PR #3193, opened 2015-04-05
provider/user/dscl: Set "comment" default value

legal90:
I suggest to set `comment` attribute to the value of `username` if it isn't specified (in Dscl provider only).

It is not critical thing in OS X, but if `comment` is not set, then this command will set "RealName" to the empty string:
https://github.com/chef/chef/blob/12.2.1/lib/chef/provider/user/dscl.rb#L201
```
run_dscl("create /Users/#{@new_resource.username} RealName '#{@new_resource.comment}'")
```
And it looks bad in OS X gui:
os_x_empty_realname

So, it will be better to prevent this case and guarantee that RealName will be always set to the readable value.

### DIFF
--- a/lib/chef/provider/user/dscl.rb
+++ b/lib/chef/provider/user/dscl.rb
@@ -195,9 +195,10 @@ user password using shadow hash.")
 
         #
         # Saves the specified Chef user `comment` into RealName attribute
-        # of Mac user.
+        # of Mac user. If `comment` is not specified, it takes `username` value.
         #
         def dscl_create_comment
+          @new_resource.comment(@new_resource.username) if @new_resource.comment.nil?
           run_dscl("create /Users/#{@new_resource.username} RealName '#{@new_resource.comment}'")
         end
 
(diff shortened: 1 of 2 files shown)


coderanger:
What does OS X do if RealName is not set at all? We could also fix it to not set RealName if comment is nil.

lib/chef/provider/user/dscl.rb
```diff
@@ -195,9 +195,10 @@ def dscl_create_user
 
         #
         # Saves the specified Chef user `comment` into RealName attribute
-        # of Mac user.
+        # of Mac user. If `comment` is not specified, it takes `username` value.
         #
         def dscl_create_comment
+          @new_resource.comment(@new_resource.username) if @new_resource.comment.nil?
```
legal90: @coderanger I've tried to replace this line by that:
```
return if @new_resource.comment.nil?
```
but got the same result as on the screenshot: "RealName" is empty and is not displayed in GUI

legal90:
Chefs, can you please merge it?

legal90:
@natewalck @mikedodge04 Could you please review this PR? Is it OK for you?

natewalck:
This looks good to me. It is far better to have the username as RealName than an empty string. Thanks for fixing this! :+1:

legal90:
@nathenharvey @lamont-granquist Could you please take a look at this PR?

lamont-granquist:
:+1:

thommay:
:+1:

PR #2338, opened 2014-11-01
fixes access keys attribute in knife show 

justanshulsharma:
this fixes #1974

PR #3195, opened 2015-04-06
Pass name by knife cil attribute

sawanoboly:
Hi,

We can't pass attribute which named `name` since this change https://github.com/chef/chef/pull/2338 via knife ssh.

I would like to fix it.

### DIFF
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -181,8 +181,8 @@ class Chef
             # Must check :[] before attr because spec can include
             #   `keys` - want the key named `keys`, not a list of
             #   available keys.
-            elsif data.respond_to?(:[])
-              data = data[attr] ? data[attr] : data.send(attr.to_sym)
+            elsif data.respond_to?(:[])  && data.has_key?(attr)
+              data = data[attr]
             elsif data.respond_to?(attr.to_sym)
               data = data.send(attr.to_sym)
             else
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -182,7 +182,7 @@ class Chef
             #   `keys` - want the key named `keys`, not a list of
             #   available keys.
             elsif data.respond_to?(:[])
-              data = data[attr]
+              data = data[attr] ? data[attr] : data.send(attr.to_sym)
             elsif data.respond_to?(attr.to_sym)
               data = data.send(attr.to_sym)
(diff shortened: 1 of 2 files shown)


ranjib:
:+1:
chef_environment and other node attributes that are computed using method will have this issue as well.

sawanoboly:
@ranjib Yes :smiley: 

I wrote a spec for this issue to cover compatibility on future.

And I'll run integration test to avoid same problem with my plugin like below. This build is scheduled at every night by my heroku scheduler :+1: 

https://circleci.com/gh/higanworks/knife-zero/20

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -182,7 +182,7 @@ def extract_nested_value(data, nested_value_spec)
             #   `keys` - want the key named `keys`, not a list of
             #   available keys.
             elsif data.respond_to?(:[])
-              data = data[attr]
+              data = data[attr] ? data[attr] : data.send(attr.to_sym)
```
danielsdeleo: Looks like this is basically duplicating the next `elsif` clause ( `elsif data.respond_to?(attr.to_sym)` ). Would it be better to change the `elsif` clause on line 184 to `data.respond_to?(:[]) && data.has_key?(attr)` ?

danielsdeleo: Otherwise we probably need to guard this with a `respond_to?` which would then completely duplicate the next `elsif` statement.

tyler-ball: Agreed about 
```ruby
elsif data.respond_to?(:[]) && data.has_key?(attr)
  data = data[attr]
elsif data.respond_to?(attr.to_sym)
  data = data.send(attr.to_sym)
```

sawanoboly:
@danielsdeleo @tyler-ball

Thanks, that is better. I'll update my branch after a few hours.

sawanoboly:
I've updated code to reflect comments :relaxed:

danielsdeleo:
:+1: from me. Would be nice to have a test that covers `--attributes` when the attribute doesn't exist and also there is no method with that name either. But I'd be okay to add that during merge.

tyler-ball:
:+1:

btm:
:+1:

I presume @danielsdeleo's comment means he's going to add a test and merge this?

danielsdeleo:
Merged w/ 182f3760dd4d81033d57cfc4d36b2fb6fc848e32

PR #3196, opened 2015-04-06
Lock to win32-api ~> 1.5.2 for Ruby 2.1 and 2.2 support

tknerr:
As of ChefDK 0.5.rc3 which ships with Ruby 2.1.5 the win32-api dependency should be locked to ~> 1.5.2

This is because 1.5.2 is the first version which [includes the prebuilt binaries for Ruby 2.1 and 2.2](https://github.com/djberg96/win32-api/issues/10#issuecomment-58742838)

ChefDK 0.5.3 actually already comes with `1.5.3 universal-mingw32`, but it might still be worth locking it to 1.5.2 minimum, just to prevent potential errors

See also related chef/chef-dk#41


lamont-granquist:
this got bumped to 1.5.3

PR #3126, opened 2015-03-24
Allow for the chef gem installation to succeed without elevated privileges

tknerr:
As discussed in #3125 the registration in the windows event log requires elevated privileges, which prevents the installation of the chef gem under a standard user account.

With this PR we do a best-effort approach and continue even if the registration in the event log fails.

/cc @jdmundrawala

PR #3203, opened 2015-04-07
Apply changes from #3126 to 12.2-stable branch

tknerr:
Applying the changes from #3126 to the 12.2-stable branch so the chef gem becomes installable again on Windows within the 12.2.x line of releases already.

If you see this as a non-risk change it would be nice to have that with 12.2.x already. No idea when 12.3 is about to be released, but 12.2.2 might be quite a bit earlier?


jdmundrawala:
The person who does the next release will take care of this. It's unlikely there will be another 12.2 release

tknerr:
this was released with 12.3.0 - no need to apply it to 12-2-stable anymore

Issue #3180, opened 2015-04-02
Exception handlers do not run if chef_run fails before @run_status is set


Igorshp:
Was digging around trying to figure out weird exception handler behaviour:

recipe does not exist on chef server - exception handler fire
role does not exist on chef server - exception handler does *NOT* fire

In the end I traced it to this line:
https://github.com/chef/chef/blob/master/lib/chef/client.rb#L438
and run_status is set here:
https://github.com/chef/chef/blob/master/lib/chef/client.rb#L244

This means that anything north of this line, does not fire a sensu handler:
https://github.com/chef/chef/blob/master/lib/chef/client.rb#L438

It feels like run_failed should always run regardless of run_status being populated fully (empty run_status instance appears to work just fine).

The goal is to get the exception handlers to fire in most (hopefully all) situations.

Any thoughts on how to best fix this?

lamont-granquist:
seems to be fixed in 12.4.0

PR #3207, opened 2015-04-09
patch to always run exception handlers

Igorshp:
Patch for issue https://github.com/chef/chef/issues/3180

Actual problem:
missing recipe present in run_list - exception handlers work as expected
missing role present in run_list - exception handlers do not fire.

Exception handlers require @run_status to be set.
Moving the declaration to initialisation block ensures it's always available.

'node' is nil at that stage and is set again here https://github.com/chef/chef/blob/master/lib/chef/client.rb#L244

big thanks to @jonlives, Coderanger and @thommay your help

### DIFF
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -152,7 +152,6 @@ class Chef
     def initialize(json_attribs=nil, args={})
       @json_attribs = json_attribs || {}
       @node = nil
-      @run_status = nil
       @runner = nil
       @ohai = Ohai::System.new
 
@@ -162,6 +161,7 @@ class Chef
       @events = EventDispatch::Dispatcher.new(*event_handlers)
       @override_runlist = args.delete(:override_runlist)
       @specific_recipes = args.delete(:specific_recipes)
+      @run_status = Chef::RunStatus.new(node, events)
 
       if new_runlist = args.delete(:runlist)
         @json_attribs["run_list"] = new_runlist
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -241,7 +241,8 @@ class Chef
     # Chef::Node:: The updated node object
     def build_node
       policy_builder.build_node
-      @run_status = Chef::RunStatus.new(node, events)
+      @run_status.node = node
+      @run_status.events = events
       node
     end
 
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -242,7 +242,6 @@ class Chef
     def build_node
       policy_builder.build_node
       @run_status.node = node
-      @run_status.events = events
       node
     end
 
--- a/spec/unit/client_spec.rb
+++ b/spec/unit/client_spec.rb
@@ -819,4 +819,24 @@ describe Chef::Client do
     end
 
   end
+
+  describe "should always attempt to run handlers" do
+    let(:e) { NoMethodError }
+    subject { client }
+    before do
+      # fail on the first thing in begin block
+      allow_any_instance_of(Chef::RunLock).to receive(:save_pid).and_raise(e)
+    end
+
+    it "should run exception handlers on early fail" do
+      expect(subject).to receive(:run_failed)
+      exception = nil
+      begin
+        subject.run
+      rescue e => ex
+        exception = ex
+      end
+      expect(exception).to be_instance_of(e)
+    end
+  end
 end
--- a/spec/unit/client_spec.rb
+++ b/spec/unit/client_spec.rb
@@ -820,7 +820,7 @@ describe Chef::Client do
 
   end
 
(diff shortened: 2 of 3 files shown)


thommay:
As I said on IRC, I think the patch is good but I'd love to see a test.

Igorshp:
@thommay added a test for exact problem that started the PR. To ensure that exception handlers get run if we fail any time inside 'begin' block of run function.

thommay:
:+1: this looks good to me

spec/unit/client_spec.rb
```diff
@@ -819,4 +819,24 @@ def stub_for_audit
     end
 
   end
+
+  describe "should always attempt to run handlers" do
+    let(:e) { NoMethodError }
```
ranjib: nitpick, remove the `should`

ranjib:
:+1:

lamont-granquist:
Are you missing dropping the second declaration out of `#build_node`?

Igorshp:
@lamont-granquist no, build node mutates @node and passes it to @run_status.
the declaration inside #build_node is correct.

This patch is simply to allow exception handlers to fire before that happens, even if full node data is not available. (from what i've seen, it doesn't impact messaging in any way)

lamont-granquist:
So build_node actually creates the node object, and then build_node is responsible for wiring it up to the RunStatus.  The concern that I've got is that to do that, you are throwing away the old RunStatus object and creating a new one to wire it up.  Right now there's no state being saved in the first RunStatus object, but it seems like its asking for trouble to be doing that.  I'd probably be better to introduce a setter on the RunStatus to send it the node object in #build_node rather than toss away the first RunStatus object and create a new one.

Igorshp:
@lamont-granquist that's a good point, I agree.

I have made a patch that adds attr_accessor to @node and @events of run_status so they can be set inside #build_node.

lib/chef/client.rb
```diff
@@ -241,7 +241,8 @@ def load_node
     # Chef::Node:: The updated node object
     def build_node
       policy_builder.build_node
-      @run_status = Chef::RunStatus.new(node, events)
+      @run_status.node = node
+      @run_status.events = events
```
lamont-granquist: the events object should be wired up correctly in the initializer already i think.

lamont-granquist:
Yeah that looks better, i think that one line can probably be dropped, but that addresses my issue...

:+1:

Igorshp:
It looked like @vents object was mutated at several places before it gets to #build_node.

I'm reading into how Eventdispatcher works to see if that line can be dropped.

Igorshp:
@lamont-granquist 
Looks like subscribes in events get updated from here
https://github.com/chef/chef/blob/4827c9d31e68eafdff8d865d40479a18e6382793/lib/chef/client.rb#L434
that function will eventually lead to:
https://github.com/chef/chef/blob/4827c9d31e68eafdff8d865d40479a18e6382793/lib/chef/client.rb#L222

So it seems like that line is required.

lamont-granquist:
Its the same events object though, so the reference that you're assigning should be the same.

Igorshp:
@lamont-granquist 
I've done some tests to see what the value of @subscribers inside the @events is at various points and you are indeed right - the contents of @events inside main loop is same as inside @run_status. There is no need to set events again.

I'll update the PR shortly

Igorshp:
@lamont-granquist do you have any other comments on this PR? how can we move this forward?

spec/unit/client_spec.rb
```diff
@@ -819,4 +819,24 @@ def stub_for_audit
...
+
+    it "should run exception handlers on early fail" do
+      expect(subject).to receive(:run_failed)
+      exception = nil
+      begin
+        subject.run
+      rescue e => ex
+        exception = ex
+      end
+      expect(exception).to be_instance_of(e)
```
btm: Could this all be just ```expect { subject.run }.to raise_error(NoMethodError)```?

Igorshp:
@btm I tried your suggestion, but I cannot get it to work. Could you show a full snippet of how it's supposed to work?
In the meantime, I have cleaned up the test a little to improve readability and reduce clutter.

lamont-granquist:
Looks fine.  :+1:

lamont-granquist:
Oh except, yeah you should use what @btm wrote.

Did you make sure to use curly braces in the `expect { }.to raise_error`?  Because that should do exactly what your code does...

Igorshp:
@lamont-granquist bah, you're right, it was the curly braces! i was doing the standard expect().to
Updated the PR, thanks!

btm:
:+1:

Igorshp:
restarting appveyor/travis builds

Issue #3146, opened 2015-03-25
change InvalidCookbookVersion exception message


anjutiwari:
The version exception message is confusing.

when the version value in metadata.rb is 
version "x.y.z"

knife cookbook upload -o mycookbook
ERROR: Chef::Exceptions::InvalidCookbookVersion: 'x.y.z' does not match 'x.y.z' or 'x.y'

using negative version now :(-2.2.0)
ERROR: Chef::Exceptions::InvalidCookbookVersion: '-2.2.0' does not match 'x.y.z' or 'x.y'

so that's mean version "x.y.z" is a acceptable and valid version value as per the above exception message.

The exception message should change , which can clearly explain from the exception message that 
x,y and z are positive numbers.
example:- 
ERROR: Chef::Exceptions::InvalidCookbookVersion: 'x.y.z' .  Version numbers always follow a simple three-number version sequence. Example (version "1.6.0")

nysthee:
I don't think that I quite understand what you mean with this. The example you're giving doesn't include a a negative number, so it shouldn't  output the given error?

anjutiwari:
@nysthee 

using negative version now :(-2.2.0)
ERROR: Chef::Exceptions::InvalidCookbookVersion: '-2.2.0' does not match 'x.y.z' or 'x.y'

nysthee:
I personally don't see this as an improvement. To me it's clear that semantic versioning http://semver.org/ is used as a standard.

PR #3212, opened 2015-04-10
Modified exception message for invalid version

anjutiwari:
Related issue : #3146


thommay:
I agree with the comments in the bug that https://semver.org is pretty clear how this should work, and I'm not sure this fixes a real problem we have.
Thank you for your contribution.

PR #3213, opened 2015-04-10
Fixes 2140 - Honor Proxy from Env

jonsmorrow:
This change adds support to Chef's core http libs to
honor a proxy uri set in the environment not just
chef/knife config. It also adds support for username
and password both in uri and env.

The order of precidence for values is uri, env, config.

### DIFF
--- a/lib/chef/http/basic_client.rb
+++ b/lib/chef/http/basic_client.rb
@@ -97,7 +97,9 @@ class Chef
 
       #adapted from buildr/lib/buildr/core/transports.rb
       def proxy_uri
-        proxy = Chef::Config["#{url.scheme}_proxy"]
+        proxy = Chef::Config["#{url.scheme}_proxy"] ||
+                env["#{url.scheme.upcase}_PROXY"] || env["#{url.scheme}_proxy"]
+
         # Check if the proxy string contains a scheme. If not, add the url's scheme to the
         # proxy before parsing. The regex /^.*:\/\// matches, for example, http://.
         proxy = if proxy.match(/^.*:\/\//)
@@ -105,7 +107,8 @@ class Chef
         else
           URI.parse("#{url.scheme}://#{proxy}")
         end if String === proxy
-        excludes = Chef::Config[:no_proxy].to_s.split(/\s*,\s*/).compact
+        no_proxy = Chef::Config[:no_proxy] || env['NO_PROXY'] || env['no_proxy']
+        excludes = no_proxy.to_s.split(/\s*,\s*/).compact
         excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
         return proxy unless excludes.any? { |exclude| File.fnmatch(exclude, "#{host}:#{port}") }
       end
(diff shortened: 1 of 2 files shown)


chef-supermarket:

* @jonsmorrow

lib/chef/http/basic_client.rb
```diff
@@ -97,7 +97,7 @@ def request(method, url, req_body, base_headers={})
 
       #adapted from buildr/lib/buildr/core/transports.rb
       def proxy_uri
-        proxy = Chef::Config["#{url.scheme}_proxy"]
+        proxy = ENV["#{url.scheme}_proxy"] || Chef::Config["#{url.scheme}_proxy"]
```
danielsdeleo: We should find out if the uppercase version might be a thing, maybe on windows? @lamont-granquist added uppercase env vars to our existing code that set proxy env vars in here: https://github.com/chef/chef/commit/b6b0a7ab02212a98ac6c3e100bcd9dafcf604aca but there's nothing in the commit about why uppercase was necessary.

jonsmorrow: Let me just add it.

lamont-granquist:
I would sort of naively expect the precedence to be uri, config, env because the config is a bit more specific and i'd think it would override the global env setting.  Other than that, it looks fine.

(And I don't know what the diff is between http_proxy and HTTP_PROXY, but it doesn't seem to hurt).

jonsmorrow:
@lamont-granquist hmmm I guess my thought was that the config file is more static and as such we should honor what a user has set in the ENV. But I am happy to swap it if that is better in the chef world sense.

lamont-granquist:
As a user invoking chef-client from the command line I can see that, but from a system administration perspective the env vars that that the daemon gets would be set in /etc/environments or the root .bashrc or something similar to that which was more 'global' while i'd expect to be able to override that in the chef config.

jonsmorrow:
@lamont-granquist I buy that let we change that ordering and re-submit.

danielsdeleo:
FWIW I checked the codebase and we don't have any other place I could find where we set config both by environment and config, which means we don't have any concerns about consistency to worry about. Carry on :smile:

danielsdeleo:
A unit test would be awesome, we already have tests for the proxy stuff here so it should be easy enough to add one or more that demonstrate the correct behavior when using env vars: https://github.com/chef/chef/blob/master/spec/unit/http/basic_client_spec.rb#L42

chef-supermarket:

danielsdeleo:
I added some tests, I think this is good to go @chef/client-core -- :+1: from me. I haven't personally had to deal with proxies very much, so if the way I set the environment in the tests doesn't test what people would normally do, please let me know.

btm:
FYI @carpnick

jonsmorrow:
@danielsdeleo We will have one additional update. We need to do the same thing for no proxy.

lib/chef/http/basic_client.rb
```diff
@@ -105,7 +107,8 @@ def proxy_uri
         else
           URI.parse("#{url.scheme}://#{proxy}")
         end if String === proxy
-        excludes = Chef::Config[:no_proxy].to_s.split(/\s*,\s*/).compact
+        no_proxy = Chef::Config[:no_proxy] || ENV['NO_PROXY'] || ENV['no_proxy']
```
lamont-granquist: can we use `#env` here as well?

tyler-ball:
:+1:

afiune:
:+1:

PR #3214, opened 2015-04-10
Chef Key Object

tylercloke:
Implemented and added spec for the Key object corresponding to the key rotation API endpoint.

Blocked pending chef/oc_erchef#126 merge.

Ping @chef/lob

### DIFF
--- a/lib/chef/key.rb
+++ b/lib/chef/key.rb
@@ -1,0 +1,251 @@
+#
+# Author:: Tyler Cloke (tyler@chef.io)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/json_compat'
+require 'chef/mixin/params_validate'
+require 'chef/exceptions'
+
+class Chef
+  # Class for interacting with a chef key object. Can be used to create new keys,
+  # save to server, load keys from server, list keys, delete keys, etc.
+  #
+  # @author Tyler Cloke
+  #
+  # @attr [String] actor           the name of the client or user that this key is for
+  # @attr [String] name            the name of the key
+  # @attr [String] public_key      the RSA string of this key
+  # @attr [String] private_key     the RSA string of the private key if returned via a POST or PUT
+  # @attr [String] expiration_date the ISO formatted string YYYY-MM-DDTHH:MM:SSZ, i.e. 2020-12-24T21:00:00Z
+  # @attr [String] rest            Chef::REST object, initialized and cached via chef_rest method
+  # @attr [string] api_base        either "users" or "clients", initialized and cached via api_base method
+  #
+  # @attr_reader [String] actor_field_name must be either 'client' or 'user'
+  class Key
+
+    include Chef::Mixin::ParamsValidate
+
+    attr_reader :actor_field_name
+
+    def initialize(actor, actor_field_name)
+      # Actor that the key is for, either a client or a user.
+      @actor = actor
+
+      unless actor_field_name == "user" || actor_field_name == "client"
+        raise Chef::Exceptions::InvalidKeyArgument, "the second argument to initialize must be either 'user' or 'client'"
+      end
+
+      @actor_field_name = actor_field_name
+
+      @name = nil
+      @public_key = nil
+      @private_key = nil
+      @expiration_date = nil
+      @create_key = nil
+    end
+
+    def chef_rest
+      @rest ||= if @actor_field_name == "user"
+                  Chef::REST.new(Chef::Config[:chef_server_root])
+                else
+                  Chef::REST.new(Chef::Config[:chef_server_url])
+                end
+    end
+
+    def api_base
+      @api_base ||= if @actor_field_name == "user"
+                      "users"
+                    else
+                      "clients"
+                    end
+    end
+
+    def actor(arg=nil)
+      set_or_return(:actor, arg,
+                    :regex => /^[a-z0-9\-_]+$/)
+    end
+
+    def name(arg=nil)
+      set_or_return(:name, arg,
+                    :kind_of => String)
+    end
+
+    def public_key(arg=nil)
+      raise Chef::Exceptions::InvalidKeyAttribute, "you cannot set the public_key if create_key is true" if !arg.nil? && @create_key
+      set_or_return(:public_key, arg,
+                    :kind_of => String)
+    end
+
+    def private_key(arg=nil)
+      set_or_return(:private_key, arg,
+                    :kind_of => String)
+    end
+
+    def delete_public_key
+      @public_key = nil
+    end
+
+    def create_key(arg=nil)
+      raise Chef::Exceptions::InvalidKeyAttribute, "you cannot set create_key to true if the public_key field exists" if arg == true && !@public_key.nil?
+      set_or_return(:create_key, arg,
+                    :kind_of => [TrueClass, FalseClass])
+    end
+
+    def expiration_date(arg=nil)
+      set_or_return(:expiration_date, arg,
+                    :regex => /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z|infinity)$/)
+    end
+
+    def to_hash
+      result = {
+        @actor_field_name => @actor
+      }
+      result["name"] = @name if @name
+      result["public_key"] = @public_key if @public_key
+      result["private_key"] = @private_key if @private_key
+      result["expiration_date"] = @expiration_date if @expiration_date
+      result["create_key"] = @create_key if @create_key
+      result
+    end
+
+    def to_json(*a)
+      Chef::JSONCompat.to_json(to_hash, *a)
+    end
+
+    def create
+      # if public_key is undefined and create_key is false, we cannot create
+      if @public_key.nil? && !@create_key
+        raise Chef::Exceptions::MissingKeyAttribute, "either public_key must be defined or create_key must be true"
+      end
+
+      # defaults the key name to the fingerprint of the key
+      if @name.nil?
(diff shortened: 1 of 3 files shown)


lib/chef/key.rb
```diff
@@ -0,0 +1,195 @@
...
+        openssl_key_object = OpenSSL::PKey::RSA.new(@public_key)
+        data_string = OpenSSL::ASN1::Sequence([
+                                                OpenSSL::ASN1::Integer.new(openssl_key_object.public_key.n),
+                                                OpenSSL::ASN1::Integer.new(openssl_key_object.public_key.e)
+                                              ])
+        @name = OpenSSL::Digest::SHA1.hexdigest(data_string.to_der).scan(/../).join(':')
+      end
+
+      payload = {"name" => @name, "public_key" => @public_key, "expiration_date" => @expiration_date}
+      if @actor_field_name == "user"
```
stevendanna: Rather than switching like this in all of the instance methods, perhaps something like:

```ruby
def chef_rest
  @rest ||= if @actor_field_name == "user
                   Chef::REST.new(Chef::Config[:chef_server_root])
                else
                  Chef::REST.new(Chef::Config[:chef_server_url])
              end
end

def api_base
  @api_base ||= if @actor_field_name == "user
    "users"
  else
   "clients"
  end
end
```

then, in your methods you can just do:

```
chef_rest.post_rest("#{api_base}/#{@actor}/keys")
```

with no switching.

lib/chef/key.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    def create
+      if @public_key.nil? || @expiration_date.nil?
+        raise ArgumentError.new("public_key and expiration_date fields must be populated when create is called")
+      end
+
+      # defaults the key name to the fingerprint of the key
+      if @name == nil
+        # TODO: is it safe to assume we aren't dealing with certs here?
```
stevendanna: We should extract this to a fingerprint method.  It will clean up this method a bit and I can definitely imagine the fingerprint method being useful in other places.

lib/chef/key.rb
tylercloke: Yeah, that is much smarter. I was trying to do pattern matching in Ruby last night :( What is object-oriented?

danielsdeleo:
@chef/client-core

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+        @name = generate_fingerprint(@public_key)
+      end
+
+      payload = {"name" => @name, "public_key" => @public_key, "expiration_date" => @expiration_date}
+      new_key = chef_rest.post_rest("#{api_base}/#{@actor}/keys", payload)
+      Chef::Key.from_hash(new_key)
+    end
+
+    def generate_fingerprint(public_key)
+        # TODO: is it safe to assume we aren't dealing with certs here?
```
danielsdeleo: Did you find the answer to this?

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+        data_string = OpenSSL::ASN1::Sequence([
+                                                OpenSSL::ASN1::Integer.new(openssl_key_object.public_key.n),
+                                                OpenSSL::ASN1::Integer.new(openssl_key_object.public_key.e)
+                                              ])
+        OpenSSL::Digest::SHA1.hexdigest(data_string.to_der).scan(/../).join(':')
+    end
+
+    def update
+      if @name.nil?
+        raise ArgumentError.new("the name field must be populated when update is called")
```
danielsdeleo: There's a ton of bad counter-examples in this codebase, but generally we're trying to use custom exception classes since they're more informative, e.g., `InvalidKeyAttributes` or whatever.

tylercloke: Where would such a class live? Just in this file?

lib/chef/key.rb
tylercloke: Naw its the last thing I'm unsure on.

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+    end
+
+    def expiration_date(arg=nil)
+      set_or_return(:expiration_date, arg,
+                    :regex => /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z|infinity)$/)
+    end
+
+    def to_hash
+      result = {
+        @actor_field_name => @actor,
```
mcquin: You may need to remove that comma.

tylercloke: Pretty sure trailing commas are chill in Ruby hashes:

```
irb> hash = {"foo" => "bar",}
=> {"foo"=>"bar"}
```

but if it violates y'alls ruby style, that is a different issue! Lemme know.

lib/chef/key.rb
mcquin: [This file holds all the exceptions](https://github.com/chef/chef/blob/master/lib/chef/exceptions.rb)

mcquin: Then you just `require chef/exceptions` and `raise Chef::Exceptions::InvalidKeyAttribute`

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/json_compat'
+require 'chef/mixin/params_validate'
+
+class Chef
+  class Key
```
tyler-ball: T'would be rad to add some doc here that explains the purpose of this class

lib/chef/key.rb
tylercloke: Thanks!

lib/chef/key.rb
tylercloke: Confirmed that we can just use public_keys.

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+      if @name.nil?
+        @name = generate_fingerprint(@public_key)
+      end
+
+      payload = {"name" => @name, "public_key" => @public_key, "expiration_date" => @expiration_date}
+      new_key = chef_rest.post_rest("#{api_base}/#{@actor}/keys", payload)
+      Chef::Key.from_hash(new_key)
+    end
+
+    def generate_fingerprint(public_key)
```
stevendanna: Sorry to nitpick, but I think in general we prefer for this method to not take an argument, be named `fingerprint` and just operate on the instance variable.

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+                  Chef::REST.new(Chef::Config[:chef_server_root])
+                else
+                  Chef::REST.new(Chef::Config[:chef_server_url])
+                end
+    end
+
+    def api_base
+      @api_base ||= if @actor_field_name == "user"
+                            "users"
+                          else
```
stevendanna: Spacing here is a bit odd, I think github destroyed the formatting of my original comment.

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+      if @name.nil?
+        raise ArgumentError.new("the name field must be populated when update is called")
+      end
+
+      new_key = chef_rest.put_rest("#{api_base}/#{@actor}/keys/#{@name}", to_hash)
+      Chef::Key.from_hash(self.to_hash.merge(new_key))
+    end
+
+    def save
+      begin
```
stevendanna: Minor style point: The begin block here is unnecessary:

```
def save
   ...
rescue Net::HTTPServerException =>
  ...
end
```

lib/chef/key.rb
```diff
@@ -0,0 +1,203 @@
...
+      result
+    end
+
+    def to_json(*a)
+      Chef::JSONCompat.to_json(to_hash, *a)
+    end
+
+    def create
+      if @public_key.nil? || @expiration_date.nil?
+        raise ArgumentError.new("public_key and expiration_date fields must be populated when create is called")
```
marcparadise: The server has ruls for this in place, and - unless there's significant server load benefit - I'm reluctant to duplicate our non-client-specific validation rules in more than one place since that means the rules  it in multiple places.  Would it make sense to just let the server report back on this kind of error?

marcparadise: related thoughts:

* expiration_date is optional on the server, and will cause the key to be created with infinity date. Do we want to not provide this feature to the caller?
*  in the soon-to-be-merged department, we also support server side key generation by providing `"create_key": true` instead of `public_key`.

tyler-ball:
:+1: other than the minor comments already listed

lib/chef/key.rb
danielsdeleo: I think very basic duplication like ensuring required fields are present is okay, and gives you a place to explain what's wrong in terms that are relevant to the client implementation code. Trying to synchronize exact format rules and such is where you're more likely to get into trouble, IMO.

tylercloke:
Hrm, this PR is kind of blocked by https://github.com/chef/oc_erchef/pull/126 as I'll need to extend this functionality as soon as that is merged (server-side key generation). I might wait and update it once I have a build of that. Thanks for all the review!

lib/chef/key.rb
tylercloke: Maybe I'll add a utility class function self.generate_fingerprint and a similar object function that calls the class function but passes in @public_key. Seems like this is nice to have as a utility function and you can't do it that way if you use an instance variable.

lib/chef/key.rb
tylercloke: 1) I'll remove the need to have expiration_date defined.
2) I think basic client-side key validation is a sane level of duplication and nice to have.
3) Yeah, I'm going to wait for you to merge then update this PR.

tylercloke:
Ping @stevendanna ^ added some of the create_key stuff we discussed.

lib/chef/key.rb
```diff
@@ -0,0 +1,236 @@
...
+
+      payload = {"name" => @name}
+      payload['public_key'] = @public_key unless @public_key.nil?
+      payload['create_key'] = @create_key if @create_key
+      payload['expiration_date'] = @expiration_date unless @expiration_date.nil?
+      new_key = chef_rest.post_rest("#{api_base}/#{@actor}/keys", payload)
+      Chef::Key.from_hash(new_key)
+    end
+
+    def generate_fingerprint_from_public_key_instance
```
stevendanna: I think we may need to agree to disagree on the name of this function :)

stevendanna:
Overall :+1:.

lib/chef/key.rb
tylercloke: I'm open to suggestions :P

stevendanna: I would just call it `fingerprint`.

jdmundrawala:
:+1:

PR #3217, opened 2015-04-13
explicitly require node presenter

thommay:
Fixes race condition on presenter loading in status


stevendanna:
:+1: The tests failures appear related to https://github.com/bundler/bundler/pull/3553 which @jdmundrawala has fixed but likely hasn't made it into the ruby we are using in appveyor.

danielsdeleo:
:+1:

PR #3227, opened 2015-04-15
Show Chef::VERSION at prompt_c and prompt_i on shell session

sawanoboly:
Hi, I want to always show version on shell session.

```
$ chef-shell 
loading configuration: none (standalone session)
Session type: standalone
Loading........done.

This is the chef-shell.
 Chef Version: 12.3.0.dev.0
 http://www.chef.io/
 http://docs.chef.io/

run `help' for help, `exit' or ^D to quit.

Ohai2u sawanoboriyu@!
chef (12.3.0.dev.0)> 
chef (12.3.0.dev.0)>   
chef (12.3.0.dev.0)>   recipe_mode
chef:recipe (12.3.0.dev.0)>   
```


lib/chef/shell.rb
```diff
@@ -108,9 +108,9 @@ def self.configure_irb
     irb_conf[:IRB_RC] = lambda do |conf|
       m = conf.main
 
-      conf.prompt_c       = "chef#{leader(m)} > "
+      conf.prompt_c       = "chef#{leader(m)} (#{Chef::VERSION})> "
```
btm: Looking at http://ruby-doc.org/stdlib-2.0/libdoc/irb/rdoc/IRB.html, it seems like prompt_c "prompt for continuated statement" would be excessive?

btm:
not sure about prompt_c, but :+1: otherwise.

sawanoboly:
I've found one more idea of implement. 
We'll have `leader(m)` append "(#{Chef::VERSION})" by defaut.

That implement appends Chef::VERSION to all prompt, but users can overwrite by setting Shell.env .

For example, below.

```
  def self.configure_irb
    irb_conf[:HISTORY_FILE] = Chef::Util::PathHelper.home(".chef", "chef_shell_history")
    irb_conf[:SAVE_HISTORY] = 1000

    irb_conf[:IRB_RC] = lambda do |conf|
      m = conf.main

      conf.prompt_c       = "chef#{leader(m)} > "
      conf.return_format  = " => %s \n"
      conf.prompt_i       = "chef#{leader(m)} > "
      conf.prompt_n       = "chef#{leader(m)} ?> "
      conf.prompt_s       = "chef#{leader(m)}%l> "
      conf.use_tracer     = false
    end
  end

  def self.leader(main_object)
-    env_string = Shell.env ? " (#{Shell.env})" : ""
+    env_string = Shell.env ? " (#{Shell.env})" : "(#{Chef::VERSION})"
    LEADERS[main_object.class] + env_string
  end
```

Which idea do you like ?

- Just remove from prompt_c.
- make leader(m) returns by default.

sawanoboly:
Ah, `Shell.env` includes `chef_environment`... ? I shoud not to use it.

sawanoboly:
I've removed from prompt_c, Thanks :smile:

btm:
@chef/client-core review?

jaymzh:
Love it. :+1:

danielsdeleo:
:+1:  on the code change. Personally, I'm ambivalent on the change so if other people think it's cool then let's ship it.

tyler-ball:
:+1: Lets do it unless someone doesn't like it

ranjib:
:+1:

Issue #1622, opened 2014-07-09
git resource Errno::EIO on windows


blueo:
### Version:
chef client 11.12.8

### Environment: 
Windows 7 x86 and x64, git 1.9.2.msysgit.0.

### Scenario:
I'm trying to use chef to check out a git repo to a windows client node.

### Steps to Reproduce:
I've got the following resource definition in one of my recipes:
```
git "C:\\pathtocheckout" do
    repo "https://gitserver/repo.git"
    action [ :checkout, :sync]
end
```
and including this in a recipe with nothing else will cause the error below.

### Actual Result:
I get the following error:
`Errno::EIO: git[C:\pathtocheckout] (cookbook_name::test line 21) had an error: Errno::EIO: Input/output error - CreateProcessW`

So I've manage to work around this by changing Chef::Provider::Git to include `.exe` in its calls to git eg `git ls-remote` becomes `git.exe ls-remote`. This is obviously not a great solution but I'm not sure how to go about fixing or reporting this...

smurawski:
There may be an alias or something similar to the local setup that may conflict.  We should append the .exe though and be more explicit.

docwhat:
I think the problem is actually a bug in Ruby...

According to the documentation for [CreateProcessW](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx) the contents of `lpCommandLine` may be modified for `CreateProcessW()`:

> The Unicode version of this function, `CreateProcessW`, can modify the contents of this string. Therefore, this parameter cannot be a pointer to read-only memory (such as a const variable or a literal string). If this parameter is a constant string, the function may cause an access violation.

The ruby code uses the passed in cmdline verbatim as a `const WCHAR *cmd`.

docwhat:
I created a simple patch and filed a bug with Ruby: https://bugs.ruby-lang.org/issues/11515

docwhat:
@smurawski I suspect you should remove the `easy` tag... :grinning:

docwhat:
Hmm... this may not be the same thing I'm seeing... My traceback:

```
Chef::Exceptions::MultipleFailures: Multiple failures occurred:
* SystemCallError occurred in chef run: powershell_script[RDP: enable NLA] (mycorp-role-windows::_rdp line 30) had an error: SystemCallError: powershell_script[Guard resource] (dynamically defined) had an error: SystemCallError: Invalid access to memory location. - CreateProcessW
[...]

PR #3228, opened 2015-04-15
Use git.exe instead of git on Windows

gohanman:
Issue: #1622

Using "git.exe" instead of "git" in provider on Windows platform. I expanded usage of the existing "git" method to all command strings and added a platform check.

Travis wasn't green but appeared to have the *same* build errors as the master branch as opposed to new ones.
https://travis-ci.org/gohanman/chef/builds/58540777

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -105,7 +105,7 @@ class Chef
       end
 
       def git_minor_version
-        @git_minor_version ||= Gem::Version.new(shell_out!('git --version', run_options).stdout.split.last)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
The code looks fine but there's 30 busted specs in the unit tests for the git provider which are all about "git" -> "git.exe", if those were fixed this would be good to merge.

danielsdeleo:
:+1: once appveyor is fixed.

smurawski:
This should actually be fixed via mixlib-shellout (https://github.com/chef/mixlib-shellout/pull/107) as noted in PR https://github.com/chef/chef/pull/4039.  If the updated mixlib-shellout doesn't fix this issue we can re-open this PR or #4039.

PR #3230, opened 2015-04-15
Updating appveyor to test against Ruby 2.1.5 because we're building ChefDK with that embedded for Windows

tyler-ball:
\cc @jdmundrawala

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ platform:
 environment:
   matrix:
     - ruby_version: "200"
+    - ruby_version: "21"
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -7,6 +7,7 @@ platform:
 environment:
   matrix:
     - ruby_version: "200"
+    - ruby_version: "215"
```
jdmundrawala: should probably just be `21`

lamont-granquist:
I think we had to roll this back in Chef-DK, so I think we should probably close this for now at least.

PR #3235, opened 2015-04-16
Make multipackage and arch play nicely together

### LINKED ISSUE #3233: Multipackage support breaks 'package.arch' syntax
I'm working on a fix.

jaymzh:
Fixes #3233

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -986,11 +986,28 @@ class Chef
         # Extra attributes
         #
 
+        def arch_for_name(n)
+          if @new_resource.respond_to?("arch")
+            @new_resource.arch
+          elsif @arch
+            idx = package_name_array.index(n)
+            as_array(@arch)[idx]
+          else
+            nil
+          end
+        end
+
         def arch
           if @new_resource.respond_to?("arch")
             @new_resource.arch
           else
             nil
+          end
+        end
+
+        def set_arch(arch)
+          if @new_resource.respond_to?("arch")
+            @new_resource.arch(arch)
           end
         end
 
@@ -1005,11 +1022,12 @@ class Chef
         # Helpers
         #
 
-        def yum_arch
+        def yum_arch(arch)
           arch ? ".#{arch}" : nil
         end
 
         def yum_command(command)
+          Chef::Log.debug("#{@new_resource}: yum command: \"#{command}\"")
           status = shell_out(command, {:timeout => Chef::Config[:yum_timeout]})
 
           # This is fun: rpm can encounter errors in the %post/%postun scripts which aren't
@@ -1087,16 +1105,13 @@ class Chef
             end
           end
 
-          # Don't overwrite an existing arch
-          unless arch
-            parse_arch
-          end
 
           @current_resource = Chef::Resource::Package.new(@new_resource.name)
           @current_resource.package_name(@new_resource.package_name)
 
           installed_version = []
           @candidate_version = []
+          @arch = []
           if @new_resource.source
             unless ::File.exists?(@new_resource.source)
               raise Chef::Exceptions::Package, "Package #{@new_resource.name} not found: #{@new_resource.source}"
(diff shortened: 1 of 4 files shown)


jaymzh:
The test don't pass yet... but the code works, afact. Tests to come.

jaymzh:
Note there's ANOTHER bug this fixes too... `install_remote_package` was comparing the wrong version index, so if you were actually USING multipackage it would think it needed to install things it didn't.

jaymzh:
OK, @lamont-granquist  - the only test that doesn't pass is the one where we're supposed to set the @new_resource.arch in some special circumstances which I can do but I don't actually see why we do it... everythign else now passes and this actually fixes several small bugs in multipackage.

jdmundrawala:
@jaymzh could you rebase, master was broken so some of those failing tests are not related to your changes

jaymzh:
Alright, all tests pass, and I'm using the code locally, so I feel good about it.

CC: @chef/client-core

lib/chef/provider/package/yum.rb
```diff
@@ -1059,16 +1077,13 @@ def load_current_resource
...
-          unless arch
-            parse_arch
-          end
 
           @current_resource = Chef::Resource::Package.new(@new_resource.name)
           @current_resource.package_name(@new_resource.package_name)
 
           installed_version = []
           @candidate_version = []
+          @arch = []
```
tyler-ball: I could totally be misunderstanding this, but won't this overwrite any value set by the resource?

tyler-ball:
There should be a new (or some updated) tests which show the use of `arch` as an array

lib/chef/provider/package/yum.rb
jaymzh: No, this is a class variable in the provider, but that the user passes in is stored in the new resource object (@new_resource.arch). We only overwrite that one in the one special case below, which is preserved from the original code.

jaymzh:
I'll look into getting some test coverage onto arch as an array.

jaymzh:
Tests added. And a minor cleanup of code.

@lamont-granquist @tyler-ball @chef/client-core I'd like to get this fix in for 12.3

tyler-ball:
:+1:

lib/chef/resource/yum_package.rb
```diff
@@ -38,7 +38,7 @@ def arch(arg=nil)
         set_or_return(
           :arch,
           arg,
-          :kind_of => [ String ]
+          :kind_of => [ String, Array ]
```
btm: #3233 indicates that you are using 'package.arch' in a multi-package pattern. Is this an example of what used to not work and what would now work?

```
package [ "cups.x86_64", "vim" ]
```

This line makes me wonder if this does the right thing:

```
package [ "cups", "vim" ] do
  arch [ "x86_64", "i386" ]
end
```

The first syntax seems friendlier, but I'd expect this to work too.

jaymzh: The first syntax is what I was fixing. The second syntax won't work because arch isn't an option on package, but it will work assuming you do `yum_package`

btm: Cool. If it works with yum_package, then I presume it'll actually work now or soon via the resource resolver because it'll get Chef::Resource::YumPackage either way.

jaymzh:
ping - @lamont-granquist  I know you're particularly interested in this. This is a bugfix. Would really like to get this before I miss a 12.3.1 window too. @chef/client-core

jaymzh:
Calling Mr. @lamont-granquist . Or hell... any 2nd reviewer on @chef/client-core . Anyone at at all.

btm:
:+1:

jdmundrawala:
:+1:

jaymzh:
Thanks. Added changelog and rebased. Will merge when tests are done.

ror6ax:
Still getting "no candidate version" on chef-12.3.0-1.

jaymzh:
That's because this wasn't actually merged in 12.3.0 despite the milestone.

PR #3240, opened 2015-04-18
Use Random instead of SecureRandom in request_id

jaym:
SecureRandom is quite a bit slower on Windows

Does RequestID need to use SecureRandom?

cc @lamont-granquist @stevendanna @chef/client-core

### DIFF
--- a/lib/chef/request_id.rb
+++ b/lib/chef/request_id.rb
@@ -15,7 +15,6 @@
 # limitations under the License.
 #
 
-require 'securerandom'
 require 'singleton'
 
 class Chef
@@ -31,7 +30,11 @@ class Chef
     end
 
     def generate_request_id
-      SecureRandom.uuid
+      ary = Random.new.bytes(16).unpack("NnnnnN")
+      ary[2] = (ary[2] & 0x0fff) | 0x4000
+      ary[3] = (ary[3] & 0x3fff) | 0x8000
+      "%08x-%04x-%04x-%04x-%04x%08x" % ary
     end
+
   end
 end


lib/chef/request_id.rb
```diff
@@ -15,7 +15,6 @@
 # limitations under the License.
 #
 
-require 'securerandom'
```
tyler-ball: Shouldn't there still be a `require 'random'` in the new file?

btm: looks like it's in the standard library.

btm:
We need a unit test for generate_required_id now. it "looks like a UUID."

mmzyk:
Since I saw this, the UUID generated is used by reporting. So long as it's a UUID then all should be well. Specifically, securerandom.uuid is generating a v4 UUID as outlined on wikipedia here:

https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29 

So that's the form we'd want to test around.

This stackoverflow actually has a great answer on how securerandom.uuid works: http://stackoverflow.com/questions/16650764/securerandom-uuid-vs-uuid-gem

securerandom is taking longer b/c it's trying very heard to get a random seed.  It also tries, in order:

openssl
/dev/urandom/
Win32

so it's going to fail the first two checks on Windows and try Win32. I have no idea how fast Win32 is, but it might be worth seeing if the slowdown is Ruby trying to access this. 

I suspect this change is okay, but since I saw it I thought I'd drop in some more background on what securerandom is doing and why. I can tell you that when it was picked for use by Reporting it was just picked b/c it was the easiest way to get a UUID.

jdmundrawala:
@mmzyk thanks for digging in. I've sometimes seen this issue on osx, but it's consistent on the windows boxes. I'm going to try the win32 api function call for getting a uuid, and if that's faster, I may just use that on windows and leave the other os's untouched. I think it generates a mac/timestamp based uuid (https://msdn.microsoft.com/en-us/library/windows/desktop/aa379322%28v=vs.85%29.aspx). It does say the UUID is only valid on the box, however I'm not sure how big of a deal a collision is for this. If it's only used for reporting, I'm thinking it's going to be ok since you'll have both machine name and the uuid.

lib/chef/request_id.rb
tyler-ball: Learning opportunity for me: how can you tell it is in the standard library?

Looking at http://ruby-doc.org/core-2.2.0/Random.html it seems the only difference from http://ruby-doc.org/stdlib-2.2.0.preview1/libdoc/securerandom/rdoc/SecureRandom.html is that `Secure` is in core, while `SecureRandom` is in stdlib.  Do we not need to require things in core?

btm: I just loaded irb and typed ```Random.new``` and it worked, then I tried ```require "secure"``` and that failed. That's why I said "looks like" :)

mmzyk:
@jdmundrawala I'm not sure what you mean by there'll be both machine name and the uuid. By machine name you mean the node id? While that is true, that's not how the reporting db stores things - it uses the uuid as the primary key. 

https://github.com/chef/oc_reporting/blob/master/db/deploy/base%401.2.0.sql#L22

While that table does have org id and node id as well, the table that stores the full details of a run only uses the uuid as the primary key and reference point back to the main table: https://github.com/chef/oc_reporting/blob/master/db/deploy/base%401.2.0.sql#L47

This is why I'm suggesting we be hesitant in straying too far from true uuids, since we don't want to naively create something that might introduce collisions at a higher rate. While I get that the chances of collisions is likely very low, I just want to make sure we use the right amount of caution around changing this. (But I also understand we need to do something to make run times not be abnormally high for windows users).

lamont-granquist:
I'm satisfied that the randomness of Random will be good enough to avoid any issues with birthday collisions in the uuid generation.  With either implementation we should still require close enough to 2^61 uuids to generate a collision.

lamont-granquist:
:+1: if we get tests.

thommay:
guess the tests should just be that we set the version number and the reserved bits correctly. Otherwise I'm happy to do this.

jaym:
this still scares me a little, let's come back to this when we've fixed all other perf issues related to windows

PR #3242, opened 2015-04-19
[RFC-039] chef handler dsl

ranjib:
Allows
```ruby
Chef.event_handler do
  on :run_failed do |exception|
    # ... hipchat_notify exception.message
  end
end
```


jdmundrawala:
This is awesome :+1:

thommay:
Awesome to the max! :+1:

lib/chef/event_dispatch/dsl.rb
```diff
@@ -0,0 +1,46 @@
...
+      # with user defined block against a chef event
+      #
+      # @return [Chef::EventDispatch::Base] a base handler object
+      def on(event_type, &block)
+        validate!(event_type)
+        handler = Chef::EventDispatch::Base.new
+        handler.define_singleton_method(event_type) do |*args|
+          block.call(args)
+        end
+        Chef::Config[:event_handlers] << handler
```
danielsdeleo: This will only work in a config file, right? You should be able to see if `Chef.run_context` exists yet and then register the handler in the middle of a run, as we discussed on the RFC.

danielsdeleo:
This LGTM as is, but IIRC you wanted to be able to define these in library files, too.

jonlives:
YES! :+1:

lib/chef/event_dispatch/dsl.rb
ranjib: should i raise an exception if  Chef.run_context is nil?

ranjib:
@danielsdeleo yup, i was intending to use it from libraries. what about core chef features, if i do write a telemetry system based on event handlers, do you think i should use this API ? or stick to raw event handlers api similar to doc formatters etc.

lib/chef/event_dispatch/dsl.rb
danielsdeleo: If you intend for this to also be usable from, say, client.rb (would be necessary to catch events very early in the run) then you will need to tolerate `Chef.run_context` being `nil`. IIRC we no longer support `--no-fork` with any kind of looping chef run, so you should be okay to both append to `Chef::Config[:event_handlers]` (works for the `client.rb`, case, ignored by the cookbook case) and do something like:

```ruby
# method names might be wrong, doing this from memory.
Chef.run_context.events.formatter.register(handler) if Chef.run_context
```

...to handle the load-from-cookbook case.

danielsdeleo:
@ranjib My mental model for this is it's just a DSL for event handler classes that also takes care of registering the handler for you. It sounds perfectly fine to me to do the same for telemetry stuff, e.g., at the core there's a thing where you write a class with some methods, but you can use a DSL instead if you prefer.

ranjib:
@danielsdeleo i have added Chef.run_context checks, and use events.register when available. I have added logic to fall back to Chef::Config[:event_handlers] in case we use the same API for defining handlers in client config file. Added a spec as well

lib/chef/event_dispatch/dsl.rb
```diff
@@ -0,0 +1,54 @@
...
+      # Adds a new event handler derived from base handler
+      # with user defined block against a chef event
+      #
+      # @return [Chef::EventDispatch::Base] a base handler object
+      def on(event_type, &block)
+        validate!(event_type)
+        handler = Chef::EventDispatch::Base.new
+        handler.define_singleton_method(event_type) do |*args|
+          block.call(args)
+        end
```
danielsdeleo: It's a little tricky to follow the different block contexts here. What happens if you do:

```ruby
Chef.event_handler do
  on(:run_start) { @analytics_connection = MyAnalyticsSystem::Connection.new }

  on(:other_event) { |data| @analytics_connection.send(data) }
end
```
Also, do methods defined inside `Chef.event_handler` resolve in the way you'd expect?

lib/chef/event_dispatch/dsl.rb
```diff
@@ -0,0 +1,54 @@
...
+        # Use current event dispacth system is run_context and associated event
```
danielsdeleo: typo "dispacth"

lib/chef/event_dispatch/dsl.rb
danielsdeleo: Also, it'd be nice to have tests for this kind of thing.

lib/chef/event_dispatch/dsl.rb
```diff
@@ -0,0 +1,54 @@
...
+class Chef
+  module EventDispatch
+    class DSL
+      # Adds a new event handler derived from base handler
+      # with user defined block against a chef event
+      #
+      # @return [Chef::EventDispatch::Base] a base handler object
+      def on(event_type, &block)
+        validate!(event_type)
+        handler = Chef::EventDispatch::Base.new
```
danielsdeleo: For debugging purposes, tt would be nice to take a name for the handler so it's not an anonymous class. That also would make it easier to test outside of a chef-client run.

lib/chef/event_dispatch/dsl.rb
```diff
@@ -0,0 +1,54 @@
...
+          Chef::Log.debug('Registering handler using config, this will only work inside config file')
+          Chef::Config[:event_handlers] << handler
+        end
+        handler
+      end
+
+      private
+      def validate!(event_type)
+        all_event_types = (Chef::EventDispatch::Base.instance_methods - Object.instance_methods)
+        raise Chef::Exceptions::UnknownEventType unless all_event_types.include?(event_type)
```
danielsdeleo: This should include a message stating what the invalid event type is, and probably a list of what the valid values are. Otherwise a simple typo could be frustrating to debug.

lib/chef/event_dispatch/dsl.rb
ranjib: since each handler code block is invoked inside a new instance of Chef::EventDispatch::Base, so i think this wont work. If you want to persist state you can use a custom object, i expect that too work due to closures. i.e. 
```ruby
Chef.event_handler do
foo = Foo.new
  on(:run_start) { foo.analytics_connection = MyAnalyticsSystem::Connection.new }
  on(:other_event) { |data| foo.analytics_connection.send(data) }
end
```
i'll validate this with specs, but does that work or you think setting instance variables is required. Cause in that route we'll have to share a single handler instance for all user defined callbacks.

lib/chef/event_dispatch/dsl.rb
ranjib: i can subclass Chef::EventDispatch::Base and introduce a name instance variable there. You want users to specify the name, or Chef internally to construct it? This is tied with the earlier strategy. i,e  it will be only useful if we want to share a single handler class for overriding all user defined callbacks

lib/chef/event_dispatch/dsl.rb
ranjib: yup, i'll fix it, this is not explanatory :-/
thanks for the review :-) .

lib/chef/event_dispatch/dsl.rb
danielsdeleo: The other way to go would be to create a class for each invocation of `Chef.event_handler`. For example,

```
def initialize(handler_class_name)
  @handler_class = Class.new(Chef::EventDispatch::Base)
  set_handler_class_const_in_sandbox_namespace(@handler_class)
  @handler_instance = @handler_class.new
end
```

Then the blocks you pass to the `on` method would just be the method bodies and instance variables and such should work.

ranjib:
@danielsdeleo addressed most of your comments. I have changed the logic to use a single handler instance for all call backs for individual `Chef.event_handler` invocation. I went on using `@name` instance variable instead of a class variable. Let me know what you think of the current implementation.

ranjib:
@danielsdeleo bump.

danielsdeleo:
Ok, I think this is good.

ranjib:
@chef/client-core this has gone through decent amount of changes since all of yours first review.. i think its good to go, a final round of checks?

ranjib:
@jaym can this go in 12.4.1?

coderanger:
@ranjib New feature means waiting for 12.5.

jaym:
12.4.1 is regressions only. While this looks safe, we've said no to other things and we even have changes on master that could go. Sorry. 12.5 shouldn't be too far out.

ranjib:
ack. no issues :-) . thanks for the quick reply.

danielsdeleo:
@ranjib when this is done, can you make a PR to Chef Docs? Our docs guy is looking for:

> so he should do a PR to chef-docs that has in his words the overview, list of methods (the :on_whatever things), and some examples. he can leverage the RFC. basically he just needs to point me to the right stuff.
and this topic: http://docs.chef.io/handlers.html
is where it'll end up as published
he does NOT need to get it right per chef-docs authoring guidelines
I will respond back in that PR with updates, changes, whatnots.
after it's "shipped" then he can do specific/legit PRs for tweaks going forward
but like always, I am happy to help get it over the wall

Ping @jamescott if you need a hand. Thanks!

jaym:
Also, release notes if you want to say anything in the blog post

jamescott:
I have created the placeholder docs for the 12.5 release: docs.chef.io/release/12-5/ so we can start doc'ing this now if you want @ranjib

tyler-ball:
:+1:

PR #3260, opened 2015-04-22
Update kitchen tests to use latest official test-kitchen

jdmundrawala:
Update kitchen tests to use latest official test-kitchen

cc @chef/client-core @schisamo @mcquin

### DIFF
--- a/kitchen-tests/.kitchen.travis.yml
+++ b/kitchen-tests/.kitchen.travis.yml
@@ -1,35 +1,36 @@
 ---
-driver_config:
+driver:
+  name: ec2
   aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
   aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
   aws_ssh_key_id: <%= ENV['AWS_KEYPAIR_NAME'] %>
-  iam_profile_name: <%= ENV['IAM_PROFILE_NAME'] %>
+  region: "us-west-2"
+  availability_zone: "us-west-2a"
+  security_group_ids: ["travis-ci"]
+
 
 provisioner:
-  name: chef_zero
-  github: <%= ENV['TRAVIS_REPO_SLUG'] %>
-  branch: <%= ENV['TRAVIS_COMMIT'] %>
-  require_chef_omnibus: 12.0.0-rc.2
+  name: chef_github
+  github_owner: "chef"
+  github_repo: "chef"
+  refname: <%= ENV['TRAVIS_COMMIT'] %>
   data_path: test/fixtures
 # disable file provider diffs so we don't overflow travis' line limit
   client_rb:
     diff_disabled: true
 
+transport:
+  ssh_key: <%= ENV['EC2_SSH_KEY_PATH'] %>
+
 platforms:
   - name: ubuntu-12.04
-    driver_plugin: ec2
-    driver_config:
-      region: "us-west-2"
-      availability_zone: "us-west-2a"
-      ssh_key: <%= ENV['EC2_SSH_KEY_PATH'] %>
-      security_group_ids: ["travis-ci"]
-  - name: centos-6.4
-    driver_plugin: ec2
-    driver_config:
-      region: "us-west-2"
-      availability_zone: "us-west-2a"
-      ssh_key: <%= ENV['EC2_SSH_KEY_PATH'] %>
-      security_group_ids: ["travis-ci"]
+    driver:
+      # http://cloud-images.ubuntu.com/locator/ec2/
+      image_id: ami-c15379f1
+  - name: rhel-6
+    driver:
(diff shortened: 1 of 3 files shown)


jdmundrawala:
This seems to be an issue with the latest test-kitchen and kitchen-ec2
cc @fnichol @tyler-ball

tyler-ball:
Can you try this with [my kitchen-ec2 branch](https://github.com/test-kitchen/kitchen-ec2/pull/109)?  If that works for you, I'll merge 109 and you can depend on master for now until I get 0.9 released.  I need to write unit tests and update the linters, then it will be ready for release.

tyler-ball:
@jdmundrawala I updated the yaml with configs that work for me locally.  The last travis error I saw looked like an issue reading the IAM instance profile name out of the environment variables.  Ping me tomorrow and I'll take a look at it with you.

danielsdeleo:
Looks awesome :+1:

tyler-ball:
:+1: @jdmundrawala you can remove the `echo` and squash all these commits

kitchen-tests/.kitchen.travis.yml
```diff
@@ -1,35 +1,37 @@
...
   - name: centos-6.4
-    driver_plugin: ec2
-    driver_config:
-      region: "us-west-2"
-      availability_zone: "us-west-2a"
-      ssh_key: <%= ENV['EC2_SSH_KEY_PATH'] %>
-      security_group_ids: ["travis-ci"]
+    driver:
+      # https://aws.amazon.com/marketplace/seller-profile/ref=srh_res_product_vendor?id=16cb8b03-256e-4dde-8f34-1b0f377efe89
+      image_id: ami-b3bf2f83
```
schisamo: This AMI should come from the official Canonical list at:
http://uec-images.ubuntu.com/releases/precise/release/

I believe we want the 64-bit instance root store AMI: `ami-c15379f1`.

tyler-ball: This is Centos though :confused: Although if you had the official list for that, I would appreciate it!  Is http://cloud-images.ubuntu.com/locator/ec2/ a bad resource to find Ubuntu images?

schisamo: @tyler-ball totally missed that this was CentOS. In that case we should just use the RHEL AMIs we use in CI. Take a look here for a list:
https://github.com/chef/releng-chef-repo/blob/master/script/ci#L93-L96

Also `uec-images.ubuntu.com == cloud-images.ubuntu.com`.

tyler-ball: @schisamo What is the preference for using `instance` storage over `ebs`?  Is it cheaper?

btm: My memory on this was vague, but I remembered recently getting an error when trying to use an m1.small for something random and seeing the instance types had changes. Something's changing related to instance storage over at AWS, but I don't see any clear statements about it. The "m1" instances are being replaced by "t2" instances which don't support instance storage per [this page](http://aws.amazon.com/ec2/previous-generation/). Other instances still have instance storage but they're calling it "SSD" or "HDD" now on [the pricing page](http://aws.amazon.com/ec2/pricing/.

It's always seemed to me the value of instance storage over ebs storage is that you pay extra for how much ebs storage you used, but given that you don't have a choice in the small instances anymore, I guess you have to compare the cost of a t1 to an m3+ebs?

tyler-ball:
I updated Centos to RHEL using the RelEng AMI, and updated Ubuntu to use the instance store rather than EBS.  If Travis passes, this should be :ship:able

jdmundrawala:
hmm, I'm not sure this is correct... it seems be running the wrong provisioner :(

jdmundrawala:
Maybe we should add an audit to the cookbook?

tyler-ball:
@jdmundrawala I think adding an audit would be nice (it could even validate that 2+2=4)

What do you mean its running the wrong provisioner?

tyler-ball:
Also!  Test Kitchen and kitchen-vagrant have been released - lets use those latest versions!

jdmundrawala:
It was using chef_zero instead of chef_github

btm:
@jdmundrawala:

```
Resolving dependencies...
Bundler could not find compatible versions for gem "test-kitchen":
  In Gemfile:
    kitchen-vagrant (~> 0.17.0.beta.2) ruby depends on
      test-kitchen (~> 1.4) ruby
    kitchen-ec2 (>= 0) ruby depends on
      test-kitchen (= 1.4.0.rc.1) ruby
    test-kitchen (~> 1.4.0.rc.1) ruby
The command "eval bundle install --jobs=3 --retry=3" failed 3 times.
The command "bundle install --jobs=3 --retry=3" failed and exited with 6 during .
Your build has been stopped.
```

jdmundrawala:
@btm https://github.com/test-kitchen/kitchen-ec2/pull/112

PR #3262, opened 2015-04-22
add Chef::Log::Syslog class

lamont-granquist:
See:

* https://www.kickflop.net/blog/2013/06/27/syslog-output-for-chef-runs/
* http://lists.opscode.com/sympa/arc/chef/2014-12/msg00152.html
* https://github.com/higanworks-cookbooks/chef-client_syslog

It is relatively embarrassing to have a system administration tool that
doesn't support syslog out of the box.


lamont-granquist:
@jblaine @sawanoboly any chance you could do the work to actually test this and make sure it works?

jdmundrawala:
@lamont-granquist This is cool, people have been asking for event logs in a similar manner for windows

jdmundrawala:
Along those lines, maybe this needs to get broken up into its own file so we can similarly add it for windows event logger

lamont-granquist:
Dunno, as it is it was small enough I obviously didn't think it warranted doing that...  but can obviously easily extract...

This adds a dep on syslog-logger BTW, and that appears to be BSD licensed, so that should be fine.

It also will not collide with the symbols in the chef-client_syslog cookbook and avoids monkeypatching the Syslog::Logger class directly and instead inherits and overrides which should work and be cleaner...

lamont-granquist:
Not sure about unit tests, if it works i'll have to poke around and see what i can do...

jblaine:
I will absolutely test this. I'll try to get to it 4/23.

sawanoboly:
Yes, I can test it too.

sawanoboly:
@lamont-granquist 

hmm... I seem that write(message) sends text only.

The better way of using syslog is set severity level to each message.


In last year, I thought about using syslog and wrote implementation for v11.8.0 in my book (The chef guide book[japanese]), like below.



```
## in `client.rb`
require ’syslog-logger’
Logger::Syslog.class_eval do
  attr_accessor :sync, :formatter, :close
  def write(msg)
    self.send(data[2].downcase.to_sym, data[3])
  end
end

log_level :info
log_location Logger::Syslog.new("chef-client", Syslog::LOG_DAEMON)
```

sawanoboly:
I've tried to update and write spec. Please see https://github.com/chef/chef/pull/3265 .

jblaine:
FWIW, it doesn't seem to work with 11.18.0, or I'm doing something stupid/wrong.

```
[@dev:~/src] $ sudo chef-client --version
Chef: 11.18.0
[@dev:~/src] $ sudo which chef-client
/usr/bin/chef-client
[@dev:~/src] $ sudo chef-client
[2015-04-23T16:31:17-04:00] FATAL: Configuration error TypeError: no implicit conversion of String into Integer
[2015-04-23T16:31:17-04:00] FATAL:   /etc/chef/client.rb:9:in `new'
[2015-04-23T16:31:17-04:00] FATAL:   /etc/chef/client.rb:9:in `from_string'
[2015-04-23T16:31:17-04:00] FATAL: Aborting due to error in '/etc/chef/client.rb'
[@dev:~/src] 2 $ sudo cat /etc/chef/client.rb
chef_server_url "https://api.opscode.com/organizations/jblaine"
validation_client_name "jblaine-validator"
node_name "dev"

Dir.glob(File.join("/etc/chef", "client.d", "*.rb")).each do |conf|
  Chef::Config.from_file(conf)
end

log_location Chef::Log::Syslog.new("::Syslog::LOG_DAEMON", "chef-client")
[@dev:~/src] $
```

lamont-granquist:
thanks @sawanoboly -- @jblaine try again?

jblaine:
Yup, works now.

lamont-granquist:
lolwindows...

sawanoboly:
Ah..

```
`require': cannot load such file -- syslog (LoadError)
```

sawanoboly:
Windows doesn't have syslog.

https://github.com/ruby/ruby/blob/trunk/ext/extmk.rb#L498-L503

sawanoboly:
I've created new PR to avoid windows.

lamont-granquist:
@chef/client-core @chef/client-engineers ready for review.

spec/unit/log/syslog_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  after do
+    Chef::Log.level = @old_log_level
+    Chef::Log.use_log_devices(@old_loggers)
+  end
+
+  it "should send message with severity info to syslog." do
+    expect(syslog).to receive(:info).with("*** Chef 12.4.0.dev.0 ***")
+    Chef::Log.info("*** Chef 12.4.0.dev.0 ***")
```
thommay: is this going to be hideously fragile whenever we bump versions?

jdmundrawala: it should be fine, the line after the expect has that exact string

sawanoboly: @thommay Ah, I can relate to what you are saying.
You can remove a version from this line by pull request.

thommay: Eh, no. it's fine. Read too fast and just saw the version.

jdmundrawala:
I don't really like the fact that we have to parse text to figure out the type of message (info, warn, etc). At some point, we have these going into their own methods, so we really shouldn't need to do that.

That being said, I think I've looked at this before and it'd require some heavy refactoring, so I'm going to :+1:

lamont-granquist:
@chef/client-core @chef/client-engineers ping need one more for review

jonlives:
:+1:

lamont-granquist:
@jblaine @sawanoboly merged, you might want to give master a run and make sure it all still works before it ships...

sawanoboly:
@lamont-granquist  I've tried this with local_mode on Yosemite and Ubuntu14. It's working fine.

I added line below to client.rb.

```
log_location Chef::Log::Syslog.new("chef-client", ::Syslog::LOG_DAEMON)
```

PR #3263, opened 2015-04-22
value_for_platform should use Chef::VersionConstraint::Platform

jordane:
It currently uses `Chef::VersionConstraint`, which doesn't work with some platforms, such as Fedora, where versions are reported as a single number like `21`. When `value_for_node` tries to create
a `Chef::VersionConstraint` with a single digit version constraint, it throws `Chef::Exception::CookbookInvalidVersion`


lamont-granquist:
LGTM :+1:

thommay:
:+1:

Issue #3266, opened 2015-04-23
securable_resource Windows shared_examples assume incorrect default owner/groups


dbjorge:
About a third of the Windows-desktop-specific functional test failures tracked by #3106 that I can repro on my machine are the result of:

    spec/support/shared/functional/securable_resource.rb:

    shared_examples_for "a securable resource without existing target" do
      context "on Windows", :windows_only do
        it "sets group to None on create if group is not specified" do

The example expects a group of None, but the implementation does not explicitly set any group at all on newly created files when no group is specified, which isn't quite the same. In practice, the current implementation results in the resource taking on a primary group based on the calling process token's primary group SID, which in practice is the user SID the process is running in the context of.

I think it probably makes more sense to update the shared_examples' expectations to match the actual default system behavior, rather than updating the implementation to match the shared_examples' current expectations.

There is a related issue with the tests' assumptions around default behavior for ownership - the tests assume that the owner should always be the Administrators group, but that's only mostly accurate (see https://technet.microsoft.com/en-us/library/cc961992.aspx). Both sets of examples should be updated to check chef's behavior against the actual defaults.

Issue #3106, opened 2015-03-18
Functional test failures on Windows desktops


btm:
Compiling this for @mwrock, @fnichol. It'd be helpful to know the results of `bundle exec rspec -cb spec/functional/*` for other people on Windows. @jdmundrawala? @ksubrama?  

About a year ago I was tracking down a functional test failure in [CHEF-5193](https://tickets.opscode.com/browse/CHEF-5193) and I was seeing a bunch of test failures on my desktop that I wasn't seeing on a 2012vm. The output of those failures is here: https://gist.github.com/btm/802f31fba4cad8d38b4a

Today, in an admin powershell from `C:\Users\btm_000\documents\github\chef`, I get:

```
Finished in 16 minutes 37 seconds (files took 7.65 seconds to load)
3548 examples, 40 failures, 95 pending
```

I got the same errors in an admin powershell from `C:\ruby\chef` though:

```
Finished in 14 minutes 36 seconds (files took 7.28 seconds to load)
3548 examples, 40 failures, 95 pending
```

Most of the errors are repeating shared examples for file resources. Of course, the concern is that if these fail for other people it's going to make contribution harder because they're not starting from a point of all the tests passing.

mwrock:
Took a look at this today. I pulled latest `master`, `bundle install`ed and then `bundle exec rspec -cb spec/functional/*`.

My results were:

```
Finished in 12 minutes 58 seconds (files took 9.37 seconds to load)
3552 examples, 33 failures, 91 pending
```

I am running Windows 10 build 10041. Looking at some of the failures it seems like the tests really need to run on a specific os version or even image since many errors were due to bad SIDs or pointer addresses.

dbjorge:
#3266 tracks and #3267 addresses the SID issues in the securable_resource-based specs, which is about 1/3 of what I'm seeing on my Windows 8.1 desktop. Remaining failures are https://gist.github.com/dbjorge/432105bd9a3e3ae3fc8d, haven't looked deeply into any of them.

stuartpreston:
This is so old I'm closing it.  We're down to 21 functional test failures on Windows 10, and 20 of those are probably due to a WinRM configuration that needs to be documented.  Tracking those in #7439 and #7440 instead.
```
Finished in 24 minutes 49 seconds (files took 36.29 seconds to load)
3745 examples, 21 failures, 66 pending
```

PR #3267, opened 2015-04-23
#3266 Fix bad Windows securable_resource functional spec assumptions for default file owners/groups

dbjorge:
See also issue #3106.

It wasn't 100% clear what the most correct fix path was here. The options I considered were:

1) Update the Windows file providers to explicitly set newly created files with no user-specified group to use the None group SID.
2) Update the spec to always assume the group should be SID.current_user
3) Update the spec to query for the actual defaults and check against them

I avoided 1 because there didn't seem to be any compelling reason to prefer the None-group behavior - there isn't backwards compatibility concern (because the implementation doesn't do it today) and it would be confusing for Chef to make up its own default behavior on top of the OS's already-confusing default behavior.

I avoided 2 because I was concerned about execution cases that I'm apparently not seeing - presumably this test passes in a test kitchen in its current form, which makes me suspect perhaps current_user isn't as accurate as MSDN suggests if, say, a non-interactive service is creating the resource.

This PR implements option 3. It was a little bit more code, but I'm happier with how it turned out. And it does fix about 1/3 of the functional spec failures #3106 mentions (of those I repro on my desktop, at least).


btm:
@chef/client-windows @ksubrama review?

ksubrama:
:+1:  I read the MSDN specs for the function calls made and this pull request makes sense to me.

Thanks for the work!

thommay:
@chef/client-windows can we get another review on this please?

btm:
:+1:

dbjorge:
Thanks all!

PR #3269, opened 2015-04-23
Remove method_missing and make Chef::Resource unspecial

jkeiser:
Implements https://github.com/chef/chef-rfc/pull/117


jkeiser:
@coderanger it looks like this passes all desired specs except Poise; but it doesn't appear to have made Poise any I can't find a set of Poise tests that pass against master of `chef`.  Would you mind looking into that when you get a chance?  Or tell me what git sha / tag / branch I should run?  Thanks :)

jkeiser:
@lamont-granquist @danielsdeleo I think this is ready for review.  Seems to be consensus so I wouldn't mind being ready to check it in when the RFC is approved next Thursday :)

lib/chef/dsl/recipe.rb
```diff
@@ -31,49 +33,62 @@ module DSL
...
-          rescue NameError
-            raise NameError, "No resource, method, or local variable named `#{method_symbol}' for #{describe_self_for_error}"
-          end
+        #
+        # If there is already DSL for this, someone must have called
+        # method_missing manually. Not a fan. Not. A. Fan.
+        #
+        if respond_to?(method_symbol)
+          Chef::Log.warn("Calling method_missing(#{method_symbol.inspect}) directly is deprecated in Chef 12 and will be removed in Chef 13.")
+          Chef::Log.warn("Use public_send() or send() instead.")
```
lamont-granquist: Chef::Log.deprecation

lib/chef/dsl/recipe.rb
```diff
@@ -31,49 +33,62 @@ module DSL
...
-      def evaluate_resource_definition(definition_name, *args, &block)
 
-        # This dupes the high level object, but we still need to dup the params
-        new_def = run_context.definitions[definition_name].dup
+        #
+        # If a definition exists, then Chef::DSL::Definitions.add_definition was
+        # never called.  DEPRECATED.
+        #
+        if run_context.definitions.has_key?(method_symbol.to_sym)
+          Chef::Log.warn("Definition #{method_symbol} (#{run_context.definitions[method_symbol.to_sym]}) was added to the run_context without calling Chef::DSL::Definitions.add_definition(#{method_symbol.to_sym.inspect}).  This will become required in Chef 13.")
```
lamont-granquist: Chef::Log.deprecation

lib/chef/dsl/recipe.rb
```diff
@@ -31,49 +33,62 @@ module DSL
...
+        if resource_class
+          #
+          # If the DSL method was *not* added, this is the case where the
+          # matching class implements 'provides?' and matches resources that it
+          # never declared "provides" for (which means we would never have
+          # created DSL).  Anything where we don't create DSL is deprecated.
+          #
+          if !respond_to?(method_symbol)
+            Chef::Log.warn("#{resource_class} is marked as providing DSL #{method_symbol}, but provides #{method_symbol.inspect} was never called!")
+            Chef::Log.warn("In Chef 13, this will break: you must call provides to mark the names you provide, even if you also override provides? yourself.")
```
lamont-granquist: Chef::Log.deprecation

lib/chef/resource.rb
```diff
@@ -811,7 +812,10 @@ def updated=(true_or_false)
     #
     # @return [String] The DSL name of this resource.
     def self.dsl_name
-      convert_to_snake_case(name, 'Chef::Resource')
```
lamont-granquist: This should also most likely Chef::Log.deprecation in favor of `#resource_name`

lib/chef/dsl/resources.rb
```diff
@@ -0,0 +1,22 @@
...
+  module DSL
+    #
+    # Module containing a method for each globally declared Resource
+    #
+    # Depends on declare_resource(name, created_at, &block)
+    #
+    # @api private
+    module Resources
+      def self.add_resource_dsl(dsl_name)
+        module_eval <<-EOM, __FILE__, __LINE__+1
```
danielsdeleo: Can we use parens to be consistent with the rest of the code? This and driving are the only things I don't like "Seattle style" :stuck_out_tongue_winking_eye:

lib/chef/dsl/resources.rb
```diff
@@ -0,0 +1,22 @@
...
+          def #{dsl_name}(name, created_at=nil, &block)
+            declare_resource(#{dsl_name.inspect}, name, created_at || caller[0], &block)
+          end
+        EOM
+      end
+      def self.remove_resource_dsl(dsl_name)
+        remove_method dsl_name if method_defined?(dsl_name)
```
danielsdeleo: same

lib/chef/resource/lwrp_base.rb
```diff
@@ -35,32 +35,64 @@ class LWRPBase < Resource
...
 
-      # Set the resource name for this LWRP
+      def self.create_deprecated_class_in_chef_resource(resource_class)
+        # Create a class in Chef::Resource::MyResource with deprecation
+        # warnings if you try to instantiate or inherit it (for Chef 12 compatibility)
+        class_name = convert_to_class_name(resource_class.resource_name)
+        if Chef::Resource.const_defined?(class_name, false)
+          Chef::Log.debug "#{class_name} already exists!  Cannot create deprecation class #{resource_class}"
+        else
+          Chef::Resource.const_set(class_name, Class.new(resource_class) do
```
danielsdeleo: I'm not sure it will be, but I wonder if this would be easier to implement via `const_missing`

lib/chef/resource_resolver.rb
```diff
@@ -84,9 +69,48 @@ def maybe_dynamic_resource_resolution(resource)
...
+      class_name = convert_to_class_name(resource.to_s)
+      if Chef::Resource.const_defined?(class_name)
+        # If Chef::Resource::X already exists, and is *not* already marked as
+        # providing this resource, mark it as providing the resource and add it
+        # to the list of handlers for next time.
+        resource_class = Chef::Resource.const_get(class_name)
+        if resource_class <= Chef::Resource && !enabled_handlers.include?(resource_class)
+          enabled_handlers << resource_class
+          Chef::Log.warn("Class Chef::Resource::#{class_name} does not declare `provides #{resource.inspect}`.")
+          Chef::Log.warn("This will no longer work in Chef 13: you must use `provides` to provide DSL.")
```
danielsdeleo: Same `Chef::Log.deprecation` thing as elsewhere.

lib/chef/resource.rb
```diff
@@ -960,6 +964,11 @@ def method_missing(method_symbol, *args, &block)
       end
     end
 
+    def self.provides(name, *args, &block)
+      super
```
coderanger: Should be `super.tap { Chef::DSL::Resources.add_resource_dsl(name) }` or similar to avoid blocking the return value. Most of my resources use `.provides()` as an accessor too.

DOC_CHANGES.md
```diff
@@ -5,3 +5,36 @@ Example Doc Change:
...
+module MyModule
+  class MyResource < Chef::Resource
+    provides :my_resource
+  end
+end
+```
+
+Users are encouraged to declare resources in their own namespaces.
+
+This is deprecated (and will emit a warning) in Chef 12, and will be removed in
```
coderanger: I think this is phrased wrong. Explicit provides isn't being deprecated?

lib/chef/resource.rb
```diff
@@ -1090,25 +1099,23 @@ def should_skip?(action)
     # === Returns
     # <Chef::Resource>:: returns the proper Chef::Resource class
     def self.resource_for_node(short_name, node)
-      require 'chef/resource_resolver'
```
lamont-granquist: i think i did that deliberately so that you could require this file without slurping in the resource_resolver which slurps in all of the resources.

lib/chef/resource_resolver.rb
```diff
@@ -84,9 +69,48 @@ def maybe_dynamic_resource_resolution(resource)
       handlers[0]
     end
 
-    # try the old static lookup of resources by mangling name to resource klass
-    def maybe_chef_platform_lookup(resource)
-      Chef::Resource.resource_matching_short_name(resource)
```
lamont-granquist: So why not just throw a deprecation inside of `resource_matching_short_name()` instead of refactoring all of this code?  Then when the deprecation time is over, just remove this bit of code and remove resource_matching_short_name instead of trying to wire that up to the resource resolver.  I'm not sure what your refactoring is offering that wasn't here before and there might be edge conditions with the way that you're always adding the resource class now.

lib/chef/resource.rb
```diff
@@ -1090,25 +1099,23 @@ def should_skip?(action)
...
     # === Returns
     # <Chef::Resource>:: returns the proper Chef::Resource class
     def self.resource_matching_short_name(short_name)
-      begin
-        rname = convert_to_class_name(short_name.to_s)
-        Chef::Resource.const_get(rname)
-      rescue NameError
-        nil
-      end
+      Chef::ResourceResolver.new(Chef::Node.new, short_name).resolve
```
lamont-granquist: This should probably be left with the old code but add Chef::Log.deprecation.  I think that would also eliminate to do all the work you did in the resource_resolver as well.

lib/chef/resource/lwrp_base.rb
```diff
@@ -35,32 +35,64 @@ class LWRPBase < Resource
...
+          Chef::Log.debug("Overriding resources with LWRPs is not supported anymore starting with Chef 12.")
+          return existing_class
+        end
 
-          Chef::Resource.const_set(class_name, resource_class)
-          resource_class.resource_name = rname
-          resource_class.run_context = run_context
-          resource_class.class_from_file(filename)
+        # We load the class first to give it a chance to set its own name
+        resource_class = Class.new(self)
```
lamont-granquist: This will reintroduce the Chef 11 garbage collection / memory pressure problems back into ChefSpec which will make @sethvargo very sad.  The point of the old code was that the constant defined on resource was a trigger to let us know that we'd already loaded the LWRP and we didn't need to reload it.  We got rid of interval runs and got rid of the use case where we needed to reload LWRPs in the same chef process.

You're just backing out all that work we did here.

lib/chef/resource.rb
jkeiser: I would like to normalize *_name, but it didn't seem like the time :)  resource_name doesn't exist on Provider, at any rate--it's only on the class in LWRPBase.

lib/chef/resource.rb
jkeiser: Should we deprecate the entire method?  It doesn't seem useful to have a method whose sole purpose is to look up `Chef::Resource::FooBar`.

jkeiser: And yeah,I added a super useful comment about what it actually does, and then changed what it did to something that made a bit more sense.  We should probably just leave it and deprecate.

lib/chef/resource_resolver.rb
jkeiser: Sounds like an excellent plan to me.  I was probably thinking "it seems like all the resolution should actually happen in the resolver," but given that we're deprecating this thing anyway, may as well not care.

lib/chef/resource/lwrp_base.rb
jkeiser: So, I'm totally cool just backing this back out, but I added it on what I *think* was a conversation with you, which I clearly didn't understand as well as I thought at the time :)  You were saying you wished LWRPs could be kicked out of Chef::Resource, and I was OK with that, so I wrote this.  Is that accurate?  (Obviously you didn't also mean "and hey, could you make Chef take gobs and gobs of memory reloading LWRPs over and over again while you're at it," that was all me :)

If so, it seems like the right thing to do is to have this thing ask the ResourceResolver if the LWRP already exists instead of `Chef::Resource`, which should have a similar effect to what you are saying.  I'll implement that till your response, so I can have something.

I'll go dig through the source to understand what was triggering the issue, I'm curious.

jkeiser: Er.  I *think* the code I wrote there does exactly what @sethvargo's did before, except now it uses the existence of a resource named `cookbookname_resourcename` as a marker, instead of looking for `Chef::Resource::CookbooknameResourcename`.  If anything, it seems like we will load the LWRP even less than we did before, because it's doing a real resolve that will take into account similarly-named resources that live in other namespaces.  Am I missing something?

lib/chef/resource/lwrp_base.rb
jkeiser: What did you have in mind?  I'm not sure we'd want to make it *always* create a new Chef::Resource class in `const_missing`, and I'm not sure what else I could do in there.

DOC_CHANGES.md
jkeiser: Hah!  Nice catch.

lib/chef/resource.rb
jkeiser: I put it back in.  Not sure why I felt I needed it at the time.

lib/chef/resource/lwrp_base.rb
jkeiser: And yeah, you're right, and it did help :)

jkeiser:
All comments handled except @lamont-granquist's one about LWRP class creation (we can chat on Monday, I am likely missing something).

jkeiser:
I also added in the deprecation warning for Chef::Provider and made sure all providers and resources in Chef have an appropriate `provides`.

lib/chef/resource.rb
lamont-granquist: Thing is that dsl_name is pretty shitty.  The actual dsl name is in the `#delcared_type` on the insetance which is actually created.  I only see a couple of places where this gets used internally, and I doubt its getting much use externally.  This seems like a fine time to just make this go away.  If I'm wrong we can always back out, but keeping this method around doesn't make any sense to me.  It makes a bit more sense as a name for the semantic of `#declared_type` but its on the class, not the instance, so we can't ever make it work right.

lib/chef/resource/lwrp_base.rb
```diff
@@ -35,32 +35,42 @@ class LWRPBase < Resource
...
 
-          Chef::Resource.const_set(class_name, resource_class)
-          resource_class.resource_name = rname
-          resource_class.run_context = run_context
-          resource_class.class_from_file(filename)
+        if existing_class
+          Chef::Log.info("Skipping LWRP resource #{filename} from cookbook #{cookbook_name}.  #{existing_class} already defined.")
+          Chef::Log.debug("Overriding resources with LWRPs is not supported anymore starting with Chef 12.")
+          return existing_class
+        end
```
lamont-granquist: Okay, not sure what code I was reading Friday, but I missed this.  Might be why I don't like returns out of the middle of functions....

I think this might work, but it seems a little roundabout going through the ResourceResolver to look up the resource name.   I'm actually a bit surprised that we don't have a bit of inception since the ResourceResolver is depending on all the provides lines getting setup but we don't setup the default provides and slurp in the class_from_file until a few lines later....

lamont-granquist:
So by adding all the provides lines I'm a bit concerned that you're going to bypass the platform_map for some providers that still require that and they'll fail.  The platform_map really needs to be properly emptied out in order to get this done.

jkeiser:
@lamont-granquist what I actually did was avoid adding the provides lines for things in the provider mapping.  I *think* it can be done as a followon and doesn't need to be in the patch.

The big reason I didn't do it was I wanted to add individual `provides` to the actual classes, and to do that we would need to add an implicit priority to our mapping similar to the one the provider mapping uses: more specific providers are preferred over more general.  So, if A specifies a platform version, and B specifies a platform family, and both match, we prefer A.  os > platform_family > platform > platform_version, I would think.  But, it's also possible we could do it by adding explicitly to the priority map.  I'll tinker when I get a moment.

lib/chef/resource/lwrp_base.rb
jkeiser: Hmm, I can find a way to work out the middle-of-function return, and at the least make it clearer why we're skipping so early in the function.  The reason we don't set anything up before skipping is that we want to check if it's already there *before* we waste time setting it up (and chewing through memory).  The first time through, nobody has called `provides :x_y`, so we don't skip.

jkeiser: And, if there is a better way to find out whether we are redefining a resource than to use the ResourceResolver, I'm not sure what it is. If we just looked at LWRPs that happened to be defined in the past (which we don't have a list of anyway), we could still redefine resources created in other ways, and at least as written in the description, that is not intended behavior.

lamont-granquist: So, is this going to break the windows_package and windows_service cookbook resources which expect to be able to fire up and overwrite the built in classes?  And more broadly we support (and have to support) multiple classes handling the same resource in the dsl, and i think whichever one is parsed first will win based on this logic?

lamont-granquist: (and i don't really care about the return in the middle that much, i was trolling a completely different PR with that comment...)

jkeiser: `windows_package` and `windows_service` should be fine, because they will add `provides` and the `package` and `service` classes do not (they rely on the provider mapping).

The current logic is, if there are multiple `provides` the winner is the first in alphabetical order (unless there is an entry in the priority array).

Here's the summary of how it works right now (before this patch).  We try all these in order:

1. **Resource.provider:** If `resource.provider` is non-nil, we return that.
2. **Priority array:** If there is a priority array for this platform, we pick the first Provider that responds. Only the last-loaded priority array for this platform is used (so that cookbooks can override the system).
3. **Provides?:** If any descendant of `Provider` responds favorably to `provides?`, we return the first match *in alphabetical order by class name*.
4. **Platform mapping:** If the old platform mapping has an entry, we return that.  If there are multiple matches, we take the most specific (os first, then platform, then platform version).
5. **Chef::Provider::Blah:** If we find a `Chef::Provider` in `Chef::Provider::Blah`, we return that.

After the patch, behavior does not change except deprecation warnings.  However, after Chef X, `provides` will become required for `Chef::Resource` classes to participate in DSL.  When you add `provides` to a class, it is picked by step 3, and it now becomes possible (depending on alphabet) for the superclass to be picked where the subclass used to be picked, if you were relying on that implicit ordering.

The smallest suggestion I have to fix that is to modify step 3 so that any Resource or Provider with `provides :blah, <anything>` is preferred over `provides :blah` with no arguments or block.  I think there are further steps we can take to improve the selection overall, but I think we should probably take those steps separately.

jkeiser: Until we're ready to change the priority system, though, I think we should have documentation for people that are encountering this.  The blessed way to do it is to use a priority array:

```ruby
class Chef::Resource::Blarghle
  # Added provides!
  provides :blarghle
end
class Linux::A
  provides :blarghle, os: 'linux'
  # Since I was relying on the real blarghle's priority before, now I need to *specify* it:
  Chef.set_resource_priority_array :blarghle, [Linux::A, Chef::Resource::Blarghle]
end
```

jkeiser: I'm going to add a note about this in the RFC for explicitness.

coderanger: A thing to check is what happens if all the providers in the priority array returns false from `provides?`. Currently it just falls forward to the next phase I think. I use this in a few places for plugins where only the core ones have a priority order.

jkeiser: @coderanger yep, we do the same thing as before--it proceeds to the next phase.

jkeiser:
@lamont-granquist I have added the two things you requested, and a third besides:

- Deprecated class.dsl_name
- Made the LWRP reload protection a lot saner by setting and checking loaded_lwrps[filename] == true
- Added the same Chef::Provider::ResourceName deprecation for LWRP providers that we have for resources (with sane reload protection enabled)

There is one thing left that likely needs to be deprecated: `provider(:name_of_provider)` resolves `Chef::Provider::NameOfProvider`.  I left it off this because at least for LWRPs the deprecation is already there (since it grabs the constant) and I'm unsure about the breadth of its use.  Regardless, I don't see a way to replace it within this patch: you really need a node, an actual resource, and an action to look up a provider.  When we move to a central node map it will be a lot easier.

CHANGELOG.md
```diff
@@ -25,10 +25,13 @@
 * [pr#3295](https://github.com/chef/chef/pull/3295): Stop mutating `new_resource.checksum` in file providers.  Fixes some ChecksumMismatch exceptions like [issue#3168](https://github.com/chef/chef/issues/3168)
 * [pr#3320] Sanitize non-UTF8 characters in the node data before doing node.save().  Works around many UTF8 exception issues reported on node.save().
 * Implemented X-Ops-Server-API-Version with a API version of 0, as well as error handling when the Chef server does not support the API version that the client supports.
+<<<<<<< HEAD
```
lamont-granquist: stray merge conflict tag left in

DOC_CHANGES.md
```diff
@@ -5,3 +5,35 @@ Example Doc Change:
...
+module MyModule
+  class MyResource < Chef::Resource
+    provides :my_resource
+  end
+end
+```
+
+Users are encouraged to declare resources in their own namespaces instead of putting them in the special `Chef::Resource` namespace.
+
+The `Chef::Resource` namespace will no longer get automatic DSL in a future Chef, and will emit a deprecation warning in Chef 12.
```
lamont-granquist: This could probably be re-worded a little better:

Authors of HWRPs need to be aware that in the future all resources and providers will be required to include a provides line.   Starting with Chef 12.4.0 any HWRPs that do not have provides lines will trigger deprecation warning messages to add them.  The LWRPBase code does this automatically so LWRP authors and users who write classes that inherit from LWRPBase do not need to explicitly include provides lines.

Or something like that.  Spell out who needs to pay attention to what you're doing by what they're doing (writing LWRPs/HWRPs/LHWRPs) and not by what you're doing (removing Chef::Resource magic)...

DOC_CHANGES.md
```diff
@@ -5,3 +5,35 @@ Example Doc Change:
...
+`mycookbook/resources/myresource.rb`, you will no longer be able to extend or
+reference `Chef::Resource::MycookbookMyresource` in Ruby code.  LWRP recipe DSL
+does not change: the LWRP will still be available to recipes as
+`mycookbook_myresource`.
+
+You can still get the LWRP class by calling `Chef::Resource.resource_matching_short_name(:mycookbook_myresource)`.
+
+The primary aim here is clearing out the `Chef::Resource` namespace.
+
+References to these classes is deprecated (and will emit a warning) in Chef 12, and will be removed in Chef 13.
```
lamont-granquist: Again this should probably be reworded, since its sounds like a breaking change.  Should lead with the fact that the auto-Chef::Resource namespace is being deprecated.

`Chef::Resource.resource_matching_short_name` is also hella wordy.  Can we get something a little tighter like `Chef::Resource.class_for(:package)`?

lib/chef/node_map.rb
```diff
@@ -78,6 +78,12 @@ def get(node, key)
       nil
     end
 
+    def clear
```
lamont-granquist: I usually think of this as `#reset!` when acting on singleton-ish objects like this, but `#clear` makes more sense from the hash-like interface perspective.  Could we maybe alias `:reset!` to this?

lib/chef/platform/provider_mapping.rb
```diff
@@ -292,6 +292,8 @@ def platforms
               :user => Chef::Provider::User::Useradd,
               :group => Chef::Provider::Group::Gpasswd,
               :ifconfig => Chef::Provider::Ifconfig,
+              :package => Chef::Provider::Package,
+              :service => Chef::Provider::Service,
```
lamont-granquist: Those are base classes that should never be instantiated alone.

lib/chef/provider_resolver.rb
```diff
@@ -20,6 +20,93 @@
...
+  # which wins over the default.
+  #
+  # ### Chef::Provider::FooBar
+  #
+  # As a last resort, if there are *still* no classes, the system transforms the
+  # DSL name `foo_bar` into `Chef::Provider::FooBar`, and returns the class if
+  # it is there and descends from `Chef::Provider`.
+  #
+  # NOTE: this behavior is now deprecated.
+  #
```
lamont-granquist: :+1:

lib/chef/resource.rb
lamont-granquist: I sorta suspect this'll cause a dependency loop somewhere for someone like chef-sugar or chefspec.

spec/data/lwrp/providers/buck_passer.rb
```diff
@@ -1,12 +1,28 @@
...
+  ensure
+    Chef::Config[:treat_deprecation_warnings_as_errors] = old_treat_deprecation_warnings_as_errors
+  end
+end
+
 action :pass_buck do
   lwrp_foo :prepared_thumbs do
     action :prepare_thumbs
-    provider :lwrp_thumb_twiddler
+    # We know there will be a deprecation error here; head it off
```
lamont-granquist: why do we get deprecation warnings?

lamont-granquist: nm, i sorta see what you're doing, although checking for the deprecation warnings like this concerns me a little bit just due to once this is actually deprecated we need to track down ripping it out...

DOC_CHANGES.md
jkeiser: This was the existing method :)  I'd like to fix the interface for resource resolution as well, but I think we need to clear a few more things out of the way before we can clearly see the right interface.

lib/chef/node_map.rb
jkeiser: Actually, it looks like I didn't need it (searching for "clear" in the patch didn't turn anything up.  Cruft from an earlier revision perhaps.

spec/data/lwrp/providers/buck_passer.rb
jkeiser: Yeah, that's going to be a thing--however, we should be able to search for anywhere we turn deprecation warnings off.

jkeiser:
@lamont-granquist all changes made except the resource_matching_short_name change--not sure if we should add more public API until further thought.

lamont-granquist:
The resource_matching_short_name thing is in your DOC_CHANGES though so that means you're declaring a public API and you're also kind of demanding that users start to type that out.  Time would be now to figure out something a little bit better.

jkeiser:
What I'd like is this:

```ruby
Chef.resource_types.get(:file, node: run_context.node)
Chef.resource_types.add(:file, Chef::Resource::File, os: 'linux')
Chef.provider_types.get(:file, node: run_context.node)
Chef.provider_types.add(:file, Chef::Resource::File, os: 'linux')
```

jkeiser:
Well, it works for resources anyway. What I really want is for there not to be a global registry of provider types at all.  Resources and Providers are tightly coupled objects (there may be many providers for one resource, but in general a provider won't work against multiple resources--the fact of the amtter is different resources have different interfaces).  It doesn't make sense to register them with names.

jkeiser:
I could see `Chef::ResourceResolver.resolve(name)` as roughly the most minimal interface change.

jkeiser:
@lamont-granquist implemented.

lamont-granquist:
:shipit:

PR #3270, opened 2015-04-23
Pull Config from external Gem

danielsdeleo:
## Extract Chef's Config to a Gem ##

(edited this description since the old description doesn't apply).

In https://github.com/chef/chef-rfc/pull/118 we want to be able to configure ohai. This would be best if we can just re-use Chef's config, so you don't need a `client.rb` and an `ohai.rb`, we can just add an ohai config context to Chef's config and use the same config file. Also, I believe that other projects would benefit from being able to read Chef's config without having to depend on all of Chef.

To that end, this PR creates a `ChefConfig` gem which is totally self-contained and depends only on mixlib-shellout (for locale detection) and mixlib-config. It is located in the main chef git repo so you don't have to deal with a bunch of Gemfile editing just to make a simple fix or add a new default config setting; you just make your change.

Currently both Chef's tests and ChefConfig's tests pass, so we _could_ just merge this. However, we probably should put a bit of thought into how much we want to synchronize the version numbers, etc. I'll detail those questions in a comment below.

@lamont-granquist @chef/client-core


danielsdeleo:
To summarize some discussion from private chat:

We would like to rev the version of chef-config either 100% in lockstep with Chef, or keep major and minor versions in lockstep. If we leave chef-config in a separate repo, that's going to cause development/release hassles where we have to do a bunch of "digital paperwork" just to synchronize versions. Bundler supports gems that are nested within a git repo as long as they're not nested too deeply (one or two dirs max). So we can make everything a lot more convenient if we put chef-config in this (Chef) repo. One way to do that is to go back to the layout we had in the chef-10 times, where chef would be in `chef/chef` and chef-config would be in `chef/chef-config`. I think it should be possible to alternatively just add the subprojects as directories without altering the structure of the repo otherwise, but I would want to check to make sure bundler behaves well. Thoughts on this issue would be welcome.

danielsdeleo:
Also, I created some issues for the ways in which `chef-config` currently depends on code from Chef: https://github.com/chef/chef-config/issues

* `Chef::FileCache`: this seems to be dead OpenID code.
* `Chef::Platform.windows?`: May need to suck it up and duplicate the windows check.
* `Chef::PathHelper`: I haven't looked at this not sure of the best way to proceed.
* `Chef::Mixin::ShellOut`: only used once, for locale detection. May be able to just use stdlib command execution methods (backticks/Process.spawn) or have Chef "monkey patch" locale config into the config class.

danielsdeleo:
@chef/client-core I believe the basic code here is correct or at least ready for final review. `chef-config` has been imported via subtree merge so its history is preserved (though this _does_ break history from `lib/chef/config.rb`--if that's a showstopper, that could be reworked).

We should probably decide on the process/metadata stuff:

* do we want to release both chef and chef-config together at the same version number all the time?
* Along with the above, what should the version constraint on the chef-config dependency be? FWIW, RSpec does something like `~> MAJOR.MINOR.0` for normal releases and exact equality for prereleases.
* What automation do we need in place to ensure everything ships when it's supposed to? Just update `rake release` to also ship `chef-config`?
* We need to make sure travis and appveyor run chef-config's tests. I think the one downside of the single-repo approach is it's difficult to determine if you only need to run the tests from one project or another. OTOH, the chef-config tests run in ~ 1s, so probably not a big deal for now.

lamont-granquist:
Easiest thing to do would probably be to make them always match exactly in version number and always equality pin.

danielsdeleo:
@lamont-granquist that would be my preference, at least to start. Off the top of my head, the way I think we do that is have one VERSION_NUMBER file in the root of the repo and have that code-gen the version.rb files (including a comment describing how to modify the version number w/ rake). That way we're less likely to forget to bump one but not the other. Reasonable?

lamont-granquist:
Yep, you are reading my mind...

danielsdeleo:
Fingers crossed, I think my latest commit should move the appveyor issues. Travis will still fail because appbundler and appbundle-updater are not going to work out of the box with this change. We will need some sort of build step that builds the component gems (just chef-config) for now. Any preferences on how that should work? @chef/test-driven-infrastructure ?

danielsdeleo:
Okay, I updated the rake tasks. One annoying bit is I think you'll have to run `rake version` and then separately you can run `rake package` because we load the gem specs when the Rakefile is evaluated. We could maybe switch to just running `gem build` to fix that if we want.

danielsdeleo:
Appveyor issue is specs stubbing `Chef::Platform.windows?` but that not changing `Chef::Config`'s behavior because it's looking at `ChefConfig.windows?`. Annoying but easily fixable.

danielsdeleo:
Specs are passing everywhere, but appbundle updater will need to be updated to install the chef-config gem before running the kitchen tests. cf. https://github.com/jdmundrawala/chef-appbundle-updater/blob/master/providers/appbundle.rb

jdmundrawala:
Yep, we're going to have to add a step that checks if the chef-config directory is there, and gem installs if it finds it.

tyler-ball:
@danielsdeleo I think the Rake tasks for managing chef-config make sense.

:+1:

danielsdeleo:
Okay, now travis is completely happy. @chef/client-core any review thoughts? Last item to complete is to update omnibus-software to install components as gems to make appbundler happy.

lamont-granquist:
:shipit:

PR #3271, opened 2015-04-23
Implement `knife user key create` and `knife client key create`

tylercloke:
This branch changes the following (see relevant commit message for details on each):
+ Fixed bug in Chef::Key related to create.
+ Set sane default for chef_server_root.
+ Implements `knife user key create` and `knife client key create`.

Ping @chef/lob


danielsdeleo:
@chef/client-core

lib/chef/knife/client_key_create.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      def initialize(argv=[])
+        super(argv)
+        @service_object = nil
+      end
+
+      def run
+        apply_params!(@name_args)
+        service_object.run
+        0
```
jaymzh: eh?

lib/chef/knife/key_create.rb
```diff
@@ -0,0 +1,128 @@
...
+      def edit_data(key)
+        @ui.edit_data(key)
+      end
+
+      def display_info(input)
+        @ui.info(input)
+      end
+
+      def display_private_key(private_key)
+        @ui.msg private_key
```
jaymzh: nit: this is the only time you're not putting paryns around passed arguments...

lib/chef/knife/key_create.rb
```diff
@@ -0,0 +1,128 @@
...
+      def create_key_from_hash(output)
+        Chef::Key.from_hash(output).create
+      end
+
+      def run
+        key = Chef::Key.new(@actor, @actor_field_name)
+        if !@config[:public_key] && !@config[:key_name]
+          raise Chef::Exceptions::KeyCommandInputError, public_key_or_key_name_error_msg
+        elsif !@config[:public_key]
+          key.create_key true
```
jaymzh: or at least it was...

spec/unit/key_spec.rb
```diff
@@ -496,17 +518,17 @@
...
+          let(:key)        { user_key }
         end
       end
 
-      context "when creating a client key" do
+      context "when updating a client key" do
         it_should_behave_like "update key" do
-          let(:url) { "clients/#{client_key.actor}/keys/#{key.name}" }
-          let(:key) { client_key }
+          let(:url)        { "clients/#{client_key.actor}/keys/#{key.name}" }
```
jaymzh: why the sudden tabs? nothing else in this file (afaict) does this.

jaymzh:
My stuff is style nits - other than the `0` which perhaps is just some part of the knife plugin infrastructure I don't know about. This looks pretty solid to me. I looked largely because I'm interested in this feature, but someone with more knife knowledge should comment on your redundancy question and general knife-y-ness. But in general I'm :+1:

lib/chef/knife/key_create.rb
tylercloke: :trollface: I'll update them both.

lib/chef/knife/client_key_create.rb
tylercloke: Ah, missing an `exit` I think. Probably don't even need it. I think knife will do the right thing and exit zero if it gets no errors.

tylercloke:
Thanks for the review!

lib/chef/knife/client_key_create.rb
jaymzh: Ah. I see. Yeah, that shouldn't be needed knife should do the right thing. Also, I'm guessing it's probably looking for truthy values in determining it's exit status, so while 0 certainly is truthy, it's seems misleading.

lib/chef/knife/key_create.rb
jaymzh: If you do, just FYI, I stopped commenting, because the rest of the file went back and forth, so there's more.

lib/chef/knife/user_key_create.rb
```diff
@@ -0,0 +1,90 @@
...
+      def actor_missing_error
+        'You must specify a user name'
+      end
+
+      def apply_params!(params)
+        @actor = params[0]
+        if @actor.nil?
+          show_usage
+          ui.fatal(actor_missing_error)
+          return 1
```
stevendanna: I think you probably want `exit 1` here so that the processes exits with one.  I think that the current code will continue onto execute service_object.run even in the case of actor.nil?

lib/chef/knife/client_key_create.rb
```diff
@@ -0,0 +1,90 @@
...
+      def actor_missing_error
+        'You must specify a client name'
+      end
+
+      def apply_params!(params)
+        @actor = params[0]
+        if @actor.nil?
+          show_usage
+          ui.fatal(actor_missing_error)
+          return 1
```
stevendanna: Same comment re exit 1 in the other location.

lib/chef/knife/user_key_create.rb
tylercloke: I tested the difference between `exit 1` and `return 1` there. Knife exits `100` on failure no matter which I do.

stevendanna:
Overall this is good.  I really like getting chef_server_root into our core config.

The amount of duplication between the two knife commands makes me wonder if your key_create class should actually just be a class that we include in each plugin.  You could then use the `included` method to manage ensure the joint options got set:

https://github.com/chef/knife-ec-backup/blob/master/lib/chef/knife/ec_base.rb#L26-L80

However, a little duplication in knife commands isn't the end of the world, so the only only comment of mine that I see as blocking is the return 1 vs exit 1 issue.

tylercloke:
Travis refuses to run...*sigh*

danielsdeleo:
@stevendanna The reasoning behind the way the code is structured is to have a clear separation of code that deals with handling CLI options/args and code that handles an action. In ChefDK I've found that it makes it much easier to build a command that combines a few operations by composing the "service" objects. It also makes testing a lot easier because you only need to verify that the CLI-level code talks to the service code correctly, while it's easy to use DI to pass dependencies (like HTTP API client stuff) into the service object.

Finally, I think this sort of design will make it more attractive for 3rd party authors to write scripts/libraries in a way we feel comfortable supporting. I find it pretty terrible that people use knife CLI code as library code, because knife operates at the CLI layer where it's acceptable to print directly to the output streams, exit if input is invalid, etc. That kind of behavior is generally unacceptable for a library. Having a separate service object fixes that issue by creating a clear line between the two kinds of code.

lib/chef/config.rb
```diff
@@ -319,7 +319,17 @@ def self.path_accessible?(path)
       default :host, 'localhost'
       default :port, 8889.upto(9999) # Will try ports from 8889-9999 until one works
     end
+
     default :chef_server_url,   "https://localhost:443"
+    default(:chef_server_root) do
+      # if the chef_server_url is a path to an organization, aka
+      # 'some_url.../organizations/*' then remove the '/organization/*' by default
+      if self.configuration[:chef_server_url] =~ /\/organizations\/\S*$/
```
danielsdeleo: It would be nice to have a test for this. I tried it manually with both `https://chef.example/extra/root/organizations/myorg` and `https://chef.example/extra/root/organizations/myorg/` (without/with trailing slash) and it seemed to work okay, but if there happens to be a bug there it'd be nice to already have a place to add a regression test.

danielsdeleo:
:+1: test for config stuff would be nice to have and it looks like most of the infrastructure is in place in `spec/unit/config_spec.rb`

lib/chef/config.rb
tylercloke: Sure, I can add a test.

stevendanna:
@danielsdeleo Thanks for the summary there.  That makes total sense.

Issue #3272, opened 2015-04-23
knife subcommands do not work anymore on Windows (since Chef 12.3.0.rc.0)


tknerr:
Im trying to run `knife cookbook` or `knife cookbook --help` to get more info about the cookbook subcommands, but it fails like this:
```
C:\Repos\_github\bills-kitchen\target\build\repo\vagrant-workflow-tests\playground\sample-toplevel-cookbook>bundle exec knife cookbook --help
FATAL: Cannot find sub command for: 'cookbook --help'
Available subcommands: (for details, knife SUB-COMMAND --help)

** KNIFE COMMANDS **
Usage: C:/Repos/_github/bills-kitchen/target/build/home/.chefdk/gem/ruby/2.1.0/bin/knife (options)

** PATH-BASED COMMANDS **
knife xargs [COMMAND]
```

My `Gemfile` looks like this:
```
source 'https://rubygems.org'
gem 'chef', '12.3.0.rc.0'
```

and it's using the right version, too:
```
C:\Repos\_github\bills-kitchen\target\build\repo\vagrant-workflow-tests\playground\sample-toplevel-cookbook>bundle exec knife -v
Chef: 12.3.0.rc.0
```

This is happens on my Windows machine only, on my Ubuntu it works like a charm.

Can anyone with a Windows at hand reproduce this?

tknerr:
P.S. I am using bundler here because of the ChefDK knife being in the front of my PATH otherwise

Calling the `knife` executable directly leads to a different (but also bad) result: it lists only the `knife cookbook metadata`, but no other cookbook commands?!

So where is it?
```
C:\Repos\_github\bills-kitchen\target\build\repo\vagrant-workflow-tests\playground\sample-toplevel-cookbook>bundle exec which knife
C:/Repos/_github/bills-kitchen/target/build/home/.chefdk/gem/ruby/2.1.0/bin\knife
```

Ok, let's try this one:
```
C:\Repos\_github\bills-kitchen\target\build\repo\vagrant-workflow-tests\playground\sample-toplevel-cookbook>C:/Repos/_github/bills-kitchen/target/build/home/.chefdk/gem/ruby/2.1.0/bin\knife cookbook --help
FATAL: Cannot find sub command for: 'cookbook --help'
Available cookbook subcommands: (for details, knife SUB-COMMAND --help)

** COOKBOOK COMMANDS **
knife cookbook metadata COOKBOOK (options)

```

tknerr:
[...]

PR #3277, opened 2015-04-24
Fix knife subcommands on windows

tknerr:
This fixes https://github.com/chef/chef/issues/3272

The underlying issue was that "-x86-mingw32"-platform suffixed chef gem (as it is installed on windows) was filtered out by the regex, thus it was detected as being "not from this chef version ".

The PR adapts the regex and adds some valid / invalid specs for it

### DIFF
--- a/lib/chef/knife/core/subcommand_loader.rb
+++ b/lib/chef/knife/core/subcommand_loader.rb
@@ -23,7 +23,7 @@ class Chef
     class SubcommandLoader
 
       MATCHES_CHEF_GEM = %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
-      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-x86-mingw32)?/}.freeze
+      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-\w+)?(-\w+)?/}.freeze
 
       attr_reader :chef_config_dir
       attr_reader :env
--- a/lib/chef/knife/core/subcommand_loader.rb
+++ b/lib/chef/knife/core/subcommand_loader.rb
@@ -23,7 +23,7 @@ class Chef
     class SubcommandLoader
 
       MATCHES_CHEF_GEM = %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
-      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}/}.freeze
+      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-x86-mingw32)?/}.freeze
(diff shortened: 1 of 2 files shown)


lib/chef/knife/core/subcommand_loader.rb
```diff
@@ -23,7 +23,7 @@ class Knife
     class SubcommandLoader
 
       MATCHES_CHEF_GEM = %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
-      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}/}.freeze
+      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-x86-mingw32)?/}.freeze
```
jdmundrawala: Can we make this match any platform specific gem so it does not break again (say if we have a 64 bit build)

tknerr: Sure. I looked into https://rubygems.org/gems/chef/versions and found only "-x86-mingw32" gems, but that might change in the future.

I think it would be safe to relax the pattern a bit so we will catch 64bit builds as well once we have them.

tknerr: @jdmundrawala on a second thought: not sure what to actually match here, i.e. no clue how other valid platforms the chef gem will be built for in the future would look like.

Is it "x86-mingw64" or "i386-mingw64" or just "mingw64" etc?

Maybe it would be better to have an integration test which ensures that `knife cookbook --help` returns the expected subcommands instead?

jdmundrawala: I think if you remove some of your invalid tests, and allow it to match to anything like `(-\s*)?(-\s*)?` it should work.
Not sure of the regex is correct, but the idea is optional cpu, optional os

tknerr: Ah I see, yep that's a good approach. So "chef-12.3.0-foo-bar" would actually be invalid, but we don't care because it won't happen realistically. So it's really only about preventing "12.3.0" matching "12.3.0.rc.0" or "12.3.0.dev" etc (i.e. the realisitc options), right?

I'll update the PR later tonight...

jdmundrawala: I think that sounds right.
Thanks for tracking this down.

tknerr:
@jdmundrawala relaxed the regex a bit and added some more specs. 

Btw: TravisCI does not run on this branch here, does it?

jdmundrawala:
@tknerr can you open the pull request against master. We will cherry pick onto the stable branches if needed.

tknerr:
Created #3281 with the commits rebased / cherry-picked against master. TravisCI is happy too ;-)

PR #3278, opened 2015-04-24
Switch over Windows builds to universal builds.

ksubrama:
Updated dependencies and refactored platform gemspecs.


ksubrama:
@btm @jdmundrawala 

Can one of you also tag chef-client-core because I'm not on that team yet.

jdmundrawala:
cc @chef/client-core
Also @tknerr

LGTM... thoughts on merging this even without Chef working with 64 bit ruby on windows?

danielsdeleo:
I'm cool with it, people can do what they want outside of omnibus but we don't support it, so no need to worry about someone stubbing their toe on 64bit ruby for now. :+1:

chef-windows.gemspec
```diff
@@ -0,0 +1,22 @@
+gemspec = eval(IO.read(File.expand_path("../chef.gemspec", __FILE__)))
+
+gemspec.platform = Gem::Platform.new ["universal", "mingw32", nil]
```
btm: style?

```
 Gem::Platform.new(["universal", "mingw32"])
```

jdmundrawala: Will that works:
https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb#L54

ksubrama: Hmm - weirdly enough, it does seem to work.  I just assigns nil to the last item.  Ruby is whacky...

chef-windows.gemspec
```diff
@@ -0,0 +1,22 @@
+gemspec = eval(IO.read(File.expand_path("../chef.gemspec", __FILE__)))
+
+gemspec.platform = Gem::Platform.new ["universal", "mingw32"]
```
btm: Mostly my style comment was around putting parentheses in there, more than dropping the nil. I just saw you could in testing/reading the docs.

You're passing an array as a single argument and my brain has to stop to understand that without the parentheses.

```
gemspec.platform = Gem::Platform.new(["universal", "mingw32"])
```

chef-windows.gemspec
ksubrama: Oh yay!  I like the extra parens to make function calls explicit but that supposedly offends the ruby sensibility of some others.  I'm quite happy to put it back in.

tknerr:
What's the implication of this change? Sorry, didn't get it...

jdmundrawala:
@tknerr It would have affected your knife PR (gems will no longer be x86-mingw32)

tknerr:
ah I see, thanks @jdmundrawala

tyler-ball:
This LGTM but

btm:
@tyler-ball left shark is the laissez-faire shark.

btm:
@ksubrama so I wonder if the Travis test-kitchen ec2 failure is because we try to do this:

```
sudo -E rm -f pkg/chef-*-x86-mingw32.gem
sudo -E $chef_bin/gem install --local pkg/chef-*.gem             -n /opt/chef/bin --no-rdoc --no-ri
```

And now the filename of mingw32 gem is different?

@mcquin helped me track down the code that's doing that, it's way off in here still: https://github.com/mcquin/test-kitchen/blob/mcquin/install_chef_from_github/lib/kitchen/provisioner/chef_base.rb#L276

@jdmundrawala is https://github.com/jdmundrawala/kitchen-appbundle-updater ready to replace the test-kitchen branch here?

btm:
(we should make sure Travis is green before merging)

jdmundrawala:
@btm https://github.com/chef/chef/pull/3260

Rakefile
```diff
@@ -40,7 +40,7 @@ task :uninstall do
 end
 
 desc "Build it, tag it and ship it"
-task :ship => :gem do
+task :ship => [:clobber_package, :gem] do
```
jdmundrawala: Don't think you need to clobber package

ksubrama: The :gem task doesn't seem to update/rebuild the gem if one is already present and there are changes.  I found my self constantly rake repackage or rake clobber_package and then rake gem when testing things out.  Is this a fairly common gem/shipping pattern?  I don't actually know if the repackage task is different from the gem task since they both seem to be doing the same thing otherwise (except for the extra cleanup in the beginning).

ksubrama:
And travis now passes!

btm:
What bumped the versions in chef-windows.gemspec compared to the remove gemspec? I see a lot of bumps in there and at least one that increased the minor version on a pessimistic constraint.

ksubrama:
I manually did.  I went to see the changelog for each repo and most of them included some helpful bugfix or the other - mostly involving 64-bit support (8-byte vs 4-byte values in system calls etc.).  Nothing breaking and radical.  The tests all passed (well - when run in 32-bit ruby...  I think there was 1-2 issues with 64-bit itself), so I decided to bump the versions up.

Rakefile
danielsdeleo: That's my experience as well, and it only takes a little while to rebuild

tknerr:
So this PR would fix #3272 too (and thus make #3281 obsolete), right?

ksubrama:
@tknerr I think you'd still need those.  The gem generated would be chef-xxx-universal-mingw32 but subsequent dependencies can still be x86-mingw32 or x64-mingw32.  The same goes for plugins I suppose.  I haven't looked at #3281 too closely, but I'd think that you'd need to still soft match the plugins you want to load based on the current chef version as you are doing, and then filtering out the platforms in some creative way (I haven't looked at that code - maybe it already does that).

ksubrama:
I just read through that subcommander loader and I think I now understand what it's trying to do.  Ignore my previous comment.  Yes you'd still need #3281 as is as you'd still need to soft-match the gem name  chef-xxx-universal-mingw32 on windows.

PR #3281, opened 2015-04-24
Fix knife on windows (rebased against master)

### LINKED ISSUE #3272: knife subcommands do not work anymore on Windows (since Chef 12.3.0.rc.0)

tknerr:
This fixes #3272 (knife subcommands don't work on windows)

See https://github.com/chef/chef/pull/3277 for the original PR against 12.3-stable

### DIFF
--- a/lib/chef/knife/core/subcommand_loader.rb
+++ b/lib/chef/knife/core/subcommand_loader.rb
@@ -23,7 +23,7 @@ class Chef
     class SubcommandLoader
 
       MATCHES_CHEF_GEM = %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
-      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-x86-mingw32)?/}.freeze
+      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-\w+)?(-\w+)?/}.freeze
 
       attr_reader :chef_config_dir
       attr_reader :env
--- a/lib/chef/knife/core/subcommand_loader.rb
+++ b/lib/chef/knife/core/subcommand_loader.rb
@@ -23,7 +23,7 @@ class Chef
     class SubcommandLoader
 
       MATCHES_CHEF_GEM = %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
-      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}/}.freeze
+      MATCHES_THIS_CHEF_GEM = %r{/chef-#{Chef::VERSION}(-x86-mingw32)?/}.freeze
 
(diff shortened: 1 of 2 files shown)


jdmundrawala:
:+1:

tyler-ball:
:+1:

tyler-ball:
Can you update these examples to include 'chef-xxx-universal-mingw32' like @ksubrama mentions in #3278?

tknerr:
@tyler-ball @ksubrama yep, I will add another test case for the "universal" gems.

I'm also considering to relax the regex a bit, to support more than two dashes. I have no list of available / existing `RbConfig['RUBY_PLATFORM']` values available, but I [suspect](http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/17139) they might contain more than two `-` inbetween

tknerr:
@tyler-ball @jdmundrawala @ksubrama what do you think about relaxing the regex?

From: `%r{/chef-#{Chef::VERSION}(-\w+)?(-\w+)?/}.freeze`
To: `%r{/chef-#{Chef::VERSION}(-\w+)*/}.freeze`

ksubrama:
That seems reasonable to me.  From what I know, the Gem::Platform class (that's used in gemspec) has 3 fields: cpu, os and version which may or may not be present and if they are present, they are '-' separated.  You could just add another (-\w+)? but *shrug* I'm also not really worried about accidentally over-matching another gem that just magically happens to start with the name "chef-<version-string>".  Is this some weird security scenario we'd need to worry about?  Somebody going out and making gems like chef-12.2.1-1-lololpwnd-whatever ?

ksubrama:
Merged as e19aacce268f37e147228a379396b3adb4f648ce

PR #3080, opened 2015-03-15
Issue 2247: powershell_script exit status should be nonzero for syntax errors

adamedx:
See #2247 @jdmundrawala, @smurawski, @btm, @carpnick.

If a script processed by the `powershell_script` resource has a syntax error, we are returning 0, i.e. success, and the chef-client run will continue with no error even though the script did not actually execute (#2247).

To address this issue, we're going to add an explicit "syntax verification-only" similar to what was proposed in #2247.

### Root cause description

The root cause of this entire issue centers around the desire to preserve the quite useful conventions around script exit codes honored by the Unix script interpreters and their associated resources, and even honored by cmd.exe: the interpreter exits with a non-zero error code if the last command (internal or external) executed by the script returned a nonzero exit code OR if the script itself is invalid. The PowerShell interpreter violates this in multiple ways:

1.  `powershell.exe -command` correctly returns a non-zero exit status (`1`) if the script is not well-formed, but does not propagate non-zero process exit codes from Win32 processes -- those exit codes are always returned as `1`, even if a script is terminated explicitly via PowerShell's `exit` statement with a specific exit code. There is a pretty large dependency in existing Windows (and non-Windows) cookbooks on exit codes being propagated with full fidelity specifically because this is a very useful convention and there is a significant body administration practice built on it over decades.
2.  `powershell.exe -file` correctly handles Win32 exit statuses -- if the `exit` statement is used in the script to set the exit status. Also, it does not propagate cmdlet exit status (i.e the value of `$?`) nor does it return a non-zero exit for non well-formed scripts. In both cases it will return 0.
3.  `powershell.exe` in either the `-command` or `-file` cases does not support the idea of returning a status code based on a boolean test result as can be done with Unix shells like `sh` with `test` and related commands -- this is the feature enabled by the `convert_boolean_return` attribute for the `powershell_script` resource.

All of these issues requirements around returning useful status codes according to conventions supported by other interpreters / Chef script resources require explicit code in the `powershell_script` resource to implement these desired behaviors. Workarounds for all these issues are implemented there, except for the syntax issue, which is addressed by this pull request.

### Implementation notes

The following options were considered as solutions:

1. Continue to launch powershell.exe with `-file` and find a way to evaluate the syntax without executing the script more than once.
 <p>a. The script can be placed inside a script block and then evaluated in a separate `powershell.exe` invocation. **This is the solution implemented in this pull request.**</p>
[...]

PR #3283, opened 2015-04-25
Adamedx/x ps exitstatus rebase

adamedx:


spec/functional/resource/powershell_spec.rb
```diff
@@ -56,14 +56,13 @@
       resource.run_action(:run)
     end
 
-    it "returns the -27 for a powershell script that exits with -27", :windows_powershell_dsc_only do
-      # This is broken on Powershell < 4.0
+    it "returns the exit status 27 for a powershell script that exits with 27" do
       file = Tempfile.new(['foo', '.ps1'])
       begin
-        file.write "exit -27"
+        file.write "exit 27"
```
jdmundrawala: nice

lib/chef/provider/powershell_script.rb
```diff
@@ -24,71 +24,147 @@ class PowershellScript < Chef::Provider::WindowsScript
 
       provides :powershell_script, os: "windows"
 
+      def initialize (new_resource, run_context)
+        super(new_resource, run_context, '.ps1')
+        normalize_script_exit_status
+      end
+
+      def action_run
+        valid_syntax = validate_script_syntax!
```
jdmundrawala: couldn't this be `super if validate_script_syntax!`?

jdmundrawala:
:cake:

lib/chef/provider/powershell_script.rb
adamedx: Now I'm trying to understand why I even need to care about the return value. That's what happens when you leave a PR for a month or two :frowning: .

Seems like I should just let it raise an exception or keep going.

adamedx:
going to close in favor of #3080.

PR #3284, opened 2015-04-25
avoid platform matched /mswin|mingw|windows/ for Chef::Log::Syslog

sawanoboly:
Don't load `log/syslog` when platform is windows.


spec/unit/log_spec.rb
```diff
@@ -23,21 +23,23 @@
 describe Chef::Log do
 end
 
-describe Chef::Log::Syslog do
-  let(:logger) { Chef::Log::Syslog.new }
+unless Chef::Platform.windows?
```
jdmundrawala: I think this needs to be rethought as this is a unit test

sawanoboly: Hmm, shall I move this to functional test ? @jdmundrawala

sawanoboly:
@jdmundrawala move syslog specs from unit to functional and then update for call Chef::Log methods.

spec/unit/log_spec.rb
lamont-granquist: In this case you can't use the syslog-logger gem on windows at all which makes it difficult to write a unit test that works on windows (not without getting crazy and stubbing require to return true and other horrible things).

lamont-granquist:
@sawanoboly i'd say it was fine before as a unit test.

sawanoboly:
@lamont-granquist So I'll just move spec/functional/log/syslog_spec.rb
 to spec/unit/log/syslog_spec.rb of 169e4de.

Would that be OK?

lamont-granquist:
yep that works.

sawanoboly:
Whoops! I have merged master with habit. I'll create new PR.

sawanoboly:
@lamont-granquist Sorry, please refer #3322 . I'll close it.

PR #3289, opened 2015-04-27
Git provider should not force branch on master

### LINKED ISSUE #3025: Git provider fails to force update when you are already on the branch
Consider the following invocation of the git provider:
```ruby                                            
 git '/opt/fenestra' do
  action :sync
  repository 'https://github.com/osuosl/fenestra'
  checkout_branch 'master'
end 
```
By default, git will check out the master branch. The provider will then attempt to force update to the `checkout_branch`, but will fail because you can't force checkout a branch you are on.
https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L160
An example on the command line:
```shell
$ git checkout master
$ git branch -f master
fatal: Cannot force update the current branch.
```
I'm interested in fixing this issue. What would be the best way to go about it?
Would I want to change the provider to look something vaguely like the following?

```ruby
def checkout
  sha_ref = target_revision
  converge_by("checkout ref #{sha_ref} branch #{@new_resource.revision}") do
    # checkout into a local branch rather than a detache@new_resource.checkout_branchd HEAD
    current_branch = shell_out!("git rev-parse --abbrev-ref HEAD") # I don't know what this returns, but let's pretend it just returns stdout
    if current_branch != @new_resource.checkout_branch then
        shell_out!("git branch -f #{@new_resource.checkout_branch} #{sha_ref}", run_options(:cwd => @new_resource.destination))
    end
    shell_out!("git checkout #{@new_resource.checkout_branch}", run_options(:cwd => @new_resource.destination))
    Chef::Log.info "#{@new_resource} checked out branch: #{@new_resource.revision} onto: #{@new_resource.checkout_branch} reference: #{sha_ref}"
  end
end
```

Thanks @jordane for helping me find where the issue was.

iankronquist:
By default, git will check out the master branch when a repository is first
cloned (unless the repo specifies a different default branch). The git provider
will then attempt to force update to the `checkout_branch`, but will fail
because you can't force checkout a branch you are on.  An example on the
command line:

```
$ git checkout master
$ git branch -f master abcde12345
fatal: Cannot force update the current branch.
```

Instead, run:

```
$ git checkout -fB master abcde12345
```

Fixes #3025

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -157,8 +157,7 @@ class Chef
 
         converge_by("checkout ref #{sha_ref} branch #{@new_resource.revision}") do
           # checkout into a local branch rather than a detached HEAD
-          shell_out!("git branch -f #{@new_resource.checkout_branch} #{sha_ref}", run_options(:cwd => @new_resource.destination))
-          shell_out!("git checkout #{@new_resource.checkout_branch}", run_options(:cwd => @new_resource.destination))
+          shell_out!("git checkout -fB #{@new_resource.checkout_branch} #{sha_ref}", run_options(:cwd => @new_resource.destination))
           Chef::Log.info "#{@new_resource} checked out branch: #{@new_resource.revision} onto: #{@new_resource.checkout_branch} reference: #{sha_ref}"
         end
       end
--- a/spec/unit/provider/git_spec.rb
+++ b/spec/unit/provider/git_spec.rb
@@ -311,9 +311,7 @@ SHAS
   end
 
   it "runs a checkout command with default options" do
-    expect(@provider).to receive(:shell_out!).with('git branch -f deploy d35af14d41ae22b19da05d7d03a0bafc321b244c', :cwd => "/my/deploy/dir",
-                                                             :log_tag => "git[web2.0 app]").ordered
-    expect(@provider).to receive(:shell_out!).with('git checkout deploy', :cwd => "/my/deploy/dir",
+    expect(@provider).to receive(:shell_out!).with('git checkout -fB deploy d35af14d41ae22b19da05d7d03a0bafc321b244c', :cwd => "/my/deploy/dir",
                                                              :log_tag => "git[web2.0 app]").ordered
     @provider.checkout
   end


iankronquist:
Sorry about commit 8468a38. I forgot to rebase, so I had to force push, and github isn't smart enough to know that.
I would like to write a regression test for this behavior. Would it be appropriate to stick something in `kitchen-tests`?

iankronquist:
@danielsdeleo You seem to be someone with authority over this repo. I made the pull request a while ago and nobody has taken a look at it.

danielsdeleo:
I think this is what `enable_checkout false` is for. Are you doing a one-time clone of a bunch of repos or are you actually managing the revision but trying to stay on master?

iankronquist:
@danielsdeleo The latter. Code which is merged onto `master` is ready for production. Chef should pick up new changes on the next chef run.

danielsdeleo:
@iankronquist why does the checkout branch name matter then? As long as the revision is the same, you have the same code.

iankronquist:
The revision will change. We want the code in production to track the tip of the master branch. If new commits get pushed to master we want chef to update the code to the tip of master.

danielsdeleo:
That's exactly how chef works by default, just the branch name on the host is "deploy" and not "master." If you tell chef to track master, it'll pull in updates as they are available and force overwrite the "deploy" branch with the new commits.

iankronquist:
Yes, however, consider the following invocation of the git provider:
```ruby                                            
 git '/opt/fenestra' do
  action :sync
  repository 'https://github.com/osuosl/fenestra'
  checkout_branch 'master'
end 
```

Chef will
1. Clone the repo and check out master
2. [Force branch to master](https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L160), but you can't force branch to a branch you're already on, so it fails.

Maybe there's a better way to do this, but when I use the git resource in a way very similar to the one above it fails.

danielsdeleo:
You get exactly the behavior you want if you just omit `checkout_branch 'master'`. If you want to be explicit that you're tracking the upstream's master branch, then `revision "master"` is what you want.

danielsdeleo:
BTW, I'm not fundamentally opposed to your change, it just seems to me like what you need to do is already supported. Maybe it's a docs issue?

iankronquist:
I'm going to have to put together a cookbook which demonstrates my issue. It may well be a docs issue. I'll be out of town for a little while so I'll bother you in a week or two.

iankronquist:
@danielsdeleo FYI, more other people have run into the issue and left comments recently.

As promised, [Here](https://github.com/iankronquist/chef-bug-demo) is the cookbook that demonstrates the issue. To see the bug in action just run `kitchen test`. The README has more information.

It could be a documentation bug, but I don't think so.

danielsdeleo:
Right, so the reason I say it's a documentation issue is that the git provider is designed to work two ways:

1. You want to always sync a particular commit (which can change over time based on use of attributes, etc.). In this case you identify what commit you want from the upstream repo using the `revision` property of the git resource. The git provider will locally maintain its own branch and force it to whatever commit you want. This way, if you want to switch from one branch to another, or only deploy certain tags (maybe you set with attributes or a data bag item), you just set the `revision` and the provider will always ensure you're on the correct commit by checking `git ls-remote` on the upstream repo.

2. You want to do a one-time checkout. Usually this happens when you're maintaining developers' laptops with Chef. In this case, you set `enable_checkout` to false which prevents the git provider from attempting to checkout master to master. This usage was sort of bolted on to the git provider after use case (1), so it's a bit unwieldy, I admit.

You seem to have use case (1), so you just need to set `revision` to "master" and not use the `checkout_branch` option and it should work the way you expect.

About it being a docs issue, the docs don't explain these two use cases (or much about how things work under the hood) so it seems like you're confusing resource properties for use case (2) when trying to implement use case (1).

danielsdeleo:
That is to say, 

change:

```ruby
git '/opt/fenestra' do
  action :sync
  repository 'https://github.com/osuosl/fenestra'
  checkout_branch 'master'
end 
```

to 

```ruby
git '/opt/fenestra' do
  action :sync
  repository 'https://github.com/osuosl/fenestra'
  revision 'master'
end
```

And each chef run will set the checked out code to the latest on master, as determined by running `git ls-remote REPO_URL master`

iankronquist:
It looks like it's a docs issue then.

danielsdeleo:
@chef/client-core thoughts? I'm happy to just copy the text from above to an issue on chef-docs and call this done. On the other hand, it's not really that big a deal if we let people set the checkout branch to master. But then again, I think people only want to do that because of a misunderstanding of how the provider actually works and maybe they'll get into trouble if they later try to change what branch/tag/whatever they track on the upstream by setting `checkout_branch` and it doesn't work.

thommay:
Agree we should just get the docs fixed.

danielsdeleo:
https://github.com/chef/chef-docs/issues/552

PR #3295, opened 2015-04-28
don't mutate the new resource

### LINKED ISSUE #3168: Exceptions::ChecksumMismatch with different checksum found
Test cookbook `test-notify.tar.gz` attached in ZenDesk 3636
Relates to #2689 

On the first run ever of the cookbook, when the `/tmp/template.file` is notified by the `/tmp/template2.file` of a change either `:immediately` or `:delayed`, a failure occurs in Chef 12.0.1+ . Works fine in Chef Client 11.16.4.

It appears the problem is that the `template` resource is thinking "Hey, my content still has the same checksum I had when I came into existence." However, when the notification from `/tmp/template2.file` goes out, then the content of `/tmp/template2.file` actually does change to `89065a`, but the resource did not expect it.

A workaround for this issue is
```
    template '/tmp/template.file' do
      source 'template.file.erb'
    end

    template '/tmp/template2.file' do
      source 'template2.file.erb'
      notifies :run, 'ruby_block[clear_checksum]', :immediately
    end

    ruby_block "clear_checksum" do
      block do
	template_r = run_context.resource_collection.find(:template => '/tmp/template.file')
	template_r.checksum = nil
      end
      action :nothing
      notifies :create, 'template[/tmp/template.file]', :immediately
    end
```

```
$ shasum -a 256 /tmp/template.file
89065ac828e9f1211cf80c8fd66c4c6777b9935c5dcc321b7b7f7b0caf4ef9fb  /tmp/template.file

$ cat /tmp/template.file
Hello
CHEF!
```

**Cookbook templates and recipe**
```
#templates/default/template.file.erb
Hello
<%="CHEF!" if ::File.exist?('/tmp/template2.file')%>

#/templates/default/template2.file.erb
chef!
```

```
#recipes/default.rb

template '/tmp/template.file' do
  source 'template.file.erb'
end

template '/tmp/template2.file' do
  source 'template2.file.erb'
  notifies :create, 'template[/tmp/template.file]', :immediately
end
```

**Example First Run with clear cache and failure**
```
Starting Chef Client, version 12.0.1
resolving cookbooks for run list: ["test-notify"]
Synchronizing Cookbooks:
  - test-notify
Compiling Cookbooks...
Converging 2 resources
Recipe: test-notify::default
  * template[/tmp/template.file] action create
    - create new file /tmp/template.file
    - update content in file /tmp/template.file from none to 27d5d8
    --- /tmp/template.file	2015-03-30 21:39:52.759586197 +0000
    +++ /tmp/chef-rendered-template20150330-15052-r4219x	2015-03-30 21:39:52.759586197 +0000
    @@ -1 +1,3 @@
    +Hello
    +
  * template[/tmp/template2.file] action create
    - create new file /tmp/template2.file
    - update content in file /tmp/template2.file from none to 101150
    --- /tmp/template2.file	2015-03-30 21:39:52.763586197 +0000
    +++ /tmp/chef-rendered-template20150330-15052-14omhyn	2015-03-30 21:39:52.763586197 +0000
    @@ -1 +1,2 @@
    +chef!
  * template[/tmp/template.file] action create

    Error executing action `create` on resource 'template[/tmp/template.file]'

    Chef::Exceptions::ChecksumMismatch
    Checksum on resource (27d5d8) does not match checksum on content (89065a)

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/test-notify/recipes/default.rb

     15: template '/tmp/template.file' do
     16:   source 'template.file.erb'
     17: end
     18:

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/test-notify/recipes/default.rb:15:in `from_file'

    template("/tmp/template.file") do
      action "create"
      updated true
      retries 0
      retry_delay 2
      path "/tmp/template.file"
      backup 5
      atomic_update true
      source "template.file.erb"
      cookbook_name "test-notify"
      recipe_name "default"
      owner "root"
      group "root"
      checksum "27d5d8efee1938c9a80bd286b5213b953a05636f17b1bdd28ce62e6973084ca0"
      mode "0644"
    end
```

**Example second run after the diff is resolved does not fail**
```
Starting Chef Client, version 12.0.1
resolving cookbooks for run list: ["test-notify"]
Synchronizing Cookbooks:
  - test-notify
Compiling Cookbooks...
Converging 2 resources
Recipe: test-notify::default
  * template[/tmp/template.file] action create
    - update content in file /tmp/template.file from 27d5d8 to 89065a
    --- /tmp/template.file	2015-03-30 21:39:52.759586197 +0000
    +++ /tmp/chef-rendered-template20150330-15247-11areh2	2015-03-30 21:44:45.227586781 +0000
    @@ -1,3 +1,3 @@
     Hello
    -
    +CHEF!
  * template[/tmp/template2.file] action create (up to date)

Running handlers:
Running handlers complete
Chef Client finished, 1/2 resources updated in 1.827960795 seconds
```

lamont-granquist:
this line violates the rule that we never mutate the new resource.

in Chef 12 this causes real problems because if a file resource is
notified (run twice) then the rendered content is loaded into the
new resource and if the file resource content changes (i.e. its a
template with logic which changes when its notified) then it will
cause a failure because the rendered content will not match the
'requested' checksum.  where the 'requested' checksum is actually
the checksum loaded by this line.

fundamentally we have three different states that we're trying to
track:

- current state
- initial state
- requested state

the removed line tries to use the @new_resource for reporting initial
state, which then bleeds over into what chef thinks is requested state
in the next invokation.

without constructing a third @initial_resource and using that for
resource reporting comparison we can't solve this problem "right".

the tradeoff is we either break reporting here or break chef-client
runs.  this patch sacrifices reporting in order to make chef-client
work.

fixes #3168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@
 
 * Convert bootstrap template to use sh #2877
 * [Issue #3316](https://github.com/chef/chef/issues/3316) Fix idempotency issues with the `windows_package` resource
+* [pr#3295](https://github.com/chef/chef/pull/3295): Stop mutating `new_resource.checksum` in file providers.  Fixes some ChecksumMismatch exceptions like [issue#3168](https://github.com/chef/chef/issues/3168)
 
 ## 12.3.0
+
 * [pr#3160](https://github.com/chef/chef/pull/3160): Use Chef Zero in
   socketless mode for local mode, add `--no-listen` flag to disable port
   binding
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -388,9 +388,6 @@ class Chef
         do_backup unless needs_creating?
         deployment_strategy.deploy(tempfile.path, ::File.realpath(@new_resource.path))
         Chef::Log.info("#{@new_resource} updated file contents #{@new_resource.path}")
-        if managing_content?
-          @new_resource.checksum(checksum(@new_resource.path)) # for reporting
-        end
       end
 
       def do_contents_changes
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -386,8 +386,12 @@ class Chef
 
       def update_file_contents
         do_backup unless needs_creating?
-        deployment_strategy.deploy(tempfile.path, ::File.realpath(@new_resource.path))
-        Chef::Log.info("#{@new_resource} updated file contents #{@new_resource.path}")
+        deployment_strategy.deploy(tempfile.path, ::File.realpath(new_resource.path))
+        Chef::Log.info("#{new_resource} updated file contents #{new_resource.path}")
+        if managing_content?
+          # save final checksum for reporting.
+          new_resource.final_checksum = checksum(new_resource.path)
+        end
       end
 
       def do_contents_changes
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -467,12 +467,21 @@ class Chef
     #
     # @return [Hash{Symbol => Object}] A Hash of attribute => value for the
     #   Resource class's `state_attrs`.
-    def state
+    def state_for_resource_reporter
       self.class.state_attrs.inject({}) do |state_attrs, attr_name|
         state_attrs[attr_name] = send(attr_name)
         state_attrs
       end
     end
+
+    #
+    # Since there are collisions with LWRP parameters named 'state' this
+    # method is not used by the resource_reporter and is most likely unused.
+    # It certainly cannot be relied upon and cannot be fixed.
+    #
+    # @deprecated
+    #
+    alias_method :state, :state_for_resource_reporter
 
     #
     # The value of the identity attribute, if declared. Falls back to #name if
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -38,6 +38,15 @@ class Chef
 
       attr_writer :checksum
 
+      #
+      # The checksum of the rendered file.  This has to be saved on the
+      # new_resource for the 'after' state for reporting but we cannot
+      # mutate the new_resource.checksum which would change the
+      # user intent in the new_resource if the resource is reused.
+      #
+      # @returns [String] Checksum of the file we actually rendered
+      attr_accessor :final_checksum
+
       provides :file
 
       def initialize(name, run_context=nil)
@@ -128,6 +137,15 @@ class Chef
         else
           @verifications
         end
+      end
+
+      def state_for_resource_reporter
+        state_attrs = super()
+        # fix up checksum state with final_checksum saved by the provider
+        if checksum.nil? && final_checksum
+          state_attrs['checksum'] = final_checksum
+        end
+        state_attrs
       end
     end
   end
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -143,7 +143,7 @@ class Chef
         state_attrs = super()
         # fix up checksum state with final_checksum saved by the provider
         if checksum.nil? && final_checksum
-          state_attrs['checksum'] = final_checksum
+          state_attrs[:checksum] = final_checksum
         end
         state_attrs
       end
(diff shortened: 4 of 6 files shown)


lamont-granquist:
@chef/client-engineers @chef/client-core assuming travis passes this needs review.

this a good example of the 'whys' behind the principle that we should never be updating the new resource from within the provider (in this case because notifications may invoke the the resource again and we wind up mutating what was 'requested').

this might be an API breaking change to reporting but we have three options:

1. break reporting of checksum diffs (this change)
2. leave this broken behavior in the chef-client (do nothing)
3. do a fairly epic change to introduce a third resource associated with providers that tracks initial state and replaces the `new_resource` for the purposes of reporting.

that last one should ideally get done, but its both a very large change and potentially breaking.

jaymzh:
well... this is a bug we run into often. Thanks for tracking it down.

Also... I imagine this will break some very serious assumptions.

Also, we already do this a LOT in `Chef::Provider::Package::Yum`

danielsdeleo:
So, a compromise: use a different parameter name for content checksums that has no meaning in the file resource DSL for reporting purposes, patch the resource reporter to use that.

lamont-granquist:
That's a little hacky, but I guess its on par with how hacky it is now...

lamont-granquist:
and @jaymzh now you know why that's bad and can look at fixing it before it turns into bug reports...

danielsdeleo:
> That's a little hacky, but I guess its on par with how hacky it is now...

IMO, there's a ton of places where the resource object does a lot more than describe the desired final state of a component, e.g., specifying the path to a package installed from disk, setting start/stop/restart commands on a service, etc. all describe _how_ not what and conceptually are more like provider customizations. But we deal with it because the other options are not pragmatic. What's one more hack between friends? :sunglasses:

lamont-granquist:
Well, the new_resource is the user-requested state, though.  When you mutate that, you lose the accurate intent of the user, which is what causes these issues when it gets reused and you've lost that accurate intent.  Which is different from strictly declarative state.

lamont-granquist:
@danielsdeleo what do you think about this?  I went ahead and solved the Chef::Resource#state problem by just renaming the method, then i can override it in the File resource class.

jdmundrawala:
that's cool

danielsdeleo:
Not the most heartwarming code, but fixes the bug and doesn't break anything else (AFICT). :+1:

danielsdeleo:
Just to be clear, my comment, "Not the most heartwarming code" is referring to the hacky solution we have to put up with because of the constraints mentioned in this thread.

lamont-granquist:
Lol, yep.

@chef/client-core @chef/client-engineers should be ready for review.

jaymzh:
LGTM
:+1: 

There's other resources that use `file` though (like `cookbook_file` and `template`)... I looked through the code and I don't see anything that would be affected by these changes.. but something to think about.

lamont-granquist:
Everything inherits so it should be all good.  The `if managing_content?` bit is already a hook that gets overridden by those resources to define what semantics are for if they're dropping content or not so that logic is already protected against inheritance.

lamont-granquist:
And I just checked for the use of `new_resource.checksum` and its either used in `managing_content?` or its used in `current_resource_matches_target_checksum?` in the remote_file/content.rb code which both happen before the code path which changed.  I don't see any consumers of that who execute after the code path that changed that would need to be changed to use `new_resource.final_checksum`.

btm:
I was looking for something in Travis and saw this failed in master after being merged (and running the integration tests that aren't run on PRs). But I can't tell that it was related, could be an intermittent failure? Gotta pay attention to something else right now.

https://travis-ci.org/chef/chef/jobs/60869991

lamont-granquist:
looks like travis can't run tests against PRs after they're merged or something which is why there's a red x here.  the failure on master looks like some unrelated networking related hiccup.

btm:
The travis failure on master is unrelated, but it's not a hiccup. All of the builds on master (which run integration tests) have been failing for around three weeks. I've described it over here: https://github.com/chef/chef/issues/3323

PR #3297, opened 2015-04-29
Implemented `knife user key list` and `knife client key list`.

tylercloke:
Abstracted out a bunch of common tests with the key create code to reduce redundancy.

Ping @chef/lob

### DIFF
--- a/lib/chef/knife/key_list.rb
+++ b/lib/chef/knife/key_list.rb
@@ -1,0 +1,88 @@
+#
+# Author:: Tyler Cloke (<tyler@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/key'
+require 'chef/json_compat'
+require 'chef/exceptions'
+
+class Chef
+  class Knife
+    # Service class for UserKeyList and ClientKeyList, used to list keys.
+    # Implements common functionality of knife [user | org client] key list.
+    #
+    # @author Tyler Cloke
+    #
+    # @attr_accessor [Hash] cli input, see UserKeyList and ClientKeyList for what could populate it
+    class KeyList
+
+      attr_accessor :config
+
+      def initialize(actor, list_method, ui, config)
+        @actor = actor
+        @list_method = list_method
+        @ui = ui
+        @config = config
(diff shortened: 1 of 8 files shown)


danielsdeleo:
@chef/client-core

spec/unit/knife/key_create_spec.rb
```diff
@@ -75,26 +51,22 @@
...
-        allow(command).to receive(:service_object).and_return(service_object)
-        allow(command).to receive(:name_args).and_return(["charmander"])
-      end
-
-      context "when the command is successful" do
-        before do
-          expect(service_object).to receive(:run)
-        end
-      end
-    end
```
tylercloke: All these deleted tests moved to common file to run against multiple classes.

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -216,7 +216,6 @@ def format_cookbook_list_for_display(item)
             end
           end
         end
-
```
jdmundrawala: Can you undo this

lib/chef/knife/key_list.rb
```diff
@@ -0,0 +1,88 @@
...
+          max_length = 0
+          keys.each do |key|
+            key['name'] = key['name'] + ":"
+            max_length = key['name'].length if key['name'].length > max_length
+          end
+          keys.each do |key|
+            next if !key['expired'] && @config[:only_expired]
+            next if key['expired'] && @config[:only_non_expired]
+            display = "#{colorize(key['name']).ljust(max_length)} #{key['uri']}"
+            display = display + " (expired)" if key["expired"]
```
jdmundrawala: I like "#{display} (expired)" better

lib/chef/knife/key_list.rb
```diff
@@ -0,0 +1,88 @@
...
+
+class Chef
+  class Knife
+    # Service class for UserKeyList and ClientKeyList,
+    # Implements common functionality of knife [user | org client] key list.
+    #
+    # @author Tyler Cloke
+    #
+    # @attr_accessor [Hash] cli input, see UserKeyList and ClientKeyList for what could populate it
+    class KeyList
```
jdmundrawala: nitpicky...you dont have to change this, but is the a list of keys or a thing that lists keys?

lib/chef/knife/core/generic_presenter.rb
tylercloke: Oops.

lib/chef/knife/key_list.rb
tylercloke: A thing that lists keys. I can see how that would be confusing. I was following the convention I set with `KeyCreate` for the key creation service class.

jdmundrawala:
:+1:

lib/chef/knife/key_list.rb
btm: What about this? ;)
```
display << " (expired)" if key["expired"]
```

btm:
:+1: 

I do wonder if the API allows listing keys without providing an actor, which could lead to something like ```knife user key list --all``` or even ```knife key list``` and ```knife key search```, but it looks like Chef::Key doesn't support that now anyway.

tylercloke:
Yeah, the API is scoped to actors at the moment, but if that's something we would see as a useful feature in the future we should chat about it and look into prioritization :)

PR #3090, opened 2015-03-16
We should not change how exceptions are raised if audit mode is disabled

jdmundrawala:
Patch for #3078

This is not the final solution, we need to think hard about the way we do this now that we know how people will break.

cc @chef/client-core

PR #3299, opened 2015-04-29
Remove experimental warning on audit mode.

juliandunn:
I'm going to suggest we remove this to avoid freaking out customers since we have been shipping this mode for some time.

/cc @chef/analytics-data-integrations

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -320,12 +320,6 @@ class Chef::Application::Client < Chef::Application
       unless expected_modes.include?(mode)
         Chef::Application.fatal!(unrecognized_audit_mode(mode))
       end
-
-      unless mode == :disabled
-        # This should be removed when audit-mode is enabled by default/no longer
-        # an experimental feature.
-        Chef::Log.warn(audit_mode_experimental_message)
-      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


jdmundrawala:
Did we solve the issue around audit mode mangling exceptions?

juliandunn:
What do you mean by "mangling exceptions"?

jdmundrawala:
https://github.com/chef/chef/commit/625373d40febafa1c6621aaf179852601d303e6c#diff-c34d08382bb25f61e44ecbd5f50f4166
This was one of the regressions in 12.2 and we ended up patching it. By making audit mode the default, those same people will break again.

jdmundrawala:
I think the example we got was SystemExit and how this broke people's CI. Not sure if there is anything else that we need to consider.
cc @tyler-ball

juliandunn:
My PR doesn't make audit mode the default; I don't know that we need to do that right now. I'm just removing warnings here.

jdmundrawala:
My bad, I guess this is harmless then.

tyler-ball:
@jdmundrawala If audit mode is disabled (the default) we don't wrap any raised exceptions, we just re-raise them - https://github.com/chef/chef/pull/3090

tyler-ball:
:+1: On this change.  I could see people being concerned this message gets changed in a patch version.  Can we hold off on it until the next major version (12.4)?

juliandunn:
@tyler-ball Sure, that's fine

juliandunn:
Per @tyler-ball's message -- looks like we're on the cusp of releasing 12.4. Can I get this warning removal in there? /cc @chef/client-maintainers

danielsdeleo:
:+1: as long as we're sufficiently confident that the design is fairly stable and we don't need to change it in incompatible ways in the future.

jtimberman:
The main thing off the top of my head is this 

https://github.com/chef/chef/issues/3482

It's a bug fix but if people did the should syntax in their tests that change would break them.

juliandunn:
@jtimberman I'm going to argue that whether that's a bug or not doesn't generally change the shape of the design of audit mode, and at this point we've got enough usage & published enough articles about it that we should just say it's shipped.

danielsdeleo:
Well then you shouldn't break any working code. OTOH, that may be a moot point: https://github.com/chef/chef/issues/3482#issuecomment-108708561

PR #3314, opened 2015-04-30
Updating to released test kitchen and kitchen vagrant

tyler-ball:
@jdmundrawala This is a follow up to #3260


jdmundrawala:
my bad, i must not have pushed this
:shipit:

Issue #3316, opened 2015-04-30
windows_package broken in core in Chef 12.2.0+


btm:
Simple recipe:
```
package "7zip" do
  source 'c:/ruby/7z920-x64.msi'
end
```

Works on Chef 12.1.2:
```
PS C:\ruby> chef-apply install_7zip.rb
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * windows_package[7zip] action install (up to date)
PS C:\ruby> chef-apply -v
Chef: 12.1.2
```

Fails on Chef 12.2.1:
```
PS C:\ruby> chef-apply -v
Chef: 12.2.1
PS C:\ruby> chef-apply install_7zip.rb
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * windows_package[7zip] action install
    * No version specified, and no candidate version available for 7zip
    ================================================================================
    Error executing action `install` on resource 'windows_package[7zip]'
    ================================================================================

    Chef::Exceptions::Package
    -------------------------
    No version specified, and no candidate version available for 7zip

    Resource Declaration:
    ---------------------
    # In install_7zip.rb

      1: package "7zip" do
      2:   source 'c:/ruby/7z920-x64.msi'
      3: end

    Compiled Resource:
    ------------------
    # Declared in install_7zip.rb:1:in `run_chef_recipe'

    windows_package("7zip") do
      action :install
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      package_name "7zip"
      source "c:\\ruby\\7z920-x64.msi"
      version "9.20.00.0"
      timeout 600
      returns [0]
      declared_type :package
      cookbook_name "(chef-apply cookbook)"
      recipe_name "(chef-apply recipe)"
    end

[2015-04-30T15:49:33-04:00] FATAL: Stacktrace dumped to C:/chef/cache/chef-stacktrace.out
[2015-04-30T15:49:33-04:00] FATAL: Chef::Exceptions::Package: windows_package[7zip] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: Chef::Exceptions::Package: No version specified, and no candidate version available for 7zip
```

The regression appears to be caused by https://github.com/chef/chef/pull/3034.

btm:
Fixed by #3317, in master, will be in the next release.

mrwizard82d1:
I still see this issue running chef-client 12.3.0:

$ chef-client --local-mode --log_level info --runlist 'recipe[test-3316]'
[...]

PR #3034, opened 2015-03-06
Remove trailing null character from msi properties

gh2k:
When reading properties from an MSI, variables are returned with a trailing null byte. I don't think that should happen.

Results in compiled windows_package resources with an incorrect version string, for example.

PR #3317, opened 2015-04-30
windows_package is idempotent again

jdmundrawala:
This was broken in #3034. The issue is that the package does not get detected
as installed because the version that is installed has \x00 at the end,
while the version from the msi does not. This fails comparison, and we fall
into code that does not use source and thus requires candidate version and we
die. (Or something like that)

Solves #3316


ksubrama:
:+1:

smurawski:
:+1:

btm:
:+1:

PR #3318, opened 2015-04-30
Modify windows package provider to allow url

jdmundrawala:
This patch allows people to specify a url for source for the `windows_package` resource.

cc @btm @smurawski @ksubrama

### DIFF
--- a/lib/chef/mixin/uris.rb
+++ b/lib/chef/mixin/uris.rb
@@ -1,0 +1,33 @@
+#
+# Author:: Jay Mundrawala (<jdm@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'uri'
+
+class Chef
+  module Mixin
+    module Uris
+      def uri_scheme?(source)
+        begin
+          !URI.split(source).first.nil?
+        rescue URI::InvalidURIError
+          return false
+        end
+      end
+    end
+  end
+end
--- a/lib/chef/mixin/uris.rb
+++ b/lib/chef/mixin/uris.rb
@@ -21,6 +21,9 @@ require 'uri'
 class Chef
   module Mixin
     module Uris
+      # uri_scheme? returns true if the string starts with
+      # scheme://
+      # For example, it will match http://foo.bar.com
       def uri_scheme?(source)
         # From open-uri
         !!(%r{\A[A-Za-z][A-Za-z0-9+\-\.]*://} =~ source)
--- a/lib/chef/provider/package/windows.rb
+++ b/lib/chef/provider/package/windows.rb
@@ -16,15 +16,16 @@
 # limitations under the License.
 #
 
+require 'chef/mixin/uris'
 require 'chef/resource/windows_package'
 require 'chef/provider/package'
 require 'chef/util/path_helper'
-require 'uri'
 
 class Chef
   class Provider
     class Package
       class Windows < Chef::Provider::Package
+        include Chef::Mixin::Uris
 
         provides :package, os: "windows"
         provides :windows_package, os: "windows"
@@ -38,7 +39,7 @@ class Chef
         # load_current_resource is run in Chef::Provider#run_action when not in whyrun_mode?
         def load_current_resource
           @current_resource = Chef::Resource::WindowsPackage.new(@new_resource.name)
-          if should_download?
+          if download_file_missing?
             Chef::Log.debug("We do not know the version of #{new_resource.source} because the file is not downloaded")
             current_resource.version(:unknown.to_s)
           else
@@ -77,7 +78,7 @@ class Chef
         end
 
         def action_install
-          if should_download?
+          if uri_scheme?(new_resource.source)
             download_source_file
             load_current_resource
           end
@@ -102,8 +103,8 @@ class Chef
 
         private
 
-        def should_download?
-          is_url?(new_resource.source) && !::File.exists?(source_location)
+        def download_file_missing?
+          uri_scheme?(new_resource.source) && !::File.exists?(source_location)
         end
 
         def resource_for_provider
@@ -129,7 +130,7 @@ class Chef
 
         def source_location
           @source_location ||= begin
-            if is_url?(new_resource.source)
+            if uri_scheme?(new_resource.source)
               uri = ::URI.parse(new_resource.source)
               filename = ::File.basename(::URI.unescape(uri.path))
               file_cache_dir = Chef::FileCache.create_cache_path("package/")
@@ -139,17 +140,6 @@ class Chef
             end
           end
         end
-
-        def is_url?(source)
-          begin
-            scheme = URI.split(source).first
-            return false unless scheme
-            %w(http https ftp file).include?(scheme.downcase)
-          rescue URI::InvalidURIError
-            return false
-          end
-        end
-
       end
     end
   end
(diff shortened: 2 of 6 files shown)


lib/chef/provider/package/windows.rb
```diff
@@ -36,19 +37,23 @@ class Windows < Chef::Provider::Package
 
         # load_current_resource is run in Chef::Provider#run_action when not in whyrun_mode?
         def load_current_resource
-          @new_resource.source(Chef::Util::PathHelper.validate_path(@new_resource.source))
-
           @current_resource = Chef::Resource::WindowsPackage.new(@new_resource.name)
-          @current_resource.version(package_provider.installed_version)
-          @new_resource.version(package_provider.package_version)
-          @current_resource
+          if should_download?
```
btm: So the obvious design debate here is, should we download the file during load_current_resource?

jdmundrawala: What are the advantages of doing it during load_current_resource?

lib/chef/provider/package/windows.rb
```diff
@@ -80,6 +93,63 @@ def install_package(name, version)
...
+              filename = ::File.basename(::URI.unescape(uri.path))
+              file_cache_dir = Chef::FileCache.create_cache_path("package/")
+              Chef::Util::PathHelper.cleanpath("#{file_cache_dir}/#{filename}")
+            else
+              Chef::Util::PathHelper.cleanpath(new_resource.source)
+            end
+          end
+        end
+
+        def is_url?(source)
```
btm: I see this example code on the tubes, but it is also basically the same code as ```Chef::Provider::Package::Rpm.uri_scheme?```. I wonder if we have to hard code the list of schemes here? It'd be best if we could just have that in Chef::Provider::RemoteFile::Fetcher and let that raise the error.

URI.scheme_list seems limited, like ```URI.scheme_list.include?(URI.parse(source).scheme.upcase)```. Maybe a constant like ```Chef::Provider::RemoteFile::Fetcher::SUPPORTED_SCHEMES```? Maybe it doesn't matter.

lib/chef/provider/package/windows.rb
btm: During compile, without the file we don't know the product id (for MSI), so we don't know if the package is installed or not. Aren't we going to always decide that it's not installed and always run the install action during converge?

lib/chef/provider/package/windows.rb
```diff
@@ -80,6 +93,63 @@ def install_package(name, version)
...
+            r.options(new_resource.options)
+          end
+        end
+
+        def download_source_file
+          source_resource.run_action(:create)
+          Chef::Log.debug("#{@new_resource} fetched source file to #{source_resource.path}")
+        end
+
+        def source_resource
```
btm: how do you set checksum on the remote_file resource? Can we have an attribute that we pass a hash of attributes that gets set on the remote_file resource?

lib/chef/resource/windows_package.rb
```diff
@@ -69,10 +69,26 @@ def source(arg=nil)
...
+            @source = arg
+          else
+            @source = ::File.absolute_path(arg).gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
+          end
         end
       end
+
+      private
+
+      def is_url?(source)
```
btm: Whatever we do, this makes three places we have this method (this resource/provider, plus rpm provider), it should probably be a mixin.

lib/chef/provider/package/windows.rb
jdmundrawala: Kind of... this resource will become 2 resources, and one of those 2 will need to be converged.

lib/chef/provider/package/windows.rb
jdmundrawala: Yep, that shouldn't be hard to do. However, as I went down this road, it sort of felt like if you wanted to do anything fancy like that, you should just use remote file. Not sure if that is a valid argument as I do feel like I'm a little biased here.

I originally wanted to let people pass a remote file resource as the source attribute (with all the chef dsl-ing), however at that point, you may as well just declare it with the dsl that's already there.

lib/chef/provider/package/windows.rb
```diff
@@ -80,6 +93,63 @@ def install_package(name, version)
...
+        # @return [Array] new_version(s) as an array
+        def new_version_array
+          # Because the one in the parent caches things
+          [new_resource.version]
+        end
+
+        private
+
+        def should_download?
+          is_url?(new_resource.source) && !::File.exists?(source_location)
```
btm: If we have a URL and an old version of the package on disk?

jdmundrawala: yep, that's wrong. good catch

lib/chef/provider/package/windows.rb
jdmundrawala: I've mixed too many concepts here. `uri_scheme?` shouldn't care about what schemes are supported, seems like remote_file's job. I'm going to update it for that, let me know what you think.

jdmundrawala:
@btm added mixin for Uris and a way to override remote_file attributes

ksubrama:
Left shark gives a :+1:

smurawski:
@jdmundrawala How do the remote file attributes get surfaced in the DSL?  Are they top level attributes with the standard package ones?  Or are they nested?  If nested, we should really surface at least checksum as a top-level attribute (to smooth the transition from the Windows cookbook).

lib/chef/mixin/uris.rb
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'uri'
+
+class Chef
+  module Mixin
+    module Uris
+      def uri_scheme?(source)
```
btm: Can you add a comment here about what this returns? I'm guessing it's true if there is a scheme in the ```scheme://``` format?

lib/chef/provider/package/windows.rb
```diff
@@ -80,6 +94,60 @@ def install_package(name, version)
...
+          source_resource.run_action(:create)
+          Chef::Log.debug("#{@new_resource} fetched source file to #{source_resource.path}")
+        end
+
+        def source_resource
+          @source_resource ||= Chef::Resource::RemoteFile.new(default_download_cache_path, run_context).tap do |r|
+            r.source(new_resource.source)
+            r.backup(false)
+
+            if new_resource.remote_file_attributes
```
btm: Can we get a RELEASE_NOTES.md entry with an example of how to use this?

lib/chef/mixin/uris.rb
```diff
@@ -0,0 +1,30 @@
...
+        # From open-uri
+        !!(%r{\A[A-Za-z][A-Za-z0-9+\-\.]*://} =~ source)
```
btm: This is nice. It looks like it works correctly in all the cases I was worried about before.

lib/chef/resource/windows_package.rb
```diff
@@ -69,10 +71,22 @@ def source(arg=nil)
           @source
         else
           raise ArgumentError, "Bad type for WindowsPackage resource, use a String" unless arg.is_a?(String)
-          Chef::Log.debug("#{package_name}: sanitizing source path '#{arg}'")
-          @source = ::File.absolute_path(arg).gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
+          if uri_scheme?(arg)
+            @source = arg
+          else
+            @source = ::File.absolute_path(arg).gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
```
btm: Probably should switch to ```Chef::Util::PathHelper.canonical_path``` now, which would also add long path support with the \\?\ business.

lib/chef/provider/package/windows.rb
```diff
@@ -80,6 +94,60 @@ def install_package(name, version)
...
+
+        # @return [Array] new_version(s) as an array
+        def new_version_array
+          # Because the one in the parent caches things
+          [new_resource.version]
+        end
+
+        private
+
+        def download_file_missing?
```
btm: downloadable_file_missing?

lib/chef/resource/windows_package.rb
jdmundrawala: msiexec does not like it

jdmundrawala:
surfaced checksum attribute

lib/chef/resource/windows_package.rb
btm: The \\\\?\\ prepend? You can pass an argument to canonical_path to not add it.

jdmundrawala: done

smurawski:
:+1: Rock on Jay.

btm:
:+1: :+1: 

Awesome work. /cc @jeremiahsnapp

PR #3319, opened 2015-04-30
Implemented X-Ops-Server-API-Version in Chef requests

tylercloke:
Ping @chef/lob @marcparadise 

### Description

Implemented `X-Ops-Server-API-Version` header and error handling for unsupported server API versions. This takes advantage of `Chef::HTTP::Authenticator.handle_request`. After the `authentication_headers` have been merged, it adds:

`{'X-Ops-Server-API-Version' => Chef::HTTP::Authenticator::SERVER_API_VERSION}`

`Chef::HTTP::Authenticator::SERVER_API_VERSION` should correspond to the max version of the server the client is compatible with. We are starting at `"0"`.

#### rest_spec.rb fixup

I also removed outer level lets `:base_header` and `:req_with_body_headers` from `rest_spec.rb` as they were never used anywhere. Another `:base_header` was defined in a tighter scope of the spec and was only used in that scope, and `:req_with_body_headers` was never used anywhere.

#### Included In This Code
+ `X-Ops-Server-API-Version` sent to server in Chef requests.
+ Nice error formatting for unsupported API version in Knife.
+ Nice error handling for unsupported API version in formatters/*.rb.
+ Specs for my work and a few parts of the code that was missing coverage.

#### Example Knife Output
```
++> knife node list
ERROR: The version of Chef that Knife is using is not supported by the Chef server you sent this request to
This version of Chef requires a server API version of 100
The Chef server you sent the request to supports a min API verson of 0 and a max API version of 1
Please either update your Chef client or server to be a compatible set
```

#### Example chef-client output
```
# chef-client
Starting Chef Client, version 12.4.0.dev.0

Chef encountered an error attempting to load the node data for "charmander"

Incompatible server API version:
--------------------------------
This version of Chef is not supported by the Chef server you sent this request to
This version of Chef requires a server API version of 100
The Chef server you sent the request to supports a min API version of 0 and a max API version of 1
Please either update your Chef client or server to be a compatible set


Running handlers:
[2015-05-01T21:39:11+00:00] ERROR: Running exception handlers
Running handlers complete
[2015-05-01T21:39:11+00:00] ERROR: Exception handlers complete
[2015-05-01T21:39:11+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
Chef Client failed. 0 resources updated in 1.281854907 seconds
[2015-05-01T21:39:11+00:00] ERROR: 406 "Not Acceptable"
[2015-05-01T21:39:12+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

### TODO
- [x] Integration tests with server? (answer: none exist yet)
- [x] We do indeed want 0 as the current server version right?

### DIFF
--- a/spec/unit/knife_spec.rb
+++ b/spec/unit/knife_spec.rb
@@ -130,7 +130,8 @@ describe Chef::Knife do
                     "Accept-Encoding"=>"gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
                     'X-Chef-Version' => Chef::VERSION,
                     "Host"=>"api.opscode.piab",
-                    "X-REMOTE-REQUEST-ID"=>request_id}}
+                    "X-REMOTE-REQUEST-ID"=>request_id,
+                    'X-Ops-Server-API-Version' => Chef::HTTP::Authenticator::SERVER_API_VERSION}}
 
     let(:request_id) {"1234"}
 
--- a/spec/unit/knife_spec.rb
+++ b/spec/unit/knife_spec.rb
@@ -375,6 +375,16 @@ describe Chef::Knife do
       expect(stderr.string).to match(%r[Response: nothing to see here])
     end
 
+    it "formats 406s (non-supported API version error) nicely" do
+      response = Net::HTTPNotAcceptable.new("1.1", "406", "Not Acceptable")
+      response.instance_variable_set(:@read, true) # I hate you, net/http.
+      allow(response).to receive(:body).and_return(Chef::JSONCompat.to_json(:error => "sad trombone", :min_version => "0", :max_version => "1"))
+      allow(knife).to receive(:run).and_raise(Net::HTTPServerException.new("406 Not Acceptable", response))
+      knife.run_with_pretty_exceptions
+      expect(stderr.string).to match(%r[The version of Chef that Knife is using is not supported by the Chef server you sent this request to])
+      expect(stderr.string).to match(%r[This version of Chef requires a server API version of #{Chef::HTTP::Authenticator::SERVER_API_VERSION}])
+    end
+
     it "formats 500s nicely" do
       response = Net::HTTPInternalServerError.new("1.1", "500", "Internal Server Error")
       response.instance_variable_set(:@read, true) # I hate you, net/http.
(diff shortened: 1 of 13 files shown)


danielsdeleo:
@chef/client-core

tyler-ball:
I think storing it as a const makes sense because we don't want it to be configurable.  I'm surprised there isn't an `authenticator_spec.rb` - would it be hard to add one that has a few tests?

tylercloke:
Naw I can totally add more tests, and there is still work to do around handing 406s, I just wanted to push this up to make sure I was going in the right direction in general.

marcparadise:
:+1: LGTM , though I'd like to see us (at some point) factor up the common error inspection stuff. 

Re: 0 - that's the baseline version.  12.1.0RC will introduce version  1, which we should move to - but only after the necessary changes to client handling of clients and users are complete.

spec/unit/knife_spec.rb
```diff
@@ -374,6 +375,16 @@ class SuperAwesomeCommand < Chef::Knife
...
     end
 
+    it "formats 406s (non-supported API version error) nicely" do
+      response = Net::HTTPNotAcceptable.new("1.1", "406", "Not Acceptable")
+      response.instance_variable_set(:@read, true) # I hate you, net/http.
+      allow(response).to receive(:body).and_return(Chef::JSONCompat.to_json(:error => "sad trombone", :min_version => "0", :max_version => "1"))
+      allow(knife).to receive(:run).and_raise(Net::HTTPServerException.new("406 Not Acceptable", response))
+      knife.run_with_pretty_exceptions
+      expect(stderr.string).to match(%r[The version of Chef that Knife is using is not supported by the Chef server you sent this request to])
+      expect(stderr.string).to match(%r[This version of Chef requires a server API version of #{Chef::HTTP::Authenticator::SERVER_API_VERSION}])
```
danielsdeleo: When you don't need any regex features, it's better to do

```ruby
expect(actual_string).to include("expected part of string")
```

Just so you don't stub your toe on any special regex characters (or have to worry about escaping them).

danielsdeleo:
Minor rspec usage thing, but :+1: otherwise. 

As for the error handling stuff, probably some of it could be shared between knife and chef-client. It's tricky though because the messaging has to be tuned to say whether the validator key or regular client key are being used, etc. It'd be cool if knife did have the messaging about what config values could be at fault though.

PR #3320, opened 2015-04-30
Lcg/node utf8 sanitize

lamont-granquist:
This patch cleans up any bad utf8 data in the node.save so that node.save will never throw a UTF8 exception.

### DIFF
--- a/chef.gemspec
+++ b/chef.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
   s.add_dependency "mixlib-shellout", ">= 2.0.0.rc.0", "< 3.0"
   s.add_dependency "ohai", "~> 8.0"
 
-  s.add_dependency "ffi-yajl", ">= 1.2", "< 3.0"
+  s.add_dependency "ffi-yajl", "~> 2.2"
   s.add_dependency "net-ssh", "~> 2.6"
   s.add_dependency "net-ssh-multi", "~> 1.1"
   # CHEF-3027: The knife-cloud plugins require newer features from highline, core chef should not.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Convert bootstrap template to use sh #2877
 * [Issue #3316](https://github.com/chef/chef/issues/3316) Fix idempotency issues with the `windows_package` resource
 * [pr#3295](https://github.com/chef/chef/pull/3295): Stop mutating `new_resource.checksum` in file providers.  Fixes some ChecksumMismatch exceptions like [issue#3168](https://github.com/chef/chef/issues/3168)
+* [pr#3320] Sanitize non-UTF8 characters in the node data before doing node.save().  Works around many UTF8 exception issues reported on node.save().
 
 ## 12.3.0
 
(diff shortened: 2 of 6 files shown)


lamont-granquist:
@chef/client-core @chef/client-engineers this should stop all the node.save utf8 exceptions.

jdmundrawala:
how is bad utf8 data cleaned up? just ignored?

lamont-granquist:
via `str.encode!("utf-8", "binary", :undef => :replace)`

https://github.com/chef/ffi-yajl/blob/master/lib/ffi_yajl/encoder.rb#L44

lamont-granquist:
so it gets the funny <?> unicode symbol

jdmundrawala:
Is that safe?
Maybe it'd be safe to replace that with an object where we base64 encode the content?
something like
```
{ '__type__': 'binary', content: '...'}
```
However that would make decoding more complicated

danielsdeleo:
I think replacing is optimal, it at least remains human readable for anyone who cares to look at it, and realistically this is probably data the user doesn't care about, like some field in AD set to ISO-whateveritis encoding on a machine configured as UTF-8.

lamont-granquist:
define 'safe'?  its valid utf8 json that we render so it passes validation in the server, and gets stored in the database fine.  you lose data, but its just the bad characters.  when you view the fields later you'll get a '?' for a single bad character in a name (which is easy to diagnose and fix) or else you'll see a pile of garbage for binary that got rendered to utf8.  and at this level we don't know if its binary and should be hex encoded.  the use case i care more about is someone's GECOS name in europe with a latin-1 character that isn't valid utf8 and base64 encoding that is real silly.

if someone really cares about the binary data being rendered accurately then ohai should be fixed to base64 encode it.  this enables a good bug report on that since we'll get "hey, node['windows']['registry']['silliness'] is garbage, looks like its binary data being managled by utf8 can we fix that to base64 encode it?" rather than "why am i getting bad utf8 token ???a?r4???#$? at the end of my chef run?"

jdmundrawala:
That makes sense. I thought you were going after the windows registry binary thing.

chef.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
   s.add_dependency "mixlib-shellout", ">= 2.0.0.rc.0", "< 3.0"
   s.add_dependency "ohai", "~> 8.0"
 
-  s.add_dependency "ffi-yajl", ">= 1.2", "< 3.0"
```
tyler-ball: Aren't there a bunch of other places we could update this?  Off the top of my head there is the [chefdk](https://github.com/chef/chef-dk/blob/master/chef-dk.gemspec#L42).  It would be nice to have as few ffi-yajl versions in our omnibus packages as possible.

lamont-granquist: Well the other ones are fine with 1.2.x, when we install chef we should get 2.2 which will satisfy the ohai and chef-zero deps and that should be the only one we wind up with.  I don't think this will result in multiple versions being installed, and generally I don't like bumping lower version numbers until I have to in order to keep the compatibility matrix as wide as possible to prevent conflicts.

tyler-ball: :+1:

tyler-ball:
:+1: other than Travis

lamont-granquist:
Yeah, travis failures are whack tracking then down now..

lamont-granquist:
so @danielsdeleo @jdmundrawala are you :+1: ?

danielsdeleo:
Yep, :+1:

jdmundrawala:
:+1:

Issue #3153, opened 2015-03-27
chef-client 12.2.0 fails to start without HOME environment variable


dgivens:
Just upgraded my Debian Wheezy systems to 12.2.0 and now get the following when trying to start the chef-client service.

```
[production-dfw] root@api-n01a:~ # service chef-client start
Starting chef-client : chef-client/opt/chef/embedded/apps/chef/lib/chef/util/path_helper.rb:199:in `home': couldn't find HOME environment -- expanding `~' (ArgumentError)
        from /opt/chef/embedded/apps/chef/lib/chef/util/path_helper.rb:199:in `all_homes'
        from /opt/chef/embedded/apps/chef/lib/chef/util/path_helper.rb:161:in `home'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:36:in `gather_plist_dirs'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:40:in `<class:Macosx>'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:27:in `<class:Service>'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:26:in `<class:Provider>'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:25:in `<class:Chef>'
        from /opt/chef/embedded/apps/chef/lib/chef/provider/service/macosx.rb:24:in `<top (required)>'
        from /opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        from /opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        from /opt/chef/embedded/apps/chef/lib/chef/providers.rb:93:in `<top (required)>'
        from /opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        from /opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        from /opt/chef/embedded/apps/chef/lib/chef.rb:25:in `<top (required)>'
        from /opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
[...]

PR #3321, opened 2015-05-01
Add an integration test of chef-client with empty ENV.

ksubrama:
This is an integration test to catch the chef-12.2.0 regression that caused #3153 .  This was one of the identified corrective actions.

### DIFF
--- a/spec/integration/client/client_spec.rb
+++ b/spec/integration/client/client_spec.rb
@@ -47,6 +47,8 @@ describe "chef-client" do
   # cf. CHEF-4914
   let(:chef_client) { "ruby '#{chef_dir}/chef-client' --minimal-ohai" }
 
+  let(:critical_env_vars) { %w(PATH RUBYOPT BUNDLE_GEMFILE).map {|o| "#{o}=#{ENV[o]}"} .join(' ') }
+
   when_the_repository "has a cookbook with a no-op recipe" do
     before { file 'cookbooks/x/recipes/default.rb', '' }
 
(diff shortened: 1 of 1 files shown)


ksubrama:
/cc @btm @jdm @smurawski 

Can one of you /cc chef-client-core as well.

jdmundrawala:
awesome. Just a heads up, I don't think this test will run in travis until it is merged.
:+1: 
cc @chef/client-core

spec/integration/client/client_spec.rb
```diff
@@ -60,6 +60,23 @@ def stop_tiny_server
...
+
+      critical_env_vars = ""
+      %w(PATH RUBYOPT BUNDLE_GEMFILE).each do |option|
+        critical_env_vars += "#{option}=#{ENV[option]} "
+      end
+
+      Chef::Log.info "Bare invocation will have the following load-path."
+      Chef::Log.info shell_out("env -i #{critical_env_vars} ruby -e 'puts $:'").stdout
+      result = shell_out("env -i #{critical_env_vars} #{chef_client} -c \"#{path_to('config/client.rb')}\" -o 'x::default'", :cwd => chef_dir)
+      result.error!
```
thommay: I'd really like to see an expectation here. (And through the rest of this file - there are isolated tests that use expectations, but it's not consistent.)

jdm:
Please remember to cc @jdmundrawala and not me :)

ksubrama:
Dang it!  I was hoping autocomplete would save me but apparently it likes Mozilla.  @jdmundrawala .

jdmundrawala:
@jdm this seems to happen a lot. Is it possible to ignore this project. According to the button, unwatch means you still got at-notifications. Ignored says that you will never get notified.

spec/integration/client/client_spec.rb
ksubrama: I think most of them use the implicit test of error! because there's really not much else to verify other than "did this command succeed".  If it didn't, you'd want to see the error log printed (which error! does).  If it did, running a second expect on the return code is redundant because error! already does that.

ksubrama:
I made one change to restrict it to !windows.  env -i is probably not going to work there :-P.

spec/integration/client/client_spec.rb
```diff
@@ -56,7 +58,18 @@ def stop_tiny_server
...
+    end
+
+    it "should complete successfully with no other environment variables", :skip => (Chef::Platform.windows?) do
+      file 'config/client.rb', <<EOM
+local_mode true
+cookbook_path "#{path_to('cookbooks')}"
+EOM
+
+      Chef::Log.info "Bare invocation will have the following load-path."
+      Chef::Log.info shell_out!("env -i #{critical_env_vars} ruby -e 'puts $:'").stdout
```
danielsdeleo: Do we still need these log lines?

ksubrama: Not necessarily - but if that test fails (on some CI environment or elsewhere), I would have no idea how to debug it without either jumping on the box and figuring out the environment or looking at that.  I figured I'd just help my future self by leaving the command to dump the lookup path in there.

danielsdeleo: In that case, I'd do:

```ruby
begin
  run_test_code
rescue
  debugging_statements
  raise
end
```

ksubrama: Done.  Thanks!

danielsdeleo:
Awesome, thanks for doing this! :100: :+1: :ship: :icecream:

PR #3324, opened 2015-05-01
Able to use custom ssl certificates when uploading to Supermarket

nellshamrell:
More and more users are spinning up private versions of the Supermarket community cookbooks sites.  Private Supermarkets use a self signed ssl certificate.  Though this is easy to pull down onto a local workstation through knife ssl fetch, it's currently not possible to tell Chef to use that certificate when trying to upload to the private Supermarket.

This pull request will allow a user to define an ssl_cert_file to use in their knife.rb file like this:
```bash
  ssl_cert_file '/root/.chef/trusted_certs/nell-supermarket-server.crt'
```
Then, when the user tries to upload a cookbook, Chef will use this ssl_cert_file to authenticate to the Private Chef Server.

### DIFF
--- a/lib/chef/cookbook_site_streaming_uploader.rb
+++ b/lib/chef/cookbook_site_streaming_uploader.rb
@@ -146,6 +146,13 @@ class Chef
           http.use_ssl = true
           http.verify_mode = verify_mode
         end
+
+        if Chef::Config[:ssl_cert_file]
+          http.cert_store = OpenSSL::X509::Store.new
+          http.cert_store.set_default_paths
+          http.cert_store.add_file(Chef::Config[:ssl_cert_file])
+        end
+
         res = http.request(req)
         #res = http.start {|http_proc| http_proc.request(req) }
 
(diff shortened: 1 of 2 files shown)


afiune:
@nellshamrell Wondering if there is a way to don't specify the exact name/path of the cert? I think this should work the same way we use other certs. (using `trusted_certs_dir` instead) 

But I could be wrong.. :smile: Thoughts?

cc/ @schisamo @cwebberOps @christophermaier

nellshamrell:
I tried using the trusted_certs_dir in my knife.rb, but unfortunately it still fails at the same place.  It's not finding and submitting the certificate.

nellshamrell:
Actually...trying something out.  Will update momentarily.

nellshamrell:
Still having no luck with using trusted_certs_dir

irvingpop:
After doing some digging it's pretty clear to me that `cookbook_site_streaming_uploader ` should use `Chef::HTTP::BasicClient` instead of calling `Net::HTTP` directly.   That way it should automatically get SSL policies,  proxy support, etc.

nellshamrell:
That would explain the massive issues we're seeing.  Taking a look.

nellshamrell:
Closing this, as it is taken care of in #3333

PR #3325, opened 2015-05-02
Enforce passing a node name with validatorless bootstrapping

### LINKED ISSUE #3204: `knife bootstrap` fails if `-N` option is not passed
The client_builder uses the value of the `-N` option as the node name:

https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap/client_builder.rb#L79-L81

if that is nil since the user didn't pass it, the user will receive a confusing HTTP 405 Method Not Allowed Error here:

https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap/client_builder.rb#L157

This command should use the first argument of the bootstrap command as the node name if `-N` was not passed.

ryancragun:
When bootstrapping with chef-vault or with user credentials we create
the node and client on the Chef Server before we bootstrap the node. As
it is possible to specify a server address that is not the node name we
need to enforce that the user pass a node name with -N.

At present if you omit the the node name you'll be prompted to delete
nonexistent clients and nodes.

cc @lamont-granquist 

closes #3204

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -316,6 +316,12 @@ class Chef
         # new client-side hawtness, just delete your validation key.
         if chef_vault_handler.doing_chef_vault? || 
             (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
+
+          unless config[:chef_node_name]
+            ui.error("You must pass a node name with -N when bootstrapping with user credentials")
+            exit 1
+          end
+
           client_builder.run
 
           chef_vault_handler.run(node_name: config[:chef_node_name])
--- a/spec/unit/knife/bootstrap_spec.rb
+++ b/spec/unit/knife/bootstrap_spec.rb
@@ -531,6 +531,7 @@ describe Chef::Knife::Bootstrap do
   describe "when running the bootstrap" do
     let(:knife_ssh) do
       knife.name_args = ["foo.example.com"]
+      knife.config[:chef_node_name] = "foo.example.com"
       knife.config[:ssh_user]      = "rooty"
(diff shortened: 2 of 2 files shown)


stevendanna:
I believe this solves: https://github.com/chef/chef/issues/3204

It is a little annoying that -N has become a required option.  An alternative would be to grab the fqdn from the node and use that by default.

ryancragun:
@stevendanna I thought about doing that, but right now we do the node and client creation prior to actually bootstrapping the node.  We don't know the FQDN for sure until we SSH into the node to run the bootstrap.  If we wanted to keep the order of operations we'd probably end up doing two separate SSH commands: first to find the FQDN and second to actually bootstrap the node.  If we think it's worth it I'd be happy to do that.

lamont-granquist:
And on windows you'd have to winrm into the node and do something.

I thought about that, but its a bunch of work, and I'd rather do this, and then let a feature request drive doing the work to try to guess the node name by using the transport mechanism into the node.

lamont-granquist:
:+1:

lamont-granquist:
we also still need this check anyway if the ssh fails to find a fqdn.  i'm also worried about edge cases like 'localhost.localdomain'.

stevendanna:
:+1: as this is definitely improvement.  We should update any documentation and banners to indicate that the -N option isn't optional.

stevendanna:
Re making 2 SSH connections to get the hostname, I wouldn't be too concerned about that as we can just keep the SSH connection around; however, I'm not sure if that is possible in the windows case, which I hadn't considered.

lamont-granquist:
Yeah technically this doesn't support windows yet at all, I think knife-windows needs a template update

btm:
knife-windows support for validatorless bootstrapping is being worked in in https://github.com/chef/knife-windows/pull/224.

PR #3327, opened 2015-05-02
Replace unreliable service group parsing mechanism.

### LINKED ISSUE #3248: AIX invalid test for group of subsystems in Chef::Provider::Service::Aix
This is not a valid test of whether or not a service is part of a subsystem group in chef/provider/service/aix.rb

Lines 96 - 97:

    services = shell_out!("lssrc -a | grep -w #{@new_resource.service_name}").stdout.split("\n")
    is_resource_group?(services)

Then lines 115 - 118:

    def is_resource_group? (services)
      if services.length > 1
        Chef::Log.debug("#{@new_resource.service_name} is a group")
        @is_resource_group = true

Say there are two subsystems, neither part of a process group:

chef
chef-client

The logic in Chef::Provider::Service::Aix will detect "chef" as being part of a subsystem group because the grep returns more than one line. This could be false. The `grep -w` does not work because the `-` in the subsystem name.

    (echo "blah"; echo "blah-bad") | grep -w blah

Probably need to remove the grep and use some ruby to perform the comparison.

juliandunn:
This is to fix #3248.

### DIFF
--- a/lib/chef/provider/service/aix.rb
+++ b/lib/chef/provider/service/aix.rb
@@ -91,15 +91,18 @@ class Chef
 
         protected
         def determine_current_status!
-          Chef::Log.debug "#{@new_resource} using lssrc to check the status "
+          Chef::Log.debug "#{@new_resource} using lssrc to check the status"
           begin
-            services = shell_out!("lssrc -a | grep -w #{@new_resource.service_name}").stdout.split("\n")
-            is_resource_group?(services)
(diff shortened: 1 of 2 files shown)


juliandunn:
Can I get a review on this @lamont-granquist ?

lamont-granquist:
I'll have to trust you on the AIXisms, but :+1:

PR #3329, opened 2015-05-03
Use target, not name, if it is specified.

### LINKED ISSUE #3250: ifconfig resource does not respect 'target' option properly
It looks like 'target' properly is not respected in resource 'interface'.

While documentatoin says :
> target 	The IP address that will be assigned to the network interface. Default value: the name of the resource block. (See “Syntax” section above for more information.)

The code and behaviour looks like the variables are used separately. 

for example resource like :
````ruby
  ifconfig 'test-name' do
    device 'lo:1'
    target '1.2.3.4'
    mask '255.255.255.255'
    action [:add, :enable]
  end
````
will throw error :
````
           Error executing action `add` on resource 'ifconfig[test-name]'
           
           Chef::Exceptions::Exec
           ----------------------
           ifconfig lo:1 test-name netmask 255.255.255.255 returned 1, expected 0
           
           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/XXX/recipes/default.rb
           
           147:   ifconfig 'test-name' do
           148:     device 'lo:1'
           149:     target '1.2.3.4'
           150:     mask '255.255.255.255'
           151:     action [:add, :enable]
           152:   end
           153: 
````

juliandunn:
Fixes #3250

### DIFF
--- a/lib/chef/provider/ifconfig.rb
+++ b/lib/chef/provider/ifconfig.rb
@@ -194,7 +194,7 @@ class Chef
 
       private
       def add_command
-        command = "ifconfig #{@new_resource.device} #{@new_resource.name}"
+        command = "ifconfig #{@new_resource.device} #{@new_resource.target}"
         command << " netmask #{@new_resource.mask}" if @new_resource.mask
         command << " metric #{@new_resource.metric}" if @new_resource.metric
         command << " mtu #{@new_resource.mtu}" if @new_resource.mtu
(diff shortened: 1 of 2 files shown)


juliandunn:
Review? @chef/client-maintainers

btm:
change looks fine, add two unit tests?

danielsdeleo:
:+1: pending regression test(s)

sethvargo:
Seems like we should have a test for this, but :+1:

lamont-granquist:
:+1:

martinisoft:
Agreed on adding some unit tests to this before merging.

PR #3335, opened 2015-05-05
let the ruby patchlevels float

lamont-granquist:
test latest 2.0.x, 2.1.x, 2.2.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,14 @@ env:
 
 matrix:
   include:
-  - rvm: 2.0.0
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core @chef/client-engineers i don't think anyone is actively bumping these so we should probably just let travis bump them automatically.

jdmundrawala:
go for it

danielsdeleo:
:+1:

tyler-ball:
:+1:

PR #3336, opened 2015-05-05
remote_file support for windows network shares

jdmundrawala:
```ruby
remote_file 'C:\Foo.tar.gz' do
  source "\\\\foohost\\fooshare\\Foo.tar.gz"
end
```

cc @btm @ksubrama @smurawski

### DIFF
--- a/lib/chef/provider/remote_file/fetcher.rb
+++ b/lib/chef/provider/remote_file/fetcher.rb
@@ -23,15 +23,29 @@ class Chef
       class Fetcher
 
         def self.for_resource(uri, new_resource, current_resource)
-          case uri.scheme
-          when "http", "https"
-            Chef::Provider::RemoteFile::HTTP.new(uri, new_resource, current_resource)
-          when "ftp"
-            Chef::Provider::RemoteFile::FTP.new(uri, new_resource, current_resource)
-          when "file"
-            Chef::Provider::RemoteFile::LocalFile.new(uri, new_resource, current_resource)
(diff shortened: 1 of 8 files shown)


lib/chef/provider/remote_file/fetcher.rb
```diff
@@ -23,15 +23,29 @@ class RemoteFile
...
+              raise ArgumentError, "Invalid uri, Only http(s), ftp, and file are currently supported"
+            end
+          end
+        end
+
+        # Windows network share: \\computername\share\file
+        def self.network_share?(source)
+          case source
+          when String
+            !!(%r{\A\\\\[A-Za-z][A-Za-z0-9+\-\.]*} =~ source)
```
ksubrama: I believe \\192.168.x.x is a valid way to reach a share (you might want to double check that).  In that case, we should allow numbers in the first character.

jdmundrawala: Do we have to support that? Is it different than '\\192.168.x.x'?
Also, is `\fooshare` valid?

ksubrama: Sorry - I meant \\IPaddress

https://msdn.microsoft.com/en-us/library/gg465305.aspx

jdmundrawala: oh i see

ksubrama:
:+1:

smurawski:
:+1:

Issue #1687, opened 2014-07-24
knife commands in Windows can have unexpected results because of argument globbing


juliandunn:
I ported this over from [CHEF-2486](https://tickets.opscode.com/browse/CHEF-2486)

This is enough of a delight blocker issue that I felt we should move it here. The real fix is probably in mingw and/or conversation with MSFT about the purported MSVC bug, but we would need someone from engineering to chase that down.

The problem is the quotes around the run_list are not preventing globbing of file/directory names in the current working directory.

For example, if your current working directory is a chef-repo with a subdirectory named `roles` then setting a node's run_list to `"role[webserver]"` acts as a glob which matches the name of the `roles` subdirectory.

Reference: http://sourceforge.net/p/mingw/bugs/1621/

There are a few workarounds documented in the ticket, but they are non-obvious and users will get frustrated long before they find them.

When using Powershell you have to use triple single quotes:

```
knife node run_list set test-node '''role[ssssssomething]'''
```

When using the Command prompt you have to use single quotes:

```
knife node run_list set test-node 'role[ssssssomething]'
```

jeremiahsnapp:
Here's the workaround from CHEF-2486

When using Poweshell you have to use single quotes and escape them with backticks:

```
knife node run_list set test-node `'role[ssssssomething]`'
```

When using the Command prompt you have to use single quotes:

```
knife node run_list set test-node 'role[ssssssomething]'
```

jeremiahsnapp:
The problem is the arguments of the knife command are seen by the shell (Powershell or command prompt) as a [string with wildcard characters](https://msdn.microsoft.com/en-us/library/aa717088%28v=vs.85%29.aspx). So the shell does its job and finds files that match the pattern. This is often referred to as [globbing](http://en.wikipedia.org/wiki/Glob_%28programming%29).

The typical way to prevent globbing is to surround the string with single quotes to prevent the shell from interpreting the string as a wildcard pattern. This works when using knife in the command prompt but it does not work when using knife in Powershell.

It seems that, for some reason, when using knife in Powershell the single quotes get eaten by knife leaving the bare string available for interpretation as a pattern internally. **The way to prevent this in Powershell is to use triple single quotes.**

Do the following to reproduce the problem and show the solution.

Create a few directories that will match the wildcard pattern `recipe[iptables]`

```
mkdir recipes
mkdir recipep
mkdir recipel
```

Run the following in the command prompt.

```
[...]

PR #3339, opened 2015-05-06
Powershell command wrappers to make argument passing to knife/chef-client etc. easier.

ksubrama:
Make argument parsing into ruby/chef binaries easier.


ksubrama:
/cc @btm @jdmundrawala 

Please cc chef-client-core.

smurawski:
It doesn't appear to actually wrap the commands to fix the issue https://github.com/chef/chef/issues/1687

ksubrama:
Ok, I think I have another version which handles that case.  It doesn't handle pipeline objects yet.  I'll do that in a second pass.

jdmundrawala:
This looks fine to me, however we might want to consider automatically generating these for all ruby bins. See https://github.com/chef/omnibus-chef/blob/master/config/software/shebang-cleanup.rb
:+1:

ksubrama:
Yes that'd definitely be optimal.  I went with only these initially to be conservative and not accidentally affect the semantics of a whole bunch of binaries that might otherwise show up on the path such as with ChefDK.

smurawski:
:-1:   trying any commands with more than one subcommand fails.  It's creating one string from the remaining args.
```powershell
Chef-PS chef-repo> knife node list
FATAL: Cannot find sub command for: '"node list"'
```

ksubrama:
How are you running this?  I run knife node list and it works for me.

ksubrama:
Apparently PS 5 changed some of the parsing behavior I was "working around".  Will work around the work around.

ksubrama:
@smurawski This should be ready for another look.

smurawski:
:+1: With limited testing on Server 2008 R2, Server 2012, and Server 2012 R2 this seems to do it and it appears to address #1687 Yay!

PR #3340, opened 2015-05-06
Allow inspection of event dispatch's subscribers

danielsdeleo:
I need this so I can properly assert that the `ChefRunner` class in chef-dk has correctly configured a 'doc' formatter for its embedded chef-client run. Currently I have to use `instance_variable_get` as a workaround :hankey: 

/cc @chef/client-core


jdmundrawala:
:+1:

tyler-ball:
:+1: You also get :star: for specs

PR #3342, opened 2015-05-07
Check if proxy env_var is empty

jonsmorrow:
We need to check if the env variable is set to empty string. If we don't we can get in an edge case where we blow up trying to call URI.parse.

### DIFF
--- a/lib/chef/http/basic_client.rb
+++ b/lib/chef/http/basic_client.rb
@@ -101,12 +101,16 @@ class Chef
                 env["#{url.scheme.upcase}_PROXY"] || env["#{url.scheme}_proxy"]
 
         # Check if the proxy string contains a scheme. If not, add the url's scheme to the
-        # proxy before parsing. The regex /^.*:\/\// matches, for example, http://.
-        proxy = if proxy.match(/^.*:\/\//)
-          URI.parse(proxy)
-        else
-          URI.parse("#{url.scheme}://#{proxy}")
-        end if String === proxy
+        # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
+        # here since we are really just trying to get the string built correctly.
+        if String === proxy && !proxy.strip.empty?
+          if proxy.match(/^.*:\/\//)
+           proxy = URI.parse(proxy.strip)
+          else
+           proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
+          end 
+        end
+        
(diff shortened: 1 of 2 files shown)


oferrigni:
LGTM

tyler-ball:
Paging @chef/client-maintainers

lib/chef/http/basic_client.rb
```diff
@@ -101,12 +101,16 @@ def proxy_uri
...
         # Check if the proxy string contains a scheme. If not, add the url's scheme to the
-        # proxy before parsing. The regex /^.*:\/\// matches, for example, http://.
-        proxy = if proxy.match(/^.*:\/\//)
-          URI.parse(proxy)
-        else
-          URI.parse("#{url.scheme}://#{proxy}")
-        end if String === proxy
+        # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
+        # here since we are really just trying to get the string built correctly.
+        if String === proxy && !proxy.strip.empty?
```
tyler-ball: I don't think the `String === proxy` check is necessary

jaymzh:
tests are failing... but other than that I'm :+1:

lib/chef/http/basic_client.rb
danielsdeleo: You would need it if someone set `Chef::Config["https_proxy"]` to an `URI` object. Not sure if that happens in the real world or not.

jonsmorrow:
Looking at the tests it was such a small change i admit not testing locally :(

danielsdeleo:
Failing test is an issue with the kitchen tests, not sure what's going on there but it's not related to this change, so :+1:

danielsdeleo:
That said, a regression test would be great.

jonsmorrow:
Interesting no failures locally. Will look into a regression test.

danielsdeleo:
@jonsmorrow looks like the failing test is some intermittent kitchen-ec2 problem that was "fixed" by re-running.

jdmundrawala:
Trying to track it down: https://github.com/chef/chef/pull/3346

martinisoft:
Seems to be passing now with a new regression test in place. Thank you @jonsmorrow :heart:

danielsdeleo:
Awesome, :+1:

tyler-ball:
:+1:

jonsmorrow:
ok merging.

PR #3344, opened 2015-05-07
Changing Net User things to use ffi instead of win32-api

jdmundrawala:
This was a mess of memory corruption, so I've rewritten using the saner ffi approach. It now works with 64-bit rubies.

cc @chef/client-windows @adamedx @ksubrama


lib/chef/win32/api/net.rb
```diff
@@ -67,6 +83,52 @@ class USER_INFO_3 < FFI::Struct
...
             :usri3_password_expired, :DWORD
+
+          def set(key, val)
+            val = if val.is_a? String
+              encoded = if val.encoding == Encoding::UTF_16LE
+                val
+              else
+                val.to_wstring
+              end
+              val = FFI::MemoryPointer.from_string(encoded)
```
ksubrama: If you're setting val =  directly here, you don't need the val = on line 88 or the else block below.

lib/chef/win32/api/net.rb
```diff
@@ -67,6 +83,52 @@ class USER_INFO_3 < FFI::Struct
...
+                  val.read_wstring
+                end
+              else
+                val
+              end
+            end
+          end
+
+          def usri3_logon_hours
+            self[:usri3_logon_hours].read_bytes(21)
```
ksubrama: Does this work if the field is null?

lib/chef/win32/net.rb
```diff
@@ -0,0 +1,216 @@
...
+            usri3_home_dir_drive: nil,
+            usri3_password_expired: 0
+          }.each do |(k,v)|
+            s.set(k, v)
+          end
+        end
+      end
+
+      def self.net_api_error!(code)
+        msg = case code
```
ksubrama: Shouldn't all of these already already be handled by Win32::Error.get_last_error?

lib/chef/win32/net.rb
```diff
@@ -0,0 +1,216 @@
...
+      def self.net_user_get_info_l3(server_name, user_name)
+        server_name = wstring(server_name)
+        user_name = wstring(user_name)
+
+        ui3_p = FFI::MemoryPointer.new(:pointer)
+
+        rc = NetUserGetInfo(server_name, user_name, 3, ui3_p)
+
+        if rc != NERR_Success
+          if Chef::ReservedNames::Win32::Error.get_last_error != 0
```
ksubrama: Did you run into a situation where this wasn't getting set even though rc != 0?  If that's the case, we shouldn't trust get_last_error at all and modify net_api_error! to do the same work Error.raise! but with the the code manually supplied.  We shouldn't have to supply our own error strings.

lib/chef/win32/net.rb
```diff
@@ -0,0 +1,216 @@
...
+        formatted_message << "---- Begin Win32 API output ----\n"
+        formatted_message << "Net Api Error Code: #{code}\n"
+        formatted_message << "Net Api Error Message: #{msg}\n"
+        formatted_message << "---- End Win32 API output ----\n"
+
+        raise Chef::Exceptions::Win32APIError, msg + "\n" + formatted_message
+      end
+
+      def self.net_user_add_l3(server_name, args)
+        param_err = FFI::Buffer.new(:long)
```
ksubrama: Can use nil for the error parameter since we don't actually use the returned index value anywhere.

lib/chef/win32/net.rb
jdmundrawala: nope, it does not set any thread local variable

lib/chef/win32/net.rb
jdmundrawala: In general, last error is not set. It probably only gets set for ERROR_ACCESS_DENIED. As for not supplying the string, I dont really know how else to get a reasonable message. I'm open to suggestions

ksubrama: Oh interesting.  I hadn't considered that LastError gets set conditionally.  If the return code is always non-zero whenever LastError does get set, then I would recommend refactoring/changing raise! in https://github.com/chef/chef/blob/master/lib/chef/win32/error.rb to instead take the error code you have in rc and do what its currently doing.  I'm newb at ruby - so I'll let you figure out if it's better to make a new method that takes the code or add an optional argument or something of that sort.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms681386(v=vs.85).aspx seems to indicate that the existing code in Error.format_message should still work for these error codes.  This way, you might not even have to specifically declare them.  This would let you remove all the (if get_last_error != 0) log everywhere and simply raise on the return code you have and will also handle the odd/weird situation where the documentation is incomplete like for https://msdn.microsoft.com/en-us/library/windows/desktop/aa370649%28v=vs.85%29.aspx where the possible return codes don't include ERROR_INVALID_PARAMETER but the documentation for parm_err mentions a situation where that might occur.

ksubrama: This is all assuming that thing work the way I think they work...  if reality doesn't match, feel free to ignore me.

ksubrama: https://support.microsoft.com/en-us/kb/256348 has a snippet on how to fetch the error messages from a netmsg.dll.

ksubrama:
:+1: Loading a library to fetch error messages is too insane for this pull-request.  Leaving that for another time.  The rest of this looks awesome!

lib/chef/util/windows/net_user.rb
```diff
@@ -120,49 +91,32 @@ def initialize(username)
     @name = multi_to_wide(username)
   end
 
-  LOGON32_PROVIDER_DEFAULT = 0
-  LOGON32_LOGON_NETWORK = 3
+  LOGON32_PROVIDER_DEFAULT = Security::LOGON32_PROVIDER_DEFAULT
+  LOGON32_LOGON_NETWORK = Security::LOGON32_LOGON_NETWORK
   #XXX for an extra painful alternative, see: http://support.microsoft.com/kb/180548
   def validate_credentials(passwd)
```
btm: in retrospect, this should have been something like ```valid_credentials?``` but wutevs. #breakfastwhenevs

lib/chef/win32/api.rb
```diff
@@ -117,6 +117,7 @@ def self.extended(host)
         host.typedef :uint32,  :LCID # Locale identifier. For more information, see Locales.
         host.typedef :uint32,  :LCTYPE # Locale information type. For a list, see Locale Information Constants.
         host.typedef :uint32,  :LGRPID # Language group identifier. For a list, see EnumLanguageGroupLocales.
+        host.typedef :pointer, :LMSTR
```
btm: "A LMSTR is a 32-bit pointer to a null-terminated string of 16-bit Unicode characters."

jdmundrawala: We agreed to listen to the typedef. really just a pointer

btm: I guess I meant to imply a comment would be nice, because this :hankey: is confusing to me.

jdmundrawala: oooh, i see

btm:
:+1: nice work man. it's a big change, so let's hope we don't have any regressions where we failed to have tests before.

the functional tests pass on my 8.1 workstation too :rocket:.

smurawski:
:+1:  Looks good and I can make tests pass.  Squash before merge?

jdmundrawala:
I think I'm ok with the commits (except the last one)... they tell the story I want them to and each one should be working point in time

smurawski:
cool beans

PR #3346, opened 2015-05-07
Diagnose failing tests

jdmundrawala:


danielsdeleo:
Yes please. :+1:

.travis.yml
```diff
@@ -45,6 +45,8 @@ matrix:
     script:
 # FIXME: we should fix centos-6 against AWS and then enable it here
     - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen test ubuntu; fi
+    after_failure:
+    - cat .kitchen/logs/kitchen.log
```
tyler-ball: There are also instance-specific logs - you'll want to cat those as well.  They're in the same directory, just named `<instance>.log`

tyler-ball: http://serverfault.com/questions/172284/linux-cat-with-separators-among-files

PR #3348, opened 2015-05-07
fixes the timing on the chef-shell specs

lamont-granquist:
also makes timing errors hard errors

### DIFF
--- a/spec/functional/shell_spec.rb
+++ b/spec/functional/shell_spec.rb
@@ -29,6 +29,8 @@ describe Shell do
   describe "smoke tests", :unix_only => true do
     include Chef::Mixin::Command::Unix
 
+    TIMEOUT=30
+
     def read_until(io, expected_value)
       start = Time.new
       buffer = ""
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core should knock a minute off of the current spec runs.

danielsdeleo:
It'd be a good idea to do an ad-hoc build just to be sure this won't blow up on Solaris or whatever.

tyler-ball:
:+1: other than fails

danielsdeleo:
The errors are weird, looks like we broke knife bootstrap, but that's not related to the change here. :+1: on this patch.

lamont-granquist:
Yeah master is broken due to #3365 at least.  I don't /think/ the test failures are related, but i'm hesitant to merge to master and break it worse right now.

I did actually run this through an ad-hoc, though, and it passed modulo two cosmic rays after i fixed the jankiness around draining stdout.

jdmundrawala:
yes please :+1:

PR #3354, opened 2015-05-08
Log cause chain for exceptions

jaym:
Thoughts?
cc @chef/client-maintainers

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -352,6 +352,13 @@ class Chef
     class << self
       def debug_stacktrace(e)
         message = "#{e.class}: #{e}\n#{e.backtrace.join("\n")}"
+
+        cause = e.cause if e.respond_to?(:cause)
+        while cause != nil
+          message << "\n\n>>>> Caused by #{cause.class}: #{cause}\n#{cause.backtrace.join("\n")}"
+          cause = cause.respond_to?(:cause) ? cause.cause : nil
+        end
+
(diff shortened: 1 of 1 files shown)


lamont-granquist:
We still ship and support ruby 2.0 don't we so you'd need to `respond_to?(:cause)` check this or something?

danielsdeleo:
I think Lamont is correct. :+1: once that's fixed up

jdmundrawala:
:(
but i wanted it for windows. I'm guessing there's no way for ruby 2.0?

danielsdeleo:
@jdmundrawala it's way harder, so depends on how motivated you are. Anyway, you can roll your own nested exceptions with something like:

```ruby

class SomeException
  attr_reader :cause
  def initialize(*args)
    @cause = $!
    super
  end
end
```

You could extract that into a module and then use module prepending to add it to all of our exceptions.

But even then we can only do this for Chef's own exceptions, so stuff from inside libraries won't have this.

tyler-ball:
@jdmundrawala are we still building chef with Ruby 2.0 for Windows?  We use 2.1.5 for the ChefDK

jdmundrawala:
still 2.0

tyler-ball:
:cry:

thommay:
:+1: LGTM

lamont-granquist:
yeah now that we're off of 2.0 everywhere we should :shipit:

Issue #3343, opened 2015-05-07
No error output on windows node


z0c:
I keep getting the following message whenever a error is raised on my cookbook: 
```
==> default: [2015-05-07T13:01:04+01:00] INFO: Running queued delayed notifications before re-raising exception
==> default:
==> default: [2015-05-07T13:01:04+01:00] ERROR: Running exception handlers
==> default: [2015-05-07T13:01:04+01:00] ERROR: Exception handlers complete
==> default: [2015-05-07T13:01:04+01:00] FATAL: Stacktrace dumped to C:/var/chef/cache/chef-stacktrace.out
==> default: [2015-05-07T13:01:04+01:00] FATAL: ArgumentError: too short control escape
Chef never successfully completed! Any errors should be visible in the
output above. Please fix your recipes so that they properly complete.
```

Seems that Chef solo is not handling the backslashes on physical paths on a windows node

Chef client version is 12.3.0 and the node is running Windows 2012 R2.

Here is the full stack trace:
```
Generated at 2015-05-07 13:01:04 +0100
ArgumentError: too short control escape
C:/opscode/chef/embedded/apps/chef/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb:66:in `recipe_snippet'
C:/opscode/chef/embedded/apps/chef/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb:43:in `add_explanation'
C:/opscode/chef/embedded/apps/chef/lib/chef/formatters/error_mapper.rb:72:in `resource_failed'
C:/opscode/chef/embedded/apps/chef/lib/chef/formatters/base.rb:139:in `resource_failed'
C:/opscode/chef/embedded/apps/chef/lib/chef/event_dispatch/dispatcher.rb:29:in `block (2 levels) in def_forwarding_method'
C:/opscode/chef/embedded/apps/chef/lib/chef/event_dispatch/dispatcher.rb:29:in `each'
C:/opscode/chef/embedded/apps/chef/lib/chef/event_dispatch/dispatcher.rb:29:in `block in def_forwarding_method'
C:/opscode/chef/embedded/apps/chef/lib/chef/resource.rb:574:in `rescue in run_action'
C:/opscode/chef/embedded/apps/chef/lib/chef/resource.rb:583:in `run_action'
C:/opscode/chef/embedded/apps/chef/lib/chef/runner.rb:49:in `run_action'
[...]

PR #3357, opened 2015-05-08
Escape string inside regex

### LINKED ISSUE #3055: Failure to print resource error message on Windows
### Version:
Chef-Client 12.1.0

### Environment:
Windows Server 2012 R2

### Scenario:
Execute a chef-solo run & view the chef log

### Steps to Reproduce:
Add an resource with a forced error, where the cookbooks are in a subdirectory starting with "p" like "C:\Windows\Temp\packer\cookbooks"

### Expected Result:
Get the nice display listing the source code of the resource & its values

### Actual Result:
C:/opscode/chef/embedded/apps/chef/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb:66: warning: invalid Unicode Property \p

What I think is happening is that the "file" variable in the below line has Windows back-slashes.  Hence when used as a regex direct (without other processing or escaping) parts will be interpreted as special escape codes & so block the desired functionality.  Not sure but would hope there's an easy way in ruby to regex-escape an arbitrary string.

https://github.com/chef/chef/blob/master/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb#L66

jdmundrawala:
Should fix #3055
The unit tests repros the issue, but does not fail. Ruby seems to be logging a warning somewhere internally

### DIFF
--- a/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb
+++ b/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb
@@ -63,7 +63,7 @@ class Chef
         def recipe_snippet
           return nil if dynamic_resource?
           @snippet ||= begin
-            if file = resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1] and line = resource.source_line[/^#{Regexp.escape(file)}:([\d]+)/,1].to_i
+            if file = parse_source and line = parse_line(file)
               return nil unless ::File.exists?(file)
               lines = IO.readlines(file)
 
@@ -110,6 +110,16 @@ class Chef
           line_nr_string = (line_nr + 1).to_s.rjust(3) + ": "
           line_nr_string + line
         end
+
+        def parse_source
+          resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1]
+        end
+
+        def parse_line(source)
+          resource.source_line[/^#{Regexp.escape(source)}:([\d]+)/,1].to_i
+        end
+
+
 
       end
     end
--- a/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb
+++ b/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb
@@ -63,7 +63,7 @@ class Chef
         def recipe_snippet
           return nil if dynamic_resource?
           @snippet ||= begin
-            if file = resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1] and line = resource.source_line[/^#{file}:([\d]+)/,1].to_i
+            if file = resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1] and line = resource.source_line[/^#{Regexp.escape(file)}:([\d]+)/,1].to_i
               return nil unless ::File.exists?(file)
               lines = IO.readlines(file)
 
--- a/spec/unit/formatters/error_inspectors/resource_failure_inspector_spec.rb
+++ b/spec/unit/formatters/error_inspectors/resource_failure_inspector_spec.rb
@@ -126,6 +126,13 @@ describe Chef::Formatters::ErrorInspectors::ResourceFailureInspector do
         expect(@inspector.recipe_snippet).to match(/^# In C:\/Users\/btm/)
       end
 
+      it "parses a Windows path" do
+        source_line = "C:\\Windows\\Temp\\packer\\cookbooks\\fake_file.rb:2: undefined local variable or method `non_existent' for main:Object (NameError)"
+        @resource.source_line = source_line
+        @inspector = Chef::Formatters::ErrorInspectors::ResourceFailureInspector.new(@resource, :create, @exception)
+        expect(@inspector.recipe_snippet).to match(/^# In C:\\Windows\\Temp\\packer\\/)
+      end
+
       it "parses a unix path" do
         source_line = "/home/btm/src/chef/chef/spec/unit/fake_file.rb:2: undefined local variable or method `non_existent' for main:Object (NameError)"
         @resource.source_line = source_line


jdmundrawala:
and probably also https://github.com/chef/chef/issues/3343, however I'm not 100% sure about that

thommay:
LGTM :+1:

lib/chef/formatters/error_inspectors/resource_failure_inspector.rb
```diff
@@ -63,7 +63,7 @@ def add_explanation(error_description)
         def recipe_snippet
           return nil if dynamic_resource?
           @snippet ||= begin
-            if file = resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1] and line = resource.source_line[/^#{file}:([\d]+)/,1].to_i
+            if file = resource.source_line[/^(([\w]:)?[^:]+):([\d]+)/,1] and line = resource.source_line[/^#{Regexp.escape(file)}:([\d]+)/,1].to_i
```
stevendanna: :+1: on the change in general.  If you have some time to clean this up a bit, it might be nice to refactor this into well-named functions.

jdmundrawala: refactored

stevendanna:
:+1:

thommay:
@jdmundrawala please update the changelog :)

rojomisin:
I'm using vagrant testing a windows cookbook using 12.3 chef-client, is there a workaround?  Do I have to build my own chef-client, and if so how do I tell vagrant and/or chef-dk to use it (it always pulls from downloads repo).

lonniev:
Same plea as @rojomisin : it's great to here this is fixed and now when will it be released?

lonniev:
Would be sweet if:

```ruby
  config.omnibus.chef_version = '12.4.0.rc.0'
```

worked. It tries to but gets a messy 404 when trying to download that revision.

There a suggestions online for running a shell provisioner to gem install particular chef solo versions. I'll try that advice.

jdmundrawala:
@lonniev not sure what tool you are using, but i think if you can specifiy `prerelease` to be true, it will work. You might also have to change `chef_version` to be 12.4.0:
https://www.getchef.com/chef/metadata?v=12.4.0&prerelease=true&nightlies=false&p=el&pv=6&m=x86_64

lonniev:
Here's the Windows shell provisioner approach:

```ruby
    vmh.vm.provision :shell, :inline => "choco install ruby -fy"
    vmh.vm.provision :shell, :inline => "choco install rubygems -fy"

    vmh.vm.provision :shell, :inline => "gem install chef --version 12.4.0.rc.0 --no-rdoc --no-ri --conservative"
```

This tactic runs into the Ruby Gems SSL issue:

```shell
==> vb-tt-sync: ERROR:  Could not find a valid gem 'chef' (= 12.4.0.rc.0), here is why:
==> vb-tt-sync:           Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)
```

I'll try @jdmundrawala's :prerelease suggestion.

lonniev:
Here's the hope:

https://docs.chef.io/install_omnibus.html

And now to figure out how to pass arguments to the vagrant omnibus handler. ( Options -p and -s are needed. )

lonniev:
Looks like the omnibus code will pass along anything given as the version string:

```ruby
            if windows_guest?
              install_cmd = "cmd.exe /c #{install_script_name} #{version}"
```

and so this might work:

```ruby
    config.omnibus.chef_version = '12.4.0.rc.0 -p -s'
```

jdmundrawala:
ew, but ok. if it get's you moving forward. Ill see about fixing up omnitruck

lonniev:
Not yet. The string is getting parsed for supported versions. I need to read that code more closely.

Also, it looks like I could curl install chef using a shell provisioner line for what the Chef Omnibus link offers.

lonniev:
Any advice on what is the URL to the prerelease versions? I don't see a way to get vagrant omnibus to go fetch it and the chef site does its best to redirect me to the 12.3.0 latest release.

lonniev:
This looked promising but I don't see any msi installers within it. At least it suggests that the 12.4.0 RC versions are somewhere available:

https://opscode-omnibus-packages.s3.amazonaws.com/

lonniev:
Even better: https://docs.chef.io/api_omnitruck.html

lonniev:
Here it is:

```bash
$ curl -H "Accept: application/json" 'http://www.chef.io/chef/metadata?p=windows&v=2012&m=x86_64&v=latest&prerelease=true'
{
  "relpath": "/windows/2008r2/x86_64/chef-client-12.4.0-rc.0-1.msi",
  "md5": "3e5b1de34c6d8b373cdecee87018a64c",
  "sha256": "346196b526bdd09e5ea550711addaa0ff87d524c99dbcb4c5adc600cea5c9903",
  "url": "http://opscode-omnibus-packages.s3.amazonaws.com/windows/2008r2/x86_64/chef-client-12.4.0-rc.0-1.msi"
}
```

lonniev:
Here's a machine-specific working config snippet:

```ruby
  config.omnibus.install_url = "http://opscode-omnibus-packages.s3.amazonaws.com/windows/2008r2/x86_64/chef-client-12.4.0-rc.0-1.msi"
  config.omnibus.chef_version = "12.4.0.rc.0"
```

lonniev:
Interesting. All this time, here's the reason my choco resources were failing:

```bash
==> vb-tt-sync: [2015-06-03T16:16:57-07:00] FATAL: RuntimeError: chocolatey[wget] (chocolatey-wget::default line 28) had an error: RuntimeError: Wrong package name wget != Wget
```

The good news is that the darn Regex parsing error in the windows pathname for the recipe is gone.

lonniev:
I probably could revert to chef-solo :latest now, knowing that the exception behind the exception is an overly careful check in the Chocolatey recipe.

rojomisin:
thanks for digging into this @lonniev , I worked around by going back to 11, but I'll try the install_url/chef_version config snippet you posted above.

PR #3358, opened 2015-05-09
Separate audit and converge failures

mcquin:
Separate converge failures from run failures. The audit reporter only reports failed audits and audit phase exceptions, the resource reporter only reported converge and run failures. Audit, converge, and unexpected run failures are wrapped together and raised at the end of the run for remaining handling and exit status.

---
Previously, on LOST:
Adds a new configuration option `:audit_as_warning`

From [chef/config.rb](https://github.com/chef/chef/blob/377174723c2a5dbc9959abdaab213fd67899c6ea/lib/chef/config.rb#L356-L360):
```ruby
# When set to true (and audit mode is :enabled or :audit_only), one or more
# failed controls will cause the client run to be marked as failed.
# When set to false, only converge or unexpected errors will cause the chef
# client run to be marked as failed.
default :audit_as_warning, false
```

Additionally, failures which occur during the audit phase (that aren't non-passing controls) will cause the run to fail.

Q: Does this need a CLI option?

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -253,6 +253,11 @@ class Chef::Application::Client < Chef::Application
     :description    => "Enable audit-mode with `enabled`. Disable audit-mode with `disabled`. Skip converge and only perform audits with `audit-only`",
     :proc           => lambda { |mo| mo.gsub("-", "_").to_sym }
 
+  option :audit_as_warning,
+    :long           => "--audit-as-warning",
+    :description    => "Interpret failed controls as warnings.",
+    :boolean        => true
+
   option :minimal_ohai,
     :long           => "--minimal-ohai",
     :description    => "Only run the bare minimum ohai plugins chef needs to function",
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -253,11 +253,6 @@ class Chef::Application::Client < Chef::Application
     :description    => "Enable audit-mode with `enabled`. Disable audit-mode with `disabled`. Skip converge and only perform audits with `audit-only`",
     :proc           => lambda { |mo| mo.gsub("-", "_").to_sym }
 
-  option :audit_as_warning,
-    :long           => "--audit-as-warning",
-    :description    => "Interpret failed controls as warnings.",
-    :boolean        => true
-
   option :minimal_ohai,
     :long           => "--minimal-ohai",
     :description    => "Only run the bare minimum ohai plugins chef needs to function",
--- a/lib/chef/audit/audit_reporter.rb
+++ b/lib/chef/audit/audit_reporter.rb
@@ -167,10 +167,6 @@ class Chef
       def iso8601ify(time)
         time.utc.iso8601.to_s
       end
-
-      def error_message_for_run_data(error)
-        "#{err.class.to_s}: #{err.message}\n#{err.backtrace.join("\n")}"
-      end
     end
   end
 end
--- a/lib/chef/audit/audit_reporter.rb
+++ b/lib/chef/audit/audit_reporter.rb
@@ -34,6 +34,7 @@ class Chef
         @rest_client = rest_client
         # Ruby 1.9.3 and above "enumerate their values in the order that the corresponding keys were inserted."
         @ordered_control_groups = Hash.new
+        @exception = nil
       end
 
       def run_context
@@ -59,6 +60,7 @@ class Chef
       # known control groups.
       def audit_phase_failed(error)
         # The stacktrace information has already been logged elsewhere
+        @exception = error
         Chef::Log.debug("Audit Reporter failed.")
         ordered_control_groups.each do |name, control_group|
           audit_data.add_control_group(control_group)
@@ -116,8 +118,10 @@ class Chef
         Chef::Log.debug("Sending audit report (run-id: #{audit_data.run_id})")
         run_data = audit_data.to_hash
 
-        if error
-          run_data[:error] = "#{error.class.to_s}: #{error.message}\n#{error.backtrace.join("\n")}"
+        if @exception || error
+          errors = [@exception, error].uniq.compact
+          errors_messages = errors.map  { |err| "#{err.class.to_s}: #{err.message}\n#{err.backtrace.join("\n")}" }
+          run_data[:error] = errors_messages.join("\n")
         end
 
         Chef::Log.debug "Audit Report:\n#{Chef::JSONCompat.to_json_pretty(run_data)}"
@@ -164,6 +168,9 @@ class Chef
         time.utc.iso8601.to_s
       end
 
+      def error_message_for_run_data(error)
+        "#{err.class.to_s}: #{err.message}\n#{err.backtrace.join("\n")}"
+      end
     end
   end
 end
--- a/lib/chef/audit/audit_reporter.rb
+++ b/lib/chef/audit/audit_reporter.rb
@@ -120,7 +120,11 @@ class Chef
 
         if @exception || error
           errors = [@exception, error].uniq.compact
-          errors_messages = errors.map  { |err| "#{err.class.to_s}: #{err.message}\n#{err.backtrace.join("\n")}" }
+          errors_messages = errors.map  do |err|
+            msg = "#{err.class.to_s}: #{err.message}"
+            msg << "\n#{err.backtrace.join("\n")}" if err.backtrace
+            msg
+          end
           run_data[:error] = errors_messages.join("\n")
         end
 
--- a/lib/chef/audit/audit_reporter.rb
+++ b/lib/chef/audit/audit_reporter.rb
@@ -34,7 +34,7 @@ class Chef
         @rest_client = rest_client
         # Ruby 1.9.3 and above "enumerate their values in the order that the corresponding keys were inserted."
         @ordered_control_groups = Hash.new
-        @exception = nil
+        @audit_phase_error = nil
       end
 
       def run_context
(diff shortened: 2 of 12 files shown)


ranjib:
@mcquin the documentation is correct? it says "When set to true failed controls will cause the client run to be marked as failed", its should warn instead of failure ? what about naming something 'permissive' and 'enforced' similar to selinux, instead of having different boolean flags, we can have different levels of audits that way.

lib/chef/config.rb
```diff
@@ -353,6 +353,12 @@ def self.path_accessible?(path)
     # This can be removed when audit-mode is enabled by default.
     default :audit_mode, :disabled
 
+    # When set to true (and audit mode is :enabled or :audit_only), one or more
```
mcquin: "When set to false..."

lib/chef/config.rb
```diff
@@ -353,6 +353,12 @@ def self.path_accessible?(path)
...
+    # failed controls will cause the client run to be marked as failed.
+    # When set to false, only converge or unexpected errors will cause the chef
```
mcquin: "When set to true..."

spec/unit/client_spec.rb
```diff
@@ -432,179 +437,232 @@ def stub_for_run
...
+              expect(error.backtrace).to include(*converge_error.backtrace)
             end
           end
         end
 
-        context "when audit mode is disabled" do
-          include_context "a client run" do
-            before do
-              Chef::Config[:audit_mode] = :disabled
+        context "when audit phase errors" do
```
mcquin: Make this a `shared_context`.

spec/unit/client_spec.rb
```diff
@@ -407,13 +407,18 @@ def stub_for_sync_cookbooks
       end
     end
 
-    describe "when converge fails" do
+    context "when converge errors" do
```
mcquin: Make this a `shared_context`.

mcquin:
Good catch, @ranjib !

@chef/client-maintainers thoughts on the configuration option? I like the extensibility a non-boolean option provides. If we go that way we'll either need to use String keys or modify Test Kitchen to allow Symbol keys in kitchen.yml.

I don't think much discussion has gone into audit-mode levels yet. Maybe it would be cool to be able to flag some controls or control groups as 'permissive' or 'enforced', then set a global audit-mode level to use as the default (untagged controls/control groups will run at the globally configured level) or override (tagged audits will run at the globally configured level, regardless of tag).

The above isn't within the scope of this PR but could be good to think about when deciding this configuration option.

jtimberman:
Symbols in config for test kitchen is already [merged](https://github.com/test-kitchen/test-kitchen/pull/652), and released in 1.4.0! :cake:

lib/chef/audit/audit_reporter.rb
```diff
@@ -116,8 +118,10 @@ def post_auditing_data(error = nil)
         Chef::Log.debug("Sending audit report (run-id: #{audit_data.run_id})")
         run_data = audit_data.to_hash
 
-        if error
-          run_data[:error] = "#{error.class.to_s}: #{error.message}\n#{error.backtrace.join("\n")}"
+        if @exception || error
+          errors = [@exception, error].uniq.compact
+          errors_messages = errors.map  { |err| "#{err.class.to_s}: #{err.message}\n#{err.backtrace.join("\n")}" }
```
tyler-ball: You can leverage your `error_message_for_run_data` method you defined below in this map.  I'm pretty sure ruby allows you to do something like `errors.map &error_message_for_run_data` but someone with more ruby knowledge could chime in on that.

tyler-ball: Actually, if this is the only place you're going to have this formatting logic, I would just get rid of the `error_message_for_run_data` method.

mcquin: Method removed https://github.com/chef/chef/commit/cce10a71681b4011d00ecdd3048486f9618d0913

mcquin:
Doesn't look like these failures are related to this PR.

lib/chef/client.rb
```diff
@@ -537,6 +538,19 @@ def has_admin_privileges?
       Chef::ReservedNames::Win32::Security.has_admin_privileges?
     end
 
+    def wrap_exceptions(converge_error, audit_error)
+      if audit_error && !(audit_error.is_a?(Chef::Exceptions::AuditsFailed) && Chef::Config[:audit_as_warning])
```
tyler-ball: How would you feel about renaming this config key to `:error_on_audit_failure`?

tyler-ball:
:+1: to this change.  Nice cleanup in the specs too.

I _think_ historically we introduce a CLI option for the config options, but maybe @jtimberman could say if this is true?

jtimberman:
@tyler-ball Not all Chef::Config options have command-line arguments, IIRC. Generally that's a thing we try to do though. I think it makes sense for this particular setting. I propose `--dont-give-me-a-fatal-exception-for-audit-mode-failures`, or more succinctly, `--audit-as-warning`.

lib/chef/application/client.rb
```diff
@@ -253,6 +253,11 @@ class Chef::Application::Client < Chef::Application
     :description    => "Enable audit-mode with `enabled`. Disable audit-mode with `disabled`. Skip converge and only perform audits with `audit-only`",
     :proc           => lambda { |mo| mo.gsub("-", "_").to_sym }
 
+  option :audit_as_warning,
```
sersut: @mcquin I am not sure if this option is needed. Ideally in the future we would like to have a way to let certain audit failures not fail the run but I am not sure yet what is the best way to do it. 

Unless there is a need to I will vote to not include this option in this PR.

sersut:
@ranjib / @mcquin I think when we have decided to fail the client run when we have an audit failure. Captured in the [RFC](https://github.com/chef/chef-rfc/blob/master/rfc035-audit-mode.md#failures). 

I think for simplicity we should not introduce `audit_as_warning` flag. We can think more and discuss about how we want to translate audit failures into warnings. 

Other than that I think this PR is :+1:. 

/cc: @chef/client-core

lib/chef/client.rb
mcquin: audit failure is a bit ambiguous: was there an error during the audit phase or did some control not pass?

lib/chef/audit/audit_reporter.rb
```diff
@@ -116,8 +118,14 @@ def post_auditing_data(error = nil)
...
 
-        if error
-          run_data[:error] = "#{error.class.to_s}: #{error.message}\n#{error.backtrace.join("\n")}"
+        if @exception || error
+          errors = [@exception, error].uniq.compact
+          errors_messages = errors.map  do |err|
+            msg = "#{err.class.to_s}: #{err.message}"
+            msg << "\n#{err.backtrace.join("\n")}" if err.backtrace
+            msg
+          end
```
mcquin: Does joining the errors together like this make sense? Or should only the error captured during the audit phase (the one passed to `audit_phase_failed`) be sent with this data?

lib/chef/audit/audit_reporter.rb
```diff
@@ -59,6 +60,7 @@ def audit_phase_complete
       # known control groups.
       def audit_phase_failed(error)
         # The stacktrace information has already been logged elsewhere
+        @exception = error
```
tyler-ball: Having `exception` and `error` below is a little confusing.  I think renaming `@exception` to `@audit_phase_failed_error` makes it more clear what we stored.

lib/chef/audit/audit_reporter.rb
tyler-ball: If audit phase is enabled and an error is passed to this method, it will either be 1) the audit phase error, 2) the converge phase error or 3) the wrapped error containing both.

If its the audit error or the wrapped error, it will duplicate `@exception`.  If its the converge error, I feel like that should be reported by some converge-oriented reporter.

So I vote no, lets only send the `@exception` error stored earlier from `audit_phase_failed`.

sersut: @tyler-ball reading my mind from over there :smile: 

We had a chat with @mcquin and that's exactly what we think we should do.

mcquin:
Can I get eyes on this again @chef/client-core?

mcquin:
Removed configurability. Behavior is to only mark a run as failed on converge (non-audit phase) errors. Configurability requires an RFC.

ranjib:
:+1:

sersut:
:+1:

lib/chef/client.rb
```diff
@@ -500,6 +496,23 @@ def run
...
+          Chef::Exceptions::RunFailedWrappingError.new(converge_error, audit_error)
+        else
+          Chef::Exceptions::RunFailedWrappingError.new(run_error, converge_error, audit_error)
+        end
+        error.fill_backtrace
+        Chef::Application.debug_stacktrace(error)
+        raise error
+      end
+
+      run_error = nil
```
btm: cruft?

btm:
:+1: 

@chef/client-core, @mcquin would appreciate your review.

spec/unit/audit/audit_reporter_spec.rb
```diff
@@ -88,6 +88,29 @@
         reporter.run_completed(node)
       end
 
+      context "when audit phase failed" do
+
+        let(:audit_error) { double("AuditError", :class => "Chef::Exceptions::AuditError",
+          :message => "Audit phase failed with error message: derpderpderp",
```
danielsdeleo: "herpderp" ? :laughing: :sob:

danielsdeleo:
:+1:

tyler-ball:
:+1:

btm:
Rebased and merged to master.

PR #3360, opened 2015-05-11
Add check_resource_semantics! lifecycle method to provider

### LINKED ISSUE #3075: Package resource should not allow multi-package and source to both be specified
As an example,

```
package ['lspci', 'iptstate'] do
  source '/tmp/lspci.rpm'
end
```

does not make sense. We should not allow loading of resources specified by these, and fail in a sensible way.

jdmundrawala:
Added a lifecycle method to check if the resource makes sense for the provider. That way, we can fail before loading it. Will add specs if we want to solve the problem this way :)
Fixed #3075 as an example

cc @jaymzh @chef/client-core

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -83,6 +83,9 @@ class Chef
       new_resource.cookbook_name
     end
 
+    def check_resource_semantics!
+    end
+
     def load_current_resource
       raise Chef::Exceptions::Override, "You must override load_current_resource in #{self.to_s}"
     end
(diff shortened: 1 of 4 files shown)


lamont-granquist:
Yeah I guess we need to do this.  This should ideally go into the resource, but with the compile/converge mode split and with the possibility of resources being passed `lazy {}` we can't really do this until provider/converge time.

Yet more ammunition that compile/converge mode needs to go away and Resources and Providers need to be fused together.  We're starting to lazy up parameter validation and the Resource can't even do its job properly...

:+1: modulo a detour into (getoffmylawn)

smurawski:
:+1: Merge it

btm:
:+1: thanks for cleaning this task up.

PR #3364, opened 2015-05-12
Update ui.rb

rreyes:
I think this is a bug.

### DIFF
--- a/lib/chef/knife/core/ui.rb
+++ b/lib/chef/knife/core/ui.rb
@@ -66,7 +66,7 @@ class Chef
       # the logger API.
       def msg(message)
         begin
-          stdout.puts message
+          @stdout.puts message
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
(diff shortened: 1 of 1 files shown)


chef-supermarket:

* @rreyes

ranjib:
@rreyes whats not working?
https://github.com/chef/chef/blob/master/lib/chef/knife/core/ui.rb#L35

lamont-granquist:
using `stdout` and going through the `attr_reader :stdout` declaration is generally preferred than going directly to `@stdout` because in the former a typo like `stdoutt` will raise a NoSuchMethod error exception while `@stdoutt` will create a new instance variable.

rreyes:
Hello. It might be my limited knowledge of Ruby but I believe attr_reader will create the instance variables which are reachable inside the object using the @ prefix. So stdout and @stdout is not the same. If you use stdout then it will not be referring to the variable created by attr_reader nor the one injected by the constructor hence your code is not allowing the user to inject their own stdout. But I could be wrong. Don't quote me.

ranjib:
attr_reader creates a method which returns an instance variable with same name. it does not create the 

```ruby
class Foo
  attr_reader :x
end
```
irb
```sh
f= Foo.new
f.instance_variables # => []
f.x # => nil
f.instance_variables #=> []
```

so.. attr_reader does not create anything , from what i understand.

lamont-granquist:
`attr_reader :x` creates:

def attr_reader
  @attr_reader
end

the ivar is autovivified the first time it is referenced like i mentioned.

closing this since its specifically against the code recommendations that we like to use (see lots of my old PRs on providers where i rip out most of the `@`-signs).

PR #3366, opened 2015-05-12
Reduce provider/resource resolution stages down to 1

jkeiser:
- Adds `platform_version` filter to NodeMap, using the Chef::Version::Platform class to parse:
```ruby
class R < Chef::Resource
  provides :r, platform: 'ubuntu', platform_version: '>= 10.1'
end
```
- Adds specificity preferences to NodeMap:
```ruby
class A < Chef::Resource
  provides :a, platform: 'ubuntu', platform_version: '>= 12.0'
end
class B < Chef::Resource
  provides :a, platform: 'ubuntu'
end
class C < Chef::Resource
  provides :a
end
class BOverride < Chef::Resource
  provides :a, platform: 'ubuntu'
end
# No matter what order the above are declared in, preference order is A > B > C
# BOverride is declared *after* B, and so is preferred over B.
```
- Makes Resource/ProviderPriorityMap look through *all* matching arrays for a supported provider, not just the first matching array.  If `provides :service, platform: 'ubuntu'` fails to implement `provides?`, we still search `provides :service, os: 'linux'`.
- Removes node-specific NodeMaps and use the global Resource/ProviderPriorityMaps to implement `provides`
- Removes all ProviderMapping entries and uses `provides` instead
- Moves all requires of providers and resources out of ProviderMapping and Resource/ProviderPriorityMaps and into the relevant classes


jkeiser:
Closing this in favor of a merge to master now that `jk/missing_method_missing` is in.

PR #3372, opened 2015-05-13
openbsd_service provider utilizing rcctl(8)

joemiller:
Detect and use the rcctl(8) service management interface on OpenBSD nodes if available. If rcctl is not available, the existing OpenBSD service provider that works by modifying the /etc/rc.conf.\* files will be used instead.

The rcctl provider should be much simpler than trying to manage the modification the /etc/rc.conf.\* files and coordinating amongst other actors that may also be editing them. Additionally this provider supports setting service flags, eg: `dhcp_flags=em1`, which may be necessary in order for services that require flags to be managed by Chef.

Flags are supported by passing a flags key to the service parameters attribute, eg:

```
service 'dhcpd' do
  parameters(flags: 'em1')
  action [:enable, :start]
end
```

Caveats:
- Not all rc.d scripts will support reload. We should detect this and remove it from the list of :supported actions. As written, if you try to reload a service that doesn't support it the underlying rc.d/rcctl calls will fail and the resource will be fail to converge.
- Right now parameters are only checked and set on action_enable since rcctl(8) only allows setting parameters on enabled services. It might be desirable to also set parameters in action_start in the case that action_start is processed before parameters are set and causes the daemon startup to fail. Same for action_restart.


chef-supermarket:

* @joemiller

chef-supermarket:

lamont-granquist:
So what you can do here is use the provider resolver stuff in Chef-12 in order to pick the right provider based on the presence of rcctl.  I can help you out with that one, but you should rename this and keep the old provider to start with, then ping me and I can help you with the provides line.

joemiller:
Hi @lamont-granquist. Looking at some of the existing code and facilities I think the best approach may be to add a service_helper such as:

```
if ::File.exist?("/usr/sbin/rcctl")
  service_resource_providers << :rcctl
end
```

And then in the new `service/openbsd_rcctl.rb` provider use the helper as such:

```
provides :service, platform_family: 'openbsd', override: true do |node|
  Chef::Platform::ServiceHelpers.service_resource_providers.include?(:rcctl)
end
```

Does that sound like the best approach to you?

lamont-granquist:
Yeah that looks fine.

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+        include Chef::Mixin::ShellOut
+
+        def initialize(new_resource, run_context)
+          super
+          new_resource.supports[:status] = true
+          new_resource.status_command("rcctl check #{new_resource.service_name}")
+          new_resource.start_command("rcctl start #{new_resource.service_name}")
+          new_resource.stop_command("rcctl stop #{new_resource.service_name}")
+          new_resource.restart_command("rcctl restart #{new_resource.service_name}")
+          new_resource.reload_command("rcctl reload #{new_resource.service_name}")
```
lamont-granquist: This shouldn't be mutating the new_resource and not allowing the user any chance to override these.

The supports one should be using the superclass method `#supports` which dups a copy of the supports hash in the new_resource.  It should then be more like `supports[:status] = false if supports[:status].nil?` so that the user can set true/false and override.

you should also take a similar approach with the commands where they should not mutate the new resource but should be ivars on the provider -- and those values you have should be defaults and the new_resource should be able to override them.

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+        end
+
+        def load_current_resource
+          @current_resource = Chef::Resource::Service.new(new_resource.name)
+          current_resource.service_name(new_resource.service_name)
+
+          determine_current_status!
+          determine_current_parameters!
+
+          current_resource.enabled is_enabled?
```
lamont-granquist: this method should return the current_resource by convention.

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+            Chef::Log.debug("#{new_resource} already enabled - nothing to do")
+          else
+            converge_by("enable service #{new_resource}") do
+              update_params!
+              enable_service
+              Chef::Log.info("#{new_resource} enabled")
+            end
+          end
+          load_new_resource_state
+          new_resource.enabled(true)
```
lamont-granquist: you should not mutate the new_resource here, and don't need to set this flag at all.

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+        #
+        # As of OpenBSD 5.7 the following variables can be set (rcctl(8)):
+        #
+        #   flags, timeout, user.
+        #
+        # This code does not strictly enforce these variables since rcctl(8) will
+        # exit non-zero if you give it something it does not recognize.
+        #
+        def action_enable
+          if current_resource.enabled && !parameters_need_update?
```
lamont-granquist: so if parameters_need_update is false then the current_resource is probably not 'enabled' -- at least not enabled correctly so we then need to fix it.

if you make that change to the semantics of current_resource.enabled, then you can probably stop overriding action_enable and just provide your enable_service sublcass method, which is the correct way to go rather than hacking up the superclass.

lib/chef/provider/service/openbsd_rcctl.rb
lamont-granquist: this is awful and needs to be fixed in the superclass, which is an example of why you don't want to mostly copypasta bad code from the superclass so that we can fix this and you can just inherit the fixes...

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+        #   sshd_timeout=30
+        #   sshd_user=root
+        def determine_current_parameters!
+          current_resource.parameters Hash.new
+          shell_out!("rcctl get #{new_resource.service_name}", returns: [0, 1]).stdout.each_line do |line|
+            if line =~ /^#{Regexp.escape(new_resource.service_name)}_(.+?)=(.*)/
+              current_resource.parameters[Regexp.last_match(1)] = Regexp.last_match(2)
+            else
+              Chef::Log.warn("Problem parsing '#{line}' from 'rcctl get #{new_resource.service_name}'")
+              next
```
lamont-granquist: this 'next' is unnecessary.

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+            end
+          end
+          Chef::Log.debug("current params from #{new_resource}: #{current_resource.parameters}")
+        end
+
+        # check the current parameters against any newly specified parameters.
+        def parameters_need_update?
+          if new_resource.parameters && current_resource.parameters
```
lamont-granquist: It looks like current_resource.parameters is always at least the empty hash, so testing if its set isn't necessary (and confused me for a second that you had a logic bug here).

lib/chef/provider/service/openbsd_rcctl.rb
```diff
@@ -0,0 +1,157 @@
...
+            new_resource.parameters.each do |k, v|
+              return true if current_resource.parameters[k.to_s] != v.to_s
+            end
```
lamont-granquist: This is obviously additive and wont wipe out old parameters.  That means that if you set something and then remove it off the new_resource it won't be actually removed off of the running config.  That is most likely not the behavior people want, although I don't know how bad the default parameters usually are and how annoying it would be to have to always type them all into the resource (which is what you might be trying to work around?).

But either this needs to be strict or else you need to add another remove_parameters property in order to remove the old ones, or you could look at somehow initializing the properties list with a sane set of defaults and doing a hash merge of what the user wants to construct the parameters to apply strictly (which is not quite in order of preference, i like just being strict best, then being strict but merging with a set of default options, then adding some extra property to the resource to remove parameters least).

lamont-granquist:
ping?

thommay:
@joemiller if you'd like to take this up again we'd be happy to help, but until then we don't have cycles for OpenBSD sadly.

PR #3374, opened 2015-05-13
Reduce provider/resource resolution stages down to 1

jkeiser:
- Adds `platform_version` filter to NodeMap, using the Chef::Version::Platform class to parse:
```ruby
class R < Chef::Resource
  provides :r, platform: 'ubuntu', platform_version: '>= 10.1'
end
```
- Adds specificity preferences to NodeMap:
```ruby
class A < Chef::Resource
  provides :a, platform: 'ubuntu', platform_version: '>= 12.0'
end
class B < Chef::Resource
  provides :a, platform: 'ubuntu'
end
class C < Chef::Resource
  provides :a
end
class BOverride < Chef::Resource
  provides :a, platform: 'ubuntu'
end
# No matter what order the above are declared in, preference order is A > B > C
# BOverride is declared *after* B, and so is preferred over B.
```
- Makes Resource/ProviderPriorityMap look through *all* matching arrays for a supported provider, not just the first matching array.  If `provides :service, platform: 'ubuntu'` fails to implement `provides?`, we still search `provides :service, os: 'linux'`.
- Removes node-specific NodeMaps and use the global Resource/ProviderPriorityMaps to implement `provides`
- Removes all ProviderMapping entries and uses `provides` instead
- Moves all requires of providers and resources out of ProviderMapping and Resource/ProviderPriorityMaps and into the relevant classes


lib/chef/provider/mount.rb
```diff
@@ -24,7 +24,8 @@
 class Chef
   class Provider
     class Mount < Chef::Provider
-
+      Mount.provides :mount
```
thommay: why is this sometimes `Foo.provides :foo` and other times `provides :foo`?

jkeiser: Copypasta.  It should never have to do Mount.provides unless you're doing it from outside the class.  I'll search through and fix.

jkeiser: @thommay fixed.

lib/chef/provider/group/gpasswd.rb
```diff
@@ -22,6 +22,7 @@ class Chef
   class Provider
     class Group
       class Gpasswd < Chef::Provider::Group::Groupadd
+        provides :group, platform: %w(suse default)
```
lamont-granquist: There's no `default` platform.   This should probably be `os: linux` which via your herirarchy of specificity would get overridden by more precise `platform` or `platform_version` overrides.  Although this might simply be `provides :group` without any specificity since many OSen would be getting this via the default in the platform_map (everything that doesn't declare a group provider needs this provider).

lib/chef/provider/ifconfig/redhat.rb
```diff
@@ -22,6 +22,7 @@ class Chef
   class Provider
     class Ifconfig
       class Redhat < Chef::Provider::Ifconfig
+        provides :ifconfig, platform: %w(centos fedora redhat ibm_powerkvm cloudlinux parallels)
```
lamont-granquist: use platform_family here to tighten this up to fedora and redhat?

lib/chef/provider/group/usermod.rb
```diff
@@ -24,6 +24,8 @@ class Group
       class Usermod < Chef::Provider::Group::Groupadd
 
         provides :group, os: "openbsd"
+        provides :group, platform: %w(opensuse), platform_version: '>= 12.3'
+        provides :group, platform: %w(openindiana opensolaris nexentacore omnios solaris2 smartos hpux)
```
lamont-granquist: it looks like you can use `os: %w{solaris2 hpux}` here

lib/chef/provider/mount/solaris.rb
```diff
@@ -27,6 +27,8 @@ class Provider
     class Mount
       # Mount Solaris File systems
       class Solaris < Chef::Provider::Mount
+        provides :mount, platform: %w(openindiana opensolaris nexentacore omnios solaris2 solaris2 < 5.11 smartos)
```
lamont-granquist: "solaris2 < 5.11" doesn't look valid here.

lib/chef/provider/package.rb
```diff
@@ -490,6 +490,38 @@ def allow_downgrade
...
+      require 'chef/provider/package/pacman'
+      require 'chef/provider/package/ips'
+      require 'chef/provider/package/solaris'
+      require 'chef/provider/package/smartos'
+      require 'chef/provider/package/aix'
+      require 'chef/provider/package/paludis'
+
+      Chef.set_provider_priority_array :package, [ Homebrew, Macports ], os: "darwin"
+
+      Chef.set_provider_priority_array :package,  Dpkg,     os: "linux"
```
lamont-granquist: dpkg isn't a default for all linuxen, this should only be `platform_family: "debian"`

lib/chef/provider/package.rb
```diff
@@ -490,6 +490,38 @@ def allow_downgrade
...
+      Chef.set_provider_priority_array :package,  Apt,      platform: %w(ubuntu gcel linaro raspbian linuxmint debian)
```
lamont-granquist: `platform_family: %w{rhel fedora}`

lib/chef/provider/package.rb
```diff
@@ -490,6 +490,38 @@ def allow_downgrade
...
+      Chef.set_provider_priority_array :package,  Yum,      platform: %w(xenserver xcp centos amazon scientific fedora oracle redhat ibm_powerkvm cloudlinux parallels)
```
lamont-granquist: `platform_family: %w{rhel fedora}`

lib/chef/provider/package.rb
```diff
@@ -490,6 +490,38 @@ def allow_downgrade
...
+      Chef.set_provider_priority_array :package,  Zypper,   platform: %w(opensuse suse)
```
lamont-granquist: `platform_family: "suse"` i think

lib/chef/util/path_helper.rb
```diff
@@ -16,11 +16,208 @@
 # limitations under the License.
 #
 
-require 'chef-config/path_helper'
-
 class Chef
   class Util
-    PathHelper = ChefConfig::PathHelper
```
lamont-granquist: You've got some kind of rebase conflict here you resolved incorrectly?

lib/chef/platform/provider_mapping.rb
```diff
@@ -34,267 +29,7 @@ class << self
...
-                :package => Chef::Provider::Package::Paludis,
-                :service => Chef::Provider::Service::Systemd,
-              }
-            },
-            :default => {
-              :mount => Chef::Provider::Mount::Mount,
-              :user => Chef::Provider::User::Useradd,
-              :group => Chef::Provider::Group::Gpasswd,
-              :ifconfig => Chef::Provider::Ifconfig,
-            }
```
lamont-granquist: You should probably leave the default values here for now.  I don't think we have good enough coverage in the provider resolver unit tests to prove that we won't still ever need to hit this.

lamont-granquist:
If you revert the cleanup changes that you made to the provider_resolver unit specs does it all still pass?  I think that you didn't modify the logic at all, but its a very large patch.  You should still get green before and after the changes to that file as far as I can see.

lib/chef/provider/group/gpasswd.rb
jkeiser: Thanks! I decided on `provides :group`.

lib/chef/provider/ifconfig/redhat.rb
jkeiser: i.e. remove that line and add `platform_family: %w(fedora redhat)`?

jkeiser: Are all those platforms variants of fedora and redhat?

lib/chef/provider/group/usermod.rb
jkeiser: Again, not super sure myself, but happy to make the change :)  (I did this all by carefully keeping everything the same when I moved them.

lib/chef/provider/mount/solaris.rb
jkeiser: Ooh, thanks.  There were a few like that, thought I had them all.

lib/chef/provider/ifconfig/redhat.rb
lamont-granquist: yes to both.  and you're missing oracle and scientific, both of which should be here, which is why platform_family is better.

lamont-granquist: and it should be 'fedora' and 'rhel', not 'redhat'

lib/chef/provider/group/usermod.rb
lamont-granquist: Yeah, the old platform map was buggy, though, because it was based on platform and couldn't do platform_family or os.

lib/chef/util/path_helper.rb
jkeiser: Oy!  Restoring.

lib/chef/platform/provider_mapping.rb
jkeiser: Hmm.  I can do that.  Though I'll say, everything that I moved incorrectly, there were tests for.

lib/chef/provider/package.rb
jkeiser: I'm confused about these last two ... what exactly should be on dpkg and what should be on apt?  It's weird that apt and dpkg would both say debian right now.

jkeiser:
@lamont-granquist ran the prior version of the provider_resolver specs, all green.

jkeiser:
@lamont-granquist all changes made with the exception of restoring the default provider_mapping, which should be straightforward.

lamont-granquist:
:+1:

lib/chef/dsl/recipe.rb
```diff
@@ -36,58 +37,6 @@ module Recipe
       include Chef::Mixin::ShellOut
       include Chef::Mixin::PowershellOut
 
-      # method_missing must live for backcompat purposes until Chef 13.
-      def method_missing(method_symbol, *args, &block)
```
tyler-ball: :tada:

lib/chef/node_map.rb
```diff
@@ -68,74 +70,90 @@ def set(key, value, filters = {}, &block)
...
+
+    # List all matches for the given node and key from the NodeMap, from
+    # most-recently added to oldest.
+    #
+    # @param node [Chef::Node] The Chef::Node object for the run
+    # @param key [Object] Key to look up
+    # @return [Object] Value
+    #
+    def list(node, key)
+      # FIXME: real exception
```
tyler-ball: It would be nice to see this FIXME solved instead of propogated

spec/unit/provider_resolver_spec.rb
```diff
@@ -50,6 +54,51 @@
     allow(resource).to receive(:is_a?).with(Chef::Resource).and_return(true)
   end
 
+  def self.on_platform(platform, *tags,
+    platform_version: '11.0.1',
```
tyler-ball: why is `platform_version` defaulted here?

jkeiser: The others are as well, just they are defaulted to other parameters (see the call later on where we use ||)

jkeiser:
@lamont-granquist I have added tests for all the provider mappings and they passed (found a few bugs, see the latest commit).  I went and ran them against master as well, and [here are the changes](https://github.com/chef/chef/commit/3306ee516061ccaed278a5c497f9c51347f970e7) I made to master to get them to pass (indicating that these are changes the provider resolver branch is making now):

- Add Ifconfig::Redhat as providers for oracle, scientific, amazon, xcp and xenserver instead of the base Ifconfig provider.
- Have Package::Paludis provide :package on the exherbo platform.
- Have :portage_package provided by Package::Portage.

I don't actually know whether the first two things are correct or not; the ifconfig thing in particular was one of the things we changed around here with platform_family instead of platform, I think.

- Stop the provider_mapping from modifying its input (it was doing name.downcase! instead of name = name.downcase).  Frozen strings in my constants caught this.
- Make platform_provider

jkeiser:
If you can let me know whether the ifconfig and package provider changes are positive or negative, I will get this puppy in.

lamont-granquist:
Those changes look good.  The redhat stuff looks like a bugfix falling out of this cleanup rather than a backcompat break.

:shipit:

PR #3381, opened 2015-05-15
warn on cookbook self-deps

lamont-granquist:
cookbooks with self-deps:

name 'foo'
depends 'foo'

are useless and can waste cycles in the depsolver (particularly in
the non-solution case), and likely limit the possible choices of
depsolvers that we could pick at some point to replace gecode.

filtering out the self-dep here will prevent the chef server depsolver
from seeing the self-dep and needing to do the work to strip it.

also warn the user so that they can remove the self-dep.

in the future this will be a hard error.

foodcritic also has a [PR in for a] rule about removing these.

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -286,9 +286,13 @@ class Chef
       # === Returns
       # versions<Array>:: Returns the list of versions for the platform
       def depends(cookbook, *version_args)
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@danielsdeleo here you go first commit fixes up knife logging

ranjib:
:+1:

danielsdeleo:
:+1:

PR #3383, opened 2015-05-18
gem_package should install to the systemwide Ruby when using ChefDK.

### LINKED ISSUE #3382: gem_package installs to embedded Ruby instead of systemwide Ruby when using ChefDK
I dug into this, and it looks like the problem is that `is_omnibus?` is returning false here: https://github.com/chef/chef/blob/master/lib/chef/provider/package/rubygems.rb#L404. The relevant regex is:

```ruby
 %r!/opt/(opscode|chef)/embedded/bin!
```

but the embedded ruby with ChefDK has chefdk in the path, not opscode or chef:

```ruby
[368, 377] in /opt/chefdk/embedded/apps/chef/lib/chef/provider/package/rubygems.rb
   368:         def initialize(new_resource, run_context=nil)
   369:           super
   370:           @cleanup_gem_env = true
   371: require 'byebug'#<<<
   372: byebug #<<<
=> 373:           if new_resource.gem_binary
   374:             if new_resource.options && new_resource.options.kind_of?(Hash)
   375:               msg =  "options cannot be given as a hash when using an explicit gem_binary\n"
   376:               msg << "in #{new_resource} from #{new_resource.source_line}"
   377:               raise ArgumentError, msg
(byebug) RbConfig::CONFIG['bindir']
"/opt/chefdk/embedded/bin"
```

It seems to me that the fix here is as simple as expanding the above regex. I can submit a PR if this looks good.

Thanks!

jfly:
This fixes #3382. Let me know if I should add a test to `spec/unit/provider/package/rubygems_spec.rb`, and if so, what it should look like.

Thanks!

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -401,7 +401,7 @@ class Chef
         end
 
         def is_omnibus?
-          if RbConfig::CONFIG['bindir'] =~ %r!/opt/(opscode|chef|chefdk)/embedded/bin!
+          if RbConfig::CONFIG['bindir'] =~ %r!/(opscode|chef|chefdk)/embedded/bin!
             Chef::Log.debug("#{@new_resource} detected omnibus installation in #{RbConfig::CONFIG['bindir']}")
             # Omnibus installs to a static path because of linking on unix, find it.
             true
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -401,7 +401,7 @@ class Chef
         end
 
         def is_omnibus?
-          if RbConfig::CONFIG['bindir'] =~ %r!/opt/(opscode|chef)/embedded/bin!
+          if RbConfig::CONFIG['bindir'] =~ %r!/opt/(opscode|chef|chefdk)/embedded/bin!
             Chef::Log.debug("#{@new_resource} detected omnibus installation in #{RbConfig::CONFIG['bindir']}")
             # Omnibus installs to a static path because of linking on unix, find it.
             true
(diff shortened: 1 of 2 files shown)


chef-supermarket:

* @jfly

chef-supermarket:

lamont-granquist:
The simple way to test this in isolation is to just stub the RbConfig values and make sure is_omnibus? returns true or false correctly.

eric-tucker:
+1 !

jdmundrawala:
:+1: 
cc @chef/client-core

danielsdeleo:
LGTM

lamont-granquist:
:shipit: !

lib/chef/provider/package/rubygems.rb
```diff
@@ -401,7 +401,7 @@ def initialize(new_resource, run_context=nil)
         end
 
         def is_omnibus?
-          if RbConfig::CONFIG['bindir'] =~ %r!/opt/(opscode|chef)/embedded/bin!
+          if RbConfig::CONFIG['bindir'] =~ %r!/opt/(opscode|chef|chefdk)/embedded/bin!
```
tyler-ball: It would also be nice to make this work on Windows machines - by default it installs to `C:\opscode\chefdk`

jfly: I'm on a big trip right now, and won't be able to look into this for a while. Let me know how you'd like to proceed from here.

danielsdeleo: Hrm, this will be tricky if people relocate the install on windows. Should we just use `ANYTHING/embedded/bin` as the heuristic? @tyler-ball does `RbConfig::CONFIG['bindir']` normalize the path separator to forward slash?

tyler-ball: @danielsdeleo I have no idea.  But I say we update the regex to `%r!\/(opscode|chef|chefdk)\/embedded\/bin!` to handle all cases

jfly: What do the backslashes do here?

jfly: Updated in https://github.com/jfly/chef/commit/d27a5f8257eed6383f33acd0ab2f07655a69da19.

tyler-ball:
:+1: :shipit:

PR #3388, opened 2015-05-18
Integration tests fix ipv6

danielsdeleo:
* Fix IPv6 tests via Chef-Zero fix (TODO: merge that and use a released version of chef zero)
* Runs integration tests on travis
* Fix an issue where class "Foo" and module "Foo" were both defined.

Talking to John about the chef-zero thing https://github.com/chef/chef-zero/pull/133 Once that's done I'll redo the commit that pulls in the chef-zero fix.

@chef/client-core

### DIFF
--- a/spec/integration/client/client_spec.rb
+++ b/spec/integration/client/client_spec.rb
@@ -3,34 +3,35 @@ require 'chef/mixin/shell_out'
 require 'tiny_server'
 require 'tmpdir'
 
-def recipes_filename
-  File.join(CHEF_SPEC_DATA, 'recipes.tgz')
-end
-
-def start_tiny_server(server_opts={})
-  recipes_size = File::Stat.new(recipes_filename).size
-  @server = TinyServer::Manager.new(server_opts)
-  @server.start
-    @api = TinyServer::API.instance
-  @api.clear
-  #
-  # trivial endpoints
-  #
-  # just a normal file
-  # (expected_content should be uncompressed)
-  @api.get("/recipes.tgz", 200) {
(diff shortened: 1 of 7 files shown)


danielsdeleo:
I think the last update should work around a test that's not working in the appveyor environment. One maybe controversial thing is that `rake spec` now runs all the specs; personally I never run the specs via `rake` but maybe other people do? I could change the travis and appveyor configs to run `spec:all` if you'd rather.

spec/integration/client/client_spec.rb
```diff
@@ -341,7 +342,8 @@ class ::Blah
     end
   end
 
-  context "when using recipe-url" do
+  # Fails on appveyor, but works locally on windows and on windows hosts in Ci.
+  context "when using recipe-url", :skip_appveyor do
```
tyler-ball: It would be awesome to not need this.  Maybe updating Rubygems will magically fix this issue?

lamont-granquist:
Yeah now that the integ specs don't take half an hour you can make rake spec run them all again

:+1:

jdmundrawala:
:+1:

spec/integration/client/client_spec.rb
danielsdeleo: Yeah, I totally agree, but I just want to move on at this point. Seems quite unlikely that rubygems would be related, there's some kind of IO error when tar attempts to unzip the tarball. We're loading the tarball from disk in binary mode on the server, and writing it to disk in binary mode, so it's not the obvious thing. And we can't repro locally. :disappointed:

PR #3393, opened 2015-05-19
bump timeout up to 300 seconds

lamont-granquist:
seems to fix the timeouts on the exceedingly slow sparc boxes in CI

### DIFF
--- a/spec/functional/shell_spec.rb
+++ b/spec/functional/shell_spec.rb
@@ -29,7 +29,7 @@ describe Shell do
   describe "smoke tests", :unix_only => true do
     include Chef::Mixin::Command::Unix
 
-    TIMEOUT=30
+    TIMEOUT=300
 
     def read_until(io, expected_value)
       start = Time.new


lamont-granquist:
This basically lets the tests run as long as they need to.  Since it raises now we can be certain that if it works we are never hitting this timeout.  If it takes longer than 300 seconds something is wrong, it will go red, and someone should fix it -- we won't start silently leaking 5 minutes of time into all our test runs.  On our sparc boxes 30 seconds is too short (which is overall a problem, and I suspect that in the port to manhattan we lost some tuning that was done to the old-ci sparc boxen, but we have visibility into that simply via the whole test phases taking over an hour to run on sparc).

lamont-granquist:
Yep, so the sparc cores that we've got just suck that badly and we're single threaded so we're gonna have to suck it.  We're CPU bound on a single core on a server that has like 100 of them.  This is the right patch and there's nothing else to be done...

*le sigh*

lamont-granquist:
@client-core review and weep...

danielsdeleo:
ugh. This is okay to merge but I can't make myself give it the thumbs up emoji.

lamont-granquist:
appveyor fail is unrelated...

PR #3396, opened 2015-05-19
Show trace for RecipeNotFound errors when it originates from include recipe

danielsdeleo:
If you have code like `include_recipe "cookbook_exists::RECIPE_DOESNT_EXIST"`, you don't get the cookbook trace or error context because `RecipeNotFound` is assumed to have originated from a bad run list in chef-solo (I think). What we should be checking for is whether we are able to locate the proper recipe context or not. I found this relevant commit https://github.com/chef/chef/commit/abefd661954fd533c227e17ff74065bf114e8b24, but didn't find the PR it was from, so I could not locate the original bug report. I added a regression test for what I suspect the original bug was, so I think this should not regress the original issue.

@chef/client-core

### DIFF
--- a/lib/chef/formatters/error_inspectors/compile_error_inspector.rb
+++ b/lib/chef/formatters/error_inspectors/compile_error_inspector.rb
@@ -30,15 +30,16 @@ class Chef
 
         def initialize(path, exception)
           @path, @exception = path, exception
+          @backtrace_lines_in_cookbooks = nil
+          @file_lines = nil
+          @culprit_backtrace_entry = nil
+          @culprit_line = nil
         end
 
         def add_explanation(error_description)
-          case exception
-          when Chef::Exceptions::RecipeNotFound
-            error_description.section(exception.class.name, exception.message)
-          else
-            error_description.section(exception.class.name, exception.message)
+          error_description.section(exception.class.name, exception.message)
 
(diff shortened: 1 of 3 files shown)


lamont-granquist:
yeah LGTM.

lib/chef/formatters/error_inspectors/compile_error_inspector.rb
```diff
@@ -30,15 +30,16 @@ class CompileErrorInspector
 
         def initialize(path, exception)
           @path, @exception = path, exception
+          @backtrace_lines_in_cookbooks = nil
+          @file_lines = nil
```
jdmundrawala: Is this used anywhere?

danielsdeleo: In the `#file_lines` method. It's technically okay because the code is `@file_lines ||= IO.readlines(culprit_file)` which doesn't trigger a ruby warning for uninitialized variables, but generally I like to create all the instance vars in `initialize` so you can see them all at a glance.

jdmundrawala: Ahh, forgot I wasn't look at the entire file

jdmundrawala:
:+1:

PR #3399, opened 2015-05-20
Only run functional and unit tests in appveyor

jdmundrawala:


thommay:
seems reasonable :+1:

lamont-granquist:
this still took 20 mins and the 40 min timeouts started on PRs before dan merged in the change that make rake spec run the integ tests again.  there's still something that changed, since appveyor used to be really fast before yesterday...

appveyor.yml
```diff
@@ -31,4 +31,4 @@ build_script:
 
 test_script:
   - SET SPEC_OPTS=--format progress
-  - bundle exec rake spec
+  - bundle exec rake spec:appveyor
```
btm: Do we need a special task? How about `bundle exec rake spec:unit spec:functional`? That should run the unit tests and if they pass run the functional tests.

PR #3402, opened 2015-05-20
Don't call get_last_error for net api

jdmundrawala:
This should fix our internal CI
cc @thommay @chef/client-windows

### DIFF
--- a/lib/chef/win32/net.rb
+++ b/lib/chef/win32/net.rb
@@ -117,11 +117,7 @@ END
 
         rc = NetUserAdd(server_name, 3, buf, nil)
         if rc != NERR_Success
-          if Chef::ReservedNames::Win32::Error.get_last_error != 0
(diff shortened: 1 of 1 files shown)


thommay:
well, appveyor likes it, and it looks like the MS docs agree with you, so :+1:

btm:
:+1: 

As an aside, I noticed NERR_BadUsername from #net_api_error! isn't listed on: 
https://msdn.microsoft.com/en-us/library/windows/desktop/aa370659%28v=vs.85%29.aspx

ksubrama:
@btm  That was them reusing a system error code from https://msdn.microsoft.com/en-us/library/windows/desktop/ms681386(v=vs.85).aspx

In the original PR, I had figured out how to fetch the various error codes and string from their various sources but it was pretty insane how hard it was to do something that simple - so it's on our backlog instead :).

ksubrama:
:+1:

PR #3406, opened 2015-05-20
add unicode WM_SETTINGCHANGE broadcast

lamont-granquist:
helps powershell pick up the env change settings.

### DIFF
--- a/lib/chef/mixin/windows_env_helper.rb
+++ b/lib/chef/mixin/windows_env_helper.rb
@@ -21,11 +21,11 @@ require 'chef/exceptions'
 require 'chef/platform/query_helpers'
 require 'chef/win32/error' if Chef::Platform.windows?
 require 'chef/win32/api/system' if Chef::Platform.windows?
+require 'chef/win32/api/unicode'
 
 class Chef
   module Mixin
     module WindowsEnvHelper
-
       if Chef::Platform.windows?
         include Chef::ReservedNames::Win32::API::System
       end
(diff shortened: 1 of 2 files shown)


lamont-granquist:
might fix opscode-cookbooks/windows#111

lamont-granquist:
@jdmundrawala this code looks correct but it still doesn't work in test-kitchen

the rubyinstaller also adds its own path already (the windows_path call i had in my recipe isn't executed because of idempotency) but also fails to get added to powershell's path when i switch to a window, so it doesn't look like this is specific to chef any more.

lib/chef/mixin/windows_env_helper.rb
```diff
@@ -39,7 +39,16 @@ module WindowsEnvHelper
...
       def broadcast_env_change
         flags = SMTO_BLOCK | SMTO_ABORTIFHUNG | SMTO_NOTIMEOUTIFNOTHUNG
-        SendMessageTimeoutA(HWND_BROADCAST, WM_SETTINGCHANGE, 0, FFI::MemoryPointer.from_string('Environment').address, flags, 5000, nil)
+        # for why two calls, see:
+        # http://stackoverflow.com/questions/4968373/why-doesnt-sendmessagetimeout-update-the-environment-variables
+        if ( SendMessageTimeoutA(HWND_BROADCAST, WM_SETTINGCHANGE, 0, FFI::MemoryPointer.from_string('Environment').address, flags, 5000, nil) == 0 )
+          Chef::ReservedNames::Win32::Error.raise!
+        end
+        if ( SendMessageTimeoutW(HWND_BROADCAST, WM_SETTINGCHANGE, 0, FFI::MemoryPointer.from_string(
+            Chef::ReservedNames::Win32::Unicode.utf8_to_wide('Environment')
```
jdmundrawala: String has a monkey patch that does this:
https://github.com/chef/chef/blob/master/lib/chef/win32/unicode.rb#L53

lib/chef/mixin/windows_env_helper.rb
```diff
@@ -21,11 +21,11 @@
 require 'chef/platform/query_helpers'
 require 'chef/win32/error' if Chef::Platform.windows?
 require 'chef/win32/api/system' if Chef::Platform.windows?
+require 'chef/win32/api/unicode'
```
jdmundrawala: Needs guard

jdmundrawala:
I'm :+1: for this. It seems the internet says we should be doing it this way.

lib/chef/mixin/windows_env_helper.rb
lamont-granquist: yes monkey patching string is vile.

PR #3407, opened 2015-05-21
Make native mode the default for policyfiles

danielsdeleo:
Policyfile APIs are going to be enabled by default in the next Chef Server release (and they're enabled in hosted now). This should be the default. Technically this is a breaking change but this has been labeled an experimental feature since the beginning so IMO we can follow the pre-1.0 semver rules.

@chef/client-core


lamont-granquist:
:+1:

thommay:
:shipit:

Issue #3416, opened 2015-05-22
Behavior change in 12.4: nameless resources


coderanger:
As part of @jkeiser's new DSL work, the behavior of a resource with no name `file { ... }` has changed. I'm hesitant to call this a compat break because the old behavior was probably a bug for everyone except me. In the current release that code would result in a resource with name set to `nil`, in master that is a syntax error. This is because the old `method_missing` code uses `args[0]` as the name, which is `nil` if there are no args. The new code has real methods defined for each resource in the DSL and those require a name parameter.

I'm cool with fixing this on the Poise side of things if we can find a non-terrible way to do that, and if we agree the old `nil` behavior was a bug and not an API compat break to change it.

coderanger:
A solution for this on the Chef side would be to make the DSL stub methods have a default value for the name parameter (probably `nil`) and then in `declare_resource` check that the name is something valid (not `nil` or `''`) and raise an exception if it isn't. Kind of roundabout though, and this will probably only benefit me (I override `declare_resource` in this case and it fills in a default name in some situations). An alternative would be for Poise to create overrides for every DSL stub method, which seems tricky and still gross.

coderanger:
Another advantage of doing the error check in `declare_resource` is we could show a nicer error message, more like the runtime error display that shows the snippet of recipe code, rather than a standard `SyntaxError`. There is already a check in `ResourceBuilder` for `name.nil?` so that's a thing too.

jkeiser:
You mean we allowed resource name to default to nil?  That seems like an
unintended compat change.

What I really want is to involve the resource in its own construction, I.e.
implement build_resource in the resource class and let it take whatever
arguments it wants.

coderanger:
Yes, I use this in nested resources to allow for simpler syntax:

```ruby
application 'foo' do
  bundle_install do
    deployment true
  end
end
```

In that case `bundle_install` gets a default name, but this is now broken because the DSL stub method for it requires a name. I created #3417 with a potential fix on the Chef side to restore the old behavior.

lamont-granquist:
this got fixed and merged

PR #3417, opened 2015-05-22
Re-allow nameless resources

coderanger:
A potential fix for #3416. Probably check that ticket for discussion as to if this is the right approach.


coderanger:
To confirm, this does fix my failing tests in Poise and Poise Application.

```
Finished in 0.44149 seconds (files took 2.87 seconds to load)
21 examples, 0 failures
```

jkeiser:
Why name.empty? Is that disallowing something that wasn't disallowed before?

What I would love is if the DSL could be overridden in provides, to take any arguments you want. Like `provides :blah, invoke: proc { ... }`.  Or perhaps something on the resource class, like `MyResource.declare(parent, *args, &block)`. Obviously desiring to exactly what we do now. Then you could override yours with `name=nil` without affecting anyone else or forcing special checking.

But it shouldn't block this. Other than the name empty question, this looks good to me.

coderanger:
Yeah, currently `package []` would be allowed I think, which could happen if you are trying to do a multi-package install but your code is wrong (or somesuch). That also checks for `package ''` which should similarly not be allowed for reasons of general sanity.

As to adding this to the DSL, my issue with that is I want this to work on all resources not just my own (in an optimal world at least). Allowing it to be a per-resource thing would solve it more cleanly for my own application_* resources but I would have to build more wrappers for core resources and 3rd party cookbooks to make them act the same way, certainly not the end of the world but the inconsistency would be frustrating.

jkeiser:
If possible I'd rather not add more restrictions on things, both for backwards compatibility purposes, and for the fact that there can be genuine things whose identity is an empty string.

jkeiser:
As for working on all resources, you mean if name is declared with a default value, honoring that when instantiating the resource? That seems like a great feature to add and I would totally support such a patch.

jkeiser:
Though it would be super nice if we used some other method than a `nil` sentinel to detect the missing argument :)

coderanger:
I don't think it could be an upstream patch, the default name is a feature of the application resource's override of `declare_resource`. If it sees no name was provided, it fills one in there based on the application config.

coderanger:
Nuked the `empty?` check, though perhaps it should be done as a deprecation warning?

jkeiser:
If there were lazy defaults, I bet it could ... :)

coderanger:
The default name is not a feature of the resource being named. It is part of the subresource DSL code https://github.com/poise/poise/blob/master/lib/poise/helpers/subresources/container.rb#L94-L111

jkeiser:
Oh! This is all about nesting, gotcha. I'm still not resolved on the right way to get nesting onto mainline (only that I really, really want it).

coderanger:
Yahr, that situation I use this is stuff like

```ruby
application '/path'
  rails do
     # some options
  end
end
```

`rails` gets a default name of `'/path'` so the author doesn't have to type it twice (or do something weird looking like `rails name do`.

danielsdeleo:
@coderanger can you add a regression test somewhere?

:+1: from me on the fix itself.

@chef/client-core

danielsdeleo:
@thommay BTW this looks like a regression that we should fix before 12.4 final

danielsdeleo:
https://www.youtube.com/watch?v=zSAJ0l4OBHM

coderanger:
Added some specs, for `Chef::DSL::Resources` in general as well as this case.

danielsdeleo:
Travis failure is unrelated, see https://github.com/chef/chef/pull/3454

danielsdeleo:
:+1:

danielsdeleo:
A rebase should fix up the travis tests

PR #3419, opened 2015-05-22
Fix cli issue with unset chef_repo_path

scotthain:
Fixes issue for when environment variable is not set and chef_repo_path expects one.

tl;dr - if you have something like this in a config.rb/knife.rb:
```chef_repo_path ENV['SOMETHING']```
and $SOMETHING is not set, when running ```chef-client -z``` it will throw a wonderful error like so:
```[2015-05-22T13:11:58-07:00] FATAL: NoMethodError: undefined method `map' for nil:NilClass```
Which gives you nothing to go on. This issues a warning and resets the config to default value.

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -279,6 +279,12 @@ class Chef::Application::Client < Chef::Application
     Chef::Config[:chef_server_url] = config[:chef_server_url] if config.has_key? :chef_server_url
 
     Chef::Config.local_mode = config[:local_mode] if config.has_key?(:local_mode)
+
+    if Chef::Config.has_key?(:chef_repo_path) && Chef::Config.chef_repo_path.nil?
+      Chef::Config.delete(:chef_repo_path)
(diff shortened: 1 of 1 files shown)


danielsdeleo:
It'd be cool if we had a more rigorous approach to validating config, but we don't and this is where all the rest of this kind of logic is. :+1:

scotthain:
@danielsdeleo I'd love to see a Chef::Config.validate type method, but that seems like a pretty big refactor and logic crazyness...

tyler-ball:
:+1:

danielsdeleo:
:tada:

lamont-granquist:
@scotthain you did miss a CHANGELOG.md update tho

scotthain:
@lamont-granquist Oh sorry! I wasn't aware of that procedure - my bad. Want me to open another PR with the change? I can do that in a jiffy?

danielsdeleo:
@scotthain best to do that after merging b/c git merge doesn't like the CHANGELOG format very much. You can just do an informal review by pinging ppl in hipchat with the text you want to add

lamont-granquist:
yeah, i usually don't bother with review on CHANGELOG.md updates

PR #3420, opened 2015-05-23
RSpec isn't a bug database or kanban board

danielsdeleo:
Mass purge of pending specs. These have been around and not been worked on for a very long time, so they're an unfunded mandate that vomit a bunch of test output and don't do much else.

@chef/client-core


lamont-granquist:
fair enough.  :+1:

ranjib:
:+1:

PR #3425, opened 2015-05-26
Fix overridden method `skip` causing arity failures

danielsdeleo:
This was causing this error in Ci:

```
  1) Chef::Provider::User::Useradd action :create when the user does not exist beforehand and the username contains a single quote ensures the user exists
     Failure/Error: skip(reason)
     ArgumentError:
       wrong number of arguments (1 for 0)
     # ./spec/functional/resource/user/useradd_spec.rb:172:in `block (4 levels) in <top (required)>'
```

@chef/client-core

### DIFF
--- a/spec/functional/resource/user/useradd_spec.rb
+++ b/spec/functional/resource/user/useradd_spec.rb
@@ -162,15 +162,10 @@ describe Chef::Provider::User::Useradd, metadata do
     end
   end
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
lol :+1:

lamont-granquist:
(works fine on my ubuntu box)

PR #3428, opened 2015-05-26
Move `skip` for useradd test to metadata

danielsdeleo:
Otherwise we run useradd provider prior to the before block with the
example-level `skip` being called.

@chef/client-core


lamont-granquist:
still passes for me

Issue #3394, opened 2015-05-19
Chef Client 12.3.0 Ignoring File Resource Backup Counts on Windows 2012R2


sean-horn:
Zendesk 4271
Chef Client 12.3.0
Windows 2012R2

As a customer using a windows platform if I run chef-client six times with the following recipe, I see six backed up files of three different filenames. This is not correct behavior

On windows, it appears that only `backup false` is working properly.

The default max number of backups to retain when modifying a file resource file is 5.

On linux with this version of chef-client and the specified recipe, I see the proper outcome after six runs...

* 5 copies of `defaultbkups` 
* 3 copies of `threebkups`
* 3 copies of `threebkups_same_mtime`
* 0 copies of `bkupsfalse`

```
require 'fileutils'

source = "#{Chef::Config[:file_cache_path]}/sourcefile"

dest1 = "#{Chef::Config[:file_cache_path]}/defaultbkups"
dest2 = "#{Chef::Config[:file_cache_path]}/threebkups"
dest3 = "#{Chef::Config[:file_cache_path]}/threebkups_same_mtime"
dest4 = "#{Chef::Config[:file_cache_path]}/bkupsfalse"

if File.exist?(source)
  FileUtils.cp source, dest1 unless File.exist?(dest1)
  FileUtils.cp source, dest2 unless File.exist?(dest2)
  FileUtils.cp source, dest3, preserve: true unless File.exist?(dest3)
  FileUtils.cp source, dest4 unless File.exist?(dest4)
end

cookbook_file source do
  action :create
end

file dest1 do
  action :delete
end

file dest2 do
  backup 3
  action :delete
end

file dest3 do
  backup 3
  action :delete
end

file dest4 do
  backup false
  action :delete
end
```

However, on w2k12r2 with Chef Client 12.3.0 after 6 runs, I see

* 6 copies of `defaultbkups` 
* 6 copies of `threebkups`
* 6 copies of `threebkups_same_mtime`
* 0 copies of `bkupsfalse`

```
C:\chef>dir c:\chef\backup\chef\cache
 Volume in drive C has no label.
 Volume Serial Number is 08C8-7B16

 Directory of c:\chef\backup\chef\cache

05/19/2015  04:15 PM    <DIR>          .
05/19/2015  04:15 PM    <DIR>          ..
05/19/2015  04:10 PM                39 defaultbkups.chef-20150519161054.907382
05/19/2015  04:12 PM                39 defaultbkups.chef-20150519161214.844916
[...]

PR #3431, opened 2015-05-26
Fix backups on windows for the file resource

jdmundrawala:
The backup utility was using Dir.[], which on breaks with \ in
the path name. This code is replaces with listing the directory and
matching the correct files
Also, on windows, the creation part stripped the drive name, while the
part that searched for old backups did not.

This should solve #3394.

### DIFF
--- a/lib/chef/util/backup.rb
+++ b/lib/chef/util/backup.rb
@@ -78,8 +78,16 @@ class Chef
         Chef::Log.info("#{@new_resource} removed backup at #{backup_file}")
       end
 
+      def unsorted_backup_files
+        # If you replace this with Dir[], you will probably break Windows.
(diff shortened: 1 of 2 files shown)


jdmundrawala:
cc @chef/client-windows

btm:
Add a regression test that uses backslashes in case someone tries to simplify that with Dir[]?

jdmundrawala:
@btm modified the test and added a comment

btm:
:+1: 

@chef/client-core this fixes a windows bug but is in core code. Need another thumb.

danielsdeleo:
:+1: regression test is well written :stars:

btm:
Rebased and merged to master.

Issue #3435, opened 2015-05-27
chef-client, in daemon mode, when called with a HUP doesn’t release old logging file handles


cachamber:
**Issue:**  When chef-client is run in daemon mode, with log rotation, it holds onto all logging file handles after a rotation, as it never closes the old log.
**Reproduction:**
1. Start chef-client in daemon mode
2. Validate log file handle with lsof 
3. Move the logfile to another name to simulate log rotation
4. Validate log file handle has moved to the new filename
5. Use init script to reload chef-client (Sends a HUP to the process)
6. Verify, using lsof that chef-client now has 2 log file handles
```
[root@xldcac01 ~]# chef-client --version
Chef: 12.3.0
[root@xldcac01 ~]# ps -efa |grep chef-client 
root     25450     1  0 13:29 ?        00:00:00 /opt/chef/embedded/bin/ruby /usr/bin/chef-client -d -c /etc/chef/client.rb -L /var/log/chef/client.log -P /var/run/chef/client.pid -i 1800 -s 20 -f
[root@xldcac01 ~]# lsof -p 25450 |grep client.log
chef-clie 25450 root   10w   REG  253,2      244     1875 /var/log/chef/client.log
[root@xldcac01 ~]# mv /var/log/chef/client.log /var/log/chef/client.log-1
[root@xldcac01 ~]# lsof -p 25450 |grep client.log
chef-clie 25450 root   10w   REG  253,2      244     1875 /var/log/chef/client.log-1
[root@xldcac01 ~]# /etc/init.d/chef-client reload
Reloading chef-client:                                     [  OK  ]
[root@xldcac01 ~]# lsof -p 25450 |grep client.log
chef-clie 25450 root   10w   REG  253,2      309     1875 /var/log/chef/client.log-1
chef-clie 25450 root   12w   REG  253,2        0     1858 /var/log/chef/client.log
[root@xldcac01 ~]# 
```
**Location:** Found in https://github.com/chef/chef/blob/master/lib/chef/application.rb
**Analysis:**
The HUP handler is on lines 77-80, calls “reconfigure” on line 79.
“reconfigure” defined on lines 47-53  the issue is that reconfigure calls “configure_logging” on line 49
“configure_logging” is defined on line 137, the offending line is line 138: 
```
Chef::Log.init(MonoLogger.new(Chef::Config[:log_location]))
```
There isn't a check to see if a logger already exists.
**Fix:**
Before creating a new instance of the logger, check to see if one exists, and close it.
```
[...]

PR #3436, opened 2015-05-27
ISSUE #3435 - Close old logfile before creating a new one

cachamber:
Possible fix for https://github.com/chef/chef/issues/3435

In essence, check to see if a logging instance exists before creating a new instance.  If so, close the old instance before opening the new one.


lamont-granquist:
Is there a better way to check than Chef::Log.info? that looks like you're getting a side effect and its not the API that i'd expect to call and it fails my does-it-make-sense-still-in-12-months test.  Nothing about the name '#info?' suggests that you're directly testing for presence.

lamont-granquist:
We'd also need a  spec test around this.

I also wonder if its better to put that logic into Chef::Log.init rather than in Chef::Application#configure_logging so that its more generally applied anytime Chef::Log is used.

irvingpop:
hey @cachamber any progress on this one?

cachamber:
@irvingpop  - No update.  I haven't had cycles to spend on it yet.  I expect to have some time in the next week or so.

thommay:
@cachamber hey, are you still interested in getting this finished up?

thommay:
Just merged a change to mixlib-log that should resolve this.

PR #3437, opened 2015-05-27
Add "action" to Resource (RFC 50)

jkeiser:
As per https://github.com/chef/chef-rfc/blob/master/rfc050-resource-actions.md

### DIFF
--- a/lib/chef/dsl/recipe.rb
+++ b/lib/chef/dsl/recipe.rb
@@ -122,9 +122,9 @@ class Chef
 
       def describe_self_for_error
         if respond_to?(:name)
-          %Q[`#{self.class.name} "#{name}"']
+          %Q[`#{self.class} "#{name}"']
         elsif respond_to?(:recipe_name)
-          %Q[`#{self.class.name} "#{recipe_name}"']
+          %Q[`#{self.class} "#{recipe_name}"']
         else
           to_s
         end
@@ -175,6 +175,24 @@ class Chef
         rescue NameError
           raise NameError, "No resource, method, or local variable named `#{method_symbol}' for #{describe_self_for_error}"
         end
+      end
+
+      module Everything
+        require 'chef/dsl/data_query'
+        require 'chef/dsl/platform_introspection'
+        require 'chef/dsl/include_recipe'
+        require 'chef/dsl/registry_helper'
+        require 'chef/dsl/reboot_pending'
+        require 'chef/dsl/audit'
+        require 'chef/dsl/powershell'
+        include Chef::DSL::DataQuery
+        include Chef::DSL::PlatformIntrospection
+        include Chef::DSL::IncludeRecipe
+        include Chef::DSL::Recipe
+        include Chef::DSL::RegistryHelper
+        include Chef::DSL::RebootPending
+        include Chef::DSL::Audit
+        include Chef::DSL::Powershell
       end
     end
   end
--- a/lib/chef/dsl/recipe.rb
+++ b/lib/chef/dsl/recipe.rb
@@ -177,7 +177,7 @@ class Chef
         end
       end
 
-      module Everything
+      module FullDSL
         require 'chef/dsl/data_query'
         require 'chef/dsl/platform_introspection'
         require 'chef/dsl/include_recipe'
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -269,24 +269,35 @@ class Chef
       # def notifications_resolved
       # end
 
+      #
+      # Resource events and ordering:
+      #
+      # 1. Start the action
+      #    - resource_action_start
+      # 2. Check the guard
+      #    - resource_skipped: (goto 7) if only_if/not_if say to skip
+      # 3. Load the current resource
+      #    - resource_current_state_loaded
+      #    - resource_current_state_load_bypassed (if not why-run safe)
+      # 4. Check if why-run safe
+      #    - resource_bypassed: (goto 7) if not why-run safe
+      # 5. During processing:
+      #    - resource_update_applied: For each actual change
+      # 6. Processing complete status:
+      #    - resource_failed if the resource threw an exception while running
+      #    - resource_failed_retriable: (goto 3) if resource failed and will be retried
+      #    - resource_updated if the resource was updated (resource_update_applied will have been called)
+      #    - resource_up_to_date if the resource was up to date (no resource_update_applied)
+      # 7. Processing complete:
+      #    - resource_completed
+      #
+
       # Called before action is executed on a resource.
       def resource_action_start(resource, action, notification_type=nil, notifier=nil)
       end
 
-      # Called when a resource fails, but will retry.
-      def resource_failed_retriable(resource, action, retry_count, exception)
-      end
-
-      # Called when a resource fails and will not be retried.
-      def resource_failed(resource, action, exception)
-      end
-
       # Called when a resource action has been skipped b/c of a conditional
       def resource_skipped(resource, action, conditional)
-      end
-
-      # Called when a resource action has been completed
-      def resource_completed(resource)
       end
 
       # Called after #load_current_resource has run.
@@ -302,20 +313,33 @@ class Chef
       def resource_bypassed(resource, action, current_resource)
       end
 
-      # Called when a resource has no converge actions, e.g., it was already correct.
-      def resource_up_to_date(resource, action)
-      end
-
       # Called when a change has been made to a resource. May be called multiple
       # times per resource, e.g., a file may have its content updated, and then
       # its permissions updated.
       def resource_update_applied(resource, action, update)
       end
 
+      # Called when a resource fails, but will retry.
+      def resource_failed_retriable(resource, action, retry_count, exception)
+      end
+
+      # Called when a resource fails and will not be retried.
+      def resource_failed(resource, action, exception)
+      end
+
       # Called after a resource has been completely converged, but only if
       # modifications were made.
       def resource_updated(resource, action)
       end
+
+      # Called when a resource has no converge actions, e.g., it was already correct.
+      def resource_up_to_date(resource, action)
+      end
+
+      # Called when a resource action has been completed
+      def resource_completed(resource)
+      end
+
 
       # A stream has opened.
       def stream_opened(stream, options = {})
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -282,7 +282,7 @@ class Chef
       # 4. Check if why-run safe
       #    - resource_bypassed: (goto 7) if not why-run safe
       # 5. During processing:
-      #    - resource_update_applied: For each actual change
+      #    - resource_update_applied: For each actual change (many per action)
       # 6. Processing complete status:
       #    - resource_failed if the resource threw an exception while running
       #    - resource_failed_retriable: (goto 3) if resource failed and will be retried
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -200,12 +200,14 @@ class Chef
       # manipulating notifies.
 
       converge_by ("evaluate block and run any associated actions") do
-        saved_run_context = @run_context
-        @run_context = @run_context.dup
-        @run_context.resource_collection = Chef::ResourceCollection.new
-        instance_eval(&block)
-        Chef::Runner.new(@run_context).converge
-        @run_context = saved_run_context
+        saved_run_context = run_context
+        begin
+          @run_context = run_context.create_child
+          instance_eval(&block)
+          Chef::Runner.new(run_context).converge
+        ensure
+          @run_context = saved_run_context
+        end
       end
     end
 
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -26,6 +26,7 @@ require 'chef/mixin/powershell_out'
 require 'chef/mixin/provides'
 require 'chef/platform/service_helpers'
 require 'chef/node_map'
+require 'forwardable'
 
 class Chef
   class Provider
@@ -65,6 +66,7 @@ class Chef
 
       @recipe_name = nil
       @cookbook_name = nil
+      self.class.include_resource_dsl_module(new_resource)
     end
 
     def whyrun_mode?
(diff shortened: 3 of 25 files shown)


lib/chef/dsl/recipe.rb
```diff
@@ -176,6 +176,24 @@ def method_missing(method_symbol, *args, &block)
           raise NameError, "No resource, method, or local variable named `#{method_symbol}' for #{describe_self_for_error}"
         end
       end
+
+      module Everything
```
tyler-ball: Can we call this `Dependencies` or something more descriptive?

lib/chef/event_dispatch/base.rb
```diff
@@ -269,26 +269,37 @@ def control_example_failure(control_group_name, example_data, error)
...
+      #    - resource_action_start
+      # 2. Check the guard
+      #    - resource_skipped: (goto 7) if only_if/not_if say to skip
+      # 3. Load the current resource
+      #    - resource_current_state_loaded
+      #    - resource_current_state_load_bypassed (if not why-run safe)
+      # 4. Check if why-run safe
+      #    - resource_bypassed: (goto 7) if not why-run safe
+      # 5. During processing:
+      #    - resource_update_applied: For each actual change
```
tyler-ball: Is this called for each action?  Or for each resource in an action?

lib/chef/provider.rb
```diff
@@ -183,6 +183,116 @@ def self.provides?(node, resource)
       Chef::ProviderResolver.new(node, resource, :nothing).provided_by?(self)
     end
 
+    #
+    # Include attributes, public and protected methods from this Resource in
+    # the provider.  Will delegate to
```
tyler-ball: `Will delegate to` whom?

lib/chef/provider.rb
```diff
@@ -200,12 +310,14 @@ def recipe_eval(&block)
...
       converge_by ("evaluate block and run any associated actions") do
-        saved_run_context = @run_context
-        @run_context = @run_context.dup
-        @run_context.resource_collection = Chef::ResourceCollection.new
-        instance_eval(&block)
-        Chef::Runner.new(@run_context).converge
-        @run_context = saved_run_context
+        saved_run_context = run_context
+        begin
+          @run_context = run_context.create_child
```
tyler-ball: Does `create_child` create an empty resource collection or does it keep all the existing resources?

lib/chef/provider.rb
```diff
@@ -183,6 +183,116 @@ def self.provides?(node, resource)
...
+    #
+    # Include attributes, public and protected methods from this Resource in
+    # the provider.  Will delegate to
+    #
+    # The actual include does not happen until the first time the Provider
+    # is instantiated (so that we don't have to worry about load order issues).
+    #
+    # @param include_resource_dsl [Boolean] Whether to include resource DSL or
+    #   not (defaults to `false`).
+    def self.include_resource_dsl(include_resource_dsl)
```
tyler-ball: Can you explain what the idea of the resource DSL is?  I think its only the `action` keyword at this point, right?

lib/chef/provider.rb
```diff
@@ -183,6 +183,116 @@ def self.provides?(node, resource)
...
+      @include_resource_dsl = include_resource_dsl
+    end
+
+    # Create the resource DSL module that forwards resource methods to new_resource
+    #
+    # @api private
+    def self.include_resource_dsl_module(resource)
```
tyler-ball: It would be nice to see a test that validates that the action_provider instance has all the wrapping resource methods

lib/chef/resource.rb
```diff
@@ -1012,6 +1010,75 @@ def default_action(action_name=NULL_ARG)
...
+      # Resources in the action recipe may notify and subscribe to other resources
+      # within the action recipe, but cannot notify or subscribe to resources
+      # in the main Chef run.
+      #
+      # Resource actions are *inheritable*: if resource A defines `action :create`
+      # and B is a subclass of A, B gets all of A's actions.  Additionally,
+      # resource B can define `action :create` and call `super()` to invoke A's
+      # action code.
+      #
+      # @param name [Symbol] The action name to define.
```
tyler-ball: I think you should update this comment to include `The first non-:nothing action defined is the default action unless the default action is explicitly defined`.  You had some good wording for this somewhere...

lib/chef/resource.rb
```diff
@@ -1012,6 +1010,75 @@ def default_action(action_name=NULL_ARG)
...
+      #
+      #   - The resource's public and protected methods (including attributes)
+      #   - The Chef Recipe DSL (file, etc.)
+      #   - super() referring to the parent version of the action (if any)
+      #
+      # @return The Action class implementing the action
+      #
+      def action(action, &recipe_block)
+        action = action.to_sym
+        create_action_provider_class.action(action, &recipe_block)
```
tyler-ball: It feels really weird to be calling `create_action_provider_class` here, but I think thats mostly because there are so many `:create` actions.

Can the `action_provider_class` be updated to be better about memoizing the `@action_provider_class` and the `create_action_provider_class` then inlined there?

lib/chef/resource.rb
```diff
@@ -1012,6 +1010,75 @@ def default_action(action_name=NULL_ARG)
...
+      end
+
+      #
+      # The created action provider class for this resource, or nil if it has
+      # not been / does not need to be created.
+      #
+      # @api private
+      def action_provider_class
+        @action_provider_class ||
+          # If the superclass needed one, then we need one as well.
```
tyler-ball: This comment doesn't explain _why_.  When would we not need a `action_provider_class`?

lib/chef/resource.rb
```diff
@@ -1012,6 +1010,75 @@ def default_action(action_name=NULL_ARG)
...
+          if superclass.respond_to?(:action_provider_class) && superclass.action_provider_class
```
tyler-ball: When would `superclass` not be a `Chef::Resource`?

lib/chef/resource.rb
```diff
@@ -1169,7 +1236,9 @@ def validate_action(action)
     end
 
     def provider_for_action(action)
-      provider = Chef::ProviderResolver.new(node, self, action).resolve.new(self, run_context)
+      provider_class = self.class.action_provider_class
+      provider_class ||= Chef::ProviderResolver.new(node, self, action).resolve
```
tyler-ball: We should definitely document that using the `action` keyword in the resource means it will no longer do a provider lookup using the ProviderResolver.  I think this behavior makes sense, though.

lib/chef/resource/reboot.rb
```diff
@@ -28,6 +28,8 @@ class Reboot < Chef::Resource
 
       allowed_actions :request_reboot, :reboot_now, :cancel
 
+      allowed_actions :request_reboot, :reboot_now, :cancel
```
tyler-ball: Looks like copypasta error

lib/chef/run_context.rb
```diff
@@ -25,118 +25,228 @@
...
       nr = notification.notifying_resource
       if nr.instance_of?(Chef::Resource)
-        @immediate_notification_collection[nr.name] << notification
+        # TODO is there any point at all to keying on name?  Do we really want
+        # to categorize notifications from execute[do it] with file[do it]
+        # and directory[do it]?
+        immediate_notification_collection[nr.name] << notification
       else
-        @immediate_notification_collection[nr.declared_key] << notification
+        # TODO this is only declared on Chef::Resource.  Does it even run?
```
tyler-ball: Well, I did have unit tests which explained why we had this, but apparently I never merged them to master.  I'll open up a PR to do that now.

https://github.com/chef/chef/blob/2101461d0f6f0ab8480e367dc570b3ce487b2816/spec/unit/run_context_spec.rb#L151-L191

There were times when the notifying resource was only a `Chef::Resource` and thus wouldn't have a declared_key (the string used, like `package` for `YumPackage`).  Thats why I had these different tests.  `declared_key` will look like `package[my_package]`.

tyler-ball:
I would expect to see more updates to `provider_spec.rb`

lib/chef/event_dispatch/base.rb
jkeiser: For each update that happens during an action.  (many per action)

lib/chef/dsl/recipe.rb
jkeiser: Not sure what the right word is.  This is the full recipe DSL (rather than a partial one, which is what Chef::DSL::Recipe is right now).

lib/chef/provider.rb
jkeiser: It is all attributes and public and protected methods on the Resource.  Updating it to this, which perhaps makes more sense:

```ruby
    #
    # Include attributes, public and protected methods from this Resource in
    # the provider.
    #
    # If this is set to true, delegate methods are included in the provider so
    # that you can call (for example) `attrname` and it will call
    # `new_resource.attrname`.
    #
    # The actual include does not happen until the first time the Provider
    # is instantiated (so that we don't have to worry about load order issues).
    #
    # @param include_resource_dsl [Boolean] Whether to include resource DSL or
    #   not (defaults to `false`).
    #

File: lib/chef/provider.rb (hunk 6 above)
Comment (jkeiser) [inline lib/chef/provider.rb]: That would be here: https://github.com/chef/chef/blob/jk/resource_actions/spec/integration/recipes/resource_action_spec.rb#L38-L92

File: lib/chef/resource.rb (hunk 7 above)
Comment (jkeiser) [inline lib/chef/resource.rb]: How about:

```ruby
    # The first action defined (besides `:nothing`) will become the default
    # action for the resource.
```

lib/chef/resource.rb
jkeiser: The issue is we need some kind of flag so that `self.action` can say "no really, we need an `action_provider_class` now."  `action_provider_class` itself will return `nil` unless one is already there (or if the superclass already has one).  I split it into two methods because I usually try to shy away from behavior flags.

lib/chef/resource.rb
jkeiser: How about:

```ruby
    #
    # The action provider class is an automatic `Provider` created to handle
    # actions declared by `action :x do ... end`.
    #
    # This class will be returned by `resource.provider` if `resource.provider`
    # is not set. `provider_for_action` will also use this by default.
    #
    # If the user has not declared actions on this class or its superclasses
    # using `action :x do ... end`, then there is no need for this class and
    # `action_provider_class` will be `nil`.
    #
    # @api private
    #
```
As for the second question, Resources can inherit from each other--it feels like this is not the right place to talk about that specifically, though.

jkeiser:
@tyler-ball what would you expect to see in provider_spec.rb and resource_spec.rb?

lib/chef/run_context.rb
jkeiser: Oh!  `instance_of?` literally means "you are an instance of `Chef::Resource` and not a subclass thereof?"  What were the circumstances where that happens?  Something still feels wrong (like, I don't get how we can instantiate something with no resource_name--in fact, I think I just disabled that ability).  Let's chat, I think I might still not understand but I'm getting closer :)

lib/chef/dsl/recipe.rb
tyler-ball: I don't care what the right word is, but `Everything` isn't it - anyone else have an opinion?

lib/chef/provider.rb
tyler-ball: :+1:

lib/chef/resource.rb
tyler-ball: :+1:

lib/chef/resource.rb
tyler-ball: :+1: to the comment - much more descriptive

lib/chef/run_context.rb
tyler-ball: http://stackoverflow.com/a/3893305/2570231

Yeah, hit me up and lets chat about this!

lib/chef/resource.rb
tyler-ball: I think I only have an issue with the `create` word - what about `new_action_provider_class`?

lib/chef/provider.rb
jkeiser: Empty resource collection.  The idea is you are entering a new converge.

lib/chef/resource.rb
jkeiser: Okey.  Changed to that :)

lib/chef/resource.rb
jkeiser: Actually, since we're setting `def provider` we don't really need to do that (same behavior, no actual resolution even though it calls the Resolver), so I'm going to remove that line.  Thanks!

lib/chef/run_context.rb
jkeiser: I'll remove the note for now--my question is more, "in what situations are you really going to be instantiating `Chef::Resource` directly?"

lib/chef/run_context.rb
```diff
@@ -198,76 +337,152 @@ def resolve_attribute(cookbook_name, attr_file_name)
...
-    # An Array of all recipes that have been loaded. This is stored internally
-    # as a Hash, so ordering is predictable.
     #
-    # Recipe names are given in fully qualified form, e.g., the recipe "nginx"
-    # will be given as "nginx::default"
+    # A list of all recipes that have been loaded.
+    #
+    # This is stored internally as a Hash, so ordering is predictable.
+    #
+    # TODO is the above statement true in a 1.9+ ruby world?  Is it relevant?
```
tyler-ball: In Ruby 2.0+ `Hashes enumerate their values in the order that the corresponding keys were inserted.` - http://ruby-doc.org/core-2.0.0/Hash.html

Your statement is true and this TODO can be removed.

lib/chef/run_context.rb
```diff
@@ -198,76 +337,152 @@ def resolve_attribute(cookbook_name, attr_file_name)
...
-    # An Array of all attributes files that have been loaded. Stored internally
-    # using a Hash, so order is predictable.
     #
-    # Attribute file names are given in fully qualified form, e.g.,
-    # "nginx::default" instead of "nginx".
+    # A list of all attributes files that have been loaded.
+    #
+    # Stored internally using a Hash, so order is predictable.
+    #
+    # TODO is the above statement true in a 1.9+ ruby world?  Is it relevant?
```
tyler-ball: Ditto to https://github.com/chef/chef/pull/3437/files#r32032113

tyler-ball:
Other than 2 minor comments, you have addressed everything I commented on, so I am :+1:

Issue #3010, opened 2015-03-03
Knife user creation doesn't work in Chef 12


luckymike:
```
knife user create foo --admin -p xxxxxxxxxxxxxxxx
```
Opens an editor with the following:
```
{
  "name": "foo",
  "public_key": null,
  "admin": true,
  "password": "xxxxxxxxxxxxxxxx"
}
```   
Saving gives this error:
```                                          
ERROR: The data in your request was invalid
Response: Field 'username' missing
```
Changing `name` to `username` then saving results in the same error.

joerg:
Same here. I also tried setting "name" and "username"...

btm:
Does this work for any version of Chef 12? That is, did 12.0.3 work, or just 11.x.x?

stevendanna:
The short answer: You probably want to either (1) use chef-server-ctl user-create or (2) temporarily configure your knife client to such that the chef_server_url is the root of the chef server (no "organization/ORGNAME" in the URL) and the client is pivotal.

In Chef 11 Server: This command should work as expected.

In Chef 12 Server: Whether `knife user create` will work depends on the user making the request and the chef_server_url.  Chef 12 has the following user endpoints that expect POST:

- `POST /users`:

This endpoint creates a new user.  With `chef_server_url` set to the root of your Chef server, and using the pivotal user, this command should work without modification.

- `POST /organizations/ORGNAME/users`:

This endpoint force-associates an *existing* user with an organization (without the user accepting the invite).  This only works for the pivotal user.  The expected POST body is:

    { "username": "USERNAME" }

`knife user create` does not post the raw JSON that you enter into the editor.  Rather, it uses it to construct a valid Chef::User object where the name field is "name" and not "username".  In fact, it looks like it will probably ignore the username field altogether and will always pass the "name" field which isn't accepted by the organization-specific version of this endpoint.

dcws:
This also doesn't work.  Then error morphs to "Field 'display name' missing.

jblaine:
I can't see how this isn't a bug, somewhere, from many perspectives, regardless of the choice of fix path. Could someone with privs add the bug label to this issue, please?

stevendanna:
@jblaine I doubt anyone disagrees with you.  I don't think the bug vs feature tags are used very reliably at the moment.  

[...]

PR #3438, opened 2015-05-27
API V1 Support

tylercloke:
## Changes in this Branch
+ Brings V1 API support to `Chef::User` and `Chef::ApiClient`.
+ API requests now default to API Version 1.
+ Brings V1, V0, and backwards compatible support to `knife user` and `knife client` (see Backwards Compat Note below for full details for `knife user`).
+ Repairs `knife user` support for `CS12` (fixing https://github.com/chef/chef/issues/3010 even though only the Superuser can create users in non-OSC).
+ **Vastly** updates and improves `Chef::User`, `Chef::ApiClient`, `knife user`, and `knife client` spec tests (although there is still work to do there, especially on the client side of things).
+ Moves original `Chef::User` object that only supported `OSC` to `Chef::OscUser` namespace, with `CS12` supported object now occupying `Chef::User` namespace.
+ Moves original `knife user` commands that only supported `OSC` to `knife osc_user`, but with `knife user` support for detecting `OCS 11` servers with proper forwarding to `knife osc_user` in that case.
+ Lots of comments on how to clean up **all** the User code once `OSC 11` support is dropped (which will greatly simplify code).

## The Gritty Details

Ping @chef/lob @danielsdeleo

**`osc_*.rb` files do not need review. They are the old version of the files I modified. See the note on backwards compatibility.**

Note: I generally like to work in small batch sizes, but this is a significant change that I think should be discussed all at once, so apologies to reviewers for the giant PR size.

### TODO
- [x] User V1 Support
- [x] Client V1 Support
- [x] Make sure `Chef::OscUser` migration didn't break any external code.

### Manual testing
- [x] Client on CS12
- [x] User on CS12
- [x] Client on OSC 11
- [x] User on OSC 11
- [x] Client on EC 11
- [x] User on EC 11
- [x] Pul into CS12 and run pedant ([passing build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/567/downstreambuildview/))
- [ ] Test OSC 11 -> CS 12 containing this code upgrade

### Open Source 11 Server Backwards Compatibility Note

Based on feedback from @stevendanna and @jdmundrawala, I made the `knife user` commands backwards compatible in this [commit](https://github.com/chef/chef/commit/768ef158fc6666dddf2fea7f8fda4613f5b31ab7).

Now, all the `knife user` commands are smart enough to detect if an `OCS 11` server was used, and if so it will:

1) Print a warning message about deprecated `knife user` `OSC 11` support and that the command is being forwarded to `knife osc_user`.
2) Forward the request to the proper `knife osc_user` command.

So essentially, `knife user` is fully backwards compatible with `OSC 11`. It just prints a warning message to the user in that case.


danielsdeleo:
@chef/client-core

tyler-ball:
:+1:

lib/chef/user.rb
```diff
@@ -144,12 +287,17 @@ class << self
     end
 
     def self.list(inflate=false)
-      response = Chef::REST.new(Chef::Config[:chef_server_url]).get_rest('users')
-      users = if response.is_a?(Array)
```
stevendanna: will this still work against an 11 server?

tylercloke: Just added this commit for backwards compatibility support: https://github.com/chef/chef/commit/55834783ee26a53cfad113433a57c691b3cbc114

I also just wrote up my thoughts and what I actually changed in at the top of this PR. Please read and let me know what you think.

tylercloke: I'll probably end up emailing this PR to dev or the client team or something. Definitely want a lot of review on this. Would like to nail this API versioning stuff the first time.

lib/chef/versioned_rest.rb
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module VersionedRest
+    # Helper for getting a sane interface to passing an API version to Chef::REST
+    # api_version should be a string of an integer
+    def get_versioned_rest_object(url, api_version)
+      Chef::REST.new(url, Chef::Config[:node_name], Chef::Config[:client_key], {:api_version => api_version})
```
danielsdeleo: Can we use this instead? https://github.com/chef/chef/blob/master/lib/chef/server_api.rb

`Chef::REST` does the magical object inflation thing where it auto-converts JSON responses into Chef model objects. I think as we touch each of the different API consumer classes it'd be good to stop relying on that behavior.

lib/chef/user.rb
danielsdeleo: We should at least figure out what happens if you run the new `knife user create` against CS 11. There's a chance the HTTP request will just work, in which case a little bit more effort should get us to a decent place compatibility-wise. If it doesn't work, then we should handle that error and be very clear about what the user needs to do instead.

lib/chef/knife/user_create.rb
```diff
@@ -30,61 +33,72 @@ class UserCreate < Knife
...
 
-        if @user_name.nil?
-          show_usage
-          ui.fatal("You must specify a user name")
-          exit 1
-        end
+        test_mandatory_field(@name_args[1], "display name")
+        user.display_name @name_args[1]
+
+        test_mandatory_field(@name_args[2], "first name")
```
jdmundrawala: So instead of failing here if the rest are not specified, could we just assume that we need to use osc_user and call that. That way, we don't break backwards compatibility. We could also spit out a deprecation warning of some sort.

lib/chef/versioned_rest.rb
tylercloke: Yeah, totally. I just wanted to be able to pass options easily without specifying `Chef::Config[:node_name], Chef::Config[:client_key` every time.

lib/chef/user.rb
tylercloke: Yeah, Jay gave a great suggestion on how to handle `knife user create` for CS 11 that I will implement. Basically, if you only pass the one arg (username) it will just try against `Chef::OscUser` with a warning that the command has been moved to `knife osc_user create`.

stevendanna: It might also be possible to choose by introspecting on chef_server_url.  Either way, being able to have a single command would be a big win I think.

tylercloke: All `knife user` commands are now backwards compatible, see https://github.com/chef/chef/commit/768ef158fc6666dddf2fea7f8fda4613f5b31ab7 and the PR description for details.

lib/chef/api_client.rb
```diff
@@ -1,7 +1,7 @@
 #
-# Author:: Adam Jacob (<adam@opscode.com>)
-# Author:: Nuo Yan (<nuo@opscode.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Author:: Adam Jacob (<adam@chef.io>)
+# Author:: Nuo Yan (<nuo@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
btm: ```Copyright:: Copyright (c) 2008,2015 Chef Software, Inc.```

I'm not sure exactly what relationship the old company has with the new one, but since we're not replacing the entire file the old copyright still applies.

tylercloke: Ah mmmk, I'll update all my updates then.

lib/chef/api_client.rb
```diff
@@ -37,6 +43,25 @@ def initialize
...
+
+    def chef_rest_v0
+      @chef_rest_v0 ||= Chef::ServerAPI.new(Chef::Config[:chef_server_url], {:api_version => "0"})
+    end
+
+    def chef_rest_v1
+      @chef_rest_v1 ||= Chef::ServerAPI.new(Chef::Config[:chef_server_url], {:api_version => "1"})
+    end
+
+    # will default to the current version
```
btm: What does this mean? The "latest" version? Where is that set?

lib/chef/api_client.rb
```diff
@@ -194,18 +230,93 @@ def save
...
+      # it was never implemented, we will simply ignore that functionality
+      # as it is being deprecated.
+      # Delete this comment after V0 support is dropped.
+      payload = { :name => name }
+      payload[:validator] = validator unless validator.nil?
+
+      # DEPRECATION
+      # This field is ignored in API V1, but left for backwards-compat,
+      # can remove after API V0 is no longer supported.
+      payload[:admin] = admin unless admin.nil?
```
btm: admin is either true or false, and should never be nil. If it ever could be, we'd want the same guard in #create?

lib/chef/api_client.rb
```diff
@@ -194,18 +230,93 @@ def save
...
+        # get the private_key out of the chef_key hash if it exists
+        if new_client['chef_key']
+          if new_client['chef_key']['private_key']
+            new_client['private_key'] = new_client['chef_key']['private_key']
+          end
+          new_client['public_key'] = new_client['chef_key']['public_key']
+          new_client.delete('chef_key')
+        end
+
+        rescue Net::HTTPServerException => e
```
btm: You've got an extra level of indentation in this rescue block.

lib/chef/api_client.rb
```diff
@@ -194,18 +230,93 @@ def save
...
-      http_api.post("clients", self)
+      payload = {
+        :name => name,
+        :validator => validator,
+        # this field is ignored in API V1, but left for backwards-compat,
+        # can remove after OSC 11 support is finished?
+        :admin => admin
+      }
+      begin
+        # try API V1
```
btm: where do we try API V0? I see some notes below about rescuing V0. I would expect some clear delineation here as to what code is part of which protocol version, maybe even factored out to separate create_v0 and create_v1 methods.

lib/chef/formatters/error_inspectors/api_error_formatting.rb
```diff
@@ -68,13 +68,16 @@ def describe_400_error(error_description)
...
+          version_header = Chef::JSONCompat.from_json(response["x-ops-server-api-version"])
+          client_api_version = version_header["request_version"]
+          min_server_version = version_header["min_version"]
+          max_server_version = version_header["max_version"]
+
           error_description.section("Incompatible server API version:",<<-E)
-This version of Chef is not supported by the Chef server you sent this request to
-This version of Chef requires a server API version of #{Chef::HTTP::Authenticator::SERVER_API_VERSION}
-The Chef server you sent the request to supports a min API version of #{min_version} and a max API version of #{max_version}
+This version of the API that this Chef request specified is not supported by the Chef server you sent this request to
```
btm: The version?

btm: Period at the end of the line.

lib/chef/formatters/error_inspectors/api_error_formatting.rb
```diff
@@ -68,13 +68,16 @@ def describe_400_error(error_description)
...
+The Chef server you sent the request to supports a min API version of #{min_server_version} and a max API version of #{max_server_version}
```
btm: Given the above line, this could probably be shorted to "The server supports..."

lib/chef/knife/client_create.rb
```diff
@@ -28,58 +28,81 @@ class ClientCreate < Knife
...
-        :boolean => true
+             :long  => "--validator",
+             :description => "Create the client as a validator.",
+             :boolean => true
 
-      banner "knife client create CLIENT (options)"
+      option :public_key,
+             :short => "-p FILE",
+             :long  => "--public-key",
+             :description => "Set the initial default key for the client from a file on disk (cannot pass with --create-key)."
```
btm:
```
(cannot pass with --create-key)
```

there is no --create-key option.

lib/chef/knife/osc_user_create.rb
```diff
@@ -0,0 +1,98 @@
...
+# limitations under the License.
+#
+
+require 'chef/knife'
+
+# DEPRECATION NOTE
+# This code only remains to support users still operating with
+# Open Source Chef Server 11 and should be removed once support
+# for OSC 11 ends. New development should occur in the Chef Server 12
+# version of this file.
```
btm: New development should occur in client_create.rb.

lib/chef/knife/osc_user_delete.rb
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+#
+
+require 'chef/knife'
+
+# DEPRECATION NOTE
+# This code only remains to support users still	operating with
+# Open Source Chef Server 11 and should	  be removed once support
+# for OSC 11 ends. New development should occur in the Chef Server 12
+# version of this file.
```
btm: there's a couple tabs in there. Also I think there's just "the normal file" not the "CS12" version, so "New development should occur in user_delete.rb"  ?

lib/chef/knife/osc_user_edit.rb
```diff
@@ -0,0 +1,59 @@
...
+# limitations under the License.
+#
+
+require 'chef/knife'
+
+# DEPRECATION NOTE
+# This code only remains to support users still	operating with
+# Open Source Chef Server 11 and should	  be removed once support
+# for OSC 11 ends. New development should occur in the Chef Server 12
+# version of this file.
```
btm: Same as osc_user_delete.rb

lib/chef/knife/user_create.rb
```diff
@@ -30,63 +34,121 @@ class UserCreate < Knife
...
+        @user_field ||= Chef::User.new
+      end
+
+      def create_user_from_hash(hash)
+        Chef::User.from_hash(hash).create
+      end
+
+      def osc_11_warning
+<<-EOF
+IF YOU ARE USING CHEF SERVER 12, PLEASE FOLLOW THE INSTUCTIONS UNDER knife user create --help.
```
btm: CHEF SERVER 12+
Typo in INSTRUCTIONS.

lib/chef/knife/user_edit.rb
```diff
@@ -29,6 +29,21 @@ class UserEdit < Knife
 
       banner "knife user edit USER (options)"
 
+      def osc_11_warning
+<<-EOF
+The Chef Sever you are using does not support the username field.
```
btm: Chef Server

lib/chef/knife/user_edit.rb
```diff
@@ -29,6 +29,21 @@ class UserEdit < Knife
...
+This means it is an Open Source 11 Server.
+knife user edit for Open Source 11 Server is being deprecated.
+Open Source 11 Server user commands now live under the knife opc_user namespace.
```
btm: knife osc_user

lib/chef/knife/user_list.rb
```diff
@@ -34,8 +34,35 @@ class UserList < Knife
         :long => "--with-uri",
         :description => "Show corresponding URIs"
 
+      def osc_11_warning
+<<-EOF
+knife user list failed in a way that indicates that you are using an Open Source 11 Server.
+knife user list for Open Source 11 Server is being deprecated.
+Open Source 11 Server user commands now live under the knife opc_user namespace.
```
btm: knife osc_user

btm: this error is copy-pasted around, I'm going to stop mentioning that. :)

lib/chef/mixin/api_version_request_handling.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+      true
+    end
+
+    def reregister_only_v0_supported_error_msg(max_version, min_version)
+<<-EOH
+The reregister command only supports server API version 0.
+The server that received the request supports a min version of #{min_version} and a max version of #{max_version}.
+User keys are now managed via the key rotation commmands.
+Please refer to the documentation on how to manage your keys via the key rotation commands.
```
btm: This documentation? https://docs.chef.io/server_security.html#key-rotation

btm:
:+1: to big PR.

lib/chef/api_client.rb
tylercloke: https://github.com/chef/chef/blob/tc/api-versioning/lib/chef/http/authenticator.rb#L27

lib/chef/api_client.rb
tylercloke: Well only specific versions of the API support `admin`, so it would be nil in that case.

lib/chef/api_client.rb
tylercloke: It tries it in the rescue block right below:

https://github.com/chef/chef/pull/3438/files#diff-e8f47cdcac29663d504b3f119e0d53f7R317

lib/chef/knife/osc_user_delete.rb
tylercloke: Ugh, that tabbing is strange. Looks fine in my editor.

lib/chef/knife/user_list.rb
tylercloke: Yeah, in general I'm not down with the copy-pasta and I would have used a smarter solution, but this code is going away in a few months when 11 support stops, so I didn't wanna put too much effort into the parts that were going away. Hard to balance but the stuff that is sticking around I tried to make as high-quality as possible (with many many tests).

lib/chef/knife/user_create.rb
```diff
@@ -30,63 +34,121 @@ class UserCreate < Knife
...
+You only passed a single argument to knife user create.
+For backwards compatibility, when only a single argument is passed,
+knife user create assumes you want Open Source 11 Server user creation.
+knife user create for Open Source 11 Server is being deprecated.
+Open Source 11 Server user commands now live under the knife opc_user namespace.
```
danielsdeleo: "knife osc_user namespace" ?

lib/chef/knife/user_delete.rb
```diff
@@ -29,6 +29,37 @@ class UserDelete < Knife
 
       banner "knife user delete USER (options)"
 
+      def osc_11_warning
+<<-EOF
+The Chef Sever you are using does not support the username field.
+This means it is an Open Source 11 Server.
+knife user delete for Open Source 11 Server is being deprecated.
+Open Source 11 Server user commands now live under the knife opc_user namespace.
```
danielsdeleo: More `opc_user`

lib/chef/mixin/api_version_request_handling.rb
```diff
@@ -0,0 +1,59 @@
...
+  module ApiVersionRequestHandling
+    # takes in an http exception, and a min and max supported API version and
+    # handles all the versioning cases
+    #
+    # it will return false if there was a non-versioning related error
+    # or the server and the client are not compatible
+    #
+    # if the server does not support versioning, then it will return true, and you
+    # can assume API v0 is safe to send
+    def handle_version_http_exception(exception, min_client_supported_version, max_client_supported_version)
```
danielsdeleo: Maybe could be named `should_retry_with_lower_api_version` or something like that?

lib/chef/mixin/api_version_request_handling.rb
```diff
@@ -0,0 +1,59 @@
...
+        max_server_version = Integer(header["max_version"])
+
+        # if the min API version the server supports is greater than the min version the client supports
+        # and the max API version the server supports is less than the max version the client supports
+        if min_server_version > min_client_supported_version || max_server_version < max_client_supported_version
+          # if it had x-ops-server-api-version header, return false
+          return false
+        end
+      end
+      true
```
danielsdeleo: Should this be the highest supported API version so you can write `server_api_version = should_retry_at_api_version` or whatever?

lib/chef/knife/user_create.rb
```diff
@@ -30,63 +34,121 @@ class UserCreate < Knife
...
         :long  => "--admin",
-        :description => "Create the user as an admin",
+        :description => "DEPRECATED: Open Source Chef 11 only. Create the user as an admin.",
         :boolean => true
 
       option :user_password,
         :short => "-p PASSWORD",
         :long => "--password PASSWORD",
-        :description => "Password for newly created user",
+        :description => "DEPRECATED: Open Source Chef 11 only. Password for newly created user.",
```
marcparadise: Password is supported (possibly even required) at time of user creation in EC11 & CS12.

marcparadise: Nevermind, I just saw the rest of this :)

marcparadise:
One thing we talked about that I don't think I saw here is the scenario where the server is 11.x or <=12.0.7.  In this case a request may succeed when we send x-ops-api-version:1 because the server is ignoring/unaware of that field - but the results may not be what is expected.  

The example I can think of is POST to users with create_key = true. create_key would be ignored since we don't recognize it prior to v1, but we will not have specified private_key = true. The result here would be no key being generated when one was expected. 

I think the client counterpart would be safe, because the default behavior is to generate a new key, even if we don't pass private_key: true. 

Other than that, this looks good to me.

tylercloke:
@marcparadise All of the `knife user` commands except `knife user list` are forwarding to `Chef::OscUser` in the `OSC 11` case. `Chef::User` is generally not backwards-compatible with `OSC 11`, but `knife user` is. `EC 11` never worked with `Chef::User` creation / `knife user create`, so fixing it is a separate issue (that we will probably never tackle, since it isn't that useful as you have to be the superuser). For example, using the newest released `knife` against latest `EC 11` as pivotal:

```
$ knife --version
Chef: 12.3.0
$ knife user create testuser -p password
USAGE: knife user create USER (options)
    -a, --admin                      Create the user as an admin
<...more usage...>
ERROR: The data in your request was invalid
Response: Display name must not be blank, Username must not be blank, Username has an invalid format (valid characters are a-z, 0-9, hyphen and underscore), Username is too short (minimum is 1 characters), Email must not be blank
```

Everything that `knife user` used to support in `EC` still works, but I didn't fix what was already broken for `EC`.

`Chef::ApiClient` should work with everything. At least that was the intension. I did manual testing on OSC 11, EC 11, and CS 12 for `knife client`. I think this is good to go.

marcparadise:
:+1:

lib/chef/knife/user_create.rb
```diff
@@ -30,63 +34,121 @@ class UserCreate < Knife
...
+<<-EOF
+IF YOU ARE USING CHEF SERVER 12+, PLEASE FOLLOW THE INSTRUCTIONS UNDER knife user create --help.
+You only passed a single argument to knife user create.
+For backwards compatibility, when only a single argument is passed,
+knife user create assumes you want Open Source 11 Server user creation.
+knife user create for Open Source 11 Server is being deprecated.
+Open Source 11 Server user commands now live under the knife osc_user namespace.
+For backwards compatibility, we will forward this request to knife osc_user create.
+If you are using an Open Source 11 Server, please use that command to avoid this warning.
+If you are not using an Open Source Chef 11 Server install, please read knife user create --help for proper usage.
```
jdmundrawala: i feel like you already told me this

lib/chef/knife/user_create.rb
```diff
@@ -30,63 +34,121 @@ class UserCreate < Knife
...
+        if @name_args.length == 1
+          ui.warn(osc_11_warning)
 
-        if @user_name.nil?
-          show_usage
-          ui.fatal("You must specify a user name")
-          exit 1
-        end
+          # run osc_user_create with our input
+          ARGV.delete("user")
```
jdmundrawala: These modifiecations to `ARGV` might read better if they are inside `run_osc_11_user_create`

tylercloke: That's a good change.

lib/chef/knife/user_create.rb
tylercloke: Yeah probably a bit redundant. Trying to be super super clear for new users but might be overkill.

Issue #2175, opened 2014-10-06
Resources run longer than timeout length 


Tech356:
### Version:

Chef Client 11.14.6
### Environment:

Windows 7 x64
### Scenario:

Use the `timeout` attribute to set a timeout on a resource
### Steps to Reproduce:

Set a timeout on an resource and have the resource run longer than the timeout
### Expected Result:

The resource should timeout after the amount of seconds set by the timeout attribute
### Actual Result:

The resource times out a significant amount of time after the the length of the timeout. In my example below it took 1014 seconds (17 minutes) for the resource to timeout when I set the timeout to 10 seconds. I have also seen this with the `windows_package` resource, which I assume uses the same timeout code as the `execute` resource.

Here is the console output:

```
C:\Windows\system32>chef-client -o "test"
Starting Chef Client, version 11.14.6
resolving cookbooks for run list: ["test"]
Synchronizing Cookbooks:
  - test
Compiling Cookbooks...
Converging 1 resources
Recipe: test::default
  * execute[ping 127.0.0.1 -n 1000 > nul] action run

    ================================================================================
    Error executing action `run` on resource 'execute[ping 127.0.0.1 -n 1000 > nul]'
    ================================================================================

    Mixlib::ShellOut::CommandTimeout
    --------------------------------
    command timed out:
    ---- Begin output of ping 127.0.0.1 -n 1000 > nul ----
    STDOUT:
    STDERR:
    ---- End output of ping 127.0.0.1 -n 1000 > nul ----

    Resource Declaration:
    ---------------------
    # In C:/chef/cache/cookbooks/test/recipes/default.rb

     10: execute "ping 127.0.0.1 -n 1000 > nul" do
     11:     timeout 10
     12: end
     13:

    Compiled Resource:
    ------------------
    # Declared in C:/chef/cache/cookbooks/test/recipes/default.rb:10:in `from_file'

    execute("ping 127.0.0.1 -n 1000 > nul") do
      action "run"
      retries 0
      retry_delay 2
      guard_interpreter :default
      command "ping 127.0.0.1 -n 1000 > nul"
      backup 5
      returns 0
      timeout 10
      cookbook_name "test"
      recipe_name "default"
    end


Running handlers:
Running handlers complete
Chef Client failed. 0 resources updated in 1065.033792 seconds
```

Here is the log output:

```
[2014-10-06T15:57:26-04:00] INFO: *** Chef 11.14.6 ***
[...]

PR #3440, opened 2015-05-27
Reduce the levels of timeouts in execute timeout test

btm:
Issue #2175 found that execute resource timeouts weren't working on Windows
but we were waiting 600 seconds for the test to fail. #2686 added a Timeout
to make sure we didn't run too long, but this option simplifies the test by
reducing the number of timeouts involved.

We're currently seeing intermittant failures on a FreeBSD tester, which is
likely a performance issue. But it's possible that we're seeing another issue
that we don't yet understand, so I don't want to throw away the regression
test yet.

### DIFF
--- a/spec/functional/resource/execute_spec.rb
+++ b/spec/functional/resource/execute_spec.rb
@@ -137,9 +137,16 @@ describe Chef::Resource::Execute do
     end
   end
 
+  # Ensure that CommandTimeout is raised, and is caused by resource.timeout really expiring.
+  # https://github.com/chef/chef/issues/2985
+  #
+  # resource.timeout should be short, this is what we're testing
+  # resource.command ruby sleep timer should be longer than resource.timeout to give us something to timeout
+  # Timeout::timeout should be longer than resource.timeout, but less than the resource.command ruby sleep timer,
+  #   so we fail if we finish on resource.command instead of resource.timeout, but raise CommandTimeout anyway (#2175).
   it "times out when a timeout is set on the resource" do
-    Timeout::timeout(5) do
-      resource.command %{ruby -e 'sleep 600'}
+    Timeout::timeout(30) do
+      resource.command %{ruby -e 'sleep 300'}
       resource.timeout 0.1
       expect { resource.run_action(:run) }.to raise_error(Mixlib::ShellOut::CommandTimeout)
     end


ranjib:
:+1: 
i wonder if the behavior will change when the timeout block is inside expect block

btm:
@lamont-granquist tried replicating the failure manually on the CI tester and couldn't, which makes it really hard/time consuming to pin it down. If the sleep timer completes and the command exits successfully, the expectation should fail. So I'm not seeing the benefit of a separate timeout block compared to lowering the sleep timer, unless we were making a choice that involved the time it takes for the internal ruby process to spin up, but I don't think we are.

The [internal tester](http://manhattan.ci.chef.co/job/chef-test/151/architecture=x86_64,platform=freebsd-9,project=chef,role=tester/) just passed. @chef/client-core review?

jdmundrawala:
I would feel better if the timeout was bumped higher or the test was disabled for freebsd. This test is kind of pointless without the timeout.

lamont-granquist:
Yeah, if it fails it hangs forever.  You can up it to 300 seconds or something silly that we should never hit, then we only burn 300 seconds when its really buggy and never ever times out.

I had to do that with some other test recently that was trolling solaris.

btm:
@jdmundrawala how is the test pointless?

a) the resource timeout passed to mixlib-shellout works properly, and raises CommandTimeout before the ruby sleep completes, meeting the expectation, and the test passes

b) something goes wrong with the resource timeout, the ruby sleep completes without raising CommandTimeout, it does not meet the expectation, and the test fails, which is the value of the test.

@lamont-granquist What's the case where it hangs forever?

Given this case:
```
 resource.command %{ruby -e 'sleep 60'}
 resource.timeout 120
```

The command completes after 60 seconds and the test correctly fails because CommandTimeout was not received. The behavior is the same if you don't pass a timeout at all in the test (simulating the timeout not getting passed).

```
  1) Chef::Resource::Execute times out when a timeout is set on the resource
     Failure/Error: expect { resource.run_action(:run) }.to raise_error(Mixlib::ShellOut::CommandTimeout)
       expected Mixlib::ShellOut::CommandTimeout but nothing was raised
     # ./spec/functional/resource/execute_spec.rb:142:in `block (2 levels) in <top (required)>'
```

I don't see the value in having both the ruby sleep timeout and and ruby Timeout wrapping it.

jdmundrawala:
Sure it raises the right exception. However it does not do the thing it is supposed to, which is to actually time out. That's why the additional check of how long it took is there.

ksubrama:
:+1:  You can lower the ruby sleep timer even further if you want.  60 seconds should be more than enough.

btm:
[fbsd 9](http://manhattan.ci.chef.co/job/chef-test/153/architecture=x86_64,platform=freebsd-9,project=chef,role=tester/) passed again. there was an unrelated [fbsd 10](http://manhattan.ci.chef.co/job/chef-test/153/architecture=x86_64,platform=freebsd-10,project=chef,role=tester/) failure we may need to look at later.

jdmundrawala:
:+1:

btm:
Merged as 678d696b169d6fbcf5966124f41847bfc3282c13

Issue #3037, opened 2015-03-06
Core Windows resources are different from Windows cookbook resources


jeremiahsnapp:
A number of Windows Chef resources are defined differently in two places, core and windows cookbook. As far as I know there isn’t any documentation that helps users understand this.

This is **very confusing and frustrating** to users who start with core resources and then add the windows cookbook to their run_list.

Here's an example related issue.
https://github.com/opscode-cookbooks/windows/issues/185

It seems that there is no solid effort to identify and attack this problem head on.

Thankfully a number of other windows pains are being worked on. It would be really nice to put some serious effort into moving the windows cookbook resources completely into core in a non-breaking way as soon as possible.

adamedx:
I think windows_package is the worst offender. There are duplicates, but that's the only one I'm aware of that causes problems with existing recipes because of naming conflicts. That conflict might be the first one we should tackle.

adamedx:
In general, here are resources that are duplicated in client and in cookbooks -- in all cases but windows_package, the core resource is not only preferred but better. We can probably start the deprecation process with those now:

* Registry: Use `registry_key` in core instead of `registry` from the Windows cookbook
* Batch: Use `batch` from core instead of `windows_batch` from the Windows cookbook.
* Reboot: Use `reboot` from core instead of `windows_reboot` from the Windows cookbook
* PowerShell: Use `powershell_script` instead of `powershell` from the PowerShell cookbook
* Package: Use `package` instead of `windows_package`. Note: if you try to use `windows_package`, core's Windows package wins, and this breaks cookbooks.

So today we should be able to deprecate `windows_reboot`, `windows_batch`, `registry`, and `powershell` from the Windows and powershell cookbooks now. The newer resources have more / better features.

jeremiahsnapp:
This is great information. Thanks @adamedx !

It would be nice to see this deprecation announced in the appropriate channels as well as in the [windows cookbook](https://github.com/opscode-cookbooks/windows) itself.

Thanks!

btm:
```Package: Use package instead of windows_package. Note: if you try to use windows_package, core's Windows package wins, and this breaks cookbooks.```

If you have the windows cookbook and you use ```windows_package``` the cookbook will win. In either case (with the windows cookbook or without) using just ```package``` on windows will use the provider in core.

sean-horn:
[...]

PR #3441, opened 2015-05-27
add powershell_out mixin

lamont-granquist:
Add powershell_out mixin to core chef

### DIFF
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -85,7 +85,7 @@ class Chef
           "-NoProfile",
           # always set the ExecutionPolicy flag
           # see http://technet.microsoft.com/en-us/library/ee176961.aspx
-          "-ExecutionPolicy RemoteSigned",
+          "-ExecutionPolicy Unrestricted",
           # Powershell will hang if STDIN is redirected
           # http://connect.microsoft.com/PowerShell/feedback/details/572313/powershell-exe-can-hang-if-stdin-is-redirected
           "-InputFormat None"
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'chef/mixin/shell_out'
+require 'chef/mixin/windows_architecture_helper'
+
 class Chef
   module Mixin
     module PowershellOut
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -1,0 +1,95 @@
+#--
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+class Chef
+  module Mixin
+    module PowershellOut
+      include Chef::Mixin::ShellOut
+      include Chef::Mixin::WindowsArchitectureHelper
+
+      # Run a command under powershell with the same API as shell_out.  The
+      # options hash is extended to take an "architecture" flag which
+      # can be set to :i386 or :x86_64 to force the windows architecture.
+      #
+      # @param script [String] script to run
+      # @param options [Hash] options hash
+      # @return [Mixlib::Shellout] mixlib-shellout object
+      def powershell_out(*command_args)
+        script = command_args.first
+        options = command_args.last.is_a?(Hash) ? command_args.last : nil
+
+        run_command_with_os_architecture(script, options)
+      end
+
+      # Run a command under powershell with the same API as shell_out!
+      # (raises exceptions on errors)
+      #
+      # @param script [String] script to run
+      # @param options [Hash] options hash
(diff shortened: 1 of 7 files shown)


lamont-granquist:
@someara @jdmundrawala @chef/client-core pulling powershell_out in from windows cookbooks.

Would like to be able to use this in core chef.  Cookbook library will need a guard.

chef-supermarket:

chef-supermarket:

lib/chef/mixin/powershell_out.rb
```diff
@@ -0,0 +1,94 @@
...
+      def powershell_out!(*command_args)
+        cmd = powershell_out(*command_args)
+        cmd.error!
+        cmd
+      end
+
+      private
+
+      def run_command_with_wow64(script, options = nil)
+        architecture = node_windows_architecture(node)
```
jdmundrawala: thoughts on not making this depend on node?

lamont-granquist: yes, but I'm not doing that today, this is enough of a yak shave.

lamont-granquist: actually i think it just fell out of the yak shave...

lamont-granquist:
okay @jdmundrawala i think the code is nicer and more stable now

jdmundrawala:
sweet :+1:

adamedx:
@lamont-granquist so what happens if someone loads the Windows (or maybe powershell) cookbook? Won't this conflict since it already has that mixin?

lamont-granquist:
It'll monkeypatch it over the top of this because its ruby, and they'll get the cookbook version.

Once this goes in, I'll submit a PR to guard the cookbook version to not overwrite the module if it already exists, and if it gets executed to warn the user that they need to upgrade their chef version to get new code.

btm:
We plan to do some similar cookbook -> core deprecation soon for #3037. Whichever of these gets done first we should leverage the other.

lamont-granquist:
Yeah, I'll aggressively deprecate the cookbook version.  This one is API compatible, so as long as there's no new additions to the cookbook it can substitute transparently and completely.  With annoying deprecation warnings and banners added to the code, along with favoring the core chef library over the cookbook we should not wind up in the windows_package situation.

ksubrama:
:+1: w00t!

adamedx:
I think the aggressive deprecation is fine then, I assume the Windows cookbook will still allow it for older versions of Chef? @btm should be signed off on the plan here. 

I want to be very careful to avoid the issue we had with name collision for `windows_package` -- its caused us a few months of confusion and support tickets. I'm fine with awkward workarounds such as the one we used for the PowerShell script resource and provider when we implemented equivalent functionality in the PowerShell cookbook within Chef -- in that case, we just used a new name `powershell_script` instead of `powershell` so there would be zero collisions and no compatibility breaks.

In this case, I think either implementation is equivalent in functionality, unlike `powershell_script` and `windows_package` cases.

adamedx:
Is there some common code that factoring that should occur with this: https://github.com/chef/chef/blob/master/lib/chef/util/powershell/cmdlet.rb ?

btm:
windows_package in core didn't have feature parity (and is still far from it), so we couldn't be aggressive about switching.

I'm :+1: on this and the plan. If the cookbook is changed to not load powershell_out when it's already found from core, and they have feature/api parity (for now), it should be seamless. People will keep getting the cookbook version until they upgrade their windows cookbook, so I think we should add a comment to both about where powershell_out came from and where it has gone to, as appropriate.

The issue with this plan would come a couple months down the road where someone changes powershell_out in core to do something new, it starts getting documented, and people using an old windows cookbook don't understand why it isn't working.

Really the only plan that avoids this situation is one where we use a different name and put deprecation messages in the cookbook.

lamont-granquist:
I think I can solve the confusion issue just by putting a Chef::Log.warn in the powershell_out routine in the cookbook that nags people to upgrade to Chef 12.whatever.0 to get the core chef one.  Can't do anything about people who don't upgrade the cookbook, but at least they should not be broken and they'll get the same behavior they had before.

lamont-granquist:
@adamedx i don't see enough common code to warrant trying to refactor, most of the common code is already handled by the windows architecture helper.

lib/chef/mixin/powershell_out.rb
```diff
@@ -0,0 +1,98 @@
...
+        flags = [
+          # Hides the copyright banner at startup.
+          "-NoLogo",
+          # Does not present an interactive prompt to the user.
+          "-NonInteractive",
+          # Does not load the Windows PowerShell profile.
+          "-NoProfile",
+          # always set the ExecutionPolicy flag
+          # see http://technet.microsoft.com/en-us/library/ee176961.aspx
+          "-ExecutionPolicy RemoteSigned",
```
adamedx: This should be changed to unrestricted. Otherwise scripts on remote servers won't be able to run. Additionally, on ps 3.0 and higher we should probably use 'bypass" as you can still run into weird issues with any script code that was downloaded from the Internet.

lamont-granquist: Do we want to change that now?  Trying to keep it identical with the cookbook version to start with.

adamedx:
Agree with your comment on refactoring @lamont-granquist, I just wanted there to be awareness there was similar code.

Separately (not in this pr), we should probably change the powershell provider to call into this code. It ostensibly calls into mixlib::shellout, and this just adds a different "mode" to mixlib::shellout.

lib/chef/mixin/powershell_out.rb
adamedx: Also unrestricted (or better!) is what we have in the existing powershell_script provider. Changing it to unrestricted won't break anything since it's more permissive.

lamont-granquist:
Okay execution policy bumped to unrestricted.  I just looked at the powershell_script provider and it is very similar but has extra code for normalizing the exit status of the command, and does not seem to have the with_os_architecture stuff.

ksubrama:
That's handled by its parent in windows_script.rb.  And Adam just fixed up a bunch of other stuff in there with https://github.com/chef/chef/pull/3080

ksubrama:
Merged in 0594046ea04be724556fcaa12d76534fbc099a42

PR #3442, opened 2015-05-27
Move resource_name to Resource, automatically populate it from `provides`

jkeiser:


lib/chef/resource.rb
```diff
@@ -871,23 +864,197 @@ def after_created
...
+    class<<self
+      #
+      # The DSL name of this resource (e.g. `package` or `yum_package`)
+      #
+      # @return [String] The DSL name of this resource.
+      #
+      # @deprecated Use resource_name instead.
+      #
+      def dsl_name
+        Chef::Log.deprecation "Resource.dsl_name is deprecated and will be removed in Chef 11.  Use resource_name instead."
```
randomcamel: This seems like the wrong log message now that we're on Chef 12.

lib/chef/resource.rb
```diff
@@ -871,23 +864,197 @@ def after_created
...
+      #
+      # Defaults to the :nothing.
+      #
+      # @param action_name [Symbol,Array<Symbol>] The default action (or series
+      #   of actions) to use.
+      #
+      # @return [Symbol,Array<Symbol>] The default actions for the resource.
+      #
+      def default_action(action_name=NULL_ARG)
+        unless action_name.equal?(NULL_ARG)
```
randomcamel: [Above](https://github.com/chef/chef/pull/3442/files#diff-135510b4e9e7ac268144bf6a7479a94dR891) (line 891) you use `!= NULL_ARG`, but here it's `#equal?`. Any reason?

jkeiser: One piece of code I wrote, another piece of code was written by someone else :)

lib/chef/resource.rb
```diff
@@ -55,6 +55,8 @@ class Resource
     # This lets user code do things like `not_if { shell_out!("command") }`
     include Chef::Mixin::ShellOut
 
+    NULL_ARG = Object.new
```
randomcamel: Why this instead of `nil`, which you use elsewhere in this patch?

lib/chef/resource.rb
```diff
@@ -871,23 +864,197 @@ def after_created
...
+
+      #
+      # The display name of this resource type, for printing purposes.
+      #
+      # @return [Symbol] The name of this resource type (e.g. `:execute`).
+      #
+      attr_accessor :resource_name
+      def resource_name(value=NULL_ARG)
+        if value != NULL_ARG
+          self.resource_name = value.to_sym
```
randomcamel: This looks like `nil` can be passed in as the resource name, which produces the dread `NoMethodError: undefined method `to_sym' for nil:NilClass` error that we should want to avoid.

lib/chef/resource.rb
jkeiser: Mostly I am not converting over old things.  Main reason to do this is it lets you distinguish between "setting something to nil" and "getting the value."

lib/chef/resource.rb
jkeiser: Making it silently fail seems worse :)

randomcamel: Keeping to our "ease of use" idea, why not have it fail with a useful error message?

lib/chef/resource/windows_package.rb
```diff
@@ -26,13 +26,12 @@ class Resource
     class WindowsPackage < Chef::Resource::Package
       include Chef::Mixin::Uris
 
-      provides :package, os: "windows"
       provides :windows_package, os: "windows"
+      provides :package, os: "windows"
```
randomcamel: Should this be deleted? Up above you have `provides :openbsd_package` only.

lib/chef/resource/windows_script.rb
```diff
@@ -31,8 +31,8 @@ class WindowsScript < Chef::Resource::Script
       def initialize(name, run_context, resource_name, interpreter_command)
         super(name, run_context)
         @interpreter = interpreter_command
-        @resource_name = resource_name
-        @default_guard_interpreter = resource_name
+        @resource_name = resource_name if resource_name
```
randomcamel: Does this allow the silent failure you were trying to avoid above?

lib/chef/resource/windows_service.rb
```diff
@@ -25,8 +25,8 @@ class WindowsService < Chef::Resource::Service
       # Until #1773 is resolved, you need to manually specify the windows_service resource
       # to use action :configure_startup and attribute startup_type
 
-      provides :service, os: "windows"
       provides :windows_service, os: "windows"
+      provides :service, os: "windows"
```
randomcamel: Same question as above for `:windows_package` and `:openbsd_package`.

lib/chef/resource/windows_package.rb
jkeiser: You mean should we remove `os: "windows"` from `provides :windows_package`?

lib/chef/resource/windows_script.rb
jkeiser: This is a different case because it isn't an optional arg.  This one is about the subclass passing resource_name--and I have changed the subclasses to pass `nil` since resource_name is already being set automatically.  However, I didn't want to remove the argument entirely, because I don't know if someone else somewhere may be overriding this and passing 4 args.

lib/chef/resource/windows_package.rb
jkeiser: (I think the answer to that particular question is "we could, but it's unclear which is better and I'm trying not to fix things I don't have a clear answer for right now :)")

lib/chef/resource.rb
jkeiser: Sure, I'll add it to the constructor.  It's worth thinking about how we can do validation in general without overwhelming the code.

spec/unit/lwrp_spec.rb
```diff
@@ -229,14 +229,6 @@ def get_lwrp_provider(name)
         expect(klass.resource_name).to eq(:foo)
       end
 
-      context "when creating a new instance" do
-        it "raises an exception if resource_name is nil" do
-          expect {
-            klass.new('blah')
-          }.to raise_error(Chef::Exceptions::InvalidResourceSpecification)
-        end
-      end
```
randomcamel: Maybe I'm not keeping everything in my head, but why remove this?

spec/unit/resource_spec.rb
```diff
@@ -336,6 +336,109 @@ class ResourceTestHarness < Chef::Resource
...
+          provides :self_resource_name_test_3
+        end
+
+        r = c.new('hi')
+        r.declared_type = :d
+        expect(c.resource_name).to eq :self_resource_name_test_2
+        expect(r.resource_name).to eq :self_resource_name_test_2
+        expect(r.declared_type).to eq :d
+      end
+      it "and there are no provides lines, resource_name is nil" do
```
randomcamel: If it's not valid to have a `nil` resource name, is this testing that we successfully get ourselves to an invalid state without an error? It might be worth a comment, at least.

lib/chef/resource/windows_package.rb
randomcamel: `os: "freebsd"` was removed above this, so if they're both correct, I'm curious why. =)

lib/chef/resource/windows_script.rb
randomcamel: Got it.

lib/chef/resource/windows_package.rb
jkeiser: @randomcamel I'm not seeing where that happened--the one instance I can find looks like it wasn't removed.  Maybe a different bsd?

spec/unit/lwrp_spec.rb
jkeiser: Because LWRPs are now using the same code as Resources, and resources don't worry about this.  I'm going to add this back in at the Resource level, though.

lib/chef/resource/windows_package.rb
randomcamel: I think that was an artifact of the changes that were mixed in earlier, so: moot.

spec/unit/resource_spec.rb
jkeiser: Fixed in the latest commit with an explicit error.  Thanks a ton for the UI think :)

jkeiser: Though I'll point out that this test is still here--it tests that we can get far enough to toss the real error.

lib/chef/resource/windows_package.rb
lamont-granquist: These lines are correct, imo.  `:windows_package` is only usable on windows, and its wired up as `:package` only on windows as well.   At some point it might be nice to unload the windows_package class on linux/solaris/aix/bsd since its just clutter there.

Note that people can do things like install apt-get on redhat, so pruning apt_package to platform_family of :debian is wrong and too strict (and some crazy person might have ported it to solaris even, so i was not overly strict when i set these things up).  The window stuff was deliberately restricted to windows though.

randomcamel:
:shipit:

lib/chef/resource.rb
lamont-granquist: I generally use `@resource_name=` rather than `self.resource_name =`

lib/chef/resource.rb
```diff
@@ -871,23 +864,53 @@ def after_created
...
+      #
+      # @example
+      #   class MyResource < Chef::Resource
+      #     provider_base Chef::Provider::Deploy
+      #     # ...other stuff
+      #   end
+      #
+      def provider_base(arg=nil)
+        @provider_base ||= arg || Chef::Provider
+      end
```
lamont-granquist: can you please throw a deprecation warning here and then remove the call from the deploy resource?  This is more gross class heirarchy crap that nobody uses now.

lamont-granquist:
The biggest issue I can see here is that we're making the first provides line magical and having it set the resource_name without tagging it in any way to make it obvious that its special.

This is leading to all the changes in the resources swapping around the two lines to make sure the correct one comes first.

lamont-granquist:
There was also some offline discussion about replacing `resource_name` with `declared_type` (declared type is the thing which the user types on the terminal).   I didn't particularly like that because then when a yum package resource is invoked with yum_package it'll be a `:yum_package` resource_name but when its invoked with package it'll be a `:package` resource_name, and I liked having a label which uniquely referred to the thing that wasn't the class name.  But that is also open for discussion if people disagree.

lamont-granquist:
(I also like the behavior in Chef 12 that `package "foo"` displays `yum_package[foo]` on redhat so that you see what is really getting done and not just what you typed).

jkeiser:
@lamont-granquist it seems weird to have to type the resource name twice.  Perhaps we should make the (new) `resource_name` automatically do `provides` instead.

lib/chef/resource.rb
jkeiser: I can deprecate it, but how should it work if someone writes this:

```ruby
deploy 'blah' do
  provider :revision
end
```

jkeiser: Deprecated, and added back to the deploy resource without deprecation.

lib/chef/resource/macosx_service.rb
```diff
@@ -22,8 +22,9 @@ class Chef
   class Resource
     class MacosxService < Chef::Resource::Service
 
-      provides :service, os: "darwin"
+      use_automatic_resource_name
       provides :macosx_service, os: "darwin"
```
lamont-granquist: can drop this too

jkeiser:
@lamont-granquist I think the patch is finished.  I did not do what we discussed at the end--automatically declare the resource_name as a class name--because it requires unprovides, and the provider resolver patch will interact with that a lot.  I'd like, if possible, to check this in, then check in the provider resolver patch, and *then* make it so automatic_resource_name is ... automatic.

lamont-granquist:
:+1:

:shipit:

PR #3444, opened 2015-05-28
Add in retry for possible transient SSL errors

glasschef:


nwatnickchef:
+1 would like to see this merged soon

gdbtek:
+1 please!

adamedx:
@lamont-granquist, @danielsdeleo, I thought our existing retries handled all of the transient cases that were safe to retry on. Do you agree this is an area we missed?

gdbtek:
I saw more random error like this an I'm wondering this PR will fix issue:

```
#<OpenSSL::SSL::SSLError: SSL_connect SYSCALL returned=5 errno=0 state=SSLv3 read finished A> had an error:
SSL_connect SYSCALL returned=5 errno=0 state=SSLv3 read finished A
```

adamedx:
I'd definitely like an inspection from @danielsdeleo or @lamont-granquist. I am wary of anything involving potential downgrade attacks with ssl or other funny business to get additional retries that may provide enough information to crack a key. I think that's probably not an issue, but I'd like additional eyes here.

adamedx:
Here's where the tests would go, though it looks like we are missing specs for the existing retries as well: https://github.com/chef/chef/blob/master/spec/unit/http_spec.rb.

lib/chef/http.rb
```diff
@@ -329,6 +329,13 @@ def retrying_http_errors(url)
           retry
         end
         raise Timeout::Error, "Timeout connecting to #{url}, giving up"
+      rescue OpenSSL::SSL::SSLError
+        if http_retry_count - http_attempts + 1 > 0
+          Chef::Log.error("SSL Error connecting to #{url}, retry #{http_attempts}/#{http_retry_count}")
+          sleep(http_retry_delay)
+          retry
+        end
+        raise Timeout::Error, "SSL Error connecting to #{url}, giving up"
```
danielsdeleo: Changing the error class here will mess with some stuff in knife that gives you hints about what to do next when you get an SSL error. The most common cause of this error is that you set up a Chef Server using the default self-generated, self-signed certificate, which cannot be verified with any cert in the default CA bundle. Knife rescues `SSLError` and tells you that you can possibly fix it by running `knife ssl check` and `knife ssl fetch`. I'd also like to add that kind of check and error messaging to chef-client. I think it would also be a good idea to skip the retries for certificate verification failures, since we can be confident that a retry will not fix the issue, so there is no point in forcing the user to wait for the retries to all fail.

It would probably be fine to change the error class to something under `Chef::Exceptions` though, which would maybe give us the ability to store enough information in the exception to fetch the server's certificate and then run a more useful SSL verification function against it (see https://gist.github.com/danielsdeleo/1ddecfaf638a8b23b8bb for a basic idea of the implementation). If we do that, then we should also update knife to accommodate the new error type. That code is here: https://github.com/chef/chef/blob/master/lib/chef/knife.rb#L439-L445

danielsdeleo:
RE: @adamedx's comment about retries being a possible security issue, I would imagine it's possible now for a MITM to, say, send a RST packet and trigger the retry mechanism we already have (plus a daemonized chef-client would just retry after a few minutes anyway) so I think this does not change the amount of risk there.

I'm definitely in favor of making our retries exhaustive as possible, but we can't change the exception class unless we make modifications elsewhere. The current code only raises `Timeout::Error` for backwards compatibility purposes, but that's not a concern here, and it's best to avoid raising other code's exception classes in most cases. Also, the error needs to be distinguishable as an SSL issue so code higher up the stack can give a more helpful error message. There's a couple of ways to do that with different pros/cons.

lamont-granquist:
Yeah, this is a useful thing to do, but I agree with dan about not mutating the exception class.

lamont-granquist:
I'm looking at all the `Errno::` exceptions that socket.c in ruby can throw and I'm kind of inclined towards us simply retrying all of them.  It may be highly unlikely to work for some of them, but the retries also won't hurt anyone.    We already retry on the common ones like ECONNREFUSED, and if we get back ENOBUFS from one request and the next one succeeds, then thats one less failed CCR and one less possible ticket cut to us.  Don't really see the downside...   Exceptions like EPROTOTYPE and EAFNOSUPPORT should simply never happen so if we retry them at worst its just a code branch we never hit.

So, I'd go with `rescue SystemCallError` and retry.

danielsdeleo:
I'd prefer to stay on the conservative side. We've had pretty much the exact same retry logic for years now and have only seen SSL stuff fall through (now that we're encouraging everyone to go SSL) and connection resets. We sometimes stream output from the network to files and I wouldn't want to accidentally change the behavior in the case it's actually writing to the FS that causes the error, or trying to get too large an object and hitting `Errno::ENOMEM`.

lamont-granquist:
Yeah, I guess I don't see a problem with retrying Errno::ENOMEM, if its our fault we should get some OOM killer love kicking in.  If its someone else, and they get killed, though, then we could retry and it might just work.  And I'm not too worried about retrying killing the box if you're in an ENOMEM kind if situation because the first time you get this back you've probably killed the perf on the box anyway and all the apps or the database or whatnot will be super unhappy and probably swapped out...

Anyway, #4133 i think is fine to move things along and fix SSL negotiation and RSTs...

danielsdeleo:
Replacing with https://github.com/chef/chef/pull/4133 which includes your commit @glasschef Thanks for contributing!

PR #3448, opened 2015-05-28
Fix dsc_resource to work with wmf5 april preview

jdmundrawala:
It looks like the April preview returns data in a different structure than the February preview


jdmundrawala:
Manually tested and the functional tests pass
cc @chef/client-windows

btm:
@jdmundrawala do you think a test would have failed if we ran our tests on wmf5 april without this patch?

jdmundrawala:
yes

btm:
:+1:

ksubrama:
:+1:

PR #3451, opened 2015-05-29
Run `gem install` tasks w/o bundler's env

danielsdeleo:
When bundler 1.10 is installed, running these tasks w/ bundler causes
`gem install` to fail with no output

@chef/client-core

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -108,9 +108,19 @@ Dir[File.expand_path("../*gemspec", __FILE__)].reverse.each do |gemspec_path|
   Gem::PackageTask.new(gemspec).define
 end
 
+def with_clean_env(&block)
+  if defined?(Bundler)
+    Bundler.with_clean_env(&block)
+  else
+    block.call
+  end
+end
+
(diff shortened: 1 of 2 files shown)


lamont-granquist:
:facepalm:

:shipit:

Rakefile
```diff
@@ -108,9 +108,19 @@ Dir[File.expand_path("../*gemspec", __FILE__)].reverse.each do |gemspec_path|
   Gem::PackageTask.new(gemspec).define
 end
 
+def with_clean_env(&block)
+  if defined?(Bundler)
+    Bundler.with_clean_env(&block)
```
tyler-ball: Just trying to think of weird issues - is there anytime Bundler wouldn't have `with_clean_env` defined?  It was added back in 2010 so I'm guessing not...

tyler-ball:
One question, otherwise :+1:

Rakefile
danielsdeleo: It's been around as long as I can remember.

PR #3455, opened 2015-05-31
Issue #3455: powershell_script: do not allow suppression of syntax errors

adamedx:
@ksubrama, @btm, @jdmundrawala  thanks for merging #3080. Here's an addendum based on a few pieces of feedback I didn't follow up on:

1. Took @jdmundrawala 's feedback that preserving behavior of other script resources that let you suppress syntax errors just added complexity at the expense of obscuring a much more likely scenario: user expects a (failure) return code of 1, but the script itself has a syntax error. In that case, the script needs to be fixed by the user, but the user isn't aware because Chef reports no error. The `powershell_script` resource hasn't been around long enough to need such strict backwards compatibility to justify the complexity in code (evidenced by numerous questions in the #3080 conversation) and the difficulty of the aforementioned scenario.
2. Feedback from @ksubrama and @jdmundrawala  around it not being obvious how we were actually doing the syntax check by embedding user code in a PowerShell script block. That's actually the core idea of the syntax validation, so to make that more obvious, I've restructured the code generation string so that instead of a terse one-liner, it's in multiple lines and looks closer to how you'd write idiomatic PowerShell code where the block was more obvious.

### DIFF
--- a/lib/chef/provider/powershell_script.rb
+++ b/lib/chef/provider/powershell_script.rb
@@ -30,8 +30,8 @@ class Chef
       end
 
       def action_run
-        valid_syntax = validate_script_syntax!
-        super if valid_syntax
+        validate_script_syntax!
+        super
       end
 
       def flags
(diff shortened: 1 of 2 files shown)


jdmundrawala:
:+1:

ksubrama:
:+1:

PR #3459, opened 2015-06-01
Support bootstrap attribute from json file ( replace of #3458 )

sawanoboly:
This PR is replace of #3458 .

When passed `@string` via --json-attributes, try read bootstrap attributes from file .

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -146,8 +146,8 @@ class Chef
       option :first_boot_attributes,
         :short => "-j JSON_ATTRIBS",
         :long => "--json-attributes",
-        :description => "A JSON string to be added to the first run of chef-client",
-        :proc => lambda { |o| Chef::JSONCompat.parse(o) },
+        :description => "A JSON string or @file to be added to the first run of chef-client",
+        :proc => lambda { |o| Chef::JSONCompat.parse(o.start_with?('@') ? File.read(o.delete('@')) : o) },
         :default => {}
 
       option :host_key_verify,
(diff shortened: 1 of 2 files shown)


patcon:
:+1:

lamont-granquist:
Would prefer this as a separate option like --json-file (with -J if its available, or whatever).  We don't use the `@file` notation magic anywhere else.

patcon:
Could we consider starting to use it, and if so, is this not as good a place as any to start? It's a widely understood pattern, and chef already has a perhaps overwhelming number of flags.

lamont-granquist:
I'm not particularly attracted to it, and @thommay wasn't very hot on it either.  If you feel strongly about it, it'd be best to open an rfc against https://github.com/chef/chef-rfc/issues/new and show up to the community meeting and get it through the rfc process.  Then we'd have a policy document that said we should do this to point to where it was written down after we'd all gotten to argue about it.

patcon:
:+1: Skipped around the code for other places this could be used, and this seemed to be the only relevant place. In light of that, let's just make the change @sawanoboly?

patcon:
also, probably `--json-attributes-file`, right @sawanoboly? Since the arg name would be a bit ambigious if reading a script that used just `--json-file`

sawanoboly:
I see. Is it OK for you that I rewrite this feature as new option `--json-attributes-file` ? @lamont-granquist

sawanoboly:
Hmm...

I would like to merge json-attributes into json-attribute-file if set both to avoid conflict.
But I have no idea when json-attribute string is passed as Array. I think that it can't merge simply. What should I do?

lamont-granquist:
If both are set you should either pick one or the other (and tell the user which one you picked), or just raise an exception.

sawanoboly:
OK. Thanks.

sawanoboly:
I've opened as new PR. #3900 
Thanks,

PR #3467, opened 2015-06-02
Add missing require

jdmundrawala:
This should fix the issue where doing
```ruby
require 'chef/resource'
```
throws an exception because Chef::Mixin::ShellOut is not loaded


thommay:
:+1:

PR #3469, opened 2015-06-02
Make ssh_port default to 22 for knife bootstrap.

ksubrama:
This is the behavior of knife-windows.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -59,6 +59,7 @@ class Chef
         :short => "-p PORT",
         :long => "--ssh-port PORT",
         :description => "The ssh port",
+        :default => 22,
         :proc => Proc.new { |key| Chef::Config[:knife][:ssh_port] = key }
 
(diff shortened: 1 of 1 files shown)


ksubrama:
/cc @chef/client-windows .
Could one of you also cc chef-client-core?

btm:
@ksubrama seems simple on the surface, but why are we adding it? Presumably all the underlying libraries make the right choice. I think we fixed our code where mixlib-cli defaults no longer override .ssh/config values, but I still worry about those kinds of repercussions with this kind of change.

@chef/client-core review?

tyler-ball:
I agree with @btm - I don't think that adding this will change anything because I think the underlying libraries already default to 22.  Do we have any evidence to the contrary?

lamont-granquist:
another agreement with btm.

ksubrama:
This is good to know.  knife-windows explicitly sets it to 22.  Perhaps the better solution is to remove the default value there and check that the underlying library defaults correctly.

PR #3470, opened 2015-06-02
add lazy require for 'chef/config'

lamont-granquist:
this is probably pulled in by chef/http at the top, but make it
explict (lazy to break circular deps).


lamont-granquist:
simple, needs review, @chef/client-core and @danielsdeleo in particular.

danielsdeleo:
:+1:  I couldn't remember if ruby required default args to be valid at file eval time, but it doesn't so this is g2g.

PR #3472, opened 2015-06-02
Allow directory resource to accept arrays for paths

eherot:
This would make it a bit cleaner to create directories in bulk. My Ruby coding skills are not the best so this may not be the best way to do this but I'd appreciate some feedback on the validity of the idea and perhaps how to code it better. If people are happy with it I'll also submit a PR to [chef/chef-docs](https://github.com/chef/chef-docs) explaining how to use it.

### DIFF
--- a/lib/chef/provider/directory.rb
+++ b/lib/chef/provider/directory.rb
@@ -36,7 +36,9 @@ class Chef
       def load_current_resource
         @current_resource = Chef::Resource::Directory.new(@new_resource.name)
         @current_resource.path(@new_resource.path)
-        if ::File.exists?(@current_resource.path) && @action != :create_if_missing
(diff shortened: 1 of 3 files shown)


lamont-granquist:
Since this is perfectly composable, I think the better solution here is to either:

- just wrap the directory provider with an `#each` loop as is common in cookbooks.
- write and LWRP or definition that takes an array of directories and constructs the directory resources

Both solutions keep the additional complexity out of the directory provider itself.

This also isn't an identical problem to the package provider taking an array since we had two issues there that made the problem not composable:

- individual yum/apt commands were very slow, and batching them was substantially faster
- some sets of packages would not install with individual commands but needed to be installed on the same command line

There's no analogous situation here that I can see.

Issue #3463, opened 2015-06-02
12.4.0.rc.0 LWRP Regression


cwebberOps:
So... I haven't been able to track down exactly what is broken, but the LWRP used by the chef_handler cookbook broke between 12.3.0 and 12.4.0.rc.0. I have created a way to reproduce the bug: https://github.com/cwebberOps/issue_repro. `kitchen converge current` will install 12.3.0 and works fine. `kitchen converge rc` will install 12.4.0.rc.0 and blows up. You will need the kitchen-appbundle-updater gem to run the test.

The error I see is:
```
         * chef_handler[BuildCookbook::HipChatHandler] action enable/tmp/kitchen/cache/cookbooks/chef_handler/providers/default.rb:71: warning: toplevel constant ChefHandler referenced by Chef::Resource::ChefHandler


           ================================================================================
           Error executing action `enable` on resource 'chef_handler[BuildCookbook::HipChatHandler]'
           ================================================================================

           NoMethodError
       -------------
           undefined method `new' for ChefHandler:Module

           Cookbook Trace:
           ---------------

           /tmp/kitchen/cache/cookbooks/issue_repro/recipes/default.rb:29:in `from_file'

           Resource Declaration:
       ---------------------
           suppressed sensitive resource output

           Compiled Resource:
           ------------------
       suppressed sensitive resource output


         ================================================================================
         Recipe Compile Error in /tmp/kitchen/cache/cookbooks/issue_repro/recipes/default.rb
         ================================================================================

         NoMethodError
         -------------
         chef_handler[BuildCookbook::HipChatHandler] (issue_repro::default line 23) had an error: NoMethodError: undefined method `new' for ChefHandler:Module

         Cookbook Trace:
         ---------------
           /tmp/kitchen/cache/cookbooks/chef_handler/providers/default.rb:71:in `load_current_resource'
           /tmp/kitchen/cache/cookbooks/issue_repro/recipes/default.rb:29:in `from_file'

         Relevant File Content:
         ----------------------
         /tmp/kitchen/cache/cookbooks/chef_handler/providers/default.rb:

          64:      converge_by("disable #{new_resource} as a #{type} handler") do
          65:        unregister_handler(type, new_resource.class_name)
          66:      end
          67:    end
          68:  end
          69:
          70:  def load_current_resource
          71>>   @current_resource = Chef::Resource::ChefHandler.new(new_resource.name)
          72:    @current_resource.class_name(new_resource.class_name)
[...]

PR #3473, opened 2015-06-02
Fix issue #3463: use real constants instead of const_missing

jkeiser:
Issue #3463 happens because the ChefHandler cookbook defines a `::ChefHandler` global module, and when you search for `Chef::Resource::ChefHandler`, it finds the top level one and so doesn't call `const_missing`.  Instead of using `const_missing`, we create a subclass of the LWRP to assign to `Chef::Resource::ChefHandler` that warns when you use it, and set is_a/kind_of/=== to work as if the two classes were equivalent.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1280,6 +1280,19 @@ class Chef
         Chef::Log.warn "#{class_name} already exists!  Deprecation class overwrites #{resource_class}"
         Chef::Resource.send(:remove_const, class_name)
       end
+
+      # In order to generate deprecation warnings when you use Chef::Resource::MyLwrp,
+      # we make a special subclass (identical in nearly all respects) of the
+      # actual LWRP.  When you say any of these, a deprecation warning will be
+      # generated:
+      #
+      # - Chef::Resource::MyLwrp.new(...)
+      # - resource.is_a?(Chef::Resource::MyLwrp)
+      # - resource.kind_of?(Chef::Resource::MyLwrp)
+      # - case resource
+      #   when Chef::Resource::MyLwrp
+      #   end
+      #
       resource_subclass = class_eval <<-EOM, __FILE__, __LINE__+1
         class Chef::Resource::#{class_name} < resource_class
           def initialize(*args, &block)
@@ -1289,7 +1302,7 @@ class Chef
           self
         end
       EOM
-      # Make case, is_a and kind_of work with the new subclass, for backcompat
+      # Make case, is_a and kind_of work with the new subclass, for backcompat.
       # Any subclass of Chef::Resource::ResourceClass is already a subclass of resource_class
       # Any subclass of resource_class is considered a subclass of Chef::Resource::ResourceClass
       resource_class.class_eval do
--- a/lib/chef/resource/lwrp_base.rb
+++ b/lib/chef/resource/lwrp_base.rb
@@ -58,10 +58,7 @@ class Chef
           resource_class.class_from_file(filename)
 
           # Make a useful string for the class (rather than <Class:312894723894>)
-          resource_class.class_eval do
-            define_method(:kind_of?)      { |other| other.class <= resource_class }
-            define_method(:is_a?)         { |other| other.class <= resource_class }
-
+          resource_class.instance_eval do
             define_singleton_method(:to_s) do
               "LWRP resource #{resource_name} from cookbook #{cookbook_name}"
             end
@@ -73,7 +70,7 @@ class Chef
           LWRPBase.loaded_lwrps[filename] = true
 
           # Create the deprecated Chef::Resource::LwrpFoo class
-          resource_subclass = Chef::Resource.register_deprecated_lwrp_class(resource_class, convert_to_class_name(resource_name))
+          Chef::Resource.register_deprecated_lwrp_class(resource_class, convert_to_class_name(resource_name))
           resource_class
         end
 


tyler-ball:
@jkeiser wanna ping me and go through this with me?  I have no idea what I'm looking at

lamont-granquist:
:+1:

lib/chef/resource/lwrp_base.rb
```diff
@@ -58,7 +58,10 @@ def build_from_file(cookbook_name, filename, run_context)
           resource_class.class_from_file(filename)
 
           # Make a useful string for the class (rather than <Class:312894723894>)
-          resource_class.instance_eval do
+          resource_class.class_eval do
+            define_method(:kind_of?)      { |other| other.class <= resource_class }
```
tyler-ball: Lets add a comment that explains why we need this.  Something along the lines of...

If you have a subclass of the LWRP class (like we make to support the deprecated class lookup under `Chef::Resource`) and you call `subclass.kind_of? LwrpClass` that will return false... Unless! we override those methods to make sure compare `kind_of?` with the real resource class.

@jkeiser Is my understanding of this correct?

tyler-ball: @jkeiser are these necessary if we overwrite them in the subclass (above)?

lib/chef/resource.rb
```diff
@@ -1274,31 +1274,48 @@ def self.resource_matching_short_name(short_name)
...
-          Chef::Log.warn "#{class_name} already exists!  Cannot create deprecation class for #{resource_class}"
-        else
-          deprecated_constants[class_name.to_sym] = resource_class
+    # @api private
+    def self.register_deprecated_lwrp_class(resource_class, class_name)
+      if Chef::Resource.const_defined?(class_name, false)
+        Chef::Log.warn "#{class_name} already exists!  Deprecation class overwrites #{resource_class}"
+        Chef::Resource.send(:remove_const, class_name)
+      end
+      resource_subclass = class_eval <<-EOM, __FILE__, __LINE__+1
```
tyler-ball: Is the `,` after `EOM` a typo?

lib/chef/resource.rb
```diff
@@ -1274,31 +1274,48 @@ def self.resource_matching_short_name(short_name)
...
+        define_method(:kind_of?) do |other|
+          other.is_a?(Module) && other === self
         end
       end
-
-      def const_missing(class_name)
-        if deprecated_constants[class_name.to_sym]
+      resource_subclass.class_eval do
+        define_singleton_method(:===) do |other|
           Chef::Log.deprecation("Using an LWRP by its name (#{class_name}) directly is no longer supported in Chef 13 and will be removed.  Use Chef::Resource.resource_for_node(node, name) instead.")
```
tyler-ball: This deprecation method is now in the subclass's initialize method above - did you mean to have it here too?  What is the benefit of having it in both places?

lib/chef/resource.rb
```diff
@@ -1274,31 +1274,48 @@ def self.resource_matching_short_name(short_name)
...
+            Chef::Log.deprecation("Using an LWRP by its name (#{class_name}) directly is no longer supported in Chef 13 and will be removed.  Use Chef::Resource.resource_for_node(node, name) instead.")
+            super
+          end
+          self
+        end
+      EOM
+      # Make case, is_a and kind_of work with the new subclass, for backcompat
+      # Any subclass of Chef::Resource::ResourceClass is already a subclass of resource_class
+      # Any subclass of resource_class is considered a subclass of Chef::Resource::ResourceClass
+      resource_class.class_eval do
```
tyler-ball: I think this comment could be expanded some.  What we're trying to say is that we want to support `resource_class.kind_of?(Chef::Resource::ResourceClass) == true` right?

lib/chef/resource.rb
```diff
@@ -1274,31 +1274,48 @@ def self.resource_matching_short_name(short_name)
...
-      # @api private
-      def register_deprecated_lwrp_class(resource_class, class_name)
-        if Chef::Resource.const_defined?(class_name, false)
-          Chef::Log.warn "#{class_name} already exists!  Cannot create deprecation class for #{resource_class}"
-        else
-          deprecated_constants[class_name.to_sym] = resource_class
+    # @api private
+    def self.register_deprecated_lwrp_class(resource_class, class_name)
+      if Chef::Resource.const_defined?(class_name, false)
+        Chef::Log.warn "#{class_name} already exists!  Deprecation class overwrites #{resource_class}"
```
tyler-ball: Can you add a comment explaining why its important to overwrite the existing class with the deprecation class?

lib/chef/resource.rb
tyler-ball: I also think it would be helpful to have some comments above this explaining why we want to subclass the anonymous `resource_class`.  IE, this is all necessary so that `Chef::Resource::LwrpClass` still works (until Chef 13)

lib/chef/resource/lwrp_base.rb
jkeiser: Will get rid of this :)

lib/chef/resource.rb
jkeiser: Nope, not a typo.  <<-EOM, x, y means "the first argument is the stuff until EOM.  Also there are these two other arguments."

Will do on comments!

lib/chef/resource.rb
jkeiser: It needs to be in both places because you can compare if (resource.is_a?(Chef::Resource::MyLwrp)) without ever having instantiated one.  That code will break too, when we remove the class.

lib/chef/resource.rb
jkeiser: Because users who currently write

```ruby
x = my_lwrp 'blah' do
end
if x.kind_of?(Chef::Resource::MyLwrp)
```
will see their code break if we don't :)  It's just enough on the edge of possibility that I felt we should account for it.

lib/chef/resource.rb
jkeiser: Honestly it's not, it just made tests easier to write.

PR #3479, opened 2015-06-03
try m3.medium over m1.small

lamont-granquist:
m3 has newer CPUs on SSDs.  should be substantially faster.

there is no m3.small


lamont-granquist:
@chef/client-core looks like this takes tests down to 17 minutes from 20 mins.  not as much as i was hoping but its a bit.

btm:
:+1: 

Makes me curious about where the performance bottleneck is.

lamont-granquist:
bundle install is quite a bit of it, trying to add caching so we don't rebuild gecode every single time.

PR #3480, opened 2015-06-03
We don't need to run ruby 2.0 in travis.

jdmundrawala:
It only matters on Windows and so it gets run in
appveyor


danielsdeleo:
:+1:

lamont-granquist:
:+1:

PR #3481, opened 2015-06-03
Lcg/bundle cache

lamont-granquist:
cuts the end-to-end tests down to 9 mins from 17 mins due to being able to reuse libgecode from a previous bundle install.

this should operate the same way as caching bundles across app deployments with shared directories in the deploy provider.  if we bump libgecode then without a Gemfile.lock we'll still need a new version so we'll build it once and add the new version to the cache -- we shouldn't wind up accidentally version pinned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 
 sudo: false
 # Early warning system to catch if Rubygems breaks something


lamont-granquist:
@chef/client-core once this finally goes green, i want to get this merged, its a huge speed boost.

danielsdeleo:
FWIW, chef-dk uses the pure-ruby Molinillo backend in Solve 2, so we could use policyfiles instead if we want. That said, :+1:

lamont-granquist:
would love to murder gecode at some point, but will still would want to use this anyway...

btm:
yay! :+1:

PR #3483, opened 2015-06-03
allow include_recipe from LWRP provider code

lamont-granquist:
allows include_recipe "build-essential" to work from provider
code.


lamont-granquist:
@danielsdeleo @someara this is what we talked about.  seems to work fine.

@chef/client-core for review if anyone is interested.

someara:
Looks good. I've been doing this in providers for a while now.

PR #3484, opened 2015-06-03
These tests were accidently left out when I moved #2621 to master

tyler-ball:
I found these when answering someone's questions.  Mo tests, mo better!

\cc @chef/client-core


btm:
:+1: pending travis

smurawski:
:+1: Yay tests!

PR #3488, opened 2015-06-04
switch to an HVM ssd based AMI

lamont-granquist:


lamont-granquist:
@chef/client-core this is a two-fer.  first one looks like it might shave another minute off of the TK runs, the second one is the one that fixes the TK runs by pinning kitchen-ec2 to the last working version.

danielsdeleo:
Conflicted about pinning kitchen-c2 to the older one. IIRC we were getting race condition failures before the last update, so it will be kinda unstable. Also, running off master is finding bugs which is good. Contrarily, this gets us working (ish) again and we can just go back to using master once it's fixed.

Anyway, :+1:  provided we get back on master later.

lamont-granquist:
Yeah, this is intended to be temporary.  I looked at the kitchen-ec2 bug and I can guess at the problem, but don't know what tyler's idea would be for how to fix it right, and he's AFK at the moment, and I'm just unfamiliar with the TK codebase...   So, pinning seems like the best way to get us back on track this morning...  The fact that we immediately found the regression in kitchen-ec2 is actually quite useful so we should go back to tracking master...

PR #3489, opened 2015-06-04
Hyphenated LWRP

danielsdeleo:
has a regression test for provider priority map setting with cookbooks with hyphens in the name. Currently doesn't have the fix. I did some refactoring in the first commit, so you might want to focus on the second commit.

@sdelano @jkeiser

### DIFF
--- a/lib/chef/mixin/convert_to_class_name.rb
+++ b/lib/chef/mixin/convert_to_class_name.rb
@@ -23,9 +23,7 @@ class Chef
       extend self
 
       def convert_to_class_name(str)
-        str = str.dup
-        str.gsub!(/[^A-Za-z0-9_]/,'_')
-        str.gsub!(/^(_+)?/,'')
+        str = normalize_snake_case_name(str)
         rname = nil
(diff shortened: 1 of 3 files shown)


jkeiser:
I just pushed the actual fix to this (thanks to pointers from Dan!), plus a full-scale test for resource+provider in a cookbook to confirm.

sdelano:
LGTM - as long as the CI runs pass this looks great. I'm running a build of chef-server through with these changes which will be exercised via the `private-chef` cookbook. That'll take a couple hours to build since the git cache is cleared with a project override: [build](http://wilson.ci.chef.co/job/chef-server-12-trigger-ad_hoc/558/downstreambuildview/)

stevendanna:
I'm going to merge this despite the travis failures.  The travis failures are EC2-related and this is blocking forward progress on getting some of the other pipelines working.  I will circle back around to the EC2 failures once I get the chef-server pipeline back up.

thommay:
retrospectively :+1:

PR #3493, opened 2015-06-04
Allow properties to be defined on resources

jkeiser:
Implements most of the [https://github.com/chef/chef-rfc/blob/jk/property/new/property.md](Property RFC), which is set to be accepted shortly.

- `property` can now be written in Resource:
  ```ruby
  class Foo < Chef::Resource
    resource_name :foo
    property :x, String
  end

  foo 'blah' do
    x 'hi'
  end
  ```
- It is now possible to set a property to `nil` *if* the type allows it:
  ```ruby
  class Foo < Chef::Resource
    resource_name :foo
    property :x, [ String, nil ], default: 'hi'
  end

  foo 'blah' do
    x nil
  end
  ```
- lazy defaults are now possible and run in the context of the instance:
  ```ruby
  class Foo < Chef::Resource
    resource_name :foo
    property :base, Integer, default: 1
    property :x, Integer, default: lazy { 10 + base }
  end

  foo 'blah' do
    puts x #=> 11
  end
  ```
- coercion is implemented:
  ```ruby
  class Foo < Chef::Resource
    resource_name :foo
    property :x, Integer, coerce: proc { |v| v.to_i + 1 }
  end

  foo 'blah' do
    x '100'
    puts x #=> 101
  end
  ```
- `resource.property_is_set?(:x)` can tell you whether a property has been set by the user.
  ```ruby
  class Foo < Chef::Resource
    resource_name :foo
    property :x, Integer
  end

  foo 'blah' do
    puts x.property_is_set?(:x) #=> false
  end
  ```

It does not yet:
- add properties to `state_attrs`
- allow sharing and specializing types

### DIFF
--- a/lib/chef/mixin/params_validate.rb
+++ b/lib/chef/mixin/params_validate.rb
@@ -252,6 +252,7 @@ class Chef
 
       # Compare the way "case" would (i.e. `===`)
       def _pv_is(opts, key, to_be, raise_error: true)
+        return true if !opts.has_key?(key.to_s) && !opts.has_key?(key.to_sym)
         value = _pv_opts_lookup(opts, key)
         to_be = [ to_be ].flatten(1)
         to_be.each do |tb|
--- a/lib/chef/mixin/params_validate.rb
+++ b/lib/chef/mixin/params_validate.rb
@@ -164,6 +164,17 @@ class Chef
         end
       end
 
+      #
+      # List of things values must be equal to.
+      #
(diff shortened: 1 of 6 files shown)


jkeiser:
This PR can be most effectively reviewed going commit-by-commit.  One feature added for each commit.

ksubrama:
This is quite cool and should make writing providers much much cleaner.

:+1:

PR #2434, opened 2014-11-17
Update `dpkg` provider to remove `DPKG_INFO` regular expression and let `dpkg-dep` handle package name.

kwilczynski:
This is as per the "Debian Policy Manual, Chapter 5 - Control files and their fields", see:

  https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version

Fixes #2417.

Signed-off-by: Krzysztof Wilczynski <krzysztof.wilczynski@linux.com>

PR #3498, opened 2015-06-06
Use dpkg-deb directly rather than regex

thommay:
This is based on @kwilczynski's work in #2434


thommay:
ping @chef/client-core

lamont-granquist:
yeah that looks fine to me.

you're updating the `@new_resource` which should be avoided but you're just preserving the behavior of the old code.  it could probably use a `#FIXME` above it if you don't have time to actually fix it now, those are kind of bad bugs which are often exposed by resource cloning in horrible ways.

thommay:
@lamont-granquist i updated to not hit `new_resource`

lamont-granquist:
if that works, its better, not sure what that code was doing or if you're gonna hit an edge case....

lamont-granquist:
:+1:

ranjib:
:+1:

PR #3499, opened 2015-06-06
Improve performance of method_missing

jkeiser:
by not repeatedly sorting and calling enabled_handlers
(This was causing major slowdown in tests)


thommay:
:+1:

Issue #3496, opened 2015-06-05
rpm_package silently fails on packages with a `~` character in the version


irvingpop:
Recent packages created by Chef CI have a tilde character in the version, like so:  `supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm`  - and installation of these is silently failing on Chef 12.3.0
(note: I'm not 100% sure it is the tilde, but this is my best guess)

Given the following cookbook code:

```ruby
case node['platform_family']
when 'rhel'
  package = 'https://web-dl.packagecloud.io/chef/stable/packages/el/5/supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm'
when 'debian'
  package = 'https://web-dl.packagecloud.io/chef/stable/packages/ubuntu/precise/supermarket_1.10.1~alpha.0-1_amd64.deb'
end

pkgname = ::File.basename(package)
cache_path = ::File.join(Chef::Config[:file_cache_path], pkgname)

remote_file cache_path do
  source package
  mode '0644'
end

if node['platform_family'] == 'rhel'
  rpm_package cache_path
  execute 'rpm -qi supermarket'
elsif node['platform_family'] == 'debian'
  dpkg_package cache_path
  execute 'dpkg -L supermarket'
end
```

It works correctly on Ubuntu, but on RHEL (tested on CentOS 5.11 and 6.6) the following happens:
```
           * rpm_package[supermarket] action install[2015-06-05T23:20:53+00:00] INFO: Processing rpm_package[supermarket] action install (/tmp/kitchen/cache/cookbooks/chef-server-ingredient/libraries/chef_server_ingredients_provider.rb line 62)
       [2015-06-05T23:20:53+00:00] DEBUG: providers for generic rpm_package resource enabled on node include: [Chef::Provider::Package::Rpm]
       [2015-06-05T23:20:53+00:00] DEBUG: providers that refused resource rpm_package[supermarket] were: []
       [2015-06-05T23:20:53+00:00] DEBUG: providers that support resource rpm_package[supermarket] include: [Chef::Provider::Package::Rpm]
       [2015-06-05T23:20:53+00:00] DEBUG: providers that survived replacement include: [Chef::Provider::Package::Rpm]
       [2015-06-05T23:20:53+00:00] DEBUG: rpm_package[supermarket] checking rpm status
       supermarket 1.10.1~alpha.0-1.el5
       [2015-06-05T23:20:53+00:00] DEBUG: rpm_package[supermarket] checking install state
[...]

PR #3076, opened 2015-03-13
Lets add coverage for the things we broke

jaym:
cc @btm @lamont-granquist @chef/client-core

PR #3503, opened 2015-06-08
fix rpm_package when sourced packages have a tilde character in the version

irvingpop:
Fixes the issue described in #3496 

The tilde character '~' is valid in package versions and has some useful benefits and was implemented in our Omnibus builds last December: https://github.com/chef/omnibus/pull/401

The regex we use to examine the package version in `rpm_package` silently fails on these packages because it doesn't look for that ~ character: http://rubular.com/r/u63jqN7OJI

This PR adds the ~ character to the regex.  

CC: @chef/client-maintainers @chef/release-engineers

### DIFF
--- a/spec/unit/provider/package/rpm_spec.rb
+++ b/spec/unit/provider/package/rpm_spec.rb
@@ -203,3 +203,73 @@ describe Chef::Provider::Package::Rpm do
     end
   end
 end
+
+# adding tests for #3503
+describe Chef::Provider::Package::Rpm do
+  let(:provider) { Chef::Provider::Package::Rpm.new(new_resource, run_context) }
+  let(:node) { Chef::Node.new }
+  let(:events) { Chef::EventDispatch::Dispatcher.new }
+  let(:run_context) { Chef::RunContext.new(node, {}, events) }
+  let(:new_resource) do
+    Chef::Resource::Package.new("supermarket").tap do |resource|
+      resource.source "/tmp/supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm"
+    end
+  end
+  let(:exitstatus) { 0 }
+  let(:stdout) { "supermarket 1.10.1~alpha.0-1.el5" }
+  let(:status) { double('Process::Status', exitstatus: exitstatus, stdout: stdout) }
+
+  before(:each) do
+    allow(::File).to receive(:exists?).and_return(true)
+    allow(provider).to receive(:shell_out!).and_return(status)
+  end
+
+  describe "when determining the current state of the package with a tilde (~) character in the version" do
+    context "installation exists" do
+      it "should get the source package version from rpm if provided" do
+        expect(provider).to receive(:shell_out!).with("rpm -qp --queryformat '%{NAME} %{VERSION}-%{RELEASE}\n' /tmp/supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm", timeout: 900).and_return(status)
+        expect(provider).to receive(:shell_out).with("rpm -q --queryformat '%{NAME} %{VERSION}-%{RELEASE}\n' supermarket", timeout: 900).and_return(status)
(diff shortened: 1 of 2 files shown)


danielsdeleo:
Regression test please?

:+1: otherwise. Hit me up in chat if you need help.

jdmundrawala:
:+1: 
I'm going to try to revive https://github.com/chef/chef/pull/3076

schisamo:
:+1: Please add some test coverage as @danielsdeleo mentioned.

irvingpop:
Thanks @danielsdeleo @jdmundrawala @schisamo -  I added rspec tests for this case and they're passing on travis.  I've also manually verified the install, upgrade and noop behavior on RHEL 5 and 7

tyler-ball:
:+1:

spec/unit/provider/package/rpm_spec.rb
```diff
@@ -203,3 +203,73 @@
...
+      end
+
+      it "should not attempt an rpm installation" do
+        expect(provider).to_not receive(:shell_out!).with("rpm  -i /tmp/supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm", timeout: 900)
+        expect { provider.run_action(:any) }.to raise_error(Chef::Exceptions::Package)
+      end
+
+      it "should not attempt an rpm upgrade" do
+        expect(provider).to_not receive(:shell_out!).with("rpm  -U /tmp/supermarket-1.10.1~alpha.0-1.el5.x86_64.rpm", timeout: 900)
+        expect { provider.run_action(:any) }.to raise_error(Chef::Exceptions::Package)
```
danielsdeleo: This is failing on windows, looks like. The test code that's running `provider.run_action(:any)` is really weird; I guess it's just trying to trigger the resource requirements code.

irvingpop: hm, I copy-pasted that from line 53 of the same file.  I wonder why it's not triggering on that as well!

irvingpop: One difference I just noticed is that here it's inside of a `context`.   I'll fire up a Windows box tomorrow and play around with it.

irvingpop: @danielsdeleo  Playing around with this, I couldn't get the tests to pass on Windows no matter what combination of things I tried.  

Couple of options I see:

1.  remove the tests that fail on Windows
2.  would it be acceptable to place a `skip` here, considering that rpm_package isn't applicable on Windows anyway?  how broadly should it be scoped?

danielsdeleo: That's okay as a last resort, but this seems to indicate there is some flakiness in the tests as written which is probably covered up by some other mitigating factor on unix. I'm taking a look.

danielsdeleo:
@irvingpop I found that there were calls to shellout that weren't stubbed and my hunch is that these were behaving differently on windows. I updated the tests so that they'll fail for any non-stubbed shell out call, so I expect that problem to be fixed.

danielsdeleo:
Also, I think I'm gonna fix up the existing tests with the same stuff I did here, so hold off on merging.

lamont-granquist:
:+1:

danielsdeleo:
Okay, added a ton of cleanup. In particular, made the tests call the methods that `Provider#run_action` calls so hopefully they're more likely to catch a bug where the methods work in isolation but don't pass state correctly.

irvingpop:
Thanks Dan, that looks way cleaner!

lamont-granquist:
Yeah, much better.  :+1:

If you want to go all the way turning the `:action_install`'s into `run_action(:install)`'s can be used to test even more code working together and gets rid of boilerplate in before sections.

danielsdeleo:
yep, agree, but that was enough for one day :sweat: :wink:

PR #3508, opened 2015-06-08
lazier coupling of Chef::HTTP to Chef::Config

lamont-granquist:
- remove Chef::Config symbol outside of `#config` lazy initializer
- allow for dep injecting a different hash completely for Chef::Config
- allow for updating `@config` or any of the config values from attr_writers
- move chef/config dep to the bottom

### DIFF
--- a/lib/chef/http/auth_credentials.rb
+++ b/lib/chef/http/auth_credentials.rb
@@ -28,8 +28,9 @@ class Chef
     class AuthCredentials
       attr_reader :client_name, :key
 
-      def initialize(client_name=nil, key=nil)
+      def initialize(client_name=nil, key=nil, opts={})
         @client_name, @key = client_name, key
+        @config = opts[:config] if opts.key?(:config)
       end
 
       def sign_requests?
@@ -44,12 +45,20 @@ class Chef
         request_params                 = request_params.dup
         request_params[:timestamp]     = Time.now.utc.iso8601
         request_params[:user_id]       = client_name
-        request_params[:proto_version] = Chef::Config[:authentication_protocol_version]
+        request_params[:proto_version] = config[:authentication_protocol_version]
         host = request_params.delete(:host) || "localhost"
 
         sign_obj = Mixlib::Authentication::SignedHeaderAuth.signing_object(request_params)
         signed =  sign_obj.sign(key).merge({:host => host})
         signed.inject({}){|memo, kv| memo["#{kv[0].to_s.upcase}"] = kv[1];memo}
+      end
+
+      def config
+        @config ||=
+          begin
+            Thread.exclusive { require 'chef/config' unless defined?(Chef::Config) }
+            Chef::Config
+          end
       end
 
     end
--- a/lib/chef/http/ssl_policies.rb
+++ b/lib/chef/http/ssl_policies.rb
@@ -31,15 +31,16 @@ class Chef
     # Configures SSL behavior on an HTTP object via visitor pattern.
     class DefaultSSLPolicy
 
-      def self.apply_to(http_client)
-        new(http_client).apply
+      def self.apply_to(http_client, opts = {})
+        new(http_client, opts).apply
         http_client
       end
 
       attr_reader :http_client
 
-      def initialize(http_client)
+      def initialize(http_client, opts = {})
         @http_client = http_client
+        @config = opts[:config] if opts.key?(:config)
       end
 
       def apply
@@ -102,7 +103,11 @@ class Chef
       end
 
       def config
-        Chef::Config
+        @config ||=
+          begin
+            Thread.exclusive { require 'chef/config' unless defined?(Chef::Config) }
+            Chef::Config
+          end
       end
 
       private
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -27,7 +27,6 @@ require 'uri'
 require 'chef/http/basic_client'
 require 'chef/http/socketless_chef_zero_client'
 require 'chef/monkey_patches/net_http'
-require 'chef/config'
 require 'chef/platform/query_helpers'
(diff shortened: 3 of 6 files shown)


lamont-granquist:
@chef/client-core what do you think about a bit less tight coupling here?

stevendanna:
Overall I like this idea.  When I get a chance there are two things I would like to test, I mention them here in case you know what the behavior will be off hand:

``` ruby
# Existing code does this, will it break?
Chef::Config[:chef_server_url] = "some_url"
@rest.get("some path")
Chef::Config[:chef_server_url] = "some_other_url"
@rest.get("some path")
```

```ruby
# given the changes I'd like to do things like:
f = Chef::HTTP.new(...)
f.custom_http_headers = "bar"
# what will the output of this be:
puts Chef::Config[:custom_http_headers]
```

tyler-ball:
What about calling `.dup` on the config if it is provided in the constructor?  This would mean that every rest instance would have their own snapshot of the Config and wouldn't all reference the same Config.  Is that too big of a breaking change?

The code looks good, but I think the bigger question is how this will affect existing behavior which relies on Chef::HTTP - that I am not familiar with.

lamont-granquist:
The problem with duping is that I think that breaks @stevendanna's first use case since then updates to the chef_server_url won't be reflected in the object.  However, I don't think we need to dup and the latter case will still work.  I think its equivalent to:

```ruby
[3] pry(main)> h = Hash.new
=> {}
[4] pry(main)> h[:foo] = :bar
=> :bar
[5] pry(main)> thing = h[:foo]
=> :bar
[6] pry(main)> thing = :baz
=> :baz
[7] pry(main)> h[:foo]
=> :bar
```

So when we assign `@chef_server_rest = Chef::Config[:chef_server_rest]` we're assigning the ivar to point at the chef_server_rest string.  If we were to do something destructive like `@chef_server_rest << "/foo/bar/baz" then we would be mutating the string in the Chef::Config itself, but nowhere do we attempt to mutate that and its clearly external config and its rude to update strings like that passed in externally without the code that actually mutates it doing the `#dup` itself.  We could Chef::Config[:chef_server_url].dup the string but I don't think that's necessary.

lamont-granquist:
And @tyler-ball I don't see how it would affect existing code.  None of the existing code passes in a config option argument so it'll just get Chef::Config as its `#config` hash, and nothing uses the new setters or getters so they'll just default to the Chef::Config values that they had.  It should just work.

One crazy idea I've got is to move the require into the config getter:

```ruby
def config
  @config ||= begin
    require 'chef/config'
    Chef::Config
  end
end
```

Then if you DI something in the config hash you won't even have the tight coupling to the chef/config class.  That could allow standalone use of Chef::REST without Chef::Config at all.

stevendanna:
> Chef::REST without Chef::Config at all

I would like that.

danielsdeleo:
I would like `Chef::REST` to go away, but I'm all for `Chef::HTTP` and friends to be less tightly coupled to `Chef::Config`. BTW, the ssl_policy stuff is also accessing `Chef::Config` directly, so you'll probably have to attack that (maybe take a look at the other middlewares, too) to get 100% `Chef::Config` independence.

danielsdeleo:
Anyway, :+1: to what's here, though I think the caveat that some middlewares and the SSL stuff aren't "fixed" is important to note.

lamont-granquist:
Yeah, I think its used in a couple of other places as well.  Since it looks like the idea went over well, I may try to get a bit more aggressive.

lamont-granquist:
@danielsdeleo @stevendanna  thoughts on this kind of direction?

there's still a lot of work here, i think i'm still missing some wiring (and tests), and should probably extract a mixin, but the idea would be able to inject opts[:config] all the way from the Chef::HTTP object through all the middleware, and have it lazily required to the point where it could be used without the presence of Chef::Config at all.

lib/chef/http/auth_credentials.rb
```diff
@@ -44,7 +45,7 @@ def signature_headers(request_params={})
         request_params                 = request_params.dup
         request_params[:timestamp]     = Time.now.utc.iso8601
         request_params[:user_id]       = client_name
-        request_params[:proto_version] = Chef::Config[:authentication_protocol_version]
+        request_params[:proto_version] = config[:authentication_protocol_version]
```
danielsdeleo: I wonder if "latest" is still an allowed value? We should probably have this fallback to something like that to force upstream users to set an appropriate version for their apps (when using outside of Chef).

danielsdeleo:
Yeah, I'm down with it. Just one nitpick which is sorta related to the idea that we need to start thinking of how we version `Chef::HTTP` as an independent project (especially if we're gonna take this to the logical conclusion of making it a component gem like chef-config).

jkeiser:
Loose coupling would be AWESOME.  I would love to see config stashed in run_context, and almost no one use Chef::Config directly.

danielsdeleo:
Come to think of it, I'd be cool to make passing config into the middlewares (and ssl policy and such) mandatory. The main API is really `Chef::HTTP` and subclasses. Plus then you don't have to copy the config-requiring code everywhere.

stevendanna:
:+1: Overall.  I like Dan's comment about making the config mandatory in the middleware.

lamont-granquist:
.

lib/chef/http/ssl_policies.rb
```diff
@@ -102,7 +103,11 @@ def set_client_credentials
       end
 
       def config
```
coderanger: This is repeated so many times it should probably be in a mixin somewhere.

lamont-granquist: yeah we should have standardized mixins for `def config`, `def env`, and the typical REST object API at a minimum.

PR #3512, opened 2015-06-10
Fix chef-config gem homepage

thommay:


thommay:
cc @chef/client-maintainers

danielsdeleo:
For other stuff we use `https://www.chef.io` but I could see a case made that these "component" gems should point here. :+1: either way, as all colors of bikeshed protect bikes from the rain.

lamont-granquist:
:no_bicycles: 
:+1:

Issue #3527, opened 2015-06-12
[knife] Allow ssh_user arg to be set inline via `knife bootstrap user@hostname`


patcon:
The thought was that this format might be more intuitive and aligned with expectations outside chef narnia :)

Thoughts?

patcon:
If this gets the ok, this is the only place where it would seem a test could be added (copy-pasting the scenario and tweaking), right?

https://github.com/chef/chef/blob/b08c9010c387c762a6fa82fa9ae8c2a5e5c8594d/spec/unit/knife/bootstrap_spec.rb#L398-L436

sawanoboly:
:+1:

PR #3529, opened 2015-06-12
Allow user@hostname format for knife-bootstrap.

patcon:
Pull request to deal with #3527.

This is untested, but if the general premise works, I'll do my due diligence :)

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -42,7 +42,7 @@ class Chef
         Chef::Knife::Ssh.load_deps
       end
 
-      banner "knife bootstrap FQDN (options)"
+      banner "knife bootstrap [SSH_USER@]FQDN (options)"
 
       option :ssh_user,
         :short => "-x USERNAME",
@@ -246,13 +246,25 @@ class Chef
         "chef-full"
       end
 
+      def host_descriptor
+        Array(@name_args).first
+      end
+
       # The server_name is the DNS or IP we are going to connect to, it is not necessarily
       # the node name, the fqdn, or the hostname of the server.  This is a public API hook
       # which knife plugins use or inherit and override.
       #
       # @return [String] The DNS or IP that bootstrap will connect to
       def server_name
-        Array(@name_args).first
+        if host_descriptor
+          @server_name ||= host_descriptor.split('@').reverse[0]
+        end
+      end
+
+      def user_name
+        if host_descriptor
+          @user_name ||= host_descriptor.split('@').reverse[1]
+        end
       end
 
       def bootstrap_template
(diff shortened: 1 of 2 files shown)


chef-supermarket:

* @patcon

chef-supermarket:

lib/chef/knife/bootstrap.rb
```diff
@@ -246,7 +247,19 @@ def default_bootstrap_template
...
+      def host_descriptor
+        unless @name_args.nil?
+          parts = Array(@name_args).first.split('@')
+          @host_descriptor = {
+            :host => parts.pop,
+            :user => parts.pop
+          }
+        end
+
+        @host_descriptor
```
lamont-granquist: This is probably better written something like this:

```ruby
def first_arg
  Array(@name_args).first
end

def server_name
   @server_name ||= if first_arg.include?('@')
      first_arg.split('@')[1]
    else
      first_arg
    end
end

def user_name
  @user_name ||=
    first_arg.split('@')[0] if first_arg.include?('@')
  # this will be nil in the else case
end
```

i don't quite see where you've handled the case when there is no '@' in this code either.

also not sure you need to check `@name_args.nil?` since that would cause the existing `server_name` method to explode I don't think that's a useful case to worry about and some other code must handle it.

thommay:
hey @patcon we think the overall premise is good, so if you can resolve the comments you have we'd be very happy to take this PR.

lib/chef/knife/bootstrap.rb
patcon: ah, the first pop is the host, regardless of "@", without which the second (user) will be nil. but perhaps that's too unspoken

ie.

```rb
"example.com".split('@').pop
```

patcon:
Ah oops, not sure how I missed these comments ages ago. I'll sort it out now

patcon:
Done and done.

lamont-granquist:
Looks like you have failing spec tests.

patcon:
@lamont-granquist 

> also not sure you need to check @name_args.nil? since that would cause the existing server_name method to explode

Ah, just realized I probably added that to make a unit test pass, which is now failing:
https://github.com/chef/chef/blob/master/spec/unit/knife/bootstrap_spec.rb#L525-L529

Is that a poorly-constructed unit test that should be removed, or should I add prepare for a nil case?

lamont-granquist:
No, looks like I was wrong, `Array(nil).first?` returns `nil` which is checked later via calling server_name so server_name probably does need to handle the nil case properly.

patcon:
Ok, should be good now. Also, it's kinda messy, so let me know if you'd like me to squash

lamont-granquist:
Yeah squash, rebase + force-push would probably be good.

patcon:
Ta-daaaaa

lamont-granquist:
works for me :+1: 

@thommay?

lamont-granquist:
And anyone else in @chef/client-core that would like to review.

thommay:
take it away :+1:

sethvargo:
:+1:

PR #3530, opened 2015-06-12
Allow using --sudo option with user's home folder in knife bootstrap

paulmooring:
Currently knife bootstrap with '--sudo' always sets the -H option to set ENV['HOME'] to root's home folder.  This adds a flag to change that behavior but doesn't change the default behavior.


lamont-granquist:
:+1:

thommay:
:+1:

Issue #3526, opened 2015-06-12
[knife] Assume a default key when no colon in QUERY


patcon:
The thought would be that this could be `name:` by default (or perhaps `fqdn:`?), but it could be set to something else in `knife.rb`

```
knife search example.com
knife ssh example.com "uname"
```

patcon:
On further consideration, using the fqdn key and NOT providing a means to set another default key would seem to imply that the feature in #3527 could equally apply to `knife ssh` (though not `knife search`):

```
knife ssh patcon@example.com "uname"`
```

Again, this would align with user expectations of the `ssh` command

lamont-granquist:
`knife search node "foo"` already goes through some magic to search name and fqdn and possibly a couple other fields by default.  that code could simply be reused for knife ssh.

patcon:
thanks! found it: https://github.com/chef/chef/blob/b08c9010c387c762a6fa82fa9ae8c2a5e5c8594d/lib/chef/knife/search.rb#L161-L165

I'll mess around with it later

patcon:
Hm. On consideration, "tags or role or fqdn or addresses" might be a little too general for a possibly destructive command, you think?

lamont-granquist:
I never run a destructive ssh command without validating the targets it will hit with a non-destructive command.  And the knife ssh command is necessarily a chainsaw, anyone with access to it is trusted to not run 'knife ssh "sudo rm -rf / &"'

patcon:
:) :+1: I'll look into tests then

sawanoboly:
@lamont-granquist @patcon 

I agree that the QUERY has a default key when `:` was omited.

And, It will enough for in most use cases, I think.

- for node => use `name:`,`fqdn:`, or `ipaddress:` by default
- for other than the node (such as roles, users) => use `name:` by default

patcon:
@sawanoboly Thanks for the feedback. @lamont-granquist said in the PR's inline comments that he preferred this share the same fuzziness as search, so I'm assuming that applies to all indexes. Correct me if I'm wrong though, @lamont-granquist

lamont-granquist:
Yes, and not just share the same semantics, but no code duplication.  Extract a method, if there isn't one already, and use it both places consistently.

lamont-granquist:
And dan is updating the knife search node fuzzifier method in #3913 for policyfiles now (which is precisely why the fuzzifier needs to exist only in one spot, not many)

PR #3536, opened 2015-06-15
Add fuzziness to `knife ssh` queries

patcon:
PR for #3526.

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -186,9 +186,12 @@ class Chef
       end
 
       def search_nodes
+        @query = @name_args[0]
+        fuzzify_query
+
         list = Array.new
-        query = Chef::Search::Query.new
-        @action_nodes = query.search(:node, @name_args[0])[0]
+        q = Chef::Search::Query.new
+        @action_nodes = q.search(:node, @query)[0]
         @action_nodes.each do |item|
           # we should skip the loop to next iteration if the item
           # returned by the search is nil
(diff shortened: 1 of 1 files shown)


lib/chef/knife/ssh.rb
```diff
@@ -204,6 +207,12 @@ def search_nodes
         list
       end
 
+      def fuzzify_query
+        if @query !~ /:/
+          @query = "tags:*#{@query}* OR roles:*#{@query}* OR fqdn:*#{@query}* OR addresses:*#{@query}*"
```
patcon: Should this be "weaker"? Possibilities:

1. Remove asterisks for both sides.
2. fqdn only? fqdn + name?
3. Keep as is, but have a confirmation prompt for what fqdn we're running  on. (Can be explicitly skipped with `--yes` flag.)

cc: @lamont-granquist

sawanoboly:
:+1:

lib/chef/knife/ssh.rb
lamont-granquist: It needs to be the same as knife search node.

patcon:
Also, my ruby force is apparently weak here -- any suggestions on how I could pull `#fuzzify_query` from `lib/chef/knife/search`? (this is where you all probably judge me :)

patcon:
Also, went searching for tests that cover this part, but doesn't seem we test this level, so given that I'm not comfortable enough with starting my own test approach from scratch, I think this is just waiting on feedback on the above, then it's good to merge as far as I'm aware.

lamont-granquist:
So, what I'd really like to see is knife ssh and knife node search share exactly the same code and call the same method that does fuzziness.  Eliminate code duplication and eliminate the chances that we patch in one place and not the other.

lamont-granquist:
We can't merge this without considering the problem holistically, so :-1: for now.

coderanger:
Agree, and for the future the Manage code has a similar fuzzifier and should make sure that uses the same algo.

patcon:
Totally fair. Closing for now

Issue #3078, opened 2015-03-14
Compat break in audit-mode changes


coderanger:
Making this to track https://github.com/chef/chef/commit/4cfb1e47aa8e9501f4f2a01f1d8cc0deb2cfa13b#commitcomment-10197616

This has ramifications for ChefSpec testing as you can no longer test for errors being raised in recipe code (and you get get log output during your tests too). Should be reverted pending Chef 13, if ever.

jdmundrawala:
Related to https://github.com/chef/chef/issues/3054

jdmundrawala:
cc @adamedx @btm @charlesjohnson

tyler-ball:
I understand how this is a breaking change.  For context, we decided to do this because we wanted the audit phase to run independently of the converge phase.  It should run even if the converge phase fails.

If there are errors during both the converge phase and audit phase, we don't have a way of raising both errors.  Thats why I added the wrapping exception.

Now that I am looking at this though, it seems unecessary to wrap ALL errors.  For an improvement, what would you think about always running both phases and capturing errors from each phase.  Then we only wrap both errors if both occured.  Otherwise we re-raise the error if only 1 phase had an error.

For anyone not using audit mode (a new feature introduced in Chef 12) this would maintain backwards compatibility with Chef 11 behavior.

What is the issue with increasing the logging output?  The exceptions are still being sent to the formatting system and I found the increased logging useful for development and debugging.

coderanger:
You should have something above the level of `#converge` that handles the catch and audit stuffs. `#converge_with_audit` or something, and make it behave as you describe. As for the logging, because Chef is not always used interactively which is why the formatter system was created. If you need to add a new formatter callback for output, do that (or use the existing generic ones). Logging should be for debugging only basically (/cc @danielsdeleo).

btm:
As a general rule, swallowing exceptions is bad because so many other tools in the Chef ecosystem treat us as a library. I agree with @jdmundrawala that we should ```raise e if Chef::Config[:audit_mode] == :disabled``` right now and ship that ASAP in the 12.1.2 regression release.

Then we talk about how audit mode gets implemented here when it's no longer an experimental feature.

btm:
This should be fixed in [12.1.2](https://www.chef.io/blog/2015/03/23/chef-client-12-1-2-released/).

Issue #3054, opened 2015-03-11
Chef 12.1 not allowing control of exit codes with exceptions


carpnick:
This ruby block below in Chef 11.* makes chef-client exit with exit code of 250.  In Chef 12 it exits with  1.  The need from a user perspective, is we need to be able to signal a Continuous Delivery system with exit codes out of Chef.  We were depending directly on the exit code in this example, to wait for WinRM or SSH to start running chef again.  The exit code will let us know what action was performed in Chef.  Reboot, Chef run success, etc.
```
ruby_block "ExitNow" do
  block do
     puts "Stopping chef run for an immediate reboot";       
     exit 250;  
  end
  action :run
end
```

Looks like Chef 11's implementation of [Converge](https://github.com/chef/chef/blob/11-stable/lib/chef/client.rb#L341) would raise the exception where in Chef 12 it [catches it](https://github.com/chef/chef/blob/12-stable/lib/chef/client.rb#L324) and returns it to the [run method] (https://github.com/chef/chef/blob/12-stable/lib/chef/client.rb#L448) and then [wraps it](https://github.com/chef/chef/blob/12-stable/lib/chef/client.rb#L461).

I think the potential fix is inside the code where it is wrapping it:
```
if converge_error.class == SystemExit
  raise
end
```

carpnick:
Other option is to have the [chef reboot] (https://docs.chef.io/resource_reboot.html) resource have an option on it to reboot with a specific exit code.

I would rather have this option, but I think it would cause the above comment code change anyways.

pburkholder:
Adding an explicit 'Exit' resource to Chef language would be a solution too.

jdmundrawala:
cc @tyler-ball @mcquin

tyler-ball:
We specifically catch and wrap the exception so we can run the audit phase independently of the converge phase.  So I don't want to implement a re-raise on SystemExit error because that would skip the audit phase.  Is using `exit \d` a typical pattern for flow control?  It doesn't seem to me like it should be.

Thats why I would rather have the reboot resource updated to take in an optional exit code that defaults to 0.

Other opinions?

carpnick:
Updating the reboot resource immediately solves our problem.  

However, I can see other use cases where a user might want to exit chef mid chef-run and not only because of a reboot to signal another system.  Network unavailable, system out of space, etc.

[...]

PR #3538, opened 2015-06-16
Exceptions for audits should only get wrapped if audit mode is enabled

jdmundrawala:
cc @jtimberman @sersut @mcquin

### DIFF
--- a/spec/unit/client_spec.rb
+++ b/spec/unit/client_spec.rb
@@ -238,23 +238,24 @@ describe Chef::Client do
     describe "when converge completes successfully" do
       include_context "a client run"
       include_context "converge completed"
-
-      describe "when audit phase errors" do
-        include_context "audit phase failed with error"
-        include_examples "a completed run with audit failure" do
-          let(:run_errors) { [audit_error] }
+      context 'when audit mode is enabled' do
+        describe "when audit phase errors" do
+          include_context "audit phase failed with error"
+          include_examples "a completed run with audit failure" do
+            let(:run_errors) { [audit_error] }
+          end
         end
-      end
 
-      describe "when audit phase completed" do
-        include_context "audit phase completed"
-        include_examples "a completed run"
-      end
+        describe "when audit phase completed" do
+          include_context "audit phase completed"
+          include_examples "a completed run"
+        end
 
-      describe "when audit phase completed with failed controls" do
-        include_context "audit phase completed with failed controls"
-        include_examples "a completed run with audit failure" do
-          let(:run_errors) { [audit_error] }
+        describe "when audit phase completed with failed controls" do
+          include_context "audit phase completed with failed controls"
+          include_examples "a completed run with audit failure" do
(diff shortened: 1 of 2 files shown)


mcquin:
Wrapping chef client run errors helps with consistency in the format of our
error messaging. I think it could be confusing to have two separate errors
reported (the error, and the wrapped error) depending on whether or not
audit mode is enabled. And for users who care what kind of error is thrown
(if such users exist?), it would be annoying to have to know that you have
to check for the error or a RunFailedWrappingError containing the error.

jdmundrawala:
@mcquin I think that this would need a major version bump:
https://github.com/chef/chef/issues/3054
https://github.com/chef/chef/issues/3078

mcquin:
That doesn't address my concern of breaking consistency in our error
messaging.

What I think we should be doing, is proper exit coding based off of the
contained errors.

jdmundrawala:
@mcquin that's a fair point, however we should wait until chef 13 to make this change. We counted it as a regression in 12.2, and we know it will break people by wrapping the exception

mcquin:
OK. That this was resolving the regression wasn't clear.

Where can we make a note that in Chef 13 we should unify run errors and fix
exit codes?

spec/unit/client_spec.rb
```diff
@@ -512,11 +513,26 @@ def stub_for_sync_cookbooks
...
+      it "should run exception handlers on early fail" do
+        expect(subject).to receive(:run_failed)
+        expect { subject.run }.to raise_error(Chef::Exceptions::RunFailedWrappingError) do |error|
+          expect(error.wrapped_errors.size).to eq 1
+          expect(error.wrapped_errors).to include(NoMethodError)
+        end
+      end
+    end
+
+    context 'when audit mode is enabled' do
```
tyler-ball: `'when audit mode is disabled'`

spec/unit/client_spec.rb
```diff
@@ -238,23 +238,24 @@ def stub_for_sync_cookbooks
     describe "when converge completes successfully" do
       include_context "a client run"
       include_context "converge completed"
-
-      describe "when audit phase errors" do
-        include_context "audit phase failed with error"
-        include_examples "a completed run with audit failure" do
-          let(:run_errors) { [audit_error] }
+      context 'when audit mode is not disabled' do
```
tyler-ball: (nit) double negatives are bad, mmm'kay.  `'when audit mode is enabled'`

tyler-ball:
Some minor comments, but overall :+1: for fixing the regression.  We should be able to add a test with the `:chef_gte_13_only` flag and assert that even when audit mode is disabled we have the wrapped error.  We can totally change this behavior in Chef 13, but at least then we would have a failing test to remind us that we need to have a discussion.

spec/unit/client_spec.rb
jdmundrawala: m'kay

ranjib:
:+1:

Issue #3531, opened 2015-06-12
registry_key resource is case sensitive in chef but not on windows


bigbam505:
If you use the `registry_key` resource on windows it can keep trying to run resources because of incorrect case.  For example, I was trying to set the chef-client service to run as delayed startup.  The code I had was this:

```ruby
 registry_key 'HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\chef-client' do
   action :create
   values [ {
      :name => 'DelayedAutoStart',
      :type => :dword,
      :data => 1
      } ]
end
```

However I noticed that this would keep creating the registry key on every chef run.  Upon further investigation I found that the windows service manager sets a key `DelayedAutostart` (notice the lowercase `s`).  If I change my code to the include the following it works fine.

```ruby 
not_if { registry_value_exists?('HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\chef-client', { name: 'DelayedAutoStart' }) }
```

Not sure if this has been addressed but I did not see any issues opened for this.  I ran into this with chef-client 12.2.1 and also chef-client 12.3.0 in test-kitchen

Tech356:
Also reported here: https://tickets.opscode.com/browse/CHEF-5196

jkerry:
Working on a fix on my fork.  I should be out of manual qa and get a pull request in shortly

lamont-granquist:
#3539 is merged, so assuming this is fixed.

PR #3539, opened 2015-06-16
Bug#3531

jkerry:
Fix for issue: #3531 registry_key resource is case sensitive in chef but not on windows

* Changed the registry_key provider to hash downcased keys and compare with downcased key names to prevent unnecessary convergence.
* Changed the Win32.Registry class to compare keys and value names with a case insensitive comparison.
* Added functional and unit tests to cover these changes

This was manually tested locally with the recipe described in  #3531

### DIFF
--- a/lib/chef/win32/registry.rb
+++ b/lib/chef/win32/registry.rb
@@ -200,7 +200,11 @@ class Chef
       end
 
       def SafelyDowncase(val)
-        (val == nil ? nil : val.downcase)
+        begin
+          return (val == nil ? nil : val.downcase)
+        rescue NoMethodError
+          return val
+        end
       end
 
       def value_exists?(key_path, value)
--- a/lib/chef/win32/registry.rb
+++ b/lib/chef/win32/registry.rb
@@ -203,7 +203,7 @@ class Chef
         key_exists!(key_path)
         hive, key = get_hive_and_key(key_path)
         hive.open(key, ::Win32::Registry::KEY_READ | registry_system_architecture) do |reg|
-          return true if reg.any? {|val| val.downcase == value[:name].downcase }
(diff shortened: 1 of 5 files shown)


lib/chef/win32/registry.rb
```diff
@@ -199,11 +199,19 @@ def registry_system_architecture
         ( applied_arch == :x86_64 ) ? 0x0100 : 0x0200
       end
 
+      def SafelyDowncase(val)
```
jdmundrawala: we use snake_case

lib/chef/win32/registry.rb
```diff
@@ -199,11 +199,19 @@ def registry_system_architecture
...
+        begin
+          return (val == nil ? nil : val.downcase)
+        rescue NoMethodError
```
jdmundrawala: I'm not a fan of expecting such an exception.

jkerry:
Doh! A c# developer's first crack at Ruby. I'll get the method name fixed and look into the exception. It's possible to cast the object to a string first but that seems dirtier.

jkerry:
Corrected both comments.  I'm not sure if there's a more idiomatic ruby way to safely downcase in this situation. If there is please let me know and I'll correct it

jdmundrawala:
:+1: LGTM
cc @chef/client-windows

ksubrama:
:+1: LGTM as well.  Thank you!

pburkholder:
@jkerry :dancer: :balloon: Congrats on first Chef PR.

jkerry:
ht_sam_griner_then_02_jef_150413_16x9_992

ranjib:
:+1:

ksubrama:
Merged at 8c09a658cffd8ff8c654aab1d8ccff69342737c7

PR #3542, opened 2015-06-16
Only automatically set resources that do class X < Chef::Resource

### LINKED ISSUE #3524: Resource and provider resolver fallback non-functional
The code used for the non-`provides()` lookup on resources and providers is non-functional as far as I can tell.

[The code looks like](https://github.com/chef/chef/blob/master/lib/chef/resource_resolver.rb#L143):

```ruby
@prioritized_handlers ||= super ||
```

but `super` returns an empty array, not `nil` so the fallback code never runs.

Ping @jkeiser and @thommay (I think this is probably a release blocker for RC2)

jkeiser:
not X = Class.new(Chef::Resource)

This fixes #3524

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -924,11 +924,6 @@ class Chef
         else
           @resource_name = nil
         end
-      else
-        # set resource_name automatically if it's not set
-        if !instance_variable_defined?(:@resource_name) && self.name
-          resource_name convert_to_snake_case(self.name.split('::')[-1])
-        end
       end
 
       @resource_name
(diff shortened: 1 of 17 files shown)


lib/chef/resource.rb
```diff
@@ -1111,7 +1106,10 @@ def self.sorted_descendants
     def self.inherited(child)
       super
       @sorted_descendants = nil
-      child.resource_name
+      # set resource_name automatically if it's not set
+      if child.name && !child.resource_name
```
coderanger: This should have `child.name.start_with?('Chef::Resource::')`.

lib/chef/resource.rb
```diff
@@ -1111,7 +1106,10 @@ def self.sorted_descendants
...
+        child.resource_name convert_to_snake_case(child.name.split('::')[-1])
```
coderanger: No parens is :-(

jkeiser:
Parens fixed :)  I'll take a look at start_with tomorrow.

PR #3543, opened 2015-06-16
Warn when multiple providers try to provide the same thing

jkeiser:


jkeiser:
@lamont-granquist this should make your git cookbook issue more manageable.

lamont-granquist:
pretty certain you need those provides lines in those providers.  those are ambiguous, but you must go through all of those and call the `::supports?` method in the provider resolver and then use the priority array mechanism to determine which one to use.

probably why travis is annoyed at you?

jkeiser:
@lamont-granquist those are duplicates of lines here: https://github.com/chef/chef/blob/jk/multi-provider-warning/lib/chef/provider/service.rb#L203-L212

Pretty sure there is nothing the provides could help with there.

spec/unit/provider_resolver_spec.rb
```diff
@@ -482,7 +482,6 @@ def stub_service_configs(*configs)
     python: Chef::Provider::Script,
     remote_directory: Chef::Provider::RemoteDirectory,
     route: Chef::Provider::Route,
-    rpm_package: Chef::Provider::Package::Rpm,
```
lamont-granquist: that should still be in there shouldn't it?

lib/chef/resource/macports_package.rb
```diff
@@ -16,10 +16,11 @@
 # limitations under the License.
 #
 
+require 'chef/resource/package'
+
 class Chef
   class Resource
     class MacportsPackage < Chef::Resource::Package
-      provides :package, os: "darwin"
```
lamont-granquist: here you need to still provide package and the resolve the conflict later with the explicit priority setting.

facebook uses this, its an external API that both macports and homebrew declare 'provides' package on darwin.

spec/unit/provider_resolver_spec.rb
jkeiser: Nope, it's actually only defined for certain OS's.

lamont-granquist:
Yeah, looking at the code the `def self.provides?` takes priority over the `provides` statements still doesn't it?  Those lines for the service provider were always ignored?

spec/unit/provider_resolver_spec.rb
jkeiser: (specifically, there is a line further down that says rpm_package: nil, so all OS's verify that rpm_package does *not* work, except the OS's on which it is supposed to work. This preserves behavior before the patch, no changes.)

jkeiser:
@lamont-granquist I'm not sure what you mean.  If `provides` is specified for the resource, then `provides?` must *also* return true, but `provides?` will not be triggered (except for deprecated purposes) unless there is a `provides` for the resource.

lamont-granquist:
Hmmm, we're talking providers here not resources, and that would indicate `#provides?` was gonna be broken, but then tests should have failed.

lib/chef/resource/macports_package.rb
jkeiser: Can you elaborate?  I'm super unclear what is supposed to act differently when the `provides` statements are in there, given that you have to require `chef/resource/package` to compile this, and that file includes the set_resource_priority_map.

The confusion might be in what `def self.provides?` does?  If you call `Chef::Resource::MacportsPackage.provides?(<os x node object>, :package)`, it will go to the `ResourcePriorityMap` to find out whether it is provided or not (it lists all resource classes matching the node for `package`, and checks to see if MacportsPackage is one of them). So the priority array from package will actually affect this, and `provides?` will give a correct result.

spec/unit/provider_resolver_spec.rb
lamont-granquist: cool.

lamont-granquist:
okay yeah, in the new world all the changes make sense

:shipit:

PR #3548, opened 2015-06-18
Do not set resource_name outside of Chef::Resource

jkeiser:


coderanger:
:ship: :it: :shipit:

coderanger:
By way of description, this brings master back to the same behavior as 12.3. Further discussions of how this should work in the future are over at https://github.com/chef/chef-rfc/pull/136.

PR #3551, opened 2015-06-18
A missing space in a shell_out command in install_via_gem_command

tfitch:
The install_via_gem_command method is missing a space and causing failures when a specific version of a gem is requested.  This adds that space.

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -538,7 +538,7 @@ class Chef
             src = @new_resource.source && " --source=#{@new_resource.source} --source=https://rubygems.org"
           end
           if !version.nil? && version.length > 0
-            shell_out_with_timeout!("#{gem_binary_path} install #{name} -q --no-rdoc --no-ri -v \"#{version}\"#{src}#{opts}", :env=>nil)
+            shell_out_with_timeout!("#{gem_binary_path} install #{name} -q --no-rdoc --no-ri -v \"#{version}\" #{src}#{opts}", :env=>nil)
           else
             shell_out_with_timeout!("#{gem_binary_path} install \"#{name}\" -q --no-rdoc --no-ri #{src}#{opts}", :env=>nil)
           end
--- a/spec/unit/provider/package/rubygems_spec.rb
+++ b/spec/unit/provider/package/rubygems_spec.rb
@@ -507,7 +507,7 @@ describe Chef::Provider::Package::Rubygems do
 
         it "installs the gem by shelling out when options are provided as a String" do
           @new_resource.options('-i /alt/install/location')
-          expected ="gem install rspec-core -q --no-rdoc --no-ri -v \"#{@spec_version}\" -i /alt/install/location"
+          expected ="gem install rspec-core -q --no-rdoc --no-ri -v \"#{@spec_version}\"  -i /alt/install/location"
           expect(@provider).to receive(:shell_out!).with(expected, env: nil, timeout: 900)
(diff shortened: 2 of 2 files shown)


tfitch:
Whoa - I'm going to look at these test failures.  Sorry.

lamont-granquist:
Does this actually fix anything?  If it does, could you include a test for it.

It looks like src and opts always are either nil, or have a leading space.  That gets the spaces right and doesn't leave a trailing space when they're both nil like in your last 3 tests.

tfitch:
So I'm proxying this fix from a customer.  In their environment they were using a combination of arguments that triggered this line of code and it was failing for them.  So they went in to their local copy of the client, added this space and it worked where it wasn't working before.

And like you, it baffled me, because while it's kinda hidden, there is a space prepended to the src variable even though it doesn't readily look like there is.

I have some meeting time with the customer today, I'll reaffirm they've still got this patch on their local systems.  Thanks @lamont-granquist.

coderanger:
Sounds like voodoo debugging, this probably wasn't what actually fixed the issue.

tfitch:
That's entirely possible @coderanger, there are lots of moving parts in play here.  But based on my history with the customer I trust their debugging skills in a locked down environment that triggered the use of this line of code that most people won't encounter.

btm:
If we do in fact need a space, we need to know how to reproduce this so we can add a test that explicitly ensures we have the space. Otherwise we'll lose it in the future when someone presumes the extra spaces are cruft.

It looks like there's been a double space there since almost when the feature was added. 0759ce215d54e115ca83ac923b7244ee1262b67d had a single space and a month later it became a double space without explanation in 6179d4dc7796626baf0f0563515925636be829da. Looks like cruft still.

I spent some time testing and ensuring I hit that particular call to `shell_out_with_timeout!` and I couldn't cause anything other than success. Something like this:

```
gem_package "bob" do
  source "https://rubygems.org"
  version "0.5.0"
  gem_binary "/Users/btm/.rvm/rubies/ruby-2.1.4/bin/gem"
end
```

calls this: 
```
/Users/btm/.rvm/rubies/ruby-2.1.4/bin/gem install bob -q --no-rdoc --no-ri -v "0.5.0" --source=https://rubygems.org --source=https://rubygems.org
```

With no source you get:
```
/Users/btm/.rvm/rubies/ruby-2.1.4/bin/gem install bob -q --no-rdoc --no-ri -v "0.5.0"
```

So looks fine still.

lamont-granquist:
Should probably just use something like this:

https://github.com/chef/chef/blob/lcg/dpkg-multipackage/lib/chef/provider/package.rb#L523-L531

That hasn't been merged yet, though, and I might try to just push it directly into shell_out_with_timeout if I can...

lamont-granquist:
That got merged (and is now 404'ing) the API to use is here:

https://github.com/chef/chef/blob/6be47e6db1513824af52524b4d8bc83c32849d4a/lib/chef/provider/package.rb#L554-L562

tfitch:
Well the contact at the customer has moved jobs and they're not complaining about the issue.  They had monkey patched it locally anyways and no one else seems to be complaining about this issue so I will close this.

I think filing away this example in our brains in case anyone else encounters this potential edge case in the future is the best I can do at the moment.

Thanks everyone.

lamont-granquist:
yeah at some point i'll a_to_s those which will fix any edge cases the right way

lamont-granquist:
(although if anyone finds this and wants to leap in and fix them, feel free to submit a PR...)

PR #3554, opened 2015-06-18
Add missing require statement in resource_resolver

ranjib:
/cc @lamont-granquist

### DIFF
--- a/lib/chef/resource_resolver.rb
+++ b/lib/chef/resource_resolver.rb
@@ -18,6 +18,7 @@
 
 require 'chef/exceptions'
 require 'chef/platform/resource_priority_map'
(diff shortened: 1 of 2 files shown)


lamont-granquist:
This may collide with or duplicate work that @jkeiser is doing.

ranjib:
so should i close this and monkey path on the consumer side? chef-lxc is failing due to this, i can definitely patch things on that side, add a require statement there instead.

jkeiser:
I don't think I've fixed this; it should be OK to go by itself.

jkeiser:
:+1:

thommay:
:+1:

Issue #3561, opened 2015-06-20
API reversion: ProviderPriorityMap#priority


coderanger:
Looks like this method got removed in https://github.com/chef/chef/commit/9a09df111e614d3e66639740d796c0479ab9b63d#diff-0547a1dbce8a9b0e83b5285e9a0a31ec

Might have been accidental? It should just be an alias for `#set` given the current code. This was the only public API to declare provider priority ordering in 12.3 so I think removal would count as a disallowed API change.

Ping @jkeiser @thommay @lamont-granquist.

coderanger:
Also it isn't clear to me how to use the current NodeMap code to set a priority array. The old code built the array in reverse so as long as your `priority()` call was after the classes were loaded (which is required anyway to have symbols for them) you were fine. I _think_ the intent is that you would use the new `override` filter flag for this, which is annoying both because you can only ever do this once (a wrapper couldn't easily re-order things once an override is in the array). Also because this flag would only work on 12.4+ so code declaring a priority array would need to check the Chef version carefully (less bad than the wrapper issue but annoying).

As it stands calling `Chef.provider_priority_map.set(:name, [Two, One])` would end up appending that to the end of the node map key array, `[One, Two, [Two, One]]` effectively. The provider resolver calls `.flatten(1).uniq` on this value, so the end result is `[One, Two]`, totally ignoring our requested priority. This also means my proposed fix is probably wrong :-) A hamfisted solution would be to make `#priority` work like `#set` _except_ it defaults to adding to the front of the list instead of the back. That seems like the most logical behavior for the use case of explicit priority orderings.

coderanger:
Trying to trace through the `set_priority_array` code, and I'm not sure this is going to have the expected behaviors. There is a check for `if new_matcher[:value].is_a?(Array) != matcher[:value].is_a?(Array)` but in a priority map the value should always be an array. I can't tell if this matters or not, but we should maybe quickly write down the expected behaviors for some common cases.

coderanger:
It seems like the effective behavior of `set_priority_array` given no filters et al is to just always prepend the value to the front of the array, which matches the 12.3 behavior but the code is sufficiently more complex that it doesn't seem intentional. There is some logic to alpha sort stuff, but it checks for `if new_matcher[:value].is_a?(Class)` which again should always be false since the value is an array.

jkeiser:
Was fixed with #3562

PR #3562, opened 2015-06-20
Re-add #priority.

coderanger:
Possible fix for https://github.com/chef/chef/issues/3561.

### DIFF
--- a/lib/chef/platform/priority_map.rb
+++ b/lib/chef/platform/priority_map.rb
@@ -4,7 +4,7 @@ class Chef
   class Platform
     class PriorityMap < Chef::NodeMap
       def priority(resource_name, priority_array, *filter)
-        set(resource_name.to_sym, Array(priority_array), *filter)
+        set_priority_array(resource_name.to_sym, priority_array, *filter)
       end
       
       # @api private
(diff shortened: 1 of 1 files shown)


jkeiser:
Yuck!  Is someone using it?  I thought this was @api private.

coderanger:
Back in 12.2 it was the only API to add stuff, and definitely not private: https://github.com/chef/chef/blob/12.2.0/lib/chef/platform/provider_priority_map.rb#L77-L79

coderanger:
12.3 added `set_priority_array` as basically an alias for `priority` but it still wasn't private.

jkeiser:
Right then, can't remove it. :+1:

coderanger:
Also `get_priority_array` and `set_priority_array` probably shouldn't be private.

jdmundrawala:
do we want to ask people not to use this?

coderanger:
Deprecating it would be fine but printing a warning might annoy people since the code to switch to a different-but-identical function that didn't exist in 12.2 is kind of aggravating.

jdmundrawala:
:+1:

Issue #3553, opened 2015-06-18
knife segmentation fault on YN prompt


alexpop:
I've installed chef-dk 0.6.2 in a Windows 7(64 bit) VM and commands that ask the "Do you really want" question throw a segmentation fault error.

```
PS C:\chef-repo> chef -v
Chef Development Kit Version: 0.6.2
chef-client version: 12.3.0
berks version: 3.2.4
kitchen version: 1.4.0
 
PS C:\chef-repo> knife role list
linux_web
linux_web_prod
windows_base
 
PS C:\chef-repo> knife role delete windows_base
Do you really want to delete windows_base? (Y/N) C:/opscode/chefdk/embedded/apps/chef/lib/chef/knife/core/ui.rb:235: [BUG] Segmentation fault
ruby 2.1.5p273 (2014-11-13 revision 48405) [i386-mingw32]
```

See complete stack trace here:
https://gist.github.com/alexpop/b8aa705b3abcc0861d47

lamont-granquist:
Can you replicate with the chef MSI (not chef-dk) that ships with ruby 2.0.0 instead of 2.1.5?  (I'm not sure how the windows ruby wound up on 2.1.5 when we haven't moved the chef ruby to 2.1.5 yet and that seems like it might be at the root of the problem).

alexpop:
Sure, I uninstalled chef-dk 0.6.2 and then installed chef-12.3.0-1.el6.x86_64.rpm

No more seg fault on deleting objects:
```
PS C:\chef-repo> knife role delete windows_jenkins
Do you really want to delete windows_jenkins? (Y/N) y
Deleted role[windows_jenkins]

PS C:\chef-repo> C:\opscode\chef\embedded\bin\ruby.exe -v
ruby 2.0.0p451 (2014-02-24) [i386-mingw32]

PS C:\chef-repo> knife -v
Chef: 12.3.0
```

lamont-granquist:
So this is sort of a chef-dk bug and is misfiled then.

lamont-granquist:
technically this a blocker for us to get onto a later ruby, but that's a separate yak shave, and this ticket is a bad way to track that yak shave, so i'm going to close this bug.

PR #3568, opened 2015-06-23
12.4.0.rc.3

jdmundrawala:
cc @chef/client-maintainers


lamont-granquist:
#3561 probably needs fixing.

thommay:
also we should confirm or deny #3553

jdmundrawala:
#3553 is unrelated to any of the changes that are going into 12.4.0
We shouldn't block for it

jdmundrawala:
#3561 has a fixed that is merged in

jkeiser:
:+1:

danielsdeleo:
I'm not super in the loop right now about regressions but this change itself is fine :+1:

tyler-ball:
:+1: 12.4.0 is out so I think this is only doc changes now

Issue #2666, opened 2014-12-16
Chef 12 errors on Windows when role_path doesn't include leading drive letter


sneal:
I took a working provisioned Windows box and upgraded from Chef 11 to Chef 12 (without changing anything else) and now I get a "Missing Role(s) in Run List" error. See Vagrant issue https://github.com/mitchellh/vagrant/issues/4974 for more information, but basically Chef 12 now requires that role paths include a leading drive letter which wasn't previously required.

```
==> default: Running chef-solo...
==> default: [2014-12-16T11:03:23-08:00] INFO: *** Chef 12.0.1 ***
==> default: [2014-12-16T11:03:23-08:00] INFO: Chef-client pid: 2848
==> default: [2014-12-16T11:03:49-08:00] INFO: Setting the run_list to ["role[dotnetframework]"] from CLI options
==> default: [2014-12-16T11:03:51-08:00] ERROR: Role dotnetframework (included by 'top level') is in the runlist but does not exist. Skipping expand.
==> default: 
==> default: 
==> default: ================================================================================
==> default: Error expanding the run_list:
==> default: ================================================================================
==> default: 
==> default: Missing Role(s) in Run List:
==> default: ----------------------------
==> default: * dotnetframework included by 'top level'
==> default: 
==> default: Original Run List
==> default: -----------------
==> default: * role[dotnetframework]
==> default: 
==> default: [2014-12-16T11:03:51-08:00] FATAL: Stacktrace dumped to C:/var/chef/cache/chef-stacktrace.out
==> default: [2014-12-16T11:03:51-08:00] FATAL: NoMethodError: undefined method `run_id' for nil:NilClass
```

Setting role_path to `/tmp/vagrant-chef-3/chef-solo-2/roles` doesn't work in Chef 12, but works in Chef 11.

Setting role_path to `c:/tmp/vagrant-chef-3/chef-solo-2/roles` works in Chef 11 and 12.

shackit:
Thanks @b-dean just tested that work around and it works for me.

Scythril:
I'm having the same issue with data_bag_path using vagrant and chef-solo; found this bug after figuring out that prepending the drive letter fixes it, but doesn't help me since Vagrant replaces the chef config on every provision.

Another note is that without the drive letter, it still verifies that the path exists, it just doesn't grab the files within that path without prepending the drive letter.

I'm currently using a combination of work-arounds from @b-dean and @lonniev (thank you both!), which is putting

```ruby
[...]

PR #3569, opened 2015-06-23
Fix for issue #2666

NickMRamirez:
* Added method to path_helper.rb to convert Linux-style absolute paths to Windows-style absolute paths that begin with a drive letter
* Moved methods that finds roles files into a separate method and then added call to new path_helper method to fix Windows issue [#2666](https://github.com/chef/chef/issues/2666)
* Added unit tests
* Tried with Vagrant script against Windows VM (Virtualbox), checking that add_role and add_recipe work

### DIFF
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -228,6 +228,17 @@ module ChefConfig
         joined_paths
       end
     end
+	
+    # On Windows, expands the given path so that it
+    # contains a drive letter at the beginning.
+    # [path] The path to convert
+    def self.get_full_windows_path(path)
+      if ChefConfig.windows?
+        path = cleanpath(File.absolute_path(path))
+      end
+      path
+    end
+	
   end
 end
 
--- a/chef-config/spec/unit/path_helper_spec.rb
+++ b/chef-config/spec/unit/path_helper_spec.rb
@@ -101,6 +101,18 @@ RSpec.describe ChefConfig::PathHelper do
     it "cleanpath does not remove leading double backslash" do
       expect(path_helper.cleanpath('\\\\a/b\\c/d/')).to eq('\\\\a\\b\\c\\d')
     end
+    
+    describe 'get_full_windows_path' do
+      it 'expands path to begin with drive letter' do
+        expect(File).to receive(:absolute_path).with('\\a\\b\\c').and_return('C:\\a\\b\\c')
+        expect(path_helper.get_full_windows_path('\\a\\b\\c')).to eq('C:\\a\\b\\c')
+      end
+      
+      it 'appends relative path to current directory' do
+        expect(File).to receive(:absolute_path).with('a\\b\\c').and_return('C:\\x\\y\\z\\a\\b\\c')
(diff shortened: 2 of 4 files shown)


chef-supermarket:

* @NickMRamirez

chef-supermarket:

chef-config/lib/chef-config/path_helper.rb
```diff
@@ -228,6 +228,17 @@ def self.all_homes(*args)
         joined_paths
       end
     end
+	
+    # On Windows, expands the given path so that it
+    # contains a drive letter at the beginning.
+    # [path] The path to convert
+    def self.get_full_windows_path(path)
+      if ChefConfig.windows?
+        path = cleanpath(File.absolute_path(path))
```
adamedx: This seems unreliable -- it relies on the process current directory, which may differ depending on how Chef is launched. It can also be altered by code (including Chef recipes) that changes the current working directory.

I think for this conversion to work deterministically and to be useful, an unchanging and predictable definition of a path like '/x/y/' on Windows must be defined. Without that, I would actually argue in favor of warnings when such a path is used and eventually (after a major version change) an exception. 

@btm, @jdm, @ksubrama, thoughts? How does `kernel32!CreateFile` behave in this situation?

adamedx: Here's a chef recipe that shows why `::File.absolute_path` can cause problems -- if you run chef-client from a different directory, you get a different path. This would be very difficult to debug :smile: .

```ruby
log ::File.absolute_path('/a/b/c/d')
```
Here's the output:

% adamed@SEA-ADAMED1: 0x00000000: C:\Users\edwardsa
2015.06.23 Tue 06:03:42.03-07:00> chef-apply C:\test\ambiguous_path.rb
  * log[C:/a/b/c/d] action write

```sh
% C:\Users\edwardsa
> chef-apply C:\test\ambiguous_path.rb
  * log[C:/a/b/c/d] action write
% C:\Users\edwardsa
> z:
% Z:\adamedwards\hosted\chef-repo
> chef-apply C:\test\ambiguous_path.rb
  * log[Z:/a/b/c/d] action write
```
Notice that in one case, the path starts with `c:/` and in the other it starts with `z:/`.

NickMRamirez:
In other parts of the code, I've seen the use of `env['SYSTEMDRIVE']` when figuring out paths. If we don't use absolute_path, then possibly that?

NickMRamirez:
I am now wondering if this is an issue with Vagrant when using a Windows guest and not an issue with Chef. There's a Vagrant config option called provisioning_path.

[Shared Chef Options](http://docs.vagrantup.com/v2/provisioning/chef_common.html):
> provisioning_path (string) - The path on the remote machine where Vagrant will store all necessary 
> files for provisioning such as cookbooks, configurations, etc. This path must be world writable. By 
> default this is /tmp/vagrant-chef-# where "#" is replaced by a unique counter.

Given that, in my experiments, the path `/tmp/vagrant-chef` is being used on a Windows guest, maybe Vagrant should have a check for Windows and use a different path by default when on Windows?

I will experiment some more.

PR #3582, opened 2015-06-23
Handle identity and state_properties, create real Property class

jkeiser:


lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
+require 'chef/exceptions'
```
coderanger: Needs a license header.

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+
+        if has_default?
+          value = default
+          if value.is_a?(DelayedEvaluator)
+            value = exec_in_resource(resource, value)
+            value = coerce(resource, value)
+            # We don't validate defaults
+            set_value(resource, value)
+          else
+            value = coerce(resource, value)
```
coderanger: As discussed before, this should freeze the coerced output if value was frozen to start with.

This covers the pathological case of `property(:foo, default: 1, coerce: proc { [] })` and `resource.foo << 123`. As written currently the array would be thrown away afterwards so the mutation would fizzle.

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+    # @return [Boolean] Whether this value explicitly accepts nil.
+    #
+    # @api private
+    def explicitly_accepts_nil?(resource)
+      options.has_key?(:is) && resource.send(:_pv_is, { name => nil }, name, options[:is], raise_error: false)
+    end
+
+    def get_value(resource)
+      if instance_variable_name
+        resource.send(:instance_variable_get, instance_variable_name)
```
coderanger: With this and the other `instance_variable_*` calls, not sure where there is a `send`. They aren't private methods that I know of?

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+      end
+
+      if value.is_a?(DelayedEvaluator)
+        value = coerce(resource, value)
+        validate(resource, value)
+      end
+      value
+    end
+
+    attr_reader :options
```
coderanger: Should probably go up towards the top of the class to match normal ruby style.

lib/chef/mixin/params_validate.rb
```diff
@@ -84,91 +119,8 @@ def lazy(&block)
       end
 
       def set_or_return(symbol, value, validation)
-        symbol = symbol.to_sym
```
coderanger: This sea of red make me unreasonably happy. :star2: :star2: :star2:

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+    #   @option options [Proc] :coerce A proc which will be called to
+    #     transform the user input to canonical form. The value is passed in,
+    #     and the transformed value returned as output. Lazy values will *not*
+    #     be passed to this method until after they are evaluated. Called in the
+    #     context of the resource (meaning you can access other properties).
+    #   @option options [Boolean] :required `true` if this property
+    #     must be present; `false` otherwise. This is checked after the resource
+    #     is fully initialized.
+    #
+    def initialize(**options)
```
coderanger: One thing I would bring over from my version is having a `Property.create` class method that by default just delegates to `Property.new`. That would give a relatively easy place for Interesting Code™ to call `Module.prepend` to inject some interesting behavior, specifically that they could check for specific arguments that came from `Resource.property` and return a subclass of `Property`. Poor mans plugin registration system :-)

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+    # resource.myprop # get
+    # resource.myprop value # set
+    # ```
+    #
+    # If multiple values or a block are passed, they will be passed to coerce.
+    # If there is no coerce method and multiple values are passed, we will throw
+    # an error.
+    #
+    # @param resource [Chef::Resource] The resource to get the property from.
+    # @param value The value to set the property to. If not passed or set to
```
coderanger: YARD has a nice `@overload` tag for this kind of method.

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+    #
+    # @raise Chef::Exceptions::ValidationFailed If the value is invalid for
+    #   this property, or if the value is required and not set.
+    #
+    def get(resource)
+      if is_set?(resource)
+        value = get_value(resource)
+        if value.is_a?(DelayedEvaluator)
+          value = exec_in_resource(resource, value)
+          value = coerce(resource, value)
```
coderanger: Same problem as below re: mutable coerced values, but given the situation this is much much less likely to be hit and could possibly be ignored.

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+    # Specialize this Property by making a duplicate with some added or
+    # changed options.
+    #
+    # @param options [Hash<Symbol,Object>] List of options that would be passed
+    #   to #initialize.
+    #
+    # @return [Property] The new property type.
+    #
+    def specialize(**modified_options)
+      Property.new(**options, **modified_options)
```
coderanger: I'm actually not sure what order these get merged in, maybe make it clearer with `**options.merge(modified_options)`? (I think that's the same effective behavior).

lib/chef/property.rb
```diff
@@ -0,0 +1,491 @@
...
+      end
+      value
+    end
+
+    attr_reader :options
+
+    # Used by #set_or_return to avoid emitting a deprecation warning for
+    # "value nil"
+    # @api private
+    class NonDeprecatedNilGetter < Property
```
coderanger: Any major reason to not just make this a config option on property instead of a subclass?

lib/chef/resource.rb
```diff
@@ -130,6 +130,135 @@ def initialize(name, run_context=nil)
...
+      # Inherit from the current / parent property if type is not passed
+      if type == NOT_PASSED && properties[name]
+        local_properties[name] = properties[name].specialize(declared_in: self, **options)
+      else
+        local_properties[name] = Property.new(name: name, declared_in: self, **options)
+      end
+
+      begin
+        class_eval <<-EOM, __FILE__, __LINE__+1
+          def #{name}(value=NOT_PASSED)
```
coderanger: Can we proxy through `&block` on both of these for future proofing?

coderanger:
Dance

coderanger:
Also to give a more concrete example for the Property subclass plugin-y stuffs, the Poise OptionCollector and TemplateContent helpers are both potential candidates for that model.

lib/chef/property.rb
jkeiser: Yep, or we should sticky all non-frozen values based on whatever comes out of the default+coerce.  I am thinking specifically about this case:

```ruby
property :numbers, Array, default: 1, coerce: { |v| v.is_a?(Array) ? v : [ v ] }
```

That is going to be a common coercion (so that users can specify `numbers 1`), and it might be reasonable for the user to want to still do `<< 2` and see `[ 1, 2 ]` in the resource.

lib/chef/mixin/params_validate.rb
jkeiser: If only we could move validate safely over too!

lib/chef/property.rb
jkeiser: The issue here is it's `protected`--either I'd have to make it public (might be ok, I'm just trying to be super ruthless with that, and I wanted to allow the possibility that we might store these in instance variables in the future), or make a little protected section up top.  For now, I'm moving it up to the top of the protected section.

lib/chef/property.rb
jkeiser: Yep, I generally agree there should be one of these--what is the disadvantage of them just overriding `new`?

coderanger: It seems wrong-er to have `new` return an instance of a subclass, and would mean there is no way to get around the plugin patching if you do really want an instance of a specific class.

lib/chef/property.rb
coderanger: I think it would be better to keep the stickiness linked to the default value, I would expect `property(:name, default: VALUE, coerce: PROC)` to be semantically equivalent to `property(:name, default: PROC(VALUE))`, and if you did `property(:name, default: [1])` we would freeze that.

lib/chef/resource.rb
jkeiser: We should probably pass *args too

coderanger: :+1: Might as well on the `=` methods too I guess, in case of really weird mixins :-)

jkeiser:
Closing in favor of #3624 (I will respond to your comments here though, and make changes in the new one.)

lib/chef/property.rb
jkeiser: I think we need to write out what the use cases for stickying are.  Like, if the primary use case is to be able to do things like `foo 'bar' do; list_of_stuff << 10; end`, then the constant-ness of the default is less the thing than the constantness of the coerced thing.  Likewise, if a use case is something like this:

```ruby
property :super_thingy, Thingy, default: 1, coerce: { |v| v.is_a?(Fixnum) ? Thingy.new(v) : v }
```

Where Thingy might be something you don't really *want* frozen, ever.  The fact that the input is an always-frozen thing (like a symbol or a number) shouldn't affect the fact that the coerced thing isn't.

lib/chef/property.rb
jkeiser: Well I'll be damned.  I'll fix that.

lib/chef/property.rb
jkeiser: Is `create` a typical name for factory things like this?

lib/chef/property.rb
jkeiser: Yeah, they are merged the same way hashes normally are: blah(a: 1, a: 2) == { a: 2 }

lib/chef/property.rb
jkeiser: Mainly that I don't want to imply that it's a semi-permanent property :)  The long name and class-ness here make it clear that it's not part of the main implementation.

lib/chef/resource.rb
jkeiser: The only fun thing about `&block` is when you want to instance_eval a proc and pass it *args and &block, you have to do some fun stuff with binding methods.  If I want to pass a, b, c, &d to my_proc, `instance_exec(a, b, c, &my_proc)` doesn't cut it--there's no place to put &d.

Best way I've found is to bind the proc to the instance and call it that way.

Passing the block as a parameter is possible but distasteful, just because that acts way differently than other Ruby calls with a block and would be surprising people.

jkeiser:
I definitely think of Property as something people will override to create a richer type system.  Regarding `Property.create`, there is something to be said for the alternative of "override `property` on your class."  That lets you create modules or classes that have a rich Property syntax, without polluting the world with them.  I think Poise would already be able to do something like that, since users have to opt in to Poise when declaring a Resource.

coderanger:
The problem with overriding `#property` is that has a ton of complex behavior in it and the actual `Property.new` is buried pretty deep inside it. That would probably mean a major copy-pasta to swap just the class creation out. We could move the actual object creation to a class method on the resource, but that seems like it is only mildly better.

jkeiser:
@coderanger actually, my hope there was to make a companion `property_type` method to create reusable property types; overriding that would be a ton simpler.

lib/chef/property.rb
jkeiser: If it's much much less likely to be hit, we should probably change the behavior to be identical to default stickiness :)

lib/chef/property.rb
coderanger: List of examples

lib/chef/property.rb
coderanger: :+1:

lib/chef/property.rb
coderanger: Sure, that would be the train of thought that we should sticky all coerced output to allow for mutable return values. That would be fine, just won't match the `default: PROC(VALUE)` semantics I mentioned. Either way is fine, and freezing them keeps the semantics to a smaller set of rules.

coderanger:
Sure, having that in a class method on Resource would accomplish the same thing and would allow for slightly more scoping. :+1:

PR #3584, opened 2015-06-24
Composite resource: treat a group of resource as a single resource

ranjib:
This PR adds a new resource type `composite` . Its a wrapper over LWRP with use_inline_resources. This will allow users to group resources together and send/receive notifications against them. Following is one of our current usage of composite resource:
We drop kernel module specific configs, each config file represents a module. and we want to rebuild the initramfs if any of the file is changes, immediately

```ruby
composite 'kernel modules' do
  resources do
    node['pd-kernel']['module_blacklist'].each do |mod|
      cookbook_file "/etc/modprobe.d/blacklist_#{mod}.conf" do
        owner 'root'
        group 'root'
        mode 0644
      end
    end
  end
  notifies :run, 'execute[run depmod]', :immediately
  notifies :run, 'execute[/usr/sbin/update-initramfs -u]', :immediately
end
```
we cant use delayed notification as we want the initramfs built immediately, not at the end, also we dont want to build it if the run fails (delayed notifications run even if the chef run fails).
There are several other uses (like iptables rule, nginx modules etc) where this can be useful.
let me know what you all think


lamont-granquist:
This probably collides conceptually with stuff that @jkeiser is doing with resources.  We've discussed opening up a block of resources with a `recipe "..." do .... end` block and also allowing for the recipe to take parameters and both take injected resources and be injectable (as well as subclass-able probably).

That framework would be more powerful that this and would most likely immediately deprecate this when it ships.

coderanger:
This should also probably be an RFC first given the scope.

jkeiser:
This looks sort of like resource_group (an rfc that's already been accepted) ... the idea being that you could take a non parameterized group of resources together and subscribe to the group rather than an individual resource in it. Additionally, once they are grouped this way you can set things like the execution model for the group.

I feel like recipe is a better name too, and I do think this one is worth an rfc.

jkeiser:
If it isn't super clear, I'm happy something like this is getting written! Thanks @ranjib!

lamont-granquist:
Yeah definitely everyone wants to see something like this.

ranjib:
@lamont-granquist our internal version of this uses `recipe` attribute, but i think thats misleading as you cant do include_recipe , also intention of this resource is to just aggregate other resources in an opaque way (so that you'll never get resource cloning related issues, since they are inline). If you can provide some example on what you mean by injected resources that will be helpful (like resources %w(foo[bar] foo[baz])?)

@jkeiser that RFC on resource_group is devised for parallel execution, not for making a composite, we can implement a resource group with composite may be.. but that RFC does not talk about enforcing use_inline_resources semantics.

@coderanger i didn't felt we need an RFC as its not any major feature, as it neither breaks anything nor introduces lot of code changes. Anyway, if all of you have inputs on this, its better to discuss over an RFC. i'll try to jot down one by EOD

lamont-granquist:
Not a recipe attribute, a top-level DSL method named "recipe" or "Chef.recipe" or something similar.  When you open it up it would create something like an LWRP with use_inline_resources but which was named and behaved more singleton-like similar to a recipe.  Ideally it also takes parameters and allows dependency injection of other resources.  It would probably need some better accumulator design baked in than the current design of include_recipe though.  Ideally you could fold in the parallellization of the resource_group stuff into this as well and just have the one way of grouping that satisfied the different needs.  And if you made it so that you could turn off the "singletonness" of it then it might also replace LWRPs in separate files with "RPs" declared directly into recipe code.

I don't know if the fully crazy of all that is possible, but that's the direction everyone is heading towards and it makes a lot of sense to me.  It means to me that we create something a bit more like a functional programming language out of the recipe DSL rather than dropping everything into the "main" of a scripting language like we do now.

ranjib:
@lamont-granquist aha :-) i get it now. will start working on an RFC and close this

PR #3586, opened 2015-06-25
Fix to allow LW resources to be used with HW providers

jdmundrawala:
This should fix sethvargo-cookbooks/swap#22

Here's what was happening:
The cookbooks creates a Resource using the LWRP base. This dynamically
creates a class. We used to create this class with the name
Chef::Resource::SomeResourceName. In 12.4, this change to something like
"LWRP resource some_resource_name from cookbook CookbookName". When
searching for a provider, it couldn't be found because it wasn't explicitly
set, as it would have been if it was a LW provider, and it wasn't found
in Chef::Provider, because Chef::Provider had a provider SomeResourceName
instead of "LWRP resource some_resource_name from cookbook CookbookName".

cc @jkeiser @sethvargo @chef/client-maintainers

### DIFF
--- a/lib/chef/platform/provider_mapping.rb
+++ b/lib/chef/platform/provider_mapping.rb
@@ -197,7 +197,8 @@ class Chef
 
         def resource_matching_provider(platform, version, resource_type)
           if resource_type.kind_of?(Chef::Resource)
-            class_name = resource_type.class.to_s.split('::').last
+            class_name = resource_type.class.name ? resource_type.class.name.split('::').last : 
+              convert_to_class_name(resource_type.resource_name.to_s)
 
             begin
               result = Chef::Provider.const_get(class_name)
--- a/spec/integration/recipes/recipe_dsl_spec.rb
+++ b/spec/integration/recipes/recipe_dsl_spec.rb
@@ -969,4 +969,29 @@ describe "Recipe DSL methods" do
       end
     end
   end
+
+  context "with a dynamically defined resource and regular provider" do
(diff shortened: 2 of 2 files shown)


lib/chef/platform/provider_mapping.rb
```diff
@@ -197,7 +197,8 @@ def platform_provider(platform, version, resource_type)
 
         def resource_matching_provider(platform, version, resource_type)
           if resource_type.kind_of?(Chef::Resource)
-            class_name = resource_type.class.to_s.split('::').last
+            class_name = resource_type.class.name ? resource_type.class.to_s.split('::').last :
```
jkeiser: This should really be `resource_type.class.name`: it won't be different for the classes we are trying to backcompat here, and is more consistent.

jkeiser:
Just one comment, and :+1:.  Thanks!!!

lib/chef/platform/provider_mapping.rb
jdmundrawala: For my own learning, what is the difference between class.name and class.to_s?

danielsdeleo:
:shipit:

thommay:
seems reasonable :+1:

jkeiser:
`class.name` is the name of the class, while `to_s` may be overridden by people to change what shows up when you print it (our lwrp classes do this even when they don't have a name).

ksubrama:
:+1:

smurawski:
:+1:

PR #3595, opened 2015-06-25
[ECO-451] Bootstrap node with an environment when set in first-boot.json

mcquin:
Adds the ability to bootstrap a node with an environment when one is set in first-boot.json. Order of environment preference: `-E ENV`, `-j JSON`, config file.

@chef/client-core

### DIFF
--- a/lib/chef/knife/bootstrap/client_builder.rb
+++ b/lib/chef/knife/bootstrap/client_builder.rb
@@ -80,9 +80,12 @@ class Chef
           knife_config[:chef_node_name]
         end
 
-        # @return [String] enviroment from the knife_config
+        # @return [String] environment from the knife_config or first_boot_attributes, if none specified by knife_config
         def environment
-          knife_config[:environment]
+          # prefer environment from cli over environment from first-boot attributes
+          knife_config[:environment] ||
+            (first_boot_attributes[:environment] if first_boot_attributes) ||
+            chef_config[:environment]
         end
 
         # @return [String] run_list from the knife_config
@@ -138,7 +141,13 @@ class Chef
               node = Chef::Node.new(chef_server_rest: client_rest)
               node.name(node_name)
               node.run_list(normalized_run_list)
-              node.normal_attrs = first_boot_attributes if first_boot_attributes
+              if first_boot_attributes
+                scrubbed_first_boot_attributes = first_boot_attributes.dup
+                scrubbed_first_boot_attributes.delete(:environment)
+                unless scrubbed_first_boot_attributes.empty?
+                  node.normal_attrs = scrubbed_first_boot_attributes
+                end
+              end
               node.environment(environment) if environment
               node
             end
--- a/lib/chef/knife/bootstrap/client_builder.rb
+++ b/lib/chef/knife/bootstrap/client_builder.rb
@@ -84,7 +84,7 @@ class Chef
         def environment
           # prefer environment from cli over environment from first-boot attributes
           knife_config[:environment] ||
-            (first_boot_attributes[:environment] if first_boot_attributes) ||
+            (first_boot_attributes[:chef_environment] if first_boot_attributes) ||
             chef_config[:environment]
         end
 
@@ -143,7 +143,7 @@ class Chef
               node.run_list(normalized_run_list)
               if first_boot_attributes
                 scrubbed_first_boot_attributes = first_boot_attributes.dup
-                scrubbed_first_boot_attributes.delete(:environment)
+                scrubbed_first_boot_attributes.delete(:chef_environment)
                 unless scrubbed_first_boot_attributes.empty?
                   node.normal_attrs = scrubbed_first_boot_attributes
                 end
--- a/lib/chef/knife/bootstrap/client_builder.rb
+++ b/lib/chef/knife/bootstrap/client_builder.rb
@@ -84,7 +84,7 @@ class Chef
         def environment
           # prefer environment from cli over environment from first-boot attributes
           knife_config[:environment] ||
-            (first_boot_attributes[:chef_environment] if first_boot_attributes) ||
+            (first_boot_attributes['chef_environment'] if first_boot_attributes) ||
             chef_config[:environment]
         end
 
@@ -143,7 +143,7 @@ class Chef
               node.run_list(normalized_run_list)
               if first_boot_attributes
                 scrubbed_first_boot_attributes = first_boot_attributes.dup
-                scrubbed_first_boot_attributes.delete(:chef_environment)
+                scrubbed_first_boot_attributes.delete('chef_environment')
                 unless scrubbed_first_boot_attributes.empty?
                   node.normal_attrs = scrubbed_first_boot_attributes
                 end
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -40,7 +40,16 @@ class Chef
         end
 
         def bootstrap_environment
-          @chef_config[:environment] || '_default'
+          # Check for environment in CLI, first_boot_attributes, then in the
+          # config file. Use '_default' if no environment specified.
+
+          # @chef_config picks up @config's environment setting, even when
+          # environment is set in the config file being used. However,
+          # @config[:environment] is not polluted by @chef_config[:environment].
+          @config[:environment] ||
+            (@config[:first_boot_attributes][:environment] if @config[:first_boot_attributes]) ||
+            @chef_config[:environment] ||
+            '_default'
         end
 
         def validation_key
@@ -167,7 +176,7 @@ CONFIG
         end
 
         private
-       
+
         # Returns a string for copying the trusted certificates on the workstation to the system being bootstrapped
         # This string should contain both the commands necessary to both create the files, as well as their content
         def trusted_certs_content
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -47,7 +47,7 @@ class Chef
           # environment is set in the config file being used. However,
           # @config[:environment] is not polluted by @chef_config[:environment].
           @config[:environment] ||
-            (@config[:first_boot_attributes][:environment] if @config[:first_boot_attributes]) ||
+            (@config[:first_boot_attributes][:chef_environment] if @config[:first_boot_attributes]) ||
             @chef_config[:environment] ||
             '_default'
         end
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -47,7 +47,7 @@ class Chef
           # environment is set in the config file being used. However,
           # @config[:environment] is not polluted by @chef_config[:environment].
           @config[:environment] ||
-            (@config[:first_boot_attributes][:chef_environment] if @config[:first_boot_attributes]) ||
+            (@config[:first_boot_attributes]['chef_environment'] if @config[:first_boot_attributes]) ||
             @chef_config[:environment] ||
             '_default'
         end
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -314,7 +314,7 @@ class Chef
 
         # chef-vault integration must use the new client-side hawtness, otherwise to use the
         # new client-side hawtness, just delete your validation key.
-        if chef_vault_handler.doing_chef_vault? || 
+        if chef_vault_handler.doing_chef_vault? ||
             (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
 
           unless config[:chef_node_name]
--- a/spec/unit/knife/bootstrap/client_builder_spec.rb
+++ b/spec/unit/knife/bootstrap/client_builder_spec.rb
@@ -168,11 +168,95 @@ describe Chef::Knife::Bootstrap::ClientBuilder do
       client_builder.run
     end
 
-    it "builds a node with an environment if its given" do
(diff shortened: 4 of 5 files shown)


lib/chef/knife/bootstrap/client_builder.rb
```diff
@@ -80,9 +80,12 @@ def node_name
...
         end
 
-        # @return [String] enviroment from the knife_config
+        # @return [String] enviroment from the knife_config or first_boot_attributes, if none specified by knife_config
         def environment
-          knife_config[:environment]
+          # prefer environment from cli over environment from first-boot attributes
+          knife_config[:environment] or
+            (first_boot_attributes and first_boot_attributes[:environment]) or
+            chef_config[:environment]
```
btm: when would `chef_config[:environment]` have a value?

charlesjohnson:
@mcquin I put that requirement in originally, but now I think I was wrong. I think I brainfarted on the "Chef-solo totally supports environments" thing. I've removed the requirement from the original card.

charlesjohnson:
@alexpop

lib/chef/knife/bootstrap/client_builder.rb
```diff
@@ -80,9 +80,12 @@ def node_name
           knife_config[:chef_node_name]
         end
 
-        # @return [String] enviroment from the knife_config
+        # @return [String] enviroment from the knife_config or first_boot_attributes, if none specified by knife_config
```
btm: might as well fix the typo in enviroment while you're here.

lib/chef/knife/bootstrap/client_builder.rb
btm: I guess it'd be helpful to have a comment that's like
```
# knife_config is the config merged config hash from mixlib-cli
# chef_config is the Chef::Config class that just has values loaded from config files
```

provided that's correct. O_o

mcquin:
Additionally, does not set an `environment` value in `/etc/chef/client.rb`:

```
> bundle exec knife environment create production
> bundle exec knife bootstrap localhost -p 2200 -x vagrant -P vagrant -N test00 -j "{\"environment\": \"production\"}" --sudo
> vagrant ssh
$ cat /etc/chef/client.rb
log_location     STDOUT
chef_server_url  "https://api.opscode.com/organizations/****"
validation_client_name "****-validator"
node_name "test00"
```

lib/chef/knife/bootstrap/client_builder.rb
mcquin: It has a value when the config file you use has an environment key:

- Knife doesn't complain if your knife.rb has an environment setting. The environment key will be picked up in the chef_config object.
- You can pass `-c ~/.chef/client.rb` as a CLI option, and it can read an environment key from that file.

mcquin: Yep, [this is correct](https://github.com/chef/chef/pull/3595#discussion_r33299635)

mcquin: @btm Yeah, that's generally correct. It's captured [here](https://github.com/chef/chef/blob/mcquin/first-boot-environment/lib/chef/knife/bootstrap/client_builder.rb#L30-L33).

btm:
:+1:

lib/chef/knife/bootstrap/client_builder.rb
```diff
@@ -80,9 +80,12 @@ def node_name
           knife_config[:chef_node_name]
         end
 
-        # @return [String] enviroment from the knife_config
+        # @return [String] environment from the knife_config or first_boot_attributes, if none specified by knife_config
         def environment
-          knife_config[:environment]
+          # prefer environment from cli over environment from first-boot attributes
+          knife_config[:environment] or
```
jaym: why use `and` and `or` instead of `&&` and `||` in these cases?

mcquin: `and` and `or` are control flow operators, whereas `&&` and `||` are logical (boolean) operators. See: http://devblog.avdi.org/2010/08/02/using-and-and-or-in-ruby/

mcquin: In this particular instance it doesn't really matter, I can switch back to && and || if preferred.

lib/chef/knife/bootstrap/client_builder.rb
```diff
@@ -138,7 +141,7 @@ def node
               node = Chef::Node.new(chef_server_rest: client_rest)
               node.name(node_name)
               node.run_list(normalized_run_list)
-              node.normal_attrs = first_boot_attributes if first_boot_attributes
+              node.normal_attrs = first_boot_attributes if first_boot_attributes # TODO: should environment be saved here, if present?
```
tyler-ball: I think the answer to this TODO is `no` because the node will be assigned an environment.  And nothing about these normal attributes communicate that they are first_boot_attributes unless you know the implementation of this.

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -40,7 +40,16 @@ def initialize(config, run_list, chef_config, secret = nil)
         end
 
         def bootstrap_environment
-          @chef_config[:environment] || '_default'
+          # Check for environment in CLI, first_boot_attributes, then in the
+          # config file. Use '_default' if no environment specified.
+
+          # TODO: @chef_config picks up @config's environment setting, even when
```
tyler-ball: Is there anything to do for this `TODO` or is this just a future note of caution?

tyler-ball:
:+1:

lib/chef/knife/core/bootstrap_context.rb
mcquin: I left it as a reminder for myself to ask a question, then I figured out my answer. I'll leave this as a comment (but remove the TODO).

mcquin:
Most notable update to this PR is [here](https://github.com/chef/chef/blob/mcquin/first-boot-environment/lib/chef/knife/bootstrap/client_builder.rb#L144-L149). Does this seem sane @chef/client-core ?

tyler-ball:
LGTM!  Does anyone else have opinions on how this should work?

mcquin:
Ping @chef/client-core

thommay:
LGTM - one definitely does not want one's environment to wind up in the node's client.rb because that way lies sadness and confusion. :+1:

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -40,7 +40,16 @@ def initialize(config, run_list, chef_config, secret = nil)
...
         def bootstrap_environment
-          @chef_config[:environment] || '_default'
+          # Check for environment in CLI, first_boot_attributes, then in the
+          # config file. Use '_default' if no environment specified.
+
+          # @chef_config picks up @config's environment setting, even when
+          # environment is set in the config file being used. However,
+          # @config[:environment] is not polluted by @chef_config[:environment].
+          @config[:environment] ||
+            (@config[:first_boot_attributes][:environment] if @config[:first_boot_attributes]) ||
```
coderanger: Is there a reason to use the key `environment` over `chef_environment`?

charlesjohnson: Good catch. Agreed we should be consistent on chef_environment.

mcquin: Configuration files use the key `environment`. I'm not familiar with `chef_environment` and its use.

coderanger: The node method is `Node#chef_environment` as is the search index key (`search(:node, 'chef_environment:foo')`).

mcquin: Knife was bootstrapping with `environment`, not `chef_environment`: https://github.com/chef/chef/blob/mcquin/first-boot-environment/lib/chef/knife/bootstrap/client_builder.rb#L151

coderanger: That's an old alias: https://github.com/chef/chef/blob/master/lib/chef/node.rb#L133

mcquin: :+1: Updating...

coderanger:
It is worth noting this may introduce confusion as a lot of people are already fuzzy on `node.chef_environment` vs. `node['chef_environment']`. Normally the latter is nil so at least errors are caught early, with this in place I think it would mean that the latter would always return the env the node was bootstrapped with, while future changes might make it different from `node.chef_environment`. Still :+1: on the feature, but we should bump up the priority of resolving this other issue. Probably would be good to always set `node['chef_environment']` to the current value at the start of a run. That doesn't resolve the issue of setting it inside recipe code though, or how to handle something like a role that overrides the attribute (which would have no effect on `node.chef_environment`).

mcquin:
Correct, `node['chef_environment']` and `node.chef_environment` at the start of a run are now the environment the node was bootstrapped with. 

@coderanger 
> we should bump up the priority of resolving this other issue

What exactly is "this other issue"? That `node.chef_environment` and `node['chef_environment']` may disagree at the start of a run?

coderanger:
I think you are scrubbing the actual attribute values, so more of a user expectation issue. They might think they could use `node['chef_environment']` because it was set in an attributes JSON file, but it isn't an attribute value.

mcquin:
Taking a step back to what's confusing me...

What exactly is the purpose of first boot attributes? Are they meant to be used /only/ during bootstrap/the first time chef-client runs, or do they persist (see: https://github.com/chef/chef/pull/3595#discussion_r33633076)? Should the environment a node is bootstrapped with only be applied to the node during bootstrap (i.e., running `chef-client` again on the node will use `_default` if no other environment is specified)?

coderanger:
Attribute passed in via `-j` are persistent, they are set as the value of `node.normal` and will last indefinitely. Separately from that we have the issue that `node.chef_environment` is _not_ attribute data, normally `node['chef_environment']` will be unset (`nil`) because `Chef::Node#chef_environment` is a real attr_accessor, not a use of the old method_missing fallback for attributes. I'm unclear we are trying to add a way to set the environment via attribute data when the environment of a node is not kept as an attribute, and I'm worried it will increase the confusion around this issue.

mcquin:
> I'm unclear we are trying to add a way to set the environment via attribute data when the environment of a node is not kept as an attribute, and I'm worried it will increase the confusion around this issue.

Maybe @charlesjohnson and @alexpop can help answer this question.

charlesjohnson:
The underlying pain, along with @alexpop's original proposal for this fix, is outlined here: https://feedback.chef.io/forums/301644-chef-product-feedback/suggestions/8278449-support-specifying-node-environment-in-first-boot:

> I've seen users having difficulties figuring out how to specify the chef environment while implementing an unattended chef-client bootstrap: 
> https://docs.chef.io/install_bootstrap.html#unattended-installs

> Right now, the `--environment` option needs to be used for chef-client to place the node in a specific environment.

> First instinct was to add the chef_environment to 'first-boot.json' along with the run_list and normal attributes. From an automation perspective, it makes more sense to generate a client.rb, a json file and run chef-client. So, this enhancement request is to allow the chef_environment to be specified in the json attributes and just call this on the first chef-client run: 
> `chef-client -j /etc/chef/first-boot.json`

> On failure, it's a simple command to run without having to remember how to specify the environment.

So if we need to re-state, then how about: "As a Chef User I want to bootstrap nodes from a single configuration file without additional command-line flags, so that unattended installs no longer require humans to track additional metadata when interacting with the unattended install process. This single configuration file should contain all of the data currently stored in first-boot.json, as well as the node environment if specified."

coderanger:
Hmm, so this won't allow setting a run list, which would probably be a hard requirement for that story. Perhaps we should rewind a whole bunch and make a new option which accepts an actual node JSON file? That would cover attributes, environment, run list, and anything else that gets added to nodes later (eg. the attributes 2.0 data split). That still wouldn't handle the forthcoming Ohai config data or `chef_server_url`/`encrypted_data_bag_secret` which are read from knife's config and are never part of a node object.

If we combine all that, I'm not sure how great this looks as a feature anymore, would end up looking a lot like spiceweasel maybe? Would be :+1: on have a `--json-node` to cover at least all data stored in the node object itself.

coderanger:
Also "unattended installs" usually doesn't mean `knife bootstrap`, which is all this covers (and the cloud plugins that use `knife bootstrap` under the hood). If you mean "unattended" in terms of things like autoscaling or self-healing systems, this is not going to help there. Can you clarify that piece of the use case a bit?

charlesjohnson:
That's way too big. YAGNI.

Again: Instead of `chef-client -j first-boot.json -E chef_environment`, we want: `chef-client --some-flag unattended-config-file-that-has-everything-you-would-put-in-first-boot-json-and-also-the-node-environment`.

If piggybacking the unattended install data into first-boot.json isn't the right solution then okay, but that's the requirement.

coderanger:
@charlesjohnson Okay, then we can close out, this change is about adding this feature to `knife bootstrap`, not to `chef-client`. I think there was just a miscommunication on the goals here. `-j` on the client has a different meaning vs. knife :-)

charlesjohnson:
Nope, this isn't about knife bootstrap. This is about `chef-client --some-flag unattended-config-file-that-has-everything-you-would-put-in-first-boot-json-and-also-the-node-environment.`

mcquin:
@charlesjohnson I think @coderanger means this PR is about bootstrap.

coderanger:
Yes, I mean that this PR is not what you want.

mcquin:
I will be tackling this in another PR that addresses the core issue.

coderanger:
:+1: @mcquin https://github.com/chef/chef/blob/master/lib/chef/node.rb#L316-L322 is probably the right place to add this.

mcquin:
@coderanger Looks likely. Thanks! :)

Issue #3596, opened 2015-06-25
powershell_script no output for info and debug log levels


alexpop:
On a Linux node, if I have these two resources:

```ruby
execute "multi_line_script" do
  command <<-EOH
    whoami
    ls /etc
  EOH
  action :run
end

bash 'test1' do
  code '
    ls /tmp
  '
  action :run
end
```

they will print the STDOUT of the two bash scripts when chef-client uses 'info' or 'debug' log levels.

This is not the case for the 'powershell_script' resource. The following resource will have no output when chef-client uses 'info' or 'debug' log levels:

```ruby
powershell_script 'test' do
  code '
    ls c:\
  '
  action :run
end
```

adamedx:
For triage: I think we should do this at medium priority and consider @alexpop's fix.

dougireton:
I am struggling with this right now. PowerShell script output would really help with debugging.

smurawski:
Closed with https://github.com/chef/chef/pull/4087

PR #3597, opened 2015-06-25
print STDOUT from the powershell wrapper. fixes issue #3596

alexpop:
fixes issue #3596

### DIFF
--- a/lib/chef/provider/powershell_script.rb
+++ b/lib/chef/provider/powershell_script.rb
@@ -163,6 +163,9 @@ elseif ( $LASTEXITCODE -ne $null -and $LASTEXITCODE -ne 0 )
   $exitstatus = $LASTEXITCODE
 }
 
+# Print STDOUT for the script execution
+Write-Output $chefscriptresult
+
 # If this script is launched with -File, the process exit
 # status of PowerShell.exe will be $exitstatus. If it was
 # launched with -Command, it will be 0 if $exitstatus was 0,


martinisoft:
This appears to be ok, but I'm not a powershell user (only confirmed usage via the docs). Anyone from @chef/maintainers able to lend a hand for review?

danielsdeleo:
Same here, it looks fine but I'm not a subject matter expert

jtimberman:
Paging @smurawski for :eyes:

alexpop:
I've hacking with that powershell wrapper script. I don't like that the trap is making the use of $ErrorActionPreference impossible. @adamedx are you working on rewriting the powershell wrapper used by the powershell_script resource?

adamedx:
I have some concerns about using `Write-Output` -- does that write to the pipeline? If so, that would affect the output of the script. Seems like it's passing on Windows so maybe that's ok, but I would think something that generated output without affecting the pipeline would be preferable.

adamedx:
@alexpop my own involvement here is pretty light, but the maintainers may be able to help if you are running into some issues.

adamedx:
Yeah, I think `Write-Output` is the wrong choice, it should be `Write-Host` instead. Also, we're missing some test coverage since some test should break if our wrapper alters the script output:

```ruby
% adamed@SEA-ADAMED1: 0x0100E0D0: C:\Users\edwardsa
2015.06.26 Fri 14:38:55.00-07:00> { }.invokereturnasis() -eq $null
True
% adamed@SEA-ADAMED1: 0x0100E0D0: C:\Users\edwardsa
2015.06.26 Fri 14:39:10.45-07:00> { write-output 'hi' }.invokereturnasis() -eq $null
False
% adamed@SEA-ADAMED1: 0x0100E0D0: C:\Users\edwardsa
2015.06.26 Fri 14:39:17.40-07:00> { write-host 'hi' }.invokereturnasis() -eq $null
hi
True
% adamed@SEA-ADAMED1: 0x0100E0D0: C:\Users\edwardsa
2015.06.26 Fri 14:39:22.63-07:00>
```

alexpop:
Thanks @adamedx. If I use `Write-Host`, I don't get the stdout of the powershell script from kitchen or from a direct chef-client run.

With `Write-Output`, I get the stdout when I run chef-client on the box using the same command as kitchen:

```
$env:TEMP='C:\Users\vagrant\appdata\local\Temp'
chef-client.bat --local-mode --config $env:TEMP\kitchen\client.rb --log_level info --force-formatter --no-color --json-attributes $env:TEMP\kitchen\dna.json --chef-zero-port 8889
```

With `Write-Output`, I don't get the stdout when I run `kitchen converge` from my laptop with info or debug log levels. This might be related to: https://github.com/test-kitchen/test-kitchen/issues/685

ksubrama:
I gave this a quick test and it makes sense to me.  @adamedx - the write-output is after the invokereturnasis and is printing what powershell would otherwise have sent to stdout.  It was error case that I was worried about - I didn't know if printing the output there would cause any weirdness with the output happening after the script failed but I don't see anything particularly offensive about the current behavior.

ksubrama:
:+1:

lamont-granquist:
:+1:

smurawski:
Merged with https://github.com/chef/chef/pull/4087

PR #3600, opened 2015-06-26
Add myself as a core maintainer.

coderanger:
I would really rather we split the DSL and Resource machinery in to a component because I'm definitely not able to call myself a maintainer for the "and everything else" part, but waiting to get my patches merged is getting annoying.

### DIFF
--- a/MAINTAINERS.toml
+++ b/MAINTAINERS.toml
@@ -34,6 +34,7 @@ another component.
 
       maintainers = [
         "btm",
+        "coderanger",
         "danielsdeleo",
         "fujin",
         "jaymzh",
@@ -290,3 +291,8 @@ The specific components of Chef related to a given platform - including (but not
   [people.joemiller]
     Name = "Joe Miller"
     GitHub = "joemiller"
+
+  [people.coderanger]
+    Name = "Noah Kantrowitz"
+    GitHub = "coderanger"
+


jaym:
:cake:

ranjib:
@coderanger which patches of yours are in queue ? i can take a look during the weekend

mcquin:
:+1:

@coderanger It'd be really cool if we could have a DSL-focused maintainer subteam, I'd totes join it with you. =)

coderanger:
Ping @adamhjk, what say ye about making "DSL plumbing" a component?

adamhjk:
Lets talk about it at the next meeting. I'm all for more people having commit bits - I could also see this just being part of "core". (Mostly I'm thinking about how we talk about routing PRs and such to this proposed new team)

mcquin:
@adamhjk I agree that routing PRs could be difficult.

@thommay Is there a way to ping/assign people based on which files are modified in the commit? For example, the DSL maintainers would be pinged/assigned if there were any changes to `lib/chef/dsl.rb`, `lib/chef/dsl/*`, `lib/chef/provider*.rb`, `lib/chef/provider/*`, `lib/chef/resource*.rb`, `lib/chef/resource/*`?

ranjib:
similar to chef-config we can split chef-core . that will ease this as well as lot other things ... :-) .

danielsdeleo:
We don't need a majority for any merges except when someone has protested/attempted a veto, so it's fine if the scope is a bit bigger than you feel comfortable with, you can just abstain if you don't feel qualified to review a particular patch. Anyway, :+1: to the question at hand.

lamont-granquist:
We can logically split things up without actually splitting up the code.  After all that is how Windows, Ubuntu, etc is done.

The only real burdensome issue with being a core chef maintainer is getting the mailing list pings, but that can be dealt with e-mail filters, threaded mail readers and the delete key...

:+1:  anyway

thommay:
@mcquin so the maintainers.toml has a concept of paths: https://github.com/chef/chef/blob/master/MAINTAINERS.toml#L53-L59
But they need filling out for the teams. Supermarket doesn't do anything with them (yet!).

I'm pretty neutral on adding a DSL maint team - I think as Dan says if you're only comfortable reviewing part of a PR, then just do that and make sure that your comments make it obvious that that's what's happened.

thommay:
:+1: for this though. Merging!

PR #3603, opened 2015-06-26
Fix ability to monkey match LWRP through Chef::Resource::MyLwrp

jaym:
Repro
=====
The following cookbook would repro the issue

cookbooks/lwrptest/resources/my_lwrp.rb
---------------------------------------
```ruby
```

cookbooks/lwrptest/providers/my_lwrp.rb
---------------------------------------
```ruby
```

cookbooks/lwrptest/recipes/default.rb
-------------------------------------
```ruby
module Something
  def something
    puts 'something here'
  end
end

::Chef::Resource::LwrptestMyLwrp.send(:include, Something)

lwrptest_my_lwrp 'blah' do
  something
end
```

Why it does not work
====================
Running this in 12.4.0 produces the error
```
NoMethodError
-------------
undefined method `something' for LWRP resource lwrptest_my_lwrp from cookbook lwrptest
```

The problem is that LWRP usage through `Chef::Resource::MyLwrp` and
`Chef::Provider::MyLwrp` is depreacted. In order to participate in
the deprecation, `Chef::Resource::MyLwrp` is a subclass of the actual
`MyLwrp` class with `initialize` overriden to do a Chef::Log.deprecation.

In the example above, the intention is to add a method to the resource
so that it can be used from the dsl. The reason this does not work is
because including the method on `Chef::Resource::MyLwrp` adds the method
to the subclass, and the resource that is looked up when running the
recipe is the actual resource.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1499,7 +1499,7 @@ class Chef
         Chef::Resource.send(:remove_const, class_name)
       end
 
-      eval("Chef::Resource::#{class_name} = resource_class")
+      Chef::Resource.const_set(class_name, resource_class)
 
       deprecated_constants[class_name.to_sym] = resource_class
     end
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1499,56 +1499,9 @@ class Chef
         Chef::Resource.send(:remove_const, class_name)
       end
 
-      # In order to generate deprecation warnings when you use Chef::Resource::MyLwrp,
-      # we make a special subclass (identical in nearly all respects) of the
-      # actual LWRP.  When you say any of these, a deprecation warning will be
-      # generated:
-      #
-      # - Chef::Resource::MyLwrp.new(...)
-      # - resource.is_a?(Chef::Resource::MyLwrp)
-      # - resource.kind_of?(Chef::Resource::MyLwrp)
-      # - case resource
-      #   when Chef::Resource::MyLwrp
-      #   end
-      #
-      resource_subclass = Class.new(resource_class) do
-        resource_name nil # we do not actually provide anything
-        def initialize(*args, &block)
-          Chef::Log.deprecation("Using an LWRP by its name (#{self.class.name}) directly is no longer supported in Chef 13 and will be removed.  Use Chef::Resource.resource_for_node(node, name) instead.")
-          super
-        end
-        def self.resource_name(*args)
-          if args.empty?
-            @resource_name ||= superclass.resource_name
-          else
-            super
-          end
-        end
-        self
-      end
-      eval("Chef::Resource::#{class_name} = resource_subclass")
-      # Make case, is_a and kind_of work with the new subclass, for backcompat.
-      # Any subclass of Chef::Resource::ResourceClass is already a subclass of resource_class
-      # Any subclass of resource_class is considered a subclass of Chef::Resource::ResourceClass
-      resource_class.class_eval do
-        define_method(:is_a?) do |other|
-          other.is_a?(Module) && other === self
-        end
-        define_method(:kind_of?) do |other|
-          other.is_a?(Module) && other === self
-        end
-      end
-      resource_subclass.class_eval do
-        define_singleton_method(:===) do |other|
-          Chef::Log.deprecation("Using an LWRP by its name (#{class_name}) directly is no longer supported in Chef 13 and will be removed.  Use Chef::Resource.resource_for_node(node, name) instead.")
-          # resource_subclass is a superclass of all resource_class descendants.
-          if self == resource_subclass && other.class <= resource_class
-            return true
-          end
-          super(other)
-        end
-      end
-      deprecated_constants[class_name.to_sym] = resource_subclass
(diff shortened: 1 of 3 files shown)


chef-supermarket:

* @jaym

lib/chef/resource/lwrp_base.rb
```diff
@@ -52,17 +52,37 @@ def build_from_file(cookbook_name, filename, run_context)
...
-              "LWRP resource #{resource_name} from cookbook #{cookbook_name}"
+          deprecated_resource_class = Class.new(self).tap do |resource_class|
+            resource_class.resource_name(nil)
+            resource_class.run_context = run_context
+            resource_class.instance_eval do
+              define_method(:initialize) do |*args, &block|
+                Chef::Log::deprecation("Deprecated Thing") if chef_depreacted_access
+                super(*args, &block)
+              end
+              define_method(:chef_depreacted_access) do
```
danielsdeleo: Speling :wink:

lamont-granquist:
paging @jkeiser

danielsdeleo:
So do we have a corpus of LWRP cookbooks to test with now? Is it feasible to add them to travis tests or pull them all in as a regression test suite?

lib/chef/resource/lwrp_base.rb
```diff
@@ -52,17 +52,37 @@ def build_from_file(cookbook_name, filename, run_context)
...
+                "LWRP resource #{resource_name} from cookbook #{cookbook_name}"
+              end
+              define_singleton_method(:inspect) { to_s }
+            end
+          end
+
+          resource_class = Class.new(deprecated_resource_class).tap do |resource_class|
+            resource_class.resource_name(resource_name.to_sym)
+            resource_class.instance_eval do
+              define_method(:chef_depreacted_access) do
```
jkeiser: Typo ("chef_deprecated_access")

lib/chef/resource/lwrp_base.rb
```diff
@@ -52,17 +52,37 @@ def build_from_file(cookbook_name, filename, run_context)
...
-          # Make a useful string for the class (rather than <Class:312894723894>)
-          resource_class.instance_eval do
-            define_singleton_method(:to_s) do
-              "LWRP resource #{resource_name} from cookbook #{cookbook_name}"
+          deprecated_resource_class = Class.new(self).tap do |resource_class|
+            resource_class.resource_name(nil)
+            resource_class.run_context = run_context
+            resource_class.instance_eval do
+              define_method(:initialize) do |*args, &block|
+                Chef::Log::deprecation("Deprecated Thing") if chef_depreacted_access
```
jkeiser: What if we did "if self.class == resource_class" instead? One less method to pollute the LWRP with ...

btm:
@jaym you should sign a CLA. heh.

jkeiser:
:+1: in general, one comment that we could get rid of the extra chef_deprecated_class (I think).  And, need a test (and to fix travis :)

lamont-granquist:
@jaym you are going to get confused with @jaymzh now...  =)

jaym:
:)

chef-supermarket:

lib/chef/resource.rb
```diff
@@ -1506,56 +1506,9 @@ def self.register_deprecated_lwrp_class(resource_class, class_name)
...
-          Chef::Log.deprecation("Using an LWRP by its name (#{class_name}) directly is no longer supported in Chef 13 and will be removed.  Use Chef::Resource.resource_for_node(node, name) instead.")
-          # resource_subclass is a superclass of all resource_class descendants.
-          if self == resource_subclass && other.class <= resource_class
-            return true
-          end
-          super(other)
-        end
-      end
-      deprecated_constants[class_name.to_sym] = resource_subclass
+      eval("Chef::Resource::#{class_name} = resource_class")
```
coderanger: There should be a comment here to explain why this isn't using `const_set`.

jaym: Not sure why.
cc @jkeiser @lamont-granquist @danielsdeleo

spec/unit/lwrp_spec.rb
```diff
@@ -177,6 +177,16 @@ def get_lwrp_provider(name)
       end
     end
 
+    it "should allow monkey patching of the through Chef::Resource" do
```
thommay: "the LWRP"?

thommay:
@jaym: you're OK to merge with the provisos that you add the comment that @coderanger requested, and fix the name of the test.

lib/chef/resource.rb
jkeiser: The reason it's not using const_set is because Ruby. The name of the class doesn't get set with const_set for some reason. A lot of banking things are like that in Ruby, sadly.

coderanger: I don't think this is correct:

```
>> Bar.const_set(:Baz, Class.new)
=> Bar::Baz
>> Bar::Baz.name
=> "Bar::Baz"
```

jkeiser: It's inconsistent at best. I haven't figured out the rules for when it works and when it doesn't.

PR #3605, opened 2015-06-26
Rework Resource#action to match the 12.3 API.

### LINKED ISSUE #3604: 12.4 API breakage: Resource#action no longer casts to Array
In 12.3, setting `action(:foo)` on a resource [would upgrade it to `[:foo]` internally](https://github.com/chef/chef/blob/12.3.0/lib/chef/resource.rb#L165). This [was removed in 12.4](https://github.com/chef/chef/blob/12.4.0/lib/chef/resource.rb#L177-L181). This means code that uses the output value of `Resource#action` expecting an array is now broken. For example, [the runit cookbook](https://github.com/hw-cookbooks/runit/blob/v1.6.0/libraries/provider_runit_service.rb#L112).

In 12.3 the docstring comment does specifically state `@return [Array[Symbol]] the list of actions.` so I think we should consider this bound by API compat rules. Notably the same comment is also present on the 12.4 code despite now being incorrect.

This was changed in https://github.com/chef/chef/commit/236570dc9605ac312b44508b1c3a6a0e7c19b4dc.

Ping @jkeiser @thommay @jaym.

coderanger:
This means it always coerces to an Array. Also ensures that
Resource#action= goes through the same validation.

Fixes #3604.

### DIFF
--- a/lib/chef/guard_interpreter/resource_guard_interpreter.rb
+++ b/lib/chef/guard_interpreter/resource_guard_interpreter.rb
@@ -68,7 +68,8 @@ class Chef
         run_action = action || @resource.action
 
         begin
-          @resource.run_action(run_action)
+          # Coerce to an array to be safe.
+          Array(run_action).each {|action_to_run| @resource.run_action(action_to_run) }
           resource_updated = @resource.updated
         rescue Mixlib::ShellOut::ShellCommandFailed
           resource_updated = nil
--- a/spec/functional/provider/whyrun_safe_ruby_block_spec.rb
+++ b/spec/functional/provider/whyrun_safe_ruby_block_spec.rb
@@ -43,7 +43,7 @@ describe Chef::Resource::WhyrunSafeRubyBlock do
     end
 
     it "updates the evil laugh, even in why-run mode" do
-      new_resource.run_action(new_resource.action)
+      Array(new_resource.action).each {|action| new_resource.run_action(action) }
       expect($evil_global_evil_laugh).to eq(:mwahahaha)
       expect(new_resource).to be_updated
     end
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1152,7 +1152,7 @@ class Chef
       action = action.to_sym
       new_action_provider_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
-      default_action action if default_action == [:nothing]
+      default_action action if Array(default_action) == [:nothing]
     end
 
     #
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1116,7 +1116,7 @@ class Chef
       elsif superclass.respond_to?(:default_action)
         superclass.default_action
       else
-        :nothing
+        [:nothing]
       end
     end
     def self.default_action=(action_name)
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1093,22 +1093,17 @@ class Chef
     # Setting default_action will automatially add the action to
     # allowed_actions, if it isn't already there.
     #
-    # Defaults to :nothing.
+    # Defaults to [:nothing].
     #
     # @param action_name [Symbol,Array<Symbol>] The default action (or series
     #   of actions) to use.
     #
-    # @return [Symbol,Array<Symbol>] The default actions for the resource.
+    # @return [Array<Symbol>] The default actions for the resource.
     #
     def self.default_action(action_name=NOT_PASSED)
       unless action_name.equal?(NOT_PASSED)
-        if action_name.is_a?(Array)
-          @default_action = action_name.map { |arg| arg.to_sym }
-        else
-          @default_action = action_name.to_sym
-        end
-
-        self.allowed_actions |= Array(@default_action)
+        @default_action = Array(action_name).map(&:to_sym)
+        self.allowed_actions |= @default_action
       end
 
       if @default_action
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1157,7 +1157,7 @@ class Chef
       action = action.to_sym
       new_action_provider_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
-      default_action action if default_action == :nothing
+      default_action action if default_action == [:nothing]
     end
 
     #
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -169,25 +169,23 @@ class Chef
     # @param arg [Array[Symbol], Symbol] A list of actions (e.g. `:create`)
     # @return [Array[Symbol]] the list of actions.
     #
-    attr_accessor :action
     def action(arg=nil)
       if arg
-        if arg.is_a?(Array)
-          arg = arg.map { |a| a.to_sym }
-        else
-          arg = arg.to_sym
-        end
-        Array(arg).each do |action|
+        arg = Array(arg).map(&:to_sym)
+        arg.each do |action|
           validate(
             { action: action },
             { action: { kind_of: Symbol, equal_to: allowed_actions } }
           )
         end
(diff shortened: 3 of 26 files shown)


jaym:
ugh, something is wrong with appveyor. I hit that issue as well

ksubrama:
:+1:

fnichol:
Agreed

jaym:
i think ffi 1.9.9 just broke us

jaym:
https://github.com/chef/chef/pull/3606 until we can figure out why

jkeiser:
The test failures in Travis look relevant. Some could be tests I changed that need changing back? But the first one looks alarming.

coderanger:
Should be better now, also fixed/testing the default case for `[:nothing]`.

coderanger:
Though hmm, I think the default in 12.3 would actually be a plain Symbol (`:nothing`) not an array. How do people feel about that?

coderanger:
Lulz, this is what I get for only running a subset of tests before commit O:-)

tyler-ball:
:+1: I think this looks right!

coderanger:
Some of the integration test failures look like legit errors but hard to make heads or tails of this test code. @jkeiser?

jkeiser:
I am nowhere near a computer at the moment, but all those failures look like they stem from the first one ("the default action is the first declared action"). By declared action including pretty sure it's talking about the new action syntax. So the first `action :x do ... end` will become the default action unless you say default_action explicitly.

jaym:
@coderanger 
```diff
diff --git i/lib/chef/resource.rb w/lib/chef/resource.rb
index 0679406..0240554 100644
--- i/lib/chef/resource.rb
+++ w/lib/chef/resource.rb
@@ -1157,7 +1157,7 @@ class Chef
       action = action.to_sym
       new_action_provider_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
-      default_action action if default_action == :nothing
+      default_action action if default_action == [:nothing]
     end

     #
```

jaym:
```diff
diff --git i/spec/integration/recipes/resource_action_spec.rb w/spec/integration/recipes/resource_action_spec.rb
index 4786294..cee7913 100644
--- i/spec/integration/recipes/resource_action_spec.rb
+++ w/spec/integration/recipes/resource_action_spec.rb
@@ -337,7 +337,7 @@ describe "Resource.action" do
           NoActionJackson.action_was = action
         end
       }
-      expect(NoActionJackson.action_was).to eq :nothing
+      expect(NoActionJackson.action_was).to eq [:nothing]
     end
   end
 end
```

jaym:
https://github.com/chef/chef/pull/3610

coderanger:
@jaym Mentioned on IRC but for redundancy, we need to make sure to catch the case of:

```ruby
class MyApp < Chef::Resource
  default_action :one
end

MyApp.new('test', nil).action == [:one]
``

**Comment by jkeiser** [comment]
Still a number of failures, FYI (you probably already noticed).  Some calling to_sym on nil values (need a .compact or Array() in there?), some expecting :create and getting [:create].

**Comment by coderanger** [comment]
Yeah, sorting out some weirdness in the why run code.

**Comment by coderanger** [comment]
Should be good to go now.

File: lib/chef/guard_interpreter/resource_guard_interpreter.rb [hunk 1]
```diff
@@ -68,7 +68,8 @@ def evaluate_action(action=nil, &block)
         run_action = action || @resource.action
 
         begin
-          @resource.run_action(run_action)
+          # Coerce to an array to be safe.
```
tyler-ball: Safe from what?  I think this comment is warranted, I just want to know `why` more fully

jaym:
@coderanger @jkeiser 
@tyler-ball has pointed out that this changes the behavior of `resource.action` to return an array. I'm not sure this is the safest thing to do.

spec/functional/provider/whyrun_safe_ruby_block_spec.rb
```diff
@@ -43,7 +43,7 @@
     end
 
     it "updates the evil laugh, even in why-run mode" do
-      new_resource.run_action(new_resource.action)
+      Array(new_resource.action).each {|action| new_resource.run_action(action) }
```
tyler-ball: Shouldn't `new_resource.action` already be returning an array here?  When would it not?

coderanger:
@jaym That is the whole point of this change. In 12.3 and earlier, `resource.action` always returned an array. This was changed in 12.4 which represents an API break.

lib/chef/guard_interpreter/resource_guard_interpreter.rb
coderanger: Resources that override `action` or `default_action`, or that set `@action` manually in an old-school `initialize`.

spec/functional/provider/whyrun_safe_ruby_block_spec.rb
coderanger: As before, there are a lot of ways this can still be a non-array when interacting with legacy code, forcing the up-cast is safest.

coderanger:
I should say, in 12.3 and earlier, setting an action and then looking at the value afterwards always returned an array. With this it should be much closer to "always". To get a non-array now you would either have to touch the ivar manually, or override something in a subclass.

jaym:
12.3.0
---------
```
jmundrawala@seamundrawala01 ~/workspace/chef (●)$ bundle exec chef-apply repro.rb                                                                                                                       [2.1.2]
gem install awesome_print  # <-- highly recommended

From: /Users/jmundrawala/workspace/chef/repro.rb @ line 2 self.run_chef_recipe:

    1: require 'pry'
 => 2: binding.pry

2.1.2 (#<Chef::Recipe:0x007f87a5dc5418>):0 > Chef::Resource::File.new("a", nil).action
=> "create"
```

12.4.0
---------
```
jmundrawala@seamundrawala01 ~/workspace/chef (●)$ bundle exec chef-apply repro.rb                                                                                                                       [2.1.2]
gem install awesome_print  # <-- highly recommended

From: /Users/jmundrawala/workspace/chef/repro.rb @ line 2 self.run_chef_recipe:

    1: require 'pry'
 => 2: binding.pry

2.1.2 (#<Chef::Recipe:0x007f89b9a1e5d8>):0 > Chef::Resource::File.new("a", nil).action
=> :create
```

This branch
-----------------
```
jmundrawala@seamundrawala01 ~/workspace/chef (coderanger-action●)$ bundle exec chef-apply repro.rb                                                                                                      [2.1.2]
gem install awesome_print  # <-- highly recommended

From: /Users/jmundrawala/workspace/chef/repro.rb @ line 2 self.run_chef_recipe:

    1: require 'pry'
 => 2: binding.pry

2.1.2 (#<Chef::Recipe:0x007f939f16fa50>):0 > Chef::Resource::File.new("a", nil).action
=> [:create]
```

coderanger:
Yeah, that's dealing with the default action case. The 12.3 behavior there is being classified as a bug by vote of @thommay as LT. The more important case this fixes is:

```
r = Chef::Resource::File.new('test', nil)
r.action(:touch)
r.action == ???
```

In 12.3 that would return `[:touch]` while in 12.4 it is `:touch`. This puts it back to the 12.3 behavior.

jaym:
Acceptable answer. I will let @thommay have the final call. Sucks we are doing this in a patch release, but it really does not matter between 12.4.1 and 12.5, so :+1:

thommay:
Merged. @jaym please ensure this hits 12.4.1

lib/chef/guard_interpreter/resource_guard_interpreter.rb
tyler-ball: Will you update the comment to include exactly that?  That is quite clear

tyler-ball:
:+1: Yay! :tada:

PR #3606, opened 2015-06-26
FFI 1.9.9 is causing segfaults

jaym:


coderanger:
:+1:

tyler-ball:
:+1:

ksubrama:
I have no good reason to but I suspect https://github.com/ffi/ffi/pull/439 because it just switched out the ruby compiler toolchain on us between 1.9.8 and 1.9.9.  We might have transient issues with cached binaries or maybe it's an actual bug...

In any case..  :+1:

PR #3608, opened 2015-06-28
Rename powershell spec files to match Ruby conventions

adamedx:
Several files under `spec` used to test the `powershell_script` resource and provider did not match the name of the classes being tested -- they were named `powershell_spec.rb` instead of `powershell_script_spec.rb`. The latter is correct according to convention because the source files were named `powershell_script.rb`.

This renames the files to the correct name, `powershell_script_spec.rb` to match the convention.


tyler-ball:
:+1:

jaym:
:+1:

adamedx:
Thanks everyone. :smile:.

PR #3610, opened 2015-06-29
Jdm/coderanger action

jaym:


jkeiser:
This looks good (thanks @coderanger, we just had Jay make a PR so we could take his patch and run it through Travis/etc.).

spec/unit/resource_spec.rb
```diff
@@ -977,4 +977,46 @@ def compiled_resource_data(resource, action, err)
...
+      it { is_expected.to eq [:two] }
+    end
+
+    context "with an array action" do
+      before { resource.action([:two, :one]) }
+      it { is_expected.to eq [:two, :one] }
+    end
+
+    context "with an assignment" do
+      before { resource.action = :one }
```
tyler-ball: Does assignment work with an array of actions?  Would be nice to have a test case either way.

tyler-ball:
:+1:

jaym:
let's keep the discussion on #3605

spec/unit/resource_spec.rb
coderanger: Added in the other PR.

PR #3611, opened 2015-06-29
Call provides? when resolving, reduce number of calls to provides?

jkeiser:
`provides?` is not called on resources or providers unless `provides :x` is also called - it should, for backcompat purposes.

### DIFF
--- a/lib/chef/provider/group/windows.rb
+++ b/lib/chef/provider/group/windows.rb
@@ -28,11 +28,9 @@ class Chef
 
         provides :group, os: 'windows'
 
-        if RUBY_PLATFORM =~ /mswin|mingw32|windows/
-          def initialize(new_resource,run_context)
-            super
-            @net_group = Chef::Util::Windows::NetGroup.new(@new_resource.group_name)
-          end
+        def initialize(new_resource,run_context)
+          super
+          @net_group = Chef::Util::Windows::NetGroup.new(@new_resource.group_name)
         end
 
         def load_current_resource
--- a/lib/chef/provider/user/windows.rb
+++ b/lib/chef/provider/user/windows.rb
@@ -29,11 +29,9 @@ class Chef
 
         provides :user, os: "windows"
 
-        if RUBY_PLATFORM =~ /mswin|mingw32|windows/
-          def initialize(new_resource,run_context)
-            super
-            @net_user = Chef::Util::Windows::NetUser.new(@new_resource.username)
-          end
+        def initialize(new_resource,run_context)
+          super
+          @net_user = Chef::Util::Windows::NetUser.new(@new_resource.username)
         end
 
         def load_current_resource
--- a/spec/unit/platform_spec.rb
+++ b/spec/unit/platform_spec.rb
@@ -103,7 +103,7 @@ describe Chef::Platform do
     end
 
     it "should raise an exception if a provider cannot be found for a resource type" do
-      expect { Chef::Platform.find_provider("Darwin", "9.2.2", :coffee) }.to raise_error(ArgumentError)
+      expect { Chef::Platform.find_provider("Darwin", "9.2.2", :coffee) }.to raise_error(Chef::Exceptions::ProviderNotFound)
     end
 
     it "should look up a provider for a resource with a Chef::Resource object" do
--- a/spec/unit/provider_resolver_spec.rb
+++ b/spec/unit/provider_resolver_spec.rb
@@ -106,13 +106,23 @@ describe Chef::ProviderResolver do
           end
         end
 
-        if expected_resource || expected_provider
-          it "resolves to #{[ expected_resource, expected_provider ].compact.join(" and ")}", *tags do
-            expect(resource.class).to eql(expected_resource) if expected_resource
-            expect(resolved_provider).to eql(expected_provider) if expected_provider
+        if expected_resource && expected_provider
+          it "'#{name}' resolves to resource #{expected_resource} and provider #{expected_provider}", *tags do
+            expect(resource.class).to eql(expected_resource)
+            provider = double(expected_provider, class: expected_provider)
+            expect(provider).to receive(:action=).with(action)
+            expect(expected_provider).to receive(:new).with(resource, run_context).and_return(provider)
+            expect(resolved_provider).to eql(expected_provider)
+          end
+        elsif expected_provider
+          it "'#{name}' resolves to provider #{expected_provider}", *tags do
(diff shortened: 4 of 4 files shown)


lib/chef/exceptions.rb
```diff
@@ -97,6 +97,7 @@ class GroupIDNotFound < ArgumentError; end
     class ConflictingMembersInGroup < ArgumentError; end
     class InvalidResourceReference < RuntimeError; end
     class ResourceNotFound < RuntimeError; end
+    class ProviderNotFound < ArgumentError; end
```
jaym: You're creating a new class because `NoProviderAvailable` because you want to preserve the parent class?

jkeiser: I was just creating a new class in order to have an analogue to ResourceNotFound.  Up until now, that codepath was raising `ArgumentError`, so I feel like we need it to be a subclass of that--might be overthinking it though.

I'll check on NoProviderAvailable.  I'd like the resource and provider one to be named similarly.

jkeiser: Let's just have this throw NoProviderAvailable, screw ArgumentError, and create an alias for either `NoResourceAvailable` or `ProviderNotFound`.  Got a preference for which?

jaym: I think it's fine as is

jaym: But who could be catching some as generic as ArgumentError... I'm torn

jkeiser: @jdm fixed. I think RuntimeError makes marginally more sense. And I like it when names are consistent.

lib/chef/provider_resolver.rb
```diff
@@ -114,13 +101,25 @@ def maybe_chef_platform_lookup(resource)
...
+    def priority_map
       Chef::Platform::ProviderPriorityMap.instance
     end
 
-    def prioritized_handlers
-      @prioritized_handlers ||=
-        provider_priority_map.list_handlers(node, resource.resource_name).flatten(1).uniq
+    # @api private
+    def potential_handlers
+      priority_map.list_handlers(node, resource.resource_name).flatten(1).uniq
```
jaym: Did you not memoize intentionally?

lib/chef/resource.rb
```diff
@@ -1304,7 +1304,7 @@ def self.sorted_descendants
     end
     def self.inherited(child)
       super
-      @sorted_descendants = nil
+      @@sorted_descendants = nil
```
jaym: I don't understand why this changed

lamont-granquist:
massive improvement in my test perf case.

lib/chef/resource_resolver.rb
```diff
@@ -107,7 +107,7 @@ def list
     #
     # @api private
     def provided_by?(resource_class)
-      !prioritized_handlers.include?(resource_class)
+      potential_handlers.include?(resource_class)
```
jaym: why was it `!prioritized_handers.include?` before?

jkeiser: It was wrong--prioritized_handlers is a list of classes filtered by node, while provided_by? is supposed to ignore node and just tell you whether resource_class is involved at all, on *any* node.

lib/chef/resource.rb
jkeiser: It was wrong: that code is supposed to reset the cache on the list of descendants when you create a new subclass of `Chef::Resource`, but it's resetting the wrong variable.  The tests failed once I started using sorted_descendants for real.

jaym:
:+1:

lib/chef/resource_resolver.rb
```diff
@@ -98,59 +99,70 @@ def resolve
...
+    #
+    # @api private
+    def self.includes_handler?(resource_name, resource_class)
+      priority_map.includes_handler?(resource_name, resource_class)
     end
 
     protected
 
+    def self.priority_map
+      Chef::Platform::ResourcePriorityMap.instance
```
coderanger: Should this, and below, be using `Chef.resource_priority_map` instead of the Singleton API?

lib/chef/resource_resolver.rb
```diff
@@ -56,6 +56,7 @@ def self.list(resource_name, node: nil, canonical: nil)
     attr_reader :resource_name
     # @api private
     def resource
+      raise 'omg'
```
coderanger: Guessing this was left in from manual testing?

lib/chef/resource_resolver.rb
```diff
@@ -98,59 +99,70 @@ def resolve
...
     def priority_map
       Chef::Platform::ResourcePriorityMap.instance
     end
 
-    def prioritized_handlers
-      @prioritized_handlers ||=
-        priority_map.list_handlers(node, resource_name, canonical: canonical)
+    # @api private
+    def potential_handlers
+      priority_map.list_handlers(node, resource_name, canonical: canonical).flatten(1).uniq
```
coderanger: Given that this `.flatten(1).uniq` exists in both places that use `list_handlers` and that expression is suuuuupre critical to the functioning of this code, maybe it should be folded in to `list_handlers` itself? At a minimum there should be some comments that changing the format requires updating all three places.

coderanger:
Overall :+1:. This also fixes the back-compat detection code to actually work.

lib/chef/node_map.rb
```diff
@@ -55,18 +48,18 @@ def set(key, value, platform: nil, platform_version: nil, platform_family: nil,
...
       if insert_at
-        @map[key].insert(insert_at, new_matcher)
+        map[key].insert(insert_at, new_matcher)
       else
-        @map[key] << new_matcher
+        map[key] << new_matcher
       end
-      insert_at ||= @map[key].size - 1
-      @map
+      insert_at ||= map[key].size - 1
```
mcquin: `insert_at` looks unused after this point.

jkeiser: Good catch, I'm guessing it's a remnant.

lib/chef/resource_resolver.rb
jkeiser: Looks like it already is, and I was just adding redundant code :)

lib/chef/resource_resolver.rb
jkeiser: Thanks!  Fixed.

lib/chef/resource_resolver.rb
jkeiser: I can dig it.

PR #3612, opened 2015-06-29
Move WorkstationConfigLoader into chef-config

mcquin:
This is needed for [ohai config rfc](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md) so that Ohai can find local workstation config files.

@chef/client-maintainers @chef/client-core


chef-config/lib/chef-config/exceptions.rb
```diff
@@ -20,7 +20,7 @@
 
 module ChefConfig
 
-  class InvalidPath < StandardError
-  end
+  class ConfigurationError < ArgumentError; end
```
coderanger: Do we want put subclass stubs in Chef that inherit from these for API compat?

mcquin: @danielsdeleo @jkeiser ?

mcquin: @coderanger [this should take care of it](https://github.com/chef/chef/blob/6a4f414666b3688a3dfa94b131f83235f4d2fa64/lib/chef/exceptions.rb#L28)

coderanger: :+1:

danielsdeleo:
:+1: awesome :100:

tyler-ball:
:+1:

mcquin:
I believe this requires a release of the chef-config gem for Ohai to be able to use these changes.

Issue #3621, opened 2015-06-30
Deprecation Warnings For LWRPs


jaym:
#3603 introduced more regressions. This code is very complex, and it's purpose is to display a deprecation warning. I think we need to reduce our risk here and simply this down. We have 2 proposals here, both involve using the old behavior unless `Chef::Config[:treat_deprecation_warnings_as_errors]` is `true`.

Proposal one:
-------------------
In this case, we do not set `Chef::Resource::LwrpName` if `:treat_deprecation_warnings_as_errors`. This will fail hard on any access the `Chef::Resource::LwrpName`
```ruby
def self.register_deprecated_lwrp_class(resource_class, class_name)
  if !Chef::Config[:treat_deprecation_warnings_as_errors]
    Chef::Resource.const_set(class_name, resource_class)
  end
end
```

Proposal two:
------------------
We register a class that will fail if initialized when `:treat_deprecation_warnings_as_errors` is set. There are some cases where this does not work, like `my_resource.is_a?(Chef::Resource::LwrpFoo)`
```ruby
class ClassThatFailsIfInitialized
  def initialize
    Chef::Log.deprecation("You cannot access LWRP's through Chef::Resource and Chef::Provider")
  end
end
 
def self.register_deprecated_lwrp_class(resource_class, class_name)
  if Chef::Config[:treat_deprecation_warnings_as_errors]
    Chef::Resource.const_set(class_name, ClassThatFailsIfInitialized)
  else
    Chef::Resource.const_set(class_name, resource_class)
  end
end
```
cc @chef/client-maintainers

danielsdeleo:
Option two seems preferable at first glance. :+1: on abandoning deprecation warnings for this, there's just too many edge cases that are very common and are getting broken.

lamont-granquist:
:+1: on option 2 i guess

if we have to do this it kind of scuttles the idea of doing rolling deprecations without major version bumps.

danielsdeleo:
@lamont-granquist I think there's other ways to go forward, like wrapping all LWRP creation in `Chef.create_provider(api_level = 1) do; end`, the old stuff becomes version 0 and we can detect very easily what version you're on.

jkeiser:
:+1: on option 1, for 12.4.1 at least!

The deprecated class is the thing that got us into this problem in the first place; option 2 will get users into weird places where things like is_a? don't work as expected, and don't emit a warning.

jaym:
https://github.com/chef/chef/pull/3623
[...]

PR #3623, opened 2015-06-30
Simplify LWRP Deprecations Proposal 1

jaym:
Our logic for creating the deprecation class was too complicated and had
a lot of edge cases. Simplifying here to only deprecate when
`:treat_deprecation_warnings_as_errors` is set. This means that warnings
will no longer be printed, however this is a lot less risky.

Loading the lwrp multiple times causes problems when comparing classes
due to the their dynamic nature. It worked fine when we were overriding
the things that checked.

I've left the tests mostly as is, other than the fact that the test
lwrp is loaded only once. The tests should still hold true, even
with the new implementation.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1499,56 +1499,11 @@ class Chef
         Chef::Resource.send(:remove_const, class_name)
       end
 
-      # In order to generate deprecation warnings when you use Chef::Resource::MyLwrp,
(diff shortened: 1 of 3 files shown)


tyler-ball:
:+1:

thommay:
I like pink :+1:

jaym:
closes #3621

PR #3624, opened 2015-06-30
Add "property" with identity and desired_state

jkeiser:
Adds a Property class to encapsulate property behavior with getters, setters, and defaults; stores identity and desired_state on them; drives identity_attr and state_attrs based on that.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -1,0 +1,491 @@
+require 'chef/exceptions'
+require 'chef/delayed_evaluator'
+
+class Chef
+  #
+  # Type and validation information for a property on a resource.
+  #
+  # A property named "x" manipulates the "@x" instance variable on a
+  # resource.  The *presence* of the variable (`instance_variable_defined?(@x)`)
+  # tells whether the variable is defined; it may have any actual value,
+  # constrained only by validation.
+  #
+  # Properties may have validation, defaults, and coercion, and have full
+  # support for lazy values.
+  #
+  # @see Chef::Resource.property
+  # @see Chef::DelayedEvaluator
+  #
+  class Property
+    #
+    # Create a new property.
+    #
+    # @param options [Hash<Symbol,Object>] Property options, including
+    #   control options here, as well as validation options (see
+    #   Chef::Mixin::ParamsValidate#validate for a description of validation
+    #   options).
+    #   @option options [Symbol] :name The name of this property.
+    #   @option options [Class] :declared_in The class this property comes from.
+    #   @option options [Symbol] :instance_variable_name The instance variable
+    #     tied to this property. Must include a leading `@`. Defaults to `@<name>`.
+    #     `nil` means the property is opaque and not tied to a specific instance
+    #     variable.
+    #   @option options [Boolean] :desired_state `true` if this property is part of desired
+    #     state. Defaults to `true`.
+    #   @option options [Boolean] :identity `true` if this property is part of object
+    #     identity. Defaults to `false`.
+    #   @option options [Boolean] :name_property `true` if this
+    #     property defaults to the same value as `name`. Equivalent to
+    #     `default: lazy { name }`, except that #property_is_set? will
+    #     return `true` if the property is set *or* if `name` is set.
+    #   @option options [Object] :default The value this property
+    #     will return if the user does not set one. If this is `lazy`, it will
+    #     be run in the context of the instance (and able to access other
+    #     properties) and cached. If not, the value will be frozen with Object#freeze
+    #     to prevent users from modifying it in an instance.
+    #   @option options [Proc] :computed The value this property will return if
+    #     the user does not set one. If this is `lazy`, it will be run in the
+    #     context of the instance (and able to access other properties).
+    #   @option options [Proc] :coerce A proc which will be called to
+    #     transform the user input to canonical form. The value is passed in,
+    #     and the transformed value returned as output. Lazy values will *not*
+    #     be passed to this method until after they are evaluated. Called in the
+    #     context of the resource (meaning you can access other properties).
+    #   @option options [Boolean] :required `true` if this property
+    #     must be present; `false` otherwise. This is checked after the resource
+    #     is fully initialized.
+    #
+    def initialize(**options)
+      options.each { |k,v| options[k.to_sym] = v if k.is_a?(String) }
+      options[:name_property] = options.delete(:name_attribute) unless options.has_key?(:name_property)
+      @options = options
+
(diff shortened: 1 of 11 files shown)


jkeiser:
@coderanger I believe this is now fully in line with your comments on the old PR, mind taking a look when you get a chance?

lib/chef/resource.rb
```diff
@@ -772,29 +779,88 @@ def provider=(arg)
...
+    # @example With just a type
+    #   property_type(String)
+    #
+    # @example With just options
+    #   property_type(default: 'hi')
+    #
+    # @example With type and options
+    #   property_type(String, default: 'hi')
+    #
+    def self.property_type(type=NOT_PASSED, **options)
```
coderanger: Is there a reason we are converting name to a keyword param but leaving the type as a positional at this point?

lib/chef/property.rb
```diff
@@ -0,0 +1,577 @@
...
+    #   property_type(String)
+    #
+    # @example With just options
+    #   property_type(default: 'hi')
+    #
+    # @example With type and options
+    #   property_type(String, default: 'hi')
+    #
+    def self.create(type=NOT_PASSED, **options)
+      # Inherit from the property type, if one is passed
```
coderanger: This isn't inheriting from a property type, it's using a specific property instance.

lib/chef/resource.rb
coderanger: Or to be more specific, it would probably be better for API stability to make this `(**options)` since we are expecting external code it monkey patch this method. Type can just become `options[:type]` in the code calling this.

lib/chef/property.rb
```diff
@@ -0,0 +1,577 @@
...
+            value = exec_in_resource(resource, value)
+          end
+
+          value = coerce(resource, value)
+
+          # We don't validate defaults
+
+          # If the value is mutable (non-frozen), we set it on the instance
+          # so that people can mutate it.  (All constant default values are
+          # frozen.)
```
coderanger: Not a reason to change this, but just keep in mind that strings aren't constant in Ruby by default.

lib/chef/property.rb
```diff
@@ -0,0 +1,577 @@
...
+    #   @option options [Boolean] :name_property `true` if this
+    #     property defaults to the same value as `name`. Equivalent to
+    #     `default: lazy { name }`, except that #property_is_set? will
+    #     return `true` if the property is set *or* if `name` is set.
+    #   @option options [Object] :default The value this property
+    #     will return if the user does not set one. If this is `lazy`, it will
+    #     be run in the context of the instance (and able to access other
+    #     properties) and cached. If not, the value will be frozen with Object#freeze
+    #     to prevent users from modifying it in an instance.
+    #   @option options [Proc] :computed The value this property will return if
```
coderanger: This feature was removed.

lib/chef/property.rb
```diff
@@ -0,0 +1,577 @@
...
+    def has_default?
+      options.has_key?(:default) || name_property?
+    end
+
+    #
+    # Whether this property has a computed default value.
+    #
+    # @return [Boolean]
+    #
+    def has_computed?
```
coderanger: As above, feature was removed.

coderanger:
Simpsons dancing

lib/chef/resource.rb
jkeiser: Sure, I can see that reasoning.

jkeiser: @coderanger right now, `options[:is]` is generally the equivalent to `type` (and the only difference is that if you specify both, `type` gets merged into the front of `is`).  I'm trying to decide the pros and cons of keeping them separate in `options`; it's interesting, but I'm not sure whether it ought to be done or not.  There are some tempting bits, though.

coderanger: Just make a subclass of Hash where they are aliased!</bad ideas> *(don't actually do that)*

lib/chef/property.rb
jkeiser: What this means is non-lazy default values--we freeze any non-lazy default values when the type is created (actually we freeze all default values, but freezing a DelayedEvaluator doesn't really have much effect :).  Not sure if there is a better way to put that?

lib/chef/resource.rb
jkeiser: Fixed it by moving all type derivation stuff into `property`.  `property_type` now only receives `is`.  If you pass a Property instance like `property :x, awesome_array_type, default: [ 10 ]`, for example, we will call `awesome_array_type.derive(default: [ 10 ])`.  That way you inherit coercion, defaults, and other behavior from the type.

lib/chef/property.rb
jkeiser: Fixed, thanks.

lib/chef/property.rb
jkeiser: Removed, thanks!

lib/chef/property.rb
jkeiser: Removed, thanks.

lib/chef/property.rb
coderanger: I mean if coerce returns a string, we'll sticky that even though most people would probably group strings with integers and booleans as a basic type rather than with arrays/hashes. I think the behavior is intuitive in the end though, so it is fine, but we should make the docs not rely on users knowing what is and isn't frozen by default :)

coderanger:
Looks ready to ship in my book! (give or take tests)

lib/chef/property.rb
```diff
@@ -0,0 +1,539 @@
...
+    def call(resource, value=NOT_PASSED)
+      if value == NOT_PASSED
+        return get(resource)
+      end
+
+      # myprop nil is sometimes a get (backcompat)
+      if value.nil? && !explicitly_accepts_nil?(resource)
+        # If you say "my_property nil" and the property explicitly accepts
+        # nil values, we consider this a get.
+        Chef::Log.deprecation("#{name} nil currently does not overwrite the value of #{name}. This will change in Chef 13, and the value will be set to nil instead. Please change your code to explicitly accept nil using \"property :#{name}, [MyType, nil]\", or stop setting this value to nil.")
```
poliva83: @jkeiser @coderanger This warning is showing up all over place on because of one particular resource in a default recipe. Trying to understand meaning of: `Please change your code to explicitly accept nil using \"property :#{name}, [MyType, nil]\", or stop setting this value to nil.` 

Is there any documentation around `property`? This seems like a fairly new concept and trying general understanding around it.

PR #3627, opened 2015-07-01
Re-separate priority map and DSL handler map so that provides has vet…

jkeiser:
…o power over priority

### DIFF
--- a/lib/chef/chef_class.rb
+++ b/lib/chef/chef_class.rb
@@ -28,6 +28,8 @@
 
 require 'chef/platform/provider_priority_map'
 require 'chef/platform/resource_priority_map'
+require 'chef/platform/provider_handler_map'
+require 'chef/platform/resource_handler_map'
 
 class Chef
   class << self
@@ -160,20 +162,26 @@ class Chef
       @node = nil
       @provider_priority_map = nil
       @resource_priority_map = nil
+      @provider_dsl_map = nil
+      @resource_dsl_map = nil
     end
 
     # @api private
     def provider_priority_map
-      @provider_priority_map ||= begin
-        # these slurp in the resource+provider world, so be exceedingly lazy about requiring them
-        Chef::Platform::ProviderPriorityMap.instance
-      end
+      # these slurp in the resource+provider world, so be exceedingly lazy about requiring them
+      @provider_priority_map ||= Chef::Platform::ProviderPriorityMap.instance
     end
     # @api private
     def resource_priority_map
-      @resource_priority_map ||= begin
-        Chef::Platform::ResourcePriorityMap.instance
-      end
+      @resource_priority_map ||= Chef::Platform::ResourcePriorityMap.instance
+    end
+    # @api private
+    def provider_handler_map
+      @provider_handler_map ||= Chef::Platform::ProviderHandlerMap.instance
+    end
+    # @api private
+    def resource_handler_map
+      @resource_handler_map ||= Chef::Platform::ResourceHandlerMap.instance
     end
   end
 
--- a/lib/chef/chef_class.rb
+++ b/lib/chef/chef_class.rb
@@ -162,8 +162,8 @@ class Chef
       @node = nil
       @provider_priority_map = nil
       @resource_priority_map = nil
-      @provider_dsl_map = nil
-      @resource_dsl_map = nil
+      @provider_handler_map = nil
+      @resource_handler_map = nil
     end
 
     # @api private
--- a/lib/chef/platform/provider_handler_map.rb
+++ b/lib/chef/platform/provider_handler_map.rb
@@ -1,10 +1,10 @@
 require 'singleton'
-require 'chef/platform/priority_map'
+require 'chef/platform/handler_map'
 
 class Chef
   class Platform
     # @api private
-    class ResourcePriorityMap < Chef::Platform::PriorityMap
+    class ProviderHandlerMap < Chef::Platform::HandlerMap
       include Singleton
     end
   end
--- a/lib/chef/platform/provider_handler_map.rb
+++ b/lib/chef/platform/provider_handler_map.rb
@@ -1,3 +1,21 @@
+#
+# Author:: John Keiser (<jkeiser@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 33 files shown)


coderanger:
Can you provide a little more background on "priority map" vs. "handler map"?

jkeiser:
Handler map is the provides lines--the declarations that say what an individual resource is willing to do. It's the global version of what used to be resource-specific node maps. Priority map is the list of priority arrays, just like before.

This check-in splits priority and provides back out separately because it was not backwards compatible: specifically, if the priority array for :dsl has A and B in it, and A.provides(:dsl) { false }, then we are required to pick B.  The system I built with one map would pick A and never run the block.

lib/chef/provider/package/portage.rb
```diff
@@ -25,6 +25,8 @@ class Chef
   class Provider
     class Package
       class Portage < Chef::Provider::Package
+
+        provides :package, platform: "gentoo"
         provides :portage_package
```
jaym: probably want `provides :portage_package, os: linux`?

lib/chef/provider_resolver.rb
```diff
@@ -68,27 +68,39 @@ def provided_by?(provider_class)
...
+
+    # The list of handlers, with any in the priority_map moved to the front
+    def prioritized_handlers
+      @prioritized_handlers ||= begin
+        supported_handlers = self.supported_handlers
+        if supported_handlers.empty?
+          # if none of the providers specifically support the resource, we still need to pick one of the providers that are
+          # enabled on the node to handle the why-run use case. FIXME we should only do this in why-run mode then.
+          Chef::Log.debug "No providers responded true to `supports?` for action #{action} on resource #{resource}, falling back to enabled handlers so we can return something anyway."
+          supported_handlers = enabled_handlers
```
jaym: This case causes work that's been done before to be done again. Not sure if it will have any measurable performance impact

lib/chef/provider/package/portage.rb
lamont-granquist: It doesn't really matter.  We probably do need to decide if we care or not, but right now there's no rule to follow.

lamont-granquist:
Passes in my test-kitchen test

:+1:

jaym:
:+1:

lamont-granquist:
@coderanger the priority_map in 12.0 through 12.3 was applied after searching through and finding the list of possibly acceptable providers.  So if you got `[ Init, Redhat, Systemd ]` or something like that it would sort them as a second phase.  12.4.0 broke that and if you applied a priority array it would just use that array instead of doing the first pass to see what was acceptable or not.  This still kind of worked because most of the providers that cared about this behavior override `#supports?` or `#provides?` which was still called in 12.4.0 effectively as a `global filter`.  The biggest thing this breaks is that passing a block to `provides` is no longer equivalent to overriding `#provides?` and the former gets ignored in the face of a priority_map while the latter does not, which is a breaking change from 12.3.0 to 12.4.0

lib/chef/chef_class.rb
```diff
@@ -160,20 +162,26 @@ def reset!
       @node = nil
       @provider_priority_map = nil
       @resource_priority_map = nil
+      @provider_dsl_map = nil
```
coderanger: I think you mean `handler_map` for these two.

lib/chef/platform/provider_handler_map.rb
```diff
@@ -0,0 +1,11 @@
+require 'singleton'
```
coderanger: License header.

lib/chef/platform/resource_handler_map.rb
```diff
@@ -0,0 +1,11 @@
+require 'singleton'
```
coderanger: License header.

lib/chef/resource_resolver.rb
```diff
@@ -127,19 +127,38 @@ def self.priority_map
...
     def enabled_handlers
       potential_handlers.select { |handler| !overrode_provides?(handler) || handler.provides?(node, resource_name) }
     end
 
+    def prioritized_handlers
+      @prioritized_handlers ||= begin
+        enabled_handlers = self.enabled_handlers
+
+        prioritized = priority_map.list(node, resource_name, canonical: canonical).flatten(1)
+        prioritized &= enabled_handlers # Filter the priority map by the actual enabled handlers
```
coderanger: This is very efficient, but maybe something like `enabled_handlers.sort_by {|x| prioritized.find_index(x)}` might be clearer?

coderanger: Also this sort logic is duplicated between resource and provider resolvers. Maybe move it to a method on the handler map like `list_with_priority`?

coderanger:
:+1: 

Minor jargon collision on "chef handler" vs. "resource/provider handler" but that seems fine given most users won't touch these APIs.

lib/chef/resource_resolver.rb
jkeiser: The issue (and the reason I was being so careful here) is that Ruby's sort is not stable: if there is no priority array, this particular `sort_by` will scramble the array.

coderanger: TIL :-( Lolrubby. Still would vote to move the code to the handle map base class.

lib/chef/provider_resolver.rb
```diff
@@ -68,27 +68,39 @@ def provided_by?(provider_class)
...
     end
 
-    def self.includes_handler?(resource_name, provider_class)
-      priority_map.includes_handler?(resource_name, provider_class)
-    end
-
     def enabled_handlers
-      @enabled_handlers ||=
-        potential_handlers.select { |handler| !overrode_provides?(handler) || handler.provides?(node, resource) }
+      potential_handlers.select { |handler| !overrode_provides?(handler) || handler.provides?(node, resource) }
```
lamont-granquist: yeah, do we have to drop the memoization here, and below, and why?

jkeiser: Nope. I don't think we need to do it as many places as we are, though.

lamont-granquist: Hmmm, I typically like to do it when i can so that in later refactorings I can just re-use the method and not have to worry about performance...

jkeiser: It's not too big a deal as a rule, but in general gc'ing runtimes can
optimize better when things remain entirely on the stack.

lamont-granquist: Even though this gets called significantly though I doubt its generating enough garbage to tax the GC compared to other things that we do, and the objects will be short-lived since the provider-resolver object itself is short lived.  The Ruby 2.x generational GC should handle it.

jkeiser: I'm not super attached, like I said; it's just my default to leave things
on the stack wherever possible, and optimize when I must. We certainly
don't want to redo the enabled handlers work, so we should cache that at
least.

PR #3632, opened 2015-07-02
Make required name attributes work

jkeiser:
Right now we evaluate required before name_attribute, which means we didn't know it had a value.


coderanger:
:+1:

PR #3633, opened 2015-07-02
add the word group to converge_by call for group provider

jordane:
Other `converge_by` calls specify what resource they are acting on, and other calls in the group provider also do so, but for some reason group creation doesn't.


stevendanna:
:+1:

lamont-granquist:
:+1:

smurawski:
:+1:

PR #3635, opened 2015-07-02
Revert "Freeze default constants"

jaym:
This reverts commit 705e22b975196baa6decf23f9c3a7e676d6aefc4.

I'd like to revert this change because it breaks the packagecloud
cookbook.

See https://gist.github.com/cwebberOps/ae002be2339ca843cbb3

### DIFF
--- a/lib/chef/mixin/params_validate.rb
+++ b/lib/chef/mixin/params_validate.rb
@@ -317,7 +317,6 @@ class Chef
       def _pv_default(opts, key, default_value)
         value = _pv_opts_lookup(opts, key)
         if value.nil?
-          default_value = default_value.freeze if !default_value.is_a?(DelayedEvaluator)
           opts[key] = default_value
         end
       end
--- a/spec/unit/property_spec.rb
+++ b/spec/unit/property_spec.rb
@@ -367,13 +367,9 @@ describe "Chef::Resource.property" do
           expect(resource.x.object_id).to eq(value.object_id)
         end
         it "Multiple instances of x receive the exact same value" do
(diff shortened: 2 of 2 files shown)


coderanger:
:-1: The whole point of this change was to reveal bugs in cookbook code, which it has done.

jaym:
Yea, I see it now. We don't want people modifying the default value

jaym:
I think we want to dup the default value

coderanger:
No, that would be a different bug. Mutation would be allowed but discarded. If you want a mutable default, wrap it in a `lazy{}`.

jaym:
@coderanger what example are you thinking of. Where are people trying to actually modify the value they assign to default?

coderanger:
[The example you gave](https://github.com/computology/packagecloud-cookbook/blob/master/providers/repo.rb#L212-L214).

coderanger:
That code should be `str + '/'` rather than trying to be tricky about mutating a string in place.

jaym:
Yep, and that's wrong as it stands. But when we get a default value in `set_or_return`, why couldn't we just dup it? If somebody actually expecting to be able to modify the original one?

jaym:
opened a pr against pacakge cloud cookbook:
https://github.com/computology/packagecloud-cookbook/pull/18

jkeiser:
Thanks for holding the line here guys.  I think we could get away with dup and keep writeability, since the default would be sticky; but that comes with its own weirdness since dup doesn't work on all objects.  Honestly, this particular cookbook is just itching for coercion!

PR #3636, opened 2015-07-05
Clean up some RSpec errors

juliandunn:
Things that recent rspec runs warned about:

* Using the `raise_error` matcher without providing a specific error or message risks false positives
* Hashes with duplicated keys


juliandunn:
I'll come back to this later.

PR #3639, opened 2015-07-07
12.4.1

jaym:


lamont-granquist:
That looks good to start with, I don't see anything that shouldn't be there.

lib/chef/api_client.rb
```diff
@@ -1,7 +1,7 @@
 #
-# Author:: Adam Jacob (<adam@chef.io>)
-# Author:: Nuo Yan (<nuo@chef.io>)
-# Copyright:: Copyright (c) 2008, 2015 Chef Software, Inc.
+# Author:: Adam Jacob (<adam@opscode.com>)
+# Author:: Nuo Yan (<nuo@opscode.com>)
+# Copyright:: Copyright (c) 2008 Opscode, Inc.
```
coderanger: Was this intentional?

lamont-granquist: It was a side effect of reverting the commit which updated that.

We should probably do a mass update of the copyright headers to prevent this kind of noise.

coderanger: :+1: I think rubocop has a rule you can enable to check it too.

PR #3645, opened 2015-07-07
Making it possible for delayed resources to know if the current chef run has crashed so far

wk8:
It sometimes happens that some delayed resources get notifications from several other resources,
and that sadness can ensue if some of these notifying resources have run but not others.

This patch makes it possible, in such cases, to test if a delayed resource is running after
another resource has crashed or not, with
```
node.run_context.has_crashed?
```

Updated unit tests.

Note that Travis builds will fail until https://github.com/chef/cheffish/pull/53 is merged.

### DIFF
--- a/lib/chef/run_context.rb
+++ b/lib/chef/run_context.rb
@@ -144,6 +144,8 @@ class Chef
       @reboot_info = {}
       @cookbook_compiler = nil
 
+      @has_crashed = false
+
       initialize_child_state
     end
 
@@ -522,6 +524,22 @@ ERROR_MESSAGE
       ChildRunContext.new(self)
     end
 
+    #
+    # Marks the run_context as having crashed
+    #
+    def has_crashed!
+      @has_crashed = true
+    end
+
+    #
+    # Returns true if the current run has crashed
+    #
+    # @return[True/False] The current value
+    #
+    def has_crashed?
+      @has_crashed
+    end
+
     protected
 
     attr_reader :cookbook_compiler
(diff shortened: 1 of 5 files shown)


coderanger:
This feels like a change of just big enough scope that it should probably go through the RFC process. Or at least I think we should fully list out the change to semantics.

ranjib:
fyi you can tap into event handlers for this detection, it has run_failed.

wk8:
@coderanger : is it? It's not changing any behavior, but simply just adding a new hook, and a pretty minor one at that?

wk8:
@ranjib : indeed, but it seems to me looking at the code that `run_failed` is called after delayed notifs are run? Event handlers are nice, but lack a lot of bells and whistles that come with regular resources.

ranjib:
@wk8 theres converge_failed and resource_failed as well.

wk8:
@ranjib : same. Called after running delayed notifs.

wk8:
Ping? :-)

coderanger:
@wk8 As stated above, please file an RFC as this has API design implications.

lamont-granquist:
Yeah this needs to go into an RFC on how to handle the case when we don't want to always run a delayed notification since the CCR may have crashed before some critical resource was run and the notification borks the service or whatever.  It has subtle interactions with the problem of triggering a delayed notification because something was updated and then something completely orthogonal crashes the run so the notification is never executed.  This solution is probably not what you want and just pushes the former problem into the latter problem.

PR #3650, opened 2015-07-09
Define == for node objects

dbresson:
Define the == method on Chef::Node as specified in CHEF-4159 in the old JIRA.

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -550,6 +550,14 @@ class Chef
       "node[#{name}]"
     end
 
+    def ==(other)
+      if other.kind_of?(self.class)
+         self.name == other.name
+      else
+        false
+      end
+    end
+
     def <=>(other_node)
       self.name <=> other_node.name
     end
--- a/spec/unit/node_spec.rb
+++ b/spec/unit/node_spec.rb
@@ -42,6 +42,14 @@ describe Chef::Node do
     expect([n3, n1, n2].sort).to eq([n1, n2, n3])
   end
 
+  it "should share identity only with others of the same name" do
+    n1 = Chef::Node.build('foo')
+    n2 = Chef::Node.build('foo')
+    n3 = Chef::Node.build('bar')
+    expect(n1).to eq(n2)
+    expect(n1).not_to eq(n3)
+  end
+
   describe "when the node does not exist on the server" do
     before do
       response = OpenStruct.new(:code => '404')


coderanger:
Is there a reason to add this? It seems like it would be a big footgun to have an `==` that only checks the name.

dbresson:
There was some discussion on this in the JIRA ticket:
https://tickets.opscode.com/browse/CHEF-4159.

The == method in ruby is basically an identity comparison, and for nodes, the node name is basically the "primary key". Two node objects that have the same name are referring to the same node, even if they have some different values for attributes.

The classic example of the previous, counter-intuitive behavior is
```
chef > search(:node,"name:#{node.name}").include? node
 => false 
```

coderanger:
That's a better counter example, but I feel like this still has some weird consequences. Like two nodes  with otherwise different data being `==` seems wrong.

dbresson:
As I pointed out in the original discussion, ActiveRecord does the same thing, with the same consequence.

Look at it this way: If you have two node objects, and node#save will cause them to overwrite each other, they're the same node and should be ==. If you really want to see if two node objects are the identical ruby object, you still have eql? for that.

lamont-granquist:
what about `#===` semantics?  and i think you also need to alias `#eql?` explicitly.

also it seems like there's a third equality that needs to be covered which is true if the ruby objects are different (equal? is false) but where all the fields must match.

that third equality statement might be useful for doing a node.save followed by searching in a loop until the new node state showed up.

dbresson:
#===, the case equality, uses == by default. I don't see a reason for that to be different.

Implementing a method that compares all fields is a separate issue. This PR just makes it so behavior of things like Array#include? for node objects does a better job of following the principle of least surprise.

coderanger:
Yeah, fixing `include?` seems like a more important issue than my quibbling about attribute data and whatnot.

lamont-granquist:
does `#eql?` use `#==` by default?  I thought I ran into a case where I needed to override both of those recently, but i could be misremembering...

lamont-granquist:
And yeah I thought about field-by-field comparison more and that's going to be heavy and messy and nobody has asked for it so far.

stevendanna:
From http://ruby-doc.org/core-2.0.0/Object.html#method-i-eql-3F

> The eql? method returns true if obj and other refer to the same hash key. This is used by Hash to test members for equality. For objects of class Object, eql? is synonymous with ==. Subclasses normally continue this tradition by aliasing eql? to their overridden == method, but there are exceptions. Numeric types, for example, perform type conversion across ==, but not across eql?

Testing in irb shows that eql? doesn't use == by default:

```
irb(main):014:0> class Foo
irb(main):015:1> def ==(other)
irb(main):016:2> puts 'foo'
irb(main):017:2> true
irb(main):018:2> end
irb(main):019:1> end
=> :==
irb(main):020:0> f1 = Foo.new
=> #<Foo:0x007fd7e398ec58>
irb(main):021:0> f2 = Foo.new
=> #<Foo:0x007fd7e3989410>
irb(main):022:0> f1 == f2
foo
=> true
irb(main):023:0> f1 === f2
foo
=> true
irb(main):024:0> f1.eql?(f2)
=> false
irb(main):025:0> f1.equal?(f2)
=> false
```

Do we want two nodes that have the same name but different content to be the same hash key?  My feeling is that if you want that behavior you can just use `node.name` as your hash key. And that this PR is good to go as is.

lamont-granquist:
:+1:

thommay:
:+1:

PR #3655, opened 2015-07-09
add tweakable cloning behavior

lamont-granquist:
:construction:

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -720,6 +720,14 @@ module ChefConfig
     # break Chef community cookbooks and is very highly discouraged.
     default :ruby_encoding, Encoding::UTF_8
 
+    # Default resource cloning behavior when an identically named resource is found on the resource collection.
+    #  :clone  - CHEF-3694-style resource cloning behavior
+    #  :rewind - chef-rewind style behavior where the prior resource is updated in-place in the resource collection
+    #  :none   - no cloning.
+    #
+    # Setting the default to :rewind is not recommended.
+    default :default_resource_cloning_behavior, :clone
+
     # If installed via an omnibus installer, this gives the path to the
     # "embedded" directory which contains all of the software packaged with
     # omnibus. This is used to locate the cacert.pem file on windows.
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -381,13 +381,25 @@ class Chef
       end
     end
 
+    #
     # Resource reporters send event information back to the chef server for
-    # processing.  Can only be called after we have a @rest object
+    # processing
+    #
     # @api private
+    #
     def register_reporters
+      # Audit and Resource Reporters set validate_utf8 to false so that they
+      # never fail if we accidentally send binary data through (validate_utf8
+      # will raise when set to true, and sanitizes to utf8 when false)
+      reporter_rest = Chef::REST.new(
+        config[:chef_server_url],
+        client_name,
+        config[:client_key],
+        validate_utf8: false,
+      )
       [
-        Chef::ResourceReporter.new(rest),
-        Chef::Audit::AuditReporter.new(rest)
+        Chef::ResourceReporter.new(reporter_rest),
+        Chef::Audit::AuditReporter.new(reporter_rest)
       ].each do |r|
         events.register(r)
       end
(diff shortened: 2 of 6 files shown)


lamont-granquist:
@coderanger @danielsdeleo @jkeiser does anyone here like the idea of being able to set a resource to have `:rewind`-like cloning behavior?  It obviously would be a horrible default, but can you see a use case for optionally using it?

I was thinking it would make accumulators easier to write using resources, but the problem there is that you can't really incrementally build a resource property.  If properties could be a deep mergable hash then you could something like:

```
foo "bar" do
  baz thing1: true
end

foo "bar" do
  baz thing2: true
end
```

and then if foo was a rewindable resource you'd wind up with just this in the resource collection:

```
foo "bar" do
  baz { thing1: true, thing2: true }
end
```

but I don't want deep mergeable properties and without that I don't see the utility.

I was also experimenting with a way to make "singleton resources" to replace include_recipe so that you could do this instead:

```
whatever1/default.rb:
install "build-essentials"

whatever2/default.rb:
install "build-essentials"

whatever3/default.rb:
install "build-essentials" do
  compile_time true
end
```

and just wind up with one resource in the collection with compile_time overridden to true.

i don't see a way to knock the unpleasant edges off of the concept though.

coderanger:
Echoing from IRC:

That allows configuring global or per resource instance, but not per-class, which I think is the important one. Configuring it per instance seems too confusing, and global isn't that useful since it might break in unknowable ways. I think adding new cloning modes is probably counterproductive, it doesn't add enough over just using rewind.

lamont-granquist:
Yeah I thought per-instance was clever back when I thought that `:rewind` sounded clever, so would probably drop both of those, and add per-class.  Would still keep the global setting since I think that no cloning makes more sense ultimately over the existing cloning behavior and in Chef-13 or some other major version release we should flip the default.

I'm also missing thinking about how notifies and subscribes should work in the no cloning case and still explicitly providing some kind of warnings around that case.  I believe when cloning is disabled that we can better identify the trivial cases when no warnings are necessary (since we don't clone the old resource any more before slurping in all the properties for the new resource).

coderanger:
Fair on the global, it would at least offer a way for people to test the Chef 13 behavior beforehand. I still think 13 should just remove the option entirely to keep things simpler.

ranjib:
if we do not provide any cloning support and ship lwrps with cloning disabled , it will break a lot of cookbooks. so i think we need one. worst case.. implement the old behavior and feature gate it.

That aside, i think it will be nice to have a hwrp/lwrp api for defining custom cloning logic (resource.clonable? and resource.clone) , and a clone factory , that implements deepmerge, first one win, last one wins etc. most core resources can be cloned easily, for things like ruby_block user can opt in to set any of the factory based logic (like last one wins), or provide their own.. so a recipe dsl for the same will be nice as well, like clone :ruby_block, with: ->(existing, new){|existing, new| ... }

coderanger:
The only place I've heard cloning defended is service resources. We should come up with a better solution that is specific to them if needed but the overall feature should be flat out removed. We've been warning about removing it for, I think, years. Any cookbooks broken by this (given it would be part of a major release) are not my concern at this point.

lamont-granquist:
Yep what @coderanger said.

I don't think resource cloning is right, its kind of monkeypatching, and I think the use case which it attempts to solves violates separation of concerns and the cookbooks need to get refactored (it solves a non-best-practice problem and creates horrible issues for the best-practices, that's a pretty bad use case to try to solve, i think is legit to say the cookbook should get a rewrite in that case).

We probably need to be careful about notifies and subscribes when there's duplicated resource names on the resource collection and still warn.  We could have a strict-mode where duplicated resources are banned, but then every `package "name"` that is replicated would fail, and customers simply don't want to deal with that, and they don't want to see a warning in that case, because they don't care as much as we do about unique identifiers on the resource collection (and they're right that it doesn't cause any issues in that case).

It would probably be nice so that if a notifies or subscribes was in the same cookbook as the resource that it would always prefer to target that resource rather than any other identically named resource on the resource collection, which would be new behavior we'd need to build.  And then if a third party resource tried to notify or subscribe to an ambiguous resource we'd always want to at least WARN there because its ambiguous.

lamont-granquist:
closing this for now

PR #3656, opened 2015-07-09
remove use of self.provides?

lamont-granquist:
the block to provides can handle this case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [**Renan Vicente**](https://github.com/renanvicente):
   [pr#3771](https://github.com/chef/chef/pull/3771) add depth property for deploy resource
 
+* [pr#3656](https://github.com/chef/chef/pull/3656) remove use of self.provides?
 * [pr#3455](https://github.com/chef/chef/pull/3455) powershell\_script: do not allow suppression of syntax errors
 * [pr#3519](https://github.com/chef/chef/pull/3519) The wording seemed odd.
 * [pr#3208](https://github.com/chef/chef/pull/3208) Missing require (require what you use).
--- a/lib/chef/provider/service/debian.rb
+++ b/lib/chef/provider/service/debian.rb
@@ -22,14 +22,12 @@ class Chef
   class Provider
     class Service
       class Debian < Chef::Provider::Service::Init
-        provides :service, platform_family: 'debian'
+        provides :service, platform_family: 'debian' do |node|
+          Chef::Platform::ServiceHelpers.service_resource_providers.include?(:debian)
+        end
 
         UPDATE_RC_D_ENABLED_MATCHES = /\/rc[\dS].d\/S|not installed/i
         UPDATE_RC_D_PRIORITIES = /\/rc([\dS]).d\/([SK])(\d\d)/i
-
-        def self.provides?(node, resource)
-          super && Chef::Platform::ServiceHelpers.service_resource_providers.include?(:debian)
-        end
 
         def self.supports?(resource, action)
           Chef::Platform::ServiceHelpers.config_for_service(resource.service_name).include?(:initd)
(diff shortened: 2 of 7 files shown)


coderanger:
Not sure I like this direction, makes things gross for subclassing core providers.

lamont-granquist:
Hmmm, well @jkeiser suggested this.  He also wanted to try to eliminate self.supports? as well and pass resource and action into provides blocks, but that call is buried inside of node_map, which on first glance would require surgery to the node_map API.

coderanger:
Putting in blocks like that means it isn't accessible to a subclass though, not sure why it makes things any better in the long run?

lamont-granquist:
Makes it a bit more DSLey and eliminates one more thing.  I think john also wanted to try to eliminate provides? entirely now that we're not doing descendants tracking scanning and clean up the code internally.  He'll have to pop in here after he gets back from vacay though.

coderanger:
Yeah, we should probably have a longer talk about the semantics of subclassing/extending resources and providers.

lamont-granquist:
paging @jkeiser into this discussion again.

after thinking about this, the problem is that when you subclass you only get `#provides?` and not the provides methods, which is not going to be useful to subclasses of these classes anyway.   by merging this it makes it so none of the provides logic is inherited which seems actually better than half of the provides logic being inherited.  so, i'd argue that this should go in.

there's a larger argument about the utility or wisdom of inheriting `provides` and `provides?` in subclasses and if `provides?` should still exist or not, which i think i'm slightly biased against, but ultimately agnostic about.  if `provides?` is used in poise to inherit behavior in subclasses i think thats fine, but i don't see the utility in these specific cases.

jkeiser:
If it's possible, we should remove either `provides?` or blocks sent to `provides`, just to reduce what is already a sprawlingly-large number of toggles and knobs on this feature. It seems like you *can* implement either one in terms of the other, and the DSL form of `def provides` is more granular. It seems like the model works better if we can put the block in the DSL entry rather than the provider class, just to make the lookup code wholly concerned about that entry and not concerned about the class.

Here's one thing I hadn't considered: we could make it so that if you don't specify a block to the DSL, it will default to calling `provides?`, but if you do, it won't. That gets rid of the code that implements the `provides?` knob, at least.

I'm really kind of ignorant as to what logic we put here. In the `provides` cases we already have, it sounds like most of this stuff *shouldn't* be inherited? Is that right?

coderanger:
When I talk about subclasses I mean people subclassing core providers, not anything in Poise. Like wanting to customize the behavior of your service provider, so you subclass it and tweak some methods. I think it is rare enough (the majority of those cases would be applying the provider manually thus making all the lookup plumbing moot) that I'm behind simplifying the code, but we should figure out something for the long term to make it not required to copy-pasta the whole `provides()` call in a subclass. Maybe something like a method you can call at the class level which says "I provide the same things as my base class, make it so"? I've been [experimenting with that for resources recently](https://github.com/poise/poise/blob/master/lib/poise/helpers/resource_subclass.rb), but a lot of same ideas apply.

lamont-granquist:
paging @chef/client-core and @thommay in particular into this convo

tyler-ball:
I'm partial to removing `provides?` and adding some kind of `protected` method that subclasses can call to copy the `provides` behavior from the parent (AKA, @coderanger's suggestion).

If we do add this ability, the subclass would still need to be able to modify the `provides` logic in case it needs to add or remove specifications.

lamont-granquist:
For now can we merge this?  The existing code allowing half-inheritance of `provides?` but not `provides` isn't offering anything useful.

lamont-granquist:
(BTW @tyler-ball `protected` doesn't do what you think it does -- you probably meant a `private` method--which is still accessible to subclasses in ruby...   the utility of `protected` over `private` is very narrow in ruby...)

coderanger:
:+1: on merging this as long as it isn't used as future justification to remove `#provides?` itself :)

tyler-ball:
:+1:

lamont-granquist:
Yep, will keep `#provides?` just won't be using it here.

PR #3660, opened 2015-07-10
Add ability to declare methods inside the action class

jkeiser:


coderanger:
Kind of in the middle on this as a feature. Docs should definitely come with a warning that if you need it for more than a few small methods, you should probably stop using the fused mode.

coderanger:
Also why remove the word "provider" from all of it?

jkeiser:
I'm OK with that view of things, actually.

@coderanger regarding the removal of provider, it's just too many words, and quite frankly, the word provider isn't adding anything for the user, especially if we require you to use the resource DSL to manipulate the thing.

jkeiser:
@coderanger I'm planning to modify this just to change names (because declare_action_class makes a bit more sense than new_, IMO).

lamont-granquist:
Yeah, I definitely don't like `new_action_provider_class`.  Honestly, I'd go with `provider_class` or `action_class` and keep it tighter.  I'm not sure the word `declare` is helping me much there either. 

@jkeiser looks like we need to get this in before shipping 12.5.0

jkeiser:
@lamont-granquist why is this necessary for 12.5.0?

jkeiser:
@lamont-granquist as far as `declare_`, the reason we have *both* a `declare_action_class` and an `action_class` is that `action_class` will return `nil` if you are not using a new-style resource. I didn't want to muck up existing resources with a false, unused provider.  I'm OK with a different word than `declare` though--perhaps `ensure`?

lamont-granquist:
what is the use case for using this when you're not using a new style provider?  if you've got an old style provider shouldn't you be writing this code in the provider?

jkeiser:
@lamont-granquist two use cases:

1. Asking for `action_class` lets you find out whether the user is using a new-style resource.
2. `action_class` is used as part of resolution (specifically, [when you ask for `provider`](https://github.com/chef/chef/blob/master/lib/chef/resource.rb#L711)), and I'd like there to be *zero* chance you get an empty `action_class` in an old-style provider--I'd rather resolution fail properly instead.

lamont-granquist:
So if we let action_class take a block and evaluate it in the context of the action_class, the only thing that declare_action_class offers over action_class is that if you use declare_action_class to write a provider method in a new-style-resource, but with an old-style-provider that it'll still work and won't blow up because of action_class is nil.

I'd say do not support *that* use case, and if action_class is nil and its passed a block, just throw an exception.  The user can move the provider into the resource, or move the method they're writing into the provider.

jkeiser:
Hmm, I'm convinced we can live without it, good point. OK--so we keep `declare_action_class`, but basically all this does is change the name of stuff. Yes?

lamont-granquist:
Cool, I think I'm better with this :+1:

PR #3663, opened 2015-07-13
Fix error message for providers without `provides`

### LINKED ISSUE #3614: I don't understand warning about resource_name in resource_matching_provider()
I'm getting the warning from [`resource_matching_provider()`](https://github.com/chef/chef/blob/dbb2e0c36d2585dc10eafd3c73f0af09ff0ff52a/lib/chef/platform/provider_mapping.rb#L198-L211) in Chef 12.4.0 and I don't understand it.

It says:

```
WARN: Class Chef::Provider::DiamondCollector does not declare 'resource_name :diamond
_collector'.
WARN: This will no longer work in Chef 13: you must use 'resource_name' to provide DSL.
```

But if I add `resource_name :diamond_collector` to `Chef::Provider::DiamondCollector` I get a `NoMethodError`.

I can add `resource_name :diamond_collector` to `Chef::Resource::DiamondCollector` and have, as per the [12.4.0 release notes](https://www.chef.io/blog/2015/06/24/chef-client-12-4-0-released/).  Specifically, I use this in the resource:

``` ruby
      if respond_to?(:resource_name)
        resource_name :diamond_collector
      else
        provides :diamond_collector
      end
```

For reference, here is the source of `resource_matching_provider()`:

``` ruby
        def resource_matching_provider(platform, version, resource_type)
          if resource_type.kind_of?(Chef::Resource)
            class_name = resource_type.class.name ? resource_type.class.name.split('::').last : 
              convert_to_class_name(resource_type.resource_name.to_s)

            begin
              result = Chef::Provider.const_get(class_name)
              Chef::Log.warn("Class Chef::Provider::#{class_name} does not declare 'resource_name #{convert_to_snake_case(class_name).to_sym.inspect}'.")
              Chef::Log.warn("This will no longer work in Chef 13: you must use 'resource_name' to provide DSL.")
            rescue NameError
            end
          end
          result
        end
```

I don't understand the `rescue NameError`... is it expecting a problem with `Chef::Provider.const_get(class_name)`? If so, why not check for existence instead of catching an error.  Error catching is slow and it makes the code confusing to read.

docwhat:
The warning incorrectly said that the provider was missing
`resource_name :resource` when it should have said it was
missing `provides :resource`

I also switched from using `begin`...`rescue` since it isn't needed and
needlessly slow things up.

I think I got the logic right, but would appreciate a careful review.

Fixes #3614

### DIFF
--- a/lib/chef/platform/provider_mapping.rb
+++ b/lib/chef/platform/provider_mapping.rb
@@ -200,14 +200,13 @@ class Chef
             class_name = resource_type.class.name ? resource_type.class.name.split('::').last :
               convert_to_class_name(resource_type.resource_name.to_s)
 
-            begin
-              result = Chef::Provider.const_get(class_name)
-              Chef::Log.warn("Class Chef::Provider::#{class_name} does not declare 'resource_name #{convert_to_snake_case(class_name).to_sym.inspect}'.")
-              Chef::Log.warn("This will no longer work in Chef 13: you must use 'resource_name' to provide DSL.")
-            rescue NameError
+            if Chef::Provider.const_defined?(class_name)
+              Chef::Log.warn("Class Chef::Provider::#{class_name} does not declare 'provides #{convert_to_snake_case(class_name).to_sym.inspect}'.")
+              Chef::Log.warn("This will no longer work in Chef 13: you must use 'provides' to use the resource's DSL.")
+              return Chef::Provider.const_get(class_name)
             end
           end
-          result
+          nil
         end
 
     end
--- a/spec/integration/recipes/recipe_dsl_spec.rb
+++ b/spec/integration/recipes/recipe_dsl_spec.rb
@@ -119,7 +119,7 @@ describe "Recipe DSL methods" do
             recipe = converge {
               backcompat_thingy 'blah' do; end
             }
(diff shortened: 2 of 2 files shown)


lamont-granquist:
i think you need `const_defined?` instead of `const_defined`... ?

docwhat:
Whups. Thanks.  Sorry I took so long to reply.

lamont-granquist:
spec tests still annoyed.

docwhat:
Thanks. Should be fixed.  I'm annoyed the tests don't tell me it failed. :-/

lamont-granquist:
@chef/client-core @chef/client-maintainers can we get some eyeballs on this and someone please whack merge ASAP?

lamont-granquist:
:+1:

docwhat:
I still feel weird about this. I feel like I've reverse the meaning or something.

tyler-ball:
:+1:

tyler-ball:
@docwhat What feels weird?  I think you correctly made it so `Provider` classes say they need a `provides` message

docwhat:
Because it talked about resources before. Is there another warning about resources missing?

tyler-ball:
@docwhat Yes, there is a different message when resources are missing their required methods (but I think this is more rare because it defaults to the class name)

PR #3664, opened 2015-07-13
Allow resource actions to resolve lazy blocks - Issue #3615

### LINKED ISSUE #3615: lazy doesn't work with 'action' attribute
This recipe code doesn't work:

``` ruby
directory('/tmp/foobar') do
  action lazy { [:create, :delete].sample }
end
```

The code in the `{}` is just random code for the example. I know it is useless.

I expected it to work, but instead I get:

```
NoMethodError: undefined method `to_sym' for #<Chef::DelayedEvaluator:0x00000004b1b5d0>
```

sempervictus:
Resource actions cannot resolve lazy blocks as they pack the passed
arg into an array and call :to_sym on members. Check to see if
the arg :responds_to?(:call) and resolve the block before Array
instantiation

Fixes #3615 (on the 12.4 codebase)

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -173,6 +173,7 @@ class Chef
     #
     def action(arg=nil)
       if arg
+        arg = arg.call if arg.respond_to?(:call)
         arg = Array(arg).map(&:to_sym)
         arg.each do |action|
           validate(


chef-supermarket:

coderanger:
:-1: This is not the way we should fix this. Probably best to leave this until after some of the dust around the Properties merge settles and then make `action` a straight up property. Regardless, this doesn't match the normal `lazy {}` semantics which check for `is_a?(Chef::DelayedEvaluator)` to allow passing non-lazy procs.

sempervictus:
Agreed on making action an actual property, both master and 12.4 are rather hackish in this matter.
Far as the lazy semantics go - this mechanism provides for block resolution directly prior to the attribute actually being required in the :to_sym call while allowing other blocks to :call it first if they catch it in their execution context which would resolve the block prior to passing it into action/self.action. Where did i go south in my process evaluation?

I've got the issue resolved in our code anyway via runtime override of the current behavior, so i'll close this out since its simple enough for you guys to implement without my hash in the commit log and the associated legal rigmarole with this "CLA" bit they have going (which is not a requirement of an apache licensed project and appears nothing more than awful public relations decision making as it deprives the project of mountains of otherwise easily contributed work).

PR #3668, opened 2015-07-14
Set chef_environment in attributes JSON

mcquin:
Allows users to have `"chef_environment": "some_environment"` in their JSON attributes file (e.g., `/etc/chef/first_boot.json`). Setting `"chef_environment"` in the attributes JSON has the same effect as running `chef-client -E some_environment`.

Note [this comment on CLI option precedence](https://github.com/chef/chef/blob/64e0a1da0826f5befababb5693d8bb5b8de913e9/lib/chef/node.rb#L351-L360). I've struggled to come up with a reasonable workaround which doesn't involve completely reworking the interaction between`Chef::Config` and CLI `config`.

\cc @chef/client-core @charlesjohnson @alexpop

### DIFF
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -6,3 +6,19 @@ Example Doc Change:
 Description of the required change.
 -->
 
+### chef-client -j JSON
+Add to the [description of chef-client options](https://docs.chef.io/ctl_chef_client.html#options):
+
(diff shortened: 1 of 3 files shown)


coderanger:
:+1: LGTM. Even with the order being annoying, I think this still hits the use case @charlesjohnson was talking about.

tyler-ball:
:+1:

PR #3669, opened 2015-07-15
Switch notifications from Hipchat to Slack (#open-source channel)

juliandunn:
I'm assuming the token is global to our org, so it's not tied to me... but if someone finds out otherwise please let me know.


lamont-granquist:
can we get this changed to `#chef-core-eng-team`, the `#open-source` channel no longer exists.

lamont-granquist:
pingles?

lamont-granquist:
also we probably want to send this to the same room we're going to be sending the build notifications to, not to the main chef-core-eng-team channel since that'll create too much noise, but i'm not sure if we have that room setup yet.

juliandunn:
I'm going to close this in favor of having the right people (i.e. @chef/engineering-services) generate the right token.

PR #3672, opened 2015-07-15
Fix issue where DSL is not emitted if state_properties happens before property

jkeiser:
If you call state_properties :x, and there is no
property :x yet, it assumes you are working with a custom
getter/setter like "def x". This is fine, but when you say
"property :x", it won't override that (and will treat the
property as if the getter/setter were already defined instead
of creating a new one).


lamont-granquist:
:+1:

PR #3676, opened 2015-07-16
pry-remote for consistent remote diagnostics

pburkholder:
https://chefio.slack.com/archives/test-kitchen/p1436995105000034

@fnichol " i can think of one tweak that would help us out--adding pry-remote to our chef omnibus packages. that would mean we have all the software available and making some native pair of chef resources totally doable. having this in chef core could be a powerful help all over the place"

### DIFF
--- a/chef.gemspec
+++ b/chef.gemspec
@@ -33,6 +33,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "chef-zero", "~> 4.2", ">= 4.2.2"
   s.add_dependency "pry", "~> 0.9"
+  s.add_dependency "pry-remote", "~> 0.1"
 
   s.add_dependency 'plist', '~> 3.1.0'
 


coderanger:
:-1: This shouldn't be a dependency of Chef. Pull it into the packages for chef-client or chefdk perhaps, but not the actual gem.

ranjib:
agree with noah

lamont-granquist:
Yep, agree with noah, i'd like to see this in omnibus-chef/omnibus-software.

Really we should drop the pry dep as well, we don't actually dep on pry, I don't know how that wound up in there...   That should also probably get moved to omnibus-chef/omnibus-software.

Going to close this one.

coderanger:
I think chef-shell uses pry? Or at least did at some point.

lamont-granquist:
Nope, we have talked about it lots and lots, but its still irb based.

PR #3684, opened 2015-07-19
Remove warnings about hander overrides.

coderanger:
These are high-volume, difficult to silence, and often not actually an error (in which case they may be impossible to silence forever). Worse, doing what the error says and setting override:true without understanding what that does can have unexpected effects (https://github.com/jssjr/git/issues/70). If we want to keep them, they should be moved to debug, but as they often don't indicate any error, I think they should be removed.

Opening this for discussion. /cc @jkeiser @lamont-granquist @someara

### DIFF
--- a/lib/chef/platform/handler_map.rb
+++ b/lib/chef/platform/handler_map.rb
@@ -31,11 +31,6 @@ class Chef
           # are exactly equal
           if new_matcher[:value].is_a?(Class) && !new_matcher[:override]
             cmp = compare_matcher_properties(new_matcher, matcher) { |m| m[:value].name }
-            if cmp < 0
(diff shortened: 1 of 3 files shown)


thommay:
:+1:

someara:
:+1:

jkeiser:
:+1:

coderanger:
And also from IRC, should follow up when @lamont-granquist is back and possibly re-add a warning in the resource/provider lookup code where we can check if there was actually ambiguity.

lamont-granquist:
Yeah we really still need the ambiguity check in the provider resolver compile-time check to see if we really encountered ambiguity.  Moving it to class loading time didn't make sense though since we didn't have enough information at that point to make an accurate check and it threw lots of inaccurate warnings.

PR #3689, opened 2015-07-20
Add ohai configuration context to config.

mcquin:
We need a way for `Chef::Config` to pick up `Ohai::Config.ohai` configurables so that Ohai is properly configured at the start of a `chef-client` run.

Adding an empty `ohai` configuration context to `Chef::Config` means append/modify operations (e.g., `ohai.plugin_path << some_path` or `ohai.plugin_path += [ some_path_1, ..., some_path_N ]`) won't work and it's not DRY to copy default values here.

Since the new Ohai configuration uses `ChefConfig::Config`, a `require 'ohai/config'` will make the Ohai configurables available to `Chef::Config`.

Requires a release of Ohai including (at least) [this change](https://github.com/chef/ohai/pull/574).

@chef/client-core

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,11 +5,6 @@ gem "activesupport", "< 4.0.0", :group => :compat_testing, :platform => "ruby"
 
 gem 'chef-config', path: "chef-config"
 
-# REMOVEME once there is a release of Ohai with these changes, and the
-# chef.gemspec is updated.
-gem 'ohai', github: 'chef/ohai',
-            ref: '7556a1d55808c459f3a9fad88e2a2371f361f3e0'
-
 group(:docgen) do
   gem "tomlrb"
   gem "yard"
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,11 @@ gem "activesupport", "< 4.0.0", :group => :compat_testing, :platform => "ruby"
 
 gem 'chef-config', path: "chef-config"
 
+# REMOVEME once there is a release of Ohai with these changes, and the
+# chef.gemspec is updated.
+gem 'ohai', github: 'chef/ohai',
+            ref: '7556a1d55808c459f3a9fad88e2a2371f361f3e0'
+
 group(:docgen) do
   gem "tomlrb"
   gem "yard"
--- a/chef.gemspec
+++ b/chef.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency "mixlib-log", "~> 1.3"
   s.add_dependency "mixlib-authentication", "~> 1.3"
   s.add_dependency "mixlib-shellout", ">= 2.0.0.rc.0", "< 3.0"
-  s.add_dependency "ohai", "~> 8.0"
+  # s.add_dependency "ohai", "~> 8.0"
 
   s.add_dependency "ffi-yajl", "~> 2.2"
   s.add_dependency "net-ssh", "~> 2.6"
--- a/chef.gemspec
+++ b/chef.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency "mixlib-log", "~> 1.3"
   s.add_dependency "mixlib-authentication", "~> 1.3"
   s.add_dependency "mixlib-shellout", ">= 2.0.0.rc.0", "< 3.0"
-  # s.add_dependency "ohai", "~> 8.0"
+  s.add_dependency "ohai", ">= 8.6.0.alpha.1", "< 9"
 
   s.add_dependency "ffi-yajl", "~> 2.2"
   s.add_dependency "net-ssh", "~> 2.6"
--- a/lib/chef/config.rb
+++ b/lib/chef/config.rb
@@ -30,8 +30,17 @@ ChefConfig.logger = Chef::Log
 require 'chef-config/config'
 require 'chef/platform/query_helpers'
 
+# Ohai::Config defines its own log_level and log_location. When loaded, it will
+# override the default ChefConfig::Config values. We save them here before
+# loading ohai/config so that we can override them again inside Chef::Config.
+#
+# REMOVEME once these configurables are removed from the top level of Ohai.
+LOG_LEVEL = ChefConfig::Config[:log_level] unless defined? LOG_LEVEL
+LOG_LOCATION = ChefConfig::Config[:log_location] unless defined? LOG_LOCATION
+
 # Load the ohai config into the chef config. We can't have an empty ohai
-# configuration context because `ohai.plugins_path << some_path` won't work.
+# configuration context because `ohai.plugins_path << some_path` won't work,
+# and providing default ohai config values here isn't DRY.
 require 'ohai/config'
 
 class Chef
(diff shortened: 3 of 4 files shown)


ranjib:
:+1:

thommay:
LGTM - go ahead and merge once the ohai change is merged!

mcquin:
@ranjib @thommay I found out that Ohai::Config's log_level and log_location were overriding the default log_level and log_location Chef::Config uses. [This commit](https://github.com/chef/chef/commit/567457cf110d114229857bc6bcf06a0fd0ec0cae) is a clumsy workaround.

mcquin:
Dependencies on git repositories got me down. I think we'll need to stage some kind of dual chef-config & ohai release.

adamedx:
@solverdude, this should allow users to get around https://github.com/chef/ohai/issues/420 -- feel free to ask for specifics on how it works in that use case.

Gemfile
```diff
@@ -5,6 +5,11 @@ gem "activesupport", "< 4.0.0", :group => :compat_testing, :platform => "ruby"
 
 gem 'chef-config', path: "chef-config"
 
+# REMOVEME once there is a release of Ohai with these changes, and the
+# chef.gemspec is updated.
+gem 'ohai', github: 'chef/ohai',
```
ksubrama: Just for my own sake, I'm marking locations that would need to be changed before commit.

chef.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency "mixlib-log", "~> 1.3"
   s.add_dependency "mixlib-authentication", "~> 1.3"
   s.add_dependency "mixlib-shellout", ">= 2.0.0.rc.0", "< 3.0"
-  s.add_dependency "ohai", "~> 8.0"
+  # s.add_dependency "ohai", "~> 8.0"
```
ksubrama: This would need to be re-enabled when the ohai line in Gemfile is removed right?

lib/chef/config.rb
```diff
@@ -49,5 +61,24 @@ class Config
...
+    default :log_level, LOG_LEVEL
+    default :log_location, LOG_LOCATION
+
+    # Ohai::Config[:log_level] is deprecated and warns when set. Unfortunately,
+    # there is no way to distinguish between setting log_level and setting
+    # Ohai::Config[:log_level]. Since log_level and log_location are used by
+    # chef-client and other tools (e.g., knife), we will mute the warnings here
+    # by redefining the config_attr_writer to not warn for these options.
+    #
+    # REMOVEME once the warnings for these configurables are removed from Ohai.
```
ksubrama: This seems similar to the issue dealt with in https://github.com/chef/chef/blob/master/lib/chef/knife/data_bag_secret_options.rb - am I thinking about this correctly?

mcquin: In the other case, it was that there is no way to distinguish between `--secret SECRET` on the CLI and `knife[:secret] = SECRET` in knife.rb after configs have been merged. The CLI secret option was redefined to write to a different key so `Chef::Config.knife[:secret]` could map to `knife[:secret]` and `Chef::Config.knife[:cl_secret]` could map to `--secret`.

In this case, the trouble is with multiple definitions of `default :log_level, LEVEL`. The ohai config file is the last writer of this configuration option, and therefore the winner. I thought about naming the configurable `:ohai_log_level` in `Ohai::Config` and redefining the `get` method for `:log_level` so that it will return `:ohai_log_level` if `log_level` isn't set in the config.rb. However, this method will get applied to `Chef::Config` too, since the `::Config` part of `Ohai::Config` and `Chef::Config` is really just `ChefConfig::Config`. Then we're back in essentially the same boat.

chef.gemspec
mcquin: Yes, and it would have to be pinned to a minimum version of ohai which includes these changes (probably 8.5?)

ksubrama:
Talked with Claire.  I'm :+1: on this.

ranjib:
for any shared config parameter between ohai & chef-client, ohai will win... i like this as it is inline with how attributes work :-) . if we white list specific attributes (like log_location, log_level), , we'll have to remember checking this for any new config params added in chef or ohai,,

im still :+1:  on this, :-p, cause it is better than existing

mcquin:
Last change: use ohai gem 8.6.0.alpha.0 instead of ohai from github.

smurawski:
:+1:  :shipit:

mcquin:
At last! I've figured out why the kitchen-tests are failing.

At the beginning of a chef-client run, an `Ohai::System` class is instantiated. During its initialization, deprecated ohai configuration settings /which have been set/ are merged into the ohai config context ([here](https://github.com/chef/ohai/blob/master/lib/ohai/config.rb#L53)). Configuration defaults aren't returned by the `configuration` getter method, only configurables which have been assigned a value.

Somewhere -- either in chef or test kitchen, I haven't narrowed it down -- `Chef::Config[:log_level]` is set to `:auto`. When the deprecated ohai config is merged into the ohai config context, `log_level: auto` appears in the hash from `#configuration`, causing `Ohai::Config.ohai[:log_level]` to also be set to `:auto`. This is causing an argument error to be raised when the ohai system configures its logging ([here](https://github.com/chef/ohai/blob/master/lib/ohai/system.rb#L218)).

Until the top-level ohai config is completely deprecated, the only workaround I can think of is to change `Ohai::Config` to purely use `Mixlib::Config` (drop `Config = ChefConfig::Config` part so that the ohai config doesn't get "auto-loaded" into the chef config), and load the ohai configuration context into `Chef::Config`. Something like `self.configuration_contexts[:ohai] = Ohai::Config.ohai` should do the trick. Unfortunately, that also means we have to refactor `ChefConfig::WorkstationConfigLoader` or copypasta/`:send` into it so we can avoid the parts of `#load` and other helper methods which call out to `ChefConfig::Config`.

---

I will be offline for the next week, so this was partly for myself to help me remember what I found out today and partly to provide an update on this PR and what's up with that last commit.

smurawski:
@mcquin it is it test-kitchen, in the chef-zero provisioner.  If no log level specified or if set to info it defaults to auto.

mcquin:
Waiting for https://github.com/chef/ohai/pull/593 to be merged and a new ohai alpha gem release.

SolverDude:
@adamex I just looked in the Ohai hostname plugin within the 12.5.1 chef-client, and I don't find any clues as to how I would configure it to wire in the old-school nis domain fqdn which you provided us via patch for chef/ohai#420.  So, if indeed that's in the 12.5.1 client, I DO need the assistance you offered (back in July) on how we would configure this.  Otherwise, something else is up (or down, depending on your point of view).

btm:
@SolverDude @adamedx We now need to go back and finish @lamont-granquist's [draft RFC for ohai identify attributes](https://github.com/chef/chef-rfc/pull/108). The new Ohai config here and in related PRs implemented [RFC53](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md) which came out of that draft RFC. Once we complete that "once and for all" we will have a convenient configuration attribute to solve https://github.com/chef/ohai/issues/420 for users, even if the default behavior isn't perfect for them.

SolverDude:
OK.  Thanks for providing the clarification @btm.  We can continue to patch the hostname.rb plugin for the moment (which otherwise breaks our use of Chef on Solaris).

@adamedx, can you provide a timeline for when we would see this Ohai configuration could be  realized, and we can configure to use NIS style fqdn configuration, and we could stop patching the chef-client?

adamedx:
@btm, can you suggest something we can do now so @SolverDude can move forward? It's good to see things move through the RFC process, but I feel there's enough there in Ohai / Chef Client where there should be a solution on top of those that could potentially be deployed via tricky methods like client.rb. If you don't have idea, let's get a f2f with @SolverDude can get something repeatable and reliable worked out.

lamont-granquist:
It should be reasonably easy to get the draft through RFC now and get the code written and done.

Now that the ohai config stuff has landed in ohai just need to take the comment that I made in there and fix it for the actual syntax we implemented and then rework the RFC around that text.  Then the code is mostly just a bunch of `if` statements to implement the different options.

PR #3691, opened 2015-07-21
Add Resource.load_current_value

jkeiser:
Per [RFC](https://github.com/chef/chef-rfc/pull/127).

An example:

```ruby
class MyFile < Chef::Resource
  property :path, String, name_property: true
  property :content, String

  load_current_value do
    value_does_not_exist! if !::File.exist?(path)
    content IO.read(path)
  end

  action :create do
    if !current_resource
      converge_by "Create #{path} with content #{content}" do
        IO.write(path, content)
      end
    elsif content != current_resource.content
      converge_by "Update #{path} with content #{content}" do
        IO.write(path, content)
      end
    end
  end
end
```

`current_resource` on the provider will be `nil` if the file does not exist, and will be a `MyFile` resource with `path` and `content` filled in if it does exist. Additionally, it fixes a subtle and easy-to-have issue with simple code like above, which happens when the user doesn't specify `content` (like, say they specify `mode` instead, and don't bother with `content`):

```ruby
my_file '/x.txt' do
end
```

Before this change, `content` would return `nil` because the user did not specify it, and the file would be truncated! Now, `content` defaults to the *current* value if one exists, so `content` will be the current text in the file, and it won't overwrite anything . This makes the resource *patchy*--"if I don't specify something, leave it alone"--with minimal work on the resource writer's part.

### DIFF
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -22,3 +22,271 @@ will set the node's environment to `"pre-production"`.
 > *Note that the environment specified by `chef_environment` in your JSON will
 take precedence over an environment specified by `-E ENVIROMENT` when both options
 are provided.*
+
+### Resources Made Easy
+
+Resources are central to Chef. The system is extensible so that you can write
+your own reusable resources, and use them in your recipes, and even publish them
+so that others can use them too!
+
+However, writing these resources has not been as easy as we would have liked. In
+Chef 12.5, we are fixing this with a large number of DSL improvements designed
+to reduce the number of things you need to type and think about when you create
+a resource. Resources should be your go-to solution for many Chef problems, and
+these changes make them easy enough to dash off in an instant, while retaining
+all the power you're accustomed to.
+
+The process to create a resource is now:
+
+1. Make a resource file in your cookbook (like `resources/my_resource.rb`).
+2. Add the recipes defining your actions using the `action :create <recipe>` DSL.
+3. Add properties so the user can tweak some knobs on your resource (like paths,
+   or preferences), using the `property :my_property, <type>, <options>` DSL.
+4. Use the resource in your recipe!
+
+There are other things you can do, but this is the most basic (and the first)
+thing you will start with.
+
+Let's demonstrate the new features by taking a simple recipe from the awesome
+[learnchef tutorial](https://learn.chef.io/learn-the-basics/rhel/configure-a-package-and-service/),
+and turning it into a reusable resource:
+
+```ruby
+package 'httpd'
+
+service 'httpd' do
+  action [:enable, :start]
+end
+
+file '/var/www/html/index.html' do
+  content '<html>
+  <body>
+    <h1>hello world</h1>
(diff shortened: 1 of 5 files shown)


lib/chef/provider.rb
```diff
@@ -211,10 +211,29 @@ def self.include_resource_dsl_module(resource)
           extend Forwardable
           define_singleton_method(:to_s) { "#{resource_class} forwarder module" }
           define_singleton_method(:inspect) { to_s }
+          # Add a delegator for each explicit property that will get the *current* value
+          # of the property by default instead of the *actual* value.
+          resource.class.properties.each do |name, property|
+            class_eval(<<-EOM, __FILE__, __LINE__)
+              def #{name}(*args, &block)
```
jaym: what happens if i have a property named `:'foo-bar'`

lib/chef/resource.rb
```diff
@@ -1414,8 +1442,31 @@ def self.new_action_provider_class
...
+            # We keep identity properties and non-desired-state, which are assumed to be
+            # "control" values like `recurse: true`
+            current_resource.class.properties.each do |name,property|
+              if property.desired_state? && !property.identity? && !property.name_property?
+                property.reset(current_resource)
+              end
+            end
+
+            if current_resource.method(:load_current_value!).arity > 0
+              current_resource.load_current_value!(new_resource)
```
jaym: I don't remember seeing this in the doc changes. What are the cases when you'd want to have this explicitly passed?

DOC_CHANGES.md
```diff
@@ -22,3 +22,271 @@ will set the node's environment to `"pre-production"`.
...
+```
+
+It won't do anything yet, though :)
+
+#### Declaring an Action
+
+Let's make our resource do something. To start with, we'll just have it do exactly
+what the learnchef tutorial does, but in the resource.  Put this in
+`resources/single_page_website.rb`:
+
```
jaym: is specifying `provides` not required?

spec/integration/recipes/resource_load_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+
+        it "i, name and d are passed to load_current_value, but not x" do
+          expect(resource.current_resource.x).to eq 'loaded 2 (d=desired_d, i=desired_i, name=blah)'
+        end
+      end
+    end
+
+    context "and a resource with no values set" do
+      let(:resource) do
+        e = self
```
jaym: why does this need to happen?

jaym:
:+1: 
cc @jamescott to look at the doc changes

DOC_CHANGES.md
jkeiser: @jaym correct.

lib/chef/provider.rb
jkeiser: Yeah, I can add a rescue to fall back on define_method in that case. We do that in the `property` declaration spot.

Issue #3232, opened 2015-04-16
template verify variable name mismatch


Tech356:
[RFC 27](https://github.com/chef/chef-rfc/blob/master/rfc027-file-content-verification.md) states

    The path to the temporary file with the proposed content will be available
    by using Ruby's sprinf formatting:

    "%{path}"

And the [documentation](https://docs.chef.io/resource_template.html) show an example

    template "/etc/nginx.conf" do
      verify "nginx -t -c %{path}"
    end

But the [code](https://github.com/chef/chef/blob/master/lib/chef/resource/file/verification.rb#L109) gives `file` instead of `path`

    def verify_command(path, opts)
        command = @command % {:file => path}
        interpreter = Chef::GuardInterpreter.for_resource(@parent_resource, command, @command_opts)
        interpreter.evaluate
    end

stevendanna:
Thanks @Tech356.  Let's add :path and keep :file to avoid breaking anyone else who managed to figure out the right option.  Sorry about this.

PR #3693, opened 2015-07-21
Interpolate `%{path}` in verify command

margueritepd:
See chef/chef#3232

### DIFF
--- a/lib/chef/resource/file/verification.rb
+++ b/lib/chef/resource/file/verification.rb
@@ -108,6 +108,10 @@ class Chef
         def verify_command(path, opts)
           # First implementation interpolated `file`; docs & RFC claim `path`
           # is interpolated. Until `file` can be deprecated, interpolate both.
+          Chef::Log.deprecation(
+            '%{file} is deprecated in verify command and will not be '\
+            'supported in Chef 13. Please use %{path} instead.'
+          ) if @command.include?('%{file}')
           command = @command % {:file => path, :path => path}
           interpreter = Chef::GuardInterpreter.for_resource(@parent_resource, command, @command_opts)
           interpreter.evaluate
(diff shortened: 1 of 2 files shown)


coderanger:
Can you add a comment indicating why there are two? Maybe that would give us a mild chance of remembering to remove the old one for Chef 13.

ranjib:
@coderanger RFC says the variable name should be `path`. current implementation uses `file` . Keeping file as it is will ensure nothing breaks downstream

stevendanna:
@ranjib I believe @coderanger was asking for a comment in the code for people who come across it later.  I'm :+1: on having a code comment about why we have two.

coderanger:
Yes, we can't remove `file` until Chef 13, but we should eventually revisit to clean this up.

ranjib:
@margueritepd that helper method name should be something different, in itself its not a test. may be platform_specific_command ?
@coderanger i cant think of a way to throw deprecation warning for file usage, else it would have helped..
:+1:

ranjib:
@stevendanna ah .. got it now

coderanger:
```ruby
Chef::Log.deprecate('%{file} is deprecated in favor of %{path}') if @command.include?('%{file}')
```

or something like that.

margueritepd:
@coderanger comment and deprecation warning added, tests updated :)

I didn't use `Chef::Log.deprecation` because that throws an error apparently.

mcquin:
@margueritepd there is a [configuration option](https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L518-L522) for treating [deprecation warnings as errors](https://github.com/chef/chef/blob/master/lib/chef/log.rb#L41). It's best to use `Ohai::Log.deprecation`.

margueritepd:
@mcquin back to using `deprecation`!

mcquin:
:+1:

PR #3202, opened 2015-04-06
Resolves ticket 2989 (previously CHEF-4446): Version numbers are now dis...

kungfujoe:
...played when synching cookbooks.

PR #3700, opened 2015-07-22
Make the doc formatter actually show what version of a cookbook is being used.

coderanger:
This has bugged me forever. It does change the API of the Formatter system in a ~non-back-compat way but I don't think I've actually seen any formatters outside of core Chef so how much do we want to worry about this? We could write a shim in `Chef::EventDispatcher::Dispatcher` to check the arity of the method if needed.

Ping @chef/client-core.

### DIFF
--- a/lib/chef/cookbook/synchronizer.rb
+++ b/lib/chef/cookbook/synchronizer.rb
@@ -131,7 +131,7 @@ class Chef
       files_remaining_by_cookbook[file.cookbook] -= 1
 
       if files_remaining_by_cookbook[file.cookbook] == 0
-        @events.synchronized_cookbook(file.cookbook.name)
+        @events.synchronized_cookbook(file.cookbook)
       end
     end
 
--- a/lib/chef/cookbook/synchronizer.rb
+++ b/lib/chef/cookbook/synchronizer.rb
@@ -131,7 +131,7 @@ class Chef
       files_remaining_by_cookbook[file.cookbook] -= 1
 
       if files_remaining_by_cookbook[file.cookbook] == 0
-        @events.synchronized_cookbook(file.cookbook)
+        @events.synchronized_cookbook(file.cookbook.name, file.cookbook)
       end
     end
 
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -119,7 +119,7 @@ class Chef
       end
 
       # Called when cookbook +cookbook+ has been sync'd
-      def synchronized_cookbook(cookbook)
+      def synchronized_cookbook(cookbook_name, cookbook)
       end
 
       # Called when an individual file in a cookbook has been updated
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -118,8 +118,8 @@ class Chef
       def cookbook_sync_start(cookbook_count)
       end
 
-      # Called when cookbook +cookbook_name+ has been sync'd
-      def synchronized_cookbook(cookbook_name)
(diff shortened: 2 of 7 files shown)


tyler-ball:
I'm :+1: on this change.  Can you update https://github.com/chef/chef/blob/master/spec/unit/formatters/doc_spec.rb and other affected files with tests to cover this new scenario?

danielsdeleo:
I coulda swore @ranjib added some new API to the event system that does this, but maybe it's in a different place. Also, for policyfiles, we are able to display both a fragment of the SHA and the version, but it might be only because we have the full Policyfile on hand at the time?

There's at least the nyan cat formatter outside of core and we can never be sure what people have in private repos, so I'd like to be extra sure we don't have a different means of accomplishing this before we accept a breaking change. Definitely :+1: on this information being in the output though.

lamont-granquist:
competes with #3202 ?

coderanger:
@lamont-granquist Yeah, that is a backwards-compatible version. Still not sure how much we care since searching GitHub finds only three formatters, one of which is abandoned (and would still work anyway), one is Nyan (which would work fine due to subclassing), and one that would break https://github.com/sakazuki/chef-formatters-simple.

lamont-granquist:
I'm okay with it, but I'm a bit of a SemVer anarchist.

coderanger:
I'm normally the compat stickler so I'm totally okay with calling this a disallowed breakage :)

lamont-granquist:
You could take a look at cleaning up #3202 at the very least it needs tests, and I'm not super wild about the approach it takes but I can't think of anything better offhand.  A nice pattern for being able to extend the formatters and add an argument without a breaking change would be nice.

coderanger:
So a generalized solution might be to make new arguments always be at the end of the argspec, and make the EventDispatch code trim based on arity. I'll take a stab at that tonight.

coderanger:
Okay, I updated this to have a more general solution for extending formatters without breaking compat, and then used that to add the new argument. Thoughts?

lamont-granquist:
That looks a lot better.  What happens in the case where someone writes a formatter that has too many args for the caller?

lamont-granquist:
(It looks like `#take` will fall back to all the args and then ruby should blow up appropriately?)

coderanger:
Yeah, `[1,2,3].take(4) == [1,2,3]`.

lamont-granquist:
So this adds a slight bit of magic and now when someone subclasses this and writes a formatter that only takes 1 arg (accidentally) instead of, say, 3 on any method, then it won't give them an error on that.  That might actually be a feature though.  And that's so overwhelmed by the positives of now being able to extend the method signatures of the loggers without breaking changes, that i think i'm like 1000% :+1:

coderanger:
Yeah, we could consider adding a warning when `arity < args.length` but chances are that will mostly spam people that can't fix it. I wish we had a better way to do "only show the first time" deprecation warnings, since that would fit nicely here.

danielsdeleo:
Yeah, I wish I woulda gone through and made Structs for all the arguments to these things for 12.0 like I intended, then you could easily add new things to any event, but it's too late. There's backwards compatible ways to make that happen if anyone has the cycles for it.

Anyway, :+1: on this patch :100:

juliandunn:
@coderanger this seems important enough to call out explicitly in `CHANGELOG.md`?

coderanger:
No idea on that front, I mostly care for helping people debug version solver issues.

stevendanna:
:fireworks: I think there have been 3-4 versions of this feature now.  So happy one finally got merged.  Thanks @coderanger

PR #3704, opened 2015-07-23
Add support for OS X 10.11 SIP paths

natewalck:
Currently, Chef will fail to create /usr/local as /usr is no longer writable on OS X 10.11.

To work around this, If a path cannot be written, we do the following:

* Check to see if the path is writable. If is, continue as we normally would.
* Check to see if we are on 10.11 and if it is a SIP path.
* If it is 10.11 and it is a SIP path, check to see if it is on the exception list. If so, return true.
* If it is not 10.11 or a SIP path, return false as you normally would.

I'm sure this could use some tweaking, but I wanted to get the PR up so we could start discussions on it. Without this functionality, Chef will simply break on 10.11 while trying to write /usr/local and any other folder that is an exception path for SIP.

I am with Facebook, so as soon as we are happy with this PR, @jaymzh will have to do the merge as I am not on the CLA.

### DIFF
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -228,6 +228,37 @@ module ChefConfig
         joined_paths
       end
     end
+
+    # Determine if the given path is protected by OS X System Integrity Protection.
+    def self.is_sip_path?(path, node)
+      if node['platform'] == 'mac_os_x' and Gem::Version.new(node['platform_version']) >= Gem::Version.new('10.11')
+          # todo: parse rootless.conf for this?
+          sip_paths= [
+            '/System', '/bin', '/sbin', '/usr',
+          ]
+          sip_paths.each do |sip_path|
+            ChefConfig.logger.info("This is a SIP path, checking if it in exceptions list.")
+            return true if path.start_with?(sip_path)
+          end
+          false
+      else
+        false
+      end
+    end
+    # Determine if the given path is on the exception list for OS X System Integrity Protection.
+    def self.writable_sip_path?(path)
+      # todo: parse rootless.conf for this?
+      sip_exceptions = [
+        '/System/Library/Caches', '/System/Library/Extensions',
+        '/System/Library/Speech', '/System/Library/User Template',
+        '/usr/libexec/cups', '/usr/local', '/usr/share/man'
+      ]
+      sip_exceptions.each do |exception_path|
+        return true if path.start_with?(exception_path)
+      end
+      ChefConfig.logger.error("Cannot write to a SIP Path on OS X 10.11+")
+      false
+    end
   end
 end
 
--- a/lib/chef/provider/directory.rb
+++ b/lib/chef/provider/directory.rb
@@ -64,7 +64,13 @@ class Chef
               is_parent_writable = lambda do |base_dir|
                 base_dir = ::File.dirname(base_dir)
                 if ::File.exists?(base_dir)
-                  Chef::FileAccessControl.writable?(base_dir)
+                  if Chef::FileAccessControl.writable?(base_dir)
+                    true
+                  elsif Chef::Util::PathHelper.is_sip_path?(base_dir, node)
+                    Chef::Util::PathHelper.writable_sip_path?(base_dir)
+                  else
+                    false
+                  end
                 else
                   is_parent_writable.call(base_dir)
                 end
@@ -74,7 +80,13 @@ class Chef
               # in why run mode & parent directory does not exist no permissions check is required
               # If not in why run, permissions must be valid and we rely on prior assertion that dir exists
               if !whyrun_mode? || ::File.exists?(parent_directory)
-                Chef::FileAccessControl.writable?(parent_directory)
+                if Chef::FileAccessControl.writable?(parent_directory)
+                  true
+                elsif Chef::Util::PathHelper.is_sip_path?(parent_directory, node)
+                  Chef::Util::PathHelper.writable_sip_path?(@new_resource.path)
+                else
+                  false
+                end
               else
                 true
               end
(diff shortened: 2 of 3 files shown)


chef-supermarket:

* @natewalck

natewalck:
Oh yea, if @jtimberman and @tyler-ball Could take a gander, I'd be much obliged.

jaymzh:
CC @chef/client-core and @chef/client-maintainers

chef-config/lib/chef-config/path_helper.rb
```diff
@@ -228,6 +228,29 @@ def self.all_homes(*args)
         joined_paths
       end
     end
+
+    # Determine if the given path is protected by System Integrity Protection
```
tyler-ball: `Determine if the given path is protected by OSX System Integrity Protection`

lib/chef/provider/directory.rb
```diff
@@ -64,7 +64,13 @@ def define_resource_requirements
               is_parent_writable = lambda do |base_dir|
                 base_dir = ::File.dirname(base_dir)
                 if ::File.exists?(base_dir)
-                  Chef::FileAccessControl.writable?(base_dir)
+                  if Chef::FileAccessControl.writable?(base_dir)
+                    true
+                  elsif Gem::Version.new(node['platform_version']) >= Gem::Version.new('10.11') && Chef::Util::PathHelper.is_sip_path?(base_dir)
```
tyler-ball: This should also check for `Darwin`

tyler-ball:
This should have some specs added.  Other than my minor comments, this LGTM.

I think it would be helpful to add some debug output saying `SIP path detected` so users will know why it failed.

lib/chef/provider/directory.rb
natewalck: Putting a darwin check into path_helper is pretty trivial, but the node object isn't available to path_helper as far as In can tell.

I don't see a ruby constant that shows the kernel info for the system, otherwise I could look for the kernel version and map that to 10.11+.

I'll keep looking, but if you know of a good place I could grab this info that is available within chef-config, I'll use that.

natewalck: I was going to try to get the OS  version using RbConfig, but ran into this: #3708 

Any ideas on how to get at this info? I could shell out, but that seems super wasteful in this use case.

jaymzh: You can always pass the node object into `is_sip_path?` - depending on how many call sites there are to update...

natewalck:
Once the code looks good, I'll start on the specs. Any specific specs you want to see for this code?

natewalck:
Added some specs to test the functionality that was added. Let me know if anything else should be added or if I can refactor it to make it look less fugly. @jtimberman @tyler-ball

natewalck:
Waiting on code review from @chef/client-core and @chef/client-maintainers.

jaymzh:
LOL, fail github. @chef/client-core @chef/client-maintainers

tyler-ball:
:+1:

spec/unit/provider/directory_spec.rb
```diff
@@ -197,6 +197,37 @@
         expect { directory.run_action(:create) }.to raise_error(Chef::Exceptions::EnclosingDirectoryDoesNotExist)
       end
     end
+
+    describe "on OS X", mac_osx_only: true do
```
tyler-ball: Lack of knowledge on my part - because we're mocking out the `Dir` responses below, do we need `mac_osx_only: true` here?  My thought is that since we're mocking we could run this test on all platforms.  Can someone with more core chef spec knowledge chime in here?  Is `mac_osx_only` for integration tests?

natewalck: Should I remove this because I am mocking it all out?

danielsdeleo: Yes, please. We use these tags for tests that can only pass on the given platform, i.e., the non-mocked tests in `spec/functional/`

natewalck:
Well then, taking that out seems like it has revealed some incorrectly mocked stuff.

natewalck:
Fixed the spec error. /usr/bin/test exists on Linux machines ;)

natewalck:
This test is failing saying /usr/bin does not exist. How do I work around that? I'm really keen on getting this merged.

danielsdeleo:
It's this part of the code that's getting triggered and creating the error:

```ruby
         requirements.assert(:create) do |a|
           # Make sure the parent dir exists, or else fail.
           # for why run, print a message explaining the potential error.
           parent_directory = ::File.dirname(@new_resource.path)
           a.assertion { @new_resource.recursive || ::File.directory?(parent_directory) }
           a.failure_message(Chef::Exceptions::EnclosingDirectoryDoesNotExist, "Parent directory #{parent_directory} does not exist, cannot create #{@new_resource.path}")
           a.whyrun("Assuming directory #{parent_directory} would have been created")
         end
```

If you change the base path in your examples from `/usr/bin` to a thing that doesn't exist, you should repro the error, and from there you should be able to stub `::File.directory?(parent_directory)` to solve it, which should fix the issue for all platforms.

natewalck:
Looks like it is passing now (I hope because it is properly fixed).

Let me know if we need anything else!

tyler-ball:
:+1: This should get squashed before merge.  Thanks for the PR!

jaymzh:
Just got back from DEFCON. @natewalck  if you want to squash these all and force-push I can click-button merge. Otherwise, I'm happy to cli-merge and squash myself, but you'll have to wait a few days for me to catch up on everything.

natewalck:
I'll squash and force push, it will be an hour or two though.




Nate

natewalck:
So...I did the squash but this looks quite wrong.

natewalck:
I think I fixed it. Let me know if I did it wrong ;)

natewalck:
@jaymzh @tyler-ball @jtimberman  How do I ensure this gets included in the 12.5.0 release?

jaymzh:
You or I should send a PR to add a Changelog docs for all yoru changes and I THINK they now always pull from master, so unless we missed their RC cut off, it'll just be in there.

PR #3706, opened 2015-07-23
Update version to 12.5.0.alpha.0 for gem release.

mcquin:
Releasing a 12.5.0.alpha.0 gem which includes changes to chef-config which are needed by ohai.

@chef/client-core


danielsdeleo:
I think this is good to :ship: but remember not to merge :)

chef-config/lib/chef-config/version.rb
```diff
@@ -20,6 +20,6 @@
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 module ChefConfig
-  VERSION = '12.5.0.current.0'
+  VERSION = '12.5.0.alpha.0'
```
tyler-ball: Until we pull `ChefConfig` into its own gem, why not set `VERSION = Chef::VERSION`

tyler-ball:
:+1:

chef-config/lib/chef-config/version.rb
mcquin: Wouldn't chef-config have to depend on chef in order to do that?

danielsdeleo: Yeah, Claire's reasoning is why we don't do that. `chef-config` is its own gem, it's just in the same repo as chef (the same way rails is). So we have a `VERSION` file at the root and a rake task to do the update everywhere.

tyler-ball: Thanks for the explanation!  Carry on

PR #3720, opened 2015-07-28
Resource.action.converge_if_changed

jkeiser:
The purpose behind `converge_if_changed` is as a more functional replacement for `converge_by`. It:

- Compares `current_resource` properties against `new_resource` and does not call the block unless something has changed.
- Automatically creates "green text" so users can see what's changing.

The docs are the best place to start looking: https://github.com/chef/chef/blob/96845d2b5ba10da650653fa022db72e79974a9cf/DOC_CHANGES.md#pure-ruby-actions

The salient points are:

1. Added `Chef::Resource::ActionProvider`, moving `load_current_resource` into it to increase efficiency and make the `action_provider_class` method less meta and easier to read.
2. Added [`converge_if_changed`](https://github.com/chef/chef/pull/3720/files#diff-31a0474b150c23d5e0e8669133fce6d9R178). This encompasses both the case where the resource exists, and where it does not (`current_resource` is `nil`). It iterates over the properties of the resource, looking for desired state properties that the user has specified that are *different* from the current resource. If a property is not set on the resource, it will not be compared.

converge_if_changed

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -193,6 +193,10 @@ class Chef
     # @return [Boolean] whether the block was executed.
     #
     def converge_if_changed(*properties, &converge_block)
+      if !converge_block
+        raise ArgumentError, "converge_if_changed must be passed a block!"
+      end
+
       properties = new_resource.class.state_properties.map { |p| p.name } if properties.empty?
       properties = properties.map { |p| p.to_sym }
       if current_resource
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -2,7 +2,7 @@
 # Author:: Adam Jacob (<adam@opscode.com>)
 # Author:: Christopher Walters (<cw@opscode.com>)
 # Author:: John Keiser (<jkeiser@chef.io)
-# Copyright:: Copyright (c) 2008-2015 Opscode, Inc.
+# Copyright:: Copyright (c) 2008-2015 Chef, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -1385,14 +1385,17 @@ class Chef
     #   created resource with its identity values filled in.
     #
     def self.load_current_value(&load_block)
-      include LoadCurrentValueDSL
       define_method(:load_current_value!, &load_block)
     end
 
-    module LoadCurrentValueDSL
-      def current_value_does_not_exist!
-        raise Chef::Exceptions::CurrentValueDoesNotExist
-      end
+    #
+    # Call this in `load_current_value` to indicate that the value does not
+    # exist and that `current_resource` should therefore be `nil`.
+    #
+    # @raise Chef::Exceptions::CurrentValueDoesNotExist
+    #
+    def current_value_does_not_exist!
+      raise Chef::Exceptions::CurrentValueDoesNotExist
     end
 
     #
--- a/lib/chef/resource/action_provider.rb
+++ b/lib/chef/resource/action_provider.rb
@@ -41,6 +41,8 @@ class Chef
           # Call the actual load_current_value! method. If it raises
           # CurrentValueDoesNotExist, set current_resource to `nil`.
           begin
+            # If the user specifies load_current_value do |desired_resource|, we
+            # pass in the desired resource as well as the current one.
             if current_resource.method(:load_current_value!).arity > 0
               current_resource.load_current_value!(new_resource)
             else


lib/chef/provider.rb
```diff
@@ -175,6 +175,57 @@ def converge_by(descriptions, &block)
...
+    #
+    # @return [Boolean] whether the block was executed.
+    #
+    def converge_if_changed(*properties, &converge_block)
+      properties = new_resource.class.state_properties.map { |p| p.name } if properties.empty?
+      properties = properties.map { |p| p.to_sym }
+      if current_resource
+        # Collect the list of modified properties
+        specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
+        modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
```
ranjib: are we confident in this equality check? if the provider does not implement load_current_resource properly, this will be error prone?

jkeiser: I can imagine some scenarios where two different values mean the same thing (like `'777'` and `0777` for `mode`, and the user would have to fix with either coercion and/or a type with an equality operator (or both). Is that what you meant?

One thing to keep in mind is the implementor of the resource is in control of both `load_current_resource` and calling `converge_if_changed`. Another is that the above error manifests as updating when you don't need to, which sucks, but is a much better problem than its counterpart, *not* updating when you *should* :)

Still, I'm all ears for specifics--is that the scenario you had in mind as far as "improper load_current_resource"?

lib/chef/resource.rb
```diff
@@ -1,7 +1,8 @@
 #
 # Author:: Adam Jacob (<adam@opscode.com>)
 # Author:: Christopher Walters (<cw@opscode.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Author:: John Keiser (<jkeiser@chef.io)
+# Copyright:: Copyright (c) 2008-2015 Opscode, Inc.
```
thommay: Chef, Inc. if we're updating the dates, please.

tyler-ball:
First, my nodes from reading the README:

1. I like the 'Handling multiple operations' block - answers a question I had.  But how does `converge_if_changed` know to call its block if no properties are set on it?  Does it compare all properties between new_resource and current_resource, then call the block if any of them are different?  That could be made more clear.

lib/chef/resource/action_provider.rb
```diff
@@ -0,0 +1,67 @@
...
+          current_resource.class.properties.each do |name,property|
+            if property.desired_state? && !property.identity? && !property.name_property?
+              property.reset(current_resource)
+            end
+          end
+
+          # Call the actual load_current_value! method. If it raises
+          # CurrentValueDoesNotExist, set current_resource to `nil`.
+          begin
+            if current_resource.method(:load_current_value!).arity > 0
```
tyler-ball: What is the arity check for?  Can we get some `TODO`s around cleaning this up when we don't have to be backward compatiable?

lib/chef/resource/action_provider.rb
```diff
@@ -0,0 +1,67 @@
...
+              current_resource.load_current_value!(new_resource)
+            else
+              current_resource.load_current_value!
+            end
+          rescue Chef::Exceptions::CurrentValueDoesNotExist
```
tyler-ball: What about `CurrentResourceDoesNotExist` for the error name?  Seems to make more sense than `Value`

lib/chef/resource/action_provider.rb
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/exceptions'
+
+class Chef
+  class Resource
+    module ActionProvider
```
jaym: :+1: `resource.rb` is too big

lib/chef/resource.rb
```diff
@@ -1382,9 +1385,16 @@ def self.action(action, &recipe_block)
     #   created resource with its identity values filled in.
     #
     def self.load_current_value(&load_block)
+      include LoadCurrentValueDSL
```
jaym: Can you add a comment as to why this happens in a method?

tyler-ball: Agreed - I am curious why we don't just include it at the top level (or just define the method in the class since it looks like thats all the module is doing)

lib/chef/resource/action_provider.rb
tyler-ball: To clarify my comment here - I don't understand what `Value` is referring to.  I based my recommendation for the name change off the fact that when we rescue this exception we set `current_resource` to nil.

lib/chef/provider.rb
```diff
@@ -175,6 +175,58 @@ def converge_by(descriptions, &block)
...
+    # - Prints out automatic green text saying what properties have changed.
+    #
+    # @param properties An optional list of property names (symbols). If not
+    #   specified, `new_resource.class.state_properties` will be used.
+    # @param converge_block The block to do the converging in.
+    #
+    # @return [Boolean] whether the block was executed.
+    #
+    def converge_if_changed(*properties, &converge_block)
+      properties = new_resource.class.state_properties.map { |p| p.name } if properties.empty?
```
tyler-ball: If someone calls `converge_if_changed(nil) do ... end` what will happen?  I would like to see a test for this.

lib/chef/resource/action_provider.rb
jkeiser: It's not about backcompat, it's just that I didn't want to require people to take the parameter with current_resource. However, it might be that blocks allow me to just always pass it and not worry about that. I'll check.

lib/chef/resource/action_provider.rb
jkeiser: @tyler-ball the `current_value_does_not_exist!` will always be used inside the block in `load_current_value`, so they are symmetric:

```ruby
class MyResource < Chef::Resource
  load_current_value do
    if !File.exist?(path)
      current_value_does_not_exist!
    end
  end
end
```

lib/chef/resource.rb
jkeiser: Because it doesn't make sense (it's an appendage) unless you are actually using `load_current_value`.  Happy to add it, but it's just a method that doesn't do anything for you unless you have load_current_value.

lib/chef/provider.rb
jkeiser: It will fail `.to_sym` with a nil error. Not sure if it's a concern though, as people will not generally be passing variables to this thing--always constants.

lib/chef/resource.rb
jkeiser: (And yeah, I'll add the comment :))

lib/chef/resource/action_provider.rb
jkeiser: Yep. Basically it's making sure these can both succeed:

```
load_current_resource do
  ...
end

load_current_resource do |desired_resource|
  ...
end
```

jkeiser:
@tyler-ball regarding DOC_CHANGES, I'm unsure how to put it other than https://github.com/chef/chef/pull/3720/files#diff-4a11ee028440aff81c1f8937bec62f8dR302 ... suggestions totally welcome!

lib/chef/resource.rb
jkeiser: OK, I relented and just added the method with a note in the method that it is used inside `load_current_value`.

lib/chef/resource/action_provider.rb
jkeiser: And, the reason for using `value` instead of `resource` is that this method runs in a different context than `load_current_resource` in the provider: that method is supposed to *create* the resource object and then fill it in, while this method has the resource created for it, and is supposed to fill it in using the actual value.  `load_current_value_into_resource` might be more fitting.

I'll think through whether to make it `load_current_resource` for muscle memory's sake though. (The other reason it's called this is this is what was accepted in the RFC :).

jkeiser:
@tyler-ball @jaym I added comments and moved `current_value_does_not_exist!` out of its little module for clarity. I have not made a change from the word `value` to `resource` (see above comments) and didn't make any changes around `nil`, so there's more discussion to be had there.

lib/chef/provider.rb
tyler-ball: I think resource authors will at some point type `converge_if_changed nil` and `converge_if_changed do ... end` and `converge_if_changed`.  When they write this in their resource and go test it, I think they are going to get some confusing errors raised.  A `no method to_sym on Nil` error is more confusing than an error that says `nil is not a valid property name`.

The latter clearly tells a new resource author what they need to fix to make their recipe work, while the former leaves them scratching their head.  Thats why I would suggest argument validation.

lib/chef/resource/action_provider.rb
tyler-ball: @jkeiser and I talked and I changed my mind - I think `Value` is fine.

coderanger:
Was this merged?

tyler-ball:
@coderanger yes it was - https://github.com/chef/chef/commit/b769876511717fb7b622183eeb5013ac7e1cd644

I can't remember why it didn't get merged from this PR, but there was a reason

PR #3721, opened 2015-07-28
Lcg/attributes v1.99

lamont-granquist:
rewriting attributes.

:construction: still under construction :construction:


lamont-granquist:
adding a comment to push to the bottom of our list.

lamont-granquist:
For future reference here's the answer to "why don't we just use hashie?" or some other library like that...

```
% pry
[3] pry(main)> require 'hashie'
=> true
[4] pry(main)> mash = Hashie::Mash.new
=> {}
[5] pry(main)> mash[:array] = []
=> []
[9] pry(main)> mash[:array].push({})
=> [{}]
[10] pry(main)> mash.class
=> Hashie::Mash
[11] pry(main)> mash[:array].class
=> Array
[12] pry(main)> mash[:array][0].class
=> Hash
```

That's a show-stopper right there.  Hashie takes the approach of inheriting from Hash to create Mashes, but it does not use Dan's VividMash/AttrArray approach to also wrap Array classes, so when you traverse an Array you wind up behaving like a normal Hash object on the other side, and lose your "Mashiness" or whatever behavior you implemented.

smurawski:
More conversation on attribute-y things - https://github.com/chef/chef/issues/4930

lamont-granquist:
this can probably be closed, i seem to be doing this in pieces iteratively to discover how far i really need to go with it...   see for example #5378

PR #3729, opened 2015-07-29
Fix #3692: flatten regex validation array so nested arrays work

### LINKED ISSUE #3692: Unknown reversion on master
Started getting a [failure this morning on my master builds](https://travis-ci.org/poise/poise-service-runit/jobs/71899472). This looks to be an issue in the runit cookbook, specifically [the `#owner` property](https://github.com/hw-cookbooks/runit/blob/v1.6.0/libraries/resource_runit_service.rb#L155-L157). It also looks like nightly builds were down for two weeks, so I can only narrow this to "some time between 12.4.1 and yesterday". The actual exception is `undefined method 'match' for [/^[^-+~:,\t\r\n\f\0]+[^:,\t\r\n\f\0]*$/]:Array`, which makes me think the value passed to `_pv_regex` is nested too deeply, which matches what I see in the code, the [`user_valid_regex` is itself an array](https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L561) so putting that in an array like the runit cookbook does would be nesting too deeply. The question then is why does this work in 12.4.1? I can't find the code change that actually broke things.

jkeiser:
Fixes #3692

### DIFF
--- a/lib/chef/mixin/params_validate.rb
+++ b/lib/chef/mixin/params_validate.rb
@@ -291,7 +291,7 @@ class Chef
       def _pv_regex(opts, key, regex)
         value = _pv_opts_lookup(opts, key)
         if !value.nil?
-          Array(regex).each do |r|
+          Array(regex).flatten.each do |r|
             return true if r.match(value.to_s)
           end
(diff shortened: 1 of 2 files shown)


coderanger:
:+1:

Issue #3735, opened 2015-07-31
Provide a way to override chef-client loggers.


docwhat:
Due to the way the `Chef::Application.configure_logging()` [resets `Chef::Log`](https://github.com/chef/chef/blob/master/lib/chef/application.rb#L137), it is impossible to add aditional loggers to `Chef::Log.loggers`.

My use case is that I would like my server nodes to use [`logstash-logger`](https://github.com/dwbutler/logstash-logger) to log all warnings, errors and fatals.  While I could use `loc_location` in `client.rb` the problem is that I want to send the meta information as JSON _without_ the timestamp and severity string that gets added to messages.

This would also make the WinEvt and Syslog loggers much simpler to understand and not require `Mixlib::Unformatter` (which smells bad).

You could rewrite the `loc_location` stuff to use the default ruby file logger (sort of like `MonoLogger` but simpler) when given a string (or IO object) and include a different way to add additional custom loggers.

lamont-granquist:
@jaym could you look at this?

docwhat:
My PR #3736 actually didn't fix this problem.  It is *sort of* on the right track, but it would have to be much more complicated because `:log_location` is set by default in many places.

I have a feeling that a lot of the logging has to be simplified.  Right now it takes looking through 6+ different files across 2 different projects to understand what's going on.

lamont-granquist:
:+1:  on simplifying the logging, it is definitely overly complicated.

PR #3736, opened 2015-07-31
Only call Chef::Log.init() if needed

docwhat:
Check if loc_location is set or if Chef::Log is unconfigured
before calling init() on it.

See issue #3735 for reason/use-case.

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -134,7 +134,9 @@ class Chef
     # that a user has configured a log_location in client.rb, but is running
     # chef-client by hand to troubleshoot a problem.
     def configure_logging
-      Chef::Log.init(MonoLogger.new(Chef::Config[:log_location]))
+      if Chef::Config[:log_location] || Chef::Log.unconfigured?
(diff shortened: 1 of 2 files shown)


lib/chef/application.rb
```diff
@@ -134,7 +134,9 @@ def set_specific_recipes
     # that a user has configured a log_location in client.rb, but is running
     # chef-client by hand to troubleshoot a problem.
     def configure_logging
-      Chef::Log.init(MonoLogger.new(Chef::Config[:log_location]))
+      if Chef::Config[:log_location] || Chef::Log.unconfigured?
```
thommay: should this be an `&&` rather than `||`? It's not clear how the second branch would ever be hit. Please can you add some tests for this.

docwhat:
Actually the `unconfigured?` thing is wrong altogether.

However, this patch doesn't work anyway because `:log_location` is set all the time.  So this doesn't really fix the problem outlined in issue #3735 -- Which is to allow using custom `Logger` classes instead of the complicated structure in place.

PR #3743, opened 2015-07-31
speed improvement for remote_directory() resource

Igorshp:
***Problem:***
doing `ls(path)` for a very large directory with > 1.8mil files takes upwards of 10 minutes.

***Solution:***
change filtering stage from `files.reject!` to files.reject
switching to `files.reject` decreases the runtime of that function from 650 seconds down to 6-7 seconds.
`files.reject!` does in place filtering and it looks like it copies the remaining array for each item filtered so it's running time is around o(n^2) vs o(n) for `files.reject`

***Second part of the optimisation:**
`files.map {|f| Chef::Util::PathHelper.cleanpath(f)}` takes almost 2 minutes in the above example when going over 1.5kk rows (after 300k were filtered out).
I've tested multiple samples, it appears that cleanpath() makes no difference to the original paths what so ever. Does it need to be there?
Could someone more knowledgeable about the reasons why it was put here in the first place comment on this?

All in all, after doing these changes, my chef run now takes only 4 minutes, down from 14.

Special thanks to @eam for helping to track this down and suggesting the fix.

Note: I've originally started looking this because that same remote_directory step alone was taking ***42 minutes***. I have since cleaned up about a third of data from that directory and the stats above are from after the cleanup.

### DIFF
--- a/lib/chef/provider/remote_directory.rb
+++ b/lib/chef/provider/remote_directory.rb
@@ -67,7 +67,7 @@ class Chef
                  ::File::FNM_DOTMATCH)
 
         # Remove current directory and previous directory
-        files.reject! do |name|
+        files = files.reject do |name|
           basename = Pathname.new(name).basename().to_s
           ['.', '..'].include?(basename)
         end


lib/chef/provider/remote_directory.rb
```diff
@@ -67,13 +67,10 @@ def ls(path)
                  ::File::FNM_DOTMATCH)
 
         # Remove current directory and previous directory
-        files.reject! do |name|
+        files = files.reject do |name|
```
jaym: you don't think you need `files =` as it will be returned since it's the last expression in the method

Igorshp: Oh yes, good point. `files = ` done before the map step was removed that used it. I'll update the pr

Igorshp:
closing to retest

Igorshp:
Struggling to reproduce test failures localy. Any help is appreciated

lamont-granquist:
Yeah seems to be failing only on windows, but its failing in the functional tests which makes me very nervous to merge as-is.   Will need to look at it when I've got a bit more time.

adamedx:
Here's what's failing on Windows:
`` 1) Chef::Resource::RemoteDirectory when creating the directory with purging enabled and there are extraneous files in the directory removes unmanaged files
     Failure/Error: resource.run_action(:create)
     Errno::ENOTEMPTY:
       remote_directory[C:/test-dir/directory_spec20150731-972-1he5e6c] (dynamically defined) had an error: Errno::ENOTEMPTY: Directory not empty - C:\test-dir\directory_spec20150731-972-1he5e6c/remotesubdir
     # ./lib/chef/provider/remote_directory.rb:99:in `rmdir'
     # ./lib/chef/provider/remote_directory.rb:99:in `block in purge_directory'
     # ./lib/chef/mixin/why_run.rb:52:in `call'
     # ./lib/chef/mixin/why_run.rb:52:in `add_action'
     # ./lib/chef/provider.rb:175:in `converge_by'
     # ./lib/chef/provider/remote_directory.rb:98:in `purge_directory'
     # ./lib/chef/provider/remote_directory.rb:86:in `block in purge_unmanaged_files'
     # ./lib/chef/provider/remote_directory.rb:78:in `each'
     # ./lib/chef/provider/remote_directory.rb:78:in `purge_unmanaged_files'
     # ./lib/chef/provider/remote_directory.rb:53:in `action_create'
     # ./lib/chef/provider.rb:144:in `run_action'
     # ./lib/chef/resource.rb:585:in `run_action'
     # ./spec/functional/resource/remote_directory_spec.rb:172:in `block (4 levels) in <top (required)>'
 
```

lamont-granquist:
We're probably doing cleanpath for a reason on windows and dropping that probably regressed something.  It'd probably be useful to look back in git history and see why that line was added.

eam:
The majority of the speedup will come from switching `reject!` to `reject`, which should be a very safe change.

Perhaps wrap cleanpath in a platform-specific check? Say, `if Gem.win_platform?`?

lamont-granquist:
We already have a `Chef::Platform.windows?` helper

Igorshp:
Yeah, I thought as much. I could not find what clean path does but, as you say, it must've been put there for a reason.

I'll make a quick test patch with it back in, let's see if that will pass the CI.

Igorshp:
Ok, so it is indeed the `clean path` line.

Going to try to enable it only for windows

Igorshp:
Well, doing `cleanpath` only for windows didn't seem to work as expected. 

As @eam said, the major part of the speedup is from changing `reject!`, speeding up `cleanpath` would be icing on the cake. 

Any insight into why exactly `cleanpath` is required would be much appreciated.

I'll rebase the commits shortly.

Igorshp:
I've done some experiments, this seems to be the fastest implementation without doing some drastic changes.

Here's the current benchmark with 1.9 million files.

| Version | Time |
|------------|--------|
|Original| 664.260273s |
|This PR| 111.063156s |

thommay:
@Igorshp i think you lost the changes that @jaym requested in code review, but other than that I'm :100: :+1:  cc @chef/client-core @jkeiser

Igorshp:
@thommay his comments were for a previous patch when `cleanpath` was removed.

`array.reject` returns an array, but without the `cleanpath` line it was the last thing in the function without the need to save the output to variable, it was returning it directly.

now that `cleanpath` is back, we must save the output. I'll reply to his comment

lib/chef/provider/remote_directory.rb
Igorshp: I've put the `cleanpath` line back in so it needs to save the output for .reject to file.

thommay:
Ah, apologies. As you were :)

tyler-ball:
:+1:

Igorshp:
Squashed commits and rebased on master

danielsdeleo:
LGTM :+1:

lamont-granquist:
Suspect this is also a huge hit on memory size and pressure as well, at least I'm definitely seeing that problem with remote_directory.

lamont-granquist:
chef-gctrace

That is live slots from a chef-client run, most of the dots are forced full GCs that i instrumented chef-client with so its really showing live objects, not eden garbage that hasn't been swept yet.  The huge spike in the middle is a single remote_file provider (that doesn't really do that much, its just a couple dozen dotfiles, nothing like 1.8M files).

Igorshp:
@lamont-granquist is this from this PR branch or current master?
also could you clarify that it's remote_file or remote_directory?

lamont-granquist:
yeah its still early, i meant remote_directory

does not look like your branch fixes the memory issue, though, but I'm :+1:  on this and will merge so that I can hack on top of this one

Igorshp:
@lamont-granquist This tweak should actually make the memory use slightly worse, as it creates a new array instead of sorting in place. But I doubt it's by much relative to everything else.

The time saved is worth it in my opinion.

lamont-granquist:
Yeah but that is pretty meaningless compared to whatever is exploding memory in there, I agree that isn't the issue.

lamont-granquist:
Well, I think I see it.  In order to purge files we slurp the whole directory structure into memory decorated with a Set and possibly Pathname.  This is particularly useless if you're not doing purge at all (like I am), and could probably be better done by simply traversing the directory structure at the end and skipping transferred files, but that'll take a bit of a major rewrite of the purge algorithm...

lamont-granquist:
Okay you're merged, I may wind up refactoring this so your code disappears though before release...

lamont-granquist:
(i'll ping you on my PR when its ready so you can test your use case and make sure I didn't do anything horrible)

PR #3744, opened 2015-07-31
Remove freeze of defaults, add warning for array/hash constant defaults

jkeiser:
Relies on https://github.com/chef/mixlib-log/pull/10

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -89,9 +89,6 @@ class Chef
       options[:name_property] = options.delete(:name_attribute) if options.has_key?(:name_attribute) && !options.has_key?(:name_property)
       @options = options
 
-      if options.has_key?(:default)
-        options[:default] = options[:default].freeze
-      end
       options[:name] = options[:name].to_sym if options[:name]
       options[:instance_variable_name] = options[:instance_variable_name].to_sym if options[:instance_variable_name]
     end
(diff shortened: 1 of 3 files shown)


coderanger:
I'm guessing there was some internal discussion about this that would be good to know as context?

jkeiser:
@coderanger yeah, the issue is it showed up in another cookbook (this one using `set_or_return` directly). Much as I really want to simply fix the issue for everyone, I'm starting to worry that it will cause enough breaks that it will become an upgrade blocker for people. ("Please upgrade the cookbooks you depend on and restart your whole process.").  Upgrade blockers = less adoption, and dot releases aren't supposed to do that (much).

coderanger:
That is fair. I'll leave it in Poise as that is opt-in so it should be safe and we can put reverting this on the Chef 13 checklist?

jkeiser:
@coderanger Sounds good! I've removed the `warn_once` functionality until we get mixlib-log worked out; you mind C/Ring it?

This should be less chatty than some warnings due to the fact that it happens on declaration rather than on use.

coderanger:
:+1: shouldn't block 12.5 on the warn_once stuffs.

coderanger:
Would you say that freezing defaults should be on the slate for 13? I can add it to the list :)

jkeiser:
Yeah. It's in the RFC. I think we need to be a little more proactive about fixing cookbooks--like, add a rule to foodcritic or something. It would suck if 13 required an upgrade of all cookbooks everywhere rather than being *mostly* compat.

PR #2007, opened 2014-09-06
Speed up file system access by not listing children on every child(name) access.

jkeiser:
Also removes an intermittent warning when first downloading large numbers of
cookbooks, where sometimes a cookbook directory has been created but its children
have not.   When you run child('other_cookbook') it lists 'cookbooks'.children and
notices the empty directory.  Annoying, and probably affects knife-ec-backup.

PR #3746, opened 2015-08-02
Added caching of children to avoid traversing the file system on every file request

tpodom:
When using chef-client in local mode the file system access is extremely slow and causes the resolving of the run list cookbooks to take over 30 minutes for my team.  

When digging into why it was taking so long we discovered that every file request was re-iterating all of the files because the way the children get built.  I saw that ```ChefRepositoryFileSystemRootDir``` cached its created children so I took that concept and extended it to the other related objects.  The caching has significantly improved the performance and the entire resolving & synchronization takes under 20 seconds now.

Here are some performance numbers using Ruby Benchmark on some of the file operations before & after:

## Before
```
LIST organizations/chef/cookbooks/ntp:   0.010000   1.320000   1.330000 (  1.325998)
LIST organizations/chef/cookbooks/sudo:   0.010000   1.270000   1.280000 (  1.291669)
LIST organizations/chef/cookbooks/ssh_known_hosts:   0.000000   1.280000   1.280000 (  1.284465)
LIST organizations/chef/cookbooks/chef-timezone:   0.010000   1.330000   1.340000 (  1.344935)
LIST organizations/chef/cookbooks/postfix:   0.010000   1.340000   1.350000 (  1.367667)
GET organizations/chef/cookbooks/ntp/1.1.8:   0.030000   2.700000   2.730000 (  2.743045)
GET organizations/chef/cookbooks/sudo/2.7.1:   0.020000   2.700000   2.720000 (  2.722147)
GET organizations/chef/cookbooks/ssh_known_hosts/2.0.0:   0.020000   2.630000   2.650000 (  2.660303)
GET organizations/chef/cookbooks/chef-timezone/0.0.1:   0.030000   2.670000   2.700000 (  2.711711)
GET organizations/chef/cookbooks/postfix/3.6.2:   0.060000   2.580000   2.640000 (  2.645569)
```

## After
```
LIST organizations/chef/cookbooks/ntp:   0.000000   0.010000   0.010000 (  0.001487)
LIST organizations/chef/cookbooks/sudo:   0.000000   0.000000   0.000000 (  0.001182)
LIST organizations/chef/cookbooks/ssh_known_hosts:   0.000000   0.000000   0.000000 (  0.000966)
LIST organizations/chef/cookbooks/chef-timezone:   0.000000   0.000000   0.000000 (  0.001770)
LIST organizations/chef/cookbooks/postfix:   0.000000   0.000000   0.000000 (  0.001557)
GET organizations/chef/cookbooks/ntp/1.1.8:   0.000000   0.040000   0.040000 (  0.046823)
GET organizations/chef/cookbooks/sudo/2.7.1:   0.010000   0.030000   0.040000 (  0.042465)
GET organizations/chef/cookbooks/ssh_known_hosts/2.0.0:   0.000000   0.010000   0.010000 (  0.018355)
GET organizations/chef/cookbooks/chef-timezone/0.0.1:   0.000000   0.010000   0.010000 (  0.020508)
GET organizations/chef/cookbooks/postfix/3.6.2:   0.030000   0.040000   0.070000 (  0.063699)
```

### DIFF
--- a/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbook_dir.rb
+++ b/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbook_dir.rb
@@ -58,7 +58,7 @@ class Chef
         end
 
         def children
-          begin
+          @children ||= begin
             Dir.entries(file_path).sort.
                 select { |child_name| can_have_child?(child_name, File.directory?(File.join(file_path, child_name))) }.
                 map { |child_name| make_child(child_name) }.
--- a/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbook_entry.rb
+++ b/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbook_entry.rb
@@ -34,7 +34,7 @@ class Chef
         attr_reader :recursive
 
         def children
-          begin
+          @children ||= begin
             Dir.entries(file_path).sort.
                 select { |child_name| can_have_child?(child_name, File.directory?(File.join(file_path, child_name))) }.
                 map { |child_name| make_child(child_name) }.
--- a/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbooks_dir.rb
+++ b/lib/chef/chef_fs/file_system/chef_repository_file_system_cookbooks_dir.rb
@@ -37,7 +37,7 @@ class Chef
         attr_reader :chefignore
 
         def children
-          begin
+          @children ||= begin
             Dir.entries(file_path).sort.
                 select { |child_name| can_have_child?(child_name, File.directory?(File.join(file_path, child_name))) }.
                 map { |child_name| make_child(child_name) }.
--- a/lib/chef/chef_fs/file_system/chef_repository_file_system_entry.rb
+++ b/lib/chef/chef_fs/file_system/chef_repository_file_system_entry.rb
@@ -72,7 +72,7 @@ class Chef
 
         def children
           # Except cookbooks and data bag dirs, all things must be json files
-          begin
+          @children ||= begin
             Dir.entries(file_path).sort.
                 select { |child_name| can_have_child?(child_name, File.directory?(File.join(file_path, child_name))) }.
                 map { |child_name| make_child(child_name) }
--- a/lib/chef/chef_fs/file_system/file_system_entry.rb
+++ b/lib/chef/chef_fs/file_system/file_system_entry.rb
@@ -31,6 +31,8 @@ class Chef
         def initialize(name, parent, file_path = nil)
           super(name, parent)
           @file_path = file_path || "#{parent.file_path}/#{name}"
+          @parent = parent
+          @children = nil
         end
 
         attr_reader :file_path
@@ -40,7 +42,7 @@ class Chef
         end
 
         def children
-          begin
+          @children ||= begin
             Dir.entries(file_path).sort.select { |entry| entry != '.' && entry != '..' }.map { |entry| make_child(entry) }
           rescue Errno::ENOENT
             raise Chef::ChefFS::FileSystem::NotFoundError.new(self, $!)
@@ -61,6 +63,7 @@ class Chef
               raise Chef::ChefFS::FileSystem::AlreadyExistsError.new(:create_child, child)
             end
           end
+          @children = nil
           child
         end
 
@@ -69,7 +72,7 @@ class Chef
         end
 
         def delete(recurse)
-          if dir?
+          result = if dir?
             if !recurse
               raise MustDeleteRecursivelyError.new(self, $!)
             end
@@ -77,6 +80,10 @@ class Chef
           else
             File.delete(file_path)
           end
+          if @parent
+            @parent.child_removed(self)
+          end
+          result
         end
 
         def exists?
@@ -92,6 +99,7 @@ class Chef
         end
 
         def write(content)
+          @children = nil
           File.open(file_path, 'wb') do |file|
             file.write(content)
           end
@@ -101,6 +109,10 @@ class Chef
 
         def make_child(child_name)
           FileSystemEntry.new(child_name, self)
+        end
+
+        def child_removed(child)
+          @children = nil
         end
       end
     end


tpodom:
Looking into the test failures

jkeiser:
Hmm, we'd been trying to avoid caching children because it was causing unhappiness--we were trying to keep a mostly-immediate API and optimize the algorithms so they only called `.children` once. But, we've noticed the slowness too, and good lord this has to get fixed soon, so I'd take this if an algorithm fix didn't appear.

Do you know who the callers are that are causing this? It sounds like it happens on a simple directory list?

tpodom:
Yeah the caching is a good point.  In fact that's what's causing the test failures one of the cookbook deletes isn't getting removed from the caches causing later tests to fail.

As far as what's causing the extra iterations it's two things:

1.  In ```Chef::ChefFS::ChefFSDataStore.with_entry``` it calls ```to_chef_fs_path``` which calls ```get_single_cookbook_version``` which results in a call to ```Chef::ChefFS::FileSystem.resolve_path``` which in turn iterates each path segment asking for the child. Then after ```to_chef_fs_path``` finishes ```with_entry``` turns around and calls ```Chef::ChefFS::FileSystem.resolve_path``` again which again iterates the children to retrieve a specific child.  That results in 2 iterations of all of the cookbooks on a single GET

2.  From chef-client's perspective it's pretty much doing 1 or 2 passes through for every LIST or GET because LIST will at least call 

From looking through this more I'd say the best solution is to avoid calling children on every ```child(name)``` call as proposed by #2007.

jkeiser:
@tpodom thanks tons for the info! I think you missed the end of the sentence on your second point? I'm not clear what that one is about.

I've started updating #2007, but I'm not sure about your specific solution there--`child` is required to return something in all cases, though it doesn't *have* to be a cookbook dir, it could be a Nonexistent entry.

jkeiser:
Regarding number 1: `child(name)` is supposed to be a costless operation. Is that not the case for one of these? That would totally be the cause, as it is baked into all our assumptions.

tpodom:
@jkeiser Right now for every call to ```resolve_path``` it's iterating the entire cookbook directory structure. ```resolve_path`` is calling ```child(name)``` on ```MultiplexedDir``` which is handled by ```BaseFSDir::child(name)``` which results in calling ```MultiplexedDir::children```.  ```MultiplexedDir::children``` is then triggering calls to ```children``` on all of the structure which eventually gets to ```ChefRepositoryFileSystemCookbooksDir::children``` which walks all of the cookbook directories and calls ```ChefRepositoryFileSystemCookbookDir::children``` which walks each individual cookbook.  

My point about #2 can mostly be ignored, I was trying trying to say every time chef-client calls LIST or GET it's iterating the entire cookbook directory structure at least once because ```resolve_path``` gets called.  With the way ```ChefFSDataStore::with_entry``` is written any time that is called ```resolve_path``` is called twice resulting in two iterations of the cookbook directory structure.  In one of the cases of LIST it doesn't call ```with_entry``` but it does call ```get_single_cookbook_version``` which calls ```resolve_path``` triggering the directory walking.  All of that wouldn't really matter if the ```child(name)``` call wasn't expensive for ```MultiplexedDir``` and ```ChefRepositoryFileSystemCookbooksDir``` classes.

On #2007 I was just trying to come up with an example for implementing ```child(name)``` but you are right I was failing to return the ```NonexistentFSObject```.  I also had missed at first that you had modified the base ```FileSystemEntry``` to prevent the ```ChefRepositoryFileSystem*``` classes from iterating children on ```child(name)```.  However, I wonder if the implementation of the ```FileSystemEntry::child(name)``` should ask if the child is valid to allow for some of the filtering logic that's performed in the ```children``` functions of the subclasses?

jkeiser:
@tpodom does this happen only when synchronizing a cookbook for the first time, or is it also slow the second time? I've got me a repro and I'm trying to figure out just what's *supposed* to be slow and what isn't.

tpodom:
It happens every time chef-client does a LIST or GET on any cookbook (and I supposed on any file contained within the cookbook too but I haven't verified that).  It appears that chef-client does a LIST on the cookbook then performs a GET so that combination results in 3 complete directory traversals just to retrieve the cookbook metadata.  It then does that for every cookbook in the run list so if there are 10 cookbooks it iterates the complete cookbook directory structure 30 times.

jkeiser:
@tpodom OK cool. Was just checking. So that means if you run `chef-client` multiple times in a row, it will continue to repro, if I read it correctly.

tpodom:
@jkeiser correct, every chef-client run takes > 30 minutes for us to synchronize cookbooks.

jkeiser:
I will note that with a rather large run list (which is nonetheless stored entirely locally), it takes me upwards of 15 minutes to sync the first time, and 1.8 seconds to sync the second time. That's why I ask: it feels like the deeper problem exists on the initial sync, even though 1.8 seconds is nothing to sneeze at.

My repro:

1. `mkdir repo`
2. `cd repo`
3. Put this in `cookbooks/blah/metadata.rb`:
```ruby
name 'blah'
version "1.5.1"
description "x"
depends "java"
depends "jenkins"
depends "windows"
depends "yum"
depends "mysql"
depends "postgresql"
depends "apache2"
depends "docker"
depends "newrelic"
depends "bacon"
depends "apt"
depends "build-essential"
depends "chef-client"
```
4. Make an empty `cookbooks/blah/recipes/default.rb`
5. Put this in `Policyfile.rb`:
```ruby
name "me"
default_source :community
run_list "blah"
cookbook "blah", path: "cookbooks/blah"
```
6. Run `chef install`
7. Run `chef export repo`
8. Put this in `repo/.chef/knife.rb`:
```ruby
use_policyfile true
policy_document_native_api false
deployment_group 'me-local'
versioned_cookbooks true
```
9. `cd repo`
10. `chef-client -z`

And that will repro.

tpodom:
@jkeiser There could also be something different between your repo setup and ours.  We're still using the old style repo where we're storing all of the cookbooks, environments, roles, data bags in a single git repo.  I have a test rb file that I wrote that I can upload when I get home that just mocks up two back to back get requests on the same cookbook.

jkeiser:
That would be great!  And I really really appreciate your sleuthing tracking this down.

I'm using an old style repo as well (`chef export` creates an old style repository with your chosen cookbooks in it). It's worth running the repro to see if we're seeing the same thing, IMO.

tpodom:
@jkeiser OK I did just try out your example and the first sync was definitely slow but subsequent runs were fast.  One difference is that your example is using versioned cookbooks so I'm not sure if that is changing any different behavior.

jkeiser:
@tpodom it's possible but probably isn't. I think though, that the first-sync case is the one that we're trying to optimize here. I see a huge difference with and without patch.

jkeiser:
Docker in particular is the worst offender I've seen. I'm going to try loading just that and see what I see.

tpodom:
@jkeiser the test script i wrote that's setting up the data store isn't working correctly with the versioned cookbooks set up from the ```chef export```.  I'll see if I can get a different test working.

tpodom:
@jkeiser I've uploaded a repo to https://github.com/tpodom/slow-repo.  I added two scripts in there, ```test_get.rb``` which will perform two GET requests on a single cookbook.  You can run ```ruby test_get.rb``` to see the timings.  If you install the unroller gem you can uncomment the unroller sections and see a full code trace of the calls being made.

The other script is ```test_run.sh``` which will run ```chef-client``` pointing at ```.chef/client.rb```.  

One thing I've observed is that on my Mac the ```test_run.sh``` takes about 30 seconds but running over a shared filesystem from VMWare on my Linux VM it takes about 5 minutes.  So part of the problem we're seeing with the extremely long run time with our repo is due to the fact that the disk access is slower from our Linux VMs.  We're using Vagrant to spin up Linux VMs and it's mounting the cookbooks on Linux to run chef-zero.

I ran a few timings of ```test_run.sh``` and ```test_get.sh``` before and after applying the patch from #2007.

Before/After times of ```test_run.sh``` for initial synchronization:
Mac OS X time before patch: INFO: Chef Run complete in 324.120637 seconds
Mac OS X time after patch: INFO: Chef Run complete in 7.70032 seconds (yes it did re-download the cookbook files)

Linux VM time before patch: INFO: Chef Run complete in 2401.952649893 seconds
Linux VM time after patch: INFO: Chef Run complete in 55.87347354 seconds

Before/After times of ```test_run.sh``` after the initial synchronization:

Mac OS X time before patch: INFO: Chef Run complete in 32.776539 seconds
Mac OS X time after patch: INFO: Chef Run complete in 7.047535 seconds

Linux VM time before patch: INFO: Chef Run complete in 286.799093342 seconds
Linux VM time after patch: INFO: Chef Run complete in 50.596094974 seconds

Before/After times of ```test_get.rb``` with patch from #2007.

Mac OS X time before patch: 
```
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.530000   0.120000   0.650000 (  0.642569)
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.520000   0.110000   0.630000 (  0.632847)
```
Mac OS X time after patch: 
```
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.000000   0.000000   0.000000 (  0.001439)
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.010000   0.000000   0.010000 (  0.000871)
```

Linux VM time before patch: 
```
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.010000   5.860000   5.870000 (  5.913868)
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.050000   6.020000   6.070000 (  6.505786)
```
Linux VM time after patch: 
```
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.000000   0.000000   0.000000 (  0.009295)
GET ["organizations", "chef", "cookbooks", "a", "0.1.0"] took:   0.000000   0.010000   0.010000 (  0.003607)
```

tpodom:
Closing in favor of the fix from #2007

PR #3750, opened 2015-08-03
Use fixed vagrant-wrapper, kitchen-ec2 and test-kitchen gems

jkeiser:
@tball this incorporates kitchen-ec2 and test-kitchen master branch, as well as an as-yet-merged-or-released change to vagrant-wrapper that fixes kitchen-tests on OS/X.

### DIFF
--- a/kitchen-tests/Gemfile
+++ b/kitchen-tests/Gemfile
@@ -2,9 +2,9 @@ source "https://rubygems.org"
 
 group :end_to_end do
   gem 'berkshelf'
-  gem 'test-kitchen', '~> 1.4.0'
+  gem 'test-kitchen', '~> 1.4.0', github: 'test-kitchen/test-kitchen'
   gem 'kitchen-appbundle-updater', '~> 0.0.1'
   gem "kitchen-vagrant", '~> 0.17.0'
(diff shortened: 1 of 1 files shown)


coderanger:
Whats the need for TK master?

jkeiser:
@coderanger https://github.com/test-kitchen/test-kitchen/commit/171812f9c67561fe5cb21c52fb7a1156cba3bd82

I can't run the tests on OS/X without it.

tyler-ball:
@jkeiser the TK changes got released as 1.4.2, so you can depend on that instead of master

kitchen-tests/Gemfile
```diff
@@ -2,9 +2,9 @@ source "https://rubygems.org"
 
 group :end_to_end do
   gem 'berkshelf'
-  gem 'test-kitchen', '~> 1.4.0'
+  gem 'test-kitchen', '~> 1.4.0', github: 'test-kitchen/test-kitchen'
```
tyler-ball: @jkeiser the TK changes got released as 1.4.2, so you can depend on that instead of master

jkeiser:
Hrm, actually, if that's the case, I think I'll just have this whole thing wait until vagrant-wrapper is released. test-kitchen was the more important dep for stability's sake.

PR #3751, opened 2015-08-04
Remove old bootstrap templates

juliandunn:
* We don't support installing from gems anymore as a first-class citizen
* Now that the bootstrap template supports regular Bourne shell (not bash), the AIX bootstrap is no longer necessary.


coderanger:
:+1: for what I think you mean. We support installing from gems, just not with knife bootstrap et al :P

juliandunn:
Yeah, that's what I meant. :grin:

thommay:
:+1:

PR #3728, opened 2015-07-29
Rewrite NetLocalGroup things to use FFI

jaym:
This should now be less broken and works on both 32bit and 64 bit.

cc @adamedx @chefsalim @ksubrama @mcquin @smurawski

PR #3754, opened 2015-08-05
Fix functional tests for group resource - fix #3728

ksubrama:


ksubrama:
Related to https://github.com/chef/chef/pull/3728

ksubrama:
We figured out that the tests weren't failing on travis because this test needs either root or windows admin. That would explain why travis and appveyor were both happy with it.  I've manually run the test on a non-domain joined Windows box and on my Mac (as root).  If anyone else wants to try a different platform, you're welcome to do so :)

thommay:
Looks reasonable to me :+1:

ksubrama:
Merging to unbreak omnibus.

PR #3757, opened 2015-08-06
Add additional helpful section for frozen objects

### LINKED ISSUE #3734: Better error message when user tries to write to a badly-declared property
```ruby
class Foo < Chef::Resource::LWRPBase
  resource_name :foo
  attribute :myhash, default: {}
  attribute :mylist, default: []
end
foo 'x' do
  mylist << 'hi'
  myhash[:a] = 10
end
```

Emits this error:

`ERROR: can't modify frozen Array`

We should put something explaining the situation, if we can.

We might want to warn on property declaration in the case of empty array and empty hash anyway, as they are super common.

martinb3:
Augument runtime errors with additional information that explains why an object might be frozen, and how frozen resource properties are a good thing.

Fixes #3734, for [accepted RFC 55](https://github.com/chef/chef-rfc/pull/135).

### DIFF
--- a/lib/chef/formatters/error_inspectors/compile_error_inspector.rb
+++ b/lib/chef/formatters/error_inspectors/compile_error_inspector.rb
@@ -44,6 +44,18 @@ class Chef
             error_description.section("Cookbook Trace:", traceback)
             error_description.section("Relevant File Content:", context)
           end
+
+          if exception_message_modifying_frozen?
+            msg = <<-MESSAGE
+            Chef calls the freeze method on certain ruby objects to prevent
+            pollution across multiple instances. Specifically, resource
+            properties have frozen default values to avoid modifying the
+            property for all instances of a resource. Try modifying the
+            particular instance variable or using an instance accessor instead.
+            MESSAGE
+
+            error_description.section("Additional information:", msg.gsub(/^ {6}/, ''))
+          end
         end
 
         def context
@@ -109,6 +121,10 @@ class Chef
               Chef::Log.debug("filtered backtrace of compile error: #{r.join(",")}")
               r
             end
+        end
+
+        def exception_message_modifying_frozen?
+          exception.message.include?("can't modify frozen")
         end
 
       end
--- a/spec/unit/formatters/error_inspectors/compile_error_inspector_spec.rb
+++ b/spec/unit/formatters/error_inspectors/compile_error_inspector_spec.rb
@@ -110,6 +110,32 @@ describe Chef::Formatters::ErrorInspectors::CompileErrorInspector do
       end
     end
 
+    context "when the error is a RuntimeError about frozen object" do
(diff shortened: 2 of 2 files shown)


danielsdeleo:
@chef/client-core Ready to review

lib/chef/formatters/error_inspectors/compile_error_inspector.rb
```diff
@@ -44,6 +44,18 @@ def add_explanation(error_description)
...
             error_description.section("Relevant File Content:", context)
           end
+
+          if exception_message_modifying_frozen?
+            msg = <<-MESSAGE
+            Chef calls the freeze method on certain ruby objects to prevent
+            pollution across multiple instances. Specifically, resource
+            properties have frozen default values to avoid modifying the
+            property for all instances of a resource. Try modifying the
+            particular instance variable or using an instance accessor instead.
```
danielsdeleo: Can this be made more clear? Should we have code examples? Do we need to mention `dup`? Also, values from `ENV` can be frozen, should we mention this?

danielsdeleo:
Thanks for this. I wouldn't mind merging as-is, because it's clearly an improvement over what we have now (namely, nothing). That said, I think we can maybe make some of the error message a bit clearer.

lamont-granquist:
:+1: lets ship it

martinb3:
I'll submit a PR with further refinements next week. Thanks for the merge!

martinb3:
After doing another revision, I'm failing to come up with anything significantly better.

danielsdeleo:
So this:

> Try modifying the particular instance variable or using an instance accessor instead.

seems pretty difficult to parse. Having a code sample showing what to change would be ideal, I think.

I think it also is worth mentioning ENV variables, which folks sometimes use as an easy way to pass input from the command line, would be nice.

martinb3:
How about something like:
```
            msg = <<-MESSAGE
            Ruby objects are often frozen to prevent further modifications
            when they would negatively impact the process (e.g. values inside
            Ruby's ENV class) or to prevent polluting other objects when default
            values are passed by reference to many instances of an object (e.g.
            the empty Array as a Chef resource default, passed by value to every
            instance of the resource).

            Chef uses Object#freeze to ensure the default values of properties
            inside Chef resources are not modified, so that when a new instance
            of a Chef resource is created, Object#dup always copies a clean, un-
            modified object to a new instance of a resource.

            Instead of modifying an object that provides a default value for a
            Chef resource, create a new object and assign it to the resource's
            parameter, e.g.:
              fruit_basket.contents %w(apple)
              - instead of -
              fruit_basket.contents << 'apple'
            MESSAGE
```

danielsdeleo:
That looks really nice :100:

martinb3:
Sweet! I'll create a PR :)

PR #3759, opened 2015-08-07
Don't use shell_out! on "lssrc -g" 

### LINKED ISSUE #3730: service idempotency not correct in AIX
When using an LWRP like the following on AIX with Chef 12:

```ruby
service "xntpd" do
 supports :restart => true
 action [ :start ]
end
```

The recipe will always report it is starting the service, even when the service is already running.  Idempotency seems to not be functioning properly.

juliandunn:
It will raise an exception if the service is not a group; instead what we want is to check the return code. If 0, it's a group, otherwise, it is not.

Fixes #3730.

### DIFF
--- a/lib/chef/provider/service/aix.rb
+++ b/lib/chef/provider/service/aix.rb
@@ -116,7 +116,7 @@ class Chef
         end
 
         def is_resource_group?
-          so = shell_out!("lssrc -g #{@new_resource.service_name}")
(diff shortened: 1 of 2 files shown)


juliandunn:
Can I get a quick review @chef/client-aix

lamont-granquist:
:+1:

thommay:
:+1:

PR #3760, opened 2015-08-07
Refactor all the gem building logic into a custom rake task.

ksubrama:
I plan on pulling this into chef-dk and any other gem we release similarly later.  If you'd rather I create a separate gem/repo for this, I can do that too.


ksubrama:
/cc @mcquin @danielsdeleo @thommay @lamont-granquist 

Can one of you cc chef-client-core.  I'm not on it.

chef-config/lib/chef-config/package_task.rb
```diff
@@ -0,0 +1,214 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+# NOTE: This file is generated by running `rake version` in the top level of
+# this repo. Do not edit this manually. Edit the VERSION file and run the rake
+# task instead.
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+module #{module_name}
```
danielsdeleo: This doesn't work because a long time ago we incorrectly made the top level constant `Chef` a class and it's probably too late to fix it :crying_cat_face:

danielsdeleo:
Sadly it won't work as written, I like the general idea though. :+1: once you work around the problem I pointed out.

danielsdeleo:
And cc: @chef/client-core

chef-config/lib/chef-config/package_task.rb
ksubrama: Just noticed the appveyor failure.  Fixed with a sad extra flag.

ksubrama:
Fiddlesticks!  Travis failure is because https://github.com/jaym/chef-appbundle-updater/blob/master/recipes/default.rb invokes rake outside bundler which is fine if there was no dependency between the Rakefile and chef-config.

schisamo:
:+1:

chef-config/lib/chef-config/package_task.rb
```diff
@@ -0,0 +1,223 @@
...
+    attr_accessor :generate_version_class
+
+    # Paths to the roots of any components that also support ChefPackageTask.
+    # If relative paths are provided, they are rooted against root_path.
+    attr_accessor :component_paths
+
+    # This is the module name as it appears on the path "lib/module/".
+    # e.g. for module_name  "ChefDK", you'd want module_path_name to be "chef-dk".
+    # The default is module_name but lower-cased.
+    attr_writer :module_path_name
```
tyler-ball: I would probably call this just `module_path` since it can end up being something like `chef/provisioning`

tyler-ball:
Thanks for putting this together, @ksubrama !  :+1:

chef-config/lib/chef-config/package_task.rb
ksubrama: This name is used in gem-uninstall - so it might not work there.  I'll rename it but as we run into it, we might need to simply split module_path and take the last fragment for the uninstall task.

ksubrama:
Waiting on https://github.com/jaym/chef-appbundle-updater/pull/6 and a new release to fix travis.

ksubrama:
And it's alive!

PR #3768, opened 2015-08-11
Make reboot_pending? look for CBS RebootPending

jeremiahsnapp:
reboot_pending should look for CBS RebootPending instead of RebootRequired

Ref: https://gallery.technet.microsoft.com/scriptcenter/Get-PendingReboot-Query-bdb79542


ksubrama:
:+1:

adamedx:
@randomcamel, is this something you worked on in the past? Would be good to get another +1.

randomcamel:
@adamedx @jeremiahsnapp I'm not an expert in the registry keys, but this change matches the behavior of [opscode-cookbooks/windows](https://github.com/opscode-cookbooks/windows/blob/4b3888bbcc8255cb20e54c614e649c806eccd935/files/default/handlers/windows_reboot_handler.rb#L68), so :+1: for that.

PR #3771, opened 2015-08-12
Add support for override depth and adding test in overriding depth

renanvicente:
This change would allow to override depth on deploy.

    include_recipe 'deploy'
    r = resources('deploy[/srv/www/simpleapp]')
    r.depth 1

this also would allow to set depth when creating a deploy resource.

    deploy 'test repo with depth' do
      repo 'https://github.com/renanvicente/lab_ecommerce_simple'
      user 'root'
      depth 1
      deploy_to '/tmp/test_with_depth'
      action :deploy
    end

why we need that?
Some times you may have a git repo with a large .git and you do not want to clone everything because it will take to long. We have shallow_clone but it only allow changing it to 5.

### DIFF
--- a/lib/chef/resource/deploy.rb
+++ b/lib/chef/resource/deploy.rb
@@ -27,6 +27,7 @@
 #   migration_command "rake db:migrate"
 #   environment "RAILS_ENV" => "production", "OTHER_ENV" => "foo"
 #   shallow_clone true
+#   depth 1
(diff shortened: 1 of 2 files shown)


lamont-granquist:
this LGTM :+1:

lamont-granquist:
we need another :+1: from @chef/client-core

thommay:
:+1:

PR #3772, opened 2015-08-13
Add ps_credential dsl method to dsc_script

jaym:
You can now do the following:

```ruby
dsc_script 'create-foo-user' do
  code <<-EOH
     User FooUser00
     {
       Ensure = "Present"
       UserName = 'FooUser00'
       Password = #{ps_credential('FooBarBaz1!')}
     }
  EOH
  configuration_data <<-EOH
    @{
      AllNodes = @(
        @{
          NodeName = "localhost";
          CertificateID = 'A8DB81D8059F349F7EF19104399B898F701D4167'
        }
      )
    }
  EOH
end
```

I couldn't get it to work without specifying the certificate id (not sure if that's even possible, seems like it wouldn't be based on how one could execute the resource on any node). @smurawski may have more ideas there.

cc @adamedx @smurawski


adamedx:
@mcquin  is taking this on now. The certificate thing is something we hadn't anticipated -- I'd like to get that resolved, a key thing here is to avoid configurationdata altogether for this scenario.

adamedx:
This looks good, except that it needs a functional spec like the one described in our Jira ticket:

> An acceptance criteria would be a functional spec like the following which uses a credential in plain txt:
https://github.com/chef/chef/blob/12.5.0.current.0/spec/functional/resource/dsc_script_spec.rb#L317
except that it would not use plain text, instead PowerShell code like that below used in the example above would not use  "PSDscAllowPlainTextPassword = $true

jaym:
@adamedx functional test added

adamedx:
:+1:

ksubrama:
:+1:

adamedx:
Additional feedback: let's add some documentation to DOC_CHANGES.md that includes an example.

PR #3774, opened 2015-08-13
Lcg/yum deprecated

lamont-granquist:


lamont-granquist:
@chef/client-core @chef/enterprise-linux @thommay this adds support for the yum provider to use yum-deprecated if it detects it on the system.

users can do something like this up front in their cookbooks, and then the package/yum_package provider will work fine on fedora-22:

```ruby
if File.exist?("/usr/bin/dnf")
  execute "dnf install yum" do
    not_if { File.exist?("/usr/bin/yum-deprecated") }
  end
end
```

lamont-granquist:
Also for cookbooks that explicitly do `yum_package` on fedora-22, we're probably going to have to do something like this, unless we alias `yum_package` to the DNF provider (which i suppose we might have to do...)

lib/chef/resource/yum_package.rb
```diff
@@ -28,6 +28,7 @@ def initialize(name, run_context=nil)
         super
         @flush_cache = { :before => false, :after => false }
         @allow_downgrade = false
+        @yum_binary = ::File.exist?("/usr/bin/yum-deprecated") ? "yum-deprecated" : "yum"
```
danielsdeleo: Curious why this is in the resource rather than the provider.

danielsdeleo:
I have a curiosity question but :+1:

spec/unit/resource/yum_package_spec.rb
```diff
@@ -78,3 +78,22 @@
...
+    expect(::File).to receive(:exist?).with("/usr/bin/yum-deprecated").and_return(false)
+    expect(resource.yum_binary).to eql("yum")
+  end
+
+  it "defaults to yum-deprecated if yum-deprecated does exist" do
+    expect(::File).to receive(:exist?).with("/usr/bin/yum-deprecated").and_return(true)
+    expect(resource.yum_binary).to eql("yum-deprecated")
+  end
+
+  it "should allow you to specify whether allow_downgrade is true or false" do
```
jaymzh: the name of this test doesn't match with the test itseelf.

lib/chef/resource/yum_package.rb
lamont-granquist: Well it felt like it was a default value, but now that I think about it, that moves it to compile time rather than converge time so any resources that would install yum-deprecated would not have run yet, so I should probably move that...

lamont-granquist:
Okay that should address both comments.

danielsdeleo:
:+1:

jaymzh:
:+1:

stevendanna:
:+1:  LGTM

PR #3775, opened 2015-08-13
Enable 64-bit support for Powershell and Batch scripts

chefsalim:
Adding in support for 64-bit Ruby for Windows Powershell & Batch scripts.
Primarily, the change involves support for the :i386 architecture flag - which, if set when running under a 64-bit Ruby environment, will override the binary path for Powershell or cmd.exe to the appropriate SysWow64 location.

Sample test recipe:
powershell_script '64demo32' do
   code 'echo hello'
   architecture :i386
end

@adamedx @ksubrama @chef/client-windows

### DIFF
--- a/lib/chef/mixin/windows_architecture_helper.rb
+++ b/lib/chef/mixin/windows_architecture_helper.rb
@@ -19,18 +19,12 @@
 
 require 'chef/exceptions'
 require 'chef/platform/query_helpers'
-require 'win32/api' if Chef::Platform.windows?
-require 'chef/win32/api/process' if Chef::Platform.windows?
-require 'chef/win32/api/error' if Chef::Platform.windows?
+require 'chef/win32/process' if Chef::Platform.windows?
+require 'chef/win32/system' if Chef::Platform.windows?
 
 class Chef
   module Mixin
     module WindowsArchitectureHelper
-
-      if Chef::Platform.windows?
-        include Chef::ReservedNames::Win32::API::Process
-        include Chef::ReservedNames::Win32::API::Error
-      end
 
       def node_windows_architecture(node)
         node[:kernel][:machine].to_sym
@@ -40,6 +34,16 @@ class Chef
         desired_architecture == :x86_64 &&
           node_windows_architecture(node) == :x86_64 &&
           is_i386_process_on_x86_64_windows?
+      end
+
+      def forced_32bit_override_required?(node, desired_architecture)
+        desired_architecture == :i386 &&
+          node_windows_architecture(node) == :x86_64 &&
+          !is_i386_process_on_x86_64_windows?
+      end
+
+      def wow64_directory
+        Chef::ReservedNames::Win32::System.get_system_wow64_directory
       end
 
       def with_os_architecture(node, architecture: nil)
@@ -88,49 +92,21 @@ class Chef
 
       def is_i386_process_on_x86_64_windows?
         if Chef::Platform.windows?
-          is_64_bit_process_result = FFI::MemoryPointer.new(:int)
-
-          # The return value of IsWow64Process is nonzero value if the API call succeeds.
-          # The result data are returned in the last parameter, not the return value.
-          call_succeeded = IsWow64Process(GetCurrentProcess(), is_64_bit_process_result)
-
-          # The result is nonzero if IsWow64Process's calling process, in the case here
-          # this process, is running under WOW64, i.e. the result is nonzero if this
-          # process is 32-bit (aka :i386).
-          result = (call_succeeded != 0) && (is_64_bit_process_result.get_int(0) != 0)
+          Chef::ReservedNames::Win32::Process.is_wow64_process
         else
           false
         end
       end
 
       def disable_wow64_file_redirection( node )
-        original_redirection_state = ['0'].pack('P')
-
         if ( ( node_windows_architecture(node) == :x86_64) && ::Chef::Platform.windows?)
-          win32_wow_64_disable_wow_64_fs_redirection =
-            ::Win32::API.new('Wow64DisableWow64FsRedirection', 'P', 'L', 'kernel32')
-
-          succeeded = win32_wow_64_disable_wow_64_fs_redirection.call(original_redirection_state)
-
-          if succeeded == 0
-            raise Win32APIError "Failed to disable Wow64 file redirection"
-          end
-
+          Chef::ReservedNames::Win32::System.wow64_disable_wow64_fs_redirection
         end
-
-        original_redirection_state
       end
 
       def restore_wow64_file_redirection( node, original_redirection_state )
         if ( (node_windows_architecture(node) == :x86_64) && ::Chef::Platform.windows?)
-          win32_wow_64_revert_wow_64_fs_redirection =
-            ::Win32::API.new('Wow64RevertWow64FsRedirection', 'P', 'L', 'kernel32')
-
-          succeeded = win32_wow_64_revert_wow_64_fs_redirection.call(original_redirection_state)
-
-          if succeeded == 0
-            raise Win32APIError "Failed to revert Wow64 file redirection"
-          end
+          Chef::ReservedNames::Win32::System.wow64_revert_wow64_fs_redirection(original_redirection_state)
         end
       end
 
--- a/lib/chef/provider/batch.rb
+++ b/lib/chef/provider/batch.rb
@@ -28,6 +28,14 @@ class Chef
         super(new_resource, run_context, '.bat')
       end
 
+      def command
+        basepath = is_forced_32bit ? wow64_directory : run_context.node.kernel.os_info.system_directory
+
+        interpreter_path = Chef::Util::PathHelper.join(basepath, interpreter)
+
+        "\"#{interpreter_path}\" #{flags} \"#{script_file.path}\""
+      end
+
       def flags
         @new_resource.flags.nil? ? '/c' : new_resource.flags + ' /c'
       end
--- a/lib/chef/win32/api/system.rb
+++ b/lib/chef/win32/api/system.rb
@@ -187,6 +187,29 @@ int WINAPI GetSystemMetrics(
         safe_attach_function :GetSystemMetrics, [:int], :int
 
 =begin
+UINT WINAPI GetSystemWow64Directory(
+  _Out_ LPTSTR lpBuffer,
+  _In_  UINT   uSize
+);
+=end
+        safe_attach_function :GetSystemWow64DirectoryW, [:LPTSTR, :UINT], :UINT
+        safe_attach_function :GetSystemWow64DirectoryA, [:LPTSTR, :UINT], :UINT
+
+=begin
+BOOL WINAPI Wow64DisableWow64FsRedirection(
+  _Out_ PVOID *OldValue
+);
+=end
+        safe_attach_function :Wow64DisableWow64FsRedirection, [:PVOID], :BOOL
+
+=begin
+BOOL WINAPI Wow64RevertWow64FsRedirection(
+  _In_ PVOID OldValue
+);
+=end
+        safe_attach_function :Wow64RevertWow64FsRedirection, [:PVOID], :BOOL
+
+=begin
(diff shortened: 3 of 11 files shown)


lib/chef/mixin/windows_architecture_helper.rb
```diff
@@ -42,6 +42,24 @@ def wow64_architecture_override_required?(node, desired_architecture)
           is_i386_process_on_x86_64_windows?
       end
 
+      def forced_32bit_override_required?(node, desired_architecture)
```
adamedx: A functional spec for this one would be good, one that covered the 32-bit Ruby case and the 64-bit Ruby case, since this function would return true in one case and false in the other.

A unit test would also be good -- it might fit along with existing unit specs for this clasee.

lib/chef/provider/batch.rb
```diff
@@ -28,6 +28,14 @@ def initialize (new_resource, run_context)
         super(new_resource, run_context, '.bat')
       end
 
+      def command
+        basepath = is_forced_32bit ? wow64_directory : run_context.node.kernel.os_info.system_directory
+
+        interpreter_path = ::File.join(basepath, interpreter).gsub('/', '\\')
```
adamedx: Might be able to avoid the gsub by using one of the path helper functions like `Chef::Util::PathHelper.canonical_path` in https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/path_helper.rb.

lib/chef/mixin/windows_architecture_helper.rb
```diff
@@ -42,6 +42,24 @@ def wow64_architecture_override_required?(node, desired_architecture)
...
+        desired_architecture == :i386 &&
+          node_windows_architecture(node) == :x86_64 &&
+          !is_i386_process_on_x86_64_windows?
+      end
+
+      def wow64_directory
+        win32_get_system_wow64_directory = ::Win32::API.new('GetSystemWow64Directory', 'PI', 'I', 'kernel32')
```
jaym: We want to stop using Win32API stuff. Use FFI instead.

adamedx: Oh yeah, forgot about that.

adamedx:
looks good, just need to adjust the win32api to ffi thing.

chefsalim:
FFI conversion completed @adamedx @jaym @ksubrama

jaym:
LGTM. One day somebody will have to explain the ReservedNames namespace to me

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+    class System
+      include Chef::ReservedNames::Win32::API::System
+      extend Chef::ReservedNames::Win32::API::System
+
+      def self.get_system_wow64_directory
+        ptr = FFI::MemoryPointer.new :char, 255, true
+        succeeded = GetSystemWow64DirectoryA(ptr, 255)
+
+        if succeeded == 0
+          raise Win32APIError "Failed to get Wow64 system directory"
```
jaym: I think you need a comma after Win32APIError

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        buf = ptr.read_string
+        buf = buf.strip unless buf.strip.empty?
```
jaym: i'm not sure what is supposed to happen here.
why can't this line just be `buf.strip`

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      def self.wow64_disable_wow64_fs_redirection
+        original_redirection_state = FFI::MemoryPointer.new :pointer
+
+        succeeded = Wow64DisableWow64FsRedirection(original_redirection_state)
+
+        if succeeded == 0
+          raise Win32APIError "Failed to disable Wow64 file redirection"
```
jaym: missing comma

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+        if succeeded == 0
+          raise Win32APIError "Failed to get Wow64 system directory"
+        end
+
+        buf = ptr.read_string
+        buf = buf.strip unless buf.strip.empty?
+      end
+
+      def self.wow64_disable_wow64_fs_redirection
+        original_redirection_state = FFI::MemoryPointer.new :pointer
```
jaym: you have parens in other places but not here

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        original_redirection_state
+      end
+
+      def self.wow64_revert_wow64_fs_redirection(original_redirection_state)
+        succeeded = Wow64RevertWow64FsRedirection(original_redirection_state)
+
+        if succeeded == 0
+          raise Win32APIError "Failed to revert Wow64 file redirection"
```
jaym: missing comma

lib/chef/win32/api/system.rb
```diff
@@ -187,6 +187,29 @@ class OSVERSIONINFOEX < FFI::Struct
...
+=end
+        safe_attach_function :GetSystemWow64DirectoryW, [:pointer, :UINT], :UINT
+        safe_attach_function :GetSystemWow64DirectoryA, [:pointer, :UINT], :UINT
+
+=begin
+BOOL WINAPI Wow64DisableWow64FsRedirection(
+  _Out_ PVOID *OldValue
+);
+=end
+        safe_attach_function :Wow64DisableWow64FsRedirection, [:pointer], :BOOL
```
jaym: use PVOID instead of pointer.

lib/chef/win32/api/system.rb
```diff
@@ -187,6 +187,29 @@ class OSVERSIONINFOEX < FFI::Struct
         safe_attach_function :GetSystemMetrics, [:int], :int
 
 =begin
+UINT WINAPI GetSystemWow64Directory(
+  _Out_ LPTSTR lpBuffer,
+  _In_  UINT   uSize
+);
+=end
+        safe_attach_function :GetSystemWow64DirectoryW, [:pointer, :UINT], :UINT
```
jaym: use LPTSTR instead of pointer

lib/chef/win32/api/system.rb
```diff
@@ -187,6 +187,29 @@ class OSVERSIONINFOEX < FFI::Struct
...
+);
+=end
+        safe_attach_function :Wow64DisableWow64FsRedirection, [:pointer], :BOOL
+
+=begin
+BOOL WINAPI Wow64RevertWow64FsRedirection(
+  _In_ PVOID OldValue
+);
+=end
+        safe_attach_function :Wow64RevertWow64FsRedirection, [:pointer], :BOOL
```
jaym: use PVOID instead of pointer

lib/chef/win32/process.rb
```diff
@@ -69,6 +69,19 @@ def self.get_process_id(handle)
...
+        is_64_bit_process_result = FFI::MemoryPointer.new(:int)
+
+        # The return value of IsWow64Process is nonzero value if the API call succeeds.
+        # The result data are returned in the last parameter, not the return value.
+        call_succeeded = IsWow64Process(GetCurrentProcess(), is_64_bit_process_result)
+
+        # The result is nonzero if IsWow64Process's calling process, in the case here
+        # this process, is running under WOW64, i.e. the result is nonzero if this
+        # process is 32-bit (aka :i386).
+        result = (call_succeeded != 0) && (is_64_bit_process_result.get_int(0) != 0)
```
jaym: you dont need `result =`

lib/chef/win32/system.rb
```diff
@@ -0,0 +1,63 @@
...
+require 'ffi'
+
+class Chef
+  module ReservedNames::Win32
+    class System
+      include Chef::ReservedNames::Win32::API::System
+      extend Chef::ReservedNames::Win32::API::System
+
+      def self.get_system_wow64_directory
+        ptr = FFI::MemoryPointer.new :char, 255, true
```
jaym: parentheses?

chefsalim:
Good catch on the comma.. you are correct!

lib/chef/win32/system.rb
chefsalim: Yep, fixed!

lib/chef/win32/system.rb
chefsalim: Good question... I copied over some sample code for this line, now I don't remember exactly where I got that pattern from.  I'm fine with removing the buf.strip.empty? since the only case that looks like it could handle is a nil return, which shouldn't happen if the function succeeds.

lib/chef/win32/system.rb
chefsalim: Fixed

lib/chef/win32/system.rb
chefsalim: Fixing

lib/chef/win32/system.rb
chefsalim: Fixed

lib/chef/win32/api/system.rb
chefsalim: Fixing

lib/chef/win32/api/system.rb
chefsalim: Fixing

lib/chef/win32/api/system.rb
chefsalim: Fixing

lib/chef/win32/process.rb
chefsalim: Fixing

lib/chef/win32/system.rb
chefsalim: Fixing

jaym:
:+1:

lib/chef/mixin/windows_architecture_helper.rb
```diff
@@ -88,49 +92,21 @@ def assert_valid_windows_architecture!(architecture)
...
-
-          # The return value of IsWow64Process is nonzero value if the API call succeeds.
-          # The result data are returned in the last parameter, not the return value.
-          call_succeeded = IsWow64Process(GetCurrentProcess(), is_64_bit_process_result)
-
-          # The result is nonzero if IsWow64Process's calling process, in the case here
-          # this process, is running under WOW64, i.e. the result is nonzero if this
-          # process is 32-bit (aka :i386).
-          result = (call_succeeded != 0) && (is_64_bit_process_result.get_int(0) != 0)
+          Chef::ReservedNames::Win32::Process.is_wow64_process
```
ksubrama: Can be written more simply as Chef::Platform.windows? && Chef::ReservedNames::Win32::Process.is_wow64_process

lib/chef/mixin/windows_architecture_helper.rb
```diff
@@ -42,6 +36,16 @@ def wow64_architecture_override_required?(node, desired_architecture)
...
+        desired_architecture == :i386 &&
+          node_windows_architecture(node) == :x86_64 &&
+          !is_i386_process_on_x86_64_windows?
+      end
+
+      def wow64_directory
+        Chef::ReservedNames::Win32::System.get_system_wow64_directory
+      end
+
       def with_os_architecture(node, architecture: nil)
```
ksubrama: I'm a bit worried that with_os_architecture is left unmodified.  Both lib/chef/util/powershell/cmdlet.rb and lib/chef/mixin/powershell_out.rb  uses it to launch commands in the right architecture and this won't handle our 64/32 conversion.  It only fixes the providers.  I feel like we should perhaps refactor and re-architect all users of this mixin to not use with_os_architecture with its global redirection hammer and instead introduce a "get_system_dir(node, architecture: nil)" method which returns the correct prefix.  That prefix can then be used by each of these classes to correctly launch powershell, or regedit or cmd whatever it is they want to launch.

adamedx: Good idea for a subsequent change, though I don't think anything we're doing here makes this worse than before the change.

adamedx:
:+1:

ksubrama:
:+1:

PR #3776, opened 2015-08-14
Changes --hide-healthy to --hide-by-mins MINS

### LINKED ISSUE #3679: Change hardcoded 1hr in knife status -h
This is a pre-screen for a feature that I am committing to implementing. I don't want to do work that is just going to get rejected, so I'd like to start the conversation here.

`knife status` has:

        -H, --hide-healthy               Hide nodes that have run chef in the last hour

It's poor.

# Why?

1. For one, the definition of "healthy" is overloaded here to be "a host that has run chef". It makes no check in the code to determine if the chef run was successful, so it has the capability to provide false positives for "health". Just because a node object was saved with `ohai_time` set doesn't mean the Chef run was successful. There are exception handlers like `lastrun` that intentionally save the node object on exception.
2. The 1 hour hardcoded time is totally arbitrary. Perhaps "healthy" to others means "nodes that have run chef in the last 4 hours" (or 30 minutes, or 3 days...).

# Option 1: Least disruptive
Add an optional argument of "MINS". Not specifying it will result in the previous behavior of checking for "healthy" hosts where "healthy" means ohai_time was in the last hour. This does not solve "Why: 1" above. It also is odd in that it is taking an optional argument. I don't believe I've seen any other Chef command allow for that.

        -H, --hide-healthy MINS         Hide nodes that have run chef in the last hour (or MINS)

# Option 2: Fully corrected
This removes "health" and "healthy" from wording so does not set expectations. It keeps the short option as '-H'. It also requires MINS. Commence option name bikeshedding... in 3... 2... 1...

        -H, --hide-by-time MINS         Hide nodes that have run chef in the last MINS minutes.

Thoughts?

jblaine:
Fixes #3679

Reasoning:

The definition of "healthy" is overloaded in the old form
(--hide-healthy) to be "a host that has run chef". The code
makes no check to determine if the chef run was successful,
so it has the capability to provide false positives for
"health". Just because a node object was saved with ohai_time
set doesn't mean the Chef run was successful. There are
exception handlers like lastrun that intentionally save the node
object on exception.

The previous 1 hour hardcoded time was totally arbitrary. Perhaps
"healthy" to others means "nodes that have run chef in the last
4 hours" (or 30 minutes, or 3 days...).

### DIFF
--- a/lib/chef/knife/status.rb
+++ b/lib/chef/knife/status.rb
@@ -41,8 +41,12 @@ class Chef
         :long => "--sort-reverse",
         :description => "Sort the status list by last run time descending"
 
+      option :hide_healthy,
+        :short => "-H",
+        :long => "--hide-healthy",
+        :description => "Hide nodes that have run chef in the last hour. [DEPRECATED] Use --hide-by-mins MINS instead"
+
       option :hide_by_mins,
-        :short => "-H MINS",
         :long => "--hide-by-mins MINS",
         :description => "Hide nodes that have run chef in the last MINS minutes"
 
@@ -66,6 +70,14 @@ class Chef
         @query ||= ""
         append_to_query(@name_args[0]) if @name_args[0]
         append_to_query("chef_environment:#{config[:environment]}") if config[:environment]
+
+        if config[:hide_healthy]
+          ui.warn("-H / --hide-healthy is deprecated. Use --hide-by-mins MINS instead")
+          time = Time.now.to_i
+          # AND NOT is not valid lucene syntax, so don't use append_to_query
+          @query << " " unless @query.empty?
+          @query << "NOT ohai_time:[#{(time - 60*60).to_s} TO #{time.to_s}]"
+        end
 
         if config[:hide_by_mins]
           hidemins = config[:hide_by_mins].to_i
--- a/lib/chef/knife/status.rb
+++ b/lib/chef/knife/status.rb
@@ -41,10 +41,10 @@ class Chef
         :long => "--sort-reverse",
         :description => "Sort the status list by last run time descending"
 
-      option :hide_healthy,
-        :short => "-H",
-        :long => "--hide-healthy",
-        :description => "Hide nodes that have run chef in the last hour"
+      option :hide_by_mins,
+        :short => "-H MINS",
+        :long => "--hide-by-mins MINS",
+        :description => "Hide nodes that have run chef in the last MINS minutes"
 
       def append_to_query(term)
         @query << " AND " unless @query.empty?
@@ -67,11 +67,12 @@ class Chef
         append_to_query(@name_args[0]) if @name_args[0]
         append_to_query("chef_environment:#{config[:environment]}") if config[:environment]
 
-        if config[:hide_healthy]
+        if config[:hide_by_mins]
+          hidemins = config[:hide_by_mins].to_i
           time = Time.now.to_i
           # AND NOT is not valid lucene syntax, so don't use append_to_query
           @query << " " unless @query.empty?
-          @query << "NOT ohai_time:[#{(time - 60*60).to_s} TO #{time.to_s}]"
+          @query << "NOT ohai_time:[#{(time - hidemins*60).to_s} TO #{time.to_s}]"
         end
 
         @query = @query.empty? ? "*:*" : @query
--- a/spec/unit/knife/status_spec.rb
+++ b/spec/unit/knife/status_spec.rb
@@ -44,8 +44,8 @@ describe Chef::Knife::Status do
       @knife.run
     end
 
-    it "should filter healthy nodes" do
-      @knife.config[:hide_healthy] = true
+    it "should filter by nodes older than some mins" do
+      @knife.config[:hide_by_mins] = 60
       expect(@query).to receive(:search).with(:node, "NOT ohai_time:[1428569820 TO 1428573420]", opts)
       @knife.run
     end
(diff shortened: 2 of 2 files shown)


jblaine:
@chef/client-core @chef/client-maintainers needs review. Heads-up: There are 2 commits above with the same commit comment. This is because I did a 'git push' then code changes and 'git commit --amend' and 'git push --force' to my github branch. The commit showing as "0/2 checks OK" is actually solved by the 2nd commit's contents.

thommay:
can you do a `git rebase -i origin/master` or similar and squash the second commit, please? Other than that I am :+1:

jblaine:
There you go.

thommay:
cc @chef/client-maintainers

jonlives:
:+1:

lamont-granquist:
If someone is currently doing `--hide-healthy` though doesn't this break them, and shouldn't we keep the old long form around as a deprecated option until 13.0 can remove it?

thommay:
@lamont-granquist @jblaine it seems we can probably behave in a backwards compat manner by setting a default for `-H` of 60 mins? and aliasing `--hide-healthy` to `--hide-by-mins`?

jtimberman:
@lamont-granquist's concern was mine. I didn't know about this option, but I also haven't used `knife status` in awhile :smile_cat:. I think @thommay's idea is a good solution for the compatibility issue.

jblaine:
@thommay @lamont-granquist The problem there is that introduces what I believe to be a first in Chef: A CLI option that takes an optional value. With the current option parsing in place, I don't *believe* this is possible without doing something pretty ugly and unwanted.

Barring other ideas, this is what I know could be done:

1. Keep `--hide-healthy`
1. Allow `-H` to still be paired with `--hide-healthy` instead of carrying it over as the short option for the new `--hide-by-mins`
1. Keep the default for the above as 60 mins. **Flag both short+long as deprecated in help text as well as throw a `ui.warn("DEPRECATED...")` message if used.**
1. Add the new `--hide-by-mins MINS` which requires MINS

Acceptable? The best-right path?

lamont-granquist:
so you can always have both `config[:hide_healthy]` and `config[:hide_by_mins]` as being distinct and then have hide-by-mins default to 60, i don't think that's a problem.  you probably will need to keep -H associated with `--hide-healthy` without an arg until Chef 13 can break that.  so, i think yes, that's the right plan that you outlined...

jblaine:
Should probably remove the ready for merge tag until I get this fixed, but that's your call I guess :)

jblaine:
Should be all good now.
```
[@dev:~/src/chef-repo] master ± knife status -H
WARNING: -H / --hide-healthy is deprecated. Use --hide-by-mins MINS instead
2 hours ago, dev.home, centos 7.1.1503.
[@dev:~/src/chef-repo] master ± sudo chef-client
...
Chef Client finished, 4/71 resources updated in 10.685824342 seconds
[@dev:~/src/chef-repo] master ± knife status -H
WARNING: -H / --hide-healthy is deprecated. Use --hide-by-mins MINS instead
[@dev:~/src/chef-repo] master ± knife status
1 minute  ago, dev.home, centos 7.1.1503.
[@dev:~/src/chef-repo] master ± knife status --hide-by-mins 2
[@dev:~/src/chef-repo] master ±
```

lamont-granquist:
Yeah @chef/client-maintainers can we get eyeballs back on this and review the updated PR?

lib/chef/knife/status.rb
```diff
@@ -68,12 +72,21 @@ def run
         append_to_query("chef_environment:#{config[:environment]}") if config[:environment]
 
         if config[:hide_healthy]
+          ui.warn("-H / --hide-healthy is deprecated. Use --hide-by-mins MINS instead")
```
btm: I wonder if there's value in standardizing to using `Chef::Log.deprecation` in all these cases. It would present a unique opportunity to put a hook to do something when a deprecation is hit.

spec/unit/knife/status_spec.rb
```diff
@@ -44,8 +44,8 @@
       @knife.run
     end
 
-    it "should filter healthy nodes" do
-      @knife.config[:hide_healthy] = true
+    it "should filter by nodes older than some mins" do
+      @knife.config[:hide_by_mins] = 60
```
btm: given that hide_by_mins is now variable, it'd be good to make this 59 or 61 everywhere and update the tests accordingly so we're sure no code is still using the old `config[:hide_healthy]` which is fixed to 60.

btm:
It is kind of annoying that we can't have an optional argument with a default value.

lib/chef/knife/status.rb
jblaine: I'm sure there's value. I asked about a code style guide re: deprecation (at least) in `#chef-hacking` the other day and was informed that I'd have to bring it up during developer hours. Currently deprecation seems like it takes many flavors (based on grepping around).

I wasn't aware of `Chef::Log.deprecation`. I saw a `ui.warn` deprecation usage in some other knife code and rolled with that.

This seems like maybe something for a deprecation "get right" effort though, after there's style documentation on it so that it doesn't keep drifting.

lamont-granquist: Yeah, Chef::Log.deprecation is for core chef, I think we probably need to add a ui.deprecation method for knife to use, but nobody has hit that issue until now.

I wouldn't block on that.  As long as you do something where its clear enough to the user that its deprecated one way or another, then that should be good enough for now.

jblaine:
@btm Is e2ed6c48 what you meant?

PR #3778, opened 2015-08-14
Don't spit out newlines when nothing to puts()!

### LINKED ISSUE #3777: knife: Stop printing a blank line when no actual output to print.
This drives me nuts.

jblaine:
Fixes #3777 

The end of knife spurious newlines in output when nothing
needs to be displayed.

### DIFF
--- a/lib/chef/knife/core/ui.rb
+++ b/lib/chef/knife/core/ui.rb
@@ -66,7 +66,7 @@ class Chef
       # the logger API.
       def msg(message)
         begin
-          stdout.puts message
+          stdout.puts message unless message.empty?
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
@@ -76,7 +76,7 @@ class Chef
       # Prints a msg to stderr. Used for info, warn, error, and fatal.
       def log(message)
         begin
-          stderr.puts message
+          stderr.puts message unless message.empty?
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
@@ -156,7 +156,7 @@ class Chef
 
       def pretty_print(data)
         begin
-          stdout.puts data
+          stdout.puts data unless data.empty?
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
--- a/lib/chef/knife/core/ui.rb
+++ b/lib/chef/knife/core/ui.rb
@@ -66,7 +66,7 @@ class Chef
       # the logger API.
       def msg(message)
         begin
-          stdout.puts message unless message.empty?
+          stdout.puts message unless (message.respond_to?('empty?') && message.empty?)
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
@@ -76,7 +76,7 @@ class Chef
       # Prints a msg to stderr. Used for info, warn, error, and fatal.
       def log(message)
         begin
-          stderr.puts message unless message.empty?
+          stderr.puts message unless (message.respond_to?('empty?') && message.empty?)
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
@@ -156,7 +156,7 @@ class Chef
 
       def pretty_print(data)
         begin
-          stdout.puts data unless data.empty?
+          stdout.puts data
         rescue Errno::EPIPE => e
           raise e if @config[:verbosity] >= 2
           exit 0
--- a/lib/chef/knife/client_bulk_delete.rb
+++ b/lib/chef/knife/client_bulk_delete.rb
@@ -95,9 +95,9 @@ class Chef
       end
 
       def print_clients(clients)
-        ui.msg("")
+        ui.msg("\n")
         ui.msg(ui.list(clients.keys.sort, :columns_down))
-        ui.msg("")
+        ui.msg("\n")
       end
     end
   end
--- a/lib/chef/knife/configure.rb
+++ b/lib/chef/knife/configure.rb
@@ -105,17 +105,17 @@ EOH
           user_create.run
         else
           ui.msg("*****")
-          ui.msg("")
+          ui.msg("\n")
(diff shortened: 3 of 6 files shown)


lamont-granquist:
Looks like we had specs that tested this.

jblaine:
So someone went out of their way to write tests to confirm that incorrect newlines were present in the output. Heh. Okay.

chef-supermarket:

* @veetow
* @juliegund

jblaine:
How do I avoid having all of those older commits show up in my PR? They're from a 'git pull --rebase upstream master'. GH seems to have figured out the single relevant commit in my push, so is it just UI clutter and harmless?

lamont-granquist:
Uhm, hrm.  I dunno, but that doesn't look right.  

So looks like this will fix it:

```ruby
git rebase --onto 183f76a 54d3908~1
```

chef-supermarket:

lamont-granquist:
@chef/client-core @chef/client-maintainers needs review

danielsdeleo:
I think this is intentional behavior, so you can put empty lines in places. If you don't think there should be empty lines in a particular piece of command output, then we can debate that, but I don't think the behavior should be changed at the top level.

jblaine:
@danielsdeleo I think you misunderstand the code, or I misunderstand your comment. The code doesn't strip all newline-on-its-own from all knife output. It determines if there was ANY output to display, and if there was not, it displays nothing. Previously it would display a single newline and nothing else.

danielsdeleo:
I'm saying that there is probably usage like `ui.msg("")` where the intended outcome is to produce an empty line.

jblaine:
Ew.

```
[@dev:~/src/chef] knife-newlines 2 ± grep 'ui.msg("")' lib/chef/knife/* 2> /dev/null
lib/chef/knife/client_bulk_delete.rb:        ui.msg("")
lib/chef/knife/client_bulk_delete.rb:        ui.msg("")
lib/chef/knife/configure.rb:          ui.msg("")
lib/chef/knife/configure.rb:          ui.msg("")
lib/chef/knife/configure.rb:          ui.msg("")
lib/chef/knife/configure.rb:          ui.msg("")
lib/chef/knife/node_bulk_delete.rb:        ui.msg("")
lib/chef/knife/node_bulk_delete.rb:        ui.msg("")
lib/chef/knife/role_bulk_delete.rb:        ui.msg("")
lib/chef/knife/role_bulk_delete.rb:        ui.msg("")
[@dev:~/src/chef] knife-newlines 2 ±
```

jblaine:
The latest commit fixes the ui.msg("") cases and tests pass.

lib/chef/knife/core/ui.rb
```diff
@@ -66,7 +66,7 @@ def highline
       # the logger API.
       def msg(message)
         begin
-          stdout.puts message
+          stdout.puts message unless (message.respond_to?('empty?') && message.empty?)
```
stevendanna: I'm wondering if, instead of changing this method, we should change the `output` method below.  `output` is what *should* be called by all core knife plugin when printing an object while .msg is for more manual output control.

While you've changed all the call sites of `msg` in the core plugins, this change will still change the formatting of other plugins.

I agree that we should never print a single blank line, but the current change would change the output of a plugin such as the following:

```
class Foo < Chef::Knife
  def run
    ui.msg({})
  end
end
```

jblaine:
FWmyvoiceIW, I disagree with the idea and maintain that `ui.msg` should never print a blank line when an empty object is passed in: The current behavior of `ui.msg` is a bug, no matter how many people may have made use of its broken side-effect blank line.

By definition of its side-effect blank line, `ui.msg` is not for more manual output control as you're suggesting. Using it as-is, as seen in the core plugins for ~4 years with spurious blank lines constructed by `ui.msg("")`, demonstrates the *lack of control* it offers.

Control **is** `ui.msg("\n")` that would now be required to print a blank line.

stevendanna:
@jblaine First, I agree with you in the large.  Second, I'm sure it's getting frustrating to have to deal with so much review on what seems like such a simple change.  I'm sorry for that, our goal is to make improvements while providing people a stable platform to build on top of.

> Using it as-is, as seen in the core plugins for ~4 years with spurious blank lines constructed by ui.msg(""), demonstrates the lack of control it offers.

I said it offers "more control" than the output method which puts the input through a formatter based on the user-supplied `-F` flag.  As you can see the code, it is currently just a thin wrapper around `puts` that handle SIGPIPE so it has semantics that are understood by ruby programmers.

Anyway, I'm not here to block this improvement, but was rather offering a suggestion to how we could avoid breaking existing plugin code.  If other reviewers are OK with the plugin breakage (probably minimal?), I won't object to this solution.

jblaine:
I'm sorry for exposing my frustration. It's not directed at you. I see one *technically* right thing to do for the future of the code. If it means delaying this fix to to 13.0 in order to be nice, fair enough. That's just how I feel.

It turns out that people could be notified to fix their knife plugins *today* in prep for the future behavior. It appears that `puts` assumes a single `\n` in the input string is unwanted *in addition to* the one `puts` will spit out on its own. That is, since `puts` spits out its own newline, one could expect `puts "\n"` to print out two blank lines. It doesn't.

```
def msg(data)
  puts data
end
msg("") # old deprecated method to get a blank line
msg("\n") # new method to get a blank line
```

```
[@dev:~/src/chef/lib/chef/knife] knife-newlines ± ruby /tmp/test.rb


[@dev:~/src/chef/lib/chef/knife] knife-newlines ±
```
That's 2 blank lines, one due to `puts` appending a newline and another due to puts ignoring the explicitly passed in `\n`.

thommay:
I'm :+1: on this for Chef 13.

lamont-granquist:
yep :+1:  for 13  (:sigh:)

thommay:
So we tried this in #5852 but it proved excessively hard; I think for right now we're going to close this and if we want to take another run at this in the Chef 15 cycle we can.
Sorry @jblaine that this did not get done.

PR #3787, opened 2015-08-17
Fix failing directory unit tests on rhel

jaym:
do_selinux needs to be mocked out so it does not actually do anything.
Otherwise, it will be available on rhel and the tests are no longer
unit and will fail


lamont-granquist:
:+1:

PR #3788, opened 2015-08-17
Don't modify members of new_resource in pw group implmentation

jaym:
This was causing the

```
Chef::Resource::Group
  group modify action
    when there is a group
      behaves like correct group management
        when the users exist
          when append is not set
            when group already contains some users
              should remove all existing users and only add the new users to the group
```
spec to fail. The test passes an array which gets modified which breaks the test.


lamont-granquist:
Yep, never modify the new_resource :+1:

ksubrama:
Thank you!

PR #3797, opened 2015-08-19
Fix dsc_script spec failure on 64-bit Ruby

chefsalim:
dsc_script_spec.rb currently fails under 64-bit Ruby. The fix here is to make sure we ensure that the node architecture for the test is always set to 64-bit (since dsc is not supported on 32-bit Windows).  

@adamedx @jaym

### DIFF
--- a/spec/functional/resource/dsc_script_spec.rb
+++ b/spec/functional/resource/dsc_script_spec.rb
@@ -67,8 +67,7 @@ describe Chef::Resource::DscScript, :windows_powershell_dsc_only do
     node = Chef::Node.new
     node.automatic['platform'] = 'windows'
     node.automatic['platform_version'] = '6.1'
-    node.automatic['kernel'][:machine] =
-      is_i386_process_on_x86_64_windows? ? :x86_64 : :i386
+    node.automatic['kernel'][:machine] = :x86_64  # Only 64-bit architecture is supported
     node.automatic[:languages][:powershell][:version] = '4.0'
     empty_events = Chef::EventDispatch::Dispatcher.new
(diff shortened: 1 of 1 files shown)


jaym:
:+1:

spec/functional/resource/dsc_script_spec.rb
```diff
@@ -67,8 +67,7 @@ def delete_user(target_user)
     node = Chef::Node.new
     node.automatic['platform'] = 'windows'
     node.automatic['platform_version'] = '6.1'
-    node.automatic['kernel'][:machine] =
-      is_i386_process_on_x86_64_windows? ? :x86_64 : :i386
+    node.automatic['kernel'][:machine] = :x86_64  # Only 64-bit process is supported
```
adamedx: So isn't the key here that we only support running these specs on a 64-bit OS vs. a 64-bit process? The specs themselves have always been 32-bit to this point, but Chef (also 32-bit) shells out to 64-bit).

My note here is only about the comment, basically I think it's more accurate to say 64-bit OS instead of 64-bit process. Does that seem reasonable?

chefsalim: The failure case was under a 64-bit OS - we would specify a i386
architecture if NOT running as a 32-bit process under a 64-bit OS.  When we
ran under 32-bit, that was as a 64-bit process as you indicate.  So I think
process is correct.  Maybe we can say "62-bit architecture" instead?

chefsalim:
Updated comment to say 'architecture'

adamedx:
The existential questions of 64-bitness have been answered. :+1:

PR #3798, opened 2015-08-19
use group name instead of sid for `current_resource` in the group resource on windows

### LINKED ISSUE #3411: Group resource on windows is reporting inconsistent member changes
The group resource used on a windows machine appears to be looking at a different id to gather initial state, which makes accurate before/after diffing of changed users challenging.

```
{
  "chef_server_fqdn": "manage.blah",
  "end_time": "2015-04-30T23:29:41Z",
  "id": "8532ef19-4edb-455f-b5ec-34809d6845b3",
  "message_type": "run_converge",
  "message_version": "0.1.0",
  "node_name": "opscode1",
  "organization_name": "opscode",
  "run_id": "d110a65d-6063-4844-976c-9d25bd9e3377",
  "run_list": [
    "recipe[wingroup]"
  ],
  "start_time": "2015-04-30T23:29:39Z",
  "status": "success",
  "total_resource_count": 10,
  "resources": [
    {
      "cookbook_name": "wingroup",
      "cookbook_version": "0.1.3",
      "delta": "",
      "duration": "7",
      "final_state": {
        "members": [
          "Administrator",
          "random"
        ]
      },
      "initial_state": {
        "members": [
          "S-1-5-21-294568438-2080408112-401244387-500",
          "S-1-5-21-294568438-2080408112-401244387-1002",
          "S-1-5-21-294568438-2080408112-401244387-501"
        ]
      },
      "resource_id": "Administrators",
      "resource_name": "Administrators",
      "resource_result": "modify",
      "resource_type": "group"
    }
  ],
  "updated_resource_count": 1
}
```

jaym:
Using the SID makes things look not so nice from a reporting perspective
Fixes #3411

### DIFF
--- a/lib/chef/provider/group/windows.rb
+++ b/lib/chef/provider/group/windows.rb
@@ -39,7 +39,9 @@ class Chef
 
           members = nil
           begin
-            members = @net_group.local_get_members
+            members = current_group_user_sids.map do |sid|
+              sid.account_name
+            end
           rescue => e
             @group_exists = false
             Chef::Log.debug("#{@new_resource} group does not exist")
@@ -81,7 +83,7 @@ class Chef
 
         def has_current_group_member?(member)
           member_sid = local_group_name_to_sid(member)
-          @current_resource.members.include?(member_sid)
+          current_group_user_sids.any?{|s| s.to_s == member_sid}
         end
 
         def remove_group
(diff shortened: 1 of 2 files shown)


thommay:
ping @chef/client-windows for review please!

lamont-granquist:
My concern would be that you're changing the semantics of the field and maybe introducing issues comparing between group names and sids between new_resource and current_resource.  Generally we follow the pattern that if they can be different types that the current_resource must load the same type as the new_resource specified so that the comparison worked.

jaym:
@lamont-granquist if you take a look at the issue, it seems that new_resource is using a human readable name, and current_resource is using the sids. Before, the compairson happened by converting `new_resource` to sids. Now, we keeps the name and sid for `current_resource`, but still compare everything with sids. (I think, it's been a while)

lamont-granquist:
Okay, LGTM :+1:

lamont-granquist:
@chef/client-core @chef/client-windows needs another :+1:

adamedx:
So I'm not sure this is safe. If you convert the sid to a name, it's not clear to me that it will be unique. does the name include the domain? And even across forests, you could have the same name, so unless the name format is UPN, you could hit issues when there's a cross-forest trust.

If this is just reporting, can't we just change what reporting outputs? Alternatively, could we use something that contains both the username AND the sid? Then it's human readable, but we won't hit name collisions in the idempotence logic.

thommay:
Closing due to old age.

PR #3799, opened 2015-08-20
fix supports hash issues in service providers

### LINKED ISSUE #2168: Can't disable status check for redhat init service
I seem to be running into an odd issue on creating services for RedHat (specifically, CentOS 6.4+ and Amazon Linux)

Given a simple template for a chkconfig-aware service, and a "service" block, the behavior observed is that the service is enabled on start runlevels, but not on the stop runlevels, nor is the service "started" at converge time, but rather only on reboot time.

Testing the same recipe on Ubuntu worked as expected (enabled in start and stop runlevels + run on converge)

```
#!/bin/bash
# chkconfig: 2345 90 10
# desription: test script

. /etc/init.d/functions
my_log="/var/log/testmememe.log"

case "$1" in
        stop)
	echo "STOPPED" >> "$my_log"
	echo "OK"
        ;;
        start)
	echo "STARTED" >> "$my_log"
	echo "OK"
        ;;
*)
        echo $"Usage: $0 {start|stop}"
        exit
esac
```

```
# init.d template
template "/etc/init.d/test" do
  source "test.erb"
  mode 0755
end

service "test" do
   supports :start => true, :stop => true, :restart => false
   priority 90
   action [:enable, :start]
end
```

After converging ```/var/log/testmememe``` does not exist.  After reboot, it contains the single word "STARTED"

Issuing ```chkconfig --add test``` on a converged machine does add the service to stop runlevels.

lamont-granquist:
- redhat provider now allows the user to override :status
- gentoo provider now allows the user to override :status and :restart
- service providers now dup the status hash and mutate their private
  copy instead of mutating the new_resource

closes #2168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   [pr#3704](https://github.com/chef/chef/pull/3704): Add SIP (OS X 10.11) support
 * [**Phil Dibowitz**](https://github.com/jaymzh):
 
+* [pr#3799](https://github.com/chef/chef/pull/3799) fix supports hash issues in service providers
 * [pr#3817](https://github.com/chef/chef/pull/3817) Remove now-useless forcing of ruby Garbage Collector run
 * [pr#3805](https://github.com/chef/chef/pull/3805) LWRP parameter validators should use truthiness
 * [pr#3774](https://github.com/chef/chef/pull/3774) Add support for yum-deprecated in yum provider
--- a/lib/chef/provider/service.rb
+++ b/lib/chef/provider/service.rb
@@ -1,6 +1,6 @@
 #
 # Author:: AJ Christensen (<aj@hjksolutions.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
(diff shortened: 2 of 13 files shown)


lamont-granquist:
@chef/client-core thought this would be nearly a one-liner but as usual...

I had to switch to using nil default values in the resource in order to distinguish true/false-by-default from true/false-by-user-intention.  Fixed the mutation of the new_resource in the process.

Had to move mutating the supports hash into load_current_resource so that the new_resource ivar got set first by the initialize method, otherwise when the redhat provider tried to set its (higher priority) default before calling super the `@new_resource` would not be set.

lamont-granquist:
@chef/client-core ping

danielsdeleo:
:+1: 

I would love to eliminate this pattern entirely and just have these be non-state attributes like `supports_status(false)` but we have a bug to fix right now and this fixes it.

lamont-granquist:
@chef/client-core needs another :+1:

thommay:
:+1:

PR #3805, opened 2015-08-20
Validating is comparing to true instead of ruby truthiness

jaymzh:
It should use truthiness.


lamont-granquist:
:+1:

coderanger:
:+1:

jaymzh:
added docs. merging.

Issue #1533, opened 2014-06-23
CHEF-2707 subversion resource doesn't work on windows


cixelsyd:
Hi Chefs.  This issue (reference original bug inside jira https://tickets.opscode.com/browse/CHEF-2707 ) is still active.

I utilize the subversion provider heavily under windows.  The impact of this bug is huge for me, and requires a hack for this for each release of chef-client:

```
C:\opscode\chef\embedded\lib\ruby\gems\1.9.1\gems\chef-xxxxx-x86-mingw32\lib\chef\mixin\command.rb
```

The hack that I had previously put into place:

```
unless Array[opts[:returns], 42].include?(status.exitstatus)
```

..no longer works.  Could you please upgrade the priority on resolving this, please? I now have to revert chef-client to a version using an insecure openssl to re-apply my previous hack.

The error I receive with chef-client version 11.12.8 is:

```
NoMethodError: undefined method `exitstatus' for nil:NilClass
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/mixin/command.rb:140:in `block in output_of_command'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/mixin/command.rb:122:in `chdir'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/mixin/command.rb:122:in `output_of_command'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/provider/subversion.rb:132:in `revision_int'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/provider/subversion.rb:117:in `export_command'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/provider/subversion.rb:81:in `block in action_force_export'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/mixin/why_run.rb:52:in `call'
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.8-x86-mingw32/lib/chef/mixin/why_run.rb:52:in `add_action'
[...]

PR #3806, opened 2015-08-21
Replace output_of_command with shell_out! in subversion provider

gh2k:
output_of_command doesn't work properly on Windows (see chef/chef#1533)

### DIFF
--- a/lib/chef/provider/subversion.rb
+++ b/lib/chef/provider/subversion.rb
@@ -130,8 +130,8 @@ class Chef
             @new_resource.revision
           else
             command = scm(:info, @new_resource.repository, @new_resource.svn_info_args, authentication, "-r#{@new_resource.revision}")
-            status, svn_info, error_message = output_of_command(command, run_options)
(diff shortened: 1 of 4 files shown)


gh2k:
I'm not sure if it's better to break out the "svn.exe" changes into a separate commit. This may be related to a bug somewhere else, as I notice other providers (such as 'git') don't do this.

I'm not sure what the cause of this is, but without the '.exe', calls to svn on Windows explode with a 'CreateProcessW' I/O error.

lamont-granquist:
could you add svn_binary to the resource as a tweakable and then just default in the provider to those values?

I did something similar here with yum_binary (but with some additional complexity due to the nature of package providers):

https://github.com/chef/chef/pull/3774/files

gh2k:
@lamont-granquist This any better? I've played around with it in my test environment and it seems OK

lamont-granquist:
Yep, LGTM :+1:

lamont-granquist:

tyler-ball:
:+1:

danielsdeleo:
LGTM :+1:

gh2k:
Thanks for the merge

PR #3808, opened 2015-08-21
Fix win32 error code handling in NamedSecurityInfo

Annih:
Hello,

In certain case, NameSecurityInfo methods can fail and return an HR result diffrent from `ERROR_SUCCESS`; current implementation call `Chef::ReservedNames::Win32::Error.raise!` which tries to get the last error code.
But often the last error code has changed in the meantime... resulting to that kind of error message:
> ---- Begin Win32 API output ----
> System Error Code: 0
> System Error Message: The operation completed successfully.
> ---- End Win32 API output ----

So this PR update the `raise!` method to allow to pass the error code, and use this new feature in some methods that directly get the HR error code.

### DIFF
--- a/lib/chef/win32/error.rb
+++ b/lib/chef/win32/error.rb
@@ -57,8 +57,8 @@ class Chef
       # nil::: always returns nil when it does not raise
       # === Raises
       # Chef::Exceptions::Win32APIError:::
-      def self.raise!(message = nil)
-        code = get_last_error
+      def self.raise!(message = nil, code = nil)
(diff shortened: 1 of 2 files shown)


smurawski:
@Annih We'd like to get this merged but a test to validate the use would be helpful.  If you can add a test, that'll help us get this merged more quickly.

Aliasgar16:
@Annih I have cherry-picked the required commits from this PR into the [PR](https://github.com/chef/chef/pull/5727).

Issue #3811, opened 2015-08-21
SmartOS package upgrade selecting wrong version


hulu1522:
When you upgrade a package it is not selecting the most current version.  Example is when installing `nodejs`.  It went from version `0.12.3` to `0.10.35`.  That is a downgrade.

```
Recipe: base::node
         * smartos_package[nodejs] action install (up to date)
         * smartos_package[nodejs] action upgrade
       - upgrade package nodejs from 0.12.3 to 0.10.35
```

As you can see, Chef logically selected the last version in the list and not the latest:

```
[root@75e226e9-d915-cac6-dc0b-f42aad24d713 ~]# pkgin se nodejs
nodejs-0.8.28 >      V8 JavaScript for clients and servers
nodejs-0.12.7 <      V8 JavaScript for clients and servers
nodejs-0.12.6 <      V8 JavaScript for clients and servers
nodejs-0.12.4 <      V8 JavaScript for clients and servers
nodejs-0.12.3 =      V8 JavaScript for clients and servers
nodejs-0.12.2 >      V8 JavaScript for clients and servers
nodejs-0.12.1 >      V8 JavaScript for clients and servers
nodejs-0.12.0 >      V8 JavaScript for clients and servers
nodejs-0.10.40 >     V8 JavaScript for clients and servers
nodejs-0.10.39 >     V8 JavaScript for clients and servers
nodejs-0.10.38 >     V8 JavaScript for clients and servers
nodejs-0.10.36 >     V8 JavaScript for clients and servers
nodejs-0.10.35 >     V8 JavaScript for clients and servers
```

hulu1522:
The version log is [here](https://github.com/chef/chef/blob/master/lib/chef/provider/package/smartos.rb#L44-L58)

hulu1522:
UPDATE: This is not a specific bug of Chef but is instead a bug with `pkgin` because it is sorting by string and not version.

Here is the fix we found that helps:
```
[root@75e226e9-d915-cac6-dc0b-f42aad24d713 ~]# pkgin se nodejs | grep '<' | sort
<: package is installed but newer version is available
nodejs-0.12.4;<;V8 JavaScript for clients and servers
nodejs-0.12.6;<;V8 JavaScript for clients and servers
nodejs-0.12.7;<;V8 JavaScript for clients and servers
```

PR #3812, opened 2015-08-21
Fix for package upgrading

hulu1522:
`pkgin` has an issue where it sorts packages by string.  This fix will allow upgrading to happen successfully.

### DIFF
--- a/lib/chef/provider/package/smartos.rb
+++ b/lib/chef/provider/package/smartos.rb
@@ -65,7 +65,7 @@ class Chef
           pkg.stdout.each_line do |line|
             case line
             when /^#{new_resource.package_name}/
-              name, version = line.split(/[; ]/)[0].split(/-([^-]+)$/)
+              name, version = line.split(/[; ]/)[0].split(/-([^-]+)$/) if line =~ /.*[<=].*/
             end
           end
           @candidate_version = version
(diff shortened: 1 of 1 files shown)


hulu1522:
Fix for #3811

hulu1522:
Don't know how to fix the tests but my image only has varnish `3.0.5` and your test is failing on varnish `2.3.4`.

```
~# pkgin se varnish
varnish-3.0.5nb1     High-performace HTTP accelerator

=: package is installed and up-to-date
<: package is installed but newer version is available
>: installed package has a greater version than available package
```

lamont-granquist:
Those are unit tests which run on Linux and Windows (which definitely doesn't have varnish installable via pkgin), so the tests must be stubbing the underlying pkgin calls.  Something in your patch has broken the unit tests and needs to get fixed (likely the unit tests were incorrect and also need fixing for the new behavior).

lamont-granquist:
This looks like it'd be easy enough to salvage, but we just don't have the time to do it ourselves.

hulu1522:
I apologize for not responding to this.  I just don't know enough about the unit testing to be of assistance.  The main goal of the request was to organize the results so the latest version would be on bottom.  By default Chef will choose the last package in the list even if the version is not the latest.  I also don't have time to help solve this so I guess close it as a 'known issue' or hope someone else can pick it up.  Thanks.

PR #3815, opened 2015-08-24
Fix powershell_script validation 

jaym:
Powershell script validation did not use same architecture as
executing the script. This caused configuring the LCM to fail.


chefsalim:
The spec change looks like it is in the 32-bit Ruby section.. we probably want to pull it out so that it applies in either the 32-bit and 64-bit Ruby case.

adamedx:
moving the test to run on all platforms would be great as suggested above.

jaym:
@adamedx @chefsalim updated

PR #3817, opened 2015-08-24
remove now-useless GC

lamont-granquist:


lamont-granquist:
@chef/client-core since we never run intervals now in the same process this is just a waste of time now since the process will always subsequently exit?

coderanger:
:+1:

danielsdeleo:
I'm pretty sure your reasoning is correct. :+1:

ranjib:
:+1:

PR #3818, opened 2015-08-25
[RFC-051] Telemetry system

ranjib:
:construction: 
This is first pass towards implementing [RFC-051](https://github.com/chef/chef-rfc/blob/master/rfc051-telemetry.md). It contains the main structural components, like the API for adding custom metrics, and couple of built in metrics (time spent on resource, recipe, cookbook etc).
If i get a review of this it will be easier to implement the rest (or i dont have to rewrite a lot of things after implementing, as rest of the components will reuse this work.

Things that are still left are:
- [] statsd based publishing mechanism
- [] node attribute based publishing mechanism
- [] GC based metrics
- [] Chef client run metrics
- [] process based metrics (only for *nix platform)

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -378,7 +378,6 @@ class Chef::Application::Client < Chef::Application
 
     if !Chef::Config[:client_fork] || Chef::Config[:once]
       begin
-        Chef::Telemetry.init if Chef::Telemetry.enabled?
         # run immediately without interval sleep, or splay
         run_chef_client(Chef::Config[:specific_recipes])
       rescue SystemExit
@@ -407,7 +406,6 @@ class Chef::Application::Client < Chef::Application
         end
 
         @signal = nil
-        Chef::Telemetry.init if Chef::Telemetry.enabled?
         run_chef_client(Chef::Config[:specific_recipes])
 
         Chef::Application.exit!("Exiting", 0) if !Chef::Config[:interval]
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -25,7 +25,6 @@ require 'chef/log'
 require 'chef/config_fetcher'
 require 'chef/handler/error_report'
 require 'chef/workstation_config_loader'
-require 'chef/telemetry'
 
 class Chef::Application::Client < Chef::Application
   include Chef::Mixin::ShellOut
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -25,6 +25,7 @@ require 'chef/log'
 require 'chef/config_fetcher'
 require 'chef/handler/error_report'
 require 'chef/workstation_config_loader'
+require 'chef/telemetry'
 
 class Chef::Application::Client < Chef::Application
   include Chef::Mixin::ShellOut
@@ -36,6 +37,12 @@ class Chef::Application::Client < Chef::Application
     :short => "-c CONFIG",
     :long  => "--config CONFIG",
     :description => "The configuration file to use"
+
+  option :enable_telemetry,
+    :short => "-T",
+    :long  => "--enable-telemetry",
+    :default => false,
+    :description => "Enable chef telemetry subsystem"
 
   option :formatter,
     :short        => "-F FORMATTER",
@@ -371,6 +378,7 @@ class Chef::Application::Client < Chef::Application
 
     if !Chef::Config[:client_fork] || Chef::Config[:once]
       begin
+        Chef::Telemetry.init if Chef::Telemetry.enabled?
         # run immediately without interval sleep, or splay
         run_chef_client(Chef::Config[:specific_recipes])
       rescue SystemExit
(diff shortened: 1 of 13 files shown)


ranjib:
cookbook, recipe, resource time spent metrics are largely based on @joemiller 's work.

lamont-granquist:
I have a rewrite of node attributes in progress, so you probably want to stay away from any telemetry of attributes until that lands (at which point it will get much easier).

ranjib:
@lamont-granquist ack. i have not added anything related to attribute tracing (no plan do it any time soon). any other comment on the PR?

lib/chef/telemetry.rb
```diff
@@ -0,0 +1,76 @@
+#
+# Author:: Ranjib Dey (<ranjib@linux.com>)
+# Copyright:: Copyright (c) 2015 Chef Inc.
```
coderanger: While you are free to assign copyright to Chef Software Inc. (though you should fix the name), you are under no obligation to do so as this was not produced under work-for-hire that I'm aware of. If this was produced at $DAYJOB you should check if they retain copyright, otherwise by default it stays with you personally.

coderanger:
General comment, I would love to see YARD docstring comments on the new APIs.

lib/chef/application/client.rb
```diff
@@ -37,6 +37,12 @@ class Chef::Application::Client < Chef::Application
     :long  => "--config CONFIG",
     :description => "The configuration file to use"
 
+  option :enable_telemetry,
+    :short => "-T",
+    :long  => "--enable-telemetry",
```
coderanger: Seems like we could make a slightly more streamlined UX here. Thinking something like this: `-T`== `-T log` == `--telemetry` == `--telemetery log`. That would allow configuring the processor(s?) via the command line too.

lib/chef/telemetry/processor.rb
```diff
@@ -0,0 +1,71 @@
...
+          publisher.publish(@metrics)
+        end
+      end
+
+      def self.create(publishers)
+        processor = Chef::Telemetry::Processor.new
+        publishers.each do |publisher|
+          processor.add_publisher(publisher)
+        end
+        Chef.event_handler do
```
coderanger: Makes me nervous using a global value inside the API like this. Should at least be overridable via method args.

lib/chef/telemetry.rb
ranjib: not a day job thing. dont worry. i am assigning this to chef inc, knowingly :-)

lib/chef/application/client.rb
ranjib: im implementing the RFC as it is. I dont want the --telmetry log discussion here. CLI only enabled/disable it. use config options to set specific publishing mechanism.

lib/chef/telemetry.rb
coderanger: There is no such company as "Chef Inc", it is "Chef Software, Inc.". </legal pedant>

lib/chef/telemetry/processor.rb
ranjib: why? All Chef.* class based API are global, and they are the public facing API. It also ensures that event handler registration done properly. Event handler was implemented to ease this type of use cases. Can you provide an example on what would be better?

I can move this logic to a dedicated event handler, similar to doc and other formatter, and then register its if telemetry is enabled, similar to formatter injection. Is that what you looking for? This is similar to that but implemented using the event handler API

coderanger: Thinking about cases where you want to write tests for a publisher and you want to use a non-global event dispatcher. Also possibly with stuff like Poise's internal converges. In general it is ungood to have an API use a global variable with no way to pass in a non-global version. A simple fix would be to make it `def self.create(publishers, event_handler=Chef.event_handler)` so that a caller can pass in their own object as needed.

thommay:
(we're putting this in help wanted to ease with PR triage, not for any other reason)

lib/chef/telemetry/processor.rb
```diff
@@ -0,0 +1,71 @@
...
+
+
+#
+# Telemetry::Processor class provides API for chef run related metrics gathering
+#
+
+class Chef
+  module Telemetry
+    class Processor
+      class Metric < Struct.new(:name, :hook, :value); end
```
lamont-granquist: Inheriting from Struct.new is better done for reasonably otherwise empty classes, based on the weight of everything else you should probably be explicit here.

danielsdeleo: This is just a 3-item struct, the rest of this is `Chef::Telemetry::Processor`.

lamont-granquist: Ah, yeah, that'd be why I sorta don't like the `; end` there...

lib/chef/telemetry/processor.rb
lamont-granquist: Yeah since this is a class method it should take an arg to make that injectable and reduce tight coupling.

thommay:
hey @ranjib , are you gonna come back to this? Or how can we help you here?

thommay:
Hey,
per our RFC on workflow, I'm going to close this PR since it's been idle for a year. We'd love to see this get done though, so please do reopen this when convenient!
-Thom

PR #3819, opened 2015-08-25
Don't use routes with action :delete when writing config

petracvv:
When I was writing a cookbook that had both route deletions and additions, I noticed that the route-eth0 file written by chef had **all** the route information from **all** the route resources I had defined no matter the type of action. 

I added an additional check to only include the :add route resources.

### DIFF
--- a/lib/chef/provider/route.rb
+++ b/lib/chef/provider/route.rb
@@ -176,7 +176,7 @@ class Chef::Provider::Route < Chef::Provider
             conf[dev] = String.new if conf[dev].nil?
             case @action
             when :add
-              conf[dev] << config_file_contents(:add, :target => resource.target, :netmask => resource.netmask, :gateway => resource.gateway)
+              conf[dev] << config_file_contents(:add, :target => resource.target, :netmask => resource.netmask, :gateway => resource.gateway) if resource.action == [:add]
             when :delete
               # need to do this for the case when the last route on an int
               # is removed


lamont-granquist:
Can you spell out your use case in a little more detail?  We can't quite parse exactly what the issue is and the code look a bit odd, and it smells like you're probably solving your problem in a fairly brittle way (inspection of new_resource.action is code smell).

petracvv:
Sure. The route resource implementation is a bit weird as it inspects the entire resource collection for "route" resources and generates the route file to commit to disk every time it is called. However, it looks through **all** the route resources, including the ones with action delete. In my case I was deleting the default gateway and replacing it with a different one:
```ruby
route 'delete_eth0_gate' do
  target '0.0.0.0/0'
  action :delete
  device 'eth0'
  gateway '10.10.10.1'
end

route 'add_eth1_gateway' do
  target '0.0.0.0/0'
  action :add
  device 'eth1'
  gateway '10.10.11.1'
end
```
Even though I was not adding a route to eth0, it still wrote a route-eth0 file (in CentOS) with the `0.0.0.0/0 via 10.50.2.1` line inside. After looking at the code, it seems to have done that from the 'add_eth1_gateway' resource because it had an `:add` action. The case statement checks the action of the calling resource, **not** the one it was iterating through. 

The if statement I added seemed to be the only way of making sure that does not happen with the current implementation of the route resource in Chef.

lamont-granquist:
:+1:

thommay:
The provider is horrific, but the code is correct. :+1:

petracvv:
Looks like I didn't see this until now... Thanks for the merge!

PR #3820, opened 2015-08-25
Add RELEASE_NOTES entry for `knife rehash`

stevendanna:


stevendanna:
cc @jaym

jaym:
:+1:

coderanger:
Can we make `chef gem` run this after all commands? That should make it very rare that a user would need to do it manually.

stevendanna:
@coderanger Yes, I believe that is the plan.

PR #3821, opened 2015-08-25
Human friendly elapsed time in log

joelhandwell:
This pull request will convert elapsed time from seconds to minutes and seconds if it's more than 60 seconds, convert it to hours, minutes and seconds if it's more than 3600 seconds to contribute for more human readable output.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ group(:development, :test) do
   gem "simplecov"
   gem 'rack', "~> 1.5.1"
   gem 'cheffish', "~> 1.3"
+  gem 'timecop'
 
   gem 'ruby-shadow', :platforms => :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
 end
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,6 @@ group(:development, :test) do
   gem "simplecov"
   gem 'rack', "~> 1.5.1"
   gem 'cheffish', "~> 1.3"
-  gem 'timecop'
 
   gem 'ruby-shadow', :platforms => :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
 end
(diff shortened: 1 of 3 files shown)


joelhandwell:
Vagrant chef-zero provisioner's elapsed time got longer ( https://github.com/mitchellh/vagrant/issues/5972 ) and that led to this PR.

That said,
```
Chef Client failed. 9 resources updated in 40 minutes 35 seconds
``` 
would be more human friendly than
```
Chef Client failed. 9 resources updated in 2435.332318955 seconds
```

stevendanna:
I like this change.  I would suggest keeping `elapsed_time` as a function that returns `end_time - start_time` and moving the pretty printing into `pretty_elapsed_time`.  That'll be friendlier to formatters that might be subclasses of this and future users of this class.

joelhandwell:
@stevendanna ```pretty_elapsed_time``` was created and formatting was moved over there.

stevendanna:
:+1: here

stevendanna:
cc @chef/client-maintainers

Gemfile
```diff
@@ -14,6 +14,7 @@ group(:development, :test) do
   gem "simplecov"
   gem 'rack', "~> 1.5.1"
   gem 'cheffish', "~> 1.3"
+  gem 'timecop'
```
sethvargo: I'm fairly against adding timecop as a dependency just for testing. I think we can stub the time easily just using RSpec.

mcquin: It could at least be a development dependency, right?

sethvargo: It could be moved to a dev dep, but I don't see the need for it since it brings in another layer of indirection and a _ton_ of monkey patches into the Time and DateTime classes that could affect other parts of the system inadvertantly.

joelhandwell:
@chef/client-maintainers @sethvargo @stevendanna Removed Timecop dependency and rebased on top of 3770ec1025fd09d3a3bb393ad955ee2985252a67

sethvargo:
:+1:

stevendanna:
:+1:

joelhandwell:
Rebased on top of 1ef5f566480ba84091a12f2b6c5e85f44af54a5b

PR #3827, opened 2015-08-26
Rewriting volume code to use FFI

jaym:
This will let us rip out the windows-pr gem from chef

cc @adamedx @chefsalim @ksubrama @mcquin

### DIFF
--- a/lib/chef/util/windows/volume.rb
+++ b/lib/chef/util/windows/volume.rb
@@ -35,11 +35,10 @@ class Chef::Util::Windows::Volume < Chef::Util::Windows
   end
 
   def device
-    buffer = 0.chr * 256
-    if GetVolumeNameForVolumeMountPoint(@name, buffer, buffer.size)
-      return buffer[0,buffer.size].unpack("Z*")[0]
-    else
-      raise ArgumentError, get_last_error
+    begin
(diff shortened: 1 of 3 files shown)


chefsalim:
Looks good - only comment: in file.rb - use LPCTSTR/LPTSTR instead of LPCSTR

lib/chef/util/windows/volume.rb
```diff
@@ -18,42 +18,42 @@
...
-  include Windows::Volume
-  #XXX not defined in the current windows-pr release
-  DeleteVolumeMountPoint =
-    Windows::API.new('DeleteVolumeMountPoint', 'S', 'B') unless defined? DeleteVolumeMountPoint
-
-  public
+  attr_reader :mount_point
 
   def initialize(name)
     name += "\\" unless name =~ /\\$/ #trailing slash required
```
ksubrama: This is tangential to the rest of the code-review, but I just noticed that this line mutates the incoming argument.  Is this deliberate?  Then again, now I'm thinking about all the places where a constructor just holds onto a reference to an incoming string that is later mutated by the caller.  Oh Ruby...

ksubrama:
:+1:

PR #3828, opened 2015-08-27
Add Salim as a windows maintainer

ksubrama:


smurawski:
:+1:

ksubrama:
:+1:

btm:
https://github.com/chef/chef-rfc/blob/master/rfc030-maintenance-policy.md#maintainer-1

Technically there needs to be more than 50% positive votes from the component maintainers/LTs. Since there are four existing LT/maintainers, that means 3 votes.

So I'll toss in an ex post facto :+1: to put you back on the good side of the law.

PR #3834, opened 2015-08-28
Dont spit out stdout and stderr for execute resource failure, if its declared sensitive

### LINKED ISSUE #3813: Sensitive data exposed in execute resource
I'm seeing sensitive data on failed execute resource executions, which was reported in #1744 and was supposed to be fixed in #2013. I'm not sure if I'm misinterpreting that fix, or if this is a similar issue. I'd expect that when using the sensitive property that I wouldn't see any of the direct output from either STDOUT or STDERR, which is contrary to what I'm seeing. I'm running test kitchen converge using the 12.4.1 omnibus install and see this on both Windows and Linux (CentOS). 

Output from chef-client run:
```
* execute[sensitive-test] action run

           Error executing action `run` on resource 'execute[sensitive-test]'

           Expected process to exit with [99], but received '0'
           ---- Begin output of echo "this is sensitive" ----
           STDOUT: this is sensitive
           STDERR:
           ---- End output of echo "this is sensitive" ----
           Ran echo "this is sensitive" returned 0

           Resource Declaration:
           ---------------------
           suppressed sensitive resource output

           Compiled Resource:
           ------------------
           suppressed sensitive resource output

```

Execute resource block in recipe:
```
execute 'sensitive-test' do
  command "echo \"this is sensitive\""
  action :run
  sensitive true
  returns [99]
end
```

ranjib:
closes #3813

### DIFF
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -60,12 +60,12 @@ class Chef
         converge_by("execute #{description}") do
           begin
             shell_out!(command, opts)
-          rescue Mixlib::ShellOut::ShellCommandFailed => e
+          rescue Mixlib::ShellOut::ShellCommandFailed
             if sensitive?
               raise Mixlib::ShellOut::ShellCommandFailed,
                 "Command execution failed. STDOUT/STDERR suppressed for sensitive resource"
             else
-              raise e
+              raise
             end
           end
           Chef::Log.info("#{new_resource} ran successfully")
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -58,7 +58,16 @@ class Chef
         end
 
         converge_by("execute #{description}") do
-          result = shell_out!(command, opts)
+          begin
+            shell_out!(command, opts)
+          rescue Mixlib::ShellOut::ShellCommandFailed => e
+            if sensitive?
+              raise Mixlib::ShellOut::ShellCommandFailed,
+                "Command execution failed. STDOUT/STDERR suppressed for sensitive resource"
+            else
+              raise e
+            end
+          end
           Chef::Log.info("#{new_resource} ran successfully")
         end
       end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/execute.rb
```diff
@@ -58,7 +58,16 @@ def action_run
...
         converge_by("execute #{description}") do
-          result = shell_out!(command, opts)
+          begin
+            shell_out!(command, opts)
+          rescue Mixlib::ShellOut::ShellCommandFailed => e
+            if sensitive?
+              raise Mixlib::ShellOut::ShellCommandFailed,
+                "Command execution failed. STDOUT/STDERR suppressed for sensitive resource"
+            else
+              raise e
```
coderanger: I think we just want `raise` instead of `raise e` to preserve the original stack trace.

ranjib:
@coderanger ah.,.gotcha. fixed it :0) thanks for the CR

thommay:
i'm :+1: on the code, although this could use a test.

lamont-granquist:
You could maybe include some tests over this that stub the call to shell_out and either raise or not and check that its handled appropriately.

But :+1:  on the code.

ranjib:
i'll add a spec as @lamont-granquist mentioned

ranjib:
@thommay @lamont-granquist spec added. i think this is ready for merge

stevendanna:
:+1:

lamont-granquist:

lib/chef/provider/execute.rb
```diff
@@ -58,7 +58,16 @@ def action_run
         end
 
         converge_by("execute #{description}") do
-          result = shell_out!(command, opts)
+          begin
+            shell_out!(command, opts)
+          rescue Mixlib::ShellOut::ShellCommandFailed
+            if sensitive?
+              raise Mixlib::ShellOut::ShellCommandFailed,
```
danielsdeleo: This also modifies the stack trace and such, but given that we're only 2 lines away from where the error must have been raised in the first place, it's probably not a big deal.

danielsdeleo:
:+1:

thommay:
thanks, @ranjib

Issue #3835, opened 2015-08-28
systemd provider helper platform_has_systemd_unit? broken in 12.4.1, pending 11.18.0 release


fujin:
https://github.com/chef/chef/blob/11.18.0.r/lib/chef/platform/service_helpers.rb#L118-L138
https://github.com/chef/chef/blob/master/lib/chef/platform/service_helpers.rb#L118-L139

```
[aj@endpoint1df988a3]~% sudo cat /var/chef/cache/chef-stacktrace.out
Generated at 2015-08-28 07:29:31 +0000
NoMethodError: service[cloud-config.service] (pantheon::cloud_init line 15) had an error: NoMethodError: undefined method `+' for false:FalseClass
/opt/chef/embedded/apps/chef/lib/chef/platform/service_helpers.rb:134:in `platform_has_systemd_unit?'
/opt/chef/embedded/apps/chef/lib/chef/platform/service_helpers.rb:98:in `config_for_service'
/opt/chef/embedded/apps/chef/lib/chef/provider/service/redhat.rb:36:in `supports?'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `block in supported_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `select'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `supported_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:91:in `prioritized_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:116:in `maybe_dynamic_provider_resolution'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:61:in `resolve'
/opt/chef/embedded/apps/chef/lib/chef/resource.rb:1205:in `provider_for_action'
/opt/chef/embedded/apps/chef/lib/chef/resource.rb:584:in `run_action'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:49:in `run_action'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `block (2 levels) in converge'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `each'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `block in converge'
/opt/chef/embedded/apps/chef/lib/chef/resource_collection/resource_list.rb:83:in `block in execute_each_resource'
[...]

PR #3836, opened 2015-08-28
simplify service helpers

nathwill:
related to https://github.com/chef/chef/issues/3835


ranjib:
:+1: so much better

stevendanna:
Is it possible that people are using the service resource to ensure the state of non-service unit files?  For instance

```
service "sshd.socket" do
  enable [:enable, :start]
end
```

Perhaps we can check for either `"#{cfg_base}/systemd/system/#{svc_name}.service"` or `"#{cfg_base}/systemd/system/#{svc_name}"`

As you mentioned this will miss loaded units that aren't backed by files.  From what I've seen most dynamically generated units are going to be device, mount, automount, or swap units.  Perhaps it is OK that we miss those?

ranjib:
is it ok to use service resource define socket? we can explicitly mention this, that the service resource only controls *.service units

stevendanna:
> is it ok to use service resource define socket?

The code I posted works currently with Chef.  I haven't seen much usage like that to know if we'd be breaking a large or small number of cookbooks.

ranjib:
hmm.. then we have to check for both i guess. but its kinda not idiomatic, is it acceptable if we add a deprecation warnings for non-service units, cause systemd supports a wide array of unit types, we wont be able to address the resulting corner cases within a single service provider. we can recommend folks to use the systemd cookbook for non service units

nathwill:
a socket is not a service, and if anyone's using service resources to manage non-service units, it's their own fault if it breaks; if it works at present, it's only because chef isn't doing what it ought to and specifying the unit type when shelling out to systemd for the service resource (i.e. `systemctl start #{service_name}` instead of `systemctl start #{service_name}.service`.

nathwill:
fix for systemd service provider overloading proposed in #3838

nathwill:
ugh, and i realize that https://github.com/chef/chef/pull/3836#issuecomment-135852297 sounded way crankier than i meant it to... i meant "own fault" in the "shouldn't be surprised" sense, more than the "finger wagging" sense... apologies.

joemiller:
>  From what I've seen most dynamically generated units are going to be device, mount, automount, or swap units.

Not sure this is a safe assumption to make. While we as a large systemd user don't do this currently, we have talked about it before. Others might as well. You can for instance have a single `redis@.service` unit file but instantiate dozens of redis instances using `redis@1234.service` as the unit name without making new files (I think...=). The 1234 can then be subbed into the unit file at runtime by systemd to use as the port number for that redis instance.

spheromak:
I like this approach because it doesn't slam dbus, but this would break for us do to the aforementioned  `service "foo.socket"` We use heaps of sockets. Thats ok in this patch tho if you include .socket files  (for us at least).  

There has to be a way to ask systmed 'is this a unit even if its not loaded' without invoking systemctl --list-all-units. I was trying to figure that out last night.

 I personally wouldn't mind being able to tell chef, hey we are a systemd shop, don't go through these shenanigans to try  and be smart. Just run systemctl status on this string I am passing you. I don't get why the platform detection is concerned about a specific service anyhow. thats the providers job isn't it?

FWIW  chef 11.16 works great with ~20k systemd services.

nathwill:
@joemiller interesting. using the `name@.service` model wouldn't actually work for the `systemctl list-unit-files` method of detecting either, unless you stripped everything after the `@` off the resource name (doable i suppose, but eww). we could try to change the `has_systemd_service_unit?` method here to check for the unit file after having stripped the service name of anything after the `@`, but i dunno... feels like a more targeted resource would be more appropriate at that point :/.

@spheromak have you looked at using the `systemd_socket` resource in the systemd cookbook?

spheromak:
@nathwill A socket is a service in that if you use socket activation to run services it is a service :). Pedantically sockets aren't the same type to systemd, but from a chef perspective a socket should be considered a service. You still run the same lifecycle commands on them.

spheromak:
@natewalck no we don't use the systemd cookbook, but sounds interesting.

nathwill:
@spheromak we'll have to agree to disagree there ;) it works only at a *very* shallow level, if you try to do anything more involved, the illusion falls apart quickly (e.g. wtf does "reload" do on a socket?).

spheromak:
@nathwill as long as theres a path for managing socket units I am not bothered. Just that the current patch would break on the existing chef code I have inherited. Guess thats better than the current master that times out dbus :)

nathwill:
i think my main objection here is that management of non-service units via the service resource has only worked in the past/present via exploitation of what is clearly a bug in Chef (no-one intended to support random unit types via the systemd service resource/provider).

i understand some (many?) folks may have been exploiting that bug in the absence of am 'official' way to manage misc units, but continued support for doing so (by not closing the bug) will only muddy the waters if/when chef later adds proper support for other unit-based resources (which i assume is inevitable), and it inherently sanctions continued misuse of the service resource.

to my mind at least, it's better to add a changelog warning and enforce the systemd-service provider's area of responsibility, and then add true support for other units if they're needed/wanted.

fujin:
@nathwill hey just tried this on one of our machines with 10k units; it doesn't work with .automount units, generates an error about no providers responding to supports, then kind of looks like it hangs. I only had a brief opportunity to test this then had to roll it back. I'll keep messing with it and see if I can get what you have to work.

```
[2015-08-29T01:35:53+00:00] INFO: Processing service[srv-bindings-55fc383fa9e041bdafc5d0567eb87b2b-files.automount] action start (endpoint::_appserver_bindings line 415)
[2015-08-29T01:35:53+00:00] DEBUG: Providers for generic service resource enabled on node include: [Chef::Provider::Service::Redhat, Chef::Provider::Service::Systemd, Chef::Provider::Service::Init]
[2015-08-29T01:35:53+00:00] DEBUG: No providers responded true to `supports?` for action start on resource service[srv-bindings-55fc383fa9e041bdafc5d0567eb87b2b-files.automount], falling back to enabled handlers so we can return something anyway.
[2015-08-29T01:35:53+00:00] DEBUG: Provider for action start on resource service[srv-bindings-55fc383fa9e041bdafc5d0567eb87b2b-files.automount] is Chef::Provider::Service::Systemd
```

fujin:
@nathwill I'd guess that is probably something to do with the previously discussed non-services-as-service-resources. I've scheduled some time to look at your lwrps, which are super promising, not sure how I'd missed them so far :)

nathwill:
@joemiller so the `redis@6380` type service should work now as well. that look alright?

nathwill:
@fujin yeah, definitely related, sorry for the hassle. this seems to me like an opportune time to clean this up, but maybe the maintainers will want to deprecate it over a couple releases? :8ball:

stevendanna:
I definitely think we need to continue to support cookbooks doing "server 'sshd.scoket'" until some deprecation period is over.  Breaking straightforward cookbooks in a minor release is a no-go.

nathwill:
@stevendanna alright, this should support non-service-unit "services" now as well.

lamont-granquist:
LGTM, everyone :+1: on shipping?

smurawski:
:+1:

nathwill:
i'm obviously :+1: on it as well :)

nathwill:
alright, straightened out some merge conflicts and got the specs working again. still interested in merging?

PR #3837, opened 2015-08-28
refactor remote_directory provider

lamont-granquist:
- Huge speed and memory perf boost.  In the prior code if you were
  copying 10 dotfiles to a home directory with a million files in it,
  would slurp all million directory entries into a ruby Set object even if you were
  not ultimately going to purge the unmanaged files.  This code inverts
  the logic and tracks managed files and then iterates through the
  filesystem without slurping a list into memory.
- Only do file purging logic if purging is set.
- Fixes mutation of new_resource.overwrite.
- Fixes mutation of new_resource.rights (subtle).
- Adds helper delegators to the new_resource properties.
- Deprecates (instead of removes) now-unused methods.
- Renamed a method (with deprecated alias preserved) for consistency.
- Adds YARD for everything.
- Changes protected to private because protected is largely useless
  in ruby.
- Removes whyrun_supported? because the superclass sets that.

### DIFF
--- a/lib/chef/deprecation/provider/remote_directory.rb
+++ b/lib/chef/deprecation/provider/remote_directory.rb
@@ -1,0 +1,52 @@
+#
+# Author:: Serdar Sutay (<serdar@opscode.com>)
+# Copyright:: Copyright (c) 2013-2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Deprecation
+    module Provider
+      module RemoteDirectory
+
+        def directory_root_in_cookbook_cache
+          Chef::Log.deprecation "the Chef::Provider::RemoteDirectory#directory_root_in_cookbook_cache method is deprecated"
+
+          @directory_root_in_cookbook_cache ||=
+            begin
+              cookbook = run_context.cookbook_collection[resource_cookbook]
+              cookbook.preferred_filename_on_disk_location(node, :files, source, path)
+            end
+        end
+
+        # List all excluding . and ..
+        def ls(path)
+          files = Dir.glob(::File.join(Chef::Util::PathHelper.escape_glob(path), '**', '*'),
+                           ::File::FNM_DOTMATCH)
+
+          # Remove current directory and previous directory
+          files = files.reject do |name|
+            basename = Pathname.new(name).basename().to_s
+            ['.', '..'].include?(basename)
+          end
+
+          # Clean all the paths... this is required because of the join
+          files.map {|f| Chef::Util::PathHelper.cleanpath(f)}
+        end
+
+      end
+    end
+  end
+end
--- a/lib/chef/provider/remote_directory.rb
+++ b/lib/chef/provider/remote_directory.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@opscode.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Copyright:: Copyright (c) 2008-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


lamont-granquist:
@Igorshp can you give this a test on your use case?

lamont-granquist:
There's no spec changes here because AFAIK this a pure refactor and is functionally identical, but has perf CPU + Mem fixes, so unsure right now of what specs to add.

There's a FIXME I added with a reference to an old CHEF ticket that turned up when I was researching stuff for documentation on why we search the way we search.  Fixing that was out of scope with the timebox that I want around this PR since this is a squirrel project that is distracting me from other work right now.

lamont-granquist:

lamont-granquist:
remote_directory_refactor

This is the before and after of memory pressure on a CCR on one of my servers with the big red spike being a single remote_directory provider before this refactor was done.

lamont-granquist:
Also removes a 4 second delay on my cookbooks as it no longer walks my home dir at all, because I do not have purge set:

https://gist.github.com/lamont-granquist/def55172542a8c0b3200

It should also not cause the memory explosion with purge set and should be slightly faster and use less memory than @Igorshp's fix since we stop mangling large arrays in memory and just walk the filesystem dirents once and during the process skip skippable files.

lib/chef/deprecation/provider/remote_directory.rb
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Deprecation
+    module Provider
+      module RemoteDirectory
```
danielsdeleo: This is for folks who've subclassed `RemoteDirectory` ?

lib/chef/provider/remote_directory.rb
```diff
@@ -16,178 +16,268 @@
...
-          ['.', '..'].include?(basename)
+      # Add a file and its parent directories to the managed_files Hash.
+      #
+      # @param [String] cookbook_file_relative_path relative path to the file
+      # @api private
+      #
+      def add_managed_file(cookbook_file_relative_path)
+        if purge
+          Pathname.new(Chef::Util::PathHelper.cleanpath(::File.join(path, cookbook_file_relative_path))).descend do |d|
+            managed_files[d.to_s] = true
```
danielsdeleo: You could use a `Set` here but nbd.

spec/unit/provider/remote_directory_spec.rb
```diff
@@ -79,7 +79,7 @@ class Chef::CFCCheck
     end
 
     it "configures access control on intermediate directorys" do
-      directory_resource = @provider.send(:resource_for_directory, File.join(Dir.tmpdir, "intermediate_dir"))
+      directory_resource = @provider.send(:directory_resource, File.join(Dir.tmpdir, "intermediate_dir"))
```
danielsdeleo: Probably a little cleaner to have this be `@api private` but ruby-public. nbd though.

lib/chef/provider/remote_directory.rb
```diff
@@ -16,178 +16,268 @@
...
+      def_delegators :@new_resource, :files_rights, :files_mode, :files_group, :files_owner, :files_backup
+      def_delegators :@new_resource, :rights, :mode, :group, :owner
+
+      attr_accessor :overwrite
+
+      # The overwrite property on the resource.  Delegates to new_resource but can be mutated.
+      #
+      # @return [Boolean] if we are overwriting
+      #
+      def overwrite
```
danielsdeleo: Probably a bit more idiomatic for this to be `overwrite?`.

danielsdeleo:
A few nitpicky things that I don't care deeply about, :+1: though.

smurawski:
LGTM :+1:

lib/chef/deprecation/provider/remote_directory.rb
lamont-granquist: Yep

lib/chef/provider/remote_directory.rb
lamont-granquist: After seeing what the other `Set` did to my memory I'm a little gun-shy of `Set`, but that may have just been all the entries...

spec/unit/provider/remote_directory_spec.rb
lamont-granquist: Yeah, I still can't decide how I feel about that, but I just left that how I found it.   Also kinda smells like the specs might need some refactoring to test the public api instead of the inner hooks, but that's very much out of scope.

lib/chef/provider/remote_directory.rb
lamont-granquist: Thought about that, but felt like it was a delegator to `@new_resource.overwrite` that was safe to mutate...

lib/chef/provider/remote_directory.rb
danielsdeleo: `Set` actually just wraps a hash. For example, the `<<` method is an alias for `add` which is defined as:

```
def add(o)
  @hash[o] = true
  self
end
```

lib/chef/provider/remote_directory.rb
danielsdeleo: I kinda wish we'd also have more non-DSL-y methods in resources, e.g., having normal setters in addition to the DSL-style ones, though that's out of scope for this PR. Outside of DSL things, I'd prefer to be a bit more idiomatic, though I can accept your argument. I'm fine to merge it either way.

lamont-granquist:
Okay, converted to Set and switched to `#overwrite?` per @danielsdeleo 

@Igorshp any chance you could test this code against your case?

Igorshp:
@lamont-granquist sure thing, i'll report back with results shorlty

Igorshp:
@lamont-granquist 

Before: `DEBUG:   109.663956  remote_directory[/var/lib/jenkins]`
After: `DEBUG:     0.442248  remote_directory[/var/lib/jenkins]`

That's some impressive performance boost :)

:+1:

danielsdeleo:
:shipit:

thommay:
:clap: :tada: :shipit:

Igorshp:
:clap:

PR #3838, opened 2015-08-28
systemd service provider should be restricted to its own area of responsibility

nathwill:
specify the unit type to prevent inappropriate overloading of the systemd service provider to manage non-service units.

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -74,7 +74,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
       if new_resource.start_command
         super
       else
-        shell_out_with_systems_locale!("#{systemctl_path} start #{new_resource.service_name}")
+        shell_out_with_systems_locale!("#{systemctl_path} start #{new_resource.service_name}.service")
       end
     end
   end
@@ -86,7 +86,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
       if new_resource.stop_command
         super
       else
-        shell_out_with_systems_locale!("#{systemctl_path} stop #{new_resource.service_name}")
+        shell_out_with_systems_locale!("#{systemctl_path} stop #{new_resource.service_name}.service")
       end
     end
   end
@@ -95,7 +95,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
     if new_resource.restart_command
       super
     else
-      shell_out_with_systems_locale!("#{systemctl_path} restart #{new_resource.service_name}")
+      shell_out_with_systems_locale!("#{systemctl_path} restart #{new_resource.service_name}.service")
     end
   end
 
@@ -104,7 +104,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
       super
     else
       if current_resource.running
-        shell_out_with_systems_locale!("#{systemctl_path} reload #{new_resource.service_name}")
+        shell_out_with_systems_locale!("#{systemctl_path} reload #{new_resource.service_name}.service")
       else
         start_service
       end
@@ -112,19 +112,19 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
   end
 
   def enable_service
-    shell_out!("#{systemctl_path} enable #{new_resource.service_name}")
+    shell_out!("#{systemctl_path} enable #{new_resource.service_name}.service")
   end
 
   def disable_service
-    shell_out!("#{systemctl_path} disable #{new_resource.service_name}")
+    shell_out!("#{systemctl_path} disable #{new_resource.service_name}.service")
   end
 
   def is_active?
-    shell_out("#{systemctl_path} is-active #{new_resource.service_name} --quiet").exitstatus == 0
+    shell_out("#{systemctl_path} is-active #{new_resource.service_name}.service --quiet").exitstatus == 0
   end
 
   def is_enabled?
-    shell_out("#{systemctl_path} is-enabled #{new_resource.service_name} --quiet").exitstatus == 0
+    shell_out("#{systemctl_path} is-enabled #{new_resource.service_name}.service --quiet").exitstatus == 0
   end
 
   private
--- a/spec/unit/provider/service/systemd_service_spec.rb
+++ b/spec/unit/provider/service/systemd_service_spec.rb
@@ -153,13 +153,13 @@ describe Chef::Provider::Service::Systemd do
         end
 
         it "should call '#{systemctl_path} start service_name' if no start command is specified" do
-          expect(provider).to receive(:shell_out_with_systems_locale!).with("#{systemctl_path} start #{service_name}").and_return(shell_out_success)
+          expect(provider).to receive(:shell_out_with_systems_locale!).with("#{systemctl_path} start #{service_name}.service").and_return(shell_out_success)
           provider.start_service
         end
 
         it "should not call '#{systemctl_path} start service_name' if it is already running" do
           current_resource.running(true)
-          expect(provider).not_to receive(:shell_out_with_systems_locale!).with("#{systemctl_path} start #{service_name}")
+          expect(provider).not_to receive(:shell_out_with_systems_locale!).with("#{systemctl_path} start #{service_name}.service")
           provider.start_service
         end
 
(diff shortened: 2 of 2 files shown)


lamont-granquist:
@stevendanna you seem to be our systemd expert?  :trollface:

coderanger:
Would we count this as a compat break?

nathwill:
@coderanger i think yes based on comments in #3836

lamont-granquist:
So is the problem here using a 'service' resource to manage a systemd thing that isn't a 'service'?

Do we have a replacement resource to be able to manage the non-service systemd things?

Can we file this in the same drawer that we keep the trick about using `file://` urls in remote_file in order to accomplish local copying?

nathwill:
@lamont-granquist the problem is that it encourages people to misuse the resource for non-service units, which then interferes with improving service-unit handling in chef, as in #3836 (why the hell should we have to check for non-service units in the service unit handler).

we actually have a whole [cookbook](https://github.com/nathwill/chef-systemd) for non-service systemd units and other systemd things, which is fairly new, but at least provides somewhere for the folks currently misusing the service resource to turn for proper xyz-unit type support.

my personal hope would be that in the long-term chef will end up with strong support for other systemd resources, hopefully by absorbing a lot of the systemd cookbook code once it's been vetted for a while outside of core. but keeping the path clear for doing that by clearly marking the boundaries of the service unit-related code is gonna be important for being able to do that down the line.

lamont-granquist:
Okay that sounds like a decent rationale, but it still seems backcompat breaking.

jaymzh:
:-1:  This breaks using the service provider sanely. @davide125 is writing up details as we speak.

davide125:
Yeah, this would break all the things for us. While using `service` for non-service units is maybe a misuse, there's no sensible alternative at the moment. That cookbook provides a ton of facilities for dropping and managing various types of units (to the point of basically reimplementing the systemd DSL in chef), which is fine if that's what you want to do, but makes it hard to manage units dropped by packages or as cookbook_file / templates. If having to support non-service units in `service` is a problem, then maybe we need a `unit` resource for that (though I'm not personally that convinced, as it would basically duplicate most of the code for the `service` provider for systemd).

davide125:
Also, this will break people using socket activated services (which are effectively services, even if they're managed through socket units). See #4307 for related work we're doing on the systemd service provider.

lamont-granquist:
Sounds like I may switch to :-1:

paging @stevendanna and @thommay

nathwill:
@davide125 i'd like to hear more about the difficulty encountered with managing package-provided units with the systemd cookbook, it should just be as easy as using the appropriate {enable,disable,start,stop} actions. the [provider](https://github.com/nathwill/chef-systemd/blob/master/libraries/unit.rb#L137-L181) for those actions doesn't try to alter the unit files at all. maybe you could open an issue on that cookbook with your use-case?

jaymzh:
But the goal is to NOT have to use a separate resource for services on different platforms. Who wants:

```
if node['platform_family'] == 'rhel' and node['platform_version'].split('.')[0].to_i > 6
  systemd_socket 'sshd' do
    action [:enable, :start]
  end
else
  # EVERY OTHER PLATFORM
  service 'sshd' do
    action [:enable, :start]
  end
end
```

That sucks. The whole POINT of the service abstration (and the package abstraction and...) is to NOT have to do that. A service is a service is a service. You want to tell chef to start the sshd service in the most appropriate way for the that platform. See the PR that Davide references. Just like I don't want to say:

```
if node['platform_family'] == 'debian'
  apt_package 'sshd' do
   action :upgrade
  end
elsif node['platform_family'] == 'rhel'
  if node['platform_version'].split('.')[0].to_i < 8
    yum_package 'sshd' do
      action :upgrade
    end
  else
    dnf_package 'sshd' do
      action :upgrade
    end
  end
else ...
```

and so on. you just want:

```
package 'sshd' do
  action :upgrade
  notifies :restart, 'service[sshd]'
end

service 'sshd' do
  action [:enable, :upgrade]
end
```

Maybe with a `value_for_platform` call to abstract out the right name of the package. Separating this out is the antipattern.

nathwill:
@jaymzh if you wanted socket activation, you'd need a different resource (or some kind of guarded name override) anyways (i.e. `service 'sshd' do... end` != `service 'sshd.socket' do...end`), and if you didn't want socket-activation, this change doesn't impact your use case. `service 'sshd'` works the same regardless.

in order to achieve what you're describing, you'd have to have the service provider prefer socket units when present, which just sounds like a world of pain.

btw, i'm in full agreement with you about preferring the `service` resource for x-platform support, which is why the systemd cookbooks [README](https://github.com/nathwill/chef-systemd#systemd-service) explicitly recommends using  the enable,start, etc actions on it instead of the `systemd_service` resource.

nathwill:
> sounds like a world of pain

to elaborate more on why i think this kind of change (preferring to activate socket units over service units from the service provider on systemd platforms) would be a Bad Idea, imagine the unsuspecting admin upgrading from non-systemd platform (e.g. ubuntu 14.04) to systemd-based platform (e.g. ubuntu 16.04).

Due to regular brute-force attacks against ssh, this user's been updating their sshd_config to listen on 2222 (security through obscurity, but this is the world we live in); picture their surprise and dismay were they to discover that their ssh service is *not* in fact even running on 16.04, and that their beloved CM tools have decided instead to spin up some fancy new systemd feature *that uses a totally different port binding configuration* (ListenStream in the socket unit), and is in fact back to listening on port 22!

this is exactly the kind of situation that makes me say "sockets aren't services", and believe that we need to acknowledge and accomodate unit distinctions in Chef.

davide125:
@nathwill: about chef-systemd: you're right, I missed that this could be used with existing units and it'd leave them unmodified, mostly because the README is heavily skewed towards the "let's build a full unit from chef" usecase. That would work, but IMO it's still not ok to break functionality in chef and expect people to use an external cookbook instead. One thing we could do, that might address the "sockets aren't services" concern, is add a new `unit` provider, that would behave like `service` does currently and support enabling/starting/masking any kind of systemd unit. We could then gut the systemd implementation in `service` and reimplement in on top of `unit`, and add some checks in it to warn people they should move to `unit` for non-service units. Then once that's been out in the wild for a while, we could turn the warnings into errors and stop supporting non-services in `service` altogether. 

I'm still not a fan of this approach though, as @jaymzh said it'll force people to use different resources depending on their platform, which seems suboptimal.

nathwill:
> it's still not ok to break functionality in chef and expect people to use an external cookbook instead

yeah, i totally understand this perspective, and if we want to say no to this PR on those grounds i'm fine with it. i'm still concerned that the service resource is likely to get bogged down in a mine-field of unknowable use-cases without enforcing scope, but if we prefer to favor back-compat, so be it.

> add a new unit provider, that would behave like service does currently and support enabling/starting/masking any kind of systemd unit.

i like the idea... it seems like it's going to be essential to have a unit base class that resources for more specific unit types can build on for systemd support, so makes sense as a good first step regardless of whether the SystemdService provider limits its scope to service units.

jaymzh:
I prefer to take the yum_package / package approach. By default you just do `service` and then ti resolves to `systemd_service` internally if necessary which provides extra options as necessary. Things like systemd shouldn't be in an external cookbook (except maybe for esoteric features?)

nathwill:
> By default you just do service and then ti resolves to systemd_service

i don't think anyone's proposing changing this. this PR just enforces that systemd_service only acts on services, not other random unit types.

> Things like systemd shouldn't be in an external cookbook (except maybe for esoteric features?)

i agree! the absence of substantive support for systemd's way of doing things in chef core required a solution, and vetting the ideas we had outside of chef core seemed like a good first step, but i'm 100% in favor of mainlining that stuff now if there's a desire to do so.

jaymzh:
Indeed - that's what Davide's PR does. I want to get this stuff into core.

nathwill:
cool, it sounds like we're in violent agreement here, then. there seem to be two non-overlapping issues at question:

- should systemd_service limit it's scope to restrict itself to acting only on service resources?

the general feeling i get at this point is that we should leave it be for now so as to avoid back-compat breaks, but add a deprecation warning ASAP for when we determine that it's being used for non-service units.

- should we start supporting systemd more in chef-core (i.e. adding mask/unmask actions for systemd_service as Davide's PR addresses), and maybe add other systemd resources as appropriate

i'd say "hell yes", and it sounds like others agree. i'm all for merging Davide's PR as a first step.

i'd suggest at this point that we close this PR in light of the back-compat concerns, and continue the discussion on how to best improve systemd support in a separate issue or via separate PRs. sound reasonable?

davide125:
@nathwill yup, works for me -- thanks!

PR #3853, opened 2015-09-01
move warning to debug

lamont-granquist:
1.  the deploy resource and other internal chef resources trigger this.

2.  this is largely a non-problem since people have been happily using
the deploy resource for years and nobody has showed up in any of our
bugtrackers due to mutating the default hash.  we simply don't need to
annoy people this badly over this.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -808,10 +808,6 @@ class Chef
         property = property_type(**options)
       end
 
-      if !options[:default].frozen? && (options[:default].is_a?(Array) || options[:default].is_a?(Hash))
-        Chef.log_deprecation("Property #{self}.#{name} has an array or hash default (#{options[:default]}). This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes. Either freeze the constant with `.freeze` to prevent appending, or use lazy { #{options[:default].inspect} }.")
-      end
-
       local_properties = properties(false)
       local_properties[name] = property
 
--- a/spec/integration/client/client_spec.rb
+++ b/spec/integration/client/client_spec.rb
@@ -350,8 +350,6 @@ EOM
       expect(run_complete).to be >= 0
 
       # Make sure there is exactly one result for each, and that it occurs *after* the complete message.
-      expect(match_indices(/MyResource.x has an array or hash default/, result.stdout)).to match([ be > run_complete ])
-      expect(match_indices(/MyResource.y has an array or hash default/, result.stdout)).to match([ be > run_complete ])
       expect(match_indices(/nil currently does not overwrite the value of/, result.stdout)).to match([ be > run_complete ])
     end
   end
--- a/spec/unit/property_spec.rb
+++ b/spec/unit/property_spec.rb
@@ -480,12 +480,6 @@ describe "Chef::Resource.property" do
         end
       end
 
-      it "when a property is declared with default: {}, a warning is issued" do
-        expect(Chef::Log).to receive(:deprecation).with( /^Property .+\.x has an array or hash default \(\{\}\)\. This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes\. Either freeze the constant with \`\.freeze\` to prevent appending, or use lazy \{ \{\} \}\.$/, /property_spec\.rb/ )
-        resource_class.class_eval("property :x, default: {}", __FILE__, __LINE__)
-        expect(resource.x).to eq({})
-      end
-
       with_property ':x, default: lazy { {} }' do
         it "when x is not set, it returns {}" do
           expect(resource.x).to eq({})


lamont-granquist:
@chef/client-core @chef/client-maintainers review plz

smurawski:
:+1:

danielsdeleo:
Do we even need it at all? I thought the top-level issue is when you define a default in a LWRP, which is in the scope of the class definition and therefore it is shared between all instances of the class, which is a bad thing, _but_ we now auto-freeze those so the actual problem is fixed. Or am I missing some important detail?

lamont-granquist:
I'm a bit confused as well...

coderanger:
We removed the auto-freeze I thought, pending Chef 13?

danielsdeleo:
I guess we did: https://github.com/chef/chef/pull/3744

danielsdeleo:
But now I'm more confused about how the deploy resource even hits this case. Looking at https://github.com/chef/chef/blob/master/lib/chef/resource/deploy.rb and searching for "default", I don't see anything relevant.

lamont-granquist:
@danielsdeleo looks like its coming from application_ruby not deploy resource directly.

danielsdeleo:
Ok, so based on my understanding of the change history (and memory of a ML thread or bug report that I can't easily find right now):

1. User did actually report a bug where their LWRP behaved unexpectedly because of mutating defaults set via the LWRP DSL. IIRC this can result in non-error incorrect behavior (i.e., chef does something the user doesn't expect but there isn't necessarily an exception raised).
2. We tried to "fix" it by freezing defaults
3. That wasn't a good fix because it broke cases where users hadn't triggered the bad behavior and the messaging about why things broke was unclear and we didn't present a clear alternative.
4. We got rid of the auto-freeze but there's a warning (I'd say the warning text also could be improved but we'd have to make it longer and the infrastructure to do that well is in-progress but not here yet).

This correct?

My opinion is that a debug message isn't very useful as a response to the bug, we might as well just get rid of the log message and preferably try to find a different way to address it.

lamont-granquist:
I would suggest dup'ing or Marshall.dump/load if we want to try to fix it.  The other solution is just education.  If this happened once its still not very common.

danielsdeleo:
There's no perfect general solution since you can't Marshal IO objects and we can't know if dup is implemented correctly for custom types. We could extend Hash or Array with a module that overrides every mutating method to warn and then calls super. Or we could phase out `default` and make you write a real method in that case or modify the DSL to require a block or whatever.

danielsdeleo:
Anyway, I'm just in favor of removing the error entirely, rather than dropping to debug. The debug message won't be anywhere near the actual error in time or in the log.

coderanger:
I still think freezing non-proc defaults is a good solution so :-1: on this. We should get people to fix their footgun code.

danielsdeleo:
I think it's wrong that the DSL makes it easy to do the wrong thing and freeze makes it pretty much impossible to give clear error messages with good context, though I did prefer that to letting LWRPs silently have incorrect behavior.

coderanger:
It's definitely not a good solution, but it gives at least some warning that a footgun is being created.

lamont-granquist:
It creates pretty massive spam right now though and the impact is a lot higher than the footgun risk and it is going to be mostly warning when there's absolutely no actually buggy code involved (it warns anytime when the defaults 'could' be mutated, not when the defaults actually are mutated).

Having Chef 13 freeze the defaults would be better, then you'd only get an error if someone really mutated the defaults without needing to change a lot of cookbook code when there's no bug involved, that means waiting until Chef 13 to do anything about this.

danielsdeleo:
I agree on the warning spam, it's worse than the actual problem it's trying to warn you about. I just don't think a debug message does any good, so my preference is to delete lines 811-813 and the corresponding test. I would like to have some way to address the LWRP mutate bug that doesn't suck but I'm not going to block this on that.

lamont-granquist:
Updated to just remove the warning.

danielsdeleo:
:+1:

coderanger:
What is the rationale for removing this? That we are worried about warning-spam for community cookbooks so it would be non-actionable for most users? This is something we need to find a way to tell users to avoid The Bad Thing™ and just removing it without adding a better notification seems like a step backwards.

stevendanna:
I initially advocated for the warning over freezing the attributes.  After introducing the freeze, we saw cookbook breakages that I felt were unfair to inflict on users with a minor release.

The problem with the warnings is that they are often non-actionable.  In addition to community cookbooks, often the warning simply reports that the problem is with `#<Class:0x007f81fcc4aad8>.some_property`, which might be easily findable in a small system, but is not in a large system even when the user authored most of the cookbooks.

One possibility is to move this to a "deprecation" so that it uses the new rollup feature.  If the `caller` is reported correctly, that would also help make them more actionable.

I thought about wrapping defaults in some class such as:

```
require 'forwardable'
class PropertyDefault
  def initialize(data)
    @data = data.freeze
  end

  Object.new.methods.each do |method|
    def_delegator :@data, method
  end

  def method_missing(method, *args, &block)
    @data.send(method, *args, &block)
  rescue RuntimeError => e
    if is_frozen_exception?(e)
      puts "WARNING: You are modifying a blah blah blah this is bad because helpful helpful helpful you should do this this and this"
      @data = @data.dup
      @data.send(method, *args, &block)
      @data.freeze
    else
      raise
    end
  end

  private

  def is_frozen_exception?(e)
    if e.to_s =~ /can't modify frozen/
      true
    else
      false
    end
  end
end
```

But that is likely to break more code than freezing would because `#class` will return PropertyDefault still.

danielsdeleo:
For one, the warning is completely indiscriminate. For example, having a default value of a Hash is fine if you always use the resource in a way that sidesteps the bug:

```ruby
some_lwrp "name" do
  # BAD:
  thing[:foo] = "bar"
  # TOTALLY FINE:
  thing :foo => "bar"
end
```

Secondly, I agree with the argument here that the cure is worse than the disease. This warning triggers far too often, to the point that it's a usability issue.

Thirdly, I don't think we've done enough work on improving the LWRP attribute/property DSL to warrant a blanket warning for this case. I would like us to design an improved DSL where the easiest thing to do is the right thing before we make everyone change all their LWRPs.

And finally, I think just changing the message from warning to debug makes it pretty much useless. The message will not occur anywhere close to where the bug is exposed, and debug is so verbose that your chances of seeing it and learning that it's the cause of your problem some 1000s of lines of output later are very low.

I do think we should do _something_ to address the mutated-default bug, but we should not have so much collateral damage. Perhaps we could make a `WarnOnHashMutation` module, with code like:

```ruby
def []=(*args)
  warn_about_mutations
  super
end
```
...and `extend` that into any Hash defaults we get passed (and similar for Array). Or really any other thing that gives you a warning or error with good locality to the actual problem.

Likewise, if we come up with a better property/attribute DSL that doesn't default to doing the wrong thing, then I think it would be fair to deprecate the old thing and encourage everyone to adopt the new thing. I don't know if that's method chaining or something else, but I think `property :thing, default: lazy { {} }` isn't good enough.

coderanger:
Saying "this really bug doesn't affect you if you happen to avoid it" doesn't mean there isn't a bug. The freezing semantics are currently the accepted RFC-level correct behavior, so warning people about a pending deprecation seems very reasonable. If we want to do something else that should go back through through the RFC process.

danielsdeleo:
That's fine, but this warning as constituted affects everyone who has a LWRP with a Hash or Array default value, not just folks who would get bonked trying to modify a frozen value.

coderanger:
That's always the case with deprecation warnings though. We add them in the hope that either the original author notices or someone bothers them about it.

coderanger:
Also remember we just merged a change so that deprecation warnings are now shown in a de-dup'd summary at the end of the run, so it shouldn't spam output during the converge.

lamont-granquist:
Marking this as 12.5.0 since we should either merge it or close it before shipping, probably up to @thommay at this point.

thommay:
I agree with Lamont and Dan here; the cure is definitely worse than the disease. @lamont-granquist please go ahead and merge this, but please also spend some time on an RFC that would allow us to detect and warn about this class of problem when we actually encounter them.

PR #3854, opened 2015-09-01
Don't print deprecations to the console until the end

jkeiser:
The purpose of this is to dedupe and collect deprecation warnings and print them all at the end of the run. What we do:

1. Add a `deprecation` method to `EventHandler`. This means that Chef formatters (like the doc formatter that outputs to the console) now have control over how deprecations are printed. By default, deprecations are printed to `Chef::Log` exactly as they are now.
2. Add a `Chef.log` method that will return either the current run's `EventDispatcher` or `Chef::Log` if there is none. This should be expanded later, but is @api private for now because it does not support the other log methods.
3. Make the doc formatter handle deprecations differently: store all locations we noticed each separate deprecation message. Then at the end of the run, print the deprecation messages to the console.

Things that would be nice but are *not* part of this to make it a reasonable sized MVP:
1. A switch to turn off deprecation printing globally
2. Switches to turn off specific deprecation messages
3. De-dupes for other things like warnings
4. `Chef.log.debug`, `info`, `error` and `warn`.
5. Make the location of the deprecation easier to understand--if it's in a recipe, format it that way. Right now we show the file, which is better than nothing.
6. Class deprecations more generally: instead of printing a separate message for `Property MyResource.x has an array or hash default ([])` and `Property MyResource.x has an array or hash default ([])`, print  this:
   ```
   Properties with array or hash defaults:
     - MyResource.x ([]) - recipe[x::default]:4
     - MyResource.y ({}) - recipe[x::default]:5
   ```

A sample run (this is the output of the test):

```
Starting Chef Client, version 12.5.0.current.0
[2015-08-31T16:58:22-07:00] WARN: Run List override has been provided.
[2015-08-31T16:58:22-07:00] WARN: Original Run List: []
[2015-08-31T16:58:22-07:00] WARN: Overridden Run List: [recipe[x::default]]
resolving cookbooks for run list: ["x::default"]
Synchronizing Cookbooks:
  - x (0.0.0)
Compiling Cookbooks...
Converging 1 resources
Recipe: x::default
  * my_resource[blah] action nothing (skipped due to action :nothing)
[2015-08-31T16:58:22-07:00] WARN: Skipping final node save because override_runlist was given

Running handlers:
Running handlers complete

Deprecated features used!
  Property MyResource.x has an array or hash default ([]). This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes. Either freeze the constant with `.freeze` to prevent appending, or use lazy { [] }. at 1 location:
    - C:/Users/jkeiser/AppData/Local/Temp/chef_repo20150831-39476-j4xt6m/config/local-mode-cache/cache/cookbooks/x/recipes/default.rb:3:in `<class:MyResource>'
  Property MyResource.y has an array or hash default ({}). This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes. Either freeze the constant with `.freeze` to prevent appending, or use lazy { {} }. at 1 location:
    - C:/Users/jkeiser/AppData/Local/Temp/chef_repo20150831-39476-j4xt6m/config/local-mode-cache/cache/cookbooks/x/recipes/default.rb:4:in `<class:MyResource>'
  x nil currently does not overwrite the value of x. This will change in Chef 13, and the value will be set to nil instead. Please change your code to explicitly accept nil using "property :x, [MyType, nil]", or stop setting this value to nil. at 1 location:
    - C:/Users/jkeiser/src/chef/lib/chef/property.rb:419:in `x'

Chef Client finished, 0/1 resources updated in 7.278835 seconds
```


lamont-granquist:
it seems good to me.

lib/chef/event_dispatch/dispatcher.rb
```diff
@@ -28,6 +28,9 @@ def register(subscriber)
       # Define a method that will be forwarded to all
       def self.def_forwarding_method(method_name)
         define_method(method_name) do |*args|
+          if method_name == :deprecation && args.size == 1
```
coderanger: Can we not special case this?

danielsdeleo:
Are these two not de-duplicated because the messages are slightly different?

```
  Property MyResource.x has an array or hash default ([]). This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes. Either freeze the constant with `.freeze` to prevent appending, or use lazy { [] }. at 1 location:
    - C:/Users/jkeiser/AppData/Local/Temp/chef_repo20150831-39476-j4xt6m/config/local-mode-cache/cache/cookbooks/x/recipes/default.rb:3:in `<class:MyResource>'
  Property MyResource.y has an array or hash default ({}). This means that if one resource modifies or appends to it, all other resources of the same type will also see the changes. Either freeze the constant with `.freeze` to prevent appending, or use lazy { {} }. at 1 location:
    - C:/Users/jkeiser/AppData/Local/Temp/chef_repo20150831-39476-j4xt6m/config/local-mode-cache/cache/cookbooks/x/recipes/default.rb:4:in `<class:MyResource>'
```
?

lib/chef/chef_class.rb
```diff
@@ -190,6 +190,22 @@ def provider_handler_map
...
+    # @overload log
+    # Get the current log object.
+    #
+    # @return An object that supports `deprecation(message)`
+    #
+    # @example
+    #     run_context.log.deprecation("Deprecated!")
+    #
+    # @api private
+    def log
```
coderanger: Really uncomfortable calling this `log`, it doesn't return a logging object so the comments are directly wrong.

lib/chef/event_dispatch/dispatcher.rb
jkeiser: deprecation() is the only thing that has an argument with a default relating to the caller, and we need to trigger that default early in the process.

When we have a better error matcher thingy on the other end--something that will figure out what recipe was invoked, for example--we can have that logic scan the stack and skip the dispatcher.

jkeiser:
@danielsdeleo yep, the two messages are different because they have different text. I think we're in a *better* space now, but I'd like to make a followup patch that will separate the `MyResource.x` and the rest of the error message (see the 6-thing list in the description).

lib/chef/chef_class.rb
jkeiser: Yeah, I agree at the moment; that's why it's `@api private`. The intent is that it actually return an object that has the same interface as `Chef::Log` though--it's just `run_context`-specific and can be intercepted by a formatter. I can rename it `deprecator` for now until that's implemented.

lib/chef/event_dispatch/dispatcher.rb
coderanger: It doesn't need to, `caller` can be called from the method itself just fine.

jkeiser:
To be clear, we shouldn't approve this if we think it (a) makes the world worse than it is now or (b) doesn't make it any better. I'm trying to break the work into little bits, but we should C/R this and decide whether we're willing to have it in the code forever if I get run over by a bus tomorrow.

lib/chef/event_dispatch/dispatcher.rb
jkeiser: @coderanger that won't make a difference though, the issue is we need to know who *originally* called `deprecation` and if we wait until after dispatcher invokes the "real" `deprecation`, the caller will just be the dispatcher.

danielsdeleo: If we do really need a special case, let's not do it inside the metaprogramming. We try to keep the logic inside the metaprogramming stuff as simple as possible. We could instead just define the method normally after the mass-defining all the normally-behaved methods.

jkeiser: My only issue there was copy/pastaing the large amount of stuff in the metaprogramming ... but I can just stick all that in a helper method in this case.

jkeiser: (Though that would make the dispatch stack traces bigger.)

lib/chef/deprecation/warnings.rb
```diff
@@ -26,9 +26,8 @@ def add_deprecation_warnings_for(method_names)
           define_method(name) do |*args|
             message = []
             message << "Method '#{name}' of '#{self.class}' is deprecated. It will be removed in Chef 12."
```
stevendanna: Not part of your changes, but should we s/12/13/ here?

lib/chef/event_dispatch/dispatcher.rb
jkeiser: OK, I think I fixed the metaprogramming thing.

lib/chef/deprecation/warnings.rb
jkeiser: Sure, I just went through and fixed all instance of Chef 12 in the codebase (I'll submit here just for fun).

lib/chef/chef_class.rb
jkeiser: @coderanger renamed `log_deprecation` until I can get the rest of that done :)

danielsdeleo:
LGTM now

lib/chef/event_dispatch/dispatcher.rb
```diff
@@ -25,28 +25,30 @@ def register(subscriber)
...
-        def_forwarding_method(method_name)
+        class_eval <<-EOM
+          def #{method_name}(*args)
+            call_subscribers(#{method_name.inspect}, *args)
+          end
+        EOM
       end
 
+      # Special case deprecation, since it needs to know its caller
+      def deprecation(message, location=caller(2..2)[0])
```
coderanger: Definitely better, though I still think it could be done without this by using `caller(3..3).first` in the actual method.

lib/chef/formatters/doc.rb
```diff
@@ -336,6 +356,16 @@ def provider_requirement_failed(action, resource, exception, message)
         end
       end
 
+      def deprecation(message, location=caller(2..2)[0])
+        if Chef::Config[:treat_deprecation_warnings_as_errors]
```
coderanger: This could probably do with a comment to explain that `super` will eventually result in the exception-ing.

coderanger:
+0.9 on this. good feature and no longer actually opposed to the patch, just don't love the special case in the event dispatch :)

lib/chef/event_dispatch/dispatcher.rb
jkeiser: This assumes there will always be a dispatcher, and always only one. And it would need to be caller(4..4), because there are two method calls in between. Therein lies madness :)

coderanger: Fair, down this path is a stack trace filter. That grossness is probably greater than the gross of having a special case here so ¯\_(ツ)_/¯

thommay:
Yep, this is goodness. :+1:

PR #3856, opened 2015-09-01
Rubygems package provider memory performance improvements

lamont-granquist:


lamont-granquist:
@danielsdeleo what do you think about this?

This has a large impact on mem performance.  181,000 live slots at the end of a run vs. 400,000.

The idea is that the dependency_installer API that we're using pulls down rubygems equivalent of /universe and caches everything.  This changes our code to use the 'gem install' API that posts the gem dep you're trying to solve to rubygems and it only gives you back the gems that you need to satisfy your dep.  Of course this means that instead of hitting rubygems once on the first chef_gem/gem_package that we hit rubygems for every chef_gem/gem_package, this is why I made it a config option.

- I'm not sure I'm using rubygems right
- I'm open to suggestions about the config option and if it should be the default or not, etc, but I think this is right for a minor version patch leaving the default behavior
- I'm unclear about what is going on with this code and what the 'spec_with_source && spec_with_source` line is supposed to be doing....

lamont-granquist:
And needs tests, but I'll try to look at those after deciding if the code looks right or not...

danielsdeleo:
I haven't poked at that area of rubygems code probably since I rewrote this provider to talk to rubygems via ruby, but it seems sane to me. Can probably nuke the pre-rubygems-2.0 code also, I don't think we support anything that far back now.

danielsdeleo:
Also, we should set the default to whatever rubygems does.

lamont-granquist:
Yeah, this is now much closer to what rubygems does with a 'gem install'.  It doesn't pull all the gems like it used to in 1.8.x.

danielsdeleo:
Okay, so I'd like to see:

- Add tests
- nuke pre-2.0 code
- set defaults to match "normal" rubygems behavior

I'm in favor thus far.

lamont-granquist:
I am slightly abusing the API because what I'm hitting gives back a "RequestSet" which is assuming that I'm going to turn around and install the gems as part of a 'gem install' request.  Instead I extract the first spec and toss the results away because what I'm doing is just querying to see which gem satisfies the gem dependency.  I dug through all the DependencyInstaller code though and if you pass a gem_dependency there's no hook where it doesn't pull down all the gems from rubygems when you're trying to just do a simple query.

lamont-granquist:
@danielsdeleo i think this addresses all your concerns.

for the specs, it turns out our unit tests are more end-to-end tests that call load_current_resource and really hit rubygems.org -- and while that might be considered bad, it does mean that there's inherently coverage -- and a lot better coverage than stubbing out the implementation.  that turned up that i wasn't catching the exception and returning nil correctly, so the existing tests seem to work fine to find bugs in the new code...

there's also the functional resource tests which cover this as well.

danielsdeleo:
Looks good, :+1:

lamont-granquist:
@chef/client-core anyone else with review?

thommay:
:+1: sorry I took so long over this. Go for it!

lamont-granquist:
While it works, I'm blocked on perf issues that I don't want to ship.

lamont-granquist:
(computational perf issues rather than memory perf issues)

lamont-granquist:
stalled.

lamont-granquist:
I think this is probably fine as-is.  We inherently have a trade-off between RAM and network round trips here.  The improvements we put in the package provider to make it more idempotent will prevent a lot of the round trips that this was causing when i was testing it before.   This patch probably just needs to be rebased and merged.

lamont-granquist:
rebased and closed in #5203

PR #3858, opened 2015-09-02
Remove duplicate 'Accept' header in spec

juliandunn:
Corrects [this error](https://travis-ci.org/chef/chef/jobs/78307017#L598-L599) in Travis


lamont-granquist:
link to the error is gone now, so we'll just have to trust you...

:+1:

PR #3870, opened 2015-09-03
Monkeypatch Win32::Registry methods delete_key, delete_value

mcquin:
- Adds monkeypatches for Win32::Registry#delete_key and Win32::Registry#delete_value. 
- Rewrites Chef::Win32::Registry#delete_key to use Win32::Registry#delete_key.

(Note: Once this is merged, I will move the [Win32::Registry#write monkeypatch](https://github.com/chef/chef/blob/master/lib/chef/win32/unicode.rb#L62-L87) into `lib/chef/monkey_patches/win32/registry.rb` for consistency. Doing so here is outside the scope of this PR.)

### DIFF
--- a/lib/chef/monkey_patches/win32/registry.rb
+++ b/lib/chef/monkey_patches/win32/registry.rb
@@ -23,8 +23,7 @@ module Win32
       # ::Win32::Registry#delete_key uses RegDeleteKeyW. We need to use
       # RegDeleteKeyExW to properly support WOW64 systems.
       def DeleteKey(hkey, name)
-        arch_mask = win64? ? 0x0100 : 0x0200
-        check RegDeleteKeyExW(hkey, name.to_wstring, KEY_WRITE | arch_mask, 0)
+        check RegDeleteKeyExW(hkey, name.to_wstring, 0, 0)
       end
       
     end
--- a/lib/chef/monkey_patches/win32/registry.rb
+++ b/lib/chef/monkey_patches/win32/registry.rb
@@ -1,0 +1,23 @@
+
+require 'chef/win32/api/registry'
+require 'chef/win32/unicode'
+require 'win32/registry'
+
+module Win32
+  class Registry
+    module API
+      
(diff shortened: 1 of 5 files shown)


mcquin:
Most, if not all, of these spec failures are related to a bad chef-zero gem. I will rebase and re-run tests once the issue is resolved.

Please review @chef/client-windows

lib/chef/monkey_patches/win32/registry.rb
```diff
@@ -0,0 +1,31 @@
+
```
chefsalim: Should have file header

lib/chef/monkey_patches/win32/registry.rb
```diff
@@ -0,0 +1,31 @@
...
+      if RUBY_VERSION =~ /^2\.1/
+        # ::Win32::Registry#delete_value is broken in Ruby 2.1 (up to Ruby 2.1.6).
+        # This should be resolved in a later release (see note #9 in link below).
+        # https://bugs.ruby-lang.org/issues/10820
+        def DeleteValue(hkey, name)
+          check RegDeleteValueW(hkey, name.to_wstring)
+        end
+      end
+
+      # ::Win32::Registry#delete_key uses RegDeleteKeyW. We need to use
```
chefsalim: Should this be under the RUBY_VERSION check as well?

mcquin: No. Even on Ruby 2.0, Win32::Registry#delete_key is implemented with `RegDeleteKey` (see: [Win32::Registry::API#DeleteKey source](http://ruby-doc.org/stdlib-2.0.0/libdoc/fiddle/rdoc/Win32/Registry/API.html))

chefsalim: Ah ok, that is why we had the RegDeleteKeyEx stub

chefsalim:
Looks good! :+1:

PR #3882, opened 2015-09-04
Document LCM certificate configuration example for ps_credential

adamedx:
I added an example to make the DSC LCM configuration prerequisite a little more obvious. cc @chefsalim, @jaym, @smurawski, @claire, @ksubrama.

### DIFF
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -98,7 +98,32 @@ dsc_script 'create-foo-user' do
 end
 ```
 
-Note, you still need to configure the CertificateID in the LCM.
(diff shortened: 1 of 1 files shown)


smurawski:
This example should either call out the other LCM features to configure (configurationmode and rebootnodeifneeded).  Configuration of the lcm is not cumulative, so if you just run the example, any other custom configuration will be reverted to the defaults.

smurawski:
Example of the rest of the LCM configuration - https://github.com/smurawski/chefconf2015_dsc/blob/master/cookbooks/basic_website/recipes/lcm_setup_dsc_script.rb

adamedx:
Ah, didn't know it wasn't cumulative, thank you. I will update.

btm:
@adamedx is this finished?

adamedx:
It's better than finished, it's obsolete since docs were already merged with this.

PR #3894, opened 2015-09-08
Add ruby 2.1

ksubrama:


jaym:
this will take a while to run, we should add more capacity for appveyor

ksubrama:
/cc @chef/client-windows, @jkeiser (for wanting ruby 2.2).

I'll merge when we have enough concurrent instances to not hurt everyone else.

jaym:
Let's rebase this to see how far it gets. Also, you can drop 2.2 for now as we're not shipping it with anything

jaym:
Moving this along :+1: 
We'll do 2.1x64 if we ever decide to ship it, otherwise just jump to 2.2

PR #3896, opened 2015-09-08
Add monkey patch for webrick

jaym:
Ruby 2.1 introduces a regression on Windows in WEBrick.
create_listeners does not throw an exception when we're
already listening on a port. This seems to only be an
issue on Windows. This patch reverts it back to what it
was in Ruby 2.0

It seems the regression was introduced in
https://github.com/ruby/ruby/commit/b1f493dcd1092fe17cccec998e175516ed5c6e47#diff-4b178393150b2b3a5ec9d77eced1f09e

This should fix the failures in the knife integration tests (see #3894)

cc @chef/client-windows @jkeiser

### DIFF
--- a/lib/chef/local_mode.rb
+++ b/lib/chef/local_mode.rb
@@ -15,6 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 require 'chef/config'
+if Chef::Platform.windows?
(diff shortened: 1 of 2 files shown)


mcquin:
:+1:

mwrock:
:+1:

PR #3898, opened 2015-09-09
Modify enforce_ownership_and_permissions_spec to be more unit-like

jaym:
This should fix the remaining failures for ruby 2.1 (#3894). Seems like the passwd struct changed. `describe_changes` also called platform specific things on windows, which started failing.

cc @chef/client-windows


mwrock:
:+1:

mcquin:
LGTM

PR #3900, opened 2015-09-09
Add new option json attributes file to bootstraping

sawanoboly:
replace of #3459

- We can pass jsonfile to bootstrap via --json-attribute-file option.
- raise an exception if both set  json-attributes and json-attribute-file.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -321,6 +321,7 @@ EOS
       end
 
       def render_template
+        @config[:first_boot_attributes].merge!(@config[:first_boot_attributes_from_file]) if @config[:first_boot_attributes_from_file]
         template_file = find_template
         template = IO.read(template_file).chomp
         Erubis::Eruby.new(template).evaluate(bootstrap_context)
@@ -330,7 +331,6 @@ EOS
         if @config[:first_boot_attributes].any? && @config[:first_boot_attributes_from_file]
           raise Chef::Exceptions::BootstrapCommandInputError, jsonstring_and_jsonfile_msg
         end
-        @config[:first_boot_attributes].merge!(@config[:first_boot_attributes_from_file]) if @config[:first_boot_attributes_from_file]
 
         validate_name_args!
 
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -154,7 +154,7 @@ class Chef
         :long => "--json-attributes",
         :description => "A JSON string to be added to the first run of chef-client",
         :proc => lambda { |o| Chef::JSONCompat.parse(o) },
-        :default => {}
+        :default => nil
 
       option :first_boot_attributes_from_file,
         :long => "--json-attribute-file FILE",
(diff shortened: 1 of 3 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -238,6 +244,13 @@ def initialize(argv=[])
         )
       end
 
+      def jsonstring_and_jsonfile_msg
+<<EOS
+You cannot pass both --json-attributes and --json-attribute-file.
+Please pass one or none.
+EOS
+      end
```
lamont-granquist: this text should go into the exception class.

lib/chef/knife/bootstrap.rb
```diff
@@ -308,12 +321,17 @@ def bootstrap_context
       end
 
       def render_template
+        @config[:first_boot_attributes].merge!(@config[:first_boot_attributes_from_file]) if @config[:first_boot_attributes_from_file]
```
lamont-granquist: what would be better here would be to extract an accessor:

```ruby
def first_boot_attributes
   @config[:first_boot_attributes].any? ? @config[:first_boot_attributes] : @config[:first_boot_attributes_from_file]
end
```

If you can make @config[:first_boot_attributes] default to nil if its not passed then you could simplify this to:

```ruby
def first_boot_attributes
   @config[:first_boot_attributes] || @config[:first_boot_attributes_from_file] || {}
end
```

Which would be much more readable.

lamont-granquist:
:+1:

smurawski:
:+1:

thommay:
:+1:

yoshiwaan:
```
knife ec2 server create --help
...
-j, --json-attributes JSON       A JSON string to be added to the first run of chef-client
```
```
knife bootstrap --help
...
-j JSON_ATTRIBS,                 A JSON string to be added to the first run of chef-client
        --json-attributes
```
```
chef solo --help
...
-j JSON_ATTRIBS,                 Load attributes from a JSON file or URL
        --json-attributes
```
```
chef-client --help
-j JSON_ATTRIBS,                 Load attributes from a JSON file or URL
        --json-attributes
```
```
chef-shell --help
    -j JSON_ATTRIBS,                 Load attributes from a JSON file or URL
        --json-attributes
```
I know you just made a separate PR for having a separate command line option for loading a file or using inline JSON but this goes against all of the other Chef commands which load JSON from a file when using -j/--json-attributes.
If anything the new attribute should load inline and -j/--json-attributes should load from a file like all the other commands. Everyone hates switch changes like that though, so I think it's better to check if the value is valid json and use it if it is, otherwise check if it's a file. That way both work.

danielsdeleo:
@yoshiwaan Your point is valid and I would certainly like to be consistent. That said, given the number of arguments you have to pass to `knife bootstrap` and friends, it's guaranteed that folks will have written scripts that run these commands so compatibility has to be maintained. I wouldn't mind some sort of detection mechanism, but we'd need to be _very_ careful with the heuristic we use. For example, if the user is trying to input JSON as a string on the command line, it's quite likely they will have a typo that could make the JSON invalid, but if we aren't careful, we'd emit a confusing error message like `no such file '{"some_invalid_json": [ }'`. My preference is to merge this, then see if we can figure out an auto detect mechanism which will have a good experience for both success and failure cases. We could then make a decision about how to clean things up for Chef 13, if it's worth it.

PR #3901, opened 2015-09-09
Add tmux-split parameter to knife ssh

permyakovsv:
*Add ability to use tmux split window


thommay:
this feels more like it ought to be an argument than a completely separate method of running. (ie: `knife ssh --tmux-split true ...` rather than `knife ssh ... tmux-split`)

thommay:
perfect, thanks :+1:

ranjib:
:+1:

PR #3907, opened 2015-09-10
Allow windows_service_spec to only run in appveyor

jaym:
We thing that these tests may be causing instability on our
jenkins slaves.

cc @sersut @chef/client-windows


sersut:
:+1:

PR #3909, opened 2015-09-10
remove pending reboot check for HKLM\SOFTWARE\Microsoft\Updates\UpdateExeVolatile

mwrock:
The pending reboot functional tests fail on checks for HKLM\SOFTWARE\Microsoft\Updates\UpdateExeVolatile with the value of 1-3. A clean 2k8r2 sp1 machine has a 3 which persists after reboots. The only way to remove it is to do so manually.

This key was introduced in the windows package installer on windows server 2003 and windows XP. It indicates that a previously installed update requires a reboot in order to complete. See https://support.microsoft.com/en-us/kb/832475

There are many folks running into issues installing various products and being told that a reboot is required and they get stuck because they receive this message after every reboot. This seems to occur most commonly with older office or sharepoint products. The guidance tends to be to delete this key. Apparently it can be orphaned and even Microsoft includes this tip in their support sites. See https://technet.microsoft.com/en-us/library/Cc164360(v=EXCHG.80).aspx

The more [popular modern reboot detection scripts](http://gallery.technet.microsoft.com/scriptcenter/Get-PendingReboot-Query-bdb79542) do not include this key and I think the fix is going to be removing it from our reboot check expecially since we no longer support server 2003.

### DIFF
--- a/lib/chef/dsl/reboot_pending.rb
+++ b/lib/chef/dsl/reboot_pending.rb
@@ -49,7 +49,8 @@ class Chef
 
           # The mere existence of the UpdateExeVolatile key should indicate a pending restart for certain updates
           # http://support.microsoft.com/kb/832475
-          (registry_key_exists?('HKLM\SOFTWARE\Microsoft\Updates\UpdateExeVolatile') &&
+          Chef::Platform.windows_server_2003? &&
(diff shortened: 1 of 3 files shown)


chefsalim:
LGTM :+1:

mwrock:
Per suggestion fro @adamedx I kept this check if on server 2003. I also ended up refactoring the functional tests because the former state cleanup was flawed.

jaym:
:+1:

PR #3911, opened 2015-09-11
We shouldn't be subclassing Struct.new - if the file is loaded twice we get a superclass mismatch error

tyler-ball:
This caused an incident, so this is a simple cleanup task.

\cc @randomcamel @danielsdeleo @lamont-granquist

### DIFF
--- a/lib/chef/resource/resource_notification.rb
+++ b/lib/chef/resource/resource_notification.rb
@@ -20,7 +20,15 @@ require 'chef/resource'
 
 class Chef
   class Resource
-    class Notification < Struct.new(:resource, :action, :notifying_resource)
+    class Notification
+
+      attr_accessor :resource, :action, :notifying_resource
+
+      def initialize(resource, action, notifying_resource)
+        @resource = resource
+        @action = action
+        @notifying_resource = notifying_resource
+      end
 
       def duplicates?(other_notification)
         unless other_notification.respond_to?(:resource) && other_notification.respond_to?(:action)
@@ -102,6 +110,11 @@ is defined near #{resource.source_line}
         F
         err.set_backtrace(e.backtrace)
         raise err
+      end
+
+      def ==(other)
+        return false unless other.is_a?(self.class)
+        other.resource == resource && other.action == action && other.notifying_resource == notifying_resource
       end
 
     end
--- a/spec/unit/resource/resource_notification_spec.rb
+++ b/spec/unit/resource/resource_notification_spec.rb
@@ -19,149 +19,148 @@ require 'spec_helper'
 require 'chef/resource/resource_notification'
 
 describe Chef::Resource::Notification do
-  before do
-    @notification = Chef::Resource::Notification.new(:service_apache, :restart, :template_httpd_conf)
-  end
+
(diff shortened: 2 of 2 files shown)


coderanger:
:+1: lolrubby

danielsdeleo:
IMO this isn't necessary, we just need to not load identical files (see https://github.com/chef/appbundler/pull/18) and we'll be fine. The superclass mismatch thing is just one symptom (of many; for example there's hundreds of redefined constant warnings and such) of this problem. Even if we change the chef code so things "work" if we load the same files multiple times, it'll still be an unacceptably poor experience.

I think subclassing `Struct.new` is a valid thing to do and we don't need to try to avoid it just so we can load the same file twice, because we should never be loading the same file twice.

coderanger:
+1 for fixing this upstream, but its a relatively minor bit of bloat and const redefs are warnings vs. superclass mismatch which I think is a fatal?

danielsdeleo:
IMO it doesn't matter that this causes an exception which leads to the termination of the program, because it only does so in the presence of a ship-blocking bug. I think it's actually kind of helpful, because it ensures that a multi-load bug will crash hard and be found (rather than spewing warnings which may not be noticed if they occur in unattended operation).

coderanger:
Fair point :)

lib/chef/resource/resource_notification.rb
```diff
@@ -20,7 +20,15 @@
 
 class Chef
   class Resource
-    class Notification < Struct.new(:resource, :action, :notifying_resource)
+    class Notification
+
+      attr_accessor :resource, :action, :notifying_resource
+
+      def initialize(resource, action, notifying_resource)
```
lamont-granquist: if back-compat is important these should all default to nil.

lamont-granquist:
Dunno that's idiomatic use of Struct.new, but at the same time this class is more than just a Struct.new and general practice is that you shouldn't take this shortcut for things that aren't solely simple containers...  I'm kinda 50/50.

tyler-ball:
@danielsdeleo I realize this helped us catch the underlying issue but I would rather set up a specific test for that.

I think when I originally wrote the class that it was simple enough to make extending a `Struct` make sense.  But there is enough implementation in it now that I think its should just be its own class.  Thats really why I am removing the superclass.

lib/chef/resource/resource_notification.rb
tyler-ball: I think they should be required args.  The only place we use it is https://github.com/chef/chef/blob/master/lib/chef/resource.rb#L1527-L1535.  I know this is an API change but I don't think anyone would be using this class besides us.

randomcamel:
This seems like a reasonable enough change: it normalizes an anomalous corner of our considerable OO hierarchy. It's not the worst thing in the world to have something in `chef/chef` that explodes when used in a bad way (like loading 2 different copies of the source), but if the only way to do that is to have a `Struct.new` descendant somewhere, we could add one such that its defensive purpose is explicit.

lamont-granquist:
:+1:

thommay:
yeah, let's do it :+1:

PR #3913, opened 2015-09-12
Policyfile node integration

danielsdeleo:
This patch integrates `policy_name` and `policy_group` into the node object, with the following behavior changes:

- Policyfile attributes can be set on `Chef::Node` objects
- Policyfile attributes are included in the serialization format (JSON)
- Policyfile attributes are persisted between runs, when using a Chef Server that supports them
- Policyfile attributes can be set in the JSON file passed to `chef-client -j JSON_FILE`
- Policyfile attributes are shown by `knife search` and `knife node show`, if present
- Policyfile attributes are searched in the `knife search` "fuzzy" query (which is invoked when the given query doesn't match the `key:value` format).
- The `use_policyfile` feature flag in config is effectively a no-op.

See also the related Chef Server change, which will be released in Chef Server 12.3: https://github.com/chef/chef-server/pull/514

#### Tasks:

- [x] Add getters and setters for `policy_name` and `policy_group` to Node
- [x] Add `policy_name` and `policy_group` to node json, with fallback for older chef servers
- [x] dynamically select policyfile mode or not based on presence of policyfile attributes on node, `-j JSON`, etc.
- [x] propagate `policy_name` and `policy_group` to and from `-j JSON`, `Chef::Config` and node
- [x] add policyfile stuff to `knife node show`
- [x] add policyfile stuff to default query in `knife search`
- [x] ensure compatibility mode still works (e.g., test kitchen)

#### Punted:
- warn if `use_policyfile true` is in config (deprecate and remove in 13): I don't think there's an overwhelming case to nag and eventually break users over this.
- support policy_name/policy_group in bootstrap: This will be a separate pull request, as this one is already very large.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -339,10 +339,27 @@ module ChefConfig
     # most of our testing scenarios)
     default :minimal_ohai, false
 
+    ###
+    # Policyfile Settings
+    #
     # Policyfile is a feature where a node gets its run list and cookbook
     # version set from a single document on the server instead of expanding the
     # run list and having the server compute the cookbook version set based on
     # environment constraints.
+    #
+    # Policyfiles are auto-versioned. The user groups nodes by `policy_name`,
+    # which generally describes a hosts's functional role, and `policy_group`,
+    # which generally groups nodes by deployment phase (a.k.a., "environment").
+    # The Chef Server maps a given set of `policy_name` plus `policy_group` to
+    # a particular revision of a policy.
+
+    default :policy_name, nil
+    default :policy_group, nil
+
+    # During initial development, users were required to set `use_policyfile true`
+    # in `client.rb` to opt-in to policyfile use. Chef Client now examines
+    # configuration, node json, and the stored node to determine if policyfile
+    # usage is desired. This flag is still honored if set, but is unnecessary.
     default :use_policyfile, false
 
     # Policyfiles can be used in a native mode (default) or compatibility mode.
@@ -355,6 +372,16 @@ module ChefConfig
     # compatibility mode. Compatibility mode remains available so you can use
     # policyfiles with servers that don't yet support the native endpoints.
     default :policy_document_native_api, true
+
+    # When policyfiles are used in compatibility mode, `policy_name` and
+    # `policy_group` are instead specified using a combined configuration
+    # setting, `deployment_group`. For example, if policy_name should be
+    # "webserver" and policy_group should be "staging", then `deployment_group`
+    # should be set to "webserver-staging", which is the name of the data bag
+    # item that the policy will be stored as. NOTE: this setting only has an
+    # effect if `policy_document_native_api` is set to `false`.
+    default :deployment_group, nil
+
 
     # Set these to enable SSL authentication / mutual-authentication
     # with the server
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -510,6 +510,14 @@ class Chef
         #Render correctly for run_list items so malformed json does not result
         "run_list" => @primary_runlist.run_list.map { |item| item.to_s }
       }
+      # Chef Server rejects node JSON with extra keys; prior to 12.3,
+      # "policy_name" and "policy_group" are unknown; after 12.3 they are
+      # optional, therefore only including them in the JSON if present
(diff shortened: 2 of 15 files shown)


danielsdeleo:
For the commit titled "Add policyfile attrs to node create/save": in chat, it was suggested that since policyfiles are "beta" it would be okay to just not implement compatibility at all. I wouldn't mind doing that, but, on the other hand, it's not a ton of work to implement the compatibility fallback and it doesn't look that dangerous to do so. Therefore, what I've implemented is backcompat with Chef Server 12.2, but also added a test to assert we remove that compatibility in 13.0, on the grounds that we don't need to maintain full major version cross-compat since policyfiles are beta when we're making the initial change. I'm okay to change it if there's a strong feeling that we'd rather not accept the added complexity in node save right now.

lamont-granquist:
Just a nit on where to put the comments about the backcompat.  I think it goes better in the path where the branching logic is than on the backcompat methods.

danielsdeleo:
K, I'll move the comments, no problem. Agree we don't really *need* to have back compete here, but also it was pretty easy to do and doesn't seem to add a lot of risk or a lot of future cleanup effort, so that pushed me over the edge to "sure, why not."

danielsdeleo:
@chef/client-core this one is ready for review now. I have attempted to make the pull request reviewable on a commit-by-commit basis since the aggregate change is fairly large.

chef-config/lib/chef-config/config.rb
```diff
@@ -339,10 +339,27 @@ def self.path_accessible?(path)
...
     # environment constraints.
+    #
+    # Policyfiles are auto-versioned. The user groups nodes by `policy_name`,
+    # which generally describes a hosts's functional role, and `policy_group`,
+    # which generally groups nodes by deployment phase (a.k.a., "environment").
+    # The Chef Server maps a given set of `policy_name` plus `policy_group` to
+    # a particular revision of a policy.
+
+    default :policy_name, nil
+    default :policy_name, nil
```
lamont-granquist: policy_group?

lib/chef/node.rb
```diff
@@ -132,6 +137,26 @@ def chef_environment=(environment)
 
     alias :environment :chef_environment
 
+    def policy_name(arg=NULL_ARG)
```
lamont-granquist: YARD for all these methods?

lib/chef/policy_builder/dynamic.rb
```diff
@@ -0,0 +1,154 @@
+#
+# Author:: Adam Jacob (<adam@opscode.com>)
+# Author:: Tim Hinderliter (<tim@opscode.com>)
+# Author:: Christopher Walters (<cw@opscode.com>)
```
lamont-granquist: uh, lol =)

lib/chef/policy_builder/dynamic.rb
```diff
@@ -0,0 +1,154 @@
...
+          end
+        select_implementation(node)
+        implementation.finish_load_node(node)
+        node
+      rescue Exception => e
+        events.node_load_failed(node_name, e, config)
+        raise
+      end
+
+      ## Delegated Methods ##
```
lamont-granquist: `extend Forwardable` and use `def_delegators`?

lamont-granquist: ah i think i see, you want to go in through the accessor not the ivar and raise appropriately, nm...

lamont-granquist:
:+1: 

mostly could just use sprinkling some more YARD around on it i think.

lib/chef/policy_builder/dynamic.rb
danielsdeleo: Forwardable can do that, too, you just use `def_delegator :method_for_delegated_thing, :delegated_method`. I'd thought about that and wasn't sure it was better at the time I wrote this, but I've changed my mind, and will use Forwardable here.

danielsdeleo:
@lamont-granquist added a ton of comments. GTG?

danielsdeleo:
Also need another +1

ranjib:
:+1: i'll test it out in compat mode as well

danielsdeleo:
Thanks. I will merge when travis is happy w/ the rebase. Let me know if you find anything in your testing

PR #3918, opened 2015-09-14
Fix failing specs on Windows

jaym:
The verify tests were not correctly mocking things out. This was causing the tests to actually shellout
for verify, it this started breaking because of an implementation detail that was part of the mixlib-shellout 2.2.1 release

### DIFF
--- a/spec/support/shared/unit/provider/file.rb
+++ b/spec/support/shared/unit/provider/file.rb
@@ -465,8 +465,8 @@ shared_examples_for Chef::Provider::File do
         t
       }
 
-      let(:verification) { double("Verification") }
-      let(:verification_fail) { double("Verification Fail") }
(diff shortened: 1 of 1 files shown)


tyler-ball:
:+1:

mwrock:
:+1:

spec/support/shared/unit/provider/file.rb
```diff
@@ -466,10 +466,12 @@ def unlink
       }
 
       let(:verification) { double("Verification") }
+      let(:verification_fail) { double("Verification Fail") }
```
danielsdeleo: I'd like to start using `instance_double` wherever possible. https://www.relishapp.com/rspec/rspec-mocks/v/3-3/docs/verifying-doubles/using-an-instance-double

danielsdeleo:
I'd prefer to use `instance_double` wherever we can, but :+1:

Issue #3897, opened 2015-09-08
12.5 regressions


jaymzh:
While trying to test @lamont-granquist 's new attribute code, I also was playing around with 12.5 current and noticed that code we have in a provider that does:

```
def load_current_resource
  @current_resource.foo = false
end
...

action :foo do
  execute 'something' do
    only_if { current_resource.foo }
  end
end
```

Fails with `undefined method 'foo' for nil:NilClass` - but works in 12.4.1

lamont-granquist:
Looks like it might be related to resources and maybe @jkeiser?

The unsimplified example had some references to Mixlib::ShellOut so I was wondering if it might have something to do with @jaym's changes, but if the simplified example fails like that it's probably nothing to do with that...

lamont-granquist:
Tagging @thommay as well.

coderanger:
I think we changed the binding context on not_if/only_if? Can you check if this works?

```ruby
action :foo do
  r = current_resource
  execute 'something' do
    only_if { r.foo }
  end
end
```

If that still works, it would definitely point at a context issue, or something with the enclosing_provider forwarding code.

jkeiser:
I suspect it has to do with [this](https://github.com/chef/chef/blob/master/lib/chef/resource.rb#L1413). I'll take a look at a patch when my head clears.

adamedx:
@mcquin is going to have a look.

lamont-granquist:
So the problem is that adding Chef::Resource::Execute#current_resource has shadowed over the outer Chef::Provider#current_resource in the binding.

Can we rename Chef::Resource#current_resource instead of ripping it out, or think of some other solution?

jaymzh:
This is still broken with the latest nightly.

chef-12.5.0~current.0+20150821082812.git.165.c832f38-1.el6.x86_64

lamont-granquist:
nightlies are busted right now, that's 8/21, not 9/21...

jaymzh:
ah. well... if someone wants to throw a nightly at me I can test. Should the 12.4.2 RPM I was testing for Mathilde include this?

jkeiser:
@jaymzh try http://opscode-omnibus-packages-current.s3.amazonaws.com/el/6/x86_64/chef-12.5.0~current.0+20150909213019.git.314.9cf93f1-1.el6.x86_64.rpm

jaymzh:
@jkeiser `ERROR 404: Not Found`

jkeiser:
[...]

PR #3921, opened 2015-09-15
Rename current_resource method on Chef::Resource as current_value.

mcquin:
Resolves https://github.com/chef/chef/issues/3897

@jkeiser @jaymzh @chef/client-core

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1406,7 +1406,7 @@ class Chef
     # @return A new copy of the resource, with values filled in from the actual
     #   current value.
     #
-    def current_resource
+    def current_value
       provider = provider_for_action(Array(action).first)
       if provider.whyrun_mode? && !provider.whyrun_supported?
         raise "Cannot retrieve #{self.class.current_resource} in why-run mode: #{provider} does not support why-run"
--- a/spec/integration/recipes/resource_load_spec.rb
+++ b/spec/integration/recipes/resource_load_spec.rb
@@ -63,17 +63,17 @@ describe "Resource.load_current_value" do
       end
 
       it "current_resource is passed name but not x" do
-        expect(resource.current_resource.x).to eq 'loaded 2 (name=blah)'
+        expect(resource.current_value.x).to eq 'loaded 2 (name=blah)'
       end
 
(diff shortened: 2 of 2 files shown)


lamont-granquist:
dammit i really want that method, we also RFC'd this and approved it.

coderanger:
@lamont-granquist Go go gadget Chef 13? :-(

mcquin:
We can't name it `current_resource` and maintain its desired behavior without introducing confusion with `Chef::Provider`'s `current_resource`. @jkeiser and I thought about renaming it to `current_value`, but we were hesitant to introduce a new method name because method names are forever.

What do you think about calling it `current_value`?

coderanger:
I was the one that requested it be `resource` to reinforce the connect with `load_current_resource`. I still stand by that :-/

mcquin:
On a provider, it's `load_current_resource` and `current_resource`. On a resource, we have `load_current_value`. Why shouldn't it be `current_value`?

lamont-granquist:
well, there's something a little odd with it:

```
res = file "/tmp/foo.txt" do
  mode 0644
end
```

returns a resource -- the new_resource

```
res = file "/tmp/foo.txt" do
  mode 0644
end.resource
```

then returns a resource -- the current_resource

but when i read that last one it reads to me like it should just return itself...

"if i build a resource and call the resource method to return a resource i get...  a different resource?"

lamont-granquist:
This morning 'current_value' is making sense to me.  What is being returned isn't really the current_resource either since its a resource that we construct which is very much like the current_resource but its not `equal?` to the current_resource so calling it something else seems fine.

mcquin:
Method renamed `current_value`.

jkeiser:
Better than not having it. And it fits reasonably well with what you'd need to do for other types. :+1: and *sooooo* glad we don't have to remove it, it makes the world so nice.

thommay:
:+1:

lamont-granquist:
@jkeiser you should probably amend the RFC

jaymzh:
Late to the game here... but :+1: 
I'll test a nightly build today if I get a chance.

PR #3931, opened 2015-09-18
Create 12.4.2 of Chef with ohai dep restriction

jkeiser:


schisamo:
Assuming Travis is happy :+1:. I assume you'll need to push a 12.4.2 to Rubygems.org also right?

chef.gemspec
```diff
@@ -15,13 +15,13 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.0.0"
 
-  s.add_dependency "chef-config", "= #{Chef::VERSION}"
+  s.add_dependency "chef-config", "= 12.4.1"
```
lamont-granquist: can't you just release chef-config 12.4.2 at the same time and avoid additional confusion here?

lamont-granquist:
And the problem is that you need to release both the gem and the omnibus artifact or else you break people still using vagrant-omnibus and such, you can't take a shortcut around the release process.

lamont-granquist:
(vagrant-omnibus still has some ancient code in it which pulls rubygems to get a list of valid chef versions and then will pick the latest and attempt to install it with omnibus, which will fail if you release to rubygems and do not release the omnibus artifact, its very gross, but still trolls us every time we release to rubygems ahead of the omnibus release)

jkeiser:
@lamont-granquist what I'm hearing is that we need to release rubygems and omnibus at the same time for it to work, yes? If we do that we'll be OK?

chef.gemspec
jkeiser: Sure, I can do that.

lamont-granquist:
Correct.

lamont-granquist:
Just do a full release of 12.4.2 and you'll be fine, don't cut corners.

jkeiser:
No corners shall be cut :) Also, patch is updated to release chef-config 12.4.2 as well.

lamont-granquist:
coolness

:+1:

tyler-ball:
:+1:

Issue #3934, opened 2015-09-21
Service resource fails if systemd unit files are not present on ubuntu 15.04


ranjib:
```ruby
service 'go-agent' do
    action [:enable, :start]
end
```
trace
```sh
Generated at 2015-09-20 19:53:02 -0700
NoMethodError: service[go-agent] (go_cd::agent line 76) had an error: NoMethodError: undefined method `+' for false:FalseClass
/opt/chef/embedded/apps/chef/lib/chef/platform/service_helpers.rb:134:in `platform_has_systemd_unit?'
/opt/chef/embedded/apps/chef/lib/chef/platform/service_helpers.rb:98:in `config_for_service'
/opt/chef/embedded/apps/chef/lib/chef/provider/service/debian.rb:35:in `supports?'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `block in supported_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `select'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:79:in `supported_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:91:in `prioritized_handlers'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:116:in `maybe_dynamic_provider_resolution'
/opt/chef/embedded/apps/chef/lib/chef/provider_resolver.rb:61:in `resolve'
/opt/chef/embedded/apps/chef/lib/chef/resource.rb:1619:in `provider_for_action'
/opt/chef/embedded/apps/chef/lib/chef/resource.rb:582:in `run_action'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:49:in `run_action'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `block (2 levels) in converge'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `each'
/opt/chef/embedded/apps/chef/lib/chef/runner.rb:81:in `block in converge'
/opt/chef/embedded/apps/chef/lib/chef/resource_collection/resource_list.rb:83:in `block in execute_each_resource'
/opt/chef/embedded/apps/chef/lib/chef/resource_collection/stepable_iterator.rb:116:in `call'
/opt/chef/embedded/apps/chef/lib/chef/resource_collection/stepable_iterator.rb:116:in `call_iterator_block'
[...]

PR #3935, opened 2015-09-21
Return empty array if no service  unit files found on systemd backed distros

ranjib:
https://github.com/chef/chef/issues/3934

### DIFF
--- a/lib/chef/platform/service_helpers.rb
+++ b/lib/chef/platform/service_helpers.rb
@@ -126,7 +126,7 @@ class Chef
           # this splits off the suffix after the last dot to return "sshd"
           services += services.select {|s| s.match(/\.service$/) }.map { |s| s.sub(/(.*)\.service$/, '\1') }
         rescue Mixlib::ShellOut::ShellCommandFailed
-          false
+          []
         end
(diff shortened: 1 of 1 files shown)


lamont-granquist:
:+1: 

could use a spec test to not regress.

lamont-granquist:
@chef/client-core could use this getting reviewed and merged for Chef 12.5.0

danielsdeleo:
Yup, could use a spec to prevent regressions. Also, does the command have specific exit codes for different failure conditions or any other characteristics we could use to make sure when we `rescue Mixlib::ShellOut::ShellCommandFailed` we're only dealing with the case that no services are found?

jkeiser:
@ranjib @lamont-granquist I went through and broadened the provider_resolver tests so they could include this case. Now instead of mocking the service_provider results, we mock the files and commands that get run by them (mocking the outside world). This allowed me to add a test for "systemctl list-unit-tests returns exit code 1".

This branch has your fix and the test fixes: https://github.com/chef/chef/pull/3972

ranjib:
i'll close this then.

PR #3936, opened 2015-09-21
Windows cli tools should have color true by default

adamedx:
Older terminal applications on Windows displayed garbage characters instead of altering the color of text in the terminal as a response to color escape sequences. In Win2k12 and newer, the default Windows terminal displays colored text correctly. To reflect this, the defaults on chef-cient, chef-apply, and knife are being changed to be like all other operating systems and use color by default. cc @jaym, @mwrock, @chefsalim, @ksubrama, @McQuin.

### DIFF
--- a/spec/unit/application/client_spec.rb
+++ b/spec/unit/application/client_spec.rb
@@ -288,9 +288,9 @@ describe Chef::Application::Client, "configure_chef" do
     ARGV.replace(@original_argv)
   end
 
-  it "should set the colored output to false by default on windows and true otherwise" do
+  it "should set the colored output to true by default on windows and true on all other platforms as well" do
     if windows?
-      expect(Chef::Config[:color]).to be_falsey
+      expect(Chef::Config[:color]).to be_truthy
     else
       expect(Chef::Config[:color]).to be_truthy
     end
--- a/lib/chef/application/apply.rb
+++ b/lib/chef/application/apply.rb
@@ -100,7 +100,7 @@ class Chef::Application::Apply < Chef::Application
   option :color,
     :long         => '--[no-]color',
     :boolean      => true,
-    :default      => !Chef::Platform.windows?,
+    :default      => true,
     :description  => "Use colored output, defaults to enabled"
 
   option :minimal_ohai,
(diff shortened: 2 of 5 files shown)


coderanger:
Does appveyor support colored output? Not sure how much it matters, but if they don't we could bug them a bunch :)

spec/unit/application/client_spec.rb
```diff
@@ -275,9 +275,9 @@
     ARGV.replace(@original_argv)
   end
 
-  it "should set the colored output to false by default on windows and true otherwise" do
+  it "should set the colored output to true by default on windows and true on all other platforms as well" do
```
mwrock: I don't think we need a conditional here at all anymore. The appveyor run should adequately test this scenario.

mwrock:
I'd make a spall change to the spec's but even still LGTM. :+1:

lamont-granquist:
We still support 2k8r2 and such though which still gets this wrong?

Ideally, we'd switch on node[:platform_version] but that would have to be done after ohai runs...

adamedx:
I'll see what happens on win2k8r2 @lamont-granquist, but even if it looks bad there, it's ok -- Win2k12 and higher are the priority. If the output looks funny, it can always be turned off.

adamedx:
@coderanger actually, I think appveyor does have colored output, which is a great thing.

lamont-granquist:
Okay, doesn't affect me since if I'm using 2008r2 or whatevs I'll certainly be fixing that problem...  =)

:+1:

adamedx:
Yeah, I want a pic of you using Win2k8R2 @lamont-granquist -- writing batch scripts.

So I tested Win2k8R2 with color, and it actually works to the extent that green shows up for actions executed during a Chef run. I didn't get any strange interpolations of escape sequences, though I'm sure there are cases that aren't handled by Windows. It's reasonable enough that I'm not worried about it as a default, and it can always be turned off at any  rate.

Thanks everyone!

thommay:
:+1:

smurawski:
:+1:

PR #3938, opened 2015-09-21
only apt and yum support Arrays of packages

lamont-granquist:
Supporting Arrays in the subclass means that it 'looks' like other
providers all support Arrays when they don't.  By correcting this
we make package resources that aren't multipackage-aware to blow up
at compile time, instead of having some random awful exception when
converging the provider.  This both gets the input validation correct
and allows rescuing the exception and falling back to handle it.

(See the multipackage cookbook for how we use this to duck-type chef 11
vs chef 12 yum_package/apt_package)

### DIFF
--- a/lib/chef/resource/apt_package.rb
+++ b/lib/chef/resource/apt_package.rb
@@ -30,6 +30,24 @@ class Chef
         @default_release = nil
       end
 
+      # override superclass and support arrays
+      def package_name(arg=nil)
+        set_or_return(
+          :package_name,
+          arg,
+          :kind_of => [ String, Array ]
+        )
+      end
+
+      # override superclass and support arrays
+      def version(arg=nil)
+        set_or_return(
+          :version,
+          arg,
+          :kind_of => [ String, Array ]
+        )
+      end
+
       def default_release(arg=nil)
         set_or_return(
           :default_release,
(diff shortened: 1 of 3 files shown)


jaymzh:
This won't work because it means you can't juyst do `package [...]` right?

lamont-granquist:
no, its an abstract base class on 12.0

`package [ "foo" ]` will create a Chef::Resource::YumPackage or Chef::Resource::AptPackage

lamont-granquist:
of course the unit tests are still based on Chef 11 and completely wrong now...

coderanger:
This is not correct, both `python_package` and `node_package` in my code support multi-installs and this would break both of them I think. If multi-installs are going to be the standard (and I see no reason for it not to be) we should put the exceptions in the exceptional cases.

jaymzh:
:100:  to what @coderanger said. We should be pushing to multipackage not away from it.

coderanger:
Also it is worth noting that stuff like this should probably be valid still?

```ruby
# On non-RHEL/Debuntu
package %w{rake rails} do
  provider Chef::Provider::Package::Rubygems
end
```

lamont-granquist:
worked around this in my cookbook rather than in core chef.

PR #3937, opened 2015-09-21
Safely clean up Win32 namespace after specs

mcquin:
Calling `Object.send(:remove_const, 'Win32')` in spec after blocks causes failures in specs which include libraries with rely on Win32.

PR #3939, opened 2015-09-21
Lazy load MSI provider, add check for MSI support

chefsalim:
This change loads the MSI Installer API only if the platform supports it.  This is currently mostly an issue for Windows Nano.  The current check for MSI uses the registry directly - a longer term change will probably use a Ohai plug-in or other mechanism.

cc @chef/client-windows @chef/client-core

### DIFF
--- a/lib/chef/platform/query_helpers.rb
+++ b/lib/chef/platform/query_helpers.rb
@@ -58,6 +58,22 @@ class Chef
         return nano == 1
       end
 
+      def supports_msi?
+        return false unless windows?
+        require 'win32/registry'
+
+        key = "System\\CurrentControlSet\\Services\\msiserver"
+        access = ::Win32::Registry::KEY_QUERY_VALUE
+
+        begin
+          ::Win32::Registry::HKEY_LOCAL_MACHINE.open(key, access) do |reg|
+            true
+          end
+        rescue ::Win32::Registry::Error
+          false
+        end
+      end
+
       def supports_powershell_execution_bypass?(node)
         node[:languages] && node[:languages][:powershell] &&
           node[:languages][:powershell][:version].to_i >= 3
--- a/spec/unit/platform/query_helpers_spec.rb
+++ b/spec/unit/platform/query_helpers_spec.rb
@@ -102,6 +102,55 @@ describe "Chef::Platform#windows_nano_server?" do
   end
 end
 
+describe "Chef::Platform#supports_msi?" do
+  include_context "Win32"  # clear and restore Win32:: namespace
+
+  let(:key) { "System\\CurrentControlSet\\Services\\msiserver" }
+  let(:key_query_value) { 0x0001 }
+  let(:access) { key_query_value }
+  let(:hive) { double("Win32::Registry::HKEY_LOCAL_MACHINE") }
+  let(:registry) { double("Win32::Registry") }
+
+  before(:all) do
+    Win32::Registry = Class.new
+    Win32::Registry::Error = Class.new(RuntimeError)
+  end
+
+  before do
+    Win32::Registry::HKEY_LOCAL_MACHINE = hive
+    Win32::Registry::KEY_QUERY_VALUE = key_query_value
+  end
+
+  after do
+    Win32::Registry.send(:remove_const, 'HKEY_LOCAL_MACHINE') if defined?(Win32::Registry::HKEY_LOCAL_MACHINE)
+    Win32::Registry.send(:remove_const, 'KEY_QUERY_VALUE') if defined?(Win32::Registry::KEY_QUERY_VALUE)
+  end
+
+  it "returns false early when not on windows" do
+    allow(ChefConfig).to receive(:windows?).and_return(false)
+    expect(Chef::Platform).to_not receive(:require)
+    expect(Chef::Platform.supports_msi?).to be false
+  end
+
+  it "returns true when the registry key exists" do
+    allow(ChefConfig).to receive(:windows?).and_return(true)
+    allow(Chef::Platform).to receive(:require).with('win32/registry')
+    expect(Win32::Registry::HKEY_LOCAL_MACHINE).to receive(:open).
+      with(key, access).
+      and_yield(registry)
+    expect(Chef::Platform.supports_msi?).to be true
+  end
+
+  it "returns false when the registry key does not exist" do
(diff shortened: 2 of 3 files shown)


lib/chef/platform/query_helpers.rb
```diff
@@ -58,6 +58,22 @@ def windows_nano_server?
...
+      def supports_msi?
+        return false unless windows?
+        require 'win32/registry'
+
+        key = "System\\CurrentControlSet\\Services\\msiserver"
+        access = ::Win32::Registry::KEY_QUERY_VALUE
+
+        begin
+          ::Win32::Registry::HKEY_LOCAL_MACHINE.open(key, access) do |reg|
+           true
```
mcquin: [nitpick] one more space, for consistent indenting

spec/unit/platform/query_helpers_spec.rb
```diff
@@ -114,6 +114,26 @@
...
+  let(:error_code) { 0 }
+
+  it "returns false early when not on windows" do
+    allow(ChefConfig).to receive(:windows?).and_return(false)
+    expect(Chef::Platform.supports_msi?).to be false
+  end
+
+  it "returns true when the registry key exists" do
+    allow(ChefConfig).to receive(:windows?).and_return(true)
+    allow(Win32::Registry::HKEY_LOCAL_MACHINE).to receive(:open).
```
mcquin: You'll need to define `Win32::Registry::HKEY_LOCAL_MACHINE`. Because `:open` is a private method for some objects, I chose to use a double [defined here](https://github.com/chef/chef/blob/master/spec/unit/platform/query_helpers_spec.rb#L38) and [used here](https://github.com/chef/chef/blob/master/spec/unit/platform/query_helpers_spec.rb#L56)

mcquin: Additionally, you'll either have to wait for [this PR](https://github.com/chef/chef/pull/3937) to be merged and do an `include_context "Win32"` [like this](https://github.com/chef/chef/blob/mcquin/remove-const/spec/unit/platform/query_helpers_spec.rb#L34-L35), or you need to include these [before](https://github.com/chef/chef/blob/master/spec/unit/platform/query_helpers_spec.rb#L41-L58) and [after](https://github.com/chef/chef/blob/master/spec/unit/platform/query_helpers_spec.rb#L60-L68) blocks.

It'd be easiest if you waited until the PR is merged. ;)

chefsalim: Hmm, the test seems to succeed on Windows though?

Travis failure shows Win32 undefined - is that a separate issue?
  1) Chef::Platform#supports_msi? returns true when the registry key exists
     Failure/Error: allow(Win32::Registry::HKEY_LOCAL_MACHINE).to receive(:open).
     NameError:
       uninitialized constant Win32

spec/unit/platform/query_helpers_spec.rb
```diff
@@ -114,6 +114,26 @@
   end
 end
 
+describe "Chef::Platform#supports_msi?" do
+  let(:key) { "System\\CurrentControlSet\\Services\\msiserver" }
+  let(:key_query_value) { 0x0001 }
+  let(:access) { key_query_value }
+  let(:error_code) { 0 }
```
mcquin: It doesn't look like this is being used.

spec/unit/platform/query_helpers_spec.rb
chefsalim: Sure, I can wait for the other PR to be merged.

mcquin: re: https://github.com/chef/chef/pull/3939#discussion_r40037890 that's due to `require 'win32/registry'` only happening on windows.

mcquin: Merged https://github.com/chef/chef/pull/3937. After you rebase all you'll need is to add `include_context "Win32"` at the top of this `describe` block, and then mock `Win32::Registry::HKEY_LOCAL_MACHINE` in some way.

danielsdeleo: @mcquin's suggestion about the mocks is good. You can make it a little bit more "realistic" using verifying doubles as described here: https://www.relishapp.com/rspec/rspec-mocks/v/3-3/docs/verifying-doubles

danielsdeleo:
:+1: overall once the tests are fixed up.

adamedx:
@danielsdeleo, can you say a little bit about why we don't just use 'provides' DSL to filter out this provider altogether on NanoServer? I think a good reason is because while NanoServer is different than "regular" Windows, it's still Windows. In such a case, testing for features, which is @chefsalim's implementation here, is a more reliable way to gracefully degrade. 

Also, we're already doing some platform testing in the provider (we check for Windows explicitly) anyway, so this is consistent with our existing approach.

Separately, I'd like us to get to the point in the future where we no longer distribute code that will never run on a platform, e.g. providers, which are typically not universal.

danielsdeleo:
For better or worse, right now we (mostly) load all the code we have and then you the provides mapping stuff picks the implementations. I'm assuming from this patch that ruby is throwing an error at code loading time due to missing constants and such, which would imply that this is necessary to get things to work at all. The `supports_msi?` platform helper could be reused in the provider mapping stuff to hide the MSI package on nano as a subsequent improvement.

Long term, I agree it would be advantageous to use a strategy similar to what ChefDK does for it's CLI mapping, see:

* https://github.com/chef/chef-dk/blob/50d8c41e5c0d3921143352a3af3b9dad4fbf29b0/lib/chef-dk/commands_map.rb
* https://github.com/chef/chef-dk/blob/50d8c41e5c0d3921143352a3af3b9dad4fbf29b0/lib/chef-dk/builtin_commands.rb

This would reduce our memory footprint and make Chef commands load a lot faster. That said, this patch lets us make forward progress now.

spec/unit/platform/query_helpers_spec.rb
chefsalim: I'd like to use a verifying double - currently it is more painful than it should be (since the verifying double makes the open method private).  For now, I'm going to follow Claire's pattern of using a non-verifying double, and we can re-factor later to use verifying doubles.

mwrock:
:+1:

PR #3941, opened 2015-09-22
allow reboot by reboot resource with chef-apply

sawanoboly:
The Chef-Apply ignores reboot resource currently.
This PR turns on rebooting by Chef-Apply.


lamont-granquist:
The code seems reasonable so :+1:, even though I think the use case should use chef-zero instead.

smurawski:
:+1:

Issue #3942, opened 2015-09-22
remote_directory produces inconsistent dir permissions


andy-dufour:
Moved from:

https://tickets.opscode.com/browse/CHEF-5080

Dir to be synced:
```
./broken1/aaa
./broken1/dir/file
./broken2/dir/file
./broken2/xxx
```
with:
```
remote_directory "/tmp/remote_something" do
  source "something"
  files_owner "fileowner"
  files_group "fileowner"
  files_mode 0700
  owner "dirowner"
  group "dirowner"
  mode 0750
end
```
Produces:
```
/tmp/remote_something/
drwxr-x--- 6 dirowner root 4.0K Feb 25 21:35 .
drwxrwxrwt 4 root     root 4.0K Feb 25 21:35 ..
drwxr-xr-x 3 root     root 4.0K Feb 25 21:35 broken1
drwxr-x--- 3 dirowner root 4.0K Feb 25 21:35 broken2

/tmp/remote_something/broken1:
drwxr-xr-x 3 root      root 4.0K Feb 25 21:35 .
drwxr-x--- 6 dirowner  root 4.0K Feb 25 21:35 ..
-rwx------ 1 fileowner root    0 Feb 25 21:35 aaa
drwxr-x--- 2 dirowner  root 4.0K Feb 25 21:35 dir

/tmp/remote_something/broken2:
drwxr-x--- 3 dirowner  root 4.0K Feb 25 21:35 .
drwxr-x--- 6 dirowner  root 4.0K Feb 25 21:35 ..
drwxr-x--- 2 dirowner  root 4.0K Feb 25 21:35 dir
-rwx------ 1 fileowner root    0 Feb 25 21:35 xxx
```

Owner of `/tmp/remote_something/broken1` is wrong (not managed), while `/tmp/remote_something/broken2` gets expected owner (managed).

Instead of doing alphabetical ordering on the paths while creating files, which results in undefined result, it should traverse breadth-first.

andy-dufour:
PR open for this: https://github.com/chef/chef/pull/3944

jaymzh:
YAY!!!! :+1:

jkeiser:
Fixed with #3944

PR #3944, opened 2015-09-22
Fix for #3942 - change remote_directory resource file discovery to tr…

andy-dufour:
This is a one line fix to correct some inconsistent file permissions that happen with the remote_directory resource when certain patterns in the file and directory names are met.

There's a repro cookbook for one of the patterns I've assembled here:

https://github.com/andy-dufour/remote-dir-repro

### DIFF
--- a/lib/chef/provider/remote_directory.rb
+++ b/lib/chef/provider/remote_directory.rb
@@ -161,7 +161,7 @@ class Chef
       def files_to_transfer
         cookbook = run_context.cookbook_collection[resource_cookbook]
         files = cookbook.relative_filenames_in_preferred_directory(node, :files, source)
-        files.sort!.reverse!
+        files.sort_by! { |x| x.split(::File::SEPARATOR).count }
       end
 
       # Either the explicit cookbook that the user sets on the resource, or the implicit
--- a/lib/chef/provider/remote_directory.rb
+++ b/lib/chef/provider/remote_directory.rb
@@ -161,7 +161,7 @@ class Chef
       def files_to_transfer
         cookbook = run_context.cookbook_collection[resource_cookbook]
         files = cookbook.relative_filenames_in_preferred_directory(node, :files, source)
-        files.sort_by! { |x| x.split(::File::SEPARATOR).count }
+        files.sort_by! { |x| x.count(::File::SEPARATOR) }
       end
 
       # Either the explicit cookbook that the user sets on the resource, or the implicit
--- a/spec/integration/recipes/remote_directory.rb
+++ b/spec/integration/recipes/remote_directory.rb
@@ -1,0 +1,74 @@
+require 'support/shared/integration/integration_helper'
+
+describe Chef::Resource::RemoteDirectory do
+  include IntegrationSupport
+  include Chef::Mixin::ShellOut
+
+  # Until Cheffish::RSpec has cookbook support, we have to run the whole client
+  let(:chef_dir) { File.join(File.dirname(__FILE__), "..", "..", "..", "bin") }
+
+  # Invoke `chef-client` as `ruby PATH/TO/chef-client`. This ensures the
+  # following constraints are satisfied:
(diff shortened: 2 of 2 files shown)


lib/chef/provider/remote_directory.rb
```diff
@@ -161,7 +161,7 @@ def purge_file(file)
       def files_to_transfer
         cookbook = run_context.cookbook_collection[resource_cookbook]
         files = cookbook.relative_filenames_in_preferred_directory(node, :files, source)
-        files.sort!.reverse!
+        files.sort_by! { |x| x.split(::File::SEPARATOR).count }
```
lamont-granquist: can we just `x.count(::File::SEPARATOR)` instead of splitting it?  that should provide a large speed boost and this is fairly speed critical code if you've got tons of files in the directory.

also, should check on windows that this still works correctly, but i think this is right (IIRC, `::File::SEPARATOR` on Windows is still '/' and we should be passing those back to ourselves -- but if not, we'd need to use `ALT_SEPARATOR` instead...)

lamont-granquist:
paging in @Igorshp as well -- you should check that this code doesn't regress your performance too much (unfortunately, i think you're gonna have to take at least a little hit for correctness).

andy-dufour:
Do we have any idea what performance bench marks we need to hit? I could performance test this pretty easily with several thousand files and compare speed to the old code but a metric to hit would be useful.

jkeiser:
@andy-dufour even a speed *comparison* will do absolute wonders. Let's talk metrics next week.

andy-dufour:
I tested with 10k files of 100 bytes each in a remote_directory and compared old code to new (with x.count instead of split)

Old code
Chef Client finished, 10012/10012 resources updated in 60.816926114 seconds
New code
Chef Client finished, 10013/10013 resources updated in 56.702165621 seconds

(also notice the +1 extra resource is managed in the new code, where it wasn't in the old)

I'll sync up with @jkeiser and see what else he was thinking.

jkeiser:
If the new code is quicker, I don't see any reason this should be blocked (@lamont-granquist and @Igorshp might have different ideas, not sure what the conversation is there?). No reason not to address @lamont-granquist's suggestion about `count()` though!

Thanks :)

andy-dufour:
I've updated the branch using a count instead of splitting and counting. This was what my speed comparisons were based off of.

jkeiser:
Awesome! I'll work with you on getting tests today.

jkeiser:
I added a test to the branch. All looks good! :+1: for my part, but would like @lamont-granquist 's :+1: before we check in.

thommay:
I'm happy with this :+1:

lamont-granquist:
slightly faster by my synthetic benchmark:

```ruby
#!/usr/bin/env ruby

require 'benchmark'
require 'tempfile'
require 'pp'

Benchmark.bm do |x|
  a = `find /usr/local`.split("\n").shuffle
  x.report("sort! reverse!") do
    a.dup.sort!.reverse!
  end
  x.report("count") do
    a.dup.sort_by! { |x| x.count(::File::SEPARATOR) }
  end
end
```

```
% /tmp/foo.rb
       user     system      total        real
sort! reverse!  0.140000   0.000000   0.140000 (  0.134576)
count  0.110000   0.000000   0.110000 (  0.116210)
```

(using the 84,000-ish files in my /usr/local mac homebrew stuff)

:+1:

danielsdeleo:
:+1: from me also

jaymzh:
`sort!.reverse!` is not the same as `sort.count`

jaymzh:
I nevermind, I thought you were proposing that as a solution, but you were comparing the old code to the new code. Ignore me.

lamont-granquist:
Yep, just making sure that by fixing the bug we didn't regress the performance too bad, and it looks like it may actually get a little faster...

jaymzh:
:+1:

PR #3951, opened 2015-09-23
Honor the ordering of whichever `name_attribute` or `default` comes first

### LINKED ISSUE #3950: Chef resources no longer respect the ordering of `default` and `name_attribute`
Consider the following recipe:

```ruby
class Blah < Chef::Resource::LWRPBase
  attribute :x, default: 'default!', name_attribute: true
  attribute :y, name_attribute: true, default: 'default!'
end

b = Blah.new('name!')
puts b.x.inspect
puts b.y.inspect
```

In 12.4.1, this prints:

```
"default!"
"name!"
```

In master, this prints:
```
"default!"
"default!"
```

12.4.1 used to honor whichever attribute happened to come first. 12.5 (master) no longer does. We can either

1. Honor the old ordering behavior and keep backwards compatibility in all cases.
2. Raise an error if you use both, hurting backwards compatibility in all cases where both are specified.
3. Pick `name_attribute` over `default`, sometimes hurting backwards compatibility when both are specified.
4. Pick `default` over `name_attribute`, sometimes hurting backwards compatibility when both are specified.

I may whip up (1) to see if it can be accomodated (it probably can). But damn if it doesn't seem weird :)

cc @coderanger @jaymzh @lamont-granquist

jkeiser:
Fixes #3950 . The issue is that in 12.4.1, if you specify both `default` and `name_attribute` in an attribute, it will pick whichever comes first. In 12.5, it always picks `default`. This restores the old (weird) behavior for backwards compatibility's sake.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -86,6 +86,21 @@ class Chef
     #
     def initialize(**options)
       options.each { |k,v| options[k.to_sym] = v if k.is_a?(String) }
+      # Only pick the first of :default, :name_property and :name_attribute if
+      # more than one is specified.
+      found_default = false
+      options.reject! do |k,v|
+        if [ :name_property, :name_attribute, :default ].include?(k)
+          if found_default
+            true
+          else
+            found_default = true
+            false
+          end
+        else
+          false
+        end
+      end
       options[:name_property] = options.delete(:name_attribute) if options.has_key?(:name_attribute) && !options.has_key?(:name_property)
       @options = options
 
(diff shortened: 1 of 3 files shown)


jaymzh:
:+1:

coderanger:
Would add a warning, but :+1: on the code.

jkeiser:
@coderanger warning added.

coderanger:
:shipit: :fireworks:

jaymzh:
:+1:  again (yay for the warning)

Issue #3948, opened 2015-09-22
Windows2008r2, unable to start chef-client service [v12.4.2] 


dlutsch:
Since the update to chef-client version 12.4.2 we are no longer able to bootstrap our windows server 2008 nodes. The chef-client windows service no longer starts and throws a system call error: "The service did not respond to the start or control request in a timely fashion". We had no issues with this on 12.4.1

```
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

C:\Windows\system32>chef-client
Starting Chef Client, version 12.4.2
[2015-09-22T23:18:22+00:00] INFO: *** Chef 12.4.2 ***
[2015-09-22T23:18:22+00:00] INFO: Chef-client pid: 2708
[2015-09-22T23:18:32+00:00] INFO: Run List is [role[windows-base]]
[2015-09-22T23:18:32+00:00] INFO: Run List expands to [role-windows-base]
[2015-09-22T23:18:32+00:00] INFO: Starting Chef Run for build19.qa.cpf.com
[2015-09-22T23:18:32+00:00] INFO: Running start handlers
[2015-09-22T23:18:32+00:00] INFO: Start handlers complete.
[2015-09-22T23:18:32+00:00] INFO: HTTP Request Returned 404 Object Not Found:
resolving cookbooks for run list: ["role-windows-base"]
[2015-09-22T23:18:32+00:00] INFO: Loading cookbooks [role-windows-base@0.1.0, wi
ndows@1.37.0, chef_handler@1.1.6, webpi@1.1.0, chef-client@4.3.0, cron@1.2.4, lo
grotate@1.3.0]
Synchronizing Cookbooks:
  - role-windows-base
  - windows
  - chef_handler
  - webpi
  - chef-client
  - cron
  - logrotate
Compiling Cookbooks...
[2015-09-22T23:18:32+00:00] WARN: You are overriding windows_package on {:os=>"w
indows"} with Chef::Resource::WindowsCookbookPackage: used to be Chef::Resource:
:WindowsPackage. Use override: true if this is what you intended.
[2015-09-22T23:18:32+00:00] WARN: chef_gem[win32-api] chef_gem compile_time inst
allation is deprecated
[2015-09-22T23:18:32+00:00] WARN: chef_gem[win32-api] Please set `compile_time f
alse` on the resource to use the new behavior.
[...]

PR #3953, opened 2015-09-23
Run the chef service executable from the bin directory

jkeiser:
we are currently running from instead of guessing where it is


jaym:
This will fix https://github.com/chef/chef/issues/3948

PR #3954, opened 2015-09-23
Run the chef service executable from the bin directory

jkeiser:
we are currently running from instead of guessing where it is


ksubrama:
:+1:

PR #3957, opened 2015-09-23
Re-upgrade chef-zero to latest

jkeiser:


tylercloke:
So just for context, do we want chef-zero to be serving up API V1? This is where supporting multiple API versions gets tough...

tylercloke:
I think your ChefFS changes are going to break other things. If you look at `chef/user.rb` and `chef/api_client.rb` you'll notice that there is handling for _both_ versions. I think that is something we'll have to do here too.

jkeiser:
@tylercloke we don't support both versions yet in ChefFS or chef-zero, do we? At the moment I think it's all v0.

jkeiser:
What are you worried will break here? I can download all the data from one of my orgs via `knife`, if that's the concern.

jkeiser:
More specifically, I can successfully do this sequence from my Hosted Chef org:

```
$ knife download /
$ knife diff /
$ knife upload / --force
```

tylercloke:
:+1:

PR #3958, opened 2015-09-23
Add policyfile support to `knife bootstrap`

danielsdeleo:
Adds `--policy-name` and `--policy-group` options to knife bootstrap, which will be passed on to the node via the first-boot JSON file.

Note that this changes the bootstrap behavior for non-policyfile usage when the user does _not_ specify an environment. Before this patch, knife would pass `-E _default` to the node, whereas now it does not give a `-E` option in the default case. The node should still default to an environment of `_default`. Looking in the git history, I didn't find a specific reason for the existing behavior.

### DIFF
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -164,7 +164,12 @@ CONFIG
 
         def first_boot
           (@config[:first_boot_attributes] || {}).tap do |attributes|
-            attributes.merge!(:run_list => @run_list)
+            if @config[:policy_name] && @config[:policy_group]
+              attributes.merge!(:policy_name => @config[:policy_name], :policy_group => @config[:policy_group])
+            else
+              attributes.merge!(:run_list => @run_list)
+            end
+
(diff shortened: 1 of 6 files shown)


danielsdeleo:

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -164,7 +165,12 @@ def latest_current_chef_version_string
 
         def first_boot
           (@config[:first_boot_attributes] || {}).tap do |attributes|
-            attributes.merge!(:run_list => @run_list)
+            if @config[:policy_name] && @config[:policy_group]
```
ranjib: given we are validating presence of both, do we have to check presence of both?

ranjib:
:+1:

jaym:
:+1:

lib/chef/knife/core/bootstrap_context.rb
danielsdeleo: Not strictly necessary, but to me it feels a bit better to check both since we are using both.

PR #3963, opened 2015-09-24
Derive config locations from absolute path to config file

danielsdeleo:
This resolves an issue where running `chef-client -c client.rb -z` will
attempt to create the local mode cache at the filesystem root with an
error like:

```
ERROR: Permission denied @ dir_s_mkdir - /local-mode-cache
```

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -68,7 +68,7 @@ module ChefConfig
 
     default(:config_dir) do
       if config_file
-        PathHelper.dirname(config_file)
+        PathHelper.dirname(PathHelper.canonical_path(config_file, false))
       else
         PathHelper.join(user_home, ".chef", "")
       end
--- a/chef-config/spec/unit/config_spec.rb
+++ b/chef-config/spec/unit/config_spec.rb
@@ -301,14 +301,36 @@ RSpec.describe ChefConfig::Config do
 
         describe "setting the config dir" do
 
+          context "when the config file is given with a relative path" do
+
+            before do
+              ChefConfig::Config.config_file = "client.rb"
+            end
+
+            it "expands the path when determining config_dir" do
+              # config_dir goes through PathHelper.canonical_path, which
+              # downcases on windows because the FS is case insensitive, so we
+              # have to downcase expected and actual to make the tests work.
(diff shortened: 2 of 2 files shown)


danielsdeleo:
For the "windows" tests, I am seeing:

```
       expected: "\\Users\\ddeleo\\oc\\chef\\chef-config\\local-mode-cache"
            got: "\\\\?\\\\users\\ddeleo\\oc\\chef\\chef-config\\local-mode-cache"
```

The default behavior of `PathHelper.canonical` path is to add that weird prefix thing: https://github.com/chef/chef/blob/33509c1bbb7366d3e2a50f61d8d53161e51eb6fa/chef-config/lib/chef-config/path_helper.rb#L118-L119

I don't remember what the prefix thing is for, so I would like to hear from a windows expert about whether I should use it or not.

danielsdeleo:
I asked around and it was decided I should not use the `\\?\\` thing in windows paths. Also, I had to muck around a bit with upper/lower case in the tests because PathHelper.canonical_path downcases on windows. Also had to stub `File.absolute_path` because giving that a windows path on non-windows gets you something like `"/Users/ddeleo/oc/chef/chef-config/C:\\chef\\client.rb"` I confirmed that on windows, `File.absolute_path` works correctly with paths with drive letters and such:

```
irb(main):042:0> File.absolute_path("C:\\chef\\client.rb")
=> "C:/chef/client.rb"
```

@chef/client-windows if y'all want to make sure I'm doing the right things for windows, that'd be nice.

cc @chef/client-core

jaym:
:+1:

stevendanna:
:+1: Was just battling this today.

jkeiser:
The canonical thing is there so that you can use long paths, or unicode paths.

jkeiser:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

You need `\\?\` for paths:
- Longer than 260 characters
- With unicode in them
- With reserved characters in them

The first restriction is the one that trips me up the most, but the second one would be terrible for people in non-European countries, I imagine.

jkeiser:
But that shouldn't block this. We should really make canonical_path append \\?\ only if it's needed.

ksubrama:
:+1:

PR #3965, opened 2015-09-25
Do not prefer default: nil over name_attribute: nil

### LINKED ISSUE #3964: Chef resources with `default: nil` honor the default even if you also specify `name_attribute: true`
```ruby
class Blah < Chef::Resource::LWRPBase
  attribute :x, default: nil, name_attribute: true
end

b = Blah.new('name!')
puts b.x.inspect
```

In 12.4.3, this prints:

```
"name!"
```

In master (12.5), this prints:

```
nil
```

The root cause is that `default: nil` is simply ignored by Chef 12.4.3, so it will treat an explicit `default: nil` the same as if you never specified it at all. Thus, if you specify both `default: nil` and `name_attribute: true` _in that order_, 12.5's behavior will change your default to `nil`, where before it was the name of the attribute.

Changing this is distasteful--if you say `default: nil`, it seems like we should treat it as if you *meant* what you said. My belief in the statement "backcompat is king" is being sorely challenged here.

This is branched from #3951.

@jaymzh @coderanger @lamont-granquist

jkeiser:
Fixes #3964.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -99,7 +99,7 @@ class Chef
         if preferred_default == :default && options[:default].nil?
           preferred_default = found_defaults[1]
         end
-        Chef::Log.deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}--only one (#{preferred_default}) will be obeyed. In Chef 13, specifying multiple defaults will become an error.", caller(5..5)[0])
+        Chef.log_deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}#{options[:resource_class] ? " of resource #{options[:resource_class].resource_name}" : ""}. Only one (#{preferred_default}) will be obeyed. In Chef 13, specifying multiple defaults will become an error.")
         # Only honor the preferred default
         options.reject! { |k,v| found_defaults.include?(k) && k != preferred_default }
       end
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -89,18 +89,19 @@ class Chef
 
       # Only pick the first of :default, :name_property and :name_attribute if
       # more than one is specified.
-      found_defaults = []
-      options.reject! do |k,v|
-        if [ :name_property, :name_attribute, :default ].include?(k)
-          found_defaults << k
-          # Reject all but the first default key you find
-          found_defaults.size > 1
-        else
-          false
-        end
-      end
+      found_defaults = options.keys.select { |k| [ :default, :name_attribute, :name_property ].include?(k) }
       if found_defaults.size > 1
-        Chef::Log.deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}--only the first one (#{found_defaults[0]}) will be obeyed. Please pick one.", caller(5..5)[0])
+        preferred_default = found_defaults[0]
+        # We do *not* prefer `default: nil` even if it's first, because earlier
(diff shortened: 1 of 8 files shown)


coderanger:
I'll just state for the record, I don't think we should support compat on this. It really is a bug IMO and cookbooks should get fixed instead :)

jkeiser:
@coderanger yeah, but it's consistent with our treatment of `nil` in other places (like saying `resource 'x' { myattr nil }`: we warn if you do it, but still support the Chef 12 Way.

coderanger:
Fair, just as long as we remember to clean it up.

lamont-granquist:
i made an attempt at nillable attributes earlier this year and backed it out when application_ruby blew up on me...

coderanger:
@lamont-granquist Sorry ... :-(

jkeiser:
@coderanger @lamont-granquist this should make a lot more sense now (and be a lot shorter), I accidentally had some `run_lock` changes I was making in there.

lib/chef/property.rb
```diff
@@ -89,18 +89,19 @@ def initialize(**options)
...
-        Chef::Log.deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}--only the first one (#{found_defaults[0]}) will be obeyed. Please pick one.", caller(5..5)[0])
+        preferred_default = found_defaults[0]
+        # We do *not* prefer `default: nil` even if it's first, because earlier
+        # versions of Chef (backcompat) treat specifying something as `nil` the
+        # same as not specifying it at all. In Chef 13 we can switch this behavior
+        # back to normal, since only one default will be specifiable.
+        if preferred_default == :default && options[:default].nil?
+          preferred_default = found_defaults[1]
+        end
+        Chef::Log.deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}--only one (#{preferred_default}) will be obeyed. In Chef 13, specifying multiple defaults will become an error.", caller(5..5)[0])
```
coderanger: Should include the resource and cookbook names in addition to the property name.

jkeiser:
@coderanger @jaymzh I fix the deprecations to print the resource name (cookbook name is hard, but will show up in the path), and for deprecations to search for the actual point where the user called into Chef so you can figure out what cookbook and file is the culprit.

lib/chef/property.rb
```diff
@@ -89,18 +89,19 @@ def initialize(**options)
...
-      end
+      found_defaults = options.keys.select { |k| [ :default, :name_attribute, :name_property ].include?(k) }
       if found_defaults.size > 1
-        Chef::Log.deprecation("Cannot specify keys #{found_defaults.join(", ")} together on property #{options[:name]}--only the first one (#{found_defaults[0]}) will be obeyed. Please pick one.", caller(5..5)[0])
+        preferred_default = found_defaults[0]
+        # We do *not* prefer `default: nil` even if it's first, because earlier
+        # versions of Chef (backcompat) treat specifying something as `nil` the
+        # same as not specifying it at all. In Chef 13 we can switch this behavior
+        # back to normal, since only one default will be specifiable.
+        if preferred_default == :default && options[:default].nil?
```
ksubrama: Just a quick sanity check: is it (was it) ever legal to say :name_attribute => nil or something equally ridiculous?  Presumably that'll get checked and failed for not being TrueClass | FalseClass but I'm paranoid.

ksubrama:
:+1:

PR #3966, opened 2015-09-25
Report expanded run list json tree to reporting

kmacgugan:
Added run list expansion json conversion. Creates a json tree from the run list expansion trace.
Added event to report expanded run list

TODO: Unclear how to test the new event.

### DIFF
--- a/lib/chef/json_compat.rb
+++ b/lib/chef/json_compat.rb
@@ -41,6 +41,7 @@ class Chef
     CHEF_RESOURCECOLLECTION = "Chef::ResourceCollection".freeze
     CHEF_RESOURCESET        = "Chef::ResourceCollection::ResourceSet".freeze
     CHEF_RESOURCELIST       = "Chef::ResourceCollection::ResourceList".freeze
+    CHEF_RUNLISTEXPANSION   = "Chef::RunListExpansion".freeze
 
     class <<self
 
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -112,6 +112,7 @@ class Chef
       @exception = nil
       @rest_client = rest_client
       @error_descriptions = {}
+      @expanded_run_list = {}
     end
 
     def run_started(run_status)
@@ -217,6 +218,10 @@ class Chef
       end
     end
 
+    def run_list_expanded(run_list_expansion)
+      @expanded_run_list = run_list_expansion
+    end
+
     def post_reporting_data
       if reporting_enabled?
         run_data = prepare_run_data
@@ -271,6 +276,7 @@ class Chef
       run_data["data"] = {}
       run_data["start_time"] = start_time.to_s
       run_data["end_time"] = end_time.to_s
+      run_data["expanded_run_list"] = Chef::JSONCompat.to_json(@expanded_run_list)
 
       if exception
         exception_data = {}
--- a/lib/chef/run_list/run_list_expansion.rb
+++ b/lib/chef/run_list/run_list_expansion.rb
@@ -22,6 +22,7 @@ require 'chef/mixin/deep_merge'
 
 require 'chef/role'
 require 'chef/rest'
+require 'chef/json_compat'
 
 class Chef
   class RunList
@@ -54,6 +55,13 @@ class Chef
       # * Duplicate roles are not shown.
       attr_reader :run_list_trace
 
+      # Like run list trace but instead of saving the entries as strings it saves their objects
+      # The to_json method uses this list to construct json.
+      attr_reader :better_run_list_trace
+
+      attr_reader :all_missing_roles
+      attr_reader :role_errors
+
       def initialize(environment, run_list_items, source=nil)
         @environment = environment
         @missing_roles_with_including_role = Array.new
@@ -68,6 +76,9 @@ class Chef
 
         @applied_roles = {}
         @run_list_trace = Hash.new {|h, key| h[key] = [] }
+        @better_run_list_trace = Hash.new {|h, key| h[key] = [] }
+        @all_missing_roles = {}
+        @role_errors = {}
       end
 
       # Did we find any errors (expanding roles)?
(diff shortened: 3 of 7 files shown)


lib/chef/run_list/run_list_expansion.rb
```diff
@@ -131,6 +138,11 @@ def errors
         @missing_roles_with_including_role.map {|item| item.first }
       end
 
+      def parse_into_json
+        @seen_items = {:recipe => {}, :role => {}}
+        {:id => @environment, :run_list => convert_run_list_trace('top level')}.to_json
+      end
```
lamont-granquist: This should probably be `def to_json` and we don't want to be using the `Hash#to_json` monkeypatch from json and want to use `Chef::JSONCompat.to_json()` instead.

There's examples of the pattern to follow in most of the Chef API objects, an example that is roughly similar in complexity using a `#to_hash` method inside of `#to_json`:

https://github.com/chef/chef/blob/master/lib/chef/api_client.rb#L109-L131

kmacgugan: Awesome, this is just what I was looking for. I had a feeling I should use JSONCompat but I couldn't find an example of how to do it with a complex hash building method as well. Thanks for the lead.

lib/chef/resource_reporter.rb
```diff
@@ -271,6 +276,12 @@ def prepare_run_data
       run_data["data"] = {}
       run_data["start_time"] = start_time.to_s
       run_data["end_time"] = end_time.to_s
+      if @expanded_run_list
+        run_data["expanded_run_list"] = @expanded_run_list.parse_into_json
+      else 
+        run_data["expanded_run_list"] = {}
+      end
```
lamont-granquist: I think if you initialize `@expanded_run_list = {}` and then just `run_data["expanded_run_list"] = @expanded_run_list` that it'll encode to JSON correctly.

lib/chef/run_list/run_list_expansion.rb
```diff
@@ -22,6 +22,7 @@
 
 require 'chef/role'
 require 'chef/rest'
+require 'json'
```
lamont-granquist: should be `require 'chef/json_compat'`

lib/chef/run_list/run_list_expansion.rb
```diff
@@ -156,8 +169,26 @@ def expand_run_list_items(items, included_by="top level")
         end
       end
 
+      # Recursive helper to decode the non-nested hash form back into a tree
+      def convert_run_list_trace(base)
+        @run_list_trace[base].map do |item|
+          skipped = @seen_items[item.type][item.name]
+          @seen_items[item.type][item.name] = true
```
lamont-granquist: that feels a little bit dirty, since you have persistent state in the object and if you call this twice you'll 'remember' all the state from the previous call.  you probably want to have seen_items be a hash accumulator 2nd argument (default to {}) that you pass through to all the recursive child calls, then it'll be local state, and once you exit from the top level it'll go out of scope.

lib/chef/run_list/run_list_expansion.rb
```diff
@@ -140,8 +152,9 @@ def applied_role(role_name)
       end
 
       def expand_run_list_items(items, included_by="top level")
+
         if entry = items.shift
-          @run_list_trace[included_by.to_s] << entry.to_s
+          @run_list_trace[included_by.to_s] <<  entry
```
lamont-granquist: This line does potentially change a visible API external to this object.  Someone could be calling `#run_list_trace` to get back this hash and be expecting string arguments.

kmacgugan: Ok, I didn't see any code using it in chef-client. Wasn't sure of the policy on this, I could just add a new field to do this with and leave the old one, it just felt crufty.

lib/chef/run_list/run_list_expansion.rb
kmacgugan: Yea, sounds better.

kmacgugan:
@lamont-granquist @danielsdeleo Updated this PR based on lamont's comments, can you review again? Thanks.

lamont-granquist:
LGTM, :+1: 

lib/chef/json_compat.rb
```diff
@@ -138,6 +139,8 @@ def class_for_json_class(json_class)
           Chef::Node
         when CHEF_ROLE
           Chef::Role
+        when CHEF_RUNLISTEXPANSION
+          Chef::RunListExpansion
```
danielsdeleo: You should only need to add this if you are setting `"json_class" => "Chef::RunListExpansion"` in your JSON representation. If you did that, then `Chef::JSONCompat.from_json(string_o_json)` would return a `Chef::RunListExpansion` object instead of a Hash. See for example, nodes: https://github.com/chef/chef/blob/8f65f75b2ae48ed33cfce1853edcdcc1a949b9c3/lib/chef/node.rb#L504

That said, I would recommend against adding the `json_class` thing because while it's sometimes convenient, sometimes it's not and it's difficult to turn it off when you want to. If maintaining compatibility weren't so hard, we'd have ripped out all of this long ago.

lib/chef/resource_reporter.rb
```diff
@@ -271,6 +276,7 @@ def prepare_run_data
       run_data["data"] = {}
       run_data["start_time"] = start_time.to_s
       run_data["end_time"] = end_time.to_s
+      run_data["expanded_run_list"] = Chef::JSONCompat.to_json(@expanded_run_list)
```
danielsdeleo: Kinda weird that this is double-JSON-encoded but it's consistent with the other stuff so it's okay I guess.

lib/chef/json_compat.rb
kmacgugan: Ok, I ripped it out.

danielsdeleo:
:+1:

thommay:
Appveyor failures seem entirely unrelated

danielsdeleo:
Yeah, appveyor stuff is https://github.com/chef/chef/pull/3993

kmacgugan:
Rebased on master to get appveyor fixes in and because this PR is ancient.

danielsdeleo:
Cool, I think mine and Lamont's :+1:'s are still valid so merge when ready

PR #3967, opened 2015-09-25
Make race condition tests explicit

jkeiser:
Fixes issue where tests expected a file to have a PID, but sometimes it would race and be empty

### DIFF
--- a/spec/functional/run_lock_spec.rb
+++ b/spec/functional/run_lock_spec.rb
@@ -35,15 +35,17 @@ describe Chef::RunLock do
 
     # make sure to start with a clean slate.
     before(:each){ FileUtils.rm_r(random_temp_root) if File.exist?(random_temp_root) }
-    after(:each){ FileUtils.rm_r(random_temp_root) }
+    after(:each){ FileUtils.rm_r(random_temp_root) if File.exist?(random_temp_root) }
 
+    WAIT_ON_LOCK_TIME = 1.0
     def wait_on_lock
-      tries = 0
-      until File.exist?(lockfile)
-        raise "Lockfile never created, abandoning test" if tries > 10
-        tries += 1
-        sleep 0.1
+      Timeout::timeout(WAIT_ON_LOCK_TIME) do
+        until File.exist?(lockfile)
+          sleep 0.1
+        end
       end
+    rescue Timeout::Error
+      raise "Lockfile never created, abandoning test"
     end
 
     ##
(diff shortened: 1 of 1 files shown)


spec/functional/run_lock_spec.rb
```diff
@@ -267,20 +311,120 @@ def results
...
+      @name = name
+      @read_from_process, @write_to_tests = IO.pipe
+      @read_from_tests, @write_to_process = IO.pipe
+    end
+
+    attr_reader :example
+    attr_reader :name
+    attr_reader :pid
+
+    def event(event)
```
ksubrama: Could you call the argument here something different from the method name.  It confuses us poor mortals when you do "event == @run_to_event".

spec/functional/run_lock_spec.rb
```diff
@@ -267,20 +311,120 @@ def results
...
+      # Tell the process what to stop at (also means it can go)
+      write_to_process.puts event
+
+      # Run the background block
+      background_block.call if background_block
+
+      # Wait until it gets there
+      Timeout::timeout(CLIENT_PROCESS_TIMEOUT) do
+        until last_event == event
+          @last_event = read_from_process.gets.strip
```
ksubrama: I'm confused about needing to do both blocking and non-blocking calls here.  If you aren't guaranteed to get the event, then this doesn't protect you that much.  You still need to worry about calling read_from_process.gets after last_event has returned because the pipe was closed.  If you are guaranteed to get the event, the last_event is unnecessary.  Just make this last_event a local variable initialized to nil and do blocking calls.

jkeiser: Oh, this is about last_event vs. @last_event? Damn straight, that's just plain a bug. I'll update for @last_event.

spec/functional/run_lock_spec.rb
```diff
@@ -267,20 +311,120 @@ def results
...
     end
 
+    def readline_nonblock(fd)
+      buffer = ""
+      buffer << fd.read_nonblock(1) while buffer[-1] != "\n"
+
+      buffer
+    #rescue IO::EAGAINUnreadable
+    rescue IO::WaitReadable
+      unless buffer == ""
```
ksubrama: This is a blocking call then - just doesn't block on line boundaries.  That's a bit misleading.  Why not just use gets?

jkeiser: This method's point is:

1. If *anything* is in the buffer, read a whole line. Because that's our protocol.
2. If *nothing* is in the buffer, exit and return nil.

#1 shouldn't really happen, but I never want to return anything other than a whole line.

spec/functional/run_lock_spec.rb
jkeiser: @ksubrama fixed.

spec/functional/run_lock_spec.rb
jkeiser: Sure, fire_event it is.

ksubrama:
:+1:

PR #3970, opened 2015-09-25
Create empty config context for chefdk

danielsdeleo:
Allows `knife` and such to parse config files with `chefdk.settting` in
them without error.

I have tested this via manually patching my chefdk, and I am able to set `config_strict_mode true` in chef-dk while allowing the config context to be non-strict when knife evaluates it.


danielsdeleo:

lamont-granquist:
:+1:

ranjib:
:+1:

danielsdeleo:
A change from earlier today broke compat w/ ruby 2.0 and made appveyor mad. Will merge as soon as the builds are green. Thanks for the review!

PR #3972, opened 2015-09-26
Test more of provider resolution by mocking the filesystem and commands

### LINKED ISSUE #3934: Service resource fails if systemd unit files are not present on ubuntu 15.04

jkeiser:
Fixes #3934, based off Ranjib's fix in #3935.

This:
1. Incorporate's Ranjib's fix to make provider resolution work in Ubuntu 12.05, where systemctl returns a nonzero exit code.
2. Modifies the ProviderResolver service tests so that they test provider resolution more deeply, mocking the filesystem and shellout commands to pretend to be various init systems. No actual changes are made to the set of tests, just how the `stub_service_*` methods are implemented.
3. [Adds a test](https://github.com/chef/chef/commit/290c9a0727b5ffa4bf47da7c00e40996c6630a3e) for when the systemctl command returns a nonzero exit code.


jkeiser:
@ranjib @lamont-granquist @dan (cc @chef/client-core) could use a review on the test changes here. (See description.)

lib/chef/platform/service_helpers.rb
```diff
@@ -126,7 +131,7 @@ def extract_systemd_services(command)
           # this splits off the suffix after the last dot to return "sshd"
           services += services.select {|s| s.match(/\.service$/) }.map { |s| s.sub(/(.*)\.service$/, '\1') }
         rescue Mixlib::ShellOut::ShellCommandFailed
```
danielsdeleo: Are there other exit statuses besides `1` that we should maybe not be rescuing (or treating as success, anyway)? Perhaps adding `return: [0,1]` to the `shell_out` call would be safer?

spec/support/shared/unit/double_world.rb
```diff
@@ -0,0 +1,79 @@
...
+#
+# world.files["/etc/rc.d/init.d"] = "blah" means that /etc/rc.d/init.d is a file with contents "blah."
+# world.commands["shellme"] = "result" means that when shell_out! calls "shellme", it will return "result" in stdout.
+# world.commands["shellme"] = 10 means that when shell_out! calls "shellme", it will return exit code 10
+# world.commands["shellme"] = { code: 10, stdout: "result", stderr: "result2"}
+#
+shared_context "double world" do
+  let(:world) { DoubleWorld.new }
+  before do
+    allow(::File).to receive(:exist?)      { |f| world.files.has_key?(f) }
```
danielsdeleo: I'd prefer a stricter approach where you have to say which files you expect the code to try to access that _don't_ exist. Maybe you would do `world.files["/etc/rc.d/init.d"] = false` or something like that. That way you'd have to be explicit about what the state of the filesystem is and you're less likely to accidentally rely on the implicit file-doesn't-exist assumption built in to this.

jkeiser: @danielsdeleo I think it's doing what you expect, if I understand you right: *only* the files you specify are considered to exist. Like, `::File.exists?('/')` will actually fail on all systems with the code the way it is now (and I'm OK with that).

lib/chef/platform/service_helpers.rb
jkeiser: @ranjib what is the exit code we want to catch anyway? Is there a correct one?

lamont-granquist:
After the file provider unit tests, I really dislike mocking out posix filesystems.  There is simply way too many possible calls the code can make to poke the filesystem so from the standpoint of unit testing you have two systems with way too much collaboration between them.  Either you need to refactor your API to make it simpler (which you inherently can't because you have to use files), or you need to expand your definition of the system-under-test to include the filesystem.

If you look at what I did with the debian ifconfig provider, I injected real tempfiles from the specs (via define_const, which you might not like, but i thought it made clear sense to me):

https://github.com/chef/chef/blob/master/spec/unit/provider/ifconfig/debian_spec.rb#L58-L68

Then it doesn't matter what calls or call sequence the code makes, you don't have to mock anything and if the file starts with one contents and winds up with the correct contents, then you know it worked -- doesn't matter what route someone chose to get there.

For this, though, you're just checking existence and if you stub the filenames its kind of halfway pointless.  I'd suggest probably making the filenames depend on a 'chroot' variable that was '/' and then you could dep inject a tmpdir in the specs and a.  validate that the filename was the chroot + the real path, and b. validate that if it exists the method returns the right value.

IMO, we don't need to go down the road of trying to write our own MockFS/FakeFS.  That's looking at the problem wrong.

jkeiser:
@lamont-granquist went ahead and did it as you suggested (roughly).

jkeiser:
Lamont :+1:'d externally, closing.

PR #3973, opened 2015-09-26
better error reporting

h4ck3rm1k3:
Improve upon the error message to include the actual absolute path of the file in question.

### DIFF
--- a/lib/chef/api_client/registration.rb
+++ b/lib/chef/api_client/registration.rb
@@ -68,7 +68,8 @@ class Chef
 
       def assert_destination_writable!
         if (File.exists?(destination) && !File.writable?(destination)) or !File.writable?(File.dirname(destination))
-          raise Chef::Exceptions::CannotWritePrivateKey, "I cannot write your private key to #{destination} - check permissions?"
+          abs_path = File.expand_path(destination)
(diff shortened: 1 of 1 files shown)


thommay:
LGTM :+1: @chef/client-maintainers

lib/chef/api_client/registration.rb
```diff
@@ -68,7 +68,8 @@ def run
 
       def assert_destination_writable!
         if (File.exists?(destination) && !File.writable?(destination)) or !File.writable?(File.dirname(destination))
-          raise Chef::Exceptions::CannotWritePrivateKey, "I cannot write your private key to #{destination} - check permissions?"
+          abs_path = File.expand_path(destination)
+          raise Chef::Exceptions::CannotWritePrivateKey, "I cannot write your private key to #{destination} (#{abs_path})- check permissions?"
```
lamont-granquist: could probably just replace destination with abs_path here, i don't think the user needs to see both?

h4ck3rm1k3: ok, that is fine

h4ck3rm1k3:
improved conciseness

danielsdeleo:
:+1: The appveyor failures are not related to this patch. There were two run lock failures also, I tried re-running to see if they are intermittent. I know we did change those slightly because the test was triggering a race, but perhaps the fix was not complete enough.

PR #3977, opened 2015-09-27
Improve messaging in knife

tas50:
Also switch some URL and references from Community Site -> Supermarket.

### DIFF
--- a/lib/chef/knife/client_bulk_delete.rb
+++ b/lib/chef/knife/client_bulk_delete.rb
@@ -71,7 +71,7 @@ class Chef
           else
             ui.msg("The following validators will be deleted:")
             print_clients(validators)
-            if ui.confirm_without_exit("Are you sure you want to delete these validators")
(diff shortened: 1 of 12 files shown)


lib/chef/knife/client_bulk_delete.rb
```diff
@@ -71,7 +71,7 @@ def check_and_delete_validators(validators)
           else
             ui.msg("The following validators will be deleted:")
             print_clients(validators)
-            if ui.confirm_without_exit("Are you sure you want to delete these validators")
+            if ui.confirm_without_exit("Are you sure you want to delete these validators?")
```
stevendanna: I believe all of these confirm methods already append the "?": https://github.com/chef/chef/blob/master/lib/chef/knife/core/ui.rb#L220

tas50: Yep let me close this and fix that

PR #3979, opened 2015-09-27
Undeprecate RunContext#resource_collection=

coderanger:
I use this in Poise and would like to make sure it continues to be a supported API. This is required because I use subcontexts with different semantics from Chef's isolated ones. The alternative to un-deprecating this would be to allow passing in an alternate factory other than `Chef::ResourceCollection.new()`.

Ping @jkeiser and @chef/client-core for review.

### DIFF
--- a/lib/chef/run_context.rb
+++ b/lib/chef/run_context.rb
@@ -92,7 +92,7 @@ class Chef
     #
     # @see CookbookCompiler
     #
-    attr_reader :resource_collection
+    attr_accessor :resource_collection
 
     #
     # The list of control groups to execute during the audit phase
@@ -532,11 +532,6 @@ ERROR_MESSAGE
       ###
       # These need to be settable so deploy can run a resource_collection
       # independent of any cookbooks via +recipe_eval+
-      def resource_collection=(value)
-        Chef.log_deprecation("Setting run_context.resource_collection will be removed in a future Chef.  Use run_context.create_child to create a new RunContext instead.")
-        @resource_collection = value
-      end
-
       def audits=(value)
         Chef.log_deprecation("Setting run_context.audits will be removed in a future Chef.  Use run_context.create_child to create a new RunContext instead.")
         @audits = value


jkeiser:
We certainly should get something to work here. I'm personally more in favor of construction with a separate collection, simply because there are so many things dependent on the resource collection (notifications among them) that the ability to pull the rug out from under that collection seems like it will hamper attempts at optimization. Does a child run context do the job for you, BTW?

I'm not necessarily opposed to this (poise needs to work!) but if a different approach might make optimization easier by keeping constraints lower, that would be awesome.

coderanger:
https://github.com/poise/poise/blob/master/lib/poise/helpers/subcontext_block.rb#L32-L47 is the relevant code. I'm avoiding using `ChildRunContext` because it won't work in <=12.4 but I do call `initialize_child_state` if available since that doesn't affect overall code flow in the same way as swapping around classes would. If `initialize_child_state` (and `create_child` along with it) took an optional `resource_collection` kwarg that would work for my needs I think.

jkeiser:
@coderanger I'm curious what would happen if you replaced the bit about changing the resource collection module with `class<<child_run_context.resource_collection; prepend SubcontextResourceCollection; end`? Would that make it more likely to work with all comers? Might even let you use `create_child`, not that that's strictly necessary.

coderanger:
I would have to rearchitect a bunch of stuff, which I do not have the cycles to do at this point. This PR is the minimum to unblock the 12.5 release and at the moment that is my priority.

jkeiser:
Understood, and I don't want to dump work on you: here is a PR. https://github.com/poise/poise/pull/21

In the meantime, I think I'd be OK with marking this @api private in 12.5 by simply adding `# @api private` to the setter ( @lamont-granquist what do you think?). I think we've gathered enough data since 12.4 to be reasonably confident that Poise is the main thing depending on this, if we want to remove it in Chef 13.

lamont-granquist:
I don't really understand the rationale for removing it.  I'll be in the office later today, so you should grab me and chat about this after lunch.

lamont-granquist:
I think I'm good either way you two decide :+1:

jkeiser:
I'll make a quick PR to remove the warning and add # @api private

coderanger:
So to record some chatter from IRC I think the immediate plan is to make `resource_collection=` not deprecated but mark it as `@api private`. This means we could still remove it in Chef 13, but Poise users won't get deprecation spam in the mean time. Long term there are a few more options: a strong contender is to not have a the setters, but still allow passing in a collection when you create a run context (either a normal one or a child context). Alternatively we could try to merge the Poise nested context stuffs back to core and then drop it on the Poise side (eventually). Another option would be to encourage a pattern of doing instance mixins combined with the core child context support to extend the existing objects without needing to provide different classes.

jkeiser:
I just submitted https://github.com/chef/chef/pull/3989 for @api private

coderanger:
Also for the record (aka people that find this via Google in a year) we collectively reserve the right to move it back to being deprecated when there is a more concrete proposal for it not being needed anymore.

PR #3984, opened 2015-09-28
Improve rubygems provider for gems without version

### LINKED ISSUE #3983: Rubygems provider download metadata from rubygems even if unnecessary.
rubygem provider always downloads metadata from rubygems when version attribute is nil.

This fetch is only useful when action is `:upgrade` or if there is no version installed.

When action is `:upgrade`, we should avoid querying rubygems.

kamaradclimber:
When a resource has no target version and a version is already
installed, we should only download metadata from rubygems when action is
:upgrade.

Fixes #3983

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -494,7 +494,7 @@ class Chef
 
         def target_version_already_installed?(current_version, new_version)
           return false unless current_version
-          return false if new_version.nil?
+          return false if new_version.nil? && action == :upgrade
 
           Gem::Requirement.new(new_version).satisfied_by?(Gem::Version.new(current_version))
         end
--- a/spec/unit/provider/package/rubygems_spec.rb
+++ b/spec/unit/provider/package/rubygems_spec.rb
@@ -345,10 +345,16 @@ describe Chef::Provider::Package::Rubygems do
   describe "when new_resource version is nil" do
     let(:target_version) { nil }
 
-    it "target_version_already_installed? should return false so that we can search for candidates" do
+    it "target_version_already_installed? should return false so that we can search for candidates if action is :upgrade" do
+      @provider.action = :upgrade
       @provider.load_current_resource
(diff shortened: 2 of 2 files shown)


kamaradclimber:
cc @aboten

lamont-granquist:
This doesn't quite smell right to me.

What happens when we are doing :upgrade with new_resource.version nil, but we've already been upgraded to the latest version.  Don't we wind up not only downloading the list from rubygems, but then also forcing the upgrade to the version we already have installed?

kamaradclimber:
I am not sure to understand what you mean.

To me, the expected behavior, when new_resource.version is nil, are:

current_resource.version |new_resource.version | action | behavior
-----------------------------------|-------------|----------|-------------
nil or 1.0 |nil           | :upgrade| download metadata, install the latest gem
nil           | nil          | :install | download metadata, install the latest gem
1.0          | nil          | :install | do not download anything

This latest case was not correct before the patch (metadata were downloaded for :upgrade and :install action)

lamont-granquist:
When the action is :upgrade, and new_resource.version is nil, and current_resource.version is 1.0 and when you download from rubygems and the latest version there is 1.0, then you should not 'upgrade' from 1.0 to 1.0.

With your code what happens?

kamaradclimber:
With the patch activated.

```
gem_package 'mosespa' do
  action :upgrade
end
```

First run normally install the gem:
```
- upgrade package mosespa from uninstalled to 1.4.4
```

Second run, does nothing:
```
gem_package[mosespa] action upgrade (up to date)
```

lamont-granquist:
It still feels like this should go into the parent package provider, and this is sort of how all providers should work, and the patch 'feels' kind of side-effect-ey.   But part of the problem is that the package providers are a mazy of twisty, badly documented APIs between the parent and the subclasses so its not clear what the right way to implement this should be offhand.

thommay:
OK, for right now I think we're going to choose not to take this PR, in favour of doing the work described in #3996 to make sure this works across the board.

Issue #3671, opened 2015-07-15
rpm_package silently fails to install packages with a `+` (plus) character in the version


irvingpop:
Very similar to the issue described in #3496 and fixed in #3503 

This line of code: https://github.com/chef/chef/blob/master/lib/chef/provider/package/rpm.rb#L64
is proving brittle because I've found yet another character that it misses:
http://rubular.com/r/kdkr3EYxOY

Using Chef 12.4.1:

```
chef:recipe (12.4.1)> run_chef
[2015-07-15T17:38:39+00:00] INFO: Processing rpm_package[/tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm] action install ((irb#1) line 1)
[2015-07-15T17:38:39+00:00] DEBUG: Providers for generic rpm_package resource enabled on node include: [Chef::Provider::Package::Rpm]
[2015-07-15T17:38:39+00:00] DEBUG: Provider for action install on resource rpm_package[/tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm] is Chef::Provider::Package::Rpm
[2015-07-15T17:38:39+00:00] DEBUG: No package source specified, but /tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm exists on the filesystem, copying to package source
[2015-07-15T17:38:39+00:00] DEBUG: rpm_package[/tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm] checking rpm status
chef-server-core 12.2.0+20150713220422-1.el6
[2015-07-15T17:38:39+00:00] DEBUG: rpm_package[/tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm] checking install state
package /tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm is not installed
[2015-07-15T17:38:39+00:00] DEBUG: rpm_package[/tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm] /tmp/ecm_cache/chef-server-core-12.2.0+20150713220422-1.el6.x86_64.rpm not installed, installing 
[...]

PR #3985, opened 2015-09-28
Use much simpler regex for determining the rpm version

irvingpop:
See #3671 for discussion

cc: @lamont-granquist


thommay:
this looks much saner! :+1: cc @chef/client-enterprise-linux

lamont-granquist:
lol, yeah, we define the query string and its space delimited.

LGTM

:+1:

PR #3986, opened 2015-09-28
fix errors when files go away during chown

aub:
The deploy resource attempts to change the owner of all files in the
deploy directory, but it breaks when files are removed during the
process. This is because FileUtils::chown_R gets a list of the files and
then walks through and updates them, and if that takes any time at  all
then files may be deleted during the process. When this happens, the
deploy will fail with Errno::ENOENT. chown_R takes an options hash with
:force as one of the available options, and the only thing it does is
stop errors from being raised under these circumstances.

NOTE: when I originally pushed this I hadn't updated the test, so the build failed, but it should be passing now.

### DIFF
--- a/lib/chef/provider/deploy.rb
+++ b/lib/chef/provider/deploy.rb
@@ -276,7 +276,7 @@ class Chef
 
       def enforce_ownership
         converge_by("force ownership of #{@new_resource.deploy_to} to #{@new_resource.group}:#{@new_resource.user}") do
-          FileUtils.chown_R(@new_resource.user, @new_resource.group, @new_resource.deploy_to)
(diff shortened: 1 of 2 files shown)


ranjib:
@aub what is causing the files to be gone while chown_R is being invoked. wont it be better to address that?

aub:
It's pretty standard to store pid files in {deploy_dir}/tmp/pids. I'm doing
that with pid files for our workers, which tend to be pretty transient, and
as a result this error pops up when a worker dies and is replaced during
the deploy.

ranjib:
:+1:

thommay:
seems fair :+1:

lamont-granquist:
Yeah, I do agree with fixing it upstream, but the change looks fine :+1:

PR #3990, opened 2015-09-29
Use SHA256 instead of MD5 for registry_key when data is not displayable

jaym:
Getting rid of md5 where we dont need it. See #3945 for more info


lamont-granquist:
I still think we should use hex instead of hashing it, but :+1:

thommay:
:+1:

PR #3991, opened 2015-09-29
Modify remote_file cache_control_data to use sha256 for its name 

jaym:
We want to support a fips mode, and doing MD5 with fips mode enabled
is wrong/hard. In this case, the type of checksum does not
matter, so let's just use sha256 since fips mode will be happy
with that.

For cases where the cache control data exists, we update it
to provide a seamless upgrade.

### DIFF
--- a/lib/chef/digester.rb
+++ b/lib/chef/digester.rb
@@ -38,7 +38,11 @@ class Chef
     end
 
     def generate_checksum(file)
-      checksum_file(file, OpenSSL::Digest::SHA256.new)
+      if file.is_a?(StringIO)
+        checksum_io(file, OpenSSL::Digest::SHA256.new)
+      else
+        checksum_file(file, OpenSSL::Digest::SHA256.new)
+      end
     end
 
     def self.generate_md5_checksum_for_file(*args)
--- a/lib/chef/provider/remote_file/cache_control_data.rb
+++ b/lib/chef/provider/remote_file/cache_control_data.rb
@@ -145,18 +145,51 @@ class Chef
         end
 
         def load_json_data
-          Chef::FileCache.load("remote_file/#{sanitized_cache_file_basename}")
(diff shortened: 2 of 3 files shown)


lib/chef/provider/remote_file/cache_control_data.rb
```diff
@@ -145,18 +145,51 @@ def load_data
         end
 
         def load_json_data
-          Chef::FileCache.load("remote_file/#{sanitized_cache_file_basename}")
+          path = sanitized_cache_file_path(sanitized_cache_file_basename)
```
coderanger: Maybe instead of doing this as a one off, we should put some kind of version number in the metadata this time so we  can do this more cleanly next time?

jaym: In general, that sounds like a good idea. I don't think it would have helped this case as the only the file name is incorrect, the data inside is exactly the same. Will look into adding this, may make the solution cleaner

lib/chef/digester.rb
```diff
@@ -38,7 +38,11 @@ def checksum_for_file(file)
     end
 
     def generate_checksum(file)
-      checksum_file(file, OpenSSL::Digest::SHA256.new)
+      if file.is_a?(StringIO)
```
danielsdeleo: Would it make sense to check for `respond_to?(:read)` ? I don't know why you'd need it, but it would allow you to `File.open(thing, opts) { |f| generate_checksum(f) }` (maybe that would be helpful if dealing with a gzipped file?).

danielsdeleo:
:+1: there are some improvements that can be made, but I won't block merge on those if this work is time sensitive.

thommay:
@jaym is this waiting on anything?

lamont-granquist:
:+1: 

if you want to address Dan's `respond_to?(:read)` comment that'd be cool as well, that's an idiom that I've seen before.

PR #3993, opened 2015-09-29
Fix provider_resolver tests on Windows

jkeiser:


danielsdeleo:
:+1: if appveyor likes it

jaym:
Why do we need the `if windows?`? Could we just mock it out for everybody?

lamont-granquist:
lol, :+1:

PR #3995, opened 2015-09-29
fix omnitruck url

thommay:


thommay:
cc @lamont-granquist @tas50

PR #3997, opened 2015-09-29
Add event logging and more race condition tests

jkeiser:


danielsdeleo:
Also there's some `puts` in the run lock test that might not be part of this change, but definitely should be replaced with `print "#{text}\n"`.

danielsdeleo:
Nice. :+1:

Issue #3992, opened 2015-09-29
overriding 'status_command' in services is broken:


jaymzh:
You can no longer override 'status_command' for services, it'll break with:

```
service[test]: unable to locate the init.d script!
```

Sample code:

```
service 'test' do
  status_command '/bin/true status'
  start_command '/bin/true start'
  stop_command '/bin/true stop'
  restart_command '/bin/true restart'
end

file '/tmp/wakakaka' do
  notifies :restart, 'service[test]'
end
```

This will fail with the above error. This worked in 12.0.3 and every release before but seems to be broken as of at least 12.4.1, maybe before.

andy-dufour:
I can't reproduce using the above sample - i successfully converge on ubuntu 14 and centos 7. What platform did this error on for you?

jaymzh:
CentOS 6. I can repro on 12.4.1 and 12.5.0-daily from a few days ago.

Almost certainly this is limited to Redhat-like OSes, and probably < 7 since C7 uses Systemd.

andy-dufour:
Thanks, I can reproduce on CentOS 6.

jaymzh:
CC @lamont-granquist  I think he was changing the way this code inherited recently...

lamont-granquist:
I can't see where this changed, certainly it was an unintended side-effect of something.  The patch looks obviously correct.

thommay:
Fixed in master, hence closing!

PR #4000, opened 2015-09-29
Fix for #3992: Add check for custom command in redhat service provider

andy-dufour:
Fixing https://github.com/chef/chef/issues/3992 

Instead of just checking for the missing service, check if a custom command was provided for the action as well. If neither is provided throw Chef::Exceptions::Service pointing at a missing init.d script or custom command for action.

### DIFF
--- a/lib/chef/provider/service/redhat.rb
+++ b/lib/chef/provider/service/redhat.rb
@@ -64,7 +64,7 @@ class Chef
             a.assertion do
               custom_command_for_action?(action) || !@service_missing
             end
-            a.failure_message Chef::Exceptions::Service, "#{new_resource}: unable to locate the init.d script!"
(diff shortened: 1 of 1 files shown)


jaymzh:
Seems reasonable. :+1:

lib/chef/provider/service/redhat.rb
```diff
@@ -61,8 +61,10 @@ def define_resource_requirements
           end
```
jaymzh: wtf is dbleoes?

jaymzh:
fwiw, I tested this locally and it works for our case, as I expected it would.

adamedx:
@lamont-granquist, @jkeiser, can you take a look?

lamont-granquist:
LGTM :+1: 

@chef/client-core could use review.

thommay:
LGTM :+1:

PR #4002, opened 2015-09-29
Make sure name_attribute works on derived properties

jkeiser:
Currently this will fail:

```
class X < Chef::Resource
  property :x, name_property: true
end

class Y < X
  property :x, name_attribute: true
end
```

Because when `name_property` is inherited to `name_attribute`, the properties conflict.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -422,7 +422,13 @@ class Chef
     # @return [Property] The new property type.
     #
     def derive(**modified_options)
-      Property.new(**options.merge(**modified_options))
+      # Since name_property and name_attribute are aliases, if you specify either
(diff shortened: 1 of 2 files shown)


coderanger:
I'm confused, why are you only allowed to have one name attribute/property?

coderanger:
Or put another way, I wouldn't expect those two properties to interact at all.

jkeiser:
@coderanger my example (inline) sucked. Fixed.

coderanger:
There we go, :+1:

coderanger:
We should also remember to deal with this for:

```ruby
class X < Chef::Resource
  property :x, name_property: true
end

class Y < X
  property :x, default: 'something'
end
```

PR #4003, opened 2015-09-29
Ensure that our list of recipes is backwards compat

### LINKED ISSUE #3767: chef-client breaking change to recipe list in node object
There appears to be a breaking change between `12.1.1` and `12.4.1` in how the recipe list is posted back to the chef server.

In `12.1.1`, the recipes posted back are both expanded and unexpanded. In other words, if you have the recipe `int-service::default` in your run list, the recipe list will have `int-service` and `int-service::default`.

In `12.4.1`, the unexpanded version (`int-service`) is no longer posted back. As such, if you had a search like: `recipes:int-service` the node will no longer be returned in the search results.

This looks to be a breaking change in `chef-client`. Should this change be made for a minor release version (i.e., `12.1` to `12.4`) or should it be a change introduced in `13.x`?

thommay:
Prior to chef 12.2 we included unexpanded 'cookbook' names for default
recipes. In 12.2, we moved to expanded ('cookbook::default') names,
which broke some searches.
However, some of our users have now moved to searching for expanded,
so we need to cater for both.
Fixes #3767

cc @lamont-granquist @chef/client-core

### DIFF
--- a/lib/chef/run_list/versioned_recipe_list.rb
+++ b/lib/chef/run_list/versioned_recipe_list.rb
@@ -82,6 +82,21 @@ class Chef
           qualified_recipe
         end
       end
+
+      # Get an array of strings of both fully-qualified and unexpanded recipe names
+      # in response to chef/chef#3767
+      # Chef-13 will revert to the behaviour of just including the fully-qualified name
+      #
+      # @return [Array] Array of strings with fully-qualified and unexpanded recipe names
+      def with_duplicate_names
+        self.map do |recipe_name|
+          if recipe_name.include?('::')
+            recipe_name
+          else
+            [recipe_name, "#{recipe_name}::default"]
+          end
+        end.flatten
+      end
     end
   end
 end
--- a/spec/unit/run_list/versioned_recipe_list_spec.rb
+++ b/spec/unit/run_list/versioned_recipe_list_spec.rb
@@ -187,4 +187,9 @@ describe Chef::RunList::VersionedRecipeList do
     end
   end
 
+  context "with duplicated names", :chef_gte_13_only do
+    it "should fail in Chef 13" do
+      expect(list).to_not respond_to(:with_duplicate_names)
+    end
+  end
 end
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -440,7 +440,7 @@ class Chef
 
       self.tags # make sure they're defined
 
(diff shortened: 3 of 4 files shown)


lib/chef/run_list/versioned_recipe_list.rb
```diff
@@ -82,6 +82,21 @@ def with_fully_qualified_names_and_version_constraints
           qualified_recipe
         end
       end
+
+      # Get an array of strings of both fully-qualified and unexpanded recipe names
+      # in response to chef/chef#3767
+      # Chef-13 will revert to the behaviour of just including the fully-qualified name
+      #
+      # @return [Array] Array of strings with fully-qualified and unexpanded recipe names
+      def with_duplicate_names
```
danielsdeleo: I'd expect this to be flattened instead of having to flatten later.

spec/unit/run_list/versioned_recipe_list_spec.rb
```diff
@@ -187,4 +187,9 @@
     end
   end
 
+  context "with duplicated names", :chef_gte_13_only do
+    it "should fail in Chef 13" do
+      expect(list.with_duplicate_names).to raise_error
```
danielsdeleo: this needs to be `expect {}` with a code block.

ranjib:
:+1:

spec/unit/run_list/versioned_recipe_list_spec.rb
danielsdeleo: Also, if we're expecting `NoMethodError`, it's best to specify that explicitly, or you could change it to `expect(thing).to_not respond_to(:method)`

danielsdeleo:
I like the behavior change/fix, have a few nitpicks.

spec/unit/run_list/versioned_recipe_list_spec.rb
thommay: Ah, I like the latter much more. Thanks

thommay:
@danielsdeleo responded to both points, thanks

danielsdeleo:
Awesome. :+1:

danielsdeleo:
Added to the 12.5 milestone, since we should ship this in the next release.

lamont-granquist:
LGTM :+1:

theckman:
:+1: thank you.

PR #4007, opened 2015-09-30
Add external tests for chefspec, chef-sugar, chef-rewind, foodcritic, halite and poise

jkeiser:


Gemfile
```diff
@@ -25,6 +25,14 @@ group(:development, :test) do
   gem 'cheffish', "~> 1.3", "!= 1.3.1"
 
   gem 'ruby-shadow', :platforms => :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
+
+  # For external tests
+  gem 'chefspec', github: 'jkeiser/chefspec', branch: 'jk/chefspec-12.5'
+  gem 'chef-sugar'
+  gem 'poise', github: 'poise/poise', branch: 'deeecb890a6a0bc2037dfb09ce0fd0a8931519aa'
```
coderanger: Guessing there were issues here?

jkeiser: @coderanger nope, just pinning to a Last Known Good.

jkeiser: You can pretty easily run what it runs with `bundle exec rake poise_spec` and `bundle exec rake halite_spec`, BTW. The halite tests didn't pass, not sure if it's master or the way that I'm munging the Gemfile.

coderanger: Halite tests against Chef master nightly (eg. https://travis-ci.org/poise/halite/jobs/82850534) and on every commit, so I think it should be working. Can you send me the errors? It is sensitive to odd bundler setups because of the way it hooks in to Rubygems so that might be related.

jkeiser: @coderanger https://gist.github.com/jkeiser/437679a61edb174e8473

coderanger: Hmm, those gems get added via [the Halite Gemfile](https://github.com/poise/halite/blob/master/Gemfile#L31-L34), but I don't see anything in your Gemfile munger that should be disrupting that.

jkeiser: @coderanger the munger creates a new Gemfile in a temp location, so ../specs will not do what one expects.

I *could* create a new Gemfile in the gem directory itself, though. Probably would be fine.

coderanger: Ahh, hmm. Maybe replace `__FILE__` in the munger with a path equivalent to where the gemfile would be normally?

jkeiser:
@coderanger got halite in there, generating a gemfile in the right location worked. If you wouldn't mind reviewing, I'll get this sucker in :)

coderanger:
Looks good! You shouldn't actually *need* the `dev_gem` munger but it makes the code a little safer so :+1: :shipit:

PR #4013, opened 2015-09-30
Example basic acceptance tests work with chef-acceptance

patrick-wright:
PR for peer review.  The PR represents some experimental work for running Chef acceptance tests.


acceptance/Gemfile
```diff
@@ -0,0 +1,9 @@
+source 'https://rubygems.org'
+
+gem 'chef-acceptance', github: 'chef/chef-acceptance'
+gem 'berkshelf'
```
sersut: Do we need to add all of these stuff into the Gemfile? I think we are launching `kitchen` via CLI so these doesn't necessarily need to be gem dependencies. We can assume that the runner of `chef-acceptance` has installed Chef DK.

sersut:
This is all good @patrick-wright. But I do not think we want to move the kitchen-tests into acceptance. I think we want to copy them. We can leave it to the `chef` project to decide if those tests should continue to run in travis or not. Let's restore the `kitchen-tests` directory as it is and copy the files we need into chef-acceptance to provide an example to write a test with cookbooks.

acceptance/Gemfile
patrick-wright: My assumption is that chefdk will be installed.  Which I think is a good thing to enforce for acceptance.  We can always add them here if we run into issues.  I'll add that to the README.

patrick-wright:
@sersut I restored `kitchen-tests`.

acceptance/Gemfile
sersut: :+1:

sersut:
:+1: for me. 

@chef/client-core This is the very first example of the acceptance tests that we are going to hook into chef-acceptance. 

Do you see any side effects of introducing this directory?

lamont-granquist:
This seems to duplicate the purpose of kitchen-tests so we should merge them into one directory and run them both from travis and delivery.  Having two of these just means that one or both of them will be kinda shitty depending on which one gets attention.

acceptance/.kitchen.yml
```diff
@@ -0,0 +1,35 @@
+---
+driver:
+  name: vagrant
+  customize:
+    cpus: 4
+    memory: 2048
+
+provisioner:
+  name: chef_zero
```
lamont-granquist: Don't you need to use chef_github here and kitchen-appbundle-updater gem in order to run the PR against the new chef code?  If we're assuming that the build has to go through all the way and generate an omnibus artifact, then I'd still expect to see configuration here that would pull that build out of artifactory (hooks in test-kitchen that I do not think exist yet).

acceptance/Gemfile
lamont-granquist: Per other comment would like to see all this duplicated code shared with the travis tests, so we should not make any assumptions about ChefDK being installed.

sersut:
@lamont-granquist our aim here is to write an example test that will be executed under the context of `chef-acceptance` framework. In the future this framework will be hooked into our CI so that we can run the acceptance tests we are running locally in the CI. 

These tests will not be testing against a PR commit. They will be testing against full omnibus artifacts. Hence we are not targeting the github driver. They will be able to pull the version to be tested from any channel (:stable, :current or :unstable when it is ready). 

I think leaving `kitchen-tests` is the right way to go until we start executing this suite in CI. 

Ping me and we can have a chat if there are still concerns :smile:

lamont-granquist:
There still seems to be considerable overlap in the cookbooks then.  Can we just have two different kitchen files like the .kitchen.yml and kitchen.cloud.yml pattern in the community cookbooks?

chef-supermarket:

sersut:
@lamont-granquist overlap is OK for now. I think the contents of this directory will change in the future. As we evolve chef-acceptance in the future, `.kitchen.yml` may even go away and everything might be driven by `.acceptance.yml`. We will be developing chef-acceptance iteratively and the tests we have copied over there are just examples for now.

thommay:
OK, what I'd like to see here is an approach that generates a set of cookbooks that we can use for acceptance tests across the board - both for travis to run against PRs and for delivery/manhattan to run at the test phase of omnibus pipelines.

patrick-wright:
@lamont-granquist @sersut I've removed the copied kitchen-tests from acceptance and left in the basic specs.  Our primary goal is to validate the acceptance framework which we can do without the additional tests.  @thommay We're less concerned with the actual tests and more concerned with validating the acceptance structure.  The idea is to create a few examples of how to setup the acceptance tests.  Hope that clears it up a bit.

acceptance/Gemfile
danielsdeleo: Just curious, could this be done with policyfiles instead? :dog: :cake:

sersut:
I think that's OK for now @patrick-wright. @lamont-granquist / @thommay let us know if this looks good. 

To reiterate what we are trying to do is to prove the infrastructure to be able to write a set of acceptance tests that will work with our internal CI and from developer laptops. Our intention is not to determine and write those tests.

sersut:
Had a conversation with @patrick-wright and @lamont-granquist and we will not merge this right now. We will revisit once the rest of the chef-acceptance is in place.

PR #4015, opened 2015-10-01
Un-remove ExpandNodeObject#load_node, deprecate it

danielsdeleo:
Some third-party tools are using this, we need it to work.

In particular, we are seeing these in ChefDK's verification tests for ChefSpec:

```
ERROR: Chef failed to converge: undefined method `load_node' for #<Chef::PolicyBuilder::ExpandNodeObject:0x007fb49bbf8c48>
```

cc @chef/client-core

### DIFF
--- a/lib/chef/policy_builder/expand_node_object.rb
+++ b/lib/chef/policy_builder/expand_node_object.rb
@@ -33,6 +33,9 @@ class Chef
     # expands the run_list on a node object and then queries the chef-server
     # to find the correct set of cookbooks, given version constraints of the
     # node's environment.
+    #
+    # Note that this class should only be used via PolicyBuilder::Dynamic and
+    # not instantiated directly.
     class ExpandNodeObject
 
       attr_reader :events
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Yeah, I've seen that code before in ChefSpec.

:+1:

lamont-granquist:
@thommay this is probably a 12.5.0 blocker.

lamont-granquist:
And tagging @jkeiser since he seems to be doing the 12.5.0 release

danielsdeleo:
Yeah, John had said that he tested ChefSpec and it worked, so I'm not sure what's up, but the ChefDK build is definitely angry about it.

lamont-granquist:
That seems like we might not be testing against it correctly somehow?

Maybe @jkeiser is testing master and @sethvargo ninja patched ChefSpec already but it just hasn't been released?

Either way, we should click merge on this.

thommay:
LGTM :+1:

PR #4029, opened 2015-10-05
Fix search result pagination

### LINKED ISSUE #4027: Chef::Search::Query#search might return duplicate results or loop forever
PR #3135 introduced a bug where search() might list results more than once or never complete.

This issue is here:

https://github.com/chef/chef/blob/master/lib/chef/search/query.rb#L92

Incrementing by the count of the rows you got back seems correct, but the problem is that the start and rows parameters are passed directly to solr on the backend.  Results from solr, however, might included deleted nodes that are filtered from the results.  Thus, you might ask for the first page of 10 results, get back 8 nodes because 2 of the 10 results returned by solr were filtered.  However, you should still start your next page at 10 because the "pages" that matter are the results coming back from solr.

NB: The author of the PR original change would have been fine. Unfortunately, I gave a braindead recommendation that led to this bug.

Fix incoming.

stevendanna:
The start and rows parameter that are passed as part of the search
request are passed directly to Solr on the backend. Results from Solr
may contain deleted nodes no longer in the erchef database.  Erchef will
filter such nodes from the results. Thus, a user may receive fewer rows
than they asked for. Incrementing 'start' only by the number of rows
received will then result in the next Solr response overlapping the
first, which can lead to duplicate results. In the case of a Solr
results page that was completely filtered, it would lead to an infinite
loop.

This commit changes the code to always increment by the requested page
size (args[:rows]) when it is non-nil. Incrementing by the length of the
response set is still wrong in the case when the args[:rows] is nil, but
the server doesn't give us anything else to increment by.

Fixes #4027

### DIFF
--- a/lib/chef/search/query.rb
+++ b/lib/chef/search/query.rb
@@ -88,8 +88,21 @@ WARNDEP
 
         if block
           response["rows"].each { |row| block.call(row) if row }
-          unless (response["start"] + response["rows"].length) >= response["total"]
-            args_h[:start] = response["start"] + response["rows"].length
+          #
+          # args_h[:rows] and args_h[:start] are the page size and
+          # start position requested of the search index backing the
+          # search API.
+          #
+          # The response may contain fewer rows than arg_h[:rows] if
+          # the page of index results included deleted nodes which
+          # have been filtered from the returned data. In this case,
+          # we still want to start the next page at start +
+          # args_h[:rows] to avoid asking the search backend for
+          # overlapping pages (which could result in duplicates).
+          #
+          next_start = response["start"] + (args_h[:rows] || response["rows"].length)
+          unless next_start >= response["total"]
+            args_h[:start] = next_start
             search(type, query, args_h, &block)
           end
           true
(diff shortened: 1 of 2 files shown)


stevendanna:
cc @chef/client-core 

This is still problematic in the case where the user doesn't pass a rows attribute.  Perhaps we want to deprecate searches without rows specified?

danielsdeleo:
:+1:

lib/chef/search/query.rb
```diff
@@ -88,8 +88,9 @@ def search(type, query='*:*', *args, &block)
 
         if block
           response["rows"].each { |row| block.call(row) if row }
-          unless (response["start"] + response["rows"].length) >= response["total"]
-            args_h[:start] = response["start"] + response["rows"].length
+          next_start = response["start"] + (args_h[:rows] || response["rows"].length)
+          unless next_start >= response["total"]
```
lamont-granquist: might want to have a comment here to explain these fields correctly.

lamont-granquist:
:+1:

lib/chef/search/query.rb
danielsdeleo: good point, this is a good place to make a note that the results may not be the same size as the amount of rows we asked for.

stevendanna:
Added a lengthy comment.

jaym:
excellent comment :+1:

danielsdeleo:
Merge whenever.

PR #4033, opened 2015-10-05
Fix dispatch when there are different receivers

jkeiser:
with different numbers of arguments.
Fixes https://github.com/chef/chef-dk/issues/546

### DIFF
--- a/lib/chef/event_dispatch/dispatcher.rb
+++ b/lib/chef/event_dispatch/dispatcher.rb
@@ -32,8 +32,11 @@ class Chef
           mth = s.method(method_name)
           # Trim arguments to match what the subscriber expects to allow
           # adding new arguments without breaking compat.
-          args = args.take(mth.arity) if mth.arity < args.size && mth.arity >= 0
-          mth.call(*args)
+          if mth.arity < args.size && mth.arity >= 0
+            mth.call(*args.take(mth.arity))
+          else
+            mth.call(*args)
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


danielsdeleo:
Wow, nice catch. :+1:

danielsdeleo:
@chef/client-core This is a blocker for 12.5 final

coderanger:
I'm a little confused what this actually changed?

lib/chef/event_dispatch/dispatcher.rb
```diff
@@ -32,8 +32,11 @@ def call_subscribers(method_name, *args)
           mth = s.method(method_name)
           # Trim arguments to match what the subscriber expects to allow
           # adding new arguments without breaking compat.
-          args = args.take(mth.arity) if mth.arity < args.size && mth.arity >= 0
-          mth.call(*args)
+          if mth.arity < args.size && mth.arity >= 0
```
chefsalim: hmm, seems like the same if expression?

jkeiser:
@coderanger the issue was that we were modifying the input `args` variable before; so if we truncated them to pass to a formatter that took 1 arg, then we'd have to pass 1 args to everyone (even those who expect 2).

lib/chef/event_dispatch/dispatcher.rb
jkeiser: It is indeed! The only change is we don't modify `args`.

jkeiser:
The attached tests fail without the change, and succeed with them, fyi.

jaym:
:+1:

coderanger:
Got it :+1:  :)

lamont-granquist:
:+1: FWIW

PR #4034, opened 2015-10-06
add optional ruby-profiling with --profile-ruby

lamont-granquist:
dumps a large call graph into /var/chef/cache/graph_profile.out

### DIFF
--- a/lib/chef/application/solo.rb
+++ b/lib/chef/application/solo.rb
@@ -54,7 +54,7 @@ class Chef::Application::Solo < Chef::Application
 
   option :profile_ruby,
     :long         => "--[no-]profile-ruby",
-    :description  => "Output ruby execution profile graph",
+    :description  => "Dump complete Ruby call graph stack of entire Chef run (expert only)",
     :boolean      => true,
     :default      => false
 
--- a/lib/chef/application/solo.rb
+++ b/lib/chef/application/solo.rb
@@ -1,7 +1,7 @@
 #
 # Author:: AJ Christensen (<aj@opscode.com>)
 # Author:: Mark Mzyk (mmzyk@opscode.com)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Copyright:: Copyright (c) 2008-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -49,6 +49,12 @@ class Chef::Application::Solo < Chef::Application
   option :force_formatter,
     :long         => "--force-formatter",
     :description  => "Use formatter output instead of logger output",
+    :boolean      => true,
+    :default      => false
+
+  option :profile_ruby,
+    :long         => "--[no-]profile-ruby",
+    :description  => "Output ruby execution profile graph",
     :boolean      => true,
     :default      => false
 
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -53,11 +53,6 @@ require 'chef/platform/rebooter'
 require 'chef/mixin/deprecation'
 require 'ohai'
 require 'rbconfig'
-begin
-  require 'ruby-prof'
-rescue LoadError
-  # ruby-prof is optional.
-end
 
 class Chef
   # == Chef::Client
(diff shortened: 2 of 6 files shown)


lamont-granquist:
I've been using this a lot lately.  Seems it would be good to get it into PR shape and let other people play with it.

@chef/client-core review?

lib/chef/client.rb
```diff
@@ -53,6 +53,11 @@
 require 'chef/mixin/deprecation'
 require 'ohai'
 require 'rbconfig'
+begin
+  require 'ruby-prof'
```
jaym: why not do this in start_profiling or profiling_prereqs

coderanger:
I feel like maybe this shouldn't show in the `--help` output as it is likely to be a red herring for users that want to profile their recipe code? Or at least if we want this to be a part of user workflows, we would need some tools to make it easier to filter user code vs. core code, etc.

jaym:
:+1: huge fan

ranjib:
agree with @coderanger , given chef-client does not have dependency on ruby-prof, it feels strange to expose this as CLI flag. a config option should be good enough
:+1:

jaym:
Are there any downsides to making it a dependency? I've had to walk people through profiling chef runs and having this just work would be awesome. I agree that it's not the most perfect feature, but it's still very useful.

coderanger:
@jaym I would say it would fall under the kind of thing to include with the omnibus packages but not actually be a dependency (for those of us left that still install via gems).

lamont-granquist:
Its a dev dependency for that reason.  'gem install chef' won't pull it in, but 'bundle install' (not --without deployment) will, so it'll get shipped in omnibus.

And whats the API to hide it from --help?

lib/chef/client.rb
lamont-granquist: 6 of one, half dozen of the other?  I certainly can...

danielsdeleo:
There isn't a way to hide things from help in mixlib-cli (or else we would have done this with some other flags we added just for tests). I suppose you could detect whether the gem is available at file load time and then surround the `option` call with an if statement. Otherwise you could add more explanatory text to the description.

lamont-granquist:
If I omit the description it still shows up in --help.  I really want it a CLI option because temporarily adding it to /etc/chef/client.rb is fussy, particularly if you have chef managing client.rb and have to edit it on every run.  I avoided adding it as a config option, since I don't think people will want to have this option running on every run, particularly since the outfile gets truncated and its not a log (and that seems like a good way to DoS your disk space if we made it a log since the call graph is huge).

lamont-granquist:
yeah, i could `if defined?` it, but if the gem is shipped with the omnibus packages, then it'll load for nearly everyone anyway, which defeats the purpose...

danielsdeleo:
I guess it depends on the purpose. If we just don't want to make a hard dep on the profiler gem, that'd be fine, if we want only expose this to users who know it's there, then it wouldn't help.

thommay:
I definitely agree this makes much more sense as a CLI flag than a config option. If you make with some typey-typey to make it really obvious what this does in the description I think we're good.

danielsdeleo:
I'm in agreement with @thommay's suggestion.

coderanger:
Yeah, not worth a ton of work to add hidden CLI args, but something to warn off peoples would be nice. Also would be good to move the require I think so it doesn't add to the load time for the 99% of users that don't need it (but will have the gem thanks to omnibus et al). That's a minor issue though, a few more requires aren't going to be super noticeable.

lamont-granquist:
Made the description substantially more scary.

I'll move the load in a sec, lazying the load time does marginally win in my mind over grouping the requires together.

coderanger:
:+1:

lib/chef/client.rb
```diff
@@ -891,6 +896,30 @@ def run_failed_notifications
     attr_reader :override_runlist
     attr_reader :specific_recipes
 
+    def profiling_prereqs!
+      begin
+        require 'ruby-prof'
+      rescue LoadError
+        raise "You must have the ruby-prof gem installed in order to use --profile-ruby"
+      end
```
thommay: as a truly trivial nitpick, the `begin` is redundant here no?

thommay:
I'm :+1: regardless of my comment

lib/chef/application/solo.rb
```diff
@@ -52,6 +52,12 @@ class Chef::Application::Solo < Chef::Application
     :boolean      => true,
     :default      => false
 
+  option :profile_ruby,
+    :long         => "--[no-]profile-ruby",
+    :description  => "Output ruby execution profile graph",
```
danielsdeleo: Might want to copy the scary here, too?

danielsdeleo:
Probably want to update the description for the CLI option for solo and apply also. :+1: aside from that.

PR #3959, opened 2015-09-24
pass ssh_password and some options provided by Net::SSH.configuration_for to session

sawanoboly:
option `--ssh-password` will be ignored by ssh session.

----

- I'll add spec for session_options
- And, I think that we should not truncate options from ssh_config.

PR #4038, opened 2015-10-07
keep configrations provided by Net::SSH.configuration_for and write s…

sawanoboly:
I want to use ssh options which are loaded from ssh/config.

This PR is replace of #3959

### DIFF
--- a/spec/unit/knife/ssh_spec.rb
+++ b/spec/unit/knife/ssh_spec.rb
@@ -176,7 +176,7 @@ describe Chef::Knife::Ssh do
       expect(@knife.session_options("the.b.org", nil)).to a_hash_including(:password => "mysekretpassw0rd")
     end
 
-    it "should placed set identity_file avove password" do
+    it "should prefer identity_file to password" do
       @knife.config[:identity_file] = "/tmp/mykey"
       @knife.config[:ssh_password] = "mysekretpassw0rd"
       expect(@knife.session_options("the.b.org", nil)).not_to a_hash_including(:password => "mysekretpassw0rd")
@@ -202,7 +202,7 @@ describe Chef::Knife::Ssh do
       end
     end
 
-    context "remove few options from ssh_config provides by NET::SSH" do
+    context "ignore some defaults from Net::SSH" do
       before do
         ssh_config = {:compression_level => 9, :timeout => 50, :user => "locutus", :port => 23, :send_env=>[/^LANG$/, /^LC_.*$/]}
         allow(Net::SSH).to receive(:configuration_for).with('the.b.org').and_return(ssh_config)
--- a/spec/unit/knife/ssh_spec.rb
+++ b/spec/unit/knife/ssh_spec.rb
@@ -149,6 +149,71 @@ describe Chef::Knife::Ssh do
     end
   end
 
+  describe "#session_options" do
+    let(:ssh_default_config) do
+      {:compression_level => 9, :timeout => 50, :user => "locutus", :port => 23}
+    end
+
+    before do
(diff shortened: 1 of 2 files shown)


spec/unit/knife/ssh_spec.rb
```diff
@@ -149,6 +149,71 @@ def self.should_return_specified_attributes
...
+        :keys_only => true
+      )
+    end
+
+    it "should set password" do
+      @knife.config[:ssh_password] = "mysekretpassw0rd"
+      expect(@knife.session_options("the.b.org", nil)).to a_hash_including(:password => "mysekretpassw0rd")
+    end
+
+    it "should placed set identity_file avove password" do
```
thommay: "should prefer identity_file to password"

spec/unit/knife/ssh_spec.rb
```diff
@@ -149,6 +149,71 @@ def self.should_return_specified_attributes
...
+        expect(@knife.session_options("the.b.org", nil)).to a_hash_including(:user => "dog")
+      end
+
+      it "should override port" do
+        @knife.config[:ssh_port] = 9022
+        expect(@knife.session_options("the.b.org", nil)).to a_hash_including(:port => 9022)
+      end
+    end
+
+    context "remove few options from ssh_config provides by NET::SSH" do
```
thommay: "ignore some defaults from Net::SSH"

thommay:
Please fix the failing tests and I've got a couple of grammar nits.

sawanoboly:
Fixed! Thanks @thommay

thommay:
Sweet, :+1:

thommay:
@coderanger did you have concerns about back compat on this?

coderanger:
It is probably fine, just a big enough feature-level change that I think we should all agree this is something we want to support. It means we could end up with a lot more funky usage patterns of `knife ssh` out in the wild. Like what does `Net::SSH` do with `Protocol 1`? Hopefully it is ignored, but if supported by `Net::SSH`, do we want to deal with compat with SSH 1.0? It is an improvement to Principle of Least Surprise UX and I'm :+1: on that basis, but it is opening a can of worms and we should acknowledge that before merge :)

sawanoboly:
If this policy is accepted, I'd like to create another PR to import this policy for `knife bootstrap`.

https://github.com/chef/chef/blob/ececa3438f122d25986fa34557694fa1e86527d2/lib/chef/knife/bootstrap.rb#L388-L400

coderanger:
@sawanoboly bootstrap already uses this code, so no further changes would be needed, ditto it would automatically apply to all the cloud plugins.

sawanoboly:
@coderanger Oh, I see Thanks,

sawanoboly:
@coderanger 

https://github.com/chef/chef/blob/ececa3438f122d25986fa34557694fa1e86527d2/lib/chef/knife/bootstrap.rb#L51

I probably seem that we should remove 'root' by default from bootstrap to use setting from ssh/config?

lamont-granquist:
@coderanger @sawanoboly where we at with this one?

lamont-granquist:
re-pingles?

sawanoboly:
What do you think about reducing configuration from `Net::SSH.configuration_for` ?

- Now, `chef/knife/ssh.rb` picks up options to be used.
- Tthis PR is discarding unused options.

thommay:
I think it's unclear that this PR will ever get merged at this point, so with regrets I'm going to close it. If we want to come back round and come to some kind of consensus please feel free to reopen.
Thanks for your submission and sorry we weren't able to accept it.

PR #4039, opened 2015-10-07
use git.exe on windows

### LINKED ISSUE #1622: git resource Errno::EIO on windows

thommay:
Based on the work in #3228 with fixed tests. 

Fixes #3228 Fixes #1622 
cc @chef/client-windows @gohanman

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -105,7 +105,7 @@ class Chef
       end
 
       def git_minor_version
-        @git_minor_version ||= Gem::Version.new(shell_out!('git --version', run_options).stdout.split.last)
+        @git_minor_version ||= Gem::Version.new(shell_out!(git('--version'), run_options).stdout.split.last)
       end
 
       def existing_git_clone?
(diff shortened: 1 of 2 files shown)


mwrock:
I think this is fixed by https://github.com/chef/mixlib-shellout/pull/107 which should find its way to chef soon. I think it would be better to have this fixed in mixlib-shellout rather than adding a code path for windows here.

thommay:
@mwrock ah, awesome work.

smurawski:
Since this is fixed via mixlib-shellout, I'm going to close this PR.  If we still see the behavior, we can bring this back at that time.

PR #2788, opened 2015-01-23
Implement live streaming for execute resources

adamhjk:
This brings live streaming of execute resource output to the
output formatters. It also adds a mechanism for checking to
see if an output formatter is in use through the event dispatch
system.

It adds a new configuration option, "always_stream_execute", which
does what it says on the tin.

PR #2100, opened 2014-09-22
Mixlib's shellout expects streams to have a shift operator

johnewart:
In metal we use the EventsOutputStream and pass that to shellout. Shellout does the following:

```ruby
@live_stdout << chunk if @live_stdout
```
here: https://github.com/opscode/mixlib-shellout/blob/master/lib/mixlib/shellout/unix.rb#L256 - where @live_stdout is an EventsOutputStream object, which then bombs because the object doesn't have a shift operator.

cc @jkeiser

PR #4040, opened 2015-10-07
Implement live streaming for execute resources

thommay:
This brings live streaming of execute resource output to the output
formatters. It also adds a mechanism for checking to see if an output
formatter is in use through the event dispatch system.

It adds a new configuration option, "always_stream_execute", which does
what it says on the tin.

(This is a rebase and fixup of Adam's work in #2788, cc @adamhjk @chef/client-core)
Fixes #2788

### DIFF
--- a/lib/chef/event_dispatch/dispatcher.rb
+++ b/lib/chef/event_dispatch/dispatcher.rb
@@ -23,9 +23,7 @@ class Chef
       # Check to see if we are dispatching to a formatter
       def formatter?
         @subscribers.each do |s|
-          if s.class <= Chef::Formatters::Base && s.class != Chef::Formatters::NullFormatter
-            return true
-          end
+          return s.is_formatter? if s.respond_to?(:is_formatter?)
         end
         false
       end
--- a/lib/chef/event_dispatch/dispatcher.rb
+++ b/lib/chef/event_dispatch/dispatcher.rb
@@ -20,6 +20,16 @@ class Chef
         @subscribers << subscriber
       end
 
+      # Check to see if we are dispatching to a formatter
+      def formatter?
+        @subscribers.each do |s|
+          if s.class <= Chef::Formatters::Base && s.class != Chef::Formatters::NullFormatter
+            return true
+          end
+        end
+        false
+      end
+
       ####
       # All messages are unconditionally forwarded to all subscribers, so just
       # define the forwarding in one go:
--- a/lib/chef/event_dispatch/dispatcher.rb
+++ b/lib/chef/event_dispatch/dispatcher.rb
@@ -22,10 +22,7 @@ class Chef
 
       # Check to see if we are dispatching to a formatter
       def formatter?
-        @subscribers.each do |s|
-          return s.is_formatter? if s.respond_to?(:is_formatter?)
-        end
-        false
+        @subscribers.any? { |s| s.respond_to?(:is_formatter?) && s.is_formatter? }
       end
 
       ####
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -78,6 +78,14 @@ class Chef
         !!new_resource.sensitive
       end
 
+      def stream_to_formatter?
+        Chef::Config[:always_stream_execute] || run_context.events.formatter?
+      end
+
+      def stream_to_stdout?
+        STDOUT.tty? && !Chef::Config[:daemon] && Chef::Log.info?
+      end
+
       def opts
         opts = {}
         opts[:timeout]     = timeout
@@ -89,10 +97,12 @@ class Chef
         opts[:umask]       = umask if umask
         opts[:log_level]   = :info
         opts[:log_tag]     = new_resource.to_s
-        if (Chef::Config[:always_stream_execute] || run_context.events.formatter?) && !sensitive?
-          opts[:live_stream] = Chef::EventDispatch::EventsOutputStream.new(run_context.events, :name => :execute)
-        elsif STDOUT.tty? && !Chef::Config[:daemon] && Chef::Log.info? && !sensitive?
-          opts[:live_stream] = STDOUT
+        unless sensitive?
+          if stream_to_formatter?
+            opts[:live_stream] = Chef::EventDispatch::EventsOutputStream.new(run_context.events, :name => :execute)
+          elsif stream_to_stdout?
+            opts[:live_stream] = STDOUT
+          end
         end
         opts
       end
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -79,11 +79,7 @@ class Chef
       end
 
       def live_stream?
-        !!new_resource.live_stream
-      end
-
-      def stream_to_formatter?
-        Chef::Config[:stream_execute_output] && run_context.events.formatter?
+        Chef::Config[:stream_execute_output] || !!new_resource.live_stream
       end
 
       def stream_to_stdout?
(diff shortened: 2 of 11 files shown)


thommay:
(I'll rebase before merge)

thommay:
Fixes #2100

lib/chef/provider/execute.rb
```diff
@@ -80,8 +88,12 @@ def opts
         opts[:umask]       = umask if umask
         opts[:log_level]   = :info
         opts[:log_tag]     = new_resource.to_s
-        if STDOUT.tty? && !Chef::Config[:daemon] && Chef::Log.info? && !sensitive?
-          opts[:live_stream] = STDOUT
+        unless sensitive?
+          if stream_to_formatter?
+            opts[:live_stream] = Chef::EventDispatch::EventsOutputStream.new(run_context.events, :name => :execute)
```
coderanger: Do we want to store a reference to the actual resource object for the future? Seems like the kind of thing that might be useful to formatters.

danielsdeleo:
I haven't reviewed yet, but thanks for getting on this. SO AWESOME.

lib/chef/provider/execute.rb
```diff
@@ -69,6 +69,14 @@ def sensitive?
         !!new_resource.sensitive
       end
 
+      def stream_to_formatter?
+        Chef::Config[:always_stream_execute] || run_context.events.formatter?
```
danielsdeleo: Did we want to make streaming an option on the resource? Seems like `run_context.events.formatter?` will be true whenever you run `chef-client` with default options on the command line; you could disable it with `sensitive(true)` but that hides outputs in the exception case, too, which wouldn't be what you want if you just want to suppress a really verbose command for only the success case.

danielsdeleo:
:+1: on the code that's here, though let's be sure that the behavior is what we want. AFAICT it will pretty much always stream the output, which makes the config option not-so-useful and doesn't give you a good option for suppressing streams if you have a really verbose command that you want to run.

lib/chef/provider/execute.rb
coderanger: I would think for compat with the current behavior this should be `&&`.

danielsdeleo: That makes sense, though I could see wanting to have a per-resource setting, some executes are more interesting than others.

coderanger: Yeah, also because we have an `elsif` below, this would totally change the behavior of `-l debug` as far as showing command output. I think we should probably rewind and outline the desired semantics in English before working with them as code.

schisamo:
@thommay Let's address the last few comments and get this :ship:!

coderanger:
@schisamo I think that is putting the cart before the horse. We still need to actually describe how we want this system to work.

danielsdeleo:
I think it's pretty straightforward:

* If `Chef::Config.always_stream_execute` is true and the formatter supports streaming and the resource doesn't set `sensitive`, then stream.
* executes should have per-resource streaming opt-in
* If `Chef::Config.always_stream_execute` is false, and `resource.live_stream` is true, and the formatter supports streaming and the resource doesn't set `sensitive`, then stream (sensitive trumps the live_stream setting).

coderanger:
@danielsdeleo How does that interact with `-l debug`?

lib/chef/provider/execute.rb
btm: Isn't this going to cause the formatter to always receive and thus print stream output, whereas currently we only do so when log level is info or higher?

It seems to me like the model is that log level determines what is sent to the formatters, rather than making decisions about what to print in the formatter based on log level.

```
( Chef::Config[:always_stream_execute] || Chef::Log.info? ) && run_context.events.formatter?
```

btm:
@danielsdeleo I generally agree with that, but I'd like to understand where log level fits in. Is it inappropriate by design to switch on log level for event output streams?

Are we saying `Chef::Config.always_stream_execute` means "Always send execute streams to events?"

danielsdeleo:
I think we shouldn't make a large change to the current behavior, so having a log level of info should stream to stdout if you have a tty and debug should always stream (correct me if I'm wrong about what the current behavior is, I wrote it but I could be misremembering). The exception to this I think is that we shouldn't double-stream. Sound correct?

btm:
`Chef::Log.info?` returns true if your log level is >= info, so currently we stream to STDOUT if your log_level is info or debug. I agree we shouldn't double-stream. My concern is how do you switch to having the execute output streams to go through events instead of to STDOUT, but only have them sent when you want debugging? 

I think we use the existing logic: only send output (to STDOUT or events) when log level is >= info and sensitive? is false, and have a config setting that controls if that goes to STDOUT or events.

Then, if you have a feature where you always want the output of some resource to get sent to events, you have a resource parameter for that specific case that doesn't care what the log level or aforementioned config setting are (but still is overridden by sensitive).

danielsdeleo:
@btm's proposal sounds good to me.

thommay:
Yup, I'll implement @btm's proposal.

schisamo:
:clap: :metal:

thommay:
@danielsdeleo @btm @coderanger so I think this should reflect the discussion now

lib/chef/provider/execute.rb
```diff
@@ -78,6 +78,18 @@ def sensitive?
         !!new_resource.sensitive
       end
 
+      def live_stream?
+        !!new_resource.live_stream
```
danielsdeleo: Seems like this should be true if either:
* `Chef::Config[:stream_execute_output]` is true
* `new_resource.live_stream` is true

Right? I'm thinking that `Chef::Config[:stream_execute_output]` means "stream every execute resource (except if marked sensitive)" and `new_resource.live_stream` means "stream this one execute resource, even if streaming for execute is disabled globally". Is that right?

btm: :+1: on @danielsdeleo thinking. #live_stream? should be true in those cases. Then the conditional below around new line 104 changes to:

```
if (Chef::Log.info? || live_stream?) && !sensitive?
  if run_context.events.formatter?
    opts[:live_stream] = Chef::EventDispatch::EventsOutputStream.new(run_context.events, :name => :execute)
  elsif stream_to_stdout?
    opts[:live_stream] = STDOUT
  end
end
```

lib/chef/event_dispatch/dispatcher.rb
```diff
@@ -20,6 +20,14 @@ def register(subscriber)
         @subscribers << subscriber
       end
 
+      # Check to see if we are dispatching to a formatter
+      def formatter?
+        @subscribers.each do |s|
```
coderanger: This has odd semantics with multiple subscribers. Wouldn't we want to check if either `any?` or `all?` rather than just the first that happens to respond?

coderanger: From the code below, I think this should be an `any?`:

```ruby
@subscribers.any? {|s| s.respond_to?(:is_formatter?) && s.is_formatter? }
```

danielsdeleo: Agreed, I think `any?` is the correct logic here. On my first reading I mistakenly thought this was logically equivalent to that (after some coffee I see that it's not).

thommay:
@danielsdeleo @coderanger done and done.

danielsdeleo:
:+1:

lib/chef/resource/execute.rb
```diff
@@ -101,6 +102,13 @@ def group(arg=nil)
         )
       end
 
+      def live_stream(arg=nil)
+        set_or_return(
+          :live_stream,
+          arg,
+          :kind_of => [ TrueClass, FalseClass ])
```
coderanger: This should default to `false` instead of `nil`, as `nil` isn't a valid value.

coderanger: Never mind, I see it does above. Any reason to not use the `default:` key with set_or_return?

coderanger:
:+1: :shipit:

lamont-granquist:
:+1: awesomesauce

PR #4049, opened 2015-10-09
Add gemspec files to allow bundler to run from the gem

ksubrama:
/cc @jkeiser @chef/client-core

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ gemspec :name => "chef"
 
 gem "activesupport", "< 4.0.0", :group => :compat_testing, :platform => "ruby"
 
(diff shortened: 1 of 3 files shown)


danielsdeleo:
:+1: please just run through a gem build/install to make sure we don't have any issues like extraneous files (like in the ChefDK version of this).

lamont-granquist:
:+1:

ksubrama:
@danielsdeleo Yep - again one extra issue identified (chef-config is not available at the path specified in the Gemfile) that wouldn't have hurt gem build/install but did cause bundler to freak out.  Fixed that.

danielsdeleo:
Cool, :+1:

PR #4054, opened 2015-10-13
suppressing sensitive properties

chasebolt:
make use of the sensitive property to suppress sensitive output from converge_if_changed.

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -204,7 +204,7 @@ class Chef
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          properties_str = if @sensitive
+          properties_str = if sensitive
             '(suppressed sensitive properties)'
           else
             specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")
@@ -216,7 +216,7 @@ class Chef
         # Print the pretty green text and run the block
         property_size = modified.map { |p| p.size }.max
         modified.map! do |p|
-          properties_str = if @sensitive
+          properties_str = if sensitive
             '(suppressed sensitive properties)'
           else
             "#{new_resource.send(p).inspect} (was #{current_resource.send(p).inspect})"
@@ -231,7 +231,7 @@ class Chef
         property_size = properties.map { |p| p.size }.max
         created = properties.map do |property|
           default = ' (default value)' unless new_resource.property_is_set?(property)
-          properties_str = if @sensitive
+          properties_str = if sensitive
             '(suppressed sensitive properties)'
           else
             new_resource.send(property).inspect
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -204,26 +204,39 @@ class Chef
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")}.")
+          properties_str = if @sensitive
+            '(suppressed sensitive properties)'
+          else
+            specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")
+          end
+          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{properties_str}.")
           return false
         end
 
         # Print the pretty green text and run the block
         property_size = modified.map { |p| p.size }.max
-        modified = modified.map { |p| "  set #{p.to_s.ljust(property_size)} to #{new_resource.send(p).inspect} (was #{current_resource.send(p).inspect})" }
+        modified.map! do |p|
+          properties_str = if @sensitive
+            '(suppressed sensitive properties)'
+          else
+            "#{new_resource.send(p).inspect} (was #{current_resource.send(p).inspect})"
+          end
+          "  set #{p.to_s.ljust(property_size)} to #{properties_str}"
+        end
         converge_by([ "update #{current_resource.identity}" ] + modified, &converge_block)
 
       else
         # The resource doesn't exist. Mark that we are *creating* this, and
         # write down any properties we are setting.
         property_size = properties.map { |p| p.size }.max
-        created = []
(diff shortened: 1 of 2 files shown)


lib/chef/provider.rb
```diff
@@ -204,13 +204,21 @@ def converge_if_changed(*properties, &converge_block)
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")}.")
+          if @sensitive == true
+            Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties (suppressed sensitive properties).")
+          else
+            Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")}.")
+          end
```
lamont-granquist: Can we do this more like:

```ruby
  properties_str = if sensitive
    "(suppressed sensitive properties)"
  else
    specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")
  end
 Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{properties_str}.")
```

lamont-granquist:
(and some comment about the other two blocks as well...)

smurawski:
Other than @lamont-granquist 's style comments :+1:

chasebolt:
alright style has been updated to your suggestion @lamont-granquist

lamont-granquist:
looks like travis got angry about something in that last commit though...

chasebolt:
looks good now except for appveyor failing out on what appears to be timeouts with a call to chef-server via knife.

chasebolt:
@lamont-granquist all green

lib/chef/provider.rb
```diff
@@ -204,26 +204,39 @@ def converge_if_changed(*properties, &converge_block)
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")}.")
+          properties_str = if @sensitive
```
jkeiser: Please use `sensitive` rather than `@sensitive`, so that we can change up how things are implemented in the future.

lib/chef/provider.rb
```diff
@@ -204,26 +204,39 @@ def converge_if_changed(*properties, &converge_block)
...
+          end
+          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{properties_str}.")
           return false
         end
 
         # Print the pretty green text and run the block
         property_size = modified.map { |p| p.size }.max
-        modified = modified.map { |p| "  set #{p.to_s.ljust(property_size)} to #{new_resource.send(p).inspect} (was #{current_resource.send(p).inspect})" }
+        modified.map! do |p|
+          properties_str = if @sensitive
```
jkeiser: Ditto from above

lib/chef/provider.rb
```diff
@@ -204,26 +204,39 @@ def converge_if_changed(*properties, &converge_block)
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          Chef::Log.debug("Skipping update of #{new_resource.to_s}: has not changed any of the specified properties #{specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect}" }.join(", ")}.")
```
jkeiser: For this one, can you change the `if sensitive` branch to just not print the values? Like this:

```ruby
if sensitive
  properties_str = specified_properties.map { |p| "#{p}=#{new_resource.send(p).inspect" }.join(", ")
else
  properties_str = specified_properties.join(", ")
end
```

This way, you still get the useful information about what properties changed, just not what their values were.

lib/chef/provider.rb
```diff
@@ -204,26 +204,39 @@ def converge_if_changed(*properties, &converge_block)
...
         # The resource doesn't exist. Mark that we are *creating* this, and
         # write down any properties we are setting.
         property_size = properties.map { |p| p.size }.max
-        created = []
-        properties.each do |property|
-          if new_resource.property_is_set?(property)
-            created << "  set #{property.to_s.ljust(property_size)} to #{new_resource.send(property).inspect}"
+        created = properties.map do |property|
+          default = ' (default value)' unless new_resource.property_is_set?(property)
+          properties_str = if @sensitive
```
jkeiser: @sensitive -> sensitive

lib/chef/provider.rb
```diff
@@ -204,26 +204,39 @@ def converge_if_changed(*properties, &converge_block)
...
+            '(suppressed sensitive properties)'
           else
-            created << "  set #{property.to_s.ljust(property_size)} to #{new_resource.send(property).inspect} (default value)"
+            new_resource.send(property).inspect
           end
+          "  set #{property.to_s.ljust(property_size)} to #{properties_str}" + default ||= ''
```
jkeiser: Why not `"#{property.to_s.ljust(property_size)} to #{properties_str}#{default}"`? If default is nil, it should just not print anything in the string. The ||= at the end of this is a little confusing.

jkeiser:
This needs a test. At a minimum, I'd have a `sensitive` version of these two:

- https://github.com/chef/chef/blob/master/spec/integration/recipes/resource_converge_if_changed_spec.rb#L91-L111
- https://github.com/chef/chef/blob/master/spec/integration/recipes/resource_converge_if_changed_spec.rb#L397-L420

That covers the update and the create case.

jkeiser:
@chasebolt thanks so much! I love the patch. Very last thing, and I will totally do it myself before merge if you don't get to it: `- set state1 to (suppressed sensitive properties)` in the plural makes less sense than `- set state1 to (suppressed sensitive property)`, so changing it to the singular might help.

jkeiser:
Checked in as  cc51b2f0504a32d7e55d2c445f22baf9ed30b031 . Thanks!

chasebolt:
That does makes more sense! Thanks for updating it! Can you make sure to
get this into compact_resource as well so we can close the chef-docker
issue?

jkeiser:
The cookbook shipped, FYI :)

PR #4062, opened 2015-10-15
Add immediately_before notification

jkeiser:
Implements https://github.com/chef/chef-rfc/pull/154


lamont-granquist:
Another way to do this would be to put an always_why_run flag on to all resources and then if its true its always why-run but then you can chain a notify to a service restart and then to another package upgrade resource.  The problem there is that then the user must duplicate two resources exactly other than the first one does the why-run mode, along with the ceremony of setting up the chain.  That seems more error prone and more ceremonial.

We could also solve this in the DSL with helpers like `package_installable?` and `package_upgradable?` (which i want anyway) that would look something like this:

```
def package_installable?(name)
  !build_resource(:package, name).current_value.candidate_version.nil?
end

def package_upgradable?(name)
  cv = build_resource(:package, name).current_value
  cv.installed_version != cv.candidate_version
end

if package_upgradable?('foo')
  service "foo" do
    action :stop
  end

  package "foo" do
    action :upgrade
  end
end
```

lamont-granquist:
I'm :+1:  on this idea though.

thommay:
@jkeiser can you update this one in the light of the RFC comments and then we'll get it merged.

lamont-granquist:
cheffish really hates this PR for some method_missing reason....

jkeiser:
@lamont-granquist I fixed the cheffish issue on this side, because I'm reluctant to release a new cheffish just yet for independent reasons (will explain in a bit). Updated the PR.

lamont-granquist:
should do the thing :+1:

lamont-granquist:
don't forget to manually add a CHANGELOG as well...

PR #4064, opened 2015-10-16
Use the initializer to avoid NoMethodError on nil.include?

josb:
Fixes ``` undefined method `include?' for nil:NilClass``` when using tag() in the recipe DSL.

### DIFF
--- a/lib/chef/knife/core/node_presenter.rb
+++ b/lib/chef/knife/core/node_presenter.rb
@@ -125,7 +125,7 @@ ROLES
             summarized << <<-SUMMARY
 #{key('Recipes:')}     #{Array(node[:recipes]).join(', ')}
 #{key('Platform:')}    #{node[:platform]} #{node[:platform_version]}
-#{key('Tags:')}        #{Array(node[:tags]).join(', ')}
+#{key('Tags:')}        #{node.tags.join(', ')}
 SUMMARY
             if config[:medium_output] || config[:long_output]
               summarized +=<<-MORE
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -376,12 +376,12 @@ class Chef
       normal[:tags]
     end
 
-    def tag(*tags)
-      tags.each do |tag|
-        self.normal[:tags].push(tag.to_s) unless self[:tags].include? tag.to_s
+    def tag(*args)
+      args.each do |tag|
+        tags.push(tag.to_s) unless tags.include? tag.to_s
       end
 
-      self[:tags]
+      tags
     end
 
     # Extracts the run list from +attrs+ and applies it. Returns the remaining attributes
--- a/lib/chef/recipe.rb
+++ b/lib/chef/recipe.rb
@@ -97,8 +97,6 @@ class Chef
     # true<TrueClass>:: If all the parameters are present
     # false<FalseClass>:: If any of the parameters are missing
     def tagged?(*tags)
-      return false if run_context.node[:tags].nil?
-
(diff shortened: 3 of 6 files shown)


lamont-granquist:
fix is right, but could use a test.

josb:
A problem with this issue is that recipe_spec.rb does things like
```
Chef::Node.new.tap {|n| n.normal[:tags] = [] }
```
This initializes the object to ```[]```, whereas it would appear that the "regular" code does not (hence the issue I am seeing). So the irony here is that the test framework initialization hides this bug.

lamont-granquist:
node.normal[:tags] is a VividMash and will autovivify to a Mash, so it may be getting initialized on access somewhere.

probably everything should be cleaned to go in through Chef::Node#tags and Chef::Node#tag as the APIs an nothing else should access tags as a hash/mash/attribute directly.

the line in recipe_spec looks pretty wrong and should most likely just be dropped.  looks kind of like someone found the bug and then 'fixed' the test instead of fixing the bug...

josb:
Thanks for the feedback, Lamont.

This should probably be a different issue (i.e. not block this one from landing) but it sounds like we want s/(node|normal)[:tags]/tags/ except for inside the tags method itself? There are quite a few occurrences of the existing access pattern, hence my question.

btm:
@josb in `lib/chef/recipe.rb` the method `Chef::Recipe#tagged?` does `return false if run_context.node[:tags].nil?`, which is correct. We expect methods ending in ? to return boolean values.

`chef$ git grep '\[:tags\]' | grep -v ^spec | grep -v 'config\[:tags\]' | wc` => 11 lines, and 4 of those are the lines inside node that we're talking about saving. That is, I think we should fix tags as @lamont-granquist suggested since it's only a few lines in the code and the associated tests, and see how that changes the behavior so we can come out with a regression test.

josb:
@btm: ```node[:tags]``` being ```nil``` is a bug. We should not have to check for that value, as it's not a valid value for this particular variable. (I'm aware of the Boolean return value convention, btw).

I fixed the other occurrences, please see the current diff. Thanks for your feedback!

lib/chef/knife/core/node_presenter.rb
```diff
@@ -125,7 +125,7 @@ def summarize(data)
             summarized << <<-SUMMARY
 #{key('Recipes:')}     #{Array(node[:recipes]).join(', ')}
 #{key('Platform:')}    #{node[:platform]} #{node[:platform_version]}
-#{key('Tags:')}        #{Array(node[:tags]).join(', ')}
+#{key('Tags:')}        #{node.tags.join(', ')}
```
btm: We're super sure now that node.tags will always give us an array, because that's part of the bug, right?

josb: Correct. I fixed all the places where we access node[:tags] which was the only way we'd never initialize it to an Array. Chef::Node#tags makes sure we initialize properly.

btm:
:+1: 

@chef/client-core, review?

lamont-granquist:
I just did a quick skim, :+1: on the overall shape, I didn't get into details.

lamont-granquist:
:+1:  on the details as well

thommay:
:+1: , thanks @josb!

lamont-granquist:
:+1: 

if you want to address Dan's `respond_to?(:read)` comment that'd be cool as well, that's an idiom that I've seen before.

josb:
Thanks, everyone :)

PR #4068, opened 2015-10-19
Update messaging to match LWRP -> Custom Resource rename

tas50:
If you showed up in a post-LWRP world these would be confusing


smurawski:
Merged in #4085

PR #4069, opened 2015-10-19
Improvements to log messages

tas50:


mcquin:
Needs unit tests. :trollface: 

There's a few more capitalization issues to find in the provider directory. A sloppy search with `Chef::Log.[a-z]+[\(|\s]"([a-z])` turned those up.

PR #4078, opened 2015-10-20
Added profile provider for OS X

natewalck:
New provider to install configuration profiles on OS X. 

Once fully reviewed and approved, @jaymzh will merge the changes under his CLA.


natewalck:
I added the gem to the chef.gemspec file, yet it still says it cannot require the gem in the test. Not entirely sure how to fix that.

jaymzh:
@natewalck Do the tests run on your localhost? Start there, its much easier to debug :)

natewalck:
@jaymzh Yep! All the tests I wrote work locally. The actual error in the log is a fail on:
```
  require 'uuidtools'
```

natewalck:
Actually, there may be a name collision with 'profile' as my resource name. Hrmmm.

jaymzh:
Well yes, I got your tests work - when you run just them. It's worth running the suite and seeing if failures fall out of that though... I can always help debug they do.

natewalck:
@jaymzh I forgot to run all the tests at once. I'll do that and figure out what is going on.

natewalck:
Woot! Tests are now passing.

@jtimberman @tyler-ball @danielsdeleo Feel free to tell me how bad this is and what I need to fix :laughing:

jaymzh:
@chef/client-core for review.

natewalck:
Hey all! I was hoping to get some feedback on this. Let me know what needs to be changed.

thommay:
@chef/client-os-x would be good for y'all to take a peak.

lib/chef/provider/profile.rb
```diff
@@ -0,0 +1,253 @@
...
+
+      def define_resource_requirements
+        requirements.assert(:remove) do |a|
+          if @new_profile_identifier
+            a.assertion {
+              !@new_profile_identifier.nil? and
+              !@new_profile_identifier.end_with?('.mobileconfig') and
+              /^\w+(?:\.\w+)+$/.match(@new_profile_identifier)
+            }
+            a.failure_message Chef::Exceptions::UnsupportedAction, "when removing using the identifier attribute, it must match the profile identifier"
```
btm: `Chef::Exceptions::UnsupportedAction` is used when `resource.action` is not supported. These errors are more specific to profile, so I'd add a `Chef::Exceptions::Profile` exception or just use RuntimeError if we don't think we'd use profile in another provider.

natewalck:
Changed it to RuntimeError as configuration profiles are an OS X specific service.

btm:
It'd be good to review the instance variables in `lib/chef/provider/profile.rb`. Some look like they can just be local variables, like `@all_profiles`. Others I think are a code smell that the code would be more understandable if you made your methods perform a single specific action that's clear by their name which act against arguments and return the result.

For instance, `get_new_profile_hash` is also a bit overloaded. I'd change `get_new_profile_hash` to something like `lookup_profile` and pass it `@new_resource.profile_name` as an argument,  then `@new_profile` doesn't need to be an instance variable. Take the logic about the action out of `get_new_profile_hash` and put it back into `load_current_resource`. For example:

```
# we should have an assertion that makes sure new_profile is going to be a String or Hash.
new_profile = @new_resource.profile || lookup_profile(@new_resource.profile_name)
new_profile['PayloadUUID'] = config_uuid(new_profile)
@new_resource.identifier = new_profile['PayloadIdentifier'] || 
    @new_resource.identifier ||
    @new_resource.profile_name
end
```

natewalck:
@btm I've made some suggested changes. Because of the behavior difference when install vs remove, I still need to check to see is new_resource-hash exists or not so it does not error out on removal.

I'll look over some other ways I can simplify the code, but suggestions are definitely welcome ;)

lib/chef/provider/profile.rb
```diff
@@ -0,0 +1,258 @@
...
+            install_profile(profile_path)
+            get_installed_profiles(update=true)
+          end
+        end
+      end
+
+      def action_remove
+        # Clean up profile after removing it
+        if @current_resource.installed
+          converge_by("remove profile #{@new_resource.profile_name}") do
```
btm: The install message prints `@new_profile_hash['PayloadIdentifier']` but we print `@new_resource.profile_name`. Would one or the other for both be better?

lib/chef/provider/profile.rb
```diff
@@ -0,0 +1,258 @@
...
+
+      def load_profile_hash(new_profile)
+          # file must exist in cookbook
+          unless new_profile.end_with?('.mobileconfig')
+            error_string = "'#{new_profile}' is not a valid profile"
+            raise RuntimeError, error_string
+          end
+          unless cookbook_file_available?(new_profile)
+            error_string = "'#{new_profile}' not found in cookbook"
+            raise Chef::Exceptions::FileNotFound, error_string
```
btm: depending on log level, this could be a vague error. Consider something like `"#{self.to_s}: '#{new_profile}' not found in cookbook"`

btm: that is, you wouldn't really know where it came from. unless your profile name was "my_profile" you might not realize it had anything to do with the profile provider.

natewalck:
@btm Quite right! 

I am using the global @new_profile_identifier as it is accurately set for both installs and removals.

lib/chef/provider/profile.rb
natewalck: @btm Quite right! 

I am using the global @new_profile_identifier as it is accurately set for both installs and removals.

natewalck:
Fixed the vague error message.

lamont-granquist:
`profile` is a fairly generic resource name and doesn't make a lot of sense when this provider is so specific to mac, and its not clear that this is a generic cross-platform term like `package` where we could reasonably expect to also have Ubuntu or FreeBSD providers for it.

a better naming would probably be to start out with `mac_profile` or `darwin_profile` or `osx_profile`.

at some point in the future if we did wind up with other `XXX_profile` resources that had similar shape then we could wire up a pseudo-resource `profile` on the DSL, but I don't think we should do that yet (and kind of expect we shouldn't do that at all, but I will leave that open).

natewalck:
I had thought the naming might be a little too generic. Would ```config_profile``` work, or is that still too generic?

lamont-granquist:
That still feels a bit too generic or awkward or something.

natewalck:
```osx_config_profile``` ? Alias to ```osx_profile```?

**Comment by natewalck** [comment]
Let me know if either of the above names would work. I'd like to make whatever changes are necessary to get this ready for merging. Thanks @lamont-granquist @btm for the code reviews!

**Comment by chef-supermarket** [comment]

* @natewalck

**Comment by natewalck** [comment]
Hey All! Hoping to get the review completed on this so I can make whichever changes are necessary to get it merged.

**Comment by natewalck** [comment]
This has been renamed to osx_profile to make it clear.

Also, could client maintainers for OS X take a look? Thanks!

**Comment by danielsdeleo** [comment]
@chef/client-os-x

**Comment by jaymzh** [comment]
pinging @chef/client-core  and @chef/client-os-x

File: lib/chef/resource/osx_profile.rb [hunk 4]
```diff
@@ -0,0 +1,84 @@
...
+
+      def profile_name(arg=nil)
+        set_or_return(
+          :profile_name,
+          arg,
+          :kind_of => [ String ]
+        )
+      end
+
+      def installed(arg=nil)
```
tyler-ball: Totally showing my newb status here - I see that this is a `state_attr` so it will be sent to the reporting server.  But it seems weird to have a set-able property that seems to me to be read_only.

Its been a while since I've written a non-provisioning resource - is this just a common pattern?

lib/chef/provider/osx_profile.rb
```diff
@@ -0,0 +1,258 @@
...
+        )
+        remote_file.cookbook_name = @new_resource.cookbook_name
+        remote_file.source(cookbook_file)
+        remote_file.backup(false)
+        remote_file.run_action(:create)
+        remote_file.path
+      end
+
+      def get_profile_hash(new_profile)
+        if @new_resource.action.include?(:install)
```
tyler-ball: I'm unsure why this check is needed - I would think this method would just convert a string_or_hash into a hash?

lib/chef/provider/osx_profile.rb
```diff
@@ -0,0 +1,258 @@
...
+        tempfile = ::Dir::Tmpname.create("allprofiles.plist") {}
+      end
+
+      def write_installed_profiles(tempfile)
+        cmd = "profiles -P -o '#{tempfile}'"
+        shell_out!(cmd)
+      end
+
+      def read_plist(xml_file)
+        installed_profiles = Plist::parse_xml(xml_file)
```
tyler-ball: This can just be `Plist::parse_xml(xml_file)`

tyler-ball:
@natewalck Before I go through here and make a bunch of comments, I want to understand better - what is the 'behavior difference when install vs remove' (besides the obvious fact that one action is trying to install and the other is trying to remove)?

I ask this question because it seems like you're setting a bunch of instance variables in `load_current_resource` that don't make sense to me.  In my naive view, I would think `load_current_resource` (and associated methods) would look something like:

```ruby
def new_profile_hash
  @new_profile_hash ||= begin
    h = get_profile_hash(new_resource.profile || new_resource.profile_name)
    h['PayloadUUID'] = config_uuid(h) if h
    h
  end
end

def new_profile_identifier
  @new_profile_identifier ||= begin
    if new_profile_hash
      new_profile_hash['PayloadIdentifier'] || new_resource.identifier || new_resource.profile_name
    else
      new_resource.identifier || new_resource.profile_name
    end
  end
end

def load_current_resource
  @current_resource = Chef::Resource::OsxProfile.new(new_resource.name)
  @current_resource.profile_name(new_resource.profile_name)

  all_profiles = get_installed_profiles
  current_profile = all_profiles['_computerlevel'].find {
    |item| item['ProfileIdentifier'] ==
       new_profile_identifier
  }

  @current_resource.profile(current_profile) if current_profile
  
  # Profile Identifier and UUID must match a currently installed profile
  if @current_resource.profile.nil? or @current_resource.profile.empty?
    @current_resource.installed(false)
  else
    if new_resource.action.include?(:remove)
      @current_resource.installed(true)
    else
      match = @current_resource.profile['ProfileUUID'] == new_profile_hash['PayloadUUID']
      @current_resource.installed(match)
    end
  end
end

# I don't understand why this was switching on the action
def get_profile_hash(new_profile)
  if new_profile.is_a?(Hash)
    new_profile
  elsif new_profile.is_a?(String)
    load_profile_hash(new_profile)
  else
    nil
  end
end
```

natewalck:
@tyler-ball To remove a profile on OS X, you only need the Profile Identifier. To install a profile on OS X, you need the entire configuration profile. Supplying the entire profile just to remove it is inefficient and if the profile was nuked from orbit and no longer exists, you still need to be able to remove it.

For instance, removing a profile can be as simple as:

```
  profile 'com.cpe.profiles.screensaver' do
    action :remove
  end
```

Whereas to install a profile, you need to provide the name of the config profile that lives in the cookbook or the profile as a hash. 

```
  # Use a profile from the cookbook
config_profile 'nwalck/com.cpe.profiles.screensaver.mobileconfig'
```

```
  # Install using a hash
  config_profile 'Install screensaver profile' do
    profile profile_hash
  end
```

When it comes to OS X config profiles, it should not be a hard requirement to supply the original profile in order to remove it (as just the identifier is sufficient).

Hope that clarifies how it works a bit. I'm happy to answer any other questions about the mechanism.

lib/chef/provider/osx_profile.rb
natewalck: I broke this out into a hash so it was easier to mock in rspec. I may be doing it wrong, so if there is a proper way to mock it without moving it to a function, I don't mind doing so.

lib/chef/resource/osx_profile.rb
natewalck: I have no idea. #makeituntilyoumakeit

lib/chef/provider/osx_profile.rb
jaymzh: he's asking why you're checking the action.

Which is what you and I talked about before. Check if the user passed in the data, if they did, return it, if not, don't. Don't do it based on action. You check check the person passed in the right data for the right action in `define_resource_requirements` - but don't change how you handle the data based on that. This can just be:

```
if new_profile.is_a?(Hash)
  return new_profile
elsif new_profile.is_a?(String)
  return load_profile_hash(new_profile)
end
```

And done.

natewalck: I see the logic here, but in the case of a removal, if the profile_name is set to the identifier, it should use that as the profile_identifier.

This allows a smaller resource call as such:
``` 
 profile 'com.cpe.profiles.screensaver' do
    action :remove
  end
```

Otherwise, you would need to specify:
```
  profile 'Remove some profile' do
    identifier 'com.cpe.profiles.screensaver'
    action :remove
  end
```

If it is a removal, it should know the resource name should be used as the identifier instead of profile name. 

One solution is if it ends in .mobileconfig, assume it is an install, if it does not, assume it is an identifier? This also seems kind of janky. I am assuming nothing with the current code, but if I have a better way to do it in a predictable fashion, I am happy to do it. (Perhaps I am missing something?)

natewalck: With the change suggested by @jaymzh, This should work if I also modify load_profile_hash like so: 

```
      def load_profile_hash(new_profile)
          # file must exist in cookbook
          if new_profile.end_with?('.mobileconfig')
            unless cookbook_file_available?(new_profile)
              error_string = "#{self.to_s}: '#{new_profile}' not found in cookbook"
              raise Chef::Exceptions::FileNotFound, error_string
            end
            cookbook_profile = cache_cookbook_profile(new_profile)
            return read_plist(cookbook_profile)
          end
      end
```

natewalck:
@tyler-ball @jaymzh I believe I have addressed all the suggestions presented except for the state_attr. This feels a lot cleaner as I am linting the input based on action in define requirements as PhilD suggested.

jaymzh:
It *sounds* like the identifier (`com.cpe.profiles.screensaver`) should be the name_parameter, except I'm *guessing* here that when you add you don't need to specify the name parameter at all because it's somewhere in the file and/or hash? Which is where the disconnect is coming from.

But that's really unrelated to the action, it's just about pulling the identifier from the best place, like:

```
if config_profile
  if config_profile.is_a?(Hash)
    profile = load_config(config_profile)
  elsif config_profile.is_a?(String)
    profile = load_config_file(config_profile)
  end
  @identifier = profile['PayloadIdentifier']
else
  @identifier = new_resource.name
end
```

Right? I think your new commit is doing something similar, though you seem to do that same fall through in a lot more places. (to be clear, I'm not necessarily asking for a change, mostly making sure I'm following).

natewalck:
That is exactly correct. The identifier is always inside the profile (either hash or file) and should always be used from there. If that identifier is missing from inside the profile, it will not install.

The only time we really care about passing identifier is for removals, in which case we cannot do the removal if we are missing the identifier.

I like how concise your code was, so I'll probably do another diff because mine is far less clear.

natewalck:
I decided to leave the logic mostly in tact, but removed the extra identifier logic. It will now fail if you don't have a PayloadIdentifier within the profile.

natewalck:
Let me know what else needs to be done on this. I'd like to get this finished up.

jaymzh:
@chef/client-core and @chef/client-os-x - ya'll wanna do another pass over it? I'm :+1:  now.

lib/chef/resource/osx_profile.rb
jtimberman: It's a common pattern. For example:

* https://github.com/chef/chef/blob/master/lib/chef/resource/package.rb#L51-L57
* https://github.com/chef/chef/blob/master/lib/chef/resource/file.rb#L80-L86

There are others.

jtimberman:
:+1: - no objections from me.

spec/unit/provider/osx_profile_spec.rb
```diff
@@ -0,0 +1,251 @@
...
+      expect(
+        @provider.instance_variable_get(:@new_profile_identifier)
+        ).to eql(@test_profile['PayloadIdentifier'])
+    end
+
+    it 'should install when not installed' do
+      @new_resource.profile @test_profile
+      allow(@provider).to receive(:get_installed_profiles).and_return(@no_profiles)
+      @provider.load_current_resource
+      expect { @provider.run_action(:install) }
```
tyler-ball: There should be an expectation here, even if it is `expect { ... }.to_not raise_error`

spec/unit/provider/osx_profile_spec.rb
```diff
@@ -0,0 +1,251 @@
...
+               "PayloadType"=>"com.apple.ManagedClient.preferences",
+               "PayloadUUID"=>"73fc30e0-1e57-0131-c32d-000c2944c108",
+               "PayloadVersion"=>1}],
+            "ProfileOrganization"=>"Facebook",
+            "ProfileRemovalDisallowed"=>"false",
+            "ProfileType"=>"Configuration",
+            "ProfileUUID"=>"1781fbec-3325-565f-9022-8aa28135c3cc",
+            "ProfileVerificationState"=>"unsigned",
+            "ProfileVersion"=>1},
+          {"ProfileDisplayName"=>"ScreenSaver Settings",
```
tyler-ball: It looks like the 2 profiles in this list are exactly the same - am I missing something or is that on purpose?

spec/unit/provider/osx_profile_spec.rb
tyler-ball: I would also expect that the `shell_out` code would have an expectation here.

I think there should also be a test `it 'does not install if the profile is already installed'`.  Basically, test the `if` statement thats the first check in `action_install`

spec/unit/provider/osx_profile_spec.rb
```diff
@@ -0,0 +1,251 @@
...
+      @provider.load_current_resource
+      allow(@provider).to receive(:write_profile_to_disk).and_return(profile_path)
+      expect(@provider).to receive(:shell_out).with("profiles -I -F '#{profile_path}'").and_return(shell_out_success)
+      @provider.action_install()
+    end
+
+    it 'should fail if there is no identifier inside the profile' do
+      @test_profile.delete('PayloadIdentifier')
+      @new_resource.profile @test_profile
+      expect{@provider.run_action(:install)}.to raise_error(RuntimeError)
```
tyler-ball: Because `RuntimeError` is so broad I think you should check the error message here - `... raise_error(RuntimeError, /message/)`

tyler-ball:
The resource and provider LGTM but I had some remaining notes about the tests.  Can you also run with [coverage](https://github.com/chef/chef/blob/6d7adf9b39459541250049ae03d9e2c6abe94b90/spec/spec_helper.rb#L38) enabled and check the test coverage?  It looks like it will be light.

spec/unit/provider/osx_profile_spec.rb
natewalck: The identifier is different, which is what matters. I'll update the DisplayName so it is also different (having the same displayname is very confusing).

lib/chef/resource/osx_profile.rb
lamont-granquist: huh?  checksum and version are both set'able by the user on the new_resource there.

if i understand the purpose of this, its more like the enabled attribute of the service resource, which I just cut this one against to remove it in Chef 13:  https://github.com/chef/chef/issues/4204

lamont-granquist: And you should be able to tell if the action runs and the current_resource had nil in the appropriate field that it was not installed.

natewalck:
@tyler-ball Sure thing! I fixed all the tests as requested, doing a full unit test run with COVERAGE=true. I'll push the new diff as soon as that finishes running.

lib/chef/provider/osx_profile.rb
```diff
@@ -0,0 +1,255 @@
...
+      def config_uuid(profile)
+        # Make a UUID of the profile contents and return as string
+        UUIDTools::UUID.sha1_create(
+          UUIDTools::UUID_DNS_NAMESPACE,
+          profile.to_s
+        ).to_s
+      end
+
+      def write_profile_to_disk
+        @new_resource.path(Chef::FileCache.create_cache_path("profiles"))
```
lamont-granquist: the mutating of the new resource here is not good.  you probably want an accessor on the provider that gives you either the `@new_resource.path` or the `Chef::FileCache.create_cache_path("profiles")` as appropriate.

natewalck:
@tyler-ball For the coverage test:

```
lib/chef/provider/osx_profile.rb	83.74 %	255	123	103	20	3.1
lib/chef/resource/osx_profile.rb	100.0 %	84	26	26	0	11.7
``

File: lib/chef/resource/osx_profile.rb (hunk 4 above)
Comment (natewalck) [inline lib/chef/resource/osx_profile.rb]: @lamont-granquist Could you show me an example of what you are looking for? I'm not quite certain what needs to be added.

File: lib/chef/provider/osx_profile.rb [hunk 11]
```diff
@@ -0,0 +1,255 @@
...
+          @new_resource.profile_name
+        end
+
+        current_profile = all_profiles['_computerlevel'].find {
+          |item| item['ProfileIdentifier'] ==
+             @new_profile_identifier
+        }
+        @current_resource.profile(current_profile)
+
+        @current_resource.installed(profile_installed?)
```
lamont-granquist: following up on the other comment in the resource.  this property on the resource needs to be deleted.   and just use `profile_installed?` directly.  you're not even setting the new_resource value so reporting is going to get the wrong values and round tripping state from the provider through the current_resource back to the provider again isn't useful.

lib/chef/resource/osx_profile.rb
lamont-granquist: Just delete this property off the resource.  Delete the line where you set the current_resource.installed in load current resource.  Then every reference to current_resource.installed can just be replaced by a direct call to `#profile_installed?` instead.

lib/chef/provider/osx_profile.rb
natewalck: @lamont-granquist Would this cover it?

```
diff --git a/lib/chef/provider/osx_profile.rb b/lib/chef/provider/osx_profile.rb
index c674875..23fac9e 100644
--- a/lib/chef/provider/osx_profile.rb
+++ b/lib/chef/provider/osx_profile.rb
@@ -59,8 +59,6 @@ class Chef
              @new_profile_identifier
         }
         @current_resource.profile(current_profile)
-
-        @current_resource.installed(profile_installed?)
       end
 
@@ -99,7 +97,7 @@ class Chef
       end
 
       def action_install
-        unless @current_resource.installed
+        unless profile_installed?
           converge_by("install profile #{@new_profile_identifier}") do
             profile_path = write_profile_to_disk
             install_profile(profile_path)
@@ -110,7 +108,7 @@ class Chef
 
       def action_remove
         # Clean up profile after removing it
-        if @current_resource.installed
+        if profile_installed?
           converge_by("remove profile #{@new_profile_identifier}") do
             remove_profile
             get_installed_profiles(update=true)
diff --git a/lib/chef/resource/osx_profile.rb b/lib/chef/resource/osx_profile.rb
index 91883b9..26b834a 100644
--- a/lib/chef/resource/osx_profile.rb
+++ b/lib/chef/resource/osx_profile.rb
@@ -26,8 +26,6 @@ class Chef
 
       identity_attr :profile_name
 
-      state_attrs :installed
-
       default_action :install
       allowed_actions :install, :remove
 
@@ -47,14 +45,6 @@ class Chef
         )
       end
 
-      def installed(arg=nil)
-        set_or_return(
-          :installed,
-          arg,
-        )
-      end
-
       def profile(arg=nil)
         set_or_return(
           :profile,

```

lib/chef/provider/osx_profile.rb
natewalck: I'm not using @new_resource.path anywhere at all. Should I just get rid of it and call Chef::FileCache.create_cache_path("profiles")?

natewalck: Actually this will not work as the resource must have a path for the Chef::FileContentManagement::Tempfile.new function to work as expected.

I don't want people setting the cache path for this...I want it to always be a tempfile made at runtime. Should I not expose path in the resource with a setter to prevent this?

natewalck:
Rebased on master, added fixes @lamont-granquist requested for removing  @current_resource.installed.

lamont-granquist:
:+1:

lib/chef/provider/osx_profile.rb
lamont-granquist: Oh, yeah, I kinda see.  You probably shouldn't be poking around in `Chef::FileContentManagement::Tempfile` directly.

I would suggest using a Chef::Resource::Template directly.  But are you basically working around the fact that templates only pull from cookbook_files and what you want to do is pull from a remote_file URI?  (which is a feature that has been requested before, but we've never gotten around to implementing...)

I think there might be some kind of workaround, though...

lamont-granquist: Yeah what about using a real Template sub-resource and then passing `local true` to the template resource and setting the source to Chef::FileCache.create_cache_path("profiles") or something that rhymes with that?

natewalck: Actually, I only want to pull whole files from the cookbook. I have no desire to use templates (if it needs to be dynamic, it should be a hash stored in an attribute or the like).

I'll try using Chef::Resource::CookbookFile.new like package.rb does and see where that gets me.

natewalck: The line that uses FileContentManagement is for writing the file to disk temporarily so it can be installed.

Basically, we read the profile either from the cookbook or the hash passed to the provider. We create a sha1 of the profile contents and save the resulting UUID out to the hash, then write the hash to disk.

It is then installed from the file that was written out to disk.

Hope that makes sense. The usage of Chef::FileContentManagement::Tempfile.new has nothing to do with writing a file from the cookbook, it is to get a hash into a tempfile that we can install from.

spec/unit/provider/osx_profile_spec.rb
tyler-ball: Thanks!

spec/unit/provider/osx_profile_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+      allow(@provider).to receive(:get_installed_profiles).and_return(@no_profiles)
+      @provider.load_current_resource
+      expect { @provider.run_action(:install) }.to_not raise_error
+    end
+
+    it 'does not install if the profile is already installed' do
+      @new_resource.profile @test_profile
+      allow(@provider).to receive(:get_installed_profiles).and_return(@all_profiles)
+      @provider.load_current_resource
+      expect(@provider.action_install).to_not receive(:install_profile)
```
tyler-ball: You're making sure that `@provider` does not receive `:install_profile`, correct?  If so, I don't think you have this expectation setup correctly.  I believe it needs to look like

```ruby
expect(@provider).to_not receive(:install_profile)
expect { @provider.action_install }.to_not raise_error
```

spec/unit/provider/osx_profile_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+
+require 'spec_helper'
+
+describe Chef::Provider::OsxProfile do
+  let(:shell_out_success) do
+    double('shell_out', :exitstatus => 0, :error? => false)
+  end
+  describe 'action_create' do
+    before(:each) do
+      @node = Chef::Node.new
```
tyler-ball: D'oh.  I should have posted this earlier but just realized.

The desired way of writing RSpec tests is to use `let` instead of making instance variables.  This `before(:each)` block should be rewritten as:

```ruby
let(:node) { Chef::Node.new }
let(:events) { Chef::EventDispatch::Dispatcher.new }
let(:run_context) { Chef::RunContext.new(node, {}, events) }
let(:new_resource) { Chef::Resource::OsxProfile.new("Profile Test", run_context) }
let(:provider) { Chef::Provider::OsxProfile.new(new_resource, run_context) }
let(:all_profiles) do
  {"_computerlevel"=>
      [{"ProfileDisplayName"=>"Finder Settings",
        "ProfileIdentifier"=>"com.apple.finder",
        "ProfileInstallDate"=>"2015-11-08 23:15:21 +0000",
        "ProfileItems"=>
         [{"PayloadContent"=>
            {"PayloadContentManagedPreferences"=>
              {"com.apple.finder"=>
                {"Forced"=>[{"mcx_preference_settings"=>{"ShowExternalHardDrivesOnDesktop"=>false}}]}}},
           "PayloadDisplayName"=>"Custom: (com.apple.finder)",
           "PayloadIdentifier"=>"com.apple.finder",
           "PayloadType"=>"com.apple.ManagedClient.preferences",
           "PayloadUUID"=>"a017048f-684b-4e81-baa3-43afe316d739",
           "PayloadVersion"=>1}],
        "ProfileOrganization"=>"Facebook",
        "ProfileRemovalDisallowed"=>"false",
        "ProfileType"=>"Configuration",
        "ProfileUUID"=>"e2e09bef-e673-44a6-bcbe-ecb5f1c1b740",
        "ProfileVerificationState"=>"unsigned",
        "ProfileVersion"=>1},
      {"ProfileDisplayName"=>"ScreenSaver Settings",
        "ProfileIdentifier"=>"com.testprofile.screensaver",
        "ProfileInstallDate"=>"2015-10-05 23:15:21 +0000",
        "ProfileItems"=>
         [{"PayloadContent"=>
            {"PayloadContentManagedPreferences"=>
              {"com.apple.screensaver"=>
                {"Forced"=>[{"mcx_preference_settings"=>{"idleTime"=>0}}]}}},
           "PayloadDisplayName"=>"Custom: (com.apple.screensaver)",
           "PayloadIdentifier"=>"com.apple.screensaver",
           "PayloadType"=>"com.apple.ManagedClient.preferences",
           "PayloadUUID"=>"73fc30e0-1e57-0131-c32d-000c2944c110",
           "PayloadVersion"=>1}],
        "ProfileOrganization"=>"Facebook",
        "ProfileRemovalDisallowed"=>"false",
        "ProfileType"=>"Configuration",
        "ProfileUUID"=>"1781fbec-3325-565f-9022-8aa28135c3cc",
        "ProfileVerificationState"=>"unsigned",
        "ProfileVersion"=>1}],
  }
end

let(:test_profile) do
  {
  'PayloadIdentifier' => 'com.testprofile.screensaver',
  'PayloadRemovalDisallowed' => false,
  'PayloadScope' => 'System',
  'PayloadType' => 'Configuration',
  'PayloadUUID' => '1781fbec-3325-565f-9022-8aa28135c3cc',
  'PayloadOrganization' => 'Chef',
  'PayloadVersion' => 1,
  'PayloadDisplayName' => 'Screensaver Settings',
  'PayloadContent'=> [
    {
      'PayloadType' => 'com.apple.ManagedClient.preferences',
      'PayloadVersion' => 1,
      'PayloadIdentifier' => 'com.testprofile.screensaver',
      'PayloadUUID' => '73fc30e0-1e57-0131-c32d-000c2944c108',
      'PayloadEnabled' => true,
      'PayloadDisplayName' => 'com.apple.screensaver',
      'PayloadContent' => {
        'com.apple.screensaver' => {
          'Forced' => [
            {
              'mcx_preference_settings' => {
                'idleTime' => 0,
              }
            }
          ]
        }
      }
    }
  ]
  }
end
let(:no_profiles) do
  { "_computerlevel"=> [] }
end

before(:each) do
  allow(provider).to receive(:cookbook_file_available?).and_return(true)
  allow(provider).to receive(:cache_cookbook_profile).and_return('/tmp/test.mobileconfig.remote')
  allow(provider).to receive(:get_new_profile_hash).and_return(test_profile)
  allow(provider).to receive(:get_installed_profiles).and_return(all_profiles)
  allow(provider).to receive(:read_plist).and_return(all_profiles)
  allow(::File).to receive(:unlink).and_return(true)
end
```

Then you can replace all references to `@new_resource` with `new_resource`, `@provider` with `provider`, etc. in the examples and context blocks below.

tyler-ball:
Some refactoring for readability in the tests, but otherwise I'm :+1:

natewalck:
Finished the requested rspec refactoring. Let me know if there is anything else!

lib/chef/provider/osx_profile.rb
```diff
@@ -0,0 +1,254 @@
...
+          end
+        end
+      end
+
+      def action_remove
+        # Clean up profile after removing it
+        if profile_installed?
+          converge_by("remove profile #{@new_profile_identifier}") do
+            remove_profile
+            get_installed_profiles(update=true)
```
jaymzh: This is a bug. You're passing in the result of `update = true` - ruby doesn't have the same kwargs that python does. This technically works only because the result of an assignment is that assignment:

```
irb(main):016:0> puts f='one'
one
```

But kwargs don't work:

```
irb(main):012:0> def foo(bar=nil, baz=nil)
irb(main):013:1> puts baz
irb(main):014:1> end
=> :foo
irb(main):015:0> foo(baz='one')
```

spec/unit/resource/osx_profile_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'spec_helper'
+
+describe Chef::Resource::OsxProfile do
+  before(:each) do
+    @resource = Chef::Resource::OsxProfile.new("Test Profile Resource")
```
tyler-ball: Same note here with `let` - rather than doing this in a `before` block, simple have `let(:resource) { Chef::Resource::OsxProfile.new("Test Profile Resource") }`

tyler-ball:
@natewalck Something useful for reviewers is to wait to squash the commits until the review is done and ready for merge.  Especially with a big review, it makes it a lot easier to see which individual comments are addressed per commit.  Squashing is great, it should just be done as the last step.

natewalck:
@tyler-ball Sure thing! I was rebasing as I went as well, which is probably unnecessary. /ocd

I'll make sure to squash at the end for easier reviews.

natewalck:
Also addressed @jaymzh's comment on kwargs.

Let me know if there is anything else. Thanks again for the feedback!

jaymzh:
:+1:

natewalck:
Added a documentation PR for this over in chef-web-docs: https://github.com/chef/chef-web-docs/pull/773

tyler-ball:
:+1:

natewalck:
Rebased and squashed.

jaymzh:
The failures are all knife related... wat? @tyler-ball  thoughts?

natewalck:
I fixed a typo in the rspec test for osx_profile resource and rebased on master.

I've added documentation as mentioned...is this accepted?

@jaymzh Needs to merge this if there is consensus.

jaymzh:
You have 3 thumbs up which is the requirement, but still have a test failing.

natewalck:
The failing tests do not seem to be mine:

```
rspec ./spec/unit/provider/directory_spec.rb:64 # Chef::Provider::Directory#load_current_resource on unix when the directory exists when the new_resource.group is a string, describes the group as a string
rspec ./spec/unit/provider/package/easy_install_spec.rb:49 # Chef::Provider::Package::EasyInstall easy_install_binary_path should set the current resources package name to the new resources package name
```

I'll rebase again and hope for the best.

natewalck:
Rebase knocked out the EasyInstall fail, but directory is still failing a test.

natewalck:
Woot! Looks like we are passing now. Let me know if you need anything else, @jaymzh.

Thanks all and have a great holidays!

jaymzh:
LGTM. I"ll merge now and add Changelog stuff on Monday.

PR #4079, opened 2015-10-20
add logger to windows service shellout

mwrock:
This adds a logger to shellout when spawning a client run in the windows service so that any debug logging regarding process creation/destruction can be captured. This accompanies but can be deployed independently of https://github.com/chef/mixlib-shellout/pull/110

### DIFF
--- a/lib/chef/application/windows_service.rb
+++ b/lib/chef/application/windows_service.rb
@@ -189,7 +189,11 @@ class Chef
           config_params += " -c #{Chef::Config[:config_file]}" unless  Chef::Config[:config_file].nil?
           config_params += " -L #{Chef::Config[:log_location]}" unless Chef::Config[:log_location] == STDOUT
           # Starts a new process and waits till the process exits
-          result = shell_out("chef-client #{config_params}", :timeout => Chef::Config[:windows_service][:watchdog_timeout])
+          result = shell_out(
+            "chef-client #{config_params}",
+            :timeout => Chef::Config[:windows_service][:watchdog_timeout],
+            :logger => Chef::Log
+          )
           Chef::Log.debug "#{result.stdout}"
           Chef::Log.debug "#{result.stderr}"
         rescue Mixlib::ShellOut::CommandTimeout => e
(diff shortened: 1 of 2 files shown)


ksubrama:
:+1:

lib/chef/application/windows_service.rb
```diff
@@ -189,7 +189,11 @@ def run_chef_client
           config_params += " -c #{Chef::Config[:config_file]}" unless  Chef::Config[:config_file].nil?
           config_params += " -L #{Chef::Config[:log_location]}" unless Chef::Config[:log_location] == STDOUT
           # Starts a new process and waits till the process exits
-          result = shell_out("chef-client #{config_params}", :timeout => Chef::Config[:windows_service][:watchdog_timeout])
+          result = shell_out(
+            "chef-client #{config_params}",
+            :timeout => Chef::Config[:windows_service][:watchdog_timeout],
+            :logger => Chef::Log
```
btm: I think you want `Chef::Log.logger` here.

mwrock: `Chef::Log` is what I tested with and worked. Just looked at mixlib-log again and looks like `logger` returns the internal logging instance which should also work. calling the various `warn`, `debug`, etc calls directly to `Chef::Log` sends those to logger. So either `Log` or `.logger` appear to be effectively the same.

btm: Welp. It looks like most of our use of `Chef::Log.logger` predated multiple logger support.

AFAICT, the only place we let you use the multiple logger feature (without your own tomfoolery) is when we're using a logger on STDOUT instead of a formatter, and you still want a log location. In that case, if you did `Chef::Log.logger.warn("bob")` it would only go to the default logger, which looks like it would be the logfile instead of STDOUT. So you would want `Chef::Log.warn("bob")` there. But that appears to be an edge case for people who've been around a long time and are scared of the output formatters.

So I'd leave it as `Chef::Log`. But, I would be curious if there's another use case for for multiple loggers. @danielsdeleo ?

danielsdeleo: @btm your memory is correct (as best I can remember myself, heh). In the olden times, users had a bad time when they set the log location in `client.rb` and then logged in to run chef-client manually for debugging or whatever, because the output would go only to the specified logfile and they wouldn't see anything. So we added a feature to make chef-client also log to STDOUT if it detected a TTY. Nowadays this would be less of a problem because we have the formatters, too, but it could be an issue if you needed to get the debug logs.

btm:
:+1:

PR #4080, opened 2015-10-20
Make property modules possible

jkeiser:
It'd be nice to be able to write:

```ruby
module SecureThing
  include Chef::Property::Mixin
  property :mode
  property :user
  property :group
end

class File < Chef::Resource
  property :path, name_property: true
  include SecureThing
  property :content, String
end

class Directory < Chef::Resource
  property :path, name_property: true
  include SecureThing
  property :content, String
end
```

Also, I'd love to separate out this functionality so we can use it in things that aren't resources.


lamont-granquist:
:+1:

PR #4081, opened 2015-10-20
RFC-037:  add chef_version and ohai_version metadata

lamont-granquist:
This allows knife to upload cookbooks containing chef_version and ohai_version.  When the chef-client downloads cookbooks with constraints it checks them and throws an exception if the constraints do not match.  The knife command will upload cookbooks which do not match the version constraint (this is deliberate in the case where knife workstation versions differ from chef-client production versions -- injecting the production chef-client version into knife to prevent uploads is out-of-scope).

Also out of scope - including these into /universe, teaching berkshelf and the policyfile depsolver to avoid them, injecting production chef-version values into policyfiles, teaching the cookbook_versions endpoint to filter based on chef_version, etc, etc, etc, etc, etc....  This is only step number one.

This PR does go slightly beyond the RFC and includes the ability to not only pin via a simple constraint like '~> 12.3' but also to allow multiple pins to handle features that were back-ported:

```ruby
chef_version ">= 11.20.0", "< 12.0"
chef_version ">= 12.6.0"
```

That will be satisfied by a chef version that is 11.x >= 11.20.0, or 12.x >= 12.6.0, or >= 13.

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -55,19 +55,23 @@ class Chef
       SOURCE_URL             = 'source_url'.freeze
       ISSUES_URL             = 'issues_url'.freeze
       PRIVACY                = 'privacy'.freeze
+      CHEF_VERSIONS          = 'chef_versions'.freeze
+      OHAI_VERSIONS          = 'ohai_versions'.freeze
 
       COMPARISON_FIELDS = [ :name, :description, :long_description, :maintainer,
                             :maintainer_email, :license, :platforms, :dependencies,
                             :recommendations, :suggestions, :conflicting, :providing,
                             :replacing, :attributes, :groupings, :recipes, :version,
-                            :source_url, :issues_url, :privacy ]
+                            :source_url, :issues_url, :privacy, :chef_versions, :ohai_versions ]
 
-      VERSION_CONSTRAINTS = {:depends     => DEPENDENCIES,
-                             :recommends  => RECOMMENDATIONS,
-                             :suggests    => SUGGESTIONS,
-                             :conflicts   => CONFLICTING,
-                             :provides    => PROVIDING,
-                             :replaces    => REPLACING }
+      VERSION_CONSTRAINTS = {:depends      => DEPENDENCIES,
+                             :recommends   => RECOMMENDATIONS,
+                             :suggests     => SUGGESTIONS,
+                             :conflicts    => CONFLICTING,
+                             :provides     => PROVIDING,
+                             :replaces     => REPLACING,
+                             :chef_version => CHEF_VERSIONS,
+                             :ohai_version => OHAI_VERSIONS }
 
       include Chef::Mixin::ParamsValidate
       include Chef::Mixin::FromFile
@@ -83,6 +87,8 @@ class Chef
       attr_reader :groupings
       attr_reader :recipes
       attr_reader :version
+      attr_reader :chef_versions
+      attr_reader :ohai_versions
 
       # Builds a new Chef::Cookbook::Metadata object.
       #
@@ -118,6 +124,8 @@ class Chef
         @source_url = ''
         @issues_url = ''
         @privacy = false
+        @chef_versions = []
+        @ohai_versions = []
 
         @errors = []
       end
@@ -386,6 +394,18 @@ class Chef
         @replacing[cookbook]
       end
 
+      # FIXME
+      def chef_version(*version_args)
+        @chef_versions << Gem::Dependency.new('chef', *version_args)
+        @chef_versions
+      end
+
+      # FIXME
+      def ohai_version(*version_args)
+        @ohai_versions << Gem::Dependency.new('ohai', *version_args)
+        @ohai_versions
+      end
+
       # Adds a description for a recipe.
       #
       # === Parameters
@@ -481,6 +501,20 @@ class Chef
         @groupings[name]
       end
 
+      def gem_requirements_to_hash(*deps)
+        deps.map do |dep|
+          dep.requirement.requirements.map do |op, version|
+            "#{op} #{version}"
+          end.sort
+        end
+      end
+
+      def gem_requirements_from_hash(what, hash)
+        hash.map do |dep|
+          Gem::Dependency.new(what, *dep)
+        end
+      end
+
       def to_hash
         {
           NAME                   => self.name,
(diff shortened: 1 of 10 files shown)


lamont-granquist:
Tested on hosted and this works as designed.

- knife upload with new code works fine even if the version pin doesn't match
- chef-client run with new code throws the exception
- chef-client run with old code ignores the directive when its present on the server
- knife upload with old code throws an exception

Because of that latter point, we still need to do this:

```ruby
chef_version ">= 12.0" if respond_to?(:chef_version)
```

So that old versions of knife can upload the cookbook.

We might want to change knife so that unknown metadata is ignored (which means a method_missing that warns and then no-ops?  *mild sigh*...), but that's out of scope for this.

lamont-granquist:

lib/chef/cookbook/metadata.rb
```diff
@@ -612,8 +650,28 @@ def privacy(arg=nil)
...
+        unless gem_dep_matches?("chef", Gem::Version.new(Chef::VERSION), *chef_versions)
+          raise Exceptions::CookbookChefVersionMismatch.new(Chef::VERSION, name, version, *chef_versions)
+        end
+      end
+
     private
 
+      def gem_dep_matches?(what, version, *deps)
+        return true unless deps.length > 0
+        deps.each do |dep|
```
coderanger: I don't think this is the logic we want. What you have here is equiv to `deps.any? {|dep| dep.match?(what, version) }`. I think we want `all?` there.

spec/unit/cookbook/metadata_spec.rb
```diff
@@ -684,9 +788,14 @@
...
         :display_name => "You have nothing"
       metadata.version "1.2.3"
+      metadata.chef_version ">= 11.14.2", "< 11.18.10"
+      metadata.chef_version ">= 12.2.1", "< 12.5.1"
+      metadata.ohai_version ">= 7.1.0", "< 7.5.0"
+      metadata.ohai_version ">= 8.0.1", "< 8.6.0"
     end
 
     it "should produce the same output from to_json and Chef::JSONCompat" do
+      # XXX: fairly certain this is testing ruby method dispatch
```
danielsdeleo: Yeah, what I think you want here is to prove that you're meeting some definition of API stability that you're willing to sign up for. You can probably do that by parsing `metadata.to_json` to a Hash and then checking that to whatever level of exactness you can pull off.

lib/chef/cookbook/metadata.rb
coderanger: Ahh, I think I see how this works. Each "row" is loaded into a single dependency object so that covers the `all?`, and then we want an `any?` between each of those? That seems maybe a bit overcomplicated but I can dig it. I would still rewrite this as an `any?` for clarity.

lamont-granquist: Heh, I think this was the last bit of code I wrote before demoing it on the last day of the summit.  Its very obviously a reimplementation of #any? so I can certainly fix that...

spec/unit/cookbook/metadata_spec.rb
lamont-granquist: Yeah, but that's done in the later test which basically checks that the 'almost-round-trip' of `metadata.to_json.to_hash` is stable and that that the round trip of `metadata.from_json(metadata.to_json)` is stable.  I'm tempted to just delete this test, but wasn't positive it had no value, but wanted to document my doubts about it...

lib/chef/cookbook/metadata.rb
lamont-granquist: And yeah, this is the code that handles the "we backported a feature" use case.

spec/unit/cookbook/metadata_spec.rb
danielsdeleo: FWIW, the test comes from https://github.com/chef/chef/commit/1c7b4629240c49e02c482da6de1a6a2cc574304e

lib/chef/cookbook/metadata.rb
```diff
@@ -386,6 +394,18 @@ def replaces(cookbook, *version_args)
         @replacing[cookbook]
       end
 
+      # FIXME
```
danielsdeleo: Is this to remind yourself to add docs?

lamont-granquist: lol yes, i believe it is.

danielsdeleo:
Code and tests look good. Need to figure out the travis issue and add the docs you meant to write, then :+1:

lamont-granquist:
travis was timing out on something unrelated, rebooting the test fixed it.

lamont-granquist:
Added docs, discovered that the ...to_hash and ...from_hash methods were acting on Arrays in the process so cosmetically fixed those names...

lib/chef/cookbook/metadata.rb
```diff
@@ -386,6 +397,28 @@ def replaces(cookbook, *version_args)
...
       end
 
+      # Metadata DSL to set a valid chef_version.  May be declared multiple times
+      # with the result being 'OR'd such that if any statements match, the version
+      # is considered supported.  Uses Gem::Requirement for its implementation.
+      #
+      # @param version_args [Array<String>] Version constraint in String form
+      # @return [Array<Gem::Dependency>] Current chef_versions array
+      def chef_version(*version_args)
+        @chef_versions << Gem::Dependency.new('chef', *version_args)
```
coderanger: The fact that this returns a value makes it sound like you expect it to be used as a reader too? Calling `chef_versions()` would shift on an empty dependency, maybe add `unless version_args.empty?` here?

lamont-granquist: fixed

mcquin:
:+1:

jonlives:
:+1: Thanks @lamont-granquist!

btm:
What about chef-shell? People still use that, right? Seems like we should validate there too so you get the same behavior.

lamont-granquist:
Hmmm?  The point of testing solo is because there's a solo-specific block in the `policy_builder/expand_node_object.rb` file so we need to test both sides of that conditional.  With chef-shell you'll wind up on one side or the other of that expression based on how you invoked chef-shell so we've got coverage there and it'll work.

lamont-granquist:
Turns out I did an integration test for the solo-side of that conditional, but forgot the client-side.

thommay:
:+1:

lamont-granquist:
I'll push the CHANGELOG after meeting....

PR #4091, opened 2015-10-22
Allow downloading of root_files in a chef repository

stevendanna:
A FileSystemEntry with a nil parent is a "root directory" and thus
should exists so long as its filesystem path exists. This was preventing
the download of files such as invitation.json whose parent directory is
a root directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,10 @@
 
 * [pr#4034](https://github.com/chef/chef/pull/4034) add optional ruby-profiling with --profile-ruby
 * [pr#3119](https://github.com/chef/chef/pull/3119) allow removing user, even if their GID isn't resolvable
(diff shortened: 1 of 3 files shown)


stevendanna:
This is currently breaking `knife-ec-backup`.  I'm not particularly thrilled with this solution, but I've exhausted the time I can put into it at the moment. My "not thrilled"-ness comes from the fact that it seems like we should always be representing the root chef repository dir with ChefRepositoryFileSystemRootDir or FileSystemRootDir objects.  I also think it would make more sense if we defined the root dir as being its own parent which would save us from carting around nils.

stevendanna:
cc @chef/lob @jkeiser @chef/maintainers

danielsdeleo:
LGTM :+1:

btm:
:+1:

PR #4098, opened 2015-10-23
provider/user/dscl: Set "comment" default value

lamont-granquist:
rebase and md files for #3193 

closes #3193

### DIFF
--- a/lib/chef/provider/user/dscl.rb
+++ b/lib/chef/provider/user/dscl.rb
@@ -195,9 +195,10 @@ user password using shadow hash.")
 
         #
         # Saves the specified Chef user `comment` into RealName attribute
-        # of Mac user.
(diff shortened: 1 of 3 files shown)


lib/chef/provider/user/dscl.rb
```diff
@@ -195,9 +195,10 @@ def dscl_create_user
 
         #
         # Saves the specified Chef user `comment` into RealName attribute
-        # of Mac user.
+        # of Mac user. If `comment` is not specified, it takes `username` value.
         #
         def dscl_create_comment
+          @new_resource.comment(@new_resource.username) if @new_resource.comment.nil?
```
coderanger: Mutating the resource like this is kind of weird. Very unlikely to cause issues, but seems non-standard.

legal90:
Thanks!

PR #4101, opened 2015-10-23
package/solaris: Adds alias method for upgrade_package.

johnbellone:
This follows the same convention that the [AIX package provider](https://github.com/johnbellone/chef/blob/master/lib/chef/provider/package/aix.rb#L122) has done and adds an alias_method.

/cc @pburkholder @acaiafa

### DIFF
--- a/lib/chef/provider/package/solaris.rb
+++ b/lib/chef/provider/package/solaris.rb
@@ -125,6 +125,8 @@ class Chef
           end
         end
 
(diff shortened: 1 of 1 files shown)


thommay:
:+1:

lamont-granquist:
This is fine :+1: 

We need to fix load_current_resource since I suspect that you're seeing that `:install` with this provider behaves like `:upgrade` anyway...

Issue #3570, opened 2015-06-23
zypper cannot install multiple packages


furlongm:
The following snippet fails on opensuse:

```
    package ['vim', 'vim-data'] do
      action :install
    end
```

Here is the failure:

```
Recipe: vim::default
  * zypper_package[vim, vim-data] action install
    * No candidate version available for vim-data
    ================================================================================
    Error executing action `install` on resource 'zypper_package[vim, vim-data]'
    ================================================================================

    Chef::Exceptions::Package
    -------------------------
    No candidate version available for vim-data

    Resource Declaration:
    ---------------------
    # In /srv/chef/cookbooks/vim/recipes/default.rb

     18:     package ['vim', 'vim-data'] do
     19:       action :install
     20:     end
     21:     cookbook_file 'vimrc.Suse' do

    Compiled Resource:
    ------------------
    # Declared in /srv/chef/cookbooks/vim/recipes/default.rb:18:in `from_file'

    zypper_package("vim, vim-data") do
      action :install
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      package_name ["vim", "vim-data"]
      declared_type :package
      cookbook_name :vim
      recipe_name "default"
    end


Running handlers:
[2015-06-23T09:34:41+00:00] ERROR: Running exception handlers
Running handlers complete
[2015-06-23T09:34:41+00:00] ERROR: Exception handlers complete
Chef Client failed. 0 resources updated in 17.682571971 seconds
[2015-06-23T09:34:42+00:00] FATAL: Stacktrace dumped to /srv/chef/chef-stacktrace.out
[2015-06-23T09:34:42+00:00] ERROR: Found 1 errors, they are stored in the backtrace
[2015-06-23T09:34:42+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

I've tried with 12.3.0 and 12.4.0rc2.

furlongm:
Both packages are available and are actually already installed:

```
# zypper se vim
Loading repository data...
Reading installed packages...

S | Name                      | Summary                                                             | Type
--+---------------------------+---------------------------------------------------------------------+--------

[SNIP]

i | vim                       | Vi IMproved                                                         | package
i | vim-data                  | Vi IMproved                                                         | package
```

I'm able to reproduce with any array of packages, not just the above. It's always the second package that is reported as having no candidate version. It doesn't matter if that package is installed or not.

[...]

PR #4102, opened 2015-10-24
Support for Zypper multiple package install

gsaslis:
See #3570 for details

### DIFF
--- a/lib/chef/provider/package/zypper.rb
+++ b/lib/chef/provider/package/zypper.rb
@@ -33,46 +33,80 @@ class Chef
         provides :zypper_package, os: "linux"
 
         def load_current_resource
-          @current_resource = Chef::Resource::ZypperPackage.new(new_resource.name)
-          current_resource.package_name(new_resource.package_name)
+          @current_resource = Chef::Resource::ZypperPackage.new(@new_resource.name)
+          @current_resource.package_name(@new_resource.package_name)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I'd prefer to not see another package provider written in this 'style'.   In #4106 I'm working on moving more of the logic into the base provider, cleanly, and not having the mess with arrays-vs-not-arrays polluting every single package provider.  I'll try to get that one into shape to be reviewed.

gsaslis:
ok, not sure what you have in mind exactly, but it would definitely make sense to treat this type of feature on the `package` provider level, rather than an apt / yum specific feature. 

In fact, given the fact that the multi-package feature was only very recently introduced, I assumed this was already the case and that the common interface would be the array of package names (and versions) passed in, and it would then be the job of each of the providers to make sure it is correctly handled. 
In fact, this is how it is handled in apt & yum (if I'm not mistaken), so I just followed the existing pattern.  

In any case, if you see this as something that's not high in your priority list, I believe this PR would very well do for the meantime.. 

P.S: Thanks a lot for reviewing!!  ; )

lamont-granquist:
I'm actually going to start working on making this easier.

Got distracted today because I'm working on `chocolatey_package` and I need to be able to easily do windows builds and was fixing my tooling.

lamont-granquist:
accepted minor for now, it'll get closed when the other PR is merged.

PR #4106, opened 2015-10-26
add chocolatey_package to core chef

lamont-granquist:
* Adds `chocolatey_package` as a core-chef resource
* Does not conflict with the `chocolatey` provider in the cookbook
* Uses the semantics of the core-chef package resource instead of chocolately (:remove instead of :uninstall)
* Implements multipackage install/remove/upgrade to the extent possible

Bonus is that it adds features to the core package provider to make it much easier to write multipackage-enabled subclasses.

One difference between this provider and the chocolatey provider in the cookbook is that we only take a string of options not a hash, again for consistency with the base package provider code in core chef.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -1,0 +1,230 @@
+#
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider/package'
+require 'chef/resource/chocolatey_package'
+require 'chef/mixin/powershell_out'
+
+class Chef
+  class Provider
+    class Package
+      class Chocolatey < Chef::Provider::Package
+        include Chef::Mixin::PowershellOut
+
+        provides :chocolatey_package, os: "windows"
+
+        # Declare that our arguments should be arrays
+        use_multipackage_api
+
+        # Responsible for building the current_resource.
+        #
+        # @return [Chef::Resource::ChocolateyPackage] the current_resource
+        def load_current_resource
+          @current_resource = Chef::Resource::ChocolateyPackage.new(new_resource.name)
+          current_resource.package_name(new_resource.package_name)
+          current_resource.version(build_current_versions)
+          current_resource
+        end
+
+        # Lazy initializer for candidate_version.  A nil value means that there is no candidate
+        # version and the package is not installable (generally an error).
+        #
+        # @return [Array] list of candidate_versions indexed same as new_resource.package_name/version
+        def candidate_version
+          @candidate_version ||= build_candidate_versions
+        end
+
+        # Install multiple packages via choco.exe
+        #
+        # @param names [Array<String>] array of package names to install
+        # @param versions [Array<String>] array of versions to install
+        def install_package(names, versions)
+          name_versions_to_install = desired_name_versions.select { |n, v| names.include?(n) }
+
+          name_nil_versions = name_versions_to_install.select { |n,v| v.nil? }
+          name_has_versions = name_versions_to_install.reject { |n,v| v.nil? }
+
+          # choco does not support installing multiple packages with version pins
+          name_has_versions.each do |name, version|
+            choco_command("install -y -version", version, cmd_args, name)
+          end
+
+          # but we can do all the ones without version pins at once
+          unless name_nil_versions.empty?
+            cmd_names = name_nil_versions.keys
+            choco_command("install -y", cmd_args, *cmd_names)
+          end
+        end
+
+        # Upgrade multiple packages via choco.exe
+        #
+        # @param names [Array<String>] array of package names to install
+        # @param versions [Array<String>] array of versions to install
+        def upgrade_package(names, versions)
+          name_versions_to_install = desired_name_versions.select { |n, v| names.include?(n) }
+
+          name_nil_versions = name_versions_to_install.select { |n,v| v.nil? }
+          name_has_versions = name_versions_to_install.reject { |n,v| v.nil? }
+
+          # choco does not support installing multiple packages with version pins
+          name_has_versions.each do |name, version|
+            choco_command("upgrade -y -version", version, cmd_args, name)
+          end
+
+          # but we can do all the ones without version pins at once
+          unless name_nil_versions.empty?
+            cmd_names = name_nil_versions.keys
+            choco_command("upgrade -y", cmd_args, *cmd_names)
+          end
+        end
+
+        # Remove multiple packages via choco.exe
+        #
+        # @param names [Array<String>] array of package names to install
+        # @param versions [Array<String>] array of versions to install
+        def remove_package(names, versions)
+          choco_command("uninstall -y", cmd_args, *names)
+        end
+
+        # Support :uninstall as an action in order for users to easily convert
+        # from the `chocolatey` provider in the cookbook.  It is, however,
+        # already deprecated.
+        def action_uninstall
+          Chef::Log.deprecation "The use of action :uninstall on the chocolatey_package provider is deprecated, please use :remove"
+          action_remove
+        end
+
+        # Choco does not have dpkg's distinction between purge and remove
+        alias_method :purge_package, :remove_package
+
+        # Override the superclass check.  The semantics for our new_resource.source is not files to
+        # install from, but like the rubygem provider's sources which are more like repos.
+        def check_resource_semantics!
+        end
+
+        private
+
+        # Magic to find where chocolatey is installed in the system, and to
+        # return the full path of choco.exe
+        #
+        # @return [String] full path of choco.exe
+        def choco_exe
+          @choco_exe ||=
+            ::File.join(
+              powershell_out!(
+                "[System.Environment]::GetEnvironmentVariable('ChocolateyInstall', 'MACHINE')"
+              ).stdout.chomp,
+              'bin',
+              'choco.exe'
+          )
+        end
+
+        # Helper to dispatch a choco command through shell_out using the timeout
+        # set on the new resource, with nice command formatting.
+        #
+        # @param args [String] variable number of string arguments
+        # @return [Mixlib::ShellOut] object returned from shell_out!
+        def choco_command(*args)
+          shell_out_with_timeout!(args_to_string(choco_exe, *args))
+        end
+
+        # Use the available_packages Hash helper to create an array suitable for
+        # using in candidate_version
+        #
+        # @return [Array] list of candidate_version, same index as new_resource.package_name/version
+        def build_candidate_versions
+          new_resource.package_name.map do |package_name|
+            available_packages[package_name.downcase]
+          end
+        end
+
+        # Use the installed_packages Hash helper to create an array suitable for
+        # using in current_resource.version
+        #
+        # @return [Array] list of candidate_version, same index as new_resource.package_name/version
+        def build_current_versions
+          new_resource.package_name.map do |package_name|
+            installed_packages[package_name.downcase]
+          end
+        end
+
+        # Helper to construct Hash of names-to-versions, requested on the new_resource.
+        # If new_resource.version is nil, then all values will be nil.
+        #
+        # @return [Hash] Mapping of requested names to versions
+        def desired_name_versions
+          desired_versions = new_resource.version || new_resource.package_name.map { nil }
+          Hash[*new_resource.package_name.zip(desired_versions).flatten]
+        end
+
+        # Helper to construct optional args out of new_resource
+        #
+        # @return [String] options from new_resource or empty string
+        def cmd_args
+          cmd_args = [ new_resource.options ]
+          cmd_args.push( "-source #{new_resource.source}" ) if new_resource.source
+          args_to_string(*cmd_args)
+        end
+
+        # Helper to nicely convert variable string args into a single command line.  It
+        # will compact nulls or empty strings and join arguments with single spaces, without
+        # introducing any double-spaces for missing args.
+        #
+        # @param args [String] variable number of string arguments
+        # @return [String] nicely concatenated string or empty string
+        def args_to_string(*args)
+          args.reject {|i| i.nil? || i == "" }.join(" ")
+        end
+
+        # Available packages in chocolatey as a Hash of names mapped to versions
+        # (names are downcased for case-insensitive matching)
+        #
+        # @return [Hash] name-to-version mapping of available packages
+        def available_packages
+          @available_packages ||=
+            begin
+              cmd = [ "list -r #{package_name_array.join ' '}" ]
+              cmd.push( "-source #{new_resource.source}" ) if new_resource.source
+              parse_list_output(*cmd)
+            end
+        end
+
+        # Installed packages in chocolatey as a Hash of names mapped to versions
+        # (names are downcased for case-insensitive matching)
+        #
+        # @return [Hash] name-to-version mapping of installed packages
+        def installed_packages
+          @installed_packages ||= parse_list_output("list -l -r")
+        end
+
+        # Helper to convert choco.exe list output to a Hash
+        # (names are downcased for case-insenstive matching)
+        #
+        # @param cmd [String] command to run
+        # @return [String] list output converted to ruby Hash
+        def parse_list_output(*args)
+          hash = {}
+          choco_command(*args).stdout.each_line do |line|
+            name, version = line.split('|')
+            hash[name.downcase] = version.chomp
+          end
+          hash
+        end
+      end
+    end
+  end
+end
--- a/lib/chef/resource/chocolatey_package.rb
+++ b/lib/chef/resource/chocolatey_package.rb
@@ -1,0 +1,37 @@
+#
+# Author:: Adam Jacob (<adam@opscode.com>)
+# Copyright:: Copyright (c) 2008-2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/resource/package'
+
+class Chef
+  class Resource
+    class ChocolateyPackage < Chef::Resource::Package
+
+      provides :chocolatey_package, os: "windows"
+
+      def initialize(name, run_context=nil)
+        super
+        @resource_name = :chocolatey_package
+      end
+
+      property :package_name, [String, Array], coerce: proc { |x| [x].flatten }
+
+      property :version, [String, Array], coerce: proc { |x| [x].flatten }
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


mwrock:
Awesome to see this. I think you want to move the provider under `providers/windows` alongside of `MSI` and add logic to https://github.com/chef/chef/blob/master/lib/chef/provider/package/windows.rb#L66 to determine when to use chocolatey.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,190 @@
...
+#
+
+require 'chef/provider/package'
+require 'chef/resource/chocolatey_package'
+require 'chef/mixin/powershell_out'
+
+class Chef
+  class Provider
+    class Package
+      class Chocolatey < Chef::Provider::Package
```
mwrock: Should be nested in the `windows` provider class

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,190 @@
...
+        provides :chocolatey_package, os: "windows"
+
+        # Declare that our arguments should be arrays
+        package_class_supports_arrays
+
+        # Responsible for building the current_resource and as a (necessary)
+        # side effect this populates the candidate_version Array.
+        #
+        # @return [Chef::Resource::ChocolateyPackage] the current_resource
+        def load_current_resource
```
btm: load_current_resource isn't going to run until converge, in Chef::Provider#run_action, right? Can we check if chocolately is installed here and raise if it isn't?

Is it "necessary" to populate candidate_version here because we're going to run #define_resource_requirements next where we make sure there are candidates for for "all" packages?

lib/chef/resource/chocolatey_package.rb
```diff
@@ -0,0 +1,34 @@
+#
```
mwrock: Not sure if this file is necessary at all. I believe if we derrive from windows_package, this can go.

mwrock:
One thing we will definitely want to do either in this PR or a subsequent iteration is support the `source` attribute. Many host their chocolatey feeds either on a network share or some other privately hosted URL. This is key for enterprise scenarios.

lamont-granquist:
Does the existing chocolatey provider support `source` or similar functionality?  I mostly copied what I found there.

lib/chef/resource/chocolatey_package.rb
lamont-granquist: We need to set the resource name and need to declare our `provides` and this provider is written more inline with the rest of the package providers, not like windows_package.

lib/chef/provider/package/chocolatey.rb
lamont-granquist: Yeah we should probably do a check for choco.exe existence.  Its really fussy getting that right, though.

I feel its now necessary to to populate candidate_version in load_current_resource for inspecting with `current_value`.  I'm having some active discussion with @jkeiser about what to do about 'derived desired state' like this and playing around with ways to get at provider state off of `current_value`.  Currently, we don't call `define_resource_requirements` or `process_resource_requirements` in `current_value`, though, so the right place to set `candidate_version` seems to be here.

lib/chef/provider/package/chocolatey.rb
lamont-granquist: Nope, this is the correct class heirarchy for package resources, the windows_package provider is in there as well:

```ruby
class Chef
  class Provider
    class Package
      class Windows < Chef::Provider::Package
```

lamont-granquist: Also same comment that the provider should not go into `provider/windows` and should live in `provider/packages` like the rest of them.

lamont-granquist:
@mwrock this is not written as a `windows_package` `installer_type :chocolatey` kind of package, it is its own `chocolatey_package` resource on the DSL

the reason for that is that the pattern for the core chef packages now works pretty well -- there's 1:1 mappings from resource-to-provider for yum_package, apt_package, dpkg_package, rpm_package, etc.  we no longer have the situation where you get a generic package resource wrapping a yum_package provider and the yum-specific validation is broken.  or like in the service resource where every time we add a new option to one of the providers then we have to add that option to the service resource itself and now we have options validation issues on every other service provider which will happily accept those options.

the N providers for 1 resource pattern broke pretty badly, and its not sustainable and does not end well and you can't ever get the validation right for it.

so this is written in the 1:1 resource-to-provider model, which is the way we should be using going forwards.  the windows_package installer_type is yet another bad example of that pattern, which if anything we need to pull out resources like `msi_package` rather than continue trying to make one fascade try to do even more work.

mwrock:
@lamont-granquist thanks for the explanation on why we are not going `windows_package` style.

Re `source`, the existing provider does support it: https://github.com/chocolatey/chocolatey-cookbook/blob/master/providers/default.rb#L73

lamont-granquist:
interesting.  we've got the ability to inject arbitrary args via `options` and i think i dropped that to simply the overall shape of the API.

i'll have to check tomorrow to see if choco's semantics for `source` matches the way the other package providers use `source`.  want to make sure we don't wind up with another `attributes` (node) vs `attributes` (resources) semantic mixup...

lib/chef/provider/package/chocolatey.rb
lamont-granquist: fixed this via 'lazying' the initializer.

lamont-granquist:
@chef/client-core @chef/client-windows i think this is ready for final review and shipping.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+require 'chef/resource/chocolatey_package'
+require 'chef/mixin/powershell_out'
+
+class Chef
+  class Provider
+    class Package
+      class Chocolatey < Chef::Provider::Package
+        include Chef::Mixin::PowershellOut
+
+        provides :chocolatey_package, os: "windows"
```
jaymzh: Shouldn't it also provide :package on windows ?

lamont-granquist: No, the windows_package provider already does that, and switching that would be a breaking change, probably needs an RFC and/or needs to be Chef 13, definitely a yak for another day...

jaymzh: Wait, I thought the whole point if this dynamic provider stuff is that more than one provider can support a resource, you can keep windows_package at a higher priority and then I can just call:

```
Chef::Platform::ProviderPriorityuMap.instance.priority :package Chef::PRovider::Package::Chocolaty
```

in my config somewhere to change the default for me....

lamont-granquist: Oh yeah, totally forgot about that...

lamont-granquist: I *think* these days it doesn't really matter, though, and if someone just does one of these in a library file:

```
Chef::Resource::ChocolateyPackage.provides :package, platform: :windows
```

That it'll pick that up, and most-specific-and-last-writer-wins (after john's 12.4.x changes) so I think all the ProviderPriorityMap stuff is now deprecated-ish, or at least unnecessary to get the job done...

mwrock:
cc @ferventcoder @guilhem

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+          desired_versions = new_resource.version || new_resource.package_name.map { nil }
+          Hash[*new_resource.package_name.zip(desired_versions).flatten]
+        end
+
+        # Helper to construct optional args out of new_resource
+        #
+        # @return [String] options from new_resource or empty string
+        def cmd_args
+          cmd_args = [ new_resource.options ]
+          cmd_args.push( "-source #{new_resource.source}" ) if new_resource.source
```
mwrock: :+1:

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+        # @return [Mixlib::ShellOut] object returned from shell_out!
+        def choco_command(*args)
+          shell_out_with_timeout!(args_to_string(choco_exe, *args))
+        end
+
+        # Use the available_packages Hash helper to create an array suitable for
+        # using in candidate_version
+        #
+        # @return [Array] list of candidate_version, same index as new_resource.package_name/version
+        def build_candidate_versions
```
mwrock: Consider tacking a `package_name` parameter and then calling `choco #{package_name} -r` which will be much faster. Otherwise this returns thousands of results.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+        # @return [String] nicely concatenated string or empty string
+        def args_to_string(*args)
+          args.reject {|i| i.nil? || i == "" }.join(" ")
+        end
+
+        # Available packages in chocolatey as a Hash of names mapped to versions
+        # (names are downcased for case-insensitive matching)
+        #
+        # @return [Hash] name-to-version mapping of available packages
+        def available_packages
```
mwrock: Consider taking a `package_name` parameter and then calling `choco #{package_name} -r` which will be much faster. Otherwise this returns thousands of results.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+          shell_out_with_timeout!(args_to_string(choco_exe, *args))
+        end
+
+        # Use the available_packages Hash helper to create an array suitable for
+        # using in candidate_version
+        #
+        # @return [Array] list of candidate_version, same index as new_resource.package_name/version
+        def build_candidate_versions
+          new_resource.package_name.map do |package_name|
+            available_packages[package_name.downcase]
```
mwrock: see comment below on `available_packages`

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,220 @@
...
+          end
+        end
+
+        # Upgrade multiple packages via choco.exe
+        #
+        # @param names [Array<String>] array of package names to install
+        # @param versions [Array<String>] array of versions to install
+        def upgrade_package(names, versions)
+          unless desired_name_versions.select { |n, v| names.include?(n) }.all? { |n,v| v.nil? }
+            raise Chef::Exceptions::Package, "Chocolatey Provider does not support version pins on upgrade command, use install instead"
```
ferventcoder: Chocolatey itself doesn't care. It supports `--version` on upgrade as well as install.

lib/chef/provider/package/chocolatey.rb
ferventcoder: Is this required? Until choco has local pkg indexes, this will be a horrible hit to performance across all default sources (or the specified source). It's not something I'd recommend as enough of these could be a bad hit to chocolatey.org. I'd prefer not to make a call out to pull back in all results.

lib/chef/provider/package/chocolatey.rb
mwrock: In line with @ferventcoder 's comment, rather than using `available_packages` here, maybe simply return the given version (or nil if none given). While we risk bogus versions, this value does not seem to exceed the cost.

mwrock:
Tested this today. It may be tha tchocolatey is haveing perf issues or that I am on the road with flakey internet but the `available_packages` call is consistently dog slow. I played with filtering it to package name which was a big improvement but still took several seconds.

I think we'd be better off not trying to resolve candidate versions against the chocolatey feed. If a user provides a bogus version that is unfortunate but the cost here seems exorbitant.

lib/chef/provider/package/chocolatey.rb
jaymzh: can we like... figure all that out and document the right way of having multiple providers that can support a resource and document it? I feel like we go through "wait, how do we do this, again?" every week or two.

lib/chef/provider/package/chocolatey.rb
ferventcoder: Best case scenario, available_packages takes a package name parameter.

lib/chef/provider/package.rb
```diff
@@ -207,6 +209,15 @@ def action_reconfig
       end
 
       # @todo use composition rather than inheritance
+
+      def with_subclass_array_wrapper(name, version)
+        if supports_arrays?
+          yield [name].flatten, [version].flatten
+        else
+          yield name, version
```
btm: Is this a convenient generic place to raise for resource that don't support arrays when an array is passed? #supports_arrays? is new here, so it'd have to be added to all the other classes. Probably not worth it.

lamont-granquist: Yeah, but we still can't tell yet.  If yum+apt are upgraded to us this (which they should) then we could think about that, but its still going to be a converge-time, not a compile-time raise since its in the provider.

lib/chef/provider/package/chocolatey.rb
lamont-granquist: Yeah, I hope we're done with rewriting all this though, i don't think it'll change again (we also don't have any plans to drop the old way).

lib/chef/provider/package/chocolatey.rb
lamont-granquist: Did that change reasonably recently?  I started writing this about 4+ months ago and I didn't think it did...

lib/chef/provider/package/chocolatey.rb
lamont-granquist: Does it work with multiple package args with -r as well as just one?

lamont-granquist:
I think we're going to find that we need at least a filtered version of available_packages in order to do good idempotency checks.  The check for `upgradable?` in the chocolatey cookbook also seems fairly bad since it just calls upgrade with `--noop` which seems like it'd be equally as expensive, if not worse.

We want to have an idempotency check in order to check the current_resource.version against the candidate_version.  If we set candidate_version to nil then we always upgrade on every CCR and that's a guaranteed bug report against this provider since the resource will never not be updated.

What other providers (yum, apt, gem) do is to get the index and then cache it for the lifetime of the CCR so that future providers don't have to go across the wire, and i'm working on making the gem provider only cache partial results (#3856).

ferventcoder:
> What other providers (yum, apt, gem) do is to get the index and then cache it for the lifetime of the CCR so that future providers don't have to go across the wire

Right, package indexes are what we have yet to put into Choco, but it's on the list and super important to do.

lamont-granquist:
Just did a little testing and `list git -r` produces a lot of output, looks like its searching against much more than just the program name index and is probably pulling out substrings?

I think I'm seeing why the existing cookbook does an `upgrade --noop` but that's kind of a messy solution, it'd be nice to get a better API to query here.

Not sure what to do about this provider in the meantime, I need to think about it a bit...

lib/chef/provider/package/chocolatey.rb
lamont-granquist: @mwrock 'nil' means 'not available' and the superclass throws an exception.   the way the superclass is written it assumes you'll can give it accurate information...

lib/chef/provider/package/chocolatey.rb
ferventcoder: Not that I know of. We've always allowed upgrades to particular versions.

lamont-granquist:
Yeah, I'm a bit stalled on this now.  The Chef::Provider::Package superclass really expects subclasses to behave this way.  We could probably hack it all up, but it'd be a massive violation of Liskov Substitution part of SOLID and while we /could/ do it, we probably /shouldn't/ do it.

mwrock:
Yeah I can see the dilemma. I do think that `list <package> -r` is a step in the right direction but not perfect. The `git` example I think may be more egregious than the average package in terms of number of results returned, but I don't have good data to back that up with. Even `git` returns 274(331ms) results compared to the current 2875 (38570msec) which is a significant (but not ideal) improvement.

I think adding the package filter would be reasonable to release as long as we do not risk taking down @ferventcoder 's chocolatey database with the queries.

ferventcoder:
> I think adding the package filter would be reasonable to release as long as we do not risk taking down @ferventcoder 's chocolatey database with the queries.

Not risk taking down Chocolatey.org with queries. It seems to do pretty well for the most part.

lamont-granquist:
One thing that I'll throw out there is that the /universe endpoint on supermarket.chef.io delivers a lot more than that.   It returns 3.2MB of rendered JSON with 12579 cookbook versions in it.  It can do that because it caches the rendered JSON and mostly just serves that up.  Every 60 seconds or so it'll get hit by a request which will decide the JSON is too stale, and it'll asynchronously start a single thread that does the full table scan of the database and regenerates the JSON--cant remember what we do for the first request after a restart when there's no cached JSON (you can either send back a 4xx retriable code or else block and queue).

And that endpoint gets pounded both by client side use of berkshelf and by berkshelf-api mirrors.

We actually 'should' be supporting an API more like rubygems where bundler will make a query with the list of gems to depsolve against, and rubygems replies with only a slice of the universe.  The API calls that `gem install chef --verbose` makes looks like:

```
HEAD https://api.rubygems.org/api/v1/dependencies
200 OK
GET https://api.rubygems.org/api/v1/dependencies?gems=chef
200 OK
GET https://api.rubygems.org/api/v1/dependencies?gems=chef-config,chef-zero,diff-lcs,erubis,ffi-yajl,highline,mixlib-authentication,mixlib-cli,mixlib-log,mixlib-shellout,net-ssh,net-ssh-multi,ohai,plist,pry,rspec-core,rspec-expectations,rspec-mocks,rspec_junit_formatter,serverspec,specinfra,syslog-logger
200 OK
GET https://api.rubygems.org/api/v1/dependencies?gems=builder,coderay,ffi,hashie,ipaddress,libyajl2,method_source,mime-types,mixlib-config,multi_json,net-scp,net-ssh-gateway,net-telnet,rack,rake,rspec,rspec-its,rspec-support,sfl,slop,systemu,uuidtools,wmi-lite
200 OK
```

That's closer to the 'right' way to do this kind of thing, but we haven't had the need (yet) to start filtering /universe results.

ferventcoder:
> One thing that I'll throw out there is that the /universe endpoint on supermarket.chef.io delivers a lot more than that. It returns 3.2MB of rendered JSON with 12579 cookbook versions in it. It can do that because it caches the rendered JSON and mostly just serves that up. Every 60 seconds or so it'll get hit by a request which will decide the JSON is too stale, and it'll asynchronously start a single thread that does the full table scan of the database and regenerates the JSON (meanwhile threads continue to serve the old JSON until the new JSON is available)--cant remember what we do for the first request after a restart when there's no cached JSON (you can either send back a 4xx retriable code or else block and queue).

Yep, I've started down a similar path with caching. I inherited the inner workings of the Chocolatey Gallery (Chocolatey.org) by starting with the NuGet Gallery and working from there. It wasn't the most optimized thing at the time I forked the codebase and while I've made things better it is still not complete yet. We are doing a total rewrite of the web site, although I had to put bits and pieces over into it over the month of October because traffic was very high, plus the database had some very fragmented indexes that caused everything to get worse.  I wrote it all up in our recent [newsletter](http://us8.campaign-archive2.com/?u=86a6d80146a0da7f2223712e4&id=0165685e1f).

lamont-granquist:
Heh, cool, victim of your own success.

So I think I'm going to put this on pause until chocolatey.org get the API+caching improvements, and I'll cherry-pick the commit that adds better multipackage support into core and try building out the Zypper provider on top of that one...

ferventcoder:
> So I think I'm going to put this on pause until chocolatey.org get the API+caching improvements

This is already done. Now we are just fixing up small things related to the changes.

lamont-granquist:
So what API should I use?  The whole list or the pruned down list?

mwrock:
Pruned.

lamont-granquist:
okay probably need to give this a spin, should be using the pruned down list -r

lamont-granquist:
@chef/client-core this is ready for review

lib/chef/provider/package/chocolatey.rb
```diff
@@ -0,0 +1,230 @@
...
+          choco_command("uninstall -y", cmd_args, *names)
+        end
+
+        # Support :uninstall as an action in order for users to easily convert
+        # from the `chocolatey` provider in the cookbook.  It is, however,
+        # already deprecated.
+        def action_uninstall
+          Chef::Log.deprecation "The use of action :uninstall on the chocolatey_package provider is deprecated, please use :remove"
+          action_remove
+        end
```
smurawski: `action_uninstall` never gets called when the action is set to :uninstall.  Action validation returns:
```
FATAL: Chef::Exceptions::ValidationFailed: Option action must be equal to one of: nothing, install, upgrade, remove, purge, reconfig!  You passed :uninstall.
```

smurawski:
Other than that :+1:

mwrock:
:+1:

jamescott:
Here's the draft reference page for 12.7: https://docs.chef.io/release/12-7/resource_chocolatey_package.html

lamont-granquist:
@jamescott looks like you mostly just need to add the remaining actions, and add some multipackage examples.

PR #4109, opened 2015-10-26
Run Chef unit tests in delivery.

tylercloke:


coderanger:
As mentioned on the mailing list :-1: :-1: :-1: :-1: :-1: :-1: :-1: until there is some kind of story for how non-employees interact with Delivery.

PR #4111, opened 2015-10-27
launchd for osx

mikedodge04:
@jaymzh will merge under his CCLA once it's passed review by @chef / @client-core

### launchd
This resource installs/removes Launch Daemons and Agents

#### Actions
- :create
- :create_if_missing
- :delete

#### Parameter attributes:
- `label` - This required key uniquely identifies the job to launchd.
- `path` - This is used as an override. By default the provider will generate a
path for you based on service_type. Default: /Library/LaunchDaemons/#{label}
- `hash` - key, value pairs to create launchd plist.
- `source` - Source path to launchd plist.
- `cookbook` - Cookbook which source file exists. Default: current cookbook
- `service_type` - is the type of launchd plist you would like to create.
Options are [system, user], System with create LaunchDaemon, and User will
create a LaunchAgent. Default: system

##### Apple Params
These parameters are Apples keys/values that you can add to a launch daemon or
agent. For more details see:

https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man5/launchd.plist.5.html#//apple_ref/doc/man/5/launchd.plist

- `program` - This key maps to the first argument of execvp(3). If this key is
missing, then the first element of the array of strings provided to the
ProgramArguments will be used instead. This key is required in the absence of
the ProgramArguments key.
- `program_arguments` - This key maps to the second argument of execvp. This key
is required in the absence of the Program key. Please note: many people are
confused by this key. Please read execvp very carefully!
- `abandon_process_group` - When a job dies, launchd kills any remaining
processes with the same process group ID as the job. Set-ting Setting ting
this key to true disables that behavior.
- `debug` - This optional key specifies that launchd should adjust its log mask
temporarily to LOG_DEBUG while dealing with this job.
- `disabled` - This optional key is used as a hint to launchctl(1) that it
should not submit this job to launchd when loading a job or jobs.
- `enable_globbing` - This flag causes launchd to use the glob(3) mechanism to
update the program arguments before invocation.
- `enable_transactions` - This flag instructs launchd that the job promises to
use vproc_transaction_begin(3) and vproc_transaction_end(3) to track
outstanding transactions that need to be reconciled before the process can
safely terminate. If no outstanding transactions are in progress, then launchd
is free to send the SIGKILL signal.
- `environment_variables` - This optional key is used to specify additional
environmental variables to be set before running the job.
- `exit_timeout` - The amount of time launchd waits before sending a SIGKILL
signal. The default value is 20 seconds. The value zero is interpreted as
infinity.
- `group` - This optional key specifies the group to run the job as. This key is
only applicable when launchd is running as root. If UserName is set and
GroupName is not, the the group will be set to the default group of the user.
- `hard_resource_limits` - Resource limits to be imposed on the job. These
adjust variables set with setrlimit(2).
- `inetd_compatibility` - The presence of this key specifies that the daemon
expects to be run as if it were launched from inetd.
- `init_groups` - This optional key specifies whether initgroups(3) should be
called before running the job. The default is true in 10.5 and false in 10.4.
- `keep_alive` - This optional key is used to control whether your job is to be
kept continuously running or to let demand and conditions control the
invocation.
- `launch_only_once` - This optional key specifies whether the job can only be
run once and only once. In other words, if the job cannot be safely respawned
without a full machine reboot, then set this key to be true.
- `limit_load_from_hosts` - This configuration file only applies to hosts NOT
listed with this key
- `limit_load_to_hosts` - This configuration file only applies to the hosts
listed with this key
- `limit_load_to_session_type` - This configuration file only applies to
sessions of the type specified
- `low_priority_io` - This optional key specifies whether the kernel should
consider this daemon to be low priority when doing file system I/O.
- `mach_services` - This optional key is used to specify Mach services to be
registered with the Mach bootstrap sub-system.
- `nice` - This optional key specifies what nice(3) value should be applied to
the daemon.
- `on_demand` - This key was used in Mac OS X 10.4 to control whether a job was
kept alive or not.
- `process_type` - This optional key describes, at a high level, the intended
purpose of the job. The system will apply resource limits based on what kind
of job it is. [Background, Standard, Adaptive, Interactive]
- `queue_directories` - Much like the WatchPaths option, this key will watch the
paths for modifications. The difference being that the job will only be
started if the path is a directory and the directory is not empty.
- `root_directory` - This optional key is used to specify a directory to
chroot(2) to before running the job.
- `run_at_load` - This optional key is used to control whether your job is
launched once at the time the job is loaded. The default is false.
- `sockets` - This optional key is used to specify launch on demand sockets that
can be used to let launchd know when to run the job.
- `soft_resource_limits` - Resource limits to be imposed on the job. These
adjust variables set with setrlimit(2).
- `standard_error_path` - This optional key specifies what file should be used
for data being sent to stderr when using stdio(3).
- `standard_in_path` - This optional key specifies what file should be used for
data being supplied to stdin when using stdio(3).
- `standard_out_path` - This optional key specifies what file should be used for
data being sent to stdout when using stdio(3).
- `start_calendar_interval` - This optional key causes the job to be started
every calendar interval as specified.
- `start_interval` - This optional key causes the job to be started every N
seconds.
- `start_on_mount` - This optional key causes the job to be started every time a
filesystem is mounted.
- `throttle_interval` - This key lets one override the default throttling policy
imposed on jobs by launchd. The value is in seconds, and by default, jobs will
not be spawned more than once every 10 seconds.
- `time_out` - The recommended idle time out (in seconds) to pass to the job. If
no value is specified, a default time out will be supplied by launchd for use
by the job at check in time.
- `umask` - This optional key specifies what value should be passed to umask(2)
before running the job.
- `username` - This optional key specifies the user to run the job as. This key
is only applicable when launchd is running as root.
- `wait_for_debugger` - This optional key specifies that launchd should instruct
the kernel to have the job wait for a debugger to attach before any code in
the job is executed.
- `watch_paths` - This optional key causes the job to be started if any one of
the listed paths are modified.
- `working_directory` - This optional key is used to specify a directory to
chdir(2) to before running the job.

#### Examples
Create a LaunchDaemon from cookbook file
```
launchd 'com.chef.every15' do
  source 'com.chef.every15.plist'
end
```

Create a LaunchDaemon using keys
```
launchd 'call.mom.weekly' do
  program '/Library/scripts/call_mom.sh'
  start_calendar_interval 'Weekday' => 7, 'Hourly' => 10
  time_out 300
end
```

Remove LaunchDaemon
```
launchd 'com.chef.every15' do
  action :delete
end

```

### DIFF
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -72,7 +72,7 @@ class Chef
       def manage_plist(action)
         if @source
           res = cookbook_file_resource
-        elsif
+        else
           res = file_resource
         end
         res.run_action(action)
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -1,0 +1,205 @@
+#
+# Author:: Mike Dodge (<mikedodge04@gmail.com>)
+# Copyright:: Copyright (c) 2015 Facebook, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider'
+require 'chef/resource'
+require 'chef/resource/file'
+require 'chef/resource/cookbook_file'
+require 'chef/resource/macosx_service'
+
+class Chef
+  class Provider
+    class Launchd < Chef::Provider
+
+      def load_current_resource
+        @current_resource = Chef::Resource::Launchd.new(@new_resource.name)
+        @name = @new_resource.name
+        @label = @new_resource.label ? @new_resource.label : @name
+        @backup = @new_resource.backup
+        @cookbook = @new_resource.cookbook
+        @group = @new_resource.group
+        @mode = @new_resource.mode
+        @owner = @new_resource.owner
+        @path = @new_resource.path
+        @source = @new_resource.source
+        @content = content?
+      end
+
+      def action_create
+        manage_plist(:create)
+      end
+
+      def action_create_if_missing
+        manage_plist(:create_if_missing)
+      end
+
+      def action_delete
+        # If you delete a service you want to make sure its not loaded or
+        # the service will be in memory and you wont be able to stop it.
+        if ::File.exists?(@path)
+          manage_service(:disable)
+        end
+        manage_plist(:delete)
+      end
+
+      def action_enable
+        if manage_plist(:create)
+          manage_service(:restart)
+        else
+          manage_service(:enable)
+        end
+      end
+
+      def action_disable
+        manage_service(:disable)
+      end
+
+      def manage_plist(action)
+        if @source
+          res = cookbook_file_resource
+        elsif
+          res = file_resource
+        end
+        res.run_action(action)
+        new_resource.updated_by_last_action(true) if res.updated?
+        res.updated
+      end
+
+      def manage_service(action)
+        res = service_resource
+        res.run_action(action)
+        new_resource.updated_by_last_action(true) if res.updated?
+      end
+
+      def service_resource
+        res = Chef::Resource::MacosxService.new(@label, run_context)
+        res.name(@label)
+        res.service_name(@label)
+        res.plist(@path)
+        res.session_type(@session_type) if @session_type
+
+        res
+      end
+
+      def file_resource
+        res = Chef::Resource::File.new(@path, run_context)
+        res.name(@path)
+        res.backup(@backup) if @backup
+        res.content(@content) if @content
+        res.group(@group) if @group
+        res.mode(@mode) if @mode
+        res.owner(@owner) if @owner
+        res.source(@source) if @source
+
+        res
+      end
+
+      def cookbook_file_resource
+        res = Chef::Resource::CookbookFile.new(@path, run_context)
+        res.name(@path)
+        res.backup(@backup) if @backup
+        res.content(@content) if @content
+        res.cookbook_name = @cookbook if @cookbook
+        res.group(@group) if @group
+        res.mode(@mode) if @mode
+        res.owner(@owner) if @owner
+        res.source(@source) if @source
+
+        res
+      end
+
+      def define_resource_requirements
+        super
+        requirements.assert(
+          :create, :create_if_missing, :delete, :enable, :disable
+        ) do |a|
+          type = @new_resource.type.to_s
+          a.assertion { !['daemon', 'agent'].include?(type) }
+          a.failure_message(
+            Chef::Exceptions::UnsupportedAction,
+            "service_type must be 'daemon' or 'agent'"
+          )
+        end
+      end
+
+      private
+
+      def content?
+        if @new_resource.hash
+          @new_resource.hash.to_plist
+        elsif @new_resource.program || @new_resource.program_arguments
+          gen_plist
+        end
+      end
+
+      def gen_plist
+        plist_hash = {}
+        {
+          'label' => 'Label',
+          'program' => 'Program',
+          'program_arguments' => 'ProgramArguments',
+          'abandon_process_group' => 'AbandonProcessGroup',
+          'debug' => 'Debug',
+          'disabled' => 'Disabled',
+          'enable_globbing' => 'EnableGlobbing',
+          'enable_transactions' => 'EnableTransactions',
+          'environment_variables' => 'EnvironmentVariables',
+          'exit_timeout' => 'ExitTimeout',
+          'ld_group' => 'GroupName',
+          'hard_resource_limits' => 'HardreSourceLimits',
+          'inetd_compatibility' => 'inetdCompatibility',
+          'init_groups' => 'InitGroups',
+          'keep_alive' => 'KeepAlive',
+          'launch_only_once' => 'LaunchOnlyOnce',
+          'limit_load_from_hosts' => 'LimitLoadFromHosts',
+          'limit_load_to_hosts' => 'LimitLoadToHosts',
+          'limit_load_to_session_type' => 'LimitLoadToSessionType',
+          'low_priority_io' => 'LowPriorityIO',
+          'mach_services' => 'MachServices',
+          'nice' => 'Nice',
+          'on_demand' => 'OnDemand',
+          'process_type' => 'ProcessType',
+          'queue_directories' => 'QueueDirectories',
+          'root_directory' => 'RootDirectory',
+          'run_at_load' => 'RunAtLoad',
+          'sockets' => 'Sockets',
+          'soft_resource_limits' => 'SoftResourceLimits',
+          'standard_error_path' => 'StandardErrorPath',
+          'standard_in_path' => 'StandardInPath',
+          'standard_out_path' => 'StandardOutPath',
+          'start_calendar_interval' => 'StartCalendarInterval',
+          'start_interval' => 'StartInterval',
+          'start_on_mount' => 'StartOnMount',
+          'throttle_interval' => 'ThrottleInterval',
+          'time_out' => 'TimeOut',
+          'umask' => 'Umask',
+          'username' => 'UserName',
+          'wait_for_debugger' => 'WaitForDebugger',
+          'watch_paths' => 'WatchPaths',
+          'working_directory' => 'WorkingDirectory',
+        }.each do |key, val|
+          if @new_resource.send(key)
+            plist_hash[val] = @new_resource.send(key)
+          end
+        end
+        plist_hash.to_plist
+      end
+
+    end
+  end
+end
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -17,27 +17,43 @@
 #
 
 require 'chef/provider'
-require 'chef/resource'
+require 'chef/resource/launchd'
 require 'chef/resource/file'
 require 'chef/resource/cookbook_file'
 require 'chef/resource/macosx_service'
+require 'plist'
+require "forwardable"
 
 class Chef
   class Provider
     class Launchd < Chef::Provider
+      extend Forwardable
+      provides :launchd, os: "darwin"
+
+      def_delegators :@new_resource, *[
+        :backup,
+        :cookbook,
+        :group,
+        :label,
+        :mode,
+        :owner,
+        :path,
+        :source,
+        :session_type,
+        :type
+      ]
 
       def load_current_resource
-        @current_resource = Chef::Resource::Launchd.new(@new_resource.name)
-        @name = @new_resource.name
-        @label = @new_resource.label ? @new_resource.label : @name
-        @backup = @new_resource.backup
-        @cookbook = @new_resource.cookbook
-        @group = @new_resource.group
-        @mode = @new_resource.mode
-        @owner = @new_resource.owner
-        @path = @new_resource.path
-        @source = @new_resource.source
-        @content = content?
+        current_resource = Chef::Resource::Launchd.new(new_resource.name)
+        @path = path ? path : gen_path_from_type
+      end
+
+      def gen_path_from_type
+        types = {
+          'daemon' => "/Library/LaunchDaemons/#{label}.plist",
+          'agent' => "/Library/LaunchAgents/#{label}.plist"
+        }
+        types[type]
       end
 
       def action_create
@@ -70,7 +86,7 @@ class Chef
       end
 
       def manage_plist(action)
-        if @source
+        if source
           res = cookbook_file_resource
         else
           res = file_resource
@@ -87,37 +103,34 @@ class Chef
       end
 
       def service_resource
-        res = Chef::Resource::MacosxService.new(@label, run_context)
-        res.name(@label)
-        res.service_name(@label)
-        res.plist(@path)
-        res.session_type(@session_type) if @session_type
-
+        res = Chef::Resource::MacosxService.new(label, run_context)
+        res.name(label) if label
+        res.service_name(label) if label
+        res.plist(@path) if @path
+        res.session_type(session_type) if session_type
         res
       end
 
       def file_resource
         res = Chef::Resource::File.new(@path, run_context)
-        res.name(@path)
-        res.backup(@backup) if @backup
-        res.content(@content) if @content
-        res.group(@group) if @group
-        res.mode(@mode) if @mode
-        res.owner(@owner) if @owner
-
+        res.name(@path) if @path
+        res.backup(backup) if backup
+        res.content(content) if content
+        res.group(group) if group
+        res.mode(mode) if mode
+        res.owner(owner) if owner
         res
       end
 
       def cookbook_file_resource
         res = Chef::Resource::CookbookFile.new(@path, run_context)
-        res.cookbook_name = @cookbook if @cookbook
-        res.name(@path)
-        res.backup(@backup) if @backup
-        res.group(@group) if @group
-        res.mode(@mode) if @mode
-        res.owner(@owner) if @owner
-        res.source(@source) if @source
-
+        res.cookbook_name = cookbook if cookbook
+        res.name(@path) if @path
+        res.backup(backup) if backup
+        res.group(group) if group
+        res.mode(mode) if mode
+        res.owner(owner) if owner
+        res.source(source) if source
         res
       end
 
@@ -125,23 +138,24 @@ class Chef
         requirements.assert(
           :create, :create_if_missing, :delete, :enable, :disable
         ) do |a|
-          type = @new_resource.type.to_s
-          a.assertion { ['daemon', 'agent'].include?(type) }
-          error_msg = 'type must be daemon or agent'
+          type = new_resource.type
+          a.assertion { ['daemon', 'agent'].include?(type.to_s) }
+          error_msg = 'type must be daemon or agent.'
           a.failure_message Chef::Exceptions::ValidationFailed, error_msg
         end
       end
 
       def content?
-        if @new_resource.hash
-          @new_resource.hash.to_plist
-        elsif @new_resource.program || @new_resource.program_arguments
-          gen_plist
-        end
+        !!content
       end
 
-      def gen_plist
-        plist_hash = {}
+      def content
+        plist_hash = new_resource.hash || gen_hash
+        Plist::Emit.dump(plist_hash) unless plist_hash.nil?
+      end
+
+      def gen_hash
+        return nil unless new_resource.program || new_resource.program_arguments
         {
           'label' => 'Label',
           'program' => 'Program',
@@ -185,14 +199,10 @@ class Chef
           'wait_for_debugger' => 'WaitForDebugger',
           'watch_paths' => 'WatchPaths',
           'working_directory' => 'WorkingDirectory',
-        }.each do |key, val|
-          if @new_resource.send(key)
-            plist_hash[val] = @new_resource.send(key)
-          end
+        }.each_with_object({}) do |(key, val), memo|
+          memo[val] = new_resource.send(key) if new_resource.send(key)
         end
-        plist_hash.to_plist
       end
-
     end
   end
 end
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -104,17 +104,15 @@ class Chef
         res.group(@group) if @group
         res.mode(@mode) if @mode
         res.owner(@owner) if @owner
-        res.source(@source) if @source
 
         res
       end
 
       def cookbook_file_resource
         res = Chef::Resource::CookbookFile.new(@path, run_context)
+        res.cookbook_name = @cookbook if @cookbook
         res.name(@path)
         res.backup(@backup) if @backup
-        res.content(@content) if @content
-        res.cookbook_name = @cookbook if @cookbook
         res.group(@group) if @group
         res.mode(@mode) if @mode
         res.owner(@owner) if @owner
@@ -124,20 +122,15 @@ class Chef
       end
 
       def define_resource_requirements
-        super
         requirements.assert(
           :create, :create_if_missing, :delete, :enable, :disable
         ) do |a|
           type = @new_resource.type.to_s
-          a.assertion { !['daemon', 'agent'].include?(type) }
-          a.failure_message(
-            Chef::Exceptions::UnsupportedAction,
-            "service_type must be 'daemon' or 'agent'"
-          )
+          a.assertion { ['daemon', 'agent'].include?(type) }
+          error_msg = 'type must be daemon or agent'
+          a.failure_message Chef::Exceptions::ValidationFailed, error_msg
         end
       end
-
-      private
 
       def content?
         if @new_resource.hash
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -16,12 +16,12 @@
 # limitations under the License.
 #
 
-require 'chef/provider'
-require 'chef/resource/launchd'
-require 'chef/resource/file'
-require 'chef/resource/cookbook_file'
-require 'chef/resource/macosx_service'
-require 'plist'
+require "chef/provider"
+require "chef/resource/launchd"
+require "chef/resource/file"
+require "chef/resource/cookbook_file"
+require "chef/resource/macosx_service"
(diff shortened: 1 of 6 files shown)


jaymzh:
CC @chef/client-core and @chef/client-os-x

coderanger:
Hawtttttt

ranjib:
no specs?

mikedodge04:
Working on those next. But would like to get some feedback on what I have so far.

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider/file'
+
+class Chef
+  class Provider
+    class Launchd < Chef::Provider::File
```
btm: Why `Chef::Provider::File` and not `Chef::Provider::Service`?

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,59 @@
...
+        super
+      end
+
+      def define_resource_requirements
+        super
+        requirements.assert(:create, :create_if_missing, :touch) do |a|
+          service_type = @new_resource.service_type.to_s
+          a.assertion { !['system', 'user'].include?(service_type) }
+          a.failure_message(
+            Chef::Exceptions::UnsupportedAction,
```
btm: I'd save UnsupportedAction for just when an *action* is provided that the provider doesn't support. Probably best to use `Chef::Exceptions::Service` here.

lib/chef/resource/launchd.rb
```diff
@@ -0,0 +1,285 @@
...
+require 'chef/resource/file'
+require 'chef/provider/launchd'
+require 'chef/mixin/securable'
+
+class Chef
+  class Resource
+    class Launchd < Chef::Resource::File
+      include Chef::Mixin::Securable
+
+      provides :launchd
```
btm: This shouldn't be necessary since you're not providing an `os` constraint, this happens when you declare the resource_name.

https://github.com/chef/chef/blob/master/lib/chef/resource.rb#L941

lib/chef/resource/launchd.rb
```diff
@@ -0,0 +1,285 @@
...
+        @action = 'create'
+        @cookbook = nil
+        @source = nil
+        @label = name
+        @disabled = false
+        @service_type = 'system'
+        @path = "/Library/LaunchDaemons/#{name}.plist"
+      end
+
+      def label(label = @label)
```
btm: I'm pretty sure you want `def label(label=nil)` here. set_or_return will return the value of the instance variable if you pass nil, or set it if you pass non-nil. In this case, any time you call `label` without arguments you're going to be resetting it to itself (and returning that value), which is unnecessary.

This example might help:

```
require 'chef/mixin/params_validate'

include Chef::Mixin::ParamsValidate

@val_nil = "initialized value"
@val_passed = "initialized value"

def val_nil(arg=nil)
  set_or_return(:val_nil, arg, :kind_of => String)
end

def val_passed(arg = "method argument")
  set_or_return(:val_passed, arg, :kind_of => String)
end

puts val_nil
puts val_passed

puts val_nil("set")
puts val_passed("set")

puts val_nil
puts val_passed
```

```
$ ruby set_or_return.rb
initialized value
method argument
set
set
set
method argument
```

lib/chef/resource/launchd.rb
```diff
@@ -0,0 +1,285 @@
...
+        set_or_return(:hash, hash, :kind_of => Hash)
+      end
+
+      def service_type(service_type = @service_type)
+        service_type = service_type.downcase
+        if service_type == 'system'
+          @path = "/Library/LaunchDaemons/#{name}.plist"
+        elsif service_type == 'user'
+          @path = "/Library/LaunchAgents/#{name}.plist"
+        end
```
btm: I know the assertion would normally catch it for a run, but I'd add an else here to make sure this method isn't passed an invalid service type and raise.

lib/chef/resource/launchd.rb
```diff
@@ -0,0 +1,285 @@
...
+
+      def path(arg = @path)
+        set_or_return(:path, arg, :kind_of => String)
+      end
+
+      def hash(hash = nil)
+        set_or_return(:hash, hash, :kind_of => Hash)
+      end
+
+      def service_type(service_type = @service_type)
```
btm: Name this more appropriately for what it does, e.g. `path_for_service_type`.

lib/chef/resource/launchd.rb
```diff
@@ -0,0 +1,285 @@
...
+      def start_on_mount(start_on_mount = nil)
+        set_or_return(
+          :start_on_mount, start_on_mount, :kind_of => [TrueClass, FalseClass]
+        )
+      end
+
+      def throttle_interval(throttle_interval = nil)
+        set_or_return(
+          :throttle_interval, throttle_interval, :kind_of => Integer
+        )
```
btm: it's weird to me visually that a bunch of these aren't on one line. I get that with whitespace they're over 80 characters though. Not saying to change it, I just kept expecting these ones to be different somehow from the ones on a single line.

lib/chef/resource/launchd.rb
btm: also just realized that this has a side-affect of directly setting path. `set_path_for_service_type`?

btm:
hrm. okay, I see now that you're inheriting from Chef::Provider::File instead of Chef::Provider::Service because you're reusing the file provider code to create the plist file. That's confusing... 

I see `Chef::Provider::Launchd::Content#get_cookbook_file` is also duplicate of `Chef::Provider::CookbookFile::Content#file_for_provider`.

Oh, let's back way up. I had presumed this was a platform-independant launchd service provider, and was going to eventually ask how this would replace Chef::Provider::Service::Macosx. But this isn't a service provider, this manages plist files for launchd. If that's actually what you want, this needs to be much more clear about that. I would expect the resource to be something like launchd_file or plist_file, not launchd. 

I'm pretty sure this should be create action added to the Macosx service.

btw, putting :construction: in the PR description is code for not done yet.

lib/chef/resource/launchd.rb
tyler-ball: I don't think its bad to include this, even if its already the default.  I think it makes resources/providers more obvious (and follows a pattern) if they all call `provides`

lamont-granquist:
Inheriting from Chef::Provider::File in order to reuse the file provider code is not appropriate.  Objects should only inherit when they've got is_a relationships.  Mixins or composition are more appropriate just for code-reuse.

In order to use a file provider from another provider, this is the pattern to follow:

https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig.rb#L174-L186

We should really break out build_resource into its own mixin so that we get better syntax without mixing back in the whole DSL, but nobody has done that yet.

jaymzh:
@btm  I don't think `launchd_file` is appropriate unless you think we should rename the `cron` resource to `cron_file`... Yes, it technically lays down a file, but that's just implementation details of how launchd works... it's actually setting up launchd jobs in the same way that cron sets up cronjobs.

btm:
@jaymzh I think this needs to part of a launchd service provider, and we answer the question as to the relationship between that and the existing macosx service provider.

the cron resource provides everything you need for managing cron jobs. launchd is "a unified, open-source service-management framework, starts, stops and manages daemons, applications, processes, and scripts in Apple OS X environments" , but the code presented here only manages launchd configuration files.

danielsdeleo:
We don't really manage the service script/file/whatever for other service types, do we? Unless I'm missing something, the only prior art I could find is https://github.com/nathwill/chef-systemd I have to say that I would probably have named the resources differently than they did though.

Anyway, is "job" the correct term for "things that exist in launchd"? `launchd_job` seems to make sense and would be distinct from a hypothetical `launchd_service`, if it follows the official terminology.

btm:
@danielsdeleo I can't think of any time we create the configurations for services in core in the past either. The packages and cookbooks typically did that. Sometimes we've modified them, like in early upstart to "disable" them.

mikedodge04:
launchd is a little weird, and you wouldn't want to join it with the service provider, because there are instances where you would want a launchd plist created but never loaded. Depending on what is contained in the launchd plist, and where it's placed, OS X may load it for you.

lib/chef/resource/launchd.rb
mikedodge04: When I wrote it this way I wanted to abstract the fact that launchd plists have two different types, and the only difference between the two are where they're placed. One is place in a location(/Library/LaunchDaemons/) and ran as root, the other is placed in a separate location(/Library/LaunchAgents/) and ran as user. And yes this is just overriding "path", because for some reason if someone wanted to create a launchd plist somewhere other than the normal locations I wanted to support that.

btm: I think the abstraction is fine. If was reading some code that called #service_type, I would expect to get service_type back, so I think this should be renamed to something that better indicates what it does.

natewalck:
Expanding on what @mikedodge04 said, with launchd, it is common to lay down a LaunchAgent that only runs at login, so you do *not* want to load the launchd, only place it.

Also, it is common to manage the state of a service without managing the launchd content. (for instance, ensuring the bluetooth service is always unloaded)

LaunchD is different than most other services and I think it is appropriate to separate out launchd creation/management from managing the status of the service. 

Eventually, we will have a wrapper cookbook that can be used to abstract the creation/loading of a service to one resource call as a shortcut, but in reality creation and management are two distinct management steps.

danielsdeleo:
Yeah, we're all clear about what launchd does and the general shape of how you'd manage it with chef using the proposed functionality here. We're just trying to decide what we think would be a clear and unambiguous name for this resource. As Bryan said:

> launchd is "a unified, open-source service-management framework, starts, stops and manages daemons, applications, processes, and scripts in Apple OS X environments"

So just calling it `launchd` doesn't indicate what is being managed (i.e., is it making files that configure launchd to do things, or is it poking `launchctl` ?). I agree with @btm that we should have one more word in there to make this unambiguous.

natewalck:
```launchd_plist``` would make the most sense. In fact, that is exactly what the man page is for the file itself: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man5/launchd.plist.5.html

**Comment by danielsdeleo** [comment]
`launchd_plist` sounds perfect to me

**Comment by chef-supermarket** [comment]

File: lib/chef/resource/launchd.rb (hunk 5 above)
Comment (mikedodge04) [inline lib/chef/resource/launchd.rb]: Do you have a recommendation on which exception I should use?

File: lib/chef/provider/launchd_plist.rb [hunk 8]
```diff
@@ -0,0 +1,58 @@
...
+        super
+      end
+
+      def define_resource_requirements
+        super
+        requirements.assert(:create, :create_if_missing, :touch) do |a|
+          type = @new_resource.type.to_s
+          a.assertion { !['daemon', 'agent'].include?(type) }
+          a.failure_message(
+            Chef::Exceptions::UnsupportedAction,
```
btm: > Do you have a recommendation on which exception I should use?

Chef::Exceptions::ValidationFailed or consider adding something like `class InvalidProperty < ArgumentError; end`

lib/chef/provider/launchd_plist.rb
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider/file'
+
+class Chef
+  class Provider
+    class LaunchdPlist < Chef::Provider::File
```
lamont-granquist: this is still a bad way to do this, better to construct the Chef::Provider::File sub resource explicitly.

lamont-granquist:
Okay the way this is implemented as a file class provider subclass with a content provider is still mildly breaking my brain.  Whats the reason for not writing this using a file/template/cookbook_file sub-resource?

mikedodge04:
@lamont-granquist 
launchd_plist is managing files, so I looked at how cookbook_file was implemented and mimicked how it was a subclass of file. Made the most sense to me, because I get all the benefits of file by subclass'ing from file. It handles some of the file magic for me. 

I'm not sure what you are referring to by 'sub-resource' do you have an example?

lamont-granquist:
Well its more like what you'd do if you wrote an LWRP/Custom Resource instead of subclassing file.  Just construct the resource and inject what you want with properties on the resource.

Here's an example of using a file sub-resource:

https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig.rb#L174-L186

Here's an example of using a directory sub-resource (bit cleaner code, but less applicable use case):

https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig/debian.rb#L65-L68

You can also use CookbookFile or Template classes instead of the File class if those are better approaches (e.g. if you find yourself copying code out of CookbookFile to inject into the content property of the File provider, you probably just want to use CookbookFile directly).  Also the syntax is a little clunky but I'm working on that.  I want to get declare_resource/build_resource added into Chef::Provider itself so that those become available (without mixing in the entire DSL), and then make it easier to open up a sub-resource collection from a 'heavy' provider and have it do the updated_by_last_action stuff for you just like LWRPs.

The problem with subclassing is that now you're brittle to the internal details of how file providers are constructed.  If you use a sub-resource then you're using the external API of the file provider and that public API is unlikely to ever change.

There's various bits of discussions about inheritance vs. composition with different biases.  A fairly neutral one seems to be here:  https://www.thoughtworks.com/insights/blog/composition-vs-inheritance-how-choose

That "How to Misuse Inheritance - Example 2" is sort of the argument I'm trying to make here.  You're solving a specific domain problem (managing a plist file) through inheriting from an implementation class (managing a file) rather than 'using' the implementation class.

Problem is what I'm asking for is a pretty major rewrite since it would likely throw away all the content class stuff -- but I think it would be ultimately simpler and less brittle -- but i don't know if there's a reason behind the current complexity.

mikedodge04:
Ok, I'm willing to do the rewrite. But doesn't that mean i'm going to have to reproduce all of the resource methods of file in order to be able to reproduce the file functionality? Of should my launchd resource still sub class from the file resource?

lamont-granquist:
if you want your parent resource to have all the file resource properties then there's no simple way to code that.  its probably more reasonable to inherit from the Chef::Resource::File class to avoid having to make the same interface declarations, and then you just have to manually send those properties to the file sub-resource when you build it.

That means that when the file resource is extended you'll aquire new properties that your provider doesn't do anything with, but we don't have a way to easily say 'build this resource, and for all its properties, copy them from mine'.  I've hit that pattern a few times, and it would be good to have a concise way of saying that, but currently we don't have that in the language.

lamont-granquist:
Actually I think @jkeiser addressed this use case in new resources...

danielsdeleo:
So the file resource has:

* atomic_update
* backup
* checksum
* content
* force_unlink
* group
* inherits (windows permission thing) 
* manage_symlink_source
* mode
* owner
* path
* rights (windows permissions)
* sensitive
* verify

As far as I can tell, only a few of these actually apply for a launchd plist. Certainly the windows stuff won't, checksum and content shouldn't be set directly by the user, you can probably just pick sensible defaults for atomic update, force unlink, and manage symlink source, and you can also either ignore verify or set a sane default. So that's half the behavior of a file resource that doesn't apply. Copying the other ones is a bit tedious but will be better in the end.

mikedodge04:
Also one last question. Since i'm not sub-classing file I was thinking of renaming the resource from launchd_plist to launchd and adding an enable and disable action so you could do something like :

# this would create
launchd 'com.chef.every15' do
  source 'com.chef.every15.plist'
  action :create
end

# this would create and enable
launchd 'com.chef.every15' do
  source 'com.chef.every15.plist'
  action :enable
end

Thoughts?

lamont-granquist:
i think that makes sense...

also i think you can do this:

```ruby
class MyResource < Chef::Resource
  Chef::Resource::File.properties.each { |p| property p.name, p }
end
```

```ruby
action :create do
  file name do
    new_resource.state.each_pair { |key, value| send(key, value) }
  end
end
```

That will get you the same effect as inheritance without using inheritance.

danielsdeleo:
That is nifty ruby/chef programming, but I think it just gives you the same problems as inheritance would--namely, the subclass is only sort-of a kind-of file, file has a lot of properties that make no sense in the context of a launchd entry. cf.

* http://tenderlovemaking.com/2014/06/02/yagni-methods-are-killing-me.html
* http://tenderlovemaking.com/2014/06/04/yagni-methods-slow-us-down.html

lamont-granquist:
So instead of slurping all the properties, just use a fixed list of what you actually want to implement?

```
[:mode, :owner, :group].each do { |p| property p, Chef::Resource::File.properties[p].options }
```

We should really add some DSL sugar to make that easier:

```
derive_properties_from(Chef::Resource::File, [:mode, :owner, :group])
```

And then same kind of thing later:

```
action :create do
  file name do
    [:mode, :owner, :group].each { |p| send(p, new_resource.p) }
  end
end
```

mikedodge04:
where would i put this ? in the provider?
```
action :create do
  file name do
    [:mode, :owner, :group].each { |p| send(p, new_resource.p) }
  end
end
```

undefined method `action' for Chef::Provider::Launchd:Class

lamont-granquist:
that's `def action_create` if you're creating a 'heavy' provider.

mikedodge04:
ok thought so. Wasn't sure if you had some magic i hadn't seen before.

lamont-granquist:
so copying properties from the file resource won't work because they're not properties, they're just methods.

mikedodge04:
@lamont-granquist
i'm using the method names, as properties to pass them to the file sub-resource. It is working in my testing. I will put some spec's up soon.

mikedodge04:
Any comments ?

jaymzh:
@mikedodge04 you need to rebase
@lamont-granquist  are you happy with this?

mikedodge04:
Any other thoughts on this PR? 
Added a PR for docs: https://github.com/chef/chef-web-docs/pull/793

jamescott:
Looks like it's close -- a rebase and approval from Lamont.

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,198 @@
...
+        @current_resource = Chef::Resource::Launchd.new(@new_resource.name)
+        @name = @new_resource.name
+        @label = @new_resource.label ? @new_resource.label : @name
+        @backup = @new_resource.backup
+        @cookbook = @new_resource.cookbook
+        @group = @new_resource.group
+        @mode = @new_resource.mode
+        @owner = @new_resource.owner
+        @path = @new_resource.path
+        @source = @new_resource.source
```
lamont-granquist: you could drop all the `@`signs off of `@new_resource` here.

it also looks more like you want to use `def_delegators` here to forward `#path` to `new_resource.path`.

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,198 @@
...
+
+      def cookbook_file_resource
+        res = Chef::Resource::CookbookFile.new(@path, run_context)
+        res.cookbook_name = @cookbook if @cookbook
+        res.name(@path)
+        res.backup(@backup) if @backup
+        res.group(@group) if @group
+        res.mode(@mode) if @mode
+        res.owner(@owner) if @owner
+        res.source(@source) if @source
```
lamont-granquist: all this kind of stuff should be using just `source` which should be an accessor delegated to `new_resource.source`.

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,198 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider'
+require 'chef/resource'
```
lamont-granquist: should be 'chef/resource/launchd'

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,198 @@
...
+          'umask' => 'Umask',
+          'username' => 'UserName',
+          'wait_for_debugger' => 'WaitForDebugger',
+          'watch_paths' => 'WatchPaths',
+          'working_directory' => 'WorkingDirectory',
+        }.each do |key, val|
+          if @new_resource.send(key)
+            plist_hash[val] = @new_resource.send(key)
+          end
+        end
```
lamont-granquist: slightly tighter ruby:

```
def gen_plist
  { 
    ...big hash...
  }.each_with_object({}) do |k, v, memo|
    memo[val] = new_resource.send(key) if new_resource.respond_to?(:key)
  end.to_plist
end
```

jamescott:
Why is `abandon_process_group` a `:kind_of => String`? Should it be `:kind_of => [TrueClass, FalseClass]`?

lib/chef/provider/launchd.rb
```diff
@@ -0,0 +1,198 @@
...
+          a.failure_message Chef::Exceptions::ValidationFailed, error_msg
+        end
+      end
+
+      def content?
+        if @new_resource.hash
+          @new_resource.hash.to_plist
+        elsif @new_resource.program || @new_resource.program_arguments
+          gen_plist
+        end
```
lamont-granquist: so:

gen_plist should probably just return a hash so that we can call to_plist in one spot here.  this also seems to be overloaded as a boolean check even though it returns a plist.  plus we should explicitly require plist in this file, and we should not use the monkeypatch (really whatever introduced that `Hash#to_plist` monkeypatch should get ripped out).

```ruby
require 'plist'

def content?
  !!content
end

def content
  plist_hash = new_resource.hash || gen_plist_hash
  Plist::Emit.dump(plist_hash) unless plist_hash.nil? 
end

def gen_plist_hash
  return nil unless @new_resource.program || @new_resource.program_arguments
  [...exactly like gen_plist but returns a hash...]
end
```

jamescott:
Also, should be `nice` `:kind_of => Integer`?

lamont-granquist:
the overall shape of how you create the sub-resources is fine now, just some nit picking left.

lamont-granquist:
YARD docs on the methods would also be good

and you can use properties now in resources code see Chef::Resource::File and Chef::Resource::Package, at some point we should replace all the old `set_or_return` attributes with `property` and stop accepting new ones into core.

mikedodge04:
Can you point me to an example of a method with YARD docs ?

lamont-granquist:
I usually pull up lib/chef/resources.rb and copy the style used in that file.

mikedodge04:
I'm not sure which could use more docs. Each of the methods seem self explanatory. Please let me know which methods you think could use a YARD doc and I'll add it.

thommay:
@mikedodge04 @jaymzh please fix the specs for this and then we're ready to merge.

lamont-granquist:
yeah lets just fix the specs and merge

mikedodge04:
After I fixed my broken spec and rebased there is still one broken test, but it isn't related to anything I touched. Let me know if you need anything else.

lamont-granquist:
its actually chefstyle complaining about this PR, just run 'chefstyle -a' over it all and commit the result and that should fix it.

lamont-granquist:
@mikedodge04 make sure to "bundle update && bundle exec chefstyle -a" you'll need to pull the latest version from git

mikedodge04:
@lamont-granquist Let me know if you need anything else

jaymzh:
@thommay or @lamont-granquist  can you give an explicit :+1:  and then I'll merge?

lamont-granquist:
:+1:

mikedodge04:
Thanks everyone for all your help. Now that this is merged, do we know what version this will ship with?

tas50:
@mikedodge04 It'll ship with the next Chef release.  Probably 12.8.something

jamescott:
Here are the preview docs for 12.8 for the launchd resource:

https://docs.chef.io/release/12-8/resource_launchd.html

I would appreciate a review from the folks in this thread for the properties, especially those specific to the Apple launchd.plist properties. There are still a few `xxxxx` placeholders in there for descriptions, known issue, just part of getting the docs moving forward.

A GitHub issue against this file: https://github.com/chef/chef-web-docs/blob/master/includes_resources/includes_resource_launchd_attributes.rst

would be sufficient for getting any feedback to my way.

Thanks!

james

jaymzh:
@mikedodge04 ^^

mikedodge04:
@jamescott There is a lot that needs to be changed. It seems that you combined the profile and launchd documentation. I had a PR for the docs that seemed to get taken down. Do you want me to put it up again?

btm:
@mikedodge04 We've been targeting shipping the client on the second Wednesday of every month, so watch for 12.8 around 3/9.

mikedodge04:
@jamescott the docs you linked to are incorrect 

syntax, actions, and the examples are not the ones I put up in my PR please see:
https://github.com/chef/chef-web-docs-2016/pull/793/files

PR #4112, opened 2015-10-28
Update knife bootstrap command to honor --no-color flag in chef-client run that is part of the bootstrap process.

tfitch:
A customer noticed this issue when using `knife bootstrap` with vCO and the color bits messing up the logging of the chef-client's output.

@btm suggested the line to add in `bootstrap_context.rb`.  I updated some existing tests to pass and added a new one "when color is true".

Submitting PR for merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 * [pr#4021](https://github.com/chef/chef/pull/4021) add missing requires for Chef::DSL::Recipe to LWRPBase
 * [pr#3597](https://github.com/chef/chef/pull/3597) print STDOUT from the powershell_script
 * [pr#4091](https://github.com/chef/chef/pull/4091) Allow downloading of root_files in a chef repository
+* [pr#4112](https://github.com/chef/chef/pull/4112) Update knife bootstrap command to honor --no-color flag in chef-client run that is part of the bootstrap process.
 
 ## 12.5.1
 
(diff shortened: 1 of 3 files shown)


btm:
I'd change the setup for the tests in `spec/unit/knife/core/bootstrap_context_spec.rb` to default to color being true, as that's the default for knife. This way, the tests that you modified here that test something like passing environment aren't affected by color. I would guess a single change like this would make all the tests pass, then add a new test specific to this feature that ensures --no-color shows up when color is false, like the other ones.

```
diff --git a/spec/unit/knife/core/bootstrap_context_spec.rb b/spec/unit/knife/core/bootstrap_context_spec.rb
index 4d69d63..e42ac5c 100644
--- a/spec/unit/knife/core/bootstrap_context_spec.rb
+++ b/spec/unit/knife/core/bootstrap_context_spec.rb
@@ -20,7 +20,7 @@ require 'spec_helper'
 require 'chef/knife/core/bootstrap_context'

 describe Chef::Knife::Core::BootstrapContext do
-  let(:config) { {:foo => :bar} }
+  let(:config) { {:color => :false } }
   let(:run_list) { Chef::RunList.new('recipe[tmux]', 'role[base]') }
   let(:chef_config) do
     {
```

ksubrama:
What @btm said.  Besides that, :+1:

tfitch:
So reverse my tests.  Got it.  Hahaha.  Thanks you two.

lamont-granquist:
Yep :+1: on @btm's comment, :+1: also on the code tho

tfitch:
@btm I updated the tests so the existing `expect` clauses remain unchanged and then the new test proves `--no-color` happens when `:color => false`

Good to merge?

PR #4118, opened 2015-10-29
Make Property.derive create derived properties of the same type

jkeiser:
The scenario: I'd like to use a custom property type so I can override `emit_dsl` and make the property's method a little different. But Property.derive doesn't retain the class by default--it just creates a Property, losing the class information.


ranjib:
:+1:

coderanger:
:+1:

PR #4120, opened 2015-10-29
Correct capitalization of GitHub username of a maintainer

robbkidd:
Supermarket uses this file to track maintainers and is currently
case-sensitive. `agentmeerkat` is [`AgentMeerkat`](https://github.com/agentmeerkat)
under the hood.

### DIFF
--- a/MAINTAINERS.toml
+++ b/MAINTAINERS.toml
@@ -194,7 +194,7 @@ The specific components of Chef related to a given platform - including (but not
         maintainers = [
           "Aevin1387",
           "tBunnyMan",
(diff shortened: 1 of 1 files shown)


ksubrama:
:+1:

danielsdeleo:
:+1: on getting this fixed now, but I wonder if we generate this via the API. Or do we sync the teams *to* github from this list?

ksubrama:
@danielsdeleo https://github.com/chef/chef/blob/master/tasks/maintainers.rb does the work.  It actually syncs from this file to github.

PR #4125, opened 2015-10-30
just playing around with using docker-vs-ec2

lamont-granquist:


lamont-granquist:
yeah, not enough time to play with this, and don't think it'll ultimately work because service resource yet again..

PR #4130, opened 2015-11-03
lazy the socketless require in Chef::HTTP

lamont-granquist:
if this is always required then everything that uses Chef::HTTP will
load up chef-zero which will load up Webrick, which does DNS lookups
of the current host which causes a DNS lookup and possible timeout.

it also likely isn't doing pretty things to chef-client memory usage
and load times there as well.

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -5,7 +5,7 @@
 # Author:: Christopher Brown (<cb@opscode.com>)
 # Author:: Christopher Walters (<cw@opscode.com>)
 # Author:: Daniel DeLeo (<dan@opscode.com>)
-# Copyright:: Copyright (c) 2009, 2010, 2013 Opscode, Inc.
+# Copyright:: Copyright (c) 2009, 2010, 2013-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -25,7 +25,6 @@ require 'tempfile'
 require 'net/https'
 require 'uri'
 require 'chef/http/basic_client'
-require 'chef/http/socketless_chef_zero_client'
 require 'chef/monkey_patches/net_http'
 require 'chef/config'
 require 'chef/platform/query_helpers'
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core @danielsdeleo can i get some :+1: 's?

lib/chef/http.rb
```diff
@@ -198,6 +197,9 @@ def streaming_request(path, headers={}, &block)
     def http_client(base_url=nil)
       base_url ||= url
       if chef_zero_uri?(base_url)
+        # PERFORMANCE CRITICAL: *MUST* lazy require here otherwise we load up all of
+        # chef-zero and webrick when we mostly never want to or have to
```
danielsdeleo: Probably should mention the DNS timeout thing?

lamont-granquist:
This addresses at least part of:

https://discourse.chef.io/t/knife-performance-on-mac-os-x/7085/19

I'm still seeing some bad load times with knife on my base ruby-2.2.2 rvm gemset outside of bundles (got tons of gems installed in it and requires just seem to take horribly long).

lib/chef/http.rb
```diff
@@ -198,6 +197,10 @@ def streaming_request(path, headers={}, &block)
     def http_client(base_url=nil)
       base_url ||= url
       if chef_zero_uri?(base_url)
+        # PERFORMANCE CRITICAL: *MUST* lazy require here otherwise we load up webrick
+        # via chef-zero and that hits DNS (at *require* time) which may timeout,
+        # when for most knife/chef-client work we never need/want this loaded.
+        require 'chef/http/socketless_chef_zero_client'
```
danielsdeleo: maybe add `unless defined?(SocketlessChefZeroClient)` here? I think the perf for a redundant require has improved in recent Ruby but I know it has been an issue in the past.

ranjib:
:+1:

lamont-granquist:
@danielsdeleo went ahead and made it threadsafe as well...

danielsdeleo:
:+1: BTW is that the fix we're gonna go with for now for the rubygems/bundler thing?

lamont-granquist:
it is the fix, but i'll merge the other PR and then drop that commit off this one...

stevendanna:
:+1:

iroller:
Do you guys have a plan when/which version to release chef with this fix merged?

lamont-granquist:
next version.   no plan right now, the delay with 12.5.1 was probably way too long...

martinb3:
:+1:

PR #4133, opened 2015-11-03
SSL and Connection Reset Error Retries

danielsdeleo:
Includes https://github.com/chef/chef/pull/3444 and fixes pull request comments, adds retries for connection resets, and adds tests for retries for the Chef::HTTP base class.

@chef/client-core


lamont-granquist:
good 'nuff :+1:

btm:
:+1:

PR #4134, opened 2015-11-03
update bundler continuously

lamont-granquist:


danielsdeleo:
LGTM :+1:

.travis.yml
```diff
@@ -4,7 +4,7 @@ cache: bundler
 sudo: false
 # Early warning system to catch if Rubygems breaks something
 before_install:
-  gem update --system
+#  gem update --system
```
danielsdeleo: Can we document which versions of things aren't working together so we'll know when we could revert this change?

PR #4135, opened 2015-11-04
windows service uses log file location from config if none is given on commandline

mwrock:
Currently the windows service manager passes a log_location to the windows service at startup. Once the service is installed, this lof_location cannot be changed and overrides any log_location set in the client.rb. This PR allows a windows service to be installed with no log_location passed to the windows service. In such a case, the service will always use what is in the client.rb or the typical default path if none is configured.

### DIFF
--- a/lib/chef/application/windows_service.rb
+++ b/lib/chef/application/windows_service.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Christopher Maier (<maier@lambda.local>)
-# Copyright:: Copyright (c) 2011 Opscode, Inc.
+# Copyright:: Copyright (c) 2011-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -45,8 +45,7 @@ class Chef
       option :log_location,
         :short        => "-L LOGLOCATION",
         :long         => "--logfile LOGLOCATION",
-        :description  => "Set the log file location",
-        :default => "#{ENV['SYSTEMDRIVE']}/chef/client.log"
+        :description  => "Set the log file location"
 
       option :splay,
         :short        => "-s SECONDS",
@@ -59,6 +58,8 @@ class Chef
         :long         => "--interval SECONDS",
         :description  => "Set the number of seconds to wait between chef-client runs",
         :proc         => lambda { |s| s.to_i }
+
+      DEFAULT_LOG_LOCATION ||= "#{ENV['SYSTEMDRIVE']}/chef/client.log"
 
       def service_init
         @service_action_mutex = Mutex.new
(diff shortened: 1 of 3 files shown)


ksubrama:
:+1:

lib/chef/application/windows_service.rb
```diff
@@ -264,6 +266,14 @@ def auto_log_level?
         Chef::Config[:log_level] == :auto
       end
 
+      def resolve_log_location
+        if Chef::Config[:log_location] && Chef::Config[:log_location].is_a?(String)
```
chefsalim: What happens if Chef::Config[:log_location] == STDOUT? (the path in the old code)

lib/chef/application/windows_service.rb
```diff
@@ -60,6 +59,8 @@ class WindowsService < ::Win32::Daemon
         :description  => "Set the number of seconds to wait between chef-client runs",
         :proc         => lambda { |s| s.to_i }
 
+      DEFAULT_LOG_LOCATION = "#{ENV['SYSTEMDRIVE']}/chef/client.log"
```
chefsalim: Some day we really need to rip out hardcoded directory names..

lib/chef/application/windows_service.rb
mwrock: Then it logs to `DEFAULT_LOG_LOCATION` since `STDOUT` fails the `.is_a?(String)` test. I previously was just checking if it was equal to `STDOUT` but the unit tests were fiddling with `STDOUT` and so I just want to make sure it is not an instance of IO.

btm: `# Check if we have a log location and it's not STDOUT`

spec/unit/windows_service_spec.rb
```diff
@@ -58,6 +59,23 @@
       tempfile.unlink
     end
 
+    it "uses the default log location if none is configured" do
+      Chef::Config.merge!({:config_file => "test_config_file", :log_level => :info})
+      expect(instance).to receive(:configure_chef).twice
+      instance.service_init
+      allow(instance).to receive(:running?).and_return(true, false)
+      allow(instance.instance_variable_get(:@service_signal)).to receive(:wait)
+      allow(instance).to receive(:state).and_return(4)
```
btm: can we move some of these stubs up into a new before for the context or the top before block? It'd be easier to see what's unique about this test if we didn't have to compare each line.

lib/chef/application/windows_service.rb
btm: It seems like in the old code we didn't pass -L at all if log_location was set to STDOUT, and now we're passing -L with DEFAULT_LOG_LOCATION if log_location is STDOUT?

mwrock: yeah I explicitly remember starting to add the conditional and then removing it last night but I think keeping it is the right thing. I'll add that.

spec/unit/windows_service_spec.rb
mwrock: Yeah. Good call.  this spec is a mess. I'm cleaning it up.

mwrock:
I basically rewrote the specs.

btm:
:+1: 

Recommend updating the copyrights:

```
windows_service.rb + windows_service_manager.rb:
Copyright:: Copyright (c) 2011-2015 Chef Software, Inc.

windows_service_spec.rb:
Copyright:: Copyright (c) 2013, 2015 Chef Software, Inc.
```

mwrock:
updated copytrights

PR #4137, opened 2015-11-04
don't squash Chef::Config[:verbosity] on subsequent instances of Chef::Knife::Bootstrap

btm:
When something goes wrong in knife it is common to get an error like:

```Exception: NoMethodError: undefined method `empty?' for nil:NilClass```

Without a file and line number, you can't tell even which project this came
from, e.g. core chef or knife-ec2?

The current message suggests collecting -VV output for filing a bug report,
which will now include the backtrace and be incredibly helpful.

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -306,7 +306,7 @@ class Chef
 
       # copy Mixlib::CLI over so that it can be configured in knife.rb
       # config file
-      Chef::Config[:verbosity] = config[:verbosity]
+      Chef::Config[:verbosity] = config[:verbosity] if config[:verbosity]
     end
 
     def parse_options(args)
--- a/spec/unit/knife_spec.rb
+++ b/spec/unit/knife_spec.rb
@@ -63,6 +63,12 @@ describe Chef::Knife do
     Chef::Knife.reset_config_loader!
   end
 
+  it "does not reset Chef::Config[:verbosity to nil if config[:verbosity] is nil" do
+    Chef::Config[:verbosity] = 2
+    Chef::Knife.new
+    expect(Chef::Config[:verbosity]).to eq(2)
+  end
+
   describe "after loading a subcommand" do
     before do
       Chef::Knife.reset_subcommands!
@@ -309,7 +315,7 @@ describe Chef::Knife do
         expect(Chef::Config[:listen]).to be(false)
       end
 
-      context "verbosity is greater than zero" do
+      context "verbosity is one" do
         let(:fake_config) { "/does/not/exist/knife.rb" }
 
         before do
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ranjib** [comment]
:+1:

**Comment by danielsdeleo** [comment]
Um, were you not getting a backtrace with `-VV` before?

**Comment by stevendanna** [comment]
-VV should raise an exception here: https://github.com/chef/chef/pull/4137/files#diff-1220e994288a7f7d9ae83196ce7fad9eR421

before the humanize_exceptions is reached.

**Comment by btm** [comment]
No backtrace. Chef::Config[:verbosity] => '' there. debugging output below:

```
$ knife ec2 server create -I ami-11f9aa74 --bootstrap-url https://omnitruck.chef.io/chef/install.sh --prerelease -x ubuntu -c ~/.chef/knife.ln -VV
in #initialize: Chef::Config[:verbosity]: 2, config[:verbosity]: 2
start of apply_computed_confg: Chef::Config[:verbosity]: 2, Chef::Config[:log_level]: auto
end of apply_computed_confg: Chef::Config[:verbosity]: 2, Chef::Config[:log_level]: debug
[snip]
DEBUG: Looking for key use_sudo_password and found value 
caught undefined method `empty?' for nil:NilClass, verbosity is: ''
This may be a bug in the 'ec2 server create' knife command or plugin
Exception: NoMethodError: undefined method `empty?' for nil:NilClass
```

**Comment by danielsdeleo** [comment]
Seems like `verbosity is: ''` is the bug then.

**Comment by btm** [comment]
Yeah, it's this: https://github.com/chef/knife-ec2/blob/master/lib/chef/knife/ec2_server_create.rb#L716

**Comment by danielsdeleo** [comment]
Is there a reasonable way to fix that? `Chef::Config[:verbosity]` is supposed to be one of `0`, `1`, or `2` and nothing else.

**Comment by btm** [comment]
force pushed a different solution based on further understanding. please re-review.

commit message:
```
We create an instance of Chef::Knife::Bootstrap in knife-ec2, but we've
already processed ARGV so we substitute and process an empty argv.

config[:verbosity] comes from Chef::Application::Knife. Normally we
merge those options with the ones from the knife subcommand in
Chef::Knife.run. Since we don't in this case, we don't set it to the
default of 0 that is specified there, leaving it nil.

We set Chef::Config[:verbosity] to config[:verbosity] in Chef::Knife#initialize
which now becomes nil.

This change makes it so we do not update Chef::Config[:verbosity] if
config[:verbosity] is nil, so that later when we rescue we don't call
humanize_exception if Chef::Config[:verbosity] is 2.
```

stevendanna:
:+1: Seems reasonable to me.

danielsdeleo:
:+1:

PR #4146, opened 2015-11-05
Update chef.gemspec to include pry-byebug

vinyar:


coderanger:
:-1: Just like last time this came up, this belongs in the ChefDK packages, not as a dependency of Chef.

Having `pry` in there at all is weird, it is used as part of the implementation of `chef-shell`, not for debugging.

vinyar:
@danielsdeleo I defer to the experts on this one.

danielsdeleo:
@coderanger `chef-shell` does not use pry, though if I had the time I would rewrite it to be an extension of pry, since IRB is actually pretty hostile to customization and the way that `chef-shell` hooks into chef is pretty fragile.

Anyway, pry is included so people can use it as a debugging tool. It could potentially be moved to omnibus packaging (modulo how we interpret SemVer for dependency removals). I do think you'd want it in the Chef (not ChefDK) package so you could use it on a VM created by test kitchen, for example.

thommay:
I tend to agree that the most appropriate place to do this in the omnibus packaging

lamont-granquist:
:-1:  

pry and pry-byebug are not dependencies of chef, they should both be removed.

PR #4147, opened 2015-11-05
Improve knife unit test isolation

ksubrama:
:construction: I withdraw this PR for now - I got a deeper understanding of what's going on but need to investigate a bit more to figure out how to clean the test suites.

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -77,11 +77,14 @@ class Chef
     def self.reset_config_loader!
       @@chef_config_dir = nil
       @config_loader = nil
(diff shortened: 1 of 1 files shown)


ksubrama:
/cc @chef/client-core 

I'm not super-sure of the implications of resetting only one vs. the other.  I do know that @subcommand_loader does not get reset anywhere at all currently which results in an older loader to be used even when reset_config_loader! is called in spec tests.  This can result in spec/unit/application/knife_spec.rb to interfere with spec/unit/knife_spec.rb (as an example).

Issue #4131, opened 2015-11-03
knife bootstrap fails without run list


btm:
When bootstrapping without a run list (I was testing bootstrapping and didn't need a run) knife fails with an exception:

```
ERROR: knife encountered an unexpected error
This may be a bug in the 'ec2 server create' knife command or plugin
Please collect the output of this command with the `-VV` option before filing a bug report.
Exception: NoMethodError: undefined method `empty?' for nil:NilClass
```

It looks like #policyfile_and_run_list_given? added in #3958 fails on `!config[:run_list].empty?` because `config[:run_list]` isn't set at all.

Backtrace. Ugh. We need to not swallow the backtrace in knife.
```
ERROR: /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife/bootstrap.rb:425:in `policyfile_and_run_list_given?'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife/bootstrap.rb:381:in `validate_options!'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife/bootstrap.rb:328:in `run'
/Users/btm/.chefdk/gem/ruby/2.1.0/gems/knife-ec2-0.12.0/lib/chef/knife/ec2_server_create.rb:529:in `run'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife.rb:405:in `block in run_with_pretty_exceptions'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/local_mode.rb:44:in `with_server_connectivity'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife.rb:404:in `run_with_pretty_exceptions'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/knife.rb:203:in `run'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/lib/chef/application/knife.rb:142:in `run'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.5.1/bin/knife:25:in `<top (required)>'
/usr/local/bin/knife:54:in `load'
[...]

PR #4149, opened 2015-11-06
Handle nil run list option in knife bootstrap

danielsdeleo:
IMO it's a bug in knife-ec2 but it worked before and there could be lots of plugins relying on the behavior so it needs to just work.

Fixes https://github.com/chef/chef/issues/4131

@btm


stevendanna:
:+1:

btm:
:+1:

Fixing in knife-ec2 in https://github.com/chef/knife-ec2/pull/374.

PR #4151, opened 2015-11-07
fix log location resolution in windows service

mwrock:
The changes made here in 20728416579f40c30e0a9c08bdfd86ecd7391cbe had an incorrect assumption about the default `Chef::Config[log_location]`. It assumed `nil` when it is actually `STDOUT`. This PR uses the following logic to determine where the windows service tells `chef-client` to send logging:

* If a location is sent on the command line, that is merged with `Chef::Config` and used for logging
* ELSE If a file path is configured in the config file, that is used
* ELSE if `STDOUT` (the default) is configured, then the `DEFAULT_LOG_LOCATION` path is used
* ELSE we have some other logging class configured (not a file path) like an event viewer and no log_location is sent to chef-client

### DIFF
--- a/lib/chef/application/windows_service.rb
+++ b/lib/chef/application/windows_service.rb
@@ -188,7 +188,11 @@ class Chef
           # Pass config params to the new process
           config_params = " --no-fork"
           config_params += " -c #{Chef::Config[:config_file]}" unless  Chef::Config[:config_file].nil?
-          config_params += " -L #{resolve_log_location}" unless Chef::Config[:log_location] == STDOUT
+          # log_location might be an event logger and if so we cannot pass as a command argument
+          # but shed no tears! If the logger is an event logger, it must have been configured
+          # as such in the config file and chef-client will use that when no arg is passed here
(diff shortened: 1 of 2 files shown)


chefsalim:
:+1:

ksubrama:
I just remembered the @mcquin had to deal with log_location and config defaults and probably has sensible insight into your change.  @mcquin - could you give your old ecosystem friends a quick review?

ksubrama:
@mwrock FYI, it's the comments in https://github.com/chef/ohai/blame/master/lib/ohai/config.rb that reminded me that we have some gazillion log_location options that need to be consolidated and I forget how they all interact with each other.

btm:
Is there a way that we can be clear in the it "passes config params to new process with default options" test that we expect log location to end up being a path even though we set up the test with log_location being STDOUT? 

Otherwise someone might try to "fix" that down the road.

otherwise, makes sense. :+1:

mwrock:
Thanks. yes i edited the "it".

PR #4153, opened 2015-11-11
Require ShellOut before Knife::SSH definition

matschaffer:
I ran into a NameError when trying to run https://github.com/matschaffer/knife-solo/blob/master/test/knife_bootstrap_test.rb#L14 that this fixes.

My suspicion is that under normal operation shell_out gets required via the application or platform. And prior to moving config into its own project it was being required in chef/config.rb

In my tests since I end up loading knife ssh directly without the surrounding application things don't work out.

I'll likely just add a line to my test helper for now but wanted to open this for comment or future improvement.

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+require 'chef/mixin/shell_out'
 require 'chef/knife'
 
 class Chef
@@ -29,7 +30,6 @@ class Chef
         require 'readline'
         require 'chef/exceptions'
         require 'chef/search/query'
-        require 'chef/mixin/shell_out'
         require 'chef/mixin/command'
         require 'chef/util/path_helper'
         require 'mixlib/shellout'


matschaffer:
I was able to bisect this error to 6a4f414666b3 using this this ruby code:

```ruby
require 'chef/knife/bootstrap'

Chef::Knife::Bootstrap.load_deps
```

matschaffer:
It seems partly coincidental, but I'm guessing the workstation config loader was the last bit of config to get moved out to chef-config which is why this change gets marked during bisection.

The caller path to `lib/chef/mixin/shell_out.rb` before 6a4f414 looks like this:
```
❯ ruby -Ilib tester.rb | jq . | grep -v kernel
[
  "/Users/mat/code/chef/lib/chef/provider.rb:24:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/platform/provider_mapping.rb:23:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/platform.rb:21:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/application.rb:27:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/config_fetcher.rb:1:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/workstation_config_loader.rb:19:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/knife.rb:23:in `<top (required)>'",
  "/Users/mat/code/chef/lib/chef/knife/bootstrap.rb:19:in `<top (required)>'",
  "tester.rb:1:in `<main>'"
]
```

This change removes the workstation_config_loader which was linking `Knife::SSH` to it's dependency on `Chef::Mixin::ShellOut` at definition time.

lamont-granquist:
okay :+1: this looks obviously correct.   that dep can't be lazied and we should be explicit rather than relying on a side effect of it being pulled in through chef/knife.

btm:
:+1:

thommay:
:+1:

PR #4165, opened 2015-11-13
Multipackage internal API improvements

lamont-granquist:
This allows a subclass of Chef::Provider::Package to declare package_class_supports_arrays which will cause the superclass to only pass arrays into the child class.  This gets rid of most of the need for the subclass to have `.is_a?(Array)` checks everywhere and just use Arrays internally.

I developed this against #4106, but that got stalled due to unrelated chocolatey.org issues, and want to use this in the Zypper provider to rework #4102 (along with eventually using this in the yum and apt providers, but those are large yaks for another day...)

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@opscode.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Copyright:: Copyright (c) 2008-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -86,11 +86,10 @@ class Chef
           end
         end
 
-        # XXX: mutating the new resource is generally bad
-        @new_resource.version(versions_for_new_resource)
-
         converge_by(install_description) do
-          install_package(package_names_for_targets, versions_for_targets)
+          multipackage_api_adapter(package_names_for_targets, versions_for_targets) do |name, version|
+            install_package(name, version)
+          end
           Chef::Log.info("#{@new_resource} installed #{package_names_for_targets} at #{versions_for_targets}")
         end
       end
@@ -113,11 +112,10 @@ class Chef
           return
         end
 
-        # XXX: mutating the new resource is generally bad
-        @new_resource.version(versions_for_new_resource)
-
         converge_by(upgrade_description) do
-          upgrade_package(package_names_for_targets, versions_for_targets)
+          multipackage_api_adapter(package_names_for_targets, versions_for_targets) do |name, version|
+            upgrade_package(name, version)
+          end
           log_allow_downgrade = allow_downgrade ? '(allow_downgrade)' : ''
           Chef::Log.info("#{@new_resource} upgraded#{log_allow_downgrade} #{package_names_for_targets} to #{versions_for_targets}")
         end
@@ -138,12 +136,13 @@ class Chef
 
       private :upgrade_description
 
-      # @todo: ability to remove an array of packages
       def action_remove
         if removing_package?
           description = @new_resource.version ? "version #{@new_resource.version} of " :  ""
           converge_by("remove #{description}package #{@current_resource.package_name}") do
-            remove_package(@current_resource.package_name, @new_resource.version)
+            multipackage_api_adapter(@current_resource.package_name, @new_resource.version) do |name, version|
+              remove_package(name, version)
+            end
             Chef::Log.info("#{@new_resource} removed")
           end
         else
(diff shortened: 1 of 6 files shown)


lamont-granquist:
@chef/client-core needs some review love

lamont-granquist:
Looks like an unrelated travis error at first glance, smells like we might have picked up a new serverspec dep....

lib/chef/provider/package.rb
```diff
@@ -86,11 +86,10 @@ def action_install
           end
         end
 
-        # XXX: mutating the new resource is generally bad
-        @new_resource.version(versions_for_new_resource)
-
         converge_by(install_description) do
-          install_package(package_names_for_targets, versions_for_targets)
+          with_subclass_array_wrapper(package_names_for_targets, versions_for_targets) do |name, version|
```
jaymzh: I hate the name. s/with_//g ?

jaymzh:
Other than the naming of that wrapper, I love it. So much cleaner.
:+1:

lamont-granquist:
I've seen the `with_` thing used a lot in ruby-land for those kinds of helper wrappers that do stuff and yield...  I don't have particularly strong feelings about it though...

@chef/client-core anyone else up for reviewing?

danielsdeleo:
Haven't reviewed yet, but do we know what's up with the travis failure? Looks unrelated, would a rebase fix, or is master broken?

lib/chef/provider/package.rb
danielsdeleo: Could call it something like `multipackage_api_adapter` or similar?

lib/chef/resource/package.rb
```diff
@@ -31,14 +31,8 @@ class Package < Chef::Resource
...
         super
-        @candidate_version = nil
-        @options = nil
-        @package_name = name
-        @response_file = nil
+        package_name(name) unless name.nil?
         @response_file_variables = Hash.new
-        @source = nil
-        @version = nil
-        @timeout = nil
```
danielsdeleo: Why are these removed?

lamont-granquist: i didn't really see the point to them...

lib/chef/provider/package.rb
lamont-granquist: i kind of like that, `multipackage_api_wrapper` maybe?  and then maybe `use_multipackage_api` for the class method?

lamont-granquist:
I tried a rebase, not sure if master is busted or not..

lib/chef/resource/package.rb
danielsdeleo: One reason is to avoid implicit instance variable creation, which is a ruby warning (some day I would like Chef to run without warnings). The other is as pseudo documentation of which instance variables exist in a class.

lamont-granquist:
yeah, master is busted, bunch of stuff @thommay merged last week was broke when it ran on master, not sure if a merged PR broke stuff or we pulled in a new gem dep...

lib/chef/resource/package.rb
```diff
@@ -31,14 +31,13 @@ class Package < Chef::Resource
...
+        options                 nil
+        package_name            name
+        response_file           nil
+        response_file_variables Hash.new
+        source                  nil
+        version                 nil
+        timeout                 nil
```
lamont-granquist: @danielsdeleo how about this?  the annoying thing is that when people bypass the accessor for non-nil values in the initializer it can cause problems (I specifically need to set the package_name this way in order to be able to subclass and then coerce it into an array--code not shown in this commint).  there was also the case of the log resource when people started calling the log resource with random objects and the initializer bypassed the strict type checking of set_or_return and so stored random non-string classes internally in the ivar.

the problem with what i'm doing though is that i'm still not explicitly setting nil because we can't have nillable properties until Chef 13, and relying on the getter to create the ivar.  that might still throw a warning somewhere under set_or_return, but we could fix it there.

danielsdeleo: That makes pragmatic sense and still gives us the benefit of initializing our state at creation time and documents the properties of the object in a single place, so :+1:

lamont-granquist: Yeah it'll also magically turn into a setter in 13.0 which will DTRT, and if it blows up because the setter doesn't support setting a nil then that's also a clear bug to fix since any property which starts out as nil needs to be nillable so that the user can reset it (so needs NilClass added or whatever).

lib/chef/provider/package.rb
jaymzh: :+1:

thommay: i kind of prefer `multipackage_api_adapter` but eh :)

thommay:
yeah, I'm :+1: modulo the bikeshedding over the function name.

lamont-granquist:
@btm can you look at that last commit 1e74aab428a4e31c00218eab6c1e95e9f49048e0 into the windows_package resource?  I removed some code there and I'm not sure what the intent was.  Since the source attribute gets set to nil in the initializer of the superclass we were never in a state when the instance variable check would have been false, so I think its dead code -- although it just became active (and trolled me) due to the changes I made in this PR to set all the default values through the accessors rather than through hitting the ivars directly...

thommay:
:shipit:

PR #4167, opened 2015-11-13
WMF 5 and Win 10 Threshold 2 Allow dsc_resource with the LCM enabled.

### LINKED ISSUE #4156: dsc_resource doesn't have a return property
We are using the chef dsc_resource to invoke an experimental dsc resource xSharePoint, and when running the pre-req installer it returns an error code which means we need to reboot, we are unable to use the return property like with the chef powershell_script. the error we get is: **undefined method 'returns' for chef::resource::dscresource.**

## Example with powershell_script:
```
powershell_script "install-#{cab}" do
    code <<-EOH
      DISM.exe /Online /Add-Package /PackagePath:#{node['stores-ISP-common']['chef_temp']}\\#{cab} /quiet /norestart
       Del #{node['stores-ISP-common']['chef_temp']}\\#{cab.split('.')[0]}.*
     EOH
     returns [0, 3010]
     notifies :reboot_now, "reboot[reboot-computer]", :immediately if returns 3010
     returns 0
     not_if "wmic qfe | findstr /i '#{cab.split('-')[2]}'"
   end
```
## **Failed Example with dsc_resource:**
```
dsc_resource 'Install_SP' do
   resource :xSPInstallPrereqs
   property :path, "C:\Users\Public\Documents\example.zip"
   returns [0, 30066]
   notifies :reboot_now, "reboot[reboot-computer]", :immediately if returns 30066
   returns 0
 end
```
I wanted to open an issue to see if the returns property should be general functionality to all chef resources, or if we should find a different way to handle our reboots.

smurawski:
Closes #4156 

Adds the reboot_action to dsc_resource

```
dsc_resource 'Install AD' do
  resource :windowsfeature
  property :name, 'ad-domain-services'
  reboot_action :reboot_now
end
```
and 

Allows the LCM to stay enabled for PowerShell builds of 10586 or newer.

### DIFF
--- a/lib/chef/platform/query_helpers.rb
+++ b/lib/chef/platform/query_helpers.rb
@@ -85,10 +85,12 @@ class Chef
       end
 
       def supports_dsc_invoke_resource?(node)
-        require 'rubygems'
         supports_dsc?(node) &&
-          Gem::Version.new(node[:languages][:powershell][:version]) >=
-            Gem::Version.new("5.0.10018.0")
+          supported_powershell_version?(node, "5.0.10018.0")
+      end
+
+      def supports_refresh_mode_enabled?(node)
+        supported_powershell_version?(node, "5.0.10586.0")
       end
 
       def dsc_refresh_mode_disabled?(node)
@@ -97,6 +99,15 @@ class Chef
         metadata = cmdlet.run!.return_value
         metadata['RefreshMode'] == 'Disabled'
       end
+      
+
+      def supported_powershell_version?(node, version_string)
+        return false unless node[:languages] && node[:languages][:powershell]
+        require 'rubygems'
+        Gem::Version.new(node[:languages][:powershell][:version]) >=
+          Gem::Version.new(version_string)
+      end
+
     end
   end
 end
--- a/lib/chef/provider/dsc_resource.rb
+++ b/lib/chef/provider/dsc_resource.rb
@@ -15,40 +15,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 require 'chef/util/powershell/cmdlet'
 require 'chef/util/dsc/local_configuration_manager'
 require 'chef/mixin/powershell_type_coercions'
 require 'chef/util/dsc/resource_store'
-
 class Chef
   class Provider
     class DscResource < Chef::Provider
       include Chef::Mixin::PowershellTypeCoercions
-
       provides :dsc_resource, os: "windows"
-
       def initialize(new_resource, run_context)
         super
         @new_resource = new_resource
         @module_name = new_resource.module_name
+        @reboot_resource = nil
       end
-
       def action_run
         if ! test_resource
           converge_by(generate_description) do
             result = set_resource
+            reboot_if_required
           end
         end
       end
-
       def load_current_resource
       end
-
       def whyrun_supported?
         true
       end
-
       def define_resource_requirements
         requirements.assert(:run) do |a|
           a.assertion { supports_dsc_invoke_resource? }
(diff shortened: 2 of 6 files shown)


adamedx:
@btm, @mwrock, @jaym, I like the fix for lcm, however how much should we really try to support old versions of wmf5? Seems like it might be best to just let things fail and not special case any particular build #. We could print warnings if anyone is running earlier than 10586...

mwrock:
I think eventually it will make sense to remove these but for now it may be that the cost of this pr will be offset by the savings from less confused customers and chef support folks.

mwrock:
I'm :+1:

lib/chef/platform/query_helpers.rb
```diff
@@ -97,6 +99,13 @@ def dsc_refresh_mode_disabled?(node)
         metadata = cmdlet.run!.return_value
         metadata['RefreshMode'] == 'Disabled'
       end
+      
+      def supported_powershell_version?(node, version_string)
+        require 'rubygems'
+        Gem::Version.new(node[:languages][:powershell][:version]) >=
```
btm: This will raise `NoMethodError: undefined method `[]' for nil:NilClass` if the powershell ohai plugin didn't find powershell.exe. I see a couple options here. 

* You could `return false unless node[:languages][:powershell]` because no, we don't support that powershell version, because we have no powershell
* Or you could `raise unless node[:languages][:powershell]` because we're considering powershell things and we don't have powershell.

I think the current code paths wouldn't get here, because you have assertions that lead to supports_dsc?, but it seems easy to lose those or other people might call #supported_powershell_version? without ensuring first that powershell existed.

lib/chef/resource/dsc_resource.rb
```diff
@@ -31,6 +31,7 @@ def initialize(name, run_context)
         super
         @properties = {}
         @resource = nil
+        @allow_reboot = :nothing
```
btm: weird things will happen if someone sets allow_reboot to true or false. I'd expect this property to take boolean and then have another method choose the appropriate action to pass to `@reboot_resource.run_action` in `invoke_reboot`.

althernatively I'd rename the property to reboot_action, but I think that's unnecessarily complicated.

lib/chef/provider/dsc_resource.rb
```diff
@@ -32,12 +32,14 @@ def initialize(new_resource, run_context)
...
         @new_resource = new_resource
         @module_name = new_resource.module_name
+        @reboot_resource = nil
       end
 
       def action_run
         if ! test_resource
           converge_by(generate_description) do
             result = set_resource
+            invoke_reboot
```
btm: this was scary when I first saw it. could this be `reboot_if_required`?

lib/chef/provider/dsc_resource.rb
```diff
@@ -123,18 +130,14 @@ def test_resource
...
-          # WMF April 2015 Preview
-          result.return_value["InDesiredState"]
-        end
+        return_dsc_resource_result(result, "InDesiredState")
       end
 
       def set_resource
         result = invoke_resource(:set)
+        if return_dsc_resource_result(result, 'RebootRequired') 
+          notify_reboot
```
btm: there's no notification happening here, how about `create_reboot_resource`?

lib/chef/provider/dsc_resource.rb
```diff
@@ -155,6 +158,31 @@ def invoke_resource(method, output_format=:object)
...
+          "Reboot for #{@new_resource.name}", 
+          run_context
+        ).tap do |r|
+          r.reason("Reboot for #{@new_resource.resource}.")
+        end
+      end
+
+      def invoke_reboot
+        unless @reboot_resource.nil?
+          Chef::Log.debug("Requesting node reboot with #{@new_resource.allow_reboot}")
```
btm: We may not actually request a reboot a reboot here, depending on `@new_resource.allow_reboot`. Can you put two log messages here, one for if we're going to actually trigger the reboot, and another for if one was requested but we're ignoring it due to the value of allow_reboot?

lib/chef/resource/dsc_resource.rb
btm: I see now you're trying to allow :reboot_now and :request_reboot. I'd probably change this something like reboot_action.

mwrock: yeah. I was initially confused by the booleanish name. `reboot_action` is more clear.

lib/chef/platform/query_helpers.rb
smurawski: I still need to address this one @btm.  Will get there shortly.

jaym:
:+1:

smurawski:
The one travis failure is an intermittent kitchen error.

PR #4173, opened 2015-11-16
Invitations and members

jkeiser:
Fixes bugs in latest local mode pedant:
1. Support for /organizations/NAME/association_requests (invitations.json) and /organizations/NAME/users (members.json) chef-zero endpoints. Because chef-zero stores "X is a member of Y" as a single file, and ChefFS stores "here is a list of Y's members," we have to translate between the two.
2. Fix ACLs for containers.

This is tested by `bundle exec rake cheffs` in chef-zero (separate PR). We will update chef Travis to run cheffs out of there directly as a followon PR.


danielsdeleo:
Code changes look sensible to me, :+1: 

I think you need to rebase to get some test fixes that Lamont committed today.

jkeiser:
Rebased and pushed!

PR #4179, opened 2015-11-17
Get external tests running again

jkeiser:
When `cache: bundler` was added, Travis started running `bundle install --path bundle/vendor`, but bundler does not get installed in that directory and therefore the tests can't run `bundle install`. Adding bundler to the `Gemfile` does not change this, either.

The fix is to stop caching for external tests.


tyler-ball:
:+1:

btm:
:+1:

jkeiser:
Merged as 2d786b7

PR #4182, opened 2015-11-18
Restore rspec 3.4 by setting project_source_dirs

jkeiser:
to be the cookbooks (see http://rspec.info/blog/2015/11/rspec-3-4-has-been-released/, Better Failure Source Detection, for more info)


lamont-granquist:
i don't really have the background to understand the implications, but the code looks right on the surface :+1:

Issue #4117, opened 2015-10-28
Remove the use of method_missing to populate the params hash of definitions


lamont-granquist:
This may go under the heading of definitions needing a complete rethink.

But fundamentally definitions are pretty simple and just a method created on the DSL.  But under the covers, it uses method_missing to populate its params hash:

https://github.com/chef/chef/blob/8d32fdd4377476a9f7dc36a864ccdaa17c32b3a1/lib/chef/resource_definition.rb#L57-L62

This is awful when you accidentally find that you've wound up with a method named `#timeout` polluting your namespace (because of some awful legacy ruby code) and now method_missing never fires when you try to pass a 'timeout' option to a definition:

https://github.com/lamont-cookbooks/multipackage/issues/6

The definitions block needs to be parsed by a real object which has methods on it corresponding to the parameters that the definition wants.  But currently definitions do not declare the parameters that they want to receive.

What we probably want to do is leverage properties and have a definitions v2 API that makes users specify their interface for the definition with properties and then does proper parsing and input validation of the properties.

Since fundamentally a definition is just a method on the DSL it make make more sense to just write some useful mixins or base classes so that people can write methods-on-the-DSL-which-take-blocks-and-use-properties-for-validation.  Don't even call them definitions to avoid the stink associated with that name.

Ideally you just open up a class and do something like:

``` ruby
class Multipackage < Chef::DSL::Extension
  provides :multipackage

  property :package_name, [String, Array], :name_property => true
  property :timeout, [String, Integer]

  code do
    [...stuff...]
  end
end
```

coderanger:
How is this not a resource?

lamont-granquist:
it'll still run at compile-time and doesn't have any actions, and can't be notified or subscribed, its just a method with strong typing to its block arguments.

lamont-granquist:
this'll also be RFC'd, this is a TODO for me, and I'm going to play around with this in a cookbook for awhile and validate that something like i just wrote really does work...

lamont-granquist:
So this could work as a resource but it needs an after_created block to force it to compile time like chef_gem, and then we'd need hooks to be able to break out of the use_inline_resources sub-resource collection and manipulate either the parent or the root resource collection.

lamont-granquist:
nobody should use definitions

Issue #4006, opened 2015-09-30
Property declarations can silently mask DSL methods


lamont-granquist:
With new-style converged resources in 12.5.0:

``` ruby
property :name, String, name_property: true
property :template, String  # <--- this is badly named

action :whatever do
  template "/tmp/#{name}" do
    source "#{name}.erb"
  end
end
```

That is just `new_resource.template = "/tmp/#{name}` with a block that is silently ignored.  It never creates a resource.

Because we dropped the need to do `new_resource.whatever` or `params[:whatever]` and the properties are methods they conflict with the DSL resources which are also methods.

The way that it works is probably fine, and declare_resource or renaming the property works around the issue, but I lost 2 hours debugging this and we should probably try to warn.

coderanger:
The solution I went with in `Poise::Fused` is that inside `action` you get the actual provider context, not the resource. That is kind of gross though :-/

lamont-granquist:
But, I suspect that being able to do `my_property` instead of `new_resource.my_property` is a design goal of the new style resources, @jkeiser?

lamont-granquist:
As far as warning goes, though, we're going to have an issue there as well.  If we're warning when properties are defined, we're still parsing through LWRPs and we don't know if a property will overwrite some resource which is later defined.  We could at least reliably catch properites overwriting core chef resources, which would still be useful.

We could do some kind of horrible hook after parsing LWRPs which walked them all and checked all their properties against the DSL methods and warned about any namespace overlap, but that is mildly awful.

If we track everything we add to the DSL we could check every time we assign to a property, but that is a bit more overhead to do that every time, and it'll emit a warning every time the bad property is used -- we could emit only if a property is passed a block to discover attempted use of a property-as-a-resource, but not all resource declarations have a block so that misses `package "foo"` when you've got package-as-a-property.

jkeiser:
I believe we can make it warn--make it so that when you are declaring an inline resource, if you *read* a property, it warns if it's a dup of the enclosing provider.

jkeiser:
Should warn with https://github.com/chef/chef/pull/4269

jkeiser:
@lamont-granquist oh! There is a way to fix this one that we should really do anyway. If you pass a block to a property and it doesn't need one, we should raise an error.

lamont-granquist:
Yes, that sounds better than all the other approaches I was thinking of.

jkeiser:
[...]

PR #4183, opened 2015-11-18
extract declare_resource to a mixin and extend API

lamont-granquist:
- extract to its own mixin so it can be included without
  method_missing and the rest of the resources DSL
- add ability to inject different run_context
- add ability to create_if_missing into the resource_collection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * [**Maxime Brugidou**](https://github.com/brugidou)
   [pr#4052](https://github.com/chef/chef/pull/4052) Add make_child_entry in ChefFS CookbookSubdir
 
+* [pr#4183](https://github.com/chef/chef/pull/4183) extract declare_resource to a mixin and extend API
 * [pr#4185](https://github.com/chef/chef/pull/4185) dpkg provider cleanup
 * [pr#4165](https://github.com/chef/chef/pull/4165) Multipackage internal API improvements
 * [pr#4081](https://github.com/chef/chef/pull/4081) RFC-037: add `chef_version` and `ohai_version` metadata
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core and @thommay @coderanger and @jkeiser in particular, plus maybe @danielsdeleo 

want to get thoughts on this.  i'll write specs if everyone is okay with it.

this is something that i've discussed with john before, which is pulling out declare_resource and build_resource into their own mixin so that classes could include them without pulling in the whole recipe DSL.   i've wanted this in my own classes that i've written.  i've also strongly considered adding this mixin to Chef::Resource and Chef::Provider directly (which i think was the direction dan wanted to take declare_resource and build_resource eventually).

by mixing this into Chef::Resource that is also one way to work around #4006 when you have a property called 'template' that overrides using the template resource in the DSL.

plus, there's the extensions to build_resource/declare_resource to better support writing accumulator patterns in resources instead of definitions.  this addresses #4117.  i was going to include a pointer to the multipackage resource that used this code, but i appear to not have pushed that remotely and my computer at home is powered down right now due to the windstorm power outage this AM...

comments?  suggestions?  better name for 'create_if_missing'?

danielsdeleo:
> which i think was the direction dan wanted to take declare_resource and build_resource eventually

Mostly I wanted to make the method_missing shim as thin as possible, and expose the functionality so people could do whatever crazy things they wanted with recipe macros in a more responsible way.

That said, I do like the idea of being able to use a thing that's in between the full recipe DSL (which has the problems you describe, things like `template` and `file` get mixed in) and having to do fully manual resource instantiation (this would be less of a pain if the resource constructors took blocks, though now we have a whole new layer of indirection there, which would be nice to wrap up).

thommay:
Needs tests for the new bits

lamont-granquist:
@chef/client-core new bits now have tests

tyler-ball:
:+1:

PR #4185, opened 2015-11-19
dpkg provider cleanup

lamont-granquist:
- :update and :install are now treated the same way and throw the
  same exceptions
- :remove and :purge don't require the source at all, so don't do
  any checking on that
- fix some convoluted side-effecty logic in load_current_resource
- load_current_resource now correctly gets the dpkg state on
  :remove and :purge when the file does not exist (pretty sure
  the old logic did not)
- fixed the FIXME about using en_US.UTF-8 (the default for shell_out!)
- just use shell_out! to throw exceptions
- clean up all the specs and remove all the instance vars from the code

### DIFF
--- a/lib/chef/provider/package/dpkg.rb
+++ b/lib/chef/provider/package/dpkg.rb
@@ -34,83 +34,61 @@ class Chef
 
         def define_resource_requirements
           super
-          requirements.assert(:install) do |a|
-            a.assertion{ not @new_resource.source.nil? }
-            a.failure_message Chef::Exceptions::Package, "Source for package #{@new_resource.name} required for action install"
+
+          requirements.assert(:install, :upgrade) do |a|
+            a.assertion { !new_resource.source.nil? }
+            a.failure_message Chef::Exceptions::Package, "#{new_resource} the source property is required for action :install or :upgrade"
           end
 
-          # TODO this was originally written for any action in which .source is provided
-          # but would it make more sense to only look at source if the action is :install?
-          requirements.assert(:all_actions) do |a|
-            a.assertion { @source_exists }
-            a.failure_message Chef::Exceptions::Package, "Package #{@new_resource.name} not found: #{@new_resource.source}"
-            a.whyrun "Assuming it would have been previously downloaded."
+          requirements.assert(:install, :upgrade) do |a|
+            a.assertion { source_file_exist? }
+            a.failure_message Chef::Exceptions::Package, "#{new_resource} source file does not exist: #{new_resource.source}"
+            a.whyrun "Assuming it would have been previously created."
           end
         end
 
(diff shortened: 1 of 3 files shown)


lamont-granquist:
@chef/client-core more review

lamont-granquist:
travis failure is unrelated: https://github.com/chef/chef-server/pull/624

thommay:
this looks just peachy :+1:

lib/chef/provider/package/dpkg.rb
```diff
@@ -119,22 +97,65 @@ def upgrade_package(name, version)
...
+        end
+
         # Runs command via shell_out_with_timeout with magic environment to disable
         # interactive prompts. Command is run with default localization rather
         # than forcing locale to "C", so command output may not be stable.
-        #
-        # FIXME: This should be "LC_ALL" => "en_US.UTF-8" in order to stabilize the output and get UTF-8
         def run_noninteractive(command)
-          shell_out_with_timeout!(command, :env => { "DEBIAN_FRONTEND" => "noninteractive", "LC_ALL" => nil })
+          shell_out_with_timeout!(command, :env => { "DEBIAN_FRONTEND" => "noninteractive" })
```
btm: why was it originally left as nil? was it a parsing issue on systems with foreign character sets?

lamont-granquist: It seems to have been in response to this bug:

https://tickets.opscode.com/browse/CHEF-281

The root cause of that bug however was using the "en_US" locale which didn't exist on systems without en_US installed.  We've got complicated workarounds now which favor all kinds of different en_US locales and then fallback to to the 'C' locale as default.

The commit which introduced it just says:

Author: Diego Algorta <diego@oboxodo.com>
Date:   Fri Aug 14 19:54:31 2009 -0300

    Make sure all Package providers run with the system's locale and not C locale.
    
    Surely people want they're packages to corrently detect the system's
    locale when configuring themselves.

And I disagree completely with that logic.  If that logic was right, then if you installed a package with your machine set to en_US.UTF-8 and then changed your locale to de_DE.UTF-8 you would have to reinstall all your packages to get the "locale detection on package installation" right.  Packages can't have a requirement to 'detect' the locale on installation or else locale switching would be completely broken.

lamont-granquist: At some point we also tore it out of all the rest of the package providers, including the apt provider so that now they call shell_out_with_timeout! which calls shell_out! and this is the only one that still overrides its LC_ALL setting.  Pretty certain this line of code is just a cargo cult now.  If it wasn't then the changes introduced in the apt provider would have broken the world a long time ago.

btm:
:+1:

PR #4188, opened 2015-11-19
Update dsc_resource to use verbose stream output

chefsalim:
This fixes a regression in the Nov 2015 update of Windows 10 where we are not putting any output in the converge description.  This is because the Nov 2015 update has started using the verbose stream instead of stdout. This is actually a good thing, as the use of stdout was a hack.  But now we need to check for both conditions.

@btm @adamedx @smurawski @jaym

### DIFF
--- a/lib/chef/provider/dsc_resource.rb
+++ b/lib/chef/provider/dsc_resource.rb
@@ -122,10 +122,15 @@ class Chef
 
       def test_resource
         result = invoke_resource(:test)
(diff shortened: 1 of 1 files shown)


chefsalim:
Note: AppVeyor just seems to be in a bad state not related to this change.

smurawski:
:+1:  lgtm and matches my playing around yesterday.

mwrock:
:+1:

btm:
:+1:

Issue #4194, opened 2015-11-20
Chef-solo is broken on windows


youngjl1:
Here's the offending commit:
https://github.com/chef/chef/commit/2b02ca90ee35031b0735a7f952ae46177875010f#commitcomment-14536316

The problem is that PathHelper.escape_glob() method calls cleanpath() which converts forward slashes to backslashes. This is normally what you want for windows but in this particular case the path is being passed into Dir.glob as an argument, and according to that method's documentation: "Note that this means you cannot use backslash in windows as part of a glob"[1] 

This the source of the bug. The call to Dir.glob returns nothing and you get "Data Bag Items must contain a Hash or Mash." This bug is affecting people [2][3].

[1] http://ruby-doc.org/core-1.9.3/Dir.html
[2] http://stackoverflow.com/questions/28324588/data-bag-items-must-contain-a-hash-or-mash-error
[3] http://stackoverflow.com/questions/28287611/chef-vagrant-encrypted-data-bags-windows

PR #4195, opened 2015-11-20
windows chef-solo fix

youngjl1:
Issue: https://github.com/chef/chef/issues/4194

### DIFF
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -141,6 +141,7 @@ module ChefConfig
       canonical_path(path1) == canonical_path(path2)
     end
 
+    # Note: this method is deprecated. Please use escape_glob_dirs
     # Paths which may contain glob-reserved characters need
     # to be escaped before globbing can be done.
     # http://stackoverflow.com/questions/14127343
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -149,6 +149,13 @@ module ChefConfig
       path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
     end
 
+	# This function does not switch to backslashes for windows
+	# This is because only forwardslashes should be used with dir (even for windows)
+    def self.escape_glob_dir(*parts)
+      path = Pathname.new(join(*parts)).cleanpath.to_s
+      path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
+    end	
+	
     def self.relative_path_from(from, to)
       Pathname.new(cleanpath(to)).relative_path_from(Pathname.new(cleanpath(from)))
     end
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -149,13 +149,13 @@ module ChefConfig
       path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
     end
 
-	# This function does not switch to backslashes for windows
-	# This is because only forwardslashes should be used with dir (even for windows)
+    # This function does not switch to backslashes for windows
+    # This is because only forwardslashes should be used with dir (even for windows)
     def self.escape_glob_dir(*parts)
       path = Pathname.new(join(*parts)).cleanpath.to_s
       path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
-    end	
-	
+    end
+
     def self.relative_path_from(from, to)
       Pathname.new(cleanpath(to)).relative_path_from(Pathname.new(cleanpath(from)))
     end
--- a/chef-config/spec/unit/path_helper_spec.rb
+++ b/chef-config/spec/unit/path_helper_spec.rb
@@ -267,6 +267,23 @@ RSpec.describe ChefConfig::PathHelper do
     end
   end
 
+  describe "escape_glob_dir" do
+    it "escapes characters reserved by glob" do
+      path = "C:\\this\\*path\\[needs]\\escaping?"
+      escaped_path = "C:\\\\this\\\\\\*path\\\\\\[needs\\]\\\\escaping\\?"
+      expect(path_helper.escape_glob_dir(path)).to eq(escaped_path)
+    end
+
+    context "when given more than one argument" do
+      it "joins, cleanpaths, and escapes characters reserved by glob" do
+        args = ["this/*path", "[needs]", "escaping?"]
+        escaped_path = "this/\\*path/\\[needs\\]/escaping\\?"
+        expect(path_helper).to receive(:join).with(*args).and_call_original
+        expect(path_helper.escape_glob_dir(*args)).to eq(escaped_path)
+      end
+    end
+  end  
+
   describe "all_homes" do
     before do
       stub_const('ENV', env)
--- a/lib/chef/data_bag.rb
+++ b/lib/chef/data_bag.rb
@@ -92,8 +92,12 @@ class Chef
           unless File.directory?(path)
             raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' is invalid"
           end
-
-          names += Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(path), "*")).map{|f|File.basename(f)}.sort
+	  temp_path = Chef::Util::PathHelper.escape_glob(path)
+	  if ChefConfig.windows?
+	    #Have to convert backslashes to forward for Dir.glob to work 	  
+	    temp_path = temp_path.gsub('\\','/')	
+	  end
+          names += Dir.glob(File.join(temp_path, "*")).map{|f|File.basename(f)}.sort
         end
         names.inject({}) {|h, n| h[n] = n; h}
       else
@@ -118,8 +122,12 @@ class Chef
           unless File.directory?(path)
             raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' is invalid"
           end
-
-          Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(path, name.to_s), "*.json")).inject({}) do |bag, f|
+	  temp_path = Chef::Util::PathHelper.escape_glob(path, name.to_s)
+	  if ChefConfig.windows?
+	    #Have to convert backslashes to forward for Dir.glob to work 	  
+	    temp_path = temp_path.gsub('\\','/')	
+	  end
+          Dir.glob(File.join(temp_path, "*.json")).inject({}) do |bag, f|
             item = Chef::JSONCompat.from_json(IO.read(f))
 
             # Check if we have multiple items with similar names (ids) and raise if their content differs
--- a/lib/chef/data_bag.rb
+++ b/lib/chef/data_bag.rb
@@ -92,8 +92,8 @@ class Chef
           unless File.directory?(path)
             raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' is invalid"
           end
-		  
-          names += Dir.glob(File.join(Chef::Util::PathHelper.escape_glob_dir(path), "*")).map{|f|File.basename(f)}.sort		
+            
+          names += Dir.glob(File.join(Chef::Util::PathHelper.escape_glob_dir(path), "*")).map{|f|File.basename(f)}.sort
         end
         names.inject({}) {|h, n| h[n] = n; h}
       else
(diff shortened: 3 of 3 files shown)


jaym:
Perhaps `escape_glob` should know about this. The fix is repeated in 2 places, and looking through the chef code base, this problem should be present in many places, for example:
https://github.com/chef/chef/blob/8d32fdd4377476a9f7dc36a864ccdaa17c32b3a1/lib/chef/knife/core/gem_glob_loader.rb#L50

Thoughts @chef/client-core @chef/client-windows

lamont-granquist:
I'm feeling deja vu.  I think @mcquin fought with this before.  I suspect its a previously solved problem in Chef::Utils helpers.

jaym:
```
[3] pry(#<Chef::Knife::SubcommandLoader>)> Dir["C:\\*"]
=> []
[4] pry(#<Chef::Knife::SubcommandLoader>)> Dir["C:\\\\*"]
=> []
[5] pry(#<Chef::Knife::SubcommandLoader>)> Dir["C:\\\\/*"]
=> ["C:\\/$Recycle.Bin",...]
[7] pry(#<Chef::Knife::SubcommandLoader>)> Dir["C:\\/*"]
=> []
[8] pry(#<Chef::Knife::SubcommandLoader>)> Dir["C:\\//*"]
=> ["C://$Recycle.Bin",...]
```
this is weird

youngjl1:
I think it could be handled in escape_glob, since it seems like that method is only supposed to be called when dealing with Dir.glob.

lamont-granquist:
Okay, so I thought escape_glob actually did this?  I've got tons of deja vu and I swear this was an already solved problem.  Did someone accidentally back out a fix somewhere or something?

lamont-granquist:
yeah, so the problem here is File.join isn't it, which defeats the code in escape_glob?

does this work?

```
Dir.glob(Chef::Util::PathHelper.escape_glob(path) + "*")
```

lamont-granquist:
So, i think that's wrong and File.join() does the right thing here and only uses '/' to join which is what Dir.glob wants?  So the problem is the original input `path` is bad?

What we want is the inverse of PathHelper.cleanpath then that converts backslashes to slashes (and probably only on Windows, being careful to think through possible edge cases with backslashes meaning escape sequences?).  Then escape_glob should probably always use that function.

And I suspect that you want to compose the functions so that escape_glob takes the output of the inverse-cleanpath function, rather than the way this is written to take the double-backslashes outputted by escape_glob and turn them into slashes.  The latter approach would be tightly coupled to escape_glob while the former could be reusable elsewhere...

youngjl1:
I think the problem is escape_glob is changing the path to backslashes, and Dir specifically needs forward slashes, even for Windows.

    chef (12.4.1)> path = "chef/bin/*"
     => "chef/bin/*"
    chef (12.4.1)> Dir[path]
     => ["chef/bin/chef-apply", "chef/bin/chef-client", "chef/bin/chef-service-manager", "chef/bin/chef-shell", "chef/bin/chef-solo", "chef/bin/chef-windows-service", "chef/bin/knife"]
    chef (12.4.1)> Chef::Util::PathHelper.escape_glob(path)
     => "chef\\\\bin\\\\\\*"
    chef (12.4.1)> Dir[Chef::Util::PathHelper.escape_glob(path)]
     => []

youngjl1:
@lamont-granquist sounds good to me. I can modify my PR to include the inverse-cleanpath function you proposed.

lamont-granquist:
So what you show escape_glob doing there makes no sense to me...  If it really doing that, it should probably stop...

And probably name it something better than "inverse-clearpath" if you do add that function...

lamont-granquist:
Oh I see, it explicitly cleanpaths()'s first...

lamont-granquist:
So did some historical digging, you need to make sure this kind of thing doesn't get broken:

https://github.com/chef/chef/pull/2087#issuecomment-56124376

The addition of cleanpath in escape_glob is looking kind of wrong to me, and I'm real confused as to why we accepted that patch, but also confused as to why it hasn't caused incredible amount of chaos and didn't immediately break the entire universe...

jaym:
@lamont-granquist i think the reason is because sometimes thing just happen to work with backslash if you have the right amount of forward slashes mixed in.

lamont-granquist:
yeah but cleanpath 'cleans' all the forward slashes to backslashes and then escape_glob escapes them all, and it seems like that works 'most' of the time...

@youngjl1 your example is also broken there...

```
Chef::Util::PathHelper.escape_glob("chef/bin/*")
```

That results in the '*' getting escaped, so there's no glob and its looking for a literal filename named '*' in the chef/bin directory.

```
Dir[Chef::Util::PathHelper.escape_glob("chef/bin") + "/*"]
```

Is how you have to write it.  What is the result out of that?

lamont-granquist:
I think what we need is to know is *exactly* what the contents of 'path' is in the failing case in question.

youngjl1:
The exact contents of the path in the case where it's failing is "/tmp/vagrant-chef-3/chef-solo-3/data_bags"

Looking at it again I'm pretty sure that the thing it is failing on is the leading forward-slash. Example:
    chef (12.4.1)> Dir[Chef::Util::PathHelper.escape_glob("/tmp/") + "/*.ps1"]
     => []
    chef (12.4.1)> Dir[Chef::Util::PathHelper.escape_glob("tmp/") + "/*.ps1"]
     => ["tmp/vagrant-elevated-shell.ps1"]
    chef (12.4.1)> Dir["/tmp/" + "/*.ps1"]
     => ["/tmp//vagrant-elevated-shell.ps1"]
    chef (12.4.1)>

Notice how removing the leading forward slash fixes it. Also, here's the results of escape_glob
    chef (12.4.1)> Chef::Util::PathHelper.escape_glob("/tmp/")
     => "\\\\tmp"
    chef (12.4.1)> Chef::Util::PathHelper.escape_glob("tmp/")
     => "tmp"

lamont-granquist:
so it seems like producing cleanpath'd paths out of escape_glob is a feature of the API though.  if we just convert all the backwhacks to forward slashes then we'll break any code which expects clean paths and will force those callers to change (probably via bug reports that we broke stuff) and include explicit calls to cleanpath.

youngjl1:
so maybe we should create a wrapper function for Dir that handles things correctly?

btm:
> if we just convert all the backwhacks to forward slashes then we'll break any code which expects clean paths

It looks like everywhere that we use escape_glob we're passing it to Dir[] or Dir.glob. Is there a case where we are using escape_glob that works correctly on Windows? It must be working based on relative paths where it's working, based on @youngjl1's results? 

[Dir.glob](http://ruby-doc.org/core-2.2.0/Dir.html#method-c-glob) explicitly says that backslashes are special escape characters so windows paths need to use forward slashes. Can we just change `escape_glob` to use `Pathname.new.cleanpath` instead of `PathHelper.cleanpath` which does the additonal conversion?

```
chef (12.5.1)> def self.escape_glob_new(*parts)
chef ?>   path = Pathname.new(join(*parts)).cleanpath.to_s
chef ?>   path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
chef ?> end
 => nil
chef (12.5.1)> Dir[escape_glob_new("/temp") + "/*.ps1"]
 => ["/temp/foo.ps1"]
chef (12.5.1)>
```

lamont-granquist:
So @btm the concern that I've got is that you've changed the API of what `Dir[]` returns.  What you need to test is that everywhere that we use `Dir[escape_glob()]` that none of the subsequent code is relying on getting back 'cleanpath'd paths and can handle either cleanpath'd or not.   I suspect we put that in there for a reason, though, and that at least some of the use cases will require cleanpath, so you'll probably break some code.

One idea is that we could write an `escape_glob2()` helper which did the inverse-cleanpath instead of the cleanpath.  Then we convert code over to use that new API.   Otherwise it'll be a big patch to update everything and test everything.

I also don't see how we can write a straight forwards general purpose wrapper.  We might want to write a wrapper function for the common case where the last path item is not escape-glob'd but have that annoying issue where most of the path needs to be escape-glob'd only the real globs should not be.  It may be better just to continue with the current shape of the API since it keeps things more explicit and less magical.

btm:
> What you need to test is that everywhere that we use Dir[escape_glob()] that none of the subsequent code is relying on getting back 'cleanpath'd paths and can handle either cleanpath'd or not.

Can you specific if you're referring to `Pathname.cleanpath` or `PathHelper.cleanpath` or both?

If we change `PathHelper.escape_glob` as in my example above in https://github.com/chef/chef/pull/4195#issuecomment-161030437, the only code path that effectively changes is on Windows, where `PathHelper.cleanpath` does the gsub converting slashes to backslashes.. We'd still be providing "cleanpath'd" paths.

If we only use `PathHelper.escape_glob` to input to `Dir.glob` (or `Dir[]`), where backslashes as a path separator are explicitly not supported, then I'd say removing that behavior on Windows is fixing a bug, not changing a functional part of the API.

lamont-granquist:
So I guess I don't quite understand because your example above shows forward slashes coming out of the API:

```
chef (12.5.1)> Dir[escape_glob_new("/temp") + "/*.ps1"]
 => ["/temp/foo.ps1"]
chef (12.5.1)>
```

Which is what I'm saying would be an API break.  Currently I believe those are all backwhacks?

Although are we mostly using File.join on escape_glob mostly everywhere?  Because I think that also converts to forward slashes if I remember correctly?

lamont-granquist:
It definitely joins using forward slashes.

lamont-granquist:
So the issue is that something like this:

```
Dir[Chef::Util::PathHelper.escape_glob(prefix, ".#{@new_resource.path}") + ".chef-*"]
```

That usage should have only backwhacks (double-escaped backwhacks so that the glob 'works' outside of these weird edge cases.  If you change everything to forward slashes that /might/ break.

With these uses, forward slashes are clearly not something the subsequent code cares about because they use File.join:

```
Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(@chef_config[:trusted_certs_dir]), "*.{crt,pem}"))
Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(path), "**", "**"))
```

So the new code almost certainly would not break those.

lamont-granquist:
We might be having confusion because it looks like Pathname#cleanpath and Chef::Util::PathHelper.cleanpath are practically inverses of each other?

lamont-granquist:
Yeah so what we need to do:

1.  Leave escape_glob alone
2.  Create escape_glob2 (or a better name) which produces only forward slashes
3.  Where the caller does something like Dir[File.join(escape_glob())] those can all be converted to escape_glob2 since the inner loop does not care about forwards or backwards slashes.
4.  Where the caller cleanly does something like Dir[escape_glob()] then we must either:
    a.  leave it alone
    b.  convert to escape_glob2 and add a Chef::Util::PathHelper.cleanpath inside the loop
    c.  determine if there are really any Win32 API calls or cmd.exe calls inside the loop which require
         conversion of slashes to backslashes

Anything else risks regressions.

lamont-granquist:
So I would argue this PR as it stands is very definitively incorrect.  It either needs to get updated per the above, or should probably be closed.

youngjl1:
I will try to update it tonight with the above proposed changes

youngjl1:
FYI, I've tested the new changes by running chef-solo, and it no longer fails.

chef-config/lib/chef-config/path_helper.rb
```diff
@@ -149,6 +149,13 @@ def self.escape_glob(*parts)
       path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
     end
 
+	# This function does not switch to backslashes for windows
+	# This is because only forwardslashes should be used with dir (even for windows)
```
lamont-granquist: need to knock off 2 leading spaces here.

more importantly it would be good to note that escape_glob was deprecated with a comment above that function, i think it is always wrong now.

lamont-granquist:
code looks correct, i'm not super wild about the name of the new function but i don't have any better ideas...

it would probably be good to add unit tests for the new function to make sure it doesn't regress anything, there should be unit tests for the old function which can be straight up copied and adapted.

lamont-granquist:
so needs:

- fix the indentation
- add some spec tests

coderanger:
Having multiple copies of the same logic with extremely subtle changes seems like a bad idea. We usually handle the backslash conversions at the edge I thought?

youngjl1:
I added spec tests and fixed indentation

lamont-granquist:
@coderanger we previously were converting forward slashes to backwhacks before the globbing in order to (presumably?) provide backwhacks to later code inside the block that expected backwhacks.  Although clearly from the code most of our usage of this doesn't require backwhacks inside the later block.

The problem is that this breaks globbing completely for certain use cases on windows, so we need to invert that and convert all the backwhacks to forward slashes.  That must happen to get the globbing right.  However, we can't just convert everyone mechanically all at once since some consumers of the previous API might have really needed backwhacks for later code inside the block.

So the former version is totally deprecated and we should convert everything to the new version, but need to be careful as we do so.  There's also no way to eliminate the code duplication as the two APIs are totally incompatible.

youngjl1:
Is there anything else for this PR at this point? Can it be merged?

thommay:
I think we're good to go here :+1:

airhorns:
Would be great to get this merged, spending lots of time trying to get ohai working on ruby 2.2 only to discover this :)

thommay:
Closing since #4569 replaced this.

PR #4196, opened 2015-11-21
multipackage dpkg_package and bonus fixes

lamont-granquist:
dpkg multipackage and bonus fixes
    
    - multipackages dpkg_package
    - fixes edge conditions in load-current-resource around purging/removing
      packages that threw errors before
    - fixes the ability to purge packages that have been removed
    - adds a lot of functional tests for dpkg_package


lamont-granquist:
well that's annoying, coercing to arrays in the resource is going to break every chefspec matcher against dpkg_package out there....    fuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu....

lamont-granquist:
@chef/client-core this is functionally correct and want to get it merged in the current state.

i'm not finished, and i know i need to push more methods back up into the package provider, and will probably massage some of the names of some of the helper functions in the process.

thommay:
:+1:

lamont-granquist:
@chef/client-core anyone else?  bueller?

lib/chef/mixin/subclass_directive.rb
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Mixin
+    module SubclassDirective
+      def self.included(base)
+        base.extend(ClassMethods)
```
danielsdeleo: Since this doesn't do anything, why not get rid of the extra stuff and have users of this module just pull it in via `extend` (like `Forwardable`)?

lamont-granquist: Yep, that was due to an initial quick sketch that got refactored, fixed.

lib/chef/provider/package.rb
```diff
@@ -44,6 +51,8 @@ def whyrun_supported?
       end
 
       def check_resource_semantics!
+        # FIXME: this is not universally true and subclasses are needing to override this and no-ops it.  It should be turned into
+        # another "subclass_directive" and the apt and yum providers should declare that they need this behavior.
```
jaym: I like this idea. I was reading the dpkg code and was thinking "looks good, looks good, looks good...", until i saw `def check_resource_semantics!` and it made me sad. Don't know why, it just did.

jaym:
LGTM. The one thing I would suggest is a quick unit test for subclass directive, but :+1: anyway

lib/chef/provider/package.rb
lamont-granquist: yet another violation of Liskov Substitution Principle in our providers.

spec/unit/provider/package/dpkg_spec.rb
```diff
@@ -52,8 +51,8 @@
   end
 
   before(:each) do
-    allow(provider).to receive(:shell_out!).with("dpkg-deb -W #{new_resource.source}", timeout: 900).and_return(dpkg_deb_status)
-    allow(provider).to receive(:shell_out).with("dpkg -s #{package}", timeout: 900).and_return(dpkg_s_status)
+    allow(provider).to receive(:shell_out!).with("dpkg-deb -W #{source}", timeout: 900).and_return(dpkg_deb_status)
```
danielsdeleo: Not strictly necessary, but I like to put something like `expect(provider).to_not receive(:shell_out)` at the top if you're adding multiple `allow()`s, this way any errant shell out call that you didn't explicitly stub will immediately blow up.

lamont-granquist: That actually blows up on me in a way that makes it look like the allow()'s aren't working...

And if I just add this line at the end of the before:

```
provider.shell_out!("invalid stuff")
```

it already blows up with:

```
        #<Chef::Provider::Package::Dpkg:0x007faf4ac51038> received :shell_out! with unexpected arguments
          expected: ("dpkg-deb -W /tmp/wget_1.11.4-1ubuntu1_amd64.deb", {:timeout=>900})
               got: ("invalid stuff")
```

lamont-granquist: I think there's some kind of odd interaction with expects always overriding allows in the new syntax which didn't happen in the old should/stub syntax...

lamont-granquist: Yeah I think myron explains it here:

https://github.com/rspec/rspec-mocks/issues/557#issuecomment-34338814

expect statements MUST be satisfied and they are not overridable like allows are, so dropping `expect(provider).not_to receive(:shell_out!)` means it'll always blow up if you use shell_out! and any later allow/expects can't override that...

danielsdeleo: Maybe I was thinking of some other trick, but now I can't find where I used it :disappointed:

lamont-granquist: Yeah, I think I know what you're thinking of but I can't put my finger on it either...

tyler-ball: I know you can do the opposite - you can have a broad `allow(thing).to receive(:foo)` and then narrow in with specific expects, EG `expect(thing).to_not receive(:foo).with('bar')` or `expect(thing).to receive(:foo).exactly(1000).times`

spec/unit/provider/package_spec.rb
```diff
@@ -453,17 +453,9 @@ class MyPackageProvider < Chef::Provider::Package
...
-    MyPackageProvider.supports_arrays = true
-  end
-
-  it "has supports_arrays? methods on the class and instance" do
-    expect(MyPackageProvider.supports_arrays?).to be true
-    expect(provider.supports_arrays?).to be true
+  it "has use_multipackage_api? methods on the class and instance" do
+    expect(MyPackageProvider.use_multipackage_api?).to be true
+    expect(provider.use_multipackage_api?).to be true
   end
```
tyler-ball: Can you add
```ruby
it "uses a_to_s to convert an array of strings to a single string" do
  expect(provider.a_to_s("a", nil, "b", "", "c", " ", "d e", "f-g")).to eq("a b c  d e f-g"))
end
```

lib/chef/provider/package/dpkg.rb
```diff
@@ -126,34 +137,124 @@ def get_current_version
...
+        # interactive prompts.
+        def run_noninteractive(*command)
+          shell_out_with_timeout!(a_to_s(*command), :env => { "DEBIAN_FRONTEND" => "noninteractive" })
+        end
+
+        # Returns true if all sources exist.  Returns false if any do not, or if no
+        # sources were specified.
+        #
+        # @return [Boolean] True if all sources exist
+        def source_files_exist?
```
tyler-ball: If someone subclasses `Dpkg`, are we worried about breaking them by renaming this?  Or is that subclassing not supported as part of our public facing API?

tyler-ball:
:+1:

spec/unit/provider/package/dpkg_spec.rb
danielsdeleo: Ok, found what I was looking for, here: https://github.com/chef/chef/blob/de1f684f415faa54599c6b3abbe211d64a319aa6/spec/unit/provider/package/rpm_spec.rb#L43-L49

I think the case I was interested in was that in some example groups, `shell_out` would be stubbed, and in others it would not. In the example I linked, that was also true for `File.exists?`. Which meant that you could accidentally leave out the stubs for `File.exists?` when the file didn't exist or when a command would fail and the tests would pass until you ran them on a different platform or added a conflicting stub. Anyway, that's not relevant to this code since we're putting at least one method stub at the top level.

lib/chef/provider/package/dpkg.rb
lamont-granquist: I think the rest of this patch is going to cause them bigger issues when install_package suddenly starts seeing Arrays, and this will be the least of their issues with this whole patch.

And subclassing is inherently not a public API.  That is why people should use composition and use the public API of the dpkg_package resource directly instead of inheriting and overriding.

PR #4198, opened 2015-11-23
Add proxy test to Travis

jkeiser:


.travis.yml
```diff
@@ -121,6 +121,25 @@ matrix:
...
+      - sudo apt-get -y install squid3 git
+    env: PROXY_TESTS_DIR=/tmp/proxy_tests
+    script:
+      - which chef-client
+      - chef-client --version
+      - git clone https://github.com/chef/proxy_tests.git
+      - cd proxy_tests
+      - chef-client -z -o proxy_tests::render
+      #- sh /tmp/proxy_tests/setup.sh
+      - sudo -E bash /tmp/proxy_tests/run_tests.sh chef_client \* \* /tmp/out.txt
```
tyler-ball: is there some way we could make this use the chef_client from the current checkin?  That seems the only advantage of embedding these tests into the `chef` repo rather than running them from the `proxy_tests` repo

jkeiser:
@tyler-ball I updated the tests to run from the bundle, which appears to be working :) Mind taking a re-look?

adamedx:
:+1:

PR #4200, opened 2015-11-23
Prevent inspect on PsCredential from printing out plain text password

jaym:
See the following use of a DSC resource -- note that it dumps out the password (which itself was not obvious in the code because an environment variable was used to obtain it). On failure (failure was expected in this case), the password, in this case '\\sea-adamed1', is displayed on the screen.

```
    Resource Declaration:
    ---------------------
    # In C:\test\chefdsc\ps5dscuser.rb

      4: dsc_resource 'Create Test User' do
      5:   resource :user
      6:   #  property :username, 'testuserps5'
      7:   property :username, 'testuserps5__\\'
      8:   property :password, ps_credential(ENV['COMPUTERNAME'])
      9:   property :ensure, 'Present'
     10: end
     11:

    Compiled Resource:
    ------------------
    # Declared in C:\test\chefdsc\ps5dscuser.rb:4:in `run_chef_recipe'

    dsc_resource("Create Test User") do
      action [:run]
      retries 0
      retry_delay 2
      properties {:username=>"testuserps5__\\", :password=>#<Chef::Util::Powershell::PSCredential:0x75c5088 @username="placeholder", @password="SEA-ADAMED1">, :ensure=>"Present"}
      resource :user
      declared_type :dsc_resource
    end
```

With this change:
```
    Resource Declaration:
    ---------------------
    # In .\foo.rb

      1: dsc_resource 'Create Test User' do
      2:   resource :user
      3:   property :username, 'testuserps5__\\'
      4:   property :password, ps_credential(ENV['COMPUTERNAME'])
      5:   property :ensure, 'Present'
      6: end

    Compiled Resource:
    ------------------
    # Declared in .\foo.rb:1:in `run_chef_recipe'

    dsc_resource("Create Test User") do
      action [:run]
      retries 0
      retry_delay 2
      properties {:username=>"testuserps5__\\", :password=>New-Object System.Management.Automation.PSCredential('placeholder',('01000000d08c9ddf0115d1118c7a00c04fc297eb010000006083790c2fc2a24c946799335d38282b0400000002000000000010660000000100002000000086b05f5a3d8ba25dcf328f5cce1bab8343b4ebc981947f186ebb0f215bba0e1b000000000e800000000200002000000060624c80a246a5e8374fb5e9d9aa87129d16a5afd65eab7ca10942a68a4d8b9f3000000082ae38decc68a5f8ee03914b5a4b7cb72683c7ab31be65443e932646b99f82639ae112d90b81cdeaa11edc81cccc73e84000000025cc0420658c4c9e72a355a0262a2b1e9dbac9ccbaedfd2e1d87e9ab27e73950261ffc4d2dd350d27b51e14adc2afc9e8a78935ec4d83c12112916f8353840b7' | ConvertTo-SecureString)), :ensure=>"Present"}
      resource :user
      reboot_action :nothing
      declared_type :dsc_resource
    end
```
cc @adamedx @chef/client-windows

### DIFF
--- a/lib/chef/resource/dsc_resource.rb
+++ b/lib/chef/resource/dsc_resource.rb
@@ -20,8 +20,26 @@ require 'chef/dsl/powershell'
 class Chef
   class Resource
     class DscResource < Chef::Resource
-
       provides :dsc_resource, os: "windows"
+
+      # This class will check if the object responds to
+      # to_text. If it does, it will call that as opposed
+      # to inspect. This is useful for properties that hold
+      # objects such as PsCredential, where we do not want
+      # to dump the actual ivars
+      class ToTextHash < Hash
+        def to_text
+          descriptions = self.map do |(property, obj)|
+            obj_text = if obj.respond_to?(:to_text)
+                         obj.to_text
+                       else
+                         obj.inspect
+                       end
+            "#{property}=>#{obj_text}"
+          end
+          "{#{descriptions.join(', ')}}"
+        end
+      end
 
       include Chef::DSL::Powershell
 
@@ -29,7 +47,7 @@ class Chef
 
       def initialize(name, run_context)
         super
-        @properties = {}
+        @properties = ToTextHash.new
         @resource = nil
         @reboot_action = :nothing
       end
--- a/lib/chef/util/powershell/ps_credential.rb
+++ b/lib/chef/util/powershell/ps_credential.rb
@@ -29,9 +29,8 @@ class Chef::Util::Powershell
       "New-Object System.Management.Automation.PSCredential('#{@username}',('#{encrypt(@password)}' | ConvertTo-SecureString))"
     end
 
-    def to_s
-      to_psobject
-    end
+    alias to_s to_psobject
+    alias to_text to_psobject
 
     private
 
--- a/spec/unit/util/powershell/ps_credential_spec.rb
+++ b/spec/unit/util/powershell/ps_credential_spec.rb
@@ -21,7 +21,7 @@ require 'chef/util/powershell/ps_credential'
 
 describe Chef::Util::Powershell::PSCredential do
   let (:username) { 'foo' }
-  let (:password) { 'password' }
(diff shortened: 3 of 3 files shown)


coderanger:
Do we care about `to_text`? (read: can this end up as something displayed from a resource?)

jaym:
@coderanger there wont be any calls to `to_text` on this object as it is not a resource. However, this change comes about because this object can be owned by a resource. Being that this describes a password (or has one), we shouldn't let people print it out. Basically, it implicitly marks the field on the resource as sensitive. I'll add a better description... see if I described what you were looking for

btm:
Is there a specific case where #inspect is being called, such as debugging output? There's a bunch of methods that would print out the password, from `instance_variable_get(:@password)` to `to_yaml`.

jaym:
@btm inspect gets called to describe the resource that failed. In this case, ps credential is used on `dsc_resource` and `dsc_script`, so if either of those were provided a `ps_credential` and not marked as `sensitive`, they would print the password in plain text. I think this is one of those cases where we know the contents are sensitive, so we should treat it as such regardless of if the user marked their resource as sensitive.

btm:
:+1:

mwrock:
:+1:

coderanger:
@jaym My point was do we want to define `to_text` or `inspect`. Either would fix the `Resource#to_text` output to be redacted. Overriding `inspect` to be unlike most other objects seems worse but I'm not sure what case we are trying to protect.

lib/chef/resource/dsc_resource.rb
```diff
@@ -20,16 +20,34 @@
...
     class DscResource < Chef::Resource
-
       provides :dsc_resource, os: "windows"
 
+      # This class will check if the object responds to
+      # to_text. If it does, it will call that as opposed
+      # to inspect. This is useful for properties that hold
+      # objects such as PsCredential, where we do not want
+      # to dump the actual ivars
+      class ToTextHash < Hash
```
coderanger: We should probably come up with a better long-term solution for this but for now :+1:

PR #4206, opened 2015-11-25
fix rspecs-ctrl-c

### LINKED ISSUE #2520: Chef's rspec tests not easily stopable
When running Chef's rspec tests via either:

    rake spec

or 

    rspec

The test run can become very hard to kill at the terminal.  Sending an interrupt (^c) simply causes the currently running test to fail and then proceeds to the next test.  Suspending the job (^z) and sending a SIGKILL to the rspec process works, but feels pretty ugly.

lamont-granquist:
we can't fork off tinyserver without a lot of work in the test
codebase, so instead save a copy of the old rspec signal handler and then
after tinyserver has started (and we're sure that rack has scribbled
over the signal handler), restore the rspec signal handler.

there's an obvious race if someone hits ctrl-c after tinyserver has
scribbled over the sigint handler but before we restore it, but this is
99.9% better of a solution than nothing.

closes #2520

### DIFF
--- a/spec/tiny_server.rb
+++ b/spec/tiny_server.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Daniel DeLeo (<dan@opscode.com>)
-# Copyright:: Copyright (c) 2010 Opscode, Inc.
+# Copyright:: Copyright (c) 2010-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@chef/client-core review plz

stevendanna:
omg yes

jkeiser:
:+1:

mcquin:
needs spec

thommay:
:rocket: :clap:

PR #4226, opened 2015-12-02
Converge actions all together instead of in action :x (fixes #4124)

### LINKED ISSUE #4124: 12.5.1 breaks LWRP code that works with <= 12.4.3/4
The following LWRP code: https://gist.github.com/Sauraus/6b699f7726957bb8f577 works 100% in 12.4.3/4 and does absolutely nothing in 12.5.1.

The worst part about it is that in 12.5.1 there is no error thrown or anything chef-client just says (up-to-date) and continues on without actually executing the action method.

jkeiser:
Fixes #4124

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -351,10 +351,20 @@ class Chef
     # @api private
     module InlineResources
 
-      # Our run context is a child of the main run context; that gives us a
-      # whole new resource collection and notification set.
-      def initialize(resource, run_context)
-        super(resource, run_context.create_child)
+      # Create a child run_context, compile the block, and converge it.
+      #
+      # @api private
+      def compile_and_converge_action(&block)
+        old_run_context = run_context
+        @run_context = run_context.create_child
+        return_value = instance_eval(&block)
+        Chef::Runner.new(run_context).converge
+        return_value
+      ensure
+        if run_context.resource_collection.any? { |r| r.updated? }
+          new_resource.updated_by_last_action(true)
+        end
+        @run_context = old_run_context
       end
 
       # Class methods for InlineResources. Overrides the `action` DSL method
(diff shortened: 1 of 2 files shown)


lib/chef/provider.rb
```diff
@@ -357,6 +357,16 @@ def initialize(resource, run_context)
         super(resource, run_context.create_child)
       end
 
+      def converge_action(action)
```
tyler-ball: Can you add a comment explaining why this is defined both in this module and on the provider class?

tyler-ball:
:+1:

coderanger:
Will need to figure out how this will work with Poise's implementation and with `use_inline_resources`. Definitely the right move but might be a breaking change?

lamont-granquist:
I'm still digesting, but I think this restores the old use_inline_resources behavior that we've had in 11.x and 12.x for a long time...

jkeiser:
@coderanger the specs on poise master pass with this change in place--not sure if that's enough to tell if it breaks anything?

jkeiser:
@lamont-granquist @coderanger @tyler-ball updated this to restore the old behavior more closely than before. Not a fan of the old behavior, but we can change that in an RFC.

coderanger:
@jkeiser I'll know once this gets merged and all my master-branch tests either fail or don't :) Either way though, this removes the old pre-`use_inline_resources` mode of operation which I think would be a compat break.

lamont-granquist:
I think the latest code does not.  We're back to having `compile_and_converge_action` called from the action method which is only included into the class when InlineResources is mixed in.

There should actually be a test as well in the "this is totally a bug, but for backcompat purposes, it adds the resources to the main resource collection and does not get marked updated" test.  Can argue about if its a "bug" or not, but that's the behavior of pre-use_inline_resources...

coderanger:
Yeah, I hadn't checked the updated patch before saying that, I agree I think this is now compatible :)

PR #4227, opened 2015-12-03
Make resource.to_hash include default property values

### LINKED ISSUE #4127: LWRP non-string default values of new_resource are missing after hash conversion in chef client>12.4.3
The following code doesn't work in chef client 12.5.1 https://gist.github.com/wp-vtuska/9b093542a6e03f37e709
It is only a gutted example. We are using the hash format of the new_resource object as a template variable. It worked before 12.5.1. 

It looks like the undefined non-string default resource attribute values are missing from instance_variables even if we tried to access them explicitly before the object hash conversion. Is it an intentional modification in the newer chef client?

Current workarounds:
1. new_resource.variable = new_resource.variable
2. using only string attribute values

jkeiser:
Fixes #4127.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -664,13 +664,18 @@ class Chef
     end
 
     def to_hash
+      # Grab all current state, then any other ivars (backcompat)
+      result = {}
+      self.class.state_properties.each do |p|
+        result[p.name] = p.get(self)
+      end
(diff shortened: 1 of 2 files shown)


lamont-granquist:
:+1:

jkeiser:
This was checked in last night (rebased, so it didn't pick it up). Should be in last night's build.

PR #4228, opened 2015-12-03
fix couple of issues in dpkg cleanup

lamont-granquist:
this is just code rearrangement, no changes to behavior at
all.  pushes the source_array and resolved_source_array up
into the superclass so that other providers can use it (as it
was intended, but that slipped out of the last PR).  also
renamed it to something a little longer and less likely to
clash with variables (which it was already doing).

the block comment also refers to an earlier iteration that used
coercions, which had to be backed out because of chefspec.

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -481,6 +481,37 @@ class Chef
         [ new_resource.version ].flatten.map { |v| v.to_s.empty? ? nil : v }
       end
 
+      # TIP: less error prone to simply always call resolved_source_array, even if you
+      # don't think that you need to.
+      #
+      # @return [Array] new_resource.source as an array
+      def source_array
+        if new_resource.source.nil?
+          package_name_array.map { nil }
+        else
+          [ new_resource.source ].flatten
+        end
+      end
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@chef/client-core review should only take a second

lib/chef/provider/package.rb
```diff
@@ -481,6 +481,37 @@ def new_version_array
...
+
+      # Helper to handle use_package_name_for_source to convert names into local packages to install.
+      #
+      # @return [Array] Array of sources with package_names converted to sources
+      def resolved_source_array
+        @resolved_source_array ||=
+          begin
+            source_array.each_with_index.map do |source, i|
+              package_name = package_name_array[i]
+              # we require at least one '/' in the package_name to avoid dpkg_package 'foo' breaking due to a random 'foo' file in cwd
```
thommay: `package` rather than `dpkg_package`

thommay:
:+1:

lamont-granquist:
@chef/client-core pingles again for review

btm:
:+1:

PR #4229, opened 2015-12-03
Convert all remaining consumers of Chef::REST to Chef::ServerAPI

### LINKED ISSUE #2737: knife should use Chef::REST or Chef::ServerAPI correctly
It would be beneficial to certain customers if `knife` (including number of sub-commands e.g., `ssl fetch`, etc.), would respect HTTP/HTTPS proxy when such is present and set.

This is a feature request driven by questions from our customers who have heavily policed network environments (i.e., enterprises, banks, etc.).

### LINKED ISSUE #3518: Remove json_class from all the things and use Chef::ServerAPI
The json_class auto-inflation is terribad and needs to go away.

Chef::REST also needs to be replaced with Chef::ServerAPI.

thommay:
:construction:
Also clobber the json auto expansion guff. Fixes #2737, fixes #3518
:construction:


thommay:
cc @lamont-granquist

lamont-granquist:
holy awesomeness...

lib/chef/data_bag.rb
```diff
@@ -70,11 +70,11 @@ def to_json(*a)
     end
 
     def chef_server_rest
-      @chef_server_rest ||= Chef::REST.new(Chef::Config[:chef_server_url])
+      @chef_server_rest ||= Chef::ServerAPI.new(Chef::Config[:chef_server_url])
```
lamont-granquist: kind of vaguely curious why we don't have a require change up top for all these...  not sure if the answer to that question will make me sad though...

lib/chef/server_api.rb
```diff
@@ -31,6 +31,7 @@ def initialize(url = Chef::Config[:chef_server_url], options = {})
       options[:client_name] ||= Chef::Config[:node_name]
       options[:signing_key_filename] ||= Chef::Config[:client_key]
       options[:signing_key_filename] = nil if chef_zero_uri?(url)
+      options[:inflate_json_class] = false
```
lamont-granquist: is this also going to wind up slurping in deprecation of json_class everywhere into the scope?

lib/chef/data_bag.rb
thommay: i was thinking about going through and require'ing server_api explicitly, but we don't actually require rest anywhere right now. So yes, I think it'll make you sad, and I'll do the explicit require and make you less sad.

lib/chef/server_api.rb
thommay: nah, it just flips from `JSONCompat.from_json` to `JSONCompat.parse` (which is just a wrapper over `FFI_Yajl::Parser.parse(source, opts)` )

lib/chef/data_bag.rb
lamont-granquist: Yeah, I suspect this is another case where we're relying on some hidden side effect of how we require the universe in order to make everything work...

thommay:
I'm gonna do a follow up PR to add deprecations to `Chef::REST`and any `json_create` methods, but I think this should be ready to review, @chef/client-core (travis failures are ipv6 and kitchen, which are known busted and being worked on)

thommay:
it turned into a bit of a monster, happy to split up if need be…

lib/chef/http.rb
```diff
@@ -154,6 +154,25 @@ def request(method, path, headers={}, data=false)
       raise
     end
 
+    # Makes an HTTP request to +path+ with the given +method+, +headers+, and
+    # +data+ (if applicable). Does not apply any middleware, besides that
+    # needed for Authentication.
+    def raw_request(method, path, headers={}, data=false)
```
danielsdeleo: I think this either needs to go one level deeper in the class hierarchy (i.e., on `ServerAPI`) or maybe in a `BasicServerAPI` which only includes the necessary middleware. Regular `Chef::HTTP` shouldn't know about any middleware stuff at all and should only make vanilla HTTP requests.

spec/unit/knife_spec.rb
```diff
@@ -156,8 +156,8 @@ class AwesomeCheffsCommand < Chef::ChefFS::Knife
                     "Accept-Encoding"=>"gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
                     'X-Chef-Version' => Chef::VERSION,
                     "Host"=>"api.opscode.piab",
-                    "X-REMOTE-REQUEST-ID"=>request_id,
-                    'X-Ops-Server-API-Version' => Chef::HTTP::Authenticator::DEFAULT_SERVER_API_VERSION}}
```
danielsdeleo: Where'd this go?

thommay: It's not an auth'd request (rest is an instance of NoAuthHTTP (above)), so it's not hitting the Authenticator at all. We only use the noauth_rest stuff for get requests to supermarket.

thommay:
@danielsdeleo addressed your requests

danielsdeleo:
LGTM now :+1: Not sure what's going on with appveyor, we should fix that but it seems unrelated to this patch.

jaym:
rebase to fix appveyor failure

thommay:
ping @chef/client-core

lib/chef/knife.rb
```diff
@@ -26,14 +26,24 @@
...
 class Chef
   class Knife
 
-    Chef::REST::RESTRequest.user_agent = "Chef Knife#{Chef::REST::RESTRequest::UA_COMMON}"
+    class NoAuthHTTP < Chef::HTTP
+      use Chef::HTTP::JSONInput
+      use Chef::HTTP::JSONOutput
+      use Chef::HTTP::CookieManager
+      use Chef::HTTP::Decompressor
+      use Chef::HTTP::RemoteRequestID
```
lamont-granquist: `use ValidateContentLength` should be added here (after Decompressor)

probably should move this to Chef::HTTP::JSONSimple / Chef::HTTP::Simple::JSON or something for re-use..

lamont-granquist:
okay one nit to pick other than that :+1:

lamont-granquist:
:+1:

PR #4231, opened 2015-12-03
zypper multipackage patch

lamont-granquist:
replaces #4102

### DIFF
--- a/spec/unit/provider/package/zypper_spec.rb
+++ b/spec/unit/provider/package/zypper_spec.rb
@@ -261,4 +261,40 @@ describe Chef::Provider::Package::Zypper do
       end
     end
   end
+
+  describe "when installing multiple packages" do # https://github.com/chef/chef/issues/3570
+    it "should install an array of package names and versions" do
+      allow(Chef::Config).to receive(:[]).with(:zypper_check_gpg).and_return(false)
+      shell_out_expectation!(
+        "zypper --non-interactive --no-gpg-checks install "+
+          "--auto-agree-with-licenses emacs=1.0 vim=2.0"
+      )
+      provider.install_package(["emacs", "vim"], ["1.0", "2.0"])
+      end
+
+    it "should install an array of package names without versions" do
+      allow(Chef::Config).to receive(:[]).with(:zypper_check_gpg).and_return(false)
+      shell_out_expectation!(
+        "zypper --non-interactive --no-gpg-checks install "+
+          "--auto-agree-with-licenses emacs vim"
+      )
+      provider.install_package(["emacs", "vim"], nil)
+    end
+
+    it "should remove an array of package names and versions" do
+      allow(Chef::Config).to receive(:[]).with(:zypper_check_gpg).and_return(false)
+      shell_out_expectation!(
+        "zypper --non-interactive --no-gpg-checks remove emacs=1.0 vim=2.0"
+      )
+      provider.remove_package(["emacs", "vim"], ["1.0", "2.0"])
+      end
+
+    it "should remove an array of package names without versions" do
+      allow(Chef::Config).to receive(:[]).with(:zypper_check_gpg).and_return(false)
+      shell_out_expectation!(
+        "zypper --non-interactive --no-gpg-checks remove emacs vim"
+      )
+      provider.remove_package(["emacs", "vim"], nil)
+    end
+  end
 end
--- a/spec/unit/provider/package/zypper_spec.rb
+++ b/spec/unit/provider/package/zypper_spec.rb
@@ -34,7 +34,7 @@ describe Chef::Provider::Package::Zypper do
 
   before(:each) do
     allow(Chef::Resource::Package).to receive(:new).and_return(current_resource)
-    allow(provider).to receive(:shell_out).and_return(status)
+    allow(provider).to receive(:shell_out!).and_return(status)
     allow(provider).to receive(:`).and_return("2.0")
   end
 
@@ -60,7 +60,7 @@ describe Chef::Provider::Package::Zypper do
     end
 
     it "should run zypper info with the package name" do
-      shell_out_expectation(
+      shell_out_expectation!(
         "zypper --non-interactive info #{new_resource.package_name}"
       ).and_return(status)
       provider.load_current_resource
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@gsaslis can you look this over?

it uses the new use_multipackage_api declaration, and it cleans up some of the tortured logic in load_current_resource.  i think its much more readable than where you were heading.

about the only question i've got (apart from if it even works or not, because i don't have a suse box easily at hand right now) is what that exception in load_current_resource is all about, because it looked like well intentioned but incorrect code:

- if zypper info was causing that exception to be thrown if both the package was not installed and not available then that was wrong and we simply need to set both to nil and let the superclass deal with the exception in define_resource_requirements

- if this was trying to catch `zypper: command not found` kinds of things then maybe we could covert it to a shell_out_with_timeout! with maybe `returns: [0, 1]` and then if the whole zypper universe is fatally broken on the box then we just let shell_out explode and report the exact failure to the user.

lamont-granquist:
tested this on an opensuse virt and it seems to work fine.

gsaslis:
@lamont-granquist yes - that use_multipackage_api certainly makes things much simpler and much more readable (!!) 

:+1: looks good to me

Noobie question, (but I am one with Chef, so it's ok) : how would I go about testing this chef branch with my own VM / box ? is there a link / doc somewhere? (I was expecting to find something on https://github.com/chef/chef/blob/master/CONTRIBUTING.md , but it doesn't seem to be mentioned there... )

lamont-granquist:
check out:

http://lists.opscode.com/sympa/arc/chef/2015-10/msg00010.html

probably you want something like:

% gem install appbundle-updater
% sudo appbundle-updater chefdk chef master

or

% gem install appbundle-updater
% sudo appbundle-updater chef chef master

depending on if you're updating chef inside omnibus-chef or omnibus-chefdk...

lamont-granquist:
oh except obviously this isn't merged to 'master' so use 'lcg/zypper-multipackage' instead...

lamont-granquist:
BTW, If you can flame-test this really quickly we might be able to get this into 12.6.0, otherwise it'll get delayed probably until next year...

lamont-granquist:
(looks like this'll slip to january)

gsaslis:
@lamont-granquist: yeah, sorry dude, won't have much time in the week to look at this (looks like you're not the only ones pushing for releases before xmas - and there's a day job as well ;) )

lamont-granquist:
okay, that's cool...   we're trying to get a monthly tempo on releases, so hopefully it won't slip too much...

lamont-granquist:
@gsaslis do you have any time to look at this?

gsaslis:
I will this weekend, yes! (if it can wait ; )  )

-- 
Yorgos Saslis
Software Engineer

lamont-granquist:
totally, it looks like i need to rebase it anyway...

spec/unit/provider/package/zypper_spec.rb
```diff
@@ -154,14 +145,14 @@ def shell_out_expectation!(command, options=nil)
         "zypper --non-interactive --no-gpg-checks install "+
         "--auto-agree-with-licenses emacs=1.0"
       )
-      provider.upgrade_package("emacs", "1.0")
+      provider.upgrade_package(["emacs"], ["1.0"])
     end
     it "should run zypper upgrade without gpg checks" do
```
gsaslis: I'm not sure what 
`"should run zypper upgrade without gpg checks"` really offers over `"should run zypper update without gpg checks"` - perhaps the latter should be removed altogether?

(unless one of the two was meant to test the `dist-upgrade, dup Perform a distribution upgrade.` feature of zypper)

spec/unit/provider/package/zypper_spec.rb
```diff
@@ -136,7 +127,7 @@ def shell_out_expectation!(command, options=nil)
       shell_out_expectation!(
         "zypper --non-interactive install --auto-agree-with-licenses emacs=1.0"
```
gsaslis: when i first read through this, I thought the `install` here was a bug (was expecting to read `update`, since we're in the upgrade section of the tests). 
Had to go through the code and actually test on the command-line that the `zypper install` actually does upgrade packages (even when a version isn't specified and there's a newer one available... : / 
Maybe we could add a line comment in the `upgrade_package` method of the zypper provider to explain why `install_package` is called, or am I the only one who would have thought `zypper install` wound not upgrade the package if it's already installed...?

spec/unit/provider/package/zypper_spec.rb
```diff
@@ -209,14 +200,14 @@ def shell_out_expectation!(command, options=nil)
       shell_out_expectation!(
```
gsaslis: I believe the title here has a typo - should be purge_package instead of remove_package

gsaslis:
@lamont-granquist with regards to your question about that exception being thrown: 
`zypper info ` can return an error code in a few cases where it cannot complete successfully, but not when it cannot find the package (in that case, the return code is 0, unless you pass in some weird input, such as `$1` as the package name ) .
I think that probably leads us on to your 2nd bullet, and if you really just needed a command you could run, so you could be sure that "zypper is all set up nicely" and that your issue is irrelevant with zypper info and packages, etc. then according to the [zypper manual](https://en.opensuse.org/SDB:Zypper_manual_(plain)) you can use `zypper ping` that does just that and returns 0  ; ) 

Hope that helps!

lamont-granquist:
Yeah, I think I answered those bullet points for myself.  I just changed it to shell_out_with_timeout! so that the exception (which really truly is exceptional in this case and indicates zypper has gone insane) will just blow up and it'll get reported to the user.  An exception is the correct behavior, but there was nothing useful gained from decorating the exception from shell out with a Chef::Exceptions::Package error.

spec/unit/provider/package/zypper_spec.rb
lamont-granquist: on tests with the Chef::Config variable set to false, the other tests the default.

spec/unit/provider/package/zypper_spec.rb
lamont-granquist: no, the title is indicating that purge_package just calls remove_package.

lamont-granquist: oh i see, it doesn't quite though...

spec/unit/provider/package/zypper_spec.rb
gsaslis: ok I see, then perhaps just changing the test's title would help?

spec/unit/provider/package/zypper_spec.rb
lamont-granquist: i think we get the correct behavior through the idempotency checks in the parent provider.   if the package is already installed then we won't call either install or upgrade on it if the action is install.  if the action is upgrade then we'll issue the command.  i'd argue 'zypper install' has broken behavior, but it doesn't actually matter to us...

lamont-granquist:
@chef/client-core i think this is ready for review

spec/unit/provider/package/zypper_spec.rb
gsaslis: :+1: 

(and I would agree with you on that argument... it wasn't at all obvious to me that it should upgrade when running `zypper install`... That's why I'm proposing we add a line comment there to document this non-obvious behaviour)

lamont-granquist: added

lamont-granquist:
@chef/client-core still needing review...

thommay:
:+1: merge once rebased

PR #4234, opened 2015-12-03
Fix minor typo

tas50:


jaym:
:+1: 
think you should just merge it as obvious

PR #4236, opened 2015-12-04
Add timeout attribute to dsc_resource

### LINKED ISSUE #4232: dsc_resource has a fixed 10 minute timout which causes long running dsc resources to fail
Using DSC resources with Chef-client 12.5.1 and WMF 5 "Production Preview" will timeout and fail if the resource takes longer than 10 minutes to complete a change. This is especially problematic when using the xSQLServerSetup DSC resource.
Unlike dsc_script there is no property within the dsc_resource that allows the timeout value to be modified.

jaym:
`dsc_resource` was missing a timeout resource. Some resources,
such as `xSQLServerSetup` were taking longer than the default
of 10 minuites set by mixlib-shellout.

Fixes #4232 

cc @chef/client-windows

### DIFF
--- a/lib/chef/provider/dsc_resource.rb
+++ b/lib/chef/provider/dsc_resource.rb
@@ -154,7 +154,7 @@ class Chef
           "Invoke-DscResource #{switches}",
           output_format
         )
-        cmdlet.run!
+        cmdlet.run!({}, {:timeout => new_resource.timeout})
       end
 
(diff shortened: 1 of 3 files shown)


smurawski:
:+1:

btm:
It'd be swell if we tested that this made it all the way to shellout, but :+1:

Issue #4215, opened 2015-12-01
knife ssl check does not honor proxy environment variables


jkeiser:
When a proxy is set, knife ssl check and fetch do not honor it.

```
export http_proxy=http://proxy.com:3168
knife ssl check
knife ssl fetch
```

PR #4237, opened 2015-12-04
Enabling 'knife ssl check/fetch' commands to respect proxy environment variables and moving proxy environment variables export to Chef::Config

tyler-ball:
https://github.com/samuelkadolph/ruby-proxifier is a gem which gives the ability to have `TCPSocket` respect the proxy environment variables.  It does this by sending a `CONNECT` request to the proxy - see https://github.com/samuelkadolph/ruby-proxifier/blob/master/lib/proxifier/proxies/http.rb

\cc @jkeiser @chef/client-maintainers

Fixes https://github.com/chef/chef/issues/4215

**EDIT**
To get the tests passing for this I needed to modify Knife to run the logic which exports `Chef::Config[:http_proxy]` to `ENV['http_proxy']`.  I decided to move the exporting logic into `Chef::Config` because that is where we want it to be able to leverage that logic in Test Kitchen and Berkshelf.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,114 +26,27 @@ env:
 
 matrix:
   include:
-  - rvm: 2.1
-    sudo: true
-  - rvm: 2.2
-    sudo: true
   - rvm: rbx
     sudo: true
   - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'chef-zero', github: 'chef/chef-zero'\""
-    script: bundle exec rake chef_zero_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'cheffish', github: 'chef/cheffish'\""
-    script: bundle exec rake cheffish_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'chef-provisioning', github: 'chef/chef-provisioning'\""
-    script: bundle exec rake chef_provisioning_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'chef-provisioning-aws', github: 'chef/chef-provisioning-aws'\""
-    script: bundle exec rake chef_provisioning_aws_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'chefspec'\""
-    script: bundle exec rake chefspec_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'chef-sugar'\""
-    script: bundle exec rake chef_sugar_spec
-  # Requires vagrant
-  # - rvm: 2.2
-  #   cache:
-  #   env: "GEMFILE_MOD=\"gem 'chef-rewind'\""
-  #   script: bundle exec rake chef_rewind_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'foodcritic', github: 'acrmp/foodcritic', branch: 'v5.0.0'\""
-    script: bundle exec rake foodcritic_spec
-  - rvm: 2.2
-    before_install:
-    env: "GEMFILE_MOD=\"gem 'halite', github: 'poise/halite'\""
-    script: bundle exec rake halite_spec
-  - rvm: 2.2
-    env: "GEMFILE_MOD=\"gem 'poise', github: 'poise/poise'\""
-    script: bundle exec rake poise_spec
-  ### START TEST KITCHEN ONLY ###
-  - rvm: 2.2
-    gemfile: kitchen-tests/Gemfile
-    before_install:
-    - echo -n $DO_KEY_CHUNK_{0..30} >> ~/.ssh/id_aws.base64
-    - cat ~/.ssh/id_aws.base64 | tr -d ' ' | base64 --decode > ~/.ssh/id_aws.pem
-    before_script:
-    - cd kitchen-tests
-    script:
-# FIXME: we should fix centos-6 against AWS and then enable it here
-    - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen test ubuntu; fi
-    after_failure:
-    - cat .kitchen/logs/kitchen.log
-    after_script:
-    - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen destroy ubuntu; fi
-    env:
-    - KITCHEN_YAML=.kitchen.travis.yml
-    - EC2_SSH_KEY_PATH=~/.ssh/id_aws.pem
-    - secure: VAauyVnAMWhqvnhJOJ/tCDn3XAdWqzbWiDVQPNBkqtm2SBIvhmZl2hlrusvw6YLU31Prdf8fSFhOSysVQQs/rJYrmD/1BfV79p6M7cGXYZ0nGWwldF81N296lyFoZLyrqtmG4G0cx3Pw2ojADFgFe+B5eTGlqJFD+z371g4RF/Y=
-    - secure: A+qtUF2LPJGkUAdvt04AwZMt69rzaeTyR0/1XEOAuntBKKXSCzddUzr5ePDc9QQ/57AWywKxhVLpnxk3QzKN7r7zerDxyIJBgklNDpNAKkeQjP3T6FpaKEIN9ROcpPtsM6FJ5Agb+bEQoRJF7s+ampO3wLV3XpTiWNuWkcAhv9A=
-    - secure: J8JIg15trrPgc8X/1DsaUWDQCdDWTvN/AorXzZ/ReudHS6G/KpoynZ5lTmKjlgFiFNE/TGMDv486pStGtIcarTKTuIEmNADdEWlAVH7bxclpayMjtppVuapRCkZWccs5gz5CJyhX7yhQCFTYoqVox9Y4qHGCluF3oqCcPRtCOOw=
-    - secure: NJYn0blTMwIoFxZlsoMWK8hPO/fi45rgWOqEImnjvSRk++5WL+GgjLBgLvEi7wCMkBijhIMWtnva60ojd4MrxeS7evrmGRjJKXnPuSKEsrGbArZPskBjCAcg+3PlnQQUkFf6hvbGD3HZlJtcbs4hrx8tbDT2Ie7bmQfqpsawKY4=
-    - secure: FipoX1VzZkzPUP6Gxd05DEva7cX6xKK2Wdq+Y18nNkyW2afPLXCNl5kCsNrgvbqAzbjKaP2M8+b0zwKjrFzNebqmmx1RRfZUJWUkNRF1EgE+tHytmMZW6tNcQlTlvA0KqXi4Dt6SIQ0l/DhwwNKZ80jmpiyYi/ErxIXzbVgVtYA=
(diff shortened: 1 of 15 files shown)


lib/chef/knife/proxified_socket.rb
```diff
@@ -0,0 +1,38 @@
+# Author:: Tyler Ball (<tball@chef.io>)
```
jkeiser: Let's put this in Chef::Mixin (IMO).

lib/chef/knife/ssl_fetch.rb
```diff
@@ -28,6 +28,8 @@ class SslFetch < Chef::Knife
         require 'socket'
         require 'uri'
         require 'openssl'
+        require_relative 'proxified_socket'
```
jkeiser: require_relative is an anti-pattern because it bypasses the Ruby "don't require a file multiple times" check.

coderanger: No it doesn't, in fact it's generally considered good to use because it is much much faster than `require` since it doesn't process the load paths.

lib/chef/mixin/proxified_socket.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Chef
+  module Mixin
+    module ProxifiedSocket
+
+      # This looks at the environment variables and leverages Proxifier to
+      # make the TCPSocket respect ENV['https_proxy'] or ENV['http_proxy'] if
+      # they are present
+      def proxified_socket(host, port)
+        proxy = ENV['https_proxy'] || ENV['http_proxy'] || false
```
danielsdeleo: So maybe this is part of what you're fixing, but in the past we allowed you to set proxy and no_proxy via config or environment variable. Is that still relevant?

.travis.yml
```diff
@@ -137,6 +137,7 @@ matrix:
       - bundle exec chef-client --version
       - git clone https://github.com/chef/proxy_tests.git
       - cd proxy_tests
+      - git checkout tball/knife_tests
```
danielsdeleo: Assuming this is temporary?

danielsdeleo:
A few questions and travis seems grumpy, but the code overall looks good.

.travis.yml
tyler-ball: correct - its going to be a merge/fail/merge/pass dance

lib/chef/mixin/proxified_socket.rb
tyler-ball: In the future we are still going to tell users to set their proxy settings in the Chef `config.rb` but behind the scenes that will export these environment variables.

This is because we also have the requirement that all the Chef workflows should still work only if the ENV variables are set and the config.rb isn't.  The ENV variables and what is in Chef::Config should be in sync after Chef starts up and makes them in sync, but that code doesn't exist yet.  So I'm looking at the ENV for now.

danielsdeleo: gotcha

chef-config/lib/chef-config/config.rb
```diff
@@ -710,6 +711,64 @@ def self.windows_home_path
...
+
+    # Public method that users should call to export proxies to the appropriate
+    # environment variables.  This method should be called after the config file is
+    # parsed and loaded.
+    # TODO add some post-file-parsing logic that automatically calls this so
+    # users don't have to
+    def self.export_proxies
+      export_proxy("http", http_proxy, http_proxy_user, http_proxy_pass) if http_proxy
+      export_proxy("https", https_proxy, https_proxy_user, https_proxy_pass ) if https_proxy
+      export_proxy("ftp", ftp_proxy, ftp_proxy_user, ftp_proxy_pass ) if ftp_proxy
```
btm: there's an extra space before the closing parenthesis here and the line above.

chef-config/lib/chef-config/config.rb
```diff
@@ -710,6 +711,64 @@ def self.windows_home_path
...
+    #   pass = password
+    def self.export_proxy(scheme, path, user, pass)
+      path = "#{scheme}://#{path}" unless path.include?('://')
+      # URI.split returns the following parts:
+      # [scheme, userinfo, host, port, registry, path, opaque, query, fragment]
+      parts = URI.split(URI.encode(path))
+      # URI::Generic.build requires an integer for the port, but URI::split gives
+      # returns a string for the port.
+      parts[3] = parts[3].to_i if parts[3]
+      if user
```
btm: Will user and pass be nil if we don't set them? If you pass empty strings for user and pass, this doesn't work right.

```
export_proxy("http", "localhost:8080", "", "") => "http://:@localhost:8080"
export_proxy("http", "localhost:8080", nil, nil) => "http://localhost:8080"
```

it seems like we should do the right thing with an empty string anyway, I could see a user setting http_proxy_user to an empty string thinking that means "no user".

btm:
I don't see a test for the case of `export_proxy("https", "localhost:8080", nil, nil)`. 

That is, nothing tests this line: `path = "#{scheme}://#{path}" unless path.include?('://')`

chef-config/lib/chef-config/config.rb
```diff
@@ -710,6 +711,64 @@ def self.windows_home_path
...
+        userinfo = URI.encode(URI.encode(user), '@:')
+        if pass
+          userinfo << ":#{URI.encode(URI.encode(pass), '@:')}"
+        end
+        parts[1] = userinfo
+      end
+
+      path = URI::Generic.build(parts).to_s
+      ENV["#{scheme}_proxy"] = path unless ENV["#{scheme}_proxy"]
```
btm: I realize you don't intend for someone to call `export_proxy` directly, but consider using `downcase` here for consistency just in case?

spec/unit/mixin/proxified_socket_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    it "returns a plain TCPSocket" do
+      expect(TCPSocket).to receive(:new).with(host, port).and_return(socket_double)
+      expect(test_instance.proxified_socket(host, port)).to eq(socket_double)
+    end
+  end
+
+  context "when https_proxy is set" do
+    before do
+      ENV['https_proxy'] = https_uri
+      ENV['http_proxy'] = http_uri
```
btm: The http_proxy shouldn't be used in this test, right?

spec/unit/mixin/proxified_socket_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    let(:proxy_uri) { https_uri }
+    include_examples "proxified socket"
+
+    context "when no_proxy is set" do
```
btm: Did you mean to include this within the https_proxy context? It's not clear from the names.

chef-config/lib/chef-config/config.rb
tyler-ball: fixed, thanks

chef-config/lib/chef-config/config.rb
tyler-ball:
```
[chef]$ rando= bundle exec pry
[1] pry(main)> ENV['rando']
=> ""
```

Normally if it isn't specified, `ENV['rando']` will be nil.  But because it looks pretty easy to pass them in as empty strings, I'll convert empty strings I find to nil values.  Thanks!

tyler-ball: I'm actually only going to check if user is an empty string, because its possible that someone's password could be `"        "`

chef-config/lib/chef-config/config.rb
tyler-ball: sure

tyler-ball:
Added a test for `path = "#{scheme}://#{path}" unless path.include?('://')`

spec/unit/mixin/proxified_socket_spec.rb
tyler-ball: I'm purposefully setting both of these because we prefer the https variable

spec/unit/mixin/proxified_socket_spec.rb
tyler-ball: Yes.  This is testing that no_proxy is also provided to Proxified (in addition to the proxy URI) when it is set.  Added a comment to that effect

spec/unit/mixin/proxified_socket_spec.rb
btm: Where is our preference? Is there a test here for, `it "prefers https over http"`?

btm:
:+1: on green.

tyler-ball:
Travis is green and there is a ridiculous Appveyor queue right now - I'm going to squash and merge

MathanRamadurai:
I am getting below error while i run "knife ssl check -VV" from my ubuntu linux (16.0).. please guide me to solve this error..

mramadurai@mramadurai-W840AU:~/chef-repo$ knife ssl check -VV
INFO: Using configuration from /home/mramadurai/chef-repo/.chef/knife.rb
DEBUG: Checking SSL cert on https://poc-chef-master/organizations/silica
Connecting to host poc-chef-master:443
/opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/proxifier-1.0.3/lib/proxifier.rb:20:in `rescue in Proxy': unknown proxy scheme `https' (ArgumentError)
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/proxifier-1.0.3/lib/proxifier.rb:17:in `Proxy'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/mixin/proxified_socket.rb:34:in `proxified_socket'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife/ssl_check.rb:80:in `verify_peer_socket'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife/ssl_check.rb:130:in `verify_cert'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife/ssl_check.rb:249:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife.rb:443:in `block in run_with_pretty_exceptions'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/local_mode.rb:44:in `with_server_connectivity'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife.rb:442:in `run_with_pretty_exceptions'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/knife.rb:219:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/lib/chef/application/knife.rb:156:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/chef-12.16.42/bin/knife:25:in `<top (required)>'
	from /usr/bin/knife:58:in `load'
	from /usr/bin/knife:58:in `<main>'

thommay:
@MathanRamadurai please don't post on old Pull Requests. 

At Chef, we use GitHub Issues to track bugs and feature requests. Questions like "How do I .... with Chef?" are better answered by our awesome community on the Chef mailing lists, or on StackOverflow. (Please pick only one place to ask your question, however, as it's considered impolite to post the same question to multiple forums.)

You can find information about joining and asking questions on our mailing lists [here](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#chef-community). On StackOverflow, please tag your question with the "chef" label so that it shows up in [this list](https://stackoverflow.com/questions/tagged/chef).

Please get in touch with the community via the above mechanisms in order to find a solution to your problem.

PR #4238, opened 2015-12-04
Revert lib/chef/api_client/registration.rb to use API V0 since it doe…

tylercloke:
…sn't support V1.

### DIFF
--- a/lib/chef/api_client/registration.rb
+++ b/lib/chef/api_client/registration.rb
@@ -123,9 +123,13 @@ class Chef
       end
 
       def http_api
-        @http_api ||= Chef::REST.new(Chef::Config[:chef_server_url],
-                                     Chef::Config[:validation_client_name],
-                                     Chef::Config[:validation_key])
+        @http_api ||= Chef::ServerAPI.new(Chef::Config[:chef_server_url],
+                                          {
+                                            :api_version => "0",
+                                            :client_name => Chef::Config[:validation_client_name],
+                                            :signing_key_filename => Chef::Config[:validation_key]
+                                          }
+                                         )
       end
 
       # Whether or not to generate keys locally and post the public key to the


andy-dufour:
My only concern here is that one day I assume the V0 API will be deprecated and then decommissioned. The registration class should change at that point to support re-registration via the V1 (or newer) API, as re-registration is still an active use-case for several customers.

:+1: if the use case of re-registration passes though, we just need to be conscientious of this in the future

/cc @jaymzh

tylercloke:
@andy-dufour Yeah that is a much larger issue still under discussion. Most of the client code has been edited to support both V0 and V1 but there is still a lot of work to do there before the whole client supported V1, and this is a place where that work would have to happen also.

jaym:
:+1:

jaymzh:
@andy-dufour should we change our monkeypatch to this instead?

lamont-granquist:
Yeah, I'd consider this a regression cause by the default going to v1, and we need to roll back to v0 here to get the old behavior.

The right thing to do is to write the code to use v1 and the keys endpoint in the future.

But lets get this merged right now so we can ship it in 12.6.0 and facebook can stop monkeypatching (and anyone else out there that does this doesn't have a blocker to upgrading).

lamont-granquist:
:+1:

jaymzh:
seems to work for me against 12.3.1 server with 12.4.1 client (with this monkey patched into my client.rb, but without @andy-dufour's other fix). Will roll this out.

tylercloke:
@jaymzh Yeah this was closed instead of @andy-dufour's fix. This is the correct fix.

chilicheech:
are there any tests for this? why was this not caught by regression/functional tests? should we add them now to prevent future breakage?

PR #4240, opened 2015-12-04
focus snuck into a PR

lamont-granquist:
this reverts the focus and allows all the tests to run again

### DIFF
--- a/spec/integration/recipes/lwrp_inline_resources_spec.rb
+++ b/spec/integration/recipes/lwrp_inline_resources_spec.rb
@@ -75,7 +75,7 @@ describe "LWRPs with inline resources" do
       end
     end
 
-    it "resources declared in b are executed immediately inline", :focus do
+    it "resources declared in b are executed immediately inline" do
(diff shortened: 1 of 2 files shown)


ksubrama:
:+1: yay!

coderanger:
Protip `config.filter_run(:focus) unless ENV['CI']`

lamont-granquist:
This seems a little better since it'd actually fail the PR that introduced the `:focus`:

```
  config.before(:example, :focus) do
    raise 'This example was committed with `:focus` and should not have been'
  end if ENV['CI']
```

lamont-granquist:
ah, that's just before syntax, just not used to reading it in the config block...

lamont-granquist:
a799d90 is just for testing 652c773 to ensure it fails in CI and i'll be removing that commit before merging

lamont-granquist:
#4226 broke the build but at the same time disabled all the tests from one, so merging this PR even though its red because now the red state of master will be accurate.

PR #4241, opened 2015-12-05
disable tests on 32-bit

lamont-granquist:


lamont-granquist:
these tests use some of the same artifacts as this test:

https://github.com/chef/chef/blob/master/spec/functional/resource/package_spec.rb#L92

so can't run on 32-bit.

this should fix jenkins 32-bit

jaym:
:+1: i'm ok with this

lamont-granquist:
testing this on ubuntu 64-bit is a bit useless (although i validated that they still run), so just merging since i can't make jenkins any worse...

jaym:
also agree with that unless there's a typo in the code i reviewd

PR #4243, opened 2015-12-06
Fix resource `super` (broken because of test :focus)

jkeiser:


lib/chef/provider.rb
```diff
@@ -376,7 +376,18 @@ module ClassMethods
         # compile the resources, converging them, and then checking if any
         # were updated (and updating new-resource if so)
         def action(name, &block)
-          define_method("action_#{name}") { compile_and_converge_action(&block) }
```
lamont-granquist: what if you pull the implementation of compile_and_converge_action into the defined method?  also does wrapping the block in a Proc or a lambda help at all?

jkeiser: @lamont-granquist It can be pulled in, but will not fix the `super` problem. The *only* way for `super` with no methods to work inside the `action :create do ... end` block is for it to be the top level of a defined method. We can remove the `super`-with-no-arguments capability but it will break some cookbooks based on 12.5.

lamont-granquist: Yeah I was trying to understand what the magic was here.  So you simply have to wrap the block with a define_method and nothing else...

lamont-granquist:
:+1:

thommay:
:+1:

PR #4248, opened 2015-12-07
expose HTTP options

thommay:
and use it in registration client specs

### DIFF
--- a/spec/unit/api_client/registration_spec.rb
+++ b/spec/unit/api_client/registration_spec.rb
@@ -94,6 +94,9 @@ describe Chef::ApiClient::Registration do
 
   it "has an HTTP client configured with validator credentials" do
     expect(registration.http_api).to be_a_kind_of(Chef::ServerAPI)
+    expect(registration.http_api.options[:client_name]).to eq("test-validator")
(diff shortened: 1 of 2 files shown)


lamont-granquist:
:+1:

spec/unit/api_client/registration_spec.rb
```diff
@@ -94,6 +94,7 @@
 
   it "has an HTTP client configured with validator credentials" do
     expect(registration.http_api).to be_a_kind_of(Chef::ServerAPI)
+    expect(registration.http_api.options[:client_name]).to eq("test-validator")
```
danielsdeleo: Might be worth getting the authenticator middleware and confirming that it has the client_name that you expect (cf. https://github.com/chef/chef/blob/9fd0dfdb1ce2e89c4e232f159d31a48b1ed014b1/lib/chef/http/authenticator.rb#L71-L73 ). It's kinda reaching into implementation detail, but since the API for middleware initialization is a free-form Hash, I'd prefer the extra assurance we're really configuring it correctly over unit test purity.

danielsdeleo:
:+1: I have a further suggestion but this is an improvement either way.

thommay:
@danielsdeleo I added that additional verification - is that the proper way of doing it?

danielsdeleo:
Yes, that's exactly what I was thinking of

PR #4249, opened 2015-12-07
Get bundler cache working

jkeiser:


lamont-granquist:
:+1: pending @danielsdeleo

jaym:
:+1: this is better. Gets us back to floating for bundler and ruby

danielsdeleo:
Like I said in chat, all I really care about is ensuring we run `chef-config` tests in a clean Ruby VM, to ensure we don't accidentally depend on constants or other code that are only there when Chef is loaded, this does that, so :+1:

PR #4254, opened 2015-12-08
knife/ChefFS support for policyfiles and policy_groups

randomcamel:


jkeiser:
:+1:

We'll work on smaller increments on this next patch, though.

randomcamel:
https://travis-ci.org/chef/chef/builds/96378496 is the failure we expect prior to releasing `chef-zero`, since this PR's specs rely on `ChefZero::RSpec` code currently in `chef-zero` master.

PR #4255, opened 2015-12-08
Reinstate bundler caching for Halite and Poise tests

jkeiser:


danielsdeleo:
yup. :+1: maybe it would be worth adding one extra commit just to be sure it runs correctly twice, but you could just merge it and deal with anything that comes up.

lamont-granquist:
:+1:

PR #4258, opened 2015-12-08
Fix dpkg specs to run on debian 6

jaym:
The exit codes on debian 6 do not match what was expected.
Purging seems to behave differently... metadata is not purged.
We now check the contents of the output to see if the output
has been removed or purged. There are also places where we
dont care if it should be purged or removed... either would
suffice. This case is also represented, and all checks
go through `should_be_purged_or_removed`.

### DIFF
--- a/spec/functional/resource/dpkg_package_spec.rb
+++ b/spec/functional/resource/dpkg_package_spec.rb
@@ -45,11 +45,15 @@ describe Chef::Resource::DpkgPackage, :requires_root, :debian_family_only, arch:
     dpkg_package.package_name name
   end
 
-  def should_be_purged_or_removed(package, action)
-    if action == :purge
-      shell_out!("dpkg -s #{package}", returns: [1])
+  def should_be_purged_or_removed(package, action=nil)
+    status = shell_out("dpkg -s #{package}")
+    output = status.stdout + status.stderr
+    if action.nil? || action == :purge
+      expect(output).to match(/no info|not-installed|not installed/)
+    elsif action == :remove
+      expect(output).to match(/deinstall ok config-files/)
     else
-      expect(shell_out("dpkg -s #{package}").stdout).to match(/deinstall ok config-files/)
+      raise "Unknown action"
     end
   end
 
@@ -187,13 +191,13 @@ describe Chef::Resource::DpkgPackage, :requires_root, :debian_family_only, arch:
       shell_out!("dpkg -i #{test1_0}")
       dpkg_package.run_action(action)
       expect(dpkg_package).to be_updated_by_last_action
-      expect(shell_out('dpkg -s chef-integration-test').exitstatus).to eql(1)
+      should_be_purged_or_removed('chef-integration-test')
     end
 
     it "should do nothing if the package is not installed when the name is a source" do
       dpkg_package.run_action(action)
       expect(dpkg_package).not_to be_updated_by_last_action
-      expect(shell_out('dpkg -s chef-integration-test').exitstatus).to eql(1)
+      should_be_purged_or_removed('chef-integration-test')
     end
 
     it "should remove a package that is installed when the name is the package name and source is nil" do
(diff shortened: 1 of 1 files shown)


spec/functional/resource/dpkg_package_spec.rb
```diff
@@ -45,11 +45,17 @@ def set_dpkg_package_name(name)
     dpkg_package.package_name name
   end
 
-  def should_be_purged_or_removed(package, action)
+  def should_be_purged_or_removed(package, action=nil)
     if action == :purge
-      shell_out!("dpkg -s #{package}", returns: [1])
+        expect(shell_out(
+          'dpkg -s chef-integration-test').stdout).to match(/not-installed|no info/)
```
lamont-granquist: so this is where you want to use something like `.not_to match(/deinstall|installed/)` -- or else you need to invert that and test for either exit(1) or `match(/not-installed/)`.   This is the 'i want to know if the package is really not installed' clause.  You could also make this the default, since this is what you're wanting to test in the 'nil' case as well.

spec/functional/resource/dpkg_package_spec.rb
```diff
@@ -45,11 +45,17 @@ def set_dpkg_package_name(name)
...
+    elsif action == :remove
+      expect(shell_out("dpkg -s #{package}").stdout).to match(/(deinstall ok config-files|not-installed)/)
```
lamont-granquist: Down here this should have been left the way it was since i really wanted to assert that the package was in the deinstalled state, and not in the not-installed/exit(1) state.

spec/functional/resource/dpkg_package_spec.rb
```diff
@@ -45,11 +45,19 @@ def set_dpkg_package_name(name)
...
   end
 
-  def should_be_purged_or_removed(package, action)
+  def should_be_purged_or_removed(package, action=nil)
+    status = shell_out("dpkg -s #{package}")
+    output = status.stdout + status.stderr
     if action == :purge
-      shell_out!("dpkg -s #{package}", returns: [1])
+      not_installed = !(output =~ /deinstall/) &&
+        (!!(output =~ /not installed/) || !!(output =~ /not-installed/))
```
lamont-granquist: i'd actually say the `!!(foo =~ /bar/)`'s here are not improving readability.  even though =~ doesn't return a boolean the || will sort that out just fine.

spec/functional/resource/dpkg_package_spec.rb
```diff
@@ -45,11 +45,19 @@ def set_dpkg_package_name(name)
     dpkg_package.package_name name
   end
 
-  def should_be_purged_or_removed(package, action)
+  def should_be_purged_or_removed(package, action=nil)
+    status = shell_out("dpkg -s #{package}")
+    output = status.stdout + status.stderr
     if action == :purge
-      shell_out!("dpkg -s #{package}", returns: [1])
+      not_installed = !(output =~ /deinstall/) &&
```
lamont-granquist: should be !~ and then drop the extra parens here as well.

spec/functional/resource/dpkg_package_spec.rb
```diff
@@ -45,11 +45,19 @@ def set_dpkg_package_name(name)
...
+        (!!(output =~ /not installed/) || !!(output =~ /not-installed/))
+      expect(not_installed).to be(true)
+    elsif action == :remove
+      expect(output).to match(/deinstall ok config-files/)
+    elsif action == nil
+      expect(output).to match(/deinstall|no info|not-installed/)
```
lamont-granquist: so the fact that you've got a case for :purge and a case for nil when both of those are testing the same thing ("the package is 100% purged") and the implementation is different means you've at least overly complicated something here.

lamont-granquist: deinstall is also clearly wrong here.  its not triggering because everywhere that i assert this you're winding up with really purged packages and not 'deinstalled' (dpkg -r) packages so this never matches, but its not the same as what i was testing.

lamont-granquist: If you made this `match(/no info|not-installed/not installed/)` then I think it would work and you could use this for both the :purge case and this nil case.

lamont-granquist:
okay code looks good :+1:

PR #4264, opened 2015-12-09
Don't activate cheffish and chef-provisioning until its DSL is used

jkeiser:


lamont-granquist:
i'm still not super-wild about the implciations for the test-matrix between cheffish/chef-provisioning and chef-client and still think we should be pinning them on minor releases closer to chef-config.

but, yeah, this does the thing you want -- the simpler way of using a cheftie.rb will not delay gem resolution the way that you crave...

:+1:

Issue #4250, opened 2015-12-07
Chef 12.5 custom resources property collapse with built-in resources


maoueh:
Using Chef custom resources syntax introduced in Chef 12.5 series (ChefDK 0.10.0 Windows 7, x64, cookbook depends on `compat_resource ~> 12.5`). 

Assuming the following resource:

```ruby
# resources/custom.rb (cookbook simple)
property :name, String, name_property: true
property :revision, String

git '/tmp/repo' do
  repository 'git@url.com:project.git'
  revision revision
  action :sync
end

# recipes/default.rb
simple_custom 'test' do
  revision 'branch'
end
```

This will result in the following `git` resource:

```ruby
git '/tmp/repo' do
  repository 'git@url.com:project.git'
  revision 'HEAD'
  action :sync
end
```

Changing property `revision` to `revision_` make it work correctly. It seems that due to how custom resource scope is evaluated collapse with built-in properties somehow. It's unclear for me which one does not resolve correctly: the right-hand side or the left-hand side. I would guess the later but it's unclear.

My current workaround is to use `new_resource.revision`. If this is unfixable due to being a Ruby limitation, the documentation should have a note about this.

coderanger:
Indeed this isn't fixable AFAIK (ping @jkeiser) since `#revision` will always look at local real methods before scope promotion to the provider.

lamont-granquist:
this is kind of the same as #4006, but it appears to be even worse...

+1 on ping @jkeiser 

i'm starting to consider that actions should not be evaluated in the context of the resource and the user should be forced to type new_resource.revision in all cases...  the overlapping of namespaces is really pretty messy...

jkeiser:
Yep. sub-resources can't access properties of the parent with the same name. `new_resource.revision` is correct.

lamont-granquist:
closed by #4269

maoueh:
I think I might have hit another scoping problem, but seems highly more weird now. Currently tested with Chef 12.5.1 but I will try later with latest Chef to see if it's a problem or not. Assume this:

```
# Resource
resource_name 'my_resource'

property :name, name_property: true

action :execute do
  cookbook_file '/tmp/#{name}/custom.conf' do
    source "#{name}/custom.conf"
  end
end

# Recipe
my_resource 'mp3'
```

[...]

PR #4269, opened 2015-12-10
Warn when user sets a property of an inline resource to itself.

jkeiser:
(User will expect "x x" to grab the parent property.)

### DIFF
--- a/lib/chef/log.rb
+++ b/lib/chef/log.rb
@@ -37,6 +37,20 @@ class Chef
       end
     end
 
+    #
+    # Get the location of the caller (from the recipe). Grabs the first caller
+    # that is *not* in the chef gem proper (allowing us to weed out internal
+    # calls and give the user a more useful perspective).
+    #
+    # @return [String] The location of the caller (file:line#) from caller(0..20), or nil if no non-chef caller is found.
+    #
+    def self.caller_location
+      # Pick the first caller that is *not* part of the Chef gem, that's the
+      # thing the user wrote.
+      chef_gem_path = File.expand_path("../..", __FILE__)
+      caller(0..20).select { |c| !c.start_with?(chef_gem_path) }.first
+    end
+
     def self.deprecation(msg=nil, location=caller(2..2)[0], &block)
       if msg
         msg << " at #{Array(location).join("\n")}"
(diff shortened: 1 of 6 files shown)


lamont-granquist:
Yeah so this still doesn't address #4006 if I read this correctly, but we'll still need this one.  This would close #4250 

:+1:

lamont-granquist:
@chef/client-core needs another review :+1:

coderanger:
I don't really understand what case this is catching.

coderanger:
Ahh I think I see, this is the `self.x` vs enclosing provider's `new_resource.x` issue?

lib/chef/property.rb
```diff
@@ -292,6 +292,15 @@ def get(resource)
         value
 
       else
+        # If we are still compiling this resource (rather than running an action
+        # on it), reading this property before it's been set is likely the wrong
+        # thing. Warn if it's a duplicate of the enclosing provider.
+        if resource.respond_to?(:enclosing_provider) && resource.enclosing_provider && !resource.currently_running_action && !name_property?
+          if resource.enclosing_provider.respond_to?(name)
```
coderanger: Second nested condition could go on the `&&` chain above.

lib/chef/resource.rb
```diff
@@ -608,15 +609,18 @@ def run_action(action, notification_type=nil, notifying_resource=nil)
...
+    ensure
+      @currently_running_action = nil
+      @elapsed_time = Time.now - start_time
+      # Reporting endpoint doesn't accept a negative resource duration so set it to 0.
+      # A negative value can occur when a resource changes the system time backwards
+      @elapsed_time = 0 if @elapsed_time < 0
+      events.resource_completed(self)
     end
 
+    attr_reader :currently_running_action
```
coderanger: Docstring comment plz.

lib/chef/log.rb
```diff
@@ -37,6 +37,13 @@ def self.show_time=(*args)
       end
     end
 
+    def self.caller_location
```
coderanger: Should have a docstring comment or be marked private.

jkeiser:
@coderanger yep! That's the one.

I could actually fix it so it does what 99% of users expect ... but I think it's worth thinking *that* decision through some more before implementing.

coderanger:
@jkeiser I would probably put a much more descriptive comment about what that warning condition check is doing, wasn't obvious at first glance and it looks like something likely to break in the future.

PR #4271, opened 2015-12-10
Get rid of ambiguity with `template 'x' do ...`

jkeiser:
If a user has a property named `template` and then tries to declare a template with the same name, we silently set the property on the resource and ignore the block. This patch causes us to at least toss an exception.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -116,6 +116,10 @@ class Chef
       options[:instance_variable_name] = options[:instance_variable_name].to_sym if options[:instance_variable_name]
     end
 
+    def to_s
+      name
+    end
+
     #
     # The name of this property.
     #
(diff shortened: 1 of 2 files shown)


lib/chef/property.rb
```diff
@@ -448,18 +452,22 @@ def emit_dsl
       # stack trace if you use `define_method`.
       declared_in.class_eval <<-EOM, __FILE__, __LINE__+1
         def #{name}(value=NOT_PASSED)
+          raise "Property #{name} of \#{self} cannot be passed a block!" if block_given?
```
lamont-granquist: Since this is going to be hit when the user does the `property :template` and later calls `template "foo" do..end` this error message should probably contain more useful information about what to do in that case.  From the user's perspective this is kinda shitty because they typed a template resource and property is blowing up, and they're likely to WTF around for awhile before the lightbulb goes on...

coderanger:
This doesn't actually get rid of the ambiguity, it just makes it an error. I'm concerned about the number of band aids this syntax is needed to add that are leaking out into the non-fused-mode code.

jkeiser:
@coderanger this only applies when you are using fused-mode `action` code. The delegators that let you type `myproperty` are only added there.

jkeiser:
Oh! Yes.

We need to think it through more. Right now we do need to support what we've already got, but if there is any time to reevaluate, it's now.

lib/chef/property.rb
jkeiser: For now, I think I'll explain and let them know they need to rename the property. I'll work on a separate PR letting them use the recipe DSL unhooked from everything else. i.e. @coderanger's point of mixing DSLs; even if we do mix, we need to provide the DSLs separately in at least one form so users can disambiguate if they want.

jkeiser:
@coderanger however, I do not consider this a band aid in any form. Whether or not this ambiguity existed, specifying a block to a method that ignores it is a problem we can usefully inform the user of.

coderanger:
@jkeiser That is fair, this will also catch the long-dreaded `template { foo: 'bar' }` case which is enough to make me :+1: on its own :) The error message should probably include that case as possible fix though.

jkeiser:
Ooh! Will do.

jkeiser:
@lamont-granquist fixed the error message, I think. Added "If you meant to create a resource named #{name} instead, you'll need to first rename the property."

@coderanger as mentioned in IRC, looks like we can't catch that case because it is a compile error. Sad! That would have been nice.

lamont-granquist:
:+1:

PR #4273, opened 2015-12-10
Use signing protocol 1.1 by default

jaym:
All supported Chef servers support the 1.1 signing protocol.
There is no reason to continue using 1.0, and removing it
as the default allows us to remove a bunch of code that tries
to upgrade 1.0 to 1.1 when the node name is too long.

If the user specifies 1.0 as the auth protocol version from
this point on, they will have to guarantee that the node
name is not too long.

cc @chef/client-core


danielsdeleo:
:+1: when travis is happy

btm:
yay! :+1:

thommay:
:+1:

PR #4277, opened 2015-12-11
non msi packages must explicitly provide a source attribute on install

mwrock:
Currently one can use a package `source` location as a package name:
```
package 'http://mercurial.selenic.com/release/windows/Mercurial-3.6.1-x64.exe' do
  checksum 'febd29578cb6736163d232708b834a2ddd119aa40abc536b2c313fc5e1b5831d'
end
```

In the case of non-`:msi` `installer_type`s, this will result in package resources reconverging an `:install` action even though the package has been previously installed.

The reason is that when the `windows_package` scans the registry for all current software installations, it cannot search by install path or url. So it searches by the `DisplayName` property in the registry keys mentioned above. If you use a source path as the name of a non `:msi` `windows_package` it will not find any currently installed package and will therefore attempt to reinstall the package. `:msi` packages get around this because there is a well known API for extracting the id of the installed MSI package from a `.msi` file.

Requiring an explicit `source` for non `:msi` packages will reduce confusion here around idempotency.

### DIFF
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -88,11 +88,19 @@ to better align with other ssh related options.
 
 `windows_package` now supports more than just `MSI`. Most common windows installer types are supported including Inno Setup, Nullsoft, Wise and InstallShield. The new allowed `installer_type` values are: `inno`, `nsis`, `wise`, `installshield`, `custom`, and `msi`.
 
+**Non `:msi` package installation**
+When installing non `:msi` packages, the `package_name` should match the display name used for the installed package in the "Add/Remove Programs" settings application. This value can also be found in the following registry locations:
+* HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Uninstall
+* HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Uninstall
+* HKEY_LOCAL_MACHINE\Software\Wow6464Node\Microsoft\Windows\CurrentVersion\Uninstall
+
+ Further, non `:msi` packages must explicitly include a package `source` attribute when using the `:install` action. Unlike `:msi` packages, they will not default to the package name if missing. Without the name matching the software's display name, non `:msi` packages will always reconverge on `:install`.
+
 Also, while being able to download remote installers from a `HTTP` resource is not new, it looks as though the top of the docs page is incorrect stating that only local installers can be used as a source.
 
 Example Nullsoft (`nsis`) package resource:
 ```
-windows_package 'Mercurial 3.6.1 (64-bit)' do
+package 'Mercurial 3.6.1 (64-bit)' do
   source 'http://mercurial.selenic.com/release/windows/Mercurial-3.6.1-x64.exe'
   checksum 'febd29578cb6736163d232708b834a2ddd119aa40abc536b2c313fc5e1b5831d'
 end
@@ -100,7 +108,7 @@ end
 
 Example Custom `windows_package` resource:
 ```
-windows_package 'Microsoft Visual C++ 2005 Redistributable' do
+package 'Microsoft Visual C++ 2005 Redistributable' do
   source 'https://download.microsoft.com/download/6/B/B/6BB661D6-A8AE-4819-B79F-236472F6070C/vcredist_x86.exe'
   installer_type :custom
   options '/Q'
@@ -108,15 +116,18 @@ end
 ```
 Using a `:custom` package is one way to install a non `.msi` file that embeds an `msi` based installer.
 
+**`windows_package` removal**
 Packages can now be removed without the need to include the package `source`. The relevent uninstall metadata will now be discovered from the registry.
 ```
-windows_package 'Mercurial 3.6.1 (64-bit)' do
+package 'Mercurial 3.6.1 (64-bit)' do
   action :remove
 end
 ```
-It is important that the package name used when not including the `source` is EXACTLY the same as the display name found in "Add/Remove programs" or the `DisplayName` property in the appropriate registry key:
+For non `:msi` packages, it is important that the package name used is EXACTLY the same as the display name found in "Add/Remove programs" or the `DisplayName` property in the appropriate registry key:
 * HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Uninstall
 * HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Uninstall
 * HKEY_LOCAL_MACHINE\Software\Wow6464Node\Microsoft\Windows\CurrentVersion\Uninstall
+
+When removing `:msi` packages, this same package naming rule applies if the `source` is omitted.
 
 Note that if there are multiple versions of a package installed with the same display name, all packages will be removed unless a version is provided in the `version` attribute or can be discovered in the `source` installer file.
--- a/lib/chef/provider/package/windows.rb
+++ b/lib/chef/provider/package/windows.rb
@@ -34,9 +34,16 @@ class Chef
 
         require 'chef/provider/package/windows/registry_uninstall_entry.rb'
 
+        def define_resource_requirements
+          requirements.assert(:install) do |a|
+            a.assertion { new_resource.source unless package_provider == :msi }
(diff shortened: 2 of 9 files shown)


chefsalim:
:+1:

btm:
```
def uninstall_registry_entries
  @uninstall_registry_entries ||= Chef::Provider::Package::Windows::RegistryUninstallEntry.find_entries(new_resource.name)
end
```

I think this should look at new_resource.package_name so that the name can be something less crazy, e.g.

```
package "remove the burning foxes" do
  package_name "FireFox Internet Browser of Webs v11.2.0 sp12"
  action :uninstall
end
```

It'd actually be way more intuitive if this was registry_name or registry_key or something. hrm.

DOC_CHANGES.md
```diff
@@ -83,11 +83,13 @@ to better align with other ssh related options.
 
 `windows_package` now supports more than just `MSI`. Most common windows installer types are supported including Inno Setup, Nullsoft, Wise and InstallShield. The new allowed `installer_type` values are: `inno`, `nsis`, `wise`, `installshield`, `custom`, and `msi`.
 
+Non `:msi` package names should match the softwares `DisplayName` property expected to be found in the registry keys mentioned below. Further, they must explicitly include a package `source` attribute when using the `:install` action. It will not default to the package name if missing. Without the name matching the software display name, non `:msi` packages will always reconverge on `:install`.
```
btm: `DisplayName` is a registry key property, not a chef resource property. we should be more clear about that. It's not that we don't say that, but `DisplayName` sticks out (besides being camel case) to me as something I might set on my resource.

For proper behavior, the `package_name` property for Windows packages other than MSIs must be set to match ...

On second thought, we should remove/combine this paragraph with the one below that says almost the same thing, lists the registry key locations, but also mentions that the add/remove name might be more accessible.

lib/chef/provider/package/windows.rb
```diff
@@ -129,6 +129,7 @@ def action_install
         # Chef::Provider::Package action_install + action_remove call install_package + remove_package
         # Pass those calls to the correct sub-provider
         def install_package(name, version)
+          fail Chef::Exceptions::NoWindowsPackageSource, "installer_type '#{installer_type}' requires a source" if new_resource.source.nil?
```
btm: We do in fact require a source for install, the catch is that in the cookbook you could provide a source url as name, and it would work, but wouldn't be idempotent.

The meaningful "fix" here is probably in the cookbook. We change the cookbook to evaluate the `name` property to see if it is a URI, and if it is, move it to `source` before we give them the core provider but WARN that they need to fix that with a nice friendly message about how it wouldn't have worked.

what about this:

```
  requirements.assert(:install) do |a|
    a.assertion { @new_resource.source unless @package_provider == :msi }
    a.failure_message Chef::Exceptions::Package, "Source for package #{@new_resource.name} must be specified in the resource's source property for package to be installed because the package_name property is used to test for package installation state for this package type."
  end
end
```

mwrock: > We do in fact require a source for install, the catch is that in the cookbook you could provide a source url as name, and it would work, but wouldn't be idempotent.

In core we do not require source, but w/o this PR the same logic as the cookbook applies defaulting the source to the name.

Not sure I get your comment about the fix being in the cookbook. The goal here is to fix core assuming the cookbook is not used.

I do like using the `requirement.assert` (forgot that existed) over the fail.

mwrock:
One thing I noticed just a bit ago is that the cookbook does require an explicit source and does not defaut  a missing source to be the name. That was one of the slight differences in the resource that did not copy over. So this PR should better mimic what we had in the cookbook.

lib/chef/provider/package/windows.rb
mwrock: Not sure the `requirement_assert` will work here because the provider must be evaluated first which soes not occur until during the action. Just not sure how that ordering is setup but will play with it.

btm: If you look at some of the other providers you'll see that they have assertions that also depend on values that come from load_current_resource, which in this case contains a call to package_provider, so we should know by then.

btm: > Not sure I get your comment about the fix being in the cookbook. The goal here is to fix core assuming the cookbook is not used.

The fix in behavior, e.g., making it idempotent is here. but if the cookbook didn't require source (apparently it does), and core did, the most pain would be in the transition for existing users. So that's the biggest band-aid/fix/hug. It's kind of moot of the behavior is fairly similar. We probably should have made a behavior matrix somewhere on paper earlier in the week.

mwrock:
I've adjusted to review comments.

lib/chef/provider/package/windows.rb
```diff
@@ -34,6 +34,13 @@ class Windows < Chef::Provider::Package
 
         require 'chef/provider/package/windows/registry_uninstall_entry.rb'
 
+        def define_resource_requirements
+          requirements.assert(:install) do |a|
+            a.assertion { @new_resource.source unless package_provider == :msi }
```
btm: Use the `new_resource` accessor rather than the instance variable.

lib/chef/provider/package/windows/exe.rb
```diff
@@ -99,7 +99,7 @@ def current_installed_version
 
           def install_file_version
             @install_file_version ||= begin
-              if ::File.exist?(@new_resource.source)
+              if !@new_resource.source.nil? && ::File.exist?(@new_resource.source)
```
btm: use the `new_resource` reader.

lib/chef/provider/package/windows/msi.rb
```diff
@@ -44,7 +44,7 @@ def expand_options(options)
 
           # Returns a version if the package is installed or nil if it is not.
           def installed_version
-            if ::File.exist?(new_resource.source)
+            if !@new_resource.source.nil? && ::File.exist?(new_resource.source)
```
btm: `new_resource` everywhere.

btm:
:+1: check the use of instance variables vs accessors, otherwise :trophy:

mwrock:
argh well thats embarassing. should have caught those. consider them smitten

btm:
:+1:

lib/chef/provider/package/windows.rb
```diff
@@ -34,9 +34,16 @@ class Windows < Chef::Provider::Package
 
         require 'chef/provider/package/windows/registry_uninstall_entry.rb'
 
+        def define_resource_requirements
+          requirements.assert(:install) do |a|
+            a.assertion { new_resource.source unless package_provider == :msi }
+            a.failure_message Chef::Exceptions::NoWindowsPackageSource, "Source for package #{new_resource.name} must be specified in the resource's source property for package to be installed because the package_name property is used to test for package installation state for this package type."
```
chefsalim: This message could probably be shortened :)
Also, I think the sentence is missing a 'the' in "source property for package"

btm: Yeah, it's a little tough. MSI doesn't require source, so I think this is an attempt at proactive documentation.

mwrock: Not seeing the missing "the" but to @btm's point, this is tough to get just right and make it concise. We could just say that the source is required and leave it at that but if the user uses a url/path as the package_name, then they might be confused why they need to specify a `source` which they don't have to do for MSIs.

btm: I guess you could add a "the" here:

"Source for package #{new_resource.name} must be specified in the resource's source property for package to be installed because the package_name property is used to test for **the** package installation state for this package type."

chefsalim:
:+1:  again

chefsalim:
Also Matt, you may need to wait to update the DOC changes for the required source, as those will be for 12.7

mwrock:
do we do doc changes for patch releases?

chefsalim:
Per James, we should update the doc_changes if the change is significant enough.  
(Note: the process _may_ change for 12.7)

mwrock:
There was some merge conflict fallout here I had to address with the `properties` changes. There is still one app veyor test failing but it is related to easy_install and completely unrelated to this PR. So merging.

PR #4278, opened 2015-12-11
make file resource use properties

lamont-granquist:
Start dogfooding properties

### DIFF
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -27,8 +27,6 @@ class Chef
     class File < Chef::Resource
       include Chef::Mixin::Securable
 
-      identity_attr :path
-
       if Platform.windows?
         # Use Windows rights instead of standard *nix permissions
         state_attrs :checksum, :rights, :deny_rights
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -53,7 +53,7 @@ class Chef
       property :content, [ String, nil ], desired_state: false
       property :backup, [ Integer, false ], desired_state: false, default: 5
       property :checksum, [ String, nil ], regex: /^[a-zA-Z0-9]{64}$/
-      property :path, String, name_property: true
+      property :path, String, name_property: true, identity: true
       property :diff, [ String, nil ], desired_state: false
       property :atomic_update, [ true, false ], desired_state: false, default: Chef::Config[:file_atomic_update]
       property :force_unlink, [ true, false ], desired_state: false, default: false
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -53,7 +53,7 @@ class Chef
       property :content, [ String, nil ], desired_state: false
       property :backup, [ Integer, false ], desired_state: false, default: 5
       property :checksum, [ String, nil ], is: /^[a-zA-Z0-9]{64}$/
-      property :path, [ String ], name_property: true
+      property :path, String, name_property: true
       property :diff, [ String, nil ], desired_state: false
       property :atomic_update, [ true, false ], desired_state: false, default: Chef::Config[:file_atomic_update]
       property :force_unlink, [ true, false ], desired_state: false, default: false
--- a/lib/chef/resource/file.rb
+++ b/lib/chef/resource/file.rb
@@ -52,7 +52,7 @@ class Chef
 
       property :content, [ String, nil ], desired_state: false
       property :backup, [ Integer, false ], desired_state: false, default: 5
-      property :checksum, [ String, nil ], regex: /^[a-zA-Z0-9]{64}$/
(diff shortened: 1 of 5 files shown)


lib/chef/resource/file.rb
```diff
@@ -52,78 +52,17 @@ class File < Chef::Resource
...
-        )
-      end
+      property :content, [ String, NilClass ], desired_state: false
+      property :backup, [ Integer, FalseClass ], desired_state: false, default: 5
+      property :checksum, [ String, NilClass ], is: /^[a-zA-Z0-9]{64}$/
+      property :path, [ String ], name_property: true
+      property :diff, [ String, NilClass ], desired_state: false
+      property :atomic_update, [ TrueClass, FalseClass ], desired_state: false, default: Chef::Config[:file_atomic_update]
+      property :force_unlink, [ TrueClass, FalseClass ], desired_state: false, default: false
+      property :manage_symlink_source, [ TrueClass, FalseClass ], desired_state: false
```
jkeiser: This can be `property :manage_symlink_source, [ true, false ], desired_state: false`. Same trick works with `NilClass` -> `nil`.

lib/chef/resource/file.rb
```diff
@@ -52,78 +52,17 @@ class File < Chef::Resource
 
       def initialize(name, run_context=nil)
         super
-        @path = name
-        @backup = 5
-        @atomic_update = Chef::Config[:file_atomic_update]
-        @force_unlink = false
-        @manage_symlink_source = nil
-        @diff = nil
         @verifications = []
```
jkeiser: Don't block this on my account, but it seems like `verifications` is the real property and `verify` is a setter (appender) method.

jkeiser:
Note: `Chef::Mixin::Securable` will need to be done as well. `Chef::Mixin::Properties` should give you the DSL you need in there; include will bring the properties in correctly.

jkeiser:
:+1: on this though, whether you make the other changes or not.

lib/chef/resource/file.rb
```diff
@@ -52,78 +52,17 @@ class File < Chef::Resource
...
-        set_or_return(
-          :manage_symlink_source,
-          arg,
-          :kind_of => [ TrueClass, FalseClass ]
-        )
-      end
+      property :content, [ String, nil ], desired_state: false
+      property :backup, [ Integer, false ], desired_state: false, default: 5
+      property :checksum, [ String, nil ], is: /^[a-zA-Z0-9]{64}$/
+      property :path, [ String ], name_property: true
```
jkeiser: One more thing worth noting: this can be written

`property :path, String, name_property: true`

lamont-granquist:
@chef/client-core passes on my desktop now and /should/ pass in appveyor/travis unless there's a sunspot...  ready for review

lamont-granquist:
@jkeiser can you look at those last three commits, i have questions, in order:

- is that the right way to create a new Array every time for a default?
- was i using the `is:` syntax wrong or is that the correct fix?
- why the hell is `identity` different from `name`?  should we be specifying both of those everywhere?

lamont-granquist:
okay done tweaking for realz now

jkeiser:
I think we've covered everything there, and this looks good! :+1:

lamont-granquist:
okay finally made windows happy.

@chef/client-core now this is really 100% ready for review

lib/chef/resource/file.rb
```diff
@@ -50,80 +48,15 @@ class File < Chef::Resource
...
-      end
-
-      def manage_symlink_source(arg=nil)
-        set_or_return(
-          :manage_symlink_source,
-          arg,
-          :kind_of => [ TrueClass, FalseClass ]
-        )
-      end
+      property :content, [ String, nil ], desired_state: false
```
coderanger: This is going to apply to all the similar pull reqs, but we should probably be more explicit that we are okay with not using parens here. The usual style for the rest of the Chef code is to use parens in most cases, and these classes are complex enough that I don't think anyone would mistake them for a pseudo-declarative thang. The plus side is it would be nice to have the Chef code look style-wise like we tell people to write LWRPs.

coderanger: Also on the list for style-lawyering, I usually alpha-sort my properties with the name prop at the top. Would be good to have a standard as this gets rolled out.

lamont-granquist: Yeah, I think for property calls we generally shouldn't use them.  There might be some case where for operator precedence issues it might make it clearer to wrap it with parens, but I'd bet in those case it'd be more readable to
rethink the expression.

Agreed on alpha-sorting, I nearly did that, but defaulted to the order they were already in...

ranjib:
LGTM

thommay:
:+1:

PR #4279, opened 2015-12-11
Use properties for packages

jkeiser:


lamont-granquist:
:+1:

thommay:
you made appveyor mad :( code looks good modulo that, though.

jkeiser:
@lamont-granquist could use a re-review for the second commit; I had to go in and fix a lot of resources to stop setting `version` to `nil`. Seems like the right thing though, especially since in a lot of cases we were *explicitly* setting it to `nil`--obviously we thought it would do something.

One thing I'm not sure of is if there are some places there where that was covering a bug--where it might have a version and we wanted to clear it to `nil`, and therefore need to do a `reset_property(:version)` to clear the version. I don't feel like `nil` is a valid version ever, so I don't think we should make it an allowable value personally--the presence or absence of the property should be enough.

jkeiser:
@thommay yeah, there was a bug in windows package still! Yay appveyor! I've updated.

lib/chef/provider/package/aix.rb
```diff
@@ -47,7 +47,6 @@ def define_resource_requirements
         def load_current_resource
           @current_resource = Chef::Resource::Package.new(@new_resource.name)
           @current_resource.package_name(@new_resource.package_name)
-          @new_resource.version(nil)
```
lamont-granquist: holy shit that is some bad code right there that you deleted.

lib/chef/provider/package/easy_install.rb
```diff
@@ -94,10 +93,10 @@ def load_current_resource
 
           if package_version == @new_resource.version
             Chef::Log.debug("#{@new_resource} at version #{@new_resource.version}")
-            @current_resource.version(@new_resource.version)
+            @current_resource.version(@new_resource.version) if @new_resource.version
```
lamont-granquist: The argument that I'd have here is that nil is perfectly valid, and is the default that we must have on the new_resource in order to indicate that the user had no preference.  We are copying those values to the current_resource, and I think you're getting fairly lost in philosophical semantics here.  You should add 'nil' to the allowed values, and then not only don't you need to guard every single one of these, but the user can 'un-set' a value back to the cleared default.  Otherwise there's no way to clear something once you've set it (which is hugely annoying in spec tests for one thing).

lamont-granquist:
so generally if the default is a nil, i want nil added to the validation specs, and i feel extremely strongly about this as a practical matter.  you're making everyone protect all their arguments like that when copying between new_resource and current_resource everywhere and then forcing people to go through reset_value to get back to the default, and i'm pretty much confident that means you've got your philosophy incorrect.   stop thinking that hard about what 'nil' means and it'll make life easier for everyone...

lib/chef/provider/package/easy_install.rb
jkeiser: @lamont-granquist if the only purpose is to denote the presence or absence of a version, I'd argue for @current_resource.reset_property(:version) in the few cases where you want to *remove* a version (though I'm curious what those cases are).

I get the argument about guards for sure--I'm starting to like @dan's `NoContent` suggestion more and more for that reason.

jkeiser:
I think we should find a way for `nil` to be a valid value, simply because it often is. I'm willing to change this one for now, now that we've gotten rid of the bad code, but I think we need to explore the `NoContent` route harder, and solve this problem once and for all.

lamont-granquist:
@jkeiser you force pushed something really odd here...

jkeiser:
@lamont-granquist I have force pushed what should be the final (or damn near). I added `nil` as a valid value for all properties (with a callback to the bug in question so we can fix them), and removed the conditional sets. I still removed the spots where we set the value to `nil` explicitly, since that does nothing.

PR #4280, opened 2015-12-11
Use properties for service

jkeiser:


jkeiser:
Hrm. Still interested. I'll try to get this updated.

lamont-granquist:
another one that i'll pick up

tas50:
We really do want to do this, but at this point we're looking to take another angle.

PR #4286, opened 2015-12-11
bump up timeout for travis

lamont-granquist:


lamont-granquist:
Even if this burns roughly 100 extra seconds (4 failed tests * 25 seconds extra per test), that is vastly less than every time this timeout kicks off and we click 'reload' on a PR and we suck up a queue slot for another 20 minutes or so...

Maybe there's a better fix, but this is better than clicking reload...

smurawski:
:+1:

jkeiser:
:+1:

jkeiser:
File a bug on it sometimes being super slow?
Also, might not fix it--it could be the issue is the server not starting at all.

lamont-granquist:
I figure the fastest way to find out if the server isn't starting is to merge this...

lamont-granquist:
(i'd also argue that this patch is correct anyway -- we should always be patient before timing out tests and forcing the user to click reload in the travis UI which is horribly expensive)

lamont-granquist:
lol, dammit wrong branch...

PR #4289, opened 2015-12-11
Improve validation for proxy in http_client_builder

### LINKED ISSUE #4252: http_proxy in Chef::HTTP could have better validation
```
  * remote_file[/var/chef/cache/elastic.gpg] action create

      Error executing action `create` on resource 'remote_file[/var/chef/cache/elastic.gpg]'

      NoMethodError
      -------------
      undefined method `host' for "":String

      Resource Declaration:
      ---------------------
      # In /var/chef/cache/cookbooks/apt/providers/repository.rb

       90:     remote_file cached_keyfile do
       91:       source new_resource.key
       92:       mode 00644
       93:       sensitive new_resource.sensitive if respond_to?(:sensitive)
       94:       action :create
       95:     end
       96:   else

      Compiled Resource:
      ------------------
      # Declared in /var/chef/cache/cookbooks/apt/providers/repository.rb:90:in `install_key_from_uri'

      remote_file("/var/chef/cache/elastic.gpg") do
        provider Chef::Provider::RemoteFile
        action [:create]
        retries 0
        retry_delay 2
        path "/var/chef/cache/elastic.gpg"
        backup 5
        atomic_update true
        source ["http://10.1.24.15/mirror/packages.elastic.co/elastic.gpg"]
        use_etag true
        use_last_modified true
        declared_type :remote_file
        cookbook_name "neteng"
        mode 420
      end


    Error executing action `add` on resource 'apt_repository[elasticsearch]'

    NoMethodError
    -------------
    remote_file[/var/chef/cache/elastic.gpg] (/var/chef/cache/cookbooks/apt/providers/repository.rb line 90) had an error: NoMethodError: undefined method `host' for "":String

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/neteng/recipes/elasticsearch.rb

     13: apt_repository 'elasticsearch' do
     14:   uri db['elasticsearch']['repo']
     15:   components ['main']
     16:   distribution 'stable'
     17:   key db['elasticsearch']['repo_gpg']
     18:   arch 'amd64'
     19:   action :add
     20: end
     21: 

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/neteng/recipes/elasticsearch.rb:13:in `from_file'

    apt_repository("elasticsearch") do
      action [:add]
      retries 0
      retry_delay 2
      declared_type :apt_repository
      cookbook_name "neteng"
      recipe_name "elasticsearch"
      uri "http://10.1.24.15/mirror/packages.elastic.co/elasticsearch/2.x/debian/"
      components ["main"]
      distribution "stable"
      key "http://10.1.24.15/mirror/packages.elastic.co/elastic.gpg"
      arch "amd64"
      cache_rebuild true
    end
```

smith:
Checks to make sure the object used for proxy\_uri is something we can
use to create a proxy object.

Fixes #4252.

### DIFF
--- a/lib/chef/http/basic_client.rb
+++ b/lib/chef/http/basic_client.rb
@@ -103,14 +103,16 @@ class Chef
         # Check if the proxy string contains a scheme. If not, add the url's scheme to the
         # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
         # here since we are really just trying to get the string built correctly.
-        if String === proxy && !proxy.strip.empty?
+        #
+        # If `proxy` is not a string or is an empty string, the value will be nil.
+        proxy = if proxy.is_a?(String) && !proxy.strip.empty?
           if proxy.match(/^.*:\/\//)
-           proxy = URI.parse(proxy.strip)
+            proxy = URI.parse(proxy.strip)
           else
-           proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
-          end 
+            proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
+          end
         end
-        
+
         no_proxy = Chef::Config[:no_proxy] || env['NO_PROXY'] || env['no_proxy']
         excludes = no_proxy.to_s.split(/\s*,\s*/).compact
         excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
--- a/spec/unit/http/basic_client_spec.rb
+++ b/spec/unit/http/basic_client_spec.rb
@@ -39,19 +39,57 @@ describe "HTTP Connection" do
     end
   end
 
+  describe "#http_client_builder" do
+    subject(:http_client_builder) { basic_client.http_client_builder }
+
+    context "when the http_proxy is an URI" do
+      before :each do
+        allow(basic_client).to receive(:proxy_uri).and_return(URI.parse(
+          "http://user:pass@example.com:1234"
+        ))
+      end
+
+      it "has a proxy_address" do
+        expect(http_client_builder.proxy_address).to eq "example.com"
+      end
+
+      it "has a proxy_pass" do
+        expect(http_client_builder.proxy_pass).to eq "pass"
+      end
+
+      it "has a proxy_port" do
+        expect(http_client_builder.proxy_port).to eq 1234
+      end
+
+      it "has a proxy_user" do
+        expect(http_client_builder.proxy_user).to eq "user"
+      end
+    end
+
+    context "when the http_proxy is nil" do
+      before :each do
+        allow(basic_client).to receive(:proxy_uri).and_return(nil)
+      end
+
+      it "returns Net::HTTP" do
+        expect(basic_client.http_client_builder).to eq Net::HTTP
+      end
+    end
+  end
+
   describe "#proxy_uri" do
+    subject(:proxy_uri) { basic_client.proxy_uri }
+
     shared_examples_for "a proxy uri" do
       let(:proxy_host) { "proxy.mycorp.com" }
       let(:proxy_port) { 8080 }
       let(:proxy) { "#{proxy_prefix}#{proxy_host}:#{proxy_port}" }
 
       it "should contain the host" do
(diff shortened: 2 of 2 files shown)


lib/chef/http/basic_client.rb
```diff
@@ -139,7 +139,8 @@ def env
 
       def http_client_builder
         http_proxy = proxy_uri
-        if http_proxy.nil?
+        if !http_proxy.respond_to?(:host) || !http_proxy.respond_to?(:port) ||
+           !http_proxy.respond_to?(:user) || !http_proxy.respond_to?(:password)
```
lamont-granquist: `if !http_proxy.is_a?(URI)`

or for bonus points drop the '!' and invert the if-else clauses

the #proxy_uri method also looks like it should be documented that it returns either a URI object or it looks like...  false?  and tests added around that if they have to be...

lamont-granquist: actually it should return nil, and as far as i can see the old code should pass the validation and it looks like some code might have gotten snagged from another project and stuffed into `#proxy_uri` which already fixed this use case?

lamont-granquist:
another thing to consider is that if http_proxy is set but not parsable into a URI that we should probably raise on that since its likely malformed input and that falling back to non-proxy behavior is most likely not what is wanted...

in the case of "" you might still want to fall back to non-proxy behavior, but in the case of "httpss://10.1.1.1" or something you probably want to raise.

smith:
@lamont-granquist updated this so that we return `nil` if the proxy is anything other than a non-empty string, which should fix the problem the reporter was having.

lib/chef/http/basic_client.rb
```diff
@@ -103,14 +103,17 @@ def proxy_uri
...
-           proxy = URI.parse(proxy.strip)
+            proxy = URI.parse(proxy.strip)
           else
-           proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
-          end 
+            proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
+          end
+        # If proxy is anything other than a non-empty string, set it to nil
+        else
+          proxy = nil
```
lamont-granquist: if the if failed, though, proxy would be unset and be nil anyway wouldn't it?

lib/chef/http/basic_client.rb
```diff
@@ -103,14 +103,17 @@ def proxy_uri
...
         # here since we are really just trying to get the string built correctly.
-        if String === proxy && !proxy.strip.empty?
+        if proxy.is_a?(String) && !proxy.strip.empty?
           if proxy.match(/^.*:\/\//)
-           proxy = URI.parse(proxy.strip)
+            proxy = URI.parse(proxy.strip)
           else
-           proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
-          end 
+            proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
```
lamont-granquist: what if either of these URI.parse statements raise?

lib/chef/http/basic_client.rb
smith: It's initially set on line 100, where it could end up being an empty string, which is what happened to the reporter.

lib/chef/http/basic_client.rb
smith: A bad URI is raised, which I think is what was happening before, and the error message would be tolerable.

lib/chef/http/basic_client.rb
lamont-granquist: okay what about:

```ruby
proxy = if proxy.is_a?(String) && !proxy.strip.empty?
  if proxy.match(/^.*:\/\//)
    URI.parse(proxy.strip)
  else
    URI.parse("#{url.scheme}://#{proxy.strip}")
  end
end
```

The return value when the outer if fails is 'nil'.

smith:
@lamont-granquist this has been updated based on your comments.

lamont-granquist:
@chef/client-core i'm on vacay, can anyone else pick this up?

smith:
No rush. I'm working on another PR now that might make this one unnecessary.

lamont-granquist:
@smith closable?

PR #4291, opened 2015-12-11
Reduce appveyor matrix

jaym:
Our tests are getting backed up. 2 hours is unreasonable.
Removing ruby 2.0 32 bit from the matrix


lamont-granquist:
:+1:

PR #4295, opened 2015-12-11
Implement action_class do ... end

jkeiser:


coderanger:
Predicting future "well it worked for me before" regressions, we should test running it twice for the same class and test it with subclassing.

jkeiser:
subclass permutations are already tested elsewhere, but it's easy enough to build one here, so here ya go.

jkeiser:
@coderanger updated, I *think* it has the set of tests you were talking about now.

coderanger:
:+1:

PR #4296, opened 2015-12-12
Don't warn when setting a property to nil unless its value would actually change

jkeiser:
Don't warn when setting a property to nil unless its value would actually change. Gets rid of cases where we are initializing a resource with values from another resource.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -251,16 +251,19 @@ class Chef
         return get(resource)
       end
 
-      # myprop nil is sometimes a get (backcompat)
       if value.nil? && !explicitly_accepts_nil?(resource)
-        # If you say "my_property nil" and the property explicitly accepts
-        # nil values, we consider this a get.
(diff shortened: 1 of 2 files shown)


jkeiser:
@thommay @lamont-granquist @coderanger @danielsdeleo this looks ready to go. I think this will actually get rid of 99% of the warnings about setting things to `nil` and allow for some more pleasant resources that don't have to declare themselves as accepting `nil`. (It's actually rare for people to overwrite the value, most of the time it's initialization anyway).

thommay:
:+1:

stevendanna:
:+1: I think this will be very helpful.

lamont-granquist:
so i think what this doesn't capture is that we want to be able to set some resources in Chef-12 to explicitly accept nil -- but have their behavior still be that they are getters in Chef-12 -- and then switch their behavior to setters in Chef-13.  so we need to log a deprecation warning in that case.

i think...  or else i've got that backwards...  but i think you're clearly missing a case here.

this also worries me a little that we just turned some properties into nil-setters on the file resource in 12.6.0 and we'll get bugs about that, if i'm reading it correctly...

jkeiser:
This was done in a separate checkin. Closing.

PR #4297, opened 2015-12-13
Windows alternate user support for execute resources

adamedx:
### Windows alternate user identity execute support

The `execute` resource and simliar resources such as `script`, `batch`, and `powershell_script`
now support the specification of credentials on Windows so that the resulting process
is created with the security identity that corresponds to those credentials. This gives these resources on Windows the same capability they have on *nix systems to run processes as an alternate identity.
#### Properties

The following properties are new or updated for the `execute`, `script`, `batch`, and
`powershell_script` resources and any resources derived from them:
-   `user`</br>
  **Ruby types:** String</br>
  The user name of the user identity with which to launch the new process.
  Default value: `nil`. The user name may optionally be specifed
  with a domain, i.e. `domain\user` or `user@my.dns.domain.com` via Universal Principal Name (UPN)
  format. It can also be specified without a domain simply as `user` if the domain is
  instead specified using the `domain` attribute. On Windows only, if this property is specified, the `password`
  property **must** be specified.
-   `password`</br>
  **Ruby types** String</br>
  _Windows only:_ The password of the user specified by the `user` property.
  Default value: `nil`. This property is mandatory if `user` is specified on Windows and may only
  be specified if `user` is specified. The `sensitive` property for this resource will
  automatically be set to `true` if `password` is specified.
-   `domain`</br>
  **Ruby types** String</br>
  _Windows only:_ The domain of the user user specified by the `user` property.
  Default value: `nil`. If not specified, the user name and password specified
  by the `user` and `password` properties will be used to resolve
  that user against the domain in which the system running Chef client
  is joined, or if that system is not joined to a domain it will resolve the user
  as a local account on that system. An alternative way to specify the domain is to leave
  this property unspecified and specify the domain as part of the `user` property.

### DIFF
--- a/lib/chef/mixin/user_identity.rb
+++ b/lib/chef/mixin/user_identity.rb
@@ -1,0 +1,81 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Mixin
+    module UserIdentity
+
+      def validate_identity(specified_user, password = nil, specified_domain = nil)
+        validate_identity_platform(specified_user, password, specified_domain)
+        validate_identity_syntax(specified_user, password, specified_domain)
+      end
+
+      def validate_identity_platform(specified_user, password = nil, specified_domain = nil)
+        if ! Chef::Platform.windows?
+          if password || specified_domain
+            raise Exceptions::UnsupportedPlatform, "Values for `domain` and `password` are only supported on the Windows platform"
+          end
+        else
+          if specified_user && password.nil?
+            raise ArgumentError, "A value for `password` must be specified when a value for `user` is specified on the Windows platform"
+          end
+        end
+      end
+
+      def validate_identity_syntax(specified_user, password = nil, specified_domain = nil)
+        identity = qualify_user(specified_user, specified_domain)
+
+        if ( password || identity[:domain] ) && identity[:user].nil?
+          raise ArgumentError, "A value for `password` or `domain` was specified without specification of a value for `user`"
+        end
+      end
+
+      def qualify_user(specified_user, specified_domain = nil)
+        domain = specified_domain
+        user = specified_user
+
+        if specified_user.nil? && ! specified_domain.nil?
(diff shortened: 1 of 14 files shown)


lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,81 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
tyler-ball: 2016

lib/chef/provider/execute.rb
```diff
@@ -19,15 +19,22 @@
...
       extend Forwardable
 
+      include Chef::Mixin::UserIdentity
+
       provides :execute
 
-      def_delegators :@new_resource, :command, :returns, :environment, :user, :group, :cwd, :umask, :creates
+      def_delegators :@new_resource, :command, :returns, :environment, :user, :domain, :password, :group, :cwd, :umask, :creates
+
+      def initialize(new_resource, run_context)
```
tyler-ball: This is unnecessary if all it does is call `super`

lib/chef/provider/script.rb
```diff
@@ -69,7 +70,38 @@ def set_owner_and_group
...
-        FileUtils.chown(new_resource.user, new_resource.group, script_file.path)
+        if ! Chef::Platform.windows?
+          FileUtils.chown(new_resource.user, new_resource.group, script_file.path)
+        else
+          grant_alternate_user_read_access
+        end
+      end
+
+      def grant_alternate_user_read_access
+        return if new_resource.user.nil?
```
tyler-ball: So in a Windows world, if `user` is nil it will just keep the same owner as whoever create it, correct?

adamedx: Yes, that's correct (and that's what is currently happening in Chef + Windows today for script resources).

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,81 @@
...
+          end
+        end
+
+        { domain: domain, user: user }
+      end
+
+      protected(:validate_identity)
+      protected(:validate_identity_platform)
+      protected(:validate_identity_syntax)
+      protected(:qualify_user)
```
lamont-granquist: `protected` is super-weird in ruby, you probably mean `private` here (`private` methods are still accessible from subclasses).

lamont-granquist: http://stackoverflow.com/questions/3534449/why-does-ruby-have-both-private-and-protected-methods

"It is important to note that this is different from the way Java or C++ works. private in Ruby is similar to protected in Java/C++ in that subclasses have access to the method. In Ruby, there is no way to restrict access to a method from its subclasses like you can with private in Java."

lib/chef/provider/execute.rb
```diff
@@ -52,6 +55,7 @@ def timeout
       end
 
       def action_run
+        validate_identity(new_resource.user, new_resource.password, new_resource.domain)
```
lamont-granquist: should this go into define_resource_requirements instead?

lib/chef/mixin/user_identity.rb
adamedx: Got it, will switch to `private` @lamont-granquist.

btm:
Replaced by https://github.com/chef/chef/pull/5764

PR #4298, opened 2015-12-13
Combined remote_file and execute alternate user on Windows

adamedx:


lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,62 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
jaym: it's 2016

lib/chef/provider/execute.rb
```diff
@@ -19,15 +19,22 @@
...
       extend Forwardable
 
+      include Chef::Mixin::UserIdentity
+
       provides :execute
 
-      def_delegators :@new_resource, :command, :returns, :environment, :user, :group, :cwd, :umask, :creates
+      def_delegators :@new_resource, :command, :returns, :environment, :user, :domain, :password, :group, :cwd, :umask, :creates
+
+      def initialize(new_resource, run_context)
```
jaym: Does this need to be explicit?

lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,118 @@
...
+
+        def initialize(username, password, domain=nil)
+          if username.nil? || password.nil?
+            raise ArgumentError, 'The logon session must be initialize with non-nil user name and password parameters'
+          end
+
+          @username = username
+          @password = password
+          @domain = domain
+          @token = FFI::Buffer.new(:pointer)
```
jaym: We have a class that wraps tokens. It deals with the lifetime of the underlying handle:
https://github.com/chef/chef/blob/master/lib/chef/win32/security/token.rb

lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,118 @@
...
+        def open
+          if @session_opened
+            raise RuntimeError, 'Attempted to open a logon session that was already open.'
+          end
+
+          username = wstring(@username)
+          password = wstring(@password)
+          domain = wstring(@domain)
+
+          status = Chef::ReservedNames::Win32::API::Security.LogonUserW(username, domain, password, Chef::ReservedNames::Win32::API::Security::LOGON32_LOGON_NETWORK, Chef::ReservedNames::Win32::API::Security::LOGON32_PROVIDER_DEFAULT, @token)
```
jaym: Instead of dealing with FFI, is it possible to use https://github.com/chef/chef/blob/b169e169f9af598ae9d4af5bfc5289b1255d93e7/lib/chef/win32/security.rb#L647 instead?

lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,118 @@
...
+
+          if ! @session_opened
+            raise RuntimeError, 'Attempted to set the user context before opening a session.'
+          end
+
+          if @impersonating
+            raise RuntimeError, 'Attempt to set the user context when the user context is already set.'
+          end
+
+          status = Chef::ReservedNames::Win32::API::Security.ImpersonateLoggedOnUser(@token.read_ulong)
```
jaym: It might be better to add this to the token class if it's possible to use that class in this case

btm:
The execute portions of this were merged in https://github.com/chef/chef/pull/5764. I'm closing this as we just need to work on the remote_file portions, which are represented in https://github.com/chef/chef/pull/4299 still.

PR #4299, opened 2015-12-13
Windows alternate user source access for remote_file

adamedx:
### Windows `remote_file` resource with alternate credentials

The `remote_file` resource now supports the use of credentials on Windows when accessing
a remote UNC path on Windows such as `\\myserver\myshare\mydirectory\myfile.txt`. This
allows access to the file at that path location  even if the Chef client process identity does
not have permission to access the file. The new properties `remote_user`,
`remote_domain`, and `remote_user_password` may be used to specify credentials
with access to the remote file so that it may be read.
#### Properties

The following properties are new for the `remote_file` resource:
-   `remote_user`</br>
  **Ruby types:** String</br>
  _Windows only:_ The user name of a user with access to the remote file specified
  by the `source` property. Default value: `nil`. The user name may optionally be specifed
  with a domain, i.e. `domain\user` or `user@my.dns.domain.com` via Universal Principal Name (UPN)
  format. It can also be specified without a domain simply as `user` if the domain is
  instead specified using the `remote_user_domain` attribute. Note that this property is ignored
  if `source` is not a UNC path. If this property is specified, the `remote_user_password`
  property **must** be specified.
-   `remote_user_password`</br>
  **Ruby types** String</br>
  _Windows only:_ The password of the user specified by the `remote_user` property.
  Default value: `nil`. This property is mandatory if `remote_user` is specified and may only
  be specified if `remote_user` is specified. The `sensitive` property for this resource will
  automatically be set to `true` if `remote_user_password` is specified.
-   `remote_user_domain`</br>
  **Ruby types** String</br>
  _Windows only:_ The domain of the user user specified by the `remote_user` property.
  Default value: `nil`. If not specified, the user and password properties specified
  by the `remote_user` and `remote_user_password` properties will be used to authenticate
  that user against the domain in which the system hosting the UNC path specified via `source`
  is joined, or if that system is not joined to a domain it will authenticate the user
  as a local account on that system. An alternative way to specify the domain is to leave
  this property unspecified and specify the domain as part of the `remote_user` property.

### DIFF
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -49,7 +49,7 @@ class Chef
             logon_session.open
             logon_session.set_user_context
           end
-          block.call
+          yield
         ensure
           logon_session.close if logon_session
         end
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -1,0 +1,62 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/util/windows/logon_session" if Chef::Platform.windows?
+require "chef/mixin/user_identity"
+
+class Chef
+  module Mixin
+    module UserContext
+
+      include Chef::Mixin::UserIdentity
+
+      def with_user_context(specified_user, password, specified_domain = nil, &block)
+        if ! Chef::Platform.windows?
+          raise Exceptions::UnsupportedPlatform, "User context impersonation is supported only on the Windows platform"
+        end
+
+        if ! block_given?
+          raise Exceptions::ArgumentError, "You must supply a block to `with_user_context`"
+        end
+
+        validate_identity(specified_user, password, specified_domain)
+
+        identity = qualify_user(specified_user, specified_domain)
+
+        user = identity[:user]
+        domain = identity[:domain]
+
+        login_session = nil
+
+        begin
+          if user
+            logon_session = Chef::Util::Windows::LogonSession.new(user, password, domain)
+            logon_session.open
+            logon_session.set_user_context
+          end
+          block.call
+        ensure
+          logon_session.close if logon_session
+        end
+      end
+
+      protected(:with_user_context)
+
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,62 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
tyler-ball: 2016

lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      include Chef::Mixin::UserIdentity
+
+      def with_user_context(specified_user, password, specified_domain = nil, &block)
+        if ! Chef::Platform.windows?
+          raise Exceptions::UnsupportedPlatform, "User context impersonation is supported only on the Windows platform"
+        end
+
+        if ! block_given?
+          raise Exceptions::ArgumentError, "You must supply a block to `with_user_context`"
```
tyler-ball: I think this should just be `raise ArgumentError` as it is in the other mixin

lib/chef/provider/remote_file/network_file.rb
```diff
@@ -35,13 +42,23 @@ def initialize(source, new_resource, current_resource)
         # Fetches the file on a network share, returning a Tempfile-like File handle
         # windows only
         def fetch
-          tempfile = Chef::FileContentManagement::Tempfile.new(new_resource).tempfile
-          Chef::Log.debug("#{new_resource} staging #{@source} to #{tempfile.path}")
-          FileUtils.cp(@source, tempfile.path)
-          tempfile.close if tempfile
+          validate_identity(new_resource.remote_user, new_resource.remote_user_password, new_resource.remote_user_domain)
```
tyler-ball: `with_user_context` already calls `validate_identity` so this seems unnecessary here.

Also, it seems like the `UserIdentity` and `UserContext` mixins could be merged if the only method providers need to call is `with_user_context`

lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,118 @@
...
+          @username = username
+          @password = password
+          @domain = domain
+          @token = FFI::Buffer.new(:pointer)
+          @session_opened = false
+          @impersonating = false
+        end
+
+        def open
+          if @session_opened
```
tyler-ball: It is more ruby-ish to add `attr_reader :session_opened` and use the method rather than directly accessing `@session_opened`

lib/chef/provider/remote_file/network_file.rb
tyler-ball: Looking at https://github.com/chef/chef/pull/4297 I now see why they are 2 different mixins

adamedx: Yeah, this was originally one giant PR, and I refactored it, so that is indeed the reasoning.

lib/chef/mixin/user_context.rb
adamedx: Good point, I will update.

lib/chef/util/windows/logon_session.rb
adamedx: Thanks @tyler-ball :) -- I will clean that up in my next rebase and remember for the future.

lib/chef/provider/remote_file/network_file.rb
smurawski: I would think `validate_identity` would happen in `define_resource_requirements` along the lines of @lamont-granquist's comment https://github.com/chef/chef/pull/4297/files#r64492533

adamedx:
Hey, just curious why this isn't merged -- it's still open. :)

btm:
This is replaced by #5832, so I'll close this and respond to your message there.

PR #4300, opened 2015-12-13
Always allow nil set; only warn if not using "is"

jkeiser:
- Aways allow nil to be set.
- If "is" is not specified, treat setting to nil as a get and warn if it would change the value.
- If nil fails validation on set, treat it as a reset.


jkeiser:
Closing this: the right behavior is to always accept `nil`.

coderanger:
@jkeiser Why is that the right behavior?

coderanger:
Ahh I see there is another ticket about this now.

PR #4303, opened 2015-12-14
Don't crash if the node has no tags set

thommay:
cc @chef/client-core

### DIFF
--- a/lib/chef/knife/core/node_presenter.rb
+++ b/lib/chef/knife/core/node_presenter.rb
@@ -125,8 +125,12 @@ ROLES
             summarized << <<-SUMMARY
 #{key('Recipes:')}     #{Array(node[:recipes]).join(', ')}
 #{key('Platform:')}    #{node[:platform]} #{node[:platform_version]}
+SUMMARY
+            unless node.tags.nil?
+              summarized << <<-SUMMARY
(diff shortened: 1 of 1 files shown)


lamont-granquist:
i thought we fixed this so that tags were always an array?

thommay:
There are nodes in one of my hosted orgs that cause us to crash. So we should probably nail down where this isn't happening, but do we want to take this as well?

danielsdeleo:
:\ https://github.com/chef/chef/blob/b60ad97446c064c71e5e1b03b94344eb0b5c136d/lib/chef/node.rb#L373-L377

lamont-granquist:
am i missing the edge condition there?  although that could have been `normal_unless[:tags] = []` to use the correct API that was written and tested for that case...

coderanger:
`attribute?` is (eventually) an alias for `has_key?` so if you had `{"tags": null}` in your node data it would fail.

coderanger:
Or `{"tags": {}}`, but that seems less likely to have happened.

lamont-granquist:
okay and previously that was perfectly legal and likely common...

lamont-granquist:
^ fixed that issue in Chef::Node and an identical issue in set_unless in #4309

lamont-granquist:
We should probably close this one now?

PR #4307, opened 2015-12-14
Extend service resource to support masking

davide125:
On systemd machines, services can be stopped, disabled or masked (see http://0pointer.de/blog/projects/three-levels-of-off.html and http://www.freedesktop.org/software/systemd/man/systemctl.html#is-enabled%20NAME... for details). 

This patch adds support for managing the masking state of a service, e.g. doing
```
service 'foo' do
   action :mask 
end
```
will make the 'foo' service masked. The implementation is based off the existing logic for enable/disable.


coderanger:
This needs default implementation on the service base class that raises NotImplementedError. Also I don't think we can do anything about it but state_attrs are literally the worst (´༎ຶД༎ຶ`)

davide125:
@coderanger I've used UnsupportedAction instead for consistency with the rest of the code.

coderanger:
:+1: on that.

What other service providers are going to implement this? If "mask" is a systemd-specific action I'm not super gung-ho about adding it here vs. the systemd cookbook.

davide125:
@coderanger I think on regular sysv init the equivalent action would be doing chmod -x on the init script. I'm not terribly familiar with other init implementations but I think the concept (force disabling a service so that it cannot be started unless manually reactivated) is a reasonably common one. @jaymzh  was also suggesting offline that if we want to make this systemd-specific another way to implement it could be with a subclass (akin to how package / yum_package is done).

coderanger:
There is precedent with `:purge` on the package resource but I think that was a mistake and not something the emulate :-/

jaymzh:
@coderanger  should we do what we did with like `allow_downgrade` on `yum_package`? Make another resource that resolves to providing `package` on CentOS7?

coderanger:
Yeah, that's probably a better model since this isn't going to be a generic thing for all service providers.

lib/chef/resource/systemd_service.rb
```diff
@@ -0,0 +1,49 @@
...
+
+require 'chef/resource/service'
+require 'chef/provider/service/systemd'
+
+class Chef
+  class Resource
+    class SystemdService < Chef::Resource::Service
+      provides :service, os: "linux" do |node|
+        Chef::Platform::ServiceHelpers.service_resource_providers.include?(:systemd)
+      end
```
lamont-granquist: This is probably not the right approach to take for the service resource.  We lazy the evaluation of what kind of provider to invoke to converge time so that we can inspect the service and determine the correct provider.  With this change you may wind up with systemd resources wrapping non-systemd providers.

Another way of looking at this is that instead of doing provider resolver stuff with mutiple object classes for the service provider, what we maybe should have done is collapse the service provider down to a single resource which at converge time would choose the correct implementation to run based on inspecting the system.  That would keep the model 1:1 but delay the decision of the implementation to converge time.

In either case you lose the ability to do systemd-specific validation on the resource, early.  But you can't ever really do that because you don't know if you've really got a systemd service or not until converge time, long after the resource has been constructed.

So this is trying to do the right thing to get validation correct in the resource -- and normally i would approve of moving to a more 1:1 mapping between resource and provider -- but this is the one case that breaks that model.

ranjib:
i like the patch in its current form. is there any easy way to improve the lazy evaluation so that we can suppress it for cases like this.

lamont-granquist:
You have to pick a resource class at compile time to create the object to put it on the resource collection, and the validations happen at that time.

If we eliminated compile/converge mode entirely then this patch as its written would be more acceptable.

In the absence of that what we do is move validation logic into the provider and raise in define_resource_requirements after we know which kind of provider it really is.

I don't like the way it gets done, but this solution isn't any better and will create an even more complicated mess with different edge conditions.

Lets stick with how we already solve this problem, which is slightly bad, but at least has understood edge conditions.

lamont-granquist:
So, :-1:

And to be clear, this is a *very* context dependent argument that I'm making.  Its just the service provider that is special like this, because of the lazy compile-time selection of the correct provider to use.  For nearly every other case, I think the package provider model is vastly superior and gets the validation right (for chef >= 12).

coderanger:
@lamont-granquist Is that :-1: on the specifics or the style being used? I can't tell if you are advocating going back to the "add them as actions to `service`" or just a different way to implement this mode.

lamont-granquist:
I don't like this code:

```
    class SystemdService < Chef::Resource::Service
      provides :service, os: "linux" do |node|
```

service needs to resolve to Chef::Resource::Service for now.

lib/chef/provider/service/systemd.rb
```diff
@@ -67,6 +76,32 @@ def define_resource_requirements
...
+    end
+    load_new_resource_state
+    @new_resource.masked(true)
+  end
+
+  def action_unmask
+    if @current_resource.masked
+      converge_by("unmask service #{@new_resource}") do
+        unmask_service
+        Chef::Log.info("#{@new_resource} masked")
```
andrewhsu: Should probably log to info "unmasked" instead of "masked".

lib/chef/resource/systemd_service.rb
jaymzh: I see your objection, but I see no better way to do this. It's the same as `package 'foo' do allow_downgrade true end` which only works on yum but you can now do via `package`.

How do you actually suggest doing this?

jaymzh: Actually - don't we already do this?

from `service/debian.rb`

```
        provides :service, platform_family: 'debian' do |node|
          Chef::Platform::ServiceHelpers.service_resource_providers.include?(:debian)
        end
```

This seems to be following the existing pattern.

FTR, this is an actively blocker for us. We're monkeypatching this PR in now, but every day we go on adding uses of this is massively more painful if the upstream solution is different, so the sooner we can get an idea of what you think the proper solution is the better.

Also, i think other people are probably just manually making symlinks to /dev/null in /etc/ (which is what this does internally on `:mask`) which is... terrible and we should try to give them something more sane sooner.

lamont-granquist: That is from `provider/service/debian.rb` not from `resource/debian_service.rb`.  I don't have an issue with that on the provider.

I'd prefer to keep the service resource a single class and eat the `allow_downgrade` problem for now.

jaymzh: OK, so what's the right way to do that?

`yum_package` **resource** also responds to `package`:

```
    class YumPackage < Chef::Resource::Package

      provides :yum_package
      provides :package, os: "linux", platform_family: [ "rhel", "fedora" ]
```

so that you can allow_downgrade. You're saying you don't want stuff but you're not telling us what you DO want.

`service` as a resource needs to work on systemd systems and have the right actions available without people having to specify `systemd_service`. It's unclear what mechanism you want Davide to use to make that happen.

lamont-granquist: Just add the actions to the Chef::Resource::Service and ignore the problem that the other providers that aren't systemd services don't implement those actions and with NoMethodError instead of failing validation...

davide125: @lamont-granquist I think that's what I did originally in https://github.com/davide125/chef/commit/f433e35cb072ff99cb8c2673c740c13ade37b021 and later factored out per @coderanger advice. Is this what you mean here?

coderanger: Yeah, the original hope was to not add this to the overall `service` resource API, since it isn't an operation that really exists in any other system (arguable for sysvinit, but thats about it).

lamont-granquist: Yeah, but then you wind up hardcoding the `system` resource to the systemd service resource on the DSL for any system which has systemd and you lose the ability to lazy evaluate to a different provider later -- or worse you might wind up picking a systemd resource and pairing it up with a non-systemd resource at some later point.  Its not a big deal now because currently systemd assimilates distros like the borg, but i wouldn't necessarily assume that will always be the case forever.  And I'd rather put up with the service resource sprouting inaccurate validations than remove that flexibility.

davide125:
@lamont-granquist: ok, PR updated to not do the resource subclassing, lemme know if this looks better. I also fixed the spec and the tests pass fine for me now.

jaymzh:
Heya - @lamont-granquist

lamont-granquist:
:+1:

@thommay?

thommay:
looks good! :+1:

jaymzh:
:+1: 
Merging under my CCLA.

chef-supermarket:

* @davide125

PR #4308, opened 2015-12-14
Only warn about potentially duplicate properties during the resource initializer.

jkeiser:
There are a number of cases where people will construct a resource, initialize it, and then use properties from the resource to calculate other things during compilation. We want to not warn on these; we are really only trying to get rid of ambiguities during the `resource :x do ... end` block.


btm:
:+1:

PR #4309, opened 2015-12-15
tags always an array; fix set_unless

lamont-granquist:
previously if the node had a { normal: { tags: nil } } that we read
since the key existed (but was nil) we would not initialize with an
array.  replacing the code with a call to node.set_unless revealed that
set_unless was similarly buggy.  fixed both issues by fixing set_unless.

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -1,4 +1,3 @@
-#
 # Author:: Adam Jacob (<adam@opscode.com>)
 # Author:: Christopher Brown (<cb@opscode.com>)
 # Author:: Christopher Walters (<cw@opscode.com>)
@@ -372,7 +371,7 @@ class Chef
 
     # Lazy initializer for tags attribute
     def tags
-      normal_unless[:tags] = []
+      normal[:tags] = Array(normal[:tags])
       normal[:tags]
     end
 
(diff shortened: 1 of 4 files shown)


lamont-granquist:
the other side of #4303

coderanger:
This would still fail for `{"tags": "foo"}` or `{"tags": {}}`. I think we should probably return `Array(normal[:tags])` instead to catch both of these.

lamont-granquist:
Hmmm, I don't think anyone does that or else we would have heard about it...

OTOH, likely nobody does that so its probably safe to change that behavior...

coderanger:
If we want to define that `#tags` returns an array, we should make sure that's all it can return :)

coderanger:
The only way I could imagine those happening is someone mis-editing the JSON via manual `knife node edit`.

danielsdeleo:
sounds good :+1:

coderanger:
:+1: I think. This is a bit weird because now `Node#tags` has a side effect rather than being an accessor, but I think that should be okay.

coderanger:
To clarify for the future when someone complains, I think it is fair to say that we (Chef core) "own" `node.normal['tags']` so doing a transparent format upgrade is within scope for Chef's code.

lamont-granquist:
Yeah, I thought about the "side effect" for a bit and pondered how much of a no-op Array.new was or not, and then just went for the trivial looking code.  I also like that it makes a strong statement - if you use the accessor you always get an Array and it will also always exactly match the underlying ivar (even if we have to fix it).

I also can't really imagine anyone complaining, but the only way to wind up complaining about this is to not use the public API...

lamont-granquist:
travis failures are GCE's lack of IPv6 trolling us...

coderanger:
:+1:

thommay:
Thanks Lamont. :+1:

PR #4311, opened 2015-12-15
Don't run IPv6 tests on GCE

jkeiser:


danielsdeleo:
As you said in chat, it's worth considering marking these as pending instead, so we'll be alerted when GCE starts supporting ipv6. But I'm okay with this, too.

PR #4315, opened 2015-12-15
Add extra tests around whether to skip with multiple guards

bbbco:
I was trying to answer a question I had today about whether a resource would be executed when multiple guards are assigned. Unfortunately, I could not find any info on this subject within the web docs. So I checked the specs, and found the resource spec with some relevant tests. 

I noticed that there were a few extra tests that could be included to fully answer my question. This pull request adds them.

### DIFF
--- a/spec/unit/resource_spec.rb
+++ b/spec/unit/resource_spec.rb
@@ -743,6 +743,12 @@ describe Chef::Resource do
       expect(resource.should_skip?(:purr)).to be_truthy
     end
 
+    it "should return false when only_if is met and also not_if is not met" do
+      resource.only_if { true }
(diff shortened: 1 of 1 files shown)


bbbco:
So I'm pretty sure the AppVeyor build failing is not because of my commit. It looks like it may be failing intermittently on other pull requests. Do we have anyone looking into what my be going on?

lamont-granquist:
:+1: 

nice, finally a PR that we don't have to ask for more specs around the use case...

i'm pretty confident appveyor failures aren't your fault...

bbbco:
:+1:

thommay:
:+1:

PR #4316, opened 2015-12-15
chef-client: add --[no]skip-cookbook-sync option

josb:
Use with caution. Useful for patching a set of cookbooks on a machine
when iterating during development.

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -285,6 +285,11 @@ class Chef::Application::Client < Chef::Application
     :description    => "DANGEROUS: does what it says, only useful with --recipe-url",
     :boolean        => true
 
+  option :skip_cookbook_sync,
+    :long           => "--[no-]skip-cookbook-sync",
+    :description    => "Whether to skip cookbook synchronization",
+    :boolean        => false
+
   IMMEDIATE_RUN_SIGNAL = "1".freeze
 
   attr_reader :chef_client_json
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -287,7 +287,7 @@ class Chef::Application::Client < Chef::Application
 
   option :skip_cookbook_sync,
     :long           => "--[no-]skip-cookbook-sync",
-    :description    => "Whether to skip cookbook synchronization",
+    :description    => "Use cached cookbooks without overwriting local differences from the server",
     :boolean        => false
 
   IMMEDIATE_RUN_SIGNAL = "1".freeze
--- a/distro/common/markdown/man8/chef-client.mkd
+++ b/distro/common/markdown/man8/chef-client.mkd
@@ -40,6 +40,8 @@ __chef-client__ _(options)_
     Set the PID file location, defaults to /tmp/chef-client.pid
   * `--once`:
     Cancel any interval or splay options, run chef once and exit
+  * `--skip-cookbook-sync`:
+    Skip cookbook synchronization
   * `-v`, `--version`:
     Show chef version
   * `-h`, `--help`:
--- a/lib/chef/cookbook/synchronizer.rb
+++ b/lib/chef/cookbook/synchronizer.rb
@@ -263,6 +263,7 @@ class Chef
     end
 
     def cached_copy_up_to_date?(local_path, expected_checksum)
+      return true if Chef::Config[:skip_cookbook_sync]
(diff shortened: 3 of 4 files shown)


coderanger:
:-1: :-1: :-1: :-1: :-1: 

This is not what `chef-client` is for, will break unknowable things.

josb:
Tools, not policies.

How else would you do this with just chef-client installed?

coderanger:
Use `chef-solo` with the cache folder as the cookbook root.

josb:
Thanks for the suggestion, @coderanger !

But doesn't ```chef-solo``` turn off remote/chef-server data bag access? If so, that won't work for me.

In fact, I think I tried this (```chef-client -z```), which prompted me to whip up this patch. I used it yesterday to debug yet another DSL-related Chef problem on a machine.

lamont-granquist:
heh, i've hacked up the cookbook synchronizer before this way to do 'testing in prod' before...  sometimes faster than trying to get a replication case in a virt, and when adding debugging and pinning a single box to a debug version of a cookbook is both fussy and possibly error prone...

thommay:
I think if we take this (and I've definitely done similar for bugs that are hard to reproduce outside of certain environments), we should add some ENORMOUS warnings and log messages. But I'm mostly :+1: for that 1-in-a-100 time this is useful.

smurawski:
:+1: with some additional warning messages when this flag is included as part of the chef run.

lamont-granquist:
So I'm actually :+1: 

It is the wrong thing to do 99 times out of 100.

But that 1 timeout of 100 when its difficult to replicate in TK or preprod and trying to push out a one box test without causing accidental chaos across all of production is annoying, and you really just want to login to this one prod box and test a fix...   then I totally use it.

Definite footgun.

But I believe all of Chef is founded on the Right to Keep and Bear Footguns.

josb:
Thanks for the feedback!

Any suggestions as to the form of the warnings/log messages? I'd be happy to add them.

youngjl1:
this would be great

thommay:
Warning at the start of the run that says "Using cookbooks from cache without syncing" or something.

lamont-granquist:
maybe:

```ruby
if Chef::Config[:skip_cookbook_sync]
  Chef::Log.warn "skipping cookbook synchronization!  DO NOT LEAVE THIS ENABLED IN PRODUCTION!!!"
  return true
end
```

josb:
I updated the pull request per @lamont-granquist 's suggestion. What else is needed before it can be merged? Thanks!

thommay:
@coderanger are you still -1 on this? 

@josb essentially, because there's a -1 on this PR, we need to either get Noah to remove his -1, or have a majority of maintainers +1 it. Obviously we'd prefer the former happens.

lib/chef/application/client.rb
```diff
@@ -285,6 +285,11 @@ class Chef::Application::Client < Chef::Application
     :description    => "DANGEROUS: does what it says, only useful with --recipe-url",
     :boolean        => true
 
+  option :skip_cookbook_sync,
+    :long           => "--[no-]skip-cookbook-sync",
+    :description    => "Whether to skip cookbook synchronization",
```
danielsdeleo: Can we tighten up this wording? Maybe like this?

> Use cached cookbooks without overwriting local differences from the server

coderanger:
@thommay Lamont convinced me to be a -0. Most people that use this are going to be doing the wrong thing and the docs should reflect that, but right to keep and bear footguns :)

lib/chef/application/client.rb
josb: @thom understood, no worries.

@danielsdeleo done!

@coderanger thanks for your support!

thommay:
@josb that wording should be in two places `application/client.rb` and `shell.rb`, you only got it in `shell.rb`. As soon as you add that I'll whack merge. Thanks for your patience!

josb:
@thommay good catch, fixed!

josb:
By the way, the warning message as it is in the current patch is emitted for every cookbook. Is this what we want? Or do we want to only emit it once? The latter seems less noisy.

thommay:
I'm fine with the noise. better to reinforce the potential foot gunnery.

josb:
@thommay fair enough. And thanks for accepting my patch!

PR #4318, opened 2015-12-16
Allow package version to be nil

jaym:
This feels like the wrong fix. If `default_value` is not specified, a property should probably implicitly allow `kind_of` to contain `nil` as it's going to default to nil.

Fixes failing test:

```
  1) Chef::Provider::Package::EasyInstall easy_install_binary_path should set the current resources package name to the new resources package name
     Failure/Error: @provider.load_current_resource

     Chef::Exceptions::DeprecatedFeatureError:
       version nil currently does not overwrite the value of version. This will change in Chef 13, and the value will be set to nil instead. Please change your code to explicitly accept nil using "property :version, [MyType, nil]", or stop setting this value to nil. at C:/projects/chef/spec/unit/provider/package/easy_install_spec.rb:52:in `block (3 levels) in <top (required)>'
     # ./lib/chef/log.rb:61:in `deprecation'
     # ./lib/chef/chef_class.rb:214:in `log_deprecation'
     # ./lib/chef/property.rb:262:in `call'
     # ./lib/chef/property.rb:484:in `version'
     # ./lib/chef/provider/package/easy_install.rb:99:in `load_current_resource'
     # ./spec/unit/provider/package/easy_install_spec.rb:52:in `block (3 levels) in <top (required)>'

Finished in 24 minutes 22 seconds (files took 10.92 seconds to load)
16178 examples, 1 failure, 104 pending
```

### DIFF
--- a/lib/chef/provider/package/easy_install.rb
+++ b/lib/chef/provider/package/easy_install.rb
@@ -93,10 +93,10 @@ class Chef
 
           if package_version == @new_resource.version
             Chef::Log.debug("#{@new_resource} at version #{@new_resource.version}")
-            @current_resource.version(@new_resource.version)
+            @current_resource.version = @new_resource.version
           else
             Chef::Log.debug("#{@new_resource} at version #{package_version}")
-            @current_resource.version(package_version)
+            @current_resource.version = package_version
           end
 
           @current_resource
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@jkeiser has a pull in flight to fix this i think.

jaym:
closing in favor of https://github.com/chef/chef/pull/4320

PR #4321, opened 2015-12-16
fix run_as_user of windows_service

mwrock:
Currently adding a user in the form of `{domain or host}\{user name}` is broken for the `run_as_user` attribute of `windows_service`. It only works with user names prepended with `.\`. This fixes that.

This also now uses the win32 LSA functions and adds the `SeServiceLogonRight` to the user and only if that user does not already have the right. Currently the service clobbers all current users with that right with the `run_as_user`.

### DIFF
--- a/lib/chef/win32/api/security.rb
+++ b/lib/chef/win32/api/security.rb
@@ -207,6 +207,21 @@ class Chef
         LOGON32_PROVIDER_WINNT40 = 2;
         LOGON32_PROVIDER_WINNT50 = 3;
 
+        # LSA access policy
+        POLICY_VIEW_LOCAL_INFORMATION = 0x00000001
+        POLICY_VIEW_AUDIT_INFORMATION = 0x00000002
+        POLICY_GET_PRIVATE_INFORMATION = 0x00000004
+        POLICY_TRUST_ADMIN = 0x00000008
+        POLICY_CREATE_ACCOUNT = 0x00000010
+        POLICY_CREATE_SECRET = 0x00000020
+        POLICY_CREATE_PRIVILEGE = 0x00000040
+        POLICY_SET_DEFAULT_QUOTA_LIMITS = 0x00000080
+        POLICY_SET_AUDIT_REQUIREMENTS = 0x00000100
+        POLICY_AUDIT_LOG_ADMIN = 0x00000200
+        POLICY_SERVER_ADMIN = 0x00000400
+        POLICY_LOOKUP_NAMES = 0x00000800
+        POLICY_NOTIFICATION = 0x00001000
+
         ###############################################
         # Win32 API Bindings
         ###############################################
@@ -381,6 +396,23 @@ class Chef
           end
         end
 
+        # https://msdn.microsoft.com/en-us/library/windows/desktop/ms721829(v=vs.85).aspx
+        class LSA_OBJECT_ATTRIBUTES < FFI::Struct
+          layout :Length, :ULONG,
+                 :RootDirectory, :HANDLE,
+                 :ObjectName, :pointer,
+                 :Attributes, :ULONG,
+                 :SecurityDescriptor, :PVOID,
+                 :SecurityQualityOfService, :PVOID
+        end
+
+        # https://msdn.microsoft.com/en-us/library/windows/desktop/ms721841(v=vs.85).aspx
+        class LSA_UNICODE_STRING < FFI::Struct
+          layout :Length, :USHORT,
+                 :MaximumLength, :USHORT,
+                 :Buffer, :PWSTR
+        end
+
         ffi_lib "advapi32"
 
         safe_attach_function :AccessCheck, [:pointer, :HANDLE, :DWORD, :pointer, :pointer, :pointer, :pointer, :pointer], :BOOL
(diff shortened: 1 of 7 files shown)


ksubrama:
:+1:

spec/unit/provider/service/windows_spec.rb
```diff
@@ -417,37 +426,19 @@
...
+      expect(Chef::ReservedNames::Win32::Security).to receive(:add_account_right).with(username, service_right)
       expect(provider.grant_service_logon(username)).to equal true
     end
 
-    it "raises an exception when the grant command fails" do
-      expect_any_instance_of(Mixlib::ShellOut).to receive(:exitstatus).and_return(1)
-      expect_any_instance_of(Mixlib::ShellOut).to receive(:stdout).and_return(failure_string)
+    it "strips '.\' from user name when sending to win32 api" do
+      expect(Chef::ReservedNames::Win32::Security).to receive(:add_account_right).with(username, service_right)
+      expect(provider.grant_service_logon(".\\#{username}")).to equal true
```
btm: what is/are the valid syntax(s) for username now?

```
  # the security policy file only seems to accept \\username, so fix .\username or .\\username.
  # TODO: this probably has to be fixed to handle various valid Windows names correctly.
  def canonicalize_username(username)
    username.sub(/^\.?\\+/, '')
  end
```

mwrock: The win32 APIs used don't understand `.\username`. However they work fine with all of these:
* `username`
* `domain\username`
* `computer\\username`

We still allow users to use the `.\username` format and strip off the `.\` here. That makes sense since its a generally accepted format elsewhere.

btm: Could you update that comment since we're using the API instead of calling out to secedit with a file now?

btm:
:+1: /cc @chef/client-windows review please

spec/unit/provider/service/windows_spec.rb
mwrock: ah right. yeah

lib/chef/win32/api/security.rb
```diff
@@ -415,6 +447,12 @@ def privilege(index)
         safe_attach_function :LookupPrivilegeNameW, [ :LPCWSTR, :PLUID, :LPWSTR, :LPDWORD ], :BOOL
         safe_attach_function :LookupPrivilegeDisplayNameW, [ :LPCWSTR, :LPCWSTR, :LPWSTR, :LPDWORD, :LPDWORD ], :BOOL
         safe_attach_function :LookupPrivilegeValueW, [ :LPCWSTR, :LPCWSTR, :PLUID ], :BOOL
+        safe_attach_function :LsaAddAccountRights, [ :pointer, :pointer, :pointer, :ULONG ], :NTSTATUS
```
jaym: instead of `:pointer`, can we create aliases for the types in the documentation. https://github.com/chef/chef/blob/master/lib/chef/win32/api.rb#L49-L237

lib/chef/win32/security.rb
```diff
@@ -104,6 +104,22 @@ def self.add_access_denied_ace_ex(acl, sid, access_mask, flags = 0, revision = A
         end
       end
 
+      def self.add_account_right(name, privilege)
+        privilege_pointer = FFI::MemoryPointer.new LSA_UNICODE_STRING, 1
+        privilege_lsa_string = LSA_UNICODE_STRING.new(privilege_pointer)
+        privilege_lsa_string[:Buffer] = FFI::MemoryPointer.from_string(privilege.to_wstring)
+        privilege_lsa_string[:Length] = privilege.length * 2
```
jaym: would `privilege.size` give the right value here?

jaym:
:+1:

smurawski:
:+1:

Thanks for fixing the rights assignment mangling.

lib/chef/win32/security.rb
mwrock: no. I fiddled with this quite alot. Even `privilege.to_wstring.size` is too small

jaym: sorry, `privilege_pointer`

mwrock: yeah. that doesn't fit

mwrock:
pushed up some changes from review. Will merge after it goes green.

ShawInnes:
I'm trying to work out if this has been released in a client release yet.  Is there an easy way for me to find out?  

I'm currently running chef client 12.6.0 and having issues with the following resource use.  The error references a policyfile 'Logon-as-service grant failed with policy file' so I'm assuming this commit isn't in my current client.  

```
service 'TCBuildAgent' do
  run_as_user 'MYDOMAIN\svcTeamCity'
  run_as_password "..."
  action :start
end
```

mwrock:
It will be available in chef 12.7. see [this pre-release announcment](https://discourse.chef.io/t/pre-release-announcement-chef-client-12-7/7747) with details on how to get it now.

Techcadia:
If your using a local user account make sure that the run_as_user has the 'HOSTNAME\Username' not just 'Username'

It was throwing a
`the account name is invalid or does not exist or the password is invalid for the account specified` Error which is exactly right.

Took me a good hour to figure that out.

Issue #4294, opened 2015-12-11
group#members naively coerced into an Array


nvwls:
Given the following test:

```
stooges = %w(moe larry curly)

stooges.each do |name|
  user name
end

group 'actors' do
  members stooges
end

group 'stooges' do
  members 'moe,larry,curly'
end
```

The groups are created properly.  However, subsequent runs will update 'stooges'

```
# getent group stooges
stooges:x:501:larry,moe,curly

# chef-apply /vagrant/bug
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * user[larry] action create (up to date)
  * user[moe] action create (up to date)
  * user[curly] action create (up to date)
  * group[actors] action create (up to date)
  * group[stooges] action create
    - alter group stooges
    - replace group members with new list of members
```

The notion that members is coerced into an Array is nice but naive. In other places an Array is required, a string is passed in, the type check would fail thus giving the user feedback.

Instead of
```
converted_members = arg.is_a?(String) ? [].push(arg) : arg
```
the following would be better:
```
converted_members = arg.is_a?(String) ? arg.split(',') : arg
```
Or to be even more thorough:
```
converted_members = arg.is_a?(String) ? arg.split(/\s*,\s*/).reject { |u| u.empty? } : arg
```
Let me know which (simple or complex) and I can create a PR.

lamont-granquist:
middle one i think.  last one i think is probably trying to do a bit too much with bad input.

PR #4322, opened 2015-12-17
[chef/4294] group: split #members/#excluded_members on commas

nvwls:
Given the following test:
```
stooges = %w(moe larry curly)

stooges.each do |name|
  user name
end

group 'actors' do
  members stooges
end

group 'stooges' do
  members 'moe,larry,curly'
end
```
The groups are created properly. However, subsequent runs will update 'stooges'
```
# getent group stooges
stooges:x:501:larry,moe,curly

# chef-apply /vagrant/bug
  * user[larry] action create (up to date)
  * user[moe] action create (up to date)
  * user[curly] action create (up to date)
  * group[actors] action create (up to date)
  * group[stooges] action create
    - alter group stooges
    - replace group members with new list of members
```

By splitting on commas, it will convert into an array and not update the resource each run.

### DIFF
--- a/lib/chef/resource/group.rb
+++ b/lib/chef/resource/group.rb
@@ -55,7 +55,7 @@ class Chef
       end
 
       def members(arg=nil)
-        converted_members = arg.is_a?(String) ? [].push(arg) : arg
+        converted_members = arg.is_a?(String) ? arg.split(',') : arg
         set_or_return(
           :members,
           converted_members,
@@ -66,7 +66,7 @@ class Chef
       alias_method :users, :members
 
       def excluded_members(arg=nil)
-        converted_members = arg.is_a?(String) ? [].push(arg) : arg
+        converted_members = arg.is_a?(String) ? arg.split(',') : arg
         set_or_return(
           :excluded_members,
           converted_members,


coderanger:
Is there some backstory on this? Nothing else in Chef behaves like this.

stevendanna:
Given that we can't remove the array coercion altogether without breaking backwards compat, :+1: on this.  However, I do think we should consider being more strict about user-inputs in the future.  

> Is there some backstory on this?

The [related bug report](https://github.com/chef/chef/issues/4294) has some more detail it seems.  A key point for me is that the comma-separated list doesn't cause a problem on create but then causes the resource to unnecessarily update on subsequent runs. 

@nvwls It would be good if your commit message had some of the context from the bug report about what this change accomplishes.

>  Nothing else in Chef behaves like this.

Many resources have ad-hoc coercion rules.  The mount resources `options` property splits on both `,` and spaces.  The most tragic I know about is the remote_file's source property:

```
      # source can take any of the following as arguments
      # - A single string argument
      # - Multiple string arguments
      # - An array or strings
      # - A delayed evaluator that evaluates to a string
      #   or array of strings
      # All strings must be parsable as URIs.
      # source returns an array of strings.
      def source(*args)
        arg = parse_source_args(args)
        ret = set_or_return(:source,
                            arg,
                            { :callbacks => {
                                :validate_source => method(:validate_source)
                              }})
        if ret.is_a? String
          Array(ret)
        else
          ret
        end
      end
```

lamont-granquist:
:+1:

nvwls:
I'm not sure how to resolve the test failure:

```
ERROR: You authenticated successfully to https://api.chef.io/organizations/chef_proxy_tests as client-creator but you are not authorized for this action
```

This seems unrelated to my change.

lamont-granquist:
Not sure about that last commit, could you just rebase on master to eliminate the test failure and then force push instead of merging master back into here?

lamont-granquist:
Spec would be nice

thommay:
:+1: needs spec

lamont-granquist:
@nvwls if you wanna write a spec that'd be cool, otherwise we should write one

PR #4323, opened 2015-12-17
Included the message part to the body of :delete request

gorazdzagar:
While RFC does not explicitly suggest that, there are services like CloudFlare that utilize the data part of DELETE request as well. It actually does not execute the requests with out it making Chef http_request unsuitable for executing actions like purge requests. I added the message part to the body of :delete.

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -131,8 +131,8 @@ class Chef
     #
     # === Parameters
     # path:: path part of the request URL
-    def delete(path, headers={})
-      request(:DELETE, path, headers)
+    def delete(path, json, headers={})
+      request(:DELETE, path, headers, json)
     end
 
     # Makes an HTTP request to +path+ with the given +method+, +headers+, and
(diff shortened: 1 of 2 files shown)


lib/chef/http.rb
```diff
@@ -131,8 +131,8 @@ def post(path, json, headers={})
     #
     # === Parameters
     # path:: path part of the request URL
-    def delete(path, headers={})
-      request(:DELETE, path, headers)
+    def delete(path, json, headers={})
```
lamont-granquist: this changes the API and breaks the world with dozens of

```
ArgumentError:
        wrong number of arguments (1 for 2..3)
```

errors in travis.  either this needs to be an optional argument or a new 'delete_with_body' method.

lamont-granquist:
absolutely can't merge this without fixes

thommay:
@gorazdzagar thanks for your pull request. We'd be interested in taking this but it needs a fair bit of work around API compatibility and tests. Please feel free to reopen if you'd like.

PR #4324, opened 2015-12-17
nil value handling

lamont-granquist:


lamont-granquist:
@jkeiser i think this gets the code right.

if 'nil' is not a valid value then in Chef-13 we will always throw an exception.

if 'nil' is a valid value and we are passing nil then in Chef-13 will overwrite instead of no-op.

i fear you shipped 12.5.1 with code that this backs out.

if we don't back that out, though, we can't use 'nil' in validations inside core chef without changing behavior (so we can't use it until chef-13).

that means users will always get the deprecation warning because without nil being valid it'll always break in chef-13.

if we accept this patch then we can sprinkle nil throughout our core resources, and users get the less painful deprecation warning, and chef-13 migration is easier.

jkeiser:
We decided on this in our talk:

- `nil` will become a valid value, always and forever, and will never be coerced or validated. Thus we will *always* warn on `x(nil)` if the value would change, and `explicitly_accepts_nil?` goes away (yay!)
- Defaults on resources will validate in 13, and we will warn if you define a property with an invalid default in Chef 12.

coderanger:
So I object that `nil` should always be a valid value. I have plenty of things like `property(:foo, kind_of: String, default: 'foo')` where I don't expect to ever see a nil in there.

lamont-granquist:
We are actually in violent agreement.

`property :foo, [ true, false ], default: true` does not accept nil at all

`property :foo, [ true, false ]` does implicitly accept nil

`property :foo, [ true, false ], default: nil` will fail validation and blow up on declaration

`property :foo, [ true, false, nil ], default: nil` is just writing out the implicit case explicitly

`property :foo, [ true, false ], default: :whatever` will fail validation on declaration

`property:foo, [ true, false, :whatever ], default: :whatever` will not accept nil

The nil value is the only one that has special treatment, and if and only if it is the implicit default (and it is the only implicit default value).

I went down the road of wanting defaults to always get merged into the validations, but that hit the problem of lazy defaults, which implied that you /could/ have values that would get set which would fail the validation tests.  That meant that properties might fail to be copied.  So, we backed that out and required that defaults need to be explicitly added to validations -- but added just enough sugar around nil so that the extremely common case of an implicit nil default value was easy to type.

lamont-granquist:
closing in favor of #4331

PR #4328, opened 2015-12-18
DISCUSSION: nil values for properties

jkeiser:
This is just a place for discussion and comments. Do not merge.

### DIFF
--- a/nil-discussion.md
+++ b/nil-discussion.md
@@ -1,0 +1,174 @@
+# nil values and validation
+
+`nil` setters and property defaults are a big snarl with our system because of
+this conflict:
+
+1. People who don't specify a default value for their property expect the
+   default to be `nil`.
+2. Some people want `nil` to be an invalid value for their property so that they
+   can simplify their code by not having to handle it.
+3. Copying a property from one resource to another is a common thing, and the
+   combination of the above rules currently causes `property :x, String` to
+   have an invalid default value, meaning it can't be copied. This is bad.
+
+In coming up with solutions, there are a number of things around default values
+and validation that have to be considered. I've written down these things (the
+things I think we want to be true) here, and enumerated all of the solutions
+I've heard so far, along with which rules they violate and surprises they
+create.
+
+## Rules
+
+Things we want to be true:
+
+1. A property with no default value successfully returns `nil` as its default,
+   regardless of validation.
+
+   This is the crux of the issue, the exception that means we have to do
+   *something* special.
+
+   - `property :x, String` has `nil` as its default value.
+   - `property :x` has `nil` as its default value.
+
+2. The default value of a property must be settable (the copy rule).
+
+   Corollary: Setters and default values are validated, so nothing can get *into*
+   the property unless it is valid.
+
+   - `x("10")` for `property :x, Integer` is INVALID.
+   - `property :x, Integer, default: "10"` is INVALID.
+   - `property :x, Integer, default: lazy { "10" }` is INVALID.
+
+3. Properties must never return invalid values.
+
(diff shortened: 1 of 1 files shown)


nil-discussion.md
```diff
@@ -0,0 +1,180 @@
...
+   - `property :x` has `nil` as its default value.
+
+2. The default value of a property must be settable (the copy rule).
+
+   Corollary: Setters and default values are validated, so nothing can get *into*
+   the property unless it is valid.
+
+   - `x("10")` for `property :x, Integer` is INVALID.
+   - `property :x, Integer, default: "10"` is INVALID.
+   - `property :x, Integer, default: lazy { "10" }` is INVALID.
```
coderanger: I would accept a system where this isn't checked but I agree it should be invalid at least in the abstract.

nil-discussion.md
```diff
@@ -0,0 +1,180 @@
...
+```
+property :x, String
+property :x, String, default: nil # ValidationError when declaring this
+```
+
+This is not intuitive. It's not terrible in my opinion, as the failure of
+intuition would be revealed immediately on declaration with a clear error
+message.
+
+### Properties with no default or `default: nil` are valid
```
coderanger: I agree with your ranking, this and the above seem like the best, and I tend toward this one fur the lower surprise and improved compat with existing code.

coderanger:
:+1: We should probably chuck this in an informational RFC if/when we settle on a solution that that future-us has an easy reference for why there is so much legacy plumbing. Or we could just YOLO and release Chef 13 next month :trollface:

jkeiser:
Yeah, I honestly don't think anyone outside of you, Lamont and I really care about nil handling :) An informational RFC would be good for sure!

I lean slightly towards the other solution, myself, just for conceptual parsimony--it burns me that `default: lazy { thing_that_can_return_nil }` would validate `nil` but `default: nil` would not--but I'm sure not going to throw an exception either way.

stevendanna:
>  Yeah, I honestly don't think anyone outside of you, Lamont and I really care about nil handling :) An informational RFC would be good for sure!

I don't think that is true (I feel like these issues have been the cause of many subtle bugs I've had to fix in cookbooks over time); however, I definitely think it is true that you three are likely a submajority of the people who know the fallout of these various paths with relation to Chef internals :)

coderanger:
I feel like validating lazy defaults can have a little more conceptual wiggle room as it's already an middling-to-advanced feature and we don't have to worry as much about footguns compared to the simple cases.

nil-discussion.md
```diff
@@ -0,0 +1,180 @@
...
+
+This passes all of the rules, and is simple. If the default is not specified,
+then we are giving you a default, and we should take our own steps to make sure
+it is a valid one.
+
+The only issue is that these two would NOT be equivalent:
+
+```
+property :x, String
+property :x, String, default: nil # ValidationError when declaring this
```
stevendanna: This being a ValidationError is OK with me in theory but is exactly the kind of minor annoyance that leads to frustration I think.  "If the 'default default' is nil, why can't I set it to nil?", my imagined user says.

nil-discussion.md
stevendanna: :+1:

jkeiser:
I have fully implemented this, using the scheme where `default: nil` validates `nil` (but that will be easy to change) here: https://github.com/chef/chef/pull/4331

Warning when default values are invalid actually found a couple of bugs in the current code that could have had real implications (they are bugs I added post-12.6 when I converted packages over, but bugs nonetheless).  Happy with that one!

lamont-granquist:
closing since i think any discussion needs to happen in #4331

PR #4330, opened 2015-12-18
Disabled mount tests on AIX

scotthain:
One cannot create a ramdisk in a WPAR, since it shares filesystem with the LPAR host. Since we use ramdisk for the test, mount tests only work in LPARs. In the future we should find a way to test this code.

@btm @jaym

### DIFF
--- a/spec/functional/resource/aix_service_spec.rb
+++ b/spec/functional/resource/aix_service_spec.rb
@@ -108,8 +108,8 @@ describe Chef::Resource::Service, :requires_root, :aix_only do
     it_behaves_like "src service"
   end
 
-
-  describe "When service is a group" do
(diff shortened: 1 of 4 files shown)


btm:
oh man, those WPARs these days.

:+1:

lamont-granquist:
sigh :+1: 

if there's any way of doing loopback mounts on AIX to files, that could be used instead of a ramdisk.  

i think the only reason for a ramdisk in these is that its a temporary thing that you can mount and unmount...

scotthain:
@lamont-granquist Yeah I looked at looking a loopback mount but that's going to take some serious time and testing just to get the test in the right place. I don't want to hold up the release because of that :(

jaym:
:+1:

PR #4331, opened 2015-12-18
Fix nil with properties

jkeiser:
1. Warn when default values are invalid.
2. Never validate nil (on set or get) if there is no default.
3. Emit "will be invalid in Chef 13" warning when setting an invalid nil value.

As per https://github.com/chef/chef/pull/4328


jkeiser:
The one thing this goes backwards on is, there's no way to say "no seriously, I'm cool with `nil` being a valid value on my resource, stop warning me." @lamont-granquist had some ideas for feature flagging that could solve that rather nicely--but I consider it out of scope for this PR.

lamont-granquist:
I ran this through manually and it behaves the way that I want in all the edge cases in #4324 

:+1:

PR #4333, opened 2015-12-19
no longer wait on node search to refresh vault but pass created ApiCient instead

mwrock:
Now that chef_vault can take a ApiClient, we can pass the one generated during bootstrap along instead of having the vault wait for a client to come back from search.

### DIFF
--- a/lib/chef/api_client/registration.rb
+++ b/lib/chef/api_client/registration.rb
@@ -53,8 +53,9 @@ class Chef
       def run
         assert_destination_writable!
         retries = Config[:client_registration_retries] || 5
+        client = nil
         begin
-          create_or_update
+          client = api_client(create_or_update)
         rescue Net::HTTPFatalError => e
           # HTTPFatalError implies 5xx.
           raise if retries <= 0
@@ -64,6 +65,7 @@ class Chef
           retry
         end
         write_key
+        client
       end
 
       def assert_destination_writable!
@@ -104,6 +106,28 @@ class Chef
           @server_generated_private_key = response["private_key"]
         end
         response
+      end
+
+      def api_client(response)
+        return response if response.is_a?(Chef::ApiClient)
+
+        client = Chef::ApiClient.new
+        client.name(name)
+        client.public_key(api_client_key(response, "public_key"))
+        client.private_key(api_client_key(response, "private_key"))
+        client
+      end
+
+      def api_client_key(response, key_name)
+        if response[key_name]
+          if response[key_name].respond_to?(:to_pem)
+            response[key_name].to_pem
+          else
+            response[key_name]
+          end
+        elsif response["chef_key"]
+          response["chef_key"][key_name]
+        end
       end
 
       def put_data
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -377,7 +377,7 @@ class Chef
 
           client_builder.run
 
-          chef_vault_handler.run(node_name: config[:chef_node_name])
+          chef_vault_handler.run(client_builder.client)
 
           bootstrap_context.client_pem = client_builder.client_path
         else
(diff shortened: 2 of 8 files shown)


mwrock:
Note:  This still needs to make sure that it is being used with chef-vault 2.6 or greater.

lib/chef/api_client/registration.rb
```diff
@@ -102,10 +104,31 @@ def update
           @server_generated_private_key = response.private_key
         else # Chef 10
           @server_generated_private_key = response["private_key"]
+          response = api_client(response)
```
jaym: should this be outside the if statement? Otherwise, update's return type is not consistent?

mwrock: no because post 10 updates return the true `ApiClient` but this is super unclear. Gonna change a bit to make it clearer

mwrock:
@jaym tweaked a bit to make the conversion to a `ApiClient` more clear by pulling the conversion out of the verskion branching logic and into the `run` call. So everything goes through the conversion and it just returns the original response if it is alreadt an `ApiClient`

jaym:
:+1:

spec/unit/knife/bootstrap/client_builder_spec.rb
```diff
@@ -41,6 +41,24 @@
...
+    let(:node) { Chef::Node.new }
+
+    before do
+      # mock out the rest of #run
+      expect(client_builder).to receive(:sanity_check)
+      expect(client_builder).to receive(:create_client!).and_return(client)
+      expect(client_builder).to receive(:create_node!).and_return(node)
+    end
+
+    it "returns the created client and nodeeturns " do
```
btm: *nodeeturns

lib/chef/knife/bootstrap.rb
```diff
@@ -375,9 +375,7 @@ def run
             exit 1
           end
 
-          client_builder.run
-
-          chef_vault_handler.run(node_name: config[:chef_node_name])
+          chef_vault_handler.run(client_builder.run[:client])
```
btm: Consider storing client and node in instance variables inside Chef::Knife::Bootstrap::Clientbuilder and pulling the client [and node] via an accessor.

```
client_builder.run
expect(client_builder.node).to eq(node)
expect(client_builder.client).to eq(client)
```

mwrock: I thought about that, but the values returned by the accessors will vary based on the state of the builder (if `run` has been called or not)

btm: how about

```
def client
  run unless @client
  @client
end

def node
  run unless @node
  @node
end
```

mwrock: Hmm. I think I like accessing via simple property over that. We'd want the methods to be `client!` and `node!` which feels a little odd to me in this case. I'm not "passionately" opposed to accessors that just return `nil` if run has not been called. Do you see issues with having them emitted from `run`?

spec/unit/knife/bootstrap/client_builder_spec.rb
mwrock: was sorta hoping nodeeturns would catch on. but yeah I dont think its gonna stick. People seem attached to "node"

lib/chef/knife/bootstrap.rb
btm: Returning an arbitrary hash from run is smelly. I think the pattern we use is that run tells the object to go do its thing and all we ever return from run is an indicator of success/failure.

Here's a similar example: https://github.com/chef/chef/blob/master/lib/chef/client.rb#L713-L716

Imagine a more extreme case where we returned 15 more properties from a run method that the object generated in a hash. Now you've got the class representing the object and this other hash that you're looking through. It'd be pretty silly to have a `person` object as well as a separate `things_we_know_about_person` hash.

It seems like there's a perfectly good ClientBuilder object there to use and we should get its properties from it rather then storing them in local variables.

mwrock: well I think this is somewhat different since this is a builder that we are treating like a client factory, but thats cool. I'll make the change.

mwrock:
removed hash from `run` @btm

btm:
:+1: on green

PR #4341, opened 2015-12-25
Deprecation waring by ruby 2.3.0: Thread.exclusive is deprecated, use Mutex.

sawanoboly:
Hi, Ruby 2.3 told that `Thread.exclusive is deprecated, use Mutex`.

It works also 1.9.3.

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -88,7 +88,7 @@ class Chef
       #
       def build_resource(type, name, created_at=nil, run_context: self.run_context, &resource_attrs_block)
         created_at ||= caller[0]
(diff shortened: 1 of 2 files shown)


infertux:
Why did you close this PR @sawanoboly? I think it should be fixed.

sawanoboly:
@infertux Ah, I'd noticed that it was worng usage of Mutex in my last PR.

I'm updating to avoid this message in https://github.com/higanworks/chef/commits/ruby23_Warning_Mutex.
And I'm examining how to fix in `lib/chef/dsl/declare_resource.rb` too.

infertux:
Sounds great, thank you. Looking forward to get rid of that pesky warning :)

PR #4343, opened 2015-12-28
use platform parser with metadata supports

jordane:
Using the cookbook version parser means that some platforms cannot be properly parsed by the supports method in metadata.

The included test metadata are two cases where it fails to parse because both are invalid cookbook versions but valid platform versions.


lamont-granquist:
:+1:

thommay:
:+1:

smurawski:
:+1:

PR #4347, opened 2015-12-29
add knife-windows to external specs

mwrock:


jaym:
:+1: :100:

btm:
:+1: 

There could be a chicken or egg issue here if both projects need a merge at the same time, but we can just call that out at the time. This is an overall win and we should have done it sooner.

PR #4348, opened 2015-12-30
Make handling of proxies more consistent

smith:
* Always use `*_proxy` environment variables.
* Make a `ChefConfig::Config.proxy_uri` method that gets used by
  `Chef::Provider::RemoteFile::FTP` and `Chef::HTTP::BasicClient`.
* Remove `env` method from `Chef::HTTP::BasicClient` (using
  `stub_const("ENV", ...)` in specs instead.)
* Remove `http_proxy_user` and `http_proxy_pass` methods from
  `Chef::HTTP::BasicClient` (replaced by functionality in `ChefConfig`.)

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -856,14 +856,14 @@ module ChefConfig
       # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
       # here since we are really just trying to get the string built correctly.
       proxy = if !proxy_env_var.empty?
-        if proxy_env_var.match(/^.*:\/\//)
-          URI.parse(proxy_env_var)
-        else
-          URI.parse("#{scheme}://#{proxy_env_var}")
-        end
-      end
+                if proxy_env_var =~ /^.*:\/\//
+                  URI.parse(proxy_env_var)
+                else
+                  URI.parse("#{scheme}://#{proxy_env_var}")
+                end
+              end
 
-      excludes = ENV['no_proxy'].to_s.split(/\s*,\s*/).compact
+      excludes = ENV["no_proxy"].to_s.split(/\s*,\s*/).compact
       excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
       return proxy unless excludes.any? { |exclude| File.fnmatch(exclude, "#{host}:#{port}") }
     end
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -846,6 +846,28 @@ module ChefConfig
       ENV["NO_PROXY"] = value unless ENV["NO_PROXY"]
     end
 
+    # Given a scheme, host, and port, return the correct proxy URI based on the
+    # set environment variables, unless exluded by no_proxy, in which case nil
+    # is returned
+    def self.proxy_uri(scheme, host, port)
+      proxy_env_var = ENV["#{scheme}_proxy"].to_s.strip
+
+      # Check if the proxy string contains a scheme. If not, add the url's scheme to the
+      # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
+      # here since we are really just trying to get the string built correctly.
+      proxy = if !proxy_env_var.empty?
+        if proxy_env_var.match(/^.*:\/\//)
+          URI.parse(proxy_env_var)
+        else
+          URI.parse("#{scheme}://#{proxy_env_var}")
+        end
+      end
+
+      excludes = ENV['no_proxy'].to_s.split(/\s*,\s*/).compact
+      excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
+      return proxy unless excludes.any? { |exclude| File.fnmatch(exclude, "#{host}:#{port}") }
+    end
+
     # Chef requires an English-language UTF-8 locale to function properly.  We attempt
     # to use the 'locale -a' command and search through a list of preferences until we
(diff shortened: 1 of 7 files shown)


adamedx:
@smith, just to confirm, there is still more work here to correctly support `no_proxy` in chef-client, right? Or do we feel this covers known `no_proxy` cases for chef-client? I'm interested since we're tracking `no-proxy` in general across components and it would be good to understand if this moves the ball forward there.

smith:
@adamedx this doesn't change any `no_proxy` behavior except for that [before in `BasicClient`](https://github.com/chef/chef/pull/4348/files#diff-19d41c6437108880128d0df418f9b4dcL114) we were testing for the env vars as well as the config options, and in [`RemoteFile::FTP`](https://github.com/chef/chef/pull/4348/files#diff-efab347b56b588611a27dc2f08ecd683L159) we were only testing for the config option, but this is no longer necessary, since we [export the `no_proxy` settings in `ChefConfig`](https://github.com/chef/chef/blob/3a9b7a7b7c35ec7ffb33a4c8eb20cb13abd5a4be/chef-config/lib/chef-config/config.rb#L731).

It appears that the behavior for no_proxy is correct. There are some basic specs in `ChefConfig` now and some more in [the specs for `Chef::Rest::AuthCredentials`](https://github.com/chef/chef/blob/nls/proxy-cleanup/spec/unit/rest/auth_credentials_spec.rb#L244-L270).

adamedx:
Thx @smith. @btm, @mwrockx, @jdm, can you take a look? Also, do we need to use this from knife-windows?

chef-config/lib/chef-config/config.rb
```diff
@@ -837,6 +837,28 @@ def self.export_no_proxy(value)
...
+    # set environment variables, unless exluded by no_proxy, in which case nil
+    # is returned
+    def self.proxy_uri(scheme, host, port)
+      proxy_env_var = ENV["#{scheme}_proxy"].to_s.strip
+
+      # Check if the proxy string contains a scheme. If not, add the url's scheme to the
+      # proxy before parsing. The regex /^.*:\/\// matches, for example, http://. Reusing proxy
+      # here since we are really just trying to get the string built correctly.
+      proxy = if !proxy_env_var.empty?
+                if proxy_env_var.match(/^.*:\/\//)
```
tyler-ball: The [old usage](https://github.com/chef/chef/pull/4348/files#diff-19d41c6437108880128d0df418f9b4dcL154) would pull out `*_proxy_user` and `*_proxy_pass` to add to the URI if they weren't embedded - seems like this should also do that to maintain functionality

randomcamel: @smith  I know it's been a while, but does this ring a bell?

randomcamel: Good catch. Should be fixed now.

tyler-ball:
:+1: Solid work

lib/chef/http/basic_client.rb
```diff
@@ -95,26 +95,8 @@ def request(method, url, req_body, base_headers = {})
...
-          else
-            proxy = URI.parse("#{url.scheme}://#{proxy.strip}")
-          end
-        end
-
-        no_proxy = Chef::Config[:no_proxy] || env["NO_PROXY"] || env["no_proxy"]
-        excludes = no_proxy.to_s.split(/\s*,\s*/).compact
-        excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
-        return proxy unless excludes.any? { |exclude| File.fnmatch(exclude, "#{host}:#{port}") }
+        @proxy_uri ||= ChefConfig::Config.proxy_uri(url.scheme, host, port)
```
thommay: think this should just be `Chef::Config.proxy_uri` for consistency's sake.

lib/chef/provider/remote_file/ftp.rb
```diff
@@ -146,19 +146,8 @@ def get
...
-          if Chef::Config["ftp_proxy_user"]
-            proxy.user = Chef::Config["ftp_proxy_user"]
-          end
-          if Chef::Config["ftp_proxy_pass"]
-            proxy.password = Chef::Config["ftp_proxy_pass"]
-          end
-          excludes = Chef::Config[:no_proxy].to_s.split(/\s*,\s*/).compact
-          excludes = excludes.map { |exclude| exclude =~ /:\d+$/ ? exclude : "#{exclude}:*" }
-          return proxy unless excludes.any? { |exclude| File.fnmatch(exclude, "#{host}:#{port}") }
+          ChefConfig::Config.proxy_uri("ftp", hostname, port)
```
thommay: here too

thommay:
trivial nits, otherwise :+1:

lib/chef/http/basic_client.rb
randomcamel: Not sure I understand: the [signature](https://github.com/chef/chef/pull/4348/files#diff-d19a42ae266a2919d7cfeb43c77dbaffR843) for `ChefConfig::Config.proxy_uri` is

```ruby
def self.proxy_uri(scheme, host, port)
```

thommay: I mean in the Chef::Config namespace not the ChefConfig::Config one.

tyler-ball: When we did the [original proxy exporting](https://github.com/berkshelf/ridley/blob/def392547123f2d846c41853c8f76587b0c3f3ea/lib/ridley/chef/config.rb) work we probably set a bad precedent for using `ChefConfig::Config` in Test Kitchen and Berkshelf

chefsalim:
:+1:

randomcamel:
Once master is green, this should be good to go after a rebase.

PR #4350, opened 2015-12-30
`@name_args` may contain some duplicated items.

tonytonyjan:
It makes `upload_failures` wrongly computed, and prints some weird warning message even though all cookbooks are successfully uploaded, for example:

```
$ knife cookbook upload nodejs --include-dependencies
Uploading nodejs         [2.4.4]
Uploading yum-epel       [0.6.5]
Uploading build-essential [2.2.4]
Uploading ark            [0.9.0]
Uploading apt            [2.9.2]
Uploading homebrew       [2.0.3]
Uploading yum            [3.8.2]
Uploading windows        [1.39.0]
Uploading 7-zip          [1.0.2]
Uploading chef_handler   [1.2.0]
WARNING: Uploaded 10 cookbooks ok but 2 cookbooks upload failed.
```

### DIFF
--- a/spec/unit/knife/cookbook_upload_spec.rb
+++ b/spec/unit/knife/cookbook_upload_spec.rb
@@ -183,6 +183,7 @@ E
         knife.config[:depends] = true
         allow(knife).to receive(:cookbook_names).and_return(["test_cookbook1", "test_cookbook2", "test_cookbook3"])
         expect(knife).to receive(:upload).exactly(3).times
+        expect(knife.cookbooks_to_upload.keys).to match_array knife.name_args
         expect do
           Timeout::timeout(5) do
             knife.run
--- a/lib/chef/knife/cookbook_upload.rb
+++ b/lib/chef/knife/cookbook_upload.rb
@@ -183,6 +183,7 @@ class Chef
                 Log.debug(e)
               end
             end
+            @name_args.uniq!
             upload_set
           end
       end


spec/unit/knife/cookbook_upload_spec.rb
```diff
@@ -179,10 +179,11 @@
         c
       end
 
-      it "should upload all dependencies once" do
+      it "should upload all dependencies once", focus: true do
```
coderanger: Oops :)

tonytonyjan: oops. My bad... Orz

tonytonyjan:
Chould be the similar issue: https://tickets.opscode.com/browse/CHEF-4896

tonytonyjan:
```
Chef::Exceptions::Win32APIError: No mapping between account names and security IDs was done.
```

any idea?

lamont-granquist:
rebase should get rid of the Win32APIError junk

lamont-granquist:
The logic here is completely tortured.

After this is merged what happens if we have a a really failed upload?  From the code I don't see where the two lengths will differ, but then again I'm not sure how they ever usefully differed in the failure case before this patch either.

lamont-granquist:
Yeah okay I think this line just needs to be dropped:

```
upload_failures += @name_args.length - @cookbooks_to_upload.length
```

And then I think the need to `uniq!` goes away?

Because after the uniq I don't see how this isn't just always zero.

tonytonyjan:
I felt the logic here is quite tortured as well, but It can be simply reproduced by upload [this cookbook](https://supermarket.chef.io/cookbooks/nodejs).

But `@name_args` will finally be:

```ruby
["nodejs", "yum-epel", "build-essential", "ark", "apt", "homebrew", "yum", "windows", "7-zip", "build-essential", "chef_handler", "windows"]
```

What I found  is that `"build-essential"` and `"windows"` are duplicated since the former is required by both `homebrew` and `nodejs`, the latter is required by both `ark` and  `7-zip`.

[This line](https://github.com/tonytonyjan/chef/blob/54df251d832c5b8169fac9277735bae59d9fb838/lib/chef/knife/cookbook_upload.rb#L178) does not take care of this sitruation. It just append all the dependencies without a check. If `uniq!` can't solve this problem, maybe we can put more control flow here?

lamont-granquist:
Right but after your `#uniq!` patch here isn't `@name_args.length - @cookbooks_to_upload.length` always going to be zero?  That suggests we just drop that line.  And if we drop that line then we don't really need the `#uniq!`.  All the actual failure cases are taken care of by exception handlers that increment upload_failures.

lamont-granquist:
lets accept this, but instead of fixing it this way just nuke the other check.

PR #4351, opened 2016-01-02
Allow loading bootstrap ssh-user from ssh_config

sawanoboly:
At present, `knife ssh` can load ssh-user from ssh-comfng.
But `knife bootstrap` sets root by default, it takes precedence over ssh-comfig.

This PR it is related lightly to #4038.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -47,8 +47,7 @@ class Chef
       option :ssh_user,
         :short => "-x USERNAME",
         :long => "--ssh-user USERNAME",
-        :description => "The ssh username",
-        :default => "root"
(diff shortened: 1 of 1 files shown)


sawanoboly:
Travis fails in the test from the knife-windows. What should I do ?

lamont-granquist:
What if we push the default down into Chef::Config?

lamont-granquist:
Ah, I see what you're trying to do, unfortunately I think you're going to break the world with this.

You probably want to introduce a --no-ssh-user flag to explicitly remove the default.

sawanoboly:
I see. It seem complicated, but not be helped. I'll add new option and create as another PR. Thanks.

PR #4355, opened 2016-01-05
don't blow up when json object has no recipes

kisoku:
fixes loading brand new node objects from json

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -545,7 +545,7 @@ class Chef
 
       if o.has_key?("run_list")
         node.run_list.reset!(o["run_list"])
-      else
+      elsif o.has_key?("recipes")
         o["recipes"].each { |r| node.recipes << r }
       end
 


coderanger:
We should add a regression test for this.

lib/chef/node.rb
```diff
@@ -546,7 +546,7 @@ def self.json_create(o)
       if o.has_key?("run_list")
         node.run_list.reset!(o["run_list"])
       else
-        o["recipes"].each { |r| node.recipes << r }
+        o["recipes"].each { |r| node.recipes << r } if o.has_key?("recipes")
```
thommay: seems like the logic would be better as part of the parent `if` statement:
```ruby
elsif o.has_key?("recipes")
...
```

kisoku:
@thommay diff updated as per your comments

kisoku:
@coderanger I can work on some tests, but unsure as to how to handle the json fixtures

lamont-granquist:
:+1:

smurawski:
:+1:

thommay:
I'm :+1:, think we should just go ahead and merge

PR #4367, opened 2016-01-07
Move server and repository fs objects to their own directories

jkeiser:


lamont-granquist:
:+1: after you rebase

thommay:
@jkeiser you'll need to rebase this but I'm :+1:

btm:
Just scanned it, looks fine. I'm indifferent about the "lib-backcompat" pattern. It isn't how we've done it in the past, so if we were talking about this being permanent we should probably consider options, but we can always just nuke it in Chef 13. Just need to make sure we don't forget. :+1:

PR #4370, opened 2016-01-10
Issue 3859 Text Update to Readme. Obvious Fix

PatrickWalker:
I hit an issue when trying to run the steps for local installation. 

I saw an issue that resolved it but the documentation didn't reference this step. I made a change to include the call to bundle exec rake gem


coderanger:
Thanks for the patch!

PR #4371, opened 2016-01-11
Ensure gem dependencies are secure

thommay:


Gemfile
```diff
@@ -30,6 +30,8 @@ group(:development, :test) do
 
   gem 'ruby-shadow', :platforms => :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
 
+  gem 'bundler-audit', git: 'https://github.com/rubysec/bundler-audit.git', ref: '4e32fca'
```
coderanger: Locking it to a ref kind of defeats the point I would think?

thommay: nope; `--update` in the travis run causes it to update the vulnerabilities db.

lamont-granquist:
LGTM, we'll see how annoying and/or useful it is :+1:

smurawski:
:+1:

PR #4376, opened 2016-01-12
prepending chef to tempfile name

anthonyneto:
prepending chef to tempfile name because we monitor file modifications in specific directories.

### DIFF
--- a/lib/chef/file_content_management/tempfile.rb
+++ b/lib/chef/file_content_management/tempfile.rb
@@ -65,6 +65,7 @@ class Chef
       def tempfile_basename
         basename = ::File.basename(@new_resource.name)
         basename.insert 0, "." unless Chef::Platform.windows?  # dotfile if we're not on windows
+        basename << '-chef-'
         basename
       end
 
--- a/lib/chef/file_content_management/tempfile.rb
+++ b/lib/chef/file_content_management/tempfile.rb
@@ -64,8 +64,8 @@ class Chef
       #
       def tempfile_basename
         basename = ::File.basename(@new_resource.name)
+        basename.insert 0, "chef-"
         basename.insert 0, "." unless Chef::Platform.windows?  # dotfile if we're not on windows
-        basename << '-chef-'
         basename
       end
 


coderanger:
This would append, not prepend (I think). Not sure what the use case is, most config scanners would ignore hidden files already.

lamont-granquist:
`basename.insert 0, "chef-"` before prepending the dot on unix should do it.

chasebolt:
his commit LGTM. looks like he is appending to the base filename, but prepending before the random tmp filename. personally i think the current commit looks the cleanest here.

before
`/tmp/.somefile.rb20160111-13548-17wr679`

after
`/tmp/.somefile.rb-chef-20160111-13548-17wr679`

lamont-granquist:
hmmm...  i sorta like the prepend because then `.chef*` matches it rather than `.*-chef-*` but YMMV

chasebolt:
i can get down with it either way. i just prefer mucking with filenames to the right of the original filename so it is easier to read (if ever needed, which in this case would probably never really be). `.chef-*` would be easier to match on.

this is really just coming down to personal preference atm.

lamont-granquist:
:+1: either way

lamont-granquist:
@chef/client-core needs another :+1:

coderanger:
I disagree that this is useful. If the goal is making blacklist globs for some kind of software it would almost certainly need to be a prefix match like `exclude conf.d/.chef-*` but with that you could just ignore `conf.d/.*` (and most software does that by default because of vim/etc temp files) so I would need to see some kind of reasoning beyond "maybe useful".

chasebolt:
use case: we monitor application directories for file modifications/creation/deletion. we have a whitelist of files we do ignore but we do not want to whitelist all hidden files. this would allow us to only ignore hidden files created by chef.

lamont-granquist:
So current version is now a ".chef-*" prefix.   I'm mostly "meh" on this but I can't think of anything really bad, and this makes the regex as easy as possible.  I'm also amused by people who care about watching  dotfiles come and go in their directories and think that's going to drive you insane, but whatever...

Someone can still have their rootkit drop an /etc/.chef-youcantfindme-nyeah-nyeah-nyeah file and the tool will have to ignore that as well, you just push around the problem (and better rootkits will inject kernel modules and none of this will work at all..)

chasebolt:
security should be built in layers and this is just one of the base layers we happen to use; there is no _one perfect_ solution for catching all malicious activity.

coderanger:
With is as `.chef-` I'm -0 instead of -1, mostly useless IMO but I'm not actively against it. Also I would prefer to see some tests, the fact that we can commit this with no tests breaking is worrisome :)

lamont-granquist:
yeah since you care about it, a test would ensure it does not break in the future.

PR #4377, opened 2016-01-12
fixing candidate filtering and adding functional tests for chocolatey_package

mwrock:
Functional tests using small dummy packages against a local feed.


mwrock:
cc @lamont-granquist

mwrock:
I found some issues when adding these and files a separate pr #4378 to fix. I can add those tests here when that is merged.

Also, I created these tests based on the the doc https://docs.chef.io/release/12-7/resource_chocolatey_package.html and now see we should add tests for upgrade too

lamont-granquist:
LGTM :+1:

smurawski:
:+1:

mwrock:
In the process of adding functional tests, I stumbled accross a couple more issues:

1. The parsing of the choco list was building a hash of name/version pairs, and therefore overwriting older versions with the latest version. I missed this making my changes responding to @lamont-granquist 's comments. So now the initial parsing creates an array of name/versions which are later filtered when determining available versions.

2. There were issues with name casing. The final name/version hash uses a mixed case key and when filtered against pinned versions that key was compared to a forced lowercase package name. So I convert the name/version pairs to use lowercase names now making the casing equal on both sides of the comparison.

I have added and edited unit tests to account for these cases  in addition to adding functional tests. I'm going to request another review/thumbs up before merging this since there was a fair amount of change to the provider.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -233,12 +240,16 @@ def installed_packages
         # @param cmd [String] command to run
         # @return [String] list output converted to ruby Hash
```
lamont-granquist: doc needs to be updated with an array now

lib/chef/provider/package/chocolatey.rb
```diff
@@ -233,12 +240,16 @@ def installed_packages
...
           choco_command(*args).stdout.each_line do |line|
             name, version = line.split('|')
-            hash[name.downcase] = version.chomp
+            list << [ name.downcase, version.chomp ]
           end
-          hash
+          list
+        end
+
+        def lowercase_names(names)
```
lamont-granquist: should have yard doc for this function as well, even though its pretty trivial

lib/chef/provider/package/chocolatey.rb
```diff
@@ -211,11 +211,18 @@ def args_to_string(*args)
         def available_packages
           @available_packages ||=
             begin
+              name = 0
+              version = 1
```
lamont-granquist: if i understand it, these are array indexes, and i think this is hindering readability more than its helping it...

its also plausible that these would wind up colliding with other variables/methods later and causing a bit of scope hell..

lib/chef/provider/package/chocolatey.rb
```diff
@@ -211,11 +211,18 @@ def args_to_string(*args)
...
+
               cmd = [ "list -ar #{package_name_array.join ' '}" ]
               cmd.push( "-source #{new_resource.source}" ) if new_resource.source
-              parse_list_output(*cmd).reject do |name,version|
-                desired_name_versions[name] && desired_name_versions[name] != version
+              available = {}
+              parse_list_output(*cmd).each do |name_version|
```
lamont-granquist: i'd probably just do something like:

```ruby
parse_list_output(*cmd).each do |name_version|
  name, version = name_version
```

then `name` and `version` are really the name and version in the loop

lib/chef/provider/package/chocolatey.rb
```diff
@@ -211,11 +211,18 @@ def args_to_string(*args)
...
+                if desired_name_versions[name_version[name]].nil? || desired_name_versions[name_version[name]] == name_version[version]
+                  available[name_version[name]] = name_version[version]
+                end
               end
+              available
```
lamont-granquist: i suspect you can use a more functional notation (#each_with_object?) and get rid of the create-hash, loop-over-thing-populating-hash, return hash pattern.

mwrock:
responded to @lamont-granquist review. Great to know about `each_with_object`!

lamont-granquist:
:+1:  LGTM

looks like you need to rebase -- suspect that the double-quotes PR trolled you.

@chef/client-core and @chef/client-windows need review

danielsdeleo:
:+1:

PR #4378, opened 2016-01-12
assert candidates exist for alternate sources and when pinning versions

mwrock:
when putting together functional tests for #4377 I noticed a couple problems:

* bogus packages were allowed to be installed when using an alternate source
* bogus versions were allowed to be installed regardless of source

This ensures that candidate versions only include pinned versions when pinning to a specific version and that we assert candidates exist when using an alternate source.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -40,6 +40,19 @@ class Chef
           current_resource
         end
 
+        def define_resource_requirements
+          super
+
+          # Chocolatey source attribute points to an alternate feed
+          # and not a package specific alternate source like other providers
+          # so we want to assert candidates exist for the alternate source
+          requirements.assert(:upgrade, :install) do |a|
+            a.assertion  { candidates_exist_for_all_uninstalled? }
+            a.failure_message(Chef::Exceptions::Package, "No candidate version available for #{packages_missing_candidates.join(", ")}")
+            a.whyrun("Assuming a repository that offers #{packages_missing_candidates.join(", ")} would have been configured")
+          end
+        end
+
         # Lazy initializer for candidate_version.  A nil value means that there is no candidate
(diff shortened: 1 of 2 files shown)


mwrock:
cc @lamont-granquist

lib/chef/provider/package/chocolatey.rb
```diff
@@ -220,7 +233,10 @@ def parse_list_output(*args)
           hash = {}
           choco_command(*args).stdout.each_line do |line|
             name, version = line.split('|')
-            hash[name.downcase] = version.chomp
+            version = version.chomp
+            if desired_name_versions[name].nil? || desired_name_versions[name] == version
+              hash[name.downcase] = version
+            end
```
lamont-granquist: so the `list -ar` lists all available versions, then we're assured that if we want to install version 1.2.3 that it'll eventually match here, even though later versions may be available?

what about for installed packages?  i think it may superficially get the behavior right since if you've got a version that doesn't match it'll return nil like its uninstalled, but that does get the semantics of the current_resource.version wrong.

lamont-granquist: and you could do with a comment or two explaining all that...

mwrock:
good call on the wacked semantics for `installed_packages` I just moved the logic into `available_packages` to clear that up and added another comment line for (hopefully) further clarification.

And yes, even if not pinning to the latest version, the assertion will work as long as the pinned version is valid.

lamont-granquist:
sweet :+1: 

i might suggest adding some specific tests around the intalled_packages + available_packages helpers -- you'd want to peel off a new `describe` block for both of those and then probably use `send` if they're private APIs.

violates the principle of unit testing slightly, but it could help to catch the kinds of edge conditions we just talked about here...

although if you wanna just ship it that's cool too...

smurawski:
:+1:

PR #4379, opened 2016-01-12
Per RFC 64, do style checking of the Chef repo

thommay:
This commit enables ChefStyle, but with no active cops. Cops will be
added per the RFC (one at a time, with a supporting PR to this repo) as
we go on.

cc @chef/client-maintainers


lamont-granquist:
:+1:

PR #4381, opened 2016-01-12
chefstyle: fix Lint/StringConversionInInterpolation

lamont-granquist:
useless use of `"#{foo.to_s}"`


lamont-granquist:
@chef/client-core this should be entirely uncontroversial as well

thommay:
:+1:

stevendanna:
:+1:

PR #4384, opened 2016-01-12
really fix cops

lamont-granquist:


lamont-granquist:
@chef/client-core sorry i assumed that autofixable cops would always autofix.  sometimes they don't.

ran all the bad files through vim auto indentation fixing.  then fixed a few more manual ones that vim wouldn't fix.

thommay:
I took a look through this and it seemed fine... :+1:

danielsdeleo:
:+1:

PR #4389, opened 2016-01-13
Refactor versioned from unversioned cookbooks

jkeiser:
Should make it easier to work with; subclassing to change behavior instead of switch statements.

### DIFF
--- a/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_cookbook_dir.rb
+++ b/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_cookbook_dir.rb
@@ -28,25 +28,9 @@ class Chef
     module FileSystem
       module Repository
         class ChefRepositoryFileSystemCookbookDir < ChefRepositoryFileSystemCookbookEntry
-          def initialize(name, parent, file_path = nil)
-            super(name, parent, file_path)
-          end
-
           def chef_object
             begin
-              loader = Chef::Cookbook::CookbookVersionLoader.new(file_path, parent.chefignore)
-              # We need the canonical cookbook name if we are using versioned cookbooks, but we don't
-              # want to spend a lot of time adding code to the main Chef libraries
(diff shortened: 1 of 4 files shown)


lib/chef/chef_fs/file_system/chef_server/chef_server_root_dir.rb
```diff
@@ -141,7 +142,7 @@ def children
             @children ||= begin
               result = [
                 # /cookbooks
-                CookbooksDir.new("cookbooks", self),
+                versioned_cookbooks ? VersionedCookbooksDir.new("cookbooks", self) : CookbooksDir.new("cookbooks", self),
```
randomcamel: Does this mean we can't use versioned and unversioned cookbooks at the same time?

lib/chef/chef_fs/file_system/chef_server/cookbooks_dir.rb
```diff
@@ -29,6 +29,15 @@ class Chef
...
       module ChefServer
+        # cookbooks/
+        #   apache2/
+        #   mysql/
+        # cookbook_artifacts/
+        #   apache2/
+        #     1.0.0/
+        #     1.0.1/
+        #   mysql/
+        #     2.0.5/
```
randomcamel: For clarity, the `cookbook_artifacts` should probably have identifiers like the hash strings they'll be in practice.

lib/chef/chef_fs/file_system/chef_server/chef_server_root_dir.rb
jkeiser: This is correct. It's repo-wide.

lib/chef/chef_fs/file_system/chef_server/cookbooks_dir.rb
jkeiser: Agreed.

PR #4390, opened 2016-01-13
Correct uses of fail to raise

thommay:
Style/SignalException - 3

@chef/client-maintainers


coderanger:
:+1: (though `fail` is oddly common in Rake tasks, but I can't find any evidence that it's a DSL method or otherwise overloaded)

lamont-granquist:
:+1: 

not sure why i couldn't find these yesterday

PR #4393, opened 2016-01-13
fix Lint/DuplicatedKey

lamont-granquist:
so, this is a bit special...

the API for "priority" is clearly broken.  but this broken API comes
from the service resource itself which appears to have been broken
by design since at least 2010.

just fixing the code for how it behaves and dropping in FIXMEs
for now, since this uncovered a fairly massive API-breaking yak.


lamont-granquist:
@chef/client-core here's a good :facepalm:

thommay:
gosh. :+1:

stevendanna:
:+1:

danielsdeleo:
:+1:

PR #4398, opened 2016-01-13
autocorrecting Style/TrailingComma

lamont-granquist:
'chefstyle -a' fixed 1044 occurrences


lamont-granquist:
@chef/client-core add a whole lotta trailing commas

danielsdeleo:
Is it possible to enforce this only for multi-line arrays/hashes? After seeing the effect, I'm inclined to not care about trailing commas for single line stuff, if we can pull it off. :+1: if not.

lamont-granquist:
it actually does it only for multi-line.

the problem is that this is multi-line:

```ruby
         sip_paths= [
            '/System', '/bin', '/sbin', '/usr',
          ]
```

that could get fixed to either:

```ruby
         sip_paths= [ '/System', '/bin', '/sbin', '/usr' ]
```

or:

```
         sip_paths= [
            '/System',
            '/bin',
            '/sbin',
            '/usr',
          ]
```

but rubocop can't autocorrect those.

lib/chef/chef_fs/data_handler/acl_data_handler.rb
```diff
@@ -11,8 +11,8 @@ def normalize(acl, entry)
             'read' => {},
             'update' => {},
             'delete' => {},
-            'grant' => {}
-            })
+            'grant' => {},
+            },)
```
lamont-granquist: ^ there's another edge case here which is less than delightful, but i believe there's another style cop which flags this as a bad way to match up the trailing paren, we just haven't gotten there yet.

lamont-granquist: actually it doesn't and that is mildly gross...

danielsdeleo:
Ok, I see now. Carry on, then. :+1:

spec/functional/resource/execute_spec.rb
```diff
@@ -89,7 +89,7 @@
       resource.environment({
         "SAWS_SECRET"  => "supersecret",
         "SAWS_KEY"     => "qwerty",
-      })
+      },)
```
lamont-granquist: so i think this one gets cleaned up later by dropping the braces entirely.

lib/chef/chef_fs/data_handler/acl_data_handler.rb
lamont-granquist: the `{}` in the function call are actually optional here as well...  so all these get cleaned up later...

thommay:
:+1:

PR #4403, opened 2016-01-13
knife upload/download for cookbook artifacts

jkeiser:


chef-config/lib/chef-config/config.rb
```diff
@@ -151,6 +151,16 @@ def self.derive_path_from_chef_repo_path(child_path)
...
     default(:client_path) { derive_path_from_chef_repo_path('clients') }
 
+    # Location of containers on disk. String or array of strings.
+    # Defaults to <chef_repo_path>/containers.
+    # Only applies to Enterprise Chef commands.
+    default(:container_path) { derive_path_from_chef_repo_path('containers') }
+
+    # Location of cookbook_artifacts on disk. String or array of strings.
+    # Defaults to <chef_repo_path>/cookbook_artifacts.
+    # Only applies to Enterprise Chef commands.
```
danielsdeleo: I guess the whole "enterprise" vs. not is kinda moot now, but Cookbook artifacts are objects that aren't related to the authz system so I don't think this is accurate anyway?

lib/chef/chef_fs/file_system/chef_server/chef_server_root_dir.rb
```diff
@@ -92,6 +94,7 @@ def initialize(root_name, chef_config, options = {})
           attr_reader :repo_mode
           attr_reader :cookbook_version
           attr_reader :versioned_cookbooks
+          attr_reader :policy_document_native_api
```
danielsdeleo: Seems like this is unused? I don't think it would affect the behavior of the server anyway, just the client.

danielsdeleo:
two small nits, but :+1: as long as travis is happy

chef-config/lib/chef-config/config.rb
jkeiser: Yep! We should remove Enterprise Chef from stuff.

PR #4405, opened 2016-01-14
ServerAPI will return a raw hash, so do that

thommay:


thommay:
@chef/client-core

lamont-granquist:
:+1:

danielsdeleo:
:+1:

PR #4407, opened 2016-01-14
Fix english grammar

jorisd:
Cosmetic change for correct english when a run chef happened less than a minute ago:

```
$ knife status
[...]
0 minutes ago, flow01-uk.blah.eu, centos 6.7.
```


coderanger:
`minutes` is correct in this situation unless I'm missing something.

tas50:
Noah is right here.  0 minutes is the correct form.

thommay:
yep, 0 minutes, 1 minute, 2 minutes is correct.

thommay:
Thanks for your contribution, but we won't be accepting it since the current wording is correct.

jorisd:
Thanks, I have learnt something !

PR #4413, opened 2016-01-14
configure chef client windows service to the correct chef directory

mwrock:
If a user installs chef to a directory other than c:\opscode\chef via MSI and installs the chef-client service as well., the service will not run because it points to the default location.

I have validated this fix by building a chef client with the changed wix.


danielsdeleo:
dunno enough about windows to tell if this is right, but you tested it and the code change looks reasonable. :+1:

ksubrama:
:+1: Good catch!

mwrock:
travis is having issues unrelated to this change so merging. This change would not affect travis anyways

PR #4414, opened 2016-01-15
explicitly adding .bat to service install command in installer

mwrock:
for those who dont have `.bat` in `PATHEXT`


danielsdeleo:
Not sure what's up with appveyor, I restarted it. We should try to carve out some time to make that more resilient. :+1: for this change though

mwrock:
closing this. After further investigation and talking with Naomi. this is not the right fix.

PR #4415, opened 2016-01-15
fix chefstyle failures

lamont-granquist:
autocorrects some merges that will be broken when the next
chefstyle is released.

also runs chefstyle after the travis tests, because i'd rather know
how the code is broken rather than how its not pretty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ branches:
 # do not run expensive spec tests on PRs, only on branches
 script: "
 set -e;
-bundle exec rake style;
 echo '--color\n-fp' > .rspec;
 sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers;
 sudo -E $(which bundle) exec rake spec;
+bundle exec rake style;
 "
 
 env:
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core this mostly just fixes stuff that we've already fixed before.

we need a chefstyle release to start enforcing all these rules.

the fact that we equality pin on chefstyle while travis does not do tests against rebased master is pretty annoying since people will only get whatever chefstyle rules were present when they forked.  we may actually want to let chefstyle float while having it equality pin rubocop...

danielsdeleo:
> we may actually want to let chefstyle float while having it equality pin rubocop...

We should give this a go.

And :+1: on this PR, too

lamont-granquist:
added removing the chefstyle pin to this PR...

thommay:
I equality pinned because that's what we talked about in the RFC, but I think this makes much more sense, especially whilst we come up with a rule set. :+1:

thommay:
(we could equally float on master, with the assumption that rules only get merged to chefstyle once they're fixed here)

lamont-granquist:
I thought you were POETSing?  =)

I think we can start by floating on rubygems so that we can merge to master in chefstyle, and fix here and ohai, and then release the gem later to start enforcing.  I'm also testing rebasing chef/chef to master before running rspec in travis...

PR #4416, opened 2016-01-15
try rebasing against master first on travis

lamont-granquist:


coderanger:
This seems weird and error prone? Definitely not what I would expect.

Then again I'm a dirty rotten merge commit lover :)

lamont-granquist:
yeah so its possible that i'm just confused and don't need to do this at all...

i thought travis only built off the head of the PR branch.  the docs seem to suggest that it builds the merge commit from the branch and the root.  i don't think i've ever seen a travis failure due to inability to merge back into master though, while i definitely have for appveyor, so i'm definitely confused..

lamont-granquist:
TIL: travis actually does build against the merge commit...

PR #4418, opened 2016-01-15
Cookbook artifacts local mode

jkeiser:


randomcamel:
:+1: assuming it still works in non-local mode. :smiley:

PR #4419, opened 2016-01-16
explicitly adding .bat to service executable called by service in cae users remove .bat from PATHEXT

mwrock:


mwrock:
I also removed a `\` from the wix which is harmless but creates 2 backslashes in the service executable and looks weird

danielsdeleo:
:+1:

pburkholder:
Thanks for the quick fix, @mwrock !

randomcamel:
:+1:

PR #4423, opened 2016-01-18
Add deprecation warnings to Chef::REST and all json_creates

thommay:


lamont-granquist:
:+1:

thommay:
Not sure what's going on with the appveyor failure. ping @chef/client-core

stevendanna:
Key, Org, User, and UserV1 all alias json_create to from_hash but I don't see those modified here to include the deprecation.  I think we probably still want to since calling json_create usually implies that auto-inflation is being used.

danielsdeleo:
:+1:

stevendanna:
:+1:

PR #4424, opened 2016-01-18
another chefstyle catchup + pull master

lamont-granquist:
bunch of merges over the weekend need re-fixing again.

seems like pointing at master of chefstyle is probably a good idea to
start making people fix their code before merging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,9 @@
 source "https://rubygems.org"
-gemspec :name => "chef"
+gemspec name: "chef"
 
-gem "activesupport", "< 4.0.0", :group => :compat_testing, :platform => "ruby"
+gem "activesupport", "< 4.0.0", group: :compat_testing, platform: "ruby"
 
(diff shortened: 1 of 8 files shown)


thommay:
:+1:

lamont-granquist:
@chef/client-core this autofixes a bunch of merges that were made since friday.

also pulls master of chefstyle to start fixing these up-front rather than after-the-fact.

danielsdeleo:
Looks like the kitchen tests currently fail with git-sourced gems in your Gemfile?

lamont-granquist:
yeah i don't understand how though, and the kitchen-appbunndle-updater version we're running is ancient...   see #4426

danielsdeleo:
AFAICT it's because there is no git on the box: https://travis-ci.org/chef/chef/jobs/103156314#L368

lamont-granquist:
fixed

PR #4426, opened 2016-01-18
let kitchen-appbundle-updater float

lamont-granquist:
i'm a bit surprised this still even works at all..

### DIFF
--- a/kitchen-tests/Gemfile
+++ b/kitchen-tests/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 group :end_to_end do
   gem "berkshelf"
   gem "test-kitchen", "~> 1.4"
-  gem "kitchen-appbundle-updater", "~> 0.0.1"
+  gem "kitchen-appbundle-updater"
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core assuming this works, we should do this to automatically test new releases of kitchen-appbundle-updater and to get the new code (which should be somewhat faster and more reliable now).

lamont-granquist:
can we just !#$%!#$%@#$% shoot everyone who thinks servers need to not have compilers in the 21st century and start including them by default on all base images?

the last time it made sense any sense was sometime in the 90s when you had to install compilers from CDROMs and they weren't available via a simple 'apt-get install build-essential' or 'yum install gcc' and therefore not having them on the boxes actually provided a barrier to hackers and users installing them...

lamont-granquist:
@chef/client-core okay fixed kitchen-appbundle-updater to start installing git and build-essential

danielsdeleo:
:+1:

smurawski:
:+1:

PR #4428, opened 2016-01-18
Fix for #3888: Define method_name in Chef::Cookbook::Metadata.validate_json if undefined

### LINKED ISSUE #3888: knife cookbook site share doesn't work
I'm getting the error when trying to release a new version for the aegir3 cookbook (https://github.com/daften/cookbook-aegir3):

```
This may be a bug in the 'cookbook site share' knife command or plugin
Exception: NameError: undefined local variable or method `method_name' for Chef::Cookbook::Metadata:Class
```

andy-dufour:
It looks like method_name could never be properly scoped here? 

It's inclusion makes any cookbook with a dependency and no metadata.rb in a cookbook fail sharing to both public and private supermarkets.

Line 626 throws an exception if method_name is undefined.

For now I've assigned an empty string if its undefined. Open for input @chef/client-core 

closes #3888

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -641,8 +641,8 @@ class Chef
         VERSION_CONSTRAINTS.each do |dependency_type, hash_key|
           if dependency_group = o[hash_key]
             dependency_group.each do |cb_name, constraints|
-              if metadata.respond_to?(method_name)
-                metadata.public_send(method_name, cb_name, *Array(constraints))
+              if metadata.respond_to?(dependency_type)
+                metadata.public_send(dependency_type, cb_name, *Array(constraints))
               end
             end
           end
--- a/spec/unit/knife/cookbook_metadata_spec.rb
+++ b/spec/unit/knife/cookbook_metadata_spec.rb
@@ -24,6 +24,7 @@ describe Chef::Knife::CookbookMetadata do
     @knife.name_args = ["foobar"]
     @cookbook_dir = Dir.mktmpdir
     @json_data = '{ "version": "1.0.0" }'
+    @invalid_json_data = '{ "version": "1.0.0", {ImInvalid}}'
     @stdout = StringIO.new
     @stderr = StringIO.new
     allow(@knife.ui).to receive(:stdout).and_return(@stdout)
@@ -109,7 +110,7 @@ describe Chef::Knife::CookbookMetadata do
       @json_file_mock = double("json_file")
     end
 
-    it "should generate the metatdata json from metatdata.rb" do
+    it "should generate the metadata json from metadata.rb" do
       allow(Chef::Cookbook::Metadata).to receive(:new).and_return(@metadata_mock)
       expect(@metadata_mock).to receive(:name).with("foobar")
       expect(@metadata_mock).to receive(:from_file).with("#{@cookbook_dir}/foobar/metadata.rb")
@@ -173,6 +174,15 @@ describe Chef::Knife::CookbookMetadata do
         expect(@stderr.string).to match /in #{@cookbook_dir}\/foobar\/metadata\.json/im
         expect(@stderr.string).to match /#{description} blah/im
       end
+    end
+    it "should raise a syntax error on invalid JSON" do
+      expect(File).to receive(:exist?).with("#{@cookbook_dir}/foobar/metadata.json").
+        and_return(true)
+      expect(IO).to receive(:read).with("#{@cookbook_dir}/foobar/metadata.json").
+        and_return(@invalid_json_data)
+      expect do
+        @knife.validate_metadata_json(@cookbook_dir, "foobar")
+      end.to raise_error(Chef::Exceptions::JSON::ParseError)
     end
   end
 


lamont-granquist:
uhm....

whatever all this `method_name` code is doing its clearly terrible.  is there some closure somewhere that makes this actually defined sometimes?  on the surface it looks like some kind of bad copypasta error that is always broken...

anyway, i'd really like to see a more aggressive patch that did more than just apply a bit of duct tape...

andy-dufour:
So far I can't find any way that method_name would actually be defined, so I think any cookbook that has any version constraint and only a metadata.json will fail to upload via knife cookbook site share or knife supermarket site share.

I wanted a 2nd set of eyes on this before I tried something more verbose, I'll take a deeper look at what we're trying to validate and how we might do it tomorrow.

andy-dufour:
I've updated this to behave in what appears to be the same way as when we grab the metadata from a metadata.rb.

I did a quick test, killing my JSON name attribute and I receive the expected 

`ERROR: Chef::Exceptions::MetadataNotValid: Cookbook loaded at path(s) [/Users/andrewdufour/Development/cookbooks/push-jobs] has invalid metadata: The `name' attribute is required in cookbook metadata`

We should probably have some tests surrounding this. Also I believe we can remove that validate_json method if this is a good approach.

lamont-granquist:
Isn't the whole `#validate_json` approach doing a lot of work to solve a problem indirectly, and we can just construct the object and let it fail with a NoMethodError if the json isn't valid?  Then if we want to we can rescue the NoMethodError and give the user a 'nicer' looking error message and/or re-raise with Chef::Exceptions::MetadataNotValid that suggests to the user that their json has bad keys in it...

andy-dufour:
I tried to just pass this down to Chef::Cookbook::Metadata.from_json as a json string, I think because it isn't eval'd like from_file is it doesn't do any version constraint checks. So a version constraint in the json of 'something': '>= a.b.c' was passing through validation.

Since we probably don't want to let people bork their version constraints I think validating the json's constraints makes sense, and I've added a reworked validate_json method to the PR (it's not really validating json, it's validating the json's constraints), and fixed the failing tests.

andy-dufour:
Alright so after spending too much time on this the last couple of days I'll explain how I arrived where we are.

When you read in a metadata.rb we use Chef::Mixin::FromFile.from_file which evals from the context of the Chef::Cookbook::Metadata object. When you execute methods like 'depends' it runs validation against the version constraints in the metadata.rb.

We were correct in assuming we'd throw an exception in the case of an invalid JSON object. But when you read in a json file and run Chef::Cookbook:Metadata.from_json you never validate version constraints like you do with from_file.

The result is you can have invalid version constraints in your valid json - e.g., {"dependencies":{"runit":">> a.b.c"}} and this will be happily processed as valid metadata, whereas we throw an exception if the same dependency is listed in a metadata.rb.

So, this validate_json method is really validating the json's constraints. I tried a few approaches like adding this to run_validation in Chef::Cookbook::Metadata but it causes a bunch of errors from Pedant running against chef-zero, that will take a bit of effort to fix.

As it stands this seems to me to be the simplest solution that isn't a complete hack that catches both invalid JSON and invalid constraints. If we want to go deeper to patch this I'll likely need some guidance and assistance.

andy-dufour:
@chef/client-core could I get another look at this? I know it's not ideal in its current state, but it puts our share commands in a better spot.

If we want to improve this more I am willing to do the work but may need some guidance.

lib/chef/cookbook/metadata.rb
```diff
@@ -618,12 +618,11 @@ def self.from_json(string)
...
       def self.validate_json(json_str)
         o = Chef::JSONCompat.from_json(json_str)
-        metadata = new()
         VERSION_CONSTRAINTS.each do |dependency_type, hash_key|
           if dependency_group = o[hash_key]
            dependency_group.each do |cb_name, constraints|
-             if metadata.respond_to?(method_name)
-               metadata.public_send(method_name, cb_name, *Array(constraints))
+             if self.respond_to?(dependency_type)
+               self.public_send(dependency_type, cb_name, *Array(constraints))
```
lamont-granquist: I'm pretty certain you don't want to change from doing this against an instance to doing this against the class (its a class method so 'self' is the class).

I think what is going on is that the respond_to? is normally a worthless check that always passes, since you're looping over a constant that the class has set on it, and all the dependency_types that you loop over are valid methods on the instances.  But when you try to use 'self' then the respond_to check silently masks all the NoMethodErrors that you would otherwise be getting.  So I think if you drop the respond_to? check entirely that you'll see those failures.

lib/chef/knife/cookbook_metadata.rb
```diff
@@ -93,7 +93,10 @@ def generate_metadata_from_file(cookbook, file)
       def validate_metadata_json(path, cookbook)
         json_file = File.join(path, cookbook, "metadata.json")
         if File.exist?(json_file)
-          Chef::Cookbook::Metadata.validate_json(IO.read(json_file))
+          ui.info("Generating metadata for #{cookbook} from #{json_file}")
+          md = Chef::Cookbook::Metadata.new
+          md.name(cookbook)
+          md.validate_json(IO.read(json_file))
```
lamont-granquist: oh yeah you're trying to convert this to an instance method from the class.  so you gotta go back up there and change `def self.validate_json` to `def validate_json`

spec/unit/knife/cookbook_metadata_spec.rb
```diff
@@ -139,12 +139,17 @@
...
+      allow(Chef::Cookbook::Metadata).to receive(:new).and_return(@metadata_mock)
+      expect(@metadata_mock).to receive(:name).with("foobar")
       expect(File).to receive(:exist?).with("#{@cookbook_dir}/foobar/metadata.json").
-                                   and_return(true)
-      expect(IO).to receive(:read).with("#{@cookbook_dir}/foobar/metadata.json").
-                               and_return(@json_data)
-      expect(Chef::Cookbook::Metadata).to receive(:validate_json).with(@json_data)
+                                      and_return(true)
+      expect(IO).to receive(:read).with("#{@cookbook_dir}/foobar/metadata.json").and_return(@json_data)
+      expect(@metadata_mock).to receive(:validate_json).with(@json_data)
```
lamont-granquist: this isn't blowing up because metadata_mock needs to be an "instance_double(Chef::Metadata)"  if you start by just changing the double to an instance_double you'll see you can't mock validate_json because that is a method on the class not the instance.  you mocked it here, so rspec is obliviously happy with it, because its not validating that the underlying object actually has a method to mock.

lib/chef/cookbook/metadata.rb
```diff
@@ -618,12 +618,11 @@ def self.from_json(string)
 
       def self.validate_json(json_str)
```
lamont-granquist: per other comments below this should be `def validate_json(json_str)` -- however you should keep this API intact and still have a class method that builds an instance and validates it so that you don't break any other consumers of it (consider the class method a public API that you can't drop).

lamont-granquist:
actually even after all that, the shape of the API doesn't make a lot of sense, since your debug message states that you're loading the metadata, but it looks like you're trying to call validate_json to load the metadata, but it both doesn't do that, and that's an odd side-effect for a 'validate' method to have.

jorhett:
Just curious why this issue from 2015 is still not addressed?

lamont-granquist:
time

DPD-30:
what is the status of the above PR?

andy-dufour:
I was finally able to pick this back up and look at it today, and noticed that @mrigeshpriyadarshi in https://github.com/chef/chef/issues/5201 was able to glean the original developers intent, that I missed.

This is a simplified solution, that, although not pretty, does work without reworking the validation methods of the class. I was also able to add a spec test to ensure we _**actually**_ test some invalid JSON to make sure we're parsing things.

@lamont-granquist if you want me to take this further and rewrite the validation methods I can give it another stab later this week. It would be great if we could merge this as a fix for anyone with only a metadata.json though.

spec/unit/knife/cookbook_metadata_spec.rb
```diff
@@ -174,6 +175,15 @@
         expect(@stderr.string).to match /#{description} blah/im
       end
     end
+    it "should raise a syntax error on invalid JSON" do
+      expect(File).to receive(:exist?).with("#{@cookbook_dir}/foobar/metadata.json").
+        and_return(true)
+      expect(IO).to receive(:read).with("#{@cookbook_dir}/foobar/metadata.json").
+        and_return(@invalid_json_data)
```
lamont-granquist: i really dislike stubbing the filesystem like this.   if you try to read the file provider unit tests you'll discover where this madness winds up.

its generally preferable to consider the filesystem to be part of the "system-under-test" (SUT) and to just create a real file.  either dynamically generate a real tempfile, or use a cookbook fixture which has a real metadata.rb file in it with bad data.

that violates the principle that "unit tests should only test the object in question" and expanding the unit test to cover other objects (like the filesystem on the whole) but the amount of collaboration between objects and the underlying filesystem is huge -- a POSIX filesystem is an extremely flawed API from the perspective of testing, and just gets worse when you layer all of ruby's file classes on top of it (FileUtils, File, IO, Pathname, etc, etc) and going down the route of a MockFS or FakeFS style mocking of the entire filesystem API is pretty much madness.

it could be argued that the resultant tests using real files aren't strictly "unit" tests, but that gets into the philosophy of unit testing -- and we have lots of "unit" testing that expands the SUT to encompass other objects.  when i was learning about unit testing, i tried to 'fix' them and it wound up extremely messy and brittle.   people could argue that is because the design of the code should be bad and it should all be rewritten to have less collaboration between objects and simpler interfaces, and i'd totally agree, but lacking the time to rewrite the codebase from scratch its better just to be flexible in the definition of unit testing.

so, best practice i'd say would be to almost never stub filesystem calls.

of course we stub filesystem calls all over the place in tests, but there's a million things in our code that are horrible examples and need cleaning up...

jorhett: A link or two showing what you consider to be better implementations would be really helpful.

lamont-granquist: see the spec/data/cookbooks directory and any spec that reads from it.  CHEF_SPEC_DATA can be used to grab that path.  `File.expand_path(File.join(CHEF_SPEC_DATA, "cookbooks")),` is the cookbooks dir.  there's already invalid metadata in that cookbooks dir, and there's tests in spec/unit/cookbook_loader_spec.rb for example.   if you modify that directory to have a cookbook with bad metadata.json it might cause other specs to blow up, but that might actually be good in that it would give better "functional-ish" coverage to not only test the metadata class but also test that the cookbook_loader uses that code and appropriately blows up.

andy-dufour: Just as an update I've created some metadata.json files in the spec/data/metadata directory and kept some pressure on this over the last few days but I have a test that isn't throwing an exception, when it should. Need some time to figure out whats up there and I'll rebase & push another commit up.

lamont-granquist:
I'm working on rewriting all the specs to be less stubby in a branch right now

PR #4432, opened 2016-01-19
add get_rest etc calls to ServerAPI

thommay:
and note in rspec that we want these gone in 13
cc @chef/client-core

### DIFF
--- a/lib/chef/server_api.rb
+++ b/lib/chef/server_api.rb
@@ -48,6 +48,12 @@ class Chef
     # responses.
     use Chef::HTTP::ValidateContentLength
 
+    # for back compat with Chef::REST, expose `<verb>_rest` as an alias to `<verb>`
+    alias :get_rest :get
+    alias :delete_rest :delete
+    alias :post_rest :post
+    alias :put_rest :put
+
     # Makes an HTTP request to +path+ with the given +method+, +headers+, and
     # +data+ (if applicable). Does not apply any middleware, besides that
     # needed for Authentication.


coderanger:
:+1:

tylercloke:
:+1:

lamont-granquist:
:+1:

jaym:
:+1:

danielsdeleo:
What are these for? To help folks switch from `Chef::REST`? If we're planning to nuke them, probably should have a deprecation warning?

tylercloke:
@danielsdeleo Yeah, to make defaulting to ServerApi not a breaking change. Is there any reason to ever remove these methods? Seems pretty simple to just alias them and be backwards compatible.

jaym:
i dont think these were ever part of ServerAPI. The reason knife-opc broke is because the `chef_rest` method started returning ServerAPI instead of REST, and it was calling the `_rest` methods on it.

tylercloke:
Yeah, but we are removing `Chef::REST` long term in favor of ServerApi, so if we can easily make ServerApi backwards compatible with `Chef::REST`, why wouldn't we?

danielsdeleo:
I think the `_rest` methods are really ugly and it'd be cool if they didn't exist, but realistically they're not hurting anything so if it makes everyone's lives easier to keep them, that's ok by me. Contrarily, returning un-inflated objects from the API calls is where the "real breakage" (i.e., requires actual thought to fix) is, and doing a find/replace for, e.g., `s/get_rest/get/g` would be the easy part, so maybe we take the opportunity to clean up our API.

In any case, the thing I really want is for deprecated methods to emit deprecation warnings. I'm fine with either way that could be made to happen.

tylercloke:
I'm fine with listing them as deprecated if we actually do wanna get rid of these methods long term, but even if they are ugly, I feel like it is good to be backwards compatible unless there is a really compelling reason not to be.

This goes back to our lack of strong interfaces / policy around the interfaces in many places in the code, so I defer to you where you think that edge is here :)

thommay:
> Contrarily, returning un-inflated objects from the API calls is where the "real breakage" (i.e., requires actual thought to fix) is

This is all done now, I think? ServerAPI returns un-inflated objects, and all the classes on top inflate as part of load.

Regarding the `_rest` calls, I think Tyler's point about stability is a fair point, so I'll just remove the deprecation test.

danielsdeleo:
I'm totally fine with this outcome, just wanted to clarify my point. You can't really just gsub `s/Chef::REST/Chef::ServerAPI/g`, even if the `*_rest` methods are there, because the return types are different and not duck types of each other (except for cases where the server returns data without `json_class` stuff). Since users of `Chef::REST` have to deal with that, adding an additional gsub `s/get_rest/get/g`(and so on) isn't that big a deal.

Anyway, :+1: from me on un-deprecating.

PR #4433, opened 2016-01-19
add a comment on exception triggers

lamont-granquist:
this is almost always a real networking issue and there is no bug in
chef-client.


lamont-granquist:
@chef/client-core easy review

issues cut against this exception typically cannot be replicated and wind up being networking issues and chef-client is functioning exactly as designed.

coderanger:
:+1: Networking is hard let's go shopping.

PR #4434, opened 2016-01-19
adds EOFError message to handlers

lamont-granquist:
i think this still misses the biggest EOFError culprit which is the
node.save at the end, but we don't seem to have a node_save_failed
event to hook.


lamont-granquist:
:construction: needs tests.

lamont-granquist:
i looked at hooking run_failed up to an error mapper but decided that was a bridge too far for today..

danielsdeleo:
O_o :+1: 

Also, I'd be totally cool with adding a new event for the final node save if that makes things easier. I think I just skipped it because you normally shouldn't get a 401/403 there since you have to have made successful API calls before then.

lamont-granquist:
yeah, just adding a new event might be simpler, but i was worried that it would duplicate error messages with the later run_failed hook

thommay:
Nice :+1:

PR #4438, opened 2016-01-20
die die die if functional test fails to install chocolatey

mwrock:
This should help diagnose if build machines fail to install chocolatey


randomcamel:
:+1:

lamont-granquist:
:+1:

PR #4439, opened 2016-01-20
Sometimes chocolately doesn't appear on the path

tyler-ball:
Add choco to the path if it isn't already there.

\cc @mwrock @jaym


schisamo:
:+1:

mwrock:
So #4438 did not fix?

mwrock:
:+1: just occured to me that the choco installer adjusts the path in a child process so it does make sense here

lamont-granquist:
makes sense :+1:

PR #4447, opened 2016-01-21
use name for Chef::RunList::RunListItem

scalp42:


thommay:
The change looks reasonable but please can you look into the test explosions? Might be as simple as a rebase

scalp42:
Rebased.

scalp42:
```
13) Chef::Node::Attribute to_hash mutating strings should not mutate the attributes
      # this is a bug that should be fixed
      Failure/Error: expect(@attributes.default["foo"]).to eql({"bar"=>{"baz"=>"fizz"}})
      
        expected: {"bar"=>{"baz"=>"fizz"}}
             got: {"bar"=>{"baz"=>"fizzbuzz"}}
      
        (compared using eql?)
      
        Diff:
        @@ -1,2 +1,2 @@
        -"bar" => {"baz"=>"fizz"},
        +"bar" => {"baz"=>"fizzbuzz"},
        
      # ./spec/unit/node/attribute_spec.rb:515:in `block (3 levels) in <top (required)>'
```

thommay:
The one you pasted is expected (it's a pending item), but the run is actually blowing up rspec, and there's a heap of errors in the test output.

scalp42:
Gotcha but I cant find how it's related.

Feel free to close the PR.

thommay:
I'm gonna close this, not sure it can work the way it's written. We'll definitely whack that bug though.

PR #4449, opened 2016-01-22
Bump omnibus lock pins and fix the solaris mapfile.

ksubrama:


schisamo:
:+1: 

Can we please ensure this change is also replicated in the push-jobs-client and chefdk repos?

ksubrama:
Will do.

scotthain:
:+1:

PR #4451, opened 2016-01-22
Enable policies in local mode by default

danielsdeleo:
The primary goal here is to enable use of policyfile endpoints in local mode without requiring additional configuration on the user's part. Due to the various compatibility constraints around ChefFS (because of `knife download` being a backup/migration mechanism for many folks), this means we have to change the default behavior of Chef Zero to Chef 12 mode for Chef Client local mode operation.

We still default to single-org mode, which means that there's no way to serve user data via local mode with the default configuration, because Chef Zero will map a request for `/users` to `/organizations/chef/users`, but that can be fixed with additional config and is a suspect use case (e.g., `knife list users -z` and similar). This limitation is described in more detail in a comment in the tests.

Additionally, this adds `cookbook_artifacts` as a "sentinel directory" that we look for when finding the chef repo path. This makes local mode correctly find the root of a policyfile-based chef repo, which only has cookbook artifacts (without this change, if a parent directory of CWD has a cookbooks directory, that would be selected as the root, which is incorrect).

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -106,6 +106,9 @@ module ChefConfig
     # that upload or download files (such as knife upload, knife role from file,
     # etc.) work.
     default :chef_repo_path do
+      # TODO: this should also look at the cookbook_artifacts path, for the
+      # case when we have a repo from `chef export` that has cookbook artifacts
+      # and no cookbooks.
       if self.configuration[:cookbook_path]
         if self.configuration[:cookbook_path].kind_of?(String)
           File.expand_path("..", self.configuration[:cookbook_path])
@@ -306,6 +309,30 @@ module ChefConfig
     default :diff_output_threshold,   1000000
     default :local_mode, false
 
+    # Configures the mode of operation for ChefFS, which is applied to the
+    # ChefFS-based knife commands and chef-client's local mode. (ChefFS-based
+    # knife commands include: knife delete, knife deps, knife diff, knife down,
+    # knife edit, knife list, knife show, knife upload, and knife xargs.)
+    #
+    # Valid values are:
+    # * "static": ChefFS only manages objects that exist in a traditional Chef
+    #   Repo as of Chef 11.
+    # * "everything": ChefFS manages all object types that existed on the OSS
+    #   Chef 11 server.
+    # * "hosted_everything": ChefFS manages all object types as of the Chef 12
+    #   Server, including RBAC objects and Policyfile objects (new to Chef 12).
+    #
+    # TODO: add a test
+    default :repo_mode do
+      if local_mode
+        "hosted_everything"
+      elsif chef_server_url =~ /\/+organizations\/.+/
+        "hosted_everything"
+      else
+        "everything"
+      end
+    end
+
     default :pid_file, nil
 
     # Whether Chef Zero local mode should bind to a port. All internal requests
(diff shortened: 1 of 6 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -305,6 +308,30 @@ def self.path_accessible?(path)
...
+    # * "static": ChefFS only manages objects that exist in a traditional Chef
+    #   Repo as of Chef 11.
+    # * "everything": ChefFS manages all object types that existed on the OSS
+    #   Chef 11 server.
+    # * "hosted_everything": ChefFS manages all object types as of the Chef 12
+    #   Server, including RBAC objects and Policyfile objects (new to Chef 12).
+    #
+    # TODO: add a test
+    default :repo_mode do
+      if local_mode
```
jkeiser: I would make this `if local_mode && !chef_zero.osc_compat`, so that if someone *does* go into osc_compat mode, they don't get errors.

danielsdeleo: From my tests, most things actually work in that combination (RBAC stuff doesn't), but I think it makes sense that if you want Chef 11 mode we give it to you.

lib/chef/chef_fs/config.rb
```diff
@@ -47,6 +47,34 @@ class Config
...
+      #   everything in Chef Server 12 and later, including RBAC objects and
+      #   Policyfile objects.
+      #
+      # The "static" and "everything" modes are used for backup and
+      # upgrade/migration of older Chef Servers, so they should be considered
+      # frozen in time.
+
+      CHEF_11_OSS_STATIC_OBJECTS = %w{cookbooks cookbook_artifacts data_bags environments roles}.freeze
+      CHEF_11_OSS_DYNAMIC_OBJECTS = %w{clients nodes users}.freeze
+      RBAC_OBJECT_NAMES = %w{acls containers groups }
```
mwrock: missed `feeze`

mwrock:
:+1:

danielsdeleo:
@chef/client-core @jkeiser ready for review now. This gets us part way to implementing https://github.com/chef/chef-rfc/blob/master/rfc025-multitenant-chef-client-support.md though I left single-org-mode on.

chef-config/lib/chef-config/config.rb
```diff
@@ -320,6 +344,21 @@ def self.path_accessible?(path)
       default(:enabled) { ChefConfig::Config.local_mode }
       default :host, "localhost"
       default :port, 8889.upto(9999) # Will try ports from 8889-9999 until one works
+
+      # When set to a String, Chef Zero disables multitenant support.  This is
+      # what you want when using Chef Zero to serve a single Chef Repo. Setting
+      # this to `false` enables multi-tenant.
```
randomcamel: It can't be set to `true`?

danielsdeleo: Maybe it could work by accident, but the effect of this is to prepend requests with `/organizations/$single_org_option`, as is done here: https://github.com/chef/chef-zero/blob/57f2536f25b06930dbbf3560dfc6e2b0c644af03/lib/chef_zero/server.rb#L624 It's also passed in to some of the data store stuff, I didn't look deeper there but you could start here if you wanted to look into it further: https://github.com/chef/chef-zero/blob/57f2536f25b06930dbbf3560dfc6e2b0c644af03/lib/chef_zero/server.rb#L181-L187

chef-config/lib/chef-config/config.rb
```diff
@@ -113,6 +113,8 @@ def self.is_valid_url? uri
             File.expand_path("..", path)
           end
         end
+      elsif configuration[:cookbook_artifact_path]
```
jaym: should this be `self.configuration`?

danielsdeleo: I don't know what that is used elsewhere here, it's not a local variable in this scope so you don't need an explicit receiver. Works fine in my testing.

jaym:
:+1:

tyler-ball:
:+1:

jkeiser:
:+1:

PR #4455, opened 2016-01-25
Seeing a useradd failure in Nexus7 functional tests

tyler-ball:
This fixes the test failure we were seeing on Nexus7 bots.  Already went through Manhattan.  Because Nexus7 is a derivative of RHEL then this behavior is correct.

\cc @chef/maintainers


lamont-granquist:
yeah i argued with them not to create their own platform family...

anyway...

:+1:

danielsdeleo:
:+1:

jaym:
:+1:

PR #4458, opened 2016-01-25
Require declare_resource statically

jkeiser:


coderanger:
Explanation for the change?

jkeiser:
@coderanger unnecessary complexity. It appears to be a vestige of a circular dependency of yore (we're not 100% sure though even though we can't make it fail anymore--thus travis).

coderanger:
:+1: just wanted a reason for the record so if something happens again we might have a chance of finding this :)

stevendanna:
FWIW, I have found it *very* helpful to have those type of explanations in the body of the commit message.

lamont-granquist:
well that found the require loops in cheffish and chef-provisioning-aws...

jkeiser:
@lamont-granquist yep! Thanks. My summary of the loop:

1. `chef/resource`
   ```
   class Chef::Resource
     require "chef/provider"
     self.provider_base: Chef::Provider
   end
   ```
2. chef/provider
   ```
   class Chef::Provider
     require 'chef/provider/inline_resources'
     self.use_inline_resources
       include Chef::Provider::InlineResources
     end
end
```
3. chef/provider/inline_resources
   ```
   require 'chef/provider'

   module Chef::Provider
     module InlineResources
       require 'chef/dsl/recipe'
       include Chef::DSL::Recipe::FullDSL
     end
   end
4. chef/dsl/recipe
   ```
   module Chef::DSL::Recipe
     require "chef/dsl/declare_resource"
     include Chef::DSL::DeclareResource
     module FullDSL
       include Chef::DSL::Recipe
     end
   end
5. chef/dsl/declare_resource
   ```
   module Chef::DSL::DeclareResource
     require "chef/resource_builder"
     <runtime:build_resource()> Chef::ResourceBuilder.new()
   end
   ```
6. chef/resource_builder
   class Chef::ResourceBuilder
     require "chef/resource"
     <runtime:resource_class> Chef::Resource.new()
   end
   ```

@lamont-granquist, I'm curious: why is the `require` hidden behind `if defined?` Can we remove the last bit without destroying perf?

lamont-granquist:
at least at one point in time that `if defined?` was necessary for perf.

lamont-granquist:
and AFAIK that is just conventional idiom for doing lazy requires correctly, e.g.:

https://github.com/berkshelf/berkshelf/blob/51676f32cb6e8b241d0025dc5e7fffd226d76e2d/lib/berkshelf/downloader.rb#L78

lamont-granquist:
you could try lazying `require "chef/provider"` at the end of the resources.rb file.   if the magic to convert from the resource class to the provider class simply got removed in Chef 13 then the tight coupling between those classes would go away.

lamont-granquist:
@jkeiser minimum viable hackery to fix this i think

lamont-granquist:
@jkeiser do you still want to fix?

jkeiser:
Want to, but probably won't for a little bit.

b-dean:
If you do fix this, will you yank older chef releases that did this? Or will the compat_resource cookbook have to know that some chef versions might still do the lazy thing and others won't?

PR #4459, opened 2016-01-25
use require_paths and not path so bundler grabs all paths from a git reference

mwrock:
Using bundler to install from git via `gem 'chef', github: 'chef/chef'` only the last path is added to `$LOAD_PATH` when using `require_path` in gemspec. `require_paths` fixes this. This fixes knife-windows builds.


danielsdeleo:
:+1:

jaym:
:shipit:

lamont-granquist:
:+1:

PR #4460, opened 2016-01-26
Set range of system user/group id to max of 200

jaym:
According to https://abcofaix.wordpress.com/tag/usermod/, the value
can be up to 200. We were seeing tests on our aix testers fail with

```
  1) Chef::Provider::User::Useradd action :create when the user does not exist beforehand when a system user is specified ensures the user has the properties of a system user
     Failure/Error: expect(pw_entry.uid.to_i).to be < uid_min.to_i

       expected: < 100
            got:   101
     # ./spec/functional/resource/user/useradd_spec.rb:306:in `block (5 levels) in <top (required)>'
```

cc @chef/client-aix @chef/client-core

### DIFF
--- a/spec/functional/resource/user/useradd_spec.rb
+++ b/spec/functional/resource/user/useradd_spec.rb
@@ -287,9 +287,9 @@ describe Chef::Provider::User::Useradd, metadata do
         let(:uid_min) do
           case ohai[:platform]
           when "aix"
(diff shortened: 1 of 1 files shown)


lamont-granquist:
:+1:

danielsdeleo:
:+1:

btm:
:+1:

btm:
this won't run on any of the travis/appveyor testers so I'll mash it.

PR #4461, opened 2016-01-26
debian-6 is EOL next month

lamont-granquist:
t1.micros are also beyond terrible for performance.


lamont-granquist:
@jkeiser

tas50:
We should remember to pull it from https://github.com/chef/chef-rfc/blob/master/rfc021-platform-support-policy.md in a month

PR #4463, opened 2016-01-26
Allow recent omnibus-software ruby version convention.

ksubrama:
This allows master to build at https://github.com/chef/omnibus-software/pull/556

### DIFF
--- a/omnibus/config/projects/chef-fips.rb
+++ b/omnibus/config/projects/chef-fips.rb
@@ -34,9 +34,10 @@ else
   install_dir "#{default_root}/#{name}"
 end
 
+override :'ruby-windows', version: "2.0.0-p647"
+
 # Global FIPS override flag.
 override :fips, enabled: true
-override :ruby, version: "2.1.7"
 
 override :chef, version: "local_source"
 override :ohai, version: "master"
--- a/omnibus/config/projects/chef-fips.rb
+++ b/omnibus/config/projects/chef-fips.rb
@@ -36,7 +36,7 @@ end
 
 # Global FIPS override flag.
 override :fips, enabled: true
-override :'ruby-windows', version: "2.0.0-p647"
+override :ruby, version: "2.1.7"
 
 override :chef, version: "local_source"
 override :ohai, version: "master"
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -33,13 +33,16 @@ else
 end
 
 override :bundler,        version: "1.10.6"
-override :ruby,           version: "2.1.6"
 
-override :'ruby-windows', version: "2.0.0-p645"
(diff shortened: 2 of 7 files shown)


jaym:
can you explain the cbgb change. Having things silently eat the exit code could suck if somebody is using that in any automated way.

omnibus/config/projects/chef.rb
```diff
@@ -33,13 +33,16 @@
...
 override :bundler,        version: "1.10.6"
-override :ruby,           version: "2.1.6"
 
-override :'ruby-windows', version: "2.0.0-p645"
-# Leave dev-kit pinned to 4.5 because 4.7 is 20MB larger and we don't want
-# to unnecessarily make the client any fatter.
-if windows_arch_i386?
-  override :'ruby-windows-devkit', version: "4.5.2-20111229-1559"
+if windows?
+  override :'ruby-windows', version: "2.0.0-p645"
```
jaym: The fact that if people pull in the new omnibus-software, they have to do something like this is a little scary. @schisamo do you know if anybody besides chef uses omnibus for their Windows projects?

omnibus/config/projects/chef-fips.rb
```diff
@@ -36,7 +36,7 @@
 
 # Global FIPS override flag.
 override :fips, enabled: true
-override :'ruby-windows', version: "2.0.0-p647"
+override :ruby, version: "2.1.7"
```
jaym: awesome, is this in our cache?

ksubrama: The source for it is :) - we've had in our cache from non-windows builds for a long time.

ksubrama:
If you look at the maintainers task, it performs the same sort of check.  The assumption is that if someone is setting up a task that needs cbgb or updating the maintainers file or whatever, they wouldn't bundle install --without maintainers.  On the other hand, when vendoring it in chef-client, we don't need any of those libraries.

Hence the soft-recovery if tomlrb isn't present.

omnibus/config/projects/chef.rb
btm: do what? override ruby-windows? weren't we doing that already? --^

btm:
this is significantly more readable via https://github.com/chef/chef/pull/4463/files?w=1

omnibus/config/projects/chef.rb
jaym: override 'ruby-windows' and not 'ruby' on windows

jaym:
:+1: if this passes the chef-fips pipeline. I think you mentioned some concerns there, which is why I bring it up.

ksubrama:
Lamont and others trying to work on omnibus-software at master need this fix.  I'm leaving chef-fips building from yakyak for now and bump that separately.

lamont-granquist:
*achm* merge conflict with my branch?

PR #4465, opened 2016-01-26
Change shell for locked user.

artursitarski:
Disallow locked user to log in setting /bin/false as a shell value (without it user is still able to log in using ssh key + authorized keys, even with locked password). 
Reset shell for unlocked user.

### DIFF
--- a/lib/chef/provider/user/useradd.rb
+++ b/lib/chef/provider/user/useradd.rb
@@ -93,13 +93,11 @@ class Chef
         end
 
         def lock_user
-          shell_out!("usermod", "-L", new_resource.username)
-          shell_out!("chsh", "-s", "/bin/false", new_resource.username)
+          shell_out!("usermod", "-L", "-s", "/bin/false", new_resource.username)
         end
 
         def unlock_user
-          shell_out!("usermod", "-U", new_resource.username)
-          shell_out!("chsh", "-s", new_resource.shell, new_resource.username)
+          shell_out!("usermod", "-U", "-s", new_resource.shell, new_resource.username)
         end
 
         def compile_command(base_command)
--- a/lib/chef/provider/user/useradd.rb
+++ b/lib/chef/provider/user/useradd.rb
@@ -94,10 +94,12 @@ class Chef
 
         def lock_user
           shell_out!("usermod", "-L", new_resource.username)
+          shell_out!("chsh", "-s", "/bin/false", new_resource.username)
         end
 
         def unlock_user
           shell_out!("usermod", "-U", new_resource.username)
+          shell_out!("chsh", "-s", new_resource.shell, new_resource.username)
         end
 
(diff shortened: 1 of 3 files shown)


artursitarski:
Travis says:
>Bundler could not find compatible versions for gem "poise-profiler":
>  In Gemfile:
>    poise-boiler was resolved to 1.5.2.pre, which depends on
>      poise-profiler (~> 1.0.pre)

Can someone take a look?

coderanger:
@artursitarski Being fixed elsewhere.

artursitarski:
@coderanger thanks a lot, now it's ok. I see that continuous-integration/appveyor checks got stuck in some weird state. Could you poke them?

artursitarski:
@chef/client-maintainers

artursitarski:
Ping

thommay:
Think this looks fine. :+1:

lamont-granquist:
only issue i've got with this is that we need to validate that it works on ubuntu, rhel and solaris -- all of which have slightly differing useradd implementations...

lamont-granquist:
other than that :+1: 

although it sounds like thom is all over the O/S compat matrix so sounds like this is good 2 go

smurawski:
:+1:

thommay:
Yeah, I have validated-by-man-page, and I think that's good enough.

legal90:
@artursitarski @lamont-granquist @thommay 
:warning: FYI: this PR potentially breaks the backward compatibility.
In our case on RHEL 6 we've managed а user and performed `lock` action expecting that it will be literally locked (`usermode -L`), but nothing more:

```ruby
user 'beaver' do
  # ...
  action :lock
end
```

But now this user got a shell set to `/bin/false`, which causes failures to start the beaver daemon. Its init script contains the following line: `su - $user -c kill -TERM $pid`, so now it's failed because of disabled shell.

We strongly suggest not to break the behavior of user account lock, at least for `useradd` provider, and left only `-L` option here. 
In case if operator wants this user account to be restricted to login via SSH private key, he should manage `AllowUsers` option of ssh deamon and/or just set shell explicitly:

```ruby
user 'my_user' do
  # ...
  shell `/bin/false`
  action :lock
end
```

Kasen:
@artursitarski @lamont-granquist @thommay Guys, usermod man page tales us:
```
       -L, --lock
           Lock a user's password. This puts a '!' in front of the encrypted password, effectively disabling the password. You can't use this option with -p or -U.

           Note: if you wish to lock the account (not only access with a password), you should also set the EXPIRE_DATE to 1.
```

So if `user` resource executes only `usermod -L` is most obvious thing.
If you have trouble with ability to login into server by ssh, you must resolve it by configuring sshd. Also you may specify explicitly `/bin/false` as shell in your `user` resources.

I think what this PR will break workflow not only for us.

PR #4466, opened 2016-01-27
Fix bug in sh script that runs test in jenkins

jaym:
Some testers fail with

```
ci/verify-chef.sh: test: ] missing`
```

This regression was introduced by
https://github.com/chef/chef/commit/2bcb7d06dcb49c953b9e666b24c8bd3cb56da2c5

This patch should fix it, though I can't be 100% sure as the pipeline seems to pick up the ci folder from master.

### DIFF
--- a/ci/verify-chef.sh
+++ b/ci/verify-chef.sh
@@ -86,9 +86,10 @@ if [ ! -f "Gemfile.lock" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jaym:
the test failure is unrelated, coderanger is working on fixing

btm:
:+1:

thommay:
:+1:

PR #4469, opened 2016-01-27
Force config file setting to expanded path in chef-client

danielsdeleo:
I attempted to fix this previously in https://github.com/chef/chef/pull/3963 but it regressed; I wasn't able to track down the commit that broke it, but in any case that fix wasn't very stable because quite a few config paths are evaluated relative to the config file path, which is expanded relative to cwd, and chef-client chdirs to root right after the config file is read.

@chef/client-core

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -90,6 +90,13 @@ class Chef
     # Parse the config file
     def load_config_file
       config_fetcher = Chef::ConfigFetcher.new(config[:config_file])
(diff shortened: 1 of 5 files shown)


danielsdeleo:
Travis issue is incompatible gems in the tests that use poise, Noah is on the case there.

tyler-ball:
:+1:

thommay:
:+1:

PR #4472, opened 2016-01-27
Fix inverted filter logic for not_fips rspec filter

jaym:
The logic here is inverted, and I think the `not_wpar` one is as well. We'll fix that in a different PR.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -169,7 +169,7 @@ RSpec.configure do |config|
   config.filter_run_excluding :aes_256_gcm_only => true unless aes_256_gcm?
   config.filter_run_excluding :broken => true
   config.filter_run_excluding :not_wpar => true unless wpar?
-  config.filter_run_excluding :not_fips => true if fips?
+  config.filter_run_excluding :not_supported_under_fips => true if fips?
 
   running_platform_arch = `uname -m`.strip unless windows?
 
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -169,7 +169,7 @@ module Shell
   config.filter_run_excluding :aes_256_gcm_only => true unless aes_256_gcm?
   config.filter_run_excluding :broken => true
   config.filter_run_excluding :not_wpar => true unless wpar?
-  config.filter_run_excluding :not_fips => true unless fips?
+  config.filter_run_excluding :not_fips => true if fips?
```
btm: what is not_fips exactly? a test we want to exclude if we are not running under fips? it might be easier to make sure this isn't backwards if we named it better.

```
config.filter_run_excluding :with_fips_only => true unless fips?
config.filter_run_excluding :fips_enabled => true unless fips?
config.filter_run_excluding :fips_disabled => true if fips?
config.filter_run_excluding :without_fips_only => true if fips?
config.filter_run_excluding :not_supported_under_fips => true if fips?
```

jaym:
@btm updated

btm:
:+1: on green

btm:
travis failure was poise.

jaym:
poise is known to be broken atm. coderanger is wronging on fixing it

lamont-granquist:
:+1:

smurawski:
:+1:

mwrock:
:+1:

PR #4474, opened 2016-01-27
Bump version to 12.7.0, and make "rake version" use just one VERSION file

jkeiser:
@chef/client-maintainers @danielsdeleo @btm 

We'll generate a release build with this, but will probably not promote to stable. When we generate another release build, we'll increment to 12.7.1. The final promoted 12.7 build will be a higher number and so won't conflict with the current-channel 12.7.0 build.


btm:
fyi for those not following along, this is a manual step towards the new versioning that doesn't use RC builds as specified in [rfc047](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#versioning).

danielsdeleo:
The special case in the rake task is better than having two VERSION files IMO, :+1:

lamont-granquist:
:+1:

PR #4476, opened 2016-01-27
bundle install without maintenance and others

lamont-granquist:
if we don't test this here, then when the jenkins ci cluster runs it
we may error in the Rakefile when it tries to require a gem which is
not in the bundle that is installed there --without several gems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,13 @@ matrix:
   include:
   - rvm: 2.1
     sudo: true
+    bundler_args: --without server docgen maintenance
   - rvm: 2.2
     sudo: true
+    bundler_args: --without server docgen maintenance
   - rvm: rbx
     sudo: true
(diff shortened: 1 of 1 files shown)


jaym:
:+1: for the tests, though i'm not sure why this actually broke

lamont-granquist:
```
- tasks/cbgb.rb was added without LoadError protection so the Rakefile will error hard if toml.rb is not available.
- `--without maintenance` was added by @ksubrama to omnibus-software so that toml.rb is not installed in omnibus builds any more.
```

ksubrama:
:+1:

lamont-granquist:
well that's a big odd, the first travis tests kicked off after i pushed the second commit so it looks like both runs are testing the end result and i don't see the failures that i expected...

must be a result of travis testing the merge commit and evaluating that at the time that the job first gets a worker on travis...

PR #4477, opened 2016-01-27
Remove poise from the test matrix

jaym:
Poise is currently having it's tests refactored and thus
not passing tests. We can revert this commit once that's
all fixed up.

cc @coderanger


btm:
:+1: if @coderanger is :+1:

coderanger:
:+1: will be another 1-2 days before I get everything pieced back together. Sorry for the noise!

under construction

under construction under construction under construction under construction under construction under construction under construction under construction under construction under construction

coderanger:
Going to revert this, poise's tests should be good to go.

PR #4478, opened 2016-01-27
RFC-060 gem metadata MVP

lamont-granquist:
Implements RFC-60 'gem' keyword in metadata.rb

### DIFF
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -1,7 +1,7 @@
 #--
 # Author:: Christopher Walters (<cw@opscode.com>)
 # Author:: Tim Hinderliter (<tim@opscode.com>)
-# Copyright:: Copyright (c) 2010-2015 Chef Software, Inc.
+# Copyright:: Copyright (c) 2010-2016 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -1,0 +1,60 @@
+#--
+# Author:: Christopher Walters (<cw@opscode.com>)
+# Author:: Tim Hinderliter (<tim@opscode.com>)
+# Copyright:: Copyright (c) 2010-2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'tmpdir'
+require 'bundler/inline'
+
+class Chef
+  class Cookbook
+    class GemInstaller
+      attr_accessor :cookbook_collection
+
+      def initialize(cookbook_collection)
+        @cookbook_collection = cookbook_collection
+      end
+
+      def install
+        cookbook_gems = []
+
+        cookbook_collection.each do |cookbook_name, cookbook_version|
+          cookbook_gems += cookbook_version.metadata.gems
+        end
+
+        return if cookbook_gems.empty?
+
+        gemfile(true) do
+          source 'https://rubygems.org'
+          cookbook_gems.each do |args|
+            gem *args
+          end
+        end
+      end
+
+      class ChefBundlerUI < Bundler::UI::Silent
+        def confirm(msg, newline = nil)
+          Chef::Log.warn("CONFIRM: #{msg}")
+        end
+
+        def error(msg, newline = nil)
+          Chef::Log.warn("ERROR: #{msg}")
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


ranjib:
neat :-)

lamont-granquist:
https://github.com/bundler/bundler/pull/4245 would be useful to leverage bundler/inline while still cleaning up the output format.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -0,0 +1,112 @@
...
+        cookbook_collection.each do |cookbook_name, cookbook_version|
+          cookbook_gems += cookbook_version.metadata.gems
+        end
+
+        events.cookbook_gem_start(cookbook_gems)
+
+        unless cookbook_gems.empty?
+          begin
+            inline_gemfile do
+              source "https://rubygems.org"
```
coderanger: This should be an option in the Chef config for people behind a firewall.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -0,0 +1,112 @@
...
+          Chef::Log.warn(msg)
+        end
+      end
+
+      private
+
+      def inline_gemfile(&block)
+        # requires https://github.com/bundler/bundler/pull/4245
+        gemfile(true, ui: ChefBundlerUI.new(events), &block)
+      rescue ArgumentError  # Method#arity doesn't inspect optional arguments, so we rescue
```
coderanger: It does, they should show up as negative numbers or something.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -0,0 +1,112 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "tmpdir"
+begin
+  require "bundler/inline"
+rescue LoadError
+  raise RuntimeError, "The RFC060 metadata gem feature requires bundler 1.10.0 or greater."
```
coderanger: With the runtime dependency I think we can drop this.

lib/chef/cookbook/gem_installer.rb
lamont-granquist: yep, not done yet.

lib/chef/cookbook/gem_installer.rb
lamont-granquist: Shows up as -1 for both cases.  Its the number of non-optional arguments minus one, so adding more optional arguments doesn't change the arity.

lib/chef/cookbook/gem_installer.rb
lamont-granquist: Yeah, that's most likely overly paranoid...

lamont-granquist:
@chef/client-core this is probably good for review now.

travis will be busted right now due to github-changelog-generator.

bundler hasn't yet done a release, but it'll work fine without the bundler release, it just won't have nice output.

halite still has the busted spec due to needing to update bundler (test failure for halite, everything works fine) so it is moved to excluded for now.

ranjib:
:+1:

thommay:
Awesome! :+1:

lib/chef/cookbook/gem_installer.rb
```diff
@@ -0,0 +1,118 @@
...
+      private
+
+      # Helper to handle older bundler versions that do not support injecting the UI
+      # object.  On older bundler versions, we work, but you get no output other than
+      # on STDOUT.
+      #
+      def inline_gemfile(&block)
+        # requires https://github.com/bundler/bundler/pull/4245
+        gemfile(true, ui: ChefBundlerUI.new(events), &block)
+      rescue ArgumentError # Method#arity doesn't inspect optional arguments, so we rescue
```
danielsdeleo: `Method#arity` returns a negative number if there are optional arguments:

```

danielsdeleo: Also, how old is this bundler? What's the cost of not supporting it?

lamont-granquist: The bundler version that supports better logging hasn't been released yet, its only in master.  They've got a pre release out on rubygems so I'm suspecting they'll do a minor release within a month or so if their historical pattern holds up.

Since the first argument to gemfile is optional, the new optional argument doesn't change the arity and it returns -1 in both cases.

I'll look into `#parameters`, looks like that's supported in >= 2.0 so we could use it without having to fallback...

danielsdeleo:
:+1: Side note, would like to get a workaround for the changelog generator thing so we can trust travis. We did this for ChefDK: https://github.com/chef/chef-dk/pull/704/files

coderanger:
@lamont-granquist We still need to do the Bundle.require bit secondarily, right?

lamont-granquist:
you mean like `gem "foo", require: "foo.rb"`?  yeah, it does no automatic chef-tie-ing...

coderanger:
:+1: It's a good thing for a second round PR, just shouldn't forget about it :)

lamont-granquist:
I thought we decided not to cheftie automagically, though?

You've got the whole Bundler `gem` API available, so `required` does work.  Even arguments like `git` and `github` should work provided the images that are used are prebaked with git.

coderanger:
We decided to not do any weird file-scanning like my code does in favor of a single `Bundler.require` call after installation. Without that, no code from the installed gems is actually loaded.

lamont-granquist:
Ah, so Bundler#require is here and is basically 

https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L101-L103

I believe we're getting the same Bundler::Require.setup.require call sequence through here:

https://github.com/bundler/bundler/blob/master/lib/bundler/inline.rb#L59-L60

We definitely get the gems installed, and I (manually) tested that requiring worked...

PR #4481, opened 2016-01-28
Allow use of command line fips switch for knife

jaym:
This somewhat worked before. However, it was just for knife bootstrap.
It also didn't support --no-fips in the case where it was in your
knife.rb and you decided that you didn't want to use fips for
that one call.

The assumption here is fips mode you run knife with is the fips
mode the node will get. This has the nice property that validatorless
bootstraps will talk to the chef server in fips mode if the node
is requested to be in fips mode.

### DIFF
--- a/spec/unit/application/knife_spec.rb
+++ b/spec/unit/application/knife_spec.rb
@@ -93,7 +93,7 @@ describe Chef::Application::Knife do
       it "does not initialize fips mode when no flags are passed" do
         with_argv(*%w{noop knife command}) do
           expect(@knife).to receive(:exit).with(0)
-          expect(OpenSSL).not_to receive(:'fips_mode=')
+          expect(Chef::Config).not_to receive(:enable_fips_mode)
           @knife.run
           expect(Chef::Config[:fips]).to eq(false)
         end
@@ -102,7 +102,7 @@ describe Chef::Application::Knife do
       it "overwrites the Chef::Config value when passed --fips" do
         with_argv(*%w{noop knife command --fips}) do
           expect(@knife).to receive(:exit).with(0)
-          expect(OpenSSL).to receive(:'fips_mode=').with(true)
+          expect(Chef::Config).to receive(:enable_fips_mode)
           @knife.run
           expect(Chef::Config[:fips]).to eq(true)
         end
@@ -117,7 +117,7 @@ describe Chef::Application::Knife do
       it "initializes fips mode when passed --fips" do
         with_argv(*%w{noop knife command --fips}) do
           expect(@knife).to receive(:exit).with(0)
-          expect(OpenSSL).to receive(:'fips_mode=').with(true)
+          expect(Chef::Config).to receive(:enable_fips_mode)
           @knife.run
           expect(Chef::Config[:fips]).to eq(true)
         end
@@ -126,7 +126,7 @@ describe Chef::Application::Knife do
       it "overwrites the Chef::Config value when passed --no-fips" do
         with_argv(*%w{noop knife command --no-fips}) do
           expect(@knife).to receive(:exit).with(0)
-          expect(OpenSSL).not_to receive(:'fips_mode=')
+          expect(Chef::Config).not_to receive(:enable_fips_mode)
           @knife.run
           expect(Chef::Config[:fips]).to eq(false)
         end
--- a/spec/unit/application/knife_spec.rb
+++ b/spec/unit/application/knife_spec.rb
@@ -82,6 +82,58 @@ describe Chef::Application::Knife do
     end
   end
 
+  context "when given fips flags" do
+    context "when Chef::Config[:fips]=false" do
+      before do
+        # This is required because the chef-fips pipeline does
+        # has a default value of true for fips
+        Chef::Config[:fips] = false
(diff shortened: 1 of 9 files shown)


spec/unit/knife/core/bootstrap_context_spec.rb
```diff
@@ -20,6 +20,10 @@
 require "chef/knife/core/bootstrap_context"
 
 describe Chef::Knife::Core::BootstrapContext do
+  before do
+    Chef::Config[:fips] = false
```
danielsdeleo: If this is the default, you shouldn't have to do this. We reset `Chef::Config` before every test via a thing in `spec_helper`

spec/unit/application/knife_spec.rb
```diff
@@ -82,6 +82,60 @@ def run
     end
   end
 
+  context "when given fips flags" do
+    context "when Chef::Config[:fips]=false" do
+      before do
+        Chef::Config[:fips] = false
```
danielsdeleo: Same as comment below.

spec/unit/knife/core/bootstrap_context_spec.rb
jaym: interesting, did not know that, let me try

jaym: @danielsdeleo  Ok, so it's coming back to me: This was required to get tests to pass in both pipelines. The environment CHEF_FIPS is set to 1 for the chef-fips pipeline, which causes the default value to change to true for that pipeline.

danielsdeleo: That makes sense. Probably worth adding a comment about that in the tests since it behaves differently (or at least appears to) than other config settings.

ksubrama:
What happens if the user bootstraps but gives a client version that's too old?  I don't remember what happens if extra flags are passed that chef-client doesn't understand.  If it yells - that's fine.  Otherwise, we might want to check for a version (or make ourselves a bug/card to check for a version before GA).

danielsdeleo:
```

**Comment by danielsdeleo** [comment]
```
chef-client --fips
/Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/mixlib-cli-1.5.0/lib/mixlib/cli.rb:191:in `parse_options': invalid option: --fips (OptionParser::InvalidOption)
	from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/chef-12.6.0/lib/chef/application.rb:84:in `configure_chef'
	from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/chef-12.6.0/lib/chef/application.rb:47:in `reconfigure'
	from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/chef-12.6.0/lib/chef/application/client.rb:284:in `reconfigure'
	from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/chef-12.6.0/lib/chef/application.rb:56:in `run'
	from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/chef-12.6.0/bin/chef-client:26:in `<top (required)>'
	from /Users/ddeleo/.rbenv/versions/2.1.5/bin/chef-client:23:in `load'
	from /Users/ddeleo/.rbenv/versions/2.1.5/bin/chef-client:23:in `<main>'
```

We should make this not dump a stack trace, but for the sake of the discussion here, it meets your needs

jaym:
eh, bootstrap puts it in the client.rb. So it'll probably happily continue along. That being said, if we want to do this, we should make a story card for it as we don't really do this for any other switches so it'd require a good deal of plumbing.

jaym:
I take the plumbing thing back, but we should still talk about this as a new feature

danielsdeleo:
Currently mixlib-config supports a strict mode where unrecognized config settings are ignored and this is the default mode for the top level config context. As nice as it would be to turn that on, lots of folks are in a habit of putting user-defined stuff in there and we'd break them. You could work around that by making a new config context, so that you do something like `fips.enabled = true` and that would blow up with a "no method on nil"error on older chefs. That said, we generally just document that you need version X of the software to get new feature Y and call it a day. The only place this really causes issues is with folks who don't upgrade chef-client but want to use community cookbooks.

ksubrama:
The silent client.rb behavior was what I was worried about.  Since FIPS is more about compliance than actually helpful features, I wanted to make sure we weren't silently succeeding the knife bootstrap if someone pointed it to an older build or a custom build path which didn't support that flag.

I'm cool with tracking that separately - it smells of a product/feature discussion.
:+1: to this otherwise.

danielsdeleo:
Yeah, it might be possibly easier and better from a user POV to have ohai check the openssl config and then you could have a recipe to abort the chef run if (e.g.) `node[:openssl_config][:fips]` isn't set as desired.

jaym:
The feature to check the chef version has been added to the backlog

thommay:
:+1:

PR #4491, opened 2016-01-29
Fix dpkg interruption

rettigs:
If dpkg gets interrupted while installing a package (such as if it gets killed for exceeding the timeout when installing a large package like `texlive-full`), subsequent runs of the `apt_package` resource will fail with an error like the following:
```
Mixlib::ShellOut::ShellCommandFailed: apt_package[mplayer] (workstation::packages line 21) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '100'
---- Begin output of apt-get -q -y install mplayer ----
STDOUT:
STDERR: E: dpkg was interrupted, you must manually run 'dpkg --configure -a' to correct the problem.
---- End output of apt-get -q -y install mplayer ----
Ran apt-get -q -y install mplayer returned 100
```
This PR should cause Chef to detect when this error happens and automatically run `dpkg --configure -a`, rather than continually failing Chef runs until `dpkg --configure -a` is manually run on the node.

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -176,7 +176,18 @@ class Chef
         # interactive prompts. Command is run with default localization rather
         # than forcing locale to "C", so command output may not be stable.
         def run_noninteractive(command)
-          shell_out_with_timeout!(command, :env => { "DEBIAN_FRONTEND" => "noninteractive", "LC_ALL" => nil })
+          cmd = shell_out_with_timeout!(command, :env => { "DEBIAN_FRONTEND" => "noninteractive", "LC_ALL" => nil }, :returns => [0, 100] )
+          if cmd.stderr == "E: dpkg was interrupted, you must manually run 'dpkg --configure -a' to correct the problem. \n"
+            # If the issue was that dpkg was interrupted,
+            # run 'dpkg --configure -a' to fix it.
+            Chef::Log.warn("dpkg was interrupted on the last run; running 'dpkg --configure -a' to correct the problem.")
+            shell_out_with_timeout!('dpkg --configure -a', :env => { "DEBIAN_FRONTEND" => "noninteractive", "LC_ALL" => nil })
+          else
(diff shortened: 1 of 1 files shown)


coderanger:
How safe do we think it is to run this automatically?

thommay:
My feeling is that this ought to be an exceptional situation that you ought not to fix automatically. So :-1: on the face of it.

coderanger:
Also this would run all failed apt commands twice, which is a huge non-starter.

lamont-granquist:
I'm pretty :-1: on hacking up the dpkg provider to do this, would prefer a more cookbook-driven solution in the apt-cookbook to work around it.

rettigs:
@thommay What is the potential for automatically fixing this to go wrong?  I'm not familiar enough with this kind of situation to know for sure.

@coderanger That was just the simplest way I could see to write a working PR.  I agree that it's not ideal, but neither is trying to make everything perfect before I know whether the PR has a chance to be merged.

@lamont-granquist If this functionality were to be implemented in the apt cookbook, how would you recommend integrating it?  This PR runs the fix when a package install fails, but as far as I know, the apt cookbook doesn't actually deal with the installation of packages.

lamont-granquist:
Can we test-and-repair this outside of apt installation?

Lets say we started solving this problem with this at the beginning of every run:

```ruby
execute "dpkg --configure -a"
```

What guard would you slap on there to make the execute resource idempotent?

And to be really clear I don't think you should solve the problem of repairing the existing chef-client run once a dpkg install hits this error message.  I think its fine and entirely appropriate to error back to the user.  Then the next run can try to repair the condition.

If the chef-client run asserts that this is cleaned up to start with, then any error can only come from some kind of bug in the current chef-client run itself (which the user should know about) or through interaction between the running chef-client and another process on the server.  Both of those conditions I think really should fail back to the user, and not hide the issue with magic.

thommay:
On further consideration we won't be taking this PR; it's an exceptional situation that the admin should rectify manually. Thanks for your submission!

PR #4482, opened 2016-01-28
Support policy ACLs in knife upload/download

jkeiser:

PR #4492, opened 2016-01-29
Allow all Chef Zero 4.x after 4.4

danielsdeleo:
We require 4.4+ to support policyfile APIs with ChefFS and local mode,
but we don't want to be pessimistic about minor versions, only major.

It looks like the minor version pessimism was accidentally introduced
here:
https://github.com/chef/chef/commit/6e4d70826d8ee6977348bd071d0ce543686b89f6

Now that Chef Zero 4.5 is released, this is breaking stuff where we load chef and chef zero from master.

@chef/client-core


danielsdeleo:
Could be fixed by https://github.com/chef/chef/pull/4482 instead...

lamont-granquist:
retroactive :+1:

PR #4493, opened 2016-01-30
Fix defaults on property_type

jkeiser:
@dan @lamont-granquist @someara `property_type(is: [:a, :b], default: :c)` was broken because ParamsValidate requires its input to have a name and property types don't have one.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -129,7 +129,7 @@ class Chef
     end
 
     def to_s
-      "#{name}#{declared_in ? " of resource #{declared_in.resource_name}" : ""}"
+      "#{name || "<property type>"}#{declared_in ? " of resource #{declared_in.resource_name}" : ""}"
     end
 
     #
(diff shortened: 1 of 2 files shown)


jkeiser:
@chef/client-maintainers

lib/chef/property.rb
```diff
@@ -463,7 +463,12 @@ def coerce(resource, value)
     def validate(resource, value)
       # If we have no default value, `nil` is never coerced or validated
       unless value.nil? && !has_default?
-        (resource || Chef::Mixin::ParamsValidate).validate({ name => value }, { name => validation_options })
+        if resource
+          resource.validate({ name => value }, { name => validation_options })
+        else
+          name = self.name || :property_type
+          (resource || Chef::Mixin::ParamsValidate).validate({ name => value }, { name => validation_options })
```
stevendanna: Can we remove the first half of this `||`?  Isn't that captured by the if/else we are in?

stevendanna:
If you get a chance, it would be nice if your commit message included the explanation that you've included in your pull request message.

PR #4497, opened 2016-02-01
Chocolatey: Raise a helpful MissingLibrary exception if it's not there

randomcamel:


lib/chef/provider/package/chocolatey.rb
```diff
@@ -136,14 +136,27 @@ def check_resource_semantics!
         #
         # @return [String] full path of choco.exe
         def choco_exe
-          @choco_exe ||=
+          @choco_exe ||= begin
+            path = choco_install_path
+
+            # GetEnvironmentVariable returns "" on failure.
+            if path.to_s.empty?
+              raise Chef::Exceptions::MissingLibrary.new("Could not find Chocolatey. Is it installed?")
```
danielsdeleo: How about adding some info about this powershell command we run to find it, in case the user actually has chocolatey but there is something else that's broken?

mwrock: The best advise here is to tell them to include the default recipe of the Chocolatey cookbook which will do the install.

danielsdeleo: I'm thinking of the kind of case where a user knows just enough to be dangerous and changes some setting that breaks it. If we tell them to run `[System.Environment]::GetEnvironmentVariable('ChocolateyInstall', 'MACHINE')` and it should produce output like `x` then they can repro outside of Chef and figure out what they broke.

mwrock: oh I didn't read @danielsdeleo 's first note very carefully. So first I would add a note about running the [chocolatey cookbook](https://github.com/chocolatey/chocolatey-cookbook) to install it if we cant find it. Its officially supported under the chocolatey org and I'm its primary maintainer. 

Perhaps logging to debug that the Machine scoped `ChocolateyInstall` variable either was not present or was found to be X could suffice for troubleshooting purposes. If the user changes this variable after choco is installed and choco.exe is no longer found under that path, they have basically broken all chocolatey functionality. The more possibly risky scenario is that they install chocolatey and it is there, but the environment variable has not been propperly broadcast and is not seen by this resource. However the chocolatey installer should do that.

danielsdeleo: I think I like all of the above, e.g., "Could not locate your Chocolatey install. To install chocolatey, we recommend the chocolatey cookbook (URL?). If Chocolatey is installed, ensure that the 'ChocolateyInstall' environment variable is correctly set. You can verify this with the powershell command `[System.Environment]::GetEnvironmentVariable('ChocolateyInstall', 'MACHINE')`".

mwrock: :+1:

lamont-granquist:
the exception should probably get raised up in define_resource_requirements to keep `def choco_exe` simpler.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -136,14 +136,27 @@ def check_resource_semantics!
...
               "bin",
               "choco.exe",
-          )
+              )
+          end
+        end
+
+        # lets us mock out an incorrect value for testing.
+        def choco_install_path
+          powershell_out!(
```
lamont-granquist: should `@choco_install_path ||=` this so you don't have to create local variables to protect against double-invocation elsewhere...

danielsdeleo:
nice. :+1:

lamont-granquist:
:+1:

PR #4500, opened 2016-02-01
explicitly specify user permissions during msi install of chef directory

mwrock:
windows client SKUs default to giving more permissions than necessary

### DIFF
--- a/omnibus/resources/chef/msi/source.wxs.erb
+++ b/omnibus/resources/chef/msi/source.wxs.erb
@@ -18,6 +18,14 @@
     <Package InstallerVersion="200" InstallPrivileges="elevated"
              Compressed="yes" InstallScope="perMachine" />
 
+    <!--
+      Create property references for the well known SIDs of the 
+      accounts we want to restrict for the project location folder
+    -->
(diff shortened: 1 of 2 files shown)


btm:
:+1:

omnibus/resources/chef/msi/source.wxs.erb
```diff
@@ -78,6 +87,13 @@
         </Directory>
         <Directory Id="INSTALLLOCATION" Name="opscode">
           <Directory Id="PROJECTLOCATION" Name="$(var.ProjectLocationDir)" >
+            <Component Id="ProjectLocationPermissions" Guid="{75f50556-efae-4ede-beb2-a2c9b1a4d43f}" >
```
danielsdeleo: Maybe worth a note describing why this is needed and/or what happens instead if this section isn't present.

danielsdeleo:
:+1: As a person who's not very familiar with MSIs it's not 100% obvious to me why exactly the permissions section is there, would be nice to have a brief comment.

jaym:
:+1: I canceled your appveyor test. They probably would have failed anyway, you can merge whenever you want

PR #4502, opened 2016-02-01
make rubygems unit specs unit like

jaym:
These tests used to take 390ish seconds to run in appveyor

### DIFF
--- a/spec/unit/provider/package/rubygems_spec.rb
+++ b/spec/unit/provider/package/rubygems_spec.rb
@@ -480,16 +480,30 @@ describe Chef::Provider::Package::Rubygems do
         @provider.current_resource = @current_resource
         @gem_dep = Gem::Dependency.new("rspec-core", @spec_version)
         allow(@provider).to receive(:load_current_resource)
+        version = Gem::Version.new(@spec_version)
+        allow(@provider.gem_env).to receive(:candidate_version_from_remote).
+            with(@gem_dep).
+            and_return(version)
       end
 
       describe "in the current gem environment" do
         it "installs the gem via the gems api when no explicit options are used" do
+          version = Gem::Version.new(@spec_version)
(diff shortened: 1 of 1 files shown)


ksubrama:
Not :construction: any more.  /cc @chef/client-windows

ksubrama:
With this refactor, I have killed all the sequential state stored in instance variables like @gem_dep, @provider and @new_resource.  These are now changed to be very "rspec"-y, allowing each test to override values such as gem_name and bindir without having to re-initialize the provider over and over in each test.

The core speedup is seen in the `before` block of the "gem install" tests (line 527).  It stubs out "candidate_version_from_remote" and returns a preset version, allowing this test to bypass costly network operations.  The target version is set to 9000 just to further emphasize that the stub is effective - a real call out to rubygems would fail for such a version.

spec/unit/provider/package/rubygems_spec.rb
```diff
@@ -327,357 +327,457 @@ def gemspec(name, version)
...
     end
 
     describe "when installing a gem" do
+      let(:target_version) { "9000.0.2" }
+      let(:current_version) { nil }
+      let(:candidate_version) { "9000.0.2" }
+      let(:current_resource) do
+        current_resource = Chef::Resource::GemPackage.new(gem_name)
+        current_resource.version(current_version)
+        #current_resource.gem_binary(gem_binary) if gem_binary
```
jaym: commented out code

mwrock:
:+1: I like it. I also like to use rspec's `subject` for the actual "subject under test" like `provider` here but not a big deal. This is a nice improvement.

jaym:
I don't have timings for how long this test now takes in appveyor, but it wasn't in the top 10, so it is significantly faster

:+1:

PR #4503, opened 2016-02-01
Use minimal Ohai for tests

chefsalim:
Use minimal Ohai in order to speed up the AppVeyor tests.

@chef/client-maintainers @jkeiser @jaym


lamont-granquist:
can we rip out all the `Ohai::System.new`s and just re-use the minimal ohai that spec_helper already setup for everyone to use?

https://github.com/chef/chef/blob/master/spec/spec_helper.rb#L92-L104

danielsdeleo:
I didn't see why we need to make the tests conditional on `Chef::Config[:minimal_ohai]`? Is there a reason why we can't just do that everywhere?

PR #4505, opened 2016-02-01
Don't rerun ohai for ohai_spec when not needed

jaym:
Running ohai is slow.

This is still not as fast as I hoped, but shaves more than 2/3's of the time off.
Still at 26 seconds on my machine


chefsalim:
LGTM :+1:

lamont-granquist:
:+1: 

relates to #4503

PR #4509, opened 2016-02-02
Modified all windows paths to not reload anything from ohai

jaym:
Ohai on windows is really slow. While we should fix this, we should also make tests fast and not keep redoing work. This was caused by https://github.com/chef/ohai/issues/715

I've also marked the windows_service_spec as broken. It is extremely unreliable. It causes the entire process to die in the 64 bit tests. Still unclear why this has become so unstable lately.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,5 +32,4 @@ build_script:
   - bundle install || bundle install || bundle install
 
 test_script:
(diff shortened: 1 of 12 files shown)


smurawski:
:+1:  LGTM

mwrock:
:+1:

ksubrama:
:+1: Can you squash?

PR #4516, opened 2016-02-02
fixes #4515 and use FormatMessage to get correct error messages from native API instead of retyping them with our fingers

### LINKED ISSUE #4515: FormatMessage for Win32 API error code is not returning the proper error message
### Version:
12.6.0

### Environment:
Windows 7 sp1

### Scenario:
Whenever there is a problem using one of the windows resources that interact with the Win32 API, the formatting of the message done in https://github.com/chef/chef/blob/e245519d02c3de655e9ddf8e833275719dc35af9/lib/chef/win32/error.rb#L61 just outputs the error code instead of the human readable message.

For example, trying to mount a network drive on Windows using domain credentials without the domain part:

```
mount 'T:' do
  action    :mount
  device    '\\\\server-name\folder'
  username  'DOMAIN\username'
  password  'password'
  #domain   'domain'
end
```

Would produce the following error:

```
ArgumentError: mount[T:] (windows_common::default line 59) had an error: ArgumentError: ---- Begin Win32 API output ----
Net Api Error Code: 87
Net Api Error Message: Received unknown error code (87)
```
Which doesn't produce valuable search results when googling it. The [proper message](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx) that should be outputted is:

```
ERROR_INVALID_PARAMETER
87 (0x57)
The parameter is incorrect.
```

### Expected Result:
Human-readable error message displayed

### Actual Result:
API error code

mwrock:


randomcamel:
:+1:

jaym:
`Chef::ReservedNames::Win32::Error.raise!` did the wrong thing in the context at one point. Is that now fixed?

mwrock:
fixed failing tests.

mwrock:
issues with chef-provisioning-aws causing redness unrelated to this change

jaym:
:+1:

PR #4519, opened 2016-02-03
add Resource#apply for applying blocks

lamont-granquist:
Example:

```ruby
  modes = proc { mode "0644"; group "root"; owner "root" }
  file "/tmp/whatever" do
    apply modes
  end
```

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1461,6 +1461,19 @@ class Chef
       end
     end
 
+    # Apply a proc or block to the resource.
+    #
+    # modes = proc { mode "0644"; group "root"; owner "root" }
+    # file "/tmp/whatever" do
+    #   apply modes
+    # end
+    #
+    # @method apply(proc, &block)
+    def apply(prok=nil, &block)
+      block ||= prok
+      self.instance_eval(&block)
+    end
+
     #
     # Preface an exception message with generic Resource information.
     #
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@chef/client-core instead of scopes or defaults, here's a trivial way to pass codeblocks to resources and just instance_eval them.

ranjib:
- Users can use straight instance_eval, why another wrapper ? to support both block/proc ? This does not solves the default's use case. Consumers has to explicitly opt in for apply.
- Will it make sense to call the block with arguments ? like recipe/cookbook name etc..

lamont-granquist:
it is DSL sugar and looks cleaner than hanging an `instance_eval` off of the resource.

didn't mean to imply it solves the default use case, it clearly does not, it only provides a way to make a recipe DRY-er and clean looking.

its also minimally invasive and doesn't have long-distance effects the way that defaults and scopes do.

block arguments are fine, i think it already gets the resource (self) as the first arg, and that could be extended.

coderanger:
I don't see the value over `instance_eval`. This seems like a hugely advanced thing to do in general and if someone is writing code that complex I don't think shielding them from Ruby is helpful.

jkeiser:
Yep, I'd prefer defaults too, but this is a pattern that at least lets people do the thing until we implement them (and defaults are still undefined enough that they'll require a real push before we can make them).

coderanger:
The pattern is fine, I just don't think it's common enough to need a DSL addition which is almost literally `instance_eval`.

coderanger:
Especially with a common word like `apply` which we might want for something more interesting in the future :)

PR #4520, opened 2016-02-03
Fix few dsc_resource bugs

smurawski:
* Chef::Mixin::PowerShellTypeCoercions can translate a hash, but cannot translate an ImmutableMash which causes it to serialize it as a string.  This breaks DSC resources that are expecting a hash.

* Chef::Mixin::PowerShellTypeCoercions can translate a array, but cannot translate an ImmutableArray which causes it to serialize it as a string.  This breaks DSC resources that are expecting a array.

* `dsc_resource` does not append the verbose output from the set method when updating a resource that was not in the desired state.


smurawski:
review please @chef/client-windows

jaym:
I think looking those up by hash is probably the wrong approach, and type coercions should have a list of things that it checks is_a? against.

Anyway, this also works, so :+1:

mwrock:
:+1:

PR #4521, opened 2016-02-03
Pin to ohai 8.10.0 for upcoming chef client release

chefsalim:
@chef/client-core @tyler-ball

### DIFF
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -57,7 +57,7 @@ end
 
 # Chef Release version pinning
 override :chef, version: "local_source"
-override :ohai, version: "master"
+override :ohai, version: "8.10.0"
 
 
 dependency "preparation"


lamont-granquist:
pins should go into a 12.7-stable release branch, master should float on master.

lamont-granquist:
sigh, nevermind...

:+1: 

forgot the whole release pipeline is constructed around the idea we only ship master, but we don't follow that so we have to commit pins to master every time we release...

omnibus/config/projects/chef.rb
```diff
@@ -57,7 +57,7 @@
 
 # Chef Release version pinning
 override :chef, version: "local_source"
-override :ohai, version: "master"
+override :ohai, version: "8.10.0"
```
tyler-ball: This should be `v8.10.0` because that is what the [tag](https://github.com/chef/ohai/tree/v8.10.0) is named

tyler-ball:
:+1: after my comment is addressed

jkeiser:
What problem does this PR solve? I'm not intrinsically opposed, but not seeing what it actually does for us besides complicating our process.

PR #4522, opened 2016-02-03
remove Gemfile.lock from omnibus

lamont-granquist:
CI/CD - let it float, fix it when it breaks


btm:
:rocket: :guitar: :+1:

PR #4526, opened 2016-02-04
whitespace fixes

lamont-granquist:
4174 Style/SpaceInsideHashLiteralBraces
1860 Style/SpaceAroundOperators
1336 Style/SpaceInsideBlockBraces
1292 Style/AlignHash
 997 Style/SpaceAfterComma
 860 Style/SpaceAroundEqualsInParameterDefault
 310 Style/EmptyLines
 294 Style/IndentationConsistency
 267 Style/TrailingWhitespace
 238 Style/ExtraSpacing
 212 Style/SpaceBeforeBlockBraces
 166 Style/MultilineOperationIndentation
 144 Style/TrailingBlankLines
 120 Style/EmptyLineBetweenDefs
 101 Style/IndentationWidth
  82 Style/SpaceAroundBlockParameters
  40 Style/EmptyLinesAroundMethodBody
  29 Style/EmptyLinesAroundAccessModifier
   1 Style/RescueEnsureAlignment

omitted:

Style/EmptyLinesAroundBlockBody
Style/EmptyLinesAroundClassBody
Style/EmptyLinesAroundModuleBody
Style/IndentHash


lamont-granquist:
@chef/client-core some of these are probably more debatable than others.

Style/EmptyLinesAroundBlockBody and Style/EmptyLinesAroundClassBody i think i would like to drop.

stevendanna:
:+1: Confirmed locally with `git show` and `git show -w` that it is indeed only whitespace changes.

stevendanna:
Personally, I like the space between the start of the class/module declaration and the first method declaration.  I'm not sure which rule that would mean dropping.  But I can live with no space too.

danielsdeleo:
Yeah, I don't like removing all the space around class bodies, especially the space on the first line after `class Foo`. It makes everything run together and makes logical grouping of things less obvious.

lamont-granquist:
I updated the offenses count with some examples of the errors that get thrown back (a lot of them are paired up with start/end, last/first or whatever, so i just copied example output into the description -- uniq'ing and sorting them created quite a bit of a duplicated mess).  Sometimes its not entirely clear from the cop name if its taking whitespace away or adding it and the description should help...

danielsdeleo:
Yeah, so I don't really like these:

* Style/EmptyLinesAroundClassBody - I prefer to have this
* Style/EmptyLinesAroundModuleBody - same
* Style/IndentHash - I think it's fine and readable to have:
```ruby
thing = {
  a: "a"
  b: "b"
}
```

I'm still not sure what Style/AlignHash is, is that the keys or the values?

lamont-granquist:
i think there's a style tweak for begin/end alignment that allows that kind of alignment, but i don't see the option for indenthash i don't think...

lamont-granquist:
```
Style/IndentHash:
  Description: Checks the indentation of the first key in a hash literal.
  Enabled: true
  EnforcedStyle: special_inside_parentheses
  SupportedStyles:
  - special_inside_parentheses
  - consistent
  - align_braces
```

maybe we could use 'consistent' here instead?  i'm not sure what those options actually mean though...

lamont-granquist:
AlignHash also has a lot of options:

```
Style/AlignHash:
  Description: Align the elements of a hash literal if they span more than one line.
  Enabled: true
  EnforcedHashRocketStyle: key
  EnforcedColonStyle: key
  EnforcedLastArgumentHashStyle: always_inspect
  SupportedLastArgumentHashStyles:
  - always_inspect
  - always_ignore
  - ignore_implicit
  - ignore_explicit
```

danielsdeleo:
I don't really care to enforce hash indentation either way, as long as the keys are aligned (I figured out that key alignment is what that other one is about, which is good).

Also, some of the cases of `Style/TrailingBlankLines` I dislike, in particular when it's example cookbook DSL stuff in the specs. For regular ruby code I don't care a lot either way, but my brain likes a trailing whitespace in Chef code like metadata.rb and such.

danielsdeleo:
Yeah, the instances of align hash I could find in the code are ok. I'm a little unhappy with how they look in some method calls and I also wonder if rubocop is capable of distinguishing (or even cares to distinguish) between opts and keyword args, but I can tolerate it if it prevents code like:
```ruby
h = {
  a: "a"
    b: "b"
}
```

lamont-granquist:
PR updated

chef-config/spec/unit/config_spec.rb
```diff
@@ -289,9 +286,7 @@ def to_platform(*args)
         end
 
         describe "ChefConfig::Config[:repo_mode]" do
-
           context "when local mode is enabled" do
-
             before { ChefConfig::Config[:local_mode] = true }
```
lamont-granquist: i think whichever one is doing this that i don't want it either....

lamont-granquist:
PR updated again, dropped Style/EmptyLinesAroundBlockBody as well

danielsdeleo:
Everything in the diff looked reasonable to me.

lamont-granquist:
Yeah, same here.  Should satisfy @stevendanna as well, so I think I'll merge and push the chefstyle merge as well, before it gets conflictey

PR #4527, opened 2016-02-04
Lock php cookbook in kitchen tests to solvable version

danielsdeleo:
Seems we're hitting a molinillo bug that prevents us from solving
dependencies correctly.

cc @jaym


jaym:
yea, a bug. :+1: 
We can also include gecode

Issue #4368, opened 2016-01-08
Coercing attributes inside an array


coderanger:
An odd report from Slack. If you are using Solo, and loading role or environment data from `.rb` files, and in those files you are using symbols as keys (either directly or via the `{a:1}` short syntax), you can end up with a situation where data isn't coerced so that only one of string or symbol keys work.

Specifically when you have an array of hashes in attribute data, we don't seem to coerce those inner hashes to `Mash`/`VividMash`. I feel like this used to work, but I can't find any evidence for that. Regardless, it probably should. Thoughts?

/cc @chef/client-core

coderanger:
A clearer example:

```ruby
# myrole.rb
default_attributes({
  foo: [{bar: 1}],
})

# recipe code
node['foo'][0][:bar] == 1
node['foo'][0]['bar'] == nil
```

lamont-granquist:
might be fixed by #3721

lamont-granquist:
Actually, tried to replicate this on master and I got '1' for both of those.

lamont-granquist:
I think there's no fundamental difference between solo and client when it comes to how the attributes get merged, and `lib/chef/run_list/run_list_expansion.rb` has a `#apply_role_attributes` which is responsible in both code paths for the merge:

```
      def apply_role_attributes(role)
        @default_attrs = Chef::Mixin::DeepMerge.merge(@default_attrs, role.default_attributes)
        @override_attrs = Chef::Mixin::DeepMerge.merge(@override_attrs, role.override_attributes)
      end
```

And if calling DeepMerge.merge(<Mash>, <Hash>) didn't work, that would break -j arguments as well and I'd think we'd have heard about this (plus it just doesn't make a whole lot of sense, we definitely get a Mash out of it, and any values merged in will get convert_value'd inside of the Chef::Mash.

lamont-granquist:
(although i guess with -j you only pass in JSON so you never get symbols, but still I don't see how DeepMerge would be bypassing Chef::Mash#convert_value...)

gravesb:
Could the issue be on [this line of VividMash](https://github.com/chef/chef/blob/master/lib/chef/node/attribute_collections.rb#L198) where it converts arrays to AttrArrays?  I don't see where the values of that attr array would be mashified.

lamont-granquist:
oh i see, yeah i think i derp'd when i tried to replicate it and didn't actually test an array.

lamont-granquist:
[...]

Issue #3802, opened 2015-08-20
/var/chef completely wiped by running chef-solo with default configs


gionn:
https://github.com/chef/chef/blob/master/lib/chef/application/solo.rb#L212

It isn't sufficient to just rm_r cookbooks_path?

Is it really a feature? It just destroyed my environments, data_bags, backups, etc under /var/chef 

/cc @vtolstov

lamont-granquist:
Yeah I ripped that line out of the "port" of this code from chef-solo to chef-client/local-mode.  It actually wiped out my entire chef git checkout when it wasn't stubbed correctly and ran an rm -rf on the directory that I ran rspec from.

I would be :+1: on a PR to remove that line and make it entirely the responsibility of the "caller" (the script that invokes chef-solo/client).

ranjib:
yup, i've been bitten by this as well. @gionn you want to work on it? else i can send a PR

vtolstov:
cookbooks path is not enough, because databags may have old data.
May be add new option to cleanup chef work dir or something like this?

gionn:
@vtolstov This feature is useful when using recipe_url (cookbooks.tar.gz served via https), so cleaning data_bags folder is out of scope, since something else should put them in the right place (and so it can freely empty such folder).

For me cleaning `Chef::Config[:cookbook_path]` is enough.

As a side note, I just saw:
```
Mixlib::ShellOut.new("tar zxvf #{tarball_path} -C #{recipes_path}").run_command
```

Why impose with `-z` that the archive should be gzipped? Tar is smart enough to do the right thing without any particular option (enabling us to switch to bz2 or xz at any time).

If is good for all, I will make a PR for both.

Thanks.

lamont-granquist:
The '-z' probably predates tar being smart enough to just DTRT if you leave it off.  I still have more than a decade of reflex (might be more than 20 years of reflex, can't recall now) of automatically including that flag to tar--its a hard habit to break...

vtolstov:
@gionn in case of not removing data_bags for example may bring problems, when some recipe use * from some databag. May be create two step unpack? first get root dirs from archive, and if dir present in tar, remove it from disk?

lamont-granquist:
We should rip out the rm_rf completely in Chef 13.0 and make it the responsibility of the 'caller' to do the correct thing for their environment.

PR #4529, opened 2016-02-05
Implement knife bootstrap client.d RFC

jaym:
This implements the RFC described in https://github.com/chef/chef-rfc/pull/183

There are some minor differences:
- `client.d` is natively interpreted by chef-client in this implementation. It does not rely on a block at the end of the client.rb that loads the files in client.d
- `client.d` is not found in the bootstrap directory, instead just the repo path.
- used `client_d_dir` instead of `client_d_path` to keep it consistent with `trusted_certs_dir`

cc @chef/client-core


coderanger:
How does this interact with chef-solo and local mode/knife? Do we want to make a `solo.d` and `knife.d`?

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -194,6 +198,26 @@ def trusted_certs_content
           content
         end
 
+        def client_d_content
+          content = ""
+          if @chef_config[:client_d_dir] && File.exist?(@chef_config[:client_d_dir])
+            root = Pathname(@chef_config[:client_d_dir])
```
stevendanna: Should we add a `require "pathname"` to this file?

jaym: yes, good catch

jaym:
@coderanger good question. My intention was that this does not get loaded with knife as it shouldn't be reading client.rb, and I don't personally see it useful to have for knife.

As for solo, I'm open to suggestions. Would people who use solo find it useful? I think solo.d would probably make the most sense for a directory name.

danielsdeleo:
One use case for `knife.d` would be for folks in the same org to share most of their knife config in a git repo and gitignore the actual `knife.rb` where they'd set username and other stuff.

jaym:
Interesting. That seems useful.
Perhaps we add that one.

What about solo?

thommay:
Just spoke to Jay about this. We agree with Dan that this would be useful for knife users, and the directory for that should be `.chef/config.d`. 

However, my feelings are that we should *not* do this work for solo.

coderanger:
The party line is that solo is fully supported, if we do it for client we should do it for solo.

btm:
@coderanger chef-solo being supported doesn't mean guaranteed feature parity. we don't require all features we add to the client mode to also work in chef-solo.

Further, [RFC031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md) is accepted as the path to deprecate chef-solo by ensuring local-mode has feature parity and then having chef-solo commands be an alias for local-mode.

I don't see any reason to require new feature for chef-solo, only that by supporting it we're promising to not break it.

coderanger:
@btm Given how little work it would be (the impl is the same as for chef-client), I don't see any reason not to do it other than trying to convince people that they shouldn't use `chef-solo` and that is something we should never do.

btm:
@coderanger that's a different argument. If @jaym or whoever implements it looks and thinks the code paths are similar enough that it's easy to add .d support to chef-solo, that's fine, go wild. But it's not required.

coderanger:
@btm I'm saying it should be required as long as `chef-solo` is a supported tool/workflow. AFAIK it still is.

jaym:
Added conf.d for knife things

btm:
@coderanger then I'm disagreeing. There is no chef-solo based workflow in [RFC019](https://github.com/chef/chef-rfc/blob/master/rfc019-chef-workflows.md) and chef-solo is effectively deprecated except for it's existing UX by [RFC031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md). There's no reason to add additional UX to chef-solo when all it is going to be is a backwards-compatible interface to local-mode, which does support .d configs.

coderanger:
RFC 19 is both not an exhaustive list and not fully implemented and RFC 31 hasn't been implemented at all. The whole point of RFC 31 is that this would be supported by way of solo being client internally, so why not support it from the start as literally any `chef-solo` user (and yes, they exist) would expect. How would you justify not adding new features to solo other than "we don't think you should be using it"?

coderanger:
Or alternatively "we don't want to support it", which is fine but it has to be supported until RFC 31 is fully implemented, that was the whole point of RFC 31.

thommay:
I'm pretty sure we've never said nor intended that support implies complete feature parity.

coderanger:
@thommay What does it mean then? Remember that solo is *explicitly not deprecated*. Keeping a high-quality user experience should mean that non-server-related features should go in both unless there is a compelling reason not to and no one has mentioned a reason other than "we don't want people using chef-solo anymore" which is not a reason I accept.

btm:
I hate the word support. Support means we don't break something. Everything beyond that is trouble. I can't formulate how support relates to new features. We certainly haven't required resource parity across all supported platforms before adding a new resource for one or more platforms. We've added a few features to the client that didn't make it into solo until someone added some kind of compatibility later, e.g. support in solo or via an external cookbook.

There's an issue here with our differing development priorities. Let's talk on chef-dev.

luckymike:
@btm @thommay Solo was actually just broken (existing functionality completely changed) in 12.7 to ensure that its flags had parity with client. 

>Chef Solo -r (–recipe-url) changes
https://www.chef.io/blog/2016/02/12/chef-client-12-7-released/

If breaking existing (if rarely or accidentally used) functionality was in order, why not supporting the conf directory? 

Is the goal for solo to be a debasing experience that encourages switching to client?

danielsdeleo:
@luckymike 

> Instead invokes the --recipe-url code, which had the side effect of **running an immediate unprompted rm -rf * in the current working directory** of the user. Due to this problem and other issues around this rm -rf * behavior it has been removed from the --recipe-url code in chef-solo. The use of -r in chef-solo to mean --recipe-url has also been deprecated.

So basically we screwed up, and merged an "improvement" that was actually a net negative for Chef Solo users, including ourselves, who were `rm -rf *`-ing their home directories. I guess in some pedantic sense this change violates SemVer, and it happens to only impact solo users, but tbqf I would rather delete all copies of SemVer from the internet than delete your home directory on accident.

> Is the goal for solo to be a debasing experience that encourages switching to client?

Is accidentally deleting a bunch of your files "a debasing experience"? If so, wouldn't making it less likely that you do this on accident be the opposite of that?

In any case, this discussion is about adding a feature.

btm:
@luckymike the issue that @danielsdeleo explained aside, the debate is about policy, i.e. by policy do we require feature parity to merge a new feature. It's less about adding the feature this PR represents for solo (I think there's more work/shed-building to figure out the right medium-term UX decisions given [RFC031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md) than there is writing the code for solo.d support).

luckymike:
I wasn't arguing for deleting home directories, I was trying to offer a clear example of rightfully favoring parity over policy (sem ver). Turning @coderanger's argument for including support in solo into a larger policy debate effectively decides this issue without anyone taking responsibility for it.

btm:
@luckymike by "this issue" you mean #4529, not something related to the `-r` stuff, right?

This feature is part of the plan I envisioned in the ohai-config RFC discussion for getting ohai UX happy and @coderanger happy about ohai UX being happy. So I'm responsible for it [and @coderanger's happiness!]. But we're in the middle of another hairy chef-client release so it's not my top priority at the moment.

luckymike:
Yes, the issue #4529, the issue we're commenting on :)

lamont-granquist:
We can argue about the semantics of the word "deprecated" but I think the reality is that chef-solo is "deprecated".  Time spent on chef-solo-specific codepaths is wasted time since chef-zero eventually will be replacing chef-solo via [RFC 031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md).  That also completely forces our hand when it comes to conflicting features that have been caused by chef-solo and chef-client/chef-zero having different codepaths.  RFC 31 means that at some point in the future the conflicting meanings of the '-r' flag need to be resolved.  Generally chef-solo is the smaller userbase and is going to lose that argument.  In the case in question its just laughable since making the solo behavior the new meaning of '-r' would mean that everyone using '-r' on the client to mean '--run-list' and doing elastic scaling would have their argument changed to '--recipe-url' and get a complimentary `rm -rf` run in the direction they're executing chef-client in.  We have to deprecate the meaning of '-r' for chef-solo users and I don't think there can be any meaningful debate about that.

So there isn't parity here.  And in fact I think that "solo.d" may buy us a worse compatibility issue since how is that supposed to work when we change chef-solo to invoke chef-client local-mode?  Not only is that doing work to create "solo.d" in codepaths that we'll throw away, but we need to have zero pick up solo.d for backcompat when RFC 31 is executed, which is just going to be digging ourselves a deeper hole to climb out of, it doesn't move us any closer towards executing RFC 031.

And removal of the rm_rf is entirely because when I was debugging #4368 recently, which is a chef-solo specific issue, I took my usual chef-zero command line argument:

```
chef-client -z -r 'recipe[test]'
```

And blindly converted that into a chef-solo command line:

```
chef-solo -r 'recipe[test]'
```

Which immediately wiped out all the cookbooks in my current working directory.

Which made me more or less like this:

Rage Quit

I was actually the one that committed the rm_rf line initially:

https://github.com/chef/chef/pull/2418

We've also had chef-solo user requests to remove that line:

https://github.com/chef/chef/pull/3802

And not only that but I went and added the "--delete-entire-chef-repo" line and added the rm_rf *TO* chef-client from chef-solo which didn't exist there before.  I just did it in a much safer way where users of "--recipe-url" who do NOT want that behavior do not get it forced down their throat and where a simple command line option like "-r" is not brutally destructive.

And I don't see how having to change a command line from:

```
chef-solo -r http:///whatever.com/recipies.tgz
```

To:

```
chef-solo --recipe-url http:///whatever.com/recipies.tgz --delete-entire-chef-repo
```

And getting _exactly_the_same_behavior_as_you_had_before_ is "breaking functionality" or a "debasing experience"

btm:
someone please file an issue about `chef-solo -r` if they want to keep talking about it beyond this point.

lamont-granquist:
To be clear, buried in that text was a relevant argument that I'm :-1: on solo.d because it just digs a deeper hole for RFC 031.

luckymike:
My point was that `policy > parity` is a bad argument, when parity results in a better user experience (as I think we all agree, regarding `-r`. I think that solo.d would provide parity that would be beneficial to the user. Some variety of solo.d support should also improve the RFC 031 UX for solo users, as they would be able to adopt this pattern ahead of time. 

If it isn't expedient for CHEF to support solo users in that fashion, it seems best to clearly state that and take responsibility for that decision rather than leaving things in limbo and shunting the conversation to chef-dev.

btm:
> If it isn't expedient for CHEF to support solo users in that fashion, it seems best to clearly state that and take responsibility for that decision rather than leaving things in limbo and shunting the conversation to chef-dev.

What's wrong with having policy discussion on the mailing list instead of a feature PR?

luckymike:
I don't care where you discuss policy; I care about the inclusion of this feature in solo.

It seems like you're offering a policy debate in response to: 
>The party line is that solo is fully supported, if we do it for client we should do it for solo.

Which just serves to leave this in limbo. 

So if the answer is "Not at this time" "Maybe in the future" or "No" someone should state that firmly here so the community has a clear understanding of CHEF's priorities regarding the solo UX.

btm:
@luckymike okay, I'll make sure you know here when we figure it out.

coderanger:
The plan from talking to Thom yesterday is to move forward ASAP with shooting *old-solo* in the face, https://github.com/chef/chef/issues/4567 is tracking the odds and ends we need to do for that. Once that's done, local mode will be reborn as solo and all will be well.

smurawski:
:+1:

chef-config/lib/chef-config/workstation_config_loader.rb
```diff
@@ -124,6 +126,27 @@ def locate_local_config
...
+
+    def conf_d_files
+      @conf_d_files ||=
+        begin
+          entries = if Config[:conf_d_dir]
+                      Dir.glob(File.join(PathHelper.escape_glob(
+                        Config[:conf_d_dir]), "*.rb"))
+                    else
+                      []
+                    end
```
lamont-granquist: can `Array.new()` clean this up a bit?

lamont-granquist:
one nit, code LGTM :+1:

chef-config/lib/chef-config/workstation_config_loader.rb
jaym: Maybe, but I'm not seeing it. What did you have in mind?

btm:
```
entries = Array.new
(entries << Dir.glob(File.join(PathHelper.escape_glob(Config[:conf_d_dir]), "*.rb"))).flatten! if Config[:conf_d_dir]
```

jaym:
I think we need #4569 to fix windows

jaym:
@chef/client-maintainers I added solo.d, could probably use another round of reviews

chef-config/lib/chef-config/config.rb
```diff
@@ -490,6 +490,16 @@ def self.path_accessible?(path)
...
 
+    # A directory that contains additional configuration scripts to load for chef-client
+    default(:client_d_dir) { PathHelper.join(config_dir, "client.d") }
+
+    # A directory that contains additional configuration scripts to load for solo
+    default(:solo_d_dir) { PathHelper.join(config_dir, "solo.d") }
+
+    # A directory that contains additional configuration scripts to load for
+    # the workstation config
+    default(:conf_d_dir) { PathHelper.join(config_dir, "conf.d") }
```
btm: I think this is supposed to be `config.d` because the workstation config is `config.rb`.

coderanger: Isn't workstation config usually `knife.rb`?

danielsdeleo: `knife.rb` is very softly deprecated. There's no real upside to forcing people to switch, but we decided to use `config.rb` going forward because it's used by many tools other than knife.

jaym: `config.rb` is supposed to be used in preference to `knife.rb` according to the docs

coderanger: TIL.

btm:
:+1: with the s/conf_d/config_d/ change.

smurawski:
confirming my :+1: after the latest updates

PR #4533, opened 2016-02-05
Explicitly use before_install on each entry in travis matrix

jaym:
Something in travis seems to have change. Perhaps we were never
running our `before_install`, and just relying on the fact that
bundler was there. See https://github.com/travis-ci/travis-ci/issues/5578

Or `before_install` broke.

We can try reverting this change next week to see if the issue
is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,33 +32,63 @@ matrix:
   - rvm: 2.1
     sudo: true
     bundler_args: --without server docgen maintenance
+    before_install:
+      - gem update --system
+      - gem install bundler
   - rvm: 2.2
     sudo: true
     bundler_args: --without server docgen maintenance
+    before_install:
+      - gem update --system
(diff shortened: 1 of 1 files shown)


lamont-granquist:
halite has a bug:  https://github.com/poise/halite/pull/3

someone threw the empty before_install: into there deliberately in order to pin bundler+rubygems to the default travis versions i think

lamont-granquist:
we definitely were running our before_install previously, just not on everything

lamont-granquist:
looking at the last failure on a master build:

https://travis-ci.org/chef/chef/builds/107293443

not sure why you're doing this.  the only failure was odd shit in ruby 2.1 compiling gems and the before_install definitely ran...

jaym:
Yea, this is the wrong change, it looked similar to the one i had this morning so I thought i'd try it out

PR #4535, opened 2016-02-05
ChefFS inline

danielsdeleo:
Just for illustration at this point. I personally have the hardest time working with ChefFS because of the many levels of class hierarchy, so I wanted to see what it would look like to totally remove that. If we did this for everything we might find a different way to build abstractions around composition or even with inheritance but with fewer levels of it.


randomcamel:
Hmm. This seems like it would copy-paste a lot of code which, when it's different, is often only subtly so (and a lot of that could be DRYed up with some minor object introspection). There's over a dozen objects in `repository`, so it's non-trivial from a maintenance perspective.

- Is the current/prospective level of DRY not as good as I think it is?

- If it is, what are the options for keeping the DRY, but in a more comprehensible object hierarchy.

:+1: to rethinking ChefFS--@jkeiser has been pondering it as well.

randomcamel:
One other way to enhance clarity might be to use superclasses, but have subclasses that strictly speaking don't need to be there. For example, there's `PoliciesDir` but no `PoliciesEntry`; it's possible an empty `PoliciesEntry` might make things a little more apparent.

danielsdeleo:
The idea is to start over in terms of abstractions and code re-use. Right now there's a lot of special cases and such sprinkled throughout the various levels of class hierarchy. I'm pretty radical about wanting composition instead of inheritance, but even if inheritance is the right thing, I think I'd want no more than a _purely abstract_ base (for doc purposes more than anything), implementation base, and a leaf node.

I'd have to go a little further to say what exactly that might look like, but one thing that seems pretty clear from what's done so far is that giving directory things a separate hierarchy removes a lot of confusing branches from the base-r classes which makes it more readable. For example, this intermediate commit: https://github.com/chef/chef/commit/96fb2353e3cf16e395c7f7c0ba00953f960ff586

danielsdeleo:
Ok, I did a second one, so you can take a look at the commonality. So far it looks pretty clear how these could all be collapsed to one thing that would represent a top-level directory in a chef repo by moving the special cases somewhere else.

In any case I should be super clear that the end result isn't to remove all code reuse, I want to eventually get to the "right" code reuse. I'm following this technique: http://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

danielsdeleo:
Okay, after doing a few of these types and looking at it, I think we can get some good readability benefits if we do the following:

* make a type for everything
* move responsibility for creating a thing into the type that represents that thing

That should allow us to make all of the directory types completely identical, except that they have different child types. Customizations like the various data handlers end up being defined in the type that is relevant to them instead of being found via the parent object, which is where you'd expect them to be. And also classes that represent dirs don't have any code that is only relevant to files, and likewise classes that represent files don't need to have any code about children or directories.

An example of what that would look like is here: (though I didn't extract any code yet): https://github.com/chef/chef/commit/3b4f345b03933672197c2a869f9a36f7382fe984

ranjib:
looking good :-) . I would drop the ChefRepository bits from class name, since they are under ChefFS::Filesystem::Repository namespace

randomcamel:
@danielsdeleo Nice. Those are great design principles, the code in https://github.com/chef/chef/commit/3b4f345b03933672197c2a869f9a36f7382fe984 especially is much more readable.

danielsdeleo:
Okay, I'd like to merge what I have in order to avoid rebase hell, and then come up with a plan to implement a few more improvements and get to a good stopping place. The next things I would plan to do are:

* Make types for dirs representing `nodes/`, `clients/`, etc. so that all these things can be made consistent with the other stuff: https://github.com/chef/chef/blob/a5afeecc19170cac9d430f7bee128d05e9bca5a9/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb#L182-L202 I can probably move the data handler stuff into specific classes for each type and remove abstracted support for it after that.
* Make a file-specific class, same as I did for directories, and use it at least for the simple cases of basic JSON files (i.e., things that aren't complicated the way cookbooks are).
* Review the feasibility of simplifying the cookbook case. Can it be composed of a file and directory object, and delegate based on some information it has? Could the top-level cookbook object determine what's a file and what's a directory?
* Review the places where the old "base" types are used, (`BaseFSDir`, `BaseFSObject`, etc.) and look for ways to de-duplicate with the new stuff where I inlined their methods, if possible.

@jkeiser @randomcamel thoughts?

jkeiser:
I'm all for expanding code where code needs expanding, but I don't understand what exactly this copy paste is accomplishing--is the point that you don't know what methods exist on these subclasses, and you want to know the method names? If so, would it possibly be better to add methods that do this to make it clear what's happening?

```ruby
def create(...)
  super
end
```

jkeiser:
After talking, I think it's worth a shot assuming you continue on :) The separation of entry from directory looks like a Good Thing :+1:

PR #4536, opened 2016-02-05
autofixing some missed cops

lamont-granquist:
second round of autofixing for some reason...


lamont-granquist:
missing from #4526 somehow...

PR #4541, opened 2016-02-08
fix a batch of cops

lamont-granquist:
applying auto-fixes for:

Style/NegatedWhile
Style/ParenthesesAroundCondition
Style/WhileUntilDo
Style/WordArray
Performance/ReverseEach
Style/ColonMethodCall

will be disabled:

Style/TrivialAccessors - buggy should IgnoreClassMethods: true at least if we ever use it.
Style/UnneededPercentQ - buggy

### DIFF
--- a/bin/chef-shell
+++ b/bin/chef-shell
@@ -33,5 +33,5 @@ require "chef/shell"
 
 # On Windows only, enable irb --noreadline because of input problems --
 # See CHEF-3284.
-IRB.conf[:USE_READLINE] = false if Chef::Platform::windows?
+IRB.conf[:USE_READLINE] = false if Chef::Platform.windows?
 Shell.start
--- a/chef-config/lib/chef-config/package_task.rb
+++ b/chef-config/lib/chef-config/package_task.rb
@@ -189,7 +189,7 @@ end
         IO.write(version_rb_path, contents)
       end
 
-      Dir[File.expand_path("*gemspec", root_path)].reverse.each do |gemspec_path|
+      Dir[File.expand_path("*gemspec", root_path)].reverse_each do |gemspec_path|
         gemspec = eval(IO.read(gemspec_path))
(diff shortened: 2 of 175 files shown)


lamont-granquist:
@chef/client-core todays batch of rubocop fixes.

lamont-granquist:
looks like i broke rubocop and i'm having a hard time pull in the current released version.

danielsdeleo:
I'm a big fan of the Style/ColonMethodCall rule but I'm not sure about the rest of them. I would like the trivial accessor thing a lot more if you could exclude class methods. I don't find:

```ruby
class << self
  attr_whatever :foo
end
```

particularly better than the `def self.foo` way

lamont-granquist:
Yeah TrivialAcessor broke, and looks kinda rough, the UnneededPercentQ one broke as well.

lamont-granquist:
@chef/core this is better now and passing tests.

thommay:
:+1:

danielsdeleo:
I don't really see the value in Style/WordArray either, e.g., if you change `["foo", "bar", result_of_method_call]` to just `["foo", "bar"]`, then it kinda feels like having to change single quotes to doubles to add interpolation. Not really motivated to argue about it though.

:+1: either way

lamont-granquist:
hmmm, feels like that happens a lot less often to me, and %w is pretty commonly considered a best-practice (one that i try to practice...)

lamont-granquist:
@chef/client-core polling anyone else for opinions?

lamont-granquist:
Welp, I talked to @thommay about this and he was gonna :+1: this PR with some words, but he's busy CBGB'ing.

@danielsdeleo we gave your argument quite a bit of thought, but in the end felt that enforcing WordArray is a really common ruby style guide and the exception doesn't happen with anywhere near the frequency that the string interpolation ones does...

And if it pisses us off too much we can always rip it back out again...

PR #1751, opened 2014-08-05
Make solo tarballs work on windows; fixes #1515

adamhjk:
Adds Guard support to Chef.

Makes solo tarballs work on windows, by converting to ShellOut and adding smarter path separator support to the regex that detects the cookbook directory to untar into.

PR #4545, opened 2016-02-09
removing rm -rf in chef-solo recipe_url

### LINKED ISSUE #1515: Can't use remote cookbooks.tar.gz with chef-solo on windows
https://tickets.opscode.com/browse/CHEF-5332
https://tickets.opscode.com/browse/CHEF-5333

`cookbooks.tar.gz` can not be used to make a chef-solo run on windows.

```
chef-solo -j C:\chef\first-boot.json -r http://foo/bar.tar.gz -E dev -l debug --logfile C:\chef\chef-solo.log
```

fails with: 

```
C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.4-x86-mingw32/lib/c
hef/application/solo.rb:190:in `join': can't convert nil into String (TypeError)

        from C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.4-x86-
mingw32/lib/chef/application/solo.rb:190:in `reconfigure'
        from C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.4-x86-
mingw32/lib/chef/application.rb:65:in `run'
        from C:/opscode/chef/embedded/lib/ruby/gems/1.9.1/gems/chef-11.12.4-x86-
mingw32/bin/chef-solo:25:in `<top (required)>'
        from C:/opscode/chef/bin/chef-solo:23:in `load'
        from C:/opscode/chef/bin/chef-solo:23:in `<main>'
```

### LINKED ISSUE #3802: /var/chef completely wiped by running chef-solo with default configs

lamont-granquist:
- deprecates '-r' used for the recipe_url in chef-solo
- adds --delete-entire-chef-repo option for users who want the
  old behavior back.
- cleans up some old code

closes #3802
closes #1515
closes #1751

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -25,6 +25,7 @@ require "chef/log"
 require "chef/config_fetcher"
 require "chef/handler/error_report"
 require "chef/workstation_config_loader"
+require "chef/mixin/shell_out"
 
 class Chef::Application::Client < Chef::Application
   include Chef::Mixin::ShellOut
@@ -279,6 +280,11 @@ class Chef::Application::Client < Chef::Application
     :description  => "Enable fips mode",
     :boolean      => true
 
+  option :delete_entire_chef_repo,
+    :long           => "--delete-entire-chef-repo",
+    :description    => "DANGEROUS: does what it says, only useful with --recipe-url",
+    :boolean        => true
+
   IMMEDIATE_RUN_SIGNAL = "1".freeze
 
   attr_reader :chef_client_json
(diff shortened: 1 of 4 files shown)


btm:
alternative for consideration, exit if the thing passed isn't a valid url?

lamont-granquist:
We don't do the rm_rf in the client code, and in order to synchronize the 'features' i strongly vote that the client version of this code wins -- either now, or in 13.0

lamont-granquist:
see #3802

lamont-granquist:
also the lack of stubs around the rm_rf in the unit tests blew away my chef github checkout once here: #2810

lamont-granquist:
So, it has trolled me twice and destroyed data -- only one of which times it would have been avoided by the 'check for a real url', and its trolled @ranjib and @gionn when using it 'correctly' as well.  The 'feature' is really indefensible.

See also https://github.com/ValveSoftware/steam-for-linux/issues/3671

thommay:
I would be in favour of removing the `rm -rf` now, and then dumping the `-r` argument in 13.

danielsdeleo:
The way it behaves now is terri-bad and I'm all for changing it. I just wonder how we even got to this situation. Did we change something in the config to make local mode work better and now `recipes_path` defaults to `pwd` where it didn't before? Can we dig up the reason this `rm_rf` exists in the first place? I'm :+1: on this now but would be + :100: if we could figure out the point of this code and find a good way of dealing with whatever the original problem was (e.g., extract the tarball to a cache dir and fixup `Chef::Config`'s paths). I would also be +90 if we could at least detect the failure mode ahead of time and give the user an explanation instead of something like shell out failing on `tar zxvf`. This code in ChefDK is kinda along the lines of what I'm trying to describe: https://github.com/chef/chef-dk/blob/23d880e44dc36a60ea036f57ff44850418639abb/lib/chef-dk/policyfile_services/export_repo.rb#L339

lamont-granquist:
see #2400

i forgot that i was the one that merged this...

shell_out doesn't fail on the untar, but if there's already a cookbooks dir that you're extracting over then odd things could happen.

lamont-granquist:
@chef/client-core okay this is a bit less angry now

- the -r option to solo is deprecated and just complains to users to use --recipe-url
- a --delete-entire-chef-repo is added (with no short option) to do the old FileUtils.rm_rf, users that want the old behavior can add it back with this option
- also uses cleanpath to fix the windows path issue in --recipe-url due to multiple cookbook paths that has been kicking around since forever

lamont-granquist:
also a bit of cleanup to use shell_out! more consistently, etc.  there's probably a common method to extract here and put it in the superclass, but at the same time the local mode check and the cookbook-array junk in chef-solo make it different enough that i think i've spent enough time on this bit of code for now...

btm:
:+1:

danielsdeleo:
:+1: modulo whatever travis trolling is going on

lamont-granquist:
s/travis trolling/rubocop trolling/

PR #4548, opened 2016-02-09
tell knife's edit_data what the object is

thommay:
rather than guessing through json autoinflation

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -510,8 +510,8 @@ class Chef
       response.body
     end
 
-    def create_object(object, pretty_name = nil, &block)
-      output = edit_data(object)
+    def create_object(object, pretty_name = nil, object_class: nil, &block)
+      output = edit_data(object, object_class: object_class)
 
       if Kernel.block_given?
         output = block.call(output)
--- a/lib/chef/knife/core/ui.rb
+++ b/lib/chef/knife/core/ui.rb
@@ -165,13 +165,13 @@ class Chef
 
       # Hash -> Hash
       # Works the same as edit_data but
-      # returns a hash rather than a JSON string/Fully infated object
+      # returns a hash rather than a JSON string/Fully inflated object
       def edit_hash(hash)
         raw = edit_data(hash, false)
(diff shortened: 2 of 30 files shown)


thommay:
now with bonus kwargs thanks to mr Keiser.

lamont-granquist:
yep :+1:

jkeiser:
:+1:

stevendanna:
:+1: Left minor comments but they aren't blocking.

lib/chef/knife/core/ui.rb
```diff
@@ -184,13 +184,22 @@ def edit_data(data, parse_output = true)
           end
         end
 
-        parse_output ? Chef::JSONCompat.from_json(output) : output
+        if parse_output
+          if klass.nil?
+            Chef::JSONCompat.from_json(output)
+          else
+            hsh = Chef::JSONCompat.parse(output)
+            klass.send(:from_hash, hsh)
```
stevendanna: Is the send needed here?  `klass.from_hash(hsh)`

lib/chef/knife.rb
```diff
@@ -510,8 +510,8 @@ def format_rest_error(response)
       response.body
     end
 
-    def create_object(object, pretty_name = nil, &block)
-      output = edit_data(object)
+    def create_object(object, pretty_name = nil, klass: nil, &block)
```
stevendanna: I know `klass` is pretty standard but since the data for the object is called `object`, `object_class` might be clear and has the benefit of being made up of real words :)

thommay:
Not yet ready for merge.

thommay:
There was some pretty side effects driven programming going on here that took me a while to unpick - we were depending on the behaviour that `#from_json` would give us a hash back if it didn't know what the class was, and then simply passing a json object that didn't include `chef_class` in in those cases. I've tested all the major knife commands with this branch and they all work, and none give me a deprecation error.
I've also included @stevendanna's feedback.
 So @chef/client-core, please re-review.

lamont-granquist:
:+1:

danielsdeleo:
:+1:

PR #4551, opened 2016-02-09
Update MSI ManufacturerName to be a cleaner string

chefsalim:
Our PublisherName shows up in Add/Remove Control Panel as a quoted string - this looks ugly, and is not the standard way to represent the package in the ARP.  The reason is we are doing an XML escape of the Manufacturer property in the MSI.  This PR puts a cleaner version of the string in the loc erb - sans the maintainers email address (which also doesn't belong in the PublisherName property).

@chef/client-core @chef/omnibus-maintainers @adamedx


danielsdeleo:
I know basically nothing about MSIs but this makes sense to me :+1:

jaym:
:+1:

btm:
:+1:

PR #4550, opened 2016-02-09
Use a streaming request to download cookbook

thommay:
fixes #4549

PR #4552, opened 2016-02-10
VERSION change and CHANGELOG for 12.7.1 release

tyler-ball:
@thommay found an [regression](https://github.com/chef/chef/pull/4550) in 12.7.0 and fixed it

\cc @chef/maintainers 

You should probably [ignore whitespace](https://github.com/chef/chef/pull/4552/files?w=1) on this diff


btm:
A thought, even though 12.7.0 wasn't promoted, we still could have a separate section in the changelog for 12.7.1, right? that way it obviously communicates the difference between 12.7.0 and 12.7.1?

tyler-ball:
I'm fine for doing that here since we are still manually incrementing the patch version and changing the CHANGELOG.  But when we have automatic patch revision changes, are we going to manage the CHANGELOG like that?  It seems weird to have 10 sections under 12.8 each with 2 PRs in them.

btm:
I think what we'll start doing when everything is automated is having a changelog section for only x.y, because every build will bump build/patch.

I'm thinking about what we do after a 12.8.8 goes out, and we need to ship a regression fix. We probably add a 12.8.8 section, and then go back to having just a 12.9 section. It feels confusing, but in the context of basically every release is going to bump the minor because every release is going to have features for a [long] while as we get on a monthly release cadence, except for the quick turnaround regression fixes....

tyler-ball:
@btm Split into `12.7.0` and `12.7.1`

btm:
:+1:

PR #4565, opened 2016-02-11
Disable unit tests in later stages of CD pipeline

schisamo:
Travis/Appveyor already run the unit tests at the appropriate gate earlier in the build process. Running the tests on each platform we build a package for should give the same result and really just serves to slow the overall delivery process down.

This is per a meeting we had earlier to improve build speeds.

/cc @chef/engineering-services @chef/ship-it @chef/workflow @chef/ecosystem-team

### DIFF
--- a/ci/verify-chef.bat
+++ b/ci/verify-chef.bat
@@ -57,7 +57,7 @@ IF NOT EXIST "Gemfile.lock" (
 IF "%PIPELINE_NAME%" == "chef-fips" (
 	set CHEF_FIPS=1
 )
-call bundle exec rspec -r rspec_junit_formatter -f RspecJunitFormatter -o %WORKSPACE%\test.xml -f documentation spec/unit spec/functional
+call bundle exec rspec -r rspec_junit_formatter -f RspecJunitFormatter -o %WORKSPACE%\test.xml -f documentation spec/functional
 
 REM ; Destroy everything at the end for good measure.
 RMDIR /S /Q %TEMP%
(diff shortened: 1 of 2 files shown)


chefsalim:
:+1:

scotthain:
Does this mean that the unit tests will not be run on platforms that Travis/Appveyor don't support?

schisamo:
@scotthain Yep. The unit tests should run the same on every platform...if they don't any tests should be moved to the functional or acceptance suites.

scotthain:
Eh. Ok. This makes me super uncomfortable but I'm sure y'alls talked it through...

schisamo:
@scotthain It shouldn't make you uncomfortable. This is just a case of running the right test at the right gate. This is not the appropriate gate to run unit tests and it is just slowing builds down and giving a false sense of coverage.

tyler-ball:
:+1:

lamont-granquist:
Yeah, I wouldn't be surprised to find out that we had 'unit' tests that only ran on solaris or aix, i know we've got ones that only run on windows.  I'm not sure that this will make anything run that much faster as well, since unit tests are inherently faster.

Overall I think I'm only 'meh' either way, though.  We can declare the platform-pinned to be functional/integration tests and that'll solve that issue.  We'll also get unit test coverage with both windows and linux which is what is mostly going to go wrong from a practical standpoint.

btm:
@schisamo can we still have them run once a day, maybe on a "daily build" ;) ?

schisamo:
@btm I love when you troll me from afar.

lamont-granquist:
it might be useful to update this ticket for posterity with the numbers on how much of an effect this had

schisamo:
@lamont-granquist Great idea. I'll be sure to do that after we have some data.

PR #4569, opened 2016-02-12
Fix databag globbing issues for chef-solo on windows

### LINKED ISSUE #4194: Chef-solo is broken on windows

jaym:
Backslashes should not be passed to Dir, as it does not work
when globbing in many cases.

Created new function that can be used with dir on windows.
`escape_glob_dir` should be used when the result is going
to be passed into Dir.

Fixes #4194
Replaces #4195

### DIFF
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -154,7 +154,7 @@ module ChefConfig
     # This is because only forwardslashes should be used with dir (even for windows)
     def self.escape_glob_dir(*parts)
       path = Pathname.new(join(*parts)).cleanpath.to_s
-      path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\"+x }
+      path.gsub(/[\\\{\}\[\]\*\?]/) { |x| "\\" + x }
     end
 
     def self.relative_path_from(from, to)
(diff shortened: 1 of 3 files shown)


thommay:
Looks like there's a spec failure on windows and then chefstyle updates, but please let's get this merged! :)

thommay:
@lamont-granquist and I have already :+1: the original, so I think you can merge as soon as the above are happy

lamont-granquist:
@jaym it looks to me like the code is correct, but your spec is testing the wrong behavior.  if you just fix the spec you can merge.  :+1:

jaym:
I think we need https://github.com/chef/chef/pull/4569 to fix windows

PR #4570, opened 2016-02-12
do not include source parameter when removing a chocolatey package and ensure source is used on all functional tests

mwrock:
This covers 2 problems:

* Ensures we specify the internal source repo location on chocolatey functional tests
* Remove the `--source` parameter when calling `choco uninstall` since that is illegal. Note that the resource `:remove` action still needs the source for the idempotence check

I ran the tests with a wireshark trace running to ensure nothing was leaking to the public choco repo with the exception of the initial choco.exe install


lamont-granquist:
looks reasonable :+1:

randomcamel:
:+1:

PR #4573, opened 2016-02-12
pull rubocop 0.37.2 into chefstyle

lamont-granquist:
this is from the same ruleset that we had, but the new code catches more
conditions.


lamont-granquist:
@chef/client-core this is entirely from bumping rubocop from 0.35.1 to 0.37.2, there's no actual ruleset changes in chefstyle.

81   Style/SpaceAroundOperators
25   Style/MultilineOperationIndentation
3    Style/ExtraSpacing
3    Style/RedundantFreeze
1    Style/SpaceBeforeFirstArg
--
113  Total

lib/chef/cookbook/cookbook_collection.rb
```diff
@@ -36,7 +36,7 @@ class CookbookCollection < Mash
     def initialize(cookbook_versions = {})
       super() do |hash, key|
         raise Chef::Exceptions::CookbookNotFound, "Cookbook #{key} not found. " <<
-          "If you're loading #{key} from another cookbook, make sure you configure the dependency in your metadata"
+                                                  "If you're loading #{key} from another cookbook, make sure you configure the dependency in your metadata"
```
lamont-granquist: kinda hate this, looking for an alternative.

lamont-granquist:
@chef/client-core looks better any objections?

lib/chef/application/knife.rb
```diff
@@ -63,7 +63,7 @@ class Chef::Application::Knife < Chef::Application
     :long         => "--disable-editing",
     :description  => "Do not open EDITOR, just accept the data as is",
     :boolean      => true,
-    :default       => false
+    :default => false
```
danielsdeleo: Did this get moved back because it was supposed to line up with the stuff above it, but was off by 1?

lib/chef/knife/ssh.rb
```diff
@@ -186,8 +186,8 @@ def get_ssh_attribute(node)
           Chef::Log.debug("Using node attribute #{Chef::Config[:knife][:ssh_attribute]}")
           attribute = Chef::Config[:knife][:ssh_attribute]
         elsif node[:cloud] &&
-              node[:cloud][:public_hostname] &&
-              !node[:cloud][:public_hostname].empty?
```
danielsdeleo: What rule is this? I don't understand the logic of it in this case.

danielsdeleo:
I really don't know what's going on with the spacing of multi-line conditionals, but it seems very counterintuitive.

lib/chef/application/knife.rb
lamont-granquist: Yep that looks like it.

lib/chef/knife/ssh.rb
```diff
@@ -186,8 +186,8 @@ def get_ssh_attribute(node)
...
+            node[:cloud][:public_hostname] &&
+            !node[:cloud][:public_hostname].empty?
           Chef::Log.debug("Using node attribute 'cloud[:public_hostname]' automatically as the ssh target")
```
lamont-granquist: @danielsdeleo if you look down here, you can see the alignment of code inside the block.  the multiline conditional is being indented another level so its clear that the two lines of code there belong to the conditional above it, and not to the code block inside the conditional.

the 'indent' style means that for multiline operator expressions if you want all the arguments to align you generally need to put the first argument on a line by itself:

```ruby
  raise FooBarException,
    "stuff" <<
    "other stuff" <<
    "yet more stuff"
```

So instead you get:

```ruby
  raise FooBarException,
    "stuff" <<
    "other stuff" <<
    "yet more stuff"
```

but by not using 'indent' what rubocop likes to do instead is:

```ruby
      raise SomeReallyLongExceptionClass, "here's the start of a long comment" <<
                                                                     "and here's the continuation where someone tried to break it to make the lines shorter."
```

and that looks kinda awful to me and I'd prefer:

```ruby
      raise SomeReallyLongExceptionClass, "here's the start of a long comment" <<
        "and here's the continuation where someone tried to break it to make the lines shorter."
```

danielsdeleo: Okay, I see what's going on. I agree with you about the `raise SomeReallyLongExceptionClass` example, but I kinda don't like it for multi-line conditionals. OTOH, I think multiline conditionals should nearly always be extracted to a method with a descriptive name, for example in this case we're just checking if the node has a cloud public hostname but have to avoid a possible train wreck, so I would change it to `if has_cloud_hostname?(node)` or similar; likewise for the other cases. So I guess I can live with this in that it won't bite you if you DTRT and make descriptive predicate methods.

lamont-granquist: yeah the `&.` operator in ruby-2.3.0 might help with the long trainwreck guards as well.

lamont-granquist:
test failures are expected, since the version of chefstyle travis pulls has MultilineOperationIndentation with the default enforcement style, can't keep master unbroken, so merging here then heading over to chefstyle to merge there...

PR #4574, opened 2016-02-12
Extend cookbook shadowing deprecation warnings more broadly

lamont-granquist:
Since 2011 we've been warning that cookbooks shadowing was deprecated, but only in the 'knife cookbook upload' command.

See https://github.com/chef/chef/commit/5a9fee8edaede311ac03a15d52fbc66dad83b576

This inserts the warning directly into the CookbookLoader class which should cause it to be emitted every time the code in question is actually exercised.

### DIFF
--- a/lib/chef/cookbook_loader.rb
+++ b/lib/chef/cookbook_loader.rb
@@ -66,11 +66,19 @@ class Chef
       merged_cookbook_paths
     end
 
+    def warn_about_cookbook_shadowing
+      unless merged_cookbooks.empty?
+        Chef::Log.deprecation "The cookbooks: #{merged_cookbooks.join(', ')} exist in multiple places in your cookbook_path. " +
+          "A composite version has been compiled.  In a future version of Chef this behavior will be removed."
+      end
+    end
+
     def load_cookbooks
       preload_cookbooks
       @loaders_by_name.each do |cookbook_name, _loaders|
         load_cookbook(cookbook_name)
       end
+      warn_about_cookbook_shadowing
       @cookbooks_by_name
     end
 
--- a/lib/chef/cookbook_loader.rb
+++ b/lib/chef/cookbook_loader.rb
@@ -69,7 +69,7 @@ class Chef
     def warn_about_cookbook_shadowing
       unless merged_cookbooks.empty?
         Chef::Log.deprecation "The cookbook(s): #{merged_cookbooks.join(', ')} exist in multiple places in your cookbook_path. " +
-          "A composite version has been compiled.  In a future version of Chef this behavior will be removed."
+          "A composite version has been compiled.  This has been deprecated since 0.10.4, in Chef 13 this behavior will be REMOVED."
       end
     end
 
--- a/lib/chef/cookbook_loader.rb
+++ b/lib/chef/cookbook_loader.rb
@@ -85,6 +85,8 @@ class Chef
     def load_cookbook(cookbook_name)
       preload_cookbooks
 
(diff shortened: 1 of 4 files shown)


coderanger:
I thought we removed this feature a long time ago?

lamont-granquist:
it looks like the spec tests use cookbook shadowing and most of them failed...

lamont-granquist:
nope, just deprecated it.  never been removed.

lamont-granquist:
Added the old 2011 commit to the description...  https://github.com/chef/chef/commit/5a9fee8edaede311ac03a15d52fbc66dad83b576

lamont-granquist:
That is very nearly my new-hire date, so we've been deprecating this feature for as long as I've been working here...

thommay:
:+1:

coderanger:
TIL :+1:

lamont-granquist:
@danielsdeleo can you check my logic?

It looks like we allowed calling load_cookbook on the same cookbook multiple times -- particularly in a tested use case of calling load_cookbook on one cookbook and then calling load_cookbooks on the lot of them.  This actually exercised what looks like a bug in the shadow cookbooks detection such that load_cookbook'ing the same non-shadowed cookbook twice would result in detecting a shadowed cookbook.  So I made load_cookbook a bit more idempotent, and should have also made load_cookbooks more idempotent as well.  We'll still get two deprecation warnings if load_cookbooks is called twice (not sure if that is a real use case or not, but it exercises the reloaded-cookbook-misidentified-as-shadowed bug pretty nicely).  I could move the idempotency check before preload_cookbooks as well, but it looks like that has a (kinda goofy) idempotency check in it already...

lib/chef/cookbook_loader.rb
```diff
@@ -66,17 +66,27 @@ def merged_cookbook_paths # for deprecation warnings
       merged_cookbook_paths
     end
 
+    def warn_about_cookbook_shadowing
+      unless merged_cookbooks.empty?
+        Chef::Log.deprecation "The cookbook(s): #{merged_cookbooks.join(', ')} exist in multiple places in your cookbook_path. " +
+          "A composite version has been compiled.  In a future version of Chef this behavior will be removed."
```
danielsdeleo: Do knife users see this twice?

danielsdeleo:
Had to re-read through it to see how all the pieces fit together, but the logic seems correct to me. :+1:

lib/chef/cookbook_loader.rb
lamont-granquist: Probably?  Possibly?  I'm still debating in my head what to do about that.   This has to be Chef::Log and can't use the knife-ui stuff, but the other stuff is knife-ui...

I could just delete the knife-ui stuff but its nice and yellow and attention getting...

danielsdeleo: We could throw in an option to turn it off, in `#initialize` or something. Generally I'm the biggest hater of methods like `def do_something(enable_disable_some_weird_behavior = true)`, but it's easy enough to nuke when 13.0 time comes. Or else you can have `#load_cookbooks_without_shadow_warning` and then have 

```ruby
def load_cookbooks
  result = load_cookbooks_without_shadow_warning
  warn_about_shadowing
  result
end
```

lamont-granquist: hmmm, maybe we should throw in a Chef::Config conditional to remove shadow cookbooks now and have it default to enabled...

that'd really be the right way to do this, but not at 6pm on a friday before a 3 day weekend...

PR #2752, opened 2015-01-13
Make sure that we have a valid object before calling close!

cmluciano:
This fixes [#2171](https://github.com/opscode/chef/issues/2171) when a
tempfile cannot be created or proper filesystem permissions do not
exist. This allows the proper exception to be thrown.

PR #4579, opened 2016-02-16
Make sure we have a valid object before calling close!

### LINKED ISSUE #2171: Clarify the probable cause of tempfile creation failure during cookbook sync
Could we make the possible problem here more clear during an exception? 
https://github.com/opscode/chef/blob/master/lib/chef/http.rb#L385

This result from chef-client while trying to sync cookbook files is not very helpful and requires digging to discover the real issue

    Unexpected Error: 
    ----------------- 
    NoMethodError: undefined method `close!' for nil:NilClass

The real problem is almost always going to be

1. /tmp filesystem full/inodes used up. Presents as 

        bash-3.2# touch /tmp/chef-rest20141003-29957-1iayumf.lock 
        touch: cannot touch `/tmp/chef-rest20141003-29957-1iayumf.lock': No space left on device

2. /tmp filesystem read-only

thommay:
This is a fix up of #2752 with added tests.
Fixes #2752, fixes #2171 cc @chef/client-core

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -406,7 +406,7 @@ class Chef
       tf.close
       tf
     rescue Exception
-      tf.close!
+      tf.close! if tf
       raise
     end
 
--- a/spec/unit/http_spec.rb
+++ b/spec/unit/http_spec.rb
@@ -28,6 +28,8 @@ end
 
 describe Chef::HTTP do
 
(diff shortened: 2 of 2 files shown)


smurawski:
:+1:

lamont-granquist:
:+1:

PR #4583, opened 2016-02-16
Split group members on commas

### LINKED ISSUE #4294: group#members naively coerced into an Array

thommay:
Fixes #4294, fixes #4322 cc @chef/client-core
Replaces #4322

### DIFF
--- a/lib/chef/resource/group.rb
+++ b/lib/chef/resource/group.rb
@@ -55,7 +55,7 @@ class Chef
       end
 
       def members(arg = nil)
-        converted_members = arg.is_a?(String) ? arg.split(',') : arg
+        converted_members = arg.is_a?(String) ? arg.split(",") : arg
         set_or_return(
           :members,
           converted_members,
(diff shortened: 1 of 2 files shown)


smurawski:
:+1:

lamont-granquist:
:+1:

PR #4587, opened 2016-02-16
Add client keys handling to ChefFS

jrunning:
Supports work to bring chef-zero closer to parity with chef-server. See chef/chef-zero#199.

### DIFF
--- a/lib/chef/chef_fs/config.rb
+++ b/lib/chef/chef_fs/config.rb
@@ -31,6 +31,7 @@ class Chef
       # out here:
       INFLECTIONS = {
         "acls" => "acl",
+        "client_keys" => "client_key",
         "clients" => "client",
         "cookbooks" => "cookbook",
         "cookbook_artifacts" => "cookbook_artifact",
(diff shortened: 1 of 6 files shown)


lib/chef/chef_fs/config.rb
```diff
@@ -65,7 +66,7 @@ class Config
       # upgrade/migration of older Chef Servers, so they should be considered
       # frozen in time.
 
-      CHEF_11_OSS_STATIC_OBJECTS = %w{cookbooks cookbook_artifacts data_bags environments roles}.freeze
+      CHEF_11_OSS_STATIC_OBJECTS = %w{cookbooks cookbook_artifacts data_bags environments roles client_keys}.freeze
```
danielsdeleo: This should be in Chef 12 objects if possible.

danielsdeleo:
:+1: modulo the static/everything/hosted_everything stuff, and getting a happy travis run after rebase.

danielsdeleo:
@chef/client-core ready for review.

jkeiser:
@jrunning code looks good; need tests though. `spec/integration/knife/download_spec` (as well as upload and list) are generally the most useful ones.

danielsdeleo:
@jkeiser We don't expect the behavior of knife upload/download/etc. to change at this time (though we did find a few bugs running the tests here with the Chef Zero changes, so that was good). Multi-key is a Chef 12 feature and we're not ready to convert knife to Chef 12 mode (and deal with all the test updates and such) right now, we just want to get Chef Zero at parity with Erchef.

jkeiser:
Ah! OK then. All good. :+1:

jkeiser:
Er wait--no, we need to make sure knife download and upload still work. These things will show up at the top level. Can you point knife at Manage or a Chef 12 server, and do `knife list /` and see if `client_keys` shows up?

danielsdeleo:
FTR, Chatted with John in a different venue. `client_keys` doesn't show up in `knife list /` (I tested this against hosted chef, and created some extra keys for clients just to be sure) and though that would be a valuable thing, it's outside of the scope of what we're trying to do. Therefore there aren't any behavior changes in chef/knife so no new tests.

PR #4590, opened 2016-02-17
add an issue template

lamont-granquist:
See https://github.com/dear-github/dear-github/issues/125

I think this keeps the ask fairly doable, while making sure to
cover the bases that we need covered 90+% of the time.


issue_template.md
```diff
@@ -0,0 +1,31 @@
...
+
+The complete output of the chef-client run or a link to a gist of it, if there is one.
+
+The debug output (chef-client -l debug) may be useful, but please link to a gist, or truncate it.
+
+## Stacktrace
+
+Please include the stacktrace.out output or link to a gist of it, if there is one.
+
+### NOTE: CHEF CLIENT BUGS ONLY
```
coderanger: Knife issues go here too, but not sure how to word that. (also solo :trollface: )

smurawski:
:+1:

issue_template.md
lamont-granquist: I think we can fix that the first time users ask 'where do kinife bugs go?'  There's also chef-apply and chef-zero should go in the chef-zero repo, while local-mode should go here.

Its also particularly hard because knife plugins need to go to the plugins and knife bugs need to go here, so wording would need to get that right, too.

But really 90+% of the misfiled bugs are server bugs winding up in this repo.  I'd prefer to keep it simple until someone posts that they're confused.

coderanger: Fair nuff. Maybe just drop the ONLY?

coderanger:
Actual docs on the feature are up https://github.com/blog/2111-issue-and-pull-request-templates

danielsdeleo:
Hope this doesn't turn into a bikeshed, but do we want to direct feature requests to https://feedback.chef.io and keep them out of issues?

lamont-granquist:
I don't think so, the feature requests aren't really a problem.

When I think of the issues that I have with issues that people cut they are:

- cutting chef server issues against chef/chef (and its almost exclusively chef server issues for some reason)
- no indication of which chef-client version the problem is against
- no stacktrace
- no output of the error
- poor repro steps

This is good enough to address all five of those, and should encourage an increase in the quality of submissions.  Doubt its perfect, and don't think it needs to be.

coderanger:
Some thoughts from the Rust team that I think are relevant https://github.com/rust-lang/rust/pull/31732

lamont-granquist:
Yeah, I don't mind constructing some barriers to entry.  We've already got 384 issues, with 120 triaged to "Help Wanted" and another 41 triaged to "Low" priority in Accepted Minor.  That's nearly half of the open issues we'll almost certainly never get to.  Doesn't could the submissions that are poor in some way or another that we've just closed.  If the template causes someone to not submit an issue and nobody else comes along and hits the same problem and can be bothered to fill out the template, then it almost certainly just triaged itself down into one of those buckets.

I think this template also mostly fits our use case as well.  We could argue that "expected behavior/actual behavior" is what we want to ask for, since there might not be an actual exception/error thrown -- but I'd like to avoid reading walls of text describing the problem.  I'd rather see a chef-client run where it does / doesn't do the thing in question.  And then that output gives me more clues about what is going on than a text description.  So I really want to encourage copying output into tickets...

lamont-granquist:
tweaked a few things, i went ahead and added a sentence about what is/isn't covered, added a pointer to chef-dk as well.

also converted to upcase as well to fit the other md files, and then got trolled by a case insensitive filesystem and had to switch to linux to fix that...

danielsdeleo:
:+1: let's go with it and we can change it if we need to.

PR #4591, opened 2016-02-17
zypper multipackage performance fix

lamont-granquist:
obvious fix to remember the results so we don't re-compute every single time on access

### DIFF
--- a/lib/chef/provider/package/zypper.rb
+++ b/lib/chef/provider/package/zypper.rb
@@ -54,7 +54,7 @@ class Chef
         end
 
         def versions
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@thommay can i just whack merge?  can't test this via public API behavior so would have to throw in a brittle unit test around this one method in order to make sure it only calls out X numbers of times, and we just don't bother spec testing these memoizing methods anywhere else, and its pretty unlikely to regress now its fixed...

thommay:
:+1:

PR #4592, opened 2016-02-17
skip uid check for aix in user add functional tests

mwrock:
Bug in AIX auto increments UIDs. So after 200 builds, this test fails.

### DIFF
--- a/spec/functional/resource/user/useradd_spec.rb
+++ b/spec/functional/resource/user/useradd_spec.rb
@@ -281,24 +281,17 @@ describe Chef::Provider::User::Useradd, metadata do
         end
       end
 
-      context "when a system user is specified" do
+      context "when a system user is specified", skip: aix? do
         let(:system) { true }
         let(:uid_min) do
-          case ohai[:platform]
-          when "aix"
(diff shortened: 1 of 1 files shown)


mwrock:
cc @scotthain @jaym

scotthain:
:+1:

spec/functional/resource/user/useradd_spec.rb
```diff
@@ -281,24 +281,17 @@ def try_cleanup
         end
       end
 
-      context "when a system user is specified" do
+      context "when a system user is specified", skip: (ohai[:platform] == "aix") do
```
thommay: skip: aix?

thommay: ie, use the platform helper, rather than ohai directly.

thommay:
one trivial thing but :+1: either way

spec/functional/resource/user/useradd_spec.rb
mwrock: oh thats way better. I'll change to that and merge after it goes green. Thanks!

PR #4598, opened 2016-02-18
Need to run apt update on EC2 ubuntu images for acceptance tests

tyler-ball:
\cc @mwrock @chef/client-maintainers 

We need to run apt becuase thats what the [cookbook also does](https://github.com/chef-cookbooks/git/blob/master/.kitchen.yml#L18-L21)

### DIFF
--- a/acceptance/top-cookbooks/.kitchen.git.yml
+++ b/acceptance/top-cookbooks/.kitchen.git.yml
@@ -1,6 +1,7 @@
 suites:
   - name: git-default
-    run_list: ["recipe[git]"]
+    # Ubuntu images need to run apt update
+    run_list: ["recipe[apt]","recipe[git]"]
     includes: [ubuntu-14.04]
   - name: git-source
     run_list: ["recipe[git::source]"]


mwrock:
:+1:

chefsalim:
:+1:

lamont-granquist:
it'd be preferable to run the apt_update resource early in the run_list and exercise that going forwards

tyler-ball:
@lamont-granquist doesn't the default recipe do that? https://github.com/chef-cookbooks/apt/blob/master/recipes/default.rb#L62

lamont-granquist:
Nope, that's an execute resource running 'apt-get update'.

I mean use the built-in resource to do this:  https://github.com/chef/chef/pull/4422/files

mwrock:
To (hopefully) get the pipeline green, I'm gonna merge as is. Based on what I can tell here, we'd want to either add the `apt_update` resource to the git cookbook or add a recipe in acceptance that runs the resource prior to the others.

PR #4601, opened 2016-02-18
adding pry and pry-byebug to dev dependencies

mwrock:


coderanger:
Any particular reason? If it's for interactive debugging of failed tests it might be good to put it in the Gemfile in a group that CI won't install, just to save some cycles. Might not matter much given that travis caches gems though.

mwrock:
Yeah just for dev. I'll move to Gemfile.

coderanger:
pry is, I think pure Ruby but byebug has some C heft to it so makes me pay slightly more attention to CI overhead :)

mwrock:
glad someone is paying attention :)

mwrock:
stripped out commented external test entries but I can put them back if they were left commented for a reason.

coderanger:
I think that's still under the `:test` group, but upon further investigation this is moot for now. Appveyor doesn't use `--without development` and Travis caches compiled gems so benefit on either.

:shipit:

lamont-granquist:
you could add pry-stack-explorer as well.   we already ship all three of these in the omnibus-chef package, so we know they're portable enough.

mwrock:
added that @lamont-granquist had not heard of that - looks very cool.

jkeiser:
:+1:

PR #4605, opened 2016-02-19
Allow chef-fips to deploy older devkit and build using environmental build tools.

ksubrama:


omnibus/config/projects/chef.rb
```diff
@@ -59,6 +71,12 @@
 dependency "shebang-cleanup"
 dependency "version-manifest"
 dependency "openssl-customization"
+dependency "rb-readline"
```
ksubrama: You only need this on windows right?  Other platforms use libedit or whatever else we compile in.

ksubrama:
Superceded by https://github.com/chef/chef/pull/4607

PR #4608, opened 2016-02-20
Update gem_package with version constraint on :upgrade action

### LINKED ISSUE #3867: gem_package "upgrade" action may be broken/confusing
According to the [docs](https://docs.chef.io/resource_gem_package.html#actions) `action :upgrade` is supposed to do the following:

"Install a package and/or **ensure that a package is the latest version.**"

Here's my problem: If you set the version constraint on a package to, say, `'~> 1.0'` and you have version `1.1.6` installed but there is a version `1.1.7` available, no upgrade is performed, because:

    DEBUG: chef_gem[your_gem] found installed gem your_gem version 1.1.6 matching your_gem (~> 1.0)
    DEBUG: chef_gem[your_gem] your_gem has no candidate_version to upgrade to
    DEBUG: chef_gem[your_gem] no versions to upgrade - nothing to do

Is this the expected behavior? Because it seems different from what the documentation implies (and altogether too close to the behavior already provided by `:install`).

wjordan:
(fixes #3867)

Additional tests might be needed to cover various edge-cases I haven't thought of, but I think something like this should allow fuzzy-version-constrained gems to continue to be upgraded when new versions are released.

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -256,9 +256,16 @@ class Chef
         options ? " #{options}" : ""
       end
 
-      # this is public and overridden by subclasses (rubygems package implements '>=' and '~>' operators)
-      def target_version_already_installed?(current_version, new_version)
-        new_version == current_version
+      # this is public and overridden by subclasses
+      def target_version_already_installed?(current_version, target_version)
+        current_version && target_version &&
+          current_version == target_version
+      end
+
+      # this is public and overridden by subclasses
+      # (rubygems package implements '>=' and '~>' operators)
+      def version_requirement_satisfied?(current_version, version_requirement)
+        target_version_already_installed?(current_version, version_requirement)
       end
 
       # @todo: extract apt/dpkg specific preseeding to a helper class
@@ -355,11 +362,12 @@ class Chef
               case action
               when :upgrade
 
-                if !candidate_version
-                  Chef::Log.debug("#{new_resource} #{package_name} has no candidate_version to upgrade to")
+                if target_version_already_installed?(current_version, new_version) ||
+                    target_version_already_installed?(current_version, candidate_version)
+                  Chef::Log.debug("#{new_resource} #{package_name} #{new_version} is already installed")
                   target_version_array.push(nil)
-                elsif current_version == candidate_version
-                  Chef::Log.debug("#{new_resource} #{package_name} the #{candidate_version} is already installed")
+                elsif candidate_version.nil?
+                  Chef::Log.debug("#{new_resource} #{package_name} has no candidate_version to upgrade to")
                   target_version_array.push(nil)
                 else
                   Chef::Log.debug("#{new_resource} #{package_name} is out of date, will upgrade to #{candidate_version}")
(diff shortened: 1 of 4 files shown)


lib/chef/provider/package/rubygems.rb
```diff
@@ -482,7 +482,7 @@ def cleanup_after_converge
 
         def candidate_version
           @candidate_version ||= begin
-                                   if target_version_already_installed?(@current_resource.version, @new_resource.version)
+                                   if action != :upgrade && target_version_already_installed?(@current_resource.version, @new_resource.version)
```
lamont-granquist: this seems incorrect, the candidate_version should be the condidate_version no matter :install or :upgrade.

most likely this conditional should just be dropped.

if that breaks :install then there's probably a bug in the superclass to fix (and i think there is at least one around this kind of thing).

wjordan: Dropping the conditional would cause the spec [`"when determining the candidate version to install" it "does not query for available versions when the current version is the target version"`](https://github.com/wjordan/chef/blob/gem_upgrade_fix_3867/spec/unit/provider/package/rubygems_spec.rb#L494-L501) to fail. The fact that this behavior is explicitly tested makes me believe this implementation was intended, at least for the `:install` case (probably to avoid an extra RubyGems API round-trip when not necessary).

lamont-granquist: Yeah, that's the bug in the superclass I was referring to.

Its really the wrong way to go about solving this problem by lying to the superclass that there's no candidate_version.  The superclass should be smart enough to see that you're trying to `:install` and therefore if the package is installed and satisfies the resource it should never check the candidate_version.

Otherwise this check gets pushed down into *all* of the subclasses and they all have to lie to the superclass, and I'm certain that they do not.  If we want to maintain this hack then really there should be a candidate_version in the superclass that implements this check for all the subclasses, and then the subclasses should call super before adding their implementation (or the superclass should call some wrapper around candidate_version that does the lying when action is :install).

wjordan:
@lamont-granquist I've rebased+updated the PR with an approach that uses the existing `Provider::Package` superclass logic to determine whether or not to check the candidate_version, as suggested. Two main changes were needed to make this refactoring possible while preserving existing behavior:

- First, I added an extra `version_requirement_satisfied?` method separate from the existing `target_version_already_installed?` method. The implementation is identical in the superclass default implementation, but is different when the package subclass supports version-ranges as RubyGems does. The `:install` action checks `version_requirement_satisfied?` for a fuzzy (range-version) match, while `:upgrade` checks `target_version_already_installed?` for an exact match.

- Second, I wrapped `Package#candidate_version` in a lazy-initialization object, in order to preserve the existing behavior where the RubyGems API would only be queried if/when `candidate_version` was actually needed to satisfy the action. Some of the existing package-provider logic needed to be re-ordered slightly so that `candidate_version` was never queried when not necessary.
  - I couldn't figure out how to modify the multipackage code to support this approach, so the lazy-initialization is currently only applied where `#use_multipackage_api?` is false.
  - Alternatives would be to either move the lazy-initialization code into the rubygems-specific implementation, or remove the lazy-initialization of `candidate_version` altogether and just accept the possibility of extra unnecessary RubyGems API queries.

lib/chef/provider/package.rb
```diff
@@ -468,7 +476,32 @@ def package_name_array
 
       # @return [Array] candidate_version(s) as an array
       def candidate_version_array
-        [ candidate_version ].flatten
+        use_multipackage_api? ?
+          [ candidate_version ].flatten :
+          [ LazyObject.new { candidate_version } ]
+      end
```
lamont-granquist: would it be possible to fix this so that when we're using the multipackage api here that we get back an array of lazy objects?

PR #4611, opened 2016-02-22
RFC062 exit status chef client

smurawski:
- [x] Define and validate approved exit codes

- [x] Deprecate unapproved exit codes

- [x] Control exit status behavior via Chef::Config

- [x] Test each of the exit code responses with and without the exit status validation

- [x] Re-plumb REBOOT_NEEDED checks.

- [x] Support SIGTERM/SIGINT

To test the behavior, 

```ruby
bundle exec rspec ./spec/integration/client/exit_code_spec.rb
```

There are two integration tests marked pending that should be safe but appear not to cancel the reboot on travis in time.

### DIFF
--- a/lib/chef/application/apply.rb
+++ b/lib/chef/application/apply.rb
@@ -137,11 +137,11 @@ class Chef::Application::Apply < Chef::Application
 
   def read_recipe_file(file_name)
     if file_name.nil?
-      Chef::Application.fatal!("No recipe file was provided", 1)
+      Chef::Application.fatal!("No recipe file was provided", Chef::Exceptions::RecipeNotFound.new)
     else
       recipe_path = File.expand_path(file_name)
       unless File.exist?(recipe_path)
-        Chef::Application.fatal!("No file exists at #{recipe_path}", 1)
+        Chef::Application.fatal!("No file exists at #{recipe_path}", Chef::Exceptions::RecipeNotFound.new)
       end
       recipe_fh = open(recipe_path)
       recipe_text = recipe_fh.read
@@ -183,7 +183,7 @@ class Chef::Application::Apply < Chef::Application
     else
       if !ARGV[0]
         puts opt_parser
-        Chef::Application.exit! "No recipe file provided", 1
+        Chef::Application.exit! "No recipe file provided", Chef::Exceptions::RecipeNotFound.new
       end
       @recipe_filename = ARGV[0]
       @recipe_text, @recipe_fh = read_recipe_file @recipe_filename
@@ -208,7 +208,7 @@ class Chef::Application::Apply < Chef::Application
       raise
     rescue Exception => e
       Chef::Application.debug_stacktrace(e)
-      Chef::Application.fatal!("#{e.class}: #{e.message}", 1)
+      Chef::Application.fatal!("#{e.class}: #{e.message}", e)
     end
   end
 
--- a/lib/chef/application/exit_code.rb
+++ b/lib/chef/application/exit_code.rb
@@ -1,0 +1,226 @@
+#
+# Author:: Steven Murawski (<smurawski@chef.io>)
+# Copyright:: Copyright 2016, Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  class Application
+
+    # These are the exit codes defined in Chef RFC 062
+    # https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md
+    class ExitCode
+
+      # -1 is defined as DEPRECATED_FAILURE in RFC 062, so it is
+      # not enumerated in an active constant.
+      #
+      VALID_RFC_062_EXIT_CODES = {
+        SUCCESS: 0,
+        GENERIC_FAILURE: 1,
+        SIGINT_RECEIVED: 2,
+        SIGTERM_RECEIVED: 3,
+        REBOOT_SCHEDULED: 35,
+        REBOOT_NEEDED: 37,
+        REBOOT_FAILED: 41,
+        AUDIT_MODE_FAILURE: 42,
+      }
+
+      DEPRECATED_RFC_062_EXIT_CODES = {
+        DEPRECATED_FAILURE: -1,
+      }
+
+      class << self
+
+        def normalize_exit_code(exit_code = nil)
+          if normalization_not_configured?
+            normalize_legacy_exit_code_with_warning(exit_code)
+          elsif normalization_disabled?
+            normalize_legacy_exit_code(exit_code)
+          else
+            normalize_exit_code_to_rfc(exit_code)
+          end
+        end
+
+        def enforce_rfc_062_exit_codes?
+          !normalization_disabled? && !normalization_not_configured?
+        end
+
+        def notify_reboot_exit_code_deprecation
+          return if normalization_disabled?
+          notify_on_deprecation(reboot_deprecation_warning)
+        end
+
+        def notify_deprecated_exit_code
+          return if normalization_disabled?
+          notify_on_deprecation(deprecation_warning)
+        end
+
+        private
+
+        def normalization_disabled?
+          Chef::Config[:exit_status] == :disabled
+        end
+
+        def normalization_not_configured?
+          Chef::Config[:exit_status].nil?
+        end
+
+        def normalize_legacy_exit_code_with_warning(exit_code)
+          normalized_exit_code = normalize_legacy_exit_code(exit_code)
+          unless valid_exit_codes.include? normalized_exit_code
+            notify_on_deprecation(deprecation_warning)
+          end
+          normalized_exit_code
+        end
+
+        def normalize_legacy_exit_code(exit_code)
+          case exit_code
+          when Fixnum
+            exit_code
+          when Exception
+            lookup_exit_code_by_exception(exit_code)
+          else
+            default_exit_code
+          end
+        end
+
+        def normalize_exit_code_to_rfc(exit_code)
+          normalized_exit_code = normalize_legacy_exit_code_with_warning(exit_code)
+          if valid_exit_codes.include? normalized_exit_code
+            normalized_exit_code
+          else
+            VALID_RFC_062_EXIT_CODES[:GENERIC_FAILURE]
+          end
+        end
+
+        def lookup_exit_code_by_exception(exception)
+          if sigint_received?(exception)
+            VALID_RFC_062_EXIT_CODES[:SIGINT_RECEIVED]
+          elsif sigterm_received?(exception)
+            VALID_RFC_062_EXIT_CODES[:SIGTERM_RECEIVED]
+          elsif normalization_disabled? || normalization_not_configured?
+            if legacy_exit_code?(exception)
+              # We have lots of "Chef::Application.fatal!('', 2)
+              # This maintains that behavior at initial introduction
+              # and when the RFC exit_status compliance is disabled.
+              VALID_RFC_062_EXIT_CODES[:SIGINT_RECEIVED]
+            else
+              VALID_RFC_062_EXIT_CODES[:GENERIC_FAILURE]
+            end
+          elsif reboot_scheduled?(exception)
+            VALID_RFC_062_EXIT_CODES[:REBOOT_SCHEDULED]
+          elsif reboot_needed?(exception)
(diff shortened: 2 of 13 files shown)


smurawski:
Looks like the reboot cancel isn't quick enough on linux and there is a bug in the reboot pending check.

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,181 @@
...
+    # These are the exit codes defined in Chef RFC 062
+    # https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md
+    class ExitCode
+
+      # -1 is defined as DEPRECATED_FAILURE in RFC 062, so it is
+      # not enumerated in an active constant.
+      #
+      VALID_RFC_062_EXIT_CODES = {
+        SUCCESS: 0,
+        GENERIC_FAILURE: 1,
```
lamont-granquist: we should keep 2 and 3 for SIGINT and SIGTERM, RFC needs to be updated to keep them.

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,218 @@
...
+          return true if skip_validation
+          return true if valid_exit_codes.include? exit_code
+
+          notify_on_deprecation(deprecation_warning)
+          allow_deprecated_exit_code
+        end
+
+        def allow_deprecated_exit_code
+          ## After the new exit code behavior default changes
+          ## TODO: change the check to just
```
tyler-ball: It looks like you can go ahead and do this TODO. Or am I missing something?

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+
+          after do
+            if Chef::Platform.windows?
+              shell_out!("shutdown /a")
+            else
+              shell_out!("shutdown -c")
+            end
+          end
+
+          xit "exits with SUCCESS, 0" do
```
tyler-ball: `xit` - typo?

tyler-ball: There is another one further down too

lib/chef/application/exit_code.rb
smurawski: I believe we'll want to announce the new optional behavior, then after a few releases we can default it.  Doing the todo now will start issuing deprecation warnings on current exit behavior.

spec/integration/client/exit_code_spec.rb
smurawski: No, they work when run locally, but Travis doesn't cancel the reboot in time.  I could add an explanation in a comment as to why they are pended.

smurawski: @tyler-ball I changed the integration tests so they were safe to run and added another scenario.

spec/functional/rebooter_spec.rb
```diff
@@ -101,5 +101,6 @@ def test_rebooter_method(method_sym, is_windows, expected_reboot_str)
         end
       end
     end
+
```
btm: bonus

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+            run_chef_client_and_expect_exit_code 41
+          end
+        end
+
+      end
+
+    end
+
+  end
+
```
btm: Can you delete the extra lines between end statements and between context statements in this file?

lib/chef/exceptions.rb
```diff
@@ -66,6 +76,9 @@ class User < RuntimeError; end
     class Group < RuntimeError; end
     class Link < RuntimeError; end
     class Mount < RuntimeError; end
+    class Reboot < RuntimeError; end
+    class RebootPending < RuntimeError; end
```
btm: Reboot and RebootPending aren't really failures are they? Subclass from Exception instead of RuntimeError so if you're catching RuntimeError for a failure you wouldn't catch these?

spec/integration/client/exit_code_spec.rb
smurawski: will do

lib/chef/exceptions.rb
smurawski: Thanks - will do.

spec/integration/client/exit_code_spec.rb
thommay: i think it's clearer to do `it "...", pending: true` than `xit`; easier to grep for too.

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+  when_the_repository "does has exit_status enabled" do
```
thommay: `does have` or just `has`

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+        end
+
+      end
+
+      context "with a recipe" do
+
+        context "which throws an error" do
+          before { file "cookbooks/x/recipes/default.rb", "raise 'BOOM'" }
+
+          it "exits with GENERAL_FAILURE, 1" do
```
thommay: `GENERIC_FAILURE` not `GENERAL_FAILURE`.

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+
+      context "with a library" do
+
+        context "which cannot be loaded" do
+          before do
+            file "cookbooks/x/recipes/default.rb", ""
+            file "cookbooks/x/libraries/error.rb", "require 'does/not/exist'"
+          end
+
+          it "exits with GENERAL_FAILURE, 1" do
```
thommay: here too

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+            setup_client_rb
+            run_chef_client_and_expect_exit_code 1
+          end
+        end
+
+        context "with a recipe which calls Chef::Application.fatal with a non-RFC exit code" do
+          before { file "cookbooks/x/recipes/default.rb", "Chef::Application.fatal!('BOOM', 123)" }
+
+          it "exits with the GENERAL_FAILURE exit code, 1" do
```
thommay: and here

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+            setup_client_rb
+            run_chef_client_and_expect_exit_code 1
+          end
+        end
+
+        context "with a recipe which calls Chef::Application.exit with a non-RFC exit code" do
+          before { file "cookbooks/x/recipes/default.rb", "Chef::Application.exit!('BOOM', 231)" }
+
+          it "exits with the GENERAL_FAILURE exit code, 1" do
```
thommay: and here

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+
+      context "with a library" do
+
+        context "which cannot be loaded" do
+          before do
+            file "cookbooks/x/recipes/default.rb", ""
+            file "cookbooks/x/libraries/error.rb", "require 'does/not/exist'"
+          end
+
+          it "exits with GENERAL_FAILURE, 1" do
```
thommay: GENERIC

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+            file "cookbooks/x/recipes/default.rb", <<-RECIPE
+control_group "control group without top level control" do
+  it "should fail" do
+    expect(2 - 2).to eq(1)
+  end
+end
+RECIPE
+          end
+
+          it "exits with GENERAL_FAILURE, 1" do
```
thommay: here too

spec/integration/client/exit_code_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+        end
+
+      end
+
+      context "with a recipe" do
+
+        context "which throws an error" do
+          before { file "cookbooks/x/recipes/default.rb", "raise 'BOOM'" }
+
+          it "exits with GENERAL_FAILURE, 1" do
```
thommay: think that's all of them

btm:
the relationship between validate_exit_code`, `resolve_exit_code`, and `resolve_exit_code_from_exception` is a hard to grok.

`validate_exit_code` doesn't look like it has a case where it raises that the exit code is invalid. Consider renaming this to `normalize_exit_code`.

I'd completely drop `resolve_exit_code` and push most of it's logic into `normalize_exit_code` as 2 of it's 3 lines aren't resolving anything and it doesn't look like `resolve_exit_code` is called directly elsewhere.

I'd paint the `resolve_exit_code_from_exception` bikeshed by naming it `lookup_exit_code_by_exception`. It's clearer to me that lookup would look in a list than resolve.

So you could end up with something like this:
```
 def normalize_exit_code(exit_code = nil)
  case exit_code
  when nil
    nil
  when is_a?(Fixnum)
    exit_code
  when is_a?(Exception)
    lookup_exit_code_by_exception(exit_code)
  else
    default_exit_code
  end
end
```

Note that I didn't use `valid?` here. The path to that seems inconsistent?

smurawski:
Thanks for the feedback @btm - I'll try something like that - I agree the naming could use some work. ;)

smurawski:
@btm I think my refactor addresses most of your naming and pathing concerns.  Would appreciate if you give it another look over when you have a chance.

btm:
:+1:

lib/chef/application/apply.rb
```diff
@@ -137,11 +137,11 @@ def parse_json
 
   def read_recipe_file(file_name)
     if file_name.nil?
-      Chef::Application.fatal!("No recipe file was provided", 1)
+      Chef::Application.fatal!("No recipe file was provided")
```
mwrock: should you be specifying ` Chef::Exceptions::RecipeNotFound.new` as yo do below?

lib/chef/application/apply.rb
```diff
@@ -137,11 +137,11 @@ def parse_json
...
     else
       recipe_path = File.expand_path(file_name)
       unless File.exist?(recipe_path)
-        Chef::Application.fatal!("No file exists at #{recipe_path}", 1)
+        Chef::Application.fatal!("No file exists at #{recipe_path}")
```
mwrock: Should this too use ` Chef::Exceptions::RecipeNotFound.new`?

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,226 @@
...
+          "Per RFC 062 (https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md)" \
+          ", when a reboot is requested Chef Client will exit with an exit code of 35, REBOOT_SCHEDULED." \
+          " To maintain the current behavior (an exit code of 0), you will need to set `exit_status` to" \
+          " `:disabled` in your client.rb"
+        end
+
+        def default_exit_code
+          if normalization_disabled? || normalization_not_configured?
+            return DEPRECATED_RFC_062_EXIT_CODES[:DEPRECATED_FAILURE]
+          else
```
mwrock: could be
```
if enforce_rfc_062_exit_codes?
  VALID_RFC_062_EXIT_CODES[:GENERIC_FAILURE]
else
  return DEPRECATED_RFC_062_EXIT_CODES[:DEPRECATED_FAILURE]
end
```

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,226 @@
...
+          begin
+            Chef.log_deprecation(message)
+          rescue Chef::Exceptions::DeprecatedFeatureError
+            # Have to rescue this, otherwise this unhandled error preempts
+            # the current exit code assignment.
+          end
+        end
+
+        def deprecation_warning
+          "Chef RFC 62 (https://github.com/chef/chef-rfc/master/rfc062-exit-status.md) defines the" \
```
mwrock: nit: `062` instead of `62` for consistency

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,226 @@
...
+            VALID_RFC_062_EXIT_CODES[:GENERIC_FAILURE]
+          end
+        end
+
+        def lookup_exit_code_by_exception(exception)
+          if sigint_received?(exception)
+            VALID_RFC_062_EXIT_CODES[:SIGINT_RECEIVED]
+          elsif sigterm_received?(exception)
+            VALID_RFC_062_EXIT_CODES[:SIGTERM_RECEIVED]
+          elsif normalization_disabled? || normalization_not_configured?
```
mwrock: could be `!enforce_rfc_062_exit_codes?`

lib/chef/application/exit_code.rb
```diff
@@ -0,0 +1,226 @@
...
+            e.is_a? Chef::Exceptions::SigInt
+          end
+        end
+
+        def sigterm_received?(exception)
+          resolve_exception_array(exception).any? do |e|
+            e.is_a? Chef::Exceptions::SigTerm
+          end
+        end
+
```
mwrock: I think you could get avoid needing each of these blah_received? methods by either expanding `resole_exception_array` or creating a new method: `resolves_to?(exception, type)`

mwrock:
I'm :+1: but I did leave a few comments and questions that are not blocking. I initially found the ExitCodes class a little hard to follow and left some comments about drying it up but after I groced the class I retracted the comments and preferred it as is. Getting this added to the client will be a very nice enhancement.

PR #4612, opened 2016-02-22
fixes exe package downloads

mwrock:
A regression introduced [here](https://github.com/chef/chef/commit/09df8eea8bdbca4b7f1f0db43033d8d473860131) has caused `exe` based windows packages to fail when pulling from a remote source. This is because of the added `define_resource_requirements` method which ensures that a `source` is included for `exe` packages. 

This is called before converging and therefore also before any remote source is downloaded. If the source is remote, we do not download it (and rightly so) until we determine that the package is elligible to be installed. However, in order to determine the type of package (in this case if the package is not an `msi`), we must beable to read the package installer file. If we do not have that locally, this fails with an error stating that the file cannot be found.

However, we don't really need to know the exact installer type at this point. We just need to know if it is an `msi` or not. Fortunately, we can tell that without actually downloading the file. So this PR adds a `msi?` method that simply checks that and uses it in the `define_resource_requirements` method.

### DIFF
--- a/lib/chef/provider/package/windows.rb
+++ b/lib/chef/provider/package/windows.rb
@@ -36,7 +36,7 @@ class Chef
 
         def define_resource_requirements
           requirements.assert(:install) do |a|
-            a.assertion { new_resource.source unless package_provider == :msi }
+            a.assertion { new_resource.source unless msi? }
             a.failure_message Chef::Exceptions::NoWindowsPackageSource, "Source for package #{new_resource.name} must be specified in the resource's source property for package to be installed because the package_name property is used to test for the package installation state for this package type."
           end
         end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/windows.rb
```diff
@@ -254,6 +254,15 @@ def validate_content!
           end
         end
 
+        def msi?
+          return true if new_resource.installer_type == :msi
+
+          if source_location.nil?
+            inferred_registry_type == :msi
+          else
+            ::File.basename(source_location).split(".").last.downcase == "msi"
```
jaym: `::File.extname(source_location) == '.msi'`

mwrock: yeah it should. probably a copy/paste oversight. Thats better.

jaym:
:+1: minor comment

danielsdeleo:
:+1:

PR #4607, opened 2016-02-20
Make the chef pipeline build chef + optional fips mode

ksubrama:
This also makes the compiler/ruby version explicit.

PR #4616, opened 2016-02-23
Build ruby from source on all platforms.

ksubrama:


ksubrama:
This is the alternative to https://github.com/chef/chef/pull/4607 until the refactoring/cleanup story is figured out.

@schisamo @jaym @btm @chef/omnibus-maintainers

omnibus/config/projects/chef.rb
```diff
@@ -51,15 +51,27 @@
...
+
+# Global FIPS override flag.
+if windows? || rhel?
+  override :fips, enabled: true
+end
 
 dependency "preparation"
-dependency "chef"
-dependency "pry"
+dependency "rb-readline" if windows?
```
jaym: should this go closer to the end in the `if windows?` block?

ksubrama: I was optimizing for build-time.  I was putting everything that could go before "chef" in front of it.

omnibus/config/projects/chef.rb
```diff
@@ -51,15 +51,27 @@
 # Chef Release version pinning
 override :chef, version: "local_source"
 override :ohai, version: "master"
+override :"rb-readline", version: "v0.5.3"
```
schisamo: Is this override needed? Shouldn't just make the `default_version` of the `rb-readline` software def `v0.5.3`.

jaym:
:+1:

omnibus/config/projects/chef.rb
schisamo: @jaym That seems like a good idea.

omnibus/config/projects/chef.rb
ksubrama: I left the default version pointing to master like a lot of other software definitions do in omnibus-software.  It felt like the natural "default" for anyone wanting rb-readline and makes the version used here explicit.  I can swap it out if that's more sensible.

omnibus/config/projects/chef.rb
```diff
@@ -51,15 +51,27 @@
...
+
+# Global FIPS override flag.
+if windows? || rhel?
+  override :fips, enabled: true
+end
```
scotthain: Does this imply that we're always building fips on windows and rhel or is this overridden in a pipeline/env var somewhere else?

jaym: we only build on windows and rhel. Nothing special in the pipeline

btm: Yeah, we're always building OpenSSL FIPS support on those platforms, but you have to enable it for the client.

btm:
:+1: I restarted the appveyor job that died. wasn't clear what happened.

omnibus/config/projects/chef.rb
scotthain: Ah ok - I was concerned about the performance implications of always building support in, but I assume that isn't an issue unless it's enabled.

ksubrama: We think there isn't any.  The we ask ruby to build internal md5 but I don't anticipate anybody using chef to crack shitty encryption from the 90s and caring about raw md5 performance.

schisamo:
@ksubrama Do we need to wait for https://github.com/chef/omnibus-software/pull/593 to merge before merging this PR?

jaym:
omnibus-software needs to go first

ksubrama:
Yeah - and I'm waiting for a good chef build on jenkins - i'm getting weird cache checksum failures again and that's annoying.

omnibus/config/projects/chef.rb
lamont-granquist: can we use rb-readline on unix as well?  that might get us onto ruby 2.2.x on unix

scotthain: I want to say there was some issue with the native gem for rb-readline on the wacky platforms, but it's worth a shot. I could be very wrong.

omnibus/config/projects/chef.rb
ksubrama: Looks like @jkeiser is actually changing this in https://github.com/chef/omnibus-software/pull/584

lamont-granquist: if there's no reason to override it, then you're just creating meaningless technical debt

omnibus/config/projects/chef.rb
ksubrama: rb-readline is purely ruby.  It was made to avoid native gems.  I've verified by installing it without a compiler on my path on windows (unless some sneaky magic was happening).

That said...  *shrug*  who knows.  We can try turning it on everywhere.  If it works, it would save us some effort trying to build libedit for ruby.

scotthain: Ohhh that's right, by bad. :+1:

btm: Let's get something merged, and later we can consider the benefits of that so we aren't biting off a ton of work for maybe's in our final week.

PR #4617, opened 2016-02-23
Autofixing new Perf cops in 0.37.2

lamont-granquist:
6 Performance/Casecmp
  18 Performance/Detect
   1 Performance/RangeInclude
  27 Performance/RedundantBlockCall
   6 Performance/RedundantMatch
   5 Performance/RedundantMerge
  18 Performance/StringReplacement

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,7 +30,7 @@ group(:development, :test) do
   # gem 'chefstyle', github: 'chef/chefstyle'
   gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
 
-  gem "ruby-shadow", platforms: :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
+  gem "ruby-shadow", platforms: :ruby unless RUBY_PLATFORM.downcase =~ /(aix|cygwin)/
 
   # For external tests
   #  gem 'chef-zero', github: 'chef/chef-zero'
--- a/lib/chef/chef_fs/data_handler/container_data_handler.rb
+++ b/lib/chef/chef_fs/data_handler/container_data_handler.rb
@@ -15,10 +15,16 @@ class Chef
           return key == "containername"
         end
 
+        # Verify that the JSON hash for this type has a key that matches its name.
+        #
+        # @param object [Object] JSON hash of the object
+        # @param entry [Chef::ChefFS::FileSystem::BaseFSObject] filesystem object we are verifying
+        # @yield  [s] callback to handle errors
+        # @yieldparam [s<string>] error message
         def verify_integrity(object, entry)
           base_name = remove_dot_json(entry.name)
           if object["containername"] != base_name
-            yield("Name in #{entry.path_for_printing} must be '#{base_name}' (is '#{object['name']}')")
+            yield("Name in #{entry.path_for_printing} must be '#{base_name}' (is '#{object['containername']}')")
           end
         end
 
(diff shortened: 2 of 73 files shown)


lamont-granquist:
travis is angrysauce about unrelated issues with a bad github-changelog-generator release

@chef/client-core thoughts?

lib/chef/provider/service/upstart.rb
```diff
@@ -62,7 +62,7 @@ def initialize(new_resource, run_context)
           end
 
           platform, version = Chef::Platform.find_platform_and_version(run_context.node)
-          if platform == "ubuntu" && (8.04..9.04).include?(version.to_f)
+          if platform == "ubuntu" && (8.04..9.04).cover?(version.to_f)
```
danielsdeleo: What is this?

Gemfile
```diff
@@ -30,7 +30,7 @@ group(:development, :test) do
   # gem 'chefstyle', github: 'chef/chefstyle'
   gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
 
-  gem "ruby-shadow", platforms: :ruby unless RUBY_PLATFORM.downcase.match(/(aix|cygwin)/)
+  gem "ruby-shadow", platforms: :ruby unless RUBY_PLATFORM.downcase =~ /(aix|cygwin)/
```
danielsdeleo: Is there more explanation of this? I'm assuming it only kicks in if you don't use the match data return... Wondering how much of an actual improvement this is.

lib/chef/chef_fs/data_handler/container_data_handler.rb
```diff
@@ -15,10 +15,10 @@ def preserve_key?(key)
           return key == "containername"
         end
 
-        def verify_integrity(object, entry, &on_error)
+        def verify_integrity(object, entry)
```
danielsdeleo: I'm a bit hesitant about these ones. I personally like the `yield` keyword for non-logical reasons, but having the block argument makes the method signature more descriptive, especially for longer methods where the `yield` may be harder to spot on a quick reading. You can of course fix that with yard/rdoc/whatever but I don't know if we find that worthwhile.

lamont-granquist: The block argument (alone) is 4x slower than just `yield`ing with no proc:

https://github.com/JuanitoFatas/fast-ruby#proccall-and-block-arguments-vs-yieldcode

Sorta feel like documentation should be fixed with YARD.

Gemfile
lamont-granquist: Not a lot:  https://github.com/JuanitoFatas/fast-ruby#regexp-vs-stringmatch-vs-string-code-

I kind of like =~ better than match -- i've probably been infected by writing too many lines of Perl in my life though....

lib/chef/provider/service/upstart.rb
lamont-granquist: IDK, good eye...

lamont-granquist: wow, this one is a lot faster:

https://github.com/JuanitoFatas/fast-ruby#range

although for Float, i'd think `#include?` would have to be an alias to `#cover?`

lib/chef/chef_fs/data_handler/container_data_handler.rb
thommay: I tend to agree we should do this with documentation. @lamont-granquist could you add some minimal YARDage to this PR for the things that used to take blocks?

Gemfile
thommay: Personally, I think =~ is a bit more intention-revealing, but it's a bit of a wash.

thommay:
I'm :+1: modulo documentation for things that are block-y

danielsdeleo:
I'm okay with these then. TBH I don't think the block thing will actually change much for us unless we're calling these things in a tight enough loop to trigger significantly more GCs but as long as we add doc to make it clear that these methods require block arguments, it's not really a big deal.

:+1:

PR #4621, opened 2016-02-24
Fix the Chocolatey-missing error again

randomcamel:
The previous implementation (in `define_resource_requirements`) failed acceptance testing because `choco.exe` gets called before that, in `load_current_resource`, so the detection needs to be a bit lower-level.

This includes a spec that replicates the acceptance test (by calling `load_current_resource`).

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -150,12 +150,12 @@ EOS
           @choco_exe ||=
               # if this check is in #define_resource_requirements, it won't get
               # run before choco.exe gets called from #load_current_resource.
-              exe_path = ::File.join(choco_install_path, "bin", "choco.exe")
-              if !::File.exists?(exe_path.to_s)
-                raise Chef::Exceptions::MissingLibrary, CHOCO_MISSING_MSG
-              else
-                exe_path
-              end
+              exe_path = ::File.join(choco_install_path.to_s, "bin", "choco.exe")
+          if !::File.exist?(exe_path)
+            raise Chef::Exceptions::MissingLibrary, CHOCO_MISSING_MSG
+          else
+            exe_path
+          end
         end
 
         # lets us mock out an incorrect value for testing.
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -147,15 +147,13 @@ EOS
         #
         # @return [String] full path of choco.exe
         def choco_exe
-          @choco_exe ||=
+          @choco_exe ||= begin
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/chocolatey.rb
```diff
@@ -152,11 +147,15 @@ def check_resource_semantics!
...
-          @choco_exe ||= ::File.join(
-            choco_install_path,
-            "bin",
-            "choco.exe"
-            )
+          @choco_exe ||=
+              # if this check is in #define_resource_requirements, it won't get
+              # run before choco.exe gets called from #load_current_resource.
+              exe_path = ::File.join(choco_install_path.to_s, "bin", "choco.exe")
+          if !::File.exist?(exe_path)
```
mwrock: looks like some indentation issues here. Also `unless` may be clearer than `if !`.

lamont-granquist: yeah that'll be a rubocop violation at some point in the not too distant future...

lib/chef/provider/package/chocolatey.rb
```diff
@@ -152,11 +147,15 @@ def check_resource_semantics!
...
+            raise Chef::Exceptions::MissingLibrary, CHOCO_MISSING_MSG
+          else
+            exe_path
+          end
```
lamont-granquist: could we do this a bit more like...

```
  @choco_exe ||=
    begin
      exe_path = ...
      raise unless exist?
      exe_path
   end
```

its both cleaner to read, and also if we've set choco_exe successfully we don't need to re-test if we should raise again the next time we hit it, and the return value buried inside the if is just a bit hard to read (much like this run on sentence...)

lib/chef/provider/package/chocolatey.rb
```diff
@@ -152,11 +147,15 @@ def check_resource_semantics!
...
         def choco_exe
-          @choco_exe ||= ::File.join(
-            choco_install_path,
-            "bin",
-            "choco.exe"
-            )
+          @choco_exe ||=
+              # if this check is in #define_resource_requirements, it won't get
+              # run before choco.exe gets called from #load_current_resource.
+              exe_path = ::File.join(choco_install_path.to_s, "bin", "choco.exe")
```
randomcamel: Somehow `exe_path` comes up `nil` here sometimes.

lib/chef/provider/package/chocolatey.rb
randomcamel: Yep, good call.

mwrock:
:+1:

btm:
:+1:

PR #4622, opened 2016-02-24
Don't print fips warning in test environment

ksubrama:


ksubrama:
@jaym @btm

btm:
Actually, I would like to remove that warning altogether now. We've another thing we want to do, but there's a complete set of functionality.

btm:
somewhere else we print that we're running in FIPS mode too.

ksubrama:
Yeah we do print that we've entered FIPS mode elsewhere.

jaym:
:+1:

btm:
:+1:

PR #4625, opened 2016-02-25
Update path_helper.rb

scalp42:
Steps:

- cd in a directory `/foo`
- delete the directory `/foo`
- run chef-client


lamont-granquist:
Personally I think we should raise anyway when `pwd` disappears.  Just pushes the problem down to the first call that tries to access PWD or ".." or do the "walk up the filesystem to find the config file" dance.   Then we have to protect that, etc...

lamont-granquist:
@chef/client-core thoughts?

thommay:
I agree with Lamont here; the problem will surface at some point down the line. I'm gonna close this, but thanks for the submission.

PR #4626, opened 2016-02-25
ensure paths maintain utf-8ness in non ascii encodings

mwrock:
On windows (I'm assuming this is limited to windows) when using anon ascii encoding like (CP850) and with non ASCII characters (like like an umlaut `\x81`) in `ENV['PATH']` or `ENV['HOME']`, chef runs fail.

I have found 2 key areas that lead to this failure:
* `path_helper`
* `path_sanity`

`path_helper` fails because `Dir.home` has a `Windows-1252` encoding and when a character like `\x81` is present, checking the directory with `File.exists?` fails due to an invalid byte sequence.

`path_sanity` creates a path that is far from sane if a character like `\x81` is in the path. This results in creating a `ENV['PATH']` with a `Encoding:8BIT` that can wreak havoc in several possible places throughout a chef run.

Note that this can only be reproduced when the node has a non US langiage pack. I used German.

### DIFF
--- a/chef-config/lib/chef-config/path_helper.rb
+++ b/chef-config/lib/chef-config/path_helper.rb
@@ -224,7 +224,7 @@ module ChefConfig
       paths = paths.map { |home_path| home_path.gsub(path_separator, ::File::SEPARATOR) if home_path }
 
       # Filter out duplicate paths and paths that don't exist.
-      valid_paths = paths.select { |home_path| home_path && Dir.exists?(home_path) }
+      valid_paths = paths.select { |home_path| home_path && Dir.exists?(home_path.force_encoding("utf-8")) }
       valid_paths = valid_paths.uniq
 
       # Join all optional path elements at the end.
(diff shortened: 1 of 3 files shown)


randomcamel:
Do we want to limit this to Windows? I want to say "no," but OTOH there have been many challenging iterations of trying to make these path functions work on both Unix and Windows.

mwrock:
The path_helper only applies to windows but the path_sanity function change would be applicable to both. My guess is that the way the system handles locale and encoding is fundamentally different on windows and linux and will not really be helped/affected by this. However, I can test a chef run on linux with this change to ensure nothing breaks. Not sure if our functional travis suites test a full chef run and `enforce_path_sanity` is called in `Client#run`.

randomcamel:
:+1:

mwrock:
FYI: Also tested this on a linux converge and nothing died.

danielsdeleo:
:+1:

PR #4629, opened 2016-02-25
List all of the unignored files when loading a cookbook

danielsdeleo:
My tactical goal here is to restore some functionality that broke in ChefDK/policyfiles. Previously, ChefDK would do a simple file copy to copy cookbooks when doing `chef export` (which is what the test kitchen provisioner uses to get the cookbook code on the guest VM). That didn't honor chefignore, so in some cases folks could end up copying a VM into itself. Therefore, that behavior was changed to use `CookbookVersionLoader`. However, this had the side-effect of only copying files that belong to a "cookbook segment," so test files were not copied. So the goal here is to make it possible for ChefDK to get the list of all of a cookbook's files (there are some caveats to this which are documented in code comments), so that test/spec/etc. files will be copied.

@chef/client-core This is "working" now, but has the downside of double-globbing the cookbook directory. Looking for some thoughts about a possible fix for that, which I will highlight in an inline code comment.

### DIFF
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -230,6 +230,7 @@ class Chef
             next if dir_relpath.to_s.start_with?(".")
 
             Dir.glob(File.join(fs_entry, "**/*"), File::FNM_DOTMATCH).each do |file|
+              next if File.directory?(file)
               file = Chef::Util::PathHelper.cleanpath(file)
               name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
               cookbook_settings[:all_files][name] = file
@@ -261,7 +262,6 @@ class Chef
         glob_pattern = File.join(Chef::Util::PathHelper.escape_glob_dir(cookbook_path, category_dir), "**", glob)
         select_files_by_glob(glob_pattern, File::FNM_DOTMATCH).each do |file|
           file = Chef::Util::PathHelper.cleanpath(file)
-          next if File.directory?(file)
           name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
           cookbook_settings[category][name] = file
         end
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -81,13 +81,7 @@ class Chef
 
         load_all_files
 
-        # TODO:
-        # 1. Can we safely pick out the segment files with fnmatch against the
-        #    file names in all_files (?)
-        # 2. If so, can we avoid running chefignore twice by running it against
-        #    all_files before segmenting?
-        #
-        # See: https://shane.io/2014/07/13/sobbing-with-ruby-file-globbing.html
+        remove_ignored_files
 
         load_as(:attribute_filenames, "attributes", "*.rb")
         load_as(:definition_filenames, "definitions", "*.rb")
@@ -98,8 +92,6 @@ class Chef
         load_recursively_as(:resource_filenames, "resources", "*.rb")
         load_recursively_as(:provider_filenames, "providers", "*.rb")
         load_root_files
-
-        remove_ignored_files
 
         if empty?
           Chef::Log.warn "Found a directory #{cookbook_name} in the cookbook path, but it contains no cookbook files. skipping."
@@ -302,10 +294,8 @@ class Chef
       end
 
       def remove_ignored_files
-        cookbook_settings.each_value do |file_list|
-          file_list.reject! do |relative_path, full_path|
-            chefignore.ignored?(relative_path)
-          end
+        cookbook_settings[:all_files].reject! do |relative_path, full_path|
+          chefignore.ignored?(relative_path)
         end
       end
 
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -253,8 +253,8 @@ class Chef
       end
 
       def load_recursively_as(category, category_dir, glob)
-        file_spec = File.join(Chef::Util::PathHelper.escape_glob(cookbook_path, category_dir), "**", glob)
-        Dir.glob(file_spec, File::FNM_DOTMATCH).each do |file|
+        glob_pattern = File.join(Chef::Util::PathHelper.escape_glob(cookbook_path, category_dir), "**", glob)
+        select_files_by_glob(glob_pattern, File::FNM_DOTMATCH).each do |file|
           file = Chef::Util::PathHelper.cleanpath(file)
           next if File.directory?(file)
           name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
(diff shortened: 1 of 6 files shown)


lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -78,6 +79,16 @@ def load
...
 
+        load_all_unignored_files
+
+        # TODO:
+        # 1. Can we safely pick out the segment files with fnmatch against the
+        #    file names in all_unignored_files (?)
+        # 2. If so, can we avoid running chefignore twice by running it against
+        #    all_unignored_files before segmenting?
+        #
+        # See: https://shane.io/2014/07/13/sobbing-with-ruby-file-globbing.html
```
danielsdeleo: I seem to recall getting bitten by the differences between `Dir.glob` and `File.fnmatch` in the past. The blog post I linked says that the differences just come down to the two functions having different default option flags, which can be changed. But it still feels a bit risky. Thoughts?

lamont-granquist:
Did you just build the first part of `all_files` from https://github.com/chef/chef-rfc/blob/master/rfc067-cookbook-segment-deprecation.md?

And would `all_unignored_files` be better called `all_files`?

spec/unit/cookbook_version_spec.rb
```diff
@@ -77,77 +79,85 @@
...
-      @cookbook[:library_filenames]     = Dir[File.join(@cookbook_root, "libraries", "**", "*.rb")]
-      @cookbook[:resource_filenames]    = Dir[File.join(@cookbook_root, "resources", "**", "*.rb")]
-      @cookbook[:provider_filenames]    = Dir[File.join(@cookbook_root, "providers", "**", "*.rb")]
-      @cookbook[:root_filenames]        = Array(File.join(@cookbook_root, "README.rdoc"))
-      @cookbook[:metadata_filenames]    = Array(File.join(@cookbook_root, "metadata.json"))
-    end
+    let(:cookbook_paths_by_type) do
+      {
+        # Dunno if the paths here are representitive of what is set by CookbookLoader...
+        all_unignored_files:  Dir[File.join(cookbook_root, "**", "*.rb")],
```
lamont-granquist: Yeah, I think I'd rather call it `all_files` and consider the concept of ignoring to be implicit in how the cookbook loader should work.

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -212,6 +224,24 @@ def chefignore
         @chefignore ||= Chefignore.new(File.basename(cookbook_path))
       end
 
+      def load_all_unignored_files
+        # List all directories that are not dotdirs, then list everything in
+        # those, *including* dotfiles and dotdirs.
+        # Finally, list all the _files_ at the root, including dotfiles, and
+        # merge those in.
```
lamont-granquist: could probably simplify to just state that dotdirs (but not dotfiles) in the root are magically excluded even if they're not chefignored.

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -212,6 +224,24 @@ def chefignore
...
+        Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(cookbook_path), "*")).each do |dir|
```
lamont-granquist: you should use escape_glob_dir here instead of escape_glob (check the pull request that added escape_glob_dir for the long story...)

lamont-granquist: and as a rule anywhere you see `Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(...)))` that signature implies that the inside of the loop takes forward slashes just fine (because of the File.join outside of the escape_glob) so that should always be converged to escape_glob_dir and everything inside of the Dir.glob should be using forward slashes.

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -212,6 +224,24 @@ def chefignore
...
+          next unless File.directory?(dir)
+
+          Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(dir), "**/*"), File::FNM_DOTMATCH).each do |file|
+            name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
+            cookbook_settings[:all_unignored_files][name] = file
+          end
+        end
+        load_root_files
+        cookbook_settings[:all_unignored_files].merge!(cookbook_settings[:root_filenames])
+      end
```
lamont-granquist: I think I'd start more like this:

```ruby
def load_all_files
  root_entries = Dir.glob(File.join(Chef::Util::PathHelper.escape_glob_dir(cookbook_path), "*"), File::FNM_DOTMATCH).reject { |f|  File.basename(f).start_with?(".") && File.directory?(f) }

  root_entries.each do |path|
    Dir.glob(File.join(Chef::Util::PathHelper.escape_glob_dir(path), "**/*"), File::FNM_DOTMATCH).each do |file|
      name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
      cookbook_settings[:all_files][name] = file
    end
  end
end
```

Then instead of loading the root_files in you could go the other way and extract the root_files from the all_files if you want to optimize...

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -212,6 +224,24 @@ def chefignore
         @chefignore ||= Chefignore.new(File.basename(cookbook_path))
       end
 
+      def load_all_unignored_files
+        # List all directories that are not dotdirs, then list everything in
+        # those, *including* dotfiles and dotdirs.
+        # Finally, list all the _files_ at the root, including dotfiles, and
+        # merge those in.
+        Dir.glob(File.join(Chef::Util::PathHelper.escape_glob(cookbook_path), "*")).each do |dir|
+          dir = Chef::Util::PathHelper.cleanpath(dir)
```
lamont-granquist: not necessary to cleanpath just to pass to File.directory? and its what you don't want with Dir.glob later

lib/chef/cookbook/cookbook_version_loader.rb
lamont-granquist: bit of a run-on sentence there probably `root_entries.reject!` which would also avoid the extra allocation...

lib/chef/cookbook/cookbook_version_loader.rb
lamont-granquist: I think that's the algorithm I'd prefer to see.  Construct all the files, apply chefignore, then extract the segmented files from the superset...

danielsdeleo: Yeah, after a day or two, I feel like I'm being too paranoid being worried about this. I'll take an in-depth look at the tests and see if I need to make a cookbook with every combination of dot-file/dot-dir/etc. to be more confident.

lib/chef/cookbook/cookbook_version_loader.rb
danielsdeleo: I'll take a look, didn't know that existed. Thanks for the tip.

lib/chef/cookbook/cookbook_version_loader.rb
danielsdeleo: Copy-pasta'd and modified from the other similar methods here, but yeah it's obviously doing work and then undoing it two lines later. I'll remove.

lib/chef/cookbook/cookbook_version_loader.rb
danielsdeleo: I've been mulling this over, and wonder if this weird exception is really necessary. On the plus side, it prevents you from enumerating a `.git` or `.svn`, which is nice. Also, if you don't have a `chefignore` you still get your `.git` dir ignored for free and I think vagrant+VMWare fusion puts stuff in a dotdir also, which would be nice to auto-ignore. OTOH, if you actually wanted to keep a dotdir in the root, there would be nothing you could do. Not sure what that use case would be though; I can't think of anything.

spec/unit/cookbook_version_spec.rb
danielsdeleo: Cool, I wanted your opinion on this specific thing. I'm maybe 51/49 in favor of `all_files` after thinking about it some more.

lib/chef/cookbook/cookbook_version_loader.rb
lamont-granquist: yeah, IDK either.  obviously you can't undo it if you ever wanted.  but all the things i can think of (`.git`, `.svn`, `.kitchen`) are always awful to actually upload anywhere.

there is the use case of something like `files/homedirs/lamont/.ssh/` but that's covered.

and if we went ahead and uploaded dotdirs that weren't in chefignore i think we'd create a flag day where everyone had to update their chefignore files to ignore all the dotdirs they didn't have to care about before...  lots of "why the @#$@#$ would i ever want to upload .git???"

so i think this makes sense, and i think 'fixing' it would be a chef-13 breaking change kind of thing where we'd need really good reasons/issues to put everyone else through the trouble....

spec/unit/cookbook_version_spec.rb
lamont-granquist: Yeah, I'm a bit more than 51% -- not super wild about it, but it does feel like its the thing we decided to name in that RFC for our $REASONS, so we should do that until we discover its unworkably awkward...

danielsdeleo:
@chef/client-core Ready for review now

coderanger:
Hell yeah step towards shooting segments in the face. :+1:

This might break some of the Halite synthetic cookbook loading tests, but I can fix them easily.

lamont-granquist:
LGTM :+1:

Issue #4614, opened 2016-02-23
Chef::DataBagItem.to_hash is modifying Chef::DataBagItem.raw_data


rb1whitney:
## Description

Chef::DataBagItem.to_hash to modifying raw_data and causing corruption of data bags to upload to chef server

## Chef Version

We have been experiencing issue with 12.06 but more many versions below this as well.

## Platform Version

Red Hat Enterprise Linux Server release 6.7

## Replication Case

to_hash has a bug and it has been corrupting my data bags. When I use to_hash to compare if a data bag is identical to what is on the server. The chef code modifies the raw_data being compared and causes chef_type and data_bag to be added to my raw_data. The above code is modifying self.raw_data since the object is a pointer and not a copy. 

https://github.com/chef/chef/blame/master/lib/chef/data_bag_item.rb
def to_hash
      result = self.raw_data
      result["chef_type"] = "data_bag_item"
      result["data_bag"] = self.data_bag
      result
end

If you try to Chef::DataBagItem.save with the data_bag item, it will upload this content to Chef Server. For an encrypted data bag, Chef thinks the data is un-encrypted as a result.

knife data bag show test_environment datasource:
Unencrypted data bag detected, ignoring any provided secret options.
chef_type: data_bag_item
data_bag:  test_environment
id:        datasource
password:
  cipher:         aes-256-cbc
  encrypted_data: ...
  
  hmac:           ...
  
  iv:             ...
  
  version:        2
url:
  cipher:         aes-256-cbc
  encrypted_data: ...
  
  hmac:           ...
  
  iv:             ...
  
  version:        2
username:
  cipher:         aes-256-cbc
  encrypted_data: ...
  
  hmac:           ...
  
  iv:             ...
  
  version:        2
## Client Output

None

## Stacktrace

Code to test:
puts "Before: #{local_bag_item.to_json}"
local_bag_item.to_hash
puts "After: #{local_bag_item.to_json}"

Before running to_hash, json output: 

{"name":"test_environment_datasource","json_class":"Chef::DataBagItem","chef_type":"data_bag_item","data_bag":"test_environment","raw_data":{"id":"...","password":{"encrypted_data":"...","hmac":"...","iv":"...","version":2,"cipher":"aes-256-cbc"},"username":{"encrypted_data":"...","hmac":"...","iv":"...","version":2,"cipher":"aes-256-cbc"}}}

[...]

PR #4631, opened 2016-02-25
Fix a bug that was causing DataBagItem.to_hash to mutate the data bag item

itmustbejj:
Fixes issue #4614.

Make a copy of the raw_data (that is then mutated with "data_bag" and "chef_type" keys) to return, rather than the actual raw_data object.

### DIFF
--- a/lib/chef/data_bag_item.rb
+++ b/lib/chef/data_bag_item.rb
@@ -107,9 +107,9 @@ class Chef
     end
 
     def to_hash
-      result = self.raw_data
+      result = self.raw_data.dup
       result["chef_type"] = "data_bag_item"
-      result["data_bag"] = self.data_bag
+      result["data_bag"] = self.data_bag.to_s
       result
     end
 
--- a/spec/unit/data_bag_item_spec.rb
+++ b/spec/unit/data_bag_item_spec.rb
@@ -146,6 +146,8 @@ describe Chef::DataBagItem do
       data_bag_item
     }
 
+    let!(:original_data_bag_keys) { data_bag_item.keys }
+
     let(:to_hash) { data_bag_item.to_hash }
 
     it "should return a hash" do
@@ -163,6 +165,11 @@ describe Chef::DataBagItem do
 
     it "should have the data_bag set" do
       expect(to_hash["data_bag"]).to eq("still_lost")
+    end
+
+    it "should not mutate the data_bag_item" do
+      data_bag_item.to_hash
+      expect(data_bag_item.keys).to eq(original_data_bag_keys)
     end
   end
 


lib/chef/data_bag_item.rb
```diff
@@ -107,9 +107,10 @@ def self.object_name(data_bag_name, id)
     end
 
     def to_hash
-      result = self.raw_data
+      result = self.raw_data.dup
+      #result = self.raw_data
```
coderanger: I don't think you meant to leave this.

lib/chef/data_bag_item.rb
```diff
@@ -107,9 +107,10 @@ def self.object_name(data_bag_name, id)
...
       result["chef_type"] = "data_bag_item"
-      result["data_bag"] = self.data_bag
+      result["data_bag"] = self.data_bag.to_s
```
coderanger: Is this ever not a string? Probably fine either way but good to know when it would be something else.

itmustbejj: I had to change this to a string, when I added the `.dup` to self.raw_data on line 110, the data bag key became a symbol instead of a string and caused a unit test to fail.

lib/chef/data_bag_item.rb
itmustbejj: I definitely did not :)

lamont-granquist:
could probably use a spec test.  you can probably copypasta some code from other examples.  all it has to do is setup a data bag item, then call `#to_hash`, and then validate that the data bag item didn't mutate the raw_data.

danielsdeleo:
Same, :+1: on the code that's there but should have a test.

itmustbejj:
@danielsdeleo @lamont-granquist test added. Good to merge?

coderanger:
:+1:

goacid:
chef 12.10.24 hit the bug again

itmustbejj:
@goacid Can you file a new issue with details to repro it and link to it as a comment on here? The code from this PR is in place in 12.10.24, so I would be curious what your use case is that is still seeing this.

epetrovich:
I dunno what's wrong here. I've added no data to the databags by myself. Just close the editor without any modifications on create/edit.

$ knife  data bag create  testbag testapp --secret-file ~/.ssh/encrypted_data_bag_secret
Created data_bag[testbag]
WARN: Auto inflation of JSON data is deprecated. Please pass in the class to inflate or use #edit_hash at /opt/chefdk/embedded/lib/ruby/2.1.0/forwardable.rb:183:in `edit_data'
Created data_bag_item[testapp]
$ knife  data bag edit  testbag testapp --secret-file ~/.ssh/encrypted_data_bag_secret
Encrypting data bag using provided secret.
Saved data_bag_item[testapp]
$ knife  data bag show  testbag testapp --secret-file ~/.ssh/encrypted_data_bag_secret
ERROR: Chef::Exceptions::ValidationFailed: Option data_bag's value {"encrypted_data"=>"ddG+zJRBIvZfD7dtl9hq4lgC4SeUGqRNsYvZddvC7vc=\n", "iv"=>"rsw+NRuKz/VC98mCXoC2eQ==\n", "version"=>1, "cipher"=>"aes-256-cbc"} does not match regular expression /^[\-[:alnum:]_]+$/
$ knife  data bag edit  testbag testapp --secret-file ~/.ssh/encrypted_data_bag_secret
ERROR: Chef::Exceptions::ValidationFailed: Option data_bag's value {"encrypted_data"=>"ddG+zJRBIvZfD7dtl9hq4lgC4SeUGqRNsYvZddvC7vc=\n", "iv"=>"rsw+NRuKz/VC98mCXoC2eQ==\n", "version"=>1, "cipher"=>"aes-256-cbc"} does not match regular expression /^[\-[:alnum:]_]+$/
$ ruby --version
ruby 2.1.8p440 (2015-12-16 revision 53160) [x86_64-darwin13.0]
$ chef --version
Chef Development Kit Version: 0.16.28
chef-client version: 12.12.15
delivery version: master (921828facad8a8bbbd767368bfc72f19bd30e7bd)
berks version: 4.3.5
kitchen version: 1.10.2

PR #4633, opened 2016-02-25
Allow MD5 only for cookbook upload / checksum in FIPS mode

jaym:
This patch requires explicitly asking that MD5 be patched in for FIPS. The tests that used MD5 but didn't require it were modified to use SHA256 instead. The tests that need it include a context that enables it.

This disables upgrading the cache control data for remote file in fips mode as it requires MD5. Since there have been no releases of this feature, you cannot possibly need to upgrade your cache control data file.


btm:
This is on hold for a few weeks until we decide if we're doing a bigger piece of work that would negate the need to do this. We want to keep the branch around for now in case we still want to do this though.

PR #4638, opened 2016-02-27
if no module name is found for a valid dsc resource default to PSDesiredStateConfiguration

mwrock:
When converging a `dsc_resource`, we call `get-dscresource` to determine if the resource is present and if it belongs to a module. Some resources report no module associated but in these cases we should default to using `PSDesiredStateConfiguration` as the module to prevent some modules from failing.

This also allows functional tests to run on RTM wmf 5 with LCM enabled.

### DIFF
--- a/lib/chef/provider/dsc_resource.rb
+++ b/lib/chef/provider/dsc_resource.rb
@@ -110,7 +110,7 @@ class Chef
                   "sure that it shows up when running Get-DscResource"
               when 1
                 if found[0]["Module"].nil?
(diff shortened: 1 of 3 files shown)


mwrock:
cc @smurawski

smurawski:
:+1:

chefsalim:
LGTM :+1:

tyler-ball:
:+1:

Looks like this is behavior changing - should only go in a minor version bump?

PR #4640, opened 2016-02-29
Support libraries with sub directories

thommay:
:construction: For Discussion :construction:
Some widely used chef-solo cookbooks, like edelight/chef-solo-search, have subdirectories in their libraries directory.
The libraries directory can also contain non-ruby files (in the chef-solo-search case, the treetop search parser from chef 10).
This PR adds support for this, and would be necessary for chef-zolo support.

### DIFF
--- a/lib/chef/chef_fs/file_system/chef_server/cookbook_dir.rb
+++ b/lib/chef/chef_fs/file_system/chef_server/cookbook_dir.rb
@@ -53,7 +53,7 @@ class Chef
             :attributes => { :ruby_only => true },
             :definitions => { :ruby_only => true },
             :recipes => { :ruby_only => true },
-            :libraries => { :ruby_only => true },
+            :libraries => { :recursive => true },
             :templates => { :recursive => true },
             :files => { :recursive => true },
             :resources => { :ruby_only => true, :recursive => true },
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -86,7 +86,7 @@ class Chef
         load_as(:attribute_filenames, "attributes", "*.rb")
         load_as(:definition_filenames, "definitions", "*.rb")
         load_as(:recipe_filenames, "recipes", "*.rb")
-        load_recursively_as(:library_filenames, "libraries", "*.rb")
+        load_recursively_as(:library_filenames, "libraries", "*")
         load_recursively_as(:template_filenames, "templates", "*")
         load_recursively_as(:file_filenames, "files", "*")
         load_recursively_as(:resource_filenames, "resources", "*.rb")
--- a/lib/chef/run_context/cookbook_compiler.rb
+++ b/lib/chef/run_context/cookbook_compiler.rb
@@ -186,6 +186,7 @@ class Chef
 
       def load_libraries_from_cookbook(cookbook_name)
         files_in_cookbook_by_segment(cookbook_name, :libraries).each do |filename|
+          next unless File.extname(filename) == ".rb"
           begin
             Chef::Log.debug("Loading cookbook #{cookbook_name}'s library file: #{filename}")
             Kernel.load(filename)
--- a/spec/integration/knife/chef_repository_file_system_spec.rb
+++ b/spec/integration/knife/chef_repository_file_system_spec.rb
@@ -231,6 +231,10 @@ EOM
 /cookbooks/cookbook1/files/c/e.json
 /cookbooks/cookbook1/libraries/
 /cookbooks/cookbook1/libraries/a.rb
+/cookbooks/cookbook1/libraries/b.json
+/cookbooks/cookbook1/libraries/c/
+/cookbooks/cookbook1/libraries/c/d.rb
+/cookbooks/cookbook1/libraries/c/e.json
 /cookbooks/cookbook1/providers/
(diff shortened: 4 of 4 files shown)


thommay:
@chef/client-core discussion, vomiting, etc appreciated.

jkeiser:
Question: in what order will these library files be loaded?

coderanger:
Do we need to do anything to handle Vim/Emacs temp files or is that handled at a higher level?

coderanger:
Also we should double check that this works correctly thought a full `knife cookbook upload && chef-client` cycle on the file transfer.

lamont-granquist:
supporting chef-solo-search under chef-zero makes my brain kind of asplodey...

i don't see any real issue with this, however, and it moves us more towards tarball-like cookbooks...

coderanger:
@lamont-granquist Another example is https://github.com/target/f5-bigip-cookbook/tree/master/libraries

lamont-granquist:
for that one we only technically need the recursive fix to chef_fs, though.

thommay:
@lamont-granquist yeah, it's moderately crazy but if we want to support people's workflows unchanged this needs to happen.
@coderanger need to check, thanks!
@jkeiser right now I guess it's alphabetically but with the ability to require from deeper in the directory structure as long as the search path is set correctly

jkeiser:
@thommay after testing it, this does in fact cause us to automatically load all library files, including those under subdirectories. I'd love if we could avoid that behavior--but, what does `chef-solo` do?

jkeiser:
A concern here is that we could actually change the behavior of cookbooks that worked before, by starting to load some library files. I have no idea, however, if that will actually happen--was it possible for people to upload cookbooks with nested libraries before this? How did those avoid getting loaded?

jkeiser:
`chef-solo` loads all cookbooks as well (just tested).

jkeiser:
I can confirm that if you have a cookbook with nested libraries, and you `knife cookbook upload` said cookbook, and then run it, nested libraries are indeed run.

So @thommay, you are clear on the "should we autoload these, and when" front.

lamont-granquist:
we're already shipping all libraries files recursively and loading them all in the real non-chef-zero client aren't we?  

the recursive loading fix is only in chef_fs and only applies to chef-zero i think?

coderanger:
I don't think so, and I'm not sure this is the right behavior either. From talking to someone that was trying to use the F5-bigip cookbook, the behavior that cookbook expects is that only files directly in `libraries/` will be autoloaded, and chef-solo-search seems to be the same. Currently the nested files don't end up loaded to the server, so they don't get sync'd down to clients either. I think at least :)

lamont-granquist:
So the recursive/nested files should have been getting sync'd with knife cookbook upload and chef-client's cookbook synchronizer starting with 12.0.0:

56246eef11d1c578f2bc9dbc92259565964f03f3

The cookbook compiler code hasn't changed since 2012, and if I'm reading it right it just goes through the manifest, so will autoload libraries in nested directories:

https://github.com/chef/chef/commit/20fd1383020261c3756b54c26f4c2ea9e652d43c

Is it possible that the F5-bigip cookbook has not been updated for chef 12.0.0?

/me goes and attempts to verify behavior against hosted...

coderanger:
I think we can take the chef-solo-search code as evidence that currently solo does not recursively auto-load.

lamont-granquist:
So I just did this:

```
# cat sk_test/libraries/foo/bar.rb
puts "LOADED by #{caller[0]}"
```

And changed the debug in the cookbook compiler to a load and got this:

```
[2016-03-02T11:41:21-08:00] WARN: Loading cookbook sk_test's library file: /var/chef/cache/cookbooks/sk_test/libraries/foo/bar.rb
LOADED by /opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.6.0/lib/chef/run_context/cookbook_compiler.rb:191:in `load'
```

So the client in chef-server mode is synching up nested libraries, and is loading them.

Dropping that behavior would be a breaking change to the client.

lamont-granquist:
I'm also seeing nested libraries loaded by chef-solo 12.x:

```
% chef-solo -j ./dna.json -c ./solo.rb
Starting Chef Client, version 12.6.0
Compiling Cookbooks...
LOADED by /Users/lamont/.rvm/gems/ruby-2.2.4/gems/chef-12.7.2/lib/chef/run_context/cookbook_compiler.rb:191:in `load'
```

Nobody using chef-solo-search has upgraded to chef 12?

jkeiser:
@coderanger @lamont-granquist I tested and we do in fact autoload recursively in both chef-client and chef-solo. I made a cookbook with `libraries/a.rb` and `libraries/b/aa.rb`, and both of them were loaded when I ran an empty recipe from that cookbook.

Since there is parity, however, it's a wash.

jkeiser:
Aaand @lamont-granquist did the same test :)

coderanger:
Huh, I wonder when it changed then. Oh well, happy to maintain as-is then :D

lamont-granquist:
Pretty certain it changed in 12.0.0 via 56246eef11d1c578f2bc9dbc92259565964f03f3

lamont-granquist:
:+1:

smurawski:
:+1:

danielsdeleo:
:+1:

PR #4641, opened 2016-02-29
use File.basename rather than reimplementing

thommay:
cc @chef/client-core am I missing something obvious?

### DIFF
--- a/lib/chef/chef_fs/data_handler/data_handler_base.rb
+++ b/lib/chef/chef_fs/data_handler/data_handler_base.rb
@@ -28,10 +28,7 @@ class Chef
         # Takes a name like blah.json and removes the .json from it.
         #
         def remove_dot_json(name)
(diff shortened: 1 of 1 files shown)


thommay:
Right now this loses the error handling if the name doesn't end in .json, but that feels we're relying on a side effect to enforce file naming.

lamont-granquist:
You could also add `raise "Invalid name #{name}: must end in .json" unless name =~ /\.json$/`

lamont-granquist:
:+1:

lamont-granquist:
wtf is wrong with appveyor?

coderanger:
This would also change things if you feed it a path with `/` in it. The original code leaves it intact, this does not.

coderanger:
Should probably be `File.extname` instead of `basename`.

thommay:
Gonna close this for now, this pattern is all over ChefFS.

PR #4642, opened 2016-02-29
Run the integration tests in FIPS mode

ksubrama:


ksubrama:
@btm @adamedx

jkeiser:
FIPS acceptance tests is over 1.5hrs already, and our limit is 1. We need to fix that before we run integration tests, sadly :( However, if we ran the integration tests simultaneously with the unit+functional tests, it might not increase overall time, in which case this would be fine. This could be done by adding a `fips-integration` to `.kitchen.yml` and running them in those separate instances.

ksubrama:
The FIPS tests with integration seem to be throwing some memory allocation errors.  I have made a cursory investigation but I need to look at it more closely.  Consider this PR WIP until then.  The tests seem to pass locally but something about the way we run it in acceptance is making it unhappy.

I'll update this when I've dug into the issue some more.

ksubrama:
I have finished cleaning up the tests and the log outputs.  Running the tests in parallel brings FIPS overall test timing down to 15 minutes.  This is now ready for review /cc @mwrock @PrajaktaPurohit @randomcamel @btm

chef-config/lib/chef-config/config.rb
```diff
@@ -960,7 +960,9 @@ def self.enable_fips_mode
       require "digest"
       require "digest/sha1"
       require "digest/md5"
+      Digest.send(:remove_const, "SHA1") if Digest.const_defined?("SHA1")
       Digest.const_set("SHA1", OpenSSL::Digest::SHA1)
+      OpenSSL::Digest.send(:remove_const, "MD5") if OpenSSL::Digest.const_defined?("MD5")
```
mwrock: might be worth commenting you are suppressing warning spam

mwrock:
:+1:

PR #4644, opened 2016-02-29
use Pathname#cleanpath instead of Chef::Util::PathHelper.cleanpath

mwrock:
Fixes build failures on windows comparing backslashed paths with forward slashed glob patterns using `::File.fnmatch?`. This now ensures that the paths being compared have forward slashes.

### DIFF
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -231,12 +231,12 @@ class Chef
 
             Dir.glob(File.join(fs_entry, "**/*"), File::FNM_DOTMATCH).each do |file|
               next if File.directory?(file)
-              file = Chef::Util::PathHelper.cleanpath(file)
+              file = Pathname.new(file).cleanpath.to_s
               name = Chef::Util::PathHelper.relative_path_from(@cookbook_path, file)
               cookbook_settings[:all_files][name] = file
             end
           elsif File.file?(fs_entry)
-            file = Chef::Util::PathHelper.cleanpath(fs_entry)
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 gemspec name: "chef"
-
+gem 'pry'
```
lamont-granquist: derp

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -231,12 +231,12 @@ def load_all_files
 
             Dir.glob(File.join(fs_entry, "**/*"), File::FNM_DOTMATCH).each do |file|
               next if File.directory?(file)
-              file = Chef::Util::PathHelper.cleanpath(file)
+              file =  ::Pathname.new(file).cleanpath
```
lamont-granquist: indentation.  i don't think you need the :: in front of Pathname either, i was just being pedantic to be clear i was referring to the stdlib on slack...

lamont-granquist:
:shipit:

lib/chef/cookbook/cookbook_version_loader.rb
```diff
@@ -231,12 +231,12 @@ def load_all_files
 
             Dir.glob(File.join(fs_entry, "**/*"), File::FNM_DOTMATCH).each do |file|
               next if File.directory?(file)
-              file = Chef::Util::PathHelper.cleanpath(file)
```
jkeiser: Need a `.to_s` after each of these two changes so we don't change the type of the stuff in `all_files`.

jkeiser:
:+1:

PR #4646, opened 2016-02-29
Handle negative content length headers too.

coderanger:
This is equivalent to not passing one, we can't check for bad downloads.

For an example, run `curl -ivs -o /dev/null "https://www.isc.org/downloads/file/dhcp-4-3-3-p1/?version=tar-gz"`

Curl displays `Negative content-length: -1, closing after transfer`. I can't find any standard for this behavior, but it seems to be shared by all the browsers I tried. ¯\_(ツ)_/¯

### DIFF
--- a/lib/chef/http/validate_content_length.rb
+++ b/lib/chef/http/validate_content_length.rb
@@ -86,7 +86,7 @@ class Chef
         transfer_encoding = http_response["transfer-encoding"]
         content_encoding  = http_response["content-encoding"]
 
-        if content_length.nil?
+        if content_length.nil? || content_length < 0
           Chef::Log.debug "HTTP server did not include a Content-Length header in response, cannot identify truncated downloads."
           return true
         end
--- a/lib/chef/http/validate_content_length.rb
+++ b/lib/chef/http/validate_content_length.rb
@@ -86,8 +86,13 @@ class Chef
         transfer_encoding = http_response["transfer-encoding"]
         content_encoding  = http_response["content-encoding"]
 
-        if content_length.nil? || content_length < 0
+        if content_length.nil?
           Chef::Log.debug "HTTP server did not include a Content-Length header in response, cannot identify truncated downloads."
+          return true
+        end
+
+        if content_length < 0
+          Chef::Log.debug "HTTP server responded with a negative Content-Length header (#{content_length}), cannot identify truncated downloads."
           return true
         end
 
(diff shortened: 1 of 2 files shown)


whiteadam:
+1

mrvinti:
@coderanger thanks for fast PR!

This is the exception raised during convergence.

However *I believe* Net::HTTP yields the 1st byte only in this case and not reading all the bytes from the socket.

    * remote_file[/var/chef/cache/dhcp-4-3-3-p1.tar.gz] action create

    =========================================================================
    Error executing action `create` on resource 'remote_file[/var/chef/cache/dhcp-4-3-3-p1.tar.gz]'
    =========================================================================
    
    Chef::Exceptions::ContentLengthMismatch
    ---------------------------------------
    Response body length 1 does not match HTTP Content-Length header -1.
    This error is most often caused by network issues (proxies, etc) outside of chef-client.

coderanger:
@mrvinti Are you saying you still see that error with this?

lamont-granquist:
:lolwut:

:+1:

lamont-granquist:
need to rub a little `chefstyle -a` on it tho

coderanger:
Not waiting for CI to check the style fix because yolo.gif :)

mrvinti:
@coderanger I did not use Chef from source so currently I cannot tell if the problem still exists.

I was suggesting that Chef doesn't fetch the entire body since the reported length is 1: `Response body length 1`

So even if it skips content-length check with the #4646, the downloaded file will have a 1byte size.
(I cannot see the file in `/var/chef/cache/*`. It seems it's either deleted by Chef when the exception is raised or maybe the Net::HTTP does not download it at all.)

lamont-granquist:
if Net::HTTP only reads a single byte, then its a net-http bug and we can't fix it in this repo.

and remote_file will download the file to a dotfile in the target directory and only after the validation check will it copy it into place, and it'll remove tempfile when it cleans up on exit (we don't clobber your destination with a corrupted download).

mrvinti:
> if Net::HTTP only reads a single byte, then its a net-http bug and we can't fix it in this repo.

Yeah, this is my suspicion. If I get the time I'll use commit 8b94d23 to asses whether this is the case.

Thanks for the `remote_file` explanation.

lamont-granquist:
(i'd also be somewhat surprised if net-http looked at content-length at all, because the whole point of us having to handle content-length is that net-http completely ignored it...  i suppose its entirely possibly they've implemented it by now, though, and now we're getting double-checking that has this issue in two places...)

PR #4649, opened 2016-03-02
WIP: read roles from ruby files

thommay:
:construction: Would appreciate a review of the approach :construction:
Essentially, we can read ruby roles (and environments) from the filesystem, but they must map in to json objects in Chef Zero.

### DIFF
--- a/lib/chef/chef_fs/data_handler/data_handler_base.rb
+++ b/lib/chef/chef_fs/data_handler/data_handler_base.rb
@@ -190,7 +190,7 @@ class Chef
         # Calls the on_error block with the error, if there is one.
         #
         def verify_integrity(object, entry, &on_error)
-          base_name = remove_dot_json(entry.name)
+          base_name = File.basename(entry.name, ".*")
           if object["name"] != base_name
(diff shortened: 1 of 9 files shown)


thommay:
So `knife diff` gets a bit weird here - when we download a role "`foo`" we want to write `roles/foo.json`, because we're downloading json. But the canonical version of that role might be `roles/foo.rb`, and when we diff we want to compare `roles/foo.rb` to `/roles/foo.json` on the server. But it's ok for us not to have `roles/foo.json` on the filesystem, and we shouldn't mention it, as long as we have a valid source form of `foo`.

thommay:
@jkeiser thoughts on this would be appreciated

jkeiser:
@Thom may the biggest thing about the code: I would use the data handler to do to/from ruby so you don't have to specialize all the entry types.  Will comment more later

thommay:
@jkeiser yeah, i didn't like the specialisation although I'm not entirely sure it won't be necessary anyway to deal with name mangling between the repo and the server.

thommay:
Closing this, it's totally different now

PR #4652, opened 2016-03-02
ChefDK was mistakenly left out of the Chef maintenance policy

tyler-ball:
\cc @chef/maintainers 

ChefDK should have been listed in here under Dev Tools - this just bring the doc in line with reality

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -46,7 +46,7 @@ To mention the team, use @chef/client-core
 
 ## Dev Tools
 
-Chef Zero, Knife, Chef Apply and Chef Shell.
+ChefDK, Chef Zero, Knife, Chef Apply and Chef Shell.
 To mention the team, use @chef/client-dev-tools
 
 ### Maintainers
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I sort of would have expected chef-dk to have its own maintainers.md file...

lamont-granquist:
(i'm okay if we want to do this as well for expediency though)

coderanger:
So far we've been putting everything in this one file, but "so far" is only a few things (ohai, chefspec, chef-zero) so :+1:

If we want to split stuff out later then we could do that, I'm ambivalent overall.

lamont-granquist:
i could have sworn we just added a maintainers file to ohai at least, now i think i'm just confused...

:+1: however we do it...

coderanger:
@lamont-granquist We started to and the pivoted to doing it all in-line https://github.com/chef/ohai/pull/687

sethvargo:
:+1:

stevendanna:
:+1:

tyler-ball:
If it ever gets confusing enough, we could always add a `MAINTAINERS.md` file to other projects and just have it redirect to here.  But I don't think its urgent

PR #4653, opened 2016-03-02
release 12.8.0 changelog

mwrock:


coderanger:
Can we make whatever script generates this force initial caps on each line? It makes my ocdaesthetics glads cranky ;-)

coderanger:
(not actually a :-1: just grumble grumble getoffmylawn :) )

jkeiser:
@coderanger it just uses the titles of the bugs, so if you click on each bug and capitalize its title (or rename), next time we generate it will use the new titles :)

jkeiser:
:+1:

PR #4654, opened 2016-03-03
Cleanup static libs at build time

chefsalim:
Add step for unused static lib cleanup at build time

@chef/client-core @lamont-granquist


ksubrama:
:+1:

PR #4655, opened 2016-03-03
Clearer exception for loading non-existent data bag items in solo mode.

coderanger:
So on one hand it would be nice if this raised the same exception for both client and solo modes, on the other hand that would be a compat break and this is better than nothing. On the client code path is ends up raising `Net::HTTPServerException` with a 404. Mostly this matters because of 

If we don't fail here, it blows up a few lines later when running `from_hash(nil)`. This gets used in https://github.com/chef/chef-vault/blob/master/lib/chef-vault/item_keys.rb#L114-L127, found via a [report on StackOverflow](https://stackoverflow.com/questions/35761065/knife-solo-and-vault-data-bag-item-rb129in-from-hash-undefined-method-d).

### DIFF
--- a/lib/chef/data_bag_item.rb
+++ b/lib/chef/data_bag_item.rb
@@ -150,6 +150,7 @@ class Chef
     def self.load(data_bag, name)
       if Chef::Config[:solo]
         bag = Chef::DataBag.load(data_bag)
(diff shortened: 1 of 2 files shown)


thommay:
:+1:

PR #3082, opened 2015-03-16
Remote file download progress

brandocorp:
I took a stab at implementing this for #2812. I'm not sure how much I like the ProgressBar class, and it could definitely be trashed for something like ruby-progressbar.

I'm looking for feedback, let me know what you think and what needs to be better.

PR #4658, opened 2016-03-03
Remote file download progress

### LINKED ISSUE #2812: Feature request: Show remote_file progress
Hi,

Using `remote_file` on a large file results in a recipe that looks like it's stuck because nothing seems to be happening (I'm pulling down 546 Mb over http). It'd be very useful to show the progress as it downloads the file (assuming I haven't missed an option that makes this possible).

And thanks for making Chef!

thommay:
Based on @brandocorp's work in #3082 rebased, with some tweaks around configuration and tests.

cc @chef/client-core
Fixes #3082, Fixes #2812

### DIFF
--- a/lib/chef/formatters/doc.rb
+++ b/lib/chef/formatters/doc.rb
@@ -23,6 +23,7 @@ class Chef
         @start_time = Time.now
         @end_time = @start_time
         @skipped_resources = 0
+        @progress = {}
       end
 
       def elapsed_time
@@ -256,6 +257,21 @@ class Chef
         # TODO: info about notifies
         start_line "* #{resource} action #{action}", :stream => resource
         indent
+      end
+
+      def resource_action_progress(resource, current, total, interval)
+        @progress[resource] ||= 0
+
+        percent_complete = (current.to_f / total.to_f * 100).to_i
+
+        if percent_complete > @progress[resource]
+
+          @progress[resource] = percent_complete
+
+          if percent_complete % interval == 0
+            start_line " - Progress: #{percent_complete}/100 %", :green
+          end
+        end
       end
 
       # Called when a resource fails, but will retry.
--- a/lib/chef/formatters/doc.rb
+++ b/lib/chef/formatters/doc.rb
@@ -259,7 +259,7 @@ class Chef
         indent
       end
 
-      def resource_action_progress(resource, current, total, interval)
+      def resource_update_progress(resource, current, total, interval)
         @progress[resource] ||= 0
 
         percent_complete = (current.to_f / total.to_f * 100).to_i
@@ -269,7 +269,7 @@ class Chef
           @progress[resource] = percent_complete
 
           if percent_complete % interval == 0
-            start_line " - Progress: #{percent_complete}/100 %", :green
+            start_line " - Progress: #{percent_complete}%", :green
           end
         end
       end
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -74,8 +74,6 @@ class Chef
     attr_reader :redirect_limit
 
     attr_reader :options
-    attr_reader :show_progress
-    attr_reader :progress_interval
 
     attr_reader :middlewares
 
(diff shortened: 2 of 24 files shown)


lib/chef/http.rb
```diff
@@ -74,6 +74,9 @@ def self.use(middleware_class)
     attr_reader :redirect_limit
 
     attr_reader :options
+    attr_reader :show_progress
```
coderanger: Are these being used?

thommay:
@coderanger nope, gone.

lib/chef/formatters/doc.rb
```diff
@@ -258,6 +259,21 @@ def resource_action_start(resource, action, notification_type = nil, notifier =
...
+        @progress[resource] ||= 0
+
+        percent_complete = (current.to_f / total.to_f * 100).to_i
+
+        if percent_complete > @progress[resource]
+
+          @progress[resource] = percent_complete
+
+          if percent_complete % interval == 0
+            start_line " - Progress: #{percent_complete}%", :green
```
tyler-ball: Will this continually overwrite the same line, or will it output multiple lines for the `- Progress:`?

thommay: multiple lines; there's no real way to do the update in a cross platform way that I know of (it requires a lot of terminal ANSI codes and nasty to make it work)

lamont-granquist:
@thommay https://github.com/jfelchner/ruby-progressbar ?

thommay:
interesting, thanks. Not sure it's actually necessary though, either.

lamont-granquist:
yeah, but it might be a nice way of leveraging curses stuff to make progressbars on the same line without having to mess with curses and ANSI codes yourself...

stevendanna:
I don't think we need in-place updating right now, but in general if you are updating a single line and you are the only one writing to stdout, you don't need anything fancy. Write the line without a newline appending, write subsequent lines starting with a carriage return.

danielsdeleo:
I'd really prefer not to have multiple lines, it makes the output too cluttered. If that's the only way we can do it then at least it should have some additional logic (like bucketing by 5% intervals or something) to reduce clutter.

lib/chef/formatters/doc.rb
```diff
@@ -258,6 +259,21 @@ def resource_action_start(resource, action, notification_type = nil, notifier =
...
+      def resource_update_progress(resource, current, total, interval)
+        @progress[resource] ||= 0
+
+        percent_complete = (current.to_f / total.to_f * 100).to_i
+
+        if percent_complete > @progress[resource]
+
+          @progress[resource] = percent_complete
+
+          if percent_complete % interval == 0
```
coderanger: Doesn't this miss printing if it jumps from, say, 19% to 21%?

coderanger:
:+1: for more bucketing, I would also do it by time, like only print an update every 5 seconds. That should cut down on useless log spam for small files or big files on internal gigabit.

lib/chef/formatters/doc.rb
tyler-ball: :+1:

lib/chef/formatters/doc.rb
brandocorp: Yes, I believe it would

brandocorp:
@danielsdeleo, @coderanger 
The current behavior is to update at a user defined percentage-interval, with the default at 20%. This should limit the log spam from this to a maximum of 5 events per remote_file resource. Is this what you meant by 'bucketing'? 

Time based eventing could be done instead, but for very large files, or very slow transfers it could even more noisy.

@thommay 
I lost track of this one, but let me know if I can help!

smurawski:
:+1:

lamont-granquist:
:+1: good enough to ship as MVP, its off by default and we can make it prettier later...  :shipit:

lib/chef/formatters/doc.rb
thommay: Yup, but I'm not sure it matters. I think this is fine as a best effort, just to give people some idea of what's happening for really big files.

danielsdeleo:
Yeah, I think 10% is fine. And we can improve later, as well.

PR #4661, opened 2016-03-03
Extend service resource to support systemd user services

davide125:
systemd support the concept of user services (https://wiki.archlinux.org/index.php/Systemd/User), which are like system services, but are run and supervised within a user session.

This patch adds support for managing user services within the `service` resource, e.g. by doing
```
service 'foo' do
  user 'bar'
  action [:enable, :start]
end
```
the  service `foo` will be enabled and started whenever the user `baz` logs in.

### DIFF
--- a/lib/chef/provider/service.rb
+++ b/lib/chef/provider/service.rb
@@ -62,7 +62,16 @@ class Chef
         end
       end
 
+      # subclasses should override this if they do implement user services
+      def user_services_requirements
+        requirements.assert(:all_actions) do |a|
+          a.assertion { @new_resource.user.nil? }
+          a.failure_message Chef::Exceptions::UnsupportedAction, "#{self} does not support user services"
+        end
+      end
+
       def shared_resource_requirements
+        user_services_requirements
       end
 
       def define_resource_requirements
--- a/lib/chef/provider/service/solaris.rb
+++ b/lib/chef/provider/service/solaris.rb
@@ -49,6 +49,11 @@ class Chef
           @current_resource
         end
 
+        def define_resource_requirements
+          # FIXME? need reload from service.rb
+          shared_resource_requirements
+        end
+
         def enable_service
           shell_out!(default_init_command, "clear", @new_resource.service_name) if @maintenance
           shell_out!(default_init_command, "enable", "-s", @new_resource.service_name)
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -72,17 +72,17 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
 
   def get_systemctl_options_args
     if new_resource.user
-      uid = node['etc']['passwd'][new_resource.user]['uid']
+      uid = node["etc"]["passwd"][new_resource.user]["uid"]
       options = {
-        'environment' => {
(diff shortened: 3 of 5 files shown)


coderanger:
I am really not comfortable with adding so much to the core `service` resource that is entirely specific to systemd. This is not even remotely generic, you should be using a `systemd_service` resource.

davide125:
@coderanger: from the discussion in #4307 with @lamont-granquist, my understanding is that we do not want to make the `service` resource resolve to anything other than `Chef::Resource::Service` for now. Given that, how do you propose we do this? Yes, the implementation of this is systemd specific, but user services are services and are modeled well by the existing `service` resource. While we could make a completely separate `systemd_service` resource and provider, that would end up duplicating a lot of code and logic from `service`; it does not feel the right solution to me and it will end up being harder to maintain in the future.

coderanger:
You should make something that subclasses `Chef::Resource::Service` and `provides(:systemd_service)` in my opinion. This should be separate from `service` in both resource and provider give how much it is deviating from the model of `service`.

coderanger:
But to be clear, this is not an instruction to change the code, just my view as one maintainer :)

lamont-granquist:
Right, the issue is that I view that as the lesser of two evils.  We can either:

- break the validation model and the service resource will allow systemd-specific actions to hit non-sytemd providers

- break the lazy provider resolver mechanism so that if you have a system running systemd you will only (ever) be able to manage systemd service resources

I really don't like the second breakage at all and feel its more potentially 'crippling' breakage that ties our hands in the future -- while the first breakage only affects people who try to use non-systemd resources like systemd resources, which should be an edge condition, and the effect is they get an uglier error message or we have to inject some defensive programming into the service provider base class to fix up the error messages.

lamont-granquist:
paging @thommay

coderanger:
If you know you are using systemd-specific features, I don't see why it's a problem that they only work with systemd.

coderanger:
To be clear, I'm not talking about remapping `service`, I'm saying this should be it's own resource a la `package` and `deb_package`/`rpm_package`.

lamont-granquist:
so the prior code in the PR was implemented that way but it remapped `service "foo"` to `systemd_service "foo"` in the DSL and hardwired all services to be systemd services

jaymzh:
I would say in the realm of choosing between two evils, we should (a) be consistent and (b) pick the one that provides more flexibility in the short term. So, for now, lets keep this the way we did #4307 which is consistent and provides more flexibility even though it bloats the service resource a bit, and then we can revisit the resolution code if we want to be able to abstract that out later.

coderanger:
If we do that, we should figure out a way to at least warn, preferably error, if this property is set for any other provider. With actions that is relatively easy, but currently this would be silently ignored for other providers.

davide125:
@coderanger yeah adding checks is a good idea. Ideally we'd do  that in `lib/chef/provider/service.rb` so it doesn't have to be reimplemented in every other provider, but I'm not sure how to do that.

thommay:
consistency is definitely king here. I'm :+1: on this approach

lamont-granquist:
:+1:

thommay:
@jaymzh feel free to merge :)

chef-supermarket:

* @davide125

PR #4663, opened 2016-03-04
better chef+ruby rspec constraints

lamont-granquist:
lets us use Gem::Dependency constraints directly as rspec
constraints for matching Chef::Version and RUBY_VERSION.

copies some magic sauce from bundler that does something
similar


lamont-granquist:
@chef/client-core i got jealous of stuff i saw in the bundler rspec tests and wanted it for myself, so i stole it for us...

thommay:
SO much better! :+1:

Issue #4392, opened 2016-01-13
Data bags item no longer support "name" key


arr-dev:
Chef version: 12.6.0

Data bags no longer support `name` key. 
This worked on version 12.4.3 
~~and seems to be caused by this commit d99e306a41b1402209d320cb7119b12a3bbb962f,  specifically this change https://github.com/chef/chef/commit/d99e306a41b1402209d320cb7119b12a3bbb962f#diff-143e680ed6057de0f7066c7b539a7d28R131~~
edit, not true, see thommay's comment

This is a sample of data bag that worked in 12.4.3 and doesn't work with 12.6.0

```
{
  "id": "app_staging",
  "name": "app_staging",
  "environment_name": "staging",
...
}
```

With 12.4.3, `data_bag_item("deploy_setup", "app_staging")['name']` returned `app_staging`, 
but now it returns `nil`

thommay:
hi @arr-dev,
that commit isn't in 12.6, so there must be something else going on. I'm surprised that this ever worked, but I'll need to look further.

coderanger:
The code in `knife data bag from file` seems to handle this correctly, maybe something in ChefFS?

arr-dev:
@thommay you are right, it's not caused by this.
btw, I am using `chef-solo` if that has anything to do with it.
I've tried from irb, with `Chef::DataBagItem.load` and it behaves the same.

thommay:
@arr-dev "it behaves the same" - it displays the bug, or it works?

arr-dev:
Sorry, should have been clearer, it works on 12.4 and doesn't work on 12.6.

lamont-granquist:
I have no idea what broke, but I replicated it...

works:

```ruby
Starting Chef Client, version 12.4.3
resolving cookbooks for run list: ["test"]
Synchronizing Cookbooks:
  - test
Compiling Cookbooks...
OBJECT:
{"data_bag_item('foo', 'bar')"=>
  {"id"=>"bar",
   "name"=>"bar",
   "test"=>"junk",
   "chef_type"=>"data_bag_item",
   "data_bag"=>"foo"}}
NAME:
"bar"
Converging 0 resources

Running handlers:
Running handlers complete
Chef Client finished, 0/0 resources updated in 2.429475 seconds
```

fails:

```
Starting Chef Client, version 12.6.0
resolving cookbooks for run list: ["test"]
Synchronizing Cookbooks:
  - test (0.0.1)
Compiling Cookbooks...
[...]

PR #4664, opened 2016-03-04
Data bag item hash can have name key

chefsalim:
This PR resolves https://github.com/chef/chef/issues/4392

In cases where the data bag item is loaded with json that does not have the canonical knife format (with the user fields in "raw_data"), then the loading code was explicitly deleting the "name" key.  It is not clear whether the deletion of the "name" key actually served any purpose (in the case where the raw_data IS present, deleting "name" is a no-op), so removing it.

@chef/client-core @thommay @tas50 @lamont-granquist

### DIFF
--- a/lib/chef/data_bag_item.rb
+++ b/lib/chef/data_bag_item.rb
@@ -128,7 +128,6 @@ class Chef
     def self.from_hash(h)
       h.delete("chef_type")
       h.delete("json_class")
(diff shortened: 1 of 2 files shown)


lamont-granquist:
seems odd that was ever there, still kinda wonder how that happened...  but :+1:

thommay:
:+1:

PR #4666, opened 2016-03-04
Use ruby 2.1.8 on both windows and unix

jkeiser:
This has passed through the build+test matrix. Windows upgrades from 2.0 to 2.1, and unix gets a patch bump from 2.1.6 to 2.1.8.

NOTE: we will not merge this until after 12.8 is released, to give us more time before the next release to sort out any issues that crop up.

### DIFF
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -34,17 +34,10 @@ else
   install_dir "#{default_root}/#{name}"
 end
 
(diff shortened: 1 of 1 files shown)


ksubrama:
I see you ninja fixed your devkit change :)
:+1:

lamont-granquist:
:+1:

tyler-ball:
:+1:

PR #4667, opened 2016-03-04
adding docker to top_cookbooks

mwrock:


adamedx:
I'm curious @mwrock  -- I see that it's exercising recipes, but what exactly does it do? Is inspec / serverspec being used to validate the result of the recipes?

mwrock:
no this basically puts docker "through its paces" so to speak. Since the docker cookbook has no `default` recipe and is rather a collection of resources, it seemed best to tun it through the same recipes that its own kitchen yml uses. Otherwise we would have to create a separate "wrapper" cookbook to excercise its resources.

chefsalim:
:+1:

adamedx:
:+1:

PR #4673, opened 2016-03-07
preferred_manifest_record: fix pretty print.

### LINKED ISSUE #2561: CookbookVersion#preferred_manifest_record can destroy file lists
Hi

`CookbookVersion#preferred_manifest_record` uses `map!` to strip the root_dir which leaves the `CookbookVersion` with a broken file list. This is often not a problem since an exception is raised right afterward. However, the exception may be caught or the `CookbookVersion` may be inspected later and then it gets weird. I think this is pretty easy to address by using `map` and a local variable.

Cheers'
Hannes

dduponchel:
The `preferred_manifest_record` function pretty prints the list of
existing files when the asked file can't be found. To do that, the root
path of the cookbook is removed from each existing file but the update
is done on the `existing_files` variable, not on a copy. When called
several times, this variable will eventually be equals to `[nil]`,
leading to a NoMethodError.

Without the fix, the new unit test fails with

> expected Chef::Exceptions::FileNotFound, got #<NoMethodError: undefined method `[]' for nil:NilClass> with backtrace:
>   # ./lib/chef/cookbook_version.rb:321:in `block in preferred_manifest_record'
>   # ./lib/chef/cookbook_version.rb:319:in `map!'
>   # ./lib/chef/cookbook_version.rb:319:in `preferred_manifest_record'
>   # ./spec/unit/cookbook_version_file_specificity_spec.rb:318:in `block (3 levels) in <top (required)>'
>   # ./spec/unit/cookbook_version_file_specificity_spec.rb:317:in `block (2 levels) in <top (required)>'

Fixes #2561.

### DIFF
--- a/lib/chef/cookbook_version.rb
+++ b/lib/chef/cookbook_version.rb
@@ -316,11 +316,17 @@ class Chef
           error_message << error_locations.join("\n")
           existing_files = segment_filenames(segment)
           # Strip the root_dir prefix off all files for readability
-          existing_files.map! { |path| path[root_dir.length + 1..-1] } if root_dir
+          pretty_existing_files = existing_files.map { |path|
+            if root_dir
+              path[root_dir.length + 1..-1]
+            else
+              path
+            end
+          }
           # Show the files that the cookbook does have. If the user made a typo,
           # hopefully they'll see it here.
-          unless existing_files.empty?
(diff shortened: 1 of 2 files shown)


thommay:
LGTM :+1: @chef/client-core

lamont-granquist:
:+1:

Issue #2234, opened 2014-10-15
considering making "include" part of recipe DSL


juliandunn:
A lot of people get trolled by the fact that Chef recipes look like regular Ruby, so they think they can just do things like:

``` ruby
include Opscode::OpenSSL::Password
```

to mixin modules. But of course that doesn't work, 'cause `include` isn't part of the DSL. Thus users have to dive into ugly Ruby metaprogramming, with syntax like:

``` ruby
::Chef::Recipe.send(:include, Opscode::OpenSSL::Password)
```

to effectively monkeypatch the Recipe class to mixin the module.

It gets even gnarlier if they want `ruby_block` to mixin modules; now they need to know the provider class:

``` ruby
Chef::Provider::RubyBlock.send(:include, Some::Thingy)
```

Is there a way we could make the user experience better? Say, by making `include` part of the DSL? Other ideas?

lamont-granquist:
`not_if` blocks are evaluated in the context of the Chef::Resource::Whatever object so its reasonable that people would want that as well.  There should probably be an easy way to mixin to Chef::Recipe, Chef::Resource and Chef::Provider all in one go, although that's a rather large footgun.

Something like:

`use Opscode::OpenSSL::Password, in_resource: :ruby_block`

Could be used to wire up to the resource and provider that :ruby_block evaluates to, but with service providers that'd be problematic to determine.  Could also take a class name and something like :all_resources to engage the full footgun mode....

ranjib:
i love it :-). may be use 
```ruby
use Foo::Bar
```
for inclusion in all resources & recipes. and 
```ruby
use Foo::Bar, in: :ruby_block, :execute
```
if user want to mixin their module only in one resource, they can do it directly, without `use` DSL
```ruby
ruby_block 'mixin_demo' do
  exten Foo::Bar
  block do
  end
end
```
we can a generic middle ware API for these mixins using the top level Chef.* DSL methods.

lamont-granquist:
I'd probably want `use Foo::Bar` to only hit the recipe DSL.  For the footgun I think you should have to type a bit more.  Its pretty dangerous because if you collide with any internal method names or collide with any properites of any Resources or LWRPs then you may get very odd behavior, and you may see that odd behavior very distant from where you launched the footgun at.

lamont-granquist:
per discussion in #4674 we probably want to be more targeted with this.  so by default `use Foo::Bar` should ideally only infect the instance of the object of the current DSL context.

so:

```
[...]

PR #4674, opened 2016-03-07
implement use DSL method and lazy module inclusion

lamont-granquist:
WHAT THIS SOLVES

I want a delightful way to do something similar to this:

```ruby
module MyMethods
  def stuff
  end
end
Chef::Recipe.send(:include, MyMethods)
```

And get the semantics truly correct.

PROBLEM 1 / SOLUTION 1 

- this doesn't cover Chef::Provider::InlineResources

so lets try:

```ruby
Chef::Recipe.send(:include, MyMethods)
Chef::Provider::InlineResources.send(:include, MyMethods)
```

PROBLEM 2 / SOLUTION 2

- this doesn't cover arbitrary classes that someone has mixed the FullDSL into

so lets try:

```ruby
Chef::DSL::Recipe::FullDSL.send(:include, MyMethods)
```

PROBLEM 3 / SOLUTION 3

- That doesn't work at all because classes are patched by modules at class parsing time, they don't late-bind and you don't inherit the monkey patches.

so lets try creating a `use MyModule` DSL method per #2234 and implement some tracking of modules that include the FullDSL so that we can update all of them.

PROBLEM 4

- The behavior of `Chef::DSL::Recipe::FullDSL.send(:include, MyMethods)` is /never/ what I want in that context, so lets just `@#$@#$` fix ruby....  And lets implement `use` on the DSL as well, because that will be significantly more pleasant, but now its just sugar around that call...

And that's how I wound up here.....


coderanger:
Which PR?

lamont-granquist:
i'm furiously typing a new description for this, gimme a few more minutes...

lamont-granquist:
there ya go

coderanger:
This has been on my to-do list for a while too, but I'm not sure I'm sold on this. We should start out stating that almost no one should be using this because global DSL modification is deep into footgun territory, but I've implemented enough terrible versions of this to agree it is needed anyway. Can you include how your example would look with the actual `use` method? I would think most places that would be calling this would be in `libraries/` files, so having the `use` method itself isn't super clear from that context.

lamont-granquist:
i think in a library you'd want to use the `Chef::DSL::Recipe::FullDSL.send(:include, MyMethods)` form.  otherwise you'd have to:

```ruby
require 'chef/dsl/use'
Module MyMethods
  extend Chef::DSL::Use

  def whatever
  end

  use self
end
```

Which might be crossing the streams and might not even work?  I wouldn't want to inject `use` into main, and I'm not super hot on changing the receiving context of library parsing to something else (at least I'm assuming we're in main, i haven't checked...)

coderanger:
So where would you see `use` being used? I _really_ don't think we should advocate doing global DSL mutation from recipe files.

lamont-granquist:
and the thing is this isn't any worse than writing your own resource or definition...

lamont-granquist:
its the use case in #2234

the most popular one is:

```ruby
Chef::Recipe.send(:include, OpenSSLCookbook::RandomPassword)
node.set['my_secure_attribute'] = random_password
```

which becomes:

```ruby
use OpenSSLCookbook::RandomPassword
node.set['my_secure_attribute'] = random_password
```

terrible for any number of reasons, but the patching-the-dsl horse is already out of the gate there...

if we noodle on that for too long i think i'll start going down the road of wanting `random_password` to just be slurped up into core chef because having that as a library is just asking for a terrible namespace collision at some point, so we might as well decisively stomp on it once and for all...

lamont-granquist:
that doesn't change the fact that its an instance of an overall pattern people use...

coderanger:
I'm saying that pattern is not something we should support. The DSL mutation should happen in the `libraries/whatever.rb` file, not the recipe. That was a bug in the `openssl` cookbook from day one.

lamont-granquist:
yeah i guess i could extract that from this problem.  thinking about it more i think `use` might better target the instance of Chef::Recipe that you're executing in, so that you have some seperability.  there's also the issue where people want to inject methods into resources to be able to use them in not_if/only_if blocks and ruby_block codeblocks.  i can see how that's all a separate concern from 'how do we globally patch the DSL' though.

coderanger:
Yeah, `use Foo` in a recipe doing non-global DSL manipulation I am hugely :+1: on, and that should be a slightly separate case from global DSL stuffs.

lamont-granquist:
okay use is gone, we're just concerned with global DSL manipulation in libraries now.

coderanger:
Okay, so :+1: for use case and impl, last question is do we want to call this `include`? We could make it some other name, but I think the semantics are probably fine. Trying to think of the ways this could break something where someone actually did want the normal `include` behavior, but I can't think of anything.

lamont-granquist:
Yeah, I dunno.  I wasn't certain if that was a bridge too far or not.

I probably need to go back and pull some of the use cases from the `use` tests I just deleted back and repurpose them so this isn't ready to merge anyway.  Will definitely sit on it for a few days.

lamont-granquist:
@chef/client-core persuant to adding a couple more end-to-end-ish tests this could use more eyeballs.

lamont-granquist:
there is also the impl question of this is yet another DescendantsTracker kind of thingy, but it will get hit exceedingly infrequently compared to the ProviderResolver stuff...  i'm not sure i can see another way of implementing it other than requiring that class/modules not directly mix in the DSL but have a layer of indirection to "properly" mix in the DSL, but all that'll do is register them with some central authority anyway...  i can't see any way out of fundamentally maintaining an array of classes to update somewhere...

ranjib:
what if we name the dsl method `use` instead of redefining `include`, and implement it via extend, i.e   descendant.extend klass. instead of send(include, klass), am i missing some failure modes here?

lamont-granquist:
there's no DSL method any more, this is now completely orthogonal to the `use` use-case.

we have to include it as well because we want instance methods not class methods.

PR #4681, opened 2016-03-09
Fix property coercion

coderanger:
Because they were different and one in Chef::Mixin::Properties would result in duplicated values for string keys.

Very specifically not adding a new test for string keys on `property` because I think this is not part of our declared API (we only state support for symbol keys) and we probably shouldn't test formally undefined behavior.


lamont-granquist:
Did the string handling get copied over from attributes?  It should probably be deprecated and removed.

But :+1:

PR #4682, opened 2016-03-09
Improved method for getting currently installed profiles

natewalck:
This improves how I was getting currently installed profiles. Previously, I was dumping all profiles to disk and then reading them back to figure out what was installed. This was pretty stupid as I could just read them out directly using stdout-xml (in the man page for the profiles command, but not --help. Thanks Apple).

As usual, I am with Facebook and this PR must be merged by PhilD.

### DIFF
--- a/lib/chef/provider/osx_profile.rb
+++ b/lib/chef/provider/osx_profile.rb
@@ -169,7 +169,7 @@ class Chef
 
       def get_profile_hash(new_profile)
         if new_profile.is_a?(Hash)
-          return new_profile
+          return new_profile.to_hash
         elsif new_profile.is_a?(String)
           return load_profile_hash(new_profile)
         end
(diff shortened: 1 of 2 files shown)


natewalck:
Tagging @jaymzh for the future merge.

jaymzh:
As I mentioned on your internal diff - your API expects a string or hash - so users should probably pass in a string or a hash... Calling .to_hash seems like the wrong solution - the user should pass in an actual hash, not an attribute.

natewalck:
@jaymzh Pointed out that if you pass an attribute directly, its not a hash...it is an attribute object. This change would have masked that issue, so I dropped the to_hash.

The improvement should be merged upstream, so I'll keep this PR for that.

jaymzh:
@natewalck  you'll want to edit the description of the PR for easier reviewing of other folks.

Also you have tests failing.

natewalck:
Fixed the lint error and updated the description. I'll squash once approved.

lib/chef/provider/osx_profile.rb
```diff
@@ -214,23 +214,18 @@ def get_installed_profiles(update = nil)
...
-        cmd = "profiles -P -o '#{tempfile}'"
-        shell_out!(cmd)
+      def dump_installed_profiles
+        cmd = "profiles -P -o stdout-xml"
+        shellout_results = shell_out(cmd)
+        if shellout_results.exitstatus == 0
+          shellout_results.stdout
+        else
+          raise Mixlib::ShellOut::ShellCommandFailed, "profiles command failed"
+        end
```
lamont-granquist: just use `shell_out!` here?

natewalck: It appears the only difference between shell_out and shell_out! is that shell_out! will raise errors, whereas shell_out will not, correct?

I misunderstood the use case for each, so this appears to work as expected.

lamont-granquist: yeah but its awkward code that looks like it could be all be replaced just by:

```ruby
shell_out!(cmd).stdout
```

natewalck: I need to check exit status as well. I do not want to assume it ran successfully.

lamont-granquist: that is what `shell_out!` does

natewalck:
This has been updated and is waiting for further review. Could we get another look at it? @lamont-granquist @jaymzh

thommay:
:+1:

lamont-granquist:
:+1:

although apple's `-o stdout-xml` makes my brain hurt a bit, but whatever apple...

chef-supermarket:

* @natewalck

natewalck:
Do NOT merge this. Apparently there is an edge case on 10.10.5 where Apple's profile tool does not output the profiles to XML correctly. It seems related to one of their security updates that was posted for a short amount of time and then pulled.

We need to leave this functionality as it was, sadly.

coderanger:
Can we at least put the `unlink` in an `ensure` so we don't leak tempfiles?

PR #4684, opened 2016-03-09
Fix issue 4334: eliminate spurious warnings.

randomcamel:
Because set(property, nil) gets turned into get(property), add a 'nil_set' parameter to get() to skip a spurious warning when that happens.

### DIFF
--- a/spec/integration/recipes/lwrp_inline_resources_spec.rb
+++ b/spec/integration/recipes/lwrp_inline_resources_spec.rb
@@ -45,6 +45,33 @@ describe "LWRPs with inline resources" do
     end
   end
 
+  context "with an inline resource with a property that shadows the enclosing provider's property" do
+    class LwrpShadowedPropertyTest < Chef::Resource::LWRPBase
+      PATH = ::File.join(Dir.tmpdir, "shadow-property.txt")
+      use_automatic_resource_name
+      actions :fiddle
+      property :content
+      action :fiddle do
+        file PATH do
+          content new_resource.content
+          action [:create, :delete]
+        end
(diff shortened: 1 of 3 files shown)


jkeiser:
:+1: . The extra parameter can break extenders in some cases, but I haven't seen extenders in the wild yet. We're going to need to update compat_resource ASAP though.

spec/integration/recipes/lwrp_inline_resources_spec.rb
```diff
@@ -45,6 +45,33 @@ def action_a
...
+      property :content
+      action :fiddle do
+        file PATH do
+          content new_resource.content
+          action [:create, :delete]
+        end
+      end
+    end
+
+    after { File.delete(LwrpShadowedPropertyTest::PATH) rescue nil }
```
tyler-ball: This seems a potential way to bleed state around - why not let the exception get raised?

randomcamel: I wanted to make sure the file got deleted even if something went wrong with the `file` resource's `:delete` action. That will mostly never happen, so the FileNotFound exception case is normal.

As I type that, I like it less, when I can just do `if File.exists?`.

tyler-ball:
:+1:

tyler-ball:
Actually, another note:  How do we ensure that `compat_resource` gets updated?  I would like to see a card or an issue (or even better, a PR with the fix) to make sure we don't forget that.

jkeiser:
@tyler-ball I'll get it updated right after

PR #4692, opened 2016-03-10
Make mount idempotent on Windows

chefsalim:
Mounting a device was not idempotent on Windows, and would fail with an error:

    Chef::Exceptions::UnsupportedAction
    -----------------------------------
    #<Chef::Provider::Mount::Windows:0x36abe70> does not implement #mount_options_unchanged?

The options check functionality was added by this change, but missed Windows: https://github.com/chef/chef/commit/43620a7e3d663ca83bc4394a5baaf724fecf06d8

Fix is to add the missing function.

@chef/client-core @mwrock

### DIFF
--- a/lib/chef/provider/mount/windows.rb
+++ b/lib/chef/provider/mount/windows.rb
@@ -80,6 +80,12 @@ class Chef
           end
         end
 
+        private
+
+        def mount_options_unchanged?
+          @current_resource.device == @new_resource.device
+        end
+
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/mount/windows.rb
```diff
@@ -80,6 +80,12 @@ def umount_fs
           end
         end
 
+        private
+
+        def mount_options_unchanged?
```
mwrock: I don't see where this gets called.

chefsalim: It gets called by action_mount in chef/lib/chef/provider/mount.rb

mwrock:
:+1:

tyler-ball:
:+1: after a spec is updated - seems like it would be pretty easy to cover with a unit test

PR #4693, opened 2016-03-11
remove deprecated Thread.exclusive

### LINKED ISSUE #4359: Ruby 2.3.0 Thread.exclusive deprication warning
After upgrading to Ruby 2.3.0, we get the following warning: 

```
Thread.exclusive is deprecated, use Mutex
/Users/pivotal/.gem/ruby/2.3.0/gems/chef-12.6.0/lib/chef/dsl/declare_resource.rb:91:in `build_resource'
/Users/pivotal/.gem/ruby/2.3.0/gems/chef-12.6.0/lib/chef/dsl/declare_resource.rb:62:in `declare_resource'
...
```

Problem comes from: https://github.com/chef/chef/blob/master/lib/chef/dsl/declare_resource.rb#L91

Background: 
https://bugs.ruby-lang.org/issues/11904

Other Fixes to this problem: 
https://github.com/middleman/middleman-livereload/pull/85/files

nathwill:
currently, Chef blows up under ruby 2.3 (e.g. [this travis build](https://travis-ci.org/nathwill/chef-hekad/builds/115222568)), this fixes that

closes #4359

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -22,6 +22,7 @@ require "chef/exceptions"
 class Chef
   module DSL
     module DeclareResource
+      BUILD_RESOURCE_SEMAPHORE = Mutex.new
 
       #
       # Instantiates a resource (via #build_resource), then adds it to the
@@ -88,7 +89,7 @@ class Chef
       #
       def build_resource(type, name, created_at = nil, run_context: self.run_context, &resource_attrs_block)
         created_at ||= caller[0]
-        Thread.exclusive do
+        BUILD_RESOURCE_SEMAPHORE.synchronize do
           require "chef/resource_builder" unless defined?(Chef::ResourceBuilder)
         end
 
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -22,7 +22,6 @@ require "chef/exceptions"
 class Chef
   module DSL
     module DeclareResource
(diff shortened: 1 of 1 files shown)


coderanger:
This wouldn't mutex it. You would need to put the Mutex object in a const outside of the method so it is created statically at code load time.

nathwill:
ah, thanks @coderanger; fixed it up.

coderanger:
:+1: from me, but would like @lamont-granquist or @jkeiser to take a look if possible.

lamont-granquist:
it probably could just be removed, i think `require` is thread-safe and i'm just cargo-culting that at this point.

nathwill:
@lamont-granquist google's not very helpful about require's thread-safety, just a bunch of very old comments stating that it isn't. is there any reason the require shouldn't just be moved outside of the method before the class declaration?

lamont-granquist:
According to bundler core maintainer:

"Requiring is already tread-safe, and exclusive is deprecated."

https://github.com/bundler/bundler/pull/4245#discussion_r51481284

coderanger:
That's good enough for me, kill it with fire :)

ranjib:
yup. no need to use mutex
https://github.com/ruby/ruby/blob/3dbb390180a0e9f98623b6db0d71b0213359c541/load.c#L957

nathwill:
:+1: cool, mutex go :boom:

ranjib:
:+1:

lamont-granquist:
:+1:

PR #4694, opened 2016-03-11
Vbox's shared filesystem driver uses non-file device names too.

coderanger:


coderanger:
Kind of surprised this hasn't come up before.

thommay:
:+1:

PR #4696, opened 2016-03-11
Warn users not to tweak Chef::Config[:file_atomic_update] setting

lamont-granquist:
people are actually doing this in order to work around issues with
writing /etc/hosts in docker containers.  they should seriously
stop doing it and fix this problem in the cookbooks.


lamont-granquist:
@chef/client-core nobody should really be tweaking this variable, but people are, i kinda regret adding it now

coderanger:
Sighhhh :+1:

danielsdeleo:
:+1:

jaymzh:
:laughing: :+1:

PR #4699, opened 2016-03-11
moving pry gems to their own very special group

mwrock:


Gemfile
```diff
@@ -17,6 +17,12 @@ group(:maintenance) do
   gem "netrc"
 end
 
+group(:pry) do
+  gem "pry"
+  gem "pry-byebug"
+  gem "pry-stack_explorer"
+end
```
danielsdeleo: I had some questions about whether excluding this group from omnibus builds would have the effect of removing pry from the package, but I learned that it will not.

danielsdeleo:
:+1:

PR #4700, opened 2016-03-12
add systemd_unit resource

nathwill:
this still needs tests and such, but i wanted to get this open early for feedback. the intent is for this resource to provide a pretty basic systemd unit resource that can be built on by resources modeling specific systemd unit types.

i'd also like to hear opinions about whether systemd resources in chef-core should be able to create/delete the unit-files and support directive-based resource-properties like the systemd cookbook does. i'm obviously in favor, as i think it's incredibly powerful, but before i put all the effort in, i'd like some indication of whether it's got a chance of getting merged :smile:


lib/chef/resource/systemd_unit.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      actions :enable, :disable, :mask, :unmask,
+              :start, :stop, :restart, :reload
+
+      default_action :nothing
+
+      attribute :unit_name, kind_of: String, default: lazy { "#{name}.#{type}" }
+      attribute :mode, kind_of: String, default: 'system',
+                       equal_to: %w( system user )
+      attribute :type, kind_of: String, required: true, default: 'service',
```
davide125: I don't think this attribute is actually needed for anything. If it is, we should autopopulate it based on the `unit_name` (e.g. foo.timer implies `type` = `timer`).

davide125:
I think this is fine as a cleaner way to manage non-service units. About generating unit files from chef resources the way chef-systemd does, I don't think we should do this in core. While creating unit files as resources is elegant, in practice it means reimplementing the systemd DSL inside chef, which makes then it tightly coupled to systemd specifics (i.e. you then need to code to figure out this version supports this option this way). Another way of building a unit ex nihilo from chef could be to just have a `content` attribute that's just a hash of the full unit. This leaves you the ability to define it within the Chef recipe, but doesn't tie you to the DSL and makes it independent from systemd version quirks. If we want to support this usecase, I think it'd be a cleaner and more supportable way overall to do it.

FWIW though, what we've seen in practice is that in the vast majority of cases a `cookbook_file` is more than enough to drop units (as you can use stuff like `EnvironmentFile` for the dynamic parts, and have chef drop that with a template).

jaymzh:
So in and of itself this PR doesn't bother me, but I have some concerns with your larger design.

Having an end-goal of a "heavy" systemd wrapper where there's a property for every single thing you can put in a unit file is fool's errand - don't be that tightly integrated, you'll be chasing systemd forever.

Alternatively, consider the approach we took with the OSX profile provider in #4078 - we accept a hash of stuff and transform it into the right XML. We verify we have the necessary pieces, but that's it... the rest is transformed, and if the user passes in something that OSX doesn't like, that's what testing is for. But we're not Apple, nor are we Systemd, and we don't want chef to have to move in lockstep with other software.

So if we want to create a very thin wrapper that can generally create syntactically correct systemd files (and maybe environment files?) I can see that being generally useful... just to abstract out the same templates we're all already writing... but I'm generally opposed to a very thick 1:1 systemd mapping.

nathwill:
:+1: on using more hash->config conversion, that should be really straight-forward, and like you noted, has the advantage of not having to chase upstream. i'll rework this to add a content property and create/delete actions accordingly.

lib/chef/resource/systemd_unit.rb
nathwill: it's just about enforcing the fact that it's for unit management, and not a generic systemctl command pass-through. if we were going to support keyword->property mappings like chef-systemd it's a useful property to have for signaling what the supported properties should be for child-resources, but if we genericize unit like we're suggested in the other comments, i agree that it's not necessary.

nathwill: further, if we genericize this like suggested, we probably won't need child resources at all, so double win?

ranjib: if theres a default, do we need required true as well ?

lib/chef/resource/systemd_unit.rb
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/resource/lwrp_base'
+
+class Chef
+  class Resource
+    class SystemdUnit < Chef::Resource::LWRPBase
```
coderanger: Nothing should still be using this pattern. Please, pretty please.

lib/chef/resource/systemd_unit.rb
```diff
@@ -0,0 +1,43 @@
...
+      resource_name :systemd_unit
+      identity_attr :unit_name
+
+      actions :enable, :disable, :mask, :unmask,
+              :start, :stop, :restart, :reload
+
+      default_action :nothing
+
+      attribute :unit_name, kind_of: String, default: lazy { "#{name}.#{type}" }
```
coderanger: Should be `property` now.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,82 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'chef/provider/lwrp_base'
+require 'mixlib/shellout'
+
+class Chef
+  class Provider
+    class SystemdUnit < Chef::Provider::LWRPBase
```
coderanger: As with the other, this is not correct.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      %w(
+        enable disable start stop restart reload mask unmask
+      ).map(&:to_sym).each do |a|
+        action a do
+          r = new_resource
+
+          cmd = "systemctl --#{r.mode}"
+
+          unless defined?(ChefSpec)
```
coderanger: This doesn't look correct.

lib/chef/resource/systemd_unit.rb
davide125: Yep, I think in the generic model this isn't needed anymore, and neither are the child resources, which leads to a much simpler and cleaner model.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,82 @@
...
+            state = case a.to_sym
+                    when :enable, :disable, :mask, :unmask
+                      Mixlib::ShellOut.new(
```
coderanger: Use the `shell_out` helper.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,82 @@
...
+
+class Chef
+  class Provider
+    class SystemdUnit < Chef::Provider::LWRPBase
+      use_inline_resources
+
+      provides :systemd_unit
+
+      def whyrun_supported?
+        true
```
coderanger: I don't think you actually have support for whyrun here?

coderanger:
This looks like something copy-pasted out of a cookbook using a mismash of deprecated patterns. We should decide, for example, what standard we are using for defining actions in core providers.

nathwill:
@coderanger how about something helpful then, like a reference to your preferred pattern?

coderanger:
@nathwill The standard by how the rest of Chef's provider are implemented is to not use inline resources at all, in favor of manual control using `shell_out` and `shell_out!`, which is probably what I would do here too since the inner resources are just that one execute I think?

ranjib:
@coderanger are you saying that we should not use LWRPBase subclassing for core resource providers? I cant understand the view given the known pitfalls. But I wont say this is a standard :-) .

coderanger:
@ranjib It is the correct way, we removed all major DSL methods from the LWRPBase classes in 12.5. You can use the normal DSL stuffs (for the resource at least) without it. It is only not "the standard" because we have a huge inertia problem.

nathwill:
> we have a huge inertia problem.

as a heavy Chef user and sometime-contributor, i don't see this as an inertia problem, but a communication problem. i don't have time to follow all the activity on chef core, or listen in on PR discussions i'm not involved in. i try to keep up with the mailing list, IRC, the chef blog, the food fight show, but from the outside, all i see is *at least* a half-dozen ways to create resources, and no clear guidance on what's advantageous or problematic about any of the given options. using LWRPBase provides (provided?) some of the niceties of full-blown LWRPs while still retaining the control available in HWRPs, so it became my go-to. now it's apparently out of fashion?

i feel bad that this is written an unacceptable style, and i'm more than happy to change it, but if you expect the community to fall in line with evolving best-practices, somebody's gotta tell us what they are.

coderanger:
@nathwill That is all true, but there is a deeper issue that the code style/standard for Chef core vs. cookbooks is super different. We've done some work to bring them a bit closer together, but still you should not expect best practices from cookbook development to carry over to Chef core at this time.

davide125:
This looks much better, thanks for reworking it. Can you also add the user support stuff from #4661 so we get feature parity with `service` ?

lamont-granquist:
No resources in core chef have ever used LWRPBase.  There's a problem there because we don't get the niceties of use_inline_resources, but I lean towards the right way of doing that as adding something like notifying_block rather than inheriting from LWRPBase.  We already had all kinds of issues with the DSL being mixed into core resources and ripped it out.

nathwill:
@davide125 yep, absolutely! i was looking at the way you did the cmd options & args and used the presence of the user property to signal `--user` mode, definitely planning on copying you there :). out of curiosity, where'd you find the docs for `DBUS_SESSION_BUS_ADDRESS`? i couldn't find that anywhere last night.

nathwill:
@davide125 one other thing that's needed here, and something we might want to port back over to the systemd_service implementation too, is that the `enabled` handling isn't quite right, as it doesn't account for static units that can't really be enabled/disabled.

```bash
[root@default-centos-72 ~]# systemctl is-enabled systemd-tmpfiles-clean.timer
static
[root@default-centos-72 ~]# echo $?
0
[root@default-centos-72 ~]# systemctl status systemd-tmpfiles-clean.timer
● systemd-tmpfiles-clean.timer - Daily Cleanup of Temporary Directories
   Loaded: loaded (/usr/lib/systemd/system/systemd-tmpfiles-clean.timer; static; vendor preset: disabled)
   Active: inactive (dead) since Tue 2016-03-15 17:04:49 UTC; 2h 25min ago
     Docs: man:tmpfiles.d(5)
           man:systemd-tmpfiles(8)

Mar 15 16:59:44 default-centos-72 systemd[1]: Started Daily Cleanup of Temporary Directories.
Mar 15 16:59:44 default-centos-72 systemd[1]: Starting Daily Cleanup of Temporary Directories.
Mar 15 17:04:49 default-centos-72 systemd[1]: Stopped Daily Cleanup of Temporary Directories.
Mar 15 17:04:49 default-centos-72 systemd[1]: Stopping Daily Cleanup of Temporary Directories.
```

my assumption here is that enabled/disabled should just be a NOP for static units, since they lack an `[Install]` and can't really be explicitly enabled or disabled, but just get activated based on dependencies.

nathwill:
> We already had all kinds of issues with the DSL being mixed into core resources and ripped it out.

@lamont-granquist out of curiosity, what kinds of issues? thinking some context can help me understand the preference here.

davide125:
@nathwill yep, NOP for static units sounds reasonable, we can make it print a warning maybe so the user is aware. I don't remember where I found out about `DBUS_SESSION_BUS_ADDRESS`, likely it was either by reading the sources or on some random forum (the user services stuff is regrettably poorly documented upstream).

lamont-granquist:
@nathwill namespace issues.  i can't recall now.  they were referenced in the issue/PR that pulled the DSL out of Chef::Provider back in Nov/Dec 2014 for 12.0.0

nathwill:
ok, i think this is probably ready to go, though i welcome any additional feedback. i'm going to leave unit content management for a separate PR, since that's likely to be a bit more involved, and this is already a fairly large bit of code to have to review.

davide125:
:+1: this looks good to me, nice work!

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,172 @@
...
+        if current_resource.static
+          Chef::Log.warn("#{new_resource.name} is a static unit, enabling is a NOP.")
+        end
+
+        if current_resource.enabled || current_resource.static
+          new_resource.updated_by_last_action(false)
+        else
+          systemctl_execute!(:enable, new_resource.name)
+          new_resource.updated_by_last_action(true)
+        end
```
lamont-granquist: The manual twiddling of updated_by_last_action isn't necessary and eventually points to a why-run bug that you've got with this code.

systemctl_execute calls shell_out_with_systems_locale! directly without wrapping in a converge_by so there's no protection in why-run mode against running that code.

you also don't need to set updated_by_last_action since converge_by does that automatically, and false is the default value.  

so code should always look like:

```ruby
   if some_idempotent_condition?
    converge_by("do a thing") do    # will output "would do a thing" in why-run mode and then no-ops the block
       shell_out!("/usr/bin/do a thing")
    end
  end
```

generally code should never manually twiddle updated_by_last_action any more.

of course that statement relies on the fact that i just added use_inline_resources into Chef::Provider last week -- you'll still see a lot of core chef code which declares a sub-resource and manually copies off the updated status onto the new resource, but all of that can be converted now (and should be converted eventually) to use_inline_resources.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,172 @@
...
+        current_resource.masked(masked?)
+        current_resource.static(static?)
+
+        current_resource
+      end
+
+      def action_enable
+        if current_resource.static
+          Chef::Log.warn("#{new_resource.name} is a static unit, enabling is a NOP.")
+        end
```
lamont-granquist: is it really necessary to warn the user here?  if a NOP and its always a NOP does anyone care that much?  its vaguely 'unclean' code but it feels like we can safely make the user's life easier here by not bugging them about something where the action we can take is obvious...

nathwill:
@lamont-granquist great, thanks for the feedback. i've incorporated and squashed the commits down, let me know if there's anything else here that needs fixing.

lamont-granquist:
cool, code looks clean :+1:

lamont-granquist:
@chef/client-core other review?

jaymzh:
I'm confused. This does nothing that `service` doesn't do today. This isn't a unit at all, this is just... a re-implementation of service with only support for systemd.

I though thte point of this was to generate unit files for you.

nathwill:
@jaymzh as mentioned in https://github.com/chef/chef/pull/4700#issuecomment-200672967, i'd like to leave content management for a separate PR to keep this PR and the content-mgmt PR more focused.

> This does nothing that service doesn't do today

as it is right now, sure, but even left just like it is, i think this is preferable to what we have now:

- it properly handles static units, which service does not.
- it gives users an officially supported resource that's actually targeted at  *units*, so we can stop overloading the service concept for non-service units like sockets, timers, mounts, etc.
- it lays a solid foundation for drop-in config support once content-management is added, which wouldn't make sense anywhere but a unit-focused resource.

we can also probably purge the systemd-service provider and have the systemd-unit provider provide service mgmt on systemd platforms, but i think that's also more suited for a separate PR as well.

> This isn't a unit at all

i disagree. units are the resource representation inside systemd, not just the files on disk.

nathwill:
ok, since it looks like chef already depends on the iniparse gem, it's easier than i expected to add content management, so i got that started. i'll holler when it's ready for review.

nathwill:
ok, sorry for the slow updates, the weekend seems to be the only time i can get anything done... create/delete w/ ini-conversion's added, i think this is gtg.

lib/chef/provider/systemd_unit.rb
```diff
@@ -0,0 +1,210 @@
...
+        current_resource.masked(masked?)
+        current_resource.static(static?)
+        current_resource.triggers_reload(new_resource.triggers_reload)
+
+        current_resource
+      end
+
+      def action_create
+        if current_resource.content != new_resource.to_ini
+          converge_by("create the unit file") do
```
stevendanna: Do we want to add the name of the unite file that will be created here?

spec/unit/provider/systemd_unit_spec.rb
```diff
@@ -0,0 +1,797 @@
+#
+# Author:: Nathan Williams (<nath.e.will@gmail.com>)
+# Copyright:: Copyright (c) 2016 Chef Software Inc.
```
stevendanna: Any reason this is copyright Chef and not you?

spec/unit/resource/systemd_unit_spec.rb
```diff
@@ -0,0 +1,133 @@
+#
+# Author:: Nathan Williams (<nath.e.will@gmail.com>)
+# Copyright:: Copyright 2016, Chef Software Inc.
```
stevendanna: Same question here

stevendanna:
👍 This is looking really good to me.  Code is clean and self-documenting, good unit test coverage  I've left a few minor comments but I think this will be a great addition to Chef.

lib/chef/provider/systemd_unit.rb
nathwill: good call

spec/unit/provider/systemd_unit_spec.rb
nathwill: nope, just copy-pasta

nathwill:
thanks, @stevendanna, updated with your suggestions.

thommay:
@jaymzh are you opposed to this being merged? otherwise we think it's good from a code point of view...

lamont-granquist:
👍

jaymzh:
Sorry for the delay I've been on the road for weeks...

No objections - this provides something we don't have today and I think that's valuable.

We should consider the overlap though and how we'd like to best handle this as a community.

jaymzh:
:+1:

thommay:
Thanks, @nathwill!

nathwill:
awesome, thanks for all the feedback everyone, excited to start using this!

Issue #3540, opened 2015-06-16
Windows: env LWRP with action :modify overwrites env variables


npiguet:
I've been trying to provision some windows VMs, and I'm trying to add a folder to the `PATH` environment variable. So far, this has not worked so well. In fact, it seems that chef completely ovewrites the environment variables even when I use `action :modify`

Here is the block I use:

``` ruby
env "Add java binaries to PATH" do
    action :modify
    key_name "PATH"
    value "#{java_home}\\bin"
end
```

Here is how PATH looks like before the block runs

```
PS C:\Windows\system32> $Env:PATH
C:\Windows\system32\WindowsPowerShell\v1.0\;C:\ProgramData\Oracle\Java\javapath;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\opscode\chef\bin
```

and here is how it looks like after the block

```
> $Env:PATH
C:\Windows\system32\WindowsPowerShell\v1.0\;C:\softwares\extern\jvm\jdk1.8.0_45\bin
```

It seems that powershell always prepends itself to `PATH`, so that's why it's still there.

npiguet:
As a workaround, I'm currently using
```ruby
unless ENV['PATH'].include?(java_home)
  env "Add java binaries to PATH" do
    action :create
    key_name "PATH"
    value "#{java_home}\\bin;#{ENV['PATH']}"
  end
end
```

jdmundrawala:
there is a `windows_path` resource in the windows cookbook that you can use.
https://github.com/opscode-cookbooks/windows/blob/master/providers/path.rb#L25-L30 is how you'd do it if you didn't want to use the cookbook

npiguet:
Thanks for that. But that sounds more like a workaround than the way it really should be, no?

jdmundrawala:
It's either a bug in the code or the documentation. I feel like this one has come up before

jdmundrawala:
Maybe we change the documentation and add the ability to specify prepend or append?
Thought?
cc @chef/client-windows

npiguet:
That might be a good idea. The order does matter since I believe windows stops at the first match in `PATH`

ksubrama:
I just grepped in the chef repo for "delim" and couldn't find any place where we set it.  Shouldn't we just set it to "," on linux and ";" on windows?

MichaelPereira:
[...]

Issue #2262, opened 2014-10-20
Env resource implementation and documentation inconsistent


jdmundrawala:
## Version:
master

## Scenario
When using the :modify action, https://docs.getchef.com/resource_env.html says that the path is appended. This is not aligned with the implementation:
```ruby
      def modify_env
        if @new_resource.delim
          #e.g. add to PATH
          @new_resource.value(@new_resource.value + @new_resource.delim + @current_resource.value)
        end
        create_env
      end
```
where the new value is being prepended.

## Expected
The documentation should specify prepend. We should also provide a way the specify where to element should be added (beginning or end).

thommay:
Closing in favour of chef/chef-docs#520

PR #4703, opened 2016-03-13
Allow values to be appended when modifying environment values

MichaelPereira:
Hi,
After hitting this problem and seeing that this issue was brought up several times ( https://github.com/chef/chef/issues/2262 https://github.com/chef/chef/issues/3540 https://feedback.chef.io/forums/301644-chef-product-feedback/suggestions/8786671-add-support-for-appending-values-in-env-resource ) without being implemented, I've made the change myself in this Pull Request.

Let me know if this needs anything more.

### DIFF
--- a/lib/chef/provider/env.rb
+++ b/lib/chef/provider/env.rb
@@ -150,7 +150,11 @@ class Chef
 
       def modify_env
         if @new_resource.delim
-          @new_resource.value((new_values + current_values).uniq.join(@new_resource.delim))
+          if @new_resource.append
+            @new_resource.value((current_values + new_values).uniq.join(@new_resource.delim))
+          else
+            @new_resource.value((new_values + current_values).uniq.join(@new_resource.delim))
+          end
         end
         create_env
       end
--- a/lib/chef/provider/env.rb
+++ b/lib/chef/provider/env.rb
@@ -151,7 +151,7 @@ class Chef
       def modify_env
         if @new_resource.delim
           if @new_resource.append
-            @new_resource.value((current_values + new_values).uniq.join(@new_resource.delim))
+            @new_resource.value((current_values.delete_if{|value| new_values.include?(value) } + new_values).uniq.join(@new_resource.delim))
           else
             @new_resource.value((new_values + current_values).uniq.join(@new_resource.delim))
           end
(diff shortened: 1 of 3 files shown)


lamont-granquist:
needs `bundle exec chefstyle -a` run on it

MichaelPereira:
@lamont-granquist thanks for the tip, I've run it and fixed the reported issues.

lamont-granquist:

danielsdeleo:
Just wondering it it would be more consistent with other things in Chef for it to be action append or have an append flag as done here. :+1: on the code as written, assuming we like the shape of it.

ksubrama:
I'm actually a bit confused with the PR.

```
(current_values.delete_if { |value| new_values.include?(value) } + new_values).uniq.join(@new_resource.delim)
```

should produce the exact same results as

```
(new_values + current_values).uniq.join(@new_resource.delim)
```

The issue I had identified in the original bug report was that I didn't see us setting `delim` anywhere because both `current_values` and `new_values` only get split when `delim` is defined.

ksubrama:
The harder question to figure out is whether to prepend or append - that's why I hadn't gone in and fiddled with it.  Currently, `new_values + current_values` creates prepended values because `uniq` maintains order.  It would actually move any values to be added to the beginning of the list.  This might be a strange behavior for someone who tries to append "foo:bar" to "baz:foo:bar:qux" because regardless of whether we're using the code path in this PR or the older code path, as long as delim is defined, we'll get "foo:bar:baz:qux" or "baz:qux:foo:bar" - which to me is fairly confusing.

MichaelPereira:
@danielsdeleo : I thought about creating a new property with the `:modify` action, where it should be defined to one of the three following options: `:prepend`, `:append`, `:replace`. But it looks a lot more complicated to implement than the current proposition just to add the append mode

@ksubrama : Now that the default behavior has been to prepend, wouldn't it be dangerous to change it and possibly break existing cookbooks if the chef-client is upgraded?

lamont-granquist:
So I think if I was writing this today I'd want an action that took `:prepend` or `:append` and a property of `uniq` that took a boolean and twiddled the condition that @ksubrama is concerned about.

I'm kind of thinking we go down that route, and deprecate `:modify` entirely...

thommay:
Hi,
we're really grateful for this PR, but as Lamont says, we're more likely to reshape this provider and deprecate `:modify`. With that in mind, I'm going to close this PR. If you'd like to have a bash at creating a prepend action, that would be amazing and we'd love to help you there.
Thanks again,
-Thom

PR #4706, opened 2016-03-14
chefstyle Style/AndOr fixes

lamont-granquist:
this is part of our informal style guide, lets make it formal since
clearly its not getting followed very well.

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -480,7 +480,7 @@ class Chef::Application::Client < Chef::Application
   end
 
   def client_sleep(sec)
-    IO.select([ SELF_PIPE[0] ], nil, nil, sec) or return
+    IO.select([ SELF_PIPE[0] ], nil, nil, sec) || (return)
     @signal = SELF_PIPE[0].getc.chr
   end
 
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -480,7 +480,7 @@ class Chef::Application::Client < Chef::Application
   end
 
   def client_sleep(sec)
-    IO.select([ SELF_PIPE[0] ], nil, nil, sec) || (return)
+    return unless IO.select([ SELF_PIPE[0] ], nil, nil, sec)
     @signal = SELF_PIPE[0].getc.chr
   end
 
--- a/lib/chef/chef_fs/config.rb
+++ b/lib/chef/chef_fs/config.rb
@@ -270,7 +270,8 @@ class Chef
           end
           object_names.each do |object_name|
             # cookbooks -> cookbook_path
-            (singular_name = INFLECTIONS[object_name]) || raise("Unknown object name #{object_name}")
+            singular_name = INFLECTIONS[object_name]
+            raise("Unknown object name #{object_name}") unless singular_name
             variable_name = "#{singular_name}_path"
             paths = Array(@chef_config[variable_name]).flatten
             result[object_name] = paths.map { |path| File.expand_path(path) }
--- a/lib/chef/chef_fs/config.rb
+++ b/lib/chef/chef_fs/config.rb
@@ -271,7 +271,7 @@ class Chef
           object_names.each do |object_name|
             # cookbooks -> cookbook_path
             singular_name = INFLECTIONS[object_name]
-            raise("Unknown object name #{object_name}") unless singular_name
+            raise "Unknown object name #{object_name}" unless singular_name
             variable_name = "#{singular_name}_path"
             paths = Array(@chef_config[variable_name]).flatten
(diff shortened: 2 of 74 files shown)


thommay:
maximal :+1:

lamont-granquist:
@chef/client-core anyone have any thoughts or objections?  @danielsdeleo ?

lib/chef/application/client.rb
```diff
@@ -480,7 +480,7 @@ def interval_sleep(sec)
   end
 
   def client_sleep(sec)
-    IO.select([ SELF_PIPE[0] ], nil, nil, sec) or return
```
danielsdeleo: This is actually the correct idiomatic usage of `or`, for flow control instead of joining multiple conditions.

danielsdeleo:
A fair number of the things that got "fixed" were actually the correct usage (as a kind of postfix if statement), though lots of them were wrong. I don't know that you can actually enforce correct usage in an automated way. Though if the argument is that the correct usage of and/or is too subtle and we want to swear if off entirely, I suppose I could be convinced. Though in that case it would be nice to use a control flow statement, like `return unless IO.select(*args)` in the example I commented on above.

lamont-granquist:
i think i always use `return unless ...` -- at the beginning of the line i'm more likely to 'see' the return and its easier to scan down the 'inside' of the code to find early returns than it is to scan down the right margin.

lamont-granquist:
@danielsdeleo @thommay can you both look over d4579addae1dafa6b9e39b286995b97dac450aa2 ?

rubocop wasn't smart enough to autofix that, and i'm not sure i'm smart enough to fix the yum change correctly.

and i think that's an excellent example of why i'd prefer to ban AndOr entirely because i really don't want to have to try to decipher that.

lamont-granquist:
it looks like we only used `|| return` that one time in the code here.

lib/chef/chef_fs/config.rb
```diff
@@ -270,7 +270,7 @@ def object_paths
           end
           object_names.each do |object_name|
             # cookbooks -> cookbook_path
-            singular_name = INFLECTIONS[object_name] or raise "Unknown object name #{object_name}"
```
danielsdeleo: This one also needs a manual fix-up since it was a control flow usage

lib/chef/event_loggers/base.rb
```diff
@@ -42,8 +42,8 @@ def self.available_event_loggers
     end
 
     def self.new(name)
-      event_logger_class = by_name(name.to_s) or
-        raise UnknownEventLogger, "No event logger found for #{name} (available: #{available_event_loggers.join(', ')})"
+      (event_logger_class = by_name(name.to_s)) ||
+        raise(UnknownEventLogger, "No event logger found for #{name} (available: #{available_event_loggers.join(', ')})")
```
danielsdeleo: `raise unless`

lib/chef/formatters/base.rb
```diff
@@ -51,8 +51,8 @@ def self.available_formatters
     #--
     # TODO: is it too clever to be defining new() on a module like this?
     def self.new(name, out, err)
-      formatter_class = by_name(name.to_s) or
-        raise UnknownFormatter, "No output formatter found for #{name} (available: #{available_formatters.join(', ')})"
+      (formatter_class = by_name(name.to_s)) ||
```
danielsdeleo: raise unless

lib/chef/http/socketless_chef_zero_client.rb
```diff
@@ -177,7 +177,7 @@ def req_to_rack(method, url, body, headers)
 
       def to_net_http(code, headers, chunked_body)
         body = chunked_body.join("")
-        msg = STATUS_MESSAGE[code] or raise "Cannot determine HTTP status message for code #{code}"
+        (msg = STATUS_MESSAGE[code]) || raise("Cannot determine HTTP status message for code #{code}")
```
danielsdeleo: raise unless

lib/chef/policy_builder/policyfile.rb
```diff
@@ -256,10 +256,10 @@ def cookbook_lock_for(cookbook_name)
       # @api private
       def run_list
         if named_run_list_requested?
-          named_run_list or
-            raise ConfigurationError,
+          named_run_list ||
+            raise(ConfigurationError,
```
danielsdeleo: This whole file should be changed to `raise unless`

danielsdeleo:
https://github.com/chef/chef/commit/d4579addae1dafa6b9e39b286995b97dac450aa2 looks correct to me

lamont-granquist:
this is better, i had to leave some `@foo || raise("stuff")` style code because it was in accessors.  i kinda lean towards wanting parens anyway...

PR #4709, opened 2016-03-14
Setting init_command should be accepted instead of specific command overrides

coderanger:
Also remove `:enable` from that list because `enable_command` isn't a thing on the service resource. Not clear to me why `:enable` ever worked there.

Alternatively we could modify `custom_command_for_action?` to always return true if `init_command` is set, but that feels weirder to me.

Ping @chef/client-core @chef/client-enterprise-linux for review.

### DIFF
--- a/lib/chef/provider/service/redhat.rb
+++ b/lib/chef/provider/service/redhat.rb
@@ -60,6 +60,12 @@ class Chef
             a.failure_message Chef::Exceptions::Service, "#{chkconfig_file} dbleoes not exist!"
           end
 
+          requirements.assert(:enable) do |a|
+            a.assertion { !@service_missing }
+            a.failure_message Chef::Exceptions::Service, "#{new_resource}: Service is not known to chkconfig."
+            a.whyrun "Assuming service would be disabled. The init script is not presently installed."
+          end
+
           requirements.assert(:start, :reload, :restart) do |a|
             a.assertion do
               new_resource.init_command || custom_command_for_action?(action) || !@service_missing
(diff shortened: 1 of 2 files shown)


lamont-granquist:
:+1:

lib/chef/provider/service/redhat.rb
```diff
@@ -60,9 +60,15 @@ def define_resource_requirements
             a.failure_message Chef::Exceptions::Service, "#{chkconfig_file} dbleoes not exist!"
           end
 
-          requirements.assert(:start, :enable, :reload, :restart) do |a|
+          requirements.assert(:enable) do |a|
+            a.assertion { !@service_missing }
+            a.failure_message Chef::Exceptions::Service, "#{new_resource}: Service is not known to chkconfig."
+            a.whyrun "Assuming service would be disabled. The init script is not presently installed."
```
lamont-granquist: minor text nit that this should be "Assuming service would be *enabled*" rather than "disabled"

lib/chef/provider/service/redhat.rb
```diff
@@ -60,9 +60,15 @@ def define_resource_requirements
             a.failure_message Chef::Exceptions::Service, "#{chkconfig_file} dbleoes not exist!"
```
thommay: when you fix lamont's nit, can you please fix this message too? :)

lamont-granquist:
:+1: still

thommay:
:+1: once those nits are deloused.

PR #4711, opened 2016-03-15
ensure we always can rake

thommay:
cc @danielsdeleo @chef/client-core

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,9 @@ gem "activesupport", "< 4.0.0", group: :compat_testing, platform: "ruby"
 
 gem "chef-config", path: "chef-config" if File.exist?(__FILE__ + "../chef-config")
 
+# Ensure that we can always install rake, regardless of gem groups
+gem "rake"
+
 group(:docgen) do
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -5,6 +5,8 @@ gem "activesupport", "< 4.0.0", group: :compat_testing, platform: "ruby"
 
 gem "chef-config", path: "chef-config" if File.exist?(__FILE__ + "../chef-config")
 
+gem "rake"
```
danielsdeleo: Probably should have a comment to explain that we need this so rake is still available after bundling without the development group.

danielsdeleo:
:+1:

lamont-granquist:
should probably be ripped out of the gemspec or i think bundler whines about being defined twice...

but :+1:

thommay:
@lamont-granquist looks like bundler doesn't whine, so I'm gonna go ahead and merge

coderanger:
Bundler only objects if it's listed in the Gemfile itself twice.

PR #4714, opened 2016-03-15
Log platform along with client version and pid.

randomcamel:


coderanger:
Might be worth including it in the "pretty" error output too, since people usually just paste that into bug reports.

jrunning:
I've updated this PR re: @coderanger's comments. It now appends a `Platform` section to the pretty error output, e.g.:

```text
Platform:
---------
x86_64-darwin14.0
```

I also added some tests.

tyler-ball:
:+1:

randomcamel:
:+1: That is slick. =)

PR #4717, opened 2016-03-16
reverting version to 12.8.1 temporarily to avoid rerelease of ohai

mwrock:
want to get a chefdk build out quickly and ohai is causing a probe for chef-config 12.9.0. Will revert this back after successful chefdk build.


danielsdeleo:
:+1:

PR #4719, opened 2016-03-16
deprecate Chef::DSL::Recipe::FullDSL

lamont-granquist:
this module was created out of a bit of excessive paranoia instead of
simply adding mixins that we always use with Chef::DSL::Recipe directly
to Chef::DSL::Recipe.  lets try and YAGNI this class and just add it
directly.  these modules are already injected into everyone's recipes
and providers, so I can't imagine a huge amount of backcompat breaks.

### DIFF
--- a/lib/chef/dsl/recipe.rb
+++ b/lib/chef/dsl/recipe.rb
@@ -120,7 +120,7 @@ class Chef
         end
       end
 
-      # This module is deprecated and will be removed in Chef 13
+      # @deprecated Use Chef::DSL::Recipe instead, will be removed in Chef 13
       module FullDSL
         include Chef::DSL::Recipe
         extend Chef::Mixin::LazyModuleInclude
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@coderanger @jkeiser and anyone else in @chef/client-core ?

lamont-granquist:
@chef/client-core review?  anyone?  anyone?  bueller?

lib/chef/dsl/recipe.rb
```diff
@@ -105,27 +120,10 @@ def method_missing(method_symbol, *args, &block)
         end
       end
 
+      # This module is deprecated and will be removed in Chef 13
```
coderanger: If you make this `@deprecated Use Chef::DSL::Recipe instead, will be removed in Chef 13` it will show a nice banner in the YARD docs.

lamont-granquist: done

coderanger:
:+1:

lamont-granquist:
chef-zero build failure is unrelated -- it passed yesterday before i changed the comment...

PR #4722, opened 2016-03-18
use rb-readlin on unix as well as windows

lamont-granquist:
necessary for ruby-2.2.4


lamont-granquist:
@chef/client-core @chef/client-windows needs chef/omnibus-software#627

mwrock:
:+1:

ksubrama:
:+1:  Does that mean you can change the ruby software def to leave out libedit?

lamont-granquist:
@ksubrama i think you already found that -- yep.

PR #4725, opened 2016-03-18
rearrange dsl mixins

lamont-granquist:
- creates Chef::DSL::MethodMissing which will be nuked in Chef 13
- creates Chef::DSL::Core which is the mixins into core providers
- adds build_resource/define_resource to core providers

The intent here is to create Chef::DSL::Core which is intended to
become the DSL minus method_missing and things that are only intended
for recipes.  Right now I've only added build_resource/declare_resource
to core providers.  We should probably push more into the "Core" DSL
so that core providers could do things like use the search API.

I would have moved everything other than resources and definitions over
but the audit mode stuff is really only appropriate for recipes and not
provider code and probably should not be moved over.  We should probably
have an intermediate mixin which is the "Custom Provider" DSL that has
Resources and Definitions mixed into it, but which does not mix in the
audit stuff that should only go into the Recipe DSL.

There may also be a "Resources DSL" which has mixins for use with
not_if/only_if and I'm unclear if that should be the "Core DSL" or
another fourth mixin of DSL files.

### DIFF
--- a/lib/chef/dsl/core.rb
+++ b/lib/chef/dsl/core.rb
@@ -1,0 +1,42 @@
+#--
+# Author:: Adam Jacob (<adam@chef.io>)
+# Author:: Christopher Walters (<cw@chef.io>)
+# Copyright:: Copyright 2008-2016, 2009-2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@chef/client-core this is what I was thought `Chef::DSL::Recipe` and `Chef::DSL::Recipe::FullDSL` was trying to accomplish.

as we migrate things from Chef::DSL::Recipe into Chef::DSL::Core they will become available inside of core chef providers, and that'll reduce some of the confusion between how those are written.

lamont-granquist:
@chef/client-core review ping?  should be reasonably straight forwards...

thommay:
yeah, I think this is a sensible split :+1:

PR #4726, opened 2016-03-18
Allow any 3.x version of plist

danielsdeleo:
This matches the constraint in ohai, which will make Chef and ohai use the same version.


adamedx:
:+1:

PR #4727, opened 2016-03-18
Client option no_proxy does not work as advertised

chefsalim:
Update no_proxy usage in chef-config to conform to documentation:
https://docs.chef.io/proxies.html#no-proxy
http://www.gnu.org/software/emacs/manual/html_node/url/Proxies.html

@chef/client-core @randomcamel @tyler-ball

### DIFF
--- a/chef-config/chef-config.gemspec
+++ b/chef-config/chef-config.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "mixlib-shellout", "~> 2.0"
   spec.add_dependency "mixlib-config", "~> 2.0"
+  spec.add_dependency "fuzzyurl", '~> 0.8.0'
 
   spec.add_development_dependency "rake", "~> 10.0"
 
--- a/chef-config/chef-config.gemspec
+++ b/chef-config/chef-config.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "mixlib-shellout", "~> 2.0"
   spec.add_dependency "mixlib-config", "~> 2.0"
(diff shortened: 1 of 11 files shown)


chef-config/chef-config.gemspec
```diff
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "mixlib-shellout", "~> 2.0"
   spec.add_dependency "mixlib-config", "~> 2.0"
+  spec.add_dependency "fuzzyurl", '~> 0.8.0'
```
tyler-ball: This is released under the MIT license, is that okay for our usage?

tyler-ball:
:+1:

chef.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |s|
   s.add_dependency "diff-lcs", "~> 1.2", ">= 1.2.4"
 
   s.add_dependency "chef-zero", "~> 4.5"
+  s.add_dependency "fuzzyurl", '~> 0.8.0'
```
randomcamel: Is this necessary, since `chef` requires `chef-config` which already declares it, and it's really `chef-config` that uses it?

randomcamel:
One question, otherwise :+1:

chef-config/chef-config.gemspec
chefsalim: Yes, confirmed.

tyler-ball:
Nice!

sergiopatino:
Any idea what version of chef-client this made it into ?

tyler-ball:
@sergiopatino It made it into `12.9.38` - I'm not sure why it isn't in the CHANGELOG.md though...

PR #4730, opened 2016-03-21
Make example resource classes return strings for Class.to_s

danielsdeleo:
You only see these test failures if you run the recipe integration tests before running the provider resolver tests. I can't say why these did not start failing until recently, since there didn't seem to be any related change in the test configuration that triggered the test failures.


danielsdeleo:
ok, this can be repro'd locally by running `bundle exec rspec spec/integration/recipes/resource_load_spec.rb spec/unit/provider_resolver_spec.rb`

chefsalim:
:+1:

jkeiser:
:+1:

lamont-granquist:
:+1: :+1: :+1: :+1: :+1: :+1: :balloon: :birthday:

PR #4732, opened 2016-03-21
allow use_inline_resources for core chef providers

lamont-granquist:
* removes the DSL from InlineResources class
* ActionClass is now responsible for mixing the DSL into
  action classes
* apt_update provider converted over to use new syntax
   - does not need to mixin DeclareResource DSL
   - declares use_inline_resources
   - uses declare_resource instead of build_resource
   - converts def action_stuff to action :stuff
   - uses an execute resource instead of shell_out!
   - does not need the converge_by to update the action

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -390,9 +390,6 @@ class Chef
           end
         end
       end
-
-      require "chef/dsl/recipe"
-      include Chef::DSL::Recipe
     end
 
     protected
--- a/lib/chef/provider/apt_update.rb
+++ b/lib/chef/provider/apt_update.rb
@@ -62,17 +62,17 @@ class Chef
 
       def do_update
         [STAMP_DIR, APT_CONF_DIR].each do |d|
-          declare_resource(:directory, d, caller[0]) do
+          declare_resource(:directory, d) do
             recursive true
           end
         end
 
-        declare_resource(:file, "#{APT_CONF_DIR}/15update-stamp", caller[0]) do
+        declare_resource(:file, "#{APT_CONF_DIR}/15update-stamp") do
           content "APT::Update::Post-Invoke-Success {\"touch #{STAMP_DIR}/update-success-stamp 2>/dev/null || true\";};"
           action :create_if_missing
         end
 
-        declare_resource(:execute, "apt-get -q update", caller[0])
+        declare_resource(:execute, "apt-get -q update")
       end
 
     end
--- a/lib/chef/provider/apt_update.rb
+++ b/lib/chef/provider/apt_update.rb
@@ -38,12 +38,16 @@ class Chef
 
       action :periodic do
         if !apt_up_to_date?
-          do_update
+          converge_by "update new lists of packages" do
+            do_update
+          end
         end
       end
 
       action :update do
-        do_update
+        converge_by "force update new lists of packages" do
+          do_update
+        end
       end
 
       private
--- a/lib/chef/provider/apt_update.rb
+++ b/lib/chef/provider/apt_update.rb
@@ -22,7 +22,7 @@ require "chef/dsl/declare_resource"
 class Chef
   class Provider
     class AptUpdate < Chef::Provider
-      include Chef::DSL::DeclareResource
+      use_inline_resources
 
       provides :apt_update, os: "linux"
 
@@ -36,18 +36,14 @@ class Chef
       def load_current_resource
       end
 
-      def action_periodic
+      action :periodic do
         if !apt_up_to_date?
-          converge_by "update new lists of packages" do
-            do_update
-          end
+          do_update
         end
       end
 
-      def action_update
-        converge_by "force update new lists of packages" do
-          do_update
-        end
+      action :update do
+        do_update
       end
 
       private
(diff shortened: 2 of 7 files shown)


lamont-granquist:
@chef/client-core use_inilne_resources for core chef providers

lib/chef/provider/apt_update.rb
```diff
@@ -36,18 +36,14 @@ def whyrun_supported?
       def load_current_resource
       end
 
-      def action_periodic
+      action :periodic do
         if !apt_up_to_date?
-          converge_by "update new lists of packages" do
```
coderanger: Is there a reason to remove this? It's nice to have in the output.

lib/chef/provider/apt_update.rb
```diff
@@ -62,16 +58,17 @@ def apt_up_to_date?
...
 
-        build_resource(:file, "#{APT_CONF_DIR}/15update-stamp", caller[0]) do
+        declare_resource(:file, "#{APT_CONF_DIR}/15update-stamp", caller[0]) do
           content "APT::Update::Post-Invoke-Success {\"touch #{STAMP_DIR}/update-success-stamp 2>/dev/null || true\";};"
-        end.run_action(:create_if_missing)
+          action :create_if_missing
+        end
 
-        shell_out!("apt-get -q update")
+        declare_resource(:execute, "apt-get -q update", caller[0])
```
coderanger: Is there a strong reason to do this instead of using the DSL? Feels weird and excessively verbose.

lib/chef/resource.rb
```diff
@@ -1316,7 +1316,11 @@ class << self
     # life as well.
     @@sorted_descendants = nil
     def self.sorted_descendants
-      @@sorted_descendants ||= descendants.sort_by { |x| x.to_s }
+      # so it turns out Class.to_s just blindly returns the name of the class,
+      # which can be a Symbol in addition to being a String.  as a result, we
+      # may get a Symbol back from Class.to_s, so we call #to_s on that again.
+      # (buggy on at least ruby-2.4.4)
+      @@sorted_descendants ||= descendants.sort_by { |x| x.to_s.to_s }
```
coderanger: This is my new favorite line of code.

lib/chef/provider/apt_update.rb
lamont-granquist: It wasn't necessary.  I can add it back though.

lib/chef/resource.rb
danielsdeleo: This comment is incorrect. We have classes in the tests that define `resource_name` as a symbol and define `self.to_s` to return the resource name. See: https://github.com/chef/chef/pull/4730

lib/chef/provider/apt_update.rb
coderanger: It shows up in the formatter stream so I would vote to keep it.

lib/chef/provider/apt_update.rb
lamont-granquist: We don't have the Resources DSL mixed into core chef providers.  We tore it out of Chef 12.  There was one reason which was because the errors back from method missing were horrid.  We no longer use method_missing (or won't in Chef-13) so that rationale will be going away.  I thought that there was also some other use case that @danielsdeleo determined where having the DSL mixed into core providers allowed cookbooks to troll the internals of core providers.  I have forgotten the specifics of that, but until we dig that up again I'm not very inclined to mixin the DSL into core providers -- particularly in light of all the confusion caused by the 12.5 resource syntax.  For now I'm liking the separation that declare_resource provides.

coderanger: That's fair. If we could remove the `caller[0]` it would be nice though.

lib/chef/resource.rb
lamont-granquist: Isn't it though?  I tore out so much hair this morning I look like I could be a presidental candidate...

lib/chef/provider/apt_update.rb
lamont-granquist: @thommay why is that there?

lib/chef/resource.rb
lamont-granquist: ah okay that explains it...

lamont-granquist:
@chef/client-core okay fixed issues

lib/chef/provider.rb
```diff
@@ -390,9 +390,6 @@ def action_#{name}
           end
         end
       end
-
-      require "chef/dsl/recipe"
-      include Chef::DSL::Recipe
```
coderanger: Is this a compat break to remove?

coderanger: Like if you did `use_inline_resources` but then didn't actually use the `action` helper this would change the behavior. I'm okay with that, but it is a change I think.

coderanger: Also as I think about this, does this change the behavior on non-12.5-style (i.e. non-fused) providers using `use_inline_resources`?

lamont-granquist: Those should be covered by inheriting from LWRPBase:

https://github.com/chef/chef/blob/master/lib/chef/provider/lwrp_base.rb#L32

coderanger: Okay, but a `libraries/` based provider using `use_inline_resources` would see a behavior change I think. We should make sure that ends up in the changelog at a minimum.

lamont-granquist: I guess someone could have discovered that if you only inherit from Chef::Provider but you declare `use_inline_resources` that it mixes in the DSL.  But `use_inline_resources` was only defined on LWRPBase before 12.5 so that usage would be broken on 11.0-12.4.  We've also never documented that and always documented inheriting from LWRPBase in order to get the DSL.  So this would be reverting an unintended undocumented feature to the behavior of prior revisions, which should not be semver violating.

If it causes chaos we can back this out and introduce another directive that means the same thing but avoids mixin in the DSL, but I'd prefer not to go there until we prove that it actually is a problem (we are accumulating a large amount of cruft trying to solve backcompat issues that aren't issues anyone has ever hit before, and its making the code a lot harder to follow).

coderanger: It's super minor, shouldn't be a blocker just worth noting in case we get support requests :)

lamont-granquist:
i think this'll pass travis now that i've pulled dan's fixes to master in....

jkeiser:
I'm down. :+1:

PR #4734, opened 2016-03-21
We still test Chef with versions that do not include our Registry monkeypatches, so better documenting when we can remove these

tyler-ball:
\cc @randomcamel @chefsalim @btm @chef/client-windows

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,8 @@ platform:
 
 environment:
   matrix:
-    - ruby_version: "200-x64"
+    # 21-x64 is failing right now
+    #- ruby_version: "21-x64"
     - ruby_version: "21"
 
 clone_folder: c:\projects\chef
--- a/lib/chef/monkey_patches/win32/registry.rb
+++ b/lib/chef/monkey_patches/win32/registry.rb
@@ -28,25 +28,28 @@ module Win32
 
       module_function
 
-      if RUBY_VERSION =~ /^2\.1/
-        # ::Win32::Registry#delete_value is broken in Ruby 2.1 (up to Ruby 2.1.6).
-        # This should be resolved in a later release (see note #9 in link below).
-        # https://bugs.ruby-lang.org/issues/10820
-        def DeleteValue(hkey, name)
-          check RegDeleteValueW(hkey, name.to_wstring)
-        end
+      # ::Win32::Registry#delete_value uses RegDeleteValue which
+      # is not an imported function after bug 10820 was solved.  So
+      # we overwrite it to call the correct imported function.
+      # https://bugs.ruby-lang.org/issues/10820
+      # Still a bug in trunk as of March 21, 2016 (Ruby 2.3.0)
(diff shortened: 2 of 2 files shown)


lib/chef/monkey_patches/win32/registry.rb
```diff
@@ -22,15 +22,20 @@
...
     module API
 
       extend Chef::ReservedNames::Win32::API::Registry
 
       module_function
 
-      if RUBY_VERSION =~ /^2\.1/
-        # ::Win32::Registry#delete_value is broken in Ruby 2.1 (up to Ruby 2.1.6).
-        # This should be resolved in a later release (see note #9 in link below).
+      if GEM_VERSION_RUBY < GEM_VERSION_217 || GEM_VERSION_RUBY = GEM_VERSION_220
```
ksubrama: Why do you force set GEM_RUBY_VERSION = GEM_VERSION_220?

mwrock:
:+1: even though this class makes me want to cry

lib/chef/monkey_patches/win32/registry.rb
mwrock: oops. bet that was meant to be `==`

tyler-ball: :doh:

lib/chef/monkey_patches/win32/registry.rb
```diff
@@ -22,15 +22,20 @@
...
     module API
 
       extend Chef::ReservedNames::Win32::API::Registry
 
       module_function
 
-      if RUBY_VERSION =~ /^2\.1/
-        # ::Win32::Registry#delete_value is broken in Ruby 2.1 (up to Ruby 2.1.6).
-        # This should be resolved in a later release (see note #9 in link below).
+      if GEM_VERSION_RUBY < GEM_VERSION_217 || GEM_VERSION_RUBY == GEM_VERSION_220
```
ksubrama: Unfortunately, this is still hopelessly broken.  It seems the core ruby folks simply don't try out the windows registry API at all.

See https://github.com/ruby/ruby/blob/trunk/ext/win32/lib/win32/registry.rb

As of Feb 2015, they import the correct functions now - https://github.com/ruby/ruby/commit/a6516ff5f7f0daab88b44c64c8bd9dfa5317d6d0

But DeleteKey and DeleteValue now call non-existent methods because they weren't updated to call RegDeleteKeyW or RegDeleteValueW.  That's why your AppVeyor tests are failing with undefined constants.

mwrock:
:+1:

appveyor.yml
```diff
@@ -6,7 +6,6 @@ platform:
 
 environment:
   matrix:
-    - ruby_version: "200-x64"
```
ksubrama: Can you actually change this to 21-x64 so that we're prepared for first-class 64-bit client support.

ksubrama:
:+1:

PR #4735, opened 2016-03-21
Pass windows architecture to kitchen provisioner

mwrock:
This will allow us to force `i386` when calling from chefdk build


jkeiser:
:+1:

danielsdeleo:
:+1:

PR #4738, opened 2016-03-21
reset release notes for 12.9

mwrock:


coderanger:
Is there a reason we're wiping this file each release?

jkeiser:
There are release notes on the docs website, so we didn't want to dup the effort. :+1: (and we can definitely have a wider discussion about where release notes belong.)

coderanger:
The release notes on the docs are still one page for each release. Would be good to keep a single archive for faster search et al maybe?

mwrock:
Perhaps we could add a note at the top that adds a link to the online release notes and states that this is a placeholder for the current version. I personally think the changelog suffices as a one stop location for change history.

mwrock:
Added a comment at the top of the file to clarify the intent.

coderanger:
:+1: for now but I still this is weird :)

PR #4741, opened 2016-03-22
Notifications from LWRPS/sub-resources can trigger resources in outer run_context scopes

### LINKED ISSUE #4017: Add ability to notify from inside LWRP to wrapping resource_collections
This has wound up finally irritating me.

In the case where you've got multiple statements to manage config files (e.g plugins) but the underlying system correctly supports conf.d style management, there's no need to use an accumulator and no need to use a definition.  I'd like to be able to easily use an LWRP to get all the resource goodness out of that.

But if we break up our resources into service and plugin LWRPs then our recipe code looks like this:

```ruby
foo_plugin "bar1" do
  # stuff
  notifies :restart, "foo_service[instance]"
end

foo_plugin "bar2" do
  # stuff
  notifies :restart, "foo_service[instance]"
end

[...repeated for all the plugins...]

foo_service "instance" do
  action [ :enable, :start ]
end
```

I'd prefer to wrap that repeated notification into the LWRP to keep it DRY and so the user doesn't need to remember it and doesn't forget it.  While the code could be simplified by a loop in the trivial case, if the `#stuff` varies in each plugin then you can't easily use looping constructs to build the plugin resources.  We also *could* use a definition here instead to avoid building another run_context/resource_collection, but I'd prefer not to.

So, I'd like the ability to bubble up the notifications.

cc: @jkeiser @coderanger @danielsdeleo

lamont-granquist:
This is similar to poise's approach but has a few differences.

Similarly to poise, the base behavior of notifications and find() and
lookup() on the resource collection is changed to be 'recursive' and
to search in outer contexts for resources and will return them by
default.

There are find_local() and lookup_local() methods added to allow for
bypassing the recursion and making sure to throw exceptions if the
current run_context does not have any matching resources.

The CHEF-3694 resource cloning code has been modified to call the
lookup_local() API and not to be recursive because we believe that
nobody in their right mind would want that behavior (and resource
cloning should eventually be removed).  So the behavior of resource
cloning should remain unchanged.

The behavior of delayed notifications to resources outside of the current
run_context is slightly different than what Poise has been implementing.
The delayed notification will run in the run_context of the resource
that is being notified.  I think Poise tends to bubble up to the nextmost
wrapping resource context (as opposed to Poise's subcontext_block or
notifying_block contexts).  This code I think is conceptually simpler to
reason about, and I think it gets the use case right where if you're
notifying a service resource in the outermost run_context from within
multiple wrapping resources that it correctly bubbles out to the
outermost run context and will notify with all the other delayed
notifications at the end of the chef client run.

Another useful feature of the delayed notification behavior is that if
we do implement notifying_block or subcontext_block that each block can
get its own delayed notification run and any resources that are inside
of that block can run in the delayed notification phase of that block
(while still being able to notify resources outside of the block and
having those delayed notifications run in the receiving resources
run_context).  This will let us implement an often-requested feature for
having "notifications delayed to the end of a block/recipe" instead of
having to do all notifications absolutely immediately or delayed to the
end of the chef run.

This code also cleans up the object model a little bit.  All of the
state about notification collection is now hanging off of the
run_context -- the delayed_actions have been moved from the Chef::Runner
to the Chef::RunContext.  Hanging it off of the Chef::Runner would have
been very difficult to 'target' from other run_context's without adding
a pointer back from the RunContext to the Runner and that feels like the
wrong object model.  The RunContext is now responsible for all of its
notification state, while the Runner is responsible for wiring up
the notifications across different run_contexts.

Note that it will not be possible to send a notification to a
run_context which has already been converged.  That seems to make sense
to me and the search API on the resource collection does not support
returning resources from run_contexts that are children, only parents
(and we don't actually hold onto pointers to child run_contexts and
they may be garbage collected).

closes #4017

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,11 +149,11 @@ matrix:
       - sudo cat /var/log/squid3/access.log
 
   allow_failures:
+    - rvm: 2.2
+      env: "GEMFILE_MOD=\"gem 'poise', github: 'poise/poise'\""
+      script: bundle exec rake poise_spec
     - rvm: 2.3.0
     - rvm: rbx
-    - rvm: 2.2
-      env: "GEMFILE_MOD=\"gem 'halite', github: 'poise/halite'\""
-      script: bundle exec rake halite_spec
 
 notifications:
   on_change: true
--- a/lib/chef/resource_builder.rb
+++ b/lib/chef/resource_builder.rb
@@ -137,7 +137,7 @@ class Chef
       @prior_resource ||=
         begin
           key = "#{type}[#{name}]"
-          run_context.resource_collection.lookup(key)
+          run_context.resource_collection.lookup_local(key)
         rescue Chef::Exceptions::ResourceNotFound
           nil
         end
--- a/lib/chef/resource_collection.rb
+++ b/lib/chef/resource_collection.rb
@@ -33,9 +33,12 @@ class Chef
     extend Forwardable
 
     attr_reader :resource_set, :resource_list
-    private :resource_set, :resource_list
+    attr_accessor :run_context
 
-    def initialize
+    protected :resource_set, :resource_list
+
+    def initialize(run_context = nil)
+      @run_context = run_context
       @resource_set = ResourceSet.new
       @resource_list = ResourceList.new
     end
@@ -79,11 +82,50 @@ class Chef
 
     resource_list_methods = Enumerable.instance_methods +
       [:iterator, :all_resources, :[], :each, :execute_each_resource, :each_index, :empty?] -
-      [:find] # find needs to run on the set
-    resource_set_methods = [:lookup, :find, :resources, :keys, :validate_lookup_spec!]
+      [:find] # find overridden below
+    resource_set_methods = [:resources, :keys, :validate_lookup_spec!]
 
     def_delegators :resource_list, *resource_list_methods
     def_delegators :resource_set, *resource_set_methods
 
+    def lookup_local(key)
+      resource_set.lookup(key)
+    end
+
+    def find_local(*args)
+      resource_set.find(*args)
+    end
+
+    def lookup(key)
+      if run_context.nil?
+        lookup_local(key)
+      else
+        lookup_recursive(run_context, key)
+      end
+    end
+
+    def find(*args)
+      if run_context.nil?
+        find_local(*args)
+      else
+        find_recursive(run_context, *args)
+      end
+    end
+
+    private
+
+    def lookup_recursive(rc, key)
+      rc.resource_collection.resource_set.lookup(key)
+    rescue Chef::Exceptions::ResourceNotFound
+      raise if rc.parent_run_context.nil?
+      lookup_recursive(rc.parent_run_context, key)
+    end
+
+    def find_recursive(rc, *args)
+      rc.resource_collection.resource_set.find(*args)
+    rescue Chef::Exceptions::ResourceNotFound
+      raise if rc.parent_run_context.nil?
+      find_recursive(rc.parent_run_context, *args)
+    end
   end
 end
--- a/lib/chef/run_context.rb
+++ b/lib/chef/run_context.rb
@@ -130,6 +130,14 @@ class Chef
     #
     attr_reader :delayed_notification_collection
 
+    #
+    # An Array containing the delayed (end of run) notifications triggered by
+    # resources during the converge phase of the chef run.
+    #
+    # @return [Array[Chef::Resource::Notification]] An array of notification objects
+    #
+    attr_reader :delayed_actions
+
     # Creates a new Chef::RunContext object and populates its fields. This object gets
     # used by the Chef Server to generate a fully compiled recipe list for a node.
     #
@@ -152,6 +160,7 @@ class Chef
       @loaded_attributes_hash = {}
       @reboot_info = {}
       @cookbook_compiler = nil
+      @delayed_actions = []
 
       initialize_child_state
     end
@@ -172,10 +181,11 @@ class Chef
     #
     def initialize_child_state
       @audits = {}
-      @resource_collection = Chef::ResourceCollection.new
+      @resource_collection = Chef::ResourceCollection.new(self)
       @before_notification_collection = Hash.new { |h, k| h[k] = [] }
       @immediate_notification_collection = Hash.new { |h, k| h[k] = [] }
       @delayed_notification_collection = Hash.new { |h, k| h[k] = [] }
+      @delayed_actions = []
     end
 
     #
@@ -217,6 +227,18 @@ class Chef
         delayed_notification_collection[nr.name] << notification
       else
         delayed_notification_collection[nr.declared_key] << notification
+      end
+    end
+
+    #
+    # Adds a delayed action to the +delayed_actions+.
+    #
+    def add_delayed_action(notification)
+      if delayed_actions.any? { |existing_notification| existing_notification.duplicates?(notification) }
+        Chef::Log.info( "#{notification.notifying_resource} not queuing delayed action #{notification.action} on #{notification.resource}"\
+                       " (delayed), as it's already been queued")
+      else
+        delayed_actions << notification
       end
     end
 
@@ -640,6 +662,8 @@ ERROR_MESSAGE
         audits
         audits=
         create_child
+        add_delayed_action
+        delayed_actions
         delayed_notification_collection
         delayed_notification_collection=
         delayed_notifications
--- a/lib/chef/runner.rb
+++ b/lib/chef/runner.rb
@@ -30,13 +30,14 @@ class Chef
 
     attr_reader :run_context
 
-    attr_reader :delayed_actions
-
     include Chef::Mixin::ParamsValidate
 
     def initialize(run_context)
-      @run_context      = run_context
-      @delayed_actions  = []
+      @run_context = run_context
+    end
+
+    def delayed_actions
+      @run_context.delayed_actions
     end
 
     def events
@@ -48,16 +49,11 @@ class Chef
     def run_action(resource, action, notification_type = nil, notifying_resource = nil)
       # If there are any before notifications, why-run the resource
       # and notify anyone who needs notifying
-      # TODO cheffish has a bug where it passes itself instead of the run_context to us, so doesn't have before_notifications. Fix there, update dependency requirement, and remove this if statement.
-      before_notifications = run_context.before_notifications(resource) if run_context.respond_to?(:before_notifications)
-      if before_notifications && !before_notifications.empty?
-        whyrun_before = Chef::Config[:why_run]
-        begin
-          Chef::Config[:why_run] = true
+      before_notifications = run_context.before_notifications(resource) || []
+      unless before_notifications.empty?
+        forced_why_run do
           Chef::Log.info("#{resource} running why-run #{action} action to support before action")
           resource.run_action(action, notification_type, notifying_resource)
-        ensure
-          Chef::Config[:why_run] = whyrun_before
         end
 
         if resource.updated_by_last_action?
@@ -65,8 +61,8 @@ class Chef
             Chef::Log.info("#{resource} sending #{notification.action} action to #{notification.resource} (before)")
             run_action(notification.resource, notification.action, :before, resource)
           end
+          resource.updated_by_last_action(false)
         end
-
       end
 
       # Actually run the action for realsies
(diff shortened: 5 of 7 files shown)


lamont-granquist:
POC for discussion.

coderanger:
So I'm overall :+1: on this feature (my helpers have worked like this forever). This does have the weird problem that `notifies :restart, 'service[foo]'` works but `notifies :restart, resources(service: 'foo')` does not. We're not advertised the latter syntax in a long time buuuuut it's still out there. The other question is how to handle "scope shadowing", should a resource with a name matching another resource in a higher scope do some kind of cloning? Not doing it would be weirdly inconsistent in some ways but I hate resource cloning and want to purge it with fire anyway (and I think all other maintainers agree on that point). We should also sort out compat with the existing Poise helpers too since it would be nice if the two were cross-compatible.

lamont-granquist:
hmmm, yeah, i'm pretty 50/50 torn on how i feel like making `resources(service: 'foo')` be recursive.

i was trying to avoid making `#find` itself be recursive by default since that felt like a more sweeping change.  i could be convinced that we need one canonical way to search through the resource_collection(s) and retrieve a resource and should just make `#find` work this way and have consistency.  it makes for one thing to learn.

i would vastly prefer ripping out the existing resource cloning than adding more.   i can't see a good use for it, and it just injects providers with CHEF-3694 resource naming uniqueness issues.  one thing i like about sub-resource-collections is that 'execute "whatever"` being locally unique is good enough and it doesn't get into fights with identically named resources in other scopes.

coderanger:
For comparison, the way I did it was to patch `#find` so it always works the same and then to disable resource cloning for any resource using the Poise helpers. I basically want providers to work like local variables in a closure: when you create a resource it's always 100% local but if you try to read something that isn't in local scope it checks outer scopes. I think that's pretty intuitive all around.

lamont-granquist:
@coderanger this changes ResourceCollection#find to search recursively, keeps ResourceSet#find non-recursive (which was only accessible through a private attr_accessor :resource_set previously)

lamont-granquist:
@chef/client-core needs specs and such.  thoughts on the code?  i think the rationale here is that when we use `find` we set it up so that it  has the same semantics as notification--which is most likely what everyone wants by default.  we could expose "no, really just my own run_context" as an API but i'm unclear when you'd want that, and we can expose that after we find out what that use case actually is...

thommay:
I'm fine with `resources(service: 'foo')`just quietly breaking in this case. It's something we ought to properly deprecate in the 13 cycle.

lamont-granquist:
actually i think i just fixed that by the change to default `find` semantics...

lamont-granquist:
Oh, so the comment about resource cloning here is that now the resource cloning API will `find` resources in the outer context?  Yeah, okay...   Not Want...

coderanger:
:+1: This is the implementation I would have written :)

coderanger:
The only other thing I added in mine was a way to get the "global context" but I think that's better handled but `Chef.node.run_context.resource_collection` for the super rare situations that it is needed. I think I only use it for the `include_recipe` wrapper to make sure all the resources from that end up in the top-level collection.

lamont-granquist:
yeah, i was thinking about a helper that did something like:

```
def run_context(rc) 
  saved = @run_context
  @run_context = rc
  yield
  @run_context = saved
end
```

(so far that compiles in my brain, haven't tried it in code yet...)

Then you could wrap stuff with `run_context(Chef.run_context) do ... end` and all the code inside the block would twiddle with the root run_context...

probably add some sugar so that `run_context(:parent) do ... end` and `run_context(:root) do ... end` worked because that is most likely the two use cases people care about...

danielsdeleo:
My opinion is that you want to opt-in to notifying outside of your own scope for the same reason we end up using `module.const_get(:Name, inherit=false)` everywhere. The majority use case will still be, as it is now, for notifications in your own scope and it's generally when you are doing something more complicated that you want to notify outside of your scope. Of course when the programmer does everything right, this will just be extra "crap" you have to add to make your thing work, and that's annoying. What I consider worse, however, is the possibility that things accidentally "work," or at least compile and execute, without doing what you actually intended, because you got the name of a resource incorrect (maybe a copy-paste mistake or you misremembered it, or just didn't have enough coffee that day). Stuff like that is generally very frustrating to debug. If you're just explicit about wanting to notify to an enclosing scope, a lot of those cases become compile errors which you'll detect the first time you test your code.

lamont-granquist:
Actually i think the majority use case is where its notifying in your own scope, but you don't care either way.  Then after that what people will want is for it to bubble up.  I think the least-used use case will be to target only resources in your sub-resource collection and not bubble up.  I think in this case "inherit=false" will be used much less than "inherit=true".  We've also got some sugar around declare_resource(...create_if_missing: true) and i think we should make that a bit better and /that/ use case is probably the common one for not wanting to bubble up.

coderanger:
I think the service notification use case is so common and it is so expected that it work that that should form that guide for how we build the feature. With resource cloning removed as a concern (yay) I think the risk of typo collision is pretty low.

lamont-granquist:
@chef/client-core i think i've got coverage now.

one interesting thing is that notifications execute relative to the run_context of the triggerer.  so all the notifications fired off in a sub-resource_collection are executed before going back to execution in the parent run context.

some people might think that a `:delayed` notification from within a provider should execute the resource only at the end of the overall run and might be surprised to see it converge at the end of the provider's run?   but if you need to do that the solution is probably to idempotently add a resource to the root run_context and fire off a single delayed notification at the same time.  then later code can execute the same code and not re-add and re-trigger the resource.  and you shouldn't be trying to build that kind of operation by natively sending notification signals to resources from within LWRPs.

coderanger:
"idempotently add a resource to the root run_context" is not an easy thing to do. https://github.com/poise/poise/blob/master/lib/poise/subcontext/runner.rb#L34 is my code to move delayed notifications up to the parent runner.

coderanger:
Also not moving them up to the parent context means delayed notifications aren't correctly de-dup'd, you can end up with a service restarting a lot of times, etc.

jaymzh:
I'm :+1:  on the idea and the refactoring and the overall code... but I'm with @coderanger  - delayed notifications should happen in the parent context... for deduping and because that's what one expects.

lamont-granquist:
yeah i added specs for de-dup'd notifications to the parent run_context

jaymzh:
@lamont-granquist I owe you a beer for this. After this we can probably start using `use_inline_resources` and drop `notifying_action`. I'm excited. :)

lamont-granquist:
@coderanger i didn't copy your implementation.  it looked a little odd to me since it looks like you assumed that the notification should run in the parent context?  instead i queue'd it up in the run_context of the resource that is being notified.  i also moved the notifying_actions array off of the runner and onto the run_context -- its difficult to get a handle onto the runner that wraps the run_context and having the notification state split up feels very wrong.  i also didn't want to introduce another backpointer from the run_context to the runner.

code is ugly and copypasta'd a bit right now, and i think i want to take the resource de-duplication code and have it live on the run_context with all the various notification collections.

i don't think i've broken any APIs since the old delayed_actions array on the runner delegates to the run_context now and any callers shouldn't be able to tell the difference.

coderanger:
I'm almost definitely going to have to fix some stuff on the Poise side, I wouldn't worry about it if you're happy with the Chef side of things :)

lamont-granquist:
i need to clean up the API a bit today anyway and i'll try to sort out what you're doing vs. what i'm doing and see if there's a way to be graceful and not break things...

lamont-granquist:
@coderanger okay i think this is working code.  you've got a failing poise test.  i'm still looking at it, but somewhat confused about why notification in your subcontexts aren't firing...

lamont-granquist:
Okay so here's the deal I guess, what happens when we do this within a resource:

```ruby
action :foo do
  service "whatever" do
    action :nothing
  end
  template "/etc/whatever/stuff1.conf" do
    source "stuff.conf.erb"
    notifies :restart, 'service[whatever]', :delayed
  end
  template "/etc/whatever/stuff2.conf" do
    source "stuff.conf.erb"
    notifies :restart, 'service[whatever]', :delayed
  end
end
```

two things to consider:

1.  i really don't want to hang an :immediately off of the stuff1.conf template resource since without executing the stuff2.conf template the results may be crashy-crashy
2.  i doubt i want to throw the delayed notification all the way to the end of the chef-client run, since later recipes may depend on this service having been configured correctly and then bounced to pick up its config

i think we're all in agreement that this:

```ruby
action :foo do
  log "always" do
    # something in some outer run_context defines service[whatever] for me already
    notifies :restart, "service[whatever]", :delayed
  end
end
```

Should never run the restart as part of its own delayed notifications.  If `service[whatever]` is defined in the outermost run_context then that's where the notification should occur and it should get de-dup'd there.

I think where i'm going with this is that `:delayed` should be run with the delayed notifications of the run_context that it was defined in.  Making all delayed notifications run at the end of the chef client run is probably not right.

I /can/ see an edge case where users might want to override that though:

```ruby
action :foo do
  service "whatever" do
    action :nothing
  end
  log "always" do
    notifies :restart, "service[whatever]", :delayed
  end
end

action :bar do
  service "whatever" do
    action :nothing
  end
  log "always" do
    notifies :restart, "service[whatever]", :delayed
  end
end
```

When I'm trying to write that code I probably mean for the restart to only happen once, even if both actions are hit.  It may also be a bit complicated since the different actions may be called at different 'levels' so one might be called at the root run_context and one might be called in a sub-run_context from some wrapping resource or two.

That can be solved with this though:

```ruby
action :foo do
  declare_resource(:service, "whatever", run_context: Chef.run_context, create_if_missing: true) do
    action :nothing
  end
  log "always" do
    notifies :restart, "service[whatever]", :delayed
  end
end

action :bar do
  declare_resource(:service, "whatever", run_context: Chef.run_context, create_if_missing: true) do
    action :nothing
  end
  log "always" do
    notifies :restart, "service[whatever]", :delayed
  end
end
```

Then the service resource always gets declared in the root run_context, and the delayed notifications will always be delayed to the end of the run.

jaymzh:
I think that's pretty reasonable. As someone who uses these sorts of pattern a lot I almost always fall into the second category.

I think the syntax to force a resource up like that is useful, but in reality I think people who want that delayed notification only are putting that resource in the resource because you can't, today, notify outside of it. Once that's changed I think those people will put the resource in the parent context naturally anyway and you'll have almost none of the 3rd case anyway.

FWIW, in our case, for #1, we'd actually want the super-delayed behavior, but #1 is just a special case of #3 and we'd move the service resource.

coderanger:
Hmm, so what happens if the notification target is in a context that has already finished executing? I feel like that is a thing that could happen but my brain is struggling to figure out how.

lamont-granquist:
@coderanger that is *exactly* the bug that happens in the poise test that is breaking.

its a side effect of the way that you add the notification to the enclosing new_resource.    if you add the notification to the delayed_actions of the enclosing run_context, though, then your code could look something like this:

```ruby
      def run_delayed_notifications(error=nil)
        # If there is an error, just do the normal thing. The return shouldn't
        # ever fire because the superclass re-raises if there is an error.
        return super if error
        delayed_actions.each do |notification|
          if @resource.run_context.respond_to?(:add_delayed_action)
            @resource.run_context.add_delayed_action(notification)
          else
            notifications = run_context.parent_run_context.delayed_notifications(@resource)
            if notifications.any? { |existing_notification| existing_notification.duplicates?(notification) }
              Chef::Log.info( "#{@resource} not queuing delayed action #{notification.action} on #{notification.resource}"\
                             " (delayed), as it's already been queued")
            else
              notifications << notification
            end
          end
        end
      end
```

lamont-granquist:
that still works, the code still supports targeting notification from any arbitrary run_context (that you can get a handle onto) to any arbitrary resource in any run_context (that you can get a handle onto).  but the default behavior of the Chef::Runner class is to target the run_context of the resource that is the target.

coderanger:
:+1: I need to clean up my tests because TK 1.7 broke everything, but will fix in the morning.

lamont-granquist:
okay i think i like this.

i also like the fact that we can fix it in the Chef::Runner and that becomes the class responsible for "dispatching notifications" -- while managing the notification collections and state all move to the RunContext along with the api that consistently de-dups actions.  That feels like the separation of responsibilities is clearer now, and we can patch the behavior later in a single class (and Poise can continue to override the behavior and get basically the same effect as before).

coderanger:
I think the behavior you outlined is better than what I had, so I don't mind adapting my stuff to it. If the code accounts for the "delayed notifications in converged contexts" thang I think it should cover all the use cases I care about in a way that makes sense.

lamont-granquist:
@chef/client-core i think this is ready for review then

lamont-granquist:
@coderanger i think the only thing blocking this from merging now is that it'll make poise go red...  The change to your patch to #run_delayed_notifications would allow you to 100% preserve backcompat behavior.  If you want to break your own API and adopt this behavior I'm not certain how you'd want to go about that...

thommay:
:+1:

jaymzh:
:+1:

coderanger:
:+1: I'll take care of fixing my stuff tomorrow. This is awesome!

ranjib:
:+1:

lamont-granquist:
okay i'm throwing poise in the allowed failures matrix on travis so master doesn't go red when i whack merge, go ahead and take it back out whenever you're ready.

i also noticed halite was still there even though we made it green awhile back, so took that out.

Issue #4214, opened 2015-12-01
Subversion resource does not respect http_proxy settings


tyler-ball:
http://subversion.apache.org/faq.html#proxy

Unlike most Linux utilities which respect `ENV['http_proxy']` (and associated variables) the subversion binary does not.  It requires proxy settings to be added to the subversion servers configuration file.  On my machine this is located at `~/.subversion/servers`.  It requires adding the keys `http-proxy-host` and `http-proxy-port` under the `[global]` group.

If `Chef::Config[:http_proxy]` is set we need to push this down to the subversion configuration when a subversion resource is invoked.

thommay:
can we do `----config-option servers:global:http-proxy-port <port>` as a short term fix?

tyler-ball:
@thommay where are you reading about that??  That is much better than modifying a file!  That should just be our fix if it works.

I had no idea you could specify config on the command line - I'm not familiar with subversion and just used the first link when I googled `subversion http proxy`.

thommay:
yeah it's gnarly but do `svn co --help` and it's in there, and then you can figure out what the key is based on what you wrote above.

thommay:
(actually forget short term fix; if we don't have to poke around in `.subversion/servers` then we shouldn't

tanahlayang:
Hi pals,
Did we think that not all users need this proxy argument extension, (My Environment doesn't need it)?
I think the params *svn_arguments* has useful enough if someone *want override the svn configs*

The issue now,
If these arguments has forced in source code to add in chef subversion resource, How I want to disable it? cause as I said not everyone need these http-proxy

thanks

tas50:
It's not realistic for us to update the config on the fly in order to work around subversion. Instead we should make sure we get this into the documentation so that users understand what they need to do to use proxy servers with subversion.

PR #4744, opened 2016-03-22
Subversion provider should respect proxies

chefsalim:
See issue at https://github.com/chef/chef/issues/4214

This change allows the subversion provider to respect proxy environment variables, by adding an appropriate proxy configuration to the svn command line.

@chef/client-core @randomcamel @tyler-ball

### DIFF
--- a/lib/chef/provider/subversion.rb
+++ b/lib/chef/provider/subversion.rb
@@ -21,6 +21,7 @@
 require "chef/log"
 require "chef/provider"
 require "chef/mixin/command"
+require "chef-config/mixin/fuzzy_hostname_matcher"
 require "fileutils"
 
 class Chef
@@ -32,6 +33,7 @@ class Chef
       SVN_INFO_PATTERN = /^([\w\s]+): (.+)$/
 
       include Chef::Mixin::Command
+      include ChefConfig::Mixin::FuzzyHostnameMatcher
 
       def whyrun_supported?
         true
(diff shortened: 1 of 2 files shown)


tyler-ball:
:+1: Nice job!  Will you update the `proxy_tests` project to [uncomment](https://github.com/chef/proxy_tests/blob/ddbc8d67a40e79a2e4c6dad3db9544b2d3383f4b/files/default/scripts/repo/cookbooks/test/recipes/default.rb#L35-L37) the svn resource test after this is merged to master?

chefsalim:
Yep, will do.

lib/chef/provider/subversion.rb
```diff
@@ -194,6 +196,21 @@ def authentication
         result
       end
 
+      def proxy
+        %w{http https}.each { |p|
+          if @new_resource.repository.start_with?("#{p}:") &&
+              ENV["#{p}_proxy"] != nil &&
```
randomcamel: Don't we want to pull this from `Chef::Config` instead of `ENV`?

lib/chef/provider/subversion.rb
```diff
@@ -194,6 +196,21 @@ def authentication
...
+              !fuzzy_hostname_match_any?(URI.parse(@new_resource.repository).host, ENV["no_proxy"])
+            proxy_uri = URI.parse(ENV["#{p}_proxy"])
```
randomcamel: `Chef::HTTP::BasicClient#proxy_uri` does different processing for the proxy URI, it'd be good to at least check out the difference.

lib/chef/provider/subversion.rb
chefsalim: The Chef::Config proxy values will get exported, so we want to pick them up from env (to handle the case where chef config does not have a specified value).

randomcamel: Huh, I had it in my head it was bi-directional and Config would return ENV values, but it looks not.

randomcamel:
:+1:

PR #4745, opened 2016-03-22
don't remove trailing newlines on json output

kisoku:
permits ChefFS to produce json objects that don't make git complain

### DIFF
--- a/lib/chef/json_compat.rb
+++ b/lib/chef/json_compat.rb
@@ -114,7 +114,7 @@ class Chef
         options_map = {}
         options_map[:pretty] = true
         options_map[:indent] = opts[:indent] if opts.has_key?(:indent)
-        to_json(obj, options_map).chomp
+        to_json(obj, options_map)
       end
 
       # Map +json_class+ to a Class object. We use a +case+ instead of a Hash


kisoku:
This was suggested by @jkeiser after I complained about ChefFS gobbling my newlines when running 'knife download'

coderanger:
I feel like this is going to wreck all kinds of havoc given how deep this is in the code. Is there somewhere closer to the edge where we can re-add the newline?

jkeiser:
@coderanger I'm curious what sort of havoc? It feels like newlines are a good thing for editor software/etc.

@kisoku OTOH, if you want to just affect knife download and upload, go through all uses of this in `lib/chef/chef_fs` and add the newline at the end there.

lamont-granquist:
@jkeiser it causes 28 failing rspec tests at a minimum.

kisoku:
I can take a stab at fixing those 28 tests if you think it's worth the time and effort. If not I will focus on knife download

thommay:
Definitely feels like it would be better to fix up `knife` rather than something so deep in the core (that could be considered part of the API contract).

lamont-granquist:
yeah, i feel like for better or for worse this is part of our Chef::JSONCompat API and we can't touch it without making a breaking change, so :-1: on this approach.

i think the spec failures are likely a symptom of how much chaos this will cause if we let it loose in the world.

i'm more okay with fixing the output on the knife download with a more restricted change.

thommay:
I think given the concerns around bustage and our API we're gonna close this PR. thanks mate!

Issue #4043, opened 2015-10-07
remote_file resource needs to support sftp


adeconsulting:
I'm attempting to bootstrap an Ubuntu 12.04 LTS node using Chef 11.10.4.  The remote_file resource is failing because it does not support sftp.

coderanger:
This seems out of scope for a ticket. If someone makes a provider in a gem/cookbook we could talk about integrating it to core.

jkerry:
Attempting to answer this here.  feedback / pull requests are welcome
https://github.com/jkerry/sftp_provider

btm:
@jkerry any interest in submitting a PR to merge that into core?

yoshiwaan:
@coderanger I think this is something that is worth adding to core under the remote_file resource instead of a separate resource/provider. I feel that ideally the remote_file resource would be an abstraction over multiple drivers for different sources (like the service resource abstracts over different init systems).
It already supports http, https and ftp, why not scp, ftps and sftp also? It's certainly the most user friendly and they're commonly used protocols. There's also easy determination of type as prefixes are already used in the source attribute (i.e. http/https/ftp/file, so scp/ftps/sftp could be added).

coderanger:
@yoshiwaan Supporting it in `remote_file` is :+1: but that can be done in an external gem/cookbook for the initial development and to prove out that this is something people want enough to include in core.

jkerry:
Sorry for leaving you folks hanging.  I'm back on this now and a PR for core will be provided shortly.  working on getting test coverage of the code first.

PR #4750, opened 2016-03-24
Sftp remote file support

jkerry:
This is a PR to fulfill #4043 .  The sftp_provider cookbook has gotten at least one PR since creation and sftp usage is strictly required for my employer's security policy. I think there's enough community need to consider a PR into core now. Additionally the sftp_provider cookbook creates an awkward gem dependency race condition that is addressed if that dependency is delivered with the chef-client install. I structured the spec to be as similar to the ftp spec as possible where appropriate to ensure the interface is similar.

Notable gaps include:
* network proxy capability
* ssh key usage ( username and password in uri are enforced )

These might be addressable but I'm not sure at this time if they are inside the capability of the net-sftp gem.

### DIFF
--- a/lib/chef/provider/remote_file/sftp.rb
+++ b/lib/chef/provider/remote_file/sftp.rb
@@ -1,6 +1,6 @@
 #
-# Author:: Jesse Campbell (<hikeit@gmail.com>)
-# Copyright:: Copyright 2013-2016, Jesse Campbell
+# Author:: John Kerry (<john@kerryhouse.net>)
+# Copyright:: Copyright 2013-2016, John Kerry
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/lib/chef/provider/remote_file/sftp.rb
+++ b/lib/chef/provider/remote_file/sftp.rb
@@ -77,8 +77,6 @@ class Chef
           if filename.length == 0 || filename.end_with?( "/" )
             raise ArgumentError, "no filename: #{path.inspect}"
           end
-
-          @directories, @filename = directories, filename
         end
 
         def validate_userinfo!
--- a/lib/chef/provider/remote_file/sftp.rb
+++ b/lib/chef/provider/remote_file/sftp.rb
@@ -36,26 +36,19 @@ class Chef
           @new_resource = new_resource
           @current_resource = current_resource
           validate_path!
+          validate_userinfo!
         end
 
         def hostname
           @uri.host
         end
 
-        def user
-          if uri.userinfo
-            URI.unescape(uri.user)
-          else
-            'anonymous'
-          end
+        def port
+          @uri.port
         end
 
-        def pass
-          if uri.userinfo
-            URI.unescape(uri.password)
-          else
-            nil
-          end
+        def user
+          URI.unescape(uri.user)
         end
 
         def filename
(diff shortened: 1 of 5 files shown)


mhedgpeth:
John is right, the gem dependency would be great to be handled by the core chef & this would make things for us at NCR much more simple with dealing with SFTP resources.

jkerry:
Doh!  I'll get on those build failures shortly

coderanger:
As I say every time this shows up: put it in a cookbook, if it becomes popular then nominate it for core inclusion. With the new gem metadata there is literally no reason to add this to core as a first step.

jkerry:
Does the gem metadata support gem installation from local file system cache? I can't find anything that says so. Our PCI compliant products are still stuck in the same situation since they have to include gem dependencies in the cookbook payload and are restricted from network resources.

It feels like a no-brainer to expand remote_file access protocols, especially more secure protocols but I'm sure every PR you guys get looks like a no-brainer to the author

jkerry:
and the cookbook is here:
https://github.com/jkerry/sftp_provider

coderanger:
Adding any feature to Chef core is an increased management and maintenance load, so it always has to be weighed against the magnitude of the benefit to users. Given the limitations of this implementations (notably that it requires plain-text passwords be floating around in the recipe) I can't see this being a security win for most people and so the end benefit is going to be minimal. Using this effectively would require having an existing secrets management/distribution platform and in most cases you should just use that for access secure data instead of SFTP :) There are totally cases where you would want SFTP that aren't amenable to tools like Vault or KeyWhiz, like moving big files around, but that is a less common use case (most people would use HTTPS for that situation). Anyways, hope that explains my resistance.

coderanger:
Basically mutually authenticated file transfer under heavy automation is a Hard Problem and I don't think this fits the bill well as it requires a lot of support services Chef does not provide (well).

lamont-granquist:
@thommay we should look at this in PR review

jkerry:
@lamont-granquist if there's any information I can provide let me know.  I closed it out for the time being but I can re-open a PR with the rubocop warnings cleaned up at a later date if there's need.

jkerry:
The rubocop failures are on files not touched in this PR.  What's the best practice for that?  I don't mind fixing them but I don't want to pollute the PR either.

lamont-granquist:
could you rebase on top of master and force push?  the merge commits are making this PR pretty ugly...

jkerry:
Yessir, got caught up in holiday stuff.  working on it

jkerry:
Back to just my changes.  My apologies for the mess.

jkerry:
Known issue:  the copyright information on lib/chef/provider/remote_file/sftp.rb reads Jesse Campbell as it does in the ftp resource.  I'll switch that to my name when I get back to the desk.

lib/chef/provider/remote_file/sftp.rb
```diff
@@ -0,0 +1,118 @@
...
+        end
+
+        def user
+          URI.unescape(uri.user)
+        end
+
+        def filename
+          parse_path if @filename.nil?
+          @filename
+        end
```
lamont-granquist: this looks copied from the ftp class

the thing is that in `#get` you've removed the one reference to this method.  the `filename` in `parse_path` is actually a local variable that gets constructed (hella confusingly -- the ftp class could use a bit of cleanup to make it clearer) and therefore both this `#filename` code and the `#parse_path` code are unused in this class that you've written and could simply be deleted.

lamont-granquist:
other than the dead code i think this looks fine.  :+1:

thommay:
Yeah. I'm :+1: after a bit of a spring clean.

jkerry:
It is used as a path validation mechanism (utilized directly in the tests).  I can probably validate the path more directly and fulfill the same interface

jkerry:
@thommay @lamont-granquist is this what you had in mind?

lib/chef/provider/remote_file/sftp.rb
```diff
@@ -0,0 +1,108 @@
...
+            d.gsub!(/%([0-9A-Fa-f][0-9A-Fa-f])/) { [$1].pack("H2") }
+          }
+          unless filename = directories.pop
+            raise ArgumentError, "no filename: #{path.inspect}"
+          end
+          if filename.length == 0 || filename.end_with?( "/" )
+            raise ArgumentError, "no filename: #{path.inspect}"
+          end
+
+          @directories, @filename = directories, filename
```
thommay: think you can still ditch this line.

jkerry: Derp.  Pushing a removal now

thommay:
:+1: LGTM

lamont-granquist:
:shipit:

chef-supermarket:

PR #4751, opened 2016-03-24
Add shorthand :syslog and :win_evt for log_location config

jrunning:
Adds `:syslog` and `:win_evt` as possible values for the `log_location` option for easier configuration via node attributes.

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -141,6 +141,7 @@ class Chef
     # that a user has configured a log_location in client.rb, but is running
     # chef-client by hand to troubleshoot a problem.
     def configure_logging
+      configure_log_location
       Chef::Log.init(MonoLogger.new(Chef::Config[:log_location]))
       if want_additional_logger?
         configure_stdout_logger
@@ -149,6 +150,20 @@ class Chef
     rescue StandardError => error
       Chef::Log.fatal("Failed to open or create log file at #{Chef::Config[:log_location]}: #{error.class} (#{error.message})")
       Chef::Application.fatal!("Aborting due to invalid 'log_location' configuration", 2)
+    end
+
+    # Turn `log_location :syslog` and `log_location :win_evt` into the
+    # appropriate loggers.
+    def configure_log_location
+      log_location = Chef::Config[:log_location]
+      return unless log_location.respond_to?(:to_sym)
+
+      Chef::Config[:log_location] =
+        case log_location.to_sym
+          when :syslog then Chef::Log::Syslog.new
+          when :win_evt then Chef::Log::WinEvt.new
+          else log_location # Probably a path; let MonoLogger sort it out
+        end
     end
 
     def configure_stdout_logger
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -247,6 +247,33 @@ describe Chef::Application do
 
         it_behaves_like "log_level_is_auto"
       end
+
+      describe "log_location" do
+        shared_examples("sets log_location") do |config_value, expected_class|
+          context "when the configured value is #{config_value.inspect}" do
+            let(:logger_instance) { instance_double(expected_class).as_null_object }
+
+            before do
+              allow(expected_class).to receive(:new).and_return(logger_instance)
+              Chef::Config[:log_location] = config_value
+            end
+
+            it "it sets log_location to an instance of #{expected_class}" do
+              expect(expected_class).to receive(:new).with no_args
+              @app.configure_logging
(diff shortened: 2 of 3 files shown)


tyler-ball:
:+1: Code LGTM

coderanger:
Is there a reason to coerce to syms instead of a strings? Usually better to go the other way because symbols are permanent. Doesn't really matter though.

lib/chef/application.rb
```diff
@@ -151,6 +152,20 @@ def configure_logging
       Chef::Application.fatal!("Aborting due to invalid 'log_location' configuration", 2)
     end
 
+    # Turn `log_location :syslog` and `log_location :win_evt` into the
+    # appropriate loggers.
+    def configure_log_location
+      log_location = Chef::Config[:log_location]
+      return unless log_location.respond_to?(:to_sym)
```
randomcamel: Doesn't this represent some kind of invalid configuration that we should explode on, so the user isn't surprised?

spec/unit/application_spec.rb
```diff
@@ -247,6 +247,31 @@
 
         it_behaves_like "log_level_is_auto"
       end
+
+      describe "log_location" do
+        shared_examples("sets log_location") do |config_value, expected_class|
```
randomcamel: Nice.

lib/chef/application.rb
```diff
@@ -151,6 +152,20 @@ def configure_logging
...
+
+      Chef::Config[:log_location] =
+        case log_location.to_sym
```
randomcamel: Same question: if we're not intending to allow something like `log_location "win_evt"`, I feel like we should reject it.

coderanger: We should definitely support that because `chef-client -L syslog` should be a thing :)

lib/chef/application.rb
jrunning: No, this just returns early when `Chef::Config[:log_location]` isn't a string or symbol, so if it's e.g. STDIN or a Logger object it won't be changed.

lib/chef/application.rb
jrunning: ^Yeah, that was my thinking. Does that make sense, @randomcamel?

btm: in that case could you add a test for the string behavior too? looks super easy with the test framework you wrote.

randomcamel: @jrunning Yup, thanks. Also what btm said.

randomcamel:
:+1: with a spec for the string behavior.

jrunning:
@randomcamel @btm I added specs for the string behavior. Tests were failing because of Rubocop failures fixed in #4755. Fingers crossed they'll pass again now.

jrunning:
This (and other PRs) are failing on a poise issue right now, coderanger says a fix will be released later today.

coderanger:
Should be fixed now if you kick the build.

btm:
This looks unrelated to your changes:

```
1) ChefSpec::Macros#stub_node sets the automatic attributes from a JSON data path
     Failure/Error: fauxhai = Fauxhai.mock(options).data
     
     Fauxhai::Exception::InvalidPlatform:
       You specified a path to a JSON file on the local system that does not exist: '/path/to/json'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai/mocker.rb:50:in `block in fauxhai_data'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai/mocker.rb:78:in `call'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai/mocker.rb:78:in `fauxhai_data'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai/mocker.rb:26:in `initialize'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai.rb:12:in `new'
     # /home/travis/.rvm/gems/ruby-2.2.0/gems/fauxhai-3.2.0/lib/fauxhai.rb:12:in `mock'
     # ./lib/chefspec/macros.rb:176:in `stub_node'
     # ./spec/unit/macros_spec.rb:81:in `block (3 levels) in <top (required)>'
```

jrunning:
@btm Yeah, ChefSpec and Fauxhai are having a diagreement; I filed an issue: https://github.com/sethvargo/chefspec/issues/687

PR #4752, opened 2016-03-24
ruby 2.0.0 is EOL

lamont-granquist:
Ruby 2.0.0 maintenance ended Feb. 24th, 2016.

### DIFF
--- a/chef.gemspec
+++ b/chef.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.email = "adam@chef.io"
   s.homepage = "http://www.chef.io"
 
-  s.required_ruby_version = ">= 2.0.0"
+  s.required_ruby_version = ">= 2.1.0"
 
   s.add_dependency "chef-config", "= #{Chef::VERSION}"
 


lamont-granquist:
@chef/client-core RIP ruby 2.0.0

lamont-granquist:
we'll want to make similar commits to ohai and the mixlibs and such...

danielsdeleo:
Did we ever make an official RFC or something about our policy on EOL'd ruby versions?

lamont-granquist:
https://github.com/chef/chef-rfc/blob/master/rfc034-ruby-193-eol.md

and

https://github.com/chef/chef-rfc/blob/master/rfc015-chef-12.md

its not written quite as clear as i'd like right now, but the waffling seems to surround the fact that we were creating wiggle room to support 2.0.0 on windows for as long as we were stuck with it.

the way the RFC's are written it looks like we could drop support on 2.1 as well since 2.2 is what we seem to be calling 'N-1'.  i don't think we need to do that until 2.1 EOL's though (and its been less than an hour since i merged the ruby-2.2 branch...)

danielsdeleo:
:+1:

tyler-ball:
:+1:

btm:
I'm so glad we don't support Ruby 1.9.5 anymore.

:+1:

jkeiser:
We should dump the Travis 2.0 tests too if we haven't already

lamont-granquist:
Those are already gone

lamont-granquist:
(which is actually a good reason to get this done -- because we have no coverage and don't want to ship broken 2.0 code that 2.0 clients can still slurp down)

PR #4753, opened 2016-03-24
Set inherit=false on the fallback provider constant lookup.

coderanger:
This prevents really weirdo behavior with relative constant lookups like `Chef::Provider.const_get('Resource')` == `ChefVaultCookbook::Resource` which causes all kinds of fun.

Ruby docs tells me this param should exist on all supported rubies so probably just an oversight that we didn't do this before. We should also possibly consider restricting this fallback check to classes under `Chef::Resource` if `class.name` is set (/cc @jkeiser @lamont-granquist for that bit?). Ping @chef/client-core for review.


lamont-granquist:
yeah this sounds familiar...   probably just oversight...

lamont-granquist:
:+1:

PR #4755, opened 2016-03-25
supresses parser gem errors

lamont-granquist:
see https://github.com/whitequark/parser/issues/283

if rubocop works around this bug then we can revert this patch.


lamont-granquist:
yolo

PR #4759, opened 2016-03-28
Add optional integer argument for --daemonize option

### LINKED ISSUE #3305: Need Config/CLI options to move interval+splay sleep to end of client loop
Chef Client 12 does not Fork Chef instance to converge when running under daemon mode. It runs after specified interval of 1800 seconds but not as soon as we start service
With Chef Client 11. It runs as soon as we start chef-client service
We were having chef 11.4.4 on node. We did chef client 12 upgrade by installing omnibus package

We have chef-client running as daemon every 30 mins.

We are expecting chef-client to run as soon as we bootstrap node or start service but it seems that with chef 12 client it is not Forking chef instance to converge as soon as we start service. I have captured debug logs 

Below is example with Chef 12

Example:

```text
chef-client -d -c /etc/chef/client.rb -P /var/run/chef/client.pid -l debug -L /tmp/chef.log

# Logfile created on 2015-04-29 12:27:13 -0400 by logger.rb/44203
[2015-04-29T12:27:13-04:00] INFO: Daemonizing..
[2015-04-29T12:27:13-04:00] INFO: Forked, in 7978. Privileges: 0 0
[2015-04-29T12:27:13-04:00] DEBUG: Sleeping for 1800 seconds
```

I think we should be expecting  it to Fork chef instance to converge and should not wait for 1800 seconds

 issue seems to be at logic https://github.com/chef/chef/blob/12.2.1/lib/chef/application/client.rb#L384

It should run immediately for first time in daemonized mode

jrunning:
The `--daemonize` option takes an optional integer argument for the number of seconds to wait before the first daemonized run.

Closes #3305

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -105,10 +105,12 @@ class Chef::Application::Client < Chef::Application
 
   unless Chef::Platform.windows?
     option :daemonize,
-      :short => "-d",
-      :long => "--daemonize",
-      :description => "Daemonize the process",
-      :proc => lambda { |p| true }
+      :short => "-d [WAIT]",
+      :long => "--daemonize [WAIT]",
+      :description =>
+        "Daemonize the process. Accepts an optional integer which is the " \
+        "number of seconds to wait before the first daemonized run.",
+      :proc => lambda { |wait| wait =~ /^\d+$/ ? wait.to_i : true }
   end
 
   option :pid_file,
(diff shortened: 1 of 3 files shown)


lib/chef/application/client.rb
```diff
@@ -105,10 +105,12 @@ class Chef::Application::Client < Chef::Application
 
   unless Chef::Platform.windows?
     option :daemonize,
-      :short => "-d",
-      :long => "--daemonize",
-      :description => "Daemonize the process",
-      :proc => lambda { |p| true }
+      :short => "-d [WAIT]",
```
chefsalim: Will this have backward compatibility implications and break existing scripts?

lib/chef/application/client.rb
```diff
@@ -105,10 +105,12 @@ class Chef::Application::Client < Chef::Application
...
+      :long => "--daemonize [WAIT]",
+      :description =>
+        "Daemonize the process. Accepts an optional integer which is the " \
+        "number of seconds to wait before the first daemonized run.",
+      :proc => lambda {|wait| wait =~ /^\d+$/ ? wait.to_i : true }
```
chefsalim: Feels a bit wierd to have both bool or int valued option...

jrunning: It doesn't feel weird to me (lots of programs have command line options with optional arguments), but I don't have a strong opinion if others think it should be a separate option. @tyler-ball and I tried `--daemonize-run-immediate` before settling on this implementation.

lamont-granquist:
I'm not sure this addresses #3305 though does it?  The problem there is that if you have a non-zero splay and/or interval that you'll always sleep-before-running and this is the behavior that those users do not want to have at all -- they still want splay/interval options, but when present, they want the first run to be immediate and then for the chef-client to sleep afterwards.  So a '--sleep-after-run' kind of flag to shift the interval/splay sleeps to the end of the loop.

jrunning:
@lamont-granquist Unless I'm misunderstanding, this addresses the issue. If they pass `--daemonize 0` then the first run will happen immediately, followed by the first sleep.

lamont-granquist:
Ah okay, so i think the problem here is that you're missing all the exception decoration that we normally do around the chef-client run:

https://github.com/chef/chef/blob/jr/daemonize-immediate-run/lib/chef/application/client.rb#L457-L467

I think when you're sleeping you also need to have the sleep wrapped around with this logic or i suspect there's some edge conditions you might break:

https://github.com/chef/chef/blob/jr/daemonize-immediate-run/lib/chef/application/client.rb#L446-L453

jrunning:
@lamont-granquist Okay, that makes sense to me. Thanks!

jrunning:
@lamont-granquist Updated re: your comments; please let me know if you have any other feedback.

lamont-granquist:
:+1:

PR #4766, opened 2016-03-29
add ruby-2.3.0 to travis matrix

lamont-granquist:


coderanger:
Bring on the pain :+1:

PR #4767, opened 2016-03-29
Add option to set knife ssh timeout

chefsalim:
This adds a config to 'knife ssh' to allow a connection timeout to be specified.
This will prevent knife ssh from hanging if one of the hosts is down or not reachable.

@chef/client-core @tyler-ball


tyler-ball:
There should be a corresponding `DOC_CHANGES` modification but otherwise :+1:

chefsalim:
Yep, will add DOC_CHANGES

Issue #4126, opened 2015-11-01
git submodules are not downloaded with git resource


puneetloya:
_From @puneetloya on October 29, 2015 1:29_

git tests_dir do
  repository node['xyz']['repo']
  revision node['xyz']['branch']
  enable_submodules true
  action :sync
end

It is downloading the main parent repository but not the submodules. I am getting only the .gitmodules file and nothing else. Am I missing something very basic?

_Copied from original issue: jssjr/git#83_

puneetloya:
_From @someara on October 29, 2015 16:15_

Hello! The "git" resource ships in Core Chef, not this cookbook. Can you please file this at github.com/chef/chef ?

lamont-granquist:
Could use more information - version of chef-client, debug output from when the resource runs, etc.

joshburt:
I believe I understand the problem here.  I'm working on a re-write of the git HWRP and have included the fix in that code base.

A stand-alone version of the git HWRP is here:  https://github.com/joshburt/jgit

Once the work is complete, I'll create pull requests for the enhancements.  If you'd like to review the changes before-hand they are here:
https://github.com/joshburt/jgit/blob/master/libraries/provider_jgit.rb#L190
https://github.com/joshburt/jgit/blob/master/libraries/provider_jgit.rb#L207

**Problem Description**
We are currently preping the submodules like this:
`git submodule sync`
and we should be doing this:
`git submodule init`

Additionally passing `--recursive` to clone commands has the same affect and has been added to my code base.  Updates were made with respect to this: https://git-scm.com/book/en/v2/Git-Tools-Submodules

Timeline:
The code base is complete.  Tests are being updated now.  Once tests are updated I'll submit the pull request and comment back here.

Thanks!

joshburt:
I originally thought this was an issue.  I reviewed the code and the documentation, and then tested it.  I no longer believe this is an issue with Chef.


So, first a recap of how we are handling submodules within the git HWRP.  The below logic is derived from https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L169

On action sync, or checkout we run these git commands:
```
git submodule sync
git submodule update --init --recursive
```


From the official documentation: https://git-scm.com/docs/git-submodule

> 
> init
[...]

Issue #3929, opened 2015-09-18
git resource should prune remote


rmoriz:
The git resource currently does not prune the remote and will crash on conflicts.

Steps to reproduce:
- create a dummy git repo
- create a branch called "crap"
- deploy the repo/branch with chef:

``` ruby
git '/tmp/crap' do
  repository 'https://github.com/…/repo'
  revision 'crap'
  enable_checkout false
  action :sync
end
```
- delete the remote branch "crap"
- create a new remote branch "crap/feature"
- deploy the repo/branch with chef:

``` ruby
git '/tmp/crap' do
  repository 'https://github.com/…/repo'
  revision 'crap/feature' # <--- 
  enable_checkout false
  action :sync
end
```

Error:

```
       [2015-09-18T13:23:16+00:00] ERROR: git[/tmp/crap] (cookbook line 1) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
       ---- Begin output of git fetch origin && git fetch origin --tags && git reset --hard b30662fb7fcd4a6c38c9d7113ff8dccfea844625 ----
       STDOUT: 
       STDERR: error: 'refs/remotes/origin/crap' exists; cannot create 'refs/remotes/origin/crap/new'
       From https://github.com/…/repo
        ! [new branch]      crap/new   -> origin/crap/new  (unable to update local ref)
       error: some local refs could not be updated; try running
        'git remote prune origin' to remove any old, conflicting branches

       Ran git fetch origin && git fetch origin --tags && git reset --hard b30662fb7fcd4a6c38c9d7113ff8dccfea844625 returned 1
       [2015-09-18T13:23:17+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)

```

IMHO either by default or by option there should be a way to tell the resource to do a "git remote prune origin" before fetching.

Fix suggestion:

``` patch
diff --git i/lib/chef/provider/git.rb w/lib/chef/provider/git.rb
index 8418f22..c60801a 100644
--- i/lib/chef/provider/git.rb
+++ w/lib/chef/provider/git.rb
@@ -181,7 +181,7 @@ class Chef
         setup_remote_tracking_branches(@new_resource.remote, @new_resource.repository)
         converge_by("fetch updates for #{@new_resource.remote}") do
           # since we're in a local branch already, just reset to specified revision rather than merge
[...]

PR #4771, opened 2016-03-30
git HWRP Code Quality Enhancements

joshburt:
This PR applies a number of code quality enhancements to the git HWRP.

There are no business logic changes in this PR.  It's simply the basis of 3 more PRs to come.
The acceptance/merging of this PR will allow the other three to be submitted independently.

Future PRs Will Cover:
     A) Support for uploadpack.allowReachableSHA1InWant introduced in git 2.5
     B) Bug Fix: CHEF-3929 https://github.com/chef/chef/issues/3929
     C) Bug Fix: CHEF-4126 https://github.com/chef/chef/issues/4126

I'd love some help with validation, feedback, etc. Thanks,
Josh

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -107,7 +107,7 @@ class Chef
           git_standard_executor(["checkout", @new_resource.checkout_branch], run_opts)
           Chef::Log.info "#{@new_resource} checked out branch: #{@new_resource.revision} onto: #{@new_resource.checkout_branch} reference: #{sha_ref}"
         end
-end
+      end
 
       def enable_submodules
         if @new_resource.enable_submodules
@@ -231,7 +231,7 @@ end
         else
           Chef::Log.debug "#{@new_resource} checkout destination #{cwd} already exists or is a non-empty directory"
         end
-end
+      end
 
       def action_export
         action_checkout
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -1,5 +1,6 @@
 #
 # Author:: Daniel DeLeo (<dan@kallistec.com>)
+# Author:: Joshua Burt (<joshburt@shapeandshare.com>)
 # Copyright:: Copyright 2008-2016, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
@@ -24,7 +25,6 @@ require "fileutils"
 class Chef
   class Provider
     class Git < Chef::Provider
-
       provides :git
 
       def whyrun_supported?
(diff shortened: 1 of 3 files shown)


joshburt:
It looks like the travis ci run had an issue downloading dependencies and failed.  I'm going to trigger a new build by closing and reopening this PR.  Thanks.

joshburt:
Any update on this?
Thanks!

lib/chef/provider/git.rb
```diff
@@ -34,99 +34,55 @@ def whyrun_supported?
       def load_current_resource
         @resolved_reference = nil
         @current_resource = Chef::Resource::Git.new(@new_resource.name)
-        if current_revision = find_current_revision
-          @current_resource.revision current_revision
-        end
+
+        current_revision = find_current_revision
+        @current_resource.revision current_revision if current_revision
```
lamont-granquist: i actually like the previous form of this better, i don't think there's any edge conditions that it improves.

lib/chef/provider/git.rb
```diff
@@ -218,21 +161,115 @@ def remote_matches?(remote_url, check_remote_command_result)
...
+        git_command = git(args)
+        Chef::Log.debug "> #{git_command}"
+        shell_out!(git_command, run_options(run_opts))
+      end
+
+      def sha_hash?(string)
+        string =~ /^[0-9a-f]{40}$/
+      end
+
+      def action_checkout
```
lamont-granquist: moving this down here makes this hard to review.  you also seem to have made this private when it should be public?

lib/chef/provider/git.rb
```diff
@@ -218,21 +161,115 @@ def remote_matches?(remote_url, check_remote_command_result)
...
       end
 
-      alias :revision_slug :target_revision
+      def git(*args)
+        ["git", *args].compact.join(" ")
+      end
+
+      def git_standard_executor(args, run_opts = {})
+        git_command = git(args)
+        Chef::Log.debug "> #{git_command}"
```
lamont-granquist: this should probably say "running #{git_command}" or something slightly more verbose

lib/chef/provider/git.rb
```diff
@@ -218,21 +161,115 @@ def remote_matches?(remote_url, check_remote_command_result)
...
+      def cwd
+        @new_resource.destination
       end
 
-      alias :revision_slug :target_revision
+      def git(*args)
+        ["git", *args].compact.join(" ")
+      end
+
+      def git_standard_executor(args, run_opts = {})
```
lamont-granquist: this is a mouthful, it could just be `git_command` or `run_git_command`

lib/chef/provider/git.rb
```diff
@@ -137,28 +93,18 @@ def add_remotes
...
-          args << "-o #{remote}" unless remote == "origin"
-          args << "--depth #{@new_resource.depth}" if @new_resource.depth
-          args << "--no-single-branch" if @new_resource.depth && git_minor_version >= Gem::Version.new("1.7.10")
-
-          Chef::Log.info "#{@new_resource} cloning repo #{@new_resource.repository} to #{@new_resource.destination}"
-
-          clone_cmd = "git clone #{args.join(' ')} \"#{@new_resource.repository}\" \"#{@new_resource.destination}\""
-          shell_out!(clone_cmd, run_options)
+        converge_by("clone from #{@new_resource.repository} into #{cwd}") do
+          clone_by_advertized_ref
```
lamont-granquist: should be `clone_by_advertised_ref` (oddly enough British and American English both agree here)

lib/chef/provider/git.rb
lamont-granquist: also should use *args and then drop all the arrays off of the calls to it.

lamont-granquist: i think we can use `def foo(*args, **opts)` now since we only support ruby-2.1 or greater.

lamont-granquist:
this is pretty difficult to review.  the code rearrangement makes it very difficult to tell from the diff what is really going on.  stuff like pulling clone_by_advertsed_ref out into its own method call when there's no duplicated code just means that there's big blocks of red and green with a lot of distance between them.  there's other style fixes which are fairly meaningless which distract from the rest of the cleanup and its not clear which ones you really care about for later code and which are just personal reference (like tweaking `+` to concat strings vs. `\` to concat strings).

we are already cleaning up the entire chef codebase with `http://github.com/chef/chefstyle` and will most likely eventually hit the style issues you're tweaking here and will make a decision then about the right way to do the style fixes so changing the style either way just adds more lines to process in the PR and we may wind up making a decision to revert all those kinds of fixes anyway...

lib/chef/provider/git.rb
```diff
@@ -282,49 +312,58 @@ def rev_search_pattern
       end
 
       def git_ls_remote(rev_pattern)
-        command = git(%Q{ls-remote "#{@new_resource.repository}" "#{rev_pattern}"})
-        shell_out!(command, run_options).stdout
+        stdout_obj = git_standard_executor ["ls-remote", "\"#{@new_resource.repository}\"", "\"#{rev_pattern}\""]
```
lamont-granquist: you should be able to make this look like `git_standard_executor("ls-remote", new_resource.repository, rev_pattern)` and eliminate all the interpolation

lib/chef/provider/git.rb
```diff
@@ -249,11 +286,8 @@ def remote_resolve_reference
         # Using such a degenerate annotated tag would be very
         # confusing. We avoid the issue by disallowing the use of
         # annotated tags named 'HEAD'.
-        if rev_search_pattern != "HEAD"
-          found = find_revision(refs, @new_resource.revision, "^{}")
-        else
-          found = refs_search(refs, "HEAD")
-        end
+
+        found = rev_search_pattern != "HEAD" ? find_revision(refs, @new_resource.revision, "^{}") : refs_search(refs, "HEAD")
```
lamont-granquist: 👍

lib/chef/provider/git.rb
```diff
@@ -266,11 +300,7 @@ def find_revision(refs, revision, suffix = "")
       end
 
       def rev_match_pattern(prefix, revision)
-        if revision.start_with?(prefix)
-          revision
-        else
-          prefix + revision
-        end
+        revision.start_with?(prefix) ? revision : prefix + revision
```
lamont-granquist: 👍

joshburt:
Lots of feedback!  
I know this code base is old, and it needs some love.  In addition to this clean up I want to add a lighter weight shallow clone option using features introduced in git 2.5.  This was the real motivation behind the cleanup represented within this PR.  

Above and beyond the new feature I took the time to review all open issues related to git and address those (there are 2).  I’ve submitted one bug fix, and one enhancement, but the last one relies on more than one PR before it can be applied.  I’ll get that submitted afterwards.

Ultimately, the code break down allowed for better understanding of what this code base was doing; which was very difficult before (i.e. such things as multiple routes to shell out to git, inconstant usage of some functions).  It was hard to determine how high level constructs were expressed as raw git commands especially when consumed by other resources such as ‘deploy’.  Additionally this breakdown allows for easier test creation as there are no longer large blocks of logic.  As Chef practices TDD this should be a huge win. 

As for the items you consider individual styling choices, all modifications were a result of feedback from the standard rubocop tests. While the chef project ignores many rubocop tests as per chef-style it doesn’t mean that they aren’t valid, or that the resulting code isn’t better for adhering to them.

Not only does this PR passes all spec tests and chef-style tests, it also passes many more standard rubocop tests.  Anyone who works on this later will be hands down be better off even it more changes are applied to it later.  Its code; I’d be surprised it wasn’t enhanced over time. ;)

I’ll respond to your individual code comments in-line, and based on your feedback I’ll put together an actionable list of tasks to bring into our next sprint

lamont-granquist:
So I agree that its much better to consolidate the multiple routes to shell_out, etc.  We definitely need those fixes.

The biggest issues I had in readability were things like the way that `clone` had its body abstracted out to `clone_by_advertized_ref` when nothing else called it, and then it was moved to a distant location in the file.  That creates big blocks of red and green in the diff which aren't remotely close to each other which simply makes the diff hard to read.  At the very least it'd be nice to keep the code together with the extracted `clone_by_advertized_ref` immediately following `clone` (you can mark just that method private as well).  I think that actually makes it more readable since you don't have to flip between the wrapper and the implementation as you're reading it.  I also don't think its particularly necessary though since if nothing else calls it the abstraction isn't terribly useful -- the body of `clone` becomes super trivial and i don't think `clone_by_advertized_ref` adds a whole lot of readability to the class.

Here's sort of an example of both things:

https://github.com/chef/chef/blob/686113531d23f30e9973d659c456ae33eb9cff1f/lib/chef/provider/package.rb#L83-L116

I didn't bother breaking out the logic in action_install because its still basically three "paragraphs" and they're readable.  You could extract three methods out of there to 'simplify' action_install but then it becomes too simple so that its just three method calls and wouldn't really add anything of value.  Of course if preseed_package and install_package were included in this method then it would start to become unweildy so even apart from the fact that those need to be subclass implementations, it makes sense to extract those out.  With your method and only a simple converge_by wrapping the implementation it doesn't offer much to extract the implementation.  And then you see there how I've kept install_description close to the only place that its used and just marked it private, which makes it easier to read the whole thing, and you don't have to go surfing through the file.

And the problem with the rubocop fixes is that it becomes really distracting with style changes.  It might be better to submit a PR just with the rubocop style changes run across the file -- ideally two commits with the autocorrected ones separated out from the manually fixed one (if there are any) and then we could see what that looks like and merge that first.  Then we could analyze the bigger rearrangements you want to do to clean stuff up here.

And part of the problem is that I absolutely don't trust our spec tests in this provider, so we really need to read the code and grok what all the changes do before hitting merge even if things are green...

joshburt:
@lamont-granquist thanks for taking the time to elaborate for my benefit.  I know it wasn't the most exciting use of your time.

I've started the work to break down this PR into smaller PRs for review based on your feedback.

joshburt:
Work to break this into other PRs in underway.  Taking this as advise for submission of those new requests.  Closing this PR.  Thanks!

PR #4776, opened 2016-03-31
Spec break on Windows due to temp dir and short path names

adamedx:
This issue only surfaces on Windows with Ruby 2.2 only when the user name is longer than 8 characters -- the Windows TEMP environment variable will be set to a path that includes the user's name since it's contained in the user's profile directory, but it will force the usage of an 8 character file name. This causes an issue in the test because while Ruby has always used TEMP when creating temp dirs. used by our tests, some Chef code that operates on it used to also return such (improperly) shortened names, but as of Ruby 2.2 which has been used for Chef only starting in the last week it does not, perhaps due to some improvement in other parts of Ruby.

This fixes the tests so that long file names are expected. However, there may be a need to investigate whether the new Ruby 2.2 behavior could cause a regression in a separate issue.

### DIFF
--- a/spec/support/chef_helpers.rb
+++ b/spec/support/chef_helpers.rb
@@ -84,6 +84,28 @@ def canonicalize_path(path)
   windows? ? path.tr("/", '\\') : path
 end
 
+# Makes a temp directory with a canonical path on any platform.
+# Only really needed to work around an issue on Windows where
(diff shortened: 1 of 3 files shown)


mwrock:
:+1: ... :cry:

tyler-ball:
:+1:

danielsdeleo:
:+1:

PR #4779, opened 2016-04-01
Revert "bump to ruby 2.2.4"

chefsalim:
This reverts commit 3cb10016d142c45d5b2409ecb337b1b7bc295046.

Going back to Ruby 2.1 until we can get chef-acceptance passing.

@adamedx @lamont-granquist @chef/client-core


adamedx:
Change looks good, but I think we need to revert this also @chefsalim: https://github.com/chef/chef/commit/e25b75e9fcd26ccc9c9094a43b614306de1047db. It assumes that Chef methods are going to return long path names, which is true in 2.2, false in 2.1.

PR #4780, opened 2016-04-01
acceptance tests doesn't work for me without this patch

lamont-granquist:
I'm using rvm, backed off to ruby-2.1.8, I have the magical line:

```
export APPBUNDLER_ALLOW_RVM=true
```

But it fails with:

```
TRIVIAL::PROVISION::[2016-03-31 18:00:21 -0700] /Users/lamont/.rvm/rubies/ruby-2.1.8/lib/ruby/site_ruby/2.1.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'chef' (>= 0) among 149 total gem(s) (Gem::LoadError)
```

Which makes perfect sense to me if you don't have chef itself in the
bundle in this directory...


lamont-granquist:
@chef/client-core i find this a bit confusing since i don't understand how they run at all without this...

tyler-ball:
Paging @sersut @patrick-wright @chefsalim - what was the reason that we intentionally didn't include Chef in this Gemfile in the first place?

chefsalim:
Yeah we used to but it broke something else - I can't recall exactly what though

sersut:
Do not recall what was the reason :frowning:. Sorry.

patrick-wright:
@chefsalim You removed it: https://github.com/chef/chef/commit/abe85b3bdf5e50518e9be959ba3accf6afb8da70#diff-cf70f0f005146a26bb099acdb735c41fL4

PR #4782, opened 2016-04-01
Add apt_repository resource

thommay:
:construction: needs more tests, but basic review appreciated :construction:

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -1,0 +1,255 @@
+#
+# Author:: Thom May (<thom@chef.io>)
+# Copyright:: Copyright (c) 2016 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/resource"
+require "chef/dsl/declare_resource"
+require "chef/mixin/shell_out"
+require "chef/http/simple"
+require "chef/provider/noop"
+
+class Chef
+  class Provider
+    class AptRepository < Chef::Provider
+      use_inline_resources
+
+      include Chef::Mixin::ShellOut
+
+      provides :apt_repository do
+        uses_apt?
+      end
+
+      def whyrun_supported?
+        true
+      end
+
+      def load_current_resource
+      end
+
+      action :add do
+        unless new_resource.key.nil?
+          if is_key_id?(new_resource.key) && !has_cookbook_file?(new_resource.key)
+            install_key_from_keyserver
+          else
+            install_key_from_uri
+          end
+        end
+
+        declare_resource(:execute, "apt-cache gencaches") do
+          ignore_failure true
+          action :nothing
+        end
+
+        declare_resource(:apt_update, new_resource.name) do
+          ignore_failure true
+          action :nothing
+        end
+
+        components = if is_ppa_url?(new_resource.uri) && new_resource.components.empty?
+                       "main"
+                     else
+                       new_resource.components
+                     end
+
+        repo = build_repo(
+          new_resource.uri,
+          new_resource.distribution,
+          components,
+          new_resource.trusted,
+          new_resource.arch,
+          new_resource.deb_src
+        )
+
+        declare_resource(:file, "/etc/apt/sources.list.d/#{new_resource.name}.list") do
+          owner "root"
+          group "root"
+          mode "0644"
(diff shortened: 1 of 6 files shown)


lib/chef/provider/apt_repository.rb
```diff
@@ -0,0 +1,244 @@
...
+
+      def whyrun_supported?
+        true
+      end
+
+      def load_current_resource
+      end
+
+      action :add do
+        converge_by "add apt repository for #{new_resource.name}" do
```
coderanger: This means this resource will always be marked as updated which doesn't seem right?

lib/chef/resource/apt_repository.rb
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+#
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class AptRepository < Chef::Resource
+      resource_name :apt_repository
+      provides :apt_repository, os: "linux", platform_family: "debian"
```
coderanger: Is there a reason we keep doing this on platform-specific resources? It makes unit testing a little more annoying and I'm not sure of the gain.

thommay:
@chef/client-core ready for review…

lib/chef/resource/apt_repository.rb
```diff
@@ -0,0 +1,47 @@
...
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class AptRepository < Chef::Resource
+      resource_name :apt_repository
+      provides :apt_repository
+
+      property :repo_name, String, name_property: true, regex: [/^([a-z]|[A-Z]|[0-9]|_|-|\.)+$/]
```
lamont-granquist: this smells a bit like another brittle regex?

lib/chef/resource/apt_repository.rb
```diff
@@ -0,0 +1,47 @@
...
+      property :uri, String
+      property :distribution, String, default: lazy { node["lsb"]["codename"] }
```
lamont-granquist: as a bit of an aside why do we need to lazy these?  this really shouldn't ever be getting updated on a node.   i could see this with `node['fqdn']` and stuff that might be changed during the chef-client run, but mutating this attribute during a chef-client run would be really, really weird...

lamont-granquist:
looks like you addressed https://github.com/chef-cookbooks/apt/issues/191 ?

lib/chef/provider/apt_repository.rb
```diff
@@ -0,0 +1,253 @@
...
+        paths.any? { |path| ::File.executable?(::File.join(path, "apt-get")) }
+      end
+
+      def is_key_id?(id)
+        id = id[2..-1] if id.start_with?("0x")
+        id =~ /^\h+$/ && [8, 16, 40].include?(id.length)
+      end
+
+      def extract_fingerprints_from_cmd(cmd)
+        so = Mixlib::ShellOut.new(cmd, env: { "LANG" => "en_US", "LANGUAGE" => "en_US" })
```
lamont-granquist: not entirely clear why you're going to Mixlib::ShellOut here instead of shell_out()?

thommay:
it should fix chef-cookbooks/apt#191 ; chef-cookbooks/apt#192 ; chef-cookbooks/apt#200 at least. But I don't want to actually close them until i port this code back to the cookbook.

lamont-granquist:
just a couple of questions, but :+1: overall

lib/chef/provider/apt_repository.rb
```diff
@@ -0,0 +1,255 @@
...
+        paths.any? { |path| ::File.executable?(::File.join(path, "apt-get")) }
+      end
+
+      def is_key_id?(id)
+        id = id[2..-1] if id.start_with?("0x")
+        id =~ /^\h+$/ && [8, 16, 40].include?(id.length)
+      end
+
+      def extract_fingerprints_from_cmd(cmd)
+        so = shell_out(cmd, env: { "LANG" => "en_US", "LANGUAGE" => "en_US" })
```
lamont-granquist: i think you can actually drop the lang stuff since this code should have you completely covered:

https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L872-L922

including handling the case of servers that have no `en_US.utf-8` or `C.utf-8` support and falls back to C and yells at the user:

https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L911

lib/chef/resource/apt_repository.rb
```diff
@@ -0,0 +1,47 @@
...
+
+class Chef
+  class Resource
+    class AptRepository < Chef::Resource
+      resource_name :apt_repository
+      provides :apt_repository
+
+      property :repo_name, String, name_property: true
+      property :uri, String
+      property :distribution, String, default: lazy { node["lsb"]["codename"] }
```
eherot: I love that this was added to core Chef, but...

@thommay Having this as a default is not a good idea because some Apt repos do not use a distribution and having this here makes it impossible to use those things. Also, it is a different behavior than the existing `apt` cookbook, which causes it to break backwards compatibility.

lamont-granquist: i belive this property is nillable so its not impossible?

lamont-granquist: fixing here: https://github.com/chef/chef/pull/4832

PR #4784, opened 2016-04-01
if password provided at command line, pass into sudo command

### LINKED ISSUE #4783: If a password is provided on the command line, sudo commands should not prompt for a password
When I run a sudo command using `knife ssh`, I am prompted for a password [(`Enter your password`)](https://github.com/chef/chef/blob/master/lib/chef/knife/ssh.rb#L337). If I logged into the node with the password argument to the `-P` flag rather than using an ssh key, the utility should already know my password and it should be passed into the sudo prompt automatically.

I dont know of a situation where a user's sudo password would be different than a password used to login to via ssh, so I think sudo should never prompt for a password when its provided at the command line.

Current state:
```shell
$ knife ssh name:mynode 'hostname' -P mypassword
mynode mynode.mydomain.com
$ knife ssh name:mynode 'sudo hostname' -P mypassword
mynode knife sudo password:
Enter your password:
mynode
mynode mynode.mydomain.com
```

Desired state:
```shell
$ knife ssh name:mynode 'hostname' -P mypassword
mynode mynode.mydomain.com
$ knife ssh name:mynode 'sudo hostname' -P mypassword
mynode knife sudo password:
using sudo password provided at command line...
mynode
mynode mynode.mydomain.com
```

I will take a look at implementing this, if I did not explain the problem clear enough, please tag me in a comment and I will try to clarify

atheiman:
resolves #4783

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -74,6 +74,13 @@ class Chef
         # without a value
         :default => false
 
+      option :ssh_password_for_sudo,
+        :short => "-S",
+        :long => "--ssh-password-for-sudo",
+        :description => "EXPERIMENTAL. If ssh password given as argument to -P and a sudo COMMAND is " +
+                        "specified, this password will be passed into sudo using the -S flag",
+        :default => false
+
       option :ssh_port,
         :short => "-p PORT",
         :long => "--ssh-port PORT",
@@ -279,7 +286,13 @@ class Chef
       end
 
       def fixup_sudo(command)
-        command.sub(/^sudo/, 'sudo -p \'knife sudo password: \'')
+        if config[:ssh_password_ng] && config[:ssh_password_for_sudo]
+          # if ssh password provided on command line, then pass it into sudo
+          command.sub(/^sudo/, "echo '#{config[:ssh_password_ng]}' | sudo -S -p ''")
+        else
+          # prompt for sudo password
+          command.sub(/^sudo/, 'sudo -p \'knife sudo password: \'')
+        end
       end
 
       def print_data(host, data)


atheiman:
im still going to add a flag to knife ssh to enable this functionality, just wanted to get this PR started.

atheiman:
based on the [`get_password`](https://github.com/chef/chef/blob/master/lib/chef/knife/ssh.rb#L333) function, `knife ssh` already assumes that the ssh login password and the sudo password will always be the same thing.

lib/chef/knife/ssh.rb
```diff
@@ -74,6 +74,13 @@ class Ssh < Knife
         # without a value
         :default => false
 
+      option :ssh_password_for_sudo,
+        :short => "-S",
+        :long => "--ssh-password-for-sudo",
+        :description => "EXPERIMENTAL. If ssh password given as argument to -P and a sudo COMMAND is " +
+                        "specified, this password will be passed into sudo using the -S flag",
```
atheiman: this looks kind of ugly when its printed out as help because its so long. Is there an easy way to indent a second line of an option description or should I write a shorter description? Any thoughts on what that description should be?

atheiman:
I think all i need is a test and then this is ready for review.

lib/chef/knife/ssh.rb
lamont-granquist: You should drop EXPERIMENTAL from the description.

lib/chef/knife/ssh.rb
```diff
@@ -279,7 +286,13 @@ def session_from_list(list)
       end
 
       def fixup_sudo(command)
-        command.sub(/^sudo/, 'sudo -p \'knife sudo password: \'')
+        if config[:ssh_password_ng] && config[:ssh_password_for_sudo]
```
lamont-granquist: can you do `@password = config[:ssh_password_ng] if config[:ssh_password_for_sudo]`?

that will set `@password` and will avoid prompting, and then you'll fall through all the logic that we already have to find the `knife sudo password: ` string on the pty and send the password (which you definitely should not be bypassing -- its all there for $REASONS).

the code line to set `@password` probably should not be in fixup_sudo() and that command should be left alone.

not positive there won't be other side effects but you should go down that path rather than the one you're going down here.

atheiman: Yes i can make that change and try it out, thanks for the feedback!

lamont-granquist:
closed via #5427

PR #4785, opened 2016-04-01
Remove override of number of workers to be used.

edolnx:
This should be set in the make scripts and cookbooks for the
esoteric platforms that need it. Overriding this to 10 workers
on all platforms is slowing things down and causing massive
resource constrains and failures on smaller platforms (like ARM).

### DIFF
--- a/omnibus/omnibus.rb
+++ b/omnibus/omnibus.rb
@@ -44,10 +44,6 @@ build_retries 3
 fetcher_retries 3
 fetcher_read_timeout 120
 
-# We limit this to 10 workers to eliminate transient timing issues in the
-# way Ruby (and other components) compiles on some more esoteric *nixes.
-workers 10
-
 # Load additional software
 # ------------------------------
 # software_gems ['omnibus-software', 'my-company-software']


edolnx:
cc @chef/engineering-services @chef/client-core

stevendanna:
Note for reviews: This defaults to the number of CPUs (as reported by ohai) + 1

edolnx:
Also, if ohai can't figure out the number of CPUs (like running in a Solaris Zone) then the code falls back to a sane default of 3. This can also be set via the command line now as well.

thommay:
@edolnx can you run this branch through manhattan and give an estimate of what it does to build times (if anything), please.

edolnx:
Build 375 is still running, most of the builds are similar time or faster. The SPARC and Windows builds are slower, but looks like they are falling back to 3 workers and may need an override in the build cookbook for those platforms.

lamont-granquist:
this is definitely the correct thing to do -- but the windows and sparc builds being slower though is hugely bad, we really need to get those defaults fixed...

scotthain:
Looks like one of the windows builders went offline (unrelated issue) during 375, I'm 👍 on this change.

@thommay @lamont-granquist we've got some projects going to resolve the SPARC and Windows speed issues.

PR #4786, opened 2016-04-01
use latest released test-kitchen in chef-acceptance

mwrock:


tyler-ball:
:+1: and it passed in Manhattan :happydance:

acceptance/Gemfile
```diff
@@ -9,5 +9,5 @@ gem "inspec"
 # puts in a box_url for bento when a vagrant box in atlas is specified
 gem "kitchen-vagrant", github: "test-kitchen/kitchen-vagrant"
 gem "windows_chef_zero"
-gem "winrm-fs", "~> 0.4.0"
+gem "winrm-fs"
```
danielsdeleo: I forget: was this just here because we wanted to pin the version? Does kitchen bring this in on its own?

acceptance/Gemfile
```diff
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 gem "chef-acceptance", github: "chef/chef-acceptance"
-gem "test-kitchen", github: "test-kitchen/test-kitchen"
+gem "test-kitchen"
```
danielsdeleo: Do we normally pin this or nah?

danielsdeleo:
I have some questions, but :+1: to merge as-is if you're in a hurry.

chefsalim:
:+1:

acceptance/Gemfile
tyler-ball: Test Kitchen [pins](https://github.com/test-kitchen/test-kitchen/blob/637ec6693a618272d9f6a5291373facaf32979f1/test-kitchen.gemspec#L37) this now

acceptance/Gemfile
tyler-ball: I know we wanted to start using released version of Test Kitchen, but I don't know if we are okay with taking every new version.  Since chef acceptance is still moving pretty fast I'm inclined to leave this as-is (especially since it fixes Manhattan).

acceptance/Gemfile
danielsdeleo: The point I was getting at is that we don't need to declare the dependency at all if TK will pull it in on its own.

PR #4789, opened 2016-04-03
Latest alternate user remote file only rebase

adamedx:


adamedx:
Not needed, changes are in a different branch.

Issue #4793, opened 2016-04-04
Discussion: Verification command output


coderanger:
Filing this as a ticket because I don't see a straightforward implementation. Right now when we run a verification command and it fails all the user sees is something like this:

``````
   Error executing action `create` on resource 'monit_check[nginx]'
    ================================================================================

    Chef::Exceptions::ValidationFailed
    ----------------------------------
    file[/etc/monit/conf.d/nginx.conf] (/var/chef/cache/cookbooks/poise-monit/files/halite_gem/poise_monit/resources/monit_config.rb line 90) had an error: Chef::Exceptions::ValidationFailed: Proposed content for /etc/monit/conf.d/nginx.conf failed verification #<Chef::Resource::File::Verification:0x00000003dc3ae0>

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/poise/files/halite_gem/poise/helpers/notifying_block.rb:69:in `notifying_block'
    /var/chef/cache/cookbooks/poise-monit/files/halite_gem/poise_monit/resources/monit_config.rb:72:in `action_create'```
``````

Notably we don't show the actual verification command if there was one (easy to fix, I can handle that) and we don't show the command output. Most tools with a syntax check option (`nginx -t`, `monit -t`) will print some debugging message indicating what the error is, which would be valuable to bubble up the chain. This is difficult though, because we are re-using the resource guard system and that doesn't expose command output as part of its API. A possible work-around is that good cookbooks shouldn't use the command form of `verify` and should instead use a block with `shell_out!` more directly, and raise the exception themselves. That's a bit weird though. What say we all.

coderanger:
Ping @chef/client-core for thought :)

lamont-granquist:
Yeah definitely :+1: on showing the verification command.

The command output would also be very good, although 'sensitive' resources should probably have that suppressed as well.  If its difficult to disentangle that code, i'd be happy with only showing the verification command for now, and having a TODO to add the output.

coderanger:
Yeah, showing the command would at least hint people that it isn't my code doing the wrong thing. I think that's an easy first step :) The output is the dicey bit. We would either need to not use the guard APIs, or extend those APIs to expose command output somehow (probably not possible given how they work internally).

lamont-granquist:
in #8249 phil mentions he'd like to be able to get at the file contents in addition to the output.

tas50:
[...]

PR #4795, opened 2016-04-04
Slight improvements to validation failures. 

coderanger:
Shows the failing command if possible and doesn't show anything for sensitive resources to be safe.

Refs https://github.com/chef/chef/issues/4793 and ping @chef/client-core for review.

### DIFF
--- a/lib/chef/resource/file/verification.rb
+++ b/lib/chef/resource/file/verification.rb
@@ -122,6 +122,16 @@ class Chef
           v = verification_class.new(@parent_resource, @command, @command_opts, &@block)
           v.verify(path, opts)
         end
+
+        def to_s
+          if @block
+            '<Proc>'
+          elsif @command.is_a?(Symbol)
+            "#{@command.inspect} (#{Chef::Resource::File::Verification.lookup(@command).name})"
+          elsif @command.is_a?(String)
+            @command
+          end
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/chef/resource/file/verification.rb
```diff
@@ -122,6 +122,16 @@ def verify_registered_verification(path, opts)
           v = verification_class.new(@parent_resource, @command, @command_opts, &@block)
           v.verify(path, opts)
         end
+
+        def to_s
+          if @block
+            '<Proc>'
```
coderanger: Do we want to try and do a code parse/display like we do for recipe exceptions?

lamont-granquist: should probably at least print out Proc#source_location?   seems to be a bit annoying to get text out of a Proc...

coderanger: I feel like that is redundant with the `Resource#to_text` this will almost always be displayed with.

lamont-granquist: 'k

thommay:
seems good, please `chefstyle` and then :+1:

lamont-granquist:
pingles, just needs a chefstyle -a pass and merged i think?

lamont-granquist:
re-pingles, travis seems to be failing

PR #4798, opened 2016-04-04
Add a NoOp provider

thommay:
This allows one to have a resource which is really only appropriate for
one platform, but available everywhere. Then you simply allow noop to
provide the resource everywhere besides where it's supposed to be.


lamont-granquist:
need respond_to_missing? instead of respond_to? (and i think you're missing the '?')

https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

lib/chef/provider/noop.rb
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  class Provider
+    class Noop < Chef::Provider
+      def load_current_resource; end
```
lamont-granquist: don't really need this boilerplate.

thommay: seems like you do, at least my tests blow up claiming that I've not implemented it if not.

coderanger:
What's the planned use case?

lamont-granquist:
Might help to display the kind of use case where it could be used:

```ruby
class Chef
  class Provider
    class AptUpdate < Chef::Provider
      use_inline_resources

      provides :apt_update do |node
        which("apt-get")
      end

      [...]
    end
  end
end

Chef::Provider::Noop.provides :apt_update
```

The motivation here is that this is very annoying:

```
apt_update "packages" if node[:platform_family] == "debian"
```

And it'd be nicer to not have to include that statement.  Also it may turn out to simply be wrong when someone installs apt-get on RHEL and wants to use it (or god help us all if microsoft ships apt-get on windows or something like that).

So the provider should figure out when it should run and should no-op itself.

Decorating actions with guards everywhere is ugly:

```ruby
   def action_run
      if node[:platform_family] == "debian"
         # really do the thing with an extra level of indentation and logic everywhere
     end
  end
```

So the solution here is to use the provider-resolver to do that since it already has all the machinery to do that.  The missing bit of machinery is some way to assign a no-op provider that consumes the action and does nothing at all.

coderanger:
I feel like `apt_update` is a very specific case where a silent no-op is the right thing to do. Most platform-specific resources should probably explode if you use them in the wrong place (eg. `env`, `windows_service`, `yum_package`).

lamont-granquist:
Yep, agreed.  You have to explicitly wire this up with that last statement:

```
Chef::Provider::Noop.provides :apt_update
```

coderanger:
Sure, no argument that it would be a bad implementation. Just that if we're only using it for the one thing, this might be overkill :) Can we come up with more? I don't disagree there might be some more that would make sense for this layout, I'm just too brain-fried to think of any.

lamont-granquist:
thom is working on apt_repository and i think there's an argument to be had there.  my first reflex was not to do it, but i'm very on the fence.  by analogy with apt_update i can see not caring about it and not wanting to decorate my recipes with checks for platform_family.  it also seems like a useful thing to let cookbooks experiment with.  if someone wants /their/ resources that they write to no-op on other platforms, i don't see why there can't be an easy way to write that...

coderanger:
I don't see this as saving enough work vs. the one line `return unless node.platform?(...)` to put it in that experimental category I guess. But I'm not strongly against it :)

thommay:
I sort of like the idea of writing:
```
apt_repository "foo" do
…
end
yum_repository "foo" do
…
end
```
and letting chef sort it out. But that's neither here nor there - I think the ergonomics of using the provider system are much cleaner than having to guard around each action.

coderanger:
My biggest worry is that the only reason the `apt_update` example works is because of the ranking system withing the provider mapping, which isn't hugely intuitive for complex setups. Like anything with a block ranks higher than without a block, etc. Have we ever documented that ranking? I don't think it's a huge deal, any mixups should be really easily caught in tests because your custom resources won't actually do anything, but the provider mapping system isn't as obvious as the direct guard logic.

lamont-granquist:
I know that the precedence levels with the NodeMap were deliberate and are tested.  That got rid of a lot of the provider_priority_map nonsense.  I don't know if its been documented, but it was designed that way, and there should be tests around all that behavior, and I think we leverage that internally in some of the core service provider ugliness.

lib/chef/provider/noop.rb
lamont-granquist: i must be having a senior moment, yes that raise has been in there for a very long time...

coderanger:
Yeah, I don't think it's "internal"-y, just more obvious to us as people that deal with it often than to "normal authors". But this doesn't preclude the other approach and if we don't see much use by Chef 13 we could zap it then :)

lamont-granquist:
:+1:

PR #4802, opened 2016-04-05
Explicitly pin json gem

scotthain:
Based on the issue describe here: https://github.com/chef/chef-dk/pull/775


tyler-ball:
\cc @chefsalim @randomcamel @mwrock

omnibus/Gemfile
```diff
@@ -7,6 +7,12 @@ gem "omnibus-software", git: "https://github.com/chef/omnibus-software.git"
 # bundler very unhappy. Remove this when upstream has merged zed-0xff/pedump#6 .
 gem "pedump", git: "https://github.com/ksubrama/pedump.git", branch: "patch-1"
 
+# `json_pure` has a bug in it that is failing ChefDK builds.  We include the
```
tyler-ball: `failing Chef builds`

mwrock:
:+1:

tyler-ball:
:+1:

omnibus/Gemfile
scotthain: COPY PASTA

lamont-granquist:
foo

so this is really not the direction we should be taking...    i'm :+1: but ultimately it looks like there's a fairly simple bug in json_pure that i'm not sure that we've even bothered to cut an issue about?  and this is why all the JSON APIs and monkeypatching is bad, and why *cough* ffi-yajl *cough* exists...

tyler-ball:
@lamont-granquist I like the idea of updating omnibus to use ffi_yajl instead of `.to_json` - I'll see if we can get a card added to our backlog

PR #4809, opened 2016-04-07
MVP to fix gem metadata command

### LINKED ISSUE #4710: chef_gem and gem metadata don't play well
## Description

If one has a gem section in metadata and a chef_gem resource, the resource implodes. `chef_gem` needs to bundle bust.

## Chef Version

12.8.x

## Client Output

https://gist.github.com/thommay/1151fc72d256edd57c5e

lamont-granquist:
The gem command in the cookbook metadata uses the bundler API but needs to not leave the chef-client in a 'bundled' state.

Trying to do this in-process always results in a bundled state.  The result is that /other/ cookbooks that access other gems such as a cookbook with:

```ruby
chef_gem "mysql"
require "mysql"
```

Will fail.  Since we can't convert every cookbook from using chef_gem to get metadata (for backcompat reasons, practicality reasons, and because of native gems and build-essentials deps) we need a workaround.

Best workaround I can come up with is to simply fork off a process.  Since windows does not support fork, the easiest way to implement this is to shell_out to `bundle install`.  When the bundle install returns, we are guaranteed that the spawned process hasn't twiddled at all with our own internal state.

This, however, breaks the ability to get `requires` lines in the gem metadata to work since those would only execute in the subprocess and are flushed with the rest of the state.

closes #4710

### DIFF
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -15,12 +15,13 @@
 # limitations under the License.
 #
 
-require "bundler"
-require "bundler/inline"
+require "tmpdir"
+require "chef/mixin/shell_out"
 
 class Chef
   class Cookbook
     class GemInstaller
+      include Chef::Mixin::ShellOut
 
       # @return [Chef::EventDispatch::Dispatcher] the client event dispatcher
       attr_accessor :events
@@ -45,10 +46,16 @@ class Chef
 
         unless cookbook_gems.empty?
           begin
-            inline_gemfile do
-              source Chef::Config[:rubygems_url]
-              cookbook_gems.each do |args|
-                gem(*args)
+            Dir.mktmpdir("chef-gem-bundle") do |dir|
+              File.open("#{dir}/Gemfile", "w+") do |tf|
+                tf.puts "source '#{Chef::Config[:rubygems_url]}'"
+                cookbook_gems.each do |args|
+                  tf.puts "gem #{args.map { |i| "'#{i}'" }.join(' ')}"
+                end
+                tf.close
+                Dir.chdir(dir) do
+                  so = shell_out!("bundle install")
+                end
               end
             end
           rescue Exception => e
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -53,9 +53,7 @@ class Chef
                   tf.puts "gem #{args.map { |i| "'#{i}'" }.join(' ')}"
                 end
                 tf.close
-                Dir.chdir(dir) do
-                  so = shell_out!("bundle install")
-                end
+                so = shell_out!("bundle install", pwd: dir)
               end
             end
           rescue Exception => e
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -66,60 +66,6 @@ class Chef
 
         events.cookbook_gem_finished
       end
-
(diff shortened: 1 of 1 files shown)


lamont-granquist:
this'll break 'requires' in the gem metadata itself.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -45,10 +46,16 @@ def install
...
-                gem(*args)
+            Dir.mktmpdir("chef-gem-bundle") do |dir|
+              File.open("#{dir}/Gemfile", "w+") do |tf|
+                tf.puts "source '#{Chef::Config[:rubygems_url]}'"
+                cookbook_gems.each do |args|
+                  tf.puts "gem #{args.map { |i| "'#{i}'" }.join(' ')}"
+                end
+                tf.close
+                Dir.chdir(dir) do
+                  so = shell_out!("bundle install")
```
coderanger: Do we want to make sure to specify Chef's bundler binary somehow?

lib/chef/cookbook/gem_installer.rb
```diff
@@ -45,10 +46,16 @@ def install
...
           begin
-            inline_gemfile do
-              source Chef::Config[:rubygems_url]
-              cookbook_gems.each do |args|
-                gem(*args)
+            Dir.mktmpdir("chef-gem-bundle") do |dir|
+              File.open("#{dir}/Gemfile", "w+") do |tf|
+                tf.puts "source '#{Chef::Config[:rubygems_url]}'"
+                cookbook_gems.each do |args|
+                  tf.puts "gem #{args.map { |i| "'#{i}'" }.join(' ')}"
```
coderanger: Are we planning support stuff like `git` sources? I think that would have worked with the prior code but not this.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -45,10 +46,16 @@ def install
...
+                end
+                tf.close
+                Dir.chdir(dir) do
```
coderanger: For the hell of it, you can use `cwd: dir` on the shell out :)

lib/chef/cookbook/gem_installer.rb
lamont-granquist: Yeah, I'm just drawing a blank today on how to render those cases...

lib/chef/cookbook/gem_installer.rb
lamont-granquist: I think we probably do?  Maybe via the omnibus-detection stuff inside the rubygems provider that we use for chef_gem and then run bundler from that path...

lib/chef/cookbook/gem_installer.rb
coderanger: Change it to `"gem(*#{args.inspect})"`?

lib/chef/cookbook/gem_installer.rb
coderanger: That sounds right. We can rely on the appbundle stub on `chef-client` itself to bust rvm envvars I think. I'm sure there is some way this can still be broken but ¯\_(ツ)_/¯

coderanger:
:+1: This is unfortunate but I think this is better than what we have now given the constraints.

coderanger:
One thing I might add is some debugging output stuffs. Like echo the generated gemfile to debug log :)

lamont-granquist:
Yeah this gets us from "totally broken feature" to "usable with issues" and is probably the best I can come up with before the 12.9.x release next week.

Gemfile to the debug log is probably a good idea..

Issue #4362, opened 2016-01-07
Deploy code that's not in SCM


ameir:
The `deploy` resource is so close to what I need, except that it appears to require git or svn.  In my environment, I use Jenkins to pull code from git, compile and/or do other magic, and push a tarball to S3.  That tarball is the production-ready code, without the .git directory.

Currently, I'm using `aws_s3_file` to pull the tarball, and an `execute` block to untar it and manage directories, etc.  It works, but it's not as clean as I'd like it to be (hence me wanting to use this resource).

Are there workarounds to working with a local file instead of an SCM repo?

I'd be happy to work on supporting this in whatever spare time I can scrape up, and use the file's hash instead of the SCM commit/revision for the directory name.

I'm also aware of https://github.com/RiotGamesCookbooks/artifact-cookbook, but it does not appear to be maintained.

zhelyan:
+1

ameir:
Closing since updates to `deploy` are not being considered at this time, per #4810.

PR #4810, opened 2016-04-07
Add `deploy_directory` resource.

ameir:
This is a simple solution to #4362.  It works just the same as the current `deploy` resource, but skips the SCM bits.  Instead of copying to the `cached-copy` directory, the code is copied directly from `repo` to the current release path.  I'd be happy to provide documentation once this has been reviewed and 👍 .


coderanger:
What is the use case for this? I'm not super clear how you would actually use it.

ameir:
Sure, I'd be happy to explain.  I'm at a fairly large company that runs thousands of hosts.  The deployment pattern varies a tad across teams, but much of it looks something like:

```
aws_s3_file '/tmp/VERSION' do
  bucket 'my-project-artifacts'
  remote_path "my-project/#{ENVIRONMENT}/api/VERSION"
end

build_version = -> { ::File.read('/tmp/VERSION') }
directory 'create directory for build' do
  user 'appuser'
  group 'appuser'
  path lazy { "#{basedir}/builds/#{build_version.call}" }
  recursive true
end

execute 'fix permissions' do
  command "chown -R appuser:appuser #{basedir}"
  cwd basedir
  only_if 'namei -mo builds | grep root'
end

aws_s3_file '/tmp/build.tgz' do
  owner 'appuser'
  bucket 'my-project-artifacts'
  remote_path lazy { "my-project/#{ENVIRONMENT}/api/#{build_version.call}.tgz" }
  notifies :run, 'execute[extract build]', :immediately
end

execute 'extract build' do
  user 'appuser'
  group 'appuser'
  command lazy { "tar xfz /tmp/build.tgz -C #{basedir}/builds/#{build_version.call}/" }
  action :nothing
end

link 'update symlink' do
  user 'appuser'
  group 'appuser'
  to lazy { "#{basedir}/builds/#{build_version.call}" }
  target_file "#{basedir}/my-project"
  notifies :reload, 'service[php-fpm]'
end
```

We basically download an artifact tarball from S3, extract it, and update a symlink.  We don't have support for deleting old builds, and have a lot of this type of code across many recipes, which isn't ideal.

With this update, we can have the `deploy` resource manage everything after the code is extracted.

We don't use the `deploy` resource as-is because it's not scalable; we can't have hundreds or thousands of hosts doing a `git clone` concurrently, and then running the builds locally.  We instead have Jenkins do that once per build, push an artifact to S3, and have the hosts pull that down.

I hope that clarifies things a bit; let me know if you need more details.

coderanger:
I'm a little confused why you want/need the `deploy` resource logic in here at all. Just do what you are doing now without the symlinks. The symlink structure of Capistrano (which `deploy` copied) makes sense for interacting with humans, but not really for fully automated deployments. You could probably simplify this down to one `aws_s3_file` resource notifying a `poise_archive` resource and leave it at that. I would also really recommend looking at switch to fpm and debs/RPMs instead of tarballs since that handles permissions, cleanup, repo formatting, etc.

ameir:
The Capistrano influence is very apparent, so I do `symlink_before_migrate({})` and a few other things to turn off some of those "freebies" that conflict with my flow.

The symlinks are there for two reasons: atomicity and the ability to rollback.  Granted, rolling back doesn't work with `deploy_directory`, but it can be supported fairly easily down the road.

`poise_archive` seems similar to `ark`, which I do use fairly often.  If these instances were fully-immutable, your suggestion would work fine, but they are long-lived and updated often.

lamont-granquist:
I understand the use case, but the deploy resource is a hot mess of old code with architectural bugs that are likely to never get fixed short of a complete rewrite.  I don't really like the idea of building on top of a bad foundation.

lamont-granquist:
Yeah I'm leaning strongly towards anything like this that builds on the existing deploy resource should go into a cookbook, I don't think we would want to accept the responsibility of maintaining more stuff on top of the existing deploy resource in core.

ameir:
Are there any plans to give the `deploy` resource any love in the near future?  I figured an update like this would be great for those folks who don't deploy from SCM and want to use an easy, stable core resource, but I do understand your concerns.

I can split this up into its own cookbook; I haven't looked into extending core resources before, but is it as straightforward as I'd expect?

coderanger:
@ameir I gave it a bunch of love by providing examples on how to not ever use it again :) The atomicity argument is real, but most systems read the code into memory so the real atomicity of upgrades is on when you restart the service. This doesn't apply to some tools though, notably PHP. Deploying from static archives is cool and should be well supported by the ecosystem, but I don't think we need to keep doing the symlink dance in the general case. If you need it specifically, it's super easy to implement unpacking each rev of the artifact to its own folder and use a single `link` resource to manage things :) That avoids the whole mess of the `deploy` resource and keeps things mostly in convergent territory.

lamont-granquist:
Noah has his complete reimagining of what the deploy resource does.  I've been noodling around with a half-written prototype.  But yeah in its simplest case, what @coderanger just ninja posted there while i was typing this -- unpack to versioned directories and use a link resource and avoid all the buggy framework that gets dragged along with the deploy resource.  Half of what the deploy resource does is buggy code that attempts to implement capestrano and if you're not deploying a rails app its not doing you any good.  The other half of the code is interfacing with the git scm provider in a horribly buggy fashion.   If you don't need to deploy a rails app and aren't deploying from git there's just not a lot left in the deploy resource that you're buying other than a dumptruck load of bugs around a symlink resource...

ameir:
Thanks, guys.  Looks like I'll just write a little wrapper around what I need to get done, and avoid all the overhead inherited from `deploy`.  

Thinking out loud: it sounds like `deploy` itself should be generic, and the Capistrano clone should be called `rails_deploy` or `cap_deploy` or something of the sort.  That's a BC break, of course, but it definitely seems more logical :) .

coderanger:
Also just for the record: Amazon does hosted Git repos as a service so you could use those. It will be waaaay more efficient than tarballs as far as transfer times :) I still prefer artifact-based deploys in most cases for the simplicity of it, but it's another option.

coderanger:
@ameir A more generic vision of an app deploy resource is exactly what the `application` resource does :)

ameir:
I looked at https://github.com/poise/application before, but it looked like the source _had_ to be cloned from git.  I am familiar with AWS CodeCommit, but the build process _after_ cloning the repo is what takes a lot of time (grunt/gulp tasks, npm installs, composer, etc.), which is why I went the artifact route.  I'll gladly take another look if I'm missing something.  Great cookbook marketing, btw ;) .

ameir:
Unless you mean that I do all of the building and then push that to a separate repo that I deploy from, which is an interesting idea.

coderanger:
@ameir The `git` you see in all those examples is a normal resource. You could replace that with your existing s3 file + unpack logic to the same end and still use all the same application plugins. This might not be super useful depending on how all-encompassing your artifact is.

ameir:
Oh, that's pretty neat.  I'll start giving that a whirl now; thanks!

PR #4812, opened 2016-04-07
Removing non-existent members from group should not fail

chefsalim:
This resolves the following issue : https://chefio.atlassian.net/browse/TRI-96

Currently, if removing users from a group, where the user is malformed or does not exist, we error out and fail the recipe. The new behavior is to not fail in this case - we now just log a warning and keep going.

@chef/client-core @tyler-ball @jkeiser

### DIFF
--- a/lib/chef/provider/group/windows.rb
+++ b/lib/chef/provider/group/windows.rb
@@ -100,7 +100,7 @@ class Chef
           begin
             Chef::ReservedNames::Win32::Security.lookup_account_name(locally_qualified_name(account_name))[1].to_s
           rescue Chef::Exceptions::Win32APIError
-            Chef::Log.warn("SID for '#{locally_qualified_name}' could not be found")
+            Chef::Log.warn("SID for '#{locally_qualified_name(account_name)}' could not be found")
             ""
           end
         end
--- a/lib/chef/provider/group/windows.rb
+++ b/lib/chef/provider/group/windows.rb
@@ -61,7 +61,7 @@ class Chef
           if @new_resource.append
             members_to_be_added = [ ]
             @new_resource.members.each do |member|
-              members_to_be_added << member if ! has_current_group_member?(member)
+              members_to_be_added << member if ! has_current_group_member?(member) && validate_member!(member)
             end
 
             # local_add_members will raise ERROR_MEMBER_IN_ALIAS if a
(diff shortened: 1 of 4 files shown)


coderanger:
@chefsalim This bug seems pretty self evident but remember your JIRA is not public so make sure to copy over any important details :)

coderanger:
Also do the various unix-y providers have the same behavior?

chefsalim:
@coderanger here are the relevant details from the ticket.  The issue is specific to Windows due to the conversion from username to SID.

Customer windows system joined to an AD domain has trouble removing users that no longer exist from groups. An exception is raised, the following, which occurs when the windows group provider tries to do `Chef::ReservedNames::Win32::Security.lookup_account_name()` on a non-existent username.

ecipe: c2c::c2c_account_remove
ohai[reload_c2c_get_existing_accounts] action nothing (skipped due to action :nothing)
group[Administrators] action modify
Error executing action `modify` on resource 'group[Administrators]'
Chef::Exceptions::Win32APIError
-------------------------------
No mapping between account names and security IDs was done.
System Error Code: 1332
System Error Message: No mapping between account names and security IDs was done.
Resource Declaration:

lib/chef/provider/group/windows.rb
```diff
@@ -80,18 +80,31 @@ def manage_group
...
 
-        def local_group_name_to_sid(group_name)
-          locally_qualified_name = group_name.include?("\\") ? group_name : "#{ENV['COMPUTERNAME']}\\#{group_name}"
-          Chef::ReservedNames::Win32::Security.lookup_account_name(locally_qualified_name)[1].to_s
+        def locally_qualified_name(account_name)
+          account_name.include?("\\") ? account_name : "#{ENV['COMPUTERNAME']}\\#{account_name}"
         end
+
+        def validate_member!(member)
+          Chef::ReservedNames::Win32::Security.lookup_account_name(locally_qualified_name(member))[1].to_s
```
tyler-ball: Does this raise an error if the member cannot be found?

tyler-ball: :doh: I see it does, looking at the method below

spec/functional/resource/group_spec.rb
```diff
@@ -267,14 +267,14 @@ def create_group
     end
 
     describe "when removing members" do
-      it "raises an error for a non well-formed domain name" do
+      it "does not raise an error for a non well-formed domain name" do
```
tyler-ball: Is this behavior change desired?  I understand about not raising an error for a non-existant member, but why not raise an error if the member name has a space in it?

jkeiser:
:+1:

spec/functional/resource/group_spec.rb
chefsalim: I guess we could special case malformed vs. non-existent, however at the API level it doesn't really care and has the same result (raise error).  It seemed simplest / safest to just warn and continue when checking membership for removal.

adamedx:
:+1:, and I'm guessing I introduced the failing behavior in the non-well-formed case when adding the conversion to SID.

PR #4817, opened 2016-04-08
12.9 release prep

tyler-ball:
\cc @chef/maintainers 

If you need to make some changes to DOC_CHANGES or RELEASE_NOTES, now is the time!


tyler-ball:
Passed Manhattan

tas50:
@tyler-ball I tagged a bunch of the PRs that were merged.  If you regenerate the changelog it'll be a lot more useful to users now.

chefsalim:
👍

jkeiser:
:+1: with regenerated changelog :)

lamont-granquist:
@tyler-ball can we remove issues entirely from the changelog?  Closed issues are not *CHANGES*, and all of the closed issues that are accurate are also duplicates with the PRs that actually fixes the issues and which are the *CHANGES*.

tas50:
I've been marking a lot of these issues to avoid getting them into the changelog, but the whole section seems to provide little customer value and probably introduces some confusion as "not a bug" sort of closures look like changes to someone reading the changelog.

lamont-granquist:
Yeah, and we're also dependent upon your particular "OCD"ishness in getting an accurate changelog, which means we have a bus factor of 1 in generating accurate changelogs this way.  In general if people do not do the work you do at all then the changelog /will/ become littered with notabug/wontfix closed issues that look like they were fixed.  We'll also see bugs closed that were "fixed in previous releases please update" which will be completely misleading, etc.

PR #4818, opened 2016-04-08
add notifying_block and subcontext_block to chef

lamont-granquist:


lamont-granquist:
@coderanger thoughts?

lib/chef/mixin/notifying_block.rb
```diff
@@ -0,0 +1,54 @@
...
+class Chef
+  module Mixin
+    module NotifyingBlock
+
+      def notifying_block(&block)
+        begin
+          subcontext = subcontext_block(&block)
+          Chef::Runner.new(subcontext).converge
+        ensure
+          new_resource.updated_by_last_action(
```
coderanger: So a thing I've never fixed in Poise but we probably should here is that using `notifying_block` twice where the first block updates and the second doesn't ends up with `updated_by_last_action` being false. We should probably make this more like `updated_by_last_action(true) if <the condition>` so we don't overwrite an existing true with a false.

lib/chef/mixin/notifying_block.rb
```diff
@@ -0,0 +1,54 @@
...
+          @run_context = sub_run_context
+          instance_eval(&block)
+        ensure
+          @run_context = outer_run_context
+        end
+
+        sub_run_context
+      end
+
+      def global_resource_collection
```
coderanger: This is only needed if we want to port over the context-smashing `include_recipe` helper code, and it can probably just live inside that since it is so simple now. I would leave it out.

coderanger:
:+1: :+1: :+1:

lamont-granquist:
So here's the kind of thing this lets us do, among other things:

```ruby
cache1 = "#{Chef::Config[:file_cache_path]}/content1"
cache2 = "#{Chef::Config[:file_cache_path]}/content2"

notifying_block do
  remote_file cache1 do
    source "http://source.url/content1"
    notifies :create, "file[/my/combined/file]", :delayed
  end

  remote_file cache2 do
    source "http://source.url/content2"
    notifies :create, "file[/my/combined/file]", :delayed
  end

  file "/my/combined/file" do
    content IO.read(cache1) + IO.read(cache2)
  end
end
```

This delays the notification only to the end of the notifying_block and not to the end of the chef-client run.  Other recipes can rely on the file being dropped into place.  If the notifies had used `:immediately` then it would trigger the file resource twice--the first time with partial information.

Obviously you can do the same thing with resources, but you don't necessarily want to be forced to create a resource every time you want to group a couple of resources for notification purposes.

This pattern gets worse when it comes to dropping config files and bouncing services and sanely handling delayed notifications but leaving the service correctly running when the recipe exits:

https://tickets.opscode.com/browse/COOK-3358

I believe there's an RFC somewhere else for doing this kind of thing and there's often requested :delayed_until_end_of_recipe notification styles that this should cover.

These methods are not resources.  Its possible we might look at introducing a 'recipe' resource or pseudo-resource that would accept notifications to it.  Considering that out of scope.

coderanger:
I tweaked your example slightly, you still need the lazy eval bits because inner compile/converge cycle.

lamont-granquist:
ah yeah, totes...

PR #4820, opened 2016-04-11
Lock dependencies of chef through a `Gemfile.lock`

jkeiser:
This all works the same as the chef-dk version. The README explains how it all works and fits together. The intent of this change is to get us out of the daily grind of "some dependency broke us" (which happens a lot) and make changes more deliberately. To avoid the obvious issue of not updating enough, we will periodically (through a bot) run `rake dependencies`, send the PR through, and auto-merge to `master` if it is green.


coderanger:
Very :-1:. Chef is still a gem at heart to at least some degree. We can't leave gem installs broken for unknown periods of time. The build breaking is exactly what it it supposed to do.

coderanger:
If you want to snapshot compiled dependencies into the gemspec, that's at least not going to break everyone else consuming Chef as a gem from master (yes, I am aware I might be literally the only other person that does this O:-) ).

jkeiser:
@coderanger the plan is to periodically (every day) update deps, check if they are red, and fix them over the next few days if they are not. Beyond that, you also have `master`.

This is not going to break people depending on released Chef gems. We have plenty of protections against that: the build will fail if gems that chef depend on are not solved to the latest version each time we make a build.

coderanger:
Sure, I'm talking about people that continuously consume Chef from master as a gem. all my builds include a matrix slot like this to ensure they work with Chef master. That being broken for a day at a time is super uncool. If we need to fix something, fine, but this is the wrong solution as it only fixes it for Chef, not the broader ecosystem. My vote would be to write something that converts the solved dependency graph into some code that can be inserted (or dynamically evaluated) in `chef.gemspec` so we still get pinning but in a way that cascades to gem users (which, again, might just be me but we don't know).

jkeiser:
@coderanger hmm. I don't think pinning gemspecs will really solve you, due to `git` and `path` dependencies in many Gemfiles (including ours). I think what you really want is for your `Gemfile` to be pinned to the `Gemfile.lock` of `chef`, so that you use the same tested versions. I bet I could come up with a not completely unreasonable snippet for that. (It should be a bundler feature, and I now have a backlog of bundler features I'd like to create :)).

coderanger:
AFAIK bundler doesn't allow using an existing lockfile in that manner. I can't use Chef's lockfile as-is given that I need my own gems too.

jkeiser:
@coderanger it doesn't, but I implemented it for the chef-dk as part of this, because (a) I wanted to guaranteed that `Gemfile.windows` pinned to the same versions as `Gemfile` no matter what other deps came in, and (b) I wanted `chef verify`, which has to `bundle install` some things like `test-kitchen` and `berkshelf` to bring in test dependencies, to make *damn* sure that all runtime deps were pinned to the version in the chef-dk so it didn't "upgrade its way out of a fix" and invalidate the test. So I know how, at least, and I'll see if I can't get you something practical :)

coderanger:
Also I intentionally don't pin my dependencies since I want exactly the breakages you are trying to avoid here. I really think this at best a band-aid over either using too many external deps (or lack of vendoring for external deps) or insufficient QA for "internal" dependencies. We should fix those instead.

jkeiser:
I've thought about vendoring our deps. I don't think it would help you though, would it? If we vendored the deps for `chef`, I don't think bundler will automatically use them when you do `gem 'chef', github: 'chef/chef'`, will it?

Whether you pin in the gemspec or pin in the Gemfile.lock seems like the same thing. Either you pin and take new dependencies in safely and quickly, or you let them run wild and take unplannable interrupts from your week that badly stall in-progress feature development.

coderanger:
It would if they were vendored inside the gem. For example, how Django pulls in dependencies https://github.com/django/django/blob/master/django/utils/six.py

Easy to over-do, but if there are specific libraries that we don't control and are causing breakage a lot, it's an option. Do we have any data on what kinds of things are causing build "downtime"?

danielsdeleo:
Just want to fill in some details about how the dependency bumping stuff is expected to work. You can see the first 1/2 of the code here: https://github.com/chef/lita-versioner/pull/11 The tl;dr if you don't want to go through that is that it's designed so it can run very frequently but won't spam the build system if the dependency update results in a broken configuration. So for now I'm planning to run it every 15 minutes. What that code doesn't do yet is actually check the build results, but the plan is to add that next and then it'll be an expedite thing for someone to fix it if the updated dependencies are breaking the build.

coderanger:
If y'all are willing to commit to `gem 'chef', github: 'chef/chef'` not breaking for a long enough period of time (i.e. more than a few hours) then I am happy with any solution. If at any point the prior assertion is broken, we are doing something Very Wrong™. How we get there, I don't much care, but if this means gem installs from master won't be under CI then I don't want to be the one picking up the pieces.

danielsdeleo:
I can't promise that any given bug will get fixed in a fixed period of time, of course. It will get assigned to a specific person to look at in the same or maybe a better way than it is now (more on that below)

The goal here isn't about reducing the level of support for the gem install scenario, we just want a more orderly process of pulling in updates so that, for example, we're not blocked on shipping an OpenSSL patch because some other gem broke something else. It's less of problem for Chef, but for ChefDK we found that we had problems precisely because some components were de facto locked (because their versions were baked into omnibus) and we would do a big bang update before a release, only to find that some combination of the new things didn't work. But users don't like it when we ship stable releases with things (like kitchen or berks) built from master, so pinning omnibus to master isn't a viable solution. So the overall goal is to pull in new stuff as quickly as possible, but have every build on the `current` channel also meet the same requirements we have for `stable`. We have these kinds of problems a lot less frequently with Chef, but would like both projects to have a similar build process.

Anyway, we currently have a business hours pager duty rotation for handling build breaks, and failures from the dependency updater will feed into that, so someone will be interrupted if the current combination of latest available gems breaks something. And the builds will all be going through the build cluster, which does scenarios like running the top N cookbooks via test kitchen, which means we'll catch things like WinRM bugs and so on that we wouldn't catch in travis/appveyor.

The outcome I'm expecting is that normally the response to having master plus the latest combination of gems being broken in some way will be pretty much the same as now. There may be situations where there is some higher priority emergency and we choose to do something else (e.g., address OpenSSL or the occasional Ruby security issue) before fixing dependency breaks, but that will be the exception.

README.md
```diff
@@ -116,6 +116,151 @@ tests in spec/functional/ and spec/unit/ respectively. These will be run on Ubun
...
+
+To build chef as a standalone package (with ruby and all dependent libraries included in a .deb, .rpm, .pkg or .msi), we use the omnibus system. Go to the [omnibus README](omnibus/README.md) to find out how to build!
+
+## Updating Dependencies
+
+If you want to change our constraints (change which packages and versions we accept in the chef), there are several places to do so:
+
+* To add or remove a gem from chef, or update a gem version, edit [Gemfile](Gemfile).
+* To change the version of binary packages, edit [version_policy.rb](version_policy.rb).
+* To add new packages to chef, edit [omnibus/config/projects/chefdk.rb](omnibus/config/projects/chefdk.rb).
```
randomcamel: Is `chefdk` correct here?

omnibus/config/software/chef.rb
```diff
@@ -0,0 +1,92 @@
...
+dependency "rubygems"
+dependency "bundler"
+
+# Install all the native gems separately
+# Worst offenders first to take best advantage of cache:
+dependency "chef-gem-ffi-yajl"
+# dependency "chef-gem-json" (only a development dep)
+dependency "chef-gem-nokogiri"
+dependency "chef-gem-libyajl2"
+# dependency "chef-gem-ffi" (only a development dep)
```
randomcamel: Should these commented-out lines be deleted?

omnibus/config/software/chef.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  # Install all the things. Arguments are specified in .bundle/config (see create_bundle_config)
+  block { log.info(log_key) { "" } }
+  bundle "install --verbose", env: chef_build_env
+
+  # For whatever reason, nokogiri software def deletes this (rather small) directory
+  block { log.info(log_key) { "" } }
+  block "Remove mini_portile test dir" do
+    mini_portile = shellout!("#{bundle_bin} show mini_portile").stdout.chomp
+    remove_directory File.join(mini_portile, "test")
```
randomcamel: Not sure of the logic here: the nokogiri def deletes a directory, but then we call `remove_directory`?

coderanger:
If this won't affect the response to dependency breakages, I change my vote to -0 :) I still think this is silly and is a patch-y solution to an increasingly unwieldy build/QA system but given that I am not volunteering to fix said build system I won't do more than point that out. If we haven't already, it would probably be good to have some people sign up for [Gemnasium updates](https://gemnasium.com/chef/chef) (or similar system) to track when dependencies change, though that won't help for stuff being pulled in from git but then again we should probably stop doing that as much :shipit:

README.md
jkeiser: Nope! Thanks :)

omnibus/config/software/chef.rb
jkeiser: Sure. Don't mind at all.

omnibus/config/software/chef.rb
jkeiser: @randomcamel we aren't calling the `nokogiri` software def anymore at all, so we have to dup its logic.

randomcamel:
👍

acceptance/fips/test/integration/fips/serverspec/Gemfile.lock
```diff
@@ -0,0 +1,13 @@
+GEM
```
tyler-ball: What is this doing here?

jkeiser: @tyler-ball I just went through every Gemfile I could find and locked it, on the principle that tests need to be locked just as much as the code.

tyler-ball:
👍

PR #4822, opened 2016-04-12
Move all ChefFS exceptions into a single file

thommay:
This leaves back compat requires to ensure nothing breaks.
cc @chef/client-core


lamont-granquist:
👍

danielsdeleo:
Should we deprecate the old files so we can delete them in 13.0?

danielsdeleo:
:+1: either way though

jkeiser:
:+1:, I like this!

thommay:
@danielsdeleo now with bonus deprecations

PR #4824, opened 2016-04-13
Add a warning about using --interval on Windows.

coderanger:
I think I got the if logic right, but a spot check would be good. Basically `--interval` on Windows works but is a really bad idea for all the reasons we hard-blocked non-forking interval runs on other platforms. We probably shouldn't actually break it on Windows, but from talking to Matt it sounds like nobody should be doing this so let's give them warning now and remove it in Chef 13 or something.

/cc @chef/client-windows for review.

### DIFF
--- a/lib/chef/application/solo.rb
+++ b/lib/chef/application/solo.rb
@@ -333,9 +333,9 @@ EOH
   end
 
   def unforked_interval_error_message
-    "Unforked chef-client interval runs are disabled in Chef 12." +
+    "Unforked chef-solo interval runs are disabled in Chef 12." +
       "\nConfiguration settings:" +
       "#{"\n  interval  = #{Chef::Config[:interval]} seconds" if Chef::Config[:interval]}" +
-      "\nEnable chef-client interval runs by setting `:client_fork = true` in your config file or adding `--fork` to your command line options."
+      "\nEnable chef-solo interval runs by setting `:client_fork = true` in your config file or adding `--fork` to your command line options."
   end
 end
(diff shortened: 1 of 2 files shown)


mwrock:
I definitely like having a more perscriptive message. However I think it should raise and not simply warn. Its realy more than discouraged. Its unsupported. Based on the conditionals above this change, it looks like this not only protects against using `:interval` on windows but also `:daemonize`. I don't think we want to give the impression that these are worth pursuing on Windows.

lamont-granquist:
Bizzare, and I appear to have :+1:'d the code that added the windows check as well.  I would just remove it and raise, because unforked interval running isn't going to work right at this point.  If we need to fix the 'forked' side of things so that we shell_out to a one-shot run that does the work we should do that.

lib/chef/application/solo.rb
```diff
@@ -326,9 +333,9 @@ def fetch_recipe_tarball(url, path)
   end
 
   def unforked_interval_error_message
-    "Unforked chef-client interval runs are disabled in Chef 12." +
+    "Unforked chef-solo interval runs are disabled in Chef 12." +
```
thommay: this seems to have made travis unhappy.

tas50: Update the spec
https://github.com/chef/chef/blob/c1a389c2a8452e9b796aa1d34c4d9e51f4af30c7/spec/unit/application/solo_spec.rb#L70

lamont-granquist:
So, to be really clear, IMO we broke interval runs in the same process completely in 12.0 -- windows people just haven't yet been clearly informed of that, and anyone running interval runs in the same process is getting some truly _fascinating_ monkeypatching as they upgrade their cookbooks and don't bounce their chef-client 'service' on their nodes.  I would not consider a 'raise' here to be a breaking change because it is breaking something that is already hella broken.

lamont-granquist:
also just needs to be cleaned up and merged

PR #4826, opened 2016-04-13
Refactor ChefFS directories to be directories

thommay:
Internally, this is FLOW-252:

Currently path types in ChefFS are handled inconsistently. 
While some paths have their own type, other paths are implemented by injecting a variable into a generic class, which makes it hard to troubleshoot behavior with just one type of object. For example, if one API call is behaving unexpectedly, additional debug code will likely apply to many other API calls, which makes printf debug output overwhelming and pry debugging annoying. 
ChefFS' design can be improved to make the code clearer to read and modify, which we expect will make future improvements easier.

Some Caveats:
- i've not added tests for the subclasses yet. I'm not sure how much value tests there will be right now.
- The cookbook directories are special snowflakes currently.

### DIFF
--- a/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_acls_dir.rb
+++ b/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_acls_dir.rb
@@ -17,6 +17,7 @@
 #
 
 require "chef/chef_fs/file_system/repository/chef_repository_file_system_entry"
+require "chef/chef_fs/file_system/repository/directory"
 require "chef/chef_fs/file_system/chef_server/acls_dir"
 require "chef/chef_fs/data_handler/acl_data_handler"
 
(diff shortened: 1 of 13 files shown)


thommay:
@danielsdeleo review appreciated.

lamont-granquist:
makes things look more symmetrical so 👍  i guess?

lib/chef/chef_fs/file_system/repository/chef_repository_file_system_acls_dir.rb
```diff
@@ -17,21 +17,25 @@
 #
 
 require "chef/chef_fs/file_system/repository/chef_repository_file_system_entry"
```
danielsdeleo: Probably can remove all these requires in the files you changed.

danielsdeleo: Never mind, you need them in the `#make_child_entry` methods.

danielsdeleo:
:+1:

PR #4827, opened 2016-04-13
Updating version after Chef 12.9 release

tyler-ball:
\cc @danielsdeleo @randomcamel @chefsalim @mwrock


omnibus/config/projects/chef.rb
```diff
@@ -45,7 +45,7 @@
 
 # Chef Release version pinning
 override :chef, version: "local_source"
-override :ohai, version: "v8.14.0"
+override :ohai, version: "master"
```
danielsdeleo: do we actually want to ride master here?

danielsdeleo:
:+1: thought we wanted to keep everything, including ohai, on some kind of current or release version, but we don't really have a current for ohai. I'm happy either way, just wondering what the plan is.

omnibus/config/projects/chef.rb
randomcamel: What he said.

randomcamel:
👍

tyler-ball:
Once @jkeiser gets his one-gemfile-to-rule-them-all changes merged in, it will float on the latest released version.  But for now it is still set on master

PR #4830, opened 2016-04-14
Rename and add backcompat requires for ChefFS dirs

thommay:
Policies, ACLs, and Client Keys directories are renamed to match the
other directories, and backwards compatible requires are added.
Ensure that they all properly inherit from Repository::Directory, too


thommay:
@chef/client-core

lib-backcompat/chef/chef_fs/file_system/repository/file_system_root_dir.rb
```diff
@@ -24,6 +24,7 @@ module FileSystem
       module Repository
         class FileSystemRootDir < FileSystemEntry
           def initialize(file_path)
+            Chef.log_deprecation "This class is deprecated."
```
danielsdeleo: Probably should include the class name so if you happen upon this in a log you have something to grep for.

danielsdeleo:
:+1:

lib-backcompat/chef/chef_fs/file_system/repository/file_system_root_dir.rb
thommay: @danielsdeleo heh, right. Done :)

lamont-granquist:
👍

PR #4832, opened 2016-04-14
add nillable apt_repository and nillable properties

lamont-granquist:
in Chef-13 nillable will become 'true' by default and we'll have
to deprecate and remove all nillable properties, but for now
this lets us opt-in, which was can't currently do in Chef-12.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -72,6 +72,9 @@ class Chef
     #     property defaults to the same value as `name`. Equivalent to
     #     `default: lazy { name }`, except that #property_is_set? will
     #     return `true` if the property is set *or* if `name` is set.
+    #   @option options [Boolean] :nillable `true` opt-in to Chef-13 style behavior where
+    #     attempting to set a nil value will really set a nil value instead of issuing
+    #     a warning and operating like a getter
     #   @option options [Object] :default The value this property
     #     will return if the user does not set one. If this is `lazy`, it will
     #     be run in the context of the instance (and able to access other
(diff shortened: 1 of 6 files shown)


lamont-granquist:
@chef/client-core fixes apt_repository resource's distribution property to be nillable, and lets us opt-in to Chef-13 nillable behavior in properties.

i did not go down the route of making explicit nils in validations act like nillable because of the subtleties in handling explicit and implicit nils where making implicit nils nillable by default is something we can't do.  so adding an option was done because it'll make everyone's head less explode-ey and intent becomes abundantly clear...

coderanger:
I thought the opt-in was declaring `nil` as a valid value for the property? Or was that in one of the patches that got reverted somewhere along the line.

lamont-granquist:
yeah but we did all that stuff with the implicit nils being added to the validation, and i think that made the logic very confusing so it got dropped.

it definitely does not currently work that way, and this patch is vastly simpler than trying to implement that logic and getting the implicit-vs-explicit nil validation behavior correct (and then explaining it to everyone).

coderanger:
Fair, no objection as long as we don't end up with two systems :)

lamont-granquist:
Nope, pretty sure in Chef-13 that whole nasty if block in properties just gets deleted and we drop "nilllable: true" off of this property and any others it gets added to, and then have properties spit out a deprecation warning for people who set "nillable: true" that its now the default and is a no-op and will go away.  The end result should ultimately be clean, and I think the road ahead is halfway sane to reason about...

coderanger:
Only other thought is that any non-boolean property that is nilable should also support `false` because it makes life easier sometimes.

lamont-granquist:
@coderanger that look okay?  i made other things nillable (and falseable) as well.

if not, i can move the coercsions privately into some pseudo-delegators on the provider...

coderanger:
:ship: :shipit: :ship:

PR #4834, opened 2016-04-15
add better resource manipulation API

lamont-granquist:
deprecates chef_rewind functionality completely and adds a few
more features

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -227,7 +227,7 @@ class Chef
 
         if create_if_missing
           Chef::Log.deprecation "build_resource with a create_if_missing flag is deprecated, use edit_resource instead"
-          # midly goofy since we call edit_resoruce only to re-call ourselves, but that's why its deprecated...
+          # midly goofy since we call edit_resource only to re-call ourselves, but that's why its deprecated...
           return edit_resource(type, name, created_at, run_context: run_context, &resource_attrs_block)
         end
 
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -1,6 +1,6 @@
 #--
 # Author:: Adam Jacob (<adam@chef.io>)
-# Author:: Christopher Walters (<cw@chef.io>)
+# Author:: Christopher Walters
 # Copyright:: Copyright 2008-2016, 2009-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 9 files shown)


lamont-granquist:
So now for code that accumulates state in a resource in the root run_context and does its work via a delayed notification we can:

```ruby
action :run do
  with_run_context :root do
    edit_resource(:my_thing, "accumulated state") do
      action :nothing
      my_array_property << accumulate_some_stuff
    end
  end
  log "kick it off" do
    notifies :run, "my_thing[accumulated state], :delayed
  end
end
```

the log resource to kick off the delayed notification is midlly fugly and it feels like there should be some way to create-a-nothing-resource-and-fire-a-delayed-notification in one statement without invoking another resource -- but i think this gets the job done.  the accumulated notifications should all get de-duped.  any nested resources/subcontexts that call this action should bubble their state up into the accumulated resource in the root run_context and whatever the :my_thing resource does should only happen exactly once at the end of the chef-client run.

lamont-granquist:
the chef_rewind editing case is also taken care of here, as well as the case of deleting resources.

i added `#find_resource!` and `#find_resource` so that there's a consistent API to use since `#resources` raises without having a `!` on the end of it, so it doesn't really fit this pattern.  also `find_resource` with a block could be useful to insert a resource with default properties, but not to touch it if it already exists (as opposed to edit_resource which will update the resource if it finds it).

lamont-granquist:
and should be able to convert a definition-based accumulator to a resource like this:

```ruby
action :run do
  with_run_context :parent do
    edit_resource(:my_thing, "accumulated state") do
      my_array_property << accumulate_some_stuff
    end
  end
end

def after_created
  Array(action).each do |action|
    run_action(action)
  end
end
```

that forces the resource we're writing to run at compile-time and to edit its parent run_context -- which is all that definition-based accumulators do.  but in the process all the stupidity around the params hash in definitions is avoided, and you get strong validation of your properties on the resource which the params hash completely lacks.

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+      # will raise Chef::Exceptions::ResourceNotFound if the resource is not found.
+      #
+      # @param type [Symbol] The type of resource (e.g. `:file` or `:package`)
+      # @param name [String] The name of the resource (e.g. '/x/y.txt' or 'apache2')
+      # @param run_context [Chef::RunContext] the run_context of the resource collection to operate on
+      #
+      # @return [Chef::Resource] The resource
+      #
+      # @example
+      #   delete_resource(:template, '/x/y.txy')
```
thommay: ought to have a bang!

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+      def edit_resource!(type, name, created_at = nil, run_context: self.run_context, &resource_attrs_block)
+        resource = find_resource!(type, name, run_context: run_context)
+        resource.instance_eval(&resource_attrs_block) if block_given?
+        resource
+      end
+
+      # Lookup a resource in the resource collection by name.  If it exists,
+      # return it.  If it does not exist, create it.  This is a useful function
+      # for accumulator patterns.  In CRUD terminology this is an "upsert" operation and is
+      # used to assert that the resource must exist with the specified properties.
```
thommay: <3 this!

lib/chef/resource_collection/resource_list.rb
```diff
@@ -67,6 +67,16 @@ def insert(resource)
         end
       end
 
+      def delete(key)
+        raise ArgumentError, "Must pass a Chef::Resource or String to delete" unless key.is_a?(String) || key.is_a?(Chef::Resource)
+        key = key.to_s
+        ret = @resources.reject! { |r| r.to_s == key }
+        if ret.nil?
+          raise Chef::Exceptions::ResourceNotFound, "Cannot find a resource matching #{key} (did you define it first?)"
```
thommay: Not sure the message in the exception is correct here?

lamont-granquist: I think it should be the same text as find...   its just find+delete...

lib/chef/dsl/declare_resource.rb
lamont-granquist: knew i'd miss at least one of those...

lib/chef/resource_collection/resource_list.rb
thommay: ok, now i understand what `reject!` returns, 👍

thommay:
:+1: LGTM

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+            rc
+          when :root
+            Chef.run_context
+          when :parent
+            run_context.parent_run_context
+          else
+            raise "bad argument to run_context helper, must be :root, :parent, or a Chef::RunContext"
+          end
+        ret = yield
+        @run_context = @old_run_context
```
coderanger: This should be in an `ensure` for safety.

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+      #   # creates/returns a 'service[foo]' resource in the root run_context
+      #   resource = with_run_context(:root)
+      #     edit_resource(:service, "foo") do
+      #       action :nothing
+      #     end
+      #   end
+      #
+      def with_run_context(rc, &block)
+        raise ArgumentError, "with_run_context is useless without a block" unless block_given?
+        @old_run_context = @run_context
```
coderanger: This isn't safe to use in a nested fashion, should save to a local variable instead.

coderanger: (alternatively error is this is already set, but that seems worse)

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+      # @example
+      #   # creates/returns a 'service[foo]' resource in the root run_context
+      #   resource = with_run_context(:root)
+      #     edit_resource(:service, "foo") do
+      #       action :nothing
+      #     end
+      #   end
+      #
+      def with_run_context(rc, &block)
+        raise ArgumentError, "with_run_context is useless without a block" unless block_given?
```
coderanger: You are mixing styles, either use the `&block` argument or `block_given?`/`yield`, both seems weird.

lib/chef/dsl/declare_resource.rb
```diff
@@ -23,7 +23,180 @@ class Chef
...
+        @run_context =
+          case rc
+          when Chef::RunContext
+            rc
+          when :root
+            Chef.run_context
+          when :parent
+            run_context.parent_run_context
+          else
+            raise "bad argument to run_context helper, must be :root, :parent, or a Chef::RunContext"
```
coderanger: Missing exception type.

coderanger:
This was already merged but there are some things we should follow up on.

PR #4837, opened 2016-04-18
Refactor ChefFS files to be files

thommay:
Internally, this is FLOW-253:
> Currently the ChefFS code has many different branches to handle files and directories differently,
> which makes it difficult to understand. We can increase the usability of ChefFS by splitting
> directories and files into specific implementations, making those base classes do less.
> 
> Similar to the class Chef::ChefFS::FileSystem::Repository::Directory class, this ticket is
> about creating a class Chef::ChefFS::FileSystem::Repository::File and using it where appropriate.

(it turned out to be BaseFile because File is just annoying)


lamont-granquist:
👍

thommay:
@chef/client-core this is ready for review.

danielsdeleo:
woot! 👍

PR #4838, opened 2016-04-18
add global and per-resource toggles for resource cloning behavior

lamont-granquist:
let us turn on and off resource cloning

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -1053,7 +1053,10 @@ module ChefConfig
 
     default :rubygems_url, "https://rubygems.org"
 
-    # For Chef-12 this is still the default, in Chef-13 the default will be false
+    # This controls the behavior of resource cloning (and CHEF-3694 warnings).  For Chef < 12 the behavior
+    # has been that this is 'true', in Chef 13 this will change to false.  Setting this to 'true' in Chef
+    # 13 is not a viable or supported migration strategy since Chef 13 community cookbooks will be expected
+    # to break with this setting set to 'true'.
     default :resource_cloning, true
 
     # If installed via an omnibus installer, this gives the path to the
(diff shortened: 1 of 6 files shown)


lamont-granquist:
still need to think through the behavior of notifications and should probably add a warning to notifications when it finds more than one resource and be a bit more verbose about which one gets picked.

also need to add a couple more tests

coderanger:
I don't think offering a global option is helpful. Cookbook code will be written expecting one behavior or the other and changing it in a way the cookbook author can't control seems bad :-/

lamont-granquist:
Lets you toggle it though and see what breaks in order to fix it (I'll certainly run with that setting and see what explodes).

coderanger:
The problem is the person toggling it is very rarely the person that could or will fix things. If, right before Chef 13, we wanted to have some kind of "send me to the future to see what happens" feature flag I could get more behind that but I don't think this has more value than the existing deprecation warnings and the ability to make deprecation warnings into errors.

lamont-granquist:
So I sorta view the Chef::Config option here as exactly the 'send me to the future' option...  I'll set it on my runs, I can set it to run it through internal CI acceptance testing, etc.  Even if its literally just an option for me to test with, i kinda want that option for me...

lamont-granquist:
apparently @thommay just told me i volunteered to write another RFC...  *sigh*...   to the thursday meeting then....

thommay:
In a lightning change of mind, let's just land this so we can use it for testing. Let's put some BIG FUCKING WARNINGS on this toggle if we document it.

tas50:
We really need a way to turn cloning off so we can perform some extensive testing across the cookbook ecosystem including private customer cookbooks. This is a pretty big change and I'm afraid we're going to push it out with good intentions, but cause a huge amount of customer pain by accident. We don't need to document the flag for general usage. We just need to enable the ability to turn off cloning so we can test community cookbooks and customer support teams can test private cookbooks.

lamont-granquist:
let me look through it again before mashing merge...

lamont-granquist:
simplified PR, added scarier banner, we can also add a deprecation warning or remove the toggle entirely in Chef 13.

PR #4839, opened 2016-04-18
fix details in with_run_context

lamont-granquist:


coderanger:
:+1: 

Some day we should get chefstyle to standardize us on one block passing style (probably `&block`) but that can be a battle for another day.

lamont-granquist:
I thought we actually did and I'm surprised that this didn't piss it off.  The `&block` thing actually has a perf overhead.

thommay:
👍

PR #4840, opened 2016-04-19
Don't use berks when bumping version--can't possibly change.

jkeiser:
Just change the top level bundle.

I suspect this is the issue causing version bumper to fail.


ksubrama:
:+1:

PR #4841, opened 2016-04-19
Run in fips mode if node is fips enabled

mwrock:
This checks ohai fips data for a default fips enabled flag.

Note the synchronization of the ohai check. This was added after witnessing odd behavior during integration tests. When ohai would attempt to parse the plugin name from its class constant, the parcing would intermittently raise exceptions and upon closer inspection the class name was stripped leaving only a hash. After much debugging and nashing of teeth, I discovered that ohai removes all class constants upon instantiating an individual `Ohai::System` object. The multithreaded integration tests were there fore corrupting the plugin maps of neghboring ohai instances. This could certainly happen in "real life" too.

So we perform the lookup inside a `Mutex` and only once per ruby process.


chef-config/lib/chef-config/config.rb
```diff
@@ -513,7 +513,32 @@ def self.path_accessible?(path)
...
+    default(:fips) do
+      !ENV["CHEF_FIPS"].nil? || check_fips_via_ohai
+    end
+
+    # we want to synchronize this ohai call because ohai is not thread safe
+    # if this gets called in a mulithreaded context, each thread's ohai instance
+    # will call reset_system while other threads are loading plugins
+    # the destructive power of reset_system is scoped to the module and not to the instance
+    def self.check_fips_via_ohai
+      @@sync ||= Mutex.new
```
danielsdeleo: `||=` in ruby is two ruby instructions which can get context switched, so it's best to avoid it when using threads. I'd recommend instead putting `@@sync = Mutex.new` in the class body.

Also, do you need the extra behaviors of class variables here (instead of a class instance variable)?

mwrock: ok thats good to know. That was the one line I was most concerned about.

You mean using `@var` instead of `@@var` right. Because `default` is a class scoped method, I assumed this was the right approach.My understanding (perhaps misguided) is that `@var` is instance scoped and thus not a good fit here.

danielsdeleo: This is just one of the weird things about everything in ruby being an object. Since a class is an instance of the `Class` class, it can have its own instance variables, which are separate from instance variables of instances of that class. E.g.,

```ruby
class AClass

  puts self.inspect

  @class_ivar = "a class instance variable"

  def self.class_ivar
    puts "class_ivar is: #{@class_ivar.inspect}"
  end

  def cant_see_class_ivar
    puts "class_ivar is: #{@class_ivar.inspect}"
  end

end

AClass.class_ivar

AClass.new.cant_see_class_ivar
```

That will output:

```
AClass
class_ivar is: "a class instance variable"
class_ivar is: nil
```

Class variables in ruby are shared by all subclasses, so there are a few times when you really need them (keeping track of plugin loading in knife is one example), but usually you want a variable scoped to just that one class, which is what class instance variables will do.

mwrock: Thanks for the explanation! I'll change to instance scope.

mwrock:
Using instance variables now and moving mutex assignment into class.

danielsdeleo:
LGTM 👍

ksubrama:
:+1:

lamont-granquist:
@mwrock @danielsdeleo @ksubrama isn't this a bit completely circularly insane now?

we use chef-config in ohai, so what happens when we create an ohai object and then call load_plugins on it from the chef-config that is being included from ohai? 

i think i'm :-1: on this, and we should probably revert and rethink this, and not pull the dependency in here.

thommay:
I'm reverting this, it explodes violently on OS X at least since there's no FIPS Ohai plugin there (presumably also every platform other than linux and windows).

lamont-granquist:
So IMO if you want code sharing this is the wrong way to do it.  We've already got Chef::Utils stuff in chef-config and you should exploit that to drop common code in there to test for fips.  Then ohai can populate its value with a call into that helper and chef-config can also use that helper.  This approach I'm pretty certain is crazy and we should never be calling ohai plugins from chef-config, and should not have ohai as a dep of chef-config.  That just smells like all kinds of wrong, and the dependency needs to be inverted and the common code needs to be moved into the gem/class that is already the dependency.

Then there's also the OSX issue to fix.

PR #4843, opened 2016-04-19
Refreeze windows

jkeiser:
We already install it frozen elsewhere; this makes Windows do the same as other platforms.


mwrock:
:+1:

randomcamel:
👍

PR #4844, opened 2016-04-19
Update changelog generator to latest

jkeiser:
(we have a lockfile now, so we are protected against inadvertent updates)


randomcamel:
👍

Issue #3456, opened 2015-06-01
Chef 12.3.0 and chef_gem install issues behind firewall


avsh:
My node doesn't have an internet access at all.

I'm trying to install gems from local file system but the setup fails due to gem binary wants to connect to rubygems.org

The command 
`/opt/chef/embedded/bin/gem install /var/chef/cache/gems/mail-2.6.3.gem -q --no-rdoc --no-ri -v "2.6.3"`
fails to run.

I run the command locally and it returned the warning

`WARNING:  Unable to pull data from 'https://rubygems.org/': no such name (https://rubygems.org/specs.4.8.gz)`

I assume it shouldn't try to connect to rubygems if `source` is in use.

pburkholder:
Hi @avsh 

Use the attribute `clear_sources true` to remove rubygems.org from the sources list.

The old default behavior is it only add the 'source' to the list of available sources, and then use rubygems.org as the last one on the list.

The docs have only recently been updated, https://docs.chef.io/resource_gem_package.html,
and see also https://github.com/chef/chef/pull/2867

HTH,

Peter

avsh:
Hi @pburkholder 

Thank you for the info. 

I've tried to use the ``clear_sources true`` as you advised but it fails.

Here is my code

```ruby
chef_gem 'mail' do
  compile_time true if respond_to?(:compile_time)
  source "#{::Chef::Config[:file_cache_path]}/gems/mail-2.6.3.gem"
  clear_sources true
end
```

and Chef failed because my squid didn't allow it to make outbound connection

```sh
192.168.35.199     Mixlib::ShellOut::ShellCommandFailed
192.168.35.199     ------------------------------------
192.168.35.199     Expected process to exit with [0], but received '1'
192.168.35.199     ---- Begin output of /opt/chef/embedded/bin/gem install /var/chef/cache/gems/mail-2.6.3.gem -q --no-rdoc --no-ri -v "2.6.3" ----
192.168.35.199     STDOUT:
192.168.35.199     STDERR: ERROR:  While executing gem ... (Net::HTTPServerException)
192.168.35.199         403 "Forbidden"
192.168.35.199     ---- End output of /opt/chef/embedded/bin/gem install /var/chef/cache/gems/mail-2.6.3.gem -q --no-rdoc --no-ri -v "2.6.3" ----
[...]

PR #4847, opened 2016-04-19
Update rubygems provider to support local install of gems if so specified

PrajaktaPurohit:
```
➜  chef git:(ksubrama+praj/FLOW-298/local_gem) ✗ gem uninstall a
Successfully uninstalled a-0.1.1

##Disable internet before testing

➜  chef git:(ksubrama+praj/FLOW-298/local_gem) ✗ bundle exec sudo bin/chef-apply -e "gem_package '/Users/prajaktapurohit/Downloads/a-0.1.1.gem'"

  * gem_package[/Users/prajaktapurohit/Downloads/a-0.1.1.gem] action install[2016-04-19T12:32:39-07:00] WARN:  gem package /Users/prajaktapurohit/Downloads/a-0.1.1.gem does not satisfy the requirements /Users/prajaktapurohit/Downloads/a-0.1.1.gem (>= 0)

    - install version  of package /Users/prajaktapurohit/Downloads/a-0.1.1.gem

➜  chef git:(ksubrama+praj/FLOW-298/local_gem) ✗ gem list | grep "a ("
a (0.1.1)
aruba (0.14.1, 0.11.2, 0.10.2)
```
/cc #@chef/client-core @ksubrama @mwrock

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -517,7 +517,9 @@ class Chef
               install_via_gem_command(name, version)
             end
           elsif @new_resource.gem_binary.nil?
-            @gem_env.install(@new_resource.source)
+            # domain is used by Gem::DependencyInstaller rather than by Chef code
+            # domain can be :local, :remote or :both
+            @gem_env.install(@new_resource.source, domain: :local)
           else
             install_via_gem_command(name, version)
           end
@@ -531,6 +533,7 @@ class Chef
         def install_via_gem_command(name, version)
           if @new_resource.source =~ /\.gem$/i
             name = @new_resource.source
+            src = " --local" unless source_is_remote?
           elsif @new_resource.clear_sources
             src = " --clear-sources"
             src << (@new_resource.source && " --source=#{@new_resource.source}" || "")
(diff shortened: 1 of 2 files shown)


ksubrama:
Fixes https://github.com/chef/chef/issues/3456

lib/chef/provider/package/rubygems.rb
```diff
@@ -517,7 +517,7 @@ def install_package(name, version)
               install_via_gem_command(name, version)
             end
           elsif @new_resource.gem_binary.nil?
-            @gem_env.install(@new_resource.source)
+            @gem_env.install(@new_resource.source, domain: :local)
```
randomcamel: It'd be nice to have a comment here pointing out that `domain` is used by `Gem::DependencyInstaller` rather than by Chef code.

randomcamel:
👍

btm:
I want to turn off emoji feature. :+1: because I'm here.

sczizzo:
This change breaks the behavior experienced prior to Chef v12.10.24, as brought up in #4955. I maybe wouldn't mind this, except with this change I don't see any way to remove the `--local` option added when installing a gem from the filesystem.

Prior to this change, you'd "opt-in" by passing `options '--local'` to the resource; now I don't have any way to opt out AFAICT.

ksubrama:
Ouch!  Thank you for reporting this @sczizzo.  We'll try to find a good middle ground.

PR #4855, opened 2016-04-21
Add layout option support for device creation to mdadm resource provider

kbruner:


lamont-granquist:
👍

thommay:
👍 let's make sure that travis job passes now

lamont-granquist:
If anyone digs this up in the future and is looking at adding more options to mdadm, it might be a good idea to consider adding an `options` property to the mdadm resource so that you could `options "--whatever foo --whateverelse bar"` and dump it all into the command.

PR #4860, opened 2016-04-23
The easy_install provider and resource are deprecated and will be removed in Chef 13

coderanger:
This is because easy_install itself is long-since deprecated by the Python community and has been de-facto unmaintained for around 5 years.

Ping @chef/client-core for review.


lamont-granquist:
seems fine to me...  👍

PR #4864, opened 2016-04-25
Update override path

scotthain:
@chef/client-core @chef/engineering-services

### DIFF
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -22,6 +22,9 @@ license "Apache-2.0"
 license_file "../LICENSE"
 
 build_iteration 1
+# Do not use __FILE__ after this point, use current_file. If you use __FILE__
+# after this point, any dependent defs (ex: angrychef) that use instance_eval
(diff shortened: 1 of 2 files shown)


thommay:
Some explanation in the comment as to why we shouldn't use `__FILE__` would be very helpful here.

lamont-granquist:
yeah what happens to `__FILE__` that makes it so we can't use it?  i'm puzzled...

scotthain:
@thommay @lamont-granquist updated comment as follows:
```
# Do not use __FILE__ after this point, use current_file. If you use __FILE__
# after this point, any dependent defs (ex: angrychef) that use instance_eval
# will fail to work correctly.
```

thommay:
👍 thanks @scotthain

lamont-granquist:
oh right...

👍

scotthain:
@thommay y'alls do the merging yourself in chef/chef these days, right?

schisamo:
@scotthain You can merge after you get enough 👍 's (which I think you have).

Issue #4862, opened 2016-04-24
knife cookbook site install broken on windows


willoucom:
## Description

On windows, the command knife cookbook site install "package" return an error with tar

The error seems to occur because the tar.exe installed doesn't support the option --force-local

## Chef Version

Irrelevant, client-side issue

## Platform Version

Windows 10
knife --version : Chef: 12.8.1

## Replication Case

Download latest chef-dk for windows (chefdk-0.12.0-1-x86.msi), install it.

Run : knife cookbook site install "package"

## Client Output

$ knife cookbook site install docker
[snip]
Uncompressing docker version 2.6.8.
ERROR: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
---- Begin output of tar zxvf "D:/Chef/chef-repo/cookbooks/docker.tar.gz" --force-local ----
STDOUT:
STDERR: tar: Option --force-local is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
---- End output of tar zxvf "D:/Chef/chef-repo/cookbooks/docker.tar.gz" --force-local ----
Ran tar zxvf "D:/Chef/chef-repo/cookbooks/docker.tar.gz" --force-local returned 1

## Quickfix and Workaround

You can bypass this error by commenting a single line in : 
/opscode/chefdk/embedded/lib/ruby/gems/2.1.0/gems/chef-12.8.1-universal-mingw32/lib/chef/knife/cookbook_site_install.rb

Comment the line 147 : 
            extract_command << " --force-local"

Retry the command : 
knife cookbook site install docker

Installing docker to D:/Chef/chef-repo/cookbooks
Checking out the master branch.
Pristine copy branch (chef-vendor-docker) exists, switching to it.
Downloading docker from Supermarket at version 2.6.8 to D:/Chef/chef-repo/cookbooks/docker.tar.gz
Cookbook saved: D:/Chef/chef-repo/cookbooks/docker.tar.gz
Removing pre-existing version.
Uncompressing docker version 2.6.8.
Removing downloaded tarball
43 files updated, committing changes

willoucom:
I have done some test and research, and it seems this issue occured after PR #3968 

From my humble point of view, quoting the path should have solved the initial issue.

I have made more testing and the same issue occurs using tar.exe from git for windows or tar.exe from chef-dk.

[...]

PR #4867, opened 2016-04-26
cookbook site install : tar error on windows

willoucom:
tar.exe  bundled with chef-dk on windows don't support the --force-local option

> STDERR: tar: Option --force-local is not supported

- fixes : #4862

### DIFF
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -144,11 +144,11 @@ class Chef
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
         if Chef::Platform.windows?
           # FIXED: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
-          tar_version = shell_out("tar --version").stdout.tr("\n"," ")
-          if /GNU tar/.match(tar_version)
+          tar_version = shell_out("tar --version").stdout.tr("\n", " ")
+          if tar_version =~ /GNU tar/
             ui.info("GNU tar detected")
             extract_command << " --force-local"
-          elsif /bsdtar/.match(tar_version)
+          elsif tar_version =~ /bsdtar/
             ui.info("BSD tar detected")
           end
         end
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -141,8 +141,17 @@ class Chef
 
       def extract_cookbook(upstream_file, version)
         ui.info("Uncompressing #{@cookbook_name} version #{version}.")
-        # FIXME: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
+        if Chef::Platform.windows?
+          # FIXED: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
+          tar_version = `tar --version`.tr("\n"," ")
+          if /GNU tar/.match(tar_version)
+            ui.info("GNU tar detected")
+            extract_command << " --force-local"
+          elsif /bsdtar/.match(tar_version)
+            ui.info("BSD tar detected")
+          end
+        end
         shell_out!(extract_command, :cwd => @install_path)
       end
 
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -143,9 +143,6 @@ class Chef
         ui.info("Uncompressing #{@cookbook_name} version #{version}.")
         # FIXME: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
-        if Chef::Platform.windows?
-          extract_command << " --force-local"
-        end
         shell_out!(extract_command, :cwd => @install_path)
       end
 
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -145,10 +145,10 @@ class Chef
         if Chef::Platform.windows?
           tar_version = shell_out("tar --version").stdout.tr("\n", " ")
           if tar_version =~ /GNU tar/
-            ui.debug("GNU tar detected, adding --force-local")
+            Chef::Log.debug("GNU tar detected, adding --force-local")
             extract_command << " --force-local"
           else
-            ui.debug("non-GNU tar detected, not adding --force-local")
+            Chef::Log.debug("non-GNU tar detected, not adding --force-local")
           end
         end
         shell_out!(extract_command, :cwd => @install_path)
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -143,13 +143,12 @@ class Chef
         ui.info("Uncompressing #{@cookbook_name} version #{version}.")
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
         if Chef::Platform.windows?
-          # FIXED: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
           tar_version = shell_out("tar --version").stdout.tr("\n", " ")
           if tar_version =~ /GNU tar/
-            ui.info("GNU tar detected")
+            ui.debug("GNU tar detected, adding --force-local")
             extract_command << " --force-local"
-          elsif tar_version =~ /bsdtar/
-            ui.info("BSD tar detected")
+          else
+            ui.debug("non-GNU tar detected, not adding --force-local")
           end
         end
         shell_out!(extract_command, :cwd => @install_path)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Yeah, we ship BSD tar in embedded/bin with the omnibus msi

Unfortunately, this undos the 'fix' in 12.6 for gnu tar which deliberately added this flag on windows.

I believe we're caught in some warfare here between people who have put embedded/bin in their $PATH and then added a GNU tar in front of that and people who have not.

I think we need another fix to path_sanity to make it a bit more assertive and fix chef-client's PATH so that embedded/bin is always first and you can't opt-out of it any more.

Of course that'll get into a fight with people who have deliberately put BSD tar first in their PATH and then expect execute resources and such to pick it up -- so we probably need to enforce in shell_out that embedded/bin comes first, while in shell_out_with_systems_locale that embedded is not added (or that we don't apply any PATH fixes at all).

lamont-granquist:
I don't know if we hit merge on this one and move forwards or not, the yak here goes exceedingly deep...

The existing code, though, is particularly bad because without both:

1.  adding embedded/bin to the path so that path_sanity will not add it first
2.  adding gnu tar in front of that to the path so that gnu tar is picked up first

Then knife cookbook site install is broken on windows.

This is the correct fix for a vanilla chef/chef-dk install on windows.

So I think we need to merge this, and break everyone that was 'fixed' by the change in 12.6

coderanger:
Alternatively we could stop using an external tar binary and use the support baked in to Ruby: https://github.com/poise/poise-archive/blob/master/lib/poise_archive/archive_providers/tar.rb as an example. For the file sizes involved in cookbook sharing, it should be fine.

coderanger:
Do we use the included `tar.exe` for anything other than supermarket downloads?

lamont-granquist:
Dunno.  I do think we need to sort out the embedded path issue.  But converting to pure-ruby tar where we can also makes a hell of a lot of sense to just side step all these issues...

willoucom:
Indeed, i think this is a good opportunity and a good idea to use native tar.rb. 
However i believe this will be a lot of work to do just to make happy the people who fiddle with the $PATH too much.
From my own experience, using cygwin/babun gave me such headaches that i switched back to the old batch scripts under cmd.exe

Could it be possible to merge the fix anyway ?
Thanks in advance

lamont-granquist:
The problem is that if we merge this we fix you, but we break the person who submitted the PR to 12.6.0 to add this.

So we're stuck with flipping a coin to decide which set of users to break.

In that case, probably not.  Particularly when using pure-ruby tar will fix both use cases.

willoucom:
I have made more research and i found a possible alternate solution.

- tar.exe bundled with chef-dk on windows doesn't support the option --force-local 
- tar.exe in cygwin or mingw32 (in git for windows) support the option --force-local

i'm not sure, but could it be possible to "update" tar.exe bundled in chef-dk to a version compatible with --force-local ?

Using the --force-local option for fix an incorrectly escaped $path is not the right solution, but hopefully, someone already fixed escaping in a previous release.

See man tar : 

> --force-local
> Forces tar to interpret the file name given to --file as a local file, even if it looks like a remote tape drive name.

Maybe the good solution was to remove --force-local and upgrading tar.exe

lamont-granquist:
Anyone that puts git for windows in front of the chef-dk embedded path will wind up with a tar that doesn't support --force-local and will be broken.  That is likely the exact set of users I'm talking about would be broken.

lamont-granquist:
And if we remove --force-local then its the other way around.   That will work with the existing chef-dk tar.exe but anyone that puts a GNU tar.exe first in the path will break.  We definitely wouldn't want to switch the tar.exe in chef-dk in that case since those changes would break it again.

willoucom:
I understand, but for the moment the tar.exe embedded with chef-dk DON'T support --force-local making 'knife cookbook site install' unusable without installing another tar.exe and fiddling with the windows path

coderanger:
As a short term fix we could remove the option and use the absolute path to the included tar executable.

willoucom:
Gentlemens, i have made some changes.

With this new version, tar is usable with both versions of tar.exe

Embeded (Bsd) tar

> D:\Chef\chef-repo> knife cookbook site install httpd
> Installing httpd to D:/Chef/chef-repo/cookbooks
> Checking out the master branch.
> Creating pristine copy branch chef-vendor-httpd
> Downloading httpd from Supermarket at version 0.3.6 to D:/Chef/chef-repo/cookbooks/httpd.tar.gz
> Cookbook saved: D:/Chef/chef-repo/cookbooks/httpd.tar.gz
> Removing pre-existing version.
> Uncompressing httpd version 0.3.6.
> BSD tar detected
> Removing downloaded tarball
> 1 files updated, committing changes
> Creating tag cookbook-site-imported-httpd-0.3.6
> Checking out the master branch.
> Updating 49c1f1b..6a2a0dd
> Fast-forward
>  cookbooks/httpd/CHANGELOG.md                       | 153 ++++
[snip]
> Cookbook httpd version 0.3.6 successfully installed

Github/gitforwindows/cygwin (Gnu) tar

> D:\Chef\chef-repo> knife cookbook site install java
> Installing java to D:/Chef/chef-repo/cookbooks
> Checking out the master branch.
> Creating pristine copy branch chef-vendor-java
> Downloading java from Supermarket at version 1.39.0 to D:/Chef/chef-repo/cookbooks/java.tar.gz
> Cookbook saved: D:/Chef/chef-repo/cookbooks/java.tar.gz
> Removing pre-existing version.
> Uncompressing java version 1.39.0.
> GNU tar detected
> Removing downloaded tarball
> 1 files updated, committing changes
> Creating tag cookbook-site-imported-java-1.39.0
> Checking out the master branch.
> Updating 6a2a0dd..31b1530
> Fast-forward
>  cookbooks/java/.gitignore                          |  47 +++
[snip]
> Cookbook java version 1.39.0 successfully installed
 
I think printing the current "version" of tar will be helpfull for users

feel free to test with many version of tar for windows as possible, i have only found 2 distinct versions for the moment

lib/chef/knife/cookbook_site_install.rb
```diff
@@ -141,8 +141,17 @@ def download_cookbook_to(download_path)
 
       def extract_cookbook(upstream_file, version)
         ui.info("Uncompressing #{@cookbook_name} version #{version}.")
-        # FIXME: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
+        if Chef::Platform.windows?
+          # FIXED: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
+          tar_version = `tar --version`.tr("\n"," ")
```
coderanger: We don't use backticks for command execution in Chef given it has many many problems. See also, the use of `shell_out` right below this.

coderanger:
Not sure why github things that was an outdated diff comment, but ^^ we can't merge this as is.

MichaelPereira:
@willoucom I'm the one who implemented the previous PR, and I had stumbled into this problem a few days back. You beat me to the correct implementation which handles both case, by running tar first to look if it will support that option. I support the changes made in 9856652

willoucom:
i have corrected the backticks

willoucom:
@MichaelPereira thanks for your support

lib/chef/knife/cookbook_site_install.rb
```diff
@@ -141,10 +141,16 @@ def download_cookbook_to(download_path)
 
       def extract_cookbook(upstream_file, version)
         ui.info("Uncompressing #{@cookbook_name} version #{version}.")
-        # FIXME: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
         extract_command = "tar zxvf \"#{convert_path upstream_file}\""
         if Chef::Platform.windows?
-          extract_command << " --force-local"
+          # FIXED: Detect if we have the bad tar from git on Windows: https://github.com/opscode/chef/issues/1753
```
thommay: this comment can go away now :)

lib/chef/knife/cookbook_site_install.rb
```diff
@@ -141,10 +141,16 @@ def download_cookbook_to(download_path)
...
+          tar_version = shell_out("tar --version").stdout.tr("\n", " ")
+          if tar_version =~ /GNU tar/
+            ui.info("GNU tar detected")
+            extract_command << " --force-local"
+          elsif tar_version =~ /bsdtar/
+            ui.info("BSD tar detected")
```
lamont-granquist: that should probably be debug level output rather than ui.info.

the elsif is also not terribly useful since its mostly a no-op, and there's the third case which is a missing else, and it doesn't really communicate what its doing when it detects it...

```
if tar_version =~ /GNU tar/
  ui.debug("GNU tar detected, adding --force-local")
  extract_command << " --force-local"
else
  ui.debug("non-GNU tar detected, not adding --force-local")
end
```

lamont-granquist: something like that?

lamont-granquist:
👍  on the approach, just a few nits

thommay:
👍 after nit eradication

willoucom:
Gentlemens, thanks a lot for your time

lamont-granquist:
oh i guess there's no `ui.debug` sorry, i think you want `Chef::Log.debug`

PR #4869, opened 2016-04-26
Lazy'ing candidate_version in package provider

### LINKED ISSUE #3867: gem_package "upgrade" action may be broken/confusing

### LINKED ISSUE #4122: Performance Issue with Chef::Provider::Package hitting candidate_version when action is :install and the package is already installed
folks, i’m wondering if there’s a bug with the chef_gem resource.
if i don’t specify a version and the gem is already installed i get this:
```
DEBUG: chef_gem[knife-lastrun] found installed gem knife-lastrun version 0.0.5 matching knife-lastrun (>= 0)
DEBUG:  found gem knife-lastrun version 0.0.5 for platform ruby from #<Gem::Source:0x00000004cfcd18>
DEBUG: chef_gem[knife-lastrun] knife-lastrun 0.0.5 already installed
DEBUG: chef_gem[knife-lastrun] is already installed - nothing to do (up to date)
```
if i do specify the version and it’s already installed i get this:
```
DEBUG: chef_gem[knife-lastrun] found installed gem knife-lastrun version 0.0.5 matching knife-lastrun (= 0.0.5)
DEBUG: chef_gem[knife-lastrun] knife-lastrun 0.0.5 satisifies 0.0.5 requirement
DEBUG: chef_gem[knife-lastrun] is already installed - nothing to do (up to date)
```
the key thing to notice here is the `found gem knife-lastrun version 0.0.5 for platform ruby from #<Gem::Source:0x00000004cfcd18>`. when i don’t specify a version it seems it always searches for the gem even though the gem is already installed. is this the intended behavior for action install? it seems like this should be the behavior for action upgrade to always search for the latest gem...

the declared resource is as follows:
```
  chef_gem 'knife-lastrun' do
    version '0.0.5'
    clear_sources true if respond_to?(:clear_sources)
    source 'http://my.awesome.private.rubygems.cool'
    compile_time true if respond_to?(:compile_time)
  end
```

here's what i expected to see:
chef_gem action install with no version specified:
- install the latest gem if it's not already installed. this requires to search for the gem
- do nothing, not even search for the gem, if it's already installed

### LINKED ISSUE #4691: slow performance in yum_package provider in 12.7.x through 12.9.x
## Description

* running chef-solo from the chef11 gem completes in roughly 14 seconds
* running chef-solo from the chef12 gem completes in over 2 minutes

## Chef Version

11.8.2 -> 12.8.1

## Platform Version

CentOS release 6.7 (Final)

## Replication Case

run the exact same recipes on the exact same hardware under both versions. 

behind the scenes, i noticed that yum-dump.py was running a LOT more under chef12. it appears to run on every single package idempotency check.

## Client Output

https://gist.github.com/downneck/d019bfa4e9cccc65f7ad

## Stacktrace

Please include the stacktrace.out output or link to a gist of it, if there is one.

lamont-granquist:
This does many things:

1.  Fixes a bug in handling `:update` actions where partial version matches in the new_resource.version like `~> 1.0` wouldn't update 1.1 to 1.2 when it came out.
2.  The rubygems provider was abusing the crap out of setting the candidate_version to nil and using side effects of lying about it to achieve its goals, which has been removed
3.  There's a lot of documentation added around the two badly-named subclass hooks now
4.  The candidate_version is now Lazy'd in the package provider so that we can walk all the arrays in parallel, while still not accessing the candidate_version and doing expensive operations even though we might wind up being completely idempotent.
5.  The LazyArray is necessary for getting multipackage's to be Lazied as well.
6.  The seeming complexity of the Decorator base class is me pulling in work from the attributes-v1.99 branch, it isn't strictly necessary to solve this problem, but it breaks things up a bit.

The Lazy and LazyArray stuff almost smells wrong to me.  We may be able to solve the problem using the lazy iterators already in ruby.  However to do that probably means re-thinking the approach I took to walking through the arrays with the `#each_package` helper, and I'd prefer to make the minimal fix possible to the overall superclass engine here.  Instead we can make one single one-line change to the superclass to implement the laziness and keep the rest of it the same.

The Decorator base class gives me a platform to start to build things on like the decorators that I have written in that attributes-v1.99 branch to unchain method calls and handle set_unless and the Chef-12 attributes APIs much better.  We could also build a consistent Monad class on top of that at some point.

closes #4608 
closes #3867
closes #4122 

I suspect this does not completely fix #4691, but might make things better, and certainly lays the groundwork for getting that fixed.

There were reasons I had in #4608 for wanting to address both #3867 and #4122 together and I think it was all the nasty magical coupling between the rubygems provider and the candidate_version in the superclass which had to be removed to make #3867 work, but make #4122 much worse...

### DIFF
--- a/lib/chef/decorator.rb
+++ b/lib/chef/decorator.rb
@@ -1,0 +1,81 @@
+#--
+# Copyright:: Copyright 2016 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "delegate"
+
+class Chef
+  class Decorator < SimpleDelegator
+    NULL = ::Object.new
+
+    def initialize(obj: NULL)
+      super(obj) unless obj == NULL
+      @__defined_methods__ = []
+    end
+
+    # if we wrap a nil then decorator.nil? should be true
+    def nil?
+      __getobj__.nil?
+    end
+
+    # if we wrap a Hash then decorator.is_a?(Hash) should be true
+    def is_a?(klass)
+      __getobj__.is_a?(klass) || super
+    end
+
+    # if we wrap a Hash then decorator.kind_of?(Hash) should be true
+    def kind_of?(klass)
+      __getobj__.kind_of?(klass) || super
+    end
+
+    # reset our methods on the instance if the object changes under us (this also
+    # clears out the closure over the target we create in method_missing below)
+    def __setobj__(obj)
+      __reset_methods__
+      super
+    end
+
+    # this is the ruby 2.2/2.3 implementation of Delegator#method_missing() with
+    # adding the define_singleton_method call and @__defined_methods__ tracking
+    def method_missing(m, *args, &block)
+      r = true
+      target = self.__getobj__ { r = false }
+
+      if r && target.respond_to?(m)
+        # these next 4 lines are the patched code
+        define_singleton_method(m) do |*args, &block|
+          target.__send__(m, *args, &block)
+        end
+        @__defined_methods__.push(m)
+        target.__send__(m, *args, &block)
+      elsif ::Kernel.respond_to?(m, true)
+        ::Kernel.instance_method(m).bind(self).call(*args, &block)
+      else
+        super(m, *args, &block)
+      end
+    end
+
+    private
+
+    # used by __setobj__ to clear the methods we've built on the instance.
+    def __reset_methods__
+      @__defined_methods__.each do |m|
+        singleton_class.send(:undef_method, m)
+      end
+      @__defined_methods__ = []
+    end
+  end
+end
--- a/lib/chef/decorator.rb
+++ b/lib/chef/decorator.rb
@@ -21,9 +21,9 @@ class Chef
   class Decorator < SimpleDelegator
     NULL = ::Object.new
 
-    def initialize(obj: NULL)
-      super(obj) unless obj == NULL
+    def initialize(obj = NULL)
       @__defined_methods__ = []
+      super unless obj.equal?(NULL)
     end
 
     # if we wrap a nil then decorator.nil? should be true
--- a/lib/chef/decorator/lazy.rb
+++ b/lib/chef/decorator/lazy.rb
@@ -23,15 +23,25 @@ class Chef
     # called against the object.
     #
     # @example
-    #   a = Chef::Decorator::Lazy.new { puts "allocated" }
-    #   puts "start"
-    #   puts a.class
+    #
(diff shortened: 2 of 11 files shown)


lamont-granquist:
@wjordan if you're still around could you take a look at this?

you can skip over the overly-engineered decorator class, i have $REASONS behind do all that extra leg work that have nothing to do with this PR

the LazyArray is my solution to supporting multipackage as well.  i'm reasonably confident it works, although i think there has to be a better way, ultimately, if we get a chance to overhaul all this logic.

i added a lot of notes about the two comparison hooks because their names and how they were used were seriously confusing me.

i changed the matching logic in the rubygems provider which is where i could really use your eyeballs.   you seemed to be doing a lot of work, and i'm not quite certain if i missed some subtle untested edge case or not...

lamont-granquist:
@chef/client-core should be ready for review (hopefully appveyor doesn't troll me again)

thommay:
👍

wjordan:
Thanks for following up on this @lamont-granquist, I really appreciate it!

Changes look very good to me, extracting the Lazy-array pattern into a reusable class is an excellent approach. The important cases should still have the correct behavior considering the test suite is passing.

Regarding the matching logic in the rubygems provider, If I recall the reason for overriding the `target_version_already_installed?` method was to ensure that an "exact" target-version constraint such as `version '= 0.1.2'` would return `true` when the installed version is `0.1.2`, even though the string `'= 0.1.2'` does not strictly match `'0.1.2'` which would return `false` according to the superclass logic. This is a very minor edge-case (that I hadn't yet written a spec for), so it should be possible to patch in a followup PR if needed.

Also, maybe somewhat related, I don't consider the use case where `:upgrade` action is called against an equality-pinned version constraint to be too uncommon- for example, consider a third-party cookbook that passes a `node['cookbook']['gem_version']` attribute into an `:upgrade` action, where the `'version'` attribute should support any valid RubyGems version-constraint string (`'0.1.2'`, `'= 0.1.2'`, `'>= 0.1.0'`, `'~> 0.1'`, etc) and either update or pin the installed gem version appropriately. This is actually the use case I intended to support with my initial PR.

lamont-granquist:
Okay I think I see where you're coming from.  So we definitely do support "0.1.2" with :upgrade, but currently do not support "= 0.1.2" with :upgrade.

I think a better approach there might be to add a "exact: true" flag to `version_requirement_satisfied?` which is set when its called from the :upgrade path.  Then the version could be inflated to a Gem::Requirement exact match.  I think that makes sense if we go about renaming the API.  actually that makes `target_version_already_installed` only a check against current_version and candidate_version...

```ruby
current_version_equals_candidate_version?
new_version_satisfied?
new_version_satisfied?(exact: true)
```

That makes perfect sense to me because the current_version and candidate_version should never be fuzzy and should always equality check.  And the new_version may be fuzzy and should always go through the fuzzy handling code...

PR #4871, opened 2016-04-27
add tests for ChefFS path formatting

thommay:
@chef/client-core

### DIFF
--- a/spec/unit/chef_fs/config_spec.rb
+++ b/spec/unit/chef_fs/config_spec.rb
@@ -103,4 +103,133 @@ describe Chef::ChefFS::Config do
       expect(local_fs.child_paths["users"]).to eq([platform_path("/base_path/users")])
     end
   end
+
+  describe "formats paths", :unix_only do
+
+    let(:single_repo_path) do
+      Mash.new({
+        chef_repo_path: "/base_path",
+      })
+    end
+
+    let(:double_repo_path) do
+      Mash.new({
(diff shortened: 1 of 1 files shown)


spec/unit/chef_fs/config_spec.rb
```diff
@@ -103,4 +103,133 @@ def platform_path(*args)
...
+
+      it "returns the entry's path if the cwd isn't in the config" do
+        cfg = Chef::ChefFS::Config.new(config, "/my_repo/cookbooks")
+        expect(cfg).to receive(:base_path).and_return(nil)
+        expect(cfg.format_path(entry)).to eq(path)
+      end
+
+      it "returns . if the cwd is the same as the entry's path" do
+        cfg = Chef::ChefFS::Config.new(config, "/base_path/roles/foo.json")
+        expect(cfg).to receive(:base_path).and_return("/roles/foo.json").exactly(4).times
```
randomcamel: I usually see `exactly` as a test smell: say it breaks and someone looks at the test, is there an intuitive way to connect the number of times `base_path` gets called with what the code is doing? (Especially if our hypothetical debugger looks at the next 2 tests and sees different numbers for `exactly`.)

thommay: fair, can just make these `at_least`.

thommay:
@chef/client-core @randomcamel ready for more review thank you, please

lamont-granquist:
:+1:

PR #4872, opened 2016-04-27
Revert "Run in fips mode if node is fips enabled"

thommay:
Reverts chef/chef#4841

Per mine and Lamont's comments there, this is bad due to both circular dependency and that it kills every non linux/windows platform.


lamont-granquist:
👍

thommay:
for the record, on OS X:

```
3623 examples, 616 failures, 2 pending
```

PR #4873, opened 2016-04-27
Consolidate shell_out usage when calling git within it's HWRP.

joshburt:
@lamont-granquist has been gracious enough to provide specific feedback around https://github.com/chef/chef/pull/4771  I’ll be breaking out changes into the smallest atomic PRs that can be created.  

This pull request is _just_ the changes specific to the consolidation of the git shell_out usage.

**Overview**
The git HWRP currently has a number of instances where git is called directly via ‘shell_out’.  This PR attempts to consolidate the number of code paths to shell_out.

As always feedback is much appreciated.
Thanks! :)

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -199,10 +199,10 @@ class Chef
             #   which we can fix by replacing them all with our target url (hence the --replace-all option)
 
             if multiple_remotes?(remote_status) || !remote_matches?(remote_url, remote_status)
-                run_git_command(["config", "--replace-all", "remote.#{remote_name}.url", remote_url], { cwd: cwd })
+              run_git_command(["config", "--replace-all", "remote.#{remote_name}.url", remote_url], { cwd: cwd })
             end
           when 1
-              run_git_command(["remote", "add", remote_name, remote_url], { cwd: cwd })
+            run_git_command(["remote", "add", remote_name, remote_url], { cwd: cwd })
           end
         end
       end
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -105,7 +105,7 @@ class Chef
       end
 
       def git_minor_version
-        @git_minor_version ||= Gem::Version.new(shell_out!("git --version", run_options).stdout.split.last)
+        @git_minor_version ||= Gem::Version.new((run_git_command ["--version"]).stdout.split.last)
       end
 
       def existing_git_clone?
@@ -120,7 +120,7 @@ class Chef
         Chef::Log.debug("#{@new_resource} finding current git revision")
         if ::File.exist?(::File.join(cwd, ".git"))
           # 128 is returned when we're not in a git repo. this is fine
-          result = shell_out!("git rev-parse HEAD", :cwd => cwd, :returns => [0, 128]).stdout.strip
+          result = (run_git_command(["rev-parse", "HEAD"], { cwd: cwd, returns: [0, 128] })).stdout.strip
         end
         sha_hash?(result) ? result : nil
       end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/git.rb
```diff
@@ -323,6 +319,12 @@ def git(*args)
         ["git", *args].compact.join(" ")
       end
 
+      def run_git_command(args, run_opts = {})
```
lamont-granquist: so here's where you need to do `git_run_command(*args, **opts)` and then you can remove the arrays and hashes from the callers

lib/chef/provider/git.rb
```diff
@@ -282,8 +279,7 @@ def rev_search_pattern
       end
 
       def git_ls_remote(rev_pattern)
-        command = git(%Q{ls-remote "#{@new_resource.repository}" "#{rev_pattern}"})
-        shell_out!(command, run_options).stdout
+        (run_git_command ["ls-remote", "\"#{@new_resource.repository}\"", "\"#{rev_pattern}\""]).stdout
```
lamont-granquist: should be `run_git_command(...).stdout`

lib/chef/provider/git.rb
```diff
@@ -105,7 +105,7 @@ def action_sync
       end
 
       def git_minor_version
-        @git_minor_version ||= Gem::Version.new(shell_out!("git --version", run_options).stdout.split.last)
+        @git_minor_version ||= Gem::Version.new((run_git_command ["--version"]).stdout.split.last)
```
lamont-granquist: `run_git_command("--version").stdout.split.last`

lib/chef/provider/git.rb
```diff
@@ -323,6 +319,12 @@ def git(*args)
         ["git", *args].compact.join(" ")
       end
 
+      def run_git_command(args, run_opts = {})
+        git_command = git(args)
```
lamont-granquist: I'd actually suggest pulling the args mangling directly into here and then renaming `git_run_command` to just `git` -- the two methods don't really add a lot and the verbose naming could be dropped and `git("ls-remote", ...)` is pretty readable and understandable.

joshburt: That's an awesome suggestion!  I'll get this updated. :)

lib/chef/provider/git.rb
joshburt: Should we go one step further here and write it like this?

`@git_minor_version ||= Gem::Version.new git("--version").stdout.split.last`

lamont-granquist: we typically leave parens around method calls to make it clearer (like `#new()` and `#git()`).  i think there's a few exceptions where its typical ruby style to leave them off and its unambiguous...

so:

```
@git_minor_version ||= Gem::Version.new( git("--version").stdout.split.last )
 ```

i like leaving the space around the slightly complicated argument so that visually its clear and the expression can 'breathe' a little more--which is why the cop that removes those spaces is disabled.

Comment (lamont-granquist) [inline lib/chef/provider/git.rb]: and we obviously often leave parens off when there's no argument -- so we don't use `.stdout.split().last()` or whatever...

dsl methods are another good case where we obviously don't bother with parens:

```ruby
file("/tmp/whatever") do
  owner("root")
  group("root")
  mode("0644")
end
```

nobody wants to type that...

joshburt:
Looks like something inside appveyor broke - kicking the checks off again.

joshburt:
Second pass is complete.  When you get some time can you review style and functional changes?  Thank you sir! :)

lib/chef/provider/git.rb
```diff
@@ -319,8 +315,10 @@ def cwd
         @new_resource.destination
       end
 
-      def git(*args)
-        ["git", *args].compact.join(" ")
+      def git(args, run_opts = {})
```
lamont-granquist: you should still be able to use `git(*args, **opts)` here

lib/chef/provider/git.rb
```diff
@@ -157,8 +157,8 @@ def checkout
 
         converge_by("checkout ref #{sha_ref} branch #{@new_resource.revision}") do
           # checkout into a local branch rather than a detached HEAD
-          shell_out!("git branch -f #{@new_resource.checkout_branch} #{sha_ref}", run_options(:cwd => @new_resource.destination))
-          shell_out!("git checkout #{@new_resource.checkout_branch}", run_options(:cwd => @new_resource.destination))
+          git(["branch", "-f", @new_resource.checkout_branch, sha_ref], { cwd: cwd })
+          git(["checkout", @new_resource.checkout_branch], { cwd: cwd })
```
lamont-granquist: ...and that'll let you write this as `git("checkout", @new_resource.checkout_branch, cwd: cwd)` along with removing the rest of the array and hash decoration

lib/chef/provider/git.rb
```diff
@@ -167,12 +167,10 @@ def enable_submodules
         if @new_resource.enable_submodules
           converge_by("enable git submodules for #{@new_resource}") do
             Chef::Log.info "#{@new_resource} synchronizing git submodules"
-            command = "git submodule sync"
-            shell_out!(command, run_options(:cwd => @new_resource.destination))
+            git(%w{submodule sync}, { cwd: cwd })
```
lamont-granquist: `git(submodule, sync, cwd: cwd)`

lib/chef/provider/git.rb
joshburt: I have this updated (and it does look much cleaner).  I'll be pushing an update shortly.

lib/chef/provider/git.rb
joshburt: I have this updated (and it look much cleaner).  I'll be pushing an update shortly.

joshburt:
Looks like AppVeyor died again, will submit again to re-trigger.  Thanks!

lamont-granquist:
bundler is broken on windows

lamont-granquist:
:+1:

thommay:
👍

thommay:
Thanks for your patience here!

lamont-granquist:
@joshburt you want to rip rubocop/lcgstyle across the file and send that one in next?

PR #4879, opened 2016-04-29
remove tasks/bin/run_chef_tests; remove commented chef_rewind test; document bundler args

lamont-granquist:
- the bundler args look entirely identical, it is very hard to detect that 'integration' is added to the --without args.
- chef_rewind is on the path to deprecation, we should focus on deprecating it, not fixing the test
- bin/run_chef_tests is removed

there's no additional code duplication with this (the run_chef_tests line gets replaced by the rake spec line everywhere) while the common code is moved back into the .travis.yml file.  IMO, this is good because then all the information is contained withing the .travis.yml file and there's no need for the additional context switch between .travis.yml to the run_chef_tests file when reading it -- removed an unnecessary layer of indirection

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,11 @@ before_install:
 
 bundler_args: --without changelog development docgen guard maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
 
+before_script:
+ - echo --color > .rspec
+ - echo -fp >> .rspec
+ - sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers;
+
 # do not run expensive spec tests on PRs, only on branches
 branches:
   only:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -55,11 +50,6 @@ matrix:
       TEST_GEM: chef-provisioning-aws
     script: tasks/bin/run_external_test $TEST_GEM rake spec
     rvm: 2.2
-# requires vagrant
```
coderanger: Intentional comment? (vs. removing it)

lamont-granquist: chef-rewind needs to be deprecated now, we shouldn't fix this.

coderanger: Works for me. Are there docs for the (way better) APIs you added?

lamont-granquist: yard docs for now.  since we just release 12.9.x it'll be a little while before 12.10.x ships.  i also want to compat_resource those APIs and port chef-rewind over to use them instead of it applying its own monkeypatches to the resource collection.   and we don't even have compat_resource for 12.9.x and i'm waiting for john to help out with getting the recursive notifications resource_collection patches merged in there before tackling the 12.10.x compat_resource version...   then i'd like to doc them and blog about using them, and maybe even yap a bit about them all at chefconf...

lamont-granquist:
@chef/client-core assuming travis + appveyor finally pass i think this is reviewable

jkeiser:
:+1:

PR #4886, opened 2016-05-02
fixes knife ssl check on windows

mwrock:
On windows when we `glob` the trusted certs directoty, we always return an empty list of files even if there are certs present because we don't properly delimit the path for `glob`

### DIFF
--- a/spec/unit/knife/ssl_check_spec.rb
+++ b/spec/unit/knife/ssl_check_spec.rb
@@ -114,6 +114,22 @@ E
       allow(ssl_check).to receive(:verify_cert_host).and_return(true)
     end
 
+    context "when the trusted certificates directory is not glob escaped", :windows_only do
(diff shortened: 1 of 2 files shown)


danielsdeleo:
👍

spec/unit/knife/ssl_check_spec.rb
```diff
@@ -114,6 +114,22 @@ def stdout
...
 
+    context "when the trusted certificates directory is not glob escaped", :windows_only do
+      let(:trusted_certs_dir) { File.join(CHEF_SPEC_DATA.tr("/", "\\"), "trusted_certs") }
+
+      before do
+        allow(ssl_check).to receive(:trusted_certificates).and_call_original
+        allow(store).to receive(:verify).with(certificate).and_return(true)
+      end
+
+      it "escpaes the trusted certificates directory" do
```
randomcamel: Typo.

PR #4887, opened 2016-05-03
Support Ruby Files in ChefFS

thommay:
Part of the work to move Chef Solo to local mode is supporting the care and feeding of ruby roles and environment files. This necessitates a fairly major amount of rework in ChefFS to support files with more than one file extension.


lamont-granquist:
:+1:

thommay:
@chef/client-core would appreciate initial review.

PR #4888, opened 2016-05-03
Make apt package provider use_multipackage_api aware

lamont-granquist:
general modernization of apt_package

- removing/purging virtual packages works now ( at least a virtual package with only one option -- which is all that we support installing)

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -558,7 +558,7 @@ class Chef
       # @param args [String] variable number of string arguments
       # @return [String] nicely concatenated string or empty string
       def a_to_s(*args)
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@chef/client-core i think this is good for today

thommay:
👍

lamont-granquist:
for future reference - this still does not fix the purge-vs-remove issue (need to pull the pattern out of the dpkg provider that i used to fix that)

lamont-granquist:
Cool, I do promise that I'll get to FreeBSD eventually, but RHEL is a bit of a disaster these days and takes precedence unfortunately...

PR #4889, opened 2016-05-03
Add inspec in Chef client's embedded gems

andy-dufour:
Currently the Chef audit cookbook tries to install inspec from rubygems.org.

Bundling the inspec gem in with chef-client significantly simplifies the inspec gem install for customers that reside behind a firewall.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ group(:omnibus_package) do
   gem "appbundler"
   gem "rb-readline"
   gem "nokogiri"
+  gem "inspec"
 end
 group(:omnibus_package, :pry) do
   gem "pry"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -157,7 +157,11 @@ GEM
       thread_safe (~> 0.3, >= 0.3.1)
     diff-lcs (1.2.5)
     docile (1.1.5)
+    docker-api (1.26.2)
+      excon (>= 0.38.0)
+      json
     erubis (2.7.0)
+    excon (0.49.0)
     faraday (0.9.2)
       multipart-post (>= 1.2, < 3)
     fauxhai (3.4.0)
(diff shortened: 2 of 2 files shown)


coderanger:
Why not vendor it into the cookbook?

coderanger:
Or better, just publish them as cookbooks.

andy-dufour:
I'm not a fan of vendoring gems into cookbooks, especially because once the Audit cookbook is feature complete it likely won't be released often. At least chef-client has a steady release cadence which can keep customers somewhat up to date.

In reality people should be running gem mirrors behind their firewall to take care of this use-case but it's simply not happening or takes a long time to get the required resources setup in some Enterprises.

In the end, we're trying to accomplish a similar thing that we did with Chef audit mode here.

@chris-rock @arlimus for further comment

coderanger:
I guess this just feels like solving the problem at the wrong level given the relative immaturity of InSpec as a project. It should be seeing relatively rapid releases still, which means bundling it with Chef itself is probably the wrong move.

coderanger:
My vote would be to run InSpec and Train through Halite and release them as cookbooks so people can "mirror" them to a Chef Server.

chris-rock:
At first of all I'd like to thank @andy-dufour for the effort to put this up. Its great to see it coming into chef-client. @coderanger thank you very much for your honest feedback. I agree, that InSpec has rapid releases and they are shorter than chef-client.

In general, I would like to decouple the discussion of InSpec in chef-client from the audit-cookbook. The cookbook already works without InSpec in chef-client. It has some drawbacks and we are going to add a vendored gems anyway for current chef-client releases. Halite looks very promising to help us in that respect.

I still think, that InSpec has a valid place in chef-client. Going forward, I think we should delay the integration until InSpec gets 1.0. In addition, we should start thinking about the evolution of the audit-mode that uses InSpec instead of ServerSpec.

@coderanger What do you think it?

coderanger:
Yep, that sounds about right. Once InSpec's and Train's release cadence is closer to that of Chef it will make more sense to ship it as part of the Chef installers. I defer to others as to if we should do that via the Gemfile or the Omnibus config but either way, same effect :) Then after a few releases of that, we can back out the vendoring in the cookbooks.

thommay:
So I think there's general agreement that we shouldn't do this _yet_. I'm gonna close this PR; please feel free to create a new one once InSpec is 1.0 and the pace has slowed.

PR #4890, opened 2016-05-04
Fix four issues with FreeBSD ports installation

sekrett:
1. Fix `package_name`. If port name is different from package name Chef will fail to detect that port was already installed, it will compile again and fail to install it.
2. According to documentation a provider may be specified in resource so it should not be overridden by Chef.

### DIFF
--- a/lib/chef/provider/package/freebsd/port.rb
+++ b/lib/chef/provider/package/freebsd/port.rb
@@ -33,13 +33,22 @@ class Chef
             shell_out_with_timeout!("make deinstall", :timeout => 300, :env => nil, :cwd => port_dir).status
           end
 
+          # The name of the package (without the version number) as understood by pkg_add and pkg_info.
+          def package_name
+            if makefile_variable_value("PKGNAME", port_dir) =~ /^(.+)-[^-]+$/
+              $1
+            else
+              raise Chef::Exceptions::Package, "Unexpected form for PKGNAME variable in #{port_dir}/Makefile"
+            end
+          end
+
           def current_installed_version
             pkg_info = if @new_resource.supports_pkgng?
-                         shell_out_with_timeout!("pkg info \"#{@new_resource.package_name}\"", :env => nil, :returns => [0, 70])
+                         shell_out_with_timeout!("pkg info \"#{@current_resource.package_name}\"", :env => nil, :returns => [0, 70])
                        else
-                         shell_out_with_timeout!("pkg_info -E \"#{@new_resource.package_name}*\"", :env => nil, :returns => [0, 1])
+                         shell_out_with_timeout!("pkg_info -E \"#{@current_resource.package_name}*\"", :env => nil, :returns => [0, 1])
                        end
-            pkg_info.stdout[/^#{Regexp.escape(@new_resource.package_name)}-(.+)/, 1]
+            pkg_info.stdout[/^#{Regexp.escape(@current_resource.package_name)}-(.+)/, 1]
           end
 
           def candidate_version
--- a/lib/chef/provider/package/freebsd/port.rb
+++ b/lib/chef/provider/package/freebsd/port.rb
@@ -33,13 +33,8 @@ class Chef
             shell_out_with_timeout!("make deinstall", :timeout => 300, :env => nil, :cwd => port_dir).status
           end
 
-          # The name of the package (without the version number) as understood by pkg_add and pkg_info.
           def package_name
-            if makefile_variable_value("PKGNAME", port_dir) =~ /^(.+)-[^-]+$/
-              $1
-            else
-              raise Chef::Exceptions::Package, "Unexpected form for PKGNAME variable in #{port_dir}/Makefile"
-            end
+            get_package_name_from_ports
           end
 
           def current_installed_version
--- a/lib/chef/resource/freebsd_package.rb
+++ b/lib/chef/resource/freebsd_package.rb
@@ -49,13 +49,13 @@ class Chef
       end
 
       def assign_provider
-        @provider = if source.to_s =~ /^ports$/i
-                      Chef::Provider::Package::Freebsd::Port
-                    elsif supports_pkgng?
-                      Chef::Provider::Package::Freebsd::Pkgng
-                    else
-                      Chef::Provider::Package::Freebsd::Pkg
-                    end
+        @provider ||= if source.to_s =~ /^ports$/i
+                        Chef::Provider::Package::Freebsd::Port
+                      elsif supports_pkgng?
+                        Chef::Provider::Package::Freebsd::Pkgng
+                      else
+                        Chef::Provider::Package::Freebsd::Pkg
+                      end
       end
     end
   end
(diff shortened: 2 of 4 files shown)


lamont-granquist:
so this should all be done with provides lines on the provider.

the provider should own when it should run.  the new_resource.provider can still always be used to override that.  the code in the resource should be broken out into lines on the individual providers.  the code deep inside install_package is also wrong and should get moved into correct logic on the provider.

this was a use case that was part of the whole motivation for the 12.0 provider resolver feature, but i think i got distracted by all the 12.5 resources stuff and never got around to fixing it right.

sekrett:
Sorry, but I am new to Chef and don't understand your best practices, and of course I don't understand how I should rewrite this. What is provider lines? What is resolver feature? Do you have documentation explaining all this or one should read the source (which sometimes may be bad example of course) to figure out himself the right way to write recipes? I even don't know how to test all this locally. I had to wait Travis finishes all test, I rebased several times, it takes so much time, very difficult.

BTW, I found another issue which I have no idea how to solve. The package name may be different, for example, if I install www/mod_rpaf, pkg will report that package mod_rpaf is not installed, but it is being registered as ap22-mod_rpaf and it IS installed. Maybe inside Makefile it is possible to find the real package name.

sekrett:
I have an idea, a much better condition would be `unless @current_resource.version`. Anyway, can you accept second patch and I will rebase after that?

sekrett:
There is another problem. freebsd_package fails if package name is an array and there are many cookbooks which supply arrays and they do not work, very frustrating. If I take apt provider as example, is it good? apt supports arrays.

sekrett:
@lamont-granquist, please take a look now, I did it better, I just fill the current_version and some magic happens, Chef realises the port is already installed.

sekrett:
Why simple system call `whereis -s zsh` fails??? It did not fail previous time.

sekrett:
I think it was a very bad idea to assign provider based on `source`, it does not work with multiple package names. Ports collection is not a source but an installation method, so one should switch provider, not source.

sekrett:
I took a look at pkg provider, it uses `source` property, so it will not support array arguments, in other words I don't know how to patch pkg providers, if it is possible. Anyway any intelligent man have already switched to pkgng on FreeBSD 9, so it is not so important fix that.

I think I will stop here. Maybe you want to refactor, but this code works, really works and fixes 4 issues, very important issues. Package management is essential part of Chef, it is very frustrating if it does not work, people may decide not to use Chef (on FreeBSD) at all because of this. Please give some priority to this.

lib/chef/resource/freebsd_package.rb
```diff
@@ -49,13 +49,11 @@ def ships_with_pkgng?
...
-                    elsif supports_pkgng?
-                      Chef::Provider::Package::Freebsd::Pkgng
-                    else
-                      Chef::Provider::Package::Freebsd::Pkg
-                    end
+        @provider ||= if supports_pkgng?
+                        Chef::Provider::Package::Freebsd::Pkgng
+                      else
+                        Chef::Provider::Package::Freebsd::Pkg
+                      end
```
lamont-granquist: this is still wrong, you should be using provides with a block on the pkgng and pkg providers instead.  also where did port go?

lib/chef/provider/package/freebsd/port.rb
```diff
@@ -26,32 +26,44 @@ class Port < Base
...
-          def current_installed_version
-            pkg_info = if @new_resource.supports_pkgng?
-                         shell_out_with_timeout!("pkg info \"#{@new_resource.package_name}\"", :env => nil, :returns => [0, 70])
-                       else
-                         shell_out_with_timeout!("pkg_info -E \"#{@new_resource.package_name}*\"", :env => nil, :returns => [0, 1])
-                       end
-            pkg_info.stdout[/^#{Regexp.escape(@new_resource.package_name)}-(.+)/, 1]
+          # The name of the package (without the version number) as understood by pkg_add and pkg_info.
+          def package_name
+            [@new_resource.package_name].flatten.map do |name|
```
lamont-granquist: multipackage supporting providers need to declare use_multipackage_api and then they'll always be passed arrays instead of scalars.  you need to also use some other conventions like always using package_name_array from the superclass.

i would suggest dropping the multipackage support out of this PR, its going to complicate it all and probably cause a lot of frustration.

lib/chef/provider/package/freebsd/pkgng.rb
```diff
@@ -25,27 +25,32 @@ module Freebsd
...
+              shell_out_with_timeout!("pkg install -y#{expand_options(@new_resource.options)} #{[name].flatten.join(' ')}", :env => { "LC_ALL" => nil }).status
             end
           end
 
           def remove_package(name, version)
             options = @new_resource.options && @new_resource.options.sub(repo_regex, "")
             options && !options.empty? || options = nil
-            shell_out_with_timeout!("pkg delete -y#{expand_options(options)} #{name}#{version ? '-' + version : ''}", :env => nil).status
+            name_array = [name].flatten
+            version_array = [version].flatten
```
lamont-granquist: same comments here about this being the wrong way to implement multipackage

smurawski:
@sekrett In regards to the failed tests finding "zsh", the output that is expected is not matching, so it is likely that you are missing a mock or stub somewhere.  The failed tests are unit tests, so they shouldn't actually be inspecting the system the tests are running, they are just to validate the logic of your workflow.

lib/chef/resource/freebsd_package.rb
sekrett: I don't know that provides syntax, sorry. I also asked to give some example or link to documentation, if you want me to rewrite something.

I removed Port because it was failing with multiple packages, an exception is being raised that you cannot specify source with multiple packages. If you suggest to drop multiple support from this pr, this patch becomes nonsense and I will revert it. The idea was to do:
<pre>package 'bash', 'zsh' do
  provider Chef::Provider::Package::Freebsd::Port
end</pre>
instead of specifying `source`.

sekrett:
@lamont-granquist, yesterday I upgraded my chef-client and it gave me warning that multiple package names are deprecated and will not work in Chef 13. I agree this is intelligent, it is easy to write: `arr.each { |pkg| package pkg }` instead of complicating Chef internals. So can you confirm Chef decided to remove that? If yes, I will remove issues 2,3,4 from this PR and rebase. Many cookbooks will need to be rewritten, but for me it is correct.

@smurawski, thanks for your comment. I understand that unit tests need mocking but I don't understand why my first patches were all green and later I had fails. I did not change anything regarding `shell_out`.

lib/chef/provider/package/freebsd/port.rb
sekrett: BTW, I was looking at apt_package implementation and I wrote the similar way. I Did not know it is wrong. Anyway, without arrays it is much better.

sekrett:
There is an opinion that multiple packages are faster: https://www.chef.io/blog/2015/03/03/phil-dibowitz-chef-client-12-1-0-multipackage-installs/

lamont-granquist:
package name arrays are not deprecated, i'm slowly working through converting them all over to arrays.

the apt_package resource was an incredibly bad example before this work i merged 2 weeks ago:

https://github.com/chef/chef/pull/4888

sekrett:
Wow, you are doing a great job of rewriting these classes. I should look at master version, not release, my fault.

I am convinced multiple package support you will do better than me. I rebased and left the minimum I need to make my cookbooks work. All explanations I have written before.

lamont-granquist:
Cool, I do promise that I'll get to FreeBSD eventually, but RHEL is a bit of a disaster these days and takes precedence unfortunately...

lamont-granquist:
closed in favor of just tracking the issue in #5158

PR #4893, opened 2016-05-04
DNF Provider

lamont-granquist:
BUGS:

- haven't checked downgrades

- virtual packages are broken (they get called with the version of the real package and it dies)

- `dnf provides` reorders what it returns randomly, can't just pick the bottom one on the list for the best match to populate candidate_version
    - the way dnf selects a package when there are multiples is simply the highest version -- which is why mariadb has an epoch of '1' so it is 'later' than every community-mysql version
    - f23 does not have the python rpm libraries installed by default

- no arch support


lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,111 @@
...
+
+          current_resource.version(get_current_versions)
+
+          current_resource
+        end
+
+        def candidate_version
+          @candidate_version ||= begin
+            package_name_array.map do |package_name|
+              provides = dnf("provides", package_name)
```
jaymzh: I'm pretty sure we're going to want something like YumCache here (in fact, we can just use YumCache, the metadata for DNF is identical... they still use the same `createrepo`

lamont-granquist: I don't think we do.  With making candidate_version array lazy, this is never hit if its idempotent and only hit if we're installing packages (in which case we have to hit yum-dump.pl again anyway).

If this works I am probably ripping the yum cache out of the yum_package provider as well...

lamont-granquist:
hrmmm, now I'm DDoS'ing travis and appveyor though....

lamont-granquist:
oops, see #4894

Issue #4266, opened 2015-12-10
package resource should raise when package_name.length != version.length != arch.length


jsmartt:
This snippet fails on at least the standard centos-7.1 boxes I use with kitchen:

``` ruby
package 'glibc' do
  arch ['x86_64', 'i686']
end
```

I get the following error:

```
Converging 1 resources
Recipe: test2::yum

    * No candidate version available for glibc
    ================================================================================
    Error executing action `install` on resource 'yum_package[glibc]'
    ================================================================================

    Chef::Exceptions::Package
    -------------------------
    No candidate version available for glibc

    Resource Declaration:
    ---------------------
    # In /tmp/kitchen/cache/cookbooks/test2/recipes/yum.rb

      7: package 'glibc' do
      8:   arch ['x86_64', 'i686']
      9: end
     10:

    Compiled Resource:
    ------------------
    # Declared in /tmp/kitchen/cache/cookbooks/test2/recipes/yum.rb:7:in `from_file'

    yum_package("glibc") do
      action [:install]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      package_name "glibc"
      flush_cache {:before=>false, :after=>false}
      declared_type :package
      cookbook_name "test2"
      recipe_name "yum"
      arch ["x86_64", "i686"]
    end


Running handlers:
[2015-12-09T23:57:16+00:00] ERROR: Running exception handlers
Running handlers complete
[2015-12-09T23:57:16+00:00] ERROR: Exception handlers complete
Chef Client failed. 0 resources updated in 03 seconds
[2015-12-09T23:57:16+00:00] FATAL: Stacktrace dumped to /tmp/kitchen/cache/chef-stacktrace.out
[2015-12-09T23:57:16+00:00] ERROR: yum_package[glibc] (test2::yum line 7) had an error: Chef::Exceptions::Package: No candidate version available for glibc
```

however, the following works:

``` ruby
package 'glibc' do
  arch 'x86_64'
end

package 'glibc' do
  arch 'i686'
end
```

So I know there is a package available. It doesn't seem to handle an array of architectures as shown in the documentation [here](https://docs.chef.io/resource_yum.html).

lamont-granquist:
i'm leaning towards either that being the correct syntax or it should use (i think) `package [ 'glibc.i686', 'glibc.x86_64' ]` to leverage the arch-in-the-name and multipackage functionality that is already in there.

[...]

Issue #4123, opened 2015-10-30
yum downgrade install method not respected for local package


Sauraus:
If you specify the `source` attribute on a yum package chef-client will default to `localinstall` as the installation method because it's hardcoded here: https://github.com/chef/chef/blob/master/lib/chef/provider/package/yum.rb#L1255-L1260

lamont-granquist:
What should the correct behavior be?

Sauraus:
IMHO it should allow downgrade with a local package, why should this only be supported for remote package installs?

lamont-granquist:
yeah, i haven't admin'd a CentOS box in over 4 years, just need to understand what commands to issue, that code is very useful...

lamont-granquist:
#6540 fixes this now

Issue #4103, opened 2015-10-24
package resource should honor version without release


someara:
This is non-intuitive.

Yum from the command line works with

```
 yum install docker-engine-1.7.1
```

Chef forces you to investigate an enabled package repository and then specify an exact release

``` ruby
package 'docker-engine' do
  version '1.7.1-1.el6'
end
```

^ The problem with that, is that it might not be the latest revision available to a repository, which is probably not what you want on RHELish systems.

Googling 'Did you specify both version and release?' yields chaos and confusion.
Example of chaos and confusion - https://github.com/someara/chef-docker/issues/93

pkang:
+1.

Would love it if we could get this enhancement. Like @someara mentioned, it's completely unintuitive given yum command line works without having to specify release portion of package.

phspagiari:
:+1:

joshuawscott:
👍

hatt:
How is this not standard behaviour? The underlying package management supports it and every other "major" configuration management system has supported this for a long time.

someara:
Lack of this feature basically makes the version property useless if you're not managing your own repository mirrors.

Upstream repositories (for some reason) often yank releases when new ones come out.
This is causing all sorts of chaos in the mysql cookbook

https://github.com/chef-cookbooks/mysql/issues/443

someara:
This applies to apt_package as well

someara:
https://github.com/chef-cookbooks/docker/issues/781

lamont-granquist:
#6540 fixes this

PR #4894, opened 2016-05-05
DNF Provider PR #2

### LINKED ISSUE #1640: yum_package flush_cache shouldn't need to be an array
The documentation doesn't make it clear that flush_cache needs to be an array http://docs.opscode.com/resource_yum.html.

Works:

```
flush_cache [:before]
```

Doesn't:

```
flush_cache :before
```

```
NoMethodError
-------------
undefined method `any?' for :before:Symbol
```

Surely this doesn't need to be an array of one?

### LINKED ISSUE #3201: Fedora 22 defaults to dnf for package management
Fedora 22 is [switching to a new package manager, dnf](https://fedoraproject.org/wiki/Changes/ReplaceYumWithDNF). They provide a legacy package (installed by default) which will provide an executable named `yum`, but [there are a number of differences](http://dnf.readthedocs.org/en/latest/cli_vs_yum.html).

As far as I can tell, most of the differences won't affect the yum provider, but I wanted to open an issue on this to increase awareness. I'm planning on writing a DNF provider in the near future, but if someone else gets to it first that is fine with me.

lamont-granquist:
- should fully support globbing, arches, versions and all other syntax in the name of the package:

```ruby
dnf_package "tcpd*-4*"
dnf_package "tcpd\*.x86_64"
```

- supports provides syntax in the name:

```ruby
dnf_package "tcpdump >= 4.7"
```

- supports version and arch properties, with globs in the version

```ruby
dnf_package "tcpdump" do
  version "4.7*"
  arch "x86_64"
end

dnf_package "tcpdump" do
  version "4.7.4"
end

dnf_package "tcpdump" do
  version "4.7.4-4.fc24"
end

dnf_package "tcpdump" do
  version "0:4.7.4-4.fc24"
end
```

NOTE: release globs are not supported ("4.7.4-4*") but that is because it doesn't work on the command line with DNF either

- supports source argument and paths to rpms:
   (NOTE: multipackage with source arguments is busted, it is also busted with yum, the stuff i did for dpkg needs to be rethought.)

```ruby
dnf_package "/a/path/to/tcpdump-4.7.10-1.fc23.x86_64.rpm"
dnf_package "tcpdump" do
  source  "/a/path/to/tcpdump-4.7.10-1.fc23.x86_64.rpm"
end
```

- NOTE: will not support groups

closes #3201
closes #1640


lamont-granquist:
As far as idempotent performance goes, this is much faster than the yum_provider even though it has no yum cache.

Wrapping the chef-apply Chef::Runner call (to avoid timing ohai and everything else):

```
# ./apply.rb
  * dnf_package[strace] action install (up to date)
  * dnf_package[lsof] action install (up to date)
  * dnf_package[tcpdump] action install (up to date)
  * dnf_package[git] action install (up to date)
  * dnf_package[dmidecode] action install (up to date)
  * dnf_package[ltrace] action install (up to date)
  * dnf_package[bc] action install (up to date)
  * dnf_package[zsh] action install (up to date)
  * dnf_package[vim-minimal] action install (up to date)
  * dnf_package[systemtap] action install (up to date)
  * dnf_package[telnet] action install (up to date)
  * dnf_package[mtr] action install (up to date)
  * dnf_package[subversion] action install (up to date)
  * dnf_package[nmap] action install (up to date)
  * dnf_package[gdb] action install (up to date)
  * dnf_package[iptraf-ng] action install (up to date)
  * dnf_package[tmux] action install (up to date)
  * dnf_package[zlib-devel] action install (up to date)
  * dnf_package[wget] action install (up to date)
  * dnf_package[postfix] action install (up to date)
  * dnf_package[sysstat] action install (up to date)
  0.020000   0.020000   0.180000 (  0.591286)
```

for 12.9.38:

```ruby
# ./apply.rb
  * yum_package[strace] action install (up to date)
  * yum_package[lsof] action install (up to date)
  * yum_package[tcpdump] action install (up to date)
  * yum_package[git] action install (up to date)
  * yum_package[dmidecode] action install (up to date)
  * yum_package[ltrace] action install (up to date)
  * yum_package[bc] action install (up to date)
  * yum_package[zsh] action install (up to date)
  * yum_package[vim-minimal] action install (up to date)
  * yum_package[systemtap] action install (up to date)
  * yum_package[telnet] action install (up to date)
  * yum_package[mtr] action install (up to date)
  * yum_package[subversion] action install (up to date)
  * yum_package[nmap] action install (up to date)
  * yum_package[gdb] action install (up to date)
  * yum_package[iptraf-ng] action install (up to date)
  * yum_package[tmux] action install (up to date)
  * yum_package[zlib-devel] action install (up to date)
  * yum_package[wget] action install (up to date)
  * yum_package[postfix] action install (up to date)
  * yum_package[sysstat] action install (up to date)
  0.280000   0.010000   0.820000 (  0.858275)
```

shelling out to `rpm -q` is idiotically fast:

```
# time rpm -q --whatprovides mysql
mariadb-10.0.23-1.fc23.x86_64
rpm -q --whatprovides mysql  0.00s user 0.00s system 92% cpu 0.005 total
```

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,115 @@
...
+
+          current_resource
+        end
+
+        def candidate_version
+          @candidate_version ||=
+            begin
+              @real_name = {}
+              package_name_array.map do |package_name|
+                provides = dnf("repoquery -q --whatprovides", package_name)
```
jaymzh: (continuing comment from old PR)

Hmmm I imagine that's going to be expensive and cause way, way more IO.

jaymzh: Your perf numbers are interesting. A few things:

* You said we wont' hit this unless we need to install packages... how is that possible? How do you know if we need to install a package for `:upgrade` unless you do this? (99% of our packages are :upgrade ftr)?
* Doing this once per package will be much more IO which is what I"m worried about

Maybe I'm misunderstanding something.

lamont-granquist:
note that actually doing a dnf install is idiotically slow, because dnf is just slow, even from the command line:

yum:

```
# time for i in strace lsof tcpdump git dmidecode ltrace bc zsh vim-minimal systemtap telnet mtr subversion nmap gdb iptraf-ng tmux zlib-devel wget postfix sysstat; do yum -qy install $i; done
real    0m37.676s
user    0m15.834s
sys    0m3.654s
```

dnf:

```
# time for i in strace lsof tcpdump git dmidecode ltrace bc zsh vim-minimal systemtap telnet mtr subversion nmap gdb iptraf-ng tmux zlib-devel wget postfix sysstat; do dnf -qy install $i; done

real    1m45.886s
user    0m41.036s
sys    0m7.000s
```

so dnf install is inherently almost 3x slower than yum (i tried disabling ipv6 and using fastestmirror but neither helped much).

lib/chef/provider/package/dnf.rb
lamont-granquist: Okay `:upgrade` might be an issue...

jaymzh: Yeah as I said this is nearly everything for us...

lamont-granquist: sweet, yeah, okay this is terrible for you, 42 seconds to run vs 0.86 seconds to run with `:upgrade`...

lamont-granquist:
and while `dnf repoquery` sorts it output it uses lexicographic string sorting, not rpm version sorting:

```
z3-devel-0:4.4.0-3.fc23.i686
z3-devel-0:4.4.0-3.fc23.x86_64
z3-devel-0:4.4.1-2.fc23.i686
z3-devel-0:4.4.1-2.fc23.x86_64
```

```
zookeeper-0:3.4.6-14.fc23.i686
zookeeper-0:3.4.6-14.fc23.x86_64
zookeeper-0:3.4.6-7.fc23.i686
zookeeper-0:3.4.6-7.fc23.x86_64
```

jaymzh:
There is an RPMVersion class from the yum provider you can use to compare RPM versions sanely... pass that to sort and bob's your uncle. But cache it.

jaymzh:
(tag me in here when you get around to adding a DnfCache thing that'll let this work for :upgrade... I'm not really pay attention to the various changes here atm)

lamont-granquist:
yeah don't look at this, pay no attention to the man behind the curtain...  i'm trying out crazy ideas...

lamont-granquist:
With this version running `:upgrade` on already-up-to-date rpms runs through my list of 23 rpms in 1.16 seconds using one-resource-per-package with is competitive with 0.86seconds in centos7 using yum.  With one muiltipackage resource around those 23 rpms it takes 0.49 seconds.

jaymzh:
Is that your way of saying this ready for review? or should we still be ignoring you?

lamont-granquist:
I'm not done with it, but you might want to start paying attention.  I'm going to try to get it done next weekish.

lamont-granquist:
the flush_cache property API and allow_downgrade (testing at least) are in flight.

lamont-granquist:
Open yum provider bugs that might be relevant:

- [ ] #5254 (arch)
- [ ] #5228 (repos)
- [ ] #5163 (arch)
- [ ] #5112 (multipackage)
- [ ] #4266 (multi-arch)
- [x] #4123 (downgrades with localinstall)
- [ ] #4103 (version+release handling)

lamont-granquist:
@jaymzh you want to give d1a349b4e04998b4f5712973a6d8ffa47e091c45 a shot and see what breaks for you?

lamont-granquist:
@jaymzh pingles can you break this for me?

lamont-granquist:
guess i have some puts's to fix up and a FIXME in define_resource_requirements to still address....

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+            end
+            hash["epoch"] = epoch unless epoch.nil?
+            hash["release"] = release unless release.nil?
+            hash["version"] = version
+          end
+
+          # @returns Array<Version>
+          def query(action, provides, version = nil, arch = nil)
+            with_helper do
+              hash = { "action" => action }
```
stevendanna: Minor Style: Might be nice to put 113-117 into a `make_message` function that would return the json. Similarly, 121-122 could go into a `parse_response` method.  So this function would then read:

```
with_helper do
   message = make_message(action, provides, version, arch)
   stdin.syswrite message
   response = stdout.sysread(4096)
   parse_response(response)
end
```

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+            @stdin, @stdout, @stderr, @wait_thr = Open3.popen3(DNF_COMMAND)
+          end
+
+          def reap
+            unless wait_thr.nil?
+              Process.kill("KILL", wait_thr.pid) rescue nil
+              stdin.close unless stdin.nil?
+              stdout.close unless stdout.nil?
+              stderr.close unless stderr.nil?
+              wait_thr.value
```
lamont-granquist: that is correct

spec/unit/resource/dnf_package_spec.rb
```diff
@@ -0,0 +1,100 @@
+#
+# Author:: AJ Christensen (<aj@chef.io>)
```
lamont-granquist: lol

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+          attr_accessor :stdout
+          attr_accessor :stderr
+          attr_accessor :wait_thr
+
+          DNF_HELPER = ::File.expand_path(::File.join(::File.dirname(__FILE__), "dnf_helper.py")).freeze
+          DNF_COMMAND = "#{which("python3")} #{DNF_HELPER}"
+
+          def start
+            ENV["PYTHONUNBUFFERED"] = "1"
+            @stdin, @stdout, @stderr, @wait_thr = Open3.popen3(DNF_COMMAND)
```
lamont-granquist: The other thing about having the python helper using stdin/stdout/stderr is that it works trivially from the command line and i used that a lot to debug it, and you either lose that flexibility or else that becomes a special debugging case, and it all feels a bit like adding complexity if it needs to be able work in this fashion from the command line anyway to debug.

stevendanna: Good point on the "working from the command line" bit.  I agree, let's keep it using stdin and stdout and have a TODO for collecting and reporting stderr.

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+        allow_nils
+
+        class Version
+          attr_accessor :name
+          attr_accessor :version
+          attr_accessor :arch
+
+          def initialize(name, version, arch)
+            @name = name
+            @version = ( version == "nil" ) ? nil : version
```
lamont-granquist: fixing

lib/chef/provider/package/dnf.rb
lamont-granquist: noted in comment

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+            with_helper do
+              hash = { "action" => action }
+              hash["provides"] = provides
+              add_version(hash, version) unless version.nil?
+              hash["arch" ] = arch unless arch.nil?
+              json = FFI_Yajl::Encoder.encode(hash)
+              puts json
+              stdin.syswrite json + "\n"
+              output = stdout.sysread(4096)
+              puts output
```
lamont-granquist: fixing

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+          end
+
+          def restart
+            reap
+            start
+          end
+
+          def with_helper
+            max_retries ||= 5
+            Timeout.timeout(60) do
```
lamont-granquist: bumped to 600s

spec/unit/resource/dnf_package_spec.rb
lamont-granquist: fixing

lib/chef/provider/package/dnf_helper.py
```diff
@@ -0,0 +1,87 @@
...
+
+signal.signal(signal.SIGINT, exit_handler)
+signal.signal(signal.SIGHUP, exit_handler)
+signal.signal(signal.SIGPIPE, exit_handler)
+signal.signal(signal.SIGCHLD, exit_handler)
+
+while 1:
+    ppid = os.getppid()
+    if ppid == 1:
+        sys.exit(0)
```
lamont-granquist: yeah i actually found i was getting orphaned subprocesses due to bugs when i was hammering on it with tests, and this code cleaned them all up (plus i went and fixed the bugs)

the dnf helper kind of tries to be very 'brittle' and die if anything goes wrong -- hence this code, plus the exit handler, etc.  trying very hard to avoid littering customer's process tables...

lamont-granquist:
so there's a couple of deliberately-kept FIXMEs here:

- cache flushing doesn't currently work and we just kill the process because i tried to implement that and created massive memory leaks in the python process.   performance is still acceptable.  i also have plans to implement installing and removing packages using the dnf helper itself (which should update the in-memory cache of the dnf helper itself thereby mitigating the need to flush caches entirely) which may render this problem mostly moot for the flush-installed use case that occurs after we install or remove something -- then we can use killing the process as the appropriate way to flush the cache if the user modifies the yum repos and needs the process to reload them completely.

- there are some fairly hideous use cases around what to do if you're asked to install e.g. "chef_rpm" on a 64-bit system and you have the 32-bit version of the package installed.  i think the right-est answer is that customers need to fully specify their intent.  i think we need to see if we get any bug reports around this to determine what customers would actually expect the right thing is to do here.  i more than halfway hope that nobody cares about this use case and its just excessive naval gazing at use cases that don't practically occur (there's nothing open on this for the existing yum provider).

lamont-granquist:
TODO:

- [ ] pondering length prefix and 4096 fixed length read buffering
- [ ] debug logging stderr back from the python helper so we can get stacktraces at least

lib/chef/provider/package/dnf.rb
```diff
@@ -0,0 +1,298 @@
...
+          def query(action, provides, version = nil, arch = nil)
+            with_helper do
+              hash = { "action" => action }
+              hash["provides"] = provides
+              add_version(hash, version) unless version.nil?
+              hash["arch" ] = arch unless arch.nil?
+              json = FFI_Yajl::Encoder.encode(hash)
+              puts json
+              stdin.syswrite json + "\n"
+              output = stdout.sysread(4096)
```
lamont-granquist: Yeah, I'm not too worried about overflowing the 4k buffer because we only ever send one request at a time and the name+version+arch on the new resource is unlikely to ever get that large.  On the way back we also only ever receive one package to install (all the filtering down to the one package nevra happens python-side -- even though the ruby code would handle multiple versions, it never actually receives them).

And I think I have the same feelings about this as the file descriptor which is that it makes it more difficult to play with the helper on the command line.

lamont-granquist:
i investigated a super-optimization to eliminate cache flushing entirely on installs and use the dnf-helper python script to do the installs (which patches its own cache so it doesn't have to regenerate caches at all) plus removes all the shellouts to dnf to install the packages and the speed boost was not impressive enough i think:

```
    with the patch:
      0.030000   0.060000  18.070000 ( 19.413037)

    without the patch:
      0.060000   0.130000  24.470000 ( 27.511077)
```

that's only a 25% bump in speed boost for additional added complexity and the patch had some bugs that still needed to be worked through:  0c13033821905019a8d05ef5c1bb47efb638d57f

and if that isn't worth it, then the speed boost from not killing the process to flush the cache probably won't be worth it either, since that's also baked into there.

DNF is just slow.

jaymzh:
FWIW, I just got 7.3 on a box in our environment (which has dnf)... I was going to try to carve out time to test this work this week or right after xmas.

thommay:
I think we should merge this now and get a full month of baking before the next release.

lamont-granquist:
so, the only DNF i can find for centos 7.3 is the dnf-0.6 ancient sauce that is in epel.

jaymzh:
Yeah - I see the same thing. I guess they changed their mind. :( I'll chat with the RH guys in Brno in 2 weeks.

jaymzh:
Hey @lamont-granquist talked to the RPM folks... try this:

https://copr.fedorainfracloud.org/coprs/jsilhan/centos-dnf/

It got pushed out, but it's expected to land for 7.5 - Facebook is now planning to move to these builds sooner rather than later.

lamont-granquist:
coolness.  i installed it.  its broken.  looking at fixing...

PR #4896, opened 2016-05-05
Deprecate knife-supermarket gem

thommay:
Include all knife supermarket sub commands here, and allow all knife
cookbook site commands to have a configurable supermarket URL.

cc @chef/client-dev-tools @robbkidd

### DIFF
--- a/lib/chef/knife/cookbook_site_download.rb
+++ b/lib/chef/knife/cookbook_site_download.rb
@@ -37,6 +37,13 @@ class Chef
         :long => "--force",
         :description => "Force download deprecated version"
 
+      option :supermarket_site,
+        :short => "-m SUPERMARKET_SITE",
+        :long => "--supermarket-site SUPERMARKET_SITE",
+        :description => "Supermarket Site",
+        :default => "https://supermarket.chef.io",
+        :proc => Proc.new { |supermarket| Chef::Config[:knife][:supermarket_site] = supermarket }
+
       def run
         if current_cookbook_deprecated?
           message = "DEPRECATION: This cookbook has been deprecated. "
@@ -59,7 +66,7 @@ class Chef
       private
 
       def cookbooks_api_url
-        "https://supermarket.chef.io/api/v1/cookbooks"
+        "#{config[:supermarket_site]}/api/v1/cookbooks"
       end
 
       def current_cookbook_data
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -59,6 +59,13 @@ class Chef
         :boolean => true,
         :default => false
 
+      option :supermarket_site,
+        :short => "-m SUPERMARKET_SITE",
+        :long => "--supermarket-site SUPERMARKET_SITE",
+        :description => "Supermarket Site",
+        :default => "https://supermarket.chef.io",
+        :proc => Proc.new { |supermarket| Chef::Config[:knife][:supermarket_site] = supermarket }
+
       attr_reader :cookbook_name
       attr_reader :vendor_path
 
@@ -134,6 +141,7 @@ class Chef
       def download_cookbook_to(download_path)
         downloader = Chef::Knife::CookbookSiteDownload.new
         downloader.config[:file] = download_path
+        downloader.config[:supermarket_site] = config[:supermarket_site]
         downloader.name_args = name_args
         downloader.run
         downloader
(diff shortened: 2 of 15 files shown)


lamont-granquist:
👍

robbkidd:
:+1: `:hugefan:`

freimer:
So, let me make sure I understand this PR.  The suggestion is to get rid of knife supermarket.  It is really convenient, for me at least and I assume others, that have their own supermarket site to be able to use knife cookbook site to refer to the public Supermarket site, and knife supermarket (share mostly) to refer to their own private Supermarket site, configured in config.rb.  Are we going to have some flexibility to be able to use a simple command like this in order to refer to the public or private Supermarket site, rather than having to override with a typed-in URL every time we want to use the opposite of what is configured in config.rb?   I'm not a super Ruby expert yet, but I don't see any new config options, or command line options to easily switch between the two like we have with knife cookbook site and knife supermarket.  Unless this functionality is included and I'm just missing it, I'm definitely 👎

polliard-jmfe:
What is the problem that this PR is trying to solve?

thommay:
@freimer yes, the exact same configuration options that are available in `knife supermarket` will be preserved in this merging of the code. 
@polliard-jmfe not having two codebases that are basically identical in functionality. reducing our maintenance burden and user confusion. 

To be clear, this PR means that both `knife cookbook site ...` and `knife supermarket ...` will behave exactly the same as `knife supermarket ...` does now.

freimer:
So how do we download cookbooks with knife from the public server without having to type in the public server URL every time?

robbkidd:
Proposal:

To consolidate the logic of interacting with the Supermarket API and to encourage workflows to change over to using `knife supermarket`:

* move the implementation from the `cookbook_site_*` command classes to `supermarket_*` classes
* add `supermarket --public` CLI option that will target the hard-coded public site and ignore any `config[:supermarket_site]` coming in from a knife config _and_ ignore a URL given via `-m/--supermarket-site` CLI-option
    * `--public` ignoring a site given on the CLI seems a bit weird and we might find ourselves in situations where we ask "why did you specify these conflicting CLI options?" Maybe include a warning about the custom site being ignored when using `--public` and a custom `config[:supermarket_site]` is present?
* add `supermarket --private` CLI option that will target the supermarket-site in knife-config/CLI-option
* `supermarket` defaults to `supermarket --private` behavior
* `cookbook site` defaults to `supermarket --public` behavior

lamont-granquist:
So leaving this PR unmerged is the worst of all possible worlds.  I'd prefer to merge it and solve the problem of having two shitty half broken implementations that troll everyone, and break the very few people who adopted using `knife cookbook site` to refer to the public site and `knife supermarket` to refer to their internal supermarket.  I suspect if we merged this, it would provide a forcing function to add the config options in a new PR.  As it is this is just stalled and no progress is being made, and is sort of being held hostage due to an edge condition.

smurawski:
:+1:

thommay:
I'd semi forgotten about this. I think it's reasonable that people who really want different behaviour should just alias `knife supermarket -m https://foo` to something reasonable.

lamont-granquist:
if we can already solve that problem with a command alias to add the url, then i vote we mash merge.

👍

robbkidd:
This is exciting. Is the next step to publish a new version of the [knife-supermarket gem](https://rubygems.org/gems/knife-supermarket) with the deprecation notice (at least in the README, but possibly elsewhere) and a version constraint on the chef gem being `<=` the latest release version (prior to this merge).

charlesjohnson:
Now that this is in, would a PR that makes `knife cookbook site` an undocumented synonym for `knife supermarket` be accepted / a good idea? This seems like an unnecessary and confusing place for TIMTOWTDI.

Also I've always hated the wording of the `cookbook site` command.

lamont-granquist:
i dont think you can hide options from --help is that is what you mean

charlesjohnson:
I'm saying that 'cookbook site' just becomes an undocumented alias for 'supermarket'. If hiding it winds up more trouble than it's worth, then a documented alias.

But not two separate features, with two separate code paths, and two separate syntaxes. One feature that can be invoked with two names.

lamont-granquist:
that is exactly what this PR does...  impossible to hide it from help and make it undocumented though.

we can and should probably add language / undocument the "cookbook site" stuff in docs.chef.io

charlesjohnson:
Ugh. This is what I get for pre-coffee internet. I saw `supermarket_download.rb` and `cookbook_site_download.rb` existing, and got twisted around because I didn't look close enough.

Mea culpa.

lamont-granquist:
I feel your pain...  every day....

PR #4898, opened 2016-05-05
Update usage of @new_resource.destination to `cwd` within the git hwrp

joshburt:
We have access to a method within the git HWRP to handle resolution of the cwd (current working directory) to @new_resource.destination.

Would it make sense to make more use of this method and replace instances of @new_resource.destination with it?

Would enjoy some feedback on this. Thanks!


coderanger:
Is there a reason to not go the other way and remove the helper? Notably use of `@new_resource` is kind of gross and breaks all kinds of scoping stuff :)

joshburt:
We certainly could!  The result would be some other code that gets longer / less clear in a number of conditions:



**Examples**

From:
`	result = git("rev-parse", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip`
To:
`	result = git("rev-parse", "HEAD", cwd: @new_resource.destination, returns: [0, 128]).stdout.strip`

From:
`	FileUtils.rm_rf(::File.join(cwd, ".git"))`
To:
`	FileUtils.rm_rf(::File.join(@new_resource.destination, ".git"))`

smurawski:
:+1:

lamont-granquist:
i'd just suggest to use Forwardable and `def_delegator :@new_resource, :destination, :cwd`

joshburt:
The Forwardable / def_delegator was a nice suggestion!  I've updated the implementation accordingly.  Thanks everyone!

thommay:
:+1: LGTM

joshburt:
Looks like the travis-ci run had a few instances of `Timeout::Error`.  I'll kick this off again as it doesn't seem related to the code changes within this PR.  Thanks!

joshburt:
Second run went through just fine!

PR #4901, opened 2016-05-05
break yum classes out into files

lamont-granquist:
doesn't change anything or break anyone, this is just a copypasta change


lamont-granquist:
@chef/client-core just a lot of copypasta

coderanger:
:+1:

lamont-granquist:
might double check that i got all the requires lines right, and i left most of the rpm utility stuff in the yum/rpm_utils.rb file and didn't split them into individual files...

hrm, i wonder if the yum_cache can still find yum-dump.py tho...

jaymzh:
:+1: <3

lamont-granquist:
^ so that is option #1 of moving yum-dump.py into the same subdir as yum_cache.rb

i could also just "../yum-dump.py" it and treat the path into yum-dump.py as an "external API" -- i could concieve that someone might be calling that script directly, but feel kinda YAGNI about worrying about it unless someone speaks up...

lamont-granquist:
mildly amusing that all the tests pass even though i'm pretty damn certain i broke it completely with that first commit...

lamont-granquist:
yeah travis is lying, this PR is still broken, fails a manual test...

Issue #2860, opened 2015-02-04
Git resource's user attribute does not properly support whyrun mode


zwant:
### Version:

12.0.3
### Environment:

CentOS 6.5
### Scenario:

If a recipe is using the git-resource with the "user" attribute:

```
git /tmp/mydir do
  repository 'ssh://git@somehost.tld/myrepo.git'
  revision 'master'
  user user_created_in_chef_run
  group user_created_in_chef_run
  action :sync
end
```

And the "user" (in this case user_created_in_chef_run) does not yet exist (since we're in whyrun mode), the git provider will error out with:

```
Could not determine HOME for specified user 'user_created_in_chef_run' for resource 'git[/tmp/mydir/]
```

It seems the error lies in a workaround done here: https://github.com/chef/chef/blob/59236f77e5ce61dea27bb99ac259e59ef9436705/lib/chef/provider/git.rb#L305

```
if @new_resource.user
  run_opts[:user] = @new_resource.user
  # Certain versions of `git` misbehave if git configuration is
  # inaccessible in $HOME. We need to ensure $HOME matches the
  # user who is executing `git` not the user running Chef.
  env['HOME'] = begin
    require 'etc'
    Etc.getpwnam(@new_resource.user).dir
  rescue ArgumentError # user not found
    raise Chef::Exceptions::User, "Could not determine HOME for specified user '#{@new_resource.user}' for resource '#{@new_resource.name}'"
  end
end
```

The 'Etc.getpwnam' call needs to take into account whyrun mode as well.

lamont-granquist:
yeah the exception needs to be raised in define_resource_requirements

PR #4902, opened 2016-05-06
Handle numeric id for the user value in the git resource

MichaelPereira:
## Description

The user property of the git resource fails when passing the user id instead of the user name. Similarly to what is described in #2860. The problem comes from the fact that the function used to retrieve the user's home only takes a String as input. The changes in this PR allows to retrieve the user's home when the user id is being used. 

## Chef Version

Latest available RPM installed automatically via vagrant

## Platform Version

Centos 6.5

## Replication Case

```
git /tmp/mydir do
  repository 'ssh://git@somehost.tld/myrepo.git'
  revision 'master'
  user <user id>
  group <group id>
end
```
## Client Output

From memory (will update later with proper output I can't access now):

```
TypeError: no implicit conversion of Integer into String
```
Raised at this line: https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L300

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -307,7 +307,7 @@ class Chef
             when Integer
               Etc.getpwuid(@new_resource.user).dir
             else
-              Chef::Log.error("The `owneuser` parameter of the #@new_resource resource is set to an invalid value (#{new_resource.owner.inspect})")
+              Chef::Log.error("The `user` parameter of the #@new_resource resource is set to an invalid value (#{new_resource.owner.inspect})")
               raise ArgumentError, "cannot resolve #{new_resource.owner.inspect} to uid, owner must be a string or integer"
             end
             
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -302,13 +302,10 @@ class Chef
           env["HOME"] = begin
             require "etc"
             case @new_resource.user
-            when String
-              Etc.getpwnam(@new_resource.user).dir
             when Integer
               Etc.getpwuid(@new_resource.user).dir
             else
-              Chef::Log.error("The `user` parameter of the #@new_resource resource is set to an invalid value (#{new_resource.owner.inspect})")
-              raise ArgumentError, "cannot resolve #{new_resource.owner.inspect} to uid, owner must be a string or integer"
+              Etc.getpwnam(@new_resource.user.to_s).dir
             end
             
           rescue ArgumentError # user not found
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -301,7 +301,16 @@ class Chef
           # user who is executing `git` not the user running Chef.
           env["HOME"] = begin
             require "etc"
-            Etc.getpwnam(@new_resource.user).dir
+            case @new_resource.user
+            when String
+              Etc.getpwnam(@new_resource.user).dir
+            when Integer
+              Etc.getpwuid(@new_resource.user).dir
+            else
+              Chef::Log.error("The `owneuser` parameter of the #@new_resource resource is set to an invalid value (#{new_resource.owner.inspect})")
+              raise ArgumentError, "cannot resolve #{new_resource.owner.inspect} to uid, owner must be a string or integer"
+            end
+            
           rescue ArgumentError # user not found
             raise Chef::Exceptions::User, "Could not determine HOME for specified user '#{@new_resource.user}' for resource '#{@new_resource.name}'"
           end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/git.rb
```diff
@@ -301,7 +301,16 @@ def run_options(run_opts = {})
           # user who is executing `git` not the user running Chef.
           env["HOME"] = begin
             require "etc"
-            Etc.getpwnam(@new_resource.user).dir
+            case @new_resource.user
```
coderanger: I would probably change this to `when Integer; Etc.getpwuid(@new_resource.user).dir else Etc.getpwnam(@new_resource.user.to_s).dir` and trust in the resource to handle validation.

MichaelPereira: I see, because the resource type is already constrained in https://github.com/chef/chef/blob/master/lib/chef/resource/scm.rb#L72, there is no reason it would go into the else case right?

MichaelPereira:
cc @schisamo who wrote the original code

schisamo:
:+1:

MichaelPereira:
The failure of the travis build is the following:

```
Failures:
  1) Chef::Provider::Git with a user id with a specific home clones a repo with amended git options with specific home
     Failure/Error: @provider.clone
     
       #<Chef::Provider::Git:0x00000018225200> received :shell_out! with unexpected arguments
         expected: ("git clone  \"git://github.com/opscode/chef.git\" \"/my/deploy/dir\"", {:user=>123, :environment=>{"HOME"=>"/home/masterNinja"}, :log_tag=>"git[web2.0 app]"})
              got: ("git clone \"git://github.com/opscode/chef.git\" \"/my/deploy/dir\"", {:user=>123, :log_tag=>"git[web2.0 app]", :environment=>{"HOME"=>"/home/masterNinja"}})
       Diff:
       
       @@ -1,5 +1,5 @@
       -["git clone  \"git://github.com/opscode/chef.git\" \"/my/deploy/dir\"",
       +["git clone \"git://github.com/opscode/chef.git\" \"/my/deploy/dir\"",
         {:user=>123,
       -  :environment=>{"HOME"=>"/home/masterNinja"},
       -  :log_tag=>"git[web2.0 app]"}]
       +  :log_tag=>"git[web2.0 app]",
       +  :environment=>{"HOME"=>"/home/masterNinja"}}]
```

Which is the unit test I created and ran successfully.
Since it's a Hash that is passed as parameters and that I can't be sorted as far as I know, is there a way to make the verification look at the hash values?

spec/unit/provider/git_spec.rb
```diff
@@ -267,6 +267,42 @@
     end
   end
 
+  context "with a user id" do
+    let(:deploy_user)  { 123 }
+    let(:expected_cmd) { 'git clone  "git://github.com/opscode/chef.git" "/my/deploy/dir"' }
```
thommay: there's a stray space between `clone` and `"git://` which won't be helping the specs.

spec/unit/provider/git_spec.rb
```diff
@@ -267,6 +267,42 @@
...
+          :environment => { "HOME" => "/home/masterNinja" },
+          :log_tag => "git[web2.0 app]",
+        }
+      end
+      before do
+        @resource.environment(override_home)
+      end
+      before { @resource.environment(override_home) }
+      it "clones a repo with amended git options with specific home" do
+        expect(@provider).to receive(:shell_out!).with(expected_cmd, overrided_options)
```
thommay: `hash_include(overriding_options)`

lamont-granquist:
if you flatten the hash into named arguments both in the code and in the test does it work?

```
expect(@provider).to receive(:shell_out!).with(expected_cmd, **overrided_options)
```

and then tack **run_options into the code (and go fix all the other test failures that probably blow up because of that...)

you also have that space to fix.

thommay:
@MichaelPereira hopefully this clears up the specs, and we'll get this merged.

MichaelPereira:
@thommay @lamont-granquist 

The new unit test is passing locally with @thommay changes. I hope it will resolves the appveyor/travis testing as well.

Unfortunately, the extra space seems needed to pass this unit test as well as others in the current form as it is expected in the command received by shell_out!:
https://gist.github.com/MichaelPereira/72c4879f24eb296a4e5562d30655ee43

MichaelPereira:
Removing the extra space fixed the unit testing issue, thanks for spotting it.

Any idea as to why the extra space in the unit test above is not causing any issue? https://github.com/MichaelPereira/chef/blob/321106fbf5e92f6bcad8bba78e5ed59642777747/spec/unit/provider/git_spec.rb#L221

thommay:
👍

smurawski:
👍

lamont-granquist:
👍

PR #4903, opened 2016-05-06
UID now starts at 501, uses createhomedir instead

nmcspadden:
I am a Facebook employee, so please do not merge this - Phil Dibowitz (PhilD) will do the final merge.

Brief summary:
Changed the default UID from starting at 500 to 501.
Home directory is created by using Apple's createhomedir command rather than ditto-ing the User Template.

Explanation:
UID 500 -> 501
By default, Chef's provider for creating users on Mac OS X, dscl.rb, creates a new user starting at UID 500. This should be changed to 501.

There's a lot of historic and current evidence that this should always be the default behavior:

1. The very first user account created when a new machine is setup is always UID 501 (and is always an admin account). Apple never creates an account through any user-facing means with a UID of less than 501.
2. Accounts with UID 500 or less do not show up in the System Preferences' "Users & Groups" pane. 
3. Apple's Server.app is unable to register a sub-501 account for any services, since it requires only "real" local accounts to do that - which means a UID of 501 or higher. An account with a UID of 500 causes Server.app to not function properly when used as a service owner. This gives more credence to the idea that UID 500 is bad.

Evidence here that Apple expects you to use a UID of appropriate value:
https://developer.apple.com/library/mac/documentation/Porting/Conceptual/PortingUnix/additionalfeatures/additionalfeatures.html#//apple_ref/doc/uid/TP40002856-BBCDBFBD
Step 4 of that example says:

Choosing a user ID is not trivial to do programmatically. For sample code, see the Starting Points appendix in Shell Scripting Primer.

Here in User & Group Management:
https://developer.apple.com/library/mac/documentation/OpenSource/Conceptual/ShellScripting/StartingPoints/StartingPoints.html#//apple_ref/doc/uid/TP40004268-CH2-SW24
Listing D-10's code sample shows:

# UID: The user ID for the new user. Leave blank for the script to automatically
# choose the first unused ID at or above MINUID (currently 501).
There are other examples in Apple documentation all indicating that they expect 501 to be the starting UID for "normal" user accounts.

createhomedir

Currently, the dscl provider does a ditto of the User Template ( /System/Library/User Template/English.lproj) into the location specified by the home directory property. While this is perfectly fine behavior, Apple provides a tool that does this directly, called createhomedir, which dates back to 10.6.2 (it may be earlier than that, but that's the first time I can find evidence of it online).

createhomedir does all the work of copying the user template, changing permissions and ownership, and filling out any appropriate post-copy actions. This is the same behavior used a new user account is added by the System Preferences (or Setup Buddy).

The call to create a home directory for a specific local user can be done this way:
/usr/sbin/createhomedir -c -u <username

I think it would be best to use the Apple-provided tools to accomplish this task rather than trying to mimic the behavior. If at any point Apple changes the behavior of what happens when someone creates a new user account, their own frameworks are the best tools to use, rather than hoping we get it right.

In order for createhomedir to function, the user's NFSHomeDirectory record has to be created first, so I moved the call to dscl before the call to ditto_home that actually creates the home directory.

### DIFF
--- a/lib/chef/provider/user/dscl.rb
+++ b/lib/chef/provider/user/dscl.rb
@@ -208,7 +208,7 @@ user password using shadow hash.")
         #
         # Sets the user id for the user using dscl.
         # If a `uid` is not specified, it finds the next available one starting
-        # from 200 if `system` is set, 500 otherwise.
+        # from 200 if `system` is set, 501 otherwise.
         #
         def dscl_set_uid
           # XXX: mutates the new resource
@@ -223,11 +223,11 @@ user password using shadow hash.")
 
         #
         # Find the next available uid on the system. starting with 200 if `system` is set,
-        # 500 otherwise.
+        # 501 otherwise.
         #
         def get_free_uid(search_limit = 1000)
           uid = nil
-          base_uid = new_resource.system ? 200 : 500
+          base_uid = new_resource.system ? 200 : 501
           next_uid_guess = base_uid
           users_uids = run_dscl("list /Users uid")
           while next_uid_guess < search_limit + base_uid
@@ -291,6 +291,8 @@ user password using shadow hash.")
             return
           end
 
+          run_dscl("create /Users/#{new_resource.username} NFSHomeDirectory '#{new_resource.home}'")
+
           if new_resource.supports[:manage_home]
             validate_home_dir_specification!
 
@@ -302,7 +304,6 @@ user password using shadow hash.")
               move_home
             end
           end
-          run_dscl("create /Users/#{new_resource.username} NFSHomeDirectory '#{new_resource.home}'")
         end
 
         def validate_home_dir_specification!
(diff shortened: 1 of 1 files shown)


coderanger:
tl;dr some apple employee 10 years ago screwed up `<` vs. `<=` :laughing:

thommay:
You'll need to fix the failing specs on both windows and linux, @nmcspadden .

chef-supermarket:

* @nmcspadden

nmcspadden:
I will fix the rspec tests soon.  Thanks.

thommay:
@nmcspadden ping!

nmcspadden:
Sorry for the huge delay on this. Need to find time to teach myself the spec language so I can fully understand how it works.  To be honest, I don't really get how the tests were working previously, nor why my changes would have suddenly broken them. I understand that the change to the home directory requires a change in spec testing, but  I don't understand why "dscl is not found" is a suddenly new spec fail.

lamont-granquist:
code looks fine, but does need spec fixes, can't merge with the angry red

thommay:
@nmcspadden @jaymzh any chance of rescuing this? :)

nmcspadden:
Sorry I let this sit for so long. I'll rebase this whole thing and fix the spec tests, should have this done by Wednesday.

thommay:
No worries, thanks @nmcspadden

PR #4907, opened 2016-05-06
reject malformed ini content in systemd_unit resource

nathwill:
was thinking about this the other day, and wanted to send a PR to raise the subject:

currently, we pass the content string through as-is, which means it's possible to give bad input to systemd. this fixes it so that it will at least be enforced as a well-formed ini-formatted string, but it's not perfect...

pro: ensure well-formed ini content string by attempting to parse as ini before converting back to string

con: `IniParse.parse(str).to_s` reformats the string with spaces between the `=` and an empty line between headings, so it's possible that a valid ini string may not match the exact string as rendered on disk, thus breaking idempotency :frowning:...

maybe this is a case that should just be covered by integration testing, but i figured i'd raise the subject and see what Chef maintainers thought about how best to handle it.

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -20,6 +20,7 @@ require "chef/provider"
 require "chef/mixin/which"
 require "chef/mixin/shell_out"
 require "chef/resource/file"
+require "iniparse"
 
 class Chef
   class Provider
@@ -41,6 +42,15 @@ class Chef
         current_resource.triggers_reload(new_resource.triggers_reload)
 
         current_resource
+      end
+
+      def define_resource_requirements
+        super
+
+        requirements.assert(:create) do |a|
+          a.assertion { IniParse.parse(new_resource.to_ini) }
+          a.failure_message "Unit content is not valid INI text"
+        end
       end
 
       def action_create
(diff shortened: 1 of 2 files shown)


thommay:
it seems like IniParse could just be a validator, rather than involved in writing out the content to a file?

lib/chef/resource/systemd_unit.rb
```diff
@@ -53,7 +55,12 @@ def to_ini
             end
           end.to_s
         else
-          content.to_s
+          begin
+            IniParse.parse(content).to_s
+          rescue IniParse::ParseError => error
+            Chef::Log.fatal "Malformed content property for #{self}: #{error}"
+            raise error
```
lamont-granquist: i don't think that decorating the parser error here gives any more information than what we get out of the exception decoration that chef-client already does and this could probably be simplified by just letting the exception trickle up

lamont-granquist:
another approach would be to just add "IniParse.parse(new_resource.content)" to define_resource_requirements in the provider.

nathwill:
oh cool, define_resource_requirements looks perfect for this :+1:

nathwill:
hmmm, what about putting IniParse#parse in coerce on the property declaration?

lamont-granquist:
@thommay was concerned about it mutating the desired content and idempotency issues.   its also a good idea to let resources for things like 'content' act as accumulators and build complete state from partial state, and if you validate in the resource you'll do that too early and possibly validate on partial content.  if you do it late in define_resource_requirements then callers can `resource.content << stuff` multiple times to build the content before calling the action.

nathwill:
@thommay @lamont-granquist does this look more in line with what you were suggesting?

lamont-granquist:
👍 

@thommay?

thommay:
👍

PR #4908, opened 2016-05-06
add systemd_unit try-restart, reload-or-restart, reload-or-try-restart actions

nathwill:
systemd has some fancy unit commands that do roughly the equivalent of the service resource's "supports" behavior for reloads, where you can say "reload-or-restart" and it'll issue a restart if the unit doesn't support reloading (as defined by the ExecReload directive);

they're defined as:

- try-restart: Restart one or more units if active
- reload-or-restart: Reload one or more units if possible, otherwise start or restart
- reload-or-try-restart: Reload one or more units if possible, otherwise restart if active

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -133,6 +133,24 @@ class Chef
         end
       end
 
+      def action_try_restart
+        converge_by("try-restarting unit: #{new_resource.name}") do
+          systemctl_execute!("try-restart", new_resource.name)
+        end
+      end
+
+      def action_reload_or_restart
+        converge_by("reload-or-restarting unit: #{new_resource.name}") do
+          systemctl_execute!("reload-or-restart", new_resource.name)
+        end
+      end
+
+      def action_reload_or_try_restart
+        converge_by("reload-or-try-restarting unit: #{new_resource.name}") do
+          systemctl_execute!("reload-or-try-restart", new_resource.name)
+        end
+      end
+
       def active?
         systemctl_execute("is-active", new_resource.name).exitstatus == 0
       end
--- a/lib/chef/resource/systemd_unit.rb
+++ b/lib/chef/resource/systemd_unit.rb
@@ -29,7 +29,9 @@ class Chef
                       :enable, :disable,
                       :mask, :unmask,
                       :start, :stop,
-                      :restart, :reload
+                      :restart, :reload,
+                      :try_restart, :reload_or_restart,
+                      :reload_or_try_restart
 
       property :enabled, [TrueClass, FalseClass]
       property :active, [TrueClass, FalseClass]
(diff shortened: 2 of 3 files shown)


thommay:
i guess it would be too confusing a behaviour change to just default `restart` to `try-restart`, but I'm fine with adding these. (And `reload-or-try-restart` as the default `reload`)

lamont-granquist:
our general semantics are that `:restart` falls back to `:start` and if there's no `reload_command` i think we revert to `:restart` don't we?  in general we try to DTRT, and i'd prefer not to add more fussy actions and to just make the actions that we have be more resiliant.

nathwill:
> in general we try to DTRT

to my mind this means correctly reflecting the default behaviors of what we're interacting with, without surprises; in the case where the system we're modeling provides explicit commands for the behaviors that other resources are providing as "extra" behaviors that aren't actually in the underlying utilities (i.e. service resource action fallback modes), is there any reason we wouldn't want to expose the "native" facilities directly as resource actions?

lamont-granquist:
Actually that's not what we tend to do.  If the underlying actions are brittle, then we usually wrap those with idempotent actions that are more resilient.  If that result looks really awful then we tend not to.  But for example if a sysv init script doesn't implement `:restart` we fall back to `:start`+`:stop` for the user rather than just bailing with an error and forcing the user to do both of those actions manually.

thommay:
I don't think we lose anything by including these, and users may well expect them 👍

lamont-granquist:
okay so i guess these are trivial.

for the future record if anyone finds this and agrees with me that this is overly fussy and `:restart` should just hit `try-restart` consider this notice that i'd like to see that PR.

👍

nathwill:
@lamont-granquist thanks; i can definitely see some justification for deviating from the default system behavior by falling back to restart if a reload fails or similar rescuing behaviors (maybe even just using reload-or-restart instead of reload). my concern with replacing restart with try-restart in particular is that try-restart is a NOP if the unit isn't already active, which is pretty contrary to what i think most people reasonably expect: that issuing a restart would result in the service running.

```bash
[nathwill@wyrd ~]$ systemctl status redis.service
● redis.service - Redis persistent key-value database
   Loaded: loaded (/usr/lib/systemd/system/redis.service; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/redis.service.d
           └─limit.conf
   Active: inactive (dead)
[nathwill@wyrd ~]$ sudo systemctl try-restart redis.service
[nathwill@wyrd ~]$ systemctl status redis.service
● redis.service - Redis persistent key-value database
   Loaded: loaded (/usr/lib/systemd/system/redis.service; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/redis.service.d
           └─limit.conf
   Active: inactive (dead)
```

that's definitely weird, and i don't care for it, but it makes a little sense in light of systemd being api-driven with the unit having an internal state, instead of being a script that's being executed.

PR #4912, opened 2016-05-09
the =~ in ruby is awful

lamont-granquist:
if you decorate/delegate =~ you can't update the capture variables
which are local to the calling context.

maybe something could be done to update the $~ in the binding of
the caller, but i don't want to do that today...

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -46,7 +46,7 @@ class Chef
               lead = 0
               tail = evr.size
 
-              if evr =~ %r{^([\d]+):}
+              if %r{^([\d]+):}.match(evr)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core this works now.

i'd prefer to add unit spec tests for the rpmutils on top of #4901 since that is already in flight

coderanger:
Does this pass tests? I thought `match()` doesn't set `$1` et al because it returns the matched values in a MatchData object.

coderanger:
Looks like I'm wrong, it does return the MatchData object but also sets the pseudo-globals.

randomcamel:
Not super-excited about deferring the specs, but 👍 for solving this particular problem.

PR #4922, opened 2016-05-10
Add CentOS/RHEL coverage to travis testing

lamont-granquist:
- updates some travis config
- adds centos-6 testing
- adds a bit more coverage of some other community cookbooks


lamont-granquist:
@chef/client-core okay, this is probably in reviewable shape, unless i added a typo or something before the rebase+force-push it should go green...

.travis.yml
```diff
@@ -124,12 +124,64 @@ matrix:
       - cd kitchen-tests
     script:
       # FIXME: we should fix centos-6 against AWS and then enable it here
```
thommay: that comment can go now ;)

.travis.yml
```diff
@@ -124,12 +124,64 @@ matrix:
...
+    gemfile: kitchen-tests/Gemfile
+    before_install:
+      - gem update --system $(grep rubygems omnibus_overrides.rb | cut -d'"' -f2)
+      - gem install bundler -v $(grep bundler omnibus_overrides.rb | cut -d'"' -f2)
+      - echo -n $DO_KEY_CHUNK_{0..30} >> ~/.ssh/id_aws.base64
+      - cat ~/.ssh/id_aws.base64 | tr -d ' ' | base64 --decode > ~/.ssh/id_aws.pem
+    before_script:
+      - cd kitchen-tests
+    script:
+      # FIXME: we should fix centos-6 against AWS and then enable it here
```
thommay: and here

thommay:
👍 with micro-nits

.travis.yml
```diff
@@ -124,12 +124,64 @@ matrix:
       - cd kitchen-tests
     script:
       # FIXME: we should fix centos-6 against AWS and then enable it here
-      - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen test ubuntu; fi
+      - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen test ubuntu-1404 -l debug; fi
     after_failure:
       - cat .kitchen/logs/kitchen.log
     after_script:
-      - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen destroy ubuntu; fi
+      - if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then bundle exec kitchen destroy ubuntu-1404 -l debug; fi
```
btm: these two scripts have `-l debug`, but the separate centos block below does not.

kitchen-tests/.kitchen.yml
```diff
@@ -7,6 +7,8 @@ driver:
 
 provisioner:
   name: chef_github
+  chef_omnibus_url: "https://omnitruck.chef.io/current/install.sh"
+  chef_omnibus_install_options: "-n"
```
btm:
```
$ diff www.install.sh omnitruck.install.sh
565c565
< metadata_url="http://www.chef.io/$channel/$project/metadata?v=$version&p=$platform&pv=$platform_version&m=$machine"
---
> metadata_url="https://omnitruck.chef.io/$channel/$project/metadata?v=$version&p=$platform&pv=$platform_version&m=$machine"
```

Huh. That's kind of confusing. `-n` is backcompat that resolves to `-c current`.

What URL is canonical / most permanent? @sersut?

I presume `chef_omnibus_url` defaults to something like `https://www.chef.io/chef/install.sh`, in which case it would seem like settings `chef_omnibus_install_options` to `-c current` would be the least likely to get deprecated/changed in the future.

btm:
:+1: with minor feedback

kitchen-tests/.kitchen.yml
schisamo: We should be using the following [per our documentation](https://docs.chef.io/packages.html#omnitruck-install-scripts):

```yaml
chef_omnibus_url: "https://omnitruck.chef.io/install.sh"
chef_omnibus_install_options: "-c current"
```

The default project is chef so adding a `-P chef` is not required but is more explicit so you could add it. The `-n` option only exists for backward compatibility and should not be used in any of our code/repos/products/documentation:
https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/generator/bourne/scripts/script_cli_parameters.sh#L24-L25

I also opened a PR in Test Kitchen to ensure it is using `https://omnitruck.chef.io/install.sh` as its default value for `chef_omnibus_url`: https://github.com/test-kitchen/test-kitchen/pull/1028

PR #4923, opened 2016-05-10
Disable Docker cookbook acceptance tests until the cookbook or Chef is fixed

adamedx:
Acceptance tests for Docker are currently failing due to a problem starting a service. For now we'll disable this test and file an issue to track fixing the cookbook or Chef if it's a product issue, or otherwise fixing the test if it's a test issue (which seems more likely given that simply starting the service is failing).

### DIFF
--- a/acceptance/top-cookbooks/.kitchen.docker.yml
+++ b/acceptance/top-cookbooks/.kitchen.docker.yml
@@ -6,7 +6,4 @@ suites:
     run_list:
       - recipe[apt]
       - recipe[apt-docker]
-      - recipe[docker_test::installation_package]
-      - recipe[docker_test::service_upstart]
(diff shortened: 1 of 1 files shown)


ksubrama:
They work fine on my local box - that's what's been confounding me about all these tests.  We keep turning them off one at a time due to flakiness when really, we should just turn it all off if it's not providing us with value.  Currently we can't debug any of them and they're all flaky in some aspect or the other.

adamedx:
@ksubrama  I agree that we need more investigation, but the priority is to keep the pipeline unblocked. If the tests are unreliable, they don't really provide value. We will pay for those missing tests with the cost of having to explicitly launch this verification on builds already in current.

randomcamel:
👍

PR #4926, opened 2016-05-12
Skip node attribute saving with named run lists

ustuehler:
[Policyfile.rb documentation](https://docs.chef.io/config_rb_policyfile.html) suggests named run lists can be used as an alternative to an override run list. A natural assumption then is that node attribute saving also behaves the same with an override run list as with named run lists.

Quoting the Policyfile.rb documentation:

> named_run_list "NAME", "ITEM1", "ITEM2", ...
> Specify a named run-list to be used as an alternative to the override run-list. This setting should be used carefully and for specific use cases, like running a small set of recipes to quickly converge configuration for a single application on a host or for one-time setup tasks.

This change marks the policy as temporary if a named run list is requested, so that chef-client will not save node attributes at the end of the run, just as with an override run list.
### Version:

Chef 12.10.45
### Environment:

chef-client runs in policyfile mode
### Scenario:

We want to run a subset of recipes (e.g., `cookbook_name::deploy`) using a named run list and not save the node's run_list at the end of the chef-client run in order not to break Chef searches such as `recipes:cookbook_name\:\:appserver`. This works with an override run list, but not in policyfile mode with named run lists.
### Steps to Reproduce:
1. Run `chef-client` with `policy_name` and `policy_group` set in `/etc/chef/client.rb`, then run `knife node show <node_name>`.
2. Run `chef-client -n <named_run_list>` (where the named run list uses a different set of recipes than the default run list) and run `knife node show <node_name` again.
3. Compare the output of `knife node show` before and after the second chef-client run.
### Expected Result:

The list of recipes and other node attributes have not changed between client runs.
### Actual Result:

The list of recipes and possibly node attributes have changed.

### DIFF
--- a/lib/chef/policy_builder/policyfile.rb
+++ b/lib/chef/policy_builder/policyfile.rb
@@ -194,12 +194,13 @@ class Chef
         cookbooks_to_sync
       end
 
-      # Whether or not this is a temporary policy. Since PolicyBuilder doesn't
-      # support override_runlist, this is always false.
+      # Whether or not this is a temporary policy, which means a
+      # named_run_list was provided. Chef::Client uses this to decide whether
+      # to do the final node save at the end of the run or not.
       #
       # @return [false]
       def temporary_policy?
-        false
+        named_run_list_requested?
       end
 
       ## Internal Public API ##
--- a/spec/unit/policy_builder/policyfile_spec.rb
+++ b/spec/unit/policy_builder/policyfile_spec.rb
@@ -583,6 +583,10 @@ describe Chef::PolicyBuilder::Policyfile do
               expect(Chef::CookbookCacheCleaner.instance.skip_removal).to be(true)
             end
 
+            it "gives `true` for #temporary_policy?" do
+              expect(policy_builder.temporary_policy?).to be(true)
+            end
+
           end
 
         end


coderanger:
Technically a compat change, but probably a good one? /cc @chef/client-core

lamont-granquist:
paging @danielsdeleo except he's OOTO...  hmmm...

lamont-granquist:
so i know for certain that dan didn't just overlook this and was well aware of the parallels between override run lists and named run lists.

unfortunately he's on PTO until June 5th.

however, i would really not want to merge this and start going down this route without his architectural oversight.

danielsdeleo:
It's been a while since I wrote it the way it is now, so I don't remember precisely my reasoning, but I think it was basically:

1. From experience, I don't think searching by run list item is a good pattern. For example, if you refactor a cookbook by breaking a recipe (say, the default recipe) into components, then you need to do extra work somewhere to prevent that from breaking run list item searches. With policyfiles, the `policy_name` and `policy_group` are stable over time as the run list changes (and regardless of which named run list you use), so I think it's better to change your searches to look for those. That would make the specific problem listed in the original post go away. If you need to look for multiple node types that share a given recipe, you should use node tags which are implemented with normal attributes and will persist across runs with different run lists (though you do have to do some work to untag if you wish them to go away).
2. Skiping node saves is a double-edged sword that makes a lot of other things complicated and is definitely wrong for some use cases. I know there are folks using named run lists on a more or less permanent basis, and this change would be bad for them. For another example, suppose you had a named run list that bootstrapped the installation of C libraries that you wanted to use with `chef_gem`, so you run `chef-client -j dna.json -n bootstrap && chef-client` (or do something else equivalent). The second chef run would be depending on the first one saving the policy name and group in that case.

Of course, there are downsides to the way it works now, which as noted are:

1. It breaks run list searches and as much as I don't like them, that pattern is used pretty widely
2. It will break searches for attributes set by cookbooks that are not included by the named run list.

So neither way is definitively *correct*, but to me the weight of the arguments for keeping it the way it is now feel a bit stronger. That said, I'm open to discussion, but we should consider this carefully and probably any change should wait for a major version bump (but could be a opt-in option in the interim).

lamont-granquist:
I guess I'm 👎 based on Dan's comment -- feel like people should probably circle around at chefconf and debate this one and try to come to some agreement...

danielsdeleo:
Seems like there should be some other kind of compromise, like stuffing all the run list items from all the run lists into some other ohai attribute or something (wouldn't address the disappearing cookbook attribute problem though).

lamont-granquist:
@coderanger looks we found that PR i was trying to think of the other day.

lamont-granquist:
Based on discussion it seems like the way to rescue this would be to have a setting in the Policyfile.rb, and possibly a setting on each named_run_list that would disable saving the node data to let users select which behavior they want.

As this hasn't been getting much attention nothing is getting done.  Users that want this feature should get involved (not just in this ticket, adding to the flood of everyone's github e-mails will just be yelling into the void -- get involved on slack, engage our support group if you pay for chef, come to the chef community meetings and propose an RFC, etc...)

lamont-granquist:
I think I'm going to close this.

If there's a stakeholder here, they need to show up to the thursday community meetings and propose an RFC or otherwise get involved in the RFC process, or at least start up a discussion on the community slack in the #policyfiles channel.  So far the existing policyfiles community has been using this feature without a lot of people piping up that they need this feature.

(Also anyone that has a support contract can engage us about this, so far nobody has).

PR #4928, opened 2016-05-12
convert kitchen tests to docker + other fixes

lamont-granquist:
- use kitchen-dokken instead of kitchen-ec2
- add tests for gem metadata + chef-sugar
- convert to inspec and add some tests

this disables the webapp stuff completely and its testing for now, that
needs to be updated to more modern httpd/mysql cookbooks and have its
testing converted to inspec.

more O/S distros are also in the pipeline.


lamont-granquist:
@chef/client-core want to get this merged to make some iterative progress, there's enough here that shipping it will be mostly a major level up.

smurawski:
Let's do it! 👍

kitchen-tests/.kitchen.travis.yml
```diff
@@ -20,22 +22,90 @@ provisioner:
...
+    intermediate_instructions:
+      - RUN /usr/bin/apt-get update
+      - RUN /usr/bin/apt-get -y install zlib1g-dev sudo net-tools
+
+- name: centos-5
+  driver:
+    image: centos:5
+    platform: rhel
+    run_command: /sbin/init
+    provision_command:
```
tas50: This should be intermediate_instructions

kitchen-tests/.kitchen.travis.yml
```diff
@@ -20,22 +22,90 @@ provisioner:
...
+      - RUN yum clean all
+      - RUN yum install -y which initscripts net-tools sudo
+
+- name: centos-6
+  driver:
+    image: centos:6
+    intermediate_instructions:
```
tas50: Probably wanna add     run_command: /sbin/init

PR #4933, opened 2016-05-13
wire up chef handlers directly from libraries

lamont-granquist:
smear just a little bit of minimal DSL wiring into the Chef::Report
class so that classes can wire themselves up automatically as
handlers.

this also extends the Chef::Config[*_handler] arrays to take classes
as well as instances.

in the case of passing a class that is not a singleton, a new instance
will be created when the report is run.  it will obviously be difficult
to pass any state into this instance.

if the class that is passed supports calling `#instance` we assume it
is a Singleton and get the instance and use that.  this is to support
other chef code grabbing/creating the singleton instance and poking
it with data (probably configuring where the reporter goes and what
creds it needs to publish and other stuff).

### DIFF
--- a/lib/chef/handler.rb
+++ b/lib/chef/handler.rb
@@ -57,9 +57,39 @@ class Chef
   #
   class Handler
 
+    def self.handler_for(*args)
+      if args.include?(:start)
+        Chef::Config[:start_handlers] ||= []
+        Chef::Config[:start_handlers] |= [self]
+      end
+      if args.include?(:report)
+        Chef::Config[:report_handlers] ||= []
+        Chef::Config[:report_handlers] |= [self]
+      end
+      if args.include?(:exception)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
this eliminates the need for the chef_handler cookbook and resource.

it pushes loading the handlers up to class parsing time in library files (and if we ever get requires: working in gem metadata installation this will come along for the ride).

as it is right now chef_handler is a resource so the exception handlers don't get wired up until you're down in resource convergence phase -- although you could force the resource to compile time, but still...

this still doesn't fix start_handlers or super-early-exception handling.

lamont-granquist:
@chef/client-core should be ready for review

thommay:
This is super nice! 👍

coderanger:
This seems a little bit redundant with the event handler DSL stuff, we should probably re-redundant-ize those two systems at some point :)

PR #4940, opened 2016-05-16
Add licenses to all software defs

jkeiser:


omnibus/config/software/chef-gem-ruby-shadow.rb
```diff
@@ -4,3 +4,7 @@
 # another gem's installation.
 require_relative "../../files/chef-gem/build-chef-gem/gem-install-software-def"
 BuildChefGem::GemInstallSoftwareDef.define(self, __FILE__)
+
+license "PublicDomain"
```
btm: I think you want `Public-Domain`

https://github.com/chef/omnibus/blob/master/lib/omnibus/licensing.rb#L391

lamont-granquist:
👍

smurawski:
👍  with btm's update

PR #4943, opened 2016-05-17
Add a bunch of knife integration tests

thommay:
There's some bonus work to allow input on stdin to be passed to knife

### DIFF
--- a/spec/integration/knife/client_bulk_delete_spec.rb
+++ b/spec/integration/knife/client_bulk_delete_spec.rb
@@ -1,0 +1,130 @@
+#
+# Copyright:: Copyright 2013-2016, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "support/shared/integration/integration_helper"
+require "support/shared/context/config"
+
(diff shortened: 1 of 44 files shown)


thommay:
🚧 still in progress 🚧

thommay:
@chef/client-maintainers review appreciated; apologies for the enormitude.

spec/integration/knife/client_bulk_delete_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+  include KnifeSupport
+
+  include_context "default config options"
+
+  when_the_chef_server "has some clients" do
+    before do
+      client "cons", {}
+      client "car", {}
+      client "cdr", {}
+      client "cat", {}
```
stevendanna: Would `concat` be a useful test case?

spec/integration/knife/client_bulk_delete_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    it "deletes all matching clients" do
+      knife("client bulk delete ^ca.*", input: "Y").should_succeed <<EOM
```
stevendanna: Reading this test, my primary question is "what happens if I don't specify the ^ anchor.

spec/integration/knife/client_create_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    it "saves the private key to a file" do
+      Dir.mktmpdir do |tgt|
+        knife("client create -f #{tgt}/bah.pem bah").should_succeed stderr: out
+        expect(File).to exist("#{tgt}/bah.pem")
+      end
+    end
+
+    it "reads the public key from a file" do
```
stevendanna: Can we test that the public key was actually sent up to the server?

smurawski:
:+1:

spec/integration/knife/client_bulk_delete_spec.rb
thommay: added a test for that case

spec/integration/knife/client_create_spec.rb
thommay: not easily - zero doesn't store the default key since it doesn't really care about it

PR #4953, opened 2016-05-19
update readme regarding minor version bump

mwrock:


btm:
:+1: does/will the internal release checklist just link to this as a step?

mwrock:
yes! @btm just waiting to push this so I get a good link

PrajaktaPurohit:
+1

PR #4958, opened 2016-05-20
Update .mailmap with a few more entries

stevendanna:


lamont-granquist:
:+1:

@thommay should probably chime in about his e-mail addy there.

PR #4962, opened 2016-05-20
Update contributing doc to better reflect reality

tas50:
Point to the new mailing list
Remove the issue template since that's handled automatically now
Remove office hours since we're not doing that anymore
https a few more URLs


ranjib:
:+1: @tas50 do you know what happened to slack migration efforts? If anything is done on that front, should we include that

coderanger:
@ranjib Nathen and I worked out a plan, I'm blocked on getting some cycles from the Chef folks to set up the landing page.

smurawski:
👍

PR #4963, opened 2016-05-22
Adding myself as a Chef Maintainer

thehar:
Spoke with Nathen Harvey about being added as a maintainer.


coderanger:
Hi there, generally we like to see a history of contribution but I suppose Provisioning is kind of in a weird state right now. You'll want to add yourself in the MAINTAINERS.toml file, not the markdown. The latter is generated automatically by the former.

PR #4965, opened 2016-05-24
fix bundle locking on bundler 1.12

mwrock:
As of bundler 1.12.0, running our rake bundler tasks wipes out the windows dependencies when run on mac/linux and wipes out the non-windows dependencies when run on windows. This change can be narrowed down to a change to the `gemspec` DSL in [this one line](https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L71). v1.12.0.pre.1 added the `platform` parameter. Removing that parameter duplicates the behavior we see in 1.11.2. We can get this same behavior by not using `gemspec` and calling `gem` instead via `path: "."`.

I'm not crazy about this trickery but it appears to be the best way to support 1.12.4 and beyond.

### DIFF
--- a/tasks/bundle_util.rb
+++ b/tasks/bundle_util.rb
@@ -57,7 +57,14 @@ module BundleUtil
 
         # Run the bundle command
         ruby_platforms = platform ? PLATFORMS[platform].join(" ") : "ruby"
-        cmd = Shellwords.join([Gem.ruby, "-S", bundle_platform, ruby_platforms, *args])
+        cmd = Shellwords.join([
+          Gem.ruby,
+          "-S",
+          bundle_platform,
+          ruby_platforms,
+          "_#{desired_bundler_version}_",
+          *args,
+        ])
         puts "#{prefix}#{Shellwords.join(["bundle", *args])}#{platform ? " for #{platform} platform" : ""}:"
         with_gemfile(gemfile) do
           puts "#{prefix}BUNDLE_GEMFILE=#{gemfile}"
@@ -65,7 +72,7 @@ module BundleUtil
           if extract_output
             `#{cmd}`
           else
-            unless system(bundle_platform, ruby_platforms, *args)
+            unless system(bundle_platform, ruby_platforms, "_#{desired_bundler_version}_", *args)
               raise "#{bundle_platform} failed: exit code #{$?}"
             end
           end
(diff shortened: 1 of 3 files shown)


mwrock:
cc @jkeiser @ksubrama

ksubrama:
Seems reasonable enough to me.

btm:
@mwrock did you look for an associated issue filed against bundler we could link against?

mwrock:
oh yes. I put the commit in the commens and here is [a link](https://github.com/bundler/bundler/pull/4166) to the PR @btm

btm:
That's the cause, right? Is there a new bug or is this a breaking change we have to eat long term?

mwrock:
No I think that PR was a "feature" and what we are seeing is a byproduct of doing something that bundler is not supporting right now. Looking through vasious issues and PRs, I get the sense that the bundler maintainers want to support "work everywhere" lock files (or something like that) but just have not set about tackling it.

So we'd be filing an issue in defence of an unsupported workflow and it looks like the original PR was motivated to fix an issue where this breaks windows and thats something we have seen here too. For example our `bundle:install` rake task does not work on windows in the chef-dk repo due to certain gems it depends on.

tasks/bundle_util.rb
```diff
@@ -91,4 +93,23 @@ def with_gemfile(gemfile)
   def platforms
     PLATFORMS.keys
   end
+
+  def validate_bundler_version!
+    actual_version = `#{Gem.ruby} #{Gem.bin_path("bundler", "bundle")} -v`.chomp
+    unless actual_version.end_with?(desired_bundler_version)
+      raise "Running #{actual_version}. Please install the required"\
```
btm: might call actual_version something like `version_string` since it looks like `Bundler version 1.11.2`

btm:
:+1:

tasks/bundle_util.rb
```diff
@@ -91,4 +93,23 @@ def with_gemfile(gemfile)
...
+        " version #{desired_bundler_version} before continuing!"
+    end
+  end
+
+  def desired_bundler_version
+    @desired_bundler_version ||= begin
+      omnibus_overrides = File.join(project_root, "omnibus_overrides.rb")
+      File.readlines(omnibus_overrides).each do |line|
+        if line.start_with?("override :bundler,")
+          return line.scan(/"([^"]*)"/).flatten.first
```
jkeiser: These two lines may read cleaner as:

```
if line =~ /^override :bundler, version: "(.+)"$/
  return $1
end
```

mwrock:
after discussing with @jkeiser I replaced the `raise` if the user did not have the right version of bundler and instead now request the desired version of bundler to be loaded: `load Gem.bin_path("bundler", "bundle", desired_version)`. This has the added benefit of using the gem on the users machine if they have it but not requiring it to be the latest one.

This also adds some additional bundler/platform hacking to make sure bundler identifies the calls as coming from the intended platform.

jkeiser:
:+1: and thanks so much for your willingness to dig deep on this!

PR #4966, opened 2016-05-24
Fix portage provider to support version with character

crigor:
I ran into an issue where chef 12 installs openssl 1.0.2d even if it's already installed. It doesn't support the character on the version, `d`.

According to `man 5 ebuild`

> 
Versions  are  normally made up of two or three numbers separated by periods, such as 1.2 or 4.5.2.  This string may be followed by a character such as 1.2a or 4.5.2z.  Note that this letter is not meant to indicate alpha,  beta,  etc...  status.   For  that,  use  the  optional  suffix;  either  _alpha,  _beta,  _pre (pre-release),  _rc  (release candidate), or _p (patch).

I added a test which failed before I added the fix. `bundle exec rspec spec/unit/provider/package/portage_spec.rb ` shows passing tests.

This is my first PR on chef so let me know if I need to change anything.


lamont-granquist:

lamont-granquist:
👍

thommay:
👍

smurawski:
👍

PR #4967, opened 2016-05-24
add a few more distros to kitchen testing

lamont-granquist:
adds debian, centos-7 and fedora-23


lamont-granquist:
@chef/client-core expanding the test-kitchen testing in travis to a few more distros.

centos-5 and opensuse are the only distros i think that docker covers that we don't have covered.   centos-5 because its a bit of a yak to install epel and git before kitchen-appbundle-updater runs, and opensuse because i'm just a bit tired of looking at test-kitchen at this point...

smurawski:
👍

PR #4969, opened 2016-05-25
Remove the extra fetch operation from the git resource

joshburt:
It appears that with each fetch we are performing an extra unneeded fetch operation.

Within the code (https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L186) we are performing:

```
git fetch {remote}
git fetch {remote} --tags
git reset {remote} --hard 
```
According to the documentation on fetching: https://git-scm.com/docs/git-fetch
> --tags
> Fetch all tags from the remote (i.e., fetch remote tags refs/tags/* into local tags with the same name), in addition to whatever else would otherwise be fetched. Using this option alone does not subject tags to pruning, even if --prune is used (though tags may be pruned anyway if they are also the destination of an explicit refspec; see --prune).

Based on the above adding '--tags' not only fetches remote tags but also **whatever else would otherwise be fetched.**

This seems to imply that we don't need to perform the first fetch operation, and only need to perform the fetch which includes the '--tags'.

This PR removed the extra fetch and updates the tests accordingly.
Thanks!

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -183,7 +183,6 @@ class Chef
         converge_by("fetch updates for #{@new_resource.remote}") do
           # since we're in a local branch already, just reset to specified revision rather than merge
           Chef::Log.debug "Fetching updates from #{new_resource.remote} and resetting to revision #{target_revision}"
-          git("fetch", @new_resource.remote, cwd: cwd)
           git("fetch", @new_resource.remote, "--tags", cwd: cwd)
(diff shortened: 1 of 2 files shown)


thommay:
I'm not sure if this was true historically - my recollection is that older git versions only fetched tags with `--tags`

joshburt:
Then that would explain why we are performing this twice! :)  If we don't know for sure we should probably keep this around as the impact is small.

PR #4973, opened 2016-05-27
Creation of the new DataCollector reporter

adamleff:
The DataCollector reporter is a new method for exporting data about your
Chef run. The details of this new feature can be found in [RFC 077](https://github.com/chef/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md).

Using the existing `EventDispatch` mechanics, the DataCollector reporter collects data about a Chef run (when it starts, when it ends, what resources were modified, etc.) and then POSTs them to a Data Collector server URL that can be specified in your Chef configuration.

While similar functionality exists using the `ResourceReporter` and Chef Reporting, a new implementation was done to decouple the reporting of this data from requiring the use of a Chef Server (in the case of Chef Reporting), opening the door to users being able to implement their own webhook-style receiver to receive these messages and analyze them accordingly.


lib/chef/data_collector.rb
```diff
@@ -0,0 +1,277 @@
+#
+# Author:: Adam Leff (<adamleff@chef.io>)
+# Auther:: Ryan Cragun (<ryan@chef.io>)
```
stevendanna: Author :D

adamleff:
http://manhattan.ci.chef.co/job/chef-trigger-ad_hoc/22/downstreambuildview/

lib/chef/data_collector/serializers/base.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        def data_collector_organization
+          Chef::Config[:data_collector_organization] || "chef_solo"
+        end
+
+        def chef_server_organization
+          return nil unless Chef::Config[:chef_server_url]
+
+          Chef::Config[:chef_server_url].match(%r(/organizations/(\w+))).nil? ? "unknown_organization" : $1
```
stevendanna: You may want a `+` after the /'s

lib/chef/data_collector/serializers/base.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        def collector_source
+          solo_run? ? "chef_solo" : "chef_client"
+        end
+
+        def solo_run?
```
stevendanna: Not blocking, but I feel like this method would be helpful generally, maybe as a method on Chef::Config itself?

lib/chef/data_collector/serializers/base.rb
```diff
@@ -0,0 +1,101 @@
...
+          Chef::Config[:solo] || Chef::Config[:local_mode]
+        end
+
+        def node_uuid
```
stevendanna: Would it make sense for this to live on the Chef::Node object itself?  That object could then be passed into the serializer (or maybe is already accessible via the run context).

stevendanna:
I know there is an RFC for this, but for PRs this big it is often nice for the commit message to have a paragraph or two high level description of the overall design to aid with review and future debugging.

stevendanna:
The original RFC mentions an optional token, is that planned for a follow up PR?

lib/chef/data_collector/serializers/base.rb
adamleff: If others think this is a good idea, I'm happy to add it. We're being very opinionated here in that "solo" includes both "solo (legacy)" and "new solo (chef client local mode)". If others agree that we can categorize that all as "solo" I'm happy to add it.

lib/chef/data_collector/serializers/base.rb
adamleff: Would others find this helpful? To date, nothing else seemed to require the need to uniquely identify nodes like this, but because of how a Data Collector server could receive messages from a variety of different sources where node name + org isn't unique enough, I added this here and wanted to limit the blast radius. :)

adamleff:
@stevendanna re: commit message, I'll definitely update it and the PR description with more words, and also link to the RFC. Thanks for that suggestion.

adamleff:
@stevendanna Are you referring to the x-data-collector-token header? If so, that is implemented here:

https://github.com/chef/chef/blob/adamleff/data_collector/lib/chef/data_collector.rb#L215-L220

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,277 @@
...
+            exception
+          ).for_json
+        )
+      end
+
+      private
+
+      def disable_reporter_on_error
+        yield
+      rescue Timeout::Error, Errno::EINVAL, Errno::ECONNRESET, Errno::ECONNREFUSED, EOFError, Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, Net::ProtocolError => e
```
stevendanna: You may also want `OpenSSL::SSL::SSLError` for cases where the user has used HTTPs but something is wonky.

stevendanna:
> The original RFC mentions an optional token, is that planned for a follow up PR?

Nvm! Found it, somehow skimmed past it.

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,277 @@
...
+            updated_resources: updated_resources,
+            status: status,
+            error_descriptions: error_descriptions
+          )
+        )
+      end
+
+      def headers(additional_headers = {})
+        headers = { "Content-Type" => "application/json" }
+        headers["x-data-collector-token"] = data_collector_token unless data_collector_token.nil?
```
stevendanna: There was some discussion about this in the PR for the RFC where Adam mentioned:

> this should be a shared secret that gets hashed based on the payload. 

There was also the sense that we should version the protocol somehow.  I'm personally fine with the auth scheme being figured out in follow up PRs, but if we think we are gonna ship a version with it in the current state we might want to emit a warning to anyone who sets it that the auth is still in flux.

adamleff: Could you give an example? This feels like it would be one of those deprecation warnings that people either ignore or are annoyed by when a different option doesn't exist and it's performing "as designed right now." Thoughts?

adamleff:
@stevendanna I have updated the PR description. When I squash/rebase prior to merge, I'll also reword the top-level commit. I just don't want to do that yet lest any commit/line comments get lost in the shuffle.

lib/chef/data_collector.rb
stevendanna: Eh, I guess I'm probably worried about a thing that isn't going to happen.  It would just be unfortunate if someone wrote an implementation against this and then we changed it on them.

adamleff: I'd actually propose that for many users, this token header is "good enough" and, if we do implement something new, it should be an either/or decision rather than us ripping this option away from them. Users can either choose the new auth hawtness, or continue using a simple PSK.

stevendanna:
Overall I think this looks good.  The "Serializers" class feels like it is more accurately a set of value objects for the various DataCollector message types which then have serialization methods on them, but that is just a naming bikeshed, not likely worth changing.

My only hesitation is that the token header doesn't seem fully fleshed out.  As I mentioned in my inline comment, maybe I'm worrying about a case that doesn't matter since the reality is that we will have the only server-side implementation for this until it settles.  However, I don't love sending the token itself, especially since we don't require HTTPs. Personally, I think the header should be something like:  

```
X-Data-Collector-Auth: sig_version=1.0 sig=SIG
```

where `SIG` is something like `OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'), data_collector_token, message_body)`.

stevendanna:
To be clear, I'm OK with this going in without my token concerns addressed.  We should just make it clear in the docs that users should be using this over HTTPs if they are using a server implementation that uses the token.

chef-config/lib/chef-config/config.rb
```diff
@@ -796,6 +796,15 @@ def self.windows_home_path
     config_context :chefdk do
     end
 
+    # Configuration options for Data Collector reporting. These settings allow
+    # the user to configure where to send their Data Collector data, what token
+    # to send, and whether Data Collector should report its findings in client
+    # mode vs. solo mode.
+    default :data_collector_server_url,       nil
```
jkeiser: It might be better for our namespaces to do this:

```
config_context :data_collector do
  default :server_url, nil
  default :token, nil
  default :mode, :both
  default :raise_on_failure, false
end
```

They can then be set in the config file with

```
data_collector.server_url = "https://..."
data_collector.mode = :both
```

etc.

Also, please comment new config values with acceptable values and what they affect.

acceptance/data-collector/test/integration/default/serverspec/default_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+end
+
+describe "CCR, local mode, config in client mode" do
+  include_examples "successful chef run", "chef-client -z -c /etc/chef/client-mode.rb"
+  include_examples "counter checks", { "run_start" => nil, "run_converge.success" => nil, "run_converge.failure" => nil }
+end
+
+describe "CCR, local mode, config in both mode" do
+  include_examples "successful chef run", "chef-client -z -c /etc/chef/both-mode.rb"
+  include_examples "counter checks", { "run_start" => 1, "run_converge.success" => 1, "run_converge.failure" => nil }
```
jkeiser: Is it OK that we're not testing the actual payloads? Like, I'm OK with this as a start and won't block checkin, but it feels like we need to test that the stuff being sent is something a server should expect, not just that something was sent.

adamleff: I feel like we'd be dangerously close to duplicating the unit tests I've written that already do this.

lib/chef/client.rb
```diff
@@ -415,6 +417,13 @@ def register_reporters
       end
     end
 
+    # Register the data collector reporter to send event information to the
+    # data collector server
+    # @api private
```
jkeiser: Can you actually put this in the `private` section rather than making it `@api private`? I think the existing APIs we kept public in case someone was using, but we don't want the new APIs to be used.

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,279 @@
...
+
+require "uri"
+require "chef/event_dispatch/base"
+require "chef/data_collector/resource_report"
+require "chef/data_collector/serializers/node_update"
+require "chef/data_collector/serializers/run_end"
+require "chef/data_collector/serializers/run_start"
+
+class Chef
+  class DataCollector
```
jkeiser: Please comment new classes and methods if public, or mark @api private / private if not.

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,279 @@
...
+    class Reporter < EventDispatch::Base
+      attr_reader :updated_resources, :status, :exception, :error_descriptions,
+                  :expanded_run_list, :run_status, :http, :resource_count,
+                  :current_resource_report, :enabled
+
+      def initialize
+        @updated_resources       = []
+        @resource_count          = 0
+        @current_resource_loaded = nil
+        @status                  = "success"
```
jkeiser: It seems strange for status to default to success before the report is finished. Is there an event or events which actually denote success, and can we mark it `nil` or `"unknown"` until then?

lib/chef/data_collector/serializers/base.rb
```diff
@@ -0,0 +1,101 @@
...
+        def organization
+          solo_run? ? data_collector_organization : chef_server_organization
+        end
+
+        def data_collector_organization
+          Chef::Config[:data_collector_organization] || "chef_solo"
+        end
+
+        def chef_server_organization
+          return nil unless Chef::Config[:chef_server_url]
```
jkeiser: Did you intend this to be `nil` or `unknown_organization`?

lib/chef/data_collector/serializers/base.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        def to_json
+          document.to_json
+        end
+
+        def chef_server_fqdn
+          return URI(Chef::Config[:chef_server_url]).host unless Chef::Config[:chef_server_url].nil?
+          return "localhost" unless defined?(run_status)
+
+          run_status.node["fqdn"]
```
jkeiser: Should we default to "localhost" if there is no fqdn in the node?

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,279 @@
...
+        http.post(nil, message.to_json, headers)
+      end
+
+      def send_run_completion
+        # If run_status is nil we probably failed before the client triggered
+        # the run_started callback. In this case we'll skip updating because
+        # we have nothing to report.
+        return unless run_status
+
+        send_to_data_collector(Serializers::NodeUpdate.new(run_status))
```
jkeiser: This will do the serialization work (including uuid generation and metadata caching) even `data_collector_accessible?` is false. Is that intended / acceptable?

acceptance/data-collector/test/integration/default/serverspec/default_spec.rb
jkeiser: The integration tests are the more important ones either way, IMO.

adamleff: I've added some payload tests to the acceptance tests. :)

adamleff:
@chef/client-core can I please have a review of this? I've got +1s from @stevendanna and @jkeiser but need maintainers to review. Thank you!

tyler-ball:
@adamleff I tried running the acceptance tests locally and got a bunch of failures. It looks like the Sinatra app is not starting correctly, but I'm not 100% sure about that

adamleff:
@tyler-ball silly question, but did you configure the right environment variables to pull a recent unstable build that has the new code in it before running the acceptance tests? For example:

```
export KITCHEN_CHEF_VERSION="12.11.15+20160601183514"
export KITCHEN_CHEF_CHANNEL=unstable
```

lib/chef/client.rb
```diff
@@ -957,6 +959,12 @@ def has_admin_privileges?
 
       Chef::ReservedNames::Win32::Security.has_admin_privileges?
     end
+
+    # Register the data collector reporter to send event information to the
+    # data collector server
+    def register_data_collector_reporter
+      events.register(Chef::DataCollector::Reporter.new) if Chef::DataCollector.register_reporter?
```
tyler-ball: Instead of making a new method I think it makes more sense to just move this into the [`register_reporters`](https://github.com/chef/chef/blob/247bfc6c26520aa0e46332b13dc98f77d22b95e5/lib/chef/client.rb#L411) method

adamleff: `register_reporters` is not called if you're in Chef Solo legacy mode, and this feature needs to work with all run-types (Solo Legacy, Solo / Client Local Mode, and Client/Server mode). That's why we broke it up and call this method directly in `#run`.

lib/chef/data_collector.rb
```diff
@@ -0,0 +1,342 @@
...
+      def increment_resource_count
+        @resource_count += 1
+      end
+
+      def add_updated_resource(resource_report)
+        @updated_resources << resource_report
+      end
+
+      def disable_data_collector_reporter
+        # TODO: should the event dispatcher support de-registering?
```
tyler-ball: I'm confused - what does this `TODO` mean?

adamleff: If we have trouble reaching the DataCollector endpoint at the start of the run, we make the decision to not try and reach it at the end of the run either. However, throughout the run, The Event Dispatcher is still sending events to this instance, which is collecting data and stuff, and then at the end of the run, just decides not to send it.

The thought was perhaps it would be nice if there was a way for a event handler to unregister itself so the Chef run isn't still invoking it for all the events triggered. It's a nice-to-have, but not a necessity. I can remove the TODO for clarity as it's out-of-scope for this change if you think that's prudent.

tyler-ball: Yeah, I say remove the TODO - we can address that if it becomes an issue. I don't know of anywhere we de-register handlers currently.

acceptance/data-collector/.acceptance/data-collector-test/recipes/default.rb
```diff
@@ -0,0 +1,38 @@
...
+pid_file    = "/var/run/api.pid"
+running_pid = ::File.exist?(pid_file) ? ::File.read(pid_file).strip : nil
+
+execute "kill existing API process" do
+  command "kill #{running_pid}"
+  not_if { running_pid.nil? }
+end
+
+execute "start API" do
+  command "bundle exec rackup -D -P #{pid_file}"
```
tyler-ball: This can just be `bin/rackup` since you did a `--binstubs` earlier

lib/chef/data_collector/messages.rb
```diff
@@ -0,0 +1,125 @@
...
+      # @param run_status [Chef::RunStatus] The RunStatus instance for this node/run.
+      #
+      # @return [Hash] A hash containing the run start message data.
+      #
+      def self.run_start_message(run_status)
+        {
+          "chef_server_fqdn"  => chef_server_fqdn(run_status),
+          "entity_uuid"       => node_uuid,
+          "id"                => run_status.run_id,
+          "message_version"   => "1.0.0",
```
tyler-ball: This should definitely be refactored into a constant instead of being duplicated in all the messages

tyler-ball: Ohhhhh, nevermind - I see these are unique per message. 😶 🌴

adamleff: Yup - we want to be able to iterate and bump them individually.

tyler-ball:
A few minor comments but 👍 once those are responded to or addressed. Nice work!

adamleff:
Thanks for taking a look, @tyler-ball! I've addressed and/or responded to all your concerns.

thommay:
👍 Looks good!

lib/chef/data_collector.rb
lamont-granquist: Yeah, I tend to feel like this will almost certainly change, and it should at least bake the protocol version into it for when it does.

lib/chef/data_collector/messages.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      #
+      # Message payload that is sent to the DataCollector server at the
+      # start of a Chef run.
+      #
+      # @param run_status [Chef::RunStatus] The RunStatus instance for this node/run.
+      #
+      # @return [Hash] A hash containing the run start message data.
+      #
+      def self.run_start_message(run_status)
```
lamont-granquist: these should probably be their own classes and you should create instances of them.  even if currently all its /currently/ doing is creating a short-lived garbage object that gets immediately thrown away.

adamleff: @lamont-granquist That's how I originally had them and @jkeiser strongly suggested that I move to this model.

lib/chef/data_collector.rb
adamleff: OK, if I add a new header:

```
X-Data-Collector-Auth: version=1.0
```

... and leave the existing token header, will that satisfy everyone's concerns? When we decide to do signed payloads, we can bump the version and include the signature in the header, deprecating the token header (i.e. servers will not read the token header unless the version == 1.0)?

lamont-granquist: yeah, that'd be good enough for me.

lib/chef/data_collector/messages.rb
lamont-granquist: well, for the record i tend to disagree.  we can create other objects, though, which just call the methods here.

lib/chef/data_collector/messages/helpers.rb
```diff
@@ -0,0 +1,161 @@
...
+        #
+        # @param run_status [Chef::RunStatus] The RunStatus object for this Chef Run.
+        #
+        # @return [String] FQDN of the configured Chef Server, or node/localhost if not found.
+        #
+        def chef_server_fqdn(run_status)
+          if !Chef::Config[:chef_server_url].nil?
+            URI(Chef::Config[:chef_server_url]).host
+          elsif run_status
+            run_status.node["fqdn"]
```
lamont-granquist: node["fqdn"] is not guaranteed to be defined.  if someone has broken dns it'll be nil.  this happens a lot -- guaranteed bug factory here.

you should either use the node_name, or else you should use `ohai[:fqdn] || ohai[:machinename] || ohai[:hostname]` out of this expression:

https://github.com/chef/chef/blob/master/lib/chef/client.rb#L579

lamont-granquist: or you should always fall through to "localhost" if node["fqdn"] is nil or something...

lib/chef/data_collector/messages/helpers.rb
```diff
@@ -0,0 +1,161 @@
...
+        # has no knowledge of an organization or to which organization is belongs.
+        #
+        # If we cannot determine the organization, we return "unknown_organization"
+        #
+        # @return [String] shortname of the Chef Server organization
+        #
+        def chef_server_organization
+          return "unknown_organization" unless Chef::Config[:chef_server_url]
+
+          Chef::Config[:chef_server_url].match(%r{/+organizations/+(\w+)}).nil? ? "unknown_organization" : $1
```
lamont-granquist: should this be "_default"?

lamont-granquist:
👍

lib/chef/data_collector/messages/helpers.rb
lamont-granquist: ^ to clarify my thoughts there...

node_name is solving a problem of reliably finding a name for a thing so it uses an algorithm which is guaranteed to not wind up being nil.

you have a problem of needing a name which will connect to a server, which is actually different, so the node_name algorithm probably is bad.  if someone has busted DNS its likely that not only will node['fqdn'] be nil, but node['machinename'] and node['hostname'] won't be any more useful to you.

but it should still have a shape more like `URI(Chef::Config[:chef_server_url]).host || node['fqdn'] || 'localhost'`

adamleff: In actuality, we're not using this value to connect to anything... it's simply to report what chef server (if in client mode) we used... otherwise, what host are we on. `node_name` may actually do the trick here.

PR #4977, opened 2016-05-27
fixes #4968 and only retrieves the latest version of packages from chocolatey

### LINKED ISSUE #4968: Chocolatey package incorrectly identifies latest version of package.
Please see issue created on wrong recipe here
https://github.com/chocolatey/chocolatey-cookbook/issues/86#issuecomment-221428759

mwrock:
This should fix retrieving the wrong package version when expecting the latest and should also significantly improve package query performance. It also supports some feeds that do not implement the all versions capability.

We now only query the chocolatey feed for the latest version and use that as the candidate version unless an explicit version is requested and then that is the candidate.

This does prevent us from validating valid versions and allows a resource with a bogus version. In these cases we will bubble up the error from choco.exe.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -229,13 +229,11 @@ EOS
         def available_packages
           @available_packages ||=
             begin
-              cmd = [ "list -ar #{package_name_array.join ' '}" ]
+              cmd = [ "list -r #{package_name_array.join ' '}" ]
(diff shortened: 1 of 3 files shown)


mwrock:
cc @lamont-granquist @ferventcoder

lamont-granquist:
LGTM 👍

smurawski:
👍

ferventcoder:
👍

PR #4979, opened 2016-05-28
Fix #4949 and Avoid Errno::EBUSY on docker containers

### LINKED ISSUE #4949: Automatically set atomic_updates to false in Chef::Provider::File when updating /etc/resolv.conf or /etc/hosts on a docker guest?
These give EBUSY errors on docker.

Fixing them in cookbooks is a bit of a whack-a-mole.

The worst things that I can think of happening is:

- docker detection code fails to detect docker, so the updates fail, in which case manually setting atomic_updates to false works, and the user isn't much worse off than they are now.

- docker no longer requires these settings, in which case its still going to work, outside of the edge condition where you happen to run the server out of disk space right when you're doing this (but its docker, so immutable infrastructure magically makes this problem go away #amirite?)

It does mean that to fix bugs in the docker detection code you need to upgrade chef and not just upgrade a cookbook.  This doesn't feel like docker is going to mess with this that much though.

andrewjamesbrown:
This commit fixes https://github.com/chef/chef/issues/4949 by adding
ohai-based docker detection code and statically checking for
/etc/hostname or /etc/resolv.conf.

cc @lamont-granquist

### DIFF
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -67,13 +67,7 @@ class Chef
       def initialize(new_resource, run_context)
         @content_class ||= Chef::Provider::File::Content
         if new_resource.respond_to?(:atomic_update)
-          # Resolves https://github.com/chef/chef/issues/4949
-          if docker_guest?(run_context.node) && (new_resource.path == "/etc/resolv.conf" ||
-                                                 new_resource.path == "/etc/hostname")
-            @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(false)
-          else
-            @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(new_resource.atomic_update)
-          end
+          @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(new_resource.atomic_update)
         end
         super
       end
@@ -198,12 +192,6 @@ class Chef
       end
 
       private
-
-      # Return true if we are running in a docker guest, or false otherwise.
-      def docker_guest?(node)
-        node[:virtualization] && node[:virtualization][:systems] &&
-          node[:virtualization][:systems][:docker] && node[:virtualization][:systems][:docker] == "guest"
-      end
 
       # What to check in this resource to see if we're going to be actively managing
       # content (for things like doing checksums in load_current_resource).  Expected to
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -67,7 +67,13 @@ class Chef
       def initialize(new_resource, run_context)
         @content_class ||= Chef::Provider::File::Content
         if new_resource.respond_to?(:atomic_update)
-          @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(new_resource.atomic_update)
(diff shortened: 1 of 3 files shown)


lib/chef/provider/file.rb
```diff
@@ -193,6 +199,12 @@ def manage_symlink_access?
 
       private
 
+      # Return true if we are running in a docker guest, or false otherwise.
+      def docker_guest?(node)
```
lamont-granquist: ideally i'd like to see `#docker?` pulled into Chef::DSL::Universal from chef-sugar:  

https://github.com/sethvargo/chef-sugar/blob/b9b3b643c27a0828f56d7355bb4d05577bd90a38/lib/chef/sugar/docker.rb

lamont-granquist: (i don't know exactly what to do about back-compat with chef-sugar and how to deal with the way that chef-sugar has that method which takes a node as an argument or what that use case covers...)

lib/chef/provider/file.rb
```diff
@@ -67,7 +67,13 @@ class File < Chef::Provider
       def initialize(new_resource, run_context)
         @content_class ||= Chef::Provider::File::Content
         if new_resource.respond_to?(:atomic_update)
-          @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(new_resource.atomic_update)
+          # Resolves https://github.com/chef/chef/issues/4949
+          if docker_guest?(run_context.node) && (new_resource.path == "/etc/resolv.conf" ||
+                                                 new_resource.path == "/etc/hostname")
+            @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(false)
+          else
+            @deployment_strategy = Chef::FileContentManagement::Deploy.strategy(new_resource.atomic_update)
```
lamont-granquist: this could probably look more like:

```ruby
atomic_updates = special_docker_files? ? false : new_resource.atomic_updates
@deployment_strategy = Chef::FileContentManagement::Deploy.strategy(atomic_updates)
```

also should probably compare against `Pathname.cleanpath(new_resource.path)` and/or use `File.expand_path()`

coderanger: I feel like an explicit value in `new_resource.atomic_updates` should always override, just that the default becomes path-dependent.

lamont-granquist: yeah, probably should.

andrewjamesbrown: Added this behaviour with d801443.

andrewjamesbrown:
I've updated the implementation so that `new_resource.atomic_updates` always overrides the default, and also moved the implementation of `docker?` into `lib/chef/dsl/platform_introspection.rb`.  I wasn't able to use the same code as chef-sugar, as Travis was complaining bitterly about the `File.exists?` and mocking that everywhere with docker-specific code didn't seem to be the best option, so I'm just looking at Ohai data.

lamont-granquist:
mmm yeah, file provider unit tests and mocking `File.exists?`...   there be dragons...

lamont-granquist:
👍

smurawski:
👍

Issue #4978, opened 2016-05-27
resource_block to lazy a notifying_block


lamont-granquist:
notifying_block is a DSL method so at compile-time it creates a subresource collection and eagerly converges it.

to lazy the execution of the subcollection we could offer a primitive like this:

``` ruby
resource_block "stuff" do
  block do
    template "/etc/whatever/thing.conf1" do
      source "thing.conf1.erb"
      notifies :restart, "service[whatever]"
    end
    template "/etc/whatever/thing.conf2" do
      source "thing.conf2.erb"
      notifies :restart, "service[whatever]"
    end
    service "whatever" do
      action :start
    end
  end
  action :run
end
```

of course i believe this is entirely equivalent to:

``` ruby
whyrun_safe_ruby_block "stuff" do
  block do
    nofiying_block do
      template "/etc/whatever/thing.conf1" do
        source "thing.conf1.erb"
        notifies :restart, "service[whatever]"
      end
      template "/etc/whatever/thing.conf2" do
        source "thing.conf2.erb"
        notifies :restart, "service[whatever]"
      end
      service "whatever" do
        action :start
      end
    end
  end
  action :run
end
```

But I can potentially see a lot of repetitive conversations about the 'right' way to do the latter and it might be simpler to just be able to say `use a resource_block, it does exactly what you need...`

lamont-granquist:
@coderanger thoughts?  up? down?  meh?

coderanger:
I don't love the syntax :-/

I think something like this is probably more grokkable:

```
recipe 'whatever' do
 template '...' do end
  file '...' etc etc
end
```

This is further from the usual resource syntax but probably more easily understandable in the short term.

coderanger:
I'm also not totally sold that this is enough better than the existing LWRP stuffs to merit syntax.

lamont-granquist:
Yeah, I was thinking about going down that road.    The thing in the block doesn't quite act like a resource when it comes to the content of the block though.  Instead of instance_eval'ing the block against the resource, we probably need to have a special option to save the whole block for use by the provider later.   Not that hard to code, but it starts to abuse the 'shape' of what a resource is.

lamont-granquist:
closing this since we closed #4982

PR #4982, opened 2016-05-31
add recipe resource

lamont-granquist:
🚧


coderanger:
What's the end goal of this? Just being able to run a sequence of resources from one notification?

lamont-granquist:
@coderanger this is an implementation of #4978 with your syntax.

reusing LWRPBase also gets the enclosing_provider correct and gets the resource DSL mixed into the provider and other things.  it needs more than just a notifying_block.

it does not (and cannot?) support not_if/only_if or notifies/subscribes on the recipe itself.

there was a recipe ivar accessor on Chef::Recipe which trolled me, and i nuked it and i think it was not being used at all...

coderanger:
Yeah, I'm trying to get a zoom'd out view of what this will be for. It feels like something living between several use cases maybe?

lamont-granquist:
its the resource grouping with its own delayed notification block that i see as the most useful:

```ruby
recipe "stuff that goes together" do
  template "/etc/whatever/conf1" do
    source "conf1.erb"
    notifies :restart, "service[whatever]"
  end
  # we now may be in a completely inconsistent state -- we cannot have the previous notify be :immedately or we may crash a running service.
  template "/etc/whatever/conf2" do
    source "conf2.erb"
    notifies :restart, "service[whatever]"
  end
  # okay now we're good, but we can't count on conf1 and conf2 always being updated at the same time, so while conf2 could notify immediately
  # there's still the case where conf1 updates where a :delayed notification off of conf1 will be too late, but :immedate will be too early.
  service "whatever" do
    action :start
  end
  # because this is in a sub-resource context the delayed notifications will all run "here"
end

execute "whateverdb add tables" do
  command "whatever add tables to things and this requires a running whatever service"
end
```

Also I imagine people could have uses for grouping resources and sending them notifications.

lamont-granquist:
there's the shitty pattern of using the log resource to collect notifications and then fire off multiple resources in parallel, which i think this could replace -- although the fact that you can't hang a notify or subscribes off of a recipe resource written this way and you always have to invert it is a bit annoying...

coderanger:
Delayed notifications is a good enough argument for it in my book :+1:. I've talked with @jkeiser before about using this syntax for parallelism in the future (`recipe('foo', parallel: true)` or something) so having the base semantics in seems like a good step, and then we can add more options and modes.

lamont-granquist:
yeah, we've been kicking this idea around for awhile now, seems like this would be an iterative step forwards...

coderanger:
I'm willing to bet this will break at least some LWRPs that use `Recipe` instead of `Chef::Recipe` because wooo Ruby relative const lookups, but I will shed 0 tears over this.

jaymzh:
This is kinda like what we were talking about when we talked about making the recipe DSL available to ruby blocks... this solves the notify problem you mentioned, but *also* solves the problem if wanting to schedule a runtime thing that uses resources, so that's kinda neat. Ultimately though, this is just an inline syntax for custom resource creation-and-calling, afaict.

jaymzh:
also, why can't this support not_if/only_if ? it's just a resource like any other, right?

coderanger:
@jaymzh Because we're using the whole "resource block" bit as the recipe DSL. The block won't be evaluated until the action is run.

jaymzh:
couldn't we check it manually in action :run ?

lamont-granquist:
you're not going to parse the not_if/only_if until you've executed all the statements before it.

i guess we could have those mean "conditionally abort the converge phase" with the understanding that the compile phase has to happen.   that would take some monkeying about with how it fires of the sub-resource compile-and-converge phases though.

notifies could also conceivably be made to work as well, just wire up the notification to the enclosing provider's resource and then when it gets updated_by_last_action then it'll fire.

subscribes can't really work though because without parsing the block when the provider is already firing in converge mode of the outer resource, you aren't going to know about the subscribes to send notifications to it from previously declared resources.

one thing we could do is wire up both options (this way of writing it and the way with the resource that takes `block do ... end` as a property) and name them something slightly longer, then just see which one eventually 'wins' and we promote to the `recipe` resource after there's a bit of consensus.

that means there might be some unnecessary code we eventually need to deprecate, but that's better that getting stuck with the API we don't want for `recipe` blocks and wind up wanting to change the API and break every early adopter...

lamont-granquist:
@chef/client-core I think what I'd like to propose here is that we merge the `recipe` resource as-is.

It gives the best experience for "i just want a group of resources to run" without a lot of clutter.

I think when and if we see the requests start coming in for things that this recipe resource does not support, then we introduce a `recipe_block` resource which takes a block and looks like:

```ruby
recipe_block "whatever" do
  block do
    file "/etc/whatever.conf"
    ....
  end
  not_if { ... }
  only_if { ... }
  notifies ...
  subscribes ...
end
```

But I don't think we expand the scope until we see the bug report come in.

lamont-granquist:
should probably rename this to "resource_group" instead of "recipe" since that overloads the meaning.  also "resource_block" could be used for the other thing that takes a block if we eventually need to build that.

PR #4991, opened 2016-06-03
Don't mask directory deletion errors

### LINKED ISSUE #4988: Recursive directory failure ignores failure
## Description

when deleting a directory using the `directory` resources, if `recursive` is set, chef does:

```
FileUtils.rm_rf(@new_resource.path)
```

Which ignores all errors, which means if this fails Chef falsely reports success.

Instead Chef should use ```FileUtils.rm_r(@new_resource.path)``` and catch errors.

You can simulate this by `mkdir -p /tmp/bananas; chattr +i /tmp/bananas` and then removing it with Chef. It'll fail but Chef will say it succeeded leading to unpredictable results later in the run. This is bad news bears.

## Chef Version

12.6.0

## Platform Version

CentOS 6 & 7

## Replication Case

mkdir -p /tmp/bananas
chattr +i /tmp/bananas
Run chef with `directory '/tmp/bananas' do action :delete, recursive true end`
notice /tmp/bananas is still there.

jaymzh:
closes #4988

### DIFF
--- a/lib/chef/provider/directory.rb
+++ b/lib/chef/provider/directory.rb
@@ -145,7 +145,9 @@ class Chef
         if ::File.exists?(@new_resource.path)
           converge_by("delete existing directory #{@new_resource.path}") do
             if @new_resource.recursive == true
-              FileUtils.rm_rf(@new_resource.path)
+              # we don't use rm_rf here because it masks all errors, including
+              # IO errors or permission errors that would prvent the deletion
+              FileUtils.rm_r(@new_resource.path)
               Chef::Log.info("#{@new_resource} deleted #{@new_resource.path} recursively")
             else
               ::Dir.delete(@new_resource.path)
--- a/spec/unit/provider/directory_spec.rb
+++ b/spec/unit/provider/directory_spec.rb
@@ -227,7 +227,7 @@ describe Chef::Provider::Directory do
     end
   end
 
-  describe "#run_action(:create)" do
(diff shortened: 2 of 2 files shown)


jaymzh:
CC @chef/client-core

lamont-granquist:
yeah, that makes sense :+1:

thommay:
seems good 👍

btm:
Can you add a regression unit test to make sure we don't revert back and fix the existing copy-pasta on line 230 of the unit test while you're in there?


```
  describe "#run_action(:create)" do
    describe "when the directory exists" do
      it "deletes the directory" do
        directory.run_action(:delete)
        expect(File.exist?(tmp_dir)).to be false
      end
```

jaymzh:
The typo is easy - but I have no idea how to write a test to simulate the OS failing the delete... :(

btm:
being a unit test I was just thinking something like 

```
it "does not use rm_rf which silently consumes errors" do
  expect(FileUtils).not_to receive(:rm_rf)
end
```

jaymzh:
Oh. I can do that. sure :)

jaymzh:
Tests fixed/added.

btm:
:+1: on green

smurawski:
The appveyor build failure is not related - I looks like the appveyor build images just got a newer version of powershell which requires a different key usage type for the cert for password decryption.

PR #4994, opened 2016-06-04
WIP: Release changelog

mwrock:


coderanger:
Can we filter out stuff that's not relevant to users? We've had a lot of churn on things like Gemfile changes that are more noise than signal in here. I understand we want to auto-generate the changelog, but I think it's resulting in an increasingly unusable document.

mwrock:
Yup. Note this is WIP. I'm just getting started here :)

mwrock:
ok maybe not perfect but WIP removed

coderanger:
:+1:

smurawski:
👍

PR #4998, opened 2016-06-06
Skip `ps_credential` test for `dsc_script` until proper test cert

smurawski:


spec/unit/resource/dsc_script_spec.rb
```diff
@@ -71,6 +71,7 @@
     end
 
     it "has the ps_credential helper method" do
+      skip
```
thommay: is it possible to do `skip("until we get a new cert")` the same way you can with `pending`? If so, please update with the reason. (and a date or something)

thommay:
👍

mwrock:
:+1:

adamedx:
👍

PR #5006, opened 2016-06-07
Bug fix: updated_resource_count should only include updated resources

adamleff:
After adding up-to-date and skipped resources to the DataCollector
resource list, the logic that computed the number of resources
updated needed to change accordingly. This change fixes that
bug and also eliminates an unnecessary counter that tracks
the total number of resources.

### DIFF
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -51,13 +51,12 @@ class Chef
     # and exports its data through a webhook-like mechanism to a configured
     # endpoint.
     class Reporter < EventDispatch::Base
-      attr_reader :updated_resources, :status, :exception, :error_descriptions,
-                  :expanded_run_list, :run_status, :http, :resource_count,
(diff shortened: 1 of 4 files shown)


jkeiser:
@adamleff it looks good, but I don't see a test that confirms the behavior change? Did I miss it?

adamleff:
@jkeiser yup, you missed it. :)

https://github.com/chef/chef/blob/c4cf0bbf1836fb70b0dc3aaf2f5df0399f5633ee/spec/unit/data_collector/messages_spec.rb#L121-L125

adamleff:
@chef/client-core please review - thanks!

mwrock:
:+1: I think the test failure is unrelated

adamleff:
Agreed re: test failure. Repeatedly fails on fedora when trying to build nokogiri, likely due to latest nokogiri release. I'm going to mash the merge button now. Thanks gentlemen!

PR #5014, opened 2016-06-09
Updating omnibus dependencies to take advantage of new msys2 toolchain

tyler-ball:


kitchen-tests/cookbooks/base/Berksfile.lock
```diff
@@ -0,0 +1,64 @@
+DEPENDENCIES
```
ksubrama: We shouldn't be adding these right?

kitchen-tests/cookbooks/webapp/Berksfile.lock
```diff
@@ -0,0 +1,68 @@
+DEPENDENCIES
```
ksubrama: And this one

tyler-ball:
@ksubrama I removed the `Berksfile.lock` files I added, since we don't lock anything that isn't going through CI

ksubrama:
:+1:

ksubrama:
The travis failure is expected - it's warning us of a vulnerability in nokogiri...  and we are working around that for now.

danielsdeleo:
👍 but let's not let the nokogiri thing fester

mwrock:
:+1: agreed @danielsdeleo I have a nokogiri branch in the wings but like the idea of getting things green before adding that in.

Issue #4728, opened 2016-03-21
Failure to post the audit report should not be an error when running from Test Kitchen


tpetchel:
## Description

When developing a cookbook for use with audit mode, you may want to first run it through Test Kitchen. You'll likely get an ERROR message when chef-client attempts to post the audit report to the  server:

```
[...]
       Synchronizing Cookbooks:
         - audit_web_rhel (0.1.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       Starting audit phase

       Finished in 0.00021 seconds (files took 0.46834 seconds to load)
       0 examples, 0 failures
       Auditing complete

       Running handlers:
       Running handlers complete
       Chef Client finished, 0/0 resources updated in 01 seconds
       [2016-03-21T13:32:43+00:00] ERROR: Failed to post audit report to server. Saving report to /tmp/kitchen/cache/failed-audit-data.json
       Finished converging <default-centos-72> (0m22.46s).
-----> Kitchen is finished. (2m15.26s)
```

I feel that ERROR is not the correct severity for this case because I don't expect there to be a server. The ERROR leads me to believe that something failed in one of my audit controls.

Perhaps WARN or INFO would be more appropriate here, or skip the upload process if if we can detect that we're running through Test Kitchen.

## Chef Version

12.8.1

## Platform Version

Discovered on CentOS 7.2, but applies to any platform.

## Replication Case

Step 1: Create a cookbook.

```
$ chef generate cookbook TEST
```

Step 2: Modify .kitchen.yml to enable audit mode.

```
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  client_rb:
    audit_mode: :audit_only

platforms:
  - name: centos-7.2

suites:
  - name: default
    run_list:
      - recipe[TEST::default]
    attributes:
```

Step 3: Run `kitchen converge`.

```
$ kitchen converge
```

## Client Output

https://gist.github.com/tpetchel/69cc5f17692ec93e174b

## Stacktrace

N/A

thommay:
This seems more like a test-kitchen enhancement (add `enable_reporting = false` to the generated config.rb) than a chef-client bug. Please file this request there.

erichelgeson:
The bug report though explicitly says they enabled reporting in their `client_rb:`

In my scenario we run chef-zero on our machines (not in tk), with audit reporting enabled and see this ERROR each run.

[...]

PR #5016, opened 2016-06-10
Don't log error for reporting audit data in when in chef-zero

erichelgeson:
This changes the log message to `debug` when running in local-mode as it's not an error that the audit-data could not be sent to the server.

I'm not very familiar with the internals of `chef/chef` so any feedback is appreciated.

Fixes https://github.com/chef/chef/issues/4728

### DIFF
--- a/lib/chef/audit/audit_reporter.rb
+++ b/lib/chef/audit/audit_reporter.rb
@@ -140,7 +140,11 @@ class Chef
               # Save the audit report to local disk
               error_file = "failed-audit-data.json"
               Chef::FileCache.store(error_file, Chef::JSONCompat.to_json_pretty(run_data), 0640)
-              Chef::Log.error("Failed to post audit report to server. Saving report to #{Chef::FileCache.load(error_file, false)}")
+              if Chef::Config.chef_zero.enabled
+                Chef::Log.debug("Saving audit report to #{Chef::FileCache.load(error_file, false)}")
+              else
+                Chef::Log.error("Failed to post audit report to server. Saving report to #{Chef::FileCache.load(error_file, false)}")
+              end
             end
           else
             Chef::Log.error("Failed to post audit report to server (#{e})")


lib/chef/audit/audit_reporter.rb
```diff
@@ -138,9 +138,13 @@ def post_auditing_data
               return
             else
               # Save the audit report to local disk
-              error_file = "failed-audit-data.json"
+              error_file = "audit-data.json"
```
lamont-granquist: can we just leave the reporting name the way it was and not change this "api"?  in production when you're likely to care, it really will be a "failed-audit-data.json" file and on chef-zero/zolo it won't matter and you won't care...

smurawski:
I'm in favor of this PR, but I'd like to see one more spec test validating that we get the debug logging when local mode.

lib/chef/audit/audit_reporter.rb
erichelgeson: Right, the "api" would change, but the file name makes me think the audit failed, instead of failed to post. I'll defer to you if this should be changed back, but I think it's a change for the better.

erichelgeson:
I'll try to write up that test. I couldn't figure out how to check the debug logging in a test - but I'll look on monday and maybe jump on IRC if I can't find figure it out. Was on vacation last week :) Thanks for reviewing this PR!

lib/chef/audit/audit_reporter.rb
lamont-granquist: Yeah, I tend to disagree, I think the old name was fine for the people who will actually need to consume it.  I think its only because you're seeing it when you don't want it that you're getting worried about the name.  It also goes along nicely with the error message "failed to post audit..." so the meaning of failed-audit-data should be clear.

lamont-granquist:
still just waiting for the un-fix to the filename..

erichelgeson:
Done. Thanks!

thommay:
Appveyor failure is "expected". This looks good to go 👍

lamont-granquist:
👍

PR #5019, opened 2016-06-10
bug fix: correctly write out data collector metadata file

adamleff:
When we changed the messages from class instances to a module, the
reading of metadata into an instance variable was removed and
introduced a bug when updating metadata. This bug causes the
metadata to be properly read from disk, updated, but then the
metadata from disk is simply re-written to disk without the
intended updates.


mwrock:
:+1:

PR #5022, opened 2016-06-10
Convert the 3694 warning to a deprecation so it will be subject to the usual deprecation formatting (collected at the bottom, can be made an error, etc).

coderanger:
Ping @chef/client-core  and @lamont-granquist in particular.

### DIFF
--- a/lib/chef/resource_builder.rb
+++ b/lib/chef/resource_builder.rb
@@ -124,9 +124,10 @@ class Chef
     end
 
     def emit_cloned_resource_warning
-      Chef::Log.warn("Cloning resource attributes for #{resource} from prior resource (CHEF-3694)")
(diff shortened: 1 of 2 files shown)


coderanger:
It should be noted I haven't actually run this with a big, complex converge, we might want to tweak the message formatting to work better with how deprecation errors get shown. Thoughts?

lamont-granquist:
i'm OOTO until mid-next-week...

lamont-granquist:
( 👍 to the general idea, dunno about the formatting, i can check on some of my runs that still have a few 3694 errors and see how they look)

lamont-granquist:
BTW, i'm generally not attached at all to the three line "banner" messages and certainly don't like what they do when they're transported to the bottom of the run, seems to create a mess...

lamont-granquist:
LGTM :+1:

PR #5029, opened 2016-06-17
Attributes v1.1

### LINKED ISSUE #5005: Use of node.default_unless mutates other attributes
## Description

In specific situations, calling `node.default_unless` may change the value returned by `node[...]` of a _different_ attribute.

```ruby
raise "Before _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)

node.default_unless['a']['r1']['g']['r'] = 'r'

raise "After _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)
```

This code will raise the following exception in the situation described below:

```
       RuntimeError
       ------------
       After _unless: a/r1/g/u is {}!
```

Before the call to `default_unless` the node access returned a string, after the call it becomes an empty immutable mash.

## Chef Version

Chef Client, version 12.10.24

## Platform Version

CentOS 7.2

## Replication Case

Create an empty cookbook using `chef generate cookbook unless`. Change the default recipe to:

```ruby
raise "Before _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)

node.default_unless['a']['r1']['g']['r'] = 'r'

raise "After _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)
```

Create a JSON environment `env/kitchen.json` for use with kitchen

```json
{
    "name": "kitchen",
    "description": "kitchen example",
    "cookbook_versions": {
    },
    "chef_type": "environment",
    "default_attributes": {
        "a" : {
            "r1" : {
                "g" : {
                    "u" : "u1"
                }
            }
        }
    }
}
```

Adapt the `.kitchen.yml` file to use the environment:

```yaml
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  environments_path: env
  client_rb:
    environment: kitchen

platforms:
  - name: centos-7.2

suites:
  - name: default
    run_list:
      - recipe[unless::default]
    attributes:
```

Run `kitchen converge`

## Client Output

```
...       
       RuntimeError
       ------------
       After _unless: a/r1/g/u is {}!
       
       Cookbook Trace:
       ---------------
         /tmp/kitchen/cache/cookbooks/unless/recipes/default.rb:10:in `from_file'
       
       Relevant File Content:
       ----------------------
       /tmp/kitchen/cache/cookbooks/unless/recipes/default.rb:
       
         3:  # Recipe:: default
         4:  #
         5:  # Copyright (c) 2016 The Authors, All Rights Reserved.
         6:  raise "Before _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)
         7:  
         8:  node.default_unless['a']['r1']['g']['r'] = 'r'
         9:  
        10>> raise "After _unless: a/r1/g/u is #{node['a']['r1']['g']['u']}!" unless node['a']['r1']['g']['u'].is_a?(String)
        11:  
       
       Platform:
       ---------
       x86_64-linux
...
```

lamont-granquist:
### code cleanup

- removes the awful node.set_unless_value_present hack that I wrote ages ago
- removes the awful "MultiMash" I wrote for Chef 12.
- introduces Chef::Decorator::Unchain to convert chained method calls into functional method calls

### introduces functional read/write/unlink/exist? APIs

these API additions were required in order to (cleanly) implement the code cleanup.

on the node object:

- `node.read("foo", "bar", "baz")` equals `node["foo"]["bar"]["baz"] `but with safety (nil instead of exception)
- `node.read!("foo", "bar", "baz")` equals `node["foo"]["bar"]["baz"]` and does trainwreck with NoMethodError

- `node.write(:default, "foo", "bar", "baz")` equals `node.default["foo"]["bar"] = "baz"` and autovivifies and replaces intermediate non-hash objects (very safe) 
- `node.write!(:default, "foo", "bar", "baz")` equals `node.default["foo"]["bar"] = "baz"` and while it autovivifies it can trainwreck if you hit a non-hash on an intermediate key (NoMethodError)
- there is still no non-autovivifying writer, and i don't think anyone really wants one.
- `node.exist?("foo", "bar")` can be used to see if `node["foo"]["bar"]` exists

on node levels:

- `node.default.read/read!("foo")` operates similarly to `node.read("foo")` but only on default level
- `node.default.write/write!("foo", "bar")` is `node.write/write!(:default, "foo", "bar")`
- `node.default.unlink/unlink!("foo")` is `node.unlink/unlink!(:default, "foo")`
- `node.default.exist?("foo", "bar")` can be used to see if `node.default["foo"]["bar"]` exists

on asymmetry:

`node.read(`) does not take a level and only returns the deep merged result.  there is no `node.read(:default, ...)` function and users must use `node.default.read(...)` instead.  this is because we have to support the merged attributes, and i want to do so with simple syntax, rather than forcing something like `node.read(:merged, ...)` onto users in order to support levels.

on the other hand for write/unlink those are mutators and do not support access to the deep merged result.  there cannot be any `node.write(:merged, ...)` or `node.unlink(:merged, ...)` so the logical way to handle thing is that `node.write(:default, ...)` is just sugar for `node.default.write(...)`.

naming concerns:

- we cannot use 'node#delete' because it collides with 'Hash#delete' in very bad ways
- i don't like 'node#remove' for the new API because it is similar but confusingly different to the 'node#rm' API family we introduced in 12.0.0
- we cannot use 'node#set' because we already have that API (and i want to deprecate it), so i went with 'node#write'

### Deprecations

- node.set is deprecated
- node.set_unless is deprecated

both of these are deprecated due to confusion.  people go looking for "how to set a node attribute?" and wind up using "node.set" because that seems super logical, and now they're setting "normal" attributes with persistence which is 99% of the time not what anyone wants to begin with.  the deprecation message pushes people towards node.default/node.override instead of node.normal.

- node.foo.bar.baz

the method_missing code is terrible for many reasons:  node.class vs node["class"] among, the way that IO.puts calls `#to_ary` and expects to rescue NoMethoError, and the inability to change the node API without stomping people.

closes #5005

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -43,6 +43,8 @@ class Chef
     def_delegators :attributes, :keys, :each_key, :each_value, :key?, :has_key?
     def_delegators :attributes, :rm, :rm_default, :rm_normal, :rm_override
     def_delegators :attributes, :default!, :normal!, :override!, :force_default!, :force_override!
+    def_delegators :attributes, :default_unless, :normal_unless, :override_unless, :set_unless
+    def_delegators :attributes, :read, :read!, :write, :write!, :unlink, :unlink!
 
     attr_accessor :recipe_list, :run_state, :override_runlist
 
@@ -196,35 +198,18 @@ class Chef
     # might be missing
     def normal
       attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = false
       attributes.normal
     end
 
-    alias_method :set, :normal
-
-    # Set a normal attribute of this node, auto-vivifying any mashes that are
-    # missing, but if the final value already exists, don't set it
-    def normal_unless
-      attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = true
-      attributes.normal
+    def set
+      Chef.log_deprecation("node.set is deprecated and will be removed in Chef 14, please use node.default/node.override (or node.normal only if you really need persistence)")
+      normal
     end
-
-    alias_method :set_unless, :normal_unless
 
     # Set a default of this node, but auto-vivify any Mashes that might
     # be missing
     def default
       attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = false
-      attributes.default
-    end
-
-    # Set a default attribute of this node, auto-vivifying any mashes that are
-    # missing, but if the final value already exists, don't set it
-    def default_unless
-      attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = true
       attributes.default
     end
 
@@ -232,15 +217,6 @@ class Chef
     # might be missing
     def override
       attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = false
-      attributes.override
-    end
-
-    # Set an override attribute of this node, auto-vivifying any mashes that
-    # are missing, but if the final value already exists, don't set it
-    def override_unless
-      attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = true
       attributes.override
     end
 
@@ -262,7 +238,6 @@ class Chef
 
     def automatic_attrs
       attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = false
       attributes.automatic
     end
 
@@ -290,8 +265,14 @@ class Chef
     end
 
     # Only works for attribute fetches, setting is no longer supported
-    def method_missing(symbol, *args)
-      attributes.send(symbol, *args)
+    # XXX: this should be deprecated
+    def method_missing(method, *args, &block)
+      attributes.public_send(method, *args, &block)
+    end
+
+    # Fix respond_to + method so that it works with method_missing delegation
+    def respond_to_missing?(method, include_private = false)
+      attributes.respond_to?(method, false)
     end
 
     # Returns true if this Node expects a given recipe, false if not.
--- a/lib/chef/decorator/unchain.rb
+++ b/lib/chef/decorator/unchain.rb
@@ -1,0 +1,43 @@
+class Chef
+  class Decorator < SimpleDelegator
+    #
+    # This decorator unchains method call chains and turns them into method calls
+    # with variable args.  So this:
+    #
+    #   node.set_unless["foo"]["bar"] = "baz"
+    #
+    # Can become:
+    #
+    #   node.set_unless("foo", "bar", "baz")
+    #
+    # While this is a decorator it is not a Decorator and does not inherit because
+    # it deliberately does not need or want the method_missing magic.  It is not legal
+    # to call anything on the intermediate values and only supports method chaining with
+    # #[] until the chain comes to an end with #[]=, so does not behave like a hash or
+    # array...  e.g.
+    #
+    #   node.default['foo'].keys is legal
+    #   node.set_unless['foo'].keys is not legal now or ever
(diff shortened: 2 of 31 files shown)


lamont-granquist:
@chef/client-core if anyone has any comments on the work so far it'd be useful.

jaymzh:
@lamont-granquist last I checked this still won't work on our codebase... or did I miss an update somewhere?

lamont-granquist:
This is just an iterative bite towards implementing that.  Its not the whole enchalada of the rewrite.  It should work fine for you.

lamont-granquist:
@chef/client-core okay this is about where i think it needs to be and should be ready for review.

the only thing I'm likely to touch here is that the #write/#write!/exist? API seems like there might be a helper to extract there, but after a solid week of looking at it I'm going to have to leave that one till monday.  i don't see changing the algorithm or behavior, though, i'll just see if i can do a pure refactoring to eliminate some duplication and make it cleaner.

i updated the deprecation notices to be explicit about timelines.   i put in that node.set/node.set_unless is allowed until Chef 14, because i that one is more pushing towards cultural change and i really don't need everyone to drop what they're doing and fix everything before Chef 13.  it can wait.  getting the deprecation notice in there is useful, however, to steer newbies away from the mistake of reaching for node.set without understanding what else it does.   the method_missing one I slated for removal in Chef 13.  that one really is horribly bad and I think its legit to light a fire under people's butts.   its also a case of where we can either suffer through fixing the deprecation warnings, suffer through breaking people when we change the node API or suffer with a node API that sucks and we can't change.  and the longer we leave the method_missing problem lying around the worse the problem gets through accretion.

lamont-granquist:
squashed and rebased.  @coderanger your `write(*keys) { value }` api is added.  @chef/client-core speak now or forever hold your peace.

thommay:
The appveyor failures *look* unrelated, 👍

lib/chef/node.rb
```diff
@@ -196,51 +198,25 @@ def [](attrib)
...
-    alias_method :set, :normal
-
-    # Set a normal attribute of this node, auto-vivifying any mashes that are
-    # missing, but if the final value already exists, don't set it
-    def normal_unless
-      attributes.top_level_breadcrumb = nil
-      attributes.set_unless_value_present = true
-      attributes.normal
+    def set
+      Chef.log_deprecation("node.set is deprecated and will be removed in Chef 14, please use node.default/node.override (or node.normal only if you really need persistence)")
```
tyler-ball: Perhaps a link here to talk about how `node.normal` is different in case people don't understand what is meant by `persistence`?

tyler-ball:
One minor comment but otherwise 👍

lib/chef/node.rb
lamont-granquist: So, I really like how rust does that, but I don't want to block this PR on starting to do that better.

lamont-granquist: (actually rust does it one better with the short error message with the code in it, and then the ability to expand the code into a wall of text, and then sometimes with a shout out to a URL -- i'm insanely jealous of them, but googling 'chef attributes normal persistence' in this case looks entirely reasonable)

smurawski:
👍

PR #5033, opened 2016-06-18
Do not use bundle install to update Gemfile.lock after version bump

jkeiser:
Using bundle install to update the Gemfile.lock to update the Chef version will cause bundle to "unfreeze" chef and all its dependent gems, making them fair gem to update. We do not want that. Therefore, we modify the Gemfile.lock ourselves because we just roll that way.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,12 +29,13 @@ require_relative "tasks/cbgb"
 require_relative "tasks/dependencies"
 require_relative "tasks/changelog"
 
-ChefConfig::PackageTask.new(File.expand_path("..", __FILE__), "Chef") do |package|
+ChefConfig::PackageTask.new(File.expand_path("..", __FILE__), "Chef", "chef") do |package|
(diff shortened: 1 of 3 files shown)


lamont-granquist:
👍

smurawski:
:+1:

coderanger:
I am getting deeply concerned about how complex our bundler "integration" is getting. I understand each step along the way made sense but I think we may need to reconsider this direction very soon given how ungainly things are getting.

lamont-granquist:
I'd agree.  It does not remove any work and was never designed to remove any work.  If the bumpbot finds gems that are broken there's still the necessarily human job of figuring out why and what to do about it.  All it does is attempt to mitigate interrupt-driven work, but in the process it causes substantially more work to maintain it all.  As long as the team who was getting hammered by the interrupt-driven work feels that this is still better than the old alternative, that is fine and its their choice to maintain it.

PR #5034, opened 2016-06-18
Revert deprication commit to fix travis

mwrock:
Reverts chef/chef#5022 which is breaking the poise travis tests.


coderanger:
:-1: This should be fixed on the Poise side.

coderanger:
(and I'm working on that :)

mwrock:
sweet @coderanger if you are on that and can get it green fairly quickly then I agree its the way to go.

coderanger:
Poise master is green again.

Issue #5027, opened 2016-06-16
Undefined local variable lookup in multiplexed_dir.rb


tomykaira:
## Description

https://github.com/chef/chef/blob/master/lib/chef/chef_fs/file_system/multiplexed_dir.rb#L44 accesses `seen` variable, but it is not defined there.

## Chef Version

Chef Client, version 12.11.18

## Platform Version

amazon linux

## Stacktrace

```
Exception raised! #<NameError: undefined local variable or method `seen' for #<Chef::ChefFS::FileSystem::MultiplexedDir:0x00000003c77060>>
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system/multiplexed_dir.rb:44:in `block in make_child_entry'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system/multiplexed_dir.rb:40:in `each'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system/multiplexed_dir.rb:40:in `make_child_entry'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system/base_fs_object.rb:103:in `child'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system.rb:103:in `block in resolve_path'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system.rb:102:in `each'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/file_system.rb:102:in `resolve_path'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/chef_fs_data_store.rb:828:in `get_single_cookbook_version'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.11.18/lib/chef/chef_fs/chef_fs_data_store.rb:549:in `list'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-zero-4.6.2/lib/chef_zero/data_store/v1_to_v2_adapter.rb:83:in `block in list'
[...]

PR #5037, opened 2016-06-20
Fix for #5027 - Remove bad variable reference

robdimarco:
A previous change introduced what appears to be a copy/paste bug. A guard condition from the children
method was used in the make_child_entry incorrectly.

This removes the reference to the non-existent variable `seen`

### DIFF
--- a/lib/chef/chef_fs/file_system/multiplexed_dir.rb
+++ b/lib/chef/chef_fs/file_system/multiplexed_dir.rb
@@ -41,7 +41,7 @@ class Chef
             child_entry = dir.child(name)
             if child_entry.exists?
               if result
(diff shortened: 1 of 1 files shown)


thommay:
Hi,
thanks for the submission. I think the right fix would be to use `seen[child_entry.name]` rather than deleting the condition - which could give a lot of debug spam!

robdimarco:
Well, since there is only one child in this method (as opposed to in the `children` method)`, I would expect that the `child_entry.name

thommay:
ok, yeah, agreed. 👍

smurawski:
👍

mwrock:
fixes #5027

PR #5038, opened 2016-06-20
adding directory caching

youngjl1:
I recently started using chef version 12.11.18 and noticed that running chef-client on a windows vagrant instance went from 30 seconds to over 5 minutes. My company has a lot of cookbooks (~200) and the children method below was taking 10 seconds to run through them all. It was being called 23 times (I guess because the runlist included a lot of cookbooks/dependencies). Anyway, that's what accounted for the increased time. When I added the caching the run time went back down to 30 seconds.

### DIFF
--- a/lib/chef/chef_fs/file_system/repository/directory.rb
+++ b/lib/chef/chef_fs/file_system/repository/directory.rb
@@ -31,6 +31,8 @@ class Chef
           alias_method :display_path, :path
           alias_method :display_name, :name
           alias_method :bare_name, :name
+          
+          @@dir_cache = Hash.new
 
           def initialize(name, parent, file_path = nil)
             @parent = parent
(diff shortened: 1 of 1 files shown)


lamont-granquist:
This introduces global state, and as a result it broke several tests -- which is really not surprising.

Either some better engineering needs to go into this -- like creating a real singleton class and having a reset method on it, which is called from the spec tests to reset the state between tests -- or else the problem should be solved further up the stack by not calling this method so much.

youngjl1:
Thanks. It would be difficult to reduce the number of calls further up the stack because they are coming from a different project (chef-zero) which is verifying each and every cookbook dependency exists. Some sort of cache makes the most sense to me. Weirdly you see logic in chef_fs to check a "memory_store" object before it goes and manually iterates through the FS

```ruby
      def exists_dir?(path)
        if use_memory_store?(path)
          @memory_store.exists_dir?(path)
```
source: https://github.com/chef/chef/blob/master/lib/chef/chef_fs/chef_fs_data_store.rb
For searching cookbooks though it always bypasses the memory store and I'm not even sure what the purpose is.

I would interested in implementing a singleton class as you mention but I'm not very experienced with ruby. Is there an example in the chef code I could use as a template?

For now I am able to get around the slowness by running "chef-solo --legacy-mode"

lamont-granquist:
@thommay maybe you could look at this since it does seem to be a particularly bad perf bug in zero

chef-supermarket:

PR #5039, opened 2016-06-20
Bump deps to the latest

tas50:
This is to bring in the latest ohai, but we also get net-ssh 3.2 and various things for acceptance


lamont-granquist:
i thought we had a bumpbot?  @jkeiser ?

PR #5041, opened 2016-06-21
FLOW-335: Add a counterpart to "skipped due to only_if/not_if" so users know that a  resource had a guard that allowed resource execution.

randomcamel:
This issue came from [user feedback](https://feedback.chef.io/forums/301644-chef-product-feedback/suggestions/10728330-logging-for-guard-resource-should-show-whether-t).  Here's part of the blurb from the internal ticket on it:

> As a person who writes Chef Cookbooks containing guard resources, I should be able to tell from reading the log whether an evaluated guard resource was inclusive (only_if) or exclusive (not_if), so that I can better understand the actions taken by my code.

### DIFF
--- a/lib/chef/formatters/minimal.rb
+++ b/lib/chef/formatters/minimal.rb
@@ -182,6 +182,9 @@ class Chef
         print "S"
       end
 
+      def resource_not_skipped(resource, action, conditional)
+      end
+
       # Called after #load_current_resource has run.
       def resource_current_state_loaded(resource, action, current_resource)
       end
--- a/lib/chef/formatters/minimal.rb
+++ b/lib/chef/formatters/minimal.rb
@@ -182,9 +182,6 @@ class Chef
         print "S"
       end
 
-      def resource_not_skipped(resource, action, conditional)
-      end
-
       # Called after #load_current_resource has run.
       def resource_current_state_loaded(resource, action, current_resource)
       end
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1493,6 +1493,8 @@ class Chef
       conditionals = [ conditional_action ] + only_if + not_if
       conditionals.find do |conditional|
         if conditional.continue?
+          events.resource_not_skipped(self, action, conditional) if events
+          Chef::Log.debug("Not skipping #{self}, passed guard #{conditional.description}")
           false
         else
           events.resource_skipped(self, action, conditional)
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1494,7 +1494,6 @@ class Chef
       conditionals.find do |conditional|
         if conditional.continue?
           events.resource_not_skipped(self, action, conditional) if events
-          Chef::Log.debug("Not skipping #{self}, passed guard #{conditional.description}")
           false
         else
           events.resource_skipped(self, action, conditional)
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -181,6 +181,9 @@ class Chef
       @pending_update = nil unless nested_resource?(resource)
     end
 
+    def resource_not_skipped(resource, action, conditional)
+    end
+
     def resource_updated(new_resource, action)
       @total_res_count += 1
     end
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -181,9 +181,6 @@ class Chef
       @pending_update = nil unless nested_resource?(resource)
     end
 
-    def resource_not_skipped(resource, action, conditional)
-    end
-
     def resource_updated(new_resource, action)
       @total_res_count += 1
     end
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -147,6 +147,9 @@ class Chef
         update_current_resource_report(resource_report)
       end
 
+      def resource_not_skipped(resource, action, conditional)
+      end
+
       # see EventDispatch::Base#resource_updated
       # Flag the current ResourceReport instance as updated (as long as it's
       # a top-level resource).
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -325,6 +325,10 @@ class Chef
       def resource_skipped(resource, action, conditional)
       end
 
+      # Called when a conditional has passed and the resource action is not skipped.
+      def resource_not_skipped(resource, action, conditional)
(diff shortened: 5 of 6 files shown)


lib/chef/formatters/minimal.rb
```diff
@@ -182,6 +182,9 @@ def resource_skipped(resource, action, conditional)
         print "S"
       end
 
+      def resource_not_skipped(resource, action, conditional)
```
coderanger: This should be inherited from base already.

lib/chef/resource.rb
```diff
@@ -1493,6 +1493,8 @@ def should_skip?(action)
       conditionals = [ conditional_action ] + only_if + not_if
       conditionals.find do |conditional|
         if conditional.continue?
+          events.resource_not_skipped(self, action, conditional) if events
+          Chef::Log.debug("Not skipping #{self}, passed guard #{conditional.description}")
```
coderanger: This is excessive debug spam.

lib/chef/resource_reporter.rb
```diff
@@ -181,6 +181,9 @@ def resource_skipped(resource, action, conditional)
       @pending_update = nil unless nested_resource?(resource)
     end
 
+    def resource_not_skipped(resource, action, conditional)
```
coderanger: This will be inherited from the base class.

coderanger:
Overall I'm not sure we should add this to the default `doc` formatter. It seems a little excessive on the output.

coderanger:
👍 on adding the callback, just needs to be cleaned up a bunch.

PrajaktaPurohit:
Recipe: Create file /var/www/html/baz.html only if foo exists

in -lwarn mode:
```
foo does not exist -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create (skipped due to only_if)
foo exists -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create (passed guard only_if)
  - create new file /var/www/html/baz.html
  - update content in file /var/www/html/baz.html from none to e9700a
  - suppressed sensitive resource
  - change mode from '' to '0755'
foo exists -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create (passed guard only_if)
   (up to date)
foo does not exist -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create (passed guard only_if)
   (up to date)
```

in -ldebug mode:
```
foo does not exist -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create[2016-07-11T13:19:40-07:00] INFO: Processing file[/var/www/html/baz.html] action create (remote_directory::default line 24)
 (skipped due to only_if)
foo exists -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create[2016-07-11T13:20:32-07:00] INFO: Processing file[/var/www/html/baz.html] action create (remote_directory::default line 24)
 (passed guard only_if)
foo exists -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create[2016-07-11T13:21:13-07:00] INFO: Processing file[/var/www/html/baz.html] action create (remote_directory::default line 24)
 (passed guard only_if)
foo does not exist -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create[2016-07-11T13:22:07-07:00] INFO: Processing file[/var/www/html/baz.html] action create (remote_directory::default line 24)
 (skipped due to only_if)
``` 

/cc @randomcamel @coderanger @loganmeetsworld

coderanger:
Very :-1: on that output. It adds nothing and clutters the screen. If you want to add the event callback for use with Insights or whatever that's cool but this doesn't belong in any human output.

adamedx:
Ok, I'm 👍  😄 . @coderanger, I've updated the PR description with the blurb from the internal ticket that triggered this as well as the user feedback that generated it. The ticket is asserting that multiple users are not clear on the type of guard being executed when the resource is not skipped. Users cared enough to use their votes for this feature. Assuming that need is true, do you have a suggestion on how they can figure this out without finding the particular cookbook to validate?

coderanger:
I guess I'm not sure what you think this will clarify. All guards are always executed, so this isn't indicating anything more than the code.

coderanger:
Upon further reading I don't think this addresses the user comment at all. What they are looking for is more detail in the `skipped due to ...` output to maybe show what actual condition caused it to be skipped.

thommay:
I'm 👎 on this as it stands; I don't think this implementation matches the user request, and I agree with @coderanger that this doesn't provide any additional information to the user.

coderanger:
I would agree that adding some output at `debug` that has specifics of the mode (not vs only), type (command string vs block), and actual output of the guard clause would be a help for debugging, but that should go in log data instead of the formatter probably :)

coderanger:
Or maybe put a very abbreviated version of that in the formatter data.

mwrock:
So I'm a bit torn on this. I agree with @thommay and @coderanger that the added output has no added value and basically just spits back the guard type specified in the recipie. On the other hand, reading the comment this seems to be exactly what the user is asking for (I am NOT saying that means we should do it). So I'm a bit confused on what @coderanger and @thom think might be better output or how they interpret the comment. In my experience I have often wished I could find out which specific clause of the guard triggered the converge especially in the case of complex guards. Obviously that would be much more involved work.

I think we need to be hyper aware of log real estate and even though this adds only a few words to the output:
```
foo exists -> Recipe: remote_directory::default
  * file[/var/www/html/baz.html] action create (passed guard only_if)
```
maybe its best if we ONLY add it for debug output.

thoughts?

coderanger:
The problem is that formatter output doesn't have levels, that's only for logging.

thommay:
I guess my problem is that I can't see what the use case is here, and I don't think we've described it adequately. I've certainly thought "why has this resource run", but the information presented here doesn't tell me that - it just tells me that the guard happens to have passed (which I know already). I still have to go back to the cookbook to find out what the guard was.

charlesjohnson:
My understanding of the user case is:
"Skipped due to guard resource" is not descriptive enough, they're looking for either, "Skipped due to not_if" or "skipped due to only_if." I agree that this is a useful change, people want to know what caused the behavior.

However, if a resource proceeds because a guard didn't fire, then it's implicit in the log already that the guard didn't fire. I don't think we need to be explicit in standard output, although it'd be useful in debug.

coderanger:
@charlesjohnson :+1: on that change, this patch is unrelated though.

mwrock:
Yeeah I think we all agree with :-1:  on the change in this PR

PR #5042, opened 2016-06-21
Update Nokogiri

mwrock:
needs to merge together with https://github.com/chef/omnibus-software/pull/682


omnibus/.kitchen.yml
```diff
@@ -93,7 +94,9 @@ platforms:
       # gets cranky if the mounted Chef source folder is a symlink. This
       # mounts at `C:\vagrant\code` and the Chef source folder is available
       # at `C:\vagrant\code\chef`
-      - ['../..', '/vagrant/code']
+      - ['..', '/vagrant/code/chef']
```
schisamo: This change causes the following error:

```
                          I | 2016-06-21T07:54:07-07:00 | fatal: Invalid symlink
 'C:/vagrant/code/chef': Function not implemented
The following shell command exited with status 128:

    $ git ls-remote "http://git.savannah.gnu.org/r/config.git" "master*"

Output:

    (nothing)

Error:

    fatal: Invalid symlink 'C:/vagrant/code/chef': Function not implemented

C:/languages/ruby/2.1.8/lib/ruby/gems/2.1.0/bundler/gems/omnibus-56b8e507df41/li
b/omnibus/util.rb:101:in `rescue in shellout!'
```

Please change it back to `- ['../..', '/vagrant/code']`. I'll let @ksubrama and @jkeiser give more of the why.

mwrock:
that filled up my vm with my entire dev directory but I'll just regigger manually when using this

tyler-ball:
@mwrock Can you also remove [this](https://github.com/chef/chef/blob/31a8c4ddfef5e8ec5c42967a97196e81bb855347/version_policy.rb#L76) comment?

Otherwise, 👍

mwrock:
changed per feedback @schisamo and I'll squash when I adjust the omnibus Gemfile.lock

omnibus/.kitchen.yml
btm: @mwrock is this feedback resolved? I saw https://github.com/chef/chef/pull/5042#issuecomment-227488991, but this note didn't go away, so I'm feeling like this line wasn't changed after this feedback.

mwrock:
changed it back for real this time.

btm:
:+1:

omnibus/.kitchen.yml
```diff
@@ -11,7 +11,7 @@ driver:
     cpus: 4
     memory: 4096
   synced_folders:
-    - ['..', '/home/vagrant/chef']
+    - ['../..', '/vagrant/code']
```
lamont-granquist: is this stuff intended to be included here?

omnibus/config/software/chef.rb
```diff
@@ -69,13 +69,6 @@
   block { log.info(log_key) { "" } }
   bundle "install --verbose", env: project_env
 
-  # For whatever reason, nokogiri software def deletes this (rather small) directory
```
lamont-granquist: pretty certain this was done to avoid files-with-spaces which i think still breaks either aix or solaris 10 builds.

mwrock: after upgrading nokogiri this started breaking with complaints in the `bundle show` stating it could not find mini_portile. Jenkins has built this full green so we should be ok.

lamont-granquist: we may have stopped shipping dev gems in the build the way we used to?  i dunno..  as long as you test through manhattan.ci on solaris + aix its fine...

PR #5045, opened 2016-06-21
First pass on --config-option handling.

coderanger:
This adds a command line argument to override any arbitrary config key/value pair from the command line a la `ssh -o`. This is intended for later use in the worker processes for multiprocessing but it seems like it might have value beyond that.

Depends on https://github.com/chef/mixlib-cli/pull/28. Ping @chef/client-core for review.


coderanger:
An example of usage would be something like `chef-client --config-option no_lazy_load=false`

lib/chef/application.rb
```diff
@@ -108,7 +108,32 @@ def load_config_file
...
+        extra_parsed_options = extra_config_options.inject({}) do |memo, option|
+          # Sanity check value.
+          Chef::Application.fatal!("Unparsable config option #{option.inspect}") if option.empty? || !option.include?("=")
+          # Split including whitespace if someone does truly odd like
+          # --config-option "foo = bar"
+          key, value = option.split(/\s*=\s*/, 2)
+          # Call to_sym because Chef::Config expects only symbol keys. Also
+          # runs a simple parse on the string for some common types. Not using
+          # a YAML parser or similar because security. Could maybe use JSON?
+          memo[key.to_sym] = case value
```
lamont-granquist: nil?

lamont-granquist:
:+1:

lib/chef/application.rb
coderanger: Good call, do you think there is a way to use a yaml parse for this that isn't going to be worse than what I have already?

coderanger: `YAML.parse(option).to_ruby` does work but I'm concerned this could have unexpected consequences, but then again if you can inject a value into this config option it's probably game over?

coderanger: Looks like `YAML.safe_load(option)` might be good enough :)

lamont-granquist:
We already use json-formatted options in at least two other places on the command line.  We should probably stick with that instead of YAML (honestly i think they're both horrible for orthogonal reasons).

I'd also suggest that there be a `--config-options` that took text and massaged it and tried to produce the easiest to use command line for setting simple things.  Then have `--config-options-json` that just parsed it from JSON.  Could also have `--config-options-yaml` if your heart is set on YAML, but I think it should be either both or it should be JSON...

lamont-granquist:
Here's some prior art for the `--bootstrap-vault-*` options:

https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L234-L251

lamont-granquist:
being able to pass a json file and/or yaml file in instead of config.rb might be amusing, too...

coderanger:
I'm only using `YAML.safe_load` to short-circuit the little bit of parser logic, not for actually deep nested values. I don't think JSON is the right way to go since most config options are simple key/value pairs.

lamont-granquist:
👍

thommay:
👍 once the tests pass :)

coderanger:
Tests will pass once we merge the mixlib-cli change :)

thommay:
@coderanger I just released mixlib-cli 1.7.0 if you'd like to do the Gemfile dance and update this?

lamont-granquist:
pingles.

thommay:
(mixlib-cli 1.7.0 fails some chef tests, not sure if that ever got resolved?)

coderanger:
We've currently got mixlib-cli locked to exclude 1.7 (`s.add_dependency "mixlib-cli", "~> 1.4", "< 1.7"`). I could remove that, but I don't know the history of it.

coderanger:
Did some research, the pin was due to a bug that was already fixed in master. See https://github.com/chef/chef/commit/6fd98696863b83449354a93a79918f5c399c7cce https://github.com/chef/mixlib-cli/issues/29 and https://github.com/chef/chef/commit/b50d7dd216c51f7b3aa09b4c3f2bc4031a60344e for details. Going to update the dependency in the branch and let tests run again.

PR #5046, opened 2016-06-22
correctly select current gem version

### LINKED ISSUE #4954: gem_package upgrade is always upgrading if there is any old version installed.
## Description

If you have:

```
gem_package "p" do
  action :upgrade
end
```

and if you have an old version of the package installed from previously, chef will upgrade the package every run.

The root cause of this is in the rubygems provider, the call stack looks something like this:

1. there is a line looking for the installed versions of the gem by [calling essentially `Gem::Specifications.find_all_by_name`](https://github.com/chef/chef/blob/v12.10.24/lib/chef/provider/package/rubygems.rb#L90)
2. which in turn [calls `Gem::Dependency#matching_specs`](https://github.com/rubygems/rubygems/blob/v2.6.3/lib/rubygems/specification.rb#L1016)
3. which [calls `Gem::Specification.to_stubs`](https://github.com/rubygems/rubygems/blob/v2.6.3/lib/rubygems/dependency.rb#L279)
4. This method sorts the gem matching the specification that you specified in the gem_package resource. On the version of Rubygem that is included with Chef 12.10.24 (which as far as I can tell is 2.6.3), gives the gems in a list where the newest gem version comes first. This maybe bugged, but I'm not sure. See: [line that calls `_resort!`, the sorting function for the matched gem specifications](https://github.com/rubygems/rubygems/blob/v2.6.3/lib/rubygems/specification.rb#L852) and the [definition for `_resort!`](https://github.com/rubygems/rubygems/blob/v2.6.3/lib/rubygems/specification.rb#L859-L865).

This means that when chef gets the [current version of the gem and takes the last element](https://github.com/chef/chef/blob/v12.10.24/lib/chef/provider/package/rubygems.rb#L437-L440) of the list returned from the stack that I outlined above, it takes the oldest version, which results in gem_package always upgrading every time chef runs.

While most of the time people might not notice as everything works fine, any subscription or notifications will be sent every time as chef sees that the resource has been changed.

Not sure if this is a rubygem bug or a chef bug. I also see: https://github.com/rubygems/rubygems/pull/1601

## Chef Version

- **Chef**: v12.10.24, all the links to code above is to that version as well.
- **Rubygems embedded in Chef**: 2.6.3 as far as I can tell (from reading `/opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems.rb` for `Gem::VERSION`).

## Platform Version

Ubuntu trusty

## Replication Case

1. Have the version above installed, have a cookbook with `gem_package 'something' { action :upgrade }`.
2. Install an old version of `something` first before running chef.
3. Run chef to get the latest version.
4. Run chef again to see it reinstalls the latest version

## Client Output

Chef client output (replaced actual gem name as it is internal):

```
 DEBUG: gem_package[something] detected omnibus installation in /opt/chef/embedded/bin
 DEBUG: gem_package[something] searching for 'gem' binary in path: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/chef/embedded/bin:/opt/chef/embedded/bin
 DEBUG: gem_package[something] using gem '/usr/bin/gem'
 DEBUG: gem_package[something] found installed gem something version 0.9.0 matching something (>= 0)
 DEBUG:  found gem something version 0.9.1 for platform ruby from #<Gem::Source:0x00000003ed6f68>
 DEBUG: gem_package[something] something is out of date, will upgrade to 0.9.1
 INFO: gem_package[something] upgraded something to 0.9.1
 DEBUG: gem_package[something] resetting gem environment to default
```

`$ gem list` output:

```
something (0.9.1, 0.9.0)
```

thommay:
Fixes #4954

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -431,17 +431,23 @@ class Chef
         end
 
         def current_version
-          #raise 'todo'
+          # rubygems 2.6.3 ensures that gem lists are sorted newest first
+          pos = if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("2.6.3")
+                  :first
+                else
+                  :last
+                end
+
           # If one or more matching versions are installed, the newest of them
           # is the current version
           if !matching_installed_versions.empty?
-            gemspec = matching_installed_versions.last
+            gemspec = matching_installed_versions.send(pos)
             logger.debug { "#{@new_resource} found installed gem #{gemspec.name} version #{gemspec.version} matching #{gem_dependency}" }
             gemspec
             # If no version matching the requirements exists, the latest installed
             # version is the current version.
           elsif !all_installed_versions.empty?
-            gemspec = all_installed_versions.last
+            gemspec = all_installed_versions.send(pos)
             logger.debug { "#{@new_resource} newest installed version of gem #{gemspec.name} is #{gemspec.version}" }
             gemspec
           else


smurawski:
👍

PR #5048, opened 2016-06-22
Turn off fips with an empty environment var

mwrock:
This PR fixes build breaks in the chef-dk jenkins debian and redhat testers when they run chef unit tests. They currently fail due to a mocking expectationi error on `::File.read` because `ChefConfig` reads `/proc/sys/crypto/fips_enabled` early in the run. and the breaking tests expect a different path.

As an alternative to mocking the call to `read` on the fips path which would be irrelevant to the test and likely raise questions as to why that is being mocked, this PR allows the fips check to have a forced `false` value if the `FIPS_ENABLED` variable is present but empty and if so, the check of the fips file is avoided.

So the following `FIPS_ENABLED` values have the following impact:

value | impact
------- | ----------
`nil` | unknown and the systems settings are checked incurring a call to read `/proc/sys/crypto/fips_enabled`
`empty` | FIPS is NOT enabled
Not empty | Fips is enabled
`

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -519,7 +519,16 @@ module ChefConfig
 
     # Set to true if Chef is to set OpenSSL to run in FIPS mode
     default(:fips) do
-      !ENV["CHEF_FIPS"].nil? || ChefConfig.fips?
+      # CHEF_FIPS is used in testing to override checking for system level
+      # enablement. There are 3 possible values that this variable may have:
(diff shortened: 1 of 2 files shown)


btm:
is CHEF_FIPS used for anything other than our tests? Either way, a comment in chef-config that answers that and includes the table of expectations that are in the PR would be good.

mwrock:
good call @btm. comment added.

btm:
:+1:

PR #4831, opened 2016-04-14
Port a change made to chef-cookbooks/windows to fix issue chef-cookbooks/windows#89.

westhusing:
* Fix and simplify the parsing of the uninstall string for removing a Windows package.

@mwrock This is the change I was attempting to submit into chef-cookbooks/windows.

PR #5050, opened 2016-06-23
fixes some uninstall scenarios by calling uninstall string directly

mwrock:
Replaces PR #4831 largely keeping its implementation but quoting the string if it is a existing file. Researching this it looks like NullSoft often (but not always) does not quote the uninstall string. Most other installers do provide quotes unless they truly are not needed.


btm:
:+1:

PR #5053, opened 2016-06-23
typo: winevt debug messages had type INFO

docwhat:
Someone just had a copy-pasta error.

### DIFF
--- a/lib/chef/log/winevt.rb
+++ b/lib/chef/log/winevt.rb
@@ -69,7 +69,7 @@ class Chef
 
       def debug(msg)
         @eventlog.report_event(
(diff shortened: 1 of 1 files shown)


smurawski:
It doesn't appear that Win32::EventLog has a DEBUG_TYPE - https://github.com/djberg96/win32-eventlog/blob/ffi2/lib/win32/eventlog.rb

docwhat:
Oh frell... it doesn't exist in Windows. Really?  https://msdn.microsoft.com/en-us/library/windows/desktop/aa363679(v=vs.85).aspx

*sigh*

PR #5058, opened 2016-06-24
Expand data_collector resource list to include all resources

adamleff:
Historically when a Chef run fails, the event handlers only report
on resources that have been processed during the run. This change
allows the Data Collector to report those resources in the resource
collection that have not yet been processed so users can better
ascertain how much of their run was completed, and specifically
what resources were not processed as a result of the Chef failure.

Instead of building up a list of resources as they are processed,
this change instead creates a resource report for each resource+action
in the resource collection and then modifies each of those reports
once the resource is processed.


thommay:
👍

lamont-granquist:
👍

jkeiser:
:+1:

Issue #5051, opened 2016-06-23
Correctly load default paths in ChefFS


SaltwaterC:
## Description

chef-solo in local mode isn't behaving like chef-solo in legacy mode. Some configuration bits that are taken for granted aren't actually working with chef-solo in local mode.

## Chef Version

12.11.18

## Platform Version

Irrelevant

## Replication Case

Minimal solo.rb config:

```ruby
log_level :info
log_location '/var/log/chef-client.log'
environment 'foo'
```

Environment 'foo' declared in /var/chef/environments/foo.rb
Cookbooks in /var/chef/cookbooks.

chef-solo --json-attributes /etc/chef/node.json # works before 12.11.x
chef-solo --legacy-mode --json-attributes /etc/chef/node.json # works after 12.11.x

## Client Output

```
[2016-06-23T15:22:56+00:00] INFO: Forking chef instance to converge...
[2016-06-23T15:22:56+00:00] INFO: *** Chef 12.11.18 ***
[2016-06-23T15:22:56+00:00] INFO: Platform: x86_64-linux
[2016-06-23T15:22:56+00:00] INFO: Chef-client pid: 19315
[2016-06-23T15:22:57+00:00] INFO: Setting the run_list to ["coookbook::default"] from CLI options
[2016-06-23T15:22:57+00:00] INFO: HTTP Request Returned 404 Not Found: Object not found: chefzero://localhost:8889/environments/foo
```

## Explanation

chef-solo in local mode isn't taking environment_path '/var/chef/environments' as default (contrary to the docs). The same can be said about cookbook_path '/var/chef/cookbooks' (although, to be fair, this value isn't documented).

coderanger:
Can you be more specific about what isn't working, such as filling in some of the template fields you left blank?

SaltwaterC:
@coderanger sorry, I've hit Command + Enter accidentally while I was writing the ticket. Filled in the rest of the blanks.

SaltwaterC:
If I only fill in environment_path, it fails with something like:

```
[2016-06-23T17:20:53+01:00] INFO: #<ChefZero::RestErrorResponse: 412: No such cookbook: cookbook>
```

These two configuration bits are in clear breakage of 100% backwards compatibility to legacy mode.

coderanger:
Ping @thommay, maybe something in the ChefFS path mounting?

PR #5059, opened 2016-06-24
Ensure that chef-solo uses the expected repo dir

thommay:
Solo has always defaulted to `/var/chef` but local-mode uses the user's
home directory, so we need to ensure that we set the correct path before
entering local-mode.
Fixes: #5051

cc @chef/client-maintainers

### DIFF
--- a/lib/chef/application/solo.rb
+++ b/lib/chef/application/solo.rb
@@ -241,6 +241,13 @@ class Chef::Application::Solo < Chef::Application
         ARGV[dash_r] = "--recipe-url"
       end
 
+      # For back compat reasons, we need to ensure that we try and use the cache_path as a repo first
+      Chef::Log.debug "Current chef_repo_path is #{Chef::Config.chef_repo_path}"
+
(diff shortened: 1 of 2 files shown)


smurawski:
👍

lamont-granquist:
👍

lib/chef/application/solo.rb
```diff
@@ -241,6 +241,13 @@ def reconfigure
         ARGV[dash_r] = "--recipe-url"
       end
 
+      # For back compat reasons, we need to ensure that we try and use the cache_path as a repo first
+      Chef::Log.debug "Current chef_repo_path is #{Chef::Config.chef_repo_path}"
```
btm: this seems likely to confuse someone if it says the path is a certain thing and then immediately switches to a different path?

thommay: it should almost never switch - if the user's configured it, no switch; this just ensures the setting is set and not a default.

PR #5063, opened 2016-06-27
service - upstart restart does not uses latest job config, need start/stop

abhishekkr:
latest upstart job config is not used by `/sbin/restart` but retains oiginal configuration from running instance
as noted under
http://upstart.ubuntu.com/cookbook/#restart

thus use-cases like
- chef updating upstart job config
- notifying service restart

wouldn't bring in desired changes, I tried using `initctl reload-configuration` as a pre-req but it didn't had any impact in my trials

`/sbin/restart` is more elegant way of restarting upstart service, but not safer for users who don't know of this behavior


lamont-granquist:
specs are busted, code looks broken and probably needs `new_resource.start_command`/`new_resource.stop_command` and needs a spec around the new behavior.

abhishekkr:
@lamont-granquist will fix

chef-supermarket:

* @abhishekkr

coderanger:
#5066 has a different implementation of the same feature.

chef-supermarket:

abhishekkr:
#5066 is an effective implementation, it will be good if `upstart` resource documentation on Chef can then clearly state the difference on `stop_start` and `restart` action
either would work till users have the info, even without these we have been getting around using a different `restart_command` ...the key here is to highlight this behavioral difference

lamont-granquist:
So, `start_service` and `stop_service` are not updating `current_resource.running` which both use as idempotency checks.

They could be changed to update this state, but they really shouldn't be updating the current_resource.  We use that for reporting to document the "pre" state of the system to do comparisons with and once that is constructed, it should not really be mutated (in an ideal world after load_current_resource should construct a frozen object, but we'd most likely cause chaos by doing that so we just need to be careful and consider it frozen).  What should happen here is that the state should be copied to an accessor on the provider, which the provider can then update however it likes to track its own internal state.

You could try to extract the non-idempotent bits of start_service and stop_service but since they call super i don't think that approach will work well and turns into spaghetti in my head when I think about it...

abhishekkr:
@lamont-granquist thanks for pointing out this flow, I think approach to have an accessor on provider shall work fine and stay clean... will soon apply, check and commit

lamont-granquist:
any update here?

abhishekkr:
@lamont-granquist got caught up in parallel track of works :(
shall push in required change within few days

abhishekkr:
most failed test were due to `git clone` failure, comment to re-trigger

abhishekkr:
trying to re-trigger tests

thiagomael:
+1

abhishekkr:
can some repo owner please re-trigger tests for this PR, thanks

ranjib:
@abhishekkr done

thommay:
hi @abhishekkr; it looks like you need to rebase on to current master to get travis passing.

abhishekkr:
@thommay thanks the earlier issue is gone now
for DCO, I did pushed in a gpg signed commit as last... do I need to follow anything more

thommay:
Existing PRs are fine with the CLA, but for the future DCO requires a "Signed-off-by" line (`git commit -s` ) rather than a GPG signature (although those are good too!).

abhishekkr:
@thommay I did followed this https://help.github.com/articles/signing-commits-using-gpg/
and did last commit using `git commit -S`

lamont-granquist:
last commit isn't sufficient -- all commits need DCO sign off (although we've got a grace period now for commits that still were covered under the old CLA, so this PR should be fine)

abhishekkr:
@lamont-granquist ok, thanks

abhishekkr:
let me know if there is anything I need to do here

tas50:
@lamont-granquist Can we merge this one since it was under the CLA at the time of completion?

abhishekkr:
I'm confused, this new `Bug` label is because it fixes a bug or due to my mistake it created a new bug.

thommay:
@abhishekkr because it fixes a bug. It's to help automatically generate the changelog.

abhishekkr:
@thommay thanks for clarification

PR #5065, opened 2016-06-28
Adding node object to Data collector run_converge message

adamleff:
Many use cases that involve consuming the run_converge messages and
displaying them to end users include needing additional data about
the node that generated the run_converge message. This change
consolidates the run_converge and node_update message into a single
message making it easier for users to filter run_converge messages
based on attributes and criteria of the node itself.

### DIFF
--- a/lib/chef/data_collector/messages.rb
+++ b/lib/chef/data_collector/messages.rb
@@ -68,6 +68,7 @@ class Chef
           "id"                     => run_status.run_id,
           "message_version"        => "1.0.0",
           "message_type"           => "run_converge",
+          "node"                   => run_status.node,
           "node_name"              => run_status.node.name,
           "organization_name"      => organization,
           "resources"              => reporter_data[:resources].map(&:report_data),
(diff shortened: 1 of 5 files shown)


adamleff:
Note this is pending acceptance of chef/chef-rfc#208.

lib/chef/data_collector/messages.rb
```diff
@@ -68,6 +68,7 @@ def self.run_end_message(reporter_data)
           "id"                     => run_status.run_id,
           "message_version"        => "1.0.0",
           "message_type"           => "run_converge",
+          "node"                   => run_status.node,
```
lamont-granquist: you should probably make sure -- explicitly with a spec test -- that the node.run_context and node.run_context.resource_collection are not getting rendered into the json (and that you're actually hitting node.to_json/for_json correctly).

i'm reasonably certain that its behaving correctly, but it would be really horrible if that broke at some point in the future (possibly through some bugginess in JSON rendering).

lamont-granquist: another way to put this is to test that the rendered node object looks like a sane node object rather than just testing that the key is there.

adamleff: Thanks, @lamont-granquist - added tests in 5137634.

jkeiser:
:+1:

adamleff:
@chef/client-core can I get another person to look at this change, please? One more +1 is needed. Thank you!

adamleff:
chef/chef-rfc#208 accepted - awaiting one more maintainer approval.

btm:
:+1:

Issue #4877, opened 2016-04-28
Upstart service provider should provide a way to restart with a stop + start


mconigliaro:
## Description

According to [the Upstart docs](http://upstart.ubuntu.com/cookbook/#restart):

> restart is not the same as running stop followed by start since the restart command will retain the original job configuration whereas stopping the job and restarting it will load the latest job configuration from disk.

This means notifying a service to restart after updating the Upstart job configuration may not work the way you expect. You might be able to work around this problem with the `restart_command` property, but it would be a lot simpler if something like a `:stop_start` action was available.

This was previously reported in [CHEF-4829](https://tickets.opscode.com/browse/CHEF-4829).
## Chef Version

```
# chef-client --version
Chef: 12.9.38
```
## Platform Version

```
# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.3 LTS
Release:    14.04
Codename:   trusty
```
## Replication Case

Updating the template will cause the service to be restarted, but not with the updated job configuration:

```
template '/etc/init/foo.conf' do
  notifies :restart, 'service[foo]'
end

service 'foo'
```

coderanger:
Obligatory plug for poise-service: https://github.com/poise/poise-service/blob/master/lib/poise_service/service_providers/upstart.rb#L38-L46

That said, you can do this:

```ruby
template '/etc/init/foo.conf' do
  notifies :stop, 'service[foo]'
  notifies :start, 'service[foo]'
end
```

mconigliaro:
@coderanger 👍  for the slightly better workaround. After some Googling, it looks like maybe `supports restart: false` might also force a stop + start?

thommay:
Fixed in #5063

PR #5066, opened 2016-06-29
Added stop_start action to service resource

anniehedgpeth:
> #4877: Upstart service provider should provide a way to restart with a stop + start 

I added a stop_start action to the service provider, added a unit test and tested it, and added stop_start to the allowable actions.

I did not run it because I'm not certain of the process, therefore there was no functional testing done for this change. 

Additionally, I did not document the change because I was uncertain of where to do so. 

This is my first pull request into Chef, and I'm new to Ruby, so any guidance would be much appreciated.

### DIFF
--- a/lib/chef/provider/service.rb
+++ b/lib/chef/provider/service.rb
@@ -161,7 +161,7 @@ class Chef
         load_new_resource_state
         @new_resource.running(false)
       end
-      
+
       def action_stop_start
         converge_by("stop then start service #{@new_resource}") do
           stop_service
@@ -172,7 +172,7 @@ class Chef
         load_new_resource_state
         @new_resource.running(true)
       end
-      
+
       def action_restart
         converge_by("restart service #{@new_resource}") do
           restart_service
--- a/lib/chef/provider/service.rb
+++ b/lib/chef/provider/service.rb
@@ -161,7 +161,18 @@ class Chef
         load_new_resource_state
         @new_resource.running(false)
       end
-
+      
+      def action_stop_start
+        converge_by("stop then start service #{@new_resource}") do
+          stop_service
+          Chef::Log.info("#{@new_resource} stopped")
+          start_service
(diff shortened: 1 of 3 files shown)


coderanger:
This is one of those cases where it's hard to wrangle a single `service` interface for multiple underlying tools. For `poise_service` I made the `:restart` action do stop+start on Upstart because I couldn't think of any case where Upstart's behavior is actually better, but then that might be surprising to people that don't know Upstart well :-/

anniehedgpeth:
@coderanger I'm taking easy issues that are approved in the chef queue and coding them as a way to practice my newfound ruby/chef skills. Since the issue was approved I thought I could implement it as suggested. Do you have any specific suggestions for what I should change?

anniehedgpeth:
How would I know if the AppVeyor build failed because of my changes?

coderanger:
@lamont-granquist Do you have opinions on this approach vs #5063? Both have downsides, I'm not sure which I like better for the long term but we should probably go with one of them.

lamont-granquist:
i tend to like #5063 better because it doesn't introduce another public API, and seems to be more in line with what we do elsewhere with stop/start.  i will freely admit i don't know if systemd has some bizzare edge case where this approach is actually better though...

coderanger:
I think I'm leaning that way too, this is more flexible but it relegates the "probably what you meant" behavior to a secondary API that is harder to find.

anniehedgpeth:
No problems here. I was honestly just using it for practice, so you can use your best judgment.

thommay:
Hi @anniehedgpeth , thank you very much for your contribution.
It's really weird that two people chose to attack the same bug on pretty much the same day, but in this case I think we prefer #5063 just because it's a bit more in line with user expectations.
Thanks again, and I look forward to seeing more PRs from you in future.
Cheers,
-Thom

Issue #2304, opened 2014-10-26
yum groupinstall


ror6ax:
Have found this on jira https://tickets.opscode.com/browse/CHEF-1392 in opened state.
Looks like it was not migrated here. 

Will have a look at it myself, but if anybody has valuable insights - feel free to guide the lost soul.

smurawski:
Channeling @someara - Group installs are a tough thing to track because it is difficult to verify what is in the group versus what is currently installed on the box.

ror6ax:
Do we necessary need to track? I can live without this functionality, just curious.

smurawski:
@ror6ax I believe so.  Yum isn't my strong point, but @someara had some thoughts there - as this has come up a lot in the past.  He'll chime in on this thread at some point in the near future.

docwhat:
From [CHEF-1392](https://tickets.opscode.com/browse/CHEF-1392) by Matthew Kent:

> I think the components are there for breaking this into it's own yum_group provider - there's a concept of installed and available groups, commands to install and remove them, even updating by running the install again.

I'm no yum expert, but that looks like enough.

thenoid:
:+1: Anyone associated with this come up with a interim solution?

lamont-granquist:
we have no plans to add this functionality to core in the near future, the cookbook looks like right solution here.  i'm going to close this since we're not planning on taking any action with it.

PR #5071, opened 2016-06-30
adding logic for handling multi fileset bff packages

mwrock:
bff packages can include multiple "filesets" under a single "software product". The individual filesets are the actual installed artifact and nodes can cherry pick these filesets or just provide the product name thereby installing all filesets in the product.

As the bff/aix pacjkage provider stands today, it only treats package names that refer to individual filesets idempotently. I can install an entire product but when given the product name, it is not idempotent and will  converge the resource on every run.

This PR attempts to add idempotency to bff install scenarios where users specify a product name that may include multiple filesets to install. Here is the basic logic used:
* Scan the source for all filesets matching provided package name on either product name or fileset name
* Filter this list to the latest version in case there are more than one
* Scan the filesets installed for those matching provided package name
* filter these installed filesets to those that match the filesets in the source
Now if all filesets installed are the same version as source, Rejoice and do nothing. Otherwise, converge.

In the process of tacking this issue, a couple other potential (likely edge) bugs are addressed:
1. source device locations may have multiple versions and now we ensure the latest is chosen
2. We were searching candates using a `*<package_name>` pattern potentially finding false matches

### DIFF
--- a/lib/chef/provider/package/aix.rb
+++ b/lib/chef/provider/package/aix.rb
@@ -102,6 +102,9 @@ class Chef
               @candidate_version = candidate
               new_resource.version(candidate)
               Chef::Log.debug("#{new_resource} setting install candidate version to #{@candidate_version}")
+              candidate
+            else
+              nil
             end
           end
         end
@@ -146,6 +149,10 @@ class Chef
           end
           ret.stdout.each_line do |line|
             case line
+            # installp will return lines with product:file_set:version
+            # So we are searching for products or filesets that match the package name
+            # Example line:
+            # samba.base:samba.base.rte:1.6.0.3::I:C:::::N:Network Authentication Service Client::::0::
             when /(^|:)#{new_resource.package_name}:/
               fields = line.split(":")
               file_sets.push([fields[1], fields[2]])
--- a/lib/chef/provider/package/aix.rb
+++ b/lib/chef/provider/package/aix.rb
@@ -34,13 +34,13 @@ class Chef
         def define_resource_requirements
           super
           requirements.assert(:install) do |a|
-            a.assertion { @new_resource.source }
-            a.failure_message Chef::Exceptions::Package, "Source for package #{@new_resource.name} required for action install"
+            a.assertion { new_resource.source }
+            a.failure_message Chef::Exceptions::Package, "Source for package #{new_resource.name} required for action install"
           end
           requirements.assert(:all_actions) do |a|
-            a.assertion { !@new_resource.source || @package_source_found }
-            a.failure_message Chef::Exceptions::Package, "Package #{@new_resource.name} not found: #{@new_resource.source}"
-            a.whyrun "would assume #{@new_resource.source} would be have previously been made available"
+            a.assertion { !new_resource.source || @package_source_found }
+            a.failure_message Chef::Exceptions::Package, "Package #{new_resource.name} not found: #{new_resource.source}"
+            a.whyrun "would assume #{new_resource.source} would be have previously been made available"
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/aix.rb
```diff
@@ -83,21 +96,14 @@ def load_current_resource
...
-          ret = shell_out_with_timeout("installp -L -d #{@new_resource.source}")
-          ret.stdout.each_line do |line|
-            case line
-            when /\w:#{Regexp.escape(@new_resource.package_name)}:(.*)/
-              fields = line.split(":")
-              @candidate_version = fields[2]
-              @new_resource.version(fields[2])
-              Chef::Log.debug("#{@new_resource} setting install candidate version to #{@candidate_version}")
+          @candidate_version ||= begin
+            candidate, fs = filter_source_device
```
btm: if we don't have "candidate" we're leaving `@candidate_version` as nil?

lib/chef/provider/package/aix.rb
```diff
@@ -83,21 +96,14 @@ def load_current_resource
...
+            if candidate
+              @candidate_version = candidate
+              new_resource.version(candidate)
+              Chef::Log.debug("#{new_resource} setting install candidate version to #{@candidate_version}")
```
btm: I don't think this is ever setting `@candidate_version`, I think you're getting mostly the expected behavior because of `new_resource.version(candidate)` but line 102 where `@candidate_version = candidate` is going to get overwritten with whatever `Chef::Log.debug` returns.

```
Is this really setting `@candidate_version`? I see youirb(main):025:0> @foo
=> nil
irb(main):026:0> @foo ||= begin
irb(main):027:1* @foo = "the candidate version"
irb(main):028:1> puts "this is a debug statement"
irb(main):029:1> end
this is a debug statement
=> nil
irb(main):030:0> @foo
=> nil
```

lib/chef/provider/package/aix.rb
mwrock: that's correct and the same as before. if no matching filesets are in source, candidate is not set

lib/chef/provider/package/aix.rb
mwrock: yup you are right. will also add a unit test so thats propperly caught

lib/chef/provider/package/aix.rb
btm: gotcha. I'd recommend making this explicit by having the conditional below have an else statement that returns nil. That would help avoid the circumstance like below where we're setting `@candidate_version` to something unintended.

lib/chef/provider/package/aix.rb
```diff
@@ -108,28 +114,63 @@ def candidate_version
...
+          return unless new_resource.source
+
+          file_sets = []
+          ret = shell_out_with_timeout("installp -L -d #{new_resource.source}")
+          unless ret.exitstatus == 0
+            raise Chef::Exceptions::Package, "installp -L -d #{new_resource.source} - #{ret.format_for_exception}!"
+          end
+          ret.stdout.each_line do |line|
+            case line
+            when /(^|:)#{new_resource.package_name}:/
```
btm: I have a personal preference for including an example of a line that we're matching as a comment above a regex like this. It helps make your intentions more clear to the next person.

mwrock:
fixed `candidate_version` @btm and added a unit test to cover that. Also added some clarification and example on the regex

lamont-granquist:
I'm 👎  on putting this directly in the package provider and complicating the idempotency checks particularly when we think about combining this with mutipackage.

This provider doesn't have multipackage yet, but ideally it should.  And then we need to add group support to yum/dnf installs as well.

I don't want to see this go into base package providers since they're already more complicated than they need to be.

We've already got tickets open like #2304, https://tickets.opscode.com/browse/CHEF-1392, https://tickets.opscode.com/browse/CHEF-3462 and there's a cookbook that claims to do yum group installs:

https://supermarket.chef.io/cookbooks/yumgroup

I'm pretty certain we've closed similar issues to try to add group installs directly into the yum provider before as wontfix or dontfix-in-the-package-provider.

scotthain:
I'm with @lamont-granquist on this one. Multiple fileset installation fits the 'group' install pattern, and a group provider for the moment may be better suited as a provider in https://github.com/chef-cookbooks/aix

All that aside, fantastic work @mwrock

mwrock:
ok. Yeah that makes sense to me @scotthain and @lamont-granquist. This definitely complicates bff and if a "group provider" has cross packager appeal and it sounds like it does, its the right approach.

We'll probably mark the underlying ticket for this as "Won't Fix" and address groups in a separate PR.

mwrock:
I think what I'll do here is sniff if the `new_package` is not a file_set and then warn that installs will not be idempotent if its not.

mwrock:
closing in favor of #5093

PR #5072, opened 2016-06-30
Depend on rack < 2

tas50:
Rack 2.0 was released today and it requires Ruby 2.2+ now, which is
breaking Travis runs on anything that depends on Chef. This should get
our builds running again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ end
 
 group(:development, :test) do
   gem "simplecov"
-  gem "rack", "< 2.0"
+  gem "rack", "< 2.0" # 2.0 requires Ruby 2.2+
 
   # for testing new chefstyle rules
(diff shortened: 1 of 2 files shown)


coderanger:
👍 Maybe with a comment reminding us to remove that when we migrate to Ruby 2.2 or 2.3?

coderanger:
Also we need to do the same for the cher-zero gem (via Success Slack).

coderanger:
:shipit:

tas50:
https://github.com/chef/chef-zero/pull/219 Fixes chef-zero

tas50:
@chef/client-core can I get some 👍 s on this?

lamont-granquist:
👍   plz to whack merge when it passes

PR #5075, opened 2016-07-01
tweak 3694 warnings

lamont-granquist:
- clean up "ZenMaster" resource
- clean up 3694 detection to use properties
- unlazy the resource_name in the trivial resource check
- fixes an issue with resources-as-definitions pattern emitting
  3694 errors for trivial resources

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -531,8 +531,6 @@ class Chef
       end
     end
 
-    protected
-
     #
     # The options this Property will use for get/set behavior and validation.
     #
@@ -583,6 +581,7 @@ class Chef
         (options.has_key?(:is) && resource.send(:_pv_is, { name => nil }, name, options[:is], raise_error: false))
     end
 
+    # @api private
     def get_value(resource)
       if instance_variable_name
         resource.instance_variable_get(instance_variable_name)
@@ -591,6 +590,7 @@ class Chef
       end
     end
 
+    # @api private
     def set_value(resource, value)
       if instance_variable_name
         resource.instance_variable_set(instance_variable_name, value)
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@chef/client-core fairly simple.  i ran into this with lamont-cookbooks/multipackage#7

lamont-granquist:
@coderanger poise gets angry at this -- but it looks like in your test you're creating an object that just has a method defined on it, and doesn't declare an attribute or a property...

i ran into this with the old ZenMaster resource and converted the methods to properties and figured anyone not using the properties/attributes interface was not following the API guidelines... IDK if you've got a use case i'm not aware of though...

coderanger:
Just that Poise has to support before those APIs existed so I have to test with them. We also support that for Chef though, so you'll need to fix the code :)

lamont-granquist:
But attributes are ancient...  like 0.8.x era or something...  do we seriously need to support syntax back that old?

coderanger:
Which API do you mean? You're using `.properties` which is very new. `def whatever() set_or_return_value end` is definitely something we should support given that until recently it was the only way implement most fancy transforms, lazy values, etc.

lamont-granquist:
oh yeah, good point, all of core chef did that until recently...  meh...

coderanger:
Sorry for buzzkill :) Some day we can have nice things? A few strategic `respond_to?` might be enough here though.

lamont-granquist:
i can revert the cleanup, i didn't actually need that, i suspect that most of the problem.

lamont-granquist:
@coderanger fixed

coderanger:
:+1: Can add this to the "Chef 13 Happy Fun Times" list.

lamont-granquist:
probably not, i think we still need to clean up a bunch of uses of methods directly on resources.  until that gets done we can't do the cleanup here.

PR #5076, opened 2016-07-01
Data Collector server URL validation, and disable on host down

adamleff:
If a user configured data_collector.server_url in their Chef config
as an empty string, we blissfully passed it along to Chef::HTTP
which would eventually raise a TypeError when trying to dup the
URI's host. This change validates the server_url when setting up
the Data Collector and gives helpful error messages to the user.

We also count Errno::EHOSTDOWN as an error worthy of disabling
the Data Collector reporter for a given run if the user so chooses.

### DIFF
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -382,16 +382,16 @@ class Chef
 
       def validate_data_collector_server_url!
         raise Chef::Exceptions::ConfigurationError,
-          "data_collector.server_url is configured but empty. Please supply a valid URL." if data_collector_server_url.empty?
+          "Chef::Config[:data_collector][:server_url] is empty. Please supply a valid URL." if data_collector_server_url.empty?
 
         begin
           uri = URI(data_collector_server_url)
         rescue URI::InvalidURIError
-          raise Chef::Exceptions::ConfigurationError, "data_collector.server_url (#{data_collector_server_url}) is not a valid URI."
+          raise Chef::Exceptions::ConfigurationError, "Chef::Config[:data_collector][:server_url] (#{data_collector_server_url}) is not a valid URI."
         end
 
         raise Chef::Exceptions::ConfigurationError,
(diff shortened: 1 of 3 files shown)


lib/chef/data_collector.rb
```diff
@@ -376,6 +379,20 @@ def detect_unprocessed_resources
       def nested_resource?(new_resource)
         @current_resource_report && @current_resource_report.new_resource != new_resource
       end
+
+      def validate_data_collector_server_url!
+        raise Chef::Exceptions::ConfigurationError,
+          "data_collector.server_url is configured but empty. Please supply a valid URL." if data_collector_server_url.empty?
```
btm: "is configured but empty" sounds a little technical, can we just say "is empty." ?

lib/chef/data_collector.rb
```diff
@@ -376,6 +379,20 @@ def detect_unprocessed_resources
...
+
+        begin
+          uri = URI(data_collector_server_url)
+        rescue URI::InvalidURIError
+          raise Chef::Exceptions::ConfigurationError, "data_collector.server_url (#{data_collector_server_url}) is not a valid URI."
```
btm: Instead of `data_collector.server_url` in the error messages, how about `Chef::Config[:data_collector][:server_url]`? That would infer that it should be in the config file.

lib/chef/data_collector.rb
```diff
@@ -376,6 +379,20 @@ def detect_unprocessed_resources
...
+        end
+
+        raise Chef::Exceptions::ConfigurationError,
+          "data_collector.server_url (#{data_collector_server_url}) is a URI with no host. Please upply a valid URL." if uri.host.nil?
```
btm: typo *supply

mwrock:
:+1:

btm:
:+1: with nits

thommay:
@btm got all the nits ; 👍 once they're done

adamleff:
appveyor error appears unrelated, internal discussion agrees we can merge, so merge I shall...

PR #5079, opened 2016-07-04
Fix #5078 using cwd parameter instead of Dir.pwd

### LINKED ISSUE #5078: Default location for zero mode
## Description

Chef-solo default to zero mode from 12.11.18 (ref #4919).
Issue linked to Berkshelf [#1483](https://github.com/berkshelf/berkshelf/issues/1483)  and Chef client's #4186

The location to untar the tarball is current directory and not solo default of `/var/chef/cookbooks` and it can overwrite owner and modes of parents directory. 

## Chef Version

Since 12.11.18 with chef-solo mode (not harmfull when using legacy mode as the cookbook path is defaulted to `/var/chef/cookbooks`)

## Platform Version

Any

## Replication Case

Create a tarball with berks package, see berkshelf issue [#1483](https://github.com/berkshelf/berkshelf/issues/1483)

## Client Output

Working directory inherit tarball `.` permissions, creating side effects on target system when tarball user, group and permissions are restricted.

## (Debatable) Propositions:

1. Use tar parameters to drop user and modes from tarball (safest for target node IMHO) ([Here if I read properly](https://github.com/chef/chef/blob/7e3a0aa1ed3559dd599b95cfb90599e2aeedbe6f/lib/chef/application/client.rb#L337):
     ```
     --no-same-owner
           extract files as yourself (default for ordinary users)

     --no-same-permissions
           apply the user's umask when extracting permissions from the archive (default for ordinary users)
     ```
    Adding those two parameters at end of line should do.

2. Exclude `.` from extraction:
[Here again](https://github.com/chef/chef/blob/7e3a0aa1ed3559dd599b95cfb90599e2aeedbe6f/lib/chef/application/client.rb#L337)

    with `--exclude '.'` at end of line

3. Make chef-zero use a defined default path (/var/chef/repo for example) instead of current directory if run as root without configuration:

I assume this should be done [here](https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L128-L142)

##In my point of view
 - 1. is the best option as it prevents chef from modifying permissions by a corrupted tarball,drawbak is that I didn't ensure which versions of tar allow those parameters.
 - 2. is the safe one but we still write the tarball anywhere on the target system which could be abused (but if the tarball is compromised, it could do a lot from the recipes within anyway).
 - 3. Is probably the most conservative one but also the more complex to implement and the most subject to break someone use case soemwhere in my opinion.

## Workarounds:

1. Run chef-solo with `--legacy` command line option to get the previous behavior (using `/var/chef/cookbooks` as target)
2. Create `§etc§chef/client.rb` file with content:
`chef_repo_path '/var/chef/repo'`
3. (Highly discouraged) Monkey patch chef to avoid this behavior

## Conclusion

Even if fixing the untar of the recipes tarball won't guarantee there won't be malicious code executed, it makes harder to corrupt a system than a simple tar file with malicious relative path.

Root cause is within berkshelf `Archive::Tar::Minitar` call [here](https://github.com/berkshelf/berkshelf/blob/master/lib/berkshelf/packager.rb#L45), sadly the issue is there for more than one year now and I'm unsure whether it should be a double fix (Archive and berkshelf) or only berkshelf (probabaly using `./*` instead of `.` would do the trick). 

P.S: I'll try to bake some PR for each behavior.

Tensibai:
Ref #5078

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -132,8 +132,8 @@ module ChefConfig
       until File.directory?(PathHelper.join(path, "cookbooks")) || File.directory?(PathHelper.join(path, "cookbook_artifacts"))
         new_path = File.expand_path("..", path)
         if new_path == path
-          ChefConfig.logger.warn("No cookbooks directory found at or above current directory.  Assuming #{Dir.pwd}.")
-          return Dir.pwd
+          ChefConfig.logger.warn("No cookbooks directory found at or above current directory.  Assuming #{cwd}.")
+          return cwd
         end
         path = new_path
       end


thommay:
@chef/client-core this seems like the correct behaviour. chef-client in local mode already passes Dir.pwd as the argument, so behaviour won't change there. 👍

lamont-granquist:
👍

PR #5080, opened 2016-07-05
Use Mixlib::Archive to extract tarballs

thommay:
this allows us to be truely cross platform, and also to ignore unsafe
paths and permissions.

right now this is locked to a specific branch of mixlib-archive, which needs merging before we merge this.


lamont-granquist:
👍

mwrock:
:+1:  after the mixlib-archive merge

PR #5082, opened 2016-07-05
In Travis use latest ruby releases + trusty

tas50:
The nice rvm aliases in Travis (2.1, 2.2, etc) have never been updated
and point to some really old releases. We should specify the latest
releases of these feature releases. Also if we require sudo then we're
running the GCE environment and we can also specify that we want to run
on the Trusty "beta" image which has much newer deps all around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 sudo: false
 cache: bundler
+dist: trusty # this is only honored where sudo is true also
 
 # Early warning system to catch if Rubygems breaks something
 before_install:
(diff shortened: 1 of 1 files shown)


coderanger:
I spent some time yesterday tracking this down https://github.com/travis-ci/travis-rubies/issues/19

.travis.yml
```diff
@@ -30,35 +30,39 @@ env:
...
-  - rvm: 2.3.0
+  - rvm: 2.3.1
     sudo: true
+    dist: trusty
     script: sudo -E $(which bundle) exec rake spec;
     # also remove integration / external tests
     bundler_args: --without changelog development docgen guard integration maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
   - rvm: rbx
     sudo: true
+    dist: trusty
```
lamont-granquist: can we move `dist: trusty` to the global part and then override later if we need precise?

tas50: Yes.  We can actually specify trusty everywhere, but it will only take effect where we're requiring sudo (GCE vs. containers)

lamont-granquist:
i think this is now colliding with work that keiser is doing to be doing, at any rate should probably get done after that lands...

PR #5085, opened 2016-07-06
WIP: include top level gemfile for chef-config tests

mwrock:


chef-config/chef-config.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rake", "~> 10.0"
 
   %w{rspec-core rspec-expectations rspec-mocks}.each do |rspec|
-    spec.add_development_dependency(rspec, "~> 3.2")
+    spec.add_dependency(rspec, "~> 3.4")
```
lamont-granquist: strictly speaking this is wrong, though, the chef-config gem has no runtime dependency on rspec.

mwrock: I'll mark this wip. It's all very experimental/mashing fingers on keyboard now

Issue #5088, opened 2016-07-06
knife-cookbook-site-share and knife-supermarket producing 404 errors


loganmeetsworld:
## Description

When I use the command `knife cookbook site share -o <location> <cookbook-name> -k <key-location> -VV` or the same command with `knife supermarket share`, I get the error: 
`FATAL: Unable to reach Supermarket: 404 "Not Found". Increase log verbosity (-VV) for more information.`

## Chef Version

chef -v: 
`Chef Development Kit Version: 0.15.16
chef-client version: 12.12.14
delivery version: master (444effdf9c81908795e88157f01cd667a6c43b5f)
berks version: 4.3.5
kitchen version: 1.10.0`

## Platform Version

Mac 10.11.5

## Replication Case

1. Uninstall knife supermarket: `gem uninstall knife-supermarket`
2. Install appbundle-updater: `chef gem install appbundle-updater`
3. Install chef v12.12.13: `sudo chef exec appbundle-updater chefdk chef v12.12.13`
4. Run  `knife cookbook site share -o <location> <cookbook-name> -k <key-location> -VV`
5. In terraformed supermarket knife.rb, point supermarket-site to the public ip of the terraform supermarket. 

## Stacktrace

Short: 
`FATAL: Unable to reach Supermarket: 404 "Not Found". Increase log verbosity (-VV) for more information.`

All output: 
`DEBUG: Chef::HTTP calling Chef::HTTP::JSONInput#handle_request
DEBUG: Chef::HTTP calling Chef::HTTP::JSONOutput#handle_request
DEBUG: Chef::HTTP calling Chef::HTTP::CookieManager#handle_request
DEBUG: Chef::HTTP calling Chef::HTTP::Decompressor#handle_request
DEBUG: Chef::HTTP calling Chef::HTTP::RemoteRequestID#handle_request
DEBUG: Chef::HTTP calling Chef::HTTP::ValidateContentLength#handle_request
DEBUG: Initiating GET to https://supermarket.chef.io/api/v1/cookbooks/dino
DEBUG: ---- HTTP Request Header Data: ----
DEBUG: Accept: application/json
DEBUG: Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
DEBUG: X-REMOTE-REQUEST-ID: 8f85fe7f-e416-4093-b160-a52f8476834b
DEBUG: ---- End HTTP Request Header Data ----
DEBUG: ---- HTTP Status and Header Data: ----
[...]

PR #5091, opened 2016-07-07
The user must specify a category for a new cookbook

thommay:
Bonus fix for a hardcoded URL.

Fixes: #5088, #5092, #5095

### DIFF
--- a/lib/chef/knife/cookbook_site_share.rb
+++ b/lib/chef/knife/cookbook_site_share.rb
@@ -113,19 +113,13 @@ class Chef
       end
 
       def get_category(cookbook_name)
-        begin
-          data = noauth_rest.get("https://supermarket.chef.io/api/v1/cookbooks/#{@name_args[0]}")
-          if !data["category"] && data["error_code"]
-            ui.fatal("Received an error from Supermarket: #{data["error_code"]}. On the first time you upload it, you are required to specify the category you want to share this cookbook to.")
-            exit(1)
-          else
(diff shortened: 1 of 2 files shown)


thommay:
This is one of those "did that ever work", @chef/client-core

thommay:
In common with Stove, let's just say "Other" if it's unspecified, since we don't really care anymore

lamont-granquist:
since we don't care could we force everything to "Other" and avoid the rescue/404 logic entirely and null out the command line option?

thommay:
I didn't want to do that just in case people are actually using categories

lib/chef/knife/cookbook_site_share.rb
```diff
@@ -113,19 +113,16 @@ def run
...
+        data["category"]
+      rescue => e
+        if e.response.code == "404"
+          "Other"
+        else
           ui.fatal("Unable to reach Supermarket: #{e.message}. Increase log verbosity (-VV) for more information.")
           Chef::Log.debug("\n#{e.backtrace.join("\n")}")
-          exit(1)
         end
+        exit(1)
```
coderanger: This seems like it is in the wrong place, don't we want to return the `"Other"`?

lamont-granquist:
so this is where knife-supermarket got fixed:

https://github.com/chef/knife-supermarket/pull/10/files

it just hardcoded an empty-string category and didn't give a choice.

PR #5093, opened 2016-07-07
Warn if not installing an individual bff fileset

mwrock:
bff packages can include multiple "filesets" under a single "software product". The individual filesets are the actual installed artifact and nodes can cherry pick these filesets or just provide the product name thereby installing all filesets in the product.

As the bff/aix pacjkage provider stands today, it only treats package names that refer to individual filesets idempotently. I can install an entire product but when given the product name, it is not idempotent and will converge the resource on every run.

#5071 was an attempt to "fix" this but it was agreed that shoehorning a "group" concept into this provider was not the right path. So this PR basically extends an olive branch to users who wonder why their packages always converge. It warns them that the package is not an individual fileset and will not be idempotent and suggests that they install by fileset instead.

### DIFF
--- a/lib/chef/provider/package/aix.rb
+++ b/lib/chef/provider/package/aix.rb
@@ -55,7 +55,11 @@ class Chef
               ret = shell_out_with_timeout("installp -L -d #{@new_resource.source}")
               ret.stdout.each_line do |line|
                 case line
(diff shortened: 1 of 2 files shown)


scotthain:
👍

btm:
:+1:

PR #5097, opened 2016-07-08
Fix #5094 node.default_unless issue in 12.12.13

### LINKED ISSUE #5094: _unless attribute methods broken in chef 12.12.13
## Description

_unless attribute methods broken in chef 12.12.13

## Chef Version

12.12.13

## Platform Version

Centos 6, Centos 7

## Replication Case

$ git clone https://github.com/acondrat/chef-12.12.13-issue
$ cd chef-12.12.13-issue/wrappercookbook/
$ bundle
$ kitchen converge

There are two test kitchen instances:
default-centos-chef-121118 - converged successful
default-centos-chef-121213 - failed

lamont-granquist:
closes #5094

### DIFF
--- a/lib/chef/node/common_api.rb
+++ b/lib/chef/node/common_api.rb
@@ -32,6 +32,7 @@ class Chef
       # - autovivifying / autoreplacing writer
       # - non-container-ey intermediate objects are replaced with hashes
       def write(*args, &block)
+        root.top_level_breadcrumb = nil if respond_to?(:root)
         value = block_given? ? yield : args.pop
         last = args.pop
         prev_memo = prev_key = nil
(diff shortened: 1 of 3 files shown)


lamont-granquist:
the `if respond_to?(:root)` is mildy gross.  its necessary to work around the fact that i'm mixing this "common api" into both vivid mashes and immutable mashes.  i could extend immutable mashes to have the root node and do all this garbage but a.  it isn't necessary for immutable mashes because you can never mutate them so you never have to worry about resetting the cache so they don't have to track their breadcrumb state and b.  the breadcrumb state needs to be burned with fire in the next release and replaced with some proper path tracking.

lamont-granquist:
@chef/client-core reviewable

lamont-granquist:
@jaymzh most likely rm_override and all those APIs were similarly buggy in 12.12.13, so you'll probably want to avoid this release.   i believe by fixing the all the lower level apis, including `#unlink` (which wasn't strictly necessary to fix #5094) that i'll have fixed that behavior as well...

btm:
:+1:

mwrock:
:+1: on green

PR #5106, opened 2016-07-13
bump gems + cookboks for kitchen-tests

lamont-granquist:


thommay:
Angry Travis is angry

PR #5103, opened 2016-07-11
add back method_missing support to set_unless

lamont-granquist:
closes #5102

PR #5107, opened 2016-07-13
Bump to 12.13

ksubrama:


lib/chef/version.rb
```diff
@@ -21,7 +21,7 @@
 
 class Chef
   CHEF_ROOT = File.expand_path("../..", __FILE__)
-  VERSION = "12.12.24"
+  VERSION = "12.13.1"
```
btm: 12.13.1?

coderanger:
I think we are planning another 12.12 bug fix still.

lib/chef/version.rb
ksubrama: Oops..  fixed.

btm:
@coderanger We're looking at what's easiest to get through the pipeline right now, well, @ksubrama is, since many of us are at chefconf. #5103 could ship in a 12.12.15+ fork or a 12.13.x build off master.

We need to bump minor on master anyway, regardless of how we get #5103 out, because master is getting unreleased features in it. (we're a little stuck on when is the best time to bump given that development continues after a "release candidate build" is triggered, but that may not be the promoted build either.)

coderanger:
I guess? It seems like fixing our reversions should be higher priority than worrying about new features or 12.13 since we have a release that is live and this is causing harm.

lamont-granquist:
we're either cherrypicking to produce a 12.12.x build or we're shipping a 12.13.x aren't we, so this change doesn't block either one?

i'm also fine with just moving forwards and fixing the regression in the 12.13.x release, that seems reasonable.

ksubrama:
I'm bumping to 12.13 so that autobumper doesn't get confused when I tag a release for 12.12 on the release branch because the bot just goes off of the numerical version in master.

coderanger:
Okay, if this is required for the 12.12 release to move forward 👍 but that seems weird :)

PR #5108, opened 2016-07-13
Save acceptance test logs when failing

ksubrama:
Modified:  I'm turning off all acceptance tests for now because they fail intermittently.


ksubrama:
@adamedx @PrajaktaPurohit

PR #5109, opened 2016-07-14
Fixed regex for zypper version 1.13.*. 

yeoldegrove:
This is backward compatible with zypper version 1.12.* (used by SLES11SP4).


lamont-granquist:
needs tests added for the new format somewhere around here:

https://github.com/chef/chef/blob/master/spec/unit/provider/package/zypper_spec.rb#L75-L89

yeoldegrove:
Do we need tests for both formats or do we simply replace the old tests?

yeoldegrove:
Tests added.

lamont-granquist:
👍

thommay:
👍

PR #5111, opened 2016-07-14
adding chef_data_bag_item method to load cheffish

deepsihag:


thommay:
Thanks for this PR, as it turns out there was another PR to fix this one and for various reasons we took that one.

Issue #5113, opened 2016-07-15
knife bootstrap --node-ssl-verify-mode


dldinternet:
## Description

The command line argument is not passed down and placed into the /etc/chef/client.rb

## Chef Version

12.12.15

## Platform Version

Ubuntu 14.04 LTS

## Replication Case

Use a self-signed certificate on the Chef server
Spin a new instance and ensure ability to SSH to the instance
Issue a knife bootstrap sequence and use --node-ssl-verify-mode
After convergence look in /etc/chef/client.rb on the instance.

## Client Output

The client fails SSL cert verification when a self-signed cert is used on the Chef server

## Stacktrace

N/A

coderanger:
Relevant code is here https://github.com/chef/chef/blob/96f9995561bdcf6e5af06303ae4cb4c7e8b876f8/lib/chef/knife/core/bootstrap_context.rb#L89-L110 and looks correct to me. Also really please don't use that option. We provide `knife ssl fetch` to at least make it mostly easy to manage self-signed certs w.r.t to the Chef Server.

dldinternet:
No that is not the code with the problem. The problem is here https://github.com/chef/chef/blob/96f9995561bdcf6e5af06303ae4cb4c7e8b876f8/lib/chef/knife/bootstrap.rb#L219-L227 
The Proc does not return the option value back to the option parser

PR #5114, opened 2016-07-15
Ensure that the valid option is given back to the option parser

dldinternet:
This fixes https://github.com/chef/chef/issues/5113


coderanger:
Good catch, thanky!

coderanger:
(should be merged on green)

thommay:
👍

PR #5115, opened 2016-07-15
launchd: start_calendar_interval validity checking

ErebusBat:
I found that the values produced for`StartCalendarInterval` were not actually correct and that there are some gotchas with this key.

This PR:
- Adds the ability to pass an array of hashes (in addition to the pre-existing single hash) for multiple start intervals.
- Adds sanity check to make sure that your key names are correct; they are incorrect in the current documentation/examples and launchd will silently ignore invalid keys
- Adds sanity check to make sure that launchd won't silently ignore your values (and treat them as wildcards)

According to `man 5 launchd.plist`:

>  StartCalendarInterval [is a] dictionary of integers or array of dictionaries of integers
>    ... Missing arguments are considered to be wildcard.

What the man page doesn't state, but what was observed (OSX 10.11.5, launchctrl v3.4.0) is that key that are specified, _but with invalid values_, will also be silently treated as a wildcard values.

This means that an entry like: `{ "Hour"=>0, "Weekday"=>"6-7"}` will not just run on midnight of Sat and Sun, rather it will run _every_ midnight.

We check to make sure that the values are integers and not strings, adhering to the manpage documentation.

This also ensures that only keys that `StartCalendarInterval` supports will be passed.
This eliminates passing a key (like `:hour` or `'hourly'`) and thinking it works when in reality launchd is just treating it as a wildcard.

NOTE: This is a "breaking" change in that: it will break existing cookbooks that are using the current docs: https://docs.chef.io/resource_launchd.html (See `start_calendar_interval`); however the generated plist doesn't appear to actually do what was intended anyway (at least on El Capitan).

### DIFF
--- a/lib/chef/resource/launchd.rb
+++ b/lib/chef/resource/launchd.rb
@@ -68,7 +68,7 @@ class Chef
 
         # Make sure the hashes don't have any incorrect keys/values
         array.each do |entry|
-          allowed_keys = %w(Minute Hour Day Weekday Month)
+          allowed_keys = %w{Minute Hour Day Weekday Month}
           error_msg = "Invalid key for start_calendar_interval, must be one of: #{allowed_keys.join(", ")}"
           if entry.keys.any? { |key| !allowed_keys.include?(key) }
             raise Chef::Exceptions::ValidationFailed, error_msg
--- a/lib/chef/resource/launchd.rb
+++ b/lib/chef/resource/launchd.rb
@@ -46,6 +46,48 @@ class Chef
       property :source, String
       property :session_type, String
 
+      # StartCalendarInterval has some gotchas so we coerce it to help sanity
+      # check.  According to `man 5 launchd.plist`:
+      #   StartCalendarInterval <dictionary of integers or array of dictionaries of integers>
+      #     ... Missing arguments are considered to be wildcard.
+      # What the man page doesn't state, but what was observed (OSX 10.11.5, launchctrl v3.4.0)
+      # Is that keys that are specified, but invalid, will also be treated as a wildcard
+      # this means that an entry like:
+      #   { "Hour"=>0, "Weekday"=>"6-7"}
+      # will not just run on midnight of Sat and Sun, rather it will run _every_ midnight.
+      property :start_calendar_interval, [Hash, Array], coerce: proc { |type|
+        # Coerce into an array of hashes to make validation easier
+        array = type   if type.is_a?(Array)
+        array = [type] unless array
+
+        # Check to make sure that our array only has hashes
+        if array.any? { |obj| !obj.is_a?(Hash) }
+          error_msg = "start_calendar_interval must be a single hash or an array of hashes!"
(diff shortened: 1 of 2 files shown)


thommay:
The code looks fine, but it would be good to get some verification that the values have not changed from previous versions of OS X. cc @chef/client-os-x @natewalck

ErebusBat:
If there is an official archive of past versions of the launchd docs I will happily cross check.  

I also have code to correct `KeepAlive` paths and `LowPriorityBackgroundIO` (not ready for a PR yet).  How far back does it need to support?

natewalck:
Tagging @mikedodge04

thommay:
@ErebusBat we support back to 10.8 ( https://docs.chef.io/platforms.html#chef-client-title-foundational-platforms ) . I'm not sure how one might go about finding docs of old versions, though.

lib/chef/resource/launchd.rb
```diff
@@ -46,6 +46,48 @@ def initialize(name, run_context = nil)
...
+
+        # Check to make sure that our array only has hashes
+        if array.any? { |obj| !obj.is_a?(Hash) }
+          error_msg = "start_calendar_interval must be a single hash or an array of hashes!"
+          raise Chef::Exceptions::ValidationFailed, error_msg
+        end
+
+        # Make sure the hashes don't have any incorrect keys/values
+        array.each do |entry|
+          allowed_keys = %w{Minute Hour Day Weekday Month}
```
mikedodge04: Dude this is awesome! Good work!

nmcspadden:
Looks like there's an unofficial copy of the launchd docs here: http://www.manpagez.com/man/5/launchd.plist/osx-10.8.php

lib/chef/resource/launchd.rb
```diff
@@ -46,6 +46,48 @@ def initialize(name, run_context = nil)
...
+      #     ... Missing arguments are considered to be wildcard.
+      # What the man page doesn't state, but what was observed (OSX 10.11.5, launchctrl v3.4.0)
+      # Is that keys that are specified, but invalid, will also be treated as a wildcard
+      # this means that an entry like:
+      #   { "Hour"=>0, "Weekday"=>"6-7"}
+      # will not just run on midnight of Sat and Sun, rather it will run _every_ midnight.
+      property :start_calendar_interval, [Hash, Array], coerce: proc { |type|
+        # Coerce into an array of hashes to make validation easier
+        array = type   if type.is_a?(Array)
+        array = [type] unless array
```
lamont-granquist: `array = [ type ].flatten` works here i think, even if its hash, then you get an array wrapping a hash which is what i think you're trying to get?

can the array contain an array already which is why you can't flatten or something?

lamont-granquist:
one comment, but either way 👍

lib/chef/resource/launchd.rb
lamont-granquist: or `Array.new(type)` i think

PR #5117, opened 2016-07-15
Verify checksum before cookbook is served to user

PrajaktaPurohit:
```
➜  chef git:(praj/FLOW-339/checksums) ✗ bundle exec rspec spec/unit/cookbook/synchronizer_spec.rb
                              :
                              :
                              :
  when syncing cookbooks with the server
    when the cookbook checksum does not match the original checksum
      does not download cookbook and raise error
    when the cache does not contain the desired files
      Chef::Config[:no_lazy_load] is false
        fetches eagerly loaded files
        downloads cookbooks if checksum matches
                             :
                             :
Finished in 0.05532 seconds (files took 1.77 seconds to load)
18 examples, 0 failures
```

REF: https://feedback.chef.io/forums/301644-chef-product-feedback/suggestions/9580350-chef-server-file-integrity

### DIFF
--- a/lib/chef/cookbook/synchronizer.rb
+++ b/lib/chef/cookbook/synchronizer.rb
@@ -252,7 +252,7 @@ class Chef
       # (remote, per manifest), do the update. This will also execute if there
       # is no current checksum.
       if !cached_copy_up_to_date?(cache_filename, file.manifest_record["checksum"])
-        download_file(file.manifest_record["url"], cache_filename)
+        download_file(file, cache_filename)
         @events.updated_cookbook_file(file.cookbook.name, cache_filename)
       else
         Chef::Log.debug("Not storing #{cache_filename}, as the cache is up to date.")
@@ -278,11 +278,16 @@ class Chef
     # Unconditionally download the file from the given URL. File will be
     # downloaded to the path +destination+ which is relative to the Chef file
     # cache root.
-    def download_file(url, destination)
-      raw_file = server_api.streaming_request(url)
-
-      Chef::Log.info("Storing updated #{destination} in the cache.")
-      cache.move_to(raw_file.path, destination)
+    def download_file(file, destination)
+      raw_file = server_api.streaming_request(file.manifest_record["url"])
+      raw_file_md5 = Chef::Digester.generate_md5_checksum_for_file(raw_file.path)
+      Chef::Log.debug("The md5 hash of the cookbook file to be served: #{raw_file_md5}")
+      if raw_file_md5 == file.manifest_record["checksum"]
+        Chef::Log.debug("Storing updated #{destination} in the cache.")
+        cache.move_to(raw_file.path, destination)
+      else
+        raise Chef::Exceptions::FileIntegrityCompromise.new(raw_file.path, file.manifest_record["checksum"], raw_file_md5)
+      end
     end
 
     # Marks the given file as valid (non-stale).
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -520,5 +520,11 @@ This error is most often caused by network issues (proxies, etc) outside of chef
         super "Found multiple matching resources. #{matches_info.join("\n")}"
       end
     end
+
+    class FileIntegrityCompromise < RuntimeError
+      def initialize(file, orig_chksum, current_chksum)
+        super "For the file: #{File.basename(file)} original checksum: (#{orig_chksum}) does not match existing content checksum: (#{current_chksum})"
+      end
+    end
   end
 end
--- a/spec/unit/cookbook/synchronizer_spec.rb
+++ b/spec/unit/cookbook/synchronizer_spec.rb
@@ -213,6 +213,36 @@ describe Chef::CookbookSynchronizer do
            :path => "/tmp/cookbook_a_template_default_tempfile")
   end
 
+  def setup_common_files_missing_expectations_for_downloads
+    # Files are not in the cache:
+    expect(file_cache).to receive(:has_key?).
+      with("cookbooks/cookbook_a/recipes/default.rb").
+      and_return(false)
+    expect(file_cache).to receive(:has_key?).
+      with("cookbooks/cookbook_a/attributes/default.rb").
+      and_return(false)
+
+    # Fetch and copy default.rb recipe
+    expect(server_api).to receive(:streaming_request).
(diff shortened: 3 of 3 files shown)


lib/chef/cookbook/synchronizer.rb
```diff
@@ -278,11 +278,16 @@ def cached_copy_up_to_date?(local_path, expected_checksum)
...
-      cache.move_to(raw_file.path, destination)
+    def download_file(file, destination)
+      raw_file = server_api.streaming_request(file.manifest_record["url"])
+      raw_file_md5 = Chef::Digester.generate_md5_checksum_for_file(raw_file.path)
+      Chef::Log.info(raw_file_md5)
+      if raw_file_md5 == file.manifest_record["checksum"]
+        Chef::Log.info("Storing updated #{destination} in the cache.")
+        cache.move_to(raw_file.path, destination)
+      else
+        Chef::Log.error("File integrity has been compromised. \n Original checksum: #{file.manifest_record["checksum"]} \n Current checksum: #{raw_file_md5}")
```
coderanger: This needs to raise an exception at a minimum.

coderanger:
I'm guessing this is another one from a customer request? Please make sure to include all the relevant data for those of us that can't see Chef's internal ticketing systems.

lib/chef/cookbook/synchronizer.rb
PrajaktaPurohit: agreed!

thommay:
👍

lib/chef/cookbook/synchronizer.rb
```diff
@@ -278,11 +278,16 @@ def cached_copy_up_to_date?(local_path, expected_checksum)
...
     # cache root.
-    def download_file(url, destination)
-      raw_file = server_api.streaming_request(url)
-
-      Chef::Log.info("Storing updated #{destination} in the cache.")
-      cache.move_to(raw_file.path, destination)
+    def download_file(file, destination)
+      raw_file = server_api.streaming_request(file.manifest_record["url"])
+      raw_file_md5 = Chef::Digester.generate_md5_checksum_for_file(raw_file.path)
+      Chef::Log.info(raw_file_md5)
```
mwrock: I'd log this to debug and not info and also state it is the file md5

lib/chef/cookbook/synchronizer.rb
```diff
@@ -278,11 +278,16 @@ def cached_copy_up_to_date?(local_path, expected_checksum)
...
+      if raw_file_md5 == file.manifest_record["checksum"]
+        Chef::Log.info("Storing updated #{destination} in the cache.")
```
mwrock: also debug

lib/chef/exceptions.rb
```diff
@@ -520,5 +520,11 @@ def initialize(resources_found)
         super "Found multiple matching resources. #{matches_info.join("\n")}"
       end
     end
+
+    class FileIntegrityCompromise < RuntimeError
+      def initialize(orig_cksum, current_cksum)
+        super "Original checksum for cookbook file (#{orig_cksum}) does not match existing content checksum (#{current_cksum})"
```
mwrock: I'd include the file name if its readily available

btm:
:+1: with @mwrock's debugging nits

mwrock:
:+1:

coderanger:
Again, this should _not_ be merged until someone explains why we are doing this. I can't see whatever internal ticket this is referencing.

coderanger:
If this is for checking server-side corruption, that should probably be a server-side process. If this is to avoid truncation attacks/errors we can just check the file size and not pay the CPU cost of the hash. Anyone +1'ing this please explain why you are doing so.

mwrock:
The motivation was voiced in a uservoice request [here](https://feedback.chef.io/forums/301644-chef-product-feedback/suggestions/9580350-chef-server-file-integrity). Its basically to ensure that the file has not been tampered with on the server.

I feel like its a reasonable requsest and this is the right implementation. We want to keep the workload on the server as minimal as possible and performing the md5 on the client is inexpensive on modern processors and only occurs at download time - only when downloading the first time or when a cookbook's version is bumped.

coderanger:
Anyone with root-level file access on the Chef Server that is being malicious is a Game Over level scenario no matter what we do. This doesn't add security but it does burn CPU cycles so I would want to see a lot more of a justification. The user request is reasonable for a user, but the correct answer is to explain to them how Chef Server's security model works and that this isn't a threat we can defend against.

tl;dr This doesn't help, anyone with access to modify these files can also change the MD5 in the database.

coderanger:
If you want to actually implement anti-tamper protections we would have to talk about adding something like GPG signature checking on policyfile objects or cookbook metadata, which is not what this addresses and would be a much more complex change.

coderanger:
Also for bonus points MD5 is unsafe for use in this context by all current security standards, this would violate FIPS I think :)

mwrock:
So technically, I think @coderanger is right - if someone can futz with the files on the server then they can certainly futz with the manifest. This PR just makes things "a bit more secure" and makes it a bit harder to mess with things and its super easy to implement. But it could never be advertised as a  security wall. I'm not as worried about cpu cost of MD5 hashing (not to trivialize that) as I am about the cognitive burden of extra code paths with little value. Even if there might be high "perceived" value.

coderanger:
The value of this kind of check would only be to catch bitrot of files stored on bookshelf. That use case seems like it would be better served by some kind of nightly maintenance script or similar like `chef-server-ctl fsck`.

ksubrama:
Well - with bit rot or network issues or fs issues...  I would think that doing it at the client end is probably more sensible in case a download was truncated or not written out correctly.  If we trust the intermediaries to transfer and persist the file correctly, we shouldn't care.  On the other hand, if we're going so far as to double check it and we have and md5 on hand, might as well do the checksum right?  It's not cryptographically secure - but neither is the TCP packet checksum.

coderanger:
@ksubrama TLS does strongly checksum all messages while in-flight though. Truncation attacks/errors can be checked without nearly as much CPU or IO load by looking at the file size :) Remember the non-zero number of people that host very big files (I've heard of multi-GB JARs) via `cookbook_file` which this would read at startup, a non-trivial slowdown. Given that I don't think I've ever heard of a single case of a truncation error with Chef cookbook transfers, I'm not super worried about that as a risk factor :)

PR #604, opened 2013-01-24
[CHEF-2205] Set '--no-create-home'

johnwilliams:
http://tickets.opscode.com/browse/CHEF-2205

Set '--no-create-home' so we don't attempt to create a home directory on RHEL/CentOS systems

PR #2713, opened 2015-01-03
Fix useradd on centos when manage_home is false.

sathlan:
On Centos 6.4/6.5/6.6/7.0 using this code block fails, provided that the
/tmp/foo directory doesn't exist.

    user 'my_test' do
      supports :manage_home => false
      home '/tmp/foo/bar'
    end

    [2015-01-03T05:54:03+00:00] FATAL: Stacktrace dumped to /tmp/kitchen/cache/chef-stacktrace.out
    Chef Client failed. 0 resources updated in 1.425418119 seconds
    [2015-01-03T05:54:03+00:00] ERROR: user[my_test] (test::default line 1) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '12'
    ---- Begin output of ["useradd", "-d", "/tmp/foo/bar", "my_test"] ----
    STDOUT:
    STDERR: useradd: cannot create directory /tmp/foo/bar
    ---- End output of ["useradd", "-d", "/tmp/foo/bar", "my_test"] ----
    Ran ["useradd", "-d", "/tmp/foo/bar", "my_test"] returned 12
    [2015-01-03T05:54:04+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)

This code runs properly on ubuntu/debian.

Adding the "-M" flag solves the problem.

PR #5122, opened 2016-07-19
user provider manage_home behavior and refactor

lamont-granquist:
This fixes the long-standing bug with user provider manage_home behavior on linux.

In the process it also refactors the user provider into 1:1 resources so we now have e.g. linux_user and aix_user similar now to yum_package and apt_package.

A lot of the resource classes are stubs but it allows us to override behavior in the classes.  Previously to this if we wanted to change the default behavior of the manage_home or supports[:manage_home] setting for linux it would mean changing it for linux, solaris, aix, etc -- or doing silly tricks with nil defaults being remapped in the providers (which is the road i travelled down first before going to all this trouble -- but that starts meaning the resource is lying about state which is awful -- falsey things on the resource magically become truthy things in the provider).

This deliberately breaks a bit of the API where the linux useradd code was the default provider for everyone.  Now users on HP-UX  will get failed resource resolution instead of getting the linux based adduser provider.  That is entirely by design.   We should fail to resolve the user resource on platforms that don't make any sense rather than just giving them the linux code.

Note that this also starts removing the `supports[:manage_home]` capabilities in the resource+provider.  Those date back to a Chef 0.6.x idea from @adamhjk which looks like it was going down the route of having tweakable feature flags to control one monolithic adduser provider.  That is a super useful API for the service provider where on the same box you may have differing init scripts with different capabilities so the chef-user needs to be able to tell chef that an init script `supports[:restart]` or not.  The user provider is strictly platform-dependent, though, and we went through the process of forking off different classes to handle the major different flavors of adduser.  Its also possible that Ubuntu 8.x didn't support some of these flags, but we've long since dropped support for those distros.

And even if there's regressions that occur here, this sets a much better foundation to fixing them:

- if AIX gets upset we can fix that in the aix_user resource or provider directly without trying to fix code in an inherited superclass or single user resource which affects linux and the rest of the distros.

Note that there's still cleanup that needs to occur here.  The base adduser base class needs some work on it to move the linux-specific code to the linux provider and turn the adduser class into a real base class between linux, aix and solaris.  The "shared" examples it turns out have only been running on the linux provider for ages -- so they're now linux-specific but still need to get 
un-refactored out of the class that they're stuffed into.

This may break chefspec users who do not specify a platform and version in tests and then use user resources.   It was basically decided in https://github.com/customink/fauxhai/pull/249 that one way or the other that is correct behavior.  Users who hit that bug just need to specify a platform of 'ubuntu' and version of '16.04' or something like that one way or another.  It may be easiest to follow the chefspec README.md directions to setup a sane default platform and platform version in your spec_helper.rb for all examples and copy that to all repos as a best practice.

closes #604


thommay:
travis is grouchy

lib/chef/resource/user.rb
```diff
@@ -42,7 +42,7 @@ def initialize(name, run_context = nil)
         @force = false
         @non_unique = false
         @supports = {
-          :manage_home => false,
+          :manage_home => true,
```
coderanger: This is a compat break, no?

thommay: There's fairly good discussion in #604 as to why this has been done, and I think it's the correct thing to do - it actually preserves current behaviour in most cases, since by default login.defs will cause home directories to be created. 
Is it a compat break? By the book, yes. In practice, I suspect not.

More concretely, looking at ubuntu, CREATE_HOME was added to login.defs in 2009, and is true by default - so this doesn't change behaviour for any of our supported versions.

Red Hat, and hence CentOS, have carried a patch since at least CentOS 5 to add CREATE_HOME and set it to true, so out of the box this doesn't change behaviour.

lib/chef/provider/user/useradd.rb
```diff
@@ -141,6 +141,7 @@ def update_options(field, option, opts)
         def useradd_options
           opts = []
           opts << "-r" if new_resource.system
+          opts << "-M" if !managing_home_dir?
```
arnonhongklay: unless is better?

lib/chef/resource/user.rb
coderanger: :+1: Should make sure to call it out in the release notes though.

lamont-granquist: Yeah, if I understand #604 correctly the problem is that we've always behaved like manage_home was forced to true and the button has never worked.  This fixes it, but we need to set manage_home to false to not have an API break.

thommay:
tests still need fixing up, @lamont-granquist , but merge when green

lamont-granquist:
yeah i got distracted by that DNS fix yesterday

lamont-granquist:
see also #2713

lamont-granquist:
okay now i'm using nil(default)/true/false for values for `supports[:manage_home]` and the `manage_home` setting itself.  the truthiness of the default behavior
has been preserved in the superclass/resource.   there's a 3x3 test matrix now that asserts that we're getting the setting right.

lamont-granquist:
dammit, ninja'd myself

lamont-granquist:
@chef/client-core reviewage?  there are some massive FIXMEs here, but i don't really want to spend the next 2 weeks refactoring all the user providers and then dump one 3,000 line diff into everyone's laps...

lamont-granquist:
meh, rebase screwed it up, but its after 7pm, fixing tomorrow...

thommay:
:+1: LGTM

Issue #2894, opened 2015-02-11
Nameserver changes don't get picked up by Chef


roman-yepishev-enoc:
This bug makes it impossible to do full Chef configuration of the node in case it was not previously configured with DNS enabled to resolve internal resources if these are later referenced in the recipes.

The use case is simple:
 1. Configure nameservers
 2. Fetch some content from a server which hostname can be resolved with nameserver configured previously.

Versions tested:
 * 11.16.2
 * 12.0.3

STR using Test Kitchen:
 1. On host: add '`54.175.219.8 randomhost.chef.io`' to /etc/hosts, effectively making randomhost.chef.io an alias for httpbin.org.
 2. Run dnsmasq
 3. Make sure randomhost.chef.io now resolves:
    ```
$ dig +short randomhost.chef.io @127.0.0.1
54.175.219.8
```
 4. Using centos-6.5 vagrant box, converge the following recipe:
   ```
file '/etc/resolv.conf' do
  content "nameserver 10.0.2.2"
end

remote_file '/tmp/user-agent'  do
  source 'http://randomhost.chef.io/user-agent'
end
```

**Expected results:**
Chef creates /tmp/user-agent with the content from http://httpbin.org/user-agent

**Actual results:**
```
[2015-02-11T16:46:06+00:00] WARN: remote_file[/tmp/user-agent] cannot be downloaded from http://randomhost.chef.io/user-agent: Error connecting to http://randomhost.chef.io/user-agent - getaddrinfo: Name or service not known
       
           ================================================================================
           Error executing action `create` on resource 'remote_file[/tmp/user-agent]'
           ================================================================================
           
           SocketError
           -----------
           Error connecting to http://randomhost.chef.io/user-agent - getaddrinfo: Name or service not known
           
           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/resolver_bug/recipes/default.rb
           
            14: remote_file '/tmp/user-agent'  do
            15:   source 'http://randomhost.chef.io/user-agent'
            16: end
           
           Compiled Resource:
           ------------------
           # Declared in /tmp/kitchen/cache/cookbooks/resolver_bug/recipes/default.rb:14:in `from_file'
           
           remote_file("/tmp/user-agent") do
             provider Chef::Provider::RemoteFile
             action "create"
             retries 0
             retry_delay 2
             guard_interpreter :default
             path "/tmp/user-agent"
             backup 5
             atomic_update true
[...]

PR #5123, opened 2016-07-19
replace glibc resolver with ruby resolver

lamont-granquist:
also wires up file providers to reload /etc/resolv.conf when it changes.

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -28,6 +28,8 @@ require "mixlib/cli"
 require "tmpdir"
 require "rbconfig"
 require "chef/application/exit_code"
+require "resolv"
+require "resolv-replace"
 
 class Chef
   class Application
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -29,7 +29,9 @@ require "tmpdir"
 require "rbconfig"
 require "chef/application/exit_code"
 require "resolv"
-require "resolv-replace"
+# on linux, we replace the glibc resolver with the ruby resolv library, which
+# supports reloading.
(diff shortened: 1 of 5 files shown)


thommay:
😨 :+1:

lamont-granquist:
code somewhat stolen from comment in #2894

coderanger:
Do we need to do anything special for Windows?

lamont-granquist:
appveyor seems to have died in flames, so i think so...

but really we only need this workaround for Linux/glibc AFAIK...

coderanger:
👍

coderanger:
One thing to note: this will break DNS for anyone using NIS for hostname lookups I think. That said, those people can literally DIAF.

lamont-granquist:
Yeah, NIS for hostnames in addition to users and groups.

I'm considering this use case, and I think we fix it when that one person complains.

lamont-granquist:
#5141 is the blocking issue that caused the revert of this PR

maokt:
Instead of replacing the glibc resolver, we could use ffi to tell glibc to reload resolv.conf itself. I've been using this in a ruby_block in recipes where resolv.conf is updated:

```
        require 'ffi'
        module libc
            extend FFI::Library
            ffi_lib FFI::CURRENT_PROCESS
            # res_init in GNU libc is called __res_init, so try both
            [ :res_init, :__res_init ].detect do |ff|
                begin
                    attach_function :res_init, ff, [], :int
                    true
                rescue FFI::NotFoundError
                    false
                end
            end
        end
        libc.res_init() if defined? libc.res_init
```

Issue #4900, opened 2016-05-05
cron resource mailto property should support setting an empty string


griffint61:
## Description

If you don't specify a mailto setting for a cron resource, it inherits whatever MAILTO setting is previously defined in the crontab file, which could be for something completely unrelated.

## Chef Version

12.4.1

## Platform Version

CentOS 6.7

## Replication Case

Use the cron resource twice. In the first case explicitly set the mailto option. In the second case don't set the mailto option at all. In both cases use commands that generate some output.

When the first cron job fires, email will be sent to the specified address, as expected.

When the second cron job fires, mail will be sent to the address specified for the first cron resource, instead of having no mail sent at all, which is likely what was intended.

Note also that you can't workaround this problem by passing an empty string as the mailto setting for second cron resource. According to the crontab(5) man page on CentOS 6, An empty string is a valid MAILTO setting, but the cron resource bombs out when passed an empty mailto string.

## Client Output

N/A

## Stacktrace

N/A

lamont-granquist:
Doing this globally to everyone's crontabs would most likely be a breaking change.

The bug where you can't pass empty string to the cron resource is certainly a bug.

pwalz:
I'd also like to see this one fixed. It looks like it was fixed in #311, but the PR never got merged due to a lack of regression tests.

For the moment, I've worked around the issue by adding shell output redirection (` >/dev/null 2>&1`) to the `command` property of my cron resource.

mconigliaro:
I think this is just how cron works (i.e. setting `MAILTO` causes every subsequent job to inherit that setting). This would be fine except for the fact that Chef does not necessarily write the crontab in the order in which jobs are defined (i.e. new jobs are always just appended to the end of the crontab). This means even if you only want to set `mailto` for a single job, you end up having to explicitly set it back to the default on every other job to ensure that mail gets delivered to the correct recipient.

PR #5127, opened 2016-07-20
Support setting an empty string for cron attrs

thommay:
Fixes: #4900, #311

### DIFF
--- a/lib/chef/provider/cron.rb
+++ b/lib/chef/provider/cron.rb
@@ -237,7 +237,7 @@ class Chef
         newcron = ""
         newcron << "# Chef Name: #{new_resource.name}\n"
         [ :mailto, :path, :shell, :home ].each do |v|
-          newcron << "#{v.to_s.upcase}=#{@new_resource.send(v)}\n" if @new_resource.send(v)
+          newcron << "#{v.to_s.upcase}=\"#{@new_resource.send(v)}\"\n" if @new_resource.send(v)
         end
         @new_resource.environment.each do |name, value|
           newcron << "#{name}=#{value}\n"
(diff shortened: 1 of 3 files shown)


thommay:
cc @chef/maintainers

lib/chef/provider/cron.rb
```diff
@@ -237,7 +237,11 @@ def get_crontab_entry
         newcron = ""
         newcron << "# Chef Name: #{new_resource.name}\n"
         [ :mailto, :path, :shell, :home ].each do |v|
-          newcron << "#{v.to_s.upcase}=#{@new_resource.send(v)}\n" if @new_resource.send(v)
```
jkeiser: This isn't caused by you, but for consistency can you use `new_resource` instead of `@new_resource`?

lib/chef/provider/cron.rb
```diff
@@ -237,7 +237,11 @@ def get_crontab_entry
...
+          if @new_resource.send(v)
+            cron_val = @new_resource.send(v)
+            cron_val = '""' if cron_val.empty?
+            newcron << "#{v.to_s.upcase}=#{cron_val}\n"
```
jkeiser: Should this be using shellwords? That might fix the "" thing as well as any escaping issues ... what if the value has a space in it?

lib/chef/provider/cron.rb
thommay: Pretty much the entire file is using `@new_resource`, so I wanted to stay consistent. There's definite scope for a rewrite, but I wanted to fix this first.

jkeiser: Ah! OK. I just noticed that the previous line (238) didn't.

lib/chef/provider/cron.rb
thommay: The value's allowed to have spaces in it, and doesn't need quoting unless it has leading or trailing spaces that are significant (which we've never supported and no-one's ever noticed :) ).

stevendanna:
LGTM

lib/chef/provider/cron.rb
lamont-granquist: can you:

```
"#{v.to_s.upcase}=\"#{cron_val}\"\n"
```

and avoid the special casing of empty?

PR #5128, opened 2016-07-20
Forward package cookbook_name to underlying remote_file

Annih:
Hello,

The package providers is instantiating the remote_file resource manually, i.e. not using the DSL.
This means the cookbook_name attribute of the resource is not set; it's sometimes useful to have this information.

I patched the windows_package using the DSL (see chef-cookbooks/windows#318), but this time I decided to only forward the cookbook_name to the newly created resources.
That means maybe other attributes may be missing. I can change my PR if you want :)

Regards!

Cc: @aboten

### DIFF
--- a/lib/chef/provider/package/windows.rb
+++ b/lib/chef/provider/package/windows.rb
@@ -201,6 +201,7 @@ class Chef
         def resource_for_provider
           @resource_for_provider = Chef::Resource::WindowsPackage.new(new_resource.name).tap do |r|
             r.source(Chef::Util::PathHelper.validate_path(source_location)) unless source_location.nil?
+            r.cookbook_name = new_resource.cookbook_name
             r.version(new_resource.version)
             r.timeout(new_resource.timeout)
             r.returns(new_resource.returns)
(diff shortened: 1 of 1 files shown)


lib/chef/provider/package/windows.rb
```diff
@@ -201,6 +201,7 @@ def downloadable_file_missing?
         def resource_for_provider
           @resource_for_provider = Chef::Resource::WindowsPackage.new(new_resource.name).tap do |r|
             r.source(Chef::Util::PathHelper.validate_path(source_location)) unless source_location.nil?
+            r.cookbook_name = new_resource.cookbook_name
```
btm: this should match the existing style, i.e. `r.cookbook_name(new_resource.cookbook_name)`

Annih: It's what I tried first, but the tests failed with a `wrong # of arguments(1 for 0)`

coderanger: cookbook_name is not a resource property like the others :)

lamont-granquist:
this should be change to use_inline_resources + declare_resource now.

e.g.:

https://github.com/chef/chef/blob/master/lib/chef/provider/apt_update.rb

after converting it over to use_inline_resources if it is still getting the cookbook_name wrong in the subresource it should then get fixed (and fix it correctly across the board for all core use_inline_resources consumers)

coderanger:
That's a much bigger change, we should still merge this for now IMO. Anyone actually -1 on this?

lamont-granquist:
👍

PR #5131, opened 2016-07-22
Cache ChefFS children

thommay:
We were seeing thousands of calls to `can_have_child?` and `make_child_entry` whilst starting a chef-zero run. In the long term we should investigate why this is so slow, but for now, let's just cache the list of children.

I've made the API for the cache a bit more flexible than it currently needs to be in case we decide to cache more down the line.

Performance:

Previously, with 300 empty cookbooks, all depended upon by another empty cookbook:

```
Chef Client finished, 0/0 resources updated in 03 minutes 15 seconds

real    3m17.167s
user    3m3.185s
sys     0m12.185s
```

Now:
```
Chef Client finished, 0/0 resources updated in 07 seconds

real    0m8.578s
user    0m7.773s
sys     0m0.664s
```

Based on the work done by @youngjl1 . cc @chef/client-core
Fixes: #5038


thommay:
All the kitchen tests will fail until chef/chef-zero#226 is merged and released, thanks to the wonders of git branches and such. Once that happens we can revert 172910323aefd72c36ff486e8c4a1ea94e07f72a .

lamont-granquist:
👍

PR #5133, opened 2016-07-24
Make windows acceptance tests pass again

mwrock:
This restores and fixes windows acceptance tests adding these key changes:
* Uses new winrm-v2 test-kitchen/winrm/inspec branches to avoid intermittent winrm-fs errors
* Increases chef-acceptance's shellout timeout from the 10 minute default to 30m minutes. Top-cookbooks and fips may very well take longer than 10 minutes
* Removes unnecesary winrm config from userdata which was causing the winrm service to restart after a converge was taking place and sometimes failing a run


mwrock:
This produces a green jenkins adhoc build

acceptance/.shared/kitchen_acceptance/.kitchen.ec2.yml
```diff
@@ -213,18 +213,7 @@ platforms:
...
-        & winrm.cmd set winrm/config '@{MaxTimeoutms="1800000"}' >> $logfile
-        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
-        & winrm.cmd set winrm/config/winrs '@{MaxShellsPerUser="50"}' >> $logfile
-        #Server settings - support username/password login
-        & winrm.cmd set winrm/config/service/auth '@{Basic="true"}' >> $logfile
-        & winrm.cmd set winrm/config/service '@{AllowUnencrypted="true"}' >> $logfile
-        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
-        #Firewall Config
-        & netsh advfirewall firewall set rule name="Windows Remote Management (HTTP-In)" profile=public protocol=tcp localport=5985 remoteip=localsubnet new remoteip=any  >> $logfile
+        & netsh advfirewall firewall set rule name="Windows Remote Management (HTTP-In)" profile=public protocol=tcp localport=5985 remoteip=localsubnet new remoteip=any
```
btm: all this goes away because?

mwrock: that was point #3 above but I should have provided more detail:
1. Don't need to log anything because all of stdout/err gets logged to the amazon provided logs on the instance
2. The default winrm settings on win2012r2 for things like memory and timeouts are much higher than previously in 2k8r2 and sufficint for our purposes
3. Latest TK and winrm use negotiate auth so basic auth without encryption is not necessary or desired
4. winrm is on by default on 2012r2 so there is no need to enable it

btm: Thanks

btm:
👍

PR #5144, opened 2016-07-28
Bump ruby to 2.2.5

jkeiser:


lamont-granquist:
reverting the revert of 2.2.4 needs to happen as well:

https://github.com/chef/chef/pull/4779/files

plus we need to drop 2.1.x out of the matrix and bump the required_ruby_version

lamont-granquist:
the ffi github branch can be replaced by pinning to `>= 1.9.11` which is where my patch to fix the solaris builds got released.

lamont-granquist:
👍

danielsdeleo:
I think we need to get https://github.com/chef/chef-rfc/pull/211 accepted before we ship anything. If this is urgent, let's try to expedite that?

mwrock:
:+1: with aggreement to @danielsdeleo

danielsdeleo:
I'm 👍 on the actual patch once we vote to let ourselves merge it

lamont-granquist:
looks like history is not rewritten by the victors, it is rewritten by the force push...

thommay:
once the necessary bits of #4779 are pulled back in 👍

jkeiser:
I think we figured out 2.3; leaving this open while figuring that out.

PR #5145, opened 2016-07-28
Clean up subscribes internals and notification storage.

coderanger:
John K and I are 99% sure those branches have never been used outside of
unit tests. Fingers crossed!

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -260,6 +260,18 @@ class Chef
     end
 
     #
+    # Token class to hold an unresolved subscribes call with an associated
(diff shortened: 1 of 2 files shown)


lamont-granquist:
yeah i believe i've stumbled across this before 👍

coderanger:
@lamont-granquist Changed enough to warrant re-review if you want :) Let's see if the tests pass now.

lamont-granquist:
ah i think i see, extends a string with #resource_key basically so it quacks more like a resource.

i think that's probably fine, and we can extend as necessary.

PR #5146, opened 2016-07-29
Also clear notifications when deleting a resource.

coderanger:
/cc @lamont-granquist @jkeiser @thommay 

This implements https://github.com/thommay/chef-rewind/commit/b69c9a510f1e7a85132dff0ee26fd883192ea12d for the new fancy API.

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -71,7 +71,15 @@ class Chef
       #   delete_resource!(:template, '/x/y.txy')
       #
       def delete_resource!(type, name, run_context: self.run_context)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
dangling references are clearly bad...

👍

coderanger:
I'm trying to think through what kind of weirdness could exist with nested contexts any notifications in a different run_context. Also this could still leave dangling `subscribes` I think?

thommay:
Sweet, thanks!

lamont-granquist:
The way i wrote the nested contexts notification stuff, at least for delayed notifies it should only queue up on the run_context that contains the resource.

PR #1792, opened 2014-08-11
use net/http/persistent for keepalives

lamont-granquist:
- converts all Chef connections to use keepalives

anything that uses Chef::REST or Chef::HTTP will benefit from this now.

:warning: :construction: Still WIP :warning: :construction:

PR #5151, opened 2016-07-29
HTTP 1.1 keepalives re-re-redux

### LINKED ISSUE #1617: support persistent HTTP connections in chef-client
This came from ChefCo's internal feature request board. I think it's a useful feature to have, and is smaller than a roadmap item, so we should do it.

Right now Chef Client frequently closes and opens HTTP connections to the Chef server. This causes performance problems for some customers, as they have some high-latency links and this overhead of establishing and tearing down HTTP connections slows down their client runs considerably.

lamont-granquist:
- leverages ruby 2.x so no need for net/http/persistent
- does not try to do everything and just solves the problem for clients that correctly reuse their Net::HTTP objects
- patches the cookbook synchronizer to reuse Net::HTTP objects which is probably 90% of the battle

closes #1617 
replaces #1792

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,4 @@
 AllCops:
-  TargetRubyVersion: 2.2
   Exclude:
     - "spec/data/**/*"
     - "vendor/**/*"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 AllCops:
+  TargetRubyVersion: 2.2
   Exclude:
     - "spec/data/**/*"
     - "vendor/**/*"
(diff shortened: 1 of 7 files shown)


lamont-granquist:
@chef/client-core its green, lets merge this (and its been 2 years and 1 day now since i opened the original keepalives PR in #1789)

danielsdeleo:
👍 I could see bringing the threading stuff into (and fixing the thread-safety of) `Chef::HTTP` but this fixes the one place where we really need it (except for uploads maybe).

lamont-granquist:
yep, scaled this all the way down to just fixing the one issue we get complaints about...

PR #5152, opened 2016-07-31
Fix specify members of group on suse/openbsd/solaris2/hpux

### LINKED ISSUE #3780: Chef 12.4.0 and 12.4.1 - Chef::Exceptions::UnsupportedAction: Setting members directly is not supported by
```
Chef::Exceptions::UnsupportedAction: group[root] (/var/chef/cache/cookbooks/users/providers/manage.rb line 160) had an error: Chef::Exceptions::UnsupportedAction: Setting members directly is not supported by #<Chef::Provider::Group::Usermod:0x1989ca8>
```
this occurs in conjunction with users cookbook - only this is done:

```
users_manage 'root' do
    action [:remove, :create]
end
```

going back to chef 12.4.0-1 fixes this.

tas50:
Action is an array.  We need to check to see if create is in the action
array instead of checking to see if it's ";create" which it never will
be.

Fixed #3780

### DIFF
--- a/lib/chef/provider/group/usermod.rb
+++ b/lib/chef/provider/group/usermod.rb
@@ -57,7 +57,7 @@ class Chef
           # This provider only supports adding members with
           # append. Only if the action is create we will go
           # ahead and add members.
-          if @new_resource.action == :create
+          if @new_resource.action.include?(:create)
             members.each do |member|
               add_member(member)
             end


thommay:
:+1: LGTM

tas50:
@lamont-granquist this work for you?

tas50:
@chef/client-core any thoughts?

lamont-granquist:
👍

PR #5154, opened 2016-08-01
Invalidate the file system cache on deletion

thommay:
@chef/client-core

### DIFF
--- a/lib/chef/chef_fs/file_system/repository/directory.rb
+++ b/lib/chef/chef_fs/file_system/repository/directory.rb
@@ -84,6 +84,7 @@ class Chef
             if child.exists?
               raise Chef::ChefFS::FileSystem::AlreadyExistsError.new(:create_child, child)
             end
+            FileSystemCache.instance.delete!(child.file_path)
             if file_contents
               child.write(file_contents)
             else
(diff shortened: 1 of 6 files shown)


lib/chef/chef_fs/file_system/repository/directory.rb
```diff
@@ -149,6 +150,10 @@ def exists?
 
           protected
 
+          def write(data)
+            raise "called write on #{file_path}"
```
danielsdeleo: Is there a more specific exception class we could use?

danielsdeleo:
Is this fixing a real-world bug? If so, is there any reasonable way to get the repro case in as a test?

lib/chef/chef_fs/file_system/repository/directory.rb
thommay: oof, that was debug spew that shouldn't have stayed in. But now you mention it, I think it's reasonable to raise a `Chef::ChefFS::FileSystem::FileSystemError` if someone does try to write to a directory entry.

thommay:
Running pedant against chef-zero without `--skip-policies` found this one.

danielsdeleo:
Cool, is it possible to remove `--skip-policies` from the pedant config or is there more work to do there?

thommay:
Yeah, that PR is coming shortly :)

danielsdeleo:
👍

thommay:
https://github.com/chef/chef-zero/pull/231 is the chef zero PR

PR #5155, opened 2016-08-01
Root ACLs are a top level json file not a sub-directory

thommay:
Ensure that we correctly write to them and manage them.

cc @chef/client-core


danielsdeleo:
👍

PR #5157, opened 2016-08-02
Update Ruby to 2.3

jkeiser:
This also bumps all dependencies (and thus incorporates the changes from #5129).


lamont-granquist:
webrick is super unhappy with you.

getting ruby to build on 2.3 is only the start of the 2.3 yakshave and since this is still exploding so badly on 2.3 in travis, i'm certain you haven't see what it does on a chef-client-test run in manhattan.ci

jkeiser:
@lamont-granquist fyi, all tests are fixed (confirming with latest commit).

lib/chef/node/attribute_collections.rb
```diff
@@ -154,6 +154,8 @@ def []=(key, value)
 
       alias :attribute? :has_key?
 
+      RUBY_EQUAL_OPERATORS = %w{=== == != >= <= }
```
lamont-granquist: cruft?

lamont-granquist:
just a bit of cruft to delete 👍

ksubrama:
:+1: modulo possibly retargeting things to omnibus-software master and your special branch in omnibus with my critical changes.

lib/chef/node/attribute.rb
```diff
@@ -523,12 +523,16 @@ def method_missing(symbol, *args)
           else
             raise NoMethodError, "Undefined method or attribute `#{symbol}' on `node'"
           end
-        elsif symbol.to_s =~ /=$/
-          Chef.log_deprecation %q{method setting of node attributes (node.foo="bar") is deprecated and will be removed in Chef 13, please use bracket syntax (node["foo"]="bar")}
-          key_to_set = symbol.to_s[/^(.+)=$/, 1]
-          self[key_to_set] = (args.length == 1 ? args[0] : args)
+        elsif symbol.to_s =~ /^\w/
```
lamont-granquist: I kind of think we shouldn't try to fix this anymore.  All we're doing is possibly breaking some people that are using dot assignment notation here with something that doesn't match `\w` and we can't give people the magic behavior they desired when they're using `>=` on values that need to be autovivified.  The addition of `Hash#>=` just exposes bad code that never really worked correctly before -- and all this patch does is change that from an error due to hitting `>=` to an NoMethodError so those people get raises either way.

Issue #5150, opened 2016-07-29
osx_profile cannot handle identifiers with spaces


erikng:
@natewalck 
## Description

If you accidentally (or intentionally) create an osx_profile with an identifier containing a space, `action :remove` will fail
## Chef Version

12.12.15
## Platform Version

OS X 10.11.6
## Replication Case

Create a profile:

```
profile_hash = {
  'PayloadIdentifier' => 'com.test.profile.chef.global preferences',
  'PayloadRemovalDisallowed' => false,
  'PayloadScope' => 'System',
  'PayloadType' => 'Configuration',
  'PayloadUUID' => '1781fbec-3325-565f-9022-8aa28135c3cc',
  'PayloadOrganization' => 'Chef',
  'PayloadVersion' => 1,
  'PayloadDisplayName' => 'Test Settings',
  'PayloadContent'=> [
    {
      'PayloadType' => 'com.apple.ManagedClient.preferences',
      'PayloadVersion' => 1,
      'PayloadIdentifier' => 'com.company.test',
      'PayloadUUID' => '73fc30e0-1e57-0131-c32d-000c2944c108',
      'PayloadEnabled' => true,
      'PayloadDisplayName' => 'test,
      'PayloadContent' => {
        'com.apple.test' => {
          'Forced' => [
            {
              'mcx_preference_settings' => {
                'Test' => 1,
              }
            }
          ]
        }
      }
    }
  ]
}

osx_profile 'test' do
  profile profile_hash
end
```

attempt to remove it with a cookbook:

```
osx_profile 'com.test.profile.chef.global preferences' do
  action :remove
end
```
## Client Output

```
Starting Chef Client, version 12.12.15
resolving cookbooks for run list: ["profile_tests"]
Synchronizing Cookbooks:
  - profile_tests (0.0.0)
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 1 resources
Recipe: profile_tests::default
  * osx_profile[com.test.profile.chef.global preferences] action remove
    * when removing using the identifier attribute, it must match the profile identifier
    ================================================================================
    Error executing action `remove` on resource 'osx_profile[com.test.profile.chef.global preferences]'
    ================================================================================

    RuntimeError
    ------------
    when removing using the identifier attribute, it must match the profile identifier

    Resource Declaration:
    ---------------------
    # In /Users/test/.chef/local-mode-cache/cache/cookbooks/profile_tests/recipes/default.rb

[...]

PR #5159, opened 2016-08-02
Added fix for spaces in profile identifiers

natewalck:
Adds support for profiles with spaces in the identifiers. Not great to use spaces, but Apple supports it, so we must as well.

### DIFF
--- a/lib/chef/provider/osx_profile.rb
+++ b/lib/chef/provider/osx_profile.rb
@@ -69,15 +69,16 @@ class Chef
             a.assertion do
               !@new_profile_identifier.nil? &&
                 !@new_profile_identifier.end_with?(".mobileconfig") &&
-                /^\w+(?:\.\w+)+$/.match(@new_profile_identifier)
             end
+                /^\w+(?:(\.| )\w+)+$/.match(@new_profile_identifier)
+            }
             a.failure_message RuntimeError, "when removing using the identifier attribute, it must match the profile identifier"
           else
             new_profile_name = @new_resource.profile_name
             a.assertion do
               !new_profile_name.end_with?(".mobileconfig") &&
-                /^\w+(?:\.\w+)+$/.match(new_profile_name)
-            end
+                /^\w+(?:(\.| )\w+)+$/.match(new_profile_name)
+            }
             a.failure_message RuntimeError, "When removing by resource name, it must match the profile identifier "
           end
         end
--- a/lib/chef/provider/osx_profile.rb
+++ b/lib/chef/provider/osx_profile.rb
@@ -69,16 +69,15 @@ class Chef
             a.assertion do
               !@new_profile_identifier.nil? &&
                 !@new_profile_identifier.end_with?(".mobileconfig") &&
-            end
                 /^\w+(?:(\.| )\w+)+$/.match(@new_profile_identifier)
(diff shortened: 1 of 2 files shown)


natewalck:
This fixes #5150

chef-supermarket:

* @natewalck

natewalck:
Added a spec, but it doesn't seem to work as I expect it to (and it seems to work in OTHER tests, like freebsd_service_spec.rb).

jaymzh:
CC @chef/client-os-x

lamont-granquist:
can we just remove the regexp and allow the shell command to explode?   generally this kind of validation of things to pass to commands is a whack-a-mole.

the error message also seems to be completely misleading, and doesn't give any indication of how to fix the problem.  As an example of how this can all go horribly wrong, the user_valid_regex that we have -- in the name of producing more user-understandable error messages -- has resulted in this abomination:

```
Chef::Exceptions::ValidationFailed: Option owner's value foo:bar does not match regular expression [/^[^-+~:,\t\r\n\f\0]+[^:,\t\r\n\f\0]*$/]
```

Which is 100% correct, and 100% worse UX than just letting the system call to set the user name fail...

natewalck:
@lamont-granquist This name isn't being directly handed to the shellout command. We need to make sure it meets certain standards, otherwise we can get into a situation where the profile can be installed but never uninstalled and other nastiness (thanks Apple).

Also, I'm not sure where tht error came from...it *should* be throwing:

```
"...when removing using the identifier attribute, it must match the profile identifier"
```

lamont-granquist:
okay then also fixing the error text while we're in here would be nice (plus travis busted, plus now a merge a conflict)

but :+1: on adding space to the regex

natewalck:
@lamont-granquist Regarding the rspec I started on, what would your guidance be for testing this particular behavior? I tried modeling how freebsd service does it, but it does not work.

I am doing: 

```
    it "should work with spaces in the identifier" do
      new_resource.identifier "com.testprofile.screensaver settings"
      new_resource.action(:remove)
      provider.load_current_resource
      provider.define_resource_requirements
      expect { provider.process_resource_requirements }.not_to raise_error
    end
```

And if I tweak the regex to now allow spaces again, it still does not raise an error. I'm not sure what I am missing here.

lamont-granquist:
yeah i don't see anything wrong with that test, it looks completely correct to me...

all i can think of is that there's some other state (from some `let` block?) way above you that is bleeding into your test and changing things?

natewalck:
I was setting the action in the wrong place it seems. I'll confirm all tests as passing now.

I did have one struggle...when I tried to override `new_resource.name` thusly:

```
new_resource.name "com.testprofile.screensaver settings"
```

and then do an interactive pry sessions in the provider itself, it never got the new resource name. `@new_resource.profile_name` was always set to what was used on line 184 of this test:

```
let(:new_resource) { Chef::Resource::OsxProfile.new("Profile Test", run_context) }
```

I'm not sure why I couldn't change it, but it seems to be working as expected now. I suspect these unit tests may need some further love before too long.

natewalck:
Needs final approval (and maybe tweaks to the rspec?) @jaymzh

jaymzh:
:+1:

PR #5165, opened 2016-08-03
Make Chef work with Ruby 2.3, update Ruby to 2.1.9

jkeiser:
The plan is to check this in now, and update to Ruby 2.3 promptly after the next minor.


lib/chef/file_access_control/windows.rb
```diff
@@ -128,7 +128,7 @@ def securable_object
       end
 
       def should_update_dacl?
-        return true unless ::File.exists?(file)
+        return true unless ::File.exists?(file) || ::File.symlink?(file)
```
mwrock: ugh. `File.exist?` no longer follows symlinks?

omnibus/Gemfile
```diff
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
-gem "omnibus", git: "https://github.com/chef/omnibus.git"
-gem "omnibus-software", git: "https://github.com/chef/omnibus-software.git"
+gem "omnibus", github: "chef/omnibus", branch: "ksubrama/gcc_investigate"
```
mwrock: Are these intentional?

.travis.yml
```diff
@@ -52,13 +52,13 @@ matrix:
     bundler_args: --without changelog development docgen guard integration maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
   - env:
       CHEFSTYLE: 1
-    rvm: 2.1
+    rvm: 2.2.5
```
mwrock: Curious, if we plan to move to 2.3.1, why not set these to that instead of 2.2.5?

lib/chef/file_access_control/windows.rb
jkeiser: It *does* now, and we still want to be able to update the dacl of a symlink that goes nowhere (because it could have a file dropped in later).

omnibus/Gemfile
jkeiser: Yes. These will be on master until we've solved the msys2 problem entirely. They have been reviewed and worked at, but there is a desire not to check them in since we think we may be able to make things work better with an omnibus-toolchain for windows.

.travis.yml
jkeiser: Because they were all originally set to 2.2 (I don't know where the 2.1 is coming from), so I'm just adding a patch version.

mwrock: I just wonder if we'd be better off with them at 2.3.1 since thats what we plan to roll with.

mwrock:
:+1:

ksubrama:
:+1:

PR #5166, opened 2016-08-04
Add `create_user_group` attribute to Useradd provider

mfischer-zd:
Add a `create_user_group` attribute to Chef::Provider::User::Useradd.
The default value is `nil`, so there's no change in existing behavior.
But setting it to `true` or `false` will append the `-U` or `-N`
optional arguments, respectively, to the `useradd` commandline.

This may or may not be compatible with all useradd(1) variants (it
works on Ubuntu 12+ and CentOS 6+, at least), so additional OS/distro
compatibility testing may be desirable.

### DIFF
--- a/lib/chef/provider/user/useradd.rb
+++ b/lib/chef/provider/user/useradd.rb
@@ -141,6 +141,14 @@ class Chef
         def useradd_options
           opts = []
           opts << "-r" if new_resource.system
+          # The default behavior is to follow whatever is in the
+          # USERGROUPS_ENAB variable in /etc/login.defs.
+          case new_resource.create_user_group
+          when true
+            opts << "-U"
+          when false
+            opts << "-N"
+          end
           opts
         end
 
--- a/lib/chef/resource/user.rb
+++ b/lib/chef/resource/user.rb
@@ -41,6 +41,7 @@ class Chef
         @manage_home = false
         @force = false
         @non_unique = false
+        @create_user_group = nil
         @supports = {
           :manage_home => false,
           :non_unique => false,
@@ -155,6 +156,13 @@ class Chef
         )
       end
 
+      def create_user_group(arg = nil)
+        set_or_return(
+          :create_user_group,
+          arg,
+          :kind_of => [ TrueClass, FalseClass, NilClass ]
+        )
+      end
     end
   end
 end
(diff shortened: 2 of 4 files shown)


lamont-granquist:
kind of worried that because of the existing awfulness with useradd as the base class that this will troll solaris and aix.  might work better to build this on top of current master where the linux_user resource + provider are broken out for exactly this kind of purpose -- but the while the class has been broken out the hard work of refactoring the code is not done.

lamont-granquist:
i also wonder if we just shouldn't have a `command_options` property so that users can inject arbitrary flags into the useradd/userdel command and be done with it all for good...

mfischer-zd:
I'm fine with the `command_options` or `extra_args` attribute idea, but I would like not to kick the can down the road for too long.  

> this will troll solaris and aix

I'm not sure what you mean by this.

lamont-granquist:
you're editing the base class which is inherited by aix and solaris, which means you're injecting code into those, which may blow up -- possibly in our build system that you (unfortunately) don't have any visibility into that runs all the tests on solaris and aix boxes.  adding arbitrary options to the commands would be more general and something i'd be more likely to support going into the base class.

mfischer-zd:
Who gets to make the UX/behavioral decision here?

lamont-granquist:
With the chef 12.14.x changes to the user provider the useradd provider is now the base class only for solaris and aix (with a future TODO to eliminate the base class entirely).  The `lib/chef/provider/user/linux.rb` file is now where this code lives and it should be much easier to patch it now (and easier to read and reason about) and not have to worry about clobbering inherited code on other operating systems (which aren't tested enough in travis).

thommay:
Hi,
the route we've been discussing for other providers that can take a bunch of arguments (`mdadm` is the stand out one) is to allow them to take an options hash. I agree with Lamont that that's the better option here, too - and that's what I'd like to see for us to accept this PR.
Thanks for the submission!

mfischer-zd:
> the route we've been discussing for other providers that can take a bunch of arguments (mdadm is the stand out one) is to allow them to take an options hash. I agree with Lamont that that's the better option here, too - and that's what I'd like to see for us to accept this PR.

I'm not super opposed to this, since such an attribute can help operate as a last option to cover missing provider functionality.  But even if such an option existed, I might have proffered a PR for this anyway: one of the main benefits of Chef is to provide an abstraction layer between the user and the underlying implementation.  

So I think it totally makes sense to support a new attribute anyway, since the question of whether to use a user-private-group (UPG) scheme is a site or machine policy that Chef can help enforce.  How it does it under the covers is an implementation decision that may vary across OSes, and the translation between policy and implementation is just the sort of thing we use Chef for.

mfischer-zd:
@charlesjohnson do you have an opinion here?

thommay:
As discussed, we're not happy with adding more options to useradd, and so are closing this PR. Thanks for your contribution!

PR #5170, opened 2016-08-05
Bump dependencies to bring in Ohai 8.18

btm:
Replaces #5168

Manually added back ffi line in kitchen-tests/Gemfile.lock that was failing travis.


btm:
#5168 had a couple :+1:. If this goes green I'll merge it.

PR #5173, opened 2016-08-05
noop apt_update similar to apt_repository

lamont-granquist:
makes apt_update behave similarly to apt_repository when invoked on non-apt-using distros

### DIFF
--- a/lib/chef/provider/apt_update.rb
+++ b/lib/chef/provider/apt_update.rb
@@ -16,15 +16,17 @@
 # limitations under the License.
 #
 
-require "chef/resource"
-require "chef/dsl/declare_resource"
+require "chef/provider"
+require "chef/provider/noop"
 
 class Chef
   class Provider
     class AptUpdate < Chef::Provider
       use_inline_resources
 
-      provides :apt_update, os: "linux"
+      provides :apt_update do
+        uses_apt?
+      end
 
       APT_CONF_DIR = "/etc/apt/apt.conf.d"
       STAMP_DIR = "/var/lib/apt/periodic"
@@ -75,6 +77,13 @@ class Chef
         declare_resource(:execute, "apt-get -q update")
       end
 
+      def self.uses_apt?
+        ENV["PATH"] ||= ""
+        paths = %w{ /bin /usr/bin /sbin /usr/sbin } + ENV["PATH"].split(::File::PATH_SEPARATOR)
+        paths.any? { |path| ::File.executable?(::File.join(path, "apt-get")) }
+      end
     end
   end
 end
+
+Chef::Provider::Noop.provides :apt_update
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -254,4 +254,4 @@ class Chef
   end
 end
 
-Chef::Provider::Noop.provides :apt_resource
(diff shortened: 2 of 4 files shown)


lib/chef/provider/apt_update.rb
```diff
@@ -75,6 +77,13 @@ def do_update
         declare_resource(:execute, "apt-get -q update")
       end
 
+      def self.uses_apt?
+        ENV["PATH"] ||= ""
+        paths = %w{ /bin /usr/bin /sbin /usr/sbin } + ENV["PATH"].split(::File::PATH_SEPARATOR)
+        paths.any? { |path| ::File.executable?(::File.join(path, "apt-get")) }
```
coderanger: We have a `which` helper for this :)

coderanger:
I still think it's a bit weird to use the provider resolution system for this vs. `return unless uses_apt?` in the action method.

lib/chef/provider/apt_update.rb
lamont-granquist: yeah, i'm copying what the apt_repository provider does, i'm not entirely sure why @thommay wasn't using that, but the code is slightly different here.

coderanger: We should probably fix both to use the `which` helper?

lamont-granquist: there's another task here to clean up the code duplication in uses_apt? between the two providers, and use the which helper and/or fix the which helper, along with ideally adding uses_apt? to some kind of DSL helper so that we can use it everywhere, but that's getting well outside of the scope of what i wanted to do here and raises namespace concerns and i'm just not feeling creative enough today to handling naming things...

lamont-granquist: i seem to recall suggesting the `which` helper last time and don't recall why we went with this code...

thommay: I'm pretty sure i tried to use the which helper and it didn't work, but i'm hazy on the details right now. And yes, eventually we should extract the `uses_apt?` method.

thommay:
You have to use the provider resolution system don't you - on debian family you want to run the provider, everywhere else you want to no-op, not to just bail out of the provider.

lamont-granquist:
yeah but if you return'd you'd no op the action and the resource wouldn't update, which is about the same effect...

i kind of like that the you don't have to guard every single action with the checking code, but i can also see how it might just be some OOP fetishism going on my head...   right now this approach feels vaguely 'cleaner' to me, but not in any particularly objective sense...

coderanger:
It feels mildly less discoverable to do it this way, but only mildly. `¯\_(ツ)_/¯`

lamont-granquist:
I think I'm just going to mergify this one.

Fixes the bug in apt_repository not actually using the no-op provider at all, gets the similar behavior right for apt_update, and adds tests so we know that both behaviors are correct and lets @tas50 start using them, which is good enough incremental updating.

the which/uses_apt? code de-dup we should address, but I have too much stacked up to get to it now...

coderanger:
Yeah, I'll see if I can do a debt-paydown PR tonight or tomorrow to clean it all up. This is good enough for now :) Any changes won't be user visible so the only ones it hurts are us (aka the lost souls)

lamont-granquist:
yeah, and i there's many worse sins we've committed...

PR #5174, opened 2016-08-05
Remove race conditions from run_lock_spec

jkeiser:


spec/functional/run_lock_spec.rb
```diff
@@ -427,6 +433,8 @@ def create_lock
       end
     end
 
+    attr_writer :run_lock
```
danielsdeleo: What's this for? I don't see `self.run_lock = thing` elsewhere in the diff, or see any `@run_lock` references.

danielsdeleo:
One line looks like maybe a leftover from debugging, 👍 on the rest

spec/functional/run_lock_spec.rb
jkeiser: Cruft! I'll remove.

lamont-granquist:
:+1:

PR #5180, opened 2016-08-08
Revert "Merge pull request #5123 from chef/lcg/resolv-replace"

### LINKED ISSUE #5141: Connections to 'localhost' fail unexpectedly because IPv6 localhost is used
## Description

For some users, the use of the `resolv-replace` may change the behavior of core ruby libraries when connecting to `localhost`.   Without `resolv-replace` many ruby network operations will resolv `localhost` to an IPv4 address even on a machine with an IPv6 loopback available.  With `resolv-replace` in place, the IPv6 address is preferred.  While this may be technically correct, it may also break 3rd party code using ruby networking primitives such as cookbook code or knife plugins.

## Chef Version

This problem exists on chef-master since e8877cd363642ed6757c48d1ed5ab35509d87e22

## Platform Version

Linux

## Replication Case

Consider the following two scripts

```ruby
#!/opt/opscode/embedded/bin/ruby
# ./no_work: This one doesn't work

require 'resolv'
require 'resolv-replace'
require 'uri'
require 'open-uri'
require 'openssl'

uri = URI.parse("https://localhost/version")
puts open(uri, {ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE}).each_line.first
```

```ruby
#!/opt/opscode/embedded/bin/ruby
# ./works: This one works

require 'uri'
require 'open-uri'
require 'openssl'

uri = URI.parse("https://localhost/version")
puts open(uri, {ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE}).each_line.first
```

On the same machine, we get two different results because of the include of `resolv-replace`:

```
[centos@ip-10-194-11-219 test-data-dir]$ ./works
chef-server 12.8.0+20160725093028
[centos@ip-10-194-11-219 test-data-dir]$ ./no_work
/opt/opscode/embedded/lib/ruby/2.1.0/resolv-replace.rb:23:in `initialize': Connection refused - connect(2) for "::1" port 443 (Errno::ECONNREFUSED)
        from /opt/opscode/embedded/lib/ruby/2.1.0/resolv-replace.rb:23:in `initialize'
        from /opt/opscode/embedded/lib/ruby/2.1.0/net/http.rb:879:in `open'
        from /opt/opscode/embedded/lib/ruby/2.1.0/net/http.rb:879:in `block in connect'
        from /opt/opscode/embedded/lib/ruby/2.1.0/timeout.rb:76:in `timeout'
        from /opt/opscode/embedded/lib/ruby/2.1.0/net/http.rb:878:in `connect'
        from /opt/opscode/embedded/lib/ruby/2.1.0/net/http.rb:863:in `do_start'
        from /opt/opscode/embedded/lib/ruby/2.1.0/net/http.rb:852:in `start'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:313:in `open_http'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:724:in `buffer_open'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:210:in `block in open_loop'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:208:in `catch'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:208:in `open_loop'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:149:in `open_uri'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:704:in `open'
        from /opt/opscode/embedded/lib/ruby/2.1.0/open-uri.rb:30:in `open'
        from ./no_work:10:in `<main>'
```

@lamont-granquist and I discussed this issue briefly.  One option we discussed is providing our own version of `resolv-replace` which prefers IPv4 addresses.  Another option would be to revert the commit that adds `require 'resolv-replace'` into chef/application.rb.  A third would be to instruct users to change `localhost` to `127.0.0.1` or `localhost4` if they want to ensure IPv4 will be used.

jkeiser:
Fixes #5141. This reverts commit e8877cd363642ed6757c48d1ed5ab35509d87e22, reversing
changes made to 162d988d3bcfce2773ef2831fa33e416f53731f6.

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -29,10 +29,6 @@ require "tmpdir"
 require "rbconfig"
 require "chef/application/exit_code"
 require "yaml"
-require "resolv"
-# on linux, we replace the glibc resolver with the ruby resolv library, which
-# supports reloading.
-require "resolv-replace" if RbConfig::CONFIG["host_os"] =~ /linux/
 
 class Chef
   class Application
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -154,7 +154,6 @@ class Chef
         do_contents_changes
         do_acl_changes
         do_selinux
-        do_resolv_conf_fixup
         load_resource_attributes_from_file(@new_resource)
       end
 
@@ -443,13 +442,6 @@ class Chef
           else
             Chef::Log.debug "selinux utilities can not be found. Skipping selinux permission fixup."
           end
-        end
-      end
-
-      def do_resolv_conf_fixup
-        # reload /etc/resolv.conf after we edit it -- only on linux -- and see lib/chef/application.rb
-        if new_resource.path == "/etc/resolv.conf" && RbConfig::CONFIG["host_os"] =~ /linux/
-          Resolv::DefaultResolver.replace_resolvers [Resolv::DNS.new("/etc/resolv.conf")]
         end
       end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -154,8 +154,6 @@ RSpec.configure do |config|
   config.filter_run_excluding :unix_only => true unless unix?
   config.filter_run_excluding :aix_only => true unless aix?
   config.filter_run_excluding :debian_family_only => true unless debian_family?
-  config.filter_run_excluding :linux_only => true unless linux?
-  config.filter_run_excluding :non_linux_only => true if linux?
   config.filter_run_excluding :supports_cloexec => true unless supports_cloexec?
   config.filter_run_excluding :selinux_only => true unless selinux_enabled?
   config.filter_run_excluding :requires_root => true unless root?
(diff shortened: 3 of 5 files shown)


coderanger:
😿 Some day we can have nice things.

lamont-granquist:
yeah, part of the reason why i just shipped the apt_update noop stuff without bothering to fix the `which` stuff is that i'd rather try to clean off my plate and resurrect this one...

lamont-granquist:
👍

lamont-granquist:
#5141 is the blocking issue here causing the revert

stevendanna:
👍 Until we have a solution for the use cases we'll break

jkeiser:
Excluding from changelog, since no stable release has this in it.

PR #5187, opened 2016-08-10
Create and delete yum repositories

thommay:
Bonus fix to make Chef::Mixin::Which work in class methods, so we can use it for provider resolution.
cc @chef/client-maintainers


kitchen-tests/cookbooks/base/recipes/default.rb
```diff
@@ -16,8 +16,13 @@
   include_recipe "selinux::disabled"
 end
 
-if node["platform_family"] == "rhel"
-  include_recipe "yum-epel"
+yum_repository "epel" do
```
lamont-granquist: 👍

lib/chef/mixin/which.rb
```diff
@@ -19,6 +19,10 @@ class Chef
   module Mixin
     module Which
       def which(cmd, opts = {})
+        Chef::Mixin::Which.which(cmd, opts)
+      end
+
+      def self.which(cmd, opts = {})
```
lamont-granquist: doesn't this not work without the `included?` hook?  and can't you `extend` vs `include` to mix in the module methods into the class instead of the instance?

thommay: i think you just triple-negative'd my brain into a loop. This definitely works; but I don't really know enough about `include`/`extend` to say definitively that it's worse.

lamont-granquist: i think i'm saying you can leave the module alone and use extend instead of include to mix the module into a class and it'll solve this problem.

i just can't quite figure out how this is even working for you right now, though, because i think the class method in the module will get ignored...    which is to say that having `def self.foo` in a module usually doesn't work at all...

lamont-granquist:
yep 👍

danielsdeleo:
Could maybe use some debug logging in `action_create` in the provider, but 👍 as is

martinisoft:
It's about time 👍

PR #5193, opened 2016-08-11
bump all dependencies

lamont-granquist:
result of `rake dependencies`

most importantly removes compat_resource as a gem dep of chef


jkeiser:
:+1:

PR #5194, opened 2016-08-11
Keeping static libraries previously deleted by jenkins builds

mwrock:
From what I can tell by looking at jenkins build logs, Windows is the only platform where actual files are deleted by this build delendency defined [here](https://github.com/chef/omnibus-software/blob/master/config/software/clean-static-libs.rb). The files deleted on Windows are:
```
Deleting C:/opscode/chef/embedded/lib/ruby/gems/2.1.0/gems/libyajl2-1.2.0/ext/libyajl2/libyajldll.a
Deleting C:/opscode/chef/embedded/lib/ruby/gems/2.1.0/gems/libyajl2-1.2.0/lib/libyajl2/vendored-libyajl2/lib/libyajldll.a
```
These files are each about 32k. Further, the `ffi-yajl` gem depends on these libraries. So if a user attempts to install a different version of lib-yajl or a gem that depends on it, they will receive compilation errors blocking the install.

### DIFF
--- a/omnibus/config/software/chef-complete.rb
+++ b/omnibus/config/software/chef-complete.rb
@@ -17,5 +17,3 @@ if windows?
   dependency "ruby-windows-devkit"
   dependency "ruby-windows-devkit-bash"
 end
-
-dependency "clean-static-libs"


mwrock:
running a jenkins build on this now.

lamont-granquist:
https://github.com/chef/omnibus-software/pull/609 is where it got added.

i can't recall what i was thinking when i was writing comments in that thread...

unless it really moves the needle a lot on the filesize of the download i'm generally opposed to little tweaks like removing *.a files that generally nobody will notice and only have downside in causing problems like this one.

better to spend effort on stuff like actively removing large chunks of software from builds (libiconf, ncurses and stuff like perl/python in the server packages).

so :+1: on this.  not sure where my head was when i was :+1:'ing the other PR...

and people have hit this bug and complained in the ffi-yajl repo:

https://github.com/chef/ffi-yajl/issues/81

PR #5196, opened 2016-08-12
Add more cops to chefstyle enforcement

lamont-granquist:
Enables 22 additional cops.

These are all cops that had less than 10 offenses in the whole chef/chef codebase, so most of these are not common, plenty of them are pretty awful though.

In a few cases I left cops disabled via comments that were difficult to fix but enabling them globally will prevent new offenses from creeping into the codebase.


lamont-granquist:
travis is failing because these cops aren't enabled, but they're being disabled by comments, but that doesn't fit the chefstyle workflow at all, so i dunno...

lib/chef/chef_fs/file_system.rb
```diff
@@ -68,7 +68,7 @@ def list_from(entry, &block)
                 list_from(exact_child, &block)
               end
 
-            # Otherwise, go through all children and find any matches
+              # Otherwise, go through all children and find any matches
```
tas50: This is odd

lamont-granquist: its consistent, it keeps everything inside the block at the same indentation level.

thommay:
I don't love RescueModifier ; doesn't seem like that change benefits readability to me.

thommay:
What's the reason for AsciiComments in this day and age?

lamont-granquist:
I kind of like AsciiComments because it prevents stuff like the unicode `-` sign that was gratuitously added (probably due to some bad copypasta from a webpage or something).  The downside is that you can't use special characters like the euro or pound symbol, but we shouldn't need those much in comments, and it impacts author names in comments, but I'm leaning towards :-1: for tracking all the authors in headers anyway -- i stopped doing that awhile ago...

lamont-granquist:
And actually, yeah, we should turn off RescueModifier, that one will annoy me...

lamont-granquist:
And as an example, trying to fix each AsciiComments warning was super annoying.  I have my vim tabline setup to give me the character under the cursor, so i had to go across every character in the line until I found one that was > 0xFF and found the unicode minus signs in the text -- it would have been more difficult without having that information right on my tabline (which isn't the default tabline in vim).  That impacts ability to grep the text, find text in editors, search and replace, etc.

lamont-granquist:
RescueModifier removed

thommay:
Yeah, moving to not putting authors in files is the right thing to do, but butchering peoples names is not. I vote that we should fix the ones that make sense to fix and then turn the rule off.

lamont-granquist:
dropped AsciiComments cop as well to get this moving

lamont-granquist:
travis will go red because of chefstyle checks failing that won't fail once https://github.com/chef/chefstyle/pull/31 is released (which we will immediately pull since we track master, so if we whack the green merge button on both of these fast enough, master of chef/chef shouldn't go red...)

thommay:
👍

PR #5197, opened 2016-08-12
Bumping ohai and mixlib-log to fix regression

mwrock:
Regresion reported in https://github.com/chef/mixlib-log/issues/17


kitchen-tests/Gemfile.lock
```diff
@@ -59,9 +59,8 @@ GEM
     excon (0.51.0)
     faraday (0.9.2)
       multipart-post (>= 1.2, < 3)
-    ffi (1.9.14)
```
btm: @mwrock I believe this is a recurring bug (when working from a windows box?) and you need to add this line back.

mwrock: i noticed that. I actually ran `rake dependencies:update` on ubuntu. Was gonna just let it be unless it breaks.

mwrock: Looking at the whole gemfile.lock I think thats the right thing. windows is the only platform that needs that gem and it kept the mingw version

btm: Will see if this happens again, this is why I added 2da9c1b20c11ce944835ea818a3c81566af7d739 last time.

https://travis-ci.org/chef/chef/jobs/149830214

btm: actually, looks like it already happened again: https://travis-ci.org/chef/chef/jobs/151909377

mwrock: yup. guess we need to look into why that is getting yanked but will add it manually now

PR #5202, opened 2016-08-15
Transitive dependency licensing information for Chef Client

sersut:
This PR goes with https://github.com/chef/omnibus/pull/705 and https://github.com/chef/omnibus-software/pull/714 and adds licensing information collection for the transitive dependencies of Chef Client. 

An ad_hoc build with all these branches are tested at:
http://manhattan.ci.chef.co/job/chef-trigger-ad_hoc/157/downstreambuildview/

Note that the PRs linked from here do not contain the msys2 changes that are currently sitting in the private branches of `omnibus` and `omnibus-software` pointed to by this project. Once they are merged we will need to rebase the private branches on top of master and re-test.

### DIFF
--- a/omnibus/Gemfile
+++ b/omnibus/Gemfile
@@ -1,7 +1,8 @@
 source "https://rubygems.org"
 
-gem "omnibus", github: "chef/omnibus", branch: "ksubrama/gcc_investigate"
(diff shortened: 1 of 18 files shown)


mwrock:
:+1:  on green

PR #5205, opened 2016-08-15
Wait until inner webrick http server is confirmed running

mwrock:
Sometimes when a jenkins slave aborts a build during functional tests when we are using the rack based `TinyServer`, future builds die on that slave during http server based tests. The aborted run has not completed its shutdown and therefore is still listening for HTTP requests on the expected port. So in this state the subsequent run thinks its `TinyServer` is ready but tests fail due to 500 errors returned from the server (probably because it is shutting down) and then the subsequent run crashes when it tries to shutdown its own unstarted server.

This PR grabs the `Webrick::HTTPServlet` yielded by Rack's `run` call which would fail if run fails and checks for that to see if its server has started rather than the parent `Rack::Server` which is likely to be present even in a failed `run` call.

### DIFF
--- a/spec/tiny_server.rb
+++ b/spec/tiny_server.rb
@@ -20,7 +20,7 @@ require "rubygems"
 require "webrick"
 require "webrick/https"
 require "rack"
-#require 'thin'
+require "thread"
 require "singleton"
 require "open-uri"
 require "chef/config"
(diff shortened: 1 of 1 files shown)


mwrock:
This also increases the timeout for waiting for the server to start from 20 seconds to a minute.

spec/tiny_server.rb
```diff
@@ -78,12 +80,13 @@ def url
     end
 
     def block_until_started
-      200.times do
-        if started? && !@server.nil?
+      600.times do
+        if !@http_server.nil? && started?
```
thommay: isn't `if @http_server && started` the same and somewhat more readable?

thommay:
1 comment but 👍 either way

spec/tiny_server.rb
mwrock: I switched it because no need to make HTTP calls in `started?` if there is no server. Not a critical reordering.

spec/tiny_server.rb
```diff
@@ -78,12 +80,13 @@ def url
     end
 
     def block_until_started
```
jkeiser: I only recently realized this, but `Queue` is a better way to do this:

```ruby
require "thread"
started = Queue.new
t = Thread.new { sleep 1; started << true }
# thread-safely waits until started << true happens
started.pop
```

mwrock: unfortunately @jkeiser this is not an option in this case (but a nice trick to stash in my memory banks). `block_until_started` is not intended to block until the thread completes but to block until the webrick server is started. So we "want" the thread to run until the tests finish.

PR #5206, opened 2016-08-15
Clear test directory before file resource examples

mwrock:
This is another fix for ensuring aborted functional tests do not corrupt subsequent runs. Here we make sure that the test directory is deleted in case it was not cleared out by the aborted run.


thommay:
the `if` seems somewhat unnecessary in the face of an `rm_rf`...

thommay:
but 👍 either way

mwrock:
Oh I was thinking `rm_rf` would fail if it did not exist but looks like thats not the case. Thanks!

PR #5209, opened 2016-08-16
Support indented heredoc as systemd_unit file content

mal:
A little sugar for defining unit files using a heredoc until Chef moves to `2.3` and the `<<~` syntax becomes available.

It works by scaning the lines of the heredoc and gathering the leading spaces before finding the shortest (excluding completely blank lines) and removing the common whitespace from the start of all lines. Given the following resource:

```rb
systemd_unit 'foo' do
  content <<-EOF
    [Unit]
    Description=bar
  EOF
  action :create
end
```

Before:
```ini
    [Unit]
    Description=bar
```

After:
```ini
[Unit]
Description=bar
```

### DIFF
--- a/lib/chef/resource/systemd_unit.rb
+++ b/lib/chef/resource/systemd_unit.rb
@@ -56,6 +56,8 @@ class Chef
           end.to_s
         else
           content.to_s
+            .gsub(/^#{content.to_s.scan(/^[^\S\n]+/).min_by { |l| l.length }}/, "")
+            .strip + "\n"
         end
       end
     end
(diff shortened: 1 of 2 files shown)


coderanger:
:-1: for supporting this in chef. It's a major departure from everything else.

coderanger:
This also might not always be equivalent with certain types of line continuation syntax? Overall this is best done at the level of user code, not Chef.

```ruby
systemd_unit 'foo' do
  content <<-EOF.gsub(/^\s+/, '')
    [Unit]
    Description=bar
  EOF
  action :create
end
```

That usually does the trick :)

mal:
Yup, that's typically what I've been doing, but even with it all over the place, I'm not the only one writing Chef and it's often forgotten when new units are created. :disappointed: 

I take your point about it not really being a Chef level thing though, I'll just grit my teeth and bear it until Chef on `2.3`!

PR #5211, opened 2016-08-16
Changes that get Jenkins running against Ruby 2.3

jkeiser:
I want to commit these separately in an attempt to minimize the differences in the Ruby 2.3 branch.


mwrock:
:+1:

PR #5216, opened 2016-08-17
Fix cookbook upload of symlinked cookbooks in Ruby 2.3 on Windows

jkeiser:
In Ruby 2.3 on Windows, `Find.find("x/")` will not traverse `x` if it is a symlink (while it does on other platforms). This fixes that to be consistent across platforms.


mwrock:
:+1: on green

lamont-granquist:
:+1:

jkeiser:
Merging because this doesn't make the problem in `master` worse and it does fix Ruby 2.3.

PR #5218, opened 2016-08-17
fix Style/BlockDelimiters, Style/MultilineBlockLayout and 0.42.0 engine upgrade

lamont-granquist:


lamont-granquist:
@chef/client-core this is all a requirement to upgrade to the rubocop 0.42.0 engine -- without these rules enabled it gets into an infinite loop on one of the dsc files using let blocks with parens around heredocs.

lib/chef/http.rb
```diff
@@ -248,8 +248,6 @@ def http_client(base_url = nil)
       end
     end
 
-    private
```
danielsdeleo: Is this because rubocop dislikes the `public` we had down below? Probably better to move the one method that affected and leave the rest `private` ?

lib/chef/knife.rb
```diff
@@ -233,8 +233,6 @@ def self.load_deps
       end
     end
 
-    private
```
danielsdeleo: I think this `private` didn't have the effect the author intended; I recall reading that you have to do `class << self` along with private to get private class methods. In any case, probably an `# @api private` would at least be good there?

danielsdeleo:
👍 aside from the public/private stuff

danielsdeleo:
Awesome, thanks for tidying up

lamont-granquist:
yeah i'm not sure how that got in there, i think the engine upgrade must have found a few more cases to complain about because that should have been in the previous PR...

PR #5220, opened 2016-08-17
remove ruby 2.1 support

lamont-granquist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,6 @@ env:
 
 matrix:
   include:
-  - rvm: 2.1.9
-    sudo: true
-    script: sudo -E $(which bundle) exec rake spec;
-    # also remove integration / external tests
-    bundler_args: --without changelog development docgen guard integration maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
   - rvm: 2.2.5
     sudo: true
     script: sudo -E $(which bundle) exec rake spec;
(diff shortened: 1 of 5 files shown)


coderanger:
Hrm, so this is going to make Travis testing a lot harder for me :-/ Not sure if that's a reason to block it but test matrix management is going to be even worse than it is now.

lamont-granquist:
the whack-a-mole of gems dropping support for 2.1 already is only going to get worse, and for berkshelf it is already pretty intractable to get ruby 2.3/2.4 support while supporting 2.1

and ruby 2.2 has been passing clean on travis for ages and is a pretty suitable ruby to test 12.x on if you need a single ruby to test on (which will inherently get some tests incorrect).   i'm reasonably certain we had clean tests on 2.2 long before we ever managed to get windows and omnibus-chef onto 2.1

for a pattern to use to test old chef versions against the version of chef they were released against you can copy this approach:

https://github.com/chef-cookbooks/compat_resource/blob/master/.travis.yml

now that i look at that i think we can inject `ruby 2.1.9` into the old rubies and modern bundler will actually be able to figure it out so i think that the old crap there can all be simplified a bit.

coderanger:
Yeah, that's what I think I need to move to, just need to write some helpers to generate the travis matrix for each project because I have so many. Probably not an issue for anyone else though so :+1:

lamont-granquist:
looks like we need to wait until ruby-2.3 lands in -current to make the kitchen tests happy

lamont-granquist:
right so the way that i've used kitchen-appbundle-updater over kitchen-dokken means we're using the docker image with the prebaked chef-client rather than installing current from omnitruck.

PR #5222, opened 2016-08-18
speed up solo test

lamont-granquist:
removes the requirement that the second wait for the first, just that
one of them waits for the other, so if we ever somehow lose that race
and the second wins the run_lock as long as the first sleeps then its
all good...

this now removes all the sleeps and just uses the logfile for IPC so that the first
run only exits after the second run writes its logline that its waiting on the first.

on my mac this results in exactly zero calls to sleep now.

upper bound is still that it'll wait for 20 seconds for insanely slow hosts, but it now exits as soon as it can rather than taking 25 minutes no matter how slow the host is.


thommay:
:+1: LGTM

PR #5224, opened 2016-08-19
Fix TinyServer races by waiting for stop() and start() callbacks

jkeiser:
1. Wait for the start callback rather than the do block, as it happens
   later. This prevents us from getting into many conditions where the
   start returns before the server is fully initialized, where if stop()
   is called too soon after start returns, it would cause exceptions.
2. Wait for the thread to finish completely
3. Use Thread.current.abort_on_exception` to show the actual listen errors,
   so that problems like this can be diagnosed.
4. Use :each to start and stop TinyServer so that race conditions like this
   are exposed earlier and more often (this will let us rid ourselves of them
   more quickly).
5. Use WEBrick::HTTPServer directly, which gets rid of the need for the
   complicated trap workaround (rack is the one that traps).

### DIFF
--- a/spec/functional/http/simple_spec.rb
+++ b/spec/functional/http/simple_spec.rb
@@ -26,11 +26,11 @@ describe Chef::HTTP::Simple do
   let(:http_client) { described_class.new(source) }
   let(:http_client_disable_gzip) { described_class.new(source, { :disable_gzip => true } ) }
 
-  before(:all) do
+  before(:each) do
     start_tiny_server
   end
 
-  after(:all) do
+  after(:each) do
     stop_tiny_server
   end
 
--- a/spec/functional/knife/cookbook_delete_spec.rb
+++ b/spec/functional/knife/cookbook_delete_spec.rb
@@ -20,9 +20,13 @@ require "spec_helper"
 require "tiny_server"
 
 describe Chef::Knife::CookbookDelete do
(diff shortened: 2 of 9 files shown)


mwrock:
Looks like CI still isn't happy. I wonder if we want to have `start` retry n times or until n seonds have elapsed if it fails. If a another process did not propperly exit, and the server did not shut down clenly, it might (just a guess) not start successfully but perhaps it eventualy will. Another thought is to randomize the port so that in theory two servers could peacefully coexist.

jkeiser:
start does have a 5 second timeout; but the main point of the patch is to shut down the server cleanly! I'll investigate and see what's up.

jkeiser:
@mwrock I found a spot in the tests where I was starting the server and had no corresponding stop. Fixed, and tests are green.

mwrock:
travis and appveyor are still being difficult

jkeiser:
Indeed. Windows, and Ruby 2.1, are failing. Will look Monday.

jkeiser:
This fix depends on #5220 , as it (temporarily) increases the Ruby 2.1 test time beyond the timeout.

lamont-granquist:
👍

mwrock:
:+1:  on green

jkeiser:
All succeeded except a known intermittent error in kitchen tests.

PR #5225, opened 2016-08-19
Add monkey patch to fix bundle locks for platforms

mwrock:
This is a temporary fix until bundler 1.14 releases see https://github.com/bundler/bundler/issues/4896


jkeiser:
:+1: with sadness!

PR #5230, opened 2016-08-22
Fix "URI.escape is obsolete" warnings

jkeiser:
We'll use `Addressable::URI` instead. See http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape#answer-13059657 for details on why.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -28,6 +28,7 @@ PATH
   remote: .
   specs:
     chef (12.14.20)
+      addressable
       bundler (>= 1.10)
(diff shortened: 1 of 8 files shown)


stevendanna:
Overall 👍. I wonder if there is any sense wrapping this in our own class so we can make changes to it more easily in the future.  Not sure how likely it is we would need to.

mwrock:
:+1:

lamont-granquist:
pretty certain URI.escape bit me in the ass before 👍

PR #5231, opened 2016-08-22
fix omnibus test breaks

lamont-granquist:
should fix AIX and mac


lamont-granquist:
@jkeiser this should at least make the test failures better..

lamont-granquist:
(possibly differently worse, but we'll have to see)

PR #5233, opened 2016-08-23
Add a warning for guard blocks that return a non-empty string

coderanger:
This will hopefully catch errors like this:

```ruby
myresource 'name' do
  not_if { 'some command' }
end
```

Anyone that doesn't want to see the warning can throw a `!!` in their block. I special cased empty strings as a gut feeling but I can't back that up so no argument on changing that behavior. Ping @chef/client-core for review.

### DIFF
--- a/lib/chef/resource/conditional.rb
+++ b/lib/chef/resource/conditional.rb
@@ -103,7 +103,15 @@ class Chef
       end
 
       def evaluate_block
-        @block.call
+        @block.call.tap do |rv|
+          if rv.is_a?(String) && !rv.empty?
+            # This is probably a mistake:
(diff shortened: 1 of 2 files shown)


jaymzh:
I'm fine with this. Though I feel like this should be a foodcritic rule.

coderanger:
Probably should do that too, but I think I see this at least once a week in IRC so probably prevalent enough to be in core :)

jaymzh:
I see it too. I'm not saying it shouldn't also be here, but I think people largely ignore chef warnings... linters I can easily force people to fix before committing/merging. :)

thommay:
:+1: LGTM

PR #5236, opened 2016-08-23
fix the useradd test filters

lamont-granquist:
try to document it a bit better so i don't accidentally delete it
again.


thommay:
👍

PR #5237, opened 2016-08-23
Remove remaining chef-config test warnings

jkeiser:
This fixes these two warnings:

```
rspec spec/unit/workstation_config_loader_spec.rb:284 spec/unit/workstation_config_loader_spec.rb:216
Run options:
  include {:focus=>true, :locations=>{"./spec/unit/workstation_config_loader_spec.rb"=>[284, 216]}}
  exclude {:windows_only=>true}

Randomized with seed 50388

ChefConfig::WorkstationConfigLoader
  loading the config file
    when no explicit config is specifed and no implicit config is found
/Users/jkeiser/src/chef/chef-config/lib/chef-config/path_helper.rb:238: warning: Dir.exists? is a deprecated name, use Dir.exist? instead
      skips loading
    when the config file exists
      and raises a ruby exception during evaluation
/var/folders/gf/0cklbzl144xg_pbtgpnj_4vc0000gq/T/Chef-WorkstationConfigLoader-rspec-test20160823-71386-1si9j4v:1: warning: unused literal ignored
/var/folders/gf/0cklbzl144xg_pbtgpnj_4vc0000gq/T/Chef-WorkstationConfigLoader-rspec-test20160823-71386-1si9j4v:2: warning: unused literal ignored
        raises a ConfigurationError

Finished in 0.04604 seconds (files took 0.25896 seconds to load)
2 examples, 0 failures
```

PR #5238, opened 2016-08-23
Don't dup the Cheffish "with a chef repo" context

jkeiser:
This fixes this warning at the beginning of the tests:

```
WARNING: Shared example group 'with a chef repo' has been previously defined at:
  /usr/local/var/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/cheffish-3.0.0/lib/cheffish/rspec/repository_support.rb:12
...and you are now defining it at:
  /Users/jkeiser/src/chef/spec/support/shared/integration/integration_helper.rb:111
The new definition will overwrite the original one.
```

### DIFF
--- a/spec/support/shared/integration/integration_helper.rb
+++ b/spec/support/shared/integration/integration_helper.rb
@@ -64,41 +64,6 @@ module IntegrationSupport
     Chef::ServerAPI.new
   end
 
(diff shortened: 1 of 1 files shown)


jkeiser:
Consolidated into #5237.

PR #5239, opened 2016-08-23
Get rid of "can't find win32 dll" warning in tests on non-Windows platforms

jkeiser:


jkeiser:
Consolidated into #5237.

PR #5240, opened 2016-08-23
Update all the things

jkeiser:


version_policy.rb
```diff
@@ -75,8 +75,8 @@
 # rubocop - chef-style pins to 0.39.0 in 0.3.1
 #
 ACCEPTABLE_OUTDATED_GEMS = [
-  "json",       # aws-sdk-v1 disallows JSON 2.x (no fix pending yet)
-  "rack",       # Rack 2.0+ requires Ruby 2.2
+  "json",       # inspec pins this because Ruby 2.0, no eta on fix
+  "rack",       # chef-zero pins this because Ruby 2.0, no eta on fix
```
tas50: We should fix this prior to a 5.0 release of chef-zero

version_policy.rb
```diff
@@ -75,8 +75,8 @@
...
   "rubocop",    # chef-style pins to 0.39.0 in 0.3.1
```
tas50: We should update this comment since we're on 0.42 in 0.4.0 now

version_policy.rb
jkeiser: @tas50 prior to? I don't think it's relevant to this patch, but I don't think we can pin rack to < 2.0 without a major chef-zero version bump, can we? `chef` isn't pinning here, we're just ignoring the fact that this gem is outdated because `chef-zero` is pinning it.

version_policy.rb
jkeiser: Will do.

jkeiser:
When http://manhattan.ci.chef.co/job/chef-build/368/ passes (or more particularly, shows the same results as the current master builds), I will merge, so this doesn't rot any more.

PR #5241, opened 2016-08-23
Use symbols instead of strings as keys for systemd user property

joshuamiller01:
Using the user property of the service resource when using systemd (on omnibus chef-12.13.30-1 (possibly older versions as well)) causes:
TypeError: no implicit conversion of Hash into String

This appears to be stemming from https://bugs.ruby-lang.org/issues/10708 - the type of hash key used influences how args are handled when splat / doublesplat are in play - both of which are used in newer mixin/shell_out function definitions (older "def shell_out(*command_args)" versus newer "def shell_out(*args, **options)").  Changing the hash keys to symbols in get_systemctl_options_args addresses this.

Sample code to demonstrate problem:
https://gist.github.com/joshuamiller01/43f4ccc86d69d1e3ba8f5dc357b7062e

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -78,10 +78,10 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
     if new_resource.user
       uid = node["etc"]["passwd"][new_resource.user]["uid"]
       options = {
-        "environment" => {
+        :environment => {
           "DBUS_SESSION_BUS_ADDRESS" => "unix:path=/run/user/#{uid}/bus",
         },
-        "user" => new_resource.user,
+        :user => new_resource.user,
       }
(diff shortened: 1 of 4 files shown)


joshuamiller01:

davide125:
I think https://github.com/chef/chef/blob/master/lib/chef/provider/systemd_unit.rb might a similar fix for `systemctl_opts`.

lamont-granquist:
travis angry, but 👍  on fixing this

jaymzh:
@lamont-granquist - travis is happy now but appveyer isn't... what's up with the windows world?

lamont-granquist:
dunno man, i didn't break it this time...

lamont-granquist:
magic...

jaymzh:
OK that's two :+1: s. Merging.

PR #5245, opened 2016-08-24
Don't let knife cookbook delete tests print to stdout

jkeiser:
This removes a little extra STDOUT spew from the tests:

```
Chef::Knife::CookbookDelete
  when the cookbook doesn't exist
    logs an error and exits
  when there is only one version of a cookbook
    asks for confirmation, then deletes the cookbook
    asks for confirmation before purging
  when there are several versions of a cookbook
Deleted cookbook[obsolete-cookbook][1.0.0]
Deleted cookbook[obsolete-cookbook][1.1.0]
Deleted cookbook[obsolete-cookbook][1.2.0]
    deletes all versions of a cookbook when given the '-a' flag
Deleted cookbook[obsolete-cookbook][1.0.0]
    asks which version to delete and deletes that when not given the -a flag
Deleted cookbook[obsolete-cookbook][1.0.0]
Deleted cookbook[obsolete-cookbook][1.1.0]
Deleted cookbook[obsolete-cookbook][1.2.0]
    deletes all versions of the cookbook when not given the -a flag and the user chooses to delete all

Finished in 0.1047 seconds (files took 2.63 seconds to load)
6 examples, 0 failures
```

### DIFF
--- a/spec/functional/knife/cookbook_delete_spec.rb
+++ b/spec/functional/knife/cookbook_delete_spec.rb
@@ -20,86 +20,75 @@ require "spec_helper"
 require "tiny_server"
 
 describe Chef::Knife::CookbookDelete do
-  before(:each) do
-    @server = TinyServer::Manager.new
-    @server.start
+  let(:server) { TinyServer::Manager.new }
+  let(:api) { TinyServer::API.instance }
+  let(:knife_stdout) { StringIO.new }
+  let(:knife_stderr) { StringIO.new }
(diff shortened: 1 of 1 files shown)


thommay:
Pretty sure we should just move this whole file over to use the same code as the other functional knife tests.

jkeiser:
@thommay pretty sure we should, yes! But there were just a couple of things in there that made me hesitate and think the cost would be too high for the time I have right now :)

PR #5248, opened 2016-08-24
rewrite linux_user provider check_lock

lamont-granquist:
this removes the kinda brittle kinda insane parsing of the installed
rpm version.  now we're just less strict about the exit 1 code and
fail on the output not being what is expected.

this also likely fixes a bug in rhel in why-run mode where shell_out!
would have failed on the user-not-found and raised an exception that
would have ended the run.  the code before certainly looks buggy.


lamont-granquist:
@chef/client-core finishing up the rewrite of linux_user

stevendanna:
👍 this seems much saner

Issue #5257, opened 2016-08-26
possible regression in 12.14.36


lamont-granquist:
```
Recipe: sk_ruby::default
  * chef_gem[fpm] action install

    ================================================================================
    Error executing action `install` on resource 'chef_gem[fpm]'
    ================================================================================

    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    Expected process to exit with [0], but received '2'
    ---- Begin output of /opt/chef/embedded/bin/gem install fpm -q --no-rdoc --no-ri -v "1.8.3" ----
    STDOUT:
    STDERR: ERROR:  Could not find a valid gem 'fpm' (= 1.8.3) in any repository
    ERROR:  Possible alternatives: fpm
    ---- End output of /opt/chef/embedded/bin/gem install fpm -q --no-rdoc --no-ri -v "1.8.3" ----
    Ran /opt/chef/embedded/bin/gem install fpm -q --no-rdoc --no-ri -v "1.8.3" returned 2

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:41:in `run_action'

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/sk_ruby/recipes/default.rb

     36: chef_gem "fpm" do
     37:   compile_time false if Chef::Resource::ChefGem.instance_methods(false).include?(:compile_time)
     38: end
     39:

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/sk_ruby/recipes/default.rb:36:in `from_file'

    chef_gem("fpm") do
      package_name "fpm"
      action [:install]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      declared_type :chef_gem
      cookbook_name "sk_ruby"
      recipe_name "default"
      gem_binary "/opt/chef/embedded/bin/gem"
    end

    Platform:
    ---------
    x86_64-linux


Running handlers:
[2016-08-26T08:43:34-07:00] ERROR: Running exception handlers
Running handlers complete
[2016-08-26T08:43:34-07:00] ERROR: Exception handlers complete
Chef Client failed. 1 resources updated in 25 seconds
[2016-08-26T08:43:35-07:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[...]

PR #5258, opened 2016-08-26
Fix flush_cache issues in yum_package

### LINKED ISSUE #4691: slow performance in yum_package provider in 12.7.x through 12.9.x

jaymzh:
Fixes #4691. I hope.

### DIFF
--- a/lib/chef/resource/yum_package.rb
+++ b/lib/chef/resource/yum_package.rb
@@ -27,18 +27,16 @@ class Chef
 
       # Install a specific arch
       property :arch, [ String, Array ]
-      property :flush_cache, Hash, default: { before: false, after: false }, coerce: proc { |v|
-        # TODO these append rather than set. This is probably wrong behavior, but we're preserving it until we know
-        if v.is_a?(Array)
-          v.each { |arg| flush_cache[arg] = true }
-          flush_cache
-        elsif v.any?
-          v
-        else
-          # TODO calling flush_cache({}) does a get instead of a set. This is probably wrong behavior, but we're preserving it until we know
-          flush_cache
+      property :flush_cache, Hash, default: { before: false, after: false },
+        coerce: proc do |v|
+          if v.is_a?(Array)
+            v.each_with_object({}) { |arg, obj| obj[arg] = true }
+          elsif v.any?
+            v
+          else
+            { before: v, after: v }
+          end
         end
-      }
       property :allow_downgrade, [ true, false ], default: false
       property :yum_binary, String
 
--- a/lib/chef/resource/yum_package.rb
+++ b/lib/chef/resource/yum_package.rb
@@ -28,16 +28,18 @@ class Chef
       # Install a specific arch
       property :arch, [ String, Array ]
       # the {} on the proc here is because rspec chokes if it's do...end
-      property :flush_cache, Hash, default: { before: false, after: false },
-        coerce: proc { |v|
-          if v.is_a?(Array)
-            v.each_with_object({}) { |arg, obj| obj[arg] = true }
-          elsif v.any?
-            v
-          else
-            { before: v, after: v }
-          end
-        }
+      property :flush_cache,
+               Hash,
+               default: { before: false, after: false },
+               coerce: proc { |v|
+                 if v.is_a?(Array)
+                   v.each_with_object({}) { |arg, obj| obj[arg] = true }
+                 elsif v.any?
+                   v
+                 else
+                   { before: v, after: v }
+                 end
+               }
       property :allow_downgrade, [ true, false ], default: false
       property :yum_binary, String
 
--- a/lib/chef/resource/yum_package.rb
+++ b/lib/chef/resource/yum_package.rb
@@ -27,8 +27,9 @@ class Chef
 
       # Install a specific arch
       property :arch, [ String, Array ]
+      # the {} on the proc here is because rspec chokes if it's do...end
       property :flush_cache, Hash, default: { before: false, after: false },
-        coerce: proc do |v|
+        coerce: proc { |v|
           if v.is_a?(Array)
             v.each_with_object({}) { |arg, obj| obj[arg] = true }
           elsif v.any?
@@ -36,7 +37,7 @@ class Chef
           else
             { before: v, after: v }
           end
-        end
+        }
       property :allow_downgrade, [ true, false ], default: false
       property :yum_binary, String
 


lib/chef/resource/yum_package.rb
```diff
@@ -30,13 +30,13 @@ class YumPackage < Chef::Resource::Package
       property :flush_cache, Hash, default: { before: false, after: false }, coerce: proc { |v|
         # TODO these append rather than set. This is probably wrong behavior, but we're preserving it until we know
         if v.is_a?(Array)
-          v.each { |arg| flush_cache[arg] = true }
-          flush_cache
+          x = {}
+          v.each { |arg| x[arg] = true }
```
lamont-granquist: should probably be each_with_object or collect

lamont-granquist: also you're breaking the behavior in the TODO (which i think is fine)

lib/chef/resource/yum_package.rb
```diff
@@ -30,13 +30,13 @@ class YumPackage < Chef::Resource::Package
...
+          x
         elsif v.any?
           v
         else
-          # TODO calling flush_cache({}) does a get instead of a set. This is probably wrong behavior, but we're preserving it until we know
-          flush_cache
+          { before: v, after: v}
```
lamont-granquist: shouldn't this just be v?  although i think it makes my brain hurt...   i don't think anything you do here turns this from a getter into a setter does it?

i suspect nobody uses flush_cache({}) as a no-op getter.

this code is now generally trying way to hard to be backcompat and screwing up everything in the process, i'd sort of scrap it and rewrite it cleanly, then ship fixes if we actually do break anyone (because currently we've broken /everyone/ and the code is very hard to follow).

lamont-granquist:
yeah i think delete the two TODO lines and delete the else clause entirely.

my bet is that doing so doesn't cause any regression.

if it does cause a regression, it'll be way less bad than the regression we're currently suffering under.

if we get that regression, then we can pick the best way to fix it after we see it.

jaymzh:
I don't think we can delete the else clause as we have to make it a hash if if it's not. But I'll nuke the todo.

lib/chef/resource/yum_package.rb
jaymzh: collect won't work cause need a hash, but each_with_object is fine (and the same)

jaymzh:
When I change this code it fixes it, but monkey patching it from a library doesn't seem to... not sure why. Still playing.

jaymzh:
OK @lamont-granquist I don't think I can monkey patch that using `property` keyword... property does magic...  so either I need to get a version cut with this ASAP or I need help with magic above my head :)

lamont-granquist:
uh hrm, yeah i suspect property does do lots of magic.

dunno what to tell you about getting new versions cut, i think we're still about 2 weeks-ish away from the next scheduled release based on when the last one went out.  beat on your support contract a bit and you might make it happen tho.  although we've got a bad outstanding regression in master i just cut today #5257, but i don't think it'll be anything insurmountable to fix, just another bug.

jaymzh:
`chef_gem` is evil, we don't use it so not worried. Installing gems via `gem` on prod systems is asinine.

Any ideas on how to monkey patch wahtever methods are being jammed into the `YumPackage` class? A new release seems unlikely but we're a bit fucked here. ANd I CC'd you on the ZenDesk ticket already.

lamont-granquist:
so we don't document using this property as anything other than a hash or an array.  i don't think boolean true/false have been an option?

although i don't mind adding that -- but i thought that else clause was intended to only catch the empty-hash '{}' case and attempt to coerce it into returning the same value...

lamont-granquist:
and no, my suggestion would just be to open the class and copypasta this one 'line' into it...

its possible coerce arguments get pushed into an array and you need to blow away the array?

lamont-granquist:
`Chef::Resource::YumPackage.properties.delete(:flush_cache)`?

jaymzh:
Haha I tried something simialr, but it doesn't work yet:

```
  Chef::Resource::YumPackage.properties[:flush_cache][:coerce] = proc { |v|
          if v.is_a?(Array)
            v.each_with_object({}) { |arg, obj| obj[arg] = true }
          elsif v.any?
            v
          else
            { before: v, after: v}
          end
          }
```

lamont-granquist:
yeah i'd try just nuking the property and then using the DSL to rebuild it all...

jaymzh:
Ha! This does it:

```
  opts = Chef::Resource::YumPackage.properties[:flush_cache].options
  opts[:coerce] = proc { |v|
          if v.is_a?(Array)
            v.each_with_object({}) { |arg, obj| obj[arg] = true }
          elsif v.any?
            v
          else
            { before: v, after: v}
          end
          }
  Chef::Resource::YumPackage.properties[:flush_cache] = opts
```

jaymzh:
delete doesn't work:

```
chef (12.13.37)> Chef::Resource::YumPackage.properties.delete(:flush_cache)
 => #<Chef::Property:0x00000000e03b48 @options={:default=>{:before=>false, :after=>false}, :coerce=>#<Proc:0x00000000e04598@/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.13.37/lib/chef/resource/yum_package.rb:30>, :instance_variable_name=>:@flush_cache, :name=>:flush_cache, :declared_in=>Chef::Resource::YumPackage, :is=>Hash}> 

chef (12.13.37)> Chef::Resource::YumPackage.properties[:flush_cache]
 => #<Chef::Property:0x00000000e03b48 @options={:default=>{:before=>false, :after=>false}, :coerce=>#<Proc:0x00000000e04598@/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.13.37/lib/chef/resource/yum_package.rb:30>, :instance_variable_name=>:@flush_cache, :name=>:flush_cache, :declared_in=>Chef::Resource::YumPackage, :is=>Hash}> 
```

jaymzh:
OK this seems to work....
```
  Chef::Log.info('Monkey-patching yum_package resource')
  opts = Chef::Resource::YumPackage.properties[:flush_cache].options
  opts[:coerce] = proc { |v|
          if v.is_a?(Array)
            v.each_with_object({}) { |arg, obj| obj[arg] = true }
          elsif v.any?
            v
          else
            { before: v, after: v}
          end
          }
  Chef::Resource::YumPackage.property(:flush_cache, opts)
```

If I specify a type, the validator becomes `... is => [Hash, Hash]` which is weird.

jaymzh:
OK, @lamont-granquist pushing that out to the 12.13.37 hosts. We'll see how that goes. You OK merging this meantime?

jaymzh:
replaced {} with do/end cause... style.

jaymzh:
I just figured out the actual problem!!!!

On the tier effected one of the earliest packages was setting 'flush_cache [:before]` - and since the bug was that it was additive and not per-package... that's what fucked us.

jaymzh:
wtf is this:

```
/home/travis/build/chef/chef/lib/chef/resource/yum_package.rb:31:in `proc': tried to create Proc object without a block (ArgumentError)
```

that's not true.

lamont-granquist:
whoah it wasn't per-package?  so it like updated the default value more or less?  yikes...

lamont-granquist:
yeah and i was just gonna suggest do..end, its different in subtle and annoying ways...

jaymzh:
so `do..end` breaks things. I'll go back to `{}` because... I don't even know. rubocop requires do..end for multi-line blocks. and do..end *works* but it chokes rspec. so whatever.

jaymzh:
fucking hell ruby

jaymzh:
and yes, it was effectively updating the default

jaymzh:
from 2550 second runs to 120 second runs. That's better. :)

lamont-granquist:
that explains why i couldn't replicate either, nobody reported that you had to call flush_cache once and then it flushed the cache every time after that...

jaymzh:
wasn't exactly easy to figure out :) @joshuamiller01 mentioned it was only happening on one of our tiers, so I grepped the code-base for `flush_cache` and then saw it was only in that cookbook, and that combined with the "we're appending, which seems odd" TODO comment, made it all click.

jaymzh:
OK, @lamont-granquist all the tests pass. I'm gonna have a fucking drink now.

lamont-granquist:
🍺

jaymzh:
CC @chef/client-core just for coverage.

lamont-granquist:
i'm 👍 

since its friday late, lets hang off on merging till everyone else can look at it

lamont-granquist:
(and for the record this is way too tortured to write a spec test for, i doubt it'll ever recur)

coderanger:
👍 Accidentally mutable defaults bite us again :-( Maybe for bonus "and the horse you rode in on" call `.freeze` on the default value?

thommay:
oof :( thanks for the debugging and nailing this, @jaymzh  👍

PR #5259, opened 2016-08-28
add a hook for compat_resource

lamont-granquist:
this is to address some of the regressions in compat_resource -- when we run this as a monkeypatch over old run_context classes we need to be able to skip these checks or it blows up if we have cheffish or chef-provisioning installed due to $REASONS.


thommay:
:+1: LGTM

PR #5264, opened 2016-08-29
Only support Solaris 10u11 and newer

rhass:
We no longer require the mapfile the Solaris 10u11 and newer.


lamont-granquist:
👍  i will not miss solaris mapfiles

mwrock:
looks like you need a rebase. I believe we are on mixlib-config 2.2.4 now and also make sure you have the latest commits from lcg/ruby23 in omnibus-software.

jkeiser:
:+1:

PR #5267, opened 2016-09-01
Added count_log_resource_updates config attribute

NimishaS:


NimishaS:
@btm, currently if `update_resource_count=false`, this message is show on running the recipe that contains log resource:
```
Chef Client finished, 0/1 resources updated in 28 seconds
```
So even though log resource is not counted in the updated resources, it is counted in the total no. of resources.
Is that ok?

btm:
@NimishaS yes, I think that is fine. it is still a resource, and it's probably more trouble than not to make it not be part of the total resource count. I believe the goal for the user is to be able to run chef-client twice in a row and always see 0 resources changed on the second run because there was nothing to change.

btm:
:+1:

thommay:
I wonder if this would be more useful as a configuration setting than a per-resource attribute? It seems like it's more of a policy decision than a cookbook authoring decision. I'm 👍 on the code as it stands though

btm:
@thommay yeah, I could see that. If I wanted this behavior I probably would always want it.

btm:
@NimishaS we should see if anyone else has an opinion, but I'm leaning toward we should make this a configuration setting rather than a resource attribute.

NimishaS:
@btm , please confirm the final approach that we should take.

btm:
@NimishaS please write this to be a configuration option instead of a resource attribute.

btm:
https://github.com/chef/chef-server git failures should be fixed upstream, you can rebase and push again and it should fix travis.

btm:
*suppress

this isn't a knife setting, so it shouldn't go in `Chef::Config[:knife]`. How about `Chef::Config[:count_log_resource_updates]` and default to true, as that is the current behavior?

You'll want to make an entry in `chef\lib\chef-config\config.rb` for the setting.

btm:
@NimishaS You still need:

> You'll want to make an entry in chef\lib\chef-config\config.rb for the setting.

btm:
:+1: @thommay, rereview?

lib/chef/provider/log.rb
```diff
@@ -45,7 +45,7 @@ def load_current_resource
         # true:: Always return true
         def action_write
           Chef::Log.send(@new_resource.level, @new_resource.message)
-          @new_resource.updated_by_last_action(true)
+          @new_resource.updated_by_last_action(true) unless Chef::Config[:count_log_resource_updates] == false
```
thommay: `@new_resource.updated_by_last_action(true) if Chef::Config[:count_log_resource_updates]`

NimishaS: I hadn't put `default :count_log_resource_updates, true` earlier. That's why made the condition this way.
Anyway, will make the change

Issue #5104, opened 2016-07-12
registry resource is updated when using dword 4294967295 (ffffffff)


lbn-aaio-team:
## Description

At each start of chef-client on windows, the registry key resource, when a dword max size (4294967295) value is set, is always updating the resource despite that the value is correctly set in the windows registry.

## Chef Version

chef-client x64 for windows v12.12.15 (tested with 12.11.18 too)

## Platform Version

Windows 2012 R2 x64 En-US (on vmware)

## Replication Case

registry_key "HKLM\\SYSTEM\\CurrentControlSet\\Control\\SecurityProviders\\SCHANNEL\\Ciphers\\AES 256/256" do
  values  [{:name => 'Enabled', :type => :dword, :data => '4294967295'}]
  action  :create
  recursive  true
end

Run first chef-client : the registry value is successfuly updated.
Run second chef-client command: the regitry value is updated again
The value "Enabled" is applied at each chef-client run.

## Client Output

Starting Chef Client, version 12.12.15
[2016-07-12T11:54:55+02:00] INFO: *** Chef 12.12.15 ***
[2016-07-12T11:54:55+02:00] INFO: Platform: x64-mingw32
[...]
Converging 1 resources
Recipe: win-security::security_providers
  * registry_key[HKLM\SYSTEM\CurrentControlSet\Control\SecurityProviders\SCHANNEL\Ciphers\AES 256/256] action create[201
6-07-12T11:55:05+02:00] INFO: Processing registry_key[HKLM\SYSTEM\CurrentControlSet\Control\SecurityProviders\SCHANNEL\Ciphers\AES 256/256] action create (win-security::security_providers line 1)
    - set value {:name=>"Enabled", :type=>:dword, :data=>"4294967295"}
[2016-07-12T11:55:06+02:00] INFO: Chef Run complete in 2.500202 seconds

Running handlers:
[2016-07-12T11:55:06+02:00] INFO: Running report handlers
[2016-07-12T11:55:07+02:00] INFO: Sending attributes to foreman
  - ChefHandlerForeman::ForemanFacts
Running handlers complete
[2016-07-12T11:55:07+02:00] INFO: Report handlers complete
[...]
Chef Client finished, 1/1 resources updated in 12 seconds

## Client full trace
[trace.txt](https://github.com/chef/chef/files/359025/trace.txt)

lamont-granquist:
i feel like i've read a similar bug recently, might be a dup?

lbn-aaio-team:
[...]

PR #5268, opened 2016-09-01
Fix for idempotent issue for some registry types in registry_key resource

dheerajd-msys:
Fix for issue https://github.com/chef/chef/issues/5104, similiar issue occurs for additional registry types, fixed those also.

### DIFF
--- a/spec/unit/provider/registry_key_spec.rb
+++ b/spec/unit/provider/registry_key_spec.rb
@@ -111,6 +111,13 @@ shared_examples_for "a registry key" do
         @provider.load_current_resource
         @provider.action_create
       end
+      it "should not set the value if the key exists but the datatype of data does not match" do
+        @new_resource.values( testval3_data_string )
+        expect(@double_registry).to receive(:get_values).with(keyname).and_return( testval3 )
+        expect(@double_registry).not_to receive(:set_value)
+        @provider.load_current_resource
+        @provider.action_create
+      end
       it "should set the value if the key exists but the type does not match" do
         expect(@double_registry).to receive(:get_values).with(keyname).and_return( testval1_wrong_type )
         expect(@double_registry).to receive(:set_value).with(keyname, testval1)
@@ -279,7 +286,8 @@ describe Chef::Provider::RegistryKey do
     let(:testval1_wrong_type) { { :name => "one", :type => :multi_string, :data => "1" } }
     let(:testval1_wrong_data) { { :name => "one", :type => :string, :data => "2" } }
     let(:testval2) { { :name => "two", :type => :string, :data => "2" } }
-
+    let(:testval3) { { :name => "three", :type => :dword, :data => 12345 } }
+    let(:testval3_data_string) { { :name => "three", :type => :dword, :data => "12345" } }
     it_should_behave_like "a registry key"
   end
 
@@ -289,6 +297,8 @@ describe Chef::Provider::RegistryKey do
     let(:testval1_wrong_type) { { :name => "one", :type => :string, :data => 255.chr * 1 } }
     let(:testval1_wrong_data) { { :name => "one", :type => :binary, :data => 254.chr * 1 } }
     let(:testval2) { { :name => "two", :type => :binary, :data => 0.chr * 1 } }
+    let(:testval3) { { :name => "three", :type => :binary, :data => 0.chr * 1 } }
+    let(:testval3_data_string) { { :name => "three", :type => :binary, :data => 0.chr * 1 } }
 
     it_should_behave_like "a registry key"
   end
--- a/spec/unit/provider/registry_key_spec.rb
+++ b/spec/unit/provider/registry_key_spec.rb
@@ -19,23 +19,6 @@
 require "spec_helper"
 
 shared_examples_for "a registry key" do
-  before(:each) do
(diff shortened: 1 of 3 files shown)


spec/unit/provider/registry_key_spec.rb
```diff
@@ -111,6 +111,13 @@
         @provider.load_current_resource
         @provider.action_create
       end
+      it "should not set the value if the key exists but the datatype of data does not match" do
```
btm: Can this be a more specific description? 

````
it "does not make a change when the key exists and the data matches, but the data type differs, e.g. string vs dword."
```

spec/unit/provider/registry_key_spec.rb
```diff
@@ -289,6 +297,8 @@
     let(:testval1_wrong_type) { { :name => "one", :type => :string, :data => 255.chr * 1 } }
     let(:testval1_wrong_data) { { :name => "one", :type => :binary, :data => 254.chr * 1 } }
     let(:testval2) { { :name => "two", :type => :binary, :data => 0.chr * 1 } }
+    let(:testval3) { { :name => "three", :type => :binary, :data => 0.chr * 1 } }
+    let(:testval3_data_string) { { :name => "three", :type => :binary, :data => 0.chr * 1 } }
```
btm: does this test do anything? in the other example we're passing a string, but here we're passing the same data as we are receiving.

btm: if we're not doing anything with binary cases, we should move the case into an appropriate context rather than having the test run for other cases that are meaningless.

btm:
It would be preferred to use a more descriptive variable name than `testval3`. if it becomes a single test rather than part of the shared examples it matters less, but it's good practice.

something like `dword_passed_as_string` and `dword_passed_as_integer`.

btm:
this file is lacking a lot of whitespace around the blocks too. no need to fix it all, but it'd be good to put whitespace (preceeding and following blank lines) around the blocks/tests that you add.

spec/unit/provider/registry_key_spec.rb
dheerajd-msys: @btm, yes this binary test case is not doing anything. i can add another context here for dword examples, but this would mean all the test cases(27) run for the dword sample values. currently 54 test cases are run(27 for safe and 27 for unsafe), if i add additional context for dword type, it would mean additional 27 test cases, is that fine with you?

spec/unit/provider/registry_key_spec.rb
dheerajd-msys: @btm, 

This test case is applicable when the type is  `dword, qword or dword_big_endian`  and the datatype of the data differs(the data value set is in integer and the data value passed by user is string)   
```
current_value = {:name => 'Enabled', :type => :dword, :data => 4294967295}   #current value set in registry key
value = {:name => 'Enabled', :type => :dword, :data => "4294967295"}         #value passed by user which needs to be set
```
So, updating the description as follows:- 

```
it "does not make a change when the key exists and the type matches, but the datatype of data value differs"
```

spec/unit/provider/registry_key_spec.rb
btm: Let's refactor these tests. This compromise means the test structure has us backed into a corner.

I'll try to get some time for thoughts on that later today.

dheerajd-msys:
Hi @btm, 

Refactored the code as per your suggestion.
Facing following travis failure, can someone help here.

```
Retrying `git clone 'https://github.com/chef/chef-server' "/home/travis/build/chef/chef/"vendor/bundle"/cache/bundler/git/chef-server-599ad228ae86d14671a45c681d0e8272f8342dec" --bare --no-hardlinks --quiet` due to error (4/4): Bundler::Source::Git::GitCommandError Git error: command `git clone 'https://github.com/chef/chef-server' "/home/travis/build/chef/chef/"vendor/bundle"/cache/bundler/git/chef-server-599ad228ae86d14671a45c681d0e8272f8342dec" --bare --no-hardlinks --quiet` in directory /home/travis/build/chef/chef/vendor/bundle/ruby/2.3.0/gems/chef-provisioning-aws-2.0.0 has failed.fatal: destination path '/home/travis/build/chef/chef/vendor/bundle/cache/bundler/git/chef-server-599ad228ae86d14671a45c681d0e8272f8342dec' already exists and is not an empty directory.
Git error: command `git clone 'https://github.com/chef/chef-server'
```

btm:
https://github.com/chef/chef-server git failures should be fixed upstream, you can rebase and push again and it should fix travis.

btm:
Otherwise looks good. There's more refactoring that could be done here, but happy to just get this bug fixed now. :+1:

mwrock:
I'm :+1: here but would be great to add a test to `functional/resource/registry_spec.rb` as well.

btm:
Does the functional test fail without this patch? It doesn't look like it would, as it's validating the data, not if there was a resource updated or not. I'd expect something like `expect(@new_resource).not_to be_updated_by_last_action`

dheerajd-msys:
@btm, yes the functional spec doesn't fail without the patch. Updated functional spec as per your suggestion

btm:
:+1:

PR #5276, opened 2016-09-02
tweak chef-solo run_lock timing test

lamont-granquist:
- fix the 'last' statements (my perl brain occassionally crops up
  in ruby code) and replace with a raise.  of course this just swaps
  one exception for another, but this time its really intended to
  work this way.

- swap the order of the expecations so that if we fails to see the
  waiting on the other process statement, but we blow up in both
  threads in the exception that was just added, we know we actually
  have a legit run_lock race condition that this code just found.

does not fix this bug, just gets us a bit more information if it
reoccurs.

### DIFF
--- a/spec/integration/solo/solo_spec.rb
+++ b/spec/integration/solo/solo_spec.rb
@@ -115,7 +115,7 @@ ruby_block "sleeping" do
     retries = 200
     while IO.read(Chef::Config[:log_location]) !~ /Chef client [0-9]+ is running, will wait for it to finish and then run./
       sleep 0.1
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core completely trivial tweaking to a spec

lamont-granquist:
turns out i can also replicate the race condition and it looks like Process.spawn may be buggy, so real fix might be coming shortly...

lamont-granquist:
https://bugs.ruby-lang.org/issues/10583 is likely the bug in Process.spawn.

PR #5277, opened 2016-09-02
flip chef_gem compile_time to false by default

lamont-granquist:
this may be a breaking change for some users, but the flags are
still there for users to fix on a case-by-case or global basis.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -781,7 +781,7 @@ module ChefConfig
     # true then the user will get backcompat behavior but with a single nag warning that cookbooks
     # may break with this setting in the future.  The false setting is the recommended setting and
     # will become the default.
-    default :chef_gem_compile_time, nil
+    default :chef_gem_compile_time, false
 
     # A whitelisted array of attributes you want sent over the wire when node
     # data is saved.
--- a/spec/unit/resource/chef_gem_spec.rb
+++ b/spec/unit/resource/chef_gem_spec.rb
@@ -43,7 +43,7 @@ describe Chef::Resource::ChefGem, "gem_binary" do
   end
 
   it "should set the gem_binary based on computing it from RbConfig" do
-    expect(resource.compile_time).to be nil
+    expect(resource.compile_time).to be false
   end
 
   context "when building the resource" do
@@ -59,10 +59,7 @@ describe Chef::Resource::ChefGem, "gem_binary" do
       Chef::Recipe.new("hjk", "test", run_context)
     end
 
-    let(:chef_gem_compile_time) { nil }
-
     let(:resource) do
-      Chef::Config[:chef_gem_compile_time] = chef_gem_compile_time
       Chef::Resource::ChefGem.new("foo", run_context)
     end
 
@@ -70,22 +67,13 @@ describe Chef::Resource::ChefGem, "gem_binary" do
       expect(Chef::Resource::ChefGem).to receive(:new).and_return(resource)
     end
 
-    it "runs the install at compile-time by default", chef: "< 13" do
-      expect(resource).to receive(:run_action).with(:install)
-      expect(Chef::Log).to receive(:deprecation).at_least(:once)
-      recipe.chef_gem "foo"
-    end
-
-    # the default behavior will change in Chef-13
-    it "does not runs the install at compile-time by default", chef: ">= 13" do
+    it "by default does not install at compile-time" do
       expect(resource).not_to receive(:run_action).with(:install)
-      expect(Chef::Log).not_to receive(:deprecation)
       recipe.chef_gem "foo"
     end
 
     it "compile_time true installs at compile-time" do
       expect(resource).to receive(:run_action).with(:install)
-      expect(Chef::Log).not_to receive(:deprecation)
       recipe.chef_gem "foo" do
         compile_time true
       end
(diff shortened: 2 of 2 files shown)


lamont-granquist:
probably some specs i need to fix that'll be blowing up shortly if i recall correctly....

lamont-granquist:
i guess we should probably start the deprecation of the config flag as well...

coderanger:
This setting was added in 12.1 so I think this constitutes a back compat break. Is there a reason other than footgun reduction?

coderanger:
It was even noted at the time that this would be flipped in Chef 13. https://github.com/chef/chef/commit/b4b5c07515cdde5acdf94803760f6043f4b17386#diff-f54e6ba45d43eee7cdc25838b79b012cR110

lamont-granquist:
There's some internal discussion around lets-never-do-chef-13-and-just-slowly-deprecate-and-break-things.  This is an example PR of what we'd have to do if there's no Chef 13.

I think @thommay is going to write up an RFC soon, so there will be an opportunity to have external discussion about it.

I'm just getting mildly annoyed at people pontificating about to not ever doing Chef 13, so I'm trying to press the point here and force that discussion out in the open and crystallize it...

Either we do stuff like this, in minor releases of 12, or else we'll need to do a Chef 13 release which is mostly just breaking shit at some point...

Legitimately, people are concerned about where that "...or else..." winds up.   But at the same time Chef 12 wasn't super compelling either.

coderanger:
I've also been pondering a move to CalVer instead of SemVer, so :+1: on discussing that but should consider this PR blocked on that for now :)

lamont-granquist:
Oh yeah, this is definitely a PR cut out of some grouchiness that I don't expect to see merged anytime soon...

coderanger:
Roger, also no argument that this is a positive footgun reduction just for the record. SemVer is a harsh master.

stevendanna:
> I'm just getting mildly annoyed at people pontificating about to not ever doing Chef 13, so I'm trying to press the point here and force that discussion out in the open and crystallize it...

> Either we do stuff like this, in minor releases of 12, or else we'll need to do a Chef 13 release which is mostly just breaking shit at some point...

Eh. I think most of us who have said "maybe no Chef 13" recognize that because of past decisions that we have to do 13 and are mostly looking towards a more user-friendly way to do breaking changes in the future that don't involve lots of unrelated breakages towards one when every so often (and with lots of notice) you have to change 1 thing.  It is interesting to talk about in the context of 12 vs 13 precisely because we have examples like this.

I think saying other maintainers are "pontificating" is pretty shitty, but I guess that is an issue for a different day.

lamont-granquist:
Well, I keep hearing this argument and its never been couched in those words, and its definitely been in the context of "never doing Chef 13", and it historically goes back months long before the discussion we had the other day.   So there's just more than a bit of irritation here that its very cheap to say that we should just maintain backcompat or make breaking feature via feature flags and never need to do Chef 13 when the people saying that aren't the ones trying to figure out how to write the code.  And again, I've been hearing those comments a lot lately, and its been in the context of more than just trying to deprecate with feature flags (which itself is pretty hard if you're talking about patches to core classes like the run context or resource collection which are global and where i'd prefer we didn't have multiple different chef-config flags tweaking their behavior in incompatible ways -- or just large refactorings like trying to get the knife commands to be consistent -- meaning we'd probably need to park the inconsistent code in classes somewhere where they were deprecated, which i'm not sure really makes the product better for users or devs).

stevendanna:
> So there's just more than a bit of irritation here that its very cheap to say that we should just maintain backcompat or make breaking feature via feature flags and never need to do Chef 13 when the people saying that aren't the ones trying to figure out how to write the code

Yeah, I get that frustration. Apologies for taking it too personally, I was grumpy over some more different ruby wtf's this morning and it leaked into this ticket. You are correct that all that talk is cheap and you are definitely at the tip of the spear on chef-client.

>  or just large refactorings like trying to get the knife commands to be consistent -- meaning we'd probably need to park the inconsistent code in classes somewhere where they were deprecated

On this specific problem, yes, I agree, there are lots of examples like this.  Importing some of the knife-acl and knife-opc stalled out pretty early mostly because it was so unclear what we even want the primary idiom to be in knife atm (i.e. ChefFS-style shit vs more traditional objects with methods).

lamont-granquist:
Yes, I'm neck deep in nasty case statements I don't understand and am similarly very grouchy.

I do think there's a fundamental disagreement that I have with feature flags in general in that it always forks code in the codebase which then users can use and has to get maintained and supported.  Even if we declare it is deprecated, users can still wind up on it, on the latest version of chef-client and support tickets can be cut.  That could lead to some highly unpleasant interactions where users are angrysauce at support over some bug and they need to know about different flags that control behavior that the user might have stumbled across and "oh that flag that your chef admin from 18 months ago flipped?  that is currently screwing you, you need to flip that the other way -- sorry you did it to yourself".  I mean I'm assuming our support team can deliver that message better, but it boils down to roughly the same thing -- and that's after support stumbles through asking devs to beat their head against a problem that they vaguely remember someone solved 18 months ago and digging through ever-expanding code to understand where the user might have wound up.

So, for example, we really need to kill off legacy mode for chef-solo and delete the code associated with that so that eventually we can assert that no users are using that code and support and the devs never have to think about supporting it again.  Generally I see writing more code as always ultimately being worse, and we need an answer for how to delete code.  By introducing any kind of feature flag we introduce more code, and I'd argue we're making the code worse -- for devs, for support, and ultimately for users.

I do like the idea of utilizing cookstyle/rubocop to possibly auto-fix recipe code for backcompat issues, or write a gem like transpec for chef code.   That would shift the code burden onto the linting/fixing gem and get it out of the chef/chef codebase.  That gives us a path to a better user experience, while not littering the chef/chef codebase itself with options and deprecated code.  It also doesn't require maintaining something like compat_resource with old broken chef/chef code living on in it.

PR #5279, opened 2016-09-02
use threads to workaround Process.spawn buginess in chef-solo unit test

lamont-granquist:
there's a WONTFIX closed bug in ruby-lang.org on this bug:

https://bugs.ruby-lang.org/issues/10583

they seem to be favoring speed over usability with the result
that if you're doing I/O synchronization (like our run_lock?)
inside of a Process.spawn chain like this that you have to
wrap the spawn calls inside of threads.

probably a best practice to wrap every spawn call with a
Thread.

this probably also explains why we had the silly-long 20
second timeout?

### DIFF
--- a/spec/integration/solo/solo_spec.rb
+++ b/spec/integration/solo/solo_spec.rb
@@ -133,16 +133,21 @@ EOM
         Timeout.timeout(120) do
           chef_dir = File.join(File.dirname(__FILE__), "..", "..", "..")
 
+          threads = []
+
           # Instantiate the first chef-solo run
-          s1 = Process.spawn("#{chef_solo} -c \"#{path_to('config/solo.rb')}\" -o 'x::default' \
--l debug -L #{path_to('logs/runs.log')}", :chdir => chef_dir)
+          threads << Thread.new do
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core brace yourselves, more than you ever wanted to know about Process.spawn...

lamont-granquist:
@btm this is the fix

lamont-granquist:
@btm @thommay mashing the green button will make our tests better, you know you wanna...

stevendanna:
The most satisfying part of this PR is `--l debug` -> `-l debug`.

stevendanna:
I'm 👍.  Mixing threads+fork is usually something I'm wary of (maybe urban legends?) but this is in test code and we we are doing pretty straightforward.

lamont-granquist:
http://www.linuxprogrammingblog.com/threads-and-fork-think-twice-before-using-them

so spawn isn't exactly fork and i'm not sure what exactly it is on unix, but thread+fork()+execve() works mostly right and we don't care about cloexec races.

lamont-granquist:
I'm also pretty certain that ruby >= 2.0.0 sets the cloexec flag by default when it opens fds so even that race is removed

PR #5281, opened 2016-09-03
Hook up the recipe_file_loaded event which was defined but not actually called

coderanger:
This also extends two of the recipe events to get the recipe name in addition to the path because that's usually useful for display. These arguments are both added at the end so the auto-magic argument trimmer will keep backwards compat.

Ping @chef/client-core for review. We should probably do a general audit on all the events to see if any others were defined and never cabled.

### DIFF
--- a/lib/chef/event_dispatch/base.rb
+++ b/lib/chef/event_dispatch/base.rb
@@ -232,11 +232,11 @@ class Chef
       end
 
       # Called after the recipe has been loaded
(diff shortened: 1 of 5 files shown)


danielsdeleo:
Code looks good but the tests are angry. 👍 on green

coderanger:
Test failure on chef-zero looks unrelated, going to push buttan unless someone thinks otherwise.

PR #5285, opened 2016-09-06
Adding support for cab files to Chef package provider on windows

Vasu1105:
Added resource and provider for cab_package


btm:
@mwrock you talked with @lamont-granquist about this pattern when you wrote the chocolately provider, yeah? any opinions on continuing with this pattern or writing a completely separate resource and provider for new functionality?

lamont-granquist:
+1 on completely separate cab_package resource + provider.

btm:
@Vasu1105 I thought I linked to these somewhere for you to reference, but not sure where. Maybe slack. Anyway, this is the most recent package provider on Windows. It's got newer patterns that you should use.

https://github.com/chef/chef/blob/master/lib/chef/resource/chocolatey_package.rb
https://github.com/chef/chef/blob/master/lib/chef/provider/package/chocolatey.rb

I'd only use my old old cab branch as a guide on the systems integration portion of the provider.

lib/chef/provider/package/cab.rb
```diff
@@ -0,0 +1,147 @@
...
+        def install_package(name, version)
+          dism_command("/Add-Package /PackagePath:\"#{@new_resource.source}\"")
+        end
+
+        def remove_package(name, version)
+          dism_command("/Remove-Package /PackagePath:\"#{@new_resource.source}\"")
+        end
+
+        def dism_command(command)
+          powershell_out("dism.exe /Online #{command} /NoRestart", { :timeout => @new_resource.timeout })
```
mwrock: I would use `Mixlib:ShellOut` over `powershell_out` since we are not leveraging powershell commands and powershell_out just adds command layers on top of shellout.

btm: It's confusing that this diff still shows even though a subsequent commit made the requested change.

lib/chef/provider/package/cab.rb
```diff
@@ -0,0 +1,150 @@
...
+                package_data[in_section][k] = v
+              else
+                package_data[k] = v
+              end
+            end
+          end
+          unless errors.empty?
+            if !(errors & ["0x80070003\r", "0x80070002\r"]).empty?
+              raise Chef::Exceptions::Package, "DISM: The system cannot find the path or file specified."
+            elsif errors.include?("740\r")
```
btm: Is the \r being used as a delimiter to match 740 at the end of a line? If not, do we need it? If so, maybe `errors.match(/740$/)` ? 

Carriage returns are kind of a weird character and make me nervous.

spec/unit/provider/package/cab_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+  before do
+    new_resource.source = "C:\\Temp\\Test6.1-KB2664825-v3-x64.cab"
+    installed_package_list_obj = double(stdout: installed_package_list_stdout)
+    allow(provider).to receive(:dism_command).with("/Get-Packages").and_return(installed_package_list_obj)
+    package_version_obj = double(stdout: package_version_stdout)
+    allow(provider).to receive(:dism_command).with("/Get-PackageInfo /PackagePath:\"#{new_resource.source}\"").and_return(package_version_obj)
+  end
+
+  def allow_package_info(package_path = nil, package_name = nil)
+    get_pakage_info_stdout = <<-EOF
```
btm: there are a few misspellings in variable names here, so I'm not sure if they were intentional. But we should spell them correctly so the next person to work on the code doesn't have to remember how to correctly misspell them.

*get_package_info_stdout

spec/unit/provider/package/cab_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+Package information:
+Package Identity : Package_for_KB2664825~31bf3856ad364e35~amd64~~6.1.3.0
+Applicable : Yes
+Copyright : Microsoft Corporation
+Company : Microsoft Corporation
+State : Installed
+Dependency : Language Pack
+The operation completed successfully
+    EOF
+    get_pakage_info_obj = double(stdout: get_pakage_info_stdout)
```
btm: *get_package_info_obj 
*get_package_info_stdout

spec/unit/provider/package/cab_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+    if package_path
+      allow(provider).to receive(:dism_command).with("/Get-PackageInfo /PackagePath:\"#{package_path}\"").and_return(get_pakage_info_obj)
+    else
+      allow(provider).to receive(:dism_command).with("/Get-PackageInfo /PackageName:\"#{package_name}\"").and_return(get_pakage_info_obj)
+    end
+  end
+
+  def allow_get_packags
```
btm: *allow_get_packages

lib/chef/provider/package/cab.rb
```diff
@@ -0,0 +1,150 @@
...
+              if in_section
+                package_data[in_section] = Hash.new unless package_data[in_section]
+                package_data[in_section][k] = v
+              else
+                package_data[k] = v
+              end
+            end
+          end
+          unless errors.empty?
+            if !(errors & ["0x80070003\r", "0x80070002\r"]).empty?
```
btm: this line could be easier to read. that said, we need some tests for these exceptions, I'm not sure they work as written.

```
irb(main):029:0> [ "something happened", "a bad thing with 740 happened", "another bad thing happened with 0x80070003" ].include?("0x80070003")
=> false
```

btm:
Please rebase and signoff after incorporating feedback.

lib/chef/provider/package/cab.rb
Vasu1105: @btm errors is a Array containing error codes, I am verifying here if any one of the code are present in 'errors' since both are raising similar kind of errors. With include? I can verify only one code at a time.  I want if any one of the error code present in the 'errors' it should raise 'file or path not found error' which I am not able to achieve with include? method.

btm: Ah, I missed the regex on line 107. If it's just error codes, that's fine matching an array with `includes?`. But this line still could be easier to read, even just:

`if errors.include?("0x80070002") || errors.include?("0x80070003")`

would be easier than parsing the existing operators.

Still want tests. looking up higher again, you should chomp on line 108 rather than include \r in the strings you are matching.

Vasu1105:
@btm Fixed review comments, rebase branch and signed off

PR #5287, opened 2016-09-06
fix typo to make knife bootstrap -j work

lamont-granquist:


lamont-granquist:
user reported on slack, i haven't tested.

lamont-granquist:
apparently doesn't fix it?  but this is certainly one bug of potentially several...

coderanger:
I don't think that's a typo https://github.com/chef/chef/blob/master/lib/chef/knife/core/bootstrap_context.rb#L192

lamont-granquist:
ah you are correct, nevermind then...

PR #5290, opened 2016-09-07
Normalize full key name to avoid resource update on identical key names

bai:
### Description

We have multiple apt repositories that have key names similar enough so the cookbook treats them as the same when performs `remote_file` cache (due to regex below).

Related: https://github.com/chef-cookbooks/apt/pull/207

### Issues Resolved

This PR makes sure that `remote_file` caches remote key with a unique name and thus avoids updating resources on every chef run.

### Check List
- [ ] All tests pass. 
- [x] The CLA has been signed.

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -153,7 +153,7 @@ class Chef
       end
 
       def install_key_from_uri
-        key_name = new_resource.key.split(%r{\/}).last
+        key_name = new_resource.key.gsub(/[^0-9A-Za-z\-]/, "_")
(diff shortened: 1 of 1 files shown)


thommay:
hey @bai, you just need to run `rake style:auto_correct` to fix up the chefstyle test. The chef-zero test is known broken at the moment.

thommay:
Once that's done I'm 👍

bai:
Pushed a fix

thommay:
Appveyor looks like system breakage not our breakage @chef/client-debian

bai:
I rebased from `master` but it seems to be failing anyways due to timeouts. Could you re-trigget the build please @thommay?

lamont-granquist:
👍

PR #5295, opened 2016-09-08
yet another dependency update PR

mwrock:


mwrock:
We'll remove the poise-boiler master pin once https://github.com/poise/poise-boiler/pull/6 is released

PR #5296, opened 2016-09-08
Pin oc-chef-pedant to prevent unintended test slippage

jkeiser:
Because chef-zero is pinned, oc-chef-pedant must be pinned too (as its test suite). They must move forward together or not at all. This difference caused a Travis breakage the other day.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,6 +114,7 @@ matrix:
     before_install:
       - gem update --system $(grep rubygems omnibus_overrides.rb | cut -d'"' -f2)
       - gem install bundler -v $(grep bundler omnibus_overrides.rb | cut -d'"' -f2)
+    bundler_args: --without changelog development docgen guard integration maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
     before_script:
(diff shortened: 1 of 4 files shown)


stevendanna:
@jkeiser I thought that the problem was more about the Gemfile in the `chef-zero` gem itself, since during the test we cd into that gem and bundle install from there?

jkeiser:
@stevendanna it absolutely is, but when we run chef-zero's tests, we honor any pins from `chef` itself if it has them. The previous PR from this branch had fixed `chef-zero` by pinning `oc-chef-pedant` to its older version.

These are `integration` pins only and will only affect the tests, so this is a way to get us where we want to be cheaply. Long run, we probably want to keep pins for the actual downstreams as well, but I'm not ready to make *another* pinning system right now.

stevendanna:
> we honor any pins from chef itself if it has them.

Ahhh, missed that bit. Cool 👍

mwrock:
If travis likes it I'm :+1:

jkeiser:
Appveyor succeeded earlier, the rebase just cancelled. Merging.

PR #5298, opened 2016-09-09
removing option -r for solaris machines

acaiafa:
disabling opt -r from being passed if the host is a solaris machine.

### DIFF
--- a/lib/chef/provider/group/groupadd.rb
+++ b/lib/chef/provider/group/groupadd.rb
@@ -126,7 +126,7 @@ class Chef
 
         def groupadd_options
           opts = ""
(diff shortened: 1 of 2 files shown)


tas50:
@coderanger Should be closed?

coderanger:
@tas50 Good catch, shouldn't have closed this :) Forgot github would parse that from the commit message. I resolved it only for myself.

coderanger:
This probably needs to be expanded to include the `user` resource too, and needs tests. On my list unless @acaiafa objects to me taking over :)

thommay:
Fixed by #5355

PR #5299, opened 2016-09-09
ignore unknown metadata fields in metadata.rb

lamont-granquist:
if we're just using chef-zero there's no metadata.json file so
chef-client doesn't do json parsing (which ignores unknown metadata
fields already) but parses metadata.json which explodes.

i believe since TK goes through berks it renders the rb to json before
copying to the virt so we don't see this in TK or in running chef-client
from a real server.

this only affects the chef-client -z or chef-zolo mode of operation (and
probably legacy chef-solo)

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -726,7 +726,7 @@ class Chef
         if block_given?
           super
         else
-          Chef::Log.warn "ignoring method #{method} on cookbook with name #{name}, possible typo or future metdata?"
+          Chef::Log.debug "ignoring method #{method} on cookbook with name #{name}, possible typo or future metadata?"
         end
       end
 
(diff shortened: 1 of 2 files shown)


coderanger:
Maybe include the path to the metadata.rb in the warning so its clearer, and maybe make it info instead of warn because we expect the majority of these to forward compat thangs?

lamont-granquist:
we don't have that information from within the Metadata class -- we don't even necessarily have the name of the cookbook set, i'm assuming/hoping that people will set that earlier than they'll be setting new options.

i can info instead of warn, easy enough but normal chef-client runs suppress info and only show warn with the formatter.

coderanger:
Right, and if we expect these mostly to be forward compat-y, better to not show the message I think?

lamont-granquist:
that's acceptable, i can drop it all the way down to debug then...

PR #5301, opened 2016-09-09
update release notes for 12.14.60

mwrock:


lamont-granquist:
you might want to hold off, there's a bug in compat_resource that i'm concerned may be a regression in master (and if its a regression it'll likely break a good sized chunk of our universe).  i'm chasing it and nearly have it nailed down, but don't quite have the root cause yet, but right now it seems more than a bit disconcerting.

lamont-granquist:
^ okay nevermind, i've found the issue now and its localized entirely to compat_resource

PR #5310, opened 2016-09-12
Install apt-transport-https when needed in apt_repository

tas50:
Right now a system lacking apt-transport-https will fail if you define a
https repo. We should run apt-get update and then install
apt-transport-https before defining anything with https in the URI

This is a scenario we should get into acceptance testing. If someone wants to point me in the right direction there I can make sure we test out a few apt scenarios. The apt cookbook had a healthy set of tests in it.

Signed-off-by: Tim Smith tsmith@chef.io

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -54,9 +54,13 @@ class Chef
         # make sure we have apt-transport-https installed before we try to
         # apt-get update with a https repo defined
         if new_resource.uri =~ /^https/
+          declare_resource(:apt_update, new_resource.name) do
+            ignore_failure true
+            action :periodic
+          end
+
           declare_resource(:package, "apt-transport-https") do
             action :install
-            notifies :update, "apt_update[#{new_resource.name}]", :before
           end
         end
 
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -51,6 +51,15 @@ class Chef
           end
         end
 
+        # make sure we have apt-transport-https installed before we try to
+        # apt-get update with a https repo defined
+        if new_resource.uri =~ /^https/
+          declare_resource(:package, "apt-transport-https") do
+            action :install
+            notifies :update, "apt_update[#{new_resource.name}]", :before
+          end
+        end
+
         declare_resource(:execute, "apt-cache gencaches") do
           ignore_failure true
           action :nothing
--- a/lib/chef/resource/apt_repository.rb
+++ b/lib/chef/resource/apt_repository.rb
@@ -22,7 +22,7 @@ class Chef
   class Resource
     class AptRepository < Chef::Resource
       resource_name :apt_repository
-      provides :apt_repository
+      provides :apt_repository, os: "linux"
 
       property :repo_name, String, name_property: true
       property :uri, String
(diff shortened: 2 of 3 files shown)


coderanger:
This makes me feel a bit off :-/ Like it's maybe overreaching what a core resource should do? There should at least be a way to disable it.

tas50:
Not including it potentially leaves a node in a bad broken state due to the joys of apt not supporting https out of the box. It was the main thing done in the apt::default cookbook previously. Moving this provider over to core chef so sorta forgot how people were using the apt cookbook.

coderanger:
Yeah, it's probably a common enough thing to break the mold a bit. Trying to think is there is a reason to allow disabling it since I don't think you can have a case where this would be undesirable. Only thing I can think of is if someone wrote their own replacement transport.

coderanger:
Thought on it for a day and I can't see any situation in which this isn't The Right Thing To Do. Someone somewhere will want different behavior and for them, writing a template into `sources.d/` seems fine :)

thommay:
Note that `ubuntu-standard` recommends `apt-transport-https`, so the vast majority of users will have it installed - in fact, I'd suspect that only people using docker won't have it there by default.

thommay:
👍

tas50:
@chef/client-core I reworked this one a bit to avoid the odd output that before gives. It just does a periodic apt_update now to make sure it has package data before installing the https package. I also set the provides to linux to match other apt providers.

lib/chef/provider/apt_repository.rb
```diff
@@ -51,6 +51,19 @@ def load_current_resource
...
 
+        # make sure we have apt-transport-https installed before we try to
+        # apt-get update with a https repo defined
+        if new_resource.uri =~ /^https/
+          declare_resource(:apt_update, new_resource.name) do
+            ignore_failure true
+            action :periodic
+          end
+
+          declare_resource(:package, "apt-transport-https") do
```
danielsdeleo: I wonder if we should have some kind of prefix for the resource name so we avoid triggering the duplicate resource warning.

coderanger: This is using an inline context so it shouldn't be a problem unless its also declared on a parent context.

lib/chef/provider/apt_repository.rb
```diff
@@ -51,6 +51,19 @@ def load_current_resource
           end
         end
 
+        # make sure we have apt-transport-https installed before we try to
+        # apt-get update with a https repo defined
+        if new_resource.uri =~ /^https/
+          declare_resource(:apt_update, new_resource.name) do
+            ignore_failure true
+            action :periodic
+          end
```
lamont-granquist: feel like this apt_update should probably not go here.  if we want to be defensive around call apt_update for the user before installing packages, then we should do that directly in apt_package.   kind of feels like its trying to backdoor that feature, and either we should really do that in apt_package or never do that for consistency and require the user to always apt_update first.

the `ignore_failure true` is also making me a bit twitchy.  every time i've seen `apt-get update` fail it has been for decent reasons and ignoring it would not have been good.

coderanger: I see `apt-get update` fail on Travis all the time due to transient network oopses, I throw `retries 5` on basically everything that touches the network at this point because ¯\_(ツ)_/¯.

lamont-granquist:
so I'm okay with not shipping a flag to disable this until someone hits a real issue that requires it (in fact i like that better on YAGNI grounds).

lib/chef/provider/apt_repository.rb
lamont-granquist: `retries 5` would make a lot more sense to me.   that is very different from `ignore_failure true` though...

lamont-granquist: still think we should convert this over into a PR to talk about adding that as a feature to apt_package in general tho....

tas50:
After chatting with @lamont-granquist on this one I'm going to revisit it in a simpler form after getting apt_package to auto update the package repos before installation so we get debian to behave the same way yum/zypper behave (sanely). Doing that avoids the need to sneak in an apt_update when you setup a repo.

chewi:
Be aware that if a server automatically redirects from http to https then that will also cause breakage when apt-transport-https is missing. In this case, you should probably put https in the first place but the server may change at a later date.

Issue #5309, opened 2016-09-12
support timeout attribute for service resource


ebgc:
## Description

services which need long controlled restart/start cycle (more then 10min) are not supported due to mixlib 600sec default timeout
## Chef Version

12.13.37 and any other which uses chef::mixlib
## Platform Version

any unix version using chef::mixlib
## Replication Case

have a unix service script which has an action (start,stop, reload ...) a 'sleep 610' set and it will fail.
## Mixlib::ShellOut::CommandTimeout

Command timed out after 600s:
Command exceeded allowed execution time, process terminated
## Cookbook Trace:

/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:78:in `run_action'
/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:136:in`run_delayed_notification'
/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:124:in `block in run_delayed_notifications'
/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:123:in`each'
/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:123:in `run_delayed_notifications'
/var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:113:in`converge'

ebgc:
suggestion:

chef/lib/chef/provider/service/*  need update per service action to send a 'timeout' option to mixlib which allows to override the default timeout of 600sec.

ebgc:
see also PR https://github.com/chef/chef/pull/5311

myoung34:
Any bumps to this? We use chef to deploy workers, and our service respects a timeout for threads to finish (the init.d file allows up to 45min to finish - rare but allowed).

```================================================================================
Error executing action `stop` on resource 'service[stop circusd]'
================================================================================
 
Mixlib::ShellOut::CommandTimeout
--------------------------------
Command timed out after 600s:
Command exceeded allowed execution time, process terminated
---- Begin output of /sbin/service circusd stop ----
STDOUT: Sending 'stop' to all workers
ok
There are still active worker threads. Retrying in 30 seconds
There are still active worker threads. Retrying in 30 seconds
There are still active worker threads. Retrying in 30 seconds
[...]

PR #3475, opened 2015-06-02
fix package timeout attribute

lamont-granquist:
updates all the shell_out calls to respect the timeout which is passed
on the new_resource.

the old `Chef::Config` yum timeout values are still respected and will
override the default.

the default of 900 seconds is preserved.

closes #2943 
closes #2944 
closes #2441 
closes #2444

PR #5311, opened 2016-09-12
Issue #5309 - allow timeout for service resource

ebgc:
this pr is to allow timeout to be set for service resource

test:
create a service resource and set 'timeout 1' and 'restart_command 'sleep 10''
notify this service to restart

test will fail due to timeout ... which is 1 and not 600 as by default (mixlib::shellout)

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,2 +1,4 @@
 ---
 BUNDLE_FROZEN: '1'
+BUNDLE_PATH: "/tmp/bundles/"
+BUNDLE_DISABLE_SHARED_GEMS: true
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,4 +1,4 @@
 ---
 BUNDLE_FROZEN: '1'
-BUNDLE_PATH: "/tmp/bundles/"
+BUNDLE_PATH: "/tmp/bundleschef/"
 BUNDLE_DISABLE_SHARED_GEMS: true
--- a/lib/chef/provider/service.rb
+++ b/lib/chef/provider/service.rb
@@ -26,6 +26,9 @@ class Chef
 
       include Chef::Mixin::Command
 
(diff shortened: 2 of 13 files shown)


.bundle/config
```diff
@@ -1,2 +1,4 @@
 ---
 BUNDLE_FROZEN: '1'
+BUNDLE_PATH: "/tmp/bundleschef/"
+BUNDLE_DISABLE_SHARED_GEMS: true
```
lamont-granquist: ^^^ that stuff needs to not be in this PR

lib/chef/provider/service.rb
```diff
@@ -26,6 +26,9 @@ class Service < Chef::Provider
 
       include Chef::Mixin::Command
 
+      # depends on 'require mixlib' from above 'include mixin::command'
+      TIMEOUT = Mixlib::ShellOut::DEFAULT_READ_TIMEOUT
```
lamont-granquist: if you default to nil, you can make it pick up mixlib-shellout's timeout and avoid this.

lamont-granquist:
better solution here would be to not reinvent the wheel:

https://github.com/chef/chef/pull/3475/files

extract shell_out_with_timeout to lib/chef/mixin/shell_out_with_timeout.rb and mix that into both the service and package resources and then convert the service resource over to use it.

thommay:
Hi,
we'd love to see this get merged, but right now we've got some concerns that would need to be dealt with first. Are you interested in continuing to work on this PR?
Thanks for the submission!

ebgc:
hi

will see if i find time during weekend


___
GrabCAD
Enrico B.

lamont-granquist:
this is going to get done somewhat differently via the shell_out refactoring i'm doing for chef-13

PR #5312, opened 2016-09-12
Update acceptance docs

tas50:
Format the markdown and expand on some things I ran into while going through the process


lamont-granquist:
👍  but just drop the lockfile bump

tas50:
Bump has been nuked

PR #5315, opened 2016-09-14
Adding support for Powershell package manager to Chef package provider

dheerajd-msys:
In-Progress


mwrock:
So I have been thinking of this resource and am worried that it won't provide much value. I've been using the powershell packages on a few other projects and find that `install-package` provides only base package download functionality.

However, users may want the functionality present in other cmdlets like `Install-Module` and others. In the case of powershell gallery modules (likely a more popular use case) one can use `Install-Package` and the module will be downloaded but usages of the module will not behave as expected because Install-Module sets up the paths.

We should consider how we want to expose this resource and leverage the functionality of other derivative package managers.

btm:
@mwrock does `install-module` do more than just run `install-package -provider psmodule` as indicated here? https://www.powershellgallery.com/GettingStarted?section=FAQ

Could we just have a provider property for this resource that you could pass psmodule to? maybe we'd have another resource wrapper resource ourselves?

mwrock:
I took a deeper look today and I think I was using the wrong provider name this weekend with `install-package` After using `-Provider PowerShellGet` `install-package` is working fine. My apologies for the noise here.

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+              version_list.push(stdout[stdout.index(name) + 1])
+            end
+          end
+          version_list
+        end
+
+        def find_version(stdout)
+          if stdout.empty?
+            nil
+            #raise Chef::Exceptions::Package, "Invalid Package name pecified or proper version not passed"
```
btm: Is this cruft? Is it expected we would get here?

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+              version_list.push(nil)
+            else
+              stdout = stdout.split(" ")
+              version_list.push(stdout[stdout.index(name) + 1])
+            end
+          end
+          version_list
+        end
+
+        def find_version(stdout)
```
btm: what are we really passing to `find_version`? Can we name `stdout` something more appropriate? it's overused in this file. maybe this method would be more aptly named `parse_version` and the argument is just a string and could be named string or output?

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+          if stdout.empty?
+            nil
+            #raise Chef::Exceptions::Package, "Invalid Package name pecified or proper version not passed"
+          else
+            stdout.split(" ")[-2]
```
btm: I think a comment here with an example of the line of text we're parsing would help.

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+          end
+        end
+
```
btm: don't need this extra newline.

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+              else
+                stdout = powershell_out("Get-Package -Name #{name}", { :timeout => @new_resource.timeout }).stdout
+              end
+            else
+              stdout = powershell_out("Get-Package -Name #{name}", { :timeout => @new_resource.timeout }).stdout
+            end
+            if stdout.empty?
+              version_list.push(nil)
+            else
+              stdout = stdout.split(" ")
```
btm: this is something different once you've `split` it, please name it accordingly.

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        #return array of currently installed version
+        def build_current_versions
+          version_list = []
+          new_resource.package_name.each_with_index do |name, index|
+            if !new_resource.version.nil?
+              if new_resource.version[index] != nil
+                stdout = powershell_out("Get-Package -Name #{name} -RequiredVersion #{new_resource.version[index]}", { :timeout => @new_resource.timeout }).stdout
+              else
+                stdout = powershell_out("Get-Package -Name #{name}", { :timeout => @new_resource.timeout }).stdout
```
btm: This line is the same as the one a little further down. Combine the two conditionals into one so you can use a single `else` .

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+                powershell_out( "Uninstall-Package #{name} -Force -RequiredVersion #{versions[index]}", { :timeout => @new_resource.timeout })
+              else
+                powershell_out( "Uninstall-Package #{name} -Force", { :timeout => @new_resource.timeout })
+              end
+            end
+            if new_resource.version.nil?
+              stdout = 0
+              while stdout != ""
+                stdout = powershell_out( "Uninstall-Package #{name} -Force", { :timeout => @new_resource.timeout }).stdout
+                if !stdout.empty?
```
btm: what does it mean if stdout is empty? should we raise?

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,136 @@
...
+          names.each_with_index do |name, index|
+            if !versions.nil?
+              if versions[index] != nil
+                powershell_out( "Uninstall-Package #{name} -Force -RequiredVersion #{versions[index]}", { :timeout => @new_resource.timeout })
+              else
+                powershell_out( "Uninstall-Package #{name} -Force", { :timeout => @new_resource.timeout })
+              end
+            end
+            if new_resource.version.nil?
+              stdout = 0
```
btm: 0?

lib/chef/provider/package/powershell.rb
dheerajd-msys: When no version is available online for the package, expected behaviour is to return nil to candidate_version

lib/chef/provider/package/powershell.rb
dheerajd-msys: As per @mwrock suggestion that if no version is passed during remove action, then we will uninstall all the versions of the package installed on the system. Until we keep getting some value in stdout, it means some version is getting removed of the package, and when we get stdout empty it means no version is left to be uninstalled.

lib/chef/provider/package/powershell.rb
dheerajd-msys: I am setting stdout with some random value for initialization of while loop.

dheerajd-msys:
@btm, all the review changes are done and tested. 
Let me know if any more changes are needed or we can merge the PR

btm:
@chef/client-windows review please

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,122 @@
...
+        def parse_version(output, name)
+          if output.empty?
+            nil
+            #raise Chef::Exceptions::Package, "Invalid Package name specified or proper version not passed"
+          else
+            #sample value of output variable
+            #Name                           Version          Source           Summary
+            #----                           -------          ------           -------
+            #xCertificate                   2.1.0.0          PSGallery        This module includes DSC resources that simplify administration of certificates on a Windows Server
+            output_list = output.split(" ")
```
btm: I think this pattern of splitting on a space is going to be a problem. `Get-Package | Format-List` on my Windows 10 machine has this format:

```
Microsoft Visual C++ 2012 R... 11.0.51106.1                                      Programs
```

It looks like those might be tabs between the columns which you could split on, but in either case we're going to have matching problems on packages with spaces in their names and long package names.

```
PS C:\Users\btm_000\Downloads> Get-package -Name "Picture Control Utility x64"

Name                           Version          Source                           ProviderName
----                           -------          ------                           ------------
Picture Control Utility x64    1.5.1            C:\Program Files\Common Files... msi
```

It seems likely that some day we're going to have some helper methods for specifying the format of powershell commands and then parsing the output of the objects in that format, but that doesn't need to be today.

spec/unit/provider/package/powershell_spec.rb
```diff
@@ -0,0 +1,367 @@
...
+    <<-EOF
+"\r\nName                           Version          Source           Summary                                               \r\n----                           -------          ------           -------                                               \r\nxNetworking                   2.12.0.0          PSGallery        Module with DSC resources for Networking Area"
+    EOF
+  end
+
+  let(:package_version_stdout_xnetworking) do
+    <<-EOF
+"\r\nName                           Version          Source           Summary                                               \r\n----                           -------          ------           -------                                               \r\nxNetworking                   2.12.0.0          https://www.powershellgallery... PowerShellGet"
+    EOF
+  end
```
btm: Please add a test for a package with a long name, e.g. `Security Update for Adobe Flash Player for Windows 10 Version 1607 (for x64-based Systems) (KB3194343)` so we make sure the code still works in that case.

spec/unit/provider/package/powershell_spec.rb
```diff
@@ -0,0 +1,367 @@
...
+    it "should return the correct class" do
+      expect(provider).to be_kind_of(Chef::Provider::Package::Powershell)
+    end
+  end
+
+  describe "#candidate_version" do
+
+    it "should set the candidate_version to the latest version when not pinning" do
+      my_double = double("powershellout_double")
+      allow(my_double).to receive(:stdout).and_return(package_version_stdout)
```
btm: there's a lot of duplication in the doubles in this file, and the variable names aren't very descriptive. Can you try to remove duplication and make the difference between these more clear?

On idea would be to do something like this to set up common doubles early.
```
let(:package_xnetworking_installed) do 
  stdout =  <<-EOF
"\r\nName                           Version          Source           Summary                                               \r\n----                           -------          ------           -------                                               \r\nxNetworking                   2.12.0.0          PSGallery        Module with DSC resources for Networking Area"
EOF

  double("powershell_out", :stdout => stdout)
end

let(:package_xnetworking_not_installed) do
  double("powershell_out", :stdout => "")
end
```

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,122 @@
...
+            end
+            version_list.push(parse_version(stdout.downcase, name.downcase))
+          end
+          version_list
+        end
+
+        def parse_version(output, name)
+          if output.empty?
+            nil
+            #raise Chef::Exceptions::Package, "Invalid Package name specified or proper version not passed"
```
btm: > When no version is available online for the package, expected behaviour is to return nil to candidate_version

Then we should remove this commented out code.

lib/chef/provider/package/powershell.rb
```diff
@@ -0,0 +1,122 @@
...
+          else
+            #sample value of output variable
+            #Name                           Version          Source           Summary
+            #----                           -------          ------           -------
+            #xCertificate                   2.1.0.0          PSGallery        This module includes DSC resources that simplify administration of certificates on a Windows Server
```
btm: I think it's generally more readable if you always follow a `#` for a comment with an empty space. Maybe I'm just used to it.

```
# sample value of output variable
# Name                           Version          Source           Summary
# ----                           -------          ------           -------
# xCertificate                   2.1.0.0          PSGallery        This module includes DSC resources that simplify administration of certificates on a Windows Server
```

lib/chef/provider/package/powershell.rb
mwrock: It looks like there are a few spots where the only thing you want from stdout is the name/version pairs. I'd add ` | select name, version | Format-Table -HideTableHeaders | Out-String` and trim it. Then you are guaranteed to just have names and versions delimited by a single space.

btm: The key here is there shouldn't be a space in the version, so if you split on space, you take the last value as version and everything before it as the package name to handle packages with spaces, right?

mwrock: That's right.

mwrock: Now that I'm looking at this again, it looks like all calls to `parse_version` are done after interrogating the packages of a single name so we'd expect all returned versions to have the same package name. If that's the case, we should use:
```
(XXX-Package -Name #{name} | select version | Format-Table -HideTableHeaders | Out-String).Trim()
```
Then each line in stdout is just a version and we should be able to omit this function entirely.

spec/unit/provider/package/powershell_spec.rb
dheerajd-msys: @btm, added test cases for space separated package name

mwrock:
But great job here @dheerajd-msys ! This is a fantastic addition to our suite of recources!!

btm:
@dheerajd-msys please dig into the appveyor failure and see if you have figure it out.

btm:
@mwrock release notes: https://github.com/chef/chef/pull/5491

dheerajd-msys:
@btm, the appveyor failure is caused after rebasing my code with master branch.
The failures are in Template,RemoteFile, File and CookbookFile provider specs.

btm:
This passed appveyor in a rebase in https://github.com/chef/chef/pull/5536, so should be fine.

PR #5320, opened 2016-09-14
Allow the :delete action for yum_repository + fix old property support

tas50:
The :delete action wasn't working and this is the documented action so this is a pretty big break. Also we weren't allowing the legacy properties that a good chunk of people are still using.

### DIFF
--- a/lib/chef/resource/yum_repository.rb
+++ b/lib/chef/resource/yum_repository.rb
@@ -29,7 +29,7 @@ class Chef
       property :cost, String, regex: /^\d+$/
       property :clean_headers, [TrueClass, FalseClass], default: false # deprecated
       property :clean_metadata, [TrueClass, FalseClass], default: true
-      property :description, String, regex: /.*/, default: "YUM Repository"
+      property :description, String, regex: /.*/, default: "Yum Repository"
       property :enabled, [TrueClass, FalseClass], default: true
       property :enablegroups, [TrueClass, FalseClass]
       property :exclude, String, regex: /.*/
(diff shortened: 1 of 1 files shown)


lib/chef/resource/yum_repository.rb
```diff
@@ -29,7 +29,7 @@ class YumRepository < Chef::Resource
       property :cost, String, regex: /^\d+$/
       property :clean_headers, [TrueClass, FalseClass], default: false # deprecated
       property :clean_metadata, [TrueClass, FalseClass], default: true
-      property :description, String, regex: /.*/, default: "Ye Ole Rpm Repo"
+      property :description, String, regex: /.*/, default: "Yum Repository"
```
shortdudey123: Is this change to differentiate between the Chef resource and the yum cookbook resource?

lib/chef/resource/yum_repository.rb
```diff
@@ -66,11 +66,14 @@ class YumRepository < Chef::Resource
...
-
       property :options, Hash
 
       default_action :create
-      allowed_actions :create, :remove, :make_cache, :add
+      allowed_actions :create, :remove, :make_cache, :add, :delete
+
+      # provide compatibility with the yum cookbook < 3.0 properties
+      alias_method :url, :baseurl
+      alias_method :keyurl, :gpgkey
```
shortdudey123: Add deprecation notice for these 2 alias's? or support them long term?

lib/chef/resource/yum_repository.rb
tas50: It seemed like a more professional default. The yum cookbook should probably get updated as well.

lib/chef/resource/yum_repository.rb
tas50: For now we're just supporting them until the end of time. If we really care to get everyone on the new properties we can introduce some sort of deprecation logging, but that would add complexity here. 2 lines for backwards compat is simpler.

lamont-granquist:
👍

noslowerdna:
We would like to request a release of https://github.com/chef/chef/commits/v12.14.65 to https://packages.chef.io/stable/el/6/chef-12.14.65-1.el6.x86_64.rpm

lamont-granquist:
we're working on one or two other regressions as well, it'll be shipped when its ready

noslowerdna:
@lamont-granquist ok, thanks

PR #5322, opened 2016-09-14
revert supports[:manage_home] behavior

lamont-granquist:
even though i violently disagree that this is correct behavior and we're
just going to break everyone in one massive go when we hit Chef 13.

### DIFF
--- a/lib/chef/resource/user/linux_user.rb
+++ b/lib/chef/resource/user/linux_user.rb
@@ -35,19 +35,6 @@ class Chef
           @manage_home = false
         end
 
-        def supports(args = {})
-          if args.key?(:manage_home)
-            Chef.log_deprecation "supports { manage_home: #{args[:manage_home]} } on the user resource is deprecated and will be removed in Chef 13, set manage_home: #{args[:manage_home]} instead"
-          end
-          if args.key?(:non_unique)
-            Chef.log_deprecation "supports { non_unique: #{args[:non_unique]} } on the user resource is deprecated and will be removed in Chef 13, set non_unique: #{args[:non_unique]} instead"
-          end
-          super
-        end
-
-        def supports=(args)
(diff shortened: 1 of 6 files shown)


lamont-granquist:
@thommay probably up to you as to if this is a regression or not.

should probably get feedback from an affected user that this reverts to the old behavior of supports[:manage_home]

coderanger:
👍 from me, but I also defer to @thommay's wisdom.

bradenwright:
@lamont-granquist lol at the "violently" disagree, I may steal that

lamont-granquist:
pretty certain i stole it from someone on Usenet around 1992 (or some dragcit BBS or something back in the dawn of time)

lib/chef/resource/user/linux_user.rb
```diff
@@ -29,21 +29,24 @@ class LinuxUser < Chef::Resource::User
...
 
         def supports(args = {})
-          Chef.log_deprecation "setting supports on the linux_user resource is deprecated"
-          # setting is deliberately disabled
-          super({})
+          if args.key?(:manage_home)
+            Chef.log_deprecation "supports { manage_home: #{args[:manage_home]} } on the user resource is deprecated and will be removed in Chef 13, set manage_home: #{args[:manage_home]} instead"
+          end
+          if args.key?(:non_unique)
+            Chef.log_deprecation "supports { manage_home: #{args[:non_unique]} } on the user resource is deprecated and will be removed in Chef 13, set manage_home: #{args[:non_unique]} instead"
```
gregkare: This should be `non_unique` everywhere on this line, not `manage_home`

lamont-granquist: fixed

mwrock:
:+1:

PR #5323, opened 2016-09-14
Update the changelog and release notes for 12.14.60/65

tas50:
We missed a few important things in the release notes and we needed to call out the yum_repository work in a way that it could be cmd-F'd in the changelog.


mfischer-zd:
Thanks, Tim.

tas50:
I'm going to remove the .65 changelog portion of this for now since it looks like that won't be the new build

PR #5325, opened 2016-09-15
Fix for ifconfig resource's functional spec.

Aliasgar16:
Modified `ifconfig` resource's functional spec to handle `persistent stateless network interface names (deprecation of eth0)`, introduced in `Ubuntu 16.04` OS.

### DIFF
--- a/spec/functional/resource/ifconfig_spec.rb
+++ b/spec/functional/resource/ifconfig_spec.rb
@@ -51,7 +51,7 @@ describe Chef::Resource::Ifconfig, :requires_root, :skip_travis, :external => in
     end
   end
 
-  def interface_name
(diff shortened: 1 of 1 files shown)


Aliasgar16:
Done.

spec/functional/resource/ifconfig_spec.rb
```diff
@@ -51,11 +51,17 @@ def lo_interface_for_test
     end
   end
 
+  def interface_name
```
btm: this should be called something like `fetch_first_interface_name`

Aliasgar16: @btm I have modified the method name.

mwrock:
👍

Issue #5324, opened 2016-09-15
systemd_unit resource not creating units anymore


fphilippon:
## Description

Following #5210 the systemd_unit resource seems not be able to create systemd units anymore.

## Chef Version

Chef 12.14.60

## Platform Version

CentOS 7.2
Chef: 12.14.60

## Replication Case

Creating systemd unit using the systemd_unit resource.

I suspect the new following method in lib/chef/provider/systemd_unit.rb to always return false:

```
def systemd_analyze_cmd
  @systemd_analyze_cmd ||= "#{systemd_analyze_path} verify %{path}"
end
```
Furthermore i did a test without the %{path} variable and the check works as expected (success when systemd unit file looks good and fail when it's not).

## Client Output
```

 * systemd_unit[jmx_exporter_kafka_521707.service] action create
  Recipe: <Dynamically Defined Resource>
    * file[/etc/systemd/system/jmx_exporter_kafka_521707.service] action create
      
      ================================================================================
      Error executing action `create` on resource 'file[/etc/systemd/system/jmx_exporter_kafka_521707.service]'
      ================================================================================
      
      Chef::Exceptions::ValidationFailed
      ----------------------------------
      Proposed content for /etc/systemd/system/jmx_exporter_kafka_521707.service failed verification #<Chef::Resource::File::Verification:0x000000071c5f00>
      
      Cookbook Trace:
      ---------------
      /var/chef/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/runner.rb:41:in `run_action'
      
      Compiled Resource:
      ------------------
      # Declared in 
      
      file("/etc/systemd/system/jmx_exporter_kafka_521707.service") do
        action [:create]
        retries 0
        retry_delay 2
        default_guard_interpreter :default
        mode "0644"
        content "[Unit]\nDescription = jmx exporter\nAfter = network.target\n\n[Service]\nType = simple\nUser = prometheus\nGroup = prometheus\nRestart = on-failure\nExecStart = /usr/bin/java -Dcom.sun.management.jmxremote.ssl=false   -Dcom.sun.management.jmxremote.authenticate=false   -Dcom.sun.management.jmxremote.port=5555 -jar                 /opt/jmx_exporter/jmx-exporter.jar                 5556                 /opt/jmx_exporter/kafka_521707.yml\n\n[Install]\nWantedBy = multi-user.target\n"
        path "/etc/systemd/system/jmx_exporter_kafka_521707.service"
        owner "root"
[...]

PR #5210, opened 2016-08-16
Verify systemd_unit file during create

mal:
Fixes #5200 

I realise there are no tests for this feature. That is because I wasn't able to find any pre-existing tests relating to the internal file block on which to base any new ones.

PR #5326, opened 2016-09-15
Revert "Verify systemd_unit file during create"

mwrock:
Reverts chef/chef#5210 since this appears to be a breaking change. See https://github.com/chef/chef/issues/5324

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -193,7 +193,6 @@ class Chef
           f.group "root"
           f.mode "0644"
           f.content new_resource.to_ini
-          f.verify systemd_analyze_cmd if systemd_analyze_path
         end.run_action(action)
       end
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
well, my reflexive :-1: on reflexively reverting any change.

do we understand why this broke and if we can ship a simple fix that maintains the feature?

stevendanna:
> do we understand why this broke and if we can ship a simple fix that maintains the feature?

The core issue seems to be that systemd-analyze verify requires the extension to indicate the unit type:

```
[root@localhost ~]# systemd-analyze verify ./foobar
Failed to load ./foobar: Invalid argument
[root@localhost ~]# systemd-analyze verify ./foobar.service
[root@localhost ~]# echo $?
0
```

One possible solution would be to create a copy of the tempfile with the right extension, verify that, and then clean it up.

stevendanna:
Currently some discussion about a new feature that allows you to specify the extension of the tempfile that we create.

mwrock:
Thanks. I'm closing.

mwrock:
reopenning since the temp file fix, though still of value, has not completely fixed this issue.

stevendanna:
👍

lamont-granquist:
👍

nathwill:
verified this is working again in 12.14.90, thanks @mwrock!

PR #5327, opened 2016-09-15
preserve the extension of the file in the rendered tempfile in File providers

lamont-granquist:
helps out with verification steps where the tempfile may be passed to a program
which gets annoyed if the extension isn't correct.

closes https://github.com/chef/chef/issues/5324

### DIFF
--- a/lib/chef/file_content_management/tempfile.rb
+++ b/lib/chef/file_content_management/tempfile.rb
@@ -40,7 +40,7 @@ class Chef
 
         tempfile_dirnames.each do |tempfile_dirname|
           begin
-            tf = ::Tempfile.open(tempfile_basename, tempfile_dirname)
+            tf = ::Tempfile.open([tempfile_basename, tempfile_extension], tempfile_dirname)
             break
           rescue SystemCallError => e
             message = "Creating temp file under '#{tempfile_dirname}' failed with: '#{e.message}'"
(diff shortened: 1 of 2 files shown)


lamont-granquist:
🚧  no tests

lib/chef/file_content_management/tempfile.rb
```diff
@@ -63,12 +63,16 @@ def tempfile_open
       # as the arguments to Tempfile.new() consistently.
       #
       def tempfile_basename
-        basename = ::File.basename(@new_resource.name)
+        basename = ::File.basename(@new_resource.name, File.extname(@new_resource.name))
```
marcparadise: Should this use `tempfile_extension` as well, instead of `File.extname`?

lamont-granquist: yeah, i had to modify tempfile_extension and therefore had to fix that as well

lamont-granquist:
having been involved in this code from its inception in 11.6.0 i'm reasonably certain that we're not changing anything that was ever a public API here.  we did have the ".chef-" prefix as a public API which was deliberately added some time ago and that is unchanged.  i'm pretty certain that if anyone cared about the extension that they would have wanted this fix and we just discovered the very first consumer that cares.

lamont-granquist:
@chef/client-core reviewage

lib/chef/file_content_management/tempfile.rb
```diff
@@ -63,12 +64,18 @@ def tempfile_open
       # as the arguments to Tempfile.new() consistently.
       #
       def tempfile_basename
-        basename = ::File.basename(@new_resource.name)
+        basename = ::File.basename(@new_resource.path, tempfile_extension)
+        # the leading "[.]chef-" here should be considered a public API and should not be changed
```
coderanger: I'm confused why the extension is used here too.

lamont-granquist: that strips it so we get the prefixey part of the basename so that "foo.html.erb" becomes "chef-foo" here and the extension becomes ".html.erb"

coderanger: Derp, I was reading that as `File.join`. Got it :)

PR #5335, opened 2016-09-16
fix non-utf8 filename issues in the file provider (again)

lamont-granquist:
it looks like we broke this awhile back with the docker specific check in the file provider (but only broke it in docker containers)

and the tempfile fix that went into master just broke this as well for everyone, but that version was never released.

adds a test (which conveniently runs in a docker container) so that hopefully we don't regress again.


mwrock:
:+1:

PR #5339, opened 2016-09-16
Make 12.5+ custom resources whyrun-safe by default

jkeiser:


lamont-granquist:
i could have sworn they were this way from the start  👍

coderanger:
@marcparadise This will only affect the 12.5-style fused syntax, most of which are written using Chef's core resources which are generally why-run safe.

lamont-granquist:
yeah "custom resources" != "LWRPs"

and generally we beat on people about how to properly write resources that it makes more sense to mark even LWRPs as why-run safe by default -- but probably not until chef 13.

jkeiser:
@marcparadise yes, this won't affect LWRPs that declare actions in `providers/blah.rb` or HWRPs that do `class Blah < Chef::Provider`. Only resources.

marcparadise:
Okay, makes sense - thanks for the clarification.

PR #5342, opened 2016-09-17
cleanup groups at the start of the tests

mwrock:


rhass:
Is there any way we can set the user id range in a higher range such as >= 30000?

rhass:
:+1:

spec/functional/resource/group_spec.rb
```diff
@@ -169,8 +173,11 @@ def create_group
 
     describe "when the users exist" do
       before do
+        high_uid = 60000
```
lamont-granquist: little worried that could interact badly/surprisingly with systems that have 32-bit signed ints for uids.   i can't recall when linux switched.  no idea when/if solaris or aix is off of those.  dunno...

mwrock: It passes all of our test nodes. I ran an ad hoc build on this.

lamont-granquist: make that 16-bit signed ints...   i can't even think that small any more...

mwrock: ok i think thats fine. knocked it down to 30000

PR #5346, opened 2016-09-19
Alias older :makecache rather than new :make_cache only.

mjuarez:
## Problem
The recent port of the yum_repository resource into chef 12.14.60 has caused issues with the make cache action on a repo. yum@4.0.0 [appears to use](https://github.com/chef-cookbooks/yum/blob/master/resources/repository.rb#L21) `:makecache` as the action, however the new port is using `:make_cache`. This appears to be breaking cookbooks that utilize the older resource from the yum Cookbook when the newer core cookbook takes precedence.

## Solution
This adds in `:makecache` as an alias to the `:make_cache` action in the new chef core resource.

### DIFF
--- a/kitchen-tests/cookbooks/base/recipes/default.rb
+++ b/kitchen-tests/cookbooks/base/recipes/default.rb
@@ -24,6 +24,7 @@ yum_repository "epel" do
   gpgcheck true
   mirrorlist "https://mirrors.fedoraproject.org/metalink?repo=epel-#{node['platform_version'].to_i}&arch=$basearch"
   only_if { node["platform_family"] == "rhel" }
+  action [ :add, :makecache ]
 end
 
 include_recipe "build-essential"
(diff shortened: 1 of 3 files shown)


mjuarez:
Just looked again, and not convinced that the `:make_cache` action would have ever succeeded due to a mismatch in provider/resources:
https://github.com/chef/chef/blob/v12.14.60/lib/chef/resource/yum_repository.rb#L73
https://github.com/chef/chef/blob/v12.14.60/lib/chef/provider/yum_repository.rb#L97

Allowed action is `:make_cache` but only `:makecache` is provided.

Should this fix be resolved to just `:makecache` to conform to previous yum cookbook, or provide both forms? Since 12.14.60 was a formal release, I'd imagine people could be consuming this new form and corrected accordingly where possible.

tas50:
Thanks for bringing this to our attention. It looks like makecache is the correct action name and make_cache was a typo based on the action of the same name.  I've opened a PR to fix the allowed_actions to match what we have in the yum cookbook.

#5348

PR #5347, opened 2016-09-19
Actually clean up bundler files

jkeiser:
We currently clean up bundler's `git`-based gem directory in the `chef` software definition. However, the `chef-appbundler` software definition runs afterward and can put things back in there. To fix this, we now clean bundler's directories after chef-appbundler, in a separate step. We merged this with the chef-remove-docs step.


scotthain:
👍

mwrock:
👍

PR #5351, opened 2016-09-20
fix method_access and array handling in node presenter

### LINKED ISSUE #5349: "knife node show -a" incorrectly warns about method access for attributes
## Description

When searching for an attribute that does not exist, knife warns about using method access even if the method access syntax was not used to search for nested attributes.

Here is the actual output:
```
[birdsnest ~]% knife node show -l mynode -a 'my_non_existent_attr'
WARN: method access to node attributes (node.foo.bar) is deprecated and will be removed in Chef 13, please use bracket syntax (node["foo"]["bar"]) at /opt/chefdk/embedded/lib/ruby/2.1.0/forwardable.rb:183:in `format_for_display'
mynode:
  my_non_existent_attr:
```
Here is the expected output:
```
[birdsnest ~]% knife node show -l mynode -a 'my_non_existent_attr'
mynode:
  my_non_existent_attr:

```

## Chef Version

```
[birdsnest ~]% pacman -Qi chef-dk
Name            : chef-dk
Version         : 0.17.17-1
Description     : The Chef Development Kit contains everything you need to start using Chef
                  along with the tools essential to managing the code that runs your business.
Architecture    : x86_64
URL             : https://downloads.chef.io/chef-dk/
Licenses        : Apache
Groups          : None
Provides        : None
Depends On      : None
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : chef  chef-solo
Replaces        : None
Installed Size  : 273.89 MiB
Packager        : Unknown Packager
Build Date      : Thu 18 Aug 2016 10:01:09 AM MDT
Install Date    : Thu 18 Aug 2016 10:03:24 AM MDT
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : None
```

## Platform Version

Arch Linux

lamont-granquist:
closes #5349 

### DIFF
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -187,7 +187,7 @@ class Chef
                 nil
               end
           end
-          # necessary for coercing hashable non-attribute objects hanging off the node into real hashes
+          # necessary (?) for coercing objects (the run_list object?) to hashes
           ( !data.kind_of?(Array) && data.respond_to?(:to_hash) ) ? data.to_hash : data
         end
 
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -175,23 +175,19 @@ class Chef
 
         def extract_nested_value(data, nested_value_spec)
           nested_value_spec.split(".").each do |attr|
-            if data.nil?
-              nil # don't get no method error on nil
-              # Must check :[] before attr because spec can include
-              #   `keys` - want the key named `keys`, not a list of
-              #   available keys.
-            elsif data.respond_to?(:[]) && data.has_key?(attr)
-              data = data[attr]
-            elsif data.respond_to?(attr.to_sym)
-              data = data.send(attr.to_sym)
-            else
-              data = begin
-                       data.send(attr.to_sym)
-                     rescue NoMethodError
-                       nil
-                     end
-            end
+            data =
+              if data.is_a?(Array)
+                data[attr.to_i]
+              elsif data.respond_to?(:[], false) && data.key?(attr)
+                data[attr]
+              elsif data.respond_to?(attr.to_sym, false)
+                # handles -a chef_environment and other things that hang of the node and aren't really attributes
+                data.public_send(attr.to_sym)
+              else
+                nil
+              end
           end
+          # necessary for coercing hashable non-attribute objects hanging off the node into real hashes
           ( !data.kind_of?(Array) && data.respond_to?(:to_hash) ) ? data.to_hash : data
         end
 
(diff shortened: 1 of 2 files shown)


lamont-granquist:
the return value for the array in the spec looks odd, but the presenter seems to do the right thing with it:

```
% knife node show crash.scriptkiddie.org -a ohai.disabled_plugins.0
crash.scriptkiddie.org:
  ohai.disabled_plugins.0: linux::block_device
```

vs. what we have in master now:

```
% knife node show crash.scriptkiddie.org -a ohai.disabled_plugins.0
This may be a bug in the 'node show' knife command or plugin
Exception: NoMethodError: undefined method `has_key?' for #<Chef::Node::ImmutableArray:0x007fbf9aa7ee58>
```

lamont-granquist:
hrm, solo spec died -- please don't restart that one, i want to investigate

lamont-granquist:
@chef/client-core reviewage needed

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -175,23 +175,19 @@ def formatting_subset_of_data?
...
-                       nil
-                     end
-            end
+            data =
+              if data.is_a?(Array)
+                data[attr.to_i]
+              elsif data.respond_to?(:[], false) && data.key?(attr)
+                data[attr]
+              elsif data.respond_to?(attr.to_sym, false)
+                # handles -a chef_environment and other things that hang of the node and aren't really attributes
```
coderanger: 😢 but can't fix that today.

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -175,23 +175,19 @@ def formatting_subset_of_data?
...
-            elsif data.respond_to?(attr.to_sym)
-              data = data.send(attr.to_sym)
-            else
-              data = begin
-                       data.send(attr.to_sym)
-                     rescue NoMethodError
-                       nil
-                     end
-            end
+            data =
```
coderanger: I feel like this could be simplified using the new `.read` since non-attribute data will never be nested so we can constrain that wonkiness to `nested_value_spec.split(".").length == 1` or somesuch?

```ruby
value_path = nested_value_spec.split(".")
if value_path.length == 1 && node.respond_to?(value_path.first)
   node.send(value_path.first)
else
  node.read(*value_path)
end
```

That might be more invasive than you want to go for this change though.

coderanger: Also this would only work with nodes, not sure if `-a` exists on any non-node data sources. Probably does, just to make things difficult.

lamont-granquist: i think your logic has an issue thought that we actually want `-a class` to be `node[:class]` and not `node.class`

which is kind of why i'm hesitant to muck with the algorithm that much...

i should add a test for that....

lamont-granquist: yeah and i think "data" can be just a plain-old-hash, so can't assume node.read() exists.

coderanger: Sounds about right, a fight for another day :)

lamont-granquist:
^ added a spec for the attributes-must-win-over-methods edge case.  tried the node.read() version of the code and it made quite a mess in the tests for other reason as well...

PR #5352, opened 2016-09-20
Updated location for latest release notes

davidwrede:
Release notes are now aggregated for all minor releases, so changing the docs.chef.io location to reflect that change.


lamont-granquist:
LGTM

PR #5355, opened 2016-09-20
Don't use -r for users or groups on Solaris.

coderanger:
Solaris has no concept of "system users" or "system groups" and so doesn't support these options. Also includes test refactors because they made me sad.

Ping @chef/client-solaris for review. Replaces #5298.


lib/chef/provider/user/useradd.rb
```diff
@@ -140,7 +140,8 @@ def update_options(field, option, opts)
 
         def useradd_options
           opts = []
-          opts << "-r" if new_resource.system
+          # Solaris doesn't support system users.
+          opts << "-r" if new_resource.system && !node.platform?("solaris2")
```
lamont-granquist: hrmmm, i seem to have done a github review fail....   

this should really go into the user/solaris.rb file rather than special casing it here.

the useradd class here is in the process of getting abandoned and i'm going to extract the code out to solaris, aix, and linux providers that inherit only from User.   you could start by just copying useradd_options directly into the solaris provider and then changing it.

also its probably not only the solaris2 platform that needs this but all the solarisish platforms as well.

coderanger: Works for me, will fix up.

lib/chef/provider/group/groupadd.rb
```diff
@@ -126,7 +126,8 @@ def set_options
 
         def groupadd_options
           opts = ""
-          opts << " -r" if @new_resource.system
+          # Solaris doesn't support system groups.
+          opts << " -r" if @new_resource.system && !node.platform?("solaris2")
```
coderanger: Think I should add `, "omnios"` here? I don't know those Solarish platforms well enough.

coderanger:
@lamont-granquist Failure look unrelated and I think I've seen you working on that in another PR, yes? (chef-solo timeout issues)

lamont-granquist:
chef-solo timeout should be fixed.

PR #5356, opened 2016-09-20
Update pedump to latest released; update all dependencies

jkeiser:


lamont-granquist:
chefspec will go red, being fixed in https://github.com/sethvargo/chefspec/pull/762

PR #5359, opened 2016-09-21
Allow deletion of registry_key without the need for users to pass data key in values hash.

Aliasgar16:
In-progress.

### DIFF
--- a/lib/chef/provider/registry_key.rb
+++ b/lib/chef/provider/registry_key.rb
@@ -70,26 +70,50 @@ class Chef
         end
       end
 
+      def key_missing?(values, name)
+        values.each do |v|
+          return true unless v.has_key?(name)
+        end
+        false
+      end
+
       def define_resource_requirements
         requirements.assert(:create, :create_if_missing, :delete, :delete_key) do |a|
           a.assertion { registry.hive_exists?(@new_resource.key) }
           a.failure_message(Chef::Exceptions::Win32RegHiveMissing, "Hive #{@new_resource.key.split("\\").shift} does not exist")
         end
+
         requirements.assert(:create) do |a|
           a.assertion { registry.key_exists?(@new_resource.key) }
           a.whyrun("Key #{@new_resource.key} does not exist. Unless it would have been created before, attempt to modify its values would fail.")
         end
+
         requirements.assert(:create, :create_if_missing) do |a|
-          #If keys missing in the path and recursive == false
+          # If keys missing in the path and recursive == false
           a.assertion { !registry.keys_missing?(@current_resource.key) || @new_resource.recursive }
           a.failure_message(Chef::Exceptions::Win32RegNoRecursive, "Intermediate keys missing but recursive is set to false")
           a.whyrun("Intermediate keys in #{@new_resource.key} do not exist. Unless they would have been created earlier, attempt to modify them would fail.")
         end
+
         requirements.assert(:delete_key) do |a|
-          #If key to be deleted has subkeys but recurssive == false
+          # If key to be deleted has subkeys but recurssive == false
           a.assertion { !registry.key_exists?(@new_resource.key) || !registry.has_subkeys?(@new_resource.key) || @new_resource.recursive }
           a.failure_message(Chef::Exceptions::Win32RegNoRecursive, "#{@new_resource.key} has subkeys but recursive is set to false.")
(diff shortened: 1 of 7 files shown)


Aliasgar16:
Done.

mwrock:
I think we should also ignore `:type` when deleting keys.

Aliasgar16:
@mwrock I will make the changes.

Aliasgar16:
Done with the changes.

mwrock:
👍

lib/chef/provider/registry_key.rb
```diff
@@ -70,6 +70,13 @@ def values_to_hash(values)
         end
       end
 
+      def key_missing?(values, name)
+        values.each do |v|
+          return true unless v.has_key?(name)
+        end
+        return false
```
btm: this could be just `false`

https://github.com/bbatsov/ruby-style-guide#no-explicit-return

lib/chef/provider/registry_key.rb
```diff
@@ -91,6 +98,18 @@ def define_resource_requirements
           a.failure_message(Chef::Exceptions::Win32RegNoRecursive, "#{@new_resource.key} has subkeys but recursive is set to false.")
           a.whyrun("#{@current_resource.key} has subkeys, but recursive is set to false. attempt to delete would fails unless subkeys were deleted prior to this action.")
         end
+        requirements.assert(:create, :create_if_missing) do |a|
+          #If type key missing in the RegistryKey values hash
+          a.assertion { !key_missing?(@new_resource.values, :type) }
+          a.failure_message(Chef::Exceptions::RegKeyValuesTypeMissing, "Missing type key in RegistryKey values hash")
+          a.whyrun("Type key does not exist. Attempt would fail unless the complete values hash containing all the keys does not exist for registry_key resource's create action.")
+        end
+        requirements.assert(:create, :create_if_missing) do |a|
```
btm: I realize this is the pattern of the whole file, but putting a blank line between the requirement blocks would make this easier to read.

lib/chef/provider/registry_key.rb
```diff
@@ -91,6 +98,18 @@ def define_resource_requirements
...
+          #If data key missing in the RegistryKey values hash
```
btm: A space after the # would improve readability, e.g. `# If the data key is missing in the RegistryKey values hash`

spec/functional/resource/registry_spec.rb
```diff
@@ -285,6 +309,30 @@ def create_deletable_keys
         @new_resource.run_action(:create)
         expect(@registry.key_exists?(reg_child + '\Slitheen')).to eq(false)
       end
+      it "does not throw an exception if the action create and type key missing in values hash" do
+        @new_resource.key(reg_child + '\Slitheen')
+        @new_resource.values([{ :name => "BriskWalk", :data => "my_data" }])
+        @new_resource.run_action(:create) # should not raise_error
+        expect(@registry.key_exists?(reg_child + '\Slitheen')).to eq(false)
+      end
+      it "does not throw an exception if the action create and data key missing in values hash" do
```
btm: add whitespace around these blocks too please.

spec/functional/resource/registry_spec.rb
```diff
@@ -285,6 +309,30 @@ def create_deletable_keys
         @new_resource.run_action(:create)
         expect(@registry.key_exists?(reg_child + '\Slitheen')).to eq(false)
       end
+      it "does not throw an exception if the action create and type key missing in values hash" do
```
btm: since we're actually raising, we should be pedantic and say that it `does not raise an exception` rather than throw.

https://coderwall.com/p/lhkkug/don-t-confuse-ruby-s-throw-statement-with-raise

spec/functional/resource/registry_spec.rb
```diff
@@ -337,6 +385,30 @@ def create_deletable_keys
       expect { @new_resource.run_action(:create_if_missing) }.to raise_error(Chef::Exceptions::Win32RegNoRecursive)
     end
 
+    it "gives error if action create_if_missing and type key missing in values hash" do
```
btm: ```it "raises an error...```

Aliasgar16:
@btm I have fixed the review comments, squashed the commits and signed-off.

Issue #2871, opened 2015-02-06
Chef 12 Attribute Regression


jonlives:
## Summary
Chef 12 has introduced behaviour which alters the handling of attributes which contain arrays of hashes such that they no longer work as they did in Chef 11

## To Replicate

To replicate the observed regressing behaviour, the following spec should be added to *spec/unit/node_spec.rb* on **master**:

```ruby
      it "correctly fetches string and symbol attribute values from an array of hashes" do
        node.default['apache']['vhosts']     = [{'host' => 'test', 'port' => '666'}]
        expect(node.default['apache']['vhosts'].first['port']).to eql('666')
        expect(node.default['apache']['vhosts'].first[:port]).to eql('666')
        expect(node.default[:apache]['vhosts'].first['port']).to eql('666')
        expect(node.default[:apache]['vhosts'].first[:port]).to eql('666')
      end
```
The equivalent (working) behaviour in Chef 11 can be reproduced by adding the following spec to *spec/unit/node_spec.rb* on **11-stable**

```ruby
    it "correctly fetches string and symbol attribute values from an array of hashes" do
      node.default["apache"]["vhosts"] = [{'host' => 'test', 'port' => '666'}]
      node['apache']['vhosts'].first['port'].should == '666'
      node['apache']['vhosts'].first[:port].should == '666'
      node['apache'][:vhosts].first['port'].should == '666'
      node['apache'][:vhosts].first[:port].should == '666'
    end
```
## What I expected to Happen

I expected Chef 12 to behave as Chef 11 did, whereby the keys of the Hash inside the array ```node['apache']['vhosts’]``` should be accessible using either symbol or string values.

## What Actually Happened

Under Chef 12, accessing the ‘port’ key of the Hash inside the array ```node['apache']['vhosts’]``` using a symbol results in the value returned being nil, instead of the expected ‘666'

## The Cause

What I *have* identified so far is that the attribute value in question is a different object type in Chef 12 than it was in Chef 11.

In Chef 11, ```node.default["apache"]["vhosts”]``` was an **ImmutableArray** containing an **ImmutableMash**. This mash meant that of course symbols and strings could be used interchangeably to access Mash keys.

In Chef 12 however, ```node.default["apache"]["vhosts”]``` is now an **AttrArray** containing a plain old **Hash**. This now means that when a string key is requested using a symbol, it does not exist and returns nil.

[...]

PR #5360, opened 2016-09-21
Chef 12 Attribute Regression

### LINKED ISSUE #4368: Coercing attributes inside an array

gbagnoli:
## Summary

Chef-12 changes the behaviour of how attributes that are defined as literal arrays are handled, such as they do not longer work as they did in Chef-11

given an attributed defined as an array of hashes

```
node['foo'] = [ {'key' => 'value'} ]
```

in Chef-11 accessing using symbols or strings works seamlessly, while in Chef-12 yields different results, i.e.

```
node['foo'].first['key'] == node['foo'].first[:key]  

```
returns `true` in Chef-11 and `false` in Chef 12, as in the latter `node['foo'].first[:key]` is `nil`

I created [a gist with a really basic cookbook](https://gist.github.com/gbagnoli/7fc007c3f67869faf8081f9f584268b4) that shows the problem -- `result_unpatched.txt` is the rendered template using code from master and `result_patched.txt` contains the same template rendered with the patched code from this PR

## To replicate

Simply add the spec from this PR to master and see it fail.
Add the same spec to `11-stable` branch and it passes :)

## Explanation

`VividMash`, unlike `Mash`, does not recursively convert values inside
arrays, just the array itself.
As such, hashes inside arrays are not converted to *Mash, introducing
a behaviour that differs from Chef 11

This PR fixes `VividMash` so arrays are recursively converted and the
behaviour restored, plus it `immutablize` `merge_all` return value, so to be sure that hashes contained in Arrays are properly immutablized as well, ensuring that non-existing attributes
are **not** automatically vivified (which would make them return an empty `VividMash`
instead of `nil`)

This problem was first reported in chef/chef#2871.

Signed-off-by: Giacomo Bagnoli <gbagnoli@gmail.com>

closes #4368

### DIFF
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -452,9 +452,7 @@ class Chef
        #
 
       def merged_attributes(*path)
-        # immutablize(
-        merge_all(path)
-        # )
+        immutablize(merge_all(path))
       end
 
       def combined_override(*path)
--- a/lib/chef/node/attribute_collections.rb
+++ b/lib/chef/node/attribute_collections.rb
@@ -73,6 +73,7 @@ class Chef
       def initialize(root, data)
         @root = root
         super(data)
+        map! { |e| convert_value(e) }
       end
 
       # For elements like Fixnums, true, nil...
@@ -84,6 +85,23 @@ class Chef
 
       def dup
         Array.new(map { |e| safe_dup(e) })
+      end
+
+      private
+
+      def convert_value(value)
+        case value
+        when VividMash
+          value
+        when AttrArray
+          value
+        when Hash
+          VividMash.new(root, value)
+        when Array
+          AttrArray.new(root, value)
+        else
+          value
+        end
       end
 
     end
@@ -184,10 +202,12 @@ class Chef
         case value
         when VividMash
           value
+        when AttrArray
+          value
         when Hash
           VividMash.new(root, value)
         when Array
-          AttrArray.new(root, value.map { |e| convert_value(e) })
+          AttrArray.new(root, value)
         else
           value
         end
--- a/lib/chef/node/attribute_collections.rb
+++ b/lib/chef/node/attribute_collections.rb
@@ -187,7 +187,7 @@ class Chef
         when Hash
           VividMash.new(root, value)
         when Array
-          AttrArray.new(root, value)
+          AttrArray.new(root, value.map { |e| convert_value(e) })
         else
           value
         end
--- a/spec/unit/node_spec.rb
+++ b/spec/unit/node_spec.rb
@@ -784,6 +784,50 @@ describe Chef::Node do
         expect(node["passenger"]["root_path_2"]).to eql("passenger-4.0.57")
         expect(node[:passenger]["root_path_2"]).to eql("passenger-4.0.57")
       end
+
+      it "should deep merge array attributes defined as literals" do
+        node.default["attr_literal_hash"] = { "key" => { "inner" => "value" } }
+        expect(node[:attr_literal_hash][:key]).not_to be_nil
+        expect(node[:attr_literal_hash][:key]).to be_a_kind_of(Mash)
+        expect(node[:attr_literal_hash]["key"]).not_to be_nil
+        expect(node["attr_literal_hash"]["key"]).not_to be_nil
+        expect(node["attr_literal_hash"][:key]).not_to be_nil
+        expect(node[:attr_literal_hash][:key][:inner]).to eql("value")
+        expect(node[:attr_literal_hash]["key"][:inner]).to eql("value")
+        expect(node[:attr_literal_hash]["key"]["inner"]).to eql("value")
+        expect(node[:attr_literal_hash][:key]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key][:inner]).to eql("value")
+        expect(node["attr_literal_hash"]["key"][:inner]).to eql("value")
+        expect(node["attr_literal_hash"]["key"]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key]["not_existing"]).to be_nil
+
+        node.default["attr_literal_array"] = [{ "key" => { "inner" => "value" } }]
+        expect(node[:attr_literal_array]).to be_a_kind_of(Array)
+        expect(node[:attr_literal_array].first[:key]).not_to be_nil
+        expect(node[:attr_literal_array].first["key"]).not_to be_nil
+        expect(node["attr_literal_array"].first["key"]).not_to be_nil
+        expect(node["attr_literal_array"].first[:key]).not_to be_nil
+        expect(node[:attr_literal_array].first[:key][:inner]).to eql("value")
+        expect(node[:attr_literal_array].first["key"][:inner]).to eql("value")
+        expect(node[:attr_literal_array].first["key"]["inner"]).to eql("value")
+        expect(node[:attr_literal_array].first[:key]["inner"]).to eql("value")
+        expect(node["attr_literal_array"].first[:key][:inner]).to eql("value")
+        expect(node["attr_literal_array"].first["key"][:inner]).to eql("value")
+        expect(node["attr_literal_array"].first["key"]["inner"]).to eql("value")
+        expect(node["attr_literal_array"].first[:key]["inner"]).to eql("value")
+        expect(node["attr_literal_array"].first[:key]["not_existing"]).to be_nil
+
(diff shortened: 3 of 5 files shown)


lib/chef/node/attribute.rb
```diff
@@ -452,9 +452,7 @@ def unlink!(level, *path)
        #
 
       def merged_attributes(*path)
-        # immutablize(
-        merge_all(path)
-        # )
+        immutablize(merge_all(path))
```
lamont-granquist: 👍  seems i commented out some code in debugging and forgot to uncomment?  the fact that nothing blew up in tests around this is mildly disturbing.

lib/chef/node/attribute_collections.rb
```diff
@@ -187,7 +187,7 @@ def convert_value(value)
         when Hash
           VividMash.new(root, value)
         when Array
-          AttrArray.new(root, value)
+          AttrArray.new(root, value.map { |e| convert_value(e) })
```
lamont-granquist: this seems like its being patched in the wrong place and that AttrArray should be calling convert_value in the constructor.   in fact, i'm really surprised this changed between 11 and 12.   pretty sure this is the wrong way to patch this and that trying to fix it more correctly should surface what really went wrong here.

gbagnoli: Perfect, I'll move the code to the AttrArray constructor :)

coderanger:
Related https://github.com/chef/chef/issues/4368

lamont-granquist:
i'm still curious about how this broke from 11 to 12?

gbagnoli:
well I'm not entirely sure, tbh, looking at the code it seems like that `merged_attributes` [is using a Mash to merge into](https://github.com/chef/chef/blob/11-stable/lib/chef/node/attribute.rb#L317) and the Mash object [recursively converting values](https://github.com/chef/chef/blob/11-stable/lib/chef/mash.rb#L220) similarly to my first patch attempt.
The mash is then [immutablize](https://github.com/chef/chef/blob/11-stable/lib/chef/node/attribute.rb#L320)d, converting Arrays to [ImmutableArray](https://github.c/chef/chef/blob/11-stable/lib/chef/node/immutable_collections.rb#L11) which [recurse in the constructor](https://github.com/chef/chef/blob/11-stable/lib/chef/node/immutable_collections.rb#L70-L72).
Seems like this code path has changed significantly between the two major versions (but I bet you already knew :P)

anyway, I do have a [branch with the spec added to chef-11](https://github.com/gbagnoli/chef/tree/11-stable-literals-ok) and the tests are passing, if someone wants to double check.

lamont-granquist:
confirmed that you get an array wrapping a Mash:

```ruby
10:06 coredump oc/chef3 % pry
[1] pry(main)> require "chef/mash"
=> true
[2] pry(main)> h = Chef::Mash.new
(pry):2: warning: toplevel constant Mash referenced by Chef::Mash
=> {}
[3] pry(main)> h[:foo] = [ { bar: true } ]
=> [{:bar=>true}]
[4] pry(main)> h[:foo].class
=> Array
[5] pry(main)> h[:foo][0].class
=> Mash
[6] pry(main)>
```

lamont-granquist:
so i checked the code out at pre-chef-12 commit of 68f99fc and found that VividMash is still buggy:

```ruby
[2] pry(main)> require 'chef/mash'
=> true
[3] pry(main)> require 'chef/node/attribute_collections'
/Users/lamont/oc/chef3/lib/chef/node/attribute_collections.rb:28: warning: already initialized constant Chef::Node::AttrArray::MUTATOR_METHODS
/Users/lamont/oc/chef3/lib/chef/node/attribute_collections.rb:28: warning: previous definition of MUTATOR_METHODS was here
=> true
[4] pry(main)> h = Chef::Node::VividMash.new(Chef::Node::Attribute.new({},{},{},{}))
NameError: uninitialized constant Chef::Node::Attribute
from (pry):4:in `__pry__'
[5] pry(main)> require 'chef/node/attribute'
=> true
[6] pry(main)> h = Chef::Node::VividMash.new(Chef::Node::Attribute.new({},{},{},{}))
=> {}
[7] pry(main)> h[:foo] = [ { bar: true } ]
=> [{:bar=>true}]
[8] pry(main)> h.class
=> Chef::Node::VividMash
[9] pry(main)> h[:foo].class
=> Chef::Node::AttrArray
[10] pry(main)> h[:foo][0].class
=> Hash
[11] pry(main)>
```

spec/unit/node_spec.rb
```diff
@@ -784,6 +784,50 @@
...
+        expect(node[:attr_literal_hash]["key"]["inner"]).to eql("value")
+        expect(node[:attr_literal_hash][:key]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key][:inner]).to eql("value")
+        expect(node["attr_literal_hash"]["key"][:inner]).to eql("value")
+        expect(node["attr_literal_hash"]["key"]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key]["inner"]).to eql("value")
+        expect(node["attr_literal_hash"][:key]["not_existing"]).to be_nil
+
+        node.default["attr_literal_array"] = [{ "key" => { "inner" => "value" } }]
+        expect(node[:attr_literal_array]).to be_a_kind_of(Array)
```
lamont-granquist: ^ this still seems to be wrong, it should be an ImmutableArray when you get it through the merged `node[:foo]` accessor and should be an AttrArray when going through the `node.default[:foo]` accessor.

lamont-granquist:
yeah vividmashes still fail in chef-11 but going in through the immutablemash seems to not fail, its kind of weird.

gbagnoli:
@lamont-granquist please let me know if you need anything else I could provide (either info or code or changes).
I think with the new features you should be able to push commits to this PR as I've set the `allow edits from maintainers` checkbox when opening the PR, in case it's needed.

This is fairly important for us as it stopped the upgrade to chef-12, since our cookbooks have this kind of access pattern (mixed strings and symbols) to hashes inside arrays. 

Thanks!

lamont-granquist:
sweet!  yeah, i went through this and convinced myself that your code is correct (i had to go down a few dead ends myself before i wound up with basically the same code), but i think i want to pull the tests i just wrote instead -- and i was just gonna ask if you could click that sweet new button =)

spec/unit/node_spec.rb
lamont-granquist: oh i see, kind_of? instead of instance_of? still we can do a little better.... one sec...

gbagnoli:
yes, it should be all set! Thanks again!

lamont-granquist:
still getting this:

```
%     git push -u gbagnoli gbagnoli-array_literals_attributes
Counting objects: 8, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 1.95 KiB | 0 bytes/s, done.
Total 8 (delta 7), reused 0 (delta 0)
remote: Resolving deltas: 100% (7/7), completed with 7 local objects.
To https://github.com/gbagnoli/chef.git
 ! [remote rejected] gbagnoli-array_literals_attributes -> gbagnoli-array_literals_attributes (permission denied)
error: failed to push some refs to 'https://github.com/gbagnoli/chef.git'
```

lamont-granquist:
oh i see have to use `git push gbagnoli HEAD:array_literals_attributes` and it even told me that, i just didn't comprehend why...

gbagnoli:
I can totally get you write access to my fork if needed :)

lamont-granquist:
fixed it, just waiting on travis to hopefully go green now...

lamont-granquist:
(i still don't understand how this ever worked in chef-11 because vividmashes were definitely broken then -- but this PR is certainly correct and has the tests to prove it...)

gbagnoli:
:pray:

lamont-granquist:
@chef/client-core could use another thumb

danielsdeleo:
👍

lamont-granquist:
@gbagnoli so you missed the release window and unless we ship another regression and have to do yet another 12.14.x fix this will only be accessible in the current channel (assuming jenkins doesn't barf on the nightlies, which is a pretty big assumption some weeks).  next release will probably be 2nd wednesday-ish of next month give or take.

gbagnoli:
@lamont-granquist got it, thanks!

PR #5367, opened 2016-09-21
fix chef-solo spec timeouts

lamont-granquist:
caused by a race condition (not actually fixed) where the pid
can be read by the process which lost the race to lock it before the
pid has been written by the process which got the lock.

the result is a message like:

"WARN: Chef client  is running, will wait for it to finish and then run."

without any numerical pid.

i'm not super concerned with fixing the race, this will unbreak a lot
of red travis failures though.


lamont-granquist:
i'm running this in a loop 100 times on my workstation, so far no hangs/failures

lamont-granquist:
@chef/client-core need a thumb from someone

mwrock:
👍

Issue #4534, opened 2016-02-05
DSC Failure incorrectly reported as WARN


webtonize:
I'm all new to Chef and such, so it could be something I still have to learn about, but I feel i ran in to something unusual with Chef this weekend.

I intentionally setup a dsc_script resource to fail, but the webinterface just reported everything went fine and there are no errors to report. The log on the chef-client gives a warning

the resource:
dsc_script 'MustThrowError' do
code <<-EOH
WindowsFeature InstallIISConsole
{
Name = "MustThrowError"
Ensure = "Present"
}
EOH
end

And the chef-client log:

[2016-02-01T18:05:15+01:00] WARN: Received error while testing configuration:
PowerShell DSC resource MSFT_RoleResource failed to execute Test-TargetResourc e functionality with error message: The requested feature MustThrowError is not found on the target machine. + CategoryInfo : InvalidOperation: (:slightly_smiling: [], CimException + FullyQualifiedErrorId : ProviderOperationExecutionFailure + PSComputerName : . The SendConfigurationApply function did not succeed. + CategoryInfo : NotSpecified: (root/Microsoft/...gurationManager :String) [], CimException + FullyQualifiedErrorId : MI RESULT 1 + PSComputerName : .

Is there some kind of configuration change I need to make to the server or client here to get this "warning", which to me seems pretty fatal enough to get some higher classification, show up in the Web UI and command-line tools?

lamont-granquist:
This looks more like a bug where a warn should be a fatal, rather than getting the warn to show up in the server

btm:
Fixed by https://github.com/chef/chef/pull/5377.

PR #5377, opened 2016-09-23
Invalid `dsc_scripts` should fail the run

NimishaS:
w.r.t https://github.com/chef/chef/issues/4534

### DIFF
--- a/spec/unit/util/dsc/local_configuration_manager_spec.rb
+++ b/spec/unit/util/dsc/local_configuration_manager_spec.rb
@@ -59,10 +59,6 @@ EOH
         allow(lcm).to receive(:run_configuration_cmdlet).and_return(lcm_status)
       end
       context "that returns successfully" do
-        before(:each) do
-          allow(lcm).to receive(:run_configuration_cmdlet).and_return(lcm_status)
-        end
-
         let(:lcm_standard_output) { normal_lcm_output }
         let(:lcm_standard_error) { nil }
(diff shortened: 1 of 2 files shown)


NimishaS:
The warning specified in https://github.com/chef/chef/issues/4534 is being raised from this line: https://github.com/chef/chef/blob/master/lib/chef/util/dsc/local_configuration_manager.rb#L107
But we can't raise this as an error here as this method is returning an array which is expected by some methods.
Exception should be caught here: https://github.com/chef/chef/blob/master/lib/chef/provider/dsc_script.rb#L101, but the previous line is not raising any exception even though the PS command fails because `-erroraction` was `continue`

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -59,10 +59,6 @@
         allow(lcm).to receive(:run_configuration_cmdlet).and_return(lcm_status)
       end
       context "that returns successfully" do
-        before(:each) do
-          allow(lcm).to receive(:run_configuration_cmdlet).and_return(lcm_status)
-        end
-
```
NimishaS: Removed this as it's written twice. Please expand the code to check.

mwrock:
I'm 👍 on this but please fix style (might just be a matter of rebasing)

NimishaS:
Rebased and signed off.

PR #5378, opened 2016-09-23
Add Chef::Node __path tracking, change root to __root, drop top_level_breadcrumb hack, fix long-standing Chef 12.0 deep merge cache bugs

lamont-granquist:
Fixes some long-standing horribleness introduced in Chef 12.0 with the top_level_breadcrumb state and confusion in the deep merge cache when that global state is whacked in the middle of an operation.

This modifies some APIs.  I would argue nobody should ever have been using them, they were undocumented, should not have been considered covered by SemVer.  Any affected consumer should modify their code.  Summary of changes:

- added the :__path accessor (can't use :path because of method_missing)
- removed the :root accessor and changed it to __root for consistency
- dropped the :top_level_breadcrumb accessor in favor of the :__path array (this is entirely an implementation detail of the deep merge cache, and its an API that is broken by design)
- changed the initializer of VividMash and ImmutableMash to be consistent and have a sane signature.

The latter change is where there might be a possibility of disruption.  However, the ImmutableMashes are also an implementation detail of the Chef::Node deep merge cache, and I have a hard time imagining how people would be using that standalone.

The VividMash has more potential to be used, but this change is actually geared towards improving its usability.  Right now you have to give two arguments to the constructor which is the root object and the initial value.  This relaxes that and rearranges the constructor so that `VividMash.new()` with no arguments is now legal, while `VividMash.new({ foo: "bar" })` initializes the hash.

The rationale for allowing the use of VividMash as a better public API is this particular abomination:

https://github.com/chef-cookbooks/mysql/blob/61cee7a9220854280f823957875d8d1c24db49c8/libraries/helpers.rb#L262-L263

That code wants an autovivifying hash from Chef, so it reuses a Chef::Node object (with attribute precedence levels and deep merge horrors all thrown in) -- this change will make it possible to remove that eventually and simply use a VividMash in that place.  And I think that code is the strongest argument that VividMash is most likely not being used outside of the attributes internals right now.

As far as the attributes fixes goes, this code is still not perfect and has flaws:

The __path state is stored in the node attribute itself.  Handles to node attributes that have been rooted in multiple places in the tree may see their __path state mutate.  That could be partially solved by moving the __path state to a decorator.

The issue of the node being able to be pointed at from multiple points in the attribute tree gives a consistency problem, particularly for the deep_merge_cache that may not be possible to fix.  It may always be possible to 'trick' the reporting of what path is being updated when a node is assigned to.  Since this is used for the deep merge cache that affects correctness as well.  Its unclear how severe the impact will be (the current examples of abusing the top_level_breadcrumb will be fixed by this change, this would be even more perverse use of attributes).  We might have to consider always dup'ing as data is assigned to the node (which we do sometimes when we assign a Hash or Array and have to run through convert_value() and dup the whole object to convert it).  We might also consider doing away with the full deep merge cache, and doing deep-merge-as-we-go in order to gain correctness -- and then simply cache often-used values like platform/platform_family/etc -- possibly any non-container leaf value or only automatic/ohai attributes.

So, this is "better", its not "perfect" (i'm fairly certain that chef-attributes-that-behave-exactly-like-hashes is literally impossible to implement now without starting to drop features).  Hopefully this is "good enough" as far as the deep merge cache bugfixing goes.

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -197,7 +197,6 @@ class Chef
     # Set a normal attribute of this node, but auto-vivify any Mashes that
     # might be missing
     def normal
-      attributes.top_level_breadcrumb = nil
       attributes.normal
     end
 
@@ -209,14 +208,12 @@ class Chef
     # Set a default of this node, but auto-vivify any Mashes that might
     # be missing
     def default
-      attributes.top_level_breadcrumb = nil
       attributes.default
     end
 
     # Set an override attribute of this node, but auto-vivify any Mashes that
     # might be missing
     def override
-      attributes.top_level_breadcrumb = nil
       attributes.override
     end
 
@@ -237,7 +234,6 @@ class Chef
     end
 
     def automatic_attrs
-      attributes.top_level_breadcrumb = nil
       attributes.automatic
     end
 
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -17,6 +17,7 @@
 # limitations under the License.
 #
 
+require "chef/node/mixin/deep_merge_cache"
 require "chef/node/mixin/immutablize_hash"
 require "chef/node/mixin/state_tracking"
 require "chef/node/immutable_collections"
@@ -36,8 +37,17 @@ class Chef
     class Attribute < Mash
 
       include Immutablize
-
+      # FIXME:  what is include Enumerable doing up here, when down below we delegate
+      # most of the Enumerable/Hash things to the underlying merged ImmutableHash.  That
+      # is, in fact, the correct, thing to do, while including Enumerable to try to create
+      # a hash-like API gets lots of things wrong because of the difference between the
+      # Hash `each do |key, value|` vs the Array-like `each do |value|` API that Enumerable
+      # expects.  This include should probably be deleted?
       include Enumerable
+
+      include Chef::Node::Mixin::DeepMergeCache
+      include Chef::Node::Mixin::StateTracking
+      include Chef::Node::Mixin::ImmutablizeHash
 
       # List of the component attribute hashes, in order of precedence, low to
       # high.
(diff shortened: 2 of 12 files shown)


lamont-granquist:
🚧  obviously needs specs, and now a rebase since i ninja'd myself...

lamont-granquist:
There's a curious issue here with the __path being stored off of every node in the "attributes tree" and ways to attack this by moving nodes around in the tree, or inserting them into two different places.

The basic idea is that you could take one node, move it, that would invalidate the tree, but the node wouldn't know its path had changed, then access the top level of the attributes tree so that the deep merge cache was updated (it got invalidated by the write), then update the node that you still had the handle onto.

Its also fairly trivial to see that if there's a pointer to a node in two different locations in the 'tree' that the __path can only point to one of them, and on every access through the two different paths to the node that __path value will thrash (and writes will only invalidate one of the deep merge caches, which is perhaps the worst side-effect).  At least we get reads correct in that case though by updating the __path on the node on every access through `#[]=`

We could try to solve this by deep-dup'ing subnodes as they're assigned as values through `#[]=` and then walk through the dup'd tree and 'fix' the __path values for the new location.

However, that means that if you re-use the handle you'll be using the original version and not what you just inserted into the tree which is kind of unusual:

```
a = node["foo"]["bar"]
node["baz"]["bar"] = a
a.__path = [ "foo", "bar" ]
a.object_id = 0x123
node["foo"]["bar"].object_id = 0x123
node["baz"]["bar"].object_id = 0x456
a["buzz"] = "fizz"
node["baz"]["bar"]["buzz"] => nil
node["foo"]["bar"]["buzz"] => "fizz"
```

That seems like worse behavior than the behavior it is trying to fix.

lamont-granquist:
As an aside this whole thing made me think fully through going down the road of not using Mashes in Chef at all and the implications to `def []=(key, value)` and how we call convert_value() and dup() the argument (when its not already a Mash-like thing in the value).  That result would be very un-delightful though because we can do arbitrarily deep hash key assignment:

```
node['foo'] = { foo: { bar: { baz: true } } }
```

We'd need to do something both computationally as expensive and even more annoying of walking through the value and yelling at the user if they used symbols instead of strings anywhere.  Or else we'd wind up storing it internally in different formats which would be brittle if some code decided to use strings while some code decided to use symbols.  Then it gets worse when you think about keys doing a
round-trip through JSON where they lose any identity over if they're strings or symbols and we just have to pick one or the other to convert from json and hope the user is using that.

The net result of all this thought is that if we had to do it all over again, we'd probably wind up using Mashes again.

The alternative would be to block any use of assignment like that and constrain the API and make it a bit clearer that the object you're building isn't really a Hash, and have to do something like:

```
node.set('foo', NodeThingy.new( { foo: { bar: { baz: true } } } ) )
```

But effectively you're still walking through the deep data structure and converting it and dup'ing it for the user.  Mashes do the same thing, only with quite  lot simpler of an API.

So, I'm thinking @adamhjk got it right those many years ago using a Mash.

When it comes down to the edge conditions in this PR related to how the path can get confused by nodes being rooted in multiple locations, we should probably move towards API consistency such that `node#[]=(key, value)` would always dup the value so that nodes would have precisely one location where they were rooted in the tree.

lamont-granquist:
<sup><sup>although a strongly typed language with coercions would probably be better...</sup></sup>

lamont-granquist:
@chef/client-core i think this is decent enough to ship.  it (ab)uses `prepend` because i've only half-extracted the behaviors of node attributes to mixins.  if the code that was in the classes were extracted into a mixin then it could be included before the mixins that i'm having to `prepend` here and then it would look things up the right order.  but doing all that work would turn an already heavy PR into a complete rewrite of node attributes.

PR #5098, opened 2016-07-08
Fixes GH-4955, allowing local gems with remote dependencies

jyaworski:
Hi all. This should fix #4955.

Issue #4955, opened 2016-05-20
Issue with new behavior of Chef Gem in 12.10.24-1 


jsywulak:
## Description

The behavior of `chef_gem` was changed when installing a local gem in #4847. If you're installing a local gem with remote dependencies, the installation fails because it only looks for those dependencies locally.

## Chef Version

12.10.24-1 

## Platform Version

Ubuntu 14.04.3 LTS

## Replication Case

You'll need to [build a gem](http://guides.rubygems.org/make-your-own-gem/) with some remote dependencies, sample `Gemfile`:

```
source 'https://rubygems.org'

gemspec

group :test do
  gem 'activesupport'
```

You'll need to build that gem and have it exist locally on your filesystem (`gem build`).

Then you need a cookbook that tries to install that gem.

```
chef_gem 'common-utils' do
  compile_time false	
  source '/tmp/gems/common-utils.gem'
  action :install
end
```

## Client Output

Client will error out complaining it can't find one of the gems (my experience yields the first one in alphabetical order, but if you have a more complicated set of dependencies that may not be true for everyone).

## Stacktrace

```
1463702555,,ui,message,    amazon-ebs: [2016-05-20T00:02:40+00:00] DEBUG: chef_gem[common-utils] common-utils not installed%!(PACKER_COMMA) installing 0.0.1
1463702555,,ui,message,    amazon-ebs:
1463702555,,ui,message,    amazon-ebs: ================================================================================
1463702555,,ui,message,    amazon-ebs: Error executing action `install` on resource 'chef_gem[common-utils]'
1463702555,,ui,message,    amazon-ebs: ================================================================================
1463702555,,ui,message,    amazon-ebs:
1463702555,,ui,message,    amazon-ebs: Mixlib::ShellOut::ShellCommandFailed
1463702555,,ui,message,    amazon-ebs: ------------------------------------
1463702555,,ui,message,    amazon-ebs: Expected process to exit with [0]%!(PACKER_COMMA) but received '2'
1463702555,,ui,message,    amazon-ebs: ---- Begin output of /opt/chef/embedded/bin/gem install /tmp/gems/common-utils.gem -q --no-rdoc --no-ri -v "0.0.1" --local ----
1463702555,,ui,message,    amazon-ebs: STDOUT:
1463702555,,ui,message,    amazon-ebs: STDERR: ERROR:  Could not find a valid gem 'activesupport' (>= 0) in any repository
[...]

PR #5379, opened 2016-09-24
Revert --local filter for gems installed from  paths

mwrock:
pr #5098 was intended tom revert #4847 however that revert left behind the addition of the `--local` argument when installing gems from a local path. So from what I can tell, the original bug reported in #4955 remains. If I try to install a gem locally that has remote dependencies, the converge fails.

This essentially completes the revert and allows remote dependencies to to be installed.


moos3:
Any chance of this getting put out in the wild soon?

jyaworski:
Yes, this was mistakenly left out of the original PR. Please merge this ASAP.

Issue #5049, opened 2016-06-22
powershell_out with a code block requires double escape for double quotes


cordovinian:
## Description

If I have a script block defined in a library and then pass that to powershell_out only double-escaped double quotes will end up as valid PowerShell syntax. Since the double quotes are missing PowerShell will fail trying to run the code that is no longer in a string.

I think the issue is in the powershell_out mixin where the script block is passed into a command to run PowerShell.

[powershell_out.rb](https://github.com/chef/chef/blob/master/lib/chef/mixin/powershell_out.rb#L94)

_Note_: This is happening inside of a LWRP
## Chef Version

``` powershell
> chef -v
Chef Development Kit Version: 0.14.25
chef-client version: 12.10.24
berks version: 4.3.3
kitchen version: 1.8.0
```
## Platform Version

Windows 10 latest

``` powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.0.10586.122
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.10586.122
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```
## Replication Case

``` ruby
   def correct_version_running?
      script_block = <<-EOF
      $myvar= 'variable'
      Write-Verbose 'Only single escape needed for other characters: (\\, \#)' -Verbose
      Write-Verbose \\"String with $myvar interpolation\\" -Verbose
      Write-Verbose \"String with $myvar interpolation\" -Verbose
      Write-Verbose "String with $myvar interpolation" -Verbose
      EOF

      ::Chef::Log.debug("Script block:\n#{script_block}")
      result = powershell_out(script_block)
   end
```
## Client Output

``` cmd
    Mixlib::ShellOut::ShellCommandFailed
    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    Expected process to exit with [0], but received '1'
    ---- Begin output of powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "      $myvar= 'variable'
          Write-Verbose 'Only single escape needed for other characters: (\, #)' -Verbose
          Write-Verbose \"String with $myvar interpolation\" -Verbose
          Write-Verbose "String with $myvar interpolation" -Verbose
          Write-Verbose "String with $myvar interpolation" -Verbose
    " ----
[...]

PR #5381, opened 2016-09-26
Handling double quotes in powershell_out

NimishaS:
w.r.t. https://github.com/chef/chef/issues/5049

### DIFF
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -34,7 +34,14 @@ class Chef
         script = command_args.first
         options = command_args.last.is_a?(Hash) ? command_args.last : nil
 
-        run_command_with_os_architecture(script, options)
+        result = run_command_with_os_architecture(script, options)
+
+        if result.stderr.include? "A positional parameter cannot be found that accepts argument"
+          raise "Please use single escape for special characters in powershell_out."
+        elsif result.stderr != ""
+          raise result.stderr
+        end
+        result
       end
 
       # Run a command under powershell with the same API as shell_out!
@@ -91,7 +98,8 @@ class Chef
           "-InputFormat None",
         ]
 
-        "powershell.exe #{flags.join(' ')} -Command \"#{script}\""
+        # without gsub user has to add double escape for double quotes in the recipe
+        "powershell.exe #{flags.join(' ')} -Command \"#{script.gsub('"', '\"')}\""
       end
     end
   end
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -37,9 +37,9 @@ class Chef
         result = run_command_with_os_architecture(script, options)
 
         if result.stderr.include? "A positional parameter cannot be found that accepts argument"
-          raise "Please use single escape for special characters in powershell_out."
+          raise Mixlib::ShellOut::ShellCommandFailed, "Please use single escape for special characters in powershell_out. \n #{result.stderr}"
         elsif result.stderr != ""
-          raise result.stderr
+          raise Mixlib::ShellOut::ShellCommandFailed, result.stderr
         end
         result
       end
--- a/spec/unit/mixin/powershell_out_spec.rb
+++ b/spec/unit/mixin/powershell_out_spec.rb
@@ -18,7 +18,7 @@
 require "spec_helper"
 require "chef/mixin/powershell_out"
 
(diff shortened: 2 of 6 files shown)


NimishaS:
Need to add specs.

spec/unit/mixin/powershell_out_spec.rb
```diff
@@ -44,27 +44,21 @@
...
-  describe "#powershell_out!" do
-    it "runs a command and returns the shell_out object" do
-      mixlib_shellout = double("Mixlib::ShellOut")
-      expect(object).to receive(:shell_out).with(
-        "powershell.exe #{flags} -Command \"Get-Process\"",
-        {}
-      ).and_return(mixlib_shellout)
-      expect(mixlib_shellout).to receive(:error!)
-      expect(object.powershell_out!("Get-Process")).to eql(mixlib_shellout)
-    end
```
NimishaS: Whole `describe "#powershell_out!"` block was defined twice. Hence deleted it.

spec/unit/mixin/powershell_out_spec.rb
```diff
@@ -44,27 +44,21 @@
...
-      mixlib_shellout = double("Mixlib::ShellOut")
-      expect(object).to receive(:shell_out).with(
-        "powershell.exe #{flags} -Command \"Get-Process\"",
-        timeout: 600
-      ).and_return(mixlib_shellout)
-      expect(mixlib_shellout).to receive(:error!)
-      expect(object.powershell_out!("Get-Process", timeout: 600)).to eql(mixlib_shellout)
+      it "raises error if double quote is passed with double escape characters", :windows_only do
+        res = object.powershell_out("Write-Verbose \\\"Some String\\\" -Verbose")
+        expect(res.stderr).to include("A positional parameter cannot be found that accepts argument")
```
btm: Can we catch this error and raise one that says you're probably using double escaped double quotes?

NimishaS: Raising error by reading `stderr` of the `shell_out` object. While fixing the specs I found that it is an expected behavior that `powershell_out` should not raise errors even for incorrect input. `powershell_out!` should raise.
According to current changes, error will be raised for `powershell_out` method too.

lib/chef/mixin/powershell_out.rb
```diff
@@ -34,7 +34,14 @@ def powershell_out(*command_args)
         script = command_args.first
         options = command_args.last.is_a?(Hash) ? command_args.last : nil
 
-        run_command_with_os_architecture(script, options)
+        result = run_command_with_os_architecture(script, options)
+
+        if result.stderr.include? "A positional parameter cannot be found that accepts argument"
+          raise "Please use single escape for special characters in powershell_out."
```
btm: We should raise ArgumentError or something more specific from `lib/chef/exceptions.rb`. Is there some argument we could include in this message to be more specific about where the error was. Without looking, maybe `script` or `options`?

lib/chef/mixin/powershell_out.rb
```diff
@@ -34,7 +34,14 @@ def powershell_out(*command_args)
...
+        elsif result.stderr != ""
+          raise result.stderr
```
btm: since this in generic, raise `Mixlib::ShellOut::ShellCommandFailed` ?

lib/chef/mixin/powershell_out.rb
NimishaS: Appended `stderr` object returned by shell_out to the raised error. `stderr` shows the specific area where the code fails. An example of raised error by `stderr` is shown below:
```
[2016-10-03T09:13:45+00:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: Please use single escape for special characters in powershell_out.
 Write-Verbose : A positional parameter cannot be found that accepts argument 'String\'.
At line:1 char:1
+ Write-Verbose \Some String\ -Verbose
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Write-Verbose], ParameterBindingException
    + FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.WriteVerboseCommand
```

lib/chef/mixin/powershell_out.rb
NimishaS: Fixed

btm:
@chef/client-windows review?

PR #5390, opened 2016-09-28
bumping chef minor version

mwrock:


tas50:
We should do this now

PR #5393, opened 2016-09-28
Add a default GID for user creation on MacOS

cgerke:
new_resource.gid.to_s will be empty if a recipe calls the user resource without gid and changes the home property. Will result in an exception and if chef is run by root, the home dir will end up being owned by root from the mv.

### DIFF
--- a/lib/chef/provider/user/dscl.rb
+++ b/lib/chef/provider/user/dscl.rb
@@ -51,6 +51,11 @@ class Chef
         provides :dscl_user
         provides :user, os: "darwin"
 
+        # Just-in-case a recipe calls the user dscl provider without specifying
+        # a gid property. Avoids chown issues in move_home when the manage_home
+        # property is in use. #5393
+        STAFF_GROUP_ID = 20
+
         def define_resource_requirements
           super
 
@@ -264,12 +269,12 @@ user password using shadow hash.")
         #
         # Sets the group id for the user using dscl. Fails if a group doesn't
         # exist on the system with given group id. If `gid` is not specified, it
-        # sets a default Mac user group "staff", with id 20.
+        # sets a default Mac user group "staff", with id 20 using the CONSTANT
         #
         def dscl_set_gid
           if new_resource.gid.nil?
             # XXX: mutates the new resource
-            new_resource.gid(20)
+            new_resource.gid(STAFF_GROUP_ID)
           elsif !new_resource.gid.to_s.match(/^\d+$/)
             begin
               possible_gid = run_dscl("read /Groups/#{new_resource.gid} PrimaryGroupID").split(" ").last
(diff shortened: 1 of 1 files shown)


lib/chef/provider/user/dscl.rb
```diff
@@ -329,7 +329,7 @@ def ditto_home
 
         def move_home
           Chef::Log.debug("#{new_resource} moving #{self} home from #{current_resource.home} to #{new_resource.home}")
-
+          new_resource.gid(20) if new_resource.gid.nil?
```
thommay: I see on my MacOS box that 20 is staff; is this a reasonable default everywhere? And, more importantly, do we document what 20 is anywhere?

cgerke: For a Mac system imho 20 is a reasonable default, because if anyone is creating a "special" system user or admin user they would most likely be explicit and set the gid.

This is certainly an edge case but my point is, the user resource allows anyone to create a user with his being required even when a managed home is in use.

You are already setting it to 20 if gid is nil in the dscl_set_gid function.

Chef doesn't document it anywhere I can see.

I think an alternative to my PR would be to raise an exception if gid is not specified, ie. make it a requirement in the user resource.

thommay: Yeah, I thought we were already using it somewhere. I'm 👍 either way now, but it'd be really nice if that were a constant like STAFF_GROUP_ID or something.

cgerke: Agreed.

cgerke: Would be really nice if the user resource natively supported the osx IsHidden property too.

thommay: To be clear, can you add the constant and make `move_home` and `dscl_set_gid` use it, please. Once you're done that I'll merge, thanks.

cgerke:
Better?

cgerke:
How can I sign DCO after the fact?

thommay:
@cgerke https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco has the details. Since you have multiple commits I'd run `git rebase -i` on your branch and add the Signed-off-by line.

coderanger:
Or `chef gem install dco && dco sign`

cgerke:
I've royally screwed up my repo now, shall I just redo another PR?

cgerke:
Think I need some help with this.

thommay:
@cgerke can you state the commands you ran so we can write better instructions?

 I can just take the commits from this branch and merge them without difficulty.

cgerke:
Yeah will do. Let me sit down in the morning with a clear head and I will run through it properly (almost midnight here).

lamont-granquist:
that last merge commit you've got probably isn't helping any, you can `git reset HEAD~1 --hard` and force push and that might help (make sure `git log` has your two commits on top before doing the force push)

lamont-granquist:
and instead of merge committing to try to pull in master you can just do this on your feature branch (after nuking the merge commit):

```ruby
git checkout patch-1
git fetch origin
git rebase origin/master
# fix up any rebase conflicts here
git log  # make sure you've still got your two commits
git push -f cgerke patch-1
```

cgerke:
Thanks mate, these steps actually was what I did originally. I did it in the following order though.

(I have aliases setup).

1. I noted my master was behind by 67 commits, before I do any work on any forked repo I always sync master to upstream. So I did the following on my master.

	sync-master-to-origin = !git fetch origin && git merge origin/master

2. I then switched to patch-1 branch and this is where it went wrong i think because I did the same step on my branch

	sync-master-to-origin = !git fetch origin && git merge origin/master

I've just followed your steps and I think its good now.

cgerke:
ok this is done now and I seem to be back to my original issue for the DCO.

cgerke:
@thommay thanks. @lamont-granquist thanks for the git-fu

Issue #5274, opened 2016-09-02
package resource not uninstalling MSI without `source` property defined


jmery:
## Description

This works properly and uninstalls the package:
```
package 'ActivePerl Enterprise 5.16.3 Build 1603.4 (64-bit)' do
  action :remove
  installer_type :msi
  source 'https://s3.amazonaws.com/jmery/pepsi-poc/ActivePearl.msi'
  checksum 'e6aca518673e37606990a3db368d401dc860182a3da09a9d77770ba16f24dfe9'
  options '/q'
end
```

This does NOT work.  The client converges and says it's uninstalling, but no action taken.
```
package 'ActivePerl Enterprise 5.16.3 Build 1603.4 (64-bit)' do
  action :remove
  installer_type :msi
#  source 'https://s3.amazonaws.com/jmery/pepsi-poc/ActivePearl.msi'
#  checksum 'e6aca518673e37606990a3db368d401dc860182a3da09a9d77770ba16f24dfe9'
  options '/q'
end
```

It appears that either the `package` resource or the MSI itself are not performing a `/quiet` uninstall.  This was verified with other MSI packages from the command line (e.g. grab the MSI `UninstallString` from the registry and run from command prompt).  Tested with the chef-client install as well with the same results.

## Chef Version

chef-client version  `Chef: 12.13.37`

## Platform Version

Windows 2012 Standard x64 (note:  **Not** 2012 R2)

## Replication Case

Install the package above in the manner of your choice (I used the `package` resource to install).  Run the above `package` resource code to attempt to uninstall.

The package `source` above is wide open for use and the `checksum` is valid.

## Client Output

No errors reported

## Stacktrace

N/A

CC @mwrock

Shafiud:
i don't see any difference in the code between "This works properly" Vs "This does NOT work".

jmery:
`source` and `checksum` are commented on the non-working resource.  Essentially, `source` must be specified for `action :remove` to work.

thommay:
@chef/client-windows please take a look…

Aliasgar16:
[...]

PR #5394, opened 2016-09-28
Newly creating uninstall string using the product code value retrieved from the uninstall_string of the Windows registry to avoid using invalid uninstall_string.

Aliasgar16:
w.r.t. https://github.com/chef/chef/issues/5274
In-progress.


mwrock:
Hmm. It sounds to me like the issue in #5274 is a case of true vendor error and I wonder if we alter/harden the string here, if that could just cause problems with otherwise valid uninstall strings. Might be best just to let this be.

Aliasgar16:
@mwrock I have tested these changes with `30 MSI installers` (most of the installers had `/I` in their `uninstall_strings` while some other MSI installers had `/X` in their `uninstall_strings`) and it worked successfully for each of them.
I have attached the `test_results` document on the [Card-399](https://chefio.atlassian.net/browse/CLOGENG-399) (refer the doc with the latest date appended) which explains the overall behavior.

PR #5395, opened 2016-09-28
package locking for apt, yum and zypper

yeoldegrove:
This introduces package locking for the apt, yum and zypper package providers.

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -218,25 +218,12 @@ class Chef
         end
       end
 
-      def package_locked?
-        if !current_version_array.any?
-          # ! any? means it's all nil's, which means nothing is installed
-          false
-        elsif !new_version_array.any?
-          true # remove any version of all packages
-        elsif have_any_matching_version?
-          true # remove the version we have
-        else
-          false # we don't have the version we want to remove
-        end
-      end
-
       def action_lock
         multipackage_api_adapter(@current_resource.package_name, @new_resource.version) do |name, version|
-          if package_locked(name) == true
+          if package_locked(name, version) == true
             Chef::Log.debug("#{new_resource} is already locked")
           else
-            lock_package(name)
+            lock_package(name, version)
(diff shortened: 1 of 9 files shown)


lib/chef/provider/package.rb
```diff
@@ -218,6 +218,28 @@ def action_reconfig
...
       end
 
+      def action_lock
+        multipackage_api_adapter(@current_resource.package_name, @new_resource.version) do |name, version|
+          if package_locked(name, version) == true
+            Chef::Log.debug("#{new_resource} is already locked")
+          else
+            lock_package(name, version)
+            Chef::Log.info("#{@new_resource} locked")
+          end
```
lamont-granquist: both these actions need to be wrapped by converge_by in order to have output in the formatter and to correctly why-run (and update the resource, and fire notifications).

you should probably follow the pattern in `action_remove` for how to write these methods.

i'm a bit concerned that i followed this pattern for a reason:

```
   if idempotency_check?
    converge_by "doing things to package(s) x, y, z" do
      multipackage_api_adapter(...) do |name, version|
        do_the_thing()
      end
    end
  end
```

the idempotency check certainly has to come around the converge_by() and if you moved the multipackage_api_check outside of that you'd have multiple converge_by()s which might have been what i was trying to avoid when i wrote that pattern?  i'm not certain why, but i fear there was some reason otherwise wrapping the whole thing in the multipackage_api_adapter call would have been much cleaner.

yeoldegrove: Thanks for that, didn't notice it!
Added the converge_by... formatter output and why-run are now also working.
In additon to that I added more tests, added yum and fixed some issues I noticed.

tas50:
@yeoldegrove This looks pretty useful. Thanks for putting this together

yeoldegrove:
Any change this gets merged?

thommay:
@lamont-granquist please to re-review

Issue #5397, opened 2016-09-28
gpgcheck default on new yum_repository


shortdudey123:
## Description

In the yum cookbook, the [gpgcheck attribute](https://github.com/chef-cookbooks/yum/blob/master/resources/repository.rb#L36) defaults to `true`
However, the new [yum_repository resource](https://github.com/chef/chef/blob/master/lib/chef/resource/yum_repository.rb#L38) has no default (i.e. `nil` which causes [repo files](https://github.com/chef/chef/blob/master/lib/chef/provider/support/yum_repo.erb#L29-L33) to change when updating to chef 12.14

## Chef Version

```
$ chef-client --version
Chef: 12.14.89
$ 
```

## Platform Version

CentOS

## Replication Case

* Run the `yum_repository` resource on Chef < 12.14 using the yum cookbook
* Upgrade chef-client to 12.14.x
* Run chef-client
* See change of `gpgcheck` in yum repo file

## Client Output

```
      --- /etc/yum.repos.d/custom-repo.repo	2016-09-15 01:23:50.339576954 +0000
      +++ /etc/yum.repos.d/.custom-repo20160928-12795-vyne84.repo	2016-09-28 23:31:28.177930190 +0000
      @@ -5,6 +5,6 @@
       name=custom repo for Red Hat Enterprise Linux $releasever
       baseurl=https://custom.local/yum/redhat/$releasever/$basearch/latest
       enabled=1
      -gpgcheck=1
      +gpgcheck=0
       gpgkey=https://custom.local/yum/redhat/$releasever/$basearch/latest/CUSTOM-GPG-KEY.pub
```

## Stacktrace

N/A

PR #5398, opened 2016-09-28
Set yum_repository gpgcheck default to true

shortdudey123:
Fixes https://github.com/chef/chef/issues/5397


ranjib:
:+1:

tas50:
Thanks for noticing this. You are correct that there currently isn't any testing of yum_repository. We're looking into getting that into integration testing in travis so we can avoid regressions here.

PR #5407, opened 2016-10-04
Fix centos 6 kitchen tests with chef-client cookbook 6.0.0

mwrock:
Signed-off-by: Matt Wrock <matt@mattwrock.com>


lamont-granquist:
@tas50 isn't this a bug in the cookbook?

tas50:
It all seems to work in Vagrant, but in Docker land things aren't behaving. The init_package attribute isn't getting set. I need to fire up my local docker and figure it out because it seems like an Ohai bug. It should fail back to init, but it could be in the cookbook somewhere as well.

PR #5410, opened 2016-10-05
Add the policyfile revision_id as an automatic attribute.

johnbellone:
/cc @coderanger @danielsdeleo

We have been looking for an easy way to get this value into our error
reporting application and it surprised me that it isn't already
captured. If there's a better way to do this totally open for thoughts!

### DIFF
--- a/lib/chef/policy_builder/policyfile.rb
+++ b/lib/chef/policy_builder/policyfile.rb
@@ -173,6 +173,7 @@ class Chef
         CookbookCacheCleaner.instance.skip_removal = true if named_run_list_requested?
 
         node.run_list(run_list)
+        node.automatic_attrs[:policy_revision] = revision_id
         node.automatic_attrs[:roles] = []
         node.automatic_attrs[:recipes] = run_list_expansion_ish.recipes
         run_list_expansion_ish


danielsdeleo:
This looks fine to me. @lamont-granquist does this cause any future problems with desired state vs. last observed state?

coderanger:
Another option would be to attach some kind of `node.policy` API which can expose more than just the name/group. But this seems fine for now. No more broken with desired vs. current than anything else we do :)

danielsdeleo:
Yeah, I just don't want to dig the hole any deeper, especially if someone was planning to work on fixing it soon.

lamont-granquist:
just based on a quick skim this looks okay, since the revision_id is coming from the policy object and the node isn't changing that.  the node is just publishing an automatic attr announcing its state in its node attributes which seems fine.

of course the issue here is that the node can be hacked and therefore lie about its revision, and other parts of the infrastructure that search on that may trust the revision and it may be possible to do-bad-things -- in which case you'd really want to present a merged facade of the node object and the policy object to users, or else to allow searching based on the policy attributes (`knife policy search`)?  feels like we should probably write those other APIs if we need to better lock down policy information as desired state, but i don't think that warrants blocking this change...

coderanger:
Needs tests and release_notes though.

lamont-granquist:
travis is sad and needs tests + release_notes

PR #5413, opened 2016-10-05
Updated release notes for cab_package resource and provider

Vasu1105:


tas50:
We need to nuke out the old notes that are from 12.14

PR #5418, opened 2016-10-06
add learnchef testing to travis matrix

lamont-granquist:


lamont-granquist:
@chef/client-core gets us some coverage in TK so we at least don't totally break these two learnchef cookbooks

lamont-granquist:
we can go back and actually add some inspec validation in the wrapper cookbook, but this gets us converge testing at the bare minimum.

PR #5419, opened 2016-10-06
Lcg/learnchef tests

lamont-granquist:


lamont-granquist:
just creating a cleaner PR for posterity

PR #5420, opened 2016-10-06
Revert "Handling double quotes in powershell_out"

mwrock:
Reverts chef/chef#5381

I had to revert this because its causing the chocolatey functional tests to fail on Jenkins 2008R2 testers. 

The above failure is partly the fault of something in the chocolatey tests which I will get to in a bit. The key problem with this PR is that it bubbles up exceptions from stderr in `powershell_out` but the bubbling of exceptions should only happen automatically from `powershell_out!`. Any error conditions are meant to be  suppressed in `powershell_out` so that the caller can examine them and act as it sees fit.

So it ends up that on 2008R2, the chocolatey installer fails and that failure causes `powershell_out` to raise an exception where it previously suppressed it. This was not a problem since chocolatey is already installed. We do need to investigate that problem separately. However we need to move the exception logic added by this PR to `powershell_out!`.

### DIFF
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -34,14 +34,7 @@ class Chef
         script = command_args.first
         options = command_args.last.is_a?(Hash) ? command_args.last : nil
 
-        result = run_command_with_os_architecture(script, options)
-
-        if result.stderr.include? "A positional parameter cannot be found that accepts argument"
-          raise Mixlib::ShellOut::ShellCommandFailed, "Please use single escape for special characters in powershell_out. \n #{result.stderr}"
-        elsif result.stderr != ""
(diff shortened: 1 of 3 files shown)


NimishaS:
@mwrock , I had also pointed this out https://github.com/chef/chef/pull/5381#discussion_r81338197, but wasn't aware of it's impact.

mwrock:
@NimishaS my suggestion would be to simmply ommit the inspection of `stdout` and the `raise` alltogether. It seems like the intent was to protect against other edge cases beyond the inclusion of quotes. If we scope this work agreeing that we are only attempting to solve for quotes and not other edge cases, then we can consider the error handling to be addressing a broader concern.

That is certainly a valid concern, but I think its going to be easiest to ger this improvement through if we isolate to just the quote escaping. and then address other edge cases separately.

PR #5422, opened 2016-10-06
Mac: ensure launchd's start_calendar_interval gets valid settings

thommay:
This work is @ErebusBat's, I'm simply finishing it off so we can get it merged.

Closes: #5115

### DIFF
--- a/lib/chef/resource/launchd.rb
+++ b/lib/chef/resource/launchd.rb
@@ -68,7 +68,7 @@ class Chef
 
         # Make sure the hashes don't have any incorrect keys/values
         array.each do |entry|
-          allowed_keys = %w(Minute Hour Day Weekday Month)
+          allowed_keys = %w{Minute Hour Day Weekday Month}
           error_msg = "Invalid key for start_calendar_interval, must be one of: #{allowed_keys.join(", ")}"
           if entry.keys.any? { |key| !allowed_keys.include?(key) }
             raise Chef::Exceptions::ValidationFailed, error_msg
--- a/lib/chef/resource/launchd.rb
+++ b/lib/chef/resource/launchd.rb
@@ -46,6 +46,48 @@ class Chef
       property :source, String
       property :session_type, String
 
+      # StartCalendarInterval has some gotchas so we coerce it to help sanity
+      # check.  According to `man 5 launchd.plist`:
+      #   StartCalendarInterval <dictionary of integers or array of dictionaries of integers>
+      #     ... Missing arguments are considered to be wildcard.
+      # What the man page doesn't state, but what was observed (OSX 10.11.5, launchctrl v3.4.0)
+      # Is that keys that are specified, but invalid, will also be treated as a wildcard
+      # this means that an entry like:
+      #   { "Hour"=>0, "Weekday"=>"6-7"}
+      # will not just run on midnight of Sat and Sun, rather it will run _every_ midnight.
+      property :start_calendar_interval, [Hash, Array], coerce: proc { |type|
+        # Coerce into an array of hashes to make validation easier
+        array = type   if type.is_a?(Array)
+        array = [type] unless array
+
+        # Check to make sure that our array only has hashes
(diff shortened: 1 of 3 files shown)


thommay:
@lamont-granquist

lib/chef/resource/launchd.rb
```diff
@@ -46,6 +46,51 @@ def initialize(name, run_context = nil)
...
+      # this means that an entry like:
+      #   { "Hour"=>0, "Weekday"=>"6-7"}
+      # will not just run on midnight of Sat and Sun, rather it will run _every_ midnight.
+      property :start_calendar_interval, [Hash, Array], coerce: proc { |type|
+        # Coerce into an array of hashes to make validation easier
+        array = if type.is_a?(Array)
+                  type
+                else
+                  [type]
+                end
```
lamont-granquist: can't you `Array.new(type)` or `[type].flatten`?

lib/chef/resource/launchd.rb
```diff
@@ -46,6 +46,51 @@ def initialize(name, run_context = nil)
...
+
+        # Check to make sure that our array only has hashes
+        if array.any? { |obj| !obj.is_a?(Hash) }
```
lamont-granquist: negating the condition actually reads better to me:

`unless array.all? { |obj| obj.is_a?(Hash) }`

lib/chef/resource/launchd.rb
thommay: can probably type.flatten assuming we know we're not gonna get [[foo]] deliberately. We can't `Array.new` or `Array()` because they both do entirely the wrong thing - `Array.new()` creates an array of `len(type)` `nil`s, and `Array()` splatters a hash into a set of array pairs.

lamont-granquist: right i forgot about hash to array conversion

thommay: probably safer to keep it as it stands.

lib/chef/resource/launchd.rb
```diff
@@ -46,6 +40,51 @@ def initialize(name, run_context = nil)
...
+        # Check to make sure that our array only has hashes
+        unless array.all? { |obj| obj.is_a?(Hash) }
+          error_msg = "start_calendar_interval must be a single hash or an array of hashes!"
+          raise Chef::Exceptions::ValidationFailed, error_msg
+        end
+
+        # Make sure the hashes don't have any incorrect keys/values
+        array.each do |entry|
+          allowed_keys = %w{Minute Hour Day Weekday Month}
+          error_msg = "Invalid key for start_calendar_interval, must be one of: #{allowed_keys.join(", ")}"
```
btm: It'd be nice if the error message here and the one below contained the invalid value so you could grep for it.

spec/unit/provider/launchd_spec.rb
```diff
@@ -100,12 +129,61 @@ def run_resource_setup_for_action(action)
...
+          expect do
+            new_resource.start_calendar_interval "Hourly" => 1
+          end.to raise_error(/Invalid key/)
+        end
+
+        it "should not allow non-integer values" do
+          new_resource.program "/Library/scripts/call_mom.sh"
+          new_resource.time_out 300
+          expect do
+            new_resource.start_calendar_interval "Hour" => "1-2"
```
btm: I realize it has zero affect on the code, but I would test `Weekday` instead of `Hour`, as that is the most likely place someone might assume you could pass a non-integer value in the future.

btm:
@mikedodge04 any thoughts?

mikedodge04:
Looks Good to me ! This is implemented well IMO

tduffield:
@thommay did we want to add anything to the RELEASE_NOTES about this?

thommay:
I'm enclined to view this as a bug fix; we'll simply work harder to give people valid launchd unit files, so not sure it's RELEASE_NOTE worthy.

PR #5423, opened 2016-10-06
Fix manage_home provider inconsistency for Mac and FreeBSD User providers

lamont-granquist:
on freebsd and mac we're now telling people to use `manage_home` instead
of `supports[:manage_home]` but it doesn't actually work.  fixes that so
that it works now.


lamont-granquist:
@chef/client-core out deprecation warnings tell people to use an API that doesn't work on freebsd or mac.  fixes that.

PR #5424, opened 2016-10-06
Mac: Ensure that launchd limit_load_to_session_type is properly handled

erikng:
The key `LimitLoadToSessionType` is actually an array and not a string. This causes chef to fail or create an incorrect launchd.

### DIFF
--- a/lib/chef/resource/launchd.rb
+++ b/lib/chef/resource/launchd.rb
@@ -73,7 +73,7 @@ class Chef
       property :ld_group, String
       property :limit_load_from_hosts, Array
       property :limit_load_to_hosts, Array
-      property :limit_load_to_session_type, String
+      property :limit_load_to_session_type, Array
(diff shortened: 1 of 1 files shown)


thommay:
@erikng please add a DCO signoff to your commit. 
@chef/client-os-x please take a look.

erikng:
Added sign off.

lamont-granquist:
@jaymzh this look okay?

natewalck:
This is actually a string *or* an array of strings. This is still a bug ;) 

As per the man:

```
    LimitLoadToSessionType <string or array or strings>
     This configuration file only applies to sessions of the type(s) specified. This key only applies to jobs which are agents. There are no distinct sessions in the privileged system context.
```

nmcspadden:
Also, the Chef web documentation for launchd is now wrong, because it still states that it takes a string, when the code now requires an array. It should take both.

PR #5427, opened 2016-10-06
kinfe ssh: use the command line prompt for sudo if set

### LINKED ISSUE #4783: If a password is provided on the command line, sudo commands should not prompt for a password

lamont-granquist:
replaces #4784
closes #4783

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -541,6 +541,7 @@ class Chef
 
         configure_user
         configure_password
+        @password = config[:ssh_password] if config[:ssh_password]
         configure_ssh_identity_file
         configure_gateway
         configure_session


lamont-granquist:
technical details:

- since we're after configure_password `config[:ssh_password]` has already been demangled and set correctly from `config[:ssh_password_ng]` if its required
- since the code already does `@password ||= prompt_for_password` then by setting the `@password` here we bypass the prompt

this seems pretty minimally invasive and does seem like what people would want.  there's the open question of if it might concievably be breaking, but it doesn't really seem like it to me -- anyone who would have been 'broken' by it would by definition not been in a script and would be interactively typing their sudo prompt.  i guess its possible someone would have expect scripted around this and that expect script might break, but it feels like i'm well into noodling about incredibly silly edge conditions at that point -- and that person would probably be happy to remove all the expect scripting and just use this feature instead.

so i'm thinking this can go in without having it be defaulted to off, or have additional flags to turn it off (we can clearly add them later, but i'm leaning towards YAGNI at this point).

@chef/client-core thoughts?

thommay:
Yeah I follow your reasoning but I really can't see this being breaking in a meaningful sense. Let's do it.

tas50:
Should we note that in the release notes

PR #5439, opened 2016-10-10
Rhass/changelog 12.15

rhass:
Update changelog for upcoming release.

/cc @mwrock

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 **Enhancements:**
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
i updated all the titles of the PRs with some tweaks earlier today, are we not using `rake changelog` at all now?

mwrock:
We do for the initial generation @lamont-granquist . Since we moved to a tag per commit (or almost every commit) there is quite a bit of "doctoring" that needs to be done to the generated content. W've recognized that definitely needs to be addressed but in the meantime, we tend to edit the changelog "in place".

mwrock:
we continue to ensure that the GH commits reflect those edits so that when we do improve the generation process, these changes are kept in sync.

Issue #5438, opened 2016-10-10
Accumulator Patterns as Resources are still Somewhat Unsatisfying


lamont-granquist:
This is the current correct way to write a delayed accumulator as a resource pattern:

```ruby
provides :email_alias
resource_name :email_alias

property :address, String, name: true, identity: true
property :recipients, Array

default_action :create

action :create do
  with_run_context :root do
    nr = new_resource   # break out of the closure over the first call
    edit_resource(:template, "/tmp/aliases") do
      source 'aliases.erb'
      variables[:aliases] ||= {}
      variables[:aliases][nr.address] ||= []
      variables[:aliases][nr.address] += nr.recipients
      action :nothing
    end
  end
  log "force delayed notification" do
    notifies :create, "template[/tmp/aliases]", :delayed
  end
end
```

Two issues here:

1.  the scoping issue is fucking horribly confusing, the first time the resource is created the new_resource is frozen to the email_alias resource that first created it and subsequent calls pick up that resource out of the binding, and not the resource that is currently executing (and this isn't due to a 3694 error, it also has nothing to do with the custom resources scope confusion and happens with LWRPs as well).  And to be clear the binding closure here is not in the edit_resource block that is being created directly but that block is instance_eval'd against the template resource and the template resource has the binding of when it was first created which has a new_resource from the first invocation.

2.  the create-it-with-action-nothing-and-force-a-delayed-notification-via-a-log-resource pattern is mildly awful.  ideally there would be a way to do that when you create the resource so that it sends itself a delayed notification when its first created.... `delayed_action :create`?

lamont-granquist:
The way this is documented with find_resource works better because the block is only passed to the resource when its created, but is slightly less satisfying due to the need to assign the resource to a variable and manipulate it:

```ruby
provides :email_alias
resource_name :email_alias

property :address, String, name: true, identity: true
property :recipients, Array

default_action :create

action :create do
  r = with_run_context :root do
    find_resource(:template, "/tmp/aliases") do
      source 'aliases.erb'
      variables[:aliases] = {}
      action :nothing
    end
  end
  r.variables[:aliases][address] ||= []
  r.variables[:aliases][address] += recipients
  log "force delayed notification" do
    notifies :create, "template[/tmp/aliases]", :delayed
  end
end
```

thommay:
It feels like all this should get sugared into an `accumulated_resource` or something. But yes, agree that we should have a way to trigger a delayed notification automatically.

[...]

PR #5441, opened 2016-10-10
Core: pass new_resource to edit_resource through instance_exec

lamont-granquist:
helps with scoping issues, best practice would be to always use
the |new_resource| argument to the block when using edit_resource
inside of resources

addresses part of https://github.com/chef/chef/issues/5438

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -119,7 +119,7 @@ class Chef
       #
       def edit_resource!(type, name, created_at = nil, run_context: self.run_context, &resource_attrs_block)
         resource = find_resource!(type, name, run_context: run_context)
-        if block_given?
+        if resource_attrs_block
           if defined?(new_resource)
             resource.instance_exec(new_resource, &resource_attrs_block)
           else
@@ -206,7 +206,7 @@ class Chef
       def find_resource(type, name, created_at: nil, run_context: self.run_context, &resource_attrs_block)
         find_resource!(type, name, run_context: run_context)
       rescue Chef::Exceptions::ResourceNotFound
-        if block_given?
+        if resource_attrs_block
           declare_resource(type, name, created_at, run_context: run_context, &resource_attrs_block)
         end # returns nil otherwise
       end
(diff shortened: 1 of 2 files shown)


lib/chef/dsl/declare_resource.rb
```diff
@@ -119,7 +119,13 @@ def delete_resource(type, name, run_context: self.run_context)
       #
       def edit_resource!(type, name, created_at = nil, run_context: self.run_context, &resource_attrs_block)
         resource = find_resource!(type, name, run_context: run_context)
-        resource.instance_eval(&resource_attrs_block) if block_given?
+        if block_given?
```
coderanger: Is there a reason to mix the `block_given?` syntax with the `&block`?

lamont-granquist: well you have to protect that because the block is optional and instance_eval/instance_exec will otherwise get confused...   not sure if that's what you were referring to though...

lamont-granquist:
also noodling around with some tests around this -- although in my manual testing this works fine.  had to add the check around new_resource being defined because we might be in a recipe context with no wrapping resource, and tested that this is still backcompat and edit_resource without the new_resource arg in a provider still works fine, while edit_resource with a new_resource arg in a recipe context parses (but will be nil).  none of the permutations can blow this code up with an ArgumentError...

lib/chef/dsl/declare_resource.rb
coderanger: Usually you would do `if resource_attrs_block` in this syntax, `block_given?` is from the days of `yield`, before you could get the block as an actual object param.

lamont-granquist: had a failure to use pry correctly so thought that didn't work, fixing...

coderanger: Heh, not actually a problem per se, just usually you use one syntax or the other :)

lamont-granquist:
@chef/client-core if anyone else wants to take a look

rhass:
@lamont-granquist or @thommay  Can you PR the merge as `enhancement` or `bug` for changelog generation in the next release? Thanks!

afiune:
I second what Ryan just said. We need to make sure to tag the PR before merging please. 😄

PR #5443, opened 2016-10-11
Core: add delayed_action to the Chef::Resource API

lamont-granquist:
effectively creates a resource which sends itself a delayed notification

useful to create a resource which runs exactly once at the end of a run:

```ruby
with_run_context :root do
  find_resource(:execute, "/bin/true") do
    action :nothing
    delayed_action :run
  end
end
```

avoids using a log resource to force a resource to update and force a
delayed notification which avoids the extra output from the log
resource, and avoids the resource updated count always being updated.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -181,6 +181,31 @@ class Chef
     alias_method :action=, :action
 
     #
+    # Force a delayed notification into this resource's run_context.
+    #
+    # This should most likely be paired with action :nothing
+    #
+    # @param arg [Array[Symbol], Symbol] A list of actions (e.g. `:create`)
+    # @return [Array[Symbol]] the list of actions.
+    #
+    def delayed_action(arg = nil)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
probably needs some release_notes or doc_updates or whichever one we're doing these days, but we're right in the middle of getting 12.15 out right now, which will not include this, so i think i might collide with those edits...

coderanger:
Mentioned in Slack, but I don't love this API. Specifically this:

```ruby
action :foo do
  new_resource.delayed_action :one
  new_resource.delayed_action :two
  new_resource.delayed_action == :two
  # But both notifications will happen.
end
```

Probably remove the instance variable and make the return value undefined?

afiune:
@lamont-granquist @thommay Next time please try to label the PR before merge as `enhancement` or `bug` for changelog generation in the next release. Thanks! 😺

PR #5445, opened 2016-10-12
msu_package resource for Microsoft Update packages

NimishaS:
### Description

Resource/Provider for msu_package
### Check List
- [x] All tests pass
- [x] Documentation, especially RELEASE_NOTES.md, has been updated if
  required


NimishaS:
Travis is failing with error ` Failed to complete #converge action: [tar is corrupt, name contains null byte] on webapp-centos-6` while running `bundle exec kitchen test webapp-centos-6`, but this doesn't seem to be related to current changes.

Travis error started occurring after rebase.

NimishaS:
@btm , fixed review comments

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,148 @@
...
+          extract_msu_contents(@new_resource.source, @temp_directory)
+          @cab_files = read_cab_files_from_xml(@temp_directory)
+
+          unless @cab_files.empty?
+            # set the package version
+            # Here is a catch: How to set package version in case of multiple cab files?
+            @new_resource.source = @cab_files.first
+            cab_pkg = Chef::Provider::Package::Cab.new(@new_resource, nil)
+            new_resource.version(cab_pkg.package_version)
+            current_resource.version(cab_pkg.installed_version)
```
NimishaS: @btm, please have a look at this part of code where version needs to be set. Does this look ok, or is there a better way to do this?

NimishaS:
Release docs for this: https://github.com/chef/chef/pull/5501

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,148 @@
...
+          @temp_directory = Dir.mktmpdir("chef")
+          extract_msu_contents(@new_resource.source, @temp_directory)
+          @cab_files = read_cab_files_from_xml(@temp_directory)
+
+          unless @cab_files.empty?
+            # set the package version
+            # Here is a catch: How to set package version in case of multiple cab files?
+            @new_resource.source = @cab_files.first
+            cab_pkg = Chef::Provider::Package::Cab.new(@new_resource, nil)
+            new_resource.version(cab_pkg.package_version)
```
lamont-granquist: this shouldn't mutate the new_resource since you're mutating the user-intent.

somehow that needs to be state on this provider object (an accessor that gets the right thing in the correct conditions?)

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+
+        def download_source_file
+          source_resource.run_action(:create)
+          Chef::Log.debug("#{new_resource} fetched source file to #{source_resource.path}")
+          @new_resource.source = source_resource.path
+        end
+
+        def source_resource
+          @source_resource ||= Chef::Resource::RemoteFile.new(default_download_cache_path, run_context).tap do |r|
```
lamont-granquist: since this is a package resource subclass you should rebase on top of #5483 to get use_inline_resources.

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,148 @@
...
+            cab_pkg = Chef::Resource::CabPackage.new(@new_resource.name, run_context)
+            cab_pkg.source = cab_file
+            cab_pkg.run_action(:install)
+          end
+        end
+
+        def remove_package(name, version)
+          #use cab_package provider to remove the extracted cab packages
+          @cab_files.each do |cab_file|
+            cab_pkg = Chef::Resource::CabPackage.new(@new_resource.name, run_context)
```
lamont-granquist: same here that these two cab package resources should use `declare_resource(:cab_package, ....)`

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,148 @@
...
+        def read_cab_files_from_xml(msu_dir)
+          # get the file with .xml extension
+          xml_files = Dir.glob("#{msu_dir}/*.xml")
+          cab_files = []
+
+          if xml_files.empty?
+            raise Chef::Exceptions::Package, "Corrupt MSU package: MSU package doesn't contain any xml file"
+          else
+            # msu package contains only single xml file. So using xml_files.first is sufficient
+            file = ::File.new(xml_files.first)
```
lamont-granquist: probably `::File.open(xml_files.first) do |file|` here?

lib/chef/provider/package/msu.rb
NimishaS: @lamont-granquist , there isn't any command to get the package version for `msu` package. Please check https://msdn.microsoft.com/fr-ch/windows/hardware/commercialize/manufacture/desktop/dism-operating-system-package-servicing-command-line-options#_Get-PackageInfo___PackageName___name_in_image_____PackagePath___path_to_cabfile__. `Get-PackageInfo` command works only for `cab` files.

So in order to set the version for `msu_package`, we can either call the methods of `cab_package` OR we can re-write those methods in msu_package.
I have modified the code a bit to avoid mutation of `@new_resource.source`. Please let me know if there is a better way to achieve this. It would be helpful if you can share some example where this has been done.

NimishaS:
@lamont-granquist , please check my comment here: https://github.com/chef/chef/pull/5445#discussion_r86138365

Please let me know what approach should be taken to set the version for `msu_package`,

btm:
@chef/client-windows review please

spec/spec_helper.rb
```diff
@@ -139,6 +139,7 @@ module Shell
   config.filter_run_excluding :not_supported_on_gce => true if gce?
   config.filter_run_excluding :not_supported_on_nano => true if windows_nano_server?
   config.filter_run_excluding :win2k3_only => true unless windows_win2k3?
+  config.filter_run_excluding :windows2012r2_only => true unless windows_2012r2?
```
btm: I would name this filter `win2012r2_only` to match the above filter's pattern, or alternatively `windows_2012r2_only` to match the below filter, but not this pattern in between.

spec/support/platform_helpers.rb
```diff
@@ -67,6 +67,13 @@ def windows_2008r2_or_later?
   components.length >= 2 && components[0] >= 6 && components[1] >= 1
 end
 
+def windows_2012r2?
+  return false unless windows?
+  wmi = WmiLite::Wmi.new
+  host = wmi.first_of("Win32_OperatingSystem")
+  (host["version"] && host["version"].start_with?("6.3"))
```
btm: There are three methods here now checking `host["version"]`, please create another method that uses WMI and returns `host["version"]`, and update these three methods to use it to avoid the duplicated pattern.

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,146 @@
...
+            # set the package version
+            # Here is a catch: How to set package version in case of multiple cab files?
+            cab_resource = @new_resource
+            cab_resource.source = @cab_files.first
+            cab_pkg = Chef::Provider::Package::Cab.new(cab_resource, nil)
+
+            new_resource.version(cab_pkg.package_version)
+            current_resource.version(cab_pkg.installed_version)
+          else
+            raise Chef::Exceptions::Package, "Corrupt MSU package: MSU package XML doesn't contain any cab file"
```
btm: *does not contain any cab files

lib/chef/provider/package/msu.rb
```diff
@@ -0,0 +1,146 @@
...
+          @cab_files = read_cab_files_from_xml(@temp_directory)
+
+          unless @cab_files.empty?
+            # set the package version
+            # Here is a catch: How to set package version in case of multiple cab files?
+            cab_resource = @new_resource
+            cab_resource.source = @cab_files.first
+            cab_pkg = Chef::Provider::Package::Cab.new(cab_resource, nil)
+
+            new_resource.version(cab_pkg.package_version)
```
btm: > @lamont-granquist , there isn't any command to get the package version for msu package. 

I think the question here is, why are we setting `new_resource.version`, not why are we using the version from the CAB instead of the version form the MSU.

Normally `new_resource.version` is the version requested by the user by setting `version` on the resource. If the MSU version is meaningless, we should be able to pass the CAB version in `new_resource.version`. If the action is set to install, with a version set, and the underlying CAB provider cannot install that exact version, we should fail.

NimishaS: @btm , I went through the code to understand how `candidate_version` works. Also referred `chocolatey_package`. I found that `candidate_version` is usually nil and needs wrapping https://github.com/chef/chef/blob/master/lib/chef/provider/package.rb#L545. So even if I wrap `build_candidate_version` and `build_current_version` methods (the way they are wrapped in `chocolatey_package`), they will still be set using cab_package's `package_version` and `installed_version` methods.
So the approach will still remain same.

btm: To start with, you want to remove this line, the `new_resource.version(cab_pkg.package_version)`.

Then I'd look at something like `Chef::Provider::Package::Apt#get_current_versions` To answer the question of setting package version for multiple cab files, you could imitate something like the multi-package code in other providers.

In any case, all the code that opens up the cab files should be factored out to another method. You want a method that returns your `cab_pkg` , maybe `get_cab_package` and then in `load_current_resource` you set `current_resource.version(get_cab_package.installed_version)`.

Then you want to create a `candidate_version` method. this should return `get_cab_package.package_version`. `Chef::Provider::Package#target_version_array` when action is install, will set us up to try to install `new_version` if it is specified, that is if the user specific a version on the resource and it is not the current version, or if `new_version` was not set, it will try to install the `candidate_version`. The existing logic in `Chef::Provider::Package` should do all the right things for other situations like if you specify a version that differs than what's in the cab file, e.g. no matching candidates.

btm:
@mwrock release notes are here: https://github.com/chef/chef/pull/5501

btm:
@mwrock in `functional/resource/windows_package_spec.rb` it looks like you download files from URLs? Or maybe they're just being used but not downloaded, but IIRC it would be downloading them and parsing them to decide what package type they are. I don't see us using an S3 bucket anywhere else, and I believe @NimishaS has URL support here, so maybe we get by [for now] using the download URLs?

https://download.microsoft.com/download/5/2/B/52BE95BF-22D8-4415-B644-0FDF398F6D96/IE10-Windows6.1-KB2859903-x86.msu
https://download.microsoft.com/download/3/B/3/3B320C07-B7B1-41E5-81F4-79EBC02DF7D3/Windows8.1-KB2959977-x64.msu

@NimishaS You'll need to rebase anyway with the powershell package provider having been merged. Wherever we get the MSU files from that isn't here, you'll want to make sure that they're never added in your get history as part of that rebase so there isn't a huge commit in the history that adds them and then they are later removed.

mwrock:
Yeah I think the ms URLs are fine.

NimishaS:
@btm, @mwrock Functional specs were working fine locally after replacing the sample MSU files by URLs. But Appveyor is failing with:
```
OpenSSL::SSL::SSLError: SSL Error connecting to https://download.microsoft.com/download/3/B/3/3B320C07-B7B1-41E5-81F4-79EBC02DF7D3/Windows8.1-KB2959977-x64.msu - SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed
```

btm:
@NimishaS I think we need to add something like this to our `appveyor.yml`:

```
appveyor DownloadFile http://curl.haxx.se/ca/cacert.pem -FileName C:\cacert.pem
set SSL_CERT_FILE=C:\cacert.pem
```

https://github.com/chef/license_scout/blob/master/appveyor.yml#L12-L13

NimishaS:
@btm, @mwrock , please review

mwrock:
LGTM. Please squash commits.

NimishaS:
@mwrock , I had squashed some commits. Do you want to keep just a single commit?

mwrock:
yes please @NimishaS

NimishaS:
Done @mwrock

PR #5446, opened 2016-10-12
Core: make delayed_action argument mandatory, remove misleading ivar introduced in PR that added delayed_action

lamont-granquist:
see https://github.com/chef/chef/pull/5443#issuecomment-253279021

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -187,7 +187,7 @@ class Chef
     #
     # @param arg [Array[Symbol], Symbol] A list of actions (e.g. `:create`)
     #
-    def delayed_action(arg = nil)
+    def delayed_action(arg)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@coderanger `Array.new(nil)` gives `[]` which is just a no-op,  IDK if you had better behavior in mind...

lamont-granquist:
i suppose we could just drop the default value and make it ArgumentError if you don't pass it an action.

coderanger:
:+1: for making it required. That basically reduces this to being (well needed) sugar for `add_delayed_notification`.

lamont-granquist:
done, travis willing...

rhass:
@lamont-granquist Can you PR the merge as `enhancement` or `bug` for changelog generation in the next release? Thanks!

PR #5453, opened 2016-10-13
update PR template

lamont-granquist:
users don't understand what we mean by "all documentation", so
clarify that what we mean to be RELEASE_NOTES.md and that's it, they
need to stay away from CHANGELOG.md

i believe we've dropped DOC_CHANGES.md in favor of RELEASE_NOTES.md


tas50:
This could use more clarification but this is a step in right direction

lamont-granquist:
@thommay this is correct for our current process?

PR #5460, opened 2016-10-14
Core: Structure deprecations with additional metadata

thommay:
Primarily adding URLs so that deprecation messages can point to full
documentation on the deprecation and how to fix it.

cc @lamont-granquist @tas50


thommay:
@coderanger this.

thommay:
Goes rather hand in hand with chef/chef-web-docs#231

thommay:
@chef/client-core @adamleff

lib/chef/deprecated.rb
```diff
@@ -0,0 +1,199 @@
...
+    class InternalApi < Base
+      def id
+        0
+      end
+      def target
+        "internal_api.html"
+      end
+    end
+
+3694
```
lamont-granquist: derp?

thommay: derp indeed.

lib/chef/deprecated.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    class InternalApi < Base
+      def id
+        0
+      end
+      def target
+        "internal_api.html"
+      end
+    end
+
```
thommay: Yeah, de-duped now :)

adamleff:
Not a maintainer, but 👍 from me! The changes to the data collector stuff will work just fine for us too. Well done, @thommay!

PR #5462, opened 2016-10-15
Rebuild HTTP headers on each retried request

sds:
### Description

Depending on how you configure your Chef client, it's possible that a
later retry would have an authentication timestamp header
(`X-OPS-TIMESTAMP`) that was more than 900 seconds old. Since the Chef
server treats timestamps older than this threshold as unauthorized
requests, you would potentially get seemingly spurious "401
Unauthorized" responses from the server.

A solution is to generate the headers on each retry so that we have a
new authentication timestamp for each retried request.

I'm not sure if this change of behavior requires additional thought. Hoping someone with more context can weigh in here.

### Issues Resolved

- [Intermittent "401 Unauthorized" failures from Chef server using Chef client](https://github.com/chef/chef-server/issues/968)

### Check List

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -329,10 +329,9 @@ class Chef
     # Runs a synchronous HTTP request, with no middleware applied (use #request
     # to have the middleware applied). The entire response will be loaded into memory.
     # @api private
-    def send_http_request(method, url, headers, body, &response_handler)
-      headers = build_headers(method, url, headers, body)
-
+    def send_http_request(method, url, base_headers, body, &response_handler)
       retrying_http_errors(url) do
+        headers = build_headers(method, url, base_headers, body)
         client = http_client(url)
         return_value = nil
         if block_given?


sds:
The build failure on Ubuntu 16.04 seems spurious: https://travis-ci.org/chef/chef/jobs/167933858

lamont-granquist:
so, you do realize that we normally treat failing chef-client runs when the server is undergoing maintenance as being something perfectly acceptable?  the general solution is "run it again" -- which relies on the idempotent, convergent design so that prior runs can fail at random points in the middle of the run, but subsequent runs will continue to reconfigure the node towards the state it should be in (which is more or less the cfengine philosophy from the 90s that chef, puppet and cfengine all share).

at the same time, this PR looks fine...

lamont-granquist:
@z3usy, we'd take small patches to make chef incrementally more reliable (which this PR is), we would not take large patches to try to make chef perfectly reliable (which is unattainable and picks a fight with over 20+ years of config management philosophy -- at some point chef has to be able to fail sometimes).

z3usy:
@lamont-granquist I totally understand and agree, although your rant just seemed a bit out of place to me as this fixes auth-header timestamps and doesn't have anything to do with run failure or retry logic, hence the confused emoji.

I'm just glad that it's accepted and look forward to the patch landing. Thanks for your quick acceptance!

PR #5463, opened 2016-10-18
[Knife] Fix deprecation warning when creating data bag items

thommay:
cc @chef/client-core
Fixes chef/chef-vault#215


danielsdeleo:
Haven't looked at the code yet, but I'd suggest looking at adding fatal deprecation warnings to chef-vault if possible.

PR #5471, opened 2016-10-19
convert svn_password to property + make sensitive

lamont-granquist:
should always suppress the password output by default since typically nobody wants to see that in their logs when we crash.

### DIFF
--- a/lib/chef/resource/scm.rb
+++ b/lib/chef/resource/scm.rb
@@ -89,7 +89,7 @@ class Chef
         )
       end
 
-      property :svn_password, String, sensitive: true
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/core  so this looks like a correct thing to do, but it doesn't quite achieve the goals.  this will suppress the password from the resource reporter by default, but it will not affect property being output in crashes of the resource, there needs to be a patch here as well to suppress individual properties if the property itself is marked sensitive:

https://github.com/chef/chef/blob/11d64c2438d6e387986adce3d4740f2bc7efaab9/lib/chef/resource.rb#L668-L683

that is a bit more of a distraction than i want to take on right now though.

PR #5472, opened 2016-10-19
Address test failures in SUSE 11.4 and 12.1

tduffield:
### Description

With the recent addition of SLES 11.4 and SLES 12.1 builders/testers to our internal build matrix, we have uncovered some bugs / false assumptions about the behavior of these two systems. This change updates our code base to address those issues.

### Issues Resolved

Internal issue COOL-597

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/group/suse.rb
+++ b/lib/chef/provider/group/suse.rb
@@ -17,6 +17,7 @@
 #
 
 require "chef/provider/group/groupadd"
+require "etc"
 
 class Chef
   class Provider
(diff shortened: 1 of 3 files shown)


lib/chef/provider/group/suse.rb
```diff
@@ -51,7 +52,12 @@ def set_members(members)
         end
 
         def add_member(member)
-          shell_out!("groupmod -A #{member} #{@new_resource.group_name}")
+          begin
+            Etc.getpwnam(member)
+            shell_out!("groupmod -A #{member} #{@new_resource.group_name}")
+          rescue ArgumentError => e
+            raise Chef::Exceptions::Group, "Could not add #{member} to #{@new_resource.group_name}: user does not exist"
```
lamont-granquist: that looks like it should get moved to define_resource_requirements

tduffield: The reason I ask is I _could_ iterate through `new_resource.members` and bomb out of one of those members isn't created, but that'd only be reliable if the assertion is evaluated during the execution (i.e. the chef recipe creates the user).

danielsdeleo: For the why-run case, I think that's what you want, e.g., Chef should say that the user doesn't exist, assuming it would have been created by a prior step.

lamont-granquist: you should probably add at least a `def to_add { members - @current_resource.members }` which you can call from `define_resource_requirements` and then `to_add.all? { |member| Etc.getpwnam(member) }` with some kinda rescue that fails it more or less.

lamont-granquist:
👍

PR #5478, opened 2016-10-20
Windows: search for config on same drive as executable location

### LINKED ISSUE #5477: Default config file drive on windows doesn't support non-system drive installations
## Description

If you install chef-client on a drive != $env:SYSTEMDRIVE ( c: ) then a standard chef-client parameter'less launch will fail.  This is because the default config file location is $env:SYSTEMDRIVE/chef/client.rb
## Chef Version

12.14.89
## Platform Version

Any windows-  tested windows 7, server 2k8r2, server 2k12
## Replication Case

Install chef on a drive other than C: / env[SystemDrive] and launch chef-client with no parameters
## Client Output

none immediately available.  will update the ticket when I have it available. 
## Stacktrace

none immediately available.  will update the ticket when I have it available.
### NOTE: CHEF CLIENT BUGS ONLY

This issue tracker is for the code contained within this repo -- `chef-client`, base `knife` functionality (not
plugins), `chef-apply`, `chef-solo`, `chef-client -z`, etc.
- [Server issues](https://github.com/chef/chef-server/issues/new)
- [ChefDK issues](https://github.com/chef/chef-dk/issues/new)
- Cookbook Issues (see the https://github.com/chef-cookbooks repos or search [Supermarket](https://supermarket.chef.io) or GitHub/Google)

jkerry:
### Description

This change causes the platform_specific_path function to root its's path in the drive `__FILE__` is found in. This is arguably more accurate than $env:SystemDrive since in some security contexts a chef-client installation on a drive other than the OS is preferred. 

### Issues Resolved

fixes #5477
### Check List
- [x] All tests pass
- [x] RELEASE_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -67,7 +67,11 @@ module ChefConfig
     end
 
     def self.windows_installation_drive
-      File.expand_path(__FILE__).split('/', 2)[0] if ChefConfig.windows?
+      if ChefConfig.windows?
+        drive = File.expand_path(__FILE__).split('/', 2)[0]
+        drive = ENV["SYSTEMDRIVE"] if drive.to_s == ''
+        drive
+      end
     end
 
     def self.add_formatter(name, file_path = nil)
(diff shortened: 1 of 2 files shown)


thommay:
cc @chef/client-windows

jkerry:
I suppose another way of doing this would be to set and accept a CHEF_HOME environment variable but I'm not sure if the cost of adding that dependency is more or less evil than inspecting the executable location.

thommay:
Thanks for your contribution, @jkerry !

PR #5483, opened 2016-10-21
Core: set use_inline_resources in package superclass

lamont-granquist:
allows subclasses to use_inline_resources (hab_package
will most likely want this)

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -31,6 +31,8 @@ class Chef
       include Chef::Mixin::ShellOut
       extend Chef::Mixin::SubclassDirective
 
+      use_inline_resources
+
       # subclasses declare this if they want all their arguments as arrays of packages and names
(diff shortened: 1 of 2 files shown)


lamont-granquist:
i don't believe this changes any existing behavior.

coderanger:
I subclass the package provider a whole bunch, so it might break some of my stuff.

lamont-granquist:
hrm, can you test and get back if it fails and how it fails?

i'd mostly guess that you turn on something similar which might get into a fight with this, but we should find out what that is and resolve any impedence mismatch there before chef 13 anyway...

coderanger:
I just did a quick lookover of the code for `python_package` and `node_package`, both look like they should be okay. Neither uses `notifying_block` anywhere, it's all straight `shell_out` calls. :+1: to merge and see if anything breaks :)

lamont-granquist:
looks like i need to add a mock to some windows unit tests...

PR #5486, opened 2016-10-24
Properly check lock status of user on solaris2

tduffield:
### Description

On Solaris, the 'shadow' database does not exist within `getent`, so the
checking for the username there won't return accurate results. As the
Solaris provider assumes user management via /etc/shadow, we can very
easily parse the contents of the file directly

``` shell
tduffield@chef-solaris-11-i86pc:~$ getent shadow tduffield
Unknown database: shadow
usage: getent database [ key ... ]
```
### Issues Resolved
- Internal ticket COOL-413
### Check List
- [ ] All tests pass
- [ ] RELEASE_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)

### DIFF
--- a/lib/chef/provider/user/solaris.rb
+++ b/lib/chef/provider/user/solaris.rb
@@ -46,21 +46,14 @@ class Chef
         end
 
         def check_lock
-          shadow_line = shell_out!("getent", "shadow", new_resource.username).stdout.strip rescue nil
+          user = IO.read(@password_file).match(/^#{Regexp.escape(@new_resource.username)}:([^:]*):/)
 
-          # if the command fails we return nil, this can happen if the user
-          # in question doesn't exist
-          return nil if shadow_line.nil?
+          # If we're in whyrun mode, and the user is not created, we assume it will be
+          return false if whyrun_mode? && user.nil?
 
-          # convert "dave:NP:16507::::::\n" to "NP"
-          fields = shadow_line.split(":")
+          raise Chef::Exceptions::User, "Cannot determine if #{@new_resource} is locked!" if user.nil?
 
-          # '*LK*...' and 'LK' are both considered locked,
-          # so look for LK at the beginning of the shadow entry
-          # optionally surrounded by '*'
-          @locked = !!fields[1].match(/^\*?LK\*?/)
-
-          @locked
+          @locked = !!(user[1] =~ /^\*?LK\*?/)
         end
 
         def lock_user
--- a/lib/chef/provider/user/solaris.rb
+++ b/lib/chef/provider/user/solaris.rb
@@ -53,7 +53,7 @@ class Chef
 
           raise Chef::Exceptions::User, "Cannot determine if #{@new_resource} is locked!" if user.nil?
 
-          @locked = !!(user[1] =~ /^\*?LK\*?/)
+          @locked = user[1].start_with?("*LK*")
         end
 
         def lock_user
--- a/spec/unit/provider/user/solaris_spec.rb
+++ b/spec/unit/provider/user/solaris_spec.rb
@@ -123,11 +123,11 @@ describe Chef::Provider::User::Solaris do
 
       # locked shadow lines
       [
-        "adam:LK:::::::",
         "adam:*LK*:::::::",
         "adam:*LK*foobar:::::::",
         "adam:*LK*bahamas10:::::::",
         "adam:*LK*goonawaLK:::::::",
+        "adam:*LK*LKgir:::::::",
         "adam:*LK*L....:::::::",
       ].each do |shadow|
         context "for user 'adam' with entry '#{shadow}'" do
@@ -146,6 +146,7 @@ describe Chef::Provider::User::Solaris do
         "adam:foobar:::::::",
         "adam:bahamas10:::::::",
         "adam:goonawaLK:::::::",
+        "adam:LKgir:::::::",
         "adam:L...:::::::",
       ].each do |shadow|
         context "for user 'adam' with entry '#{shadow}'" do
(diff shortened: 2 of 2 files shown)


tduffield:
cc @chef/client-core for review

tduffield:
maybe also @chef/solaris as well.

lib/chef/provider/user/solaris.rb
```diff
@@ -46,7 +46,11 @@ def manage_user
         end
 
         def check_lock
-          shadow_line = shell_out!("getent", "shadow", new_resource.username).stdout.strip rescue nil
+          shadow_line = if node["platform"] == "solaris2"
+                          shell_out!("grep", new_resource.username, @password_file)
```
stevendanna: What if you have a user named `greg` and a user named `gregory`?  Perhaps we should ancor the search a bit: `"^#{new_resource.username}:"` to ensure that we only pick up exact matches.  Alternatively, we could read in the whole file and parse it directly.

coderanger: Yeah, never shell out to grep, implement it in Ruby.

stevendanna:
> Does this introduce behavior differences between omnios and OSes we report as solaris2 in cases where they have a non-file-based shadow database configured in nsswitch.conf? I'm unsure which solaris's support that, perhaps the solaris maintainers could comment there.

Reading the provider more carefully reveals this already assume /etc/shadow access, so I think we can ignore that.

tduffield:
@stevendanna @coderanger after doing a little bit more research, I was able to consolidate the lock evaluation into a single Ruby-based code branch.

lib/chef/provider/user/solaris.rb
```diff
@@ -46,19 +46,17 @@ def manage_user
...
-          shadow_line = shell_out!("getent", "shadow", new_resource.username).stdout.strip rescue nil
-
-          # if the command fails we return nil, this can happen if the user
-          # in question doesn't exist
-          return nil if shadow_line.nil?
-
-          # convert "dave:NP:16507::::::\n" to "NP"
-          fields = shadow_line.split(":")
+          @locked = nil
+          ::File.readlines(@password_file).each do |entry|
```
coderanger: A possibly simpler option?

```ruby
def check_lock
  # '*LK*...' and 'LK' are both considered locked
  locked = IO.read(@password_file) =~ /^#{RegExp.escape(@new_resource.username)}:\*?LK\*?:/
  return false if whyrun_mode? && locked.nil?
  locked
end
```

tduffield: So, in this case wouldn't locked always be `true` or `false`? If the user does not exist, then it wouldn't be in the file and the matcher would simply return false?

tduffield: So I gave this a try, and I couldn't get it to work. Everything kept coming back nil. My Friday afternoon brain couldn't figure out why.

lamont-granquist: My Friday night brain reads that in Ethel Mermen's voice singing "Everything's coming up nil"...

spec/unit/provider/user/solaris_spec.rb
```diff
@@ -109,25 +129,29 @@
...
+          let(:user_lock) { shadow }
+
+          it "should return true" do
+            expect(provider.check_lock).to eql(true)
+          end
         end
       end
 
-      # unlocked shadow lines
+      # unlcoked shadow lines
```
stevendanna: minor typo here.

lib/chef/provider/user/solaris.rb
```diff
@@ -46,21 +46,14 @@ def manage_user
...
-          fields = shadow_line.split(":")
+          raise Chef::Exceptions::User, "Cannot determine if #{@new_resource} is locked!" if user.nil?
 
-          # '*LK*...' and 'LK' are both considered locked,
-          # so look for LK at the beginning of the shadow entry
-          # optionally surrounded by '*'
-          @locked = !!fields[1].match(/^\*?LK\*?/)
-
-          @locked
+          @locked = !!(user[1] =~ /\*?LK\*?/)
```
coderanger: You mean `/^\*?LK\*?$/`, this will otherwise match any password hash that happens to have `LK` in it, which we should add as a test probably :)

lib/chef/provider/user/solaris.rb
```diff
@@ -46,21 +46,14 @@ def manage_user
         end
 
         def check_lock
-          shadow_line = shell_out!("getent", "shadow", new_resource.username).stdout.strip rescue nil
+          user = IO.read(@password_file).match(/^#{Regexp.escape(@new_resource.username)}:(\*?\w+\*?)/)
```
coderanger: If you made this `/^#{Regexp.escape(@new_resource.username)}:([^:]*):/` it would be slightly more robust since you are checking the hash bit in a second regexp below and just want to extract the field I think.

lamont-granquist:
if a linux using sysadmin writes "!!" manually to /etc/shadow does Solaris behave as if it is locked, and should we be able to unlock that account (and does `passwd -u` unlock it)?

tduffield:
According to multiple man pages I found for Solaris, the account is *only* locked if the first 4 characters are `*LK*`. Source: http://docs.oracle.com/cd/E23824_01/html/821-1473/shadow-4.html. Not sure what Solaris would do if a sysadmin manually wrote `!!` into /etc/shadow, but the documentation suggests it would not be considered locked.

lamont-granquist:
Yeah, but what else it is going to do other than act like a locked account?  It feels like we should apply Postel's Law here and be liberal in what we accept.

lamont-granquist:
Looks like we've validated that `*LK*` is the only string which causes solaris to behave like its a locked account.  Any other string behaves like its unlocked.

PR #5487, opened 2016-10-25
Enable Signed Header Auth for Data Collector, and Configure the Data Collector Automatically

danielsdeleo:
### Description
- Allow the Chef Automate data collector to authenticate using signed headers with the client key.
- Attempt to automatically configure the data collector when running in client-server mode
- Token auth still available for either mode if configured
- Check for the presence of both the data collector URL and token to enable the collector in solo mode.
### Issues Resolved

Internal Chef Ticket
### Check List
- [x] All tests pass
- [x] RELEASE_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -857,7 +857,7 @@ module ChefConfig
       # Full URL to the endpoint that will receive our data. If nil, the
       # data collector will not run.
       # Ex: http://my-data-collector.mycompany.com/ingest
-      default :server_url,       nil
+      default(:server_url) { File.join(config_parent.chef_server_url, "/data_collector") }
 
       # An optional pre-shared token to pass as an HTTP header (x-data-collector-token)
       # that can be used to determine whether or not the poster of this
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -857,7 +857,7 @@ module ChefConfig
       # Full URL to the endpoint that will receive our data. If nil, the
       # data collector will not run.
       # Ex: http://my-data-collector.mycompany.com/ingest
-      default(:server_url) { File.join(config_parent.chef_server_url, "/data_collector") }
+      default(:server_url) { File.join(config_parent.chef_server_url, "/data-collector") }
 
       # An optional pre-shared token to pass as an HTTP header (x-data-collector-token)
       # that can be used to determine whether or not the poster of this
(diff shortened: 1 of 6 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -857,7 +857,7 @@ def self.windows_home_path
       # Full URL to the endpoint that will receive our data. If nil, the
       # data collector will not run.
       # Ex: http://my-data-collector.mycompany.com/ingest
-      default :server_url,       nil
+      default(:server_url) { File.join(config_parent.chef_server_url, "/data_collector") }
```
stevendanna: Chef server currently has a minimal proxy config for the data collector at `/data-collector`

https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-cookbooks/private-chef/templates/default/nginx/nginx_chef_api_lb.conf.erb#L48-L57

danielsdeleo:
@sersut I added a bunch of logging to the data collector enable/disable decision-making. For case (2) I decided not to add anything because the initial data collector POST will fail with an informative error message that includes the configured URL, which should be enough information for a customer to figure out the problem.

sersut:
👍

danielsdeleo:
Note: I am planning to do end-to-end tests for a few scenarios once the server piece is done, so please consider this PR on hold for now.

spec/unit/data_collector_spec.rb
```diff
@@ -23,6 +23,14 @@
 require "chef/resource_builder"
 
 describe Chef::DataCollector do
+
+  # TODO:
```
srenatus: Are these still open? (I'm not sure what exactly you tracked here, but it seems like these have been taken care of by this change, doesn't it?)

chef-config/lib/chef-config/config.rb
```diff
@@ -857,7 +857,7 @@ def self.windows_home_path
       # Full URL to the endpoint that will receive our data. If nil, the
       # data collector will not run.
       # Ex: http://my-data-collector.mycompany.com/ingest
-      default :server_url,       nil
+      default(:server_url) { File.join(config_parent.chef_server_url, "/data-collector") }
```
srenatus: Present clients will have this one configured, right? (I.e., they're not using the default value.)

danielsdeleo: Yes, before this patch is applied, you must explicitly configure this in order for the data collector to be enabled (it's in the old behavior of `DataCollector.register_reporter?()` ).

spec/unit/data_collector_spec.rb
danielsdeleo: Good catch, forgot to grep my diff for `TODO`. Will remove.

danielsdeleo:
@chef/client-core This should be G2G. My team will be doing end-to-end testing for a few scenarios including not having Automate/Chef Server Authenticated proxy available.

PR #5488, opened 2016-10-25
Support knife attribute filters with dots

tduffield:
### Description

Some node attribute keys have dots in them, i.e. our current windows
package resolves to `packages.Chef Client v12.12.15`. Currently, when
parsing our `--attribute` flags we simply split on a `.` to indicate
nesting, which would result in the data not coming back to the user as
expected. This change adds support for quoted attributes in the filter
flags, so that you can specify keys that include special characters like
dots.

```shell
knife node show NODE --attribute 'packages."Chef Client v12.12.15"'
knife node show NODE --attribute "packages.'Chef Client v12.12.15'"
```

Because the shell processes the strings before we even get the chance to
see it, the user will need to make sure they escape or use a combination
of single or double quotes to preserve the quoting as appropriate.

### Issues Resolved

* Internal ticket COOL-599

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -174,7 +174,7 @@ class Chef
         end
 
         def extract_nested_value(data, nested_value_spec)
-          nested_value_spec.split(".").each do |attr|
+          extract_attribute_chain(nested_value_spec).each do |attr|
             data =
               if data.is_a?(Array)
                 data[attr.to_i]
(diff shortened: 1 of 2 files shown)


coderanger:
:-1: on the quoting unless we actually update the RFC first and agree on some standard. The sprawl of different attribute processing styles is already way out of hand.

coderanger:
https://github.com/chef/chef-rfc/blob/master/rfc018-attribute-syntax.md#command-line-access-knife for those that don't want to go find it. The fix we approved back then was to add a `-F` option to specify the split character. It's cool if the community wants to change that, but for now this is out of spec.

tduffield:
Ah, I totally missed that RFC. Thanks @coderanger!

PR #5489, opened 2016-10-25
Add the `--field-separator` flag to knife show commands

tduffield:
### Description

For the commands that allow you to filter which attributes it shows you
using the `--attribute` flag, add the `--field-separator` flag to allow
the user to customize which character(s) they want to use to delineate the
attribute hierarchy.

``` shell
knife node show NODE -S: -a "packages:Chef Client v12.12.15"
```
### Issues Resolved
- Internal ticket COOL-599
### Check List
- [ ] All tests pass
- [ ] RELEASE_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

### DIFF
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -28,12 +28,14 @@ class Chef
         # :nodoc:
         def self.included(includer)
           includer.class_eval do
-            @attrs_to_show = []
             option :attribute,
               :short => "-a ATTR1 [-a ATTR2]",
               :long => "--attribute ATTR1 [--attribute ATTR2] ",
-              :proc => lambda { |val| @attrs_to_show << val },
-              :description => "Show one or more attributes"
+              :description => "Show one or more attributes",
+              :proc => Proc.new { |a|
+                Chef::Config[:knife][:attribute] ||= []
+                Chef::Config[:knife][:attribute].push(a)
+              }
           end
         end
       end
--- a/lib/chef/knife/core/generic_presenter.rb
+++ b/lib/chef/knife/core/generic_presenter.rb
@@ -28,6 +28,11 @@ class Chef
         # :nodoc:
         def self.included(includer)
           includer.class_eval do
+            option :field_separator,
+              :short => "-S SEPARATOR",
+              :long => "--field-separator SEPARATOR",
+              :description => "Character separator used to delineate nesting in --attribute filters (default \".\")"
+
             option :attribute,
               :short => "-a ATTR1 [-a ATTR2]",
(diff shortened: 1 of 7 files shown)


lib/chef/knife/client_show.rb
```diff
@@ -42,6 +42,7 @@ def run
 
         client = Chef::ApiClientV1.load(@client_name)
         output(format_for_display(client))
+        self.class.attrs_to_show = []
```
stevendanna: I see we are doing this in environment_show and other commands as well.  Why is it it needed, isn't this command about to exit right after we call this?  Perhaps a comment would be appropriate here?

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -29,11 +29,21 @@ module MultiAttributeReturnOption
         def self.included(includer)
           includer.class_eval do
             @attrs_to_show = []
+
+            option :field_separator,
+              :short => "-S SEPARATOR",
+              :long => "--field-separator SEPARATOR",
+              :description => "Character separator used to delineate nesting in --attribute filters"
```
stevendanna: Might be good to list the default in the description.

tduffield:
cc @chef/client-core

lamont-granquist:
the clear_attrs_to_show call should probably be the responsibility of the rspec tests to run.

its global state so it should go in the global state clearing that we do where Chef::Config gets reset to its default (hanging off of the spec_helper).

tduffield:
@lamont-granquist What about the case where someone (maybe another Knife plugin) is coding against those libraries directly?

lamont-granquist:
actually it would be better if that class variable was eliminated and the option was changed to accumulate state in `Chef::Config[:knife]`, for example:

https://github.com/chef/chef/blob/01ed0309458e69a43cf984b87961baba0857fc3d/lib/chef/knife/bootstrap.rb#L251-L260

then you just leverage the existing code that resets the Chef::Config state on every rspec run.

lamont-granquist:
i'm not sure what that use case would be, but in that case, they need to worry about resetting global state like Chef::Config anyway (which is why i think it should go into Chef::Config)

tduffield:
Okay. I'll look into that.

lamont-granquist:
i think you can just accumulate the array directly into `Chef::Config[:knife][:attribute]` and any subclasses should be using `option[:attribute]` and not poking that class variable directly -- and certainly if anything breaks that is backwards-compatible API that they should have been using all along so that broken knife plugin should just get fixed.

lamont-granquist:
👍

PR #5495, opened 2016-10-28
add attribute_changed hook to event handlers

lamont-granquist:
🚧 needs tests


lamont-granquist:
@chef/client-core should be ready for review

PR #5496, opened 2016-10-28
Add deprecations to Data Collector competion message

adamleff:
By adding deprecation warnings to Data Collector, receivers of
Data Collector messages can easily identify nodes that would need
attention to their versions/cookbooks before a major version Chef
upgrade would be successful.

### DIFF
--- a/lib/chef/data_collector/messages.rb
+++ b/lib/chef/data_collector/messages.rb
@@ -66,7 +66,7 @@ class Chef
           "entity_uuid"            => node_uuid,
           "expanded_run_list"      => reporter_data[:expanded_run_list],
           "id"                     => run_status.run_id,
-          "message_version"        => "1.0.0",
+          "message_version"        => "1.1.0",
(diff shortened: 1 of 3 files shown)


adamleff:
@thommay fixed up that chefstyle issue - all green! Thanks for taking a look!

lib/chef/data_collector/messages.rb
```diff
@@ -80,6 +80,7 @@ def self.run_end_message(reporter_data)
           "status"                 => reporter_data[:status],
           "total_resource_count"   => reporter_data[:resources].count,
           "updated_resource_count" => reporter_data[:resources].select { |r| r.report_data["status"] == "updated" }.count,
+          "deprecations"           => reporter_data[:deprecations],
```
shortdudey123: looks like this hash is alphabetical, should we keep it that way?

thommay:
Thanks @adamleff ; looking forward to seeing the results of this!

Issue #2919, opened 2015-02-16
knife ssl fetch needs better cert validation logic and should not throw NoMethodError on nil


isuftin:
When trying to figure out an issue with my SSL cert not being verified even though I have the cert in a trusted cert directory, I am seeing this issue:

```
$ knife ssl fetch -VV https://chef.owicloud.org
INFO: Using configuration from /Users/isuftin/.chef/knife.rb
DEBUG: Checking SSL cert on https://chef.owicloud.org
WARNING: Certificates from chef.owicloud.org will be fetched and placed in your trusted_cert
directory (/Users/isuftin/.chef/trusted_certs).

Knife has no means to verify these are the correct certificates. You should
verify the authenticity of these certificates after downloading.

/opt/chefdk/embedded/apps/chef/lib/chef/knife/ssl_fetch.rb:92:in `cn_of': undefined method `[]' for nil:NilClass (NoMethodError)
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife/ssl_fetch.rb:118:in `write_cert'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife/ssl_fetch.rb:137:in `block in run'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife/ssl_fetch.rb:136:in `each'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife/ssl_fetch.rb:136:in `run'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife.rb:417:in `block in run_with_pretty_exceptions'
        from /opt/chefdk/embedded/apps/chef/lib/chef/local_mode.rb:38:in `with_server_connectivity'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife.rb:416:in `run_with_pretty_exceptions'
        from /opt/chefdk/embedded/apps/chef/lib/chef/knife.rb:213:in `run'
        from /opt/chefdk/embedded/apps/chef/lib/chef/application/knife.rb:139:in `run'
        from /opt/chefdk/embedded/apps/chef/bin/knife:25:in `<top (required)>'
        from /usr/bin/knife:40:in `load'
        from /usr/bin/knife:40:in `<main>'
```

Chef version: 12.0.3

ranjib:
looks like the ssl cert being [fetched](https://github.com/chef/chef/blob/master/lib/chef/knife/ssl_fetch.rb#L73) has malformed /empty data. Can you provide some information? what happens when you do:
[...]

PR #5498, opened 2016-10-31
Updating knife ssl fetch to correctly store certificate when it does not have a CN

tyler-ball:
### Description

If a certificate in the chain does not contain a common name we currently fail hard. It looks like the only reason we fetch the CN is so we can write out a meaningful certificate file name locally.

My current code uses a combination of URI host and timestamp to write out the certificate file if it does not have a CN. I added a timestamp to address the situation where:

1) The URI host is `my.custom.url.com`
2) The first certificate in the chain has a CN of `my.custom.url.com`
3) The next certificate in the chain does not have a CN

In this case I did not want to overwrite the primary `my.custom.url.com.crt` certificate with the second certificate in the chain. Any better recommendations on certificate file naming are welcome.

### Issues Resolved

Fixes https://github.com/chef/chef/issues/2919

### Check List

- [x] All tests pass


danielsdeleo:
Woot!

PR #5499, opened 2016-10-31
Bump deps to pull in latest Ohai

tas50:
Full dep bump. Berks dep job fails so that one isn't bumped.


tas50:
Ping @chef/client-core

PR #5500, opened 2016-10-31
Core: improve readability of property-resource namespace collision exception message

lamont-granquist:
@tas50 here's a slightly improved error message for this case.


tas50:
This is a welcome improvement.  The messaging as it stands now is pretty confusing even for a Chef veteran.

Issue #1452, opened 2014-05-22
Allow specification of default log_location in knife.rb


bknowles:
Also prompt for default log_location when doing an interactive "knife configure".

btm:
This is fixed in https://github.com/chef/chef/issues/5502 and https://github.com/chef/knife-windows/pull/407.

The Chef changes are in 12.17.25+, with a stable release targeted for release on Monday. The knife-windows changes will be an upcoming release.

btm:
https://github.com/chef/knife-windows/pull/407 was released in knife-windows 1.8.0 today.

PR #5502, opened 2016-11-01
Knife Bootstrap: Passing config_log_level and config_log_location from config.rb

dheerajd-msys:
### Description

Passing config_log_level and config_log_location from config.rb

### Issues Resolved
https://github.com/chef/chef/issues/1452

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -69,7 +69,8 @@ class Chef
 
         def config_content
           client_rb = <<-CONFIG
-log_location     STDOUT
+log_level        :#{@chef_config[:config_log_level]}
+log_location     "#{@chef_config[:config_log_location]}"
 chef_server_url  "#{@chef_config[:chef_server_url]}"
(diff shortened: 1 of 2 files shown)


thommay:
@dheerajd-msys you'll need to update the specs that have blown up travis, but otherwise this is good, thank you!

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -69,7 +69,8 @@ def trusted_certs
 
         def config_content
           client_rb = <<-CONFIG
-log_location     STDOUT
+log_level        :#{@chef_config[:config_log_level]}
+log_location     "#{@chef_config[:config_log_location]}"
```
NimishaS: There are two possible issues that may require fix:
1. What happens if `config_log_level` and `config_log_location` are not passed? Are the values of `log_level` and `log_location` getting passed in that case?
2.  Is `config_log_location  STDOUT` getting passed successfully?

btm: We should make sure we keep the default log location of STDOUT.

PR #5503, opened 2016-11-01
update docs for attribute_changed

lamont-granquist:
add release notes for attribute_changed, also deletes the doc_changes since we're not doing that any more.


lamont-granquist:
@chef/client-core  docs

PR #5508, opened 2016-11-03
Core: Ensure that the sensitive property is correctly accessed

axos88:
### Description

sensitive is the property of the resource, not the provider.

### Issues Resolved

Fixes exceptions using converge_if_changed

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -205,7 +205,7 @@ class Chef
         specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
-          properties_str = if sensitive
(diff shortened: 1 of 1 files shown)


lamont-granquist:
this could really use a spec to not regress, needs a DCO signoff as well.

thommay:
@axos88 please sign the DCO (https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) and then we're good to go. cc @chef/client-core

axos88:
According to that doc obvious fixes don't need DCO, and this is pretty obvious, is it not? I'd rather avoid amending and force-pushing.

thommay:
obvious fix actually requires you to say that it is one, which you've now done :)

axos88:
Huh. Thought I did. Anyhoo, glad that it's okay this way.

PR #5511, opened 2016-11-03
Enable data collector w/o token for solo, but require explicit URL

danielsdeleo:
### Description

Third-party data collector implementations are supported and may not
require a token for authentication. Therefore we must allow the case
where the token is nil in Chef Solo mode. To reduce the chance that the
data collector is accidentally enabled, only set the data collector URL
to a default value in Chef Client mode.

### Issues Resolved

Acceptance is failing in the pipeline because it tests this exact case

### Check List

- [x] All tests pass

Signed-off-by: Daniel DeLeo <dan@chef.io>

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -857,7 +857,13 @@ module ChefConfig
       # Full URL to the endpoint that will receive our data. If nil, the
       # data collector will not run.
       # Ex: http://my-data-collector.mycompany.com/ingest
-      default(:server_url) { File.join(config_parent.chef_server_url, "/data-collector") }
+      default(:server_url) do
+        if config_parent.solo || config_parent.local_mode
+          nil
+        else
+          File.join(config_parent.chef_server_url, "/data-collector")
+        end
+      end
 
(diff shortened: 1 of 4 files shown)


spec/integration/client/client_spec.rb
```diff
@@ -89,7 +89,7 @@ def stop_tiny_server
 
     it "should be able to node.save with bad utf8 characters in the node data" do
       file "cookbooks/x/attributes/default.rb", 'default["badutf8"] = "Elan Ruusam\xE4e"'
-      result = shell_out("#{chef_client} -z -r 'x::default' --disable-config", :cwd => path_to(""))
+      result = shell_out("#{chef_client} -z -r 'x::default' --disable-config -ldebug", :cwd => path_to(""))
```
afiune: Do we want the logs to be in `debug` mode? or this was just for testing?

afiune:
I have a little comment but code LGTM. Lets wait for travis to pass...

spec/integration/client/client_spec.rb
danielsdeleo: You're right, nice catch. Will fix.

danielsdeleo:

PR #5515, opened 2016-11-04
Update changelog generator to better fill our pattern

tduffield:
### Description

Configure the changelog generator to better fit our current pattern of
changelogs. By utilizing the HISTORY.md file, we can generate a combined
changelog of the latest version from HISTORY.md to now. Then, when we
cut a release we can simply move our current changelog into
HISTORY.md (using the appropriate rake task) and start the process over
again.

This might even be a first step to having the Chef Versioner generate
the changelog automatically for us each time it bumps the version, and
then we just need to groom it as we go along.

### Issues Resolved

Some personal pain for those who have to artisanally craft the current CHANGELOG out of yack hair, toothpicks, bubble gum, and tears.

### Check List

- [ ] All tests pass

### DIFF
--- a/tasks/changelog.rb
+++ b/tasks/changelog.rb
@@ -1,12 +1,29 @@
 begin
   require "github_changelog_generator/task"
 
-  GitHubChangelogGenerator::RakeTask.new :changelog do |config|
-    config.issues = false
+  # Take the current changelog and move it to HISTORY.md. Should be done when
+  # cutting a release
+  task :archive_changelog do
+    changelog = File.readlines("CHANGELOG.md")
+    File.open("HISTORY.md", "w+") { |f| f.write(changelog[2..-1].join("")) }
+  end
+
+  # Run this to just update the changelog for the current release. This will
+  # take what is in History and generate a changelog of PRs between the most
+  # recent tag in HISTORY.md and HEAD.
+  GitHubChangelogGenerator::RakeTask.new :update_changelog do |config|
     config.future_release = Chef::VERSION
+    config.between_tags = ["v#{Chef::VERSION}"]
+    config.max_issues = 0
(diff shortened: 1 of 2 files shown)


tas50:
Any reason not to just make it all part of the same command in Rake?

thommay:
@tas50 i think so - you want all of 12.y to ship with one changelog, but if the changelog gets generated by the bump bot on each patch bump you'd lose all that.

tasks/changelog.rb
```diff
@@ -1,13 +1,30 @@
...
+    File.open("HISTORY.md", "w+") { |f| f.write(changelog[2..-1].join("")) }
+  end
+
+  # Run this to just update the changelog for the current release. This will
+  # take what is in History and generate a changelog of PRs between the most
+  # recent tag in HISTORY.md and HEAD.
+  GitHubChangelogGenerator::RakeTask.new :update_changelog do |config|
     config.future_release = Chef::VERSION
+    config.between_tags = ["v#{Chef::VERSION}"]
+    config.max_issues = 0
```
thommay: Does this mean we won't show any issues, and just PRs? I am +MAXINT if so :)

tas50: Looks like the workaround for disabling issues being broken

afiune:
did somebody test this? 🤔

I just ran it and it generates a `CHANGELOG-NEW.md` and I don't know what is it.. I guess we still need to move it to the actual file?

tasks/changelog.rb
tduffield: Right now it shows Closed issues, but they are grouped together and we can easily clean them out manually or (ideally) with automation. Also, I'm sure with a minor PR upstream we could fix it properly.

PR #5519, opened 2016-11-05
run the func tests on fedora, suseleap and others

lamont-granquist:
scientific linux, oracle, etc, etc...


lamont-granquist:
@chef/client-core easy one

PR #5523, opened 2016-11-07
Internal: Bump Minor Version; Automatically update the CHANGELOG when we bump the Build version

tduffield:
### Description

* Update our CI scripts to automatically update the CHANGELOG anytime we bump the version.
* Add Rake tasks for bumping major/minor versions (for use after releases)
* Bump minor version to "12.17" after "12.16" stable release.

### Issues Resolved

None

### Check List

- [x] All tests pass


lamont-granquist:
so how do we deal with patch releases to address regressions after a release?  just looking at 12.14 we had 12.14.60, 12.14.77, and 12.14.89 releases before we went to 12.15.0

most likely you'll want to have some kind of log of official release numbers that gets manually added to, and not try to trigger on the minor version bump.  there might be some other way to more automatedly pick up the correct release versions, but i haven't been able to think of it.

danielsdeleo:
👍

lamont-granquist:
^ my concerns were addressed offline -- basically we're going to branch if we need to ship regression fixes, which actually makes tons more sense to me than what we were doing before.

PR #5527, opened 2016-11-08
Add gem release badge to the readme / changelog header

tas50:
Add gem release badge to the readme / changelog header

When we update our changelog it’s a bit confusing since multiple days go by before we ship the new release. Adding a header and a readme gem release badge should help clear up what our current release is.


lamont-granquist:
only concern i've got is that you'll wind up ninja'ing tom-without-an-h

lamont-granquist:
or i guess that one is already merge conflict-ey anyway...

PR #5545, opened 2016-11-15
Add Dockerfile to root

tduffield:
### Description

This Dockerfile will allow individuals, who wish to, to build Docker
images containing the latest Chef Client. Specifically, it can be used
inside testing frameworks like kitchen-dokken.

This Dockerfile will be used by internal Chef release engineering to
automatically post Docker images to the Docker Hub as part of our
standard artifact promotion pipeline.

### Check List

- [x] All tests pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,11 @@
-FROM centos:5
+FROM busybox
 MAINTAINER Chef Software, Inc. <docker@chef.io>
 
 ARG CHANNEL=stable
 ARG VERSION=12.16.42
-ENV PATH=/opt/chef/bin:/opt/chef/embedded/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-RUN yum install -y wget && \
-    wget --content-disposition --no-check-certificate "https://packages.chef.io/files/${CHANNEL}/chef/${VERSION}/el/5/chef-${VERSION}-1.el5.x86_64.rpm" -O /tmp/chef-client.rpm && \
-    rpm -i /tmp/chef-client.rpm && \
+RUN wget "http://packages.chef.io/files/${CHANNEL}/chef/${VERSION}/el/5/chef-${VERSION}-1.el5.x86_64.rpm" -O /tmp/chef-client.rpm && \
+    rpm2cpio /tmp/chef-client.rpm | cpio -idmv && \
     rm -rf /tmp/chef-client.rpm
 
 VOLUME [ "/opt/chef" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,13 @@
-FROM ubuntu:16.04
+FROM centos:5
 MAINTAINER Chef Software, Inc. <docker@chef.io>
 
 ARG CHANNEL=stable
-ARG VERSION=latest
-ENV DEBIAN_FRONTEND=noninteractive
+ARG VERSION=12.16.42
 ENV PATH=/opt/chef/bin:/opt/chef/embedded/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-RUN apt-get update && \
-    apt-get install -y wget ssh && \
-    wget --content-disposition "https://omnitruck.chef.io/${CHANNEL}/chef/download?p=ubuntu&pv=16.04&m=x86_64&v=${VERSION}" -O /tmp/chef-client.deb && \
-    dpkg -i /tmp/chef-client.deb && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN yum install -y wget && \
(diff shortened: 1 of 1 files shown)


someara:
Can we do something like this instead?

https://github.com/someara/chef-docker-images/blob/master/recipes/default.rb

^ That unpacks the EL5 build (the oldest glibc) and shoves it into an image from SCRATCH.

That's how I built chef/chef in the past.

someara:
The Dockerfile in the PR won't work with dokken, and will only be useful for bind mounting on Ubuntu 16.04

tduffield:
Why won't it work with dokken?

tduffield:
The reason we did it this way is that it works natively with Docker. Anyone could pull it down and build their own Chef image.

someara:
For a Chef container to be useful, it needs to be bind mounted by other containers.
It's missing two critical lines:

```
ADD opt /opt
VOLUME [ '/opt/chef' ]
```

Even if those were present, the resulting bits are built against the glibc that ships with Xenial.
If you were to try and run it from a Trusty image it'd explode.

someara:
If only we had some sort of packaging technology we could use to describe dependencies and yield sparse images.... *rubs chin*

tduffield:
Hmm, I'll have to think about this. The pipeline we have right now won't really support the chef recipe.

someara:
You don't have to use the Chef recipe... that was for my own convenience.

The technique used by the Chef recipe is the important bit: 
1) Unpack the EL5 archive
2) shove into a container from scratch
3) expose `/opt/chef`

-s

tduffield:
Right, the pipeline we have won't support that stuff. It's pretty simplistic. It just runs Dockerfiles.

tduffield:
@someara what if we did this process, but on centos:5?

tduffield:
I mean, we'd be shipping centos:5 (which isn't super optimized) but would it fix the "works with dokken" issue?

someara:
I'll have to test it out... 

```
someara/chef  latest              db4fa58c4d18        5 weeks ago         150.2 MB
centos              5                    1ae98b2c895d        11 weeks ago        284.7 MB
```

someara:
I'll have to test it out...  The overhead is pretty large. This would almost triple the pull time for CI systems.

```
someara/chef  latest              db4fa58c4d18        5 weeks ago         150.2 MB
centos              5                    1ae98b2c895d        11 weeks ago        284.7 MB
```

someara:
I think the better thing to do would be to augment the build pipeline to have it spit out the optimized versions to Dockerhub as well as Omnitruck. I can't think of a way to have the Dockerhub automated build thing use FROM SCRATCH without shipping a copy of the RPM in the source repo

tduffield:
@someara the first pull would be a little longer, sure. But subsequent builds _should_ only pull down the latest layer which should be the same size as your from scratch. Also, something is better than nothing, and right now there's nothing. If people are concerned about download time / image size they can continue to use alternatives, but I think the opportunity to have the latest packages readily available automatically provides great value. Once we have something then we can work on optimizing.

someara:
Also, Dockerhub is going to build on each mater commit pushed to Github, so unless you already have the archive available at packages.chef.io, it'll fail to fetch it.

someara:
True. Warm caches are nice. However, you don't get them on Travis and friends.

tduffield:
We won't be using the automated builds from Docker Hub. We're using an internal build system that is triggered by events in artifactory. That build system, at the moment, isn't built to support the "from scratch" method you've mentioned. It would require work, which I'm not against doing, but at the moment I'd rather hook this up to that build system and move from there rather than continue to have nothing while we go back and pretty much rewrite the build system from scratch.

someara:
PS: kitchen-dokken will continue to use the optimized version published at `someara/chef`

tduffield:
That's fine. This is simply an alternative.

tduffield:
FWIW I ran through a series of kitchen-dokken tests using the wheezy-slim image, one of those tests being centos:5, and Chef ran just fine :)

someara:
I just did a quick spot check by against the httpd cookbook's config suites.

It works on all platforms except EL5.

```
/opt/chef/embedded/bin/ruby: /lib64/libc.so.6: version `GLIBC_2.8' not found (required by /opt/chef/embedded/lib/libruby.so.2.3)
/opt/chef/embedded/bin/ruby: /lib64/libc.so.6: version `GLIBC_2.6' not found (required by /opt/chef/embedded/lib/libruby.so.2.3)
/opt/chef/embedded/bin/ruby: /lib64/libc.so.6: version `GLIBC_2.9' not found (required by /opt/chef/embedded/lib/libruby.so.2.3)
```

tduffield:
@someara I _think_ that should work. Can you confirm?

tduffield:
I've built a test image using this Dockerfile. If you want to test it, you can use `tduffield/chef:test1`.

btm:
I have no docker opinions. 👍 on whatever others want to do.

PR #5548, opened 2016-11-15
Chefstyle updates

lamont-granquist:
These are new proposed cops

### DIFF
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -76,10 +76,10 @@ class Chef
     class Group < RuntimeError; end
     class Link < RuntimeError; end
     class Mount < RuntimeError; end
-    class Reboot < Exception; end
-    class RebootPending < Exception; end
+    class Reboot < Exception; end # rubocop:disable Lint/InheritException
+    class RebootPending < Exception; end # rubocop:disable Lint/InheritException
     class RebootFailed < Mixlib::ShellOut::ShellCommandFailed; end
-    class ClientUpgraded < Exception; end
+    class ClientUpgraded < Exception; end # rubocop:disable Lint/InheritException
     class PrivateKeyMissing < RuntimeError; end
     class CannotWritePrivateKey < RuntimeError; end
     class RoleNotFound < RuntimeError; end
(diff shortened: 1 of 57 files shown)


lamont-granquist:
@chef/client-core thoughts and feels?

thommay:
I dislike EmptyCaseCondition but don't really care.

I find MultilineArrayBraceLayout a step backwards in terms of readability.

lamont-granquist:
so @thommay the default for MultilineArrayBraceLayout is 'symmetrical' and supports either of these:

```ruby
# symmetrical: good
# new_line: bad
# same_line: good
[ :a,
  :b ]

# symmetrical: good
# new_line: good
# same_line: bad
[
  :a,
  :b
]
```

i think the problem you're seeing is that you don't like that the opening brace did not start on its own line in those cases.

thommay:
ok, I mean the first example is awful (which is what you had in this PR)

lamont-granquist:
right but that's a function of the first line being written incorrectly, MultilineArrayBraceLayout is behaving well, we just need to get around to the cop that will fix the first line.  if you look at the code it is fixing it is worse:

```
[ :a,
  :b,
]
```

that is something nobody wants.

lamont-granquist:
and I think I wrote both of the code fragments that EmptyCaseCondition caught.  i'm real tempted to drop that one.

lib/chef/exceptions.rb
```diff
@@ -76,10 +76,10 @@ class User < RuntimeError; end
     class Group < RuntimeError; end
     class Link < RuntimeError; end
     class Mount < RuntimeError; end
-    class Reboot < Exception; end
```
danielsdeleo: Seems like we may have made this inherit from `Exception` intentionally?

danielsdeleo:
I'm not sure we want to auto-fix the `InheritException` ones. In that case I'm okay with enabling it and using the magic comments to disable for specific cases where we know it's what we want.

lamont-granquist:
EmptyCaseCondition gone
MultilineArrayBraceLayout gone (too much trouble)

lamont-granquist:
@chef/client-core okay fixes applied i think

danielsdeleo:
Looks ok to me now

Issue #4045, opened 2015-10-07
powershell module does not pass through output


mwrock:
If using knife via `chef.psm1` out put cannot be captured via stream redirection or variable capture. It always leaks to the console.

```
knife node list > myfile.txt
knife node list | more
$out = knife node list
```

None of these work and the first just creates an empty file.

smurawski:
@mwrock is this only on wmf5 or have you tested other wmf versions?  I ask because WMF5 introduced a new stream (information) which parallels info out to the host.

mwrock:
Just tested on my win 10 and not sure what Jason was testing on. I don't think this is the result of the information stream because but the way we are shelling out from the module to ruby.

smurawski:
@mwrock I just confirmed with jason - 2012 R2 with wmf 5

james-s-nduka:
Is there any updates to this or workarounds? I have hit a roadblock due to this issue.

mwrock:
The workaround is not to use the module.

btm:
This should be fixed in 12.17.49+

PR #5549, opened 2016-11-15
PowerShell module output pass through

coolmacool:
### Description

The PowerShell module functions, Load-Win32Bindings & Run-ExecutableAndWait, were modified to utilize an anonymous pipe for writing stdout and stderr of the spawned child process, which provides the parent PowerShell process the ability to read from the pipe and then pass through the output in the normal way.

### Issues Resolved

This commit fixes [#4045](https://github.com/chef/chef/issues/4045) where the output from any of the chef executables wrapped by the module functions would not allow redirection or variable capture.

### DIFF
--- a/distro/powershell/chef/chef.psm1
+++ b/distro/powershell/chef/chef.psm1
@@ -109,6 +109,12 @@ public enum StandardHandle : int
   Error = -12
 }
 
+public enum HandleFlags : int
+{
+  HANDLE_FLAG_INHERIT = 0x00000001,
+  HANDLE_FLAG_PROTECT_FROM_CLOSE = 0x00000002
+}
+
 public static class Kernel32
 {
   [DllImport("kernel32.dll", SetLastError=true)]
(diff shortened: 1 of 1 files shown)


btm:
@chef/client-windows this is good stuff, review please?

ksubrama:
This looks great to me.  Perhaps a comment describing the caveats?  I see both stdout and stderr merged into one pipe and then the entire pipe is read into memory.  I think it fine - I don't expect a lot of scenarios where a chef tools output is going to cause any significant memory pressure but it is worth documenting in case someone somewhere has ohai running with all the bells and whistles turned on and generated a boat load of output only to note the powershell process terminating or something weird like that.

coolmacool:
Thanks guys! Let me know if I should still add the comments regarding the caveats with this solution. I am happy to help with this, or any of the other open issues.

btm:
@coolmacool there are a number of triaged windows issues that can always use help, if that's a thing you're into: https://github.com/chef/chef/labels/Windows

PR #5550, opened 2016-11-16
Remove debug write

atward:
### Description

Removes debug write introduced in previous commit: e57e29b96b75ff7fc04abb7f9db73920b2a5894a

### Issues Resolved

Removes erroneous output during compilation phase, eg:
```
Compiling Cookbooks...
python
platform
python
platform
platform
```

Obvious fix.


coderanger:
Merging before tests because oops. Thanks for catching this.

lamont-granquist:
i should really setup a precommit hook that looks for "puts"

PR #5551, opened 2016-11-16
smf_recursive_dependencies: Allow solaris services to start recursively.

MarkGibbons:
### Description

I have several cases where I would like to start solaris services with the -sr option rather than -s. -r will start services recursively.  I have cookbooks where the service enable command fails because the services dependencies have not been met.   The problem resolved is pretty minor, the second chef run usually picks up the service start up issue.  I'd rather get things started with one run however.

### Issues Resolved

None reported.

### Check List

- [ ] All tests pass

Svcadm enable -s service only starts the named service. This commit adds the
recursive option to start service dependencies recursively.  The resulting command
"svcadm enable -sr service"  will try to start service dependencies before
starting the named service.


lib/chef/provider/service/solaris.rb
```diff
@@ -56,7 +56,8 @@ def define_resource_requirements
 
         def enable_service
           shell_out!(default_init_command, "clear", @new_resource.service_name) if @maintenance
-          shell_out!(default_init_command, "enable", "-s", @new_resource.service_name)
+          start_flags = @new_resource.recursive ? "-sr" : "-s"
+          shell_out!(default_init_command, "enable", start_flags , @new_resource.service_name)
```
lamont-granquist: can you just pass `@new_resource.options` here directly instead so that arbitrary options can be passed -- same presumably with other actions.

lib/chef/resource/service.rb
```diff
@@ -212,6 +213,14 @@ def supports(args = {})
         end
       end
 
+      def options(arg = nil)
+        set_or_return(
+          :user,
```
thommay: This should be `:options`, not `:user`

lamont-granquist:
don't think this works, if you use the argv form of shell_out! you need to have everything split:

```
[6] pry(main)> shell_out!("/bin/ls", "-l", "-a").run_command.stdout
[...works...]
[7] pry(main)> shell_out!("/bin/ls", "-l -a").run_command.stdout
Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
---- Begin output of ["/bin/ls", "-l -a"] ----
STDOUT:
STDERR: /bin/ls: illegal option --
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
---- End output of ["/bin/ls", "-l -a"] ----
Ran ["/bin/ls", "-l -a"] returned 1
from /Users/lamont/.rvm/gems/ruby-2.3.3/gems/mixlib-shellout-2.2.7-universal-mingw32/lib/mixlib/shellout.rb:289:in `invalid!'
```

so this line is invalid:

```
 expect(@provider).to receive(:shell_out!).with("/usr/sbin/svcadm", "enable", "-s -r", @current_resource.service_name).and_return(@success)
```

works because you stubbed it, but if it wasn't a stubbed unit test that would fail.

lib/chef/resource/service.rb
```diff
@@ -49,6 +49,7 @@ def initialize(name, run_context = nil)
         @run_levels = nil
         @user = nil
         @supports = { :restart => nil, :reload => nil, :status => nil }
+        @recursive = nil
```
thommay: This should be `@options` now.

spec/unit/resource/service_spec.rb
```diff
@@ -124,6 +124,22 @@
     end.to raise_error(ArgumentError)
   end
 
+  it "should accept true for recursive" do
+    @resource.recursive true
```
thommay: These should all be `@resource.options`, and you're expecting them to take an Array or a String, rather than a bool.

MarkGibbons:
I fixed the problems with the shell_out options.  Renaming recursive to options should also be complete.  Thank you for your patience.  I built the chef client from the changed code and the resulting gem works in real life.

thanks,
Mark

MarkGibbons:
Added DCO

thommay:
@lamont-granquist re-review?

lib/chef/resource/service.rb
```diff
@@ -152,6 +153,14 @@ def masked(arg = nil)
         )
       end
 
+      def options(arg = nil)
+        set_or_return(
+          :options,
+          arg.respond_to?(:split) ? arg.split(/\s+/) : arg,
```
lamont-granquist: should use `#shellsplit` i think from `shellwords` module here:

https://github.com/chef/chef/pull/5838/files

MarkGibbons: Working on it.  Thanks

Issue #5533, opened 2016-11-10
several unit tests fail to run when workstation is FIPS enabled


ghatwala:
## Description
Getting test case failures.
```
# Run All the Tests 
bundle exec rake spec
```

While tying out the test cases , getting below errors
```
Finished in 13 minutes 15 seconds (files took 12.58 seconds to load)
16974 examples, 3 failures, 31 pending

Failed examples:

rspec ./spec/unit/application/client_spec.rb:431 # Chef::Application::Client configure_chef should set the colored output to true by default on windows and true on all other platforms as well
rspec ./spec/unit/application_spec.rb:134 # Chef::Application when there are no configuration errors configure_chef when a config_file is present should configure chef::config from a file
rspec ./spec/unit/application_spec.rb:139 # Chef::Application when there are no configuration errors configure_chef when a config_file is present should merge the local config hash into chef::config

```

## Chef Version
```
jenkins@1e4fff739f96:~/chef> chef-solo --version
Chef: 12.17.1

```
## Platform Version
```
jenkins@1e4fff739f96:~/chef> arch
ppc64le
```
```
jenkins@1e4fff739f96:~/chef> lscpu
Architecture:          ppc64le
Byte Order:            Little Endian
CPU(s):                16
On-line CPU(s) list:   0-15
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             16
NUMA node(s):          1
Model:                 IBM pSeries (emulated by qemu)
Hypervisor vendor:     KVM
Virtualization type:   full
L1d cache:             64K
L1i cache:             32K
NUMA node0 CPU(s):     0-15

```
Bundle version
```
jenkins@1e4fff739f96:~/chef> bundle --version
Bundler version 1.13.6

```
```
jenkins@1e4fff739f96:~/chef> ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [powerpc64le-linux]
```
gem version
```
jenkins@1e4fff739f96:~/chef> gem --version
2.6.8
```

chef client
```
jenkins@1e4fff739f96:~/chef> chef-client --version
Chef: 12.17.1

```

OS version
```
jenkins@1e4fff739f96:~/chef> cat /etc/os-release
NAME="SLES"
VERSION="12-SP1"
VERSION_ID="12.1"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP1"
ID="sles"
ANSI_COLOR="0;32"
[...]

PR #5559, opened 2016-11-17
Test case execution incase FIPS  exists

ghatwala:
### Description

As requested in the issue to raise a PR to edit the readme so that the test cases pass on any linux box which has FIPS enabled.

chef#5533

@thommay - please kindly review the PR.

### Issues Resolved
All test cases pass now in a linux box which has FIPS enabled once path has been exported .

Test cases were failing as mentioned in issue : https://github.com/chef/chef/issues/5533

### Check List
All tests pass

Signed-off-by: Amit Ghatwal ghatwala@us.ibm.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ We use RSpec for unit/spec tests. It is not necessary to start the development
 environment to run the specs--they are completely standalone.
 
 ```bash
+#Ensuring for any linux box which has FIPS enabled , the chef test cases will pass using below:
+export CHEF_FIPS=""
+
 # Run All the Tests
 bundle exec rake spec
 


ghatwala:
Signed-off-by: Amit Ghatwal <ghatwala@us.ibm.com>

lamont-granquist:
I'm not sure that's really the right way to go about fixing this.

Its very one-off and its likely that not very many people will have FIPS enabled and will hit this (certainly you're the first).  For 99% of the people reading that it is just additional cognitive overhead.

Really the specs should get fixed so that they just-work when FIPS is enabled on the box.

I'd actually kind of prefer just to close both issues, and let someone who google searches find the issue and find the workaround, rather than formally documenting the workaround.

ghatwala:
Hi @lamont-granquist  ,

I had raised this PR on the comment given by @thommay  ( chef contributor ) 

**@ghatwala we'd be happy to accept a PR that adds that.** 

@thommay  -  could u comment on the above if this PR is needed  ?

shortdudey123:
> Really the specs should get fixed so that they just-work when FIPS is enabled on the box.

Completely agree with this part, working around spec failures is never really a good idea

ghatwala:
@shortdudey123, @lamont-granquist , @thommay  - I understand that working around specs failures isn't a good way to resolve the test case failures issue .

Are we planning to looking at failing test cases on a linux box where FIPS is enabled , if yes , do let me know if u guys need any more information from my end ?

Thanks for considering this issue.

thommay:
After reflection, I agree with Lamont; let's just close this and let google turn up the issue. We clearly need to redo the specs to cope with FIPS mode, but that's a rather larger kettle of fish.
@ghatwala thanks for your patience with this, and thanks for the PR!

ghatwala:
@thommay : Thanks for concluding no this for now . Please do include me in future threads about FIPS mode.

PR #5561, opened 2016-11-18
check to see if a unit is-active after starting

kisoku:
With certain types of systemd units we cannot depend systemctl's
exit code to determine if the unit is active.

Use systemctl is-active to explicitly check if the unit has started

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -115,6 +115,8 @@ class Chef
         unless current_resource.active
           converge_by("starting unit: #{new_resource.name}") do
             systemctl_execute!(:start, new_resource.name)
+            sleep 0.5
+            systemctl_execute!("is-active", new_resource.name)
           end
         end
       end
@@ -130,6 +132,8 @@ class Chef
       def action_restart
         converge_by("restarting unit: #{new_resource.name}") do
           systemctl_execute!(:restart, new_resource.name)
+          sleep 0.5
+          systemctl_execute!("is-active", new_resource.name)
         end
       end
 


coderanger:
Wouldn't that fail with a one-shot that runs for less than 500ms?

kisoku:
You are correct. Perhaps this check should be flagged using a resource property ?

stevendanna:
If we are going to add something like this, I would like to see the commit message and perhaps a code comment which outlines what kind of units require this follow up check and the error that would be raised if this weren't in place.  Also, is there a test we could add for the problematic unit types?

coderanger:
So I think this reflects an overall issue with how we handle services (and service-like things). The problem is that we generally define the success or failure of the `:start`/`:stop`/`:restart`/`:reload` actions in terms of the underlying system's command for that action. In some cases, the `start` command can succeed even though the service crashes immediately. The question is does that count as a valid service start? The underlying system did accept a start command and claim that it started a thing, that thing just happened to die immediately. We could definitely try to unify this behavior but Chef has generally chosen to have core resources map as closely as possible to the underlying semantics.

coderanger:
Also to more fully define the problem, imagine a service consisting only of `/bin/false` or of `/bin/bash -c "sleep 10 && false"`. What should be the behavior of each of those under systemd, Upstart, etc?

coderanger:
This patch would make the first fail the converge but the second would not. I would say that the first one being an error is more user friendly but its a hard line to draw. Also this behavior would have very limited use outside of service units so definitely needs to be behind either an option or a new action or something else.

kisoku:
I'm not arguing that chef should be used as a monitoring tool. In my case chef is considering python processes that have been passed bad command line arguments as converged and running. I'm just looking for similar sematics that I get from using init based startup scripts or runit_service out of systemd_unit

lamont-granquist:
yeah the pattern of `sleep 0.5 and then check` is going to be brittle in general, and i don't see any way around the inherent asynchronousness of systemd units, and i think i concur that chef isn't a monitoring tool.

the services need to be monitoring, if they're scripts then tests need to get written and test-kitchen'd or something, or the python scripts could just be moved into chef itself, etc.  lots of different ways to skin this particular cat, but this seems like the wrong way to go about it.  i don't think chef should be responsible for this.

also you can roll this yourself with a ruby_block if you need to.  in the general case, though, this is going to cause more problems than it solves.

PR #5585, opened 2016-11-26
Add warning in knife node from file

jeunito:
### Description
Unlike ```knife environment from file``` since this functionality doesn't update, 
there should be a confirmation warning before overriding a node in knife node from file if the node exists.

### DIFF
--- a/lib/chef/knife/node_from_file.rb
+++ b/lib/chef/knife/node_from_file.rb
@@ -36,6 +36,15 @@ class Chef
 
       def run
         @name_args.each do |arg|
+          node_name = File.basename(arg.gsub(/\.(json|rb)/, ""))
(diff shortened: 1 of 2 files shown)


coderanger:
Can you be more specific about what you think is or isn't happening? All of the `from file` commands are pretty literally just `PUT`ing whatever data you give which does a full object overwrite.

jeunito:
@coderanger I could have sworn that at more than one time in past after running `node from file` on a node that was existing, all the node attributes outside of the template were erased and the chef-client runs were failing.

In replying to your question, I tried to replicate again the issue I saw on a different type of server but I am not able to anymore. 

I am closing this PR for now and will open it again if and when I encounter the issue.

PR #5590, opened 2016-11-28
Update Data Collector to use Chef::JSONCompat

adamleff:
Data Collector uses the json gem to read/write the metadata file
in the file cache. Let's change that to use Chef::JSONCompat instead.


lamont-granquist:
👍

Issue #5594, opened 2016-11-29
reboot! does not work well with Tomcat


Xoph:
## Description

I have Tomcat 8 installed. When Tomcat is installed, it is added at the beginning of the PATH environment variable, so the 'shutdown...' command-line in the reboot! method of the rebooter.rb module uses shutdown.bat in C:\Program Files\Tomcat-8\bin\ instead of the built-in Windows shutdown command.

My recipes update the version of Java installed, and therefore set the JRE_HOME environment variable, so using the reboot resource returns an error because the 'shutdown...' command-line is still using the old value of the JRE_HOME environment variable (which references a location that no longer exists) when running the reboot.

Synopsis: If the PATH environment variable contains a location with a shutdown.exe or shutdown.bat file that is evaluated before C:\Windows\System32, that file will be used and it can cause an issue.

Solution: Use a full path depending on OS architecture.

Additional funkiness comes into play with this issue as well, because shutdown.bat for Tomcat does not support /c so a message cannot be specified to the user.

## Chef Version

12.13.37

## Platform Version

Windows Server 2012 R2

## Replication Case

1. Create a file C:\Temp\shutdown.bat that contains the text 'exit 1'.
2. Add C:\Temp\ as the first item in the list in the PATH environment variable.
3. Create a recipe that uses the reboot resource.
4. Run Chef for the recipe and see it fail.

## Client Output

Chef Client finished, 50/105 resources updated in 10 minutes 19 seconds
  3/3 controls succeeded
[2016-11-29T13:31:01-05:00] WARN: Rebooting server at a recipe's request. Details: {:delay_mins=>1, :reason=>"Rebooting
the server!.", :tim
estamp=>2016-11-29 13:26:06 -0500, :requested_by=>"Reboot the server at the end of the run"}

Running handlers:
[2016-11-29T13:31:01-05:00] ERROR: Running exception handlers
Running handlers complete
[2016-11-29T13:31:01-05:00] ERROR: Exception handlers complete
Chef Client failed. 50 resources updated in 10 minutes 19 seconds
  3 controls succeeded
[2016-11-29T13:31:01-05:00] FATAL: Stacktrace dumped to c:/support/radiologyservicetools/installautomation/local-mode-ca
che/cache/chef-stacktrace.out
[2016-11-29T13:31:01-05:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[...]

PR #5596, opened 2016-11-29
Ensures correct version of shutdown is called when using the reboot resource on Windows

Xoph:
### Description

Use 'C:/Windows/System32/shutdown.exe...' instead of 'shutdown...' when rebooting.

### Issues Resolved

https://github.com/chef/chef/issues/5594

### Check List

- [X] All tests pass
- [N/A] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)

### DIFF
--- a/lib/chef/platform/rebooter.rb
+++ b/lib/chef/platform/rebooter.rb
@@ -36,7 +36,7 @@ class Chef
           cmd = if Chef::Platform.windows?
                   # should this do /f as well? do we then need a minimum delay to let apps quit?
                   # Use explicit path to shutdown.exe, to protect against https://github.com/chef/chef/issues/5594
-                  windows_shutdown_path = "#{ENV['SystemDrive']}/Windows/System32/shutdown.exe"
+                  windows_shutdown_path = "#{ENV['SYSTEMROOT']}/System32/shutdown.exe"
                   "#{windows_shutdown_path} /r /t #{reboot_info[:delay_mins] * 60} /c \"#{reboot_info[:reason]}\""
                 else
                   # probably Linux-only.
--- a/lib/chef/platform/rebooter.rb
+++ b/lib/chef/platform/rebooter.rb
@@ -36,7 +36,7 @@ class Chef
           cmd = if Chef::Platform.windows?
                   # should this do /f as well? do we then need a minimum delay to let apps quit?
                   # Use explicit path to shutdown.exe, to protect against https://github.com/chef/chef/issues/5594
-                  windows_shutdown_path = 'C:/Windows/System32/shutdown.exe'
+                  windows_shutdown_path = "#{ENV['SystemDrive']}/Windows/System32/shutdown.exe"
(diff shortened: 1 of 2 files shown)


lib/chef/platform/rebooter.rb
```diff
@@ -35,7 +35,9 @@ def reboot!(node)
 
           cmd = if Chef::Platform.windows?
                   # should this do /f as well? do we then need a minimum delay to let apps quit?
-                  "shutdown /r /t #{reboot_info[:delay_mins] * 60} /c \"#{reboot_info[:reason]}\""
+                  # Use explicit path to shutdown.exe, to protect against https://github.com/chef/chef/issues/5594
+                  windows_shutdown_path = 'C:/Windows/System32/shutdown.exe'
```
tas50: This will break if the system is installed on a non-C drive.

lamont-granquist: or yeah not, we probably need a patch that contains more `%`s like `%SYSTEMROOT%` or whatever...

coderanger: I don't think so (re: `Which`), the original bug was because there was something else higher up on the path that happens to be called `shutdown`. 👍 that this should be using `ENV['SystemDrive']` or whatever it is now though.

Xoph:
Not sure what's exactly failing with the build.

lib/chef/platform/rebooter.rb
```diff
@@ -35,7 +35,9 @@ def reboot!(node)
 
           cmd = if Chef::Platform.windows?
                   # should this do /f as well? do we then need a minimum delay to let apps quit?
-                  "shutdown /r /t #{reboot_info[:delay_mins] * 60} /c \"#{reboot_info[:reason]}\""
+                  # Use explicit path to shutdown.exe, to protect against https://github.com/chef/chef/issues/5594
+                  windows_shutdown_path = "#{ENV['SystemDrive']}/Windows/System32/shutdown.exe"
```
btm: Can you use `#{ENV['SYSTEMROOT']/system32` instead? This would ensure this works if windows is installed to a directory named something other than 'windows'.

spec/functional/rebooter_spec.rb
```diff
@@ -43,7 +43,7 @@ def create_resource
 
   let(:expected) do
     {
-      :windows => 'shutdown /r /t 300 /c "rebooter spec test"',
+      :windows => "#{ENV['SystemDrive']}/Windows/System32/shutdown.exe /r /t 300 /c \"rebooter spec test\"",
```
btm: I guess the test that uses this will weirdly work on non-windows platforms because the variable will be empty in both cases, so, maybe that's okay, but I'd really expect the test below to be filtered using `:windows_only`

btm:
@Xoph thanks. in the future it'd be great to do an interactive rebase against a branch like this to squash all those commits into a single commit. After doing so you have to force push your branch since your modifying the git history.

Xoph:
Honestly this was my first change with a couple files on Git so I didn't know much of what I was doing. I gotta get up on Git, I use Perforce for work.

PR #5598, opened 2016-11-30
Remove Maintainer subsections

thommay:
Besides the platfrom ones, which require some touch up to the
maintenance policy.

Per the developers' meeting last week, the subsections are not really working, so let's get rid of them.

cc @adamhjk @chef/maintainers


adamhjk:
I'm the decider, and I approve this. Go for it!

PR #5601, opened 2016-11-30
Core: Ensure chef-solo creates node files w/ correct permissions

scottopherson:
### Description

This commit ensures that the `nodes` dir and the node files within
it are created with the correct permissions by chef-solo.

### Issues Resolved

Internal ticket

### Check List

- [x] All tests pass

cc @chef/client-core @chef/client-windows

### DIFF
--- a/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
+++ b/lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
@@ -44,6 +44,7 @@ require "chef/chef_fs/data_handler/role_data_handler"
 require "chef/chef_fs/data_handler/user_data_handler"
 require "chef/chef_fs/data_handler/group_data_handler"
 require "chef/chef_fs/data_handler/container_data_handler"
+require "chef/win32/security" if Chef::Platform.windows?
 
 class Chef
   module ChefFS
@@ -109,7 +110,17 @@ class Chef
             else
               child_paths[name].each do |path|
                 begin
-                  Dir.mkdir(path)
+                  Dir.mkdir(path, 0700)
+                  if Chef::Platform.windows?
+                    all_mask = Chef::ReservedNames::Win32::API::Security::GENERIC_ALL
+                    owner = Chef::ReservedNames::Win32::Security::SID.current_user
+                    dacl = Chef::ReservedNames::Win32::Security::ACL.create([
+                      Chef::ReservedNames::Win32::Security::ACE.access_allowed(owner, all_mask),
+                    ])
+                    so = Chef::ReservedNames::Win32::Security::SecurableObject.new(path)
+                    so.owner = owner
+                    so.set_dacl(dacl, false)
+                  end
                 rescue Errno::EEXIST
                 end
               end
--- a/lib/chef/chef_fs/file_system/repository/nodes_dir.rb
+++ b/lib/chef/chef_fs/file_system/repository/nodes_dir.rb
@@ -20,6 +20,7 @@
 require "chef/chef_fs/file_system/repository/node"
 require "chef/chef_fs/file_system/repository/directory"
 require "chef/chef_fs/file_system/exceptions"
+require "chef/win32/security" if Chef::Platform.windows?
 
 class Chef
   module ChefFS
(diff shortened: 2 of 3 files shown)


lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
```diff
@@ -110,6 +110,7 @@ def create_child(name, file_contents = nil)
               child_paths[name].each do |path|
                 begin
                   Dir.mkdir(path)
+                  File.chmod(0700, path) if name == "nodes"
```
thommay: is there any reason not to do this for everything?

stevendanna: If we are going to do it for everything, I believe `mkdir` takes a second argument where you can set the mode on creation.

scottopherson: It also just hit me that these permission changes only work in unix land so I gotta go talk to some Windows folks.

lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
```diff
@@ -44,6 +44,7 @@
 require "chef/chef_fs/data_handler/user_data_handler"
 require "chef/chef_fs/data_handler/group_data_handler"
 require "chef/chef_fs/data_handler/container_data_handler"
+require "chef"
```
lamont-granquist: this is most likely going to create dependency loops where chef requires chef_fs which requires chef.

i'd prefer to go back to simpler ruby code than trying to re-use chef provider code here, it's very likely to have unintended side effects.

lamont-granquist: and I'm even :-1: on trying to require the resource+provider code directly and bypassing the mess that `require "chef"` will pull in and be more targetted.  we simply haven't architected resources+providers to be that reusable.

lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
```diff
@@ -109,7 +110,14 @@ def create_child(name, file_contents = nil)
             else
               child_paths[name].each do |path|
                 begin
-                  Dir.mkdir(path)
+                  events = Chef::EventDispatch::Dispatcher.new
+                  node = Chef::Node.new
+                  run_context = Chef::RunContext.new(node, {}, events)
+                  dir = Chef::Resource::Directory.new(path, run_context)
```
lamont-granquist: and all of this framework to build a resource is an indication that these objects aren't well constructed to be reusable like this, so shouldn't be used here.

coderanger: We should really make an imperative API for file perms because I've done similar atrocities for the same reasons in my own code :(

lamont-granquist: indeed, it would be nice to have that kind of API, but we don't.  its easier for you to commit these kinds of atrocities because chef doesn't have direct deps on your code, so all the deps run one direction -- but here it runs both ways and that will only lead to loops and great sadness..

lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
scottopherson: I'd prefer simpler ruby code as well but doing file permissions for windows, without using file/dir providers, will likely be more complicated that what's here. From my chats with Windows people doing file permissions on Windows from Ruby is a pain.

Would you be opposed to using using `Dir.mkdir`/`File.open` as before but then leveraging https://github.com/chef/chef/blob/master/lib/chef/win32/security/securable_object.rb directly to handle the Windows perms?

lamont-granquist: yeah the chef/win32 stuff should be sufficiently low level to not wind up in dep loop hell

scottopherson: Cool cool I'll give that a shot, thanks.

lamont-granquist:
travis seems angry, but this is a much better direction to head for now 👍

scottopherson:
Oh derp, I made rubocop angry.

lib/chef/chef_fs/file_system/repository/nodes_dir.rb
```diff
@@ -30,6 +31,24 @@ class NodesDir < Repository::Directory
           def make_child_entry(child_name)
             Node.new(child_name, self)
           end
+
+          def create_child(child_name, file_contents = nil)
+            child = super
+            File.chmod(0600, child.file_path)
```
tduffield: Should this be in an `else` statement to the windows check below?

scottopherson: I elected to skip an `else` since `File.chmod` is a no-op on Windows. I'm fine either way though if we want to put in in an else.

lib/chef/chef_fs/file_system/repository/nodes_dir.rb
```diff
@@ -30,6 +31,24 @@ class NodesDir < Repository::Directory
...
+            if Chef::Platform.windows?
+              read_mask = Chef::ReservedNames::Win32::API::Security::GENERIC_READ
+              write_mask = Chef::ReservedNames::Win32::API::Security::GENERIC_WRITE
```
btm: I think you want FILE_GENERIC_READ and FILE_GENERIC_WRITE here, but I'm out the door. I can look after dinner.

PR #5602, opened 2016-12-01
Modify `knife configure` example `chef_server_url`

jerryaldrichiii:
This modifies the example given for `chef_server_url` in `knife configure`. A proper `chef_server_url` in `knife.rb` must contain a path to an organization.


jerryaldrichiii:
Tests updated @thommay! I also updated the man doc. Let me know if I should change anything else.

PR #5604, opened 2016-12-01
Use object ID when detected unprocessed Resources

### LINKED ISSUE #5565: TypeError: no implicit conversion of nil into Integer in detect_unprocessed_resources
## Description

Chef run crashes during detect_unprocessed_resources with TypeError: no implicit conversion of nil into Integer

## Chef Version

12.16.42-1

## Platform Version

Ubuntu 14.04

## Replication Case

Dunno

## Client Output

```Recipe: role_build::_nginx
  * aws_s3_file[/etc/ssl/certs/build.piazza.com.crt] action create
    * remote_file[/etc/ssl/certs/build.piazza.com.crt] action create (skipped due to not_if)
     (up to date)
  * aws_s3_file[/etc/ssl/private/build.piazza.com.key] action create
    * remote_file[/etc/ssl/private/build.piazza.com.key] action create (skipped due to not_if)
     (up to date)
  * template[/etc/nginx/sites-available/build] action create (up to date)
  * execute[nxensite build] action run (skipped due to not_if)
  * execute[nxdissite default] action run (skipped due to only_if)

Running handlers:
[2016-11-18T12:06:43-08:00] ERROR: Running exception handlers
Running handlers complete
[2016-11-18T12:06:43-08:00] ERROR: Exception handlers complete
Chef Client failed. 15 resources updated in 11 seconds
[2016-11-18T12:06:43-08:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2016-11-18T12:06:43-08:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2016-11-18T12:06:43-08:00] ERROR: no implicit conversion of nil into Integer
[2016-11-18T12:06:43-08:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

## Stacktrace

[chef-stacktrace.out](https://gist.github.com/kpmueller/7c1d5e71fc252770bc81d28b29cd3661)

adamleff:
In the Data Collector, when detecting unprocessed resources, a Set
was built up using the resource object itself as the Set element.
Internally, Set adds this to a Hash.

We allow users to create custom resources that could contain a property
named "hash" which in turn wires up a `#hash` instance method on the
Resource. Ruby expects object's `#hash` method to return a Fixnum that
it uses internally. So if a resource had a "hash" property that returned
a String, bad things happened.

With this change, we make our own Hash and use the resource's object ID
in the key so we don't have to worry about the resource's `#hash` method
getting called and throwing an exception. I will send up a separate
change that warns users when they choose a property name that is already
an existing method name.

Fixes #5565.

### DIFF
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -430,26 +430,33 @@ class Chef
       end
 
       def detect_unprocessed_resources
-        # create a Set containing all resource+action combinations from
-        # the Resource Collection
-        collection_resources = Set.new
+        # create a Hash (for performance reasons, rather than an Array) containing all
+        # resource+action combinations from the Resource Collection
+        #
+        # We use the object ID instead of the resource itself in the Hash key because
+        # we currently allow users to create a property called "hash" which creates
+        # a #hash instance method on the resource. Ruby expects that to be a Fixnum,
+        # so bad things happen when adding an object to an Array or a Hash if it's not.
+        collection_resources = {}
         run_context.resource_collection.all_resources.each do |resource|
           Array(resource.action).each do |action|
-            collection_resources.add([resource, action])
+            collection_resources[[resource.__id__, action]] = resource
           end
         end
 
-        # Delete from the Set any resource+action combination we have
+        # Delete from the Hash any resource+action combination we have
         # already processed.
         all_resource_reports.each do |report|
-          collection_resources.delete([report.resource, report.action])
+          collection_resources.delete([report.resource.__id__, report.action])
         end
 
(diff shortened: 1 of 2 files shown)


danielsdeleo:
Can you add a test that specifically addresses the `#hash` misbehavior? Rest of the code LGTM.

/cc @chef/client-core

adamleff:
@danielsdeleo done! Thanks for the suggestion.

btm:
Appveyor failure is unrelated, I believe https://github.com/chef/chef/pull/5627 fixes it.

PR #5606, opened 2016-12-02
Deprecate creating properties whose names are already methods

adamleff:
When creating a resource, a user can create a property that is the same
name as an already-existing Ruby method, such as `#hash`. In the case of
the `#hash` method, this can cause issues when attempting to adding
resources to other data structures, such as Arrays or Hashes. In other
examples, this could cause unexpected behavior that is incredibly
difficult to troubleshoot.

This change adds a deprecation warning in the case where a user adds
a property to a resource that the resource instance already responds to.
If y'all are OK with this approach, I'll be happy to write up the
deprecation doc for this for docs.chef.io.

In addition, three resources were found to be redefining methods in this manner: `apt_repository`, `yum_repository`, and `launchd`.  `apt_repository` and `yum_repository` were redefining the `sensitive` property which is already available in `Chef::Resource`, so these property definitions were superfluous.

The `launchd` resource was defining a property named `hash` which is already an instance method in `Object`. This change also renames that property to `plist_hash` and deprecates the use of the `hash` property.

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -156,6 +156,26 @@ class Chef
       end
     end
 
+    class PropertyNameCollision < Base
+      def id
+        11
+      end
+
+      def target
+        "property_name_collision.html"
+      end
+    end
+
+    class LaunchdHashProperty < Base
+      def id
+        12
+      end
+
+      def target
+        "launchd_hash_property.html"
+      end
+    end
+
     class ChefPlatformMethods < Base
       def id
         13
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -518,6 +518,13 @@ class Chef
       # be using the existing getter/setter to manipulate it instead.
       return if !instance_variable_name
 
+      # We deprecate any attempt to create a property that already exists as a
+      # method in some Classes that we know would cause our users problems.
+      # For example, creating a `hash` property could cause issues when adding
+      # a Chef::Resource instance to an data structure that expects to be able
+      # to call the `#hash` method and get back an appropriate Fixnum.
+      emit_property_redefinition_deprecations
+
       # We prefer this form because the property name won't show up in the
       # stack trace if you use `define_method`.
       declared_in.class_eval <<-EOM, __FILE__, __LINE__ + 1
@@ -631,6 +638,30 @@ class Chef
     end
 
     private
+
+    def emit_property_redefinition_deprecations
+      # We only emit deprecations if this property already exists as an instance method.
+      # Weeding out class methods avoids unnecessary deprecations such Chef::Resource
+      # defining a `name` property when there's an already-existing `name` method
+      # for a Module.
+      return unless declared_in.instance_methods.include?(name)
+
+      # Only emit deprecations for some well-known classes. This will still
+      # allow more advanced users to subclass their own custom resources and
+      # override their own properties.
+      return unless [ Object, BasicObject, Kernel, Chef::Resource ].include?(declared_in.instance_method(name).owner)
+
+      # Allow top-level Chef::Resource proprties, such as `name`, to be overridden.
+      # As of this writing, `name` is the only Chef::Resource property created with the
+      # `property` definition, but this will allow for future properties to be extended
+      # as needed.
+      return if Chef::Resource.properties.keys.include?(name)
+
+      # Emit the deprecation.
+      resource_name = declared_in.respond_to?(:resource_name) ? declared_in.resource_name : declared_in
+      Chef.deprecated(:property_name_collision, "Property `#{name}` of resource `#{resource_name}` overwrites an existing method. " \
+        "Please use a different property name. This will raise an exception in Chef 13.")
+    end
 
     def exec_in_resource(resource, proc, *args)
       if resource
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -179,7 +179,7 @@ GEM
       mixlib-log (~> 1.3)
       rack (~> 2.0)
       uuidtools (~> 2.1)
-    cheffish (4.0.0)
+    cheffish (4.1.0)
       chef-zero (~> 5.0)
       net-ssh
     chefspec (5.3.0)
--- a/lib/chef/chef_class.rb
+++ b/lib/chef/chef_class.rb
@@ -30,6 +30,7 @@ require "chef/platform/provider_priority_map"
 require "chef/platform/resource_priority_map"
 require "chef/platform/provider_handler_map"
 require "chef/platform/resource_handler_map"
+require "chef/deprecated"
 require "chef/event_dispatch/dsl"
 require "chef/deprecated"
 
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -150,7 +150,7 @@ class Chef
       end
 
       def content
-        plist_hash = new_resource.hash || gen_hash
+        plist_hash = new_resource.plist_hash || gen_hash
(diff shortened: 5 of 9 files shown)


lib/chef/property.rb
```diff
@@ -518,6 +518,18 @@ def emit_dsl
       # be using the existing getter/setter to manipulate it instead.
       return if !instance_variable_name
 
+      # If the property name is already an existing Ruby method, this could be bad.
+      # For example, the resource may implement #hash which is used internally by
+      # Ruby when adding the object to other data structures, like Hashes.
+      # Let's warn the user that this is bad. Avoid warnings on the core
+      # Chef::Resource class, where properties like `name` may be implemented
+      # and are safe.
+      if declared_in.respond_to?(name) && declared_in != Chef::Resource
```
lamont-granquist: is this to prevent Chef::Resource itself from firing off deprecation warnings?

lib/chef/property.rb
```diff
@@ -518,6 +518,18 @@ def emit_dsl
...
+        resource_name = declared_in.respond_to?(:resource_name) ? declared_in.resource_name : declared_in
+        Chef.deprecated(:property_name_collision, "Property `#{name}` of resource `#{resource_name}` is already an existing Ruby method. " \
+          "This could be dangerous and cause unexpected behavior. This will raise an exception in Chef 13.")
```
lamont-granquist: Might want to just say:

"Property `#{name}` of resource `#{resource_name}` overwrites an existing method.  Please use a different property name.  This will raise an exception in Chef 13."

- it doesn't have to be a ruby method, overwriting methods that are on Chef::Resource itself are bad as well (a problem that we have when we add new methods there)
- you burned a lot of words explaining it could be bad, and i don't think that sentence adds any value
- you didn't actually tell them what to do (rename the property)

adamleff: Thanks for the feedback on that. I initially had "Please choose a different property name" in there but thought that belonged in the docs page in remediation. I'll update it with your suggestion.

lib/chef/property.rb
adamleff: Yes - it will fire off a warning for `name`

lib/chef/deprecated.rb
```diff
@@ -156,6 +156,16 @@ def target
       end
     end
 
+    class PropertyNameCollision < Base
+      def id
+        10
```
adamleff: I can bump mine to 11, didn't know there was something already in flight. Didn't know if we wanted to start a registry of these somewhere that we grab from before we implement.

adamleff:
@jaymzh could you chime in on this? I'd like to deprecate any attempt for a resource property to step on an existing method, but the core `launchd` resource is currently stepping on `Object#hash` with its own `hash` property. Because tests run with deprecations treated like errors, tests are failing.

A couple of options here that I can think of, in my own personal preference order:
 * add a deprecation for the hash property of the launchd resource, introduce the breaking change in Chef 13 and, for now, change this PR to just emit Chef::Log.warn instead of deprecations
 * introduce breaking change to change launchd resource from stepping on hash, because honestly that's A Bad Thing, and then merge this change
 * change the tests to not treat deprecations as errors

I'd love to actually just introduce the breaking change, but that feels like a not-nice thing to do. Since FB contributed the initial launchd core resource, could you please share your thoughts on this one?

lamont-granquist:
most likely you should rename the hash property to something else, then include a straight-up `def hash()` method (not a property) that emits another deprecation warning ("hash property for launchd has been renamed to XXX") and calls the new property name.

coderanger:
:-1: as written. I like the idea but I do actually use redefining of properties on subclasses in a lot of places when I am extending things. If we checked if the name being redefined was a property that would cover me I think though.

lamont-granquist:
yeah that's fair enough, we mostly care about methods like `#hash` and `#class` and non-property methods hanging off of Chef::Resource

lamont-granquist:
(this has also come up before in the past, someone was screaming at me in an issue that their shit broke and they needed this kind of warning...)

coderanger:
We could also check the method owner and only warn on things that are owned by Kernel, Object, or BasicObject (assuming I didn't forget about some place that methods live).

coderanger:
```
>> Foo.instance_method(:hash).owner
=> Kernel
>> Foo.instance_method(:x).owner
=> Foo
```

lamont-granquist:
well writing a property over `Chef::Resource#retries` is also probably pretty bad, along with `#notifies`, `#subscribes`, `#action`, `#identity`, `#ignore_failure`, `#to_json`, `#updated`, etc.

coderanger:
Yeah, probably add `Chef::Resource` to the list too :) But it would still degrade nicely when overloading stuff from resource base classes.

adamleff:
I'll tweak this next week keeping this use case in mind. Thanks for the feedback.

adamleff:
Also, I don't think I realized how lucky I am that I haven't shot myself in the foot before given this shared namespace debacle until I started digging deeper into this area of the codebase. I'd really like to see these depredations added for the not-so-lucky.

lamont-granquist:
If i could rewrite Chef from scratch I'd pay a lot more attention to shared namespaces and DSLs

adamleff:
@lamont-granquist and @coderanger - I think this addresses the concerns and suggestions y'all had. Please take a gander at your convenience.

lamont-granquist:
good 'nuff, hopefully nobody is redefining properties defined on Chef::Resource, and if that happens we can fix that...

adamleff:
@lamont-granquist These tests uncovered Cheffish redefining the name property on each of its resources. I'm going to send up a PR for that unless we want to allow redefining `name` and OK thing to do.

lamont-granquist:
please fix cheffish if you can, that seems like an insane thing to do.

adamleff:
I'm on it :)

adamleff:
PR for cheffish submitted - I think we should likely hold off on merging this until a new cheffish gem is cut and released to avoid further breaking travis. However, if a chef maintainer wants to override me, please feel free to do so. :)

lamont-granquist:
you might want to allow any property in Chef::Resource to be overridden.  the way cheffish is overriding the name to add validations actually isn't as horrible as i expected.

you should be able to query all the Chef::Resource properties (i think literally `Chef::Resource.properties` class method should return an array of them, IIRC?) and exclude them.

adamleff:
@lamont-granquist the only real "property" on Chef::Resource is `name` - everything else is defined with `#set_or_return` which doesn't appear to get appended into the `Chef::Resource.properties` hash.

I'm already about 5 yak levels deep here in adding these deprecations 😄 , so I can redefine the properties in question using `property` (retries, retry_delay, sensitive, guard_interpreter, ignore_failure, provider) if that makes sense to keep this moving. But those don't seem like the kind of things we want people to override.

lamont-granquist:
I wouldn't worry about those.  We might want to make those properties later, but I have a hard time viewing those as being something that people override or that we want to allow overriding.

And I could believe that someone might have accidentally overridden `retries` or possibly `sensitive` with a property, but in that case its the kind of thing that we actually do want to deprecate.

adamleff:
Added an exclusion for Chef::Resource properties set with the `property` method. Currently that only includes `name` but should avoid unnecessary deprecations for users using `name` as a name property in their custom resources.

lamont-granquist:
👍

adamleff:
Holding off on merging this until chef/cheffish#116 is merged and a new cheffish gem is released.

lamont-granquist:
@adamleff is this merge-able?

adamleff:
@lamont-granquist I was waiting until a new cheffish gem was released... otherwise Travis will keep complaining. I defer to y'all on that.

PR #5614, opened 2016-12-05
fix yum regexp similarly to rpm regexp in #3985

lamont-granquist:
this is a "nope, aint got time to test this one" PR.

this affects a code path that is only for localinstalls and doesn't have
any existing tests.

i've already exceeded my 5 minute timebox for fixing the problem.

i'm more likely to rewrite the yum provider from scratch and shitcan the
unit tests in their entirety than to reuse any work that i do here so
any time spent on carefully testing this is likely to end up wasted.


lamont-granquist:
approved by @thommay who is on phone.

PR #5615, opened 2016-12-05
The suggested fix for the manage_home deprecation is incorrect

tas50:
Fix the suggested fix to actually compile

### DIFF
--- a/spec/unit/provider/user/linux_spec.rb
+++ b/spec/unit/provider/user/linux_spec.rb
@@ -69,7 +69,7 @@ describe Chef::Provider::User::Linux do
 
     it "throws a deprecation warning on setting supports[:manage_home]" do
       Chef::Config[:treat_deprecation_warnings_as_errors] = false
-      expect(Chef).to receive(:deprecated).with(:supports_property, "supports { manage_home: true } on the user resource is deprecated and will be removed in Chef 13, set manage_home: true instead")
+      expect(Chef).to receive(:deprecated).with(:supports_property, "supports { manage_home: true } on the user resource is deprecated and will be removed in Chef 13, set manage_home true instead")
       @new_resource.supports( { :manage_home => true } )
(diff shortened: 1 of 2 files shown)


spec/unit/provider/user/linux_spec.rb
```diff
@@ -69,7 +69,7 @@
 
     it "throws a deprecation warning on setting supports[:manage_home]" do
       Chef::Config[:treat_deprecation_warnings_as_errors] = false
-      expect(Chef).to receive(:deprecated).with(:supports_property, "supports { manage_home: true } on the user resource is deprecated and will be removed in Chef 13, set manage_home: true instead")
+      expect(Chef).to receive(:deprecated).with(:supports_property, "supports { manage_home: true } on the user resource is deprecated and will be removed in Chef 13, set manage_home true instead")
```
coderanger: Also if I can be slightly pedantic it would have to be `supports({ manage_home: true })` because lolrubby :(

tas50: There's a few different legacy forms:

https://github.com/chef-cookbooks/djbdns/pull/27/commits/fcbcf4643b66e6417aa4225e063796d3008c173e

thommay:
You've only fixed the spec, not the actual words. But 👍 when you do.

spec/unit/provider/user/linux_spec.rb
coderanger: There are multiple ways to write it but the specific one in the message is not a valid option :)

tas50:
Git branching FTW here. Let me go cherry pick the right commit this time.

lamont-granquist:
👍

tas50:
@thommay @lamont-granquist Updated the actual file. Fixed the same problem in non_unique this time as well and added a spec for that.

PR #5616, opened 2016-12-05
Core: Fix bug where Access Controls on existing symlink resources would be ignored on first chef-client run

tduffield:
### Description

When a link resource needs to change its destination, there was a bug in
the logic flow that would cause the access controls on the resulting
symlink to revert to the default access controls. This commit closes
that hole by re-evaluating the current access controls after creating
the new symlink.

### Issues Resolved

* COOL-614 (Internal Ticket)

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/link.rb
+++ b/lib/chef/provider/link.rb
@@ -121,6 +121,12 @@ class Chef
               file_class.symlink(canonicalize(@new_resource.to), @new_resource.target_file)
               Chef::Log.debug("#{@new_resource} created #{@new_resource.link_type} link from #{@new_resource.target_file} -> #{@new_resource.to}")
               Chef::Log.info("#{@new_resource} created")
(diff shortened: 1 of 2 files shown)


lib/chef/provider/link.rb
```diff
@@ -121,6 +121,12 @@ def action_create
               file_class.symlink(canonicalize(@new_resource.to), @new_resource.target_file)
               Chef::Log.debug("#{@new_resource} created #{@new_resource.link_type} link from #{@new_resource.target_file} -> #{@new_resource.to}")
               Chef::Log.info("#{@new_resource} created")
+              # file_class.symlink will create the link with default access controls.
+              # This means that the access controls of the file could be different
+              # thank those captured during the initial evaluation of current_resource.
```
stevendanna: s/thank/than/

PR #5619, opened 2016-12-06
Bump deps to the latest

tas50:


thommay:
We should do these at the start of each monthly cycle, rather than adhoc. So -1, and let's redo it next week after the release is out.

tas50:
Works for me

PR #5622, opened 2016-12-06
Knife: cookbook site share doesn't work with some json metadata

### LINKED ISSUE #3888: knife cookbook site share doesn't work

lamont-granquist:
rewrite of the specs on top of @andy-dufour's patch

closes #4428 
closes #3888


lamont-granquist:
@chef/client-core i think this is ready, appveyor is green even though the callback to github seems to have failed to register

Issue #4803, opened 2016-04-05
Powershell not_if fails when powershell_out works


youngjl1:
## Description

A certain line of powershell code "import-module webadministration; Get-WebConfiguration –filter system.applicationhost/sites/sitedefaults/logfile" is failing when it is used inside a not_if. When that same line of code is executed in powershell_out it works fine. Incidentally it's a line of code that works in 64bit powershell and fails in 32bit powershell.
## Chef Version

12.8.1
## Platform Version

Microsoft Windows Server 2008 R2 Standard
## Replication Case

In the below case the guard_interpreter runs successfully, returning true and skipping normal execution

``` Ruby
guard_script = <<-EOH
        import-module webadministration
        $log_fields = Get-WebConfiguration –filter system.applicationhost/sites/sitedefaults/logfile | Select-Object -ExpandProperty logExtFileFlags
        ($True)
EOH

powershell_script "set iis logging fields" do
  guard_interpreter :powershell_script
  code <<-EOH
    # do something
  EOH
  #Script fails if I just use not_if. Suspect the reason is 32-bit powershell
  not_if powershell_out(guard_script).stdout
end
```

The below however uses the exact same code but the guard_interpreter doesn't work and the main is not skipped

``` Ruby
powershell_script "set iis logging fields" do
  guard_interpreter :powershell_script
  code <<-EOH
    # do something...
  EOH
  #Script fails if I just use not_if. Suspect the reason is 32-bit powershell
  not_if <<-EOH
        import-module webadministration
        Get-WebConfiguration –filter system.applicationhost/sites/sitedefaults/logfile 
        ($True)
  EOH
end
```

Furthermore I can get the above to work by commenting out the Get-WebConfiguration line e.g.

``` Ruby
powershell_script "set iis logging fields" do
  guard_interpreter :powershell_script
  code <<-EOH
    # do something...
  EOH
  #Script fails if I just use not_if. Suspect the reason is 32-bit powershell
  not_if <<-EOH
        import-module webadministration
        #Get-WebConfiguration –filter system.applicationhost/sites/sitedefaults/logfile 
        ($True)
  EOH
end
```
### NOTE: CHEF CLIENT BUGS ONLY

This is definitely a chef client issue as I am not using any other cookbooks

lamont-granquist:
well i know why they're not the same.

your first example you are:

- calling `powershell_out(guard_script).stdout` and getting back a string
- invoking the powershell guard on the stdout return from that command

the second example you are:

[...]

PR #5624, opened 2016-12-07
[Do Not Merge - On Hold] Calling the shell_out! method to raise errors in powershell_out code flow.

Aliasgar16:
In-progress.

Issue Ref: https://github.com/chef/chef/issues/4803

### DIFF
--- a/lib/chef/mixin/powershell_out.rb
+++ b/lib/chef/mixin/powershell_out.rb
@@ -64,7 +64,7 @@ class Chef
         arch = options.delete(:architecture)
 
         with_os_architecture(nil, architecture: arch) do
-          shell_out(
+          shell_out!(
             build_powershell_command(script),
             options
           )


lamont-granquist:
pretty sure this is exactly what `powershell_out!` accomplishes and you should change the caller to use that.

Aliasgar16:
@lamont-granquist  yeah right...but the difficulty I am facing is that the use of `powershell_out` (with `shell_out!` in its code flow) or `powershell_out!` method (with `not_if` or `only_if` conditional) raises an exception (when any command in the code block passed to it, results in error) and exits the execution with the particular exception raised, where in the other scenario when the same code block (which user passed inside the `powershell_out!` method) when passed without the `powershell_out!` method to the `not_if` conditional, triggers the same error but in the code flow it is evaluated to `false` and then `not_if` continues further without raising any exception and without exiting the recipe.

This is where the user sees different behaviors with the same code block passed to `not_if` with and without `powershell_out` method.

I tried to change the code flow for `powershell_out` to make it consistent with the flow where `powershell_out` is not used with `not_if` conditional, but its code flow is completely different from the later one and so it does not seems to be feasible to achieve the purpose.

btm:
Closing https://github.com/chef/chef/issues/4803, so nothing to do here for now.

PR #5630, opened 2016-12-07
Fix Data Collector organization parsing regex

adamleff:
The regex used to pluck the Chef Server Organization from the
chef_server_url config value did not permit hyphens in org names
even though they are allowed by Chef Server. This caused org names
to get incorrectly truncated when sending the payload to Data
Collector servers.

Re: Chef bug IPO-500

### DIFF
--- a/lib/chef/data_collector/messages/helpers.rb
+++ b/lib/chef/data_collector/messages/helpers.rb
@@ -76,7 +76,7 @@ class Chef
         def chef_server_organization
           return "unknown_organization" unless Chef::Config[:chef_server_url]
 
-          Chef::Config[:chef_server_url].match(%r{/+organizations/+([\w-]+)}).nil? ? "unknown_organization" : $1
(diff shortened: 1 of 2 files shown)


lamont-granquist:
can you just do non-slash characters?  that way when someone creates a unicode organization name using CJK characters will just have it work once the server supports them...

adamleff:
I had that initially and changed it back to match what's currently accepted, and didn't want to grab a query string (if for some reason someone had that in their URL - i.e. they had their own proxy service and used query strings to drive behavior).

tduffield:
Here's the regex applied when orgs are created. Maybe we could just use that? https://github.com/chef/chef-server/blob/master/src/chef-mover/apps/oc_chef_authz/src/oc_chef_organization.erl#L193

stevendanna:
👍 on either doing [^/] or @tduffield's suggestion of using the same regex chef-server uses.  Would there be any sense of putting this in chef-config which already also knows how to parse out the chef_server_root from the chef_server_url?

adamleff:
@chef/client-core I believe I've addressed the initial concern, and Travis is all fixed-up now too.

PR #5631, opened 2016-12-07
Properly deprecate old Chef::Platform methods

lamont-granquist:
These should have been deprecated formally in 12.5.1 after the
Dynamic Provider-Resolver'ing of the internal chef resources
was completed.

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -156,6 +156,16 @@ class Chef
       end
     end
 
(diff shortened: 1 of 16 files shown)


lamont-granquist:
https://github.com/chef/chef-web-docs/pull/310 related PR for the deprecation notice

lamont-granquist:
@chef/client-core should be reviewable (any travis failures that pop up ought to be restartable)

PR #5632, opened 2016-12-07
Update Release Announcements; Keep Dockerfile up-to-date

tduffield:
### Description

* Update the release announcement templates to remove kitchen language.
* Make sure that the default version in the Dockerfile is kept up to date with the latest stable release.
* Update the download URL for chef in the README.

### Check List

- [x] All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,13 @@ If you just want to use Chef, check out these resources:
 
 * [learnchef](https://learn.chef.io): Getting started guide
 * [docs.chef.io](https://docs.chef.io): Comprehensive User Docs
-* [Installer Downloads](https://downloads.chef.io/chef-client/): Install Chef as a complete package
+* [Installer Downloads](https://downloads.chef.io/chef/): Install Chef as a complete package
 
 ## Installing From Git
 
 **NOTE:** Unless you have a specific reason to install from source (to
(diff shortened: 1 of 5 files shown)


btm:
I think we should put the kitchen-dokken directions in the README.md instead along with words about what you would use it for and not the release email templates that are going to go out a couple times a month.

If someone is already using kitchen-dokken, they already have that YAML and would know where to change the version. There isn't the same convenience of a link to the download site where you could just click it to get to the latest version, i.e. someone still has to copy the YAML to the correct place manually, and they're just going to change the version number rather than use the YAML template from a release post.

tduffield:
@btm by that logic, should we remove the instructions on how to specify the version of chef for the normal chef provisioner as well?

btm:
@tduffield yeah, that makes sense to me, it looks the same and doesn't seem like it bears repeating.

btm:
@tduffield were you going to put those YAML blurbs in README.md?

tduffield:
@btm I was but there's language in the README indicating that the README was intended to document development only. I tried to find a good place for it, but I didn't see a good place that wouldn't turn the README into a "These are all the different ways to use Chef" so I put a link to the image that has documentation on how to use it.

PR #5636, opened 2016-12-08
remove Chef::Platform::HandlerMap

lamont-granquist:
seems like an entirely unnecessary level of indirection


lamont-granquist:
@coderanger unless this actually hoses some of your code, I think I'm arguing this is an implicit `@api private` that nobody should have been poking around in.

lamont-granquist:
should probably use composition and delegation rather than inheriting from NodeMap as well, but not today...

coderanger:
I'll read it tonight if I've got postconference brains left, I can also run a few of my test suites against it and see what shakes loose :)

btm:
@tduffield this is a sort of thing that I'd want the changelog generator to pick up but it fits under a different category like 'deprecations' or 'notices'

lamont-granquist:
"code cleanup"  or "tech debt"

PR #5639, opened 2016-12-09
Install rpms with the same name and different architectures - yum.rb

MarkGibbons:
### Description

When using the package resource to install two rpm packages with the same name but different architectures the install of the second package specified fails.

package %w(zlib.x86_64 zlib.i686)  will not install package zlib.i686.

### Issues Resolved

Install all the packages when an array of packages is specified.

### Check List

- [x] All tests pass

When installing multiple packages, if two of the packages (cups.x86 and cups.i686) have the same name
but different architectures the second package specified, cups.i686, will not be installed.
The current method of determining the architecture searchs a list of packages by name, cups, and
will always find the information for cups.x86 instead of the information for cups.i686.
This change moves package architecture processing mostly out of load_current_resource and
into install_remote_package. The package name with architecture specified is returned from
load_current_resource and then used by install_remote_package.

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -51,17 +51,6 @@ class Chef
         # Extra attributes
         #
 
-        def arch_for_name(n)
-          if @new_resource.respond_to?("arch")
-            @new_resource.arch
-          elsif @arch
-            idx = package_name_array.index(n)
-            as_array(@arch)[idx]
-          else
-            nil
-          end
-        end
-
         def arch
           if @new_resource.respond_to?("arch")
             @new_resource.arch
@@ -93,6 +82,7 @@ class Chef
 
         def yum_command(command)
           command = "#{yum_binary} #{command}"
+
           Chef::Log.debug("#{@new_resource}: yum command: \"#{command}\"")
           status = shell_out_with_timeout(command, { :timeout => Chef::Config[:yum_timeout] })
 
@@ -173,6 +163,7 @@ class Chef
               dep = parse_dependency(n, new_version_array[index])
               if dep
                 if @new_resource.package_name.is_a?(Array)
+                  # Remove the dependency from @NR.package_name and add the computed package name to @NR.package_name
                   @new_resource.package_name(package_name_array - [n] + [dep.first])
                   @new_resource.version(new_version_array - [new_version_array[index]] + [dep.last]) if dep.last
                 else
@@ -188,7 +179,6 @@ class Chef
 
           installed_version = []
           @candidate_version = []
-          @arch = []
           if @new_resource.source
             unless ::File.exists?(@new_resource.source)
               raise Chef::Exceptions::Package, "Package #{@new_resource.name} not found: #{@new_resource.source}"
@@ -215,10 +205,8 @@ class Chef
                 # if we parsed an arch from the name, update the name
                 # to be just the package name.
                 if parch
-                  if @new_resource.package_name.is_a?(Array)
-                    @new_resource.package_name[idx] = name
-                  else
-                    @new_resource.package_name(name)
+                  unless @new_resource.package_name.is_a?(Array)
+                    @package_name = name
                     # only set the arch if it's a single package
                     set_arch(parch)
                   end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/yum.rb
```diff
@@ -51,17 +51,6 @@ def yum_binary
         # Extra attributes
         #
 
-        def arch_for_name(n)
-          if @new_resource.respond_to?("arch")
-            @new_resource.arch
-          elsif @arch
```
MarkGibbons: I don't think this block gets entered since @new_resource should always respond to arch.

lib/chef/provider/package/yum.rb
```diff
@@ -51,17 +51,6 @@ def yum_binary
...
-            idx = package_name_array.index(n)
```
MarkGibbons: We search by name here and find the first occurrence.  We want the correct occurrence but have no way to identify which entry with a given name is the right one.

MarkGibbons:
I found a few more cases where installing packages doesn't work.  I'm working on fixing them.

spec/unit/provider/package/yum_spec.rb
```diff
@@ -2279,8 +2279,35 @@ def reset_instance
       expect(@provider).to receive(:yum_command).with(
         "-d0 -e0 -y install cups-1.2.4-11.19.el5.x86_64 vim-1.0"
       )
-      @provider.install_package(%w{cups vim}, ["1.2.4-11.19.el5", "1.0"])
+      @provider.install_package(%w{cups.x86_64 vim}, ["1.2.4-11.19.el5", "1.0"])
     end
 
+    it "should run yum install with the package name, version, and arch when names have arch" do
+      @new_resource = Chef::Resource::Package.new(["cups.x86_64", "cups.i386"])
```
lamont-granquist: If you replace this with:

```ruby
      @new_resource = Chef::Resource::YumPackage.new(["cups", "vim"])
      @new_resource.arch [ "x86_64", nil ]
```

it needs to work identically, but its blowing up on me.  IDK if its something stubbed that isn't being stubbed properly though

lamont-granquist: err wait, i seem to not be on your PR...

spec/unit/provider/package/yum_spec.rb
```diff
@@ -153,14 +153,14 @@
         allow(@yum_cache).to receive(:yum_binary=).with("yum")
         @provider = Chef::Provider::Package::Yum.new(@new_resource, @run_context)
         @provider.load_current_resource
-        expect(@provider.new_resource.package_name).to eq("testing")
+        expect(@provider.new_resource.package_name).to eq("testing.noarch")
```
lamont-granquist: i suspect that this indicates you've actually broken a use case, but i proving it, without taking a week to create a bunch of functional end-to-end tests is going to be difficult.

lamont-granquist:
@jaymzh you probably need to test this and make sure it didn't break anything.

spec/unit/provider/package/yum_spec.rb
MarkGibbons: From my viewpoint specifying the architecture for packages is really a mess.  It might be harder to fix than I have time for.  I'm working on adding tests for some more cases.  What I did above was add the architecture to the package name so that the arch was preserved after going back through package.rb method target_version_array.   If the resource package name needs to be preserved as originally specified I'll have to figure out some other way of associating the package and the architecture.   I think documenting the use cases won't be a waste at any rate.

Thanks,
  Mark

lamont-granquist: i've already got functional tests for dnf_package which are going to get merged over here:  #4894

those tests cover a lot of the bugs in the yum provider -- i tried to launch with dnf provider without all the bugginess that is here.

its a completely different approach, but the func tests could be adapted.

lamont-granquist:
The DNF provider has been merged and has a pile of func tests in it.  I want to add yum func tests by more or less a `s/dnf/yum/g` so that we get better func coverage of this provider before proceeding.

MarkGibbons:
I'll open this up again when/if have time to work on it.

PR #5647, opened 2016-12-13
Swap rbx testing in Travis for Ruby head

tas50:
Rbx was failing as it wasn't a valid rbenv name. Ruby 2.4 comes out this month so we should add it to the test matrix.


lamont-granquist:
what about just ruby-head rather than pinning the exact preview version?

tas50:
@lamont-granquist Has 2.4 been forked off yet?

lamont-granquist:
there isn't even any tags in the repo for the 2.4 preview versions, it looks like they're still working off of head.

tas50:
Swapped out rbx entirely for ruby-head.

lamont-granquist:
LGTM, its been like a year or more and i've never bothered to fix rbx.

PR #5657, opened 2016-12-17
Use the same task name used by the chef-client cookbook so that the task can be managed by its task recipie

mwrock:
This will allow the MSI created task to be managed by the chef-client cookbook. See https://github.com/chef-cookbooks/chef-client/blob/master/recipes/task.rb#L41

cc @btm 

Signed-off-by: Matt Wrock <matt@mattwrock.com>

### DIFF
--- a/omnibus/resources/chef/msi/source.wxs.erb
+++ b/omnibus/resources/chef/msi/source.wxs.erb
@@ -71,14 +71,14 @@
 
     <CustomAction Id="CreateChefClientScheduledTask"
                   Directory="TARGETDIR"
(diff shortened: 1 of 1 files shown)


tduffield:
I don't know much about Windows things, but will this result in an orphan scheduled task on Windows systems?

mwrock:
No. This just changes the name of the task the MSI creates and removes. However, without this PR you could have an orphaned or duplicate task if using the chef-client cookbook because they both use different names. The scheduled task functionality in the MSI is new to the upcoming client release so its important this makes it in before then. Currently this task name is not "in the wild" so their is no opportunity for orphans yet. :)

tas50:
We should absolutely get alignment with the MSI and the cookbook. Otherwise we'll end up with double jobs

PR #5666, opened 2016-12-20
formally deprecate run_command

### LINKED ISSUE #3014: Deprecate Chef::Mixin::Command
We need to mark `Chef::Mixin::Command` with either `Chef::Log.deprecation` or `Chef::Mixin::Deprecation`.  While doing that, we need to remove all internal uses of it (but not remove it from `Provider` because that changes the public API).  Then we can remove the class entirely for 13.0.0

Source: https://github.com/chef/chef/pull/3012

lamont-granquist:
closes #3014

also improves the groupadd provider a bit

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -166,6 +166,16 @@ class Chef
       end
     end
 
+    class RunCommand < Base
+      def id
+        14
+      end
+
+      def target
+        "run_command.html"
+      end
+    end
+
     class ResourceCloning < Base
       def id
         3694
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core fairly simple reviewage

tduffield:
Hey! This PR has broken about half of our tests in Jenkins :) We should look into fixing that!

lamont-granquist:
i suspect the next three PRs fix that.

lamont-granquist:
i need reviewage on #5674 and #5676 and after those are merged that'll clean up the rest of the uses of run_command i could find.

PR #5667, opened 2016-12-21
remove chef-provisioning-aws from travis

lamont-granquist:
blocks testing on ruby 2.4

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,2 +1,2 @@
 ---
-BUNDLE_FROZEN: '1'
+BUNDLE_FROZEN: "1"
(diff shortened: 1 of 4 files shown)


jeremydyoung:
FYI, this change appears to prevent aws-sdk-v1 (or aws-sdk 1.x) from being used with Chef 12.18.x

I realize that aws-sdk-v1 pins json to ~> 1.4, and json 2.x is required for ruby 2.4...  Not really sure if there is a way to resolve this.

lamont-granquist:
the only way to fix it would be to convince amazon to release a ruby 2.4 compatible version of aws-sdk-v1 which allowed json 2.x

since they're working on aws-sdk-v3 now i suspect that will never happen and aws-sdk-v1 is basically abondonware.

note that you can build a custom gemfile which pins aws-sdk to v1 and installs chef 12.18.x and you can bundle exec chef-client out of there, which amounts to working around the appbundler pins in /opt/chef/bin/chef-client.

PR #5668, opened 2016-12-21
Report a dummy resource if sensitive is enabled

afiune:
### Description

Before this commit we were sending sensitive resource information to
Reporting and therefor you were able to see the sensitive data on the
Run History in the Chef Manage Console.

This commit is fixing this problem and now it is transforming any
sensitive resource to display the word `*sensitive*` instead of the
actual data. As we are inserting this values directly to the data base,
it means that this change will cascade all the way to the front-end.

Catch: Old data that was already reported and is displaying sensitive
data will contineu to be displayed. Apologize.

### Issues Resolved

COOL-642/ZD 12936 - Chef Manage Run History compromises sensitive data

### Check List

- [x] All tests pass

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,7 +14,7 @@ _This file holds "in progress" release notes for the current release under devel
   compromise any sensitive data.
 
   _Note: Old data that was already sent to Reporting marked as sensitive will continue to be
-  displayed. Apologize._
+  displayed. Apologies._
 
 ## New deprecations introduced in this release:
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,16 @@ _This file holds "in progress" release notes for the current release under devel
 
 - You can now enable chef-client to run as a scheduled task directly from the client MSI on Windows hosts.
 
+## Highlighted bug fixes for this release:
+
+- Fixed exposure of sensitive data of resources marked as sensitive inside Reporting. Before you
+  were able to see the sensitive data on the Run History tab in the Chef Manage Console. Now we
+  are sending a new blank resource if the resource is marked as sensitive, this way we will not
+  compromise any sensitive data.
+
+  _Note: Old data that was already sent to Reporting marked as sensitive will continue to be
+  displayed. Apologize._
+
 ## New deprecations introduced in this release:
 
 ### Chef Platform Methods
@@ -13,3 +23,4 @@ _This file holds "in progress" release notes for the current release under devel
 - **Deprecation ID**: 13
 - **Remediation Docs**: <https://docs.chef.io/chef_platform_methods.html>
 - **Expected Removal**: Chef 13 (April 2017)
+
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -198,9 +198,30 @@ class Chef
     def resource_completed(new_resource)
       if @pending_update && !nested_resource?(new_resource)
         @pending_update.finish
+
+        # Verify if the resource has sensitive data
+        if @pending_update.new_resource.sensitive
+          sensitive_resource = transform_sensitive_resource(@pending_update.new_resource)
+          @pending_update.new_resource = sensitive_resource
+        end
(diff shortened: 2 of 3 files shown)


afiune:
Here some screenshots of the actual result on the Chef Manage Console:

screen shot 2016-12-20 at 5 16 29 pm

**Clic on the** 👁 

screen shot 2016-12-20 at 5 16 36 pm

thommay:
I don't think we should transform the name property, since that just makes it hard for users to see what's going on. 
I understand that that could result in a leak of sensitive information for execute resources, but I'd prefer to see us raise an error in the client run if the resource is set as sensitive and the `command` parameter is unset, since we're being put in a position where we can't help but leak information. 
@chef/client-core thoughts?

tduffield:
I kind of agree with Thom here. The initial card was focused around the execute command being shown: is there a solution that is specific to that case rather than the more general solution we have here?

lamont-granquist:
it sort of feels like the name property should be left unchanged, while any other property should be omitted if the resource is marked sensitive, and we shouldn't be doing one-offs for file, execute, etc.

coderanger:
This looks like it is mutating the resource objects in place. Am I missing a dup?

coderanger:
Also should this match the behavior and formatting for the existing display code, not just re: name but in general?

afiune:
@thommay @tduffield @lamont-granquist @coderanger I really appreciate the feedback!!! I am very open to suggestions around the implementation since this is the first time I am modifying Reporting (ish) - Would it make sense to create a zoom session to discuss? If not I'll ping you on Slack. 👍

tduffield:
Kay. Which means we'll want to do a resource.dup, modify :allthethingss:, and then return that dup'd resource :)

coderanger:
That's one option, another is to do the redaction when building the hash form of the data rather than where it is in the this patch. That seems less error prone as there would be no mutation of existing objects.

tduffield:
We should make sure to update the README with this.

afiune:
@tduffield You mean the RELEASE_NOTES. - Yeah good idea. On it! 👍

afiune:
I had to rebase... But I think this is ready!! 



cc/ @chef/maintainers

afiune:
@coderanger What do you think about this implementation?

RELEASE_NOTES.md
```diff
@@ -6,10 +6,21 @@ _This file holds "in progress" release notes for the current release under devel
...
 
+## Highlighted bug fixes for this release:
+
+- Fixed exposure of sensitive data of resources marked as sensitive inside Reporting. Before you
+  were able to see the sensitive data on the Run History tab in the Chef Manage Console. Now we
+  are sending a new blank resource if the resource is marked as sensitive, this way we will not
+  compromise any sensitive data.
+
+  _Note: Old data that was already sent to Reporting marked as sensitive will continue to be
+  displayed. Apologize._
```
coderanger: This should be fixed to at least be grammatically correct.

lib/chef/resource_reporter.rb
```diff
@@ -198,6 +198,17 @@ def resource_failed(new_resource, action, exception)
...
         @pending_update.finish
+
+        # Verify if the resource has sensitive data
+        # and create a new blank resource with only
+        # the name so we can report it back without
+        # sensitive data
+        if @pending_update.new_resource.sensitive
+          klass = @pending_update.new_resource.class
+          resource_name = @pending_update.new_resource.name
+          @pending_update.new_resource = klass.new(resource_name)
```
coderanger: This leaves no way to know if a resource was actually blank or sensitive:

```ruby
package 'foo'

# vs

package 'foo' do
  sensitive true
end
```

I don't really need to worry much about y'alls commercial features so I suppose it's up to you if you think that's a problem. I will also point out that we already have a method for fixing this, `Chef::Resource#state_for_resource_reporter`.

coderanger:
Also I don't think this handles sensitive props? I guess that would actually be a problem if that's something you intend to support.

afiune:
@coderanger I just fixed the typo. Thank you for the great review. I will merge this as soon as test pass.

PR #5670, opened 2016-12-21
Style/NumericPredicate comparison style

lamont-granquist:
Apparently rubocop thinks the default behavior should be that we have to
rewrite every use of `foo == 0` into `foo.zero?` which is a big pile of
NOPE for me.  After discovering that `.zero?` is actually slower, I'd
prefer to go the other direction.  Same for `positive?` and
`negative?`.

These are the only uses of `zero?` in the chef/chef codebase, while I'm
pretty sure the inverse rule would touch nearly every file.

### DIFF
--- a/lib/chef/http/json_input.rb
+++ b/lib/chef/http/json_input.rb
@@ -33,7 +33,7 @@ class Chef
 
       def handle_request(method, url, headers = {}, data = false)
         if data && should_encode_as_json?(headers)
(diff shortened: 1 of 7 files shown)


lamont-granquist:
@chef/client-core anyone have strong disagreement here?

thommay:
agreed, `.zero?` seems entirely gratuitous.

lamont-granquist:
oops, i may have done that a bit too fast...

PR #5673, opened 2016-12-21
cleanup of group provider

lamont-granquist:
- consistent use of shell_out_compact!
- remove more uses of run_command
- some other code cleanup


lamont-granquist:
@chef/client-core should be ready for reviewage

PR #5674, opened 2016-12-21
user provider cleanup

lamont-granquist:
needed to be built on top of #5673 so it might be easier to review and merge that one first.

### DIFF
--- a/lib/chef/provider/user.rb
+++ b/lib/chef/provider/user.rb
@@ -39,7 +39,7 @@ class Chef
         if @new_resource.gid.is_a? String
           @new_resource.gid(Etc.getgrnam(@new_resource.gid).gid)
         end
-      rescue ArgumentError => e
+      rescue ArgumentError
         @group_name_resolved = false
       end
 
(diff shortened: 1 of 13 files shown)


lamont-granquist:
@chef/client-core more to review

lib/chef/provider/user.rb
```diff
@@ -175,7 +172,7 @@ def action_lock
       end
 
       def action_unlock
-        if check_lock() == true
+        if check_lock == true
```
thommay: `if check_lock`?

lib/chef/provider/user.rb
```diff
@@ -137,34 +137,31 @@ def action_create
...
+        return unless compare_user
+        converge_by("modify user #{@new_resource.username}") do
+          manage_user
+          Chef::Log.info("#{@new_resource} modified")
         end
       end
 
       def action_lock
-        if check_lock() == false
+        if check_lock == false
```
thommay: `if !check_lock` ?

lamont-granquist:
yeah fixing those nits are out of scope i think

PR #5676, opened 2016-12-22
cleanup of ifconfig and route providers

lamont-granquist:
built on top of #5673 and #5674 so easier to review after those get merged.


lamont-granquist:
@chef/client-core i've been busy today, more reviewage

lamont-granquist:
confirmed that this unbreaks some of the jenkins brakes, so gonna YOLO to make progress on greenification

PR #5680, opened 2016-12-23
Use rubygems_url config option in default case

jflemer-ndp:
### Description

The (undocumented) `Chef::Config` option `rubygems_url` does not work in the default case where `source` is not specified for the `chef_gem` or `gem_package` resource.  Since this config option is undocumented (chef/chef-web-docs#206), this seems like the expected behavior.

### Issues Resolved

 - No open issues, but with out this, setting `rubygems_url` does not work for me

### Check List

- [ ] All tests pass
- [x] This commit includes same level of test and documentation as original commit adding the feature (chef/chef@1d063284f755b996a6563d7210496afd716a9f2d)

### Example

Recipe:
```ruby
chef_gem 'net-ldap' do
  compile_time true
  action :install
end
```

Command:
```
chef-client --local-mode --config-option rubygems_url=http://localhost/ -o test
```

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -540,7 +540,8 @@ class Chef
             src = " --clear-sources"
             src << (new_resource.source && " --source=#{new_resource.source}" || "")
           else
-            src = new_resource.source && " --source=#{new_resource.source} --source=#{Chef::Config[:rubygems_url]}"
+            src = new_resource.source && " --source=#{new_resource.source}" || ""
+            src << (Chef::Config[:rubygems_url] && " --source=#{Chef::Config[:rubygems_url]}" || "")
           end
           if !version.nil? && !version.empty?
             shell_out_with_timeout!("#{gem_binary_path} install #{name} -q --no-rdoc --no-ri -v \"#{version}\"#{src}#{opts}", env: nil)


jflemer-ndp:
cc @lamont-granquist

lib/chef/provider/package/rubygems.rb
```diff
@@ -540,7 +540,8 @@ def install_via_gem_command(name, version)
             src = " --clear-sources"
             src << (@new_resource.source && " --source=#{@new_resource.source}" || "")
           else
-            src = @new_resource.source && " --source=#{@new_resource.source} --source=#{Chef::Config[:rubygems_url]}"
+            src = @new_resource.source && " --source=#{@new_resource.source}" || ""
+            src << (Chef::Config[:rubygems_url] && " --source=#{Chef::Config[:rubygems_url]}" || "")
```
coderanger: I think this is the same thing as a `?:` trinary just written differently?

jflemer-ndp:
Perhaps. I made the code style consistent with the preceding clause for
readability.

lamont-granquist:
this feels like its getting convoluted.  can we just make the Chef::Config variable the default for the source property in the resource (and eliminate `--clear-sources` entirely?)

jflemer-ndp:
That solution works for my need (point chef to a internal mirror of
rubygems.org)

lamont-granquist:
Lets revisit this in early feb after we branch and fix this in Chef-13 the clean way.

jflemer-ndp:
Any chance of just merging this trivial fix for 12.x now? It is no less clean than the existing 12.x capability. It is a blocking issue that prevents using a mirror.

jflemer-ndp:
Note: Updated to latest master

jflemer-ndp:
Do you think this will be addressed in Chef 13?

lamont-granquist:
replacing with #5995

PR #5698, opened 2017-01-04
Avoid apt (dpkg) prompts for modified config files

ccope:
### Description

chef runs should not fail due to dpkg prompts in the following scenario:
 1. User installs package
 1. User overwrites config files from package with Chef-managed files
 1. User upgrades the package, which contains an new configuration file from the package maintainer

chef should default to the following behavior:
 - if the admin never changed the config file, install the updated maintainer config
 - if the admin has changed the file, preserve the existing config

> dpkg handles configuration file updates, but, while doing so, regularly interrupts its work to ask for input from the administrator. This makes it less than enjoyable for those who wish to run updates in a non-interactive manner. This is why this program offers options that allow the system to respond automatically according to the same logic:
 **--force-confold** retains the old version of the file
 **--force-confnew** will use the new version of the file (these choices are respected, even if the file has not been changed by the administrator, which only rarely has the desired effect). 
 **--force-confdef** tells dpkg to decide by itself when possible (in other words, when the original configuration file has not been touched), and only uses --force-confnew or --force-confold for other cases.

-- https://debian-handbook.info/browse/stable/sect.package-meta-information.html#sidebar.questions-conffiles

### Issues Resolved

https://gitlab.com/gitlab-com/operations/issues/254
http://serverfault.com/questions/575343/out-of-order-chef-recipes-causing-apt-package-install-to-fail
http://askubuntu.com/questions/104899/make-apt-get-or-aptitude-run-with-y-but-not-prompt-for-replacement-of-configu/104912#104912

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -145,10 +145,12 @@ class Chef
           # It doesn't make sense to install packages in a scenario that can
           # result in a prompt. Have users decide up-front whether they want to
           # forcibly overwrite the config file, otherwise preserve it.
-          if new_resource.respond_to?(:overwrite_config_files) && new_resource.overwrite_config_files
-            '-o Dpkg::Options::="--force-confnew"'
-          else
-            '-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"'
+          if new_resource.respond_to?(:overwrite_config_files)
+            if new_resource.overwrite_config_files
+              [ '-o', 'Dpkg::Options::="--force-confnew"' ]
+            else
+              [ '-o', 'Dpkg::Options::="--force-confdef"', '-o', 'Dpkg::Options::="--force-confold"' ]
+            end
           end
         end
 
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -86,7 +86,7 @@ class Chef
           package_name = name.zip(version).map do |n, v|
             package_data[n][:virtual] ? n : "#{n}=#{v}"
           end
-          run_noninteractive("apt-get", "-q", "-y", default_release_options, options, "install", package_name)
+          run_noninteractive("apt-get", "-q", "-y", config_file_options, default_release_options, options, "install", package_name)
         end
 
         def upgrade_package(name, version)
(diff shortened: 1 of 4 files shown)


ccope:
Waiting for feedback before updating tests/docs

coderanger:
Does apt not already suppress those prompts when running in non-interactive mode?

ccope:
I was very surprised, but apparently not. I hit errors when upgrading an apt package via chef 12.17.44 on Ubuntu 14.04 today ([traceback](https://gist.github.com/ccope/03f9a55ecd40f84dec83b6cab7d09832
)). The issue from the gitlab link in the PR description also looks the same.

I could have sworn that `DEBIAN_FRONTEND=noninteractive` fixed this for me in the past, but one of my Debian developer coworkers pointed me to a [manpage](http://manpages.ubuntu.com/manpages/trusty/man7/debconf.7.html) that shows that variable applies to debconf (which handles input for configuration settings), not dpkg (which manages conffiles in /etc). I also grep'd the source for the Ubuntu apt and dpkg packages and they don't seem to mention that environment variable anywhere.

I'm wondering now if these arguments have just been set in /etc/apt/conf.d everywhere I've worked before...

coderanger:
Sounds believable, but we should probably figure out why this has been working up until now for the majority of peoples to make sure we aren't subtly changing the current behavior.

ccope:
Yeah I would really like to know why this hasn't come up before. There are a lot of posts on the internet about having to pass these options in shell scripts.

I'm managing a separate chef environment from our production setup, but looking at the production repository we've been setting these options in /etc/apt/conf.d.

FWIW, puppet has similar logic going all the way back to 2006: https://github.com/puppetlabs/puppet/blob/94484df1a8d00e0eae35cbe09c0e5d2ff4b16f49/lib/puppet/provider/package/apt.rb#L69
and ansible also sets these options since early 2013: https://github.com/ansible/ansible/blob/f38ca6c92f0ee06d52c3db6135d23b4c1de308d8/lib/ansible/modules/packaging/os/apt.py#L252

thommay:
I'm astounded we've never hit this, but I guess the majority of packages now use debconf rather than relying on dpkg's configfiles behaviour directly.
I think the change is reasonable; it would be good to know what package you hit this with (assuming it's a public one) so that we can verify behaviour going back in time a bit.
We should probably do the same thing for the dpkg provider.

ccope:
I encountered this with the lxc packages from the stable ppa (probably the same one as in the trusty-backports repo). Reposting the link to the traceback with the error here: https://gist.github.com/ccope/03f9a55ecd40f84dec83b6cab7d09832

ccope:
And the package, in case anyone is interested: https://www.dropbox.com/s/66z25rjzk9qv4q4/lxc-common_2.0.6-0ubuntu1~ubuntu14.04.1~ppa1_amd64.deb?dl=0

thommay:
@ccope so I think the logic and code is correct. Please go ahead and write some tests and we'll get this merged. Thanks!

ccope:
Apologies for the delay, I got swamped by other work stuff.

I've fixed most of the tests, but ran into a couple of issues.

1. Some tests already contain these arguments in the `options` parameter, which made me think I should add validation to make sure it doesn't contain any of these arguments. What is the best way to do that?
 Copy the `options` property from the Package resource into the AptPackage resource, and add a callback?

2. The resource does not appear to have the overwrite_config_files property set when the apt_spec tests instantiate Chef::Resource::Package instances, but aren't those supposed to create AptPackage resources?

PR #5701, opened 2017-01-05
Added module_version attribute for dsc_resource.

Aliasgar16:
Added `module_version` attribute support for `dsc_resource` resource to allow users to specify which version of a module (out of the multiple versions of a module installed) to use.

### DIFF
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -525,5 +525,8 @@ This error is most often caused by network issues (proxies, etc) outside of chef
         super "Found multiple matching resources. #{matches_info.join("\n")}"
       end
     end
+
+    # exception specific to invalid usage of 'dsc_resource' resource
+    class DSCModuleNameMissing < RuntimeError; end
   end
 end
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -527,6 +527,6 @@ This error is most often caused by network issues (proxies, etc) outside of chef
     end
 
     # exception specific to invalid usage of 'dsc_resource' resource
-    class DSCModuleNameMissing < RuntimeError; end
+    class DSCModuleNameMissing < ArgumentError; end
   end
 end
--- a/lib/chef/provider/dsc_resource.rb
+++ b/lib/chef/provider/dsc_resource.rb
@@ -29,6 +29,7 @@ class Chef
         super
         @new_resource = new_resource
         @module_name = new_resource.module_name
+        @module_version = new_resource.module_version
         @reboot_resource = nil
       end
 
@@ -48,6 +49,10 @@ class Chef
         true
       end
 
+      def module_name_missing?
+        !(!@module_name && @module_version) ? true : false
+      end
+
       def define_resource_requirements
         requirements.assert(:run) do |a|
           a.assertion { supports_dsc_invoke_resource? }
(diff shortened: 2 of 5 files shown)


Aliasgar16:
Done.

lib/chef/provider/dsc_resource.rb
```diff
@@ -92,6 +101,10 @@ def supports_refresh_mode_enabled?
         Chef::Platform.supports_refresh_mode_enabled?(node)
       end
 
+      def module_usage_valid?
+        !(!@module_name && @module_version) ? true : false
```
smurawski: Since the left hand side is already going to eval to `TrueClass` or `FalseClass`, I don't think we need the ternary expression at the end.

lib/chef/exceptions.rb
```diff
@@ -525,5 +525,8 @@ def initialize(resources_found)
         super "Found multiple matching resources. #{matches_info.join("\n")}"
       end
     end
+
+    # exception specific to invalid usage of 'dsc_resource' resource
+    class DSCModuleNameMissing < RuntimeError; end
```
btm: It looks like this is user specified information, so ArgumentError?

lib/chef/provider/dsc_resource.rb
```diff
@@ -65,6 +66,14 @@ def define_resource_requirements
           a.whyrun err + ["Assuming a previous resource sets the RefreshMode."]
           a.block_action!
         end
+        requirements.assert(:run) do |a|
+          a.assertion { module_usage_valid? }
+          err = ["module_name must be supplied along with module_version."]
+          a.failure_message Chef::Exceptions::DSCModuleNameMissing,
+            err
+          a.whyrun err + ["Assuming a module_name is specified."]
```
btm: Either the module_name is supplied or it isn't, right? There isn't anything dynamic in another resource that would have specified the module_name for the user. In that case, this line should be removed.

lib/chef/resource/dsc_resource.rb
```diff
@@ -68,6 +68,14 @@ def module_name(value = nil)
         end
       end
 
+      def module_version(value = nil)
```
btm: Take a look at using `set_or_return` as it's used in other resources, e.g.

```
def module_version(arg = nil)
  set_or_return(
    :module_version,
    arg,
    :kind_of => [ String ]
  )
end
```

lib/chef/provider/dsc_resource.rb
btm: How about `!(@module_name.nil? || @module_version.nil?)` ?

Aliasgar16: @smurawski I will modify this.

@btm The scenarios I have handled here is like below:

| Module Name | Module Version | Result |
| ------------------ | --------------------- | -------- |
| Given              | Given                 | Valid   |
| Not given        | Not given           | Valid    |
| Given              | Not given           | Valid   |
| Not given        | Given                  | Invalid |

1. If **module_name** and **module_version** both are given then it invokes the given module with the given version.
2. If both are not given then default module **PSDesiredStateConfiguration** is used.
3. If **module_name** is given and **module_version** is not given, then it invokes the given module without the version. In this case, if more than one version of the given module is installed then Chef raises exception saying resource not found.
4. If **module_name** is not given and **module_version** is given, then this is an invalid case.

Let me know if you want to modify these scenarios. You can also take a look at the test scenarios document (attached on the card-434) to understand all the positive and negative scenarios with their results.

lib/chef/exceptions.rb
Aliasgar16: @btm I will modify this.

lib/chef/provider/dsc_resource.rb
Aliasgar16: @btm I will modify this.

PR #5704, opened 2017-01-05
Core: Ensure paths are correctly escaped when syntax checking

### LINKED ISSUE #5702: knfie upload - too short control escape
## Description

In the Chef::Cookbook::SyntaxCheck class there is a method called remove_uninteresting_ruby_files() that filters out cookbooks' files we don't want to treat as ruby files, and further, don't want to validate it's syntax.

```
def remove_uninteresting_ruby_files(file_list)
  file_list.reject { |f| f =~ %r{#{cookbook_path}/(files|templates)/} }
end
```

Sometimes when we want to upload cookbooks to a chef server using **knfie upload**, this method generates error:

`too short control escape (ArgumentError)`

This depends on directory, where the cookbooks are stored.

If directory path contains special characters for regexp (control escape characters), the error appears.

For example, in path:
**C:\\AGENT-HOME\\xml-data\\build-dir\\76808194-76906499\\artifact\\cookbooks/java'**

**\\xml-data** is handled as
\xnn - Matches an ASCII character, where nn is a two-digit hexadecimal character code.

**\\cookbooks** is handled as
\cX - Matches an ASCII control character, where X is the letter of the control character. For example, \cC is CTRL-C.

To eliminate this bug, I suggest to "normalize" cookbook_path using `Chef::Util::PathHelper.escape_glob_dir()` method.

escape_glob_dir() is used many times in SyntaxCheck class.

## Chef Version

Chef Development Kit Version: 0.9.0
chef-client version: 12.5.1
berks version: 4.0.1
kitchen version: 1.4.2

## Platform Version

Windows 2012r2

ceneo:
### Description

Fixed 'too short control escape' on knfie upload.

### Issues Resolved

#5702

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/cookbook/syntax_check.rb
+++ b/lib/chef/cookbook/syntax_check.rb
@@ -110,7 +110,7 @@ class Chef
       end
 
       def remove_uninteresting_ruby_files(file_list)
-        file_list.reject { |f| f =~ %r{#{Chef::Util::PathHelper.escape_glob_dir(cookbook_path)}/(files|templates)/} }
+        file_list.reject { |f| f =~ %r{#{Regexp.quote(cookbook_path)}/(files|templates)/} }
       end
 
       def ruby_files
--- a/lib/chef/cookbook/syntax_check.rb
+++ b/lib/chef/cookbook/syntax_check.rb
@@ -110,7 +110,7 @@ class Chef
       end
 
       def remove_uninteresting_ruby_files(file_list)
-        file_list.reject { |f| f =~ %r{#{cookbook_path}/(files|templates)/} }
+        file_list.reject { |f| f =~ %r{#{Chef::Util::PathHelper.escape_glob_dir(cookbook_path)}/(files|templates)/} }
       end
 
       def ruby_files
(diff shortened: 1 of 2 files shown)


thommay:
This looks good, but please add a test with one of the paths you mentioned in #5702

lib/chef/cookbook/syntax_check.rb
```diff
@@ -110,7 +110,7 @@ def remove_ignored_files(file_list)
       end
 
       def remove_uninteresting_ruby_files(file_list)
-        file_list.reject { |f| f =~ %r{#{cookbook_path}/(files|templates)/} }
+        file_list.reject { |f| f =~ %r{#{Chef::Util::PathHelper.escape_glob_dir(cookbook_path)}/(files|templates)/} }
```
lamont-granquist: this should be `Regexp.quote` instead of `...escape_glob_dir`

cheeseplus:
Had a user in community slack test/confirm this fixes the problem

cheeseplus:
Had a user in community slack test/confirm this fixes the problem

lib/chef/cookbook/syntax_check.rb
ceneo: Regex escape was my first thought. It solves the `too short control escape` error, but it also breaks functionality of `remove_uninteresting_ruby_files` method.

When debugged I found out that even Windows `file_list` have a path consisting to forward slashes and, in the other hand, `cookbook_path` consists to backslashes.

So, it's necessary to convert backlashes to forward slashes. Based on code documentation, `escape_glob_dir` is a recomended way for that.
And of course, when we convert backslashed to forward slashes `too short control escape` error disappears.

PR #5708, opened 2017-01-05
yum_repository: Allow baseurl to be an array & allow fastestmirror_enabled false

tas50:
Port two fixes that were made the yum cookbook yum_repository provider.

https://github.com/chef-cookbooks/yum/pull/165
https://github.com/chef-cookbooks/yum/pull/162


lamont-granquist:
can we kill the cookbook version, because maintaining two versions is how we wound up with the windows cookbook nightmare.

tas50:
My hope was to use compat_resource in the yum cookbook to kill it off. We need that whole patching context thing changed from cookbook to global for that to work though I believe.

PR #5710, opened 2017-01-06
WIP: Handle non-root usage of cron resource and basic test.

Tensibai:
### Description

Ensure the crontab command line used doesn't use `-u` if not run as root, fail if non-root and trying to set someone else crontab.

### Issues Resolved

- Issue 2491
- StackOverflow Q/A: [Unable to run Chef cron recipe as non-root user](http://stackoverflow.com/questions/40470685/unable-to-run-chef-cron-recipe-as-non-root-user/40475768#40475768)

### Check List

- [X] All tests pass

### TODO: 

It should have more tests for the cases:

- [x] Non-root user editing its own crontab should not raise an error.
- [x] Non-root user editing someone else crontab should raise an error.

### DIFF
--- a/lib/chef/provider/cron.rb
+++ b/lib/chef/provider/cron.rb
@@ -36,25 +36,12 @@ class Chef
       ENV_PATTERN = /\A(\S+)=(\S*)/
 
       def can_modify_cron?
-        unless Process.euid == 0
-          begin
-            nuid = Etc.getpwnam(@new_resource.user).uid
-          rescue ArgumentError
-            raise Chef::Exceptions::Cron,
-              "Error in #{@new_resource.name}, #{@new_resource.user} does not exists on this system."
-          end
-          unless nuid == Process.euid
-            raise Chef::Exceptions::Cron,
-              "Error in #{@new_resource.name}, Chef can't modify another user crontab when not running as root."
-          end
-        end
       end
 
       def initialize(new_resource, run_context)
         super(new_resource, run_context)
         @cron_exists = false
         @cron_empty = false
-        @specify_user = ""
       end
       attr_accessor :cron_exists, :cron_empty
 
@@ -62,13 +49,34 @@ class Chef
         true
       end
 
+      def check_resource_semantics!
+        etype = Chef::Exceptions::Cron
+        unless Process.euid == 0
+          begin
+            nuid = Etc.getpwnam(@new_resource.user).uid
+          rescue ArgumentError
+            emessage = "Error in #{@new_resource.name}, #{@new_resource.user} does not exists on this system."
+            events.provider_requirement_failed(@new_resource.action, @new_resource.name, etype, emessage)
+            if why_run?
+              events.whyrun_assumption(@new_resource.action,
+                                @new_resource.name, "Assuming user #{@new_resource.user} would have been created")
+            else
+              raise etype, emessage
+            end
+          end
+          unless nuid == Process.euid
+            emessage = "Error in #{@new_resource.name}, Chef can't modify another user crontab when not running as root."
+            events.provider_requirement_failed(@new_resource.action, @new_resource.name, etype, emessage)
+            raise etype, emessage
+          end
+        end
+      end
+
       def load_current_resource
         crontab_lines = []
         @current_resource = Chef::Resource::Cron.new(@new_resource.name)
         @current_resource.user(@new_resource.user)
         @cron_exists = false
-        # Raise an execption if Chef can't read or update the crontab.
-        can_modify_cron?
 
         if crontab = read_crontab
           cron_found = false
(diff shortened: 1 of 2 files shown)


Tensibai:
@chef/client-maintainers Hello, If you can give me some guidance to polish this it would be awesome.

lib/chef/provider/cron.rb
```diff
@@ -51,6 +67,9 @@ def load_current_resource
         @current_resource = Chef::Resource::Cron.new(@new_resource.name)
         @current_resource.user(@new_resource.user)
         @cron_exists = false
+        # Raise an execption if Chef can't read or update the crontab.
+        can_modify_cron?
```
lamont-granquist: this should be moved into define_resource_requirements and should support why-run mode.   so when the user doesn't exist it should result in a "assuming user would have been created" style of warning in why-run.  the other exception should most likely also fail in why-run mode, but should still be raised in d_r_r.

lib/chef/provider/cron.rb
```diff
@@ -207,7 +226,8 @@ def set_environment_var(attr_name, attr_value)
 
       def read_crontab
         crontab = nil
-        status = popen4("crontab -l -u #{@new_resource.user}") do |pid, stdin, stdout, stderr|
+        @specify_user = "-u #{@new_resource.user}" if Process.euid === 0
```
lamont-granquist: this should just be a local variable and not an ivar

lib/chef/provider/cron.rb
Tensibai: Ok, I'll have a look at this tomorrow,  I may ask for 'backup' if I don't get by myself how to switch between normal and why-run mode.

Still wondering if fixing the closed PR referenced in the issue would be better or not (overlooked the previous PR before submitting this one)

lamont-granquist: if you look at the assertions API that is used in all the define_resource_requirements methods across our providers that's all that i'm talking about.  there should be lots of examples, its just an assertion that must be true, with an exception to raise in normal mode if its not, and text for why run if its not.

Tensibai: @lamont-granquist Currently playing with the assertions, from the provider.rb the resource requirement are processed after the load_current_resource, in case of cron, the read_crontab (which this PR aim to improve) is called within load_current_resource, and the validation has to be done before. I don't see how to achieve this properly :/ I think I see how to add why-run support to the existing code, but I don't see how moving to d_r_r could help. Maybe using the code above in `check_resource_semantics!` and raising or messaging if in why-run ?

lamont-granquist: okay, sorry, it sounds we should walk that back then.

in that case i think it may be fine to go the other direction and not worry about decorating the exception.  if you do `shell_out!("crontab -u foobar")` and `foobar` doesn't exist you'll get a perfectly acceptable ShellOut error with output from crontab indicating that the user doesn't exist.  you're probably doing a lot of work worrying about validating with `getpwnam` that you don't need to bother with.

lamont-granquist:
duplicating my last message since i posted it on an out-of-date review thread:

okay, sorry, it sounds we should walk that back then.

in that case i think it may be fine to go the other direction and not worry about decorating the exception. if you do shell_out!("crontab -u foobar") and foobar doesn't exist you'll get a perfectly acceptable ShellOut error with output from crontab indicating that the user doesn't exist. you're probably doing a lot of work worrying about validating with getpwnam that you don't need to bother with.

lamont-granquist:
and you probably want to convert `popen4` to `shell_out!` which will start to make everything a bit easier, you're digging around in a very old codebase and it looks like nobody got around to doing that.  i don't see any reason `popen4` is still there other than this code may have been written before we even had `run_command`.

Tensibai:
Indeed, the previous PR was heading to use shell_out... I'm unsure how to merge both, decorating the why sounds ok for me, using shell_out doesn't sound that hard using code from previous PR,  but I'm afraid stealing someone else code. Any position from maintainers about this ?

Tensibai:
And for the records,  the problem arise when reading the crontab, writing always give a meaningful error, but we never go to this point because the read crontab would fail before with a message not really obvious. The issue clearly explain it. Hence why we have to validate the parameters before even trying to know the resource status. I'm quite confident with the actual code for validation within check_resource_semantic! in fact (minus typo, English grammar and code grammar). I can review it once more to use shell_out but I'm unsure about the side effects. 
If there's no obvious side effect I'll move using it, but to be honest I'm quite afraid with the tests changes it implies.

thommay:
Sorry this got somewhat left; we just discussed it and we're going to have to do a more or less complete rewrite of cron, so we're gonna close this for now. Thanks for your patience and for the PR!

Tensibai:
@thommay np, I was looking at another PR reviewing this point also (and at the same state) to make it 'mergeable' but I've been busy with work. That's the second PR closed for this subject, may I suggest putting this on chef's core backlog before another come hitting the wall?

Tensibai:
@lamont-granquist considering the actual state of the cron provider, the overall problem has been fixed for long. I dunno how I missed it when we've been discussing it. 

The only point is the error may not be very clear for the end user, but giving an informative message for this edge case is probably not worth the extra code to handle it.

So, my apologies, it can be closed.

Tensibai:
cloud be closed in favor of #8340 where I rebased the changes against current master

PR #5713, opened 2017-01-09
kitchen-tests: remove deprecated openssh options

lamont-granquist:


lamont-granquist:
doesn't fix the current master breakage, but removes the deprecation warnings that'll bit us eventually when they go away...

PR #5718, opened 2017-01-11
enforce 2.4.0 failures

lamont-granquist:
since its green, we should probably keep it green.


thommay:
Although given the apparent speed of 2.4 we might not want to do this right now

thommay:
( https://bugs.ruby-lang.org/issues/13113 - fix is on HEAD and should be in 2.4.1 )

PR #5731, opened 2017-01-18
Beginning of automatic API version detection

thommay:
When we make a request to a chef server, we capture the minimum and
maximum support API versions and allow them to be queried.
We then provide some infrastructure for making decisions on which class
should be used, in a middleware-ish mechanism.

### DIFF
--- a/lib/chef/mixin/versioned_api.rb
+++ b/lib/chef/mixin/versioned_api.rb
@@ -1,0 +1,49 @@
+#--
+# Copyright:: Copyright 2017, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Mixin
+    module VersionedAPI
+
+      def self.included(base)
+        # When this file is mixed in, make sure we also add the class methods
+        base.send :extend, ClassMethods
+      end
+
+      module ClassMethods
+        def versioned_interfaces
+          @versioned_interfaces ||= []
+        end
+
+        def add_api_version(klass)
+          versioned_interfaces << klass
+        end
+      end
+
+      def select_api_version
+        self.class.versioned_interfaces.select do |klass|
+          version = klass.send(:supported_api_version)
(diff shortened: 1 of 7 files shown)


thommay:
A sample for how this might be used:

```ruby
require 'chef/mixin/versioned_api'
require 'chef/server_api_versions'
require 'pp'

class BaseVersion
  include Chef::Mixin::VersionedAPI

  class Version1
    def self.supported_api_version
      0
    end
  end

  class Version2
    def self.supported_api_version
      2
    end
  end

  add_api_version Version1
  add_api_version Version2

end

# this would ordinarily be set by the HTTP middleware.
Chef::ServerAPIVersions.instance.set_versions({"min_version" => 0, "max_version" => 1})

foo = BaseVersion.new
pp foo.select_api_version
```

thommay:
@lamont-granquist @chef/maintainers

stevendanna:
Hey! Tests!

stevendanna:
This looks like a good addition to me.  Might be interesting to try to convert the User class to using it to see if the API is what we need to make that cleaner.

lib/chef/mixin/versioned_api.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def select_api_version
+        self.class.versioned_interfaces.select do |klass|
+          version = klass.send(:supported_api_version)
+          # min and max versions will be nil if we've not made a request to the server yet,
+          # in which case we'll just start with the highest version and see what happens
+          ServerAPIVersions.instance.min_server_version.nil? || (version >= ServerAPIVersions.instance.min_server_version && version <= ServerAPIVersions.instance.max_server_version)
+        end
+          .sort { |a, b| a.send(:supported_api_version) <=> b.send(:supported_api_version) }
```
thommay: `minimum_api_version` rather than `supported_api_version`

lamont-granquist:
names and shape should probably be:

```ruby
class ThingV2
  minimum_api_version 2
end

class ThingV1
  minimum_api_version 1
end

class Thing
  add_versioned_api_class ThingV2
  add_versioned_api_class ThingV1
end
```

lamont-granquist:
to answer my own question i just asked on zoom -- if the client wants to use a user object against a server it has never talked to before then it would get back a highest-numbered version of the object (ThingV2) and then would need to rescue an exception and retry, and that would be on the caller to do that (presumably we could more tightly bake this into our API objects and Chef::HTTP at some point).

thommay:
new version looks a lot better:
```ruby
class Version1
  include Chef::Mixin::VersionedAPI
  minimum_api_version  0
end

class Version2
  include Chef::Mixin::VersionedAPI
  minimum_api_version  2
end

class BaseVersion
  include Chef::Mixin::VersionedAPIFactory

  add_versioned_api_class Version1
  add_versioned_api_class Version2

end

Chef::ServerAPIVersions.instance.set_versions({ "min_version" => 0, "max_version" => 1 })

foo = BaseVersion.new
pp foo.class
=> Version1
```

lamont-granquist:
👍

PR #5736, opened 2017-01-19
Add gems for ECC algorithm support to omnibus.

rhass:
### Description

These definitions have been added to resolve build issues with ECC algorithm support on Solaris and Cisco IOS-XR.

The bcyrpt_pbkdf gem currently link against shared libs which are not installed on Cisco IOS-XR.

The rbnacl gems currently fail to compile on Solaris for both x86-64 and
SPARC.

Adding these definitions will allow us to selectively define the supported platforms as well as provide a simple means to iterate on patches as needed.

Note: This was pinned to a dev branch for testing:
https://github.com/chef/chef/compare/rhass/COOL-657_ecc-algorithm-support?expand=1#diff-ba79ba7a139ca66864e6fb2f9cbbc904R11

I am targeting master, so this pinning will need to be reverted and the patch cherry-picked until we kill those branches in a couple weeks.

- Related PR: https://github.com/chef/omnibus-software/pull/786

### Issues Resolved

- COOL-657 (Internal Tracking ID)

### Check List

- [ ] All tests pass


chef.gemspec
```diff
@@ -26,9 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "ffi-yajl", "~> 2.2"
   s.add_dependency "net-ssh", ">= 2.9", "< 5.0"
-  s.add_dependency "rbnacl-libsodium", "~> 1.0.0"
   s.add_dependency "rbnacl", "~> 4.0.0"
```
lamont-granquist: you probably want to remove this as well.

omnibus/config/software/chef.rb
```diff
@@ -44,6 +44,9 @@
 dependency "chef-gem-byebug"
 dependency "chef-gem-debug_inspector"
 dependency "chef-gem-binding_of_caller"
+dependency "chef-gem-rbnacl-libsodium" unless ios_xr?
+dependency "chef-gem-bcrypt_pbkdf-ruby" unless solaris?
```
lamont-granquist: and likely you just want to make this an `unless ios_xr? || solaris?` around both of them (or all three)

lamont-granquist:
couple comments, although i don't feel strongly about them...

tduffield:
Test build (chef 748) was successful

omnibus/config/software/chef.rb
```diff
@@ -44,6 +44,10 @@
 dependency "chef-gem-byebug"
 dependency "chef-gem-debug_inspector"
 dependency "chef-gem-binding_of_caller"
+unless ios_xr? || solaris?
```
afiune: Do we need both deps for any Cisco or Solaris OS? 🤔  I ask this because in the desc @rhass mentioned that the bcyrpt_pbkdf gem currently link against shared libs which are not installed on Cisco IOS-XR and that the rbnacl gems currently fail to compile on Solaris for both x86-64 and SPARC.

tduffield: This is how lamont and thom recommended we do it, and the builds succeeded, so unless there's a specific reason we should I'm inclined to keep it.

lamont-granquist: We don't need this feature on either Cisco or Solaris, its mostly for workstations, and Cisco's are definitely not workstations, and Solaris workstations should be reasonably miniscule as a use case.  The bigger problem here is Linux, Windows, Mac, and maybe FreeBSD.  It introduces a slight discrepancy in our omnibus philosophy, but we've had that before with ruby-shadow.

afiune: I trust y'all 👍  - Just wanted to point it out. 😄

PR #5742, opened 2017-01-20
Do not modify File's new_resource during why-run

scottopherson:
### Description

The `File` provider's `action_create` method would modify the `new_resource` object during a
`why-run`. This was problematic when a `File` resource was configured with a `notifies :before`
because resources with `before` timers have their actions executed twice. [Once](https://github.com/chef/chef/blob/a2e1d1d4edb74a968b416517fdbdf35740686467/lib/chef/runner.rb#L54-L57) in `why-run` mode to
see if the other resources need to be notified, and then executed [once again](https://github.com/chef/chef/blob/a2e1d1d4edb74a968b416517fdbdf35740686467/lib/chef/runner.rb#L68-L69) after the `before`
notifications are resolved. This behavior would result in existing File resources not getting their
ownership and permissions updated properly during a converge.

### Issues Resolved

Internal ticket

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/directory.rb
+++ b/lib/chef/provider/directory.rb
@@ -138,7 +138,7 @@ class Chef
         end
         do_acl_changes
         do_selinux(true)
-        load_resource_attributes_from_file(@new_resource)
+        load_resource_attributes_from_file(@new_resource) unless Chef::Config[:why_run]
       end
 
       def action_delete
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -154,7 +154,7 @@ class Chef
         do_contents_changes
         do_acl_changes
         do_selinux
-        load_resource_attributes_from_file(@new_resource)
+        load_resource_attributes_from_file(@new_resource) unless Chef::Config[:why_run]
       end
 
       def action_create_if_missing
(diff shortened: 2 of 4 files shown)


lamont-granquist:
For posterity:

Note that we do this in non-why-run in order to get reporting correct.

We have a longstanding problem where we really have three different kinds of state:

- the desired state
- the current state
- the ending state

We have two different objects and we (ab)use the new_resource to have two roles:

- new_resource:  the desired state and the ending state
- current_resource:  the current state

What reporting needs is to diff the current state and the ending state.   Diffing the current state and the desired state gets it wrong.  So what we do in some cases is mutate the new resource to load the ending state in order for the reporting diff to be correct.   This is bad, but this is what happens.

Note that the file checksum property caused large amounts of chaos due to this, particularly when combined with resource cloning so that the checksum would get set at the end of the converge for reporting and then it would get copied to the cloned resource which would tend to explode, so there was a workaround implemented for that to have a different property that the reporter used for the reported checksum.

Either that workaround should get applied to the file modes, or at some point we should just have a `final_resource` which we construct for all provider which probably clones the `new_resource` fixes it up and then passes it off to reporting.

Anyway, what happened here is that now that we have before blocks, we expect to why run first and then run the resource again afterwards.  This interacts badly with resource cloning and the values are read into the new_resource and mutate it and then the desired state for the actual converge to do work completely changes and it breaks before notifications completely.

So TL;DR:   the mutation of the new_resource in non-why run is there for a reason, not mutating it in why-run fixes before notifications, and fixing it right is a large yak shave due to $REASONS.

PR #5758, opened 2017-01-25
Update changelog task to match the new labels

tas50:
There might be more labels we want to add to the list, but this gets us a working changelog again


thommay:
I'm kind of against doing this on a per-repo basis, gonna drop a PR on our gh-c-g with some defaults in the morning.

tas50:
I'm all for defaults.

PR #5762, opened 2017-01-26
Install default omnibus-software OpenSSL version

tduffield:

### Description

Do not override the OpenSSL version to install. Instead, use what is default in omnibus-software. This will keep the version of OpenSSL used in chef-client insync with what is used in ChefDK.

### Check List

- [ ] All tests pass


tduffield:
We're so close to having chef and chef-dk back on master, I say we just wait for that.

Issue #2850, opened 2015-02-02
powershell_script user attribute: no method for uid


docwhat:
This is copied from [CHEF-5344](https://tickets.opscode.com/browse/CHEF-5344):

> using the `powershell_script` resource and documentation states also has the `user` attribute like execute or bash on linux.
> The `powershell_script` resource executes the script fine, unless the `user` attribute is supplied. I used a local user account which chef created for testing purposes.
> I saw a similar ticket for the powershell cookbook that julian dunn closed recently, but nothing for chef

This is still a problem in Chef-Client 12.0.3

It seems `Mixlib::ShellOut` has the ability to change user (requiring a password, and possibly a domain), but the `powershell_script` resource/provider hasn't been updated.

docwhat:
I searched, but couldn't find the ticket from the powershell cookbook mentioned.

alex-williams:
I've hit this too and it's standing in the way of some important stuff. I've got a different issue with joining a Windows 2012 domain, because SYSTEM on 2012 does not appear to be allowed to perform the action of joining the domain. This was the workaround, to use the user feature in powershell_script, to run the commands necessary for joining the domain, in a different script, as a different user.

docwhat:
If you use the `chef-client::task` instead of the windows service you can have it run as Administrator. As a work around.

alex-williams:
I'll give that a try if my current test fails. Thanks @docwhat

smurawski:
@alex-williams The domain join problem is mostly unrelated.  The only way domain join works under the local Administrator account is when the local Administrator password is the same as the domain's Administrator password and NTLM auth just matches that up.  No local user account (or service account) has rights to join a machine to a domain.  Only domain accounts with rights to join a computer account can do so.  In a default configuration, active directory will let a normal user account join 10 machines to the domain.  Most production domains I've worked on disable that and require specific accounts to be used to join machines to domains.

smurawski:
@docwhat Agreed this is a bug and should get some attention. Paging @btm, @jdmundrawala, @ksubrama :)  Likely won't make it for the 12.1.0 release, but we'll get some attention on it.

alex-williams:
[...]

PR #5722, opened 2017-01-13
Windows alternate user support for remote_file and execute resources

NimishaS:
### Description

Combined PR for https://github.com/chef/chef/pull/4299 and https://github.com/chef/chef/pull/4297

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #5764, opened 2017-01-27
Windows alternate user support for execute resources

NimishaS:
### Description

w.r.t https://github.com/chef/chef/pull/4297
Also includes fixing of review comments, bugs and specs from https://github.com/chef/chef/pull/5722

### Issues Resolved

https://github.com/chef/chef/issues/2850

### Check List

- [x] All tests pass


btm:
This replaces https://github.com/chef/chef/pull/4297. We originally started by combining the `remote_file` support in https://github.com/chef/chef/pull/4299 again, but realized there's a use case we need to fix before we release any new properties. It'd be significant progress to get this feature merged.

lib/chef/provider/execute.rb
```diff
@@ -39,7 +42,11 @@ def whyrun_supported?
       end
 
       def define_resource_requirements
-         # @todo: this should change to raise in some appropriate major version bump.
+        # @todo: this should change to raise in some appropriate major version bump.
```
afiune: do we still need this comment?

lib/chef/provider/script.rb
```diff
@@ -66,10 +67,45 @@ def action_run
       end
 
       def set_owner_and_group
-        # FileUtils itself implements a no-op if +user+ or +group+ are nil
-        # You can prove this by running FileUtils.chown(nil,nil,'/tmp/file')
-        # as an unprivileged user.
-        FileUtils.chown(new_resource.user, new_resource.group, script_file.path)
+        if ! Chef::Platform.windows?
```
afiune: should we use `unless`?

afiune: Maybe this could be read better like:
```
if Chef::Platform.windows?
  grant_alternate_user_read_access
else
  FileUtils.chown(new_resource.user, new_resource.group, script_file.path)
end
```

lib/chef/resource/execute.rb
```diff
@@ -143,6 +143,30 @@ def user(arg = nil)
...
+      def password(arg = nil)
+        set_or_return(
+          :password,
+          arg,
+          :kind_of => [ String ]
+        )
+      end
+
+      def sensitive(args = nil)
+        if ! password.nil?
```
afiune: Same here, switching them looks more readable:
```
if password.nil?
  super
else
  true
end
```

lamont-granquist: actually i think you can do `property :password, String, sensitive: true` and avoid this.

the domain should also be a property (you don't have to convert the whole file, but anything new should be a property).

lib/chef/provider/execute.rb
lamont-granquist: it should be moved down below to above the warn

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Mixin
+    module UserIdentity
```
lamont-granquist: so, generally i think this level of indirection with a mixin is probably the wrong approach and the logic would go better elsewhere...

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+      def validate_identity(specified_user, password = nil, specified_domain = nil)
+        validate_identity_platform(specified_user, password, specified_domain)
+        validate_identity_syntax(specified_user, password, specified_domain)
+      end
+
+      def validate_identity_platform(specified_user, password = nil, specified_domain = nil)
+        if ! Chef::Platform.windows?
+          if password || specified_domain
+            raise Exceptions::UnsupportedPlatform, "Values for `domain` and `password` are only supported on the Windows platform"
+          end
```
lamont-granquist: this should go in the resource where setting the password or the domain on not-windows should fail

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+        else
+          if specified_user && password.nil?
+            raise ArgumentError, "A value for `password` must be specified when a value for `user` is specified on the Windows platform"
+          end
```
lamont-granquist: this should be raised in an after_created block in the resource

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+      end
+
+      def qualify_user(specified_user, specified_domain = nil)
+        domain = specified_domain
+        user = specified_user
+
+        if specified_user.nil? && ! specified_domain.nil?
+          raise ArgumentError, "The domain `#{specified_domain}` was specified, but no user name was given"
+        end
```
lamont-granquist: this is also appropriate for the after_created block in the resource

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        # if domain is provided in both username and domain
+        if specified_user && ((specified_user.include? '\\') || (specified_user.include? "@")) && specified_domain
+          raise ArgumentError, "The domain is provided twice. Username: `#{specified_user}`, Domain: `#{specified_domain}`. Please specify domain only once."
+        end
```
lamont-granquist: also should be after_created in the resource

lib/chef/mixin/user_identity.rb
```diff
@@ -0,0 +1,90 @@
...
+            # Splitting username of format: Username@Domain
+            domain_and_user = user.split("@")
+            if domain_and_user.length == 2
+              domain = domain_and_user[1]
+              user = domain_and_user[0]
+            elsif domain_and_user.length != 1
+              raise ArgumentError, "The specified user name `#{user}` is not a syntactically valid user name"
+            end
+          end
+        end
```
lamont-granquist: most of this should be extracted into a helper function like `def syntactically_valid_username?` and should be moved to the resource, and
the raise should be in the after_created block again.

lib/chef/provider/execute.rb
```diff
@@ -39,7 +42,11 @@ def whyrun_supported?
...
+        requirements.assert(:all_actions) do |a|
+          a.assertion { validate_identity(new_resource.user, new_resource.password, new_resource.domain) }
+        end
```
lamont-granquist: yeah the shape of this looks pretty wrong.  you're not using either the why-run text or the exception to raise in not-why-run.  the conditional has the side effect of doing all the raises, and will run in both why-run and not-why-run the same, which means the whole define_resource_requirements API is not really being used and could be dropped.

since i don't think that validate_identity relies on doing user lookups and is static analysis of the parameters, though, that means it should really be moved to the resource (i don't think any of the syntax checks there could depend upon prior resources being run unless i missed something).  in a few cases the assertion is simple ("on unix, if password is specified you must raise") which are easy to do directly from the property declaration.  for the ones where you need collaboration ("if user is set on windows, password must also be set") then that needs to be done in the after_created block of the resource.

if there is anything which depends on the user actually being created then that would need to go into define_resource_requirements, but then you'd need to have the full define_resource_requirements API here with the exception to raise, and the text to produce in why-run mode.

lamont-granquist:
tl;dr is that the code looks okay except the validation really needs to be moved directly to the resource, would probably prefer not to create the additional mixin as well.

lib/chef/provider/execute.rb
NimishaS: The comment should be at the right place now since the lines which were added previously are removed now.

mwrock:
Great work on this @NimishaS!

NimishaS:
Thanks @mwrock. Yes I still need to fix the specs for mixin references.

btm:
@NimishaS the validations that you put in `after_created` of the resource should go in `define_resource_requirements` in the provider. The `requirements` class used here allows you to specify which actions (or all) that the requirements apply to as well as avoiding a failure in why-run mode when we would not be able to verify the requirements in why-run mode (typically due to a prior resource creating them).

NimishaS:
@btm , I got review comments by @lamont-granquist to move the validations in `after_created` of the resources https://github.com/chef/chef/pull/5764#discussion_r98286129. Previously I had called the validation method in `define_resource_requirements`.

NimishaS:
@btm , @lamont-granquist , please let me know where should the validations be kept. I would also like understand the difference between the purpose of `after_created` and `define_resource_requirements` method.

btm:
To help understand, here's a wall of text. :)

I figured `after_created` was a new pattern that I didn't understand, then I looked at the git history and saw that I added it 5 years ago as part of the original `chef_gem` implementation. Heh. I don't see a new pattern in use in the code. In theory starting to use `after_created` for validating arguments passed to properties by the user that wouldn't be affected by the state of the system would raise an error earlier, which would be good. `after_created` naturally exists to run code in the compilation phase tied to resource creation, but we use it sparingly.

By being in the provider, `define_resource_requirements` is well timed for the kind of validation that might not pass when chef-client starts but would later after prior resources have changed the state of the system. This is why it also has provisions for avoiding raising an error when in why-run mode. Because we're not actually making changes to the system in why-run mode, we have to ignore some of those failures for why-run to work.

Currently all of the property validation is done in `define_resource_requirements`. You can see a recent example of this in the [launchd provider](https://github.com/chef/chef/blob/master/lib/chef/provider/launchd.rb#L137) where we ensure that the 'type' property is either 'daemon' or 'agent'. 

I chose that example to further muddy the waters of this discussion though. The `property` feature replaces `set_or_return`. This is new and we're still not used to using it. I was thinking there may be a good feature in `property` to use for static validation. In the [launchd resource](https://github.com/chef/chef/blame/master/lib/chef/resource/launchd.rb#L90) we are overloading the `coerce` option to not only downcase but to duplicate the same validation that is in the provider. Obviously the validation is in `coerce` and not in the `type` so we can use `downcase` in that block to catch all capitalization variations of the argument. If we passed an array of strings like `[ "daemon", "agent" ]` to type we would miss an argument like "SomethingElse" and fail messily later.

From the perspective of using `property`, being guaranteed that `coerce` runs before `type` (aka checking `:is` in Chef::Mixin::ParamsValidate) would be nice and allow something concise like this:

```
property :type, [ "daemon", "agent" ], default: "daemon", coerce: proc { |type| type.downcase }
```

So. Using `define_resource_requirements` for validating properties that depend on system state that could change between when the run starts and the provider runs seems to continue to best. Using `property` going forward for validation when the property the user provides can be validated irrespective of system state would be a good path, but we seem likely to make a habit of overloading `coerce` to achieve this. 

There's an ordering problem if you need to compare one property with another though. Perhaps @lamont-granquist was thinking about using `after_created` as a compromise between the ordering problem and waiting until the provider action with `define_resource_requirements`.

There might be some validation that could be moved up if we overload `coerce` though, like:
```
property :domain, String, coerce: windows_only!
property :password, String, sensitive: true, coerce: windows_only!

def windows_only!
  unless Chef::Platform.windows?
    raise Exceptions::UnsupportedPlatform, "Values for `domain` and `password` are only supported on the Windows platform"
  end
end
```

@lamont-granquist ?

lamont-granquist:
so `coerce: windows_only!` reads wrong to me because you're not doing any coercions, we should implement some other keyword for that which doesn't have the coercion side effect and doesn't read as poorly.  that should be fairly straight forward since it's just a keyword that acts like a coercion without coercing anything.

i agree with you about d_r_r being a better place for validations which depend on system state, but it didn't seem to me like any of these validations depended on system state and it was all static analysis of the resource properties.  most of the conditional logic was based on `windows?` being true or false and that can't change during the run (and chefspec should get it correct--if it doesn't that should get fixed).  but the complexity was that the validation required looking at more than one property at a time.  so if you just failed if someone set a username without setting a password on windows and did validation in the property then this would fail:

```ruby
execute "whatever" do
   username "Administrator"
   password "foobar"
end
```

when the username is set, the password hasn't been set yet, which would fail validation then, so we have to lazy that check until the end block in order to be sure that the password has been supplied.  that is what moving the check to after_created accomplishes.

if there's a check in there which depends on system state, which might change during running, and can't be evaluated at declaration time then i missed it on my read-through.

so, i think after_created is still the most appropriate place to put the validations.  i like the idea of using the properties api, but abusing the `coerce` keyword makes me very twitchy.  i really like the effect, but i'd like to see that done right and extend the properties api to support it properly.

lamont-granquist:
Another way of saying the same thing is to look at define_resource_requirements:

```
    [...]
    requirements.assert(:all_actions) do |a|
      a.assertion { validate_identity(new_resource.user, new_resource.password, new_resource.domain) }
    end
    [...]
  end
```

That is code-smell because there's no `a.failure_message` or `a.whyrun` in that block.  All the raising is done by the assertion as a side effect.  Really it should just be written:

```
    validate_identity!(new_resource.user, new_resource.password, new_resource.domain)
  end
```

But that just highlights the fact that user, password and domain are being validated statically and there's no system state that is being inspected -- or else there would have needed to be an assertion with a `whyrun` statement so that chef would output "assuming user 'foo' would have been created" or something.  I can see that we do have requirements which are added with no `whyrun` statement but those are really just syntactic sugar for `raise <something> unless <assertion>`.  the ones that i see it does make sense to lazy the assertion since it does inspect the underlying system its just not likely that chef would ever fix the issue during a run (e.g. the check for `hive_exists?` in registry_key -- you could throw an "assuming hive would have been created" but i believe nobody ever creates new hives and those are predefined on windows -- but since it evaluates that state of the node then you want to lazy that onto the provider so it doesn't troll chefspec running on unix).

lamont-granquist:
And also putting the the user+password+domain validation on the resource *should* get the chefspec case correct.

If your fauxhai platform under chefspec is windows, then you want to fail validation if you give a username with no password.  If your fauxhai platform under chefspec is unix-ish then you want to fail validation if you give a password.  If you move the validation to d-r-r then it always passed.

Of course this assumes Chef::Platform.windows? is correctly checking node data, if it ain't then that's busted, but that indicates we should change that to inspect `node[:platform]`.

Aaaand that's busted:

```
module ChefConfig

  def self.windows?
    if RUBY_PLATFORM =~ /mswin|mingw|windows/
      true
    else
      false
    end
  end

end
```

should probably use `node[:os] == "windows"` instead of `Chef::Platform.windows?` and we need to fix that.

btm:
@NimishaS in summary, let's leave everything in `after_created` for now because in `property` the other properties we want to inspect may not be set yet, and `define_resource_requirements` runs later in the run than we need to. Please do change `Chef::Platform.windows?` to `node[:os] == "windows"` as @lamont-granquist indicated in his last message.

@lamont-granquist I agree we should add a validation option for validating the input in a block, and then go back and clean up the launchd resource where it is doing that in the coerce block.

lamont-granquist:
yeah i cut this issue to work on fixing `Chef::Platform.windows?` and making it all more usable, but its going to be more than a quick fix:

https://github.com/chef/chef/issues/5771

for now use `node["platform_family"] == "windows"`  (probably use `platform_family` over `os` since that is what chef-sugar has been using since forever, and i'm fairly strongly inclined to just copy that, but it _probably_ doesn't matter at all)

lib/chef/resource/execute.rb
```diff
@@ -165,7 +165,7 @@ def after_created
       end
 
       def validate_identity_platform(specified_user, password = nil, specified_domain = nil)
-        if Chef::Platform.windows?
+        if node[:os] == "windows"
```
btm: You should use `node[:platform_family] == "windows"` here because it may be more compatible with ecosystem tools.

https://github.com/chef/chef/pull/5764#issuecomment-276470185

lib/chef/provider/script.rb
```diff
@@ -18,6 +18,7 @@
 
 require "tempfile"
 require "chef/provider/execute"
+require "chef/win32/security" if Chef::Platform.windows?
```
NimishaS: @btm, I tried to use `node[:os]` here too, but it raises error that `node object is not defined`.

btm: yeah, because that's outside the class. If you move the require into `grant_alternate_user_read_access` and check `node[:platform_family] == windows` that'll work.

While I understand what @lamont-granquist is getting at, we've got to go around and fix all of this soon for https://github.com/chef/chef/issues/5771 anyway.

lib/chef/resource/execute.rb
```diff
@@ -135,12 +135,18 @@ def timeout(arg = nil)
...
-          arg,
-          :kind_of => [ String, Integer ]
-        )
+      property :user, [ String, Integer ]
+
+      property :domain, String
+
+      property :password, String, sensitive: true
+
+      def sensitive(args = nil)
```
NimishaS: `property :password, String, sensitive: true` was not setting `sensitive=true` for `Chef::Resource::Execute` even if password is specified. Hence I had to define `sensitive` again.

NimishaS:
@btm , @lamont-granquist , please review the PR.

lib/chef/provider/script.rb
lamont-granquist: this is actually correct because it happens so early that ohai will not have run, there can't be a node object, and chefspec shouldn't require this file either.

adamedx:
Very cool to see all the commentary on this -- thx for joining in @lamont-granquist  and @mwrock. Sorry I wasn't able to shepherd this through all the way (still a little too busy these days), but grateful that we could get the feature to users!

Issue #5362, opened 2016-09-21
systemd_unit provider verify property in underlying file resource breaks name-sensitive units


nathwill:
## Description

after merging of https://github.com/chef/chef/pull/5210, systemd_unit resource can no longer create valid units where systemd considers the unit file name to be an important (and prescribed) part of the unit definition (i.e. automount, mount, probably device units as well).

because the verify command is executed on the temporary path, this causes `systemd-analyze verify %{path}` to fail (as shown below). see also https://github.com/chef/chef/issues/5324.

example output where `systemd-analyze verify` blows up due to file name:

```
[root@default-fedora-24 ~]# systemd-analyze verify /etc/systemd/system/chef-proc-sys-fs-binfmt_misc.automount 
chef-proc-sys-fs-binfmt_misc.automount: Where= setting doesn't match unit name. Refusing.
chef-proc-sys-fs-binfmt_misc.automount: Failed to create chef-proc-sys-fs-binfmt_misc.automount/start: Unit chef-proc-sys-fs-binfmt_misc.automount is not loaded properly: Invalid argument.
```

## Chef Version

12.14+

## Platform Version

all systemd-based platforms

## Replication Case

try to converge a recipe that creates a mount unit, such as:

```ruby
systemd_unit("tmp.mount") do
  content ({
    "Mount"=>{
      "What"=>"tmpfs", 
      "Where"=>"/tmp", 
      "Type"=>"tmpfs", 
      "Options"=>"mode=1777,strictatime"
    },
    "Unit"=>{
       "Description"=>"temp dir", 
       "Documentation"=>"man:hier(7)", 
       "Conflicts"=>"umount.target", 
       "Before"=>"local-fs.target umount.target", 
       "DefaultDependencies"=>"no", 
       "ConditionPathIsSymbolicLink"=>"!/tmp"
    }
  })
  action [:create]
end
```

## Client Output

```
            ================================================================================
             Error executing action `create` on resource 'systemd_unit[tmp.mount]'
             ================================================================================
             
             Chef::Exceptions::ValidationFailed
             ----------------------------------
             file[/etc/systemd/system/tmp.mount] (dynamically defined) had an error: Chef::Exceptions::ValidationFailed: Proposed content for /etc/systemd/system/tmp.mount failed verification #<Chef::Resource::File::Verification:0x00000002cbcb20>
             
             Resource Declaration:
             ---------------------
             # In /tmp/kitchen/cache/cookbooks/systemd/libraries/resource_factory.rb
[...]

Issue #5200, opened 2016-08-14
Make systemd_unit verify it's unit file content


mal:
# Feature Suggestion

The `systemd_unit` resource has the potential to be improved by utilising the `verify` functionality in it's [`file` block](https://github.com/chef/chef/blob/ba819110a6643d1f65e6b20650e5c8905b083aa8/lib/chef/provider/systemd_unit.rb#L191-L196). For example, using: `systemd-analyze verify %{path}`.

ccope:
I think this should be reopened, since this change was reverted in #5362 
I just shot myself in the foot while trying to install a systemd mount unit, and had to work out how to run the systemd verify tool myself to figure out what I did wrong. I *think* that simply writing the unit file into /etc/systemd isn't enough to enable it, it then has to be symlinked when you run `systemctl enable`? So maybe Chef can just write it there, verify, and delete it if it fails? (disclaimer: i'm a systemd noob)

mal:
@thommay please could you reopen this issue? The original solution had to be reverted but [subsequent discussion](https://github.com/chef/chef/issues/5362#issuecomment-248715590) agreed that the feature in theory still had merit?

nathwill:
@thommay i think we can close this one now, as the fix is merged in #5765

thommay:
Thanks!

PR #5765, opened 2017-01-28
Verify systemd_unit file with custom verifier

mal:
Adds custom file verification class to work around file name requirements when running `systemd-analyze`.

This is a second attempt at adding this functionality after the first (#5210) had to be reverted. The problem lay in the fact that `systemd-analyze` checks that attributes within the unit match the name of the unit as determined by it's filename. Due to the way in which the Chef file provider manages temp files, the name would not match and `systemd-analyze` would fail the file.

This second approach addresses this problem by providing a custom file verification class which copies the temp file to a new temp directory under it's real name and verifies it there instead, before cleaning up the temp directory and by extension the second temp file.

### Issues Resolved

- #5200 (original feature request)
- #5362 (bug that triggered the roll back of the first attempt)

### DIFF
--- a/lib/chef/resource/file/verification/systemd_unit.rb
+++ b/lib/chef/resource/file/verification/systemd_unit.rb
@@ -1,0 +1,67 @@
+#
+# Author:: Mal Graty (<mal.graty@googlemail.com>)
+# Copyright:: Copyright 2013-2017, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/mixin/which"
+
+class Chef
+  class Resource
+    class File
+      class Verification
+
+        #
+        # Systemd provides a binary for verifying the correctness of
+        # unit files.  Unfortunately some units have constraints on the
+        # filename meaning that normal verification against temp files
+        # won't work.
+        #
+        # Working around that requires placing a copy of the temp file
+        # in a temp directory, under it's real name and running the
+        # verification tool against that file.
+        #
+
+        class SystemdUnit < Chef::Resource::File::Verification
+          include Chef::Mixin::Which
+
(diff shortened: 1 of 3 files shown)


mal:
@nathwill following https://github.com/chef/chef/issues/5362#issuecomment-248715895 it'd be good to get your feedback on this

nathwill:
@mal awesome, thanks for this! First impression is that this looks right to me; I'll try to run some local testing on it this afternoon as well.

nathwill:
so, after testing this, it definitely seems to do what's intended. it should be noted that systemd-analyze does exit with status 0 given some kinds of "ignorable" syntax errors... so this won't catch everything, but having a unit verifier in place to catch the grossest syntax errors is still a great idea IMO, so i'm 👍.

lib/chef/resource/file/verification/systemd_unit.rb
```diff
@@ -0,0 +1,67 @@
...
+      class Verification
+
+        #
+        # Systemd provides a binary for verifying the correctness of
+        # unit files.  Unfortunately some units have constraints on the
+        # filename meaning that normal verification against temp files
+        # won't work.
+        #
+        # Working around that requires placing a copy of the temp file
+        # in a temp directory, under it's real name and running the
```
stevendanna: nit: it's -> its

lamont-granquist:
code looks superclean.

stresler:
Hi, I think this is a good feature overall, and thanks for creating it. 

However, is it at all possible to bubble up the validation error that actually occurs. 

I'm just using chef-solo to build AMIs with Packer.  Our build failed, and I finally had to move to debug mode, log into the aws instance, copy the unit out of the 'content' field into a test .service file, and then run 

```
[root@ip-10-40-16-52 recipes]# /usr/bin/systemd-analyze verify /tmp/calico-test.service 
Error: org.freedesktop.systemd1.LoadFailed: Unit docker.service failed to load: No such file or directory.
Failed to create calico-test.service/start: Input/output error
calico-test.service: command /usr/bin/docker is not executable: No such file or directory
calico-test.service: command /usr/bin/docker is not executable: No such file or directory
```

I didn't really care about the order of the install, because this was just generating an image, but now I have to care about it, which is totally acceptable. The harder part was the only error I was getting was: 

```
 ERROR: systemd_unit[calico-node.service] (pr_base::calico line 26) had an error: Chef::Exceptions::ValidationFailed: file[/etc/systemd/system/calico-node.service] (dynamically defined) had an error: Chef::Exceptions::ValidationFailed: Proposed content for /etc/systemd/system/calico-node.service failed verification #<Chef::Resource::File::Verification:0x0000000b5be720>
```

Let me know if I should just open a new issue for this. Thanks.

ccope:
Hm yeah I think this should be a new issue. You can reference this one if you like.

PR #5772, opened 2017-01-31
Use CIDR notation rather than netmask in route-eth0 file

tduffield:
### Description

Convert the netmask into it's CIDR block for the route-eth0 file.

Everything I've found in terms of documentation says that the CIDR block is perfectly acceptable in the route-eth0 file. Unless I missed something that's probably what we want to use as RHEL does not accept the netmask. 

### Issues Resolved

* Internal ZD-12741

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/route.rb
+++ b/lib/chef/provider/route.rb
@@ -219,7 +219,7 @@ class Chef
         case action
         when :add
           content << (options[:target]).to_s
-          content << "/#{options[:netmask]}" if options[:netmask]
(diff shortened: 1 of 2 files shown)


lamont-granquist:
yeah it looks like we write out the CIDR anyway so that's presumably been working, and it looks like the same format back to centos 5.2:

https://www.centos.org/docs/5/html/5.2/Deployment_Guide/s1-networkscripts-static-routes.html

so LGTM

PR #5782, opened 2017-02-06
rhel7 / dnf 2.0 fixes / improved errors

lamont-granquist:
- fixes for dnf 2.0 / rhel7
- improved error messages from python helper (gives python stack
  traces instead of just EPIPE all the time)
- improved which/where interface in the which mixin.

### DIFF
--- a/lib/chef/mixin/which.rb
+++ b/lib/chef/mixin/which.rb
@@ -18,15 +18,27 @@
 class Chef
   module Mixin
     module Which
-      def which(cmd, extra_path: nil)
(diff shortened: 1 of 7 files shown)


lamont-granquist:
@chef/client-core this moves the needle a lot but rhel7/dnf2.0 still have some busted tests (and still nothing that runs in CI to cover it, i'm not certain what to do here) but this should start to unblock @jaymzh 

proxy tests will most likely fail because they're broken on master

lamont-granquist:
@chef/client-core need review + merge

PR #5793, opened 2017-02-08
Sanitize UTF-8 data sent to Data Collector

lamont-granquist:
new data collector re-introduced the same kinds of bugs that e.g. #3320 fixed


lamont-granquist:
@adamleff and @tduffield

tduffield:
@lamont-granquist is this to address the COOL ticket we were discussing?

lamont-granquist:
yeah i think this fixes that COOL ticket

PR #5794, opened 2017-02-08
Fix chef_platform_methods deprecation class name

biinari:
Deprecation warnings for Chef::Platform methods introduced in #5631 were
logged as e.g. `"Class.set is deprecated"` instead of `"Chef::Platform.set is
deprecated"`

This gets correct class name for message and adds unit specs to cover
the expected deprecation messages.

Unfortunately for these specs, some of these methods use each other,
resulting in multiple warnings. I have put an expectation for each of
these to avoid masking any other deprecations.

Signed-off-by: Bill Ruddock <bill.ruddock@gmail.com>

### DIFF
--- a/lib/chef/platform/provider_mapping.rb
+++ b/lib/chef/platform/provider_mapping.rb
@@ -35,7 +35,7 @@ class Chef
       include Chef::Mixin::ParamsValidate
 
       def find(name, version)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@thommay should be simple to review provided it goes green (modulo the red proxy tests issue)

biinari:
Thanks :smile:

PR #5784, opened 2017-02-07
Updated release docs for alternate user identity support in execute resources

NimishaS:
### Description

Release docs for https://github.com/chef/chef/pull/5764

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [x] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #5796, opened 2017-02-09
Be more specific about using the alternate user feature

btm:
Replaces https://github.com/chef/chef/pull/5784


btm:
These are RELEASE_NOTES for https://github.com/chef/chef/pull/5764.

PR #5800, opened 2017-02-10
Numerious minor tweaks to unblock the build

tduffield:
### Description

* Unblock the builds
* Fix dependency update steps so that they are more reliable when run locally 
* Update dependencies 
* Unpin gems that no longer need to be unpinned
* Update the Bundler groups 

### Issues Resolved

* Broken builds

### Check List

- [ ] All tests pass


tduffield:
Confirmed this PR passes all internal tests

acceptance/Gemfile
```diff
@@ -1,21 +1,16 @@
 source "https://rubygems.org"
 
+gem "rake" # required to build some native extensions
```
btm: oh man. so many environment issues have trolled me with not having rake in Gemfiles. 👍

ci/dependency_update.sh
```diff
@@ -4,6 +4,6 @@ set -evx
 
 . ci/bundle_install.sh
 
-bundle exec rake dependencies
+bundle exec rake dependencies_ci
```
btm: @tduffield is this intentional?

tduffield: Yes! :) dependencies is for humans, which basically triggers ci/dependency_update.sh now. If you run `rake dependencies` (now _ci) it typically won't work on your local machine because of the multi-os-ness of the gem deps

PR #5803, opened 2017-02-10
Suppress sensitive properties from resource log and reporting output

### LINKED ISSUE #5473: Properties marked sensitive should not be output in Chef::Resource#to_text
See #5471 

https://github.com/chef/chef/blob/11d64c2438d6e387986adce3d4740f2bc7efaab9/lib/chef/resource.rb#L668-L683

That would always suppress the svn_password property even if the overall resource is not marked sensitive.

tduffield:

### Description

When a property is marked sensitive, suppress it when printing a text representation of the resource for logging purposes. 

### Issues Resolved

* ZD-13635 (internal ticket)
* COOL-680 (internal ticket)
* Fixes #5473 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -670,16 +670,18 @@ class Chef
       text = "# Declared in #{@source_line}\n\n"
       text << "#{resource_name}(\"#{name}\") do\n"
 
-      props = []
-      self.class.state_properties.each do |p|
-        props << "@#{p.name}".to_sym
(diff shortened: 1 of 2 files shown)


tduffield:
@coderanger That's good feedback. It wasn't too hard to rework the logic to prevent the "grouping"

coderanger:
Just as a question: is there a reason to not do this in `Property#to_text`? That might fix it in a bunch of places at once.

coderanger:
Oh, because there is no such method, that would be a good reason.

PR #5811, opened 2017-02-11
Bump the version of chef depending on the label associated with PR

tduffield:
### Description

Introduces new functionality to our CI system that allows us to control the type of version bump we want for a given PR based on the Labels applied. With this, we can do proper semver while still automatically bumping our versions as necessary. 

As it stands, the existing behavior of bumping the patch level on every merge will continue. Once we have the necessary labels created we can migrate to using those. 

In the near future we plan to roll out a Github status check that will ensure we have the appropriate labeling applied to a PR before it can be merged. 

### Check List

- [ ] All tests pass


lamont-granquist:
Should probably go to the developer's meeting and an RFC.

tduffield:
@lamont-granquist is there a specific aspect of this that you think requires an RFC? There's nothing in RFC047 that specifically mentions the method by which we bump our version, only that we adhere to semver.

lamont-granquist:
We currently have a monthly release cadence of minor versions which people have come to expect, if we're planning on changing that it should probably go through the RFC process, even if it isn't written down anywhere yet (it really should be).

coderanger:
So major bumps should definitely get removed IMO, the risks of a bad bump seem to me to outweigh the work of doing it by hand once a year. For minor revs, this seems like it would do one minor rev per tagged PR, which is not how I think you meant this to work? Regardless, I would like to see a "plot explanation" as to the goals of the change and reason for it. An RFC works or you can write something up in here :-) I think there is a general feeling from the community that the release process is getting un-grokably complex so changes to it should get higher levels of scrutiny at this point.

thommay:
This is a change that gets us much closer to the actual stated aims of SemVer, as well as the process documented in RFC 47 - specifically, that when we add new _features_, rather than improvements or bug fixes, we'll bump the minor version. I don't think we need an RFC, but we are intending to communicate via the usual channels (developer meeting, discourse, etc) when this takes effect.

The main point to this change is that it makes it possible to safely do security point releases or regression releases - since right now a human having hand bumped the minor version number post release is the only demarcation, we've seen that new features get merged after a release but before that version bump, which means that we can't then do a security release for that minor without dragging along new features, too.

We'll still have a monthly cadence of releases that will not deprecate new features or bump major.

tduffield:
This PR was motivated by the recent security issue that affected mixlib-archive. The realization made by the internal Chef teams responsible for cutting new releases was that given our current version bumping model, which bumps the build/patch version on every merge PR,  Chef Maintainers can be put into a situation where a security or hot fix release could be complicated or result in confusion for the user. This complication arises out of a few practices and procedures that exist in how PRs against chef/chef are currently handled. 

I’ll start with a review on how Chef is released. Per RFC-087,  once a month (on an agreed upon date) we select a build from the current channel. This is typically the most recent build, but sometimes it is not. The reasons vary but there are many factors that go into deciding which build to promote and we can’t always be sure we’re going to pick the most recent build. Maybe the most recent build has an experimental feature and the consensus is that it’s not ready. Maybe X.Y.Z has a bug, but X.Y.Z-1 does not so to try and stay within the timeline provided by RFC-087 we select X.Y.Z-1. Sometimes it is a better choice _not_ to select the most recent build but instead some build from the recent past. 

After the build has been selected we bless this build and notify the community that it X.Y.Z is our release candidate. Per RFC-087, there is a waiting period before we can formally promote that build to stable. During this time, there is currently no “code freeze” on master, meaning that new feature code could be merged to master resulting in builds for X.Y.Z+1 and X.Y.Z+2. The current practice is not to immediately bump the minor version to X.Y+1 (though doing so would still not protect us against scenarios mentioned in the previous paragraph). 

So, the current scenario we are in is that X.Y.Z could have an issue that requires a security or hot-fix. With the current processes (and other factors that I outlined above) there is no guarantee that version X.Y.Z+1 is available or that X.Y.Z+N would not contain new functionality that was not given adequate time to be tested during the waiting period. The end result is that to get the security or hot-fix out we would either need to a) also include the new functionality, b) release X.Y.Z+N (but not include the functionality that was introduced in the interim), or c) release the hot fix as X.Y+1.0 and make next month’s release X.Y+2.Z. 

So when looking to address this problem a number of solutions came up but the current design was motivated around the following guideline from RFC-047:

```
Given a version number MAJOR.MINOR.BUILD:

Increment the MAJOR version when the release will break existing functionality
Increment the MINOR version when you add any new features
The BUILD version is the same as PATCH, but is increased automatically by the build system
```

The guideline regarding the MINOR version is not a guideline we currently follow. Right now the MINOR version gets bumped once per month regardless of whether any new features were added. Additionally, the BUILD/PATCH version will get bumped on every PR merge even if the resulting Chef gem/binary would have no changes (in the cases where documentation, travis, or other non-binary related files are the only ones changed) resulting in the potential situation where X.Y.Z and X.Y.Z+1 are functionally identical. 

During my analysis I determined that it would be pretty simple to add new functionality to chef/chef’s version bumping logic (housed within chef/chef’s Rakefile) that could bump the version based on the labels associated with the PR, thereby giving control to Maintainers to determine version bumping at the time of PR review. This label-based process would be functionally equivalent to manually bumping the version in that it would have the following benefits:

1. We could decide not bump the version (and thus not trigger a build) if it was not necessary (in the case of changes to non-binary files such as tests, documentation, etc). 
2. We could give Chef Maintainers more fine-grained control over whether a patch or minor bump was necessary on a per-PR basis, allowing us to be in compliance with RFC-047. 
3. Because of the above, security and hot-fixes could be easily cherry-picked to MAJ.MIN “branches” even after new functionality is merged. 

Now, using the label-based automation rather than manually bumping the version continues to give us the following benefits:

1. The process is easily repeatable and does not require an additional PR. 
2. Because every bump still results in a build, it would not increase the time it would take for users to receive fixes via the current channel.
3. It does not add additional burden on Maintainers to manually bump and trigger new releases. 
4. Whether or not a PR deserved a MINOR bump could be changed without requiring a new commit be pushed to the branch.  This could be useful if it was determined that several “new feature” PRs should be merged but only have a single MINOR version bump. 
5. This same process, if successful, could be easily applied to other Chef OSS projects, resulting in a consistent experience for maintainers across a variety of products. 

Now I understand there is some concern that the change in versioning structure could cause confusion with users. To that I might argue that the bump to Chef 13 would be a perfect time to establish a new versioning pattern. We’ve already made it clear to our users that things are going to change post Chef 13, and tweaking the rate at which MINOR versions increment I think will be relatively minor. It might even have some unintended positive benefits to users who install from current in their ability to live on the edge but still have some control on the new functionality that enters into their systems. 

I also want to emphasize that the way in which we communicate with the users about stable releases will not change. We will still make monthly releases where the release notes will contain all the changes since the previous stable release and the changelog would also continue to make stable releases it's section headers (regardless of how many minor release bumps occurred during that time). 

With resources like discourse.chef.io and  downloads.chef.io I think that users will have access to all the information they need to know what the latest version of Chef is, how many releases are they behind, and what has changed since their last release. I know that the Chef Community team, and Chef Software as a whole, is heavily invested in ensuring that users are successful with Chef. If the switch to adhering to RFC-047 leads to confusion amongst our users I’m confident that we will find ways to address that confusion. 

If you have any further questions or clarifications, please let me know and I’ll be happy to speak to them asynchronously.

lamont-granquist:
You're still going to have the problem where PRs labelled as "patches" are still being merged to master without bumping the minor version.  That means if you pick 13.2.45 as the release candidate it is entirely possible that 13.2.46, 13.2.47 and 13.2.48 are submitted after the rc has been picked and only labelled to do a minor bump.

The problem that I see with that is that those assignments are being made by humans, and there's no agreement over what constitutes a "patch" or a "minor" release.  The SemVer policy itself is fairly vague on this statement, and its particularly vague when its applied to chef-client and the definition of what is a formally documented public API and what is not becomes blurry.

And to backtrack a bit on history, RFC-047 is not the latest agreement on what our version numbering policy means.  At the time that all of the gem locking rake tasks were added into the chef/chef project there was a meeting where it was discussed as to what the version numbers on the project meant.  At that meeting it was agreed upon that we'd modify SemVer.  And one of the reasons why we would do that is that there was no reliable definition of what minor and patch release were.  Since we were adopting a more continuous delivery model then we would treat the minor number as the blessed release number, and the patch number would become a build number.   This agreement was made in a meeting in Chef HQ and apparently never made it into any RFC.

The way that I see it, if you want to really ensure that we can do security releases of the latest release of chef-client then we need to clearly think about what we're doing which is supporting a branching model.  Once we have picked an rc we need to not do any subsequent patch release and the minor number needs to be incremented.  So when 13.2.45 is picked as a release candidate master becomes 13.3.0 for the next commit.  Then if there are build issues or subsequent security issues those get cherry-picked onto 13.2.46, 13.2.47, etc.

Another model is that we simply adopt that the minor version becomes the build version and every commit is auto bumped so that 13.45.0 becomes the release candidate.  Then 13.45.1 becomes available as the next patch release.  Most releases in currently would then be dot-zero releases, and we would have very large build numbers (13.231.0 would be the 231st commit past 13.0.0 on master).

The model being proposed does not seem to me to address the problem of additional commits past the rc since it allow "patchy" commits.  And it clearly opens up the hole where someone submits a "patchy" commit and then later does an "oopsie that should have been a minor version bump" but with properly immutable artifacts there's no way to renumber them and erase that mistake.

lamont-granquist:
Also one other thing is that I brought up this problem in that meeting at Chef HQ and wanted it addressed.  I was told very clearly that the new CI/CD way of doing things was that master just rolled forwards and that we would never branch for security fixes, and the train just kept running straight down the tracks.

I had my doubts about that, and the person who was most vocal about that model has left the company and it seems that we want to address this problem now.  However, I do not see how this proposal even fixes the problem.

coderanger:
So to summarize what I said at the dev meeting, I think before we discuss the specifics of this proposal we need to drop down a level and talk about semantic information in version numbers. I think there are somewhat distinct camps on this and without some agreement (or at least someone above me on the org chart picking a side and saying we're doing that now) I think we're going to keep talking past each other. I will try to summarize both but since I am very biased please call me out if I'm strawman-ing here.

On one side is the view that version numbers are largely consumed by automated tools and that as long as we follow SemVer, the specifics aren't super important. In this view the important semantic data when comparing two version numbers (other than major rev for compat level which I think we all agree on) is just "newer" or "older" (or "same" but again, ignoring that case). I think the argument here is that in general people tend to set aside time to upgrade based on their own schedule (which was discussed a lot in the CalVer RFC and related stuffs) and are going to just pick whatever version is current and upgrade to that with whatever work that entails.

The other side is saying that having a rough "distance" measure between two version numbers is useful beyond just knowing which is more recent. Seeing 12.9 and then 12.14 you can get a rough idea of how far apart in both time and therefore code without any additional information.

Further complicating this talk about version numbers is the fact that we've already done this conversion on the patch number, effectively making it only useful for relative comparisons. I think that case is very different because the goal from the start was that we would only ever promote one (or maybe two if we need to fix a blocker) version for a given minor release. This is now reflected in how we talk about version numbers, instead of talking about 11.16.2 we just say 12.18 and that's generally unambiguous. This is very different from the effect of de-information-ing the minor number.

Given the very strong user expectations that 13.0 will be followed by 13.1, combined with the benefits from being able to judge "distance" between two stable releases, I think that we should try to preserve these properties in the version number. We could still do that with a label-based auto-bump, but I'm not sure it would be worth the extra complexity. That would mean that once a month someone would have to submit a bumper PR with a label rather than a PR that just changes the version like we do now.

I would also like to echo Lamont's concerns that even if we do end up with the minor rev being a monotonic incrementer and nothing more, that this is still not a great solution to the original problem because the "CD from master all the time" structure is itself the part that makes hotfix branching difficult. This was understood when we moved to it, that we just wouldn't do those and would require users to get other changes sometimes. I'm still not a huge fan of the CD approach and I'm totally on-board with making hotfix branches easier but I think it's going to require a much deeper re-evaluation of how we do releases.

lamont-granquist:
To throw another possible solution inspired from Applesoft BASIC into the mix:

We can continue to abuse the patch version and have it autoincrement by 10 or even 100 on every commit.  That leaves open future versions for branching after every single commit.  If we went with 100x then that leaves open two digits which could be used to communicate to users, where the ones digit could be simple "oopses" while the tens would get incremented for security releases.  So 12.19.3300 would be the RC, 12.19.3301 would be the first "oops", while 12.19.3310 would be the "openssl broke the world (again)" release.  Of course that only leaves 10 slots open for "oops"es.

Basically, that keeps the minor number human-meaningful, while abusing the patch number to accomplish whatever goals we want.

tduffield:
I think it'd be useful to shift the conversation over to https://github.com/chef/chef-rfc/pull/250.

PR #5814, opened 2017-02-13
fix specs: RedundantReturn, RedundantSelf, RedundantBegin

lamont-granquist:
department of redundancy department


lamont-granquist:
@chef/client-core hopefully this all makes sense.  need a thumb or two.

PR #5816, opened 2017-02-14
coerce immutable arrays to normal arrays in the yum_package resource

### LINKED ISSUE #5812: Node attributes are read-only on package install with 12.19
I'm seeing the following failure in the tftp cookbook with 12.19 in Travis that doesn't occur locally with 12.18.

```
Recipe: tftp::server
  * yum_package[tftp-server] action install
    
    Error executing action `install` on resource 'yum_package[tftp-server]'
    
    Chef::Exceptions::ImmutableAttributeModification
    ------------------------------------------------
    Node attributes are read-only when you do not specify which precedence level to set. To set an attribute use code like `node.default["key"] = "value"'
    
    Resource Declaration:
    ---------------------
    # In /opt/kitchen/cache/cookbooks/tftp/recipes/server.rb
    
     26: package node['tftp']['pkgs']
     27: 
    
    Compiled Resource:
    ------------------
    # Declared in /opt/kitchen/cache/cookbooks/tftp/recipes/server.rb:26:in `from_file'
    
    yum_package("tftp-server") do
      package_name ["tftp-server"]
      action [:install]
      retries 0
      retry_delay 2
      declared_type :package
      cookbook_name "tftp"
      recipe_name "server"
      flush_cache {:before=>false, :after=>false}
    end
    
    Platform:
    ---------
    x86_64-linux
    
Running handlers:
[2017-02-12T17:24:30+00:00] ERROR: Running exception handlers
Running handlers complete
[2017-02-12T17:24:30+00:00] ERROR: Exception handlers complete
Chef Client failed. 0 resources updated in 12 seconds
[2017-02-12T17:24:30+00:00] FATAL: Stacktrace dumped to /opt/kitchen/cache/chef-stacktrace.out
[2017-02-12T17:24:30+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2017-02-12T17:24:30+00:00] ERROR: yum_package[tftp-server] (tftp::server line 26) had an error: Chef::Exceptions::ImmutableAttributeModification: Node attributes are read-only when you do not specify which precedence level to set. To set an attribute use code like `node.default["key"] = "value"'
[2017-02-12T17:24:30+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Converge failed on instance <default-centos-6>.  Please see .kitchen/logs/default-centos-6.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

lamont-granquist:
as noted in comments this works around bugginess in the yum
provider that require a major refactor to fix.

closes #5812

### DIFF
--- a/lib/chef/resource/yum_package.rb
+++ b/lib/chef/resource/yum_package.rb
@@ -1,6 +1,6 @@
 #
 # Author:: AJ Christensen (<aj@chef.io>)
-# Copyright:: Copyright 2008-2016, Chef Software, Inc.
+# Copyright:: Copyright 2008-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -24,8 +24,17 @@ class Chef
       resource_name :yum_package
       provides :package, os: "linux", platform_family: %w{rhel fedora}
 
-      # Install a specific arch
-      property :arch, [ String, Array ]
+      # XXX: the coercions here are due to the provider promiscuously updating the properties on the
+      # new_resource which causes immutable modification exceptions when passed an immutable node array.
+      #
+      # <lecture>
+      # THIS is why updating the new_resource in a provider is so terrible, and is equivalent to methods scribbling over
+      # its own arguments as unintended side-effects (and why functional languages that don't allow modifcations
+      # of variables eliminate entire classes of bugs).
+      # </lecture>
+      property :package_name, [ String, Array ], identity: true, coerce: proc { |x| x.is_a?(Array) ? x.to_a : x }
+      property :version, [ String, Array ], coerce: proc { |x| x.is_a?(Array) ? x.to_a : x }
+      property :arch, [ String, Array ], coerce: proc { |x| x.is_a?(Array) ? x.to_a : x }
 
       property :flush_cache,
         Hash,
--- a/lib/chef/resource/package.rb
+++ b/lib/chef/resource/package.rb
@@ -1,7 +1,7 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
 # Author:: Tyler Cloke (<tyler@chef.io>)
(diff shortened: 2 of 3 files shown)


lib/chef/resource/yum_package.rb
```diff
@@ -24,8 +24,17 @@ class YumPackage < Chef::Resource::Package
...
-      # Install a specific arch
-      property :arch, [ String, Array ]
+      # XXX: the coercions here are due to the provider promiscuously updating the properties on the
+      # new_resource which causes immutable modification exceptions when passed an immutable node array.
+      #
+      # <lecture>
+      # THIS is why updating the new_resource in a provider is so terrible, and is equivalent to methods scribbling over
+      # its own arguments as unintended side-effects (and why functional languages that don't allow modifcations
+      # of variables eliminate entire classes of bugs).
+      # </lecture>
```
coderanger: maxresdefault-1

coderanger:
Just for others the might find this in the future, `Chef::Node::ImmutableArray#to_a` does a copy internally so this isn't just a "cast away the const-ness".

lamont-granquist:
yes i used that because i know it does an inherent `#dup`

for chef 13 we really need to freeze the string things in immutable hash/arrays and make sure that dup really dups the whole thing (because i think you'd find right now that if i mutated the strings with `<<` and not just the array elements that chef would be fine with that and the default level attr that i set would be getting updated-in-place -- which is the kind of thing which led facebook to mandate calling `#to_hash` all over the place in order to dup attributes before playing around with them...  which is probably way more than anyone ever wanted to know...)

lamont-granquist:
@chef/client-core today's version of coding horror for your review and amusement...

PR #5818, opened 2017-02-15
cleans up the remaining new/current_resource ivars in remaining provider code

lamont-granquist:
switches from using ivars to accessors for getters

(and yeah if you're wondering i've thought about adding a chefstyle/rubocop rule to enforce this style...  totally thought about it... ;) )

### DIFF
--- a/lib/chef/provider/cookbook_file.rb
+++ b/lib/chef/provider/cookbook_file.rb
@@ -36,15 +36,15 @@ class Chef
       end
 
       def load_current_resource
-        @current_resource = Chef::Resource::CookbookFile.new(@new_resource.name)
+        @current_resource = Chef::Resource::CookbookFile.new(new_resource.name)
         super
       end
 
       private
 
       def managing_content?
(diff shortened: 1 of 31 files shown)


lamont-granquist:
@chef/client-core mechanical code change.

stevendanna:
gif-keyboard-12021361866156310935

lib/chef/provider/cookbook_file.rb
```diff
@@ -36,15 +36,15 @@ def initialize(new_resource, run_context)
       end
 
       def load_current_resource
-        @current_resource = Chef::Resource::CookbookFile.new(@new_resource.name)
+        @current_resource = Chef::Resource::CookbookFile.new(new_resource.name)
```
thommay: It saddens me a little that we don't have some sugar around this pattern, given how often we repeat it.

lamont-granquist: we've got `load_current_value` which is that plus more.

PR #5832, opened 2017-02-22
Alternate user local logon authentication for remote_file resource

NimishaS:
### Description

Extending https://github.com/chef/chef/pull/4299

### Check List

- [x] All tests pass


NimishaS:
@mwrock , @btm , please review this PR.

Just a thought; does it make more sense to have the feature for local logon with alternate user credential in `file` resource rather than `remote_file` resource?
`remote_file` resource can have support for just remote logon with alternate user credential.

lib/chef/resource/remote_file.rb
```diff
@@ -131,6 +131,78 @@ def show_progress(args = nil)
...
+        if node[:platform_family] == "windows"
+          if specified_user && password.nil?
+            raise ArgumentError, "A value for `password` must be specified when a value for `user` is specified on the Windows platform"
+          end
+        else
+          if password || specified_domain
+            raise Exceptions::UnsupportedPlatform, "Values for `domain` and `password` are only supported on the Windows platform"
+          end
+        end
+      end
```
mwrock: This seems like duplicated logic with the resource requirements above.

lib/chef/provider/remote_file/fetcher.rb
```diff
@@ -24,6 +24,9 @@ class Fetcher
 
         def self.for_resource(uri, new_resource, current_resource)
           if network_share?(uri)
+            if !Chef::Platform.windows?
+              raise Exceptions::UnsupportedPlatform, "Fetching the file on a network share is supported only on the Windows platform. Please change your source: #{uri}"
+            end
```
mwrock: Actually looking closer I'm wrong. This only makes sence on Windows unless using a smbclient which we dont do here.

mwrock:
I think we want this feature on the remote_file resource because the identity running the chef client is likely not to have access to remote file resources as opposed to local resources. Am I understanding your question correctly?

NimishaS:
@mwrock , this PR supports only `local logon` i.e. this will allow access to only the local files with alternate user credentials.

mwrock:
Hmm. That does indeed seem flawed then since the whole point of this resource is to access files from a remote UNC path. If you run this pr to access `\\some other host\share` that is restricted to the credentials passed to this resource, does the resource fail? If so then we should hold off on merging. 

Can you point me to where in this PR we specify a `LOCAL_LOGON`?

NimishaS:
Yes @mwrock, the resource will fail currently for `\\some other host\share`. I have had a discussion about this with @btm previously.

The reason is that `LogonUserW` function is being used here: https://github.com/chef/chef/pull/5832/files#diff-cb5fc9c15d552c450a10d7b23f1a9260R50. This function works only for local alternate user credentials.

mwrock:
Sounds to me @NimishaS like we would need to use `REMOTE_LOGON` here for this PR to be viable. Is that a major refactoring or just a change from `LogonUserW` to a different function?

NimishaS:
@mwrock , it's a major refactoring. The card for this issue is also for `LOCAL_LOGON`. We'll have to create a separate task and separate PR for `REMOTE_LOGON`.

btm:
The purpose of this feature is to allow you to copy files to or from a remote CIFS/SMB share with `remote_file` using credentials. The local versus remote discussion here is the wrong context. We're definitely using the credentials to access a remote file, that's the goal.  The correct context of local versus remote is whether we're passing the credentials to the local system or to the remote system to gain the necessary authorization to access that remote file.

This isn't technically correct, but imagine this case:
Chef is running as Administrator as a scheduled task. It needs to access a remote file on server WEB1, using the credentials for the user 'web_service_acct'. Chef can either connect to the remote server and pass those credentials to WEB1, or it can create another processes that is logged in locally as `web_service_acct`, which then accesses the remote file.

Right now, if we can do it, we want the use case where we authenticate to the remote system instead of logging in locally. I was hoping that switching from passing LOGON32_LOGON_NEW_CREDENTIALS to LogonUserW instead of LOGON32_LOGON_NETWORK would allow this. 

I believe there will be more use cases where the remote system has a user that the local system doesn't (this wouldn't be the case in most domains) than use cases where there are restrictions in place that prevent Windows from authenticating against the remote system.

So what we need to confirm here, does this code work when this user only exists on the remote system?
```
remote_file "c:\\temp\\foo.txt" do
  source "\\\\web1\\share\\foo.txt"
  user "web_service_account"
  password 'P@ssw0rd'
end
```

NimishaS:
@btm , the code only works for localhost currently:
```
remote_file "D://test.txt"  do
  source  "\\\\localhost\\Test\\domain_test.txt"
  remote_user "username"
  remote_password "password"
end
```

I had tried passing LOGON32_LOGON_NEW_CREDENTIALS instead of LOGON32_LOGON_NETWORK to LogonUserW, but this was failing while accessing the files on localhost with alternate user.

adamedx:
This is still open -- curious as to why this isn't yet merged...

btm:
@adamedx This isn't finished. @NimishaS wasn't able to successfully test a use case where remote_file copied a file off a remote machine. 

At the time I was trying to understand the difference between LOGON32_LOGON_NEW_CREDENTIALS and LOGON32_LOGON_NETWORK to make sure we were authenticating in the right place (against the remote machine). I haven't had time to go back and troubleshoot myself.

lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,52 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
```
stevendanna: [minor] Is this author line correct?  Also the copyright line below can say 2017.

lib/chef/provider/remote_file/network_file.rb
```diff
@@ -35,13 +39,22 @@ def initialize(source, new_resource, current_resource)
...
+
+            with_user_context(new_resource.remote_user, new_resource.remote_password, new_resource.remote_domain) do
+              ::File.open(@source, "rb") do |remote_file|
+                while data = remote_file.read(TRANSFER_CHUNK_SIZE)
+                  tempfile.write(data)
+                end
+              end
+            end
+          ensure
+            tempfile.close if tempfile
```
stevendanna: Should we call fsync here?

stevendanna: Actually, since this runs on Windows only, my intuition is probably wrong here.

lib/chef/mixin/user_context.rb
stevendanna: Ah, I see from the commit history that it probably is correct.

adamedx: @stevendanna  let me know if you'd like my help moving this forward. If the issue is that we need an end to end test, I can help generate one, though honestly, the functional tests *do* actually verify that it works. 😄

adamedx: If you can get a branch with it rebased and all merge conflicts resolved, I can get the tests passing again...

NimishaS:
@adamedx , I was able to test this PR successfully for `local logon`, but I couldn't get it working for `remote UNC` path. Is this the expected behavior or am I missing something?

adamedx:
No, it should definitely work with a remote unc path. What was the stack trace when it failed? Where you using a local account on the remote machine, or were you using a domain account? If you can repro and let me know the line that failed, that would be great.

NimishaS:
@adamedx , I was using a domain account. I am creating a domain account to reproduce this issue. I will share the stack trace and other details soon.

NimishaS:
@adamedx , I tried the following recipe after setting up a domain account:
```
remote_file "E://test.txt"  do
  source  "\\\\ip-address\\test\\test.txt"
  remote_user "domain\\username"
  remote_user_password "password"
end
```
It gives me error:
```
Chef::Exceptions::Win32APIError
-------------------------------
Logon for user `<username>` failed with Win32 status 1326.
```
It seems that the code fails when `LogonUserW` method is called: https://github.com/chef/chef/pull/5832/files#diff-cb5fc9c15d552c450a10d7b23f1a9260R50. The method returns `false`.
Please find the complete stacktrace here: https://gist.github.com/NimishaS/251251f988efce68a5943a97373fa599

**Note**: The machine where from I was trying to access the file wasn't in the same domain where the remote machine was. I hope that doesn't cause any issue as suggested here https://stackoverflow.com/questions/7488786/why-doesnt-logonuser-work-for-domain-accounts in the first response.

*Separate from the issue above:*
@btm, @lamont-granquist if we run `remote_file` resource with alternate user using `chef-apply`, it will fail with error `undefined local variable or method 'node' for #<Chef::Provider::RemoteFile::NetworkFile:0x00000004d6a430>` since we are using `node["platform_family"]` to check the platform.

btm:
> Note: The machine where from I was trying to access the file wasn't in the same domain where the remote machine was. 

I suspect this is the issue. This is where we left off. In an [above comment](https://github.com/chef/chef/pull/5832#issuecomment-284463965) I said:

> Right now, if we can do it, we want the use case where we authenticate to the remote system instead of logging in locally. I was hoping that switching from passing LOGON32_LOGON_NEW_CREDENTIALS to LogonUserW instead of LOGON32_LOGON_NETWORK would allow this.

I think as written this feature requires the user on the local system that chef is running as to have sufficient privileges to authenticate to a remote system without passing a username/password, e.g. in a domain environment where you can log on to each system.

From the [MSDN article for LogonUser](https://msdn.microsoft.com/en-us/library/windows/desktop/aa378184(v=vs.85).aspx)

> The LogonUser function attempts to log a user on to the local computer. The local computer is the computer from which LogonUser was called. You cannot use LogonUser to log on to a remote computer.

Which doesn't cover the situation of "I have a username and password for a remote share."

NimishaS:
Using `LOGON32_LOGON_NEW_CREDENTIALS` works successfully for remote logon. The PR is rebased. Travis is failing for some `kitchen-tests`, which don't seem to be related to the PR.

NimishaS:
@adamedx it's merged.

Party

adamedx:
Yay! Thank you @NimishaS for persistence and @btm  and @mwrock for keeping it alive. The whole `LOGON32_LOGON_NEW_CREDENTIALS` was not something that occurred to me. :)

All right, time to grab a nightly...

PR #5826, opened 2017-02-17
Deprecate Cisco platforms

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

### Description

Chef Client will no longer be built or tested on Cisco platforms.

@chef/engineering-services 

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussion that's relevant]

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #5835, opened 2017-02-22
Removing Cisco from Maintainers files

adamleff:
Chef is no longer built on the Cisco platforms, so there is no need
for lieutenants and maintainers any longer.


tduffield:
Relevant PR: https://github.com/chef/chef/pull/5826

PR #5843, opened 2017-02-27
appveyor: uninstall bundler first

lamont-granquist:
so we can roll back when we need to

looks like we need to


lamont-granquist:
we should determine why we're so broken on bundler 1.14.5 at some point, but for now rolling back correctly to 1.12.5 which gets us essentially back to last friday's stable builds

PR #5846, opened 2017-02-28
Launchd should not load launchagents as root. 

mikedodge04:
The Launchd resource will load launchagents as root when the mac is sitting on the login window. This is never a desired outcome and is hard to recover from. This change now will now check to see if a user is logged in before loading launchagents.

### DIFF
--- a/lib/chef/provider/launchd.rb
+++ b/lib/chef/provider/launchd.rb
@@ -16,12 +16,12 @@
 # limitations under the License.
 #
 
-require 'chef/provider'
-require 'chef/resource/launchd'
-require 'chef/resource/file'
-require 'chef/resource/cookbook_file'
-require 'chef/resource/macosx_service'
-require 'plist'
+require "chef/provider"
+require "chef/resource/launchd"
+require "chef/resource/file"
+require "chef/resource/cookbook_file"
+require "chef/resource/macosx_service"
+require "plist"
 require "forwardable"
 
 class Chef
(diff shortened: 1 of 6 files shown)


mikedodge04:
@lamont-granquist when you get a chance can you take a look at this?

lamont-granquist:
you should really rebase on top of master instead of pulling a merge commit into your branch.

lamont-granquist:
shouldn't this throw an exception rather than swallowing the error and emitting a debug notice?

mikedodge04:
Sorry I missed the email about this PR, I thought it was merged with the others I had up at the time. 

>shouldn't this throw an exception rather than swallowing the error and emitting a debug notice?

No. Apples docs aren't clear about this. But when a mac is sitting at the login window, the console user is root. If chef were to run it will load any managed LaunchAgents as root. But root doesn't have access to the Aqua session and is unable to launch gul applications. Any LaunchAgent can only be loaded once, so the user will never have it loaded for them.

thommay:
@mikedodge04 you'll need to appease `chefstyle` but besides that I think this is a great change, thanks.

mikedodge04:
@thommay passes chef style now.

tas50:
@mikedodge04 Thanks for the lint update, but we're still seeing failures in AppVeyor. The specs are blowing up on Windows when it looks for /dev/console.

tas50:
Also any chance you can squash this down to a single commit so we avoid the merges in history?

lamont-granquist:
yeah unit tests need to get mocked correctly, can't get merged until those are fixed.

  i wonder if the squash + merge strategy on github could handle this PR or not...

PR #5851, opened 2017-03-01
Chef-13:  remove magic from the logger/formatter settings

### LINKED ISSUE #2514: Eliminate formatter and logger distinctions, remove STDOUT magic, reduce user confusion
### Version:

11.16.4
### Environment:

Ubuntu 14.04.1
### Scenario:

Run chef client from command line with an error in the cookbook.
### Steps to Reproduce:

In my case set logging level to :info and log location to /var/log/chef.log in client.rb.  Then have an error in the cookbook that would manifest at compile time.  In my case I tried to notify a service that hadn't been defined.
### Expected Result:

The error should log to the logfile and to the console.
### Actual Result:

Chef client just failed with no error shown in the console.  The log contained the actual error.

End of output on the console:

```
Recipe: mysql::ruby
  * chef_gem[mysql] action install (up to date)  <-- last compile time resource right here
  Converging 310 resources

Running handlers:
Running handlers complete
Chef Client failed. 0 resources updated in 6.368434115 seconds
```

Real error in the log file:

```
[2014-12-02T00:04:22+00:00] ERROR: Running exception handlers
[2014-12-02T00:04:22+00:00] ERROR: Exception handlers complete
[2014-12-02T00:04:22+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2014-12-02T00:04:22+00:00] ERROR: resource template[/etc/php5/mods-available/90-llgy.ini] is configured to notify resource service[php5-fpm] with action restart, but service[php5-fpm] cannot be found in the resource collection. template[/etc/php5/mods-available/90-llgy.ini] is defined in /var/chef/cache/cookbooks/land-web/recipes/default.rb:71:in `from_file'

[2014-12-02T00:04:22+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

lamont-granquist:
This should remove all of the magic.

This makes `chef-client > /tmp/client.out` log with the formatter.

It also makes the default log format the formatter.

It also eliminates the magical addition of a STDOUT formatter when a log formatter is configured in client.rb

closes #2514

There may be a little bit of overrreach here in that someone might be using the magic additional logger to capture logs of all command line invocations of chef-client, which will now be removed.  I tend to feel like that's a better use case for event handlers, the data collector, etc.  Its a very poor-mans solution and i think it had fairly poor features since i think it would switch log formats based on how you invoked it.  At any rate, if people show up who really want that kind of behavior back, I'd say we engineer a knob for them for config.rb which ensures that even on the command line that output goes to a specified log location in addition to STDOUT -- the magical solution was a poor solution.

Note that if people have a `log_location` set in config.rb that now we will read that and when chef-client is invoked on the command line we will send all the output to the log file and the tty will get no output.  IMO, that is simply chef-client doing precisely what you tell it to.  The right way to fix that is to delete that line from config.rb and to pass the `-L` option on the command line to the daemon process or cronjob -- because you want different behavior in those two cases and so you should have to configure it differently.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -270,9 +270,12 @@ The `use_inline_resources` provider mode is always enabled when using the
 `action :name do ... end` syntax. You can remove the `use_inline_resources`
 line.
 
-### Magic removed from logger and formatter settings
+### The logger and formatter settings are more predictable
 
-There is no more magic involved in the setting of the logger and the formatter, the default is now the formatter.
+The default now is the formatter.  There is no more automatic switching to the logger when logging or when output
+is sent to a pipe.  The logger needs to be specifically requested with `--force-logger` or it will not show up.
+
+The `--force-formatter` option does still exist, although it will probably be deprecated in the future.
 
 Setting the `log_location` in the config.rb now applies to both daemonized and command-line invocation and there is
 no magic which dups the logger in order to log to STDOUT on the command line in that case.
@@ -291,4 +294,3 @@ to the command line that invokes your daemon process.
 
 Redirecting output to a file with `chef-client > /tmp/chef.out` now captures the same output as invoking it directly on the command
 line with no redirection.
-
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -269,3 +269,26 @@ use `recipe.rb` in the root of the cookbook.
 The `use_inline_resources` provider mode is always enabled when using the
 `action :name do ... end` syntax. You can remove the `use_inline_resources`
 line.
+
+### Magic removed from logger and formatter settings
+
+There is no more magic involved in the setting of the logger and the formatter, the default is now the formatter.
+
+Setting the `log_location` in the config.rb now applies to both daemonized and command-line invocation and there is
+no magic which dups the logger in order to log to STDOUT on the command line in that case.
+
+Setting logger settings in config.rb is most likely no longer appropriate and those settings should be changed to
+command line flags on the config script or cronjob that invokes chef-client daemonized.  In other words:
+
+```
+chef-client -d -L /var/log/client.rb --force-logger
+```
+
+If your logfiles switch to the formatter, you need to include `--force-logger` for your daemonized runs.
+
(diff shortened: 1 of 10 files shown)


lamont-granquist:
Also note that I kept --force-logger and --force-formatter as the options to use, and have tests that setting both of those results in the formatter, and that preserves prior behavior.  Without the magic that makes a little less sense, but I felt like changing those flags around would be almost entirely for pedantic reasons, and would offer a lot less than the disruption that it would cause by re-engineering them.

lamont-granquist:
Some other thoughts:  I went down the road of considering how to make the behavior different by default based on magic which was not tied to the `STDOUT.tty?` check.  So that we could overload the meaning of `-d` so that it would only obey the `-L` or `log_location` flags when combined with `-d` or that `-d` would imply `--force-logger` by default or something like that.  But all of that is just to try to save typing so that `chef-client -d -L /var/log/chef.log --force-logger` becomes `chef-client -d` magically and means that `chef-client -L /var/log/chef.log` would be broken, and that command is perfectly reasonable to use if you're running from a cronjob and not using the chef-client daemon process. 

I think we need to get back to "chef does exactly what you tell it to" and away from magic.

RELEASE_NOTES.md
```diff
@@ -49,3 +49,25 @@ PolicyFile users on Chef-13 should be using Chef Server 12.3 or higher.
 
 The remediation is removing the self-dependency `depends` line in the metadata.
 
+### Magic removed from logger and formatter settings
+
+There is no more magic involved in the setting of the logger and the formatter, the default is now the formatter.
```
thommay: I feel like "magic" isn't a great word to be putting in front of our users. Maybe:

"Make the logging system more obvious" or "Clarify where log messages will be outputted to"?

lamont-granquist:
after grepping the code this turned out to be not so simple, it got baked into how the windows service stuff works, etc.

lamont-granquist:
@chef/client-core assuming appveyor ever goes green this is ready

PR #5852, opened 2017-03-01
Don't spit out newlines when nothing to puts()!

lamont-granquist:
rebase of the original PR #3778 for chef-13

closes #3778


lamont-granquist:
Not sure I'm happy with this.

`ui.msg "foo"` resulting in "foo\n" while `ui.msg ""` being suppressed and requiring `ui.msg "\n"` to output a single newline feels like the ui.msg ui is being broken to accomplish the goal.  the caller should probably be responsible for
not calling ui.msg when the result of the search or whatever is empty.

lamont-granquist:
paging @jblaine are you around?

the Chef-13 window is now open so we can do this, but i'm not certain hacking up ui.msg is the right approach now and all the `ui.msg "\n"` edits suggest to me this is going down the wrong route.  i'd expect `ui.msg` to behave like `puts` and not have magic empty-string-suppression behavior.

jblaine:
I'm around in that I'm alive. I'm unfortunately in a phase that has me the busiest I've ever been in my life. There's zero chance I can assist further toward this cause any time in the next 6 months.

PR #5853, opened 2017-03-02
support why-run by default

### LINKED ISSUE #5341: Mark all LWRPs and Custom Resources why-run safe by default
Our current policy mostly creates boilerplate code to enable why-run on an LWRP when most LWRPs that are correctly written are why-run safe naturally.

We should flip the default so that the onus is on the authors of bad code to write mark their code as being not why-run safe.

Can't do this until chef-13 though.

lamont-granquist:
this should fix several bugs in providers that look like they should
support whyrun but fail to declare it.

closes #5341

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -1,7 +1,7 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
 # Author:: Christopher Walters (<cw@chef.io>)
-# Copyright:: Copyright 2008-2016, 2009-2016 Chef Software, Inc.
+# Copyright:: Copyright 2008-2016, 2009-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -75,7 +75,7 @@ class Chef
     end
 
     def whyrun_supported?
-      false
+      true
     end
 
     def node
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Thom May (<thom@chef.io>)
-# Copyright:: Copyright (c) 2016 Chef Software, Inc.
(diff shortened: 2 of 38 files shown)


tas50:
This seems incredibly dangerous for a lot of LWRPs out there. While a well written LWRP is why-run safe a lot of the ones that Chef itself wrote are not. I support making custom resources why-run by default, but LWRP convergence was almost never done correctly by end users / employees.

shortdudey123:
Probably need to make sure this gets documented with a link on how to make sure an LWRP is why-run safe

lamont-granquist:
so @tas50 do most people actually use the flag correctly or do they copypasta setting it to true and then write a broken LWRP?

and i don't really see how why-run bugs are any more "incredibly dangerous" than any other bug in a provider.

and there's nothing magical in custom resources that prevents people from writing whyrun bugs -- you write basically the same code -- if you support it for custom resources, you should support this.

@shortdudey123 the TL;DR is to either use sub-resources to do all the work or properly use converge_by blocks around code that change the system state, thats it.  and i believe that those are documented (at least i wrote the documentation for the doc site at least once or twice).

shortdudey123:
Searching https://docs.chef.io for `converge_by` only shows it in code blocks without anything actually saying what it is or what goes in it.  Only `converge_if_changed` is really covered.

I do see it mentioned in https://blog.chef.io/2015/11/06/custom-resources-in-chef-client-12-5/ though

lamont-granquist:
this is where i wrote them:

https://github.com/chef-boneyard/chef-web-docs-2016/pull/647

lamont-granquist:
Also we flipped the default on custom resources end of last year after they'd been around for over a year and we'd been telling everyone to use them:

https://github.com/chef/chef/pull/5339

So far exactly zero people have cared.

coderanger:
It has been on for Poise stuff by default since the 11.x days, though that's probably too small a sample to be useful :)

lamont-granquist:
@chef/client-core reviewage?

PR #5854, opened 2017-03-02
Adding restart action to launchd resource 

mikedodge04:
At first we wanted only support native actions with the launchd resource and launchd does not have the concept of restart. You would unload and load. But not having a restart was causing more confusion to people.


thommay:
I'm in favour of this.

thommay:
@mikedodge04 you need to add the DCO sign off to your commits, please.

PR #5855, opened 2017-03-02
Verify data bag exists before trying to create it

afiune:
### Description
[ZD-12613]
There are users who want to customize permissions to prevent a group of
users from creating new data bags but want to allow the group of users
to create data bag items within an existing data bag.

If the user doesn't have permission to create data bags then the current
API request sequence will prevent the user from creating a data bag item
even if the data bag already exists.

This commit fixes the above problem by verifying (not trying to create)
if the data bag exist through a GET request, if it doesn't the it tries
to generate it, then it proceeds to create the data bag item.

### Issues Resolved

ZD-12613

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/data_bag_create.rb
+++ b/lib/chef/knife/data_bag_create.rb
@@ -49,13 +49,15 @@ class Chef
           exit(1)
         end
 
-        # create the data bag
+        # Verify if the data bag exists
         begin
+          rest.get("data/#{@data_bag_name}")
+          ui.info("Data bag #{@data_bag_name} already exists")
+        rescue Net::HTTPServerException => e
+          raise unless e.to_s =~ /^404/
+          # if it doesn't exists, try to create it
           rest.post("data", { "name" => @data_bag_name })
           ui.info("Created data_bag[#{@data_bag_name}]")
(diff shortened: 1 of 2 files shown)


lib/chef/knife/data_bag_create.rb
```diff
@@ -49,13 +49,15 @@ def run
           exit(1)
         end
 
-        # create the data bag
+        # Verify if the data bag exists
         begin
+          rest.get("data/#{@data_bag_name}")
+          ui.info("Data bag #{@data_bag_name} already exists")
```
rhass: Rather than do exception handling, you may want to just store the response and use something like:

```ruby
case response.code
when 409
   ui.info("Data bag #{@data_bag_name} already exists")
when 404
  rest.post("data", { "name" => @data_bag_name })
  ui.info("Created data_bag[#{@data_bag_name}]")
when >= 500
  raise Net::HTTPServerException
end 
```

rhass: Neverminds -- I just realized the 409 is on a post error. :man_facepalming:

afiune:
ping @chef/maintainers

PR #5856, opened 2017-03-02
Chef-13 remove resource cloning and 3694 warnings

lamont-granquist:
Our long nightmare has ended.

### DIFF
--- a/lib/chef/resource_builder.rb
+++ b/lib/chef/resource_builder.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Lamont Granquist (<lamont@chef.io>)
-# Copyright:: Copyright 2015-2016, Chef Software, Inc.
+# Copyright:: Copyright 2015-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core how about some thumbs?

lib/chef/resource_builder.rb
```diff
@@ -52,14 +52,6 @@ def build(&block)
       resource.source_line = created_at
       resource.declared_type = type
 
-      # If we have a resource like this one, we want to steal its state
-      # This behavior is very counter-intuitive and should be removed.
-      # See CHEF-3694, https://tickets.opscode.com/browse/CHEF-3694
-      # Moved to this location to resolve CHEF-5052, https://tickets.opscode.com/browse/CHEF-5052
-      if prior_resource && Chef::Config[:resource_cloning]
-        resource.load_from(prior_resource)
```
danielsdeleo: Does `Resource#load_from` become dead code now?

danielsdeleo:
I think you might be able to improve your deleted lines score. But 👍 on what's here.

lamont-granquist:
Also I'm going to work on a different PR to add a warn to the resource collection if it finds more than one resource -- which will capture notifies that have ambiguity against the target, or against poorly written accumulators that create multiple resources in the same run context and accumulate incorrectly.

As a concrete example 3694 errors were very useful here:

https://github.com/sous-chefs/samba/issues/56

But every time I've found them to be actually useful its been due to tracking down some shitty resource cloning bug (which will be gone now) or its been due to something searching the resource collection (notifies or accumulator or chef_rewindish kinds of patterns).

I'm going to handle that issue asynchronously tho.

lib/chef/resource_builder.rb
lamont-granquist: maybe?  i don't see it in chefspec or minitest-chef-handler which are the kinds of tools i would have expected to be poking around this deeply in things.

i don't know how i feel about it, because if anyone in the future wants to do that kind of thing, that's the way to to do it -- and if we ever implemented `def dup` we'd probably allocate a new resource and then more or less call that function.

lamont-granquist:
*cough* @coderanger *cough*

```
Poise::Helpers::ResourceCloning
  with a resource that should be cloned
    should receive warn(*(any args)) 1 time (FAILED - 1)
```

coderanger:
lulzzzzzz

coderanger:
@lamont-granquist Should be fixed now.

PR #5858, opened 2017-03-03
HTTP: add debug long for non-JSON response

lamont-granquist:
we should probably raise (?) but this at least helps us debug.

This is just a little bit of help for bugs like this one:  https://github.com/chef/chef-dk/issues/1184

### DIFF
--- a/lib/chef/http/json_output.rb
+++ b/lib/chef/http/json_output.rb
@@ -61,7 +61,9 @@ class Chef
           [http_response, rest_request, return_value]
         else
           Chef::Log.debug("Expected JSON response, but got content-type '#{http_response['content-type']}'")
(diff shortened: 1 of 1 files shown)


lib/chef/http/json_output.rb
```diff
@@ -61,6 +61,7 @@ def handle_response(http_response, rest_request, return_value)
           [http_response, rest_request, return_value]
         else
           Chef::Log.debug("Expected JSON response, but got content-type '#{http_response['content-type']}'")
+          Chef::Log.debug("Response body contains:\n#{http_response.body}")
```
coderanger: Can we do `http_response.body.length < 100 ? http_response.body : "#{http_response.body[0..100]}..."`?

coderanger: (or some other cutoff, just to avoid blowing out scrollback :)

lamont-granquist:
that work for you?  i went a little bit longer in case there was a bit more relevant error spewage...

PR #5863, opened 2017-03-07
CHEF-13:  remove supports API from Chef::Resource

lamont-granquist:
still there on service (where it makes some sense)

also still on mount (because i have no idea if that is actively being
used or if it makes any sense at all).

converts it to a property on mount + service as well.

bonus bug fix:  the old array-style code was mutating the default values, which it should not do now.


lamont-granquist:
well guess i can't do that -- integration tests for the win

lamont-granquist:
@chef/client-core i think this is going to be ready for review -- travis should be green modulo sunspots.

Issue #5542, opened 2016-11-14
Properties have too much magic


thommay:
https://github.com/chef/chef/blob/tm/deprecation_with_url/lib/chef/property.rb#L99-L101 the comment is flat out wrong, but we should make it true.

the logic here: https://github.com/chef/chef/blob/tm/deprecation_with_url/lib/chef/property.rb#L111-L112 should just be - if you specify more than one, we just error.

Issue #3705, opened 2015-07-23
Chef 13 deprecation checklist


coderanger:
This is intended to be a living list of all the things we want to change for the next major rev when we can break compat (though doing so should be carefully considered or Adam will be sad):
- [x] Remove resource cloning.
- [x] Remove `%{file}` from verification command path interpolation.
- [x] Remove old `partial_search` API.
- [x] Remove deprecated setters on `RunContext`.
- [x] Remove `Resource.dsl_name`.
- [x] Remove `Resource.provider_base`.
- [x] Remove `Resource::Execute#path`.
- [x] Remove setting LWRPs as consts.
- [x] Make `Resource::Script#code` required.
- [x] [Re-add freezing of resource property defaults](https://github.com/chef/chef/pull/3744).
- [x] [Make multiple default-y property values be an error](https://github.com/chef/chef/pull/3951).
- [x] Remove `knife cookbook site vendor`

coderanger:
I've added the two I've been poking today, @chef/client-core what else should be on (or removed from) the list?

coderanger:
Added some more from grepping for "will be removed" and similar.

juliandunn:
Can I nominate `knife cookbook site vendor` and `knife cookbook test` commands?

juliandunn:
Also wondering about all the deprecated classes in `lib/chef/deprecation`.

lamont-granquist:
There's some other issues I cut against the Chef 13 milestone, plus there's spec tests that are designed to blow up once someone changes the version to 13.x.x.x.x

thommay:
#5878

coderanger:
Punting on `knife cookbook test` because it isn't deprecated and seems reasonable to keep for the moment, the rest of these are complete.

PR #5866, opened 2017-03-07
Be a bit less keen to help properties

thommay:
This removes all the magic that attempts to allow users to write unsafe
properties - ie, ones that set a default but claim to be a name
property. This yielded different results depending on ordering. It's
better for our users to just suck up fixing this.
Closes: #5542

(Part of #3705)


thommay:
Never mind, this will break old school attributes.

PR #5869, opened 2017-03-07
Remove provisioning from the downstream tests

thommay:
I'm not sure this is providing value. Should we decide to put resources in to sorting out cheffish/provisioning etc, we can easily re-add this.


lamont-granquist:
I've hated the circular dep this creates.

PR #5871, opened 2017-03-07
convert additional resource methods to properties

lamont-granquist:
converts sensitive, retries, retry_delay and ignore_failure to
properties

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -138,9 +138,6 @@ class Chef
       @action = self.class.default_action
       @updated = false
       @updated_by_last_action = false
-      @ignore_failure = false
-      @retries = 0
-      @retry_delay = 2
       @not_if = []
       @only_if = []
       @source_line = nil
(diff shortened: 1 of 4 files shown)


lamont-granquist:
i didn't convert `action`, `allowed_actions`, `guard_interpreter` or `provider` over -- seems like those would be odd as properties, particularly action which isn't itself declarative.  IDK tho.

lamont-granquist:
@chef/client-core should go green in a few minutes and be ready for review (unless we get a Timeout Error)

tas50:
@lamont-granquist  Does this mean a cookbook using those in a custom_resource will actually work on chef 13?

lamont-granquist:
i believe a cookbook that overrides these will still continue to work and won't hit the method redefinition error, yes.  and supports was already handled in the other merged, PR so those won't actually be issues.

tas50:
That makes my life a lot easier. Thanks

PR #5872, opened 2017-03-07
fix the acceptance Gemfile

lamont-granquist:
same fix as the other Gemfile stuff to stop using github and
switch to https everywhere


lamont-granquist:
same fix as #5860

PR #5873, opened 2017-03-07
fix spec for solaris 10

lamont-granquist:
this only affects tests running as root on old solaris because ENV['HOME'] for
root is "/" there and that causes a "/.ssh" vs "//.ssh" error if we don't
File.expand_path in both the code and the spec.


lamont-granquist:
@chef/client-core not standing on ceremony for this one

Issue #5819, opened 2017-02-16
We should update the default data bag encryption version for Chef 13


coderanger:
Adding this so we don't forget. Unless someone has a really good reason we should bump the encryption version to 3 (it is currently 1) for Chef 13 because it's been long enough.

thommay:
Annoyingly our docs say 0,1,2 and the code says 1,2,3; so we should fix that too.
https://docs.chef.io/data_bags.html#encryption-versions

lamont-granquist:
should also make sure that mixlib-auth is using 1.3 by default, and consider dropping support for 1.0 and 1.1 (although probably not 1.2 since 1.3 only came out end of last year).

i don't have the history of that in my brain so don't know how disruptive any of that would be.

PR #5877, opened 2017-03-08
Use v3 data bag encryption

thommay:
Closes: #5819

cc @chef/maintainers

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -222,7 +222,7 @@ GEM
       logify (~> 0.1)
       mime-types
     chef-sugar (3.4.0)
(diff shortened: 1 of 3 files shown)


lamont-granquist:
might have actually broken a chef-zero test.

danielsdeleo:
FWIW, the original problem was lack of availability of the encryption modes we wanted to use in older versions of OpenSSL. There was some question about non-omnibus installs technically being supported that was holding us back. Should be a-ok now though.

👍 

Would be great to fix the docs also, as mentioned in the comment on original ticket: https://github.com/chef/chef/issues/5819#issuecomment-280393780

PR #5880, opened 2017-03-08
Remove all 11 era deprecations

thommay:


lamont-granquist:
gawd that's like chef 11.6.0 era deprecations that we never cleaned up in chef-12....

PR #5882, opened 2017-03-08
Chef-13:  Chef::Resource cleanup

lamont-granquist:
most of this deletes useless old code.

the change to lookup_provider_constant changes to more strictly
stop using class name based lookups and to go through the resource
resolver and provider resolver.

as a result in order to find the provider class for a given dsl name
you have to go through the resource resolver to find the resource in
order to be able to pass a resource instance through the provider
resolver.  since the provider resolver api passes resources into blocks
passed into provides api you must construct a resource instance.

that means that providers need to be associated with resources in order
to be looked up (which makes sense in Real Life(tm) use of Chef, but
breaks quite a few lazy tests we had where we constructed providers
without doing the work of wrapping them in a resource.

note that as the deploy resource shows this filters into a changed
behavior of the `provider` syntax where before `provider :revision`
would look up Chef::Provider::Deploy::Revision via class-name based
magic.  this breaks that API so that `provider :deploy_revision` is
used instead -- the symbol (or string) there is turned into a resource
first via the Chef::ResourceResolver and then looked up via the
Chef::ProviderResolver into Chef::Provider::Deploy::Revision.  this
is a breaking change but is also a bug fix so that the symbol here
goes through the same lookup that you get when you type it in the DSL.

i had considered implementing a lookup from a resource_name symbol to
a provider, but in looking at how to implement that in the
ProviderResolver the issue is that we really need to have a resource
instance order to pass to the ProviderResolver.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/chef/chef-server
-  revision: 5cdb74dc17a96cac30a6a4dc1381d3d3c7b2fb4f
+  revision: 4ec1d014c9a4945398c6baa550c71f7e3a7e69e0
   specs:
     oc-chef-pedant (2.2.0)
       activesupport (>= 4.2.7.1, < 6.0)
@@ -97,7 +97,7 @@ GIT
 
 GIT
   remote: https://github.com/poise/poise.git
-  revision: 62056469b3a19144f4bf4810abe14efa1c10463b
+  revision: 1aa3a6b7c5b3fe0c357e2f3fcc50080790772ccb
   specs:
     poise (2.7.3.pre)
       halite (~> 1.0)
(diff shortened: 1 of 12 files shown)


lamont-granquist:
@coderanger looks like this broke a fused resources test that uses the symbol version of `provider` lookup.

how 'should' this API behave in a dynamic provider-resolver world?

i was going to make it lookup based on matching the resource_name to the provides syntax but the way you have to construct a valid resource instance to send into the block-form of the provides syntax (on a provider) derailed that.

we could avoid calling those, but that feels wrong as well -- the block may not depend on the resource argument at all and bypassing it may get the behavior totally wrong.

in order to construct a resource instance to pass to it, we really should be going through the resource resolver, and to do that we should be using that symbol which is provided -- which is how this code all ended up where it did.

of course "burn the explicit provider syntax in fire" is a direction that i'd ideally like to head towards...

coderanger:
To summarize from Slack, 👍 from me on a dropping strings/symbols from the `Chef::Resource#provider` API since they kind of no longer make sense with how we map providers to resources. The broken test has been changed to pass in a class object directly.

lamont-granquist:
@chef/client-core this should be reviewable and should go green

lamont-granquist:
Also to summarize from slack, we still haven't dropped symbols/strings from Chef::Resource#provider just changed them to go through the ResourceResolver + ProviderResolver and in the process that breaks the ability to lookup a "bare" provider (without any associated resource we can lookup in the ResourceResolver) this way.  This fix to poise would have also worked:  

https://github.com/poise/poise/pull/30/commits/bbfa8cdaad3c126a05df9ea66dff675086dc1449

However, might choose to remove symbols and strings to Chef::Resource#provider at some point, but its feeling more Chef-14-ish right now.  There's definitely still sharp edges there where its better not to be using that API.

PR #5884, opened 2017-03-09
Compress debs and rpms with xz

thommay:
This gives considerable size benefits, at a cost of slightly slower decompression. xz compression has been available in dpkg and rpm since 2009 so we shouldn't see any platforms that don't support this.

### DIFF
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -53,6 +53,16 @@ dependency "chef-complete"
 
 package :rpm do
   signing_passphrase ENV["OMNIBUS_RPM_SIGNING_PASSPHRASE"]
+
+  unless rhel? && platform_version.satisfies?("< 6")
+    compression_level 1
+    compression_type :xz
+  end
+end
+
+package :deb do
+  compression_level 1
+  compression_type :xz
 end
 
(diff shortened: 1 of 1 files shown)


thommay:
For them what can see manhattan, http://manhattan.ci.chef.co/job/chef-trigger-ad_hoc/11/

thommay:
For debian packages, this drops us from 32mb to 27mb

thommay:
On EL6 and EL7 the difference is about the same, however we'll need to disable this on EL5 since apparently I was premature in my "won't see any platforms that don't support this" statement.

lamont-granquist:
came here to point out el5 was like 2007...  =)

tas50:
April...

srenatus:
Just a note -- higher levels give _drastically smaller_ artifacts, at the cost of _compression time_. Depending on our workflow, it might be worth considering to use higher levels for release builds, or something? (I haven't yet figured out where _repackaging_ would fit into your pipelines...)

srenatus:
Update: I've started the discussion here for chef-server: https://github.com/chef/chef-server/pull/1150 -- I'd think the same thing would work for chef (and chef-dk, for that matter).

Issue #5810, opened 2017-02-10
Port windows_task  resource from windows cookbook to core chef


tas50:
## Description

Scheduled task setup is one of the more common tasks on a Windows systems and doing so in cookbooks is incredibly common. We currently ship a windows_task resource in the Windows cookbook, which requires depending on that cookbook. Additionally the resource has multiple issues and is in need of a refactor:

https://github.com/chef-cookbooks/windows/issues/407
https://github.com/chef-cookbooks/windows/issues/245
https://github.com/chef-cookbooks/windows/issues/176
https://github.com/chef-cookbooks/windows/issues/353
https://github.com/chef-cookbooks/windows/issues/361
https://github.com/chef-cookbooks/windows/issues/366

There is a gem that may potentially help us reduce the complexity of this resource and resolve some of the issues listed above:

https://github.com/djberg96/win32-taskscheduler

tas50:
This was fixed by #5886. Thanks @NimishaS

PR #5886, opened 2017-03-09
Adding windows_task resource

NimishaS:
### Description

Adding support for `windows_task` resource

### Issues Resolved

https://github.com/chef/chef/issues/5810
https://github.com/chef-cookbooks/windows/issues/407
https://github.com/chef-cookbooks/windows/issues/245
https://github.com/chef-cookbooks/windows/issues/176
https://github.com/chef-cookbooks/windows/issues/353
https://github.com/chef-cookbooks/windows/issues/361
https://github.com/chef-cookbooks/windows/issues/366

### Check List

- [x] All tests pass


lib/chef/resource/windows_task.rb
```diff
@@ -0,0 +1,65 @@
...
+                                       :once,
+                                       :on_logon,
+                                       :onstart,
+                                       :on_idle], default: :hourly
+      property :start_day, String
+      property :start_time, String
+      property :day, [String, Integer]
+      property :months, String
+      property :idle_time, Integer
+      property :random_delay, String
```
tas50: Should we validate the random delay the user passes us here?

NimishaS: @tas50 I thought of that. But then the accepted values for `random_delay` are like `PT5M`, `PT3M`, `PT1H`, `PT24H`, `P1D` etc. The list is infinite. I was thinking of updating the Readme and other docs to help the user understand what values can be given here. 
Right now the recipe throws error: `The task XML contains a value which is incorrectly formatted or out of range.` if an invalid value is given.

mwrock: One option would be to accept an integer as a number of seconds. Then use something like https://github.com/WinRb/WinRM/blob/master/lib/winrm/wsmv/iso8601_duration.rb to convert it to the ISO8601 format.

Tech356:
Is the plan for this implementation to use `schtasks.exe` or will it migrate to the win32 api? Will triggers be supported (chef-cookbooks/windows#258)?

lib/chef/resource/windows_task.rb
NimishaS: @mwrock, when we accept input from the user in a different format than what is expected by the task, more parsing is required to ensure idempotency during `:create` action. It's doable but requires time.
I was working on https://github.com/chef-cookbooks/windows/issues/366. It's tricky to fix this because we are setting the properties of `@current_resource` using method `load_task_hash` to ensure idempotency: https://github.com/chef/chef/pull/5886/files#diff-208e6e972fb102d2fce9f211de0d94a8R262.
This method uses `schtasks /Query /FO LIST /V /TN <task_name>` to get the details of the task and doesn't return all the required properties. For some properties we need to inspect the XML object e.g.  `random_delay`. A lot of parsing is also required since the values that we get from the command and from the XML are not in the same format that the resource expects.
e.g. the resource expects `frequency` as let's say 15 and `frequency_modifier` as `minutes`, while the task XML will return `PT15M`.
A quick fix will be to update the task everytime `:create` action is called, but then we'll be compromising on the idempotency.

NimishaS: But I like this: https://github.com/WinRb/WinRM/blob/master/lib/winrm/wsmv/iso8601_duration.rb. We'll just require it's reverse too for maintaining idempotency.

mwrock: I was thinking that with the above method, you can convert a user input integer to the string iso format and then compare that to the actual value in the XML. Asking users to enter the delay in the ISO 8601 may be more error prone than a simple number of seconds. Also, I'd consider switching to XML in the `load_task_hash`. That way you consistently deal with the same data without the mismatch between the xml and `schtasks /Query`.

NimishaS: @mwrock , made the changes to accept `random_delay` in seconds.

lib/chef/provider/windows_task.rb
```diff
@@ -0,0 +1,396 @@
...
+        ensure
+          ::File.delete(temp_task_file)
+        end
+      end
+
+      def load_task_hash(task_name)
+        Chef::Log.debug 'Looking for existing tasks'
+
+        # we use shell_out here instead of shell_out! because a failure implies that the task does not exist
+        output = shell_out("schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout
```
jugatsu: If Windows locale is different than English, for example, Russian, the output will be in Russian language so the hash will be completely useless. Therefore, resource will be non-idempotent by design.
As a workaround we can use chcp 437 prior to run schtasks.exe
```cmd
output = shell_out("chcp 437 && schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout
```

lib/chef/provider/windows_task.rb
```diff
@@ -0,0 +1,396 @@
...
+        Chef::Log.debug('running: ')
+        Chef::Log.debug("    #{cmd}")
+        shell_out!(cmd, returns: [0])
+      end
+      # rubocop:enable Style/StringLiteralsInInterpolation
+
+      def task_need_update?
+        # gsub needed as schtasks converts single quotes to double quotes on creation
+        @current_resource.command != @new_resource.command.tr("'", '"') ||
+        @current_resource.user != @new_resource.user ||
```
jugatsu: If we run `windows_task` on non-English system, for example, Russian version of Windows, the task will be created with RunAsUser: `СИСТЕМА`. As a result task will be created on each run, as a
```ruby
@current_resource.user != @new_resource.user
```
will never match. I think, it's because encoding issues. Maybe we should use SID, instead of hardcoded system names.

NimishaS: @jugatsu in case the system is Russian, do you pass `user` in Russian in the recipe or in English?

jugatsu: If i pass `user` in English it works as expected (with chcp 437 it's also idempotent, see above my comment).
If i pass `user` in Russian, task is created on each chef-client run.

NimishaS: I haven't tried writing a recipe on a non-English server. But suppose I run the following recipe on a Russian server:
```
windows_task 'chef-client' do
  user 'СИСТЕМА'
  password 'password'
  command 'chef-client'
  run_level :highest
  frequency :hourly
  frequency_modifier 5
  random_delay '2400'
end
```
The user will be save as `СИСТЕМА` in the system for scheduled task. And if I run the same recipe again, the scheduled task XML will return me user in Russian only and hence the idempotency will work.

Please correct me if I am missing something

jugatsu: @NimishaS 
`СИСТЕМА` translates to english `SYSTEM`, it's a system user. You can't pass it without password because system username hardcoded in
```ruby
SYSTEM_USERS = ['NT AUTHORITY\SYSTEM', 'SYSTEM', 'NT AUTHORITY\LOCALSERVICE', 'NT AUTHORITY\NETWORKSERVICE'].freeze
```

Btw, I don't need to pass translated system user, because with defaut SYSTEM user task `will be` created without any problems.
**BUT** 
```cmd
schtasks /Query /FO LIST /V /TN \"#{task_name}\
```
will return `СИСТЕМА` and task will not be idempotent.

Also if i create user `Пользователь` (translates `User`) and use
```ruby
windows_task 'chef-client' do
  user 'Пользователь'
  password 'password'
  command 'chef-client'
  run_level :highest
  frequency :hourly
  frequency_modifier 5
  random_delay '2400'
``
task will be created on each converge.

Comment (jugatsu) [inline lib/chef/provider/windows_task.rb]: @NimishaS Thanks for adding chcp for fixing shell_out ountput.
The last thing to make resource completely idempotent is to fix issue with
```ruby
@current_resource.user != @new_resource.user
```
when `user` is non-English

NimishaS: @jugatsu, if I remove this line then then how will idempotency check be done when the user is being passed in the resource? eg.
```
windows_task 'chef-client' do
  user 'Nimisha'
  password 'password'
  command 'chef-client'
  run_level :highest
  frequency :hourly
  frequency_modifier 5
end
```

Can you please share the exact recipe which I can use to reproduce the issue that you are pointing at? You can probably checkout my branch and try the recipe.

jugatsu: @NimishaS you shouldn't remove this line because, as you already said, without it we can't check for idempotency for `user` property.
```ruby
load_current_value do |desired|
  pathed_task_name = desired.task_name.start_with?('\\') ? desired.task_name : "\\#{desired.task_name}"

  task_hash = load_task_hash pathed_task_name

  task_name pathed_task_name
  if task_hash.respond_to?(:[]) && task_hash[:TaskName] == pathed_task_name
    exists true
    status :running if task_hash[:Status] == 'Running'
    enabled task_hash[:ScheduledTaskState] == 'Enabled' ? true : false
    cwd task_hash[:StartIn] unless task_hash[:StartIn] == 'N/A'
    command task_hash[:TaskToRun]
    user task_hash[:RunAsUser]
  else
    exists false
  end
end
```

Here we load current `user task_hash[:RunAsUser]` value based on output of 
```ruby
output = shell_out("schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout
```

But the problem is that this output on russian systems will return `?????` instead of actual user, for example, `СИСТЕМА`. 
I think this is `shell_out` issue with codepage 866.

If I hardcoded `user` in `load_current_value do |desired|`
```ruby
load_current_value do |desired|
  pathed_task_name = desired.task_name.start_with?('\\') ? desired.task_name : "\\#{desired.task_name}"

  task_hash = load_task_hash pathed_task_name

  task_name pathed_task_name
  if task_hash.respond_to?(:[]) && task_hash[:TaskName] == pathed_task_name
    exists true
    status :running if task_hash[:Status] == 'Running'
    enabled task_hash[:ScheduledTaskState] == 'Enabled' ? true : false
    cwd task_hash[:StartIn] unless task_hash[:StartIn] == 'N/A'
    command task_hash[:TaskToRun]
    # user task_hash[:RunAsUser]
    user 'пользователь'
  else
    exists false
  end
end
```
IT WORKS AS EXPECTED, so the resource is idempotent.

Here is actual recipe
```ruby
windows_task 'task_from_name' do
  command 'dir'
  # user 'vagrant'
  # password 'vagrant'
  user 'пользователь'
  password 'Ohn9phev'
  run_level :highest
  action :create
end
```

jugatsu: Here is debug output
```ruby
[2017-03-23T10:38:58-07:00] DEBUG: Looking for existing tasks
Active code page: 437

Folder: \
HostName:                             VAGRANT-2016
TaskName:                             \task_from_name
Next Run Time:                        23.03.2017 11:38:00
Status:                               Ready
Logon Mode:                           Interactive/Background
Last Run Time:                        30.11.1999 0:00:00
Last Result:                          267011
Author:                               VAGRANT-2016\vagrant
Task To Run:                          dir
Start In:                             N/A
Comment:                              N/A
Scheduled Task State:                 Enabled
Idle Time:                            Disabled
Power Management:                     Stop On Battery Mode, No Start On Batteries
Run As User:                          ?????
Delete Task If Not Rescheduled:       Disabled
Stop Task If Runs X Hours and X Mins: 72:00:00
Schedule:                             Scheduling data is not available in this format.
Schedule Type:                        One Time Only, Hourly
Start Time:                           10:38:00
Start Date:                           23.03.2017
End Date:                             N/A
Days:                                 N/A
Months:                               N/A
Repeat: Every:                        1 Hour(s), 0 Minute(s)
Repeat: Until: Time:                  None
Repeat: Until: Duration:              Disabled
Repeat: Stop If Still Running:        Disabled
[2017-03-23T10:38:58-07:00] DEBUG: new_resource user: пользователь
[2017-03-23T10:38:58-07:00] DEBUG: current_resource user: ?????
[2017-03-23T10:38:58-07:00] DEBUG: new_resource commad: dir
[2017-03-23T10:38:58-07:00] DEBUG: current_resource command: dir
[2017-03-23T10:38:58-07:00] DEBUG: new_resource user: пользователь
[2017-03-23T10:38:58-07:00] DEBUG: current_resource user: ?????
[2017-03-23T10:38:58-07:00] DEBUG: new_resource commad: dir
[2017-03-23T10:38:58-07:00] DEBUG: current_resource command: dir
[2017-03-23T10:38:58-07:00] DEBUG: running:
[2017-03-23T10:38:58-07:00] DEBUG:     schtasks /CREATE /TN "task_from_name" /F /SC "hourly" /MO "1" /TR "dir" /RU "пользователь" /RP "Ohn9phev" /RL "HIGHEST"
WARNING: The task is registered, but may fail to start. Batch logon privilege needs to be enabled for the task principal.

    - creating a new scheduled task task_from_name
```

For debug output I hacked `def task_need_update?` to
```ruby
def task_need_update?
    # gsub needed as schtasks converts single quotes to double quotes on creation
    Chef::Log.debug "new_resource user: #{new_resource.user}"
    Chef::Log.debug "current_resource user: #{current_resource.user}"
    Chef::Log.debug "new_resource commad: #{new_resource.command}"
    Chef::Log.debug "current_resource command: #{current_resource.command}"
    current_resource.command != new_resource.command.tr("'", '"') ||
      current_resource.user.encode('external') != new_resource.user
  end
```

jugatsu: Working example
```ruby
[2017-03-23T10:34:15-07:00] DEBUG: Looking for existing tasks
Active code page: 437

Folder: \
HostName:                             VAGRANT-2016
TaskName:                             \task_from_name
Next Run Time:                        23.03.2017 11:32:00
Status:                               Ready
Logon Mode:                           Interactive/Background
Last Run Time:                        30.11.1999 0:00:00
Last Result:                          267011
Author:                               VAGRANT-2016\vagrant
Task To Run:                          dir
Start In:                             N/A
Comment:                              N/A
Scheduled Task State:                 Enabled
Idle Time:                            Disabled
Power Management:                     Stop On Battery Mode, No Start On Batteries
Run As User:                          vagrant
Delete Task If Not Rescheduled:       Disabled
Stop Task If Runs X Hours and X Mins: 72:00:00
Schedule:                             Scheduling data is not available in this format.
Schedule Type:                        One Time Only, Hourly
Start Time:                           10:32:00
Start Date:                           23.03.2017
End Date:                             N/A
Days:                                 N/A
Months:                               N/A
Repeat: Every:                        1 Hour(s), 0 Minute(s)
Repeat: Until: Time:                  None
Repeat: Until: Duration:              Disabled
Repeat: Stop If Still Running:        Disabled
[2017-03-23T10:34:15-07:00] DEBUG: new_resource user: vagrant
[2017-03-23T10:34:15-07:00] DEBUG: current_resource user: vagrant
[2017-03-23T10:34:15-07:00] DEBUG: new_resource commad: dir
[2017-03-23T10:34:15-07:00] DEBUG: current_resource command: dir
[2017-03-23T10:34:15-07:00] INFO: windows_task[task_from_name] task already exists - nothing to do
 (up to date)
```

lib/chef/resource/windows_task.rb
jugatsu: @mwrock If we switch to XML in the `load_task_hash` how we actualy will be checking  `user` property for idepmotency as XML output doesn't contain any usernames, only SID?

lib/chef/provider/windows_task.rb
```diff
@@ -0,0 +1,398 @@
...
+        ensure
+          ::File.delete(temp_task_file)
+        end
+      end
+
+      def load_task_hash(task_name)
+        Chef::Log.debug 'Looking for existing tasks'
+
+        # we use shell_out here instead of shell_out! because a failure implies that the task does not exist
+        output = shell_out("chcp 437 && schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout
```
jugatsu:
```cmd
output = shell_out("chcp 437 >nul 2>&1 && schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout
```
Pass output of `chcp` to null for supress `Active code page: 437` in debug output.

jugatsu: @NimishaS I make resource completely idempotent with this patch
```ruby
output = shell_out("chcp 65001 >nul 2>&1 && schtasks /Query /FO LIST /V /TN \"#{task_name}\"").stdout.force_encoding('UTF-8')
```
set UTF prior to run `schtasks` and the encode output to `UTF-8`

What do you think about this?

lib/chef/resource/windows_task.rb
NimishaS: @jugatsu , I am using both XML output and `schtasks /Query` command to check for idempotency. For `user` property, `schtasks /Query` command is being used.

btm:
@Tech356 the immediate effort is to obtain feature parity in core with the windows cookbook. This is one of the last, if not the last, resource we plan to migrate from the windows cookbook into core. We're trying to fix the bugs filed against the resource as well, but not trying to add features to it. The goal is to get a stable resource into Chef 13 to facilitate deprecating the resource in the windows cookbook.

lib/chef/resource/windows_task.rb
```diff
@@ -0,0 +1,237 @@
...
+        if [:once, :on_logon, :onstart, :on_idle].include? frequency
+          raise ArgumentError, "`start_day` property is not supported with frequency: #{frequency}"
+        end
+      end
+
+      def validate_start_time(start_time)
+        raise ArgumentError,"`start_time` needs to be provided with `frequency :once`" unless start_time
+      end
+
+      SYSTEM_USERS = ['NT AUTHORITY\SYSTEM', 'SYSTEM', 'NT AUTHORITY\LOCALSERVICE', 'NT AUTHORITY\NETWORKSERVICE'].freeze
```
smurawski: We should add `BUILTIN\USERS` and `USERS` per https://github.com/chef-cookbooks/windows/pull/457 as this allows users to schedule tasks that can run for the currently logged on user.  Helpful when managing client systems or embedded devices.

lib/chef/provider/windows_task.rb
```diff
@@ -213,8 +213,11 @@ def update_task_xml(options = [])
 
         Chef::Log.debug "looking for existing tasks"
 
-        # we use shell_out here instead of shell_out! because a failure implies that the task does not exist
-        xml_cmd = shell_out("schtasks /Query /TN \"#{@new_resource.task_name}\" /XML")
+        task_script = <<-EOH
+            [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+            schtasks /Query /TN \"#{@new_resource.task_name}\" /XML
+        EOH
+        xml_cmd = powershell_out("#{task_script}")
```
jugatsu: @NimishaS to make rubocop happy you can just write
```ruby
xml_cmd = powershell_out(task_script)
```

NimishaS: Sorry that's so obvious. I should have done it that way only. I'll fix it.

mwrock:
This looks good to me @NimishaS . I'd really like to see some functional tests here. In the windows cookbook we had kitchen tests that created tasks and tested that the tasks existed and ran as well as some other functionality. Having equivilent functional tests would be great.

NimishaS:
@mwrock , we had skipped the functional testcases for time being as this functionality needs to be merged for Chef 13. We had a discussion about this with @btm.
We have plans to add functional testcases later. Adding functional testcases requires 2-3 more days. Adding testcases for test-kitchen will require some more time. If that's ok then I can start work on that today.

thommay:
@NimishaS are you happy for us to release this without functional tests? If so, we need to get this merged today for the 13.0 release candidate, before the release on monday.

NimishaS:
@thommay , I have done a thorough testing for `windows_task` resource: https://gist.github.com/NimishaS/04cbc6937e32d74dc4bdbd2f72526e6a.

It's working successfully. This PR is good to merge. We will add the functional specs soon.

.bundle/config
```diff
@@ -1,2 +0,0 @@
----
-BUNDLE_FROZEN: "1"
```
thommay: Please remove this change, and use `bundle exec rake dependencies:update_gemfile_lock` to ensure the Gemfile.lock is correct.

lib/chef/provider/windows_task.rb
```diff
@@ -0,0 +1,419 @@
...
+
+require "chef/mixin/shell_out"
+require "rexml/document"
+require "iso8601"
+require "chef/mixin/powershell_out"
+
+class Chef
+  class Provider
+    class WindowsTask < Chef::Provider
+      use_inline_resources
```
thommay: `use_inline_resources` no longer does anything and can be removed.

lib/chef/resource/windows_task.rb
```diff
@@ -0,0 +1,237 @@
...
+                                       :on_logon,
+                                       :onstart,
+                                       :on_idle], default: :hourly
+      property :start_day, String
+      property :start_time, String
+      property :day, [String, Integer]
+      property :months, String
+      property :idle_time, Integer
+      property :random_delay, String
+      property :execution_time_limit
```
thommay: does this need a type?

.bundle/config
NimishaS: This is a removed file. It got added in this PR as I cherry-picked https://github.com/chef/chef/pull/5886/commits/2ff65b4f07617dbb1255b7b8f4795d69a35dcb39

@thommay , does it still affect the PR even though it's showing as a removed file?

NimishaS: Wait.. I got your point.

NimishaS:
Done @thommay

PR #5890, opened 2017-03-09
Jeremymv2/cheffs optimizations

jeremymv2:
### Description

During the course of a backup/restore operation, initially two `GET` requests are made: the first for establishing whether the objects exists, the second for comparing the object for determining whether to copy it or not.  Both requests result in the entire object being retrieved.  This is very inefficient when dealing with thousands of objects.

This change eliminates one of those `GET` requests by caching the first and using it for later comparison.
It has been tested and validated for both backup and restore operations.

### Issues Resolved

n/a

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb
+++ b/lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb
@@ -165,6 +165,9 @@ class Chef
           end
 
           def write(file_contents)
+            # free up cache - it will be hydrated on next check for exists?
+            @this_object_cache = nil
+
             begin
               object = Chef::JSONCompat.parse(file_contents)
             rescue Chef::Exceptions::JSON::ParseError => e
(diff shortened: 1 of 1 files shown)


echohack:
👍 Faster backups for large organizations are awesome! Let's do this!

lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb
```diff
@@ -85,6 +86,8 @@ def exists?
           end
 
           def delete(recurse)
+            # free up cache - it will be hydrated on next check for exists?
+            @this_object_cache = nil
```
stevendanna: Do we want something similar in `write`?

jeremymv2: Hi @stevendanna - I've added the same to `write`.  Could you re-review please?

PR #5892, opened 2017-03-10
add missing functional tests for users if applicable Do not merge

harikesh-kolekar:

### Description In progress Do not merge 

https://github.com/chef/chef/blob/master/spec/functional/win32/security_spec.rb#L33

I'm fairly certain that the mixlib-shellout user code works now, we were just talking about using it with powershell_out. If so, we create the necessary functional tests here.

### Check List

- [ ] All tests pass


spec/functional/win32/security_spec.rb
```diff
@@ -31,14 +34,36 @@
   # support for windows.
   #
   # TODO - Add negative tests once mixlib-shellout has user support
-  it "has_admin_privileges? returns false when running as non-admin" do
-    skip "requires user support in mixlib-shellout"
+  describe "running as non admin user" do
+    let(:user) { "security_user" }
+    let(:password) { "security" }
+    before do
+      include Chef::Mixin::WideString
```
btm: we should definitely add a comment that this needs to be replaced when this code ends up in a mixin and refers to #5832.

btm: you might just want to pull the mixin from #5832 in here, and then you can use the other tests there as a guide.

`with_user_context` yields a block, so you could pass a block with `Chef::ReservedNames::Win32::Security.has_admin_privileges?` and then check the return of the block for true/false.

PR #5895, opened 2017-03-13
Chef-13: remove method_missing access to node object.

lamont-granquist:
So again the reasons why we break this:

- node.class vs. node['class'] kinds of problems
- adding node#foo to Chef::Node as an extension to the API and
  breaking node.foo meaning node["foo"] (we can't make exensions
  to that class without possibly breaking someone)
- crazy things like the old CHEF-3799 issue in the old ticketing
  system where IO#puts in ruby blindly calls #to_ary on stuff and
  expects it to raise -- whereas we would potentially autovivify
  a 'to_ary' hash key and return nil which breaks the world.

This also has caused issues with the hashie gem and they've gone to
spamming warnings by default to try to deal with it:

https://github.com/berkshelf/berkshelf/issues/1665

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,3 +53,9 @@ The remediation is removing the self-dependency `depends` line in the metadata.
 
 Retained only for the service resource (where it makes some sense) and for the mount resource.
 
+### Removed deprecated `method_missing` access from the Chef::Node object
+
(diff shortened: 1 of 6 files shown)


lamont-granquist:
@chef/client-core travis willing this should be ready to review

coderanger:
yjliv51

stevendanna:
gif-keyboard-17097229626904590709

PR #5896, opened 2017-03-14
Chef-13:  properly deep dup Node#to_hash

lamont-granquist:
previously to_hash allowed mutating non-container elements, now
they get properly dup'd.

fixes a 2.5 year old pending spec test.

also fills out the API so that there is to_h/to_hash/to_a/to_array
instead of the weird mix-and-match we had before.

bonus-fix:  dup wasn't properly dup'ing as well

### DIFF
--- a/lib/chef/node/immutable_collections.rb
+++ b/lib/chef/node/immutable_collections.rb
@@ -70,19 +70,16 @@ class Chef
       end
 
       def to_a
-        a = Array.new
-        each do |v|
-          a <<
-            case v
-            when ImmutableArray
-              v.to_a
-            when ImmutableMash
-              v.to_h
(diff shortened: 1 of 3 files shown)


lamont-granquist:
@chef/client-core seems i forgot to ping last night for reviewers -- this is ready

lamont-granquist:
@client-core pingles?

lamont-granquist:
Note that this behavior in Chef-12 has caused at least one "massive outage" in a large customer.

Anyone who gets bit by incompatibility here should know that its very unlikely for us to reconsider this choice and it was not done arbitrarily.    Similarly with the freezing of the merged attributes in #5905.

PR #5900, opened 2017-03-14
Chef-13:  Remove declare_resource create_if_missing API

lamont-granquist:
Likely nobody cares about this change, I added it and then a
release or two later introduced the better edit_resource style
of API.

I also changed "created_at" to be a proper named parameter, instead
of a positional parameter, which is a breaking change in an API that
has been around for quite some time -- makes the API consistent since
I added it as a named parameter to the new APIs when I wrote them.
(This one is very internal, though, so I suspect there's a shortlist of
things that muck about in chef internals that would be all that would break).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -59,3 +59,9 @@ Previously, the syntax `node.foo.bar` could be used to mean `node["foo"]["bar"]`
 with the core ruby Object class (e.g. `node.class`) and where it collided with our own ability to extend the `Chef::Node` API.  This
 method access has been deprecated for some time, and has been removed in Chef-13.
 
+### Changed `declare_resource` API
+
+Dropped the `create_if_missing` parameter that was immediately supplanted by the `edit_resource` API (most likely nobody ever used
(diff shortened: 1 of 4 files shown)


lamont-granquist:
I'm giving roughly even odds that poise and/or chefspec goes red...

coderanger:
Heh, if it doesn't go red on poise it definitely will on the application cookbook. I'll have to ponder how to deal with positional vs named param, I wish Ruby didn't make that so hard :-(

lamont-granquist:
I don't suuuuuper care about fixing that latter one, it can get backed out.

lamont-granquist:
You can always call it one way or the other and `rescue ArgumentError` it blows up with one either if you call it with the extra positional parameter, or with the named parameter it doesn't know about.

coderanger:
That would break if we ever add another positional param in the future though, I'll probably just do an explicit version check and hold my nose :)

lamont-granquist:
call it with the named parameter -- it validates that as well and will ArgumentError on Chef 12.

lamont-granquist:
crazy it went green -- @coderanger could you throw an approval on this one if you've sorted out the application cookbook stuff?

PR #5902, opened 2017-03-14
bump ruby to 2.3.3

lamont-granquist:
can finally do this now that the branching nightmare is gone.

not bumping to 2.4.0 because of:

- perf regression in rspec and waiting for 2.4.1
- not wanting to buy that yak shave yet
- iterative development and moving the needle forwards in
  small increments


coderanger:
We should make sure to circle back on 2.4 before shipping 13 though, if we don't want to run into problems similar to 2.1 during the 13.x lifecycle >_< 👍

lamont-granquist:
Well, I'd considered 2.4.0 to be blocked waiting on 2.4.1, and if its not released by the time that 13.0 is released, we've got RFC cover to bump to 2.4.1 in 13.1

PR #5905, opened 2017-03-15
Chef-13: freeze merged node attribute

lamont-granquist:
Another oversight in creating ImmutableMash/Array but the strings were never dup'd and frozen.

I'm certain this will break bad code out there.  There's no way to issue deprecation warnings though (monkeypatching every mutator on every simple class like String and Fixnum so that you can mark a string as being deprecated-frozen so that they'll issue warnings, although dup will remove the flag and make it behave the same afterwards?  Nope...)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -82,3 +82,10 @@ n["foo"] << "buzz"
 before this would have mutated the original string in-place so that `node["foo"]` and `node.default["foo"]` would have changed to "fizzbuzz"
 while now they remain "fizz" and only the mutable `n["foo"]` copy is changed to "fizzbuzz".
 
+### Freezing immutable merged attributes
+
+Since Chef 11 merged node attributes have been intended to be immutable but the merged strings have not been frozen.  In Chef 13, in the
+process of merging the node attributes strings and other simple objects are dup'd and frozen.  In order to get a mutable copy, you can
+now correctly use the `node.dup` or `node.to_hash` methods, or you should mutate the object correctly through its precedence level like
+`node.default["some_string"] << "appending_this"`.
+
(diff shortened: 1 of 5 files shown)


lamont-granquist:
And the reason for doing this is that in complicated large environments allowing this can lead to bugs which are very difficult to debug since we have a rule that in order to mutate attributes you must go in through the precedence accessor but there exist ways to exploit the existing behavior and bypass that rule.  Every place this breaks will likely require a few minutes of fixing, but may mitigate hours and days of hard to track down debugging.

RELEASE_NOTES.md
```diff
@@ -82,3 +82,10 @@ n["foo"] << "buzz"
 before this would have mutated the original string in-place so that `node["foo"]` and `node.default["foo"]` would have changed to "fizzbuzz"
 while now they remain "fizz" and only the mutable `n["foo"]` copy is changed to "fizzbuzz".
 
+### Freezing immutable merged attributes
+
+Since Chef 11 merged node attributes have been intended to be immutable but the merged strings have not been frozen.  In Chef 13, in the
+process of merging the node attributes strings and other simple objects are dup'd and frozen.  In order to get a mutable copy, you can
+now correctly use the `node.dup` or `node.to_hash` methods, or you should mutate the object correctly through its precedence level like
+`node.default["some_string"] << "appending_this"`.
```
stevendanna: Would code that looks like this:

```
some_opts = node['default_opts']
some_opts << '  --fkermit' if is_frog?
```

would need to change? If so, we might want to put an explicit example or two somewhere on the best way to fix such code.

jaymzh:
👍

RELEASE_NOTES.md
lamont-granquist: that code would need to change.  if there was no intent to mutate the node attribute and that was a mistake then use non-mutating methods like '+' or call dup.  if there was an intent to mutate then go in through the precedence accessor.  the choice over what to do is why i opted for more words instead of examples though since there's no cookie-cutter approach to fixing it.

PR #5907, opened 2017-03-16
Chef-13:  Nillable properties

lamont-granquist:
Remove deprecations and now properties are nillable and passing a nil is
now always a set, not a get.

Note the rules around nil:

- nils are never coerced (nil is nil).  don't try to coerce nil to something else (probably use a default value instead of nil)
- nils are not validated if there's no explicit default value set

So:

```ruby
property :foo, String
property :foo, [ String, nil ]
property :foo, [ String, nil ], default: nil
```

Those are all similar and the default value in those cases is nil, and nil can be set.

```ruby
property :foo, String, default: "bar"
```

Nil is not ever a valid value there, so it is validated, and cannot be set.

There is no support for "nil means unset default value, but once set nil will fail validation".   Perhaps this only matters in my brain, but I thought that would be a useful construct at one point, but it really breaks the ability to e.g. construct a current_resource from the new_resource because you can't set the nil default value.  The better construct there is to add `required: true` since that forces the user to supply a value, then nil will raise an exception upon access -- but by the time the current_resource is being constructed it must have a value.


lamont-granquist:
@chef/client-core if it goes green its reviewable, if it doesn't i'll deal with it tomorrow...

lamont-granquist:
looks like i missed one and the cron cookbook looks like it needs a bump.

lamont-granquist:
@chef/client-core fixed, green and reviewable

PR #5908, opened 2017-03-16
add missing functional tests for users

harikesh-kolekar:

### [Do Not Merge] Add missing functional tests for users if applicable

https://github.com/chef/chef/blob/master/spec/functional/win32/security_spec.rb#L33
I'm fairly certain that the mixlib-shellout user code works now, we were just talking about using it with powershell_out. If so, we create the necessary functional tests here.

### Check List

- [ ] All tests pass


lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,126 @@
...
+            raise "Attempted to open a logon session that was already open."
+          end
+
+          username = wstring(original_username)
+          password = wstring(original_password)
+          domain = wstring(original_domain)
+
+          status = Chef::ReservedNames::Win32::API::Security.LogonUserW(username, domain, password, Chef::ReservedNames::Win32::API::Security::LOGON32_LOGON_NETWORK, Chef::ReservedNames::Win32::API::Security::LOGON32_PROVIDER_DEFAULT, token)
+
+          if !status
```
coderanger: You change the spacing of this kind of call a few times, should be consistent or just use `unless`.

lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,52 @@
...
+  module Mixin
+    module UserContext
+
+      def with_user_context(user, password, domain = nil, &block)
+        if node["platform_family"] != "windows"
+          raise Exceptions::UnsupportedPlatform, "User context impersonation is supported only on the Windows platform"
+        end
+
+        if ! block_given?
+          raise ArgumentError, "You must supply a block to `with_user_context`"
```
coderanger: Mixing old and new block syntax.

lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,52 @@
...
+            logon_session.open
+            logon_session.set_user_context
+          end
+          yield
+        ensure
+          logon_session.close if logon_session
+        end
+      end
+
+      protected(:with_user_context)
```
coderanger: We don't usually use the explicit form in Chef code

lib/chef/util/windows/logon_session.rb
```diff
@@ -0,0 +1,126 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+#
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
coderanger: It is not 2015

lib/chef/mixin/user_context.rb
```diff
@@ -0,0 +1,52 @@
+#
+# Author:: Adam Edwards (<adamed@chef.io>)
+# Copyright:: Copyright (c) 2016 Chef Software, Inc.
```
coderanger: It is not 2016

coderanger:
This adds more than just tests, please clarify the goal.

harikesh-kolekar:
@coderanger  I have referred the PR: https://github.com/chef/chef/pull/5832 to login with windows user to test admin privileges so It has more code than just the test.

btm:
@coderanger these tests aren't part of #5832, but that PR has the mixin to allow us to run code as a different (non privileged) user on Windows which is needed for these tests. The functionality is all part of an old PR #4298 by @adamedx that we've been trying to break down into smaller pieces and get finished.

Another option would be to just add the mixin without anything using it in a separate PR, and then write the tests and feature separately. The feature in #5832 is difficult to get the functionality right and needs further work.

adamedx:
Hey folks, let me know if you need my help on this. And sorry for the pain in how this was factored @btm. Originally I had two different features as part of one giant pr because they shared common code. Then I decided to retroactively separate them, but I had managed to add some shared examples between the two features and had to artificially include part of one feature in the other. This may have required an ordering in their merge. 

Anyway, it all seemed like a good idea at the time. :)

coderanger:
@btm No objection on the PR, just probably needs a new title that better reflects what this is actually for (which matters since those titles end up in the changelog) :)

btm:
@harikesh-kolekar now that https://github.com/chef/chef/pull/5832 is merged, can you rebase this against master and wrap it up?

Vasu1105:
@btm This looks to be taken care in #6425

PR #5912, opened 2017-03-16
Chef-13:  raise on properties redefining inherited methods

lamont-granquist:
flips the deprecation warning over to a raise


lamont-granquist:
@chef/client-core should be ready to review

PR #5914, opened 2017-03-17
remove old platform mapping code

lamont-granquist:
Chef::Platform.set is now removed in favor of the dynamic
ProviderResolver `provides` syntax.


lamont-granquist:
@chef/client-core this should be ready

.travis.yml
```diff
@@ -272,43 +272,43 @@ matrix:
...
+#      - sudo iptables -L DOCKER || ( echo "DOCKER iptables chain missing" ; sudo iptables -N DOCKER )
+#      - cd kitchen-tests
+#    script:
+#      - bundle exec kitchen test awesome-customers-rhel-centos-7
+#    after_failure:
+#      - cat .kitchen/logs/kitchen.log
+#    env:
+#      - AWESOME_CUSTOMERS_RHEL=1
+#      - KITCHEN_YAML=.kitchen.travis.yml
+#    ### END TEST KITCHEN ONLY ###
```
thommay: This either needs re-uncommenting or deleting.

lamont-granquist: no, that's going to get uncommented by the time that chef-13 goes live.  its blocked on chef_gem_mysql2 and mysql 8.x cookbooks being used by those tests, but learnchef needs to make changes to get onto mysql 8.x and the changes to chef_gem_myxql2 that it needs are committed to master but not yet released.  the mysql 8.x issue is also complicated by the fact that adding the repo has been removed from the mysql cookbook so learnechef needs to pick that up explicitly or something like that.  so there's a whole in-progress yak shave going on here...

thommay: OK. We need a reminder of stuff like this...

PR #5916, opened 2017-03-17
Chef-13: add Chef-14.0 deprecation for chef-solo legacy mode

lamont-granquist:
the task isn't finished until we can delete the code.

### DIFF
--- a/lib/chef/chef_class.rb
+++ b/lib/chef/chef_class.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Lamont Granquist (<lamont@chef.io>)
-# Copyright:: Copyright 2015-2016, Chef Software, Inc.
+# Copyright:: Copyright 2015-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@coderanger resource cloning spec test attacks again

coderanger:
Trololol brittle tests ftw.

coderanger:
So putting this in makes ChefSpec's SoloRunner spam warnings. I vote we add an `unless defined?(ChefSpec)` for now, until ChefSpec gets updated, which in turn requires fixing some perf problems with zero or otherwise hacking around them in ChefSpec.

coderanger:
This would also incorrectly impact people using the `chef_solo` provisioner in Kitchen, so maybe check `ENV['TEST_KITCHEN']` too.

lamont-granquist:
IDK it feels like if we don't start nagging people it'll never get done.  up to @thommay

coderanger:
So the specific issue is that (legacy) solo mode is still faster than zero mode for testing, not by a lot but if you're like me and run thousands of test runs a day even an extra 30 seconds on moving files around is a big change. I realize there very well might be no one else that this applies to, but putting in a warning that I can't silence that both spams on every spec running and makes `deprecations_as_errors` unusable (since it would always fire).

PR #5917, opened 2017-03-20
[CHEF-3775] Converting recipe loading to recursive.

Tzrlk:
This _should_ fix the issue raised on the jira a while back where recipes couldn't be referenced if they were in subfolders. The format I went for was simply recipe[cookbook::folder/recipe], though if desired I can alter it to be recipe[cookbook::folder::recipe].

### DIFF
--- a/lib/chef/cookbook/cookbook_version_loader.rb
+++ b/lib/chef/cookbook/cookbook_version_loader.rb
@@ -86,7 +86,7 @@ class Chef
 
         load_as(:attribute_filenames, "attributes", "*.rb")
         load_as(:definition_filenames, "definitions", "*.rb")
-        load_as(:recipe_filenames, "recipes", "*.rb")
+        load_recursively_as(:recipe_filenames, "recipes", "*.rb")
         load_recursively_as(:library_filenames, "libraries", "*")
         load_recursively_as(:template_filenames, "templates", "*")
         load_recursively_as(:file_filenames, "files", "*")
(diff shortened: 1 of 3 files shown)


coderanger:
This was a feature, not a bug IMO, and I bet this breaks in a lot of other places. :-1: from me.

coderanger:
Also regardless of goals, a change of something this fundamental should go through the RFC process most likely, given the broad ramifications.

lamont-granquist:
it looks like `@root_paths[0]` is blowing up because its nil in a lot of cases -- on a lot of functional tests, so its real breakage.

also no DCO, and if we're going to do this i'd prefer the more consistent foo::bar::baz syntax, and probably does need an RFC.

also the usual suggestion that if you're writing enough recipe code like this in a single cookbook that you want to organize things like this, you probably want to consider refactoring.  the `apache2` cookbook is a good example of a bad pattern where a hundred recipes really need to be replaced by one single resource that can be called arbitrarily.

Tzrlk:
@lamont-granquist My use-case is specifically for a "provision" cookbook. If node provisioning were handled with a different structure, it'd be a different story, but that'd be a waaay bigger change than this.

I'll look up the RFC procedure and put something together when I can get time.

thommay:
I think we're going to close this PR given it's age and our general nervousness about this. If you'd like to write up an RFC we can have a discussion there.

Thanks for your submission!

PR #5920, opened 2017-03-20
Chef-13: remove more deprecated provider_resolver code

lamont-granquist:
I think this nukes descendants tracker, although might need to clean it
up in a few more spots.


lamont-granquist:
@coderanger you seem to be dependent upon the descendants tracker code?

PR #5924, opened 2017-03-20
Chef-13: Remove support for the sort option to searches.

coderanger:
I'm 99% sure sorting was just never implemented in Erchef. Did I just make that up?


coderanger:
Also removing this without a deprecation because unless I made up that this was never in Erchef, it's been non-functional for a long time, like since 11.0.

lamont-granquist:
IDK @stevendanna ?

stevendanna:
👍 On this change. The sorting feature has been broken for roughly 5 years on the erchef/hosted-chef/private-chef/enterprise-chef side.  It is possible it worked before the schema change in solr/expander (moving from a field per node attribute to a single field for all node content).  I can confirm that in the current code of erchef the sort parameter that is sent to Solr/ES is hard-coded.

I think the only reason not to remove the parameter would be if we care about not breaking people who were passing the --sort flag in scripts and such.

thommay:
Better to break them so they know stuff isn't doing what they expect, i think.

PR #5925, opened 2017-03-20
Chef-13: remove deprecated run_context methods

lamont-granquist:


lamont-granquist:
your tests are passing, i'm just seeing some transient travis failures

PR #5926, opened 2017-03-21
Chef-13: remove node_map back-compat

lamont-granquist:
removes old Chef-10 era APIs that were not widely used.


lamont-granquist:
@chef/client-core also ready to be reviewed

coderanger:
Also I should do a bulk copyright year update again :P

PR #5932, opened 2017-03-21
Chef-13: Remove deprecated Chef::Client attrs

lamont-granquist:
nothing uses these, nothing should use these, no tests, deprecated for ages, goodbye...


lamont-granquist:
@chef/client-core this one should be the simplest of tap-ins

lamont-granquist:
sometimes we noodle way too hard on all the possible ways we might break someone that was doing something incredibly silly, and our capacity to imagine user silliness is nearly unbounded (because, to be fair to us, every now and then users do come up with a real whopper...)

lamont-granquist:
i think i'm skipping this one in the RELEASE_NOTES deliberately, it should not affect anyone and CHANGELOG.md should be good enough for it.

PR #5933, opened 2017-03-21
ideas for the issue template

lamont-granquist:
maybe try this?

- directly put the purpose of the issue tracker directly in front of users
- use direct links instead of dereferenced links to links to links
- also tried putting the "form test" inside of <>'s to see if more people will delete it that way.  i suspect that with the changes we'll get the big splash notice at the top of every submission, but i can suffer through deleting that for users as long as it improves the correct usage of the tracker and filters users towards the correct resources...


lamont-granquist:
@client-core thoughts?

thommay:
I think the structure is good, but the words need a little work. I'll try and think of some that i like

PR #5936, opened 2017-03-22
Use the ruby 2.4.1 release

thommay:
First, let's see if they've fixed rspec. Then we can bump omnibus-software and go from there


lamont-granquist:
looks like travis' rvm needs to get updated.  you might be able to `rvm get head && rvm install ruby-2.4.1` very early in the build, but that'll make it take a whole lot longer and kind of defeat the purpose...

mawatech:
The binary is available now:

https://s3.amazonaws.com/travis-rubies/binaries/ubuntu/12.04/x86_64/ruby-2.4.1.tar.bz2

Perhaps you can trigger a rebuild?

thommay:
@chef/maintainers assuming this goes green. Numbers look pretty tasty. I also think we should only test with 2.4.1 for 13; thoughts?

lamont-granquist:
yeah should bump up everything possible to 2.4.1 but should probably still test on both 2.3.3 and 2.4.1 for `rake spec`.

lamont-granquist:
bumping the Gemfile.lock failed on chef-zero, need to investigate that in parallel.

bumping bundler to either 1.13.x or 1.14.x is also not working right now, and gets into fights with rake tasks, there's probably some additional uncrazysaucing that needs to happen.

lamont-granquist:
(appveyor is also still on ruby-2.3 because there's no ruby-2.4)

smurawski:
Re: appveyor and ruby 2.4 - https://github.com/appveyor/ci/issues/1350

PR #5937, opened 2017-03-22
Add InSpec to chef omnibus builds

adamleff:
This change adds InSpec as a dependency of Chef, making it easier
for our community to use InSpec for testing without requiring
them to `chef_gem` install it prior to use.

This also helps our users who wish to use InSpec but are in an
air-gapped environment. Including our preferred testing library
in our Omnibus builds will make it much easier for those users
to use InSpec.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem "cheffish", "~> 13" # required for rspec tests
 group(:omnibus_package) do
   gem "appbundler"
   gem "rb-readline"
+  gem "inspec"
   # nokogiri has no ruby-2.4 version for windows so it cannot go into our Gemfile.lock
   #  gem "nokogiri", ">= 1.7.1"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/chef/chef-server
-  revision: b670bd5f2ed1f56a39ab0c211efd4204b6978422
+  revision: 07fea253ceea254af8fbd9c602a04017751cae21
   specs:
     oc-chef-pedant (2.2.0)
       activesupport (>= 4.2.7.1, < 6.0)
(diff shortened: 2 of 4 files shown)


adamleff:
Tested ad-hoc build, which succeeded except on the s390x builders due to network issues.

coderanger:
👎 Unless Chef is actually going to depend on InSpec in some way (which I would be against given what happened with audit mode) this belongs in the installers not here.

coderanger:
Specifically in https://github.com/chef/chef/blob/2fcc06f5cad0f70eb1cc58d118ba10c7999fa78a/Gemfile#L22-L26 I think, but I defer to others that know the omnibus plumbing better.

adamleff:
I'm totally fine with that approach too. I'll make the change and run another test build.

adamleff:
Doesn't appear that the chef-zero travis job failure has anything to do with this change, but if someone disagrees, please lemme know.

thommay:
@adamleff it is indeed entirely unrelated.

lamont-granquist:
travis failure is because you're also bumping cheffish and chef-zero so i think thom's PR has to get merged first.  you will break master if you mash merge before thom.

lamont-granquist:
i'm reasonably certain you need to be adding a definition to omnibus itself kind of like:

https://github.com/chef/chef/blob/master/omnibus/config/software/chef-gem-byebug.rb

and adding that to the chef definition.

adamleff:
@lamont-granquist re: bumping the other dependencies, that's certainly not intended but rather that's what `rake bundle:install` did.

Re: the omnibus software definition, it didn't appear to be *necessary*... I did an ad-hoc build and inspec+train was included. But as I read these definitions they appear to be desired for caching purposes, so I'll go ahead and follow this pattern then too.

adamleff:
http://manhattan.ci.chef.co/job/chef-trigger-ad_hoc/46/

adamleff:
Putting a brief hold on this to see if I can rip nokogiri out of InSpec.

adamleff:
Travis test failures unrelated, will be happy to rebase once #6009 is merged. But otherwise, this should be good to go.

PR #5938, opened 2017-03-22
Chef-13: Support nameless resources and remove deprecated multi-arg resources

lamont-granquist:
This makes nameless resources work in the DSL:

```ruby
apt_update
```

It does it by giving it an empty-string name of ""

It also drops support for multi-arg resources, that has been deprecated
for some time:

```ruby
some_resource "foo", "bar", "baz"
```

Note that multipackage package resources do not use multiple args, but a
single argument which is an array:

```ruby
package [ "lsof", "strace", "tcpdump" ]
```

So this change does not affect that usage.  Multi-arg has been
deprecated for some time now and its not clear that it was ever used by
anyone.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -142,3 +142,18 @@ available in the [`poise-python`](https://github.com/poise/poise-python) cookboo
 
 We've upgraded to the latest stable release of the Ruby programming
 language.
+
+### Resource can now declare a default name
+
+The core `apt_update` resource can now be declared without any name argument, no need for `apt_update "this string doesn't matter but
+why do i have to type it?"`.
+
+This can be used by any other resource by just overriding the name property and supplying a default:
+
+```ruby
+  property :name, String, default: ""
+```
+
+Notifications to resources with empty strings as their name is also supported via either the bare resource name (`apt_update` --
+matches what the user types in the DSL) or with empty brackets (`apt_update[]` -- matches the resource notification pattern).
+
--- a/lib/chef/dsl/resources.rb
+++ b/lib/chef/dsl/resources.rb
@@ -34,11 +34,10 @@ class Chef
 
       def self.add_resource_dsl(dsl_name)
         module_eval(<<-EOM, __FILE__, __LINE__ + 1)
-            def #{dsl_name}(*args, &block)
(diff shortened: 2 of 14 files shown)


lamont-granquist:
note that I implemented notifications that work with any of:  `apt_update`, `apt_update[]` or `apt_update: ""`, the canonical representation is `apt_update[]`

damacus:
Awesome. This'll tidy up a lot of custom resources! 

What happens if a nameless resource is declared more than once? Does it just collect them all and the last one wins?

lamont-granquist:
you get multiple resources in the resource collection, just like the old 3694 issue, but there's no cloning of properties any more in Chef-13, if you notify the resource then the last one will win.  if the resource is idempotent then only the first one should fire and the rest will hit their own idempotency checks.  renaming the resource works around that problem just like working around the old 3694 error.  TL;DR is "behaves exactly like any other resource"

lamont-granquist:
@coderanger you see any problems with this?  i considered nil for the default value, but that means names now become String-or-nil and that'd be a bit more invasive, while this change means names are still always Strings.

coderanger:
@lamont-granquist So my only concern, and part of why we didn't do this in 12.5, is if it should be opt-in at the resource class level. There was concern that this could lead to DSL-y confusion because the DSL includes common stuff like `user` and `group` and it gets dumped in some contexts where that could make typos or misunderstandings lead to unexpected behavior. For resources like `apt_update` which the name provides no value, totally on board, but maybe not for everything? We could do that by adding a check for `==''` in `Resource#initialize` or similar, and then override the check in specific classes to allow it.

@damacus Basically this means that `resource ''` and `resource` are synonyms, with some added niceness for the string reference syntax.

lamont-granquist:
okay nameless properties are now opt-in, bare `user` will throw a validation failed error (moved that responsibility off of the resource_builder and onto the property) and now you can just override the actual name property and supply a default value of "" and it'll work.  still basically the same fixes to the resource collection to support targetting resources with empty-strings as names, but removed support for nil names in the external API of the resource collection.

lamont-granquist:
@chef/client-core not positive it'll turn green but it should be ready for review

PR #5940, opened 2017-03-23
Don't `rescue Exception` in retryable resources

srenatus:
### Description

Without this, sending an interrupt signal (hitting Ctrl+C) during a
chef-client run that is currently executing a resources that has
`retries` set will not interrupt the run.

If the retryable resource is shelling out, the spawned process also gets
the INT signal and will stop -- however, the chef-run would go on,
merely noticing that the resource has failed and theres X-1 retries
left.

It is imaginable that users depend on some (whacky) Ruby script in a
retriable resource that raises SystemExit, and they want this to _not_
stop the chef-run. So it's up for debate if we just want to merge this
or make it configurable.

In general, I'd like Ctrl+C to _stop my chef-run_, but this expectation
could also be misled.

### Issues Resolved

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -589,7 +589,7 @@ class Chef
       begin
         return if should_skip?(action)
         provider_for_action(action).run_action
-      rescue Exception => e
+      rescue StandardError => e
         if ignore_failure
           Chef::Log.error("#{custom_exception_message(e)}; ignore_failure is set, continuing")
           events.resource_failed(self, action, e)
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,6 +53,10 @@ The remediation is removing the self-dependency `depends` line in the metadata.
 
 Retained only for the service resource (where it makes some sense) and for the mount resource.
 
+### Removed retrying of non-StandardError exceptions for Chef::Resource
+
+Exceptions not decending from StandardError (e.g. LoadError, SecurityError, SystemExit) will no longer trigger a retry if they are raised during the executiong of a resources with a non-zero retries setting.
+
 ### Removed deprecated `method_missing` access from the Chef::Node object
 
 Previously, the syntax `node.foo.bar` could be used to mean `node["foo"]["bar"]`, but this API had sharp edges where methods collided
--- a/spec/unit/resource_spec.rb
+++ b/spec/unit/resource_spec.rb
@@ -549,6 +549,18 @@ describe Chef::Resource do
       expect { retriable_resource.run_action(:purr) }.to raise_error(RuntimeError)
       expect(retriable_resource.retries).to eq(3)
     end
+
+    it "should not rescue from non-StandardError exceptions" do
+      retriable_resource.retries(3)
+      retriable_resource.retry_delay(0) # No need to wait.
+
+      provider = Chef::Provider::SnakeOil.new(retriable_resource, run_context)
+      allow(Chef::Provider::SnakeOil).to receive(:new).and_return(provider)
+      allow(provider).to receive(:action_purr).and_raise(LoadError)
+
+      expect(retriable_resource).not_to receive(:sleep)
+      expect { retriable_resource.run_action(:purr) }.to raise_error(LoadError)
+    end
   end
 
   it "runs an action by finding its provider, loading the current resource and then running the action" do
--- a/spec/unit/runner_spec.rb
+++ b/spec/unit/runner_spec.rb
@@ -122,25 +122,26 @@ describe Chef::Runner do
 
     it "should raise exceptions as thrown by a provider" do
       provider = Chef::Provider::SnakeOil.new(run_context.resource_collection[0], run_context)
-      allow(Chef::Provider::SnakeOil).to receive(:new).once.and_return(provider)
-      allow(provider).to receive(:action_sell).once.and_raise(ArgumentError)
+      expect(Chef::Provider::SnakeOil).to receive(:new).once.and_return(provider)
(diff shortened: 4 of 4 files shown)


dmccown:
I'd like to see some tests around this change.

srenatus:
@dmccown sure -- this just trying to start a discussion 😃

lib/chef/resource.rb
```diff
@@ -590,6 +590,7 @@ def run_action(action, notification_type = nil, notifying_resource = nil)
         return if should_skip?(action)
         provider_for_action(action).run_action
       rescue Exception => e
+        raise if e.class == SystemExit
```
lamont-granquist: that should just be a `rescue SystemExit; raise` case.

lamont-granquist:
is SystemExit only for ctrl-c inside of ruby code or does it get raised by anything else like system or backticks?

coderanger:
So ctrl-c isn't supposed to raise SystemExit AFAIK, it raises Interrupt which is a subclass of SignalException. SystemExit is raised by calling the `exit` method.

coderanger:
(we should probably not be swallowing either SystemExit or SignalException, and arguably NoMemoryError and SystemStackError should be in there too)

lamont-granquist:
okay so this smells a bit more like "you shouldn't rescue Exception" then, which sounds pretty reasonable.

lamont-granquist:
and ruby code inside of a ruby_block or whatever shouldn't be calling `exit` directly unless they really mean to panic everything, they should be raising something like RuntimeError if they expect to get retried.

coderanger:
@lamont-granquist Yeah but if you're calling in to a gem or something that has stupid code in it, we should probably Do What You Mean.

lamont-granquist:
shouldn't the chef code inside the resource then wrap the call with a rescue to re-raise something less dumb?  is it core chef's responsibility?

coderanger:
I guess we should establish what the meaning of calling `exit()` inside a ruby_block or providers in general is. Given how rarely this comes up I don't really care (SignalException, NoMemoryError, and SystemStackError are the more dangerous things to catch) but we should pick a semantic. I'm cool with it meaning "no really, exit like I told you to and damn the torpedoes" since the author of the code can catch SystemExit themselves if they don't want that behavior. There are sometimes situations where being able to set chef-client's exit code from within a recipe would be useful.

lamont-granquist:
Yeah generally I think I'm somewhat against "make it configurable" now since users can always catch exceptions and re-raise something less toxic if that's what they want, and that use case should be uncommon.

And I'd say that if you want retries you should `raise` something more polite that inherits from RuntimeError or whatever while `exit` should mean `exit` and shouldn't get retried.

coderanger:
So probably the thing we should add is

```ruby
rescue SystemExit, SignalException, NoMemoryError, SystemStackError
  raise
```

lamont-granquist:
`SecurityError` and `ScriptError` (and `LoadError`, `SyntaxError`, `NotImplementedError`) are pretty unlikely to fix themselves on retries.

still kind of smells to me like we should have done just a plain `rescue` and not `rescue Exception`...

i'm 👍  either way though.

and this should get fixed up soonish before Chef-13 goes live or else it'll block another year.

stevendanna:
> So ctrl-c isn't supposed to raise SystemExit AFAIK, it raises Interrupt which is a subclass of SignalException. SystemExit is raised by calling the exit method.

Ah, right. I think we were seeing the SystemExit in the stack trace perhaps because of the interrupt handler we have installed there.  But I am all for expanding this to a broader set of "Almost certainly something you don't want retried" exceptions.

srenatus:
I should have posted more details first, sorry -- indeed, I think the SystemExit came from [the trap handler installed here](https://github.com/chef/chef/blob/2757379abb86a6f698b02930b3c2d0ae515beb8d/lib/chef/application.rb#L62-L64), which calls [Process.exit](https://github.com/chef/chef/blob/2757379abb86a6f698b02930b3c2d0ae515beb8d/lib/chef/application.rb#L361), which in turn
> Initiates the termination of the Ruby script by raising the SystemExit exception. (https://ruby-doc.org/core-1.9.3/Process.html#method-c-exit)

srenatus:
I'll push another commit changing this to `rescue StandardError => e` after I've looked after the tests.

srenatus:
Ugh lot's of 💥 on travis. The audit check would be fixed by #5943 (thanks @brugidou), the rest are kind of complicated... any help would be appreciated (Just rerun? or is something going on that I'm missing? I wouldn't expect this to come from this change...)

srenatus:
Digging into this, and changing some of the `allow` to `expect` in the failing test ([my playground branch is here](https://github.com/chef/chef/tree/ssd-sr/re-raise-system-exit-playground)), I've noticed that [this doesn't get executed](https://github.com/chef/chef/blob/master/spec/unit/runner_spec.rb#L141):

```
  1) Chef::Runner when we are doing dynamic provider resolution should retry with the specified delay if retries are specified
     Failure/Error: expect(Chef::Provider::SnakeOil).to receive(:new).once.and_return(provider)

       (Chef::Provider::SnakeOil (class)).new(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments
     # ./spec/unit/runner_spec.rb:141:in `block (3 levels) in <top (required)>'
     # b/rspec:17:in `load'
     # b/rspec:17:in `<main>'
```

...so I'm not entirely sure this test works as expected on master. I'll keep digging on Monday. (Happy Friday)

srenatus:
@coderanger @lamont-granquist @dmccown By any chance, could have another look at this? 😃

lamont-granquist:
👍

PR #5941, opened 2017-03-23
Make ResourceReporter smarter to get resource identity and state

### LINKED ISSUE #5490: no implicit conversion of Chef::DelayedEvaluator into String
## Description

We've had reports downstream in https://github.com/rackspace-cookbooks/chef-slack_handler/issues/37 that releases of Chef > 12.6.x have broken some functionality in the `cookbook_file` provider.  It looks like this [same issue](https://github.com/chef/chef/pull/1361) cropped up in [CHEF-5162](https://tickets.opscode.com/browse/CHEF-5162) and some other places, but I'm not specifically aware of the fix for cookbook file.
## Chef Version

```
➜  ~ chef --version
Chef Development Kit Version: 0.18.30
chef-client version: 12.14.89
delivery version: master (d86679335580be3de22996ef294b20d525889d8d)
berks version: 5.1.0
kitchen version: 1.13.2
```
## Platform Version

Mac OS 10.x
## Replication Case

I'm actually not sure how to replicate it. I'm passing the bug upstream, but I haven't directly encountered it.
## Client Output

```
  * cookbook_file[/var/chef/handlers/slack_handler_util.rb] action create

    Error executing action `create` on resource 'cookbook_file[/var/chef/handlers/slack_handler_util.rb]'

    TypeError
    ---------
    no implicit conversion of Chef::DelayedEvaluator into String

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/slack_handler/recipes/default.rb:42:in `from_file'
    /var/chef/cache/cookbooks/moxi-slack/recipes/default.rb:14:in `from_file'

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/slack_handler/recipes/default.rb

     37: cookbook_file "#{node['chef_handler']['handler_path']}/slack_handler_util.rb" do
     38:   source 'slack_handler_util.rb'
     39:   mode "0600"
     40:   action :nothing
     41:   # end
     42: end.run_action(:create)

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/slack_handler/recipes/default.rb:37:in `from_file'

    cookbook_file("/var/chef/handlers/slack_handler_util.rb") do
      provider Chef::Provider::CookbookFile
      action [:nothing]
      retries 0
      retry_delay 2
      source "slack_handler_util.rb"
      declared_type "cookbook_file"
      cookbook_name "slack_handler"
      recipe_name "default"
      mode "0600"
      atomic_update #<Chef::DelayedEvaluator:0x00000001d8ce98@/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.13.37/lib/chef/resource/file.rb:53>
      path #<Chef::DelayedEvaluator:0x00000001d8df78@/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.13.37/lib/chef/property.rb:180>
    end

    Platform:
    ---------
    x86_64-linux
```

afiune:
### Description
Lets make the ResourceReporter smarter, we will try to get the identity
and state (after and before) of the resource but if we can't because of
an Exception, lets sends a message to the data collector containing a
static resource identity since we were unable to generate a proper one.

The problem is coming from the ResourceReporter, when it is trying to
report back the resource identity and state, it is using basic methods
from the Base class that should be secure to call from eny subscriber.

This is actually not a bug since Chef is doing the right thing, that is
to throw an error when the identity property is nil (or is not the right
kind that we configured for) so, rather than making identity properties
smarter for lazy evaluation we are going to make the ResourceReporter a
bit smarter so that it doesn't explode when the use configured a
resource with a wrong identity property.

### Issues Resolved

* COOL-691
* ZD-13808

closes #5490

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/data_collector/resource_report.rb
+++ b/lib/chef/data_collector/resource_report.rb
@@ -18,6 +18,8 @@
 # limitations under the License.
 #
 
+require "chef/exceptions"
+
 class Chef
   class DataCollector
     class ResourceReport
@@ -67,9 +69,9 @@ class Chef
         hash = {
           "type"           => new_resource.resource_name.to_sym,
           "name"           => new_resource.name.to_s,
-          "id"             => new_resource.identity.to_s,
-          "after"          => new_resource.state_for_resource_reporter,
-          "before"         => current_resource ? current_resource.state_for_resource_reporter : {},
+          "id"             => resource_identity,
+          "after"          => new_resource_state_reporter,
+          "before"         => current_resource_state_reporter,
           "duration"       => elapsed_time_in_milliseconds.to_s,
           "delta"          => new_resource.respond_to?(:diff) && potentially_changed? ? new_resource.diff : "",
           "ignore_failure" => new_resource.ignore_failure,
@@ -83,13 +85,39 @@ class Chef
           hash["recipe_name"]      = new_resource.recipe_name
         end
 
-        hash["conditional"]   = conditional.to_text if status == "skipped"
+        hash["conditional"] = conditional.to_text if status == "skipped"
         hash["error_message"] = exception.message unless exception.nil?
 
         hash
       end
       alias :to_h :to_hash
       alias :for_json :to_hash
+
+      # We should be able to call the identity of a resource safely, but there
+      # is an edge case where resources that have a lazy property that is both
+      # the name_property and the identity property, it will thow a validation
+      # exception causing the chef-client run to fail. We are not fixing this
+      # case since Chef is actually doing the right thing but we are making the
+      # ResourceReporter smarter so that it detects the failure and sends a
+      # message to the data collector containing a static resource identity
+      # since we were unable to generate a proper one.
+      def resource_identity
+        new_resource.identity.to_s
+      rescue => e
+        "unknown identity (due to #{e.class})"
+      end
+
+      def new_resource_state_reporter
+        new_resource.state_for_resource_reporter
+      rescue
+        {}
+      end
+
+      def current_resource_state_reporter
+        current_resource ? current_resource.state_for_resource_reporter : {}
+      rescue
+        {}
+      end
     end
   end
 end
--- a/spec/unit/data_collector/resource_report_spec.rb
+++ b/spec/unit/data_collector/resource_report_spec.rb
@@ -1,0 +1,145 @@
+#
+# Author:: Salim Afiune (<afiune@chef.io)
+#
+# Copyright:: Copyright 2012-2017, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "spec_helper"
+
+describe Chef::DataCollector::ResourceReport do
+  let(:cookbook_repo_path) { File.join(CHEF_SPEC_DATA, "cookbooks") }
+  let(:cookbook_collection) { Chef::CookbookCollection.new(Chef::CookbookLoader.new(cookbook_repo_path)) }
+  let(:node) { Chef::Node.new }
+  let(:events) { Chef::EventDispatch::Dispatcher.new }
+  let(:run_context) { Chef::RunContext.new(node, cookbook_collection, events) }
+  let(:resource) { Chef::Resource.new("zelda", run_context) }
+  let(:report) { described_class.new(resource, :create) }
+
+  describe "#skipped" do
+    let(:conditional) { double("Chef::Resource::Conditional") }
+
+    it "should set status and conditional" do
+      report.skipped(conditional)
+      expect(report.conditional).to eq conditional
+      expect(report.status).to eq "skipped"
+    end
+  end
+
+  describe "#up_to_date" do
+    it "should set status" do
+      report.up_to_date
+      expect(report.status).to eq "up-to-date"
+    end
+  end
+
+  describe "#updated" do
+    it "should set status" do
+      report.updated
+      expect(report.status).to eq "updated"
+    end
+  end
+
+  describe "#elapsed_time_in_milliseconds" do
+
+    context "when elapsed_time is not set" do
+      it "should return nil" do
+        allow(report).to receive(:elapsed_time).and_return(nil)
+        expect(report.elapsed_time_in_milliseconds).to eq nil
+      end
+    end
+
+    context "when elapsed_time is set" do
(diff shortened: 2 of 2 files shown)


spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  describe "#updated" do
+    it "should set status" do
+      report.updated
+      expect(report.status).to eq "updated"
+    end
+  end
+
+  describe "#elapsed_time_in_milliseconds" do
```
adamleff: This feels like we're testing the wrong thing. The only thing `#elapsed_time_in_milliseconds` does is format a number. So we should probably mock out the call to `#elapsed_time` and test for a nil return and a number return.

spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      it "should return nil" do
+        expect(report.elapsed_time_in_milliseconds).to eq nil
+      end
+    end
+
+    context "when resource finished" do
+      it "should return the elapsed_time in milliseconds" do
+        report.finish
+        # zero because computers are fast
+        expect(report.elapsed_time_in_milliseconds).to eq 0
```
adamleff: If we do want to test it this way, let's never assume that this will always be 0... this is a race-condition / unstable test waiting to happen. :) we'd need to mock the resource's `#elapsed_time` method and have it return a known number vs. assuming that the testing computer is fast.

But again, see above comment - I don't think we need to worry about that here. This is more of a functional test than a unit test IMO.

spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+          "id" => "hyrule",
+          "ignore_failure" => false,
+          "name" => "hyrule",
+          "result" => "create",
+          "status" => "unprocessed",
+          "type" => :zelda,
+        }
+      end
+
+      it "should work" do
```
adamleff: nitpick, but I'd make the example name more clear, such as "returns a hash containing the expected values"

spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      end
+    end
+
+    context "for a lazy_resource that got skipped" do
+      let(:resource) do
+        c = Class.new(Chef::Resource) do
+          resource_name "link"
+          property :sword, String, name_property: true, identity: true
+        end
+        l = c.new("hyrule")
```
adamleff: Please use better, more intentional variable names. 🙂

lib/chef/data_collector/resource_report.rb
```diff
@@ -83,13 +85,38 @@ def to_hash
...
       end
       alias :to_h :to_hash
       alias :for_json :to_hash
+
+      # We should be able to call the identity of a resource safely, but there
+      # is an edge case where resources that have a lazy property that is both
+      # the name_property and the identity property, it will thow a validation
+      # exception causing the chef-client run to fail. We are not fixing this
+      # case since Chef is actually doing the right thing but we are making the
+      # ResourceReporter smarter so that it detects the failure and sends dummy
```
adamleff: Rather than calling this "dummy data", how about we change that last part of the sentence to: "sends a message to the data collector containing a static resource identity since we were unable to generate a proper one."

spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+          "result" => "create",
+          "status" => "skipped",
+          "type" => :link,
+        }
+      end
+      let(:conditional) do
+        double("Chef::Resource::Conditional", :to_text => "because")
+      end
+
+      before { report.skipped(conditional) }
```
adamleff: Since there's only one example, we can toss this in the `it` block like you've done with the other tests, yes?

adamleff:
Oh, and many thanks for adding tests for the resource report class!

spec/unit/data_collector/resource_report_spec.rb
afiune: 👍

spec/unit/data_collector/resource_report_spec.rb
afiune: ok

spec/unit/data_collector/resource_report_spec.rb
afiune: you got it

spec/unit/data_collector/resource_report_spec.rb
afiune: why? zelda is awesome!!

spec/unit/data_collector/resource_report_spec.rb
afiune: done

lib/chef/data_collector/resource_report.rb
afiune: changed

afiune:
@adamleff Thanks for your review. I have resolved all your comments. 👍

spec/unit/data_collector/resource_report_spec.rb
adamleff: I was more talking about `l` and `c` - Zelda rules and can stay. :)

lamont-granquist:
seems okay, i'd almost consider broadening the rescue to any standard exception with `rescue` since blowing up in reporting is usually rude.

spec/unit/data_collector/resource_report_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+      it "should return nil" do
+        allow(report).to receive(:elapsed_time).and_return(nil)
+        expect(report.elapsed_time_in_milliseconds).to eq nil
+      end
+    end
+
+    context "when elapsed_time is not set" do
+      it "should return a number" do
+        allow(report).to receive(:elapsed_time).and_return(1)
+        expect(report.elapsed_time_in_milliseconds).to be_kind_of Fixnum
```
lamont-granquist: `.to eql(1000)`?

afiune: @adamleff mentioned that he wanted to test for a number. I'm ok with either one of them.

spec/unit/data_collector/resource_report_spec.rb
afiune: Oh... I didn't understand that, I'll fix them. And Yay for letting Zelda stay!! :tada:

spec/unit/data_collector/resource_report_spec.rb
adamleff: @lamont-granquist is correct... `elapsed_time_in_milliseconds` formats a number, so we should be checking to make sure it's formatting correctly :)

adamleff:
@lamont-granquist's point on broadening to any exception is a cool idea. I would just change the static identity response to something like `unknown identity (exception: #{e.class})` so we (and the user) know why is blew up.

lamont-granquist:
not /any/ exception though, just from StandardError, you should `rescue` and not `rescue Exception`

afiune:
@adamleff @lamont-granquist Ok I think I have applied all your suggestions. 🎉

lamont-granquist:
👍

PR #5942, opened 2017-03-24
Fix cookbook gem installer

brugidou:
Current gem installer does not work if multiple cookbook request the
same gem. The solution is to merge all the gem requirements for the same
gem.

If multiple cookbooks have conflicting constraints, the install will
fail anyway.

This bug affects Chef 12 too and make the feature almost unusable.

### DIFF
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -36,10 +36,12 @@ class Chef
       # Installs the gems into the omnibus gemset.
       #
       def install
-        cookbook_gems = []
+        cookbook_gems = Hash.new { |h, k| h[k] = [] }
 
         cookbook_collection.each do |cookbook_name, cookbook_version|
-          cookbook_gems += cookbook_version.metadata.gems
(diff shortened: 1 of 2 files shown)


brugidou:
Opened #5943 to fix the bundle-audit check

lamont-granquist:
needs a test

brugidou:
@lamont-granquist I added unit tests, I checked that they effectively fail without my patch.

lamont-granquist:
good 'nuff.

note that you did a bunch of work to stub everything and we're not generally that religious about our unit tests being pure unit tests of only the object under test with everything else stubbed.  in particular mocking the filesystem usually ends in tears because there's too much API surface area (read: collaboration) between the object under test and the posix filesystem.  there's mockfs/fakefs solutions that attempt to double-down on unit testing, but i usually prefer to just create real files.  also wouldn't have minded creating a real CookbookCollection instead of stubbing it with a hash.

still way cleaner tests than we usually get, i don't think it really needs to be changed to be accepted, since those are just philosophical points...

PR #5946, opened 2017-03-24
Fix action class weirdness in Chef-13

lamont-granquist:
The old action_class code was doing some magical stuff with the provider
accessor in order to determine if the class was supposed to be a custom
resource or not and have the action_class autovivifying accessor return nil
in cases when the resource wasn't a custom resource and implementing
inheritance by walking back up the tree in ways that were difficult to
grok.

This removes the magic from the provider so that there's no longer any
accessor that magically short-circuits to nil if the resource is not
supposed to be a custom resource.

There is now a simple inherited API for `Chef::Resource.custom_resource?`
which just defines if the class is a custom resource or not.  Since both
`action` and `action_class` call `define_action_class` they both wind up
setting this boolean on the class, which is then inherited to subclasses
automatically, which eliminates the need to walk up the hierarchy.

The superclass.respond_to?(:action_class) checks have also been rendered
unnecessary by removing the code that walked up the inheritance
hierarchy and also because Chef::Resource is never going to be a
custom resource itself, so will never call `define_action_class` so from
inside of `define_action_class` you can always rely on the superclass
being a resource and implementing `custom_resource?` and `action_class`.

The wiring for picking the provider is now moved explicitly to the
ProviderResolver -- even though custom resources hardcode a 1:1
resource-to-provider mapping.  This reads much clearer to me than the
magical wiring to the provider accessor off of the instance.

The bug that this fixes was that the way the magical accessor
nil-or-action_class was implemented the old way of defining action
helpers with class_eval broke:

```ruby
action_class.class_eval << EOM
  def a
    "foo"
  end
EOM
```

If that came before any action_class-with-a-block or action declaration
and the resource did not inherit from another custom resource then the
action_class would not be created and it would return nil, which was an
API which the magical wiring in the provider accessor required.

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -752,8 +752,7 @@ class Chef
               else
                 arg
               end
-      set_or_return(:provider, klass, kind_of: [ Class ]) ||
-        self.class.action_class
+      set_or_return(:provider, klass, kind_of: [ Class ])
     end
 
     def provider=(arg)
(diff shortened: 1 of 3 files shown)


lib/chef/resource.rb
```diff
@@ -1151,13 +1156,11 @@ def self.action_class(&block)
     # Ensure the action class actually gets created. This is called
     # when the user does `action :x do ... end`.
     #
-    # If a block is passed, it is run inside the action_class.
-    #
     # @api private
-    def self.declare_action_class(&block)
+    def self.declare_action_class
       @action_class ||= begin
                           if superclass.respond_to?(:action_class)
```
coderanger: Just for consistency might as well change this too.

coderanger:
Seems like the right direction 👍

lamont-granquist:
github seems to be having webhook weirdness so i'm not sure when this might turn green.

i think its reviewable now.

lamont-granquist:
and paging @chef/client-core for review

lamont-granquist:
Note that I still have no idea exactly what changed in master of Chef-13 to expose this bug.  As far as I can tell it shouldn't have worked on Chef-12 either.

The fix for the bugginess is simply that `def self.action_class` now always creates the action_class and doesn't have the behavior of returning nil if nothing else had caused the action class to come into being (which was then required behavior for the provider accessor to utilize that API in the case where the provider hadn't been set and the resource wasn't a custom resource, so the provider needed to really return nil -- but now that's all gone).

PR #5948, opened 2017-03-26
Add zypper_repository resource

tas50:
The functionality is the same as that of the zypper cookbook and this includes both property aliases and action aliases to ensure compatibility with cookbooks that used the previous format. The properties are more accurately aligned with the names within the configs. Additionally the cookbook relied on shelling out to zypper to add the repositories, which limits the fields that are available and makes determining / updating state hard. This simplified setup works similar to that of the yum_repository resource, which it's based off. It simply writes out a config to the /etc/zypp/repos.d directory and runs the refresh on that repo if necessary.


tas50:
Tested on openSUSE Leap 42.2

tas50:
Tested on SLES 11 SP4

tas50:
Tested on SLES 12 SP2

```
Recipe: test::default
  * zypper_repository[tim] action create
    * template[/etc/zypp/repos.d/tim.repo] action create
      - create new file /etc/zypp/repos.d/tim.repo
      - update content in file /etc/zypp/repos.d/tim.repo from none to fc1a15
      --- /etc/zypp/repos.d/tim.repo	2017-08-09 22:20:23.862965114 +0000
      +++ /etc/zypp/repos.d/.chef-tim20170809-14990-12ry0tl.repo	2017-08-09 22:20:23.862965114 +0000
      @@ -1 +1,15 @@
      +# This file was generated by Chef
      +# Do NOT modify this file by hand.
      +
      +[tim]
      +type=rpm-md
      +enabled=1
      +autorefresh=1
      +gpgcheck=1
      +baseurl=http://smt-ec2.susecloud.net/repo/SUSE/Products/SLE-Module-Adv-Systems-Management/12/x86_64/product?credentials=SMT-http_smt-ec2_susecloud_net
      +priority=99
      +keeppackages=0
      +mode=0644
      +refresh_cache=1
      +name=some repository
      - change mode from '' to '0644'
    * execute[zypper --non-interactive refresh tim] action run
      - execute zypper --non-interactive refresh tim
    * execute[zypper --non-interactive refresh tim] action nothing (skipped due to action :nothing)
```

lib/chef/provider/zypper_repository.rb
```diff
@@ -0,0 +1,79 @@
...
+require "chef/provider/noop"
+
+class Chef
+  class Provider
+    class ZypperRepository < Chef::Provider
+
+      extend Chef::Mixin::Which
+
+      provides :zypper_repository do
+        which "zypper"
```
coderanger: We really need to address this pattern, we do no caching AFAIK so this will run every time if there is more than one resource. Fine for now, but it's going to add up :(

lib/chef/provider/zypper_repository.rb
```diff
@@ -0,0 +1,79 @@
...
+            source ::File.expand_path("../support/zypper_repo.erb", __FILE__)
+            local true
+          end
+          sensitive new_resource.sensitive
+          variables(config: new_resource)
+          mode new_resource.mode
+          notifies :run, "execute[zypper --non-interactive refresh #{new_resource.repo_name}]", :immediately if new_resource.refresh_cache
+        end
+
+        declare_resource(:execute, "zypper --non-interactive refresh #{new_resource.repo_name}") do
```
coderanger: You can avoid some minor repeating if you move this up above and do `refresh = declare...` and then `notifies :run, refresh, :immediately`.

coderanger: Alternatively you could use `notifies :refresh, new_resource, :immediately` and skip the dup'd code entirely.

lib/chef/provider/zypper_repository.rb
```diff
@@ -0,0 +1,79 @@
...
+          action :nothing
+        end
+      end
+
+      action :delete do
+        declare_resource(:execute, "zypper removerepo #{new_resource.repo_name}") do
```
coderanger: It seems a little weird to involve resources here instead of just using `shell_out*!` directly.

lib/chef/provider/zypper_repository.rb
```diff
@@ -0,0 +1,79 @@
...
+        declare_resource(:execute, "zypper refresh #{new_resource.repo_name}") do
+          only_if "zypper lr #{new_resource.repo_name}"
+        end
+      end
+
+      alias_method :action_add, :action_create
+      alias_method :action_remove, :action_delete
+
+      def template_available?(path)
+        !path.nil? && run_context.has_template_in_cookbook?(new_resource.cookbook_name, path)
```
coderanger: I don't think this is good, if a `source` is given, it shouldn't be silently ignored just because it doesn't exist.

lib/chef/provider/zypper_repository.rb
tas50: As far as package providers go it seems like exposing the data in ohai and consuming it here would probably work

lib/chef/provider/zypper_repository.rb
tas50: That's straight out of the yum_repository provider

lib/chef/provider/zypper_repository.rb
coderanger: Or just work out a shared cache like we do for detecting which service tools are installed.

lib/chef/provider/zypper_repository.rb
tas50: It seemed like the easiest way to handle notification and the if statement that has to occur. I could double shell out there, but that seems a bit more complex than just letting the core resources do what they do best

coderanger: That's fair, the main risk factor here is the use of string execution instead of array, but probably low impact in this particular case. If you wanted to be extra fancy you could `Shellwords.escape` the repo name everywhere.

lib/chef/provider/zypper_repository.rb
coderanger: Sigh, a fight for another day then :)

tas50:
With spaces:

```
  * zypper_repository[spaces are evil] action create
    * template[/etc/zypp/repos.d/spaces\ are\ evil.repo] action create
      - create new file /etc/zypp/repos.d/spaces\ are\ evil.repo
      - update content in file /etc/zypp/repos.d/spaces\ are\ evil.repo from none to 3158ca
      --- /etc/zypp/repos.d/spaces\ are\ evil.repo	2017-08-09 23:21:05.870965114 +0000
      +++ /etc/zypp/repos.d/.chef-spaces\ are\ evil20170809-16321-ofwr17.repo	2017-08-09 23:21:05.870965114 +0000
      @@ -1 +1,15 @@
      +# This file was generated by Chef
      +# Do NOT modify this file by hand.
      +
      +[spaces are evil]
      +type=rpm-md
      +enabled=1
      +autorefresh=1
      +gpgcheck=1
      +baseurl=http://smt-ec2.susecloud.net/repo/SUSE/Products/SLE-Module-Adv-Systems-Management/12/x86_64/product?credentials=SMT-http_smt-ec2_susecloud_net
      +priority=99
      +keeppackages=0
      +mode=0644
      +refresh_cache=1
      +name=some repository
      - change mode from '' to '0644'
    * zypper_repository[spaces are evil] action refresh
      * execute[zypper refresh spaces\ are\ evil] action run
        - execute zypper refresh spaces\ are\ evil
```

PR #5949, opened 2017-03-26
Swap Ubuntu 12.04 testing for openSUSE Leap

tas50:
Swap Ubuntu 12.04 testing for openSUSE Leap

    Ubuntu 12.04 is going end of life so lets swap testing for another OS we should check.

    While we’re in here lets update a few other things
    -  There’s no longer a reason to hold back dokken to the 1.0 release.
    - Remove a dupe travis config
    - Remove centos 5 from the kitchen config since it’s going into the mostly EOL phase
    - Update platforms in the local kitchen config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ before_script:
 branches:
   only:
   - master
-  - 11-stable
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -193,7 +174,6 @@ matrix:
       - cd kitchen-tests
     script:
       - bundle exec kitchen test webapp-debian-8
-    bundler_args: --without ci development docgen guard integration maintenance omnibus_package tools aix bsd mac_os_x solaris windows --frozen
```
lamont-granquist: that line of red looks wrong?

kitchen-tests/.kitchen.travis.yml
```diff
@@ -43,16 +42,6 @@ platforms:
       - RUN /usr/bin/apt-get update
       - RUN /usr/bin/apt-get -y install zlib1g-dev sudo net-tools wget ca-certificates
 
-- name: centos-5
```
thommay: definitely we should not remove that.

kitchen-tests/.kitchen.travis.yml
```diff
@@ -2,7 +2,6 @@
 driver:
   name: dokken
   privileged: true
-  chef_image: chef/chef
```
lamont-granquist: what was this there for?  beware that i had to do some hacky shit to make dokken work with chef_github provisioner, and applying standardized dokken config to this file may cause issues...

kitchen-tests/.kitchen.travis.yml
tas50: We weren't actually using this in the config. It was just defined in the Kitchen yaml.

kitchen-tests/.kitchen.travis.yml
tas50: Old kitchen-dokken uses sean's image. The last pile of releases have defaulted to our images so this config becomes redundant.

tas50:
I'll open a PR for just the opensuse changes later. Most of this was just merged anyways in a diff PR

PR #5952, opened 2017-03-28
Chef-13: Make ActionClass a class

lamont-granquist:
since any ActionClass literally is-a Chef::Provider it should just
use inheritance and then we don't have to use the hacky hooks for
class methods in modules and it becomes more clear from the
ActionClass source what it is and what it does.

~~Bonus starts the deprecation of properties being injected into the
provider namespace.~~ Not yet.

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -267,15 +267,15 @@ class Chef
     # @param include_resource_dsl [Boolean] Whether to include resource DSL or
     #   not (defaults to `false`).
     #
-    def self.include_resource_dsl(include_resource_dsl)
-      @include_resource_dsl = include_resource_dsl
+    def self.include_resource_dsl?
+      false
     end
 
     # Create the resource DSL module that forwards resource methods to new_resource
     #
     # @api private
     def self.include_resource_dsl_module(resource)
-      if @include_resource_dsl && !defined?(@included_resource_dsl_module)
+      if include_resource_dsl? && !defined?(@included_resource_dsl_module)
         provider_class = self
         @included_resource_dsl_module = Module.new do
           extend Forwardable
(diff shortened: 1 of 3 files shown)


lamont-granquist:
are you sure we didn't already have that weirdness?  because the parent resource then wouldn't have an action class so you take Chef::Provider as your base class and mixin the ActionClass and lose all the actions on the provider associated with the resource you were subclassing.

and solving that problem in the general case would be impossible because of 1:N resources like service where you have to go in through the provider resolver in order to even find which provider you should use as the base to subclass...

but yeah, i'd just say we should care about that because its not sane and we can't identify the correct provider to subclass...

and i'm not even certain anyone is subclassing custom resources...

lib/chef/provider.rb
```diff
@@ -286,10 +286,27 @@ def self.include_resource_dsl_module(resource)
           resource.class.properties.each do |name, property|
             class_eval(<<-EOM, __FILE__, __LINE__)
               def #{name}(*args, &block)
+                # not making this a formal deprecation yet because I don't want to drive tim completely crazy, this
+                # should be upgraded to a real deprecation in 13.1 or so
+                Chef::Log.warn "Deprecated property access: please replace calling #{name} with new_resource.#{name}"
```
coderanger: Thinking more on this, I'm not sure I love this. Deprecation warnings are automatically de-dup'd to avoid spamming the logs, but this won't be.

lamont-granquist: @tas50 what do you think?  if we throw this in as a real honest deprecation warning then if you run with deprecations-are-errors like you like to you'll fail and have to fix all these...

the other alternative is to just not care about it for now and to make sure to remember later.

lamont-granquist:
I do actually see now what you lose by doing this, which is that you can't mixin ActionClass manually into an old school provider in the providers DSL, and would have to move that provider to a library and subclass and lets just not support that use case, because blarghle...

tas50:
Ideally we don't introduce any "real" deprecations right now since that makes the 13.0 migration work that much harder. I entirely support killing this feature with fire though. I've been burned by it and it wasted a lot of my time. I love sugar, but this isn't helpful sugar. The majority of the time it just leads to user pain and we should remove it to prevent that pain.

lamont-granquist:
Sounds like we should just add a real deprecation for 13.1 though.

tas50:
Drop as many deprecations into 13.1 as you want

lamont-granquist:
github seems to be having webhook issues again, appveyor is green even though github is showing it still pending...

PR #5958, opened 2017-03-29
disable nokogiri on windows for now

lamont-granquist:
nokogiri does not have a ruby-2.4 release for windows

this is slightly complicated because we still need nokogiri in the
Gemfile.lock for travis.

i'm starting to think we should have a Gemfile.travis and
Gemfile.travis.lock or something and stop using groups for that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,8 +22,8 @@ gem "cheffish" # required for rspec tests
 group(:omnibus_package) do
   gem "appbundler"
   gem "rb-readline"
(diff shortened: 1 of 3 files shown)


lamont-granquist:
well this is necessary to unbreak builds, but now foodcritic is in our gemfile which pull in nokogiri and if we have to work on 2.4 then it slurps in an ancient nokogiri version, and i gave up on solving this problem for today...

PR #5960, opened 2017-03-29
Retry API requests if using an unsupported version

thommay:
If we've not made a request to the server previously, we start by
trying the best possible version we know about; this means that in the case of a fresh connection to an old server we in general should only make two requests at worst.

cc @chef/maintainers

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -142,13 +142,26 @@ class Chef
     # Makes an HTTP request to +path+ with the given +method+, +headers+, and
     # +data+ (if applicable).
     def request(method, path, headers = {}, data = false)
+      http_attempts ||= 0
       url = create_url(path)
       method, url, headers, data = apply_request_middleware(method, url, headers, data)
 
       response, rest_request, return_value = send_http_request(method, url, headers, data)
       response, rest_request, return_value = apply_response_middleware(response, rest_request, return_value)
+
       response.error! unless success_response?(response)
       return_value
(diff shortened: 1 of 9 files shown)


lib/chef/http.rb
```diff
@@ -142,13 +142,26 @@ def delete(path, headers = {})
...
       url = create_url(path)
       method, url, headers, data = apply_request_middleware(method, url, headers, data)
 
       response, rest_request, return_value = send_http_request(method, url, headers, data)
       response, rest_request, return_value = apply_response_middleware(response, rest_request, return_value)
+
       response.error! unless success_response?(response)
       return_value
+
+    rescue Net::HTTPServerException => e
```
thommay: We have to do this here (and in the streaming requests) rather than in `retrying_http_errors` because we need to re-process all the middleware each time.

lib/chef/http.rb
```diff
@@ -177,6 +191,16 @@ def streaming_request_with_progress(path, headers = {}, &progress_block)
         response.error!
       end
       tempfile
+    rescue Net::HTTPServerException => e
+      http_attempts += 1
+      response = e.response
+      if response.kind_of?(Net::HTTPNotAcceptable) && version_retries - http_attempts > 0
```
srenatus: There's no way to rescue that _directly_, I suppose?

thommay: Nope!

PR #5964, opened 2017-03-29
fix ruby-2.4 func link tests on solaris 11

lamont-granquist:
ruby-2.4 on solaris 11 behaves the same way here

i suspect linux is the buggy one and everyone else is posix
compliant...

### DIFF
--- a/spec/functional/resource/link_spec.rb
+++ b/spec/functional/resource/link_spec.rb
@@ -654,7 +654,7 @@ describe Chef::Resource::Link do
           end
           context "and the link does not yet exist" do
             it "links to the target file" do
-              skip("OS X/FreeBSD/AIX symlink? and readlink working on hard links to symlinks") if os_x? || freebsd? || aix? || solaris?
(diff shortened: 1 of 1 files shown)


spec/functional/resource/link_spec.rb
```diff
@@ -654,7 +654,7 @@ def parent_inheritable_acls
           end
           context "and the link does not yet exist" do
             it "links to the target file" do
-              skip("OS X/FreeBSD/AIX symlink? and readlink working on hard links to symlinks") if os_x? || freebsd? || aix?
+              skip("OS X/FreeBSD/AIX symlink? and readlink working on hard links to symlinks") if os_x? || freebsd? || aix? || solaris?
```
shortdudey123: add `solaris` to the message itself too?

lamont-granquist: sure but the whole premise is still likely misleading...

shortdudey123: fair point

Issue #5931, opened 2017-03-21
Missing output from systemd-analyze makes troubleshooting impossibly hard


axelrtgs:
## Description

Missing output from systemd-analyze makes troubleshooting impossibly hard

## Chef Version

chef-client 12.19.36

## Platform Version

Ubuntu 16.04

## Replication Case

create a systemd unit with invalid data, systemd-analyze will fail but chef will only indicate failure with no help of where, the systemd-analyze output would be very helpful for troubleshooting issues as it contains the exact errors.

## Client Output

ERROR: systemd_unit[app.service] (WebApps::app line 85) had an error: Chef::Exceptions::ValidationFailed: file[/etc/systemd/user/app.service] (dynamically defined) had an error: Chef::Exceptions::ValidationFailed: Proposed content for /etc/systemd/user/app.service failed verification #<Chef::Resource::File::Verification:0x00000004f19fb0>

stresler:
This is exactly what I hit. I don't think this is a "Help Wanted" but a feature request. 

The systemd-analyze output is available it is just not making it into the chef output. My use case I described in the PR conversation here: https://github.com/chef/chef/pull/5765#issuecomment-288579727

Where it was indicated this should be a new issue.

Issue #5909, opened 2017-03-16
systemd verify fails for socket units


tpodom:
## Description

Trying to create a systemd unit for a socket fails because the related service doesn't exist yet.  I believe this is related to commit aea6988.  When I run systemd-analyze manually it outputs "Service rt.service not loaded, rt.socket cannot be started." because the rt.service is created separately.  Is there a different way to create a socket + service?

## Chef Version

12.19.36 CentOS omnibus

## Platform Version

CentOS 7.3

## Replication Case

```
    systemd_unit 'rt.socket' do
      enabled true
      content({
        'Unit' => {
          'Description' => 'Request Tracker Listen Socket'
        },
        'Socket' => {
          'ListenStream' => "0.0.0.0:9001"
        },
        'Install' => {
          'WantedBy' => 'sockets.target'
        }
      })
      action :create
    end

    systemd_unit 'rt.service' do
      enabled true
      content({
        'Unit' => {
          'Description' => 'Request Tracker Webapp',
          'After' => 'network.target',
          'Requires' => 'rt.socket'
        },
        'Service' => {
          'User' => 'rt',
          'Group' => 'rt',
          'StandardOutput' => 'null',
          'StandardError' => 'syslog',
          'ExecStart' => "/usr/local/bin/systemd-spawn-fcgi.sh /usr/local/rt/sbin/rt-server.fcgi -initial-env RT_SITE_CONFIG=/usr/local/rt/etc/RT_SiteConfig.pm",
          'Restart' => 'always'
        },
        'Install' => {
          'WantedBy' => 'multi-user.target'
        }
      })
      action :create
    end
```

## Client Output
```
==> linux:       * file[/etc/systemd/system/rt.socket] action create
==> linux: 
==> linux:         
==> linux:         ================================================================================
==> linux:         Error executing action `create` on resource 'file[/etc/systemd/system/rt.socket]'
==> linux:         ================================================================================
==> linux:         
==> linux:         Chef::Exceptions::ValidationFailed
==> linux:         ----------------------------------
==> linux:         
==> linux: Proposed content for /etc/systemd/system/rt.socket failed verification #<Chef::Resource::File::Verification:0x000000041f5aa0>
==> linux:         
==> linux:         
==> linux: Compiled Resource:
==> linux:         ------------------
==> linux:         # Declared in 
==> linux:         
==> linux:         file("/etc/systemd/system/rt.socket") do
==> linux:           action [:create]
==> linux:         
==> linux:   retries 0
[...]

PR #5971, opened 2017-03-30
systemd_unit verifier escape hatch

nathwill:
systemd's pretty strict with the verification rules. it's usually possible to make it work, but this adds an escape hatch for if the verifier is too strict, or for use in debugging, since verifier output isn't surfaced right now.

### Description

permits users to opt out of resource verification

### Issues Resolved

helps with situations like #5931 and #5909. a full fix for #5931 that actually surfaces err output is incoming, but as a separate PR (#5972) since it touches more than just the systemd_unit resource.

### Check List

- [x] New functionality includes tests (N/A)
- [x] All tests pass
- [x] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes) (N/A)

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -194,7 +194,7 @@ class Chef
           f.group "root"
           f.mode "0644"
           f.content new_resource.to_ini
-          f.verify :systemd_unit
+          f.verify :systemd_unit if new_resource.verify
         end.run_action(action)
       end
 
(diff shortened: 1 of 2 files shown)


mal:
lgtm

coderanger:
I thought we made it an error to shadow core methods using `property`?

nathwill:
> shadow core methods

what core methods? i thought #verify was just a file resource property?

i'm fine to rename it, but honestly not sure where the problem is supposed to stem from here.

lamont-granquist:
yeah, verify isn't a core property, its on the file resource.

@coderanger we actually allow overriding properties, we don't allow clobbering non-property methods.  in some cases we deliberately subclass and override properties....  not happening here though.

farvour:
This feature does not seem available in 12.21? The documentation doesn't make this clear either. I think the documentation should be updated if this is a Chef 13+ feature.

https://docs.chef.io/resource_systemd_unit.html#properties

PR #5972, opened 2017-03-30
surface default guard interpreter errors

### LINKED ISSUE #5931: Missing output from systemd-analyze makes troubleshooting impossibly hard

nathwill:
surface command stderr when guard interpreter command is unsuccessful. helps with showing why file/systemd_unit verify commands may fail.

surfacing the error in the converge output is important for debugging since the temporary files used during verification are not available for inspection after the verification command is run.

### Description

aids in debugging when guard interpreter fails

### Issues Resolved

#5931 

### Check List

- [x] New functionality includes tests (N/A)
- [x] All tests pass (N/A)
- [x] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes) (N/A)

### DIFF
--- a/lib/chef/guard_interpreter/default_guard_interpreter.rb
+++ b/lib/chef/guard_interpreter/default_guard_interpreter.rb
@@ -33,7 +33,9 @@ class Chef
       public
 
       def evaluate
-        shell_out_with_systems_locale(@command, @command_opts).status.success?
+        result = shell_out_with_systems_locale(@command, @command_opts)
+        Chef::Log.debug "Command failed: #{result.stderr}" unless result.status.success?
+        result.status.success?
       # Timeout fails command rather than chef-client run, see:
       #   https://tickets.opscode.com/browse/CHEF-2690
       rescue Chef::Exceptions::CommandTimeout


mal:
Looks good.

Part of me wondered if it'd benefit from the same sort of error handling as the `execute` resource does; dumping both `stderr` and `stdout`. Unfortunately a quick look at the code there suggests it could be a bit of a rabbit hole with regard to handling of sensitive output. So it probably doesn't make sense to do as part of this solution, more something to consider if an issue in that area gets raised.

nathwill:
> Part of me wondered if it'd benefit from the same sort of error handling as the execute resource does; dumping both stderr and stdout

yeah, i thought about dumping stdout also, but ultimately decided that since we know that we're logging an error condition, failure to write error data to STDERR is an app problem, rather than something we should worry about.

> it could be a bit of a rabbit hole with regard to handling of sensitive output

good point! i hadn't thought of that... fortunately i don't think the overlap between 'sticking sensitive data in a unit file' and 'sensitive data in unit file causing error' would be very common?

my recommendation is always to put the sensitive data in an environment file and interpolate the environment variable into the unit if it's needed in the unit definition (it's pretty convenient that systemd supports this). i won't even use `Environment=` anymore, as `EnvironmentFile=` is a clearer separation of concerns that tends to be more maintainable in my experience (and lets me suppress output of just the resource with the sensitive data).

mal:
WRT the sensitivity, I agree it probably doesn't apply to systemd units too much if at all, but I was thinking more in the broader non-systemd sense as this will be handling all verification failures if I understand correctly, where it could just be any arbitrary command. Probably worth waiting for a real-world use case before thinking about it too hard though.

lib/chef/guard_interpreter/default_guard_interpreter.rb
```diff
@@ -33,7 +33,9 @@ def initialize(command, opts)
       public
 
       def evaluate
-        shell_out(@command, @command_opts).status.success?
+        result = shell_out(@command, @command_opts)
+        Chef::Log.warn "Command failed: #{result.stderr}" unless result.status.success?
```
lamont-granquist: but "failure" here in terms of exit != 0 is a normal condition though isn't it so we shouldn't warn?  we could debug log the stderr and stdout and it would seem to be a decent idea to actually always do that.

its also odd that we rescue a timeout below and then swallow the exception and return false.  if we did that for a reason it should probably be noted in a comment as to why that got done, otherwise i'd think that any exception that raised by shell_out should propagate upwards and kill the chef run.

lamont-granquist:
implementing sensitive will be somewhat annoying because the DefaultGuardInterpreter doesn't get the resource passed to it so can't read sensitive off.   that should really get changed.

lib/chef/guard_interpreter/default_guard_interpreter.rb
nathwill: oh good point re: failure being a normal (or at least expected) condition... i'll update this to debug.

coderanger:
This is an improvement but still not the UX I would really love to see, this should get surfaced as part of the exception/error display probably. But having anything at all is 👍 since right now you are just SoL

nathwill:
rebased, squashed

PR #5977, opened 2017-03-30
apply knife search node fuzzifier to knife ssh

### LINKED ISSUE #3526: [knife] Assume a default key when no colon in QUERY

lamont-granquist:
extracts out a common 'fuzz' option to Chef::Search::Query so that anyone can
use it.

closes #3526

### DIFF
--- a/lib/chef/knife/search.rb
+++ b/lib/chef/knife/search.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright 2009-2016, Chef Software Inc.
+# Copyright:: Copyright 2009-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


lamont-granquist:
working on tests...

lib/chef/knife/search.rb
```diff
@@ -73,19 +73,18 @@ class Search < Knife
...
       def run
         read_cli_args
-        fuzzify_query
 
         if @type == "node"
           ui.use_presenter Knife::Core::NodePresenter
         end
 
         q = Chef::Search::Query.new
-        escaped_query = Addressable::URI.encode_component(@query, Addressable::URI::CharacterClasses::QUERY)
```
lamont-granquist: it looks like this was fixed here first (nuo yan era fix), then fixed correctly in Chef::Search::Query but was never removed here.

PR #5984, opened 2017-03-31
Chef-13: fix notifying array resources

### LINKED ISSUE #4349: Notify multiple packages
I'm trying to notify a resource that contains multiple packages as below

```
package ["chef", "chef-zero"] do
    action :nothing
end

log 'call a notification' do
  notifies :install, 'package[chef, chef-zero]', :immediately
end
```

The above recipe is in line with documentation at https://docs.chef.io/resource_yum.html (Multiple Packages). Output of a local run is below. Kindly let me know if this is a bug or a user error?

```
Starting Chef Client, version 11.12.8
resolving cookbooks for run list: []
Synchronizing Cookbooks:
Compiling Cookbooks...
[2015-12-30T23:18:57+05:30] WARN: Node derek-laptop has an empty run list.
Converging 2 resources

Running handlers:
[2015-12-30T23:18:57+05:30] ERROR: Running exception handlers
Running handlers complete

[2015-12-30T23:18:57+05:30] ERROR: Exception handlers complete
[2015-12-30T23:18:57+05:30] FATAL: Stacktrace dumped to /home/derek/.chef/local-mode-cache/cache/chef-stacktrace.out
Chef Client failed. 0 resources updated in 0.67967665 seconds
[2015-12-30T23:18:57+05:30] ERROR: resource log[call a notification] is configured to notify resource package[chef, chef-zero] with action install, but package[chef, chef-zero] cannot be found in the resource collection. log[call a notification] is defined in /home/derek/abc.rb:5:in `from_file'

[2015-12-30T23:18:57+05:30] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
derek@derek-laptop:~$ cat abc.rb 
package ["chef", "chef-zero"] do
    action :nothing
end

log 'call a notification' do
  notifies :install, 'package[chef, chef-zero]', :immediately
end
```

### LINKED ISSUE #5976: Chef 13.1: Deprecate old multi-resource notification syntax
Turns out notifying `package[foo, bar]` means to try to find both `package[foo]` and `package[bar]` in the resource_collection, which gets into a fight with multipackage syntax.

AFAIK this has never been documented before, and dates back 9 years to the first ever commit to the resource collection.

The way everyone does this is with multiple notifies hanging off the same resource, or with a notification to a log resource (we really need a null resource)

We probably need to deprecate this to make notifications to multiple package installs to work.

lamont-granquist:
this also deprecates the multi-resource notification syntax where
`foo[a,b]` would attempt to notify both `foo[a]` and `foo[b]` because
that is hella ambiguous (and AFAIK undocumented for the past 9 years)

closes #5976
closes #4349


lib/chef/resource_collection/resource_set.rb
```diff
@@ -161,27 +161,39 @@ def find_resource_by_hash(arg)
...
+            return [ lookup(create_key(resource_type, name)) ]
+          end
+        rescue Chef::Exceptions::ResourceNotFound => e
+          if arg =~ MULTIPLE_RESOURCE_MATCH
+            begin
+              results = Array.new
+              resource_type = $1
+              arg =~ /^.+\[(.+)\]$/
+              resource_list = $1
+              resource_list.split(",").each do |instance_name|
```
coderanger: Can do this with less global variables via `arg[/^.+\[(.+)\]$/, 1].split(',')` but doesn't really matter.

lamont-granquist: yeah its just copypasta from what was there and imo it needs to die

lamont-granquist:
hrm, poise has a busted test...

PR #5985, opened 2017-03-31
Chef-13: remove deprecated run_command API entirely

lamont-granquist:
I can't believe I nearly missed this one.


lamont-granquist:
sigh, the cron resource...

lamont-granquist:
finally green with manhattan green (as it can get) as well.

PR #5986, opened 2017-03-31
Add expeditor

tduffield:
### Description

Expeditor is the next-generation utility that we'll use for version bumping. It will include additional features that we don't currently have like the ability to not bump a version on a specific PR.

### Check List

- [ ] All tests pass

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -1,0 +1,31 @@
+# The name of the product key for this product from mixlib-install
+product_key: chef
+
+# Slack channel in Chef Software slack to send notifications about build failures, etc
+slack:
+  notify_channel: chef-notify
+
+# When a version of ChefDK hits the current channel, build a corresponding Docker image
+# and publish that image to https://hub.docker.com/r/chef/chefdk
+docker:
+  enable: true
+  build_args:
+    CHANNEL: "{{channel}}"
+    VERSION: "{{version}}"
+
+github:
+  # The file where the MAJOR.MINOR.PATCH version is kept
+  version_file: "VERSION"
+
+  # When a PR is merged, bump the PATCH version
+  bump_version_on_merge: true
+
+  # After the PATCH version has been bumped, execute this script
+  # to distribute that version to other files in the repository.
+  update_version_script: ".expeditor/update_version.sh"
+
+  # The tag format Expeditor should use when tagging version commits
+  version_tag_format: "v{{version}}"
+
+  # After the version is bumped and the tag is pushed to Github, trigger a Jenkins build
+  trigger_build_on_bump: true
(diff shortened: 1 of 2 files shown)


.expeditor/bundle_install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -evx
+
+gem environment
+bundler_version=$(grep bundler omnibus_overrides.rb | cut -d'"' -f2)
+gem install bundler -v $bundler_version --user-install --conservative
+# WITH: ci (for version bumping and changelog creation)
+export BUNDLE_WITHOUT=omnibus_package:test:pry:integration:docgen:maintenance:travis:aix:bsd:linux:mac_os_x:solaris:windows:development
+bundle _${bundler_version}_ install
```
lamont-granquist: just do `bundle install` here -- you've already asserted the bundler version you're going to be running is the correct one, and this seems to get into a fight with the 1.14.x trampolining code.

.expeditor/bundle_install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -evx
+
+gem environment
+bundler_version=$(grep bundler omnibus_overrides.rb | cut -d'"' -f2)
```
lamont-granquist: this should probably be renamed because bundler uses `ENV['BUNDLER_VERSON']` now to setup trampolining, but that code is buggy and we don't want to use it, at at a minimum this is going to cause confusion to humans later.

tduffield: so use a variable name other than bundler_version?

tasks/version.rb
```diff
@@ -39,3 +39,27 @@
     end
   end
 end
+
+task :expeditor_update_version do
```
lamont-granquist: i don't know if this is a stub, but it doesn't seem to be useful, and it seems like the caller should be responsible for swallowing/ignoring exceptions directly.

tduffield: When you say "this", what do you mean?

.expeditor/bundle_install.sh
lamont-granquist: the appveyor config used `$env:BUNDLER_VERSION=` which actively failed on bundler 1.14.x and i changed that to `$env:OMNIBUS_BUNDER=` here:

https://github.com/chef/chef/commit/a3159a69242adc94050e1664a7c1ced9d08a84a6#diff-180360612c6b8c4ed830919bbb4dd459

tasks/version.rb
lamont-granquist: the whole block of green code here.  if expeditor wants to call the rake tasks and ignore errors it should just call the rake tasks and ignore errors, rather than building additional rake tasks that ignore errors...

.expeditor/bundle_install.sh
lamont-granquist: yeah just rename the variable

tduffield:
@coderanger I've applied many of the same principles we spoke about in chef/chef-dk#1213 to this PR as well. 

@lamont-granquist - Let me know if this new PR addresses your feedback :)

PR #5987, opened 2017-04-01
Expand the system info displayed on error to give us some more data to work with when helping users

coderanger:
An example of the new block:

```
System Info:
------------
chef_version=13.0.0
platform=ubuntu
platform_version=16.04
ruby=ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
program_name=chef-client
executable=/opt/chef/bin/chef-client
```

The only odd ones that might need explanation are program_name which is the straight $0 and executable which is the un-symlinked path of the actual ruby-level program running.

### DIFF
--- a/lib/chef/formatters/error_description.rb
+++ b/lib/chef/formatters/error_description.rb
@@ -17,8 +17,7 @@
 # limitations under the License.
 #
 
-require 'chef/version'
-
+require "chef/version"
 
 class Chef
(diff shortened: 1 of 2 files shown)


coderanger:
Travis failure looks unrelated, guessing it is fallout from the Great Amazon Linux Swap, merging :)

PR #5989, opened 2017-04-02
Chef-13: remove deprecated knife subcommand APIs

lamont-granquist:


lamont-granquist:
i suspect there will be test failures...

lamont-granquist:
amazeballs there are not...

someone tell me if i'm doing this right though because i'm literally just mechanically deleting deprecated code without any idea...

lamont-granquist:
No Idea Dawg

stevendanna:
I think that we should also be able to nuke: https://github.com/chef/chef/blob/bc42427577b2874dfd1a5cf6bc65dece2c6bdec1/lib/chef/knife/core/custom_manifest_loader.rb

stevendanna:
This function could potentially move into the GemGlobLoader rather than living in the base SubcommandLoader class: 

```
lib/chef/knife/core/subcommand_loader.rb
125:      def find_subcommands_via_dirglob
```

lamont-granquist:
yeah, i've only got 4 more working days to play with Chef-13 so I think i'll just nuke that file I forgot to remove and call it good.  rest of it i think we either figure out how to do it non-breaking or it waits till Chef-14.

PR #5990, opened 2017-04-02
Chef-13: turn on zypper gpg checks by default

### LINKED ISSUE #2339: CHEF-4058: zypper package provider should not ignore gpg checks by default
we should not be doing this by default:

> ```
>     * package[nfs-utils] action install[2014-11-02T18:45:35+00:00] WARN: Chef::Config[:zypper_check_gpg] was not set. All packages will be installed without gpg signature checks. This is a security hazard.
> ```

lamont-granquist:
Adds a gpg_check property to the resource to turn them off on a
per-resource basis.  The global config is also preserved.  The
global config will warn once if you have it set to turn off
gpg checks.  The per-resource property will not warn (presumably
you know you're doing something bad when you turn it off).

closes #2339

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -644,11 +644,10 @@ module ChefConfig
     # generation (server generates client keys).
     default(:local_key_generation) { true }
 
-    # Zypper package provider gpg checks. Set to true to enable package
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core more reviewage

PR #5991, opened 2017-04-02
Chef-13: coerce package options property to an Array

### LINKED ISSUE #5763: corece the options property in the package resource base class to an Array
this will break chefspec matchers on the option property, but will clean up the code a lot.

lamont-granquist:
Windows side we still can't support this kind of API, this
is for Unix-ey stuff.

closes #5763

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -52,11 +52,7 @@ class Chef
       end
 
       def options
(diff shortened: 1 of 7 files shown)


lamont-granquist:
@chef/client-core reviewage

lamont-granquist:
derp, yes, i was wrangling with the String case on the windows side and suffered some bit corruption in the title...

PR #5995, opened 2017-04-03
Chef-13:  tweaks to rubygems source option for urls

lamont-granquist:
By default now we don't pass any --source option or set the
`Chef::Config[:rubygems_url]` to anything.  This is so that
if there is external config for rubygems sources we will simply
have rubygems pick it up and use it (the resource acts like
the command line).

There is also no more magic around `--clear-sources` and if you
set the property it sets the flag.

If the `Chef::Config[:rubygems_url]` setting is set then that
becomes the default `--source` flag sent to the command.  This
also supports Arrays now.

If the `source` option is set to a URI (it can also be set to a
filesystem path for poorly-designed-API reasons) then we do not
use the default values at all.  This may be an API break but this
much more clearly allows users to override individual resources
and is more declarative.  You get what you type.

### DIFF
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -1,5 +1,5 @@
 #--
-# Copyright:: Copyright (c) 2010-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2010-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -50,7 +50,9 @@ class Chef
           begin
             Dir.mktmpdir("chef-gem-bundle") do |dir|
               File.open("#{dir}/Gemfile", "w+") do |tf|
-                tf.puts "source '#{Chef::Config[:rubygems_url]}'"
+                Array(Chef::Config[:rubygems_url] || "https://www.rubygems.org").each do |s|
+                  tf.puts "source '#{s}'"
+                end
                 cookbook_gems.each do |gem_name, args|
                   tf.puts "gem(*#{([gem_name] + args).inspect})"
                 end
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -533,18 +533,21 @@ class Chef
         end
 
         def install_via_gem_command(name, version)
-          if new_resource.source =~ /\.gem$/i
+          src = []
+          if new_resource.source.is_a?(String) && new_resource.source =~ /\.gem$/i
             name = new_resource.source
-          elsif new_resource.clear_sources
-            src = " --clear-sources"
-            src << (new_resource.source && " --source=#{new_resource.source}" || "")
           else
-            src = new_resource.source && " --source=#{new_resource.source} --source=#{Chef::Config[:rubygems_url]}"
+            src << "--clear-sources" if new_resource.clear_sources
+            srcarry = [ new_resource.source || Chef::Config[:rubygems_url] ].flatten.compact
+            srcarry.each do |s|
+              src << "--source=#{s}"
+            end
           end
+          src_str = src.empty? ? "" : " #{src.join(" ")}"
           if !version.nil? && !version.empty?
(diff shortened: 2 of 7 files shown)


lamont-granquist:
tests incoming, but i'm not sure if what i'm doing with setting the rubygems_url to nil and just letting rubygems do its default behavior is a bridge too far here.  i think it has defaulted to https://www.rubygems.org for a very long time though...

lamont-granquist:
looks like `https://www.rubygems.org` has been the default since at least rubygems 1.4.0 some 6.5-ish years ago:

https://github.com/rubygems/rubygems/commit/2799f27e6febf5df71626fdd9758ccc7c3d4a7c1

lamont-granquist:
and per slack conversation http to www.rubygems.org has a forced redirect to https, so even if someone has their external environment setup to use "http://rubygems.org/" it'll still redirect to https securely.

only thing this breaks is people who have their external environment setup to an internal mirror, but who want chef to default to rubygems.org -- and i'm okay with breaking those people in order to by default support the people who have their external environment setup to mirror who want chef to hit the mirror by default.

lib/chef/cookbook/gem_installer.rb
```diff
@@ -48,7 +48,13 @@ def install
           begin
             Dir.mktmpdir("chef-gem-bundle") do |dir|
               File.open("#{dir}/Gemfile", "w+") do |tf|
-                tf.puts "source '#{Chef::Config[:rubygems_url]}'"
+                if Chef::Config[:rubygems_url].empty?
```
coderanger: Doesn't this also need a nil check?

lib/chef/cookbook/gem_installer.rb
```diff
@@ -48,7 +48,13 @@ def install
...
+                  tf.puts "source 'https://www.rubygems.org'"
+                else
+                  [ Chef::Config[:rubygems_url] ].flatten.compact.each do |s|
```
coderanger: The flatten-and-compact dance is probably not worth the CPU time here, `Array(Chef::Config[:rubygems_url] || 'https://www.rubygems.org').each` handles both conditions?

lib/chef/provider/package/rubygems.rb
```diff
@@ -533,18 +533,21 @@ def gem_binary_path
...
             name = new_resource.source
-          elsif new_resource.clear_sources
-            src = " --clear-sources"
-            src << (new_resource.source && " --source=#{new_resource.source}" || "")
           else
-            src = new_resource.source && " --source=#{new_resource.source} --source=#{Chef::Config[:rubygems_url]}"
+            src << "--clear-sources" if new_resource.clear_sources
+            srcarry = [ new_resource.source || Chef::Config[:rubygems_url] ].flatten.compact
+            srcarry.each do |s|
+              src << "--source=#{s}"
```
coderanger: Should technically use `Shellwords.escape` here on Unix but I don't want to think of how that would break on Windows or how terrible this would be in general with metachars in a source URL. Can heap this on the "should be using array form of shellout" debt pile.

lamont-granquist: yeah hopefully i want to lay some better groundwork this week in Chef-13 for array form of shell_out

lib/chef/cookbook/gem_installer.rb
lamont-granquist: i suspect i will find travis is annoyed about that tomorrow morning...

lib/chef/cookbook/gem_installer.rb
lamont-granquist: yeah most likely.

lamont-granquist:
FFS the multiplexing in the meaning of the "source" property is just aggravating...  words really need to have single meanings...

lib/chef/provider/package/rubygems.rb
```diff
@@ -469,7 +470,8 @@ def all_installed_versions
         end
 
         def gem_sources
-          new_resource.source ? Array(new_resource.source) : nil
+          srcs = new_resource.source || Chef::Config[:rubygems_url]
```
lamont-granquist: it seems like this also fixes a bug where gem_package installs without a binary shell_out have never used the config variable.

lamont-granquist:
master is red right now, but this didn't seem to break anything additionally in the last run (plus i keep on getting ninja'd on RELEASE_NOTES and having to fix conflicts, so now its someone else's turn...)

PR #5997, opened 2017-04-03
windows_task with Gemfile.lock change

btm:
This is #5886 after resetting the Gemfile.lock and running `rake dependencies:update` and `bundle install --no-deployment`. /cc @NimishaS


.bundle/config
```diff
@@ -1,2 +1,3 @@
 ---
 BUNDLE_FROZEN: "1"
+BUNDLE_WITHOUT: "omnibus_package:test:pry:integration:docgen:maintenance:travis:aix:bsd:linux:mac_os_x:solaris:windows:development"
```
tas50: This shouldn't be merged here

btm: Yeah, I kind of figured that but `rake dependencies:update` keeps adding it back.

lamont-granquist: `rake dependencies:update_gemfile_lock`

but you'll still pull in ohai which will break amazon linux -- working on that.

lib/chef/provider/windows_task.rb
```diff
@@ -0,0 +1,411 @@
...
+        end
+
+        task_xml = load_task_xml task_name
+        task.merge!(task_xml) if task && task_xml
+
+        task
+      end
+
+      def load_task_xml(task_name)
+        xml_cmd = shell_out("chcp 65001 >nul 2>&1 && schtasks /Query /TN \"#{task_name}\" /XML")
```
jugatsu: https://github.com/chef/chef/pull/5886#discussion_r108259818
I also made PR to upstream https://github.com/chef-cookbooks/windows/pull/461

btm: @jugatsu, I asked @NimishaS and they said that they had swapped in `powershell_out` in one place (line 207) but this place would take more changes to the code than just swapping. They're going to look at it again, but we also want to make sure this gets merged this week for Chef 13.

btm:
@NimishaS is pulling the Gemfile.lock change into https://github.com/chef/chef/pull/5886.

lib/chef/provider/windows_task.rb
NimishaS: @btm , fixed the review comment. Turned out that I only had to remove `stdout.force_encoding('UTF-8')` from `powershell_out("#{task_script}").stdout.force_encoding('UTF-8')` to make it working in the later case.

PR #5998, opened 2017-04-03
bumping Gemfile.lock to pick up ohai amazon change

lamont-granquist:
expecting this to go red and need code/spec changes in chef


lamont-granquist:
@chef/client-core i think this'll eventually go green.  there's going to need to be a bunch of cookbook work to actually make the amazonlinux tests go green.

PR #6000, opened 2017-04-03
Chef-13: Remove the consts for DSL-based resources/providers

coderanger:
Been warning for a long time, should be low impact though as there wasn't really much use for those consts that wasn't a bug in and of itself (e.g. manually instantiating resource objects).


coderanger:
Unless there is a transient spec failure, the only failing tests should be Amazon related which are being dealt with elsewhere. Should be ready for merge.

lamont-granquist:
auspicious PR number

coderanger:
I added some firmly mediocre release notes, suggestions welcome.

thommay:
I think we should do an editing pass on the release notes on thursday regardless, and then turn them in to a true upgrading doc.

PR #6001, opened 2017-04-03
Make Standardized Exit Codes The Default Behavior

smurawski:
### Description

Sets a default in Chef::Config for `exit_status` to `:enabled`.
Removes uses of deprecated exit codes from the Chef code base.
Changed the deprecation warning/error to a warning message when attempting to exit with a non-standard exit code while standard exit codes are enforced.

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/application/exit_code.rb
+++ b/lib/chef/application/exit_code.rb
@@ -72,8 +72,6 @@ class Chef
             VALID_RFC_062_EXIT_CODES[:SIGINT_RECEIVED]
           elsif sigterm_received?(exception)
             VALID_RFC_062_EXIT_CODES[:SIGTERM_RECEIVED]
-          elsif legacy_exit_code?(exception)
-            VALID_RFC_062_EXIT_CODES[:GENERIC_FAILURE]
           elsif reboot_scheduled?(exception)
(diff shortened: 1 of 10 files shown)


coderanger:
Is there a compelling reason to leave this configurable? If not, I vote to salt the earth and make it just how Chef works since it seems unilaterally better :)

lamont-granquist:
yeah i'm actually even a bit more :-1: on configurability, since anyone tweaking that and using a cookbook which expects the RFC behavior on exit codes will be broken.  should cookbooks work under both cases?  should cookbooks that require the RFC raise if its not set correctly?  better solutions is probably to actually remove the choice...

i do normally promote the right to keep and bear footguns, but this is more like a footgun that when you pull the trigger a grenade will always go off in your face, its not really useful at all...

coderanger:
If we want to leave in the ability to have manual control over exit codes, it should probably be something structural like a special exception class rather than configured.

smurawski:
@lamont-granquist @coderanger I'm totally cool with removing the opt-out. :)

coderanger:
I vote to purge with :fire: and add it back in 13.1 if someone complains hard enough

smurawski:
The rspec failures will be fixed with https://github.com/chef/chef/pull/6009

PR #6003, opened 2017-04-04
Chef-13: Freeze property defaults

coderanger:
Been planned for a while now, and should prevent a bunch of bugs like the one we had in Mount's tests :)


lib/chef/property.rb
```diff
@@ -110,6 +110,9 @@ def initialize(**options)
         raise ArgumentError, "Cannot specify both default and name_property/name_attribute together on property #{self}"
       end
 
+      # Freeze the default if it isn't a lazy value.
+      default.freeze unless default.is_a?(DelayedEvaluator)
```
lamont-granquist: what about hashes and arrays?  pretty sure the yum_package default mutation bug was on a hash key.

coderanger:
Travis failures look unrelated. Merging.

lamont-granquist:
#4691 is the "reference bug" for why we do that.  the yum_provider flush_cache default mutation bug was awful.

this also already found errors in the tomcat cookbook.

lamont-granquist:
we also discussed and could not design a way to provide deprecation warnings for this, since you'd have to give back some kind of decorator over the object which gave deprecation warnings if any mutator methods were called, which would be nearly impossible due to both ruby being bad about decorators (truthiness, hardwired =~ operator, === comparisons, etc) and knowing what the mutator methods were (i supposed you could catch the exception, unfreeze, replay the method, and refreeze or something insane like that -- but still you'd break all kinds of code.

PR #6005, opened 2017-04-04
Allow lazy{} to be used in template resource variables.

coderanger:
It's a special case, but people do this a lot so let's make it JFW.

Not adding to RELEASE_NOTES since it seems minor by comparison.


lib/chef/provider/template/content.rb
```diff
@@ -36,7 +36,26 @@ def template_location
...
+              end
+            when Array
+              # Avoid mutating arrays in the resource in case we're changing anything.
+              obj.map { |value| visitor.call(value) }
+            when DelayedEvaluator
+              new_resource.instance_eval(&obj)
+            else
+              obj
+            end
+          end
```
lamont-granquist: this smells like you could make it into a chef/mixin/visitor, didn't you just do something very similar for the ohai freezing, and i think i just asked you to try to do something similar for properties freezing?

coderanger: +1 overall but will need to think through the API commonalities, not sure how much more readable a mixin-d version would end up looking vs. this. Could probably work out a few flavors though, value-only, dup style, and map! style?

coderanger:
Travis failures look unrelated, merging.

PR #6007, opened 2017-04-04
RFC 59: Load Ohai plugins

thommay:
*Minimum Viable Ohai Plugins*

This adds a new phase in the compilation of the run context, and bubbles
up the ohai object from the client to the run_context.

We definitely need some discussion about the changes to the run context
and having the full ohai object in so many places, and
`consume_ohai_data` also would benefit from careful review.

cc @chef/maintainers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,12 +12,12 @@ source "https://rubygems.org"
 gem "chef", path: "."
 
 # tracking master of ohai for chef-13.0 development, this should be able to be deleted after release
-gem "ohai", git: "https://github.com/chef/ohai.git"
+gem "ohai", git: "https://github.com/chef/ohai.git", branch: "tm/load_additional_plugins"
 
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "rake"
 gem "bundler"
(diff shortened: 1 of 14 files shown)


thommay:
requires chef/ohai#980

lamont-granquist:
yeah adding it to the initializer of the run_context is gonna break a ton of things, plus you need to be careful about passing that into child run contexts...

lamont-granquist:
yeah so Ohai::Systems shouldn't have any state that you care about, I think you can just `Ohai::System.new.run_additional_plugins()` and don't have worry about threading a single Ohai::System object through the run_context?

lamont-granquist:
(well probably rub a bit of lazy allocation on it so that you only create one object per cookbook compiler...)

thommay:
Hmm, doesn't that kind of break dependencies though? If we use the same ohai object, cookbook plugins can depend on plugins from core ohai, and not have to try and resolve them all over again…

lamont-granquist:
It shouldn't break them should it?  They should just get re-resolved and re-run and should produce the same data -- similar to an ohai provider `:reload`?

lamont-granquist:
and i guess that's sorta what i was yapping about yesterday which was that instead of doing something particularly invasive to thread the ohai object through the run_context, i'd prefer to just reload everything once and not worry about patchiness and speed...

PR #6008, opened 2017-04-04
No more knife cookbook site vendor

coderanger:
This was made into a deprecated stub a while ago, finish the cleanup.


coderanger:
Test fail is unrelated.

Issue #3123, opened 2015-03-22
make use_inline_resources the only way to write resources in Chef 13


lamont-granquist:
should have done this for Chef 12

breaking change has to be done in a major version bump.

probably want to add some kind of dont_use_inline_resources for LWRPs that actually can't use it.

tas50:
@lamont-granquist Can we close this now that custom resources turn use_inline_resources on by default?  Seems like that's the proper way to get users into the correct pattern.

lamont-granquist:
Nope, adding the nags to foodcritic is actually in preparation for simply turning this on and making everything that inherits from LWRPBase have use_inline_resources by default.

martinb3:
I think we also need to expose a way to turn this off, right? I don't see any way to currently disable `use_inline_resources`. (e.g. the `dont_use_inline_resources` that @lamont-granquist refers to above).

lamont-granquist:
If you just define a method on a provider without using the 'action' syntax you will magically turn it off right now...

To really incorporate this better, we probably want to merge this into Chef::Provider#action though so that every method gets it.

But, I'm also not clear why anyone needs this functionality any more.  We have parent pointers on run_context's now and the root Chef.run_context and you can inject the run_context into declare_resource and build_resource to mess with your parent run_context's resource collection.

AFAIK the only reason for turning it off was to write (bad and broken) accumulators using resources.

coderanger:
@lamont-granquist I will need it disabled in all Poise-based code as I have my own (better) replacement.

lamont-granquist:
fair 'nuff then.

coderanger:
But for my purposes it could be something like `override a method on the provider class`, rather than something exposed more directly.

martinb3:
> you can inject the run_context into declare_resource and build_resource to mess with your parent run_context's resource collection.

I had assumed this was a workaround -- and not something we should be doing as a public API.

lamont-granquist:
I think I may want to take a mulligan on create_if_missing but injecting the run_context is supported as far as I'm concerned.

lamont-granquist:
so, i "took a mulligan on create_if_missing" and implemented this stuff:

https://coderanger.net/rewind/

its definitely intended that those APIs be used along with the `with_run_context` helper to be able to write definitions and resources that do not require anyone using dont_use_inline_resources.   those APIs are documented and fully intended to be used publicly.

lamont-granquist:
[...]

PR #6013, opened 2017-04-05
Remove use_inline_resources

coderanger:
This removes `use_inline_resources`. This also means the `action :name do ... end` syntax is available to all providers instead of just via LWRPBase. If you want to avoid this mode, use the `def action_name ... end` syntax instead.


tas50:
What happens if use_inline_resources is there? Not an error hopefully?

coderanger:
It's currently an error, I can add it back as a no-op though :)

coderanger:
But the alternative to `use_inline_resources if defined?(use_inline_resources)`

tas50:
Having this not break the world would be ideal. Deprecation + a FC warning will clean up most of it before Chef 14 where you can just delete it.

coderanger:
Yep, just added that.

coderanger:
Will fix #3123 when merged.

coderanger:
Should be ready for review, though it will need a deprecation page made before release.

thommay:
Might be good to have a test that ensures the deprecation warning fires as expected, otherwise 👍

thommay:
OK, so we just had a long conversation about we want to do this deprecation. If we put this deprecation in as things stand, any cookbook that wants to have tests with deprecation_warnings_as_errors and still support 12 and 13 will have to wrap `use_inline_resources` in a `Gem::Version.compare` or similar, which is pretty user hostile.
How do you feel about either making `use_inline_resources` a null method for right now, or wrapping the deprecation warning up so it'll only start firing after 13.6?

PR #6015, opened 2017-04-05
Implement RFC033: Root aliases

coderanger:
https://github.com/chef/chef-rfc/blob/master/rfc033-root-aliases.md

tl;dr this makes it so you can use `mycookbook/attributes.rb` and `mycookbook/recipe.rb` instead of folders. For attributes this should probably be the dominant use case since having multiple attribute files is almost always more confusing than anything else. For recipes, this is nice for simple use cases, though there are plenty of good reasons to have many recipes so for those cases they should continue to use the folder.


coderanger:
Not super in love with how messy this code is, but :-/

PR #6017, opened 2017-04-05
RFC 59 - Load ohai plugins

thommay:
RFC 59 - Load ohai plugins, take 2. This just uses a new ohai system context to run all the plugins.

replaces #6007


thommay:
Not actually getting release notes until Chef Server supports cookbook api v2, so this is being dark launched right now :)

PR #6023, opened 2017-04-05
Improve coverage of root alias tests

coderanger:
Mostly making this a PR to ensure tests all pass, will merge when green.

This just adds test cases for alias collisions with the older filenames.


coderanger:
Not waiting on AppVeyor since it seems backed up today.

PR #6028, opened 2017-04-06
Chef-13: better solution to gem_package source issues

lamont-granquist:
I think this is better

- Chef::Config[:rubygems_uri] is nearly always applied (but
  could be set to nil if someone really wanted to opt-out)
- the per-resources sources are additive on top of the
  global config (lets cookbooks be additive to global config
  rather than override it by default).
- there's a new flag for if you want to opt-out of the global
  config completely on a per-resource basis.

So I think this gets the most commmon and the next most common
use cases correct, while still allow people with arbitrarily
crazy needs to write resources that do it.

The RELEASE_NOTES also have somewhat better user stories in them,
which I think suggests that I'm on the right track here.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -243,16 +243,19 @@ property :x, default: lazy { {} }
 
 ### Rubygems provider sources behavior changed.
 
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@chef/client-core think this is a better solution than my one from yesterday

Issue #5771, opened 2017-01-31
Chef::Platform.windows? needs to inspect the node.


lamont-granquist:
We should also inject that method into the universal DSL.

It may be possible to have it optionally inspect the node if its mixed into a context which has a node or else fall back to the RUBY_PLATFORM based inspection.

The problem right now is that everyone blindly uses this instead of `node["platform_family"] == "windows"` but chefspec/fauxhai can mock the former but not the latter.  Most of the time we'll have a node object and should inspect that.  There are legit reasons why we'd need to inspect RUBY_PLATFORM before we have the node object and in other low level stuff that chefspec should never invoke so we probably do need both methods.  I'll bet 90% of the time we use this method the caller doesn't care though and we need to make it easy to get it right, not easy to get it wrong.

I guess technically chefspec could monkeypatch the `windows?` method directly but BAAAARF.

I don't see where chefspec is monkeypatching this now, but i might be wrong.

lamont-granquist:
maybe do this for Chef-13.

btm:
@NimishaS if we ponder this and think it's the right course, as soon as master becomes Chef 13, we should make this change to `Chef::Platform.windows?` and go find all the places we still have conditionals based on `RUBY_PLATFORM` hardcoded and have them use `Chef::Platform.windows?`. It looks like there are 16.

lamont-granquist:
this kind of code is both necessary and can't inspect the node:

```
require 'chef/win32/security' if Chef::Platform.windows?
```

that has to evaluate at class loading time, and must not be parsed by chefspec either (since we're really not on a windows box and parsing that class would barf on non-windows) and we're way before ohai has been able to run so there's no node object.

so Chef::Platform.windows? probably needs to check `responds_to?(:node)` and fallback if there's no node object?  unfortunately that means it silently fails to be correct for chefspec if the node object isn't mixed into the context.

not sure how i feel about that, could force people to write something else where when they really, really need to check `windows?` before class loading time.  or we could continue along this course and just mixin the node object as any bugs crop up in chefspec (which should be fairly minimal).

should also probably check `responds_to?(:run_context)` and pull the node off the `run_context` if we have that.

tas50:
@lamont-granquist making this part of the universal DSL would be really nice. We have a mix of platform checks throughout Chef and a single helper would be useful

lamont-granquist:
[...]

PR #6030, opened 2017-04-06
Added Feature to inspect node by Chef::Platform.windows? 

piyushawasthi:
- Added Feature to inspect node by Chef::Platform.windows?
- Re-factor all codes and specs related to Chef::Platform.windows?
- Re-factor code of provider_mapping.rb.
- Fixed failing specs. 
wrt - https://github.com/chef/chef/issues/5771

### DIFF
--- a/chef-config/lib/chef-config/windows.rb
+++ b/chef-config/lib/chef-config/windows.rb
@@ -17,8 +17,9 @@
 
 module ChefConfig
   def self.windows?
-    if @node
-      platform = @node[:platform] || @node[:platform_family] || @node[:os]
+    @node ||= nil
+    if @node && @node[:os]
+      platform = "windows"
     else
       platform = RUBY_PLATFORM
     end
--- a/chef-config/lib/chef-config/windows.rb
+++ b/chef-config/lib/chef-config/windows.rb
@@ -16,13 +16,17 @@
 #
 
 module ChefConfig
-
   def self.windows?
-    if RUBY_PLATFORM =~ /mswin|mingw|windows/
+    if @node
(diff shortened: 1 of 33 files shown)


chef-config/lib/chef-config/windows.rb
```diff
@@ -17,8 +17,9 @@
 
 module ChefConfig
   def self.windows?
-    if @node
-      platform = @node[:platform] || @node[:platform_family] || @node[:os]
+    @node ||= nil
+    if @node && @node[:os]
```
smurawski: Wouldn't this always result in returning `true` if @node is present and has an :os key, even on non-windows systems?

chef-config/lib/chef-config/windows.rb
```diff
@@ -16,13 +16,18 @@
 #
 
 module ChefConfig
-
   def self.windows?
-    if RUBY_PLATFORM =~ /mswin|mingw|windows/
+    @node ||= nil
+    if @node && @node[:os]
```
smurawski: Wouldn't this always result in returning true if @node is present and has an :os key, even on non-windows systems?

piyushawasthi: The node[:os] only present for windows

NimishaS:
@btm , @lamont-granquist , @smurawski , @mwrock , please check the comments given by @coderanger in https://github.com/chef/chef/pull/6048.

btm:
We're going to wait for Lamont to come back from vacation and work this into his plans for [rfc087](https://github.com/chef/chef-rfc/blob/master/rfc087-distro-sugar-helpers.md). I'm to close this PR and we'll keep the issue #5771 open to track further progress.

PR #6032, opened 2017-04-06
server enforced required recipe

sdelano:
when the chef-server is configured to serve a requried recipe,
chef-client shall load the recipe into the run context and
execute it as part of the converge phase.

if the chef-server is NOT configured, it will return a 404 and
chef-client will continue normally.

Signed-off-by: Stephen Delano <stephen@chef.io>

### Description

See above.

See also:
* https://github.com/chef/chef-rfc/pull/258
* https://github.com/chef/chef-server/pull/1214

### Issues Resolved

* https://github.com/chef/chef-rfc/pull/258

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -280,6 +280,8 @@ class Chef
 
         run_context = setup_run_context
 
+        load_required_recipe(@rest, run_context) unless Chef::Config[:solo_legacy_mode]
+
         if Chef::Config[:audit_mode] != :audit_only
           converge_error = converge_and_save(run_context)
         end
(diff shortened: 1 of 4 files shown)


lib/chef/data_collector.rb
```diff
@@ -384,6 +384,8 @@ def data_collector_token
       end
 
       def add_resource_report(resource_report)
+        #require 'pry'
```
thommay: let's get these removed please

lib/chef/client.rb
```diff
@@ -515,6 +517,48 @@ def setup_run_context
...
+    # @return The modified run context
+    #
+    # @api private
+    #
+    # TODO: @rest doesn't appear to be used anywhere outside
+    # of client.register except for here. If it's common practice
+    # to create your own rest client, perhaps we should do that
+    # here but it seems more appropriate to reuse one that we
+    # know is already created. for ease of testing, we'll pass
+    # the existing rest client in as a parameter
```
lamont-granquist: you should probably use `rest_clean` here, so that JSON request/response is sanitized

we probably should have converted everything over to `rest_clean` for Chef-13

sdelano: The comments around `#rest_clean` indicate that it's used for API reporters. If it is indeed related to JSON sanitization, that's also non-applicable here because neither the request nor the response contain any JSON. This is simply a `GET` to retrieve a recipe file.

lamont-granquist: i guess the point is that `rest` is kinda-sorta deprecated and `rest_clean` is the future -- at some point `rest` should just set the same options as `rest_clean` and then `rest_clean` should go away.

it might not make any difference at all, but that's kind of fine and useful to know...

Issue #6044, opened 2017-04-10
Cookbooks uploaded by Chef 13 can't be used by Chef 12


coderanger:
Still working out the details but probably related to the manifest system changes, cookbooks uploaded to Chef Server from a Chef 13 workstation end up looking like this:

<details>
<pre>
{
  "version": "0.0.0",
  "name": "wat-0.0.0",
  "cookbook_name": "wat",
  "root_files": [
    {
      "name": "metadata.rb",
      "path": "metadata.rb",
      "checksum": "00065dfaaf15f58e10660ce79799531a",
      "specificity": "default",
      "url": "https://s3-external-1.amazonaws.com:443/..."
    }
  ],
  "recipes": [
    {
      "name": "default.rb",
      "path": "recipes/default.rb",
      "checksum": "08163fbceff7b03ace5b4aefe0b08924",
      "specificity": "default",
      "url": "https://s3-external-1.amazonaws.com:443/..."
    }
  ],
  "frozen?": false,
  "chef_type": "cookbook_version",
  "json_class": "Chef::CookbookVersion",
  "metadata": {
    "name": "wat",
    "description": "",
    "maintainer": "asdf",
    "maintainer_email": "asdf",
    "license": "All rights reserved",
    "platforms": {
    },
    "providing": {
      "wat": ">= 0.0.0"
    },
    "recipes": {
      "wat": ""
    },
    "version": "0.0.0",
    "source_url": "",
    "issues_url": "",
    "privacy": false,
    "chef_versions": [
    ],
    "ohai_versions": [
    ],
    "gems": [
    ],
    "attributes": {
    },
    "dependencies": {
    },
    "long_description": ""
  }
}
</pre>
</details>

Compare to a manifest for the same cookbook from a Chef 12 workstation:

<details>
<pre>
{
  "recipes": [
    {
      "name": "default.rb",
      "path": "recipes/default.rb",
      "checksum": "08163fbceff7b03ace5b4aefe0b08924",
      "specificity": "default",
      "url": "https://s3-external-1.amazonaws.com:443/..."
    }
  ],
  "definitions": [
  ],
  "libraries": [
  ],
  "attributes": [
  ],
  "files": [
  ],
  "templates": [
  ],
  "resources": [
  ],
  "providers": [
  ],
  "root_files": [
    {
      "name": "metadata.rb",
      "path": "metadata.rb",
      "checksum": "f0fb947e7c2f4b944d4f3d49fef85423",
      "specificity": "default",
[...]

PR #6045, opened 2017-04-10
Fix for Chef 13 upload reversion.

coderanger:
Refs https://github.com/chef/chef/issues/6044

### DIFF
--- a/lib/chef/cookbook/manifest_v0.rb
+++ b/lib/chef/cookbook/manifest_v0.rb
@@ -54,6 +54,11 @@ class Chef
             end
           end
         end
+        # Ensure all segments are set to [] if they don't exist.
+        # See https://github.com/chef/chef/issues/6044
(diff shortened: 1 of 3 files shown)


coderanger:
Ping @chef/client-core for high-priority review.

coderanger:
@danielsdeleo That might be hard, the bug doesn't show up in regression-able form using Chef Zero since Zero automatically does some data massage to fix things I think.

coderanger:
I can add some unit tests for it at least though :)

coderanger:
Hmm, trying to work out how the unit tests for version-specific manifest stuff works, might need help from @thommay

danielsdeleo:
Yeah, I'm 100% cool with unit tests for this one

thommay:
I've added tests for both manifest types and fixed a buglet in Manifest v0 roundtripping.

Issue #6046, opened 2017-04-11
chef 13 yum_package options option broken


Stromweld:
## Description

chef-client 13.0.113 when using the yum package resource with options property defined throws an error on unknown method split.

## Chef Version

13.0.113

## Platform Version

CentOS 6.8 running on AWS cloud

## Replication Case

use yum_package with options '--nogpgcheck' defined.

## Client Output

```
Recipe: dse-nativex::datastax
  * yum_package[dse-full] action install

    ================================================================================
    Error executing action `install` on resource 'yum_package[dse-full]'
    ================================================================================

    NoMethodError
    -------------
    undefined method `split' for ["--nogpgcheck"]:Array

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/dse-nativex/recipes/datastax.rb

      9:     package 'dse-full' do
     10:       version node['cassandra']['dse_version']
     11:       options '--nogpgcheck'
     12:       action :install
     13:     end
     14:   else

    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/dse-nativex/recipes/datastax.rb:9:in `from_file'

    yum_package("dse-full") do
      package_name "dse-full"
      action [:install]
      default_guard_interpreter :default
      declared_type :package
      cookbook_name "dse-nativex"
      recipe_name "datastax"
      version "4.8.4-1"
      options ["--nogpgcheck"]
    end

    System Info:
    ------------
    chef_version=13.0.113
    platform=centos
    platform_version=6.8
    ruby=ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
    program_name=chef-client worker: ppid=25468;start=19:31:40;
    executable=/opt/chef/bin/chef-client


Running handlers:
[2017-04-10T19:32:02-05:00] ERROR: Running exception handlers
Running handlers complete
[2017-04-10T19:32:02-05:00] ERROR: Exception handlers complete
Chef Client failed. 8 resources updated in 21 seconds
[2017-04-10T19:32:02-05:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2017-04-10T19:32:02-05:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2017-04-10T19:32:02-05:00] ERROR: yum_package[dse-full] (dse-nativex::datastax line 9) had an error: NoMethodError: undefined method `split' for ["--nogpgcheck"]:Array
[...]

PR #6047, opened 2017-04-11
Fix some fallout from the options converstion to an array

coderanger:
We should probably audit the tests for more of this pattern of using rspec-mocks on the resource object, might be other resource-y changes that we've missed for the same reason. This should make `options` work again on Yum-y platforms.

Ping @chef/client-core and @chef/client-enterprise-linux for review, should probably get hotfix'd.

Fixes https://github.com/chef/chef/issues/6046.

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -297,7 +297,11 @@ class Chef
       def expand_options(options)
         # its deprecated but still work to do to deprecate it fully
         #Chef.deprecated(:package_misc, "expand_options is deprecated, use shell_out_compact or shell_out_compact_timeout instead")
-        options ? " #{options.is_a?(Array) ? Shellwords.join(options) : options}" : ""
+        if options
+          " #{options.is_a?(Array) ? Shellwords.join(options) : options}"
+        else
+          ""
+        end
       end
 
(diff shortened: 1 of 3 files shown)


coderanger:
A dump of everywhere else we're doing this with the `options` property specifically:

```
spec/unit/provider/package/aix_spec.rb
166:      allow(@new_resource).to receive(:options).and_return("-e/tmp/installp.log")
179:      allow(@new_resource).to receive(:options).and_return("-e/tmp/installp.log")

spec/unit/provider/package/macports_spec.rb
124:      allow(@new_resource).to receive(:options).and_return("-f")
143:      allow(@new_resource).to receive(:options).and_return("-f")
161:      allow(@new_resource).to receive(:options).and_return("-f")
194:      allow(@new_resource).to receive(:options).and_return("-f")

spec/unit/provider/package/ips_spec.rb
137:      allow(@new_resource).to receive(:options).and_return("--no-refresh")

spec/unit/provider/package/pacman_spec.rb
160:      allow(@new_resource).to receive(:options).and_return("--debug")
181:      allow(@new_resource).to receive(:options).and_return("--debug")
```

I probably won't get to those tonight so consider the PR mergable if Travis goes green and I (or someone else) will stare at the others later.

PR #6048, opened 2017-04-11
Inspecting @node object to check the platform

NimishaS:

### DIFF
--- a/chef-config/lib/chef-config/windows.rb
+++ b/chef-config/lib/chef-config/windows.rb
@@ -18,7 +18,13 @@
 module ChefConfig
 
   def self.windows?
-    if RUBY_PLATFORM =~ /mswin|mingw|windows/
+    if instance_variable_defined?(:@node) && @node && @node["os"]
+      platform = "windows"
+    end
+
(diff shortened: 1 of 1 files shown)


coderanger:
This kind of magic needs a lot more justification and a loooot more tests. We already expose `Chef.node` as a global if we wanted to break encapsulation that far, but since a bunch of config stuff is used before the node object exists (let alone ohai running) this seems like it will produce temporary inconcsistent results.

NimishaS:
@coderanger , we are trying to fix https://github.com/chef/chef/issues/5771.

There's already a PR https://github.com/chef/chef/pull/6030 for which Appveyor is failing. I created this PR mainly for debugging the issue.

coderanger:
Probably use the chefclass global then, that has a much more defined life cycle.

coderanger:
And I think we need to at least revisit this given we're adding the Chef sugar APIs to core which use only Ohai data, probably table this whole thing until Lamont is back?

NimishaS:
Fixed the Appveyor issue in #6030. Closing this PR.

PR #6055, opened 2017-04-11
Clean up more stubbed properties in tests

coderanger:
This kind of got away from me as it turns out a whole bunch of tests were assuming nillable properties. I converted the two easy ones (`env` and `route`) so that's just done, for the others I could either revert the test changes or we can live with annoying hacks for nillable stuff for the few cases where it was breaking tests.


coderanger:
We should hold merge on this until after 13.0 hotfixes (if any) are finished, nothing in here is that important.

PR #6067, opened 2017-04-17
Adding functional specs for windows_task resource

NimishaS:

### DIFF
--- a/spec/functional/resource/windows_task_spec.rb
+++ b/spec/functional/resource/windows_task_spec.rb
@@ -383,7 +383,8 @@ describe Chef::Resource::WindowsTask, :windows_only do
 
     subject do
       new_resource = Chef::Resource::WindowsTask.new(task_name, run_context)
-      new_resource.command task_name
+      new_resource.command "dir"
+      new_resource.run_level :highest
       new_resource
     end
 
@@ -391,7 +392,7 @@ describe Chef::Resource::WindowsTask, :windows_only do
       subject.run_action(:create)
       subject.run_action(:run)
       task_details = windows_task_provider.send(:load_task_hash, task_name)
-      expect(task_details[:Status]).to eq("Ready")
+      expect(task_details[:Status]).to eq("Running")
     end
   end
 
--- a/spec/functional/resource/windows_task_spec.rb
+++ b/spec/functional/resource/windows_task_spec.rb
@@ -1,0 +1,458 @@
+#
+# Author:: Nimisha Sharad (<nimisha.sharad@msystechnologies.com>)
+# Copyright:: Copyright (c) 2016 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


NimishaS:
@btm , @mwrock , this PR is ready for review.

mwrock:
This looks good. It would be great if a `:run` action test could test that the task is in the running state and also validate that the command ran.

NimishaS:
@mwrock , I had previously added `expect(task_details[:Status]).to eq("Running")` for `:run` action, which ran successfully on my local machine. But on Appveyor, it was somehow failing and receiving the task status as `Ready`. So I changes the expectation later on.
Any idea why would the Appveyor system return task status as `Ready` instead of `Running` even after running `subject.run_action(:run)` command?

spec/functional/resource/windows_task_spec.rb
```diff
@@ -0,0 +1,458 @@
...
+      new_resource = Chef::Resource::WindowsTask.new(task_name, run_context)
+      new_resource.command task_name
+      new_resource
+    end
+
+    it "runs the existing task" do
+      subject.run_action(:create)
+      subject.run_action(:run)
+      task_details = windows_task_provider.send(:load_task_hash, task_name)
+      expect(task_details[:Status]).to eq("Ready")
```
NimishaS: @mwrock , @btm, I previously had `expect(task_details[:Status]).to eq("Running")` condition here, but Appveyor was returning the Status as `Ready` for the task even after running `subject.run_action(:run)`.

Please suggest how can we debug this issue for Appveyor system.

mwrock: you could try debugging an appveyor instance via RDP. See https://www.appveyor.com/docs/how-to/rdp-to-build-worker/. You might also try running a script that creates a file or performs some action that you can check to determine if the task ran at all.

NimishaS: @mwrock , I found that when I login with `appveyor` user, I am not able to run the scheduled tasks of `SYSTEM` user manually. I tried creating the scheduled task for `appveyor` user, but it's giving error `The user name or password is incorrect.`(current appveyor failure)
If we pass `appveyor` user while task creation then we'll have to give password also in the `windows_task` recipe. This will mean that we need to update the `appveyor.yml` file as well to create the password: https://github.com/chef/chef/pull/6067/files#diff-180360612c6b8c4ed830919bbb4dd459R8
The standard way of giving password is through environment variables, but that will also mean that appveyor will expect this password to be set on every developers machine for running these tests.
@btm, @mwrock , please share your views on this.

mwrock:
You might consider creating appveyor only tests that create a new dummy admin user with a known password.

NimishaS:
@mwrock, I realized that when I had logged into the appveyor system, `chef-client` wasn't installed. The scheduled task in the specs tried to run `chef-client` command, which may result into failure. So changed the command from `chef-client` to `dir`. Also noticed that the task wasn't running with highest privileges, so updated that too.

Now specs are running fine.

Issue #6051, opened 2017-04-11
no_lazy_load false yields Chef::Exceptions::FileNotFound on cookbook_file/remote_directory resources


natewalck:
## Description

Using `no_lazy_load false` causes cookbook_file/remote_directory resources to fail on Chef 13.

## Chef Version

```
chef-client --version
Chef: 13.0.113
```
## Platform Version

macOS 10.12.4

## Replication Case

1. Edit `/etc/chef/client.rb` and set `no_lazy_load false`
2. Run a cookbook thats uses a remote_directory resource. For instance:
```
# Cookbook: default.rb
rd = remote_directory '/tmp/rd_test' do
  source 'some/folder'
  owner 'root'
  group 0
  mode 0755
  purge true
  recursive true
  action :nothing
end

rd.run_action(:create)
```

## Client Output
```
Recipe: rd_test::default
  * remote_directory[/tmp/rd_test] action create

    ================================================================================
    Error executing action `create` on resource 'remote_directory[/tmp/rd_test]'
    ================================================================================

    Chef::Exceptions::FileNotFound
    ------------------------------
    cookbook rd_test has no directory files/default/some/folder

    Cookbook Trace:
    ---------------
    /etc/chef/local-mode-cache/cache/cookbooks/rd_test/recipes/default.rb:11:in `from_file'

    Resource Declaration:
    ---------------------
    # In /etc/chef/local-mode-cache/cache/cookbooks/rd_test/recipes/default.rb

      1: rd = remote_directory '/tmp/rd_test' do
      2:   source 'some/folder'
      3:   owner 'root'
      4:   group 0
      5:   mode 0755
      6:   purge true
      7:   recursive true
      8:   action :nothing
      9: end
     10:

    Compiled Resource:
    ------------------
    # Declared in /etc/chef/local-mode-cache/cache/cookbooks/rd_test/recipes/default.rb:1:in `from_file'

    remote_directory("/tmp/rd_test") do
      action [:nothing]
      default_guard_interpreter :default
      path "/tmp/rd_test"
      recursive true
      source "some/folder"
      purge true
      files_backup 5
      files_owner nil
      files_group nil
      files_mode 420
      overwrite true
      declared_type :remote_directory
      cookbook_name "rd_test"
      recipe_name "default"
      owner "root"
      group 0
      mode "0755"
    end

    System Info:
    ------------
    chef_version=13.0.113
    platform=mac_os_x
    platform_version=10.12.4
[...]

PR #6070, opened 2017-04-18
[On Hold] fix lazy loading of cookbooks

thommay:
#6051


thommay:
I'd like to do this, still. But the API changes and everything else mean it'll need to wait for 14.

PR #6071, opened 2017-04-18
The two key list commands didn’t actually require the helper object they both use.

coderanger:
Boring PR is boring.


chilicheech:
👍

PR #6074, opened 2017-04-19
Update package resource with default for timeout.

iennae:
### Description

Currently the timeout attribute is inconsistent in behavior between windows and non-windows platforms. This was initially observed in chef-ingredient with allowing the resource to be of type NilClass. Converges would fail on Windows but succeed on other platforms. This PR moves the default value to the package superclass and removes the attribute from windows_package.

### Issues Resolved

https://github.com/chef-cookbooks/chef-ingredient/issues/155

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/package.rb
+++ b/lib/chef/resource/package.rb
@@ -40,7 +40,7 @@ class Chef
       property :response_file, String, desired_state: false
       property :response_file_variables, Hash, default: lazy { {} }, desired_state: false
       property :source, String, desired_state: false
-      property :timeout, [ String, Integer ], desired_state: false
+      property :timeout, [ String, Integer ], desired_state: false, default: 600
 
     end
   end
--- a/lib/chef/resource/windows_package.rb
+++ b/lib/chef/resource/windows_package.rb
@@ -42,7 +42,6 @@ class Chef
 
       # Unique to this resource
       property :installer_type, Symbol
-      property :timeout, [ String, Integer ], default: 600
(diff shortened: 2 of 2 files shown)


coderanger:
I hate to ask it, but do we think 10 minutes is long enough for this to not be a compat issue? :( I've seen some pretty big (multi-GB) RPMs out there which could easily take longer than that to install and if I'm reading this right, the old default was no timeout.

iennae:
It's a good question. The old default was 600 seconds in documentation for Windows. In making things less bad, what is better:

* setting a default explicitly 
* having different behaviors on cookbooks for different platforms.

I guess what I'm asking is whether a default should be set, and if not having a default timeout impacts one platform more than another which is why Windows had it configured and others don't?

coderanger:
I can't say for sure, but if I had to guess it's just because the package resource is much older and we tried to Do Things Right when making windows_package :-) Not against having a default timeout on all platforms, just worried it will negatively impact people.

lamont-granquist:
old timeout was actually 900 in most cases.

https://github.com/chef/chef/blob/master/lib/chef/mixin/shell_out.rb#L86
https://github.com/chef/chef/blob/master/lib/chef/provider/package.rb#L639

paludis updated that to an hour because it does compiles:

https://github.com/chef/chef/blob/master/lib/chef/resource/paludis_package.rb#L30

it kinda sorta looks like the freebsd ports cookbook also tries to bump it up as well, hardcoded in the provider.

i'm working on another PR to clean all this junk up properly

PR #6077, opened 2017-04-20
add a pointer from the run_context back to the runner

lamont-granquist:
because surfing through ObjectSpace makes kittens cry

### DIFF
--- a/lib/chef/run_context.rb
+++ b/lib/chef/run_context.rb
@@ -111,6 +111,11 @@ class Chef
     attr_reader :audits
 
     #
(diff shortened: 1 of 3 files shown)


lamont-granquist:
@chef/client-core reviewage, particularly from @coderanger

lamont-granquist:
i guess this turns into circular references, so maybe it could be a weak reference, but doesn't the mark-sweep GC in ruby deal with that anyway?

coderanger:
I can neither confirm nor deny that I'm doing the ObjectSpace analysis in production code. This is fine 🔥.

PR #6078, opened 2017-04-20
Looks like this is broken on SLES 12.0/x64 too

thommay:


thommay:
@chef/maintainers easy one. Green on our SLES testers with this.

PR #6079, opened 2017-04-21
simplify external tests, finally bump bundler, remove massive Gemfile.lock closure over poise+halite+chefspec+etc

lamont-granquist:
TL;DR:  The cure is worse than the disease, this reverts the cure

- all the locking we're doing cannot automate one iota of the human work required to deal with upstream gem incompatibilities
- locking gems down so that they cannot break master out of band pushes that work onto the first human that needs to bump the deps to get work done
- the maintenance of this framework takes substantial amounts of time which is always net added onto the time necessary to deal with incompatibilities
- the closure in the Gemfile.lock over all external tests and their development dependencies creates gem incompatibilities which are completely unnecessary to solve in order to ship product (we shouldn't even be shipping halite, much less worrying about halite and chefspec getting into a fight over development deps)
- pinning the externals tests means that we're not getting accurate tests of the existing state of the world -- we're testing against poise from a day, a week or months ago, not the current release.

this PR also shifts to pulling master of the upstreams deps instead of released gems.  this is a good thing because it is more agile.  it was used during the chef-13 development cycle in order to fix upstream breaks in master of repos without necessitating doing gem releases and without needing to worry about chef-13/chef-12 compat issues (master could be fixed for chef-13 quickly, chef-12 compat could be added later, then a gem release could be done a few days after that -- the first action to fix master of e.g. poise would fix chef/chef master and the rest could be done lazily).

the objection to that latter approach would be that it makes it more likely that upstream merges to master could break chef/chef travis builds out of band, but all of those gems test against recent chef, and many of those gems test against master of chef (and arguably all of them /should/ test against master of chef if they do not).

also if its desired that we do go about locking the gems then the approach we had been taking is incorrect, and other things need to happen:

- first, we need the actual bump bot which was responsible for bumping deps, auto-submitting bump PRs and merging them if green, and cutting tickets to humans if not.  this work appears to have stalled and/or been forgotten, but without that automation the solution is untenable.
- second, the mega-Gemfile.lock is wrong.  we do not want a single depsolved solution over all the dev gems and *their* dev dependencies.  if we go down this road we need to peel off individual gemfiles for each test gem and
individual gemfile.lock files for each one of them.
- in order to do transitive Gemfile.locking sanely and supportably we would need that kind of a feature in bundler via something like bundler/bundler#5615


lamont-granquist:

ranjib:
👍

PR #6081, opened 2017-04-22
Remove lib-backcompat dir from require_path

anatol:
There is no such directory, no need to require it.


thommay:
Obvious fix, thanks!

PR #6087, opened 2017-04-25
lazy load cookbook files

thommay:
Fixes: #6051 and supercedes: #6070

cc @chef/maintainers

### DIFF
--- a/spec/integration/client/client_spec.rb
+++ b/spec/integration/client/client_spec.rb
@@ -496,7 +496,6 @@ EOM
     end
   end
 
-
   when_the_repository "has a cookbook with an ohai plugin" do
     before do
       file "cookbooks/x/recipes/default.rb", <<-RECIPE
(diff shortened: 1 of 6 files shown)


spec/integration/client/client_spec.rb
```diff
@@ -306,6 +306,7 @@ class ::Blah
 cookbook_path "#{path_to('cookbooks')}"
 EOM
       result = shell_out!("#{chef_client} -c \"#{path_to('config/client.rb')}\" -o 'x::default' -z --profile-ruby", :cwd => chef_dir)
+      result.error!
```
danielsdeleo: My recollection is that `shell_out!` with the `!` should implicitly call `error!`

thommay:
@coderanger yeah, #6070 was intended to do that but it's too big for right now (and done right would break some APIs)

PR #5874, opened 2017-03-07
Fix apt_repository for latest os version 16.10

afiune:
### Description
Great description and solution of the problem made by @jeremiahsnapp in
the below issue!

Closes https://github.com/chef/chef/issues/5831

I have validated this works fine for the following os:
* Ubuntu 14.04
* Ubuntu 16.04
* Ubuntu 16.10 (latest)
* Debian-6.0.10
* Debian-8.7 (latest)

Signed-off-by: Salim Afiune <afiune@chef.io>

### Issues Resolved

* COOL-685
* https://github.com/chef/chef/issues/5831

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #6094, opened 2017-04-27
apt-key finger no longer matches extract fingerprints regex

whiteley:
Signed-off-by: Matt Whiteley <mattwhiteley@gmail.com>

### Description

https://github.com/chef/chef/pull/5874 caused keys to be retrieved from keyserver and installed on every chef run. The regex in `extract_fingerprints_from_cmd` was changed but it's called in both `install_key_from_uri` and `install_key_from_keyserver` and it will never match the command output in `apt-key finger`.

I'm not sure this is the desired way to do this, happy to change up if it's not.

### Check List

- [X] All tests pass

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -33,6 +33,9 @@ class Chef
         which("apt-get")
       end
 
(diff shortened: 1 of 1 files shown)


thommay:
@chef/maintainers PTAL

Issue #6092, opened 2017-04-26
undefined method managing_home_dir? on AIX


mgrubb:
## Description

When running chef-client  13.0.118 on AIX 7.2 with the following recipe:
```ruby
user 'root' do
  shell '/bin/ksh'
  home '/root'
  manage_home true
  action :modify
end
```
I'm getting:
```
================================================================================
Error executing action `modify` on resource 'aix_user[root]'
================================================================================

NoMethodError
-------------
undefined method `managing_home_dir?' for #<Chef::Provider::User::Aix:0x000001143edbf8>
```
## Chef Version

System Info:
------------
chef_version=13.0.118
platform=aix
platform_version=7.2
ruby=ruby 2.4.1p111 (2017-03-22 revision 58053) [powerpc-aix6.1.0.0]
program_name=chef-client worker: ppid=13304156;start=14:30:27;
executable=/opt/chef/bin/chef-client

## Platform Version

AIX 7.2

## Replication Case

Copy above resource for user 'root' into a file `root.rb` then run `chef-client --local-mode root.rb` on an AIX system using Chef version 13.0.118.

## Client Output

Full output of the chef-client run: https://gist.github.com/mgrubb/4180f0e17b79b8fa1104985c044d7c59

Gist of debug run: https://gist.github.com/mgrubb/ed1d1c98fab8a7fde45d9d4d023dd312

I'm using local-mode for this test case, but it behaves the same way when running the chef-client connected to our chef-server.

## Stacktrace

https://gist.github.com/mgrubb/42ad6cdf717bb74147c973f90799165f

PR #6097, opened 2017-04-27
Fix manage_home on AIX

thommay:
Fixes: #6092

cc @chef/maintainers


lamont-granquist:
still need to find a way to nuke codecov...

PR #6107, opened 2017-04-29
Update all the things for Chef 12.

rhass:
### Description

Update all the Gemfile and Berksfile locks for Chef 12.

Note, this may unconver some things which cannot safely be upgraded past a certain version since I found stuff like this:
https://github.com/chef/chef/blob/chef-12/version_policy.rb#L87


Gemfile.lock
```diff
@@ -345,39 +314,10 @@ GEM
...
-      test-kitchen (~> 1.7, >= 1.7.1)
-      travis (~> 1.8, >= 1.8.1)
-      vagrant-wrapper
-      winrm (>= 1.6, < 3)
-      winrm-fs (>= 0.4, < 2)
-      yard (~> 0.8)
-      yard-classmethods (~> 1.0)
-    poise-profiler (1.0.1)
-      halite (~> 1.0)
+    poise-boiler (0)
```
lamont-granquist: ^ that seems odd ^

rhass: Yeah, I was thinking that too. I am not sure what is going on here. Maybe @coderanger can provide some insight/context on this one?

lamont-granquist: did you `rake dependencies:update_gemfile_lock` to do this?  you might also try blowing away the Gemfile.lock and then redo'ing it.  i think i removed the rm_f that the code used to do of the Gemfile.lock when i was refactoring stuff, and i think that's still in the 12.x branch, but in the 13.x branch i've gone back to nuking the Gemfile.lock first in that rake task.

rhass: I did `rake dependencies:update` to update all the things. There probably remains some gems we need to constrain to ensure compatibility with Chef 12 though.

coderanger: I should go yank that version, thanks for the reminder :) I use a `0` version as a placeholder when reserving names but it can allow bad version solutions.

coderanger: Should be cleaned up now.

rhass: @coderanger Cool, thanks!

rhass:
I haven't got this to update quite right, but its old anyway.

PR #6111, opened 2017-05-01
Update Habitat plan to correctly build

elliott-davis:
### Description

Updates habitat plan to build chef 13 correctly. Also adds config settings for data-collector

### Check List

- [ ] All tests pass
***There are no integration tests to validate this***

### DIFF
--- a/habitat/config/client.rb
+++ b/habitat/config/client.rb
@@ -1,0 +1,21 @@
+cookbook_path "{{pkg.svc_data_path}}/{{cfg.cookbook_path}}"
+data_bag_path "{{pkg.svc_data_path}}/{{cfg.data_bag_path}}"
+environment_path "{{pkg.svc_data_path}}/{{cfg.environment_path}}"
+file_backup_path "{{pkg.svc_data_path}}/{{cfg.file_backup_path}}"
+node_path "{{pkg.svc_data_path}}/{{cfg.node_path}}"
+role_path "{{pkg.svc_data_path}}/{{cfg.role_path}}"
+pid_file "{{pkg.svc_data_path}}/{{cfg.pid_file}}"
+data_collector.server_url "{{cfg.data_collector.url}}"
+data_collector.token "{{cfg.data_collector.token}}"
+data_collector.mode "{{cfg.data_collector.mode}}".to_sym
+data_collector.raise_on_failure {{cfg.data_collector.raise_on_failure}}
+chef_repo_path "{{pkg.svc_data_path}}/chef"
+environment "{{cfg.environment}}"
+minimal_ohai {{cfg.minimal_ohai}}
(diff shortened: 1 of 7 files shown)


habitat/default.toml
```diff
@@ -1,11 +1,28 @@
-# The Chef run interval in seconds
-interval = 10 
+[chef-client]
```
smith: Should we be using underscores here? Would like to get consistency about this across core/chef plans. Not a blocker, but the other variables use underscores in this file.

elliott-davis: the more I think about it - I don't really like having that chef-client table there at all. I think I'll remove it unless you have any objections. After looking through the docs there are only a couple of nested keys so it's sort of redundant to have a chef-client table in the chef-client package.

habitat/config/client.rb
```diff
@@ -0,0 +1,21 @@
+cookbook_path "{{pkg.svc_data_path}}/{{cfg.cookbook_path}}"
+data_bag_path "{{pkg.svc_data_path}}/{{cfg.data_bag_path}}"
+environment_path "{{pkg.svc_data_path}}/{{cfg.environment_path}}"
+file_backup_path "{{pkg.svc_data_path}}/{{cfg.file_backup_path}}"
+node_path "{{pkg.svc_data_path}}/{{cfg.node_path}}"
+role_path "{{pkg.svc_data_path}}/{{cfg.role_path}}"
+pid_file "{{pkg.svc_data_path}}/{{cfg.pid_file}}"
```
tduffield: nit: this feels like it should go in `svc_var_path`

habitat/hooks/run
```diff
@@ -1,8 +1,6 @@
...
-export GEM_HOME="{{pkg.path}}/ruby/2.3.0"
-export GEM_PATH="$(hab pkg path core/ruby)/lib/ruby/gems/2.3.0:$(hab pkg path core/bundler):$GEM_HOME"
-export APPBUNDLER_ALLOW_RVM=true
-
 exec 2>&1
-exec chef-solo --fork -i {{cfg.interval}} -s {{cfg.splay}} -l {{cfg.log_level}} -o {{cfg.run_list}}
 
+export SSL_CERT_FILE="{{pkgPathFor "core/cacerts"}}/ssl/cert.pem"
+
+exec chef-solo --fork -c {{pkg.svc_config_path}}/client.rb
```
tduffield: I (think) this might want to be `--no-fork`?

rhass: We can't use `--no-fork` as it is required in Chef 12 and later. We can however remove the exec since it is redundant.

adamhjk: No it’s not. It means the pid doesn’t change.

PR #6112, opened 2017-05-02
Remove chef-gem software definition complexity from omnibus builds

lamont-granquist:
As far as I can tell all of this machinery is to extract individual chef-gems out to their own software definitions.  Mostly this optimizes for git-caching so that when the Gemfile.lock is updated then we don't have to rebuild all the gems.  The simpler solution here is to only break out nokogiri and libgecode to their own libraries to optimize those.  Since we never got the auto-Gemfile.lock-bumping robot and its still just me and tim manually submitting PRs every now and then, this problem is nowhere near important enough to solve.  Once nokogiri and libgecode are optimized by their own software definitions the rest of the gems are also fast.

This PR also shifts us back to using the software definition in omnibus-software instead of forking it.  I'm now extremely convinced that this is the correct way forwards.


ci/verify-chef.bat
```diff
@@ -46,8 +48,11 @@ call %EMBEDDED_BIN_DIR%\rspec --version
 
 SET PATH=C:\opscode\%PROJECT_NAME%\bin;C:\opscode\%PROJECT_NAME%\embedded\bin;%PATH%
 
-REM ; Test against the vendored chef gem (cd into the output of "bundle show chef")
-for /f "delims=" %%a in ('bundle show chef') do cd %%a
+REM ; Test against the vendored chef gem (cd into the output of "gem which chef")
+for /f "delims=" %%a in ('gem which chef') do set CHEFDIR=%%a
+call :dirname "%CHEFDIR%" CHEFDIR
+call :dirname "%CHEFDIR%" CHEFDIR
```
thommay: why twice?

thommay:
I think this is definitely the correct direction 👍

ci/verify-chef.bat
lamont-granquist: `chef-13.0.118/lib/chef.rb` is what comes back and you need to lop it off twice to get back to the parent dir.

lamont-granquist:
@chef/client-core i'll probably squash this and maybe do a bit of tweaking but this and the omnibus-software PR will be merging tomorrow

lamont-granquist:
moves the needle towards green....

ksubrama:
@lamont-granquist  This was not an optimization when it was written - it was working around a different issue which was that we were running "bundle install" under various ruby software definitions such as nokogiri or libgecode and these would install different (and possibly mutually incompatible) gems from the ones that chef would finally install when we ran bundle install on it.  The particular implementation was convoluted because bundler 1.11.x at the time didn't give us enough ability to lock down various gem versions across platforms - i.e. if you ran bundle install in the chef repo on windows, you'd get a Gemfile.lock with windows ffi and win32-* gem entries that would then cause other platforms to fail to perform bundle install.  There were workarounds but these workarounds caused different subtle gem/bundler version sensitive issues on various platforms.  A bunch of this code is probably irrelevant today given that bundler 1.14.x has added cross-platforms gem pinning support.  But not all of it is irrelevant.  We still need to ensure that multiple versions of ohai or chef-config or faraday or other things like that don't get installed in an omnibus build of chef-client because bugs related to those have wasted a lot of our time before.

lamont-granquist:
yeah but that can be solved via a small patch to the appbundling routine in omnibus:

https://github.com/chef/omnibus/pull/776/files

and then appbundling those gems against the chef-dk Gemfile.lock in the software definition:

https://github.com/chef/chef-dk/pull/1248/files#diff-6daaafd26626ee0f9885c9a4132b960dR64

ksubrama:
:+1:  Yeah modern bundler offers way better features for us to do this than the wacky hacks that went into this originally.

PR #6114, opened 2017-05-02
switch to using trusty on travis

lamont-granquist:
the travis precise openssl libs appears to hate rubygems and we should
do this anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: ruby
 sudo: false
 cache: bundler
+dist: trusty
 
 # Early warning system to catch if Rubygems breaks something
 before_install:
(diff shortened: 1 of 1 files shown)


kitchen-tests/Gemfile.lock
```diff
@@ -47,15 +47,15 @@ GEM
     celluloid-io (0.16.2)
       celluloid (>= 0.16.0)
       nio4r (>= 1.1.0)
-    chef-config (12.19.36)
+    chef-config (12.20.3)
```
tas50: 12?

lamont-granquist: that was necessary before 13 was released, it can probably be fixed now, but its not relevant to this mess.

lamont-granquist:
this already went green before i squashed and force-pushed.

PR #6119, opened 2017-05-04
[In-progress] Calling reconfigure outside trap('HUP')

NimishaS:
### Description

Ruby 2.0+ prevents you from performing a lot of actions inside a trap. Previously it would allow it but risked deadlocking if the trap interfered with resources the main thread was using.
Hence calling `reconfigure` from trap raises error if any `require` statement is given in `client.rb` file.

### Issues Resolved

https://github.com/chef/chef/issues/4578

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #6122, opened 2017-05-04
reconfigure between runs when running in a loop

### LINKED ISSUE #4578: SIGHUP kills chef-client when loading library in client.rb
Hello,

When loading a library in the /etc/chef/client.rb (using the attributes) we get the following code in the client.rb

```
["sbp_nagios_chefclient_handler/sbp_nagios_chefclient_handler"].each do |lib|
  begin
    require lib
  rescue LoadError
    Chef::Log.warn "Failed to load #{lib}. This should be resolved after a chef run."
  end
end
```

when sending a SIGHUP (kill -1) to the chef-client to reload the configuration, the chef-service client dies

```
[2016-02-15T15:13:33+01:00] FATAL: Configuration error ThreadError: can't be called from trap context
[2016-02-15T15:13:33+01:00] FATAL:   /etc/chef/client.rb:3:in `block in from_string'
[2016-02-15T15:13:33+01:00] FATAL:   /etc/chef/client.rb:1:in `each'
[2016-02-15T15:13:33+01:00] FATAL:   /etc/chef/client.rb:1:in `from_string'
[2016-02-15T15:13:33+01:00] FATAL: Aborting due to error in '/etc/chef/client.rb'
```

This only happens when loading the library, for the second time due to the sighup.
Tested with a empty library, to exclude code issues:

```
require 'chef/handler'
require 'chef/config'

class SBP
  class Nagios
    class Chefclient
      class Handler < ::Chef::Handler
        def initialize
        end
      end
    end
  end
end
```

Tested on RHEL7 with Chef-client 12.4.3 and 12.7.2
This is especially important because the SIGHUP signal gets sent every sunday in the default logrotate configuration on linux, and this crashing chef-client every week when using libraries.

Temporary workarround is doing a restart instead of reload in logrotate for rhel7 and systemd alike, but it would be better if chef-client handles this correctly.

```
override['chef_client']['log_rotation']['postrotate'] = 'systemctl restart chef-client.service >/dev/null || :'
```

Kind Regards,

btm:
We can't reconfigure in a trap context because we might open files or another activity that could cause a deadlock. Ruby 2.0+ now prevents this by raising a ThreadError.

Fixes #4578, Replaces #6119.

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -304,6 +304,7 @@ class Chef::Application::Client < Chef::Application
     :boolean        => false
 
   IMMEDIATE_RUN_SIGNAL = "1".freeze
+  RECONFIGURE_SIGNAL = "H".freeze
 
   attr_reader :chef_client_json
 
@@ -409,13 +410,14 @@ class Chef::Application::Client < Chef::Application
       SELF_PIPE.replace IO.pipe
 
       trap("USR1") do
-        Chef::Log.info("SIGUSR1 received, waking up")
+        Chef::Log.info("SIGUSR1 received, will run now or after the current run")
         SELF_PIPE[1].putc(IMMEDIATE_RUN_SIGNAL) # wakeup master process from select
       end
 
+      # Override the trap setup in the parent so we can avoid running reconfigure during a run
       trap("HUP") do
-        Chef::Log.info("SIGHUP received, reconfiguring")
-        $reconfigure = true
+        Chef::Log.info("SIGHUP received, will reconfigure now or after the current run")
+        SELF_PIPE[1].putc(RECONFIGURE_SIGNAL) # wakeup master process from select
       end
     end
   end
@@ -443,7 +445,6 @@ class Chef::Application::Client < Chef::Application
   private
 
   def interval_run_chef_client
-    reconfigure if $reconfigure
     if Chef::Config[:daemonize]
       Chef::Daemon.daemonize("chef-client")
 
@@ -460,14 +461,14 @@ class Chef::Application::Client < Chef::Application
   end
 
   def sleep_then_run_chef_client(sleep_sec)
-    @signal = test_signal
-    unless @signal == IMMEDIATE_RUN_SIGNAL
-      Chef::Log.debug("Sleeping for #{sleep_sec} seconds")
-      interval_sleep(sleep_sec)
-    end
-    @signal = nil
+    Chef::Log.debug("Sleeping for #{sleep_sec} seconds")
+
+    # interval_sleep will return early if we received a signal (unless on windows)
+    interval_sleep(sleep_sec)
 
     run_chef_client(Chef::Config[:specific_recipes])
+
+    reconfigure
   rescue SystemExit => e
     raise
   rescue Exception => e
@@ -480,10 +481,6 @@ class Chef::Application::Client < Chef::Application
     Chef::Application.fatal!("#{e.class}: #{e.message}", e)
   end
 
-  def test_signal
-    @signal = interval_sleep(0)
-  end
-
   def time_to_sleep
     duration = 0
     duration += rand(Chef::Config[:splay]) if Chef::Config[:splay]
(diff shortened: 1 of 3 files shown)


btm:
@NimishaS This is what I was thinking, but needs testing. I haven't tried it yet.

btm:
I fixed the chefspec issue but there are a couple failing tests to look at.

NimishaS:
@btm, please cherry-pick https://github.com/MsysTechnologiesllc/chef/commit/784dfe4e2779e49e1fb84dadcf56d657726a81f1

btm:
@chef/client-maintainers this is ready for review.

danielsdeleo:
This LGTM. I almost wonder if we should just reconfigure regardless, though that could be a who moved my cheese thing.

lamont-granquist:
i suspect you need to copypasta everything from client.rb to solo.rb as well still.

lib/chef/application/client.rb
```diff
@@ -454,11 +460,12 @@ def interval_run_chef_client
   end
 
   def sleep_then_run_chef_client(sleep_sec)
-    @signal = test_signal
-    unless @signal == IMMEDIATE_RUN_SIGNAL
-      Chef::Log.debug("Sleeping for #{sleep_sec} seconds")
-      interval_sleep(sleep_sec)
-    end
```
lamont-granquist: i'm a little bit confused as to what this logic used to do and what happens to it now

lamont-granquist:
what is the downside of always reconfiguring?  it seems preferable since it ensures that the running daemon always has the current config.  this solution leaves open the possibility that the "external" code which updates the config will fail to correctly signal a reconfiguration and the event will get lost and the old config will remain running.  i'm more worried about that edge case causing user issues than the edge case of people who really want to update config and not have the daemon pick up that config...

lib/chef/application/client.rb
btm: client client starts, you come into sleep_then_run_chef_client, you haven't had any signals yet because you just started so you go into interval_sleep

in interval_sleep, we hang out and watch the pipe. If something hits it, we're going to drop back out below on 466. One of two things will hit the pipe. An `RECONFIGURE_SIGNAL` (H) will cause us to reconfigure on 468. if it isn't that it is `IMMEDIATE_RUN` (1). In which case, we're about to run anyway, so let's do that (I could add a comment or a line here around 468 to be more clear about that).

Then we run on 471, and we're going to end up looping back around to the top of this method. `test_signal` used to use `interval_sleep` to look to see if anything was sitting in the pipe, but that's redundant because we're just about to call `interval_sleep` anyway, there's nothing happening in between. We don't need a quick check because `interval_sleep` with a longer timeout is still going to return as soon as you send a signal and a character gets dropped into the pipe, or is already there waiting because it was sent in the middle of a previous run.

This is a bit confusing, I tried to improve it which was part of why `test_signal` got renamed to `check_for_signal` but when I realized the above I just avoided it and it looks like I forgot to prune it.

btm:
I *guess* always reconfiguring is fine? If you're doing something to your chef config that isn't atomic, you've probably built something fragile anyway. It feels like a significant albeit minimal change. I'll send a note to the dev list.

btm:
https://discourse.chef.io/t/always-reconfigure-between-runs-when-daemonized/10931

lamont-granquist:
Yeah, although I'd say we want to know if anyone would _actually_ be broken by this change, rather than _possibly_.

btm:
Here's another fun *possibly*, could someone be depending on reconfigure running in the middle of their first chef run?

lamont-granquist:
like the daemon gets the old config and the child run gets the new, then you go back to the daemon with the old config?

btm:
Like I bootstrap with a minimal client config, and during my first run I
drop some ruby in there, and reconfigure, which is necessary for my run to
finish correctly.

But I realize now I'd just call reconfigure, I would use a signal.

btm:
If we always reconfigured there's a choice to make if we do it at the end of a run or the beginning of a run.

If you do it at the end of a run, you pick up any changes during the last run.

If you do it before a run, you are sure that you close and reopen log files before a run in case they've been rotated (the most common reason we get a HUP I think).

So we might want to still do both. Reconfigure at the end of every run, and reconfigure if we get a HUP while sleeping.

lib/chef/application/client.rb
```diff
@@ -474,31 +481,33 @@ def sleep_then_run_chef_client(sleep_sec)
...
+        reconfigure
+        interval_sleep(sleep)
+      end
     else
-      # Windows
       sleep(sec)
     end
   end
 
-  def client_sleep(sec)
```
btm: I merged this back into `interval_sleep` to have one less method with 'sleep' in it's name that wasn't very descriptive.

lib/chef/application/client.rb
btm: We didn't use `@signal` anywhere else so I moved all the signal code here into `interval_sleep` and added an explicit line for `IMMEDIATE_RUN` for clarity.

btm:
> i suspect you need to copypasta everything from client.rb to solo.rb as well still.

I'm going to leave HUP as is in `application.rb` for solo, as it lacks any of the pattern for queueing signals that the client has. We could add that to solo if/when.

I've pushed a new commit that runs `reconfigure` at the end of every run.

btm:
I can't find an un-horrific way way to add unit tests for hup behavior because of forking the process to be able to signal it.

lamont-granquist:
yeah i wouldn't overly worry about that, those kinds of tests are very hard to write

NimishaS:
@btm, I tested this PR. It's working fine.

btm:
@chef/client-maintainers re-review?

btm:
No responses to the [discourse post](https://discourse.chef.io/t/always-reconfigure-between-runs-when-daemonized/10931). I also tried pinging @jaymzh directly on slack but didn't hear back.

Issue #6128, opened 2017-05-06
"NoMethodError: private method `select' called for nil:NilClass" on downgrading chef-client from 13 to 12


akitada:
## Description

Chef installation fails with "NoMethodError: private method `select' called for nil:NilClass" on downgrading chef-client from 13 to 12.

## Chef Version

12.20.3

## Platform Version

`bento/centos-7.3`

## Replication Case

```
$ cat > Vagrantfile <<EOF
Vagrant.configure("2") do |config|
  config.vm.box = "bento/centos-7.3"
  config.vm.provision "chef_zero" do |chef|
    chef.nodes_path = "."
    chef.run_list = ["recipe[chef_client_updater::default]"]
    chef.json = {
      "chef_client_updater": {
        "prevent_downgrade": false,
        "version": "12.20.3",
        "post_install_action": "exec"
      }
    }
  end
end
EOF
$ mkdir cookboooks
$ git clone https://github.com/chef-cookbooks/chef_client_updater.git cookbooks/chef_client_updater
$ git clone https://github.com/chef-cookbooks/compat_resource.git cookbooks/compat_resource
$ vagrant up
```

## Client Output

```
$ vagrant up
...
==> default: [2017-05-06T15:40:52+00:00] INFO: Shutting down local-mode server.
==> default: [2017-05-06T15:40:52+00:00] WARN: Replacing ourselves with the new version of Chef to continue the run.
==> default: [2017-05-06T15:40:53+00:00] INFO: Started chef-zero at chefzero://localhost:8889 with repository at /tmp/vagrant-chef/d2feca91794a404d32f51c99f65d5346, /tmp/vagrant-chef
==> default:   One version per cookbook
==> default:   nodes at /tmp/vagrant-chef/3ef9d08fbcd57be47090d1e5ad4abad1/nodes
==> default: 

==> default: [2017-05-06T15:40:53+00:00] INFO: Forking chef instance to converge...
==> default: Starting Chef Client, version 12.20.3
==> default: [2017-05-06T15:40:53+00:00] INFO: *** Chef 12.20.3 ***
==> default: [2017-05-06T15:40:53+00:00] INFO: Platform: x86_64-linux
==> default: [2017-05-06T15:40:53+00:00] INFO: Chef-client pid: 4698
==> default: [2017-05-06T15:40:55+00:00] INFO: Setting the run_list to ["recipe[chef_client_updater::default]"] from CLI options
[...]

PR #6129, opened 2017-05-07
Discard caches for files not in the manifest

akitada:
### Description

cookbook_segment returns nil if the cache is not found in the manifest.
This can happen when chef client 12 runs on a node where chef client 13
generated cookbook cache files.

Currently, chef client fails in this case. This patch fixes it.

It should be safe to remove those cache files because they are not
cached in Chef 12, after all.

### Issues Resolved

https://github.com/chef/chef/issues/6128

### Check List

- [x] All tests pass


akitada:
This pull request is for Chef 12 only since Chef 13 does not have this problem.

thommay:
@chef/maintainers

akitada:
I've added tests and confirmed all unit tests pass by running `bundle exec rspec spec/unit`

PR #6142, opened 2017-05-11
[MSYS-563] Modified error message to explain connectivity issue due to Proxy

NimishaS:


NimishaS:
@btm, please review. This is for MSYS-563.

lib/chef/knife.rb
```diff
@@ -486,7 +486,8 @@ def humanize_http_exception(e)
         ui.error "Failed to authenticate to #{server_url} as #{username} with key #{api_key}"
         ui.info "Response:  #{format_rest_error(response)}"
       when Net::HTTPForbidden
-        ui.error "You authenticated successfully to #{server_url} as #{username} but you are not authorized for this action"
+        ui.error "You authenticated successfully to #{server_url} as #{username} but you are not authorized for this action.
+        This is possibly a connectivity issue. Please check your proxy settings if any."
```
btm: I'd make a separate error

```
proxy_env_vars = ENV.to_hash.keys.map(&:downcase) & [ "http_proxy", "https_proxy", "ftp_proxy", "socks_proxy", "no_proxy" ]
unless proxy_env_vars.empty?
  ui.error "There are proxy servers configured, your Chef server may need to be added to NO_PROXY."
end
```

The related 403 unit test in `spec/unit/knife_spec.rb` should be duplicated to have one that matches this message if HTTP_PROXY is set as an environment variable.

PR #6145, opened 2017-05-12
Depth Enhancements

joshburt:
**Depth Enhancements**

**Abstract**
Git resource only honors 'depth' attribute when performing an initial clone.
Git resource is unable to recover gracefully when depth is specified, but the target ref is outside the depth.

**Summary**
This PR adds support for shallow fetches, and automatically increasing the depth when the ref is outside the specified depth.

If the requested ref is outside the specified depth the client will deepen the shallow clone/fetch until enough histroy is avaiable to complete the requested operation.
There is a hard limit on the additional depth allowed (50) in order to prevent situtations involving infinite looops or overflows.

The deep is increased using a fibonacci scale:
[User specified depth] + [1, 2, 3, 5, 8, 13, 21, ... ]

**Example**
If we specify a depth of 1 on the resource:

```
git 'my stuff' do
  depth 1
  action :sync
end
```

and are unable to complete the checkout, we will see depth increase as follows up to the limit:

(initial run) 1
(initial run) 1 + 1 = 2
(initial run) 1 + 2 = 3
(initial run) 1 + 3 = 4
(initial run) 1 + 5 = 6
(initial run) 1 + 8 = 9
(initial run) 1 + 13 = 14
(initial run) 1 + 21 = 22
...

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -162,7 +162,7 @@ class Chef
             current_depth = 1
             last_depth = 0
             depth = new_resource.depth
-            while (exit_code != 0 && current_depth < 50)
+            while exit_code != 0 && current_depth < 50
               git_fetch_result = git("fetch", new_resource.remote, "--depth", depth.to_s, cwd: cwd, returns: [0, 1])
               git_fetch_tags_result = git("fetch", new_resource.remote, "--tags", "--depth", depth.to_s, cwd: cwd, returns: [0, 1])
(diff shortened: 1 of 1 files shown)


joshburt:
It looks like I'm missing the DCO portion of this request.  I'll address this and resubmit.  Thanks everyone! :)

joshburt:
I'll resubmit with the DCO requirements in place.  Thanks!

coderanger:
@joshburt You can just force push to your branch after running `dco sign`.

joshburt:
@coderanger  oooh!  Thanks, I've already redone this PR but will definetly use that for next time. :)

PR #6147, opened 2017-05-12
Wipe release notes and add new deprecations

tas50:


thommay:
I'm not sure the targeted for chef 15 deprecations should be here.

PR #6149, opened 2017-05-12
Git Resource 'Shallow' Enhancements

joshburt:
**Depth Enhancements**

**Abstract**
Git resource only honors 'depth' attribute when performing an initial clone.
Git resource is unable to recover gracefully when depth is specified, but the target ref is outside the depth.

**Summary**
This PR adds support for shallow fetches, and automatically increasing the depth when the ref is outside the specified depth.

If the requested ref is outside the specified depth the client will deepen the shallow clone/fetch until enough histroy is avaiable to complete the requested operation.
There is a hard limit on the additional depth allowed (50) in order to prevent situtations involving infinite looops or overflows.

The deep is increased using a fibonacci scale:
[User specified depth] + [1, 2, 3, 5, 8, 13, 21, ... ]

**Example**
If we specify a depth of 1 on the resource:

```
git 'my stuff' do
  depth 1
  action :sync
end
```

and are unable to complete the checkout, we will see depth increase as follows up to the limit:

(initial run) 1
(initial run) 1 + 1 = 2
(initial run) 1 + 2 = 3
(initial run) 1 + 3 = 4
(initial run) 1 + 5 = 6
(initial run) 1 + 8 = 9
(initial run) 1 + 13 = 14
(initial run) 1 + 21 = 22
...


coderanger:
This makes me nervous as a structural thing. If a user requests a depth, we should probably honor that or error clearly.

joshburt:
@coderanger you have a valid point. :)  

We use this for specific instances where both uploadarchive.allowUnreachable and uploadpack.allowReachableSHA1InWant are enabled on the git server.  This allows pulls/fetches using any commit id not just advertised references.  This type of un-shallowing is also used within some builds systems for similar conditions.

If it would be better to explicitly fail under this condition for the largest group of consumers then we probably shouldn’t change the behavior.  If we think it would be useful and warrants the addition or change to the attributes/actions on the resource I can certainly provide that as well.

Let me know either way what your opinion is on this.  We have this code as a stand-alone cookbook and leverage the DSL that way, so if this PR isn’t accepted it isn’t the end of the world.  I just wanted to provide the community with an opportunity to have this functionality available natively if desired. :)

Thanks!

lamont-granquist:
yeah i think i agree with @coderanger that increasing the depth to find the ref is too much magic and if the user gives us a ref that is outside of the depth we should bomb.

thommay:
Hey @joshburt we just came back around on this one (sorry!), and agree with @coderanger this is too much magic to support in the core resource.
Thanks for the PR, and sorry about how long this has been open!

lamont-granquist:
(also all those files that conflict just need to not get updated, it all gets autogenerated via processes)

PR #6152, opened 2017-05-13
🚧 Move Chef docs in to the Chef repo

coderanger:
So this is (hopefully obviously) still a work in progress, but I want to get a place for conversation ASAP. Basically I think we need to split up chef-web-docs and put each project's documentation in their own repo. So Chef's docs go here, Chef Server's docs go in chef/chef-server, Ohai in chef/ohai, etc. For internal stuff within Chef Software I leave it to the company to work out where best to store things, could remain in a single, shared repo or could be broken up to the various internal repos.

With that out of the way: what now? We can look at hosting each project effectively independently (possibly via Read The Docs) or use some kind of submodule or build script system to combine all the doc sets and build them into a single HTML site. Sphinx has a system called Intersphinx for linking between isolated doc sets so we can leverage that at least somewhat. This will also probably be the death of the dtags system, but I'm pretty okay with that as I think it does more harm than good at this point.

And with that, discuss.


davidwrede:
Hey @coderanger, what's the fundamental problem (or problems) you're trying to solve with such a change? Is pulling the OSS docs out of chef-web-docs the only way to achieve it/them?

As a member of the doc team, I'm strongly in favor of keeping the docs together and feel having the OSS content split into their respective repos would lead to a poor user experience overall. Scenario content that crosses multiple technologies, having a common IA, and keeping one place for people to submit doc updates are a few reasons that I have for keeping all of the content in one repo.

Instead of having this conversation in PR comments, I'm open to discussing this through zoom with folks, in-person at ChefConf, or in some other fashion.

tas50:
@coderanger Do we want to keep this one open?

coderanger:
Probably not, I'm unthrilled that there continues to be no motion on this topic after months but I did what I could.

Issue #6103, opened 2017-04-28
Chef 13.0 regression in gem_package related to finding gem binaries


lamont-granquist:
gem_package does a `which('gem')` more-or-less here:

https://github.com/chef/chef/blob/master/lib/chef/provider/package/rubygems.rb#L412-L418

that breaks because i removed path_sanity from the global `ENV['PATH']` in Chef-13 and moved it to shell_out injecting its own `ENV['PATH']` which covers most of the bases, but we have to address additional one-offs like this one by one.

we may be able to somewhat reuse the `which` helper here:

https://github.com/chef/chef/blob/master/lib/chef/mixin/which.rb

but that also does not replicate the old path_sanity.

the new path_sanity is here:

https://github.com/chef/chef/blob/master/lib/chef/mixin/path_sanity.rb

but that will be super not-appropriate to inject into the which helper as extra_paths because in chef-13 it was changed to stick the omnibus ruby binaries first, which completely defeats the entire purpose of the gem_package provider.

so we probably need to replicate the old behavior of path_sanity, and inject *that* into the which helper, but i'm still not 100% certain that's entirely correct.

its possible that after i have more coffee it will become apparent that this isn't a regression in core chef and cookbooks that depended upon the old behavior need to be updated -- i'm not entirely sure -- this is definitely a one-off use case though and we need to think through how gem_package is supposed to work fairly carefully...

lamont-granquist:
yeah, i think at a minimum we need to add the ruby bindirs from omnibus onto the tail of the path here so that if there's no system rubygems we fall back to omnibus.

i'm unsure if omitting /usr/local/bin is a bug or a feature in chef-13.  leaning towards not fixing that until there's a bug report.

thommay:
So adding `RbConfig::CONFIG['bindir']`to the `extra_path` supplied to `which` seems like it ought to be enough to mitigate this.

PR #6154, opened 2017-05-15
Ensure that we check the embedded gem binary last

thommay:
Previously we injected our expected paths in to ENV['PATH'] but we
stopped doing that in Chef 13; for gem_package, we need to ensure that
if all else fails we'll use the Omnibus provided gem binary, but we
should never pick that if we can find a different one.

Closes: #6103

cc @lamont-granquist @chef/maintainers


lamont-granquist:
travis+appveyor hate you but the code looks right

PR #6156, opened 2017-05-15
Prefer Systemd with sysvinit script over Upstart for service provider

### LINKED ISSUE #6076: Service provider for rpcbind detected as Upstart on Debian 8
## Description

Service resource detects provider as Chef::Provider::Service::Upstart for NFS service rpcbind on Debian 8. Issue is related to but not specific to or caused by (https://github.com/atomic-penguin/cookbook-nfs/issues/76). The NFS cookbook previously had service provider overrides based on platform (https://github.com/atomic-penguin/cookbook-nfs/commit/7b8b8c483119a52b3d3971dc29b8c01322a858c7). These were removed in v2.3.0 of the cookbook, deferring to the Chef provider helpers.

Unfortunately, this has exposed an issue with the helpers. The issue appears to be that because /etc/init/rpcbind.conf exists (https://packages.debian.org/jessie/amd64/rpcbind/filelist), Chef is incorrectly detecting the init provider as Upstart instead of Insserv. Removing the file results in the correct behaviour.

## Chef Version

Issue present on chef_version=13.0.118

## Platform Version

Debian 8.6

## Replication Case

Either run the nfs::default recipe (v2.3.1)

or

Install nfs-common package. 
Attempt to restart the rpcbind service using a recipe.
```
service 'rpcbind' do 
  action :restart
end
```
## Client Output & Stacktrace

Default behaviour & stacktrace reproduced in chef-shell
https://gist.github.com/gsreynolds/540a66658c99ec20358b04d40c4b8caf#file-1-upstart-detected

Same behaviour in chef-client using nfs::default
https://gist.github.com/gsreynolds/540a66658c99ec20358b04d40c4b8caf#file-2-chef-client-nfs-default

`rm /etc/init/rpcbind.conf` causes Chef to correctly identify provider and restart the service successfully:
https://gist.github.com/gsreynolds/540a66658c99ec20358b04d40c4b8caf#file-3-insserv-detected

Thanks in advance!

shortdudey123:
Signed-off-by: Grant Ridder <shortdudey123@gmail.com>

### Description

On newer Debian based systems that use systemd to have a package install a SysV init and Upstart script.  Systemd can use systemd service files or SysV init scripts.  However, on the Chef serivce provider side, this is not reflected.  So in this case, Upstart will be prefered even if Upstart itself is not installed on the system.

### Issues Resolved

Service resource failing to modify a service on a systemd based host that has SysV init and Upstart scripts.

### Check List

- [x] All tests pass

closes #6076

### DIFF
--- a/lib/chef/platform/service_helpers.rb
+++ b/lib/chef/platform/service_helpers.rb
@@ -69,7 +69,7 @@ class Chef
           configs = []
 
           if ::File.exist?(Chef.path_to("/etc/init.d/#{service_name}"))
-            configs << :initd
+            configs += [ :initd, :systemd ]
           end
 
           if ::File.exist?(Chef.path_to("/etc/init/#{service_name}.conf"))
--- a/spec/unit/provider_resolver_spec.rb
+++ b/spec/unit/provider_resolver_spec.rb
@@ -192,34 +192,34 @@ describe Chef::ProviderResolver do
           stub_service_providers(:debian, :invokercd, :upstart, :systemd)
         end
 
-        it "when only the SysV init script exists, it returns a Service::Debian provider" do
+        it "when both the SysV init and Systemd script exists, it returns a Service::Debian provider" do
           stub_service_configs(:initd, :systemd)
           expect(resolved_provider).to eql(Chef::Provider::Service::Systemd)
         end
 
-        it "when both SysV and Upstart scripts exist, it returns a Service::Upstart provider" do
+        it "when SysV, Upstart, and Systemd scripts exist, it returns a Service::Systemd provider" do
           stub_service_configs(:initd, :upstart, :systemd)
           expect(resolved_provider).to eql(Chef::Provider::Service::Systemd)
         end
 
-        it "when only the Upstart script exists, it returns a Service::Upstart provider" do
+        it "when both the Upstart and Systemd scripts exists, it returns a Service::Systemd provider" do
           stub_service_configs(:upstart, :systemd)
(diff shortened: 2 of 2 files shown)


lamont-granquist:
looks correct.  it looks like we were testing the broken behavior so existing tests just need to be changed to report Systemd correctly, and the use cases for systemd+everything else at the same time all look right with everything defaulting to systemd except in the case where only an upstart script exists (which is probably not going to work anyways, but its certainly the best we could ever do and we can let the OS fail in that case...)

tas50:
This is a super frustrating bug that I'd love to have fixed.

lamont-granquist:
i could have sworn that the fixed behavior was the way it worked all along...

lamont-granquist:
yeah in 12.4.0 we had this test:

```ruby
    shared_examples_for "an ubuntu platform with upstart, update-rc.d and systemd" do
      before do
        stub_service_providers(:debian, :invokercd, :upstart, :systemd)
      end

      it "when only the SysV init script exists, it returns a Service::Debian provider" do
        allow(Chef::Platform::ServiceHelpers).to receive(:config_for_service).with("ntp")
          .and_return( [ :initd, :systemd ] )
        expect(resolved_provider).to eql(Chef::Provider::Service::Systemd)
      end
```

so this broke in 12.5.1

shortdudey123:
Looks like it was actually always broken with the new solver
https://github.com/chef/chef/commit/97aaf5bbcdfd0810722b123bdc67e883a7ca8077#diff-78041db52b95a34ebfc3d06c264438a3R80 in `12.0.0.rc.0`

PR #6168, opened 2017-05-22
Make gem_installer generate a valid Gemfile

oclaussen:
### Description

When multiple cookbooks in the run list require the same gem from the gem_installer, and these gem declarations include keyword arguments (e.g. to install the gem from a git repository), the gem_installer will generate an invalid Gemfile.

This should fix this by handling keyword arguments separately and merging them together when necessary. This may still result in unwanted behaviour, if the cookbooks specify different arguments that do not play well together, but at least we will have a valid Gemfile in all cases.

### Issues Resolved

None?

### Check List

- [X] All tests pass

### DIFF
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -43,7 +43,7 @@ class Chef
             if cookbook_gems[args.first].last.is_a?(Hash)
               args << {} unless args.last.is_a?(Hash)
               args.last.merge!(cookbook_gems[args.first].pop) do |key, v1, v2|
-                Chef::Log.warn "Conflicting requirements for gem '#{args.first}', will use #{key.inspect} => #{v2.inspect}" if v1 != v2
+                raise Chef::Exceptions::GemRequirementConflict.new(args.first, key, v1, v2) if v1 != v2
                 v2
               end
             end
--- a/lib/chef/cookbook/gem_installer.rb
+++ b/lib/chef/cookbook/gem_installer.rb
@@ -40,6 +40,13 @@ class Chef
 
         cookbook_collection.each do |cookbook_name, cookbook_version|
           cookbook_version.metadata.gems.each do |args|
+            if cookbook_gems[args.first].last.is_a?(Hash)
+              args << {} unless args.last.is_a?(Hash)
+              args.last.merge!(cookbook_gems[args.first].pop) do |key, v1, v2|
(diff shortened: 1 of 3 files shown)


coderanger:
I would rather we just leave it as invalid and surface the bundler error to the user. It's just like you've requested an invalid cookbook dependency, should be a fatal error not a warning.

oclaussen:
Unfortunately, it would be totally valid that two cookbooks want to install the same gem with the same options (e.g. install from the same git repo), but currently it would produce an invalid Gemfile in that case.

coderanger:
We should check that the values are "compatible" then, i.e. make sure it's still an error if two cookbooks do actually request differing versions of the same gem and whatnot.

oclaussen:
I changed the request to fail directly whenever we would have directly conflicting requirements. Further checking if values are "compatible" might be quite complicated and is probably better done by Bundler.

lamont-granquist:
if you have two really identical gem declarations i think you should just drop the duplicate statement.  i think that would address the use case of what happens when two cookbook both want the same gem pulled from the same git source.

the problem that i see is `github` and `git` can still get merged and produce nonsense, and there may be other ways that merging flags produces dangerously successful nonsense, where bundler doesn't fail but it doesn't produce results that anyone would ever want.

i think i'd prefer to just de-dup based on strictly the options hashes being equivalent and raise otherwise.

lamont-granquist:
And for internal use if you're installing gems across multiple cookbooks then all your gem statements should be synchronized.  If you have 50 identical gem statements across 50 cookbooks then you should really consolidate those into one cookbook that expresses that gem dependency (or one cookbook that manages all your gem dependencies) and make all the other depend on that.

For community cookbooks, those should generally not use any arguments and just install the gem, or set a version.  To keep those in synch, while allowing pulling from a git tag (that also meets the requirements) we could merge only the version requirements.

So you could build a coherent Gemfile out of:

```
gem "foo"
gem "foo", "~> 1.1"
gem "foo", "> 1.0"
gem "foo", git: "https://github.com/foo/foo", tag: "bugfix1"
```

merging that into:

```
gem "foo", "~> 1.1",  "> 1.0", git: "https://github.com/foo/foo", tag: "bugfix1"
```

any non-version-constraint options should strictly conflict though even before we fire it off to bundler.

lamont-granquist:
yeah this is better but not perfect, i'm cool with merging, still think eventually it should get fixed better.

thommay:
Let's hold off merging this until we get the 14.1 hotfix out the door

PR #6169, opened 2017-05-23
Make Chef::VERSION be fancy

coderanger:
This is because if I have to write `Gem::Requirement.create('~> 12.0').satisfied_by?(Gem::Version.create(Chef::VERSION))` one more time I'm going to scream. Paradoxically because it's a version check API we won't actually be able to use it for a long while, but might as well start now.

This allows stuff like `Chef::VERSION =~ '~> 12.0'` or `case Chef::VERSION; when '>= 12.5' ...`.


coderanger:
Might also be cool to do this for `node['platform_version']`, in the Ohai loading we could do something like `automatic['platform_version'] = VersionString.new(automatic['platform_version'])`.

coderanger:
The `===` hack might be too much, would be cool if everyone else wants to just do the fancy `=~` operator and skip the `case` magic.

lamont-granquist:
yeah probably not the monkeypatching of `===`

coderanger:
Okay, the `===` trickery is gone and chef-config is no longer wrecked :)

thommay:
chefstyle needs care and feeding

coderanger:
And now fingers crossed that this doesn't some how jam the bumpbot :)

PR #6170, opened 2017-05-23
Update release notes for 13.1

tas50:
Just add the single ohai deprecation


shortdudey123:
Should https://github.com/chef/chef/pull/6156 get added too? i missed adding it to the release notes in my PR

PR #6175, opened 2017-05-25
Add `--allow-downgrades` to all `apt-get` commands

### LINKED ISSUE #6095: Support allow_downgrade in apt_package
## Description

The package resource has an `allow_downgrade` property that is supported for dnf/rpm/yum. Newer versions of Apt (1.1+) support this functionality along with a couple of others.

```
       --allow-downgrades
           This is a dangerous option that will cause apt to continue without prompting if it is doing downgrades. It should not be used except in very special situations. Using it can potentially destroy your system! Configuration Item:
           APT::Get::allow-downgrades. Introduced in APT 1.1.

       --allow-remove-essential
           Force yes; this is a dangerous option that will cause apt to continue without prompting if it is removing essentials. It should not be used except in very special situations. Using it can potentially destroy your system! Configuration Item:
           APT::Get::allow-remove-essential. Introduced in APT 1.1.

       --allow-change-held-packages
           Force yes; this is a dangerous option that will cause apt to continue without prompting if it is changing held packages. It should not be used except in very special situations. Using it can potentially destroy your system! Configuration Item:
           APT::Get::allow-change-held-packages. Introduced in APT 1.1.
```

I'm not sure if this can be done without causing backwards compatibility issues. I'll contribute a PR for this if I could be pointed towards a workable method to do it. Thanks!

## Chef Version

13.0.118

## Platform Version

Ubuntu 16.04

thomas88:
### Description

Adds `--allow-downgrades` to every apt command, as discussed in #6095.

Caution, `man apt-get` says this:

       --allow-downgrades
           This is a dangerous option that will cause apt to continue without prompting if it is doing downgrades. It should not be used except in very special situations.
           Using it can potentially destroy your system! Configuration Item: APT::Get::allow-downgrades. Introduced in APT 1.1.

Let me know what you think.

### Issues Resolved

resolves #6095

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -86,7 +86,7 @@ class Chef
           package_name = name.zip(version).map do |n, v|
             package_data[n][:virtual] ? n : "#{n}=#{v}"
           end
-          run_noninteractive("apt-get", "-q", "-y", default_release_options, options, "install", package_name)
+          run_noninteractive("apt-get", "-q", "-y", "--allow-downgrades", default_release_options, options, "install", package_name)
         end
 
         def upgrade_package(name, version)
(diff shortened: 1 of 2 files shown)


thommay:
This blows up 14.04 because allow-downgrades is not supported there, so you'll need to figure out the first versions of ubuntu and debian that can safely support this and conditionally add the flag

curiositycasualty:
I'd also like to see separate unit test(s) for the downgrade use-case. Possibly even OS sensitive?

tas50:
It looks like someone needs to the do the full investigation to figure out which versions of apt support allow-downgrades so we can get a solution that would allow us to apply allow-downgrades where supported and not break the older distributions. I'm going to close this out and this point since we're going to need a more robust solution here.

PR #6176, opened 2017-05-25
convert breakpoint resource to a custom resource

lamont-granquist:
mildly pointless, but it is a core resource converted to custom resource style with an action class and tests.


lamont-granquist:
@chef/client-core reviewage (lets not think too hard about the breakpoint resource itself, i just picked it because its the shortest bit of provider code out there)

PR #6178, opened 2017-05-25
Missed this which makes some smoke tests for chef-shell fail.

coderanger:
Whoops.


lamont-granquist:
omnibus builds are already in pain so merging to fix those

PR #6191, opened 2017-05-31
Add release notes related to SIGHUP changes [ci skip]

btm:
For https://github.com/chef/chef/pull/6122


tas50:
@btm you can nuke the old notes if you want. They're up on the docs site now

PR #6192, opened 2017-06-01
Remove bundle tasks, version_policy.rb and update README

lamont-granquist:
Per slack discussion this:

- removes tasks/bundle.rb
- removes version_policy.rb
- removes `rake dependencies:update_omnibus_overrides`
- omnibus_overrides.rb is now hand-managed, just edit the thing and go.

I also took a swing and updating the README.md with a focus on tasks actually useful to most users who would be reading the README and wanting to get shit done with a focus on brevity.

- throw away sentences like "Of special mention is the software we use to build omnibus itself." have been largely discarded with extreme prejudice.
- tried to put information users actually need to get things done front and center
- whacked a lot of philosophical yappage that was mostly no longer accurate or were moving targets (if anyone really cares deeply about how the omnibus software and project definitions are constructed they should investigate the
  sources, and if the sources aren't clear we should comment/document was is going on there).
- there's a bit in the middle on version bumping and regressions that i didn't touch, but still seems very overly wordy.

The sort of TL;DR is that we're back to basically using everything normally.  If you were familiar with the build process of about a year ago (forget all the crazy stuff of the past 8 months) the major changes from then would be:

- we have a Gemfile.lock checked in and other things are generally pinned via Gemfile.lock and/or Berksfile.lock
- the Gemfile.lock uses `bundle lock --platform` to lock windows stuff.  if you blow it away you'd need to re-add the windows gems.   this is all done doing normal bundler tooling.  bundler will complain at you with yellow text if you forget.
- `rake dependencies:update_gemfile_lock` handles the bundle lock'ing of the windows gems for you (you can `bundle update` against an existing Gemfile.lock just fine though -- again, this is just convenience around standard tooling).
- the `omnibus_overrides.rb` file is added which just externalizes the omnibus pins into the root directory.
- `.travis.yml` and `appveyor.yml` read `omnibus_overrides.rb` to pin bundler + rubygems against the versions we ship in omnibus so we test a bit more accurately.
- `rake dependencies` updates most of the locked stuff
- `omnibus_overrides.rb` needs to be updated by hand
- just use tools like `bundle outdated` directly to check if gems are outdated.


omnibus_overrides.rb
```diff
@@ -1,4 +1,6 @@
-# DO NOT EDIT. Generated by "rake dependencies". Edit version_policy.rb instead.
+# THIS IS NOW HAND MANAGED, JUST EDIT THE THING
```
thommay: :joy:

lamont-granquist:
I see a PR and I want it painted red

README.md
```diff
@@ -24,51 +24,55 @@ If you just want to use Chef, check out these resources:
...
-
-# Install dependencies with bundler
 bundle install
-
-# Build a gem
 bundle exec rake gem
-
-# Install the gem you just built
-gem install pkg/chef-VERSION.gem
+budnle exec rake install
```
tas50: typo on bundle

PR #6193, opened 2017-06-02
[MSYS-586] Allow windows_task create action to update tasks.

harikesh-kolekar:

### Description
Since `windows_task` moved into core, the expectation is that the `:create` action can take the place of the old `:change` action (since it has the same attributes).  In fact, the plan sounds like that `:change` action can just be aliased to `:create` to prevent legacy breakage.  However, the `:create` action will break if the `command` attribute is not specified (like it would be in a preexisting `:change` action): ` undefined method `tr' for nil:NilClass`
*Expected Results*:  The windows_task resource can be modified by resource name without having to specify the command

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -52,53 +52,74 @@ class Chef
         set_current_idle_time(task_hash[:idle_time]) if task_hash[:idle_time]
         @current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
         @current_resource.execution_time_limit(task_hash[:execution_time_limit]) if task_hash[:execution_time_limit]
-
         @current_resource.status = :running if task_hash[:Status] == "Running"
         @current_resource.enabled = true if task_hash[:ScheduledTaskState] == "Enabled"
       end
 
+      # This method checks if task and command attributes exist since those two are mandatory attributes to create a schedules task.
+      def basic_validation
+        validate = []
+        validate << "Command" if new_resource.command.nil? || new_resource.command.empty?
+        validate << "Task Name" if new_resource.task_name.nil? || new_resource.task_name.empty?
+        return true if validate.empty?
+        raise Chef::Exceptions::ValidationFailed.new "Value for '#{validate.join(', ')}' option cannot be empty"
+      end
+
+      # get array of windows task resource attributes
+      def resource_attributes
+        %w{ command user run_level cwd frequency_modifier frequency idle_time random_delay execution_time_limit start_day start_time }
+      end
+
       def action_create
-        if @current_resource.exists && !(task_need_update? || @new_resource.force)
-          Chef::Log.info "#{@new_resource} task already exists - nothing to do"
-        else
-          options = {}
-          options["F"] = "" if @new_resource.force || task_need_update?
-          options["SC"] = schedule
-          options["MO"] = @new_resource.frequency_modifier if frequency_modifier_allowed
-          options["I"]  = @new_resource.idle_time unless @new_resource.idle_time.nil?
-          options["SD"] = @new_resource.start_day unless @new_resource.start_day.nil?
-          options["ST"] = @new_resource.start_time unless @new_resource.start_time.nil?
-          options["TR"] = @new_resource.command
-          options["RU"] = @new_resource.user
-          options["RP"] = @new_resource.password if use_password?
-          options["RL"] = "HIGHEST" if @new_resource.run_level == :highest
-          options["IT"] = "" if @new_resource.interactive_enabled
-          options["D"] = @new_resource.day if @new_resource.day
-          options["M"] = @new_resource.months unless @new_resource.months.nil?
-
-          run_schtasks "CREATE", options
-          xml_options = []
-          xml_options << "cwd" if new_resource.cwd
-          xml_options << "random_delay" if new_resource.random_delay
-          xml_options << "execution_time_limit" if new_resource.execution_time_limit
-          update_task_xml(xml_options) unless xml_options.empty?
-
-          new_resource.updated_by_last_action true
-          Chef::Log.info "#{@new_resource} task created"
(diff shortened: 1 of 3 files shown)


lib/chef/provider/windows_task.rb
```diff
@@ -52,40 +52,58 @@ def set_current_resource(task_hash)
         set_current_idle_time(task_hash[:idle_time]) if task_hash[:idle_time]
         @current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
         @current_resource.execution_time_limit(task_hash[:execution_time_limit]) if task_hash[:execution_time_limit]
-
         @current_resource.status = :running if task_hash[:Status] == "Running"
         @current_resource.enabled = true if task_hash[:ScheduledTaskState] == "Enabled"
+        @current_resource.start_time = task_hash[:StartTime] if task_hash[:StartTime]
+        @current_resource.start_day = task_hash[:StartDate] if task_hash[:StartDate]
```
NimishaS: Since we are planning to do the fix for `start_day` and `start_time` in a separate PR, please remove this change.

lib/chef/provider/windows_task.rb
```diff
@@ -52,40 +52,58 @@ def set_current_resource(task_hash)
...
-          options["D"] = @new_resource.day if @new_resource.day
-          options["M"] = @new_resource.months unless @new_resource.months.nil?
-
-          run_schtasks "CREATE", options
-          xml_options = []
-          xml_options << "cwd" if new_resource.cwd
-          xml_options << "random_delay" if new_resource.random_delay
-          xml_options << "execution_time_limit" if new_resource.execution_time_limit
-          update_task_xml(xml_options) unless xml_options.empty?
+      def check_validation
```
NimishaS: The name of this method can be modified as `basic_validation` and it would be good to add a comment explaining it's purpose:
`This method checks if `task` and `command` attributes exist since those 2 a mandatory attributes to create a schedules task.`

spec/unit/provider/windows_task_spec.rb
```diff
@@ -98,6 +98,8 @@
       expect(current_resource.frequency_modifier).to eq(15)
       expect(current_resource.execution_time_limit).to eq("PT72H")
       expect(current_resource.enabled).to be(true)
+      expect(current_resource.start_time).to eq("1:12:00 PM")
+      expect(current_resource.start_day).to eq("3/30/2017")
```
NimishaS: Please remove the specs related to `start_time` and `start_day`.

btm:
@harikesh-kolekar @NimishaS we need to figure out what is happening with appveyor here.

btm:
The Appveyor failure here looks like the one we're having everywhere on this project now. We'll investigate that as part of MSYS-606.

@chef/client-windows, review please?

lib/chef/provider/windows_task.rb
```diff
@@ -52,40 +52,58 @@ def set_current_resource(task_hash)
...
+          end
+          # To merge current resource and new resource attributes
+          temp_arry = %w{ command user run_level cwd frequency_modifier frequency idle_time random_delay execution_time_limit start_day start_time }
+          temp_arry.each do |attribute|
+            new_resource_attribute = @new_resource.send(attribute)
+            current_resource_attribute = @current_resource.send(attribute)
+            @new_resource.send("#{attribute}=", current_resource_attribute ) if current_resource_attribute && new_resource_attribute.nil?
+          end
         end
+        return unless basic_validation
```
stevendanna: Do we think this should raise an error since the user has asked us to create a resource but we can't?

lib/chef/provider/windows_task.rb
```diff
@@ -52,40 +52,58 @@ def set_current_resource(task_hash)
...
-          new_resource.updated_by_last_action true
-          Chef::Log.info "#{@new_resource} task created"
+      def action_create
+        if @current_resource.exists
+          if !(task_need_update? || @new_resource.force)
+            Chef::Log.info "#{@new_resource} task already exists - nothing to do"
+            return
+          end
+          # To merge current resource and new resource attributes
+          temp_arry = %w{ command user run_level cwd frequency_modifier frequency idle_time random_delay execution_time_limit start_day start_time }
```
stevendanna: It might make sense to extract this to a small function.

lib/chef/provider/windows_task.rb
```diff
@@ -52,40 +52,58 @@ def set_current_resource(task_hash)
...
+          temp_arry.each do |attribute|
+            new_resource_attribute = @new_resource.send(attribute)
+            current_resource_attribute = @current_resource.send(attribute)
+            @new_resource.send("#{attribute}=", current_resource_attribute ) if current_resource_attribute && new_resource_attribute.nil?
+          end
         end
+        return unless basic_validation
+        options = {}
+        options["F"] = "" if @new_resource.force || task_need_update?
```
stevendanna: I noticed we are still using `@new_resource` here rather than `new_resource`.  That is a feature of the existing code, but I wonder if it is something we should clean up while here.

btm: @harikesh-kolekar to be clear `attr_accessor :new_resource` was added to `Chef::Provider` last year, which we subclass here. So we can use `new_resource` instead. This makes `new_resource` accessible from outside the class without going out of our way.

lib/chef/provider/windows_task.rb
harikesh-kolekar: https://github.com/chef/chef/pull/6193/files/510ed6f74483ba213c4e6e368f9d512d83e96f6f#diff-208e6e972fb102d2fce9f211de0d94a8R65
The above line raise an error to user

stevendanna: By raise an error I meant raise an exception that will halt the execution.  While that line will emit an error to the log, execute will continue as if nothing was wrong unless I'm missing something?

PR #6196, opened 2017-06-04
Pass authentication for chocolatey_package

astoltz:
Signed-off-by: Andrew Stoltz <astoltz@gmx.com>

### Description

```ruby
chocolatey_package 'private-package' do
  source 'https://custom/nuget'
  options '-u username -p password'
end
```
Before PR, the authentication parameters don't get passed along to the "choco list" command causing failed authentication. The PR causes these parameters to be passed along.

I'm not sure if there are options that should be passed during "choco install" but not "choco list". If there are, we may need a separate property for this purpose.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -228,8 +228,7 @@ EOS
         def available_packages
           @available_packages ||=
             begin
-              cmd = [ "list -r #{package_name_array.join ' '}" ]
-              cmd.push( "-source #{new_resource.source}" ) if new_resource.source
+              cmd = [ "list -r #{package_name_array.join ' '}", cmd_args ]
               raw = parse_list_output(*cmd)
               raw.keys.each_with_object({}) do |name, available|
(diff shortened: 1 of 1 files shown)


coderanger:
For `python_package` I ended up with `options`, `list_options`, and `install_options`.

Another option for this in particular would be to allow putting credentials in the `source` URI.

mwrock:
I think this should be fine. Any unexpected args, `list` will warn but the exit code will remain 0. Just need to fix the broken test.

btm:
@astoltz would you please look at the failing tests and rebase this?

stuartpreston:
Hi @astoltz - thanks for this submission.  As you can see the `chocolatey_package` has evolved since this PR was first created so your changes need a little bit of work.  Are you still in a position to make the requested rebase/fix the tests?  If not then we'll close this and open an issue to track and solve the original problem as I don't see one.  Thanks!

astoltz:
I'm no longer in a position that requires authenticated access to a NuGet repository.

PR #6203, opened 2017-06-08
Suppress nested causes for sensitive exceptions

### LINKED ISSUE #3903: Stdout from sensitive execute resource displays with chef-client -l debug
I have a sensitive execute block in one of my cookbooks. Normally, running chef-client suppresses the output from this block, as it should, however if I run it with `-l debug`, stdout goes straight to my terminal:

```
# chef-client -l debug
Starting Chef Client, version 12.4.1
resolving cookbooks for run list: 
```

...

```
  * execute[my-execute-block] action run
```

<em>(tons of stuff that we shouldn't be able to see)</em>

<pre>    - execute sensitive resource
  * execute[next-thing] action run</pre>

This is undesirable ಠ_ಠ

I noted <a href="https://github.com/chef/chef/issues/3813">this issue</a>, which is similar, but in this case the execution was successful, and debug mode was the variation.

TheLunaticScripter:
This will prevent sensitive data from leaking into debug_stacktrace output.

Signed-off-by: John Snow <thelunaticscripter@outlook.com>

### Description

This will prevent sensitive data from leaking into debug_stacktrace output.

### Issues Resolved

#3903
#3813

### DIFF
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -58,9 +58,12 @@ class Chef
             shell_out!(command, opts)
           rescue Mixlib::ShellOut::ShellCommandFailed
             if sensitive?
-              $! = nil
-              raise Mixlib::ShellOut::ShellCommandFailed,
-                "Command execution failed. STDOUT/STDERR suppressed for sensitive resource"
+              ex = Mixlib::ShellOut::ShellCommandFailed.new("Command execution failed. STDOUT/STDERR suppressed for sensitive resource")
+              # Forcibly hide the exception cause chain here so we don't log the unredacted version
+              def ex.cause
+                nil
+              end
+              raise ex
             else
               raise
             end


TheLunaticScripter:
Test on the way but open for review.

coderanger:
What's the concrete use case here? The sensitive flag is about the data in the resource, we don't generally consider a trace back to be security relevant.

itmustbejj:
@coderanger It looks like this was the behavior before https://github.com/chef/chef/pull/3354 was merged.

I get that the sensitive flag is set the on the resource, but the reason it's set is so that secrets don't show up in logs. That behavior was true before the above mentioned PR. With this change, you get the best of both worlds.

lamont-granquist:
This is definitely the wrong way to fix this.  This is very brittle, very one-off, lacks tests and creates spooky-action-at-a-distance where this string in one file
randomly has effects on another file a very long and surprising distance away.

You can instead edit the message in the original exception object and then rethrow that so that you don't create an e.cause chain in the first place:

```ruby
#!/usr/bin/env ruby

begin
  begin
    raise "FOO"
  rescue Exception => e
    e.message[0..-1] =  "BAR"
    raise e
  end
rescue Exception => e
  puts e
  puts e.cause
end
```

That is mildly awful due to the in-place substring mangling, but localizes the damage at least.  There might be better in-place editing APIs.

Really needs a test as well.

Maybe @coderanger can come up with something better.

I initially considered suggesting creating a unique "Chef::Exceptions::SensitiveSuppressedException" and teach this code to stop if hit one of those.  But the idea of seeing that spread all over the codebase and change the exception return classes would be very poor API design.  I'm reasonably confident that the exception following code should not be changed here -- which led me to the suggestion of just hacking it up at the source.

coderanger:
Just so we're clear, the actual stack trace is not the issue, right? It's that some resource properties are ending up in the cause string somehow? Because that seems like that part to fix.

coderanger:
Like if you put a password in the message string of an exception, we're not going to block that. And I don't think that the backtrace counts as sensitive since we log all the files being loaded by Chef at startup (at debug level, but still).

coderanger:
@TheLunaticScripter Can you give an example of output you think is incorrect (with the actual secret-y bits slightly redacted obviously)?

lamont-granquist:
The problem is that following the `e.cause` chain of exceptions defeats the purpose of this code entirely:

https://github.com/ranjib/chef/blob/master/lib/chef/provider/execute.rb#L59-L62

You can replicate with something along these lines:

```bash
#!/bin/sh

echo "ALL UR PAZZWURDZ"

exit -1
```

```ruby
#!/usr/bin/env chef-apply

execute "/Users/lamont/bin/fail" do
  sensitive true
end
```

coderanger:
`$! = nil` before the new raise should fix that :)

lamont-granquist:
much better than in-place editing of the original exception string...

TheLunaticScripter:
@lamont-granquist @coderanger @itmustbejj Thank you all for the help and input. I will add the tests and get that changed and resubmit.

lamont-granquist:
rspec is broken for relevant reasons:

```
expected Mixlib::ShellOut::ShellCommandFailed with message matching /suppressed for sensitive resource/, got #<NameError: $! is a read-only variable> 
```

coderanger:
Derp, I am fail in advice. Let me do some more research.

coderanger:
Okay, so this time with some actual testing rather than just assuming global variables work the way all other globals work:

```ruby
def oops
  raise RuntimeError.new("secret password")
end

def try_a_thing
  begin
    oops
  rescue Exception
    ex = RuntimeError.new("redacted")
    def ex.cause
      nil
    end
    raise ex
  end
end

def show_exceptions(ex)
  while ex
    puts ex
    ex = ex.cause
  end
end

begin
  try_a_thing
rescue Exception => e
  show_exceptions(e)
end
```

If you run that you should only see `redacted`. If you comment out the `def self.cause` bit you'll see both in the chain. This should work in basically the same way as fixing `$!` if a bit weirder.

coderanger:
I think I correctly figured out how to use my "write in to someone else's branch because I'm a contributor" powers, @TheLunaticScripter sorry if I messed up your repo in some way >_<

TheLunaticScripter:
@coderanger no worries big thank you to both you and @lamont-granquist for all your help.

lamont-granquist:
lol, yeah, i think one way or another gonna have to get our hands dirty here...

coderanger:
Not sure what's wrong with Appveyor.

TheLunaticScripter:
Looks like it hit the time out in Appveryor. If I'm not mistaken I still need to add test for this change correct?

lamont-granquist:
if you can?  the appveyor timeout is known issue.

PR #6204, opened 2017-06-08
Deprecate property namespace magic

### LINKED ISSUE #5974: Chef-13.1:  Deprecate magical custom resource namespace collision generating unhappysauce
right here:

https://github.com/chef/chef/blob/bae9118c612fa6d0f102b26fd3ba72cf612f7e60/lib/chef/provider.rb#L289

lamont-granquist:
This forces everyone to starting using new_resource.property_name
instead of just property_name.

closes #5974

### DIFF
--- a/lib/chef/dsl/declare_resource.rb
+++ b/lib/chef/dsl/declare_resource.rb
@@ -155,7 +155,15 @@ class Chef
       def edit_resource(type, name, created_at: nil, run_context: self.run_context, &resource_attrs_block)
         edit_resource!(type, name, created_at: created_at, run_context: run_context, &resource_attrs_block)
       rescue Chef::Exceptions::ResourceNotFound
-        declare_resource(type, name, created_at: created_at, run_context: run_context, &resource_attrs_block)
+        resource = declare_resource(type, name, created_at: created_at, run_context: run_context)
+        if resource_attrs_block
+          if defined?(new_resource)
+            resource.instance_exec(new_resource, &resource_attrs_block)
+          else
+            resource.instance_exec(&resource_attrs_block)
+          end
+        end
+        resource
       end
 
       # Lookup a resource in the resource collection by name.  If the resource is not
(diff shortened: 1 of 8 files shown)


lib/chef/dsl/declare_resource.rb
```diff
@@ -155,7 +155,15 @@ def edit_resource!(type, name, created_at: nil, run_context: self.run_context, &
       def edit_resource(type, name, created_at: nil, run_context: self.run_context, &resource_attrs_block)
         edit_resource!(type, name, created_at: created_at, run_context: run_context, &resource_attrs_block)
       rescue Chef::Exceptions::ResourceNotFound
-        declare_resource(type, name, created_at: created_at, run_context: run_context, &resource_attrs_block)
+        resource = declare_resource(type, name, created_at: created_at, run_context: run_context)
+        if resource_attrs_block
+          if defined?(new_resource)
+            resource.instance_exec(new_resource, &resource_attrs_block)
```
coderanger: So this would be:

```ruby
action :foo do
  some_resource 'asdf' do |instance_of_some_resource, parent_resource|
```

or something like that?

lib/chef/provider.rb
```diff
@@ -350,9 +350,11 @@ def #{name}(*args, &block)
                 #
```
coderanger: Should update the FIXME above this.

lib/chef/dsl/declare_resource.rb
lamont-granquist: this was a bug that was getting papered over by the resource magic.  if you look at edit_resource! above it has all the same instance_exec stuff but if that throws then we catch here and never do the instance_exec fixup.  What was happening before was that you'd get property lookups through method_missing on the resource being constructed that bubbled up to the enclosing_provider which then hit the magic on the provider to forward the property to the resource.  So we broke that, and that exposed that this code had never gotten properly written.

Since if you remove this code it blows up the integration tests for accumulators written with edit_resource, we have coverage for it already.

TL;DR:  my head hurts.

coderanger: Oh derp, right, I misread what method this was in. :+1:

lib/chef/provider.rb
coderanger: Resolved in commit below.

Issue #6165, opened 2017-05-18
Chomp errors when using Chocolatey provider


TonyChengCME:
## Description

When I attempt to install a Chocolatey package using the Chef::Provider::Package::Chocolatey provider, I am getting a "undefined method `chomp' for nil:NilClass" error during my attempt to converge.  I upgraded the client to 12.16.64 and was able to execute the recipe locally.

## Chef Version

Initially started at 12.12.15 and upgraded to 12.16.64

## Platform Version

Windows Server 2012 R2

## Replication Case

When I attempt to install a Chocolatey package using the Chef::Provider::Package::Chocolatey provider in client version 12.12.15,  I am getting a `undefined method 'chomp' for nil:NilClass` error during my attempt to converge.  I upgraded the client to 12.16.64 and was able to successfully execute the recipe locally (i.e., "chef-client -z").  I then ran a “choco uninstall java-platform-se-8" and added the JRE installation cookbook to the node’s run_list.  When I execute the chef-client.bat afterwards, it generated the `FATAL: NoMethodError: undefined method 'failed' for nil:NilClass` error again.

## Client Output
```
Recipe: cme_etpawin_javajre::install_javajre
  * chocolatey_package[java-platform-se-8] action install[2017-05-15T15:15:57-05:00] INFO: Processing chocolatey_package
[java-platform-se-8] action install (cme_etpawin_javajre::install_javajre line 1)
 
 
    ================================================================================
    Error executing action `install` on resource 'chocolatey_package[java-platform-se-8]'
    ================================================================================
 
    NoMethodError
    -------------
    undefined method `chomp' for nil:NilClass
 
    Resource Declaration:
    ---------------------
    # In C:/chef/cache/cookbooks/cme_etpawin_javajre/recipes/install_javajre.rb
 
      1: chocolatey_package 'java-platform-se-8' do
      2:   provider                   Chef::Provider::Package::Chocolatey
      3:   version                    '8.0.1210.13'
      4:   action                     :install # defaults to :install if not specified
      5: end
 
    Compiled Resource:
    ------------------
    # Declared in C:/chef/cache/cookbooks/cme_etpawin_javajre/recipes/install_javajre.rb:1:in `from_file'
 
    chocolatey_package("java-platform-se-8") do
      package_name ["java-platform-se-8"]
      provider Chef::Provider::Package::Chocolatey
      action [:install]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      declared_type :chocolatey_package
[...]

PR #6207, opened 2017-06-09
Fix chomp error for nil class

piyushawasthi:
wrt https://github.com/chef/chef/issues/6165

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -257,7 +257,7 @@ EOS
           choco_command(*args).stdout.each_line do |line|
             next if line.start_with?("Chocolatey v")
             name, version = line.split("|")
-            hash[name.downcase] = version.chomp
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@mwrock this looks correct?  once this is fixed, nil values as versions in the generated hash should turn into nil values in e.g. installed_version and that should be handled fine by the rest of the package provider machinery...

ferventcoder:
This may be incorrect upon a closer look - I'm guessing something is breaking due to a trial message showing through, even with `-r`.

ferventcoder:
But this could also be correct.
It may be better to discard any that don't have versions.

mwrock:
I think this is good. As @ferventcoder mentions, these error cases are likely because of some text completely unrelated to a individual package. We could also discard those items, but even if we don't they should not interfere with the valid packages in the hash.

lamont-granquist:
and now that we've dropped 2.2 was can finally use the ruby 2.3 safe navigation operators...

PR #6209, opened 2017-06-09
bump Gemfile.lock

lamont-granquist:
mostly want this for new chefstyle, nothing surprising here.


lamont-granquist:
@chef/client-core looks like this is going green

PR #6213, opened 2017-06-13
[MSYS-590]Make mount idempotent on Aix

NAshwini:

### Description

[Make mount idempotent on Aix]

### Issues Resolved

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/mount/aix.rb
+++ b/lib/chef/provider/mount/aix.rb
@@ -40,25 +40,32 @@ class Chef
           # Check to see if there is an entry in /etc/filesystems. Last entry for a volume wins. Using command "lsfs" to fetch entries.
           enabled = false
 
+          regex_arr = device_fstab_regex.split(":")
+          if regex_arr.size.eql?(2)
+            nodename = regex_arr[0]
+            devicename   = regex_arr[1]
+          else
+            devicename   = regex_arr[0]
+          end
           # lsfs o/p = #MountPoint:Device:Vfs:Nodename:Type:Size:Options:AutoMount:Acct
           # search only for current mount point
           shell_out("lsfs -c #{@new_resource.mount_point}").stdout.each_line do |line|
             case line
             when /^#\s/
               next
-            when /^#{Regexp.escape(@new_resource.mount_point)}:#{device_fstab_regex}:(\S+):(\[\S+\])?:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
+            when /^#{Regexp.escape(@new_resource.mount_point)}:#{devicename}:(\S+):#{nodename}:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
               # mount point entry with ipv6 address for nodename (ipv6 address use ':')
               enabled = true
               @current_resource.fstype($1)
-              @current_resource.options($5)
-              Chef::Log.debug("Found mount #{device_fstab} to #{@new_resource.mount_point} in /etc/filesystems")
+              @current_resource.options($4)
+              Chef::Log.debug("Found mount #{devicename}:#{nodename} to #{@new_resource.mount_point} in /etc/filesystems")
               next
-            when /^#{Regexp.escape(@new_resource.mount_point)}:#{device_fstab_regex}::(\S+):(\S+)?:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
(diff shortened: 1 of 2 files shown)


btm:
@NAshwini these tests were passing before, yeah? Can you add a test that would pass with this patch but not pass without it?

The AppVeyor/Travis failures looked unrelated. I've restarted them.

lib/chef/provider/mount/aix.rb
```diff
@@ -149,6 +180,18 @@ def enable_fs
           end
         end
 
+        def mount_options_unchanged?
+          current_resource_options = @current_resource.options
+          current_resource_options.pop
```
btm: what's this pop?

NAshwini: Hi @btm ,
 
It pop the extra `"rw"` which gets appended for options comming from file `/etc/filesystems`
 & which in return does not match with the options we are passing from the recipe.
Hence I am removing the last entry which is `"rw"` from the `current_resource_options` so it can match the exact entries from the `/etc/filesystems` file.

btm: It feels a little magical. a comment and something like `current_resource_options.delete_if { |x| x == "rw" }` might be more clear.

btm: The concern is that if something changed, and we start popping the wrong thing, it's going to be hard for someone else walking into this code to understand why it is there.

lib/chef/provider/mount/aix.rb
```diff
@@ -40,30 +40,61 @@ def enabled?
...
+              Chef::Log.debug("Found mount #{nodename} to #{@new_resource.mount_point} in /etc/filesystems")
+              next
+            when /^#{Regexp.escape(@new_resource.mount_point)}:(\S+)?:(\S+):#{devicename}:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
+              # mount point entry with hostname or ipv4 address
+              enabled = true
+              @current_resource.fstype($2)
               @current_resource.options($5)
-              Chef::Log.debug("Found mount #{device_fstab} to #{@new_resource.mount_point} in /etc/filesystems")
+              @current_resource.device(devicename)
+              @new_resource.device(nodename)
```
btm: Generally we shouldn't change new_resource, the user should. Can we do this without mutating new_resource?

lib/chef/provider/mount/aix.rb
```diff
@@ -40,30 +40,60 @@ def enabled?
...
               next
-            when /^#{Regexp.escape(@new_resource.mount_point)}:#{device_fstab_regex}:(\S+):(\[\S+\])?:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
+            when /^#{Regexp.escape(@new_resource.mount_point)}:#{devicename}:(\S+):#{nodename}:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
               # mount point entry with ipv6 address for nodename (ipv6 address use ':')
               enabled = true
               @current_resource.fstype($1)
-              @current_resource.options($5)
-              Chef::Log.debug("Found mount #{device_fstab} to #{@new_resource.mount_point} in /etc/filesystems")
+              @current_resource.options($4)
+              Chef::Log.debug("Found mount #{devicename}:#{nodename} to #{@new_resource.mount_point} in /etc/filesystems")
```
scotthain: Let's make all the debug lines consistent with line 90 and say `Found mount point ...`

lib/chef/provider/mount/aix.rb
```diff
@@ -40,30 +40,60 @@ def enabled?
...
               next
-            when /^#{Regexp.escape(@new_resource.mount_point)}:#{device_fstab_regex}::(\S+):(\S+)?:(\S+)?:(\S+):(\S+):(\S+):(\S+)/
+            when /^#{Regexp.escape(@new_resource.mount_point)}:#{nodename}:(\S+)::(\S+)?:(\S+):(\S+):(\S+):(\S+)/
               # mount point entry with hostname or ipv4 address
               enabled = true
               @current_resource.fstype($1)
+              @current_resource.options($4)
+              @current_resource.device("")
+              Chef::Log.debug("Found mount #{nodename} to #{@new_resource.mount_point} in /etc/filesystems")
```
scotthain: Same as above - `Found mount point`

lamont-granquist:
LGTM

PR #6216, opened 2017-06-15
Fixing build failures of Windows Server 2016

NimishaS:

### DIFF
--- a/spec/functional/mixin/powershell_out_spec.rb
+++ b/spec/functional/mixin/powershell_out_spec.rb
@@ -22,13 +22,13 @@ describe Chef::Mixin::PowershellOut, windows_only: true do
   include Chef::Mixin::PowershellOut
 
   describe "#powershell_out" do
-    context "for windows version less than 10", windows_10_or_2016: false do
+    context "for windows version less than 10", windows_lt_10: true do
       it "runs a powershell command and collects stdout" do
         expect(powershell_out("get-process").run_command.stdout).to match /Handles\s+NPM\(K\)\s+PM\(K\)\s+WS\(K\)\s+VM\(M\)\s+CPU\(s\)\s+Id\s+/
       end
     end
 
-    context "for windows version greater than 10", windows_10_or_2016: true do
+    context "for windows version greater than 10", windows_gte_10: true do
       it "runs a powershell command and collects stdout" do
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -149,6 +149,7 @@ module Shell
   config.filter_run_excluding :windows64_only => true unless windows64?
   config.filter_run_excluding :windows32_only => true unless windows32?
   config.filter_run_excluding :windows_nano_only => true unless windows_nano_server?
+  config.filter_run_excluding :windows_10_or_2016 => true unless windows_10_or_2016?
```
btm: For long term purposes, we probably mean 10 or newer? so would an more appropriate filter be something like `windows_gte_10` ?

spec/functional/resource/group_spec.rb
```diff
@@ -58,6 +58,7 @@ def group_should_not_exist(group)
     when "debian", "fedora", "rhel", "suse", "gentoo", "slackware", "arch"
       expect { Etc.getgrnam(group) }.to raise_error(ArgumentError, "can't find group for #{group}")
     when "windows"
+      sleep(5)
```
btm: if this works it would be preferable if we could add a check of some kind for readiness rather than adding 5 seconds to the tests.

spec/functional/mixin/powershell_out_spec.rb
```diff
@@ -22,8 +22,16 @@
...
   describe "#powershell_out" do
-    it "runs a powershell command and collects stdout" do
-      expect(powershell_out("get-process").run_command.stdout).to match /Handles\s+NPM\(K\)\s+PM\(K\)\s+WS\(K\)\s+VM\(M\)\s+CPU\(s\)\s+Id\s+/
+    context "for windows version less than 10", windows_10_or_2016: false do
+      it "runs a powershell command and collects stdout" do
+        expect(powershell_out("get-process").run_command.stdout).to match /Handles\s+NPM\(K\)\s+PM\(K\)\s+WS\(K\)\s+VM\(M\)\s+CPU\(s\)\s+Id\s+/
+      end
+    end
+
+    context "for windows version greater than 10", windows_10_or_2016: true do
```
lamont-granquist: this should probably be `windows_after_10_or_2016: true` and should be changed to something more like `Gem::Requirement.new(">= 10").satisfied_by?(Gem::Version(host_version))` (which handles `10.10 > 10.9` correctly, unlike `to_f`).

thommay:
I'll merge this as soon as the bump bot is playing ball

PR #6219, opened 2017-06-16
bump zlib to 1.2.11

thommay:
This is in response to CVEs:

  * CVE-2016-9841
  * CVE-2016-9842
  * CVE-2016-9843

@chef/maintainers


coderanger:
For the sake of tracking, the CVE fixes were in 1.2.9 but no reason to not go to latest while we're at it.

btm:
The zlib website does explicitly say to not use the intermediary versions
due to bugs as well.

PR #6222, opened 2017-06-16
Special case for the supports pseudo-property for Chef 12.

coderanger:
We could (and probably should) fix this in each cookbook but this has been the cause of enough issue spam that a special case seems warranted, especially since it is only needed on the 12.x branch and won't live on in 13.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -657,6 +657,10 @@ class Chef
       # as needed.
       return if Chef::Resource.properties.keys.include?(name)
 
+      # Special case for `supports` as it was moved in Chef 13 and this is causing
+      # some user confusion in cookbooks that need to support both 12 and 13.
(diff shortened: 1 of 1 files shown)


tas50:
It's hacky, but I think this will resolve a lot of pain. We advised customers to upgrade to the latest chef 12 so they could resolve all deprecations before going to 13. We eve gave them instructions that will cause deprecations warnings like the supports warning to fail their pipelines. Then we did the right thing and fixed supports which makes it a bogus deprecation warnings. Still that warning is causing the same bugs to get filed over and over on our cookbooks. It's probably causing similar issues within customer orgs. It would be great to have a chef 12 build that didn't cause this so that customers could turn on the fail on deprecations setting and not hit this snag.

lamont-granquist:
yeah minimum viable shutting up the deprecation spam...

PR #6223, opened 2017-06-19
bump acceptance provisioning timeout to 10 minutes

thommay:


thommay:
merging this since there are no functional changes.

PR #6225, opened 2017-06-19
check for a set home property before testing file existence (fixes #5777)

### LINKED ISSUE #5777: TypeError: no implicit conversion of nil into String in dscl_user create
## Description

dscl_user fails to create a user with "TypeError: no implicit conversion of nil into String" in https://github.com/chef/chef/blob/7de787be3e792a58285f089c31e2b58b995fde08/lib/chef/provider/user/dscl.rb#L321.
The problem was brought by commit https://github.com/chef/chef/commit/7de787be3e792a58285f089c31e2b58b995fde08.

```
         def current_home_exists?
 -          ::File.exist?("#{current_resource.home}")
 +          ::File.exist?(current_resource.home)
          end
```
`current_resource.home` is `nil` when the user does not exist. File.exist? now raises an exception @lamont-granquist 

## Chef Version

chef-client 12.18.31

## Platform Version

macOS 10.12.3

## Replication Case

```
dscl_user("dummy") do
  supports {:manage_home=>true, :non_unique=>false}
  username "dummy"
  home "/Users/dummy"
  shell "/bin/bash"
end
```

## Client Output

```
* dscl_user[dummy] action create[2017-02-02T15:12:42+00:00] INFO: Processing dscl_user[dummy] action create (/var/chef/cache/cookbooks/user/providers/account.rb line 120)
...
/var/db/dslocal/nodes/Default/users/dummy.plist: file does not exist or is not readable or is not a regular file (Error Domain=NSCocoaErrorDomain Code=260 "The file “dummy.plist” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/var/db/dslocal/nodes/Default/users/dummy.plist, NSUnderlyingError=0x7fd786c0b3a0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}})
...
DEBUG: dscl_user[dummy] user does not exist
...
Error executing action `create` on resource 'dscl_user[dummy]'
TypeError
---------
no implicit conversion of nil into String
```

## Stacktrace

```
TypeError: dscl_user[dummy] (/var/chef/cache/cookbooks/user/providers/account.rb line 120) had an error: TypeError: no implicit conversion of nil into String
/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.18.31/lib/chef/provider/user/dscl.rb:321:in `exist?'
/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.18.31/lib/chef/provider/user/dscl.rb:321:in `current_home_exists?'
/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.18.31/lib/chef/provider/user/dscl.rb:305:in `dscl_set_home'
/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.18.31/lib/chef/provider/user/dscl.rb:178:in `create_user'
```

nathwill:
### Description

`::File.exist?('')` returns false, but is non-obvious, which is how this got broken in the first place; rather than casting back to a string to resolve, make the check for a truthy home property explicit.

### Issues Resolved

#5777 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/user/dscl.rb
+++ b/lib/chef/provider/user/dscl.rb
@@ -318,7 +318,7 @@ user password using shadow hash.")
         end
 
         def current_home_exists?
-          ::File.exist?(current_resource.home)
+          current_resource.home && ::File.exist?(current_resource.home)
         end
 
         def new_home_exists?


lamont-granquist:
but `''` is itself also truthy.

lamont-granquist:
hrm, maybe i'm not understanding the purpose here, but that means it probably needs an explanation and definitely needs a spec test.

also having empty string for a value seems real weird, it might be better to consider converting empty string values to nil, which would make more sense to me.  i'd argue that the first thing which needs to be considered is digging into if dscl_get() can be fixed so that it returns nil instead of empty string, because empty string for things like username, etc also make no sense at all to me.  if empty strings need to be returned by that API for some purposes, then it'd be good to coerce empty strings to nils only for a targeted set of parameters where empty strings don't make sense.  that may take more surgery to the class, but this feels like a non-obvious hacky 'trick' layered on top of a poor API, and that always ends badly.

nathwill:
yeah, i think i explained it badly, `''` is fine and expected to pass the first, even though i agree it would be a weird thing for someone to set for this property. the check for truthiness is to exclude nil since that was failing on the existence check. truthiness seemed good enough to me, but maybe it should be an explicit `!current_resource.home.nil?` check instead?

sgtcoolguy:
This is a blocker for chef usage on mac. The equivalent bug exists for me in Chef 12.19, and looks to still be broken here(!). The assumption that the current_resource.home is non-nil is causing some nasty blocker bugs for me. I'm not really sure how to workaround this (it's happening for the first attempt to create a user for me).

lashomb:
Exists in 12.21.4 as well. Unfortunately going back to 12.17.44 requires dev tools.... so change in workflow if you hadn't included them previously.

jtimberman:
Hello @nathwill, thank you for contributing to Chef! Do you have cycles to continue work on this?

To fix the problem ASAP, I think this change is fine. We definitely need a test. Also, I agree with Lamont. We need to fix `dscl_get()` so it will do the right thing. That seems to be to coerce empty strings into `nil`, but we should verify whether that is the case.

@mikedodge04 said he can help you out, or if you don't have cycles, he can take over and carry this forward. 

Cheers,
Joshua

nathwill:
@jtimberman yeah sadly i'm pretty tied up with other work, if somebody else wants to take over a fix here, 👍

get9:
This fix as-is will return `nil` - I think the proper fix is to do something like:
```
def current_home_exists?
  !!current_resource.home && ::File.exist?(current_resource.home)
end
```

PR #6228, opened 2017-06-20
bump rubocop to 0.49.1

lamont-granquist:
does what it says on the tin


lamont-granquist:
appveyor still timing out after an hour...

PR #6230, opened 2017-06-21
Ensure the service_name is always escaped when using systemd

davide125:
### Description

Chef currently passes the string representing the name of a service (i.e. of the unit in systemd land) directly on the `systemctl` command line when shelling out. This causes issues for units containing escape sequences (e.g. autogenerated units like `dev-dm\x2d1.swap`), which will result in Chef calling `systemctl start dev-dm\x2d1.swap`, which doesn't work as the shell reads the `\` as a control character. This PR fixes the problem by always calling `.shellescape` when using the service name in a command line for the systemd `service` provider and the `systemd_unit` provider.

### Issues Resolved

N/A

### Check List

- [X] All tests pass

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -20,6 +20,7 @@
 require "chef/resource/service"
 require "chef/provider/service/simple"
 require "chef/mixin/which"
+require "shellwords"
 
 class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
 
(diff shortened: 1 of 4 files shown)


davide125:
AppVeyor failure seems to be unrelated

thommay:
@chef/maintainers

thommay:
Looks good, thanks Davide. We do indeed need to figure out why appveyor has recently spiked to taking an hour for builds.

jaymzh:
👍

davide125:
@lamont-granquist ping?

davide125:
@coderanger sure, updated to call `Shellwords.escape()` directly instead

PR #6229, opened 2017-06-20
Properly send run_list_expanded event in policy node case

tylercloke:
### Description

This fixes a bug where the `expanded_run_list` was not being sent to `data_collector` for policy nodes.

### Check List

- [x] New functionality includes tests
- [ ] All tests pass
- [x] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #6233, opened 2017-06-22
Properly send run_list_expanded event in policy node case

btm:
Backport #6229 from master for use in ACC pipeline.


btm:
Looks like those travis failures are already happening in the chef-12 branch.

https://travis-ci.org/chef/chef/builds/245293822

Issue #6215, opened 2017-06-15
Windows 10 Creators Update builds unable to create symbolic links as SYSTEM


svmastersamurai:
## Description

Stumbled upon something quite odd in that Windows 10 Creator's Update builds are unable to create symbolic links using the `link` resource when chef is being run as the local system account. I have done a variety of attempts to remediate this, including granting `NT AUTHORITY\SYSTEM` and my domain joined computer accounts explicit permissions for the `SeCreateSymbolicLinkPrivilege` flag, but it doesn't seem to work. Running the same exact code under my local user account as an admin works just fine.

The folks seeing this, myself included, are developers and have it enabled although I can see this behavior if I disable Windows 10's developer mode.

I know they changed some symlink privilege things around in this build so figure I'd start here to see if you guys have seen this or had any advice 😄

## Chef Version

I can confirm this behavior is consistent across chef-clients from 12.8.1 to 12.19.36.

## Platform Version

A majority of the hosts exhibiting this behavior are running Windows 10.0.15063, but I can confirm this affects up to 10.0.16199 as these are also present in our environment as well.

## Replication Case

I have constructed a basic chef recipe that exhibits the behavior, and my host is running the affected build:
```bash
PS C:\> gcim cim_operatingsystem | select version

version
-------
10.0.15063
```

Here is the chef recipe
```ruby
user = ENV['USERNAME']
has_it = Chef::ReservedNames::Win32::Security.get_account_right(user).include?('SeCreateSymbolicLinkPrivilege')
Chef::Log.info("#{user} has SeCreateSymbolicLinkPrivilege? #{has_it}")
grant = Chef::ReservedNames::Win32::Security.add_account_right(user, 'SeCreateSymbolicLinkPrivilege')
Chef::Log.info("Result of chef's granting of privilege #{grant}")

link 'C:\chef-clone' do
  to 'C:\chef'
end
```

## Client Output

```bash
# Logfile created on 2017-06-14 20:13:48 -0700 by logger.rb/54362
[2017-06-14T20:13:48-07:00] INFO: *** Chef 12.19.36 ***
[2017-06-14T20:13:48-07:00] INFO: Platform: x64-mingw32
[2017-06-14T20:13:48-07:00] INFO: Chef-client pid: 20664
[2017-06-14T20:13:49-07:00] WARN: Run List override has been provided.
[2017-06-14T20:13:49-07:00] WARN: Original Run List: []
[...]

PR #6236, opened 2017-06-23
add option to enable unprivileged symlink creation on windows

svmastersamurai:
Signed-off-by: Daniel Sedlacek <dansedlacek@fb.com>

### Description

Windows 10 Creators Update and on enables unprivileged symlink creation but in some scenarios, such as when the` chef-client` runs as a scheduled task as the local system account, any calls to create a symlink error out with:
```
System Error Code: 1314
System Error Message: A required privilege is not held by the client.
```
[Passing a newly introduced flag](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/#cpLA6xrKTwb5fOf7.97) appears to allow for `chef-client` to create symlinks from a scheduled task like in the older Windows 10 builds.

I have added this as a toggleable configuration option since in our environment we manage the chef configuration with chef so this is a fairly low barrier of entry for us to introduce a monkeypatch that can be rolled out so we can remediate the errors until the next stable client is released.

I am not married to this approach however, but definitely wanted to get the ball rolling on getting this fixed so we eventually stop getting these errors 😃 

### Issues Resolved

This should fix [#6215](https://github.com/chef/chef/issues/6215) if the option is enabled in the node's `client.rb`.


jaymzh:
CC @chef/client-windows 
CC @chef/client-core

btm:
@svmastersamurai do you know any downsides to this? The only one I can find is that you could get an error if you can't use the flag.

https://news.ycombinator.com/item?id=13141206
> The primary motivation for adding the SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag is to not disrupt the behavior of existing code. Because of pre-existing behavior, existing apps are likely to be built to assume that symlinks cannot be created when the app is running without admin rights. If the behavior changed such that the app could now create symlinks but not have admin rights, the app and/or its dependent scripts etc. will likely misbehave. Therefore, we added a new flag - SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE - which allows developers to expressly indicate that their code is specifically built to handle the ability to create symlinks should their code be running on a supporting platform (i.e. Win10 Creators Update, or later) even if the app does not have admin rights.

This sounds like a lot of work for an edge case that doesn't make sense, or I'm misunderstanding. I can't see any scenario in Chef where we'd ask it to make a symlink and want it not to do so if it could.

So I think what we want to do is not have a config option and always do this. Which means not trying if it's not going to work (check build version?) or maybe trying without it, then trying with it, then failing?Either way I think that there's an error of `ERROR_INVALID_PARAMETER` if it isn't supported and we'd want to catch that and raise some english.

svmastersamurai:
@btm as I mentioned this was a minimum viable product to get some discussion rolling, checking a build version and not making this optional sounds like a fine approach as well!

When we met with chef a couple years ago and had discussions on our Windows environment it didn't seem like an issue to have the client invoked as scheduled task as the local system account and we haven't had reason to change that behavior up until the creator's update was released. In our current setup the creation of symlinks is broken regardless of what version of chef-client you're on, and we were in the process of making a big push to upgrade our windows 10 fleet to later builds since it was going to make our development infrastructure team's lives a bit easier. Without working symlinks though we've had to hold off on that push and with more folks upgrading out of our usual channels we've been seeing the incidence of chef failures slowly creeping upwards :(

If you guys are still open to this PR I can revise the implementation to check the node's build version (and likely have to account for edition as well since server 2016 follows the kernel 10+ build numbers) and then add some more robust error handling.

btm:
@svmastersamurai I think this is right thing to do. Maybe my comment about edge cases wasn't clear enough; I think it's silly that Microsoft put this behind a flag rather than just letting it start working.

What I'm suggesting is that we check platform and/or always pass this flag and drop the Chef::Config flag. So two options to consider:

1) Check the platform, and always pass `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` if we believe the platform will accept it, and/or
2) always pass `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` and catch `ERROR_INVALID_PARAMETER` or whatever, drop the flag, and retry.

svmastersamurai:
@btm thanks for clarification, I will revise this with that flow in mind :)

svmastersamurai:
@btm I went with implementing the first one and updated the unit tests to reflect the change in logic. Feedback would be appreciated since I had to expose a few things in the `Win32::Version` class to make the check work 🤔

spec/unit/win32/link_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    context "a valid parameter is passed" do
+      before(:each) do
+        allow(File).to receive(:directory?).and_return(false)
+        allow(Chef::ReservedNames::Win32::File).to receive(:encode_path) { |a| a }
+        allow(Chef::ReservedNames::Win32::File).to receive(:CreateSymbolicLinkW).with(any_args) { "don't //actually// do this" }
+      end
+
+      it "will not pass the unpirivileged symlink flag if the node is not Windows 10" do
```
btm: s/unpirivileged/unprivileged/ in a couple descriptions here.

lib/chef/win32/file.rb
```diff
@@ -60,6 +61,7 @@ def self.symlink(old_name, new_name)
         # TODO do a check for CreateSymbolicLinkW and
         # raise NotImplemented exception on older Windows
         flags = ::File.directory?(old_name) ? SYMBOLIC_LINK_FLAG_DIRECTORY : 0
+        flags |= SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE if Chef::ReservedNames::Win32::Version.new.win_10_creators_or_higher?
```
mwrock: Another approach here instead of checking for a specific windows version would be to check the error returned from `CreateSymbolicLinkW` for `ERROR_INVALID_PARAMETER` (87). Thats the [approach taken by rust](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/fs.rs#L649), but I think this is fine too.

btm: Agree, I offered both but left it up to the implementer.

PR #6237, opened 2017-06-26
[knife] add order by and limit the number of cookbooks returned when listing

iennae:
### Description

This updates knife cookbook site list to add order by and limit number functionality. It changes the default behavior from listing all cookbooks to 100. The refactor speeds up the response.

Additionally, code is removed from knife cookbook show list that isn't used. 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/cookbook_site_show.rb
+++ b/lib/chef/knife/cookbook_site_show.rb
@@ -47,20 +47,6 @@ class Chef
           noauth_rest.get("#{supermarket_uri}/cookbooks/#{@name_args[0]}/versions/#{name_args[1].tr('.', '_')}")
         end
       end
-
-      def get_cookbook_list(items = 10, start = 0, cookbook_collection = {})
-        cookbooks_url = "#{supermarket_uri}/cookbooks?items=#{items}&start=#{start}"
-        cr = noauth_rest.get(cookbooks_url)
-        cr["items"].each do |cookbook|
-          cookbook_collection[cookbook["cookbook_name"]] = cookbook
-        end
-        new_start = start + cr["items"].length
-        if new_start < cr["total"]
-          get_cookbook_list(items, new_start, cookbook_collection)
-        else
-          cookbook_collection
-        end
-      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


iennae:
@robbkidd

thommay:
@iennae please fix the chefstyle violations in travis

lib/chef/knife/cookbook_site_show.rb
```diff
@@ -47,20 +47,6 @@ def get_cookbook_data
...
-        cr["items"].each do |cookbook|
-          cookbook_collection[cookbook["cookbook_name"]] = cookbook
-        end
-        new_start = start + cr["items"].length
-        if new_start < cr["total"]
-          get_cookbook_list(items, new_start, cookbook_collection)
-        else
-          cookbook_collection
-        end
-      end
```
robbkidd: tenor-151493477

robbkidd:
👍  These were the changes @sigje and I paired on.

**NOTE:** default behavior changed. With this change, `knife cookbook site list` won't list _all_ cookbooks at the given Supermarket URL. Defaults to listing the first 100 of a given sort order. If changing the default behavior isn't acceptable, something could be done with some new option switches and two API trips: one to get the total and another to retrieve that number items.

stevendanna:
Hard to say whether this behavior change is the right one without know how people use that command.  Some thoughts:

- By removing the pagination, it means that if someone sets `-n 5000` to ensure they can list all of the cookbooks on the community site, then we make a single request asking for 5000 items.  Do we think that would be problematic for public or private supermarket sites?

- An alternative might be to keep the pagination (increasing the page size to 100 or more) and then also adding a configurable overall limit.

thommay:
I don't think we can change default behaviour in a minor version.

robbkidd:
> By removing the pagination, it means that if someone sets -n 5000 to ensure they can list all of the cookbooks on the community site, then we make a single request asking for 5000 items. Do we think that would be problematic for public or private supermarket sites?

The Supermarket v1 API currently allows someone to make a request with items = 5000. At the current number of cookbooks on public Supermarket, this hasn't been a problem.

> An alternative might be to keep the pagination (increasing the page size to 100 or more) and then also adding a configurable overall limit.

Another alternative is experimenting with chunked transfer encoding. I don't know offhand how the rest client inside chef/knife handles that, though.

PR #6241, opened 2017-06-29
Investigate length of package functional tests in Appveyor for chef/chef <Do not merge>

harikesh-kolekar:


btm:
We're back to normal as of https://github.com/chef/chef/pull/6291. Thanks for the investigation work @harikesh-kolekar.

Issue #6248, opened 2017-06-30
git config --replace-all breaks when repo url has multiple words


hoprocker:
When using a git repository with multiple words, ie with the `ext::ssh` protocol, the `git config --replace-all` command breaks. This is only a problem when a) a repository has multiple remotes, or b) the repository name doesn't match the shortname.

This appears in the most recent version of Chef (13.2.20), but [that line](https://github.com/chef/chef/blob/462ce499783771dd54d3cba576e6165643a25b11/lib/chef/provider/git.rb#L201) dates from Apr 29, 2016.

The fix should be very simple -- pass the remote url into the config as a quoted string, as is done with the similar `git_ls_remote()` function.

I also see that the in `Chef::Mixin::ShellOut`, there's a recommendation to generally use shell_out_compact() -- maybe that would work here, as well.

coderanger:
How can a URL have multiple words?

hoprocker:
Good question -- when using git's "external transport" protocol feature, ie when using different credentials for different repositories inline. For example,

```
ext::ssh -i /home/deploy-user/.ssh/all_awesome_repos_key -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no git@github.com %S /GreatOrg/one_awesome_repo
```

Gleaned from [here](https://stackoverflow.com/questions/24024783/checkout-git-repo-with-chef-with-ssh-key/33810160#33810160)

coderanger:
You might be interested in the [`poise-git` cookbook/resource](https://github.com/poise/poise-git) which handles this already.

hoprocker:
Yup, looks like that might cover at least this one scenario; I still feel like this might be worth adding, though, to support this particular edge case (which afaik is a valid usage of git)?

tas50:
Fixed in #7014 which will go out in Chef 14.2

PR #6249, opened 2017-06-30
quote git remote_url

hoprocker:
for issue #6248

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -213,7 +213,7 @@ class Chef
             #   which we can fix by replacing them all with our target url (hence the --replace-all option)
 
             if multiple_remotes?(remote_status) || !remote_matches?(remote_url, remote_status)
-              git("config", "--replace-all", "remote.#{remote_name}.url", remote_url, cwd: cwd)
+              git("config", "--replace-all", "remote.#{remote_name}.url", %("#{remote_url}"), cwd: cwd)
             end
           when 1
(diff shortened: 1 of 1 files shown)


coderanger:
Seems reasonable to support but we should use shellwords instead. And should check for weird interactions with cmd.exe on windows.

lamont-granquist:
i think this is probably acceptable as it is.  we commonly just throw quotes around args to protect windows against files-with-spaces.

what we really need to do is get *argv support into mixlib-shellout on windows and then port stuff like this to it, and solve it right at that level and then unburden consumers from having to care so deeply about it...

thommay:
@hoprocker thanks for the PR, and sorry we've been a bit slow getting it merged. We need a think from you, which is to mark your commit with a DCO sign off. Since you have a single commit, please run `git commit --amend -s`, and then `git push -f` to update the branch here. Once that's done we'll get this merged.
Thanks again!

lamont-granquist:
it has a chefstyle issue as well:

```
Offenses:
lib/chef/provider/git.rb:216:75: C: Style/PercentLiteralDelimiters: %-literals should be delimited by { and }.
              git("config", "--replace-all", "remote.#{remote_name}.url", %("#{remote_url}"), cwd: cwd)
                                                                          ^^^^^^^^^^^^^^^^^^
1611 files inspected, 1 offense detected
RuboCop failed!
```

tas50:
Merged with the chefstyle + spec fix in #7014

PR #6250, opened 2017-06-30
fix execute resource to use shell_out_with_systems_locale

### LINKED ISSUE #6239: /opt/embedded/chef/bin in PATH causes scripts to fail
## Description

Since Chef 13, the `/opt/embedded/chef/bin` path is now injected at the beginning of PATH, giving it higher priority. In previous versions, it was injected at the end. This now causes problems when running pkg-config and probably other things as well. Is there a fix (or am I doing something wrong ;-)?

## Chef Version

Chef 13.1.31

## Platform Version

RHEL 6.4

## Replication Case

    execute 'test' do
      command 'env'
      live_stream true
    end

## Client Output

    # running with chef 13.1.31
    PATH=/opt/chef/embedded/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

    # running with chef 12.21.1  
  
  PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/chef/embedded/bin:/opt/chef/embedded/bin

lamont-granquist:
This may be a breaking change for some users coming from 13.0/13.1
but this was the intended way that this API would function in Chef 13.0.

With this change we stop inserting environment variables into
execute resources.  That means that whatever users type into the
execute resource should much more exactly agree with the response
on the command line, which will close a shedload of bugs and is how
Chef 13.0 was intended to operate.  We no longer globally mangle the
PATH or the LANG/LC_ locale environment variables.

The way that it worked in Chef 13.0/13.1 resulted in the internal
chef embedded/bin directory being prepended to the PATH environment
variable in the `execute` resource, which was never intended.

closes #6239

### DIFF
--- a/lib/chef/provider/execute.rb
+++ b/lib/chef/provider/execute.rb
@@ -55,7 +55,7 @@ class Chef
 
         converge_by("execute #{description}") do
           begin
-            shell_out!(command, opts)
+            shell_out_with_systems_locale!(command, opts)
           rescue Mixlib::ShellOut::ShellCommandFailed
             if sensitive?
               raise Mixlib::ShellOut::ShellCommandFailed,
--- a/spec/functional/resource/execute_spec.rb
+++ b/spec/functional/resource/execute_spec.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Serdar Sutay (<serdar@chef.io>)
-# Copyright:: Copyright 2014-2016, Chef Software Inc.
+# Copyright:: Copyright 2014-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -51,7 +51,7 @@ describe Chef::Resource::Execute do
 
       # why_run mode doesn't disable the updated_by_last_action logic, so we really have to look at the provider action
       # to see if why_run correctly disabled the resource.  It should shell_out! for the guard but not the resource.
-      expect_any_instance_of(Chef::Provider::Execute).to receive(:shell_out!).once
+      expect_any_instance_of(Chef::Provider::Execute).to receive(:shell_out_with_systems_locale!).once
 
       resource.only_if guard
       resource.run_action(:run)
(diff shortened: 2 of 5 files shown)


lamont-granquist:
@chef/client-core reviewage

lamont-granquist:
Yeah, I'm sure it'll break some stuff.

Since execute is also the base class for the script resource and all the other script resources (bash, etc) that makes this a fairly wide-ranging change.

This is the way that I intended and thought Chef 13.0 would work though and I'm positive #6239 will trip up many, many people and this gets us to the correct future which is where execute and script resources just do-what-i-type with no magic.

lamont-granquist:
appveyor busted because appveyor is busted

lamont-granquist:
So to be very, very clear in case any users find this PR:

You were never intended to run your software off of the chef-client's omnibus install.

If you have a ruby app that you're deploying and you expect to bundle install and pick it up out of /opt/chef/embedded/bin/{ruby,bundle,gem,etc} then this PR just broke that.  Very much on purpose.  It was intended to break in 13.0

You must have your own ruby in /usr/bin/ruby or /usr/local/bin/ruby or rvm-managed somewhere or whatever.

You can continue to use the embedded ruby by applying a patch like this:  #6251 

But the fact that you're scribbling over the global path and the care with which I restored the old path there should be an indication that you're doing something incorrect there.  You're also creating a tight coupling between your software and our ruby distribution.  If we get stuck on an old version of ruby then you'll get stuck on an old version of ruby.  If we move forward to a new version of ruby and your software isn't ready for that, you'll be stuck on an old version of chef-client.  The fact that you need to do this indications that you're probably doing something wrong.

And to reiterate this fixes many bugs where things like `pkg-config` were being picked up out of the embedded path and creating cross-linking issues between the system libraries and embedded libs (particularly things like openssl).

We continue to this with our own acceptance tests because that is in the same repo as the client itself and its tightly coupled and we'll certainly take the time to fix it if we ever need to.

Issue #5552, opened 2016-11-16
Knife should give a useful error when the chef_server_url isn't a chef server API.


lioramilbaum:
## Description

Running knife list returns unclear error

Briefly describe the issue

## Chef Version

12.11.0

Tell us which version of chef-client you are using (see below for Server+ChefDK bugs).
12.15.19

## Platform Version
Ubuntu 14.04

Tell us which Operating System distribution and version chef-client is running on.
OSX 10.12.1

## Replication Case

Launched chef server with chef-server community cookbook. Self signed certificated.
Fetched the certificate with - knife ssl fetch

## Client Output

Lioras-MBP:chef liora$ knife list -R /  --config config.rb 
ERROR: /nodes: No such file or directory
ERROR: /data_bags: No such file or directory
ERROR: /clients: No such file or directory
ERROR: /roles: No such file or directory
ERROR: /users: No such file or directory
ERROR: /environments: No such file or directory
ERROR: The object you are looking for could not be found
Response: <!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
  <meta http-equiv="X-UA-Compatible" content="IE=EmulateIE7" />
  <title>Chef - 404 Not Found</title>
  <link media="all" rel="stylesheet" type="text/css" href="/css/all.css" />
  
</head>
<body>
  <div class="header-block">
    <div id="header">
      <strong class="logo"><a href="https://www.chef.io">Chef</a></strong>
    </div>
  </div>
  <div id="wrapper">
    <div id="main">
      <div class="mybox">
        <div id="content">
          <h1>404 - Not Found</h1>
          <p>Sorry, I can't find what you are looking for.</p>
        </div>
      </div>
    </div>
  </div>
  <div class="footer-block">
    <div id="footer">
      <div class="mybox">
      </div>
      <div class="footer-bottom">
        <span>&copy; 2010&thinsp;&ndash;&thinsp;2016 Chef Software, Inc. All Rights Reserved</span>
      </div>
    </div>
  </div>
</body>
</html>## Stacktrace

Please include the stacktrace.out output or link to a gist of it, if there is one.

### NOTE: CHEF CLIENT BUGS ONLY

This issue tracker is for the code contained within this repo -- `chef-client`, base `knife` functionality (not
plugins), `chef-apply`, `chef-solo`, `chef-client -z`, etc.

[...]

PR #6253, opened 2017-07-05
Knife should give a useful error when the chef_server_url isn't a chef server API

jeunito:
### Description
Raises an exception if the value of chef_server_url is not a valid chef_server_api

### Issues Resolved

https://github.com/chef/chef/issues/5552

### Check List

- [x] All tests pass


stevendanna:
Do we know what this query will return on an open source chef 11 server?

jeunito:
@stevendanna no I did not consider that. Should it be supported too?

stevendanna:
Unfortunately, we don't have great numbers on the deployed versions of Chef Server out there.  I'd definitely prefer if there was a way to do this without breaking Chef Server 11 if it isn't too onerous.

jeunito:
@coderanger I just assumed that the value of ```chef_server_url``` is always of the form 
> http://fqdn/organizations/NAME

which is what I see on a hosted chef setup. 

Hence, it is actually sending a request to https://docs.chef.io/api_chef_server.html#organizations-name. 

Could this be different in an on-premise setup?

coderanger:
@jeunito That's part of the global API though, which we don't use for anything else in chef-client that I know of. There is no definition for a `/` call in https://docs.chef.io/api_chef_server.html#organization-endpoints

coderanger:
@jeunito Basically we don't use that API from chef-client, it's part of the Erchef management API but not the "Chef Server API" which is a subset/different thing.

coderanger:
Otherwise, for example, this would fail on Chef Server 11.

stevendanna:
Because of the array of server implementations we might be talking to, an active check might be difficult.

I think my preferred alternate approach would be to avoid doing an extra request completely and add some heuristics to the error handling for failed requests to add a warning if the response looks like it might not be from a chef server.

That is, if you tried to list nodes but you got a 404 or some other error response, we look for things like whether the response was HTML rather than JSON and then print out a "This response does not
look like a response from a valid Chef Server, is your chef_server_url correct?" message.

## On GET /organizations/ORGNAME

I'm not sure I agree that this endpoint isn't part of the Chef API. Namely, it is

- Documented here: https://chef-server-api-docs.chef.io/#organizations__organization__get
- Supported by chef-zero
- Tested in oc-chef-pedant

However, the fact that it is a question at all likely means we should not add it as a dependency to every request. We've certainly treated it as an internal API as the contents have changed over time. Further, I could imagine situations were odd permissions issues could get in the way of an otherwise successful request.

## A possible active check
Unfortunately, supporting chef server 11, chef server 12, chef-zero, and goiardi might be difficult.

Perhaps we could do the following:

- GET to /clients/CLIENT_NAME where CLIENT_NAME is name of the
  client we are trying to make a request as.

- In the case of a 200, we can confirm that we got a client back.

- In the case of a 404, a compliant Chef server should still be
  serving up JSON content. If we see anything else, we can
  complain that it might not be a Chef Server.

- In the case of a 401 or 403, again we would likely have to
  check that the shape of the response looks correct.

- In the case of 50X we might want to "fail open" in case it was
  a transient failure.

## Some other thoughts

I originally thought we could use a response header. Unfortunately I couldn't not find one that is supported across implementation. Chef Server will return `x-ops-api-info` and `x-ops-server-api-version` but the former doesn't appears unsupported in chef-zero or goiardi and the later was only added in Chef Server 12.something.

We could use an endpoint that is more or less guaranteed to exist (such as `/clients`) but we would still have to have handling for authn and authz errors and this could be problematic for organizations where the return body will be large.

jeunito:
Thanks Steven. I ll look into your suggestion.

jeunito:
> In the case of a 404...If we see anything else, we can complain that it might not be a Chef Server.

@stevendanna unfortunately we can't inspect the content when we get a 404 due to https://github.com/chef/chef/blob/master/lib/chef/http.rb#L155-L163

How about just letting it fail for all of 401, 404, 403 and 5xx cases you mentioned above and putting something like the following: 

> There was an issue while trying to validate your chef_server_url. Please check that: 
> 1. The value of chef_server_url points to a chef server
> 2. This client or your user name has the appropriate permissions 
> 3. The client or user ```node or client name``` exists 

This would be one step better than just letting chef fail with an unclear error and no clues for the user as mentioned in https://github.com/chef/chef/issues/5552

jeunito:
@stevendanna nvm the comment at https://github.com/chef/chef/pull/6253#issuecomment-316843590. I think I have found an answer to the initial perceived obstacle I mentioned.

lamont-granquist:
Unfortunately I'm 100% opposed to anything which incurs an extra round trip and API call just to ping the server every time a ServerAPI object is created.  Creates more round trip delays and more load on the server.  I'd prefer to leave the problem unfixed than to fix it this way.

You could do something better with error decoration where when get raised errors like ECONNREFUSED or 404'd then ping to check if it looked like it was a chef server at all -- *after* you've already hit an error.  Ideally memoize this result so that you don't need to ping check for multiple broken requests (and maybe if you get any success set it so that it doesn't have to ping it in the future).

lamont-granquist:
^ there's going to be an obviously annoying aspect to that which is that if you rescue in get() and then call get() you're getting recursive behavior so that needs to be handled carefully so that you don't cause a ping flood / stack size overflow

stevendanna:
> You could do something better with error decoration where when get raised errors like ECONNREFUSED or 404'd then ping to check if it looked like it was a chef server at all -- after you've already hit an error. Ideally memoize this result so that you don't need to ping check for multiple broken requests (and maybe if you get any success set it so that it doesn't have to ping it in the future).

Yes, this is what I would like to see as well.  Heuristics on the /response/ to a request rather than a proactive check.

jeunito:
@stevendanna @lamont-granquist ok thanks both. Will do that instead

jeunito:
> ...ping to check if it looked like it was a chef server at all -- after you've already hit an error.

@stevendanna @lamont-granquist please see this [commit](https://github.com/chef/chef/pull/6253/commits/b3eec2dc4c27bba5256b1ae42edb70cf422997ac) that reflects this proposal

lib/chef/knife/list.rb
```diff
@@ -104,6 +104,18 @@ def add_dir_result(result)
         begin
           children = result.children.sort_by { |child| child.name }
         rescue Chef::ChefFS::FileSystem::NotFoundError => e
+          begin
+            rest.get("/clients/#{Chef::Config[:client_name]}")
+          rescue Net::HTTPServerException => e
+            if e.response.class == Net::HTTPNotFound
+              require "json"
+              begin
+                JSON.load(e.response.body())
```
lamont-granquist: that should use ffi-yajl, and the require should get moved to the top of the file.

lamont-granquist:
could probably use a rebase+squash of commits (and removal of the merge commit)

lib/chef/knife/list.rb
thommay: This should happen in `chef/http` and not in individual classes.

jeunito: @thommay Did you mean ```chef/server_api```? I reckon that ```chef/http``` could be used somewhere else and this logic is specific to a chef server.

jeunito:
@thommay @lamont-granquist please see the new commits. I have extracted the logic for the heuristic to server_api

PR #6257, opened 2017-07-07
Chef 14: Remove deprecated bootstrap options

tas50:
We said we would remove these in Chef 13 in the code. Lets do it for 14 instead.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -119,16 +119,6 @@ class Chef
         :description => "Do not proxy locations for the node being bootstrapped; this option is used internally by Opscode",
         :proc => Proc.new { |np| Chef::Config[:knife][:bootstrap_no_proxy] = np }
 
-      # DEPR: Remove this option in Chef 13
-      option :distro,
-        :short => "-d DISTRO",
-        :long => "--distro DISTRO",
-        :description => "Bootstrap a distro using a template. [DEPRECATED] Use -t / --bootstrap-template option instead.",
-        :proc        => Proc.new { |v|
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -307,10 +288,9 @@ def user_name
       end
 
       def bootstrap_template
-        # The order here is important. We want to check if we have the new Chef 12 option is set first.
-        # Knife cloud plugins unfortunately all set a default option for the :distro so it should be at
```
stevendanna: Does this imply any work for the cloud plugins or is this comment just out of date?

tas50: I'll have to check on that

Ames0831:
I'm not sure what I'm doing?

tas50:
This is going to require a bit of digging to figure out where else these are used

lamont-granquist:
Mostly I'd suspect it would be knife-windows that would be mucking about with the internals enough of knife-bootstrap to break when these are removed...

tas50:
The same options are deprecated in knife-windows so I don't see this impacting that. We'll nuke them from knife-windows and do a major version bump there as well.

lib/chef/knife/bootstrap.rb
tas50: We do need to clean this all up in knife-cloud it appears.

tas50:
I merged this on the CLI to fix the spec conflict. We've updated the knife-windows tests to not care about this CLI option. We should go through our other tools to make sure they're not trying to use these options anymore. They've been deprecated for years though and we said we were going to remove them in Chef 13.

PR #6261, opened 2017-07-10
move nokogiri config to omnibus

lamont-granquist:
this hand-builds it with the software dep, and its not a direct dep of
chef itself and shouldn't be in the Gemfile.lock anyway, plus we need
to pin via omnibus_overrides.rb and double-pinning in the Gemfile.lock
is just added fussiness


lamont-granquist:
there's nothing that is going to blow up travis or appveyor in here and we need to get omnibus builds back to green...

PR #6274, opened 2017-07-14
Fix solaris/x86 builds

thommay:
The actual fix is chef/omnibus-software#864


thommay:
merging since there's no functional changes.

PR #6275, opened 2017-07-14
Use the previous VS 2015 image in appveyor instead of 2013

btm:
@harikesh-kolekar believes we'll have better luck on the 'Previous' 2015 image rather than 2013.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 version: "master-{build}"
 
-os: Previous Visual Studio 2013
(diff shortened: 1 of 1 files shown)


btm:
Appveyor has shared that they believe our performance issues with their new images are this issue: https://github.com/appveyor/ci/issues/1657

We should use the previous 2015 images for the moment.

btm:
The 32bit build still can take ~52 minutes, but that's progress for now.

PR #6279, opened 2017-07-18
added gateway attribute to ifconfig resource on debian

tomklino:
### Description

I noticed that the ifconfig resource does not implement the "gateway" option for network interfaces. I tried some local changes to see if some copy pasting and name changing from existing attributes will do the trick. It did! Using gateway "ipaddress" will now put that ip address in the interface file.
### Issues Resolved

I did not find a relevant issue for that subject.

### Check List

NOTE: I ran into multiple issues trying to get rspec to run. something with cheffish requiring version 2.3.3 of ruby (which I've installed). Either way, I have no way to run the existing specs on my machine.
If there is a docker image with all the requirements for the tests to run please let me know and I'll be happy to run and write tests for this pull request. But as of now I've been trying to get rspec to work for more than an hour.

- [ no ] New functionality includes tests
- [ don't know ] All tests pass

### DIFF
--- a/lib/chef/provider/ifconfig/debian.rb
+++ b/lib/chef/provider/ifconfig/debian.rb
@@ -48,6 +48,7 @@ iface <%= new_resource.device %> inet static
     <% if new_resource.metric %>metric <%= new_resource.metric %><% end %>
     <% if new_resource.hwaddr %>hwaddress <%= new_resource.hwaddr %><% end %>
     <% if new_resource.mtu %>mtu <%= new_resource.mtu %><% end %>
+    <% if new_resource.gateway %>gateway <%= new_resource.gateway %><% end %>
 <% end %>
 <% end %>
           }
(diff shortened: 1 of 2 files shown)


tas50:
@tomklino Any chance you can rebase this and sign off for the DCO so we can merge this. It would be great to get included in Chef 14

tomklino:
@tas50 looks like the files in the commit have changed a lot since I initiated the pull request and it doesn't seem to make sense anymore. the diff is all over the place. I'll have a look at it when I get more time to see if I can offer the same change according to the new logic. (will welcome the help if you can spare the time)

tas50:
I've implemented this change in #7475. We modernized this resource a while back so the format of things changed around a bit. I also added support for RHEL systems as well since they support gateway too. Thanks for the PR even though we didn't end up using it. It got the ball rolling.

PR #6280, opened 2017-07-18
Handle display names with trailing spaces

### LINKED ISSUE #6217: Windows_Package should handle trailing spaces in package names
## Description

PostgreSQL 9.6.3.1 on Windows installs itself with a DisplayName of "PostgreSQL 9.6 " (note the trailing space). If the package resource has the name "PostgreSQL 9.6" then the two don't match and Chef will re-install the package on every run.

It would be nice if https://github.com/chef/chef/blob/master/lib/chef/provider/package/windows/registry_uninstall_entry.rb#L43 stripped spaces before doing the comparison

## Chef Version

12.13.37

## Platform Version

Windows 2012R2

## Replication Case

Create a recipe with this resource (and download the PostgreSQL 9.6.3 installer)
```
package 'PostgreSQL 9.6' do
  source "postgresql-9.6.3-1-windows-x64.exe"
  installer_type :custom
  options '--mode unattended '\
          '--unattendedmodeui none '\
          "--servicename Postgres "\
          "--datadir d:\\PGSqlData "\
          "--superpassword password "
  action :install
end
```

perlun:
Closes #6217.

The reason why we do `to_s.rstrip` is to handle the (potentially weird case of) `display_name` being `nil`, in which case we will otherwise get an unfriendly Ruby error message.

### Check List

- [ ] New functionality includes tests (negative: if this is a requirement, please tell me where to add them)
- [x] All tests pass
- [ ] ~RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)~ Should not be required in this case
- [ ] ~All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>~ (should not be needed for the trivial change, let me know if you want me to amend to include this)

### DIFF
--- a/lib/chef/provider/package/windows/registry_uninstall_entry.rb
+++ b/lib/chef/provider/package/windows/registry_uninstall_entry.rb
@@ -40,7 +40,7 @@ class Chef
                     begin
                       entry = reg.open(key, desired)
                       display_name = read_registry_property(entry, "DisplayName")
-                      if display_name == package_name
+                      if display_name.to_s.rstrip == package_name
                         quiet_uninstall_string = RegistryUninstallEntry.read_registry_property(entry, "QuietUninstallString")
                         entries.push(quiet_uninstall_string_key?(quiet_uninstall_string, hkey, key, entry))
                       end


thommay:
Submitter has said this is an obvious fix, no DCO required.

btm:
Added MSYS-631 to add a test.

PR #6284, opened 2017-07-19
fix retries on JSON POST requests

lamont-granquist:
on the first pass through the JSON middleware it encodes the
body.  if there's a retry, it re-encodes the body as a string
with all its metacharacters escaped.

this is a particular issue when doing a first request that is
a POST that requires negotiating the API version.  when doing a
GET it isn't a problem because there's no body payload -- but a
POST or a PUT which requires a retry will get garbled and will
cause a 500.

this happens on hosted right now if trying to POST with a v2
API since hosted is only v1, so there's a retry to downgrade.

i also made the same kind of changes to the streaming download
requests, but since they're GETs its unclear to me if there was
any impact there -- but middleware could have been double-mangling
headers on a retry.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,7 @@ end
 group(:development, :test) do
   gem "rake"
   gem "simplecov"
+  gem "webmock"
 
   # for testing new chefstyle rules
   # gem 'chefstyle', github: 'chef/chefstyle'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,6 +122,7 @@ GEM
     addressable (2.4.0)
     appbundler (0.10.0)
       mixlib-cli (~> 1.4)
+      mixlib-shellout (~> 2.0)
     ast (2.3.0)
     backports (3.8.0)
     binding_of_caller (0.7.2)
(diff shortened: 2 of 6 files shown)


lamont-granquist:
Yeah, its a problem whenever there needs to be a renegotiation.  Whenever the client is looking for N and the server is < N.

lamont-granquist:
It just concretely popped up for me when talking to hosted.

lamont-granquist:
(and yeah it looks like we sleep on any kind of retry, including version negotiation)

lamont-granquist:
also: this isn't just a renegotiation problem.  i think that any time we fail a POST/PUT we will uselessly retry and fail.

lamont-granquist:
^ last statement is not true, this does only affect protocol renegotiation.  which makes more sense.  the retries for normal GET/PUT/POST happen in send_http_request and don't retry the same loop that protocol renegotiation does.  which makes this bug substantially less impactful than i was starting to get worried about last night.

lamont-granquist:
bonus fix:  removes the sleeps from protocol negotiation.  since the retries on failure is a completely separate retry loop i believe we can just drop it here.

lamont-granquist:
okay so this seems to only affect chef-client 13.0.x through 13.2.x and only on protocol negotiation with PUT or POST requests.

PR #6287, opened 2017-07-20
Update Expeditor config to use new Merge and Artifact Actions

tduffield:
Chef Expeditor now supports built in merge and artifact actions which
allow us to reduce the burden of duplicating common processes across
all our repositories. This change introduces several of these built_in
functions, replacing functionality that was previously kept in our
Rakefile.


coderanger:
How does a non-Chef employee generate the changelog then?

coderanger:
Or more generally, this looks like it walks back on a lot of the earlier compromises made to keep things in the Chef release process not directly tied to the closed-source Lita tool.

tduffield:
By modifying the CHANGELOG.md manually 😄 

Expeditor just does some simple inplace regex modification to inject the PR title as a line item under the appropriate category header (which is based on the label you provide). You can see what the end result of a changelog manipulation would look like here: https://github.com/chef/opscode-pushy-client/blob/master/CHANGELOG.md

coderanger:
But they can't generate it manually, you're removing the rake task for it. Unless you mean actually manually, in which case that's a breaking change in my book.

coderanger:
I get that y'all want to use the nice things for Chef itself, and I respect that. But the same as I said last time, if it's closed-source you can't use it for anything important on Chef. The cost of keeping your release system proprietary is that you can't use (as much of) it on the open-source side.

tduffield:
IIRC, the compromise we came to was that [Chef Maintainers](https://github.com/orgs/chef/teams/client-maintainers/members) would have access to the code that we use to perform these automated processes: that continues to be the case. In fact, we've expanded that to include all the code necessary to launch your own instance of the software. All Chef Maintainers, whether they work for Chef Software or not, have the ability to see the code that is used to power these processes.

coderanger:
That was part of the compromise, to ensure that the maintainers can see how the sausage is made, but part of the ethos of open-source is that our users should not be beholden to any maintainer (including Chef Software as a whole). As the Lita code is still not licensed in such a way that I could fork it (even though I can see it as a maintainer), I will vote against using any features of it that leave us overly tied to the Lita software. That said, "overly" is clearly subjective. However you have done little here to explain the new features and how they work, or otherwise ameliorate the harm to our users that this causes (and yes, it is a harm to depend on something proprietary like, say, GitHub, the key is ensuring that the benefits outweigh the harms). If you would like to make that case, I would love to read it.

chef-config/lib/chef-config/version.rb
```diff
@@ -14,9 +14,7 @@
 # limitations under the License.
 
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-# NOTE: This file is generated by running `rake version` in the top level of
-# this repo. Do not edit this manually. Edit the VERSION file and run the rake
-# task instead.
+# NOTE: This file is modified automatically by Expeditor. Do not modify it manually.
```
thommay: it is probably clearer to reference `update-version.sh` here, so it's clear what mechanism causes this update.

thommay:
The use of lita/expeditor has not changed, and is still as described in chef/chef-dk#1213 .

We're not keeping any "nice things" for ourselves; we're prioritising spending time on development of chef, rather than maintaining a changelog generator that was never especially good. As we discussed the last time round, there's nothing that leaves one beholden to any software, this is an automated way to take the title of a PR and drop it into a file. I'm pretty sure that that doesn't fit any definition of "important" that I'm aware of. 
Usage, for any maintainer - Chef employee or otherwise - is exactly the same as it has been through most of the 12 series; automatic. But should someone wish to edit the changelog, they now can. If anything, there is less "harm" since a maintainer (it's completely opaque to users) can choose to add, edit or remove lines as they see fit.
I'd agree some commentary about the tags in the changelog would be useful; please make that so @tduffield .

coderanger:
@thommay If that is indeed all that this is doing, then I agree, however there was minimal information given and so while I trust y'all, I can expect every use to do the same and so have to push hard on making sure this stuff gets fully explained one way or another. It would nice if I didn't have to do this each time Lita et al comes up but as someone external it feels like there is a gulf between internal and external understanding on this whole topic. I realize the tools are likely normalized to the point of forgetting they exist internally, but this is not the case when you work with the external community.

PR #6289, opened 2017-07-20
chef server rejects metadata with nil maintainer

lamont-granquist:
chef server returns a 400.  we could prune nil values out of the
JSON but then it'd be difficult to specify a "null" JSON value if we
really needed one.  seems better to follow the pattern of using
empty string here (which i tested against Hosted and works).


lamont-granquist:
@chef/client-core fairly easy review i think

thommay:
travis is grouchy

lamont-granquist:
travis degrouchification was successful

PR #6292, opened 2017-07-24
Updating / fixing deps to fix acceptance

btm:


btm:
Building in chef-build 152

lamont-granquist:
you probably want to just leave bumping anything kitchen-tests out of it, that doesn't get used in the acceptance environment.

lamont-granquist:
omnibus/Gemfile.lock should also be installed into the omnibus-toolchain ruby and not into /opt/chef so shouldn't be the source of any duplicated gems.  AFAIK only the root Gemfile.lock and acceptance/Gemfile.lock should need bumping to fix the issue.

btm:
Well this is more correct, but doesn't appear to fix acceptance in manhattan still.

Issue #5809, opened 2017-02-10
Port windows_path from the windows cookbook to core chef


tas50:
## Description

The windows_path resource is one of the more common resources used by community cookbooks, and its use requires adding the windows cookbook as a dependency throughout the chef ecosystem. To improve the out of the box experience with chef we should move this resource into core so that users can easily write cookbooks without pulling in a dependency.

https://github.com/chef-cookbooks/windows#windows_path

PR #6295, opened 2017-07-27
Add windows_path resource from the Windows cookbook

NimishaS:
### Description

Porting `windows_path` resource from the windows cookbook to core chef

### Issues Resolved
fixes https://github.com/chef/chef/issues/5809

### Check List
- [x] All tests pass


NimishaS:
Specs are remaining.

lib/chef/provider/windows_path.rb
```diff
@@ -0,0 +1,62 @@
...
+        end
+      end
+
+      # Expands the environment variables
+      def expand_env_vars(path)
+        # We pick 32k because that is the largest it could be:
+        # http://msdn.microsoft.com/en-us/library/windows/desktop/ms724265%28v=vs.85%29.aspx
+        buf = 0.chr * 32 * 1024 # 32k
+        if Chef::Provider::WindowsPath::ExpandEnvironmentStrings.call(path.dup, buf, buf.length) == 0
+          raise Chef::Exceptions::Win32APIError, "Failed calling ExpandEnvironmentStrings (received 0)"
```
stevendanna: Would GetLastError typically return something useful for us here?

NimishaS: @stevendanna, fixed review comment.

NimishaS:
The Appveyor and Travis failure don't seem to be related to this PR.

thommay:
The travis failures aren't present on master

spec/unit/provider/windows_path_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      allow(provider).to receive(:expand_env_vars)
+      expect(provider).to receive(:declare_resource).with(:env, "path")
+      provider.run_action(:remove)
+    end
+  end
+
+  describe "#expand_env_vars", :windows_only do
+    context "when ExpandEnvironmentStrings succeeds" do
+      it "returns the expanded environment variable" do
+        expanded_var = provider.expand_env_vars("some_path")
```
stevendanna: It might be better to test this behavior on the resource as a whole rather than testing this explicit method; however, if that is overly difficult for some reason this is fine.

It might also be better to test a string that gets mutated by ExpandEnvironmentStrings().  Perhaps something that uses one of these: https://technet.microsoft.com/en-us/library/dd560744(v=ws.10).aspx

lib/chef/provider/windows_path.rb
```diff
@@ -0,0 +1,62 @@
...
+        @current_resource
+      end
+
+      action :add do
+        declare_resource(:env, "path") do
+          action :modify
+          delim ::File::PATH_SEPARATOR
+          value new_resource.path.tr("/", '\\')
+        end
+        ENV["PATH"] = expand_env_vars(ENV["PATH"])
```
stevendanna: A comment on why this is necessary might be good, especially since we don't do an equivalent in the `remove` action.

lib/chef/provider/windows_path.rb
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+#
+
+require "Win32API" if Chef::Platform.windows?
+require "chef/exceptions"
+
+class Chef
+  class Provider
+    class WindowsPath < Chef::Provider
+      ExpandEnvironmentStrings = Win32API.new("kernel32", "ExpandEnvironmentStrings", %w{ P P L }, "L") if Chef::Platform.windows? && !defined?(ExpandEnvironmentStrings)
```
btm: this is a new pattern.

https://github.com/cosmo0920/win32-api/blob/master/README#L55

>  The current Win32API library that ships with the standard library has been slated for removal from Ruby 2.0 and it will not receive any updates in the Ruby 1.8.x branch.

This ruby bug says a similar thing but isn't current: https://bugs.ruby-lang.org/issues/8601

https://github.com/cosmo0920/win32-api/blob/master/README#L109

> I really don't have future plans for this library since you should use FFI as your preferred C interface going forward.

It seems like we need to add ExpandEnvironmentStrings to `lib/chef/win32/api.rb` instead?

btm: https://github.com/ruby/ruby/blob/ruby_2_4/ext/win32/lib/Win32API.rb#L5

> warn "Warning:#{caller[0].sub(/:in `.*'\z/, '')}: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead" if $VERBOSE

lib/chef/provider/windows_path.rb
```diff
@@ -0,0 +1,65 @@
...
+          end
+
+          # The windows Env provider does not correctly expand variables in
+          # the PATH environment variable. Ruby expects these to be expanded.
+          #
+          declare_resource(:ruby_block, "fix ruby ENV['PATH']") do
+            block do
+              ENV['PATH'] = expand_path(ENV['PATH'])
+            end
+            action :nothing
```
NimishaS: Noticed that the environment variable string isn't getting expanded even after calling `expand_path` method. The reason is that when this block executes, `ENV['PATH']` still doesn't have user provided path appended. Working on it.

lib/chef/resource/windows_path.rb
```diff
@@ -0,0 +1,41 @@
...
+class Chef
+  class Resource
+    class WindowsPath < Chef::Resource
+
+      provides :windows_path, os: "windows"
+
+      allowed_actions :add, :remove
+      default_action :add
+
+      def initialize(name, run_context = nil)
```
NimishaS: @tas50 , @btm, strangely if I remove the initialize block, I get error:
```
Chef::Exceptions::ProviderNotFound
Cannot find a provider for windows_path[%WINDIR%] on windows version 6.1.7601
```
even though `provides :windows_path, os: "windows"` is defined above.

PR #6297, opened 2017-07-29
README improvement for Chef beginner.

takaya-fuj19:
### Description

Since I am a new user of Chef, I was confused about how to start Chef development.
It is better to divide User and Developer instructions clearly in README.md.
For example,

- Add "Getting Started" section for User
- Change "Installing From Git" section to "Installing From Git for Developers"

It is easy to understand to add the following link to "Getting Started" as the reference for User.
https://docs.chef.io/quick_start.html

### Issues Resolved

README.md improvement.


tas50:
Non-IP change so no DCO required for the merge

PR #6299, opened 2017-07-31
Set explicit page size for every search request

### LINKED ISSUE #4329: Chef::Search::Query getting stack level too deep (SystemStackError)
Hi guys,
we currently have 4500 nodes in our fleet all managed by Chef (which is awesome :) ) and we have a script that runs a  query to get the last time a node has reported to the chef server so we can get all those nodes that have not called home in more than 24 hours and report this through email and we can check what is going on with them. 
This script used to run ok but it is now throwing the following error:
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.4.1/lib/chef/http.rb:153: stack level too deep (SystemStackError)

any ideas what could be causing it? 
Thanks in advance

### LINKED ISSUE #5783: stack level too deep (SystemStackError)
## Description

I understand that this repo is not for Chef Vault, but I think Chef Vault's usage of the Chef Client libraries is enough to make it relevant here. Anyways...

When running knife vault refresh I get a failure due to **stack level too deep (SystemStackError)**. From what I understand this is caused by too many nested processes. I added some debug code to see just how many levels deep the error was being thrown at. When it fails, **caller.size** reports the process is at 7285. This seems incredibly deep to me and given that I'm not surprised by the error. We thought it might have be correlated to the number of nodes/clients we have, but after culling a few hundred of them the issue persists. We only just started seeing this issue. We have not upgraded recently. 

Thoughts??

## Chef Version

Chef Server 12.7, Chef Client 12.15.19

## Platform Version

Ubuntu 14.04 

## Stacktrace

/opt/chef/embedded/lib/ruby/gems/2.3.0/gems/mixlib-log-1.7.1/lib/mixlib/log/formatter.rb:45:in `msg2str': stack level too deep (SystemStackError)
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/mixlib-log-1.7.1/lib/mixlib/log/formatter.rb:36:in `call'
        from /opt/chef/embedded/lib/ruby/2.3.0/logger.rb:546:in `format_message'
        from /opt/chef/embedded/lib/ruby/2.3.0/logger.rb:434:in `add'
        from /opt/chef/embedded/lib/ruby/2.3.0/logger.rb:455:in `debug'
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/mixlib-log-1.7.1/lib/mixlib/log.rb:120:in `block in debug'
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/mixlib-log-1.7.1/lib/mixlib/log.rb:120:in `each'
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/mixlib-log-1.7.1/lib/mixlib/log.rb:120:in `debug'
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.15.19/lib/chef/http/basic_client.rb:75:in `block (2 levels) in request'
         ... 7278 levels...
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.15.19/lib/chef/application/knife.rb:156:in `run'
        from /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.15.19/bin/knife:25:in `<top (required)>'
        from /usr/bin/knife:54:in `load'
        from /usr/bin/knife:54:in `<main>'

stevendanna:
The previous code would advance the search start parameter by either
the user-provided rows parameter or (as a fallback) the number of rows
returned in that page.  Most code that calls the search() function
does not set rows, so the fallback is used often.

Since the search index may have stale results, it is possible to get
responses with a completely empty page. For example:

    {
       "start": 0,
       "total": 1,
       "rows": []
    }

In this case, if no rows parameter was passed, the search function
would infinitely recurse (eventually running out of stack).

To avoid this, we set rows to 1000 if no user-provided value was set.
This is the default page size of Chef Server so for most users it will
not be changing the default behavior.

Signed-off-by: Steven Danna <steve@chef.io>

### Issues Resolved

Fixes #4329
Fixes #5783

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/search/query.rb
+++ b/lib/chef/search/query.rb
@@ -71,6 +71,11 @@ class Chef
           args_h = args_h.reject { |k, v| k == :fuzz }
         end
 
+        # Set default rows parameter to 1000. This is the default in
+        # Chef Server, but we set it explicitly here so that we can
+        # confidently advance our start parameter.
+        args_h[:rows] ||= 1000
+
         response = call_rest_service(type, query: query, **args_h)
 
         if block
@@ -87,7 +92,7 @@ class Chef
           # args_h[:rows] to avoid asking the search backend for
           # overlapping pages (which could result in duplicates).
           #
-          next_start = response["start"] + (args_h[:rows] || response["rows"].length)
+          next_start = response["start"] + args_h[:rows]
           unless next_start >= response["total"]
             args_h[:start] = next_start
             search(type, query, args_h, &block)
--- a/spec/functional/knife/ssh_spec.rb
+++ b/spec/functional/knife/ssh_spec.rb
@@ -304,7 +304,7 @@ describe Chef::Knife::Ssh do
     Chef::Config[:client_key] = nil
     Chef::Config[:chef_server_url] = "http://localhost:9000"
 
-    @api.post("/search/node?q=*:*&start=0", 200) do
+    @api.post("/search/node?q=*:*&start=0&rows=1000", 200) do
       %({"total":1, "start":0, "rows":[{"data": {"fqdn":"the.fqdn", "config": "the_public_hostname", "knife_config": "the_public_hostname" }}]})
     end
   end
--- a/spec/unit/search/query_spec.rb
+++ b/spec/unit/search/query_spec.rb
@@ -22,9 +22,10 @@ require "chef/search/query"
 describe Chef::Search::Query do
   let(:rest) { double("Chef::ServerAPI") }
   let(:query) { Chef::Search::Query.new }
+  let(:default_rows) { 1000 }
 
   shared_context "filtered search" do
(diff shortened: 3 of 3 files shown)


stevendanna:
I think something like this is the solution to #4329 and similar issues.

stevendanna:
> Probably wouldn't turn to have an explicit check somewhere though, like if you explicitly pass in rows: 0 here I think this still blows up.

I could see wanting to special case 0 since I've seen users use it to get the count of how many nodes match a search without pulling down the full node object.

thommay:
Looks good but travis and appveyor are unhappy, @stevendanna

stevendanna:
@thommay Fixed up the tests.

PR #6312, opened 2017-08-09
[MSYS-594] windows_task resource is not idempotent when specifying start_time and start_day

### LINKED ISSUE #6093: Chef13 windows_task resource is not idempotent when specifying start_time
## Description

The new windows_task resource in Chef13 is not idempotent when specifying start_time.  The resource will always execute during a chef-client run even if it has not been changed in any way.

Unfortunately I have not tried to use the older windows_task resource from the Chef Windows cookbook.  I am not aware if this behavior existed in that cookbook.

## Chef Version

Tell us which version of chef-client you are using (see below for Server+ChefDK bugs).

## Platform Version

Windows 2012R2 Chef 13.0.118

## Replication Case

Create a windows_task resource and specify a start time.  Run chef-client twice and note that the resource executes each time.

windows_task 'test task' do
  command "echo test"
  cwd "c:/windows"
  start_time "02:00"
  frequency :daily
end
  

## Client Output

-----> Converging <default-windows-2012r2-D5GB>...
       Preparing files for transfer
       Preparing dna.json
       Preparing current project directory as a cookbook
       Removing non-cookbook files before transfer
       Preparing data_bags
       Preparing validation.pem
       Preparing client.rb
-----> Chef Omnibus installation detected (install only if missing)
       Transferring files to <default-windows-2012r2-D5GB>
       Starting Chef Client, version 13.0.118
       resolving cookbooks for run list: ["Perforce-ProxySetup::default"]
       Synchronizing Cookbooks:
         - Perforce-ProxySetup (2015.2.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       ...
       windows_task[test task] action create
       ...
Running handlers:
Running handlers complete
Chef Client finished, 1/98 resources updated in 06 seconds
Finished converging <default-windows-2012r2-D5GB> (0m13.07s).

harikesh-kolekar:
### Description
windows_task resource is not idempotent when specifying start_time
fixes #6093
### Check List

- [ ] All tests pass

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ gem "ohai", "~> 13"
 
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "cheffish", "~> 13" # required for rspec tests
+gem "american_date"
 
 group(:omnibus_package) do
   gem "appbundler"
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,6 @@ gem "ohai", "~> 13"
 
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "cheffish", "~> 13" # required for rspec tests
-gem "american_date"
 
 group(:omnibus_package) do
   gem "appbundler"
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -30,32 +30,32 @@ class Chef
       provides :windows_task, os: "windows"
 
       def load_current_resource
-        @current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
+        self.current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
         pathed_task_name = new_resource.task_name.start_with?('\\') ? new_resource.task_name : "\\#{new_resource.task_name}"
 
-        @current_resource.task_name(pathed_task_name)
+        current_resource.task_name(pathed_task_name)
         task_hash = load_task_hash(pathed_task_name)
 
         set_current_resource(task_hash) if task_hash.respond_to?(:[]) && task_hash[:TaskName] == pathed_task_name
-        @current_resource
+        current_resource
       end
 
       def set_current_resource(task_hash)
-        @current_resource.exists = true
-        @current_resource.command(task_hash[:TaskToRun])
-        @current_resource.cwd(task_hash[:StartIn]) unless task_hash[:StartIn] == "N/A"
-        @current_resource.user(task_hash[:RunAsUser])
+        current_resource.exists = true
+        current_resource.command(task_hash[:TaskToRun])
+        current_resource.cwd(task_hash[:StartIn]) unless task_hash[:StartIn] == "N/A"
+        current_resource.user(task_hash[:RunAsUser])
         set_current_run_level task_hash[:run_level]
         set_current_frequency task_hash
-        @current_resource.day(task_hash[:day]) if task_hash[:day]
-        @current_resource.months(task_hash[:months]) if task_hash[:months]
+        current_resource.day(task_hash[:day]) if task_hash[:day]
+        current_resource.months(task_hash[:months]) if task_hash[:months]
         set_current_idle_time(task_hash[:idle_time]) if task_hash[:idle_time]
-        @current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
-        @current_resource.execution_time_limit(task_hash[:execution_time_limit] || "PT72H")
-        @current_resource.status = :running if task_hash[:Status] == "Running"
-        @current_resource.enabled = true if task_hash[:ScheduledTaskState] == "Enabled"
-        @current_resource.start_time = task_hash[:StartTime] if task_hash[:StartTime]
-        @current_resource.start_day = task_hash[:StartDate] if task_hash[:StartDate]
+        current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
+        current_resource.execution_time_limit(task_hash[:execution_time_limit] || "PT72H")
+        current_resource.status = :running if task_hash[:Status] == "Running"
+        current_resource.enabled = true if task_hash[:ScheduledTaskState] == "Enabled"
+        current_resource.start_time = task_hash[:StartTime] if task_hash[:StartTime]
+        current_resource.start_day = task_hash[:StartDate] if task_hash[:StartDate]
       end
 
       # This method checks if task and command attributes exist since those two are mandatory attributes to create a schedules task.
@@ -73,7 +73,7 @@ class Chef
       end
 
       def action_create
-        if @current_resource.exists
+        if current_resource.exists
           if !(task_need_update? || new_resource.force)
             Chef::Log.info "#{new_resource} task already exists - nothing to do"
             return
@@ -81,7 +81,7 @@ class Chef
           # To merge current resource and new resource attributes
           resource_attributes.each do |attribute|
             new_resource_attribute = new_resource.send(attribute)
-            current_resource_attribute = @current_resource.send(attribute)
+            current_resource_attribute = current_resource.send(attribute)
             new_resource.send("#{attribute}=", current_resource_attribute ) if current_resource_attribute && new_resource_attribute.nil?
           end
         end
@@ -91,7 +91,7 @@ class Chef
         options["SC"] = schedule
         options["MO"] = new_resource.frequency_modifier if frequency_modifier_allowed
         options["I"]  = new_resource.idle_time unless new_resource.idle_time.nil?
-        options["SD"] = convert_user_date_to_system_date unless new_resource.start_day.nil?
+        options["SD"] = convert_user_date_to_system_date new_resource.start_day unless new_resource.start_day.nil?
         options["ST"] = new_resource.start_time unless new_resource.start_time.nil?
         options["TR"] = new_resource.command
         options["RU"] = new_resource.user
@@ -112,8 +112,8 @@ class Chef
       end
 
       def action_run
-        if @current_resource.exists
-          if @current_resource.status == :running
+        if current_resource.exists
+          if current_resource.status == :running
             Chef::Log.info "#{new_resource} task is currently running, skipping run"
           else
             run_schtasks "RUN"
@@ -126,7 +126,7 @@ class Chef
       end
 
       def action_delete
-        if @current_resource.exists
+        if current_resource.exists
           # always need to force deletion
           run_schtasks "DELETE", "F" => ""
           new_resource.updated_by_last_action true
@@ -137,8 +137,8 @@ class Chef
       end
 
       def action_end
-        if @current_resource.exists
-          if @current_resource.status != :running
+        if current_resource.exists
+          if current_resource.status != :running
             Chef::Log.debug "#{new_resource} is not running - nothing to do"
           else
             run_schtasks "END"
@@ -151,8 +151,8 @@ class Chef
       end
 
       def action_enable
-        if @current_resource.exists
-          if @current_resource.enabled
+        if current_resource.exists
+          if current_resource.enabled
             Chef::Log.debug "#{new_resource} already enabled - nothing to do"
           else
             run_schtasks "CHANGE", "ENABLE" => ""
@@ -166,8 +166,8 @@ class Chef
       end
 
       def action_disable
-        if @current_resource.exists
-          if @current_resource.enabled
+        if current_resource.exists
+          if current_resource.enabled
             run_schtasks "CHANGE", "DISABLE" => ""
             new_resource.updated_by_last_action true
             Chef::Log.info "#{new_resource} task disabled"
@@ -196,20 +196,20 @@ class Chef
 
       def task_need_update?
         return true if (new_resource.command &&
-            @current_resource.command != new_resource.command.tr("'", '"')) ||
-            @current_resource.user != new_resource.user ||
-            @current_resource.run_level != new_resource.run_level ||
-            @current_resource.cwd != new_resource.cwd ||
-            @current_resource.frequency_modifier != new_resource.frequency_modifier ||
-            @current_resource.frequency != new_resource.frequency ||
-            @current_resource.idle_time != new_resource.idle_time ||
-            @current_resource.random_delay != new_resource.random_delay ||
-            !new_resource.execution_time_limit.include?(@current_resource.execution_time_limit) ||
-            (!new_resource.start_day.nil? && start_day_updated?) ||
-            (!new_resource.start_time.nil? && start_time_updated?)
+            current_resource.command != new_resource.command.tr("'", '"')) ||
+            current_resource.user != new_resource.user ||
+            current_resource.run_level != new_resource.run_level ||
+            current_resource.cwd != new_resource.cwd ||
+            current_resource.frequency_modifier != new_resource.frequency_modifier ||
+            current_resource.frequency != new_resource.frequency ||
+            current_resource.idle_time != new_resource.idle_time ||
+            current_resource.random_delay != new_resource.random_delay ||
+            !new_resource.execution_time_limit.include?(current_resource.execution_time_limit) ||
+            (new_resource.start_day && start_day_updated?) ||
+            (new_resource.start_time && start_time_updated?)
         begin
-          return true if new_resource.day.to_s.casecmp(@current_resource.day.to_s) != 0 ||
-              new_resource.months.to_s.casecmp(@current_resource.months.to_s) != 0
+          return true if new_resource.day.to_s.casecmp(current_resource.day.to_s) != 0 ||
+              new_resource.months.to_s.casecmp(current_resource.months.to_s) != 0
         rescue
           Chef::Log.debug "caught a raise in task_needs_update?"
         end
@@ -218,18 +218,18 @@ class Chef
       end
 
       def start_day_updated?
-        current_day = convert_system_date_format_into_ruby_date(@current_resource.start_day)
+        current_day = convert_system_date_format_into_ruby_date(current_resource.start_day)
         new_day = convert_system_date_format_into_ruby_date(new_resource.start_day)
         current_day != new_day
       end
 
       def start_time_updated?
-        time = convert_system_date_format_into_ruby_date(@current_resource.start_time).strftime("%H:%M")
+        time = convert_system_date_format_into_ruby_date(current_resource.start_time).strftime("%H:%M")
         time != new_resource.start_time
       end
 
       def convert_system_date_format_into_ruby_date(date_in_string)
-        Chef::Log.debug "Looking system date format"
+        Chef::Log.debug "Converting '#{date_in_string}' to '\%Y-\%m-\%d \%H:\%M'"
         task_script = <<-EOH
           [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
           ([datetime]'#{date_in_string}').ToString('yyyy-MM-dd HH:mm')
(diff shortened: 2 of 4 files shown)


lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,35 @@ def task_need_update?
         false
       end
 
+      def compare_day
```
NimishaS: Since this function checks if the `start_day` has been updated and returns a boolean value based on that, it would be good to name it as `start_day_updated?`.

lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,35 @@ def task_need_update?
...
+        current_day = convert_system_dateformat_into_ruby_date(@current_resource.start_day).strftime("%m/%d/%Y")
+        new_day = convert_system_dateformat_into_ruby_date(new_resource.start_day).strftime("%m/%d/%Y")
+        current_day != new_day
+      end
+
+      def compare_time
```
NimishaS: It would be good to name this function as `start_time_updated?`.

lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,35 @@ def task_need_update?
...
+        time = convert_system_dateformat_into_ruby_date(@current_resource.start_time).strftime("%H:%M")
+        time != new_resource.start_time
+      end
+
+      def convert_system_dateformat_into_ruby_date(date_in_string)
+        Chef::Log.debug "Looking system date format"
+        task_script = <<-EOH
+          [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+          ([datetime]'#{date_in_string}').ToString('yyyy-MM-dd HH:mm')
+        EOH
```
NimishaS: Please add some comments to explain this code. Comments can be added here or at the place where this method is being called.

lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,35 @@ def task_need_update?
...
+        date_time = powershell_out(task_script).stdout.force_encoding("UTF-8").gsub(/[\s+\uFEFF]/, "")
+        DateTime.strptime(date_time, "%Y-%m-%d %H:%M")
+      end
+
+      def convert_ruby_date_to_system_date
+        task_script = <<-EOH
+          [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+          ([datetime]"#{new_resource.start_day}").ToString([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern)
+        EOH
```
NimishaS: Please add some comments to explain this code. Comments can be added here or at the place where this method is being called.

NimishaS:
Please add unit testcases for the new functions added.

lib/chef/provider/windows_task.rb
```diff
@@ -51,9 +51,11 @@ def set_current_resource(task_hash)
         @current_resource.months(task_hash[:months]) if task_hash[:months]
         set_current_idle_time(task_hash[:idle_time]) if task_hash[:idle_time]
         @current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
-        @current_resource.execution_time_limit(task_hash[:execution_time_limit]) if task_hash[:execution_time_limit]
+        @current_resource.execution_time_limit(task_hash[:execution_time_limit] || "PT72H")
```
btm: 72 hours seems long, but this is an arbitrary setting, right?

lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,38 @@ def task_need_update?
...
+        current_day != new_day
+      end
+
+      def start_time_updated?
+        time = convert_system_date_format_into_ruby_date(@current_resource.start_time).strftime("%H:%M")
+        time != new_resource.start_time
+      end
+
+      def convert_system_date_format_into_ruby_date(date_in_string)
+        Chef::Log.debug "Looking system date format"
```
btm: Is this debug still useful? I think it can be deleted. Otherwise I'd change this to something like `"Converting '#{date_in_string}' to system date format"`

lib/chef/provider/windows_task.rb
```diff
@@ -51,9 +51,11 @@ def set_current_resource(task_hash)
         @current_resource.months(task_hash[:months]) if task_hash[:months]
         set_current_idle_time(task_hash[:idle_time]) if task_hash[:idle_time]
         @current_resource.random_delay(task_hash[:random_delay]) if task_hash[:random_delay]
-        @current_resource.execution_time_limit(task_hash[:execution_time_limit]) if task_hash[:execution_time_limit]
+        @current_resource.execution_time_limit(task_hash[:execution_time_limit] || "PT72H")
         @current_resource.status = :running if task_hash[:Status] == "Running"
         @current_resource.enabled = true if task_hash[:ScheduledTaskState] == "Enabled"
+        @current_resource.start_time = task_hash[:StartTime] if task_hash[:StartTime]
+        @current_resource.start_day = task_hash[:StartDate] if task_hash[:StartDate]
```
btm: `current_resource` set as an accessor in Chef::Provider, so these lines could be updated while you're here to not use the instance variable.

lib/chef/provider/windows_task.rb
```diff
@@ -202,8 +204,9 @@ def task_need_update?
             @current_resource.frequency != new_resource.frequency ||
             @current_resource.idle_time != new_resource.idle_time ||
             @current_resource.random_delay != new_resource.random_delay ||
-            @current_resource.execution_time_limit != new_resource.execution_time_limit ||
-            !new_resource.start_day.nil? || !new_resource.start_time.nil?
+            !new_resource.execution_time_limit.include?(@current_resource.execution_time_limit) ||
+            (!new_resource.start_day.nil? && start_day_updated?) ||
```
btm: `new_resource.start_day` instead of `!new_resource.start_day.nil?` seems like a better pattern here.

btm: (and above and below)

lib/chef/provider/windows_task.rb
```diff
@@ -214,6 +217,38 @@ def task_need_update?
...
+        DateTime.strptime(date_time, "%Y-%m-%d %H:%M")
+      end
+
+      def convert_user_date_to_system_date
+        task_script = <<-EOH
+          [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+          ([datetime]"#{new_resource.start_day}").ToString([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern)
+        EOH
+        # Convert start_date into OS date format
+        powershell_out(task_script).stdout.force_encoding("UTF-8").gsub(/[\s+\uFEFF]/, "")
```
btm: Do we need to shell out like this to know what the system date format is? Could we do without the shelling out? It's slow.

Why can't we do something like this:
```
 DateTime.parse('8/3/2017 15:12').strftime("%Y-%m-%d %H:%M")
```

lib/chef/provider/windows_task.rb
harikesh-kolekar: @btm By default on windows 2008 and 2012 system `execution_time_limit` return "PT72H" and for windows 2016 it returns nil so idempotent check fails when the user executes resource without  `execution_time_limit` on windows 2016.
https://github.com/chef/chef/blob/master/lib/chef/resource/windows_task.rb#L77

lib/chef/provider/windows_task.rb
harikesh-kolekar: @btm The `start_day`  to execute the command, is required in system date format and if two nodes are in diff system DateTime format then the resource execution will fail so need to convert system date format.
like if node A date format in mm-dd-yyyy and node B date format in yyyy-mm-dd and user has passed date like 11-11-2015 then for node B  the resource will fail to execute.
and the command execution needs system date format so we can't do like 
```
DateTime.parse('8/3/2017 15:12').strftime("%Y-%m-%d %H:%M")
```

lib/chef/provider/windows_task.rb
harikesh-kolekar: @btm After replacing `@current_resource` as `current_resource` throws following error
```
>>>> Caused by NoMethodError: undefined method `exists=' for nil:NilClass
C:/Users/azure/chef/chef/lib/chef/provider/windows_task.rb:44:in `set_current_resource'
```
https://github.com/chef/chef/blob/master/lib/chef/provider/windows_task.rb#L44

btm: Change only line 33 to this:
```
self.current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
```

I know this is confusing. What is happening here without the `self` is you're assigning `Chef::Resource::WindowsTask.new(new_resource.name)` to a local variable named `current_resource`. Once you're outside of `load_current_resource`, that variable no longer exists.

In this case, we already have an accessor method named `current_resource`, so it seems unnecessarily confusing for Ruby to also allow the creation of a local variable named `current_resource`, but it does. So we have be explicit that we mean the accessor method. Later when we're calling methods against the object stored in `current_resource`, we don't have to be explicit.

In most of our current code we leave the @ on this one line, which also works. Using the accessor instead of the instance variable makes it easier to change behavior later. We could change the accessor to a normal method and manipulate the data that we return. In this case, we'd end up going back and changing the one line of `@current_resource...` to `self.current_resource`.

lib/chef/provider/windows_task.rb
btm: please add a comment here about this setting. Otherwise someone will come along and think 72h is too long and change it and the tests to something shorter and we might not realize until after  we release.

lib/chef/provider/windows_task.rb
btm: Do you have access to a system where the system format is different? What do these two methods return on that system in Ruby?

```
Time.new.strftime('%x %X')
Time.new.strftime('%D %T')
```

mwrock: agree with @btm. It would be preferable to do the date comparisons in ruby

harikesh-kolekar: @btm , @mwrock I have removed `convert_system_date_format_into_ruby_date` function and used `DateTime.parse` to convert string into ruby date object  
https://github.com/chef/chef/pull/6312/commits/1a8f50ac05609e7bbbb3dd70c583d11bc7414753

But to execute window_task command(schtasks) `start_day` must be in the system date format. By default windows system date is in `MM/dd/yyyy` format, but the user can change it to any other format e.g. yyyy-MM-dd or dd-MM-yyyy, etc. So we need to convert `start_day` into system date format

The Time.new.strftime('%x %X') Time.new.strftime('%D %T') gives date into `MM/dd/yy` format it's output does not depend on system date format

```
irb(main):001:0> Time.new.strftime('%x %X')
=> "08/30/17 18:51:06"
irb(main):002:0> Time.new.strftime('%D %T')
=> "08/30/17 19:00:27"
```

*Another Solution:* We can restrict user to give start_day as input to windows_task resource into  system date format 
*Problem:* If node A has default date format(M/d/yyyy) and for node B user had changed to d-M-yy and the user has passed date in dafault date format(M/d/yyyy) from recipe then the execution fails for node B because start_day required in system date format.

mwrock: Forcing the user to use the system format could prove overly restrictive and break some resources. While doing this in ruby would be ideal, resorting to shelling out is probably acceptable.

lib/chef/provider/windows_task.rb
```diff
@@ -194,26 +196,48 @@ def run_schtasks(task_action, options = {})
...
+      def start_time_updated?
+        time = DateTime.parse(current_resource.start_time).strftime("%H:%M")
+        time != new_resource.start_time
+      end
+
+      def convert_user_date_to_system_date(date_in_string)
+        Chef::Log.debug "Converting '#{date_in_string}' to system date format"
+        task_script = <<-EOH
+          [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+          ([datetime]"#{date_in_string}").ToString((((((([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern -replace "dd", "d") -replace "d", "dd") -replace "MM", "M")-replace "M","MM") -replace "yyyy","yy") -replace "y", "yy"))
```
stuartpreston: Has this been tested against multiple locales? On my system which is in English (United Kingdom) this line would not accept a date formatted in the ShortDatePattern for my locale:

```
PS C:\Users\spreston> get-culture

LCID             Name             DisplayName                                                                                                      
----             ----             -----------                                                                                                      
2057             en-GB            English (United Kingdom)  
```

```
([datetime]"30/08/2017").ToString((((((([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern -replace "dd", "d") -replace "d", "dd") -replace "MM", "M")-replace "M","MM") -replace "yyyy","yy") -replace "y", "yy"))
Cannot convert value "30/08/2017" to type "System.DateTime". Error: "String was not recognized as a valid DateTime."
At line:1 char:1
+ ([datetime]"30/08/2017").ToString((((((([Globalization.Cultureinfo]:: ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [], RuntimeException
```

But it does take an English (US) format date:
```
([datetime]"08/30/2017").ToString((((((([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern -replace "dd", "d") -replace "d", "dd") -replace "MM", "M")-replace "M","MM") -replace "yyyy","yy") -replace "y", "yy"))
30/08/2017
```

```
PS C:\Users\spreston> [Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern
dd/MM/yyyy

PS C:\Users\spreston> [Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat


AMDesignator                     : AM
Calendar                         : System.Globalization.GregorianCalendar
DateSeparator                    : /
FirstDayOfWeek                   : Monday
CalendarWeekRule                 : FirstFourDayWeek
FullDateTimePattern              : dd MMMM yyyy HH:mm:ss
LongDatePattern                  : dd MMMM yyyy
LongTimePattern                  : HH:mm:ss
MonthDayPattern                  : d MMMM
PMDesignator                     : PM
RFC1123Pattern                   : ddd, dd MMM yyyy HH':'mm':'ss 'GMT'
ShortDatePattern                 : dd/MM/yyyy
ShortTimePattern                 : HH:mm
SortableDateTimePattern          : yyyy'-'MM'-'dd'T'HH':'mm':'ss
TimeSeparator                    : :
UniversalSortableDateTimePattern : yyyy'-'MM'-'dd HH':'mm':'ss'Z'
YearMonthPattern                 : MMMM yyyy
AbbreviatedDayNames              : {Sun, Mon, Tue, Wed...}
ShortestDayNames                 : {Su, Mo, Tu, We...}
DayNames                         : {Sunday, Monday, Tuesday, Wednesday...}
AbbreviatedMonthNames            : {Jan, Feb, Mar, Apr...}
MonthNames                       : {January, February, March, April...}
IsReadOnly                       : False
NativeCalendarName               : Gregorian Calendar
AbbreviatedMonthGenitiveNames    : {Jan, Feb, Mar, Apr...}
MonthGenitiveNames               : {January, February, March, April...}
```

lib/chef/provider/windows_task.rb
```diff
@@ -194,26 +196,48 @@ def run_schtasks(task_action, options = {})
...
+        new_day = DateTime.parse(new_resource.start_day)
+        current_day != new_day
+      end
+
+      def start_time_updated?
+        time = DateTime.parse(current_resource.start_time).strftime("%H:%M")
+        time != new_resource.start_time
+      end
+
+      def convert_user_date_to_system_date(date_in_string)
```
stuartpreston: I am suspicious of the need to create a PowerShell process here?

spec/unit/provider/windows_task_spec.rb
```diff
@@ -330,6 +330,9 @@
         new_resource.frequency_modifier 15
         new_resource.user "SYSTEM"
         new_resource.execution_time_limit "PT72H"
+        new_resource.start_day "3/30/2017"
+        new_resource.start_time "13:12"
+        date_time_obj = DateTime.strptime("#{new_resource.start_day} #{new_resource.start_time}", "%m/%d/%Y %H:%M")
```
stuartpreston: Again, this spec appears to deal with English (US) systems, which may be fine for our automated testing but is not guaranteed to be the format the user provides in the resource

lib/chef/provider/windows_task.rb
```diff
@@ -194,26 +196,56 @@ def run_schtasks(task_action, options = {})
...
+      end
+
+      def start_time_updated?
+        time = to_date_time_obj(current_resource.start_time).strftime("%H:%M")
+        time != new_resource.start_time
+      end
+
+      def to_date_time_obj(date_in_string)
+        return DateTime.parse(date_in_string)
+      rescue
```
btm: Could you be more specific here with what we're rescuing?

lib/chef/provider/windows_task.rb
btm: see https://github.com/chef/chef/pull/6312#discussion_r136145254

btm:
oh, and rebase :)

lib/chef/provider/windows_task.rb
```diff
@@ -194,26 +196,56 @@ def run_schtasks(task_action, options = {})
...
+        require "american_date"
+        return DateTime.parse(date_in_string)
+      end
+
+      def convert_user_date_to_system_date(date_in_string)
+        Chef::Log.debug "Converting '#{date_in_string}' to system date format"
+        date_in_string = to_date_time_obj(date_in_string).strftime("%Y-%m-%d")
+        task_script = <<-EOH
+          [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+          ([datetime]"#{date_in_string}").ToString((((((([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern -replace "dd", "d") -replace "d", "dd") -replace "MM", "M")-replace "M","MM") -replace "yyyy","yy") -replace "y", "yy"))
```
stuartpreston: What's the value of doing the replacements in this line.  Can we not use something like the following:

```
C:\Users\spreston> ([datetime]"2017-09-06").ToString([Globalization.Cultureinfo]::CurrentCulture.DateTimeFormat.ShortDatePattern)
06/09/2017
```

stuartpreston: OK I see it myself, you're trying to protect against dates that have a single character in each section.  Such as 9/6/17.

Gemfile
```diff
@@ -11,6 +11,7 @@ gem "ohai", "~> 13"
 
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "cheffish", "~> 13" # required for rspec tests
+gem "american_date"
```
tas50: This seems like a gem we probably shouldn't be bringing in. It's about 50 lines of code and adds yet another dependency to Chef. Instead of adding this dependency can we just add the same logic to convert date stamps to the "american" style date stamp?

spuder:
Thank you for fixing this. How to use `start_day` is not yet documented. I've opened a new issue. 

https://github.com/chef/chef/issues/6401

kalpik:
This merge seems to have broken windows_task. If I don't specify a start_time now, it somehow passes a wrongly formatted start_time.

`windows_task[Chef client scheduled run] (sdl_tridion_common::create_chef_client_run_task line 37) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '-2147467259'
---- Begin output of schtasks /CREATE /TN "Chef client scheduled run" /F /SC "minute" /MO "30" /SD "10/17/2017" /ST "7:56:00 PM" /TR "C:\opscode\chef\bin\chef-client.bat -c c:\chef\client.rb -L C:/chef/log\chef_scheduled_run.log -l info" /RU "user" /RP "password" /RL "HIGHEST"  ----
STDOUT: 
STDERR: ERROR: Invalid starttime value.`

If you see, it's apparently passing `/ST "7:56:00 PM"`, which is wrong. It should be `/ST "19:56"`

harikesh-kolekar:
@kalpik, Please create a new issue for this bug as this PR is closed. We will be fixing it soon on new PR.

NimishaS:
@kalpik , fixing your issue in https://github.com/chef/chef/pull/6544.

PR #6313, opened 2017-08-10
Pass json configuration to ShellSession class

jonlives:
This commit passes json configuration to ShellSession class and thusly into ChefClient object created by the ShellSession. 

### Description

This commit fixes an issue in chef-shell whereby runlists passed in via a json file using the -j flag would not be loaded when running chef-shell in client mode using the -z flag.

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [X ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #6314, opened 2017-08-10
Pass json configuration to ShellSession class

btm:
Fixes an issue in chef-shell whereby runlists passed in via a json file using the -j flag would not be loaded when running chef-shell in client mode using the -z flag.

Adds unit tests on top of #6313


jonlives:
👍

PR #6315, opened 2017-08-10
Maintain backwards compatibility with apt_preference

tas50:
Remove did not previously require the pin and pin_priority. Also it doesn't make sense to require those for a remove action.


tas50:
The failure is actually a bug in how we display errors that was being hit because of a mismatch int he action names for apt_preference. I'll fix that elsewhere

PR #6318, opened 2017-08-10
Maintain compat with old zypper_repo resource used in cookbooks

tas50:


tas50:
@lamont-granquist mentioned they were different when I asked. We have them everywhere else. If we wanna nuke them we should nuke them all.

PR #6329, opened 2017-08-12
12: Bump omnibus-software to fix early Rubygems segfaults

btm:
Chef 12 builds were failing on Windows in under 3 minutes with a segfault running rubygems.

This passes the pipeline but eats Ruby 2.4. Not sure if we can do that.

http://manhattan.ci.chef.co/job/chef-build/202/


coderanger:
I'm pretty tepid at best on doing 2.4 for the Chef 12 and ChefDK 1 lines. Would it fix things if we upgraded to a newer point release of 2.3? If nothing else this will invalidate all installed gems.

btm:
I tried 2.3.3 at one point, I'll try 2.3.4.

coderanger:
Whatever we end up doing, we should probably keep ChefDK 1.x in sync with it, so that might complicate things further :(

btm:
@lamont-granquist is currently playing the same game with Chef DK, I bet whatever the fix here is will be the same there.

btm:
Okay, this builds now. Please review for any of the non-ruby changes that might be a concern.

After https://github.com/chef/omnibus-software/pull/876 is merged I'll fix the git reference to omnibus-software.

Gemfile
```diff
@@ -12,14 +12,12 @@ source "https://rubygems.org"
...
 
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
-gem "rake"
 gem "bundler"
 gem "cheffish" # required for rspec tests
 
 group(:omnibus_package) do
   gem "appbundler"
   gem "rb-readline"
-  gem "nokogiri", "~> 1.7.2"
```
lamont-granquist: that should still be in there, but the pin should be dropped i think

lamont-granquist: ugh, its the old stuff, right...

btm: I forget the exact message, but I needed to bump nokogiri. Now that I think about it, it might have been related to 2.4 so maybe I could back this out. I'm not sure which is better. But yeah, I copied the pattern from your big change in master.

btm:
I'm giving up on the remaining two Travis tests for now (chef-zero and halite).

PR #6332, opened 2017-08-14
Update Gemfile.lock without attempting to install Gems

tduffield:
### Description

Fix the auto-bumping by removing the bundle lock that was preventing us from updating the Gemfile.lock.

### Check List

- [ ] All tests pass


btm:
In case it's not obvious, the two travis failures are unrelated failures we haven't gotten around to fixing.

Issue #6322, opened 2017-08-11
Information from guards (not_if, only_if)


D33M:
## Description

Currently Chef displays generic information about why resource wasn't run, according to guards (`(skipped due to not_if)`). It would be nice if we could briefly describe this behaviour, without looking into cookbooks.

## Chef Version

```
Chef Development Kit Version: 2.0.26
chef-client version: 13.2.20
delivery version: master (17c1b0fed9be4c70f69091a6d21a4cbf0df60a23)
berks version: 6.2.0
kitchen version: 1.16.0
inspec version: 1.30.0
```

## Suggessted mechanism

It would be nice if guards were a little more expanded:
```
execute 'Hello' do
   command     'rm -rf /'
    not_if [
       {File::exist?('/please_dont'), :information => 'Since you are begging, it will be skipped'},
       {user_is_my_friend && user_paid, :information => 'You're my dude!'},
       {node['hostname'] == 'my.computer', :information => 'I won't give myself more work...'}
    ]
end
```

This way it would display descriptive information *why* it was skipped (without looking inside code and wondering, which part made it `true`):

```
* execute [Hello] action run  (skipped due to not_if: 'Since you are begging, it will be skipped')
```

coderanger:
Unfortunately that isn't valid Ruby code, the required gymnastics to work this into the DSL probably wouldn't be worth it.

D33M:
Yeah, I do know, it isn't valid, I just wanted to show some example how it would like - just to show what I have in mind. Syntax can be of course quite different.

PR #6340, opened 2017-08-19
Conditional Guard Description

jgorak:
### Description

Custom description option (:desc) on only_if/not_if guards for better identification in output
```
log "hello" do
  not_if(:desc => 'randomly skip') { rand < 0.5 }
end
```

### Issues Resolved

https://github.com/chef/chef/issues/6322

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -386,6 +386,10 @@ class Chef
     # @param block [Proc] A ruby block to run. Ignored if a command is given.
     #
     def only_if(command = nil, opts = {}, &block)
+      if command.is_a?(Hash) && opts.empty?
+        opts = command
+        command = nil
+      end
       if command || block_given?
         @only_if << Conditional.only_if(self, command, opts, &block)
       end
@@ -416,6 +420,10 @@ class Chef
     # @param block [Proc] A ruby block to run. Ignored if a command is given.
     #
     def not_if(command = nil, opts = {}, &block)
+      if command.is_a?(Hash) && opts.empty?
+        opts = command
+        command = nil
+      end
       if command || block_given?
         @not_if << Conditional.not_if(self, command, opts, &block)
       end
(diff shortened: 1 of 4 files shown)


coderanger:
Something this big I would want to see go through the RFC process. Personally I think this is overkill, code this complex can use `shell_out` manually to do the same thing.

jgorak:
@coderanger Thanks for the feedback.  I'm not sure I understand the shell_out reference, though.  Would you be able to shoe me how it could be used to replace the example at the top?

lib/chef/resource.rb
```diff
@@ -416,6 +420,10 @@ def only_if(command = nil, opts = {}, &block)
     # @param block [Proc] A ruby block to run. Ignored if a command is given.
     #
     def not_if(command = nil, opts = {}, &block)
+      if command.is_a?(Hash) && opts.empty?
+        opts = command
+        command = nil
+      end
```
lamont-granquist: does `def not_if(command = nil, **opts, &block)` work here without this mangling, because this code seems brittle to me...

coderanger:
@jgorak You can do something like this:

```ruby
log "hello" do
  not_if do
    (rand < 0.5).tap do |val|
      Chef::Log.info("a thing!") unless val
    end
end
```

And similarly in the case of `shell_out(...).error?` as the condition for a command string.

lib/chef/resource.rb
jgorak: I agree 100%.  I use ruby 1.9 on my daily, and really just never realized double-splat worked a little differently than the way k:v pairs implicitly load into opts={} .

Thanks for the feedback.

jgorak:
@coderanger I see.  Thought I was missing some trick that might make it cleaner.  That works, though in my opinion it's not as simple/clean as it could potentially be.

Thanks again

coderanger:
@jgorak I agree it's not as nice, but this is clearly not something that has a huge need or others would have wanted it by now so for a minor hit to complexity of the recipe code, we can avoid adding the complexity to Chef core. The main case I would imagine is identifying between multiple guard clauses on a single resource, which is itself something we caution against as it leads to code that is very difficult to read.

lib/chef/resource.rb
lamont-granquist: yeah that was going to be my other comment is that i would have expected to see something like `opts = opts.dup.delete(:desc)` here.

lamont-granquist:
closing due to inactivity

PR #6343, opened 2017-08-22
[MSYS-646]Working on fix for minimum package versioning bug

Vasu1105:
In progress

### Description
When running 
package 'binutils >= 2.20.51.0.2-5.11' 

chef client fails with exception 
`installed package binutils-2.23.52.0.1-55.el7 is newer than candidate package binutils-2.23.52.0.1-30.el7_1.2`

Though 2.23.52.0.1-55.el7 being the latest version on the system, client run is trying to update it to version 2.23.52.0.1-30.el7_1.2 lower than currently installed 2.23.52.0.1-55.el7.
### Issues Resolved

https://getchef.zendesk.com/agent/#/tickets/15964

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -445,12 +445,26 @@ class Chef
           end
 
           unless packages.empty?
-            new_package_name = packages.first.name
-            new_package_version = packages.first.version.to_s
-            debug_msg = "#{name}: Unable to match package '#{name}' but matched #{packages.size} "
-            debug_msg << (packages.size == 1 ? "package" : "packages")
-            debug_msg << ", selected '#{new_package_name}' version '#{new_package_version}'"
-            Chef::Log.debug(debug_msg)
+            installed_package = packages.find {|p| p.installed == true }
+
+            if installed_package
+              packages.each do |pkg|
+                if pkg.version.evr > installed_package.version.evr
+                  new_package_name = pkg.name
+                  new_package_version = pkg.version.to_s
+                else
+                  new_package_name = installed_package.name
+                  new_package_version = installed_package.version
+                end
+              end
+            else
+              new_package_name = packages.first.name
+              new_package_version = packages.first.version.to_s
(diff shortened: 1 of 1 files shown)


lamont-granquist:
where's the bug report here?

lamont-granquist:
and i'd prefer to remove the yum code entirely and replace it with the dnf package approach.  and i really do not want to implement some kind of an extension to the api here which requires continuing that forwards when the dnf package provider may have solved that.

Vasu1105:
@lamont-granquist  This issue reported on zendesk here is the link https://getchef.zendesk.com/agent/#/tickets/15964 and this is card link https://chefio.atlassian.net/browse/MSYS-646

Vasu1105:
@lamont-granquist  I tried using dnf_package to install the same package which is mentioned in the issue
I am testing this on Red Hat versiugon 7.2 
I am getting error
`ec2-52-91-215-65.compute-1.amazonaws.com >>>> Caused by NoMethodError: undefined method `+' for false:FalseClass
ec2-52-91-215-65.compute-1.amazonaws.com /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.2.20/lib/chef/provider/package/dnf/python_helper.rb:44:in `dnf_command'`

When I had debug the error, came to know it tries to import dnf module at this [line](https://github.com/chef/chef/blob/master/lib/chef/provider/package/dnf/python_helper.rb#L41.)  but unable to do so since dnf package is not available on the system because of which it throws above error. So I have tried installing dnf on the system using yum install dnf but it says no package dnf available. So am I missing something over here while using dnf_package resource for RHEL 7.2

lamont-granquist:
yeah i think per discussion i just had with @jonlives is that either him or me are going to entirely replace the yum provider with a new one based on the approach in the dnf provider, which should fix this issue.

lamont-granquist:
yeah #6540 will make this go away properly in Chef-14.

PR #6351, opened 2017-08-25
dnf_provider: be more specific when we provide `package`

jaymzh:
### Description

It is really bad to have Chef all of a sudden start using a different provider
on the same OS because you happen to install DNF to play with it.

For RHEL, we know 8 will default to DNF, and 7 defaults to yum, so we specify
that.

Signed-off-by: Phil Dibowitz <phil@ipom.com>

### Issues Resolved

Unexpectedly switching package managers

### DIFF
--- a/lib/chef/provider/package/dnf.rb
+++ b/lib/chef/provider/package/dnf.rb
@@ -35,9 +35,11 @@ class Chef
         use_multipackage_api
         use_package_name_for_source
 
-        provides :package, platform_family: %w{rhel fedora amazon} do
+        provides :package, platform_family: %w{fedora amazon} do
           which("dnf") && shell_out("rpm -q dnf").stdout =~ /^dnf-[1-9]/
         end
+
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core reviewage

overall i'm okay with this.  if i noodle hard enough on it, its theoretically a breaking change, but i suspect phil is about the only person that really cares about this use case.

jaymzh:
I would argue I'm fixing a breaking change. :)

jaymzh:
(and yes I need to update the tests just getting the PR up for any commentary)

tas50:
I can't see anyone out there wanting chef to use DNF on RHEL 7, but technically they could. It would be a really odd edge case.

jaymzh:
We will want to eventually, but when we do we will call `set_priority_map` directly. We backported Rawhide DNF to RHEL7 and worked with RedHat to get those RPMs published, so we'll get there, but it certainly should switch which package manager you used just because it's installed :)

PR #6353, opened 2017-08-26
Launchd should not load launchagents as root.

mikedodge04:
The Launchd resource will load launchagents as root when the mac is
sitting on the login window. This is never a desired outcome and is hard
to recover from. This change now will now check to see if a user is
logged in before loading launchagents.

Signed-off-by: Mike Dodge <mikedodge04@gmail.com>


mikedodge04:
I screwed up my git branch hard, said f it and started over.  :'(  
#5846

PR #6355, opened 2017-08-28
speed up immutabilization

lamont-granquist:
since we only inject Mash/VividMashes do not convert


lamont-granquist:
yeah ImmutableMash is not a public-facing API.

Issue #6350, opened 2017-08-25
chef chokes on git version string on apple.


kzw:
## Description

When running git resource on macbook with git provided by Apple, chef dies with the following exception
`Malformed version number string Git-81)`

## Chef Version

12.15

## Platform Version

Mac OS 10.12

## Replication Case

Run chef recipe using git resource.

## Client Output

```
===============================================================================
    Error executing action `sync` on resource 'git[/Users/redacted/arcanist/libphutil]'
    ================================================================================

    ArgumentError
    -------------
    Malformed version number string Git-81)

    Resource Declaration:
    ---------------------
    # In /Users/redacted/repo/chef/solo/cache/cookbooks/bread-local/recipes/default.rb

     19:   git File.join(arc_root, repo) do
     20:     repository File.join(github_root, repo)
     21:     revision version
     22:     depth 1
     23:   end
     24: end

    Compiled Resource:
    ------------------
    # Declared in /Users/redacted/repo/chef/solo/cache/cookbooks/bread-local/recipes/default.rb:19:in `block in from_file'
```
## Stacktrace

Please see above

### NOTE: CHEF CLIENT BUGS ONLY

This issue tracker is for the code contained within this repo -- `chef-client`, base `knife` functionality (not
plugins), `chef-apply`, `chef-solo`, `chef-client -z`, etc.

* Requests for new or alternative functionality should be made to [feedback.chef.io](https://feedback.chef.io/forums/301644-chef-product-feedback/category/110832-chef-client)
* [Chef Server issues](https://github.com/chef/chef-server/issues/new)
* [ChefDK issues](https://github.com/chef/chef-dk/issues/new)
* Cookbook Issues (see the https://github.com/chef-cookbooks repos or search [Supermarket](https://supermarket.chef.io) or GitHub/Google)

thommay:
For reference:
```
[17:01] ~/work/chef % /usr/bin/git --version 
git version 2.13.3 (Apple Git-91)
[17:01] ~/work/chef % /usr/local/bin/git --version
git version 2.14.1
```

kzw:
Yes.  A work around would be to disable apple git.  But chef should be smarter about these version strings.  I feel that this should just a be small change in git resource code.

thommay:
We'd welcome a pull request to fix this. The code in question is here:

https://github.com/chef/chef/blob/462ce499783771dd54d3cba576e6165643a25b11/lib/chef/provider/git.rb#L106-L108

[...]

PR #6359, opened 2017-08-29
handle apple git

kzw:
### Description

This resolves issue chef issue 6350

### Issues Resolved

https://github.com/chef/chef/issues/6350

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -106,7 +106,7 @@ class Chef
       end
 
       def git_has_single_branch_option?
-        @git_has_single_branch_option ||= !git_gem_version.nil? && git_gem_version >=  Gem::Version.new("1.7.10")
+        @git_has_single_branch_option ||= !git_gem_version.nil? && git_gem_version >= Gem::Version.new("1.7.10")
       end
 
       def git_gem_version
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -110,7 +110,7 @@ class Chef
       end
 
       def git_gem_version
-        defined?(@git_gem_version) and return @git_gem_version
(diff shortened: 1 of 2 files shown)


lib/chef/provider/git.rb
```diff
@@ -103,8 +105,11 @@ def action_sync
         end
       end
 
-      def git_minor_version
-        @git_minor_version ||= Gem::Version.new( git("--version").stdout.split.last )
+      def git_gem_version
+        output = git("--version").stdout
+        match = GIT_VERSION_PATTERN.match(output)
+        raise ArgumentError, "unparsable git version number #{output}" unless match
```
lamont-granquist: it'd probably make more sense to leave the version `nil` or `false` or something, while outputting a Chef::Log.warn message and then something like:

```ruby
def git_has_single_branch_option?
  !git_gem_version.nil? && git_gem_version >= Gem::Version.new("1.7.10")
end

[...]

  clone_cmd << "--no-single-branch" if new_resource.depth && git_has_single_branch_option?
```

thommay:
Since we received two PRs at pretty much the same time that both fix the problem, I'm going to make an executive decision and say that we take #6359 , since it's slightly better factored. @Happycoil @kzw thank you both for doing this work!

lib/chef/provider/git.rb
adamleff: I'm of the same opinion as @lamont-granquist. I'd love to see us assume that the `git` available to us will do most of what we need it to do, do a best-effort parsing of the version, and just be selective about the `--no-single-branch` flag.

kzw: Good suggestions implemented.  Thanks!

thommay:
Thanks for this great PR, @kzw

PR #6360, opened 2017-08-29
[MSYS-647]array support for chocolatey_package for Artifactory repo

dheerajd-msys:
### Description

chocolatey_package installs packages in array for Artifactory server as source repo.

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -538,7 +538,11 @@ class Chef
       # @yield [package_name, new_version, current_version, candidate_version] Description of block
       def each_package
         package_name_array.each_with_index do |package_name, i|
-          candidate_version = candidate_version_array[i]
+          if new_version_array[i].nil?
+            candidate_version = candidate_version_array[i]
+          else
+            candidate_version = new_version_array[i]
+          end
           current_version = current_version_array[i]
           new_version = new_version_array[i]
           yield package_name, new_version, current_version, candidate_version
--- a/spec/unit/provider/package_spec.rb
+++ b/spec/unit/provider/package_spec.rb
@@ -711,11 +711,11 @@ describe "Chef::Provider::Package - Multi" do
       expect { provider.run_action(:install) }.not_to raise_error
     end
 
-    it "raises an exception if an explicit version is asked for, an old version is installed, but no candidate" do
+    it "does not raise an exception if an explicit version is asked for, an old version is installed, but no candidate" do
       new_resource.version ["1.0", "6.2"]
       current_resource.version(["1.0", "6.1"])
       provider.candidate_version = ["1.0", nil]
-      expect { provider.run_action(:install) }.to raise_error(Chef::Exceptions::Package)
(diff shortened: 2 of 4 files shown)


btm:
@dheerajd-msys can you add a unit test and look into the build failures please?

btm:
I'm concerned about why we're making changes to the OpenBSD spec. We'll need @lamont-granquist to look at this given the change to `Chef::Provider`.

lib/chef/provider/package.rb
```diff
@@ -538,7 +538,11 @@ def forced_packages_missing_candidates
       # @yield [package_name, new_version, current_version, candidate_version] Description of block
       def each_package
         package_name_array.each_with_index do |package_name, i|
-          candidate_version = candidate_version_array[i]
+          if new_version_array[i].nil?
+            candidate_version = candidate_version_array[i]
+          else
+            candidate_version = new_version_array[i]
+          end
```
lamont-granquist: i don't think you can do this, you can't overwrite the candidate_version (what is available) with what the user desires since those are often very different concerns.

spec/unit/provider/package_spec.rb
```diff
@@ -711,11 +711,11 @@ class MyPackageProvider < Chef::Provider::Package
       expect { provider.run_action(:install) }.not_to raise_error
     end
 
-    it "raises an exception if an explicit version is asked for, an old version is installed, but no candidate" do
+    it "does not raise an exception if an explicit version is asked for, an old version is installed, but no candidate" do
       new_resource.version ["1.0", "6.2"]
       current_resource.version(["1.0", "6.1"])
       provider.candidate_version = ["1.0", nil]
-      expect { provider.run_action(:install) }.to raise_error(Chef::Exceptions::Package)
```
lamont-granquist: this test is no longer correct.

if we have "6.1" installed, and we want "6.2" then we need to perform an action, but the repo lacks any version at all, so we must blow up -- if we attempt to install then we can't install 6.2.   candidate_version of nil is perfectly correct in this case, and you can't overwrite it with "6.2"

spec/unit/provider/package_spec.rb
```diff
@@ -725,11 +725,11 @@ class MyPackageProvider < Chef::Provider::Package
       expect { provider.run_action(:install) }.not_to raise_error
     end
 
-    it "raise an exception if an explicit version is asked for, and is not installed, and no candidate" do
+    it "does not raise an exception if an explicit version is asked for, and is not installed, and no candidate" do
       new_resource.version ["1.0", "6.2"]
       current_resource.version(["1.0", nil])
       provider.candidate_version = ["1.0", nil]
-      expect { provider.run_action(:install) }.to raise_error(Chef::Exceptions::Package)
```
lamont-granquist: same thing here.

lamont-granquist:
the generic tests are actually more directly concerning.

dheerajd-msys:
Creating another PR since we do require these changes.

lamont-granquist:
closing this one out then...

PR #6362, opened 2017-08-29
immutablize properly as we deep merge

lamont-granquist:
removes a lot of excess dup'ing that was going on in deep merge

this forks our deep merge implementation since people were using deep merge as an external API, internally we no longer use the old deep merge class.

the attribute version of deep merging is about to mutate into unrecognizability though, and is no longer an external api.

### DIFF
--- a/lib/chef/mash.rb
+++ b/lib/chef/mash.rb
@@ -105,6 +105,12 @@ class Mash < Hash
     regular_writer(convert_key(key), convert_value(value))
   end
 
+  # internal API for use by Chef's deep merge cache
+  # @api private
+  def internal_set(key, value)
+    regular_writer(key, convert_value(value))
+  end
+
   # @param other_hash<Hash>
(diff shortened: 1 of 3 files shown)


jaymzh:
@adamleff yup. Particularly on attribute-heavy users like us (Facebook).

adamleff:
@jaymzh I think I grossly underestimated the amount of attributes you all use 🙂 I hope this helps you out!

lamont-granquist:
yeah it has a pile of speed up.  its still not actually sufficient, so its likely that i'm going to have to redo how deep merging works and how the merged cache works completely.

the fix in https://github.com/chef/chef/pull/5360 did a whole lot of extra work, but i think now we're down to the fact that properly immutablizing through arrays is the rest of the perf hit that facebook is seeing before and after that PR.  so i think this is as fast as we can make attributes work via 'patching' and the next step is throwing out the top-level-key merge cache and doing it at every level on-demand.

PR #6364, opened 2017-08-29
Link to the knife docs when the knife config file is missing

tas50:
This is a pretty common first error message to experience with Chef. Let's point the users to a helpful location so they can quickly get up and running.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -52,6 +52,12 @@ module ChefConfig
       configuration.inspect
     end
 
+    # given a *nix style config path return the platform specific path
+    # to that same config file
+    # @example client.pem path on Windows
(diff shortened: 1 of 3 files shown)


lamont-granquist:
ohai_hint resource snuck in here

tas50:
let me nuke that out

thommay:
If the user is using `config.rb` talking about `knife.rb` will be pretty confusing, but I like the intent.

tas50:
I'd agree, but we never actually documented config.rb in the workstation and knife docs. I'm not sure how many people are using that.

thommay:
I don't entirely understand why it's in the ChefDK docs, but the Configuration topic there is all about config.rb. So we are telling people about it.

tas50:
Updated to not mention knife.rb specifically, but still send you to the docs

PR #6372, opened 2017-08-31
Support an array of keys for apt_repository

### LINKED ISSUE #6369: Support multiple keys for apt_repository to match yum_repository
## Description

For `apt_repository` resource the key property does not accept an array of keys. The `yum_repository` resource does. 

https://github.com/chef/chef/blob/4f2bac9506f9e93dd33e0a94d61c016f6aeacbb1/lib/chef/resource/yum_repository.rb#L39
https://github.com/chef/chef/blob/4f2bac9506f9e93dd33e0a94d61c016f6aeacbb1/lib/chef/resource/apt_repository.rb#L36

For deb packages which require two keys to be installed to authenticate, it's not currently supported to install both in a single `apt_repository` resource. 

Current workaround: 
https://github.com/sous-chefs/rundeck/blob/247c4282b2be2c4fb0efcd814dde907a32a4b8ed/recipes/server_install.rb#L54-L73

Related: https://github.com/sous-chefs/rundeck/pull/178

## Chef Version

13.3.42

## Platform Version

Ubuntu 16.04, any debian based distro

## Replication Case

Install rundeck package from [bintray apt repo](https://dl.bintray.com/rundeck/rundeck-deb)
Need to install *both* bintray key `379CE192D401AB61` from keyserver.ubuntu.com and rundeck build key from `http://rundeck.org/keys/BUILD-GPG-KEY-Rundeck.org.key` to install the package without allowing unauthenticated packages

## Client Output

apt-get install execute in `apt_package` fails with
`E: There were unauthenticated packages and -y was used without --allow-unauthenticated`

gsreynolds:
### Description

Amends `apt_repository` key property to accept an array of keys to install, in the case of authenticating packages from repos signed by multiple keys.

### Issues Resolved

Fix #6369

### DIFF
--- a/lib/chef/resource/apt_repository.rb
+++ b/lib/chef/resource/apt_repository.rb
@@ -33,7 +33,7 @@ class Chef
       # whether or not to add the repository as a source repo, too
       property :deb_src, [TrueClass, FalseClass], default: false
       property :keyserver, [String, nil, false], default: "keyserver.ubuntu.com", nillable: true, coerce: proc { |x| x ? x : nil }
-      property :key, [String, Array, nil, false], default: [], coerce: proc { |x| x ? [x].flatten : nil }
+      property :key, [String, Array, nil, false], default: [], coerce: proc { |x| x ? Array(x) : nil }
       property :key_proxy, [String, nil, false], default: nil, nillable: true, coerce: proc { |x| x ? x : nil }
 
       property :cookbook, [String, nil, false], default: nil, desired_state: false, nillable: true, coerce: proc { |x| x ? x : nil }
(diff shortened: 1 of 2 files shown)


lib/chef/resource/apt_repository.rb
```diff
@@ -33,7 +33,7 @@ class AptRepository < Chef::Resource
       # whether or not to add the repository as a source repo, too
       property :deb_src, [TrueClass, FalseClass], default: false
       property :keyserver, [String, nil, false], default: "keyserver.ubuntu.com", nillable: true, coerce: proc { |x| x ? x : nil }
-      property :key, [String, nil, false], default: nil, nillable: true, coerce: proc { |x| x ? x : nil }
+      property :key, [String, Array, nil, false], default: [], coerce: proc { |x| x ? [x].flatten : nil }
```
thommay: ` x ? [x].flatten : nil`
is better written as
` x ? Array(x) : nil`

gsreynolds: Thanks! I just referenced what was used elsewhere, powershell, dnf and chocolatey package providers used `[x].flatten`. No problem to change if `Array(x)` is preferred though!

gsreynolds:
Thanks!

thommay:
Thanks for this, @gsreynolds

PR #6376, opened 2017-09-01
bump rb-readline

lamont-granquist:


lamont-granquist:
back to halite + chef-zero being busted

Issue #5588, opened 2016-11-28
dsc_script idempotence not working?


rayterrill:
## Description

I cannot get dsc_script to correctly work with idempotence. Always reporting resources are changing, even though they are not. If I code the same functionality in a powershell_script resources with a not_if guard, idempotence works correctly.

## Chef Version

12.16.42

## Platform Version

Server2012R2

## Replication Case

```
#resource definition
dsc_script 'NetFramework35Core_dscScript' do
  code <<-EOH
    WindowsFeature NetFramework35Core
    {
      Name = "Net-Framework-Core"
      Ensure = "Present"
      Source = 'C:\\Windows\\WinSxS'
    }
  EOH
end
```

run_status.updated_resources shows the above as always updated.

## Client Output

```
myhost.mydomain.com [2016-11-28T12:49:18-08:00] INFO: Processing dsc
_script[NetFramework35Core_dscScript] action run (Troubleshooting::default line
15)
myhost.mydomain.com [2016-11-28T12:49:21-08:00] WARN: Could not pars
e LCM output: Could not parse:
myhost.mydomain.com WARNING: Whatif is deprecated in this cmdlet. Pl
ease remove Whatif and try
myhost.mydomain.com again.
myhost.mydomain.com
myhost.mydomain.com [2016-11-28T12:49:26-08:00] INFO: DSC resource c
onfiguration completed successfully
myhost.mydomain.com [2016-11-28T12:49:26-08:00] INFO: Chef Run compl
ete in 13.517764 seconds
myhost.mydomain.com [2016-11-28T12:49:26-08:00] INFO: Running report
 handlers
myhost.mydomain.com https://myreportingnode.mydomain.com:444/Chef/chefCh
eckin.php?node=myhost.mydomain.com&status=0&runlist=recipe[POPBase],
recipe[Troubleshooting]&totalresources=11&updatedresources=3&elapsedtime=13.5177
64&updatedresourcestext=chef_handler[POPLogging::LogChefRuns],chef_handler[POPLo
gging::SendEmail],dsc_script[NetFramework35Core_dscScript]
myhost.mydomain.com [2016-11-28T12:49:26-08:00] INFO: Report handler
s complete
```

## Stacktrace

Please include the stacktrace.out output or link to a gist of it, if there is one.

### NOTE: CHEF CLIENT BUGS ONLY

This issue tracker is for the code contained within this repo -- `chef-client`, base `knife` functionality (not
[...]

PR #6383, opened 2017-09-07
MSYS-656 - fixed dsc_script for WMF5

piyushawasthi:
Fixes https://github.com/chef/chef/issues/5588

### Description

[fixed dsc_script for WMF5]

### Issues Resolved

[dsc_script idempotence not working?]

### DIFF
--- a/lib/chef/util/dsc/local_configuration_manager.rb
+++ b/lib/chef/util/dsc/local_configuration_manager.rb
@@ -47,15 +47,13 @@ class Chef::Util::DSC
 
     def run_configuration_cmdlet(configuration_document, apply_configuration, shellout_flags)
       Chef::Log.debug("DSC: Calling DSC Local Config Manager to #{apply_configuration ? "set" : "test"} configuration document.")
-      test_parameters = ! apply_configuration ? "-whatif; if (! $?) { exit 1 }" : ""
 
       start_operation_timing
-      command_code = lcm_command_code(@configuration_path, test_parameters)
       status = nil
 
       begin
         save_configuration_document(configuration_document)
-        cmdlet = ::Chef::Util::Powershell::Cmdlet.new(@node, "#{command_code}")
+        cmdlet = ::Chef::Util::Powershell::Cmdlet.new(@node, lcm_command(apply_configuration))
         if apply_configuration
           status = cmdlet.run!({}, shellout_flags)
         else
@@ -72,19 +70,22 @@ class Chef::Util::DSC
       status
     end
 
-    def lcm_command_code(configuration_path, test_parameters)
-      if !test_parameters.strip.empty? && is_ps_version_gte_5?
-        <<-EOH
-$ProgressPreference = 'SilentlyContinue';Test-DscConfiguration -path #{@configuration_path}
-EOH
+    def lcm_command(apply_configuration)
+      common_command_prefix = "$ProgressPreference = 'SilentlyContinue';"
+      ps4_base_command = "#{common_command_prefix} Start-DscConfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force"
+
+      if apply_configuration
+        ps4_base_command
       else
-        <<-EOH
-$ProgressPreference = 'SilentlyContinue';start-dscconfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force #{test_parameters}
-EOH
+        if ps_version_gte_5?
+          "#{common_command_prefix} Test-DscConfiguration -path #{@configuration_path}"
+        else
+          ps4_base_command + " -whatif; if (! $?) { exit 1 }"
+        end
       end
     end
 
-    def is_ps_version_gte_5?
+    def ps_version_gte_5?
       Chef::Platform.supported_powershell_version?(@node, 5)
     end
 
--- a/lib/chef/util/dsc/local_configuration_manager.rb
+++ b/lib/chef/util/dsc/local_configuration_manager.rb
@@ -47,10 +47,10 @@ class Chef::Util::DSC
 
     def run_configuration_cmdlet(configuration_document, apply_configuration, shellout_flags)
       Chef::Log.debug("DSC: Calling DSC Local Config Manager to #{apply_configuration ? "set" : "test"} configuration document.")
-      test_only_parameters = ! apply_configuration ? "-whatif; if (! $?) { exit 1 }" : ""
+      test_parameters = ! apply_configuration ? "-whatif; if (! $?) { exit 1 }" : ""
 
       start_operation_timing
-      command_code = lcm_command_code(@configuration_path, test_only_parameters)
+      command_code = lcm_command_code(@configuration_path, test_parameters)
       status = nil
 
       begin
@@ -72,10 +72,20 @@ class Chef::Util::DSC
       status
     end
 
-    def lcm_command_code(configuration_path, test_only_parameters)
-      <<-EOH
-$ProgressPreference = 'SilentlyContinue';start-dscconfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force #{test_only_parameters}
+    def lcm_command_code(configuration_path, test_parameters)
+      if !test_parameters.strip.empty? && is_ps_version_gte_5?
+        <<-EOH
+$ProgressPreference = 'SilentlyContinue';Test-DscConfiguration -path #{@configuration_path}
 EOH
+      else
+        <<-EOH
+$ProgressPreference = 'SilentlyContinue';start-dscconfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force #{test_parameters}
+EOH
+      end
+    end
+
+    def is_ps_version_gte_5?
+      Chef::Platform.supported_powershell_version?(@node, 5)
     end
 
     def log_what_if_exception(what_if_exception_output)
--- a/lib/chef/util/dsc/local_configuration_manager.rb
+++ b/lib/chef/util/dsc/local_configuration_manager.rb
@@ -73,7 +73,6 @@ class Chef::Util::DSC
     def lcm_command(apply_configuration)
       common_command_prefix = "$ProgressPreference = 'SilentlyContinue';"
       ps4_base_command = "#{common_command_prefix} Start-DscConfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force"
-
       if apply_configuration
         ps4_base_command
       else
--- a/spec/unit/util/dsc/local_configuration_manager_spec.rb
+++ b/spec/unit/util/dsc/local_configuration_manager_spec.rb
@@ -63,11 +63,63 @@ EOH
         let(:lcm_standard_error) { nil }
         let(:lcm_cmdlet_success) { true }
 
-        it "should successfully return resource information for normally formatted output when cmdlet the cmdlet succeeds" do
+        it "successfully returns resource information for normally formatted output when cmdlet the cmdlet succeeds" do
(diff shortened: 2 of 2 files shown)


lib/chef/util/dsc/local_configuration_manager.rb
```diff
@@ -78,6 +78,18 @@ def lcm_command_code(configuration_path, test_only_parameters)
 EOH
     end
 
+    def test_lcm_command_code
+      ps_version = Mixlib::ShellOut.new("powershell $PSVersionTable.PSVersion.Major").run_command
```
stuartpreston: *or* use the query helper that is already there: ```Chef::Platform.supported_powershell_version?(node, 5)```

lib/chef/util/dsc/local_configuration_manager.rb
```diff
@@ -78,6 +78,18 @@ def lcm_command_code(configuration_path, test_only_parameters)
...
+      ps_version.error!
+      if ps_version.stdout.to_i > 4
+        <<-EOH
+$ProgressPreference = 'SilentlyContinue';Test-DscConfiguration -path #{@configuration_path}
```
btm: Does this mean we still run `Start-DscConfiguration` and then `Test-DscConfiguration` to test? Wouldn't `Start-DscConfiguration` without `-whatif` actually make a change? I think `test_only_parameters` need to something more like `test_command`.

lib/chef/util/dsc/local_configuration_manager.rb
piyushawasthi: Thanks @stuartpreston,

lib/chef/util/dsc/local_configuration_manager.rb
piyushawasthi: First we run `Test-DscConfiguration` then `Start-DscConfiguration`.

lib/chef/util/dsc/local_configuration_manager.rb
```diff
@@ -72,10 +72,24 @@ def run_configuration_cmdlet(configuration_document, apply_configuration, shello
...
+      else
+        <<-EOH
+$ProgressPreference = 'SilentlyContinue';start-dscconfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force #{test_command}
+EOH
+      end
+    end
+
+    def test_lcm_command_code
+      if Chef::Platform.supported_powershell_version?(@node, 5)
+        "test"
```
NimishaS: Can we return something more specific to Powershell version 5 e.g. `test_powershell_5`? Returning `test` looks vague here.

btm:
@piyushawasthi I'd appreciate it if you would refactor this a little more.

`test_supported_powershell` should return just a boolean with one purpose, something like `powershell_version_gte_5?` or `powershell_version_greater_than_5?` The string for powershell 4 that is returned by `test_supported_powershell` should just be part of `lcm_command_code` which should be told if we're testing or not, then look up the version, and return the right string to run. (I Think. but that's why this needs to be simpler).

piyushawasthi:
@btm - I have updated code as per comment, Please review.

btm:
@piyushawasthi take a look at this commit I pushed, this is more what I had in mind: https://github.com/chef/chef/commit/ca92458609a6a475eb69f6b94aac3dcd71e6ecbe

The goal here is to take everything related to the command we run and have it be in the same method. having the `"-whatif; if (! $?) { exit 1 }"` string in `run_configuration_cmdlet` and the rest of the strings in `lcm_command_code` made it hard to follow.

Can you add a unit test or two around the new PS5 behavior too? If you cherry-pick my commit you can just stub `ps_version_gte_5?` and write unit tests around `lcm_command` as it is now self-contained. Will probably have to update other tests due to the refactor though.

lib/chef/util/dsc/local_configuration_manager.rb
```diff
@@ -72,10 +70,22 @@ def run_configuration_cmdlet(configuration_document, apply_configuration, shello
...
-EOH
+    def lcm_command(apply_configuration)
+      @common_command_prefix = "$ProgressPreference = 'SilentlyContinue';"
+      ps4_base_command = "#{@common_command_prefix} Start-DscConfiguration -path #{@configuration_path} -wait -erroraction 'stop' -force"
+      apply_configuration ? ps4_base_command : test_lcm_configuration(ps4_base_command)
+    end
+
+    def test_lcm_configuration(ps4_base_command)
+      if ps_version_gte_5?
+        "#{@common_command_prefix} Test-DscConfiguration -path #{@configuration_path}"
```
NimishaS: We are using an instance variable `@common_command_prefix` here which is required to be set prior to calling this method. If we need to use this `test_lcm_configuration` method for any other command in future then we'll have to make sure to define `@common_command_prefix` otherwise the function will throw error for PS 5.
It would be good to pass `common_command_prefix` as an argument.

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -63,7 +63,35 @@
         let(:lcm_standard_error) { nil }
         let(:lcm_cmdlet_success) { true }
 
-        it "should successfully return resource information for normally formatted output when cmdlet the cmdlet succeeds" do
+        it "successfully return resource information for normally formatted output when cmdlet the cmdlet succeeds" do
```
NimishaS: `successfully *returns* resource information for normally formatted output when the cmdlet succeeds.`

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -63,7 +63,35 @@
...
+          test_configuration_result = lcm.test_configuration("config", {})
+          expect(test_configuration_result.class).to be(Array)
+          expect(test_configuration_result.length).to be > 0
+          expect(Chef::Log).not_to receive(:warn)
+        end
+      end
+
+      context "returns successfully for PowerShell version 5" do
```
NimishaS: Nitpicking: It's not clear what it returns successfully for PowerShell version 5. It would be a good idea to use:
```
context "when running on PowerShell version 5" do
```

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -63,7 +63,35 @@
...
+        let(:lcm_standard_output) { normal_lcm_output }
+        let(:lcm_standard_error) { nil }
+        let(:lcm_cmdlet_success) { true }
+
+        it "successfully return resource information for normally formatted output when cmdlet the cmdlet succeeds" do
```
NimishaS: successfully `returns`.

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -63,7 +63,35 @@
...
+          allow(lcm).to receive(:ps_version_gte_5?).and_return(true)
+          test_configuration_result = lcm.test_configuration("config", {})
+          expect(test_configuration_result.class).to be(Array)
+          expect(test_configuration_result.length).to be > 0
+          expect(Chef::Log).not_to receive(:warn)
+        end
+      end
+
+      context "returns successfully for PowerShell version less than 5" do
```
NimishaS: Nitpicking: It's not clear what it returns successfully for PowerShell version less than 5. It would be a good idea to use:
```
context "when running on PowerShell version less than 5" do
```

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -63,7 +63,35 @@
...
+        let(:lcm_standard_output) { normal_lcm_output }
+        let(:lcm_standard_error) { nil }
+        let(:lcm_cmdlet_success) { true }
+
+        it "successfully return resource information for normally formatted output when cmdlet the cmdlet succeeds" do
```
NimishaS: successfully `returns`. Please fix this in all the specs below from where `should` has been removed.

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -114,23 +142,23 @@
...
         end
       end
     end
 
-    it "should identify a correctly formatted error message as a resource import failure" do
+    it "identify a correctly formatted error message as a resource import failure" do
       expect(lcm.send(:dsc_module_import_failure?, dsc_resource_import_failure_output)).to be(true)
     end
 
-    it "should not identify an incorrectly formatted error message as a resource import failure" do
```
NimishaS: `does` not identify an incorrectly formatted error message as a resource import failure.

Applicable to the specs below as well.

lib/chef/util/dsc/local_configuration_manager.rb
btm: ` test_lcm_configuration` is only called from `lcm_command`. I'd combine it back into `lcm_command` similar to how I had https://github.com/chef/chef/commit/ca92458609a6a475eb69f6b94aac3dcd71e6ecbe. That way you don't have to set `@common_command_prefix` or pass `ps4_base_command` to a method that only needs it half the time.

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -114,23 +142,23 @@
         let(:lcm_standard_error) { "Abort, Retry, Fail?" }
         let(:lcm_cmdlet_success) { false }
 
-        it "should log a warning" do
+        it "log a warning" do
```
btm: *logs a warning

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -94,14 +122,14 @@
...
-        it "should log a warning if the message is formatted as expected when a resource import failure occurs" do
+        it "log a warning if the message is formatted as expected when a resource import failure occurs" do
           expect(Chef::Log).to receive(:warn).at_least(:once)
           expect(lcm).to receive(:dsc_module_import_failure?).and_call_original
           test_configuration_result = nil
           expect { test_configuration_result = lcm.test_configuration("config", {}) }.not_to raise_error
         end
 
-        it "should return a (possibly empty) array of ResourceInfo instances" do
+        it "return a (possibly empty) array of ResourceInfo instances" do
```
btm: *returns

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -94,14 +122,14 @@
         let(:lcm_standard_error) { dsc_resource_import_failure_output }
         let(:lcm_cmdlet_success) { false }
 
-        it "should log a warning if the message is formatted as expected when a resource import failure occurs" do
+        it "log a warning if the message is formatted as expected when a resource import failure occurs" do
```
btm: logs

spec/unit/util/dsc/local_configuration_manager_spec.rb
```diff
@@ -80,7 +108,7 @@
           expect(lcm.send(:whatif_not_supported?, no_whatif_lcm_output)).to be_truthy
         end
 
-        it "should should return a (possibly empty) array of ResourceInfo instances" do
+        it "return a (possibly empty) array of ResourceInfo instances" do
```
btm: returns

spec/unit/util/dsc/local_configuration_manager_spec.rb
btm: it seems like if you end up with a combined `lcm_command`, you would want a test somewhere that ensures you get `-whatif` on ps4 and `test-dscconfiguration` on ps5.

PR #6387, opened 2017-09-08
Cleanup the current changelog

tas50:
Remove release note / changelog fixes, improve a few descriptions. Move
bugs into the bug section


tduffield:
Don't forget you can use the `Meta: Exclude From Changelog` label to prevent PRs from being added to the Changelog :)

PR #6394, opened 2017-09-12
Windows: Added :none frequency to windows_task resource

### LINKED ISSUE #6200: `windows_task` should have a frequency option of `none`
## Description

Currently you have to specify a frequency for a scheduled task.  There are occasionally a need to have a scheduled task with 'no frequency'.  ( I know...I sound like a crazy person with this request).
But i have come across this need a few times where i need this type of functionality...and the chef run just needed to call `run` on the scheduled task.  But i could not use `windows_task` to create the task.

## Chef Version
13

## Platform Version
Windows 2012 R2

NAshwini:
### Description

Updated code to add `none` frequency option to windows task resource.

### Issues Resolved

Fixes #6200

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -89,11 +89,17 @@ class Chef
         basic_validation
         options = {}
         options["F"] = "" if new_resource.force || task_need_update?
-        options["SC"] = schedule
+        if schedule == :none
+          options["SC"] = :once
+          options["ST"] = "00:00"
+          options["SD"] = convert_user_date_to_system_date "12/12/2012"
+        else
+          options["SC"] = schedule
+          options["ST"] = new_resource.start_time unless new_resource.start_time.nil?
+          options["SD"] = convert_user_date_to_system_date new_resource.start_day unless new_resource.start_day.nil?
+        end
         options["MO"] = new_resource.frequency_modifier if frequency_modifier_allowed
         options["I"]  = new_resource.idle_time unless new_resource.idle_time.nil?
-        options["SD"] = convert_user_date_to_system_date new_resource.start_day unless new_resource.start_day.nil?
-        options["ST"] = new_resource.start_time unless new_resource.start_time.nil?
         options["TR"] = new_resource.command
         options["RU"] = new_resource.user
         options["RP"] = new_resource.password if use_password?
@@ -277,6 +283,7 @@ class Chef
           :daily => "Triggers/CalendarTrigger/RandomDelay",
           :weekly => "Triggers/CalendarTrigger/RandomDelay",
           :monthly => "Triggers/CalendarTrigger/RandomDelay",
+          :none => "Triggers",
         }
 
         xml_element_mapping = {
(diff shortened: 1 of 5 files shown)


btm:
@NAshwini seems fine. I don't know enough about the library to know about all the specific task logic choices here. Can you look at the failing tests?

lib/chef/provider/windows_task.rb
```diff
@@ -89,11 +89,17 @@ def action_create
         basic_validation
         options = {}
         options["F"] = "" if new_resource.force || task_need_update?
-        options["SC"] = schedule
+        if schedule == :none
+          options["SC"] = :once
+          options["ST"] = "00:00"
+          options["SD"] = convert_user_date_to_system_date "12/12/2012"
```
stevendanna: In the original issue, the reporter indicated that we could construct XML hat doesn't require this "start time in the past" trick:

https://github.com/chef/chef/issues/6200#issuecomment-324386409

Is there a reason we can't do that here?  I see that it would require a substantial change to this class since we rely on option flags for nearly everything at the moment.  I understand if that is a bit more than we want to do at the moment, but it  might be worth a comment.

lib/chef/provider/windows_task.rb
```diff
@@ -297,6 +304,12 @@ def update_task_xml(options = [])
 
         doc = REXML::Document.new(xml_cmd.stdout)
 
+        if new_resource.frequency == :none
+          doc.root.elements.delete(xml_element_mapping["random_delay"])
+          cwd_element = REXML::Element.new(xml_element_mapping["random_delay"])
```
stevendanna: Variable naming here is a bit unclear.  Why is this called `cwd_element`?

Also, can you explain what this code does, I feel like I'm missing some important detail.  It looks like we are deleting an element and then adding the same element right back.

tas50:
@stevendanna Another possibility is using https://github.com/djberg96/win32-taskscheduler which seems like a promising way to abstract away the complexity of things.

lib/chef/provider/windows_task.rb
NAshwini: Thank you @stevendanna for your review comments.
 As per original issue comments I have changed the XML file in `def update_task_xml`. But here the code flow is such that we are creating a task first using `schtasks` command and then updating the XML template if required, for value passed such as `execution time`, `random_delay`, etc.  So here we created a scheduled task with frequency `:once` for an outdated `start date`, so that it doesn't execute. Then we update it's XML for frequency `none` in `def update_task_xml`. 
**Note**: We need to do this because `SC` is a mandatory parameter in `schtasks` command.

lib/chef/provider/windows_task.rb
NAshwini: Thanks, this variable name should be `trigger_element` . I have made the changes.
As per approach explained in above comments,
This code is to delete value of start_time and startdate under `<TimeTrigger>`  from `<Triggers>` as below:
```
<Triggers>
    <TimeTrigger>
      <StartBoundary>2017-08-28T11:00:00</StartBoundary>
      <Enabled>true</Enabled>
    </TimeTrigger>
  </Triggers>
```
And then add only `<Triggers>` for frequncy `:none` as below:
` <Triggers />`

NAshwini:
@btm , I have fixed the travis and appveyor related failure of windows_task specs.
But now getting travis failure which is not related to this PR.

tas50:
Documented this here: https://github.com/chef/chef-web-docs/pull/1042

PR #6404, opened 2017-09-14
update ruby version

thommay:


omnibus/Gemfile.lock
```diff
@@ -79,20 +80,21 @@ GEM
...
       celluloid (>= 0.16.0)
       nio4r (>= 1.1.0)
-    chef-config (13.2.20)
+    chef-config (13.4.19)
       addressable
       fuzzyurl
       mixlib-config (~> 2.0)
       mixlib-shellout (~> 2.0)
     chef-sugar (3.5.0)
+    citrus (3.0.2)
```
btm: Is this inclusion going to require any changes related to license_scout?

https://github.com/mjackson/citrus#license

omnibus/Gemfile.lock
```diff
@@ -227,6 +228,8 @@ GEM
     thor (0.19.1)
     timers (4.0.4)
       hitimes
+    toml-rb (1.0.0)
```
btm: This is new as well.

https://github.com/emancu/toml-rb#license

omnibus/Gemfile.lock
thommay: we don't ship these gems, just use them to run omnibus, so no, it shouldn't

thommay: ironically it's license scout that brings those in.

lamont-granquist:
does 2.4.2 address that proxy issue that was fixed in master but not released yet?

PR #6408, opened 2017-09-15
Quiet the output of the zypper refresh and add force

tas50:
This is probably the sane set of options for an automated refresh approach.


tas50:
Test failure is unrelated

PR #6409, opened 2017-09-15
Replace which apt-get check with simple debian check in apt resources

tas50:
apt-get is a link to yast2 in suse so these resources fire there when someone wouldn't expect them to. Looking back we probably should have skipped the whole platform based no-op thing, but that ship sailed. This makes it behave the way a user would expect it to. What happens if you have apt compiled on your gentoo box? Well it's going to break. We'll just have to live with fixing it for 99.99% of users instead of the current 95%. Also: This is faster.

### DIFF
--- a/lib/chef/provider/apt_preference.rb
+++ b/lib/chef/provider/apt_preference.rb
@@ -19,17 +19,12 @@
 require "chef/provider"
 require "chef/dsl/declare_resource"
 require "chef/provider/noop"
-require "chef/mixin/which"
 require "chef/log"
 
 class Chef
   class Provider
     class AptPreference < Chef::Provider
-      extend Chef::Mixin::Which
-
-      provides :apt_preference do
-        which("apt-get")
-      end
+      provides :apt_preference, os: "linux", platform_family: [ "debian" ]
 
       APT_PREFERENCE_DIR = "/etc/apt/preferences.d".freeze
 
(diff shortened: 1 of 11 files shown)


lib/chef/provider/apt_preference.rb
```diff
@@ -19,17 +19,12 @@
...
 
 class Chef
   class Provider
     class AptPreference < Chef::Provider
-      extend Chef::Mixin::Which
-
-      provides :apt_preference do
-        which("apt-get")
-      end
+      provides :apt_preference, os: "linux", platform_family: [ "debian" ]
```
lamont-granquist: you can drop the array.

tas50: Done

lib/chef/resource/apt_preference.rb
```diff
@@ -22,7 +22,7 @@ class Chef
   class Resource
     class AptPreference < Chef::Resource
       resource_name :apt_preference
-      provides :apt_preference
+      provides :apt_preference, os: "linux", platform_family: "debian"
```
tas50: Should I leave this open ended?

lamont-granquist: resource + provider should probably match

tas50:
DCO was signed off and the test failure is a docker container issue that I'll fix in another PR

lamont-granquist:
@tas50 you broke appveyor

lamont-granquist:
yeah i had a functional test around nameless resources that utilized apt_update because it supported namelessness.  probably need to add a resource to that test which is nameless and use that instead.

PR #6414, opened 2017-09-15
Update dokken integration testing: dokken images, updated cookbook, debian 9

tas50:
Remove EOL CentOS 5 / openSUSE 13.2
Add Debian 9 / openSUSE Leap


kitchen-tests/Berksfile
```diff
@@ -1,12 +1,8 @@
 source "https://supermarket.chef.io"
 
-#cookbook "webapp", path: "cookbooks/webapp"
+cookbook "webapp", path: "cookbooks/webapp"
```
lamont-granquist: it'd probably be more useful to focus on getting the learnchef tutorial cookbooks working again (below) and i think those are fairly "webapp"-ey and would provide similar coverage and validate the tutorials again.

tas50: Yep I was going to get those working as well. This is what happens when I go in to fix a simple failure and get sucked in

tas50:
@lamont-granquist If you're still Ok with the changes things are passing now. The CentOS issue was actually fixed in an updated openssh cookbook that handles docker, but all the other stuff is still worth having and gets us a tiny bit closer to turning the webapp and learn chef testing back on.

lamont-granquist:
you need to rebase again, again

kitchen-tests/cookbooks/base/recipes/packages.rb
```diff
@@ -7,13 +7,7 @@
 
 # this is just a list of package that exist on every O/S we test, and often aren't installed by default.  you don't
 # have to get too clever here, you can delete packages if they don't exist everywhere we test.
-pkgs = %w{lsof tcpdump strace zsh dmidecode ltrace bc curl wget telnet subversion git traceroute htop tmux }
-
-# this deliberately calls the multipackage API N times in order to do one package installation in order to exercise the
-# multipackage cookbook.
-pkgs.each do |pkg|
-  multipackage pkgs
-end
```
lamont-granquist: you need to back this out -- its deliberately being mildly stupid just to exercise the multipackage cookbook to ensure that does not break.  you optimized away the use of that cookbook.

kitchen-tests/Berksfile
lamont-granquist: So to be a bit more clear -- the direction I was going was to replace the webapp cookbook with those learnchef tutorial cookbooks since those exercise mysql+httpd cookbooks and they're maintained.  I just never got around to deleting the webapp cookbook, which was the plan...

PR #6417, opened 2017-09-17
Open apt resources up to prevent breaking change

tas50:
If we want these to throw a no ops vs. a hard failure we need them to be
wide open. Since a lot of people (including all of chef-cookbooks) have
coded as if these will gracefully skip this would have been a breaking
change.


tas50:
This should fix our appveyor failures as well

tas50:
On CI system down!

PR #6421, opened 2017-09-18
Remove the webapp cookbook from integration tests

tas50:
This isn't coming back. We'll replace it with the learnchef stuff in another PR.


tas50:
Not sure what the appveyor failure is about, but it can't possibly be related to these changes

lamont-granquist:
looked like a sunspot / ran too long and got killed or something

PR #6428, opened 2017-09-21
Fixes to package upgrade behaviour

### LINKED ISSUE #6352: DNF provider doesn't handle `:upgrade` as expected
## Description

DNF provider will downgrade if we have a newer package installed than the repo.

i.e.:

* Install foo-2.3-2
* yum/dnf repo's latest version is foo-2.1-1
* `package 'foo' { action :upgrade }` will downgrade you to 2.1

In the yum provider this wasn't true.

## Chef Version

12.21.4

## Platform Version

CentOS 7.3

## Replication Case

See above

jonlives:
### Description

This pull request introduces fixes to the behaviour of the various providers of the "package" resource (and subtypes like dnf_package etc) when a more recent version of the package has been installed than that present in an available repository or file source and "action :upgrade" is called against the package.

Currently, a number of package providers will attempt to *downgrade* a package under this scenario, where the desired action should be to recognise that a more up to date version is already installed and do nothing.

The PR introduces a ```version_compare``` method to the package superclass which by default uses ```Gem::Version``` comparison with spaceship-operator semantics to ensure that an upgrade is attempted only when the candidate version is *not* older than the current version.

For package management systems (such as dnf) requiring more specific implementations of ```version_compare```, a method has been added to the relevant subclass.

This PR has been tested against the following package managers (and associated providers), and behaves correctly (that is, Chef will not attempt to upgrade to an older package version):

* apt
* dpkg
* rpm
* yum
* dnf
* gem
* zypper
* homebrew

Additionally, this PR introduces a method called ```version_equals?``` which is functionally identical to ```target_version_already_installed?``` (and that method now in fact calls ```version_equals?```. This has been done to clarify the purpose and functionality of that method, but since it's a public API, deprecation of the old method name will have to be done more gradually.

### Issues Resolved

Resolves #6352 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -323,10 +323,38 @@ class Chef
       #
       # Note that most likely we need a spaceship operator on versions that subclasses can implement
       # and we should have `version_compare(v1, v2)` that returns `v1 <=> v2`.
+
+      # This method performs a strict equality check between two strings representing version numbers
       #
+      # This function will eventually be deprecated in favour of the below version_equals function.
+
       def target_version_already_installed?(current_version, target_version)
-        return false unless current_version && target_version
-        current_version == target_version
+        version_equals?(current_version, target_version)
+      end
+
+      # Note that most likely we need a spaceship operator on versions that subclasses can implement
+      # and we should have `version_compare(v1, v2)` that returns `v1 <=> v2`.
+
+      # This method performs a strict equality check between two strings representing version numbers
+      #
+      def version_equals?(v1, v2)
+        return false unless v1 && v2
+        v1 == v2
+      end
+
+      # This function compares two version numbers and returns 'spaceship operator' style results, ie:
+      # if v1 < v2 then return -1
+      # if v1 = v2 then return  0
+      # if v1 > v2 then return  1
+      # if v1 and v2 are not comparable then return nil
+      #
+      # By default, this function will use Gem::Version comparison. Subclasses can reimplement this method
+      # for package-management system specific versions.
+      def version_compare(v1, v2)
+        gem_v1 = Gem::Version.new(v1)
+        gem_v2 = Gem::Version.new(v2)
+
+        gem_v1 <=> gem_v2
       end
 
       # Check the current_version against the new_resource.version, possibly using fuzzy
@@ -337,7 +365,7 @@ class Chef
       # `version_satisfied_by?(version, constraint)` might be a better name to make this generic.
       #
       def version_requirement_satisfied?(current_version, new_version)
-        target_version_already_installed?(current_version, new_version)
+        version_equals?(current_version, new_version)
       end
 
       # @todo: extract apt/dpkg specific preseeding to a helper class
@@ -439,15 +467,18 @@ class Chef
             each_package do |package_name, new_version, current_version, candidate_version|
               case action
               when :upgrade
-                if target_version_already_installed?(current_version, new_version)
+                if version_equals?(current_version, new_version)
                   # this is an odd use case
                   Chef::Log.debug("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
                   target_version_array.push(nil)
-                elsif target_version_already_installed?(current_version, candidate_version)
+                elsif version_equals?(current_version, candidate_version)
                   Chef::Log.debug("#{new_resource} #{package_name} #{candidate_version} is already installed")
                   target_version_array.push(nil)
                 elsif candidate_version.nil?
                   Chef::Log.debug("#{new_resource} #{package_name} has no candidate_version to upgrade to")
+                  target_version_array.push(nil)
+                elsif version_compare(current_version, candidate_version) == 1 && !new_resource.allow_downgrade
+                  Chef::Log.debug("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{candidate_version}. Skipping...)")
                   target_version_array.push(nil)
                 else
                   Chef::Log.debug("#{new_resource} #{package_name} is out of date, will upgrade to #{candidate_version}")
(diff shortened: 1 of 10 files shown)


jonlives:
Travis testing has identified a couple of version number permutations under redhat and debian that are causing issues, working on fixing that (along with one remaining chef-style issue...).

lib/chef/provider/package.rb
```diff
@@ -304,7 +304,7 @@ def expand_options(options)
         end
       end
 
-      # Check the current_version against either the candidate_version or the new_version
+      # Check the current_version against either the candidate_version or the new_version		 +
```
thommay: that trailing `+` is weird.

jonlives: Whoops, I've removed that in my branch, will commit with the fixes :)

jaymzh:
Thanks @jonlives  this is awesome!

I had another thought on a test-case that might be worth doing... did you test this with multipackage rules (both kinds[1])? From reading the code it looks like it should work fine, but it's a good thing to test before merging.

[1] Both a `package ['something', 'somethingelse'] do action :upgrade end` / `package ['something', 'somethingelse'] do action :install; version ['someversion', 'anotherversion'] end` as well as an actual `multipackage` resource from @lamont-granquist's `multipackage` cookbook. There's a lot of weird interactions between these.

jonlives:
Valid point @jaymzh! I shall make sure that's tested!

jonlives:
I've tested all three multipackage incarnations as suggested by @jaymzh above, and all are good - the one remaining issue to fix is when you ask the rpm provider to *actually* downgrade - adding that fix now.

lib/chef/provider/package.rb
```diff
@@ -337,7 +365,7 @@ def target_version_already_installed?(current_version, target_version)
       # `version_satisfied_by?(version, constraint)` might be a better name to make this generic.
       #
       def version_requirement_satisfied?(current_version, new_version)
-        target_version_already_installed?(current_version, new_version)
+        version_equals?(current_version, new_version)
```
lamont-granquist: i think you still have to call the old version here so that overriding it works.  although you might want to `responds_to?` it here, and throw a deprecation warning and call it, then call `version_equals?` -- then actually delete or comment out the definition of the method here entirely.

coderanger:
I don't think we can call this a fix, it's a compat change and would have to wait for 14. My understanding of the package code has been that when an explicit `version` is given, the action (install vs. upgrade) is effectively ignored and the version is used as per normal, or at least that's how I've characterized it to people in the past. Happy to be wrong though.

lib/chef/provider/package.rb
```diff
@@ -439,16 +467,19 @@ def target_version_array
...
                   Chef::Log.debug("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
                   target_version_array.push(nil)
-                elsif target_version_already_installed?(current_version, candidate_version)
+                elsif version_equals?(current_version, candidate_version)
                   Chef::Log.debug("#{new_resource} #{package_name} #{candidate_version} is already installed")
                   target_version_array.push(nil)
                 elsif candidate_version.nil?
                   Chef::Log.debug("#{new_resource} #{package_name} has no candidate_version to upgrade to")
                   target_version_array.push(nil)
+                elsif version_compare(current_version, candidate_version) == 1 && !new_resource.allow_downgrade
```
coderanger: Based on my limited understanding of the package machinery, I think this will catch an explicit version too? If so that would be a potentially undesirable behavior change. I very well may be misreading things though.

jonlives: It would be simple enough to skip this step if new_resource.version has been specified, which would keep behaviour of preferring an explicit version over the action - sound good?

jaymzh: I think the current behavior if you specify a version would be to crash because you can't downgrade, but I'd have to check...

jonlives: So having tested a sample size of two (yum and dnf), we don't behave consistently here. DNF will see things as up to date and do nothing, yum will act as if there's no current version and upgrade from uninstalled -> specified version. The question is whether we want to have all providers behave as they currently do (ie not consistently), or standardise this somewhat..but that then brings us to @coderanger's original point that this might be a compat change.

lamont-granquist:
@coderanger the behavior of specifying a new_version on the upgrade action seems to be almost entirely undefined, since if you fail the installed_version != new_version equality check then you will go on an install the candidate_version (which probably does not satisfy the new_version).  we should have an `if new_version` check that ultimately winds up installing the new_version, or also does a check to ensure the candidate_version is satisfied by the new_version check before installing the candidate_version.  since the current implementation is incoherent, semver can't possibly apply.

we probably should do something more or less like this:

```
 case :upgrade
   if new_version
     if new_version.satisfied_by?(current_version)
       if new_version.satisifed_by?(candidate_version) && candidate_version >= current_version
         install(candidate_version)
       else
         do_nothing
       end
     else
       install(new_version)
     end
   elsif current_version >= candidate_version  # fix for bug
     do_nothing
   elsif candidate_version.nil?
     do_nothing
   else /* out of date*/
     install(candidate_version)
 case :install
   if new_version
     if new_version.satisfied_by?(current_version)
       do_nothing
     else
       install(new_version)
     end
   elsif current_version.nil?
     install(candidate_version)
   else
     do_nothing
   end
 end
```

where the first check in the `:upgrade` block handles explicit version pins, and would handle cases like `version "< 2"` where we'd allow upgrades from 1.2.3 to 1.2.4 to 1.3 up until 2.0 got released -- which actually usefully makes sense of how a version "pin" would work with `:upgrade` when mixed with the constraint syntax (which we support for DNF and need to support for YUM and APT and have users bitching at us like crazy that we don't support in other issues).

lamont-granquist:
(but all that is outside of the scope of this PR)

lamont-granquist:
Then for the use case where we're doing the `allow_downgrade` check here, that should probably go away:

```ruby
# install foo-1.2.3 if no foo package is installed
rpm_package "foo-1.2.3.rpm" do
  action :install
end

# install foo-1.2.3 if no package is installed or an older package is installed
rpm_package "foo-1.2.3.rpm" do
  action :upgrade
end

# install foo-1.2.3 if foo-1.2.3 is not installed
rpm_package "foo-1.2.3.rpm" do
  version "1.2.3"
  action :install   # or action :upgrade
end
```

PR #6433, opened 2017-09-22
Net:SSH config uses verify_host_key instead of paranoid

petecheslock:
Signed-off-by: Pete Cheslock <petecheslock@gmail.com>

### Description

In net-ssh 4.2.0 - they are deprecating the term `paranoid` and using `verify_host_key` instead as a config option.  When using knife ssh you will get the following warning when using chef 13.4.24 and net-ssh 4.2.0
```
:paranoid is deprecated, please use :verify_host_key. Supported values are exactly the same, only the name of the option has changed.
```

### Issues Resolved

The changes in net-ssh
https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt#L11
https://github.com/net-ssh/net-ssh/pull/524/files

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -270,7 +270,7 @@ class Chef
           opts[:port] = port unless port.nil?
           opts[:logger] = Chef::Log.logger if Chef::Log.level == :debug
           if !config[:host_key_verify]
-            opts[:paranoid] = false
+            opts[:verify_host_key] = false
             opts[:user_known_hosts_file] = "/dev/null"
           end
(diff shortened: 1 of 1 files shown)


tas50:
Seems like we should also set net-ssh 4.2 as the minimum version in the gemspec since seems to break < 4.2. Fun times in deprecation land.

adamleff:
I just came here to say what @tas50 said... he beat me to it.

lamont-granquist:
yep if this is a hard break in 4.2 we need a gemspec bump here as well.

thommay:
Fixed in #6869

Issue #6303, opened 2017-08-01
Chef cannot install chocolatey packages when passed in array from Artifactory


poblahblahblah:
## Description

It looks like chef does a `downcase` on chocolatey package names, which means packages like "GoogleChrome" cannot be installed via chef depending on your nuget/chocolatey repo.

## Chef Version

12.19.36

## Platform Version

windows2012r2
windows2016

## Replication Case

Note: We are using JFrog's Artifactory product as a chocolatey repo source which is case sensitive, so this issue is specific to that product. The official chocolatey repository works fine. I can provide private access to this for testing if required.

1. Have an artifactory chocolatey (nuget) repository
2. Upload the `GoogleChrome` choco package to artifactory
3. `chocolatey_package 'GoogleChrome'` in a recipe
4. Run chef-client.

## Client Output

```
       ---- Begin output of C:\ProgramData\chocolatey/bin/choco.exe install -y --version 60.0.3112.78 googlechrome ----
       STDOUT: Chocolatey v0.10.7
       Installing the following packages:
       googlechrome
       By installing you accept licenses for the packages.
       googlechrome not installed. The package was not found with the source(s) listed.
        If you specified a particular version and are receiving this message, it is possible that the package name exists but the version does not.
        Version: "60.0.3112.78"
        Source(s): "https://our-artifactory.jfrog.io/us/api/nuget/chocolatey"

       Chocolatey installed 0/1 packages. 1 packages failed.
        See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
```

## Stacktrace

N/A

dheerajd-msys:
@poblahblahblah Please share private repo access to test, however I am working on to setup my own Artifactory repo.

dheerajd-msys:
@poblahblahblah According to https://github.com/chocolatey/choco/wiki/CreatePackages#naming-your-package and https://github.com/chocolatey/choco/wiki/CreatePackagesQuickStart#common-mistakes there shouldn't be mixed case in names by `chocolatey naming convention`.

poblahblahblah:
@dheerajd-msys understood, although unfortunately this is not enforced even for officially maintained packages:

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/googlechrome/googlechrome.nuspec#L5

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/cygwin/cygwin.nuspec#L5

[...]

PR #6437, opened 2017-09-25
[MSYS-647] array support for choco pkg from artifactory

dheerajd-msys:
### Description

chocolatey_package installs packages in array for Artifactory server as source repo

### Issues Resolved
Fixes https://github.com/chef/chef/issues/6303

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -177,7 +177,13 @@ EOS
         # @return [Array] list of candidate_version, same index as new_resource.package_name/version
         def build_candidate_versions
           new_resource.package_name.map do |package_name|
-            available_packages[package_name.downcase]
+            if new_resource.source && new_resource.source.include?("artifactory") && package_name_array.length > 1
+              available_packages_from_artifactory[package_name.downcase]
+            elsif new_resource.source && new_resource.source.include?("artifactory") && package_name_array.length == 1
+              available_packages[package_name.downcase]
+            else
+              available_packages[package_name.downcase]
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/chocolatey.rb
```diff
@@ -177,7 +177,11 @@ def choco_command(*args)
         # @return [Array] list of candidate_version, same index as new_resource.package_name/version
         def build_candidate_versions
           new_resource.package_name.map do |package_name|
-            available_packages[package_name.downcase]
+            if new_resource.source && new_resource.source.include?("artifactory")
```
lamont-granquist: are we matching against the url here?  that would be a little bit brittle.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -238,6 +242,32 @@ def available_packages
           @available_packages
         end
 
+        # Getting packages from Artifactory server and return as candidate
+        def available_packages_from_artifactory
+          return @available_packages_from_artifactory if @available_packages_from_artifactory
+          @available_packages_from_artifactory = {}
+          package_name_array.length.times do |item|
```
lamont-granquist: this looks like it should just be `package_name_array.each do |pkg|` and `list -r #{pkg}` below.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -238,6 +242,32 @@ def available_packages
...
+            available_versions =
+              begin
+                cmd = [ "list -r #{package_name_array[item]}" ]
+                cmd.push( "-source #{new_resource.source}" )
+                raw = parse_list_output(*cmd).first.each_slice(2).to_h
+                raw1 = parse_list_output(*cmd)
```
lamont-granquist: can we fold this logic into `available_packages()`, so that we inspect the output of `list -r` to determine if we need to do the extra processing?  then there'd be no magic of inspecting the `source` property for "artifactory" and it would collapse it all down to one codepath.

dheerajd-msys:
@lamont-granquist Please review.

lamont-granquist:
the ruby looks good, should probably get some review from @mwrock

PR #6442, opened 2017-09-26
fix password property is sensitive for mount resource 

### LINKED ISSUE #4850: Mount provider should support sensitive property
## Description

I am using Chef-vault to try and keep my passwords secure. I am currently using one of the vault Items to try and mount a share. 

My problem is that the error output from chef printed the RAW password, not the encrypted or hashed out version, meaning anyone can circumvent the chef-vault by creating a bad mount recipe!

Would it not be better to hash out ALL 'password' attributes/fields in when printing them so that they are not shown to users (who might not be intended to see them!) or stored in log files?
## Chef Version

Chef: 12.6.0
## Platform Version

Windows 10
## Replication Case

  mount 'O:' do
    device '\\myShare\myFolder'
    action :mount
    domain 'myDomain'
    username testUser['username']
    password testUser['password']
  end
## Client Output

Error output:
    mount("O:") do
      action [:mount]
      supports {:remount=>false}
      retries 0
      retry_delay 2
      mount_point "O:"
      device "'\\myShare\myFolder'"
      device_type :device
      fsck_device "-"
      fstype "auto"
      options ["defaults"]
      dump 0
      pass 2
      username "muUser"
      password "myPassword"
      domain "myDomain"
      declared_type :mount
      cookbook_name "sn_users"
      recipe_name "test_user"
    end

dimsh99:
Signed-off-by: dmitrys <dmitrys@northernlight.com>

### Description

Password properties become sensitive by default

### Issues Resolved

#4850

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -242,11 +242,7 @@ class Chef
     # @return [Boolean]
     #
     def sensitive?
-      if name == :password
-        options.fetch(:sensitive, true)
-      else
-        options.fetch(:sensitive, false)
-      end
+      options.fetch(:sensitive, false)
     end
 
     #
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -242,7 +242,11 @@ class Chef
     # @return [Boolean]
     #
     def sensitive?
-      options.fetch(:sensitive, false)
+      if name == :password
+        options.fetch(:sensitive, true)
+      else
+        options.fetch(:sensitive, false)
+      end
     end
 
     #
--- a/lib/chef/resource/mount.rb
+++ b/lib/chef/resource/mount.rb
@@ -149,7 +149,6 @@ class Chef
         set_or_return(
           :username,
           arg,
-          :sensitive => true,
           :kind_of => [ String ]
         )
       end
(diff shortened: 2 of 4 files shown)


coderanger:
Not a fan of this approach. It's cute but too much magic for me to be in Chef core. 4850 should be fixed in the mount resource itself IMO.

coderanger:
To expand, keying this off the name is not super robust. If a property was named `user_password`, for example, or if the code isn't written in English. It's not a bad assumption but it's overly narrow and we can't possibly parse all human-readable names into secret and not-secret. Better to use the existing explicit API of `property(:password, sensitive: true)`.

lamont-granquist:
pretty sure we're following that pattern of marking the password property as sensitive on an individual resource now somewhere as well...

it looks like i was the one that mentioned doing this globally and that was probably not one of my best ideas...

lamont-granquist:
`execute.rb:      property :password, String, sensitive: true`

lamont-granquist:
`remote_file.rb:      property :remote_password, String, sensitive: true`

coderanger:
This also makes the `username` property be sensitive, is that needed?

dimsh99:
It doesn't make it sensitive, but it's my mistake made while removing old :password definition. I'm really sorry for that.

thommay:
This is a great addition, thank you!

PR #6447, opened 2017-09-26
Force encoding to UTF_8 in chef-shell to prevent failures

### LINKED ISSUE #6446: chef-shell / irb fail to launch on SUSE/omnios/other platforms
There's a few problems here. rb-readline needs infocmp which doesn't ship out of the box on suse. If we install ncurses-devel to get that command it still fails. It's basically the same bug as #4842, but I'm expanding on it here since there's an impact platforms we fully support.

```
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:2138:in `block in _rl_read_init_file': invalid byte sequence in US-ASCII (ArgumentError)
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:2136:in `each_line'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:2136:in `_rl_read_init_file'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:2109:in `rl_read_init_file'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:2573:in `readline_initialize_everything'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:3849:in `rl_initialize'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/rbreadline.rb:4868:in `readline'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/rb-readline-0.5.5/lib/readline.rb:45:in `readline'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/input-method.rb:151:in `gets'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb.rb:473:in `block (2 levels) in eval_input'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb.rb:627:in `signal_status'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb.rb:472:in `block in eval_input'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:189:in `buf_input'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:104:in `getc'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/slex.rb:206:in `match_io'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/slex.rb:76:in `match'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:283:in `token'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:265:in `lex'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:236:in `block (2 levels) in each_top_level_statement'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:232:in `loop'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:232:in `block in each_top_level_statement'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:231:in `catch'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb/ruby-lex.rb:231:in `each_top_level_statement'
	from /opt/chef/embedded/lib/ruby/2.4.0/irb.rb:489:in `eval_input'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.4.24/lib/chef/shell.rb:82:in `block in start'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.4.24/lib/chef/shell.rb:81:in `catch'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.4.24/lib/chef/shell.rb:81:in `start'
	from /opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.4.24/bin/chef-shell:34:in `<top (required)>'
	from /opt/chef/embedded/bin/chef-shell:23:in `load'
	from /opt/chef/embedded/bin/chef-shell:23:in `<main>'
```

## Chef Version

Chef 13.4
## Platform Version

Omnios r15101
openSUSE 42.1

tas50:
Currently on some distros where the default isn't UTF (SUSE/omnios/probably others) chef-shell fails to launch with a nasty rb-readline error. It's not the best user experience. Lifting this line from Berkshelf fixes it and makes it easier to troubleshoot on those platforms.

Closes #6446 

### DIFF
--- a/bin/chef-shell
+++ b/bin/chef-shell
@@ -23,6 +23,8 @@ begin
 rescue LoadError
 end
 
+Encoding.default_external = Encoding::UTF_8
+
 require "irb"
 require "irb/completion"
 require "irb/ext/save-history"


tas50:
We should probably note this in the release notes since it may impact the behavior of chef-shell for users.

coderanger:
What does this do if the user explicitly has LC_ALL (or any other) set to something like Shift-JIS?

lamont-granquist:
@coderanger, I don't exactly know, but nobody has reported issues against berkshelf.   Since internally we're 8-bit clean, and in the cookbook space we're UTF-8 by default, this produces the least bugginess in the system.  We also force encoding to en_US.UTF-8 internally in chef-client and I have never seen someone who has needed to fiddle with the config there and as a result I'm somewhat doubtful that writing all that config code was ever necessary.  I also suspect that setting the encoding to anything other than UTF-8 will just push bugs around (precisely because cookbooks are UTF-8 by default, so anyone who thinks they solve a problem by setting the encoding to something else will likely wind up sadpanda with other bugs down the road...) and so that would boil down to telling the user they really needed to fix the problem some other way.  If we actually wind up with a customer who wants to have their whole ecosystem setup with Shift-JIS, will never consume external cookbooks with UTF-8 chars in them, and wants to write their own Shift-JIS cookbooks, then we can cross that bridge when we hit it.  So far, haven't hit that one.  The more common case is someone in Europe with ISO-8859-1 set in their environment, trying to consume cookbooks which have UTF-8 characters embedded in them (because author names) and this patch fixes that.

JSON is another external thing that forces everything to UTF-8, although we clean that for reporting and node data, so anything that doesn't transcode just gets a UTF '<?>' character in it...

PR #6460, opened 2017-09-28
Bump InSpec to v1.40.0

adamleff:


tas50:
Foodcritic failure is because cucumber was bumped and the world is failing right now. We need to fix that, but it's unrelated to this PR

Issue #6411, opened 2017-09-15
After receiving SIGHUP chef-client daemon goes to sleep and never wakes up


grekasius:
## Description

After receiving `SIGHUP`, `chef-client` daemon goes to sleep and never wakes up.

## Chef Version

Reproduced on stable releases: `13.2.20`, `13.3.42`, `13.4.19`, `13.4.24`
Seems to be introduced by: [a1e3870](https://github.com/chef/chef/commit/a1e38709047c3afd0ad6ced65e959b68a37075a5)

## Platform Version

CentOS 6.8

## Replication Case

1. Start `chef-client` in daemon mode, e.g.
```
chef-client -d -i 120 -s 0
```
2. Observe Chef runs triggering at specified interval
3. Send `SIGHUP` to `chef-client` daemon process
4. Observe Chef runs to stop triggering

Note: If you attach `strace` to `chef-client` daemon process, after sending `SIGHUP`, you will see it calling `futex` that never completes.

I suspect the problem is caused by this line: https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L504

grekasius:
Submitted PR to fix this: https://github.com/chef/chef/pull/6461

PR #6461, opened 2017-09-29
Sleep for another interval after handling SIGHUP

grekasius:
Signed-off-by: Giedrius Rekasius <giedrius.rekasius@gmail.com>

### Description

This change resolves a bug of chef-client service going into infinite sleep after handling SIGHUP. The logic is to sleep for another interval after handling SIGHUP.

### Issues Resolved

chef-client service going to infinite sleep after handling SIGHUP (Issue: [6411](https://github.com/chef/chef/issues/6411))

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -501,7 +501,7 @@ class Chef::Application::Client < Chef::Application
       # we need to sleep again after reconfigure to avoid stampeding when logrotate runs out of cron
       if signal == RECONFIGURE_SIGNAL
         reconfigure
-        interval_sleep(sleep)
+        interval_sleep(sec)
       end
     else
       sleep(sec)


lamont-granquist:
does this mask the problem by just hammering on reconfigure until the sleep time finally counts down to zero?  because it would be better to solve the problem by fixing why reconfigure is getting hammered on...

grekasius:
@lamont-granquist `reconfigure` runs once for each `SIGHUP` received. After `SIGHUP` is handled, `IO.select` gets called again and blocks for remaining interval time.

Before this fix, `chef-client` would go into infinite sleep because of this: https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L504

lamont-granquist:
Yeah I'm missing some context about the shape of the infinite loop.  It might be useful to actually see a section of the stacktrace showing the infinite loop.  Trying to compile that in my head it seems like you'd have to be repetitively reading the same value here:  https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L497  which seems like it would be more like the root cause of the bug.  The ruby compiler in my head is known to be occasionally buggy though.

grekasius:
@lamont-granquist the problem is with `sleep` rather than the shape of the loop. `sleep` is called with no arguments in https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L504, so the process never wakes up. The simplest way to see this in action is to attach `strace` to `chef-client`, send `SIGHUP` and observe all activity to stop indefinitely after process goes to sleep.

In my fix I change two things:
1. After `reconfigure`, `sleep` with no arguments is replaced with going back to `IO.select` with the timeout set to the remaining interval time. This allows `chef-client` to keep handling subsequent signals while waiting for the remaining interval to expire. Not sure whether it would make more sense to reset the interval after `reconfigure`, that might be up for debate.
2. Replace recursion with loop to avoid filling the stack if large amount of signals is received.

btm:
Can this be as simple as https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L504 being changed from `interval_sleep(sleep)` to `interval_sleep(sec)` ? I think `sleep` as an argument there is an error from the last refactor here.

grekasius:
Sure, if looping through recursion is not a concern (it would take sending thousands of `SIGHUPs` to fill the stack and cause `StackError`, shouldn't happen under normal circumstances) and we prefer to reset interval after `reconfigure`, then the fix could be as simple as `interval_sleep(sleep)` -> `interval_sleep(sec)` and I believe it should work fine then.

btm:
I think that would be simplest. Off the top of my head I don't remember all the use cases here, but I've been in this code enough to know that it's a delicate balance of them.

I *think* it's something like this:

In a run:
- HUP reconfigures after the run completes
- USR1 runs again after the run completes
- TERM exits after the run completes

Sleeping:
- HUP reconfigures and goes back to sleep.
- USR1 runs
- TERM exits cleanly

grekasius:
@btm updated PR according to your suggestion. Did some testing with this fix applied and it seems to work fine.

lamont-granquist:
LGTM

PR #6467, opened 2017-10-02
Deprecate vendor branches with `knife cookbook site install`

coderanger:
So we can kill it for Chef 14. This adds a CLI option to not vendor and warns if you don't use it. I think I did the option correctly but I should probably make some kind of test for this.

💣 💥

### DIFF
--- a/lib/chef/knife/cookbook_site_install.rb
+++ b/lib/chef/knife/cookbook_site_install.rb
@@ -101,10 +101,12 @@ class Chef
         #cookbook_path = File.join(vendor_path, name_args[0])
         upstream_file = File.join(@install_path, "#{@cookbook_name}.tar.gz")
 
-        @repo.sanity_check
-        unless config[:use_current_branch] || config[:no_vendor]
-          @repo.reset_to_default_state
-          @repo.prepare_to_import(@cookbook_name)
+        unless config[:no_vendor]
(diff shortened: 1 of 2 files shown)


lib/chef/knife/cookbook_site_install.rb
```diff
@@ -90,7 +102,7 @@ def run
         upstream_file = File.join(@install_path, "#{@cookbook_name}.tar.gz")
 
         @repo.sanity_check
```
lamont-granquist: should this also be protected by `unless config[:no_vendor]`?  and possibly `@repo` shouldn't be built unless we're vendoring (although scope...)

tas50:
@coderanger Care to rebase this one?

tas50:
There's some test failures here @coderanger, but it would be nice to get this one in before the deprecation window closes.

lamont-granquist:
unit tests are actually failing 'cuz new deprecation warnings get thrown

tas50:
@lamont @coderanger Any reason we don't just warn now and yank this in 15? Seems a lot simpler than requiring folks to put it in the configs to keep it until we ship the next major.

lamont:
You probably wanted @lamont-granquist

tas50:
Closing in favor of #7012

PR #6468, opened 2017-10-03
Deprecate the deploy resource and family

coderanger:
@tas50 has volunteered (really, I have witnesses) to make the migration cookbook.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,16 @@
 _This file holds "in progress" release notes for the current release under development and is intended for consumption by the Chef Documentation team. Please see <https://docs.chef.io/release_notes.html> for the official Chef release notes._
 
+# Chef Client Release Notes 13.6:
+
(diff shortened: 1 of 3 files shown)


danielsdeleo:
pour one out

coderanger:
@tas50 Updated :)

coderanger:
Also I put up https://github.com/chef/chef-rfc/pull/282, covering this and two other resources I think should be on a path towards removal.

jjasghar:
I still remember the first time I used `deploy` resource trying to convert from `capistrano`, I'll miss you `deploy`, you helped teach me`chef`!

coderanger:
Going to merge this even though the RFC in question is still under discussion because it doesn't seem like there is any major objection to this (compared to the other two) and I want to get the warning in as soon as possible to start turning the migration crank.

coderanger:
Hmm, except merging seems to be jammed because it thinks Travis didn't report in?

PR #6470, opened 2017-10-03
[Route unit tests]: Ensure we go down the new file path

thommay:
This should make us more robust in the face of the enemy


spec/unit/provider/route_spec.rb
```diff
@@ -236,9 +236,11 @@
         @node.automatic_attrs[:platform] = platform
 
         route_file = StringIO.new
+        allow(File).to receive(:exists?).with("/etc/sysconfig/network").and_return(false)
```
lamont-granquist: update to `File.exist?` while you're here?

PR #6471, opened 2017-10-04
Fix to use cookbook name in local mode or chef-zero

### LINKED ISSUE #4816: cookbook name not taken into account in local mode or chef-zero
## Description

In local mode, the directory name is used as the name of the cookbook, not the `name` in metadata.rb

So if you have a cookbook called `foo`, but the dir name is `foo_test` and inside of `foo` you do `include_recipe 'foo::some_recipe'` it'll complain that:

```
Cookbook foo not found. If you're loading foo from another cookbook, make sure you configure the dependency in your metadata
```

Frankly, I was a proponent of adding a Foodcritic rule that required to be the same but @lamont-granquist  (I think) chimed in that having them able to be separate is super important. I ended up using this to test something and it turns out not to work.
## Chef Version

12.8.1
## Platform Version

CentOS 7
## Replication Case

See above.
## Client Output

See above.

dimsh99:
### Description

Cookbook name and version from metadata are used in chef_fs_data_store instead of directory name

### Issues Resolved

closes #4816

### Check List

- [x] All tests pass


lamont-granquist:
should be careful and make sure travis passes as much as is possible before we whack merge here.

dimsh99:
I believe travis check is hanging since October and needs to be reset.

thommay:
Looks great, thanks!

Issue #6474, opened 2017-10-05
OpenBSD package provider fails to install packages


e100:
## Description

The package provider fails to find candidate versions to install because PKG_PATH variable is not set when calling pkg_info.
Additionally, installing packages with empty flavors always fails.

## Chef Version: 13.5.3 - Chef Zero
## Platform Version OpenBSD 6.1
## Replication Case
Try to install any empty flavor package that is currently not installed
`package 'rsync--'`
## Client Output

`Chef::Exceptions::Package                                                                          
          -------------------------                                                                          
          No candidate version available for rsync--                                                         
                                                                                                             
          Resource Declaration:                                                                              
          ---------------------                                                                              
          # In /tmp/kitchen/cache/cookbooks/test/recipes/default.rb                         
                                                                                                             
           77: package 'rsync--'                                                                             
           78: package 'iftop'                                                                               
                                                                                                             
          Compiled Resource:                                                                                 
          ------------------                                                                                 
          # Declared in /tmp/kitchen/cache/cookbooks/test/recipes/default.rb:77:in `from_file`

PR #6475, opened 2017-10-05
Fix OpenBSD package

e100:
### Description

Updated candidate_version to include PKG_PATH env var when calling pkg_info.
Modified regex for 'flavor' handling to allow installation of empty flavors

### Issues Resolved
https://github.com/chef/chef/issues/6474

### DIFF
--- a/lib/chef/provider/package/openbsd.rb
+++ b/lib/chef/provider/package/openbsd.rb
@@ -101,7 +101,7 @@ class Chef
         def candidate_version
           @candidate_version ||= begin
             results = []
-            shell_out_compact_timeout!("pkg_info", "-I", package_string(new_resource.package_name, new_resource.version), env: nil, returns: [0, 1]).stdout.each_line do |line|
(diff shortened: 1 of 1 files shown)


thommay:
the chefstyle issues are fixed on master, so @chef/client-maintainers let's get this reviewed please.

lib/chef/provider/package/openbsd.rb
```diff
@@ -57,7 +57,7 @@ def define_resource_requirements
           end
           requirements.assert(:all_actions) do |a|
             a.assertion do
-              if new_resource.package_name =~ /^(.+?)--(.+)/
+              if new_resource.package_name =~ /^(.+?)--(.?+)/
```
lamont-granquist: i'm slightly puzzled by the regular expression here since it changes it so that "foo--bar" will return "b" instead of "bar" because the match is now non-greedy, which doesn't match what i'd expect for "empty flavor" since I would have expected just ".*" or maybe "(?:--(.*))" or something like that...  definitely seems to need either some better comment documentation or ideally some tests.

cheeseplus:
@e100 any thoughts on the above comments/review? We'd love to help you get this merged.

tas50:
@e100 I'm going to close this out since we haven't heard back at this point. If you want to move forward with this feel free to open it back up with a response or updated code.

PR #6478, opened 2017-10-05
Remove cookbook_artifacts from CHEF_11_OSS_STATIC_OBJECTS

itmustbejj:
Signed-off-by: Josh Hudson <jhudson@chef.io>

Fixes an issue where newer versions of ChefFS cannot be used with Chef 11 servers because policyfiles did not exist then. I believe this was a mistaken addition (it is included in `CHEF_12_OBJECTS` properly) that only affected Chef 11 servers.

### DIFF
--- a/lib/chef/chef_fs/config.rb
+++ b/lib/chef/chef_fs/config.rb
@@ -66,7 +66,7 @@ class Chef
       # upgrade/migration of older Chef Servers, so they should be considered
       # frozen in time.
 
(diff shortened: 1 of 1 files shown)


itmustbejj:
Please hold off on merging for now. Needs more investigation.

itmustbejj:
Ok, I feel really good about the scope of this change. You need to set `repo_mode` to "static" in your knife.rb for this change to have any effect.

PR #6483, opened 2017-10-06
Remove the knife null plugin

tas50:
This is a pretty odd plugin that we're shipping. I suspect there's a reason for it so I opened this PR to get the conversation going.


lamont-granquist:
i think its for testing knife timing issues to see how much of it is just due to ruby module loading garbage

tas50:
It came along with the hash functionality so I suspect that's true, but it's in the help page, which is less than great.

lamont-granquist:
Yeah, this should stay in.   Hiding it from help is another kind of issue.

stevendanna:
I'd be 👍 on figuring out how to hide it from help or to give it a clearer help message.  We could remove it but I do think it is useful to keep around.

Alternatively, we could make it a bit more useful by renaming it and having it print some diagnostic information; but then it is less useful for benchmarking and such.

lamont-granquist:
yeah i think its utility is in benchmarking, so it really should just do nothing.

PR #6486, opened 2017-10-09
Support new CriticalOhaiPlugins

jaymzh:
### Description

We should not stack-trace, we should exit with a clean error

Signed-off-by: Phil Dibowitz <phil@ipom.com>

### Check List

- [X] All tests pass


jaymzh:
This will break until Ohai is released and I can pin it...

jaymzh:
Oh, it passes, but I probably *should* pin to >= to a version... but we haven't released one.
@tas50 ...?

jaymzh:
Turns out it's fine to not pin since you can catch exception classes that are undefined. Merging.

PR #6489, opened 2017-10-10
allow #attribute? to check deep

atheiman:
### Description

Allows `Chef::Node#attribute?` and `Chef::Node::Attribute#attribute?` to check deep into the attribute Mash for existence of attributes.

### Issues Resolved

Couldnt find any related github issues from a quick search, but these are top results on google for `chef nested attribute`:
- https://discourse.chef.io/t/node-attribute-check-nested/4191/2
- https://stackoverflow.com/questions/18986096/what-is-the-correct-way-to-check-for-the-existence-of-a-nested-attribute-in-chef

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -244,11 +244,8 @@ class Chef
 
     # Return true if this Node has a given attribute, false if not.  Takes either a symbol or
     # a string.
-    #
-    # Only works on the top level. Preferred way is to use the normal [] style
-    # lookup and call attribute?()
-    def attribute?(attrib)
-      attributes.attribute?(attrib)
+    def attribute?(*attrib)
+      attributes.attribute?(*attrib)
     end
 
     # Yield each key of the top level to the block.
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -207,16 +207,11 @@ class Chef
 
       # check for the existence of attributes
       def attribute?(*args)
-        r = nil
+        r = self
         until args.empty?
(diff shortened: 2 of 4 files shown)


lamont-granquist:
i think you're looking for `#exist?` here: https://github.com/chef/chef/blob/master/lib/chef/node/attribute.rb#L448-L450

its also a common api across both VividMash/AttrArray and ImmutableMash/Array:

https://github.com/chef/chef/blob/c2869b01147182ce6155e2fbb434b6d45a0dbf0c/lib/chef/node/common_api.rb#L70-L89

so `node.default.exist?("foo", "bar", "baz")` should work against only the default level.

atheiman:
wow thank you its even in chef 12

`exist?` does exactly what i want. i just tested out with these specs:

```ruby
      node.normal["deep"]["nested"]["attribute"] = "blue"

      expect(node.exist?("deep", "nested")).to eq(true)
      expect(node.exist?("deep", "wrong")).to eq(false)
      expect(node["deep"].exist?("nested")).to eq(true)
      expect(node["deep"].exist?("wrong")).to eq(false)
      expect(node.exist?("deep", "nested", "attribute")).to eq(true)
      expect(node.exist?("deep", "nested", "wrong")).to eq(false)
      expect(node["deep"]["nested"].exist?("attribute")).to eq(true)
      expect(node["deep"]["nested"].exist?("wrong")).to eq(false)
      expect(node.exist?("deep", "nested", "attribute", "wrong")).to eq(false)
```

PR #6435, opened 2017-09-22
dnf_resource: be more specific for rhel packages

NaomiReeves:
### Description
With #6351 and #6377, the DNF package provider was fixed to use yum by
default on RHEL 7, but the resource was still calling the DNF provider
if it happened to be installed.

We want to use the default of yum for RHEL 7, and only use DNF if we
call `set_priority_map` directly.

Signed-off-by: Naomi Reeves <naomi.c.reeves@gmail.com>

### Issues Resolved
Unexpectedly switching package managers

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #6490, opened 2017-10-10
only rhel >= 8 and fedora >= 22 get dnf

lamont-granquist:
the shell_out approach to parse the version string from rpm is a bit of
a failed experiment.

the shell_out that gets incurred on every package provider is a bit
terrible for performance.

DNF < 1.00 has also never formally landed in any distribution and its
very difficult at this point to deploy it.

when amazon deploys DNF we should add a version comparison for that.

If this patch causes issues we can go back to adding some form of

`provides :package ... { which("dnf" }`

That will be much faster than having the shell_out().

### DIFF
--- a/lib/chef/resource/dnf_package.rb
+++ b/lib/chef/resource/dnf_package.rb
@@ -29,11 +29,11 @@ class Chef
 
       allowed_actions :install, :upgrade, :remove, :purge, :reconfig, :lock, :unlock, :flush_cache
 
-      provides :package, platform_family: %w{fedora amazon} do
-        which("dnf") && shell_out("rpm -q dnf").stdout =~ /^dnf-[1-9]/
-      end
+      # all rhel variants >= 8 will use DNF
+      provides :package, platform_family: "rhel", platform_version: ">= 8"
 
-      provides :package, platform_family: %{rhel}, platform_version: ">= 8"
+      # fedora >= 22 uses DNF
+      provides :package, platform: "fedora", platform_version: ">= 22"
 
       provides :dnf_package
 
--- a/lib/chef/resource/dnf_package.rb
+++ b/lib/chef/resource/dnf_package.rb
@@ -35,6 +35,11 @@ class Chef
       # fedora >= 22 uses DNF
       provides :package, platform: "fedora", platform_version: ">= 22"
 
+      # amazon will eventually use DNF
+      provides :package, platform: "amazon" do
+        which("dnf")
+      end
+
       provides :dnf_package
 
       # Install a specific arch
--- a/lib/chef/provider/package/dnf.rb
+++ b/lib/chef/provider/package/dnf.rb
@@ -41,6 +41,11 @@ class Chef
         # fedora >= 22 uses DNF
         provides :package, platform: "fedora", platform_version: ">= 22"
 
+        # amazon will eventually use DNF
+        provides :package, platform: "amazon" do
+          which("dnf")
+        end
+
         provides :dnf_package
 
         #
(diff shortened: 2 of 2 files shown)


lamont-granquist:
This replaces #6435 and ensures that both the provider+resource provides lines match

Paging @jaymzh and @NaomiReeves

lamont-granquist:
(there are no spec changes because its the same behavior for all the platforms we care about, just less noodling about unlikely edge conditions we never tested for)

tas50:
We might want to keep this logic for amazon for now so we will have working versions of opsworks for amazon when they make that leap

danielsdeleo:
I don't think OpsWorks uses the package resource anywhere. OTOH, we do have a decent chunk of folks on Amazon Linux. I'll leave it up to y'all to decide on how to deal with that one.

lamont-granquist:
Okay, the `which("dnf")` piece is back, only for amazon linux, still removing the shell_out tho.

lib/chef/resource/dnf_package.rb
```diff
@@ -29,11 +29,16 @@ class DnfPackage < Chef::Resource::Package
...
+      # all rhel variants >= 8 will use DNF
+      provides :package, platform_family: "rhel", platform_version: ">= 8"
+
+      # fedora >= 22 uses DNF
+      provides :package, platform: "fedora", platform_version: ">= 22"
 
-      provides :package, platform_family: %{rhel}, platform_version: ">= 8"
+      # amazon will eventually use DNF
+      provides :package, platform: "amazon" do
+        which("dnf")
```
jaymzh: provides gets resolved for every single resource... are we walking the filesystem looking for DNF on every single resource (and were we forking rpm each time)? This seems... awful (though not affected by this PR just a thing I noticed)...

lamont-granquist: which just does some `File.exist?` aka `stat(2)` calls searching the `ENV['PATH']` in pure ruby, its pretty lightweight.

straight from your graph profiles:

```
   3.45%   0.01%      4.948      0.010      0.000      4.938             4265     Chef::NodeMap#block_matches?
                      4.830      0.003      0.000      4.827          124/525     Chef::Mixin::ShellOut#shell_out
                      0.055      0.004      0.000      0.051          387/387     <Class::Chef::Platform::ServiceHelpers>#service_resource_providers
                      0.046      0.002      0.000      0.044          124/254     Chef::Mixin::Which#which
                      0.005      0.005      0.000      0.000      3630/208599     NilClass#nil?
                      0.001      0.001      0.000      0.000       635/680809     Kernel#nil?
                      0.001      0.001      0.000      0.000        387/43146     Array#include?
```

So 4.83 seconds (probably about half that because of profiling overhead) in shell_out with the old code, while 0.046 seconds in `which` so almost exactly 100x faster.

And hopefully the fact that this is amazon-linux-only means it doesn't run on anything else (i would certainly *not* bet on that -- but we're now talking about tens of milliseconds total if i'm wrong and not seconds). 

That shell_out (along with the fact that the resource provides statements and the provider provides statements have drifted) is the whole reason for this.   And this should shave about 1.7 secs off your runs.

jaymzh: That's on your non-IO-loaded system. I"m hacking RPM and other utilities left-and-right to not touch the disk if they don't need to...  so saving a few stats for **every single** package_resources makes a big difference.

jaymzh: (those profiles are also on non-IO-loaded systems. We can get you one on a heavily IO loaded hadoop box if you wanna wait like...  a few days or so for it to run)

lamont-granquist: Since this is `platform: "amazon"` though there's no reason for it to be running on centos.  That would be a bug that we need to fix in the NodeMap stuff.  And this is intended to be temporary and replaced once amazon finally goes DNF at some point in the future.

PR #6494, opened 2017-10-12
Ensure package (un)locking is idempotent

### LINKED ISSUE #6361: Package locking is not idempotent when package and resource name differ
## Description

Package locking is not idempotent when package and resource name differ. This seems to be caused by the provider incorrectly using the resource name when determining if a package is locked. 

## Chef Version

12.18.31

## Platform Version

Red Hat Enterprise Linux Server release 6.8

## Replication Case

Declare a yum_package resource where the resource name and package name differ and locking is used. Example:

```
yum_package 'install-apache' do
  package_name 'httpd'
  version '2.2.15-60.el6_9.5'
  action %i(install lock)
end
```

In this case, "install-apache" is passed as a argument to `package_locked`:
https://github.com/chef/chef/blob/3c42edc942619b634a3f3dc7e44cf4699532f208/lib/chef/provider/package.rb#L223

Note that when the actual locking occurs, `package_name` is correctly used:
https://github.com/chef/chef/blob/3c42edc942619b634a3f3dc7e44cf4699532f208/lib/chef/provider/package.rb#L225

This behavior results in the package being locked on every Chef run and duplicate entries in the version lock list (`yum versionlock list`).

## Client Output

N/A

## Stacktrace

N/A

### LINKED ISSUE #6493: Package locking is not idempotent
## Description

Due to the way the `package_locked` method has been written, it will always return false as it is comparing a string to an array:
* https://github.com/chef/chef/blob/master/lib/chef/provider/package/yum.rb#L107
* https://github.com/chef/chef/blob/master/lib/chef/provider/package/zypper.rb#L83
* https://github.com/chef/chef/blob/master/lib/chef/provider/package/apt.rb#L78

## Chef Version

13.4.19

## Platform Version

Ubuntu 16.04

## Replication Case

```
package 'magic' do
  action %i[lock]
end
```
```
package %w[magic bowtie] do
  action %i[lock]
end
```

## Client Output

* The continual updating of each package resource that is being locked. 
* Locked packages will never be unlocked.

## Stacktrace

N/A

Rarian:
### Description

Due to the nature of the comparison done, the `package_locked` methods
will always return false as they presume the package name is a string,
when it is in fact always coerced into an array.

Additionally in situations where the package_name is set rather than
being inherited from the package resource's name, it will always return
false.

### Issues Resolved

Resolves #6361
Resolves #6493

### Check List

- [x] New functionality includes tests, tests were added for the package_name versus resource name, and idempotency of locking/unlocking inside apt; zypper and yum tests were not added.
- [x] All tests pass

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -220,7 +220,7 @@ class Chef
       end
 
       def action_lock
-        if package_locked(new_resource.name, new_resource.version) == false
+        unless package_locked(new_resource.package_name, new_resource.version)
           description = new_resource.version ? "version #{new_resource.version} of " : ""
           converge_by("lock #{description}package #{current_resource.package_name}") do
             multipackage_api_adapter(current_resource.package_name, new_resource.version) do |name, version|
@@ -234,7 +234,7 @@ class Chef
       end
 
       def action_unlock
-        if package_locked(new_resource.name, new_resource.version) == true
+        if package_locked(new_resource.package_name, new_resource.version)
           description = new_resource.version ? "version #{new_resource.version} of " : ""
           converge_by("unlock #{description}package #{current_resource.package_name}") do
             multipackage_api_adapter(current_resource.package_name, new_resource.version) do |name, version|
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -71,15 +71,11 @@ class Chef
         end
 
         def package_locked(name, version)
-          islocked = false
           locked = shell_out_compact_timeout!("apt-mark", "showhold")
-          locked.stdout.each_line do |line|
-            line_package = line.strip
-            if line_package == name
-              islocked = true
-            end
+          locked_packages = locked.stdout.each_line.map do |line|
+            line.strip
           end
-          islocked
+          name.all? { |n| locked_packages.include? n }
         end
 
         def install_package(name, version)
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -100,15 +100,11 @@ class Chef
 
         # @see Chef::Provider::Package#package_locked
         def package_locked(name, version)
-          islocked = false
           locked = shell_out_with_timeout!("yum versionlock")
-          locked.stdout.each_line do |line|
-            line_package = line.sub(/-[^-]*-[^-]*$/, "").split(":").last.strip
-            if line_package == name
-              islocked = true
-            end
+          locked_packages = locked.stdout.each_line.map do |line|
+            line.sub(/-[^-]*-[^-]*$/, "").split(":").last.strip
           end
-          islocked
+          name.all? { |n| locked_packages.include? n }
         end
 
         #
--- a/lib/chef/provider/package/zypper.rb
+++ b/lib/chef/provider/package/zypper.rb
@@ -76,15 +76,11 @@ class Chef
         end
 
         def package_locked(name, version)
-          islocked = false
           locked = shell_out_compact_timeout!("zypper", "locks")
-          locked.stdout.each_line do |line|
-            line_package = line.split("|").shift(2).last.strip
-            if line_package == name
-              islocked = true
-            end
+          locked_packages = locked.stdout.each_line.map do |line|
+            line.split("|").shift(2).last.strip
           end
-          islocked
+          name.all? { |n| locked_packages.include? n }
         end
 
         def load_current_resource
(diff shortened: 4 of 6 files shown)


lib/chef/provider/package/apt.rb
```diff
@@ -71,15 +71,11 @@ def candidate_version
...
-          locked.stdout.each_line do |line|
-            line_package = line.strip
-            if line_package == name
-              islocked = true
-            end
+          locked_packages = locked.stdout.each_line.map do |line|
+            line.strip
           end
-          islocked
+          name.reject { |n| locked_packages.include? n }.empty?
```
lamont-granquist: nice cleanup as well

lib/chef/provider/package.rb
```diff
@@ -220,7 +220,7 @@ def removing_package?
       end
 
       def action_lock
-        if package_locked(new_resource.name, new_resource.version) == false
+        if package_locked(new_resource.package_name, new_resource.version) == false
```
lamont-granquist: can you make this `unless package_locked` and just `if package_locked` below?  its not 100% logically equivalent but the case of not-true/not-false values like nil would be false in both cases, which is pretty buggy and incoherent anyway i think?

you could introduce:

```ruby
def package_locked?
  !!package_locked
end
```

to clean up the API a bit (subclasses in external code outside of core chef might be using `package_locked` without the `?` mark so I don't think you can change that there, its somewhat implicitly a public API and changing it for a nitpick is probably not worth it).

lamont-granquist:
so specs for the superclass are here:

https://github.com/chef/chef/blob/master/spec/unit/provider/package_spec.rb#L324-L372

and individual classes:

https://github.com/chef/chef/blob/master/spec/unit/provider/package/apt_spec.rb#L408-L428
https://github.com/chef/chef/blob/master/spec/unit/provider/package/yum_spec.rb#L778-L794
https://github.com/chef/chef/blob/master/spec/unit/provider/package/zypper_spec.rb#L262-L306

i think i'd be happy with just a test or two in the superclass, which seems to be the most critical bit.

lib/chef/provider/package.rb
Rarian: I like the change to just `unless` and `if`. 

Unsure if I'm just not following the ramifcations of a `package_locked?` method in regards to cleaning up the code, but I'm missing something there as I don't understand how it would actually clean up the API.

lamont-granquist: that would let you `if package_locked?` in the superclass where it would still call `package_locked` in the subclass but would be coerced to a boolean by `!!` and make the superclass at least read a little more idiomatically ruby.  IDK tho.  that was more of a thought.

Rarian:
Let me know if the added tests are sufficient or if I need to add some more.

lib/chef/provider/package/apt.rb
coderanger: Better clean up:

```ruby
locked_packages = locked.stdout.each_line.map(&:strip)
name.any? { |n| locked_packages.include?(n) }
```

That will stop if it finds one. Also I'm not sure what the semantics of this for situations with multi-package and partial locked input. I think you changed the logic from `any?` to `all?`

tas50:
@Rarian Can you rebase this? There's a pile of failures in Travis and Appveyor, but I believe we fixed them with new CI configs. I can't tell for sure unless we get this updated to the current master.

Rarian:
@tas50 Rebased, will see if everything is happy this time around.

lib/chef/provider/package/apt.rb
```diff
@@ -71,15 +71,11 @@ def candidate_version
...
 
         def package_locked(name, version)
-          islocked = false
           locked = shell_out_compact_timeout!("apt-mark", "showhold")
-          locked.stdout.each_line do |line|
-            line_package = line.strip
-            if line_package == name
-              islocked = true
-            end
+          locked_packages = locked.stdout.each_line.map do |line|
```
adamleff: A fun Ruby shorthand for this is:

```ruby
locked_packages = locked.stdout.each_line.map(&:strip)
```

You don't have to change it - just want to impart knowledge I've learned from much smarter people than me. :)

lib/chef/provider/package/apt.rb
```diff
@@ -71,15 +71,11 @@ def candidate_version
...
+            line.strip
           end
-          islocked
+          name.all? { |n| locked_packages.include? n }
```
adamleff: Small nitpick: I'd love to see this be `include?(n)` so we remain consistent with method params being wrapped in parentheses. Not a blocker for this, just something to think about for your next change.

PR #6496, opened 2017-10-13
registry_key: Add sensitive property support for suppressing output (fixes #5695)

### LINKED ISSUE #5695: [Enhancement] Add sensitive to registry_key
## Description

Windows registry_key resource doesn't support sensitive.  While it does write to Windows registry that is not very secure, it would be nice to suppress stdout and stderr output. Here is an example of output when setting autologin username and password:

```
* windows_autologin[set rdp_local] action enable
             * registry_key[set AutoAdminLogon for rdp_local] action create
        - set value {:name=>"DefaultUsername", :type=>:string, :data=>"rdp_local"}
        - set value {:name=>"DefaultPassword", :type=>:string, :data=>"%2wNrk6z%Y-eJrv"}
```

It would be nice to be able to hide this from the logs.

https://ci.appveyor.com/project/dhoer/chef-windows-screenresolution/build/build-28#L375

shoekstra:
### Description

This PR adds the `sensitive` property to `registry_key` but needs some help before it's ready to be merged.

I'm not sure if the `sensitive: true` sets the `values` property to default to sensitive or if when `sensitive` is true on the resource, `values` will be masked.

Re tests, looking at the `execute` resource it looks like I should be copying across something like below, but not sure where it should sit.

```
    context "when sensitive is set true" do
      it "should honor sensitive attribute" do
        new_resource.sensitive true
        # Since the resource is sensitive, it should not have :live_stream set
        opts.delete(:live_stream)
        expect(provider).to receive(:shell_out_with_systems_locale!).with(new_resource.name, opts)
        expect(provider).to receive(:converge_by).with("registry_key sensitive resource").and_call_original
        expect(Chef::Log).not_to receive(:warn)
        provider.run_action(:run)
        expect(new_resource).to be_updated
      end
    end
```

### Issues Resolved

#5695 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/registry_key.rb
+++ b/lib/chef/provider/registry_key.rb
@@ -126,12 +126,18 @@ class Chef
               value[:data] = value[:data].to_i
             end
             unless current_value[:type] == value[:type] && current_value[:data] == value[:data]
-              converge_by("set value #{value}") do
+              converge_by_value = value
+              converge_by_value[:data] = "*sensitive value suppressed*" if new_resource.sensitive
+
+              converge_by("set value #{converge_by_value}") do
                 registry.set_value(new_resource.key, value)
               end
             end
           else
-            converge_by("set value #{value}") do
+            converge_by_value = value
+            converge_by_value[:data] = "*sensitive value suppressed*" if new_resource.sensitive
+
+            converge_by("set value #{converge_by_value}") do
               registry.set_value(new_resource.key, value)
             end
           end
@@ -146,7 +152,10 @@ class Chef
         end
         new_resource.unscrubbed_values.each do |value|
           unless @name_hash.has_key?(value[:name].downcase)
-            converge_by("create value #{value}") do
+            converge_by_value = value
+            converge_by_value[:data] = "*sensitive value suppressed*" if new_resource.sensitive
+
+            converge_by("create value #{converge_by_value}") do
               registry.set_value(new_resource.key, value)
             end
           end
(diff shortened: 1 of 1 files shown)


tas50:
@shoekstra There's a large number of spec failures due to the validation change in Travis

lib/chef/resource/registry_key.rb
```diff
@@ -67,6 +67,8 @@ def initialize(name, run_context = nil)
         @values, @unscrubbed_values = [], []
       end
 
+      property :sensitive, [ TrueClass, FalseClass ], default: false, coerce: proc { |x| password ? true : x }
```
coderanger: Just to follow up from Slack since I'm now on a real keyboard, this is a weird thing to add. The `sensitive` property is supported on all resources already, and the coerce doesn't actually help with masking passwords in most cases. I think you mean to mark the `password` property sensitive instead?

lib/chef/resource/registry_key.rb
```diff
@@ -75,12 +77,12 @@ def key(arg = nil)
...
-      def values(arg = nil)
-        if not arg.nil?
-          if arg.is_a?(Hash)
-            @values = [ arg ]
-          elsif arg.is_a?(Array)
-            @values = arg
+      property :values, [ Hash, Array ], default: nil, sensitive: true, desired_state: false, coerce: proc { |v|
+        if not v.nil?
+          if v.is_a?(Hash)
+            @values = [ v ]
```
coderanger: The coerce method should be returning the modified value, rather than directly touching instance variables.

tas50:
@shoekstra Any chance you have some time to take a crack at @coderanger's feedback. It would be nice to get sensitive wired up here.

lib/chef/resource/registry_key.rb
lamont-granquist: yeah we also ran into cases where people want to pass a password and want sensitive off for debugging or something like that i think.

shoekstra:
Sorry for the delayed response!

After some conversations with @coderanger we realised I was totally off, the problem lay in the `converge_by` blocks in the provider which should be resolved now.

PR #6497, opened 2017-10-13
fix rebooter for solaris and background reboots

### LINKED ISSUE #5026: Reboot resource on linux needs to fork the shutdown process
When using the `reboot` resource on centos or ubuntu, `shutdown -r` with some time period and reason is executed and blocks until the time period runs out.  Then chef-client races to finish against the shutdown in progress.

This can break testing scenarios that require a reboot via test-kitchen as the new exit codes aren't returned consistently (depending on who wins the race).

lamont-granquist:
closes #5026

### DIFF
--- a/lib/chef/platform/rebooter.rb
+++ b/lib/chef/platform/rebooter.rb
@@ -33,14 +33,18 @@ class Chef
         def reboot!(node)
           reboot_info = node.run_context.reboot_info
 
-          cmd = if Chef::Platform.windows?
+          cmd = case
+                when Chef::Platform.windows?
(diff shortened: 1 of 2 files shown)


lamont-granquist:
note that the existing specs are a big klunky and this certainly exposes that they're a bit klunky, but i don't have the cycles to clean everything up here right now....

lamont-granquist:
i also didn't actually test the solaris code, but its super damn broken right now as it is, because solaris does not speak BSD shutdown.

lamont-granquist:
@chef/client-maintainers pingles

PR #6498, opened 2017-10-15
Apt repo cleanup and testing expansion

tas50:
So this all started when @coderanger pointed out that we didn't need to use the run_command on the shell_out helper method since it does that internally already. I fixed that here and realized the logic could be broken out to ease readability and expand testing. I added yard comments, broke out some logic into smaller methods, added more unit tests, and added a bit more debug logging.

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -200,7 +200,13 @@ class Chef
         end
       end
 
-      def install_key_from_keyserver(key, keyserver = new_resource.keyserver)
+      # build the apt-key command to install the keyserver
+      #
+      # @param [String] key the key to install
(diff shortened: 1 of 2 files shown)


tas50:
Adding yard mostly came along for free when adding simple method comments. Adding a param and return code was pretty simple at that point. It probably doesn't add a huge amount, but it's pretty simple and makes it a bit easier to read through later on when modifying that code. I don't really see anyone reading this in an actual web browser though since there's not really any need / reason.

tas50:
Hold off on this one. I'm working on a key validation issue and will get apt_repository into the Travis integration tests

tas50:
I ran this through some more tests and I'm not able to see the original problem with chef 14 now. Looks good to merge

PR #6503, opened 2017-10-17
Causes immediate exit on calls to Chef::Application.fatal!

RyanFrantz:
Signed-off-by: Ryan Frantz <ryanleefrantz@gmail.com>

### Description

This update forces the Chef run to exit immediately on a call to `Chef::Application.fatal!` rather than raising a `SystemExit` exception.

I need `chef-client` to exit immediately as I'm working on a start handler that looks for a node attribute and stops the Chef run if present. That handler looks a bit like this:

```ruby
require 'chef/handler'
require 'chef/application'

class Chef::Handler::IsChefDisabled < Chef::Handler
  def report
    Chef::Application.fatal!("chef-client is disabled (node['disable_chef'] is defined). Exiting converge!", 1) if node['disable_chef']
  end
end
```

I noticed that calling `Chef::Application.fatal!` doesn't cause the Chef run to terminate. In fact, it completes the converge phase. Calling `Process.exit!` stops the Chef run immediately. An immediate exit seems like appropriate behavior for `Chef::Application.fatal!`

For extra context, we run `chef-client` on a cron like so:

```
# Chef Name: chef-client
0,15,30,45 * * * * /usr/bin/chef-client --run-lock-timeout 300 --no-fork --lockfile /var/run/chef/chef-client.lock > /dev/null 2>&1
```

**NOTE**: This update does not modify the behavior for `Chef::Application.exit!`; it still calls `Process.exit` which will raise the `SystemExit` exception.

### Issues Resolved

N/A

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -355,7 +355,7 @@ class Chef
       # Log a fatal error message to both STDERR and the Logger, exit the application
       def fatal!(msg, err = nil)
         Chef::Log.fatal(msg)
-        Process.exit(normalize_exit_code(err))
+        Process.exit!(normalize_exit_code(err))
       end
 
       def exit!(msg, err = nil)


lamont-granquist:
That might be the way it should have been written originally, but I'm a little skeptical that we can change that behavior at this point.

We use that call several times in this file and also elsewhere in the code:

https://github.com/chef/chef/search?utf8=%E2%9C%93&q=Application.fatal%21&type=

And the problem is that I'm not sure what the implications are for all those codepaths, or if this change is appropriate for all of them.  I wouldn't be too surprised if some of the specs blow up (although I also wouldn't feel particularly confident that this patch wouldn't cause backcompat issues even if all the specs do pass).

Its most likely better if you just `Process.exit!` directly in your handler if you need to do that.  We could maybe accept some kind `#fatal_and_i_mean_it!` extension to the API (obviously with a better name), but changing the existing behavior without having bugs against the existing code is making me suspicious of backcompat.

RyanFrantz:
If not `#fatal_and_i_mean_it!`, then mayhaps `catastrophe!`?

In all seriousness, I think your recommendation of calling `Process.exit!` in the handler is best in the short term. I wanted to submit this PR because after trawling the code it felt like someone _meant_ `Process.exit!` in `fatal!`.

Thanks for the review.

coderanger:
It shouldn't exit immediately, but if it isn't aborting the run ASAP then we might be catching SystemExit somewhere that we shouldn't be. Calling `exit!` is a very big hammer and breaks all kinds of things like resources that would be released in an `ensure` block, signal handlers, etc.

lamont-granquist:
Yeah `#exit!` means leaking tempfiles in /tmp and things like that which always get reported as bugs (because destructors and ensure blocks are missed...)

PR #6508, opened 2017-10-18
Don't use the safe navigation operator which requires Ruby 2.3

tas50:
This was backported from Chef 13 where we require Ruby 2.3. This won't work in Ruby 2.2.


aalexmonteiro:
I would like to know what the forecast of this release is in The RubyGems. Because I need this fix, otherwise I will do a workaround in my cookbook.

PR #6510, opened 2017-10-18
Fix remote_file with UNC paths failing

tas50:
Our check here to see if we're on Windows uses node data that's not
available in this context. Use the same Chef::Platform.windows? check we
use above. Without this you get the following error:

[2017-10-02T21:40:42+00:00] DEBUG: Re-raising exception: NameError - remote_file[c:/foo/bar] (foo::default line 14) had an error: NameError: undefined local variable or method `node' for #<Chef::Provider::RemoteFile::NetworkFile:0x00000000064c0148>

### DIFF
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -23,7 +23,7 @@ class Chef
     module UserContext
 
       def with_user_context(user, password, domain = nil, &block)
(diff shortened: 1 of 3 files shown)


lamont-granquist:
heh so we stub the `#node` method even though it had none, but its not a `double` so we can't use `instance_double`...

Issue #6210, opened 2017-06-10
knife ssh doesn't work with different identity file for gateway and node


davidlawson:
## Description

The following command:

```
knife ssh 'name:<node-name>' 'sudo chef-client' --ssh-gateway <gateway-user>@<gateway-ip> --ssh-gateway-identity ~/.ssh/<gateway-identity>.pem --ssh-identity-file ~/.ssh/<node-identity>.pem --ssh-user <node-user>
```

attempts to use the `~/.ssh/<gateway-identity>.pem` file for both the gateway and the node. Even specifying a non-existent file for `--ssh-identity-file` does not throw a warning/error.

## Chef Version

```
Chef Development Kit Version: 2.0.12
chef-client version: 13.0.118
delivery version: master (651e8c7e37fc80d6e1867a176974bf41fb08aa98)
berks version: 6.1.0
kitchen version: 1.16.0
inspec version: 1.26.0
```

## Platform Version

Local: macOS 10.12.5
Gateway & node: Ubuntu 16.04.2 LTS

## Client Output

```
...
DEBUG: establishing connection to bastion:22
DEBUG: connection established
...
DEBUG: trying publickey (b8:d3:68:05:39:44:54:03:22:2b:63:d3:d5:2d:7f:b4)
...
DEBUG: publickey succeeded (b8:d3:68:05:39:44:54:03:22:2b:63:d3:d5:2d:7f:b4)
...
DEBUG: Using node attribute 'fqdn' as the ssh target: ip-10-0-3-133.us-west-2.compute.internal
...
DEBUG: trying publickey (b8:d3:68:05:39:44:54:03:22:2b:63:d3:d5:2d:7f:b4)
...
ERROR: all authorization methods failed (tried none, publickey)
DEBUG: read 0 bytes
WARNING: Failed to connect to ip-10-0-3-133.us-west-2.compute.internal -- Net::SSH::AuthenticationFailed: Authentication failed for user ubuntu@127.0.0.1@ip-10-0-3-133.us-west-2.compute.internal
INFO: local forwarded connection closed
DEBUG: queueing packet nr 14 type 97 len 28
DEBUG: /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/net-ssh-multi-1.2.1/lib/net/ssh/multi/server.rb:194:in `rescue in new_session'
[...]

PR #6514, opened 2017-10-20
Added Flag to distinguish between gateway and host key to fix issue #6210

erikparra:
Signed-off-by: Erik <Erik.Parra4@gmail.com>

### Description

When using option --ssh-gateway-identity and -ssh-identity, both with be utilized by the session_options definition.

### Issues Resolved

[Issue #6210 ](https://github.com/chef/chef/issues/6210) Resolves the issue of --ssh-gateway-identity being used for both gateway and target node.

### Check List

- [X] All tests pass

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -285,7 +285,7 @@ class Chef
         list.each do |item|
           host, ssh_port = item
           Chef::Log.debug("Adding #{host}")
-          session_opts = session_options(host, ssh_port, 0)
+          session_opts = session_options(host, ssh_port, gateway: false)
           # Handle port overrides for the main connection.
           session_opts[:port] = Chef::Config[:knife][:ssh_port] if Chef::Config[:knife][:ssh_port]
           session_opts[:port] = config[:ssh_port] if config[:ssh_port]
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -248,14 +248,14 @@ class Chef
       # @param host [String] Hostname for this session.
       # @param port [String] SSH port for this session.
       # @param user [String] Optional username for this session.
-      # @param gatway [Boolean] Flag: host or gateway key
+      # @param gateway [Boolean] Flag: host or gateway key
       # @return [Hash<Symbol, Object>]
-      def session_options(host, port, user = nil, gateway = false)
+      def session_options(host, port, user = nil, gateway: false)
         ssh_config = Net::SSH.configuration_for(host, true)
         {}.tap do |opts|
           # Chef::Config[:knife][:ssh_user] is parsed in #configure_user and written to config[:ssh_user]
           opts[:user] = user || config[:ssh_user] || ssh_config[:user]
-          if gateway == false && config[:ssh_identity_file]
+          if !gateway && config[:ssh_identity_file]
             opts[:keys] = File.expand_path(config[:ssh_identity_file])
             opts[:keys_only] = true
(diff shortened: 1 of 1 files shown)


lib/chef/knife/ssh.rb
```diff
@@ -247,19 +247,20 @@ def search_nodes
       # @since 12.5.0
       # @param host [String] Hostname for this session.
       # @param port [String] SSH port for this session.
+      # @param type [Integer] Flag: host or gateway key
       # @param user [String] Optional username for this session.
       # @return [Hash<Symbol, Object>]
-      def session_options(host, port, user = nil)
+      def session_options(host, port, type = 0, user = nil)
```
lamont-granquist: this should be a named parameter:

```
def session_options(host, port, user = nil, gateway = false)
```

then there's no need to pass it `false` in the normal host case, but the gateway would call it like:

```ruby
session_options(gw_host, gw_port, gw_user, gateway: true)
```

which makes it a lot more obvious what is happening in that line than having a strange '1' as an option.

cheeseplus:
Fixes #6210

cheeseplus:
@erikparra we've reviewed the code above and if you can make some changes accordingly we'd love to help you merge this PR.

lib/chef/knife/ssh.rb
```diff
@@ -247,18 +247,18 @@ def search_nodes
       # @since 12.5.0
       # @param host [String] Hostname for this session.
       # @param port [String] SSH port for this session.
-      # @param type [Integer] Flag: host or gateway key
       # @param user [String] Optional username for this session.
+      # @param gatway [Boolean] Flag: host or gateway key
```
thommay: `gateway` not `gatway`

lib/chef/knife/ssh.rb
```diff
@@ -247,18 +247,18 @@ def search_nodes
...
       # @return [Hash<Symbol, Object>]
-      def session_options(host, port, type = 0, user = nil)
+      def session_options(host, port, user = nil, gateway = false)
```
thommay: `gateway: false`, rather than `gateway = false` - in the code above you're using a named argument, but here you're using a positional argument. In general, we prefer a named argument so if you can make the two line up I think we're good to go.

lib/chef/knife/ssh.rb
```diff
@@ -247,18 +247,18 @@ def search_nodes
...
         ssh_config = Net::SSH.configuration_for(host, true)
         {}.tap do |opts|
           # Chef::Config[:knife][:ssh_user] is parsed in #configure_user and written to config[:ssh_user]
           opts[:user] = user || config[:ssh_user] || ssh_config[:user]
-          if type == 0 && config[:ssh_identity_file]
+          if gateway == false && config[:ssh_identity_file]
```
thommay: `if !gateway && config[:ssh_identity_file]` is more idiomatic ruby in this case.

lib/chef/knife/ssh.rb
```diff
@@ -284,7 +285,7 @@ def session_from_list(list)
         list.each do |item|
           host, ssh_port = item
           Chef::Log.debug("Adding #{host}")
-          session_opts = session_options(host, ssh_port)
+          session_opts = session_options(host, ssh_port, 0)
```
thommay: this should be `gateway: false`

thommay:
@erikparra ok, one last fix, and then please do a DCO signoff

thommay:
Great, thank you so much for working through this with us. Thanks for your contribution to Chef!

PR #6520, opened 2017-10-25
Use the latest libxml2, libxslt, libyaml, and openssl

tas50:
libxml2:
  A GIANT list of bugfixes and these CVEs:

CVE-2017-9050
CVE-2017-9049
CVE-2017-9048
CVE-2017-9047
CVE-2017-8872
CVE-2016-9318

https://www.cvedetails.com/vulnerability-list/vendor_id-1962/product_id-3311/Xmlsoft-Libxml2.html

libxslt:
 - Fixes bad memory handling and null derefs plus a GIANT list of bug

libyaml:
  * Fixed segfault in yaml_string_write_handler.
  * Fixed invalid simple key assertion.
  * Fixed error handling in some examples (thank to Mathias Svensson).
  * Removed obsolete VS project files.

openssl:

CVE-2017-3731 (OpenSSL advisory)  [Moderate severity] 26th January 2017:
If an SSL/TLS server or client is running on a 32-bit host, and a specific cipher is being used, then a truncated packet can cause that server or client to perform an out-of-bounds read, usually resulting in a crash. For OpenSSL 1.1.0, the crash can be triggered when using CHACHA20/POLY1305; users should upgrade to 1.1.0d. For Openssl 1.0.2, the crash can be triggered when using RC4-MD5; users who have not disabled that algorithm should update to 1.0.2k Reported by Robert Święcki of Google.

CVE-2017-3732 (OpenSSL advisory)  [Moderate severity] 26th January 2017:
There is a carry propagating bug in the x86_64 Montgomery squaring procedure. No EC algorithms are affected. Analysis suggests that attacks against RSA and DSA as a result of this defect would be very difficult to perform and are not believed likely. Attacks against DH are considered just feasible (although very difficult) because most of the work necessary to deduce information about a private key may be performed offline. The amount of resources required for such an attack would be very significant and likely only accessible to a limited number of attackers. An attacker would additionally need online access to an unpatched system using the target private key in a scenario with persistent DH parameters and a private key that is shared between multiple clients. For example this can occur by default in OpenSSL DHE based SSL/TLS ciphersuites. Note: This issue is very similar to CVE-2015-3193 but must be treated as a separate problem. Reported by OSS-Fuzz project.

CVE-2016-7055 (OpenSSL advisory)  [Low severity] 10th November 2016:
There is a carry propagating bug in the Broadwell-specific Montgomery multiplication procedure that handles input lengths divisible by, but longer than 256 bits. Analysis suggests that attacks against RSA, DSA and DH private keys are impossible. This is because the subroutine in question is not used in operations with the private key itself and an input of the attacker's direct choice. Otherwise the bug can manifest itself as transient authentication and key negotiation failures or reproducible erroneous outcome of public-key operations with specially crafted input. Among EC algorithms only Brainpool P-512 curves are affected and one presumably can attack ECDH key negotiation. Impact was not analyzed in detail, because pre-requisites for attack are considered unlikely. Namely multiple clients have to choose the curve in question and the server has to share the private key among them, neither of which is default behaviour. Even then only clients that chose the curve will be affected.ctures using a callback which do not handle NULL value are affected. Reported by Publicly reported.

### Description

### Issues Resolved

### Check List

- [ ] All tests pass


tas50:
This isn't ready to merge as it  fails builds on Windows due to the libxslt patch not properly applying.

tas50:
I rant this SHA through Jenkins and it passed so I think this is good to merge. We're still seeing unit test failures in Travis that keep this from going green.

PR #6521, opened 2017-10-25
Include Ohai 13.6

btm:
Ohai Release Notes 13.6

- Critical Plugins

Users can now specify a list of plugins which are `critical`. Critical plugins will cause Ohai to fail if they do not run successfully (and thus cause a Chef run using Ohai to fail). The syntax for this is:

```
ohai.critical_plugins << :Filesystem
```

- Filesystem now has a `allow_partial_data` configuration option

The Filesystem plugin now has a `allow_partial_data` configuration option. If set, the filesystem will return whatever data it can even if some commands it ran failed.

- Rackspace detection on Windows

Windows nodes running on Rackspace will now properly detect themselves as running on Rackspace without a hint file.

- Package data on Amazon Linux

The Packages plugin now supports gathering packages data on Amazon Linux

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -224,7 +224,7 @@ GEM
     nori (2.6.0)
     octokit (4.7.0)
       sawyer (~> 0.8.0, >= 0.5.3)
-    ohai (13.5.0)
+    ohai (13.6.0)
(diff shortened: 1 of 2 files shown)


tas50:
Merging this since the travis failure is unrelated and the integration tests pass, which is what we actually care about here

PR #6523, opened 2017-10-25
Enable Fedora integration testing in Travis

tas50:
This pulls out the Fedora bits of #6504 which turned into a yak shave. This gets us a bit more platform coverage before we pull things into Jenkins and it doesn't add to the overall Travis time since we're still limited by the time the unit tests take to run.


tas50:
@thommay any objection to merging this in since travis is happy with the change?

PR #6529, opened 2017-10-26
Update the Ohai 13.6 release notes

tas50:


tas50:
The deprecations haven't required RFCs to create or update in the past.

btm:
I was presuming there was an RFC because of the reference to a deprecation plan. I guess I was wrong.

tas50:
Merging since this is actually part of the 13.6 release which is about to go out

PR #6537, opened 2017-10-27
ignore leading whitespace changes when displaying diffs

dirkm:
### Description

This patch ignores leading whitespace-changes when displaying diffs as well. For our use cases (some groovy files as templates) it stops lots of unnecessary diff-displaying in case of indentation-changes.

I  cannot judge what the overall preference is here. Maybe it should be made configurable, but for us it stops lots of complaints from the administrators.

### DIFF
--- a/lib/chef/util/diff.rb
+++ b/lib/chef/util/diff.rb
@@ -89,8 +89,8 @@ class Chef
         diff_str = ""
         file_length_difference = 0
 
-        old_data = IO.readlines(old_file).map { |e| e.chomp }
-        new_data = IO.readlines(new_file).map { |e| e.chomp }
+        old_data = IO.readlines(old_file).map { |e| e.strip }
+        new_data = IO.readlines(new_file).map { |e| e.strip }
         diff_data = ::Diff::LCS.diff(old_data, new_data)
 
         return diff_str if old_data.empty? && new_data.empty?


coderanger:
This feels overly opinionated on file formats for Chef to do automatically. This would be better handled in a custom formatter class of some kind maybe?

onlyhavecans:
Wouldn't this hide why a file is being changed if only whitespace is being touched? For some file formats whitespace is very important.

coderanger:
Also this wouldn't affect the diff for indentation, it would only hide whitespace changes at the start and end of the file which feels very niche.

dirkm:
Well, trailing whitespace is already ignored in the current version. This patch would extend it to leading whitespace. 

But your point is taken, and we can live with simply patching our local chef.

coderanger:
Chomp only removes a single newline (if present), not arbitrary trailing whitespace.

dirkm:
ok, you are right again. In my defense, most diff-tools have an ignore-whitespace option. I am not sure how this can be meaningfully incorporated in chef.

lamont-granquist:
yeah this cannot be on by default, and i'm skeptical this is the right way of going about it.   it would have to be twiddleable via chef-config and/or per-resource and should really implement proper whitespace-ignoring functions, but that probably means adding proper whitespace ignoring functions to diff-lcs?

thommay:
Given the complexities of white space diffing, we unfortunately can't accept this PR. Thank you so much for submitting it, and we'd love to see more from you.

PR #6540, opened 2017-10-27
Yum refactor

lamont-granquist:
🚧  under construction 🚧


lib/chef/provider/package/yum.rb
```diff
@@ -16,483 +15,224 @@
...
-        # Overload the Package provider to keep track of the YumCache
-        def initialize(new_resource, run_context)
-          super
+        provides :package, platform_family: %w{fedora amazon} do
+          which("yum")
+        end
 
-          @yum = YumCache.instance
-          @yum.yum_binary = yum_binary
+        provides :package, platform_family: %w{rhel}, platform_version: ">= 8"
```
tas50: Shouldn't this be < 8? >= 8 is DNF

lamont-granquist: yeah that's bad copypasta, fixed.

lamont-granquist:
@chef/client-maintainers i guess its time to review.

please don't hit merge yet, i'm validating it won't break omnibus builds (and strongly suspect it might break el5 if i don't check)

lamont-granquist:
Locking since we do not monitor closed issues and PR, please cut a new issue for any issues.

Issue #6549, opened 2017-11-01
Latest package upgrade behaviour doesn't cope with all debian version formats.


bogdanap:
## Description
Our builds recently started failing on apt package upgrades. My best assumption would be that the changes introduced in https://github.com/chef/chef/pull/6428 don't take into account all debian version formats, that could also include characters like `~` and `:` (in addition to `+` and `_`) which make Gem::Version unhappy. 

## Chef Version

13.6.0

## Platform Version
Ubuntu 14.04

## Replication Case

```
apt_repository 'pollinate' do
  uri 'http://ppa.launchpad.net/pollinate/ppa/ubuntu'
  distribution node['lsb']['codename']
  components %w[main]
  key '43732553'
  keyserver 'hkp://ha.pool.sks-keyservers.net'
  retries 2
  retry_delay 30
end

package 'pollinate' do
  action %i[install upgrade]
end
```

This is the complete recipe that currently fails at the upgrade package step:
https://github.com/travis-ci/travis-cookbooks/blob/master/cookbooks/travis_build_environment/recipes/cloud_init.rb#L34

## Client Output
```
    googlecompute:   * apt_package[pollinate] action upgrade
    googlecompute:
    googlecompute:     ================================================================================
    googlecompute:     Error executing action `upgrade` on resource 'apt_package[pollinate]'
    googlecompute:     ================================================================================
    googlecompute:
    googlecompute:     ArgumentError
    googlecompute:     -------------
    googlecompute:     Malformed version number string 4.23-0ubuntu1~14.04
```
Failing build: https://travis-ci.org/travis-infrastructure/packer-build/jobs/295254154

onlyhavecans:
I also have this issue

```
         * apt_package[ruby2.3] action upgrade

           ================================================================================
           Error executing action `upgrade` on resource 'apt_package[ruby2.3]'
           ================================================================================

           ArgumentError
           -------------
           Malformed version number string 2.3.5-1bbox2~trusty1

           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/dnsimple-server/recipes/_ruby.rb

            18: package 'ruby2.3' do
            19:   action :upgrade
            20:   notifies :reload, 'ohai[refresh_rubies]', :immediately
            21:   notifies :run, 'execute[reinstall_bundler]', :immediately
            22: end
            23:

           Compiled Resource:
           ------------------
[...]

PR #6554, opened 2017-11-02
Fix the invalid version comparison of apt packages.

komazarari:
### Description
The strings of dpkg versions are more complex than the Gem::Version
and not compatible with it.
Using Gem::Version to compare versions of dpkg packages will cause
mistakes or raise the exeption of 'Malformed version number..'.

please see a simple example:
https://gist.github.com/komazarari/c6b33b29a0e31e9c62bbc2f0b2091647

Prefer to use the dpkg cli.
https://debian-handbook.info/browse/en-US/stable/sect.manipulating-packages-with-dpkg.html

### Issues Resolved
none
### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -128,10 +128,15 @@ class Chef
         private
 
         def version_compare(v1, v2)
-          gem_v1 =  v1.gsub(/[_+]/, "+" => "-", "_" => "-") unless v1.nil?
-          gem_v2 =  v2.gsub(/[_+]/, "+" => "-", "_" => "-") unless v2.nil?
-
-          Gem::Version.new(gem_v1) <=> Gem::Version.new(gem_v2)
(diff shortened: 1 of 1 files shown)


bogdanap:
Fwiw, this should fix: https://github.com/chef/chef/issues/6549

lamont-granquist:
this looks reasonably fine.  i'm not positive the defensive programming is necessary.  the only real concern i'd have is performance, but it looks like it gets used at the tail end of a big if-else block in the superclass, so it isn't going to be called in most of the happy paths, and at any rate if performance becomes an issue then we just need to copy the c source into ruby -- and i'd vastly prefer to avoid that work if possible.

lamont-granquist:
should also ideally include some unit tests, tho, for the cases that we broke

tas50:
I've confirmed locally that this does the right thing when the package version contains a tilde

PR #6555, opened 2017-11-02
Change a useradd_spec test for RHEL >= 6.8 and >= 7.3

jeremiahsnapp:
RHEL >= 7.3 ships with a fixed `usermod` command.
Reference: https://access.redhat.com/errata/RHBA-2016:2322

Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>


jeremiahsnapp:
This should also be backported to Chef 12. Is there anything I need to do to make that happen?

PR #6557, opened 2017-11-02
Try using the old appveyor VS 2015 image

tas50:
They released a new build and we're getting name resolution errors now.
Lets try the old one so we can use this information to open a ticket.


tas50:
Didn't fix it

PR #6558, opened 2017-11-03
Fix the invalid version comparison in apt/dpkg providers

tas50:
This is builds on the work that @komazarari did in https://github.com/chef/chef/pull/6554 to fix comparison of version strings that Gem can't handle.

If we don't have a version installed yet we shouldn't go do a version
comparison with nil. No matter what that's a waste of time and in
dpkg/apt it's going to require shelling out now. Instead just do the
install.

This also adds the same version comparison method to the dpkg resource.
At the moment it doesn't look like this is going to get called since our
upgrade action there is a bit broken (doesn't actually compare
versions), but that's going to get fixed next.

This also cleans up the version comparison to make sure we're
always dealing with strings

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -127,12 +127,16 @@ class Chef
 
         private
 
+        # compare 2 versions to each other to see which is newer.
+        # this differs from the standard package method because we
+        # need to be able to parse debian version strings which contain
+        # tildes which Gem cannot properly parse
+        #
+        # @return [Integer] 1 if v1 > v2. 0 if they're equal. -1 if v1 < v2
         def version_compare(v1, v2)
-          dpkg_v1 = v1 || '0'
-          dpkg_v2 = v2 || '0'
-          if shell_out_compact_timeout("dpkg", "--compare-versions", dpkg_v1, "gt", dpkg_v2).status.success?
+          if !shell_out_compact_timeout("dpkg", "--compare-versions", v1.to_s, "gt", v2.to_s).error?
             1
-          elsif shell_out_compact_timeout("dpkg", "--compare-versions", dpkg_v1, "eq", dpkg_v2).status.success?
+          elsif !shell_out_compact_timeout("dpkg", "--compare-versions", v1.to_s, "eq", v2.to_s).error?
             0
           else
             -1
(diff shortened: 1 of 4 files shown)


lib/chef/provider/package/apt.rb
```diff
@@ -127,11 +127,23 @@ def unlock_package(name, version)
...
+        # @return [Integer] 1 if v1 > v2. 0 if they're equal. -1 if v1 < v2
         def version_compare(v1, v2)
-          gem_v1 =  v1.gsub(/[_+]/, "+" => "-", "_" => "-") unless v1.nil?
-          gem_v2 =  v2.gsub(/[_+]/, "+" => "-", "_" => "-") unless v2.nil?
-
-          Gem::Version.new(gem_v1) <=> Gem::Version.new(gem_v2)
+          # we need to be defensive here since we can be passed a nil
+          dpkg_v1 = v1.nil? ? 0 : v1.to_s
+          dpkg_v2 = v2.nil? ? 0 : v2.to_s
+          if shell_out_compact_timeout("dpkg", "--compare-versions", dpkg_v1, "gt", dpkg_v2).status.success?
```
lamont-granquist: this should raise if either v1 or v2 are nil, it makes no sense to have it silently translate it to a '0' here.

we already have a check that the candidate_version is not nil prior to calling this function in the superclass, and its likely that we need a check that the installed_version is not nil there.  and that is the appropriate place for it to go, which would short-circuit this test (if its not installed, and there's no installed_version it makes no sense to compare the versions so that use case should never fall through to this code).

that also makes the code simpler:  `shell_out_compact_timeout("dpkg", "--compare-versions", dpkg_v1.to_s, "gt", dpkg_v2.to_s).status.success?`

lamont-granquist:
i'd prefer to not see all the unrelated cleanup to the specs on a regression fix, which we're trying to ship on a friday, while we have a disaster happening on appveyor and travis.

lib/chef/provider/package/apt.rb
tas50: when I tested it in dpkg nils were being passed in

tas50:
I was going to actually write an apt spec, but this one can go into another PR now

lamont-granquist:
i'm mildly concerned about perf, but this gets the behavior entirely correct, and it is at the tail end of a large if-else chain.  we should either abort earlier than this, or else we're likely to wind up really doing the action which will shell out to something vastly more expensive anyway.

lamont-granquist:
(the apt utils are also substantially faster than all the rpm/yum/dnf stuff)

PR #6578, opened 2017-11-08
feat: Exclude extensions with remote_directory

clburlison:
### Description

This enhancement gives the ability to include an array of file
extensions that should not be purged. Extremely useful if shipping a
python module via chef since `.pyc` files will be created. However, as
files get changed, renamed, moved, etc. the directory will become dirty
without the `purge` option.

`exclude_extensions` is optional and takes an array:

    ['.tmp', '.pyc']

I am still trying to figure out how/where to add this as a test.

### Issues Resolved

n/a

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/remote_directory.rb
+++ b/lib/chef/provider/remote_directory.rb
@@ -36,7 +36,7 @@ class Chef
 
       def_delegators :new_resource, :purge, :path, :source, :cookbook, :cookbook_name
       def_delegators :new_resource, :files_rights, :files_mode, :files_group, :files_owner, :files_backup
-      def_delegators :new_resource, :rights, :mode, :group, :owner
+      def_delegators :new_resource, :rights, :mode, :group, :owner, :exclude_extensions
 
       # The overwrite property on the resource.  Delegates to new_resource but can be mutated.
       #
@@ -111,6 +111,9 @@ class Chef
 
             # Skip files that we've sync'd and their parent dirs
             next if managed_files.include?(file)
+
+            # Skip files that have a excluded extensions
+            next if exclude_extensions.find { |ext| file.end_with?(ext) }
 
             if ::File.directory?(file)
               if !Chef::Platform.windows? && file_class.symlink?(file.dup)
--- a/lib/chef/resource/remote_directory.rb
+++ b/lib/chef/resource/remote_directory.rb
@@ -46,6 +46,7 @@ class Chef
         @files_mode = 0644 unless Chef::Platform.windows?
         @overwrite = true
         @cookbook = nil
+        @exclude_extensions = []
       end
 
       if Chef::Platform.windows?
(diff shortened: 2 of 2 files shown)


coderanger:
Use the existing Chefignore file system instead?

coderanger:
Also for this specific instance you can use `export  PYTHONDONTWRITEBYTECODE=1` or `python -B`. This behavior is also much improved in Python 3.

erikng:
This is hugely beneficial to managing macOS.

erikng:
@coderanger macOS ships with Python 2 and this assumes that the python code can be modified or even ran in another context.

coderanger:
If we were going to do this it should probably be an array of either globs or regexps rather than being hardwired to extensions. But I'm 👎 as an over-use of cookbook file distribution. If you want to ship python code in a cookbook, include the wheel and `pip install` it before use.

erikng:
@coderanger pip install is not a valid use case here and is ultimately an even poorer tool with other issues, which only adds complexity.

People who manage macOS and use chef commonly distribute python libraries/other python tools that can create pyc files - a very common use case is PyObc files. These python tool might be sub tools inside a larger tool that invokes these files (thereby creating the pyc files). 

The remote_directory is a perfect use case for installing these files, but ultimately causes chef-runs to update resources that aren't needed.

mikedodge04:
@coderanger I think there is som confusion. They are not talking about python modules, they are talking a few python scripts. let's say a dir called 'mac_tools' with a handful of py files, deployed with remote_dir. With purge set it will manage the state of that dir. but each run if the scripts were ran, it must clean up the '.pyc' each run. Does that help?  

But also I could think of plenty of situations where you would want to manage an entire directory, but exclude one file type, or like you mention, a glob.

coderanger:
`remote_directory` is not intended to replace proper file deployment systems like git, or tarballs, or Mac pkg installers. As such we try to keep `cookbook_file` and `remote_directory` very simple. My personal opinion is this crosses the line of what we are targeting, but will have to see if other maintainers disagree :)

erikng:
@coderanger fair, but package based installers have no idea if the files have changed (either by accident or nefariously). Same for tarballs and unless someone wrote a deletion/git clone script it wouldn't work either.

The files in this case are pretty important and should be tracked by chef, but again the pyc files get in the way.

coderanger:
As the author of the Chef+Python world, I would highly recommend turning your scripts into packages and installing them via normal `pip` mechanism. It's a small amount of work up-front to do the conversion, but worlds easier to manage.

erikng:
Pip doesn't address the files changing.

In the world of servers this makes sense, but when managing user devices that drift every minute of every day, things like this are not only better, but more secure.

lamont-granquist:
yeah i don't particularly like this from the perspective of creeping featurism on the remote_directory resource.

this should probably be a feature on the `zap` cookbook and that cookbook should be made to work with `remote_directory` (although the latest version might?  i suspect that the way that remote_directory doesn't properly use_inline_resources prevents it from working though).

and if instead we're going to go down this route it needs to be engineered to allow arbitrary file specification from the start so that we don't keep adding more properties to solve the same problem in different ways one by one.  only solving the problem for file extensions is limiting and not enough (which is another reason why i think it needs to go into `zap`).

clburlison:
Alright thanks for the feedback. I'll continue using this as a custom provider.

PR #6580, opened 2017-11-09
Bump dependencies to pick up InSpec 1.45.9

adamleff:


tas50:
@adamleff I assume in 20 days there's been 10 new inspec releases so this can get closed out now?

adamleff:
@tas50 hilarious... and yes, I'm going to close it an open new ones since I did a release today! :)

btm:
@tas50 I think we need to find a bunch of inspec bugs to keep @adamleff busy so he stops being so good at shipping inspec.

adamleff:
@btm don't tell anyone, but we're good at shipping because we keep fixing bugs.

Issue #6581, opened 2017-11-11
knife ssh exits on error even when `--exit-on-error` is not passed


sarkis:
## Description

When using `knife ssh` to run a ssh_command on multiple nodes and --exit-on-error is NOT passed, knife ssh will still fail out upon encountering non-zero exit status.

## Chef Version

13.6.4

## Platform Version

Ubuntu 14.04

## Replication Case
Run any command that returns a 0 exit status on a set of nodes.  
  
Run any command that returns a non-zero exit status on the same set of nodes.  
  
This returns a different number of output. It seems that --exit-on-error behavior is being used here, even though it is disabled.

example:
```
$ knife ssh "name:*.jade" "uname -a" | wc -l
17

$ knife ssh "name:*.jade" "uname -a && false" | wc -l
2
```

sarkis:
I've narrowed this issue down to this logic here: https://github.com/chef/chef/blob/93fe1aca2beff64b30910daa2aac615bc623d259/lib/chef/knife/ssh.rb#L338

The issue looks to actually be config[:on_error] actually gets set to `:skip` here: https://github.com/chef/chef/blob/93fe1aca2beff64b30910daa2aac615bc623d259/lib/chef/knife/ssh.rb#L134

I'll be sending a PR that fixes the issue with a stricter conditional that looks for `:raise` rather than just if config[:on_error] is set...

PR #6582, opened 2017-11-11
use a stricter comparison so knife ssh only fails if --exit-on-error

sarkis:
### Description

This is required due to config[:on_error] being set here: https://github.com/chef/chef/blob/93fe1aca2beff64b30910daa2aac615bc623d259/lib/chef/knife/ssh.rb#L134. This conditional will allow ssh_commands that return non-zero status to be executed on entire pool of servers returned by the knife query when --exit-on-error is NOT specified. 

### Issues Resolved

https://github.com/chef/chef/issues/6581

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -448,7 +448,7 @@ class Chef
         ssh.config[:ssh_identity_file] = config[:ssh_identity_file] || config[:identity_file]
         ssh.config[:manual] = true
         ssh.config[:host_key_verify] = config[:host_key_verify]
-        ssh.config[:on_error] = :raise
+        ssh.config[:on_error] = true
         ssh
       end
 
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -335,7 +335,7 @@ class Chef
         command = fixup_sudo(command)
         command.force_encoding("binary") if command.respond_to?(:force_encoding)
         subsession.open_channel do |chan|
-          if config[:on_error] && exit_status != 0
+          if config[:on_error] == :raise && exit_status != 0
             chan.close()
           else
             chan.request_pty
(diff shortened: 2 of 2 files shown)


coderanger:
That whole option seems like something is very wrong. It's declared as a boolean type but in some places we expect a symbol. And the `proc` on it doesn't actually check the `key` value which means it doesn't actually matter if you set the option or not.

sarkis:
@coderanger thanks for taking a look - 
I think the way it's meant to be used where --exit-on-error is set (true) or not (false) - we should keep this as a boolean. I simplified this a bit - not sure the `proc` is necessary as you mentioned in the previous comment. Let me know what you think of this approach when you have a moment.

sarkis:
UPDATE - I was able to successfully test all edge cases I could think of here - including testing ssh connection issues to one of the servers returned in the knife ssh query.

lamont-granquist:
were we trying too hard to be backcompat at some point and only half converted it from boolean to a multi-valued symbol?

stevendanna:
@lamont-granquist The other way around.  It started as multi-value config that was really only used when calling this plugin from other plugins like this:

https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L451

I'm in favor of moving it to a boolean if we don't think anyone else is using it.  If we do move it to a boolean, lets fixup the usage in bootstrap above as well.

sarkis:
Since this was set to `:raise` in bootstrap - i wanted to leave the same behavior which after moving to boolean would be the same as set to true. Hopefully, I didn't misinterpret this ... as far as I can tell that covers all instances of `ssh.confg[:on_error]` in this repo - let me know if a PR may be required elsewhere...

mal:
Constantly running into this and having to remember to add `|| true` to my commands 🙁 Tested out this patch and can confirm it fixes the basic `knife ssh` use case. Thanks for the patch, it'd be really good to see it in the next stable 👍

thommay:
@sarkis thanks for your contribution!

PR #6584, opened 2017-11-15
Add ohai_time to minimal_ohai filter

btm:
We need ohai_time to get the 'Last Check-In' time on the Chef Server.
There is really no risk to including it.


btm:
breaking CI is unrelated :(

PR #6588, opened 2017-11-16
Revert "add missing functional tests for users"

lamont-granquist:
testing to see if appveyor goes green or not.

so somewhere between v13.5.28 and v13.6.7 we broke appveyor.

looks like it was fa6e4e89b586bb8eaee767d5602b13cb75f6c073


lamont-granquist:
Yep, sounds good to me.

PR #6590, opened 2017-11-17
Knife SSH prefix option

### LINKED ISSUE #6573: Req: Option to prefix with node names rather than SSH attribute in knife ssh
## Description

The output of `knife ssh` is really useful when running a command over multiple servers, but it can be frustrating to have to convert public hostnames, or IPs back to a recognisable server name. It'd be nice to have a flag to switch out the SSH attribute value with the corresponding node name instead for these situations and a matching `knife.rb` setting to facilitate setting it as default behavior.

Possible extensions might be allowing an arbitrary attribute to be used, configurable in `knife.rb`.

## Chef Version

13.4.19

## Platform Version

Linux

## Replication Case

**Current:**
```
$ knife ssh 'name:test-server' 'echo hello'
ec2-xxx-xxx-xxx-xxx.compute-1.amazonaws.com hello
```

**Desired:**
```
$ knife ssh -n 'name:test-server' 'echo hello'
test-server hello
```

mal:
### Description

```sh
# existing usage (no change)
$ knife ssh 'name:app-server' 'echo hello' 
ec2-10-0-0-1.compute-1.amazonaws.com hello

# new functionality
$ knife ssh --prefix-attribute name 'name:app-server' 'echo hello'
app-server hello
```

Allows an attribute to be specified which will determine how a node is identified in the output of `knife ssh` commands. It uses the long flag `--prefix-attribute` ~~and the short flag `-b`. The short flag was picked as a placeholder, chosen because `-p` is already in use for `--port`, this can be changed prior to merging to align with any sort of option/flag guidelines that may exist.~~ The config setting `Chef::Config[:knife][:prefix_attribute]` is also supported.

This feature makes use of the session properties assignment functionality documented [here](https://github.com/jamis/net-ssh-multi/blob/c0b0a0b9a84d0e2ba58dd6f78e0d5e28900e18c4/lib/net/ssh/multi/session.rb#L320) to attach the prefix to the session, and renames some internal keys (`config`, `knife_config`) in order to aid understanding of the code when reading and provide clear distinctions between the attribute keys used for the SSH target and the prefix.

### Issues Resolved

Resolves #6573

---

~~Tests seem to have been broken before I got here 🙁~~ Looks like it's been fixed \o/

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -46,11 +46,10 @@ class Chef
         :default => nil,
         :proc => lambda { |o| o.to_i }
 
-      option :attribute,
+      option :ssh_attribute,
         :short => "-a ATTR",
         :long => "--attribute ATTR",
-        :description => "The attribute to use for opening the connection - default depends on the context",
-        :proc => Proc.new { |key| Chef::Config[:knife][:ssh_attribute] = key.strip }
+        :description => "The attribute to use for opening the connection - default depends on the context"
 
       option :manual,
         :short => "-m",
@@ -181,27 +180,21 @@ class Chef
         session_from_list(list)
       end
 
-      def get_ssh_attribute(node)
+      def get_ssh_attribute(item)
         # Order of precedence for ssh target
-        # 1) command line attribute
-        # 2) configuration file
-        # 3) cloud attribute
-        # 4) fqdn
-        if node["config"]
-          Chef::Log.debug("Using node attribute '#{config[:attribute]}' as the ssh target: #{node["config"]}")
-          node["config"]
-        elsif Chef::Config[:knife][:ssh_attribute]
-          Chef::Log.debug("Using node attribute #{Chef::Config[:knife][:ssh_attribute]}: #{node["knife_config"]}")
-          node["knife_config"]
-        elsif node["cloud"] &&
-            node["cloud"]["public_hostname"] &&
-            !node["cloud"]["public_hostname"].empty?
-          Chef::Log.debug("Using node attribute 'cloud[:public_hostname]' automatically as the ssh target: #{node["cloud"]["public_hostname"]}")
-          node["cloud"]["public_hostname"]
+        # 1) config value (cli or knife config)
+        # 2) cloud attribute
+        # 3) fqdn
+        msg = "Using node attribute '%s' as the ssh target: %s"
+        if item["target"]
+          Chef::Log.debug(sprintf(msg, config[:ssh_attribute], item["target"]))
+          item["target"]
+        elsif !item.dig("cloud", "public_hostname").to_s.empty?
+          Chef::Log.debug(sprintf(msg, "cloud.public_hostname", item["cloud"]["public_hostname"]))
+          item["cloud"]["public_hostname"]
         else
-          # falling back to default of fqdn
-          Chef::Log.debug("Using node attribute 'fqdn' as the ssh target: #{node["fqdn"]}")
-          node["fqdn"]
+          Chef::Log.debug(sprintf(msg, "fqdn", item["fqdn"]))
+          item["fqdn"]
         end
       end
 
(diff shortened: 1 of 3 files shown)


mal:
Test failures appear to be related to file permissions in Travis, and were already broken in appveyor rather than an of the code changes here :/

mal:
Tests seem pretty broken before I got here 🙁

mal:
Rebased on master to get passing tests \o/

lib/chef/knife/ssh.rb
```diff
@@ -59,6 +59,12 @@ class Ssh < Knife
         :description => "QUERY is a space separated list of servers",
         :default => false
 
+      option :prefix,
+        :short => "-b ATTR",
+        :long => "--prefix-attribute ATTR",
+        :description => "The attribute to use for prefixing the ouput - default depends on the context",
+        :proc => Proc.new { |key| Chef::Config[:knife][:prefix_attribute] = key.strip }
```
lamont-granquist: I'm reasonably certain that if you name your option `:prefix_attribute` and delete this proc that it merges correctly now (and has for a very long time, but we have never cleaned up this from the knife attributes), and in particular you should make sure that the Chef::Config option and the Mixlib::Cli option here match.

lib/chef/knife/ssh.rb
```diff
@@ -181,18 +187,34 @@ def configure_session
...
 
+      def get_prefix_attribute(node)
+        # Order of precedence for prefix
+        # 1) command line attribute
+        # 2) configuration file
+        # 3) nil
+        if node["prefix_cli"]
+          Chef::Log.debug("Using node attribute from CLI '#{config[:prefix]}' as the prefix: #{node["prefix_cli"]}")
+          node["prefix_cli"]
+        elsif Chef::Config[:knife][:prefix_attribute]
```
lamont-granquist: Also here you should use `config[:prefix_attribute]` to get the merging magic.

lib/chef/knife/ssh.rb
```diff
@@ -212,14 +234,24 @@ def search_nodes
...
 
+        # if we've set a prefix attribute to use on the command line
+        if config[:prefix]
+          required_attributes[:prefix_cli] = config[:prefix].split(separator)
+        end
+
+        # if we've configured a prefix attribute in our config
+        if Chef::Config[:knife][:prefix_attribute]
+          required_attributes[:prefix_config] = Chef::Config[:knife][:prefix_attribute].split(separator)
+        end
```
lamont-granquist: again this should just be `config[:prefix_attribute]` and let the knife cli/config merging code handle it.

lib/chef/knife/ssh.rb
```diff
@@ -181,18 +187,34 @@ def configure_session
...
-          Chef::Log.debug("Using node attribute '#{config[:attribute]}' as the ssh target: #{node["config"]}")
-          node["config"]
+        if node["ssh_cli"]
+          Chef::Log.debug("Using node attribute from CLI '#{config[:attribute]}' as the ssh target: #{node["ssh_cli"]}")
+          node["ssh_cli"]
         elsif Chef::Config[:knife][:ssh_attribute]
-          Chef::Log.debug("Using node attribute #{Chef::Config[:knife][:ssh_attribute]}: #{node["knife_config"]}")
-          node["knife_config"]
+          Chef::Log.debug("Using node attribute from config '#{Chef::Config[:knife][:ssh_attribute]}' as the ssh target: #{node["ssh_config"]}")
+          node["ssh_config"]
```
lamont-granquist: this looks like it changed an external API on the node?

coderanger:
This doesn't seem common enough to use a short option too, can we just do the long option?

lib/chef/knife/ssh.rb
mal: `node` in this context is the response from the custom `query.search` call we make in `search_nodes`.

mal:
All the changes here are based on mirroring the current approach taken with the ssh attribute option. I'm not at all familiar with the "merging magic" stuff mentioned, so maybe that's a module level refactor worth doing as a separate body of work?

Dropping the short flag seems like a good idea, I'll sort that out in a bit 👍

lamont-granquist:
yes, the existing code may just be broken, and may be more complicated than it needs to be, please don't copy that.

if you just name the cli option and the knife config option the same and access them through the `config` hash it works correctly and the `Proc` in the cli option can be dropped.  anything else and you're copypasta'ing dubious code around.

mal:
- Sorted out the newer config merge magic for both `:ssh_attribute` and `:prefix_attribute`
- Renamed `node` in the `get_*_attribute` methods so it's clear that they're not operating on real nodes
- Removed the short flag in favour of just having the long flag

mal:
Flaky tests are back, but not related to this PR 🙁

lamont-granquist:
coolness, merging...

PR #6594, opened 2017-11-20
DO NOT MERGE: bunch of debugging

btm:
Trying to see what's happening with the user context test on appveyor.


btm:
The issues with the particular tests breaking other tests on appveyor are still unknown. There were other underlying issues so we reverted those tests for now in https://github.com/chef/chef/pull/6588.

PR #6597, opened 2017-11-22
Don't try to uninstall bundler on appveyor

btm:
As of Rubygems 2.7 bundler is a default gem and cannot be uninstalled.

Appveyor recently (11/19) changed and now has bundler as a default gem.

https://github.com/appveyor/ci/issues/1944


appveyor.yml
```diff
@@ -29,9 +29,8 @@ install:
   - ruby --version
   - gem --version
   - bundler --version
-  - gem uninstall bundler -a -x
   - gem update --system %OMNIBUS_RUBYGEMS% || gem update --system %OMNIBUS_RUBYGEMS% || gem update --system %OMNIBUS_RUBYGEMS%
-  - gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc || gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc || gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc
+  - gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc --force || gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc --force || gem install bundler -v %OMNIBUS_BUNDLER% --quiet --no-ri --no-rdoc --force
```
thommay: I don't think we need to do this any more. Just leave the system bundler in place - appveyor is building it in now.

btm: We're still installing a specific version of bundler on Travis. I figured the point here is to always test with the version of bundler in `omnibus_overrides.rb`. This line does successfully give us 1.15.4 rather than the 1.16.0 that Appveyor is shipping.

btm: If we don't care what version of bundler is in use, then yah sure.

thommay: i think we should always ship the latest (and we should do the same in omnibus, eventually). Ruby 2.5 will take this out of our hands anyway.

btm: 🚅

thommay:
tres bon!

btm:
I don't know what travis is doing, but appveyor is green. Progress!

btm:
Fixes appveyor, but Travis is having some permission issues with ` /home/travis/`. I'm seeing the same thing on https://github.com/chef/chef/pull/6598, I think it's a new infrastructure problem.

PR #6600, opened 2017-11-22
Switch from the Travis container to the VM

btm:
If `sudo` is `false`, then Travis CI will use a container for testing. If
    `sudo` is `required, then it will use a GCE VM. We've recently been
    seeing permission errors that appeared related to the Travis CI infrastructure.
    Switching to the VM resolves those errors. The downside is that it takes 50s
    instead of 5s to start up the tester, but given that our tests take 18 minutes
    that is okay.

    https://docs.travis-ci.com/user/reference/trusty/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-sudo: false
+sudo: required
 cache: bundler
 dist: trusty
 


lamont-granquist:
green >> speed

btm:
This did pass CI before I pushed with a cleaned up commit message: https://travis-ci.org/chef/chef/builds/305996472

btm:
Travis is green. Appveyor was previously green, I'm not going to wait another 20 minutes to see it happen 3 times in a row. 🤠

PR #6602, opened 2017-11-22
Switch from the Travis container to the VM on chef-12

btm:
If `sudo` is `false`, then Travis CI will use a container for testing. If
`sudo` is `required, then it will use a GCE VM. We've recently been
seeing permission errors that appeared related to the Travis CI infrastructure.
Switching to the VM resolves those errors. The downside is that it takes 50s
instead of 5s to start up the tester, but given that our tests take 18 minutes
that is okay.

https://docs.travis-ci.com/user/reference/trusty/


tas50:
After more investigation on master it appears that Travis nodes 1 & 2 will always fail with weird permissions errors for us. There appears to be 6 nodes, but 1&2 run more often than the remaining 4 so we see a lot of failures. Travis is aware of this now and once they come back from Christmas hopefully they'll get to the bottom of it.

PR #6604, opened 2017-11-27
Remove test stuff from the changelog

tas50:
No need to add these to the changelog. Also fixup the wording of the
windows_task change since that's super important and people should know
about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,12 @@
 
 #### Merged Pull Requests
 - Makes life easier for hook authors switching from the older report handler syntax [#6574](https://github.com/chef/chef/pull/6574) ([coderanger](https://github.com/coderanger)) <!-- 13.6.16 -->
-- [MSYS-688] Fixed invalid date and Invalid starttime error [#6544](https://github.com/chef/chef/pull/6544) ([NimishaS](https://github.com/NimishaS)) <!-- 13.6.15 -->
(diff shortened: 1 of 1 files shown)


lamont-granquist:
I'm kind of philosophically opposed to manually hacking up the ChangeLog and think we need to get back to PR titles being the source of truth and being able to regen the ChangeLog (completely) from information in github.   If we just had a log of which version numbers were promoted releases we would have all the information necessary to do that.

thommay:
@lamont-granquist 🙈 we went away from that because it was such a total disaster. I think hacking up the changelog after the fact in general is not worth the time, but I definitely do not want us to change our current approach.

tas50:
In the ideal world we'd just get the names right the first time and use the nice new labels to exclude appveyor and travis changes from the changelog. I updated this one primarily because the windows_task change is super significant and we have users waiting on that. The title of the PR was cryptic and not something you could search for. I try to tag things for exclusion and fixup the names as I see them, but it's a best effort sort of thing.

lamont-granquist:
Driving the changelog entirely off the issues and PR titles worked great up until we went down the crazy CI version numbering route.   All we'd need to get back to that was a list of the promoted versions (which is 98% captured in rubygems.org) and we
could teach the changelog generator to construct changelogs entirely from the issue contents and that version log.  Where it all went horribly wrong is with the badly non-declarative approaches we tried to use to incrementally generate the changelog and mangled the crap out of it.  Before that it was working pretty well.

PR #6608, opened 2017-11-28
fix NodeMap to not throw exceptions on platform_versions

lamont-granquist:
Chef::Version and VersionConstraint are for x.y/x.y.z cookbook
versions, and are always wrong for anything other than
cookbooks.

### DIFF
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -143,7 +143,7 @@ class Chef
 
       filter_values.empty? ||
         Array(filter_values).any? do |v|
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Yeah, I'm not claiming this is correct in all circumstances for all platforms (clearly) but it doesn't throw on non-x.y.z, which unblocks significant customers and was just hella broken.

jaymzh:
@lamont-granquist  I'd merge, but yo tests iz br0k3d.

lamont-granquist:
transient failure

PR #6611, opened 2017-12-01
Add Amazon Linux testing to PRs in Travis

tas50:
Our biggest limiter in tests are the unit tests which take 20+ minutes to run. This gives us the ability to add additional integration tests w/o impacting the overall time it takes to test a build. We claim to support Amazon Linux so we should start testing PRs against it. Worst case it finds nothing. Best case we avoid a regression.


lamont-granquist:
you should nuke the old commented-out section in the .travis.yml as well.

PR #6612, opened 2017-12-01
[MSYS-726] Allow setting environment variables at the user level

harikesh-kolekar:
### Description
The `env` resource allows you to set a system environment variable, but windows also allow per-user environment variables.

https://blogs.technet.microsoft.com/heyscriptingguy/2005/03/18/how-can-i-create-an-environment-variable-using-a-script/

It looks like we could simply add a user property and pass it in obj.username in the existing provider instead of "<System>".

Acceptance criteria:
- User property for env resource

### Check List

- [x] All tests pass


coderanger:
This is not part of the abstract interface for "global env vars" so we are probably at (really: long past, but whatevs) the point that this resource needs to be renamed `windows_path` or this property should be moved into a platform-specific resource :-/ If we want to keep it in the generic one, we should make sure all other providers raise an exception if it is used so it doesn't silently Do The Wrong Thing.

btm:
>  this property should be moved into a platform-specific resource :-/ If we want to keep it in the generic one, we should make sure all other providers raise an exception if it is used so it doesn't silently Do The Wrong Thing.

What would a 'platform-specific resource' be? Just renaming it `windows_env`? The resource is currently [limited to the windows platform here](https://github.com/chef/chef/blob/master/lib/chef/resource/env.rb#L23) and thus we do raise exceptions elsewhere, e.g.:
```
2017-12-05T12:08:08-05:00] FATAL: Chef::Exceptions::NoSuchResourceType: Cannot find a resource for env on mac_os_x version 10.11.6
```

coderanger:
Yes, `windows_env` would be a better name if this is going to be a specific model for Windows environment variables.

spec/functional/resource/env_spec.rb
```diff
@@ -55,6 +79,25 @@
         expect(ENV[chef_env_test_lower_case]).to eq(env_value1)
       end
 
+      it "should create an environment variable with default user System for action create" do
+        expect(ENV[chef_env_test_lower_case]).to eq(nil)
+        test_resource.key_name(chef_env_test_lower_case)
+        test_resource.value(env_value1)
+        test_resource.run_action(:create)
+        env_obj
+        expect(@env_obj.username.upcase).to eq(default_env_user)
```
NimishaS: Can we assign the return value of `env_obj` to some variable instead of using the instance variable `@env_obj`?

spec/functional/resource/env_spec.rb
```diff
@@ -22,9 +22,27 @@
...
+    let(:env_obj) do
+      wmi = WmiLite::Wmi.new
+      environment_variables = wmi.query("select * from Win32_Environment where name = '#{test_resource.key_name}'")
+      if environment_variables && environment_variables.length > 0
+        environment_variables.each do |env|
+          @env_obj = env.wmi_ole_object
+          return @env_obj if @env_obj.username.split('\\').last.casecmp(test_resource.user) == 0
+        end
+      end
+      @env_obj = nil
```
NimishaS: Please return a value from this method instead of setting the instance variable.

spec/functional/resource/env_spec.rb
```diff
@@ -65,6 +108,20 @@
...
+        test_resource.value(env_value1)
+        test_resource.run_action(:create)
+        expect(ENV[chef_env_test_lower_case]).to eq(env_value1)
+        test_resource.value(env_value2)
+        test_resource.user(env_user)
+        test_resource.run_action(:create)
+        test_resource.key_name(chef_env_test_lower_case)
+        test_resource.user(default_env_user)
+        env_obj
+        expect(@env_obj.variablevalue).to eq(env_value1)
```
NimishaS: We can probably use a `before` block here for creating the existing variable.

spec/functional/resource/env_spec.rb
```diff
@@ -161,6 +218,53 @@
...
+        test_resource.value(env_value1)
+        test_resource.user(env_user)
+        test_resource.run_action(:create)
+        expect(ENV[chef_env_test_lower_case]).to eq(env_value1)
+        test_resource.run_action(:delete)
+        env_obj
+        expect(@env_obj).to eq(nil)
+      end
+
+      it "should not delete an user environment variable if user are not passed" do
```
NimishaS: should not delete an user environment variable if user `is` not passed.

spec/functional/resource/env_spec.rb
```diff
@@ -161,6 +218,53 @@
...
+        test_resource.key_name(chef_env_test_lower_case)
+        test_resource.value(env_value1)
+        test_resource.user(env_user)
+        test_resource.run_action(:create)
+        expect(ENV[chef_env_test_lower_case]).to eq(env_value1)
+        test_resource.user(default_env_user)
+        test_resource.run_action(:delete)
+        test_resource.user(env_user)
+        env_obj
+        expect(@env_obj).not_to be_nil
```
NimishaS: Please use `before` block or `let` for creating the existing environment variable.

spec/unit/provider/env_spec.rb
```diff
@@ -47,12 +48,17 @@
...
+      expect(@provider.new_resource.user).to eq("<System>")
+    end
+
     it "should set the key_name to the key name of the new resource" do
       @provider.load_current_resource
       expect(@provider.current_resource.key_name).to eq("FOO")
     end
 
-    it "should check if the key_name exists" do
+    it "should check if the key_name and user exists" do
```
NimishaS: "should check if the key_name and user `exist`"

spec/unit/provider/env_spec.rb
```diff
@@ -152,14 +158,20 @@
       @provider.action_modify
     end
 
+    it "should call modify_group if the key exists and user are not equal" do
```
NimishaS: "and `users` are not equal"

lib/chef/resources.rb
```diff
@@ -35,7 +35,7 @@
 require "chef/resource/dnf_package"
 require "chef/resource/dsc_script"
 require "chef/resource/dsc_resource"
-require "chef/resource/env"
+require "chef/resource/windows_env"
```
tas50: this should go in the right place alphabetically.

lib/chef/resource/windows_env.rb
```diff
@@ -21,16 +21,17 @@ class Chef
   class Resource
     # Use the env resource to manage environment keys in Microsoft Windows. After an environment key is set, Microsoft
     # Windows must be restarted before the environment key will be available to the Task Scheduler.
-    class Env < Chef::Resource
-      resource_name :env
-      provides :env, os: "windows"
+    class WindowsEnv < Chef::Resource
+      resource_name :windows_env
+      provides :windows_env, os: "windows"
```
tas50: We want to make sure to also include  provides :env here so that any windows users with previously cookbooks will continue to work.

lamont-granquist:
after some poking around i am satisfied that we never had a unix env resource before.  on unix previously it blew up with NoSuchResourceType because the resource was windows-only, then the provider that was wired up on unix was a stub that raised notimplementederror.  so busted all around, nobody could have ever used it.

want the back compat though so people on windows that were using 'env' can continue using 'env'.

thommay:
Right, so this is going in the right direction, and clearly on non-Windows this has raised out of the box so there's no concern about deprecation there.
What we should do:

the resource should add a `provides :env os: "windows"`.
There's no real need for the superclass any more, so just merge `providers/env/windows.rb` into `providers/env.rb` and then rename that to `providers/windows_env.rb`.
Delete the `providers/env/` directory.

spec/functional/resource/windows_env_spec.rb
```diff
@@ -0,0 +1,285 @@
+#
```
wrightp: I recommend doing a `git mv` here and then make the appropriate changes.

lib/chef/provider/windows_env.rb
```diff
@@ -17,14 +17,16 @@
...
 
 class Chef
   class Provider
-    class Env < Chef::Provider
+    class WindowsEnv < Chef::Provider
+      include Chef::Mixin::WindowsEnvHelper
       attr_accessor :key_exists
 
-      provides :env, os: "!windows"
+      provides :windows_env, os: "windows"
```
thommay: can you add
```
provides :env, os: "windows"
```
here too please?

lib/chef/providers.rb
```diff
@@ -28,7 +28,7 @@
 require "chef/provider/directory"
 require "chef/provider/dsc_script"
 require "chef/provider/dsc_resource"
-require "chef/provider/env"
+require "chef/provider/windows_env"
```
thommay: this file should be sorted alphabetically.

spec/functional/resource/windows_env_spec.rb
harikesh-kolekar: I have tried by `git mv ` but still, in git file changes it shows file has been deleted you can see commit at https://github.com/chef/chef/pull/6612/commits/85ad3ee585ab50d173c083325d106967d2161dab#diff-5903ffd7bafff40eae05e9c21380c6a0

lamont-granquist:
@thommay merge if you think its good

PR #6625, opened 2017-12-06
speedup lwrp tests

lamont-granquist:
reset the global resource and provider handler maps before each and
every test.

note that mutation through the top-level keys will get preserved so
this is not perfect, but deep-duping through every single test we run
becomes expensive.

this is sufficient to flush out the global state of all the test lwrp fixtures
that we setup.

it discovered several usage bugs where we either relied on this feature
or where we were testing different fixtures from the ones we thought we
were.

### DIFF
--- a/spec/integration/recipes/noop_resource_spec.rb
+++ b/spec/integration/recipes/noop_resource_spec.rb
@@ -4,7 +4,7 @@ describe "Resources with a no-op provider" do
   include IntegrationSupport
 
   context "with noop provider providing foo" do
-    before(:context) do
+    before(:each) do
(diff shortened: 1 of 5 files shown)


lamont-granquist:
So first indications are that this speeds up travis by about 2x (38/40 mins down to 19 mins)

lamont-granquist:
and looks like appveyor is coming down from like 58 mins down to 35/37ish which is 40% perf boost

lamont-granquist:
i think this should now eventually go green, modulo sunspots and DCO bot seems stuck

PR #6627, opened 2017-12-07
[MSYS-725] [chef#6073] allow forcing sensitive to false on execute and batch resource if password is set.

### LINKED ISSUE #6073: Sensitive property does not work on execute or batch resource
## Description

When using the execute or patch resource, if the password property is used, the Chef Client treats the output as sensitive, even if you set the sensitive property to false.

## Chef Version

12.19.36 x64

## Platform Version

Windows Server 2012 R2.

## Replication Case

You need to have a password stored in an encrypted data bag (or I suppose you could just hardcode in the password, but I used a data bag), and then try this in a recipe:

>batch 'run ipconfig' do
>	code 'ipconfig /null'
>	user tauser['id']
>	password tauser['pw']
>end

This will fail but the output will be suppressed.  Modifying the code to this doesn't change the output:

>batch 'run ipconfig' do
>      sensitive false
>	code 'ipconfig /null'
>	user tauser['id']
>	password tauser['pw']
>end

Vasu1105:
Allows forcing sensitive to false on execute and batch resource if password is set.

Resolves #6073 

### DIFF
--- a/lib/chef/resource/execute.rb
+++ b/lib/chef/resource/execute.rb
@@ -133,7 +133,8 @@ class Chef
 
       property :password, String, sensitive: true
 
-      property :sensitive, [ TrueClass, FalseClass ], default: false, coerce: proc { |x| password ? true : x }
+      # lazy used to set default value of sensitive to true if password is set
+      property :sensitive, [ TrueClass, FalseClass ], default: lazy { |r| r.password ? true : false }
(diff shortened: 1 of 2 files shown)


Vasu1105:
@btm and @NimishaS travis and appveyor failures are not related to my changes.

PR #6628, opened 2017-12-07
windows_task: Add additional input validation to properties

tas50:
windows_task is rather particular on the data it takes, especially dates and time values. We need to make sure to properly validate user input so we can fail and warn.

Changes in this PR:

- execution_time_limit now sets the default value in the property definition instead of injecting it via a validation method later on
- random_delay and execution_time_limit now accept Integers in addition to Strings. The value here is an Integer and this is the safest way to pass it
- random_delay and execution_time_limit validation has been improved to make sure the values are in fact numeric
- start_time is now properly validated to make sure it's a time value
- start_date is now properly validated to make sure it's a date value
- Generation of ISO8601 duration can now accept 0 seconds as a valid duration
- Testing added to the generation of ISO8601 duration strings from the seconds value the user passes
- All validation errors now ArgumentErrors not RuntimeErrors

### DIFF
--- a/lib/chef/resource/windows_task.rb
+++ b/lib/chef/resource/windows_task.rb
@@ -72,7 +72,7 @@ class Chef
 
         if execution_time_limit
           unless execution_time_limit == "PT72H" # don't double convert an iso08601 format duration
-            raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds an Integer or a String with numeric values only e.g. '60'." unless numeric_value_in_string?(execution_time_limit)
+            raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds as an Integer (e.g. 60) or a String with numeric values only (e.g. '60')." unless numeric_value_in_string?(execution_time_limit)
             duration = sec_to_dur(execution_time_limit)
             execution_time_limit(duration)
           end
@@ -101,7 +101,7 @@ class Chef
           raise ArgumentError, "`random_delay` property is supported only for frequency :minute, :hourly, :daily, :weekly and :monthly"
         end
 
-        raise ArgumentError, "Invalid value passed for `random_delay`. Please pass seconds an Integer or a String with numeric values only e.g. '60'." unless numeric_value_in_string?(random_delay)
+        raise ArgumentError, "Invalid value passed for `random_delay`. Please pass seconds as an Integer (e.g. 60) or a String with numeric values only (e.g. '60')." unless numeric_value_in_string?(random_delay)
       end
 
       def validate_start_day(start_day, frequency)
--- a/lib/chef/resource/windows_task.rb
+++ b/lib/chef/resource/windows_task.rb
@@ -117,7 +117,7 @@ class Chef
       def validate_start_time(start_time, frequency)
         if start_time
           raise ArgumentError, "`start_time` property is not supported with `frequency :none`" if frequency == :none
-          raise ArgumentError, "`start_time` property must be in the HH:mm format." unless /^[0-2][0-3]:[0-5][0-9]$/.match?(start_time)
+          raise ArgumentError, "`start_time` property must be in the HH:mm format (e.g. 6:20pm -> 18:20)." unless /^[0-2][0-9]:[0-5][0-9]$/.match?(start_time)
         else
           raise ArgumentError, "`start_time` needs to be provided with `frequency :once`" if frequency == :once
         end
--- a/lib/chef/resource/windows_task.rb
+++ b/lib/chef/resource/windows_task.rb
@@ -44,22 +44,22 @@ class Chef
       property :interactive_enabled, [TrueClass, FalseClass], default: false
       property :frequency_modifier, [Integer, String], default: 1
       property :frequency, equal_to: [:minute,
-                                       :hourly,
-                                       :daily,
-                                       :weekly,
-                                       :monthly,
-                                       :once,
-                                       :on_logon,
-                                       :onstart,
-                                       :on_idle,
-                                       :none], default: :hourly
+                                      :hourly,
+                                      :daily,
+                                      :weekly,
+                                      :monthly,
+                                      :once,
+                                      :on_logon,
+                                      :onstart,
+                                      :on_idle,
+                                      :none], default: :hourly
       property :start_day, String
       property :start_time, String
       property :day, [String, Integer]
       property :months, String
       property :idle_time, Integer
-      property :random_delay, String
-      property :execution_time_limit, String
+      property :random_delay, [String, Integer]
+      property :execution_time_limit, [String, Integer], default: "PT72H"
 
       attr_accessor :exists, :status, :enabled
 
@@ -71,15 +71,11 @@ class Chef
         end
 
         if execution_time_limit
-          raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds as a String e.g. '60'." if execution_time_limit.to_i == 0
-          duration = sec_to_dur(execution_time_limit)
-          execution_time_limit(duration)
-        else
-          # schtask sets execution_time_limit as PT72H by default
-          # We are setting the default value here so that we can do idempotency check later
-          # Note: We can't use `default` in the property
-          # because it will raise error for Invalid values passed as "PT72H" is not in seconds
-          execution_time_limit("PT72H")
+          unless execution_time_limit == "PT72H" # don't double convert an iso08601 format duration
+            raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds an Integer or a String with numeric values only e.g. '60'." unless numeric_value_in_string?(execution_time_limit)
+            duration = sec_to_dur(execution_time_limit)
+            execution_time_limit(duration)
+          end
         end
 
         validate_start_time(start_time, frequency)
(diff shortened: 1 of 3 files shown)


lib/chef/resource/windows_task.rb
```diff
@@ -44,22 +44,22 @@ def initialize(name, run_context = nil)
...
+                                      :none], default: :hourly
       property :start_day, String
       property :start_time, String
       property :day, [String, Integer]
       property :months, String
       property :idle_time, Integer
-      property :random_delay, String
-      property :execution_time_limit, String
+      property :random_delay, [String, Integer]
+      property :execution_time_limit, [String, Integer], default: "PT72H"
```
thommay: comment on what a `PT72H` is when it's at home might be nice (or a link to some docs)

lib/chef/resource/windows_task.rb
```diff
@@ -94,12 +90,18 @@ def after_created
...
+        true
+      end
+
       def validate_random_delay(random_delay, frequency)
         if [:once, :on_logon, :onstart, :on_idle, :none].include? frequency
           raise ArgumentError, "`random_delay` property is supported only for frequency :minute, :hourly, :daily, :weekly and :monthly"
         end
 
-        raise ArgumentError, "Invalid value passed for `random_delay`. Please pass seconds as a String e.g. '60'." if random_delay.to_i == 0
+        raise ArgumentError, "Invalid value passed for `random_delay`. Please pass seconds an Integer or a String with numeric values only e.g. '60'." unless numeric_value_in_string?(random_delay)
```
thommay: I think an `as` has gone missing in the text here.

lib/chef/resource/windows_task.rb
```diff
@@ -94,12 +90,18 @@ def after_created
 
       private
 
+      # Validate the passed value is numeric values only if it is a string
+      def numeric_value_in_string?(val)
+        return false if val.is_a?(String) && /\D/ =~ val # \D is any non-numeric value
```
thommay: `!!Integer(val) rescue false` (can we convert this thing automatically in to an Integer?)

lib/chef/resource/windows_task.rb
thommay: Ah, I see from Google it's an ISO8601 duration. It would be good to document that that's what we're expecting here.

lib/chef/resource/windows_task.rb
tas50: irb(main):002:0> !!Integer(5,000)
=> true
irb(main):003:0> Integer(5,000)
=> 5

This is the problem I was hoping to avoid here. We never should have allowed a String here. Numeric strings are just pain all over

lib/chef/resource/windows_task.rb
tas50: I'll add a comment for that. The original block that injected the value had a comment

lib/chef/resource/windows_task.rb
thommay: But you couldn't pass `5,000` as a bare word anyway:
```
irb(main):004:0> def foo(a); puts a; end
=> :foo
irb(main):005:0> foo 5,000
ArgumentError: wrong number of arguments (given 2, expected 1)
```

tas50: So here's the problem as it expresses itself in 13.6.4:

```ruby
windows_task 'chef-client' do
  user 'SYSTEM'
  command 'chef-client'
  run_level :highest
  frequency :minute
  frequency_modifier 15
  execution_time_limit '5000'
end
```
The execution_time_limit is properly parsed as being 5000 seconds
correct

```ruby
windows_task 'chef-client' do
  user 'SYSTEM'
  command 'chef-client'
  run_level :highest
  frequency :minute
  frequency_modifier 15
  execution_time_limit '5,000'
end
```

The  execution_time_limit is improperly parsed as being 5 seconds.
incorrect

The real life lesson here is don't set a property to String when you really want an Integer, but that 🚢 is long out of port.

thommay: Yup:
```
irb(main):001:0> Integer('5,000')
ArgumentError: invalid value for Integer(): "5,000"
```

thommay: What we see here, really, is that the conversion in `sec_to_dur` on https://github.com/chef/chef/pull/6628/files#diff-7158a2a9329d172ac577a1b48be4ff66R214 isn't right - it just blindy does `to_i`, which tries too hard. If that line were `Integer(seconds)` we'd properly blow up down there and avoid all of this.

lib/chef/resource/windows_task.rb
```diff
@@ -94,25 +90,35 @@ def after_created
...
+        raise ArgumentError, "Invalid value passed for `random_delay`. Please pass seconds as an Integer (e.g. 60) or a String with numeric values only (e.g. '60')." unless numeric_value_in_string?(random_delay)
       end
 
       def validate_start_day(start_day, frequency)
         if [:once, :on_logon, :onstart, :on_idle, :none].include? frequency
           raise ArgumentError, "`start_day` property is not supported with frequency: #{frequency}"
         end
+
+        # make sure the start_day is in MM/DD/YYYY format: http://rubular.com/r/cgjHemtWl5
+        raise ArgumentError, "`start_day` property must be in the MM/DD/YYYY format." unless /^(0[1-9]|1[012])[- \/.](0[1-9]|[12][0-9]|3[01])[- \/.](19|20)\d\d$/.match?(start_day)
```
stuartpreston: I believe this only works for US locale machines. The `start_day` property is a string, which I believe means the date would be passed to `schtasks.exe` wholesale on the command line.  On my UK english locale system this would cause schtasks.exe to fail with an error, as `schtasks.exe` follows the local system locale for command line input.  See the output on my system for an example:

```
C:\projects\other\windows_task_testing> SCHTASKS /Create /SC HOURLY /TN accountant /TR calc.exe /ST 12:00 /ET 14:00 /SD **13/12/2017** /ED 14/12/2017
SUCCESS: The scheduled task "accountant" has successfully been created.
C:\projects\other\windows_task_testing> SCHTASKS /Create /SC HOURLY /TN accountant-us /TR calc.exe /ST 12:00 /ET 14:00 /SD **12/13/2017** /ED 12/13/2017
ERROR: Incorrect Start Date.
```

As an aside, I see we don't have corresponding `end_date` and `end_time` properties (Task expiry time in Task Scheduler).

spec/unit/resource/windows_task_spec.rb
```diff
@@ -125,12 +154,41 @@
...
-      expect { resource.send(:validate_start_day, "Wed", :on_logon) }.to raise_error(Chef::Exceptions::ArgumentError, "`start_day` property is not supported with frequency: on_logon")
+    it "raise error if start_day is passed with invalid frequency (:on_logon)" do
+      expect { resource.send(:validate_start_day, "02/07/1984", :on_logon) }.to raise_error(Chef::Exceptions::ArgumentError, "`start_day` property is not supported with frequency: on_logon")
+    end
+
+    it "does not raise error if start_day is passed with valid frequency (:weekly)" do
+      expect { resource.send(:validate_start_day, "02/07/1984", :weekly) }.not_to raise_error
+    end
+
+    it "raise error if start_day is passed with invalid date format (DD/MM/YYYY)" do
```
stuartpreston: This is not an invalid date format when schtasks.exe is executed on a non-US locale system (from `schtasks.exe /create /?` on my system:

```
...
    /SD   startdate    Specifies the first date on which the task runs. The
                       format is dd/mm/yyyy. Defaults to the current
                       date. This is not applicable for schedule types: ONCE,
                       ONSTART, ONLOGON, ONIDLE, ONEVENT.
```

lib/chef/resource/windows_task.rb
```diff
@@ -44,22 +44,22 @@ def initialize(name, run_context = nil)
...
-                                       :onstart,
-                                       :on_idle,
-                                       :none], default: :hourly
+                                      :hourly,
+                                      :daily,
+                                      :weekly,
+                                      :monthly,
+                                      :once,
+                                      :on_logon,
+                                      :onstart,
```
stuartpreston: Any reason other than legacy that this symbol is inconsistent with the others? (i.e. ideally would be `:on_start`)

lib/chef/resource/windows_task.rb
NimishaS: @tas50 , `execution_time_limit` property was set as string earlier because `schtasks create` accepts input in ISO08601 duration format, which needs to be a string e.g. `PT72H`. Initially this property was accepting input as String in ISO08601 duration format only. It was later changed to accept time in seconds as it's more user friendly.
But yeah the input type should have been changed at that time. But then this [logic](https://github.com/chef/chef/blob/master/lib/chef/provider/windows_task.rb#L339), which fetches all the properties to be updated via task XML in a loop, wouldn't work and there would be more `If-Else` conditions.

spec/unit/resource/windows_task_spec.rb
tas50: From what I understood from our docs and some earlier tickets we were requiring input to be US format on the dates. If that's not correct we need to get our docs updated.

lib/chef/resource/windows_task.rb
tas50: Great question...

lib/chef/resource/windows_task.rb
tas50: @thommay I'll just move that into sec_to_dur and rescue the failure there

lib/chef/resource/windows_task.rb
stuartpreston: I take it back, the existing docs do state to use MM/DD/YYYY irrespective of the system locale and any local knowledge of schtasks.exe.

lib/chef/resource/windows_task.rb
```diff
@@ -92,8 +92,9 @@ def after_created
 
       # Validate the passed value is numeric values only if it is a string
       def numeric_value_in_string?(val)
-        return false if val.is_a?(String) && /\D/ =~ val # \D is any non-numeric value
-        true
+        return true if Integer(val)
```
thommay: if i were being super picky i'd note that `return` there is implicit :)

thommay:
So here's the thing. We're doing something that is fairly clearly not going to be expected behaviour for anyone non-US. We should clearly merge this, because it fixes a great big pile of problems. But we should create an actual *design* for this and get any breaking changes in place in the 14 window

lib/chef/resource/windows_task.rb
tas50: I always feel a bit awful returning a non Boolean for a ? method. Totally works, but feels wrong at the same time.

thommay: ah, i just meant `true if Integer(val)`

PR #6632, opened 2017-12-07
speedup node_map get and set operations

lamont-granquist:
this is the flip side of what was making the old tests so slow and this
change will actually speedup customer runs by short-circuiting some
operations.


lamont-granquist:
This takes the old spec/unit/lwrp_spec.rb code (when it was still polluting the global state) from 6.88 seconds to 0.50512 seconds

lamont-granquist:
doesn't have a measurable effect on our tests now, but for our old pathological behavior it is a 10x speedup...

PR #6636, opened 2017-12-08
Ensure data bags names can contain reserved words

### LINKED ISSUE #3058: One can create data bags with names that are reserved words
One can create a data bag called role, node, client or environment, but then cannot search that data bag. This happens with the latest chef version.

The data bag creation should probably throw an error and not allow creation of a data bag with one of these names.

EmFl:
### Description

I think there was a bug introduced by #3058 where any data bag *containing* reserved words are seen as invalid

### Issues Resolved

#3058

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/data_bag.rb
+++ b/lib/chef/data_bag.rb
@@ -33,7 +33,7 @@ class Chef
     include Chef::Mixin::ParamsValidate
 
     VALID_NAME = /^[\.\-[:alnum:]_]+$/
-    RESERVED_NAMES = /node|role|environment|client/
+    RESERVED_NAMES = /^(node|role|environment|client)$/
 
     attr_accessor :chef_server_rest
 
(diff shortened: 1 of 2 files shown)


thommay:
Obvious fix

thommay:
@EmFl do you fancy adding a test around https://github.com/chef/chef/blob/bcdb1efcb33c580c89a96b50688d6872741ed7ae/spec/unit/knife/data_bag_create_spec.rb#L75 to verify that this works?

EmFl:
@tas50 You renamed this with "not start with them" but my understanding is that a data bag named rolefoobar would be a valid name. Right ?

EmFl:
@thommay I've committed a test case but I'm not familiar at all with rspec (sorry!) so this might not be what you were expecting...

thommay:
This is great, thanks!

PR #6637, opened 2017-12-08
simplify node_map logic

lamont-granquist:
flattens out the internal data structure and removes some excessive
defensive coding.

### DIFF
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -20,18 +20,16 @@
 # example of a NodeMap entry for the user resource (as typed on the DSL):
 #
 #  :user=>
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core this just needs an approval

lib/chef/node_map.rb
```diff
@@ -222,8 +219,12 @@ def compare_matchers(key, new_matcher, matcher)
         # Sort by class name (ascending) as well, if all other properties
         # are exactly equal
         # XXX: remove this in Chef-14 and use last-writer-wins (prepend if they match)
```
thommay: Is this comment still valid?

lamont-granquist: yes, there's an issue i cut on that one as well.   it blocks us from being able to trivially override core in cookbooks.

PR #6656, opened 2017-12-11
[MSYS-727] Added support for setting node policy name and group from knife

piyushawasthi:

### Description
Added support for setting node policy name and group from knife.

### Command
knife node policy set  NODE --policy-name POLICY_NAME --policy-group POLICY_GROUP

### DIFF
--- a/lib/chef/knife/node_policy_set.rb
+++ b/lib/chef/knife/node_policy_set.rb
@@ -1,0 +1,79 @@
+#
+# Author:: Piyush Awasthi (<piyush.awasthi@chef.io>)
+# Copyright:: Copyright 2017-2018, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the License);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/knife"
+
+class Chef
+  class Knife
+    class NodePolicySet < Knife
+
+      deps do
+        require "chef/node"
+        require "chef/json_compat"
+      end
+
+      banner "knife node policy set NODE POLICY_GROUP POLICY_NAME (options)"
+
+      def run
+        validate_node!
+        validate_options!
+        node = Chef::Node.load(@name_args[0])
+        set_policy(node)
+        if node.save
+          ui.info "Successfully set the policy on node #{node.name}"
+        else
+          ui.info "Error in updating node #{node.name}"
+        end
+      end
+
+      private
+
+      # Set policy name and group to node
+      def set_policy(node)
+        policy_group, policy_name = @name_args[1..-1]
+        node.policy_name  = policy_name
+        node.policy_group = policy_group
+      end
+
+      # Validate policy name and policy group
+      def validate_options!
+        if incomplete_policyfile_options?
+          ui.error("Policy group and name must be specified together")
+          exit 1
+        end
+        true
+      end
+
+      # Validate node pass in CLI
+      def validate_node!
+        if @name_args[0].nil?
+          ui.error("You must specify a node name")
+          show_usage
+          exit 1
+        end
+      end
+
+      # True if one of policy_name or policy_group was given, but not both
+      def incomplete_policyfile_options?
+        policy_group, policy_name = @name_args[1..-1]
+        (policy_group.nil? || policy_name.nil? || @name_args[1..-1].size > 2)
+      end
+
+    end
+  end
+end
--- a/spec/unit/knife/node_policy_set_spec.rb
+++ b/spec/unit/knife/node_policy_set_spec.rb
@@ -1,0 +1,122 @@
+#
+# Author:: Piyush Awasthi (<piyush.awasthi@chef.io>)
+# Copyright:: Copyright 2017-2018, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the License);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "spec_helper"
+
+describe Chef::Knife::NodePolicySet do
+  let(:node) do
+    node = Chef::Node.new()
+    node.name("adam")
+    node.run_list = ["role[base]"]
+    node
+  end
+
+  let(:knife) do
+    Chef::Log.logger = Logger.new(StringIO.new)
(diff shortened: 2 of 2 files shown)


thommay:
I wonder if `knife node policy set NODE POLICY_NAME POLICY_GROUP` might not be a better UX. Or, better, ship the two commands included here, but add one that lets you do both.

lib/chef/knife/node_policy_group_set.rb
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/knife"
+require 'pry'
```
thommay: this needs to get removed.

piyushawasthi: Thanks @thommay - It's in-progress and I'm working on it.

coderanger:
This feels waaaaay in the weeds. We still shouldn't be recommending people edit nodes like this because it isn't safe (if there is a running converge it will overwrite the changed settings when it `node.save`s) and there is alreay `knife node edit` for interactive use and `knife exec -E "n = nodes.load('name'); n.policy_group = 'asdf'; n.save"` for people that need super flexible CLI scripting. I'm guessing some folks wanted this for symmetry with existing run list modification commands? If so, I don't think we should repeat that UX for policyfiles :)

piyushawasthi:
@thommay, Please review PR and let me know your thought about it.

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,87 @@
...
+        :description  => "Policy group name to use (--policy-name must also be given)",
+        :default      => nil
+
+      def run
+        validate_node!
+        validate_options!
+        node = Chef::Node.load(@name_args[0])
+        set_policy(node)
+        if node.save
+          ui.info "Successfully policy set on node #{node.name}"
```
NimishaS: Successfully set the policy on node #{node.name}.

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,87 @@
...
+        else
+          ui.info "Node not updated, skipping node save"
```
NimishaS: Is there any error message that can be shown if `node.save` fails?

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+      end
+
+      private
+
+      # Set policy name to node
```
NimishaS: The description can be modified since the method sets both policy name and group.

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,87 @@
...
+        if @name_args[0].nil?
+          ui.error("You must specify a node name")
+          show_usage
+          exit 1
+        end
+      end
+
+      # True if one of policy_name or policy_group was given, but not both
+      def incomplete_policyfile_options?
+        (config[:policy_group].nil? || config[:policy_name].nil?)
```
NimishaS: Can we add this condition directly in `validate_options` method?

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,87 @@
...
+      def validate_options!
+        if incomplete_policyfile_options?
+          ui.error("--policy-name and --policy-group must be specified together")
+          exit 1
+        end
+        true
+      end
+
+      # Validate node
+      def validate_node!
```
NimishaS: This method isn't modifying any object. We shouldn't use `!` here.
This is applicable to `validate_options!` method too.

spec/unit/knife/node_policy_set_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    node = Chef::Node.new()
+    node.name("adam")
+    node.run_list = ["role[base]"]
+    node
+  end
+
+  let(:knife) do
+    Chef::Log.logger = Logger.new(StringIO.new)
+    Chef::Config[:knife][:bootstrap_template] = bootstrap_template unless bootstrap_template.nil?
+    k = Chef::Knife::NodePolicySet.new(bootstrap_cli_options)
```
NimishaS: Please replace `k` with some meaningful variable name.

spec/unit/knife/node_policy_set_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      let(:bootstrap_cli_options) { %w{ adam --policy-name my-app --policy-group staging } }
+      it "returns an error node not updated successfully" do
+        allow(Chef::Node).to receive(:load).with(bootstrap_cli_options[0]).and_return(node)
+        allow(node).to receive(:save).and_return(false)
+        knife.run
+        expect(stderr.string.strip).to eq("Node not updated, skipping node save")
+      end
+    end
+
+    context "when successfully policy set on node" do
```
NimishaS: "when the policy is set successfully on the node"

lib/chef/knife/node_policy_set.rb
piyushawasthi: I am just following same development pattern for better understanding to others https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L489

btm:
Yeah, I agree with @thommay on the UX. We expect this to provide parity for this command:

```
knife node run_list set NODE ENTRIES (options)
```

btm:
@coderanger yes, like setting a run list, setting the policy group is reported to be the most common change to a node. The UX of `knife node edit` and making those changes in JSON is painful (you forgot a comma, hahaha on you!) and that's probably the place we want to discourage people from making a habit of being because it will present you with a bunch of bad pattern ideas.

Anyone other than a very experienced user wanting to set the policy group is going to not find themselves figuring out something like `knife exec -E "n = nodes.load('name'); n.policy_group = 'asdf'; n.save" short of an FAQ that lists every knife exec pattern that might be useful.

And the risk of a race when modifying the node during a run shouldn't keep us from making it easier to work with nodes, it should make us implement [node state separation](https://github.com/chef/chef-rfc/blob/master/rfc045-node_state_separation.md) already.

piyushawasthi:
@btm : currently UX like:
`knife node policy set NODE (options)`
Here you can pass options like: `--policy-name POLICY_NAME --policy-group POLICY_GROUP`. 
Please let me know if you want any changes in UX.

thommay:
i don't think policy name and policy group should be options, they should just be positional arguments:
`knife node policy set NODE GROUP NAME`

piyushawasthi:
Hi @thommay : Updated UX as per your comment please have a look and let me know you thought about that.

spec/unit/knife/node_policy_set_spec.rb
```diff
@@ -0,0 +1,122 @@
+#
+# Author:: Piyush Awasthi (<piyush.awasthi@msystechnologies.com>)
+# Copyright:: Copyright 2017-2018, Piyush Awasthi
```
thommay: Same here.

lib/chef/knife/node_policy_set.rb
```diff
@@ -0,0 +1,79 @@
...
+class Chef
+  class Knife
+    class NodePolicySet < Knife
+
+      deps do
+        require "chef/node"
+        require "chef/json_compat"
+      end
+
+      banner "knife node policy set NODE GROUP NAME (options)"
```
btm: Could this be `knife node policy set NODE POLICY_GROUP POLICY_NAME (options)`? That would be more clear as to what GROUP and the second NAME are.

piyushawasthi:
@thommay : Please review the changes.

thommay:
Thanks!

Vasu1105:
I had added the review comments way before but just missed to finish my review my bad. We can take those review comments and try to fix which make sense.

piyushawasthi:
@Vasu1105 : That's not an issue I put that comments in my mind and will fix with some other card.

PR #6658, opened 2017-12-11
Use comments in the github issue template

tas50:
This way we don't see the end users notes when they submit it. This should make it easier to read the issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,31 +1,5 @@
-## Description
-
-Briefly describe the issue
-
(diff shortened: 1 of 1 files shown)


tas50:
I think the same people will read / edit these, but we won't have to see it in partially filled out issues

lamont-granquist:
yeah like the comments.

i was going this direction:

https://github.com/chef/chef/pull/5933/files

i didn't know you could drop stuff into comments.

i also moved the links at the bottom up to the top, but was trying to make it a bit more useful and a bit less terse and indirect.

PR #6660, opened 2017-12-12
implement credential management

thommay:
per RFC 99

### DIFF
--- a/chef-config/lib/chef-config/workstation_config_loader.rb
+++ b/chef-config/lib/chef-config/workstation_config_loader.rb
@@ -31,7 +31,9 @@ module ChefConfig
 
     # Path to a config file requested by user, (e.g., via command line option). Can be nil
     attr_accessor :explicit_config_file
-    attr_reader :profile
+    # The name of a credentials profile. Can be nil
+    attr_accessor :profile
+    attr_reader :credentials_found
 
     # TODO: initialize this with a logger for Chef and Knife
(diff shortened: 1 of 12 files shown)


chef-config/lib/chef-config/workstation_config_loader.rb
```diff
@@ -138,6 +146,34 @@ def working_directory
       a
     end
 
+    def apply_credentials(creds, _profile)
+      Config.node_name = creds.fetch("node_name") if creds.key?("node_name")
+      Config.node_name = creds.fetch("client_name") if creds.key?("client_name")
```
adamleff: I wonder if this is worthy of a warning log message or similar if the user has both a `node_name` and a `client_name` in their credentials.

chef-config/lib/chef-config/workstation_config_loader.rb
```diff
@@ -138,6 +146,34 @@ def working_directory
       a
     end
 
+    def apply_credentials(creds, _profile)
```
adamleff: Is accepting the profile here just future-proofing? Or am I missing some context (totally reasonable assumption on your part).

chef-config/lib/chef-config/workstation_config_loader.rb
thommay: Yeah that really needs a raise.

lamont-granquist:
still 👍

PR #6661, opened 2017-12-12
memoize some work in the package class

lamont-granquist:
This is a 10% perf gain for lightweight use of package resources like `build_resource(:yum_package, 'lsof').current_value.version`


lamont-granquist:
fails on a few yum provider unit tests that i deleted in the great refactoring, can't merge until that one goes in

lamont-granquist:
I got the memo. And I understand the policy. And the problem is just that I forgot the one time. And I've already taken care of it so it's not even really a problem anymore....

PR #6664, opened 2017-12-13
[MSYS-692] Fix issue with PowerShell function buffer

TheLunaticScripter:
Signed-off-by: John Snow <thelunaticscripter@outlook.com>

Submitted with contributions also from @btm 

### Description

When a command, like knife, is run using the chef.psm1 function the output of the command was buffered until the command completed so that an object could be created and passed through a pipe. This prevents UI elements from being visible to the user as well as stdout during the running of the ruby command. By instead passing the output of the ruby command direct instead of storing it in a string buffer this helps with both piping commands and allow for UI elements to be visible.

Note: If a UI element is piped such as `knife client delete fake-node > output.txt` the UI element will be in the text file not in the console.

### Issues Resolved

https://github.com/chef/chef-dk/issues/1236
Sustaining Eng MSYS-692
ZenDesk 16768

### Check List

- [ ] All tests pass

### DIFF
--- a/distro/powershell/chef/chef.psm1
+++ b/distro/powershell/chef/chef.psm1
@@ -236,7 +236,6 @@ function Run-ExecutableAndWait($AppPath, $ArgumentString) {
     throw "Unable to create process [$ArgumentString].  Error code $reason."
   }
 
-  $sb = New-Object System.Text.StringBuilder
   $buffer = New-Object byte[] 1024
 
   # Initialize reference variables
@@ -270,7 +269,7 @@ function Run-ExecutableAndWait($AppPath, $ArgumentString) {
       while ([Chef.Kernel32]::ReadFile($hReadOut, $buffer, $buffer.Length, [ref] $bytesRead, 0)) {
         $output = [Text.Encoding]::UTF8.GetString($buffer, 0, $bytesRead)
         if ($output) {
-          [void]$sb.Append($output)
+          $output
         }
         if ($bytesRead -lt $buffer.Length) {
           # Partial buffer indicating the end of stream, break out of ReadFile loop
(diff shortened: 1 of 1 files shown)


btm:
@coolmacool do you have any thoughts on this solution?

vrimkus:
@btm (@coolmacool here, I merged accounts)

This looks good, and is definitely more in line with the PowerShell way of doing things.  I am trying to remember the exact reasoning, but I think initially I was worried about having partial output sent directly to stdout. But running stuff like `knife node show testnode -l -F json | ConvertFrom-Json ` with this change, is resulting in the expected behavior.

This does kinda touch on some of the thoughts posed by @ksubrama in the module comments of https://github.com/chef/chef/commit/fdf8e17abf9a0c114e3f6657968dd47548b819b7#diff-58c67b3f13e3428d18ea2971b8dba902R294

Should we still be going with this method for executing the ruby commands?

I played with it a little today and it seems like it may be possible to eliminate all the Win32 P\Invoking, in favor of splatting an array of commands & arguments when invoking ruby with the call operator (`&`). Still need more testing on lower PS versions/other Windows editions, but here is a gist:
 [chef_nowin32.psm1](https://gist.github.com/vrimkus/fd743dd22e67b4977ebee0ba45dfe0c1)

If this is something that we want to explore, I can open a new issue around that.

btm:
Travis failures look like the weirdness we've been seeing over there.

@vrimkus I'm not sure which part of the script or the whole script this comment that you linked to referred to:
>  1) Maybe this entire script is unnecessary and the bugs mentioned below have been fixed

Maybe that's referring to the portions related to handling processes only, but this entire script came about due to file globbing and escaping madness reported in #1687. I don't expect that problem has gone away on Windows yet.

tas50:
If this is rebased on master it's a bit easier to hammer on Travis until it passes. They have some bad workers in their infra that are causing failures right now.

PR #6671, opened 2017-12-14
Avoid a few initializers in resources by using the DSL we have

tas50:
Tip of the iceberg here, but it's the low hanging fruit

### DIFF
--- a/lib/chef/resource/msu_package.rb
+++ b/lib/chef/resource/msu_package.rb
@@ -24,23 +24,18 @@ class Chef
     class MsuPackage < Chef::Resource::Package
       include Chef::Mixin::Uris
 
+      resource_name :msu_package
       provides :msu_package, os: "windows"
 
       allowed_actions :install, :remove
-
-      def initialize(name, run_context = nil)
-        super
-        @resource_name = :msu_package
-        @source = name
-        @action = :install
-      end
+      default_action :install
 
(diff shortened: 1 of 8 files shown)


lib/chef/resource/msu_package.rb
```diff
@@ -24,23 +24,18 @@ class Resource
...
-      end
+      default_action :install
 
       property :source, String,
                 coerce: (proc do |s|
                   unless s.nil?
                     uri_scheme?(s) ? s : Chef::Util::PathHelper.canonical_path(s, false)
                   end
-                end)
+                end), name_property: true
```
coderanger: Might want to put this before the `coerce` so it's easier to see when scanning.

tas50: Fixed

lib/chef/resource/msu_package.rb
```diff
@@ -24,18 +24,14 @@ class Resource
...
-      def initialize(name, run_context = nil)
-        super
-        @resource_name = :msu_package
-        @source = name
-        @action = :install
-      end
+      default_action :install
 
       property :source, String,
+                name_property: true,
```
tas50: This fixes 1 of 2 bugs in how we setup this resource. I've filed #6674 for the other issue, which I'll fix and get a test around

PR #6678, opened 2017-12-15
Experiment with using CircleCI rather than travis

thommay:


lamont-granquist:
google trends approves:  https://trends.google.com/trends/explore?q=circleci,Travis%20CI

PR #6687, opened 2017-12-18
[MSYS-724] Chef::Util::Windows::LogonSession should allow having only the prescribed users permissions

NimishaS:
### Description

While finishing https://github.com/chef/chef/pull/5832,  we changed the logon type in `Chef::Util::Windows::LogonSession` from `LOGON32_LOGON_NETWORK` to `LOGON32_LOGON_NEW_CREDENTIALS` to facilitate passing credentials to a remote system without logging in locally using those credentials, to handle a use case where those credentials were valid on the remote system but not on the local system.
However, this means that `Chef::Mixin::UserContext.with_user_context` might log on to the local system as a new user but continue using the old users token and thus the old/original users privileges.
`Chef::ReservedNames::Win32::Security.has_admin_privileges?` is really looking to see if we're elevated, not 'admin' per se. We should print some debug information about the token like if it is elevated. With this information you'll see that when we run chef from an elevated shell, we still have an elevated token even if we run `has_admin_privileges?` from inside `with_user_context` using a regular user.
The above functionality was masked by this line in the before block in ```spec\functional\win32\security_spec.rb:
allow(Chef::ReservedNames::Win32::Security).to receive(:OpenProcessToken).and_return(true)```
This caused the 'running as non admin user' test to pass even though it should fail. Fixing this might be hard, because a regular user cannot use much of the functionality in `has_admin_privileges`, maybe not even calling `open_current_process_token`. One solution might be catching the access denied results and returning false, presuming that if we get an access denied in this area then we aren't elevated.

Note that we reverted the functional tests in https://github.com/chef/chef/pull/6588 due odd behavior on Appveyor.

### Acceptance Criteria

- [x] able to run a ruby block with only the privileges of another user, while preserving the functionality needed by remote_file.
- [x] `Chef::ReservedNames::Win32::Security.has_admin_privileges?` prints a debug message that contains token elevation information.
- [x] The Chef::Win32::Security functional test for running as a non-admin user that was reverted is returned and does not return a false positive.
- [x] The functional tests pass on appveyor or we have a documented reason to exclude them on appveyor. 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -31,7 +31,7 @@ class Chef
           raise ArgumentError, "You must supply a block to `with_user_context`"
         end
 
-        login_session = nil
+        logon_session = nil
 
         begin
           if user
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -22,7 +22,8 @@ class Chef
   module Mixin
     module UserContext
 
-      def with_user_context(user, password, domain = nil, &block)
+      # valid logon_type values => :remote, :local
+      def with_user_context(user, password, domain = nil, logon_type = :remote, &block)
         unless Chef::Platform.windows?
           raise Exceptions::UnsupportedPlatform, "User context impersonation is supported only on the Windows platform"
         end
@@ -35,7 +36,7 @@ class Chef
 
         begin
           if user
-            logon_session = Chef::Util::Windows::LogonSession.new(user, password, domain)
+            logon_session = Chef::Util::Windows::LogonSession.new(user, password, domain, logon_type)
             logon_session.open
             logon_session.set_user_context
           end
--- a/lib/chef/mixin/user_context.rb
+++ b/lib/chef/mixin/user_context.rb
@@ -22,8 +22,10 @@ class Chef
   module Mixin
     module UserContext
 
-      # valid logon_type values => :remote, :local
-      def with_user_context(user, password, domain = nil, logon_type = :remote, &block)
+      # valid values for authentication => :remote, :local
+      # When authentication = :local, we use the credentials to create a logon session against the local system, and then try to access the files.
+      # When authentication = :remote, we continue with the current user but pass the provided credentials to the remote system.
+      def with_user_context(user, password, domain = nil, authentication = :remote, &block)
         unless Chef::Platform.windows?
           raise Exceptions::UnsupportedPlatform, "User context impersonation is supported only on the Windows platform"
         end
@@ -36,7 +38,7 @@ class Chef
 
         begin
           if user
-            logon_session = Chef::Util::Windows::LogonSession.new(user, password, domain, logon_type)
+            logon_session = Chef::Util::Windows::LogonSession.new(user, password, domain, authentication)
             logon_session.open
             logon_session.set_user_context
           end
--- a/lib/chef/util/windows/logon_session.rb
+++ b/lib/chef/util/windows/logon_session.rb
@@ -25,7 +25,7 @@ class Chef
       class LogonSession
         include Chef::Mixin::WideString
 
-        def initialize(username, password, domain = nil, logon_type)
+        def initialize(username, password, domain = nil, logon_type = :remote)
           if username.nil? || password.nil?
             raise ArgumentError, "The logon session must be initialize with non-nil user name and password parameters"
           end
--- a/lib/chef/util/windows/logon_session.rb
+++ b/lib/chef/util/windows/logon_session.rb
@@ -25,7 +25,7 @@ class Chef
       class LogonSession
         include Chef::Mixin::WideString
 
-        def initialize(username, password, domain = nil, logon_type = :remote)
+        def initialize(username, password, domain = nil, authentication = :remote)
           if username.nil? || password.nil?
             raise ArgumentError, "The logon session must be initialize with non-nil user name and password parameters"
           end
@@ -33,7 +33,7 @@ class Chef
           @original_username = username
           @original_password = password
           @original_domain = domain
-          @logon_type = logon_type
+          @authentication = authentication
           @token = FFI::Buffer.new(:pointer)
           @session_opened = false
           @impersonating = false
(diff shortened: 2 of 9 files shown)


NimishaS:
@btm , should we also add a property called `logon_type` in `remote_file` resource so that users can set this as `:local` in case they want to access any file from local system with an alternate user?

lib/chef/mixin/user_context.rb
```diff
@@ -22,7 +22,8 @@ class Chef
   module Mixin
     module UserContext
 
-      def with_user_context(user, password, domain = nil, &block)
+      # valid logon_type values => :remote, :local
```
btm: Please add a comment like, "logon_type specifies if we use the credentials to create a logon session against the local system, and then try to access the files in the case of :local, or if we continue with the current user but pass the provided credentials to the remote system with :remote"

spec/functional/win32/security_spec.rb
```diff
@@ -26,13 +28,36 @@
...
+    let(:user) { "security_user" }
+    let(:password) { "Security@123" }
+
+    let(:domain) do
+      whoami = Mixlib::ShellOut.new("whoami")
+      whoami.run_command
+      whoami.error!
+      whoami.stdout.split("\\")[0]
+    end
+    before do
```
btm: there are a few nit-picky missing blank lines between blocks in here.

NimishaS:
Rebased the PR, but appveyor issues aren't resolved. Now we have travis failures too. The failures don't look related to the PR.

btm:
The travis failures look like the ones I've been seeing lately.

btm:
@NimishaS looks like tests are mostly passing on master now, I just see one odd travis failure. might try another rebase?

NimishaS:
Rebased the PR again. There are several failures in Travis for Ruby 2.4.3 and 2.5.0 for Unit specs.
Appveyor is also failing: https://ci.appveyor.com/project/chef/chef/build/master-11374/job/kguoj00fvb4unk0t#L2302
The failures aren't related to the PR.

lib/chef/util/windows/logon_session.rb
```diff
@@ -47,7 +48,11 @@ def open
           password = wstring(original_password)
           domain = wstring(original_domain)
 
-          status = Chef::ReservedNames::Win32::API::Security.LogonUserW(username, domain, password, Chef::ReservedNames::Win32::API::Security::LOGON32_LOGON_NEW_CREDENTIALS, Chef::ReservedNames::Win32::API::Security::LOGON32_PROVIDER_DEFAULT, token)
+          if authentication == :remote
+            status = Chef::ReservedNames::Win32::API::Security.LogonUserW(username, domain, password, Chef::ReservedNames::Win32::API::Security::LOGON32_LOGON_NEW_CREDENTIALS, Chef::ReservedNames::Win32::API::Security::LOGON32_PROVIDER_DEFAULT, token)
```
jaym: nitpick: It's hard to tell the difference between these two branches. I think it would be cleaner just to switch the changing input which looks like `LOGON32_LOGON_NEW_CREDENTIALS` vs `LOGON32_LOGON_NETWORK`

lib/chef/win32/security.rb
```diff
@@ -341,6 +341,21 @@ def self.get_token_information_primary_group(token)
...
+      def self.get_token_information_elevation_type(token)
+        token_result_size = FFI::MemoryPointer.new(:ulong)
+        if GetTokenInformation(token.handle.handle, :TokenElevationType, nil, 0, token_result_size)
+          raise "Expected ERROR_INSUFFICIENT_BUFFER from GetTokenInformation, and got no error!"
+        elsif FFI::LastError.error != ERROR_INSUFFICIENT_BUFFER
+          Chef::ReservedNames::Win32::Error.raise!
+        end
+        info_ptr = FFI::MemoryPointer.new(:pointer)
+        token_info_pointer = TOKEN_ELEVATION_TYPE.new info_ptr
+        unless GetTokenInformation(token.handle.handle, :TokenElevationType, token_info_pointer, 4, token_result_size)
```
jaym: Is there a `sizeof` function we can use instead of 4?

lib/chef/win32/security.rb
```diff
@@ -341,6 +341,21 @@ def self.get_token_information_primary_group(token)
         SID.new(group_result[:PrimaryGroup], group_result_storage)
       end
 
+      def self.get_token_information_elevation_type(token)
+        token_result_size = FFI::MemoryPointer.new(:ulong)
+        if GetTokenInformation(token.handle.handle, :TokenElevationType, nil, 0, token_result_size)
+          raise "Expected ERROR_INSUFFICIENT_BUFFER from GetTokenInformation, and got no error!"
```
jaym: Why is this check needed?

lib/chef/win32/security.rb
```diff
@@ -632,7 +647,19 @@ def self.has_admin_privileges?
 
           true
         else
-          process_token = open_current_process_token(TOKEN_READ)
+          # a regular user doesn't have privileges to call Chef::ReservedNames::Win32::Security.OpenProcessToken
+          # hence we return false if the open_current_process_token fails with `Access is denied.` error message.
+          begin
+            process_token = open_current_process_token(TOKEN_READ)
+          rescue Exception => run_error
+            return false if run_error.message =~ /Access is denied/
```
jaym: Will this cause a problem for non english locales? I'm guessing chef generally doesn't work in those cases anyway?

lib/chef/win32/security.rb
NimishaS: @jaym , `GetTokenInformation` method does return `ERROR_INSUFFICIENT_BUFFER` error if the input provided to the method isn't correct. It's not mandatory to keep this check, but I have made it consistent with other existing methods e.g. https://github.com/chef/chef/blob/master/lib/chef/win32/security.rb#L317 and https://github.com/chef/chef/blob/master/lib/chef/win32/security.rb#L332.
Please let me know your thought on that.

lib/chef/win32/security.rb
NimishaS: When `false` is returned from this method (`has_admin_privileges?`), then we raise a customized error message. If this fails for a non English locale, then the next line `Chef::ReservedNames::Win32::Error.raise!` will raise the error.
So this won't particularly break anything.

lib/chef/win32/security.rb
NimishaS: Ideally it should have been `size_of(token_info_pointer)`, but it's initially 0. So that gives error. I have introduced a variable `token_info_length` for better readability. Please check if it looks fine.

btm:
Tests in master are still messed up :(

Issue #6638, opened 2017-12-08
windows_task not fully idempotent


tas50:
In some scenarios the windows_task resource updates on every run. Here's an example resource that will update on every run:

```ruby
windows_task 'chef-client 5' do
  command 'chef-client'
  run_level :highest
  frequency :weekly
  frequency_modifier 3
  random_delay '60'
end
```

troyready:
I'm seeing this being caused by the `execution_time_limit`, e.g.:
```
windows_task "mytask" do
  cwd 'c:\chef'
  command "myscript.bat"
  run_level :highest
  frequency :weekly
  execution_time_limit '720'
```
Subsequent chef-client runs show the old resource as having time limit value `PT12M` and the new_resource as having value `PT12M0S` which triggers the resource as needing an update.

Ran into this on 13.2, and verified it's still present in 13.6.4

tas50:
Thanks for the details. That’s something we can potentially clean up easily

PR #6688, opened 2017-12-18
Fix windows_task resource not being idempotent for random_delay and execution_time_limit

Vasu1105:
windows_task resource is not idempotent for random_delay and execution_time_limit property
```
windows_task 'chef-client 5' do
  command 'chef-client'
  run_level :highest
  frequency :weekly
  frequency_modifier 3
  random_delay '60'
  day 'Mon'
end
```

```
windows_task 'chef-client 5' do
  command 'chef-client'
  run_level :highest
  frequency :weekly
  frequency_modifier 3
  execution_time_limit '60'
  day 'Mon'
end
```
#6638 partially fixed because if day is not set the resource is still not idempotent for day. We will address this issue in different PR.
 
### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -251,9 +251,9 @@ class Chef
       def random_delay_updated?
         if new_resource.random_delay.nil?
           false
-        elsif current_resource.random_delay.nil? &&  new_resource.random_delay == "PT0S" # when user sets random_dealy to 0 sec
+        elsif current_resource.random_delay.nil? && new_resource.random_delay == "PT0S" # when user sets random_dealy to 0 sec
           false
-        elsif current_resource.random_delay.nil?  && !new_resource.random_delay.nil?
+        elsif current_resource.random_delay.nil? && !new_resource.random_delay.nil?
           true
         else
           ISO8601::Duration.new(current_resource.random_delay) != ISO8601::Duration.new(new_resource.random_delay)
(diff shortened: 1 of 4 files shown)


lib/chef/provider/windows_task.rb
```diff
@@ -243,6 +242,37 @@ def task_need_update?
...
+        end
+      end
+
+      # Comparing execution_time_limit values using Ref: https://github.com/arnau/ISO8601/blob/master/lib/iso8601/duration.rb#L18-L23
+      def execution_time_limit_updated?
+        if new_resource.execution_time_limit.nil?
+          false
+        elsif current_resource.execution_time_limit.nil? && new_resource.execution_time_limit == "PT0S" # when user sets random_dealy to 0 sec
+          false
+        elsif current_resource.execution_time_limit.nil? && !new_resource.execution_time_limit.nil?
```
tas50: You've already checked to see if new_resource.execution_time_limit.nil? above so you know at this point it's not nil.

lib/chef/resource/windows_task.rb
```diff
@@ -205,38 +205,12 @@ def validate_idle_time(idle_time, frequency)
...
-          if seconds > 60 # more than a minute
-            minutes = seconds / 60
-            seconds -= (60 * minutes)
-            iso_str << "#{minutes}M"
-          end
-          iso_str << "#{seconds}S"
-        end
-
-        iso_str
+        dur = ISO8601::Duration.new(seconds.to_i).to_s
```
tas50: It shouldn't be necessary to set the value equal to anything here. Just use ISO8601::Duration.new(seconds.to_i).to_s. Also keep the YARD comment above for the return value: @return [String] ISO8601 representation of seconds

lib/chef/provider/windows_task.rb
```diff
@@ -243,6 +242,37 @@ def task_need_update?
...
+      # ds = ISO8601::Duration.new('P65707200S')
+      # dp = ISO8601::Duration.new('P2Y1MT2H')
+      # di == dp # => true
+      # di == ds # => true
+      def random_delay_updated?
+        if new_resource.random_delay.nil?
+          false
+        elsif current_resource.random_delay.nil? && new_resource.random_delay == "PT0S" # when user sets random_dealy to 0 sec
+          false
+        elsif current_resource.random_delay.nil? && !new_resource.random_delay.nil?
```
tas50: You've already returned if new_resource.random_delay.nil? above so at this point you know new_resource.random_delay is not nil and shouldn't need to check it again

Vasu1105:
@tas50 I have fixed the review comments the travis failures are not part of this changes.

PR #6689, opened 2017-12-18
Modernize cookbook_file resource and expand specs

tas50:
Don't hardcode the provider since the provider already has a provides line

Test the source logic default properly by creating the resource name as a absolute path, which is how the user would use it

### DIFF
--- a/lib/chef/resource/cookbook_file.rb
+++ b/lib/chef/resource/cookbook_file.rb
@@ -27,23 +27,12 @@ class Chef
     class CookbookFile < Chef::Resource::File
       include Chef::Mixin::Securable
 
+      resource_name :cookbook_file
+
+      property :source, [ String, Array ], default: lazy { |r| ::File.basename(r.name) }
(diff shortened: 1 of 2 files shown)


lib/chef/resource/cookbook_file.rb
```diff
@@ -27,23 +27,12 @@ class Resource
...
+      resource_name :cookbook_file
 
-      def source(source_filename = nil)
-        set_or_return(:source, source_filename, :kind_of => [ String, Array ])
-      end
-
-      def cookbook(cookbook_name = nil)
-        set_or_return(:cookbook, cookbook_name, :kind_of => String)
-      end
+      property :source, [ String, Array ], default: lazy { |r| ::File.basename(r.name) }
```
coderanger: Unless things changed this can be `lazy { ::File.basename(name) }`.

tas50: Fixed

PR #6692, opened 2017-12-19
Cleanup to some of the resource specs

tas50:
Expand some of the resource specs to test inputs, actions, and name properties, while not testing the resource class which we just defined in the let block

Remove usage of "should" to make rspec folks happy

Convert before -> let for the resources

Add a few missing package specs


thommay:
two legit travis fails but otherwise 👍

PR #6694, opened 2017-12-19
Chef 17361 - enable yum_package on AIX

kristofwillen:
### Description

We are using yum on AIX, and want to start installing packages with Chef.  When using  yum_package, however, we get the following error message : 

Error executing action `install` on resource 'yum_package[logrotate]' 
Chef::Exceptions::ProviderNotFound 
Cannot find a provider for yum_package[logrotate] on aix version 7.1

When taking a look at the source for the yum provider, it seems indeed that only linux is supported as OS.  This pull request solves this issue, and enables yum_package on AIX.  When implementing this patch on our AIX chef clients, yum_package works successfully on AIX.

Our current Chef client is v12.21.4, but this issue has also been tested and confirmed with the latest Chef-client v13. 

### Issues Resolved

Chef ticket #17361 (https://getchef.zendesk.com/hc/en-us/requests/17361)

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -29,8 +29,8 @@ class Chef
       class Yum < Chef::Provider::Package
         include Chef::Mixin::GetSourceFromPackage
 
-        provides :package, platform_family: %w{rhel fedora amazon aix}
-        provides :yum_package, os: %{linux aix}
+        provides :package, platform_family: %w{rhel fedora amazon}
+        provides :yum_package, os: "linux"
 
         # Multipackage API
         allow_nils
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -29,8 +29,8 @@ class Chef
       class Yum < Chef::Provider::Package
         include Chef::Mixin::GetSourceFromPackage
 
-        provides :package, platform_family: %w{rhel fedora amazon}
-        provides :yum_package, os: "linux"
+        provides :package, platform_family: %w{rhel fedora amazon aix}
+        provides :yum_package, os: %{linux aix}
 
         # Multipackage API
         allow_nils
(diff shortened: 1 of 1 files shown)


tas50:
@kristofwillen There's certainly some improvements that could be made to yum on AIX so thanks for bringing that up, but this PR has a few issues at the moment. The bff_package provider currently maps to package on AIX systems since that's the default package system on AIX. We can't map two providers to the package resource since that would break things. yum_package we can totally make work on AIX though, but we have a test right now to prevent it working on very old AIX systems. I'm going to check and see if we still need that test since we no longer support aix 5.6. Otherwise you can add a provides line just for AIX that did a version check. I'll get back to you on that

lib/chef/provider/package/yum.rb
```diff
@@ -29,8 +29,8 @@ class Package
       class Yum < Chef::Provider::Package
         include Chef::Mixin::GetSourceFromPackage
 
-        provides :package, platform_family: %w{rhel fedora amazon}
-        provides :yum_package, os: "linux"
+        provides :package, platform_family: %w{rhel fedora amazon aix}
```
lamont-granquist: package needs to be bff_package on AIX so this line needs to be reverted, you must use `yum_package` explicitly on AIX

lib/chef/provider/package/yum.rb
```diff
@@ -29,8 +29,8 @@ class Package
...
+        provides :yum_package, os: %w{linux aix}
```
lamont-granquist: this line should just be `provides :yum_package` there's actually little point in the os restriction and we've been removing those everywhere

lamont-granquist:
also the yum_package resource needs to be updated as well.

and if you really wish to override `package` then you can use something like this locally in a library file:

```ruby
Chef::Provider::Package::Yum.provides :package, os: :aix, override: true
Chef::Resource::YumPackage,provides :package, os: :aix, override: true
```

but we can't accept that as a PR, and that may cause issues with your ability to ingest community cookbooks in the future since it swaps around the rule for the package provider.

lamont-granquist:
i included the patch to remove the os restriction on `yum_package` in #6540

PR #6699, opened 2017-12-19
Modernize windows_path resource

tas50:

### DIFF
--- a/lib/chef/resource/windows_path.rb
+++ b/lib/chef/resource/windows_path.rb
@@ -24,19 +24,11 @@ class Chef
     #
     # @since 13.4
     class WindowsPath < Chef::Resource
-
+      resource_name :windows_path
       provides :windows_path, os: "windows"
 
       allowed_actions :add, :remove
       default_action :add
-
-      def initialize(name, run_context = nil)
-        super
-        @resource_name = :windows_path
-        @path = name
-        @provider = Chef::Provider::WindowsPath
-        @action = :add
-      end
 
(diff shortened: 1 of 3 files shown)


lamont-granquist:
windows specs are busted because of the (correct) removal of the hardcoded provider.  kind of looks like the provider is missing a provides line.

lib/chef/resource/windows_path.rb
```diff
@@ -24,20 +24,12 @@ class Resource
     #
     # @since 13.4
     class WindowsPath < Chef::Resource
-
+      resource_name :windows_path
       provides :windows_path, os: "windows"
```
lamont-granquist: should drop the "os: windows" bit here.

tas50: I was planning to nuke those in Chef 14. I'm certain there's people out there coding around the fact that these are no-ops on non-windows platforms. We can nuke them all out then and everyone is happy.

lamont-granquist: they're not no-ops, though, they blow up hard.

lamont-granquist: and i think right now chefspec testing of windows_path on anything other than windows might be problematic -- sort of a bug we should fix in chefspec, but its easy enough to fix it so that even on linux or whatever we can resolve windows_path to a resource.

tas50: I thought chefspec was one of the big reasons we kept all this stuff the way it was.

PR #4285, opened 2015-12-11
Make script resources use properties

jkeiser:

PR #6700, opened 2017-12-19
WIP: Modernize the execute command

tas50:
Cleaned up a the spec naming a bit while I was in there


tas50:
This is going to require some work to get the validation stuff happy

lamont-granquist:
yeah this is at the base of all the script resources.  really the script resources should all use composition and a execute sub-resource rather than inheriting from execute, but likelihood of subtle issues due to inheritance is high and you should make sure travis and appveyor really gets to green on this one.

tas50:
Closing this out in favor of #4285 which handles the script rework necessary to modernize execute.

PR #6703, opened 2017-12-20
Fix dscl group provider gid_used?

get9:
### Description

To check whether a requested gid is used by the system already, the macOS dscl group provider lists all gids on the system and parses the output. There's a bug wherein the parsing looks for the number and a newline, but this yields false positives, e.g. if `dscl . list /Groups gid` returns:
```
group1          100
group2          148
```
and the code `gid_used?(48)` is run currently, it will return `true` even though gid 48 isn't present.

### Issues Resolved

None

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/group/dscl.rb
+++ b/lib/chef/provider/group/dscl.rb
@@ -93,7 +93,7 @@ class Chef
           # dscl -search should not return anything if the gid doesn't exist,
           # but on the off-chance that it does, check whether the given gid is
           # in the output.
-          !!(search_gids =~ /\b#{gid.to_s}\b/)
+          !!(search_gids =~ /\b#{gid}\b/)
         end
 
         def set_gid
--- a/lib/chef/provider/group/dscl.rb
+++ b/lib/chef/provider/group/dscl.rb
@@ -88,8 +88,12 @@ class Chef
 
         def gid_used?(gid)
           return false unless gid
-          search_gids = safe_dscl('search', '/Groups', 'gid', gid.to_s)
-          !search_gids.strip.empty?
+          search_gids = safe_dscl("search", "/Groups", "PrimaryGroupID", gid.to_s)
+
+          # dscl -search should not return anything if the gid doesn't exist,
+          # but on the off-chance that it does, check whether the given gid is
+          # in the output.
+          !!(search_gids =~ /\b#{gid.to_s}\b/)
         end
 
         def set_gid
--- a/lib/chef/provider/group/dscl.rb
+++ b/lib/chef/provider/group/dscl.rb
@@ -88,8 +88,8 @@ class Chef
 
         def gid_used?(gid)
           return false unless gid
-          groups_gids = safe_dscl("list", "/Groups", "gid")
-          !!( groups_gids =~ Regexp.new("#{Regexp.escape(gid.to_s)}\n") )
+          search_gids = safe_dscl('search', '/Groups', 'gid', gid.to_s)
+          !search_gids.strip.empty?
         end
 
         def set_gid
--- a/spec/unit/provider/group/dscl_spec.rb
+++ b/spec/unit/provider/group/dscl_spec.rb
@@ -117,21 +117,25 @@ describe Chef::Provider::Group::Dscl do
     before do
       @node = Chef::Node.new
       @provider = Chef::Provider::Group::Dscl.new(@node, @new_resource)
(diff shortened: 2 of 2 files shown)


spec/unit/provider/group/dscl_spec.rb
```diff
@@ -117,20 +117,27 @@
...
 
-    it "should run safe_dscl with list /Groups gid" do
-      expect(@provider).to receive(:safe_dscl).with(*"list /Groups gid".split(" "))
+    it "should run safe_dscl with search /Groups gid" do
+      expect(@provider).to receive(:safe_dscl).with(*"search /Groups gid 500".split(" ")).and_return('')
       @provider.gid_used?(500)
     end
 
     it "should return true for a used gid number" do
+      allow(@provider).to receive(:safe_dscl).and_return(<<-eos
```
jaymzh: put this in the before block so that your other tests are actually useful...

spec/unit/provider/group/dscl_spec.rb
```diff
@@ -117,20 +117,27 @@
...
+            500
+        )
+        eos
+      )
       expect(@provider.gid_used?(500)).to be_truthy
     end
 
     it "should return false for an unused gid number" do
+      allow(@provider).to receive(:safe_dscl).and_return('')
       expect(@provider.gid_used?(501)).to be_falsey
```
jaymzh: for example here you want to make sure 50, 00, 5 don't match just because 500 is in use. so having the same mocked output is useful.

lib/chef/provider/group/dscl.rb
```diff
@@ -88,8 +88,8 @@ def get_free_gid(search_limit = 1000)
 
         def gid_used?(gid)
           return false unless gid
-          groups_gids = safe_dscl("list", "/Groups", "gid")
-          !!( groups_gids =~ Regexp.new("#{Regexp.escape(gid.to_s)}\n") )
+          search_gids = safe_dscl('search', '/Groups', 'gid', gid.to_s)
```
jaymzh: you'll need to move from our style to upstream style.

spec/unit/provider/group/dscl_spec.rb
get9: I don't quite follow how having this in the before block is helpful. The check for whether the given gid is used is whether safe_dscl returns any output or not. For the tests of truthy/falsey values it would have to mock different things.

lib/chef/provider/group/dscl.rb
```diff
@@ -88,8 +88,12 @@ def get_free_gid(search_limit = 1000)
...
         def gid_used?(gid)
           return false unless gid
-          groups_gids = safe_dscl("list", "/Groups", "gid")
-          !!( groups_gids =~ Regexp.new("#{Regexp.escape(gid.to_s)}\n") )
+          search_gids = safe_dscl("search", "/Groups", "PrimaryGroupID", gid.to_s)
+
+          # dscl -search should not return anything if the gid doesn't exist,
+          # but on the off-chance that it does, check whether the given gid is
+          # in the output.
+          !!(search_gids =~ /\b#{gid.to_s}\b/)
```
jaymzh: `#{}` implies .to_s :) Since it's no longer wrapped in another method you should be able to drop .to_s

spec/unit/provider/group/dscl_spec.rb
```diff
@@ -131,6 +136,7 @@
 
     it "should return false for an unused gid number" do
       expect(@provider.gid_used?(501)).to be_falsey
+      expect(@provider.gid_used?(18)).to be_falsey
```
jaymzh: as I mentioned before here, interesting tests with this particular output are if 50, 5000, 00, 5, or 0 return falsey, since that's the only real logic your function does - check that the output contains the gid you searched for, checking that's actually valid (so say someone else doesn't break it later) is a really good test. I can see someone checking that it say it *contains* the number or *startswith* the number (by accident).

That's why I suggested a moving the output to the before block so you can run all these various tests against that output.

jaymzh:
Something went seriously wrong with your repo....

jaymzh:
OK this looks good to me. Hey @jtimberman and @mikedodge04 - any comments from the mac folks? Also while I'm at it, CC @chef/client-os-x

mikedodge04:
I verified on my system. Looks good to me. Thanks for this.

PR #6706, opened 2017-12-20
Speed up Appveyor testing with bundler caching

tas50:
Start by shuffling around the config to not do a build. This makes it easier to follow and matches their Ruby guide. Then enable bundler caching per their docs.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,18 +26,20 @@ install:
   - echo %OMNIBUS_RUBYGEMS%
   - ps: $env:OMNIBUS_BUNDLER=$(findstr bundler omnibus_overrides.rb | %{ $_.split(" ")[3] })
   - echo %OMNIBUS_BUNDLER%
-  - ruby --version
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -26,19 +29,22 @@ install:
...
   - appveyor DownloadFile http://curl.haxx.se/ca/cacert.pem -FileName C:\cacert.pem
   - set SSL_CERT_FILE=C:\cacert.pem
-  - bundle env
-
-build_script:
+  - SET BUNDLE_WITHOUT=server:docgen:maintenance:pry:travis:integration:ci
+  - bundle config --local path vendor/bundle # use the cache we define above
   - bundle install || bundle install || bundle install
 
+build: off
```
lamont-granquist: what does this do?

tas50: We're not doing a build anymore. Just a test so this skips that build phase

PR #6707, opened 2017-12-21
Unpin Chef and kitchen-dokken in the integration tests

tas50:
Cross your fingers. It would be nice to use the last year's worth of work.


lamont-granquist:
pretty sure the chef-config line was just to work around building chef 13 before it shipped and should have been removed a long time ago.  at the same time its just the chef that is launching test-kitchen so its not critical to testing.

tas50:
Closing this in favor of a PR to do everything but dokken for now

PR #6709, opened 2017-12-21
Remove lock files and test github masters in Kitchen Tests

tas50:
Remove deps we don't actually use
Pull things we control from master
Remove the Berkshelf and Gem lock files so we test the latest things available. Otherwise we're testing things from months ago, which isn't what customers would use or what we ship in ChefDK.


kitchen-tests/Gemfile
```diff
@@ -1,13 +1,9 @@
 source "https://rubygems.org"
 
-gem "chef-config", "< 13"
 gem "rake" # required to build some native extensions
 gem "berkshelf"
```
lamont-granquist: you could try pulling berks from master here

tas50: Do we need to do that or *should* we just to test that

lamont-granquist: its a _should_ not a _need_, although technically i think we should nuke the Gemfile.lock and pull from git master on everything we control (unless there's some issue requiring pinning)

tas50:
@lamont-granquist this work for you now?

PR #6711, opened 2017-12-21
Don't check both platform_family / os in provides when platform_family will do

tas50:
We're already going specific enough here to identify the platform_family. Debian is always going to be on Linux and SmartOS on Solaris2. Don't both checking both.

### DIFF
--- a/lib/chef/provider/apt_preference.rb
+++ b/lib/chef/provider/apt_preference.rb
@@ -24,7 +24,7 @@ require "chef/log"
 class Chef
   class Provider
     class AptPreference < Chef::Provider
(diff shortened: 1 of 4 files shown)


tas50:
@lamont-granquist if you have some knowledge of why this is the way it is please drop a knowledge bomb on me

lamont-granquist:
no reason at all.  smartos should probably be "platform" its a one-off.  although it doesn't matter either way.

tas50:
Yep I can change that up. All the solaris derivatives just 1:1 match platform and platform_family.

PR #6717, opened 2017-12-30
Improve property warnings in resources

tas50:
Give more details on why people are seeing these problems so they can fix them.


thommay:
This is great, thanks Tim!

PR #6725, opened 2018-01-02
Add a warning that Chef 11 server support in knife user is deprecated

tas50:
We're currently shipping 2 of each of these plugins and rewriting the user input when we think they actually meant to support Chef 11 server. It's code that really needs to go away at this point.


thommay:
I don't think it will be removed in 14.

tas50:
15?

thommay:
maaaybe? let's say Yes.

tas50:
Updated to Chef 15 in April 2019.

PR #6736, opened 2018-01-05
Add dhparam, rsa_private_key and rsa_public_key resources

tas50:
Ported from the openssl cookbook. I've done all the major refactoring there to shake the bugs out. This is just reformatted to use the mixin instead of a cookbook helper and to be library style so it works in core chef.


lib/chef/mixin/openssl.rb
```diff
@@ -0,0 +1,118 @@
...
+        cert_file_path, cert_filename = ::File.split(cert_filename)
+        cert_filename = ::File.basename(cert_filename, ::File.extname(cert_filename))
+        cert_file_path + ::File::SEPARATOR + cert_filename + ".key"
+      end
+
+      # is the key length a valid key length
+      # @param [Integer] number
+      # @return [Boolean] is length valid
+      def key_length_valid?(number)
+        number >= 1024 && number & (number - 1) == 0
```
stevendanna: [nit] It might help future readers of this code if we wrapped  `number & (number - 1) == 0` in some clarifying parens.

tas50: Fixed.

PR #6744, opened 2018-01-16
Fix yum_repository to properly reload chef's package cache post repo setup

tas50:
Without this you cannot install a package that is in that repository. We already did this for the yum cache, but not  for the DNF cache.

### DIFF
--- a/lib/chef/provider/yum_repository.rb
+++ b/lib/chef/provider/yum_repository.rb
@@ -47,7 +47,7 @@ class Chef
           if new_resource.make_cache
             notifies :run, "execute[yum clean metadata #{new_resource.repositoryid}]", :immediately if new_resource.clean_metadata || new_resource.clean_headers
             notifies :run, "execute[yum-makecache-#{new_resource.repositoryid}]", :immediately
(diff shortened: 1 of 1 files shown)


lib/chef/provider/yum_repository.rb
```diff
@@ -79,10 +84,10 @@ def load_current_resource
 
         declare_resource(:file, "/etc/yum.repos.d/#{new_resource.repositoryid}.repo") do
           action :delete
-          notifies :create, "ruby_block[yum-cache-reload-#{new_resource.repositoryid}]", :immediately
+          notifies :create, "ruby_block[package-cache-reload-#{new_resource.repositoryid}]", :immediately
         end
 
-        declare_resource(:ruby_block, "yum-cache-reload-#{new_resource.repositoryid}") do
+        declare_resource(:ruby_block, "package-cache-reload-#{new_resource.repositoryid}") do
           block { Chef::Provider::Package::Yum::YumCache.instance.reload }
```
lamont-granquist: what about these?

tas50: That would certainly help

PR #6745, opened 2018-01-16
fix non-daemonized umask

lamont-granquist:
We are ignoring this setting in non-daemonized (--no-fork) runs.


lamont-granquist:
I think I'm going to omit testing on this one because the code fix is so simple while the tests would be either ugly or fairly useless...

Issue #6623, opened 2017-12-06
windows_task: break sec_to_duration logic into a mixin


tas50:
Windows loves ISO8601. We should move this into a mixin so we can use it elsewhere and so that authors can use it in their cookbooks. There's no reason to confine this code to just the windows_task resource.

dheerajd-msys:
Thanks @tas50 , Updated the code as per your comments.

dheerajd-msys:
@tas50 Please close this issue as this is no longer required. Thanks.

PR #6747, opened 2018-01-17
[MSYS-733]Move sec_to_dur method to chef\mixin

dheerajd-msys:
### Description

Move `sec_to_duration` logic from windows_task resource to a mixin

### Issues Resolved

https://github.com/chef/chef/issues/6623


tas50:
@dheerajd-msys Sorry for the work here, but now that we're just using the gem to handle the logic I don't think we actually  need to move this into a mixin. Before it was about a page of pure logic that could potentially be used elsewhere. Now it's just a simple method. The previous refactoring was all that was needed. I'm going to close this out at this point.

PR #6748, opened 2018-01-17
use TMPDIR correctly in initscript

lamont-granquist:
lets TMPDIR be something other than /tmp


spec/functional/resource/bff_spec.rb
```diff
@@ -1,6 +1,6 @@
 #
 # Author:: Prabhu Das (<prabhu.das@clogeny.com>)
-# Copyright:: Copyright 2013-2016, Chef Software Inc.
+# Copyright:: Copyright 2013-2017, Chef Software Inc.
```
tas50: Missed a year

PR #6751, opened 2018-01-18
Update release notes for 13.7

thommay:


tas50:
We deprecated use inline resources and managing chef 11 server with knife. We also should mention the CVE fixed by ruby 2.4.3

tas50:
We also should mention the CVE in OpenSSL we bumped

PR #6753, opened 2018-01-19
Remove erl_call and deploy resources

tas50:
We deprecated these for removal in Chef 14.


thommay:
tenor-82109686

PR #6765, opened 2018-01-22
Convert node map to last-writer-wins for ties

lamont-granquist:
We still bind preferentially by specificity, but for ties with
specificity we now prefer last-writer-wins instead of the alphabetic
by class name method we had before (that was never documented and I'm
pretty certain nobody understood).

So now `provides(:whatever) { true }` in a cookbook should always win
over core.


tas50:
You have some busted integration specs

PR #6766, opened 2018-01-22
Fail on interval runs on windows

lamont-granquist:
closes #4824


lamont-granquist:
appveyor is passing everything but failing

PR #6775, opened 2018-01-23
speed up http func tests

lamont-granquist:
For the truncated downloads the tiny server needs to hit its
RequestTimeout in order to terminate the truncated request, so that
is set to 1 second.  Then also drop down the retries and the
rest_timeout on the client side (but keep the rest_timeout to 2 to
keep it above the tiny_server timeout)

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -5,7 +5,7 @@
 # Author:: Christopher Brown (<cb@chef.io>)
 # Author:: Christopher Walters (<cw@chef.io>)
 # Author:: Daniel DeLeo (<dan@chef.io>)
-# Copyright:: Copyright 2009-2017, Chef Software Inc.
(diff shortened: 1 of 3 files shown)


lamont-granquist:
cuts the func tests down to 6 mins from 12 mins.

and the failures on travis are all red on master, nothing to do with this.

lamont-granquist:
also on my desktop this gets a full `rake spec` down to 7 mins 55 seconds, which is almost respectable.

lamont-granquist:
note that if we start getting timeout failures in these tests we can look at bumping the Chef::Config values up some

lamont-granquist:
moving the tiny_server stop/start to `before/after(:all)` blocks from `:each` blocks should help as well (and probably mitigate timeouts due to startup time)

lamont-granquist:
this is as green as this one gets i think.

PR #6776, opened 2018-01-23
Revert "Fail on interval runs on windows"

lamont-granquist:
Reverts chef/chef#6766


lamont-granquist:
broke appveyor for some reason.

PR #6778, opened 2018-01-23
update immutable API blacklist and whitelist

lamont-granquist:
adds the whitelist so that when we update to new ruby versions it forces
us to audit the new Array and Hash methods and figure out if they are
mutators or not.

this change closes the Hash#store loophole that was being abused by
some cookbook consumers to directly modify merged ImmutableHashes
(producing inherently undefined behavior that we 'broke' at one
point).


lamont-granquist:
@chef/client-core this could use some more eyes on it to see if i accidentally put any mutator methods in the whitelist

lamont-granquist:
once again, this is as green as master currently is.

PR #6781, opened 2018-01-24
Raise fatal error If FQDN duplicated

linyows:
### Description

If fqdn is duplicated, it raises fatal error. Because it sometimes does not notice that it is duplicated.

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -183,7 +183,7 @@ class Chef
           end
           exit 10
         end
-        if %i(warn fatal).include?(config[:duplicated_fqdns])
+        if %i{warn fatal}.include?(config[:duplicated_fqdns])
           fqdns = list.map { |v| v[0] }
           if fqdns.count != fqdns.uniq.count
             duplicated_fqdns = fqdns.uniq
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -177,11 +177,11 @@ class Chef
           end
           exit 10
         end
-        ips = list.map { |v| v.first }
-        if ips.count != ips.uniq.count
-          duplicated_ips = ips.uniq
-          ui.fatal("SSH #{duplicated_ips.count > 1 ? 'nodes are' : 'node is'} " +
-                   "duplicated: #{duplicated_ips.join(',')}")
+        fqdns = list.map { |v| v[0] }
+        if fqdns.count != fqdns.uniq.count
+          duplicated_fqdns = fqdns.uniq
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I tend to lean towards this being a problem with the nodes in the database with the duplicated fqdns and those need to be cleaned up.  I don't think this is the responsibility of knife ssh.

lamont-granquist:
It should also be some long option to `knife ssh` like `--error-on-duplicated-fqdns` or something like that, which is off by default (and `--skip-on-duplicated-fqdns` would the other obvious behavior to support -- maybe `--duplicated-fqdns=(skip|error)` or something like that)

linyows:
I think that's true. So what if you do not exit with warnings only?

lamont-granquist:
if you're going to go down this road i think you should do what i suggested and make it a command line option, and have different behaviors - skip / error / warn and have the default be the existing behavior.

linyows:
I understand that it should be added as such option in order not to change existing specifications.

thommay:
Can we clarify _why_ there are duplicate fqdns here? Is this just bad data in the chef database, in which case we probably should not work around the problem. Or is there a genuine usecase?

linyows:
I would like to solve that is not possible to execute chef-client as expected for duplicated fqdn on node files with [knife-zero](https://github.com/higanworks/knife-zero). Because knife-zero manages nodes as static file. So chef database does not matter.

lib/chef/knife/ssh.rb
```diff
@@ -123,6 +123,11 @@ class Ssh < Knife
         :boolean => true,
         :proc => Proc.new { :raise }
 
+      option :skip_on_duplicated_fqdns,
+        :long => "--skip-on-duplicated-fqdns",
+        :description => "Skip exits even if FQDNs are duplicated",
+        :boolean => true
```
lamont-granquist: this is still the incorrect default here.  it needs to be off by default.  it should also be expanded to support different behaviors:  `:ignore`, `:warn`, `:fatal`, `:skip`

linyows: I got it.

thommay:
@linyows this is looking good. We're gonna need you to do two things, and then we can get this merged. 
First is to fix the chefstyle failures: https://travis-ci.org/chef/chef/jobs/340760026 . You can do this by running `bundle exec chefstyle -a`.
Second is to add a DCO sign off. If you've got the chefdk installed, run `dco sign` in your branch.

linyows:
Ok, thanks.

thommay:
Thanks for sticking with this, and thanks for your contribution to Chef!

Issue #4959, opened 2016-05-20
Wrong mount defaults on FreeBSD (/etc/fstab: Inappropriate file or format)


sekrett:
## Description

/etc/fstab must de written in a different way on FreeBSD
## Chef Version

12.10.24
## Platform Version

bento/FreeBSD-10.3
## Replication Case

```
directory '/source'
directory '/destination'
mount '/destination' do
  device '/source'
  fstype 'nullfs'
  pass 0
  action [:mount, :enable]
end
```
## Client Output

It adds:
`/source /destination nullfs defaults 0 0`

This does not mount after reboot. It should be:
`/source /destination nullfs rw 0 0`

So `defaults` does not work on FreeBSD and must be replaced.
## Workaround

Add `options 'rw'`.

cramaechi:
I think the substitution should take place in `.../lib/chef/provider/mount/mount.rb`:

<img width="1204" alt="screen shot 2018-01-21 at 3 31 02 am" src="https://user-images.githubusercontent.com/24444174/35192769-c091fc02-fe5d-11e7-8a5f-7abd67f8c312.png">

My only problem now is figuring out whether this change affects all platforms.

cheeseplus:
Fixed by #6782

PR #6782, opened 2018-01-24
Update /etc/fstab on FreeBSD #4959

cramaechi:
### Description

/etc/fstab contains: /source /destination nullfs defaults 0 0

This does not mount after reboot. It should be:
/source /destination nullfs rw 0 0

So "defaults" does not work on FreeBSD and must be replaced.

Signed-off-by: Chibuikem Amaechi <cramaechi@me.com>

### Issues Resolved

No preexisting issue(s).

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/mount/mount.rb
+++ b/lib/chef/provider/mount/mount.rb
@@ -158,7 +158,7 @@ class Chef
             disable_fs
           end
           ::File.open("/etc/fstab", "a") do |fstab|
-            fstab.puts("#{device_fstab} #{@new_resource.mount_point} #{@new_resource.fstype} #{@new_resource.options.nil? ? "defaults" : @new_resource.options.join(",")} #{@new_resource.dump} #{@new_resource.pass}")
+            fstab.puts("#{device_fstab} #{@new_resource.mount_point} #{@new_resource.fstype} #{@new_resource.options.nil? ? "rw" : @new_resource.options.join(",")} #{@new_resource.dump} #{@new_resource.pass}")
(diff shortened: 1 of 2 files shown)


thommay:
Thanks for the PR. In this case, I think what is required is a specific selection so that we can still use `defaults` on linux (and anywhere else that supports that syntax, if anywhere), and something more specific for freebsd.
So I'd recommend making this a function and then returning the correct value based on `node[:os]`.

cramaechi:
Hi @thommay - I was just about to comment on that very point.

Thank you helpful notes!

cramaechi:
Pull request has been updated.

cramaechi:
Updated pull request with unit tests.

lib/chef/provider/mount/mount.rb
```diff
@@ -146,6 +146,15 @@ def remount_fs
           end
         end
 
+        # Return appropriate default mount options according to the input os.
+        def default_mount_options(os)
+          if os == "freebsd"
+            "rw"
+          else
+            "defaults"
+          end
```
lamont-granquist: you can just make this more concisely:

```ruby
def default_mount_options
  node[:os] == "freebsd" ? "rw" : "defaults"
end
```

thommay: Actually we think _only_ linux should be `defaults` and everything else is `rw`. So let's flip that logic around.

cramaechi: @lamont-granquist, @thommay - The pull request has been updated.

Thanks for the help!

PR #6783, opened 2018-01-24
Update Ohai, omnibus, and Ruby versions

tas50:
ruby to resolve 2 CVEs
ohai to 8.26.1 to fix docker detection and ec2 detection on new instance
types
omnibus to fix issues with Docker images of chef


tas50:
The powershell test is the same failure we see in chef 14 branch right now. Something changed on the appveyor side probably.  The ssl failures on test kitchen are a known issue I'm looking into right now

thommay:
We'll need to bump omnibus for the cacerts fixes.

tas50:
Powershell failures in appveyor are expected. We need to update out tests for the new version of PS they're shipping.

PR #6793, opened 2018-01-25
Add ohai_hint resource from ohai cookbook

tas50:
This makes it easier to drop off a hint and reload ohai w/o bringing in cookbook dependencies.


lib/chef/resource/ohai_hint.rb
```diff
@@ -0,0 +1,96 @@
...
+      introduced "14.0"
+
+      property :hint_name,
+               String,
+               description: "The name of hints file if different from the resource name",
+               name_property: true
+
+      property :content,
+               Hash,
+               description: "A Hash of values to include in the hint file"
```
thommay: > Data to include in the hint file

We already know it's a hash, and our docs should show the type and the description.

lib/chef/resource/ohai_hint.rb
```diff
@@ -0,0 +1,96 @@
...
+# limitations under the License.
+#
+
+class Chef
+  class Resource
+    class OhaiHint < Chef::Resource
+      resource_name :ohai_hint
+      provides :ohai_hint
+
+      description "Use the ohai_hint resource to pass hint data to Ohai to aid in configuration detection."
```
thommay: Don't think we should repeat the name of the resource in the description. We could probably use some wordsmithing and advice from @brewn here. (And on the rest of these).

tas50: This goes off the format on the docs site. Here's an example:

Use the apt_repository resource to specify additional APT repositories. Adding a new repository will update APT package cache immediately.

tas50:
@thommay Does this work with the wording fixes?

PR #6794, opened 2018-01-26
Fix "knife ssh" authentication scheme #4342

### LINKED ISSUE #4342: knife ssh fails with --ssh-password if ~/.ssh/id_rsa does not exist
Affects at least knife 12.5.1

Using `knife ssh` in "password, not key" mode fails (though it shows only a warning):

```
% knife ssh 'netgroups:mtc' 'ls /' --ssh-user jblaine --ssh-password
Enter your password:
WARNING: Failed to connect to hosthere.our.org -- Errno::ENOENT: No such file or directory @ rb_sysopen - /afs/rcf/user/jblaine/.ssh/id_rsa

...20 other hosts with same output here...
```

cramaechi:
### Description

Affects at least knife 12.5.1

When a user uses knife ssh in "password, not key" mode, it fails.

Signed-off-by: Chibuikem Amaechi <cramaechi@me.com>

### Issues Resolved

closes #4342 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -560,6 +560,11 @@ class Chef
           config[:ssh_password] = get_stripped_unfrozen_value(ssh_password ||
                              Chef::Config[:knife][:ssh_password])
         end
+
+        # CHEF-4342 Diable host key verification if a password has been given.
+        if config[:ssh_password]
+          config[:host_key_verify] = false
+        end
       end
 
       def configure_ssh_identity_file
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -560,11 +560,6 @@ class Chef
           config[:ssh_password] = get_stripped_unfrozen_value(ssh_password ||
                              Chef::Config[:knife][:ssh_password])
         end
-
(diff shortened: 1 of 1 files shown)


cramaechi:
If `--no-host-key-verify` option is absent from the command line and a password is eventually specified, I believe host key verification should be disabled manually by setting `config[:host_key_verify]` to false as soon as a password is given via command line, user prompt, or knife.rb. It bears mentioning that host key verification is enabled by default.

I think the reason why`knife ssh` is issues a warning when a host connection failure occurs is because it is operating under the assumption that more than one host is trying to be connected to.

thommay:
hi @cramaechi I think you've misunderstood the bug here. When a password is specified, we need to not check for (and fail) the user's ssh key, not the *host* ssh key.
We absolutely must check the host ssh key on every connection attempt because this is a key part of ssh's protection against man in the middle attacks.

cramaechi:
Hi @thommay - Please excuse my failure to fully understand the issue :grimacing:. 

Though I couldn't fully replicate the user's problem, I have updated the pull request with changes that I think should remediate the issue.

Please share your thoughts.

thommay:
Fixes: #4342

PR #6795, opened 2018-01-26
Add hostname resource from chef_hostname cookbook

tas50:
Add in a new resource for managing hostnames


lib/chef/resource/hostname.rb
```diff
@@ -0,0 +1,229 @@
...
+            not_if { IO.read(path).split("\n").include?(string) }
+          end
+        end
+
+        # read in the xml file used by Ec2ConfigService and update the Ec2SetComputerName
+        # setting to disable updating the computer name so we don't revert our change on reboot
+        # @return [String]
+        def updated_ec2_config_xml
+          begin
+            require "nokogiri"
```
lamont-granquist: so nokogiri is evil and this should really be rexml

lib/chef/resource/hostname.rb
```diff
@@ -0,0 +1,252 @@
...
+      provides :hostname
+      resource_name :hostname
+
+      description "Sets the systems hostname, ensures that reboot will preserve the hostname, and re-runs the ohai plugin so the hostname will be available in subsequent cookbooks."
+      introduced "14.0"
+
+      property :hostname,
+               String,
+               description: "The hostname if different than the resource's name",
+               name_property: true
```
lamont-granquist: what about slightly more compactly?

```ruby
property :hostname, String, name_property: true,
   description: "The hostname if different than the resource's name"
```

PR #6799, opened 2018-01-26
Ensure that we create a docker compatible ca-certs symlink

thommay:
(let's hold off on merging this until the master build is verified)


tas50:
Master passed and this resolved the issue for me in cookbooks pointing at current

docwhat:
Can 13.7.16 be removed from the list of available chef packages?  I've spent a day trying to track this down in my dokken instance.

PR #6806, opened 2018-01-29
Add MSIFASTINSTALL property, supported by Windows Installer 5.0

stuartpreston:

### Description

This change adds an MSIFASTINSTALL property to the Windows Installer set with a value that skips most of the costings, when used on a system that supports Windows Installer 5.0 (i.e. Windows 2008 R2 upwards) [ref](https://msdn.microsoft.com/en-us/library/windows/desktop/aa372856(v=vs.85).aspx).  It should result in a slight speed improvement on install/upgrade.

### Check List

- [ ] All tests pass

### DIFF
--- a/omnibus/Gemfile.lock
+++ b/omnibus/Gemfile.lock
@@ -9,9 +9,9 @@ GIT
 
 GIT
   remote: https://github.com/chef/omnibus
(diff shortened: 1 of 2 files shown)


thommay:
Do you also need to bump omnibus version to pick up the new fastinstall lib?

PR #6714, opened 2017-12-26
[MSYS-735]fix validation for idle_time

dheerajd-msys:
### Description
Validate if  `idle_time` is not provided or provided outside range `1-999` for `frequency :on_idle`

PR #6807, opened 2018-01-29
Fix windows_task idle_time validation

algaut:
### Description

Two improvements on the idle_time validation function for the windows_task resource:
- Compare frequency to :idle_time instead of comparing to an incomplete list of other frequencies.
- Keep all the validation logic in the function instead of splitting it between the function and where the function is called.

### Issues Resolved

I din't create an issue since it seemed like an easy fix, but the previous PR [#6714](https://github.com/chef/chef/pull/6714) on the matter had some of my windows_task resources fail when using the :onstart frequency.

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/windows_task.rb
+++ b/lib/chef/resource/windows_task.rb
@@ -82,7 +82,7 @@ class Chef
         validate_create_frequency_modifier(frequency, frequency_modifier)
         validate_create_day(day, frequency) if day
         validate_create_months(months, frequency) if months
-        validate_idle_time(idle_time, frequency) if ( !idle_time.nil? && ([:minute, :hourly, :daily, :weekly, :monthly].include? frequency)) || (idle_time.nil? || !(idle_time > 0 && idle_time <= 999)) && !([:minute, :hourly, :daily, :weekly, :monthly].include? frequency)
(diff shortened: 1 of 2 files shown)


algaut:
Closed for sign-off

tas50:
@algaut Thanks for taking the time to make sure this works in additional scenarios. Do you think you could take a peak at the current specs and add a spec or two for the scenarios you think this fixes. We're looking at doing some large scale refactoring of this codebase and without specs to handle each scenario there's a good chance we'll introduce a regression here in the future.

-Tim

algaut:
Hi Tim,
Would some specs like those work for you? Not sure if I'm supposed to use simple logic during spec definitions, but this should cover most cases for what I changed

tas50:
@algaut I think those specs will certainly help us to avoid a regression here. Thanks. Any chance you can squash this down to a single commit and rebase it off master. We fixed some testing failures since you opened this up originally and that would let us get to green in appveyor testing.

tas50:
Fixes https://github.com/chef/chef/issues/6824

algaut:
Should be good to go now @tas50

tas50:
Thanks @algaut

PR #6808, opened 2018-01-30
Fix Appveyor testing:  the format of this flametest doesn't matter

lamont-granquist:
we don't need to test the exact output of this powershell command, just
get literally anything that looks like it ran successfully.


lamont-granquist:
there's still rubygems issues on appveyor + the epic_fail unit test.

Issue #6639, opened 2017-12-08
windows_task: switch from schtasks CLI to XML


tas50:
If we directly manipulate the XML we get full control of the tasks. We should investigate what would be necessary to skip the CLI entirely

spuder:
Be aware that the XML must be encoded as UTF16 (Little endian), which means you must be very careful when editing in your typical text editor, or you will get error

    The format of the task is not valid. The following error was reported: (1,2)::ERROR: one root element

https://github.com/atom/atom/issues/8219

Using the XML would be nice because there is no way to do Custom Event filters with schtasks.exe

Today the work around is to use a template (saved as UTF16) and shell out to schtask.exe


```ruby
powershell_script 'TaskImportApplicationError' do
  guard_interpreter :powershell_script
  code <<-EOH
  schtasks.exe /create /TN 'Application Error Alert' /RU SYSTEM /XML C:\\tools\\schedtasks\\applicationError.xml
  EOH
  not_if "schtasks.exe /Query /TN 'Application Error Alert'"
end
```

Issue #6635, opened 2017-12-08
windows_task: creation fails when run twice in latest current build on non-US system


tas50:
Our latest and greatest build (13.6.26) still fails if you try to create the following task and run chef twice on a non-US date formatted system:

```ruby
windows_task 'chef-client' do
  user 'SYSTEM'
  command 'chef-client'
  run_level :highest
  frequency :minute
  frequency_modifier 15
  execution_time_limit '5000'
end
```

**The error:**
```
-----> Converging <task-us-windows-2016>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 6.3.1...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
-----> Chef Omnibus installation detected (install only if missing)
       Transferring files to <task-us-windows-2016>
       [2017-12-07T15:44:50-08:00] INFO: Started chef-zero at chefzero://localhost:1 with repository at C:/Users/vagrant/AppData/Local/Temp/kitchen, C:/Users/vagrant/AppData/Local/Temp/kitchen
         One version per cookbook

       Starting Chef Client, version 13.6.26
       [2017-12-07T15:44:50-08:00] INFO: *** Chef 13.6.26 ***
       [2017-12-07T15:44:50-08:00] INFO: Platform: x64-mingw32
       [2017-12-07T15:44:50-08:00] INFO: Chef-client pid: 4244
       [2017-12-07T15:44:50-08:00] INFO: The plugin path C:\chef\ohai\plugins does not exist. Skipping...
       [2017-12-07T15:44:56-08:00] INFO: Setting the run_list to ["recipe[task_test::us]"] from CLI options
       [2017-12-07T15:44:56-08:00] INFO: Run List is [recipe[task_test::us]]
       [2017-12-07T15:44:56-08:00] INFO: Run List expands to [task_test::us]
       [2017-12-07T15:44:56-08:00] INFO: Starting Chef Run for task-us-windows-2016
       [2017-12-07T15:44:56-08:00] INFO: Running start handlers
       [2017-12-07T15:44:56-08:00] INFO: Start handlers complete.
       resolving cookbooks for run list: ["task_test::us"]
       [2017-12-07T15:44:56-08:00] INFO: Loading cookbooks [task_test@0.1.0]
       Synchronizing Cookbooks:
       [2017-12-07T15:44:56-08:00] INFO: Storing updated cookbooks/task_test/recipes/us.rb in the cache.
[...]

Issue #6622, opened 2017-12-06
windows_task: execution_time_limit property should also accept integers


tas50:
Execution time limit takes seconds as a string. This seems really odd. We even have to later check to see if the string value .to_i equals 0. We should prefer Integers here and document it as an integer. Everyone will be better off.

tas50:
Fixed by https://github.com/chef/chef/pull/6628

tas50:
Fixed in 13.6.30, which will be shipped as 13.7.X

Issue #6615, opened 2017-12-04
invalid date error on windows_task with frequency :on_logon


sarahbakal:
## Description

invalid date error on windows_task with frequency :on_logon

## Chef Version

Chef-client 13.6.24

## Platform Version

Windows Server 2016
Chef-client 13.6.24

## Replication Case

Windows 2016, created the task on the first run and then deleted the trigger to make sure it would re-apply the :on_logon trigger

## Client Output
```
-----> Converging <default-win2016-standard>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 6.3.1...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
-----> Chef Omnibus installation detected (install only if missing)
       Transferring files to <default-win2016-standard>
       Starting Chef Client, version 13.6.24
       resolving cookbooks for run list: ["compliance_standard_baseline::default"]
       Synchronizing Cookbooks:
         - compliance_standard_baseline (23.19.0)
         - sysctl (2.0.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       Converging 188 resources
       Recipe: compliance_standard_baseline::windows2016
         * windows_task[chef-client-hku] action create
           
           ================================================================================
           Error executing action `create` on resource 'windows_task[chef-client-hku]'
           ================================================================================
           
           ArgumentError
           -------------
           invalid date
           
           Resource Declaration:
           ---------------------
           # In C:/Users/vagrant/AppData/Local/Temp/kitchen/cache/cookbooks/compliance_standard_baseline/recipes/windows2016.rb
           
             7:   windows_task 'chef-client-hku' do
             8:     task_name 'chef-client-hku'
             9:     command 'cmd /c C:/opscode/chef/bin/chef-client'
            10:     frequency :on_logon
            11:     action :create
            12:     user 'SYSTEM'
            13:     run_level :highest
            14:   end
            15: # end
           
           Compiled Resource:
           ------------------
           # Declared in C:/Users/vagrant/AppData/Local/Temp/kitchen/cache/cookbooks/compliance_standard_baseline/recipes/windows2016.rb:7:in `from_file'
           
           windows_task("chef-client-hku") do
             action [:create]
             default_guard_interpreter :default
             task_name "chef-client-hku"
             declared_type :windows_task
             cookbook_name "compliance_standard_baseline"
             recipe_name "windows2016"
[...]

PR #6815, opened 2018-01-31
[MSYS-752] windows task rewrite using win32-taskscheduler

Vasu1105:
### Description
windows_task resource rewrite using win32-taskscheduler library.

### Issues Resolved
| Issue                                    | Fixed                 | Notes                                                           |
| ----------------------------------  | -------------------  | ------------------------------------------------------ |
|[#490]( https://github.com/chef-cookbooks/chef-client/issues/490)| yes tested | Tested chef-clien::task cookbook with Winodws_task rewrite |
|MSYS-738 #6638 | yes tested | Tested recipes mentioned in card with windows_task rewrite |
MSYS-737 #6635  | yes tested | Tested recipes mentioned in card |  
MSYS-735 #6807  | yes tested | Tested with on_idle and idle_time  |
MSYS-734 #6622  | yes tested | Tested execution_time_limit as string and integer values |  
MSYS-731 | yes tested | Tested recipes mentioned in card for idempotency |  
MSYS-730 #6615  | yes tested | This will not cause an issue in Chef 14 rewrite release since the now user can set the start_day for :on_idle, :on_logon and :onstart event. |  
#6639 | yes  | windows_task is now using win32-taskscheduler library  |


lib/chef/provider/windows_task.rb
```diff
@@ -20,26 +20,153 @@
 require "rexml/document"
 require "iso8601"
 require "chef/mixin/powershell_out"
+require "chef/provider"
+require "win32/taskscheduler"
```
tas50: We need to get this dep into the gemspec

lib/chef/provider/windows_task.rb
```diff
@@ -20,26 +20,153 @@
...
+require 'pry'
```
tas50: Probably want to remove this one

Vasu1105: yes the coiding is in progress so I will be removing those while cleaning up the unnecessary code

lib/chef/provider/windows_task.rb
Vasu1105: yes I will be committing those file

chef-universal-mingw32.gemspec
```diff
@@ -15,6 +15,7 @@ gemspec.add_dependency "win32-process", "~> 0.8.2"
 gemspec.add_dependency "win32-service", "~> 0.8.7"
 gemspec.add_dependency "windows-api", "~> 0.4.4"
 gemspec.add_dependency "wmi-lite", "~> 1.0"
+gemspec.add_dependency "win32-taskscheduler", "0.3.2"
```
btm: This should have a comparison rather than pin to a specific version, unless there is a specific reason we need only this version, in which case there needs to be a comment about why.

Vasu1105: Yes I will update that

lib/chef/resource/windows_task.rb
```diff
@@ -163,27 +185,44 @@ def validate_create_frequency_modifier(frequency, frequency_modifier)
...
+              unless ["mon", "tue", "wed", "thu", "fri", "sat", "sun", "*"].include?(d.strip.downcase)
+                raise ArgumentError, "day property invalid. Only valid values are: MON, TUE, WED, THU, FRI, SAT, SUN and *. Multiple values must be separated by a comma."
+              end
             end
           end
         end
       end
 
+      def days_includes_days_of_months?(days)
+        days.each do |d|
```
piyushawasthi: This method is very costly because if days have 5 element the below array [1..31, last] create 5 times. 

This method should be.
```
def days_includes_days_of_months?(days)
    days.all? { |day| default_array.include?(day) }
end

def default_array
   ("1".."31").to_a << "last"
end
```

lib/chef/resource/windows_task.rb
```diff
@@ -163,27 +185,44 @@ def validate_create_frequency_modifier(frequency, frequency_modifier)
...
+          if days_includes_days_of_months?(days)
+            if frequency == :weekly
+              raise "day values 1-31 or last is only valid with frequency :monthly"
+            end
+
+            if frequency_modifier && frequency_modifier != 1
+              raise "Do not use frequency_modifier property when day values set as 1-31 or last"
+            end
+          else
+            days.each do |d|
```
piyushawasthi: similarly below we don't need to iterate that.

lib/chef/provider/windows_task.rb
```diff
@@ -156,11 +176,12 @@ def action_delete
       def action_end
         if current_resource.exists
           Chef::Log.debug "#{new_resource} task exists"
-          if current_resource.status != :running
+          if current_resource.task.status != "running"
```
Nimesh-Msys: we may DRY up this status check thing. may include some methods like `running?` , `ready?`, etc and can the same elsewhere.

lib/chef/resource/windows_task.rb
```diff
@@ -94,22 +97,41 @@ def numeric_value_in_string?(val)
...
+          if  frequency_modifier_includes_days_of_weeks?(frequency_modifier)
+            unless day
+              raise ArgumentError, "Please select day on which you want to run the task e.g. 'Mon, Tue'. Multiple values should be commas seprated."
+            end
+          end
+        end
+      end
+
+      def frequency_modifier_includes_days_of_weeks?(frequency_modifier)
+        frequency_modifier.split(',').each do |value|
```
piyushawasthi: same here as below.

lib/chef/resource/windows_task.rb
```diff
@@ -94,22 +97,41 @@ def numeric_value_in_string?(val)
         false
       end
 
+      def validate_frequency_monthly(frequency_modifier, months, day)
```
samshinde: Can we avoid multiple if / if / unless here? We can write it in a single statement is it?

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
-      def run_schtasks(task_action, options = {})
-        cmd = "schtasks /#{task_action} /TN \"#{new_resource.task_name}\" "
-        options.each_key do |option|
-          unless option == "TR"
-            cmd += "/#{option} "
-            cmd += "\"#{options[option].to_s.gsub('"', "\\\"")}\" " unless options[option] == ""
-          end
+      def set_start_day_and_time
+        unless new_resource.start_day
+          new_resource.start_day = Time.now.strftime("%m/%d/%Y")
```
Nimesh-Msys: we may use `one-liner unless` here

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
         end
-        # Appending Task Run [TR] option at the end since appending causing sometimes to append other options in option["TR"] value
-        if options["TR"]
-          cmd += "/TR \"#{options["TR"]} \" " unless task_action == "DELETE"
+
+        unless new_resource.start_time
+          new_resource.start_time = Time.now.strftime("%H:%M")
```
Nimesh-Msys: one-liner unless can be used

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
-          false
-        elsif current_resource.random_delay.nil?
-          true
-        else
-          ISO8601::Duration.new(current_resource.random_delay) != ISO8601::Duration.new(new_resource.random_delay)
+      def trigger
+        date = new_resource.start_day.split('/') if new_resource.start_day
+        time = new_resource.start_time.split(':') if new_resource.start_time
+        #TODO currently end_month, end_year and end_year needs to be set to 0. If not set win32-taskscheduler throwing nil into integer error.
+        trigger_hash = {}
```
Nimesh-Msys: need not to be declared, we are defining it in next line,
Declaration is not necessary

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
-        if new_resource.random_delay.nil?
-          false
-        elsif current_resource.random_delay.nil? && new_resource.random_delay == "PT0S" # when user sets random_dealy to 0 sec
-          false
-        elsif current_resource.random_delay.nil?
-          true
-        else
-          ISO8601::Duration.new(current_resource.random_delay) != ISO8601::Duration.new(new_resource.random_delay)
+      def trigger
+        date = new_resource.start_day.split('/') if new_resource.start_day
```
Nimesh-Msys: date = new_resource.start_day? ? new_resource.start_day.split('/') : []
This will optimize below code @ L#254-L#256.
Similarly for time

lib/chef/resource/windows_task.rb
```diff
@@ -94,22 +97,41 @@ def numeric_value_in_string?(val)
...
       # @todo when we drop ruby 2.3 support this should be converted to .match?() instead of =~f
       def validate_start_day(start_day, frequency)
-        if [:once, :on_logon, :onstart, :on_idle, :none].include? frequency
+        if start_day && [:on_logon, :onstart, :on_idle, :none].include?(frequency)
           raise ArgumentError, "`start_day` property is not supported with frequency: #{frequency}"
         end
 
         # make sure the start_day is in MM/DD/YYYY format: http://rubular.com/r/cgjHemtWl5
-        raise ArgumentError, "`start_day` property must be in the MM/DD/YYYY format." unless /^(0[1-9]|1[012])[- \/.](0[1-9]|[12][0-9]|3[01])[- \/.](19|20)\d\d$/ =~ start_day
+        if start_day
```
piyushawasthi: I this this should be a method for valid date format
```
require 'date'
def date_valid?(start_day)
  format = '%m-%d-%Y'
  DateTime.strptime(date, format)
  true
rescue ArgumentError
   "`start_day` property must be in the MM/DD/YYYY format."
end
```

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+        time = new_resource.start_time.split(':') if new_resource.start_time
```
Nimesh-Msys: start_time = new_resource.start_time || []
[ start_hour, start_min ] = start_day.split(':')
then use these variables in L#257 & L#258

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+        #TODO currently end_month, end_year and end_year needs to be set to 0. If not set win32-taskscheduler throwing nil into integer error.
+        trigger_hash = {}
+        trigger_hash = {
+          start_year: date.nil? ? 0 : date[2].to_i,
```
Nimesh-Msys: direct variables can be used, as explained in L# 249

lib/chef/resource/windows_task.rb
```diff
@@ -17,6 +17,7 @@
 #
 
 require "chef/resource"
+require 'pry'
```
piyushawasthi: This should not be here.

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+          start_hour: time.nil? ? 0 : time[0].to_i,
+          start_minute: time.nil? ? 0 : time[1].to_i,
+          end_month: 0,
+          end_day: 0,
+          end_year: 0,
+          trigger_type: trigger_type,
+          type: type,
+          random_minutes_interval: new_resource.random_delay
+        }
+        if new_resource.frequency == :minute
```
Nimesh-Msys: new_resource.frequency == :minute,
these conditions may directly be checked in convert_in_minutes method,
possible DRY up

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
-        date_format.sub!("yy", "%y")
-        date_format
-      end
+      #TODO : Try to optimize this method
+      # known issue : Since start_day and time is not mandatory while updating weekly frequency for which start_day is not mentioned by user idempotency
+      # is not gettting maintained as new_resource.start_day is nil and we fetch the day of week from start_day to set and its currently coming as nil and don't match with current_task
+      def task_needs_update?(task)
+        flag = false
+        current_task_trigger = task.trigger(0)
+        new_task_trigger = trigger
```
Nimesh-Msys: `trigger` is undefined. Pls check if need `current_task_trigger` here

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+        flag = true if start_day_updated?(current_task_trigger, new_task_trigger) == true ||
+        start_time_updated?(current_task_trigger, new_task_trigger) == true ||
+        current_task_trigger[:trigger_type] != new_task_trigger[:trigger_type] ||
+        current_task_trigger[:type] != new_task_trigger[:type] ||
+        current_task_trigger[:random_minutes_interval].to_i != new_task_trigger[:random_minutes_interval].to_i ||
+        current_task_trigger[:minutes_interval] != new_task_trigger[:minutes_interval] ||
+        task.account_information != new_resource.user ||
+        task.max_run_time != new_resource.execution_time_limit.to_i * 1000 ||
+        task.application_name != new_resource.command ||
+        task.principals[:run_level] != run_level
```
Nimesh-Msys: L#298 - L#312,
all the conditions may be checked by using &&
In case of any false, it will be returned then and there
Hence 'flag' variable may not be required

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+
+        flag
+      end
+
+      def start_day_updated?(current_task_trigger, new_task_trigger)
+        if new_resource.start_day
+          return true if current_task_trigger[:start_year].to_i != new_task_trigger[:start_year] ||
```
Nimesh-Msys: we may use a 'flag' here. Hence we can avoid 'return' and else.
may also use something like:
def start_day_updated?(current_task_trigger, new_task_trigger)
  (new_resource.start_day && (
      current_task_trigger[:start_year].to_i != new_task_trigger[:start_year] ||
      current_task_trigger[:start_month].to_i != new_task_trigger[:start_month] ||
      current_task_trigger[:start_day].to_i != new_task_trigger[:start_day] ))
end

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+        else
+          false
+        end
       end
 
-      def convert_system_date_to_mm_dd_yyyy(system_date)
-        system_date_format = convert_system_date_format_to_ruby_date_format
-        unless system_date_format == "%m/%d/%Y"
-          system_date = Date.strptime(system_date, system_date_format).strftime("%m/%d/%Y")
+      def start_time_updated?(current_task_trigger, new_task_trigger)
```
Nimesh-Msys: many conditions and return are used

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+          when :once, :minute, :hourly
+            TaskScheduler::ONCE
+          when :daily
+            TaskScheduler::DAILY
+          when :weekly
+            TaskScheduler::WEEKLY
+          when :monthly
+            # If frequency modifier is set with frequency :monthly we are setting taskscheduler as monthlydow
+            # Ref https://msdn.microsoft.com/en-us/library/windows/desktop/aa382061(v=vs.85).aspx
+            if new_resource.frequency_modifier != 1
```
Nimesh-Msys: Ternary operator can be used

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+              TaskScheduler::MONTHLYDATE
+            end
+          when :on_idle
+            TaskScheduler::ON_IDLE
+          when :onstart
+            TaskScheduler::AT_SYSTEMSTART
+          when :on_logon
+            TaskScheduler::AT_LOGON
+          else
+            #TODO
```
Nimesh-Msys: raise ArgumentError with "Invalid Trigger Type" ?

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+          end
+      end
 
-        if new_resource.frequency == :none
-          doc.root.elements.delete(xml_element_mapping["random_delay"])
-          trigger_element = REXML::Element.new(xml_element_mapping["random_delay"])
-          doc.root.elements.add(trigger_element)
+      def type
+        if trigger_type == TaskScheduler::ONCE
```
Nimesh-Msys: Since all the checks are with trigger_type only, case statements can be used instead of ladder

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+      end
 
-        task[:on_logon] = true if root.elements["Triggers/LogonTrigger"]
-        task[:onstart] = true if root.elements["Triggers/BootTrigger"]
-        task[:on_idle] = true if root.elements["Triggers/IdleTrigger"]
-
-        task[:idle_time] = root.elements["Settings/IdleSettings/Duration"].text if root.elements["Settings/IdleSettings/Duration"] && task[:on_idle]
+      # TODO handle run_level here
+      def run_level
+        case new_resource.run_level
```
Nimesh-Msys: ternary operator can be used

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
 
-        task[:none] = true if root.elements["Triggers/"] && root.elements["Triggers/"].entries.empty?
-        task[:once] = true if !(task[:repetition_interval] || task[:schedule_by_day] || task[:schedule_by_week] || task[:schedule_by_month] || task[:on_logon] || task[:onstart] || task[:on_idle] || task[:none])
-        task[:execution_time_limit] = root.elements["Settings/ExecutionTimeLimit"].text if root.elements["Settings/ExecutionTimeLimit"] #by default PT72H
-        task[:random_delay] = root.elements["Triggers/TimeTrigger/RandomDelay"].text if root.elements["Triggers/TimeTrigger/RandomDelay"]
-        task[:random_delay] = root.elements["Triggers/CalendarTrigger/RandomDelay"].text if root.elements["Triggers/CalendarTrigger/RandomDelay"]
-        task
+      #TODO: while creating the configuration settings win32-taskscheduler it accepts execution time limit values in ISO8601 formata
+      def config_settings
+        settings = {}
```
Nimesh-Msys: no need to declare this

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+        settings = {
```
Nimesh-Msys: pls use similar format.
```
settings = {}
settings[:execution_time_limit] = ISO8601::Duration.new(new_resource.execution_time_limit.to_i).to_s
settings[:enabled] = true
```
if we are using something like this, then above review comment can be ignored(of declaration)

lib/chef/provider/windows_task.rb
```diff
@@ -201,308 +224,303 @@ def action_disable
...
+          enabled: true
+        }
+        settings[:idle_duration] = new_resource.idle_time if new_resource.idle_time
+        settings[:run_only_if_idle] = true if new_resource.idle_time
+        settings
       end
 
-      SYSTEM_USERS = ['NT AUTHORITY\SYSTEM', "SYSTEM", 'NT AUTHORITY\LOCALSERVICE', 'NT AUTHORITY\NETWORKSERVICE', 'BUILTIN\USERS', "USERS"].freeze
+      def principal_settings
+        settings = {}
```
Nimesh-Msys: { run_level: run_level } would be enough in this method. It will return the desired hash

lib/chef/provider/windows_task.rb
Vasu1105: trigger is a method  call here

lib/chef/provider/windows_task.rb
Vasu1105: we don't need  to raise error here since we here since we are handling this in provider

lib/chef/resource/windows_task.rb
Vasu1105: I don't think this is required here since we are checking the string format since of date coming as string "mm/dd/yyyy" format and it's already working code.

lib/chef/resource/windows_task.rb
Vasu1105: I don't think that the days.all? { |d| } block is reducing the cost since that is also enumeration it may increase the readability.

Vasu1105: I used array set operation instead.

piyushawasthi: That's make sense, you can change `days.all? { |day| valid_days.include?(day) }` to `(days - valid_days).empty?`

lib/chef/resource/windows_task.rb
Vasu1105: fixed

lib/chef/resource/windows_task.rb
```diff
@@ -136,65 +154,72 @@ def password_required?(user)
...
+        raise "day property is only valid for tasks that run monthly or weekly" unless [:weekly, :monthly].include?(frequency)
+
         if day.is_a?(String) && day.to_i.to_s != day
           days = day.split(",")
-          days.each do |d|
-            unless ["mon", "tue", "wed", "thu", "fri", "sat", "sun", "*"].include?(d.strip.downcase)
-              raise ArgumentError, "day property invalid. Only valid values are: MON, TUE, WED, THU, FRI, SAT, SUN and *. Multiple values must be separated by a comma."
+          if days_includes_days_of_months?(days)
+              raise "day values 1-31 or last is only valid with frequency :monthly" if frequency == :weekly
+              raise "Do not use frequency_modifier property when day values set as 1-31 or last" if frequency_modifier && frequency_modifier != 1
```
rhass: We should add tests for these two raises given the complexity the conditional logic. If they are covered, it is difficult to correlate and understand due to the complexity.

Vasu1105: @rhass for failing test cases I hope you are having right build of 'win32-taskscheduler'. We have fixed some issues in win32-taskcheduler and window_task now depends on this gem. And the new release is not yet done for win32-taskschduler so on my local in my Gemfile so you might need to point to win32-taskscheduler ole branch on github in Gemfile and comment out dependency line for win32-taskscheduler library in chef-universal-mingw32.gemspec.

For adding tests for raise error I will work on it.

Vasu1105:
@rhass for failing test cases I hope you are having right build of 'win32-taskscheduler'. We have fixed some issues in win32-taskcheduler. window_task resource now depends on win32-taskscheduler gem. The new release is not yet done for win32-taskschduler so on local in you might need to point to win32-taskscheduler to ole branch on github in Gemfile and comment out dependency line for win32-taskscheduler library in chef-universal-mingw32.gemspec.

btm:
@Vasu1105 are these current test failures from a problem with the unreleased dependency?

Vasu1105:
@btm yes

btm:
@Vasu1105 okay. we'll want to get those green since we're adding a lot of tests. I'll ship the dep as soon as it is ready.

btm:
win32-taskscheduler 0.4.0 is released, so we can update this to `"~> 0.4.0"` now.

Vasu1105:
@btm Appveyor is green but travis is failing. I tried committing the Gemfile.lock but still the travis was failing so I reverted  back Gemfile.lock since I am not sure we should commit that or not. Need help in understanding travis issue.

Vasu1105:
@btm I have fixed all travis failures for windows_task resource. So now the failing travis issue is not related to windows_task resource.

btm:
@Vasu1105 it looks like there are some chefstyle style violations to fix: https://travis-ci.org/chef/chef/jobs/365971394

btm:
@Vasu1105 you'll need to cherry-pick this change in to get the latest 1.x version of license_scout, which I updated to add an override for win32-taskscheduler. https://github.com/chef/chef/commit/c40c9763580361c67496367015a8e849638dde04.

Also it looks like the functional tests need an update, the ad-hoc build produced this:
```
Failures:

  1) Chef::Resource::WindowsTask action :run runs the existing task
     Failure/Error: expect(current_resource.task.status).to eq("queued").or eq("running") # queued or can be running

          expected: "queued"
               got: "ready"

          (compared using ==)

       ...or:

          expected: "running"
               got: "ready"

          (compared using ==)
     # ./spec/functional/resource/windows_task_spec.rb:1358:in `block (3 levels) in <top (required)>'
```

btm:
This PR refactors the `windows_task` resource to use the win32-taskscheduler library (and FFI) rather than calling out to task scheduler XML. This should produce a much more stable experience but we really need some people to try it out in their current scenarios. 

@seththoenen, @troyready, @algaut, @sarahbakal, @spuder: You've recently filed a bug against `windows_task`, would you try this build and report back here any issues, concerns, success, or other feedback?

There's a test build of this PR here: https://packages.chef.io/files/unstable/chef/14.0.202+20180417172358/windows/2016/chef-client-14.0.202+20180417172358-1-x64.msi

lib/chef/resource/windows_task.rb
```diff
@@ -49,67 +49,118 @@ class WindowsTask < Chef::Resource
...
-            execution_time_limit(duration)
-          end
+          execution_time_limit(259200) if execution_time_limit == "PT72H"
+          raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds as an Integer (e.g. 60) or a String with numeric values only (e.g. '60')." unless numeric_value_in_string?(execution_time_limit)
+          execution_time_limit(sec_to_min(execution_time_limit))
         end
 
+        # Removed the default value :hourly for frequency in Chef 14 following code added for backward compatibility
+        chef_version = ::Chef::VERSION.split(".")
+        if chef_version[0].to_i > 13 || (chef_version[0].to_i >= 0)
```
algaut: Doesn't (chef_version[0].to_i >= 0) include all chef versions?
If that check is necessary, I don't see the reason to first test if it's > 13

Vasu1105: @algaut Thanks I am fixing that code. I want to have version check for chef 14.1. Fixing this.

btm:
Here's the latest build. I'd appreciate feedback from someone using `windows_task`.

https://packages.chef.io/files/unstable/chef/14.0.203+20180418122035/windows/2016/chef-client-14.0.203+20180418122035-1-x64.msi

seththoenen:
@btm, I've tested this on a 2012 R2 and a 2016 system in my lab. The scheduled task created by the chef-client cookbook was successful and is idempotent!

lokiwins:
@btm 
I ran through a couple of tests with the 14.0.203 build you linked and my windows_task resources executed without issue. I ran these tests in 2 fashions being dependent on windows cookbook >=4.1.4 and without. I only tested on Server 2012R2 and Server 2016 and I did not test all available options with windows_task. 

Task Resources Tested:
`windows_task 'Test1' do
    user node['WinTaskTest']['Administrator']['User']
    password node['winTaskTest']['Administrator']['Password']
    cwd "#{node['winTaskTest']['Maintenance']['dir']}"
    command "install.bat"
    run_level :highest
    frequency :none
    action :create
  end

windows_task 'Test2' do
    user node['WinTaskTest']['Administrator']['User']
    password node['winTaskTest']['Administrator']['Password']
    cwd "#{node['winTaskTest']['Maintenance']['dir']}"
    command "install.bat"
    run_level :highest
    frequency :once
    start_time '00:01'
    action :create
  end

module.Server2012R2.vsphere_virtual_machine.BaseVM-Test (chef): Starting Chef Client, version 14.0.203
module.Server2016.vsphere_virtual_machine.BaseVM-Test (chef): Starting Chef Client, version 14.0.203`

Granted I really only use windows_task to get around WinRM Access Denied issues so my tasks are usually just created, executed, then deleted after so I tend to not have to worry about frequency modifier, day, months, idle_time, or start_time. Test2 was my workaround for when frequency :none caused an error when there was a dependency on the windows cookbook. All of my tasks executed as expected regardless of any dependency on windows cookbook. Hope it helps! Thanks again!

spuder:
The two example scheduled tasks I've had trouble with in chef 13 that worked in chef 12
Windows 2016

# Test 1
**Issue**: https://github.com/chef/chef/issues/6824
**problem**: ` idle_time property is only valid for tasks that run on_idle`
**result**: fixed ✅ 
```ruby
windows_task 'test' do
  task_name 'test'
  command "set"
  run_level :highest
  frequency :onstart
  action :create
end
```

# Test 2
**Issue:** https://github.com/NetDocuments/ad-join-cookbook/issues/13
**problem**: can't set start time (schedtask.exe doesn't allow setting start time on creation)
**result**: failure ❌ 
**expected behavior**: honor start_day (if possible), or throw error if not allowed.
actual behavior: while the chef run completed successfully without errors or warning, the `start_day` and `start_time` attribute was silently ignored.
```ruby
  windows_task 'chef ad-join' do
    task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
    user 'SYSTEM'
    command 'chef-client -L C:\chef\chef-ad-join.log'
    run_level :highest
    frequency :onstart
    start_day '06/09/2016'
    start_time '01:00'
    action :create
  end
```

# Test 3
**Issue** https://github.com/chef/chef/issues/6059
**problem**: windows_task no longer has a `:change` action (work around for test #2)
**result**: Worked (with some negative consequences, see test 4) ⚠️ 
```ruby
windows_task 'chef ad-join' do
  task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
  user 'SYSTEM'
  command 'chef-client -L C:\chef\chef-ad-join.log'
  run_level :highest
  frequency :onstart
  action :create
end

windows_task 'modify sched task start time' do
  task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
  start_day '06/09/2016'
  start_time '01:00'
  action :change
end
```

# Test 4
issue: https://github.com/chef/chef/issues/6887
problem: windows task 'modify' changes the trigger type
expected_behavior: windows task runs 'onstart'. The 'start-time' is in the past
result:  Trigger is 'one time', when it should be 'onstart' :x:
result:  'repeat task every hour' should *not* be checked :x:
Windows tasks are notoriously hard to get start_time working as expected. This may be a bug in windows itself. Chef 12 with the windows cookbook was able to handle this use case as expected. A work around is provided in the linked issue
```ruby
windows_task 'chef ad-join' do
  task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
  user 'SYSTEM'
  command 'chef-client -L C:\chef\chef-ad-join.log'
  run_level :highest
  frequency :onstart
  action :create
end

windows_task 'modify sched task start time' do
  task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
  start_day '06/09/2016'
  start_time '01:00'
  action :change
end`
```
screenshot 2018-04-18 14 14 32

Vasu1105:
For Test 2 I have done some analysis https://msdn.microsoft.com/en-us/library/windows/desktop/aa446815(v=vs.85).aspx#properties

Microsoft itself is not have any property which will clear up Activate settings. To while create you should not pass the start_time or start_day. I have exported the xml after manual creation of :onstart schedule task.  And this is what the result is with comparison

screenshot from 2018-04-19 13-06-40

According to me there will be only solution to this is not to pass start_day and start_time in recipe.

lib/chef/resource/windows_task.rb
```diff
@@ -49,67 +49,118 @@ class WindowsTask < Chef::Resource
...
-            execution_time_limit(duration)
-          end
+          execution_time_limit(259200) if execution_time_limit == "PT72H"
+          raise ArgumentError, "Invalid value passed for `execution_time_limit`. Please pass seconds as an Integer (e.g. 60) or a String with numeric values only (e.g. '60')." unless numeric_value_in_string?(execution_time_limit)
+          execution_time_limit(sec_to_min(execution_time_limit))
         end
 
+        # Removed the default value :hourly for frequency in Chef 14.1 following code added for backward compatibility
+        chef_version = ::Chef::VERSION.split(".")
+        if chef_version[0].to_i > 14 || (chef_version[0].to_i == 14 && chef_version[1].to_i > 0)
```
btm: We need to remove this version check and go one way or another here. Otherwise we're testing behavior that will be different on release.

spuder:
@btm 

Test 2

More context can be gathered from [this issue](https://github.com/NetDocuments/ad-join-cookbook/issues/13). Basically if you make a scheduled task with ':onstart', windows will set 'activate' to the current date and time. If you then change your time zone (with group policy or otherwise) and reboot, then the scheduled task will not run because 'activate time' is in the future. 
With chef 12 and the 'windows' cookbook, you could simply modify the start time (activate time) to some arbitrary date/time in the past as a work around. (In my testing schtask.exe behaves differently with 'start_time' than schedtasks from xml behave). 

>if the task already exists, we set the StartBoundary based on start_day and start_time
if the task does not exist, we create it, and then separately add StartBoundary immediately thereafter

I think that would be sane logic, and would behave more like the schtask.exe command line

Test 3/4

If test 2 is resolved, I would have no need for the :change action

Vasu1105:
@btm In Test 2 we are setting up the startBoundry based on `start_day` and `start_time` as provided by user. If user not provides start_day and start_time we were setting based on current date and time. And I have tested the recipe given by @spuder with rewrite work and the provided start_day and start_time is reflecting please see my screenshot shared with @spuder in this conversation.

@spuder 1) The value you have set in start_day and start_time is getting reflected in activation date and time for at startup scheduled task is getting reflected in Desktop app ? In Test 2 you mentioned that start_day and start_time are silently ignored does that mean that what you are setting is not reflecting in the Desktop app ?
Why I am confused here because I ran the same recipe
` windows_task 'chef ad-join' do
    task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
    user 'SYSTEM'
    command 'chef-client -L C:\chef\chef-ad-join.log'
    run_level :highest
    frequency :onstart
    start_day '06/09/2016'
    start_time '01:00'
    action :create
  end
`
And the result is this.

screenshot from 2018-04-20 13-11-48

2) If above is not the case, then if I am understanding your concern correctly the issue is to clear up Activate check box which is getting activated.

spuder:
@Vasu1105 I re-ran my test2, and I am also confused because I found the same result as you. I must have made a mistake in my first test. The scheduled task is setting the start_day and start_time properly. 

```
  windows_task 'chef ad-join' do
    task_name 'chef ad-join' # http://bit.ly/1WDZ1kn
    user 'SYSTEM'
    command 'chef-client -L C:\chef\chef-ad-join.log'
    run_level :highest
    frequency :onstart
    start_day '06/09/2016'
    start_time '01:00'
    action :create
  end
```

screenshot 2018-04-20 10 12 34

btm:
@Vasu1105 okay, I think we're ready to go here. let's make another build and share it and I'll work on getting some final code review.

Vasu1105:
@btm New adhoc build url https://packages.chef.io/files/unstable/chef/14.0.210+20180424070709/windows/2016/chef-client-14.0.210+20180424070709-1-x64.msi

seththoenen:
@Vasu1105, just tested this in my lab with the latest chef-client cookbook and everything fits our needs.

lib/chef/provider/windows_task.rb
```diff
@@ -20,549 +20,568 @@
...
-            logger.trace "#{new_resource} is not running - nothing to do"
-          else
-            converge_by("#{new_resource} task ended") do
-              run_schtasks "END"
+        def action_delete
+          if current_resource.exists
+            logger.trace "#{new_resource} task exists"
+            converge_by("delete scheduled task #{new_resource}") do
+              ts = TaskScheduler.new
+              ts.delete(current_resource.name)
```
stuartpreston: Shouldn't this be `current_resource.task_name` ?

stuartpreston:
As per the comment on the code, I could not delete a task I created where the resource name and task_name differ.  resource, log, local debug and stacktrace are below:

```
PS C:\Users\Administrator> chef-apply repro.rb
  * windows_task[notepad] action delete

    Error executing action `delete` on resource 'windows_task[notepad]'

    Win32::TaskScheduler::Error
    ---------------------------
    DeleteTask: The system cannot find the file specified.

    Resource Declaration:
    ---------------------
    # In repro.rb

      1: windows_task 'notepad' do
      2:   task_name 'notepadtest'
      3:   user 'vagrant'
      4:   password 'vagrant'
      5:   command 'c:\windows\notepad.exe'
      6:   run_level :highest
      7:   frequency :onstart
      8:   action :delete
      9: end

    Compiled Resource:
    ------------------
    # Declared in repro.rb:1:in `run_chef_recipe'

    windows_task("notepad") do
      action [:delete]
      declared_type :windows_task
      task_name "notepadtest"
      user "vagrant"
      password "vagrant"
      command "c:\\windows\\notepad.exe"
      run_level :highest
      frequency :onstart
      execution_time_limit 4320
      password_required true
    end

    System Info:
    ------------
    chef_version=14.0.210
    platform=windows
    platform_version=10.0.14393
    ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x64-mingw32]
    program_name=C:/opscode/chef/bin/chef-apply
    executable=C:/opscode/chef/bin/chef-apply

[2018-04-26T08:52:43+00:00] FATAL: Stacktrace dumped to C:/chef/cache/chef-stacktrace.out
[2018-04-26T08:52:43+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2018-04-26T08:52:43+00:00] FATAL: Win32::TaskScheduler::Error: windows_task[notepad] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: Win32::TaskScheduler::Error: DeleteTask: The system cannot find the file specified.
```
```
PS C:\Users\Administrator> schtasks.exe /query /tn 'notepadtest'

Folder: \
TaskName                                 Next Run Time          Status
======================================== ====================== ===============
notepadtest                              N/A                    Ready
```
```
PS C:\Users\Administrator> gc C:/chef/cache/chef-stacktrace.out
Generated at 2018-04-26 08:52:43 +0000
Win32::TaskScheduler::Error: windows_task[notepad] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: Win32::TaskScheduler::Error: DeleteTask: The system cannot find the file specified.
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/win32-taskscheduler-0.4.0/lib/win32/taskscheduler.rb:234:in `rescue in delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/win32-taskscheduler-0.4.0/lib/win32/taskscheduler.rb:231:in `delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider/windows_task.rb:169:in `block in action_delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/mixin/why_run.rb:51:in `add_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider.rb:227:in `converge_by'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider/windows_task.rb:167:in `action_delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider.rb:182:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource.rb:576:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:70:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `block (2 levels) in converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `each'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `block in converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/resource_list.rb:94:in `block in execute_each_resource'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:114:in `call_iterator_block'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:85:in `step'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:103:in `iterate'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:55:in `each_with_index'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/resource_list.rb:92:in `execute_each_resource'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:97:in `converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:195:in `run_chef_recipe'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:204:in `run_application'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:216:in `run'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/bin/chef-apply:25:in `<top (required)>'
C:/opscode/chef/bin/chef-apply:59:in `load'
C:/opscode/chef/bin/chef-apply:59:in `<main>'

>>>> Caused by WIN32OLERuntimeError: (in OLE method `DeleteTask': )
    OLE error code:80070002 in <Unknown>
      <No Description>
    HRESULT error code:0x80020009
      Exception occurred.

C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/win32-taskscheduler-0.4.0/lib/win32/taskscheduler.rb:232:in `method_missing'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/win32-taskscheduler-0.4.0/lib/win32/taskscheduler.rb:232:in `delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider/windows_task.rb:169:in `block in action_delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/mixin/why_run.rb:51:in `add_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider.rb:227:in `converge_by'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider/windows_task.rb:167:in `action_delete'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/provider.rb:182:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource.rb:576:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:70:in `run_action'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `block (2 levels) in converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `each'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:98:in `block in converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/resource_list.rb:94:in `block in execute_each_resource'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:114:in `call_iterator_block'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:85:in `step'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:103:in `iterate'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/stepable_iterator.rb:55:in `each_with_index'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/resource_collection/resource_list.rb:92:in `execute_each_resource'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/runner.rb:97:in `converge'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:195:in `run_chef_recipe'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:204:in `run_application'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/lib/chef/application/apply.rb:216:in `run'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.0.210-universal-mingw32/bin/chef-apply:25:in `<top (required)>'
C:/opscode/chef/bin/chef-apply:59:in `load'
C:/opscode/chef/bin/chef-apply:59:in `<main>'
```

tas50:
We should make sure that https://github.com/chef/chef/issues/7169 gets resolved in this PR as well

Vasu1105:
@btm, @stuartpreston, @tas50 I have fixed the --why-run by moving code in converge block. Please verify.

Vasu1105:
@stuartpreston also fixed the :delete action.

PR #6820, opened 2018-02-01
guard against somehow being called by the package resource

thommay:


ccope:
aww, I had hoped this issue was fixed since my original pr :-(

thommay:
Yeah we really need to fix this properly.

lamont-granquist:
eh wut?  this should be only a unit testing failure?

thommay:
> eh wut? this should be only a unit testing failure?

@lamont-granquist we have a number of functional tests that create `Chef::Resource::Package` and then do stuff with it, which is obviously bad but we need to get this test back to green first.

lamont-granquist:
think should fix it, but have to see what travis thinks.

PR #6826, opened 2018-02-04
Remove testing of Debian 7 as it's going EOL

tas50:
Debian 7 goes EOL 31st May 2018. We should merge this around that time.

### DIFF
--- a/omnibus/.kitchen.yml
+++ b/omnibus/.kitchen.yml
@@ -39,10 +39,6 @@ platforms:
     run_list: yum-epel::default
   - name: centos-7.2
     run_list: yum-epel::default
-  - name: debian-6.0.8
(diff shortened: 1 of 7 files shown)


omnibus/.kitchen.yml
```diff
@@ -39,10 +39,6 @@ platforms:
     run_list: yum-epel::default
   - name: centos-7.2
     run_list: yum-epel::default
-  - name: debian-6.0.8
-    run_list: apt::default
-  - name: debian-7.9
-    run_list: apt::default
   - name: debian-8.2
     run_list: apt::default
   - name: freebsd-9.3
```
tas50: This entire file needs cleanup, but that's for a different PR

PR #6833, opened 2018-02-07
Don't rely on the Passwd Ohai plugin in resources

thommay:


thommay:
Context is as part of https://chef.github.io/chef-rfc/rfc103-disable-passwd-and-sessions-plugins.html we'll be disabling passwd by default, so we need a better way to access the user's uid.

PR #6852, opened 2018-02-14
WIP: Add cpan_module resource from the perl cookbook

tas50:


lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+               description: "The path of the cpanm binary.",
+               default: "cpanm"
+
+      action :install do
+        description "Install the module."
+
+        declare_resource(:execute, "CPAN :install #{new_resource.module_name}") do
+          cwd current_working_dir
+          command cpanm_install_cmd
+          environment "HOME" => current_working_dir, "PATH" => "/usr/local/bin:/usr/bin:/bin"
```
lamont-granquist: this should be in `({})`'s.  also its missing in the uninstall action.  also it is some very interesting hardcoding that seems a bit brittle.

lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+          # remove leading v and convert underscores to dots since gems parses them wrong
+          mod_ver.gsub!(/v_?(\d)/, '\\1')
+          mod_ver.tr!("_", ".")
+          # in the event that this command outputs whatever it feels like, only keep the first vers number!
+          version_match = /(^[0-9.]*)/.match(mod_ver)
+          version_match[0]
+        end
+
+        def module_exists?
+          !shell_out("perl", "-M", new_resource.module_name, "-e", "1").error?
```
lamont-granquist: we configure `new_resource.cpanm_binary` but just shell out to the perl we find first in the chef PATH which seems super wrong.

lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      action_class do
+        def module_exists_new_enough
+          existing_version = parse_cpan_version
+          return false if existing_version.empty? # mod doesn't exist
+          return true if new_resource.version.nil? # mod exists and version is unimportant
+          @comparator, @pending_version = new_resource.version.split(" ", 2)
+          @current_vers = Gem::Version.new(existing_version)
+          @pending_vers = Gem::Version.new(@pending_version)
+          @current_vers.method(@comparator).call(@pending_vers)
```
lamont-granquist: This should use`Gem::Requirement.new().satisfied_by?` API instead of what its doing here.

lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+          !shell_out("perl", "-M", new_resource.module_name, "-e", "1").error?
+        end
+
+        def cpanm_install_cmd
+          @cmd = "#{new_resource.cpanm_binary} --quiet "
+          @cmd += "--force " if new_resource.force
+          @cmd += "--notest " unless new_resource.test
+          @cmd += new_resource.module_name
+          @cmd += parsed_version
+          @cmd
```
lamont-granquist: both of these should use arrays

lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+          @cmd += "--force " if new_resource.force
+          @cmd += "--uninstall "
+          @cmd += new_resource.module_name
+          @cmd
+        end
+
+        # a bit of a stub, could use a version parser for really consistent experience
+        def parsed_version
+          return "~\"#{new_resource.version}\"" if new_resource.version
+          ""
```
lamont-granquist: this looks like it got mangled to death by Style/GuardClause?

lib/chef/resource/cpan_module.rb
```diff
@@ -0,0 +1,143 @@
...
+          return false if existing_version.empty? # mod doesn't exist
+          return true if new_resource.version.nil? # mod exists and version is unimportant
+          @comparator, @pending_version = new_resource.version.split(" ", 2)
+          @current_vers = Gem::Version.new(existing_version)
+          @pending_vers = Gem::Version.new(@pending_version)
+          @current_vers.method(@comparator).call(@pending_vers)
+        end
+
+        def parse_cpan_version
+          mod_ver_cmd = Mixlib::ShellOut.new("perl -M#{new_resource.module_name} -e 'print $#{new_resource.module_name}::VERSION;' 2> /dev/null")
```
lamont-granquist: again with perl hardcoding and not using `shell_out` and using mixlib-shellout directly (which is going to be super-confusing to debug since i think you get path-sanity with shell_out and you don't here.

thommay:
This is not going to make Chef 14, but will be stabilised further in the cookbook and reconsidered for Chef 15 next year.

PR #6858, opened 2018-02-15
Remove Attribute Lazy Caching, Add Per-Key Cache Flushing

lamont-granquist:
🚧  🚧  🚧  🚧  🚧  🚧  🚧  🚧  🚧  BROKEN SPECS

The changes in #6424 did two things:

- moved the cache to per-container so that resets could be sent to e.g. `node["foo"]["bar"]["baz"]` and only values under that would need to be recomputed.
- made all the caches lazy caches so that when `node["foo"]["bar"]["baz"]` was reset the cache reset code would just create an empty container there, which was marked that it needed cache population on access.

The problem with the the lazy caches was that all the cache objects inherited directly from Array or Hash, so while all the methods on those objects were intercepted and would lazily create the cache on access, it is not really possible to intercept messages to e.g. Hash#merge on actual hash objects (e.g. the `{}.merge(node['foo'])` case.  This would work better with decorators since most APIs internal to ruby call `#to_h/#to_a` on things which do not inherit from Hash/Array, but objects which inherit result in the APIs poking around directly at the internal Hash/Array structure without calling methods on the target objects themselves for speed.

This PR removes the lazy caching, so that when `node["foo"]["bar"]["baz"]` is reset it immediately rebuilds the entire cache underneath it.  As a result cache flushing needs to be accurate (previously we could bang on the cache flushing API and sending three messages to flush a cache was cheap, but now it is expensive and results in three rebuilds).

It was also necessary to implement per-key flushing on Mashes.  This is necessary to avoid a write to `node["foo"]` rebuilding all of `node` (and on down the line, but this is the really expensive case).

This is still faster than 12.21.31 or 13.0 attribute performance but is not as fast as 13.7.16 was.

### DIFF
--- a/lib/chef/node.rb
+++ b/lib/chef/node.rb
@@ -179,6 +179,9 @@ class Chef
       policy_group(policy_group)
     end
 
+    # @api private
+    attr_writer :attributes
+
     def attributes
       @attributes ||= Chef::Node::Attribute.new({}, {}, {}, {}, self)
     end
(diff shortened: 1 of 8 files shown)


thommay:
@lamont-granquist everything is red :(

lamont-granquist:
I was incorrect.  This is too slow by almost an order of magnitude.

I think the only way to make laziness work is with decorators and breaking the direct inheritance from Array/Hash.

PR #6873, opened 2018-02-20
Add output_locations functionality to data collector

jonlives:
This commit implements the recent changes to RFC-077, to support multiple output locations in the data collector.

I've tried to add in the existing functionality while preserving both the intent and functionality of the original data collector stuff that is designed to link in to Automate.

I'd particularly welcome feedback on if both output styles should all be combined in to one - I expect this code to likely change somewhat from my original PR, and am putting it up for feedback and review that that in mind.

Signed-off-by: Jon Cowie <jonlives@gmail.com>

### Description

This change adds the ability to configure multiple output_locations in the data collector as specified in modifications to https://github.com/chef/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md

### Issues Resolved

Implements the most recent modifications to https://github.com/chef/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md

### Check List

- [X] All tests pass


jonlives:
The only failing specs right now are known fails with poise and poise/halite :)

lib/chef/data_collector.rb
```diff
@@ -50,8 +50,8 @@ def self.register_reporter?
                         "#{Chef::Config[:data_collector][:mode].inspect} modes, disabling it")
         return false
       end
-      unless data_collector_url_configured?
-        Chef::Log.debug("data collector URL is not configured, disabling data collector")
+      unless data_collector_url_configured? || data_collector_output_locations_configured?
```
thommay: it seems odd, throughout this, to special case data_collector_url. Instead, might it be better to just add that url to the list of output locations?

jonlives: That is entirely possible and something I considered, however there is a bunch of behaviour for that config setting specific to being used with automate, like the initial post & disable behaviour. I erred on the side of leaving that part untouched and augmenting it with output_locations, but I'm entirely happy to merge that all in to one if we feel it'd be better.

lib/chef/data_collector.rb
```diff
@@ -286,14 +293,22 @@ def deprecation(message, location = caller(2..2)[0])
...
+          Chef::HTTP::SimpleJSON.new(url, validate_utf8: false)
         end
       end
 
+      def setup_http_output_locations
+        http_output_locations = {}
+        Chef::Config[:data_collector][:output_locations][:urls].each do |location_url|
+          http_output_locations[location_url] = setup_http_client(location_url)
+        end
+        http_output_locations
```
lamont-granquist: this should use `#each_with_object({})`

lib/chef/data_collector.rb
```diff
@@ -467,21 +512,55 @@ def nested_resource?(new_resource)
...
+          end
         end
+      end
 
-        if uri.host.nil?
+      def validate_data_collector_output_locations!
+        if data_collector_output_locations.empty?
           raise Chef::Exceptions::ConfigurationError,
-            "Chef::Config[:data_collector][:server_url] (#{data_collector_server_url}) is a URI with no host. Please supply a valid URL."
+                "Chef::Config[:data_collector][:output_loations] is empty. Please supply an hash of valid URLs and / or local file paths."
```
thommay: `locations`

lib/chef/data_collector.rb
```diff
@@ -467,21 +512,55 @@ def nested_resource?(new_resource)
...
+        end
+
+        data_collector_output_locations.each do |type, locations|
+
+          method = if type == :urls
+                     method(:validate_and_return_uri)
+                   elsif type == :files
+                     method(:validate_and_create_file)
+                   end
```
lamont-granquist: this is definitely not idiomatic ruby and looks like a C function pointer lookup table, and should probably get rewritten a bit cleaner, i'm going to need more coffee before i have a useful suggestion how to refactor it though.

lib/chef/data_collector.rb
```diff
@@ -467,21 +512,55 @@ def nested_resource?(new_resource)
...
+        return nil
+      end
+
+      def validate_and_create_file(file)
+        send_to_file_location(file, "")
+        return true
+      rescue Errno::ENOENT
+        return false
+      rescue Errno::EACCES
+        return false
```
lamont-granquist: swallowing these exceptions seems a bit odd and if i'm reading this correctly the ultimate error becomes "location is not valid" which i think is going to be less useful to the user than the error which would have been raised directly here?  at least it needs comments here.

lib/chef/data_collector.rb
jonlives: Fixed!

lib/chef/data_collector.rb
jonlives: Fixed!

lib/chef/data_collector.rb
jonlives: Fair point - I've added commenting to this, but also changed the exception handling to re-raise the caught exceptions with clearer text. Essentially, I want to give a slightly more useful error than for example an EACCESS without much context as to why it happened.

lib/chef/data_collector.rb
jonlives: Feedback welcome on making this more idiomatic :)

lamont-granquist: probably just:

```ruby
def handle_type(type, loc)
  type == :urls ? validate_and_return_uri(loc) : validate_and_create_file(loc)
end
```

PR #6877, opened 2018-02-20
Add build_essential resource from the build-essential cookbook

tas50:
This cookbook is by far the most included community cookbook we have. It would be really great if people could just use the resource directly and avoid the extra dependency.

This has the following modifications from the cookbook:

Solaris 10 no longer warns since we don't support Solaris 10. We assume 11
Windows no longer includes the windows recipe. This really should not have been added since build-essential concept doesn't translate there
macOS xcode setup is done inline vs. including the xcode command line setup resource from build-essential cookbook. It's a single execute resource and this seemed pretty silly


coderanger:
It would be a lot more moving pieces, but might want to look at https://github.com/poise/poise-build-essential/tree/master/lib/poise_build_essential/build_essential_providers too since it would be easier to include providers for only platforms we test in core and shunt the rest to a cookbook. At a minimum it would be nice to have some level of functional dispatch so a cookbook could monkey-patch in a new platform.

PR #6878, opened 2018-02-20
Add macos_user_defaults resource from mac_os_x cookbook

tas50:
Brought in as-is from the upstream cookbook


thommay:
Guess I should say explicitly this needs all the documentation goodness

tas50:
I figured we'd wait for the results of our docs team meeting before added all of that

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -0,0 +1,123 @@
...
+               description: "",
+               default: false,
+               desired_state: false
+
+      action :write do
+        description "Write the setting to the specified domain"
+
+        @userdefaults = Chef::Resource.resource_for_node(:macos_userdefaults, node).new(new_resource.name)
+        @userdefaults.key(new_resource.key)
+        @userdefaults.domain(new_resource.domain)
```
lamont-granquist:
```ruby
userdefault = build_resource(:macos_userdefaults, new_resource.name) do
  key new_resource.key
  domain new_resource.domain
end
```

tas50:
Double checked this with lamont via slack. Merging on the verbal approval

PR #6879, opened 2018-02-20
Disable sudo on unit tests

lamont-granquist:
Seems to be somewhat more stable in travis, and we do have coverage in jenkins (and the func and integ tests still run with sudo)


lamont-granquist:
@tas50 @thommay this appears to make the unit tests green.   smash merge if you approve of green tests...

PR #6888, opened 2018-02-21
The end of our long travis unit testing nightmare

lamont-granquist:
The only line here which is actually a bugfix is this one:

https://github.com/chef/chef/pull/6888/files#diff-1c826d5d40b34ba0095996a6453ca135R76

Which stubs and no-ops `:_change_privilege`.  Most of the rest of the specs in this describe block setup expectations on that method, but the first test does not, which means that it really calls that function and really changes privs.

Also added some defensive checking which will blow up if this ever happens again (and has an edge case where it throws out a false positive which is responsible for the other couple lines of spec fixes).

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright 2008-2017, Chef Software Inc.
+# Copyright:: Copyright 2008-2018, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -254,6 +254,24 @@ module Shell
     Chef.resource_priority_map.instance_variable_set(:@map, resource_priority_map.dup)
   end
 
+  # This bit of jankiness guards against specs which accidentally drop privs when running as
+  # root -- which are nearly impossible to debug and so we bail out very hard if this
+  # condition ever happens.  If a spec stubs Process.[e]uid this can throw a false positive
+  # which the spec must work around by unmocking Process.[e]uid to and_call_original in its
+  # after block.
+  if Process.euid == 0 && Process.uid == 0
+    config.after(:each) do
```
danielsdeleo: ISTR these things can take block arguments which are the RSpec objects representing the test. If that's true then you could use that to give a bit more info about the culprit test, which would probably be really awesome if you needed it.

lamont-granquist:
And the ultimate root cause of this looks to be that the specs wind up trying to setreuid to user "aj" group "aj".  On boxes without an "aj" user + group this throws, which is then caught and changed to false which the specs ignore.  When fujin wrote the test it didn't cause any chaos because he uses "fujin" as his login name(?).  On the travis boxes that fail there's an "aj" user, on the ones that do not there is no "aj" user.

This PR has been brought to you by the letters W, T and especially F....

spec/spec_helper.rb
lamont-granquist: i checked the rspec docs and couldn't find any block arguments to these....   i didn't go code surfing to see if they were hidden options...

lamont-granquist:
and i'm pretty certain now that this would have never occurred to chef-aj because mac's use 'staff' as the group by default so aj was testing with "aj:staff" on his mac, while and the specific busted spec omits the group argument when `_chage_privilege` is called so that it tries to use "aj:aj" and that would have thrown on the group lookup.  since travis images are linuxen they use user-private-groups so travis-aj's account was "aj:aj" which triggers the bug.  once her account started getting added to images this bug triggered.

PR #6890, opened 2018-02-21
Disable sudo on unit tests

tas50:
This seems to make unit tests on travis somewhat more stable


thommay:
:-1: #6888 is the real fix, we should just backport that.

tas50:
Oh yep. Even better

PR #6892, opened 2018-02-21
Link to the knife.rb docs when the knife.rb file is missing

tas50:
This is a pretty common first error message to experience with Chef. Let's point the users to a helpful location so they can quickly get up and running. Add some YARD comments while I was in the config class

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -182,7 +182,7 @@ class Chef
       config_loader.profile = profile
       config_loader.load
 
-      ui.warn("No knife configuration file found") if config_loader.no_config_found?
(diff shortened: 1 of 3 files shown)


lib/chef/knife.rb
```diff
@@ -182,7 +182,7 @@ def self.load_config(explicit_config_file, profile)
       config_loader.profile = profile
       config_loader.load
 
-      ui.warn("No knife configuration file found") if config_loader.no_config_found?
+      ui.warn("No knife configuration file found. See https://docs.chef.io/config_rb_knife.html for details.") if config_loader.no_config_found?
```
thommay: Guiding them to run `knife configure` might be more humane than that doc, which we should also fix.

tas50: This is just a backport from master. We could probably make it better, but I think this is better than just tossing a message that requires some Google searching

PR #6895, opened 2018-02-21
Add chef_handler resource from chef_handler cookbook

tas50:
Merges the helper library into the action_class, but otherwise this is
the same as the cookbook.


thommay:
Specs are legit busted though

PR #6896, opened 2018-02-22
add a utility to dump info about resources

thommay:


lib/chef/resource_inspector.rb
```diff
@@ -0,0 +1,89 @@
...
+#
+
+require "chef/cookbook_loader"
+require "chef/cookbook/file_vendor"
+require "chef/cookbook/file_system_file_vendor"
+require "chef/resource/lwrp_base"
+require "chef/run_context"
+require "chef/node"
+require "chef/resources"
+require "chef/json_compat"
```
thommay: this isn't the correct json requirement

PR #6899, opened 2018-02-23
Add more introduced and description fields to resources

tas50:
There's still more, but this adds the fields to a good chunk of additional resources

### DIFF
--- a/lib/chef/resource/osx_profile.rb
+++ b/lib/chef/resource/osx_profile.rb
@@ -28,9 +28,12 @@ class Chef
       provides :osx_profile, os: "darwin"
       provides :osx_config_profile, os: "darwin"
 
+      identity_attr :profile_name
+
       description "12.7"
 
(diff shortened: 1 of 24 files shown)


lib/chef/resource/osx_profile.rb
```diff
@@ -24,13 +24,11 @@ class Resource
     # on the macOS platform. The osx_profile resource installs profiles by using
     # the uuidgen library to generate a unique ProfileUUID, and then using the
     # profiles command to install the profile on the system.
-    #
-    # @since 12.7
     class OsxProfile < Chef::Resource
       provides :osx_profile, os: "darwin"
       provides :osx_config_profile, os: "darwin"
 
-      identity_attr :profile_name
```
tas50: This came along for the ride, but it doesn't need to be here anymore since we set name_property on this attribute anyways

PR #6900, opened 2018-02-23
better docs for Chef::Provider::Service::Upstart.supports?

bradarv90:


tas50:
Merged this in with the whitespace fix too. Thanks for the comment

PR #6902, opened 2018-02-23
Set properties in git resource using our resource DSL

tas50:
Modernize how we define this resource


tas50:
@lamont-granquist @thommay This should be good to go at this point

PR #6903, opened 2018-02-23
Modernize provides in the portage_package resource

tas50:


thommay:
red

tas50:
@lamont-granquist @thommay This should be good to go at this as well

PR #6906, opened 2018-02-23
Modernize the user resource

tas50:

### DIFF
--- a/lib/chef/resource/user.rb
+++ b/lib/chef/resource/user.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright 2008-2017, Chef Software Inc.
+# Copyright:: Copyright 2008-2018, Chef Software Inc.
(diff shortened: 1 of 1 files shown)


thommay:
failing tests

lib/chef/resource/user.rb
```diff
@@ -30,130 +30,21 @@ class User < Chef::Resource
...
-          :gid,
-          arg,
-          :kind_of => [ String, Integer, NilClass ],
-          :coerce => proc { |x| x || nil }
-        )
-      end
+      property :username, String, name_property: true
+      property :comment, String
+      property :uid, [ String, Integer, NilClass ], default: nil
+      property :gid, [ String, Integer, NilClass ], default: nil
```
lamont-granquist: both these can just be `property :uid, [ String, Integer ]`

nil is always the default and always nillable, unless you set a non-nil default.

tas50: Fixed

PR #6912, opened 2018-02-23
Revert "Revert "fixup some unit tests""

lamont-granquist:
Accidentally reverted this in #6898 but i has nothing directly to do with the attribute rewrite and is useful to keep.


lamont-granquist:
(and mostly i'm reverting this here to hopefully keep the git timelines in closer synch so that patches to master that might build on top of this patch will hopefully not merge conflict on backporting)

PR #6914, opened 2018-02-26
 Don't use supervisor process for one-shot / command-line runs

### LINKED ISSUE #6712: Change default behavior of non-interval runs to --no-fork
If we're only running in any sort of one-shot mode, we don't need to fork.

Causes issues and needless complexity with things like chef_client_updater and automate.

lamont-granquist:
without --interval we default to --no-fork, with --interval we default
to running --fork.  --once and --daemonize are handled before this code
so they will also DTRT.

closes #6712


lamont-granquist:
heh that's cute, unit tests are failing without reporting the failure

thommay:
https://ci.appveyor.com/project/chef/chef/build/master-11943#L2149

PR #6917, opened 2018-02-28
bump omnibus deps

thommay:


omnibus/Gemfile
```diff
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
-gem "omnibus", git: "https://github.com/chef/omnibus"
-gem "omnibus-software", git: "https://github.com/chef/omnibus-software"
+gem "omnibus", git: "https://github.com/chef/omnibus", branch: "lcg/improve-appbundling"
+gem "omnibus-software", git: "https://github.com/chef/omnibus-software", branch: "lcg/appbundle-0.11-without"
```
lamont-granquist: i suspect that master of both with appbundler 0.11.0 will work fine, since using `without` doesn't DTRT anyway right now...

Gemfile
```diff
@@ -56,7 +56,7 @@ group(:development, :test) do
 
   # for testing new chefstyle rules
   # gem 'chefstyle', github: 'chef/chefstyle'
-  gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
+  gem "chefstyle"
```
lamont-granquist: you'll be able to back this out in a few minutes

lamont-granquist:
due to circular dependency hell we need a build in current that has bundler 1.16 in order for master to go red with appbundler 0.11.x and we can't easily pin in the travis tests (there's no Gemfile.lock in the kitchen virts).

manhattan builds on master are also red due to circular dep hell.

PR #6932, opened 2018-03-02
pin old appbundler in chef-12

lamont-granquist:
this has no chance of working on later bundler/appbundler


lamont-granquist:
nevermind, looks like this just works.

PR #6935, opened 2018-03-02
Add windows_feature, windows_feature_powershell, and windows_feature_dism resources

tas50:
Ported from the windows cookbook 4.0


lib/chef/resource/windows_feature.rb
```diff
@@ -0,0 +1,81 @@
...
+          when :windows_feature_powershell
+            windows_feature_powershell new_resource.name do
+              action desired_action
+              feature_name new_resource.feature_name
+              source new_resource.source if new_resource.source
+              all new_resource.all
+              timeout new_resource.timeout
+              management_tools new_resource.management_tools
+            end
+          end
```
lamont-granquist: this could probably be a bit more concisely:

```ruby
provider = new_resource.install_method || :windows_feature_dism
declare_resource(provider, new_resource.name) do
  action desired_action
  feature_name new_resource.feature_name
  source new_resource.source if new_resource.source
   all new_resource.all
   timeout new_resource.timeout
   management_tools new_resource.management_tools if provider == :windows_feature_powershell
end
```

although i might be missing something, and the way that the  management_tools property is passed to one and not passed to the other gives me some pause..

kind of up to you tho...

lamont-granquist: `provider` is also inaccurately named now should probably be `resource`.

i imagine this code used to look 10 times worse tho so...

tas50: it was actually a provider at one point. Now it's more of a sub resource. I should probably just rename it and be done with it

tas50: We should probably just pass the management stuff and let it fail. We did a lot of silently helping the user out in this resource, which wasn't always the best thing / was never the right thing

lamont-granquist: yeah this is good enough to ship either way...

tas50: I updated the PR to use that and added it to the cookbook as well. It's a lot simpler now that we don't need to support 3 different types of installs.

lamont-granquist:
doubly approved

PR #6937, opened 2018-03-02
Add Ubuntu 18.04 Testing in Travis

tas50:
We have a dokken image for 18.04 based off the beta and rebuilding on a schedule in Docker Cloud.


tas50:
Looks like we need to update openssh cookbook for 18.04. I'll figure that out and get this updated

tas50:
Fixed openssh cookbook to create the priv sep directory that was missing on 18.04 docker image. This only impacts docker, but it's fixed now for anyone using 18.04 come April

PR #6941, opened 2018-03-04
New interop between Chef and PowerShell 4.0 (or higher)

stuartpreston:
### Description

This change provides a new way of working with PowerShell on Windows-based systems that support .NET Framework 4.0 and PowerShell 4.0 or higher (This is the out of box configuration included with Windows Server 2012 R2).

- The primary benefit of this approach is PowerShell is accessed through the Ruby process and not via `shell_out`.
- The output from PowerShell that is usually formatted specially for console but in the new case is coerced to JSON and available as a Hash.
- Files containing potentially sensitive data are not written to disk, they are sent via COM to the PowerShell engine.
- Errors that occurred during script block execution can be trapped and handled

### What's included:
- Chef::Powershell class (`require "chef/powershell"`)
- `powershell_exec` mixin
- Binary inclusion of COM interop component required in the Windows package.
- Registry keys required for operation added to the Windows installer package.

### Example mixin usage

```
  outcome = powershell_exec("Get-WindowsFeature | where { $_.Installed -eq $true }")
  if outcome.error?
    outcome.errors.each do |e|
      Chef::Log.error e
    end
  else
    Chef::Log.info "Result: #{outcome.result}"
  end
```

NB: powershell_exec has been included in the Universal DSL mixin list, alongside the existing powershell_out

It should be noted a full install of Chef Client is required to test this due to the components being registered by Windows installer.

### Check List

- [x] All tests pass


lib/chef/powershell.rb
```diff
@@ -0,0 +1,53 @@
...
+    # Run a command under PowerShell via a managed (.NET) COM interop API.
+    # This implementation requires the managed dll to be registered on the
+    # target machine.
+    #
+    # Requires: .NET Framework 4.0 or higher on the target machine.
+    #
+    # @param script [String] script to run
+    # @return [Object] output
+    def initialize(*command_args)
+      raise "This class can only be used on the Windows platform." unless RUBY_PLATFORM =~ /mswin|mingw32|windows/
```
tas50: We should just say what the helper is. Sometimes you're deep in Chef errors and it's hard to see what spit the actual error out so you can fix your code.

lib/chef/mixin/powershell_exec.rb
```diff
@@ -0,0 +1,39 @@
...
+      # This implementation requires the managed dll to be registered on the
+      # target machine.
+      #
+      # Requires: .NET Framework 4.0 or higher on the target machine.
+      #
+      # @param script [String] script to run
+      # @return [Chef::PowerShell] output
+      def powershell_exec(*command_args)
+        script = command_args.first
+        options = command_args.last.is_a?(Hash) ? command_args.last : nil
```
lamont-granquist: This should probably just be `powershell_exec(script)` for now.  I don't think options are being used at all here now and can safely be introduced later when there's actually options to use (and `**options` should get used rather than the `last.is_a?(Hash)` stuff which is probably some ruby 1.8 kludginess in shell_out which we've not yet dared to do away with.  The array style passing of arguments on windows is also busted (i tried adding windows support once to bake the poise_shell_out helper directly into mixlib-shellout, but its complicated and brittle -- https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/).

lib/chef/powershell.rb
stuartpreston: Yep you're right, I was struggling to find where we consistently apply this elsewhere, any recommendations?

lib/chef/mixin/powershell_exec.rb
stuartpreston: Yeah, I was following powershell_out's style of doing things here.  I'll revert back to taking the single argument for now :)

stuartpreston:
Ready for merge, the Travis failures with Chef Zero seem unrelated...

RELEASE_NOTES.md
```diff
@@ -2,6 +2,12 @@ _This file holds "in progress" release notes for the current release under devel
 
 # Unreleased
 
+# Chef Client Release Notes 14.TBC:
+
+## New Windows PowerShell mixin `powershell_exec`
+
+Since our supported Windows platforms can all run .NET Framework 4.0 and PowerShell 4.0 we have taken time to add a new interop that will allow faster and safer interactions with the system PowerShell.  You will be able to use the `powershell_exec` mixin in most places where you would have previously used `powershell_out`.
```
tas50: It would be nice to have a benchmark here so we could explain how fancy this new mixin is

PR #6943, opened 2018-03-05
Add more description fields, style fixes, add missing requires

tas50:
Make sure to require chef/resource everywhere. It would work without it but only because some other resource had already required it.

Put the type on the same line as the property, which we discussed in slack

Put introduced after description which makes differing simpler

Convert some yard into description fields


lib/chef/resource/build_essential.rb
```diff
@@ -22,7 +22,12 @@ class BuildEssential < Chef::Resource
       resource_name :build_essential
       provides :build_essential
 
-      property :compile_time, [true, false], default: false
+      description "Use the build_essential resource to install packages required for compiling C software from source"
+      introduced "14.0"
+
+      property :compile_time, [true, false],
```
stuartpreston: Shouldn't the type (and default) here be [TrueClass, FalseClass] for consistency?

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -25,12 +25,11 @@ class MacosUserDefaults < Chef::Resource
...
-      introduced "14.0"
-
       description "Use the macos_userdefaults resource to manage the macOS user defaults"\
                   " system. The properties to the resource are passed to the defaults command"\
                   " and the parameters follow convention of the macOS command. See the defaults(1)"\
                   " man page for details on how the tool works."
+      introduced "14.0"
 
       property :domain,
                String,
```
stuartpreston: Move the type up to the property line for consistency with other resources?

lib/chef/resource/ohai_hint.rb
```diff
@@ -21,17 +21,15 @@ class OhaiHint < Chef::Resource
...
                name_property: true
 
-      property :content,
-               Hash,
-               description: "Values to include in the hint file"
+      property :content, Hash,
+               description: "Values to include in the hint file."
 
       property :compile_time,
                [TrueClass, FalseClass],
```
stuartpreston: Move the type up to the property line for consistency with other resources?

lib/chef/resource/osx_profile.rb
```diff
@@ -20,20 +20,19 @@
...
+                  " on the macOS platform. The osx_profile resource installs profiles by using"\
+                  " the uuidgen library to generate a unique ProfileUUID, and then using the"\
+                  " profiles command to install the profile on the system."
+      introduced "12.7"
 
       default_action :install
-
       allowed_actions :install, :remove
 
       property :profile_name, String, name_property: true
```
stuartpreston: Should name_property be on the same line as the property name? Looks inconsistent with other modernized resources.

lib/chef/resource/rhsm_register.rb
```diff
@@ -27,43 +27,35 @@ class RhsmRegister < Chef::Resource
...
+      property :username, String,
                description: "The username to use when registering. Not applicable if using an activation key. If specified, password and environment are also required."
 
-      property :password,
-               String,
+      property :password, String,
                description: "The password to use when registering. Not applicable if using an activation key. If specified, username and environment are also required."
 
       property :auto_attach,
                [TrueClass, FalseClass],
```
stuartpreston: Why not move this one up to the property line?

tas50: I wasn't breaking them up until I added the descriptions. That bloats the lines so much it's easier to read on individual lines.

PR #6948, opened 2018-03-05
splat array arguments of shell_out_*

mbakhoff:
lib/chef/mixin/shell_out.rb contains these methods:
```
def shell_out_compact(*args, **options)`
def shell_out_compact_timeout(*args, **options)
```

Some callers, e.g. lib/chef/provider/package/macports.rb pass the command as an array:
```
command = [ "port", "installed", new_resource.package_name ]
status = shell_out_compact_timeout(command)
```

This works fine in most cases:
```
def shell_out_compact_timeout(*args, **options)
  puts "args: " + args.inspect
  puts "options: " + options.inspect
end

command = [ "port", "installed", "zip" ]
status = shell_out_compact_timeout(command)
args: [["port", "installed", "zip"]]
options: {}
```

Things get more complicated when some cookbook has run `require 'extlib'`. Extlib monkey patches Array to have a to_hash method and this is very confusing to the shell_out_ methods:

```
require 'extlib'
# same as above
status = shell_out_compact_timeout(command)
args: [{"port"=>nil, "installed"=>nil, "zip"=>nil}]
options: {}
# in the macports case, this results in: port= command not found 
```

It seems that ruby is trying to use some implicit conversion here, but things go horribly wrong. One solution is to splat the argument to avoid the conversion: `shell_out_compact_timeout(*command)`. This PR adds such spats to all usages of shell_out *args/**options methods that I could find. 

I'm not a ruby expert, so I'm not sure if that's even the right approach to solving it? 
Note that some usages of shell_out that pass the command as an array already splat it.

### DIFF
--- a/lib/chef/provider/group/dscl.rb
+++ b/lib/chef/provider/group/dscl.rb
@@ -27,7 +27,7 @@ class Chef
           argdup = args.dup
           cmd = argdup.shift
           shellcmd = [ "dscl", ".", "-#{cmd}", argdup ]
-          status = shell_out_compact(shellcmd)
+          status = shell_out_compact(*shellcmd)
           stdout_result = ""
           stderr_result = ""
           status.stdout.each_line { |line| stdout_result << line }
--- a/lib/chef/provider/group/pw.rb
+++ b/lib/chef/provider/group/pw.rb
@@ -48,7 +48,7 @@ class Chef
             command += [ "-M", new_resource.members.join(",") ]
           end
 
-          shell_out_compact!(command)
+          shell_out_compact!(*command)
         end
 
         # Manage the group when it already exists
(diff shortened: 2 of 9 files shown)


lamont-granquist:
so...  extlib is a support library for merb which hasn't been updated since 2013, and monkeypatching Array to have `#to_hash` is more than a bit awful.  the real solution here is to abandon that library and stop using it.

i'll look at this PR since it may be correct anyway, but mostly this problem is self-inflicted...

mbakhoff:
agreed. i'm looking into ways to replace/fix the dependencies that pull in extlib in my cookbooks, but that may be tricky. in any case - if this change looks pointless in the long run, then i won't mind if it gets rejected

thank you for your fast response

coderanger:
On one hand this is pretty harmless as a fix so I don't really mind per se, but I also am with Lamont that I give precisely zero [redacted]s about any bugs introduced by extlib ;-) (or anything else that monkeypatches core objects)

mbakhoff:
closing this as it doesn't really fix anything. thanks for your feedback!

PR #6950, opened 2018-03-05
Convert more set_or_returns to proper properties

tas50:
Just modernizing our resources so we can start to add descriptions to them for the documentation

### DIFF
--- a/lib/chef/resource/cron.rb
+++ b/lib/chef/resource/cron.rb
@@ -18,6 +18,7 @@
 #
 
 require "chef/resource"
+require "chef/provider/cron" # do not remove. we actually need this below
 
 class Chef
   class Resource
@@ -135,14 +136,7 @@ class Chef
         )
       end
 
-      def time(arg = nil)
-        set_or_return(
-          :time,
-          arg,
-          :equal_to => Chef::Provider::Cron::SPECIAL_TIME_VALUES
-        )
-      end
-
+      property :time, Symbol, equal_to: Chef::Provider::Cron::SPECIAL_TIME_VALUES
       property :mailto, String
       property :path, String
       property :home, String
--- a/lib/chef/resource/group.rb
+++ b/lib/chef/resource/group.rb
@@ -20,7 +20,6 @@
 class Chef
   class Resource
     class Group < Chef::Resource
-      identity_attr :group_name
       state_attrs :members
 
       description "Use the group resource to manage a local group."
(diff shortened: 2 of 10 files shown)


lib/chef/resource/cron.rb
```diff
@@ -18,6 +18,7 @@
 #
 
 require "chef/resource"
+require "chef/provider/cron" # do not remove. we actually need this below
```
tas50: We have a pile of cron constants in the provider

lib/chef/resource/cron.rb
```diff
@@ -135,14 +136,7 @@ def weekday(arg = nil)
...
 
-      def time(arg = nil)
-        set_or_return(
-          :time,
-          arg,
-          :equal_to => Chef::Provider::Cron::SPECIAL_TIME_VALUES
-        )
-      end
-
+      property :time, Symbol, equal_to: Chef::Provider::Cron::SPECIAL_TIME_VALUES
```
tas50: Technically we didn't restrict this to a symbol, but we expected it to be a symbol in the equal_to so this is the same

lib/chef/resource/group.rb
```diff
@@ -74,29 +56,9 @@ def excluded_members(arg = nil)
...
-
-      def non_unique(arg = nil)
-        set_or_return(
-          :non_unique,
-          arg,
-          :kind_of => [ TrueClass, FalseClass ]
-        )
-      end
+      property :append, [ TrueClass, FalseClass ], default: false
+      property :system, [ TrueClass, FalseClass ], default: false
```
tas50: This lacked a default before, but it seems like the default was in fact false

lib/chef/resource/link.rb
```diff
@@ -35,62 +35,23 @@ class Resource
     class Link < Chef::Resource
       include Chef::Mixin::Securable
       resource_name :link
+      provides :link
 
-      identity_attr :target_file
-
-      state_attrs :to, :owner, :group
+      state_attrs :owner # required since it's not a property below
```
tas50: This comes from Chef::Mixin::Securable which doesn't use properties so no state set

lib/chef/resource/link.rb
```diff
@@ -35,62 +35,23 @@ class Resource
...
-      def owner(arg = nil)
-        set_or_return(
-          :owner,
-          arg,
-          :regex => Chef::Config[:user_valid_regex]
-        )
-      end
+      property :target_file, String, name_property: true, identity: true
+      property :to, String
+      property :link_type, [String, Symbol], coerce: proc { |arg| arg.kind_of?(String) ? arg.to_sym : arg }, equal_to: [ :symbolic, :hard ], default: :symbolic
```
tas50: This had no type before, but it could be string or symbol so lets enforce that now

lib/chef/resource/link.rb
```diff
@@ -35,62 +35,23 @@ class Resource
...
+      property :group, [String, Integer], regex: [Chef::Config[:group_valid_regex]]
```
tas50: group/user had no type before, but our spec checks integers so clearly those should be here

thommay:
genuine red though: https://travis-ci.org/chef/chef/jobs/349650692#L1668

PR #6953, opened 2018-03-05
Avoid compile time error in apt_repository

tas50:
My previous PR to apt_repository introduced a bug where gpg keys couldn't validate. This converts a raise to an execute so it all happens during converge.


lamont-granquist:
that should probably be a `verify` hanging off the remote_file/cookbook_file instead of being its own execute resource?

Issue #3363, opened 2015-05-11
gem_package behaves like chef_gem after chef installs system ruby


duallain:
I expect that this recipe when run for the first time will install the json gem to /usr/bin/gem:

``` ruby
package 'rubygems' do
  action :install
end

gem_package 'json' do
  action :install
  gem_binary '/usr/bin/gem'
end

gem_package 'thor' do
  action :install
  gem_binary '/usr/bin/gem'
end

execute "/usr/bin/gem list |grep thor"
execute "/usr/bin/gem list |grep json"
```

thor is installed to /usr/bin/gem but json is not.  This is for chef 12.3.0 in test kitchen.

The json gem is not installed because of this message:

```
  * gem_package[json] action install[2015-05-11T22:07:19+00:00] INFO: Processing gem_package[json] action install (repo::default line 16)       
 (up to date)       
  * gem_package[thor] action install[2015-05-11T22:07:24+00:00] INFO: Processing gem_package[thor] action install (repo::default line 21)       
[2015-05-11T22:07:30+00:00] INFO: gem_package[thor] installed thor at 0.19.1
```

Testing with chef 11.16.4 the recipe runs all the way though without issue.

```
* gem_package[json] action install[2015-05-11T22:10:59+00:00] INFO: Processing gem_package[json] action install (repo::default line 16)      
- install version 1.8.2 of package json
```

lamont-granquist:
You most likely need to use `action :upgrade`.  The json gem is a default gem in ruby:

```
# gem uninstall json
ERROR:  While executing gem ... (Gem::InstallError)
    gem "json" cannot be uninstalled because it is a default gem
```

If you try to `:install` it then it will find it and won't install it.

If you try to `:upgrade` it then it'll download the latest if its available:

```
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * gem_package[json] action upgrade
    - upgrade package json from 1.8.1 to 1.8.2
```

I suspect between 11.16.4 and 12.3.0 that we may have fixed this.

duallain:
That seemed to work, but I'm not sure that it is the full/correct solution.

The json gem is installed in /opt/embedded/chef. But is not installed for /usr/bin/gem. Should the fact that  a gem is installed in one path affect if it is installed in a different one?  Especially with the difference between chef_gem and gem_package.

I'm now fixed, but it is possible that this is a bigger problem.

lamont-granquist:
[...]

PR #6955, opened 2018-03-07
Rework installed_versions for newer Rubygems

### LINKED ISSUE #6954: Weird gem issues
Opening this for tracking. Some time between 14.0.89 and 14.0.99 my integration tests for poise-ruby broke. The culprit appears to be something funky with `gem_package` causing `bundler` to show as up-to-date even if it isn't:

```
...
* ruby_gem[bundler] action upgrade (up to date)
...
kitchen@2a2d85915112:~$ /usr/bin/gem2.3 list

*** LOCAL GEMS ***

bigdecimal (1.2.8)
did_you_mean (1.0.0)
io-console (0.4.5)
json (1.8.3)
minitest (5.8.4)
net-telnet (0.1.1)
power_assert (0.2.7)
psych (2.0.17)
rake (10.5.0)
rdoc (4.2.1)
test-unit (3.1.7)
```

We should track this down :)

coderanger:
Fixes #6954, #3363. Not sure how best to test this other than the poise-ruby test suite. I put in a monkeypatch version of this and it seems to fix my suite.

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -86,7 +86,22 @@ class Chef
           # === Returns
           # [Gem::Specification]  an array of Gem::Specification objects
           def installed_versions(gem_dep)
-            if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("1.8.0")
+            rubygems_version = Gem::Version.new(Gem::VERSION)
+            if rubygems_version >= Gem::Version.new("2.7")
+              # In newer Rubygems, bundler is now a "default gem" which means
+              # even with AlternateGemEnvironment when you try to get the
+              # installed versions, you get the one from Chef's Ruby's default
+              # gems. This workaround ignores default gems entirely so we see
+              # only the installed gems.
+              stubs = gem_specification.send(:installed_stubs, gem_specification.dirs, "#{gem_dep.name}-*.gemspec")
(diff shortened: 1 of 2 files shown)


coderanger:
The monkeypatch version of this has cleared my test suite so that's a good sign. Looks like I'll need to fix some stubbing in the unit tests before we can merge this but still good to get eyeballs. If/when the upstream fix for default vs installed goes in, we can also switch this over to using the (public) `stubs_for` rather than `installed_stubs`.

lib/chef/provider/package/rubygems.rb
```diff
@@ -86,7 +86,22 @@ def gem_specification
...
+              # only the installed gems.
+              stubs = gem_specification.send(:installed_stubs, gem_specification.dirs, "#{gem_dep.name}-*.gemspec")
+              # Filter down to only to only stubs we actually want. The name
+              # filter is needed in case of things like `foo-*.gemspec` also
+              # matching a gem named `foo-bar`.
+              stubs.select! {|stub| stub.name == gem_dep.name && gem_dep.requirement.satisfied_by?(stub.version) }
+              # This isn't sorting before returning beacuse the only code that
+              # uses this method calls `max_by` so it doesn't need to be sorted.
+              stubs
+            elsif rubygems_version >= Gem::Version.new("1.8.0")
```
tas50: Do we still need to check for 1.8 at this point. That's 2011.

PR #6957, opened 2018-03-07
Remove platfom restrictions in provides and don't require providers

tas50:
Don't require providers in the resources since we map these automatically. Also remove or simplify the platform restrictions in the provides lines in the resources. We're not really getting anything from these the way we wire up resources.


tas50:
This is going to require some rework to the provider resolver specs.

coderanger:
Most of these can just be deleted. There is no need to have both `resource_name` and `provides` set to the same thing.

lamont-granquist:
do we ever want to do something where we get rid of resource_name as a statement and wire up one of the provides to set resource_name instead?  i'd kind of be inclined to have one less thing.

tas50:
@lamont-granquist @coderanger Let's come up with a concrete decision on resource_name vs. resource_name + provides. We have some differing opinions there. Let me know what I should be doing and I'll get it pushed out everywhere in time.

coderanger:
@lamont-granquist You can do that already, `provides :foo, canonical: true` IIRC. That's pretty literally what `resource_name :foo` does. Don't care which we use, just no need for both :D

lamont-granquist:
i kind of like `canonical: true` then -- its a flag instead of another dsl method, and i like that bit of explicit magic rather than hidden magic of the provides which occurs when resource_name is called.

lamont-granquist:
lets just merge this and clean up later.  i'm working on the `resource_name` thing.

PR #6959, opened 2018-03-07
Update @content_class assignment from = to ||=

jasonwbarnett:
### Description

Update the assignment of `@content_class` to use `||=` so that inheritance of
`Chef::Provider::RemoteFile` is easier to do.

### Issues Resolved

N/A

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/remote_file.rb
+++ b/lib/chef/provider/remote_file.rb
@@ -25,7 +25,7 @@ class Chef
       provides :remote_file
 
       def initialize(new_resource, run_context)
-        @content_class = Chef::Provider::RemoteFile::Content
+        @content_class ||= Chef::Provider::RemoteFile::Content
         super
       end
 


jasonwbarnett:
Build errors seem completely unrelated and temporary. Would someone check me on that?

lamont-granquist:
there's nothing wrong with this PR, but at the same time you should be using composition and not inheritance and should be writing custom resources that wrap and use the remote_file resource and not directly inheriting.

jasonwbarnett:
This was inspired because I had a need to create a resource that shares 100% of the functionality of `remote_file` with one minor modification to the `#headers` method. It's essentially dynamically generating headers on the fly with each request.

Does this seem reasonable or not?

lamont-granquist:
so that's actually the wrong approach.

you should write a new resource which uses the remote_file resource more like this:

https://github.com/lamont-cookbooks/sk_s3_file/blob/master/providers/default.rb#L28-L35

note that is some fairly gross Chef-11-era code that i need to clean up at some point so don't copy that straight like that.

the one bit of annoyance with this approach is you need to copy all your properties off your resource onto the remote_file resource.  there's one example of that there (probably old and busted), another newer example here:

https://github.com/chef-cookbooks/etcd/blob/master/resources/etcd_service.rb#L20-L29

we need to get a helper like that into core to make the right way of doing things here not take more typing.

jasonwbarnett:
@lamont-granquist My use case allows for retries and with each retry it generates a new set of headers which are time sensitive.

What is the reasoning of using one pattern in the parent class on not using that same pattern in it's inheriters?

PR #6962, opened 2018-03-07
Use node.override not node.normal in the windows_feature_dism resource

tas50:
This fix was released in the windows 4.0.1 cookbook


thommay:
automatic feels more appropriate, but that's entirely bikesheddy

tas50:
Automatic failed for me when I attempted to use it in the cookbook. I need to look into that, but this matches the cookbook behavior we have now

PR #6965, opened 2018-03-09
Add docker testing in Travis

tas50:


tas50:
Docker cookbook is currently not Chef 14 compatible. I'm slowly working towards that. This testing was great. It proved the state of the cookbook before moving it into core.

Issue #6851, opened 2018-02-14
Issue with chef mount resource not updating existing entries in /etc/fstab when enable action is used


frostygresh:
## Description
Mount resource is creating a new entry for file system that already has an entry in /etc/fstab.
If the fstab entry for the file system has a different device that is specified in the mount block, the enable action will create a new entry at the end of fstab and leaves the existing entry in fstab as well. 
Test environment:
Linux: CentOS 7.4
chef-client version: 13.6.4

Example:
```
mount '/boot' do
    device "#{node['filesystem']['by_mountpoint']['/boot']['uuid']}"
    device_type :uuid
    pass 0
    fstype "#{node['filesystem']['by_mountpoint']['/boot']['fs_type']}"
    options 'defaults'
    action :enable
 end
```

This is intended to ensure that the /boot file system on a Linux server is using UUID pathing in /etc/fstab as the device.
Below is the results of running the recipe on a test server.

/etc/fstab before:

`#`
```
# /etc/fstab
# Created by anaconda on Tue Oct 31 15:08:36 2017
#
# Accessible filesystems, by reference, are maintained under '/dev/disk'
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
#
/dev/mapper/cl-root     /                       xfs     defaults        0 0
/dev/sda1 /boot xfs defaults 0 0
/dev/mapper/cl-swap     swap                    swap    defaults        0 0
none /dev/shm                  tmpfs     rw,nosuid,nodev,noexec        0 0
/dev/mapper/vg_test-lv_test /test xfs async,auto,exec,nouser,rw,suid,nodev 0 0

```
/etc/fstab after:

```
#
# /etc/fstab
# Created by anaconda on Tue Oct 31 15:08:36 2017
#
# Accessible filesystems, by reference, are maintained under '/dev/disk'
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
#
/dev/mapper/cl-root     /                       xfs     defaults        0 0
/dev/sda1 /boot xfs defaults 0 0
/dev/mapper/cl-swap     swap                    swap    defaults        0 0
none /dev/shm                  tmpfs     rw,nosuid,nodev,noexec        0 0
/dev/mapper/vg_test-lv_test /test xfs async,auto,exec,nouser,rw,suid,nodev 0 0
UUID=1d9f79f2-e655-4a52-8456-5e3178f75aa0 /boot xfs defaults 0 0

```
[...]

PR #6969, opened 2018-03-09
update mount to use properties and fix 6851

thommay:
Fixes: #6851

### DIFF
--- a/lib/chef/provider/mount/aix.rb
+++ b/lib/chef/provider/mount/aix.rb
@@ -28,7 +28,7 @@ class Chef
           super
           # options and fstype are set to "defaults" and "auto" respectively in the Mount Resource class. These options are not valid for AIX, override them.
           if @new_resource.options[0] == "defaults"
-            @new_resource.options.clear
+            @new_resource.options([])
           end
           if @new_resource.fstype == "auto"
             @new_resource.send(:clear_fstype)
--- a/lib/chef/resource/mount.rb
+++ b/lib/chef/resource/mount.rb
@@ -28,8 +28,9 @@ class Chef
       allowed_actions :mount, :umount, :unmount, :remount, :enable, :disable
 
       # this is a poor API please do not re-use this pattern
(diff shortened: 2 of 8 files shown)


lib/chef/resource/mount.rb
```diff
@@ -24,8 +24,6 @@ class Resource
     class Mount < Chef::Resource
       description "Use the mount resource to manage a mounted file system."
 
-      identity_attr :device
```
tas50: identity needs to be set on device if this goes away

lib/chef/resource/mount.rb
```diff
@@ -24,8 +24,6 @@ class Resource
...
-
       state_attrs :mount_point, :device_type, :fstype, :username, :password, :domain
```
tas50: if we move to properties you might as well nuke this all since everything is a state attrs by default

lib/chef/resource/mount.rb
thommay: Yep, it's there :)

tas50: yep I'm blind

lib/chef/provider/mount/aix.rb
```diff
@@ -28,7 +28,7 @@ def initialize(new_resource, run_context)
           super
           # options and fstype are set to "defaults" and "auto" respectively in the Mount Resource class. These options are not valid for AIX, override them.
           if @new_resource.options[0] == "defaults"
-            @new_resource.options.clear
+            @new_resource.options([])
```
tas50: Is there a reason not to just set these correctly for AIX in the resource?

thommay: wild spaghetti code in the resource v twiddling stuff in the provider...

thommay: I mean, i think either is bad, but i think this is marginally less bad

tas50: fair enough

PR #6972, opened 2018-03-11
Remove redundant parentheses

tas50:
Rubymine was mad about these. There's a rubocop rule we should probably turn on


lamont-granquist:
pretty sure we decided not to turn this rule on, i'm definitely inclined against it since `foo()` is a nice indication that the thing is a method not a variable.

tas50:
I've been downvoted

PR #6974, opened 2018-03-11
Remove the manpages

tas50:
These are horribly outdated and we shouldn't have people using them for decision making


danielsdeleo:
/me pours one out...

I recall we also had a knife command to display these and possibly some rakefile stuff to get them built into the gem or generate them from a different format. Can you double check to make sure all that stuff is gone, too?

PR #6975, opened 2018-03-11
Use Ruby 2.4+'s .match? method

tas50:
We're not longer testing 2.3 in Travis so we should officially remove support. With that gone we should use a handy Ruby 2.4-ism: .match? It's faster than =~ because it doesn't actually store the match. This is just autocorrected from Rubocop's Performance/RegexpMatch

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -119,7 +119,7 @@ class Chef::Application::Client < Chef::Application
       :description =>
         "Daemonize the process. Accepts an optional integer which is the " \
         "number of seconds to wait before the first daemonized run.",
-      :proc => lambda { |wait| wait =~ /^\d+$/ ? wait.to_i : true }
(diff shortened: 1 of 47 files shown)


lamont-granquist:
things look busted tho

tas50:
There's going to be some cleanup to get this running, but I think it's worth doing

tas50:
Turns out there's some serious downsides to .match?. If you try to match on a nil value it blows up. =~ doesn't do that and we've coded to that in a TON of places. This isn't a change we can make unless we're sure the place where we're using it won't be nil.

```
irb(main):001:0> val = nil
=> nil
irb(main):002:0> val.match?(/something/)
NoMethodError: undefined method `match?' for nil:NilClass
	from (irb):2
	from /Users/tsmith/.rbenv/versions/2.4.3/bin/irb:11:in `<main>'
irb(main):003:0> val =~ /something/
=> nil
```

PR #6976, opened 2018-03-11
Don't use String.new in the cron provider

tas50:
This is another Rubocop performance cop. Here's their description:

```
In Ruby 2.3 or later, use unary plus operator to unfreeze a string literal instead of String#dup and String.new. Unary plus operator is faster than String#dup.

Note: String.new (without operator) is not exactly the same as +''. These differ in encoding. String.new.encoding is always ASCII-8BIT. However, (+'').encoding is the same as script encoding(e.g. UTF-8). So, if you expect ASCII-8BIT encoding, disable this cop.
```

### DIFF
--- a/lib/chef/provider/cron.rb
+++ b/lib/chef/provider/cron.rb
@@ -95,8 +95,8 @@ class Chef
       end
 
       def action_create
-        crontab = String.new
-        newcron = String.new
+        crontab = +""
+        newcron = +""
         cron_found = false
 
         newcron = get_crontab_entry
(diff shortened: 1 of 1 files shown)


coderanger:
I don't think I like this, any perf impact has got to be so tiny as to be moot given the use case and I am overall not a fan of this kind of cryptic code.

coderanger:
Like why can't it just be `crontab = ""`? I don't think we need any dup behavior here.

PR #6981, opened 2018-03-13
Add windows_adjoin resource

tas50:
Copied with permission from John Snow's win_ad cookbook. Renamed and minor changes


tas50:
Ping @stuartpreston 

What here needs to change?

lib/chef/resource/windows_adjoin.rb
```diff
@@ -0,0 +1,90 @@
...
+               required: true
+
+      property :ou_path, String,
+               description: "The path to the OU where you would like to place the host."
+
+      property :reboot, Symbol,
+               equal_to: [:immediate, :delayed, :never],
+               validation_message: "The reboot property accepts :immediate (reboot as soon as the resource completes), :delayed (reboot once the Chef run completes), and :never (Don't reboot)",
+               description: "Controls the system reboot behavior post domain joining. Reboot immediately, after the Chef run completes, or never. Note that a reboot is necessary for changes to take effect.",
+               default: :immediate
```
lamont-granquist: still think this would be easier to read:

``
      property :reboot, Symbol, equal_to: [:immediate, :delayed, :never], default: :immediate,
               validation_message: "The reboot property accepts :immediate (reboot as soon as the resource completes), :delayed (reboot once the Chef run completes), and :never (Don't reboot)",
               description: "Controls the system reboot behavior post domain joining. Reboot immediately, after the Chef run completes, or never. Note that a reboot is necessary for changes to take effect.",
```

the long ones are the ones you need new lines for, and if the shorter ones don't exceed the lengths of the longest description you might as well put those all together on the first line.

tas50: I might go shuffle them around post 14

tas50: Also I think I'm going to single line all the descriptions and just assume we're all soft wrapping because we're not doing our work on an Apple IIe

PR #6983, opened 2018-03-14
Require Ruby 2.4+

tas50:
We're removed testing on 2.3 so at this point we wouldn't know if we
introduced some Ruby 2.4+ code.


lamont-granquist:
we had discussed before dropping supporting ruby-1 on chef-client

Issue #6984, opened 2018-03-14
Authorization header isn't stripped after a redirect to a different host


bugok:
## Description
In http, the authentication header isn't stripped after a redirect to a different host.

This is how other popular http modules are doing this: 
- curl: https://github.com/curl/curl/blob/6beb0eee/lib/http.c#L710
- Python's requests module is doing it: https://github.com/request/request/pull/1184/
- gist example of how curl strips this header: https://gist.github.com/bugok/6551763aacb7b6ff2d463d2005bdb897

## Chef Version
12.21.31

## Platform Version
MacOS: 10.13.3

## Replication Case
Hitting httpbin.org with a redirect option to get redirected, observing that the auth headers are being sent upon the redirect.
```
$ cat exmaple.rb
http_request 'dummy' do
  action :get
  url 'https://httpbin.org/redirect-to?url=https://www.google.com'
  headers({'Authorization' => "Basic #{
    Base64.encode64('username:password')}",
  })
end
```
Then, I ran `chef-apply example.rb -l debug` 

## Client Output
https://gist.github.com/bugok/843f112c1171e934ce7c67c490ab3cbd
You can see that in line 414, the redirect request contains the auth header.

## Stacktrace
None

bugok:
Closing, as the fix got merged.

bugok:
As https://github.com/chef/chef/pull/6985 got reverted in https://github.com/chef/chef/pull/6996, reopening.

tas50:
This was fixed with that PR merge back in 2018

PR #6985, opened 2018-03-14
Stripping out Authorization header on redirect to a different host

bugok:
### Description

In http, strips `Authorization` header on 302 redirects, just like curl does it.

### Issues Resolved

- https://github.com/chef/chef/issues/6984

### Check List

- [ ] New functionality includes tests. (Couldn't find a way to make tiny_server return a 302 to return. Seems like https://github.com/chef/chef/blob/master/spec/functional/http/simple_spec.rb is the right place to add such a test, but I didn't find a neat way to add one. I'm open to suggestions on how to test this.)
- [ ] All tests pass - master seems to be broken. http units are in the same state as I found them.

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -381,7 +381,9 @@ class Chef
         elsif redirect_location = redirected_to(response)
           if [:GET, :HEAD].include?(method)
             follow_redirect do
-              send_http_request(method, url + redirect_location, headers, body, &response_handler)
+              redirected_url = url + redirect_location
+              headers.delete("Authorization") if url.host != redirected_url.host
+              send_http_request(method, redirected_url, headers, body, &response_handler)
             end
           else
             raise Exceptions::InvalidRedirect, "#{method} request was redirected from #{url} to #{redirect_location}. Only GET and HEAD support redirects."


lib/chef/http.rb
```diff
@@ -381,7 +381,11 @@ def send_http_request(method, url, base_headers, body, &response_handler)
         elsif redirect_location = redirected_to(response)
           if [:GET, :HEAD].include?(method)
             follow_redirect do
-              send_http_request(method, url + redirect_location, headers, body, &response_handler)
+              redirected_url = url + redirect_location
+              if url.host != redirected_url.host
+                headers.delete('Authorization') if headers['Authorization']
```
BackSlasher:
```
headers.delete('Authorization') if url.host != redirected_url.host
```
IMO, in Ruby you don't need to be that careful when removing from hashes

bugok: Fixed.

bugok:
Seems like travis is broken on master for the same reason: https://travis-ci.org/chef/chef/jobs/352934823

thommay:
> Seems like travis is broken on master for the same reason: https://travis-ci.org/chef/chef/jobs/352934823

Yeah, that's now reverted - if you rebase on to master you should be super green.

Thanks for this PR, it looks like a great fix.

rckclmbr:
Can we also get this backported to chef-12?

thommay:
> Can we also get this applied to chef-12?

chef-12 is end of life and won't get changes like this. A backport to 13 is appropriate however.

jaymzh:
:+1:

tas50:
@bugok Heads up we had to revert this since it caused failures for anyone using the http_request resource. https://github.com/chef/chef/pull/6996

bugok:
:(
I'll circle back and try to fix it in another PR.

PR #6990, opened 2018-03-15
Add swap_file resource from the swap cookbook

tas50:
This is copied as-is from the swap cookbook and included here with permission of Vargo


lib/chef/resource/swap_file.rb
```diff
@@ -0,0 +1,217 @@
...
+          rescue Mixlib::ShellOut::ShellCommandFailed => e
+            Chef::Log.warn("#{new_resource} Rescuing failed swapfile creation for #{new_resource.path}")
+            Chef::Log.debug("#{new_resource} Exception when creating swapfile #{new_resource.path}: #{e}")
+            do_create(dd_command)
+          end
+        end
+        if new_resource.swappiness
+          include_recipe "sysctl::default"
+
+          sysctl_param "vm.swappiness" do
```
coderanger: Should use `sysctl` as the new primary name.

lib/chef/resource/swap_file.rb
```diff
@@ -0,0 +1,217 @@
...
+          begin
+            Chef::Log.info "starting first create: #{node['virtualization']['system']}"
+            do_create(swap_creation_command)
+          rescue Mixlib::ShellOut::ShellCommandFailed => e
+            Chef::Log.warn("#{new_resource} Rescuing failed swapfile creation for #{new_resource.path}")
+            Chef::Log.debug("#{new_resource} Exception when creating swapfile #{new_resource.path}: #{e}")
+            do_create(dd_command)
+          end
+        end
+        if new_resource.swappiness
```
coderanger: I'm mixed on this feature. Notably if you have more than one swap, they would compete? But that's pretty rare.

lamont-granquist: Separation of concerns suggests that this option shouldn't be here.  And it creates a really weird API where this is just a facade over calling the sysctl resource to set the swappiness.

lib/chef/resource/swap_file.rb
```diff
@@ -0,0 +1,217 @@
...
+          compatible_filesystems = %w{xfs ext4}
+          parent_directory = ::File.dirname(new_resource.path)
+          # Get FS info, get second line as first is column headings
+          command = "df -PT #{parent_directory} | awk 'NR==2 {print $2}'"
+          result = shell_out(command).stdout
+          Chef::Log.debug("#{new_resource} filesystem listing is '#{result}'")
+          compatible_filesystems.any? { |fs| result.include? fs }
+        end
+
+        # we can remove this when we only support Chef 13
```
lamont-granquist: you only need to support Chef 14 now

lib/chef/resource/swap_file.rb
tas50: Done and the include is gone too

lib/chef/resource/swap_file.rb
tas50: Two fold answer:

First part: It's an optional feature and it's off by default. It makes a lot of sense to manage your swapiness when you configure your swap file. For users that want to just get swap going they can set a value and it'll just work. If they don't turn it on, then they won't get it, which seems to follow "the chef way".

Second part: I think the scenario where we have multiple swap files and the users specify different swapiness values it super edge case. You could have multiple swap files and even set the swapiness on both without issue. You could specify it on one and it would work. You just can't have 2 different values and I think it *should* be pretty clear to anyone who would want swapiness set why that isn't possible.

tas50: 3rd part: Users of the cookbook are utilizing this property so removing it breaks that API in a sad way

tas50:
Yep the recipe was a leftover from the pre-1.0 cookbook that was very much attribute driven still.

PR #6993, opened 2018-03-16
Use Chef omnibus def that includes libarchive

tas50:
This already ran through Jenkins w/o issue and gives us a build with libarchive on all platforms (even AIX!)


coderanger:
🎊✨🎊

PR #6994, opened 2018-03-16
Add more resource descriptions and convert resources to use properties

tas50:
More general resource cleanup. Add more resource descriptions and convert resources to use properties

### DIFF
--- a/lib/chef/resource/template.rb
+++ b/lib/chef/resource/template.rb
@@ -44,9 +44,6 @@ class Chef
       def initialize(name, run_context = nil)
         super
         @source = "#{::File.basename(name)}.erb"
-        @cookbook = nil
-        @local = false
-        @variables = Hash.new
         @inline_helper_blocks = {}
         @inline_helper_modules = []
         @helper_modules = []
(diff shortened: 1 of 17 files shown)


lamont-granquist:
integration tests on travis seem pissed, i don't understand what it has to do with this though

lamont-granquist:
i can't replicate the failure on master on my desktop tho...

tas50:
These failures are super odd

lib/chef/resource/template.rb
```diff
@@ -60,29 +57,9 @@ def source(file = nil)
...
-      end
-
-      def local(args = nil)
-        set_or_return(
-          :local,
-          args,
-          :kind_of => [ TrueClass, FalseClass ]
-        )
-      end
+      property :variables, Hash, default: {}
```
lamont-granquist: this is a frozen hash default.

```ruby
template "whatever.erb" do
   variables[:aliases] ||= {}
end
```

That mutates the Hash which is the default value.

You need to get a new Hash object for every new template resource.

Frozen default values probably just caught a huge bug here.

lamont-granquist: `default: lazy { Hash.new }` maybe?

lamont-granquist: yeah validated that produces different objects.

i'd almost go so far as to say that `default: {}` is always a bug and should be replaced with this (and any default hash, even with default keys should be lazied this way).

PR #6980, opened 2018-03-13
Pass pointer to LsaFreeMemory, not FFI::MemoryPointer

btm:
LsaFreeMemory takes a pointer to the struct in memory created by LsaEnumerateAccountRights.
We have to pass it the actual pointer, not the FFI::MemoryPointer object.

This fixes masked access violations which occasionally lead to a heap corruption which
presents itself as silent termination of chef-client/ruby with a return code of -1073740940
or STATUS_HEAP_CORRUPTION / 0xc0000374.

Fixes #6589

PR #6995, opened 2018-03-16
Fix regression in #6980, add functional tests

btm:
The logging refactor in #6980 should have been a class variable. This fixes that.

Also adds functional tests that would catch that and the original #6980 bug.

### DIFF
--- a/lib/chef/win32/security.rb
+++ b/lib/chef/win32/security.rb
@@ -668,7 +668,7 @@ class Chef
         Token.new(Handle.new(token.read_pointer))
       end
 
-      def test_and_raise_lsa_nt_status(result)
+      def self.test_and_raise_lsa_nt_status(result)
         win32_error = LsaNtStatusToWinError(result)
         if win32_error != 0
           Chef::ReservedNames::Win32::Error.raise!(nil, win32_error)
--- a/spec/functional/win32/security_spec.rb
+++ b/spec/functional/win32/security_spec.rb
@@ -52,6 +52,7 @@ describe "Chef::Win32::Security", :windows_only do
       delete_user.run_command
       delete_user.error!
     end
+
     it "has_admin_privileges? returns false" do
       has_admin_privileges = with_user_context(user, password, domain, :local) do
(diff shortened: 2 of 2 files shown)


spec/functional/win32/security_spec.rb
```diff
@@ -149,4 +150,32 @@
       end
     end
   end
+
+  describe "self.get_account_right" do
```
rhass: Usually the name for this describe block would be ".get_account_right" for a class method. Eg: http://www.betterspecs.org/#describe

lib/chef/win32/security.rb
```diff
@@ -668,7 +668,7 @@ def self.logon_user(username, domain, password, logon_type, logon_provider)
         Token.new(Handle.new(token.read_pointer))
       end
 
-      def test_and_raise_lsa_nt_status(result)
+      def self.test_and_raise_lsa_nt_status(result)
```
rhass: What's the reason for changing this to a class method?

spec/functional/win32/security_spec.rb
```diff
@@ -149,4 +150,32 @@
...
+    context "when given an invalid username" do
+      let(:username) { "noooooooooope" }
+
+      it "raises an exception" do
+        expect { Chef::ReservedNames::Win32::Security.get_account_right(username) }.to raise_error(Chef::Exceptions::Win32APIError)
+      end
+    end
+  end
+
+  describe "self.test_and_raise_lsa_nt_status" do
```
rhass: Same here: `.test_and_raise_lsa_nt_status" rather than `self.test_and_raise_lsa_nt_status`

lib/chef/win32/security.rb
btm: All the other methods are class methods, we never make instances of this class, so this has to be a class method to be called.

tas50:
There's a specs failure in AppVeyor that needs to be fixed

btm:
This change should fix the appveyor use case.

PR #6996, opened 2018-03-16
Revert "Stripping out Authorization header on redirect to a different host

tas50:
This introduced a failure when using the http_request resource that we experienced internally using Chef 14 builds. 

```

    Error executing action `delete` on resource 'http_request[delete-indicies]'

    FrozenError
    -----------
    can't modify frozen Hash

    Resource Declaration:
    ---------------------
    313:   http_request 'delete-indicies' do
    314:     url 'http://localhost:9200/_all'
    315:     action :delete
    316:   end
    317:

```

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -381,9 +381,7 @@ class Chef
         elsif redirect_location = redirected_to(response)
           if [:GET, :HEAD].include?(method)
             follow_redirect do
-              redirected_url = url + redirect_location
-              headers.delete("Authorization") if url.host != redirected_url.host
-              send_http_request(method, redirected_url, headers, body, &response_handler)
+              send_http_request(method, url + redirect_location, headers, body, &response_handler)
             end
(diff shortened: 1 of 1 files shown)


lamont-granquist:
the actual fix here probably needs to not only dup the hash before deleting the key, but someone out there will want auth headers to get preserved (maybe across localhost to 127.0.0.1 redirects, DNS cnames, or something like that...)

bugok:
@lamont-granquist, @tas50: Sorry for the breakage. Had I known how to test this it better, I would have.
I would like to submit another will request with a better implementation.

From reading the comment, I understand two things:
1. After deleting the header - I need to duplicate the hash. What hash? How can I do this?
2. I should make deleting the header optional. Should it be opt-in or opt-out? Does adding a property named `pass_auth_header_on_redirect` makes sense? Is there another suggestion?

Again, sorry for the mess.

PR #7005, opened 2018-03-18
Add missing installed logic for macos in build_essential

tas50:
Add the missing installed? check that makes this fast. Also use tail
instead of head in the execute so we install the latest not the oldest
xcode release if more than 1 is available on the system (upgrades).

### DIFF
--- a/lib/chef/resource/build_essential.rb
+++ b/lib/chef/resource/build_essential.rb
@@ -47,21 +47,23 @@ class Chef
           declare_resource(:package,  "devel/m4")
           declare_resource(:package,  "devel/gettext")
         when "mac_os_x"
-          # This script was graciously borrowed and modified from Tim Sutton's
-          # osx-vm-templates at https://github.com/timsutton/osx-vm-templates/blob/b001475df54a9808d3d56d06e71b8fa3001fff42/scripts/xcode-cli-tools.sh
-          declare_resource(:execute, "install XCode Command Line tools") do
-            command <<-EOH.gsub(/^ {14}/, "")
-              # create the placeholder file that's checked by CLI updates' .dist code
-              # in Apple's SUS catalog
-              touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
-              # find the CLI Tools update
-              PROD=$(softwareupdate -l | grep "\*.*Command Line" | head -n 1 | awk -F"*" '{print $2}' | sed -e 's/^ *//' | tr -d '\n')
-              # install it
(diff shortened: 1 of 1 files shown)


lib/chef/resource/build_essential.rb
```diff
@@ -112,6 +114,22 @@ class BuildEssential < Chef::Resource
...
+      action_class do
+        #
+        # Determine if the XCode Command Line Tools are installed
+        #
+        # @return [true, false]
+        #
+        def xcode_cli_installed?
+          cmd = Mixlib::ShellOut.new("pkgutil --pkgs=com.apple.pkg.CLTools_Executables")
+          cmd.run_command
+          cmd.error!
```
lamont-granquist: `cmd.error?` rather than raise-and-rescue?

lib/chef/resource/build_essential.rb
```diff
@@ -47,21 +47,23 @@ class BuildEssential < Chef::Resource
...
+                # in Apple's SUS catalog
+                touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
+                # find the CLI Tools update. We tail here because sometimes there's 2 and newest is last
+                PROD=$(softwareupdate -l | grep "\*.*Command Line" | tail -n 1 | awk -F"*" '{print $2}' | sed -e 's/^ *//' | tr -d '\n')
+                # install it
+                softwareupdate -i "$PROD" --verbose
+                # Remove the placeholder to prevent perpetual appearance in the update utility
+                rm -f /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
+              EOH
+              # rubocop:enable Metrics/LineLength
```
lamont-granquist: don't need this

tas50: Fixed

lib/chef/resource/build_essential.rb
```diff
@@ -112,6 +113,19 @@ class BuildEssential < Chef::Resource
...
+      action_class do
+        #
+        # Determine if the XCode Command Line Tools are installed
+        #
+        # @return [true, false]
+        #
+        def xcode_cli_installed?
+          cmd = Mixlib::ShellOut.new("pkgutil --pkgs=com.apple.pkg.CLTools_Executables")
+          cmd.run_command
+          cmd.error? ? false : true
```
thommay: I think that deserves a comment.

tas50: Done

PR #7011, opened 2018-03-19
Fix regression in #6980, add functional tests

btm:
The logging refactor in #6980 should have been a class variable. This fixes that.

Also adds functional tests that would catch that and the original #6980 bug.


btm:
travis failure is unrelated, has been failing in `chef-12` too.

PR #7012, opened 2018-03-19
Deprecate vendor branches with `knife supermarket install`

lamont-granquist:
closes #6467 just a rebase

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -196,7 +196,7 @@ class Chef
       target 19
     end
 
-    class CookbookInstallVendor < Base
+    class SupermarketInstallVendor < Base
       target 20
     end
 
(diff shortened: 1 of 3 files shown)


lib/chef/knife/cookbook_site_install.rb
```diff
@@ -67,12 +67,24 @@ class CookbookSiteInstall < Knife
...
       attr_reader :cookbook_name
       attr_reader :vendor_path
 
       def run
         extend Chef::Mixin::ShellOut
 
+        # Deprecation warning for vendor installs.
+        unless config[:no_vendor]
+          Chef.deprecated(:cookbook_install_vendor, "Using knife cookbook site install with vendor branching is deprecated. " \
+            "Please pass --no-vendor or add `knife[:no_vendor] = true` to your knife configuration.")
```
tas50: Is the idea that in Chef 15 we just remove this option and make it the default? If so we might want to mention that here.

lamont-granquist:
Deprecating it has proven to be too difficult, we are just going to remove the feature without deprecation.

tas50:
Closing this out since we're just going to yank it instead.

PR #7018, opened 2018-03-21
Prefer proxy configs within client.rb over environment variables

jakauppila:
### Description

Today, if you define any of the following configurations within `client.rb` they are subsequently ignored on the Chef Client execution if the environment variable(s) are already defined. 

```
http_proxy
https_proxy
ftp_proxy
no_proxy
```
I would expect the configuration as defined within the `client.rb` to take precedence over the environment variables of the machine. As it stands today, you can't override the machine environment variables.

### Issues Resolved

I ran across this issue using Kitchen where my host machine has proxy environment variables defined and was overriding the values I had defined within my guests `client.rb` configuration.

https://github.com/test-kitchen/test-kitchen/issues/1366

### Check List

- [ ] All tests pass

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -979,14 +979,14 @@ module ChefConfig
       end
 
       path = uri.to_s
-      ENV["#{scheme}_proxy".downcase] = path unless ENV["#{scheme}_proxy".downcase]
-      ENV["#{scheme}_proxy".upcase] = path unless ENV["#{scheme}_proxy".upcase]
+      ENV["#{scheme}_proxy".downcase] = path
+      ENV["#{scheme}_proxy".upcase] = path
     end
 
     # @api private
     def self.export_no_proxy(value)
-      ENV["no_proxy"] = value unless ENV["no_proxy"]
-      ENV["NO_PROXY"] = value unless ENV["NO_PROXY"]
+      ENV["no_proxy"] = value
+      ENV["NO_PROXY"] = value
     end
 
     # Given a scheme, host, and port, return the correct proxy URI based on the


jakauppila:
In terms of updating the tests, do I modify the existing tests so that we have the environment variables defined? Or do I just clone the entire `export_proxies` section where values already exist?

lamont-granquist:
Pretty certain we did this deliberately because you can just do `ENV['http_proxy'] = "whatever"` in the client.rb because its just ruby.  The only time this comes up is when you've got those env var set externally but client.rb sets something different, and in that case i think i'm okay with forcing the user to do more work in client.rb because that seems like a very unusual edge case.

jakauppila:
Reading through the wording again, it does say that it will set the value if not already defined. But from a user perspective, if there's a proxy config option in a configuration file I would expect that to override any system settings; otherwise why are those even an option? At that point, it would probably be better just to instruct users to set the environment variables directly as you mentioned.

**Automatic Proxy Config**
If `http_proxy`, `https_proxy`, `ftp_proxy`, or `no_proxy` is set in the client.rb file and _is not already set in the ENV_, the chef-client will configure the ENV variable based on these (and related) settings.

lamont-granquist:
The problem is that we generally do not overwrite your ENV vars and preserve them so that "external" commands (like firing up an actual app service) do not get polluted with ENV settings that are critical to the internals of chef-client (e.g. when we fire off ifconfig or netstat or something inside of a provider and parse the output).

Mostly this gets used in the case of LANG settings.  If the external environment is de_DE.UTF-8 or something like that we have to force that to en_US.UTF-8 internally because we have to get English strings back from shell utilities in order to parse them correctly.  But if we are launching someone's rails or nodejs app we need to preserve their settings  (which could also be Latin-1 or some of the odd non-UTF-8 japanese encodings that are still kicking around).

This is the same thing here, which is that if someone has explicitly set a proxy variable in their external environment then we have to preserve that ENV setting by default, otherwise when they start a service from chef, they will get the chef proxy setting, while if they start a service from the cli they will get the external proxy variable.  That is a really weird kind of problem to debug.

lamont-granquist:
It might help to have a better description of why this is necessary and how you stumbled across it.  More or less we're missing a traditional user-story here, and I'm have a very difficult time guessing at it.

lamont-granquist:
Closing due to inactivity.

If you've got proxy ENV vars in your system which are incorrect, you can explicitly overwrite them in your config.rb, which I think is the correct solution.  Having the client decide to stomp env vars by default is probably going to hurt some other use case someone has and make that difficult (particularly if chef-client is supposed to go through one proxy setting, while services which are started should go through the default proxy on the box, and if we stomp the ENV vars now those people need to decorate all their service resources with the proxy settings which are just the external settings on the host -- making that use case of the service resource not what people would expect seems wrong).  If your external env var is incorrect, I think its appropriate to require more typing and config.

PR #7021, opened 2018-03-21
Add caching to windows_feature_powershell and fix array handling in windows feature

tas50:
These fixes are in the 4.1 version of the windows cookbook

### DIFF
--- a/lib/chef/resource/windows_feature_dism.rb
+++ b/lib/chef/resource/windows_feature_dism.rb
@@ -55,7 +55,7 @@ class Chef
         unless features_to_install.empty?
           message = "install Windows feature#{'s' if features_to_install.count > 1} #{features_to_install.join(',')}"
           converge_by(message) do
-            install_command = "#{dism} /online /enable-feature #{features_to_install.map { |f| "/featurename:#{f}" }.join(' ')} /norestart"
+            install_command = "dism.exe /online /enable-feature #{features_to_install.map { |f| "/featurename:#{f}" }.join(' ')} /norestart"
             install_command << " /LimitAccess /Source:\"#{new_resource.source}\"" if new_resource.source
             install_command << " /All" if new_resource.all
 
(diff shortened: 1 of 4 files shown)


lib/chef/resource/windows_feature_dism.rb
```diff
@@ -55,7 +55,7 @@ class WindowsFeatureDism < Chef::Resource
         unless features_to_install.empty?
           message = "install Windows feature#{'s' if features_to_install.count > 1} #{features_to_install.join(',')}"
           converge_by(message) do
-            install_command = "#{dism} /online /enable-feature #{features_to_install.map { |f| "/featurename:#{f}" }.join(' ')} /norestart"
+            install_command = "dism.exe /online /enable-feature #{features_to_install.map { |f| "/featurename:#{f}" }.join(' ')} /norestart"
```
lamont-granquist: is that deliberate?  reads like a step backwards

lib/chef/resource/windows_feature_powershell.rb
```diff
@@ -57,77 +58,170 @@ class WindowsFeaturePowershell < Chef::Resource
...
+          unavailable = (new_resource.feature_name - all_available)
+          raise "The Windows feature#{'s' if unavailable.count > 1} #{unavailable.join(',')} #{unavailable.count > 1 ? 'are' : 'is'} not available on this version of Windows. Run 'Get-WindowsFeature' to see the list of available feature names." unless unavailable.empty?
+        end
+
+        # run Get-WindowsFeature to get a list of all available features and their state
+        # and save that to the node at node.override level.
+        # @return [void]
+        def reload_cached_powershell_data
+          Chef::Log.debug("Caching Windows features available via Get-WindowsFeature.")
+          node.override["powershell_features_cache"] = Mash.new
```
lamont-granquist: This shouldn't go in the node, it should really go into a singleton object, similar in principle to the old YumCache

lib/chef/resource/windows_feature_dism.rb
tas50: the dism method was from the cookbook and came over when I was diffing them by accident. It was to support funky 32bit vs. 64bit stuff. We discussed a while back that we could ditch this logic.

tas50:
Closing this one out since I brought the bugfixes in #7078. We need to get this fixed for Chef 14, but it can be done with the singleton method in another PR

PR #7022, opened 2018-03-21
Add sysctl_param resource from the sysctl cookbook

tas50:
Copied from the cookbook with modifications made there


lib/chef/resource/sysctl_param.rb
```diff
@@ -0,0 +1,123 @@
...
+        o = shell_out("sysctl -n -e #{key}")
+        raise "Unknown sysctl key!" if o.error!
+        o.stdout.to_s.tr("\t", " ").strip
+      end
+
+      load_current_value do
+        value get_sysctl_value(key)
+        if node.normal["sysctl"]["backup"][key].empty?
+          node.normal["sysctl"]["backup"][key] = value
+        end
```
lamont-granquist: i still think this is the weirdest feature ever.

this key could be saved on the node for three years and then `:removed`, and who the hell knows what is going to happen...

`action :remove` as a concept feels like "you're doing config management wrong".

lib/chef/resource/sysctl_param.rb
```diff
@@ -0,0 +1,123 @@
...
+# limitations under the License.
+#
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class SysctlParam < Chef::Resource
+      resource_name :sysctl_param
+      provides :sysctl_param
```
lamont-granquist: I sort of think this should be:

```
resource_name :sysctl
provides :sysctl
provides :sysctl_param  # back compat with the cookbook
```

lib/chef/resource/sysctl_param.rb
```diff
@@ -0,0 +1,123 @@
...
+        when Array
+          v.join(" ")
+        else
+          v.to_s
+        end
+      end
+
+      def get_sysctl_value(key)
+        o = shell_out("sysctl -n -e #{key}")
+        raise "Unknown sysctl key!" if o.error!
```
coderanger: Should include the key name in the error message so it is clearer.

lib/chef/resource/sysctl_param.rb
coderanger: I think `action :remove` is fine to clean up the underlying files but I too am unsure why we are storing this stuff into node attributes. What problem does this solve?

lib/chef/resource/sysctl_param.rb
coderanger: Also even if we don't want to have two names, still don't need both `resource_name` and `provides` for the same sym.

lib/chef/resource/sysctl_param.rb
```diff
@@ -0,0 +1,123 @@
...
+          set_sysctl_param(new_resource.key, new_resource.value)
+
+          directory new_resource.conf_dir
+
+          file "#{new_resource.conf_dir}/99-chef-#{new_resource.key}.conf" do
+            content "#{new_resource.key} = #{new_resource.value}"
+          end
+
+          execute "sysctl -p" do
+            command "sysctl -p"
```
coderanger: Redundant command and action.

lib/chef/resource/sysctl_param.rb
tas50: I pushed a rename commit

lib/chef/resource/sysctl_param.rb
tas50: Fixed

lib/chef/resource/sysctl_param.rb
tas50: Fixed

lib/chef/resource/sysctl.rb
```diff
@@ -0,0 +1,123 @@
...
+          set_sysctl_param(new_resource.key, new_resource.value)
+
+          directory new_resource.conf_dir
+
+          file "#{new_resource.conf_dir}/99-chef-#{new_resource.key}.conf" do
+            content "#{new_resource.key} = #{new_resource.value}"
+          end
+
+          execute "sysctl -p" do
+            command "sysctl -p"
```
coderanger: Still redundant. Doesn't hurt anything but easy to DRY.

lib/chef/resource/sysctl.rb
```diff
@@ -0,0 +1,123 @@
...
+        when Array
+          v.join(" ")
+        else
+          v.to_s
+        end
+      end
+
+      def get_sysctl_value(key)
+        o = shell_out("sysctl -n -e #{key}")
+        raise "Unknown sysctl key #{key}!" if o.error!
```
jaymzh: `error!` raises. so this is ... a bit meaningless :) did you mean `error?`

lib/chef/resource/sysctl.rb
```diff
@@ -0,0 +1,123 @@
...
+        o.stdout.to_s.tr("\t", " ").strip
+      end
+
+      load_current_value do
+        value get_sysctl_value(key)
+        if node.normal["sysctl"]["backup"][key].empty?
```
jaymzh: Don't we discourage `normal` ?

lib/chef/resource/sysctl.rb
tas50: Oh nice that never worked in the cookbook either. I'll get that fixed

PR #7023, opened 2018-03-21
rubocop fixes from engine bump to 0.54.0

lamont-granquist:
some cops got a little stricter.

### DIFF
--- a/spec/unit/knife/osc_user_reregister_spec.rb
+++ b/spec/unit/knife/osc_user_reregister_spec.rb
@@ -19,7 +19,7 @@
 require "spec_helper"
 
 # DEPRECATION NOTE
-# This code only remains to support users still	operating with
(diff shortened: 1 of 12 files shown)


spec/unit/knife/osc_user_reregister_spec.rb
```diff
@@ -19,7 +19,7 @@
 require "spec_helper"
 
 # DEPRECATION NOTE
-# This code only remains to support users still	operating with
+# This code only remains to support users still  operating with
```
tas50: these double space autocorrects are funky, but whatevs

lamont-granquist: yeah they're tabs->spaces.  less awful than having hard tab stops there.  will block them in the future.

PR #7031, opened 2018-03-22
partially revert 61e3d4bb: do not use properties for mount

thommay:


thommay:
This is due to #7037; if we change to properties we'll stop accepting nil values to mean default, which will break at least the llvm cookbook

Issue #4284, opened 2015-12-11
Make copying resource properties easier


jkeiser:
Right now, resources that don't have a default value and don't allow `nil` have to be copied over like this:

```
class Blah < Chef::Resource
  resource_name 'blah'
  property :x, String
end

hi = blah 'hi' do
  x 'boo!'
end
lo = blah 'lo'
lo.x = hi.x if hi.property_is_set?(:x)
```

Otherwise you get a `nil` property warning.

We should consider making `nil` be an official sentinel value for "not set" for any resource that does not specify `nil` as its value. Passing `nil` to the property would set it to `nil` for reals. This means getting rid of the `nil` assignment warning as well.

lamont-granquist:
I really think the syntax should just be:

```ruby
# here I mean that nil is the default/unset value
property :version, [ Integer, nil ]
# here I don't allow nil at all and distinguish between unset and should-be-cleared
property :version, [ Integer, :default, :clear ], default: :default
```

And we should maintain the 'tried to set nil but didn't include nil in the validations' warning, and in Chef 13 should make passing nil set nil for reals.

In other words, I don't think we need to change anything here that we weren't planning on doing.

coderanger:
What should the behavior be if `nil` isn't an allowed value but no default is given? Make that a declaration error?

lamont-granquist:
I'm agnostic.  I'd say leave it so that it raises as a validation error on assignment and let it bake for awhile.

Then when we see how its used, either that becomes a declaration error, or we go down a route of automatically adding the default value to validations (or only do that in the default-is-nil case), or we just leave it alone...

Would people actually correctly want to use the ability to ban setting a property to the default value?  Is that a valid use case?  I sorta feel like not, but I sorta don't want to be that proscriptive...

lamont-granquist:
I guess we could always bake the default into the validation and then have an optional opt-out -- which makes the common case short to type, and the uncommon case longer to type...

```
# setting nil passes validation
property :version, Integer
# setting :unset will pass validation (nil will not)
property :version, Integer, default: :unset
# setting nil will raise a validation exception
property :version, Integer, default_passes_validation: false
```

jkeiser:
[...]

PR #7037, opened 2018-03-23
Setting nil to properties with implicit nil sets default value

lamont-granquist:
- This makes converting core resources to properties safer

- This makes it easier to apply wrapping properties to subresources

property :foo, String, default: "foo"

This is where the change lies, and writing a nil here will now actually
write a "foo" to the variable.

property :foo, [ String, nil ], default: "foo"

This is unchanged.  Writing nil writes nil.

property :foo, String

Technically this is changed, since it writes the default value, but
since nil.equal?(nil) in a very deep way no behavior changes.

### DIFF
--- a/lib/chef/mixin/params_validate.rb
+++ b/lib/chef/mixin/params_validate.rb
@@ -131,10 +131,6 @@ class Chef
 
       private
 
-      def explicitly_allows_nil?(key, validation)
-        validation.has_key?(:is) && _pv_is({ key => nil }, key, validation[:is], raise_error: false)
-      end
-
       def _validation_message(key, default)
         @validation_message.has_key?(key) ? @validation_message[key] : default
       end
@@ -321,6 +317,7 @@ class Chef
       #   ```
       #
       def _pv_callbacks(opts, key, callbacks)
+        puts caller
         raise ArgumentError, "Callback list must be a hash!" unless callbacks.kind_of?(Hash)
         value = _pv_opts_lookup(opts, key)
         if !value.nil?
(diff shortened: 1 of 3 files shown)


lamont-granquist:
this is revisiting https://github.com/chef/chef/issues/4284 for the history on this discussion.

coderanger:
I think I'm overall -1 on this. This is basically re-neg'ing on the previous plan to have a unified behavior that if you set `nil`, it always meant "set the value to `nil`". I agree there are probably some bad behaviors with old code, but I think we should at a minimum outline which use cases we are targeting specifically, and then decide if we think those are worth breaking the consistency of the current behavior. I'm not sure on that, but I like consistency so I lean towards it not being worth it unless this will break way more stuff than I think it will.

thommay:
This is the best half way house we can come up with - if you explicitly accept nil, then we do what you tell us. But if not, and you pass nil anyway, then clearly that's the same as leaving the property unset, which should cause it to be default when there's a default.

lamont-granquist:
I think I'm having regret about this particular PR as well.  Back in #4284 we made the decision that code like that does need to break and needs to get changed (and we're shipping a dot-0 which is an appropriate time to do this).  We can do something similar to #7046 to make writing this kind of code easier in the future.

coderanger:
If the concern is specifically with the one property on the `mount` resource, we can definitely fix this on the other side (i.e. in the mount provider code, make a helper method that does `new_resource.point || new_resource.default_point` or similar). I would much rather magic behavior for compat in one place rather than everywhere.

Issue #7047, opened 2018-03-23
Convert windows_pagefile to powershell


tas50:
Instead of running wmic.exe and parsing the results we should just use powershell to make the WMI calls. It should be much easier to parse and is probably significantly faster when paired with the new powershell out work.

johnmccrae:
https://help.pdq.com/hc/en-us/community/posts/115002970512-Powershell-script-to-change-Pagefile-Virtual-memory-size-on-remote-computers-using-PDQ-deploy-

$settings = Get-CimInstance -Class Win32_PageFileUsage | Select-Object *
(Get-CimInstance -Class Win32_PageFileUsage | Select-Object *).AllocatedBaseSize
Get-CimInstance -ClassName CIM_StorageVolume -Filter "Name='$($DL):\\'" -ErrorAction Stop

# experimental
Get-CimInstance -ClassName Win32_PageFileSetting -Filter "SettingID='pagefile.sys @ $($DriveLetter):'" -ErrorAction Stop | Set-CimInstance -Property @{InitialSize = $InitialSize ;MaximumSize = $MaximumSize ;} -ErrorAction Stop

# working? But it doesn't set the size... hmmm
#  Set automanage pagefile to false; min/max pagefile size
$computerSystem = Get-WmiObject -Class Win32_ComputerSystem -EnableAllPrivileges
$computerSystem.AutomaticManagedPagefile = $false
$computerSystem.Put() | Out-Null

$pageFileSetting = Get-WmiObject -Class Win32_PageFileSetting
$pageFileSetting.InitialSize = $initialsize
$pageFileSetting.MaximumSize = $maxsize
$pageFileSetting.Put() | Out-Null

PR #7048, opened 2018-03-23
Remove path helpers from window_pagefile

tas50:
I backported locate_sysnative_cmd initially, but after chatting with Stuart this isn't something we need anymore. This removes it and instead uses wmic.exe directly, which in testing works just fine. This also removes the windows friendly path helper which is really just a gsub on the slashes. Instead I changed the coerce to get us what we need in terms of path formats. I copied this to a Windows 2016 box and it works fine with both formats of paths. I'm making this same change to the windows cookbook.


tas50:
Longterm fix for dism.exe is #7047

PR #7051, opened 2018-03-24
Ship InSpec 2

thommay:


jquick:
Manhattan fixed with the Gemfile update:
http://manhattan.cd.chef.co/job/chef-trigger-ad_hoc/91/downstreambuildview/

kitchen-tests/.kitchen.travis.yml
```diff
@@ -35,6 +35,7 @@ platforms:
     pid_one_command: /sbin/init
     intermediate_instructions:
       - RUN yum -y install sudo
+      - RUN yum -y install gcc-c++
```
tas50: Can you combine the 2 yum lines and add a comment on why we're adding gcc-c++. This reduces the number of layers that have to be created

kitchen-tests/.kitchen.travis.yml
```diff
@@ -116,6 +121,8 @@ platforms:
     pid_one_command: /bin/systemd
     intermediate_instructions:
       - RUN /usr/bin/zypper install -y sudo
+      - RUN /usr/bin/zypper install -y gcc-c++
+
```
tas50: that extra line can go away while you're in there

jquick:
Thanks @tas50. I think this is good to go.

PR #7055, opened 2018-03-26
Parser 2.5.0.4 was yanked

tas50:
We missed 2 changelog updates due to this so I'm manually adding them


lamont-granquist:
It has been [0] days since the last 'gem yank' incident.

PR #7079, opened 2018-03-27
windows_feature_dism: Be case insensitive with feature names

tas50:
We took what they user gave us and compared it to the list of available
packages from dism.exe. In doing so we were case sensitive, but the CLI
doesn't actually care. This just downcases what the user gives us and
what dism gives us so it doesn't matter.

This same change is in windows cookbook 4.1.2


tas50:
So here's the problem: This works great, but not on Windows 2008R2. I think this is the right thing to do, but we need to be aware of that and probably amend the release notes to note that.

tas50:
The last commit there fixes the issues with windows 2008 r2.

PR #7082, opened 2018-03-28
[windows_ad_join] add description for :join action

brewn:
Signed-off-by: brewn <nbrewer@chef.io>

### DIFF
--- a/lib/chef/resource/windows_ad_join.rb
+++ b/lib/chef/resource/windows_ad_join.rb
@@ -57,7 +57,7 @@ class Chef
                default: true
 
       action :join do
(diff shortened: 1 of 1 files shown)


lib/chef/resource/windows_ad_join.rb
```diff
@@ -57,6 +57,8 @@ class WindowsAdJoin < Chef::Resource
                default: true
 
       action :join do
+        description "Default. Join the Active Directory domain."
```
tas50: You can take the default part out. We'll figure that using a bit of logic in the code.

PR #7087, opened 2018-03-29
Add basic hostname validation on Windows

tas50:
alphanumeric, -, . or _
Starts with an alphanumeric

We could get a lot more complex here and that might not be a bad idea, but this is probably a step in the right direction.

### DIFF
--- a/lib/chef/resource/hostname.rb
+++ b/lib/chef/resource/hostname.rb
@@ -11,8 +11,6 @@ class Chef
       introduced "14.0"
 
       property :hostname, String,
-               regex: /^([A-Za-z0-9])+([A-Za-z0-9-._])*$/, # http://rubular.com/r/c5xKppWX4i
-               validation_message: "Hostnames can contain alphanumeric characters, -, ., and _ ,but must start with an alphanumeric character.",
                description: "The hostname if different than the resource's name",
                name_property: true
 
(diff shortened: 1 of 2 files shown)


lamont-granquist:
hrm, isn't this one of those things that we should typically let the underlying operating system just fail on and surface the error?  similar to the username and group regexps?

tas50:
It might fail on Linux, but it's pretty clear you can pass bogus data to windows the way we're doing it.

tas50:
At the very least we're going to have to have some Windows specific regex because we'll just reboot forever trying to set the hostname

lamont-granquist:
Yeah that might be better to clean up windows since it explodes the server if you don't.

tas50:
Round 2:

- Moves ohai into non-windows only since we don't use it on windows
- Raises if you put a period in the hostname on windows

lib/chef/resource/hostname.rb
```diff
@@ -203,8 +203,9 @@ def updated_ec2_config_xml
           end
 
         else # windows
+          raise "Windows hostnames cannot contain a period." if new_resource.hostname.match?(/./)
 
-        # suppress EC2 config service from setting our hostname
```
tas50: rubocop totally fails to align comments btw. it's quite odd

stuartpreston:
Here’s the reference I use for naming on Windows: https://support.microsoft.com/en-gb/help/909264/naming-conventions-in-active-directory-for-computers-domains-sites-and

This PR looks like a good gate stop though to catch the most likely cause of failure.

tas50:
@stuartpreston I started with that doc and I still think we could benefit from a nice validation here, but it would potentially get a bit tricky since it's different on windows vs *nix

PR #7104, opened 2018-04-02
scrub tempfile names

### LINKED ISSUE #7100: Illegal UTF character test fails after Ruby 2.5.1 upgrade
After the update to Ruby 2.5.1 the following test in the kitchen travis tests 

```ruby
# this caught a regression in 12.14.70 before it was released when i
# ran it in lamont-ci, so added the test here so everyone else other than
# me gets coverage for this as well.
file "/tmp/chef-test-\xFDmlaut" do
  content "testing illegal UTF-8 char in the filename"
end
```

The failure:
```
  * file[/tmp/chef-test-�mlaut] action create
    
    Error executing action `create` on resource 'file[/tmp/chef-test-�mlaut]'
    
    ArgumentError
    -------------
    invalid byte sequence in UTF-8
    
    Resource Declaration:
    ---------------------
    # In /opt/kitchen/cache/cookbooks/base/recipes/tests.rb
    
     19: file "/tmp/chef-test-\xFDmlaut" do
     20:   content "testing illegal UTF-8 char in the filename"
     21: end
    
    Compiled Resource:
    ------------------
    # Declared in /opt/kitchen/cache/cookbooks/base/recipes/tests.rb:19:in `from_file'
    
    file("/tmp/chef-test-�mlaut") do
      action [:create]
      declared_type :file
      cookbook_name "base"
      recipe_name "tests"
      content "testing illegal UTF-8 char in the filename"
      owner nil
      group nil
      mode nil
      path "/tmp/chef-test-\xFDmlaut"
      verifications []
    end
    
    System Info:
    ------------
    chef_version=14.0.183
    platform=amazon
    platform_version=2
    ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
    program_name=/opt/chef/bin/chef-client
    executable=/opt/chef/bin/chef-client
```

Note: This didn't fail in the integration tests in Jenkins. We probably want to get this test into the integration tests so we could fail before promotion.

lamont-granquist:
this keeps Tempfile.open from internally barfing on ruby-2.5.1

also bumps the ruby version in the travis matrix for testing.

closes #7100

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,31 +36,31 @@ matrix:
   include:
   - env:
       INTEGRATION_SPECS_24: 1
-    rvm: 2.4.3
+    rvm: 2.4.4
     sudo: true
     script: sudo -E $(which bundle) exec rake spec:integration;
     bundler_args: --without ci docgen guard integration maintenance omnibus_package --frozen
   - env:
       INTEGRATION_SPECS_25: 1
-    rvm: 2.5.0
+    rvm: 2.5.1
     sudo: true
     script: sudo -E $(which bundle) exec rake spec:integration;
     bundler_args: --without ci docgen guard integration maintenance omnibus_package --frozen
   - env:
       FUNCTIONAL_SPECS_24: 1
-    rvm: 2.4.3
+    rvm: 2.4.4
     sudo: true
     script: sudo -E $(which bundle) exec rake spec:functional;
     bundler_args: --without ci docgen guard integration maintenance omnibus_package --frozen
   - env:
       FUNCTIONAL_SPECS_25: 1
-    rvm: 2.5.0
+    rvm: 2.5.1
     sudo: true
     script: sudo -E $(which bundle) exec rake spec:functional;
     bundler_args: --without ci docgen guard integration maintenance omnibus_package --frozen
   - env:
       UNIT_SPECS_24: 1
-    rvm: 2.4.3
+    rvm: 2.4.4
     sudo: true
     script:
       - sudo -E $(which bundle) exec rake spec:unit;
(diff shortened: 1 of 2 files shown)


robbkidd:
Is it sensible to update `spec/functional/resource/file_spec` and/or its file relations to test for this? Or is it only a problem when dealing with temporary files?

If I put `\xFD` into `file_base` for the file resource specs, I get:

```
Errno::EILSEQ:
  file[/Users/rkidd/work/chef/spec/data/test-dir/file_?_spëc20180330-36293-1x0jo4r]
  (dynamically defined) had an error: Errno::EILSEQ: Illegal byte sequence @ rb_sysopen - 
  /Users/rkidd/work/chef/spec/data/test-dir/file_?_spëc20180330-36293-1x0jo4r
```

Which, I admit, is different than the `ArgumentError: invalid byte sequence in UTF-8` we're seeing in the integration test. 🤔

lamont-granquist:
probably a good idea?  i threw it into the kitchen tests real quick just to get some kind of coverage, getting it into the func specs would probably be better.  is that failure a failure in sysopen coming from chef-client code or only from the spec code though?  (i.e. is that a real bug or a test bug?)

robbkidd:
Updated the error to include the stacktrace. Coming from chef code.

But the `Errno::EILSEQ` is different than `ArgumentError`. I may not be triggering the same code path as the integration test.

lamont-granquist:
so i'm seeing the same errors on 2.5.0 and 2.5.1 when i throw illegal chars into `file_base` so i don't think that's a regression, but it looks like we have a bug there when taking backups -- although i seem to wind up with spec bugs after fixing it that i don't understand where they're coming from.

PR #7108, opened 2018-04-03
need to escape '.' in regex

lamont-granquist:


tas50:
Dupe of #7107

tas50:
Merged #7107 which came in 1 minute before this

PR #7122, opened 2018-04-05
add delegator for property_is_set? to providers

lamont-granquist:
This was dropped in Chef 14.0.190 by chef/chef#6952 as a side effect
of stripping out all the Chef::Resource automagic delegators.

People were using this one inside of provider code, and it seems fine
to continue to use it.

Moving it to inject it directly into the Provider class instead of
just custom resource action_classes to make our API more consistent.


lamont-granquist:
travis is broken on master, same errors.

PR #7123, opened 2018-04-05
changing to double spaces after periods

lamont-granquist:
just need to get a fresh run of master with no code changes


tas50:
This is an evil PR

lamont-granquist:
i guess i learned everything i needed to know, so you're safe...

for now...

Issue #7127, opened 2018-04-06
new macos_userdefaults resource improperly converts booleans to integers


erikng:
## Description
When passing booleans to the new `macos_userdefaults` resource and not explicitly defining a boolean type, booleans will be converted to integers.

This is because you have a `coerce_booleans` function that is ran immediately on the value returned. This means that all booleans will be immediately converted to integers. You can validate this by running `type(value)`.

I have provided a fix below. The conversion should occur for the `load_current_value` portion only and then pass the original value to the `action: write`

## Chef Version
14.0.190

## Platform Version
macOS 10.13.4

## Replication Case
Pass a simple boolean
```
macos_userdefaults 'com.apple.commerce' do
  domain '/Library/Preferences/com.apple.commerce'
  key 'AutoUpdate'
  value true
end
```

## Client Output
`- execute defaults write '/Library/Preferences/com.apple.commerce' 'AutoUpdate' -int 1`

## Notes
- I have fixed the resource [here](https://gist.github.com/erikng/79d58a66cfdd584a9ae288d34e167eca
- After combing through this resource, **I think it may be dangerous for chef to officially support this new resource** and would love to discuss at length, but would prefer if we start the conversation with this fix.

PR #7128, opened 2018-04-06
macos_userdefaults - fix improper boolean to int conversion

erikng:
### Description

This is a fix for the improper boolean to int conversion occuring on macos_userdefaults

### Issues Resolved

Fix for https://github.com/chef/chef/issues/7127


tas50:
You'll need to update the unit tests here since they're all failing now

Issue #7117, opened 2018-04-04
Chef14: windows_task frequency :none Fails to Validate 


lokiwins:
## Description
I hope I got all the information you require :) 

windows_task Option frequency fails validation when set to :none
This was working with the chef client 13, but once it started to install chef client version 14 I started to experience this issue. I am using terraform to bring up a windows server 2016 VM and everything works fine up until the chef-client starts to run, which is why I figured posting the issue here would be appropriate apologies if not. With the latest version it seems to have an issue validating the frequency option :none for a windows_task resource. 

## Chef Version
14.0.190 - Validation fails for :none frequency option
13.8.5 - Working

## Platform Version
Tested on the following: 
Windows Server 2016
Windows Server 2012R2

## Replication Case

Create windows_task with frequency set to :none and run with chef-client 14.0.190
```
winInfo = Chef::DataBagItem.load( 'win_info', 'main' )
windows_task 'Test' do
    only_if { node['wintaskTest']['DB_Stuff']['InstallDB'] == true }
    user winInfo['admin_username']
    password winInfo['admin_password']
    cwd node['wintaskTest']
    command "test.bat"
    run_level :highest
    frequency :none
    action :create
end
```
Contents of Databag

With Chef Client 14.0.190 this results in a validation error claiming frequency cannot be :none 

If I force chef client version to 13.8.5 the run is successful. I am still fairly new to Chef but I figured this should work as it is outlined in the documentation that :none is a valid setting. 
```
frequency
Ruby Type: Symbol | Default Value: :hourly

Frequency with which to run the task. Values include: :minute, :hourly, :daily, :weekly, :monthly, :none, :once, :on_logon, :onstart, :on_idle. The :once value requires the start_time property. The :none frequency requires Chef 13.6 or later.
```
## Client Output

Using latest Chef-Client Version : 14.0.190
```
module.Server2012R2-.vsphere_virtual_machine.BaseVM-Test (chef): Installing Cookbook Gems:
module.Server2012R2-.vsphere_virtual_machine.BaseVM-Test (chef): Compiling Cookbooks...
module.Server2012R2-.vsphere_virtual_machine.BaseVM-Test (chef):
module.Server2012R2-.vsphere_virtual_machine.BaseVM-Test (chef): ================================================================================
module.Server2012R2-.vsphere_virtual_machine.BaseVM-Test (chef): Recipe Compile Error in C:/chef/cache/cookbooks/wintaskTest/recipes/default.rb
[...]

PR #7134, opened 2018-04-07
Modify the provides for all resources from cookbooks so chef wins

tas50:
Due to changes in chef-client 14 old cookbooks started winning and this causes lots of problems, especially for old incompatible resources. Chef client should always win here.


tas50:
Fixes #7117

Issue #6606, opened 2017-11-27
package/apt.rb doesn't support epoch in version


Aketzu:
## Description

If a package version contains epoch (i.e. starts with `1:`) version_compare will feed it to Gem::Version.new which barfs.

## Chef Version

chef-client -v
Chef: 13.5.25

## Platform Version

Debian 9.2

## Replication Case

Following cookbook recipe and then just waiting for a new version of the package
```
    apt_repository 'zabbix' do
      distribution 'stretch'
      uri 'http://repo.zabbix.com/zabbix/3.4/debian'
      key '082AB56BA14FE591'
      keyserver 'keyserver.ubuntu.com'
      components ['main']
      notifies :run, 'execute[apt-get update]', :immediately
    end

    apt_package 'zabbix-agent' do
      options '-o Dpkg::Options::="--force-confold" -y'
      action :upgrade
    end
```

## Client Output

```
Chef run failed on server.example.com
apt_package[zabbix-agent] (server::zabbix line 75) had an error: ArgumentError: Malformed version number string 1:3.4.4-3-stretch

Backtrace
/opt/chef/embedded/lib/ruby/2.4.0/rubygems/version.rb:208:in `initialize'
/opt/chef/embedded/lib/ruby/2.4.0/rubygems/version.rb:199:in `new'
/opt/chef/embedded/lib/ruby/2.4.0/rubygems/version.rb:199:in `new'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.5.25/lib/chef/provider/package/apt.rb:134:in `version_compare'
```

Rest in https://gist.github.com/Aketzu/3e6d84fd3f7700cdf0c556bec1d51e9c

## Analysis

package/apt.rb is using Gem::Version for checking which version is never but Gem version numbers follow different format than Debian package versions. Current version numbering is defined in https://www.debian.org/doc/debian-policy/#s-f-version which also defines proper way of comparing version numbers.

https://github.com/chef/chef/blob/913ea3f7f8e3363caa71db6b4e16cbf2ed481363/lib/chef/provider/package/apt.rb#L130-L135
apt.rb version_compare should be changed to use much more complex version comparision logic

Other options are to call system `dpkg --compare-versions` or write a library which uses apt-pkg (e.g. https://anonscm.debian.org/cgit/collab-maint/ruby-debian.git/tree/ext/debian_version/Version.cpp ). Quick search didn't turn up any ruby implementations of Debian version comparision logic.

[...]

PR #7136, opened 2018-04-09
add support for lock bot

lamont-granquist:
https://github.com/dessant/lock-threads

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,1 +1,1 @@
-daysUntilLock: 30
+daysUntilLock: 60
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,1 @@
+daysUntilLock: 30


.github/lock.yml
```diff
@@ -0,0 +1 @@
+daysUntilLock: 30
```
thommay: Probably worth adding a comment too `lockComment`

lamont-granquist: I don't have any ideas for improving over the default though:  "This thread has been automatically locked because it has not had recent activity. Please open a new issue for related bugs and link to relevant comments in this thread."

thommay: oh, i missed that that was the default.
The only other comment i have is that the lock timeout should be longer (i was gonna do 60 days), because it could be almost 30 days between a fix being merged and being released.

lamont-granquist: does closing the ticket reset the timer?

lamont-granquist: (i assumed it did, IDK)

lamont-granquist: locking an issue definitely counts as an update.  e.g. https://github.com/chef/chef/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed++updated%3A%3E%3D2018-04-02 find https://github.com/chef/chef/issues/6606 because i locked that on the 2nd.

lamont-granquist: screen shot 2018-04-09 at 3 54 08 pm

yeah, the timer resets when an issue is closed -- so this is 30 days from when it was closed/fixed at a minimum.

thommay: yeah, which might be only a day or two after we do a stable release with the fix

lamont-granquist:
bumped up to 60 to get it in

PR #7139, opened 2018-04-09
We no longer build or test on OSX 10.10

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

### Description

OSX 10.10 is deprecated and has been for a while.

### Issues Resolved

### Check List

- [ ] All tests pass


thommay:
But this happened in the 14.0 timescale so it's the correct place for them.

coderanger:
@thommay Fair, but we should probably call it out as an update or somesuch then? Adding things right in the middle of published release notes seems bad :)

PR #7140, opened 2018-04-10
Add support for route metric

tomdoherty:
Signed-off-by: Tom Doherty <tom.doherty@fixnetix.com>

### Description

This allows specifying route metric. metric are already supported by the ifconfig resource so add here also

### Issues Resolved

N/A

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/route.rb
+++ b/lib/chef/resource/route.rb
@@ -29,7 +29,7 @@ class Chef
 
       property :target, String, identity: true, name_property: true
       property :comment, [String, nil]
-      property :metric, [String, nil]
+      property :metric, [Integer, nil]
       property :netmask, [String, nil]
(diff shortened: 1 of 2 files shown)


lib/chef/resource/route.rb
```diff
@@ -40,7 +41,6 @@ class Route < Chef::Resource
       property :hostname, [String, nil], desired_state: false
       property :domainname, [String, nil], desired_state: false
       property :domain, [String, nil], desired_state: false
-      property :metric, [Integer, nil], desired_state: false
```
thommay: This looks wrong. metric is clearly an Integer.

tomdoherty: fixed :-)

PR #7144, opened 2018-04-11
Fix up knife logging

thommay:
ensure that trace level is selectable if necessary via -VVV
Pass the correct object to net/ssh


lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -178,7 +178,11 @@ def start_chef
           # If the user doesn't have a client path configure, let bash use the PATH for what it was designed for
           client_path = @chef_config[:chef_client_path] || "chef-client"
           s = "#{client_path} -j /etc/chef/first-boot.json"
-          s << " -l debug" if @config[:verbosity] && @config[:verbosity] >= 2
+          if @config[:verbosity] && @config[:verbosity] >= 3
+            s << " -l trace"
+          elsif @config[:verbosity] && @config[:verbosity] >= 2
```
tas50: >= 2? Wouldn't that match above

PR #7152, opened 2018-04-13
add the resources() dsl method back to providers

lamont-granquist:
Custom resource actions were picking this up via delegation to
the wrapping resource, and that wiring was removed in 14, so that
API then got dropped accidentally.  This should fix that back
up.  It also consistently injects the resource APIs into resources
and providers both now (and consistently across core resources
and custom resources).


lamont-granquist:
this probably needs a test

Issue #2994, opened 2015-02-28
chefignore in at cookbook root not honored by `knife cookbook upload`


pburkholder:
This seems to be a change from 11.X.  Reported by @mcalb. To reproduce [see comment below, original comment deleted by @pburkholder since it was pretty unhelpful]

pburkholder:
My apologies to the triage team for my lame report on a Friday evening.  Here's the README.md from https://github.com/pburkholder/chefignore-demo, which demonstrates this issue:

This cookbook demonstrates that chef-client 12.0.3 does not honor the contents
of `chefignore`, which is a security bug since content the creator considers
secret is uploaded to the chef-server in contravention to what one expects:


The `secrets` file in this cookbook has one secret:

    MySecretPassword

and the `chefignore` file has one entry:

    secrets

In chef-client 11.X, the `knife cookbook upload` command would not upload the
secrets file per the contents of chefignore, however, observe what happens

```
> knife --version
Chef: 12.0.3
> knife cookbook list | grep chefignore
> pwd
/Users/pburkholder/Hack/chefignore-demo
> knife cookbook upload -o ~/Hack chefignore-demo
Uploading chefignore-demo [0.1.0]
Uploaded 1 cookbook.
> cd ~/tmp
> knife cookbook download chefignore-demo
Downloading chefignore-demo cookbook version 0.1.0
Downloading resources
Downloading providers
Downloading recipes
Downloading definitions
Downloading libraries
Downloading attributes
Downloading files
Downloading templates
Downloading root_files
Cookbook downloaded to /Users/pburkholder/tmp/chefignore-demo-0.1.0
> more chefignore-demo-0.1.0/secrets
MySecretPassword
> more chefignore-demo-0.1.0/chefignore
secrets
>
```

Oops, so we've uploaded the secrets, and downloaded them, despite the
instructions of the `chefignore` file.

pburkholder:
In case it's not clear above, here's the cookbook structure:

```
> tree .
.
├── README.md
├── chefignore
├── metadata.rb
├── recipes
│   └── default.rb
└── secrets

1 directory, 5 files
> more chefignore
secrets
```

Note that if you do, `cp chefignore ..` so the chefignore file is at the top of the cookbook_path, then the contents of chefignore are honored (a symlink also works).

pburkholder:
[...]

PR #7156, opened 2018-04-13
Added searching for chefignore file in cookbook root directory

LukeRoz:
Signed-off-by: Lukasz Rozok <lukescroz@gmail.com>

### Description

Fixes reading chefignore file from cookbook root directory.

### Issues Resolved

https://github.com/chef/chef/issues/2994

### Check List

- [ ] All tests pass


lamont-granquist:
For reference this is what berkshelf has always done:

https://github.com/berkshelf/berkshelf/blob/master/lib/berkshelf/berksfile.rb#L603-L620

thommay:
Hey @LukeRoz this looks great; could you fix up chefstyle (just run `chefstyle -a`) and rebase and we'll get this merged.
A potential optimization might be to use `Pathname.ascend`, but that would be a nice to have rather than a blocker.

LukeRoz:
Hi @thommay,

Sorry for late reply. One of offenses in my code is related to using "for" instead of "each" ([chefignore.rb#L64](https://github.com/LukeRoz/chef/blob/22ed661e9099343db5cee4a60a88be527a637d21/lib/chef/cookbook/chefignore.rb#L64)). To be honest "each" doesn't fit in here for me. Any ideas how could I change it?

lib/chef/cookbook/chefignore.rb
```diff
@@ -56,22 +54,32 @@ def parse_ignore_file
...
         ignore_globs
       end
 
       def find_ignore_file(path)
-        if File.basename(path) =~ /chefignore/
-          path
-        else
-          File.join(path, "chefignore")
+        ignore_path=path
+        for i in 0..2
```
coderanger: `(0..2).each do |i|` or `3.times do |i|`

lib/chef/cookbook/chefignore.rb
```diff
@@ -56,22 +54,32 @@ def parse_ignore_file
             ignore_globs << line.strip unless line =~ COMMENTS_AND_WHITESPACE
           end
         else
-          Chef::Log.trace("No chefignore file found at #{@ignore_file} no files will be ignored")
+          Chef::Log.debug("No chefignore file found. No files will be ignored!")
```
coderanger: This needs to be updated.

coderanger:
Do we not have helpers for this kind of thing already? This looks like some very intricate code that we probably shouldn't need to be so intricate :)

lib/chef/cookbook/chefignore.rb
thommay: I think this is actually correct - a `debug` level message seems reasonable.

lamont-granquist:
@coderanger no, we don't have helpers, and berkshelf does its own different thing which is why I couldn't reuse code there.

Hence my suggestion to standardize on what berkshelf does since it seems it has been doing that for a long time.

lamont-granquist:
OTOH, this is better than what we've got now though....

lamont-granquist:
looks like there may still be a chefstyle failure tho...

LukeRoz:
Son of a... OK, double-quoted strings are required. I'll update it tomorrow morning :)

tas50:
Closing this in favor of #8925 which solves this problem plus several others

PR #7161, opened 2018-04-16
Add ssh_known_hosts_entry resource from ssh_known_hosts cookbook

tas50:
Port this from the 6.1 release of the cookbook


lib/chef/resource/ssh_known_hosts_entry.rb
```diff
@@ -0,0 +1,147 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class SshKnownHostsEntry < Chef::Resource
+      resource_name :ssh_known_hosts_entry
+      provides(:ssh_known_hosts_entry) { true }
```
coderanger: Need to remove the `{ true }` trick.

coderanger: Or better yet, just remove this whole line.

lib/chef/resource/ssh_known_hosts_entry.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      description "Use the ssh_known_hosts_entry resource to append an entry for the specified host in /etc/ssh/ssh_known_hosts or a user's known hosts file if specified."
+      introduced "14.3"
+      preview_resource true
```
coderanger: This needs to be above the `resource_name`.

PR #7165, opened 2018-04-18
Add kernel_module resource from the kernel_module cookbook

tas50:
Copied from the 1.1.1 release.


lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
...
+          content "#{new_resource.modname}\n"
+          notifies :run, "execute[update initramfs]"
+        end
+
+        declare_resource(:execute, "update initramfs") do
+          command initramfs_command
+          action :nothing
+        end
+
+        new_resource.run_action(:load)
```
coderanger: I'm not sure this does what you want. This would run the entire load action before the install action's resources run.

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
...
+      provides(:kernel_module) { true }
+
+      property :modname, String, name_property: true, identity: true
+      property :load_dir, String, default: "/etc/modules-load.d"
+      property :unload_dir, String, default: "/etc/modprobe.d"
+
+      # Load kernel module, and ensure it loads on reboot
+      action :install do
+        declare_resource(:directory, new_resource.load_dir) do
+          recursive true
```
coderanger: This could create non-leaf dirs without safe permissions, probably should be removed.

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
...
+          action :nothing
+        end
+
+        new_resource.run_action(:unload)
+      end
+
+      # Load kernel module
+      action :load do
+        declare_resource(:execute, "modprobe #{new_resource.modname}") do
+          not_if "cat /proc/modules | grep ^#{new_resource.modname}"
```
coderanger: Can do this in Ruby code for better perf/testing and less weirdness with special characters (this would do funky stuff if the modname included either regex metachar or shell metachars).

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
...
+      # Unload kernel module
+      action :unload do
+        declare_resource(:execute, "modprobe -r #{new_resource.modname}") do
+          only_if "cat /proc/modules | grep ^#{new_resource.modname}"
+        end
+      end
+
+      action_class do
+        def initramfs_command
+          if platform_family?("debian")
```
coderanger: This doesn't seem exhaustive, should probably fail on unknown families.

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
+#
+# Resource:: kernel_module
+#
+# The MIT License (MIT)
+#
+# Copyright 2016-2018, Shopify Inc.
+# Copyright 22018, Chef Software, Inc.
```
coderanger: From the fuuuuuutre :)

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,100 @@
+#
+# Resource:: kernel_module
+#
+# The MIT License (MIT)
```
coderanger: This represents a new license, will need to make sure we are compliant with all the requirements. Is the new code going to also be under MIT or Apache like the rest of Chef? Does this impact our patent safety stance? (we may need to get a parent grant from Shopify).

lib/chef/resource/kernel_module.rb
tas50: That's the goal.

lib/chef/resource/kernel_module.rb
```diff
@@ -0,0 +1,128 @@
...
+      property :unload_dir, String,
+               description: "The modprobe.d directory.",
+               default: "/etc/modprobe.d"
+
+      action :install do
+        description "Load kernel module, and ensure it loads on reboot"
+
+        declare_resource(:directory, new_resource.load_dir) do
+          recursive true
+        end
```
lamont-granquist: its custom-resource style so i don't think you need to do `declare_resource`?

coderanger: Don't need to, but doesn't hurt anything so authors style choice.

tas50: Fixed

lib/chef/resource/kernel_module.rb
tas50: Lawyers have signed off on the license in this specific case

PR #7167, opened 2018-04-19
actually test against cheffs

lamont-granquist:
- revert the changes to try to lookup chef-zero cookbook using name property.  even after the perf fixes, it caused chef-zero to fail lots of pedant tests, which were masked by the fact that this repo was testing the released version of chef, and pulling master of chef-zero which was still pinned to chef 13.

- test chef-zero and cheffish correctly against PRs and against master.

this reverts the changes in #7143 and #6471

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,9 +88,10 @@ matrix:
     script: bundle exec tasks/bin/run_external_test $TEST_GEM master rake
     rvm: 2.5.1
   - env:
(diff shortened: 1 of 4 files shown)


lamont-granquist:
did not get enough sleep, lets try this again tomorrow...

lamont-granquist:
breaks cheffish now because cheffish is actually broken.

Issue #7119, opened 2018-04-05
Chef-14: service :enable does not work on ubuntu 14.04


ccrebolder:
## Description
When I use the `:enable` action to a `service` resource in Ubuntu 14.04 the service is not actually enabled. The integration test for `should be_enabled` fails and I have confirmed the service is not running after rebooting the machine.

## Chef Version
Chef: 14.0.190

## Platform Version
I am using `bento/ubuntu-14.04     (virtualbox, 2.2.9)` in Test Kitchen.

## Replication Case
I discovered this with a private cookbook, but I can replicate using the master branch of the chef-client cookbook, which contains a test to verify the service is enabled.
```
$ kitchen verify service-upstart-ubuntu-1404                                                                                                                                                                                 
-----> Starting Kitchen (v1.17.0)
-----> Setting up <service-upstart-ubuntu-1404>...
       Finished setting up <service-upstart-ubuntu-1404> (0m0.00s).
-----> Verifying <service-upstart-ubuntu-1404>...
       Loaded tests from {:path=>"/home/creboldc/cookbooks/chef-client/test/integration/service_upstart"} 

Profile: tests from {:path=>"/home/creboldc/cookbooks/chef-client/test/integration/service_upstart"}
Version: (not specified)
Target:  ssh://vagrant@127.0.0.1:2222


  Processes chef-client
     ✔  should exist
  Service chef-client
     ∅  should be enabled
     expected that `Service chef-client` is enabled
     ✔  should be installed
     ✔  should be running

Test Summary: 3 successful, 1 failure, 0 skipped
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Verify failed on instance <service-upstart-ubuntu-1404>.  Please see .kitchen/logs/service-upstart-ubuntu-1404.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

## Client Output
The chef-client run tries to enable the service every run.

Let me know if more logs or stacktrace is needed.

trappar:
Seeing this as well going. :enable worked as of chef-client 13.8.5 but is failing using 14.0.190 for me, also on Ubuntu 14.04.

PR #7170, opened 2018-04-20
[Debian/Ubuntu] Detect init script service levels

thommay:
Debian removed the no-op flag from update-rc.d, and so we're now
unlinking services rather than just querying them. Just parsing the
scripts ourselves is much safer.

Fixes: #7119

### DIFF
--- a/lib/chef/provider/service/debian.rb
+++ b/lib/chef/provider/service/debian.rb
@@ -52,8 +52,8 @@ class Chef
           end
 
           requirements.assert(:all_actions) do |a|
-            a.assertion { @so_priority.exitstatus == 0 }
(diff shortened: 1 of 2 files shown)


lib/chef/provider/service/debian.rb
```diff
@@ -70,22 +70,40 @@ def define_resource_requirements
...
+        def parse_init_file(path)
+          return [] unless ::File.exist?(path)
+          start = false
+          finish = false
+          ::File.readlines(path).each_with_object([]) do |line, acc|
+            if line =~ /^### BEGIN INIT INFO/
+              start = true
+              next
+            elsif line =~ /^### END INIT INFO/
+              finish = true
```
lamont-granquist: you can probably just set start to false here and eliminate the finish boolean.  the 'next's are likely redundant as well

PR #7192, opened 2018-04-26
sysctl: Fix ignore_error being ignored

tas50:
This came into the sysctl cookbook https://github.com/sous-chefs/sysctl/pull/127


tas50:
The raise was actually bogus there since we already passed -e. I removed that line entirely

tas50:
Test failure is unrelated and we have a fix in the works in knife-windows

PR #7194, opened 2018-04-27
Allow specifying `ignore_failure :quiet` to disable the error spew

coderanger:
Directly inspired by https://github.com/sous-chefs/sysctl/pull/127 but more generally people have been asking for this forever so why not?


jasonwbarnett:
@coderanger thank you! I literally **just** ran into an odd situation where I needed a ton of resources to have ignore_failure and many chef runs were noisier than my brain could handle.

Issue #6729, opened 2018-01-04
The gem_package provider should set clear_sources to true if either `Chef::Config[:rubygems_url]` is set or if a `source` is given on the resource


tstuber:
## Description
Given the following setup: 
* Air-gapped environment
* Local ruby gem repository (linked to rubygems.org)
* `client.rb` has a defined `rubygems_url` to the lokal rubygems repository
 
This works so far. But whenever we install a ruby gem via `chef_gem` resource, it always takes 3 minutes until completion. The only way it works immediatly is, if we use the `clear_sources` flag to set it to true.

Our investigations showed that whenever an installation source is specified, the default `rubygems url` (rubygems.org) is used. It would be possible to set the `clear_sources` flag to clean the `rubygems_url` list and to avoid this problem, but that can not be made for shared cookbooks (such as push_jobs, etc). 

## Chef Version
Client: 13.6.4

## Platform Version
Linux, Red Hat 7.4

## Replication Case

* Ensure an air-gapped client
* Define a custom `rubygems_url` (via `client.rb`)
* Run the following snipped that installs a not existing ruby gems

```
chef_gem 'chef-ruby-lvm' do
  action [ :remove, :install ]
  # clear_source true # uncomment to fix the 3min timeout. 
end
```

The same behaviour can be faced by using `gem install` directly.

**3 Minutes installation duration:**
```
gem install --source http://rubygems.mycustomer.local/repository/rubygems mixlib-install --verbose
...
unable to pull data from https://rubygems.org/': timed out (https://api.rubygems.org/specs.4.8.gz)
1 gem installed
```

**Instant installation duration:** 
```
gem install --clear-sources --source http://rubygems.mycustomer.local/repository/rubygems mixlib-install --verbose
```

## Client Output

The behaviour without the `clear_sources` (with 3 minutes duration) is as follows:

```
[2017-12-14T12:19:26+01:00] DEBUG: chef_gem[chef-ruby-lvm] no installed version found for chef-ruby-lvm (>= 0)
[2017-12-14T12:19:26+01:00] DEBUG: found gem chef-ruby-lvm version 0.3.0 for platform ruby from 
[2017-12-14T12:19:26+01:00] DEBUG: chef_gem[chef-ruby-lvm] chef-ruby-lvm not installed, installing 0.3.0
[2017-12-14T12:22:26+01:00] INFO: chef_gem[chef-ruby-lvm] installed chef-ruby-lvm at 0.3.0
[...]

PR #7196, opened 2018-04-27
Tg/clear gem sources option

nrgetik:
### Description

Added a config option called clear_gem_sources and defaulted gem_package.clear_gem_sources to it. This supports airgapped environments that want to exclude implicit remote gem sources.


itmustbejj:
I believe this address https://github.com/chef/chef/issues/6729

lamont-granquist:
looks like you have a chefstyle issue

coderanger:
A workaround for folks on older Chef that need this feature:

```ruby
class Chef::Resource::GemPackage
  def clear_gem_sources
    true
  end
end
```

Drop that or something like it in a `libraries/` file.

PR #7204, opened 2018-04-27
Convert some of remote_directory to use properties

tas50:
This isn't all of it, but it's a start


rmoriz:
Danger! This PR changed the default for `overwrite` from `true` to `false`

Issue #7185, opened 2018-04-25
no_lazy_load false network connectivity performance regression in chef 13


pudquick:
## Description
Using no_lazy_load false for cookbook_file/remote_directory resources on Chef 13 seems to cause a considerable number of additional TCP connections and bandwidth usage compared to Chef 12.

## Chef Version
```
chef-client --version
Chef: 13.8.5
```

## Platform Version
macOS 10.13.4

## Replication Case
- Edit /etc/chef/client.rb and set no_lazy_load false
- Run a cookbook thats uses a remote_directory resource. For instance:
```
# Cookbook: default.rb
rd = remote_directory '/tmp/rd_test' do
  source 'some/folder'
  owner 'root'
  group 0
  mode 0755
  purge true
  recursive true
  action :nothing
end

rd.run_action(:create)
```
- This remote_directory resource should have 50 or more files located inside it to begin to see the impact.
- Perform a packet capture during a chef-client run, then compare activity between `false` and `true` settings.

pudquick:
When we originally opened issue #6051, no_lazy_load false was not supported for remote_directory in chef 13 at all. It was very helpful to see it implemented, thank you!

However, now that we are doing client rollout with it, we've noticed a considerably different network connectivity load from chef 13 clients.

Wireshark captures, for an example run, showed 17 total TCP connections opened during the chef 13 client run with no_lazy_load `true` - and over 600 TCP connections when set to `false`.

This has resulted in considerable additional load on the chef servers themselves when attempting to use no_lazy_load false in the client.rb.

Having this setting as `false` in chef 12 did not appear to have the same performance impact. I'm working on collecting traces for it now for direct comparison with the same cookbook set.

Is there an implementation reason / difference why we would see such a performance hit for chef 13?

pudquick:
Having performed the same analysis with chef-client 12.21.31:

- `no_lazy_load true` with the same cookbook comparison used for chef 13: identical number of TCP connections as chef 13 (under 20). And similar total byte transfer / raw bandwidth.
- `no_lazy_load true` on chef 12.21.31: Only **7** TCP connections (less, not more), and a total raw byte transfer with the chef server of 10% (again less, not more) compared to the `true` run.

So in summary:
- On chef-client 12, using `no_lazy_load false` with our remote_directory configuration represented a 90% bandwidth saving and 50%+ TCP connection saving vs chef-client `no_lazy_load true` runs.
[...]

PR #7208, opened 2018-04-30
Don't always request lazy files

thommay:
Testing with a single remote directory of 100 3 byte files, this drops a chef run from 40 secs to 5 secs when the files are already in the cache (identical to lazy load being disabled).


thommay:
This is part of #7185 .

jaymzh:
Nice catch!

Can we add a test for this though?

pudquick:
@thommay This looks like a great change, thank you.

Since 13 is still supported and we see this regression in 13, will an updated 13 release also incorporate this fix?

pudquick:
Whoops, thought I was on the original issue #7185 page for the discussion. My fault.

PR #7224, opened 2018-05-05
Implement rfc107: NodeMap locking for resource and provider handlers

coderanger:
Implementation of https://github.com/chef/chef-rfc/pull/305 based on current discussion. May need more changes if that RFC is changed before acceptance.

Probably needs some tweaks to the language of the warning, but this is a start and I think implements the right behavior.

### DIFF
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -76,7 +76,14 @@ class Chef
       # 4. No previous `provides` had `allow_cookbook_override`, either set to
       #    true or with a string version matcher that still matches Chef::VERSION
       if !__core_override__ && map[key] && map[key].any? {|matcher| matcher[:locked] } && !map[key].any? {|matcher| matcher[:cookbook_override].is_a?(String) ? Chef::VERSION =~ matcher[:cookbook_override] : matcher[:cookbook_override] }
-        type_of_thing = klass.is_a?(Chef::Resource) ? 'resource' : 'provider'
+        # If we ever use locked mode on things other than the resource and provider handler maps, this probably needs a tweak.
+        type_of_thing = if klass < Chef::Resource
+          'resource'
+        elsif klass < Chef::Provider
+          'provider'
+        else
+          klass.superclass.to_s
+        end
         # For now, only log the warning.
         Chef.log_deprecation("Trying to register #{type_of_thing} #{key} on top of existing Chef core #{type_of_thing}. Check if a new version of the cookbook is available.")
         # In 15.0, uncomment this and remove the log above.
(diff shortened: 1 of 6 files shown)


lib/chef/node_map.rb
```diff
@@ -59,6 +65,31 @@ def set(key, klass, platform: nil, platform_version: nil, platform_family: nil,
       new_matcher[:block] = block if block
       new_matcher[:canonical] = canonical if canonical
       new_matcher[:override] = override if override
+      new_matcher[:cookbook_override] = allow_cookbook_override
+      new_matcher[:core_override] = __core_override__
+
+      # Check if the key is already present and locked, unless the override is allowed.
+      # The checks to see if we should reject, in order:
+      # 1. Core overide mode is not set.
```
jaymzh: override

lib/chef/node_map.rb
```diff
@@ -59,6 +65,31 @@ def set(key, klass, platform: nil, platform_version: nil, platform_family: nil,
...
+      # 2. The key exists.
+      # 3. At least one previous `provides` is now locked.
+      # 4. No previous `provides` had `allow_cookbook_override`, either set to
```
jaymzh: I'm not sure I follow this logic.

Lets be concrete:

* Package::YumPackage (builtin) is locked, and has `allow_cookbook_override
* Package::HonkyDory (builtin) is locked, and does NOT have `allow_cookbook_override`
* We now hit Package::WillyNilly - should it be able  override HonkDory even though HonkyDory has not allowed overrides?

Since only internal stuff can be `locked` (I think?), shouldn't we ensure that all things which are locked also allow overrides? Or am I misunderstanding?

coderanger: The locking happens all at once after all of core's resource and provider files are loaded, so anything in core can always do whatever it wants. Only after https://github.com/chef/chef/pull/7224/files#diff-c34d08382bb25f61e44ecbd5f50f4166R278 do we transition to the new rules. So if `Package::WillyNilly` was in a cookbook, then it would be rejected starting in 15.0 unless it uses `__core_override__`.

coderanger: Oh I think I see what you mean, the rules around `allow_cookbook_override` make less sense for providers (which I think is what you mean here?), setting that on just one of the core package providers would indeed get funky.

coderanger:
Fingers crossed all the tests pass, but I think this is ready to go. The low-level `allow_cookbook_override` and `__core_override__` flags for `provides` are still there but the normal workflow should be the new `preview_resource true` flag added to the resource when it goes into Chef core. Opinions?

PR #7226, opened 2018-05-05
Backport quiet feature

jasonwbarnett:
Backporting #7194 into Chef 13.

Literally just cherry-picked @coderanger commit from master

### DIFF
--- a/lib/chef/formatters/base.rb
+++ b/lib/chef/formatters/base.rb
@@ -137,7 +137,7 @@ class Chef
 
       def resource_failed(resource, action, exception)
         description = ErrorMapper.resource_failed(resource, action, exception)
-        display_error(description)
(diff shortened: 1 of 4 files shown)


coderanger:
Unlike the other core devs, I don't get paid to work on Chef, so I also am not signing up to support this. You will have to be on the hook for this permanently (or at least until Chef 13 goes EOL).

jasonwbarnett:
I can commit to supporting this feature until April of 2019, but I don't feel strongly about this getting merged in because I realized I can easily monkey patch this functionality in a cookbook library.

thommay:
Thanks for the PR;  we're only updating 13 for serious issues and security fixes, so I'm going to say that this isn't suitable for merge (that doesn't reflect on the value of the feature or on your request, we're just trying to be really clear about how our release cycle works).

jasonwbarnett:
@thommay completely makes sense. Thanks for getting back and explaining the why.

PR #7234, opened 2018-05-08
Switch Node#role? to use the attributes expansion instead of the run list

coderanger:
This means it understands nested roles, avoiding a common footgun. It's technically a behavior change, so perhaps it needs to wait for 15. But I would say that it errs towards what most users expect so maybe we should do it now. Discuss :)


lamont-granquist:
Yeah this is how people expect it to work and the other behavior is particularly useless with role cookbooks.

PR #7236, opened 2018-05-09
Update omnibus to pack inspec-core.

jquick:
Signed-off-by: Jared Quick <jquick@chef.io>

### Description

This change ships ChefClient with InSpec core instead of the full InSpec package. InSpec core only includes the local transport and has much less dependencies. This change will allow InSpec to be built on more systems then the full InSpec package. 

The reasoning behind this change is that we are adding dependencies to core InSpec which do not support some of the systems ChefClient supports. 

While this change is removing some functionality to the ChefClient InSpec it is leaving all the main features you would need with InSpec via ChefClient. Mainly the audit cookbook will not be affected as it only requires the local transport to run. It is being upgraded here https://github.com/chef-cookbooks/audit/pull/318 to support the new `inspec-core` gem.

What this means is everything should work as expected for the audit cookbook. If you want the newest InSpec you will need to update your audit cookbook accordingly.


jquick:
builds http://manhattan.cd.chef.co/job/chef-build/50/

jquick:
Please note that `inspec`(full version) will stop being compatible with ChefClient on May17th release.

tas50:
@jquick I see the docker-api and httpclient are no longer part of inspec-core. Does that mean the http and docker resources aren't going to work with chef once we merge this?

jquick:
Hey @tas50, the http resources just uses `faraday` which is still being included. https://github.com/chef/inspec/blob/master/inspec-core.gemspec#L37

The docker resource will still work as that is just looking at the docker setup on your system which is just running local commands: https://github.com/chef/inspec/blob/master/lib/resources/docker.rb#L156

The actual docker-api is used in Train to connect to docker instances to run inspec on. So with the ChefClient version of inspec you wont be able to do:

`inspec exec my_profile -t docker://container_id`

But you will be able to run anything on the specific system using the local transport which is mostly what the ChefClient version is used for (audit cookbook, etc).

Issue #7168, opened 2018-04-20
subcommand-specific settings support in ~/.chef/credentials


scopenco:
Thank you for implementation https://github.com/chef/chef/pull/6660 but i found a little bug in it.
If I try to set subcommand-specific settings like
```
knife[:ssh_user] = 'user1'
```
It returns me the error:
```
ERROR: CONFIGURATION ERROR:Unable to parse Credentials file: /Users/user1/.chef/credentials

parse error on value "[" ("[")
```
The workaround will be to use `case` statement in ~/.chef/knife.rb like
```
case File.basename(client_key)
when 'mykey.pm'
  knife[:ssh_user] = 'user1'
end
``` 

Anyway, I think this should be fixed in credentials parser.

```
$ chef -v
Chef Development Kit Version: 2.5.3
```

coderanger:
This is now supported via either hash keys or (preferably) subtables in the TOML:

```toml
[default]
# ...
[default.knife
ssh_user = "user1"
```

PR #7237, opened 2018-05-09
Support for toml knife config settings

smcavallo:
Support for toml knife config settings

### Description
Enable defining additional (knife) config settings via chef .credentials file.
The goal is that no conditional custom ruby code needs to exist in the client.rb or knife.rb files to support multiple chef organizations.

The new .credentials file has been fantastic and a huge improvement for maintaining one config file which can support multiple chef orgs.  See - https://chef.github.io/chef-rfc/rfc099-authentication-config-file.html
This new feature is limited in that it only currently supports setting the following config values:
node_name, chef_server_url, validation_client_name, validation_key, validator_key, client_key

There are other settings, namely the encrypted data bag secret file, which aren't yet supported.
This adds support for setting any knife configs via the .credentials file.
Inline table format can be used.

### Issues Resolved

https://github.com/chef/chef/issues/7168

### Check List

- [ X] New functionality includes tests
- [ X] All tests pass

### DIFF
--- a/chef-config/lib/chef-config/workstation_config_loader.rb
+++ b/chef-config/lib/chef-config/workstation_config_loader.rb
@@ -156,10 +156,10 @@ module ChefConfig
       Config.chef_server_url = creds.fetch("chef_server_url") if creds.key?("chef_server_url")
       Config.validation_client_name = creds.fetch("validation_client_name") if creds.key?("validation_client_name")
 
-      creds.each  do |key, value|
+      creds.each do |key, value|
         next unless key == "knife"
-        value.each do | k, v |
-          Config.knife[k.to_sym] = v
+        value.each do |k, v|
+          Config.knife.merge!({k.to_sym => v})
         end
       end
 
--- a/chef-config/lib/chef-config/workstation_config_loader.rb
+++ b/chef-config/lib/chef-config/workstation_config_loader.rb
@@ -156,10 +156,10 @@ module ChefConfig
       Config.chef_server_url = creds.fetch("chef_server_url") if creds.key?("chef_server_url")
       Config.validation_client_name = creds.fetch("validation_client_name") if creds.key?("validation_client_name")
 
-      creds.each do |cred|
-        knife_setting = /knife\[:(\w+)\]/.match(cred[0])
-        if knife_setting
-          Config.knife[knife_setting[1].to_sym] = cred[1]
+      creds.each  do |key, value|
+        next if key != "knife"
+        value.each do | k, v |
+          Config.knife[k.to_sym] = v
         end
       end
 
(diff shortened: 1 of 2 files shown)


thommay:
Hm, I think we can use inline tables, just not the dot_separated keys:

```
[default]
client_name = "foo"
[default.knife]
ssh_user = "thom"
```
or
```
[default]
client_name = "foo"
knife = {
  ssh_user = "thom"
}
```
both result in:
```
{"default"=>{"client_name"=>"foo", "knife"=>{"ssh_user"=>"thom"}}}
```

Which seems like what we need.

thommay:
That not withstanding, this is an awesome PR, thank you!

smcavallo:
@thommay - thanks for the feedback!  I've incorporated your changes and added tests for using both those formats.

chef-config/lib/chef-config/workstation_config_loader.rb
```diff
@@ -156,6 +156,13 @@ def apply_credentials(creds, profile)
       Config.chef_server_url = creds.fetch("chef_server_url") if creds.key?("chef_server_url")
       Config.validation_client_name = creds.fetch("validation_client_name") if creds.key?("validation_client_name")
 
+      creds.each  do |key, value|
```
coderanger: Isn't this the same as `Config.knife.update(creds['knife'] || {})`? Give or take dealing with symbolizing keys.

coderanger: (and I'm like 99.99% certain we don't need the `to_sym` because mixlib-config coerces keys)

smcavallo: @coderanger - can you point me in the right direction on how to use Config.knife.update?
I've tried but haven't been able to get any sample data set hashes to actually be added to the config when trying to use it.

coderanger: I was close, looks like it should be `merge!` not `update`. So to rewind a bit, `Config.knife` returns a Mixlib::Config-ized object itself. So we can use [`merge!`](https://github.com/chef/mixlib-config/blob/master/lib/mixlib/config.rb#L291) on it. And `creds` is a hash I think, so we can grab the `"knife"` directly rather than iterating. And if there is no `knife` key, merging in `{}` is a no-op. So putting it all together, something like this:

```
Config.knife.merge!(creds['knife'] || {})
```

Could also use `if creds['knife']` rather than the `||{}` trick if you prefer, up to your personal style :)

smcavallo: @coderanger - what you said makes sense.  I wasn't able to get it to work in that way.  It wouldn't work unless the keys were symbolized.
I've refactored in a way which still uses the knife.merge but symbolizes keys and handles if creds['knife'] is nil

thommay:
Thanks, this is great!

PR #7239, opened 2018-05-09
Convert reboot resource to a custom resource with descriptions

tas50:
This is a simple resource. It's way easier to understand in a single file.


tas50:
@lamont-granquist am I missing something here or is default_action busted when you specify :nothing?

PR #7240, opened 2018-05-09
Remove unused route resource properties

tas50:
Add descriptions and remove unused properties.

### DIFF
--- a/lib/chef/provider/route.rb
+++ b/lib/chef/provider/route.rb
@@ -158,9 +158,9 @@ class Chef
       end
 
       def generate_config
-        conf = {}
         case node[:platform_family]
         when "rhel", "amazon", "fedora"
+          conf = {}
           # walk the collection
           run_context.resource_collection.each do |resource|
             next unless resource.is_a? Chef::Resource::Route
--- a/lib/chef/resource/route.rb
+++ b/lib/chef/resource/route.rb
@@ -48,7 +48,7 @@ class Chef
                description: "The network interface to which the route applies.",
                desired_state: false # Has a partial default in the provider of eth0.
 
-      property :route_type, Symbol,
+      property :route_type, [Symbol, String],
                description: "",
                equal_to: [:host, :net], default: :host, desired_state: false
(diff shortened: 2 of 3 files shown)


lib/chef/provider/route.rb
```diff
@@ -154,58 +154,57 @@ def action_delete
         end
 
         # for now we always write the file (ugly but its what it is)
-        generate_config
+        write_config
       end
 
-      def generate_config
+      def write_config
```
lamont-granquist: this could be "public API" if anyone subclasses.

lib/chef/provider/route.rb
```diff
@@ -154,58 +154,57 @@ def action_delete
...
+        return unless platform_family?("rhel", "fedora", "amazon")
+
         conf = {}
-        case node[:platform]
-        when "centos", "redhat", "fedora"
```
lamont-granquist: kind of like the case statement here better than the early return, its more indicative that this is where you'd need to add support for debian and other linuxes...

lib/chef/resource/route.rb
```diff
@@ -27,20 +27,26 @@ class Route < Chef::Resource
...
-      property :gateway, [String, nil]
-      property :device, [String, nil], desired_state: false # Has a partial default in the provider of eth0.
-      property :route_type, [:host, :net], default: :host, coerce: proc { |x| x.to_sym }, desired_state: false
-
-      # I can find no evidence of these properties actually being used by Chef. NK 2017-04-11
-      property :networking, [String, nil], desired_state: false
-      property :networking_ipv6, [String, nil], desired_state: false
-      property :hostname, [String, nil], desired_state: false
-      property :domainname, [String, nil], desired_state: false
-      property :domain, [String, nil], desired_state: false
```
thommay: These are all part of the `route` public API and can't be removed until chef 15

tas50: That's totally fine. This can wait till we branch off chef-14.

PR #7253, opened 2018-05-11
Add cron_d and cron_access resources

tas50:
cron_access (previously cron_manage) is a simple, but useful resource for accumulating access to cron.

cron_d when I surveyed users was a top 5 request to include in Chef. It seems to get very widely used by Linux users that like the .d pattern vs. the monolithic pattern of our existing cron resource. If people want to manage the whole file or individual snippets we'll give them the option, just as we've always done by having the core resource and this resource. Choices are a good thing here.


lamont-granquist:
appveyor hates you for some reason

lamont-granquist:
err no that's travis

lamont-granquist:
chefstyle hates you for some reason....

PR #7254, opened 2018-05-12
remote_directory: restore overwrite default

rmoriz:
### Description

`remote_directory` used to overwrite files by default. When converting to properties, the default was changed to false in PR #7204 probably unintentionally. This PR restores the default and also provides a test case.

### Check List

- [X] All tests pass


thommay:
Thanks so much @rmoriz

Issue #7212, opened 2018-05-02
Add skip_publisher_check to powershell_package resource


kitforbes:
This resource installs a PowerShell module through the PowerShell Package Manager. There are a few modules within the PSGallery that require the `-SkipPublisherCheck` switch to be added to the `Install-Module` command. I have noticed that the Chef run will not fail when PowerShell raises the exception in this case and will make it appear as though the module was successfully installed.

Some more details on this can be found in this [PowerShell Team blog post](https://blogs.msdn.microsoft.com/powershell/2016/09/29/powershellget-and-packagemanagement-in-powershell-gallery-and-github/) from 2016.

tas50:
@Happycoil Fixed this up and it's been merged to master.

PR #7259, opened 2018-05-15
Add skip_publisher_check property to powershell_package

Happycoil:
### Description

Adds a `skip_publisher_check` property and fixes a bug where catalog signing errors would cause the install to fail silently.

### Issues Resolved

https://github.com/chef/chef/issues/7212

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/powershell.rb
+++ b/lib/chef/provider/package/powershell.rb
@@ -110,12 +110,14 @@ class Chef
 
         def build_powershell_package_command(command, version = nil)
           command = [command] unless command.is_a?(Array)
+          cmdlet_name = command.first
           command.unshift("(")
           %w{-Force -ForceBootstrap}.each do |arg|
             command.push(arg)
           end
           command.push("-RequiredVersion #{version}") if version
-          command.push("-Source #{new_resource.source}") if new_resource.source && command[1] =~ Regexp.union(/Install-Package/, /Find-Package/)
+          command.push("-Source #{new_resource.source}") if new_resource.source && cmdlet_name =~ Regexp.union(/Install-Package/, /Find-Package/)
+          command.push("-SkipPublisherCheck") if new_resource.skip_publisher_check && cmdlet_name !~ /Find-Package/
           command.push(").Version")
           command.join(" ")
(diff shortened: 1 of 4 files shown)


lib/chef/resource/powershell_package.rb
```diff
@@ -37,6 +37,7 @@ class PowershellPackage < Chef::Resource::Package
       property :package_name, [String, Array], coerce: proc { |x| [x].flatten }
       property :version, [String, Array], coerce: proc { |x| [x].flatten }
       property :source, [String]
+      property :skip_publisher_check, [true, false], default: false
```
tas50: If you add introduced: '14.2' we can make sure this displays correctly on the docs site when we start autogenerating resource docs. You should also add description: 'something something' so we know how to document this on the docs site

lib/chef/provider/package/powershell.rb
```diff
@@ -53,7 +53,9 @@ def candidate_version
         # Installs the package specified with the version passed else latest version will be installed
         def install_package(names, versions)
           names.each_with_index do |name, index|
-            powershell_out(build_powershell_package_command("Install-Package '#{name}'", versions[index]), timeout: new_resource.timeout)
+            cmd = powershell_out(build_powershell_package_command("Install-Package '#{name}'", versions[index]), timeout: new_resource.timeout)
+            next if cmd.nil?
```
tas50: Are you trying to check for an error here? next unless cmd.error? might make more sense if so

Happycoil: If I remember correctly `cmd` is `nil` if the command is successful, so this is to guard against a `NoMethodError` from the next line.

lib/chef/resource/powershell_package.rb
Happycoil: Added

lib/chef/provider/package/powershell.rb
lamont-granquist: that sounds like an enormous bug in powershell_out if its true, it should always return a mixlib-shellout object.

PR #7261, opened 2018-05-15
hostname: Anchor the hostnamectl regex so we always update when necessary

### LINKED ISSUE #7260: resource hostname: regex matches short hostnames AND fqdn
https://github.com/chef/chef/blob/c0609e449135fae43d436136a4f0fd3889a9b8f1/lib/chef/resource/hostname.rb#L119

This regex matches
`foo-bar`
as well as
`foo-bar.example.com`

This is a real deal breaker when trying to change a FQDN hostname into a short hostname

lamont-granquist:
closes #7260


lamont-granquist:
🚧 needs to be validated and needs a test

PR #7273, opened 2018-05-16
we removed the state default property in 13

thommay:
This sucks for folk upgrading where a cookbook uses a property called state, like https://github.com/chef-cookbooks/openssl/pull/70

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -23,6 +23,9 @@ github:
   # The file where our CHANGELOG is kept. This file is updated automatically with
   # details from the Pull Request via the `built_in:update_changelog` merge_action.
   changelog_file: "CHANGELOG.md"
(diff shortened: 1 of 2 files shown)


.expeditor/config.yml
```diff
@@ -23,6 +23,9 @@ github:
   # The file where our CHANGELOG is kept. This file is updated automatically with
   # details from the Pull Request via the `built_in:update_changelog` merge_action.
   changelog_file: "CHANGELOG.md"
+  # This will trigger an Omnibus build with EXPIRE_CACHE enabled after
```
coderanger: Intentional?

thommay: Yup; the pipelines are sadly fragile when changing ruby versions and this is the safest way to get working chef-12 builds.

Issue #7269, opened 2018-05-16
systemd_unit user mode exception


mal:
## Description
`systemd_unit` [documentation](https://docs.chef.io/resource_systemd_unit.html#properties) specifies `user` to be a `string`. However when using a string username, a `TypeError` is thrown:

```
no implicit conversion of String into Integer
```

## Chef Version
Chef: 14.1.1

## Platform Version
Ubuntu 18.04 Desktop

## Replication Case

```rb
systemd_unit 'testcase.service' do
  content <<~EOF
    [Unit] # junk definition, but enough to show issue
  EOF
  user ENV['USER']
  action :create
end
```

## Client Output & Stacktrace
https://gist.github.com/mal/dd155ccad07376157bed4e4206f8b12c

## Observations

`Etc.getpwuid` only appears to accept a UID, perhaps it should be using `Etc.getpwnam` instead? Or some method that decides which to use based on input type?

Potentially affects the systemd provider for the `service` resource too: https://github.com/chef/chef/pull/6833/files

mal:
Fixed by https://github.com/chef/chef/pull/7274

PR #7274, opened 2018-05-16
Fix systemd_unit user context

mal:
### Description

- Switches out `Etc.getpwuid` for `Etc.getpwnam` in `systemd_unit` provider and `systemd` provider for `service` resource in order to correctly handle usernames set on the resources.
- Fixes issue where `daemon-reload` would be called at the system level for user level unit changes which would leave the updated user units in an incomplete state.

### Issues Resolved

https://github.com/chef/chef/issues/7269

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -77,7 +77,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
 
   def get_systemctl_options_args
     if new_resource.user
-      uid = Etc.getpwuid(new_resource.user).uid
+      uid = Etc.getpwnam(new_resource.user).uid
       options = {
         :environment => {
           "DBUS_SESSION_BUS_ADDRESS" => "unix:path=/run/user/#{uid}/bus",
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -219,7 +219,7 @@ class Chef
       end
 
       def daemon_reload
-        shell_out_with_systems_locale!("#{systemctl_path} daemon-reload")
+        shell_out_with_systems_locale!("#{systemctl_cmd} daemon-reload", systemctl_opts)
       end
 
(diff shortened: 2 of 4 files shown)


lib/chef/provider/service/systemd.rb
```diff
@@ -77,7 +77,7 @@ def define_resource_requirements
 
   def get_systemctl_options_args
     if new_resource.user
-      uid = Etc.getpwuid(new_resource.user).uid
+      uid = Etc.getpwnam(new_resource.user).uid
```
thommay: This'll explode if users have previously been using UIDs here:
```
irb(main):003:0> Etc.getpwnam 501
        3: from /Users/thom/.rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        2: from (irb):3
        1: from (irb):3:in `getpwnam'
TypeError (no implicit conversion of Integer into String)
```
probably better to call the right thing if we have a string or an int.

mal:
I don't understand why the tests fail, it runs fine on my local machine. The errors read like the mocking of `Etc.getpwnam` isn't being done correctly, but it looks correct. Any pointers would be very welcome.

lib/chef/provider/service/systemd.rb
lamont-granquist: regardless we should accept both in the API.

there's fairly horrible problems with glibc caching+ruby and when chef creates the username where uids are required in apis like this in order to work around the problems (or else people need to run nscd, but everyone irrationally hates nscd).

mal: Any hope of that being a separate task? This was just meant to be a quick fix for something that's currently outright broken. Adding support and accompanying tests for allowing `user` to be a UID in the `service` and `systemd_unit` resources is not something I really want to take on.

thommay: Yeah, let's file a bug for that and merge this.

thommay:
Thanks @mal, this is a great fix!

PR #7278, opened 2018-05-17
WIP: 🚧 Experimenting with ChefSpec in core

coderanger:
Do not merge, I just need Travis.


tas50:
I assume we could pick this up now that we have the new functionality in a release of chefspec?

PR #7282, opened 2018-05-17
HTTP remote file provider should not fetch the file in --why-run mode

stuartpreston:

### Description

This change bypasses the logic that retrieves the content for remote files when used in --why-run mode.  Instead, just the outer converge_by message will be shown, i.e.:

```
Recipe: @recipe_files::C:/projects/repro/why_run_remote_file/repro.rb
  * remote_file[C:/Users/spreston\.chef\local-mode-cache\cache/push-jobs-client-2.5.0-1-x64.msi] action create
    - Would create new file C:/Users/spreston\.chef\local-mode-cache\cache/push-jobs-client-2.5.0-1-x64.msi
    - Would change dacl
    - Would change owner
    - Would change group
```

### Issues Resolved

Previously, the remote file was retrieved anyway for a checksum comparison even though it was not necessary to retrieve the whole file.  A number of cookbooks use remote_file internally to retrieve Chef installation files from our CDN and so this change benefits both Chef Software as well as customer bandwidth charges.

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/remote_file/content.rb
+++ b/lib/chef/provider/remote_file/content.rb
@@ -68,7 +68,7 @@ class Chef
 
         # Given a source uri, return a Tempfile, or a File that acts like a Tempfile (close! method)
         def grab_file_from_uri(uri)
-          Chef::Provider::RemoteFile::Fetcher.for_resource(uri, @new_resource, @current_resource).fetch
+          Chef::Provider::RemoteFile::Fetcher.for_resource(uri, @new_resource, @current_resource).fetch unless Chef::Config[:why_run]
         end
 
         def current_resource_matches_target_checksum?
--- a/lib/chef/provider/remote_file/content.rb
+++ b/lib/chef/provider/remote_file/content.rb
@@ -68,7 +68,7 @@ class Chef
 
         # Given a source uri, return a Tempfile, or a File that acts like a Tempfile (close! method)
         def grab_file_from_uri(uri)
-          Chef::Provider::RemoteFile::Fetcher.for_resource(uri, @new_resource, @current_resource).fetch unless Chef::Config[:why_run]
+          Chef::Provider::RemoteFile::Fetcher.for_resource(uri, @new_resource, @current_resource).fetch
         end
 
         def current_resource_matches_target_checksum?
--- a/lib/chef/provider/remote_file/http.rb
+++ b/lib/chef/provider/remote_file/http.rb
@@ -70,7 +70,7 @@ class Chef
             end
           else
             tempfile = http.streaming_request(uri, headers, orig_tempfile)
-          end
+          end unless Chef::Config[:why_run]
           if tempfile
             update_cache_control_data(tempfile, http.last_response)
             tempfile.close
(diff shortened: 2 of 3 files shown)


lamont-granquist:
i'm suspicious this is not the correct way to do this, but i wrote a bunch of text and then thought about it in the context of cookbook_file/template resources where you do always have access to the files on disk, and you wouldn't want this behavior, which points towards modifications not being made in Chef::Provider::File itself, but in the content related classes under remote_file.  still this feels side-effecty in that we're relying on a nil tempfile to do the right thing in why-run mode, when that would probably be a hard error in other content classes.

definitely needs unit tests -- unfortunately, because the file provider unit tests are easily the worst code i've ever written (that taught me a lot about how to not write unit tests that bad).

lamont-granquist:
So I guess the side effect of this is that why-run mode with remote_file will always report that the file would be updated even if it wouldn't?  And is there any behavior change around if the cache state exists or not?  Ideally if the cache state is there, we just trust it so if the checksum matches we report the file wouldn't be updated...

lamont-granquist:
So to be clear the use case I'm thinking of is *not* the use case which was reported.  It is the use case for the second time, when you have the cache file available, and it has the checksum of the target file in it.  In that case we would still want to blindly trust that the actual remote_file has not been updated.  But we can do a comparison between the target file and the cached checksum and if they match assume it would not be updated.  I think the existing code would report it always might be updated, so with a remote_file resource in the run list anywhere the why-run output would always be 'dirty' and never 'clean' -- which even if its a pile of lies, users are going to want to see that immediately after a successful convergence (particularly one which was fully idempotent with zero resources updated) that they get zero resources to be updated out of why-run.   I think otherwise we'll get lots of questions...

stuartpreston:
Agreed and thanks as ever for the comments - the original problem spotted with why-run mode is that it will stream the remote file *before* storing the checksum. In some cases (e.g remote_file on the Chef Client which was the original reason for this PR) this means hundreds of megs of download just to get a checksum.  If we changed to store the Etag from the HTTP response we may get further but this involves a significant refactoring.  It would be my preference in why-run more not to attempt any ‘real’ retrieval of the file concerned, but not sure what’s compatible with this second use case.

stuartpreston:
Closing this after triage discussion - significant reworking of the RemoteFile provider would be required to accomplish this.

lamont-granquist:
To capture discussion at least for my future memory self:

- we should look at the actual file on disk and compare to the checksum in the json the remote_file resource writes out -- if it fails, there's no question and we should why-run that it changes and avoid any disk access.
- similarly if the file doesn't exist on disk we can short-circuit everything and why run that it would change because that's easy to do.
- we can pull only the etags off the actual remote file and compare that to the json and if it changes then remote that it would why-run -- but this would still mean one network hit per why-run call (although we could avoid downloading the entire file).  this would be more correct, but it would mean network access so spamming 30,000 why-run CCRs would still potentially DDoS a remote service briefly.

that's a lot more than this PR was setting out to try to do though.

PR #7284, opened 2018-05-18
Mount: Fix errors on Windows when using the mount_point property

dheerajd-msys:
### Description
According to mount resource documentation for mount_type, "The directory (or path) in which the device is to be mounted. Default value: the name of the resource block. A customer had tried this and was getting errors when the name was anything but the mount_point. Looking at lib/chef/provider/mount/windows.rb, it shows that net_use is hard-wired to use only the resource name, which conflicts with the documentation. If we change it from name to mount_point, then it will work according to how the documentation specifies.

Windows mount uses `new_resource.name` instead of `new_resource.mount_point` when connecting to a CIFS share

### Resolved issue
https://getchef.zendesk.com/agent/tickets/18684

### DIFF
--- a/lib/chef/provider/mount/windows.rb
+++ b/lib/chef/provider/mount/windows.rb
@@ -40,10 +40,8 @@ class Chef
 
         def load_current_resource
           if is_volume(@new_resource.device)
-            @mount = Chef::Util::Windows::Volume.new(@new_resource.name)
-          elsif @new_resource.name.include?(":") && !@new_resource.mount_point #assume network drive
-            @mount = Chef::Util::Windows::NetUse.new(@new_resource.name)
-          else
+            @mount = Chef::Util::Windows::Volume.new(@new_resource.mount_point)
+          else #assume network drive
             @mount = Chef::Util::Windows::NetUse.new(@new_resource.mount_point)
           end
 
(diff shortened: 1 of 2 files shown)


lib/chef/provider/mount/windows.rb
```diff
@@ -41,8 +41,10 @@ def initialize(new_resource, run_context)
         def load_current_resource
           if is_volume(@new_resource.device)
             @mount = Chef::Util::Windows::Volume.new(@new_resource.name)
```
tas50: Is there a reason we don't use new_resource.mount_point here? It's the name property so it seems like we're working around the core bug without actually fixing it

lib/chef/provider/mount/windows.rb
```diff
@@ -41,8 +41,10 @@ def initialize(new_resource, run_context)
...
-          else #assume network drive
+          elsif @new_resource.name.include?(":") && !@new_resource.mount_point #assume network drive
             @mount = Chef::Util::Windows::NetUse.new(@new_resource.name)
```
tas50: This would also be mount_point not name

dheerajd-msys: I thought it similar but to make it backward compatible I committed this change. Anyways I will update this and make it `name` to `mount_point`.

btm:
obviously the change itself is simple enough that I'm fine with it. 

Can we have a unit or integration test here?

PR #7288, opened 2018-05-22
windows_task: Don't allow bad username/password to be provided to a task which will fail later

Vasu1105:
### Description
When wrong username and password provided in the windows_task recipe it creates the windows_task and raises the exception afterwards. It should not create task when username and password is incorrect.

### Issues Resolved
Updated win32-taskscheduler to set user information at the time of creation of task so that it will validate the user at that time and updated the method defination in windows_task provider.

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -133,10 +133,10 @@ class Chef
             converge_by("#{new_resource} task created") do
               task = TaskScheduler.new
               if new_resource.frequency == :none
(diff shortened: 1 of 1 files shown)


Vasu1105:
Note: This fix has dependency on https://github.com/chef/win32-taskscheduler/pull/31 so need to merge and release that fix and then merge this  Failing test cases here has the same reason. Once above PR get merged and released the we can re-run the test cases.

Vasu1105:
@tas50 & @btm this PR is ready to merge.

PR #7293, opened 2018-05-25
windows_task: Fix for task is not idempotent when task name includes parent folder

### LINKED ISSUE #7291: windows_task not idempotent when task_name includes parent folder
## Description

The windows_task not idempotent when task_name includes parent folder. The resource will converge every time as a result, even when properties have not changed.

## Chef Version

14.1.12

## Platform Version

Windows Server 2016

## Replication Case

1. Create a cookbook called windows_task_test.
2. Create a recipe called my_task.rb.
3. In recipes/my_task.rb configure a windows_task resource with a task_name that includeds a parent folder (see the example below).

  ```ruby
  windows_task 'MyTask Under MyOrg' do
    task_name '\MyOrg\MyTask'
    command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
    run_level :highest
    frequency :daily
    start_time '03:00'
    action :create
  end
  ```

4. Execute a chef-client run ensuring that the windows_task_test::my_task recipe is included in the runlist.

## Client Output

```
Starting Chef Client, version 14.1.12
[2018-05-24T11:17:44-06:00] INFO: *** Chef 14.1.12 ***
[2018-05-24T11:17:44-06:00] INFO: Platform: x64-mingw32
[2018-05-24T11:17:44-06:00] INFO: Chef-client pid: 2588
[2018-05-24T11:17:44-06:00] INFO: The plugin path C:\chef\ohai\plugins does not exist. Skipping...
Creating a new client identity for Administrator using the validator key.
[2018-05-24T11:17:47-06:00] INFO: Client key C:/ChefInstall/.chef/Administrator.pem is not present - registering
[2018-05-24T11:17:47-06:00] INFO: Setting the run_list to [#<Chef::RunList::RunListItem:0x0000000005bc0228 @version=nil, @type=:recipe, @name="windows_task_test::my_task">] from CLI options
[2018-05-24T11:17:47-06:00] INFO: Run List is [recipe[windows_task_test::my_task]]
[2018-05-24T11:17:47-06:00] INFO: Run List expands to [windows_task_test::my_task]
[2018-05-24T11:17:47-06:00] INFO: Starting Chef Run for Administrator
[2018-05-24T11:17:47-06:00] INFO: Running start handlers
[2018-05-24T11:17:47-06:00] INFO: Start handlers complete.
resolving cookbooks for run list: ["windows_task_test::my_task"]
[2018-05-24T11:17:48-06:00] INFO: Loading cookbooks [windows_task_test@0.0.0]
Synchronizing Cookbooks:
[2018-05-24T11:17:48-06:00] INFO: Storing updated cookbooks/windows_task_test/recipes/my_task.rb in the cache.
  - windows_task_test (0.0.0)
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 1 resources
Recipe: windows_task_test::my_task
  * windows_task[MyTask Under MyOrg] action create[2018-05-24T11:17:48-06:00] INFO: Processing windows_task[MyTask Under MyOrg] action create (windows_task_test::my_task line 1)

    - windows_task[MyTask Under MyOrg] task created
[2018-05-24T11:17:48-06:00] INFO: Chef Run complete in 0.792092 seconds

Running handlers:
[2018-05-24T11:17:48-06:00] INFO: Running report handlers
Running handlers complete
[2018-05-24T11:17:48-06:00] INFO: Report handlers complete
Chef Client finished, 1/1 resources updated in 03 seconds
```

## Stacktrace

NA

Vasu1105:
### Description
The windows_task not idempotent when task_name includes parent folder. The resource will converge every time as a result, even when properties have not changed.

```
windows_task 'MyTask Under MyOrg' do
task_name '\MyOrg\MyTask'
command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName 
 Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
 run_level :highest
 frequency :daily
 start_time '03:00'
 action :create
end
```

### Issues Resolved
Fixes #7291

Tested with following recipes

```
windows_task 'MyTask Under MyOrg' do
  task_name '\MyOrg\MyTask'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '04:00'
  action :create
end

windows_task 'MyTask Under MyOrg' do
  task_name '\MyOrg\MyTask'
  action :delete
end

windows_task 'MyTask Under MyOrg' do
  task_name '\MyOrg\MyTask1'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '04:00'
  action :create
end

windows_task 'MyTask Under MyOrg' do
  task_name '\MyOrg\MyTask1'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :once
  start_time '07:00'
  action :create
end

windows_task 'MyTask MyFolder MyOrg' do
  task_name '\MyOrg\MyFolder\MyTask'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '03:00'
  action :create
end

windows_task 'MyTask Under MyOrg' do
  task_name '\MyOrg\MyFolder\MyTask'
  action :delete
end

windows_task 'MyTask1 MyFolder MyOrg' do
  task_name '\MyOrg\MyFolder\MyTask1'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '07:00'
  action :create
end

windows_task 'MyTask1 MyFolder MyOrg' do
  task_name '\MyOrg\MyFolder\MyTask1'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '09:00'
  action :create
end

windows_task 'foo bar MyTask1' do
  task_name '\foo\bar\MyTask1'
  command 'powershell -NoLogo -NonInteractive -ExecutionPolicy Bypass "Write-EventLog -LogName Application -EventID 10000 -Source Chef -EntryType Information -Message MyTask_Under_MyOrg"'
  run_level :highest
  frequency :daily
  start_time '09:00'
  action :create
end

windows_task 'chef-client-test' do
 task_name 'chef-client-test11' # 1http://bit.ly/1WDZ1kn
 command "chef-client -W -L 'C:\\chef\\chef-ad-join.log'"
 run_level :highest
 frequency :daily
end
```

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -100,15 +100,13 @@ class Chef
 
         def load_current_resource
           @current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
-          task = TaskScheduler.new
-          if task.exists?(new_resource.task_name)
-            @current_resource.exists = true
+          task = TaskScheduler.new(new_resource.task_name, nil, "\\", false)
+          @current_resource.exists = task.exists?(new_resource.task_name)
+          if @current_resource.exists
             task.get_task(new_resource.task_name)
             @current_resource.task = task
             pathed_task_name = new_resource.task_name.start_with?('\\') ? new_resource.task_name : "\\#{new_resource.task_name}"
             @current_resource.task_name(pathed_task_name)
-          else
-            @current_resource.exists = false
           end
           @current_resource
         end
--- a/spec/functional/resource/windows_task_spec.rb
+++ b/spec/functional/resource/windows_task_spec.rb
@@ -1109,6 +1109,60 @@ describe Chef::Resource::WindowsTask, :windows_only do
     end
   end
 
+  context "task_name with parent folder" do
+    describe "task_name with path '\\foo\\chef-client-functional-test' " do
+      let(:task_name) { "\\foo\\chef-client-functional-test" }
+      after { delete_task }
+      subject do
+        new_resource = Chef::Resource::WindowsTask.new(task_name, run_context)
+        new_resource.command task_name
+        new_resource.run_level :highest
+        new_resource.frequency :once
+        new_resource.execution_time_limit = 259200 / 60 # converting "PT72H" into minutes and passing here since win32-taskscheduler accespts this
+        new_resource
+      end
+
+      it "creates the scheduled task with task name 'chef-client-functional-test' inside path '\\foo'" do
+        call_for_create_action
+        current_resource = call_for_load_current_resource
+        expect(current_resource.exists).to eq(true)
+        expect(current_resource.task.application_name).to eq(task_name)
+      end
+
+      it "does not converge the resource if it is already converged" do
+        subject.run_action(:create)
+        subject.run_action(:create)
+        expect(subject).not_to be_updated_by_last_action
+      end
+    end
+
+    describe "task_name with path '\\foo\\bar\\chef-client-functional-test' " do
+      let(:task_name) { "\\foo\\bar\\chef-client-functional-test" }
+      after { delete_task }
+      subject do
+        new_resource = Chef::Resource::WindowsTask.new(task_name, run_context)
+        new_resource.command task_name
+        new_resource.run_level :highest
+        new_resource.frequency :once
+        new_resource.execution_time_limit = 259200 / 60 # converting "PT72H" into minutes and passing here since win32-taskscheduler accespts this
+        new_resource
+      end
+
+      it "creates the scheduled task with task with name 'chef-client-functional-test' inside path '\\foo\\bar' " do
+        call_for_create_action
+        current_resource = call_for_load_current_resource
+        expect(current_resource.exists).to eq(true)
+        expect(current_resource.task.application_name).to eq(task_name)
+      end
+
+      it "does not converge the resource if it is already converged" do
+        subject.run_action(:create)
+        subject.run_action(:create)
+        expect(subject).not_to be_updated_by_last_action
+      end
+    end
+  end
+
   describe "Examples of idempotent checks for each frequency" do
     after { delete_task }
     context "For frequency :once" do


idflyfish:
Just finished testing and confirmed that this fixes my issue.

lib/chef/provider/windows_task.rb
```diff
@@ -100,9 +100,27 @@ class WindowsTask < Chef::Provider
...
           @current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
-          task = TaskScheduler.new
-          if task.exists?(new_resource.task_name)
+          if new_resource.task_name.split("\\").size > 1
+            *path, task_name = new_resource.task_name.split("\\")
+            path = path.join("\\")
+            begin
+              task = TaskScheduler.new(new_resource.task_name, nil, path, false)
+            rescue Exception => e
+              if e.message == "folder \'#{path}\' not found"
```
coderanger: Single quotes don't need to be escaped in double quotes.

lib/chef/provider/windows_task.rb
```diff
@@ -100,9 +100,27 @@ class WindowsTask < Chef::Provider
 
         def load_current_resource
           @current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
-          task = TaskScheduler.new
-          if task.exists?(new_resource.task_name)
+          if new_resource.task_name.split("\\").size > 1
```
coderanger: I think this is just a complex way of checking if `task_name.include?("\\")`?

lib/chef/provider/windows_task.rb
```diff
@@ -100,9 +100,27 @@ class WindowsTask < Chef::Provider
...
+            *path, task_name = new_resource.task_name.split("\\")
+            path = path.join("\\")
+            begin
+              task = TaskScheduler.new(new_resource.task_name, nil, path, false)
+            rescue Exception => e
```
coderanger: Looks like this is an `ArgumentError` upstream (https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L164) so no need to cast such a wide net.

Vasu1105: I am adding that block so that if I get folder not found error I can reset the folder to root path i.e. "\\". Otherwise just raise the error coming from the library.

coderanger: Indeed, but you can rescue `ArgumentError` instead of all exceptions it looks like. Unless that error can come from elsewhere. Slightly reduces the chances of badness :)

coderanger:
I don't really understand the goal of this change. If the task folder doesn't exist, we shouldn't ignore the folder I would think, we should create it? (which is handled by passing `true` for the force parameter)

Vasu1105:
@coderanger I updated my comment above. I did check with force parameter but its not working. And there are somethings which are not written well in win32-taskscheduler and might need refactor which might result in changes to be done at this end to. So for me this is the quick fix which is not making us to do the changes at both the side.

lib/chef/provider/windows_task.rb
Vasu1105: I am not sure if I am missing something but I am just checking if there is folder not found error then set root folder else raise the exception in else block.

coderanger:
https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L161-L162 is the code for the force option, and it seems correct given my limited knowledge of this OLE namespace. With this patch as written, it seems like we would just create a second, unrelated task object?

Vasu1105:
@coderanger https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L161 this particular line throws error undefined method CreateFolder for nil::NilClass since @root is here comes as nil. I have given a fix for this here https://github.com/chef/win32-taskscheduler/pull/32/files.  About the task object I need to do that to reset the folder path to root "//" because currently there is no other method using which I can reset the folder to root. And to answer why I need to reset it to root is,  in win32-taskscheduler library which internally calls this method in many places https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L1308 and here https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L1311 if you see it uses @root object to create or update the task so if I don't reset the @root to "//" it will remain as "//foo" in case where user is using parent folder and it results into creating another f older inside "//foo" with same name. I can'r really walk through each and every issues in win32-taskscheduler here. So as I mentioned what I think this is the quick fix right now without doing any code changes in win32-taskscheduler. We are planning refactoring of win32-taskscheduler after which we can modify the code in windows_task accordingly.

Vasu1105:
@coderanger I am not seeing I am having any orphan object I am just reinitialise the task object.  Please point me where I am doing that.

Vasu1105:
As per review comments I am throwing fatal error for folder not found.

coderanger:
It is worth reiterating that I am only one voice on this and don't speak for the whole maintainer team (and definitely not for Chef Software :)

idflyfish:
Hi @coderanger Is there anything else needed for your approval? I'd be happy to help out where I can.

coderanger:
I tend to leave final approval of Msys PRs to Chef Software employees given the complex relationships involved.

lib/chef/provider/windows_task.rb
btm: @Vasu1105 @coderanger is saying that because in [win32-taskscheduler](https://github.com/chef/win32-taskscheduler/blob/ole/lib/win32/taskscheduler.rb#L164) we are raising `ArgumentError`, here you can `rescue ArgumentError => e` instead which is slightly safer coding behavior than rescuing Exception. This way your rescue block is only run for the type of exception you are explicitly looking for, rather than all exceptions. Rescuing all exceptions can have unintended consequences. See also https://github.com/rubocop-hq/ruby-style-guide#no-blind-rescues for one example of unintended consequences.

lib/chef/provider/windows_task.rb
```diff
@@ -229,6 +230,27 @@ def action_disable
...
+        def task_exists?
+          task = TaskScheduler.new(new_resource.task_name, nil, "\\", false)
+          if new_resource.task_name.include?("\\")
+            *path, task_name = new_resource.task_name.split("\\")
+            path = path.join("\\")
+            begin
+              task = TaskScheduler.new(new_resource.task_name, nil, path, false)
+            rescue Exception => e
+              if e.message == "folder '#{path}' not found"
+                logger.error "Folder '#{path} not found calling create."
```
btm: `task_exists?` itself doesn't seem to create anything, so I'm not sure what 'calling create' implies here. Can you improve this message? I can't suggest something at the moment as I'm now sure what the behavior is yet.

btm:
> there are somethings which are not written well in win32-taskscheduler and might need refactor which might result in changes to be done at this end to. 

@Vasu1105 one way or another that sentence is acknowledging there's some debt in win32-taskscheduler that we need to go back and fix. I would expect `TaskScheduler.new('\MyOrg\MyTask').exists?` to return true if that task exists, and false if either the task does not exist or the folder does not exist. Is that really what we need to fix here?

Vasu1105:
@btm yes we need to fix the exists? which can return false if folder or task not exists? currently its just checking if task exist in given folder. so currently its always returning false for the task which has parent folder like "\\foo\\bar" since in constructor the folder set to "\\" and to "\\foo\\bar". I can go back to win32-taskscheduler and fix the exists? and work on this PR after that fix. Let me know your thoughts meanwhile I start working on fixing the exists? in win32-taskscheduler.

Vasu1105:
@btm I have updated the code as per new changes done in win32-taskscheduler library once we release the new version of lwin32-taskscheduler library specs will be green.

lib/chef/provider/windows_task.rb
```diff
@@ -100,15 +100,13 @@ class WindowsTask < Chef::Provider
 
         def load_current_resource
           @current_resource = Chef::Resource::WindowsTask.new(new_resource.name)
-          task = TaskScheduler.new
-          if task.exists?(new_resource.task_name)
-            @current_resource.exists = true
+          task = TaskScheduler.new(new_resource.task_name, nil, "\\", false)
+          @current_resource.exists = task.exists?(new_resource.task_name)
+          if @current_resource.exists == true
```
coderanger: You never need `== true` in Ruby.

Vasu1105:
@tas50 & @btm rebased this branch and this PR is now is ready for merge

btm:
@Vasu1105 thanks for solving this in the library. I think that was a good route.

PR #7294, opened 2018-05-25
[MSYS-814] - Manual cron job deletion

Nimesh-Msys:
- Commit to handle deletion of manually created cron jobs, other than chef
- Also fixed an existing issue: If chef created job is not the last job written in crontab, then even after its deletion, message was: `save unmodified crontab`

Signed-off-by: nimesh-msys <nimesh.patni@msystechnologies.com>

### Description

Cron Job Deletion process: Iteration over each line of crontab and insert the contents(line) into a local string 'crontab'.  When we found "# Chef Name: #{new_resource.name}", this line(the header comment) and the next line(with cron job) will be skipped. Hence 'crontab' string only contains the remaining line of codes. This 'crontab' then completely overwrites crontab file. Thus we were only deleting the jobs which were created by chef (by header search)
Following code changes includes: during iteration if we encounter a 'cron job pattern', which is exactly similar to what user has passed(for deletion), skip that line too. This skipping will be irrespective of resource name, hence the given job could also be deleted which was created manually.

### Issues Resolved
[MSYS-814](https://chefio.atlassian.net/browse/MSYS-814)

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/cron.rb
+++ b/lib/chef/provider/cron.rb
@@ -155,46 +155,48 @@ class Chef
       end
 
       def action_delete
-        crontab = String.new
-        cron_found = false
-        cron_job_found = false
-        manual_job_found = false
-        read_crontab.each_line do |line|
-          case line.chomp
-          when "# Chef Name: #{new_resource.name}"
-            cron_job_found = true
-            cron_found = true
-            next
-          when ENV_PATTERN
-            next if cron_found
-          when SPECIAL_PATTERN
-            if cron_found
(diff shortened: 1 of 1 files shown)


coderanger:
Also this needs a metric ass-load of tests :D

Nimesh-Msys:
These code changes may not be required. There are few complexities while handling environment variables. It may also affect cron entries created by chef.
Have sent a [request](https://github.com/chef/chef-web-docs/pull/1487) for updating the document including workaround to achieve this.

PR #7181, opened 2018-04-24
Ensure we have a sane PATH when calling sysctl -p

tomdoherty:
Signed-off-by: Tom Doherty <tom.doherty@fixnetix.com>

### Description

Fixes issue on RHEL 6/7 where sysctl isn't on PATH when chef-client is called via cron

### Issues Resolved

#7180

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #7298, opened 2018-05-25
add default_env flag to shell_out and execute resource

lamont-granquist:
this will wind up replacing shell_out_with_systems_locale.

also fixes an issue in the sysctl provider.

closes #7181

### DIFF
--- a/lib/chef/resource/execute.rb
+++ b/lib/chef/resource/execute.rb
@@ -1,7 +1,7 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
 # Author:: Tyler Cloke (<tyler@chef.io>)
-# Copyright:: Copyright 2008-2017, Chef Software Inc.
+# Copyright:: Copyright 2008-2018, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


lamont-granquist:
These have been roughly equivalent since 13.0:

```ruby
execute "ls -la"

converge_by "execute ls -la" do
   shell_out_with_systems_locale!("ls -la")
end
```

these are now roughly equivalent:

```ruby
execute "ls -la" do
  internal: true
end

converge_by "execute ls -la" do
   shell_out!("ls -la")
end
```

- generally where we use the execute resource internally we should be setting `internal true` nearly everywhere (unless we're doing something like starting a service for the user).

- `shell_out_with_systems_locale` should be replaced by using `internal false` either on shell_out! or on the execute resource.  this API is now badly named since it also does magic to the PATH variable.

lib/chef/resource/execute.rb
```diff
@@ -64,6 +64,8 @@ def command(arg = nil)
 
       property :group, [ String, Integer ]
       property :live_stream, [ TrueClass, FalseClass ], default: false
+      # internal defaults to `false` so that the command execution more exactly matches what the user gets on the command line without magic
+      property :internal, [ TrueClass, FalseClass ], desired_state: false, default: false
```
tas50: Add the description field here so we can more easily get it into the docs

lamont-granquist:
went with 'default_env'

PR #7300, opened 2018-05-30
resource_inspector: Add default values for properties

thommay:


coderanger:
We should maybe consider adding a `default_description` field or something to handle the case of lazy defaults or defaults where `nil` means "do something fancy".

tas50:
a default_description field would probably be useful, but this is a solid step in the right direction

PR #7302, opened 2018-05-30
[For Discussion] Experimental systemd provider for ifconfig

thommay:
:construction: DO NOT MERGE :construction:

Ubuntu 18.04 uses systemd-networkd to manage networking for servers, so we eventually need to support this.
(Actually they generate systemd unit files via something they NIHd called netplan, but we'll ignore that since it's unlikely anyone else'll ever use it).
Sadly, systemd-networkd is not especially sensible, and appears to be poorly supported by systemd - to the extent that you can't actually verify network unit files using the usual `systemd-analyze` command. The real problem is that you can only have one unit per network *device*, so this will end up needing to either accumulate into a single unit, or use partial unit files as needed.

The other problem is that the `ifconfig` resource itself is not really fit for purpose - we need to do a fairly serious rethink, and we might look at the `network_interfaces_2` cookbook for inspiration here. 

Using `ip` rather than `ifconfig` is so much nicer so, so there's that…

### DIFF
--- a/lib/chef/provider/ifconfig/systemd.rb
+++ b/lib/chef/provider/ifconfig/systemd.rb
@@ -1,0 +1,174 @@
+#
+# Copyright:: Copyright 2018, Chef Software, Inc
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+require "ipaddr"
+require "iniparse"
+
+require "chef/mixin/which"
+require "chef/mixin/shell_out"
+
+class Chef
+  class Provider
+    class Ifconfig
+      class Systemd < Chef::Provider::Ifconfig
+        extend Chef::Mixin::Which
+        include Chef::Mixin::ShellOut
+        extend Chef::Mixin::ShellOut
+
+        provides :ifconfig do
+          systemd_unit_enabled?("systemd-networkd")
+        end
+
+        def load_current_resource
+          @current_resource = Chef::Resource::Ifconfig.new(new_resource.name)
+
+          @status = shell_out_with_systems_locale!("ip", "-json", "addr", "list")
+          json = JSON.parse @status.stdout
+          interfaces = json.each_with_object({}) { |iface, acc| acc[iface["ifname"]] = iface }
+
+          iface = driver_name
+          interface = interfaces.fetch(iface, nil)
+          if !interface.nil?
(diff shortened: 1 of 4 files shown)


lib/chef/provider/ifconfig/systemd.rb
```diff
@@ -0,0 +1,174 @@
...
+  class Provider
+    class Ifconfig
+      class Systemd < Chef::Provider::Ifconfig
+        extend Chef::Mixin::Which
+        include Chef::Mixin::ShellOut
+        extend Chef::Mixin::ShellOut
+
+        provides :ifconfig do
+          systemd_unit_enabled?("systemd-networkd")
+        end
```
lamont-granquist: since this ultimately does a shell_out on every evaluation it is going to slow everyone down evaluating this.  you could make it a bit better by short-circuiting the shell_out if the which fails, but it'll still be slow on systems that have systemd.

its possible this needs to be an `@class_var ||= systemd_unit_enabled?("systemd-networkd")` but that would mean that nobody could use chef to enable `systemd-networkd` and then use this.

lib/chef/provider/ifconfig/systemd.rb
```diff
@@ -0,0 +1,174 @@
...
+
+        def self.systemd_unit_enabled?(unit)
+          systemctl_execute!("is-enabled", unit).exitstatus == 0
+        end
+
+        def self.systemctl_execute!(action, unit)
+          systemctl_path = which("systemctl")
+          systemctl_cmd = "#{systemctl_path} --system"
+          shell_out_with_systems_locale!("#{systemctl_cmd} #{action} #{Shellwords.escape(unit)}")
```
lamont-granquist: these should just be `shell_out(..., default_env: false)` now.

lamont-granquist: actually i think none of this should be `default_env: false` or `_with_systems_locale` and they should just be straight `shell_out`

lib/chef/provider/ifconfig/systemd.rb
thommay: Yeah, one of the points of this as an experiment was to see what needs to happen to support systemd properly. We could short circuit this some by checking if we already think that systemd is in use, but really we need to set up some methods to query systemd better and use that consistently. We're only going to need to do this more frequently in the future.

lib/chef/provider/ifconfig/systemd.rb
```diff
@@ -0,0 +1,174 @@
...
+        end
+
+        private
+
+        def unit_dir
+          "/etc/systemd/network"
+        end
+
+        def generate_network_unit
```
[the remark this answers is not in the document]
thommay: Should it? Yes. Can it? no. because why? because `systemd-analyze` can't verify network units currently afaict, and so it always fails.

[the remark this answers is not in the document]
nathwill: fwiw, systemd_unit lets you disable the verification. the next problem would be needing to adjust the unit path... we could create a systemd_unit inside the provider with a nothing action and use #to_ini as the arg to a file's content property to get the content out...

nathwill:
imo, it's probably easier and better to make a clean break with ifconfig concepts and create new systemd_link, systemd_network, and systemd_netdev resources.

PR #7313, opened 2018-05-30
convert a_to_s to shell_out_compact in DNF/yum

lamont-granquist:
work towards finishing deprecating the a_to_s api.


tas50:
Looks good with the string -> array fix

PR #7324, opened 2018-05-30
Support signing with ssh-agent

coderanger:
Depends on https://github.com/chef/mixlib-authentication/pull/36

The new mode is behind a config option (`ssh_agent_signing`). In the release notes I only mention knife but technically this should work in any of the binaries. Tested against real ssh-agent locally and it seems to work, not sure what the best way to do an integration test on this would be since zero specifically doens't check keys I think.


tas50:
If we're depending on that yet to be released mixlib-authentication PR don't we need to release that and bump the dep in chef/chef?

coderanger:
Yep, can't merge this yet. Thom approved the other PR once already but then I added a new deprecation warning so that needs a re-review probably :)

coderanger:
Fun potential side-effect of this (which I'm not going to test), because gpg-agent can pretend to be an ssh-agent, this would mean that any GPG hardware keys (Yubikey, HSMs, etc) could be used for Chef auth.

Issue #7303, opened 2018-05-30
knife configure no longer shows where the config file is written to


tas50:
If you have no config files (knife.rb or credential) and you run knife configure then knife doesn't properly write out where the config is written to.

### When there's a knife.rb file

```
Please enter the chef server URL: [https://Tim.local/organizations/myorg]
Please enter an existing username or clientname for the API: [tsmith]
*****

You must place your client key in:
  /Users/tsmith/.chef/tsmith.pem
Before running commands with Knife

*****
Configuration file written to /Users/tsmith/.chef/knife.rb
```

### Without a knife.rb file

```
WARNING: No knife configuration file found. See https://docs.chef.io/config_rb_knife.html for details.
Please enter the chef server URL: [https://Tim.local/organizations/myorg]
Please enter an existing username or clientname for the API: [tsmith]
*****

You must place your client key in:
  /Users/tsmith/.chef/tsmith.pem
Before running commands with Knife

*****
Configuration file written to
```

Issue #7304, opened 2018-05-30
When knife.rb is present knife configure incorretly says it's writing there


tas50:
knife configure always writes to the credentials file, but when a knife.rb is present it will incorrectly tell you it was writing to that file:

### Results with a knife.rb file present

```
Please enter the chef server URL: [https://Tim.local/organizations/myorg]
Please enter an existing username or clientname for the API: [tsmith]
*****

You must place your client key in:
  /Users/tsmith/.chef/tsmith.pem
Before running commands with Knife

*****
Configuration file written to /Users/tsmith/.chef/knife.rb
```

Where it actually wrote the file:

```
❰tsmith❙~/.chef❱✔≻ cat credentials                                                                                                                                                  
[default]
client_name     = 'tsmith'
client_key      = '/Users/tsmith/.chef/tsmith.pem'
chef_server_url = 'https://Tim.local/organizations/myorg'
```

PR #7325, opened 2018-05-30
Properly print path to config file to the screen in knife configure

tas50:
config[:config_file] is the knife.rb path so we'd either make it look like we were writing to a knife.rb file or it was nil and we'd print a blank file path. Toss in some yard and shuffle things around while I'm in here.


tas50:
Fixes #7303 #7304

tas50:
Proxy tests are broken and unrelated to this change

PR #7326, opened 2018-05-31
Better errors with uninstalled official knife plugins

tas50:
Remove the mention of Chef .10
Remove terremark and bluebox
Add more modern plugins
Avoid some extra work we didn't need to do

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -272,8 +272,7 @@ class Chef
         if category_commands = guess_category(args)
           list_commands(category_commands)
         elsif OFFICIAL_PLUGINS.include?(args[0]) # command was an uninstalled official chef knife plugin
-          ui.info("You can install the plugin with `(sudo) gem install knife-#{args[0]}`")
-          ui.info("Use `chef gem install knife-#{args[0]}` instead if using ChefDK")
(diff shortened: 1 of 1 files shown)


lib/chef/knife.rb
```diff
@@ -272,10 +271,9 @@ def subcommand_not_found!(args)
 
         if category_commands = guess_category(args)
           list_commands(category_commands)
-        elsif missing_plugin = ( OFFICIAL_PLUGINS.find { |plugin| plugin == args[0] } )
-          ui.info("The #{missing_plugin} commands were moved to plugins in Chef 0.10")
-          ui.info("You can install the plugin with `(sudo) gem install knife-#{missing_plugin}`")
-          ui.info("Use `chef gem install knife-#{missing_plugin}` instead if using ChefDK")
+        elsif OFFICIAL_PLUGINS.include?(args[0]) # command was an uninstalled official chef knife plugin
+          ui.info("You can install the plugin with `(sudo) gem install knife-#{args[0]}`")
```
coderanger: I would just remove this, people not using ChefDK are on their own IMO.

tas50: I was kinda leaning that way

PR #7328, opened 2018-05-31
Ensure that reboot requests on AD Join work

thommay:


tas50:
That's do it. Might wanna yank that declare_resource out since we don't need it and it just makes reading this more complex

thommay:
> Might wanna yank that declare_resource out since we don't need it and it just makes reading this more complex

Done

lib/chef/resource/windows_ad_join.rb
```diff
@@ -86,6 +86,17 @@ def on_domain?
           raise "Failed to check if the system is joined to the domain #{new_resource.domain_name}: #{node_domain.stderr}}" if node_domain.error?
           node_domain.stdout.downcase.strip == new_resource.domain_name.downcase
         end
+
+        def clarify_reboot(reboot_action)
```
tas50: Maybe a comment to why we're actually doing this also. Makes sense in this context, but might not be entirely clear in a few years

PR #7330, opened 2018-06-01
Make shell_out_compact automatically pull timeouts off the resource + remove uses of shell_out_compact_timeout

### LINKED ISSUE #6288: Portage package provider does not respect timeouts in declared package/portage_package resources
## Description

The portage package provider does not respect a timeout set in package resource declarations.   For example:

```
portage_package "dev-db/postgis" do
  version '2.2.2'
  action :install
  timeout 1200
end
```

That will timeout at 600 (which seems to be the default in the shellout mixing via its DEFAULT_READ_TIMEOUT constant) seconds if still running.  This:

```
portage_package "dev-db/postgis" do
  version '2.2.2'
  action :install
  timeout 30
end
```

Will blow past the 30 second timeout and keep running, up to the 600 second default shellout mixin timeout.

After poking around and testing it seems this can be quickly fixed by replacing the `#shell_out_compact!!` in the portage provider's `install_package` method with a call to `#shell_out_compact_timeout!` here:

https://github.com/chef/chef/blob/master/lib/chef/provider/package/portage.rb#L115

## Chef Version

12.10.24.  Also tested with 12.21 and checked that the line I fixed still uses `#shell_out_compact!` in the current public code base, i.e. 13.3.18.

## Platform Version

Engine Yard Gentoo Linux (EYGL) 2016.06

## Replication Case

Run a recipe with the above described postgis resource declaration with a 30 second timeout on a Gentoo system and watch the build blow by 30 seconds with no timeout.

## Client Output

```
This happened with the 1200 second timeout in the original resource declaration:

Error executing action `install` on resource 'portage_package[dev-db/postgis]'

Mixlib::ShellOut::CommandTimeout
--------------------------------
Command timed out after 600s:
Command exceeded allowed execution time, process terminated
---- Begin output of emerge -g -n --color n --nospinner --quiet =dev-db/postgis-2.2.2 ----
STDOUT: Local copy of remote index is up-to-date and will be used.
>>> Verifying ebuild manifests
>>> Running pre-merge checks for dev-db/postgis-2.2.2
>>> Emerging (1 of 1) dev-db/postgis-2.2.2::gentoo
STDERR:
---- End output of emerge -g -n --color n --nospinner --quiet =dev-db/postgis-2.2.2 ----
Ran emerge -g -n --color n --nospinner --quiet =dev-db/postgis-2.2.2 returned 143
```

lamont-granquist:
Collapses down the API a bit better.

I've got some possibly overly paranoid gating to make sure this wouldn't affect anyone's cookbooks, but at the same time this only affects shell_out_compact(), not shell_out() itself.

It found and fixed at least one reported bug:  #6288

Since portage is a compile-from-source I bumped the default timeout up to 3600 on that one and it now should obey user-set timeouts.

Code timeouts always take precedence now, then the new_resource.timeout, then it applies the 900 timeout if that is nil.

closes #6288

### DIFF
--- a/lib/chef/mixin/shell_out.rb
+++ b/lib/chef/mixin/shell_out.rb
@@ -62,7 +62,7 @@ class Chef
       #
 
       def shell_out_compact(*args, **options)
-        options = Chef::Mixin::ShellOut.maybe_add_timeout(options)
+        options = Chef::Mixin::ShellOut.maybe_add_timeout(self, options)
         if options.empty?
           shell_out(*clean_array(*args))
         else
@@ -71,7 +71,7 @@ class Chef
       end
 
       def shell_out_compact!(*args, **options)
-        options = Chef::Mixin::ShellOut.maybe_add_timeout(options)
+        options = Chef::Mixin::ShellOut.maybe_add_timeout(self, options)
         if options.empty?
           shell_out!(*clean_array(*args))
         else
@@ -79,12 +79,13 @@ class Chef
         end
       end
 
+      # module method to not pollute namespaces, but that means we need self injected as an arg
       # @api private
-      def self.maybe_add_timeout(options)
-        if is_a?(Chef::Provider) && !new_resource.is_a?(Chef::Resource::LWRPBase) && new_resource.respond_to?(:timeout) && !options.key?(:timeout)
+      def self.maybe_add_timeout(obj, options)
+        if obj.is_a?(Chef::Provider) && !obj.new_resource.is_a?(Chef::Resource::LWRPBase) && obj.new_resource.respond_to?(:timeout) && !options.key?(:timeout)
           options = options.dup
           # historically resources have not properly declared defaults on their timeouts, so a default default of 900s was enforced here
-          options[:timeout] = new_resource.timeout ? new_resource.timeout.to_f : 900
+          options[:timeout] = obj.new_resource.timeout ? obj.new_resource.timeout.to_f : 900
         end
         options
       end
--- a/lib/chef/mixin/shell_out.rb
+++ b/lib/chef/mixin/shell_out.rb
@@ -62,6 +62,7 @@ class Chef
       #
 
       def shell_out_compact(*args, **options)
+        options = maybe_add_timeout(options)
         if options.empty?
           shell_out(*clean_array(*args))
         else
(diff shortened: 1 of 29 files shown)


coderanger:
My only worry is the naming of `maybe_add_timeout`, do we want to leave that as an instance method or maybe make it a module method so it doesn't get dumped into everything that uses the mixin?

coderanger:
Flip side, leaving it as an instance method allows hackery like redefining it in your resource/provider?

lib/chef/mixin/shell_out.rb
```diff
@@ -70,29 +71,29 @@ def shell_out_compact(*args, **options)
...
         if options.empty?
           shell_out!(*clean_array(*args))
         else
           shell_out!(*clean_array(*args), **options)
         end
       end
 
-      # helper sugar for resources that support passing timeouts to shell_out
+      def maybe_add_timeout(options)
+        if is_a?(Chef::Provider) && !new_resource.is_a?(Chef::Resource::LWRPBase) && new_resource.respond_to?(:timeout) && !options.key?(:timeout)
```
coderanger: Should probably check for `respond_to?(:new_resource)` before trying to call it.

lib/chef/provider/package.rb
```diff
@@ -307,7 +307,7 @@ def unlock_package(name, version)
       # used by subclasses.  deprecated.  use #a_to_s instead.
       def expand_options(options)
         # its deprecated but still work to do to deprecate it fully
-        #Chef.deprecated(:package_misc, "expand_options is deprecated, use shell_out_compact or shell_out_compact_timeout instead")
+        #Chef.deprecated(:package_misc, "expand_options is deprecated, use shell_out_compact or shell_out_compact instead")
```
thommay: heh

lib/chef/mixin/shell_out.rb
lamont-granquist: Chef::Provider objects all have new_resource's though.  i had that, but it seemed redundant...

lib/chef/provider/package.rb
lamont-granquist: lulz

lamont-granquist:
hmm, right, i'll have to fix this properly but have other work to do so i'll have to come back to this...

PR #7338, opened 2018-06-05
pin kitchen tests to chef-13 on chef-13 branch

lamont-granquist:
stop pulling down ohai master / chef-14 current / ruby-2.5


lamont-granquist:
fixes what it set out to fix, does not fix everything yet...

PR #7339, opened 2018-06-05
Expand platforms we test in travis / cleanup cruft

tas50:
Remove the old awesome customers stuff that isn't coming back and isn't used in learn chef anymore anyways. Nuke some extra fluff in the kitchen configs and avoid inspec deprecation warnings. Remove Debian 7 testing and add Amazon2 / Ubuntu 18.04


lamont-granquist:
i'm very confused as to why the ruby-2.5.0 failures are back in this branch...

tas50:
Yep same here. I wonder if we're landing on different boxes again

kitchen-tests/.kitchen.travis.yml
```diff
@@ -2,8 +2,6 @@
 driver:
   name: dokken
   privileged: true
-  chef_image: chef/chef
-  chef_version: current
```
lamont-granquist: i wonder if something like this was bizzarely necessary magic and now we're somehow getting cached ruby-2.5.0 versions of chef-client in the base image?  (which seems backwards, but maybe these flags have the effect of disabling the precached images and then later we do the kitchen-appbundle-updater dance to install what we actually need?).

PR #7343, opened 2018-06-07
Expand development docs with branch/backport + more

tas50:
It came up today that we don't document the backporting. It seems like a totally valid thing to document if we want the community to drive the process. I sorta pulled this out of my own thoughts + some of how we do it now. Feel free to tell me if anything pops out as crazy. I also reworked some of the other stuff. 1 change per commit to make it easier to track why.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,20 +168,25 @@ After each "official" stable release we need to bump the minor version. To do th
 
 Submit a PR with the changes made by the above.
 
-## Addressing a Regression
+## Branch Structure
 
-Sometimes, regressions split through the cracks. Since new functionality is always being added and the minor version is bumped immediately after release, we can't simply roll forward. In this scenario, we'll need to perform a special regression release process. In the example that follows, the stable release with a regression is `1.10.60` while master is currently sitting at `1.11.30`. *Note:* To perform this process, you must be a Chef employee.
+We develop and ship the current release of Chef off the master branch of this repository. Our goal is that master should always be in a shipable state. Previous stable releases of Chef are developed on their own branches named by the major version (ex: chef-13 or chef-12). We do not perform direct development on these stable branches except to resolve build failures. Instead we backport fixes from our master branch to these stable branches. Stable branches receive critical bugfixes and security releases and stable Chef releases are made as necessary for security purposes.
 
-1. If the regression has not already been addressed, open a Pull Request against master with the fix.
-2. Wait until that Pull Request has been merged and `1.11.31` has passed all the necessary tests and is available in the current channel.
+## Backporting Fixes to Stable Releases
+
+If there is a critical fix you believe should be backported from master to a stable branch please follow these steps to backport your change:
+
+1. Ask in the #chef-dev channel on [Chef Community Slack](https://community-slack.chef.io/) if this is an appropriate change to backport.
 3. Inspect the Git history and find the `SHA` associated with the Merge Commit for the Pull Request above.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,20 +168,25 @@ After each "official" stable release we need to bump the minor version. To do th
...
-2. Wait until that Pull Request has been merged and `1.11.31` has passed all the necessary tests and is available in the current channel.
+## Backporting Fixes to Stable Releases
+
+If there is a critical fix you believe should be backported from master to a stable branch please follow these steps to backport your change:
+
+1. Ask in the #chef-dev channel on [Chef Community Slack](https://community-slack.chef.io/) if this is an appropriate change to backport.
 3. Inspect the Git history and find the `SHA` associated with the Merge Commit for the Pull Request above.
 4. Apply the fix for the regression via a cherry-pick:
-  1. Check out the stable release tag: `git checkout v1.10.60`
+  1. Check out the stable release tag: `git checkout chef-13`
```
coderanger: It's a branch, not a tag.

README.md
```diff
@@ -177,20 +168,25 @@ After each "official" stable release we need to bump the minor version. To do th
...
 
-1. If the regression has not already been addressed, open a Pull Request against master with the fix.
-2. Wait until that Pull Request has been merged and `1.11.31` has passed all the necessary tests and is available in the current channel.
+## Backporting Fixes to Stable Releases
+
+If there is a critical fix you believe should be backported from master to a stable branch please follow these steps to backport your change:
+
+1. Ask in the #chef-dev channel on [Chef Community Slack](https://community-slack.chef.io/) if this is an appropriate change to backport.
 3. Inspect the Git history and find the `SHA` associated with the Merge Commit for the Pull Request above.
```
coderanger: You don't usually cherry-pick merge commits I would think, but maybe I've just been doing it the slow way forever?

README.md
```diff
@@ -151,13 +142,13 @@ Chef is an amalgam of many components. These components update all the time, nec
...
-- `stable`: Periodically, Chef will pick a release to "bless" for folks who would like a slower update schedule than "every time a build passes the tests." When this happens, it is manually promoted to stable and an announcement is sent to the list. It can be reached at https://downloads.chef.io or installed using the `curl` command without specifying `-c current`. Packages in `stable` are no longer available in `current`.
+- `current`: If the packages pass all the tests on all supported OS's, it is promoted as a unit to `current`, and is available by running `curl https://www.chef.io/chef/install.sh | sudo bash -s -- -c current -P chef` or at <https://downloads.chef.io/chef/current>
+- `stable`: Periodically, Chef will pick a release to "bless" for folks who would like a slower update schedule than "every time a build passes the tests." When this happens, it is manually promoted to stable and an announcement is sent to the list. It can be reached at <https://downloads.chef.io> or installed using the `curl` command without specifying `-c current`. Packages in `stable` are no longer available in `current`.
```
coderanger: Did we ever fix the "Packages in `stable` are no longer available in `current`."? I remember we talked about it, but not sure what happened.

README.md
tas50: That's what I get for reusing the old content

README.md
tas50: I usually just do the individuals myself. I'll make this more generic

README.md
tas50: We copy from current to stable now vs. move.

PR #7349, opened 2018-06-07
Support windows_feature_powershell on Windows 2008 R2

tas50:
So this one was a few layers deep:

1) We didn't actually need to import the ServerManager module. This was a powershell 2.0 thing, but since we use ConvertTo-Json  we require Powershell 3.0 where this module is autoloaded.
2) Features are only case insensitive on Windows 8 / 2012+. That means we can't lowercase the features on these platforms. We still want to lowercase the features everywhere else because case shouldn't matter and that would be a breaking change.
3) Windows 2008 R2 doesn't have a concept of a removed feature (where the source is gone) so they don't return data the same way when you request features. They just return: Installed: true/false. This means we need to coerce the returned data from 2008 R2 to look like a later Windows release.

### DIFF
--- a/lib/chef/resource/windows_feature_dism.rb
+++ b/lib/chef/resource/windows_feature_dism.rb
@@ -30,7 +30,7 @@ class Chef
 
       property :feature_name, [Array, String],
                description: "The name of the feature/role(s) to install if it differs from the resource name.",
-               coerce: proc { |x| to_lowercase_array(x) },
+               coerce: proc { |x| to_formatted_array(x) },
                name_property: true
 
       property :source, String,
@@ -44,12 +44,18 @@ class Chef
                description: "Specifies a timeout (in seconds) for feature install.",
                default: 600
 
-      def to_lowercase_array(x)
+      # @return [Array] lowercase the array unless we're on < Windows 2012
+      def to_formatted_array(x)
         x = x.split(/\s*,\s*/) if x.is_a?(String) # split multiple forms of a comma separated list
 
-        # dism on windows < 2012 is case sensitive so only downcase when on 2012+
-        # @todo when we're really ready to remove support for Windows 2008 R2 this check can go away
-        node["platform_version"].to_f < 6.2 ? x : x.map(&:downcase)
+        # feature installs on windows < 2012 are case sensitive so only downcase when on 2012+
+        older_than_2012_or_8? ? x : x.map(&:downcase)
+      end
+
+      # a simple helper to determine if we're on a windows release pre-2012 / 8
+      # @return [Boolean] Is the system older than Windows 8 / 2012
+      def older_than_2012_or_8?
+        node["platform_version"].to_f < 6.2
       end
 
       action :install do
@@ -200,7 +206,7 @@ class Chef
           # dism on windows 2012+ isn't case sensitive so it's best to compare
           # lowercase lists so the user input doesn't need to be case sensitive
           # @todo when we're ready to remove windows 2008R2 the gating here can go away
-          feature_details.downcase! unless node["platform_version"].to_f < 6.2
+          feature_details.downcase! unless older_than_2012_or_8?
           node.override["dism_features_cache"][feature_type] << feature_details
         end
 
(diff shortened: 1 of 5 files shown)


lib/chef/resource/windows_feature_powershell.rb
```diff
@@ -54,9 +54,19 @@ class WindowsFeaturePowershell < Chef::Resource
                description: "",
                default: false
 
-      def to_lowercase_array(x)
+      # a simple helper to determine if we're on a windows release pre-2012 / 8
+      # @return [Boolean] Is the system older than Windows 8 / 2012
+      def older_than_2012_or_8?
```
stuartpreston: Arguably this could go into chef/platform/query_helpers as `Chef::Platform.older_than_2012_or_8?` so we can reuse it.

lib/chef/resource/windows_feature_dism.rb
```diff
@@ -44,12 +44,18 @@ class WindowsFeatureDism < Chef::Resource
...
-        # dism on windows < 2012 is case sensitive so only downcase when on 2012+
-        # @todo when we're really ready to remove support for Windows 2008 R2 this check can go away
-        node["platform_version"].to_f < 6.2 ? x : x.map(&:downcase)
+        # feature installs on windows < 2012 are case sensitive so only downcase when on 2012+
+        older_than_2012_or_8? ? x : x.map(&:downcase)
+      end
+
+      # a simple helper to determine if we're on a windows release pre-2012 / 8
+      # @return [Boolean] Is the system older than Windows 8 / 2012
+      def older_than_2012_or_8?
```
stuartpreston: Reuse via chef/platform/query_helpers ?

lib/chef/resource/windows_feature_powershell.rb
```diff
@@ -148,12 +158,16 @@ def raise_on_old_powershell
           raise "The windows_feature_powershell resource requires PowerShell 3.0 or later. Please install PowerShell 3.0+ before running this resource." if powershell_version < 3
         end
 
+        # The appropirate cmdlet to install a windows feature based on windows release
```
stuartpreston: s/appropirate/appropriate/g

lib/chef/resource/windows_feature_powershell.rb
```diff
@@ -148,12 +158,16 @@ def raise_on_old_powershell
...
+        # @return [String]
         def install_feature_cmdlet
-          node["platform_version"].to_f < 6.2 ? "Import-Module ServerManager; Add-WindowsFeature" : "Install-WindowsFeature"
+          older_than_2012_or_8? ? "Add-WindowsFeature" : "Install-WindowsFeature"
         end
 
+        # The appropirate cmdlet to remove a windows feature based on windows release
```
stuartpreston: s/appropirate/appropriate/g

lib/chef/resource/windows_feature_dism.rb
```diff
@@ -206,7 +201,7 @@ def add_to_feature_mash(feature_type, feature_string)
           # dism on windows 2012+ isn't case sensitive so it's best to compare
           # lowercase lists so the user input doesn't need to be case sensitive
           # @todo when we're ready to remove windows 2008R2 the gating here can go away
-          feature_details.downcase! unless older_than_2012_or_8?
+          feature_details.downcase! unless Chef::Platform.older_than_win_2012_or_8
```
stuartpreston: Should be `older_than_win_2012_or_8?`

lib/chef/resource/windows_feature_powershell.rb
```diff
@@ -264,7 +259,7 @@ def fail_if_removed
 
         # Fail unless we're on windows 8+ / 2012+ where deleting a feature is supported
         def raise_if_delete_unsupported
-          raise Chef::Exceptions::UnsupportedAction, "#{self} :delete action not support on Windows releases before Windows 8/2012. Cannot continue!" if older_than_2012_or_8?
+          raise Chef::Exceptions::UnsupportedAction, "#{self} :delete action not support on Windows releases before Windows 8/2012. Cannot continue!" if Chef::Platform.older_than_win_2012_or_8
```
stuartpreston: 1. s/support/supported

2. Should be `older_than_win_2012_or_8?`

PR #7353, opened 2018-06-11
windows_package: Fix package sensitive error

dheerajd-msys:
### Description

Package resource doesn't suppress all sensitive resource output even if suppress property is used

### Issues Resolved

https://getchef.zendesk.com/agent/#/tickets/18893

### DIFF
--- a/lib/chef/provider/package/windows/exe.rb
+++ b/lib/chef/provider/package/windows/exe.rb
@@ -55,27 +55,17 @@ class Chef
 
           def install_package
             logger.trace("#{new_resource} installing #{new_resource.installer_type} package '#{new_resource.source}'")
-            begin
-              shell_out!(
-                [
-                  "start",
-                  "\"\"",
-                  "/wait",
-                  "\"#{new_resource.source}\"",
-                  unattended_flags,
-                  expand_options(new_resource.options),
-                  "& exit %%%%ERRORLEVEL%%%%",
-                ].join(" "), timeout: new_resource.timeout, returns: new_resource.returns
-              )
-            rescue Mixlib::ShellOut::ShellCommandFailed
-              if sensitive?
-                ex = Mixlib::ShellOut::ShellCommandFailed.new("Command execution failed. STDOUT/STDERR suppressed for sensitive resource")
-                raise ex
-              else
-                raise
-              end
-            end
-            logger.info("#{new_resource} ran successfully")
+            shell_out!(
+              [
+                "start",
+                "\"\"",
+                "/wait",
+                "\"#{new_resource.source}\"",
+                unattended_flags,
+                expand_options(new_resource.options),
+                "& exit %%%%ERRORLEVEL%%%%",
+              ].join(" "), timeout: new_resource.timeout, returns: new_resource.returns, sensitive: new_resource.sensitive
+            )
           end
 
           def remove_package
@@ -88,10 +78,6 @@ class Chef
           end
 
           private
-
-          def sensitive?
-           !!new_resource.sensitive
-          end
 
           def uninstall_command(uninstall_string)
             uninstall_string = "\"#{uninstall_string}\"" if ::File.exist?(uninstall_string)
--- a/lib/chef/provider/package/windows/exe.rb
+++ b/lib/chef/provider/package/windows/exe.rb
@@ -55,17 +55,27 @@ class Chef
 
           def install_package
             logger.trace("#{new_resource} installing #{new_resource.installer_type} package '#{new_resource.source}'")
-            shell_out!(
-              [
-                "start",
-                "\"\"",
-                "/wait",
-                "\"#{new_resource.source}\"",
-                unattended_flags,
-                expand_options(new_resource.options),
-                "& exit %%%%ERRORLEVEL%%%%",
-              ].join(" "), timeout: new_resource.timeout, returns: new_resource.returns
-            )
+            begin
+              shell_out!(
+                [
+                  "start",
+                  "\"\"",
+                  "/wait",
+                  "\"#{new_resource.source}\"",
+                  unattended_flags,
+                  expand_options(new_resource.options),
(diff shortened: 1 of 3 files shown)


lib/chef/resource/windows_package.rb
```diff
@@ -42,7 +42,10 @@ def initialize(name, run_context = nil)
       end
 
       # windows can't take array options yet
-      property :options, String
+      property :options, String, sensitive: true
+
+      # lazy used to set default value of sensitive to true if options is set
+      property :sensitive, [ TrueClass, FalseClass ], default: lazy { |r| r.options ? true : false }
```
lamont-granquist: Neither of these options should really be changed, the `options` string is not generally sensitive -- options like "/SkipRules=RebootRequiredCheck /Action=install" don't require sensitive.  Its only if the user supplies passwords in the resource that it should be marked sensitive, and i think its upon the user to determine that the options are sensitive and to mark it so.  The general case of automatically determining that a password has been supplied to a property and marking it magically sensitive for the user has too many positive and negative edge cases and I don't think we should open that can of worms.

lib/chef/provider/package/windows/exe.rb
```diff
@@ -55,17 +55,27 @@ def package_version
...
+                  "& exit %%%%ERRORLEVEL%%%%",
+                ].join(" "), timeout: new_resource.timeout, returns: new_resource.returns
+              )
+            rescue Mixlib::ShellOut::ShellCommandFailed
+              if sensitive?
+                ex = Mixlib::ShellOut::ShellCommandFailed.new("Command execution failed. STDOUT/STDERR suppressed for sensitive resource")
+                raise ex
+              else
+                raise
+              end
```
lamont-granquist: this is actually buggy and will still throw as part of the exception call chain that helpfully gets logged in newer rubies...

https://github.com/chef/chef/blob/master/lib/chef/provider/execute.rb#L59-L70

that is the correct code to use, and the fact that we're copypasta'ing this around suggests to me that this needs to get included in the Chef::Resource class directly so that any resource with sensitive set suppresses ShellCommandFailed exceptions correctly...

coderanger: Or it should get moved into mixlib-shellout as a feature there.

lib/chef/resource/windows_package.rb
dheerajd-msys: Even I thought the same but I couldn't find a way to suppress the sensitive data provided in `options`  and wanted review and suggestions so made it sensitive. Is there a way we can handle such cases?

lamont-granquist: It has to be manual on the part of the user to mark the whole resource as sensitive.

lib/chef/provider/package/windows/exe.rb
lamont-granquist: Yep sensitive flag in mixlib-shellout itself would be good as well, then the initial exception with the sensitive info never gets thrown, so we don't have to worry about the exception chaining case.

dheerajd-msys:
Added `sensitive` as feature in `mixlib-shellout` in https://github.com/chef/mixlib-shellout/pull/156
**Recipe**
```
package "SQL install" do
action :install
sensitive true
source "C:\\Msys-830\\SQLEXPR32_x86_ENU\\SETUP.exe"
options "/Q /IACCEPTSQLSERVERLICENSETERMS /SkipRules=RebootRequiredCheck /Action=install /SAPWD=\"azure@123456\" /SQLSVCPASSWORD=\"azure@123456\" /AGTSVCPASSWORD=\"azure@123456\" /ConfigurationFile=\"C:\\Msys-830\\SQLEXPR32_x86_ENU\\SETUP.EXE.CONFIG\""
timeout 3600
installer_type :custom
returns [0, 3010]
end
```
**Output**
{code}
  * windows_package[SQL install] action install

    Error executing action `install` on resource 'windows_package[SQL install]'

    Expected process to exit with [0, 3010], but received '-2067922413'
    Command execution failed. STDOUT/STDERR suppressed for sensitive resource

    Resource Declaration:
    ---------------------
    suppressed sensitive resource output

    Compiled Resource:
    ------------------
    suppressed sensitive resource output

    System Info:
    ------------
    chef_version=14.2.2
    platform=windows
    platform_version=6.3.9600
    ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x64-mingw32]
    program_name=C:/Msys-830/chef/.bundle/ruby/2.5.0/bin/chef-apply
    executable=C:/Msys-830/chef/.bundle/ruby/2.5.0/bin/chef-apply

[2018-06-13T14:32:04+00:00] FATAL: Stacktrace dumped to C:/chef/cache/chef-stacktrace.out
[2018-06-13T14:32:04+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2018-06-13T14:32:04+00:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: windows_package[SQL install] ((chef-apply cookb
ook)::(chef-apply recipe) line 1) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0,
3010], but received '-2067922413'
Command execution failed. STDOUT/STDERR suppressed for sensitive resource
{code}

dheerajd-msys:
@thommay asked me to resolve the style issues here https://github.com/chef/mixlib-shellout/pull/156#discussion_r195416049, So will be fixing those.

lib/chef/provider/package/windows/exe.rb
btm: @lamont-granquist this got moved to mixlib-shellout. Can you re-review since you've got an ❌ on the PR?

dheerajd-msys:
@lamont-granquist Could you please review it again and merge the PR.

Thanks

coderanger:
@dheerajd-msys This can't get merged until a new release of mixlib-shellout goes out so we can update the dependency.

PR #7355, opened 2018-06-12
Update to the latest inspec and liblzma

tas50:
Pull in the latest omnibus-software that bumps liblzma with a large pile of bugfixes.


lamont-granquist:
think you need to bump versions in `omnibus_overrides.rb` though?

PR #7357, opened 2018-06-12
Updating git resource to fail when checking out to non-empty directory

### LINKED ISSUE #7347: git resource silently fails to sync or checkout when directory already exists
## Description
The `git` resource's `:sync` action does nothing and does not fail if the target directory already exists and is not a git repo.

## Chef Version
14.1.2

## Platform Version
I've reproduced this in Ubuntu & macOS. After looking at the implementation, this should reproduce anywhere the git resource is usable.

## Replication Case
I generated a cookbook with a default recipe that looked like this

```ruby
git 'test-repo' do
  repository 'https://github.com/brownmike/test-repo.git'
  revision 'master'
  action :sync
end
```

I ran `kitchen converge` then `kitchen login` to verify `/test-repo` matched what was in my remote. I then `rm -rf /test-repo/.git` and added a new file to the repo on the remote. On the next `kitchen converge` the output implied nothing was wrong.

## Client Output

```
       Starting Chef Client, version 14.1.12
       resolving cookbooks for run list: ["gitissues::default"]
       Synchronizing Cookbooks:
         - gitissues (0.1.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       Converging 1 resources
       Recipe: gitissues::default
         * git[test-repo] action sync (up to date)

       Running handlers:
       Running handlers complete
       Chef Client finished, 0/1 resources updated in 01 seconds
       Downloading files from <default-ubuntu-1604>
       Finished converging <default-ubuntu-1604> (0m4.62s).
```

## Problematic Code
From `lib/chef/provider/git.rb`

In the `action_sync` method it first checks if we're dealing with an existing git repo. If we're not, we get redirected to the `action_checkout` method.

```ruby
def action_sync
  if existing_git_clone?
    logger.trace "#{new_resource} current revision: #{current_resource.revision} target revision: #{target_revision}"
    unless current_revision_matches_target_revision?
      fetch_updates
      enable_submodules
      logger.info "#{new_resource} updated to revision #{target_revision}"
    end
    add_remotes
  else
    action_checkout
  end
end
```

In the `action_checkout` method if the target directory is not empty, we log a line about it but otherwise continue on normally.

```ruby
def action_checkout
  if target_dir_non_existent_or_empty?
    clone
    if new_resource.enable_checkout
      checkout
    end
    enable_submodules
    add_remotes
  else
    logger.trace "#{new_resource} checkout destination #{cwd} already exists or is a non-empty directory"
  end
end
```

## Potential Solutions

1. Explicitly fail when the target directory already exists
2. Overwrite pre-existing target directories and update the documentation to indicate this

brownmike:
### Description

The current behavior of the git resource is that it will silently fail to checkout if the destination directory is not empty. This change modifies the behavior in that situation to raise an explicit error.

### Issues Resolved

#7347 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -50,6 +50,12 @@ class Chef
             "Cannot clone #{new_resource} to #{cwd}, the enclosing directory #{dirname} does not exist")
         end
 
+        requirements.assert(:checkout) do |a|
+          a.assertion { target_dir_non_existent_or_empty? }
+          a.failure_message(Chef::Exceptions::DestinationAlreadyExists,
+            "Cannot clone #{new_resource} into #{cwd}, the destination is not a directory or not empty.")
+        end
+
         requirements.assert(:all_actions) do |a|
           a.assertion { !(new_resource.revision =~ /^origin\//) }
           a.failure_message Chef::Exceptions::InvalidRemoteGitReference,
@@ -72,20 +78,16 @@ class Chef
       end
 
       def action_checkout
-        if target_dir_non_existent_or_empty?
-          clone
-          if new_resource.enable_checkout
-            checkout
-          end
-          enable_submodules
-          add_remotes
-        else
-          logger.trace "#{new_resource} checkout destination #{cwd} already exists or is a non-empty directory"
+        clone
+        if new_resource.enable_checkout
+          checkout
         end
+        enable_submodules
+        add_remotes
       end
 
       def action_export
-        action_checkout
+        run_action :checkout
         converge_by("complete the export by removing #{cwd}.git after checkout") do
           FileUtils.rm_rf(::File.join(cwd, ".git"))
         end
@@ -101,7 +103,7 @@ class Chef
           end
           add_remotes
         else
-          action_checkout
+          run_action :checkout
         end
       end
 
@@ -132,7 +134,7 @@ class Chef
 
       def find_current_revision
         logger.trace("#{new_resource} finding current git revision")
-        if ::File.exist?(::File.join(cwd, ".git"))
+        if existing_git_clone?
           # 128 is returned when we're not in a git repo. this is fine
           result = git("rev-parse", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip
         end
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -141,6 +141,7 @@ class Chef
     class IllegalChecksumRevert < RuntimeError; end
     class CookbookVersionNameMismatch < ArgumentError; end
     class MissingParentDirectory < RuntimeError; end
+    class DestinationAlreadyExists < RuntimeError; end
     class UnresolvableGitReference < RuntimeError; end
     class InvalidRemoteGitReference < RuntimeError; end
     class InvalidEnvironmentRunListSpecification < ArgumentError; end
--- a/spec/unit/provider/git_spec.rb
+++ b/spec/unit/provider/git_spec.rb
@@ -19,6 +19,10 @@
 require "spec_helper"
 describe Chef::Provider::Git do
 
+  let(:parent_dir) { "/my/deploy" }
+  let(:repo_dir) { ::File.join parent_dir, "dir" }
+  let(:repo_git_dir) { ::File.join repo_dir, ".git" }
+
   before(:each) do
     allow(STDOUT).to receive(:tty?).and_return(true)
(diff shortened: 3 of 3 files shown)


lib/chef/provider/git.rb
```diff
@@ -50,6 +50,12 @@ def define_resource_requirements
             "Cannot clone #{new_resource} to #{cwd}, the enclosing directory #{dirname} does not exist")
         end
 
+        requirements.assert(:checkout) do |a|
+          a.assertion { target_dir_non_existent_or_empty? }
```
tas50: @lamont-granquist Do we want to take why-run into account here?

brownmike: Good question. I’ve heard rumors that why run is inherently flawed and possibly going away anyway in the future? I didn’t really think about it TBH.

thommay: Using "why-run" to simulate chef runs is inherently flawed, yes. However, the chef client uses why-run internally to make decisions about when to run `:before` notifications, among other things.

brownmike: Looping back on this today. What's the plan here? Should we raise an assertion during why run or not?

lib/chef/provider/git.rb
```diff
@@ -72,20 +78,16 @@ def define_resource_requirements
...
+        if new_resource.enable_checkout
+          checkout
         end
+        enable_submodules
+        add_remotes
       end
 
       def action_export
-        action_checkout
+        run_action :checkout
```
lamont-granquist: this is probably not a good idea, it re-runs define_resource_requirements and everything else, which is difficult to reason about.

lib/chef/provider/git.rb
lamont-granquist: @brownmike why-run is used in :before notifications and now can't really ever go away

lamont-granquist:
Is this idempotent?  because if you've check it out already, what prevents it from just failing on your existing checkout on the second run through?  Not sure if i'm just missing something obvious and need more coffee though...

lamont-granquist:
(wrong button, definitely need more coffee)

lib/chef/provider/git.rb
brownmike: Shouldn't that be exactly what happens, though? This action is an alias for checking out + removing the `.git/` directory when we're done. It seems we should define all the resource requirements for checking out if that's the thing we're going to be doing. Maybe I'm missing something here?

lib/chef/provider/git.rb
brownmike: > why-run is used in :before notifications and now can't really ever go away

I'm not sure what this means in the context of this review, though. Should I add an assertion during why run? Or just a message like some of the other requirements. i.e. the parent dir must exist requirement just has this:

```ruby
a.whyrun("Directory #{dirname} does not exist, this run will fail unless it has been previously created. Assuming it would have been created.")
```

brownmike:
> Is this idempotent? because if you've check it out already, what prevents it from just failing on your existing checkout on the second run through? Not sure if i'm just missing something obvious and need more coffee though...

You're right, this will fail on a second run since the target directory is now non-empty.

This raises a curious question about the checkout action: what should we do to guard against this situation? With the existing logic, checking out twice in a row is idempotent *because* checking out to a directory that already exists falsely succeeds (it fails silently by logging a message about the destination already existing, but the run continues). This, in turn, causes syncing to falsely succeed if the target directory's `.git/` folder was removed (because if it's not a git repo, the sync action delegates to the checkout action).

So should we...
1. Overwrite existing target directories on checkout?
2. Create some heuristic to determine if the target directory is indeed the repo we want already?
3. Preserve the existing behavior and close the bug as won't fix?
4. Some magical solution behind door number four that I haven't thought of?

lib/chef/provider/git.rb
lamont-granquist: Yeah, that's more or less what you need although i think if its non-existent or empty that's fine, so the message needs to be along the lines of "Directory is not a git repository, assuming someone will have cleaned that up" or "Directory contains some kind of garbage, assuming someone cleaned that up..."

The way the whyrun message is pretty oddly worded suggests that it may be better to have the git resource own the target directory and blow it away and assert control over the entire state (which is properly declarative).

lamont-granquist:
Yeah you're going to need to go down the road of parsing `git status --branch --porcelain` I think

lamont-granquist:
closing old PR due to inactivity.

PR #7358, opened 2018-06-12
Add more property descriptions to resources

tas50:
Slowly we'll get them all.

### DIFF
--- a/lib/chef/resource/ifconfig.rb
+++ b/lib/chef/resource/ifconfig.rb
@@ -35,23 +35,63 @@ class Chef
       default_action :add
       allowed_actions :add, :delete, :enable, :disable
 
-      property :target, String, name_property: true
-      property :hwaddr, String
-      property :mask, String
-      property :family, String, default: "inet"
-      property :inet_addr, String
-      property :bcast, String
-      property :mtu, String
-      property :metric, String
-      property :device, String, identity: true
-      property :onboot, String
-      property :network, String
-      property :bootproto, String
-      property :onparent, String
-      property :ethtool_opts, String
-      property :bonding_opts, String
-      property :master, String
-      property :slave, String
+      property :target, String,
+               name_property: true,
+               description: "The IP address that is to be assigned to the network interface. If not specified we'll use the resource's name."
+
+      property :hwaddr, String,
+               description: "The hardware address for the network interface."
+
+      property :mask, String,
+               description: "The decimal representation of the network mask. For example: 255.255.255.0."
+
+      property :family, String,
+               default: "inet", introduced: "14.0",
+               description: "Networking family option for Debian-based systems; for example: inet or inet6."
+
+      property :inet_addr, String,
+               description: "The Internet host address for the network interface."
+
+      property :bcast, String,
+               description: "The broadcast address for a network interface. On some platforms this property is not set using ifconfig, but instead is added to the startup configuration file for the network interface."
+
+      property :mtu, String,
+               description: "The maximum transmission unit (MTU) for the network interface."
+
+      property :metric, String,
+               description: "The routing metric for the interface."
+
+      property :device, String,
+               identity: true,
+               description: "The network interface to be configured."
+
+      property :onboot, String,
+               description: "Bring up the network interface on boot."
+
+      property :network, String,
+               description: "The address for the network interface."
+
+      property :bootproto, String,
+               description: "The boot protocol used by a network interface."
+
+      property :onparent, String,
+               description: "Bring up the network interface when its parent interface is brought up."
+
+      property :ethtool_opts, String,
+               introduced: "13.4",
+               description: "Options to be passed to ethtool(8). For example: -A eth0 autoneg off rx off tx off"
+
+      property :bonding_opts, String,
+               introduced: "13.4",
+               description: "Bonding options to pass via BONDING_OPTS on RHEL and CentOS. For example: mode=active-backup miimon=100"
+
+      property :master, String,
+               introduced: "13.4",
+               description: "Specifies the channel bonding interface to which the Ethernet interface is linked."
+
+      property :slave, String,
+               introduced: "13.4",
+               description: "When set to yes, this device is controlled by the channel bonding interface that is specified via the master property."
     end
   end
 end
--- a/lib/chef/resource/ifconfig.rb
+++ b/lib/chef/resource/ifconfig.rb
@@ -47,7 +47,7 @@ class Chef
 
       property :family, String,
                default: "inet", introduced: "14.0",
-               description: "Networking family option for Debian-based systems; for example: inet or inet6."
+               description: "Networking family option for Debian-based systems. For example: inet or inet6."
 
       property :inet_addr, String,
                description: "The Internet host address for the network interface."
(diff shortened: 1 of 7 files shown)


lib/chef/resource/ifconfig.rb
```diff
@@ -35,23 +35,63 @@ class Ifconfig < Chef::Resource
...
+
+      property :hwaddr, String,
+               description: "The hardware address for the network interface."
+
+      property :mask, String,
+               description: "The decimal representation of the network mask. For example: 255.255.255.0."
+
+      property :family, String,
+               default: "inet", introduced: "14.0",
+               description: "Networking family option for Debian-based systems; for example: inet or inet6."
```
coderanger: Can I be super pedantic that you used two different styles of `for example` here and the one above? O:-)

lib/chef/resource/ifconfig.rb
```diff
@@ -35,23 +35,63 @@ class Ifconfig < Chef::Resource
...
+               description: "Bring up the network interface on boot."
+
+      property :network, String,
+               description: "The address for the network interface."
+
+      property :bootproto, String,
+               description: "The boot protocol used by a network interface."
+
+      property :onparent, String,
+               description: "Bring up the network interface when its parent interface is brought up."
```
coderanger: Since this isn't a boolean, what kind of value would go in here?

lib/chef/resource/launchd.rb
```diff
@@ -30,16 +30,39 @@ class Launchd < Chef::Resource
...
-      property :mode, [String, Integer]
-      property :owner, [String, Integer]
-      property :path, String
-      property :source, String
-      property :session_type, String
+      property :label, String,
+               identity: true, name_property: true,
+               description: "The unique identifier for the job."
+
+      property :backup, [Integer, FalseClass],
```
coderanger: This probably needs a default value.

lib/chef/resource/launchd.rb
```diff
@@ -30,16 +30,39 @@ class Launchd < Chef::Resource
...
+
+      property :group, [String, Integer],
+               description: "When launchd is run as the root user, the group to run the job as. If the username property is specified and this property is not, this value is set to the default group for the user."
+
+      property :plist_hash, Hash,
+               introduced: "12.19",
+               description: "A Hash of key value pairs used to create the launchd property list."
+
+      property :mode, [String, Integer],
+               description: "A quoted 3-5 character string that defines the octal mode. For example: '755', '0755', or 00755."
```
coderanger: It appears this takes a integer too, and `quoted` is odd (makes it sound like you need to do `"'755'"` or something.

lib/chef/resource/launchd.rb
```diff
@@ -50,96 +73,180 @@ class Launchd < Chef::Resource
...
-        types = %w{daemon agent}
-
-        unless types.include?(type)
-          error_msg = "type must be daemon or agent"
-          raise Chef::Exceptions::ValidationFailed, error_msg
-        end
-        type
-      }
+      property :start_calendar_interval, [Hash, Array],
+               description: "A Hash (similar to crontab) that defines the calendar frequency at which a job is started.",
```
coderanger: Can also take an array.

lib/chef/resource/launchd.rb
```diff
@@ -50,96 +73,180 @@ class Launchd < Chef::Resource
...
+               description: "An array of non-empty directories which, if any are modified, will cause a job to be started."
+
+      property :root_directory, String,
+               description: "chroot to this directory, and then run the job."
+
+      property :run_at_load, [ TrueClass, FalseClass ],
+               description: "Launch a job once (at the time it is loaded)."
+
+      property :sockets, Hash,
+               description: "A Hash of on-demand sockets that notify launchd when a job should be run."
```
coderanger: If we have one, an example of the format would be good.

lib/chef/resource/launchd.rb
```diff
@@ -50,96 +73,180 @@ class Launchd < Chef::Resource
...
+               description: "Start a job every time a file system is mounted."
+
+      property :throttle_interval, Integer,
+               description: "The frequency (in seconds) at which jobs are allowed to spawn."
+
+      property :time_out, Integer,
+               description: "The amount of time (in seconds) a job may be idle before it times out. If no value is specified, the default timeout value for launchd will be used."
+
+      property :umask, Integer,
+               description: "A decimal value to pass to umask before running a job."
```
coderanger: Doesn't need to be specifically decimal I would think.

lib/chef/resource/launchd.rb
```diff
@@ -50,96 +73,180 @@ class Launchd < Chef::Resource
...
+               description: "When launchd is run as the root user, the user to run the job as."
+
+      property :wait_for_debugger, [ TrueClass, FalseClass ],
+               description: "Specify if launchd has a job wait for a debugger to attach before executing code."
+
+      property :watch_paths, Array,
+               description: "An array of paths which, if any are modified, will cause a job to be started."
+
+      property :working_directory, String,
+               description: "chdir to this directory, and then run the job."
```
coderanger: Missing initial capital letter.

lib/chef/resource/log.rb
```diff
@@ -37,8 +37,13 @@ class Log < Chef::Resource
                   " entry that is not built into the resource collection, use Chef::Log instead"\
                   " of the log resource.)"
 
-      property :message, String, name_property: true, identity: true
-      property :level, Symbol, equal_to: [ :debug, :info, :warn, :error, :fatal ], default: :info
+      property :message, String,
+               name_property: true, identity: true,
+               description: "The message to be added to a log file. If not specified we'll use the resource's name instead."
```
coderanger: We should probably come up with a standard for how to document name_props re: the "If not specified ..." bit. But this is fine for now certainly :)

lib/chef/resource/log.rb
```diff
@@ -37,8 +37,13 @@ class Log < Chef::Resource
...
+
+      property :level, Symbol,
+               equal_to: [ :debug, :info, :warn, :error, :fatal ], default: :info,
+               description: "The level of logging that is to be displayed by the Chef Client."
```
coderanger: This makes it sound like its setting `Chef::Log.level`. Maybe "The logging level to display this message at."?

lib/chef/resource/mdadm.rb
```diff
@@ -32,14 +32,35 @@ class Mdadm < Chef::Resource
...
-      property :bitmap, String
-      property :raid_device, String, identity: true, name_property: true
-      property :layout, String
+      property :chunk, Integer,
+               default: 16,
+               description: "The chunk size. This property should not be used for a RAID 1 mirrored pair (i.e. when the level property is set to 1)."
+
+      property :devices, Array,
+               default: lazy { [] },
+               description: "A comma-separated list of devices to be part of a RAID array"
```
coderanger: This makes it sound like it takes a string, rather than an array.

lib/chef/resource/mdadm.rb
```diff
@@ -32,14 +32,35 @@ class Mdadm < Chef::Resource
...
+               default: false,
+               description: "Indicates whether the RAID array exists."
+
+      property :level, Integer,
+               default: 1,
+               description: "The RAID level."
+
+      property :metadata, String,
+               default: "0.90",
+               description: "The superblock type for RAID metadata"
```
coderanger: Missing trailing period.

tas50:
So the wording is entirely copy / paste from the docs site except for some basic name property explanation. So it came over warts and all.

PR #7369, opened 2018-06-13
removing mwrock from client maintainers

mwrock:
I'm trying to reign in my GH notifications. I'm still very interested in chef but deleting the emails (yes, I know, thats how I consume GH notifications) consumes minutes of every day and I no longer have cycles to act on these.


tas50:
You'll want to regenerate the markdown too

mwrock:
oh ok. Done @tas50

PR #7370, opened 2018-06-13
windows_feature_dism: Fix errors when specifying the source

tas50:
Make sure we dupe the immutable array of the node attribute so we can
concat the additional array values into it later on.


mike1372:
Great thanks guys.

PR #7372, opened 2018-06-14
Unification of shell_out APIs

lamont-granquist:
shell_out_compact / shell_out_with_timeout / shell_out_compact_timeout are now all collapsed into simply shell_out

the `shell_out_compacted` API is a private method which can be used to setup expectations on in rspec post-flat/compaction -- so meaningless changes to the shell_out won't cause churn in the specs.

formal deprecation of the APIs comes next.


tas50:
Probably wanna add this to the release notes

lamont-granquist:
it'll make more sense to do that with the deprecation PR

PR #7375, opened 2018-06-15
Silence deprecation warnings

coderanger:
Also includes a minor overhaul of the deprecation system because it was kind of all over the place.

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -24,275 +24,189 @@ class Chef
     class << self
       include Chef::Mixin::ConvertToClassName
 
-      def create(type, message = nil, location = nil)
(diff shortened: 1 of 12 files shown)


lib/chef/deprecated.rb
```diff
@@ -24,276 +24,196 @@ class Deprecated
...
 
     class InternalApi < Base
-      def id
-        0
-      end
-
-      def target
-        "internal_api.html"
-      end
+      target 0
```
thommay: I think we need to preserve the page links for all of these, don't we?

coderanger: They are filled in by the default for all except the one non-standard case.

PR #7376, opened 2018-06-15
Don't crash on non-existent credentials profile

mivok:
### Description

If a credentials file exists, but doesn't contain the profile in
CHEF_PROFILE (or a default profile is CHEF_PROFILE is unset and there is
no context file), then an "Unable to parse Credentials file" error is
shown.

This change simply detects if there isn't a matching profile in the
credentials file and returns without trying to load the credentials any
further (as is the case when a credentials file isn't present).

### Check List

- [ ] All tests pass

### DIFF
--- a/chef-config/lib/chef-config/mixin/credentials.rb
+++ b/chef-config/lib/chef-config/mixin/credentials.rb
@@ -43,6 +43,12 @@ module ChefConfig
                   end
 
         config = Tomlrb.load_file(credentials_file)
+
+        if config[profile].nil?
+          return if profile == 'default'
+          raise ChefConfig::ConfigurationError, "Profile #{profile} doesn't exist. Please add it to #{credentials_file}."
+        end
+
         apply_credentials(config[profile], profile)
       rescue ChefConfig::ConfigurationError
         raise


thommay:
I'm not sure about this. It feels like if you've specifically requested a profile we shouldn't use a different one. Agree that if there's no default profile and you've not specified one then just doing the normal thing is right.

mivok:
We're not using a different profile, we're just not using any profile at all. This is identical behavior to if the credentials file doesn't exist, but you have a profile name set in CHEF_PROFILE.

Here's the reproduction case, and the thing I want to avoid:

* Create a blank `~/.chef/credentials` (or a `~/.chef/credentials` file with a single profile in called myprofile)
* Make sure there isn't a `~/.chef/context` file, and CHEF_PROFILE isn't set
* Run `chef install` on chefdk 3.0.36 (note: `chef install` doesn't connect to a chef server, but you still get the error. This will probably also happen with knife, but I hit the error with the chef command.)
* Get:
  ```
  ERROR: Unable to parse Credentials file: /Users/mark/.chef/credentials
  undefined method `key?' for nil:NilClass
  ```
* (optional) - spend 5 minutes trying to find a syntax error in your credentials file
* Next, set CHEF_PROFILE to `foobar`, and get the same error (because foobar isn't in the credentials file).

Now, remove ~/.chef/credentials (or move it aside) and repeat the above. You don't get an error, and it doesn't try to load a profile that doesn't exist.

coderanger:
What we probably need is a better error message, rather than silencing the error.

coderanger:
(and maybe some better handling specifically for the default profile not existing)

mivok:
Another way of looking at it is that this is similar behavior to adding `[default]` (or your profile name) to ~/.chef/credentials, but otherwise leaving the profile empty. This is actually my current workaround to get `chef install` working without having to set my profile each time.

I disagree that this is merely 'silencing the error'. It's replicating the behavior of an almost identical situation where the credentials file doesn't exist (implying that the profile inside the credentials file doesn't exist), in which case we fall back to existing credential loading methods.

I do agree that we should not be implying that there is a syntax error in the credentials file if you don't have a default profile.

It sounds like we all agree though that, in the case of the default profile not being present, we should fall back to the 'normal' way of doing things. Given that, would you both be ok with something like `return if profile == 'default' and config[profile].nil?`.

mivok:
I've modified this pull request to only return if the missing profile is the default one. If the missing profile is any other profile, it will print an error message saying the profile doesn't exist:

```
$ echo $CHEF_PROFILE

$ chef install
Error: Policyfile not found at path /Users/mark/Policyfile.rb
$ export CHEF_PROFILE=foobar
$ chef install
ERROR: Profile foobar doesn't exist. Please add it to /Users/mark/.chef/credentials.
```

mivok:
bump...

do the modifications look ok?

coderanger:
Yes but you need to fix your quotes before we can merge.

coderanger:
Fixed via #7455

mivok:
Thanks. Today I learned that chefstyle doesn't have the horrible varying quote style depending on whether you need interpolation or not, although interestingly cookstyle _does_ have that rule.

lamont-granquist:
That is a large part of why chefstyle and cookstyle both exist.  Chefstyle in general has lots of tweaks to make rubocop less annoying to people who just want to get code written.

PR #7379, opened 2018-06-18
Fixing array args in some unix providers

lamont-granquist:
- apt_repository
- apt_update
- cron/unix
- ifconfig
- mdadm
- mount (aix, solaris, linux)
- osx_profile
- route
- systemd_unit

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -112,7 +112,7 @@ class Chef
       # run the specified command and extract the fingerprints from the output
       # accepts a command so it can be used to extract both the current key's fingerprints
       # and the fingerprint of the new key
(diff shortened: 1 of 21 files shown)


lamont-granquist:
some stray heredoc chefstyle fixes also got pulled in when i rebase'd.

lib/chef/provider/apt_repository.rb
```diff
@@ -113,8 +115,8 @@ def is_key_id?(id)
       # @param [String] cmd the command to run
       #
       # @return [Array] an array of fingerprints
-      def extract_fingerprints_from_cmd(cmd)
```
tas50: isn't the yard type wrong now?

lamont-granquist: fixed

PR #7380, opened 2018-06-19
Switch powershell_exec mixin to use FFI instead of COM

stuartpreston:
### Description

This change incorporates the latest iteration of PowerShell improvements on the Windows platform, specifically we will now use FFI to invoke PowerShell via a C++ wrapper that exports a C definition and has a reference to the .NET assembly that invokes an instance of the PowerShell engine (System.Management.Automation class).  

Because Ruby internally uses UTF-8 and .NET/PowerShell uses null-terminated UTF-16LE strings, some directed marshalling is required across all components.  The existing pointer->widestring support in Chef reads from the stream correctly but did not detect the end of the bytestream that was returned, hence the similar looking method `read_utf16string` being added to `win32/unicode.rb` which offers a simplified approach to converting UTF16-LE to something readable in Ruby.

### Check List

- [ ] All tests pass


Gemfile
```diff
@@ -65,3 +65,18 @@ instance_eval(ENV["GEMFILE_MOD"]) if ENV["GEMFILE_MOD"]
...
+# We copy (and overwrite) these files every time "bundle <exec|install>" is
+# executed, just in case they have changed.
+if RUBY_PLATFORM =~ /mswin|mingw|windows/
+  instance_eval do
+    ruby_exe_dir = RbConfig::CONFIG["bindir"]
+    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", Dir.pwd) + "/*.dll")
+    FileUtils.cp_r assemblies, ruby_exe_dir, verbose: true unless ENV['_BUNDLER_WINDOWS_DLLS_COPIED']
+    ENV['_BUNDLER_WINDOWS_DLLS_COPIED'] = "1"
+  end
+end
```
lamont-granquist: this seems a little odd if i'm reading it correctly...

but the problem is that we need these dlls in the PATH or in pwd in order to have them link correctly?  and since we might be in some other directory we don't want to copy to pwd, so we copy to `RbConfig::CONFIG["bindir"]` into the ruby?

it might be better to use __FILE__ to construct a full path to the DLLs and use that path in `FFI.ffi_lib` and maybe `Fiddle.dlopen` the dependent DLLs first to load them into the process before the ffi_lib call?

stuartpreston: Yes, we could load the wrapper from any location inside the distribution, but the .NET assemblies must be loaded from the location where ruby.exe is running (or we need to deploy the assemblies into the .NET Global Assembly Cache)

lamont-granquist: right, windoze is gonna windoze.... carry on then..

PR #7384, opened 2018-06-20
Update acceptance deps to the latest

tas50:
Silence a security warning from Github even though we're not using these


lamont-granquist:
`git rm acceptance` instead?

tas50:
I was pretty tempted to do that, but I think we should wait until we have an idea what to replace it with a

lamont-granquist:
if anyone does that, they can always go back and find that info in git history...

tas50:
PR incoming then

PR #7385, opened 2018-06-20
Rename the kitchen base test suite to end-to-end

tas50:
It's our end-to-end test according to the readme. This just aligns the recipe/suite names with that so it's a bit more obvious what we're doing here. This also removes sudo installs since I moved that all to dokken-images now.


tas50:
Amazon Linux 2 failure is going to continue till we fix the installer/ohai issues

PR #7388, opened 2018-06-22
Add chocolatey_config and chocolatey_source resources

tas50:
Allow a user to fully manage chocolatey and get it setup for airgapped environments using chef out of the box.


lib/chef/resource/chocolatey_source.rb
```diff
@@ -0,0 +1,88 @@
...
+        bypass_proxy element["bypassProxy"] == "true"
+        priority element["priority"].to_i
+      end
+
+      # @param [String] id the source name
+      # @return [REXML::Attributes] finds the source element with the
+      def fetch_source_element(id)
+        require "rexml/document"
+
+        config_file = 'C:\ProgramData\chocolatey\config\chocolatey.config'
```
stuartpreston: Consider `"#{ENV['ALLUSERSPROFILE']}\\chocolatey\\config\\chocolatey.config"`

lib/chef/resource/chocolatey_source.rb
```diff
@@ -0,0 +1,88 @@
...
+            shell_out!(choco_cmd("remove"))
+          end
+        end
+      end
+
+      action_class do
+        # @param [String] action the name of the action to perform
+        # @return [String] the choco source command string
+        def choco_cmd(action)
+          cmd = "C:\\ProgramData\\chocolatey\\bin\\choco source #{action} -n \"#{new_resource.source_name}\""
```
stuartpreston: Best to use ```cmd = "#{ENV['ALLUSERSPROFILE']}\\chocolatey\\bin\\choco source ..."``` here

stuartpreston:
Also one thing I noticed, in `chocolatey_config` if you set `value ""` if there was a previous value in the configuration this causes choco.exe set to fail because a value is not provided.  It may not be clear to the end user that you need the `:unset` action.

Can we call `unset` here if the value is empty but had a prior value? 

repro:
pass1 (succeeds)
```
chocolatey_config "cacheLocation" do
  value "c:\\"
end
```
pass2 (fails)
```
chocolatey_config "cacheLocation" do
  value ""
end
```

tas50:
I think we could potentially fail with a better message when you pass an empty string. Let me try that in the cookbook.

PR #7390, opened 2018-06-24
knife config and a bunch of UX improvements

coderanger:
This adds a `knife config` command to display the current config data in various ways, mostly for debugging but also potentially useful for scripting (via `--format=json` or `--raw`). It also improves the default configuration behavior for `knife`/`chef`, `node_name` defaults to the system username, and `client_key` and `validation_key` default to some of the usual paths for them. It also adds a `$CHEF_HOME` env var and improved support for `$KNIFE_HOME`.

All put together on the config side, this means that `knife` and `chef` are totally usable with only a `credentials` file like:

```toml
[default]
chef_server_url = "https://..."
```


coderanger:
So one potential question: what I have as `knife config`, do we want to punt that down a level to something like `knife config get` so we can use `knife config` as a namespace for other commands (`knife config set-profile`, `knife config initialize`)? Given that the `knife` namespace is kind of a wreck, maybe it's not worth fixing and we should just roll this kind of functionality directly into `chef config *`?

coderanger:
Oh and also how do people feel about the control file being `~/.chef/context` (currently) vs. `~/.chef/profile` as that's the wording we use everywhere else?

lib/chef/knife/config.rb
```diff
@@ -0,0 +1,126 @@
+#
+# Author:: Joshua Timberman <opensource@housepub.org>
```
tas50: ?

coderanger: I started from https://github.com/jtimberman/knife-config/blob/master/lib/chef/knife/config.rb, though there is little of that code left I think :)

jaymzh:
I vote for `knife config get` to give us room to grow

coderanger:
Okay, I changed the command name. I want to fill out some other commands under `knife config` (pending a chat with the Workstation and UX teams) but I'm not going to do that in this PR, it's too big already :) Should be ready for review eyes.

PR #7393, opened 2018-06-25
remove some ruby 1.8.7-isms

lamont-granquist:
been a long, long time since we've needed to use iconv.

this does not address the rest of dubiousness of this mixin, but just
cleans out the old codepaths.


lamont-granquist:
@stuartpreston could use a thumb

PR #7398, opened 2018-06-27
Remove sudo/gcc-c++ package installs from kitchen tests

tas50:
Sudo was added to dokken images. gcc-c++ was added to appbundle-updater. Also use the kitchen-appbundle-updater from rubygems since it's up to date. No need to pull from the repo for this.

### DIFF
--- a/kitchen-tests/Gemfile
+++ b/kitchen-tests/Gemfile
@@ -4,7 +4,7 @@ gem "rake" # required to build some native extensions
 gem "chef", path: ".."
 gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids failures when we bump chef major
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
-gem "kitchen-appbundle-updater", git: "https://github.com/chef/kitchen-appbundle-updater", branch: "master"
+gem "kitchen-appbundle-updater"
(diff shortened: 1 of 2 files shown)


kitchen-tests/Gemfile
```diff
@@ -4,7 +4,7 @@ gem "rake" # required to build some native extensions
 gem "chef", path: ".."
 gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids failures when we bump chef major
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
-gem "kitchen-appbundle-updater", git: "https://github.com/chef/kitchen-appbundle-updater", branch: "master"
+gem "kitchen-appbundle-updater"
 gem "kitchen-dokken", "< 2.0" # 2.x fails atm: https://travis-ci.org/chef/chef/jobs/199125787
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
 gem "kitchen-vagrant", git: "https://github.com/test-kitchen/kitchen-vagrant.git", branch: "master"
```
lamont-granquist: should probably do this consistently and either pull from rubygems or pull from master...   i can't recall if there was a really good reason for pulling master here other than "CI"

tas50: They've slowly been added to resolve gem deps, but we should probably do them all or go back to just chef/ohai.

PR #7404, opened 2018-06-29
Resolve a few simple rubocop identified performance issues

tas50:
Use max instead of sort.last and key? instead of keys.include?


tas50:
Not worth troubleshooting the failures

PR #7414, opened 2018-07-02
Chefstyle fixes

lamont-granquist:
Fixes:

- Style/HashSyntax
- Layout/IndentHeredoc
- Layout/DocPosition
- Layout/LeadingCommentSpace
- Style/UnneededInterpolation
- Style/TrailingCommandInHashLiteral

note that I tried to do one-fix-per-commit, but some fixes caused other cops to fire which fixes some whitespace issues in the individual commits.


chef-config/spec/unit/workstation_config_loader_spec.rb
```diff
@@ -389,12 +389,12 @@
 
       context "and has a default profile" do
         let(:content) do
-          content = <<EOH
-[default]
-node_name = 'barney'
-client_key = "barney_rubble.pem"
-chef_server_url = "https://api.chef.io/organizations/bedrock"
-invalid_config_option1234 = "foobar"
+          content = <<~EOH
```
tas50: This is a very nice rule addition

lamont-granquist: it made me happy =)

lib/chef/application/windows_service.rb
```diff
@@ -200,8 +200,8 @@ def run_chef_client
           timeout: Chef::Config[:windows_service][:watchdog_timeout],
           logger: Chef::Log
         )
-        Chef::Log.trace "#{result.stdout}"
```
tas50: I guess it's more clear what we're trying to do here, but this seems a bit odd

lib/chef/chef_fs/chef_fs_data_store.rb
```diff
@@ -537,9 +537,9 @@ def list(path)
           if chef_fs.versioned_cookbooks || path[0] == "cookbook_artifacts"
             result = with_entry([ path[0] ]) do |entry|
               # list /cookbooks/name = filter /cookbooks/name-version down to name
-              entry.children.map { |child| split_name_version(child.name) }.
-              select { |name, version| name == path[1] }.
```
tas50: So much better

lib/chef/application/windows_service.rb
lamont-granquist: the parens are maybe unnecessary but `result.stdout.to_s` is definitely better than `"#{result.stdout}"`

PR #7417, opened 2018-07-03
chefstyle: fix Style/MutableConstant

lamont-granquist:


coderanger:
How badly does it break things if we do `# frozen_string_literal: true` instead/in addition to this? I'm guessing "very" :)

tas50:
Very

lamont-granquist:
yeah frozen string literals is a battle for another day...

PR #7423, opened 2018-07-05
Allow digging of attributes

mal:
Facilitates using [`dig`](https://ruby-doc.org/core-2.4.0/Hash.html#method-i-dig) (available since ruby 2.3) on `node` attribute objects.

This is helpful in at least two cases:
1. When dealing with attributes out of the cookbook authors control
2. When dealing with attributes that may be redefined by the end user as a `Hash`

```rb
node.default['foo']['bar']['qux'] = 'value'
[...]
node.dig('foo', 'bar', 'qux') #=> 'value'
node.dig('foo', 'baz', 'qux') #=> nil
```

Hoping this is simple and unobjectionable 🤞

### DIFF
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -89,6 +89,7 @@ class Chef
        :count,
        :cycle,
        :detect,
+       :dig,
        :drop,
        :drop_while,
        :each,
--- a/spec/unit/node/attribute_spec.rb
+++ b/spec/unit/node/attribute_spec.rb
@@ -617,6 +617,46 @@ describe Chef::Node::Attribute do
     end
   end
 
+  describe "dig" do
+    before do
+      @attributes = Chef::Node::Attribute.new(
+        {
+          "one" =>  { "two" => "three" },
+          "hut" =>  { "two" => "three" },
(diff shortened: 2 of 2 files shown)


coderanger:
This feature is already available as .read and I’m not sure we need two ways to access it.

mal:
Aha, thanks, it can be tough to find stuff like this, I know about `dig` from Ruby core, I had no idea that `read` existed let alone that it does the same thing.

coderanger:
Yeah, it's yet another thing that was never really documented very well. FWIW I'm not 100% against having `node.dig` just for folks that come in from the Ruby world but it should almost certainly be an alias to `read` so they work the same, and we should at least make sure that we want to have both (maybe support both but have a Foodcritic rule to force to one or the other)? Mostly worried we would end up with another situation like strings vs. symbols for attrs where there is a choice to make with minimal guidance or clarity to users about which to go with.

mal:
Makes sense, as someone that had to go through that symbol to string migration, I'm very much onboard with avoiding any kind of repeat. I guess for now I'll close this PR and create an issue for supporting `dig` in whatever form that ultimately takes.

PR #7432, opened 2018-07-06
Pull in latest win32-service gem

tas50:
Also bump all the omnibus gems. This new gem resolves deprecation warnings on Ruby 2.4


stuartpreston:
I'm assuming the Travis failures re: utf-8 are unrelated as they appeared previously in the build history.

PR #7443, opened 2018-07-10
Expeditor: Remove labels that don't exist / chef 12 config

tas50:
Remove labels that we don't use in chef/chef. Also remove Chef 12 branch setup.


.expeditor/config.yml
```diff
@@ -1,4 +1,4 @@
-# Documentation available at http://expeditor-docs.es.chef.io/
+# Documentation available at https://expeditor-docs.es.chef.io/
```
btm: there isn't actually an https version of this site.

scotthain: This should be https://expeditor.chef.io

PR #7455, opened 2018-07-11
Add knife config get/use-profile commands

coderanger:
Mostly because explaining the `.chef/context` file to people is hard, so hide it behind a command. These should probably grow up to be `chef` CLI commands some day but I'm not doing that until we have a better idea of the plan there.


tas50:
These would be much easier to use if there was a banner that explained what they do. Out of the box one might assume that get-profile lists all the profile. Making sure they have proper explanations on the help screen would be super useful.

Also a knife list profile command would be super useful because you need to know what profiles you can set w/o cracking open the config file.

coderanger:
Fair, opinions on `knife config list-profiles` vs `knife config profiles` (or something else entirely)?

tas50:
I am a fan of list personally

tas50:
Consider that approval me, but logged into the wrong account

PR #7456, opened 2018-07-11
Mildly speed up knife by moving deps into dep blocks

tas50:
We can lazily load the requirements for knife plugins, but we don't do that for all our plugins. We should. Is this a game changer? Nope. It does make it 4% faster though, which makes it worth doing.

Side note: We have some dupe requires in here and we also requires mixlib-shellout and mixin-shellout when we only used mixin-shellout.

### DIFF
--- a/lib/chef/knife/cookbook_site_share.rb
+++ b/lib/chef/knife/cookbook_site_share.rb
@@ -18,22 +18,19 @@
 #
 
 require "chef/knife"
-require "chef/mixin/shell_out"
 
 class Chef
   class Knife
     class CookbookSiteShare < Knife
 
-      include Chef::Mixin::ShellOut
-
       deps do
         require "chef/cookbook_loader"
         require "chef/cookbook_uploader"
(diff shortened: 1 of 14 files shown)


lib/chef/knife/cookbook_site_share.rb
```diff
@@ -18,22 +18,19 @@
...
       deps do
         require "chef/cookbook_loader"
         require "chef/cookbook_uploader"
         require "chef/cookbook_site_streaming_uploader"
-        require "mixlib/shellout"
-      end
+        require "chef/mixin/shell_out"
 
-      include Chef::Mixin::ShellOut
+        include Chef::Mixin::ShellOut
```
lamont-granquist: does that still work the same way?  `include` needs `self` to be the class, and if `deps` gets called by the instance then this would behave...  strangely...

tas50:
Closing this in favor of a 1 at a time PR that will make a bit more sense. Some of these requires don't need to be moved. They just need to be removed.

PR #7466, opened 2018-07-14
Move all knife cookbook site plugin logic into knife supermarket

tas50:
We want to deprecate knife cookbook site. This moves all the logic into
the knife supermarket plugins and has the legacy knife cookbook site
plugins inherit from that.

### DIFF
--- a/lib/chef/knife/cookbook_site_download.rb
+++ b/lib/chef/knife/cookbook_site_download.rb
@@ -17,106 +17,18 @@
 #
 
 require "chef/knife"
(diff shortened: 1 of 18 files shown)


lamont-granquist:
I could have sworn it was this way already, but I might be remembering what I suggested should happen rather that what did happen...

tas50:
I believe we pulled in the supermarket gem then killed off all the commands. That was part 1 in deprecating knife cookbook site so technically this is the 2nd part.

PR #7475, opened 2018-07-16
ifconfig: Add gateway property on RHEL/Debian based systems

tas50:
This reimplements #6279 after our conversion of this resource to use properties. It also adds support for RHEL since that's documented here:

https://www.centos.org/docs/5/html/Deployment_Guide-en-US/s1-networkscripts-interfaces.html


tas50:
Looks like there's more to do for RHEL so I'm actually going to close this out in favor of the sep PR. So many PRs...

tas50:
This is in fact the correct approach.

PR #7477, opened 2018-07-16
add back clean_array API

lamont-granquist:
broke public API that habitat package cookbook was using, so add
this back but make it deprecated.


coderanger:
An alternative would be to have the cookbook monkeypatch this in.

lamont-granquist:
the cookbook is getting patched to just call `shell_out!('hab', *command)` now which is all it actually needs.

the point of this is to unbreak any other cookbooks out there where someone might have copypasta'd clean_array around.

Issue #6403, opened 2017-09-14
windows_task: tasks will not run on battery power


spuder:
Scheduled tasks by default will not run if on battery power. This includes test kitchen vms. 

<img width="647" alt="screenshot 2017-09-13 21 17 46" src="https://user-images.githubusercontent.com/242382/30410805-7e2807ba-98ca-11e7-8dc9-7dabc364b338.png">

This should not be the default, or at least it should be configurable. If you are on a laptop, you will spend hours of fun debugging why chef works when your laptop is plugged in, but not when on battery power. (Even if your workstation is a mac)

## How

There are 3 main ways to make a windows scheduled task. 

- Schedtask.exe
- [New-ScheduledJobOption](https://docs.microsoft.com/en-us/powershell/module/psscheduledjob/new-scheduledjoboption?view=powershell-5.1)
- XML

SchedTask.exe is not able to modify the power/battery settings. You must either use the powershell module, or import the task as xml. 

    <DisallowStartIfOnBatteries>false</DisallowStartIfOnBatteries>

### Possible config options
```ruby
windows_task 'foo' do
  conditions :start_only_ac_power, false, :stop_on_battery, false 
end
```
```ruby
windows_task 'foo' do
  start_only_ac_power false
  stop_on_battery false
end
```

jaymzh:
Hey @tas50 - what's the status of this? It looks like `windows_task` (now?) uses XML... I think one could just add a property that sets DisallowStartIfOnBatteries in the XML?

This has bitten our client team pretty hard. :/

spuder:
It looks like the XML is only when updating an existing task, chef still uses the commands for creation. 

https://github.com/chef/chef/blob/master/lib/chef/provider/windows_task.rb

jaymzh:
@spuder - `def action_create` seems to use XML to me. Am I missing something?

tas50:
We're currently working on rewriting the whole thing to use the win32-taskscheduler gem, which is XML based under the hood. This will allow for all the fancy (and basic) things like running on battery and real priorities.

The WIP PR is https://github.com/chef/chef/pull/6815

tas50:
Now that we landed the rewrite of windows task to no longer uses the CLI we should figure out if we can get this fixed up. The default for the resource should definitely be to run on battery power.

PR #7483, opened 2018-07-17
Support for battery power options in windows_task resource

dheerajd-msys:
### Description

Support for running tasks when on battery. Properties `disallow_start_on_battery` and `stop_on_battery` is being added.

### Issues Resolved
Fixes https://github.com/chef/chef/issues/6403

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -327,14 +327,12 @@ class Chef
         def task_needs_update?(task)
           flag = false
           if new_resource.frequency == :none
-            # require "byebug"
-            # byebug
             flag = (task.account_information != new_resource.user ||
             task.application_name != new_resource.command ||
             task.parameters != new_resource.command_arguments.to_s ||
             task.principals[:run_level] != run_level ||
-            task.settings[:disallow_start_if_on_batteries] != new_resource.disallow_start_on_battery ||
-            task.settings[:stop_if_going_on_batteries] != new_resource.stop_on_battery)
+            task.settings[:disallow_start_if_on_batteries] != new_resource.disallow_start_if_on_batteries ||
+            task.settings[:stop_if_going_on_batteries] != new_resource.stop_if_going_on_batteries)
           else
             current_task_trigger = task.trigger(0)
             new_task_trigger = trigger
@@ -358,8 +356,8 @@ class Chef
                   task.principals[:logon_type] != logon_type ||
                   task.principals[:run_level] != run_level ||
                   PRIORITY[task.priority] != new_resource.priority ||
-                  task.settings[:disallow_start_if_on_batteries] != new_resource.disallow_start_on_battery ||
-                  task.settings[:stop_if_going_on_batteries] != new_resource.stop_on_battery
+                  task.settings[:disallow_start_if_on_batteries] != new_resource.disallow_start_if_on_batteries ||
+                  task.settings[:stop_if_going_on_batteries] != new_resource.stop_if_going_on_batteries
               if trigger_type == TaskScheduler::MONTHLYDATE
                 flag = true if current_task_trigger[:run_on_last_day_of_month] != new_task_trigger[:run_on_last_day_of_month]
               end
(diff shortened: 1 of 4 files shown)


lib/chef/resource/windows_task.rb
```diff
@@ -59,6 +59,8 @@ class WindowsTask < Chef::Resource
       property :execution_time_limit, [String, Integer], default: "PT72H" # 72 hours in ISO8601 duration format
       property :minutes_duration, [String, Integer]
       property :minutes_interval, [String, Integer]
+      property :disallow_start_on_battery, [TrueClass, FalseClass], default: false
```
tas50: We'll want to add description and introduced fields to all of these properties. We introduced the description and introduced in Chef 14.0 and we're going to be using it in the future to generate documentation.

https://github.com/chef/chef/pull/7475/files#diff-b03a778f5af81b89ff2db80b67f7adc0R100 is a good example of how we use both of them.

Vasu1105:
@tas50 We are not getting this query `Also do we need to make Chef depend on a newer windowstask gem now?`

Do you mean here depend on win32-taskscheduler gem. Since we are using win32-taskscheduler library for task creation we do need to depend on it.

lib/chef/resource/windows_task.rb
dheerajd-msys: Thank you @tas50 for nice suggestion. I will add fields `description` and `introduced` for the properties.

lib/chef/provider/windows_task.rb
```diff
@@ -324,10 +324,14 @@ def convert_hours_in_minutes(hours)
         def task_needs_update?(task)
           flag = false
           if new_resource.frequency == :none
+            # require "byebug"
```
Vasu1105: remove byebug

tas50:
@Vasu1105 If we pull in the most recent win32-taskscheduler the path tests for the task scheduler resource fail. You'll want to take a look at that logic before we can pull in the new gem.

Vasu1105:
@tas50 Fixed the breaking test for windows_task resource in #7504

Vasu1105:
@dheerajd-msys Please add functional test for this change.

lib/chef/resource/windows_task.rb
```diff
@@ -59,6 +59,12 @@ class WindowsTask < Chef::Resource
       property :execution_time_limit, [String, Integer], default: "PT72H" # 72 hours in ISO8601 duration format
       property :minutes_duration, [String, Integer]
       property :minutes_interval, [String, Integer]
+      property :disallow_start_on_battery, [TrueClass, FalseClass], default: false,
+               introduced: "14.4",
+               description: "Scheduled task option when system is on AC power."
+      property :stop_on_battery, [TrueClass, FalseClass], default: false,
```
tas50: Can we change this description to "Stop the task when the system switches to battery power"

lib/chef/resource/windows_task.rb
```diff
@@ -59,6 +59,12 @@ class WindowsTask < Chef::Resource
       property :execution_time_limit, [String, Integer], default: "PT72H" # 72 hours in ISO8601 duration format
       property :minutes_duration, [String, Integer]
       property :minutes_interval, [String, Integer]
+      property :disallow_start_on_battery, [TrueClass, FalseClass], default: false,
```
tas50: Can we change this property name to "start_on_battery" to match the property name below.

lib/chef/resource/windows_task.rb
```diff
@@ -59,6 +59,12 @@ class WindowsTask < Chef::Resource
...
+               introduced: "14.4",
+               description: "Scheduled task option when system is on AC power."
```
tas50: This can get changed to "Start the task if the system is running on battery power"

lib/chef/resource/windows_task.rb
Vasu1105: @tas50 we are setting up the property as per msdn docs so that we can identify or relate them very easily. I will suggest to @dheerajd-msys the property name should be :disallow_start_if_on_battery and :stop_if_on_battery.

lib/chef/resource/windows_task.rb
Vasu1105: My suggestion for description based on :property name is `Disallow start of the task if the system is running on battery power.`

Vasu1105:
@tas50  Please let us know if you want the property names to be changed. Otherwise please review the changes and since PR is ready for merge.

PR #7491, opened 2018-07-18
Fix dup logger bugs again

lamont-granquist:
See the inline comments attempting to explain my logic

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -202,10 +202,12 @@ class Chef
         end
     end
 
-    # Based on config and whether or not STDOUT is a tty, should we setup a
-    # secondary logger for stdout?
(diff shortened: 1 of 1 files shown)


lamont-granquist:
⚠️ completely untested and i'm peace'ing out and going on vacation.

this makes sense to me though because anytime someone sets the log_location we generally want to be setting up another logger.

the only time we want to not be doing that is when STDOUT is closed.  if we're checking STDOUT.tty? then we just create bugs about redirecting to a file, or piping to less vs. executing on the command-line.  and if we're running daemonized we never want to create a new logger.

lamont-granquist:
(my interest in fixing this is also super low, someone should add specs and/or just merge if it works)

lamont-granquist:
#7698

PR #7493, opened 2018-07-19
Fix Expeditor action set ordering

tduffield:


tduffield:
Merging obvious fix to Expeditor config

PR #7500, opened 2018-07-23
Update inspec-core and win32-taskscheduler to latest

tas50:
InSpec ships with some nice bugfixes
win32-taskscheduler supports priorities and running on battery


tas50:
This actually breaks win32-taskscheduler so we're going to have to hold off on this

PR #7509, opened 2018-07-26
Fixes rpmdb corruption errors for multiple yum package

piyushawasthi:

### Description
1: Fixes rpmdb corruption errors if package resource is given a name that contains a space.
2: Validate package_name if contains a space.
3: Fix Yum arguments so that user can add and remove multiple packages following way:
```
package 'box bag foo' do
  action :remove
end
```
```
package 'remove multiple packages' do
   package_name "box bag foo"
   action :remove
end
```
```
package 'remove multiple packages' do
   package_name '["box", "bag", "foo"]'
   action :remove
end
```
Similarly with `install`

### Issues Resolved

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -64,6 +64,9 @@ class Chef
       def load_current_resource; end
 
       def define_resource_requirements
+        # Before install and upgrade we are validating package name
+        validate_package(new_resource.package_name)
+
         # XXX: upgrade with a specific version doesn't make a whole lot of sense, but why don't we throw this anyway if it happens?
         # if not, shouldn't we raise to tell the user to use install instead of upgrade if they want to pin a version?
         requirements.assert(:install) do |a|
@@ -170,7 +173,6 @@ class Chef
       end
 
       def removing_package?
-        validate_package(new_resource.package_name)
         if !current_version_array.any?
           # ! any? means it's all nil's, which means nothing is installed
           false
@@ -306,11 +308,6 @@ class Chef
       end
 
       def validate_package(name)
-        raise( Chef::Exceptions::Package, "package name not allowed spaces, tabs, newlines, etc." ) if invalid_name?(name)
-      end
-
-      def invalid_name?(x)
-        ( x.is_a?(Array) && x.any? { |name| name.strip.index(/\s/) } ) || ( x.is_a?(String) && x.strip.index(/\s/) )
       end
 
       # used by subclasses.  deprecated.  use #a_to_s instead.
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -170,6 +170,7 @@ class Chef
       end
 
       def removing_package?
+        validate_package(new_resource.package_name)
         if !current_version_array.any?
           # ! any? means it's all nil's, which means nothing is installed
           false
@@ -302,6 +303,14 @@ class Chef
 
       def unlock_package(name, version)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :unlock" )
+      end
+
+      def validate_package(name)
+        raise( Chef::Exceptions::Package, "package name not allowed spaces, tabs, newlines, etc." ) if invalid_name?(name)
+      end
+
+      def invalid_name?(x)
+        ( x.is_a?(Array) && x.any? { |name| name.strip.index(/\s/) } ) || ( x.is_a?(String) && x.strip.index(/\s/) )
       end
 
       # used by subclasses.  deprecated.  use #a_to_s instead.
--- a/spec/functional/resource/yum_package_spec.rb
+++ b/spec/functional/resource/yum_package_spec.rb
@@ -79,6 +79,34 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
     context "vanilla use case" do
       let(:package_name) { "chef_rpm" }
 
+      it "raises error if the package name contain whitespace" do
+        flush_cache
+        yum_package.package_name("chef rpm")
+        expect { yum_package.run_action(:install) }.to raise_error(Chef::Exceptions::Package, /Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed./)
+        expect(yum_package.updated_by_last_action?).to be false
+      end
+
+      it "raises error if the package name contain tabs" do
+        flush_cache
+        yum_package.package_name("chef\trpm")
+        expect { yum_package.run_action(:install) }.to raise_error(Chef::Exceptions::Package, /Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed./)
+        expect(yum_package.updated_by_last_action?).to be false
+      end
+
+      it "raises error if the package name contain newlines" do
+        flush_cache
+        yum_package.package_name("chef\nrpm")
+        expect { yum_package.run_action(:install) }.to raise_error(Chef::Exceptions::Package, /Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed./)
+        expect(yum_package.updated_by_last_action?).to be false
+      end
+
+      it "raises error if the any package name contain spaces" do
+        flush_cache
+        yum_package.package_name(["chef_rpm", "foo bar"])
+        expect { yum_package.run_action(:install) }.to raise_error(Chef::Exceptions::Package, /Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed./)
+        expect(yum_package.updated_by_last_action?).to be false
+      end
+
       it "installs if the package is not installed" do
         flush_cache
         yum_package.run_action(:install)
@@ -308,7 +336,7 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
     context "with constraints" do
       it "with nothing installed, it installs the latest version" do
         flush_cache
-        yum_package.package_name("chef_rpm >= 1.2")
+        yum_package.package_name("chef_rpm-1.10")
         yum_package.run_action(:install)
         expect(yum_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
@@ -316,7 +344,7 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
 
       it "when it is met, it does nothing" do
         preinstall("chef_rpm-1.2-1.#{pkg_arch}.rpm")
-        yum_package.package_name("chef_rpm >= 1.2")
+        yum_package.package_name("chef_rpm-1.2")
         yum_package.run_action(:install)
         expect(yum_package.updated_by_last_action?).to be false
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.2-1.#{pkg_arch}$")
@@ -324,7 +352,7 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
 
       it "when it is met, it does nothing" do
         preinstall("chef_rpm-1.10-1.#{pkg_arch}.rpm")
-        yum_package.package_name("chef_rpm >= 1.2")
+        yum_package.package_name("chef_rpm-1.10")
         yum_package.run_action(:install)
         expect(yum_package.updated_by_last_action?).to be false
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
@@ -332,7 +360,7 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
 
       it "with nothing intalled, it installs the latest version" do
         flush_cache
-        yum_package.package_name("chef_rpm > 1.2")
+        yum_package.package_name("chef_rpm-1.10")
         yum_package.run_action(:install)
         expect(yum_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
@@ -340,7 +368,7 @@ describe Chef::Resource::YumPackage, :requires_root, external: exclude_test do
 
       it "when it is not met by an installed rpm, it upgrades" do
         preinstall("chef_rpm-1.2-1.#{pkg_arch}.rpm")
-        yum_package.package_name("chef_rpm > 1.2")
+        yum_package.package_name("chef_rpm-1.10")
         yum_package.run_action(:install)
         expect(yum_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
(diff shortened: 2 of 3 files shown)


lib/chef/mixin/shell_out.rb
```diff
@@ -101,7 +101,7 @@ def shell_out_with_systems_locale!(*args, **options)
       end
 
       def a_to_s(*args)
-        Chef.deprecated(:shell_out, "a_to_s is deprecated use shell_out with splat-args")
+        # Chef.deprecated(:shell_out, "a_to_s is deprecated use shell_out with splat-args")
```
rhass: Why are we commenting out this deprecation warning? If we don't need it we should remove it entirely since we can always restore it from version control.

lib/chef/resource/yum_package.rb
```diff
@@ -38,7 +38,7 @@ class YumPackage < Chef::Resource::Package
       # its own arguments as unintended side-effects (and why functional languages that don't allow modifcations
       # of variables eliminate entire classes of bugs).
       # </lecture>
-      property :package_name, [ String, Array ], identity: true, coerce: proc { |x| x.is_a?(Array) ? x.to_a : x }
+      property :package_name, [ String, Array ], identity: true, coerce: proc { |x| x.is_a?(Array) ? x.to_a : ((x.strip.index(/\s/)) ? x.to_s.split(" ") : x) }
```
rhass: A ternary does not seem to be needed for `x.is_a?(Array)` since the `true` case is doing a needless `to_a` against an object which should already be an array. 

Also, a ternary is probably the wrong thing to use for `x.strip.index(/\s/))`. Here is why, this condition will return true:

```ruby
[12] pry(main)> "one two three".strip.index(/\s/)
=> 3
[13] pry(main)> 0 ? "zero" : "not zero"
=> "zero"
[14] pry(main)> 1 ? "zero" : "not zero"
=> "zero"
```

Also, what is your intention with this: `x.to_s.split(" ")`? I ask because it can one of two things, one of which is effectively a no-op, and the other creates a complex nested array of strings. Eg:
```ruby
[16] pry(main)> %w(one two three).to_s.split(" ")
=> ["[\"one\",", "\"two\",", "\"three\"]"]
[17] pry(main)> "one two three".to_s.split(" ")
=> ["one", "two", "three"]
```

.rubocop.yml
```diff
@@ -32,3 +32,5 @@ Lint/ShadowedException:
 Layout/AlignHash:
   Enabled: true
   EnforcedLastArgumentHashStyle: ignore_implicit
+Style/NestedTernaryOperator:
+  Enabled: false
```
rhass: This is a style warning for a good reason, nested ternaries are hard to read and understand. Let's find a better way to do this. Maybe we need a private method that we can call to coerce out input value.

lib/chef/resource/yum_package.rb
PrajaktaPurohit: I think that coercion is to convert an immutable array into a normal array, since apparently the provider may convert it into an immutable array.

PrajaktaPurohit: Also, looks like the intention with `x.to_s.split(" ")` was to convert the string into an array of strings.
```irb(main):024:0> y = '  '
=> "  "
irb(main):025:0> y.strip.index(/\s/) ? y.to_s.split(" ") : y
=> "  "
irb(main):026:0> y.strip.index(/\s/)
=> nil```

```irb(main):029:0> y='box bag bar  '
=> "box bag bar  "
irb(main):030:0> y.strip.index(/\s/) ? y.to_s.split(" ") : y
=> ["box", "bag", "bar"]```

If the user entered %w(one two three) the to_a true case would handle it and it would not result in the complex array.

But I admit, it does not seem to do the expected thing in the following scenarios with the above assumption:

``irb(main):034:0> y = 'foo'
=> "foo"
irb(main):035:0> y.strip.index(/\s/) ? y.to_s.split(" ") : y
=> "foo"
irb(main):036:0> y = 'foo '
=> "foo "
irb(main):037:0> y.strip.index(/\s/) ? y.to_s.split(" ") : y
=> "foo "
irb(main):038:0> y = 'foo  '
=> "foo  "
irb(main):039:0> y.strip.index(/\s/) ? y.to_s.split(" ") : y
=> "foo  "```

File: .rubocop.yml (hunk 3 above)
Comment (PrajaktaPurohit) [reply .rubocop.yml]: Agreed!

File: lib/chef/mixin/shell_out.rb (hunk 1 above)
Comment (PrajaktaPurohit) [reply lib/chef/mixin/shell_out.rb]: Also not sure we can make available what what previously made known as deprecated. We should instead use the splat-args as mentioned in the suggestion for our changes.

File: lib/chef/resource/yum_package.rb [hunk 4]
```diff
@@ -62,6 +62,11 @@ class YumPackage < Chef::Resource::Package
       property :allow_downgrade, [ true, false ], default: false
 
       property :yum_binary, String
+
+      def get_package_name(x)
+        return x.to_a if x.is_a?(Array)
```
rhass: We don't need to call `to_a` on the array. `return x if x.is_a?(Array)` is all we need since the object will already be an array when this evaluates as true.

lib/chef/resource/yum_package.rb
```diff
@@ -62,6 +62,11 @@ class YumPackage < Chef::Resource::Package
       property :allow_downgrade, [ true, false ], default: false
 
       property :yum_binary, String
+
+      def get_package_name(x)
```
rhass: Let's make this a private method. There really is no reason anyone else should call this method.   Also a small comment or yard doc about the intent of the function for future reference would be awesome.

lib/chef/resource/yum_package.rb
piyushawasthi: Actually we need to add for ImmutableArray otherwise it's causing issue LIKE:
```
["one", "two", "three"] not to be a kind of Chef::Node::ImmutableArray
```
https://github.com/chef/chef/blob/master/lib/chef/node/immutable_collections.rb#L61

lamont-granquist:
I don't think we really want to support multiple packages as strings, and its not clear to me how this fixes the rpmdb corruption problem.  also you're converting the array args back to a joined string which defeats the purpose of the splat arg calling syntax of shell_out.

lamont-granquist:
yeah i think i'm :-1: on trying to make `package "foo bar baz"` install multiple packages.  that isn't required by the original MSYS-866 bug report and ultimately leads towards trying to support that syntax everywhere.

i think all we really need to do is validate the package names don't have spaces in them, if they do we should raise.  that means if we're given an array we should walk the array and raise if any of the packages have a space in them.

piyushawasthi:
Hi @lamont-granquist Thanks for review now I have added validation on package name that check if any package_name contain space so it'll stop the execution and show invalid package error to end user Like: invalid packages
```
package_name = "Foo bar"
or
package_name = ["Foo", "bar", "Invalid package"]
``` 
But still if user want to install multiple package on rhel that throwing TypeError: `no implicit conversion of Array into String` because shell_out! expecting argument in String instead of Array.

lamont-granquist:
`shell_out!` in master on not-windows definitely takes array args and not just strings.  if you're seeing that, there's something else gone wrong.

piyushawasthi:
Hi @lamont-granquist that working when update `chef-14.2.0/lib/chef/mixin/shell_out.rb` from `master -> shell_out.rb` because shell_out! code change in master for slat_args. so please review the PR and let me know if you need anything else from my end.

lib/chef/resource/yum_package.rb
```diff
@@ -62,6 +62,21 @@ class YumPackage < Chef::Resource::Package
       property :allow_downgrade, [ true, false ], default: false
 
       property :yum_binary, String
+
+      private
+
+      def get_package_name(x)
+        if is_valid_package?(x)
+          Chef::Log.error("Invalid package name #{x}")
+          exit 0
```
lamont-granquist: the exit 0 here is wrong.   this should probably be expressed as a `callback` validation argument instead of being in the `coerce` method so that it throws a proper validationfailed exception type.

lib/chef/resource/yum_package.rb
```diff
@@ -62,6 +62,21 @@ class YumPackage < Chef::Resource::Package
...
+        else
+          x.is_a?(Array) ? x.to_a : x
```
lamont-granquist: this is redundant with the coerce.

lib/chef/resource/yum_package.rb
```diff
@@ -62,6 +62,21 @@ class YumPackage < Chef::Resource::Package
...
+        end
+      end
+
+      def is_valid_package?(x)
+        (x.is_a?(Array) && x.any? { |package_name| package_name.strip.index(/\s/) }) || (x.is_a?(String) && x.strip.index(/\b \b/))
```
lamont-granquist: it is probably enough to fail on `.any? { |p| p =~ /\s/ }` or something like that, since package names should never contain whitespace (and the validation message should be something like that) spaces, tabs, newlines, etc are all not allowed.

lamont-granquist:
This logic should also most likely go into the superclass.  There's no package resource we have which couldn't use the ImmutableArray->Array coercion, and I don't think there's package managers that allow whitespace (maybe chocolatey/windows? -- but then it should be an exception for those windows specific package managers).

piyushawasthi:
Hi @lamont-granquist : Please review the changes as per your comment added exception in `package.rb`.

jaymzh:
Just weighing in here to agree with @lamont-granquist  that we shouldn't support strings with package lists in them, they should definitely be arrays. Validating someone didn't do it by accident and erroring out, though, seems like a good call.

lib/chef/provider/package.rb
```diff
@@ -304,6 +305,14 @@ def unlock_package(name, version)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :unlock" )
       end
 
+      def validate_package(name)
+        raise( Chef::Exceptions::Package, "package name not allowed spaces, tabs, newlines, etc." ) if invalid_name?(name)
+      end
+
+      def invalid_name?(x)
```
rhass: Thinking about this more, we should change `x` to `name` or `package_name` to be consistent with the rest of the code and make it a little clearer what `x` is evaluating. 

`name.strip.index(/\s/) }` can be reduced to `name =~ /\s/` since we are just checking to see if a value contains white spaces... but we also may want to check for tabs in the regex as mentioned in the exception message.

lib/chef/provider/package.rb
```diff
@@ -304,6 +305,14 @@ def unlock_package(name, version)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :unlock" )
       end
 
+      def validate_package(name)
+        raise( Chef::Exceptions::Package, "package name not allowed spaces, tabs, newlines, etc." ) if invalid_name?(name)
```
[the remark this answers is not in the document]
piyushawasthi: Hi @rhass : All done please have a look also create gist for updated regex. gist: https://gist.github.com/piyushawasthi/fd7fca5189afa5a84da77a00dfc006f6

lib/chef/provider/package.rb
```diff
@@ -304,6 +305,14 @@ def unlock_package(name, version)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :unlock" )
       end
 
+      def validate_package(name)
+        raise( Chef::Exceptions::Package, "Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed." ) if invalid_name?(name)
+      end
+
+      def invalid_name?(package_name)
+        ( package_name.is_a?(Array) && package_name.any? { |name| name.index(/\s|\\t|\\n/) } ) || ( package_name.is_a?(String) && package_name.to_s.index(/\s|\\t|\\n/) )
```
lamont-granquist: I think you can make this clearer by just doing `Array(package_name).any? {  .... }` to coerce single strings to Arrays of strings and then do the same validation?

lib/chef/provider/package.rb
```diff
@@ -170,6 +170,7 @@ def have_any_matching_version?
       end
 
       def removing_package?
+        validate_package(new_resource.package_name)
```
lamont-granquist: we should probably validate on installing packages as well.

lamont-granquist: (and update, and the rest of the APIs...)

lamont-granquist: and really these validation calls should probably not be buried in the helper method but should go directly into `action_install` and all the rest of the top-class actions so its more readable.  having `removing_package?` throw an exception as a side effect is poor, even though it may be less code.

piyushawasthi: Hi @lamont-granquist - Install and upgrade already checking and return `not found that package having space or invalid` that's why i have not added validation into that. Let me test if that required I'll add validation into that as well.

lamont-granquist: define_resource_requirements is probably a better place to add the validation then.

you probably want to name the helper method something like `def valid_package_names?` and then subclasses can inherit and override that if the validation turns out to be not appropriate for some package manager out there, or if we want to add more specific validation to subclasses at some point.

rhass:
Pinging @lamont-granquist for follow-up code review.

lib/chef/provider/package.rb
```diff
@@ -304,6 +310,14 @@ def unlock_package(name, version)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :unlock" )
       end
 
+      def validate_package(name)
+        raise( Chef::Exceptions::Package, "Package names containing whitespace characters such as spaces, tabs, or newlines are not allowed." ) if valid_package_names?(name)
+      end
+
+      def valid_package_names?(package_name)
+        Array(package_name).any? { |name| name.index(/\s|\\t|\\n/) }
+      end
```
lamont-granquist: So there are concerns that we're injecting this into every package provider now.  Particularly for package_names with source paths to files then there could be files with spaces that this would hit.

It is probably most appropriate to leave the hooks here in the superclass, but leave an empty implementation in the superclass, then move the implementation of the validation code into the yum_package provider.

lib/chef/provider/package.rb
```diff
@@ -148,6 +151,7 @@ def upgrade_description
       private :upgrade_description
 
       action :remove do
+        validate_package(new_resource.package_name)
```
lamont-granquist: not quite sure why we're still calling this directly in :remove/:purge/:reconfig when those actions all call DRR as well.

piyushawasthi: Hi @lamont-granquist  I have done with above changes also DRR is called only for :install/upgrade for :remove/:purge/:reconfig we are not calling that methods. I am testing on the changes and will update for the same.

lamont-granquist:
@tas50 take a quick look at this?

spec/functional/resource/yum_package_spec.rb
```diff
@@ -308,84 +308,95 @@ def pkg_arch
     context "with constraints" do
       it "with nothing installed, it installs the latest version" do
         flush_cache
-        yum_package.package_name("chef_rpm >= 1.2")
+        yum_package.package_name("chef_rpm")
+        yum_package.version("1.10-1")
```
lamont-granquist: yeah this is not equivalent, and these specs going red means that this change actually breaks this API.

lamont-granquist:
So what is the original rpmdb corruption error and how does it arise from what we're executing when we're given `yum_package 'foo bar baz"`?  and can we somehow better validate the package_name is a valid rpm query before taking the action that corrupts the rpmdb?

spec/functional/resource/yum_package_spec.rb
```diff
@@ -308,84 +336,84 @@ def pkg_arch
     context "with constraints" do
       it "with nothing installed, it installs the latest version" do
         flush_cache
-        yum_package.package_name("chef_rpm >= 1.2")
```
tas50: These specs need to come back since these are valid forms of a package name for yum

rhass:
This needs to be rebased before we can merge. The rebase should fix the expeditor errors.

Vasu1105:
Hello @lamont-granquist, @tas50, @rhass I am closing this PR since its really hard to track here the changes done after review. We will create clean PR which in which we will make sure that the review comments are considered.

spec/functional/resource/yum_package_spec.rb
Vasu1105: @tas50 Agreed on this we should not break which was working earlier.

Issue #7490, opened 2018-07-18
Unable to install a package via the package resource with zypper global options.


asafmgn:
## Description
Unable to install a package via the package resource with zypper global options.

    package 'sssd' do
       options '-D /tmp/repos.d/'
    end

The options in the package resource will be appended to the end of the zypper command:

    zypper install sssd -D /tmp/repos.d/

the proper way and the only way it works is lke this:

    zypper  -D /tmp/repos.d/ install sssd

    zypper --help
    Usage:
            zypper [--global-options] <command> [--command-options] [arguments]
            zypper <subcommand> [--command-options] [arguments

    Global Options:
    --reposd-dir, -D <dir>  Use alternative repository definition file

## Chef Version
    chef-solo --version
    Chef: 14.2.0

## Platform Version
    cat /etc/*release
    SUSE Linux Enterprise Server 12 (x86_64)
    VERSION = 12
    PATCHLEVEL = 3
    NAME="SLES"
    VERSION="12-SP3"
    VERSION_ID="12.3"
    PRETTY_NAME="SUSE Linux Enterprise Server 12 SP3"
    ID="sles"
    ANSI_COLOR="0;32"
    CPE_NAME="cpe:/o:suse:sles:12:sp3"

## Client Output
    chef-solo -o ad-auth::test
    [2018-07-18T14:47:25+03:00] WARN: No config file found or specified on command line, using command line options.
    [2018-07-18T14:47:25+03:00] WARN: *****************************************
    [2018-07-18T14:47:25+03:00] WARN: Did not find config file: /etc/chef/client.rb, using command line options.
    [2018-07-18T14:47:25+03:00] WARN: *****************************************
    Starting Chef Client, version 14.2.0
    [2018-07-18T14:47:34+03:00] WARN: Run List override has been provided.
    [2018-07-18T14:47:34+03:00] WARN: Original Run List: []
    [2018-07-18T14:47:34+03:00] WARN: Overridden Run List: [recipe[ad-auth::test]]
    resolving cookbooks for run list: ["ad-auth::test"]
    Synchronizing Cookbooks:
      - ad-auth (0.1.0)
    Installing Cookbook Gems:
    Compiling Cookbooks...
    Converging 1 resources
    Recipe: ad-auth::test
      * zypper_package[sssd] action install
        * No candidate version available for sssd
        ================================================================================
        Error executing action `install` on resource 'zypper_package[sssd]'
        ================================================================================
    
        Chef::Exceptions::Package
[...]

PR #7518, opened 2018-08-02
zypper_package: Add new global_options property

dheerajd-msys:

### Description

Unable to install a package via the package resource with zypper global options.
```
package 'sssd' do
   options '-D /tmp/repos.d/'
end
```
The options in the package resource will be appended to the end of the zypper command:
```
zypper install sssd -D /tmp/repos.d/
```
the proper way and the only way it works is lke this:
```
zypper  -D /tmp/repos.d/ install sssd
zypper --help
Usage:
        zypper [--global-options] <command> [--command-options] [arguments]
        zypper <subcommand> [--command-options] [arguments

Global Options:
--reposd-dir, -D <dir>  Use alternative repository definition file
```
Introduced `global_options` property to the `zypper_package` resource

### Issues Resolved
fixes https://github.com/chef/chef/issues/7490


lib/chef/provider/package/zypper.rb
```diff
@@ -109,7 +109,7 @@ def zypper_version
         end
 
         def install_package(name, version)
-          zypper_package("install", *options, "--auto-agree-with-licenses", allow_downgrade, name, version)
+          zypper_package("install", *options, allow_downgrade, name, version)
```
lamont-granquist: that looks like a behavior change

lib/chef/provider/package/zypper.rb
```diff
@@ -146,7 +146,7 @@ def zypper_package(command, *options, names, versions)
           if zypper_version < 1.0
             shell_out!("zypper", gpg_checks, command, *options, "-y", names)
```
lamont-granquist: *options here should probably be moved earlier as well.

lib/chef/provider/package/zypper.rb
```diff
@@ -157,6 +158,14 @@ def gpg_checks
...
           "--oldpackage" if new_resource.allow_downgrade
         end
+
+        def clean_deps
+          "--clean-deps" if new_resource.action == :purge
+        end
+
+        def auto_agree_licenses
+          "--auto-agree-with-licenses" if new_resource.action == :install
+        end
```
lamont-granquist: this is creating more complicated code by burying the conditionals down here.  i'm not sure why all of that needed to get refactored.  the `--auto-agree-with-licenses` flag should get moved back to install and the `--clean-deps` flag should get moved back to purge.

dheerajd-msys: Thank you @lamont-granquist for review. We may need to introduce other property `--global-options` for `package` resource If we don't do above changes. As `zypper` provides two options one `global-options` and other is `command-options`. Also if we introduce `--global-options` to `package` resource then this property will become redundant for other package managers like `yum` , `apt` etc. Since they provide only one option.
 
### Zypper options 
`zypper [--global-options] <command> [--command-options] [arguments]`

However I am verifying it again for `--auto-agree-with-licenses` to move it in install and `--clean-deps` to purge.

I am also looking into the ways how to make better use of introducing `--global-options` either to `package` OR to `zypper_package` resource.

Please suggest your views on this.

dheerajd-msys:
@lamont-granquist @tas50 @btm Please review it again. Thank you.

rhass:
You will need to rebase on master to make expeditor happy.

dheerajd-msys:
I am adding functional test cases to this package now.

lib/chef/resource/zypper_package.rb
```diff
@@ -30,6 +30,7 @@ class ZypperPackage < Chef::Resource::Package
 
       property :gpg_check, [ TrueClass, FalseClass ], default: lazy { Chef::Config[:zypper_check_gpg] }
       property :allow_downgrade, [ TrueClass, FalseClass ], default: false
+      property :global_options, [ String, Array ], coerce: proc { |x| x.is_a?(String) ? x.shellsplit : x }
```
lamont-granquist: you should probably add a test which passed multiple options in a string and which then got shellsplit and that the call to `shell_out!()` wound up looking correct

dheerajd-msys: @lamont-granquist I've added unit test case for multiple options passed here https://github.com/chef/chef/pull/7518/files#diff-b638b95483df994a2982f5a0cfade9d3R175. Please take a look.

dheerajd-msys:
Yes @rhass I will revert these changes, I just wanted to confirm if functional test cases are running for `zypper_package` in travis. They are working properly in `jenkins adhoc build` (http://manhattan.ci.chef.co/job/chef-test/architecture=x86_64,platform=sles-12,project=chef,role=tester/217/console) and locally on `suse` and `opensuse` VM. And files which are commented out was failing here in travis and kept me stuck to verify for `zypper_package` tests.

lib/chef/resource/zypper_package.rb
```diff
@@ -34,6 +34,10 @@ class ZypperPackage < Chef::Resource::Package
       property :allow_downgrade, [ TrueClass, FalseClass ],
                description: "Allow downgrading a package to satisfy requested version requirements.",
                default: false, introduced: "13.6"
+
+      property :global_options, [ String, Array ],
```
dheerajd-msys: @tas50 Please suggest if I need to add `introduced` feild here.
Thanks

dheerajd-msys:
@rhass @lamont-granquist Please re-review this PR again. 
Thanks.

lib/chef/resource/zypper_package.rb
lamont-granquist: Yeah, you're going to need to add that.

.travis.yml
```diff
@@ -340,6 +340,23 @@ matrix:
...
+    before_script:
+      - sudo iptables -L DOCKER || ( echo "DOCKER iptables chain missing" ; sudo iptables -N DOCKER )
+      - cd kitchen-tests
+    script:
+      - bundle exec kitchen test rspec-opensuse-leap
+    after_failure:
+      - cat .kitchen/logs/kitchen.log
+    env:
+     - RSPEC_OPENSUSELEAP=42
+     - KITCHEN_YAML=kitchen.travis.yml
```
lamont-granquist: Wonder if this looks okay, @tas50?

lib/chef/resource/zypper_package.rb
dheerajd-msys: Added, Thank you.

kitchen-tests/kitchen.travis.yml
```diff
@@ -104,6 +104,9 @@ platforms:
   driver:
     image: dokken/opensuse-leap
     pid_one_command: /bin/systemd
+    intermediate_instructions:
+      - RUN /usr/bin/zypper --non-interactive update
+      - RUN /usr/bin/zypper --non-interactive install cron
```
tas50: if we need cron we should just add it to the dokken-images, but I can do this later. It's not a blocker on this.

Issue #5992, opened 2017-04-02
converge_if_changed ignore missing required properties


Annih:
## Description

The `converge_if_changed` method is only checking "properties specified by the user on the resource" to determine whether the block should be called or not. Meaning that if the user forget to pass my important property, my block protected by a `converge_if_changed` won't be called, and the `requirement` validation will never happen.

I already read #4596, but I think this issue is a more specific case, which I personally qualify as a bug what do you think? :)

## Chef Version

12.13.37 & 12.19.36

## Platform Version

* CentOS 6.8, & 7.2
* Windows Server 2012R2 & 2016

## Replication Case

Create a resource

```ruby
resource_name :my_resource

property :important_attribute, String, required: true

action :try_to_do_something do
  converge_if_changed(:important_attribute) do
    puts 'do something'
  end
end
```
And use it:
```ruby
my_resource 'test'
```

**Cc:** @aboten

lamont-granquist:
yeah so the problem here is that we should only be looking at the properties that the user cares about.  consider implementing a file provider using this api where the `new_resource.mode` is `nil` (the default) but the `current_resource.mode` exists and is 0755 because the file already exists.  we definitely don't want to do the comparison between the current_resource and new_resource on `mode` and decide that its 'different'.

so you can't fix this by enforcing that all the properties are checked, because they're not being checked for a good reason.

you might be able to simply run through all the properties given to converge_if_changed in a void context to get them and force the wave function to collapse first.

Annih:
I'm not sure to understand your proposed solution/workaround. Are you saying I have to perform a validation before my `converge_if_changed(:important_and_required_property)`?

This seems very counter productive/intuitive.
Instinctively I would expect that the `converge_if_changed` method to compare the current value of the property to the provided one.
If the property validation is done in a lazy way, I don't care this is an implementation detail, but I would expect the requirement validation to be performed at this point, because I need the value!
My code is saying _do something on this property_ implying _please verify that it's provided first_.

Am I weird to think like that?

lamont-granquist:
no i'm saying that `converge_if_changed` could be modified to do a get of all the properties first (and throw the values away).

> Instinctively I would expect the converge_if_changed method to compare the current value of the property to the provided one.

right, but that gets the case where the provided value is the default/unset/nil value wrong in many cases.

something more or less like this:

```ruby
resource_name :file_perms

property :path, String, name_property: true
property :mode, Integer
property :owner, String
property :group, String

load_current_value do
[...]

Issue #4596, opened 2016-02-18
Chef doesn't fail if a resource's required properties is missing


failshell:
## Description

I'm writing a new resource in which i defined required properties. When using the resource and omitting the required attributes, Chef happily continues its run without failing. I would expect it to fail.
## Chef Version

I tested with 12.5.1 and 12.7.2.
## Platform Version

Centos 7.1
## Replication Case

Here's the resource file:

``` text
resource_name :rundeck_user

property :password, String, required: true
property :roles, Array, required: true

action :create do
  puts 'foo'
end
```

And how it's called inside another cookbook:

``` text
rundeck_user 'test'
```
## Stacktrace

As seen https://gist.github.com/failshell/17553f8b90287f5a6da8, if i raise then I get the expected result.

coderanger:
This has always been slightly confusing, but the behavior hasn't changed. The `required` flag is only checked on the first time the property is used. If you change the action code to `puts new_resource.roles` it would fail as expected.

@lamont-granquist This is definitely not an `accepted minor` IMO, if we wanted to change this behavior it would have to be in a major release.

lamont-granquist:
Wow, that's a bit lolwut.

Lets Accepted Major this and drop the priority down to Low.

failshell:
@coderanger ah cool, good to know. thought i was going nuts here :)

coderanger:
:+1: It's definitely something that catches a lot of people off guard.

failshell:
Strangely, I've been writing resources for years and never thought to check if that behavior worked before accessing the actually attribute.

Live and learn :)

sekrett:
I think old behavior is good. For example I have many properties in my resource, all of them are required for a `create` action, and only one for `destroy` action. If a property is not used, why Chef should fail? Any opinions?

lamont-granquist:
yeah this might be a feature.  the validation has to be lazy in order to deal with explicitly lazy properties and `coerce:` blocks and stuff.

the end result is that if you never use a property then it isn't validated.  the workaround is just to reference the property (in a void context should work) which will cause it to get accessed and validated.

i definitely don't see a way forwards to fixing this in Chef-13

lamont-granquist:
need to evaluate this one again in time for Chef-16 to see if there's any solution

lamont-granquist:
So probably the fix to this should be to before every action taken to run through all the properties which are required and poke them and cause them to fail validation if they're not set.

[...]

PR #7519, opened 2018-08-02
Minor fix while checking kernel parameter's value in sysctl

### LINKED ISSUE #7450: Sysctl idempotency check needs to consider the existence of the sysctl.d file, not just the value
## Description

The idempotency check in the sysctl resource only compares the existing value of the sysctl with the desired value.  If they match, it does nothing.  However, if the sysctl has *also* been changed outside of chef-client then the values may match, the /etc/sysctl.d file will not be written, and on a reboot the value will revert to whatever it was before.

If the /etc/sysctl.d file does not exist, then that should trigger the resource to fire and setup the /etc/sysctl.d file, even though the running value will not change.

## Chef Version
chef-client 14.1.12

## Platform Version
RHEL 7.5

## Replication Case
* Assume there is a sysctl parameter "foo" with a default value of "0"
* Set foo to "1" outside of Chef using `sysctl foo=1`
* Run a cookbook that contains the resource `sysctl('foo') { value(1) }`
* Reboot the system.

When the system comes back up, I expect that the sysctl parameter foo will retain the value configured by Chef (foo=1), however, it actually will change to its default value (foo=0).

I [originally reported this issue against the sysctl cookbook](https://github.com/sous-chefs/sysctl/issues/112).  I am re-reporting it here now that the sysctl resource is part of core Chef.

Nimesh-Msys:
- Added a check in sysctl.d to ensure the persisted value of a parameter
- This will allow sysctl resource to create a config file for a parameter irrespective of its current session value

Fixes: MSYS-871

Signed-off-by: [Nimesh](nimesh.patni@msystechnologies.com)

### Description

The value of a kernel parameter was previously compared with only its current session value. This was causing user to create a new file when he wanted to update the value using sysctl resource. To ensure the persisted value of this parameter, we have added a check in config file

### Issues Resolved

#7450 

### Check List

- [X] All tests pass


lib/chef/resource/sysctl.rb
```diff
@@ -116,6 +116,30 @@ def set_sysctl_param(key, value)
...
+
+      # shellout to sysctl to get the current value
+      # ignore missing keys by using '-e'
+      # convert tabs to spaces since sysctl tab deliminates multivalue parameters
+      # strip the newline off the end of the output as well
+      # and compare the value with chef-created config to maintain idempotency
+      #
+      def get_sysctl_value(key)
+        val = shell_out!("sysctl -n -e #{key}").stdout.tr("\t", " ").strip
+        raise unless val == get_sysctld_value(key)
```
tas50: Isn't this going to give us no current value if the values just don't match?

tas50: I guess we kind of have to do that since the system is in a 1/2 configured state. We should probably comment above the raise why we have to treat the systctl and sysctld values not matching as a no current value situation.

Nimesh-Msys: Yes, I referred within list of exceptions and found out this to be the best match according to the [comment](https://github.com/chef/chef/blob/master/lib/chef/exceptions.rb#L127).
Will update the comment above this raise

lamont-granquist: It might be easier to set `value` to `nil` instead of dealing with `current_resource` being `nil`.

i'm not certain i'm wild about the whole `current_value_does_not_exist!` API...

Nimesh-Msys: @lamont-granquist :+1:  Initially I thought the same, but since value is a required property, it raises a exception at runtime:
```
Chef::Exceptions::ValidationFailed: value is a required property
```

@tas50 : Ive updated the comments as per the suggestions. Could you please take a look and let me know in case of any issues

lamont-granquist: yeah, i see, and based on our bad handling of required properties (#5992, #4596) there's no really good solution here, so it looks like this is the way to fix this today...

PR #7521, opened 2018-08-02
Add API to skip resource parsing in a cookbook

lamont-granquist:
The cookbook itself can use this API from its own library segment in
order to skip loading some of its resources based on the chef-client
version.

### DIFF
--- a/lib/chef/chef_class.rb
+++ b/lib/chef/chef_class.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Lamont Granquist (<lamont@chef.io>)
-# Copyright:: Copyright 2015-2016, Chef Software, Inc.
+# Copyright:: Copyright 2015-2018, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


lamont-granquist:
For use today it would need to get back-monkeypatched to old chef-client versions:

```
unless defined?(Chef.skip_resources_on_chef_version)
  class Chef
    def self.skip_resources_on_chef_version(cookbook_name:, chef_version:, skip_resources:)
      if Gem::Requirement.new(chef_version).satisfied_by?(Gem::Version.new(Chef::VERSION))
        Chef.run_context.cookbook_collection[cookbook_name].files_for(:resources).select { |x| skip_resources.include? x['name'].split('/').last }.map {|x| x["full_path"] }.each do |x|
          Chef::Resource::LWRPBase.send(:loaded_lwrps)[x] = true
        end
      end
    end
  end
end

Chef.skip_resources_on_chef_version(cookbook_name: "chef_hostname", chef_version: ">= 14.0", skip_resources: [ "hostname.rb" ])
```

lamont-granquist:
A more generalized API might hook into the cookbook manifest itself and delete entries out of it instead of fiddling with the loaded_lwrps hash.  That would mean libraries could delete recipes, attributes, etc out of the manifest and the cookbook_compiler would then skip them...  We could do something slightly nicer by building hooks into the cookbook_compiler, but any hooks we build wouldn't be usable on older chef-client versions...

PR #7529, opened 2018-08-07
kitchen rspec testing

lamont-granquist:
🚧


kitchen-tests/kitchen.travis.yml
```diff
@@ -69,6 +69,9 @@ platforms:
     image: dokken/centos-7
     pid_one_command: /usr/lib/systemd/systemd
     intermediate_instructions:
+      - RUN yum -y install cronie util-linux e2fsprogs
```
tas50: cronie util-linux an go away now. I added it to dokken-images

lib/chef/provider/mount/mount.rb
```diff
@@ -53,6 +53,7 @@ def mountable?
         def enabled?
           # Check to see if there is a entry in /etc/fstab. Last entry for a volume wins.
           enabled = false
+          return unless ::File.exist?("/etc/fstab")
```
tas50: We should at least throw an debug level message here if we do this

.travis.yml
```diff
@@ -322,6 +321,23 @@ matrix:
       - cat /tmp/out.txt
       - sudo cat /var/log/squid3/cache.log
       - sudo cat /var/log/squid3/access.log
+  - rvm: 2.4.4
```
tas50: Can we get a brief comment here at to what this test is. The description in the PR is pretty solid

kitchen-tests/cookbooks/rspec/metadata.rb
```diff
@@ -0,0 +1,8 @@
+name             "rspec"
+license          "Apache-2.0"
+description      "Runs nothing"
```
tas50: A description of what we're trying to do with this will be useful for the future folks

kitchen-tests/cookbooks/rspec/templates/run-chef-rspec
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
tas50: Not that it totally matters but why a template if it's static?

kitchen-tests/kitchen.travis.yml
```diff
@@ -108,3 +111,9 @@ suites:
   - name: end-to-end
     run_list:
       - recipe[end_to_end::default]
+  - name: rspec
+    run_list:
+      - recipe[rspec]
```
tas50: With the lifecycle hooks couldn't we just copy this file in w/o ever needing a chef cookbook?

kitchen-tests/cookbooks/rspec/templates/run-chef-rspec
lamont-granquist: i always use templates because in the future it might not be static.

PR #7531, opened 2018-08-08
Automatically open PRs on dep gem updates

tas50:
This uses expeditor subscriptions to automatically cut a PR to update gems when a dependent Chef managed gem is pushed to rubygems.

### DIFF
--- a/.expeditor/update_dep.sh
+++ b/.expeditor/update_dep.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -evx
+
+branch="expeditor/${GEM_NAME}_${VERSION}"
+git checkout -b "$branch"
+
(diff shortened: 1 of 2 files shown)


.expeditor/update_dep.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -evx
+
+branch="expeditor/${GEM_NAME}_${VERSION}"
+git checkout -b "$branch"
+
+bundle update $GEM_NAME
```
tyler-ball: Could add `-v $VERSION`. That could potentially highlight if there is some promotion problem later down the line

tas50: Do you mean a conflict that prevents pulling in latest?

Issue #7356, opened 2018-06-12
User resource on windows applies comment as Full Name, does not allow for setting Full Name


sarahbakal:
## Description
User resource on windows applies comment as Full Name, does not allow for setting Description.

## Platform Version
Windows

## Replication Case
User resource on windows applies comment as Full Name, does not allow for setting Full Name
```
user 'testuser' do
  comment 'the comment'
  password 'password'
end
```
## Client Output

```

```

stuartpreston:
Thanks for the report, we agree this should be possible but is not today.  Using the `powershell_script` resource (`New-LocalUser` cmdlet) is probably the easiest workaround if you need to set the Description correctly.

PR #7537, opened 2018-08-10
Fix issue of setting comment for windows user

NAshwini:
Currently, User resource on windows applies `comment` as `full_name`, does not allow for setting `description`. Now able to set `comment` as `description` and not as `fullname`. Done testing with below output:
```
PS C:\Users\ashuser> $dom = $env:userdomain
PS C:\Users\ashuser> $usr = $env:username
PS C:\Users\ashuser> ([adsi]"WinNT://$dom/$usr,user").fullname

PS C:\Users\ashuser> ([adsi]"WinNT://$dom/$usr,user").description
the comment
```

Issue fixed : https://github.com/chef/chef/issues/7356

### DIFF
--- a/lib/chef/provider/user/windows.rb
+++ b/lib/chef/provider/user/windows.rb
@@ -45,7 +45,7 @@ class Chef
             user_info = @net_user.get_info
 
             current_resource.uid(user_info[:user_id])
-            current_resource.comment(user_info[:full_name])
+            current_resource.comment(user_info[:comment])
             current_resource.home(user_info[:home_dir])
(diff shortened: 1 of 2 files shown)


tas50:
@btm Does it make sense for us to add a Full Name property specifically for Windows. That's a pretty fundamental part of setting up real users on a system, but it doesn't really translate into *nix /etc/passwd world.

lamont-granquist:
If we need to add windows-specific properties to the windows_user resource those can go into the superclass here:

https://github.com/chef/chef/blob/master/lib/chef/provider/user/windows.rb

Issue #7530, opened 2018-08-07
osx_profiles resource uses relative paths to profiles tool


jessepeterson:
## Description

The `osx_profiles` resource uses relative paths to profile tool. Had a user that had different system paths and thus had failing chef on this resource.

## Chef Version
E.g. current GH master. 7c9178e5dd4ec7b5a75993c148b087bac6e5b53d

In the code: https://github.com/chef/chef/blob/e2f21177cae1da7dc462a8f17a3f73e5c915cba0/lib/chef/provider/osx_profile.rb#L191

## Platform Version
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G65
```

## Replication Case
Place a `profiles` executable higher in your path order than the system profiles command, `/usr/bin/profiles`. See as the `osx_profiles` resource will use that instead of the system path.

tas50:
Fixed by #7539. Thanks for the report

PR #7539, opened 2018-08-10
osx_profile: Use the full path to /usr/bin/profiles

tas50:
This prevents an app named profiles being later in the path, which was reported by a user.


tas50:
Fixes #7530

PR #7550, opened 2018-08-14
Add centos-6 rspec functional testing as well

lamont-granquist:
there's enough rpm/yum lib differences that el-6 blew up in isolation
in jenkins


tas50:
Closing this since it's not functional yet.

PR #7553, opened 2018-08-16
Allow remote_directory to source from /templates

thechile:
**Backstory**

I mostly use the _/files_ directory to store content that will, to the best of my knowledge, never need to be dynamically updated i.e. binary, image, tar files etc.. For everything else I will use _/templates_, even if the files don't currently use any template variables.. I tend to find that 50% of the time during cookbook development or soon after, either myself or someone else on the team will decide that in fact it would be nice if some element of the files were dynamic. Having the files already in _/templates_ just makes it easier to add some `<%= .. %>` and move on. We accept the small performance penalty of this approach. 

**Problem**

One of the things I often have in _/templates_ are many directories that in turn contain multiple template .erb files. For example in _/files/nginx/conf.d/upstreams_ there might be 10+ files that all need to be rendered with some node attributes. Currently this leads to nasty stuff like declaring an array with the 10+ file names in the recipe to be looped over or having to use stuff like `Dir.foreach("#{node[:chef_client][:cache_path]}/cookbooks/cbexample/templates/nginx/conf.d/upstreams/") do |file|` and still have to loop over the output. 

**Solution**

I have made some small changes to the _remote_directory_ resource that lets the segment type be defined as either `:files` or `:templates`. When it’s defined as `:templates` the **source** will be _/templates_ instead of _/files_ and rendered without the above hacks.

If you could review and give feedback please. If it's OK i will update the `chef-web-docs` as well. 

**Related**:
- https://tickets.opscode.com/browse/CHEF-3958


coderanger:
👎 On this, I think it causes more problems than it solves. What we need is a better API for seeing what files exist in the cookbook because people keep trying to use things like as a workaround for a lack of that API.

thechile:
Agreed but talk of improving the API in this regard goes back to discussions i found as far back as 2009.

For the problem i described, the trivial amount of code in this PR would improve many Chef users daily work experience until such times as the refactoring you suggested happens.

coderanger:
Unfortunately "this fixes it for now" as an argument breaks down when any new API needs to be supported ~indefinitely (barring future deprecation, but those are painful at the best of times).

coderanger:
"is worse than choosing to do nothing" because if we add this now and then decide to implement a better solution later, we can never remove this :D

thechile:
OK, well thanks for sharing your thoughts. Hopefully this functionality will make it into Chef some day.

In the mean time i'll try monkey patching what i need.

lamont-granquist:
Yeah, probably not this implementation.  What we'd need to do is actually render the templates with erb, and the sub-resources that remote_directory creates should switch between directory/cookbook_file to directory/template.  I would also probably block making the remote_directory resource more complex until it is cleaned up.  It should use a real child context and create its subresources in the child context in 2018/2019-style chef rather than 2010-style chef.

PR #7558, opened 2018-08-17
Restart Python yum helper before each repo enable/disable

### LINKED ISSUE #7540: chef-client yum_package install fails with "No candidate version available for ..." error
## Description

On CentOS 6 a `chef-client` run using the `yum_package` provider to install a package will fail with a `No candidate version available for ...` error under fairly easy to trigger circumstances (see details under replication case)

This seems to be CentOS 6 specific. Chef 14 on CentOS 7 does not seem to have this bug. Chef 12 is unaffected.

## Chef Version

chef-client: all versions of Chef 14 above 14.0.136 (when the refactor of yum_package landed). I have tested Chef 14.0.136 and it is unaffected by this issue. Chef 14.0.139 is affected. I've also tested several later versions including the current "Current" (14.4.18) and the current Stable (14.3.37).  All newer versions since the yum refactor seem to be affected. Chef 12 is unaffected.

## Platform Version

CentOS 6.10 x86_64

## Replication Case

```
package 'foo' do
    action [:remove]
end

package 'bar' do
    options "--enablerepo=blah"
    action [:install]
end
```
Where `bar` is in repo `blah` and `blah.repo` has `enabled=0`. If `foo` is *not* already installed when the chef run occurs then the installation of `bar` will fail with a `No candidate version available for bar` error.

This pattern is fairly common in our cookbooks, ie `foo` conflicts with with `bar` so the recipe will ensure `foo` is uninstalled before attempting to install `bar`

I had a dig through the Chef code. It does seem that there may be a bug in the yum Python API in CentOS 6. `yum_helper.py` appears to successfully enable the `blah` repo, but when it then calls `returnPackages` nothing comes back from the yum API for the requested package. The reason it works if `foo` is installed is that when Chef removes a package it restarts `yum_helper.py` (see here: https://github.com/chef/chef/blob/425d599c90b7a07ac903502b46fd3045cc41ae98/lib/chef/provider/package/yum.rb#L131-L135 by calling `flushcache` ). I think what's happening is that the search to find `foo` is leaving some stray state in `yum_helper.py` . I suppose the fix may be to more aggressively restart yum_helper.py on CentOS 6, but I assume that may affect performance.

## Client Output

Some info redacted since this is one of our production cookbooks. Note that this is running inside test kitchen during a kitchen converge

```

 * yum_package[GeoIP] action install

           * No candidate version available for GeoIP

           Error executing action `install` on resource 'yum_package[GeoIP]'

           Chef::Exceptions::Package

           -------------------------

           No candidate version available for GeoIP

           Resource Declaration:

           ---------------------

           # In /tmp/kitchen/cache/cookbooks/***redacted***

            16: package 'GeoIP' do

            17:     options "--enablerepo=***redacted***"

            18:     action [:install]

            19: end

           Compiled Resource:

           ------------------

           # Declared in /tmp/kitchen/cache/cookbooks/***redacted*** `from_file'

           yum_package("GeoIP") do

             package_name "GeoIP"

             action [:install]

             declared_type :package

             cookbook_name "***redacted***"

             recipe_name "default"

             options ["--enablerepo=***redacted***"]

           end

           System Info:

           ------------

           chef_version=14.2.0

           platform=centos

           platform_version=6.10

           ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]

           program_name=/opt/chef/bin/chef-solo

           executable=/opt/chef/bin/chef-solo
```

## Stacktrace

See above

cosinusoidally:
### Description

This restarts the yum Python helper before each repo enable/disable in order to work around a bug in the yum API that occurs on CentOS 6.

### Issues Resolved

#7540

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/yum/python_helper.rb
+++ b/lib/chef/provider/package/yum/python_helper.rb
@@ -112,10 +112,11 @@ class Chef
             parameters = { "provides" => provides, "version" => version, "arch" => arch }
             repo_opts = options_params(options || {})
             parameters.merge!(repo_opts)
+            # XXX: for now we restart before and after every query with an enablerepo/disablerepo to clean the helpers internal state
+            restart unless repo_opts.empty?
             query_output = query(action, parameters)
             version = parse_response(query_output.lines.last)
             Chef::Log.trace "parsed #{version} from python helper"
-            # XXX: for now we restart after every query with an enablerepo/disablerepo to clean the helpers internal state
             restart unless repo_opts.empty?
             version
           end


tas50:
Merging this under the obvious fix policy.

lamont-granquist:
I made the suggestion to copy this line of code, so I think we're okay without DCO here.

PR #7560, opened 2018-08-18
support repeated options in systemd_unit

### LINKED ISSUE #7559: systemd_unit does not support repeated options in content when using a hash
## Description
systemd_unit takes a hash which describes the systemd unit file. However, some options in that file can be specified more than once. The current implementation doesn't support that. I suggest allowing the value to be an array, and if it is, specify the option once for each value in the array.

## Chef Version
14.3.37

## Platform Version
Any using systemd.

dbresson:
### Description

Fixes partial support for systemd unit files by allowing options to be specified multiple times.

### Issues Resolved

Fixes #7559

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/resource/systemd_unit.rb
+++ b/lib/chef/resource/systemd_unit.rb
@@ -61,7 +61,9 @@ class Chef
             content.each_pair do |sect, opts|
               doc.section(sect) do |section|
                 opts.each_pair do |opt, val|
-                  section.option(opt, val)
+                  [val].flatten.each do |v|
+                    section.option(opt, v)
+                  end
                 end
               end
             end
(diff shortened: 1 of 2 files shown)


dbresson:
Looks like the tests that failed locally for me pass on Travis

tas50:
@dbresson Thanks for the fix here. Can you add an example to our docs here for the systemd_unit so the next person who runs into this problem can read about your fix:

https://github.com/chef/chef-web-docs/blob/master/chef_master/source/resource_systemd_unit.rst

dbresson:
https://github.com/chef/chef-web-docs/pull/1550

tas50:
Thanks for the docs update!

PR #7565, opened 2018-08-21
switch shell_out to shell_out! in func tests

lamont-granquist:
not failing here hides issues caused by e.g. mknod not working properly.


lamont-granquist:
no way this caused the busted kitchen tests....

PR #7524, opened 2018-08-03
add chef_version API to provides lines

lamont-granquist:
also adds the chef_version_for_provides helper method

PR #7566, opened 2018-08-21
lazy the default resource_name until after parsing

lamont-granquist:
only set it if the resource doesn't set one.

this means that we don't call resource_name or provides before parsing
the file so that the chef_version_for_provides API in #7524 can work
(otherwise we throw the deprecation warning before we ever parse the
file).

### DIFF
--- a/lib/chef/resource/lwrp_base.rb
+++ b/lib/chef/resource/lwrp_base.rb
@@ -65,6 +65,8 @@ class Chef
 
           LWRPBase.loaded_lwrps[filename] = true
 
+          # wire up the default resource name after the class is parsed only if we haven't declared one.
+          # (this ordering is important for MapCollision deprecation warnings)
           resource_class.resource_name resource_name.to_sym if resource_class.resource_name.nil?
 
           resource_class
--- a/lib/chef/resource/lwrp_base.rb
+++ b/lib/chef/resource/lwrp_base.rb
@@ -49,9 +49,7 @@ class Chef
 
           resource_name = filename_to_qualified_string(cookbook_name, filename)
 
-          # We load the class first to give it a chance to set its own name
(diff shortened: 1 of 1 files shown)


lamont-granquist:
not certain we can actually do this since a custom resource that does not declare a resource_name will now have the resource_name called afterwards, which will call `provides`, which may mask a `provides` set by the custom resource.  but its all complicated by the silly canonical name stuff, and i'm going to have to look at the code tomorrow, sometime after having some really stiff coffee...

lamont-granquist:
so we have some trivial use cases that this does not affect:

-  if the custom resource does not declare any resource_name or provides lines that i think this is equivalent
-  if the custom resource declares a resource_name then this is equivalent (and arguably better since we could go one step further and rip out the code which removes the canonical provides line on every subsequent resource_name call as step 1 of ripping out the magical provides here).

so we're left with the use case:

- a custom resource declares only provides lines.  furthermore, the provides line needs to be the same as the resource_name.

lamont-granquist:
ah so the provides lines also `remove_canonical_dsl` if the resource_name matches the name on the provides line.

so i think this actually is backwards compatible (or at least i can't think of any way it is not, and my testing is showing that the resulting node_map is the same in the corner cases that i think are important to consider).

lamont-granquist:
there's also the checking in `resource_name` to make sure that we wire up provides only if the class isn't already providing the name, so by calling resource_name later we're not wiping out any provides that were in the custom resource.

overall i think i've convinced myself that this is the way it should have been written and eagerly wiring up the resource_name was just doing work that we later have all this complicated code to un-do what we did.

lamont-granquist:
manually tested halite and poise against this since I think our tests here still don't properly test master...

PR #7572, opened 2018-08-22
Cleanup the rake file and make sure dep updates work

tas50:
Cleanup some of the old version logic that I believe was leftover from the github changelog generator
Remove rdoc stuff we don't need
Remove the require on rubygems which modern ruby doesn't require
Make sure we don't bundle install w/o rake (in development group) which was breaking the dependencies update task

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,10 +17,6 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


tas50:
Closing this out in favor of a larger cleanup

Issue #7568, opened 2018-08-22
`knife configure` no longer works as documented


keiththomps:
## Description

The `knife configure` command [is documented](https://docs.chef.io/knife_configure.html) to create a `knife.rb` file, but as of [this commit](https://github.com/chef/chef/commit/569841d88cdfa690dfa76af1bbbccff2746fe707#diff-6cf2d85903a8e481628dd4af8f31a511) it now creates a `~/.chef/credentials` file. Commands like `knife node list` work fine, but `knife upload` will fail with:

```
$ cd ~/chef-repo
$ knife upload cookbooks/example
ERROR: Attempt to use relative path 'cookbooks/example' when current directory is outside the repository path.

ERROR: Current working directory is '/home/user/chef-repo'.
```

## Chef Version

Chef 13.6.42 and newer

## Platform Version

CentOS 7

## Replication Case

Assuming functional Chef Server, client pem file, and a fresh workstation with ChefDK installed:

1. Generate repository via `chef generate repo chef-repo`
1. Configure knife via `knife configure`
1. `cd ~/chef-repo`
1. `knife ssl fetch`
1. `knife upload cookbooks/example`

## Commentary

The `~/.chef/credentials` file appears to be undocumented outside of [RFC-099](https://github.com/chef/chef-rfc/blob/master/rfc099-authentication-config-file.md). The simplest solution I could come up for in the case mentioned above was to create a `knife.rb` file within the repo's `.chef` directory (which I also had to generate) containing this:

```ruby
cookbook_path ["#{File.dirname(__FILE__)}/../cookbooks"]
```

I wasn't sure exactly where the *preferred* fix for this would go, but I'm more that willing to tackle this once I find out. First thought is to revert the changes to `knife configure` since there is currently not a good way to walk someone through creating a `knife.rb` file.

coderanger:
The bug is in the default value of `cookbook_repo_path` not working correctly (and also that the docs are very out of date with respect to how to configure knife).

PR #7575, opened 2018-08-23
Utilize repo discovery in Chef::ChefFS::Knife

keiththomps:
### Description

When using a newly generated chef-repo + `knife configure` the Knife
commands that are based off of `Chef::ChefFS::Knife` currently fail
because the `Chef::Config[:chef_repo_path]` and associated sub-paths
default to be based off of the `Chef::Config.cache_path` which will be
`~/.chef`. This commit checks for this case and instead utilizes `discover_chef_repo`
to get a more accurate default chef_repo_path based on where the
knife command is being run from.

### Issues Resolved

* #7568 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/chef_fs/knife.rb
+++ b/lib/chef/chef_fs/knife.rb
@@ -67,6 +67,15 @@ class Chef
           Chef::ChefFS::Config::INFLECTIONS.each_value do |variable_name|
             Chef::Config.delete("#{variable_name}_path".to_sym)
           end
+        elsif Chef::Config[:chef_repo_path] == Chef::Config.cache_path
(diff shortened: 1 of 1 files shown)


keiththomps:
I'm not exactly sure what's wrong with how the config defaults are set, and changing `ChefConfig::Config` has potential effects far outreaching this issue that is localized to commands based on `Chef::ChefFS::Knife`. I've opted to make the change here to introduce the smallest possible amount of bugs since this is my first contribution. Any guidance is appreciated.

lamont-granquist:
Yeah +1 that this is just making things more complicated to try to avoid a delicate change to make.

But that delicate change is the right one.  Gotta do the hard thing, not the easy fix that just creates more magic and wiring and options.  Closing this one.

PR #7584, opened 2018-08-24
Switch from opensuse 42 -> 15 in Travis tests

tas50:
15 is the new release and 42 goes EOL in the not so distant future. Let's test on the latest.


tas50:
Closing this since we need to get things working in omnitruck

PR #7590, opened 2018-08-27
openssl resources: Improve descriptions and fix provides for Chef 14.X

tas50:
Update the descriptions, add copyrights, add authors, make sure the cookbook wins until Chef 15 for the legacy openssl_x509 name.

### DIFF
--- a/lib/chef/resource/openssl_x509_certificate.rb
+++ b/lib/chef/resource/openssl_x509_certificate.rb
@@ -24,7 +24,7 @@ class Chef
 
       preview_resource true
       resource_name :openssl_x509_certificate
-      provides(:openssl_x509) { true } # legacy cookbook name. Cookbook will win. @todo Make this true in Chef 15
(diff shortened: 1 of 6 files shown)


lib/chef/resource/openssl_x509_certificate.rb
```diff
@@ -24,7 +26,7 @@ class OpensslX509Certificate < Chef::Resource
 
       preview_resource true
       resource_name :openssl_x509_certificate
-      provides(:openssl_x509) { true } # legacy cookbook name. Cookbook will win. @todo Make this true in Chef 15
+      provides(:openssl_x509) { false } # legacy cookbook name. Cookbook will win. @todo Make this true in Chef 15
```
lamont-granquist: that just globally disables this provides line entirely, probably not what you want.

lamont-granquist: think you want to be using noah's api here.

tas50: It's really just a reminder to get this fixed in Chef 15

PR #7632, opened 2018-09-07
Update expeditor config to use subscriptions

tas50:
I *think* I have this right.


tas50:
@tduffield Am I going to need further work here to make sure that the promote actions occur for the correct branches? If I promote a Chef 13 artifact am I going to globber master with Chef 13 dockerfiles?

tduffield:
@tas50 Yeah - you'll want to use a glob so that you don't pick up non-master release branch artifacts

PR #7633, opened 2018-09-10
Add locale resource more managing the system's locale

### LINKED ISSUE #7318: port locale resource to core
You should be able to manage a system's locale out of the box

vincentaubert:
### Description

adds the locale resource

### Issues Resolved

#7318

### Check List

- [x] All tests pass
- [?] RELEASE\_NOTES.md has been updated if required (not required for bugfixes, required for API changes)

Signed-off-by: Vincent AUBERT vincentaubert88@gmail.com


lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,93 @@
...
+        def up_to_date?(file_path, lang, lc_all = nil)
+          locale = IO.read(file_path)
+          locale.include?("LANG=#{lang}") &&
+            (node["init_package"] == "systemd" || lc_all.nil? || locale.include?("LC_ALL=#{lc_all}"))
+        rescue
+          false
+        end
+      end
+
+      allowed_actions :update
```
tas50: These 2 can go away

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,93 @@
+#
+# Cookbook:: locale
```
tas50: The cookbook and resource comments can go away

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,93 @@
...
+# limitations under the License.
+#
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class Locale < Chef::Resource
+      resource_name :locale
+      provides :locale
```
tas50: provides can go away

vincentaubert:
cleaned up

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,86 @@
...
+
+class Chef
+  class Resource
+    class Locale < Chef::Resource
+      resource_name :locale
+
+      description "Use the locale resource to set the system's locale"
+      introduced "14.5"
+
+      property :lang, String, default: "en_US.utf8"
```
tas50: You'll want to add description fields here as well that explains what these new properties do. You can check out the other resources in this same dir for an example.

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,86 @@
...
+      property :lc_all, String, default: "en_US.utf8"
+
+      action :update do
```
tas50: You'll want to add a description field that explains what this action is for. Again the other resources in this directory *should* have them as an example

tas50:
Looks like chef-config/pkg/chef-config-14.5.6.gem snuck in there too. You might want to delete that and amend the history so we can keep it out of git.

tas50:
The final thing to mention is you'll need to sign this off for the DCO so we can accept the contribution. Checkout out https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco for the details on the process. It's pretty straight forward.

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,86 @@
...
+# See the License for the specific slanguage governing permissions and
+# limitations under the License.
+#
+
+require "chef/resource"
+
+class Chef
+  class Resource
+    class Locale < Chef::Resource
+      resource_name :locale
```
tas50: Above this you'll want to put `preview_resource true` which will ensure the cookbook wins until Chef 15 is released and we make this an official resource.

vincentaubert: fixed

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,87 @@
...
+      preview_resource true
```
tas50: This actually has to go above the resource_name bit so it properly registers the resource name as a preview resource. It's one of the few things in a resource where the order matters

vincentaubert:
all fixes have been made let's have Travis advice :)

lib/chef/resource/locale.rb
```diff
@@ -0,0 +1,92 @@
...
+    class Locale < Chef::Resource
+      preview_resource true
+      resource_name :locale
+
+      description "Use the locale resource to set the system's locale"
+      introduced "14.5"
+
+      property :lang, String,
+               default: "en_US.utf8",
+               description: "set the default system language"
```
tas50: 1 more thing. We use these to auto generate the documentation so can you make sure all the description fields have sentences that are capitalized and end with periods.

PR #7649, opened 2018-09-17
Add @jjlimepoint as a maintainer for chef-provisioning

jjlimepoint:
As discussed with @tas50, add myself as a maintainer for chef-provisioning since we end up keeping it running a lot!

Signed-off-by: Jaymz Julian <jjulian@limepoint.com>


coderanger:
I think we should probably remove this from the maintainers file as we're purging it from ChefDK and other installers.

tas50:
So eventually we will probably remove this section but as of today it's an active product with a maintainer list. @jjlimepoint is actually maintaining it and has fixed quite a few critical issues in the AWS driver lately. We should note that work and the privs we gave him on Github by adding his name to the list.

PR #7653, opened 2018-09-17
Bump omnibus-software to latest

tas50:


tas50:
Doing this all in a single PR

PR #7659, opened 2018-09-18
windows_service: Remove potentially sensitive info from the log

### LINKED ISSUE #7634: windows_service does not properly support sensitive
## Description
Using the `windows_service` resource with the `sensitive true` property has no effect and exposes passwords in logs. This behaviour can be seen in Chef's code [here](https://github.com/chef/chef/blob/master/lib/chef/provider/service/windows.rb#L94)

## Chef Version
`12.21.4`
`13.8.5`

## Platform Version
Windows Server 2012 R2

## Replication Case
```ruby
windows_service 'Some Service' do
  run_as_user 'domain\\username'
  run_as_password 'my_password'
  sensitive true
  action %i( enable start )
end
```

## Client Output
```plaintext
[2018-09-11T15:59:20+01:00] INFO: windows_service[Some Service] configured with {:service_name=>"Some Service", :service_start_name=>"domain\\user", :password=>"my_password"}
```

## Stacktrace
n/a

stuartpreston:

### Description

Removes the logging of potentially sensitive information in the log when a Windows service is configured or reconfigured.

### Issues Resolved

#7634 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/service/windows.rb
+++ b/lib/chef/provider/service/windows.rb
@@ -91,7 +91,7 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
       }.reject { |k, v| v.nil? || v.length == 0 }
 
       Win32::Service.configure(new_config)
-      logger.info "#{@new_resource} configured with #{new_config.inspect}"
+      logger.info "#{@new_resource} configured."
 
       # LocalSystem is the default runas user, which is a special service account that should ultimately have the rights of BUILTIN\Administrators, but we wouldn't see that from get_account_right
       if new_config.key?(:service_start_name) && new_config[:service_start_name].casecmp("localsystem") != 0


lib/chef/provider/service/windows.rb
```diff
@@ -91,7 +91,7 @@ def start_service
       }.reject { |k, v| v.nil? || v.length == 0 }
 
       Win32::Service.configure(new_config)
-      logger.info "#{@new_resource} configured with #{new_config.inspect}"
```
tas50: Seems like we might want want to emit the full message in debug assuming sensitive isn't set? We probably never needed it in info.

stuartpreston: I don't see much point in logging anything here at all, but as the existing config/reconfig was logging to `info` I figured it would be an unexpected change to do anything other than show the non-sensitive info (which amounts to just the resource name).  So I figured I'd just remove that possibly sensitive info.

PR #7667, opened 2018-09-20
Deprecate ohai resource's ohai_name property

tas50:
This isn't actually used anywhere in the resource and there's no point. It reload's plugins based on the plugin property. We should kill this thing off and reduce the complexity a tiny bit.

### DIFF
--- a/lib/chef/resource/ohai.rb
+++ b/lib/chef/resource/ohai.rb
@@ -27,8 +27,12 @@ class Chef
 
       description "Use the ohai resource to reload the Ohai configuration on a node. This allows recipes that change system attributes (like a recipe that adds a user) to refer to those attributes later on during the chef-client run."
 
-      property :ohai_name, name_property: true, introduced: "12.17"
-      property :plugin, String
+      property :ohai_name,
+               deprecated: true,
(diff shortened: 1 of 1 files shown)


lib/chef/resource/ohai.rb
```diff
@@ -27,8 +27,12 @@ class Ohai < Chef::Resource
 
       description "Use the ohai resource to reload the Ohai configuration on a node. This allows recipes that change system attributes (like a recipe that adds a user) to refer to those attributes later on during the chef-client run."
 
-      property :ohai_name, name_property: true, introduced: "12.17"
-      property :plugin, String
+      property :ohai_name,
+               deprecated: true,
+               name_property: true, introduced: "12.17"
```
lamont-granquist: yeah but that's the name property, shouldn't we do the apt_update trick to allow it to be unnamed?

tas50: The name_property isn't actually used. It's just a dangling property that makes no sense in the context of the way the resource works.

lamont-granquist: yeah i see that `apt_update` doesn't actually  use a name_property either...  carry on...

PR #7677, opened 2018-09-24
Adds full_name property to user resource for Windows.

### LINKED ISSUE #7542: Add full_name property use user resource for Windows
We should allow configuring the full_name property in Windows particularly for shops that are managing laptops with Chef. Our net_use class already has support  for this. We just need to wire it up in the resource/provider code.

Vasu1105:

### Description

Adds full_name property to user resource for Windows.
### Issues Resolved
#7542

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/user/windows.rb
+++ b/lib/chef/provider/user/windows.rb
@@ -39,12 +39,12 @@ class Chef
             logger.warn("The 'gid' (or 'group') property is not implemented on the Windows platform. Please use the `members` property of the  'group' resource to assign a user to a group.")
           end
 
-          @current_resource = Chef::Resource::User.new(new_resource.name)
+          @current_resource = Chef::Resource::User::WindowsUser.new(new_resource.name)
           current_resource.username(new_resource.username)
           begin
             user_info = @net_user.get_info
-
             current_resource.uid(user_info[:user_id])
+            current_resource.full_name(user_info[:full_name])
             current_resource.comment(user_info[:comment])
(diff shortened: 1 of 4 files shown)


lib/chef/provider/user/windows.rb
```diff
@@ -39,12 +39,12 @@ def load_current_resource
             logger.warn("The 'gid' (or 'group') property is not implemented on the Windows platform. Please use the `members` property of the  'group' resource to assign a user to a group.")
           end
 
-          @current_resource = Chef::Resource::User.new(new_resource.name)
+          @current_resource = Chef::Resource::User::WindowsUser.new(new_resource.name)
```
Vasu1105: Note: I need to replace this I have added full_name for windows_user resource as it needs only be added for Winodws due to which its not available in base/parent class. So I need to initialise @current_resource with WindowsUser class.

lib/chef/resource/user/windows_user.rb
```diff
@@ -25,6 +25,9 @@ class WindowsUser < Chef::Resource::User
 
         provides :windows_user
         provides :user, os: "windows"
+
+        property :full_name, String,
+                  description: "The full name of the user."
```
tas50: We need to add an introduced field here for our documentation

lib/chef/provider/user/windows.rb
lamont-granquist: you've actually done the right thing here.  current_resource should have been a WindowsUser resource.

Issue #7629, opened 2018-09-07
windows_task is not idempotent on non-English versions of Windows


jugatsu:
## Description
`windows_task` resource converges each time as `user` property is not `SYSTEM` or any other from https://github.com/chef/chef/blob/master/lib/chef/resource/windows_task.rb#L119.

Task is created successfully but running under `система` or `СИСТЕМА` user. I think that similar problem exists on different versions of Windows (non-English).

https://github.com/chef/chef/blob/2c4f13a14d762a306dbe051c658cb9b848b670b6/lib/chef/provider/windows_task.rb#L330

```ruby
def task_needs_update?(task)
          flag = false
          if new_resource.frequency == :none
            flag = (task.account_information != new_resource.user ||
```

`task.account_information != new_resource.user` is always true because `account_information` is in different language and therefore not `SYSTEM`.

How we can workaround this use case?

## Chef Version
13.9.4

## Platform Version
Windows Server 2016 Datacenter Evaluation Russian Edition
```powershell
PS C:\Users\vagrant> get-winsystemlocale

LCID             Name             DisplayName
----             ----             -----------
1049             ru-RU            Русский (Россия)
```

## Replication Case
```ruby
include_recipe 'chef-client::task'
```

## Client Output
```
[2018-09-07T10:39:49-07:00] DEBUG: Providers for generic windows_task resource enabled on node include: [Chef::Provider::WindowsTask]
       [2018-09-07T10:39:49-07:00] DEBUG: Providers for generic windows_task resource enabled on node include: [Chef::Provider::WindowsTask]
       [2018-09-07T10:39:49-07:00] DEBUG: Provider for action create on resource windows_task[chef-client] is Chef::Provider::WindowsTask
       [2018-09-07T10:39:49-07:00] DEBUG: Provider for action create on resource windows_task[chef-client] is Chef::Provider::WindowsTask
       [2018-09-07T10:39:49-07:00] DEBUG: Looking for existing tasks
       [2018-09-07T10:39:49-07:00] DEBUG: Looking for existing tasks
       [2018-09-07T10:39:49-07:00] DEBUG: windows_task[chef-client] task exists.
       [2018-09-07T10:39:49-07:00] DEBUG: windows_task[chef-client] task exists.
       [2018-09-07T10:39:49-07:00] DEBUG: Finding system date format
       [2018-09-07T10:39:49-07:00] DEBUG: Finding system date format
       [2018-09-07T10:39:50-07:00] DEBUG: running: 
[...]

PR #7679, opened 2018-09-24
Add support for localized system account to windows_task resource

jugatsu:
Signed-off-by: Anton Kvashenkin <anton.jugatsu@gmail.com>

### Description

This PR adds full support (idempotent) for using localized system account in `windows_task` resource.

I think it would be better to remove hard-coded system accounts and use `Chef::ReservedNames::Win32::Security::SID` to determine if account is actually system or not. This can help us to support all localized versions of Windows.

### Issues Resolved

Closes https://github.com/chef/chef/issues/7629

### DIFF
--- a/lib/chef/provider/windows_task.rb
+++ b/lib/chef/provider/windows_task.rb
@@ -570,7 +570,16 @@ class Chef
         def logon_type
           # Ref: https://msdn.microsoft.com/en-us/library/windows/desktop/aa383566(v=vs.85).aspx
           # if nothing is passed as logon_type the TASK_LOGON_SERVICE_ACCOUNT is getting set as default so using that for comparision.
-          new_resource.password.nil? ? TaskScheduler::TASK_LOGON_SERVICE_ACCOUNT : TaskScheduler::TASK_LOGON_PASSWORD
+          user_id = new_resource.user
+          if Chef::ReservedNames::Win32::Security::SID.service_account_user?(user_id)
+            TaskScheduler::TASK_LOGON_SERVICE_ACCOUNT
+          elsif Chef::ReservedNames::Win32::Security::SID.group_user?(user_id)
+            TaskScheduler::TASK_LOGON_GROUP
+          elsif !new_resource.password.to_s.empty? # password is present
+            TaskScheduler::TASK_LOGON_PASSWORD
+          else
+            TaskScheduler::TASK_LOGON_INTERACTIVE_TOKEN
+          end
         end
 
         # This method checks if task and command properties exist since those two are mandatory properties to create a schedules task.
--- a/lib/chef/resource/windows_task.rb
+++ b/lib/chef/resource/windows_task.rb
@@ -17,318 +17,320 @@
 #
 
 require "chef/resource"
+require "chef/win32/security" if Chef::Platform.windows?
 
 class Chef
   class Resource
     class WindowsTask < Chef::Resource
-      resource_name :windows_task
-      provides(:windows_task) { true }
+      if Chef::Platform.windows?
+        resource_name :windows_task
+        provides(:windows_task) { true }
 
-      description "Use the windows_task resource to create, delete or run a Windows scheduled task. Requires Windows Server 2008 or later due to API usage."
-      introduced "13.0"
+        description "Use the windows_task resource to create, delete or run a Windows scheduled task. Requires Windows Server 2008 or later due to API usage."
+        introduced "13.0"
 
-      allowed_actions :create, :delete, :run, :end, :enable, :disable, :change
-      default_action :create
+        allowed_actions :create, :delete, :run, :end, :enable, :disable, :change
+        default_action :create
 
-      property :task_name, String, regex: [/\A[^\/\:\*\?\<\>\|]+\z/],
-               description: "The task name, such as 'Task Name' or '/Task Name'",
-               name_property: true
+        property :task_name, String, regex: [/\A[^\/\:\*\?\<\>\|]+\z/],
+                 description: "The task name, such as 'Task Name' or '/Task Name'",
+                 name_property: true
 
-      property :command, String,
-               description: "The command to be executed by the windows scheduled task."
+        property :command, String,
+                 description: "The command to be executed by the windows scheduled task."
 
-      property :cwd, String,
-               description: "The directory the task will be run from."
+        property :cwd, String,
+                 description: "The directory the task will be run from."
 
-      property :user, String,
-               description: "The user to run the task as.",
-               default: "SYSTEM"
+        property :user, String,
+                 description: "The user to run the task as.",
+                 default: Chef::ReservedNames::Win32::Security::SID.LocalSystem.account_simple_name
 
-      property :password, String,
-               description: "The user’s password. The user property must be set if using this property."
+        property :password, String,
+                 description: "The user’s password. The user property must be set if using this property."
 
-      property :run_level, Symbol, equal_to: [:highest, :limited],
-               description: "Run with ':limited' or ':highest' privileges.",
-               default: :limited
+        property :run_level, Symbol, equal_to: [:highest, :limited],
+                 description: "Run with ':limited' or ':highest' privileges.",
+                 default: :limited
 
-      property :force, [TrueClass, FalseClass],
-               description: "When used with create, will update the task.",
-               default: false
+        property :force, [TrueClass, FalseClass],
+                 description: "When used with create, will update the task.",
+                 default: false
 
-      property :interactive_enabled, [TrueClass, FalseClass],
-               description: "Allow task to run interactively or non-interactively. Requires user and password to also be set.",
-               default: false
+        property :interactive_enabled, [TrueClass, FalseClass],
+                 description: "Allow task to run interactively or non-interactively. Requires user and password to also be set.",
+                 default: false
 
-      property :frequency_modifier, [Integer, String],
-               default: 1
+        property :frequency_modifier, [Integer, String],
+                 default: 1
 
-      property :frequency, Symbol, equal_to: [:minute,
-                                              :hourly,
-                                              :daily,
-                                              :weekly,
-                                              :monthly,
-                                              :once,
-                                              :on_logon,
-                                              :onstart,
-                                              :on_idle,
-                                              :none],
-               description: "The frequency with which to run the task."
+        property :frequency, Symbol, equal_to: [:minute,
+                                                :hourly,
+                                                :daily,
+                                                :weekly,
+                                                :monthly,
+                                                :once,
+                                                :on_logon,
+                                                :onstart,
+                                                :on_idle,
+                                                :none],
+                 description: "The frequency with which to run the task."
 
-      property :start_day, String,
-               description: "Specifies the first date on which the task runs in MM/DD/YYYY format."
+        property :start_day, String,
+                 description: "Specifies the first date on which the task runs in MM/DD/YYYY format."
 
-      property :start_time, String,
-               description: "Specifies the start time to run the task, in HH:mm format."
+        property :start_time, String,
+                 description: "Specifies the start time to run the task, in HH:mm format."
 
-      property :day, [String, Integer],
-               description: "The day(s) on which the task runs."
+        property :day, [String, Integer],
+                 description: "The day(s) on which the task runs."
 
-      property :months, String,
-               description: "The Months of the year on which the task runs, such as: 'JAN, FEB' or '\*'. Multiple months should be comma delimited. e.g. 'Jan, Feb, Mar, Dec'."
+        property :months, String,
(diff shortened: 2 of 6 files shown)


jugatsu:
@Vasu1105 Can you take a look please.

Vasu1105:
Yes @jugatsu I will check once free from current task.

lib/chef/resource/windows_task.rb
```diff
@@ -227,9 +227,18 @@ def validate_user_and_password(user, password)
         end
       end
 
+      def user_is_system?(user)
+        if Chef::ReservedNames::Win32::Security::SID.from_account(user) == Chef::ReservedNames::Win32::Security::SID.LocalSystem ||
```
rhass: I haven't played with this in pry... so it may not work at all, but you might be able to clean this up a little with something like:
```
account = Chef::ReservedNames::Win32::Security::SID.from_account(user)
account == Chef::ReservedNames::Win32::Security::SID.LocalSystem
```
I am not sure if you can call the class method on the object or not, but if you can it might also work as:

```
account = Chef::ReservedNames::Win32::Security::SID.from_account(user)
account.LocalSystem || account.NtLocal || account.NtNetwork || account.BuiltinUsers
```
Again, I haven't tried this myself.... but in my head it compiles. :man_shrugging: :wink:

jugatsu: @rhass What do you think about this change?

jugatsu:
Any feedback is highly appreciated. @Vasu1105

lib/chef/provider/windows_task.rb
```diff
@@ -601,6 +603,17 @@ def run_schtasks(task_action, options = {})
         end
         # rubocop:enable Style/StringLiteralsInInterpolation
 
+        def task_user
```
tas50: Can you toss the comment on the top of the method here so it gets properly documented by yard. A yard return tag would also be useful for the web docs. https://gist.github.com/chetan/1827484

jugatsu: Done.

Vasu1105:
> Any feedback is highly appreciated. @Vasu1105

@jugatsu Thank you. We are working to fix the issue on win32-taskscheduler using the SID and once that is done we can come back to your PR because we are planning to remove hard coded values.

Nimesh-Msys:
> > Any feedback is highly appreciated. @Vasu1105
> 
> @jugatsu Thank you. We are working to fix the issue on win32-taskscheduler using the SID and once that is done we can come back to your PR because we are planning to remove hard coded values.

@jugatsu : Nice to choose SID approach. We may fix this with bare minimum changes here which will also be a clean code:
1. In `\lib\chef\win32\security\sid.rb`:
- Let's have an array `SYSTEM_USERS` here without any hardcoded values. Something like:
```
        SYSTEM_USER = [self.LocalSystem, self.NtLocal, self.NtNetwork,
                       self.BuiltinAdministrators, self.BuiltinUsers, self.Guests].map do |user_type|
          [user_type.account_name, user_type.account_simple_name]
        end.flatten.freeze
```

- And a method `system_user?` to check if the given user is a system user or not ! (may implement case insensitive match here)
```
        def self.system_user?(user)
          SYSTEM_USER.include?(user)
        end
```
2. In `\lib\chef\resource\windows_task.rb`:
- Removing `SYSTEM_USERS` form here
- Modifying default value of `user` property as:
```
      property :user, String,
               description: "The user to run the task as.",
               default: Chef::ReservedNames::Win32::Security::SID.LocalSystem.account_simple_name
```
- Last, In password_required? method:
```
@password_required ||= !Chef::ReservedNames::Win32::Security::SID.system_user?(user.upcase)
```

These changes are working but required to be tested thoroughly, but first win32-taskscheduler gem needs to be fixed.

@Vasu1105 , I have started working on [win32-taskscheduler](https://github.com/chef/win32-taskscheduler/pull/69/) for these changes. There were some issues while registration of a task by 'Users' or 'Administrators' which are fixed. Next, will work on implementing SID over there.

Thank you

rhass:
This looks a lot cleaner. I like it! :+1:

Nimesh-Msys:
> This looks a lot cleaner. I like it! 👍

Thanks @rhass , I've tested these changes with few more modifications against both Russian and English version of windows and the results are as per the expectations. May I send a PR for the same ?

@rhass  or @tas50  *Could you please share your views on this:*
Now since we are providing a system specific provision for user, will it not be good to use `система` or `СИСТЕМА` as the user on Russian version of windows instead of `SYSTEM`. Hence the correct recipe would be:

```
windows_task 'client' do
  run_level :highest
  command 'ipconfig.exe'
  user 'СИСТЕМА'
  frequency :hourly
  frequency_modifier 20
end
```

jugatsu:
> Now since we are providing a system specific provision for user, will it not be good to use система or СИСТЕМА as the user on Russian version of windows instead of SYSTEM. Hence the correct recipe would be:

@Nimesh-Msys I don't understand what do you mean by that.

We should not hard-code system accounts and should check is the user system or not (using SID approach). That's all.
User can choose `user` property whatever he wants.

jugatsu:
@Nimesh-Msys I'm waiting until  https://github.com/chef/win32-taskscheduler/pull/69 will be merged. Then I can make modification as per https://github.com/chef/chef/pull/7679#issuecomment-426669598 and test it.

Nimesh-Msys:
> > Now since we are providing a system specific provision for user, will it not be good to use система or СИСТЕМА as the user on Russian version of windows instead of SYSTEM. Hence the correct recipe would be:
> 
> @Nimesh-Msys I don't understand what do you mean by that.
> 
> We should not hard-code system accounts and should check is the user system or not (using SID approach). That's all.
> User can choose `user` property whatever he wants.

Its not hard coded @jugatsu, First we are fetching the actual user name on system based on SID and then comparing it with the input provided.
I have deliberately kept this approach. As per your [comment](https://github.com/chef/chef/issues/7629#issuecomment-420225915) when I tried

```
Get-WMIObject win32_group -filter "Name='Администраторы' AND LocalAccount=True" | select -ExpandProperty SID
```
It returns the SID `S-1-5-32-544`.  But when we pass this name in English,
```
Get-WMIObject win32_group -filter "Name='Administrators' AND LocalAccount=True" | select -ExpandProperty SID
```
It does not return any output.

There I thought, when we could not even fetch SID of English literals on Russian system, should we allow this on Chef ? Will it not overkill the app? there may be involved some security reasons!

Providing this feature would mean, If we are passing the user input in Hindi, we should be able to run this on Vietnamese Systems. Isn't it ?

@stuartpreston , @tas50 , @rhass may share their views on this.

Nimesh-Msys:
> @Nimesh-Msys I'm waiting until [chef/win32-taskscheduler#69](https://github.com/chef/win32-taskscheduler/pull/69) will be merged. Then I can make modification as per [#7679 (comment)](https://github.com/chef/chef/pull/7679#issuecomment-426669598) and test it.

@jugatsu , Could you please take a look at: [code changes and tested recipe](https://gist.github.com/Nimesh-Msys/952e0882153abb04fc95beece6be6205) for various types of users.

jugatsu:
@Nimesh-Msys What is the right way to test it? How can i bump `win32-taskscheduler` to 2.0?

Nimesh-Msys:
> @Nimesh-Msys What is the right way to test it? How can i bump `win32-taskscheduler` to 2.0?

Rebase your current branch with master. `win32-taskscheduler` is already bumped with `2.0` into chef.
To test your functionalities at Russian version of windows:
1. [Run this recipe](https://gist.github.com/Nimesh-Msys/952e0882153abb04fc95beece6be6205#file-recipe-tested-on-russian-version-of-windows) and see if it is not breaking for any system users

2. Do check for at least one non system user(With Password)

3. Please ensure that specs are running fine; Specially on English version of windows

***NOTE*** In your code changes, make sure it passes `chefstyle` and all the comments are in YARD format

jugatsu:
> > @Nimesh-Msys What is the right way to test it? How can i bump `win32-taskscheduler` to 2.0?
> 
> Rebase your current branch with master. `win32-taskscheduler` is already bumped with `2.0` into chef.
> To test your functionalities at Russian version of windows:
> 
> 1. [Run this recipe](https://gist.github.com/Nimesh-Msys/952e0882153abb04fc95beece6be6205#file-recipe-tested-on-russian-version-of-windows) and see if it is not breaking for any system users
> 2. Do check for at least one non system user(With Password)
> 3. Please ensure that specs are running fine; Specially on English version of windows
> 
> _**NOTE**_ In your code changes, make sure it passes `chefstyle` and all the comments are in YARD format

@Nimesh-Msys I know how to test this PR with recipes, etc. I can't figure it out how to test this PR with new `task-scheduler` on my Macos :) Previously I just edit gems inside `opscode` directory inside Windows WM running via Test-kitchen. A little bit hacky I know :)

Vasu1105:
@jugatsu You need to rebase your branch as told by @Nimesh-Msys and run bundle install inside this will install win32-taskscheduler new version on your system and then just test your changes with bundle exec chef-client command with your recipes.

jugatsu:
@Nimesh-Msys Can you review please.

Tested on Russian version of WIndows and it works like a charm.

jugatsu:
Need to figure out how to fix tests.

jugatsu:
There is an issue with `TASK_LOGON_PASSWORD`. `windows_task` is not idempotent when using password. Like this one:
```ruby
windows_task 'client' do
    run_level :highest
    user 'vagrant'
    password 'vagrant'
    command 'ipconfig.exe'
    frequency :hourly
    frequency_modifier 20
  end
```

Tests are also failed with
```bash
 1) Chef::Provider::WindowsTask#logon_type return logon_type bindary value as 1 as if password is not nil
     Failure/Error: expect(provider.send(:logon_type)).to be(1)

       expected #<Integer:3> => 1
            got #<Integer:11> => 5

       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/unit/provider/windows_task_spec.rb:420:in `block (3 levels) in <top (required)>'

Finished in 0.19027 seconds (files took 8.98 seconds to load)
63 examples, 1 failure

```

Working on it.

jugatsu:
I need help to fix Travis :) Appveyor is green.
```
An error occurred while loading ./spec/integration/client/client_spec.rb.
Failure/Error: default: Chef::ReservedNames::Win32::Security::SID.LocalSystem.account_simple_name
NameError:
  uninitialized constant Chef::ReservedNames::Win32::Security
```

lib/chef/resource/windows_task.rb
```diff
@@ -223,13 +222,13 @@ def validate_start_time(start_time, frequency)
 
       def validate_user_and_password(user, password)
         if password_required?(user) && password.nil?
-          raise ArgumentError, %q{Cannot specify a user other than the system users without specifying a password!. Valid passwordless users: 'NT AUTHORITY\SYSTEM', 'SYSTEM', 'NT AUTHORITY\LOCALSERVICE', 'NT AUTHORITY\NETWORKSERVICE', 'BUILTIN\USERS', 'USERS'}
+          raise ArgumentError, "Cannot specify a user other than the system users without specifying a password!. Valid passwordless users: #{Chef::ReservedNames::Win32::Security::SID::SYSTEM_USER.join(', ')}"
```
Nimesh-Msys: In the message if we use ```"'#{array.join("', '")```}'", as an interpolation; we may retain the single quotes. Please try this:

```
'#{Chef::ReservedNames::Win32::Security::SID::SYSTEM_USER.join("', '")}'
```
and then we can retain the old format.

It means, we would not need to alter single quotes in [spec#L1574](https://github.com/chef/chef/blob/master/spec/functional/resource/windows_task_spec.rb#L1574)

jugatsu:
@Nimesh-Msys I have no idea how to fix Travis (
`require "chef/win32/security" if Chef::Platform.windows?` doesn't work.

Nimesh-Msys:
> @Nimesh-Msys I have no idea how to fix Travis (
> `require "chef/win32/security" if Chef::Platform.windows?` doesn't work.

Travis is running on Ubuntu and not on Windows.
So, when we are loading files from `/lib/chef/resources.rb:142`, at this point this constant is also expected to be loaded. For that even if we use ```require "chef/win32/security"```,(even on ubuntu) it will break at FFI adapters.
To fix this, if we use this one liner check with the constant, It resolves the problem.

```
default: Chef::ReservedNames::Win32::Security::SID.LocalSystem.account_simple_name if Chef::Platform.windows?
```

But IMO, we should allow the class to be loaded but its properties and method should not consume any space on anything but windows platforms. We should wrap the entire logic to be platform specific. Hence the correct approach would be:
```
    class WindowsTask < Chef::Resource
      if Chef::Platform.windows?
       // The entire content
      end
    end
```
It won't allow travis to blow up.
Post changes, please recheck the indentation using chefstyle and also test it once on your local windows.

Vasu1105:
LGTM

tas50:
Thanks for digging in here @jugatsu

PR #7691, opened 2018-09-26
Upgrade to rspec 3.8.x

lamont-granquist:
Problem is described in detail here:

rspec/rspec-expectations#1068 (comment)

Note that this is self-inflicted damage and our pattern of having #to_hash
on objects be called by #to_json and to produce a serialized 'flavor' of
the object breaks what ruby's semantics are supposed to be and those
should all be `#to_h` methods, and we should deprecate `#to_hash` but I
imagine that will be a reasonably large breaking change (Although
hopefully most of the use by users will be Node#to_hash which is
entirely correct usage since that thing is actually a kind of glorified
Hash-like monster).


tas50:
@lamont-granquist fixer of evil Ruby things

lamont-granquist:
https://zverok.github.io/blog/2016-01-18-implicit-vs-expicit.html

Is the direct link to a good blog post explaining how we use `#to_hash` is wrong all over the place.

PR #7705, opened 2018-09-28
short circuit before the version_compare call

lamont-granquist:
combined with the other PRs this will make most of the
subclasses never even call version_compare now.

the subclasses that need allow_downgrade as a feature
should implement both the property and implement
version_compare now.


lamont-granquist:
it is possible now that version_compare in the superclass won't ever be called an we could raise an error that it must be implemented now...

IDK feel that aggressive tonight though.

PR #7712, opened 2018-10-02
chocolatey_source: add support for enable/disable and user/pw

ttdgcp:
### Description

Adds support to enable & disable chocolatey sources along with adding sources with a username and password, if desired.

### Issues Resolved

No outstanding issues- only discussed in the community Slack.

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/chocolatey_source.rb
+++ b/lib/chef/resource/chocolatey_source.rb
@@ -36,10 +36,12 @@ class Chef
                description: "The priority level of the source."
 
       property :user, String,
-               description: "The username to authenticate to the source"
+               introduced: "14.6",
+               description: "The username to authenticate to the source."
 
       property :password, String,
-               description: "The password to authenticate to the source"
+               introduced: "14.6",
+               description: "The password to authenticate to the source."
 
       load_current_value do
         element = fetch_source_element(source_name)
@@ -87,6 +89,7 @@ class Chef
       end
 
       action :enable do
+        introduced "14.6",
         description "Enables a Chocolatey source."
 
         if current_resource
@@ -97,6 +100,7 @@ class Chef
       end
 
       action :disable do
+        introduced "14.6",
         description "Disables a Chocolatey source."
 
         if current_resource
--- a/lib/chef/resource/chocolatey_source.rb
+++ b/lib/chef/resource/chocolatey_source.rb
@@ -53,7 +53,7 @@ class Chef
         bypass_proxy element["bypassProxy"] == "true"
         priority element["priority"].to_i
         user element["user"]
-        password element["password"]
+        password unprotect_password(element["password"])
       end
 
       # @param [String] id the source name
@@ -67,6 +67,12 @@ class Chef
         config_contents = REXML::Document.new(::File.read(config_file))
         data = REXML::XPath.first(config_contents, "//sources/source[@id=\"#{id}\"]")
         data ? data.attributes : nil # REXML just returns nil if it can't find anything so avoid an undefined method error
+      end
+
+      # @param [String] base64 string to unprotect
+      # @return [String] unprotected password
+      def unprotect_password(base64_protected)
+        powershell_out!("Add-Type -AssemblyName System.Security;$enc = [system.Text.Encoding]::UTF8;$input = \"#{base64_protected}\";$salt=\"Chocolatey\";$decryptedBytes=[System.Security.Cryptography.ProtectedData]::Unprotect([System.Convert]::FromBase64String($input),$enc.GetBytes($salt),[System.Security.Cryptography.DataProtectionScope]::LocalMachine); write-output $enc.GetString($decryptedBytes)").stdout.chomp
       end
 
       action :add do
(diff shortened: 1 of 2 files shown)


lib/chef/resource/chocolatey_source.rb
```diff
@@ -35,6 +35,12 @@ class ChocolateySource < Chef::Resource
       property :priority, Integer, default: 0,
                description: "The priority level of the source."
 
+      property :user, String,
+               description: "The username to authenticate to the source"
+
+      property :password, String,
```
tas50: Can you add introduced properties here for 14.6 since that's the version we'll ship this in. Here's an example: https://github.com/chef/chef/blob/429345d5a75782da0d6544bb3f60fa89ce964b94/lib/chef/resource/ifconfig.rb#L48

lib/chef/resource/chocolatey_source.rb
```diff
@@ -35,6 +35,12 @@ class ChocolateySource < Chef::Resource
...
+               description: "The password to authenticate to the source"
```
tas50: Make sure to add periods at the end of these sentences

lib/chef/resource/chocolatey_source.rb
```diff
@@ -78,6 +86,26 @@ def fetch_source_element(id)
         end
       end
 
+      action :enable do
```
tas50: can you add introduced "14.6" inside these actions as well so our documentation will pick up when we added this

ttdgcp: What is the proper syntax for introduced under an action?
```
      action :enable do
        introduced "14.6",
        description "Enables a Chocolatey source."
```

or

```
      action :disable do
        introduced: "14.6",
        description "Disables a Chocolatey source."
```?

**Comment by ttdgcp** [comment]
@stuartpreston Thanks for the feedback. I've pushed fixes for #1 and #2.

Do you have any recommendations on how to handle the encrypted password for #3?

File: lib/chef/resource/chocolatey_source.rb [hunk 4]
```diff
@@ -89,22 +90,22 @@ def fetch_source_element(id)
       end
 
       action :enable do
-        introduced "14.6",
+        introduced "14.6"
         description "Enables a Chocolatey source."
 
         if current_resource
-          converge_by("enable Chocolatey source '#{new_resource.source_name}'") do
+          converge_if_changed("enable Chocolatey source '#{new_resource.source_name}'") do
```
stuartpreston: Sorry that I probably didn't explain this clearly enough yesterday.  For converge_if_changed to work, we need to be retrieving the enabled/disable values when the resource is loaded, have a look at how other fields get retrieved and see if you can work it out, otherwise we can add it together (I'll make it work and you can see how I did it)

stuartpreston:
Well, it just so happened that I was in the Chocolatey GitHub looking at how they protect the password and unprotecting the password on the same machine is possible, so we can use that to compare.  I rewrote it in PowerShell 

```ruby
 # @param [String] base64 string to unprotect
      # @return [String] unprotected password
      def unprotect_password(base64_protected)
        powershell_out!("Add-Type -AssemblyName System.Security;$enc = [system.Text.Encoding]::UTF8;$input = \"#{base64_protected}\";$salt=\"Chocolatey\";$decryptedBytes=[System.Security.Cryptography.ProtectedData]::Unprotect([System.Convert]::FromBase64String($input),$enc.GetBytes($salt),[System.Security.Cryptography.DataProtectionScope]::LocalMachine); write-output $enc.GetString($decryptedBytes)").stdout.chomp
      end
```
You can then amend the value that is loaded into load_current_value: 
```ruby
  password element["password"]
```
to be
```ruby
  password unprotect_password(element["password"])
```
Also if you set `sensitive: true` on the password property that will solve concerns exposing the password in the logs i.e.
```ruby
      property :password, String,
               sensitive: true, 
               introduced: "14.6",
               description: "The password to authenticate to the source."
```
This probably needs some more manual testing but is definitely going in the right direction - thanks for your contribution!

ttdgcp:
@stuartpreston Sorry for the delay-- I think I now have the desired changes if I got the `converge_if_changed` syntax correct.

Thanks again for your review.

ttdgcp:
Hey @stuartpreston was hoping to bump this up this week for a review-- let me know if there's anything I can do!

stuartpreston:
@ttdgcp Sorry for the delay in replying. @tas50 and I spoke to Rob Reynolds about this at Chocolatey Fest recently and he recommended against decrypting the password to test for idempotency because that mechanism may change in an upcoming version.  At this point I think the option most consistent with other resources is to set the password field only if the user has changed, otherwise the resource will never be idempotent.  Do you think you would be able to have another look?

lamont-granquist:
For my own future sanity to explain the problem clearly:

- chocolatey commands are responsible for creating and writing this file and the format of it.

- the base64 encoding of the password field may change without warning in chocolatey at any point in the future.

- there is no capability within chocolatey to read the password and be able to use that for idempotency checks.

lamont-granquist:
I think I would support just shipping this, I think I'm okay with it breaking in the future, and then we fix it again.

tas50:
I'm going to close this out since we don't have a native way to handle idempotency here and chocolatey may change the underlying implementation without notice.

PR #7719, opened 2018-10-04
better docs for Chef::Knife::Bootstrap#validate_options!

bankair:
This PR add some documentation for method
Chef::Knife::Bootstrap#validate_options!.

- [x] All tests pass

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -409,7 +409,13 @@ class Chef
         end
       end
 
-      # make sure policyfile values are set correctly
+      # Ensure options are valid by checking policyfile values.
+      #
+      # The method call will cause the program to exit(1) if:
+      #   * Only one of --policy-name and --policy-group is specified
+      #   * Policyfile options are set and --run-list is set as well
+      #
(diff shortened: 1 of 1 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -409,7 +409,14 @@ def validate_name_args!
...
       end
 
-      # make sure policyfile values are set correctly
+      # Ensure options are valid by checking policyfile values.
+      #
+      # The method call will cause the program to exit(1) if:
+      #   * Only one of --policy-name and --policy-group is specified
+      #   * Policyfile options are set and --run-list is set as well
+      #
+      # === Returns
```
tas50: Can you change this  over to YARD format. I realize we're not YARD everywhere, but we're trying to slowly document the world. https://gist.github.com/chetan/1827484

bankair: Yeah, sure. Is there a specific for exit(1) in yard?

tas50: I think this is just @return [TrueClass] If options are valid.

bankair:
@tas50 : This way?

PR #7721, opened 2018-10-04
Honor chef_server_url for sandbox uploads

sbbowers:
### Description

Chef uploads cookbooks transactionally by generating a unique sandbox
URL to upload all the associated files to using a "POST
/organizations/<organization>/sandboxes" operation.

This POST operation returns the URL of the sandbox that should then be
used to PUT the cookbook's files to. This URL is pre-resolved, based on
the configuration of the chef server, which can be different from what
is configured in the client.

In special cases of network configuration, this discrepancy forces the
client to use a URL with a non-resolvable or unreachable host for
uploading cookbooks, rather than the scheme/host/port configured with
chef_server_url.

This change parses the sandbox URL returned by the server "POST
/organizations/<organization>/sandboxes" operation, and rewrites
it honor the chef_server_url configuration for subsequent cookbook PUT operations.

### Issues Resolved

- Does not resolve, but is semi-related to https://github.com/chef/chef-server/issues/50

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/cookbook_uploader.rb
+++ b/lib/chef/cookbook_uploader.rb
@@ -77,7 +77,8 @@ class Chef
 
       queue.process(@concurrency)
 
-      sandbox_url = new_sandbox["uri"]
+      sandbox_url = URI.parse(Chef::Config[:chef_server_url])
+      sandbox_url.path = URI.parse(new_sandbox["uri"]).path
       Chef::Log.trace("Committing sandbox")
       # Retry if S3 is claims a checksum doesn't exist (the eventual
       # in eventual consistency)


sbbowers:
Ah. I understand the use case with the cookbook storage back-end; thank you for pointing that out.  

Would a pull request for a new setting be acceptable? It could be something like "cookbook_upload_proxy", that we could configure in our client.rb/knife.rb to allow us to over-ride the host/port of the sandbox URLs provided from the server. This would also ensure continuing compatibility for the various storage back-ends by default.

I agree that this problem is ultimately a deficiency in the server code, however, the server can't (and probably shouldn't) know about every potential client configuration used to connect to it.  A cookbook proxy setting on the client side could also be used to create compatibility with various chef server configurations, as the API itself would have to change in order to fix chef/chef-server#50.

PR #7728, opened 2018-10-06
Remove audit mode from chef-client

tas50:
This just gives us a line count to the change and perhaps a starting
point for when we do this in Chef 15


lib/chef/client.rb
```diff
@@ -721,7 +676,6 @@ def converge(run_context)
           events.converge_complete
         rescue Exception => e
           events.converge_failed(e)
-          raise e if Chef::Config[:audit_mode] == :disabled
```
tas50: This bit seems odd to just remove

tas50:
@btm @lamont-granquist this is ready for 👁

lib/chef/client.rb
```diff
@@ -322,22 +294,11 @@ def run
...
-                  run_error || converge_error
-                else
-                  e = if run_error == converge_error
-                        Chef::Exceptions::RunFailedWrappingError.new(converge_error, audit_error)
-                      else
-                        Chef::Exceptions::RunFailedWrappingError.new(run_error, converge_error, audit_error)
-                      end
-                  e.fill_backtrace
-                  e
-                end
```
lamont-granquist: i think you still need this bit of code or at least something like:

```
      if run_error || converge_error 
        error =  if run_error == converge_error
                        Chef::Exceptions::RunFailedWrappingError.new(converge_error)
                      else
                        Chef::Exceptions::RunFailedWrappingError.new(run_error, converge_error) 
                      end
                  error.fill_backtrace
```

lamont-granquist: oh maybe nevermind though, i just read through this with my last comment in mind and yeah it looks like this all goes away because `:audit_mode == :disabled` now.

lib/chef/client.rb
```diff
@@ -756,50 +705,12 @@ def converge_and_save(run_context)
         begin
           save_updated_node
         rescue Exception => e
-          raise e if Chef::Config[:audit_mode] == :disabled
-          converge_exception = e
+          raise e
```
lamont-granquist: you probably still need `converge_exception = e`

lamont-granquist: okay i see this one as well now, "converge_exception = e" should not be hit if we raise.

lamont-granquist:
```
def converge_and_save(run_context)
  converge_exception = converge(run_context)
  unless converge_exception
    save_updated_node
  end
  converge_exception
end
```

i don't particularly like the way that the node save looks like it clobbers the shit out of the converge exception, but i think that is backwards compatible with what we've always done.

lib/chef/client.rb
lamont-granquist: (the next line does go away then since raise always raises and the `converge_exception = e` becomes dead code.

lamont-granquist:
For future reference:

converge_exception was only being used when audit mode was activated in order to save the converge_exception separate from any audit mode failures.

without audit mode it has always thrown and converge_exception was always nil

in converge_and_save without audit mode the actual exception thrown was just bypassing the node save through having the exception bubble all the way up.  the only point of the complicated code in the converge_and_save routine was for audit mode.

so all that complexity is now removed, converge_exception goes away, the fact that if we thrown in converge() we do not save in converge_and_save() is the correct behavior.

restored a test that we were losing in earlier revisions of this PR to check that on throwing an exception in converge() that we did not save the node (and validated that breaking that behavior broke the test).

lib/chef/client.rb
```diff
@@ -412,18 +361,6 @@ def rest_clean(client_name = node_name, config = Chef::Config)
                                                       signing_key_filename: config[:client_key], validate_utf8: false)
     end
 
-    # Resource reporters send event information back to the chef server for
-    # processing.  Can only be called after we have a @rest object
-    # @api private
-    def register_reporters
```
tas50: This just all goes away due to the new resource reporter work?

tas50:
Let's ship it and find out. We certainly made chef a lot easier to read through with this.

PR #7733, opened 2018-10-09
WIP: initial suppport for snap packages

smcavallo:
Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### Description
Snap packages are now native to ubuntu bionic.  Aws and other providers are using the snap package type as their default.  We're using amazon EKS and the kubeclient and kubelet service are deployed via snap.  
**We need a way to manage both the snap packages and the snap configs and services via chef.**

Please note -  this PR is a work in progress.   I'm hoping to get feedback on whether this is the right way to proceed before submitting an 'official' PR with everything implemented.

**I'm hoping someone can advise on the approach here and if it's following chef standards and test coverage, etc.**
If any feedback can be provided i'll take note and proceed with the updates and continuing work as appropriate.

### Issues Resolved

### Check List

- [X ] New functionality includes tests
- [X ] All tests pass


tas50:
Pinging @lamont-granquist here

lamont-granquist:
yeah, looks like this was based on the current yum/dnf provider code (with the python craziness and some rpm craziness removed), which is the correct guess on how to properly write package providers these days.

lamont-granquist:
travis and appveyor are angry about unit tests and chefstyle.  it looks good once those get cleaned up though.

lib/chef/provider/package/snap.rb
```diff
@@ -0,0 +1,318 @@
...
+require "chef/mixin/shell_out"
+require 'socket'
+require 'json'
+
+class Chef
+  class Provider
+    class Package
+      class Snap < Chef::Provider::Package
+        extend Chef::Mixin::Which
+        extend Chef::Mixin::ShellOut
```
lamont-granquist: that looks copied from the dnf provider and you don't need either of those because you're not using which/shell_out in your provides lines.

lib/chef/provider/package/snap.rb
```diff
@@ -0,0 +1,318 @@
...
+
+        allow_nils
+        use_multipackage_api
+
+        # Todo: support non-debian platforms
+        provides :package, platform_family: "debian"
```
lamont-granquist: oh yeah you can't do that.

you're wiring this up as the default package provider for everyone on debian here.

as it looks like snap do not wholesale replace apt_package on debian, you probably can't do this and need to delete this line.

lamont-granquist: (probably the source of most of the unit test issues as well)

lamont-granquist:
fixed and merged in #7999 

@smcavallo only merged for chef-15 right now, so you've got until april to debug it.

PR #7736, opened 2018-10-09
Add the timezone resource from the timezone_lwrp cookbook

tas50:
Adding this resource to core Chef with permission from @dragonsmith


kitchen-tests/cookbooks/end_to_end/recipes/default.rb
```diff
@@ -9,6 +9,8 @@
 
 apt_update
 
+timezone "UTC"
```
lamont-granquist: :+1:

lib/chef/resource/timezone.rb
```diff
@@ -0,0 +1,95 @@
...
+          package_name platform_family?("suse") ? "timezone" : "tzdata"
+        end
+
+        if node["init_package"] == "systemd"
+          # Modern Amazon, Fedora, CentOS, RHEL, Ubuntu & Debian
+          cmd_set_tz = "/usr/bin/timedatectl"
+          cmd_set_tz += " "
+          cmd_set_tz += "--no-ask-password"
+          cmd_set_tz += " "
+          cmd_set_tz += "set-timezone #{new_resource.timezone}"
```
lamont-granquist: that is crying out to be an array instead of an accumulated string

PR #7741, opened 2018-10-11
Tune appveyor skip to only version bumps

btm:
The Chef repo has Expeditor subscriptions to other repos which
automatically create a PR to bump versions of those dependencies in Chef
when they're released. We were skipping all commits in Appveyor by 'Chef
Expeditor' to avoid testing the automatic version bump commits that
happen on every merge to master.

This change will only skip the version bumps.


tas50:
@btm appveyor is quite red

PR #7752, opened 2018-10-16
Node Attributes: Build ImmutableMash properly in deep_merge!

### LINKED ISSUE #7738: Node Attribute Deep Merge Caching Is Susceptible to Caching an Out of Date Attribute Merge
## Description

Chef node attributes have various precedence levels and a deep merge is performed when node attributes are looked up.  The result of a deep merge is cached and a StateTracking mixin is used to invalid the cached result when attributes are mutated.  The crux of the issue is that the deep_merge! method in chef/node/attribute.rb can mutate the \_\_root__ StateTracking instance variable and this will break merge cache invalidation.  This doesn't show up often in practice because the \_\_root__ StateTracking variable is "fixed" by the [] method of StateTracking, so a fully-qualified lookup masks the issue (i.e. node.default['foo']['bar'] = 'something').  However, if, for example, you are holding on to a sub-level of the default attribute tree in a variable such that further assignments are done directly on that variable, the issue can show up.  Ex.

```
default['foo']['bar'].tap do |bar|
  # Other stuff up here...
  # Under certain cases this won't invalid the merge cache, see full replication steps
  # for specifics
  bar['hello'] = 'world'
end
```

## Chef Version

Reproducible starting on version 13.4.11

## Platform Version

CentOS 7

## Replication Case

I created a minimal replication example in the form of a script that can be placed in the bin folder of a chef source code checkout and run to reproduce and explain the issue.

As previously stated, the issue can be replicated in any version starting with 13.4.11, but for the purposes of this example, we will use the latest stable version at the time of writing: 14.5.33

```
# Get the code and get set up
git clone https://github.com/chef/chef.git
cd chef
git checkout v14.5.33
bundle install
bundle exec rake install
```

Here's a script that demonstrates the bug and outputs information about what's going on as it does stuff:
```
#!/usr/bin/env ruby

require "rubygems"
$:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))

require "chef"
require "chef/node/attribute"

puts '''
default = {
    "key1" => {
        "key2" => {
            "key3" => true
        }
    }
}
'''

default = {
    "key1" => {
        "key2" => {
            "key3" => true
        }
    }
}

puts '''
env_default = {
    "key1" => {
        "key2" => {
            "key4" => false
        }
    }
}
'''

env_default = {
    "key1" => {
        "key2" => {
            "key4" => false
        }
    }
}

# Keep this simple -- we only have defaults
puts "\n"
puts "a = Chef::Node::Attribute.new({}, default, {}, {})"
a = Chef::Node::Attribute.new({}, default, {}, {}) 
puts "a.env_default = env_default"
a.env_default = env_default

puts "\nFirst let's proxy some relevant StateTracking methods so we can log when they are called"

# Logs a call to the reset_cache method
module ResetCacheLogger
  def reset_cache(path = nil)
    puts "=== ResetCacheLogger ==="
    puts "Reset cache called for path #{path == nil ? 'nil': path}"
    super   
  end 
end

# Logs a call the StateTracking __root__ setter if the set changes the value
module RootChangeLogger
  def __root__=(root)
    if root != __root__
        puts "=== RootChangeLogger ==="
        puts "__root__ on #{__path__} is getting changed from object_id #{__root__.object_id} to #{root.object_id}!\nStacktrace: #{caller.join("\n")}"
    end 
    super   
  end 
end

# Stick our logger modules in where we want them
puts "a.default['key1'].__root__.extend(ResetCacheLogger)"
a.default['key1'].__root__.extend(ResetCacheLogger)
puts "a.default['key1']['key2'].extend(RootChangeLogger)"
a.default['key1']['key2'].extend(RootChangeLogger)

puts "\nWe look up a merged value to make the 'key1' subtree merge get cached..."
puts "Notice that this invokes our logger for the __root__ StateTracking variable getting changed on a.default['key1']['key2'] -- that seems weird"
puts "a['key1']"
puts
a['key1']

puts "\nNow we will use a.default['key1']['key2'].tap to add some default values. Note: This is going to reset the __root__ on a.default['key1']['key2']"
puts "to the proper value since the [] method on StateTracking sets __root__ appropriately as you navigate down the tree"
puts "a.default['key1']['key2'].tap do |key2|"
puts

a.default['key1']['key2'].tap do |key2|
    puts "\nLet's invalid the merge cache by setting a default value..."
    puts "key2['foo'] = 'bar'"
    puts
    key2['foo'] = 'bar'

    puts "\nGreat, now let's look up the value we just set using the merged attributes.  This will result in the"
    puts "'key1' subtree merge result getting cached again, but it will also mutate the __root__ StateTracking variable... this seems bad."
    puts "on a.default['key1']['key2']"
    puts
    puts "\nWe expect a['key1']['key2']['foo'] to return bar. Actual result: #{a['key1']['key2']['foo']}"

    puts "\nWe will now set a default again, and we expect it to result in a call to cache_reset, but it doesn't"
    puts "The reason is that __root__ got redefined when we did the attribute merge above.  In fact, cache_reset is getting"
    puts "called, but on some random VividMash instance that gets created during the merge"
    puts "key2['foo'] = 'bar2'"
    key2['foo'] = 'bar2'

    puts "\nNow when we look up the default value we just set using the merged attributes, we get the wrong result"
    puts "since the previously cached merge was never cleared"
    puts "\nWe expect a['key1']['key2']['foo'] to return bar2. Actual result: #{a['key1']['key2']['foo']}"
end
```
The easiest thing to do is to copy the above script into the bin folder of your chef checkout and call it attribute-bug.rb.  You then run it and it prints the relevant code it is executing and explains what exactly is going wrong:
```
cd bin
# Create attribute-bug.rb with the contents above
bundle exec attribute-bug.rb | less
```

lamont-granquist:
closes #7738

The root cause here is that there's a difference between doing

Chef::Node::VividMash.new( <another VividMash> )

And:

Chef::Node::ImmutableMash.new( <a VividMash> )

The former short circuits and does no work in convert_value.  The
latter will not short circuit and does the proper work to dup
and convert the value argument.

Since we build an ImmutableMash now, we do not need to wrap it
with the extra immutablize() call.

This should be perf neutral or very, very slightly better perf.

### DIFF
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -403,11 +403,11 @@ class Chef
       end
 
       def combined_override(*path)
-        immutablize(merge_overrides(path))
+        merge_overrides(path)
       end
 
       def combined_default(*path)
-        immutablize(merge_defaults(path))
+        merge_defaults(path)
       end
 
       def normal_unless(*args)
@@ -599,9 +599,9 @@ class Chef
         # In all other cases, replace merge_onto with merge_with
         else
           if merge_with.kind_of?(Hash)
-            Chef::Node::VividMash.new(merge_with)
+            Chef::Node::ImmutableMash.new(merge_with)
           elsif merge_with.kind_of?(Array)
-            Chef::Node::AttrArray.new(merge_with)
+            Chef::Node::ImmutableArray.new(merge_with)
           else
             merge_with
           end
--- a/spec/unit/node_spec.rb
+++ b/spec/unit/node_spec.rb
@@ -1818,4 +1818,37 @@ describe Chef::Node do
       expect(node["a"]["key"]).to eql(1)
     end
   end
+
+  describe "when abusing the deep merge cache" do
+    # https://github.com/chef/chef/issues/7738
+    it "do not corrupt VividMashes that are part of the merge set and not the merge_onto set" do
+      # need to have a merge two-deep (not at the top-level) between at least two default (or two override)
+      # levels where the lowest priority one is the one that is going to be corrupted
+      node.default["foo"]["bar"]["baz"] = "fizz"
+      node.env_default["foo"]["bar"]["quux"] = "buzz"
+      node.default["foo"]["bar"].tap do |bar|
+        bar["test"] = "wrong"
+        # this triggers a deep merge
+        node["foo"]["bar"]["test"]
+        # this should correctly write and dirty the cache so the next read does another deep merge on the correct __root__
+        bar["test"] = "right"
+      end
+      expect(node["foo"]["bar"]["test"]).to eql("right")
+    end
+
+    it "do not corrupt VividMashes that are part of the merge set and not the merge_onto set (when priorities are reversed)" do
+      # need to have a merge two-deep (not at the top-level) between at least two default (or two override)
+      # levels where the *HIGHEST* priority one is the one that is going to be corrupted
+      node.env_default["foo"]["bar"]["baz"] = "fizz"
+      node.default["foo"]["bar"]["quux"] = "buzz"
+      node.env_default["foo"]["bar"].tap do |bar|
+        bar["test"] = "wrong"
+        # this triggers a deep merge
+        node["foo"]["bar"]["test"]
+        # this should correctly write and dirty the cache so the next read does another deep merge on the correct __root__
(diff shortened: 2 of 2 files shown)


lamont-granquist:
facebook validated no perf hit from this.

PR #7755, opened 2018-10-16
Run more Travis tests on Ruby 2.5.1

tas50:
There's no reason to keep executing our TK tests against ruby 2.4.


lamont-granquist:
and yeah, the TK rspec runs can also be bumped.  we should catch 2.4.0 bugs by travis+appveyor.  the intent of the TK rspec runs is to be like manhattan so we get as close to possible to testing on what we ship (the last -current's build omnibus ruby which is 2.5.x)

PR #7758, opened 2018-10-17
Initial target_mode implementation

btm:
https://github.com/chef/chef-rfc/pull/324

Spike on target mode support.
Adds support for using train to configure remote nodes without installing Chef remotely.
Supports the execute resource and guards.


chef-config/lib/chef-config/config.rb
```diff
@@ -607,7 +623,15 @@ def self.init_openssl
...
-    default(:client_key) { chef_zero.enabled ? nil : platform_specific_path("/etc/chef/client.pem") }
+    default(:client_key) do
+      if chef_zero.enabled
+        nil
+      elsif target_mode?
+        platform_specific_path("/etc/chef/#{target_mode.host}/client.pem")
+      else
+        platform_specific_path("/etc/chef/client.pem")
+      end
+    end
```
lamont-granquist: so i forgot that the credentials stuff was all knife-side and this is going to be entirely client-side.

it still feels like we should probably support credentials in the client here to have one way of doing things, rather than two.

lib/chef/client.rb
```diff
@@ -593,6 +613,39 @@ def save_updated_node
...
+      # Train handles connection retries for us
+      @train_connection = Train.create(protocol, train_config).connection
+      train_connection.wait_until_ready
+    rescue SocketError => e # likely a dns failure, not caught by train
+      e.message.replace "Error connecting to #{train_connection.uri} - #{e.message}"
+      raise e
+    rescue Train::PluginLoadError
+      logger.error("Invalid target mode protocol: #{protocol}")
+      exit(false)
+    end
```
lamont-granquist: sort of wonder here about the inherent tight coupling here between one train connection and one client connection.

particularly when we start talking next-gen chef-provisioning-like ideas, we may want to have this more loosely coupled.

i'm leaning towards run_context.train_connection should be the primary object, you can have a delegator here for train_connection -> run_context.train_connection.

then access to train_connection should probably have a lazy initializer so that when it is hit the first time it calls setup_train in a method off of the run_context.

you should probably have a chef class like Chef::TrainThing which wraps a single train connection which is the thing which actually hangs off of the run_context, then the run_context just calls Chef::TrainThingy.build_connection() to get a new object -- then the run_context and the client go back to "knowing" very little about train.

lib/chef/dsl/universal.rb
```diff
@@ -47,6 +48,7 @@ module Universal
       include Chef::Mixin::PowershellExec
       include Chef::Mixin::PowershellOut
       include Chef::Mixin::ShellOut
+      include Chef::Mixin::TrainOrShell
```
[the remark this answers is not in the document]
btm: yup. pushed.

lib/chef/client.rb
btm: We do need a train connection before the run context is created because we use it to go get ohai-like-data at the same time as ohai would, but I suppose there's no reason to keep that particular train connection for later. some day we might have ohai go out and make that connection instead, so could split that now too.

lamont-granquist: I think I'm going to change the run_context in Chef 15 to get created way the hell early, and decouple the creation of the run_context from the compilation of the cookbooks, so it just becomes the global-ish state context of the whole run.  keep running into no way of threading state through the actual whole run being a problem.

lib/chef/mixin/train_or_shell.rb
```diff
@@ -0,0 +1,74 @@
...
+require "ostruct"
+
+class Chef
+  module Mixin
+    module TrainOrShell
+      #
+      # Train #run_command returns a Train::Extras::CommandResult which
+      # includes `exit_status` but Mixlib::Shellout returns exitstatus
+      # This wrapper makes the result look like Mixlib::ShellOut to make it
+      # easier to swap providers from #shell_out to #trian_or_shell
```
lamont-granquist: trian

lib/chef/node_map.rb
```diff
@@ -269,11 +270,23 @@ def matches_version_list?(node, filters, attribute)
...
 
+    # Succeeds if:
+    # - we are in target mode, and the target_mode filter is true
+    # - we are not in target mode, and the target_mode filter is not false
+    def matches_target_mode?(filters)
+      if Chef::Config.target_mode?
+        filters[:target_mode] == true
+      else
+        !(filters[:target_mode] == false)
+      end
```
lamont-granquist: `!!filters[:target_mode] == !!Chef::Config.target_mode?` is probably a bit clearer that you're checking boolean equality while not caring about strict true/false values and just truthy/falsey

lib/chef/run_context.rb
```diff
@@ -579,6 +580,20 @@ def reboot_requested?
...
+    #   our train transport.
+    #
+    def transport_connection
+      if @transport.nil?
+        @transport = Chef::Transport.build_transport(logger)
+        @transport.connection
+      else
+        @transport.connection
+      end
+    end
```
lamont-granquist: this should probably be:

```
def transport
  @transport ||= Chef::Transport.build_transport(logger)
end

def transport_connection
  transport.connection
end
```

lib/chef/transport.rb
```diff
@@ -0,0 +1,140 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef-config/mixin/credentials"
+require "train"
+
+class Chef
+  class Transport
```
lamont-granquist: since this looks like a pile of helper class methods and you don't create an instance of this, it should probably be a module instead of a class (and then the methods no longer being class methods).

since it is train-specific it might be better named Chef::TrainTransport as well?

i was gonna suggest injecting the methods directly into the object but i think I like the inherent message to "go look in another file" with writing `Chef::TrainTransport.build_transport` rather than having `include Chef::TrainTransport .....  @transport = build_train_transport` which gives the user no clue of where to actually find build_train_transport (and you'd have to rename it because `build_transport` would then also be too generic...)

tas50:
@btm can you rebase this?

btm:
@tas50 this is rebased.

@lamont-granquist you have since done the early run context work, I think I refactored to account for that.

btm:
This is going to need to be refactored to share more train bits with the new bootstrap I think, but given we need to figure out how we're going to refactor that still I don't know it's worth thinking about today.

chef.gemspec
```diff
@@ -38,6 +38,8 @@ Gem::Specification.new do |s|
   s.add_dependency "diff-lcs", "~> 1.2", ">= 1.2.4"
   s.add_dependency "ffi-libarchive"
   s.add_dependency "chef-zero", ">= 14.0.11"
+  s.add_dependency "train-core"
```
tas50: Do we want to keep this since we bring this in already via chef-core?

lib/chef/application/client.rb
```diff
@@ -298,6 +298,15 @@ class Chef::Application::Client < Chef::Application
     description: "Use cached cookbooks without overwriting local differences from the server",
     boolean: false
 
+  option :target,
+    short: "-t TARGET",
+    long: "--target TARGET",
+    description: "Target the Chef Client run against a remote system or device",
```
tas50: This should utilize the branding values we merged in

lib/chef/client.rb
```diff
@@ -151,8 +152,11 @@ def initialize(json_attribs = nil, args = {})
       @json_attribs = json_attribs || {}
       @logger = args.delete(:logger) || Chef::Log.with_child
 
-      @ohai = Ohai::System.new(logger: logger)
-
+      @ohai = if Chef::Config.target_mode?
+                OpenStruct.new(data: Mash.new)
+              else
+                Ohai::System.new(logger: logger)
+              end
```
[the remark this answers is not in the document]
tas50: I'd rather we let ohai abstract the details of how this happens

chef.gemspec
btm: I could go any way. I don't know of any problem caused by leaving it in, or taking it out.

When we change this implementation to use the same library that is chef-core today, we would surely remove this. But on its own, presuming that we'll get a dependency we require from another dependency doesn't feel right.

lib/chef/client.rb
```diff
@@ -555,6 +570,32 @@ def save_updated_node
...
+      ohai.data[:fqdn] = if transport_connection.respond_to?(:hostname)
+                           transport_connection.hostname
+                         else
+                           Chef::Config[:target_mode][:host]
+                         end
+      if transport_connection.respond_to?(:os)
+        ohai.data[:platform] = transport_connection.os.name
+        ohai.data[:platform_version] = transport_connection.os.release
+        ohai.data[:os] = transport_connection.os.family_hierarchy[1]
+        ohai.data[:platform_family] = transport_connection.os.family
```
tas50: I'm not sure if we want to do this actually. These platforms aren't the same and populating attributes normally handled by ohai with data from train is going to cause a bad user experience. Many of the platforms are not the same so you'd write a cookbook expecting one behavior and you'd get another.

tas50: Take a peak at the work that @burtlo did to translate train to ohai values.

lib/chef/application/client.rb
btm: done.

lib/chef/client.rb
btm: I think we have to do this for the platform mapping to not blow up on us, probably other parts of code expect `node[:platform]` to be set. Otherwise `node[:train][:os][:name]` would be logical.

Of course I agree we still need to reconcile ohai vs train data here. or https://github.com/burtlo/ohbye, or similar.

We've got a ways to go before you could use the same cookbooks for a target mode and normal node though.

lib/chef/train_transport.rb
```diff
@@ -50,26 +50,15 @@ def self.load_credentials(profile)
...
-    end
+			fqdn.split(".").reduce(hash) { |h, k|
+				v = h[k]
+				if Hash === v then
+					v
+				else
+					break false
+				end
+			}
+		end
```
zenspider: tabs?? 😛

PR #7767, opened 2018-10-23
Fix registry key bug when sensitive is true

josh-barker:
### Description

Fixes bug with `registry_key` resource when sensitive is true

Tested against Chef 14.5.33 on Windows 2012r2 & 2016

The following example results in `*sensitive value suppressed*` being written to the registry value, instead of `some value`

```ruby
registry_key "HKEY_LOCAL_MACHINE\\path-to-key\\Policies\\System\Demo" do
  values [{
    name: 'example',
    type: :string,
    data: 'some value'
  }]
  sensitive true
  action :create
end
```

### Issues Resolved

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/registry_key.rb
+++ b/lib/chef/provider/registry_key.rb
@@ -126,16 +126,22 @@ class Chef
               value[:data] = value[:data].to_i
             end
             unless current_value[:type] == value[:type] && current_value[:data] == value[:data]
-              converge_by_value = value
-              converge_by_value[:data] = "*sensitive value suppressed*" if new_resource.sensitive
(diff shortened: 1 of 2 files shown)


tas50:
@josh-barker can you squash this down to a single commit to remove the DCO issue with the merge commit

stuartpreston:
@josh-barker Thanks very much for identifying and fixing this issue!

tas50:
Unrelated travis failures that I'm digging into

josh-barker:
You're welcome @stuartpreston :-)

PR #7780, opened 2018-10-25
Add some retry/delay in HTTP functional tests

schisamo:
We are seeing some failures when running the functional tests on our new
Anka-based macOS testers. This small tweaks to the `Chef::Config` used
in the specs fix the issues.

Signed-off-by: Seth Chisamore <schisamo@chef.io>

### DIFF
--- a/spec/functional/http/simple_spec.rb
+++ b/spec/functional/http/simple_spec.rb
@@ -32,8 +32,8 @@ describe Chef::HTTP::Simple do
 
   before(:each) do
     Chef::Config[:rest_timeout] = 2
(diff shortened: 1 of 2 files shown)


lamont-granquist:
for my future memory:  this is likely papering over some issue with tinyserver, but i can't figure out what.  tests that use tiny server before the failing tests will succeed.  similar tests will pass.  remote_file was failing on a compressed file, http::simple was failing on an uncompressed file.  there's no obvious common codepath which broke similar use cases.  i'm not sure i even know what to debug, and we needed to ship this...

PR #7782, opened 2018-10-25
Fix locale on RHEL 6 / Amazon Linux

tas50:
There was a double source in the command, which would fail.


tas50:
We need to pull the chefspec stuff into chef itself so we can test our core resources

PR #7790, opened 2018-10-26
Remove legacy require_recipe DSL method

tas50:
We've been warning on this one for YEARS and we have a Foodcritic rule
for it as well, which has been around since before I was at Chef.


lamont-granquist:
no unit tests to remove?

tas50:
Did we right unit tests in 2009? :trollface:

PR #7792, opened 2018-10-26
Remove cookbook merging/shadowing from the cookbooker loader

lamont-granquist:
This has been deprecated since forever.

### DIFF
--- a/spec/unit/cookbook_loader_spec.rb
+++ b/spec/unit/cookbook_loader_spec.rb
@@ -24,7 +24,6 @@ describe Chef::CookbookLoader do
   end
   let(:repo_paths) do
     [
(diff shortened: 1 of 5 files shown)


spec/unit/cookbook_loader_spec.rb
```diff
@@ -187,7 +152,7 @@ def full_paths_for_part(cb, part)
 
     let(:repo_paths) do
       [
-        File.join(CHEF_SPEC_DATA, "kitchen"),
+#        File.join(CHEF_SPEC_DATA, "kitchen"),
```
tas50: ?

lamont-granquist: ah right that was starting off as just a test

danielsdeleo:
Yaaaaas

PR #7793, opened 2018-10-27
shell_out auto-timeout still needs to be restricted to only providers

lamont-granquist:
fixes rspec tests with a `let(:new_resource)` from blowing up

still have to do the janky string thing because descendents tracker needs to get nuked...


lamont-granquist:
got appveyor and the unit tests fully green

PR #7794, opened 2018-10-27
Refactor Cookbook loader logic now that we don't support merging

lamont-granquist:
this can be cleaned up by the removal of the multiple loaders per
cookbook (since shadowing / merging is gone) and just because the
logic was getting really weird.


lamont-granquist:
@tas50 this is now almost approaching readability

tas50:
Seems fairly straight forward now

lamont-granquist:
yeah and i'm working on a little bit more cleanup and docs for another PR

PR #7797, opened 2018-10-27
Add macOS support to the timezone resource

tas50:
Moving towards a world where we properly load the current state. Adds a method that grabs the current timezone and properly fails with a warning message if we don't have the privs we need. Using converge_by and shell_out to set the timezone so we get a nice clean converge messaging vs. execute resources. Also improves the description to give a link to where users can find all the TZ options.

### DIFF
--- a/lib/chef/resource/timezone.rb
+++ b/lib/chef/resource/timezone.rb
@@ -24,7 +24,7 @@ class Chef
     class Timezone < Chef::Resource
       resource_name :timezone
 
-      description "Use the timezone resource to change the system timezone."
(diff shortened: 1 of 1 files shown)


lib/chef/resource/timezone.rb
```diff
@@ -50,38 +53,63 @@ class Timezone < Chef::Resource
...
-            owner "root"
-            group "root"
-            mode "0644"
-            action :create
-            content %{ZONE="#{new_resource.timezone}"\nUTC="true"\n}
-          end
+        else
+          case node["platform_family"]
+          # Old version of RHEL < 7 and Amazon 201X
+          when "rhel", "amazon"
```
lamont-granquist: fedora?

tas50: Pre-systemd so no need for Fedora here

PR #7798, opened 2018-10-27
Require mixin::shellout where we use it

tas50:
We're extending this in a few places where we didn't actually require
it. We should probably just require it here.


lib/chef/knife/supermarket_install.rb
```diff
@@ -20,6 +20,7 @@
 require "chef/exceptions"
 require "shellwords"
 require "mixlib/archive"
+require "chef/mixin/shell_out"
```
lamont-granquist: this one has the lazy-load dep down below it though.

Issue #7708, opened 2018-09-30
dmg_package: Can't install dmg package with EULA


jorgeLuizChaves:
## Description
Given a dmg package that needs to accept EULA to install dmg_package can't install it 

## Chef Version
Chef: 14.4.56

## Platform Version
MacOS High Sierra 10.13.6

## Replication Case
```ruby
dmg_package 'Evernote' do
    dmg_name 'evernote'
    source 'https://cdn1.evernote.com/mac-smd/public/Evernote_RELEASE_7.5_457109.dmg'
    accept_eula true
end
```

## Client Output

```
  * dmg_package[Evernote] action install
    * remote_file[/Users/jorgeluiz/.chef/cache/evernote.dmg - Evernote] action create (up to date)
    * ruby_block[attach /Users/jorgeluiz/.chef/cache/evernote.dmg] action run

      ================================================================================
      Error executing action `run` on resource 'ruby_block[attach /Users/jorgeluiz/.chef/cache/evernote.dmg]'
      ================================================================================

      Mixlib::ShellOut::ShellCommandFailed
      ------------------------------------
      Expected process to exit with [0], but received '1'
      ---- Begin output of echo Y | PAGER=true hdiutil attach  '/Users/jorgeluiz/.chef/cache/evernote.dmg' -mountpoint '/Volumes/Evernote' -quiet ----
      STDOUT:
      STDERR:
      ---- End output of echo Y | PAGER=true hdiutil attach  '/Users/jorgeluiz/.chef/cache/evernote.dmg' -mountpoint '/Volumes/Evernote' -quiet ----
      Ran echo Y | PAGER=true hdiutil attach  '/Users/jorgeluiz/.chef/cache/evernote.dmg' -mountpoint '/Volumes/Evernote' -quiet returned 1

      Resource Declaration:
      ---------------------
      # In /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.4.56/lib/chef/resource/dmg_package.rb

      114:           ruby_block "attach #{dmg_file}" do
      115:             block do
      116:               cmd = shell_out("hdiutil imageinfo #{passphrase_cmd} '#{dmg_file}' | grep -q 'Software License Agreement: true'")
      117:               software_license_agreement = cmd.exitstatus == 0
      118:               raise "Requires EULA Acceptance; add 'accept_eula true' to package resource" if software_license_agreement && !new_resource.accept_eula
[...]

PR #7802, opened 2018-10-28
Multiple fixes to dmg_package

tas50:
1) Remove the "installed" property that was being used to track state and instead properly load the current_value
2) Set desired_state false on several properties that aren't part of the desired state
3) Add missing periods to a few descriptions
4) Remove declare_resource where it's not necessary
5) Fix dmg_name to properly take on app as the default value
6) Move some of the logic into helper methods that can be tested at some point
7) Move some values into default values instead of variables we set later that contained the same logic
8) Improved the error message you get if your package requires setting the EULA
9) Don't use grep to find the string that indicates the EULA is required
10) Use the full path to hdiutil so we're not depending on the PATH variable being correct
11) Accepting the EULA now works

### DIFF
--- a/lib/chef/resource/dmg_package.rb
+++ b/lib/chef/resource/dmg_package.rb
@@ -71,9 +71,9 @@ class Chef
                description: "Specify whether to accept the EULA. Certain dmgs require acceptance of EULA before mounting.",
                default: false, desired_state: false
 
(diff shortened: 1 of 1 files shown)


tas50:
Fixes https://github.com/chef/chef/issues/7708

PR #7806, opened 2018-10-29
Added windows support for the timezone resource

username-is-already-taken2:
Signed-off-by: username-is-already-taken2 <digitalgaz@hotmail.com>

### Description

I have added windows support to the timezone resource.  I have based this off of #7797 

### Issues Resolved

Don't know if anyone has requested this yet.

### Check List

- [x] New functionality includes tests (existing test seem to cover the resource)
- [ ] All tests pass - Really hard to validate this on a windows box can't get rake to pass cleanly on master.  I'll build an debian vm and try with that
- [ ] RELEASE\_NOTES.md has been updated if required (not required for bugfixes, required for API changes) - Happy to update if you think its required.


lib/chef/resource/timezone.rb
```diff
@@ -95,12 +95,18 @@ class Timezone < Chef::Resource
...
+                shell_out!("tzutil /s \"#{new_resource.timezone}\"")
+              end
+            end
           end
         end
       end
 
       action_class do
-        # detect the current TZ on darwin hosts
+        # detect the current TZ on darwin hosts & windows hosts
```
tas50: This comment was actually for the macos method. Can you add a comment and YARD tag to the windows method below as well

lib/chef/resource/timezone.rb
```diff
@@ -112,6 +118,14 @@ def current_darwin_tz
             /Time Zone: (.*)/.match(tz_shellout.stdout)[1]
           end
         end
+        def current_windows_tz
+          tz_shellout = shell_out!("tzutil /g")
+          if /is not recognized as an internal/.match?(tz_shellout.stderr)
```
stuartpreston: If you use the shell_out without the !
```
tz_shellout = shell_out("tzutil /g")
```
Then you can test the exitstatus rather than raising directly
```
if tz_shellout.exitstatus == 1
```
This will save you matching the stderr from the command processor.

username-is-already-taken2:
@tas50 - I'm sorry for being slow, I've been away.  You're both were very kind to give me a good steer on what you needed from me, I just need to cover this.

I think part of the problem was that I copied someone else's homework :) So if I'm honest I didn't know what the ! is doing on the shell_out (it seemed work :) or what YARD tag is?

I should have managed this PR better but I will endeavor to look these things up and get this updated.

lib/chef/resource/timezone.rb
username-is-already-taken2: thanks for the advice, I've amended it, is that OK?

lib/chef/resource/timezone.rb
username-is-already-taken2: Hi there I've put your tag back the way it was and added one to mine, is that OK?

tas50:
@username-is-already-taken2 You have a few chefstyle warnings here. You can run chefstyle -a . to fix that.

username-is-already-taken2:
@tas50 school boy error sorry, should have covered that off pushing

username-is-already-taken2:
Running rspec locally on my machine it passes my test, not sure how to fix the travis error.

Vasu1105:
> Running rspec locally on my machine it passes my test, not sure how to fix the travis error.

Making your spec to run only on windows will most probably fix the travis.

lib/chef/resource/timezone.rb
username-is-already-taken2: @stuartpreston hi there, I added a test to cover off windows.  Was that what you had in mind?

Thanks for your help

tas50:
@username-is-already-taken2 Thanks for keeping at it and working with us here. This looks good to me now.

PR #7811, opened 2018-10-29
Automated resource documentation improvements

tas50:
Backport from Chef 15 master:

Add new DSL for resource properties:
 - skip_docs: don't document a particular property
 - default_description: describe a default value since many are calculated

This also syncs some copy edits from the docs site.


stuartpreston:
Travis is failing with
```

Bundler could not find compatible versions for gem "ohai":
  In Gemfile:
    ohai
    chef was resolved to 14.6.47, which depends on
      ohai (~> 14.0)
```

PR #7819, opened 2018-10-30
Fix testing / installing on SLES 15

tas50:
Make sure we identify SLES 15 as suse for our zypper tests
Make sure we don't remove the symlinks on SLES when doing an upgrade or the new release will be broken

### DIFF
--- a/omnibus/package-scripts/angrychef/postrm
+++ b/omnibus/package-scripts/angrychef/postrm
@@ -17,8 +17,8 @@ is_darwin() {
 
 is_suse() {
   if [ -f /etc/os-release ]; then
-    source /etc/os-release
-    [ $ID_LIKE == "sles" ] || [ $ID_LIKE == "suse" ]
+    . /etc/os-release
+    [ "$ID_LIKE" = "sles" ] || [ "$ID_LIKE" = "suse" ]
   else
     [ -f /etc/SuSE-release ]
   fi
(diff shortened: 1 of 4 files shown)


tas50:
Apparently I'm too young to edit this file so calling on @btm and @lamont-granquist

omnibus/package-scripts/angrychef/postrm
```diff
@@ -11,11 +11,19 @@ is_smartos() {
...
 
-is_darwin()
-{
+is_darwin() {
   uname -v | grep "^Darwin" 2>&1 >/dev/null
 }
 
+is_suse() {
+  if [ -f /etc/os-release ]; then
+    source /etc/os-release
```
btm: `source` is a bashism, you want `. /etc/os-release`

omnibus/package-scripts/angrychef/postrm
```diff
@@ -11,11 +11,19 @@ is_smartos() {
...
+    [ $ID_LIKE == "sles" ] || [ $ID_LIKE == "suse" ]
```
btm: `==` is a bashism too. you want single `=`, i.e.
```
     [ $ID_LIKE = "sles" ] || [ $ID_LIKE = "suse" ]
```

btm: edit about the quotes around the variable --^

PR #7840, opened 2018-11-01
Chef 15 node attribute array fixes

lamont-granquist:
Doing something like:

```
node.default["foo"] = []
node.default["foo"] << { "bar" => "baz }
```

Would result in a Hash instead of a VividMash inserted into the
AttrArray, so that:

```
node.default["foo"][0]["bar"] # gives the correct result
node.default["foo"][0][:bar]  # does not work due to the sub-Hash not
                              # converting keys
```

This fixes that behavior, but is likely a breaking change to someone
who expects to ingest hashes with symbol keys and get back out symbol
keys, they will now get converted to strings.  This breaking change
is fully 100% intentional.  This makes the node attribute structure
more consistent and predictable.

### DIFF
--- a/lib/chef/node/attribute_collections.rb
+++ b/lib/chef/node/attribute_collections.rb
@@ -18,6 +18,9 @@
 
 require "chef/node/common_api"
 require "chef/node/mixin/state_tracking"
+require "chef/node/mixin/immutablize_array"
+require "chef/node/mixin/immutablize_hash"
+require "chef/node/mixin/mashy_array"
 
 class Chef
   class Node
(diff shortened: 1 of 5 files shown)


tas50:
Just a note I wouldn't even put this on the list of breaking changes. It's a bug fix in my mind. There is some slim chance somewhere that a person is relying on this behavior, but they were relying on a bug. We shouldn't freak people out by adding this to the list of breaking changes.

lamont-granquist:
Yeah but prior experience is that we've fixed very similar things and gotten bug reports, and complaints that we're changing things in stable versions.

PR #7841, opened 2018-11-01
Remove knife user support for open source Chef Server < 12

tas50:
We shipped Chef 12 server on 2014-11-25. We've maintained backwards compatibility here by shipping double plugins, constantly checking for Chef 11 and then rewriting passed args. We deprecated it with Chef 14. This fully removes support. chef-client will continue to work on chef 11, but users cannot be managed with knife from a workstation running the latest version of Chef/ChefDK.


markan:
Seems reasonable to me.

PR #7855, opened 2018-11-02
Remove the check for nil code property in the script provider

tas50:
new_resource.code comes from the execute resource and it's a required property there. This can never fire.


lamont-granquist:
no, pretty certain that does fire.

lamont-granquist:
oh i think i see, that's in the provider and it was added a long time ago, and the resource was fixed to be required back in chef-13

PR #7867, opened 2018-11-04
Add more validation_messages to properties

### LINKED ISSUE #7059: Add validation messages anywhere we do property validations
We can now give the user much more information when validation of properties in resources fail by using validation messages. We should add these fields to all existing resources to help out new users.

tas50:
Provide users with more helpful messages

Fixes #7059


tas50:
Pinging the @chef/docs-team since some of this will change the docs site.

Issue #7468, opened 2018-07-14
Knife status should return ipv4/fqdn cloud options not EC2 specific data


tas50:
We have a special case that only works for ec2 where in if you run knife status --long you get back the ipv4/fqdn from the ec2 ohai node data. It makes far more sense to use the cloud plugin here so we get the correct fqdn/ipv4 address for any cloud.

It's a breaking change so we'd need to do it in Chef 15

PR #7882, opened 2018-11-06
Update knife status --long to use cloud attributes not ec2 specific attributes

tas50:
This way we get the public IP / hostname on Azure, GCE, DigitalOcean, Softlayer, etc.


tas50:
Fixes https://github.com/chef/chef/issues/7468

PR #7885, opened 2018-11-07
Update rubygems to use --no-document so we can support Ruby 2.6

tas50:
Ruby 2.6 includes rubygems 3.0 which makes --no-ri and --no-rdoc hard fail. Rubygems 2.0 introduced --no-document so we're good here as it came out Feb 2013.

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -547,9 +547,9 @@ class Chef
           end
           src_str = src.empty? ? "" : " #{src.join(" ")}"
           if !version.nil? && !version.empty?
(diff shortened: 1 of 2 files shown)


lamont-granquist:
RHEL6 is rubygems 1.3.7

lamont-granquist:
We probably need to hit `"#{gem_binary_path} --version"` and parse it and switch the behavior

tas50:
But we wouldn't support the version of ruby in RHEL 6 anyways right. We're going to be requiring 2.4+ on Chef 14 and 2.5+ on Chef 15.

lamont-granquist:
gem_package installs into the system ruby

i think for that as well you have to use the gem binary and these codepaths because you can't use the rubygems environment that chef is running out of, but my recollection may be a bit hazy on that...

PR #7890, opened 2018-11-07
fix whitespace in node attributes [ci skip]

lamont-granquist:
no idea how this ever passed chefstyle, but the multi-line array loop was confusing everything

### DIFF
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -78,113 +78,117 @@ class Chef
         :@force_override,
       ].freeze
 
-      [:all?,
-       :any?,
-       :assoc,
-       :chunk,
-       :collect,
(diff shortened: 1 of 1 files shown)


lib/chef/node/attribute.rb
```diff
@@ -625,11 +629,11 @@ def hash_only_merge!(merge_onto, merge_with)
           end
           merge_onto
 
-        # If merge_with is nil, don't replace merge_onto
+          # If merge_with is nil, don't replace merge_onto
         elsif merge_with.nil?
           merge_onto
 
-        # In all other cases, replace merge_onto with merge_with
+          # In all other cases, replace merge_onto with merge_with
```
tas50: isn't this part wrong now?

PR #7892, opened 2018-11-07
Chef-15: add nillability to attribute deep merging

lamont-granquist:
Changes to consistently uses the "NIL" object internally to mean "not present"
instead of abusing nil for that purpose.

So now "nil means nil" and we can deep merge that so it will effectively
override lower precedence levels if you write a nil.


tas50:
Add a rough description of this to the breaking changes section of the release notes so we don't forget about it.

PR #7896, opened 2018-11-08
Fix the knife integration spec timeouts

lamont-granquist:
rely on killing the thread to terminate the server, and rely on
thread#join to correctly signal to the main thread that the
server thread is dead, don't leak the server object out of the
helper.


lamont-granquist:
just the fedora failure

PR #7914, opened 2018-11-11
use --no-tty during apt-keys fro gpg - fixes #7913

### LINKED ISSUE #7913: apt-keys call within apt_repository broken for debian >9.3
when installing a repo using apt_repository from debian >9.3 it will fail since tty is missing

using this kitchen setup for reproducing this:

```yaml
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  
verifier:
  name: inspec

platforms:
  - name: debian/stretch64
    driver:
      box: debian/stretch64
      box_version: "9.3.0"
  - name: debian/stretch64
    driver:
      box: debian/stretch64
      box_version: "9.4.0"
  - name: debian/stretch64
    driver:
      box: debian/stretch64
      box_version: "9.5.0"

suites:
  - name: default
    run_list:
    - recipe[chef-apt-docker::default]
```

For 9.3 it will work, for 9.4 / 9.5 it will fail with

```
 ---- End output of apt-key adv --recv --keyserver hkp://keyserver.ubuntu.com:80 9DC858229FC7DD38854AE2D88D81803C0EBFCD88 ----
       Ran apt-key adv --recv --keyserver hkp://keyserver.ubuntu.com:80 9DC858229FC7DD38854AE2D88D81803C0EBFCD88 returned 2
       [2018-11-10T22:01:37+00:00] ERROR: apt_repository[docker-stable] (chef-apt-docker::default line 29) had an error: Mixlib::ShellOut::ShellCommandFailed: execute[install-key 9DC858229FC7DD38854AE2D88D81803C0EBFCD88] (/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/chef-13.12.3/lib/chef/provider/apt_repository.rb line 194) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '2'
       ---- Begin output of apt-key adv --recv --keyserver hkp://keyserver.ubuntu.com:80 9DC858229FC7DD38854AE2D88D81803C0EBFCD88 ----
       STDOUT: Executing: /tmp/apt-key-gpghome.dVFKHp4W4m/gpg.1.sh --recv --keyserver hkp://keyserver.ubuntu.com:80 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
       STDERR: Warning: apt-key output should not be parsed (stdout is not a terminal)
       gpg: cannot open '/dev/tty': No such device or address
```

I would expect the critical part to be

```
gpg: cannot open '/dev/tty': No such device or address
```

Even though i use `chef-apt-docker::default` here, it would not really matter, they just add a `apt_repository` [like this](https://github.com/chef-cookbooks/chef-apt-docker/blob/master/recipes/default.rb#L29)

Talking to the `debian` people, they had a lot of upgrades on gpg in those releases, so that might be the case. it seems like the possible fix is adding `--no-tty` to add-keys would help.

If of any interested, using 

```console
apt-key adv --recv --keyserver hkp://keyserver.ubuntu.com:80 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
```
on the cli on that box which failed to converge using `chef exec kitchen login default` will work without issues, so its not a false positve

EugenMayer:
Since debian 9.4+ (and probably other distros having a similar GPG path level) gpg complains about the missing tty - we use this to tell gpg that no tty will be available.

I cannot think about a way how this should break anything - we simply have no tty and never had here. All we do is tell gpg upfront before it fails doing silly checks on it.
Thinking twice, if gpg right now has tried to get tty it would have failed anyway, e.g. for ncurses questions or such.

I think this strategy should be rather bullet proof. I tested this using `vagrant ssh` without `-i` and it worked, withtout `--no-tty` it was broken as expected

### Description

### Issues Resolved

### Check List

- [ ] All tests pass


EugenMayer:
some of the tests are broken due to unrelated reasons (centos). What i would need to fix are the specs ./spec/unit/provider/apt_repository_spec.rb

EugenMayer:
`Ruby: 2.5.3 Gemfile: kitchen-tests/Gemfile
RSPEC_OPENSUSELEAP=42 KITCHEN_YAML=kitchen.travis.yml` is broken due to other reasons, fixed code styles.

Other then that, tests are looking fine

EugenMayer:
The appveyour test seems to be failing out of very other reasons - so this PR is good to go, all other tests are passing now.

EugenMayer:
Any chance someone comments on this - its a huge blocker at least for us, since that is core and we cannot "monkey" fix this as with a cookbook. Thanks!

tas50:
On the surface this looks like the correct thing, but I want to make sure this works on all the various versions of Ubuntu/Debian out there. I'll dig in today and see if there's any reason we can't merge this in.

EugenMayer:
@tas50 right now all debian/ubuntu variants are actually broken :) I tried debian > 9.3 and trusty, you cannot install any apt_repository with a key with the latest chef. So it does not get more broken then that, probably.

cheeseplus:
I think @tas50 is talking about adding the flag and the implications on older distro releases which might know what to do with that flag so causing those versions to explode.

tas50:
I confirmed that this works on Ubuntu 10.04 so we should be good to merge it in without breaking legacy OS releases.

EugenMayer:
thank you for taking the time to verify the missing aspects!

is there any time scale on when this is going to be part of an release?

pieter-vanisacker:
@tas50 @EugenMayer Is there any workaround for this issue? This is currently preventing us to bootstrap new instances.
Or is there any update as to when a new chef version will be release that fixes this?

EugenMayer:
@pieter-vanisacker i guess i have the same question. This issue also has a huge impact on docker builds, this will be quiet a wave through a lot of devops teams which are not even using chef but docker.

@pieter-vanisacker to workarround this issue i bootstrap on 9.3 initially and the upgrade on 9.6 - that works for us ( for now )

kleini:
How is it possible to get this fix into our deployments? For Debian Stretch all omnibus scripts still seem to install 14.7.17 instead of 14.7.26. Chef 13.12.3 does not contain the fix either. Applying the fix to the apt_resource.rb using a cookbook does not work either, as Chef does not seem to reload those files without restarting it.

satiani:
For those needing a temporary fix, this is how I resolved this, I transformed a block like this:

```ruby
apt_repository 'mariadb_org_repository' do
  uri "http://mariadb.mirrors.ovh.net/MariaDB/repo/10.3/debian"
  distribution "stretch"
  components ["main"]
  key "F1656F24C74CD1D8"
  keyserver "keyserver.ubuntu.com"
end
```
To:

```ruby
package 'software-properties-common' # necessary for add-apt-repository
key = "F1656F24C74CD1D8"
bash 'mariadb_org_repository' do
  code <<-EOH
  apt-key adv --no-tty --recv-keys --keyserver keyserver.ubuntu.com 0x#{key}
  add-apt-repository 'deb [arch=amd64,i386,ppc64el] http://mariadb.mirrors.ovh.net/MariaDB/repo/10.3/debian stretch main'
  EOH
  not_if "apt-key adv --list-public-keys --with-fingerprint --with-colons | grep 'fpr.*#{key}'"
end
```

kleini:
Found a way to patch the apt_repository resource in Chef and to reload it.

theorician:
@kleini Well, would you mind sharing? https://xkcd.com/979/

kleini:
I solved this by writing a cookbook to fix the issue within an installed Chef:

    ruby_block 'fix 7913 on apt_repository resource' do
      block do
        Dir.glob('/opt/chef*/embedded/lib/ruby/gems/*/gems/chef-*/lib/chef/provider/apt_repository.rb').each do |path|
          filename = ::File.expand_path(path)
          replace = Chef::Util::FileEdit.new(filename)
          replace.search_file_replace_line(/        cmd = "apt-key adv --recv"/,
            '        cmd = "apt-key adv --no-tty --recv"')
          replace.write_file
        end
        load filename
      end
      action :run
    end

theorician:
@kleini Thank you!

ziggythehamster:
I made a slight change to @kleini's monkeypatch, to only run on Chef 12. I can confirm it solved the issue on 12.22.5.

PR #7939, opened 2018-11-15
WIP: Move response_file and response_file_variables properties out of the base package resource

tas50:
This is an apt / dpkg thing and not a generic package thing. Having these properties in the package resource pollutes our resource documentation.


tas50:
This is being handled by msys now

PR #7951, opened 2018-11-19
cab_package: Chef should fail when specified package is not applicable to the image

kapilchouhan99:

### Description

- We have use `shell_out()` instead of directly hitting Mixlib::Shellout `shellout = Mixlib::ShellOut.new("dism.exe /Online /English #{command} /NoRestart", { :timeout => @new_resource.timeout })`
- Then we are using exitstatus rather than raising an error directly
- After that We are maching `Error: 0x800f081e` and raise an error like `The specified package is not applicable to this image.`

### Issues Resolved

Fixes MSYS-918: ZD 20361 - tecRacer GmbH & Co. KG - msu_package / cab_package should check return code of call to dism.exe

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/package/cab.rb
+++ b/lib/chef/provider/package/cab.rb
@@ -74,9 +74,13 @@ class Chef
         end
 
         def dism_command(command)
-          shellout = Mixlib::ShellOut.new("dism.exe /Online /English #{command} /NoRestart", timeout: new_resource.timeout)
           with_os_architecture(nil) do
-            shellout.run_command
+            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout })
+            if result.exitstatus == -2146498530
+              raise Chef::Exceptions::Package, "The specified package is not applicable to this image." if result.stdout.include?("0x800f081e")
+              result.error!
+            end
+            result
           end
         end
 
--- a/spec/functional/resource/msu_package_spec.rb
+++ b/spec/functional/resource/msu_package_spec.rb
@@ -76,6 +76,20 @@ describe Chef::Resource::MsuPackage, :win2012r2_only do
     end
   end
 
+  context "when an msu package is not applicable to the image." do
+    def package_name
+      "Package_for_KB4019990"
+    end
+
+    def package_source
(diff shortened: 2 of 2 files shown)


lib/chef/provider/package/cab.rb
```diff
@@ -74,9 +74,13 @@ def remove_package(name, version)
         end
 
         def dism_command(command)
-          shellout = Mixlib::ShellOut.new("dism.exe /Online /English #{command} /NoRestart", timeout: new_resource.timeout)
           with_os_architecture(nil) do
-            shellout.run_command
+            begin
+              shell_out!("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout })
+            rescue Exception => e
```
tas50: Can we figure out exactly what we need to rescue so we're not rescuing everything. That has some nasty side effects.

lamont-granquist: More specifically it should probably use `shell_out` without the bang and then based on the exit code either force the shell_out object to raise or to do the custom error handling.

I know we do this somewhere already.

kapilchouhan99:
@lamont-granquist @tas50 I have fixed with review comments, Please have a look on it and let me know if you have any require changes.

tas50:
@stuartpreston Any thoughts?

lib/chef/provider/package/cab.rb
```diff
@@ -74,9 +74,13 @@ def remove_package(name, version)
         end
 
         def dism_command(command)
-          shellout = Mixlib::ShellOut.new("dism.exe /Online /English #{command} /NoRestart", timeout: new_resource.timeout)
           with_os_architecture(nil) do
-            shellout.run_command
+            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout })
+            if result.exitstatus == -2146498530
+              raise Chef::Exceptions::Package, "The specified package is not applicable to this image." if result.inspect.include?("Error: 0x800f081e")
```
stuartpreston: Do we have anything else to test other than the whole result object (inc. stderr, stdout)? I'm also concerned that the word `Error:` could be localized to something else on non-English systems.

kapilchouhan99: @stuartpreston Done with the review comments, Please have a look on it.

PR #7953, opened 2018-11-19
Chef-15: switch default of allow_downgrade to true

### LINKED ISSUE #6484: Chef-15:  flip allow_downgrade options on all package resources to true

lamont-granquist:
The fact that this:

```
package "foo" do
  version "1.2.3"
end
```

Does not allow downgrades from version "2.3.4" is surprising non-declarative magic.  The meaning of that block should be to pin that package to that version.  Instead we currently make people realize there's an edge condition and then go hunting for `allow_downgrade true` to fix this behavior.  Instead we can reverse that and now the semantics of that block is exactly what it looks like it says and people who want to do "foo > 1.2.3" will need to use `allow_downgrade false` explicitly.

closes #6484

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,29 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Package provider allow_downgrade is now true by default
+
+The behavior of the package provider without any allow_downgrade flag is now to allow downgrades.  This will mostly
+affect users of the rpm and zypper package providers.
+
+```
+package "foo" do
+  version "1.2.3"
+end
+```
+
+That code should now be read as asserting that the package `foo` must be version `1.2.3` after that resource is run.
+
+```
+package "foo" do
+  allow_downgrade false
+  version "1.2.3"
+end
+```
+
+That code is now what is necessary to specify that `foo` must be version `1.2.3` or higher.  Note that the yum provider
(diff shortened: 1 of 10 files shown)


RELEASE_NOTES.md
```diff
@@ -6,6 +6,27 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Package provider allow_downgrade is now true by default
+
+A package provider without any `allow_downgrade` flag now will allow downgrades, which is opposite from previous versions. This behavior change will mostly affect users of the rpm and zypper package providers.
```
mjingle: @lamont-granquist Is the assertion that this behavior is different from previous versions correct? (This can always be changed to be more accurate.)

lamont-granquist: Yep this is a potentially breaking change for 15.0

lamont-granquist:
travis found a failure in the rspec func testing on rhel

RELEASE_NOTES.md
```diff
@@ -6,6 +6,27 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Package provider allow_downgrade is now true by default
+
```
mjingle: Consider: 
> We reversed the default behavior to `allow_downgrade true` for our package providers. To override this setting to refuse downgrades, use the `allow_downgrade —false` flag.

PR #7954, opened 2018-11-19
Chef-15:  require instead of load libraries

### LINKED ISSUE #6450: Using Kernel.require instead of Kernel.load to load library
Cookbooks libraries are loaded using *Kernel.load* in the [cookbook loader](https://github.com/chef/chef/blob/e75454384a9763458ea7d6d0aa7472521025dcbd/lib/chef/run_context/cookbook_compiler.rb#L224). This means that one library could be loaded multiple times, producing warnings about constant redefinition and other weird behavior.

I did some archeology and it seems the `Kernel.require` call has been replaced by a `Kernel.load` a while ago (6 years) due to some [issue with the chef-client daemon](https://github.com/chef/chef/commit/c3c701f6efb5b7f8f4a2bb2b1140b9f9defd6c8e).

Would it be possible to reconsider this change? I don't know if the original issue is still present with the recent code base.
A possible workaround would be to use `Kernel.require` everywhere but empty the *$LOADED_FEATURES* array using `$LOADED_FEATURES.clear` 

In the case using `Kernel.require` is not an option, maybe you should add a comment to explain why this is not possible. This would save a lot of time to people in my situation :)

## Chef Version

All including current one (13.4.24)

*Cc.* @aboten

lamont-granquist:
This reverts a change which was working around problems caused by
unforked interval runs.  Fixes issues with chefspec, should make
it faster and reduce warning spam.

closes #6450

### DIFF
--- a/lib/chef/run_context/cookbook_compiler.rb
+++ b/lib/chef/run_context/cookbook_compiler.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Daniel DeLeo (<dan@chef.io>)
-# Copyright:: Copyright 2012-2016, Chef Software Inc.
+# Copyright:: Copyright 2012-2018, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I certainly hope that this is enough of a edge condition in behavior that it affects absolutely no-one, so think i'm going to leave this off the release notes.

Annih:
Thank you @lamont-granquist :)

kamaradclimber:
It certainly have side-effect for people toying with multiple libraries overriding same method.
https://xkcd.com/1172/ !
Adding this in the release note is probably necessary

Issue #7934, opened 2018-11-14
windows_share resource adds everyone group to share


chadmccune:
## Description
"everyone" is granted read access to shares created with the windows_share resource

## Chef Version
14.7.17

## Platform Version
Windows Server 2016

## Replication Case
windows_share 'MyShare' do
action :create
path archive_folder
full_users [ "domain\domain admins", "domain\user2" ]
read_users [ "domain\user3" ]
end

## Client Output
n/a

## Stacktrace
n/a

derekgroh:
This is the default behavior of `New-SmbShare` powershell command as a default group needs access, similar to the CREATOR OWNER permission for NTFS.

The `everyone` group can be removed using `Revoke-SmbShareAccess -Name "MyShare" -AccountName "Everyone" -Force`.

Issue #7933, opened 2018-11-14
windows_share resource is not idempotent


chadmccune:
## Description
windows_share resource succeeds on first run, but fails on following runs

## Chef Version
14.7.17

## Platform Version
Windows Server 2016

## Replication Case
windows_share 'MyShare' do
  action :create
  path archive_folder
  full_users [ "domain\domain admins", "domain\user2" ]
  read_users [ "domain\user3" ]
end

## Client Output
STDERR: Revoke-SmbShareAccess : No mapping between account names and security IDs was done.

## Stacktrace

Generated at 2018-11-13 14:51:44 -0600
Mixlib::ShellOut::ShellCommandFailed: windows_share[ArchivedBackups] (selerix_archivers::default line 288) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
---- Begin output of powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "Revoke-SmbShareAccess -Name 'ArchivedBackups' -AccountName \"KITCHEN\Domain Admins,KITCHEN\sqlarchiver,KITCHEN\intranetmonitor,KITCHEN\intranetoperator,KITCHEN\nagios\" -Force" ----
STDOUT: 
STDERR: Revoke-SmbShareAccess : No mapping between account names and security IDs was done. 

At line:1 char:1

+ Revoke-SmbShareAccess -Name 'ArchivedBackups' -AccountName "KITCHEN\D ...

+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    + CategoryInfo          : NotSpecified: (MSFT_SmbShare (...copeName = "*"):ROOT/Microsoft/Windows/SMB/MSFT_SMBShar 

   e) [Revoke-SmbShareAccess], CimException

    + FullyQualifiedErrorId : Windows System Error 1332,Revoke-SmbShareAccess
---- End output of powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "Revoke-SmbShareAccess -Name 'ArchivedBackups' -AccountName \"KITCHEN\Domain Admins,KITCHEN\sqlarchiver,KITCHEN\intranetmonitor,KITCHEN\intranetoperator,KITCHEN\nagios\" -Force" ----
[...]

PR #7956, opened 2018-11-20
windows_share: Fix idempotency by removing the "everyone" access

tas50:
This resource uses powershell under the hood and calls new-smbshare,
which defaults to adding read only access to the everyone group. With
this change when we create the share we'll remove that permission. Once
that's done we'll go about adding our desired permissions. This only
runs once so the overhead is pretty low and fixes idempotency.


tas50:
Fixes #7934 and #7933

PR #7958, opened 2018-11-20
Update python_helpers for rhel-8

sspans:
### Description

RHEL 8 uses '/usr/libexec/platform-python' for system tooling, this version should be used for dnf/yum

Obvious fix.

### Issues Resolved

This makes the dnf_package and yum_package providers work on rhel-8

For more details see:
https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/

> Careful readers might have noticed a discrepancy here: Python is not installed by default, but yum is, and yum is written in Python. What magic makes that possible?
>
> It turns out there is an internal Python interpreter called “Platform-Python”. This is what system tools use. It only includes the parts of Python needed for the system to function, and there are no guarantees that any particular feature won’t be removed from it in the future.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html-single/configuring_basic_system_settings/#the_internal_platform_python_package

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/package/dnf/python_helper.rb
+++ b/lib/chef/provider/package/dnf/python_helper.rb
@@ -37,7 +37,7 @@ class Chef
           DNF_HELPER = ::File.expand_path(::File.join(::File.dirname(__FILE__), "dnf_helper.py")).freeze
 
           def dnf_command
-            @dnf_command ||= which("python", "python3", "python2", "python2.7") do |f|
+            @dnf_command ||= which("platform-python", "python", "python3", "python2", "python2.7", extra_path: "/usr/libexec") do |f|
               shell_out("#{f} -c 'import dnf'").exitstatus == 0
             end + " #{DNF_HELPER}"
           end
--- a/lib/chef/provider/package/yum/python_helper.rb
+++ b/lib/chef/provider/package/yum/python_helper.rb
@@ -40,7 +40,7 @@ class Chef
           YUM_HELPER = ::File.expand_path(::File.join(::File.dirname(__FILE__), "yum_helper.py")).freeze
 
           def yum_command
-            @yum_command ||= which("python", "python2", "python2.7") do |f|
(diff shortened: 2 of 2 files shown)


tas50:
@sspans Nice fix. Can you sign off the DCO for this commit so we can merge it?

sspans:
Any update on getting this merged? @lamont-granquist

davide125:
I can confirm this is working well for us, and is a straight backport to both 13 and 14. cc @NaomiReeves @jaymzh

pixdrift:
A similar fix was merged here:
https://github.com/chef/chef/commit/ddebf8540d4b3870ef649c0797262ff6777a0118

In this PR, the changes to the yum provider will like fail on RHEL 8 (ie. defining `yum_package` in a recipe on RHEL 8) because it can't `import yum` as yum capabilities are provided by a DNF shim in RHEL 8.

Has anyone tested `yum_package` for RHEL 8 with this PR?

sspans:
I recall yum_package working with platform-python, but I'll double check on monday.

pixdrift:
@sspans, I commented here with explanation:
https://github.com/chef/chef/commit/ddebf8540d4b3870ef649c0797262ff6777a0118#commitcomment-31519593

I would expect the `import yum` to fail (and it does on my RHEL8b1 environment with this PR)

sspans:
I did another check on rhel 8 beta and handling yum_package via python indeed fails.
So given that #8003 has been merged this can be closed.

However I would suggest adding a few lines to yum_package / python_helper to make it 
clear that 'yum_package' isn't supported on RHEL 8. Returning 'undefined method `+' for false:FalseClass' isn't very helpful.

pixdrift:
@sspans thanks for confirming.

I have suggested similar (deprecation warning) here.
https://github.com/chef/chef/commit/ddebf8540d4b3870ef649c0797262ff6777a0118#commitcomment-31519593

I really think the ddebf8540d4b3870ef649c0797262ff6777a0118 commit should be be backed out.

PR #7975, opened 2018-11-20
add a RELEASE_NOTE on /opt/chef wiping

lamont-granquist:
this is certainly worth documenting.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,9 +6,21 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Omnibus chef packages now wipe /opt/chef before installation
+
+The intent of this is that on upgrading packages the /opt/chef directory is wiped of any stray `chef_gem` installed gem versions and other
+modifications to /opt/chef that might be preserved and cause issues on upgrades.  Due to technical details with rpm script execution order
+the way this was implemented was that a pre-installation script wipes /opt/chef before every install (done consistently this way on
+every package manager).
+
+This will not affect anyone who is properly managing customizations to /opt/chef via chef recipes itself since they will be installed by
+the new chef-client package.
+
+The package installation emits a message that /opt/chef is being removed to inform the user that this action is taking place.
+
 ### Package provider allow_downgrade is now true by default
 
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -6,9 +6,21 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Omnibus chef packages now removes /opt/chef before installation
```
tas50: Just call them Chef packages. We've mostly scrubbed the word omnibus from our various docs pages

RELEASE_NOTES.md
```diff
@@ -6,9 +6,21 @@ Chef 15 release notes will be added here as development progresses.
 
 ## Breaking Changes
 
+### Chef packages now remove /opt/chef before installation
+
+The intent of this change is that on upgrading packages the /opt/chef directory is removed of any stray `chef_gem` installed gem versions and other
```
kagarmoe: What does "stray" mean in this context? Also, what does "other modifications" mean?

kagarmoe: Consider:
The upgrade process wipes the `/opt/chef` directory before every installation and with every package manager.
You'll see a warning that the `/opt/chef` directory will be removed during the package installation process.
We've implemented this change in order to remove legacy `chef_gem` installations and any other changes to this directory that might cause issues during upgrades.
Users who are properly managing customizations to `/opt/chef` through Chef recipes won't be affected, because their customizations will still be installed by
the new chef-client package.
As to the rest of all y'alls, you should really manage your `/opt/chef` customizations with Chef recipes.
We're even here to help you.
Find us on the [Chef Community Slack](http://community-slack.chef.io/) or the [Chef Community Forum](https://discourse.chef.io/).

lamont-granquist: Well "The upgrade process wipes the /opt/chef directory before every installation and with every package manager." is sort of covered by the "Chef packages now remove /opt/chef before installation" title.  

The block of text describing the technical details of why the change was made the way it was is necessary, since it will almost certainly cause questions in users as to why it was implemented the way it was.

I'll change the last two sentences to your text though.

lamont-granquist: And "stray" can just be dropped, but "other modifications" means whatever anyone might have done -- "unknown unknowns"

PR #7980, opened 2018-11-22
minimal_ohai: Add init_package plugin as a required plugin

tas50:
This is a pretty critical bit of data and we're using it for decisions
in the timezone resource now.

### DIFF
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -604,7 +604,7 @@ class Chef
     # @api private
     #
     def run_ohai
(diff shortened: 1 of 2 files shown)


lamont-granquist:
I kind of wonder if we should though...

Although the way forwards may be to create the chef-helpers gem and have both chef and ohai depend on it, and have ohai use that to populate this attribute...

tas50:
I'm fine with a single helper in the future, but at the moment we need this attribute for core chef resources to function.

PR #7982, opened 2018-11-23
Add multipackage support to homebrew

tas50:
Issues:
- versions are pretty much ignored, but I think that makes sense with homebrew. It's a roll forward package system so you can't install a specific version. That's just not a concept of the package system.

### DIFF
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -101,6 +101,25 @@ class Chef
           end
         end
 
+        #
+        # Return the package information given a package name or package alias
+        #
+        # @param [String] name_or_alias The name of the package or its alias
+        #
+        # @return [Hash] Package information
+        #
+        def package_info(package_name)
+          # return the package name if it's in the brew info hash
+          return brew_info[package_name] if brew_info[package_name]
+
+          # check each item in the hash to see if we were passed an alias
+          brew_info.each_value do |p|
+            return p if p["aliases"].include?(package_name)
+          end
+
+          {}
+        end
+
         # Some packages (formula) are "keg only" and aren't linked,
         # because multiple versions installed can cause conflicts. We
         # handle this by using the last installed version as the
@@ -111,14 +130,16 @@ class Chef
         #
         # @returns [String] package version
         def installed_version(i)
-          if brew_info[i]["keg_only"]
-            if brew_info[i]["installed"].empty?
+          p_data = package_info(i)
+
+          if p_data["keg_only"]
+            if p_data["installed"].empty?
               nil
             else
-              brew_info[i]["installed"].last["version"]
+              p_data["installed"].last["version"]
             end
           else
-            brew_info[i]["linked_keg"]
+            p_data["linked_keg"]
           end
         end
 
@@ -136,7 +157,9 @@ class Chef
         #
         # @returns [String] package version
         def available_version(i)
-          brew_info[i]["versions"]["stable"]
+          p_data = package_info(i)
+
+          p_data["versions"]["stable"]
         end
 
         private
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -54,8 +54,7 @@ class Chef
         end
 
         def install_package(names, versions)
-          packages = names.select { |x| x unless x.nil? }
-          brew("install", options, packages)
+          brew("install", options, names.compact)
         end
 
         def upgrade_package(name, version)
@@ -69,14 +68,12 @@ class Chef
         end
 
         def remove_package(names, versions)
-          packages = names.select { |x| x unless x.nil? }
-          brew("uninstall", options, packages)
+          brew("uninstall", options, names.compact)
         end
 
         # Homebrew doesn't really have a notion of purging, do a "force remove"
         def purge_package(names, versions)
-          packages = names.select { |x| x unless x.nil? }
-          brew("uninstall", "--force", options, packages)
+          brew("uninstall", "--force", options, names.compact)
         end
 
         def brew(*args)
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -95,7 +95,7 @@ class Chef
         # @returns [Hash] a hash of package information where the key is the package name
         def brew_info
           @brew_info ||= begin
-            command_array = ["info", "--json=v1"].concat new_resource.package_name
+            command_array = ["info", "--json=v1"].concat Array(new_resource.package_name)
             # convert the array of hashes into a hash where the key is the package name
             Hash[Chef::JSONCompat.from_json(brew(command_array)).collect { |pkg| [pkg["name"], pkg] }]
           end
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -106,7 +106,7 @@ class Chef
         # @return [Hash] Package information
         #
         def package_info(package_name)
-          # return the package name if it's in the brew info hash
+          # return the package hash if it's in the brew info hash
           return brew_info[package_name] if brew_info[package_name]
 
           # check each item in the hash to see if we were passed an alias
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -92,7 +92,7 @@ class Chef
         # @returns [Hash] a hash of package information where the key is the package name
         def brew_info
           @brew_info ||= begin
-            command_array = ["info", "--json=v1"].concat Array(new_resource.package_name)
+            command_array = ["info", "--json=v1"].concat package_name_array
             # convert the array of hashes into a hash where the key is the package name
 
             cmd_output = brew_cmd_output(command_array, allow_failure: true)
--- a/lib/chef/provider/package/homebrew.rb
+++ b/lib/chef/provider/package/homebrew.rb
@@ -164,7 +164,7 @@ class Chef
 
           logger.trace "Executing 'brew #{command.join(" ")}' as user '#{homebrew_user.name}'"
           # FIXME: this 1800 second default timeout should be deprecated
-          output = shell_out!('brew', *command, timeout: 1800, user: homebrew_uid, environment: { "HOME" => homebrew_user.dir, "RUBYOPT" => nil, "TMPDIR" => nil })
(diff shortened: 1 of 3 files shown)


tas50:
This speeds up a chef run with 3 brew packages to install by 11.5%. Not bad

lib/chef/provider/package/homebrew.rb
```diff
@@ -24,31 +24,40 @@ class Chef
...
         end
 
+        def get_current_versions
+          package_name_array.map do |package_name|
+            installed_version(package_name)
+          end
+        end
+
+        def install_package(names, versions)
+          packages = names.select { |x| x unless x.nil? }
```
phiggins: If this is getting rid of potential nils, `compact` does the same thing:

```
irb(main):001:0> a = [1, nil, 2, nil, 3]
=> [1, nil, 2, nil, 3]
irb(main):002:0> a.select { |x| x unless x.nil? }
=> [1, 2, 3]
irb(main):003:0> a.compact
=> [1, 2, 3]
```

lib/chef/provider/package/homebrew.rb
```diff
@@ -83,9 +90,34 @@ def brew(*args)
...
+
+        #
+        # Return the package information given a package name or package alias
+        #
+        # @param [String] name_or_alias The name of the package or its alias
+        #
+        # @return [Hash] Package information
+        #
+        def package_info(package_name)
+          # return the package name if it's in the brew info hash
```
phiggins: Super nitpicky but this comment is kind of misleading, this method does not return the package name.

lib/chef/provider/package/homebrew.rb
tas50: Updates to use compact. Thanks

lib/chef/provider/package/homebrew.rb
tas50: Fixed

lib/chef/provider/package/homebrew.rb
```diff
@@ -24,56 +24,59 @@ class Chef
...
-            brew("uninstall", "--force", options, name)
-          end
+        # upgrades are a bit harder in homebrew than other package formats. If you try to
+        # brew upgrade a package that isn't installed it will fail so if a user specifies
+        # the action of upgrade we need to figure out which packages need to be installed
+        # and which packages can be upgrades. We do this by checking if brew_info has an entry
+        # via the installed_version helper.
+        def upgrade_package(names, versions)
+          upgrade_pkgs = names.filter_map { |x| x if installed_version(x) }
+          install_pkgs = names.filter_map { |x| x unless installed_version(x) }
```
phiggins, line 67:
```suggestion
          upgrade_pkgs, install_pkgs = names.partition { |x| !!installed_version(x) }
```

phiggins: I didn't know `filter_map` was a thing. Neat! ✨

lib/chef/provider/package/homebrew.rb
```diff
@@ -83,9 +86,34 @@ def brew(*args)
...
         # command-line interface that returns the same thing.
         #
-        # https://github.com/Homebrew/homebrew/wiki/Querying-Brew
+        # https://docs.brew.sh/Querying-Brew
+        #
+        # @returns [Hash] a hash of package information where the key is the package name
         def brew_info
-          @brew_info ||= Chef::JSONCompat.from_json(brew("info", "--json=v1", new_resource.package_name)).first
+          @brew_info ||= begin
+            command_array = ["info", "--json=v1"].concat Array(new_resource.package_name)
```
phiggins, line 94:
```suggestion
            command_array = ["info", "--json=v1", new_resource.package_name]
```

Should be the same, unless I'm misunderstanding what this is doing...

tas50: We concat here because with multipackage it may be an array and this way we avoid needed to flatten it later:

```ruby
irb(main):001:0> command_array = ["info", "--json=v1"].concat Array("foo")
irb(main):002:0> command_array
=> ["info", "--json=v1", "foo"]
irb(main):003:0> command_array = ["info", "--json=v1"].concat Array(%w(foo))
irb(main):004:0> command_array
=> ["info", "--json=v1", "foo"]
irb(main):005:0> command_array = ["info", "--json=v1"].concat Array(%w(foo bar))
irb(main):006:0> command_array
=> ["info", "--json=v1", "foo", "bar"]
irb(main):007:0> name = "foo"
irb(main):008:0> command_array = ["info", "--json=v1", name]
irb(main):009:0> command_array
=> ["info", "--json=v1", "foo"]
irb(main):010:0> name = %w(foo bar)
irb(main):011:0> command_array = ["info", "--json=v1", name]
irb(main):012:0> command_array
=> ["info", "--json=v1", ["foo", "bar"]]
irb(main):013:0>
```

phiggins: Ahh, that makes sense. 👍

lib/chef/provider/package/homebrew.rb
tas50: Sadly it's not a thing in 2.6

lib/chef/provider/package/homebrew.rb
```diff
@@ -115,19 +149,23 @@ def current_installed_version
...
+        #
+        # @param [String] package name
+        #
+        # @returns [String] package version
+        def available_version(i)
+          p_data = package_info(i)
 
-        private
+          p_data["versions"]["stable"]
+        end
```
lamont-granquist, line 160: this needs to return `nil` if there's no available_version (e.g. just a typo of the package name)

tas50: Done

lib/chef/provider/package/homebrew.rb
```diff
@@ -83,9 +87,40 @@ def brew(*args)
...
+
+          # return the package hash if it's in the brew info hash
+          return brew_info[package_name] if brew_info[package_name]
+
+          # check each item in the hash to see if we were passed an alias
+          brew_info.each_value do |p|
+            return p if p["aliases"].include?(package_name)
+          end
+
+          {}
```
lamont-granquist, line 123: here you might actually want to return `{}` instead of nil for the package_info.

i didn't see this return value on first reading

lib/chef/provider/package/homebrew.rb
```diff
@@ -115,19 +158,33 @@ def current_installed_version
...
+        # @returns [String] package version
+        def available_version(i)
+          p_data = package_info(i)
 
-        private
+          # nil means we couldn't find anything
+          return nil if p_data.nil?
 
-        def get_response_from_command(*command)
+          p_data["versions"]["stable"]
```
lamont-granquist, line 171: and actually down here p_data could be `{}` which would trainwreck this method.

lib/chef/provider/package/homebrew.rb
```diff
@@ -83,9 +87,40 @@ def brew(*args)
...
+        #
+        # @returns [Hash] a hash of package information where the key is the package name
         def brew_info
-          @brew_info ||= Chef::JSONCompat.from_json(brew("info", "--json=v1", new_resource.package_name)).first
+          @brew_info ||= begin
+            command_array = ["info", "--json=v1"].concat Array(new_resource.package_name)
+            # convert the array of hashes into a hash where the key is the package name
+
+            cmd_output = brew_cmd_output(command_array, allow_failure: true)
+            return nil if cmd_output.empty? # empty std_out == bad package queried
```
lamont-granquist, line 99: oh this is deeply weird.

so ideally you'd only return a nil for the one package_name that didn't exist.

it is certainly more efficient to only shellout once, but that makes it all strange.

you could return an array of nils  which was of the size of the package_name_array but then it would probably cause the superclass to issue an error about every package being missing.

to do this right you probably need to:

```
if cmd_output.empty?
  package_name_array.each_with_object({}) do |package_name, hsh|
    cmd_output = brew_cmd_output("info", "--json=v1", package_name, allow_failure: true)
    if cmd_output.empty?
      hsh[package_name] = nil
    else
      hsh[json["name"] = Chef::JSONCompat.from_json(cmd_output)
    end
  end
else 
  Hash[Chef::JSONCompat.from_json(cmd_output).collect { |pkg| [pkg["name"], pkg] }]
end
```

lamont-granquist: So that means when it fails, you fall back to iterating one by one through the package_names to determine which one failed and what the versions were for the rest of them.

lib/chef/provider/package/homebrew.rb
lamont-granquist: you could actually push this higher up and return an empty hash out of brew_info and just delete the nil check and it'll work fine here, that keeps the return signature of both packge_info and brew_info consistently returning a Hash and uses empty? instead of nil? as the guard condition.

lib/chef/provider/package/homebrew.rb
lamont-granquist: `hsh[package_name] = {}` per my other comments that this should return empty array instead of nil here to make package_info simpler and to contain the damage in this method.

lib/chef/provider/package/homebrew.rb
```diff
@@ -83,9 +87,40 @@ def brew(*args)
...
         # command-line interface that returns the same thing.
         #
-        # https://github.com/Homebrew/homebrew/wiki/Querying-Brew
+        # https://docs.brew.sh/Querying-Brew
+        #
+        # @returns [Hash] a hash of package information where the key is the package name
         def brew_info
-          @brew_info ||= Chef::JSONCompat.from_json(brew("info", "--json=v1", new_resource.package_name)).first
+          @brew_info ||= begin
+            command_array = ["info", "--json=v1"].concat Array(new_resource.package_name)
```
lamont-granquist, line 95: `Array(new_resource.package_name)` should just be `package_name_array`

tas50: Done

lib/chef/provider/package/homebrew.rb
tas50: Done

lib/chef/provider/package/homebrew.rb
tas50: That's handled now

lib/chef/provider/package/homebrew.rb
tas50: I used this with a minor tweak to Chef::JSONCompat.from_json(cmd_output).first since Chef::JSONCompat.from_json(cmd_output) returned and array of hashes.

PR #7989, opened 2018-11-27
Make sure which mixin requires chef_class

tas50:
It uses path_to which is in chef_class


recipe.rb
```diff
@@ -0,0 +1,13 @@
+# package %w(go rust awscli)
```
tyler-ball: I imagine this isn't supposed to be here?

tas50: ... nope

PR #7998, opened 2018-11-28
better kithen ohai pinning

lamont-granquist:
since it turns out kitchen-tests/Gemfile.lock isn't actually checked in
the prior change was still just effectively floating on master, this
now makes us parse the ohai version out of the root Gemfile.lock and
use that to install ohai from git in our virt.


lamont-granquist:
travis failures are unrelated.

PR #8000, opened 2018-11-29
windows_certificate: Fix failures in delete action fails if certificate doesn't exist

Vasu1105:

### Description

Currently  certificate delete action fails if certificate doesn't exist as mentioned here https://github.com/chef-cookbooks/windows/issues/578. This fix handles this. I will be adding same fix on the windows cookbook side too.

We need to merge https://github.com/chef/win32-certstore/pull/45 before merging this.

### Issues Resolved

https://github.com/chef-cookbooks/windows/issues/578

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/windows_certificate.rb
+++ b/lib/chef/resource/windows_certificate.rb
@@ -58,7 +58,6 @@ class Chef
 
       action :create do
         description "Creates or updates a certificate."
-
         add_cert(OpenSSL::X509::Certificate.new(raw_source))
       end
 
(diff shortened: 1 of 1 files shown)


lib/chef/resource/windows_certificate.rb
```diff
@@ -58,8 +58,9 @@ class WindowsCertificate < Chef::Resource
 
       action :create do
         description "Creates or updates a certificate."
-
-        add_cert(OpenSSL::X509::Certificate.new(raw_source))
+        converge_by("Adding certificate #{new_resource.source} from Store #{new_resource.store_name}") do
```
tas50: Without any sort of additional logic this will cause the resource to update on every chef run. You need an if statement like you have below on delete

Vasu1105: @tas50 I have removed the converge_by block for create action and address in different ticket to address idempotency.

PR #8012, opened 2018-12-05
Pin the ohai definition to use the ohai version from Gemfile.lock

tas50:
Without this we end up with chef-client's definition using the version
of ohai in the gemfile.lock and the ohai defintion using whatever is in
master. This is bad for two reasons:

1. we're shipping an unreleased master version even when we ship chef 13
or chef 14
2. we ship the ohai gem twice which takes up space on disk


lamont-granquist:
if it builds correctly, ship it...  we may still wind up with transitive dep double-gems but if we're bumping both Gemfile.lock's constantly them maybe its not going to be a practical issue...

PR #8028, opened 2018-12-07
Add a new archive_file resource from the libarchive cookbook

tas50:
This adds the archive_file resource from the libarchive cookbook. I've updated the property names in that cookbook to allow us to add a :create action later on. This code matches the code in the cookbook nearly 100%.


lib/chef/resource/archive_file.rb
```diff
@@ -0,0 +1,135 @@
...
+               default: lazy { [] }
+
+      # backwards compatibility for the legacy cookbook names
+      alias_method :extract_options, :options
+      alias_method :extract_to, :destination
+
+      action :extract do
+        description "Extract and archive file."
+
+        require "fileutils"
```
lamont-granquist: don't think you need this?

lib/chef/resource/archive_file.rb
```diff
@@ -0,0 +1,135 @@
...
+          raise Errno::ENOENT, "No archive found at #{new_resource.path}!"
+        end
+
+        unless Dir.exist?(new_resource.destination)
+          converge_by("create directory #{new_resource.destination}") do
+            FileUtils.mkdir_p(new_resource.destination, mode: new_resource.mode.to_i)
+          end
+        end
+
+        converge_by("extract #{new_resource.path} to #{new_resource.destination}") do
```
lamont-granquist: converge_by without a guard so this always fires and is always updated, isn't idempotent.

tas50: At the moment this isn't actually idempotent and should really only be used with a notification block

lib/chef/resource/archive_file.rb
tas50: I think that's a leftover. I nuked it

lib/chef/resource/archive_file.rb
```diff
@@ -0,0 +1,135 @@
...
+          extract(new_resource.path, new_resource.destination,
+            Array(new_resource.options))
+        end
+
+        if new_resource.owner || new_resource.group
+          converge_by("set owner of #{new_resource.destination} to #{new_resource.owner}:#{new_resource.group}") do
+            FileUtils.chown_R(new_resource.owner, new_resource.group, new_resource.destination)
+          end
+        end
```
lamont-granquist: this is screaming out for a feature to add to chown_R/chmod_R in the directory resource

lib/chef/resource/archive_file.rb
```diff
@@ -0,0 +1,133 @@
...
+          flags = [options].flatten.map { |option| extract_option_map[option] }.compact.reduce(:|)
+          modified = false
+
+          Dir.chdir(dest) do
+            archive = Archive::Reader.open_filename(src)
+
+            archive.each_entry do |e|
+              pathname = ::File.expand_path(e.pathname)
+              if ::File.exist?(pathname)
+                modified = true unless ::File.mtime(pathname) == e.mtime
```
lamont-granquist: should probably check file size here as well at least.

lamont-granquist: oh wait or is size coming from the archive sometimes a pile of lies and that is why it isn't here?

tas50: I'm not sure I'd trust that data

lib/chef/resource/archive_file.rb
tas50: that would be a great feature

PR #8039, opened 2018-12-12
Fix for windows_package when version number is passed with build number

vijaymmali1990:

### Description

windows_package explodes when we pass version number in SemVer format (major.minor.patch)
Added fix for this
The reason for the issue is as we are passing array to get the version instead of string due to which it throws the Malformed version number string error. Resolved this by consider first element of Array.

```
irb(main):008:0> v1 = ["10.1.15.3917699"]
=> ["10.1.15.3917699"]
irb(main):009:0> gem_v1 = Gem::Version.new(v1)
        5: from /opt/chefdk/embedded/bin/irb:11:in `<main>'
        4: from (irb):9
        3: from /opt/chefdk/embedded/lib/ruby/2.5.0/rubygems/version.rb:199:in `new'
        2: from /opt/chefdk/embedded/lib/ruby/2.5.0/rubygems/version.rb:199:in `new'
        1: from /opt/chefdk/embedded/lib/ruby/2.5.0/rubygems/version.rb:208:in `initialize'
ArgumentError (Malformed version number string ["10.1.15.3917699"])
```

### Issues Resolved
It resolves following error while running the windows_package resource 
recipe:
```
windows_package 'VMware Tools' do
source 'C:\vmtools\setup.exe'
version '10.1.15'
installer_type :custom
options '/S /qn REBOOT=R ADDLOCAL=ALL'
end
```
Log trace:

```
Generated at 2018-11-21 13:53:52 +0000
ArgumentError: windows_package[VMware Tools] (@recipe_files::C:/vmtools/vmtools.rb line 1) had an error: ArgumentError: Malformed version number string ["10.0.9.3917699"]
C:/opscode/chef/embedded/lib/ruby/2.5.0/rubygems/version.rb:208:in `initialize'
C:/opscode/chef/embedded/lib/ruby/2.5.0/rubygems/version.rb:199:in `new'
C:/opscode/chef/embedded/lib/ruby/2.5.0/rubygems/version.rb:199:in `new'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.5.33-universal-mingw32/lib/chef/provider/package/windows.rb:199:in `version_compare'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.5.33-universal-mingw32/lib/chef/provider/package.rb:507:in `block in target_version_array'
C:/opscode/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.5.33-universal-mingw32/lib/chef/provider/package.rb:590:in `block in each_package'
```
Fixes MSYS-942

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -509,7 +509,7 @@ class Chef
                   if version_requirement_satisfied?(current_version, new_version)
                     logger.trace("#{new_resource} #{package_name} #{current_version} satisifies #{new_version} requirement")
                     target_version_array.push(nil)
-                  elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
+                  elsif current_version && version_compare(current_version, new_version) == 1 && !allow_downgrade
                     logger.warn("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{new_version}. Skipping...)")
                     target_version_array.push(nil)
                   else
--- a/lib/chef/provider/package/windows.rb
+++ b/lib/chef/provider/package/windows.rb
@@ -218,6 +218,10 @@ class Chef
             return 0
           end
 
+          # Till now we are getting v1 as an array. But we need only the latest version of the package in form of string.
+          # So we pass only the first element of the array.
+          # If we pass array in `Gem::Version.new(Array)` it will throw `Malformed version number string []` error.
+          v1 = v1.first if v1.is_a?(Array)
           gem_v1 = Gem::Version.new(v1)
           gem_v2 = Gem::Version.new(v2)
 


lamont-granquist:
so first thing is that `Gem::Version.new("10.0.9.3917699")` seems to behave fine in a simple test from the commandline:

```
 :001 > Gem::Version.new("10.0.9.3917699")
 => #<Gem::Version "10.0.9.3917699">
```

and the format seems entirely fine for a rubygems format, so i'm confused about the whole bug.

second thing is that it seems like this would be better done in the windows package superclass, directly in version_compare and not by mangling the version -- that will have side effects where now "10.0.9.3917699" == "10.0.9.1" and chef will be idempotent if only the build number changes, so that's just buggy.  if it really needs to be done in the subclass it should be done by subclassing version_compare.

lamont-granquist:
This may be a bug in rubygems itself and may be fixed by an upgrade in the rubygems version.

lamont-granquist:
and this kind of change would need to be tested -- and needs to show via tests that it doesn't mess up idempotency either direction ( so that "10.0.9.3917699" == "10.0.9.1.3917699" and is idempotent and that "10.0.9.3917699" != "10.0.9.1" and is not idempotent).

Vasu1105:
> so first thing is that `Gem::Version.new("10.0.9.3917699")` seems to behave fine in a simple test from the commandline:
> 
> ```
>  :001 > Gem::Version.new("10.0.9.3917699")
>  => #<Gem::Version "10.0.9.3917699">
> ```
> and the format seems entirely fine for a rubygems format, so i'm confused about the whole bug.
> 
> second thing is that it seems like this would be better done in the windows package superclass, directly in version_compare and not by mangling the version -- that will have side effects where now "10.0.9.3917699" == "10.0.9.1" and chef will be idempotent if only the build number changes, so that's just buggy. if it really needs to be done in the subclass it should be done by subclassing version_compare.

@lamont-granquist I have added more description on the PR to understand the issue and the fix around it.

lib/chef/provider/package.rb
```diff
@@ -509,7 +509,7 @@ def target_version_array
                   if version_requirement_satisfied?(current_version, new_version)
                     logger.trace("#{new_resource} #{package_name} #{current_version} satisifies #{new_version} requirement")
                     target_version_array.push(nil)
-                  elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
+                  elsif current_version && version_compare(current_version, new_version) == 1 && !allow_downgrade
```
Vasu1105: We have shifted !allow_downgrade check at the end since it always returns true if allow_downgrade is not provided or supported due to which the version_compare code is unreachable.

Vasu1105:
@lamont-granquist we also removed any code changes related to idempotency check and keep the PR to fix specific issue and we will address idempotency issue in another PR with proper analysis done around it.

lib/chef/provider/package.rb
lamont-granquist: That change needs to get reverted.  The existing behavior is correct.

This codeblock is all about preventing downgrades.  By default package providers now set this to true, which means they do not prevent downgrades, which is correct.  If a specific package provider wants to support this then it *must* implement the allow_downgrades property API in the specific provider -- and only if the user switches it to false manually will this code get triggered.

And then the choice to short circuit the version comparison is deliberate because the version comparison is not useful if allow_downgrade will always cause it to be skipped anyway.

lamont-granquist:
This bug is caused by this issue:

https://github.com/chef/chef/blob/b316f9a92a8c132a91710618795599dcfc687456/lib/chef/provider/package/windows.rb#L173-L187

What needs to happen here is that method needs to get deleted and the root problem needs to get fixed properly.

I believe what needs to happen here is that the current_resource.version needs to never be an array.  That is being set by calling `package_provider.installed_version` which likely also needs to not be an array.  Ultimately this will cause changes to all the subclasses of windows_package.  The reason why current_version is an array, though, is that the windows providers use that in order to uninstall all the versions when there are multiple versions of a package installed.  That was implemented incorrectly.  There needs to be a new API created something like `package_provider.installed_versions_for_uninstall` which will be the array of all the installed versions.  The uninstall that removes all the installed versions needs to use that instead in order to get an Array to feed to the uninstall.  There's an open question about how to choose which installed version to return in `package_provider.installed_version` and the only guidance I can give there is that if microsoft gives guidance in their documentation somewhere (i wouldn't have any idea where) that we should follow that, otherwise that should be the highest version number which is installed.

lib/chef/provider/package.rb
Vasu1105: Got it thanks :+1:

Vasu1105:
> This bug is caused by this issue:
> 
> [chef/lib/chef/provider/package/windows.rb](https://github.com/chef/chef/blob/b316f9a92a8c132a91710618795599dcfc687456/lib/chef/provider/package/windows.rb#L173-L187)
> 
> Lines 173 to 187 in [b316f9a](/chef/chef/commit/b316f9a92a8c132a91710618795599dcfc687456)
> 
>  # @return [Array] current_version(s) as an array 
>  # this package provider does not support package arrays 
>  # However, There may be multiple versions for a single 
>  # package so the first element may be a nested array 
>  # 
>  # FIXME: this breaks the semantics of the superclass and needs to get unwound.  Since these package 
>  # providers don't support multipackage they can't put multiple versions into this array.  The windows 
>  # package managers need this in order to uninstall multiple installed version, and they should track 
>  # that in something like an `uninstall_version_array` of their own.  The superclass does not implement 
>  # this kind of feature.  Doing this here breaks LSP and will create bugs since the superclass will not 
>  # expect it at all.  The `current_resource.version` also MUST NOT be an array if the package provider 
>  # is not multipackage.  The existing implementation of package_provider.installed_version should probably 
>  # be what `uninstall_version_array` is, and then that list should be sorted and last/first'd into the 
>  # current_resource.version.  The current_version_array method was not intended to be overwritten by 
>  # sublasses (but ruby provides no feature to block doing so -- it is already marked as private). 
> What needs to happen here is that method needs to get deleted and the root problem needs to get fixed properly.
> 
> I believe what needs to happen here is that the current_resource.version needs to never be an array. That is being set by calling `package_provider.installed_version` which likely also needs to not be an array. Ultimately this will cause changes to all the subclasses of windows_package. The reason why current_version is an array, though, is that the windows providers use that in order to uninstall all the versions when there are multiple versions of a package installed. That was implemented incorrectly. There needs to be a new API created something like `package_provider.installed_versions_for_uninstall` which will be the array of all the installed versions. The uninstall that removes all the installed versions needs to use that instead in order to get an Array to feed to the uninstall. There's an open question about how to choose which installed version to return in `package_provider.installed_version` and the only guidance I can give there is that if microsoft gives guidance in their documentation somewhere (i wouldn't have any idea where) that we should follow that, otherwise that should be the highest version number which is installed.

Yes we did had a look at that comment and thought of spending time around it. We thought that the currently just picking first element from the array will resolve the issue. So added that patch. And I agree that patches are not good. So what we can do is work on root problem in this PR as you mentioned in your comments and keep this PR in progress. Let me know your thoughts.

Vasu1105:
Closing this PR as this issue has been fixed in latest version of chef-client.

PR #8046, opened 2018-12-13
fix EBUSY errors in preinst script

lamont-granquist:
I think this works, its somewhat difficult to test, I couldn't find the magic that creates the EBUSY errors.

It certainly doesn't error out on me, we'll have to see if it fixes travis

### DIFF
--- a/omnibus/package-scripts/angrychef/preinst
+++ b/omnibus/package-scripts/angrychef/preinst
@@ -7,6 +7,11 @@
 #   this programming language.  do not touch.
 # - if you are under 40, get peer review from your elders.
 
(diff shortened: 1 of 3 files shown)


lamont-granquist:
eh, needs a couple fixes...

lamont-granquist:
since we have to support AIX and ultra-thin docker installs we can't rely on mktmp or fuser or lsof or pretty much anything, so i believe we have to rely on the idempotency of mkdir.

this is the same technique we use in install.sh

lamont-granquist:
appveyor blew up for entirely unrelated reasons.  red travis is what this is attempting to fix...

PR #8063, opened 2018-12-18
Extract ActionCollection out of ResourceReporter, overhaul DataCollector

lamont-granquist:
This overhauls the ResourceReporter and DataCollector code, and extracts an ActionCollection class which is common code (which can also be used for zap-like utilities to inspect the sequence of actions chef-client takes).

- The tight coupling between Chef::Client and the DataCollector has been removed.  The DataCollector and the ResourceReporter have been made responsible for disabling themselves.

- For the ActionCollection, users are expected to register with the ActionCollection by hooking the action_collection_registration event. If no objects have registered themselves with the ActionCollection then the ActionCollection will deregister itself with the Dispatcher, which will save piles of useless work + memory consumption if nobody cares.  Users may also call the ActionCollection directly to register (e.g. at library class loading time -- if writing an event handler is too "heavy").

- Users must register by the time that recipe compile time is starting.  The cookbook_compilation_start event is hooked by the ActionCollection which is where it enqueues the action_collection_registration action, passing itself as an argument.  The Dispatcher has been modified to support this kind of event chaining.  What this means is that the action_collection_registration hook is sent at the latest possible time, and users must have made a decision at that point as to if they are interested in the action collection tracking or not.

- The DataCollector now sends its output format to a file in JSON format instead of a largely-unusable un-marshalled puts of a Hash object with complex sub-objects that cannot be parsed (considered a bugfix since the prior format was un-parseable)

- The DataCollector now sends all resources and sub-resources in its output.  The sub-resources appear before the wrapping resource (in the same order as the resource_completed calls).

- The DataCollector count of updated and total resources should match the client-client run, along with being internally consistent with the resource report.

- In the DataCollector, failed resources that successfully loaded their current state now report their current state.

- Early failures to the client in e.g. client registration, node loading, run_list  expansion, cookbook resolution and cookbook synchronization are now handled by the data collector.  Failures to parse the config will still cause nodes to disappear (with no config where should the report be sent to?  what should the token be?)

- EventDispatcher has been extended to be able for registered objects to unregister

- EventDispatcher has been extended to be able to chain an event from a current event

- chef_guid code has been extracted from the Chef::Client and made the sole responsibility of the DataCollector

- Chef::Resources now have an executed_by_runner flag which is set to true by the Runner when the resource is fired.  If execution of the ResourceCollection aborts untouched resources will have this flag unset.  This is now used by the DataCollector to report "unprocessed" resources -- considerably simplifying that logic.

- The run_context gains an action_collection accessor which is what cookbook-consumers of the action_collection are expected to use.

- Various tweaks to events and signatures and the run_state clock were necessary for resilience in the face of early failures.


lamont-granquist:
If anyone does take stab at this nuclear reactor for PR review - note that I'm clearly not done-done with this PR.  There are obvious FIXMEs, I need to add some more test coverage and need to do a lot of YARD commenting.  I'm reasonably happy with the code as it stands.  What I could use is any feedback on what major areas to focus on.

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+title: Action Collection
+---
+
+# Action Collection Design
+
+The action collection tracks all actions taken by all Chef resources.  The resources can be in recipe code, as sub-resources of custom resources or
+they may be built "by hand".  Since the action collection hooks the events which are fired from the `run_action` method on Chef::Resource it does
+not matter how the resources were built (as long as they were correctly passed the Chef `run_context`).
+
+This is complementary to the resource collection itself which as incomplete statement of what might happen or has happened in the run since there are
```
tas50: which as incomplete?

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+many common ways of invoking resource actions which are not captured by how the resource collection is built.  Replying the sequence of actions in
```
tas50: Replaying or replying?

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
+---
+title: Data Collector
+---
+
+# Data Collector RFC
+
+The Data Collector design and API is covered in:
+
+https://github.com/chef/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md
```
tas50: I'd assume this repo is going to go away in the near future and dump the contents here if they're still valid

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
+---
+title: Data Collector
+---
+
+# Data Collector RFC
```
tas50: Nuke the word RFC from orbit

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
...
+
+The Data Collector also collects:
+
+- the expanded run list
+- deprecations
+- the node
+- formatted error output for exceptions
+
+Most of this is done through hooking events directly in the Data Collector itself.  The ErrorHandlers module is broken out into a module which is directly mixed into
+the Data Collector to separate that concern out into a different file (it is straightforwards with fairly little state, but is just a lot of hooked methods).
```
tas50:
```suggestion
the Data Collector to separate that concern out into a different file (it is straightforward with fairly little state, but is just a lot of hooked methods).
```

docs/dev/data_collector.md
tas50:
```suggestion
# Data Collector Design
```

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
...
+
+The Data Collector design and API is covered in:
+
+https://github.com/chef/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md
+
+This document will focus entirely on the nuts and bolts of the Data Collector
+
+## Action Collection Integration
+
+Most of the work is done by the separate action collection to track the actions of Chef resources.  If the Data Collector is not enabled, it never registers with the
```
tas50:
```suggestion
Most of the work is done by a separate action collection to track the actions of Chef resources.  If the Data Collector is not enabled, it never registers with the
```

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
...
+action collection and no work will be done by the action collection.
```
tas50:
```suggestion
action collection and no work will be done by the action collection for data collection.
```

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+This works for chef-clients which are configured to hit a chef server, but use a custom non-Chef-Automate endpoint for reporting, or for chef-solo/zero users.
+
+XXX: There is also the `Chef::Config[:data_collector][:output_locations] = { uri: [ "https://chef.acme.local/myendpoint.html" ] }` method -- which is going to behave
+differently, particularly on non-chef-solo use cases.  In that case the data collector `server_url` will still be automatically derived from the `chef_server_url` and
+the data collector will attempt to contact that endpoint, but with the token being supplied it will use that and will not use Chef Server authentication, and the
+server should 403 back, and if `raise_on_failure` is left to the default of false then it will simply drop that failure and continue without raising, which will
+appear to work, and output will be send to the configured `output_locations`.  Note that the presence of a token flips all external URIs to using the token so that
+it is not possible to use this feature to talk to both a Chef Automate endpoint and a custom URI reporting endpoint (which would seem to be the most useful of an
```
tas50:
```suggestion
it is **not** possible to use this feature to talk to both a Chef Automate endpoint and a custom URI reporting endpoint (which would seem to be the most useful of an
```

docs/dev/data_collector.md
tas50: Also that doc still talks about audit mode and really needs to be brought here so that it's current to the technology we've built / killed in the last 3-4 years

docs/dev/data_collector.md
```diff
@@ -0,0 +1,117 @@
...
+### Configured to Log to a Non-Chef Server Endpoint
+
+Setup a server url, requiring a token:
+
+```
+Chef::Config[:data_collector][:server_url] = "https://chef.acme.local/myendpoint.html"
+Chef::Config[:data_collector][:token] = "mytoken"
+```
+
+This works for chef-clients which are configured to hit a chef server, but use a custom non-Chef-Automate endpoint for reporting, or for chef-solo/zero users.
```
tas50: What would this endpoint even be?

docs/dev/action_collection.md
tas50:
```suggestion
This is complementary to the resource collection itself which has an incomplete picture of what might happen or has happened in the run since there are
```

docs/dev/action_collection.md
tas50:
```suggestion
many common ways of invoking resource actions which are not captured by how the resource collection is built.  Replaying the sequence of actions in
```

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+the action collection would be closer to replaying the chef-client converge than trying to re-converge the resource collection (although both of
+those models are still flawed in the presence of any imperative code that controls the shape of those objects).
+
+This extracts common duplicated code from the Data Collection and old Resource Reporter, and is designed to be used by other consumers which
```
tas50:
```suggestion
This design extracts common duplicated code from the Data Collection and old Resource Reporter, and is designed to be used by other consumers which
```

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Usage
+
+## Action Collection Event Hook Registration
+
+Consumers may register an event handler which hooks the `action_collection_registration` hook.  This event is fired directly before recipes are
+compiled and converged (after library loading, attributes, etc).  This is just before the earliest point in time that a resource should fire an
+action so represents the latest point that a consumer should make a decision about if it needs the action collection to be enabled or not.
+
+Consumers should hook this method, where they will be passed the action collection instance, which can be saved to reuse, and then should
```
tas50: this sentence is quite confusing

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Action Collection
+---
+
```
tas50: a small bulleted list of what the action collection hopes to achieve would be super useful. You get it once you read the whole thing and think it through + have the historical context of problems that zap had, but it would be put into just a few bullets that would make this whole thing much more useful

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+# Action Collection Design
+
+The action collection tracks all actions taken by all Chef resources.  The resources can be in recipe code, as sub-resources of custom resources or
```
tas50:
```suggestion
Action Collection tracks all actions taken by all Chef resources.  The resources can be in recipe code, as sub-resources of custom resources or
```

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+they may be built "by hand".  Since the action collection hooks the events which are fired from the `run_action` method on Chef::Resource it does
```
tas50:
```suggestion
they may be built "by hand".  Since Action Collection hooks the events which are fired from the `run_action` method on Chef::Resource it does
```

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+not matter how the resources were built (as long as they were correctly passed the Chef `run_context`).
+
+This is complementary to the resource collection itself which as incomplete statement of what might happen or has happened in the run since there are
+many common ways of invoking resource actions which are not captured by how the resource collection is built.  Replying the sequence of actions in
+the action collection would be closer to replaying the chef-client converge than trying to re-converge the resource collection (although both of
```
tas50:
```suggestion
Action Collection would be closer to replaying the chef-client converge than trying to re-converge the resource collection (although both of
```

docs/dev/action_collection.md
```diff
@@ -0,0 +1,101 @@
...
+
+The `ActionCollection` object itself implements enumerable and returns `ActionRecord` objects (see the `ActionCollection` code for the fields exposed on
+`ActionRecords`).
+
+This would return all file resources in any state in the recipe context:
+
+```
+Chef.run_context.action_collection.filtered_collection(max_nesting: 0).select { |rec| rec.new_resource.is_a?(Chef::Resource::File) }
+```
+
```
tas50:
```suggestion
NOTE:
```

docs/dev/data_collector.md
lamont-granquist: It is just an endpoint that takes start and end run messages in JSON and ...  does whatever with it.  Roll your own A2.  Explicit design goal of the DC from the start.

Issue #4815, opened 2016-04-08
Uploading "raw" data bag items fails, and we now write raw items.


tpetchel:
## Description

Uploading an encrypted data bag to Chef server fails with this error:

```
ERROR: Chef::Exceptions::InvalidDataBagItemID: Data Bag items must have an id matching /^[\.\-[:alnum:]_]+$/, you gave: nil
```

This issue blocks our Learn Chef tutorials. Users reported the issue on the RHEL and Ubuntu flavors of the same tutorial:

https://learn.chef.io/manage-a-web-app/rhel/run-your-web-application-on-a-node/#comment-2597692112

https://learn.chef.io/manage-a-web-app/ubuntu/run-your-web-application-on-a-node/#comment-2609047548

The user in the second thread worked around the issue by manually adding the `id` field to the encrypted data bag item's .json file.

I can repo this using Chef DK 0.12.0. I'm almost certain that it does not repro from 0.10.0.
## Chef Version

```
$ chef --version
Chef Development Kit Version: 0.12.0
chef-client version: 12.6.0
berks version: 4.3.0
kitchen version: ERROR
```
## Platform Version

Applies to all platforms. Can repro from Windows 8.1 and Mac OS workstations.
## Replication Case

I believe this bug would apply to uploading any encrypted data bag item to the Chef server, but here's a quick way to repo. `knife` needs to be configured to communicate with a Chef server.

Perform [step 2](https://learn.chef.io/manage-a-web-app/rhel/bring-up-the-web-app-using-test-kitchen/#step2) from the tutorial's quick start page to set up the encrypted data bag locally.

Perform [step 3](https://learn.chef.io/manage-a-web-app/rhel/run-your-web-application-on-a-node/#step3) on this page to upload the data bag to the Chef server.
## Client Output

```
$ knife data bag from file database_passwords mysql_customers.json
ERROR: Chef::Exceptions::InvalidDataBagItemID: Data Bag items must have an id matching /^[\.\-[:alnum:]_]+$/, you gave: nil
```
## Stacktrace

```
$ knife data bag from file database_passwords mysql_customers.json -VV
INFO: Using configuration from /Users/thomaspetchel/learn-chef/.chef/knife.rb
[...]

PR #8077, opened 2018-12-21
Allow encrypting a previously unencrypted data bag

### LINKED ISSUE #8282: knife data bag edit <bag> <item> --encrypt will corrupt unencrypted data bags
when created with `knife data bag create creds whatever --encrypt` a `knife data bag edit creds whatever --encrypt` pops up an editor window to this and works:

```
{
  "id": "whatever",
  "foo": "bar"
}
```

when created with `knife data bag create creds whatever` (no --encrypt) a `knife data bag edit creds whatever --encrypt` pops up an editor window to this and fails:

```
{
  "name": "data_bag_item_creds_whatever",
  "json_class": "Chef::DataBagItem",
  "chef_type": "data_bag_item",
  "data_bag": "creds",
  "raw_data": {
    "id": "whatever",
    "foo": "bar"
  }
}
```

saving it then results in corruption:

```
% knife data bag show creds whatever
ERROR: Chef::Exceptions::ValidationFailed: Property data_bag's value {"encrypted_data"=>"tP5mvplwGL2TfsJs/G13c6VVn4ldg2hn\n", "iv"=>"7TmYwNI59Xbmuik7\n", "auth_tag"=>"pLJ9cerNIQqHSPnJzOgSSw==\n", "version"=>3, "cipher"=>"aes-256-gcm"} does not match regular expression /^[\-[:alnum:]_]+$/
```

it looks like `knife data bag edit --encrypt` is assuming that the data bag is already encrypted and then corrupting a non-encrypted data bag.  users expect the data bag to be encrypted (although do they expect it to be magically decrypted if you forget the --encrypt?  probably they expect an error in that case).

note that this has nothing to do with the format on disk and this is knife communicating with the chef-server directly and not interacting with the chef-repo/chef-fs/chef-zero.

```
% knife --version
Chef: 14.11.21
```

vijaymmali1990:
### Description

After creating an unencrypted data bag, and trying to edit/encrypt, It was raising exception as shown below.
There was a minor miss while editing. If data is not encrypted, we should be returning data only instead of returning the entire item.

```
// Creating an unencrypted Data Bag
> knife data bag create test box 
Data bag test already exists
Created data_bag_item[box]

// Edit and add encryption
> knife data bag edit test box  --secret 'ab'
Encrypting data bag using provided secret.
Saved data_bag_item[box]

// Raises Exception:
> knife data bag show test box 
ERROR: Chef::Exceptions::ValidationFailed: Property data_bag's value {"encrypted_data"=>"zuDd2Da3eyZiuPehburpE3DD+oJt2TA=\n", "iv"=>"jPNhkdghdNd+IsEP\n", "auth_tag"=>"V772V241NSkUwy+102R6Dg==\n", "version"=>3, "cipher"=>"aes-256-gcm"} does not match regular expression /^[\-[:alnum:]_]+$/
```

### Issues Resolved
Fixes #8282 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/knife/data_bag_edit.rb
+++ b/lib/chef/knife/data_bag_edit.rb
@@ -43,7 +43,7 @@ class Chef
             exit(1)
           end
         else
-          return item, false
+          return item.raw_data, false
         end
       end
 
--- a/spec/integration/knife/data_bag_create_spec.rb
+++ b/spec/integration/knife/data_bag_create_spec.rb
@@ -27,29 +27,98 @@ describe "knife data bag create", :workstation do
   let(:err) { "Created data_bag[foo]\n" }
   let(:out) { "Created data_bag_item[bar]\n" }
   let(:exists) { "Data bag foo already exists\n" }
+  let(:secret) { "abc" }
 
   when_the_chef_server "is empty" do
-    it "creates a new data bag" do
-      knife("data bag create foo").should_succeed stderr: err
+    context "with encryption key" do
+      it "creates a new data bag and item" do
+        pretty_json = Chef::JSONCompat.to_json_pretty({ id: "bar", test: "pass" })
+        allow(Chef::JSONCompat).to receive(:to_json_pretty).and_return(pretty_json)
+        knife("data bag create foo bar --secret #{secret}").should_succeed stdout: out, stderr: err
+        expect(knife("data bag show foo bar --secret #{secret}").stderr).to eq("Encrypted data bag detected, decrypting with provided secret.\n")
+        expect(knife("data bag show foo bar --secret #{secret}").stdout).to eq("id:   bar\ntest: pass\n")
+      end
+
+      it "creates a new data bag and an empty item" do
+        knife("data bag create foo bar --secret #{secret}").should_succeed stdout: out, stderr: err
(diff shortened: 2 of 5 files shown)


Vasu1105:
@tas50 and @btm This LGTM. Please review it from your side.

btm:
We had a little further discussion at code review and we could add a missing set of tests to `spec/integration/knife/data_bag_edit_spec.rb` (the file doesn't exist yet) that fake an editor by modifying JSON directly. We should do that.

lamont-granquist:
yeah if we had functional tests this would probably break those for the other use case.

and the correct patch should look like a patch to the object model itself, not to knife.  so Chef::DataBagItem would need to probably have `def self.from_hash` modified to be able to accept either format.

but then that also needs to get done for data bags, roles, and environments at a minimum.

and there needs to be some attention made to DRY'ing up this code.

lamont-granquist:
ah, i think i see now.

i think this fixes #8282 which i just cut, and i deleted all the comments in #4815 which had nothing to do with the original issue.  my prior comment only makes sense in the context of where i thought this was addressing what is now #4815.

so, yes, this might fix #8282 

could really use some func or integration tests though.

btm:
This has functional tests now. Should be good to go.

PR #8087, opened 2019-01-01
chocolatey_package: Require exact match on package name and support artifactory

vstromk:
Signed-off-by: Kelvin Huang <kelvinxjh@gmail.com>

### Description

The command `choco list -r <package name>` may returns bunch of packages which are not relating to the package at all.
And in my case where `-source=https://myartifactory/`, it doesn't even return valid package result.

By adding `-a -e` will return exactly available packages, and it works perfectly with my Jfrog Artifactory.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -234,7 +234,7 @@ EOS
           package_name_array.each do |pkg|
             available_versions =
               begin
-                cmd = [ "list -r #{pkg}" ]
+                cmd = [ "list -r -a -e #{pkg}" ]
                 cmd.push( "-source #{new_resource.source}" ) if new_resource.source
                 raw = parse_list_output(*cmd)
                 raw.keys.each_with_object({}) do |name, available|
--- a/spec/unit/provider/package/chocolatey_spec.rb
+++ b/spec/unit/provider/package/chocolatey_spec.rb
@@ -60,7 +60,7 @@ describe Chef::Provider::Package::Chocolatey do
     EOF
     remote_list_obj = double(stdout: remote_list_stdout)
     package_names.each do |pkg|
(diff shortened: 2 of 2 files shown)


stuartpreston:
@kelvinxjh Overall I'm :+1: on having more specificity here, I think that would match the end user expectation.  Do you have a good example using a public Nuget repository, where the incorrect result is returned using the existing code?  We may be able to help with some tests here.

kelvinxjh:
> @kelvinxjh Overall I'm +1 on having more specificity here, I think that would match the end user expectation. Do you have a good example using a public Nuget repository, where the incorrect result is returned using the existing code? We may be able to help with some tests here. Also thank you for this contribution!

@stuartpreston Thanks for your comment.
For example with default chocolatey comunnity repository (source=https://chocolatey.org/api/v2), if we want to list available packages of `git` by running `choco list -r git`, you will get a list of packages and most of them are not relating to git at all.
But running `choco list -r -a -e git` will return a list of git packages only.

This get worse when we are using JFrog Artifactory server, where we setup local and remote cache repositoris for Nuget. The command `choco list -r git` returns empty, while `choco list -r -a -e git` return a list of git packages.

Thanks for taking this PR into consideration, and please help with those failed tests (I have difficulty to figure out how to make them pass).

lamont-granquist:
this is broken because there's a legit failure here in appveyor:

```
rspec ./spec/functional/resource/chocolatey_package_spec.rb:93 # Chef::Resource::ChocolateyPackage upgrading a package upgrades to the latest version if no version given
```

nimDevOps:
Here is a result of choco.exe list -r for git.install running against ChocoGallery:
$  choco.exe list -r git.install
git.install|2.21.0
git|2.21.0
SuperCollider|3.9.3.0
jivkok.tools|1.1.0.7

Obviously there is nothing common between git.install and SuperCollider 👎 
If we add -e option, result is much better:

$  choco.exe list -r -e git.install
git.install|2.21.0

nimDevOps:
You may try to use ChefClient before 14.11.21 where there is no choco list command yet, for example 14.10.9.
See another issue with 14.11.21+:
https://github.com/chef/chef/issues/8437

PR #8100, opened 2019-01-07
Fix for locale resource not setting LC_ALL

### LINKED ISSUE #7904: locale resource does not set LC_ALL
## Description
`locale` resource does not set `LC_ALL`

## Chef Version
```bash
$ chef-client --version
Chef: 14.6.47
```

## Platform Version
```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
```

## Replication Case
when my recipe includes
```ruby
locale 'set locale' do
  lang 'en_US.utf8'
  lc_all 'en_US.utf8'
end
```
after convergence examine the locale, the `LC_ALL` is not set
```bash
$ locale
LANG=en_US.utf8
LANGUAGE=
LC_CTYPE="en_US.utf8"
LC_NUMERIC="en_US.utf8"
LC_TIME="en_US.utf8"
LC_COLLATE="en_US.utf8"
LC_MONETARY="en_US.utf8"
LC_MESSAGES="en_US.utf8"
LC_PAPER="en_US.utf8"
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT="en_US.utf8"
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=

$ cat /etc/default/locale
LANG=en_US.utf8

$ locale -a
C
C.UTF-8
en_US.utf8
POSIX
```

## Client Output
```bash
        * locale[update locale] action update[2018-11-09T11:48:46+00:00] INFO: Processing locale[update locale] action update
           * execute[localectl set-locale LANG=en_US.utf8] action run[2018-11-09T11:48:46+00:00] INFO: Processing execute[localectl set-locale LANG=en_US.utf8] action run (/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.6.47/lib/chef/resource/locale.rb line 41)
        (skipped due to not_if)
       [2018-11-09T11:48:46+00:00] DEBUG: Skipping execute[localectl set-locale LANG=en_US.utf8] due to not_if ruby block
            (up to date)
```

after [posting this on the `locale` cookbook](https://github.com/chef-cookbooks/locale/issues/36) and migrating to chef built-in resource, i raise it here.

kapilchouhan99:
### Description
- Before we were using utilities like `localectl set-locale` `locale-gen` and `update-locale`, but now we are directly writing the contents in required files which affects system locale. This would be a generic change and will support almost all the OS and that too without relying on the presence of an external script to achieve the same functionality.

### Issues Resolved
Fixes #7904 MSYS-946: locale resource does not set LC_ALL

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/resource/locale.rb
+++ b/lib/chef/resource/locale.rb
@@ -26,52 +26,96 @@ class Chef
       introduced "14.5"
 
       property :lang, String,
-               default: "en_US.utf8",
                description: "Sets the default system language."
 
       property :lc_all, String,
-               default: "en_US.utf8",
                description: "Sets the fallback system language."
 
       action :update do
         description "Update the system's locale."
 
-        if ::File.exist?("/etc/default/locale")
-          locale_file = "/etc/default/locale"
-        elsif ::File.exist?("/etc/sysconfig/i18n")
-          locale_file = "/etc/sysconfig/i18n"
-        elsif ::File.exist?("/etc/environment")
-          locale_file = "/etc/environment"
-        else
-          raise "#{node["platform"]} platform not supported by the chef locale resource."
+        locale_file = nil
+
+        %w{ /etc/default/locale
+            /etc/sysconfig/i18n
+            /etc/environment }.each do |file|
+          if ::File.exist?(file)
+            locale_file = file
+            break
+          end
         end
 
-        contents = IO.readlines(locale_file)
-        env_val = contents.map { |t| t.split("=") if t.include?("=") }.compact.to_h
+        if locale_file.nil?
+          raise "#{node['platform']} platform not supported by the chef locale resource."
+        end
 
-        unless up_to_date?(env_val)
-          execute "Generate locales" do
-            command "locale-gen #{new_resource.lang}"
-            only_if { locale_file == "/etc/default/locale" }
-          end
+        contents = ::File.read(locale_file)
 
-          file locale_file do
-            content replace(env_val)
+        unless up_to_date?(contents, new_resource.lang, new_resource.lc_all)
+          converge_by("Applying locale settings") do
+            file locale_file do
+              content add_or_replace(contents, new_resource.lang, new_resource.lc_all)
+            end
           end
         end
       end
 
-      action_class do
-        def up_to_date?(hash)
-          hash["LANG"] && new_resource.lang == hash["LANG"].gsub(/\n|"/, "") &&
-            hash["LC_ALL"] && new_resource.lc_all == hash["LC_ALL"].gsub(/\n|"/, "")
-        end
+      def up_to_date?(contents, lang, lc_all)
+        h = parse_and_get_hash(contents)
+        h["LANG"] == lang && h["LC_ALL"] == lc_all
+      end
 
-        def replace(hash)
-          hash["LANG"] = "\"#{new_resource.lang}\""
-          hash["LC_ALL"] = "\"#{new_resource.lc_all}\""
-          hash.to_a.map { |t| t.join("=") }.join("\n")
+      def add_or_replace(contents, lang, lc_all)
+        lang_flag = true if lang
+        lc_all_flag = true if lc_all
+        new_contents = []
+        contents.each_line do |line|
+          new_contents <<  if PARSER.match? line
+                             kv = parse_and_get_hash(line)
+                             if kv["LANG"] && lang
+                               lang_flag = false
+                               if kv["LANG"] != lang
+                                 set("LANG", lang)
+                               else
+                                 line
+                               end
+                             elsif kv["LC_ALL"] && lc_all
+                               lc_all_flag = false
+                               if kv["LC_ALL"] != lc_all
+                                 set("LC_ALL", lc_all)
+                               else
+                                 line
+                               end
+                             else
+                               line
+                             end
+                           else
+                             line
+                           end
         end
+        new_contents << set("LANG", lang) if lang_flag
+        new_contents << set("LC_ALL", lc_all) if lc_all_flag
+        new_contents.join
+      end
+
+      private
+
+      # This will parse the string and capture the data where
+      # Valid "KEY = VAL" exists. Comments and whitespaces in a line are ignored
+      # Key can be any word character (letter, number, underscore)
+      # Separated by "="
+      # Value can also contain dots and quotations
+      PARSER = /^(\s*\w+\s*)=([ \t]*[\w+."']+)/i
+
+      def parse_and_get_hash(contents)
+        # For quick matching purpose, keys are converted in upper case
+        contents.scan(PARSER)
+                .map { |line| [line.first.strip.upcase, line.last.strip] }
+                .to_h
+      end
+
+      def set(key, val)
+        "\n#{key}=#{val}\n"
       end
     end
   end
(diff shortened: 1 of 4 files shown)


kapilchouhan99:
@tas50 I have verified it on both the Debian 8 and 9 VM, the code is working fine, I am not sure why it's failing for kitchen test in travis-ci. Please let me know your thoughts on this.

tas50:
This isn't passing on Amazon Linux 201X (v1) or Debian 8/9. Can you take a look at those Travis failures and see what can be done.

lib/chef/resource/locale.rb
```diff
@@ -36,56 +36,41 @@ class Locale < Chef::Resource
...
-              locale = IO.read(locale_file_path)
-              variables = Hash[locale.lines.map { |line| line.strip.split("=") }]
-              variables["LANG"] = new_resource.lang
-              variables["LC_ALL"] =
-                variables.map { |pairs| pairs.join("=") }.join("\n") + "\n"
-            }
-            not_if { updated }
-          end
+        contents = IO.readlines(locale_file)
+        env_val = contents.map { |t| t.split("=") if t.include?("=") }.compact.to_h
```
marcparadise: ["k=1", "", "k2=2"].map { |t| t.split("=") if t.include?("=") }.to_h

3: from /home/marc/.rubies/ruby-2.5.3/bin/irb:11:in `<main>'
2: from (irb):3
1: from (irb):3:in `to_h'
TypeError (wrong element type NilClass at 1 (expected array))
irb(main):004:0> 
```

**Comment by btm** [comment]
In PR review discussion we're thinking we should remove the default values for both the properties. As written write now we would write LC_ALL even if the user only specified lang, and would write LANG even if the user only specified LC_ALL.

Functionally, I don't think that is a breaking change.

File: lib/chef/resource/locale.rb (hunk 1 above)
Comment (btm) [reply lib/chef/resource/locale.rb]: This also deletes comments, which we probably shouldn't do:

```
$ cat /etc/default/locale 
#  File generated by update-locale
LANG="en_US.UTF-8"

$ irb
2.5.1 :001 > locale = IO.readlines("/etc/default/locale")
 => ["#  File generated by update-locale\n", "LANG=\"en_US.UTF-8\"\n"] 
2.5.1 :002 > locale.map { |t| t.split("=") if t.include?("=") }.compact.to_h
 => {"LANG"=>"\"en_US.UTF-8\"\n"} 
2.5.1 :003 > 
```

> [Ruby] Hashes enumerate their values in the order that the corresponding keys were inserted.

We would have to read the comments into our hash in the same loop as the `map` though to maintain the order.

File: lib/chef/resource/locale.rb [hunk 2]
```diff
@@ -36,56 +36,41 @@ class Locale < Chef::Resource
...
-        rescue
-          false
+        def up_to_date?(hash)
+          hash["LANG"] && new_resource.lang == hash["LANG"].gsub(/\n|"/, "") &&
+            hash["LC_ALL"] && new_resource.lc_all == hash["LC_ALL"].gsub(/\n|"/, "")
+        end
+
+        def replace(hash)
+          hash["LANG"] = "\"#{new_resource.lang}\""
+          hash["LC_ALL"] = "\"#{new_resource.lc_all}\""
```
robbkidd: If `LC_ALL` should only be set if the user sets a value on the resource, default to nil could guard whether LC_ALL gets written.

```ruby
hash["LC_ALL"] = "\"#{new_resource.lc_all}\"" if new_resource.lc_all
```

Also, tip, to avoid needing to escape double-quotes when interpolating in strings, you can use `%Q{}` (or only `%{}`):

```ruby
%Q{"#{new_resource.lang}"}
```

lib/chef/resource/locale.rb
```diff
@@ -36,56 +36,41 @@ class Locale < Chef::Resource
       action :update do
         description "Update the system's locale."
 
-        if node["init_package"] == "systemd"
-          # on systemd settings LC_ALL is (correctly) reserved only for testing and cannot be set globally
```
lamont-granquist: So there seems to be a comment here describing that the bug report is actually not a bug at all but a deliberate design choice.

See: https://wiki.archlinux.org/index.php/locale#LC_ALL:_troubleshooting

And I tend to agree since setting LC_ALL routinely leads to users not understanding what they're doing and submitting bug reports in all our command line tools when they break UTF-8 encoding (`LC_ALL=C` being the prime offender).

Removing using the public localectl interface for systemd users also seems like a large step backwards, not forwards.

lamont-granquist:
I lean pretty heavily to wanting to close out the issue as WONTFIX on this one.

I think if someone really wants to bypass the safety check here designed by the systemd developers to not allow setting LC_ALL that they should just use a file resource themselves, we should not support that.

Nimesh-Msys:
> I lean pretty heavily to wanting to close out the issue as WONTFIX on this one.
> 
> I think if someone really wants to bypass the safety check here designed by the systemd developers to not allow setting LC_ALL that they should just use a file resource themselves, we should not support that.

Yes, but as per the documentation, it seems that primary functionality of [locale resource](https://docs.chef.io/resource_locale.html) is only to set/unset the values of LANG and LC_ALL and not any other LC_ values. Using a utility (like `localectl`) would be significant if we have plans to set **any LC_ values other than LC_ALL** and we might want to update the document accordingly.
But consider if we follow the doc, then we have to provide an alternative to set these ENV values.
Additionally, as a part of sanity we may check that if the input locale is valid/available to be updated or not.
Please let us know your views.

lamont-granquist:
okay, yes, the bigger issue here is that we need `lc_collate` and all the rest of them.    that should probably be an entirely separate issue and should get done first.

robbkidd:
@lamont-granquist Are you recommending that we not use this to fix the `locale` resource's current claim that it can set `LG_ALL` in favor of a more comprehensive `LC_*` management resource?

lib/chef/resource/locale.rb
Nimesh-Msys: Tried and updated the logic. Handled comments, and even any invalid lines, they will remain untouched. Added the specs for almost around all the cases.
Could you please take a look and let me know in case of any issues.

lib/chef/resource/locale.rb
Nimesh-Msys: I've updated the code. I think I've covered almost all the cases.
Not enclosing the values in double quotes. Not a breaking change so far.

btm:
A proposed "resolution."

1) We add support for the other `LC_` environment variables (what is the definitive list?)
2) For Chef 15, we remove (or just deprecate?) the `lc_all` property. Either way, we end up with a warning or error that points to a deprecation page on the docs site. That error and page would say two things:

- Setting LC_ALL is not recommended and can break Chef's parsing of command output in unexpected ways. Use one of the more specific lc_ properties as needed.
- If you still want to set LC_ALL, use this file resource:
```
file "/etc/default/locale" do
  content "LC_ALL=blah"
end
``

(with better docs than that). should suggest the alternative files than just `/etc/default/locale`.

btm:
As discussed we fixed this through adding support for other LC variables and deprecating LC_ALL in #8324.

PR #8104, opened 2019-01-11
Fix for apt_repository creating temporary files as root in running user home directory

### LINKED ISSUE #8007: chef/provider/apt_repository.rb is creating temporary files as root in the running user home directory
## Description
1. lib/chef/provider/apt_repository.rb is using gpg to verify keys when installing a new repo
2. gpg by default is creating temporary file in the user home directory (if you use sudo chef-client, chef will create files in your home directory as root).
3. for users with nfs mounted home directory with root squash (very common setup), gpg will fail to create the temp file and the chef run will fail.

A proposed fix is to change 
`verify "gpg %{path}"`
to
`verify "gpg --homedir /tmp/ %{path}"`

in chef/provider/apt_repository.rb

Output proving it works:
```
omry@robodev001:~/opsfiles/chef/cookbooks/ros (master)$ sudo gpg  /var/chef/cache/https___raw_githubusercontent_com_ros_rosdistro_master_ros_key
gpg: failed to create temporary file `/nfs/home/omry/.gnupg/.#lk0x783310.robodev001.39807': Permission denied
gpg: keyblock resource `/nfs/home/omry/.gnupg/secring.gpg': general error
gpg: failed to create temporary file `/nfs/home/omry/.gnupg/.#lk0x783820.robodev001.39807': Permission denied
gpg: keyblock resource `/nfs/home/omry/.gnupg/pubring.gpg': general error
pub  1024D/B01FA116 2009-12-24 ROS Builder <rosbuild@ros.org>
sub  2048g/8F3611A0 2009-12-24
2 omry@robodev001:~/opsfiles/chef/cookbooks/ros (master)$ echo $?
2
omry@robodev001:~/opsfiles/chef/cookbooks/ros (master)$ sudo gpg --homedir /tmp/  /var/chef/cache/https___raw_githubusercontent_com_ros_rosdistro_master_ros_key
gpg: WARNING: unsafe permissions on homedir `/tmp/'
pub  1024D/B01FA116 2009-12-24 ROS Builder <rosbuild@ros.org>
sub  2048g/8F3611A0 2009-12-24
omry@robodev001:~/opsfiles/chef/cookbooks/ros (master)$ echo $?
0
```
## Chef Version
$ chef -v
Chef Development Kit Version: 3.5.13
chef-client version: 14.7.17
delivery version: master (6862f27aba89109a9630f0b6c6798efec56b4efe)
berks version: 7.0.6
kitchen version: 1.23.2
inspec version: 3.0.52

## Platform Version
Ubuntu 16.04

## Replication Case
Create an apt repository with a key from a url:

      apt_repository("ros") do
        action [:add]
        declared_type :apt_repository
        cookbook_name "ros"
        uri "http://packages.ros.org/ros/ubuntu"
        distribution "xenial"
        components ["main"]
        key ["https://raw.githubusercontent.com/ros/rosdistro/master/ros.key"]
        repo_name "ros"
      end

## Client Output
http://paste.ubuntu.com/p/HDcN3gFGcZ/

## Stacktrace
http://paste.ubuntu.com/p/DPqwRNQ8gW/

vijaymmali1990:

### Description
apt_repository was creating temporary files as root in running user home directory as a result when it goes and verifies those files or certificates it was throwing the permission errors. So provided fix for this.

### Issues Resolved

Fixes #8007 

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/apt_repository.rb
+++ b/lib/chef/provider/apt_repository.rb
@@ -206,7 +206,7 @@ class Chef
           mode "0644"
           sensitive new_resource.sensitive
           action :create
-          verify "gpg %{path}"
+          verify "gpg --homedir /tmp/ %{path}"
         end
 
         declare_resource(:execute, "apt-key add #{cached_keyfile}") do


lib/chef/provider/apt_repository.rb
```diff
@@ -206,7 +206,7 @@ def install_key_from_uri(key)
           mode "0644"
           sensitive new_resource.sensitive
           action :create
-          verify "gpg %{path}"
+          verify "gpg --homedir /tmp/ %{path}"
```
tas50: We probably want to use tempfile here so we can get a unique path to the file. https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html

vijaymmali1990: yes we can use tempfile here and I even tried using it but as I thought it will not maintain idempotancy as tempfiles get deleted automatically when ruby interpreter shuts down.

tas50: Will that impact the idempotency of the resource? It should just recreate the .gpg directory on each Chef run, which seems fine and ideal since we'd avoid reading in a predictable directory that is globally writable.

lib/chef/provider/apt_repository.rb
```diff
@@ -200,13 +201,15 @@ def key_type(uri)
       def install_key_from_uri(key)
         key_name = key.gsub(/[^0-9A-Za-z\-]/, "_")
         cached_keyfile = ::File.join(Chef::Config[:file_cache_path], key_name)
+        tmp_dir = Dir.mktmpdir(".gpg")
+        at_exit { FileUtils.remove_entry(tmp_dir) }
```
lamont-granquist: you can probably just use `@tmp_dir` here and by assigning it to an instance variable it lives for the lifetime of the provider, which allows the converge phase to complete before the tmpdir gets removed?

oh except if you don't give mktmpdir a block it never deletes it because it only returns a string and not some object that you it could setup a destructor on...

i still feel like `at_exit {}` is kinda weird.

maybe use `ruby_block` to lazy `FileUtils.rm_rf` or something?  except that updates the outer resource and will fire every time...

okay, i think i see why you're doing this...

think i see why phil wanted a ruby_block that didn't update the outer resource and fire off notifications as well...

PR #8106, opened 2019-01-13
Update bundler to 2.0.1

tas50:
Pull in the latest bundler for Chef 15


lamont-granquist:
you might run into an unsolvable chicken-and-egg problem with the travis tests...

tas50:
yeah this is never going to work in travis due to the way we sorta hack up the old omnibus install with the new code. We never get a full new omnibus install. I need to run this through jenkins and see what pops out.

Issue #8099, opened 2019-01-06
openssl_dhparam resource does not set the specified mode if the file already exists


anewb:
## Description
The openssl_dhparam resource does not honor the specified mode if the file already exists. 

## Chef Version
Chef Client, version 14.8.12

## Platform Version
Ubuntu 18.04.1 LTS

## Replication Case
Create a recipe with
```
openssl_dhparam '/etc/ssl/certs/dhparam.pem' do
end
```
and let chef-client run once.

Then change recipe to
```
openssl_dhparam '/etc/ssl/certs/dhparam.pem' do
  mode '0600'
end
```
Chef-client will not change the mode of the file on a subsequent run. I would expect if the mode is specified, it should be set according to specification.
To achieve this, the file ressource should be executed in every case, not just if there is no valid dhparam.pem file. Should be a trivial fix.

## Client Output
`  * openssl_dhparam[/etc/ssl/certs/dhparam.pem] action create (up to date)`

dheerajd-msys:
@anewb, Thank you for pointing this issue. We have got it reproduced and will be working on to fix it soon.

PR #8108, opened 2019-01-14
Fix for openssl_dhparam resource properties update

dheerajd-msys:

The `openssl_dhparam` resource does not honor the specified mode if the file already exists.
Chef-client will not change the mode of the file on a subsequent run. If the mode is specified, it should be set according to specification.
To achieve this, the file resource should be executed in every case, not just if there is no valid dhparam.pem file

### Description

Properties `mode,owner & group` can be updated now with this fix.

### Issues Resolved

Fixes https://github.com/chef/chef/issues/8099

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/resource/openssl_dhparam.rb
+++ b/lib/chef/resource/openssl_dhparam.rb
@@ -57,17 +57,27 @@ class Chef
 
       action :create do
         description "Create the dhparam file."
+        unless dhparam_pem_valid?(new_resource.path)
+          converge_by("Create a dhparam file #{new_resource.path}") do
+            dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem
 
-        converge_by("Create a dhparam file #{new_resource.path}") do
-          dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem if !::File.exist?(new_resource.path)
-
+            file new_resource.path do
+              action :create
+              owner new_resource.owner unless new_resource.owner.nil?
+              group new_resource.group unless new_resource.group.nil?
+              mode new_resource.mode
+              sensitive true
+              content dhparam_content
+            end
+          end
+        end
+        if ::File.exist?(new_resource.path)
           file new_resource.path do
             action :create
             owner new_resource.owner unless new_resource.owner.nil?
             group new_resource.group unless new_resource.group.nil?
             mode new_resource.mode
             sensitive true
-            content dhparam_content
           end
         end
       end
--- a/lib/chef/resource/openssl_dhparam.rb
+++ b/lib/chef/resource/openssl_dhparam.rb
@@ -58,9 +58,8 @@ class Chef
       action :create do
         description "Create the dhparam file."
 
-        unless dhparam_pem_valid?(new_resource.path)
           converge_by("Create a dhparam file #{new_resource.path}") do
-            dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem
+            dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem if !::File.exist?(new_resource.path)
 
             file new_resource.path do
               action :create
@@ -71,7 +70,6 @@ class Chef
               content dhparam_content
             end
           end
-        end
       end
     end
   end
--- a/lib/chef/resource/openssl_dhparam.rb
+++ b/lib/chef/resource/openssl_dhparam.rb
@@ -58,18 +58,18 @@ class Chef
       action :create do
         description "Create the dhparam file."
 
-          converge_by("Create a dhparam file #{new_resource.path}") do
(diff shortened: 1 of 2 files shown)


lib/chef/resource/openssl_dhparam.rb
```diff
@@ -58,18 +58,16 @@ class OpensslDhparam < Chef::Resource
       action :create do
         description "Create the dhparam file."
 
-        unless dhparam_pem_valid?(new_resource.path)
-          converge_by("Create a dhparam file #{new_resource.path}") do
-            dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem
+        converge_by("Create a dhparam file #{new_resource.path}") do
```
tas50: This will converge on each run.

dheerajd-msys: This `converge_by` is just a message. File is not being created if already exist. But `mode` property can be updated. And also with this fix content of the existing file is also not changing on subsequent run. Ensured idempotency of the resource.

## Recipe
```
root@dh-openssl-param:~/chef# cat ../openssl_dhparam.rb
openssl_dhparam '/etc/ssl/certs/dhparam.pem' do
    mode '0640'
end
```
## First Run
```
root@dh-openssl-param:~/chef# bundle exec chef-apply ../openssl_dhparam.rb
  * openssl_dhparam[/etc/ssl/certs/dhparam.pem] action create
    - Create a dhparam file /etc/ssl/certs/dhparam.pem
    * file[/etc/ssl/certs/dhparam.pem] action create
      - create new file /etc/ssl/certs/dhparam.pem
      - update content in file /etc/ssl/certs/dhparam.pem from none to 302bc8
      - suppressed sensitive resource
      - change mode from '' to '0640'
root@dh-openssl-param:~/chef#
```
## Second Run
```
root@dh-openssl-param:~/chef# bundle exec chef-apply ../openssl_dhparam.rb
  * openssl_dhparam[/etc/ssl/certs/dhparam.pem] action create
    - Create a dhparam file /etc/ssl/certs/dhparam.pem
    * file[/etc/ssl/certs/dhparam.pem] action create (up to date)

root@dh-openssl-param:~/chef#
```
## If `mode` is changed
```
root@dh-openssl-param:~/chef# bundle exec chef-apply ../openssl_dhparam.rb
  * openssl_dhparam[/etc/ssl/certs/dhparam.pem] action create
    - Create a dhparam file /etc/ssl/certs/dhparam.pem
    * file[/etc/ssl/certs/dhparam.pem] action create
      - change mode from '0640' to '0600'

root@dh-openssl-param:~/chef#
```
@tas50 Please let me know your thoughts.

tas50: converge_by is not just a log message though. converge_by marks the resource as being updated, which fires off notifications and adds it to any resource reporting which goes to A2. It's pretty important that we only do this when a resource is actually changed. If not bad things happen.

lamont-granquist:
https://docs.chef.io/custom_resources_notes.html

That doc has notes on converge_by.

> A converge_by block that is not wrapped in an idempotency check will always cause the resource to be updated, and will always cause notifications to fire. To prevent this, a properly written resource should wrap all converge_by checks with an idempotency check.

lib/chef/resource/openssl_dhparam.rb
```diff
@@ -72,6 +71,15 @@ class OpensslDhparam < Chef::Resource
             end
           end
         end
+        if ::File.exist?(new_resource.path)
```
btm: unless dhparam_pem_valid?(new_resource.path)
 # generate key
  # file resource that creates key
else
  # file resource that fixes mode/owner/group
end

lib/chef/resource/openssl_dhparam.rb
```diff
@@ -72,6 +71,15 @@ class OpensslDhparam < Chef::Resource
...
+          file new_resource.path do
+            action :create
+            owner new_resource.owner unless new_resource.owner.nil?
+            group new_resource.group unless new_resource.group.nil?
```
btm: @lamont-granquist says in both of these resources we can simply do:

```
owner new_resource.owner
group new_resource.group
```

(and drop the conditional)

lib/chef/resource/openssl_dhparam.rb
marcparadise: Could we instead avoid repeating the resource with something like: 
```
        dhparam_contenxt = nil
        unless dhparam_pem_valid?(new_resource.path)
          converge_by("Generating dhparam") do
            dhparam_content = gen_dhparam(new_resource.key_length, new_resource.generator).to_pem
          end
       end
     
        file new_resource.path do
          action :create
          owner new_resource.owner unless new_resource.owner.nil
          group new_resource.group unless new_resource.group.nil?
          mode new_resource.mode
          sensitive true
          content dhparam_content unless dhparam_content.nil?
       end
```

btm:
Merged in https://github.com/chef/chef/pull/8245

PR #8112, opened 2019-01-15
changes for rubocop engine upgrades.

lamont-granquist:
this is the result of changes to rules we already previously had
enabled.

should pass on both versions of the engine.


lamont-granquist:
chef/chefstyle#68 companion engine upgrade that needs to go in and get released first.

lib/chef/cookbook/metadata.rb
```diff
@@ -467,24 +467,24 @@ def gem_requirements_from_array(what, array)
 
       def to_h
         {
-          NAME                   => name,
```
tas50: best rule ever

lib/chef/encrypted_data_bag_item/decryptor.rb
```diff
@@ -88,26 +88,28 @@ def for_decrypted_item
...
       def decrypted_data
-        @decrypted_data ||= begin
-          plaintext = openssl_decryptor.update(encrypted_bytes)
-          plaintext << openssl_decryptor.final
-        rescue OpenSSL::Cipher::CipherError => e
-          # if the key length is less than 255 characters, and it contains slashes, we think it may be a path.
-          raise DecryptionFailure, "Error decrypting data bag value: '#{e.message}'. Most likely the provided key is incorrect. #{(@key.length < 255 && @key.include?('/')) ? 'You may need to use --secret-file rather than --secret.' : ''}"
-        end
+        @decrypted_data ||=
+          begin
```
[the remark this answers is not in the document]
lamont-granquist: my feelings exactly

lib/chef/cookbook/metadata.rb
lamont-granquist: yeah i think i just lost an argument we had at rally because i'm too lazy to care...

i welcome our new non-aligned overlords...

PR #8118, opened 2019-01-16
windows_certificate: Ensure all actions are fully idempotent

Nimesh-Msys:
- Minor fixes in `:create` action
- DRYed up `verify_cert` action and reusing the same while :create
- Chefstyle maintained
- Added Functional test cases

### Description

- For `:create` action to be idempotent, we need to check if a certificate with the same thumbprint that we are going to import is already present or not. If present, we should restrict the import otherwise converge
- `:fetch` and `:verify` need not be converged and they are already idempotent
- `:delete`  was already fixed with recent changes. Only thing, if a certificate is not present, then we must inform the user.

With this, `windows_certificate` is now idempotent under `:add, :verify, :delete and :fetch` actions.

### Testing

Have manually tested this fix as follows:
Exported a certificate in various formats(base64 encoded CER, CER and PFX12) and executed the following script:
```
# Should converge: 2(Add & Delete)
#                  X 3(types of certificates)
#                  X 3(again in loop) = 18 resources
def perform_all(cert)
  add_cert(cert)
  verify_cert(cert)
  fetch_cert(cert)
  delete_cert(cert)
end

@cert.each do |cert_name|
  # Converge: 18/39 resources
  add_cert(cert_name)
  @cert.each do |cert|
    perform_all(cert)
  end

  # Converge: 18/39 resources
  verify_cert(cert_name)
  @cert.each do |cert|
    perform_all(cert)
  end

  # Converge: 18/39 resources
  fetch_cert(cert_name)
  @cert.each do |cert|
    perform_all(cert)
  end

  # Converge: 18/39 resources
  delete_cert(cert_name)
  @cert.each do |cert|
    perform_all(cert)
  end
end
```
For each method, we have written down the recipe using `windows_certificate` resource with their respective actions.
As per the expectations, the result is:

```
Chef Client finished, 72/156 resources updated in 01 minutes 48 seconds
```

### Issues Resolved

#[576](https://github.com/chef-cookbooks/windows/issues/576)

### Check List

- [x] All tests pass


lib/chef/resource/windows_certificate.rb
```diff
@@ -58,7 +58,21 @@ class WindowsCertificate < Chef::Resource
...
-        add_cert(OpenSSL::X509::Certificate.new(raw_source))
+
+        cert_obj = OpenSSL::X509::Certificate.new(raw_source) # A certificate object in memory
+        thumbprint = OpenSSL::Digest::SHA1.new(cert_obj.to_der).to_s # Fetch its thumbprint
+
+        # Check whether a certificate with this thumbprint
+        # is already present in certificate store
+        exists = verify_cert(thumbprint)
+
+        if (!!exists == exists) && exists
```
lamont-granquist: whatever this line of code is doing it absolutely needs to be made clearer

lib/chef/resource/windows_certificate.rb
```diff
@@ -94,6 +108,8 @@ class WindowsCertificate < Chef::Resource
           converge_by("Deleting certificate #{new_resource.source} from Store #{new_resource.store_name}") do
             delete_cert
           end
+        else
+          Chef::Log.info("Certificate not found")
```
tas50: We probably want this at debug level

lib/chef/resource/windows_certificate.rb
```diff
@@ -58,7 +58,21 @@ class WindowsCertificate < Chef::Resource
...
+
+        cert_obj = OpenSSL::X509::Certificate.new(raw_source) # A certificate object in memory
+        thumbprint = OpenSSL::Digest::SHA1.new(cert_obj.to_der).to_s # Fetch its thumbprint
+
+        # Check whether a certificate with this thumbprint
+        # is already present in certificate store
+        exists = verify_cert(thumbprint)
+
+        if (!!exists == exists) && exists
+          Chef::Log.info("Certificate is already present")
```
tas50: This should probably be default level

lib/chef/resource/windows_certificate.rb
```diff
@@ -58,7 +58,21 @@ class WindowsCertificate < Chef::Resource
 
       action :create do
         description "Creates or updates a certificate."
-        add_cert(OpenSSL::X509::Certificate.new(raw_source))
+
+        cert_obj = OpenSSL::X509::Certificate.new(raw_source) # A certificate object in memory
```
tas50: can we get a comment on the raw_source method explaining what's it's doing

lib/chef/resource/windows_certificate.rb
```diff
@@ -134,9 +150,9 @@ def fetch_cert
           store.get(new_resource.source)
         end
 
-        def verify_cert
+        def verify_cert(thumbprint = new_resource.source)
```
tas50: can you add a comment to this method while you're here

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Yes, actually I also didn't like the way `verify_cert` returns the output. When a certificate for the given thumbprint is not present, then it returns a `String`: "Certificate not found" from [this line of code](https://github.com/chef/win32-certstore/blob/master/lib/win32/certstore/store_base.rb#L174). But when a certificate is present, it returns `Boolean`: true.
Due to this reason, I applied the same [logic which is used in `:verify` action](https://github.com/chef/chef/blob/master/lib/chef/resource/windows_certificate.rb#L114-L117) with a thought that may be it was an acceptable solution.
As per the logic, if the return output is of Boolean type (!!out == out) AND its value is true, then certificate is present in the store.
I can put the comments to make it clear, but will it not be better to changes such tricky unplanned returns in library itself. Need your views. Thanks.

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Okay, I'll update this. Thanks

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Okay, I'll update this. Thanks

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Yes, I will do that. Thanks

Nimesh-Msys: Have added the comment. Thanks

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Added comments. Thanks

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+    end
+
+    context "Readding same certificate" do
+      before do
+        win_certificate.run_action(:create)
+      end
+      it "Does not imports certificate into store" do
+        expect(no_of_certificates).to eq(1)
+      end
+      it "Idempotent: Does not converges while addition" do
```
Vasu1105: typo `Does not converge`

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+        expect(no_of_certificates).to eq(1)
+        expect(win_certificate).not_to be_updated_by_last_action
+      end
+    end
+
+    context "Readding same certificate of other format" do
```
Vasu1105: Typo `Reading`

spec/functional/resource/windows_certificate_spec.rb
Nimesh-Msys: Done

PR #8123, opened 2019-01-16
homebrew_cask / homebrew_tap:  Properly use the cask_name and tap_name properties

tas50:
If the name properties for these resources we used we'd entirely ignore them.

### DIFF
--- a/lib/chef/resource/homebrew_cask.rb
+++ b/lib/chef/resource/homebrew_cask.rb
@@ -58,8 +58,8 @@ class Chef
         homebrew_tap "caskroom/cask" if new_resource.install_cask
 
         unless casked?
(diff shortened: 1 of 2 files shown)


tas50:
We don't really have the ability to test this the way that Chef is structured at the moment. I'll be adding tests to https://github.com/chef/bft

robbkidd:
@rhass `cask_name` is designated as the `name_property`, so it is already required sorta kinda. This change does not increase its requirementality ... which is a word I just made up.

tas50:
@rhass It does not. Since cask_name and tap_name are name_properties the value of new_resource.name becomes their default. It's a super handy thing in Chef. If someone doesn't use them they just get the resource name and otherwise they get the override.

PR #8135, opened 2019-01-18
[knife] Remove duplicate code blocks in the knife cookbook upload command

f9n:
### Description

I didn't encounter a bug. I was just reading the codes. Then, I wanted to remove the duplicate codes.

### Check List

- [X] All tests pass

### DIFF
--- a/lib/chef/knife/cookbook_upload.rb
+++ b/lib/chef/knife/cookbook_upload.rb
@@ -76,21 +76,13 @@ class Chef
 
       def run
         # Sanity check before we load anything from the server
-        unless config[:all]
-          if @name_args.empty?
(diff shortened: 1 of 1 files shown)


f9n:
I think that I need to change the conditional expression in this 'pull request' to the old conditional expression. Because the old conditional statement is readable more than that. Should I change it?

```ruby
# Old
unless config[:all]
  if @name_args.empty?
    ...
   end
end

# In Pr
if ! config[:all] && @name_args.empty?
  ...
end
```

clintoncwolfe:
I'm personally happy with two operators in an `if`.  Three is iffy and four is right out.

PR #8136, opened 2019-01-18
Sysctl: Allow slashes in key or block name

kapilchouhan99:
This change will allow user to pass slash separated key/name in `sysctl` resource.

### Issue

As per the functionality, sysctl resource creates a file in `sysctl.d` directory with the file name as the key/name passed in resource. (`"99-chef-<key>"`). Due to this reason, even if we want to pass slashes in key, then while creating a file it was being considered as a directory and raised an error.
But as per sysctl's syntax, in case an interface name contains dot in it, then other properties must be separated by slash.
```
eg: Interface: ens256.401
valid sysctl key: net/ipv4/conf/ens256.401/rp_filter
```
We should be able to take care of this necessity.

### Resolution

With these changes: 
- Name of the File which is created by chef will always be dot separated
  - This would avoid any of it's consideration to a directory. Hence any exception wouldn't arise.
- File content will be as same as the key/name passed by the user
  - Hence, user can also pass slashes as a valid separator for sysctl.

### Note

Similar to systctl, now we also support both the separators as a valid sysctl key when an interface doesn't contains dot in it. We might send another PR for document updation
```
   net/ipv4/conf/default/rp_filter
   net.ipv4.conf.default.rp_filter
```

### Issues Resolved

MSYS-931: Sysctl doesn't work when key has a . in the name

### Check List

- [x] All tests pass

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/tests.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/tests.rb
@@ -19,3 +19,11 @@ end
 file "/tmp/chef-test-\xFDmlaut" do
   content "testing illegal UTF-8 char in the filename"
 end
+
+node["network"]["interfaces"].each do |interface_data|
+  interface = interface_data[0]
+  sysctl_param "net/ipv4/conf/#{interface}/rp_filter" do
+    value 0
+    ignore_failure true
+  end
+end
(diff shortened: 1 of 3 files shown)


tas50:
Thanks @kapilchouhan99

PR #8140, opened 2019-01-21
Fix cask resource running each chef-client run

jeroenj:
### Description

This fixes an issue where casks would be installed during each chef-client run because the `casked?` check failed.

### Issues Resolved

Interpolation was not taken into account correctly in the `casked?` check.

### Check List

- [ ] All tests pass

---

@tas50 Should the changes from the [homebrew](https://github.com/chef-cookbooks/homebrew) cookbook all be backported into Chef? (in case this and #8139 don't cover everything yet)


tas50:
Hurts to see this. Thanks for fixing it

PR #8149, opened 2019-01-22
Add a deprecation warning to knife cookbook site

tas50:
We'll want to backport this to Chef 14 as well.

### DIFF
--- a/lib/chef/knife/cookbook_site_download.rb
+++ b/lib/chef/knife/cookbook_site_download.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright 2009-2016, Chef Software, Inc.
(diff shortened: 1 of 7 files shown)


lib/chef/knife/cookbook_site_download.rb
```diff
@@ -29,6 +29,12 @@ class CookbookSiteDownload < Knife::SupermarketDownload
 
       banner "knife cookbook site download COOKBOOK [VERSION] (options)"
       category "deprecated"
```
lamont-granquist: does category "deprecated" not do anything useful like this already?

tas50: Nope

tas50: if I thought we were going to do this more I'd add a deprecated method to the plugin that would handle it, but I think we're good until we look at migrating knife commands over to chef.

PR #8151, opened 2019-01-22
Remove 'attributes'  attribute from cookbook metadata

tas50:
We have a foodcritic rule to tell people not to do this
It's not being used anywhere and is just a pile of code we run for no reason
We should start warning that it's an unknown attribute when cookbooks contain these as it's wasted effort on the part of cookbook authors

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -44,7 +44,6 @@ class Chef
       PLATFORMS              = "platforms".freeze
       DEPENDENCIES           = "dependencies".freeze
       PROVIDING              = "providing".freeze
-      ATTRIBUTES             = "attributes".freeze
       RECIPES                = "recipes".freeze
       VERSION                = "version".freeze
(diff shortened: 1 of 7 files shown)


lib/chef/cookbook/metadata.rb
```diff
@@ -56,9 +55,8 @@ class Metadata
 
       COMPARISON_FIELDS = [ :name, :description, :long_description, :maintainer,
                             :maintainer_email, :license, :platforms, :dependencies,
-                            :providing, :attributes, :recipes, :version,
-                            :source_url, :issues_url, :privacy, :chef_versions, :ohai_versions,
-                            :gems ].freeze
+                            :providing, :recipes, :version, :source_url, :issues_url,
+                            :privacy, :chef_versions, :ohai_versions, :gems ].freeze
```
lamont-granquist: i spent way too long diffing these lines in my head.

tas50: that's why 1 item per line is a totally inefficient, but great way to do arrays

Issue #7434, opened 2018-07-07
Add path property alias to scm resource


mal:
## Description
The `scm` resource takes a file path under it's `destination` property, in the vast majority of other core resources similar properties are named `path`. Would it be possible to introduce `path` as an alias for `destination` in this resource, to follow the approach of least surprise.

## Chef Version
Chef 14

tas50:
@mal I'd be open to a PR to add change the resource to path with an alias so that destination still works. You'd want to add a note in description that the legacy destination properly will still function. We'd probably merge that in for Chef 15

mal:
Closing as the change was rejected even despite being non-breaking.

PR #8152, opened 2019-01-22
Make SCM resource consistent with other resources

mal:
Introduce the 'path' property and deprecate the 'destination' property.
This provides a more consistent experience for users across file based
resources.

Fixes https://github.com/chef/chef/issues/7434

Updated tests and release notes as per issue template.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -151,6 +151,10 @@ Initial support has been added for Red Hat Enterprise Linux 8. Thank you [@pixdr
 
 gem_package now supports installing gems into Ruby 2.6 or later installations.
 
+### scm / git / subversion
+
+scm now supports the path property in favour of destination. However destination will continue to work at this time.
+
 ### windows_ad_join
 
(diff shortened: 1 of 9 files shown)


RELEASE_NOTES.md
```diff
@@ -151,6 +151,10 @@ Initial support has been added for Red Hat Enterprise Linux 8. Thank you [@pixdr
 
 gem_package now supports installing gems into Ruby 2.6 or later installations.
 
+### scm / git / subversion
+
+scm now supports the path property in favour of destination. However destination will continue to work at this time.
```
```suggestion
`scm` resource now supports the `path` property rather than the `destination` property. While currently supported, the `destination` property will be deprecated eventually.
```

lamont-granquist:
this is the kind of breaking change that we probably can't accept at this point, it would need to get done as a deprecation and its probably not worth the code churn to make everyone change.

mal:
Would you consider the inverse of this `alias` in order to support the path of least-surprise, even if wishing to retain `destination` for backwards-compatibility?

https://github.com/chef/chef/pull/8152/files#diff-0a644a52e840f13a7f5bae7934d7c080R72

PR #8158, opened 2019-01-23
Remove support for unsupported opensuse < 42 from group provider

tas50:
We had a few places where we checked for legacy opensuse platforms. Chef 15 doesn't support this platform and you won't be able to find binaries via omnitruck. We might as well yank out a bit of provider logic.

Also for some reason usermod has logic in a method for suse, but the provider didn't support that.

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/metadata.rb
+++ b/kitchen-tests/cookbooks/end_to_end/metadata.rb
@@ -20,7 +20,7 @@ depends          "git"
 supports         "ubuntu"
 supports         "debian"
 supports         "centos"
-supports         "opensuse"
+supports         "opensuseleap"
 supports         "fedora"
 supports         "amazon"
 
--- a/lib/chef/provider/group/groupadd.rb
+++ b/lib/chef/provider/group/groupadd.rb
@@ -1,6 +1,6 @@
 #
 # Author:: AJ Christensen (<aj@chef.io>)
-# Copyright:: Copyright 2008-2016, Chef Software Inc.
+# Copyright:: Copyright 2008-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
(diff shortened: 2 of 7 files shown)


lamont-granquist:
i'm very confused because it seems the intention is to use the suse group provider for legacy suse, while you've ripped support out of the usermod one, where it seems more like the provides lines are broken and need to match the conditional inside the provider, rather than the other way around.

tas50:
We don't support opensuse as a platform anymore. It was replaced with opensuseleap 42 many years ago and some versions of that are even EOL at this point. opensuse is the old 13 release, which just needs to go away.

lamont-granquist:
so what i was missing was the context that suse is now picked up by the gpasswd provider.  that didn't appear in this PR because its just a default provider.

looking into the unit specs though things still look weird:

```
          "suse" => {
            group: [ Chef::Resource::Group, Chef::Provider::Group::Gpasswd ],
            "suse" => {
              "12.0" => {   # <------------------- this looks wrong.
              },
              %w{11.1 11.2 11.3} => {
                group: [ Chef::Resource::Group, Chef::Provider::Group::Suse ],
              },
            },
            "opensuseleap" => {
    #          service: [ Chef::Resource::RedhatService, Chef::Provider::Service::Redhat ],
              package: [ Chef::Resource::ZypperPackage, Chef::Provider::Package::Zypper ],
              group: [ Chef::Resource::Group, Chef::Provider::Group::Usermod ],  # <-------------   this also looks wrong?
            },
          },
```

why is opensuseleap using Usermod while suse is using gpasswd?

tas50:
This should get merged after #8165 so we can cleanly backport that PR.

tas50:
@lamont-granquist this should be ready with that spec change

PR #8162, opened 2019-01-24
Use the latest omnibus-software and nokogiri

tas50:
The new omnibus-software includes a patch I added to fix compilation
failures on non C99 compatible compilers with 2.6.0.

The omnibus release that gets pulled in is also necessary for Ruby 2.6 on macOS

The new nokogiri supports windows on Ruby 2.6, which the previous
version did not.


tas50:
No need to wait on the tests since this will only fail in jenkins and it's already been run there in the ruby26 branch.

Issue #8161, opened 2019-01-23
Group resource: "append true" doesn't work right for adding users to multiple groups on Solaris.


jjustice6:
## Description
When using :modify on a group in `solaris2` (5.11) environments, `append true` doesn't work correctly when adding a user to *additional* groups. See `-G` flag comments on https://docs.oracle.com/cd/E26502_01/html/E29031/usermod-1m.html , note that the code for the usermod resource doesn't put `+` or `-` before the group name.

We're working on a fix for this.

## Chef Version
Chef 14.7.17, though the bug exists through many versions.

## Platform Version
Solaris 5.11. 

## Replication Case

```
group1_name = 'tstgrp1'
group1_id = 6789
group2_name = 'tstgrp2'
group2_id = 9876
group group1_name do
  action :create
  gid group1_id
end
group group2_name do
  action :create
  gid group2_id
end

group 'primary' do
  action :create
end

firstbatch = %w(user1 user2)
secondbatch = %w(user3 user4)
%w(user1 user2 user3 user4).each do |username|
  user username do
    group 'primary'
    manage_home true
  end
end

group 'add firstbatch to group1' do
  group_name group1_name
  action :modify
  members firstbatch
  append true
end

group 'add secondbatch to group2' do
  group_name group2_name
  action :modify
  members secondbatch
  append true
end

# Now, add them to the other group. The end result SHOULD be that all 4 users are in both groups.
group 'add firstbatch to group2' do
  group_name group2_name
  action :modify
  members firstbatch
  append true
end

group 'add secondbatch to group1' do
  group_name group1_name
  action :modify
  members secondbatch
  append true
end

```

## Client Output
Appears normal, until you actually look at the groups.

EXPECTED: All 4 users are in both groups.
ACTUAL:
```
# getent group tstgrp1
tstgrp1::6789:user3,user4
# getent group tstgrp2
tstgrp2::9876:user1,user2
```

Make sure that you think of it as "the user should be in X groups", not "the group should have X users" - if you only have one group and attempt to add users in successive `:modify` actions, that won't cause the problem. The problem here is not that "add secondbatch to group1" wiped out the firstbatch users from group 1 - it's that "add secondbatch to group1" _took them out of group 2_.

## Stacktrace
N/A

tas50:
Fix merged to master and will be backported for the next Chef 14 release. Thanks @jjustice6

PR #8165, opened 2019-01-24
Create a separate Group provider for Solaris systems.

jjustice6:
### Description
This creates a new Group provider for Solaris systems (`solaris2` environments), as the underlying `usermod` command behaves differently than how it does on other systems due to the +/- flag preceding the group list. See [the manpage](https://docs.oracle.com/cd/E26502_01/html/E29031/usermod-1m.html) for details.

### Issues Resolved

https://github.com/chef/chef/issues/8161

### Check List

- [X] All tests pass


jjustice6:
I'm struggling to figure out exactly what needs to be where for the "other" solaris platforms.
Omnios points to http://wiki.omniosce.org/ManSections.html which suggests http://illumos.org/man/1M/usermod is the non +/- behavior.
SmartOS also has the other behavior. https://smartos.org/man/1M/usermod
OpenIndiana points to http://docs.openindiana.org/ which points to the Illumos pages.

So far I only see that actual Oracle Solaris 5.11 with the new behavior using the +/- on usermod.

So what's the best way to define the provides for this?

lamont-granquist:
My $0.02 is to retain the existing behavior unless you can find something fairly definitive stating that the other solaris platforms use the new behavior.

Since none of them appear to use the new behavior, then the only provides like that you'd setup here would be for "solaris2" itself, and the others would all default to the old behavior.

Then we ensure we don't break anyone that isn't already broken.   For the people that are already broken it is up to them to do the research and submit the PR to change the behavior to fix it.

jjustice6:
That makes sense to me.

lib/chef/provider/group/solaris.rb
```diff
@@ -0,0 +1,72 @@
...
+#
+
+require "chef/provider/group/groupadd"
+
+class Chef
+  class Provider
+    class Group
+      class Solaris < Chef::Provider::Group::Groupadd
+
+        provides :group, platform: "solaris2"
```
tas50:
```suggestion
        # this provides line is setup to only catch the solaris2 platform and not other platforms in the Solaris platform_family      
        provides :group, platform: "solaris2"
```

spec/unit/provider_resolver_spec.rb
```diff
@@ -851,8 +851,7 @@ def stub_service_configs(*configs)
               user: [ Chef::Resource::User::SolarisUser, Chef::Provider::User::Solaris ],
               "5.11" => {
```
tas50: Is there an omnios or smartos spec in the provider_resolver spec so we can make sure the existing group provider still loads on these platforms?

lib/chef/provider/group/solaris.rb
```diff
@@ -0,0 +1,72 @@
+#
+# Author:: Joshua Justice (<jjustice6@bloomberg.net>)
+# Copyright:: Copyright 2008-2016, Chef Software Inc.
```
tas50:
```suggestion
# Copyright:: Copyright 2008-2019, Chef Software Inc.
```

spec/unit/provider/group/solaris_spec.rb
```diff
@@ -0,0 +1,105 @@
+#
+# Author:: Joshua Justice (<jjustice6@bloomberg.net>)
+# Copyright:: Copyright 2008-2016, Chef Software Inc.
```
tas50:
```suggestion
# Copyright:: Copyright 2008-2019, Chef Software Inc.
```

spec/unit/provider_resolver_spec.rb
jjustice6: Yeah, there is.

lib/chef/provider/group/solaris.rb
jjustice6: Tweaked comment further and did a separate commit so i could sign-off-by without having to do a force push like the others.

tas50:
@jjustice6 Hey we merged out of order so you'll need to rebase this one now. I nuked some opensuse junk in #8158

spec/unit/provider_resolver_spec.rb
tas50: 👍

jjustice6:
Is there anything else I need to do or any concerns I need to address?

tas50:
@scotthain you have some Solaris experience can you take a peak at this?

PR #8166, opened 2019-01-24
Cleanup requires / includes in knife supermarket

tas50:
1) Move everything into the deps blocks
2) don't require shell_out in install since we're not using it. We use mixlib_archive now


lib/chef/knife/supermarket_share.rb
```diff
@@ -29,11 +28,9 @@ class SupermarketShare < Knife
         require "chef/cookbook_loader"
         require "chef/cookbook_uploader"
         require "chef/cookbook_site_streaming_uploader"
-        require "mixlib/shellout"
+        require "chef/mixin/shell_out"
       end
 
-      include Chef::Mixin::ShellOut
```
tas50: If someone wonders why this is going away it's on line 25 as well

Issue #1604, opened 2014-07-03
powershell_script: flags is confusing and hard to use


docwhat:
The documentation for `flags` for the `powershell_script` resource implies you can use an array, but it requires a string.  http://docs.opscode.com/resource_powershell_script.html#attributes

Furthermore, you don't override the default flags, you add to them: 
https://github.com/opscode/chef/blob/master/lib/chef/provider/powershell_script.rb#L63-L87

I feel like it should always append `-NoLogo, -NonInteractive -InputFormat None, -File` but that `-NoProfile, -ExecutionPolicy RemoteSigned` should both be able to be overridden.

Perhaps this should be changed to two different elements:

* `use_profile` -- default `false`
* `execution_policy` -- `:remote_signed`

And then build the flags internally.

smurawski:
The docs are correct now in specifying the flags attribute takes a string.

adamedx:
@brendonmartino, thanks for your comment on leaving the script behind for debugging. This might be useful for all script resources, so feel free to open a specific issue for that.

smurawski:
There is a second issue, about the defaulting of the executionpolicy and the noprofile and I've set up https://github.com/chef/chef/issues/3057 to track that issue.

eric-tucker:
I just stumbled upon this - how is this in any way valid that Chef ignores setting a blank flags parameter?  How the hell am I supposed to set my own values?  I don't want -NoProfile as I'm trying to run Chef to configure things specific to my user account on which I'm running Chef (configuring my local machine).

DamonStamper:
+1 to allow powershell_script to load a profile.

PR #8167, opened 2019-01-25
powershell_script: Prefer user provided flags over the defaults

### LINKED ISSUE #3057: powershell_script doesn't allow for all flags to be specified
The powershell_script resource specifies -noprofile and -executionpolicy remotesigned.  Users should be able to specify an executionpolicy and allow profile scripts. 

From an earlier report in https://github.com/chef/chef/issues/1604 :

I feel like it should always append -NoLogo, -NonInteractive -InputFormat None, -File but that -NoProfile, -ExecutionPolicy RemoteSigned should both be able to be overridden.

Perhaps this should be changed to two different elements:

use_profile -- default false
execution_policy -- :remote_signed
And then build the flags internally.

Nimesh-Msys:
- Handling `flags` property at resource level instead of provider
- Fixed the order of concatenating user defined and default flags
- Added test cases
- Chef-style maintained

### Description

The order of concatenating user input flag and the default flag was incorrect.
`Powershell` would favor the flag value which is happening at last mentioned. Thus, to give priority to user input flag, we should append them at latter place.

### Issues Resolved

#3057 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/powershell_script.rb
+++ b/lib/chef/provider/powershell_script.rb
@@ -60,8 +60,8 @@ class Chef
       def flags
         interpreter_flags = [*default_interpreter_flags].join(" ")
 
-        if ! (new_resource.flags.nil?)
-          interpreter_flags = [new_resource.flags, interpreter_flags].join(" ")
+        unless new_resource.flags.nil?
+          interpreter_flags = [interpreter_flags, new_resource.flags].join(" ")
         end
 
         interpreter_flags
--- a/lib/chef/provider/powershell_script.rb
+++ b/lib/chef/provider/powershell_script.rb
@@ -48,23 +48,13 @@ class Chef
         # end of the script, it gets changed to '1'.
         #
         # Nano only supports -Command
-        cmd = "\"#{interpreter_path}\" #{flags}"
+        cmd = "\"#{interpreter_path}\" #{new_resource.flags}"
         if Chef::Platform.windows_nano_server?
           cmd << " -Command \". '#{script_file.path}'\""
         else
           cmd << " -File \"#{script_file.path}\""
         end
         cmd
-      end
-
-      def flags
-        interpreter_flags = [*default_interpreter_flags].join(" ")
-
-        unless new_resource.flags.nil?
-          interpreter_flags = [interpreter_flags, new_resource.flags].join(" ")
-        end
-
-        interpreter_flags
       end
 
       protected
@@ -78,7 +68,7 @@ class Chef
       end
 
       def validate_script_syntax!
-        interpreter_arguments = default_interpreter_flags.join(" ")
+        interpreter_arguments = new_resource.flags
         Tempfile.open(["chef_powershell_script-user-code", ".ps1"]) do |user_script_file|
           # Wrap the user's code in a PowerShell script block so that
           # it isn't executed. However, syntactically invalid script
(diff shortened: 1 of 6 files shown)


lib/chef/provider/powershell_script.rb
```diff
@@ -60,8 +60,8 @@ def command
       def flags
         interpreter_flags = [*default_interpreter_flags].join(" ")
 
-        if ! (new_resource.flags.nil?)
```
tas50: It seems like we're making this much more complex than we need to. There may be a bigger issue here, but we should be able to do this all in the resource now that we're properly using the resource DSL in the script resource. In the powershell_script resource just add a flags resource with our desired default value and also perform the join via a proc.

lamont-granquist: `[*default_interpreter_flags]` could be simplified to just `default_interpreter_flags` -- there ought to be a rubocop rule for that, because its pretty trivially redundant.

but even more broadly, it makes more sense to concatenate the arrays and then join at the end:

```
def flags
  flags = default_interpreter_flags
  flags += new_resource.flags unless new_resource.flags.nil?
  flags.join(" ")
end
```

so that's clearer about what is going on here. 

now the broader question -- yes, overriding this delegator via inheritance is a fairly terrible pattern.  moving it to the resource would be better but there needs to be a bit of caution here in the way that the default flags are concatenated and you want to make certain not to mutate the default value when you do that concatenation.

Nimesh-Msys: Updated this logic. Using flags from `resources` and removed its entire manipulation from `provider`
Thanks for the review comment :)
Could you please check and let me know in case of any issues

Nimesh-Msys:
Travis is failing from [chefspec/run_spec.rb:19](
https://github.com/chefspec/chefspec/blob/7dbf076c0a4807fe22a49defd51d146e0464b442/examples/powershell_script/spec/run_spec.rb#L19)

The specs are required to be fixed. Default value of `flags` should be appended before the user provided inputs. Hence the argument `flags: '--flags'` does not holds true at this place

lib/chef/resource/powershell_script.rb
```diff
@@ -54,6 +69,27 @@ def convert_boolean_return(arg = nil)
       def self.get_default_attributes(opts)
         { convert_boolean_return: true }
       end
+
+      # Options that will be passed to Windows PowerShell command
+      def default_flags
+        return "" if Chef::Platform.windows_nano_server?
```
tas50: We should decide later if we want to still support all these windows nano checks

lib/chef/resource/powershell_script.rb
```diff
@@ -22,6 +22,21 @@ class Resource
...
+        coerce: proc { |input|
+          if input == default_flags
+            # Means there was no input provided,
+            # and should use defaults in this case
+            input
+          else
+            # The last occurance of a flag would override its
+            # previous one at the time of command execution.
+            [default_flags, input].join(" ")
+          end
```
lamont-granquist: you should be able to just `proc { ( default_flags + input ).join(" ") }` you might need to have an `input ||= []` as well to make sure input is always an Array.  come to think if it you might want to `input.split(/\s+/) if input.is_a?(String)`

Nimesh-Msys:
> We'll need to update ChefSpec for the changes here since that spec is no longer valid.

I've sent a minor update: https://github.com/chefspec/chefspec/pull/950
Could you please check and merge, I'll re-trigger this PR to ensure the checks once again.
Thank you

lib/chef/resource/powershell_script.rb
Nimesh-Msys: Right sir, but as per the [Document](https://docs.chef.io/resource_powershell_script.html), which was corrected after [this discussion](https://github.com/chef/chef/issues/1604#issuecomment-78310266), `flags` are of String type and not Array. Even the multiple flags are also passed within the same string simply separated by space.

tas50:
This is the right thing to do, but it's technically a breaking change so we probably shouldn't backport it. We should call it out in a positive light in the release notes for Chef 15.

PR #8168, opened 2019-01-25
Fix for write permissions were not working properly on windows

### LINKED ISSUE #4431: deny_rights :write, "Administrators" does not function properly on Windows
for the directory resource, deny_rights :write, "Administrators" adds two permissions to the resource:

a Deny "Write" permission on subfolders and files
a Deny "Write, Special" permission on this folder only - which includes Deny, "Read Permissions"

On Windows, Deny Read Permissions actually prevents the end user from opening the folder/file, so deny_rights :write is actually denying read as well.

vijaymmali1990:
- Minor fixes to use `WRITE` with `:deny_rights` and `:rights` options of `directory` resource
- Fixed test cases
- Added functional test cases for deny_rights option
- Ensured chef style
- Fixes MSYS-958

### Description

- Fixes were required in `create_ace_with_mask_and_sid` method to fix `access_allowed` and `access_denied` methods.
- Correct value of [WRITE](https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.filesystemrights?view=netframework-4.7.2#fields) must be used to allow/deny that rights of a directory.

### Issues Resolved

Fixes #4431 

### Check List

- [x] All tests pass


tas50:
@vijaymmali1990 There are appveyor spec failures at the moment

lib/chef/file_access_control/windows.rb
```diff
@@ -90,11 +90,13 @@ def acls_equal(target_acl, actual_acl)
         target_acl.each do |target_ace|
           if target_ace.flags & INHERIT_ONLY_ACE == 0
             self_ace = target_ace.dup
-            self_ace.flags = 0
+            # We need flag value which is already being set in case of WRITE permissions as 3, so we will not be overwriting it with the hard coded value.
+            self_ace.flags = 0 unless target_ace.mask == Chef::ReservedNames::Win32::API::Security::WRITE
```
marcparadise: I can't find `Security::WRITE` in `chef/win32/api/file.rb` - coudl this be intended to be GENERIC_WRITE, or FILE_GENERIC_WRITE?

marcparadise: Is the intent to ensure that only the value '::WRITE' (3) is accepted? Or do we want to mask out any other options, so if `target_ace` has `flags` of `11`, we should retain the 3 and drop the rest: 

`self_ace.flags &= ...Security::WRITE`

lib/chef/file_access_control/windows.rb
```diff
@@ -90,11 +90,13 @@ def acls_equal(target_acl, actual_acl)
...
             self_ace.mask = securable_object.predict_rights_mask(target_ace.mask)
             new_target_acl << self_ace
           end
-          if target_ace.flags & (CONTAINER_INHERIT_ACE | OBJECT_INHERIT_ACE) != 0
+          # As there is no inheritence needed in case of WRITE permissions.
+          if target_ace.mask != Chef::ReservedNames::Win32::API::Security::WRITE && target_ace.flags & (CONTAINER_INHERIT_ACE | OBJECT_INHERIT_ACE) != 0
```
marcparadise: If the original code is correct (assigning WRITE and not ANDing it) then this is fine; but if we do want to AND we'll need to update target_ace.mask comparisons to AND the WRITE flag, `(target_ace.mask & ...WRITE) == 0 && ... `

btm:
@jasonwbarnett any thoughts on this?

PR #8178, opened 2019-01-28
Convert execute_resource remaining properties to use properties 

### LINKED ISSUE #7063: Convert execute resource to setup properties via property DSL
The execute resource has not been fully converted to using property to set properties. We should convert this to setup properties via the property resource DSL.

Vasu1105:

### Description
The execute resource has not been fully converted to using property to set properties. We should convert this to setup properties via the property resource DSL.

### Issues Resolved
Fixes #7063
### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/resource/execute.rb
+++ b/lib/chef/resource/execute.rb
@@ -23,10 +23,6 @@ class Chef
   class Resource
     class Execute < Chef::Resource
       resource_name :execute
-      provides :execute
-
-      identity_attr :command
-
       description "Use the execute resource to execute a single command. Commands that"\
                   " are executed with this resource are (by their nature) not idempotent,"\
                   " as they are typically unique to the environment in which they are run."\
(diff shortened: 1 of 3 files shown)


lib/chef/resource/execute.rb
```diff
@@ -23,10 +23,6 @@ class Chef
   class Resource
     class Execute < Chef::Resource
       resource_name :execute
-      provides :execute
```
tas50: ?

Vasu1105: I didn't felt removing this is making any difference and it's required as I compared this with other resources. Let me know if I am missing something here.

lamont-granquist: it is technically redundant with resource_name, but I'd like to deprecate resource_name, so having the `provides :execute` line in here is more forwards-looking.

Vasu1105: @lamont-granquist Fixed.

PR #8186, opened 2019-01-29
Update to Bundler 2.0.1

tas50:
Bring Chef 15 up to Bundler 2.x


tas50:
Closing this since I don't think we should actually pull in Bundler 2 at this point in time

Issue #7985, opened 2018-11-23
chef-client run when providing an empty string argument produces incongruent error


denchang:
## Description
Running the command chef-client '' (with an empty string as input) produces an error message that is difficult to interpret. I consider it a bug because the empty string as input argument is innocuous and perhaps should not be accepted. However, when providing this argument to the client, we get an output error, like https://github.com/chef-cookbooks/chef-client/issues/481

It appears that the PWD environment variable value is eventually passed downstream to

variable specific_recipes, contained in file:

lib/chef/policy_builder/expand_node_object.rb

Producing the error EISDIR.

## Chef Version
Tested on chef-client version 12.19.36, 13.4.24, 14.1.7 (latest one)

## Platform Version
Ubuntu 18.04 LTS

## Replication Case
cd /root
chef-client ''
should produce an EISDIR error indicating that no recipes can be found in /root because it's a directory.

## Client Output
https://github.com/chef-cookbooks/chef-client/issues/481

The debug output (chef-client -l debug) may be useful, but please link to a gist, or truncate it. -->

```

```

PR #8200, opened 2019-02-04
Fixed issue for chef-client run was throwing error when provided empty string with it

vinay033:
Signed-off-by: vinay033 <vinay.sharma@msystechnologies.com>

### Description

If you run the chef-client with the empty string option then it throws an error. Added fix for it. Added a check if string is a file or not.

### Issues Resolved

Fixes https://github.com/chef/chef/issues/7985
### Check List

- [x] All tests pass


lamont-granquist:
what is the error message the user gets back in this case?

Vasu1105:
> what is the error message the user gets back in this case?

@lamont-granquist since @vinay033 is on leave replying here User will get message as ```[2019-02-18T09:37:36+00:00] FATAL: IOError: Cannot open or read /home/azure!``` 

Which may not still resolve the purpose of the issue we are resolving.

We identified this case while running chef-client on node.
when we do not pass any arguments, (nil), we do not set `chef_config[:specific_recepies]` https://github.com/chef/chef/blob/master/lib/chef/application.rb#L157:L159
But if we pass blank string, File.expand_path("") returns Dir.pwd. Here is the output:
```// Blank string returns Dir.pwd
irb(main):001:0> File.expand_path("")
=> "C:/Windows/System32" (output in case of Windows)
=> "/home/msys/workspace/chef" (output in case of Ubuntu 18.04)

// The case when string contains only spaces
irb(main):002:0> File.expand_path(" ")
=> "" (out put in case of Windows)
=> "/home/msys/workspace/chef/ " ( output in case of Ubuntu 18.04)```
The returned values are then set as `chef_config[:specific_recipes]` which gives uncertain results later on. Like nil, if we consider blank strings(also with only spaces) to be the same case and do not set `specific_recipes` we can seamlessly run `chef-client`

**Comment by lamont-granquist** [comment]
yeah it'd be nice to wrap the Exception being thrown to give some indication that chef is trying to read a recipe at that path.

keep the `file?` check here since that is the correct way to do that.

the set_specific_recipes method should probably do some kind of:

```ruby
  rescue => e
    raise "Cannot read recipe file '#{chef_config[:specific_recipies]}': #{e.message}"`
  end
```

that way the output should become:

```
[2019-02-18T09:37:36+00:00] FATAL:  Cannot read recipe file '':  IOError: Cannot open or read /home/azure!
```

Which is probably good enough and helps in general with any command line argument put there.

vinay033:
> ```ruby
> raise "Cannot read recipe file '#{chef_config[:specific_recipies]}': #{e.message}"`
> ```

I checked set_specific_recipe method but this method does not produce any error, in this method  cli_argument is just an array.
if you pass a blank string with the chef-client like: $ chef-client ' ' 
then cli_argument contains a blank array with a blank string like [" "] 

we can do with this method like 
`
def set_specific_recipes
   if cli_arguments.respond_to?(:map)
        if cli_arguments.size == 1 && File.file?(cli_arguments[0])
          chef_config[:specific_recipes] =
            cli_arguments.map { |file| File.expand_path(file) }
        else
          raise "No recipes can be found because it's a directory."
       end
   end
end
`

vinay033:
@lamont-granquist 
`cli_arguments` array contains strings denoting the file paths.
In some situation (ex:  $ chef-client '' ), we found that these string can even be empty('') and in that case, File.expand_path('') returns the Dir.pwd itself.
There are no exceptions in this operation, But we also want to skip this infiltered empty string from here.

One way could be, as per the suggestions, that whenever we found any invalid(or empty) string at `set_specific_recipes`, we can raise an error:
```
def set_specific_recipes
   if cli_arguments.respond_to?(:map)
      chef_config[:specific_recipes] =
      cli_arguments.map do |file|
         raise if !File.file?(file)
         File.expand_path(file)
      end
   end
rescue => e
    raise IOError, "Cannot open or read #{File.expand_path(file)}!"
end

This code might look untidy, but more important, it's unreachable code.
As per the testing, I found redesigning `from_file` in this optimum way would anyway raise the expected error.

and also one more important thing is that if we raise an error through the `set_specific_recipe` then output error display with the Traceback (most recent call last).
like: 
```
	6: from /usr/bin/chef-client:296:in `<main>'
	5: from /usr/bin/chef-client:296:in `load'
	4: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.7.17/bin/chef-client:25:in `<top (required)>'
	3: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.7.17/lib/chef/application.rb:64:in `run'
	2: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.7.17/lib/chef/application/client.rb:317:in `reconfigure'
	1: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.7.17/lib/chef/application.rb:162:in `set_specific_recipes'
/opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/chef-14.7.17/lib/chef/application.rb:171:in `rescue in set_specific_recipes': Cannot open or read /home/ad.msystechnologies.com/vinay.sharma/chef-repo! (IOError)
```

File: lib/chef/application.rb [hunk 1]
```diff
@@ -153,9 +153,13 @@ def apply_extra_config_options(extra_config_options)
     end
 
     def set_specific_recipes
-      if cli_arguments.respond_to?(:map)
+      if cli_arguments.is_a?(Array) &&
+          (cli_arguments.empty? || cli_arguments.all? { |file| File.file?(file) } )
         chef_config[:specific_recipes] =
           cli_arguments.map { |file| File.expand_path(file) }
+      else
+        Chef::Application.fatal!("IOError: Cannot open or read: \"" +
```
btm: This may not be a file, e.g. if someone ran `chef-client ""` we're getting an empty string. that could be confusing.

We probably shouldn't mention "IOError" since we aren't raising that exception.

btm:
Please add unit test coverage for `set_specific_recipes`.

lib/chef/application.rb
vinay033: Ok, I will add specs as well as changes for error message.

lib/chef/mixin/from_file.rb
```diff
@@ -30,7 +30,7 @@ module FromFile
       # Raises an IOError if the file cannot be found, or is not readable.
       def from_file(filename)
         self.source_file = filename
-        if File.exists?(filename) && File.readable?(filename)
+        if File.exists?(filename) && File.readable?(filename) && File.file?(filename)
```
marcparadise: Since we're checking .file? we can remove .exists? : 
`if File.file?(filename) && File.readable?(filename)`

lib/chef/mixin/from_file.rb
```diff
@@ -43,7 +43,7 @@ def from_file(filename)
       # Raises an IOError if the file cannot be found, or is not readable.
       def class_from_file(filename)
         self.source_file = filename
-        if File.exists?(filename) && File.readable?(filename)
+        if File.exists?(filename) && File.readable?(filename) && File.file?(filename)
```
marcparadise: As above

lib/chef/application.rb
vinay033: I have added test cases and updated the error message.

vinay033:
> Please add unit test coverage for `set_specific_recipes`.

I have added unit test cases for the `set_specific_recipes`

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
   end
 
+  describe "#set_specific_recipes" do
+    let(:app) { Chef::Application.new }
+    context "when cli arguments does not containing any value" do
+      before do
+        allow(app).to receive(:cli_arguments).and_return([])
+      end
+
+      it "returns a empty array" do
```
tas50:
```suggestion
      it "returns an empty array" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
 
   end
 
+  describe "#set_specific_recipes" do
+    let(:app) { Chef::Application.new }
+    context "when cli arguments does not containing any value" do
```
tas50:
```suggestion
    context "when cli arguments does not containing any values" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return([])
+      end
+
+      it "returns a empty array" do
+        app.set_specific_recipes
+        expect(Chef::Config[:specific_recipes]).to eq([])
+      end
+    end
+
+    context "when cli arguments containing valid recipe file path" do
```
tas50:
```suggestion
    context "when cli arguments contain valid recipe file path" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return([tempfile])
+      end
+
+      it "sets the specific recipes to config" do
+        app.set_specific_recipes
+        expect(Chef::Config[:specific_recipes]).to eq([tempfile])
+      end
+    end
+
+    context "when cli arguments containing invalid recipe file path" do
```
tas50:
```suggestion
    context "when cli arguments contain invalid recipe file path" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return([tempfile])
+        allow(Chef::Application).to receive(:fatal!).and_return(fatal)
+      end
+
+      it "raises an error with application exit" do
+        expect(app.set_specific_recipes).to eq(fatal)
+      end
+    end
+
+    context "when cli arguments containing empty string" do
```
tas50:
```suggestion
    context "when cli arguments contain empty string" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return([""])
+        allow(Chef::Application).to receive(:fatal!).and_return(fatal)
+      end
+
+      it "raises an arguments error" do
+        expect(app.set_specific_recipes).to eq(fatal)
+      end
+    end
+
+    context "when cli arguments containing any string" do
```
tas50:
```suggestion
    context "when cli arguments contain any string" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return(["test"])
+        allow(Chef::Application).to receive(:fatal!).and_return(fatal)
+      end
+
+      it "raises an arguments error" do
+        expect(app.set_specific_recipes).to eq(fatal)
+      end
+    end
+
+    context "when cli arguments containing multiple invalid strings" do
```
tas50:
```suggestion
    context "when cli arguments contain multiple invalid strings" do
```

spec/unit/application_spec.rb
```diff
@@ -386,6 +386,94 @@
...
+        allow(app).to receive(:cli_arguments).and_return(["", "test"])
+        allow(Chef::Application).to receive(:fatal!).and_return(fatal)
+      end
+
+      it "raises an arguments error" do
+        expect(app.set_specific_recipes).to eq(fatal)
+      end
+    end
+
+    context "when cli arguments containing valid recipe file path and invalid string" do
```
tas50:
```suggestion
    context "when cli arguments contain valid recipe file path and invalid string" do
```

lib/chef/application.rb
```diff
@@ -153,9 +153,13 @@ def apply_extra_config_options(extra_config_options)
     end
 
     def set_specific_recipes
```
tas50: while you're here can we get a YARD comment added to this method

vinay033: sure, I will add YARD comment for this method.

spec/functional/mixin/from_file_spec.rb
```diff
@@ -21,6 +21,7 @@
   REAL_DATA = File.join(CHEF_SPEC_DATA, "mixin", "real_data.rb")
   INVALID_DATA = File.join(CHEF_SPEC_DATA, "mixin", "invalid_data.rb")
   NO_DATA = File.join(CHEF_SPEC_DATA, "mixin", "non_existant_data.rb")
+  DIRECTORY = File.expand_path("")
```
zenspider: Huh. I've never even though of doing that. I always use `"."`. Thank you for teaching me something!

This should be the same as `Dir.pwd` which might be a bit clearer. This is just a comment tho... not a blocking request for change.

PR #8219, opened 2019-02-07
Chef::Config: Uniform config dir path separator

### LINKED ISSUE #4909: Chef::Config[:file_cache_path] mixed path style
## Description

When accessing the variable class Chef::Config[:file_cache_path] on windows the path returned is not completely windows like. This could be tricky when passing paths to legacy commands.
I would like to have an uniform path style that for windows is the legacy one.
## Chef Version

12.9.41
## Platform Version

tested on windows 8.1, same for others.
## Replication Case

open chef-shell  and in recipe_mode  write Chef::Config[:file_cache_path] 
## Client Output

``` text
"C:/Users/myuser\\.chef\\cache"
```

vsingh-msys:

### Description

 - Config dir is extracted from the `config_file` if provided otherwise it is set to the `user_home` path and based on the environments (windows or Linux) we provide canonical path but when we set `user_home` it returns as it is, So In order to make it uniform we are returning environments based uniform path for `user_home` dir as well.

### Issues Resolved

#4909

### Check List

- [x] All tests pass

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -122,7 +122,7 @@ module ChefConfig
       if config_file
         PathHelper.dirname(PathHelper.canonical_path(config_file, false))
       else
-        PathHelper.join(user_home, ".chef", "")
+        PathHelper.join(PathHelper.canonical_path(user_home, false), ".chef", "")
       end
     end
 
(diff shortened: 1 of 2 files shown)


tas50:
@vsingh-msys can we get a spec for this?

vsingh-msys:
@tas50 thanks for pointing out for spec, Got some of the minor changes in specs and also added some windows environments specs please have a look. Thanks!

PR #8229, opened 2019-02-12
windows_certificate: Add support to import Base 64 encoded CER certificates

Nimesh-Msys:
- Appending "- inform DER" only if the certificate is binary encoded. It defaults to "- inform PEM" which fetches Base64 encoded certificates.
- Added test cases.
- Ensured Chef style.

### Description

- Till now, CER certificates were only considered to be in binary format (DER).
- They can also be Base-64 encoded (PEM).
- We should only append "inform DER" only if it is a binary certificate, otherwise, default ("inform PEM") flag would allow us to fetch a Base-64 encoded certificates.

### Issues Resolved

Found this cases while testing a [windows cookbook PR](https://github.com/chef-cookbooks/windows/pull/600)

### Check List

- [x] All tests pass


lib/chef/resource/windows_certificate.rb
```diff
@@ -277,7 +277,9 @@ def acl_script(hash)
         def convert_pem(ext)
           out = case ext
                 when ".crt", ".cer", ".der"
-                  powershell_out("openssl x509 -text -inform DER -in #{new_resource.source} -outform PEM")
+                  command = "openssl x509 -text -in #{new_resource.source} -outform PEM"
```
tas50: Is this something we can just do directly in ruby vs. shelling out?

lib/chef/resource/windows_certificate.rb
```diff
@@ -277,7 +277,9 @@ def acl_script(hash)
...
+                  command += " -inform DER" if binary_cert?
+                  powershell_out(command)
```
tas50: If we can't do this with pure ruby we should use powershell_out! here so it fails if this fails

lib/chef/resource/windows_certificate.rb
```diff
@@ -299,6 +301,11 @@ def format_raw_out(out)
           end_cert = "-----END CERTIFICATE-----"
           begin_cert + out[/#{begin_cert}(.*?)#{end_cert}/m, 1] + end_cert
         end
+
+        # Checks if the certificate is binary encoded or not
+        def binary_cert?
+          powershell_out("file -b --mime-encoding #{new_resource.source}").stdout.strip == "binary"
```
tas50: I'm fairly certain we can do this in pure ruby as well and avoid the extra time it takes to spin up a powershell instance

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Yes, In ruby there is an `openssl` library which can easily perform all such operations.
```
OpenSSL::X509::Certificate.new(File.read(base64)).to_pem
=> CERTIFICATE CONTENT
```
But It doesn't seems to be working for all the cases. Like for DER(and even PFX, with keys) formats, it does not give the expected results:
```
OpenSSL::X509::Certificate.new(File.read(der))
=> OpenSSL::X509::CertificateError (not enough data)
```

Not sure, but may be due to this reason, openssl utility via powershell was preferred here.

But if it is permitted then _we can totally skip all shellouts, `openssl` and conversion operations from here_ and can use Crypt API in `win32-certstore`.

Also, due to this implementation we are importing all the certificates in PEM format into the certificate store.  and we've seen its drawback like data loss(private keys of PFX certificates).

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Actually, I deliberately removed `powershell_out!`. In case of any error, it shows all the commands and details which was not looking user friendly. Pretty printing of an error was the only reason that I used `powershell_out`. I can revert back the code in case of any issues. Please do let me know.

Here is the case:

**powershell_out!("Some invalid command")**
```
*** Mixlib::ShellOut::ShellCommandFailed Exception: Expected process to exit with [0], but received '1'
---- Begin output of powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "Some invalid command" ----
STDOUT:
STDERR: Some : The term 'Some' is not recognized as the name of a cmdlet, function, script file, or operable program. Check
the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ Some invalid command
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (Some:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
---- End output of powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "Some invalid command" ----
Ran powershell.exe -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command "Some invalid command" returned 1

```


**powershell_out("Some invalid command").stderr**

```
"Some : The term 'Some' is not recognized as the name of a cmdlet, function, script file, or operable program. Check \r\nthe spelling of the name, or if a path was included, verify that the path is correct and try again.\r\nAt line:1 char:1\r\n+ Some invalid command\r\n+ ~~~~\r\n    + CategoryInfo          : ObjectNotFound: (Some:String) [], CommandNotFoundException\r\n    + FullyQualifiedErrorId : CommandNotFoundException\r\n \r\n"
```

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Right @tas50 , one way of doing this could be by checking `String.encoding` which is not that reliable or we can also use `String.ascii_only?`. In either case, we need to read the file contents.
Other option I found is by using `Open3`. I tried them and chose the last one based on these Benchmark results.

```
def open3
  file_details, status = Open3.capture2e("file", cert_path)
  file_details.rpartition(":").last.strip == "PEM certificate"
end

def ascii_only
  File.read(cert_path).ascii_only?
end

Benchmark results:

open3 => 0.03903859999263659
ascii_only => 11.53894020000007
```

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: Hi @tas50 , I've just realized and found this, for DER(binary) certificates, we can get the contents by using **File.binread** :)

Nimesh-Msys:
Hi @tas50 :tophat: 
I have removed all the shell interactions and converted the expected operations in pure ruby way. With this, the code is really clean, reliable and optimized now.
Thank you again reviewing the code. :pizza: :fries: :beer:

Could you please take a look and let me know in case of any issues.

lib/chef/resource/windows_certificate.rb
Nimesh-Msys: `open3` didn't solved the purpose. We were still relying on shelling. powershell_out! is used since the command relies on `file.exe` which is available on powershell, but not on cmd.

tas50:
Looks good to me now. Thanks

PR #8247, opened 2019-02-21
Chef-13: add lazy module include to universal DSL

lamont-granquist:
backport of #8243


lamont-granquist:
test failures got nothing to do with this...

PR #8252, opened 2019-02-25
Fix for cron resource get confused by environement/property mismatch

### LINKED ISSUE #8241: cron resource get confused by environement/property mismatch
## Description

If you use a `environment` for an entry that can also be specified as a  `property`, the `cron` resource will update every run.

## Chef Version
13.9, but based on code, I suspect all versions

## Platform Version
RHEL7

## Replication Case

```
cron 'test' do
  command '/bin/true'
  environment({'MAILTO' => 'root'})
end
```

vijaymmali1990:
- Added some changes in cron.rb
- It now throws error if user passes both `environment` with the `{:SHELL, :HOME, :PATH, :MAILTO}` and these `properties` itself.
- It maintains `idempotency`.
- Added unit tests for this case.
- Ensured chef-style.

### Description
When we use `environment` for an entry that can also be specified as a `property` i.e. `:SHELL` `:PATH` `:MAILTO` `:HOME` it causes `idempotency` issues. Added a fix for the same. Now it throws error if we pass both `environment` with the `{:SHELL, :HOME, :PATH, :MAILTO}` and these `properties` itself. And works fine in case we pass `environment` with the `{:SHELL, :HOME, :PATH, :MAILTO}` only.

### Issues Resolved
Fixes #8241 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/cron.rb
+++ b/lib/chef/provider/cron.rb
@@ -32,6 +32,7 @@ class Chef
       CRON_PATTERN = /\A([-0-9*,\/]+)\s([-0-9*,\/]+)\s([-0-9*,\/]+)\s([-0-9*,\/]+|[a-zA-Z]{3})\s([-0-9*,\/]+|[a-zA-Z]{3})\s(.*)/.freeze
       SPECIAL_PATTERN = /\A(@(#{SPECIAL_TIME_VALUES.join('|')}))\s(.*)/.freeze
       ENV_PATTERN = /\A(\S+)=(\S*)/.freeze
+      ENVIRONMENT_PROPERTIES = %w{MAILTO PATH SHELL HOME}.freeze
 
       def initialize(new_resource, run_context)
         super(new_resource, run_context)
@@ -89,16 +90,9 @@ class Chef
       end
 
       def cron_different?
-        logger.warn("It is not idempotent to use `environment` for an entry that can be specified as a `property`") if property_with_environment?
         CRON_ATTRIBUTES.any? do |cron_var|
           new_resource.send(cron_var) != current_resource.send(cron_var)
         end
-      end
-
-      # Returns true if user is using a `environment` for an entry that can also be specified as a `property`
-      def property_with_environment?
-        common_attribute = %w{MAILTO PATH SHELL HOME} & new_resource.environment.keys
-        common_attribute.any?
       end
 
       def action_create
@@ -199,7 +193,7 @@ class Chef
       private
 
       def set_environment_var(attr_name, attr_value)
-        if %w{MAILTO PATH SHELL HOME}.include?(attr_name)
+        if ENVIRONMENT_PROPERTIES.include?(attr_name)
           current_resource.send(attr_name.downcase.to_sym, attr_value.gsub(/^"|"$/, ""))
         else
           current_resource.environment(current_resource.environment.merge(attr_name => attr_value))
@@ -228,7 +222,18 @@ class Chef
           newcron << "#{v.to_s.upcase}=\"#{new_resource.send(v)}\"\n" if new_resource.send(v)
         end
         new_resource.environment.each do |name, value|
-          newcron << "#{name}=#{value}\n"
+          if ENVIRONMENT_PROPERTIES.include?(name)
+            unless new_resource.property_is_set?(name.downcase)
+              logger.warn("#{new_resource.name}: the environment property contains the '#{name}' variable, which should be set separately as a property.")
+              new_resource.send(name.downcase.to_sym, value.gsub(/^"|"$/, ""))
+              new_resource.environment.delete(name)
+              newcron << "#{name.to_s.upcase}=\"#{value}\"\n"
+            else
+              raise Chef::Exceptions::Cron, "#{new_resource.name}: the '#{name}' property is set and environment property also contains the '#{name}' variable. Remove the variable from the environment property."
+            end
+          else
+            newcron << "#{name}=#{value}\n"
+          end
         end
         if new_resource.time
           newcron << "@#{new_resource.time} #{new_resource.command}\n"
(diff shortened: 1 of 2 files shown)


vijaymmali1990:
Do we need to change the docs for `cron` resource.

lib/chef/provider/cron.rb
```diff
@@ -220,9 +220,14 @@ def get_crontab_entry
         [ :mailto, :path, :shell, :home ].each do |v|
           newcron << "#{v.to_s.upcase}=\"#{new_resource.send(v)}\"\n" if new_resource.send(v)
         end
+        idempotency = true
         new_resource.environment.each do |name, value|
+          if %w{MAILTO PATH SHELL HOME}.include?(name)
```
jaymzh: This is certainly better than what we have today, but it feels like we could just adjust `cron_different?` to compare these better?

Vasu1105: @vijaymmali1990 I feel this should go in resource.

lib/chef/provider/cron.rb
```diff
@@ -220,9 +220,14 @@ def get_crontab_entry
         [ :mailto, :path, :shell, :home ].each do |v|
           newcron << "#{v.to_s.upcase}=\"#{new_resource.send(v)}\"\n" if new_resource.send(v)
         end
+        idempotency = true
```
Vasu1105: This idempotency is not required if you know that the resource is not idempotent if we are using the same environment variable which is declared in properties.

lib/chef/provider/cron.rb
vijaymmali1990: Yes @jaymzh I have done some changes in `cron_different?` and I have also tested these which gives us positive results.

vijaymmali1990: @Vasu1105 no we can not add this in resource as for the `environment` variable it does not go in resource file.

lib/chef/provider/cron.rb
```diff
@@ -89,11 +89,18 @@ def load_current_resource
...
 
       def cron_different?
+        property_with_environment?
         CRON_ATTRIBUTES.any? do |cron_var|
           new_resource.send(cron_var) != current_resource.send(cron_var)
         end
       end
 
+      # Raises warning if user is using a `environment` for an entry that can also be specified as a `property`
+      def property_with_environment?
```
Vasu1105: @vijaymmali1990 we need to revisit this this is wrong implementation of method in ruby method with ? should always return boolean type.

lib/chef/provider/cron.rb
```diff
@@ -89,11 +89,18 @@ def load_current_resource
       end
 
       def cron_different?
+        logger.warn("It is not idempotent to use `environment` for an entry that can be specified as a `property`") if property_with_environment?
```
clintoncwolfe: I found this error message confusing - it would be better if it was more explicit, such as "It looks like you are trying to set the MAILTO environment variable.  Please use the `mailto` resource property instead."

btm:
In `cron_different?` we compare all the properties (f.k.a. attributes) of the current resource with the new resource.

If we have this resource:
```
cron 'noop' do
  hour '5'
  minute '0'
  command '/bin/true'
  environment({'MAILTO' => 'admin@example.org'})
end
```

If that cron entry already exists, when we read it we will set the mailto property for current_resource in `set_environment_var`.

However the mailto property for new_resource will be empty, because the user didn't specify it.

So `cron_different?` will return true and we will run the action.

The solution to this (99% sure) is to make sure we parse the special variables out of the environment property before/in  `cron_different?`. Which would look something like this:

```
  ENVIRONMENT_PROPERTIES = %w{MAILTO PATH SHELL HOME}.freeze

      def cron_different?
        new_resource.environment.each do |name, value|
          if ENVIRONMENT_PROPERTIES.include?(name)
            unless new_resource.property_is_set?(name)
              logger.warn("#{new_resource.name}: the environment property contains the '#{name}' variable, which should be set separately as a property.")
              new_resource.send(name.downcase.to_sym, attr_value.gsub(/^"|"$/, ""))
              new_resource.environment.delete(name)
            else
              raise Chef::Exceptions::Cron, "#{new_resource.name}: the '{name}' property is set and environment property also contains the '#{name}' variable. Remove the variable from the environment property."
            end
          end
        end

        CRON_ATTRIBUTES.any? do |cron_var|
          new_resource.send(cron_var) != current_resource.send(cron_var)
        end
      end
```

vijaymmali1990:
@btm yes we had a thought upon it and yesterday we needed to discuss with you this thing only. Anyways Thank you!
Will test it around and update!

lib/chef/provider/cron.rb
```diff
@@ -221,7 +222,18 @@ def get_crontab_entry
           newcron << "#{v.to_s.upcase}=\"#{new_resource.send(v)}\"\n" if new_resource.send(v)
         end
         new_resource.environment.each do |name, value|
-          newcron << "#{name}=#{value}\n"
+          if ENVIRONMENT_PROPERTIES.include?(name)
+            unless new_resource.property_is_set?(name.downcase)
+              logger.warn("#{new_resource.name}: the environment property contains the '#{name}' variable, which should be set separately as a property.")
+              new_resource.send(name.downcase.to_sym, value.gsub(/^"|"$/, ""))
+              new_resource.environment.delete(name)
+              newcron << "#{name.to_s.upcase}=\"#{value}\"\n"
```
btm: why the addition of the double quotes around value? Those weren't present before. That would trigger an update for everyone?

jaymzh: Presumably to match line 222?

Nimesh-Msys: I think, may be to support spaces in values ?
Like in PATH where some directory name could contain spaces.

tas50:
Fixed in #8455

PR #8253, opened 2019-02-26
Refactor bootstrapping to use train as the transport with full Windows bootstrap support

marcparadise:
### Description
Update bootstrap to use Train/TargetHost via chef-core, and moves windows bootstrap into core knife bootstrap

### Remaining Tasks

- [x] remove windows bootstrap context support for v < 12
- [x] move options back into main class
- [x] remove --prerelease
- [x] remove install support for v < 12.20
- [x] alias flags with changed names, deprecations
Work to remove the config resolution in bootstrap will follow in a separate PR after merge. 
- [x] rename chef_core* to chef-core . Also merged chef-core-cliux into chef-core, but left actions separate in order to prevent pulling in an indirect circular dependency of ChefDK -> Chef. 
- [x] Move chef_core-* into a single repository. 
- [x] verify no chef circular dependency in chef-core  
- [x] publish chef-core gem to avoid dependency failures from gem install.  
- [x] update gem lockfile to add windows gem
- [x] add design doc to PR 

#### Post-merge

x'd items mean an appropriate issue or PR has been created and linked. Aside from the switch to train-core, the items below are not blockers for Chef 15. 

- [ ] switch to train-core - on hold, doing this requires a versioin bump to InSpec 4 which is not yet released.  
    * Train core can't be reconciled with inspec v3's constraints, so this is deferred until after inspec 4. 
- [ ] use mixlib-install api to manage creating the download script for windows/nix.  I will open a design proposal for this change. 
- [x] #8394 improvements to ensure all config keys are available in `Knife#config`
- [ ] remove Chef::Config setter procs from options in bootstrap 
  * this is a larger change because it affects bootstrap, contexts, and templates which don't know about unified config and make assumptions about where config is sourced from. 
  * depends on [chef#8934]
- [ ] determine which Windows versions in the template are valid and supported; remove ones that aren't.

### Check List

- [x] All tests pass


Gemfile
```diff
@@ -13,6 +13,16 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master"
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "cheffish", "~> 14"
 
+# The chef_core gems are sourced from git.
+# gem "chef_core", git: "https://github.com/chef/chef_core", branch: "chef-core-split"
+# gem "chef_core-actions", git: "https://github.com/chef/chef_core-actions", branch: "chef-core-split"
+# gem "chef_core-cliux", git: "https://github.com/chef/chef_core-cliux", branch: "chef-core-split"
+# Temporary for testing:
+gem "train", path: "../train"
```
tas50: you're going to want train-core here not the full blown train

Gemfile
```diff
@@ -13,6 +13,16 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master"
...
+gem "chef_core", path: "../chef_core"
+gem "chef_core-actions", path: "../chef_core-actions"
+gem "chef_core-cliux", path: "../chef_core-cliux"
```
lamont-granquist: Few issues some trivial some not so much:

- should not use underscores in gem names.
- should probably not be immediately extracting out 3 different dependent gems if possible, i'd prefer to see 1 or may be 2
- the "chef core" name is probably poor since it suggests to me something that its not clear to me that that is
- the dependent gems MUST go into the chef repo itself similar to the chef-config gem, i can push some code i was working on for the chef-helpers gem which can show how.  the way forwards for sub-gems of chef client needs to be more monorepoish and lockstepping with the versions, the mixlib style is poor (talk to @tas50 about the joys of major version bumping the universe which should be freshly burned into his brain).   this is very, very hard 'no' on the direction here with additional gem repos.
- since knife is requiring chef_core files it needs to appear in the gemspec.  that looks like it ultimately creates circular deps here:  https://github.com/chef/chef_core-actions/blob/master/chef_core-actions.gemspec#L40-L42 which is either a mistake or indicates  problem in modelling how this is being extracted.

its not "no don't do this" but i have deep concerns about the way that this is being done.

lib/chef/knife/bootstrap/options.rb
```diff
@@ -0,0 +1,356 @@
...
+              description: "The SSH identity file used for authentication"
+
+            # ssh options - train options[:verify_host_key]
+            option :ssh_verify_host_key,
+              long: "--ssh-[no-]verify-host-key",
+              description: "Verify host key, enabled by default.",
+              boolean: true
+
+            # argument to installer in chef-full, via bootstrap_context
+            option :prerelease,
```
tas50: We should have a discussion at some point if we need this option. We don't release pre-release gems anymore so it's basically a useless option.

Gemfile
lamont-granquist: Some READMEs in the dependent gems or a few paragraphs of a design doc somewhere explaining what and why would also help.

lib/chef/knife/bootstrap/options.rb
```diff
@@ -0,0 +1,356 @@
...
+                Chef::Config[:knife][:bootstrap_vault_item] ||= {}
+                Chef::Config[:knife][:bootstrap_vault_item][vault] ||= []
+                Chef::Config[:knife][:bootstrap_vault_item][vault].push(item)
+                Chef::Config[:knife][:bootstrap_vault_item]
+              }
+
+              # Windows only
+
+              # bootstrap template
+            option :install_as_service,
```
tas50: We should also remove this since we've long ago deprecated this option in favor of windows task based installations

lib/chef/knife/bootstrap/templates/windows-chef-client-msi.erb
```diff
@@ -0,0 +1,271 @@
...
+@rem If this is an unknown version of windows set the default
+@set MACHINE_OS=2012r2
+@echo Warning: Unknown version of Windows, assuming default of Windows %MACHINE_OS%
+goto architecture_select
+
+:Version6.0
+@set MACHINE_OS=2008
+goto architecture_select
+
+:Version5.2
```
tas50: Do we need windows 2003 logic in here anymore since we don't support that at all in chef or have MSIs for it

lib/chef/knife/bootstrap/templates/windows-chef-client-msi.erb
```diff
@@ -0,0 +1,271 @@
...
+@rem Currently Windows Server 2012 R2 is treated as equivalent to Windows Server 2012
+:Version6.3
+@set MACHINE_OS=2012r2
+goto architecture_select
+
+:Version10.0
+@set MACHINE_OS=2016
+goto architecture_select
+
+@rem Currently Windows Server 2016 R2 is treated as equivalent to Windows Server 2016
```
tas50: I'm really confused by this. There is no 2016 R2. Any idea what this might be @stuartpreston

tas50: Also we should figure out if we need to add 2019 here

lib/chef/knife/bootstrap.rb
```diff
@@ -22,229 +22,31 @@
...
 
       attr_accessor :client_builder
       attr_accessor :chef_vault_handler
+      attr_reader   :target_host
 
       deps do
-        require "chef/knife/core/bootstrap_context"
         require "chef/json_compat"
         require "tempfile"
-        require "highline"
```
tas50: Are we actually killing off highline entirely with this change. I know we only use it in a handful of places. If so we should nuke that dep in the gemspec.

Gemfile
marcparadise: The circular dep is a mistake - will get that straightened out.  I am +1 on merging these into a single repo with multiple gemspecs.

Gemfile
marcparadise: We'll need to provide a train-winrm gem or similar - train-core does not have any transports included.

lib/chef/knife/bootstrap.rb
```diff
@@ -22,229 +22,31 @@
...
 require "chef/util/path_helper"
+require "chef/knife/bootstrap/options"
 
 class Chef
   class Knife
     class Bootstrap < Knife
       include DataBagSecretOptions
+      # Command line flags and options for bootstrap - there's a large number of them
+      # so we'll keep this file a little smaller by splitting them out.
+      include Bootstrap::Options
```
lamont-granquist: i'm actually not too wild about this change since the bootstrap options are tightly coupled to the code and this leads to flipping between the two files.  if it isn't actually being reused somewhere it just creates a context switch between the two different files, and there's no sharp separation of concerns.  keeping them in the same file makes it easy to search for a string to determine that an option might (e.g.) be dead code and unused anywhere.

lib/chef/knife/bootstrap.rb
marcparadise: We actually weren't using highline in bootstrap

lib/chef/knife/bootstrap.rb
```diff
@@ -351,58 +156,156 @@ def render_template
...
+        perform_bootstrap(bootstrap_path)
+      ensure
+        target_host.del_file(bootstrap_path) if target_host && bootstrap_path
+      end
 
+      def register_client
         # chef-vault integration must use the new client-side hawtness, otherwise to use the
         # new client-side hawtness, just delete your validation key.
+        # 2019-04-01 TODO
+        # TODO -  should this raise if config says to use vault because json/file/item exists
```
marcparadise: more general comment, not specific to these changes.

lib/chef.rb
```diff
@@ -17,6 +17,9 @@
 #
 
 require "chef/version"
+
+require "chef_core/text"
```
tas50: Do we need this require at this top level when we require it in knife bootstrap as well?

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -0,0 +1,412 @@
...
+      #
+      class WindowsBootstrapContext < BootstrapContext
+
+        def initialize(config, run_list, chef_config, secret = nil)
+          @config       = config
+          @run_list     = run_list
+          @chef_config  = chef_config
+          @secret       = secret
+          # Compatibility with Chef 12 and Chef 11 versions
+          begin
```
tas50: We can nuke this as long as we note it in the release notes.

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -0,0 +1,412 @@
...
+        # and prefixes each line with an
+        # echo
+        def escape_and_echo(file_contents)
+          file_contents.gsub(/^(.*)$/, 'echo.\1').gsub(/([(<|>)^])/, '^\1')
+        end
+
+        private
+
+        def install_command(executor_quote)
+          if @config[:install_as_service]
```
tas50: this bit can go away if we nuke the service install option

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -0,0 +1,412 @@
...
+            chef_version_string = if knife_config[:bootstrap_version]
+                                    knife_config[:bootstrap_version]
+                                  else
+                                    Chef::VERSION.split(".").first
+                                  end
+
+            installer_version_string = "&v=#{chef_version_string}"
+
+            # If bootstrapping a pre-release version add the prerelease query string
+            if chef_version_string.split(".").length > 3
```
tas50: if we get rid of pre-release gem support this can go away

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -0,0 +1,412 @@
...
+
+          unless trusted_certs_script.empty?
+            client_rb << %Q{trusted_certs_dir "c:/chef/trusted_certs"\n}
+          end
+
+          if Chef::Config[:fips]
+            client_rb << <<~CONFIG
+              fips true
+              chef_version = ::Chef::VERSION.split(".")
+              unless chef_version[0].to_i > 12 || (chef_version[0].to_i == 12 && chef_version[1].to_i >= 8)
```
tas50: We can assume 12.8+ so this fips support check can go away

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -0,0 +1,412 @@
...
+
+        def local_download_path
+          "%TEMP%\\chef-client-latest.msi"
+        end
+
+        def msi_url(machine_os = nil, machine_arch = nil, download_context = nil)
+          # The default msi path has a number of url query parameters - we attempt to substitute
+          # such parameters in as long as they are provided by the template.
+
+          if @config[:install].nil? || @config[:msi_url].empty?
```
tas50: Should we just use mixlib-install here vs. crafting our own API call?

lib/chef/knife/bootstrap/options.rb
marcparadise: Removed after discussion in community slack

lib/chef/knife/bootstrap/options.rb
marcparadise: Removed after discussion in community slack

lib/chef.rb
marcparadise: We cdan remove it from knife bootstrap. 

This is a complicated one.  Text initializes r18n, which in turn uses the YAML lib.
However, if we hit a path that has us including rubygems first, it monkey-patches YAML 
constants (Psych) for backwards compatibility in a way that makes r18n raise an error. 

This makes it so that the  only way to actually load chef/text is  to ensure we do so before rubygems is used to parse a gemspec.

marcparadise:
Current state: 

* [x] remove windows bootstrap context support for v < 12
* [x] move options back into main class
* [x] remove --prerelease
* [x] remove install support for v < 12.20
* [x] alias flags with changed names, deprecations
* [ ] remove config merge, and use only the merged config: _this requires changin the config merge behavior to include knife keys that are not part of the CLI.  If we're   :+1:  on that, I can make that change, though it may be best as a separate PR. Otherwise what bootstrap does now is the right thing wrt config keys that can have different names in knife config, or only be present in Chef::Config. 
* [ ] remove Chef::Config setter procs from options: I'd like to make this a separate PR and not hold this up on it. This is a bigger change because of how scattered the bootstrap contexts are in sourcing their config. Note that if we go ahead with the change described for config merge behavior above, this becomes easier.

lib/chef/knife.rb
```diff
@@ -343,8 +343,15 @@ def cli_keys
...
     def config_file_settings
+      @key_sources = { cli: [], config: [] }
       cli_keys.each_with_object({}) do |key, memo|
-        memo[key] = Chef::Config[:knife][key] if Chef::Config[:knife].key?(key)
+        if config.key?(key)
+          @key_sources[:cli] << key
+        end
+        if Chef::Config[:knife].key?(key)
+          @key_sources[:config] << key
+          memo[key] = Chef::Config[:knife][key] if Chef::Config[:knife].key?(key)
```
Vasu1105: Do we need if Chef::Config[:knife].key?(key) check here again as we are already checking that at line 351 ?

btm:
Am I correct that this effectively makes these command line changes?

`knife windows bootstrap ssh` -> `knife bootstrap`
`knife windows bootstrap winrm` -> `knife bootstrap -o winrm`

marcparadise:
@btm Yes.  Alternatively it supports `knife bootstrap ssh://HOST` or  `knife bootstrap winrm:://HOST`.

Gemfile
marcparadise: I just saw that I didn't address these in this discussion.  To that end: 

* gem name changed to `chef-core`, with `chef-core-actions` as optional. chef-core-cliux is now just part of chef-core and doesn't have a separate gem. The repos for -cliux and -actions have been removed. 
* it looks like that chef dep was in chef-core-actions, which was in the Gemfile - but not the gemspec.  It shouldn't have been there -- we don't use it in bootstrap and it does have the dep chef-core-actions -> chefdk -> chef. I've confirmed that knife + chef-core does not do this. 
* We'll be  updating  the chef-core repo with a README in the coming days.

Gemfile
marcparadise: https://github.com/inspec/train/pull/433 will resolve this - once it lands and once after inspec-core 4 release, we can switch to train-core.

lib/chef/knife/bootstrap/templates/windows-chef-client-msi.erb
marcparadise: It looks like 2019 is also identified as `10.0`. I can't find any reference to 2016R2 or any versoin 10.1 of Server.

lib/chef/knife/core/windows_bootstrap_context.rb
marcparadise: Given the working state (this file pulled mostly as-is from knife-windows), I'll track that as a follow up for a separate PR.

Gemfile
marcparadise: Looking a little closer, even chef-core-actions should not have that dep - it lives only in chef-apply. chef-core-actions and chef-apply have been updated to reflect this.  `chef-core` now has no lower-level chef deps except `chef-config`. (It only uses that to get a one-line function, so that can go too if it becomes a problem.)

Gemfile.lock
```diff
@@ -1,17 +1,31 @@
+GIT
+  remote: https://github.com/chef/chef-core
```
tas50: Let's ship the gem and make it a dep in the gemspec here

Gemfile
lamont-granquist: yeah so chef can can depend on chef-core and chef-config with chef-core also depending on chef-config since that directed graph does not have a circular dep.

if you potentially start looking at chef-config needing a dep on chef-core then you'd need to remove the dep going the other direction and do something else with that function.

lib/chef/knife.rb
```diff
@@ -355,9 +362,16 @@ def config_file_settings
...
       # update-in-place.
-      config.replace(
-        default_config.merge(config_file_settings).merge(config)
-      )
+      config.replace(default_config.merge(config_file_settings).merge(config))
+    end
+
+    # Return where a config key has been sourced,
+    # :cli, :config, or nil if the key is not set.
+
```
tas50: prob wanna nuke this so yard renders correctly

Gemfile.lock
```diff
@@ -51,42 +65,6 @@ PATH
       proxifier (~> 1.0)
       syslog-logger (~> 1.6)
       uuidtools (~> 2.1.5)
-    chef (15.0.211-universal-mingw32)
```
lamont-granquist: you've lost the windows Gemfile.lock stuff here.  this is where `rake dependences:update_gemfile_lock` can get used to re-create them.

lib/chef.rb
lamont-granquist: probably shouldn't remove it from knife bootstrap.  generally we have to assume that files can get required by client code without this top level require having been hit, so we'd need it both places.

lamont-granquist: (there should probably be a comment here about why this line is here)

tas50:
@marcparadise can you rebase this too to get rid of that gemfile.lock conflict

Gemfile.lock
marcparadise: Done, and also updated to the released chef-core gem.

Gemfile
marcparadise: Based on review discussions, we'll be merging this with train 2.0, but switching over to train-core as soon as it becomes available (eg inspec 4 is available)

RELEASE_NOTES.md
```diff
@@ -24,6 +24,77 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
 
 ## Breaking Changes
 
+### Knife Bootstrap
+
+Knife bootstrap has been updated, and Windows bootstrap has been merged in Knife core bootstrap. This marks the deprecation of knife-windows plugin's `bootstrap` behavior.
+This addresses [CVE-2015-8559](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8559): The knife bootstrap command in chef leaks the validator.pem private RSA key to /var/log/messages.
+
+In order to accomodate a combined bootstrap that supports both SSH and WinRM,
+CLI flags have been added/removed/changed.  Using the changed options will
```
```suggestion
CLI flags have been added, removed, or changed.  Using the changed options will
```

RELEASE_NOTES.md
```diff
@@ -24,6 +24,77 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
 
 ## Breaking Changes
 
+### Knife Bootstrap
+
+Knife bootstrap has been updated, and Windows bootstrap has been merged in Knife core bootstrap. This marks the deprecation of knife-windows plugin's `bootstrap` behavior.
```
mjingle: Does "Knife core bootstrap" require some sort of denotation (`Knife core bootstrap`) or hyphen (Knife-core bootstrap)? Asking since I'm not sure what previous formatting has been used for this.

RELEASE_NOTES.md
```diff
@@ -24,6 +24,77 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+This addresses [CVE-2015-8559](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8559): The knife bootstrap command in chef leaks the validator.pem private RSA key to /var/log/messages.
```
```suggestion
This change also addresses [CVE-2015-8559](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8559): The knife bootstrap command in chef leaks the validator.pem private RSA key to /var/log/messages.
```

RELEASE_NOTES.md
```diff
@@ -24,6 +24,77 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+
+In order to accomodate a combined bootstrap that supports both SSH and WinRM,
+CLI flags have been added/removed/changed.  Using the changed options will
+result in deprecation warnings, but will accept those options unless otherwise noted.
+Using removed options will cause the command to fail.
+
+#### New Flags
+
+| Flag | Description |
+|-----:|:-=----------|
```
mjingle: This "New Flags" chart is not displaying correctly. Is it because of this line's formatting?

RELEASE_NOTES.md
```diff
@@ -24,6 +24,77 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+1. These flags do not have an automatic mapping of old flag -> new flag. The
+   new flag must be used.
+
+#### Removed Flags
+
+| Flag | Notes |
+|-----:|:------|
+|--kerberos-keytab-file| This option existed but was not implemented.|
+|--winrm-codepage| This was used under knife-windows because bootstrapping
+                   was performed over a `cmd` shell. It is now invoked
```
mjingle: This line breaks the table for some reason. I suspect a hard enter somewhere here?

RELEASE_NOTES.md
marcparadise: It looks like the words I should use here are "into core Knife" based on past release notes which talk about merging cookbook resources into "core Chef".

marcparadise: Nope, that's not right - Chef refers to the product name; knife is just a component...

marcparadise: That leaves me with "into core Chef's `knife bootstrap`"

lib/chef/knife/bootstrap/templates/windows-chef-client-msi.erb
tas50: Let me check on the versions in 2019, but we should fix this comment no matter what

tas50: 2016 is "10.0.14393" and 2019 is "10.0.17763" according to fauxhai data I have. That comment should just get updated to be 2019. Perhaps there was a 2016 R2 planned at some point? Who knows.

lib/chef/knife/core/windows_bootstrap_context.rb
tas50: Can you cut an design proposal issue in chef/chef for making that cutover?

lib/chef/knife/bootstrap/templates/windows-chef-client-msi.erb
```diff
@@ -70,10 +70,6 @@ goto architecture_select
 @set MACHINE_OS=2008
```
tas50: We don't produce 2008 MSIs either so we may be able to remove this as well. We'd have to see when we ended that support though.

RELEASE_NOTES.md
```diff
@@ -24,6 +24,81 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+
+```
+  knife bootstrap example.com -o ssh
+  knife bootstrap ssh://example.com
+  knife bootstrap example.com -o winrm
+  knife bootstrap winrm://example.com
+```
+
+
+#### Windows Server 2008
```
tas50: I'm not sure we need to mention this since we don't support  2008 R1.

RELEASE_NOTES.md
```diff
@@ -24,6 +24,81 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+   new flag must be used.
+
+#### Removed Flags
+
+| Flag | Notes |
+|-----:|:------|
+|--kerberos-keytab-file| This option existed but was not implemented.|
+|--winrm-codepage| This was used under knife-windows because bootstrapping was performed over a `cmd` shell. It is now invoked from `powershell`, so this option is no longer used.|
+|--winrm-shell|This option was ignored for bootstrap.|
+|--prerelease|Prerelease Chef hasn't existed for some time.|
```
tas50:
```suggestion
|--prerelease|Chef now releases all development builds to our current channel and does not perform pre-release gem releases.|
```

RELEASE_NOTES.md
```diff
@@ -24,6 +24,81 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
 
 ## Breaking Changes
 
+### Knife Bootstrap
```
tas50: This is more of a todo comment, but before releasing this we should work on rewording this release note item to focus on the positive changes. This is a huge win for users that also comes with some breaking changes. We should be super proud of all the cool parts of that work, while also making sure to warn people that there's some breaking bits.

chef.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.5.0"
 
   s.add_dependency "chef-config", "= #{Chef::VERSION}"
+  s.add_dependency "chef-core", ">= 0.0.1"
```
tas50: Should we constrain this at all? Less than 1.0 or 2.0?

docs/dev/design_documents/bootstrap_with_train.md
```diff
@@ -0,0 +1,150 @@
...
+remains largely unchanged.
+
+## Implementation
+
+### Remove Unsupported Behaviors
+
+We will also remove the following obsolete or unsupported behaviors:
+
+* `--prelease` flag - Chef hasn't been pre-released in quite some time.
+* `--install-as-service` - this is an anti-pattern on Windows that we have
```
tas50:
```suggestion
* `--install-as-service` - For many years we have suggested users not run chef-client as a service due to memory leaks in long running Ruby processes.
```

docs/dev/design_documents/bootstrap_with_train.md
```diff
@@ -0,0 +1,150 @@
...
+* `--kerberos-keytab-file` - this is not implemented in the WinRM gem we use, and so was
+passed through to no effect.
+* remove explicit support for versions of Chef older than 12.8. Versions older than the supported
+  Chef client distributions will continue to be use at your own risk.
+* Remove support for Windows 2003 in the Windows bootstrap template.
```
tas50:
```suggestion
* Remove support for Windows 2003 in the Windows bootstrap template as Chef does not support EOL Windows 2003 installs.
```

RELEASE_NOTES.md
```diff
@@ -24,6 +24,74 @@ Chef now ships with Ruby 2.6.2. This new version of Ruby improves performance an
...
+
+Knife bootstrap has been updated, and Windows bootstrap has been merged into core Chef's `knife bootstrap`. This marks the deprecation of the `knife-windows` plugin's `bootstrap` behavior.
+This change also addresses [CVE-2015-8559](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8559): The knife bootstrap command in chef leaks the validator.pem private RSA key to /var/log/messages.
```
```suggestion
This change also addresses [CVE-2015-8559](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8559): The `knife bootstrap` command in chef leaks the validator.pem private RSA key to /var/log/messages.
```

docs/dev/design_documents/bootstrap_with_train.md
```diff
@@ -0,0 +1,150 @@
...
+* remove explicit support for versions of Chef older than 12.8. Versions older than the supported
+  Chef client distributions will continue to be use at your own risk.
+* Remove support for Windows 2003 in the Windows bootstrap template as Chef does not support EOL Windows 2003 installs.
+
+### CLI Flag Changes
+
+As part of this change, CLI options from `knife bootstrap windows winrm` and `knife bootstrap`
+need to be merged.  The majority will be untouched, but we'll also take this opportunity
+to make flag names more accurately describe what they're doing, and updating several options that are
+protocol-specific to be prefixed with the protocol (e.g. --ssl-peer-fingerprint to --winrm-ssl-peer-fingerprint`)
```
mjingle: Formatting fix!
```suggestion
protocol-specific to be prefixed with the protocol (e.g. `--ssl-peer-fingerprint to --winrm-ssl-peer-fingerprint`)
```

docs/dev/design_documents/bootstrap_with_train.md
```diff
@@ -0,0 +1,150 @@
...
+|-----:|:------------|
+| --max-wait SECONDS | Maximum time to wait for initial connection to be established. |
+| --winrm-basic-auth-only | Perform only Basic Authentication to the target WinRM node. |
+| --connection-protocol PROTOCOL|Connection protocol to use. Valid values are 'winrm' and 'ssh'. Default is 'ssh'. |
+| --connection-user | user to authenticate as, regardless of protocol |
+| --connection-password| Password to authenticate as, regardless of protocol |
+| --connection-port | port to connect to, regardless of protocol |
+
+`--connection-user`, `--connection-port`, and `--connection-password` replace their protocol-specific counterparts, since
+these are applicable to all supported transports.  Their original knife config keys (`ssh\_user`, `ssh\_password`, etc) remain
```
mjingle:
```suggestion
these are applicable to all supported transports.  Their original knife config keys (`ssh\_user`, `ssh\_password`, etc.) remain
```

docs/dev/design_documents/bootstrap_with_train.md
```diff
@@ -0,0 +1,150 @@
...
+
+1. These flags do not have an automatic mapping of old flag -> new flag. The
+   new flag must be used.
+
+### Removed Flags
+
+| Flag | Notes |
+|-----:|:------|
+|--kerberos-keytab-file| This option existed but was not implemented.|
+|--winrm-codepage| This was used under knife-windows because bootstrapping was performed over a `cmd` shell. It is now invoked from `powershell`, so this option is no longer required.|
```
mjingle:
```suggestion
|--winrm-codepage| This was used under `knife-windows` because bootstrapping was performed over a `cmd` shell. It is now invoked from `powershell`, so this option is no longer required.|
```

marcparadise:
The current integration failures are due to this: 
```
Bundler could not find compatible versions for gem "net-ssh":
 In Gemfile:
   chef was resolved to 15.0.223, which depends on
     net-ssh (~> 4.2)
   test-kitchen was resolved to 2.1.0, which depends on
     net-ssh (>= 2.9, < 6.0)
Bundler could not find compatible versions for gem "train":
 In Gemfile:
   chef was resolved to 15.0.223, which depends on
     chef-core (~> 0.0) was resolved to 0.0.1, which depends on
       train (~> 2)
   kitchen-inspec was resolved to 1.1.0, which depends on
     inspec (>= 1.47, < 5.0) was resolved to 1.47.0, which depends on
       train (~> 0.30)
```

Once we upgrade to inspec 4 and relax the net-ssh "~4.2" constraint (#8380), these should be passing.  We'll likely need to update the inspec 4 constraint in `kitchen-inspec` as well.

chef.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.5.0"
 
   s.add_dependency "chef-config", "= #{Chef::VERSION}"
+  s.add_dependency "chef-core", "~> 0.0"
```
tas50: We should leave a note here that we need to add the train-core 2.0 dep as soon as its released

kitchen-tests/Gemfile
```diff
@@ -7,4 +7,5 @@ gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "mas
 gem "kitchen-dokken", "~> 2.0"
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
 gem "inspec", git: "https://github.com/inspec/inspec.git", branch: "master" # this goes away when we ship inspec 4
+gem "kitchen-vagrant", git: "https://github.com/test-kitchen/kitchen-vagrant.git", branch: "master"
```
tas50: This needs to go away again

btm:
@tas50, @lamont-granquist: is our decision to merge this, and then in a later 15.x work on a mixlib-bootstrap or mixlib-transport shaped refactor of chef-core for both chef and chef-apply (the project)?

tas50:
@btm I would like to have a larger sit down to discuss where we see shared code in the projects and where it can be extracted to. There's obviously not time for that now, but something we need to prioritize post Chef 15.

btm:
@tas50 sure. so we'll get this merged today. @jonsmorrow scheduled some time to connect later, but around ChefConf we should get the foundations teams together to talk about shared code as that has implications for everyone.

PR #8278, opened 2019-03-06
windows_service: Fix action :start to not resets credentials on service

### LINKED ISSUE #8080: windows_service action :start resets credentials on service
## Description
When we run the following in a recipe it resets the credentials on the service to Local System

```
windows_service 'myServiceName' do
	action :start
	only_if { ::Win32::Service.exists?("myServiceName")}
end
```

## Chef Version
14.7.17
## Platform Version
Windows 2016

## Replication Case
Whenever we use windows_service with action :start it replaces the credentials the service runs as with Local System.

jasonwbarnett:
### Description

Use idempotent `#action_configure` instead of using older code path that isn't actually idempotent.

### Issues Resolved

- #8080

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/service/windows.rb
+++ b/lib/chef/provider/service/windows.rb
@@ -83,22 +83,7 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
 
   def start_service
     if Win32::Service.exists?(@new_resource.service_name)
-      # reconfiguration is idempotent, so just do it.
-      new_config = {
-        service_name: @new_resource.service_name,
-        service_start_name: @new_resource.run_as_user,
-        password: @new_resource.run_as_password,
-      }.reject { |k, v| v.nil? || v.length == 0 }
-
-      Win32::Service.configure(new_config)
-      logger.info "#{@new_resource} configured."
-
-      # LocalSystem is the default runas user, which is a special service account that should ultimately have the rights of BUILTIN\Administrators, but we wouldn't see that from get_account_right
-      if new_config.key?(:service_start_name) && new_config[:service_start_name].casecmp("localsystem") != 0
-        unless Chef::ReservedNames::Win32::Security.get_account_right(canonicalize_username(new_config[:service_start_name])).include?(SERVICE_RIGHT)
-          grant_service_logon(new_config[:service_start_name])
-        end
-      end
+      configure_service_run_as_properties
 
       state = current_state
       if state == RUNNING
@@ -281,6 +266,21 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
 
   private
 
+  def configure_service_run_as_properties
+    return unless new_resource.property_is_set?(:run_as_user)
+
+    new_config = {
+      service_name: new_resource.service_name,
+      service_start_name: new_resource.run_as_user,
+      password: new_resource.run_as_password,
+    }.reject { |k, v| v.nil? || v.length == 0 }
+
+    Win32::Service.configure(new_config)
+    logger.info "#{new_resource} configured."
+
+    grant_service_logon(new_resource.run_as_user) if new_resource.run_as_user.casecmp("localsystem") != 0
+  end
+
   def current_delayed_start
     if service = Win32::Service.services.find { |x| x.service_name == new_resource.service_name }
       service.delayed_start == 0 ? false : true
@@ -290,6 +290,8 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
   end
 
   def grant_service_logon(username)
+    return if Chef::ReservedNames::Win32::Security.get_account_right(canonicalize_username(username)).include?(SERVICE_RIGHT)
+
     begin
       Chef::ReservedNames::Win32::Security.add_account_right(canonicalize_username(username), SERVICE_RIGHT)
     rescue Chef::Exceptions::Win32APIError => err
--- a/spec/unit/provider/service/windows_spec.rb
+++ b/spec/unit/provider/service/windows_spec.rb
@@ -511,6 +511,31 @@ describe Chef::Provider::Service::Windows, "load_current_resource", :windows_onl
         double("StatusStruct", current_state: "running"))
     end
 
+    context "run_as_user user is specified" do
+      let(:run_as_user) { provider.new_resource.class.properties[:run_as_user].default }
+
+      before do
+        provider.new_resource.run_as_user run_as_user
+      end
+
+      it "configures service run_as_user and run_as_password" do
+        expect(provider).to receive(:configure_service_run_as_properties).and_call_original
+        expect(Win32::Service).to receive(:configure)
+        provider.start_service
+      end
+    end
+
+    context "run_as_user user is not specified" do
+      before do
+        expect(provider.new_resource.property_is_set?(:run_as_user)).to be false
+      end
+
+      it "does not configure service run_as_user and run_as_password" do
+        expect(Win32::Service).not_to receive(:configure)
+        provider.start_service
+      end
+    end
+
     context "start_command is specified" do
       let(:start_command) { "sc start #{chef_service_name}" }
 


lib/chef/provider/service/windows.rb
```diff
@@ -83,20 +83,12 @@ def load_current_resource
...
-      # reconfiguration is idempotent, so just do it.
-      new_config = {
-        service_name: @new_resource.service_name,
-        service_start_name: @new_resource.run_as_user,
-        password: @new_resource.run_as_password,
-      }.reject { |k, v| v.nil? || v.length == 0 }
-
-      Win32::Service.configure(new_config)
-      logger.info "#{@new_resource} configured."
+      action_configure
```
jasonwbarnett: To be clear, I don't think it makes sense to call configure here, but I left it so it doesn't break the existing API.

jasonwbarnett:
@tas50 can you review the failing test suites and tell me if they're related to the PR? I reviewed them and everything seems unrelated with the exception of the following, but I am not close enough to be sure. [This failure][1] seems super odd and I'm not sure why it would be happening.

Any ideas?

```text
  1) Chef::Provider::Service::Windows load_current_resource Chef::Provider::Service::Windows start_service calls #action_configure
     Failure/Error: Win32::Service.start(@new_resource.service_name)
     NoMethodError:
       undefined method `start' for Win32::Service:Class
     # ./lib/chef/provider/service/windows.rb:108:in `block in start_service'
     # ./lib/chef/provider/service/windows.rb:331:in `block in spawn_command_thread'
```

[1]: https://ci.appveyor.com/project/chef/chef/builds/22855264/job/04sss4u6f7lp7ly6#L1021

jasonwbarnett:
> @tas50 can you review the failing test suites and tell me if they're related to the PR? I reviewed them and everything seems unrelated with the exception of the following, but I am not close enough to be sure. [This failure](https://ci.appveyor.com/project/chef/chef/builds/22855264/job/04sss4u6f7lp7ly6#L1021) seems super odd and I'm not sure why it would be happening.
> 
> Any ideas?
> 
> ```
>   1) Chef::Provider::Service::Windows load_current_resource Chef::Provider::Service::Windows start_service calls #action_configure
>      Failure/Error: Win32::Service.start(@new_resource.service_name)
>      NoMethodError:
>        undefined method `start' for Win32::Service:Class
>      # ./lib/chef/provider/service/windows.rb:108:in `block in start_service'
>      # ./lib/chef/provider/service/windows.rb:331:in `block in spawn_command_thread'
> ```

After digging a little deeper I noticed that all of the other unit tests for `#start_service` have stubbed the call to `::start`. Does this seem right to you? Obviously I could just add a stubbed response so it doesn't call out, but I want to understand why this is happening in the first place, especially given this is running on appveyor (windows), not linux.

jasonwbarnett:
My suggestion is to merge this and in a separate pull request I am going to refactor the `windows_service` provider unit tests so they are higher quality.

jasonwbarnett:
Don't merge this yet. I think this will actually put the provider in a worse spot. I'd like to remove the configure portion of the `:start` action.

@stuartpreston what are your thoughts? Any downsides to doing this?

lib/chef/provider/service/windows.rb
```diff
@@ -281,6 +266,23 @@ def action_configure_startup
 
   private
 
+  def configure_service_run_as_properties
+    return unless new_resource.property_is_set?(:run_as_user)
```
jasonwbarnett: This is the actual fix to #8080

jasonwbarnett:
@tas50 So I looked back at be28ba90b and based on that wanted to keep the whole block of code added to the same method because I figured they were part of the same "fix" or whatever it was committed for.

So this is a simple refactor that fixes #8080 and makes the code slightly more readable in the short-term.

lib/chef/provider/service/windows.rb
```diff
@@ -281,6 +266,21 @@ def action_configure_startup
...
+
+    new_config = {
+      service_name: new_resource.service_name,
+      service_start_name: new_resource.run_as_user,
+      password: new_resource.run_as_password,
+    }.reject { |k, v| v.nil? || v.length == 0 }
+
+    Win32::Service.configure(new_config)
+    logger.info "#{new_resource} configured."
```
tas50: We probably want to make this log the service_name property and not the resource name

lib/chef/provider/service/windows.rb
lamont-granquist: is `return unless new_resource.run_as_user == "LocalSystem"` the same?  its just that `property_is_set?` checking is a bit odd, and that creates the need to go look up what the default was.  also, if someone sets the property to "LocalSystem" should it behave the same way here (in which case it may actually be a bug you've written here?).

if the user specifying "LocalSystem" is different than the default of "LocalSystem" that seems, at any rate, odd and worthy of some commented documentation if that's the way things are...

lib/chef/provider/service/windows.rb
jasonwbarnett: I agree. I'll update now.

lib/chef/provider/service/windows.rb
jasonwbarnett: I'm asserting that the behavior _should_ be different whether it's the default value versus being explicitly set to that same default value. The logic behind this is that a chef user may desire to use the `windows_service` just for notifications, `:start`/`:stop`/`:restart`, etc.

I believe the current behavior is unexpected. I don't believe that someone with the following resource would actually want the service to be re-configured. This is especially true for built-in/native windows services that are drivers which don't even run as a specific user (read reference below). With that said, this is definitely in need of some clear documentation because it's going to be confusing regardless.

```ruby
windows_service 'Dhcp' do
  action :start
end
```

> If the service type is SERVICE_KERNEL_DRIVER or SERVICE_FILE_SYSTEM_DRIVER, the name is the driver object name that the system uses to load the device driver. Specify NULL if the driver is to use a default object name created by the I/O system.

Reference: https://docs.microsoft.com/en-us/windows/desktop/api/winsvc/nf-winsvc-changeserviceconfiga

spec/unit/provider/service/windows_spec.rb
```diff
@@ -511,6 +511,31 @@
         double("StatusStruct", current_state: "running"))
     end
 
+    context "run_as_user user is specified" do
+      let(:run_as_user) { provider.new_resource.class.properties[:run_as_user].default }
```
jasonwbarnett: @lamont-granquist Tied back to your original question...

> if the user specifying "LocalSystem" is different than the default of "LocalSystem" that seems, at any rate, odd and worthy of some commented documentation if that's the way things are...

I intentionally setup the test like this so that we would explicitly differentiate between the default being explicitly set and the default itself.

btm:
@jasonwbarnett what if we dropped `Win32::Service.configure(new_config)` completely from `start_service` ? It is only setting the run_as user, not the whole service configuration. Would this simplify everything?

Presuming anyone is relying on this:
```
windows_service 'foo' do
  run_as_user 'not_local_system'
  run_as_password 'secret'
  action :start
end
```

Would that then be replaced by this?
```
windows_service 'foo' do
  run_as_user 'not_local_system'
  run_as_password 'secret'
  action [ :configure, :start ]
end
```

jasonwbarnett:
> @jasonwbarnett what if we dropped `Win32::Service.configure(new_config)` completely from `start_service` ? It is only setting the run_as user, not the whole service configuration. Would this simplify everything?
> 
> Presuming anyone is relying on this:
> 
> ```
> windows_service 'foo' do
>   run_as_user 'not_local_system'
>   run_as_password 'secret'
>   action :start
> end
> ```
> Would that then be replaced by this?
> 
> ```
> windows_service 'foo' do
>   run_as_user 'not_local_system'
>   run_as_password 'secret'
>   action [ :configure, :start ]
> end
> ```

That's right, but it's a little nuanced. The way the service is being configured in the `#start_action` _only_ configures the `run_as_user` and `run_as_password` properties. The `:configure` action configures _everything_, so I  will need to look closer at what those differences are and if that will have any unintended consequences.

In general though, I think that's right.

PR #8298, opened 2019-03-14
Allow the use of `--delete-entire-chef-repo`

ABewsher:
Signed-off-by: Adam Bewsher <adam.bewsher@neos.co.uk>

### Description

Allows the use of `--delete-entire-chef-repo` in chef-client when also using `--local-mode` with `--recipe-url`

### Issues Resolved

Previously when using these options, chef-client exited with `undefined local variable or method  'recipes_path'`.  This was due to a copy/paste error in https://github.com/chef/chef/pull/4545 and no tests.

### DIFF
--- a/spec/unit/application/client_spec.rb
+++ b/spec/unit/application/client_spec.rb
@@ -91,7 +91,7 @@ describe Chef::Application::Client, "reconfigure" do
 
     # protect the unit tests against accidental --delete-entire-chef-repo from firing
     # for real during tests.  DO NOT delete this line.
-    expect(FileUtils).not_to receive(:rm_rf)
+    allow(FileUtils).to receive(:rm_rf)
   end
 
   after do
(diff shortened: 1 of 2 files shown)


tas50:
Nice catch @ABewsher. Thanks for adding all the tests so we make sure this doesn't break again.

spec/unit/application/client_spec.rb
```diff
@@ -91,7 +91,7 @@
 
     # protect the unit tests against accidental --delete-entire-chef-repo from firing
     # for real during tests.  DO NOT delete this line.
-    expect(FileUtils).not_to receive(:rm_rf)
+    allow(FileUtils).to receive(:rm_rf)
```
lamont-granquist: took me a second, but yeah you don't need the expectation here, the allow will stub it for everything which protects the other specs still, then you can override with the expectation, and it doesn't work to override an expectation with an expectation...  (mostly mentioning this for my future self if it ever comes up again..)

lamont-granquist:
(and yes, thanks for yak shaving the specs, our coverage in knife is poor)

ABewsher:
No probs.  Glad to help.

PR #8301, opened 2019-03-15
Early allocation of the Chef::RunContext

lamont-granquist:
Allow for allocation of the Chef::RunContext without its arguments and
use that to allocate it early.  The APIs are kept backcompatible since
otherwise this has a decent chance of blowing up something like
chefspec.

### DIFF
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -263,7 +263,7 @@ class Chef
         Chef.resource_handler_map.lock!
         Chef.provider_handler_map.lock!
 
-        setup_run_context(run_context)
+        setup_run_context
 
         load_required_recipe(@rest, run_context) unless Chef::Config[:solo_legacy_mode]
 
@@ -465,9 +465,11 @@ class Chef
     # @return The newly set up run context
     #
     # @api private
-    def setup_run_context(run_context)
+    def setup_run_context
       policy_builder.setup_run_context(specific_recipes, run_context)
       assert_cookbook_path_not_empty(run_context)
+      run_status.run_context = run_context # backcompat for chefspec
+      run_context
     end
 
     #
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -473,7 +473,7 @@ class Chef
     #
     # @api private
     def setup_run_context
-      run_context = policy_builder.setup_run_context(specific_recipes, run_context)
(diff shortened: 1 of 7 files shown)


lib/chef/data_bag.rb
```diff
@@ -32,8 +32,8 @@ class DataBag
     include Chef::Mixin::FromFile
     include Chef::Mixin::ParamsValidate
 
-    VALID_NAME = /^[\.\-[:alnum:]_]+$/.freeze
-    RESERVED_NAMES = /^(node|role|environment|client)$/.freeze
+    VALID_NAME = /^[\.\-[:alnum:]_]+$/
+    RESERVED_NAMES = /^(node|role|environment|client)$/
```
lamont-granquist: unrelated chefstyle that somehow crept in...

jaymzh:
Is the idea of this so that we can log earlier and have earlier event handlers?

lib/chef/client.rb
```diff
@@ -451,9 +473,9 @@ def sync_cookbooks
     #
     # @api private
     def setup_run_context
-      run_context = policy_builder.setup_run_context(specific_recipes)
+      @run_context = policy_builder.setup_run_context(specific_recipes, run_context)
       assert_cookbook_path_not_empty(run_context)
-      run_status.run_context = run_context
+      run_status.run_context = run_context # backcompat for chefspec
```
jaymzh: should this be @run_context since you just assigned to that?

lamont-granquist: no, its an accessor so it works fine and returns `@run_context`

lamont-granquist:
it ultimately gets the run_context created earlier and gets setup_run_context out of the business of creating it, and that allows event handlers to get ahold of that object earlier
and then the point is more to the hang the rest client(s) off of the run_context for reuse.  that has broader benefits, but one goal is to get all the ResourceReporter + DataCollector crap out of the Chef::Client.

lamont-granquist:
also "do less work in initializers" and now you can create a Chef::RunContext without passing a node + cookbook_collection + event handler, which it doesn't always need (noting how often we pass `{}` in for the cookbook_collection), etc.

not that i'm certain how much i want to push that, since there's about a billion run_context objects created in the unit tests and i don't really want to go through and 'fix' every single one just for philosophical reasons -- hence why this change is all backcompat and not breaking...

PR #8307, opened 2019-03-19
package: move response_file and response_file_variables out of base package resource

### LINKED ISSUE #7599: Move response_file and response_file_variables out of base package resource
The response_file and response_file_variables properties only work in apt/dpkg resources, but for historical reasons they were applied to the base package resource. When auto generating documentation this results in these properties showing up in every package resource, which needs to be fixed.

kapilchouhan99:

### Description

- Moved `response_file` and `response_file_variables` properties from `package` resource to `apt_package` and `dpkg_package` resources
- Created a `Chef::Provider::Package::Deb` module for `apt` and `dpkg` providers and included it into Apt and Dpkg classes which are the sub-classes of the Package class, and also moved common methods of `apt` and `dpkg` providers in `Chef::Provider::Package::Deb`
- Added test cases
### Issues Resolved
Fixes #7599 MSYS-986

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -88,14 +88,7 @@ class Chef
           return
         end
 
-        # @todo: move the preseed code out of the base class (and complete the fix for Array of preseeds? ugh...)
-        if new_resource.response_file
-          if preseed_file = get_preseed_file(package_names_for_targets, versions_for_targets)
-            converge_by("preseed package #{package_names_for_targets}") do
-              preseed_package(preseed_file)
-            end
-          end
-        end
+        prepare_for_installation
 
         converge_by(install_description) do
           multipackage_api_adapter(package_names_for_targets, versions_for_targets) do |name, version|
@@ -194,31 +187,6 @@ class Chef
         end
       end
 
-      action :reconfig do
-        if current_resource.version.nil?
-          logger.trace("#{new_resource} is NOT installed - nothing to do")
-          return
-        end
-
-        unless new_resource.response_file
-          logger.trace("#{new_resource} no response_file provided - nothing to do")
-          return
-        end
-
-        if preseed_file = get_preseed_file(new_resource.package_name, current_resource.version)
-          converge_by("reconfigure package #{new_resource.package_name}") do
-            preseed_package(preseed_file)
-            multipackage_api_adapter(new_resource.package_name, current_resource.version) do |name, version|
-              reconfig_package(name, version)
-
-            end
-            logger.info("#{new_resource} reconfigured")
-          end
-        else
-          logger.trace("#{new_resource} preseeding has not changed - nothing to do")
-        end
-      end
-
       def action_lock
         packages_locked = if respond_to?(:packages_all_locked?, true)
                             packages_all_locked?(Array(new_resource.package_name), Array(new_resource.version))
@@ -262,6 +230,10 @@ class Chef
         raise Chef::Exceptions::UnsupportedAction, "#{self} has no way to detect if package is locked"
       end
 
+      # Subclasses will override this to a method and provide a preseed file path
+      def prepare_for_installation
+      end
+
       # @todo use composition rather than inheritance
 
       def multipackage_api_adapter(name, version)
@@ -292,7 +264,7 @@ class Chef
         raise Chef::Exceptions::UnsupportedAction, "#{self} does not support pre-seeding package install/upgrade instructions"
       end
 
-      def reconfig_package(name, version)
+      def reconfig_package(name)
         raise( Chef::Exceptions::UnsupportedAction, "#{self} does not support :reconfig" )
       end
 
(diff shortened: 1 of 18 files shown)


lib/chef/provider/package/debian.rb
```diff
@@ -0,0 +1,135 @@
...
+
+          if new_resource.response_file
+            if preseed_file = get_preseed_file(package_names_for_targets, versions_for_targets)
+              converge_by("preseed package #{package_names_for_targets}") do
+                preseed_package(preseed_file)
+              end
+            end
+          end
+
+          converge_by(install_description) do
```
dheerajd-msys: Can we put this `converge_by` block under some conditional to maintain the idempotent behavior?

lib/chef/provider/package/debian.rb
```diff
@@ -0,0 +1,135 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/provider/package"
+
+class Chef
+  class Provider
+    class Package
+      class Debian < Chef::Provider::Package
```
lamont-granquist: This should actually be written as a module not a class since the Chef::Provider::Package::Debian class should not be used directly.

That means you'll need to use the included hook:

```
class Chef
  class Provider
    class Package
      module Debian
        def self.included(base)
          base.class_eval do
[... all the normal contents go here...]
          end
        end
      end
    end
  end
end

File: lib/chef/provider/package/debian.rb [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+        use_multipackage_api
+
+        action :install do
```
lamont-granquist: this is all a bit problematic because now we've forked the install action in the subclass, so fixes to the superclass would need to get replicated into this code.

what likely needs to get extracted is a hook in the superclass like `def prepare_for_installation` which is called from the middle of the `action :install` in the superclass but which is normally a method with no code, then the subclass can override the implementation of that to do the response_file/preseed_file stuff.   then there is no copypasta between the superclass and the subclass.

lib/chef/provider/package.rb
```diff
@@ -88,14 +88,8 @@ def define_resource_requirements
...
 
-        # @todo: move the preseed code out of the base class (and complete the fix for Array of preseeds? ugh...)
-        if new_resource.response_file
-          if preseed_file = get_preseed_file(package_names_for_targets, versions_for_targets)
-            converge_by("preseed package #{package_names_for_targets}") do
-              preseed_package(preseed_file)
-            end
-          end
-        end
+        # for apt_package and dpkg_package
```
robbkidd: Recommend removing this comment in favor of adding documentation to the `prepare_for_installation` method itself, what role does it perform for a package resource, and how it could be used generically.

lib/chef/provider/package/debian.rb
```diff
@@ -0,0 +1,129 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "chef/provider/package"
+
+class Chef
+  class Provider
+    class Package
+      module Debian
```
robbkidd: Recommend naming this module `Deb` to more closely associate it with the `.deb` package type and less with the `Debian` platform. See prior art in Windows for `Windows::MSI` and `Windows::EXE`.

lib/chef/provider/package/debian.rb
```diff
@@ -0,0 +1,129 @@
...
+
+              unless new_resource.response_file
+                logger.trace("#{new_resource} no response_file provided - nothing to do")
+                return
+              end
+
+              if preseed_file = get_preseed_file(new_resource.package_name, current_resource.version)
+                converge_by("reconfigure package #{new_resource.package_name}") do
+                  preseed_package(preseed_file)
+                  multipackage_api_adapter(new_resource.package_name, current_resource.version) do |name, version|
```
robbkidd:
```suggestion
                  multipackage_api_adapter(new_resource.package_name, current_resource.version) do |name, _version|
```
The revised `reconfig_package()` method now does not need the `version` parameter, so `version` here is not used. It would help future readers (and Rubocop) to know that `version` is _intentionally_ not used if you change `|name, version|` to `|name, _version|`

spec/unit/provider/debian_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    node = Chef::Node.new
+    node.automatic_attrs[:platform] = :just_testing
+    node.automatic_attrs[:platform_version] = :just_testing
+    node
+  end
+  let(:events) { Chef::EventDispatch::Dispatcher.new }
+  let(:logger) { double("Mixlib::Log::Child").as_null_object }
+  let(:run_context) { Chef::RunContext.new(node, {}, events) }
+  let(:new_resource) { Chef::Resource::AptPackage.new("emacs", run_context) }
+  let(:current_resource) { Chef::Resource::AptPackage.new("emacs", run_context) }
```
robbkidd: Can the abstract `MyDebianPackageResource` and `MyDebianPackageProvider` classes that appear later in this spec file be used here instead of depending on `AptPackage`?

spec/unit/provider/debian_spec.rb
```diff
@@ -0,0 +1,136 @@
+#
```
robbkidd: Because the new module file exists in `lib/chef/provider/package/<module>.rb`, I think the spec should exist in `spec//unit/provider/package/<module>.rb`.

spec/unit/provider/package/apt_spec.rb
```diff
@@ -428,6 +428,91 @@ def ubuntu1404downgrade_stubs
       end
     end
 
+    describe "when given a response file" do
```
robbkidd: These specs test the logic in the new module that is shared between the Apt and Dpkg package resources. What do you think about collecting this group of specs into a single set of `shared_examples` in the module's spec file, testing their inclusion with `MyDebianPackage{Provider|Resource`, and then using the shared examples here and in dpkg_spec?

spec/unit/provider/debian_spec.rb
lamont-granquist: yeah this should definitely get moved.

lib/chef/provider/package/debian.rb
lamont-granquist: we don't do that in the chef codebase though, and we went through that rule and intentionally and thoughtfully turned it off though.

spec/unit/provider/debian_spec.rb
kapilchouhan99: @robbkidd this test case is related to action `reconfig` and this action is used by only `apt_package` and `dpkg_package`, so currently we are checking logger like this `apt_package[emacs] is NOT installed - nothing to do` if we use the abstract `MyDebianPackageResource` and `MyDebianPackageProvider` classes then we got logger like this `[emacs] is NOT installed - nothing to do`

lib/chef/provider/package/debian.rb
robbkidd: OK. Mine wasn't a strong suggestion.

kapilchouhan99:
I have verified all the test cases are working fine on locally but its failing on Travis build and AppVeyor build.

spec/unit/provider/debian_spec.rb
kapilchouhan99: @robbkidd In specs we are setting `response_file` property, and this property is only present in `apt` and `dpkg`, so that's why here I am using `AptPackage`

kapilchouhan99:
I have fixed Travis build and AppVeyor build failure, and Please let me know if you have any required changes.

lib/chef/provider/package.rb
```diff
@@ -262,6 +230,11 @@ def package_locked(name, version)
         raise Chef::Exceptions::UnsupportedAction, "#{self} has no way to detect if package is locked"
       end
 
+      # Subclasses will override this to a method and provide a preseed file path
+      # only apt_package and dpkg_package are overriding this method
```
marcparadise: Would recommend removing hte reference to apt_package/dpkg_package - this is the kind of thing that changes over time, and this is the kind of comment that we forget to update because it's not in the path of the change.

lamont-granquist:
hmm chefstyle seems annoyed about something though

PR #8310, opened 2019-03-19
Remove the travis gem from our gemfile

tas50:
It doesn't look like we're using this anymore.


lamont-granquist:
AFAIK this was only ever used to create secrets back when we were using test-kitchen to launch AWS instances via kitchen-ec2

PR #8318, opened 2019-03-21
Add windows_service deprecation of configuring run_as properties

jasonwbarnett:
### Description

Steps towards cleaning up the `windows_service` resource.

### Issues Resolved

N/A

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -233,6 +233,10 @@ class Chef
       target 28
     end
 
+    class WindowsServiceConfigureStartup < Base
+      target 29
+    end
+
     class Generic < Base
       def url
         "https://docs.chef.io/chef_deprecations_client/"
--- a/lib/chef/provider/service/windows.rb
+++ b/lib/chef/provider/service/windows.rb
@@ -264,6 +264,13 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
   def configure_service_run_as_properties
     return unless new_resource.property_is_set?(:run_as_user)
 
+    unless new_resource.action.include?(:configure)
+      new_actions = new_resource.action.dup
+      insert_position = new_actions.find_index(:start) || 0
+      new_actions.insert(insert_position, :configure)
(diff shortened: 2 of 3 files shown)


tas50:
@jasonwbarnett You're going to need to make this deprecation 28 now since we just merged 27. You also have a pile of failures in appveyor and travis.

jasonwbarnett:
> @jasonwbarnett You're going to need to make this deprecation 28 now since we just merged 27. You also have a pile of failures in appveyor and travis.

I totally winged this and have no idea if it's even right. Which PR was for 27 so I can use it for a reference?

jasonwbarnett:
@tas50 ready for 👀

jasonwbarnett:
@tas50 ready for review!

lamont-granquist:
So the problem here is that philosophically Chef should never have had verbs at all other than "action :fixit" and it should only have properties which are nouns (more or less -- ignoring flags that determine how something should get fixed like atmoic_update on the file resource).

So, looking at the english verb "start" and trying to "make that make sense" is going to be digging the hole even deeper here.

Ultimately if you have a resource which declares start and declares a run_as user and the run_as user is ignored, then that's what nobody would expect.  If the run_as user is specified on start then you need to leave the actual configured resource on the system with the correct run_as user and to fix it so that it is running (`action :start` means something like "must be left in a running state with all the correct properties that I've specified").

It sounds like this comes from a situation where the run_as user is incorrectly taking a run_as user of System so that a simple `action :start` with nothing specified will cause the service to reconfigure to the System user in windows.  That's a bug.  The run_as_user property should be nil.  When the run as user is nil then when trying to reconfigure a system the run_as_user should be ignored.  When trying to create a user the run_as_user should still be nil, and that nil should mean System.  It sounds like from discussion that windows sets up services to run_as_user as System so that could be just left up to the O/S which sounds most correct -- whatever the O/S default is we should probably trust that (unless it is literally insane for some reason).

The bug here is that someone decided to be explicit about the System default in the run_as_user which is wrong.  That is the correct *behavior* but specifying it on the property and making that the Chef property default is incorrect.  Properties should default to nil whenever possible to just leave it up to the O/S.  Asserting the O/S default as a property default leads to bugs like this one.

jasonwbarnett:
No reason to leave this open. Thanks for the information @lamont-granquist

PR #8340, opened 2019-04-05
Handle non-root usage of cron resource and basic test. - Reset of PR #5710

Tensibai:
Signed-off-by: Tensibai <tensibai@iabis.net>

### Description

Ensure the crontab command line used doesn't use `-u` if not run as root, fail if non-root and trying to set someone else crontab.

### Issues Resolved

- Issue 2491
- StackOverflow Q/A: [Unable to run Chef cron recipe as non-root user](http://stackoverflow.com/questions/40470685/unable-to-run-chef-cron-recipe-as-non-root-user/40475768#40475768)

### Check List

- [X] All tests pass
- [x] Non-root user editing its own crontab should not raise an error.
- [x] Non-root user editing someone else crontab should raise an error.


Tensibai:
So I'm a bitlost with the functionnal test not working, taker of advice on what I'm missing :)

tas50:
@Tensibai wanna try to rebase this. We've changed a lot in our testing in a last few months

Tensibai:
@tas50 Rebased, style corrected. There's still some tests failing but they don't seem related to this change.

lib/chef/provider/cron.rb
```diff
@@ -41,11 +41,34 @@ def initialize(new_resource, run_context)
...
+          else
+            raise etype, emessage
+          end
+        end
+        return if nuid == Process.euid
+
+        emessage = "Error in #{new_resource.name}, Chef can't modify another user crontab when not running as root."
+        events.provider_requirement_failed(new_resource.action, new_resource.name, etype, emessage)
+        raise etype, emessage
+      end
```
lamont-granquist: this should probably be moved to `defined_resource_requirements` and use the typical assertion syntax there.  if the point is to protect load_current_resource from exploding, then LCR really needs to be made 'safe' for the cases when the user doesn't exist and/or the thing isn't being run as root (should just return a bunch of nil stuff for what it can't get at).

what you're doing here with directly handling your own why run assertions is waaaaaay too complicated to follow for 99% of humans.

Tensibai: Well, it's starting to be quite old :) I think this was made on 13 codebase at first.

I'll see to understand the define_resource_requirement then, any hint to bootstrap me?

Tensibai: BTW, the goal is to fail the run with an informative message if the user doesn't exist as the resource can't do it's job.

lamont-granquist: here's an example of trying to modify a group that doesn't exist, which is the closest i could find offhand:

https://github.com/chef/chef/blob/master/lib/chef/provider/group.rb#L54-L59

johnmccrae:
@Tensibai is this still active? Should we keep working on this?

Tensibai:
Unsure, I didn't had time nor will to work around latest comment from Lamont about define_resource_requirement nor to check if the long time overdue rewrite of this resource did happen.

I'd say the test case is in the pr, if you add it to current master and it fails, that's still something to work on imho.

Happy to discuss it in more details on slack if you want

PR #8341, opened 2019-04-05
initial proposal for distro constants

bobchaos:
### Description

first steps towards allowing multiple distros of Chef in compliance with the Trademark policy
### Issues Resolved

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/application/windows_service.rb
+++ b/lib/chef/application/windows_service.rb
@@ -121,7 +121,7 @@ class Chef
 
       def service_stop
         run_warning_displayed = false
-        Chef::Log.info("STOP requeschefm.")
+        Chef::Log.info("STOP request from operating system.")
         loop do
           # See if a run is in flight
           if @service_action_mutex.try_lock
--- a/lib/chef/application/windows_service.rb
+++ b/lib/chef/application/windows_service.rb
@@ -29,6 +29,7 @@ require "socket"
 require "uri"
 require "win32/daemon"
 require "chef/mixin/shell_out"
+require "chef/dist"
 
 class Chef
   class Application
@@ -66,7 +67,7 @@ class Chef
         @service_signal = ConditionVariable.new
 
         reconfigure
-        Chef::Log.info("Chef Client Service initialized")
+        Chef::Log.info("#{Chef::Dist::CLIENT} Service initialized")
       end
 
       def service_main(*startup_parameters)
@@ -78,7 +79,7 @@ class Chef
           # Grab the service_action_mutex to make a chef-client run
           @service_action_mutex.synchronize do
             begin
-              Chef::Log.info("Next chef-client run will happen in #{timeout} seconds")
+              Chef::Log.info("Next #{Chef::Dist::CLIENT} run will happen in #{timeout} seconds")
               @service_signal.wait(@service_action_mutex, timeout)
 
               # Continue only if service is RUNNING
@@ -95,7 +96,7 @@ class Chef
               # run chef-client only if service is in RUNNING state
               next if state != RUNNING
 
-              Chef::Log.info("Chef-Client service is starting a chef-client run...")
+              Chef::Log.info("#{Chef::Dist::CLIENT} service is starting a #{Chef::Dist::CLIENT} run...")
               run_chef_client
             rescue SystemExit => e
               # Do not raise any of the errors here in order to
@@ -120,7 +121,7 @@ class Chef
 
       def service_stop
         run_warning_displayed = false
-        Chef::Log.info("STOP request from operating system.")
+        Chef::Log.info("STOP requeschefm.")
         loop do
           # See if a run is in flight
           if @service_action_mutex.try_lock
@@ -130,12 +131,12 @@ class Chef
             break
           else
             unless run_warning_displayed
-              Chef::Log.info("Currently a chef run is happening on this system.")
-              Chef::Log.info("Service  will stop when run is completed.")
+              Chef::Log.info("Currently a #{Chef::Dist::GENERIC} run is happening on this system.")
+              Chef::Log.info("Service will stop when run is completed.")
               run_warning_displayed = true
             end
 
-            Chef::Log.trace("Waiting for chef-client run...")
+            Chef::Log.trace("Waiting for #{Chef::Dist::CLIENT} run...")
             sleep 1
           end
         end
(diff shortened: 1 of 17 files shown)


lamont-granquist:
you can go ahead and PR the rest of the file changes you had into here as well.

lib/chef/dist.rb
```diff
@@ -0,0 +1,21 @@
...
+    CLIENT = "chef-client"
+
+    # the server tool's name (knife)
+    KNIFE = "knife"
+
+    # Name used for certain directories. Merge with chef_executable?
+    GENERIC = "chef"
+
+    # The chef executable, as in `chef gem install` or `chef generate cookbook`
+    EXEC = "chef"
```
tas50: It seems like this one should be in DK/workstation where that tool actually exists.

bobchaos: it's referenced in knife.rb. I'm adding them as I go

bobchaos:
> you can go ahead and PR the rest of the file changes you had into here as well.

done, with some minor adjustements

lib/chef.rb
```diff
@@ -32,3 +32,5 @@
 require "chef/handler/json_file"
 require "chef/event_dispatch/dsl"
 require "chef/chef_class"
+
+require "chef/dist"
```
lamont-granquist: so this should actually go in the individual files.  not everything does a `require "chef"` at the top of it and we support stuff like `require "chef/node"` or other randomish things that have never quite been put into any kind of coherent public API.  so the general rule is that if there are class constants in the file, then there should be a require line at the top of the file.

bobchaos: done!

bobchaos:
will this require tests? And if so, any tips on the most efficient way to write those?

lib/chef/application/windows_service.rb
```diff
@@ -120,7 +121,7 @@ def service_main(*startup_parameters)
 
       def service_stop
         run_warning_displayed = false
-        Chef::Log.info("STOP request from operating system.")
+        Chef::Log.info("STOP requeschefm.")
```
coderanger: Mispaste.

lib/chef/knife.rb
```diff
@@ -182,7 +183,7 @@ def self.load_config(explicit_config_file, profile)
       config_loader.profile = profile
       config_loader.load
 
-      ui.warn("No knife configuration file found. See https://docs.chef.io/config_rb_knife.html for details.") if config_loader.no_config_found?
+      ui.warn("No config.rb (formerly #{Chef::Dist::KNIFE}.rb) configuration file found. See https://docs.chef.io/config_rb_knife.html for details.") if config_loader.no_config_found?
```
coderanger: Let's leave non-distro related changes out of this.

spec/integration/solo/solo_spec.rb
```diff
@@ -163,7 +163,7 @@
         ruby_block "sleeping" do
           block do
             retries = 200
-            while IO.read(Chef::Config[:log_location]) !~ /Chef client .* is running, will wait for it to finish and then run./
+            while IO.read(Chef::Config[:log_location]) !~ /#{Chef::Dist::PRODUCT} client .* is running, will wait for it to finish and then run./
```
coderanger: I don't think it's going to be helpful to include distro-variances in these unit tests. They care only about the project, not the distribution system. I would use `.*` in places to make the tests work under any configuration.

lib/chef/knife.rb
bobchaos: how would we formulate it so it's relevant regardless of knife's name?

lib/chef/application/windows_service.rb
bobchaos: :O fixed!

spec/integration/solo/solo_spec.rb
bobchaos: just pushed some fixes for this too

lib/chef/knife.rb
Tensibai: TBH I still wonder why knife is included in chef and not it's own product, it's mostly irrelevant for node management, I probably don't know how much Chef and knife codebase are interleaved that said

lamont-granquist: history, we could peel a knife gem out of the chef gem but it would have to dep on the chef gem and we'd ship it in the same repo and march versions in lockstep anyway, so the advantage would be fairly low.   we'd still ship it in the omnibus chef package, so it wouldn't reduce footprint any as well.  so, not a lot of practical impetus to break it out.

lamont-granquist:
This seems fine to me.

The most useful test I would think would be to ensure that the tests didn't fail when the Chef::Dist constants were changed which would be an additional grid of travis checks to modify those constants and then run all the tests.  I don't think you need to worry about automating that yet.

lib/chef/knife.rb
Tensibai: Fair enough :)

lib/chef/http/http_request.rb
```diff
@@ -40,8 +41,8 @@ class HTTPRequest
 
       engine = defined?(RUBY_ENGINE) ? RUBY_ENGINE : "ruby"
 
-      UA_COMMON = "/#{::Chef::VERSION} (#{engine}-#{RUBY_VERSION}-p#{RUBY_PATCHLEVEL}; ohai-#{Ohai::VERSION}; #{RUBY_PLATFORM}; +https://chef.io)".freeze
-      DEFAULT_UA = "Chef Client" << UA_COMMON
+      UA_COMMON = "/#{::Chef::VERSION} (#{engine}-#{RUBY_VERSION}-p#{RUBY_PATCHLEVEL}; ohai-#{Ohai::VERSION}; #{RUBY_PLATFORM}; +#{Chef::Dist::WEBSITE})".freeze
+      DEFAULT_UA = "#{Chef::Dist::PRODUCT} Client" << UA_COMMON
```
btm: This would make the UA 'Chef Infra Client', I wonder what is out there that would break?

lib/chef/knife.rb
tas50: I don't see the need to change knife.rb so I think this can same the way it is. knife.rb is the legacy config name and everyone should be on config.rb going forward which eliminates the need to worry about the file name.

lib/chef/http/http_request.rb
bobchaos: Similarly I've been wondering how many gems might look for a "Chef Client run completed" message as part of their tests or actual functions, among other things that would change :/

Tensibai: I can say my filebeat filter will break (one http_request at end of run with a bunch of retries to test the app is ok, cheap smoke test :))

bobchaos:
As discussed on Slack, I'm closing this in favor of https://github.com/chef/chef/pull/8368

PR #8342, opened 2019-04-05
Fix Debian 8 tests in Travis

tas50:
This is just a copy of @JohnRoesler's fix from haproxy cookbook


JohnRoesler:
Hmm, this actually doesn't appear to resolve the issue, despite the helpful guidance of the error message

PR #8344, opened 2019-04-05
Implement Chef::Resource#copy_properties_from

### LINKED ISSUE #7046: Port copy_properties_to from etcd cookbook into core chef
```ruby
def copy_properties_to(to, *properties)
  properties = self.class.properties.keys if properties.empty?
  properties.each do |p|
    # If the property is set on from, and exists on to, set the
    # property on to
    if to.class.properties.include?(p) && property_is_set?(p)
      to.send(p, send(p))
    end
  end
end
```

I think it would be better to have it shaped more like:

```ruby
declare_resource(:remote_file, name, &block).apply_properties(include: [ :group, :owner ], exclude: [ :headers ])
```

Maybe `include` is the obvious enough default to have it be `def apply_properties(*include, exclude: [])`

Kind of think it should hang off of Chef::Resource so it method chains better like that.  And rename it to not conflict with existing cookbooks.   Could also write a minimum viable compat_apply_properties cookbook to patch up old chef-client versions so other cookbooks could use it (not the entire crazysauce of compat_resource, but do-just-one-small-thing-only).

lamont-granquist:
Kept in the properties mixin since it is tightly coupled.

closes #7046

### DIFF
--- a/lib/chef/mixin/properties.rb
+++ b/lib/chef/mixin/properties.rb
@@ -315,6 +315,41 @@ class Chef
         raise ArgumentError, "Property #{name} is not defined in class #{self}" if !property
         property.description
       end
+
+      # Copy properties from another property object (resource)
+      #
+      # By default this copies all properties other than the name property (that is required to create the
+      # destination object so it has already been done in advance and this way we do not clobber the name
+      # that was set in that constructor).  By default it copies everything, optional arguments can be use
+      # to only select a subset.  Or specific excludes can be set (and the default exclude on the name property
+      # can also be overridden).  Exclude has priority over include, although the caller is likely better
+      # off doing the set arithmetic themselves for explicitness.
+      #
+      # action :doit do
+      #   # use it inside a block
+      #   file "/etc/whatever.xyz" do
+      #     copy_properties_from new_resource
+      #   end
+      #
+      #   # or directly call it
+      #   r = declare_resource(:file, "etc/whatever.xyz")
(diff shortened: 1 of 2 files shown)


lib/chef/mixin/properties.rb
```diff
@@ -315,6 +315,41 @@ def property_description(name)
...
+      #   r = declare_resource(:file, "etc/whatever.xyz")
+      #   r.copy_properties_from(new_resource, :owner, :group, :mode)
+      # end
+      #
+      # @param other [Object] the other object (Chef::Resource) which implements the properties API
+      # @param includes [Array<Symbol>] splat-args list of symbols of the properties to copy.
+      # @param exclude [Array<Symbol>] list of symbosl of the properties to exclude.
+      # @return the self object the properties were copied to for method chaining
+      #
+      def copy_properties_from(other, *includes, exclude: [ :name ])
```
tas50: could we add a whitelist feature as well here?

lamont-granquist: includes is the whitelist, exclude is the blacklist

README.md
```diff
@@ -1,3 +1,5 @@
+
```
tas50: Is this a Tim trolling change right here?

lib/chef/mixin/properties.rb
tas50: 🤦‍♂️

README.md
lamont-granquist: hahahaha, no this was me fighting with git aliases and it snuck in...

PR #8351, opened 2019-04-10
Prevent accidentally configuring windows_service properties

jasonwbarnett:
### Description

As I refactored the `windows_service` resource there are a couple of assumptions that I made originally that turn out to be more harmful than good. i.e. Assuming that when a user removes the `run_as_user` or `delayed_start` properties that they want those property defaults to be enforced. More verbosely stated:

If a user originally has a resource defined as such:

```ruby
windows_service 'MyApp' do
  run_as_user 'MyAppsUser'
  run_as_password 'MyAppsUserPassword'
  startup_type :automatic
  delayed_start true
  action [:configure, :start]
end
```

And then the user updates the resource to:

```ruby
windows_service 'MyApp' do
  startup_type :automatic
  action [:configure, :start]
end
```

I think it's more harmful to assume the user wants to configure the service using `LocalSystem` as the `run_as_user` than it is to assume the user does not want to configure the `run_as_user` or `delayed_start` properties at all.

So this change in combination with clearer documentation should go a long way in providing an overall better user experience.

### Issues Resolved

Doesn't directly solve #8080, but it would be re-introduced if #8318 were to be merged in and users followed the directions of the deprecation.

### Check List

- [ ] All tests pass

### DIFF
--- a/lib/chef/provider/service/windows.rb
+++ b/lib/chef/provider/service/windows.rb
@@ -208,11 +208,6 @@ class Chef::Provider::Service::Windows < Chef::Provider::Service
       return
     end
 
-    # Until #6300 is solved this is required
-    if new_resource.run_as_user == new_resource.class.properties[:run_as_user].default
-      new_resource.run_as_user = new_resource.class.properties[:run_as_user].default
-    end
-
     converge_if_changed :service_type, :startup_type, :error_control,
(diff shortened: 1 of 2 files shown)


jasonwbarnett:
@btm @tas50 ready for 👀

PR #8352, opened 2019-04-10
Avoid occasionally randomly reusing a gid in tests

btm:
Some platforms have groups in odd ranges, make sure the gid we select is
not in use.

### DIFF
--- a/spec/functional/resource/group_spec.rb
+++ b/spec/functional/resource/group_spec.rb
@@ -292,7 +292,18 @@ describe Chef::Resource::Group, :requires_root_or_running_windows do
     end
   end
 
-  let(:number) { rand(2000..9999) } # avoid low group numbers
+  let(:number) do
+    # Loop until we pick a gid that is not in use.
+    loop do
+      begin
+        gid = rand(2000..9999) # avoid low group numbers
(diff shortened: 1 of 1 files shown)


spec/functional/resource/group_spec.rb
```diff
@@ -292,7 +292,15 @@ def create_group
     end
   end
 
-  let(:number) { rand(2000..9999) } # avoid low group numbers
+  let(:number) do
+    loop do
+      gid = rand(2000..9999) # avoid low group numbers
```
jaymalasinha: Can we start at 8888?  That range should be good for testing and will avoid our AD ID range which often conflicts

btm: I originally was changing it to like `3000..6000` to avoid some groups on the rhel-7-power-le boxes where we failed recently because we hit an existing group. 

But what I did instead is to skip a gid if it exists already and pick a different one, so we don't have to play that game of trying to exclude any in use blocks. on the chance it hits one in use, it'll use another.

tas50:
You have a chefstyle here

btm:
@tas50 fixed. ran this on OS X and linux too, passes locally.

PR #8353, opened 2019-04-10
Stop printing environment in ci/verify.bat

jeremiahsnapp:


lamont-granquist:
so if you're doing this to prevent outputting secret values the issue here is that obviously anyone in the world can submit a PR to revert this commit, so this can't be a security mechanism...

(and i guess the description should really have something in it so that people know what this change was doing and why)

PR #8354, opened 2019-04-11
Add logic to require acceptannce of the Chef license to run the client

tyler-ball:
### Description

This adds a new feature to chef-client where users must accept the
EULA before using the product. Users can accept the license by
passing '--chef-license accept', setting ENV[CHEF_LICENSE]=accept or by
running 'chef-client' and following the interactive prompt.

### Issues Resolved

N/A

### Check List

- [x] All tests pass
- [x] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -143,3 +143,6 @@ subscriptions:
   - workload: ruby_gem_published:cheffish-*
     actions:
       - bash:.expeditor/update_dep.sh
+  - workload: ruby_gem_published:license-acceptance-*
+    actions:
+      - bash:.expeditor/update_dep.sh
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -143,6 +143,6 @@ subscriptions:
   - workload: ruby_gem_published:cheffish-*
     actions:
       - bash:.expeditor/update_dep.sh
-  - workload: ruby_gem_published:license-acceptance-*
-    actions:
-      - bash:.expeditor/update_dep.sh
+#  - workload: ruby_gem_published:license-acceptance-*
+#    actions:
+#      - bash:.expeditor/update_dep.sh
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem "cheffish", "~> 14"
 group(:omnibus_package) do
   gem "appbundler"
   gem "rb-readline"
-  gem "inspec-core", ">= 4.0.0.a", "< 5"
+  gem "inspec-core", "~> 4.3"
   gem "chef-vault"
   gem "ed25519" # ed25519 ssh key support done here as it's a native gem we can't put in train
   gem "bcrypt_pbkdf" # ed25519 ssh key support done here as it's a native gem we can't put in train
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,8 +16,8 @@ Chef Client requires a EULA to be accepted by users before it can run. Users can
 `CHEF_LICENSE=accept-no-persist chef-client`
 
 Finally, if users run `chef-client` without any of these options, they will receive an interactive prompt asking for
-license acceptance. If the license is accepted, a marker file will be written to the filesystem (unless `no-persist` is
-specified). Once this marker file is persisted, users no longer need to set any of these flags.
+license acceptance. If the license is accepted, a marker file will be written to the filesystem (unless
+`accept-no-persist` is specified). Once this marker file is persisted, users no longer need to set any of these flags.
 
 ### Allow Using --delete-entire-chef-repo in Chef Local Mode
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,9 +15,9 @@ Chef Client requires a EULA to be accepted by users before it can run. Users can
 `CHEF_LICENSE=accept chef-client`
 `CHEF_LICENSE=accept-no-persist chef-client`
 
-Finally, if users run `chef-client` without any of these options they will receive an interactive prompt asking for
-license acceptance. If the license is accepted a marker file will be written to the filesystem (unless `no-persist` is
-specified). Once this file is persisted users no longer need to set any of these flags.
+Finally, if users run `chef-client` without any of these options, they will receive an interactive prompt asking for
+license acceptance. If the license is accepted, a marker file will be written to the filesystem (unless `no-persist` is
+specified). Once this marker file is persisted, users no longer need to set any of these flags.
 
 ### Allow Using --delete-entire-chef-repo in Chef Local Mode
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,19 @@ Chef 15 release notes will be added here as development progresses.
 
 ## New Features / Functionality
 
+### Chef EULA
+
+Chef Client requires a EULA to be accepted by users before it can run. Users can accept the EULA in a variety of ways:
+
+`chef-client --chef-license accept`
+`chef-client --chef-license accept-no-persist`
+`CHEF_LICENSE=accept chef-client`
+`CHEF_LICENSE=accept-no-persist chef-client`
+
+Finally, if users run `chef-client` without any of these options they will receive an interactive prompt asking for
+license acceptance. If the license is accepted a marker file will be written to the filesystem (unless `no-persist` is
+specified). Once this file is persisted users no longer need to set any of these flags.
+
 ### Allow Using --delete-entire-chef-repo in Chef Local Mode
 
 ### Data Collection Ground-Up Refactor
--- a/chef-bin/bin/chef-apply
+++ b/chef-bin/bin/chef-apply
@@ -21,4 +21,4 @@
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 require "chef/application/apply"
 
-Chef::Application::Apply.new.run
+Chef::Application::Apply.new.run(true)
--- a/chef-bin/bin/chef-apply
+++ b/chef-bin/bin/chef-apply
@@ -21,4 +21,4 @@
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 require "chef/application/apply"
 
-Chef::Application::Apply.new.run(true)
+Chef::Application::Apply.new.run(enforce_license: true)
--- a/kitchen-tests/Berksfile
+++ b/kitchen-tests/Berksfile
@@ -2,3 +2,4 @@ source "https://supermarket.chef.io"
 
 cookbook "end_to_end", path: "cookbooks/end_to_end"
 cookbook "rspec", path: "cookbooks/rspec"
+cookbook "chef-client", git: "https://github.com/tyler-ball/chef-client.git", branch: "chef_license_accept"
--- a/kitchen-tests/Berksfile
+++ b/kitchen-tests/Berksfile
@@ -2,4 +2,3 @@ source "https://supermarket.chef.io"
 
 cookbook "end_to_end", path: "cookbooks/end_to_end"
 cookbook "rspec", path: "cookbooks/rspec"
-cookbook "chef-client", git: "https://github.com/tyler-ball/chef-client.git", branch: "chef_license_accept"
--- a/kitchen-tests/Gemfile
+++ b/kitchen-tests/Gemfile
@@ -6,5 +6,5 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids f
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
 gem "kitchen-dokken", "~> 2.0"
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
-gem "inspec", git: "https://github.com/inspec/inspec.git", branch: "master" # this goes away when we ship inspec 4
-gem "test-kitchen", git: "https://github.com/test-kitchen/test-kitchen.git", branch: "chef_license_accept"
+gem "inspec"
+gem "test-kitchen", git: "https://github.com/test-kitchen/test-kitchen.git", branch: "master"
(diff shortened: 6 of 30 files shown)


Gemfile
```diff
@@ -13,6 +13,8 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master"
 gem "chef-config", path: File.expand_path("../chef-config", __FILE__) if File.exist?(File.expand_path("../chef-config", __FILE__))
 gem "cheffish", "~> 14"
 
+gem "license-acceptance", "~> 0.2"
```
tas50: shouldn't this be in the gemspec?

spec/unit/application/knife_spec.rb
```diff
@@ -75,11 +77,7 @@ def run
       expect(@knife).to receive(:exit).with(0)
       @knife.run
     end
-    if windows?
```
tas50: that's not a weird if statement at all

RELEASE_NOTES.md
```diff
@@ -6,6 +6,19 @@ Chef 15 release notes will be added here as development progresses.
...
+### Chef EULA
+
+Chef Client requires a EULA to be accepted by users before it can run. Users can accept the EULA in a variety of ways:
+
+`chef-client --chef-license accept`
+`chef-client --chef-license accept-no-persist`
+`CHEF_LICENSE=accept chef-client`
+`CHEF_LICENSE=accept-no-persist chef-client`
+
+Finally, if users run `chef-client` without any of these options they will receive an interactive prompt asking for
```
```suggestion
Finally, if users run `chef-client` without any of these options, they will receive an interactive prompt asking for
```

RELEASE_NOTES.md
```diff
@@ -6,6 +6,19 @@ Chef 15 release notes will be added here as development progresses.
...
+license acceptance. If the license is accepted a marker file will be written to the filesystem (unless `no-persist` is
```
```suggestion
license acceptance. If the license is accepted, a marker file will be written to the filesystem (unless `no-persist` is
```

RELEASE_NOTES.md
```diff
@@ -6,6 +6,19 @@ Chef 15 release notes will be added here as development progresses.
...
+specified). Once this file is persisted users no longer need to set any of these flags.
```
```suggestion
specified). Once this file is persisted, users no longer need to set any of these flags.
```

spec/unit/application/knife_spec.rb
tyler-ball: Yeah - just found this randomly, figured I would do some cleanup.

tyler-ball:
@coderanger Yeah - we're actively having discussions internally about how this will work. Updates to this PR are coming. This code is all based on first pass implementation before we started having discussions about commercial vs public distributions.

🚧 🚧 🚧

tas50:
@tyler-ball we should get a dev-doc created for this that explains how the license gem interacts with chef/chef.

coderanger:
@tyler-ball Given the importance to the community, I would recommend moving those discussions to not be purely internal.

tyler-ball:
@coderanger It looks like we're continuing discussions in the community slack about how the distro will work. I would like to get this code accepted so we can get current channel omnibus builds going. This UX will be the same for the Chef Software Omnibus builds, we're just going to re-arrange the code to match the distro model later. With current channel builds we can start testing via Test Kitchen and users can start testing.

This also assumes I can get the integration tests fixed 😁

tyler-ball:
Okay - 🚧 is over! This PR implements the license-acceptance flow for the _current_ structure of the Chef repo. As the distribution discussions continue there will be changes to both this and the binstubs in this repo. We (Chef Software) will make sure to implement the changes to bring the license-acceptance flow in line with the distribution discussion outcomes. IE, we will activate the license-acceptance flow only in the Chef Software distributions of Chef Client.

kitchen-tests/Berksfile
```diff
@@ -2,3 +2,4 @@ source "https://supermarket.chef.io"
 
 cookbook "end_to_end", path: "cookbooks/end_to_end"
 cookbook "rspec", path: "cookbooks/rspec"
+cookbook "chef-client", git: "https://github.com/tyler-ball/chef-client.git", branch: "chef_license_accept"
```
lamont-granquist: that should probably get updated

kitchen-tests/Gemfile
```diff
@@ -9,4 +9,4 @@ gem "kitchen-dokken", "=1.1.1" # 2.x fails atm: https://travis-ci.org/chef/chef/
 gem "kitchen-docker", git: "https://github.com/test-kitchen/kitchen-docker.git", branch: "master"
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
 gem "kitchen-vagrant", git: "https://github.com/test-kitchen/kitchen-vagrant.git", branch: "master"
-gem "test-kitchen", git: "https://github.com/test-kitchen/test-kitchen.git", branch: "master"
+gem "test-kitchen", git: "https://github.com/test-kitchen/test-kitchen.git", branch: "chef_license_accept"
```
lamont-granquist: do we merge this first and remove the git pin here or is this chicken-and-egg?

kitchen-tests/cookbooks/end_to_end/Berksfile
```diff
@@ -1,3 +0,0 @@
-source "https://supermarket.chef.io"
-
-metadata
```
lamont-granquist: why did this disappear?

tasks/bin/run_external_test
```diff
@@ -13,7 +13,11 @@ git_thing = ARGV.shift
 
 build_dir = File.expand_path(ENV["TRAVIS_BUILD_DIR"] || Dir.pwd)
 
-env = { "GEMFILE_MOD" => "gem 'chef', path: '#{build_dir}'; gem 'ohai', git: 'https://github.com/chef/ohai.git', branch: 'master'" }
+env = {
+  "GEMFILE_MOD" => "gem 'chef', path: '#{build_dir}'; " \
+    "gem 'ohai', git: 'https://github.com/chef/ohai.git'",
+  "CHEF_LICENSE" => "accept-no-persist",
```
lamont-granquist: should this just go into a global variable in the travis configs?   this whole bit of code is hacky and probably shouldn't be extended.

tas50:
@tyler-ball we chatted about this in the foundation team review and I think there's a pretty minor change we could make which would ensure we don't break anyone trying to build chef as a distribution. Instead of always checking for license acceptance in the Application class we can pass a value into that class from the bin stub stating if the license should be validated or not. That boolean would default to false and any community distribution would just not set license enforcement in their binstub and the method would be a no-op.

kitchen-tests/Berksfile
tyler-ball: I don't have access to that repo to push a branch. If someone gives me access I can change that, otherwise this needs to stay

tyler-ball: Or we need to finish the changes to that cookbook and get the merged before this PR merges

kitchen-tests/Gemfile
tyler-ball: Its not a chicken and egg, we could merge the changes to TK first

kitchen-tests/cookbooks/end_to_end/Berksfile
tyler-ball: Because it did not seem in my testing to be used at all

tasks/bin/run_external_test
tyler-ball: It is set as a global variable in the travis config. But these tests must run in some kind of sandbox environment because I needed to set it here too

kitchen-tests/Berksfile
tyler-ball: ... Thinking about it, we should release the `chef-client` cookbook changes before releasing this, even if its only going to the `current` channel on PR merge

tyler-ball:
Moved the TK license acceptance from argument to client config
Depends on https://github.com/test-kitchen/test-kitchen/pull/1547

kitchen-tests/Gemfile
```diff
@@ -6,5 +6,5 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids f
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
 gem "kitchen-dokken", "~> 2.0"
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
-gem "inspec", git: "https://github.com/inspec/inspec.git", branch: "master" # this goes away when we ship inspec 4
-gem "test-kitchen", git: "https://github.com/test-kitchen/test-kitchen.git", branch: "master"
+gem "inspec"
+gem "test-kitchen", git: "https://github.com/tyler-ball/test-kitchen.git", branch: "test-tb"
```
markan: Let's make sure we have a checklist item to remove this once the test-kitchen change lands.

lib/chef/knife/bootstrap.rb
```diff
@@ -498,6 +507,35 @@ def render_template
         Erubis::Eruby.new(template).evaluate(bootstrap_context)
       end
 
+      # Check deprecated flags are used; map them to their new keys,
+      # and print a warning. Will not map a value to a new key if the
+      # CLI flag for that new key has also been specified.
+      # If both old and new flags are specified, this will warn
+      # and take the new flag value.
+      # This can be moved up to the base knife class if it's agreeable.
+      def warn_and_map_deprecated_flags
```
markan: Is this actually used, or is it replaced by verify_deprecated_flags!

lib/chef/knife/bootstrap.rb
```diff
@@ -18,11 +18,19 @@
 
 require "chef/knife"
 require "chef/knife/data_bag_secret_options"
+require "erubis"
+require "chef/knife/bootstrap/chef_vault_handler"
+require "chef/knife/bootstrap/client_builder"
+require "chef/util/path_helper"
+require "chef/dist"
+require "license_acceptance/cli_flags/mixlib_cli"
+# require "license_acceptance/acceptor"
```
markan: Can we remove this?

lib/chef/knife/bootstrap.rb
markan: This might be bad rebase pasta introduced by me...

lib/chef/knife/bootstrap.rb
tyler-ball: This should be un-commented because we use the acceptor below

lib/chef/knife/bootstrap.rb
tyler-ball: I think this and the added `require`s at the top of here are an issue. All this PR should be adding is the license requires and code

tyler-ball:
I want to get all the tests green then I will add back a purposefully failing test to prevent the build from promoting from unstable -> current

lib/chef/knife/bootstrap.rb
markan: I *think* you are agreeing with me that they should be removed; is that correct?

tas50:
@tyler-ball There's a conflict now due to the branding PR I just merged. Sorry about that

.expeditor/config.yml
```diff
@@ -143,3 +143,6 @@ subscriptions:
   - workload: ruby_gem_published:cheffish-*
     actions:
       - bash:.expeditor/update_dep.sh
+  - workload: ruby_gem_published:license-acceptance-*
```
tas50: Can you comment these subscriptions all out for now actually. They're not working and just create a ton of noise in the project.

lib/chef/application.rb
```diff
@@ -60,10 +61,11 @@ def reconfigure
     end
 
     # Get this party started
-    def run
+    def run(enforce_license = false)
```
tas50: It would be nice to do this as an options hash so it's more apparent what the true we're passing into new is up in the binaries. Right now you have to hop down here to see what is being set.

spec/unit/application/knife_spec.rb
```diff
@@ -75,11 +75,7 @@ def run
...
-      expect(Chef::Config[:color]).to be_truthy
-    else
-      expect(Chef::Config[:color]).to be_truthy
-    end
+    expect(Chef::Config[:color]).to be_truthy
```
tas50: that's a great spec there in the file

chef-bin/bin/chef-apply
```diff
@@ -21,4 +21,4 @@
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 require "chef/application/apply"
 
-Chef::Application::Apply.new.run
+Chef::Application::Apply.new.run(enforce_license: true)
```
tas50: Thanks. This is super clear now

kitchen-tests/Gemfile
```diff
@@ -6,5 +6,5 @@ gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids f
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
 gem "kitchen-dokken", "~> 2.0"
 gem "kitchen-inspec", git: "https://github.com/chef/kitchen-inspec.git", branch: "master"
-gem "inspec", git: "https://github.com/inspec/inspec.git", branch: "master" # this goes away when we ship inspec 4
```
tas50: part of me wonders if we want to keep this as an early warning device. If master of inspec will break us we'll know early

kitchen-tests/kitchen.yml
```diff
@@ -12,6 +12,7 @@ provisioner:
   name: dokken
   client_rb:
     diff_disabled: true
+  chef_license: "accept-no-persist"
```
tas50: Is this just in case someone runs this locally? Otherwise the env var should cover it.

tas50:
We might want to update the release notes to let folks know that they'll want to upgrade to chef-client cookbook 11.2 to manage their licenses in the client.rb

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -220,6 +220,8 @@
       let(:bootstrap) { Chef::Knife::Bootstrap.new(["--bootstrap-install-command", "chef-client"]) }
       before do
         bootstrap.config[:bootstrap_install_command] = "chef-client"
+        @old_env = ENV["CHEF_LICENSE"]
+        ENV["CHEF_LICENSE"] = "accept"
```
tyler-ball: I would prefer `accept-no-persist` if possible, so we don't accidently leave a marker file around and then have a test erroneously pass later

tyler-ball: I'll update this in the next PR I'm going to write which fixes `knife bootstrap` to only perform the license acceptance flow if the remote version is >= 15

tyler-ball:
Accepted review in zoom

PR #8358, opened 2019-04-11
Limit locale resource to linux platforms

btm:
Should fix a pipeline failure.


lib/chef/resource/locale.rb
```diff
@@ -24,6 +24,7 @@ class Locale < Chef::Resource
 
       description "Use the locale resource to set the system's locale."
       introduced "14.5"
+      provides :locale, platform_family: %w{rhel fedora amazon debian}
```
lamont-granquist: we've been going away from doing this.  stuff like bff_package being still declared on linux is useful for chefspec for example.

tas50:
We need to figure out a way to do this without reverting to lwrps, but I guess this achieves the end result for now.

lamont-granquist:
@btm you should be able to do this as a custom resource with:

```ruby
action_classs do
    [....]
  end
end
```

btm:
Replaced with a custom resource solution as suggested in: https://github.com/chef/chef/pull/8375

PR #8359, opened 2019-04-11
Don't force DSC functional tests to PS4

btm:
New Windows platforms like 2016 have Powershell 5.


tas50:
@btm there's a chefstyle failure here as well

btm:
@tas50 ha. fixed.

btm:
This passed an ad-hoc test on the windows testers alright. I'm not sure yet if it fixes our broken state though. @jaymalasinha thinks it might be happening after a chef-14 build, but rebuilt the tester again so I'm not sure if it's in a bad state or not yet.

Either way, I think this is an improvement to the functional tests, since we run different commands for PS4 and PS5.

Issue #6375, opened 2017-09-01
Chef-Solo loads configurations twice


gmeneze:
## Description

chef-solo run loads the configuration file (solo.rb) twice when it's run. This causes duplication of handlers,  resulting in the same handler being run twice.

## Chef Version
We were able to replicate this for versions 12.14.89+

## Platform Version
Red Hat Enterprise Linux Server release 6.8 (Santiago)

## Replication Case
1. Create a minimal solo.rb with an event handler : -
```
Chef.event_handler do
  on :converge_start do
    puts "Ohai! I have started a converge."
  end
end
```
2.  Run chef solo : -
```
sudo chef-solo -c ./solo.rb
```

## Solo Output
```
$ sudo chef-solo -c ./solo.rb
[2017-09-01T20:10:07+00:00] WARN: No cookbooks directory found at or above current directory.  Assuming /var/chef.
[2017-09-01T20:10:07+00:00] WARN: No cookbooks directory found at or above current directory.  Assuming /var/chef.
Starting Chef Client, version 13.2.20
resolving cookbooks for run list: []
Synchronizing Cookbooks:
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 0 resources
Ohai! I have started a converge.
Ohai! I have started a converge.

Running handlers:
Running handlers complete
Chef Client finished, 0/0 resources updated in 04 seconds
Running handlers:
Running handlers complete
Chef Client finished, 0/0 resources updated in 04 seconds
```

The converge_start event handler is run twice, resulting in the statement  "Ohai! I have started a converge" being printed twice.

## My Analysis
When the chef-solo executable is run, it first loads the configurations in its run method, here : -
https://github.com/chef/chef/blob/master/lib/chef/application/solo.rb#L222

If the legacy mode flag is not passed to the chef-solo run, then this block of code is executed later on in the run method : -
https://github.com/chef/chef/blob/master/lib/chef/application/solo.rb#L225

This calls the run method of the application class, which calls reconfiguration method again, so the configuration file is loaded for the second time here : -
https://github.com/chef/chef/blob/master/lib/chef/application.rb#L57

This is fine for configurations like checksum_path, cookbook_path etc. as they are overwritten with the same value, but for chef-handlers (start_handlers, report_handlers, exception_handlers, event_handler), which are stored as arrays, it results in two similar objects getting inserted, which causes this problem.

lamont-granquist:
closing as a dup of #8366 and should be fixed by work in #8872 whenever that lands

PR #8361, opened 2019-04-11
Drop privileges before creating files in solo mode

### LINKED ISSUE #8360: Chef solo with --user drops permissions after it has created artifacts
## Description
When using `chef-solo --user foo --recipe-url` the cache directory (and others) has to be preconfigured to be owned by 'foo'.

## Chef Version
14.11.21

## Platform Version
Ubuntu Linux 18.04.2

## Replication Case
from @teknofire:

create /tmp/solo.rb:
```
file_cache_path "/tmp/chef-solo/cache/"
file_backup_path "/tmp/chef-solo/backup/"
cookbook_path "/tmp/chef-solo/recipes/cookbooks/"
log_level :info
```
then run:
```
chef-solo --format doc --recipe-url https://github.com/teknofire/chef-solo-test/releases/download/v1.0/chef-solo.tar.gz --user opscode --override-runlist 'recipe[testing]' --config /tmp/solo.rb 
```

## Client Output
```
[2019-04-11T16:48:57-04:00] INFO: About to change privilege to opscode
[2019-04-11T16:48:57-04:00] INFO: Started chef-zero at chefzero://localhost:1 with repository at /tmp/chef-solo/recipes
  One version per cookbook

[2019-04-11T16:48:57-04:00] FATAL: Errno::EACCES: Permission denied @ dir_s_mkdir - /tmp/chef-solo/cache
```

btm:
Fixes #8360

for discussion

### DIFF
--- a/lib/chef/application/solo.rb
+++ b/lib/chef/application/solo.rb
@@ -218,12 +218,12 @@ class Chef::Application::Solo < Chef::Application
   # Get this party started
   def run
     setup_signal_handlers
+    setup_application
     reconfigure
     for_ezra if Chef::Config[:ez]
     if !Chef::Config[:solo_legacy_mode]
       Chef::Application::Client.new.run
     else
-      setup_application
       run_application
     end
   end


tas50:
Any idea why we previously kept this permission setup out of legacy mode?

btm:
@tas50 this moves it out of the legacy mode portion of the conditional. The other half of the conditional that calls `Chef::Application::Client.new.run` is the chef-zero mode, and everything in `run`, including ` setup_application` is called again a second time. That happens to be #6375 and doesn't have an easy solution 

In general when running daemonized I think it's common to read/write some files as root before you drop privileges, like secrets. I'm not sure I'd want to change that behavior in Chef::Application itself which gets inherited into Chef::Application::Client. But Solo doesn't daemonize, so I think it's fine to drop privileges then create the files in that regard. But I don't know if that would break another use case. I can't think of one that would work though.

lamont-granquist:
So why is there the same setup_application method in Chef::Application::Client which is not called from Chef::Application::Client#run?  Because now chef-solo-legacy and chef-solo-zero both are doing this, but chef-client is not.  And chef-solo (both ways) should support the same daemonization options as chef-client, at least in principle.

btm:
@lamont-granquist Chef::Application::Client inherits run from Chef::Application, which does call setup_application.

https://github.com/chef/chef/blob/master/lib/chef/application.rb#L65

PR #8362, opened 2019-04-11
Add new client_rekey resource

lamont-granquist:
Based on https://github.com/chef-boneyard/client-rekey only with none of the legacy, written as a custom resource.


lib/chef/resource/rekey.rb
```diff
@@ -0,0 +1,86 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  class Resource
+    class Rekey < Chef::Resource
+      provides :rekey
+      resource_name :rekey
```
tas50: seems like this would make more sense if we named it client_rekey

lamont-granquist: yeah that makes sense, will do.

tas50:
@lamont-granquist what's left here before we can ship this?

tas50:
@lamont-granquist anything else that needed to be done to this one?

lamont-granquist:
yeah still needed decent manual testing i think which i ran into snags trying to do

lamont-granquist:
(or decent non-manual testing but that would require a chef-zero that could do auth or spinning up a chef-server for testing)

lib/chef/resource/client_rekey.rb
```diff
@@ -0,0 +1,114 @@
...
+      description "The client_rekey resource regenerates the key used by #{Chef::Dist::PRODUCT} to communicate with the #{Chef::Dist::SERVER_PRODUCT}."
+      introduced "15.6"
+
+      examples <<~DOC
+        Rekey the node
+        ```ruby
+        client_rekey
+        ```
+
+        Rekey the node with a custom key path
```
NoamDolovichWS: noam

tas50:
Closing this one out for now since we don't have a way to test this and we shouldn't pull it in without that.

PR #8363, opened 2019-04-12
Add support for comment blocks to sysctl resource

lanky:
### Description
Adds support for comments in sysctl files under sysctl.d - permits sysctl settings to be documented/explained for troubleshooting (and informational) purposes.

The comment property is optional and accepts both Strings and Arrays (of strings). 

### Issues Resolved
N/A
### Check List

- [X] All tests pass
- [X ] RELEASE\_NOTES.md has been updated if required (not required for bugfixes, required for API changes)
- [X ] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/lib/chef/resource/sysctl.rb
+++ b/lib/chef/resource/sysctl.rb
@@ -45,6 +45,10 @@ class Chef
                coerce: proc { |v| coerce_value(v) },
                required: true
 
+      property :comment, [Array, String],
+               description: "An optional comment to prepend the resource setting.",
+               default: []
+
       property :conf_dir, String,
                description: "The configuration directory to write the config to.",
                default: "/etc/sysctl.d"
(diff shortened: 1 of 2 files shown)


lib/chef/resource/sysctl.rb
```diff
@@ -45,6 +45,10 @@ class Sysctl < Chef::Resource
                coerce: proc { |v| coerce_value(v) },
                required: true
 
+      property :comment, [Array, String],
+               description: "An optional comment to prepend the resource setting.",
```
tas50: We generate our docs from these comments so you'll want to briefly explain that multi-line comments can be accomplished by using an array of strings where each line is a value in the array.

lanky: OK, have updated and pushed.  Thanks.

lanky: hmmm, that didn't work. All tests pass locally. moving it back to a single line to test.

lib/chef/resource/sysctl.rb
```diff
@@ -45,6 +45,10 @@ class Sysctl < Chef::Resource
                coerce: proc { |v| coerce_value(v) },
                required: true
 
+      property :comment, [Array, String],
+               description: "Optional comments, placed above the resource setting in the generated file. For multi-line comments, use an array of strings, one per line.",
```
lanky: the only change in the last commit is to make this description longer. Not sure why the tests are now failing?

PR #8370, opened 2019-04-13
Allow empty strings in -o to result in empty override run list

lamont-granquist:
Previously this was not possible.  Now you can do this:

```
chef-client -o "" ./foo.rb
```

And it will only run foo.rb and not any of the recipes in your
run_list at all.  Node will not be saved at the end.

Useful to one-shot something like keyrotation (which still needs
keys and needs to talk to the chef server API, so this is a
different use case from simple chef-apply).

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -1,7 +1,7 @@
 #
 # Author:: AJ Christensen (<aj@chef.io>)
 # Author:: Mark Mzyk (mmzyk@chef.io)
-# Copyright:: Copyright 2008-2018, Chef Software Inc.
+# Copyright:: Copyright 2008-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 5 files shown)


lamont-granquist:
Note that `chef-client ./foo.rb` will always run foo.rb after all the files in the run_list and it was originally engineered this way years and years ago back in chef 11.x era, so I've preserved that deliberate API choice.  I kinda think if I was greenfielding it, I would have had that implicitly override + clear the run_list, but that isn't the choice that was made originally.  I can see the logic though since then the extra script becomes additive to the set of recipes that are already running on the node, so don't think its worth trying to break that API since it is also useful and we'd just wind up having to support that some other way.

So, the change here is instead an extension to allow explicitly clearing the run list, and should be entirely backwards compatible (there were a ton of tests which broke along the way and the implementation evolved a lot).

PR #8375, opened 2019-04-15
Limit locale resource to Linux

btm:
Platforms that don't use /etc/locale.conf aren't supported

Replaces #8358. Also fixes an intermittent failure on MacOS.


lamont-granquist:
yeah there we go.

btm:
Looks like RHEL + friends lack locale-gen and use localedef instead, which we'll need to add support for.

PR #8379, opened 2019-04-16
Add :leave action to windows_ad_join resource

jasonwbarnett:
### Description

Add `:leave` action to windows_ad_join resource

### Issues Resolved

N/A

### Check List

- [ ] All tests pass


lib/chef/resource/windows_ad_join.rb
```diff
@@ -91,11 +95,59 @@ class WindowsAdJoin < Chef::Resource
...
+                reason "Reboot to leave domain #{new_resource.domain_name}"
+              end
+            end
+          end
+        end
+      end
+
       action_class do
-        def on_domain?
+        def node_domain
```
tas50: can you add a yard comment here

jasonwbarnett: Yes.

tas50:
This sort of breaks the noun / adverb model of resources and actions, but I think the damage is minimal and I'm not opposed to this.

jasonwbarnett:
> This sort of breaks the noun / adverb model of resources and actions, but I think the damage is minimal and I'm not opposed to this.

I agree, this logically makes little sense. Maybe it should be split into its own resource `windows_ad_leave`? I haven't thought in depth about this and I'm not sure what trade-offs we would be making here.

PR #6777, opened 2018-01-23
Revert "Revert "Fail on interval runs on windows""

lamont-granquist:
Reverts chef/chef#6776

PR #8392, opened 2019-04-18
Allow unforked interval runs on Windows in solo

tas50:
This matches the behavior of chef-client. I also updated the messaging since no one cares that we allowed it in Chef 11.


lamont-granquist:
this is the wrong way, we want to disallow unforked interval runs across the board.

lamont-granquist:
https://github.com/chef/chef/pull/6777/files

tas50:
Didn't that not work at all though?

tas50:
NM I see this is the new version of the original change.

PR #8393, opened 2019-04-18
Don't require all of chef in chef-solo

tas50:
This greatly speeds up basic operations like -v which goes from 1.6 to
.3 seconds on my system.


tas50:
We actually are doing this in chef-client just differently.

PR #8397, opened 2019-04-19
Remove chef-* binstubs from chef gem

lamont-granquist:
This is for the relicensing effort.

Note that this PR leaves the knife and ohai binstubs still in the
gem(s) while that discussion is still ongoing.


chef.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_dependency "bundler", ">= 1.10"
 
   s.bindir       = "bin"
-  s.executables  = %w{ chef-client chef-solo knife chef-shell chef-apply chef-resource-inspector }
```
tas50: Do we really need to remove chef-resource-inspector and chef-shell?

lamont-granquist: Per discussion in standup, i decided those didn't matter (we have to do chef-client and chef-solo which is where the user pain will all come from) and so removing all the chef-* binstubs optimizes for consistency and clarity and making upper management happy while not inflicting substantially more pain than was already there.  And discussions with management have revolved around removing `chef*` (where there's a trademark issue) being the dividing line.  So I just didn't want to die on the hill of trying to argue that chef-resource-inspector needed to be kept, because I kinda think that it doesn't...

coderanger:
Just for the record, I'm not a fan of this change. We've never officially supported `gem install chef` as an install style, but this will make it entirely unavailable. This will make installing Chef on unusual environments (RasPis, etc) more difficult, and will undoubtedly frustrate some number of people that built their own install systems around `gem install chef`.

That said, I have no energy to argue this point, I just wanted to side effects listed so people will have an easier time finding the change which broke things for them.

lamont-granquist:
closed in favor of #8413

PR #8403, opened 2019-04-20
Replace kitchen-appbundle-updater in Travis tests with Test Kitchen lifecycle hooks

lamont-granquist:
update of kitchen end-to-end testing for the modern world

### DIFF
--- a/kitchen-tests/Gemfile
+++ b/kitchen-tests/Gemfile
@@ -4,8 +4,7 @@ gem "rake" # required to build some native extensions
 gem "chef", path: ".."
 gem "ohai", git: "https://github.com/chef/ohai.git", branch: "master" # avoids failures when we bump chef major
 gem "berkshelf", git: "https://github.com/berkshelf/berkshelf.git", branch: "master"
-gem "kitchen-appbundle-updater"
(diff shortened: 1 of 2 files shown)


lamont-granquist:
so @tas50 my only concern about this is that we're pulling straight from whatever dokken image now and i don't know if those are up to date with latest current omnibus builds or what flags to use to get them there?

lamont-granquist:
oh its right there, must have been getting late.

lamont-granquist:
So this will work well with the kitchen-dokken setup because the image comes in with the chef omnibus installation from current.

That means that we can use the pre-converge hook to hack up the image with appbundle-updater.

To do this on any other provisioner than kitchen-dokken would be difficult because you would want a lifecycle hook inside of the provisioner that I don't think exists -- because the provisioner handles the whole converge phase which normally installs omnibus-chef and runs the client-client run as a single stage -- while we would need to insert appbundle-updater commands in between
those two events inside of the converge phase.

It probably indicates that the chef provisioner needs to support its own sub-hook (or T-K could split up the converge phase but that is likely way too hard to do).  But we don't need it here 'cuz dokken bakes in the chef-client into the image.

lamont-granquist:
@tas50 this should be ready to go, RIP kitchen-appbundle-updater

PR #8409, opened 2019-04-23
Add comment block to sysctl resource

lanky:
### Description

Adds optional comments to the sysctl resource, which are prepended to any settings. Allows for documentation of sysctls on the target system.

### Issues Resolved
This is a new feature, so none.

### Check List

- [X] All tests pass
- [X] RELEASE\_NOTES.md has been updated if required (not required for bugfixes, required for API changes)
- [X] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -330,6 +330,7 @@ which results in `/etc/sysctl.d/99-chef-vw.swappiness.conf` as follows
 # than the high water mark in a zone.
 # The default value is 60.
 vm.swappiness = 10
+```
 
 ## New Resources
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -301,7 +301,7 @@ Chef Solo's `--delete-entire-chef-repo` option has been extended to work in Loca
 
 ### sysctl now accepts a comments parameter
 
-The sysctl resource has been updated to allow additional (optional) comments. Comments can be passed as an array or a string. Any comments provided are prefixed with '#' signs and precede the sysctl setting in generated files.
+The `sysctl` resource has been updated to allow the inclusion of descriptive comments. Comments may be passed as an array or a string. Any comments provided are prefixed with '#' signs and precede the `sysctl` setting in generated files.
 
 An example:
 
@@ -320,7 +320,7 @@ sysctl 'vm.swappiness' do
 end
 ```
 
-which results in `/etc/sysctl.d/99-chef-vw.swappiness.conf` as follows
+which results in `/etc/sysctl.d/99-chef-vm.swappiness.conf` as follows
 ```
 # define how aggressively the kernel will swap memory pages.
 # Higher values will increase aggressiveness
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -299,6 +299,38 @@ Our underlying SSH implementation has been updated to support the new ed25519 SS
 
 Chef Solo's `--delete-entire-chef-repo` option has been extended to work in Local Mode as well. Be warned that this flag does exactly what it states, and when used incorrectly, can result in loss of work.
 
+### sysctl now accepts a comments parameter
+
+The sysctl resource has been updated to allow additional (optional) comments. Comments can be passed as an array or a string. Any comments provided are prefixed with '#' signs and precede the sysctl setting in generated files.
+
+An example:
+
+```
+sysctl 'vm.swappiness' do
+  value 10
+  comment [
+     "define how aggressively the kernel will swap memory pages.",
+     "Higher values will increase aggressiveness",
+     "lower values decrease the amount of swap.",
+     "A value of 0 instructs the kernel not to initiate swap",
+     "until the amount of free and file-backed pages is less",
+     "than the high water mark in a zone.",
+     "The default value is 60."
+    ]
+end
+```
+
+which results in `/etc/sysctl.d/99-chef-vw.swappiness.conf` as follows
+```
+# define how aggressively the kernel will swap memory pages.
+# Higher values will increase aggressiveness
+# lower values decrease the amount of swap.
+# A value of 0 instructs the kernel not to initiate swap
+# until the amount of free and file-backed pages is less
+# than the high water mark in a zone.
+# The default value is 60.
+vm.swappiness = 10
+
 ## New Resources
 
 ### archive_file resource
--- a/lib/chef/resource/sysctl.rb
+++ b/lib/chef/resource/sysctl.rb
@@ -45,6 +45,10 @@ class Chef
         coerce: proc { |v| coerce_value(v) },
         required: true
 
+      property :comment, [Array, String],
+        description: "Comments, placed above the resource setting in the generated file. For multi-line comments, use an array of strings, one per line.",
+        default: []
+
       property :conf_dir, String,
         description: "The configuration directory to write the config to.",
         default: "/etc/sysctl.d"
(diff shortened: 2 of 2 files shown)


lanky:
The buildkite test appears to have nothing to do with my PR, AFAICT?
If so, anyone able to point out how? There doesn't appear to be any way to trigger a re-run of this?

lib/chef/resource/sysctl.rb
```diff
@@ -80,8 +84,22 @@ def coerce_value(v)
...
 
+          # construct a string, joining members of new_resource.comment
+          comment_lines = []
+          if new_resource.comment.is_a?(Array)
+            new_resource.comment.each do |line|
+              comment_lines << "# #{line.strip}"
+            end
+          else
+            comment_lines << "# #{new_resource.comment.strip}"
+          end
```
lamont-granquist: this should be done in a more ruby way, probably:

```
comment_lines = Array(new_resource.comment).map { |c| "# #{c.strip}" }
```

lib/chef/resource/sysctl.rb
```diff
@@ -80,8 +84,22 @@ def coerce_value(v)
...
+
+          comment_lines << "#{new_resource.key.tr('/', '.')} = #{new_resource.value}\n\n"
```
lamont-granquist: this is  not a comment line to the variable should be renamed.

lib/chef/resource/sysctl.rb
```diff
@@ -80,8 +84,22 @@ def coerce_value(v)
...
+
+          fcontent = comment_lines.join("\n")
+
           file "#{new_resource.conf_dir}/99-chef-#{new_resource.key.tr('/', '.')}.conf" do
-            content "#{new_resource.key} = #{new_resource.value}"
+            content fcontent
```
lamont-granquist: the join statement can be moved down here to avoid the fcontent variable

lanky:
I must apologise, I (foolishly) rebased before continuing work on this and you now have many many commits in this PR
The only changes I was making are to `lib/chef/resource/sysctl.rb` and `RELEASE_NOTES.md`

RELEASE_NOTES.md
```diff
@@ -37,6 +37,39 @@ See our [Frequently Asked Questions document](https://www.chef.io/bmc-faq/) for
 
 ## New Features / Functionality
 
+### sysctl now accepts a comments parameter
```
tas50: parameter -> property

lib/chef/resource/sysctl.rb
```diff
@@ -45,6 +45,10 @@ class Sysctl < Chef::Resource
                coerce: proc { |v| coerce_value(v) },
                required: true
 
+      property :comment, [Array, String],
+               description: "Optional comments, placed above the resource setting in the generated file. For multi-line comments, use an array of strings, one per line.",
```
tas50: You can leave out the Optional bit here.

lib/chef/resource/sysctl.rb
```diff
@@ -80,8 +84,13 @@ def coerce_value(v)
 
           directory new_resource.conf_dir
 
+          # construct a string, joining members of new_resource.comment
+          sysctl_lines = Array(new_resource.comment).map { |c| "# #{c.strip}" }
+
+          sysctl_lines << "#{new_resource.key.tr('/', '.')} = #{new_resource.value}\n\n"
```
tas50: Is there a reason you introduced the tr('/', '.')here?

To keep this no-op for existing users you should just:

sysctl_lines << "#{new_resource.key} = #{new_resource.value}"

That way everyone won't get upgraded resources with the new lines

tas50:
@lanky This looks really promising and I'm sure a lot of people would find it handy. Take a look at the comments I left and rebase on master and we should be good to go. Feel free to reach out to. me on the community Slack if you have any issues.

lanky:
due to a number of conflicting changes this is now a consolidated commit, affecting only the `sysctl.rb` and `RELEASE_NOTES.md` files with changes from current master

RELEASE_NOTES.md
```diff
@@ -299,6 +299,38 @@ Our underlying SSH implementation has been updated to support the new ed25519 SS
...
+```
+# define how aggressively the kernel will swap memory pages.
+# Higher values will increase aggressiveness
+# lower values decrease the amount of swap.
+# A value of 0 instructs the kernel not to initiate swap
+# until the amount of free and file-backed pages is less
+# than the high water mark in a zone.
+# The default value is 60.
+vm.swappiness = 10
+
```
tduffield: You appear to be missing ending codeblock ticks.

lib/chef/resource/sysctl.rb
```diff
@@ -45,6 +45,10 @@ class Sysctl < Chef::Resource
         coerce: proc { |v| coerce_value(v) },
         required: true
 
+      property :comment, [Array, String],
+        description: "Comments, placed above the resource setting in the generated file. For multi-line comments, use an array of strings, one per line.",
```
tduffield: Why an array of strings vs a single multi-line string?

RELEASE_NOTES.md
lanky: ah, you are of course correct. Fixed, will push shortly.

lib/chef/resource/sysctl.rb
lanky: Simplicity and ease of use - prepending '#' to the beginning of each entry in an array is a simple task. I'm sure it's quite trivial for multi-line strings too, but this to me is more readable.
Personal preference, admittedly.

lanky:
buildkite failures appear to be unit/integration tests on Windows (And look like they aren't related to this PR at all).

RELEASE_NOTES.md
```diff
@@ -299,6 +299,39 @@ Our underlying SSH implementation has been updated to support the new ed25519 SS
 
 Chef Solo's `--delete-entire-chef-repo` option has been extended to work in Local Mode as well. Be warned that this flag does exactly what it states, and when used incorrectly, can result in loss of work.
 
+### sysctl now accepts a comments parameter
```
mjingle: "Parameter" in this header should be "Property."

lib/chef/resource/sysctl.rb
mjingle: Descriptions for properties should ideally be full sentences. Please change the first sentence to include a verb. For example, "Comments _are_ placed above the resource setting in the generated file."? Feel free to rework/insert your own verb.

mjingle: FYI The reason why these descriptions are so important is because they are aggregated from this file and then displayed on our docs website.

RELEASE_NOTES.md
```diff
@@ -299,6 +299,39 @@ Our underlying SSH implementation has been updated to support the new ed25519 SS
...
+
+The sysctl resource has been updated to allow additional (optional) comments. Comments can be passed as an array or a string. Any comments provided are prefixed with '#' signs and precede the sysctl setting in generated files.
```
mjingle: Formatting request: Please put tick marks on `sysctl` for both "`sysctl` resource" in the first sentence, and for "the `sysctl` setting" in the last sentence.

mjingle: I think the first sentence would benefit from removing the parenthesis, so that:
> The sysctl resource has been updated to allow additional**,** optional comments.
You should use a comma as these adjectives are coordinate adjectives.

lanky: Thought I'd done these. Apologies for the delay.
I've actually reworded it to remove that part.

tas50:
Thanks for working with us on this @lanky

PR #8410, opened 2019-04-24
 Move ed25519 gems into omnibus

tas50:
We had to move this out of train as they were causing installation failures for inspec. This change makes sure we get them into our omnibus package.


lamont-granquist:
except single quoted strings

PR #8413, opened 2019-04-24
Remove chef-* bin files from chef gem

lamont-granquist:
This is for the relicensing effort.

Note that this PR leaves the knife and ohai bin files still in the
gem(s) while that discussion is still ongoing.

This PR extracts the bin files to chef-bin gem which is not intended
to be redistributed outside of the omnibus chef binaries.

A blank gem for chef-bin has been pushed to rubygems.org for
obvious security concerns and so that bundler/rubygems can resolve
the gem


chef-binstubs/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+   To apply the Apache License to your work, attach the following
+   boilerplate notice, with the fields enclosed by brackets "[]"
+   replaced with your own identifying information. (Don't include
+   the brackets!)  The text should be enclosed in the appropriate
+   comment syntax for the file format. We also recommend that a
+   file or class name and description of purpose be included on the
+   same "printed page" as the copyright notice for easier
+   identification within third-party archives.
+
+Copyright [yyyy] [name of copyright owner]
```
miah: Do we need to fill this in?

lamont-granquist: heh if we do then we need to do that in chef-config/LICENSE as well, I just did a `cp -r` and then started hacking things away and renaming them....

PR #8417, opened 2019-04-25
file: Tell people what file a link is pointing at in warning messages

jaymzh:
### Description

In the event you are managing a file `/etc/file` with a `file` resource,
but it turns out to be a link to `/etc/passwd` and you are trying to figure
out what overwrite `/etc/passwd`, there is no indication in the logs. Let's add
one.

### Check List

- [x] All tests pass

### DIFF
--- a/lib/chef/win32/file.rb
+++ b/lib/chef/win32/file.rb
@@ -89,6 +89,14 @@ class Chef
         is_symlink
       end
 
+      def self.realpath(file_name)
+        if symlink?(file_name)
+          readlink(file_name)
+        else
+          file_name
+        end
+      end
+
       # Returns the path of the of the symbolic link referred to by +file+.
       #
       # Requires Windows Vista or later. On older versions of Windows it
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -227,7 +227,7 @@ class Chef
         elsif file_class.symlink?(path) && new_resource.manage_symlink_source
           verify_symlink_sanity(path)
         elsif file_class.symlink?(new_resource.path) && new_resource.manage_symlink_source.nil?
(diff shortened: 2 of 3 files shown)


tas50:
Looks like realpath isn't going to work on Windows @jaymzh

jaymzh:
Fucking Windows. Is there... something that does?

jaymzh:
OK I think I added support

Gemfile.lock
```diff
@@ -56,47 +56,6 @@ PATH
       syslog-logger (~> 1.6)
       tty-screen (~> 0.6)
       uuidtools (~> 2.1.5)
-    chef (15.0.233-universal-mingw32)
```
tas50: Looks like you just ran bundle update on the gemfile.lock. That removes the windows specific gems. Use the rake task to update the deps instead, but make sure not to bump the bundle version line at the bottom of the file.

jaymzh: ah damnit. Thanks.

jaymzh:
@tas50  so fixing that, at lest running locally the linux stuff, I get a bunch of failures when I do `bundle exec rspec spec/unit/providers/file_spec.rb` but when I run ALL the specs, I don't. wat.

jaymzh:
OK @tas50 light is green, trap is clean!

lib/chef/win32/file.rb
```diff
@@ -89,6 +89,14 @@ def self.symlink?(file_name)
         is_symlink
       end
 
+      def self.realpath(file_name)
+        if symlink?(file_name)
+          readlink(file_name)
+        else
+          filename
```
lamont-granquist: `file_name`

jaymzh: nice catch!! Updated version pushed. Thanks.

PR #8439, opened 2019-04-30
fix chef-bin bundling in omnibus

lamont-granquist:
this file winds up in the embedded lib at:

/opt/chef/embedded/lib/ruby/gems/x.y.0/gems/chef-X.Y.Z/Gemfile

It needs to find the chef-bin gem from the same library, rather than
simply omitting it.

The chef-config gem comes in through the chef.gemspec so the lines are
different.


jaymalasinha:
Do we also need this in `omnibus/Gemfile` ?

lamont-granquist:
no it should work fine without that there.

Issue #8432, opened 2019-04-30
knife bootstrap winrm error outs when we use ip address of the VM to bootstrap along with -o option along with it. 


Vasu1105:
## Description
When I use following command it errors out `Exception: NameError: uninitialized constant Net::SSH`

I am running following command from my workstation to bootstrap VM.
```
knife bootstrap 137.117.17.107 -o winrm --node-name vj-win2012r2 --connection-user 'username' --connection-password 'password' --connection-port 5985 --winrm-auth-method "negotiate" -c ~/workspace/chef-repo/.chef/config.rb
```
Where as if I use following syntax it does not break and proceeds for further execution
```
knife bootstrap winrm://137.117.17.107 --node-name vj-win2012r2 --connection-user 'azure' --connection-password 'password' --connection-port 5985 --winrm-auth-method "negotiate" -c ~/workspace/chef-repo/.chef/knife.rb
```

## Chef Version

```Chef Infra: 15.0.238```
## Platform Version

Workstation: Ubuntu 18.04

Trying to bootstrap Windows-2012R2

## Replication Case
After running this command 
```
knife bootstrap 137.117.17.107 -o winrm --node-name vj-win2012r2 --connection-user 'username' --connection-password 'password' --connection-port 5985 --winrm-auth-method "negotiate" -c ~/workspace/chef-repo/.chef/config.rb
```
## Stacktrace
```
Connecting to 137.117.17.107
Traceback (most recent call last):
	18: from /usr/bin/knife:80:in `<main>'
	17: from /usr/bin/knife:80:in `load'
	16: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.238/bin/knife:24:in `<top (required)>'
	15: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.238/lib/chef/application/knife.rb:162:in `run'
	14: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.238/lib/chef/knife.rb:221:in `run'
	13: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.238/lib/chef/knife.rb:459:in `run_with_pretty_exceptions'
	12: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.238/lib/chef/local_mode.rb:41:in `with_server_connectivity'
[...]

PR #8440, opened 2019-04-30
[CHEF-8432] Ensure default protocol is used properly. Use correct 'require' before accessing Net::SSH constants.

marcparadise:
## Description

This PR corrects two problems found in #8432: 
1. when protocol was not given in the URL (winrm://hostname) set the default protoocol - even if '-o' was given - because it was lost in incorrect order of operations. 
2. when a `TrainConnector#connect!` failed, if protocol was not `ssh` we would  would receive undefined constant `Net::SSH` - we check for a specific `Net::SSH` exception, but if we don't hit ssh path in train `net/ssh` is not loaded. 
3. In addition, it cleans up TrainConnector a bit: 
* remove `#test_instance` -- we can do the same setup in tests
* finish moving config setup to a lazy load 
* Naming was made more consistent within the module 
* Removed a couple helper methods that didn't make sense anymore
* add yard doc

## Related Issue

#8432 

## Types of changes

## Checklist:

- [x] (na) I have updated the documentation accordingly.

### DIFF
--- a/lib/chef/knife/bootstrap/train_connector.rb
+++ b/lib/chef/knife/bootstrap/train_connector.rb
@@ -35,13 +35,27 @@ class Chef
 
         MKTEMP_NIX_COMMAND = "bash -c 'd=$(mktemp -d ${TMPDIR:-/tmp}/chef_XXXXXX); echo $d'".freeze
 
-        def initialize(host_url, default_transport, opts)
-          uri_opts = opts_from_uri(host_url)
-          uri_opts[:backend] ||= @default_transport
-          @transport_type = uri_opts[:backend]
+        def initialize(host_url, default_protocol, opts)
+          @host_url = host_url
+          @default_protocol = default_protocol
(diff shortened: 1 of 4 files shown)


lib/chef/knife/bootstrap/train_connector.rb
```diff
@@ -35,34 +35,27 @@ class TrainConnector
...
-          @transport_type = uri_opts[:backend]
+        def initialize(host_url, default_protocol, opts)
+          @host_url = host_url
+          @default_protocol = default_protocol
+          @opts_in = opts
+        end
 
-          # opts in the URI will override user-provided options
-          @config = transport_config(host_url, opts.merge(uri_opts))
+        def config
```
tas50: Can we get yard comments on these

marcparadise: Updated

PR #8429, opened 2019-04-30
[CHEF-8422] Fix incorrect deprecation warnings

marcparadise:
## Description

Fixed incorrect option mapping where deprecated flags
would have a 'long' taken from the replacement flag. This caused 
mixlib-cli to duplicate the `long` flag - which caused lots of 
assignment confusion, including warning the operator about using flags they never used. 

This also updates the behavior in case of conflicting flags
(for example --ssh-user and --connection-user) to raise an error 
instead of choosing which one to use.This seems less likely to surprise.

Renamed deprecation variables and methods for readability/consistency.

Signed-off-by: Marc A. Paradise <marc.paradise@gmail.com>

## Related Issue

Fixes #8422 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I have run the pre-merge tests locally and they pass.
- [x] I have ~added~ updated to cover my changes.
- [x] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #8442, opened 2019-04-30
Restore bootstrap pre-release support

marcparadise:
## Description

Bootstrap prerelease support was originally removed after
 discussion in community slack, but it turns out to have a little more 
utility than we thought at the time. When enabled, it will cause the 
package to be downloaded from the `current` channel - the latest by default,
but otherwise the specified `--bootstrap-version`.

`Knife#config_source` has been updated to not use a separate hash to store config sources,
and now knows about `:cli_default` value source, which maps to defaulted CLI options that are not in config or used on the CLI by the operator.  This was necessary to prevent `knife bootstrap` from thinking that `--channel` had been set by the operator when it only had a default value.

The CLI flag is now `--channel CHANNEL`; `--prerelease` has been deprecated. If use, will set `channel` to "current". Default is `stable`.

This PR otherwise restores the original behavior (minus the bits that were specific to gem prerelease)  and adds some tests.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -159,9 +159,9 @@ class Chef
 
       option :channel,
         long: "--channel CHANNEL",
-        description: "Install from the given channel. Valid values are 'current' and 'stable'. Default is 'stable'",
+        description: "Install from the given channel.  Valid values are 'stable, 'current', and 'unstable'. Default is 'stable'",
         default: "stable",
-        in: %w{stable current}
+        in: %w{stable current unstable}
 
       # client.rb content via chef-full/bootstrap_context
       option :bootstrap_proxy,
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -86,7 +86,8 @@ class Chef
         short: "-w AUTH-METHOD",
         long: "--winrm-auth-method AUTH-METHOD",
         description: "The WinRM authentication method to use. Valid choices are #{friendly_opt_list(WINRM_AUTH_PROTOCOL_LIST)}.",
-        proc: Proc.new { |protocol| Chef::Config[:knife][:winrm_auth_method] = protocol }
+        proc: Proc.new { |protocol| Chef::Config[:knife][:winrm_auth_method] = protocol },
+        in: WINRM_AUTH_PROTOCOL_LIST
 
       option :winrm_basic_auth_only,
         long: "--winrm-basic-auth-only",
@@ -156,9 +157,11 @@ class Chef
         description: "The version of #{Chef::Dist::PRODUCT} to install.",
         proc: lambda { |v| Chef::Config[:knife][:bootstrap_version] = v }
 
-      option :prerelease,
-        long: "--prerelease",
-        description: "Install from 'current' channel"
+      option :channel,
+        long: "--channel CHANNEL",
+        description: "Install from the given channel. Valid values are 'current' and 'stable'. Default is 'stable'",
+        default: "stable",
+        in: %w{stable current}
 
       # client.rb content via chef-full/bootstrap_context
       option :bootstrap_proxy,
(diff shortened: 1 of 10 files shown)


lamont-granquist:
i think the way it is trying to do this is very, very old though isn't it?

the current channel also won't have any prerelease gem versions in it, so the autodetection of the prerelease gem version is kind of dead code.

and then don't we really want to be passing in the channel option?  i don't see where you're passing in the current channel...

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -190,13 +190,24 @@ def knife_config
...
+                                    Chef::VERSION.split(".").first
+                                  end
+
+            installer_version_string = ["-v", chef_version_string]
 
-          "-v #{chef_version_string}"
+            # If bootstrapping a pre-release version add -p to the installer string
+            if chef_version_string.split(".").length > 3
+              installer_version_string << "-p"
+            end
```
btm: We no longer make "prerelease" gems (x.y.z.pre.0, x.y.z.beta.0) and just use channels now, so this conditional can be dropped.

lib/chef/knife/bootstrap.rb
```diff
@@ -146,6 +146,10 @@ class Bootstrap < Knife
         description: "The version of Chef to install",
         proc: lambda { |v| Chef::Config[:knife][:bootstrap_version] = v }
 
+      option :prerelease,
```
tas50: any reason not to just expose a channel flag. That's a term people would be familiar with and we'd avoid any sort of magic by just passing in the channel name. That would also make it future proof if we added more channels in the future.

btm: Agree with @tas50, let's add a "channel" option. Then instead of adding `-p` to the installer_version string (which is just passed to install.sh),  we would pass `-c CHANNEL`.

The prerelease option could warn it is deprecated and set the channel to "current" which is functionally our prerelease builds now.

test.rb
```diff
@@ -0,0 +1,26 @@
+def rl
```
btm: I don't think you meant to add this file?

lib/chef/knife/bootstrap.rb
marcparadise: Cool, I'd had the same thought and figured few changes are better at this point.

test.rb
marcparadise: Nope, thanks

lib/chef/knife/bootstrap.rb
btm: it's a small change that's overdue. I think it will also let us install from unstable which will let us get the license acceptance builds that we'll start making soon that we're going to intentionally keep from hitting current for a couple days.

marcparadise:
I have updated this branch with the changes above, however because I needed to touch deprecations I rebased it on top of #8429 which also includes deprecation work. way.

lib/chef/knife/bootstrap.rb
```diff
@@ -146,6 +147,12 @@ class Bootstrap < Knife
         description: "The version of Chef to install",
         proc: lambda { |v| Chef::Config[:knife][:bootstrap_version] = v }
 
+      option :channel,
+        long: "--channel CHANNEL",
+        description: "Install from the given channel. Valid values are 'current' and 'stable'. Default is 'stable'",
+        default: "stable",
+        in: %w{stable current}
```
btm: Can we allow `unstable` here too? There are testing scenarios where we would want the ease of grabbing a build from that channel and it is accessible.

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -187,16 +187,32 @@ def knife_config
...
+          installer_version_string = []
+          use_current_channel = (@config[:channel] == "current")
+          installer_version_string << "-p" if use_current_channel
+          version = if knife_config[:bootstrap_version]
+                      knife_config[:bootstrap_version]
+                    else
+                      # We will take the latest current by default,
+                      # if no specific version is given.
+                      if use_current_channel
+                        # Take the latest stable version from the current major release.
```
btm: > # Take the latest stable version from the latest major release.
or highest?

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -187,16 +187,32 @@ def knife_config
...
         def latest_current_chef_version_string
-          chef_version_string = if knife_config[:bootstrap_version]
-                                  knife_config[:bootstrap_version]
-                                else
-                                  Chef::VERSION.split(".").first
-                                end
+          # NOTE: Changes here should also be reflected in Knife::Core::BootstrapContext#latest_current_chef_version_string
+          installer_version_string = []
+          use_current_channel = (@config[:channel] == "current")
+          installer_version_string << "-p" if use_current_channel
```
btm: `-p` is actually back-compat in `install.sh`, we should pass `-c CHANNEL` install the installer_version_string here.

It is easiest to see here: https://omnitruck.chef.io/install.sh

But it comes from here: https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/generator/bourne/scripts/script_cli_parameters.sh

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -158,14 +158,35 @@ def start_chef
...
+        #
+        # Provide the query arguments to a URL that will be appeded to the URL in #msi_url().  This is used
+        # to get the latest version of Chef Infra Client via omnitruck
+        # By default, bootstrap will look for the latest version of the currently-running major release of on stable.
+        #
+        # @return [String] query arguments to append to the download request.
         def latest_current_windows_chef_version_query
+          # NOTE: Changes here should also be reflected in Knife::Core::BootstrapContext#latest_current_chef_version_string
+          use_current_channel = (@config[:channel] == "current")
+          installer_version_string = use_current_channel ? "&prerelease=true" : ""
```
btm: fixing this to pass a channel instead of just prerelease means changing the URL down in `msi_url` to use omnitruck instead of whatever that chef.io redirect is.

https://docs.chef.io/api_omnitruck.html

I don't know how much time we want to spend on that though, because what we really need to do is replace all this with https://github.com/chef/mixlib-install

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -187,16 +187,17 @@ def knife_config
...
-
-          "-v #{chef_version_string}"
+        # Returns the version of Chef to install (as recognized by the Omnitruck API)
+        #
+        # @return [String] download version string
+        def version_to_install
+          return knife_config[:bootstrap_version] if knife_config[:bootstrap_version]
+
+          if @config[:channel] == "stable"
+            Chef::VERSION.split(".").first
```
tas50: Yet another reason we need to use mixlib-install

btm:
failure is opensuseleap42, merging.

PR #8445, opened 2019-05-01
Remove old maintainer gems from the Gemfile

tas50:
These were used by our maintainer rake tasks. These are all gone now so there's no need for these gems.


lamont-granquist:
"

You are trying to install in deployment mode after changing
--
  | your Gemfile. Run `bundle install` elsewhere and add the
  | updated Gemfile.lock to version control.

"

PR #8452, opened 2019-05-03
Add connstant for Chef Server and improve help messaging

tas50:
We talk about the server in messages and we need to brand that appropriately.

In the help:

- Periods after each sentence
- Use the branding better
- Fix some weird capitalization

The final bit is the removal of a link to the RFC docs since those are in the boneyard now.

### DIFF
--- a/lib/chef/application/client.rb
+++ b/lib/chef/application/client.rb
@@ -41,11 +41,11 @@ class Chef::Application::Client < Chef::Application
   option :config_file,
     short: "-c CONFIG",
     long: "--config CONFIG",
-    description: "The configuration file to use"
+    description: "The configuration file to use."
 
   option :config_option,
     long: "--config-option OPTION=VALUE",
-    description: "Override a single configuration option",
+    description: "Override a single configuration option.",
     proc: lambda { |option, existing|
       (existing ||= []) << option
       existing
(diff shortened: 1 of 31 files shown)


tas50:
There's some specs to update here.

lib/chef/application/client.rb
```diff
@@ -238,26 +238,26 @@ class Chef::Application::Client < Chef::Application
...
 
   option :chef_zero_host,
     long: "--chef-zero-host HOST",
-    description: "Host to start chef-zero on"
+    description: "Host to start chef-zero on."
 
   option :chef_zero_port,
     long: "--chef-zero-port PORT",
-    description: "Port (or port range) to start chef-zero on.  Port ranges like 1000,1010 or 8889-9999 will try all given ports until one works."
+    description: "Port (or port range) to start chef-zero on. Port ranges like 1000,1010 or 8889-9999 will try all given ports until one works."
```
mjingle:
```suggestion
    description: "Port (or port range) to start chef-zero on. Port ranges like 1000, 1010 or 8889-9999 will try all given ports until one works."
```

lib/chef/application/knife.rb
```diff
@@ -46,112 +46,112 @@ class Chef::Application::Knife < Chef::Application
...
 
   option :chef_zero_host,
     long: "--chef-zero-host HOST",
-    description: "Host to start chef-zero on"
+    description: "Host to start chef-zero on."
 
   option :chef_zero_port,
     long: "--chef-zero-port PORT",
-    description: "Port (or port range) to start chef-zero on.  Port ranges like 1000,1010 or 8889-9999 will try all given ports until one works."
+    description: "Port (or port range) to start chef-zero on. Port ranges like 1000,1010 or 8889-9999 will try all given ports until one works."
```
mjingle:
```suggestion
    description: "Port (or port range) to start chef-zero on. Port ranges like 1000, 1010, or 8889-9999 will try all given ports until one works."
```

mjingle:
LGTM with the exception of the two suggestions I found.

lib/chef/application/client.rb
tas50: I think we need the original formatting actually

PR #8458, opened 2019-05-03
Switch Travis testing to the new opensuse-42 image

tas50:
The old image isn't being generated anymore and is based off an
unmaintained docker container from the opensuse project.


tas50:
This isn't the fit

PR #8459, opened 2019-05-03
Resolve exceptions when running knife diff

### LINKED ISSUE #4189: ArgumentError: invalid byte sequence in UTF-8 when using knife diff on binary files
### Version:

11.18.12
### Scenario:

Diff cookbooks that may contain binary files
### Steps to Reproduce:

`knife diff cookbooks/a_cookbook_containing_a_binary`

Using OS X `.DS_Store` files seem especially likely to cause the error
### Expected Result:

Some sort of non-crash diff report (e.g., "Binary files X and Y differ")
### Actual Result:

`ERROR: ArgumentError: invalid byte sequence in UTF-8`
### Notes

The problematic code seems to still be [here in `Chef::ChefFS::CommandLine.diff_text`](https://github.com/chef/chef/blob/master/lib/chef/chef_fs/command_line.rb#L263) in master.

`new_value` and `old_value` come in as strings with specific encodings (ASCII-8BIT in the case of binary files). The values then get written out to files, then read back in [here in `Chef::Util::Diff#udiff`](https://github.com/chef/chef/blob/master/lib/chef/util/diff.rb#L92) but without and encoding option. `IO.readlines` then appears to default to UTF-8.

lamont-granquist:
closes #4189

### DIFF
--- a/lib/chef/chef_fs/command_line.rb
+++ b/lib/chef/chef_fs/command_line.rb
@@ -1,6 +1,6 @@
 #
 # Author:: John Keiser (<jkeiser@chef.io>)
-# Copyright:: Copyright 2012-2016, Chef Software Inc.
+# Copyright:: Copyright 2012-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


lib/chef/chef_fs/command_line.rb
```diff
@@ -271,19 +271,18 @@ def diff_text(old_path, new_path, old_value, new_value)
...
-            result
-          ensure
-            old_tempfile.close!
-          end
+          result = Chef::Util::Diff.new.udiff(old_tempfile.path, new_tempfile.path)
+          result = result.gsub(/^--- #{old_tempfile.path}/, "--- #{old_path}")
+          result = result.gsub(/^\+\+\+ #{new_tempfile.path}/, "+++ #{new_path}")
+          result
+        rescue => e
+          "!!! Unable to diff #{old_path} and #{new_path} due to #{e}"
```
tas50: This formatting seems weird since we don't do that anywhere else.

Issue #8376, opened 2019-04-16
Make Chef distributable under a configurable name - wordmarks in outputs


bobchaos:
## Motivation
    As a community member,
    I want to distribute Chef under a different name,
    so that we can all use and contribute to chef OSS.
 
We're attempting to address the various concerns raised by the community following the announcement that all Chef products will become OSS but the Trademark policy prevents redistribution of the software as-is. In short, the code contains multiple  references to Chef wordmarks, including the binary names themselves, that need to be either substituted or removed in a simple way. 

## Context

This problem has multiple facets, at this time the following 3 broad angles have been identified:
- The `chef` gem's binstubs make it impossible to distribute while complying with the trademark policy
  - by removing the binstubs, the chef gem effectively becomes a library and classifies as "internals", which would presumably exclude it from the policy statement 
- The code itself has wordmarks sprinkled in all kinds of log messages and command outputs
  - To comply all outwardly facing messages must have wordmarks removed
- The binaries and folder structure also contain wordmarks (`chef-client`, `/etc/chef`, etc...)
  - These will likely need to have at least some modifications handled at build time

This design proposal aims at addressing the second point, the wordmarks being outputted by `chef`

### cleaning out wordmarks from outputs and logs
this can be solved by creating a new set of constants under the namespace `Chef::Dist`. These new constant would match wordmarks such as "Chef" or "chef-client" in all user-facing messages and outputs as required by https://www.chef.io/trademark-policy/ . An exhaustive list of these changes does not exist at this time.

In log messages and other ruby string objects, wordmarks would be substituted directly through interpolation ie: `"Chef Server" => "#{Chef::Dist::PRODUCT} Server"`.

Other messages may require a bit more work. Windows Event logs come to mind, they are currently present as static files in the code and would require to be made dynamic through ERB or some other solution.

Tests that fail as a result of expecting a wordmark must be modified to ignore the presence of the wordmark, or alternatively to use the distro constant if ignoring the wordmark would break the _intent_ of the test.

The notion of `Chef::Dist` constants could be applied to other projects affected by the recent announcements (Inspec, Automate, Habitat, etc...)

## Downstream Impact

At this point in time the community has identified a number of repositories that will need to be edited in order to make Chef distributable.

| Repo/Product | Patch Needed? | Reason/Example | PR(s)link | Status(es) |
| ------ | ------ | ------ | ----- | ----- |
[...]

PR #8460, opened 2019-05-05
Add distro constants for solo, zero and automate

bobchaos:
Signed-off-by: Marc Chamberland <mchamberland@pbsc.com>
Adding distro constants for Chef Infra Zero, Chef Infra Solo and their CLI counterparts. Also Automate.

## Description
More progress towards making easily Chef easily distributable in compliance with the Trademarks Policy 

## Related Issue
https://github.com/chef/chef/issues/8376

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/data_collector/run_end_message.rb
+++ b/lib/chef/data_collector/run_end_message.rb
@@ -16,6 +16,7 @@
 #
 
 require_relative "message_helpers"
+require_relative "../../dist"
 
 class Chef
   class DataCollector
@@ -61,7 +62,7 @@ class Chef
             "policy_group" => node&.policy_group,
             "start_time" => run_status.start_time.utc.iso8601,
             "end_time" => run_status.end_time.utc.iso8601,
-            "source" => solo_run? ? "chef_solo" : "chef_client",
+            "source" => solo_run? ? Chef::Dist::SOLOEXEC : CHEF::DIST::CLIENT,
             "status" => status,
             "total_resource_count" => all_action_records(action_collection).count,
             "updated_resource_count" => updated_resource_count(action_collection),
--- a/lib/chef/data_collector/run_end_message.rb
+++ b/lib/chef/data_collector/run_end_message.rb
@@ -16,7 +16,6 @@
 #
 
 require_relative "message_helpers"
-require_relative "../../dist"
 
 class Chef
   class DataCollector
@@ -62,7 +61,7 @@ class Chef
             "policy_group" => node&.policy_group,
             "start_time" => run_status.start_time.utc.iso8601,
             "end_time" => run_status.end_time.utc.iso8601,
-            "source" => solo_run? ? Chef::Dist::SOLOEXEC : CHEF::DIST::CLIENT,
+            "source" => solo_run? ? "chef_solo" : "chef_client",
             "status" => status,
             "total_resource_count" => all_action_records(action_collection).count,
             "updated_resource_count" => updated_resource_count(action_collection),
--- a/lib/chef/data_collector/run_start_message.rb
+++ b/lib/chef/data_collector/run_start_message.rb
@@ -16,6 +16,7 @@
 #
 
 require_relative "message_helpers"
+require_relative "../../dist"
 
 class Chef
   class DataCollector
@@ -50,7 +51,7 @@ class Chef
             "node_name" => node&.name || data_collector.node_name,
             "organization_name" => organization,
             "run_id" => run_status&.run_id,
-            "source" => solo_run? ? "chef_solo" : "chef_client",
+            "source" => solo_run? ? Chef::Dist::SOLOEXEC : Chef::Dist::CLIENT,
             "start_time" => run_status.start_time.utc.iso8601,
           }
         end
--- a/lib/chef/data_collector/run_start_message.rb
+++ b/lib/chef/data_collector/run_start_message.rb
@@ -16,7 +16,6 @@
 #
 
 require_relative "message_helpers"
-require_relative "../../dist"
 
 class Chef
   class DataCollector
@@ -51,7 +50,7 @@ class Chef
             "node_name" => node&.name || data_collector.node_name,
             "organization_name" => organization,
             "run_id" => run_status&.run_id,
-            "source" => solo_run? ? Chef::Dist::SOLOEXEC : Chef::Dist::CLIENT,
+            "source" => solo_run? ? "chef_solo" : "chef_client",
             "start_time" => run_status.start_time.utc.iso8601,
           }
         end
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -177,7 +177,7 @@ class Chef
       # client.rb content via bootstrap_context
       option :bootstrap_no_proxy,
         long: "--bootstrap-no-proxy [NO_PROXY_URL|NO_PROXY_IP]",
-        description: "Do not proxy locations for the node being bootstrapped; this option is used internally by Chef.",
+        description: "Do not proxy locations for the node being bootstrapped; this option is used internally by #{Chef::Dist::PRODUCT}.",
         proc: Proc.new { |np| Chef::Config[:knife][:bootstrap_no_proxy] = np }
 
       # client.rb content via bootstrap_context
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -177,7 +177,7 @@ class Chef
       # client.rb content via bootstrap_context
       option :bootstrap_no_proxy,
         long: "--bootstrap-no-proxy [NO_PROXY_URL|NO_PROXY_IP]",
-        description: "Do not proxy locations for the node being bootstrapped; this option is used internally by #{Chef::Dist::PRODUCT}.",
+        description: "Do not proxy locations for the node being bootstrapped",
         proc: Proc.new { |np| Chef::Config[:knife][:bootstrap_no_proxy] = np }
 
       # client.rb content via bootstrap_context
--- a/lib/chef/knife/bootstrap/chef_vault_handler.rb
+++ b/lib/chef/knife/bootstrap/chef_vault_handler.rb
@@ -15,7 +15,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 class Chef
   class Knife
     class Bootstrap < Knife
@@ -141,7 +140,7 @@ class Chef
         def require_chef_vault!
           @require_chef_vault ||=
             begin
-              error_message = "Knife bootstrap requires version 2.6.0 or higher of the chef-vault gem to configure chef vault items"
+              error_message = "Knife bootstrap requires version 2.6.0 or higher of the chef-vault gem to configure vault items"
               require "chef-vault"
               if Gem::Version.new(ChefVault::VERSION) < Gem::Version.new("2.6.0")
                 raise error_message
--- a/lib/chef/knife/exec.rb
+++ b/lib/chef/knife/exec.rb
@@ -18,6 +18,7 @@
 
 require_relative "../knife"
 require_relative "../util/path_helper"
+require_relative "../dist"
 
 class Chef::Knife::Exec < Chef::Knife
 
@@ -26,7 +27,7 @@ class Chef::Knife::Exec < Chef::Knife
   option :exec,
     short: "-E CODE",
     long: "--exec CODE",
-    description: "A string of Chef code to execute."
+    description: "A string of #{Chef::Dist::PRODUCT} code to execute."
 
   option :script_path,
     short: "-p PATH:PATH",
--- a/lib/chef/knife/serve.rb
+++ b/lib/chef/knife/serve.rb
@@ -26,11 +26,11 @@ class Chef
 
       option :repo_mode,
         long: "--repo-mode MODE",
-        description: "Specifies the local repository layout. Values: static (only environments/roles/data_bags/cookbooks), everything (includes nodes/clients/users), hosted_everything (includes acls/groups/etc. for Enterprise/Hosted #{Chef::Dist::PRODUCT}). Default: everything/hosted_everything."
+        description: "Specifies the local repository layout. Values: static (only environments/roles/data_bags/cookbooks), everything (includes nodes/clients/users), hosted_everything (includes acls/groups/etc. for Enterprise/Hosted Chef). Default: everything/hosted_everything."
 
       option :chef_repo_path,
         long: "--chef-repo-path PATH",
-        description: "Overrides the location of chef repo. Default is specified by chef_repo_path in the config."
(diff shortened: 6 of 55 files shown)


lib/chef/data_collector/run_end_message.rb
```diff
@@ -61,7 +62,7 @@ def construct_message(data_collector, status)
             "policy_group" => node&.policy_group,
             "start_time" => run_status.start_time.utc.iso8601,
             "end_time" => run_status.end_time.utc.iso8601,
-            "source" => solo_run? ? "chef_solo" : "chef_client",
+            "source" => solo_run? ? Chef::Dist::SOLOEXEC : CHEF::DIST::CLIENT,
```
bobchaos: any serious implications? Could these be considered as "internals"?

lib/chef/data_collector/run_start_message.rb
```diff
@@ -50,7 +51,7 @@ def construct_message(data_collector)
             "node_name" => node&.name || data_collector.node_name,
             "organization_name" => organization,
             "run_id" => run_status&.run_id,
-            "source" => solo_run? ? "chef_solo" : "chef_client",
+            "source" => solo_run? ? Chef::Dist::SOLOEXEC : Chef::Dist::CLIENT,
```
bobchaos: internals?

lib/chef/data_collector/run_end_message.rb
tas50: Leave this one the way it was

lib/chef/data_collector/run_start_message.rb
tas50: Same with this one

lib/chef/knife/bootstrap.rb
```diff
@@ -175,7 +175,7 @@ def self.friendly_opt_list(opt_array)
       # client.rb content via bootstrap_context
       option :bootstrap_no_proxy,
         long: "--bootstrap-no-proxy [NO_PROXY_URL|NO_PROXY_IP]",
-        description: "Do not proxy locations for the node being bootstrapped; this option is used internally by Chef.",
+        description: "Do not proxy locations for the node being bootstrapped; this option is used internally by #{Chef::Dist::PRODUCT}.",
```
tas50: @lamont / @btm did we mean Chef the company or Chef the product here?

lib/chef/knife/bootstrap/chef_vault_handler.rb
```diff
@@ -141,7 +142,7 @@ def load_chef_bootstrap_vault_item(vault, item)
         def require_chef_vault!
           @require_chef_vault ||=
             begin
-              error_message = "Knife bootstrap requires version 2.6.0 or higher of the chef-vault gem to configure chef vault items"
+              error_message = "Knife bootstrap requires version 2.6.0 or higher of the chef-vault gem to configure vault items"
```
tas50: Geez we should just remove this entirely. That version shipped 4 years ago

lib/chef/knife/client_create.rb
```diff
@@ -35,7 +35,7 @@ class ClientCreate < Knife
       option :admin,
              short: "-a",
              long: "--admin",
-             description: "Open Source Chef Server 11 only. Create the client as an admin.",
+             description: "Open Source #{Chef::Dist::SERVER_PRODUCT} 11 only. Create the client as an admin.",
```
tas50: Looks like I missed removing this feature.

tas50:
@bobchaos I nuked the leftover Chef 11 support in client_create so you'll want to rebase this.

lib/chef/knife/bootstrap/chef_vault_handler.rb
robbkidd: You _know_ someone out there is still running an older version, though.

lib/chef/knife/exec.rb
```diff
@@ -26,7 +27,7 @@ class Chef::Knife::Exec < Chef::Knife
   option :exec,
     short: "-E CODE",
     long: "--exec CODE",
-    description: "A string of Chef code to execute."
+    description: "A string of #{Chef::Dist::PRODUCT} code to execute."
```
robbkidd: I think would be fine to get us moving, so I'm not commenting asking for a change.

But ...

Usages like this demonstrate the muddiness of the water when it comes to things we all call "chef." The "string of code" to be executed here is the same whether it is being executed by the Infra Chef® chef-client® or by CINC's cinc-client or by some other build of this code base. Is this code maybe "recipe code" because the input to this option would look an awful lot like code someone would put in a recipe?

Names are hard.

bobchaos: perhaps the DSL needs a name, distinct from the product or any wordmarks?

lib/chef/policy_builder/policyfile.rb
```diff
@@ -103,7 +103,7 @@ def initialize(node_name, ohai_data, json_attribs, override_runlist, events)
         end
 
         if Chef::Config[:environment] && !Chef::Config[:environment].chomp.empty?
-          raise UnsupportedFeature, "Policyfile does not work with Chef Environments."
+          raise UnsupportedFeature, "Policyfile does not work with #{Chef::Dist::PRODUCT} Environments."
```
robbkidd: Could maybe omit the reference to the distro name with a rewording.

```suggestion
          raise UnsupportedFeature, "Policyfile does not work with an Environment configured."
```

Beyond the scope of minimal changes in this PR, but in the future I'd consider going further in the error message with something like `"Environment configured: #{Chef::Config[:environment]}. Policyfiles and Environments are mutually exclusive. You must use one or the other."`

lib/chef/provider/file.rb
```diff
@@ -230,7 +230,7 @@ def inspect_existing_fs_entry
         elsif file_class.symlink?(new_resource.path) && new_resource.manage_symlink_source.nil?
           logger.warn("File #{path} managed by #{new_resource} is really a symlink (to #{file_class.realpath(new_resource.path)}). Managing the source file instead.")
           logger.warn("Disable this warning by setting `manage_symlink_source true` on the resource")
-          logger.warn("In a future Chef release, 'manage_symlink_source' will not be enabled by default")
+          logger.warn("In a future #{Chef::Dist::PRODUCT} release, 'manage_symlink_source' will not be enabled by default")
```
robbkidd: This reference could probably be removed.

```suggestion
          logger.warn("In a future release, 'manage_symlink_source' will not be enabled by default")
```

lib/chef/resource/dsc_resource.rb
```diff
@@ -16,14 +16,15 @@
...
+require_relative "../dist"
 
 class Chef
   class Resource
     class DscResource < Chef::Resource
       resource_name :dsc_resource
       provides :dsc_resource
 
-      description "The dsc_resource resource allows any DSC resource to be used in a Chef recipe, as well as any custom resources that have been added to your Windows PowerShell environment. Microsoft frequently adds new resources to the DSC resource collection."
+      description "The dsc_resource resource allows any DSC resource to be used in a #{Chef::Dist::PRODUCT} recipe, as well as any custom resources that have been added to your Windows PowerShell environment. Microsoft frequently adds new resources to the DSC resource collection."
```
robbkidd: Could probably omit this reference.

```suggestion
      description "The dsc_resource resource allows any DSC resource to be used in a recipe, as well as any custom resources that have been added to your Windows PowerShell environment. Microsoft frequently adds new resources to the DSC resource collection."
```

lib/chef/resource/cron_d.rb
```diff
@@ -162,7 +163,7 @@ def self.validate_dow(spec)
 
       # warn if someone passes the deprecated cookbook property
       def after_created
-        raise ArgumentError, "The 'cookbook' property for the cron_d resource is no longer supported now that this resource ships in Chef itself." if cookbook
+        raise ArgumentError, "The 'cookbook' property for the cron_d resource is no longer supported now that this resource ships in #{Chef::Dist::PRODUCT} itself." if cookbook
```
robbkidd: Maybe another place we can omit the reference?

```suggestion
        raise ArgumentError, "The 'cookbook' property for the cron_d resource is no longer supported now that cron_d ships as a core resource." if cookbook
```

lib/chef/resource/file.rb
```diff
@@ -73,7 +73,7 @@ class File < Chef::Resource
...
-               description: "Change the behavior of the file resource if it is pointed at a symlink. When this value is set to true, the Chef client will manage the symlink's permissions or will replace the symlink with a normal file if the resource has content. When this value is set to false, Chef will follow the symlink and will manage the permissions and content of symlink's target file. The default behavior is true but emits a warning that the default value will be changed to false in a future version; setting this explicitly to true or false suppresses this warning."
+               description: "Change the behavior of the file resource if it is pointed at a symlink. When this value is set to true, the Chef client will manage the symlink's permissions or will replace the symlink with a normal file if the resource has content. When this value is set to false, #{Chef::Dist::PRODUCT} will follow the symlink and will manage the permissions and content of symlink's target file. The default behavior is true but emits a warning that the default value will be changed to false in a future version; setting this explicitly to true or false suppresses this warning."
```
robbkidd: Missed a "Chef".

```suggestion
               description: "Change the behavior of the file resource if it is pointed at a symlink. When this value is set to true, #{Chef::Dist::PRODUCT} will manage the symlink's permissions or will replace the symlink with a normal file if the resource has content. When this value is set to false, #{Chef::Dist::PRODUCT} will follow the symlink and will manage the permissions and content of symlink's target file. The default behavior is true but emits a warning that the default value will be changed to false in a future version; setting this explicitly to true or false suppresses this warning."
```

lib/chef/data_collector/run_end_message.rb
```diff
@@ -16,6 +16,7 @@
 #
 
 require_relative "message_helpers"
+require_relative "../dist"
```
tas50: ?

lib/chef/data_collector/run_start_message.rb
```diff
@@ -16,6 +16,7 @@
 #
 
 require_relative "message_helpers"
+require_relative "../dist"
```
tas50: ?

lib/chef/knife/bootstrap.rb
tas50: @bobchaos can you make this chef the company

tas50: @bobchaos Can you just take out everything from the ; on. There's no need to mention the internal nature of this.

lib/chef/knife/bootstrap/chef_vault_handler.rb
tas50: Well tell @robbkidd to upgrade

lib/chef/knife/serve.rb
```diff
@@ -25,15 +26,15 @@ class Serve < Knife
 
       option :repo_mode,
         long: "--repo-mode MODE",
-        description: "Specifies the local repository layout. Values: static (only environments/roles/data_bags/cookbooks), everything (includes nodes/clients/users), hosted_everything (includes acls/groups/etc. for Enterprise/Hosted Chef). Default: everything/hosted_everything."
+        description: "Specifies the local repository layout. Values: static (only environments/roles/data_bags/cookbooks), everything (includes nodes/clients/users), hosted_everything (includes acls/groups/etc. for Enterprise/Hosted #{Chef::Dist::PRODUCT}). Default: everything/hosted_everything."
```
tas50: I think this one needs to stay. Hosted Chef is a product and any client could be used with it. Making it Hosted Chef Infra Client doesn't make sense and making it Hosted Cinc Client certainly doesn't make sense.

lib/chef/policy_builder/policyfile.rb
```diff
@@ -103,7 +103,7 @@ def initialize(node_name, ohai_data, json_attribs, override_runlist, events)
         end
 
         if Chef::Config[:environment] && !Chef::Config[:environment].chomp.empty?
-          raise UnsupportedFeature, "Policyfile does not work with Chef Environments."
+          raise UnsupportedFeature, "Policyfile does not work with an Environment Configured."
```
tas50: Configured should be lowercase

lib/chef/resource/cron_d.rb
```diff
@@ -162,7 +163,7 @@ def self.validate_dow(spec)
 
       # warn if someone passes the deprecated cookbook property
       def after_created
-        raise ArgumentError, "The 'cookbook' property for the cron_d resource is no longer supported now that this resource ships in Chef itself." if cookbook
+        raise ArgumentError, "The 'cookbook' property for the cron_d resource is no longer supported now that it ships as a core resource." if cookbook
```
tas50: core resource in #{Chef::Dist::PRODUCT}

lib/chef/knife/serve.rb
bobchaos: good catch, will fix

lib/chef/resource/cron_d.rb
bobchaos: actually I removed that at @robbkidd 's request. Not opposed to adding it back.

lib/chef/data_collector/run_end_message.rb
bobchaos: I don't see an issue, but then again I'm noobsauce o.O What's wrong with this?

lib/chef/knife/client_create.rb
bobchaos: I'ma resolve this, I saw you removed this in a different PR

lib/chef/data_collector/run_end_message.rb
bobchaos: nm, I see it now XD

lib/chef/knife/exec.rb
tas50: I'd leave this one as-is for now.

lib/chef/resource/cron_d.rb
robbkidd: I was thinking that it will be easier if we reduce the number of references. Whether it's "Chef cron_d resource" or "CINC cron_d resource," it's the _same_ resource.

bobchaos:
windows_certificate tests started failing after rebase VS master, best guess is an issue with the pipeline itself (based on bundler errors)

spec/unit/policy_builder/policyfile_spec.rb
```diff
@@ -150,7 +150,7 @@ def initialize_pb
       expect(initialize_pb.temporary_policy?).to be_falsey
     end
 
-    context "chef-solo" do
+    context "#{Chef::Dist::SOLOEXEC}" do
```
lamont-granquist: this probably shouldn't be done since i don't think there's any technical or legal reason to do this -- but it breaks grepping through the file for strings in the rspec output, so there's a developer-workflow reason to keep it with the old string.  if it causes confusion or something later we can revisit this, but i'd rather not do it right now.

bobchaos:
Based on conversations as ChefConf I'ma take a moment to revert changes done to the spec where not immediately relevant

PR #8467, opened 2019-05-06
 Fix the dep bumping rake task to respect the bundler version

tas50:
This prevents it from running with bundler 2.x and bumping the gemfile.lock to bundler 2.x.

This also updates the announce task for the new Infra name.


tas50:
This doesn't work  :(

PR #8470, opened 2019-05-06
Remove resource collision deprecations

lamont-granquist:
Switch to only doing warnings.  Remove the features in the node_map that
we never used.

### DIFF
--- a/lib/chef/deprecated.rb
+++ b/lib/chef/deprecated.rb
@@ -207,9 +207,7 @@ class Chef
       target 23
     end
 
-    class MapCollision < Base
-      target 25
-    end
+    # id 25 was deleted
 
     # id 3694 was deleted
 
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Lamont Granquist (<lamont@chef.io>)
-# Copyright:: Copyright 2014-2018, Chef Software Inc.
+# Copyright:: Copyright 2014-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


lib/chef/deprecated.rb
```diff
@@ -207,9 +207,7 @@ class FreebsdPkgProvider < Base
       target 23
     end
 
-    class MapCollision < Base
-      target 25
-    end
+    # id 15 was deleted
```
tas50: 25 not 15

lib/chef/node_map.rb
```diff
@@ -99,11 +84,7 @@ def set(key, klass, platform: nil, platform_version: nil, platform_family: nil,
                         else
                           klass.superclass.to_s
                         end
```
tas50: Can we just call it collision warning. Versions don’t age well

lib/chef/node_map.rb
```diff
@@ -54,19 +49,13 @@ class NodeMap
...
-    #   override from cookbooks is allowed.
-    # @param __core_override__ [Boolean] Advanced-mode override to add to a key
-    #   even in locked mode.
     # @param chef_version [String] version constraint to match against the running Chef::VERSION
     #
     # @yield [node] Arbitrary node filter as a block which takes a node argument
     #
     # @return [NodeMap] Returns self for possible chaining
     #
-    def set(key, klass, platform: nil, platform_version: nil, platform_family: nil, os: nil, canonical: nil, override: nil, allow_cookbook_override: false, __core_override__: false, chef_version: nil, target_mode: nil, &block) # rubocop:disable Lint/UnderscorePrefixedVariableName
```
tas50: Does this break any existing uses of allow_cookbook_override?

lib/chef/resource_inspector.rb
```diff
@@ -43,7 +43,6 @@ def self.extract_resource(resource, complete = false)
     data[:actions] = resource.allowed_actions
     data[:examples] = resource.examples
     data[:introduced] = resource.introduced
-    data[:preview] = resource.preview_resource
```
tas50: We want to keep preview as a descriptor even if we don't use it to mean anything from the code side. It powers the docs and allows us to arbitrarily mark things as stable or not.

PR #8471, opened 2019-05-06
knife bootstrap should only request license when installing Chef 15

tyler-ball:
## Description
Also updating the transmission of the license acceptance through the
config.rb because that brings it in line with existing patterns
(Test Kitchen).

## Types of changes

## Checklist:


spec/unit/knife/bootstrap_spec.rb
```diff
@@ -40,8 +40,8 @@
   let(:knife) do
     Chef::Log.logger = Logger.new(StringIO.new)
     Chef::Config[:knife][:bootstrap_template] = bootstrap_template unless bootstrap_template.nil?
-    expect(LicenseAcceptance::Acceptor).to receive(:check_and_persist!)
 
+    expect_any_instance_of(Chef::Knife::Bootstrap).to receive(:check_license)
     k = Chef::Knife::Bootstrap.new(bootstrap_cli_options)
```
lamont-granquist: this is likely smell that you shouldn't be doing `chef_license` in the initializer, and should move that to `#run` (and really the ivars that are set in the initializer should be moved to lazy accessors so that the initializer disappears.  then you avoid having the testing against the initializer below, which reads really awkward, and that should eliminate the need to `expect_any_instance_of` and you can just set the expectation directly against `k` here.

lib/chef/knife/bootstrap.rb
```diff
@@ -428,6 +428,20 @@ def initialize(argv = [])
...
       end
 
+      # Determine if we need to accept the Chef Infra license locally in order to successfully bootstrap
+      # the remote node. Remote 'chef-client' run will fail if it is >= 15 and the license is not accepted locally.
+      def check_license
+        Chef::Log.debug("Checking if we need to accept Chef license to bootstrap node")
+        version = config[:bootstrap_version] || Chef::VERSION.split(".").first
+        acceptor = LicenseAcceptance::Acceptor.new(logger: Chef::Log, provided: Chef::Config[:chef_license])
+        if acceptor.license_required?(Chef::Dist::MIXLIB, version)
+          Chef::Log.debug("License acceptance required for chef version: #{version}")
```
tas50: if we're using the dist file already should we use it for "chef" here?

lib/chef/knife/bootstrap.rb
```diff
@@ -428,6 +428,20 @@ def initialize(argv = [])
         )
       end
 
+      # Determine if we need to accept the Chef Infra license locally in order to successfully bootstrap
+      # the remote node. Remote 'chef-client' run will fail if it is >= 15 and the license is not accepted locally.
+      def check_license
+        Chef::Log.debug("Checking if we need to accept Chef license to bootstrap node")
```
tas50: if we're using the dist file already should we use it for "chef" here?

lib/chef/dist.rb
```diff
@@ -15,5 +15,8 @@ class Dist
 
     # product website address
     WEBSITE = "https://chef.io".freeze
+
+    # The key that mixlib-install uses to refer to Chef Infra packages
+    MIXLIB = "chef".freeze
```
tas50: This seems like premature optimization here. We'd only be using it in licensing checks which the community edition wouldn't care about anyways.

tyler-ball: I'm happy to hardcode it back to `"chef"` for now instead of referencing the DIST

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -216,28 +225,26 @@
     end
   end
 
+  # TODO the 'through WinRM' and 'through SSH' tests below are exactly the same - what difference are they supposed
```
tyler-ball: I discovered that these two tests are exactly the same. Anyone know what they are supposed to be testing?

lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -171,8 +175,7 @@ def config_content
         def start_chef
           # If the user doesn't have a client path configure, let bash use the PATH for what it was designed for
           client_path = @chef_config[:chef_client_path] || "#{Chef::Dist::CLIENT}"
-          # We know we can hardcode CHEF_LICENSE because the user cannot get here without accepting the license locally
-          s = "CHEF_LICENSE=accept #{client_path} -j /etc/chef/first-boot.json"
+          s = "#{client_path} -j /etc/chef/first-boot.json"
```
tas50: 👍

PR #8473, opened 2019-05-07
Remove the Chef 11 admin flag from knife client create

tas50:
We've already removed the other support for setting up users on a Chef 11 server. This just further drops support for old Server installs.

### DIFF
--- a/lib/chef/knife/client_create.rb
+++ b/lib/chef/knife/client_create.rb
@@ -30,13 +30,7 @@ class Chef
       option :file,
              short: "-f FILE",
              long: "--file FILE",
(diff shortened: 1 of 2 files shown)


btm:
@tas50 
```
 2) Chef::Knife::ClientCreate run when clientname is passed with -a or --admin should create an admin client
     Failure/Error: expect(client.admin).to be_truthy
     
       expected: truthy value
            got: false
     # ./spec/unit/knife/client_create_spec.rb:146:in `block (5 levels) in <top (required)>'
```

tas50:
@btm I pushed up the fix for the specs

PR #8480, opened 2019-05-07
Target mode code tweaks

lamont-granquist:
The train_or_shell mixin should mixin its dependent mixins and
require them, which removes the need to do that in the calling classes.

Include TrainOrShell correctly in the Universal DSL.

Few more code tweaks ("protected" is largely worthless in ruby and was
always being misused here).

### DIFF
--- a/lib/chef/guard_interpreter/default_guard_interpreter.rb
+++ b/lib/chef/guard_interpreter/default_guard_interpreter.rb
@@ -16,23 +16,18 @@
 # limitations under the License.
 #
 
-require "chef/mixin/shell_out"
 require "chef/mixin/train_or_shell"
+require "chef/exceptions"
 
 class Chef
   class GuardInterpreter
     class DefaultGuardInterpreter
-      include Chef::Mixin::ShellOut
       include Chef::Mixin::TrainOrShell
-
-      protected
 
       def initialize(command, opts)
         @command = command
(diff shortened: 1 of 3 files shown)


lib/chef/guard_interpreter/default_guard_interpreter.rb
```diff
@@ -16,24 +16,19 @@
 # limitations under the License.
 #
 
-require "chef/mixin/shell_out"
```
tas50: Couldn't this be used outside of train_or_shellout?

lamont-granquist: train_or_shell now pulls that require in along with the mixin itself, so it is just redundant here -- and since this file doesn't call shell_out directly and only via train_or_shellout it doesn't add anything to the readability here either.

lamont-granquist: it sort of made sense before because train_or_shell didn't pull in either of them, but that creates a situation where now even though you aren't directly using shell_out or powershell_out you had to know to include the correct one in order to include train_or_shell, but that creates a weird coupling of how you need to structure your requires lines -- only to avoid having powershell_out mixed in when you don't "really" need it.  this is a lot simpler -- just require train_or_shell, include TrainOrShell and use train_or_shell.

PR #8494, opened 2019-05-08
Fix config setting which was returning nil

dheerajd-msys:

## Description

- `@original_config` was nil so return value poses `undefined method `include?' for nil:NilClass (NoMethodError)`

## Related Issue

Fixes https://github.com/chef/knife-ec2/issues/567#issuecomment-490042696

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -375,7 +375,7 @@ class Chef
     #   - :cli_default - came from a declared CLI `option`'s `default` value.
     #   - nil - if they key does not exist
     def config_source(key)
-      return :cli if @original_config.include? key
+      return :cli if config.include? key
       return :config if config_file_settings.key? key
       return :cli_default if default_config.include? key
       nil


lib/chef/knife.rb
```diff
@@ -375,7 +375,7 @@ def merge_configs
     #   - :cli_default - came from a declared CLI `option`'s `default` value.
     #   - nil - if they key does not exist
     def config_source(key)
-      return :cli if @original_config.include? key
+      return :cli if config.include? key
```
marcparadise: This test would always return true. `config` is the merged configuration, while `@original_config` is the pre-merge copy.

marcparadise: We can resolve this by initializing `@original_config` in the constructor. If `merge_config` is never called, it means `config_source` will report `nil`. 

`config_source` is a new addition to the API; bootstrap is the only one [using it](https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L676-L677), and the effect there will be that we don't show deprecation warnings if deprecated config values are used (as they are, in knife-ec2).

marcparadise:
Replaced by #8497

PR #8507, opened 2019-05-08
Use exports compatibile with /bin/sh in the bootstrap script

MarkGibbons:
The bootstrap script is explicitly using /bin/sh.
On some solaris servers the version of /bin/sh exporting environment variables
using "export id=value" fails. The form "id=value export id" does work.
Bootstrap fails when the wrong form is used.

## Description
knife bootstrap fails. The shell command creates uses a form of the export command that isn't supported by old /bin/sh programs.

## Types of changes

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.
- [x ] I have run the pre-merge tests locally and they pass.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


lamont-granquist:
yeah that tickles my braincells as being correct, although i used to see it more split across two lines, but makes sense that one line works as well.

lamont-granquist:
you can cherry-pick + PR this against chef-14 branch for the backport

MarkGibbons:
I'm running it on my local chefdk copy that way.  Works on the sh in solaris 10 so it will probably work on anything created after 1987.

PR #8508, opened 2019-05-09
Convert require to require_relative

lamont-granquist:
This gives a speed boost since rubygems does not have to scan through
every gem in the gemset in order to find the file.


lamont-granquist:
On my mac this gives a 16% speed boost:

```
/opt/chef/bin/chef-client --version  1.46s user 0.57s system 99% cpu 2.050 total
/opt/chef/bin/chef-client --version  1.32s user 0.42s system 99% cpu 1.754 total
```

Conversion was done with this script:

https://gist.github.com/d44421dea302eff20c1cb3e1552e8c19

lib/chef/application/client.rb
```diff
@@ -17,19 +17,19 @@
...
-require "chef/application"
-require "chef/client"
-require "chef/config"
-require "chef/daemon"
-require "chef/log"
-require "chef/config_fetcher"
-require "chef/handler/error_report"
-require "chef/workstation_config_loader"
-require "chef/mixin/shell_out"
+require_relative ""
```
tas50: Does this just do the current dir?

lamont-granquist: WTF?  no, that's a bug.

lamont-granquist: weird that all those mangled lines didn't cause any test failures...

lamont-granquist:
Unfortunately we're still slower than Chef-14 due to appbundling changes:

```
# time /opt/chef/bin/chef-client --version
Chef: 14.12.9
/opt/chef/bin/chef-client --version  0.97s user 0.27s system 98% cpu 1.255 total
```

PR #8509, opened 2019-05-09
Add comments to the Dockerfile explaining how it all works

tas50:
Help outsiders understand how we push things up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # About this Dockerfile:
-# Run without any arguments passed in this Dockerfile will build the latest "stable" channel release of Chef, which
-# is manually defined in the VERSION arg.
+# When run without any arguments passed, this Docker file will build the latest "stable" release of Chef. The version
+# of that release is specified in the VERSION arg in this file, and is automatically updated as described below.
 #
 # Several processes occur using this file which are kicked off by our Expeditor pipeline tooling:
 #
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,24 @@
...
+#
+# When a build makes it through our internal CI system and is promoted to our "unstable" channel Expeditor will
+# trigger a Docker image build of that version and push it to Docker Hub.
+#
+# When tests of an unstable build pass within our CI system it will be promoted to the "current" channel and
+# Expeditor will tag that image as "current" on Docker Hub.
+#
+# When a build is promoted to our "stable" channel .expeditor/update_dockerfile.sh is run to update the version
+# in this file and also tag that image as "latest" and "stable" on Docker Hub. Additionally major and minor tags will be
+# applied so 15.0.260 would be tagged as "latest", "stable", "15" and "15.0", as well as "15.0.260".
```
tduffield: We actually don't apply a "stable" tag, only "latest"

Dockerfile
```diff
@@ -1,8 +1,24 @@
+# About this Dockerfile:
+# Run without any arguments passed in this Dockerfile will build the latest "stable" channel release of Chef, which
+# is manually defined in the VERSION arg.
```
btm: > Run without any arguments passed in this Dockerfile will build the latest "stable" channel release of Chef, which is manually defined in the VERSION arg.

How VERSION is updated is kinda buried. How about:

When run without any arguments passed, this Docker file will build the latest "stable" release of Chef. The version of that release is specified in the VERSION arg in this file, and is automatically updated as described below.

Dockerfile
tas50: Fixed. That was a dokken-ism

Dockerfile
tas50: Fixed

PR #8517, opened 2019-05-10
Backwards compatibility code in Chef::Query::Search fixed (hashify_args)

fledermaus:
## Description

The number of args to check for is 3, not 4, as first 2/5 args to
search are special-cased and not passed to hashify_args.

Array.pop doesn't remove args from the beginning of the list.
Array.shift matches what the comments actually _say_ is being done.

Without these changes search stanzas in recipes fail with 400 Bad Request
when they try to call the REST service backing them.

## Types of changes

## Checklist:

- [x] I couldn't figure out how to run the tests without rake/bundle wanting to install gems system wide and/or failing to complete, but adding the patch to the debian 13.8.7 package and building it resulted in no test failures.

### DIFF
--- a/lib/chef/search/query.rb
+++ b/lib/chef/search/query.rb
@@ -127,9 +127,13 @@ class Chef
         return args.first if args.first.is_a?(Hash)
 
         args_h = Hash.new
-        # If we have 4 arguments, the first is the now-removed sort option, so
+        # If we have 3 arguments, the first is the now-removed sort option, so
         # just ignore it.
-        args.pop(0) if args.length == 4
+        if args.length == 3
+          Chef::Log.warn("Removing deprecated sort arg to search from #{args}\n");
+          args.shift()
+        end
+
         args_h[:start] = args[0] if args[0]
         args_h[:rows] = args[1]
         args_h[:filter_result] = args[2]


lamont-granquist:
this has been broken for over 2 years now and is a deprecated backcompat feature and we aren't ever going to release another chef-13 release.

if you're trying to write compatible cookbooks, those APIs need to get converted to use named parameter hash-style syntax.  that has been supported since Chef 12.1.0 (which is something around 5 years of compatibility) and we will certainly not do any work to support cookbooks running on versions older than that.

it probably makes no sense to fix this in chef-15 because that will not be compatible with chef-13 then since we'll never release a bugfix to chef-13.

lamont-granquist:
this is also most likely a breaking change to anyone who has code that is currently happily running on chef-13/14/15 even though the code doesn't match the stated intent.

fledermaus:
> this is also most likely a breaking change to anyone who has code that is currently happily running on chef-13/14/15

How? If you pass the reduced number of arguments (new API) then the code won't trigger.
If you pass the old argument set, the recipe breaks and your chef client fails to complete its run.

lamont-granquist:
So I think your test needs to be `args.length == 4` since if its 3 and you shift it then args[2] will always be nil and you cant set the `:filter_result`.

Beyond that, you have to assume that people are not using the API as intended but are using it incorrectly.  If someone has figured out that "<start>, <rows>, <filter_result>, <sort>" works (even though sort no longer works) then unshifting that array will remove the <start> arg and they will get "<rows>, <filter_result>, <sort>" and break.

But that argument does not matter.

The bigger point is that chef-13 has seen its last release, and this feature is deprecated, and the new style syntax has been supported for 5 years since 12.1.x.  And patching this won't ever be able to help someone get off of chef-12 to chef-13 at this point.  It should have been deprecated in chef-13 and removed in 14 but that was clearly missed.  That doesn't mean it is now supported and this is a bug, though.

lamont-granquist:
Meant to close this when I opened the cookstyle issue.

PR #8518, opened 2019-05-11
Pin bundler to 1.17.2 which is included in Ruby 2.6

tas50:
This prevents us installing 1.17.3 in addition to 1.17.2. This *should* reduce the overall size of the install and prevent double gems. I'm waiting on an adhoc to see how this works out.


lamont-granquist:
this'll also bandaid-fix the perf issue.

Issue #8491, opened 2019-05-08
Chef 15: Missing --session-timeout bootstrap option 


vsingh-msys:
### Describe the Enhancement:

As per the documentation https://github.com/chef/chef/blob/master/RELEASE_NOTES.md#changed-flags we have `--session-timeout` option for the both `ssh` and `winrm` but only `--winrm-session-timeout` available

### Describe the Need:

1.  `--session-timout` for both winrm & ssh
OR 
2. `--winrm-session-timeout` and `--ssh-session-timeout` for winrm and ssh respectively.
OR
3. Need to update the RELEASE_NOTE.md 

--session-timeout MINUTES | --session-timeout SECONDS | New for ssh, existing for winrm. The unit has changed from MINUTES to SECONDS for consistency with other timeouts.
-- | -- | --

### Can We Help You Implement This?:
Based on the above suggestions need clarification which one should be implemented.

marcparadise:
Let's rename winrm-sessin -timeout to session-timeout, and add winrm-session-timeout to deprecations.  You'll also need to verify that we're passing that option through to train source from get_config(:session_timeout) instead of :winrm_session_timeout

vsingh-msys:
Fixed in https://github.com/chef/chef/pull/8521

PR #8521, opened 2019-05-13
Chef 15: Add --session-timeout bootstrap option for both ssh & winrm

vsingh-msys:
## Description

 - Added bootstrap option `--session-timeout`.
 - Added `--winrm-session-timeout` to deprecations.
 - Default --session-timeout value is 60 secs(same was added for `--winrm-session-timeout`).
 - Modified existing specs related to old `--winrm-session-timeout` key and add specs for `--session-timeout`.
 - To fix `ERROR: TypeError: can't convert String into time interval` parse it to `to_i`. (Float not supported on terminal `ssh user@host -o 'ConnectTimeoit=14.5'` raise error `invalid time value`  )
## Related Issue
https://github.com/chef/chef/issues/8491

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -69,8 +69,7 @@ class Chef
       option :session_timeout,
         long: "--session-timeout SECONDS",
         description: "The number of seconds to wait for each connection operation to be acknowledged while running bootstrap.",
-        proc: Proc.new { |protocol| Chef::Config[:knife][:session_timeout] = protocol },
-        default: 60
+        proc: Proc.new { |protocol| Chef::Config[:knife][:session_timeout] = protocol }
 
       # WinRM Authentication
       option :winrm_ssl_peer_fingerprint,
@@ -839,7 +838,7 @@ class Chef
         return opts if connection_protocol == "winrm"
         opts[:non_interactive] = true # Prevent password prompts from underlying net/ssh
         opts[:forward_agent] = (config_value(:ssh_forward_agent) === true)
-        opts[:connection_timeout] = config_value(:session_timeout).to_i
+        opts[:connection_timeout] = config_value(:session_timeout)&.to_i || 60
         opts
       end
 
@@ -938,7 +937,7 @@ class Chef
           opts[:ca_trust_file] = config_value(:ca_trust_file)
         end
 
-        opts[:operation_timeout] = config_value(:session_timeout).to_i
+        opts[:operation_timeout] = config_value(:session_timeout)&.to_i || 60
 
         opts
       end
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -66,6 +66,11 @@ class Chef
         long: "--max-wait SECONDS",
         description: "The maximum time to wait for the initial connection to be established."
 
+      option :session_timeout,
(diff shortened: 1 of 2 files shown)


vsingh-msys:
Session timeout on `train-core` side:

- There is no need to add session timeout related configuration in `train-core` side. this option is already available as `operation_timeout` & `connection_timeout` (https://github.com/inspec/train/blob/827b73a66ae10b7c54de9fef9a228862f425bb43/lib/train/transports/ssh.rb#L154) for `winrm` & `ssh` respectively.

- By default `connection_timeout` value is 15 secs (https://github.com/inspec/train/blob/827b73a66ae10b7c54de9fef9a228862f425bb43/lib/train/transports/ssh.rb#L55) but previously we passed `--winrm-session-timeout` 60 secs so now it passed for ssh as well.

lib/chef/knife/bootstrap.rb
```diff
@@ -836,6 +838,7 @@ def ssh_opts
         return opts if connection_protocol == "winrm"
         opts[:non_interactive] = true # Prevent password prompts from underlying net/ssh
         opts[:forward_agent] = (config_value(:ssh_forward_agent) === true)
+        opts[:connection_timeout] = config_value(:session_timeout)&.to_i || 60
```
btm: Why the `&`?

vsingh-msys: @btm Actually if config_value(:session_timeout) is nil `to_i` return `0` 
so 
`config_value(:session_timeout).to_i || 60` evaluates to `0` 
To avoid this 
`&` used to first check nil and return nil for for expression `config_value(:session_timeout)&.to_i`
so 
`config_value(:session_timeout)&.to_i || 60` evaluates to 60

tas50: Any reason not to just make the default 60 seconds and avoid this assignment here?

vsingh-msys: Hey @tas50 As the default `connection_timeout` value is 15 secs (https://github.com/inspec/train/blob/827b73a66ae10b7c54de9fef9a228862f425bb43/lib/train/transports/ssh.rb#L55) 

And the only reason to assign a default value to 60 secs due to previously we passed for `--winrm-session-timeout` 60 secs.

Any suggestion would be helpful?

Thanks

tas50: yeah we should just set it to 60 for knife, but we can do that via CLI default unless I'm missing something

vsingh-msys: @tas50 Will check and update accordingly.

vsingh-msys:
Okay @mp Will update and revert you back accordingly thanks

vsingh-msys:
@marcparadise & @tas50 required changes have been made. Could you please review and let's know further if any issue/concern thanks!

marcparadise:
This looks good, but I don't want to leave the time unit conversion - and possible problems that might come out of it - unaddressed. 

@btm suggested adding a validation check that gives a warning if the session-timeout value is less than 30 seconds.

 It's a safe bet that someone putting in a low value is thinking in terms of minutes and not seconds, and this way if someone habitually uses `--winrm-session-timeout 1` we'll at least have given them a hint as to the cause of any issues.

PR #8524, opened 2019-05-13
Use new Net:SSH host key verify values

### LINKED ISSUE #8482: Bootstrap: If the SSH fingerprint is unknown offer to continue anyways
If we encounter an unknown ssh key we should offer to continue. Just about everyone is going to hit this and then have to start over with the flag to ignore the unknown keys.

```
Creating new client for tim_test
Creating new node for tim_test
Connecting to ec2-34-221-181-2.us-west-2.compute.amazonaws.com
ERROR: Train::Transports::SSHFailed: SSH command failed (fingerprint SHA256:ll/BKqLRTOgQXaFCFUaHx5/JxDYvOpAbSS80NVTek8Q is unknown for "ec2-34-221-181-2.us-west-2.compute.amazonaws.com,34.221.181.2")
```

btm:
Fixes #8482 by prompting to continue if the host key is not recognized.

This attempts to tell Net:SSH to accept and write the key but it is only
temporarily accepting.

This changes the parameters of --ssh-verify-host-key (which is new) to
take the Net::SSH values, which get passed through by train. This allows
the user more options than only verifying or not.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -88,7 +88,7 @@ class Chef
       option :winrm_auth_method,
         short: "-w AUTH-METHOD",
         long: "--winrm-auth-method AUTH-METHOD",
-        description: "The WinRM authentication method to use. Valid choices are #{friendly_opt_list(WINRM_AUTH_PROTOCOL_LIST)}.",
+        description: "The WinRM authentication method to use.",
         proc: Proc.new { |protocol| Chef::Config[:knife][:winrm_auth_method] = protocol },
         in: WINRM_AUTH_PROTOCOL_LIST
 
@@ -146,9 +146,9 @@ class Chef
         description: "The SSH identity file used for authentication."
 
       option :ssh_verify_host_key,
-        long: "--[no-]ssh-verify-host-key",
-        description: "Verify host key, enabled by default.",
-        boolean: true
+        long: "--ssh-verify-host-key VALUE",
+        description: "Verify host key. Default is 'always'.",
+        in: %w{always accept_new accept_new_or_local_tunnel never}
 
       #
       # bootstrap options
@@ -162,7 +162,7 @@ class Chef
 
       option :channel,
         long: "--channel CHANNEL",
-        description: "Install from the given channel.  Valid values are 'stable, 'current', and 'unstable'. Default is 'stable'",
+        description: "Install from the given channel. Default is 'stable'.",
         default: "stable",
         in: %w{stable current unstable}
 
@@ -549,7 +549,28 @@ class Chef
 
         $stdout.sync = true
         register_client
-        connect!
+        begin
+          connect!
+        rescue Train::Transports::SSHFailed => e
+          if e.message =~ /fingerprint (\S+) is unknown for "(.+)"/
+            fingerprint = $1
+            hostname,ip = $2.split(',')
+            puts "The authenticity of host '#{hostname} (#{ip})' can't be established."
+            # TODO: convert the SHA256 base64 value to hex with colons
+            # 'ssh' example output:
+            # RSA key fingerprint is e5:cb:c0:e2:21:3b:12:52:f8:ce:cb:00:24:e2:0c:92.
+            # ECDSA key fingerprint is 5d:67:61:08:a9:d7:01:fd:5e:ae:7e:09:40:ef:c0:3c.
+            puts "fingerprint is #{fingerprint}."
+            ui.confirm("Are you sure you want to continue connecting")
+            # FIXME: this should save the key to known_hosts but doesn't appear to be
+            config[:ssh_verify_host_key] = :accept_new
+            connection_opts(reset: true)
+            retry
+          end
+
+          raise e
+        end
+
         unless client_builder.client_path.nil?
           bootstrap_context.client_pem = client_builder.client_path
         end
@@ -781,8 +802,8 @@ class Chef
 
       # @return a configuration hash suitable for connecting to the remote
       # host via train
-      def connection_opts
-        return @connection_opts unless @connection_opts.nil?
+      def connection_opts(reset: false)
+        return @connection_opts unless @connection_opts.nil? || reset == true
         @connection_opts = {}
         @connection_opts.merge! base_opts
         @connection_opts.merge! host_verify_opts
@@ -824,8 +845,7 @@ class Chef
           { self_signed: config_value(:winrm_no_verify_cert) === true }
         elsif ssh?
           # Fall back to the old knife config key name for back compat.
-          { verify_host_key: config_value(:ssh_verify_host_key,
-                                          :host_key_verify, true) === true }
+          { verify_host_key: config_value(:ssh_verify_host_key, :host_key_verify, "always") }
         else
           {}
         end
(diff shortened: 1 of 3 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -88,7 +88,7 @@ def self.friendly_opt_list(opt_array)
       option :winrm_auth_method,
         short: "-w AUTH-METHOD",
         long: "--winrm-auth-method AUTH-METHOD",
-        description: "The WinRM authentication method to use. Valid choices are #{friendly_opt_list(WINRM_AUTH_PROTOCOL_LIST)}.",
```
marcparadise: Why remove the valid values? The only other way to discover them is to try them and see if you get an error.

tas50: I get why we're doing this here, but we should get a todo to fix the output of "in" cleaned up since it's not very user friendly.

tas50: Also you might be able to kill that friendly_opt_list method now. I'm not sure if I was using that anywhere else

tas50:
@btm can you add the flag changes to the release notes

lib/chef/knife/bootstrap.rb
tas50: Using in prints them to help as well. It just does so in a kind of confusing way: 

```
-w AUTH-METHOD,                  The WinRM authentication method to use. Valid choices are 'plaintext', 'kerberos', 'ssl' and 'negotiate'. (included in ['plaintext', 'kerberos', 'ssl', 'negotiate'])
        --winrm-auth-method
```

lib/chef/knife/bootstrap.rb
```diff
@@ -549,7 +549,28 @@ def run
...
+            hostname,ip = $2.split(',')
+            puts "The authenticity of host '#{hostname} (#{ip})' can't be established."
+            # TODO: convert the SHA256 base64 value to hex with colons
+            # 'ssh' example output:
+            # RSA key fingerprint is e5:cb:c0:e2:21:3b:12:52:f8:ce:cb:00:24:e2:0c:92.
+            # ECDSA key fingerprint is 5d:67:61:08:a9:d7:01:fd:5e:ae:7e:09:40:ef:c0:3c.
+            puts "fingerprint is #{fingerprint}."
+            ui.confirm("Are you sure you want to continue connecting")
+            # FIXME: this should save the key to known_hosts but doesn't appear to be
+            config[:ssh_verify_host_key] = :accept_new
```
marcparadise: train doesn't know about this value - it only expects a bool [1], so we'll need to modify there as well. 

That's probably also why `accept_new` isn't causing `known_hosts` to be updated.  

[1] https://github.com/inspec/train/blob/master/lib/train/transports/ssh.rb#L199

lib/chef/knife/bootstrap.rb
```diff
@@ -549,7 +549,28 @@ def run
...
+        rescue Train::Transports::SSHFailed => e
+          if e.message =~ /fingerprint (\S+) is unknown for "(.+)"/
+            fingerprint = $1
+            hostname,ip = $2.split(',')
+            puts "The authenticity of host '#{hostname} (#{ip})' can't be established."
+            # TODO: convert the SHA256 base64 value to hex with colons
+            # 'ssh' example output:
+            # RSA key fingerprint is e5:cb:c0:e2:21:3b:12:52:f8:ce:cb:00:24:e2:0c:92.
+            # ECDSA key fingerprint is 5d:67:61:08:a9:d7:01:fd:5e:ae:7e:09:40:ef:c0:3c.
+            puts "fingerprint is #{fingerprint}."
```
marcparadise: recommend: #"#{hostname}'s fingerprint is: #{fingerprint}"

lib/chef/knife/bootstrap.rb
btm: Yeah, we should fix that output, but it was weird having them printed twice.

lib/chef/knife/bootstrap.rb
```diff
@@ -549,7 +549,28 @@ def run
 
         $stdout.sync = true
         register_client
-        connect!
+        begin
+          connect!
```
marcparadise: Can we rescue this rescue block of  `connect!` itself?  Having the connection error handling in one place seems more manageable than splitting it across two.

lib/chef/knife/bootstrap.rb
```diff
@@ -549,7 +549,28 @@ def run
...
+          if e.message =~ /fingerprint (\S+) is unknown for "(.+)"/
+            fingerprint = $1
+            hostname,ip = $2.split(',')
+            puts "The authenticity of host '#{hostname} (#{ip})' can't be established."
+            # TODO: convert the SHA256 base64 value to hex with colons
+            # 'ssh' example output:
+            # RSA key fingerprint is e5:cb:c0:e2:21:3b:12:52:f8:ce:cb:00:24:e2:0c:92.
+            # ECDSA key fingerprint is 5d:67:61:08:a9:d7:01:fd:5e:ae:7e:09:40:ef:c0:3c.
+            puts "fingerprint is #{fingerprint}."
+            ui.confirm("Are you sure you want to continue connecting")
```
marcparadise: Does this raise if they pick no? If not, we need to check the return value before proceeding.

lib/chef/knife/bootstrap.rb
marcparadise: :+1:

lib/chef/knife/bootstrap.rb
btm: `#confirm` will exit 3 if they say no.

lib/chef/knife/bootstrap.rb
btm: That method converts specific values but passes anything else through:

```
2.4.1 :001 > given = :accept_new
 => :accept_new 
2.4.1 :002 >         {
2.4.1 :003 >               'true' => true,
2.4.1 :004 >               'false' => false,
2.4.1 :005 >               nil => false,
2.4.1 :006 >             }.fetch(given, given)
 => :accept_new 
```

btm: you can see it working here too, notice after the SNIP in the first and second connections:

```
DEBUG: [SSH] opening connection to ubuntu@ec2-3-85-23-14.compute-1.amazonaws.com<{SNIP, :verify_host_key=>:always}>
The authenticity of host 'ec2-3-85-23-14.compute-1.amazonaws.com (3.85.23.14)' can't be established.
fingerprint is SHA256:dVvJdTZrMcAr+YfN0NV4tnt5Sj5zjdJLhfzEoZvkd90.
Are you sure you want to continue connecting? (Y/N) y

Connecting to ec2-3-85-23-14.compute-1.amazonaws.com
DEBUG: [SSH] Using Agent keys as no password or key file have been specified
DEBUG: [SSH] ubuntu@ec2-3-85-23-14.compute-1.amazonaws.com<{SNIP, :verify_host_key=>:accept_new}> (cmd.exe /c ver)
```

lib/chef/knife/bootstrap.rb
btm: I was trying to match the output of `ssh` which is two lines, the first line with the hostname is above the comments here. the second one is in the example comments, but to do it we need to convert the key from base64 at least in the case of an RSA SHA256 fingerprint, and I kinda punted on that.

lib/chef/knife/bootstrap.rb
btm: https://github.com/chef/mixlib-cli/pull/59

lib/chef/knife/bootstrap.rb
marcparadise: Cool, thanks - that's what I get for skimming

lib/chef/knife/bootstrap.rb
btm: done

PR #8554, opened 2019-05-16
Add license CLI options to chef-apply command

tas50:
You could accept the license interactively but you could not do it via
the CLI flag


tas50:
@tyler-ball done

Issue #8433, opened 2019-04-30
knife  suggest to use -VVV option to collect the ouput which is not giving full Trackback of the command output.


Vasu1105:
## Description
While running knife bootstrap command its giving suggestion to use -VVV option to collect the output of the command which is not giving the full Trackback. or stacktrace. If I am not wrong here we should tell to use -VV option instead of -VVV

## Chef Version
Chef Infra: 15.0.238
## Replication Case
We can run any knife command which throws error to reproduce this.
https://github.com/chef/chef/blob/66c0fdeb65c19d267bd501550e60cd16d7bb4901/lib/chef/knife.rb#L487
## Client Output

```

```

## Stacktrace

command output when using -VVV option
```
WARNING: --winrm-user USER is deprecated. Use --connection-user USERNAME instead.
WARNING: --winrm-password is deprecated. Use --connection-password PASSWORD instead.
WARNING: --winrm-port is deprecated. Use --connection-port PORT instead.
WARNING: --winrm-authentication-protocol PROTOCOL is deprecated. Use --winrm-auth-method AUTH-METHOD instead.
TRACE: Chef::HTTP calling Chef::HTTP::JSONInput#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::JSONOutput#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::CookieManager#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::Decompressor#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::Authenticator#handle_request
TRACE: Signing the request as vasundharaj
TRACE: Chef::HTTP calling Chef::HTTP::RemoteRequestID#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::APIVersions#handle_request
TRACE: Chef::HTTP calling Chef::HTTP::ValidateContentLength#handle_request
TRACE: Initiating GET to https://api.chef.io/organizations/vasuj/nodes/vj-win2012r2
TRACE: ---- HTTP Request Header Data: ----
TRACE: Accept: application/json
TRACE: Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
TRACE: X-Ops-Server-API-Version: 1
TRACE: X-OPS-SIGN: algorithm=sha1;version=1.1;
TRACE: X-OPS-USERID: vasundharaj
TRACE: X-OPS-TIMESTAMP: 2019-04-30T07:24:21Z
[...]

PR #8569, opened 2019-05-20
Chef- 15: Change knife suggestion to use -VV instead of -VVV

vsingh-msys:
## Description
During bootstrap seems logged info to use `-VVV` trace level instead of `-VV`. `-VV` does not provide raised exception trace so it might not use for the community to debug the code to in first go so `-VV` is useful before logging the issue.

https://github.com/chef/chef/blob/66c0fdeb65c19d267bd501550e60cd16d7bb4901/lib/chef/knife.rb#L487

Well, we can document it a cleaner way. something similar to: 
 - -V: log level info
- -VV: log level debug, allow exceptions to raise
- -VVV: log level trace, do not allow exceptions to raise

instead
    -V, --verbose                    More verbose output. Use twice for max verbosity.

in `knife bootstrap --help` command?

## Related Issue
Fixed https://github.com/chef/chef/issues/8433

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -498,7 +498,7 @@ class Chef
       when NameError, NoMethodError
         ui.error "knife encountered an unexpected error"
         ui.info  "This may be a bug in the '#{self.class.common_name}' knife command or plugin"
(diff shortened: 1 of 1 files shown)


lamont-granquist:
closing this one to get it off the list, it needs a different fix from the issue.

Issue #8534, opened 2019-05-14
[CHEF 15] Bootstrapping a server now causes 'No such file or directory' all of a sudden.


breisig:
I have upgraded from Chef client 14.xx to Chef Infra Client: 15.0.293 and tried bootstrapping a server as normal [had to replace -x on the command line with --connection-user and  --connection-password] but now I keep getting the follow error message when attempting to boostrap a server.

Performing legacy client registration with the validation key at /home/test/.chef/mychefserver.pem..."
Delete your validation key in order to use your user credentials for client registration instead.
Connecting to test.mytestsite.com
ERROR: Errno::ENOENT: No such file or directory @ rb_sysopen - /tmp/d20190514-10948-1pq02vu/test.mytestsite.com.pem

marcparadise:
Thanks for reporting this @breisig. Could you let us know what your workstation OS/version is, and the OS/version of the node you're bootstrapping?  It would also help to know which cloud provider for the target node, or if the target node is internal/local.

breisig:
@marcparadise  Both the workstation and bootstrapped server is using the latest CentOS 7.6 64bit. The target server is just an internal VM running on a fresh CentOS 7.6 minimal image [with all patches].

marcparadise:
Thanks, I'll see if I can get the same to occur.  

Would you be able to include the redacted CLI  and its output when run with -VV? (Note two Vs and not three - two will not give the stack trace I'm hoping for - ref https://github.com/chef/chef/issues/8433 for the curious...)

dheerajd-msys:
@breisig Until you confirm with the complete `stacktrace` I would like to say that I was getting the similar error from Windows workstation with `chef v15` here https://github.com/chef/knife-ec2/issues/575. In my `knife.rb` there was an entry for `validation_key` which is old process for client registration. Deleting the entry and file from the path resolved my error. Could you please cross-verify if you are doing the same.

Thanks,

PR #8574, opened 2019-05-21
Chef 15: Fix order of connection before registering node

dheerajd-msys:

Correct the order of `connection` before `registration`

## Description
Updates order of `connection`  before `registration` for any node, Ensures the `connection` object to be live.

## Related Issue
Fixes https://github.com/chef/knife-ec2/issues/579

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -543,8 +543,8 @@ class Chef
         warn_on_short_session_timeout
 
         $stdout.sync = true
-        register_client
         connect!
+        register_client
 
         content = render_template
         bootstrap_path = upload_bootstrap(content)
--- a/spec/unit/knife/bootstrap_spec.rb
+++ b/spec/unit/knife/bootstrap_spec.rb
@@ -1664,8 +1664,8 @@ describe Chef::Knife::Bootstrap do
       expect(knife).to receive(:validate_policy_options!).ordered
       expect(knife).to receive(:winrm_warn_no_ssl_verification).ordered
       expect(knife).to receive(:warn_on_short_session_timeout).ordered
-      expect(knife).to receive(:register_client).ordered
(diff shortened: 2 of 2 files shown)


Vasu1105:
@dheerajd-msys we register the client first and then go for connection. So this will change the current behavior.

Vasu1105:
@dheerajd-msys I am working on with @btm in fixing the #8576. So may be we can close this PR.

dheerajd-msys:
@Vasu1105 I guess there is a mismatch, connection to the machine should be established before registration of any machine/node to the chef server.

Vasu1105:
@dheerajd-msys may be I am not sure. I have look around the code. @btm  any thoughts on this?

Vasu1105:
What I see here in new bootstrap we are using connection to identify the platform so that we can use the right bootstrap_context. Due to which just to get that connection we have to call connection first but I am not sure how much this is in relation with bootstrap process.

btm:
@dheerajd-msys is there a particular error this fixes? can you post the output here? there are a couple issues going around so we should be specific about which ones this intends to fix.

I made https://github.com/chef/chef/pull/8567 to fix https://github.com/chef/chef/issues/8534 but that caused https://github.com/chef/chef/issues/8576. I think that's what you're trying to fix here.

btm:
The problem was that I was fixing the 'validator' bootstrap path and broke the 'validatorless' bootstrap path, which @marcparadise had been testing. So we need to test both.

btm:
This tested working for me locally with both validatorless and validator bootstrap paths. :+1:

dheerajd-msys:
Yes @btm I was trying to fix https://github.com/chef/knife-ec2/issues/579 which is same as https://github.com/chef/chef/issues/8576. And I have verified that this change works for 'validator as well as 'validator-less'. 👍

PR #8575, opened 2019-05-21
Add screensaver resource

natewalck:
## Description
After talking with @tas50 and @mikedodge04, I am cutting this PR to spur discussion. This creates a Screensaver resource which can be used to manage screensaver settings. Currently it only supports macOS, but I will add Windows and Gnome support once we hash this out and know this is the direction we want to go.

## Types of changes

## Checklist:

> screensaver "macOS Screensaver Test Manage" do
>   idle_time 60
>   ask_for_password true
>   password_delay 5
>   organization "Uber"
>   prefix "com.uber"
> end
> 
> screensaver "macOS Screensaver Test Unmanage" do
>   action :unmanage
>   organization "Uber"
>   prefix "com.uber"
> end
> 
> 
> screensaver "Windows Screensaver Test Manage" do
>   idle_time 60
>   ask_for_password true
>   password_delay 5
> end
> 
> screensaver "Windows Screensaver Test Unmanage" do
>   action :unmanage
> end


lib/chef/resource/screensaver.rb
```diff
@@ -0,0 +1,116 @@
...
+
+class Chef
+  class Resource
+    class Screensaver < Chef::Resource
+      resource_name :screensaver
+      provides(:screensaver) { true }
+      default_action :manage
+
+      description "Use the screensaver resource to configure screensaver settings."
+      introduced "14.0"
```
tas50: 15.1 probably

lib/chef/resource/screensaver.rb
```diff
@@ -0,0 +1,116 @@
...
+#
+
+require_relative "../resource"
+
+class Chef
+  class Resource
+    class Screensaver < Chef::Resource
+      resource_name :screensaver
+      provides(:screensaver) { true }
+      default_action :manage
```
tas50: you can leave this out. The first one is is the default in Chef 12.5+

natewalck: Done!

lib/chef/resource/screensaver.rb
natewalck: Done!

lib/chef/resource/screensaver.rb
```diff
@@ -0,0 +1,175 @@
...
+          screensaver_payload = {
+            'PayloadType' => 'com.apple.screensaver',
+            'PayloadVersion' => 1,
+            'PayloadIdentifier' => profile_identifier,
+            'PayloadUUID' => '3B2AD6A9-F99E-4813-980A-4147617B2E75',
+            'PayloadEnabled' => true,
+            'PayloadDisplayName' => 'Screensaver'
+          }
+
+          screensaver_payload['idleTime'] = new_resource.idle_time if new_resource.idle_time
```
erikng: instead of doing this, should any value be allowed, in case Apple adds new keys later?

lib/chef/resource/screensaver.rb
```diff
@@ -0,0 +1,175 @@
...
+# limitations under the License.
+#
+
+require_relative "../resource"
+
+class Chef
+  class Resource
+    class Screensaver < Chef::Resource
+      resource_name :screensaver
+      provides(:screensaver) { true }
```
lamont-granquist: this should be the `MacosScreensaver` and `:macos_screensaver` resource that provides `:screensaver` on platform `:mac_os_x` (IIRC).

when you get to the point of code-sharing between multiple different screensaver implementations that need to share the same API you'll need to use a mixin for the common properties, and you'll
wind up needing this:

https://lamont-granquist.gitlab.io/chef/base-modules/

lib/chef/resource/screensaver.rb
natewalck: We would need to add a property to the resource to consume the new key, which means touching the code anyways. For a core resource, I think it is better to keep it tightly scoped and update as needed.

tas50:
@natewalck Anything I can do to move this along?

tas50:
@natewalck Ping. We'd love to get this into Chef 16 if you have the time.

jaymzh:
I pinged Nate on messenger...

tas50:
@natewalck I'm going to close this one out at this point since we have't heard back.

Issue #8580, opened 2019-05-23
Ability to enable/disable Chocolatey features


gep13:
### Describe the Enhancement:
In my Chef recipes, I want to be able to enable/disable Chocolatey features, in the same way that I can set Chocolatey configuration values and sources.

### Describe the Need:
Anyone currently making use of Chocolatey and Chef would benefit from being able to also configure features within Chocolatey.

This would also bring better feature parity for Chocolatey in Chef compared to some of the other Configuration Managers out there:

![image](https://user-images.githubusercontent.com/1271146/58280048-aaee9c80-7d54-11e9-92b0-ce59aa965f3c.png)

### Current Alternative
Currently possible to execute Chocolatey commands directly in recipe, but not as "nice" as having a built in resource for it.

### Can We Help You Implement This?:

This would be added as a new resource, similar to `chocolatey_config` and `chocolatey_source`

I am here at ChefConf this week, and I have been bugging @ta50 with some questions about this, and I believe I will have a PR up soon which will be a first stab at implementing this.

PR #8581, opened 2019-05-23
Add new chocolatey_feature resource for managing features in Chocolatey

gep13:
## Description

Add the ability to enable and disable a named Chocolatey feature by
creating a new Chef resource named chocolatey_feature.  This accepts a
single property either from feature_name or directly from name attribute
then, based on the defined action, will either enable or disable that
feature.  Resource inspects the current state of the chocolatey.config
in order to decide whether a change is required or not.

## Related Issue
https://github.com/chef/chef/issues/8580

## Types of changes

## Checklist:


gep13:
Couple of questions that I have about this PR...

* What should I put for the `introduced` property in the resource?
* Is there a way to "hide" the `feature_state` property within the resource?  This is only used for figuring out the current state of the feature from the configuration file, but it is not intended to be set by a user within a recipe.
* Is there documentation that needs to be updated in line with this new resource?  If so, can you point me at it, as I wasn't sure what needs to be updated where.

kitforbes:
Hey @gep13 . Coincidently, I was working on the same resource. I based my implementation off of the previous `chocolatey_config` from [here](https://github.com/chef-boneyard/chocolatey_config/blob/master/resources/default.rb). My **very hacky** attempt can be found [here](https://github.com/kitforbes/chef-repo/blob/master/cookbooks/windows_baseline/resources/chocolatey_feature.rb). Hopefully someone with more Chef knowledge will come along and correct my approach, but until then...

In my resource, I set the `feature_state` (`enabled` in my version) as not being `required: false`. Then, within each action, I forcibly overwrite `new_resource.feature_state` to suit my needs. This probably wouldn't be allowed in core Chef, but works fine for my own personal cookbook.

It can then be used as below:

```ruby
chocolatey_feature 'checksumFiles' do
  action :enable
end
```

I also discovered that you can get the current `action` from `new_resource.action` which returns an array of actions. There might be a way to use that to achieve this in a better way.

lamont-granquist:
> I forcibly overwrite `new_resource.feature_state` to suit my needs

Yeah you shouldn't ever overwrite anything on the new_resource, that works up until things get crazy fast.

Mutable state should always be hung off of the provider/action_class.

There are exceptions in core chef where we've had to do this for reasons related to reporting, but those are all working around the fact that we need an after_resource state which is read similarly to current_resource for 'diffing' purposes, and those cases all need to be handled very carefully.

lib/chef/resource/chocolatey_feature.rb
```diff
@@ -0,0 +1,80 @@
...
+        current_value_does_not_exist! if current_state.nil?
+
+        feature_name feature_name
+        feature_state current_state == "true"
+      end
+
+      # @param [String] id the feature name
+      # @return [String] the element's value field
+      def fetch_feature_element(name)
+        require "rexml/document" unless defined?(REXML::Document)
```
lamont-granquist: 👍 for using rexml over nokogiri

lamont-granquist:
this seems fine to me.

gep13:
@lamont-granquist what is the recommended way to "hide" the property on the resource so that it can't be set within a recipe/cookbook.  It is only used internally within the resource to verify the current state.

gep13:
I will try to get the build failures fixed up over the weekend, so that all the checks pass.

lib/chef/resource/chocolatey_feature.rb
gep13: @lamont-granquist can't claim that I knew what I was doing here 😄 I unashamedly borrowed the code from the `chocolatey_config` and `chocolatey_source` and just applied what was required to make it work in the context of the feature command.

gep13:
I have updated this PR to fix the errors that were appearing in the CI Builds.  This should now be ready for further review.

lib/chef/resource/chocolatey_feature.rb
```diff
@@ -0,0 +1,80 @@
+#
+# Copyright:: 2018, Chef Software, Inc.
```
tas50: This should be copyright 2019 you

lib/chef/resource/chocolatey_feature.rb
```diff
@@ -0,0 +1,80 @@
...
+# limitations under the License.
+#
+
+class Chef
+  class Resource
+    class ChocolateyFeature < Chef::Resource
+      resource_name :chocolatey_feature
+
+      description "Use the chocolatey_feature resource to enable and disable Chocolatey features."
+      introduced "x.x.x"
```
tas50: This would be 15.1

lib/chef/resource/chocolatey_feature.rb
tas50: We have rexml because you shamed me @lamont-granquist when I wrote the original resources with nokogiri

lib/chef/resource/chocolatey_feature.rb
gep13: 👍

lib/chef/resource/chocolatey_feature.rb
gep13: 👍

gep13:
@tas50 thanks for the review.  I have fixed up the points that you noted.

Any thoughts about the other questions that I asked here:

https://github.com/chef/chef/pull/8581#issuecomment-495353814

Thanks

tas50:
I'll see about the hiding of the state property. We do have a way to hide it from docs and that should get added to all the resources here. As for the docs we auto generate them (mostly) off fields here so that's on us.

PR #8584, opened 2019-05-23
Better target mode shell_out

lamont-granquist:
- integrates target-mode fully with shell_out
- removes the train_or_shellout mixin
- apt_package is now fully supported in target-mode


lib/chef/mixin/shell_out.rb
```diff
@@ -153,15 +153,23 @@ def self.clean_array(*args)
...
-        cmd = if options.empty?
-                Mixlib::ShellOut.new(*args)
-              else
-                Mixlib::ShellOut.new(*args, **options)
-              end
-        cmd.live_stream ||= io_for_live_stream
-        cmd.run_command
-        cmd
+        if Chef::Config.target_mode?
+          FakeShellOut.new(args, options, transport_connection.run_command(args.join(" "))) # FIXME FIXME FIXME: join here is horrible
```
zenspider: WRT `FIXME`: You probably want to look at `Shellwords.join` and/or `Shellwords.escape`

zenspider: They can get hairy by doing too much or having two accidental layers of escaping so this might not be the right layer or you might need to detect if it has been done already... or something?

lib/chef/mixin/shell_out.rb
```diff
@@ -153,15 +153,23 @@ def self.clean_array(*args)
...
+        else
+          cmd = if options.empty?
+                  Mixlib::ShellOut.new(*args)
+                else
+                  Mixlib::ShellOut.new(*args, **options)
```
zenspider: I don't think you need to do the conditional here. `**empty_hash` should be the same as nothing at all.

PR #8597, opened 2019-05-28
Add skip_docs to chocolatey_feature's feature_state

tas50:
We don't want this to show up in docs


lamont-granquist:
the original PR got the API shape wrong...

fix incoming...

PR #8599, opened 2019-05-28
chocolatey_feature: feature_state -> state and only support single action

lamont-granquist:
chef/chef#8597 made it obvious what was wrong about the way the
resource was written before.

the desired enabled/disable state needs to be property, not an action.

the human desire to express things imperatively instead of declaratively
strikes again.

this changes the API from:

```
chocolatey_feature "whatever" do
  action :enable
end

chocolatey_feature "whatever" do
  action :disable
end
```

to:

```
chocolatey_feature "whatever" do
  state :enabled
end

chocolatey_feature "whatever" do
  state :disabled
end
```

### DIFF
--- a/lib/chef/resource/chocolatey_feature.rb
+++ b/lib/chef/resource/chocolatey_feature.rb
@@ -1,5 +1,5 @@
 #
-# Copyright:: 2019, Chef Software, Inc.
+# Copyright:: 2019-2019, Chef Software Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lamont-granquist:
But that's why I changed it from `:enable` (verb) to `:enabled` (noun) to make it declarative.

There's one feature that is under management for every resource here.  The proper way to do this is to have one action which fixes it to the declarative state, and figures out the correct way to do that.

The service resource was widely recognized to be a poor API after it was designed.  The default behavior of that should have been to have the running/stopped state and the enabled/disabled on boot state as declarative properties.  Then the actions would only have been used from notification when you necessarily had to roll your own orchestration of how things happened.  Instead it was built ass-backwards around the orchestration case first, which means it has all kinds of ugly internal state which interacts tightly with the action.

This is way cleaner because it loads the current state into the property on the current_resource, then compares that against the desired state on the new_resource and then manages the transition, which makes reporting and everything else work correctly out of the box.

lamont-granquist:
And consider that our language is not a model of how to verb things correctly:

```ruby
# do not create an /etc/shadow file on machines from 1996 that don't have one, but fix the ones that do
file "/etc/shadow" do
  owner "root"
  group "root"
  mode 0600
  action :create
  only_if { File.exist?("/etc/shadow") }
end
```
action...  create?

that isn't what we're doing there.

lamont-granquist:
feeling now like i don't want to die on this hill

ultimately this is all gonna end in tears though

gep13:
@lamont-granquist apologies for any friction that was caused here.  This certainly wasn't my intent with the original PR.

PR #8614, opened 2019-05-29
Target mode for systemd service helper

lamont-granquist:
Makes it so that local cookbook can bounce a remote service over
train/ssh/whatever.  Works at least for ubuntu.

Introduces a fairly rudimentary TrainHelpers class, using some
inspiration from ohbye.

### DIFF
--- a/lib/chef/mixin/train_helpers.rb
+++ b/lib/chef/mixin/train_helpers.rb
@@ -1,0 +1,44 @@
+#--
+# Author:: Lamont Granquist <lamont@chef.io>
+# Copyright:: Copyright 2010-2018, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "stringio" unless defined?(StringIO)
+
+class Chef
+  module Mixin
+    module TrainHelpers
+      def file_exist?(*args)
+        if Chef::Config.target_mode?
+          Chef.run_context.transport_connection.file(args[0]).exist?
+        else
+          File.exist?(*args)
+        end
+      end
+
+      # XXX: modifications to the StringIO won't get written back
+      def file_open(*args, &block)
+        if Chef::Config.target_mode?
+          content = Chef.run_context.transport_connection.file(args[0]).content
+          string_io = StringIO.new content
+          yield string_io if block_given?
+          string_io
+        else
+          File.open(*args, &block)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/chef/mixin/train_helpers.rb
```diff
@@ -0,0 +1,44 @@
+#--
+# Author:: Lamont Granquist <lamont@chef.io>
+# Copyright:: Copyright 2010-2018, Chef Software Inc.
```
tas50: it's 2019

lib/chef/mixin/train_helpers.rb
```diff
@@ -0,0 +1,44 @@
...
+    module TrainHelpers
+      def file_exist?(*args)
+        if Chef::Config.target_mode?
+          Chef.run_context.transport_connection.file(args[0]).exist?
+        else
+          File.exist?(*args)
+        end
+      end
+
+      # XXX: modifications to the StringIO won't get written back
```
tas50: yard?

lib/chef/mixin/train_helpers.rb
```diff
@@ -0,0 +1,44 @@
...
+      def file_open(*args, &block)
+        if Chef::Config.target_mode?
+          content = Chef.run_context.transport_connection.file(args[0]).content
```
tas50: do we need any error handling here if the file can't be read or are we fine with just letting train blow up with its own error?

lamont-granquist: doesn't matter right now, the only place we use it we check file_exist? beforehand.

that's problematic, because what should we do, and should we just fake out standard File.open errors so that code doesn't have to change?  or should we throw some other Chef::Exception error?  or should we just require consumer code to catch the error(s) that train throws?

and this whole approach is super hacky and i don't know how far it'll actually go, but i can't sit down and design the correct thing without doing exploratory surgery.

lib/chef/mixin/train_helpers.rb
```diff
@@ -0,0 +1,60 @@
+#--
```
tas50: ?

tas50:
Tests?

tas50:
Also the DCO is failing here

lib/chef/node_map.rb
```diff
@@ -287,8 +287,6 @@ def canonical_matches?(canonical, matcher)
     # "provides" lines with identical filters sort by class name (ascending).
     #
     def compare_matchers(key, new_matcher, matcher)
-      cmp = compare_matcher_properties(new_matcher[:target_mode], matcher[:target_mode])
-      return cmp if cmp != 0
```
btm: why did we not need this? I think I added it out of copy/pasta.

lamont-granquist: yeah it was a copypasta error...

so what this does is establish the priority and uniqueness of two different rules.

that created problem where i think this line:

```
provides :service, target_mode: true
```

started behaving more like:

```
provides :service { true }
```

so that instead of being the least-specific, lowest-priority match it went to being the highest-priority and difficult to override.  i might not be remembering the case entirely correctly, but it definitely caused some provider_resolver tests to explode.  didn't show up until i started messing with the service provider (which is going to be where all the weird edge cases are).

now the meaning of the flag is identical to what it meant before, but the target_mode flag just means "also supports target mode".

lamont-granquist: (actually its not uniqueness, because there is still that `matches_target_mode?` check, it is just used for sort order -- otherwise deleting those lines would have no-op'd it completely)

Issue #8550, opened 2019-05-15
[chef-15] knife bootstrap ssh gives Train::PlatformDetectionFailed error if --connection-user is not provided.


Vasu1105:
## Description

knife bootstrap with ssh gives the `Train::PlatformDetectionFailed` error if --connection-user is not provided. It should give valid error message which will help user to understand the root cause.

## Chef Version

15.0.293
## Platform Version

Ubuntu 18.04
## Replication Case

I am running below command.
```
knife bootstrap <ip-address> -N vj-test-ssh -c ~/workspace/chef-repo/.chef/knife.rb --ssh-identity-file ~/.ssh/id_rsa --sudo -VV --yes
```
## Client Output

```

```

## Stacktrace

```
knife bootstrap <ip-address> -N vj-test-ssh -c ~/workspace/chef-repo/.chef/knife.rb --ssh-identity-file ~/.ssh/id_rsa --sudo -VV --yes
INFO: Using configuration from /home/msys/workspace/chef-repo/.chef/knife.rb
DEBUG: Checking if we need to accept Chef license to bootstrap node
DEBUG: Reading products and relationships...
DEBUG: Successfully read products and relationships
DEBUG: License acceptance required for chef version: 15
DEBUG: Searching for the following licenses: ["infra-client", "inspec"]
DEBUG: Found license chef_infra_client at /home/msys/.chef/accepted_licenses/chef_infra_client
DEBUG: Found license inspec at /home/msys/.chef/accepted_licenses/inspec
DEBUG: Missing licenses remaining: []
DEBUG: All licenses present
Creating new client for vj-test-ssh
Creating new node for vj-test-ssh
Connecting to <ip-address>
DEBUG: [SSH] root@<ip-address><{:user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>60, :auth_methods=>["none", "publickey"], :keys_only=>true, :keys=>["/home/msys/.ssh/id_rsa"], :password=>"<hidden>", :forward_agent=>nil, :non_interactive=>true, :verify_host_key=>:always}> (cmd.exe /c ver)
[...]

PR #8624, opened 2019-05-31
Print warning for platform detection error

dheerajd-msys:

Print warning for platform detection error if user is not provided.

## Description
Handles `Train::PlatformDetectionFailed` error

## Related Issue
fixes https://github.com/chef/chef/issues/8550

## Output
```
INFO: Using configuration from E:/Backup/Project/chef-starter/chef-repo/.chef/knife.rb
Connecting to 34.219.121.27
WARN: [SSH] PTY requested: stderr will be merged into stdout
WARNING: Could not connect with 'root'. Please provide valid user_name
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -621,6 +621,9 @@ class Chef
           # FIXME: this should save the key to known_hosts but doesn't appear to be
           config[:ssh_verify_host_key] = :accept_new
           do_connect(connection_opts(reset: true))
+        elsif e.message == "Sorry, we are unable to detect your platform" # Train::PlatformDetectionFailed
+          ui.warn("Could not connect with 'root'. Please provide valid user_name")
+          exit
         elsif ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed
           if connection.password_auth?
             raise


Vasu1105:
@dheerajd-msys I tried testing this fix but unfortunately its not fixing the error which I am getting. It still shows the same error which I mentioned in #8550

lib/chef/knife/bootstrap.rb
```diff
@@ -621,6 +621,9 @@ def connect!
           # FIXME: this should save the key to known_hosts but doesn't appear to be
           config[:ssh_verify_host_key] = :accept_new
           do_connect(connection_opts(reset: true))
+        elsif e.message == "Sorry, we are unable to detect your platform" # Train::PlatformDetectionFailed
+          ui.warn("Could not connect with 'root'. Please provide valid user_name")
```
lamont-granquist: You should be checking the exception here, not the text.

Also though if platform detection fails even though train can login to the host, but train doesn't actually know how to detect the platform it is going to fail with an incorrect message about supplying a valid user name.

Also root looks hardcoded here.   Plus there could be a lot of other cases where SSH might fail.

Most likely this must be changed in train to support better error messages about why platform detection fails.

lamont-granquist: you could do something like `ui.warn("Could not detect the platform (check your username #{config[:whatevertheusernameis]} is correct and pass the correct username with --whatvertheuserflagis)")`

and the other problem is that "root" is certainly wrong if the target is windows.

Vasu1105: Yes @lamont you are correct here. So we are working out on finding right solution and debugging more to give right fix for this issue. May today I will fix this up.

Vasu1105: The main issue was that its giving wrong error and it should not go for platform detection unless the connection is made.  

So I tried debugging few things like I ran following on irb
```irb(main):006:0> Net::SSH.start("104.211.52.56", "root", :user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>60, :auth_methods=>["none", "publickey"], :keys_only=>true, :keys=>["/home/msys/.ssh/id_rsa"], :password=>"<hidden>", :non_interactive=>true, :verify_host_key=>:accept_new) do |ssh|
irb(main):007:1*   result = ssh.exec!("ls -l")
irb(main):008:1>   puts result
irb(main):009:1> end
Please login as the user "azure" rather than the user "root".
```
Which is actually a right behavior and it gives error to use what is expected and same should be displayed to user but that's not happening so I am fixing this now.

Vasu1105:
I am closing this PR as this has fixed Train here once that's gets released and updated the version in Chef #8550 will get fixed.

PR #8626, opened 2019-05-31
Chef-15: Added deprecation check for short arguments

### LINKED ISSUE #8566: Old knife bootstrap windows winrm -x short arg should raise a deprecation
Before Chef 15 / knife-windows 3, running `knife bootstrap windows winrm` took a `-x USER` short flag. We've deprecated `--winrm-user` to `--connection-user` but we should also catch `-x` and tell the user to use `-U` instead. 

What you get right now is:
```
Error: invalid option: -x
[snip --help output]
```

Nimesh-Msys:
- Handles "-x" that was used to set "winrm-user" and is now Deprecated.
- Fixes #8566 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -348,10 +348,11 @@ class Chef
         }
 
       DEPRECATED_FLAGS = {
-        # deprecated_key: [new_key, deprecated_long]
+        # deprecated_key: [new_key, deprecated_long, replacement_value, deprecated_short]
         # optional third element: replacement_value - if converting from bool
         #   (--bool-option) to valued flag (--new-option VALUE)
         #   this will be the value that is assigned the new flag when the old flag is used.
+        # optional fourth element: deprecated_short - if the short form of flag has also been deprecated.
         auth_timeout: [:max_wait, "--max-wait SECONDS" ],
         forward_agent:
           [:ssh_forward_agent, "--forward-agent"],
@@ -360,7 +361,7 @@ class Chef
         prerelease:
           [:channel, "--prerelease", "current"],
         ssh_user:
-          [:connection_user, "--ssh-user USER"],
+          [:connection_user, "--ssh-user USERNAME"],
         ssh_password:
           [:connection_password, "--ssh-password PASSWORD"],
         ssh_port:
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -368,7 +369,7 @@ class Bootstrap < Knife
         ssl_peer_fingerprint:
           [:winrm_ssl_peer_fingerprint, "--ssl-peer-fingerprint FINGERPRINT"],
         winrm_user:
-          [:connection_user, "--winrm-user USER"],
+          [:connection_user, "--winrm-user USER", nil, "-x USERNAME"],
```
btm: seems like both options should use either USER or USERNAME here.

Nimesh-Msys: As per the previous implementation, I used [-x USERNAME](https://github.com/chef/chef/blob/chef-14/lib/chef/knife/bootstrap.rb#L48). If we choose to use USERNAME here, should we also update the same for ssh_user option ?

marcparadise: Let's make it consistently USERNAME across connection-user, winrm-user, ssh-user

lib/chef/knife/bootstrap.rb
```diff
@@ -384,18 +385,28 @@ class Bootstrap < Knife
       }.freeze
 
       DEPRECATED_FLAGS.each do |deprecated_key, deprecation_entry|
-        new_key, deprecated_long, replacement_value = deprecation_entry
+        new_key, deprecated_long, replacement_value, deprecated_short = deprecation_entry
```
marcparadise: I think this is good for now, but it's getting unwieldy. I think we're at the point where if we add more, it might make sense to have a DeprecatedOption struct. 


With  I'm POCing some mixlib-install work that will manage deprecation behaviors there; but this is good for now.

spec/unit/knife/bootstrap_spec.rb
```diff
@@ -1,6 +1,6 @@
 #
 # Author:: Ian Meyer (<ianmmeyer@gmail.com>)
-# Copyright:: Copyright 2010-2016, Ian Meyer
+# Copyright:: Copyright 2010-2019, Chef Software Inc.
```
btm: We need to leave the original copyright here. We can add another line below it.

Nimesh-Msys: Done. Thanks

btm:
We'll take this for now. I believe @marcparadise is working on adding this upstream in mixlib-cli to fix https://github.com/chef/mixlib-cli/issues/63, so later we can back all of this out for the cleaner solution.

PR #8633, opened 2019-06-04
Avoid constant warnings

tas50:
Chef has gotten quite loud with warnings. This quiets things back down a
bit.


lamont-granquist:
although why are we Kernel.loading these things more than once?

tas50:
Most are not being loaded twice but a few during bootstrapping as reported on Discourse

Issue #8634, opened 2019-06-04
Ability to further configure Chocolatey source


gep13:
### Describe the Enhancement:
In my Chef Recipes, I want to be able to configure additional properties on a Chocolatey source.  This would include the ability to:

* enable a source
* disable a source
* set additional properties for a source such as `admin_only` and `allow_self_service`. 

### Describe the Need:
Anyone currently making use of Chocolatey and Chef would benefit from being able to further configure Chocolatey sources..

This would also bring better feature parity for Chocolatey in Chef compared to some of the other Configuration Managers out there:

![image](https://user-images.githubusercontent.com/1271146/58280048-aaee9c80-7d54-11e9-92b0-ce59aa965f3c.png)

### Current Alternative
Currently possible to execute Chocolatey commands directly in recipe, but not as "nice" as having a built in resource for it.

### Can We Help You Implement This?:
This would be an addition to the existing chocolatey_source resource (along with additional unit tests).

I have a PR in the making to provide this functionality.

gep13:
Related PR has been merged, so going to go ahead and close out this issue.

PR #8635, opened 2019-06-04
chocolatey_source: Add additional actions and propereties for configuring sources

gep13:
## Description
This provides the ability to set additional properties for a Chocolatey
source with properties `admin_only` and `allow_self_service`.  As well
as the ability to enable/disable a source.

## Related Issue

https://github.com/chef/chef/issues/8634

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/chocolatey_source.rb
+++ b/lib/chef/resource/chocolatey_source.rb
@@ -17,7 +17,6 @@
 class Chef
   class Resource
     class ChocolateySource < Chef::Resource
-      preview_resource true
       resource_name :chocolatey_source
 
       description "Use the chocolatey_source resource to add or remove Chocolatey sources."
@@ -41,7 +40,7 @@ class Chef
       property :priority, Integer, default: 0,
                description: "The priority level of the source."
 
-      property :source_state, [TrueClass, FalseClass], default: false, desired_state: false, skip_docs: true
+      property :disabled, [TrueClass, FalseClass], default: false, desired_state: false, skip_docs: true
 
       load_current_value do
         element = fetch_source_element(source_name)
@@ -53,13 +52,13 @@ class Chef
         admin_only element["adminOnly"] == "true"
         allow_self_service element["selfService"] == "true"
         priority element["priority"].to_i
-        source_state element["disabled"] == "true"
+        disabled element["disabled"] == "true"
       end
 
       # @param [String] id the source name
       # @return [REXML::Attributes] finds the source element with the
       def fetch_source_element(id)
-        require "rexml/document"
+        require "rexml/document" unless defined?(REXML::Document)
 
         config_file = "#{ENV['ALLUSERSPROFILE']}\\chocolatey\\config\\chocolatey.config"
         raise "Could not find the Chocolatey config at #{config_file}!" unless ::File.exist?(config_file)
@@ -92,7 +91,7 @@ class Chef
       action :disable do
         description "Disables a Chocolatey source."
 
-        if current_resource.source_state != true
+        if current_resource.disabled != true
           converge_by("disable Chocolatey source '#{new_resource.source_name}'") do
             shell_out!(choco_cmd("disable"))
           end
@@ -102,7 +101,7 @@ class Chef
       action :enable do
         description "Enables a Chocolatey source."
 
-        if current_resource.source_state == true
+        if current_resource.disabled == true
           converge_by("enable Chocolatey source '#{new_resource.source_name}'") do
             shell_out!(choco_cmd("enable"))
           end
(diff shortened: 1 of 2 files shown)


gep13:
@tas50 @lamont-granquist what are your thoughts on the following changes?

I am aware of the conversation that is going on here:

https://github.com/chef/chef/pull/8599

regarding the API.  I wasn't sure if that discussion applied to this modification or not.  Let me know if there are changes required here.

gep13:
I attempted to add unit tests for testing the enabling/disabling of the source, and I can get this to work for one operation.  However, when I try to add the following:

```
  describe "run_action(:disable)" do
    it "when source is enabled, it disables it correctly" do
      resource.source_state false
      allow(current_resource).to receive(:fetch_source_element).with("fakey_fakerton").and_return(OpenStruct.new(disabled: "false"))
      expect(disable_provider).to receive(:shell_out!).with("C:\\ProgramData\\chocolatey\\bin\\choco source disable -n \"fakey_fakerton\"")
      resource.run_action(:disable)
      expect(resource.updated_by_last_action?).to be true
    end

    it "when source is disabled, it is idempotent when trying to disable" do
      resource.source_state true
      allow(current_resource).to receive(:fetch_source_element).with("fakey_fakerton").and_return(OpenStruct.new(disabled: "true"))
      resource.run_action(:disable)
      expect(resource.updated_by_last_action?).to be false
    end
  end
```

The unit tests fail.  I "think" it is something to do with how the resource is being mocked out, but I don't know enough about how this works to get it to work.  Could either of you provide some guidance in this area?

lib/chef/resource/chocolatey_source.rb
```diff
@@ -31,23 +32,34 @@ class ChocolateySource < Chef::Resource
...
+        allow_self_service element["selfService"] == "true"
         priority element["priority"].to_i
+        source_state element["disabled"] == "true"
       end
 
       # @param [String] id the source name
       # @return [REXML::Attributes] finds the source element with the
       def fetch_source_element(id)
-        require "rexml/document" unless defined?(REXML::Document)
+        require "rexml/document"
```
lamont-granquist: that needs to get backed out, the idempotent behavior of `require` in ruby is poor and needs the guard

lib/chef/resource/chocolatey_source.rb
```diff
@@ -77,6 +89,26 @@ def fetch_source_element(id)
...
+          converge_by("disable Chocolatey source '#{new_resource.source_name}'") do
+            shell_out!(choco_cmd("disable"))
+          end
+        end
+      end
+
+      action :enable do
+        description "Enables a Chocolatey source."
+
+        if current_resource.source_state == true
```
lamont-granquist: this has the bug that someone who does:

```ruby
chocolatey_source "whatever" do
   action :enable
end
```

and then changes that to:

```ruby
chocolatey_source "whatever" do
  allow_self_service true
   action :enable
end
```

it'll see it is already enabled and won't fix the flag.

lib/chef/resource/chocolatey_source.rb
```diff
@@ -31,23 +32,34 @@ class ChocolateySource < Chef::Resource
...
         element = fetch_source_element(source_name)
         current_value_does_not_exist! if element.nil?
 
         source_name element["id"]
         source element["value"]
         bypass_proxy element["bypassProxy"] == "true"
+        admin_only element["adminOnly"] == "true"
+        allow_self_service element["selfService"] == "true"
         priority element["priority"].to_i
+        source_state element["disabled"] == "true"
```
lamont-granquist: should probably just call this `disabled`

gep13: That makes a lot of sense 😄

lib/chef/resource/chocolatey_source.rb
gep13: I don't recall removing that 😢 I have added this back in.

lib/chef/resource/chocolatey_source.rb
gep13: This is a very good point.  If it makes any difference, the same thing will happen if you run the Chocolatey command directly as well:

image

This doesn't result in any changes to the allow_self_service property.

@lamont-granquist is there a strategy for how this should be handled?  Without creating a new resource, specifically for enabling/disabling a source, I don't see a path forward.  Is there a way to limit properties of a resource to only apply to certain actions?  This would then error if a property is set on an action that it wasn't intended?

gep13:
@lamont-granquist thank you for taking the time to review this PR, I really appreciate it!  I have addressed a couple of the comments that were made (I can rebase that commit out if required) and replied to your other comment.

gep13:
@lamont-granquist @tas50 just wanted to touch base on this PR to see if there were any other comments, or if there was any feedback on some of my questions.

Thanks

gep13:
@tas50 thank you very much for getting this merged!

Did you see my comment/question about not being able to run all the Unit Tests that I wanted to on this change?  Is this something that you can help with?

PR #8653, opened 2019-06-12
Parse YAML recipes

lamont-granquist:
YAML is now being supported.

This implementation is restrictive and does not support arbitrary ruby code embedded in the YAML.  It is being done that way to support writing a YAML-to-ruby autoconverter (so that users have an easy way to back out of YAML-land to ruby-land since YAML-land will inherently always be limited).

What this means is that the syntax is limited right now, no control flow, no looping, not even the ability to render the node object values into properties.  That would need to be done via extending the "YAML DSL" to support those, in a structured way that allowed for autoconversion to ruby later.

### DIFF
--- a/lib/chef/cookbook_version.rb
+++ b/lib/chef/cookbook_version.rb
@@ -4,7 +4,7 @@
 # Author:: Tim Hinderliter (<tim@chef.io>)
 # Author:: Seth Falcon (<seth@chef.io>)
 # Author:: Daniel DeLeo (<dan@chef.io>)
-# Copyright:: Copyright 2008-2018, Chef Software Inc.
+# Copyright:: Copyright 2008-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


lamont-granquist:
Simple cookbook/yml/recipes/default.yml:

```
---
- resources:
  - type: "file"
    name: "/tmp/xyz.out"
    content: "whatever"
  - type: "execute"
    name: "some name here"
    command: "true"
```

It also works fine as a top-level hash:

```
---
resources:
  - type: "file"
    name: "/tmp/xyz.out"
    content: "whatever"
  - type: "execute"
    name: "some name here"
    command: "true"
```

lamont-granquist:
erb also works:

```
---
- resources:
  <% ["x", "y", "z"].each do |f| %>
  - type: "file"
    name: "/tmp/<%= f %>.out"
    content: "whatever"
  <% end %>
  - type: "execute"
    name: "some name here"
    command: "true"
```

lamont-granquist:
ripped out the erb parsing.

if we go down that road that means that we can never write a YAML-to-ruby autoconverter, which means that customers would forever get stuck on in YAML-chef.

lamont-granquist:
This of course means we can no longer do even a simple thing like setting a property to the value of a node object and have to build a YAML-language feature to do that.

Issue #7389, opened 2018-06-23
Notifications fire multiple times if a resource is declared twice with the same name


teknofire:
## Description

When a a resource is declared more than once with the same name property and one of them has a notification attached, then the notification will get execute for each instances of the resource.

This appears to happen because the `immediate_notification_collection` uses `declared_key` from `notifying_resource` for its key, which would be `file[createit]` for both resources from my example below.

https://github.com/chef/chef/blob/92d85cf4c67680d778cbe8bc5c676f9a10fa1442/lib/chef/run_context.rb#L231-L236

## Chef Version

chef-client --version
Chef: 14.2.0

## Platform Version

Ubuntu 16.04

## Replication Case

Create a cookbook with the following recipe

```
# please note that this is a major simplification of a
# more complex set of cookbooks
 file 'createit' do
   path '/tmp/createit'
   content 'creating item 1'
 end

 execute 'ranit' do
   command 'echo "ran it" >> /tmp/ranit'
   action :nothing
 end

 file 'createit' do
   path '/tmp/createit2'
   content 'creating item 2'
   notifies :run, 'execute[ranit]', :immediately
 end
```

The notification from the second `file[createit]` resource will be run twice, once when the first `file[createit]` is executed and then again when the second one runs.  If you then look at the contents of  `/tmp/ranit` you will see the "ran it" appears twice.

## Client Output

```
       Starting Chef Client, version 14.2.0
       resolving cookbooks for run list: ["notifications_same::default"]
       Synchronizing Cookbooks:
         - notifications_same (0.1.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       Converging 3 resources
       Recipe: notifications_same::default
         * file[createit] action create
           - create new file /tmp/createit
           - update content in file /tmp/createit from none to eb8e3a
           --- /tmp/createit	2018-06-23 00:31:13.862529471 +0000
           +++ /tmp/.chef-createit20180623-3342-wsvyan	2018-06-23 00:31:13.862529471 +0000
           @@ -1 +1,2 @@
           +creating item 1
         * execute[ranit] action run
           - execute echo "ran it" >> /tmp/ranit
         * execute[ranit] action nothing (skipped due to action :nothing)
         * file[createit] action create
           - create new file /tmp/createit2
           - update content in file /tmp/createit2 from none to b2b8af
           --- /tmp/createit2	2018-06-23 00:31:13.882529471 +0000
[...]

PR #8668, opened 2019-06-18
Add unified_mode for resources

lamont-granquist:
This is inspired by "use_inline_resources".

Setting `unified_mode false` in a resource would be the existing
behavior with separate compile/converge phases.  This remains the
default.

Setting `unified_mode true` in a resource will eliminate the converge
phase.  Reverse notifications and delayed notifications will still
fire.  The resource action will behave like all resources are executing
at compile time.

As a aside, notifications have never worked for resources firing at
compile time.  This implementation gets that behavior correct so
that notifications will work.

Of course forward notifications to resources not yet declared will not
be possible.

Setting `resource_unified_mode_default true` in `Chef::Config` would
turn off unified mode and turn off the split compile/converge mode for every
custom resource.

NOTE: This does not affect recipe mode at all. 

As an example old annoying code:

```ruby
action :doit do
  file "/tmp/foo.xyz" do
    content "something"
    action :nothing
  end.run_action(:create)

  puts "exists" if File.exist?("/tmp/foo.xyz")

  if some_conditional?
    converge_by("do something to /tmp/foo.xyz") do
      [ ... some code here that consumes /tmp/foo.xyz... ]
    end
  end
end
```

```ruby
unified_mode true # put this anywhere in the declaration of the resource

action :doit do
  file "/tmp/foo.xyz" do
    content "something"
    # it is no longer necessary to `action :nothing` this
    # there simply IS no converge_phase
  end  # no forcing to compile_time here

  # at this point normal non-resource code will see the file as having been created

  puts "exists" if File.exist?("/tmp/foo.xyz")

  if some_conditional?
    converge_by("do something to /tmp/foo.xyz") do
      [ ... some code here that consumes /tmp/foo.xyz... ]
    end
  end
end
```

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -875,6 +875,7 @@ module ChefConfig
     #
     # NOTE: CHANGING THIS SETTING MAY CAUSE CORRUPTION, DATA LOSS AND
     # INSTABILITY.
+    #
     default :file_atomic_update, true
 
     # There are 3 possible values for this configuration setting.
@@ -882,18 +883,27 @@ module ChefConfig
     # false => file staging is done via tempfiles under ENV['TMP']
     # :auto => file staging will try using destination directory if possible and
     #   will fall back to ENV['TMP'] if destination directory is not usable.
+    #
     default :file_staging_uses_destdir, :auto
 
     # Exit if another run is in progress and the chef-client is unable to
     # get the lock before time expires. If nil, no timeout is enforced. (Exits
     # immediately if 0.)
+    #
     default :run_lock_timeout, nil
 
     # Number of worker threads for syncing cookbooks in parallel. Increasing
     # this number can result in gateway errors from the server (namely 503 and 504).
     # If you are seeing this behavior while using the default setting, reducing
     # the number of threads will help.
+    #
     default :cookbook_sync_threads, 10
+
+    # True if all resources by default default to unified mode, with all resources
+    # applying in "compile" mode, with no "converge" mode. False is backwards compatible
+    # setting for Chef 11-15 behavior.  This will break forward notifications.
+    #
+    default :resource_unified_mode_default, false
 
     # At the beginning of the Chef Client run, the cookbook manifests are downloaded which
     # contain URLs for every file in every relevant cookbook.  Most of the files
@@ -920,9 +930,9 @@ module ChefConfig
     default :no_lazy_load, true
 
     # A whitelisted array of attributes you want sent over the wire when node
-    # data is saved.
-    # The default setting is nil, which collects all data. Setting to [] will not
-    # collect any data for save.
+    # data is saved. The default setting is nil, which collects all data. Setting
+    # to [] will not collect any data for save.
+    #
     default :automatic_attribute_whitelist, nil
     default :default_attribute_whitelist, nil
     default :normal_attribute_whitelist, nil
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -509,5 +509,17 @@ class Chef
         super "Conflicting requirements for gem '#{gem_name}': Both #{value1.inspect} and #{value2.inspect} given for option #{option.inspect}"
       end
     end
+
+    class UnifiedModeImmediateSubscriptionEarlierResource < RuntimeError
+      def initialize(notification)
+        super "immediate subscription from #{notification.resource} resource cannot be setup to #{notification.notifying_resource} resource, which has already fired while in unified mode"
+      end
+    end
+
+    class UnifiedModeBeforeSubscriptionEarlierResource < RuntimeError
+      def initialize(notification)
+        super "before subscription from #{notification.resource} resource cannot be setup to #{notification.notifying_resource} resource, which has already fired while in unified mode"
+      end
+    end
   end
 end
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -1,7 +1,7 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
 # Author:: Christopher Walters (<cw@chef.io>)
-# Copyright:: Copyright 2008-2016, 2009-2018, Chef Software Inc.
+# Copyright:: Copyright 2008-2016, 2009-2019, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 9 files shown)


lamont-granquist:
Thoughts on names:

```ruby
converge_mode false
one_pass true
unified true
fused true
```

Latter has some name-conflicts with poise's fused resources though which are kinda like custom resources.

I think it should probably get inverted and be a "true" thing rather than a "false" thing, because most humans seem to like being positive.

lamont-granquist:
I'm sorta leaning `unified_mode true`

lamont-granquist:
failures seem to be SuSE stuff and the ruby-prof appveyor one.

lamont-granquist:
So per meeting notes:

- should log that it is experimental only for now
- need to investigate how it interacts with `subscribes` since in my head what happens is that the subscribes will necessarily have to happen too late, and the resource will have already have fired a long time ago so nothing will happen

lamont-granquist:
somehow an early resource which subscribes to a later resource works (for subscribes the resolution of the string to the resource name must be lazy?) while having a later resource subscribe to an earlier resource fails because the action isn't tracked (maybe this is a job for the action_collection? hrm...).

lamont-granquist:
This now works correctly with 6 out of 8 possible notification schemes:

1.  delayed notification to a resource declared either before or after the notifying resource works (2 cases)
2.  delayed subscribing to a resource decalred either before or after the subscribing resource works (2 cases)
3.  immediate notification to a resource that is declared earlier works.
4.  immediate subscribing to a resource that is declared later, and has not yet fired works.
5.  immediate notification to a resource that is declared later fails (that resource has not yet been parsed so i can't "immediately" be notified without predicting the future).
6.  immediate subscribing to a resource that is declared earlier and has fired fails (you cannot rewind time to fire the currently parsed resource when that prior resource had fired earlier).

Since 1+2 are the common case that covers most of the bases, and I think may be mildly surprising that it works, but with the lazy subscribes code that we already have it was fairly easy to lazily keep around the notification reference and only resolve it when the delayed notifications are fired after the resource is fully parsed.

The rule for immediate notifications should really be to declare resources first and then notify them which makes better intuitive sense than the subscribing to resources that are not yet declared.

I could have tried to make the 5+6 cases work, by interpreting `:immediate` to mean "immediately when the later resource is parsed" but that seem rather weird and magical and the code was gonna be *very* ugly which is usually a good indication that its a bad idea.

lamont-granquist:
So now what I don't like is this pattern is common and now breaks:

```

actions :install

default_action :install

property :source, String
property :version, String, name_property: true

action :install do
      source_url = new_resource.source

      remote_file "/tmp/redis-#{new_resource.version}.tar.gz" do
        #source "http://download.redis.io/releases/redis-#{new_resource.version}.tar.gz"
        new_resource.source source_url
        notifies :run, "execute[unzip_redis_archive]", :immediately
      end
      
      execute "unzip_redis_archive" do
        command "tar xzf redis-#{new_resource.version}.tar.gz"
        cwd "/tmp"
        action :nothing
        notifies :run, "execute[build_and_install_redis]", :immediately
      end
      
      execute "build_and_install_redis" do
        command 'make && make install'
        cwd "/tmp/redis-#{new_resource.version}"
        action :nothing
        notifies :run, "execute[install_server_redis]", :immediately
      end
      
      execute "install_server_redis" do
        command "echo -n | ./install_server.sh"
        cwd "/tmp/redis-#{new_resource.version}/utils"
        action :nothing
      end
end
```

We've broken forward immediate notifications, so this doesn't work.  Further if you just change the notifies to subscribes and bump then down to the next resource you would get equivalent pre-unified-mode behavior.  But that is also broken.

To make it work you have to flip the order resources are declared, which is now the inverse order of how they are declared:

```
unified_mode true

actions :install

default_action :install

property :source, String
property :version, String, name_property: true

action :install do
      source_url = new_resource.source

      execute "install_server_redis" do
        command "echo -n | ./install_server.sh"
        cwd "/tmp/redis-#{new_resource.version}/utils"
        action :nothing
      end
      
      execute "build_and_install_redis" do
        command 'make && make install'
        cwd "/tmp/redis-#{new_resource.version}"
        action :nothing
        notifies :run, "execute[install_server_redis]", :immediately
      end

      execute "unzip_redis_archive" do
        command "tar xzf redis-#{new_resource.version}.tar.gz"
        cwd "/tmp"
        action :nothing
        notifies :run, "execute[build_and_install_redis]", :immediately
      end

      remote_file "/tmp/redis-#{new_resource.version}.tar.gz" do
        #source "http://download.redis.io/releases/redis-#{new_resource.version}.tar.gz"
        new_resource.source source_url
        notifies :run, "execute[unzip_redis_archive]", :immediately
      end
end
```

This is a bit easier didactically though "you have to declare things first and you call backwards" is fairly simple to teach, and it will fail hard when the user gets it wrong, so it isn't annoying subtle incorrectness like compile-converge mode.  It is incorrectness that will beat you over the head.

Still it feels like we're lacking something here, and that notifications may even be a poor way to express it, because really what you want to express is a simple "if <this resource updates>, then <evaluate all these resources in order>".

lamont-granquist:
So this PR is blowing my mind a bit:

```
% cat cookbooks/test/resources/myresource.rb
provides :myresource

property :name, String

unified_mode true

action :doit do
  file "/tmp/xyz.out" do
    content "whatever"
    notifies :run, 'ruby_block[boom]', :immediate
  end

  ruby_block "boom" do
    block do
      raise "next resources goes boom"
    end
    action :nothing
  end

rescue
  file "/tmp/xyz.out" do
    action :delete
  end
end
```

results in:

```
% chef-client -z -j dna.json -c client.rb
Starting Chef Infra Client, version 15.2.22
resolving cookbooks for run list: ["test::test"]
Synchronizing Cookbooks:
  - test (0.0.1)
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 2 resources
Recipe: test::test
  * log[it] action nothing (skipped due to action :nothing)
  * test_myresource[whatever] action doit
    * file[/tmp/xyz.out] action create
      - create new file /tmp/xyz.out
      - update content in file /tmp/xyz.out from none to 85738f
      --- /tmp/xyz.out	2019-08-13 13:59:11.391832871 -0700
      +++ /tmp/.chef-xyz20190813-60134-vx3by9.out	2019-08-13 13:59:11.391636855 -0700
      @@ -1 +1,2 @@
      +whatever
    * ruby_block[boom] action nothing (skipped due to action :nothing)
    * ruby_block[boom] action run

      Error executing action `run` on resource 'ruby_block[boom]'

      RuntimeError
      ------------
      next resources goes boom

[.... debugging output ....]

    * file[/tmp/xyz.out] action delete
      - delete file /tmp/xyz.out

[.... rest of the error output ....]
```

In that kind of scenario where a remote_file creation is acting as a gate which triggers a notify chain which unzips the file and whatnot, you could catch exceptions anywhere in that chain, and then clean up the remote file, so that you don't get 'stuck' in an intermediate state where the file was downloaded but the chained resources never fully completed due to some transient error on the last run.

lamont-granquist:
(heh, although because i used two file resources that have the same name in that snippet the notification can't tell the difference between the originating resources, so the file with action :delete causes the notify to run again and the ruby_block to fire again and another exception to get run -- but that is just #7389 and nothing to do with unified_mode specifically but fallout from removing the CHEF-3694 warnings without doing something about warning about when same-named resources cause confusion in notifies).

PR #8669, opened 2019-06-18
Windows Path on Bootstrap

Xorima:
Fix issue with windows path during bootstrap

## Description
As a windows user when I bootstrap a node I expect the path to not have items removed as my cookbooks may depend on executables not stored on common system paths, such as sqlcmd which will break and cause the initial chef run to fail

## Types of changes

## Checklist:


tas50:
Another one for you @marcparadise

Xorima:
I think the reason this was changed was too long file paths but it has other unintended causes, 
I guess the question is how can we validate the length of the path without it getting too long, and as we are using a cmd file it limits the options we would have with say powershell

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -158,7 +158,7 @@ def get_log_location
 
         def start_chef
           bootstrap_environment_option = bootstrap_environment.nil? ? "" : " -E #{bootstrap_environment}"
-          start_chef = "SET \"PATH=%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem;%SYSTEMROOT%\\System32\\WindowsPowerShell\\v1.0\\;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\"\n"
+          start_chef = "SET \"PATH=%PATH%;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\"\n"
```
marcparadise: By allowing custom path elements first, we run the risk of grabbing the wrong ruby/other utility.  Could you modify this to something like: 
`"SET \"PATH=%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem;%SYSTEMROOT%\\System32\\WindowsPowerShell\\v1.0\\;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\;%PATH%"\n"
```
It'll have some redundancy, but will ensure that the chef-related paths are first during bootstrap.

Xorima:
Resolved as per feedback @marcparadise apologies for the delay

Xorima:
Resolved as per feedback @marcparadise apologies for the delay

Xorima:
tests failing on linux so doubt it is due to this one?

Xorima:
Nudge @marcparadise could this be looked at please?

PR #8670, opened 2019-06-18
Move the data collector should_be_enabled? check

lamont-granquist:
This works better as part of the module that handles all the config
wrangling.  It isn't central to the data_collector operation itself.


tas50:
Makes a lot more sense there

PR #8676, opened 2019-06-20
Improvements to omnibus buildkite pipeline

jeremiahsnapp:
This PR does NOT migrate omnibus builds to buildkite. That will be done in a subsequent PR after final tests of the buildkite pipeline are completed.

This PR enables all platforms in omnibus buildkite pipeline and ports appropriate functionality from `ci/verify-chef.*` scripts to the new `omnibus/omnibus-test.*` scripts.

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -7,52 +7,51 @@ fips-platforms:
   - el-*-x86_64
   - windows-*
 builder-to-testers-map:
-  # aix-7.1-powerpc:
-  #   - aix-7.1-powerpc
+  aix-7.1-powerpc:
+    - aix-7.1-powerpc
   debian-8-x86_64:
     - debian-8-x86_64
     - debian-9-x86_64
   el-6-i386:
     - el-6-i386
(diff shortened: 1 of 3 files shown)


.expeditor/release.omnibus.yml
```diff
@@ -7,52 +7,51 @@ fips-platforms:
...
-  #   - solaris-11-sparc
+  solaris-11-i86pc:
+    - solaris-11-i86pc
+  solaris-11-sparc:
+    - solaris-11-sparc
   ubuntu-14.04-i386:
     - ubuntu-14.04-i386
-  # ubuntu-14.04-ppc64le:
-  #   - ubuntu-14.04-ppc64le
+  ubuntu-14.04-ppc64le:
```
tas50: Not sure if we want to start out BK for chef/chef with Ubuntu 14.04 since we're trying to move off that.

jeremiahsnapp: After some discussion we've agreed to focus on migrating all current platforms to buildkite first and then followup later with removing deprecated platforms.

PR #8678, opened 2019-06-20
[chef-client] [scaffolding-chef] add new build configuration for scaffolding-chef

echohack:
Signed-off-by: echohack <echohack@users.noreply.github.com>

When having expeditor paths in the .bldr.toml that aren't the default in `/`, we need to be more explicit about which directories are being watched for new events to fire builds.

Additionally, we layered `chef-client` one more folder deeper, and I realized this would break the usage of the VERSION file. So I fixed that.

tenor-82158748

### DIFF
--- a/habitat-packages/chef-client/plan.sh
+++ b/habitat-packages/chef-client/plan.sh
@@ -3,14 +3,32 @@ pkg_origin=chef
 pkg_maintainer="The Chef Maintainers <humans@chef.io>"
 pkg_description="The Chef Client"
 pkg_license=('Apache-2.0')
-pkg_filename=${pkg_dirname}.tar.gz
+pkg_filename="${pkg_name}-${pkg_version}.tar.gz"
 pkg_bin_dirs=(bin)
-pkg_build_deps=(core/make core/gcc core/git)
-pkg_deps=(core/glibc core/ruby26 core/libxml2 core/libxslt core/libiconv core/xz core/zlib core/bundler core/openssl core/cacerts core/libffi core/coreutils core/libarchive)
+pkg_build_deps=(
+  core/make
+  core/gcc
+  core/git
+)
+pkg_deps=(
+  core/glibc
+  core/ruby26
+  core/libxml2
+  core/libxslt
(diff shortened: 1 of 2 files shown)


habitat-packages/chef-client/plan.sh
```diff
@@ -2,7 +2,7 @@ pkg_name=chef-client
 pkg_origin=chef
 pkg_maintainer="The Chef Maintainers <humans@chef.io>"
 pkg_description="The Chef Client"
-pkg_version=$(cat ../VERSION)
+pkg_version=$(cat ../../VERSION)
```
tduffield: Not a blocker, but I'd prefer we use the update_version function (https://github.com/chef/ci-studio-common/blob/master/habitat/plan.sh#L10-L17), or use Expeditor to update the version inline using sed (https://github.com/chef/expeditor/blob/master/.expeditor/update_version.sh#L7).

echohack: Whoa. Using pkg_version() like that is really dope and intuitive.

echohack: Fixed! I went with the pkg_version() route.

habitat-packages/chef-client/plan.sh
```diff
@@ -3,14 +3,32 @@ pkg_origin=chef
 pkg_maintainer="The Chef Maintainers <humans@chef.io>"
 pkg_description="The Chef Client"
 pkg_license=('Apache-2.0')
-pkg_filename=${pkg_dirname}.tar.gz
+pkg_filename="${pkg_name}-${pkg_version}.tar.gz"
 pkg_bin_dirs=(bin)
-pkg_build_deps=(core/make core/gcc core/git)
-pkg_deps=(core/glibc core/ruby26 core/libxml2 core/libxslt core/libiconv core/xz core/zlib core/bundler core/openssl core/cacerts core/libffi core/coreutils core/libarchive)
+pkg_build_deps=(
```
tas50: Thanks for this. So much easier to read

PR #8682, opened 2019-06-24
Ignore noisy outputs at create temp_dir during bootstrap

sawanoboly:
Signed-off-by: sawanoboly <sawanoboriyu@higanworks.com>

## Description

When using the sudo option, bootstrap may fail. Since pty is enabled in the SSH connection, the path acquisition of temp_dir does not work well when stderr contains unnecessary output.

for example

```
$ sudo mktemp -d
sudo: unable to resolve host hostnamelocalhost  # << this message outs on stderr but merged into stdin due to pty is true
/tmp/tmp.FgLKVeE87G
```

This PR is changed to use only the last line to ignore such output.

## Related Issue

none.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap/train_connector.rb
+++ b/lib/chef/knife/bootstrap/train_connector.rb
@@ -117,7 +117,7 @@ class Chef
           cmd = windows? ? MKTEMP_WIN_COMMAND : MKTEMP_NIX_COMMAND
           @tmpdir ||= begin
                         res = run_command!(cmd)
-                        dir = res.stdout.chomp.strip
+                        dir = res.stdout.split.last
                         unless windows?
                           # Ensure that dir has the correct owner.  We are possibly
                           # running with sudo right now - so this directory would be owned by root.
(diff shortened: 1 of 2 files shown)


tas50:
@marcparadise I'm tagging you on this one since you touched this all last.

lib/chef/knife/bootstrap/train_connector.rb
```diff
@@ -117,7 +117,7 @@ def temp_dir
           cmd = windows? ? MKTEMP_WIN_COMMAND : MKTEMP_NIX_COMMAND
           @tmpdir ||= begin
                         res = run_command!(cmd)
-                        dir = res.stdout.chomp.strip
+                        dir = res.stdout.split.last
```
tas50: A small comment explaining that we split in case extra output is there would be helpful here. The test captures that scenario, but this might be confusing to someone reading the code.

sawanoboly: Thanks, I've add comment for this.

tas50: Looks good now. Thanks

PR #8708, opened 2019-07-02
fix Layout/AlignArguments

lamont-granquist:
this is using:

Layout/AlignArguments:
  Enabled: true
  EnforcedStyle: with_fixed_indentation

the default style can use really excessive whitespace.  on starting
lines which are already long, it fully indents across to where the
arguments start and then begins the line there.

the first change here is indicative of the direction it goes when you use the other style of alignment:

https://github.com/chef/chef/pull/8708/files#diff-676e90498ac31a69a5fe6cd73f79fc91L159

we have many lines of code which are 2-3 times as long as that which currently use this `:with_fixed_indentation` style and are not present in this PR, but which will
be exposed by using the default style


lamont-granquist:
Good example of the other way of going about this cop and what it does to code:

```
   it "raises an error if both name_property and name_attribute are specified" do
     expect { resource_class.property :x, name_property: false, name_attribute: 1 }.to raise_error ArgumentError,
-      /name_attribute and name_property are functionally identical and both cannot be specified on a property at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
+                                                                                                  /name_attribute and name_property are functionally identical and both cannot be specified on a propert
y at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
     expect { resource_class.property :x, name_property: false, name_attribute: nil }.to raise_error ArgumentError,
-      /name_attribute and name_property are functionally identical and both cannot be specified on a property at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
+                                                                                                    /name_attribute and name_property are functionally identical and both cannot be specified on a prope
rty at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
     expect { resource_class.property :x, name_property: false, name_attribute: false }.to raise_error ArgumentError,
-      /name_attribute and name_property are functionally identical and both cannot be specified on a property at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
+                                                                                                      /name_attribute and name_property are functionally identical and both cannot be specified on a pro
perty at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
     expect { resource_class.property :x, name_property: true, name_attribute: true }.to raise_error ArgumentError,
-      /name_attribute and name_property are functionally identical and both cannot be specified on a property at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
+                                                                                                    /name_attribute and name_property are functionally identical and both cannot be specified on a prope
rty at once. Use just one on property x of resource chef_resource_property_spec_(\d+)./
   end
```

lamont-granquist:
So there's two philosophies over indentation here broadly:

- always indent 2 characters from the start of the previous line
- always indent to where the expression begins on the previous line

The problem with the latter is that you can wind up needing 60+ character indents.

This results in humans feeling that quickly becomes ugly and now code needs to get manually changed.  This makes me not like it personally.  But it also means that contributors wind up needing more than just ripping `chefstyle -a` across the PR.  It puts the PR reviewer in the position of asking the submitter for changes or committing to fixing it themselves, which is just one more barrier towards getting to 'yes' on PRs.

So for that reason I prefer the setting where you can rip `chefstyle -a` across the PR and its always right and mergeable.  Less work for everyone, easier to get to merged.  Takes no thought.

PR #8711, opened 2019-07-02
More Chefstyle updates

lamont-granquist:
This should probably be reviewed only after #8708 goes in (or doesn't).

Associated chefstyle PR with the updates:  https://github.com/chef/chefstyle/pull/74


lamont-granquist:
* look at Layout/ClosingParenthesisIndentation for a different enforcement style
* add negatedunless

PR #8717, opened 2019-07-07
Add runit_service resource from runit cookbook

tas50:
Ship another common resource directly in Chef


tas50:
@mjingle The description fields for this new resource will be feeding the docs site so can you take a look to make sure they look good. They're just copied from the existing cookbook readme.

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :control, Array, default: [],
+               description: "An array of signals to customize control of the service, see runsv man page on how to use this. This requires that each template be created with the name sv-service_name-signal.erb."
+
+      property :env, Hash, default: {},
+               description: "A hash of environment variables with their values as content used in the service's env directory. Default is an empty hash. When this hash is non-empty, the contents of the runit service's env directory will be managed by Chef in order to conform to the declared state."
+
+      property :options, Hash, # deprecated: true should be set in the future
+        default: lazy { default_options }, coerce: proc { |r| default_options.merge(r) if r.respond_to?(:merge) },
+        description: "Options passed as variables to templates, for compatibility with legacy runit service definition. Default is an empty hash."
```
mjingle: The default in the description ("an empty hash") doesn't seem to match the default in line 56. Does something need to be modified?

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :log, [TrueClass, FalseClass], default: true,
+               description: "Whether to start the service's logger with svlogd, requires a template sv-service_name-log-run.erb to configure the log's run script."
```
mjingle:
```suggestion
               description: "Determines whether to start the service's logger with svlogd, which requires a template sv-service_name-log-run.erb to configure the log's run script."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :cookbook, String,
+        description: "A cookbook where templates are located instead of where the resource is used. Applies for all the templates in the enable action."
```
mjingle:
```suggestion
        description: "A cookbook where the templates are located, instead of where the resource is used. Applies for all the templates in the enable action."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :check, [TrueClass, FalseClass], default: false,
+               description: "Whether the service has a check script, requires a template sv-service_name-check.erb."
```
mjingle:
```suggestion
               description: "Determines whether the service has a check script, which requires a template sv-service_name-check.erb."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :start_down, [TrueClass, FalseClass], default: false,
+               description: "Set the default state of the runit service to 'down' by creating <sv_dir>/down file. Defaults to false. Services using start_down will not be notified to restart when their run script is updated."
+
+      property :delete_downfile, [TrueClass, FalseClass], default: false,
+               description: "Delete previously created <sv_dir>/down file."
```
mjingle:
```suggestion
               description: "Deletes previously created <sv_dir>/down file."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+               description: "Whether to start the service's logger with svlogd, requires a template sv-service_name-log-run.erb to configure the log's run script."
+
+      property :cookbook, String,
+        description: "A cookbook where templates are located instead of where the resource is used. Applies for all the templates in the enable action."
+
+      property :check, [TrueClass, FalseClass], default: false,
+               description: "Whether the service has a check script, requires a template sv-service_name-check.erb."
+
+      property :start_down, [TrueClass, FalseClass], default: false,
+               description: "Set the default state of the runit service to 'down' by creating <sv_dir>/down file. Defaults to false. Services using start_down will not be notified to restart when their run script is updated."
```
mjingle:
```suggestion
               description: "Sets the default state of the runit service to 'down' by creating <sv_dir>/down file. Defaults to false. Services using start_down will not be notified to restart when their run script is updated."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :delete_downfile, [TrueClass, FalseClass], default: false,
+               description: "Delete previously created <sv_dir>/down file."
+
+      property :finish, [TrueClass, FalseClass], default: false,
+               description: "Whether the service has a finish script, requires a template sv-service_name-finish.erb."
```
mjingle:
```suggestion
               description: "Determines whether the service has a finish script, which requires a template sv-service_name-finish.erb."
```

mjingle: Is the template supposed to be named "sv-service_name-finish.erb"? We might want to clarify that in this description and in the description in line 66.

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+        regex: [Chef::Config[:group_valid_regex]]
+
+      property :enabled, [TrueClass, FalseClass], default: false,
+               deprecated: true, skip_docs: true
+
+      property :running, [TrueClass, FalseClass], default: false,
+               deprecated: true, skip_docs: true
+
+      property :default_logger, [TrueClass, FalseClass], default: false,
+               description: "Whether a default log/run script should be set up. If true, the default content of the run script will use svlogd to write logs to /var/log/service_name."
```
mjingle:
```suggestion
               description: "Determines whether a default log/run script should be set up. If true, the default content of the run script will use svlogd to write logs to /var/log/service_name."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :restart_on_update, [TrueClass, FalseClass], default: true,
+               description: "Whether the service should be restarted when the run script is updated. Defaults to true. Set to false if the service shouldn't be restarted when the run script is updated."
```
mjingle:
```suggestion
               description: "Determines whether the service should be restarted when the run script is updated. Defaults to true. Set to false if the service shouldn't be restarted when the run script is updated."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :run_template_name, String, default: lazy { service_name },
+               description: "An alternate filename of the run run script to use instead of the value provided in service_name."
```
mjingle: "run run script"?

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+               description: "An alternate filename for the check script to use instead of the value provided in service_name."
+
+      property :finish_script_template_name, String, default: lazy { service_name },
+               description: "An alternate filename for the finish script to use instead of the value provided in service_name."
+
+      property :control_template_names, Hash, default: lazy { set_control_template_names },
+               description: "A hash of control signals (see control above) and their alternate template name(s) replacing service_name."
+
+      property :status_command, String, default: lazy { "#{sv_bin} status #{service_name}" },
+               description: "The command used to check the status of the service to see if it is enabled/running (if it's running, it's enabled). This hardcodes the location of the sv program to /usr/bin/sv due to the aforementioned cookbook load order."
```
mjingle:
```suggestion
               description: "The command used to check the status of the service to see if it is enabled/running. (If the service is running, it is enabled.) This hardcodes the location of the sv program to /usr/bin/sv due to the aforementioned cookbook load order."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :sv_templates, [TrueClass, FalseClass], default: true, description: "If true, the :enable action will create the service directory with the appropriate templates. Default is true. Set this to false if the service has a package that provides its own service directory. See Usage examples."
+
+      property :sv_timeout, Integer,
+        description: "Override the default sv timeout of 7 seconds."
+
+      property :sv_verbose, [TrueClass, FalseClass], default: false, description: "Whether to enable sv verbose mode. Default is false."
```
mjingle:
```suggestion
      property :sv_verbose, [TrueClass, FalseClass], default: false, description: "Determines whether to enable sv verbose mode. Default is false."
```

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+        description: "The directory where the svlogd log service will run. Used when default_logger is true."
+
+      property :log_flags, String, default: "-tt",
+               description: "The flags to pass to the svlogd command. Used when default_logger is true."
+
+      property :log_size, Integer,
+        description: "The maximum size a log file can grow to before it is automatically rotated. See svlogd(8) for the default value."
+
+      property :log_num, Integer,
+        description: "The maximum number of log files that will be retained after rotation. See svlogd(8) for the default value."
```
mjingle: I've noticed these properties and their metadata components sometimes run to one to three lines long in this file. Will that have an effect on resource auto-generation?

lib/chef/resource/runit_service.rb
```diff
@@ -0,0 +1,693 @@
...
+
+      property :log_min, Integer,
+        description: "The minimum number of log files that will be retained after rotation (if svlogd cannot create a new file and the minimum has not been reached, it will block).",
```
mjingle:
```suggestion
        description: "The minimum number of log files that will be retained after rotation. If svlogd cannot create a new file and the minimum has not been reached, it will block.",
```

lib/chef/resource/runit_service.rb
tas50: typo for sure

lib/chef/resource/runit_service.rb
tas50: That'll be fine. It just makes it harder for humans to read here

lib/chef/resource/runit_service.rb
tas50: Fixed

tas50:
This still has some issues that need to be worked out. Hold off on merging.

tas50:
Closing this for now

PR #8718, opened 2019-07-07
Improve generation of docs site resource pages

tas50:
1) Fix the indentation
2) Avoid extra new lines
3) Show example data if we have it
4) Chef Infra Client vs. the Chef Infra Client


tas50:
@kagarmoe I believe I have all the spacing working so that the docs that spit out don't have extra or missing blank lines. That should make comparing things a lot easier. Any chance you can read over he usage of Chef Infra Client here. I'm unsure of how we want to use that exactly. A chef infra client run vs. the chef infra client run.

PR #8738, opened 2019-07-10
Use apt not apt-update in kitchen tests

tas50:
We have it to use here


tas50:
This isn't the fix

Issue #8714, opened 2019-07-04
chef-solo: Unable to locate config file


vsingh-msys:
## Description
Unable to locate config file for chef-solo  `-c` or `--config` option

Reference: https://stackoverflow.com/questions/56745844/unable-to-locate-config-files-when-i-tried-to-run-chef-solo
## Chef Version

15.0.300 & latest 15.1.37
## Platform Version

Windows & Ubuntu

## Replication Case

`chef-solo -c solo.rb`

## Client Output

Ubuntu
```
[2019-07-04T14:41:28+05:30] WARN: *****************************************
[2019-07-04T14:41:28+05:30] WARN: Did not find config file: /solo.rb. Using command line options instead.
[2019-07-04T14:41:28+05:30] WARN: *****************************************
[2019-07-04T14:41:28+05:30] WARN: No cookbooks directory found at or above current directory.  Assuming /var/chef.
[2019-07-04T14:41:28+05:30] DEBUG: Reading products and relationships...
[2019-07-04T14:41:28+05:30] DEBUG: Successfully read products and relationships
```
Windows
````
[2019-07-04T13:56:13-07:00] WARN: *****************************************
[2019-07-04T13:56:13-07:00] WARN: Did not find config file: C:/solo.rb. Using command line options instead.
[2019-07-04T13:56:13-07:00] WARN: *****************************************
[2019-07-04T13:56:13-07:00] WARN: No cookbooks directory found at or above current directory.  Assuming C:/Users/vsingh-msys\.chef.
[2019-07-04T13:56:13-07:00] WARN: *****************************************
[2019-07-04T13:56:13-07:00] WARN: Did not find config file: C:/solo.rb. Using command line options instead.
[2019-07-04T13:56:13-07:00] WARN: *****************************************
````

lamont-granquist:
The important point from this one is likely that relative paths aren't being expanded so there needs to be some `File.expand_path()` rubbed on some of these variables somewhere.

lamont-granquist:
This is also likely a dup

vsingh-msys:
Yes, the path is being reset by `setup_application` method before reconfiguration.

https://github.com/chef/chef/blob/85914b32f9010aacca472f93b13784825a67102f/lib/chef/application/solo.rb#L222-L226

Here we need to call `reconfigure` then `setup_application`.

lamont-granquist:
This is related to:  #8138

lamont-granquist:
[...]

PR #8744, opened 2019-07-16
Move chef-client and chef-solo shared code into a base class and remove duplication and skew

lamont-granquist:
Does most of the easy part of cleaning up the shared code between client in solo affecting things which I'm confident enough are not behavior breaks.

There's issues here where `chef-solo --legacy-mode` has been broken for ages (local-mode `chef-solo` probably worked due to reparsing the ARGV in the Client class).


btm:
@lamont-granquist does this fix https://github.com/chef/chef/issues/8714?

lamont-granquist:
No, this is just prep for fixing #8714 and the rest.

PR #8754, opened 2019-07-18
dnf_package fixes for RHEL8

lamont-granquist:


jaymalasinha:
https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/83 passed 👍   Thank you for fixing this so quickly

tas50:
@lamont-granquist we should backport this to chef 14 as well

Issue #8722, opened 2019-07-08
--pre option isn't passed to "choco.exe list" when using choco_package resource


schaden-fred:
## Description
Using the chocolatey_package resource to deploy a pre-release package, the recipe fails.
Although the --pre option was specified in the resource, it's not passed on correctly when choco list is run.   The result is that pre-release packages are not shown.  If the only packages available are prerelease, then the recipe fails.

## Chef Version
14.3.37

## Platform Version
Windows Server 2019 Standard Evaluation
Running in Vagrant from gusztavvargadr/windows-server

## Replication Case

1. Create your own chocolatey server.  Build one or more prerelease packages (version number has a suffix with a hyphen and text) and upload them to the server.
2. Set your test machine to use this server.
3. Create a recipe with a resource to install the chocolatey package.
`chocolatey_package 'packagename' do`
` options '--pre'`
` action :upgrade`
`end`
4. Run chef on the client.

## Client Output
`         * chocolatey_package[packagename] action upgrade`
`           `
`          ` `================================================================================`
`           Error executing action `upgrade` on resource 'chocolatey_package[packagename]' `
` `  `================================================================================`
` `           
` `
`           Mixlib::ShellOut::ShellCommandFailed`
`           ------------------------------------`
`           Expected process to exit with [0], but received '2'`
`           ---- Begin output of C:\ProgramData\chocolatey/bin/choco.exe list -r packagename ----`
`           STDOUT: `
`           STDERR: `
`           ---- End output of C:\ProgramData\chocolatey/bin/choco.exe list -r packagename ----`
`           Ran C:\ProgramData\chocolatey/bin/choco.exe list -r packagename returned 2`

Stdout is blank because no packages were found.  Stderr is blank because according to chocolatey, nothing's gone wrong.

The last line shows the command that was run, and this is where the problem is.  Because we passed --pre as an option in our recipe, the command should be 
`C:\ProgramData\chocolatey/bin/choco.exe list -r --pre packagename`

## Workaround
1. Don't build prerelease packages.
2. Replace the chocolatey_package resource with a powershell_script resource that runs Chocolatey directly.

vsingh-msys:
Fixed in #8765

schaden-fred:
> Fixed in #8765

Thanks @vsingh-msys !

PR #8765, opened 2019-07-23
Fix for chocolatey_package fails using extra options

### LINKED ISSUE #8437: chocolatey_package fails using extra options
## Description
#8230 added options parameter while fetching list of available packages.  Some options that are allowed for choco install fail when running choco list.

## Chef Version
Showed up in 14.11.21, our TK dev instances saw this as recently as 2019-04-26.

## Platform Version
Windows, chef-client 14.11.21+

## Replication Case
```
chocolatey_package '<package-name>' do
 options '--install-arguments=NO_START=1'
end
```

## Client Output
From user report, if we need to add more, let us know.
```
No candidate version available for <package-name>
```

## Stacktrace
N/A

kapilchouhan99:

## Description
- Added method `common_options(include_source: true)` and three properties `list_options`, `user` and `password`
- In the chocolatey package if we have the feature `useEnhancedExitCodes` turned on, at that time choco will provide enhanced exit codes:
   ```
    0: operation was successful, no issues detected
    –1 or 1: an error has occurred
    2: no results (enhanced)
   ```
   - If the feature is not enabled its provide normally Exit codes:
   ```
    0: operation was successful, no issues detected
    –1 or 1: an error has occurred
   ```
   - So we have handled the exception( 2: no results (enhanced) ) by setting the default value of property `returns` as [0, 2] instead of [0].
- Added test cases
## Related Issue
Fixes #8437

## Checklist:

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -210,7 +210,7 @@ class Chef
         # @return [String] options from new_resource or empty string
         def cmd_args(include_source: true)
           cmd_args = [ new_resource.options ]
-          cmd_args.push([ "-source", new_resource.source ]) if new_resource.source && include_source
+          cmd_args += common_options(include_source: include_source)
           cmd_args
         end
 
@@ -227,8 +227,8 @@ class Chef
             available_versions =
               begin
                 cmd = [ "list", "-r", pkg ]
-                cmd.push( [ "-source", new_resource.source ] ) if new_resource.source
-                cmd.push( new_resource.options ) if new_resource.options
+                cmd += common_options
+                cmd.push( new_resource.list_options ) if new_resource.list_options
 
                 raw = parse_list_output(*cmd)
                 raw.keys.each_with_object({}) do |name, available|
@@ -271,6 +271,14 @@ class Chef
         # @return [Array] same names in lower case
         def lowercase_names(names)
           names.map(&:downcase)
+        end
+
+        def common_options(include_source: true)
+          args = []
+          args.push( [ "-source", new_resource.source ] ) if new_resource.source && include_source
+          args.push( [ "--user", new_resource.user ] ) if new_resource.user
+          args.push( [ "--password", new_resource.password ]) if new_resource.password
+          args
         end
       end
     end
--- a/lib/chef/resource/chocolatey_package.rb
+++ b/lib/chef/resource/chocolatey_package.rb
@@ -33,6 +33,18 @@ class Chef
       property :options, [String, Array],
         description: "One (or more) additional options that are passed to the command."
 
+      property :list_options, String,
+        introduced: "15.3",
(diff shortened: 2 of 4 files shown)


kapilchouhan99:
@btm @tas50 
In the choco, if we have the feature `useEnhancedExitCodes` turned on, then choco will provide enhanced exit codes(2: no results).
Choco exit codes https://chocolatey.org/docs/commandsinfo#exit-codes

Here we have 2 approaches to fix this:
1.) If we get exit status as 2, at that time we can raise a custom error like: "Package does not exist."
2.) Or we can skip it instead of raising an error on exit status 2, we can set the default value of property returns as [0, 2] instead of [0]

Please let me know your thoughts on this.

nimDevOps:
My experience with https://github.com/chef/chef/issues/8437 is not about handling error code 2, but mostly splitting package parameters into two subsets:

1. Parameters applicable for "choco list" command like package Id and version;
2. Parameters applicable for package itself, for example for **git.install**:
 **--packageParameters="/NoAutoCrlf /GitAndUnixToolsOnPath"**
should not be sent to Choco Repository server;

I would say even more: only Id, version and new parameters used with authentication or whatever chef can recognize for this enhancement should be used with **choco list** command.

lib/chef/provider/package/chocolatey.rb
```diff
@@ -282,6 +283,25 @@ def parse_list_output(*args)
...
+            password = options.include?("-p") ? find_credential("-p", options) : find_credential("--password", options)
+          end
+          [username, password].compact.join(" ").strip
+        end
+
+        def find_credential(element, array)
+          index = array.find_index(element) + 1
+          next_element = array.at(index)
+          next_element ? element + " " + next_element : nil
+        end
```
lamont-granquist: This now feels like we're going in the wrong direction trying to solve this problem.

We probably need to back up and add `new_resource.list_options` as something that users can set which is separate to `new_resource.options`.  For common things like user and password those should go onto additional `new_resource.user` and `new_resource.password` fields.

I could potentially be persuaded to avoid the duplication between `list_options` and `options` but then `options` should get passed to both, and any options which cannot apply to both then must be made into properties on the new_resource.

What I don't know is what the overlap looks like between the different options that those CLI commands support.  If the union is large, then users probably want `options` passed to both, if the union is small then users probably don't.  Based on the code here it looks like the union is small so that users probably want `options` to only go to the main install or remove command and that `new_resource.list_options` should be a separate.

Either way though we should just use properties instead of doing this parsing out of the options argument.  The creds are important enough they should be top-level properties.

lamont-granquist:
so https://github.com/chef/chef/pull/7712 is closer to the correct direction to take this fix.  that's more what i'm talking about with suggesting user and password properties be explicit on the resource.

https://github.com/chef/chef/issues/8722 is also related to this problem.  we seem to have a general problem with some options that need to go to list and some options that can't.

PR #8774, opened 2019-07-24
Fix zypper test failures

jaymalasinha:
Cloud testers have  as many as 10 zypper repos configured and enabled, so the zypper searches from specs take longer to run on our testers as too many repos with most of them enabled were slowing down the zypper process, locking it during the test runs and failing the zypper package functional tests.  Since we do not use these repos in our testing we delete them.

Signed-off-by: Jaymala Sinha <jsinha@chef.io>

## Description
Resolves failing zypper package test on our infra testers

## Types of changes

## Checklist:

### DIFF
--- a/omnibus/omnibus-test.sh
+++ b/omnibus/omnibus-test.sh
@@ -49,7 +49,7 @@ sudo rm -rf "$TMPDIR"
 mkdir -p "$TMPDIR"
 
 # Verify that we kill any orphaned test processes. Kill any orphaned rspec processes.
-ps ax | grep -E 'rspec' | grep -v grep | awk '{ print $1 }' | xargs sudo kill -s KILL || true
+sudo kill -9 $(ps ax | grep 'rspec' | grep -v grep | awk '{ print $1 }') || true
 
 PATH="/opt/$product/bin:$PATH"
 export PATH
@@ -150,6 +150,18 @@ if [[ -d /etc/sudoers.d ]]; then
 elif [[ -d /usr/local/etc/sudoers.d ]]; then
   echo "Defaults:$(id -un) !secure_path, exempt_group += $(id -gn)" | sudo tee "/usr/local/etc/sudoers.d/$(id -un)-preserve_path"
   sudo chmod 440 "/usr/local/etc/sudoers.d/$(id -un)-preserve_path"
+fi
+
+# Remove the ec2 cloud zypper repos before running functional tests
+if [[ ! $(grep s390 /etc/SuSE-release) ]] > /dev/null 2>&1; then   # Don't delete repos on s390x
+  if [[ $(grep -i suse /etc/SuSE-release /etc/os-release) ]] > /dev/null 2>&1; then
+    echo "Removing zypper repos for Amazon SLES nodes"
+    sleep 5  # Allow some time for repo setup to complete
+    for i in $(zypper lr |grep -v "Alias" |awk -F "|" '{print $2}'); do
+      sudo zypper rr $i
+    done
+    sudo rm /usr/lib/zypp/plugins/services/* || true
+  fi
 fi
 
 cd "$chef_gem"


jaymalasinha:
https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/86 passed.

omnibus/omnibus-test.sh
```diff
@@ -152,6 +152,18 @@ elif [[ -d /usr/local/etc/sudoers.d ]]; then
   sudo chmod 440 "/usr/local/etc/sudoers.d/$(id -un)-preserve_path"
 fi
 
+# Remove the cloud zypper repos before running functional tests
+if [[ -f /etc/SuSE-release ]] || [[ -f /etc/os-release ]] && [[ ! $(grep s390x /etc/SuSE-release /etc/os-release) ]]; then
+  if [[ -f /etc/os-release ]] && [[ ! $(grep -i suse  /etc/os-release) ]]; then
```
lamont-granquist: this conditional seems a bit awkward and potentially brittle and could definitely use a comment or two at least.

- the s390x side of the conditional seems to be testing for our testing infrastructure rather than testing the zypper config that makes those nodes different.  can we get so that we're not treating s390x as a special case?

- since you've got mixed || and &&'s in the first conditional you should really use parens there for clarity

- the sense of the comparisons should be flipped around so that the `else` block becomes the body of the `if` block, the implicit negation is making it harder to read.

it looks like what this says is that if you have /etc/SuSE-release but not /etc/os-release" that you always prune the zypper repos, but that if you have /etc/os-release and it is suse (or really not-not-suse) and is not 390x then you prune it as well?

```
if [[ -f /etc/SuSE-release ]] || ( [[ -f /etc/os-release ]] && [[ ! $(grep s390x /etc/SuSE-release /etc/os-release) ]] && [[ $(grep -i suse  /etc/os-release) ]]; then
    echo "Removing zypper repos for Amazon SLES nodes"
    for i in $(zypper lr |grep -v "Alias" |awk -F "|" '{print $2}'); do
      sudo zypper rr $i
    done
fi
```

maybe that, unless i'm reading it wrong...

jaymalasinha:
@lamont-granquist I simplified the conditions to filter out s390x nodes.  Also added code comments.

omnibus/omnibus-test.sh
jaymalasinha: @lamont-granquist I simplified the conditions to filter out s390x nodes. Also added code comments.

jaymalasinha: https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/87 in progress to test these changes

omnibus/omnibus-test.sh
```diff
@@ -49,7 +49,7 @@ sudo rm -rf "$TMPDIR"
 mkdir -p "$TMPDIR"
 
 # Verify that we kill any orphaned test processes. Kill any orphaned rspec processes.
-ps ax | grep -E 'rspec' | grep -v grep | awk '{ print $1 }' | xargs sudo kill -s KILL || true
+sudo kill -9 $(ps ax | grep -E 'rspec' | grep -v grep | awk '{ print $1 }') || true
```
lamont-granquist: I think `$()` is technically a bash-ism and might fail on older Unixen?  Do we care any more?

Sol9 this certainly wouldn't work on...

lamont-granquist: `ps ax` also fails on solaris i think since BSD-vs-SysV flag syntax, although maybe by Sol11 they've finally supported both?

omnibus/omnibus-test.sh
```diff
@@ -152,6 +152,16 @@ elif [[ -d /usr/local/etc/sudoers.d ]]; then
...
 
+# Remove the ec2 cloud zypper repos before running functional tests
+if [[ ! $(grep s390 /etc/SuSE-release) ]] > /dev/null 2>&1; then   # Don't delete repos on s390x
+  if [[ $(grep -i suse /etc/SuSE-release /etc/os-release) ]] > /dev/null 2>&1; then
+     echo "Removing zypper repos for Amazon SLES nodes"
+    for i in $(zypper lr |grep -v "Alias" |awk -F "|" '{print $2}'); do
+      sudo zypper rr $i
+    done
+  fi
+fi
```
lamont-granquist: okay this is better although we're still using "is_390x?" as a proxy for "is_not_in_ec2?" but its a lot easier to see it now...

lamont-granquist:
i think i'm good with this as long as i'm reassured that we're sufficiently advanced in technology now that `$()` works everywhere...

omnibus/omnibus-test.sh
jaymalasinha: good point will check this on Solaris now

jaymalasinha: The `$()` works on solaris, but `grep -E ` has been failing until now. Removing `-E` here

jaymalasinha:
@lamont-granquist @tas50 https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/92 passed, this is ready for merge.

Issue #7763, opened 2018-10-19
chef user resource for macOS is not compatible with macOS 10.14 security protections


pudquick:
## Description
The user resource as currently designed for macOS performs operations which are blocked in 10.14 by default.

## Chef Version
This affects all versions, but the examples I'll give here are for 14.3.37

## Platform Version
macOS 10.14 (Mojave) build 18A391

## Replication Case
Create a user resource:

```
user '_example' do
  comment 'example user'
  system true
  shell '/usr/bin/false'
  home '/var/empty'
  action :create
end
```

Run chef once. No errors will occur. Run it again, chef will terminate with an error.

## Client Output
```
================================================================================
      Error executing action `create` on resource 'dscl_user[_example]'
================================================================================
      
      Chef::Exceptions::DsclCommandFailed
      -----------------------------------
      dscl error: <Mixlib::ShellOut#70108986308720: command: '["dscl", ".", "-create", "/Users/_example", "UniqueID", "206"]' process_status: #<Process::Status: pid 8590 exit 40> stdout: '' stderr: '<main> attribute status: eDSPermissionError
      <dscl_cmd> DS Error: -14120 (eDSPermissionError)' child_pid: 8590 environment: {"LC_ALL"=>"en_US.UTF-8", "LANGUAGE"=>"en_US.UTF-8", "LANG"=>"en_US.UTF-8", "PATH"=>"[clipped for brevity]"} timeout: 600 user:  group:  working_dir:  >

...

    System Info:
    ------------
    chef_version=14.3.37
    platform=mac_os_x
    platform_version=10.14
    ruby=ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin15]
    program_name=/usr/local/bin/chef-client
    executable=/opt/chef/bin/chef-client

```

pudquick:
Root cause for this lies in several behaviors the chef resource performs which are now denied by default on macOS 10.14.

For example: 
https://github.com/chef/chef/blob/master/lib/chef/provider/user/dscl.rb#L592-L594

```
            user_plist_file = "#{USER_PLIST_DIRECTORY}/#{new_resource.username}.plist"
            user_plist_info = run_plutil("convert", "xml1", "-o", "-", user_plist_file)
            user_info = Plist.parse_xml(user_plist_info)
```

[...]

PR #8775, opened 2019-07-24
Add mac_user resource that is compatible with macOS >= 10.14

ryancragun:
Provide a user resource that is compatible with default TCC restrictions
that were introduced in macOS 10.14.

## Description
Changes:

* This resource and the corresponding provider have been modified to
  work with default macOS TCC policies. Direct access to user binary
  plists are no longer permitted by default, thus we've chosen to use
  use newer methods of creating, modifying and removing users.

* Due to the tooling required by the provider this provider is only
  suitable for use on macOS >= 10.14. Support for older platforms has
  been removed.

New Features:

* Primary group management is now included.

* 'admin' is now a boolean property that configures a user to an admin.

* 'admin_username' and 'admin_password' are new properties that define the
  admin user credentials required for toggling SecureToken for an
  exiting user.

  The 'admin_username' must correspond to a system admin with SecureToken
  enabled in order to toggle SecureToken.

* 'secure_token' is a boolean property that sets the desired state
  for SecureToken. SecureToken token is required for FileVault full
  disk encryption.

## Related Issue
https://github.com/chef/chef/issues/7763

## Types of changes

## Checklist:


lib/chef/resource/user/mac_user.rb
```diff
@@ -0,0 +1,105 @@
...
+      #   for SecureToken. SecureToken token is required for FileVault full
+      #   disk encryption.
+      class MacUser < Chef::Resource::User
+        resource_name :mac_user
+
+        provides :mac_user
+        provides :user, os: "darwin", platform_version: ">= 10.14"
+
+        property :iterations, Integer,
+          description: "macOS platform only. The number of iterations for a password with a SALTED-SHA512-PBKDF2 shadow hash.",
```
lamont-granquist: you probably don't have to comment that its "macOS platform only" here since its in the mac_user resource.

lib/chef/resource/user/mac_user.rb
```diff
@@ -0,0 +1,105 @@
...
+          default: 57803, desired_state: false
+
+        # Overload gid so we can set our default. NilClass is for backwards compat
+        # and 20 is the macOS "staff" group.
+        property :gid, [String, Integer, NilClass], description: "The numeric group identifier.", default: 20, coerce: ->(gid) do
```
lamont-granquist: It looks like this is `property :gid, Integer` since you can't assign a nil and strings get coerced, and the validation should apply to after all coercion.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+
+        def define_resource_requirements
+          super
+
+          %w{dscl sysadminctl plutil dsimport}.each do |bin|
+            requirements.assert(:all_actions) do |a|
+              a.assertion { which(bin) }
+              a.failure_message(Chef::Exceptions::User, "Cannot find binary '#{bin}' on the system for #{new_resource}!")
+            end
+          end
```
lamont-granquist: is this really something that you ever reasonably expect to fail?

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+          requirements.assert(:all_actions) do |a|
+            a.assertion do
+              if new_resource.property_is_set?(:salt)
+                new_resource.property_is_set?(:iterations) && new_resource.password =~ /^[[:xdigit:]]{256}$/
+              else
+                true
+              end
+            end
+            a.failure_message(Chef::Exceptions::User, "Password must be a SALTED-SHA512-PBKDF2 shadow hash entropy when a shadow hash salt is given")
+          end
```
lamont-granquist: this way of expressing requirements is really for why-run mode when some state is being read from the system that might be changed by the chef client run.

for just validating the inputs of properties you should really just test them and fail, because there's nothing about this check that is going to get fixed by not running in why-run mode since the properties are pretty static.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+        end
+
+        def load_current_resource
+          convert_group_name if new_resource.property_is_set?(:gid)
```
lamont-granquist: looks like this method doesn't exist and you likely moved this to the coercion in the resource but forgot to delete it here?

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+
+          @current_resource = Chef::Resource::User::MacUser.new(new_resource.username)
+          current_resource.username(new_resource.username)
+
+          @user_plist = load_user
+          if user_plist
+            reload_with_user_plist
+          else
+            @user_exists = false
```
lamont-granquist: user_exists should really be set to true on the other side of this, which i certainly can't easily see if it is or not.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+          shadow_hash_hex = dscl_get(user_plist, :shadow_hash)[0]
+
+          return user_plist unless shadow_hash_hex && shadow_hash_hex != ""
+
+          shadow_binary_plist = [shadow_hash_hex.delete(" ")].pack("H*")
+          shadow_xml_plist = shell_out("plutil", "-convert", "xml1", "-o", "-", "-", input: shadow_binary_plist).stdout
+          dscl_set(user_plist, :shadow_hash, Plist.parse_xml(shadow_xml_plist))
+
+          user_plist
+        rescue Chef::Exceptions::PlistUtilCommandFailed, Chef::Exceptions::DsclCommandFailed
```
lamont-granquist: it might be clearer to move these exceptions to wrap only the call which can throw them, because i think they're coming from different calls.  makes for more code, but it makes it clearer what can fail and where.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+          unless res.downcase =~ /creating user/
+            raise Chef::Exceptions::User, "error when creating user: #{res}"
+          end
+
+          # Wait for the user to be created and for dscl to flush everything
+          sleep 3
+
+          # Reload with up-to-date user information
+          @user_plist = load_user
+          reload_with_user_plist
```
lamont-granquist: you shouldn't be updating the current_resource here, that has to be the 'before' state and left the way that load_current_resource found it, or else you break reporting.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+        #
+        # User Provider Callbacks
+        #
+
+        def create_user
+          cmd = [-"-addUser", new_resource.username]
+          cmd += ["-fullName", new_resource.comment]  if new_resource.property_is_set?(:comment)
+          cmd += ["-UID", new_resource.uid]           if new_resource.property_is_set?(:uid)
+          cmd += ["-shell", new_resource.shell]       if new_resource.property_is_set?(:shell)
+          cmd += ["-home", new_resource.home]         if new_resource.property_is_set?(:home)
```
lamont-granquist: for some of these like home directory you set a default and don't allow setting it to nil, so the `if new_resource.property_is_set?(:home)` check can likely be omitted, because it will always have a valid setting that should be respected.

but even for properties that could be nil you should avoid the `property_is_set?` check and just use `if new_resource.whatever` or `unless new_resource.whatever.nil?` (if its boolean-valued).

lamont-granquist: (generally you probably don't want to be using `property_is_set?` at all because like i said offhand i can't recall if a default value is 'set' or not and if i have to read the code to determine how it is implemented and how close to `nil?` the check is or not then you're likely asking too much of the reader in most cases)

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+          # sysadminctl doesn't exit with a non-zero exit code if it encounters
+          # a problem. We'll check stderr and make sure we see that it finished
+          # correctly.
+          res = run_sysadminctl(cmd)
+          unless res.downcase =~ /creating user/
+            raise Chef::Exceptions::User, "error when creating user: #{res}"
+          end
+
+          # Wait for the user to be created and for dscl to flush everything
+          sleep 3
```
lamont-granquist: if you can figure out a way to poll a bit more tightly as to when its really done instead of just a large sleep value that'd both be more correct under high load / low cpu situations and would be faster in the common case.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+
+          @user_exists = true
+        end
+
+        def compare_user
+          %i{comment shell uid gid salt password admin secure_token}.each do |m|
+            return true if diverged?(m)
+          end
+
+          false
```
lamont-granquist: `%i{comment shell uid gid salt password admin secure_token}.any? { |m| diverged?(m) }`

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+            # PrimaryGroupID for the user and move on if we decide that actual
+            # group magement should be done outside of the core resource.
+            group_name, group_id, group_action = user_group_info
+
+            declare_resource(:group, group_name) do
+              members new_resource.username
+              gid group_id if group_id
+              action group_action
+              append true
+            end
```
lamont-granquist: you probably want to `action :nothing` this one and then `.run_action(:create)` or whatever to force it to compile time.  you have a lot of converge_by blocks in here and this resource will run only after all of those complete.

lamont-granquist: yeah `run_dscl` uses a `shell_out` and not an execute resource, so you're going to want to pick your poison and either do everything with chef sub-resources, including ruby_blocks where necessary, or else you're going to want to take all the sub-resources and force them to compile time.

this would be a good example of where #8668 would be useful.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+          end
+
+          # Other fields are have been set on current resource so just compare
+          # them.
+          new_resource.property_is_set?(prop) && (new_resource.send(prop) != current_resource.send(prop))
+        end
+
+        def user_group_info
+          @user_group_info ||= begin
+            if new_resource.gid.is_a?(String)
```
lamont-granquist: similarly to other comments i think this now never happens due to coercion.

lib/chef/provider/user/mac.rb
```diff
@@ -0,0 +1,607 @@
...
+
+          result.stdout
+        end
+
+        def run_plutil(*args)
+          result = shell_out("plutil", "-#{args[0]}", args[1..-1])
+          raise(Chef::Exceptions::PlistUtilCommandFailed, "plutil error: #{result.inspect}") unless result.exitstatus == 0
+
+          if result.stdout.encoding == Encoding::ASCII_8BIT
+            result.stdout.encode("utf-8", "binary", undef: :replace, invalid: :replace, replace: "?")
```
lamont-granquist: did this one actually happen? because chef-client does pretty well at forcing itself to utf-8 mode, and unless there was a truncated read (which there shouldn't be from shell_out) it should never return 8-bit encoding.

lib/chef/provider/user/mac.rb
lamont-granquist: yeah i'm a bit confused, i was gonna say that this should be `diverged?(:home)` and push the `property_is_set?` consistently into that API, but this is creating where you don't want to worry about "is it diverged, and did the user want to fix it?" but instead you just want to create the user, so I think you want to just always set the home here -- otherwise you won't be calling with the default you specified in the resource (which might work but smells buggy one way or another).

lib/chef/resource/user/mac_user.rb
ryancragun: Yar, it was copy-pasta'd from the `dscl` resource but has been updated.

lib/chef/resource/user/mac_user.rb
ryancragun: I allowed `String` because if group lookup failed during compile time then we'd allow try resolving the group again in the provider in case a `group` resource had converged and created the group we want before our `user` resource.

Seems to me the issue is that previously we didn't do group management, so if GID resolution happened once and failed we could raise an error because a group _should_ have been created before the user which you're trying to associate it with. But now we do group management so we have to assume a user is giving us the GID or group name they want us to handle, regardless of whether or not it exists at compile time or converge time.

I've convinced myself that regardless or whether or not we want to keep group management, the property probably need to be both an `Integer` or `String` and that we should only resolve it a single time in the provider, not with coercion. The reason we'll want to do it later is to handle the case where a prior `group` resource creates a group that doesn't exist at compile time. If we used coercion to resolve it at compile time that would surely fail.

We have two options after that:

* get rid of group management. Try and resolve the group name in the provider and raise an error if it doesn't exist, associate the primary ID.
* keep group management, create the group if it doesn't resolve, append the user to the group, associate the primary ID.

I went with the second option.

lib/chef/provider/user/mac.rb
ryancragun: I haven't tested it on platforms that aren't 10.14, but I would expect this resource and provider to _mostly_ work on 10.12 -> 10.14, aside from the SecureToken bits. If people specify  `mac_user` on <= 10.11 some of those probably won't be available. I also have absolutely no idea what strange ways Apple is going to change things for 10.15 either, so I was kind of trying to future proof for that case. That said, I know there are some incompatibilities with this and the dev preview of 10.15 that I'm trying to work out, but I'm not there yet.

On the only targeted platform for this resource (10.14) those binaries should almost surely exist. I'm happy to remove the check if you think it's adding nothing.

lib/chef/provider/user/mac.rb
ryancragun: I moved the validation to the resource.

lib/chef/provider/user/mac.rb
ryancragun: It's inherited from `Chef::Provider::User`. Like I mentioned in the comment above, we should probably only resolve the group name in a single place and I think it's here in the provider. Still, since we're managing the group we shouldn't raise an error if the group isn't resolvable so I've removed that method.

lib/chef/provider/user/mac.rb
ryancragun: It defaults to `true` in `Chef::Provider::User` which we've inherited. We can set it again if it makes it more clear but I think we only need to worry about setting it for false conditions.

lib/chef/provider/user/mac.rb
ryancragun: Hrm, that's a bummer. After we create the user some of these properties can be magically changed depending on whether or not we're running `adminUser` or not, and whether or not they want SecureToken. By remapping I was able to avoid "reconverging them" by manual means if the previous command had already done so.

lib/chef/provider/user/mac.rb
ryancragun: Ironically, the original pass at this used all sub-resources and then I changed it to just to the mixture :man_facepalming:

lib/chef/provider/user/mac.rb
ryancragun: I copy-pasta'd this from the `dscl` provider. Removing it things worked just fine.

lib/chef/provider/user/mac.rb
ryancragun: I removed all updating of `current_resource` after `load_current_resource` and now update a different instance variable(s).

lib/chef/provider/user/mac.rb
lamont-granquist: yeah defensive coding that won't ever get exercised is really just cognitive overhead for anyone reading the code.

i also have a YAGNI instinct now when it comes to future-proofing...

lib/chef/resource/user/mac_user.rb
lamont-granquist: cool that makes sense.

lib/chef/provider/user/mac.rb
lamont-granquist: yeah you gotta pick one or the other, the mixture crosses the streams.  if you can do it without any converge_by and use sub-resources entirely that's probably the best approach.  but then you still need to think about compile/converge mode.  when people start using converge_by because they want control over the output formatting and think that converge_by is equivalent to a "log" statement is when things usually get real weird...

lib/chef/resource/user.rb
```diff
@@ -47,6 +47,7 @@ class User < Chef::Resource
 
       property :password, String,
         description: "The password shadow hash",
+        sensitive: true,
```
tas50: How did we not have this set?

lib/chef/resource/user/mac_user.rb
```diff
@@ -0,0 +1,117 @@
...
+      # * 'secure_token_password' is the plaintext password required to enable
+      #   or disable secure_token for a user. If no salt is specified we assume
+      #   the 'password' property corresponds to a plaintext password and will
+      #   attempt to use it in place of secure_token_password if it not set.
+      class MacUser < Chef::Resource::User
+        resource_name :mac_user
+
+        provides :mac_user
+        provides :user, os: "darwin", platform_version: ">= 10.14"
+
```
tas50: Can you stick this in here below provides on its own

```ruby
introduced "15.3"
```

PR #8776, opened 2019-07-25
Tweak data collector exception handling

### LINKED ISSUE #8749: raise_on_failure = false config doesn't catch internal server errors
## Description
In situations where there is an Automate server but it is broken due to high loads, chef-client will fail on attempting to post to data_collector even if data_collector.raise_on_failure is set to false. This looks to be due to the fact that we don't rescue on Net::HTTPInternalServerError here.
https://github.com/chef/chef/blob/7558c414030d082194d0d2b5f74279ce151517e9/lib/chef/data_collector.rb#L180-L183

This results in a failed chef run with a stacktrace on `500 "Internal Server Error"` from Automate.

## Chef Version
verified on chef-13.6.4

## Platform Version
Seen on Windows Server, but I don't think it's OS specific.

## Replication Case
You can use chef-load to overload an Automate server similar to what I describe here.
https://github.com/chef/automate/issues/924
At which point, client connections will start failing with 500 errors. Which will fail chef-client runs.

lamont-granquist:
This isn't Java and Net::HTTP can throw a billion things, so just rescue
    (nearly) everything.  I have a hard time imagining that there's a
    failure here that we wouldn't want to ignore, other than the ones like
    OOM thrown by Exception that indicate internal failures.

    Also only throw the "this is normal" info message on 404s, other
    responses like 500s should be warns even if it is set to ignore
    failures (people with data collectors correctly setup need to know about
    500s even if we keep going, since that is abnormal).

    This also fixes a bug in handling exceptions like Timeout::Error which
    do not have a response.code at all which would throw NoMethodError.

closes #8749

### DIFF
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -198,7 +198,7 @@ class Chef
           if code == "404"
             # Make the message non-scary for folks who don't have automate:
             msg << " (This is normal if you do not have #{Chef::Dist::AUTOMATE})"
-            Chef::Log.info(msg)
+            Chef::Log.debug(msg)
           else
             Chef::Log.warn(msg)
           end
--- a/lib/chef/data_collector.rb
+++ b/lib/chef/data_collector.rb
@@ -177,14 +177,17 @@ class Chef
 
         @http ||= setup_http_client(Chef::Config[:data_collector][:server_url])
         @http.post(nil, message, headers)
-      rescue Timeout::Error, Errno::EINVAL, Errno::ECONNRESET,
-        Errno::ECONNREFUSED, EOFError, Net::HTTPBadResponse,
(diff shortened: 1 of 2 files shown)


lib/chef/data_collector.rb
```diff
@@ -177,24 +177,31 @@ def send_to_data_collector(message)
...
         msg = "Error while reporting run start to Data Collector. URL: #{Chef::Config[:data_collector][:server_url]} Exception: #{code} -- #{e.message} "
 
         if Chef::Config[:data_collector][:raise_on_failure]
           Chef::Log.error(msg)
           raise
         else
-          # Make the message non-scary for folks who don't have automate:
-          msg << " (This is normal if you do not have #{Chef::Dist::AUTOMATE})"
-          Chef::Log.info(msg)
+          if code == "404"
```
tas50: This is great.

PR #8781, opened 2019-07-27
Property: Reorder comparison with NOT_PASSED.

ab:
The original comparison `value == NOT_PASSED` ends up calling `.<=>` on
`value` under the hood. This relies on the implementation of `<=>`/`cmp`
not being broken in that class.

Unfortunately, some classes like OpenSSL::X509::Name have broken
implementations of `<=>` that raise TypeError instead of returning nil
for mismatched types. https://github.com/ruby/openssl/pull/264

As a result, currently, if you attempt to set a resource property to an
OpenSSL::X509::Name object, the result is a TypeError.

```ruby
>> require 'openssl'
>> require './lib/chef/constants.rb'
>> name = OpenSSL::X509::Name.new

>> name == Chef::NOT_PASSED
TypeError: wrong argument type Object (expected OpenSSL/X509/NAME)
>> Chef::NOT_PASSED == name
=> false
```

With this change, we use the implementation of comparison from the
NOT_PASSED object instead, which we know will work no matter what.

Signed-off-by: Andy Brody <git@abrody.com>

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -347,7 +347,7 @@ class Chef
     #   `get`, the non-lazy, coerced, validated value will always be returned.
     #
     def call(resource, value = NOT_PASSED)
-      if value == NOT_PASSED
+      if NOT_PASSED == value
         get(resource)
       else
         set(resource, value)


lamont-granquist:
There's probably at least a couple dozen more cases of this and somehow this needs at least a comment so it doesn't get forgotten that this is why it is compared in this order.

tas50:
Thanks @ab for the solid fix and a great explanation

ab:
Another option would be to use `NOT_PASSED.equal?(value)`. The `Object#equal?` method checks for exact object equality, and is unlikely to be overridden by subclasses regardless of which order it's written in. That also makes it more obvious that the order is significant.

PR #8777, opened 2019-07-26
Add raise an error on Chef 14 when trying to bootstrap Chef 15

vsingh-msys:
## Description

- Validate only if chef version is less than 15.
- If the chef version is 14 or lower add validation based on `bootstrap_version` & `prerelease` 

Note: This fixes is mainly for chef-14 so need to back-port it into chef-14.

## Related Issue

https://github.com/chef/chef/issues/8623

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [x] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Signed-off-by: Vivek Singh <vivek.singh@msystechnologies.com>

Issue #8623, opened 2019-05-30
Raise an error on Chef 14 when trying to bootstrap Chef 15


btm:
### Describe the Enhancement:
`knife bootstrap` on Chef 14 should check what version we're going to install and if it is >= 15 we should raise an error.

### Describe the Need:
The license acceptance features of Chef 15 would be difficult to backport to Chef 14 given the refactoring and improvements to `knife bootstrap` in Chef 15. We should raise an error that says "You must use Chef 15 or later to bootstrap Chef 15 nodes"

### Current Alternative
It currently breaks and doesn't work.

shindemahendra:
Add raise an error on Chef 14 when trying to bootstrap Chef 15 #8777

Chef-14: Raise an error on Chef 14 when trying to bootstrap Chef 15 #8782

PR #8782, opened 2019-07-29
Chef-14: Raise an error on Chef 14 when trying to bootstrap Chef 15

vsingh-msys:
## Description

- Validate only if chef version is less than 15.
- If the chef version is 14 or lower add validation based on bootstrap_version & prerelease.
- PR against chef-14 only.

Referenced PR: https://github.com/chef/chef/pull/8777

## Related Issue

https://github.com/chef/chef/issues/8623

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -22,6 +22,7 @@ require "erubis"
 require "chef/knife/bootstrap/chef_vault_handler"
 require "chef/knife/bootstrap/client_builder"
 require "chef/util/path_helper"
+require_relative "../version_string"
 
 class Chef
   class Knife
@@ -517,7 +518,7 @@ class Chef
       def bootstrap_version_gte_15?
         config[:prerelease] ||
           (!!config[:bootstrap_version] &&
-           (config[:bootstrap_version] == "latest" || config[:bootstrap_version].to_i >= 15))
+           (config[:bootstrap_version] == "latest" || Chef::VersionString.new(config[:bootstrap_version]) >= 15.0))
       end
     end
   end
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -28,6 +28,7 @@ class Chef
     class Bootstrap < Knife
       include DataBagSecretOptions
 
+      CHEF_15 ||= 15
       attr_accessor :client_builder
       attr_accessor :chef_vault_handler
 
@@ -366,6 +367,7 @@ class Chef
 
         validate_name_args!
         validate_options!
+        validate_bootstrap_version_options!
 
         $stdout.sync = true
 
@@ -429,6 +431,22 @@ class Chef
           ui.error("Policyfile options and --run-list are exclusive")
           exit 1
         end
+        true
+      end
+
+      # Ensure options are valid by checking bootstrap version values.
+      #
+      # The method call will cause the program to exit(1) if:
+      #   * Bootstrap version is greater than or equal to 15 for chef 14 or lower
+      #   * Pre-release options is passed for chef 14 or lower
+      #
+      # @return [TrueClass] If options are valid.
+      def validate_bootstrap_version_options!
+        if target_node_gt_15?
+          ui.error("You must use Chef 15 or later to bootstrap Chef 15 nodes")
+          exit 1
+        end
+
         true
       end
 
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -28,6 +28,7 @@ class Knife
     class Bootstrap < Knife
       include DataBagSecretOptions
 
+      CHEF_15 ||= 15
```
btm: I don't know that this constant is really helpful, but since it is a constant you could do `15.freeze`

lib/chef/knife/bootstrap.rb
```diff
@@ -432,6 +434,22 @@ def validate_options!
...
 
+      # Ensure options are valid by checking bootstrap version values.
+      #
+      # The method call will cause the program to exit(1) if:
+      #   * Bootstrap version is greater than or equal to 15 for chef 14 or lower
+      #   * Pre-release options is passed for chef 14 or lower
+      #
+      # @return [TrueClass] If options are valid.
+      def validate_bootstrap_version_options!
+        if target_node_gt_15?
```
btm: this would be more understandable if both the conditions were here, e.g.

```
if (Chef::VERSION.to_i < 15) && target_version_gte_15?
```

note that since we're putting this on the Chef 14 branch, in theory Chef::VERSION.to_i should always be < 15, but I guess it's safe to be clear.

lib/chef/knife/bootstrap.rb
```diff
@@ -496,6 +514,20 @@ def incomplete_policyfile_options?
         (!!config[:policy_name] ^ config[:policy_group])
       end
 
+      # True if the bootstrap version is greater than or equal to 15 or latest.
+      def bootstrap_version_gt_15?
```
btm: this is actually greater than or equal to, so the pattern we've used is "gte" instead of "gt" in these method names, i.e. `bootstrap_version_gte_15?`

lib/chef/knife/bootstrap.rb
```diff
@@ -496,6 +514,20 @@ def incomplete_policyfile_options?
...
+        !!config[:bootstrap_version] &&
+          (config[:bootstrap_version] == "latest" ||
+           config[:bootstrap_version].split(".").first.to_i >= CHEF_15)
+      end
+
+      # Consider only if current chef version less than 15.
+      # True if both bootstrap node version greater than or equal to 15 or pre-release is set.
+      def target_node_gt_15?
```
btm: the `target_node...` part here is misleading because we're not talking about the other node, the target, we're talking about this computer, the workstation. That said, in my comment above I think we don't need this method as we could simply do `Chef::VERSION.to_i >= 15`

lib/chef/knife/bootstrap.rb
vsingh-msys: I think, it can be revemod safely.

lib/chef/knife/bootstrap.rb
vsingh-msys: Okay, thanks!

vsingh-msys:
@btm Incorporated the reviews comments. could you please have a look again?
Thanks!

lib/chef/knife/bootstrap.rb
```diff
@@ -432,6 +433,22 @@ def validate_options!
...
 
+      # Ensure options are valid by checking bootstrap version values.
+      #
+      # The method call will cause the program to exit(1) if:
+      #   * Bootstrap version is greater than or equal to 15 for chef 14 or lower
+      #   * Pre-release options is passed for chef 14 or lower
+      #
+      # @return [TrueClass] If options are valid.
+      def validate_bootstrap_version_options!
+        if Chef::VERSION.to_i < 15 && bootstrap_version_gte_15?
```
tas50: There's no need for the Chef::VERSION check since we know we're on 14

vsingh-msys: Yes, @tas50 there is no need of checking version but for sanity purchase, we may think of?

lib/chef/knife/bootstrap.rb
```diff
@@ -432,6 +434,22 @@ def validate_options!
...
+          ui.error("You must use Chef 15 or later to bootstrap Chef 15 nodes")
```
tas50: Chef -> Chef Infra Client

PR #8787, opened 2019-08-01
For Discussion:  Weirich Semantic Block style

lamont-granquist:
Style/BlockDelimiters:
  Enabled: true
  EnforcedStyle: semantic

See, eg:

https://github.com/rubocop-hq/ruby-style-guide/issues/162

http://www.virtuouscode.com/2011/07/26/the-procedurefunction-block-convention-in-ruby/

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -98,7 +98,7 @@ module ChefConfig
 
     def self.apply_extra_config_options(extra_config_options)
       if extra_config_options
-        extra_parsed_options = extra_config_options.inject({}) do |memo, option|
+        extra_parsed_options = extra_config_options.inject({}) { |memo, option|
           # Sanity check value.
           if option.empty? || !option.include?("=")
             raise UnparsableConfigOption, "Unparsable config option #{option.inspect}"
@@ -111,7 +111,7 @@ module ChefConfig
           # runs a simple parse on the string for some common types.
           memo[key.to_sym] = YAML.safe_load(value)
           memo
-        end
+        }
         merge!(extra_parsed_options)
       end
     end
@@ -153,7 +153,7 @@ module ChefConfig
     # So while this is basically identical to what method_missing would do, we pull
     # it up here and get a real method written so that things get dispatched
     # properly.
-    configurable(:daemonize).writes_value { |v| v }
+    configurable(:daemonize).writes_value do |v| v end
 
     # The root where all local chef object data is stored.  cookbooks, data bags,
     # environments are all assumed to be in separate directories under this.
@@ -203,23 +203,23 @@ module ChefConfig
 
     # Location of acls on disk. String or array of strings.
     # Defaults to <chef_repo_path>/acls.
-    default(:acl_path) { derive_path_from_chef_repo_path("acls") }
+    default(:acl_path) do derive_path_from_chef_repo_path("acls") end
 
     # Location of clients on disk. String or array of strings.
     # Defaults to <chef_repo_path>/clients.
-    default(:client_path) { derive_path_from_chef_repo_path("clients") }
+    default(:client_path) do derive_path_from_chef_repo_path("clients") end
 
     # Location of client keys on disk. String or array of strings.
     # Defaults to <chef_repo_path>/client_keys.
-    default(:client_key_path) { derive_path_from_chef_repo_path("client_keys") }
+    default(:client_key_path) do derive_path_from_chef_repo_path("client_keys") end
 
     # Location of containers on disk. String or array of strings.
     # Defaults to <chef_repo_path>/containers.
-    default(:container_path) { derive_path_from_chef_repo_path("containers") }
+    default(:container_path) do derive_path_from_chef_repo_path("containers") end
 
     # Location of cookbook_artifacts on disk. String or array of strings.
     # Defaults to <chef_repo_path>/cookbook_artifacts.
-    default(:cookbook_artifact_path) { derive_path_from_chef_repo_path("cookbook_artifacts") }
+    default(:cookbook_artifact_path) do derive_path_from_chef_repo_path("cookbook_artifacts") end
 
     # Location of cookbooks on disk. String or array of strings.
     # Defaults to <chef_repo_path>/cookbooks.  If chef_repo_path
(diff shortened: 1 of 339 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -235,39 +235,39 @@ def self.derive_path_from_chef_repo_path(child_path)
...
 
     # Location of groups on disk. String or array of strings.
     # Defaults to <chef_repo_path>/groups.
-    default(:group_path) { derive_path_from_chef_repo_path("groups") }
+    default(:group_path) do derive_path_from_chef_repo_path("groups") end
 
     # Location of nodes on disk. String or array of strings.
     # Defaults to <chef_repo_path>/nodes.
-    default(:node_path) { derive_path_from_chef_repo_path("nodes") }
+    default(:node_path) do derive_path_from_chef_repo_path("nodes") end
```
lamont-granquist: if i'm understanding it correctly this seem incorrectly applied by rubocop?  these functions just do some string wrangling and return a value.

chef-config/lib/chef-config/workstation_config_loader.rb
```diff
@@ -265,7 +265,7 @@ def find_default_key(key_names)
 
     def highlight_config_error(file, line)
       config_file_lines = []
-      IO.readlines(file).each_with_index { |l, i| config_file_lines << "#{(i + 1).to_s.rjust(3)}: #{l.chomp}" }
+      IO.readlines(file).each_with_index do |l, i| config_file_lines << "#{(i + 1).to_s.rjust(3)}: #{l.chomp}" end
```
lamont-granquist: this should be an `each_with_object` or `inject`, so i guess this is right and points out the code is ugly for other reasons?

lib/chef/application.rb
```diff
@@ -342,7 +342,7 @@ def fork_chef_client
         else
           exit 0
         end
-      end
+      }
```
lamont-granquist: this definitely looks like rubocop got it wrong.

lib/chef/chef_fs/file_system/chef_server/nodes_dir.rb
```diff
@@ -28,9 +28,9 @@ module ChefServer
         class NodesDir < RestListDir
           # Identical to RestListDir.children, except supports environments
           def children
-            @children ||= root.get_json(env_api_path).keys.sort.map do |key|
+            @children ||= root.get_json(env_api_path).keys.sort.map { |key|
               make_child_entry(key, true)
-            end
+            }
```
lamont-granquist: I'm fairly sure mike_child_entry actually does a side-effect so this looks incorrectly applied again.

lib/chef/chef_fs/file_system/repository/chef_repository_file_system_root_dir.rb
```diff
@@ -163,11 +163,11 @@ def fs_description
           def root_dir
             existing_paths = root_paths.select { |path| File.exists?(path) }
             if existing_paths.size > 0
-              MultiplexedDir.new(existing_paths.map do |path|
+              MultiplexedDir.new(existing_paths.map { |path|
                 dir = FileSystemEntry.new(name, parent, path)
                 dir.write_pretty_json = !!write_pretty_json
                 dir
-              end)
+              })
```
lamont-granquist: This looks correct but i had to go through 4 levels of super calls in the FileSystemEntry.new initializer to determine that none of them had side effects.  And while i try to put initializers on a diet whenever i come across fat ones with side-effects that definitely can't be trusted in our codebase.  (That many levels of inheritance is another thing, but we're well into crazy sections of the chef codebase here).

lib/chef/chef_fs/file_system/repository/file_system_entry.rb
```diff
@@ -82,7 +82,7 @@ def children
             # Except cookbooks and data bag dirs, all things must be json files
             Dir.entries(file_path).sort
               .map { |child_name| make_child_entry(child_name) }
-              .select { |new_child| new_child.fs_entry_valid? && can_have_child?(new_child.name, new_child.dir?) }
+              .select do |new_child| new_child.fs_entry_valid? && can_have_child?(new_child.name, new_child.dir?) end
```
lamont-granquist: This one looks side effect free to me.

lib/chef/cookbook/manifest_v2.rb
```diff
@@ -31,7 +31,7 @@ def from_hash(hash)
 
         def to_h(manifest)
           result = manifest.manifest.dup
-          result["all_files"].map! { |file| file.delete("full_path"); file }
+          result["all_files"].map! do |file| file.delete("full_path"); file end
```
lamont-granquist: I guess so but again I don't see the value in thinking about it this hard.

robbkidd:
I am personally philosophically-aligned 100% with Weirich-style semantic blocks. I _long_ for a code base that tells me immediately that a block has a return or has side-effects.

However, I am wary of applying it to Chef because it is a style preference that cannot be reliably enforced with automation.

1. `chef/chef` is an 11+ year-old code base. As @lamont-granquist demonstrates in this PR, each of its many blocks will require human review because there are enough cases of the `Style/BlockDelimiters` cop's semantic auto-correct being wrong.
2. Because the cop cannot be relied on, the review of every future PR will require a human reviewer verify the semantic meaning of blocks added/modified.
3. When a human reviewer finds a semantic mismatch in a PR, the review comments to the submitter come in the form of a style suggestion which causes frustration in some submitters for what they see as unnecessary rework.

chef-config/lib/chef-config/workstation_config_loader.rb
zenspider: This screams `map` to me.

chef-config/lib/chef-config/config.rb
zenspider: `rubocop` doesn't know that. It is just looking at the name / context. Since this isn't being assigned or returned, it looks like it is for side-effect. I'd agree based on this code here.

zenspider:
DO IT

tas50:
This syntax makes sense, but we're never going to have 100% accuracy in Rubocop here. It's probably going to be pretty poor at enforcing it and if it's not reliable syntax then the entire point of the syntax is mute. I'm very much thumbs down on this. It's going to results in a lot of work adding exclusions and tracking down rubocop bugs for the foreseeable future.

miah:
I get that it would be painful for the Chef Infra code to move to this pattern, it is something we may want to do with InSpec. 

Can we allow both patterns? Maybe that would require us disabling a rubocop check that enforces one style over the other? I can see where the pattern not being applied evenly over an entire code base could be confusing but we can still be intentional and work towards that over time. And we can be explicit in contributor documentation about patterns the code intends to follow.

zenspider:
> However, I am wary of applying it to Chef because it is a style preference that cannot be reliably enforced with automation.

@robbkidd then let's turn the damn thing off? The 1-liner distinction is meaningless and non-helpful.

robbkidd:
> let's turn the damn thing off? The 1-liner distinction is meaningless and non-helpful.

I hear you. Then the new proposal is for chefstyle to adopt this, yea?

```
Style/BlockDelimiters:
Enabled: false
```

lamont-granquist:
@miah 	@zenspider 	chefstyle picks up .rubocop.yml so you're always free to diverge:

chef/chef does that for a few things which we turned on globally but would need to get cleaned up here:

https://github.com/chef/chef/blob/master/.rubocop.yml

you can turn it off / turn on semantic enforcement in inspec/train or we can flip the default to off in chefstyle.

i think i'm okay with the latter and that's probably what we should do.

PR #8790, opened 2019-08-02
[ci] Fix omnibus pipeline

jeremiahsnapp:
## Description

This change only affects CI. A new version of omnibus is required and the artifactory gem is required for the publish part of the build stage. Platform names in `release.omnibus.yml` are also fixed to conform with what the new package publishing code in the omnibus gem expects.

## Related Issue

https://github.com/chef/omnibus-buildkite-plugin/issues/22

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -12,8 +12,8 @@ builder-to-testers-map:
   debian-8-x86_64:
     - debian-8-x86_64
     - debian-9-x86_64
-  el-6-i386:
-    - el-6-i386
+  el-6-i686:
+    - el-6-i686
   el-6-s390x:
     - el-6-s390x
   el-6-x86_64:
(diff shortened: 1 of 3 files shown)


.expeditor/release.omnibus.yml
```diff
@@ -12,8 +12,8 @@ builder-to-testers-map:
   debian-8-x86_64:
     - debian-8-x86_64
     - debian-9-x86_64
-  el-6-i386:
-    - el-6-i386
+  el-6-i686:
+    - el-6-i686
```
robbkidd: Um. Is this a dramatic change in architecture we support or only a labeling change?

.expeditor/release.omnibus.yml
```diff
@@ -29,9 +29,9 @@ builder-to-testers-map:
   el-7-x86_64:
     - el-7-x86_64
     - el-8-x86_64
-  freebsd-10-x86_64:
-    - freebsd-10-x86_64
-    - freebsd-11-x86_64
+  freebsd-10-amd64:
+    - freebsd-10-amd64
+    - freebsd-11-amd64
```
robbkidd: Is this a dramatic change in architecture we support or only a labeling change?

.expeditor/release.omnibus.yml
jeremiahsnapp: only a labeling change as mentioned in the PR comment. good catch though.

.expeditor/release.omnibus.yml
jeremiahsnapp: only a labeling change as mentioned in the PR comment. good catch though.

.expeditor/release.omnibus.yml
robbkidd: I saw "Fix platform names to conform with what the new package publishing code expects" in the commit message, but didn't interpret that as assurance that architecture wasn't actually changing. Thanks!

Issue #8764, opened 2019-07-23
Chef-15: Bootstrapping failed due to SSH connection timeout in AIX 7 


vsingh-msys:
## Description

Bootstrapping of AIX 7 node raises the `SSH command failed (Timeout, server IP_ADDRESS not responding.) (Train::Transports::SSHFailed)`

## Chef Version

15.1.36

## Platform Version

AIX Version 7.1
## Replication Case

```
knife bootstrap IP_ADDRESS --connection-user username -N aix-7 -r "recipe[starter]" --connection-password password --ssh-verify-host-key 'never' -VV
```

## Stacktrace

```
DEBUG: Checking if we need to accept Chef license to bootstrap node
DEBUG: Reading products and relationships...
DEBUG: Successfully read products and relationships
DEBUG: License acceptance required for chef version: 15
DEBUG: Searching for the following licenses: ["infra-client", "inspec"]
DEBUG: Found license chef_infra_client at /.chef/accepted_licenses/chef_infra_client
DEBUG: Found license inspec at /.chef/accepted_licenses/inspec
DEBUG: Missing licenses remaining: []
DEBUG: All licenses present
Connecting to IP_ADDRESS
WARN: [SSH] PTY requested: stderr will be merged into stdout
DEBUG: [SSH] username@IP_ADDRESS<{:user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>60, :auth_methods=>["none", "password", "keyboard-interactive"], :keys_only=>nil, :keys=>[], :password=>"<hidden>", :forward_agent=>nil, :non_interactive=>true, :verify_host_key=>:never}> (cmd.exe /c ver)
DEBUG: [SSH] opening connection to username@IP_ADDRESS<{:user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>60, :auth_methods=>["none", "password", "keyboard-interactive"], :keys_only=>nil, :keys=>[], :password=>"<hidden>", :forward_agent=>nil, :non_interactive=>true, :verify_host_key=>:never}>
[...]

PR #8795, opened 2019-08-06
Chef-15: Make temp dir using mkdir instead of mktemp

vsingh-msys:
## Description 
- In order to get rid of `mktemp` use `mkdir` for Linux and Unix host.
- `mktemp` does not exist for AIX host.
 - `mkdir` does not return the created path so return the dir itself.

Note: working on to fix `chown: /tmp/chef_Xr0QpL: Operation not permitted.` 

seems run_command! created dir with `root:system` user privileges 
but when ssh login on AIX host and run the same command to created dir it get created with `user:staff` permission.

```
drwxr-xr-x    2 vsingh   staff           256 Aug 06 08:36 chef_Xr0Q
drwxr-xr-x    2 root     system          256 Aug 06 08:35 chef_Xr0QpL
```
 
Any suggestions would be helpful?

Thanks!

## Related Issue

https://github.com/chef/chef/issues/8764

Issue found while creating temp dir https://github.com/chef/chef/blob/3284e9067aacbe42f96afc78b06e523b46b98d2f/lib/chef/knife/bootstrap/train_connector.rb#L35

ERROR: `bash: mktemp: command not found`

In order to make it work, we have to update the command for creating temp dir

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap/train_connector.rb
+++ b/lib/chef/knife/bootstrap/train_connector.rb
@@ -18,6 +18,7 @@
 require "train"
 require "tempfile" unless defined?(Tempfile)
 require "uri" unless defined?(URI)
+require "securerandom" unless defined?(SecureRandom)
 
 class Chef
   class Knife
@@ -32,7 +33,7 @@ class Chef
           $tmp.FullName
         EOM
 
-        MKTEMP_NIX_COMMAND ||= "bash -c 'd=$(mktemp -d ${TMPDIR:-/tmp}/chef_XXXXXX); echo $d'".freeze
+        DEFAULT_REMOTE_TEMP ||= "/tmp".freeze
 
         def initialize(host_url, default_protocol, opts)
           @host_url = host_url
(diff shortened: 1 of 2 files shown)


lamont-granquist:
This is more or less what we have to do with install.sh as well.

lamont-granquist:
Addressing the possible security implications for future reference:

- mkdir should fail and run_command! should raise rather than be exploitable, so its more of a DoS than an exploit, even if you could somehow guess the value of securerandom.

- and how would an attacker have access to a node in the process of bootstrapping?

So this LGTM.

vsingh-msys:
Fixed `chown: /tmp/chef_Xr0QpL: Operation not permitted.` by adding `sudo` while changing ownership.
New
```
mkdir -p /tmp/chef_qeAUFU && sudo chown user '/tmp/chef_qeAUFU'
```
Old
```
mkdir -p /tmp/chef_qeAUFU && chown user '/tmp/chef_qeAUFU'
```

PR #8801, opened 2019-08-08
Add retries to the GitHub PRs in Buildkite

tas50:
We keep seeing random failures on our PRs.  This resolves that.


.expeditor/verify.pipeline.yml
```diff
@@ -3,6 +3,9 @@ expeditor:
   defaults:
     buildkite:
       timeout_in_minutes: 30
+    retry:
```
tduffield: This needs to be one level deeper, at the same level as `timeout_in_minutes`.

PR #8842, opened 2019-08-23
Always use system rubygem to determine nodocument flag.

st0012:
(This could only happen on chef `14.x` at this moment)

Currently if the user doesn't specify custom rubygem binary path,
chef uses `Gem::VERSION` to determine if the rubygem version is
compatiable with old nodocument flags like `--no-rdoc`.

https://github.com/chef/chef/blob/7558c414030d082194d0d2b5f74279ce151517e9/lib/chef/provider/package/rubygems.rb#L601-L611

If the user doesn't specify rubygem binary path:
https://github.com/chef/chef/blob/7558c414030d082194d0d2b5f74279ce151517e9/lib/chef/provider/package/rubygems.rb#L272-L274

If the user specifies rubygem binary path:
https://github.com/chef/chef/blob/7558c414030d082194d0d2b5f74279ce151517e9/lib/chef/provider/package/rubygems.rb#L303-L305

This doesn't work because the `Gem::VERSION` always points to chef's
embedded rubygem version. But the binary we use to install gems is
the system's rubygem, which might be different from chef's embedded
version **even though the user doesn't specify custom binary path**.

So when a user upgrades his Ruby version to 2.6 (rubygem 3.0.x) with
Chef 14.x (rubygem 2.7.x), an error will occur:

```
 STDOUT:
   STDERR: ERROR:  While executing gem ... (OptionParser::InvalidOption)
       invalid option: --no-rdoc
```

This is because `Gem::VERSION` would make chef think that it's ok to use
`--no-rdoc` while it's not. The solution is to always perform the check with
system rubygem's version.

Signed-off-by: st0012 <stan001212@gmail.com>

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -604,7 +604,8 @@ class Chef
         end
 
         def needs_nodocument?
-          Gem::Requirement.new(">= 3.0.0.beta1").satisfied_by?(Gem::Version.new(gem_env.rubygems_version))
+          rubygem_version = shell_out!("#{gem_binary_path} --version").stdout.chomp
+          Gem::Requirement.new(">= 3.0.0.beta1").satisfied_by?(Gem::Version.new(rubygem_version))
         end
 
         def opts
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/rubygems.rb
```diff
@@ -604,7 +604,8 @@ def rdoc_string
         end
 
         def needs_nodocument?
-          Gem::Requirement.new(">= 3.0.0.beta1").satisfied_by?(Gem::Version.new(gem_env.rubygems_version))
+          rubygem_version = shell_out!("#{gem_binary_path} --version").stdout.chomp
```
st0012: I don't want to put another rubygem version method out there just for this case. I think that will cause confusion because we already have 2 similar methods in the same file.

tas50:
Closing this based on @lamont-granquist's response

PR #8844, opened 2019-08-23
Fix option parsing order for knife commands

### LINKED ISSUE #8834: knife bootstrap --bootstrap-version command line option not working
## Description
When running knife bootstrap with `--bootstrap-version` option, the option if set in `~/.chef/config.rb` takes precedence. According to https://docs.chef.io/config_rb.html#optional-settings the command line option should take precedence. The code at https://github.com/chef/chef/blob/master/lib/chef/knife/core/bootstrap_context.rb#L189-L205 in the `knife_config` method only appears to check `@chef_config` for the value used in `version_to_install`.

## Chef Version
```
$ chef --version
Chef Development Kit Version: 3.11.3
chef-client version: 14.13.11
delivery version: master (9d07501a3b347cc687c902319d23dc32dd5fa621)
berks version: 7.0.8
kitchen version: 1.25.0
inspec version: 3.9.3
```

## Platform Version
Ubuntu 16.04 (both workstation and client)

## Replication Case
* add `knife[:bootstrap_version] = '13.9.4'` to `~/.chef/config.rb`
* run `knife bootstrap --bootstrap-version '14.12.9' $HOST`
* expected result: `-----> Installing Chef Omnibus (-v 14.12.9)`
* actual result: `-----> Installing Chef Omnibus (-v 13.9.4)`

## Client Output
```
$ knife bootstrap -VV --bootstrap-version '14.12.9' --no-host-key-verify --ssh-user vagrant -N test 192.168.121.194
INFO: Using configuration from /home/myusername/.chef/config.rb
Creating new client for test
Creating new node for test
Connecting to 192.168.121.194
DEBUG: Adding 192.168.121.194
vagrant@192.168.121.194's password:
192.168.121.194 -----> Installing Chef Omnibus (-v 13.9.4)
192.168.121.194 downloading https://omnitruck.chef.io/chef/install.sh
192.168.121.194   to file /tmp/install.sh.1645/install.sh
```

ccrebolder:
* https://docs.chef.io/plugin_knife_custom.html#config-rb-settings

A common pattern for mixlib-cli procs is to set the Chef::Config[:knife]
setting to the same as the command line argument, however Chef::Config
has not been fully loaded at the time the options are parsed, which
means if the config file has the same option it resets to that value
after the load.

Fixes #8834

Signed-off-by: Chris Crebolder <ccrebolder@gmail.com>

## Description
The expected behaviour is for the proc to be able to manually set a value in Chef::Config, however if the same option exists in the config file, it is loaded after the proc has run, setting it back to the original.

## Related Issue
#8834

## Types of changes

## Checklist:

### DIFF
--- a/spec/data/knife_subcommand/test_yourself.rb
+++ b/spec/data/knife_subcommand/test_yourself.rb
@@ -11,6 +11,8 @@ module KnifeSpecs
 
     option :scro, :short => '-s SCRO', :long => '--scro SCRO', :description => 'a configurable setting'
 
+    option :with_proc, :long => '--with-proc VALUE', proc: Proc.new { |v| Chef::Config[:knife][:with_proc] = v }
+
     attr_reader :ran
 
     def run
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -217,8 +217,8 @@ class Chef
       subcommand_class = subcommand_class_from(args)
       subcommand_class.options = options.merge!(subcommand_class.options)
       subcommand_class.load_deps
-      instance = subcommand_class.new(args)
-      instance.configure_chef
+      instance = subcommand_class.new(args.dup)
+      instance.configure_chef(args)
       instance.run_with_pretty_exceptions
     end
 
@@ -299,7 +299,7 @@ class Chef
       command_name_words = self.class.snake_case_name.split("_")
 
       # Mixlib::CLI ignores the embedded name_args
-      @name_args = parse_options(argv)
+      @name_args = opt_parser.parse!(argv)
       @name_args.delete(command_name_words.join("-"))
       @name_args.reject! { |name_arg| command_name_words.delete(name_arg) }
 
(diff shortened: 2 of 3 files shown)


lamont-granquist:
So I believe dan fixed this ages and ages ago and we need to do a bit of historical digging here to figure out what is up...

As far as I can recall we should be able to achieve this fairly simply without a lot of modifications to core code, and with cleaner looking code for handling the option....  But that patch went in years ago so I may have some brainrot...

spec/data/knife_subcommand/test_yourself.rb
```diff
@@ -11,6 +11,8 @@ class << self
 
     option :scro, :short => '-s SCRO', :long => '--scro SCRO', :description => 'a configurable setting'
 
+    option :with_proc, :long => '--with-proc VALUE', proc: Proc.new { |v| Chef::Config[:knife][:with_proc] = v }
```
lamont-granquist: I would prefer to remove this as a way of writing cli options.  AFIAK this was the way that people tried to "make the cli win" circa chef-0.10.x and then we fixed most of that, but never went through and scrubbed all the Procs out, but if we're going to do anything I'd prefer to remove every single one of these Procs from the codebase.

tas50:
@ccrebolder I'm going to close this out at this point based on the feedback from @lamont-granquist. This seems like it's going to require a much larger change to how we handle options.

PR #8856, opened 2019-08-30
Fix Bootstrap password prompt

samshinde:
Bootstrap command was not prompting for the password when --connection-password option is not passed.

## Description
 
Bootstrap command was throwing error `ERROR: RuntimeError: password is a required option` when --connection-password option is not passed. The exception was not getting raised because of an incorrect method call.
- Done correct exception handling
- Fixed test case

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -616,7 +616,7 @@ class Chef
 
       def connect!
         ui.info("Connecting to #{ui.color(server_name, :bold)}")
-        opts = connection_opts.dup
+        opts ||= connection_opts.dup
         do_connect(opts)
       rescue Train::Error => e
         # We handle these by message text only because train only loads the
@@ -638,8 +638,10 @@ class Chef
           EOM
           # FIXME: this should save the key to known_hosts but doesn't appear to be
           config[:ssh_verify_host_key] = :accept_new
-          do_connect(connection_opts(reset: true))
-        elsif ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed
+          conn_opts = connection_opts(reset: true)
+          opts.merge! conn_opts
+          retry
+        elsif (ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed) || (ssh? && e.class == Train::ClientError && e.reason == :no_ssh_password_or_key_available)
           if connection.password_auth?
             raise
           else
@@ -650,7 +652,23 @@ class Chef
           end
 
           opts.merge! force_ssh_password_opts(password)
-          do_connect(opts)
+          retry
+        else
+          raise
+        end
+      rescue RuntimeError => e
+        if winrm? && e.message == "password is a required option"
+          if connection.password_auth?
+            raise
+          else
+            ui.warn("Failed to authenticate #{opts[:user]} to #{server_name} - trying password auth")
+            password = ui.ask("Enter password for #{opts[:user]}@#{server_name}.") do |q|
+              q.echo = false
+            end
+          end
+
+          opts.merge! force_winrm_password_opts(password)
+          retry
         else
           raise
         end
@@ -1013,6 +1031,12 @@ class Chef
           keys_only: false,
           key_files: [],
           auth_methods: %i{password keyboard_interactive},
+        }
+      end
+
+      def force_winrm_password_opts(password)
+        {
+          password: password,
         }
       end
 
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -614,9 +614,9 @@ def perform_bootstrap(remote_bootstrap_script_path)
         end
       end
 
-      def connect!
+      def connect!(opts = nil)
```
Nimesh-Msys: Any thoughts to [Keyword Arguments](https://thoughtbot.com/blog/ruby-2-keyword-arguments) ?

lib/chef/knife/bootstrap.rb
```diff
@@ -638,8 +638,10 @@ def connect!
           EOM
           # FIXME: this should save the key to known_hosts but doesn't appear to be
           config[:ssh_verify_host_key] = :accept_new
-          do_connect(connection_opts(reset: true))
-        elsif ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed
+          conn_opts = connection_opts(reset: true)
+          opts.merge! conn_opts
+          connect!(opts)
+        elsif ssh? && e.class == Train::ClientError && e.message == "Your SSH Agent has no keys added, and you have not specified a password or a key file"
```
Nimesh-Msys: Error check with string comparison could be risky. Also @ L#660, would you please confirm if this is really the last option.

lib/chef/knife/bootstrap.rb
```diff
@@ -650,7 +652,23 @@ def connect!
           end
 
           opts.merge! force_ssh_password_opts(password)
-          do_connect(opts)
+          connect!(opts)
+        else
+          raise
```
Nimesh-Msys: Best if we could have redefine [exceptions](https://github.com/chef/chef/blob/master/lib/chef/exceptions.rb) and can raise and catch them.

lib/chef/knife/bootstrap.rb
vsingh-msys: @Nimesh-Msys sure this might be an improvement area for https://github.com/inspec/train

lib/chef/knife/bootstrap.rb
lamont-granquist: that bit is actually fine, the code is just stating that it handles certain Train:Error's but not others.

lib/chef/knife/bootstrap.rb
```diff
@@ -638,8 +638,10 @@ def connect!
           EOM
           # FIXME: this should save the key to known_hosts but doesn't appear to be
           config[:ssh_verify_host_key] = :accept_new
-          do_connect(connection_opts(reset: true))
-        elsif ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed
+          conn_opts = connection_opts(reset: true)
+          opts.merge! conn_opts
```
marcparadise: Why the change to merge into old opts here?

lib/chef/knife/bootstrap.rb
```diff
@@ -638,8 +638,10 @@ def connect!
...
+          retry
+        elsif (ssh? && e.cause && e.cause.class == Net::SSH::AuthenticationFailed) || (ssh? && e.class == Train::ClientError && e.message == "Your SSH Agent has no keys added, and you have not specified a password or a key file")
```
marcparadise: This would work, but I'd prefer updating train [1] to populate the exception's `reason` field so that we don't have to worry about changes to text. This'll also keep moving train towards providing a consistent way to discover exception reasons.
 
[1] https://github.com/inspec/train/blob/master/lib/train/transports/ssh.rb#L106
to something like: 
```
        raise Train::ClientError.new(
            "Your SSH Agent has no keys added, and you have not specified a password or a key file",
            :no_ssh_password_or_key_available
        )
```

lib/chef/knife/bootstrap.rb
samshinde: Because if --connection-password is not passed and `Net::SSH::AuthenticationFailed` or `Train::ClientError` error is thrown then old opts will have the password.

lib/chef/knife/bootstrap.rb
samshinde: Updated comparison with `e.reason`. Added PR https://github.com/inspec/train/pull/517 for this change.

vsingh-msys:
@tas50 this PR is good to merge as inspec/train#517 is now in master. Also, need to pin train 3.1+
for the latest changes.
```
s.add_dependency "train-core", "~> 3.1" 
```

PR #8861, opened 2019-09-04
convert chocolatey resource to use shell_out splat args

lamont-granquist:
This improves command line argument handling on windows.  Should correctly deal with arguments-with-spaces issus along with any windows metacharacters in any command line arguments.  I don't see any open issues surrounding those, but we definitely have some questionable handling here.  If this is successful we can more broadly apply this across the codebase as a standard (finally).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,10 +8,10 @@ GIT
 
 GIT
   remote: https://github.com/chef/ohai.git
-  revision: 15315e3e36f07062111faede26f2667ead8c723f
+  revision: 9982b1e62ed1afb8db57349f536da3fed759749f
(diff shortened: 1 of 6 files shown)


tas50:
We need to raise our minimum mixlib-shellout to 3.0.4 with this change now right? We officially require the splat format to work.

lamont-granquist:
Yes probably, i need to give a little thought to how the options work here as well when they're not passed in as an array of strings to be joined, although I think its okay, but need to think about it after better sleep than I got last night...

lamont-granquist:
Gotta make sure the func tests are really running on appveyor as well and that there's enough coverage there...

PR #8865, opened 2019-09-05
Chef 16: Remove the ability to write encrypted data bag V1 / V2 formats

tas50:
We can still read existing encrypted data bags, but we shouldn't write to the legacy formats anymore. There's just no reasons to continue to support these legacy formats. V3 was introduced in Chef 12 and became the default in Chef 13.


tas50:
We're going to have to move some of the specs to hard coded values for v1 / v2 instead of using the encryption classes to make the values first. Should be faster as well.

lamont-granquist:
looks like the thing that does the stuff.

PR #8872, opened 2019-09-10
Further cleanup on Application::Client and Solo and deprecate solo_legacy_mode UX

### LINKED ISSUE #6375: Chef-Solo loads configurations twice

### LINKED ISSUE #8138: Duplicate log lines when using a formatter
## Description
When running `chef-solo` with a custom formatter, every log line is printed twice. This seems similar to #7184.

## Chef Version
v14.8.12

## Platform Version
SunOS <hostname> 5.11 joyent_20180816T001857Z i86pc i386 i86pc Solaris

## Client Output

```
[root@jpc1-anand-01 /var/chef]# chef-solo -W -c solo.rb -F doc
Starting Chef Client, version 14.8.12
Starting Chef Client, version 14.8.12
resolving cookbooks for run list: ["chef"]
resolving cookbooks for run list: ["chef"]
Synchronizing Cookbooks:
Synchronizing Cookbooks:
  - chef (0.0.0)
  - chef (0.0.0)
Installing Cookbook Gems:
Installing Cookbook Gems:
Compiling Cookbooks...
Compiling Cookbooks...
Converging 6 resources
Converging 6 resources
Recipe: chef::default
  * directory[/var/chef] action createRecipe: chef::default
  * directory[/var/chef] action create (up to date)
 (up to date)
  * directory[/var/chef/bin] action create  * directory[/var/chef/bin] action create
    - Would create new directory /var/chef/bin
    - Would create new directory /var/chef/bin

  * gem_package[chef] action install  * gem_package[chef] action install (up to date)
 (up to date)
  * gem_package[berkshelf] action install  * gem_package[berkshelf] action install (up to date)
 (up to date)
  * template[/var/chef/bin/run-chef] action create  * template[/var/chef/bin/run-chef] action create
    * Parent directory /var/chef/bin does not exist.
    * Parent directory /var/chef/bin does not exist.
    * Assuming directory /var/chef/bin would have been created
    * Assuming directory /var/chef/bin would have been created
    - Would create new file /var/chef/bin/run-chef
    - Would create new file /var/chef/bin/run-chef
    - Would change mode from '' to '0755'
    - Would change mode from '' to '0755'

  * link[/opt/local/bin/run-chef] action create  * link[/opt/local/bin/run-chef] action create
    - Would create symlink at /opt/local/bin/run-chef to /var/chef/bin/run-chef
    - Would create symlink at /opt/local/bin/run-chef to /var/chef/bin/run-chef

[2019-01-20T20:14:03+00:00] WARN: In why-run mode, so NOT performing node save.

Running handlers:

Running handlers:
Running handlers complete
Running handlers complete
Chef Client finished, 3/6 resources would have been updated
Chef Client finished, 3/6 resources would have been updated
```
## Stacktrace

### LINKED ISSUE #8366: The chef-solo binstubs need to use the Chef::Application::Client class
We really need to look at inverting the way that we invoke the Chef::Application::Solo class and then trampoline out of that into Chef::Application::Client when we are in zolo mode and not legacy chef-solo.  Instead we should invoke Chef::Application::Client and then only trampoline out when we are in legacy mode, making legacy-mode the special handling case.

This means that command line parsing will have to be made consistent between the classes (and is probably a good thing -- reducing code duplication and interface drift between the two classes).  The Chef::Application::Solo class should also likely have its CLI arguments removed, and all the code duplication should really get removed as much as possible.

In the end it would be best to have only Chef::Application::Client exist with special-casing around the legacy mode operation of chef-solo (although its also legacy and at some point we should deprecate that behavior, but that requires fixing the perf issues in ChefFS).

lamont-granquist:
🚧 

closes #8366
fixes #6375
fixes #8138

### DIFF
--- a/chef-bin/bin/chef-solo
+++ b/chef-bin/bin/chef-solo
@@ -19,6 +19,6 @@
 # limitations under the License.
 
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
-require "chef/application/solo"
+require "chef/application/client"
 
 Chef::Application::Client.new(solo: true).run(enforce_license: true)
--- a/chef-bin/bin/chef-solo
+++ b/chef-bin/bin/chef-solo
@@ -21,4 +21,4 @@
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 require "chef/application/solo"
 
-Chef::Application::Client.new(solo: true).run
+Chef::Application::Client.new(solo: true).run(enforce_license: true)
--- a/chef-bin/bin/chef-solo
+++ b/chef-bin/bin/chef-solo
@@ -21,4 +21,4 @@
 $:.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 require "chef/application/solo"
 
-Chef::Application::Solo.new.run(enforce_license: true)
+Chef::Application::Client.new(solo: true).run
--- a/lib/chef/application/base.rb
+++ b/lib/chef/application/base.rb
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require_relative "../../chef"
 require_relative "../application"
 require_relative "../client"
 require_relative "../log"
@@ -22,9 +23,12 @@ require_relative "../mixin/shell_out"
 require_relative "../config_fetcher"
 require_relative "../dist"
 require_relative "../daemon"
+require_relative "../workstation_config_loader"
 require "chef-config/mixin/dot_d"
 require "license_acceptance/cli_flags/mixlib_cli"
 require "mixlib/archive" unless defined?(Mixlib::Archive)
+require "fileutils" unless defined?(FileUtils)
+require "uri" unless defined?(URI)
 
 # This is a temporary class being used as a part of an effort to reduce duplication
 # between Chef::Application::Client and Chef::Application::Solo.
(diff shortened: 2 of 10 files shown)


lamont-granquist:
#8361 broke executing single recipes since dropping privs also does a chdir and the work done to error if the recipes don't exist happen after the chdir now and fail.  that probably needs to get addressed somehow.

floh96:
@lamont-granquist is this on the todo list for chef 17?

PR #8884, opened 2019-09-13
update syntax of `update-rc.d` commands in enable & disable actions

robuye:
Hi Chefs :wave:

I stumbled upon this bug in Kitchen + Docker environment where Chef detects `:debian` service provider and shells out to `update-rc.d` to manage runlevels. I think all full distributions ship with another init system so this bug wouldn't affect most systems traditionally provisioned with Chef.

Docker on the other hand encourages minimal system and usually does not ship with any additional init, hence Chef falls back to `:debian` and `update-rc.d`.

The fix should allow more reliable testing if nothing else by making chef-client go through the same code path in docker and non-docker environments.

I ran a rspec tests locally, but some, unrelated to my changes, were failing to begin with. Otherwise tests related to the `Debian` service provider are passing.

Let me know what you think about this P/R. Thanks!

## Description

Some `update-rc.d` commands do not work as expected because the CLI changed. As a result of this `disable` action does not work and it enables the service instead.

The `stop` and `start` commands were removed in [2.88.dsf-42](https://launchpad.net/debian/+source/sysvinit/2.88dsf-42):

```
  * sysv-rc:
    - update-rc.d no longer supports non-dependency-based boot.
      + Remove non-insserv codepaths.
      + Warn if the start or stop actions are used.
      + Skip runlevel mismatch warnings if default action is used
        (no arguments to validate).
      + Update manual page to remove start and stop actions, plus
        manual setting of boot sequence ordering; note that start
        and stop are no longer supported.  Closes: #606505.
```

and has been since ported to all currently supported Debian distributions (including Jessie, released in 2015).

Both `start` and `stop` commands now effectively call `defaults`.

As a result of this `action_disable` did not disable a service, it would enable it instead. Chef would execute the following commands on disable:

```
/usr/sbin/update-rc.d -f <service_name> remove
/usr/sbin/update-rc.d -f <service_name> stop 80 2 3 4 5 .
```

But `update-rc.d` would effectively run:

```
/usr/sbin/update-rc.d -f <service_name> remove
/usr/sbin/update-rc.d <service_name> defaults
```

And the service gets enabled instead.

With changes in this commit Chef will effectively run the following commands:

```
/usr/sbin/update-rc.d -f <service_name> remove
/usr/sbin/update-rc.d <service_name> defaults
/usr/sbin/update-rc.d <service_name> disable
```

The service is now disabled, as expected.

Additionally `priority` support has been dropped entirely from code as it's not supported by `update-rc.d` either.

It's potentially a breaking change on a very outdated distributions and it could be worked around with additional `if/else` branches, but given all non-EOLed distros have up-to-date version of `update-rc.d` available it didn't feel worth the complexity.

## Related Issue

None, or I didn't find any.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/service/debian.rb
+++ b/lib/chef/provider/service/debian.rb
@@ -149,44 +149,42 @@ class Chef
         end
 
         def enable_service
-          if new_resource.priority.is_a? Integer
-            shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
-            shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults #{new_resource.priority} #{100 - new_resource.priority}")
-          elsif new_resource.priority.is_a? Hash
+          if new_resource.priority.is_a? Hash
             # we call the same command regardless of we're enabling or disabling
             # users passing a Hash are responsible for setting their own start priorities
             set_priority
-          else # No priority, go with update-rc.d defaults
+          else # No priority or Integer. Either way go with update-rc.d defaults
             shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
             shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults")
           end
         end
 
         def disable_service
-          if new_resource.priority.is_a? Integer
-            # Stop processes in reverse order of start using '100 - start_priority'
-            shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
-            shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} stop #{100 - new_resource.priority} 2 3 4 5 .")
-          elsif new_resource.priority.is_a? Hash
+          if new_resource.priority.is_a? Hash
             # we call the same command regardless of we're enabling or disabling
             # users passing a Hash are responsible for setting their own stop priorities
             set_priority
-          else
-            # no priority, using '100 - 20 (update-rc.d default)' to stop in reverse order of start
+          else # No priority or Integer. Either way disable in all levels.
             shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
-            shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} stop 80 2 3 4 5 .")
+            shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults")
+            shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} disable")
           end
         end
 
         def set_priority
-          args = ""
-          new_resource.priority.each do |level, o|
-            action = o[0]
-            priority = o[1]
-            args += "#{action} #{priority} #{level} . "
-          end
+          # Reset priorities to default values before applying customizations. This way
+          # the final state will always be consistent, regardless if all runlevels were
+          # provided.
           shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
-          shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} #{args}")
+          shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults")
+
+          # Note the `_priority` is not used. update-rc.d does not support priorities
+          # anymore, this feature has been dropped in sysvinit 2.88dsf-42.
+          new_resource.priority.each do |level, (action, _priority)|
+            disable_or_enable = (action == :start ? "enable" : "disable")
+
+            shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} #{disable_or_enable} #{level}")
+          end
         end
       end
     end
--- a/lib/chef/provider/service/debian.rb
+++ b/lib/chef/provider/service/debian.rb
@@ -149,42 +149,102 @@ class Chef
         end
 
         def enable_service
+          # We call the same command regardless if we're enabling or disabling
+          # Users passing a Hash are responsible for setting their own stop priorities
           if new_resource.priority.is_a? Hash
-            # we call the same command regardless of we're enabling or disabling
-            # users passing a Hash are responsible for setting their own start priorities
             set_priority
-          else # No priority or Integer. Either way go with update-rc.d defaults
-            shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
-            shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults")
+            return
           end
+
+          start_priority = new_resource.priority.is_a?(Integer) ? new_resource.priority : 20
+          # Stop processes in reverse order of start using '100 - start_priority'.
+          stop_priority = 100 - start_priority
+
+          shell_out!("/usr/sbin/update-rc.d -f #{new_resource.service_name} remove")
+          shell_out!("/usr/sbin/update-rc.d #{new_resource.service_name} defaults #{start_priority} #{stop_priority}")
         end
 
         def disable_service
           if new_resource.priority.is_a? Hash
-            # we call the same command regardless of we're enabling or disabling
-            # users passing a Hash are responsible for setting their own stop priorities
+            # We call the same command regardless if we're enabling or disabling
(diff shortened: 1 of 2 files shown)


lamont-granquist:
> It's potentially a breaking change on a very outdated distributions and it could be worked around with additional if/else branches, but given all non-EOLed distros have up-to-date version of update-rc.d available it didn't feel worth the complexity.

So this is entirely correct, but we're not in Chef-15.0 right now, this would better be put on a Chef-16.0 branch, because it certainly is a breaking change to anyone who wound up using Chef-15 on those EOL'd versions.

tas50:
@robuye Is it possible to check the version of update-rc.d so that you can fall back to the old logic for ancient debian / ubuntu releases?

robuye:
Thanks for the feedback. I pushed one more commit that brings back legacy syntax and explained reasoning in the commit message & code comments. Additionally I tested it as much as I could using official docker images for Ubuntu (10.04 up to 18.04) & Debian (6-8).

Ubuntu 14 comes with old-enough `rc-update.d` to support start & stop commands and priorities:

```
$ cat /etc/lsb-release |grep DESCRI
DISTRIB_DESCRIPTION="Ubuntu 14.04.6 LTS"
$ dpkg-query -W --showformat "\${Version}\n" sysv-rc
2.88dsf-41ubuntu6.3
$ touch /etc/init.d/example && chmod +x /etc/init.d/example
$ update-rc.d -f example remove
 Removing any system startup links for /etc/init.d/example ...
$ update-rc.d example start 20 2 . stop 55 3 .  2>/dev/null
 Adding system startup for /etc/init.d/example ...
   /etc/rc3.d/K55example -> ../init.d/example
   /etc/rc2.d/S20example -> ../init.d/example
$ update-rc.d -f example remove
 Removing any system startup links for /etc/init.d/example ...
   /etc/rc2.d/S20example
   /etc/rc3.d/K55example
```

Ubuntu 10 & 12 worked the same while Ubuntu 16 & 18 require "modern" syntax as start & stop don't work. I used `rc.local` service here instead of `example` because the output is more meaningful when LSB init comments are present.

```
$ cat /etc/lsb-release |grep DESCRI
DISTRIB_DESCRIPTION="Ubuntu 16.04.6 LTS"
$ dpkg-query -W --showformat "\${Version}\n" sysv-rc
2.88dsf-59.3ubuntu2
$ update-rc.d -f rc.local remove
$ update-rc.d rc.local start 20 2 . stop 55 3 .
update-rc.d: warning: start and stop actions are no longer supported; falling back to defaults
update-rc.d: warning: start runlevel arguments (2) do not match rc.local Default-Start values (2 3 4 5)
update-rc.d: warning: stop runlevel arguments (3) do not match rc.local Default-Stop values (none)
$ update-rc.d -f rc.local remove
```

Debian 6 & 7 trigger dependency-based boot warning and ignore priorities, but links are generated as per LSB init comments (so this is identical to `rc-update.d <name> defaults`):

```
$ dpkg-query -W --showformat "\${Version}\n" sysv-rc
2.88dsf-13.1+squeeze1
$ update-rc.d -f rc.local remove
update-rc.d: using dependency based boot sequencing
$ update-rc.d rc.local start 20 2 . stop 55 3 .
update-rc.d: using dependency based boot sequencing
update-rc.d: warning: rc.local start runlevel arguments (2) do not match LSB Default-Start values (2 3 4 5)
update-rc.d: warning: rc.local stop runlevel arguments (3) do not match LSB Default-Stop values (none)
$ find /etc/rc?.d/ -name '???rc.local'
/etc/rc2.d/S01rc.local
/etc/rc3.d/S01rc.local
/etc/rc4.d/S01rc.local
/etc/rc5.d/S01rc.local
```

Debian 8 ships with up-to-date update-rc.d:

```
$ dpkg-query -W --showformat "\${Version}\n" sysv-rc
2.88dsf-59
$ update-rc.d -f rc.local remove
$ update-rc.d rc.local start 20 2 . stop 55 3 .
update-rc.d: warning: start and stop actions are no longer supported; falling back to defaults
update-rc.d: warning: start runlevel arguments (2) do not match rc.local Default-Start values (2 3 4 5)
update-rc.d: warning: stop runlevel arguments (3) do not match rc.local Default-Stop values (none)
$ find /etc/rc?.d/ -name '???rc.local'
/etc/rc2.d/S01rc.local
/etc/rc3.d/S01rc.local
/etc/rc4.d/S01rc.local
/etc/rc5.d/S01rc.local
```

I don't have access to older releases, but I think this is sufficient to confirm the version checks will select appropriate syntax. Also keep in mind I'm doing this in docker so there is no proper init system as there would be in full OS, however it's still viable for tests (`expect(@my_service).to be_disabled`).

lamont-granquist:
So the thing that would concern me is that we also support crazy things like raspbian and mint which are debian platform_family but not debian or ubuntu and don't track the platform_version of debian or ubuntu.

lamont-granquist:
And you can't necessarily assume that you only have to worry about those since anyone could patch in other platforms into ohai and map them to the debian platform_family.  I'm pretty certain we have "prior art" as well where we check the actual `--version` off of things and then use that to switch the behavior of providers.  We also have run into users who have upgraded the package manager on their system so that it isn't necessarily a given that a version of a package manager will be tied to the distro version (and i've done that before in a prior life as a system engineer).

robuye:
Yeah, I hear you and I considered other options too.

Initially I didn't include legacy syntax at all because there isn't any reliable way to detect it. There's no `--version` flag for `update-rc.d` and not much of a versioning at all to rely on. I was also considering parsing `--help` output, but that didn't feel better than the version check.

If `update-rc.d` would only be managing links in `/etc/rc?.d/` then it could be fairly easily implemented in Ruby, but at this point it would probably better fit the `init` provider. Since we ended up in `:debian` I assumed it's safe to rely on core Debian utilities such as `dpkg`. Also after another look at the code I should probably include a requirement on it for `:enable` and `:disable` actions too (not sure if this would affect provider resolution).

Keep in mind that `update-rc.d` is a core Debian package closely coupled to the init system. If another distribution diverges to mimic outdated Debian's `update-rc.d` on top of custom init then in my book it's okay if Chef gets confused, I would too (:

I think this comes down to support for run levels and priorities, but Debian itself moved away from it really long time ago (https://lists.debian.org/debian-devel-announce/2008/01/msg00004.html and also this one, linked in commit message too https://lists.debian.org/debian-devel/2013/05/msg01109.html). Currently, a command that was used to disable a service enables it instead and `:disable` action basically does the opposite of what it should, although I didn't do a system boot to be certain if links in `/etc/rc?.d/` are meaningful.

I tried to make it a forward-compatible change and if a known old package is not found it triggers "modern" syntax. If some ancient release shipped `update-rc.d` in another package that will still fail and I could also try parsing `--help` to account for that.

Yet another option is to not change the provider at all, only resolver so current & confirmed distros versions do not return `:debian`.

Lastly doing nothing is okay too, it's not difficult to work around it with simple `execute` resource. I just happened to be curious and wanted to find the root cause. It's been a pretty cool deep dive into Chef codebase, definitely fun and worthwhile regardless how we move forward!

lamont-granquist:
We're going to open master as chef-16 after shipping the next chef-15 version, this is probably a good candidate for that.

I'm real close to considering just merging it into chef-15 since it seems to be that we only support outdated distros that are EOL, but I still worry about people running EOL distros in production complaining on a minor version update that everything broke.

lamont-granquist:
(SemVer has truly horrible implications and this is why I mostly hate it, particularly when it is used blindly, which is mostly what users expect)

robuye:
> I'm real close to considering just merging it into chef-15 since it seems to be that we only support outdated distros that are EOL, but I still worry about people running EOL distros in production complaining on a minor version update that everything broke.

I think it's reasonable to hold on with merging this P/R until Chef 16, just to be on the safe side. Do you feel it would be worth to _deprecate_  the provider so it can be cleanly removed in Chef 17 or replaced with something more suitable? I'm cool to add such deprecation to the P/R, or any other changes that would fit in here, just let me know. Thanks for looking into it again.

lamont-granquist:
I think we just merge this into Chef-16.  I'm a lot less worried about that due to the distros being EOL and update-rc.d being a bit of an edge case anyway.

Issue #8859, opened 2019-09-03
Remove chef bootstrap script from command line / ps output


btm:
When bootstrapping in Chef Infra 15, the chef-full script is now run through train, for the most part fixing #3871. However https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap/templates/chef-full.erb still is wrapped by a `sh -c '... '` block which means the contents end up in ps output.

We should be able to simply remove that block from the template, but will need to test and verify tests.

lamont-granquist:
#8885 fixed

Issue #3887, opened 2015-09-07
Single quotes in attributes lead to bash syntax error in knife bootstrap


josephta:
If the initial attributes used in with `knife bootstrap` contain an apostrophe / single quote, then a bash syntax error is throw when chef client is started on the remote machine. 

Running with a "nice" set of attributes. Bootstrap connects to the node and tries to trigger the recipe (which fails as expected as the chef server isn't running):

```
$ knife bootstrap host-redacted -j "{\"x\":\"y\"}"
Bootstrapping Chef on host-redacted
host-redacted Starting Chef Client, version 12.1.2
host-redacted [2015-09-03T02:11:37-07:00] ERROR: Connection refused connecting to http://127.0.0.1:8889/nodes/host-redacted, retry 1/5
ˆC
```

Then, running the same thing with an apostrophe in a valid location:

```
$ knife bootstrap host-redacted -j "{\"x\":\"y'\"}"
Bootstrapping Chef on host-redacted
host-redacted bash: -c: line 80: unexpected EOF while looking for matching `"'
host-redacted bash: -c: line 84: syntax error: unexpected end of file
```

Spotted in versions 11.8.2 and 12.1.2.

josephta:
It looks like the issue may be with the lack of single quote escaping when generating the template at https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L390 as the default template (https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap/templates/chef-full.erb) uses single quotes.

This opens up for code injection attacks:

    $ knife bootstrap localhost -j "{\"a\": \"b'; echo Hello there; pwd; '\"}"
    Doing old-style registration with the validation key at /root/.chef/chef-validator.pem...
    Delete your validation key in order to use your user credentials instead
    
    Connecting to localhost
    localhost bash: line 80: warning: here-document at line 79 delimited by end-of-file (wanted `EOP')
    localhost Hello there
    localhost /root
    localhost bash: line 85: ","run_list":[]}
    localhost EOP
    localhost 
    localhost echo "Starting first Chef Client run..."
    localhost 
    localhost chef-client -j /etc/chef/first-boot.json -E _default: No such file or directory

lamont-granquist:
Since you're bootstrapping the host you've already got all the rights in the world to inject arbitrary code as the user you're bootstrapping the host with, so not sure I'd consider this an 'attack'.

This will be fixed by #3920 which will eliminate the horrible bash command.

[...]

PR #8885, opened 2019-09-13
Removing sh -c wrapper around the bootstrapper command string

Nimesh-Msys:
Avoids leaking of the secret key on console output while bootstrapping

## Description

- Train handles it and they are no more required here
- Tested manually found no impacts
- Ensured the test cases

## Related Issue

#8859 
#3887 

## Types of changes

## Checklist:


lamont-granquist:
I think this is a breaking change and that we should still support the validation key @marcparadise ?

The issue is that we should remove the `sh -c` since I think we now ship the script over scp and don't execute it entirely over ssh?  And that way we keep all of the contents of the file off of the ps listing, which this change wouldn't accomplish.

Nimesh-Msys:
Tested on the latest version (tested on 15.4.2), it does not display the keys. Checking the prescribed changes for prior versions.

Nimesh-Msys:
> The issue is that we should remove the `sh -c` since I think we now ship the script over scp and don't execute it entirely over ssh? And that way we keep all of the contents of the file off of the ps listing, which this change wouldn't accomplish.

Thanks for the review. Have made the required changes and tested.

PR #8887, opened 2019-09-14
kernel_module: Add new options property

ramereth:
This adds a new parameter called ``options`` which allows users to set module
specific parameters and settings. This option file needs to be run before
loading the module so that the settings get loaded properly.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/kernel_module.rb
+++ b/lib/chef/resource/kernel_module.rb
@@ -20,6 +20,9 @@ class Chef
         description: "An optional property to set the kernel module name if it differs from the resource block's name.",
         name_property: true, identity: true
 
+      property :options, Array,
+        description: "An optional property to set options for the kernel module."
+
       property :load_dir, String,
         description: "The directory to load modules from.",
         default: "/etc/modules-load.d"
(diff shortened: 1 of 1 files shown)


lib/chef/resource/kernel_module.rb
```diff
@@ -20,6 +20,9 @@ class KernelModule < Chef::Resource
         description: "An optional property to set the kernel module name if it differs from the resource block's name.",
         name_property: true, identity: true
 
+      property :options, Array,
```
tas50: This needs an introduced: 15.4

lib/chef/resource/kernel_module.rb
```diff
@@ -15,13 +15,60 @@ class KernelModule < Chef::Resource
...
+        end
+        ```
+      DOC
 
       property :modname, String,
         description: "An optional property to set the kernel module name if it differs from the resource block's name.",
         name_property: true, identity: true
 
       property :options, Array,
         description: "An optional property to set options for the kernel module."
```
tas50: missing a comma at the end here

tas50:
Thanks for the quick turnaround @ramereth

Issue #8731, opened 2019-07-10
Habitat Builds for Windows


btm:
As an enterprise user of Chef Infra, I need a Habitat package for Windows so I can move to effortless.

Acceptance Criteria:

habitat builds of chef infra client are:

- built in parallel of the omnibus builds
- cause the whole pipeline to appropriately fail if they fail
- run the same test suite as the omnibus packages
 - ~~are promoted in parallel with the omnibus builds~~ pending: https://github.com/chef/expeditor/issues/1132

Aha! Link: https://chef.aha.io/epics/CHEF-E-71

robbkidd:
A solid, official chef-infra-client Habitat package for Windows relies on a solid Ruby package. Ruby being what it is requires fiddling with various environment variables containing file paths (e.g. `GEM_PATH`, `SSL_CERT_FILE`). Habitat+Windows has an [open issue to figure out how to resolve paths at package build time to be able to set paths that are reliable at runtime](https://github.com/habitat-sh/habitat/issues/6483). There is a [design proposal](https://github.com/habitat-sh/habitat/issues/6837) for potential solutions.

robbkidd:
Update!

✅  I've got a `ruby-plus-devkit` (plus git) package that is basically the analog of Stuart's ruby-with-devkit. It can `bundle install` projects that have git references! (because I made it even bigger by baking the DevKit's (a.k.a. MSYS2's) git package into it.

✅  I've got a `chef-infra-client package` that bundle installs. 🎉 And almost appbundles ...

😢  ... but, about that appbundling, running `hab pkg exec robbkidd/chef-infra-client chef-client.bat -v` gets me an error about it being unable to find the "addressable" gem (the first one listed in the precise appbundled manifest).

I'm finally at the point where I need to figure out how appbundler works. I am distressed to discover an `unless A && B .. else ..end` block is [where the attempted use of gems](https://github.com/chef/appbundler/blob/master/lib/appbundler/app.rb#L303-L304) lives. My brain is too tired at the moment to parse that, but I suspect that the `nil`ing out of `GEM_PATH` is a problem within this Habitat package whose gems have been collected into paths that have been added to the `GEM_PATH`.

In a parallel effort—while waiting on these iterations of builds—[the King in the North](https://github.com/TheLunaticScripter) and I started experimenting with a `ruby` plan built from source with `visual-cpp-build-tools-2015` to see how far we could get with a more Hab-philosophically-pure build chain. We only started this afternoon, so progress was short but promising.

[...]

PR #8900, opened 2019-09-18
a Windows Habitat plan

robbkidd:
## Description

A Windows Habitat plan to produce Windows-flavored `chef/chef-infra-client`.

## Related Issue

#8731 

# ~Blockers~ ✅ 

- [x] resolve how to get the system's existing `%PATH%` to be included at the end of the `PATH` used in the habitat execution environments
  - **within a supervised process**: The `PATH` set by the thing that starts the hab-supervisor might already [be included in the execution environment](https://github.com/habitat-sh/habitat/blob/5d4fc16c7abce0a2ffda505ba6518d042d8f889b/components/common/src/util/path.rs#L148-L151). I think.
  - **when executed via hab-generated binlink**: Works as of 0.85.0
  - **when executed via hab pkg exec**: Might not be needed? See https://github.com/habitat-sh/habitat/issues/6657#issuecomment-530106046
- [x] an official home for the `ruby-plus-devkit` plan to live, be reviewed, and institute a pipeline around, related to (probably a sub-task of) https://github.com/chef/release-engineering/issues/831

# TODO ✔️ 

- [x] add tests to `habitat/tests/` for validating package post-build, in the spirit of https://github.com/chef/effortless/tree/master/scaffolding-chef-infra/tests
- ~Judicious testing with the effortless scaffolding~
- [x] rebase to latest master and squash commits appropriately to make a clearer story

## Checklist:

### DIFF
--- a/habitat/plan.ps1
+++ b/habitat/plan.ps1
@@ -30,6 +30,7 @@ function Invoke-Begin {
 function Invoke-SetupEnvironment {
     Push-RuntimeEnv -IsPath GEM_PATH "$pkg_prefix/vendor"
 
+    Set-RuntimeEnv APPBUNDLER_ALLOW_RVM "true" # prevent appbundler from clearing out the carefully constructed runtime GEM_PATH
     Set-RuntimeEnv -IsPath SSL_CERT_FILE "$(Get-HabPackagePath cacerts)/ssl/cert.pem"
     Set-RuntimeEnv LANG "en_US.UTF-8"
     Set-RuntimeEnv LC_CTYPE "en_US.UTF-8"
--- a/habitat/plan.ps1
+++ b/habitat/plan.ps1
@@ -1,0 +1,138 @@
+$pkg_name="chef-infra-client"
+$pkg_origin="chef"
+$pkg_version=(Get-Content $PLAN_CONTEXT/../VERSION)
+$pkg_description="Chef Infra Client is an agent that runs locally on every node that is under management by Chef Infra. This package is binary-only to provide Chef Infra Client executables. It does not define a service to run."
+$pkg_maintainer="The Chef Maintainers <maintainers@chef.io>"
+$pkg_upstream_url="https://github.com/chef/chef"
+$pkg_license=@("Apache-2.0")
+$pkg_filename="${pkg_name}-${pkg_version}.zip"
+$pkg_bin_dirs=@(
+    "bin"
+    "vendor/bin"
+)
+$pkg_deps=@(
+  "core/cacerts"
+  "chef/ruby-plus-devkit"
+)
+
+function Invoke-Begin {
+    $hab_version = (hab --version)
+    $hab_minor_version = $hab_version.split('.')[1]
+    if ( -not $? -Or $hab_minor_version -lt 85 ) {
+        Write-Warning "(╯°□°）╯︵ ┻━┻ I CAN'T WORK UNDER THESE CONDITIONS!"
+        Write-Warning ":habicat: I'm being built with $hab_version. I need at least Hab 0.85.0, because I use the -IsPath option for setting/pushing paths in SetupEnvironment."
+        throw "unable to build: required minimum version of Habitat not installed"
+    } else {
+        Write-BuildLine ":habicat: I think I have the version I need to build."
+    }
+}
+
+function Invoke-SetupEnvironment {
+    Push-RuntimeEnv -IsPath GEM_PATH "$pkg_prefix/vendor"
+
+    Set-RuntimeEnv -IsPath SSL_CERT_FILE "$(Get-HabPackagePath cacerts)/ssl/cert.pem"
+    Set-RuntimeEnv LANG "en_US.UTF-8"
+    Set-RuntimeEnv LC_CTYPE "en_US.UTF-8"
+}
+
+function Invoke-Download() {
+    Write-BuildLine " ** Locally creating archive of latest repository commit at ${HAB_CACHE_SRC_PATH}/${pkg_filename}"
+    # source is in this repo, so we're going to create an archive from the
+    # appropriate path within the repo and place the generated tarball in the
+    # location expected by do_unpack
+    try {
+        Push-Location (Resolve-Path "$PLAN_CONTEXT/../").Path
+        git archive --format=zip --output="${HAB_CACHE_SRC_PATH}/${pkg_filename}" HEAD
+    } finally {
+        Pop-Location
+    }
+}
+
+function Invoke-Verify() {
+    Write-BuildLine " ** Skipping checksum verification on the archive we just created."
+    return 0
+}
+
+function Invoke-Prepare {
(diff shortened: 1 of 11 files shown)


robbkidd:
Many commits here are unsigned because they could be temporary, interstitial commits before an interactive squashy rebase to clean up the history to document things in commit messages.

robbkidd:
Modeling chef-client hab-package verification after the omnibus build tests requires a test environment that acts like a "normal" host.

* a real Windows host (with the tests running as Administrator?)
* real Ubuntu/RHEL/etc hosts

The omnibus build validation is essentially a run of the chef-client RSpec suite scoped to the functional specs.

- https://github.com/chef/chef/blob/v15.3.14/omnibus/omnibus-test.ps1#L94
- https://github.com/chef/chef/blob/v15.3.14/omnibus/omnibus-test.sh#L157

A subset of those specs do not pass when run within a hab studio (which is the nice pattern demonstrated by the effortless scaffolding tests on linux: do it in a studio clean room).

Some studio/chef-client incompatibilities discovered so far:

😢 A Windows studio does not have access to read and/or change several system resources (e.g. the Registry).
😢 A Linux studio does not have `useradd`.
😢 `Chef::Resource::Bash` functional specs are unhappy in a Linux studio.

To replicate the build verification used for omnibus, we'll need to run the hab package verify scripts on "real" hosts. Currently figuring out how to wire that all up in a pipeline. It may seem gross, but at the moment, I think this is the correct path. We should verify that a hab-packaged chef-client's functional tests pass on a real host because the purpose of running a chef-client is to manage a real host, not a studio.

habitat/plan.ps1
```diff
@@ -0,0 +1,127 @@
...
+$pkg_maintainer="The Chef Maintainers <humans@chef.io>"
+$pkg_license=@("Apache-2.0")
+$pkg_filename="${pkg_name}-${pkg_version}.zip"
+$pkg_bin_dirs=@(
+    "bin"
+    "vendor/bin"
+)
+$pkg_deps=@(
+  "core/cacerts"
+  "robbkidd/ruby-plus-devkit"
```
btm: you should use `chef/ruby-plus-devkit` ;)

robbkidd: indeed

dec9dcb

robbkidd:
With the Verify action in here testing the Windows plan taking 30-40 minutes, maybe we should consider running that on PR merges like with omnibus builds. That means we won't know about Habitat package build breakages until merges to master ... but that's the same as with omnibus.

Also yet to be solved: how to coordinate promotion from unstable-to-current? A version's omnibus and habitat packages should all pass tests or none get promoted. The current subscription actions are set to promote the hab packages based on omnibus stability.

robbkidd:
> Also yet to be solved: how to coordinate promotion from unstable-to-current?

Some out-of-band conversations yesterday have us thinking that we'll let the omnibus packages and the habitat packages independently promote from unstable to stable based on their automated tests. The promotion from current to stable will require human coordination of versions being ready across the package repositories and lever-pulling.

btm:
@robbkidd this needs a rebase to not clobber the the 'Ruby 2.6' tests in `.expeditor/verify.pipeline.yml`, yeah? I don't think all that red is supposed to be in the diff? and squash some commits and fix the DCO?

lamont-granquist:
the appbundling and stuff looks correct.

you might want to throw `export FORCE_FFI_YAJL=ext` somewhere in the verification steps similar to:

https://github.com/chef/chef/blob/master/scripts/bk_tests/bk_linux_exec.sh#L22

that's about all i can think of other than the comments btm already made.

robbkidd:
Rebased with a mega-squash and signing. The first commit has a message with a lot of notes about life, the universe, and everything.

robbkidd:
* Regarding `export FORCE_FFI_YAJL=ext`

I see that the ffi_yajl gem says ...

> You can set the environment variable `FORCE_FFI_YAJL` to `ext` or `ffi` to control if the C extension or FFI bindings are used.

So I understand that setting is to `ext` tells the gem to use the built C-extensions. I'd like to understand why. How does that help us? Is it only something for the test environment? Is it something that should be set for production runtime, too?

* Regarding the swath of RED in the verify pipeline

I have left that commit unsigned because I don't intend to keep it. That's currently got this PR's Buildkite testing focused on verifying that this plan produces a package that works. The question is still whether we want that plan testing step—that can take 30-40 minutes—included in the `verify` pipeline that is triggered on every PR and every push to an open PR's branch.

Those are the same tests that I think would run to decide whether to promote an unstable hab package to a current channel. That "test unstable and promote to current" is not wired up in this PR (yet?).

omnibus/omnibus-test.ps1
```diff
@@ -82,14 +82,12 @@ Get-Location
...
-
 # some tests need winrm configured
 winrm quickconfig -quiet
 
 bundle
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 
-bundle exec rspec -r rspec_junit_formatter -f RspecJunitFormatter -o test.xml -f documentation spec/functional
+# chocolatey functional tests fail so disable that tag directly
+bundle exec rspec -r rspec_junit_formatter -f RspecJunitFormatter -o test.xml -f documentation --tag ~choco_installed spec/functional
```
robbkidd: This change—to directly turn off `choco_installed` specs instead of deleting the `choco.exe` to "uninstall" Chocolatey—to the omnibus test worked in an adhoc build.

robbkidd:
@btm Despite the test failures, I think this is ready. I expect the merge of this PR to build the hab packages and promote the Windows one to current. 

halloween shrug

scripts/ci/verify-plan.ps1
```diff
@@ -0,0 +1,37 @@
...
+    [string]$Plan
+)
+
+$env:HAB_ORIGIN = 'ci'
+$Plan = 'chef-infra-client'
+
+Write-Host "--- :8ball: :windows: Verifying $Plan"
+
+Write-Host "--- :habicat: Installing the version of Habitat required"
+Install-Habitat --version 0.85.0.20190916
```
btm: some day we'll probably not want this lost deep in here somehow.

robbkidd: Yup. I had a variant of this apparently lost in shuffling that would only attempt to install this version if the installed habitat version was below 0.85. I should bring that back. This and [the version test in `Invoke-Begin`](https://github.com/chef/chef/pull/8900/files#diff-54f992763c1ee2453d8e339101744a01R21-R27) are some early workarounds to demonstrate a possible missing piece of Habitat plan metadata: a minimum version of Habitat required for a plan to be built by.

robbkidd: Updated with an "install only if hab is not installed or is too old" wrapper.

robbkidd: Updated with an "install only if hab is not installed or is too old" wrapper.

robbkidd:
Force push to keep things up-to-date with master.
tenor-201774685

PR #8902, opened 2019-09-19
Remove the gemfile.local logic from the omnibus gemfile

tas50:
This allows dependabot to handle all our bumping PRs for omnibus and gives us really nice changelogs of the changes that are happening with each omnibus-software update.


lamont-granquist:
`instance_eval(ENV["GEMFILE_MOD"]) if ENV["GEMFILE_MOD"]`

that should be necessary for testing.  i don't think you can delete that even if the tests go green.

PR #8912, opened 2019-09-24
fix converge_if_changed to compare default values

### LINKED ISSUE #6300: converge_if_changed does not check desired_state of property with default that has not been accessed
## Description

When using `converge_if_changed`, Chef will not detect a change if a property is configured with a `default` and is `desired_state` if it is not explicitly defined in the resource usage.

The following line defines which properties to watch the desired state: 
- https://github.com/chef/chef/blob/264e3c5c12e1f8c15ed37aea3964957585c23601/lib/chef/provider.rb#L247

And ultimately utilizes the `is_set?` method which returns true if `The user explicitly set the value` and `The property has a default, and the value was retrieved.`
- https://github.com/chef/chef/blob/master/lib/chef/property.rb#L400

So in the event that a property has a default value, but the user has not retrieve the value, it will not be checked against the desired state.

A workaround is to simply access the value prior to `converge_if_changed` was executed, but leads to unnecessary code.

## Chef Version

Chef 12.21.3

## Platform Version

CentOS 7.2

## Replication Case

To replicate:
1. `git clone https://github.com/jakauppila/myTestCookbook`
2. `kitchen converge`
3. `kitchen exec -c "echo 'Goodbye World.' >> /home/vagrant/myfile.txt"`
4. `kitchen converge`

## Client Output

```
  [2017-07-31T03:09:40+00:00] DEBUG: Converging node default-centos-72
  Recipe: test::default_value
    * my_provider[/home/vagrant/myfile.txt] action add[2017-07-31T03:09:40+00:00] INFO: Processing my_provider[/home/vagrant/myfile.txt] action add (test::default_value line 1)
  [2017-07-31T03:09:40+00:00] INFO: my_provider[/home/vagrant/myfile.txt] already installed - nothing to do.
  (up to date)
    * my_provider[/home/vagrant/myfile.txt] action config[2017-07-31T03:09:40+00:00] INFO: Processing my_provider[/home/vagrant/myfile.txt] action config (test::default_value line 1)
  [2017-07-31T03:09:40+00:00] DEBUG: Skipping update of my_provider[/home/vagrant/myfile.txt]: has not changed any of the specified properties .
  (up to date)
```

## Expected Output

```
  [2017-07-31T03:12:37+00:00] DEBUG: Converging node default-centos-72
  Recipe: test::default_value
    * my_provider[/home/vagrant/myfile.txt] action add[2017-07-31T03:12:37+00:00] INFO: Processing my_provider[/home/vagrant/myfile.txt] action add (test::default_value line 1)
  [2017-07-31T03:12:37+00:00] INFO: my_provider[/home/vagrant/myfile.txt] already installed - nothing to do.
  (up to date)
    * my_provider[/home/vagrant/myfile.txt] action config[2017-07-31T03:12:37+00:00] INFO: Processing my_provider[/home/vagrant/myfile.txt] action config (test::default_value line 1)
  [2017-07-31T03:12:37+00:00] DEBUG: Resources for generic file resource enabled on node include: [Chef::Resource::File]
  [2017-07-31T03:12:37+00:00] DEBUG: Resource for file is Chef::Resource::File

      - update /home/vagrant/myfile.txt
      -   set contents to "Hello World." (was "Hello World.Goodbye World.\n")
      * file[/home/vagrant/myfile.txt] action create[2017-07-31T03:12:37+00:00] INFO: Processing file[/home/vagrant/myfile.txt] action create (/tmp/kitchen/cache/cookbooks/myTestCookbook/providers/my_provider.rb line 23)
```

lamont-granquist:
Fixes #6300.

This may be a breaking change.

### DIFF
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -272,11 +272,17 @@ class Chef
         raise ArgumentError, "converge_if_changed must be passed a block!"
       end
 
-      properties = new_resource.class.state_properties.map(&:name) if properties.empty?
-      properties = properties.map(&:to_sym)
+      properties =
+        if properties.empty?
+          new_resource.class.state_properties
+        else
+          properties.map { |property| new_resource.class.properties[property] }
+        end
+
       if current_resource
         # Collect the list of modified properties
-        specified_properties = properties.select { |property| new_resource.property_is_set?(property) }
+        specified_properties = properties.select { |property| property.is_set?(new_resource) || property.has_default? }
+        specified_properties = specified_properties.map(&:name).map(&:to_sym)
         modified = specified_properties.select { |p| new_resource.send(p) != current_resource.send(p) }
         if modified.empty?
           properties_str = if new_resource.sensitive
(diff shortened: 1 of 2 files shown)


tas50:
Technically breaking, but this is 100% what you would expect and is just a plain bug. People ask about this a lot and it makes the converge_if_changed something a lot of folks can't use. We can just call it out as a bug fix that potentially changes behavior.

lamont-granquist:
Yeah I'm heavily leaning that way, but at the same time it seems a little scary.

If you know of some major cookbooks with converge_if_changed that aren't broken by this or something that'd be good info to have.

Issue #8848, opened 2017-06-09
knife subcommand --help doesn't work as intended


moutons:
:evilchef:
```
# chef --version
Chef Development Kit Version: 1.3.43
...
# chef gem install knife-acl
Fetching: knife-acl-1.0.3.gem (100%)
Successfully installed knife-acl-1.0.3
1 gem installed
# knife group --help
FATAL: Cannot find subcommand for: 'group --help'
Available subcommands: (for details, knife SUB-COMMAND --help)
...
```

Vasu1105:
Latest chefdk `4.3.13` don't gives above error but it does not show the expected result. It should filter out only knife group commands from list, but it lists all the knife command when we run ```knife group --help ``` also same is the result when we run ```knife acl --help```

Vasu1105:
My bad it does give above error and then list the command

Vasu1105:
Fixed in #8915

PR #8915, opened 2019-09-25
Fix for knife subcommand --help don't work as intended. 

Vasu1105:

## Description
Fix for knife subcommand --help don't work as intended. If we run knife acl --help, knife group --help or knife opc --help command it does not list the commands even though this plugins are shipped with ChefDK. The reason behind this is the category under which these are listed does not 
match with the arguments passed by user. for e.g. acl commands are grouped under the category "OPSCODE_HOSTED_CHEF_ACCESS_CONTROL" and since acl as an argument doesn't match with this category as string, this does not gives valid output.

This methods groups the sub-command by category here https://github.com/chef/chef/blob/master/lib/chef/knife.rb#L165 and it fetches the category from the class for e.g. it has key and value like this in hash  "acl_remove"=>OpscodeAcl::AclRemove,
and klass here is OpscodeAcl::AclRemove and its fetched from here https://github.com/chef/knife-acl/blob/271ef64e30b5bad3f68a7e21cd940a2fe3bd289b/lib/chef/knife/acl_remove.rb#L22

## Related Issue

#8848 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -64,6 +64,12 @@ class Chef
     attr_accessor :name_args
     attr_accessor :ui
 
+    # knife acl subcommands are grouped in this category using this constant to verify.
+    OPSCODE_HOSTED_CHEF_ACCESS_CONTROL = %w{acl group user}.freeze
+
+    # knife opc subcommands are grouped in this category using this constant to verify.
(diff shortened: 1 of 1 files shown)


tas50:
The real fix is to get these into chef and cleanup the categories, but this fixes it for now.

Issue #8808, opened 2019-08-09
windows_firewall_rule doesn't support multiple profiles


Happycoil:
## Description
`windows_firewall_rule` doesn't support multiple profiles. The property needs to support the Array type.

## Chef Version
14.7+

## Platform Version
Windows

## Replication Case
Try to modify a standard rule with multiple profiles

## Client Output
```
Chef::Exceptions::ValidationFailed
----------------------------------------------------
Option profile must be equal to one of: :public, :private, :domain, :any, :notapplicable!  You passed :"private, public".
```

mrmarbury:
We are facing the same problem. The firewall rule we try to delete has the profile `Domain, Private` and there is no way deleting it. We get the same error message as the OP

tecracer-theinen:
Still not solved: The resources as of today's master branch will work for rules with just one associated profile, but fail validation of current value on those with multiple.

This is because the `load_firewall_state` method of the resource converts the profiles to a String in PowerShell, resulting in a current value like `:"domain, private"` for existing rules with multiple profiles. This is of course not a valid value and it fails validation even before executing the resource.

PowerShell return value:
```json
 {
  "rule_name": "WMPNSS-RME-HTTP-In-TCP",
  "description":"Inbound rule for the Windows Media Player Network Sharing Service to allow HTTP Media Streaming. [TCP 10245]",
  "group": "@FirewallAPI.dll,-31500",
  "local_address": "Any",
  "local_port": "10245",
  "remote_address": "Any",
  "remote_port": "Any",
  "direction": "Inbound",
  "protocol": "TCP",
  "firewall_action": "Allow",
  "profile": "Domain, Private",
  "program": "System",
  "service": "Any",
  "interface_type": "Any",
  "enabled": false
}
```

Failure during loading current state in Chef run:
```
 [2020-03-19T18:30:40-07:00] FATAL: Chef::Exceptions::ValidationFailed: windows_firewall_rule[WMPNSS-RME-HTTP-In-TCP] (appinstall-hardening::set_firewallrules line 28) had an error: Chef::Exceptions::ValidationFailed: Option profile's value [:"domain, private"] contains values not in :public, :private, :domain, :any or :notapplicable!
```

PR #8916, opened 2019-09-25
Support multiple profiles in windows_firewall_rule

Happycoil:
## Description
Adds Array support to the `profile` property

## Related Issue
https://github.com/chef/chef/issues/8808

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/windows_firewall_rule.rb
+++ b/lib/chef/resource/windows_firewall_rule.rb
@@ -74,7 +74,7 @@ class Chef
         callbacks: {
           "contains values not in :public, :private :domain, :any or :notapplicable" => lambda { |p|
             p.all? { |e| %i{public private domain any notapplicable}.include?(e) }
-          }
+          },
         }
 
       property :program, String,
--- a/lib/chef/resource/windows_firewall_rule.rb
+++ b/lib/chef/resource/windows_firewall_rule.rb
@@ -70,15 +70,8 @@ class Chef
       property :profile, [Symbol, String, Array],
         default: :any,
         description: "The profile the firewall rule applies to.",
-        coerce: proc { |p|
-          if p.is_a?(Array)
-            p.map(&:downcase).map(&:to_sym).sort
-          elsif p.is_a?(String)
-            Array(p.downcase.to_sym).sort
-          else
-            Array(p).sort
-          end
-        }
+        coerce: proc { |p| Array(p).map(&:downcase).map(&:to_sym).sort }
+
       property :program, String,
         description: "The program the firewall rule applies to."
 
--- a/lib/chef/resource/windows_firewall_rule.rb
+++ b/lib/chef/resource/windows_firewall_rule.rb
@@ -67,11 +67,18 @@ class Chef
         description: "The action of the firewall rule.",
         coerce: proc { |f| f.is_a?(String) ? f.downcase.to_sym : f }
 
-      property :profile, [Symbol, String],
-        default: :any, equal_to: %i{public private domain any notapplicable},
+      property :profile, [Symbol, String, Array],
+        default: :any,
         description: "The profile the firewall rule applies to.",
-        coerce: proc { |p| p.is_a?(String) ? p.downcase.to_sym : p }
-
+        coerce: proc { |p|
+          if p.is_a?(Array)
+            p.map(&:downcase).map(&:to_sym).sort
+          elsif p.is_a?(String)
+            Array(p.downcase.to_sym).sort
+          else
+            Array(p).sort
+          end
+        }
       property :program, String,
         description: "The program the firewall rule applies to."
 
@@ -158,7 +165,7 @@ class Chef
           cmd << " -Direction '#{new_resource.direction}'" if new_resource.direction
           cmd << " -Protocol '#{new_resource.protocol}'" if new_resource.protocol
           cmd << " -Action '#{new_resource.firewall_action}'" if new_resource.firewall_action
-          cmd << " -Profile '#{new_resource.profile}'" if new_resource.profile
+          cmd << " -Profile '#{new_resource.profile.join("', '")}'" if new_resource.profile
           cmd << " -Program '#{new_resource.program}'" if new_resource.program
           cmd << " -Service '#{new_resource.service}'" if new_resource.service
           cmd << " -InterfaceType '#{new_resource.interface_type}'" if new_resource.interface_type
--- a/lib/chef/resource/windows_firewall_rule.rb
+++ b/lib/chef/resource/windows_firewall_rule.rb
@@ -70,7 +70,12 @@ class Chef
       property :profile, [Symbol, String, Array],
         default: :any,
         description: "The profile the firewall rule applies to.",
-        coerce: proc { |p| Array(p).map(&:downcase).map(&:to_sym).sort }
+        coerce: proc { |p| Array(p).map(&:downcase).map(&:to_sym).sort },
+        callbacks: {
+          'contains values not in :public, :private :domain, :any or :notapplicable' => lambda {
+            |p| p.all? { |e| %i{public private domain any notapplicable}.include?(e) }
+          }
+        }
 
       property :program, String,
         description: "The program the firewall rule applies to."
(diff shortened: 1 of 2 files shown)


Happycoil:
There's a potential issue with sorting the profile names here, that means the resource will always converge unless the user also sorts the property correctly, right?

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -158,7 +165,7 @@ def firewall_command(cmdlet_type)
           cmd << " -Direction '#{new_resource.direction}'" if new_resource.direction
           cmd << " -Protocol '#{new_resource.protocol}'" if new_resource.protocol
           cmd << " -Action '#{new_resource.firewall_action}'" if new_resource.firewall_action
-          cmd << " -Profile '#{new_resource.profile}'" if new_resource.profile
+          cmd << " -Profile '#{new_resource.profile.join("', '")}'" if new_resource.profile
```
robbkidd: I have not used this command in anger, so I looked up [the documentation for New-NetFirewallRule](https://docs.microsoft.com/en-us/powershell/module/netsecurity/new-netfirewallrule?view=win10-ps#parameters). It states for the `-Profile` parameter:

> Separate multiple entries with a comma and do not include any spaces.

Have you seen this change work on a Windows host? I see that the unit test was updated (and so passes) to check that the generated command includes `-Profile 'private', 'public'`, but will that space cause an error at run time?

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -67,11 +67,18 @@ class WindowsFirewallRule < Chef::Resource
...
-        coerce: proc { |p| p.is_a?(String) ? p.downcase.to_sym : p }
-
+        coerce: proc { |p|
+          if p.is_a?(Array)
+            p.map(&:downcase).map(&:to_sym).sort
+          elsif p.is_a?(String)
+            Array(p.downcase.to_sym).sort
+          else
+            Array(p).sort
+          end
```
lamont-granquist: `coerce: proc { |p| Array(p).map(&:downcase).map(&:to_sym).sort }` will probably do this without the conditional.

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -67,11 +67,18 @@ class WindowsFirewallRule < Chef::Resource
         description: "The action of the firewall rule.",
         coerce: proc { |f| f.is_a?(String) ? f.downcase.to_sym : f }
 
-      property :profile, [Symbol, String],
-        default: :any, equal_to: %i{public private domain any notapplicable},
```
lamont-granquist: you should still validate that the array does not contain anything that isn't one of these symbols?

lib/chef/resource/windows_firewall_rule.rb
Happycoil: Didn't get that far yet, but it appears you're right. If they've written the cmdlet like that it's definitely not idiomatic PowerShell... I was about to say that this explains why LocalPort and RemotePort are written that way in the resource, but those parameters are actually documented as taking `String[]`

lib/chef/resource/windows_firewall_rule.rb
Happycoil: Yeah, that's much better

lib/chef/resource/windows_firewall_rule.rb
Happycoil: How though? I made it work with the below snippet, but is there a simpler approach?
```ruby
        callbacks: {
                      'contains values not in :public, :private :domain, :any or :notapplicable' => lambda {
                        |p| %i{public private domain any notapplicable} & p == p
                      }
                   }
```

lib/chef/resource/windows_firewall_rule.rb
Happycoil: So turns out both approaches work for `Profile`, whereas the comma delimited string the resource uses for `LocalPort` and `RemotePort` actually doesn't work.

Happycoil:
Not all review comments are addressed, but I'm happy with the PR when they're resolved @robbkidd @lamont-granquist

lib/chef/resource/windows_firewall_rule.rb
lamont-granquist: that may be overly terse.  `lambda { |p| p.all? { |e| %i{public private domain any notapplicable}.include?(e) } }` or something like that which doesn't require realizing those are set operators.

spec/unit/resource/windows_firewall_rule_spec.rb
```diff
@@ -270,6 +302,11 @@
         expect(provider.firewall_command("New")).to eql("New-NetFirewallRule -Name 'test_rule' -DisplayName 'test_rule' -Description 'Firewall rule' -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'private' -InterfaceType 'any' -Enabled 'true'")
       end
 
+      it "sets multiple Profiles" do
```
robbkidd:
```suggestion
      it "sets multiple Profiles (must be comma-plus-space delimited for PowerShell to treat as an array)" do
```

spec/unit/resource/windows_firewall_rule_spec.rb
```diff
@@ -326,7 +363,17 @@
...
-        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -LocalPort 80 -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -LocalPort '80' -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+      end
+
+      it "sets LocalPort with int" do
+        resource.local_port(80)
+        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -LocalPort '80' -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+      end
+
+      it "sets multiple LocalPorts" do
```
robbkidd:
```suggestion
      it "sets multiple LocalPorts (must be comma-plus-space delimited for PowerShell to treat as an array)" do
```

spec/unit/resource/windows_firewall_rule_spec.rb
```diff
@@ -336,7 +383,17 @@
...
-        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -RemotePort 443 -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -RemotePort '443' -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+      end
+
+      it "sets RemotePort with int" do
+        resource.remote_port(443)
+        expect(provider.firewall_command("Set")).to eql("Set-NetFirewallRule -Name 'test_rule' -Description 'Firewall rule' -RemotePort '443' -Direction 'inbound' -Protocol 'TCP' -Action 'allow' -Profile 'any' -InterfaceType 'any' -Enabled 'true'")
+      end
+
+      it "sets multiple RemotePorts" do
```
robbkidd:
```suggestion
      it "sets multiple RemotePorts (must be comma-plus-space delimited for PowerShell to treat as an array)" do
```

lib/chef/resource/windows_firewall_rule.rb
robbkidd: Well, thanks for finding that out and fixing them!

PR #8920, opened 2019-09-26
don't even think of merging or reviewing.

lamont-granquist:
i have an enormous rebase mess to figure out how to clean up and i just want to be able to look at git diffs because for some reason i like GUIs more than the command line for that...


lamont-granquist:
guess i could have just preemptively closed this

Issue #8820, opened 2019-08-14
`.to_yaml` does not do what one would expect


jaymzh:
## Description

`node['foo'].to_yaml`  dumps the entire node object and countless internal references. It should be the same as `.to_json`, and collapse the attributes below that and then generate yaml.

## Chef Version
15.1.36

## Platform Version
Ubuntu 18.04

## Replication Case
```
node.default['test'] = {'bar' => ['baz'], 'bing' => 'boo'}
node['test'].to_yaml
```

coding-blip:
I have seen this issue too, though not directly the same I don't think. What I did to get around it was to first convert to_json and then convert that to_yaml. Not sure if something like this will work for you or not. Who knows, maybe we can both learn a better way to do it! 

require 'yaml'

my_hash = {}
node['thing']['config'].each do |config_item, config_value|
  my_hash.merge!(config_item => config_value)
end

# Convert to JSON then YAML in order to process properly
file "#{node['thing']['config_dir']}/config.yml" do
  content JSON.parse(es_config.to_json).to_yaml.lines.to_a[1..-1].join
end

jaymzh:
oh my work around was simple: `node['test'].to_json.to_yaml` ... but the `to_yaml` method needs to get fixed.

mateuszkwiatkowski:
It seems that this is caused by yaml (psych) module and is not specific to Chef. Nodes attributes are instance of `Chef::Node::ImmutableMash` and they can be converted to Hash by `.to_h` method. Json module checks if object responds to `.to_h` or `.to_hash` methods and converts it before doing convertion to JSON. I checked psych source code and didn't find similar logic in there.

`node['test'].to_h.to_yaml` works too.

lamont-granquist:
likely Chef::Node::Attribute just needs:

```
def to_yaml(*opts)
  to_hash.to_yaml(*opts)
end
```

vsingh-msys:
Fixed in #8927 `v15.4.13`

PR #8927, opened 2019-09-27
Add #to_yaml method for ImmutableMash & ImmutableArray

vsingh-msys:
## Description
 - Psych module to_yaml not working as expected for ImmutableMash & ImmutableArray.
 - So parse first to Hash/Array then convert it to YAML formatted string.
eg.
1. If hash is assgined to node attribute it is returning ImmutableMash
```
> node.default['test'] = {'bar' => ['baz'], 'bing' => 'boo'}
> node['test'].class # ImmutableMash

```

2. If Array is assigned to node attribute it is returning ImmutableArray
```
> node.default['test'] = [{'bar' => ['baz']}, {'bing' => 'boo'}]
> node['test'].class # ImmutableArray

```
3. If any other value is assigned to node `to_yaml` working as expected.

## Related Issue

https://github.com/chef/chef/issues/8820

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/node/immutable_collections.rb
+++ b/lib/chef/node/immutable_collections.rb
@@ -96,7 +96,7 @@ class Chef
 
       alias_method :to_array, :to_a
 
-      # As Psych module, not respecting Immutable ImmutableArray object
+      # As Psych module, not respecting ImmutableArray object
       # So first convert it to Hash/Array then parse it to `.to_yaml`
       def to_yaml(*opts)
         to_a.to_yaml(*opts)
--- a/lib/chef/node/immutable_collections.rb
+++ b/lib/chef/node/immutable_collections.rb
@@ -96,6 +96,12 @@ class Chef
 
       alias_method :to_array, :to_a
 
+      # As Psych module, not respecting Immutable ImmutableArray object
+      # So first convert it to Hash/Array then parse it to `.to_yaml`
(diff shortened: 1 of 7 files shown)


lib/chef/node/immutable_collections.rb
```diff
@@ -168,6 +174,12 @@ def to_h
 
       alias_method :to_hash, :to_h
 
+      # As Psych module, not respecting ImmutableMash object
+      # So first convert it to Hash/Array then parse it to `.to_yaml`
+      def to_yaml(*opts)
+        to_h.to_yaml(*opts)
+      end
```
lamont-granquist: should probably add this API for the VividMash/ArrayAttr classes as well, since users can get at those out of the components.  i would think we'd need to add something in Chef::Node::Attribute as well to make this change visible.

vsingh-msys:
> should probably add this API for the VividMash/ArrayAttr classes as well since users can get at those out of the components. i would think we'd need to add something in Chef::Node::Attribute as well to make this change visible.

- Have added `to_yaml` method for VividMash/AttrArray and also for Chef::Node::Attribute.
- Fix `to_json` is missing Chef::Node::Attribute that raises error `Wrong number of an argument while calling on object`
- Refactor the specs using a shared example.

@lamont-granquist  could you please have a look again?

Thanks!

Issue #7042, opened 2018-03-23
sudo file validation using verify doesn't appear to run


tas50:
I was able to write out a bad config which tells me that visudo is not being called correctly. We need to confirm if this is running or not.

tas50:
This was resolved with the factoring of the sudo config validation in #8928

PR #8928, opened 2019-09-30
sudo: perform config validation on the overall sudoers state

### LINKED ISSUE #8763: sudo resource does not check if command aliases are already in use
## Description
The sudo resource (as of chef-client 14.4.56) does not check whether a command alias is already defined before adding it to a file within sudoers.d. This will cause sudo to break, unable to parse the sudoers config.

## Chef Version
14.4.56

## Platform Version
CentOS/RHEL 7

## Replication Case
1. Create a sudo resource in a recipe which defines some command alias along with the user who should run it
2. Make a separate file in sudoers.d which contains the identical Cmnd_Alias to what chef-client will produce in its sudoers.d file
3. Run chef-client
4. Attempt to sudo will fail, as the Cmnd_Alias is duplicated in two files

samshinde:
Signed-off-by: Amol Shinde <amol.shinde@msystechnologies.com>

## Description

`visudo` was only checking the individual file which is being generated using sudo resource, but it was not checking overall state of sudoers.
- Added overall state verification with `visudo`

## Related Issue

Fixes #8763 #7042
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/sudo.rb
+++ b/lib/chef/resource/sudo.rb
@@ -166,7 +166,7 @@ class Chef
             source new_resource.template
             mode "0440"
             variables new_resource.variables
-            verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?
+            verify "cat #{new_resource.config_prefix}/sudoers %{path} | #{new_resource.visudo_binary} -cf -" if visudo_present?
             action :create
           end
         else
@@ -185,7 +185,7 @@ class Chef
                       setenv:             new_resource.setenv,
                       env_keep_add:       new_resource.env_keep_add,
                       env_keep_subtract:  new_resource.env_keep_subtract
-            verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?
+            verify "cat #{new_resource.config_prefix}/sudoers %{path} | #{new_resource.visudo_binary} -cf -" if visudo_present?
             action :create
           end
         end


lib/chef/resource/sudo.rb
```diff
@@ -167,6 +167,7 @@ def platform_config_prefix
             mode "0440"
             variables new_resource.variables
             verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?
```
tas50: I think we can actually simplify this a bit:
```ruby
verify "#{new_resource.visudo_binary} -cf #{new_resource.config_prefix}/sudoers" if visudo_present?
```

sudoers will parse in all the .d files so there's no need to parse both the sudoers config and the .d file we're working on now:

```
$ visudo -cf /etc/sudoers
/etc/sudoers: parsed OK
/etc/sudoers.d/90-cloud-init-users: parsed OK
/etc/sudoers.d/README: parsed OK```

Comment (lamont-granquist) [reply lib/chef/resource/sudo.rb]: you have to combined the sudoers.d with the %{path} file though not the old /etc/sudoers file.

either way though that line should get replaced.

Comment (samshinde) [reply lib/chef/resource/sudo.rb]: I have removed `verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?`, because if newly added sudo resource is invalid then the overall state will be invalid.

File: lib/chef/resource/sudo.rb [hunk 2]
```diff
@@ -166,7 +166,7 @@ def platform_config_prefix
             source new_resource.template
             mode "0440"
             variables new_resource.variables
-            verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?
+            verify "cat #{new_resource.config_prefix}/sudoers %{path} | #{new_resource.visudo_binary} -cf -" if visudo_present?
```
tas50:
```suggestion
            verify " #{new_resource.visudo_binary} -c" if visudo_present?
```

lib/chef/resource/sudo.rb
```diff
@@ -185,7 +185,7 @@ def platform_config_prefix
                       setenv:             new_resource.setenv,
                       env_keep_add:       new_resource.env_keep_add,
                       env_keep_subtract:  new_resource.env_keep_subtract
-            verify "#{new_resource.visudo_binary} -cf %{path}" if visudo_present?
+            verify "cat #{new_resource.config_prefix}/sudoers %{path} | #{new_resource.visudo_binary} -cf -" if visudo_present?
```
tas50:
```suggestion
             verify " #{new_resource.visudo_binary} -c" if visudo_present?
```

samshinde: I did the suggested change and created file `foo` in sudoers.d/ with below content.
```bash
Cmnd_Alias TEST = /usr/bin/ls, /usr/bin/cat
admin ALL=(ALL) ALL
```
Now, When I am running cookbook with `sudo` resource which have the same `Cmnd_Alias`, it runs successfully, but after that, it breaks the `sudo` command
```bash
[vagrant@localhost chef]$ sudo -s
>>> /etc/sudoers.d/foo: Alias "TEST" already defined near line 4 <<<
sudo: parse error in /etc/sudoers.d/foo near line 4
sudo: no valid sudoers sources found, quitting
sudo: unable to initialize policy plugin
```
So, we have to merge the content of `/etc/sudoers` file and the current .d file using 
```ruby
cat #{new_resource.config_prefix}/sudoers %{path} | #{new_resource.visudo_binary} -cf -" if visudo_present?
```

PR #8929, opened 2019-09-30
Fix policy settings from config file

dheerajd-msys:

- Chef Client cookbook doesn't address the policy file settings like `policy_name` & `policy_group` when given in `config.rb/knife.rb`

## Description
- Updated options `policy_name` & `policy_group`  so that it can fetch the details from `config`
## Related Issue
Fixes https://github.com/chef-cookbooks/chef-client/issues/588
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -241,12 +241,14 @@ class Chef
       option :policy_name,
         long: "--policy-name POLICY_NAME",
         description: "Policyfile name to use (--policy-group must also be given).",
+        proc: Proc.new { |u| Chef::Config[:knife][:policy_name] = u },
         default: nil
 
       # runtime - client_builder - set policy group when creating node
       option :policy_group,
         long: "--policy-group POLICY_GROUP",
         description: "Policy group name to use (--policy-name must also be given).",
+        proc: Proc.new { |u| Chef::Config[:knife][:policy_group] = u },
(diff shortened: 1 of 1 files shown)


lamont-granquist:
This suffers from the same problem that I need someone to explain to me why we have to `Proc` this when my understanding of what we did ages ago is that the cli + config options are correctly merged and we should never have to do this.  So blocked on the same comments as #8844.

dheerajd-msys:
This change is no longer required considering the `policy_name` and `policy_group` gets written in `/etc/chef/first_boot.json` on the node.

So there is no need to write `policy_name` and `policy_group` into `client.rb`.
Tested thoroughly providing the `policy_name` and `policy_group` from command line as well.
In this case also these values are written into `/etc/chef/first_boot.json`

With the introduction of `policy_name` and `policy_group`, `run_list` can be neglected since Policyfile options and `--run-list` are exclusive.

## Output of /etc/chef/first_boot.json
1. When options provided from CLI
```
[azure@dh-centos71 ~]$ cat /etc/chef/first-boot.json
{"policy_name":"chef_client_test","policy_group":"dh-policy-grp"}
```
2. When options provided from `config.rb`
```
[azure@dh-centos71 ~]$ cat /etc/chef/first-boot.json
{"policy_name":"chef_client_test","policy_group":"dh-policy-grp"}
```

PR #8941, opened 2019-10-01
🚧 WIP 🛑 Blocked: refactor Linux hab plan, add testing! (based on lessons from Windows plan)

robbkidd:
From lessons learned on #8900 

Lots of things changing in here. Let's take them in a possibly weird order.

### do_prepare()

Move all the settng of build environment ..uh.. settings into the prepare function. Notable things:

- `GEM_HOME` is set to a vendor directory under `pkg_prefix` so that all gem installs (regardless of `bundle install` or `gem install`) will be installed _there_. Expect to see all gems shipped with chef-client under this directory.
- Multiple `bundle config --local` commands. These are executed in a subshell after changing directory to the `CACHE_PATH` so that the bundler configuration will be written out to a `.bundle/config` file there. bundle commands will be run from this location later, so these settings become effective for the duration of the build and obviate the need for a `_bundle` helper function.

### pkg_bin_dirs=()

Add the `vendor/bin` directory (remember, `vendor/` is where gems get installed, so the binstubs created by the installation will appear here in vendor/bin/) to the `PATH` of this package's runtime environment because there are commands in there people sometimes want to run. They won't be fast and they might end up with conflicts with gems installed by the user or by downstream pacakges  because they haven't been appbundled.

### do_setup_environment()

These are environment variables we'll want set to these defaults at
runtime.

- `GEM_PATH` gets this package's gem installation directory pushed onto it making the gems available at runtime regardless of whether bundle exe is used.
- `APPBUNDLER_ALLOW_RVM` is a flag looked for by the binstubs generated by appbundler. Setting this to "true" has nothing to do with RVM and everything to do with asking the appbundled binstub to not wipe out the carefully constructed `GEM_PATH` from this package and any possible others that "publish" to the runtime environment that gems are available within.
- `SSL_CERT_FILE` - trust the CA cert package from core, by default, but downstream packages can override this if they like with something
  custom.

### do_build()

Life is simpler in the build function now. Build environment is all setup, so here we:

- bundle install to retrieve dependencies based on the Gemfile+.lock
- run the rake install for the project which builds and installs the gems whose source are in this repo (chef, chef-bin, chef-config)
- for any gem installed via a git reference, we change to the source directory and rake install it, too, to install it like any other gem

This results in a clean collection of all the required gems installed like normal gems under "vendor/gems." With that directory on the `GEM_PATH`, we don't need to care about whether the gems were installed by path reference or via git.

### do_install()

OK. I fibbed about not needing to tell bundler anything else. For install, we're going to generate the appbundled binstubs that lockdown the versions of gems required by particular commands. appbundler uses the project's Gemfile+lock to determine those version, so while we're in the "install" directory (`pkg_prefix`), we tell bundler that the Gemfile to work with is over in `CACHE_PATH`. Then we iterate over the names of gems we want appbundled binstubs for in this package. Those will land in `pkg_prefix/bin/`.

Note: there's nothing to copy here during install because all the gem install actions were targeting `pkg_prefix/vendor` so what we would copy is here already.

### do_after()

Stuff to make the package smaller.

- We don't need the cached .gem artifacts.
- We don't need the cache of git repos retrieved by bundler (those gems are installed at this point like any other gem.)
- We don't need to ship the gem API documentation. (I imagine we could change our minds about this later.)
- We don't need the spec tests for all the gems ... except for our own chef gem which, for Reasons, we keep the spec suite around for post-build functional testing.

### pkg_deps()

Ruby comes with a version of bundler that works at the moment. We don't have to risk conflicts between the version of Ruby this package depends on and the version of Ruby core/bundler was built against.


robbkidd:
@afiune These are the changes to the chef-client plan we walked through (last week?). Probably need more review, but here they are!

robbkidd:
Opened https://github.com/habitat-sh/habitat/issues/7135 to capture the difficulty in testing the chef-infra-client package with `hab pkg exec`.

robbkidd:
Closing in favor of #10102! The parts of this that are worth saving were reapplied there to a more up-to-date mainline

PR #8960, opened 2019-10-03
add support for reinstall

ritzk:
add support for reinstall to yum and dnf package resources

## Description
Will allow admins to reinstall a corrupted pkg

## Related Issue
https://stackoverflow.com/questions/28703217/how-reinstall-a-package-using-dpkg-package-chef-resource

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -65,7 +65,7 @@ class Chef
         end
 
         def define_resource_requirements
-          requirements.assert(:install, :upgrade, :remove, :purge) do |a|
+          requirements.assert(:install, :reinstall, :upgrade, :remove, :purge) do |a|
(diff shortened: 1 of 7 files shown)


lamont-granquist:
there's no `action :reinstall` in the superclass so this isn't working, and the func tests are failing on centos + fedora.

vsingh-msys:
I think we do not support `:rename`, `reinstall` action because it never maintain idempotency

lamont-granquist:
Yeah that's going to be the next bit which is that `:reinstall` as a action is going to be impossible to use idempotently, so its only going to be useful as a trigger.

To accept it there'd need to be a much better and compelling user story.  "As a chef user, I want to .... " kinds of stuff describing what doesn't currently work and can't be made to work, and how the feature would be used.  It needs a more formal write up than just that SO post.

PR #8961, opened 2019-10-07
Avoid a PATH environment variable update before a windows package install

jeremyhage:
Avoid a PATH environment variable update before a windows package install

## Description
During a windows package install we need the PATH environment variable to be preserved. By default, shell_out updates the PATH by putting the Chef bin directory at the first place. It could break an install if a system command with the same name as a binary file embedded in Chef is used (like find.exe for example). 

Setting default_env to false into shell_out options in windows_package resource fix this issue. 

By default shell_out also updates the env vars LC_ALL, LANGUAGE and LANG which are not windows related variables.  

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/windows/exe.rb
+++ b/lib/chef/provider/package/windows/exe.rb
@@ -64,7 +64,7 @@ class Chef
                 unattended_flags,
                 expand_options(new_resource.options),
                 "& exit %%%%ERRORLEVEL%%%%",
-              ].join(" "), timeout: new_resource.timeout, returns: new_resource.returns, sensitive: new_resource.sensitive
+              ].join(" "), default_env: false, timeout: new_resource.timeout, returns: new_resource.returns, sensitive: new_resource.sensitive
             )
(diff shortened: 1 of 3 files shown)


lamont-granquist:
look correct, we need to not do the internal env var mangling here.

lamont-granquist:
./spec/unit/provider/package/windows/exe_spec.rb has a related failure

jeremyhage:
> ./spec/unit/provider/package/windows/exe_spec.rb has a related failure

I updated the test and it passes now. There are 2 more errors on windows functional tests but it does not seem to be due to my commits. 

`rspec ./spec/functional/util/powershell/cmdlet_spec.rb:88 # Chef::Util::Powershell::Cmdlet when returning json returns json format data`
`rspec ./spec/functional/util/powershell/cmdlet_spec.rb:107 # Chef::Util::Powershell::Cmdlet when constructor is given invalid arguments throws an exception`

Issue #8958, opened 2019-10-03
knife supermarket list taking too long


vsingh-msys:
## Description

`knife supermarket list` fetching the list from default supermarket `https://supermarket.chef.io` URL. Seems the default items count set to 10 and no way we can change it.

so it total items ~3,860 so it sending 3860/10 = 386 requests to get the data. After fetching all the data it's showing to the user which take around **~15mins**

Eg.
```
knife supermarket list
start_time: 2019-10-03 16:14:33 +0530
------
end_time: 2019-10-03 16:31:49 +0530
Total time: 17.2687 mins
```

## Client Output

```
TRACE: Initiating GET to https://supermarket.chef.io/api/v1/cookbooks?items=10&start=0
TRACE: ---- HTTP Request Header Data: ----
TRACE: Accept: application/json
TRACE: Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
TRACE: X-REMOTE-REQUEST-ID: 7c23aa39-f5c9-4d8d-8a78-54fe92aa1c4f
TRACE: ---- End HTTP Request Header Data ----
TRACE: ---- HTTP Status and Header Data: ----
TRACE: HTTP 1.1 200 OK
TRACE: cache-control: max-age=0, private, must-revalidate
TRACE: content-encoding: gzip
TRACE: content-type: application/json; charset=utf-8
TRACE: date: Thu, 03 Oct 2019 10:28:40 GMT
TRACE: etag: W/"b63af66a9a9cfaff2782f977db888127"
TRACE: server: nginx
TRACE: x-content-type-options: nosniff
TRACE: x-frame-options: SAMEORIGIN
TRACE: x-request-id: 365623e7-97d3-4a4c-8317-9ffa914d346f
TRACE: x-runtime: 0.041249
TRACE: x-xss-protection: 1; mode=block
TRACE: content-length: 498
TRACE: connection: Close
TRACE: ---- End HTTP Status/Header Data ----
TRACE: Chef::HTTP calling Chef::HTTP::ValidateContentLength#handle_response
TRACE: Content-Length validated correctly.
TRACE: Chef::HTTP calling Chef::HTTP::RemoteRequestID#handle_response
TRACE: Chef::HTTP calling Chef::HTTP::Decompressor#handle_response
TRACE: Decompressing gzip response
TRACE: Chef::HTTP calling Chef::HTTP::CookieManager#handle_response
[...]

PR #8971, opened 2019-10-09
Enhance knife supermarket list & search

vsingh-msys:
## Description

- Increase items count 10 to 1000.  
- Remove #sort from cookbook collection because data is already sorted.
- Add `--sort-by` & `--owned-by` options
- Update the `new_start` count if total records are less so that it can prevent an infinite loop.

**Note:** Removing the limit from the cookbooks supermarket API is not possible because it always returns with default limit 10.

eg:  `https://supermarket.chef.io/api/v1/cookbooks` 
```
{"start":0,"total":3863,"items":[ 10 records...]}
```
We can bypass it by providing items limit something  `999999` or `9999999` but it's not feasible solution. 

## Related Issue
Fixes https://github.com/chef/chef/issues/8958

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/supermarket_list.rb
+++ b/lib/chef/knife/supermarket_list.rb
@@ -37,25 +37,40 @@ class Chef
         default: "https://supermarket.chef.io",
         proc: Proc.new { |supermarket| Chef::Config[:knife][:supermarket_site] = supermarket }
 
+      option :sort_by,
+        long: "--sort-by SORT",
+        description: "Use to sort the records",
+        in: %w{recently_updated recently_added most_downloaded most_followed}
+
+      option :owned_by,
+        short: "-o USER",
+        long: "--owned-by USER",
+        description: "Show cookbooks that are owned by the USER"
+
       def run
+        url_params = []
+        url_params << "&order=#{config[:sort_by]}" if config[:sort_by]
+        url_params << "&user=#{config[:owned_by]}" if config[:owned_by]
+
+        query = url_params.join unless url_params.empty?
+        cookbooks_list = get_cookbook_list(query)
         if config[:with_uri]
-          cookbooks = {}
-          get_cookbook_list.each { |k, v| cookbooks[k] = v["cookbook"] }
-          ui.output(format_for_display(cookbooks))
+          ui.output(format_for_display(cookbooks_list))
         else
-          ui.msg(ui.list(get_cookbook_list.keys.sort, :columns_down))
+          ui.msg(ui.list(cookbooks_list.keys, :columns_down))
         end
       end
 
-      def get_cookbook_list(items = 10, start = 0, cookbook_collection = {})
+      def get_cookbook_list(query, items = 1000, start = 0, cookbook_collection = {})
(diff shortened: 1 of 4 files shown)


lib/chef/knife/supermarket_list.rb
```diff
@@ -37,25 +37,40 @@ class SupermarketList < Knife
...
         else
-          ui.msg(ui.list(get_cookbook_list.keys.sort, :columns_down))
+          ui.msg(ui.list(cookbooks_list.keys, :columns_down))
         end
       end
 
-      def get_cookbook_list(items = 10, start = 0, cookbook_collection = {})
+      def get_cookbook_list(query, items = 1000, start = 0, cookbook_collection = {})
         cookbooks_url = "#{config[:supermarket_site]}/api/v1/cookbooks?items=#{items}&start=#{start}"
+        cookbooks_url << query if query
```
robbkidd: Could we reduce the number of things changing in this by moving the `sort_by` and `owned_by` parameter additions to here?

```
cookbooks_url = "#{config[:supermarket_site]}/api/v1/cookbooks?items=#{items}&start=#{start}"
cookbooks_url << "&order=#{config[:sort_by]}" if config[:sort_by]
cookbooks_url << "&user=#{config[:owned_by]}" if config[:owned_by]
```

I think that would leave the `run` method untouched and the `get_cookbook_list` method solely responsible for URL string construction.

lib/chef/knife/supermarket_list.rb
```diff
@@ -37,25 +37,40 @@ class SupermarketList < Knife
         default: "https://supermarket.chef.io",
         proc: Proc.new { |supermarket| Chef::Config[:knife][:supermarket_site] = supermarket }
 
+      option :sort_by,
+        long: "--sort-by SORT",
+        description: "Use to sort the records",
+        in: %w{recently_updated recently_added most_downloaded most_followed}
```
robbkidd: ✨ Yes! Great guarding of [the actual options available in the API endpoint](https://github.com/chef/supermarket/blob/cc49068bc4d2affff11b77939b61e0d2bd6f9a29/src/supermarket/app/models/cookbook.rb#L30-L33). ✅

lib/chef/knife/supermarket_list.rb
vsingh-msys: Ohh Yup, Thanks

lib/chef/knife/supermarket_list.rb
```diff
@@ -37,23 +37,34 @@ class SupermarketList < Knife
...
-          ui.output(format_for_display(cookbooks))
+          ui.output(format_for_display(get_cookbook_list))
         else
-          ui.msg(ui.list(get_cookbook_list.keys.sort, :columns_down))
+          ui.msg(ui.list(get_cookbook_list.keys, :columns_down))
         end
       end
 
-      def get_cookbook_list(items = 10, start = 0, cookbook_collection = {})
+      def get_cookbook_list(items = 1000, start = 0, cookbook_collection = {})
```
tas50: Let's just set this to 9999999 and avoid the whole thing. You probably want to add a comment that we're using 9999999 to avoid pagination entirely.

Issue #8965, opened 2019-10-07
"undefined method `+' for false:FalseClass" when using package resources on CentOS 8 


javier-lopez:
# Description
`yum_package` fail to install packages on CentOS 8.

# Chef Version
15.3.14

# Platform Version
CentOS Linux release 8.0.1905 (Core)

# Replication Case
Create a recipe that installs a package, for instance:

```
yum_package 'yum-utils' do
        action :install
end
```

# Client Output

```
chef-apply yum.rb
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * yum_package[yum-utils] action install

    ================================================================================
    Error executing action `install` on resource 'yum_package[yum-utils]'
    ================================================================================

    NoMethodError
    -------------
    undefined method `+' for false:FalseClass

    Resource Declaration:
    ---------------------
    # In yum.rb

      1: yum_package 'yum-utils' do
      2:        action :install
      3: end

    Compiled Resource:
    ------------------
    # Declared in yum.rb:1:in `run_chef_recipe'

    yum_package("yum-utils") do
      package_name "yum-utils"
      action [:install]
      default_guard_interpreter :default
      declared_type :yum_package
      cookbook_name "(chef-apply cookbook)"
      recipe_name "(chef-apply recipe)"
    end

    System Info:
    ------------
    chef_version=15.3.14
    platform=centos
    platform_version=8.0.1905
    ruby=ruby 2.6.4p104 (2019-08-28 revision 67798) [x86_64-linux]
    program_name=/usr/bin/chef-apply
    executable=/opt/chef/bin/chef-apply

[2019-10-07T13:14:02-07:00] FATAL: Stacktrace dumped to /home/vagrant/.chef/cache/chef-stacktrace.out
[2019-10-07T13:14:02-07:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2019-10-07T13:14:02-07:00] FATAL: NoMethodError: yum_package[yum-utils] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: NoMethodError: undefined method `+' for false:FalseClass
```

There is a related issue: https://github.com/chef/chef/issues/7988

```
[vagrant@centos-80-x64 chef]$ python --version
-bash: python: command not found
[vagrant@centos-80-x64 chef]$ python2 --version
-bash: python2: command not found
[vagrant@centos-80-x64 chef]$ python3 --version
Python 3.6.8
[...]

PR #8972, opened 2019-10-09
Fix yum & dnf shellout if exit with nonzero status

vsingh-msys:
## Description

If shell out of python yum & dnf helper is exited with nonzero status then raise an error with stderr message
Now it showing 

```
------
FATAL: RuntimeError: yum_package[yum-utils] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: RuntimeError: Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'yum'
```
Instead of `"undefined method `+' for false:FalseClass"`
## Related Issue

Fixes https://github.com/chef/chef/issues/8965

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/yum/python_helper.rb
+++ b/lib/chef/provider/package/yum/python_helper.rb
@@ -45,6 +45,7 @@ class Chef
                 shell_out("#{f} -c 'import yum'").exitstatus == 0
               end
               raise Chef::Exceptions::Package, "cannot find yum libraries, you may need to use dnf_package" unless cmd
+
               "#{cmd} #{YUM_HELPER}"
             end
           end
--- a/lib/chef/provider/package/yum/python_helper.rb
+++ b/lib/chef/provider/package/yum/python_helper.rb
@@ -40,9 +40,16 @@ class Chef
           YUM_HELPER = ::File.expand_path(::File.join(::File.dirname(__FILE__), "yum_helper.py")).freeze
 
           def yum_command
-            @yum_command ||= which("platform-python", "python", "python2", "python2.7", extra_path: "/usr/libexec") do |f|
-              shell_out("#{f} -c 'import yum'").exitstatus == 0
-            end + " #{YUM_HELPER}"
+            @yum_command ||= begin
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/yum/python_helper.rb
```diff
@@ -40,9 +40,16 @@ class PythonHelper
...
+            @yum_command ||= begin
+              cmd = which("platform-python", "python", "python2", "python2.7", extra_path: "/usr/libexec") do |f|
+                res = shell_out("#{f} -c 'import yum'")
+                raise res.stderr if res.exitstatus != 0
+
+                true
+              end
+
+              "#{cmd} #{YUM_HELPER}"
+            end
```
lamont-granquist: So I think what I'd like better here would look more like:

```ruby
@yum_command ||= begin
    cmd = which("platform-python", "python", "python2", "python2.7", extra_path: "/usr/libexec") do |f|
              shell_out("#{f} -c 'import yum'").exitstatus == 0
            end
    raise Chef::Exceptions::Package, "cannot find yum libraries, you may need to use dnf_package" unless cmd
    "#{cmd} #{YUM_HELPER}"
  end
```

The code here in this PR has a bug where it'll only try to the first python it finds and if that doesn't have the yum/dnf libraries it'll throw.  It needs to preserve the behavior of searching through all the python executables that it finds to find the one that has the yum/dnf libs.

If you want to go the extra mile you could Chef::Log.trace log the stderr of all the shell_out's that are executed (I don't think users will generally want that output)

lamont-granquist:
👍 in general, this is the direction we should take to close out that issue.

lib/chef/provider/package/yum/python_helper.rb
vsingh-msys: @lamont-granquist Updated the required changes. Thankyou!

vsingh-msys:
Updated error message in dnf_package & yum_package vice versa. let me know if further if any feedback.

PR #8981, opened 2019-10-10
windows_service: don't update the service if the run_as_user ca…

### LINKED ISSUE #8715: Incorrect casing on run_as_user cases windows_service to restart every run
## Description

When using the `windows_service` resource, if the `run_as_user` property differs only by case, chef will attempt to apply the change and restart the service so that it takes effect. Unfortunately, Windows does not store the casing change as the value is case-insensitive. This results in Chef attempting to make the change every time it runs (thus restarting the service).

In our case, this meant the service was only online for 15 minutes at a time. To resolve this, we had to change the credentials to another user account, then change them back with the casing Chef expected.

## Chef Version

Chef: 14.2.0

## Platform Version

Windows Server 2016

## Replication Case

1. Ensure a Windows Service exists which runs as a user account. Ensure the credentials are entered in all lowercase.
2. Create a Chef recipe which includes a `windows_service` resource. Specify the same credentials in the resource definition but use all upper case.
3. Run the cookbook multiple times, the service will be restarted on each run.

## Client Output

Client output cannot be included as it contains a password but was identical on each run. The windows_service resource was configured with a new username and password and then restarted.

## Stacktrace

No error is thrown.

jasonwbarnett:
## Description

Fixes #8715 

## Types of changes

## Checklist:

### DIFF
--- a/spec/unit/provider/service/windows_spec.rb
+++ b/spec/unit/provider/service/windows_spec.rb
@@ -19,8 +19,16 @@
 
 require "spec_helper"
 
-describe Chef::Provider::Service::Windows, "load_current_resource", :windows_only do
+describe Chef::Provider::Service::Windows, "load_current_resource" do
   include_context "Win32"
+
+  before(:all) do
+    Chef::ReservedNames::Win32::Security = Class.new unless windows?
+  end
+
+  after(:all) do
+    Chef::ReservedNames::Win32.send(:remove_const, :Security) unless windows?
+  end
 
   let(:logger) { double("Mixlib::Log::Child").as_null_object }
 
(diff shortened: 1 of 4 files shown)


jasonwbarnett:
@tas50 ready for 👀

spec/unit/provider/service/windows_spec.rb
```diff
@@ -19,9 +19,17 @@
 
 require "spec_helper"
 
-describe Chef::Provider::Service::Windows, "load_current_resource", :windows_only do
+describe Chef::Provider::Service::Windows, "load_current_resource" do
   include_context "Win32"
 
+  before(:all) do
+    Chef::ReservedNames::Win32::Security = Class.new unless windows?
```
jasonwbarnett: This goodness allows running this in *nix

spec/unit/provider/service/windows_spec.rb
```diff
@@ -19,9 +19,17 @@
...
+  end
+
+  after(:all) do
+    Chef::ReservedNames::Win32.send(:remove_const, :Security) unless windows?
```
jasonwbarnett: This goodness allows running this in *nix

Issue #4213, opened 2015-12-01
knife download environments incomplete


sdarwin:
Chef: 12.5.1

knife download environments, creates files formatted as:

```
{
  "name": "staging"
}
```

alternatively, a command such as "knife environment show staging --format=json > environments/staging.json" generates:

```
{
  "name": "staging",
  "description": "",
  "cookbook_versions": {

  },
  "json_class": "Chef::Environment",
  "chef_type": "environment",
  "default_attributes": {

  },
  "override_attributes": {

  }
}
```

The latter format is more complete, and in fact test-kitchen + chef-solo requires those additional attributes in an environment, it crashes otherwise.

"knife download environments" should probably download the same data as "knife environment show", right?

thommay:
Right, we should totally verify which bits of this kitchen/solo/zero actually need and make `knife download environment` do the right thing here.

lamont-granquist:
If the problem here is that ChefFS is suppressing empty hashes on download and is not actually reflecting the state of the 'document' on the server, that should definitely get fixed there.

If its accurately reflecting the server, it might be a bit muddier as to what to do about this...

smurawski:
This should be part of https://github.com/chef/chef/issues/3368

lamont-granquist:
I'm not actually positive its part of that epic....   it might just be a core bug in ChefFS...

PR #8986, opened 2019-10-14
ChefFS & knife environment matching the output

vsingh-msys:
## Description

Added default params in `knife download /environments` so that it's matched with `knife environment show  --format=json`

## Related Issue

Fixes https://github.com/chef/chef/issues/4213

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb
+++ b/lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb
@@ -101,7 +101,7 @@ class Chef
 
           def read
             # Minimize the value (get rid of defaults) so the results don't look terrible
-            Chef::JSONCompat.to_json_pretty(minimize_value(_read_json))
+            Chef::JSONCompat.to_json_pretty(normalize_value(_read_json))
           end
 
           def _read_json
@@ -122,7 +122,11 @@ class Chef
           end
 
           def minimize_value(value)
-            data_handler.minimize(data_handler.normalize(value, self), self)
+            data_handler.minimize(normalize_value(value), self)
+          end
+
+          def normalize_value(value)
+            data_handler.normalize(value, self)
           end
 
           def compare_to(other)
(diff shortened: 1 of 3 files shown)


lamont-granquist:
Can you validate that if you create a minimal role/environment, that when you upload that to the server that `knife diff` will not produce any diffs?

In other words, does knife diff go through a different codepath where it is still being minimized?  Or does it in some other way force minimization of the json data?

Vasu1105:
As per my observations if you have other information added for environment knife download that information

Vasu1105:
As per my observation 

If we have other information set for environment knife download . / knife download /environments does download that information.

Data after download of the environments 
for default.json
```
{
  "name": "_default",
  "description": "The default Chef environment"
}
```
for development.json where I have added some more data about environment.
```
{
  "name": "development",
  "description": "Used for developement environment",
  "cookbook_versions": {
    "starter": "1.0.0"
  }
}
```
What I feel is if we are setting any data in knife download if we are having any data loss that will be bug then as we use same data to upload also. So it may be this behavior is correct to bring minimal data at the download time and to not create any additional information which is not required at the time of download. As it may increase the time for download. These are just my thoughts.

vsingh-msys:
@lamont-granquist As per the knife diff concern, the normalized environment does not produce any diff content because in `#campare_to` it minizine the both local and server content before comparing.

https://github.com/chef/chef/blob/b7025a006710bbbf9bc9c05b21a083ca7e53a65f/lib/chef/chef_fs/file_system/chef_server/rest_list_entry.rb#L145-L155


@Vasu1105  That's true, there is no loss of the data if we upload environment data keys with the respective non-empty value.

But if we upload or create the environment with:
```
{
  "name": "dev",
  "description": "",
  "cookbook_versions": {

  },
  "default_attributes": {

  },
  "override_attributes": {

  },
  "json_class": "Chef::Environment",
  "chef_type": "environment"
}

```
when downloading environments it always returns
```
{
 "name": "dev"
}
```

vsingh-msys:
@lamont-granquist, For now, I have tested/verified these changes only for the environments. 

For roles, it's testing is still pending.

vsingh-msys:
Roles also working as expected with the latest changes. This PR is good to review again Thanks!

lamont-granquist:
Marked this as breaking change since it looks breaking, it is correct but it should only go into Chef-16.0

Issue #5148, opened 2016-07-29
Using knife.rb setting knife[:script_path] as documented throws exception


wwilfingertt:
## Description

In my knife.rb, set `knife[:script_path]` to a string that is a path to Ruby scripts. Try to `knife exec` one of the scripts, exception is thrown.
## Chef Version

```
% knife --version
Chef: 12.4.1
```
## Platform Version

CentOS 6
## Replication Case

https://docs.chef.io/config_rb_knife_optional_settings.html

![screenshot_072916_111537_am](https://cloud.githubusercontent.com/assets/5647708/17255171/c7fb089c-557d-11e6-9242-12ff0a0f3860.jpg)

Add to my knife.rb as documented

```
knife[:script_path] = "/home/myself/work/scripts"
```

Throws a undefined method `each' for string exception.

This works

```
knife[:script_path] = ["/home/myself/work/scripts"]
```

This also works

```
script_path "/home/myself/work/scripts"
```

I believe it comes down to [this line here](https://github.com/chef/chef/blob/master/lib/chef/knife/exec.rb#L42). If the documentation said `knife[:script_path]` should be an array of paths, that would be fine.
## Client Output

```
% knife exec -VV a_script
INFO: Using configuration from /home/myself/.chef/knife.rb
DEBUG: Searching script_path: "/home/myself/work/scripts/home/myself/.chef/scripts/home/myself/.chef/scripts"
/opt/chef/embedded/apps/chef/lib/chef/knife/exec.rb:73:in `find_script': undefined method `each' for #<String:0x00000002013608> (NoMethodError)
        from /opt/chef/embedded/apps/chef/lib/chef/knife/exec.rb:55:in `block in run'
        from /opt/chef/embedded/apps/chef/lib/chef/knife/exec.rb:54:in `each'
        from /opt/chef/embedded/apps/chef/lib/chef/knife/exec.rb:54:in `run'
        from /opt/chef/embedded/apps/chef/lib/chef/knife.rb:425:in `block in run_with_pretty_exceptions'
        from /opt/chef/embedded/apps/chef/lib/chef/local_mode.rb:39:in `with_server_connectivity'
        from /opt/chef/embedded/apps/chef/lib/chef/knife.rb:424:in `run_with_pretty_exceptions'
        from /opt/chef/embedded/apps/chef/lib/chef/knife.rb:215:in `run'
        from /opt/chef/embedded/apps/chef/lib/chef/application/knife.rb:142:in `run'
[...]

PR #8987, opened 2019-10-14
Fix undefined method `each' for String

vsingh-msys:
- Fix `undefined method 'each' for #<String:> (NoMethodError) ` while providing script path as follow

```
knife[:script_path] = "/home/myself/work/scripts"
```

 - Better UI when no args provided for knife exec.

## Related Issue

Fixes https://github.com/chef/chef/issues/5148
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/exec.rb
+++ b/lib/chef/knife/exec.rb
@@ -40,7 +40,7 @@ class Chef::Knife::Exec < Chef::Knife
   end
 
   def run
-    config[:script_path] ||= Array(Chef::Config[:script_path])
+    config[:script_path] = Array(config[:script_path] || Chef::Config[:script_path])
 
     # Default script paths are chef-repo/.chef/scripts and ~/.chef/scripts
     config[:script_path] << File.join(Chef::Knife.chef_config_dir, "scripts") if Chef::Knife.chef_config_dir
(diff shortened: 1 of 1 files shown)


lib/chef/knife/exec.rb
```diff
@@ -40,7 +40,7 @@ class Chef::Knife::Exec < Chef::Knife
   end
 
   def run
-    config[:script_path] ||= Array(Chef::Config[:script_path])
+    config[:script_path] = Array(config[:script_path] || Chef::Config[:script_path])
```
lamont-granquist: due to the merging of configs, i think you can just use `config[:script_path]` here, `Chef::Config[:script_path]` should already be in there.

vsingh-msys: As I can verify for now we required both
1. when `knife[:script_path]` is set
```
knife[:script_path] =     "/chef/scripts-path"

=> puts config[:script_path]
/chef/scripts-path

=> puts Chef::Config[:script_path]
[]
```
2. when `script_path` is provided
```
script_path    "/chef/scripts-path"

=> puts config[:script_path]
nil
=> puts Chef::Config[:script_path]
["/chef/scripts-path"]
```
let me know if I am missing something? Thanks!

lamont-granquist: Ah right, config[:script_path] = Chef::Config[:knife][:script_path] and not Chef::Config[:script_path] and the latter is not in the knife namespace because this option is likely very old.

We should probably audit everything in knife for Chef::Config references that aren't in the knife namespace.

PR #8993, opened 2019-10-14
Build as chef-client in habitat again

tas50:
This let's folks continue to consume the old chef-client packages


tas50:
https://github.com/chef/chef/pull/8994 Instead

PR #9002, opened 2019-10-17
[knife supermarket] set tarballs to uid/gid 0 to avoid large uid/gid errors on consumer machines

robbkidd:
## Description

The tar commands default to storing the user and group IDs--and the user/group names associated with them--of the directories and files being archived. This change updates the BSD and GNU forms for the tar command to set the owner and group to 0 and to not store the user/group name mapping in the resultant archive.

There is also has a benefit of avoiding bugs in tar readers that do not behave well when large uid/gid values are stored in tar's extended metadata headers. While this won't fix cookbooks that have already been published, it will minimize the bug's appearance in clients with older dependency resolvers as fewer latest cookbook versions will include large IDs.

## Related Issue

To reiterate, this PR won't fix the issue for cookbooks already published. Any code that expands a cookbook—clients, Supermarkets, etc—must still handle tarballs with large IDs. These issues exhibit the downstream pain of the large uid/gid values being included in the cookbook tarballs.
 
* https://github.com/chef/supermarket/issues/1806
* https://github.com/chef/chef-dk/issues/1549
* https://github.com/chef/chef-dk/issues/2018
* https://github.com/chef/effortless/issues/54

## Types of changes

- ~New feature (non-breaking change which adds functionality)~
- ~Breaking change (fix or feature that would cause existing functionality to change)~
- ~Chore (non-breaking change that does not add functionality or fix an issue)~

## Checklist:

### DIFF
--- a/lib/chef/knife/supermarket_share.rb
+++ b/lib/chef/knife/supermarket_share.rb
@@ -151,11 +151,11 @@ class Chef
 
       def tar_cmd
         unless @tar_cmd
-          @tar_cmd = "tar"
+          @tar_cmd = "tar --numeric-owner --uid 0 --gid 0"
           begin
             # Unix and Mac only - prefer gnutar
             if shell_out("which gnutar").exitstatus.equal?(0)
-              @tar_cmd = "gnutar"
+              @tar_cmd = "gnutar --numeric-owner --owner 0 --group 0"
             end
           rescue Errno::ENOENT
           end


lib/chef/knife/supermarket_share.rb
```diff
@@ -151,11 +151,11 @@ def do_upload(cookbook_filename, cookbook_category, user_id, user_secret_filenam
 
       def tar_cmd
         unless @tar_cmd
-          @tar_cmd = "tar"
+          @tar_cmd = "tar --numeric-owner --uid 0 --gid 0"
```
lamont-granquist: that'll break non-freebsd non-gnutar posix tar implementations (solaris definitely and probably aix)

lamont-granquist: i'd suggest you just be explicit about case statements for linux, mac, bsd and other, but that is difficult because knife doesn't run ohai.  knife should probably at least run --minimal-ohai to get platform stuff working, but that might be a bit out of scope for you, but the lack of that makes this all hard to do right.

lamont-granquist: also i think the 'E' flag for extended metadata should be used on old posix tar to address some of your comments.

lamont-granquist:
Changed the title, since the actual breakage is much more important than the anonymization.  I was about to close this on the principle that the anonymization wasn't sufficient to warrant doing the work to fix it since it seems to be not as trivial at first glance.

lamont-granquist:
yeah this is probably better left unfixed for now, its harder than it looks at first glance

Issue #3890, opened 2015-09-07
file_cache_path setting


CMeza99:
When using a setting such as: 

``` ruby
file_cache_path '~/.chef/cache'
```

chef-client and chef-solo return:

```
[2015-09-07T14:17:06-07:00] INFO: Forking chef instance to converge...
[2015-09-07T14:17:06-07:00] ERROR: No such file or directory @ rb_sysopen - ~/.chef/cache/chef-client-running.pid
[2015-09-07T14:17:06-07:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

It does work when the full path is given. But this is unexpected as settings such as cookbook_path and data_bag_path will accept paths with ~.

chef 12.4.1

lamont-granquist:
Seems related to another one we had recently about ".." not working in cookbook_path.

We should fix all of these and make sure they all work with ".." and "~".

lamont-granquist:
Seems related to #2809 and #3079

vsingh-msys:
Fixed #3890 via #9042

Issue #5842, opened 2017-02-24
ChefConfig should expand relative paths


coderanger:
For stuff like `cookbook_path` and `chef_repo_path` it would be nice if they expanded relative paths to the current directory. This might be confusing in some cases though so I'm not sure of the best semantics. If the options in a file, I would expect it to be relative to the file, but if passed on the command line (via `--config-option chef_repo_path=.` or `--config-option cookbook_path=cookbooks/`) I would expect it to be relative to the cwd. Thoughts?

Issue #2875, opened 2015-02-07
Default file_cache_path with local-mode may break things and do not honor documentation


gionn:
Hi,

I've just found a misbehaving ark resource due to the default relative path set as file_cache_path:

```
./local-mode-cache/cache
```

Also, the documentation wrongly states:

```
When running the chef-client using the --local-mode option, ~/.chef/local-mode-cache is the default value for cache_path.
```

that looks a good path, anyway.

I was running with user root on ~:

```
chef-client -z -c client.rb
```

A workaround is to configure _file_cache_path_ in _client.rb_, e.g.:

```
data_bag_path "/tmp/chef/data_bags"
environment_path "/tmp/chef/environments"
cookbook_path "/tmp/chef/cookbooks"
file_cache_path "/tmp/chef/cache"
ssl_verify_mode :verify_peer
```

gionn:
I made a quick dig into it, and chef relies on File.dirname passing the configuration file as passed on the command-line, so I am getting a relative cache path since I passed a relative path for the configuration file.

It may be a good idea to always return an absolute path?

lamont-granquist:
think this is a dup of another issue as well, but offhand i couldn't find it.

vsingh-msys:
Fixed #2875 via #9042

PR #9006, opened 2019-10-18
[Do Not Merge] Expand paths provided in file_cache_path setting to allow for relative paths

kapilchouhan99:

## Description
- Minor fixes, now `file_cache_path` will work with ".." and "~".

## Related Issue
Fixes: #3890 

## Types of changes

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -146,6 +146,7 @@ class Chef
       end
       extra_config_options = config.delete(:config_option)
       chef_config.merge!(config)
+      chef_config[:file_cache_path] = File.expand_path(chef_config[:file_cache_path])
       apply_extra_config_options(extra_config_options)
     end
 
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -130,6 +130,7 @@ describe Chef::Application do
 
         before do
           @app.config[:config_file] = config_location
+          @app.config[:file_cache_path] = "/root/.chef/local-mode-cache/cache"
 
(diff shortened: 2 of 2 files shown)


lamont-granquist:
can we push this somewhere earlier where the lockfile config is created?

tas50:
Ideally we just handle this in the initializer for RunLock. A test for that there would also be useful

lamont-granquist:
I'm actually thinking we should go the other way though.

On "input" we should normalize things as much as possible.  I might even argue that the file_cache_path should have a `Chef::Config[:file_cache_path] = File.expand_path(Chef::Config[:file_cache_path])` somewhere early because otherwise if the pwd of the chef-client changes then the file_cache_path won't be consistent.  Consuming relative paths in the config internally seems like it would be a bug.

I'd even argue that on assignment in Chef::Config itself most paths should be normalized by some kind of helper that just called File.expand_path.  That would eliminate most of these bugs, without spraying File.expand_path defensively all over every piece of code that deals with paths.

vsingh-msys:
@lamont-granquist we do agree with your point. here one more issue https://github.com/chef/chef/issues/2875 related to chef-client relative cache path issue.

In order fixes these relative path issues we can cover this https://github.com/chef/chef/issues/5842 in Chef::Config.

kapilchouhan99:
@lamont-granquist @tas50 There is no need to merge this PR, I have fixed issue #3890 in PR: https://github.com/chef/chef/pull/9042
We can close this PR.

lamont-granquist:
yep closing

PR #9039, opened 2019-10-31
Add output for the file provider verification

### LINKED ISSUE #4793: Discussion: Verification command output

lamont-granquist:
closes #4793

### DIFF
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -341,7 +341,10 @@ class Chef
         if tempfile
           new_resource.verify.each do |v|
             unless v.verify(tempfile.path)
-              raise Chef::Exceptions::ValidationFailed.new "Proposed content for #{new_resource.path} failed verification #{new_resource.sensitive ? "[sensitive]" : v}"
+              backupfile = "#{Chef::Config[:file_cache_path]}/failed_validations/#{::File.basename(tempfile.path)}"
+              FileUtils.mkdir_p File.dirname(backupfile)
+              FileUtils.cp tempfile.path, backupfile
+              raise Chef::Exceptions::ValidationFailed.new "Proposed content for #{new_resource.path} failed verification #{new_resource.sensitive ? "[sensitive]" : "#{v}\n#{v.output}"}\nTemporary file moved to #{backupfile}"
             end
           end
         end
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -342,7 +342,7 @@ class Chef
           new_resource.verify.each do |v|
             unless v.verify(tempfile.path)
               backupfile = "#{Chef::Config[:file_cache_path]}/failed_validations/#{::File.basename(tempfile.path)}"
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@vsingh-msys or @Vasu1105 can one of you pick this up and fix the spec failures and add a spec or two for the behavior here?

vsingh-msys:
Alright @lamont-granquist, will take it up and revert you back accordingly. Thanks

lamont-granquist:
ping @vsingh-msys

lib/chef/provider/file.rb
```diff
@@ -341,7 +341,10 @@ def do_validate_content
         if tempfile
           new_resource.verify.each do |v|
             unless v.verify(tempfile.path)
-              raise Chef::Exceptions::ValidationFailed.new "Proposed content for #{new_resource.path} failed verification #{new_resource.sensitive ? "[sensitive]" : v}"
+              backupfile = "#{Chef::Config[:file_cache_path]}/failed_validations/#{::File.basename(tempfile.path)}"
+              FileUtils.mkdir_p File.dirname(backupfile)
```
vsingh-msys: Getting error 
```
 NoMethodError
    -------------
    undefined method `dirname' for Chef::Provider::File:Class
```
Need to replace `File` with `::File`

vsingh-msys:
@lamont-granquist sorry for the delayed response.

Started looking into it, will update as soon as possible. Thanks

vsingh-msys:
Hi @lamont-granquist , Could you please have a look again. I have only fixed the specs and tested the expected behavior seems its working file. Also, let me know if any further queries/concerns.

Thanks

PR #9042, opened 2019-11-01
Fix for ChefConfig should expand relative paths

kapilchouhan99:

## Description
- Expand paths provided in `--config-option` setting to allow for relative paths
- Added test cases

## Related Issue
Fixes: #5842 #3890 

## Types of changes

## Checklist:

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -107,12 +107,20 @@ module ChefConfig
           # Split including whitespace if someone does truly odd like
           # --config-option "foo = bar"
           key, value = option.split(/\s*=\s*/, 2)
+
           # Call to_sym because Chef::Config expects only symbol keys. Also
           # runs a simple parse on the string for some common types.
           memo[key.to_sym] = YAML.safe_load(value)
           memo
         end
-        merge!(extra_parsed_options)
+        set_extra_config_options(extra_parsed_options)
+      end
+    end
+
+    # We use :[]= assignment here to not bypass any coercions that happen via mixlib-config writes_value callbacks
+    def self.set_extra_config_options(extra_parsed_options)
+      extra_parsed_options.each do |key, value|
+        self[key.to_sym] = value
       end
     end
 
@@ -154,6 +162,20 @@ module ChefConfig
     # it up here and get a real method written so that things get dispatched
     # properly.
     configurable(:daemonize).writes_value { |v| v }
+
+    def self.expand_relative_paths(path)
+      unless path.nil?
+        if path.is_a?(String)
+          File.expand_path(path)
+        else
+          Array(path).map { |path| File.expand_path(path) }
+        end
+      end
+    end
+
+    configurable(:cookbook_path).writes_value { |path| expand_relative_paths(path) }
+
+    configurable(:chef_repo_path).writes_value { |path| expand_relative_paths(path) }
 
     # The root where all local chef object data is stored.  cookbooks, data bags,
     # environments are all assumed to be in separate directories under this.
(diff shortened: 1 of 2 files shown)


lamont-granquist:
That doesn't look like the correct way to solve this.  That is making a magic API where anything with a substring of "path" will get matched and get expand_path'd and that is subject to someone embedding "path" in a config option and having it be a boolean instead of a real path and expand_path will throw on that.

This should really be done by coercions on all the path variables so that when they're set they're coerced
via expand_path.  I'm not sure that mixlib-config directly supports coercions, but I think that `writes_value` may be (ab)usable to implement them, there's an example or two already in the chef config.

chef-config/lib/chef-config/config.rb
```diff
@@ -155,6 +162,22 @@ def self.is_valid_url?(uri)
...
 
+    def self.expand_relative_paths(path)
+      unless path.nil?
+        if path.is_a?(String)
+          File.expand_path(path)
+        else
+          path.map do |path|
+            File.expand_path(path)
+          end
+        end
```
lamont-granquist: You should be able to just `Array(path).map { |path| File.expand_path(path) }`

chef-config/lib/chef-config/config.rb
```diff
@@ -107,12 +107,19 @@ def self.apply_extra_config_options(extra_config_options)
...
           # runs a simple parse on the string for some common types.
           memo[key.to_sym] = YAML.safe_load(value)
           memo
         end
-        merge!(extra_parsed_options)
+        set_extra_config_options(extra_parsed_options)
+      end
+    end
+
+    def self.set_extra_config_options(extra_parsed_options)
```
lamont-granquist: You might want to include a comment here:  "# We use :[]= assignment here to not bypass any coercions that happen via mixlib-config writes_value callbacks"

lamont-granquist:
couple nitpicks, but this looks like the right approach

chef-config/lib/chef-config/config.rb
lamont-granquist: that actually replaces the whole block:

```
unless path.nil?
  Array(path).map { |path| File.expand_path(path) }
end
```

if its a string then `Array(path)` wraps an Array around it, but leaves any Array that it finds, and means that there's no `.is_a?` check anymore which means that any scalar type which File.expand_path operates on will work with `expand_relative_paths` without needing to add more types to the check, and is clearer than inverting that and doing an `.is_a?` check on the Array instead.  the use of `is_a?` is also always brittle and creates tight coupling.

lamont-granquist: actually this works:

```ruby
def self.expand_relative_paths(path)
  Array(path).map { |path| File.expand_path(path) }
end
```

because:

```
# pry
[1] pry(main)> Array(nil)
=> []
```

kapilchouhan99: @lamont-granquist If we have string value for any option eg: file_cache_path = "~/test/.chef/cache"
then it will set file_cache_path = ["/root/test/.chef/cache"]
So this might result in the bug and potentially cause other issues as well. 

If there are any more suggestions considering my comment please share.

jobsonmodesto:
Root.natural input idea
Copyright to freedom in say every part but not wise choice

jobsonmodesto:
The gear is getting on though the size in human interface detractor are sound waves measured by graphics bits which read/erase but we modify/here and I write as you understand it ref book documentation language standard first readme

chef-config/lib/chef-config/config.rb
lamont-granquist: I see.  Yeah, that might break downstream consumers external to our code that are buggy against arrays of dirs here (which would still be bugs on their side, but we'd break them hard if we did this).

Okay.

Issue #9038, opened 2019-10-31
chef_version metadata violation raises incorrect error


kimbernator:
## Description
If a run-list includes a cookbook with a chef_version constraint that is not permissive of the version of chef infra client running, the error raised is incorrect due to an error in the code to finalize the run with an end message:

```
Running handlers:
[2019-10-31T13:22:15-05:00] ERROR: Running exception handlers
Running handlers complete
[2019-10-31T13:22:15-05:00] ERROR: Exception handlers complete
Chef Infra Client failed. 0 resources updated in 10 seconds
[2019-10-31T13:22:15-05:00] FATAL: NoMethodError: undefined method `filtered_collection' for nil:NilClass
```

Running `chef-client -l trace` reveals an attempt at re-raising the correct error, which is:

```
[2019-10-31T13:22:15-05:00] TRACE: Re-raising exception: Chef::Exceptions::CookbookChefVersionMismatch - Cookbook '<cookbook_name>' version '0.5.0' depends on chef version [">= 12.14", "< 15"], but the running chef version is 15.4.45
```

In these lines: https://github.com/chef/chef/blob/b499e050480780dcd2f8d30fa3a83374ba879b40/lib/chef/client.rb#L303-L317

The exception is re-raised in a trace log, but line 308's `run_failed` call experiences its own error which is unhandled and results in all code until `ensure` (lines 310-312) not being executed and the `CookbookChefVersionMismatch` not being raised.  

The unhandled error occurs when formatting the run ending message as a result of the `updated_resources` method in chef/resource_reporter.rb: 
https://github.com/chef/chef/blob/b499e050480780dcd2f8d30fa3a83374ba879b40/lib/chef/resource_reporter.rb#L184-L187

Changing line 186 to the following results in a chef-client failure that includes the more useful CookbookChefVersionMismatch error:
```
@updated_resources ||= action_collection&.filtered_collection(max_nesting: 0, up_to_date: false, skipped: false, unprocessed: false) || {}
```
(Changes being the `&` after `action_collection` to return nil instead of throwing a NoMethodError, and `|| {}` at the end to assign `{}` to `@updated_resources`)

I am hesitant to submit a PR with this change since it feels like a cheap hack to resolve one specific error without addressing the fact that `@action_collection` is not being defined early enough for this error to be properly raised.

## Chef Version
15.4.45

## Platform Version
RHEL 6 and Windows Server 2008R2

## Replication Case
[...]

PR #9043, opened 2019-11-01
Return empty hash when action_collection is undefined

kimbernator:
## Description
Resolves an issue with error handling during chef-client runs when the chef_version metadata attribute is violated. See related issue for details.

## Related Issue
https://github.com/chef/chef/issues/9038

## Types of changes

## Checklist:

Signed-off-by: Jeremy Kimber Jeremy.Kimber@cerner.com

### DIFF
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -183,7 +183,11 @@ class Chef
 
     # get only the top level resources and strip out the subcollections
     def updated_resources
-      @updated_resources ||= action_collection&.filtered_collection(max_nesting: 0, up_to_date: false, skipped: false, unprocessed: false) || {}
+      @updated_resources ||= if action_collection.respond_to?(:filtered_collection)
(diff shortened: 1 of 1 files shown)


vsingh-msys:
LGTM :+1: 

I would suggest not to use the safe navigation operator here, might be we can think of using `respond_to?`

```
@updated_resources ||= begin
     action_collection.filtered_collection(max_nesting: 0, up_to_date: false, skipped: false, unprocessed: false) if action_collection.respond_to?(:filtered_collection)
end || {}
```
Otherwise @lamont-granquist suggest us better?

kimbernator:
I agree the readability left something to be desired, so I've updated it to what I believe is much more readable.

lamont-granquist:
safe navigation operator should be fine here, though, that's what its there for.

and i used it a lot in similar code in the data collector:

https://github.com/chef/chef/blob/master/lib/chef/data_collector/run_end_message.rb#L58-L59

PR #9046, opened 2019-11-02
systemd_unit needs to log

### LINKED ISSUE #9045: systemd_unit resource doesn't log when it takes actions
## Description
systemd_unit will take actions without ever logging anything. So in doc formatter:

```
  * systemd_unit[disable systemd-journal-remote] action disable
    - disabling unit: systemd-journal-remote.service
```

but in Logs you only get the `Processing` line and no line about it taking actions.

Anyone at scale is going to require line-oriented logs which makes the formatter-bit irrelevant and this a VERY scary bug.

## Chef Version
14.13.11

## Platform Version
Ubuntu 18

jaymzh:
Unlike other resources, `systemd` doesn't log when it does work breaking the standard Chef logging model. This fixes that. See #9045 for details.

Ccloses #9045

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -94,6 +94,7 @@ class Chef
         unless current_resource.enabled || current_resource.static
           converge_by("enabling unit: #{new_resource.unit_name}") do
             systemctl_execute!(:enable, new_resource.unit_name)
+            logger.info("#{new_resource} enabled")
           end
         end
       end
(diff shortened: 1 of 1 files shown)


tas50:
Is there a reason that the converge_by blocks aren't logging?

jaymzh:
... how did the PR template not work?! Weird. I didn't intentionally ignore it. :/ I'll fix.

jaymzh:
@tas50 converge_by blocks don't log... they Doc.

PR #9047, opened 2019-11-02
service/systemd_unit: Don't try to reenable services in an indirect status

### LINKED ISSUE #9041: service provider doesn't handle `indirect` systemd units
# Description
If a unit is "indirect":

```
● systemd-journal-remote.service - Journal Remote Sink Service
   Loaded: loaded (/lib/systemd/system/systemd-journal-remote.service; indirect; vendor preset: disabled)
```

Then `service`, when set to `:disable` it will always try to disable it, ever time. However, `systemd_unit` does not make this mistake.

## Chef Version
14.13.11

## Platform Version
Ubuntu 18

## Replication Case
On a typical Ubuntu18 box set:

```
service 'systemd-journal-remote' do
  action [:stop, :disable]
end
```

And it will re-disable it every run. But the same thing with `systemd_unit` does not.

jaymzh:
`systemd_unit` and `service` both try to disable `indirect` units on every run.

Closes #9041

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -51,6 +51,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
         current_resource.running(false)
         current_resource.enabled(false)
         current_resource.masked(false)
+        current_resource.indirect(false)
       end
     else
       current_resource.running(is_active?)
@@ -58,6 +59,7 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
 
     current_resource.enabled(is_enabled?)
     current_resource.masked(is_masked?)
+    current_resource.indirect(is_indirect?)
     current_resource
   end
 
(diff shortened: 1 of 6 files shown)


tas50:
There's a ton of unit tests failures on this change due to the --quiet flag being removed.

jaymzh:
Yup, I'll go through the tests sometime this week.

jaymzh:
Hmmm this PR doesn't work either... because it doesn't matter if you want to enable or disable, if something is `indirect` you can't do either. So whichever I set in `enabled?` won't work... it has to be inside of `enable` and `disable` methods that have to just return if it's indirect...

jaymzh:
OK, I updated this with a very different approach that should work reliably. All existing tests work and I added a few new ones.

jaymzh:
The current tests are unrelated to my changes, AFAICT.

lamont-granquist:
yeah this looks fine.

PR #9054, opened 2019-11-04
Update lixml2, libxslt, and nokogiri to the latest

tas50:
Resolve multiple CVEs in these libs


btm:
verify failure is the systemd one.

PR #9060, opened 2019-11-05
Replace hardcoded /etc/chef with Chef::Dist::CONF_DIR

bobchaos:

## Description
More work towards https://github.com/chef/chef/issues/8376

This one replaces the last known occurrences that need addressing of `/etc/chef` with a distro constant in chef-config.

I validated there's no more occurrences of `/opt/chef` too. Windows paths could also be addressed in this PR (another commit), or another PR if maintainers prefer that.

## Related Issue
https://github.com/chef/chef/issues/8376
https://github.com/chef/chef/issues/9094

## Types of changes

## Checklist:

- [NA] I have updated the documentation accordingly.
- [NA] I have added tests to cover my changes.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -34,7 +34,7 @@ require "uri" unless defined?(URI)
 require "addressable/uri" unless defined?(Addressable::URI)
 require "openssl" unless defined?(OpenSSL)
 require "yaml"
-require "chef/dist"
+require_relative "dist"
 
 module ChefConfig
 
@@ -646,9 +646,9 @@ module ChefConfig
       if chef_zero.enabled
         nil
       elsif target_mode?
-        platform_specific_path("#{Chef::Dist::CONF_DIR}/#{target_mode.host}/client.pem")
+        platform_specific_path("#{ChefConfig::Dist::CONF_DIR}/#{target_mode.host}/client.pem")
       else
-        platform_specific_path("#{Chef::Dist::CONF_DIR}/client.pem")
+        platform_specific_path("#{ChefConfig::Dist::CONF_DIR}/client.pem")
       end
     end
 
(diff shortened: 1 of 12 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -33,6 +33,7 @@
 require "addressable/uri" unless defined?(Addressable::URI)
 require "openssl" unless defined?(OpenSSL)
 require "yaml"
+require "chef/dist"
```
tas50: this gives us a circlular dep between chef-config and chef itself.

bobchaos: good point; I'll add a dist.rb to chef-config to avoid that

bobchaos: See latest commit for proposed fix, but this has me wondering if dist.rb shouldn't simply be entirely owned by chef-config. Some quick and dirty sed could address that if we want to try it.

bobchaos:
the windows tests look to have been cancelled. Did we reach some sort of execution timeout?

tas50:
Personally long term we should move the dist stuff into chefconfig and then we can rely on it in other products like Ohai which use chef config and may need to output log messages with branding

bobchaos:
> 
> 
> Personally long term we should move the dist stuff into chefconfig and then we can rely on it in other products like Ohai which use chef config and may need to output log messages with branding

When this is merged I can look into that, I want to try and finish up chef/chef ASAP to minimize the number of times we'll need to adjust CinC's pipelines.

About those Windows tests, should I worry?

chef-config/lib/chef-config/dist.rb
```diff
@@ -0,0 +1,8 @@
+module ChefConfig
+  class Dist
+    # The chef executable name. Also used in directory names.
+    EXEC = "chef".freeze
```
tas50: We shouldn't need this here. The path below isn't based on the executable name. It's really based on the company name. I'd just let CONF_DIR equal a string and avoid basing it on anything else.

lamont-granquist:
You need to address `/var/chef` as well, and you're going to need to do something about the hacktastic way that the routine in the other PR matches on the `chef` substring.  That whole platform_specific_path routine is making me very grouchy the way its brittle against that substring.  All the relevant configs probably need to do a `windows? ? "c:/chef" : "/etc/chef"` in them directly at the very least to set their defaults.  Most likely the best way would be to be able to specify the default paths for unix and windows as separate config options, which were then merged after the config was read and set.

bobchaos:
I've merged dist_windows_service from https://github.com/chef/chef/pull/9067 in here and the last commit is my proposed solution to the /var/chef and /etc/chef hackery

bobchaos:
note to self: run specs on all subprojects too :/

bobchaos:
this should be ready for another review pass. Do note I left `platform_specific_path`'s definition in the code in case it's called externally somehow, but I could add a deprecation message or something, or just plain remove it if you're all comfortable with that.

PR #9064, opened 2019-11-05
Symbolize config for ssl_verify_mode in credentials

teknofire:
## Description
When loading the configs using `~/.chef/credentials`, any value set for `ssl_verify_mode` is always ignored, and `:verify_peer` is always used. Chef expects the value to be a ruby symbol, which we can't set via the TOML `credentials` file.

This takes the value for that config and converts it to a symbol along with handling if the user specifies a string that looks like a symbol.

Example configs: 
```
[default]
client_name     = "someuser"
client_key      = "/home/user/.chef/userkey.pem"
chef_server_url = "https://CHEF_SERVER_URL/organizations/SOMEORG"
ssl_verify_mode = "verify_none"
```
or
```
[default]
client_name     = "someuser"
client_key      = "/home/user/.chef/userkey.pem"
chef_server_url = "https://CHEF_SERVER_URL/organizations/SOMEORG"
ssl_verify_mode = ":verify_none"
```

## Types of changes

## Checklist:

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -562,6 +562,16 @@ module ChefConfig
     # SSL verification settings. When set to :verify_none no HTTPS requests will
     # be validated.
     default :ssl_verify_mode, :verify_peer
+   
+    # Needed to coerce string value to a symbol when loading settings from the 
+    # credentials toml files which doesn't allow ruby symbol values
+    configurable(:ssl_verify_mode).writes_value do |value|
+      if value.is_a?(String) && value[0] == ":"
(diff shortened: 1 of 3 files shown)


lamont-granquist:
I think you can do coercsion and validation via the `writes_value` hack to mixlib-config (or if you can't, I would really like to know that because this is like the 3rd time it has come up this morning) which looks like this:

https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L141-L151

So that trick could be used to coerce ssl_verify_mode into a symbol across the board, and would get this code out of the workstation_config_loader and remove spooky action at a distance.

PR #9070, opened 2019-11-06
Remove useless search arg mangling

lamont-granquist:
This doesn't actually do anything and repetitively confuses me every
6 months when I look at it.

Array#pop(0) does nothing at all (in python it removes the first element
from the list):

```
[2] pry(main)> a = [ 1, 2, 3 ]
=> [1, 2, 3]
[3] pry(main)> a.pop(0)
=> []
[4] pry(main)> a
=> [1, 2, 3]
```

### DIFF
--- a/lib/chef/search/query.rb
+++ b/lib/chef/search/query.rb
@@ -126,9 +126,6 @@ class Chef
         return args.first if args.first.is_a?(Hash)
 
         args_h = {}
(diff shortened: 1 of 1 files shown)


lamont-granquist:
don't really need to backport it though, it literally does nothing, just makes it hard to read.

PR #9071, opened 2019-11-06
Fix search arguments for Chef-11 compatibility

lamont-granquist:
rely on identifying an integer value or a string that converts
to an integer (out of paranoia) vs. a string which is a sort
argument to be able to delete the search argument.

this makes the code backwards compatible with chef 11


lamont-granquist:
Dunno how useful this is.

We could backport to Chef-14, but 12+13 are never getting released, so anyone going Chef 11->14/15 would work but anyone going Chef 11->12/13 will never be able to be fixed.

Issue #9003, opened 2019-10-17
knife cookbook upload and probably knife upload cookbook do not validate presence of metadata.json/metadata.rb or validate important fields exist


lamont-granquist:
The way knife cookbook upload/upload cookbooks should behave is:

1.  use the metadata.json if present
2.  use the metadata.rb if present (generate the metadata.json)
3.  raise if none of them are present
4.  validate the name field was parsed from the metadata (throw if not)
5.  validate the version field was parsed from the metadata (throw if not)

right now a cookbook without any metadata can be uploaded to a chef-zero server and will get a name of `null` and a version of `"0.0.0"`.  the chef-server fails validation on the metadata.name being `null`, but we should catch that client-side.

jaymzh:
Fixed by attached PR.

PR #9073, opened 2019-11-07
Generate metadata.json from metadata.rb if not exist before knife cookbook upload or knife upload or berkshelf upload

Vasu1105:
## Description

Berkshelf generates metadata.json at the time of upload but knife cookbook upload / knife upload do not have similar behaviour.  

This change generates the metadata.json if it does not exist at the time of knife upload and knife cookbook upload and since it has the metadata.json uploaded on Chef Server it gets downloaded at the time of knife download.

## Related Issue

#9003 
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/knife/cookbook_upload.rb
+++ b/lib/chef/knife/cookbook_upload.rb
@@ -96,32 +96,42 @@ class Chef
         # to check for the existence of a cookbook's dependencies.
         @server_side_cookbooks = Chef::CookbookVersion.list_all_versions
         justify_width = @server_side_cookbooks.map(&:size).max.to_i + 2
+
+        cookbooks = []
+        cookbooks_to_upload.each do |cookbook_name, cookbook|
+          cookbooks << cookbook
+        end
+
+        tmp_cl = Chef::CookbookLoader.copy_to_tmp_dir_from_array(cookbooks)
+        tmp_cl.load_cookbooks
+        tmp_cl.compile_metadata
+        tmp_cl.freeze_versions if config[:freeze]
+
+        cookbooks_for_upload = []
+        tmp_cl.each do |cookbook_name, cookbook|
+          cookbooks_for_upload << cookbook
+          version_constraints_to_update[cookbook_name] = cookbook.version
+        end
+
         if config[:all]
-          cookbook_repo.load_cookbooks
-          cookbooks_for_upload = []
-          cookbook_repo.each do |cookbook_name, cookbook|
-            cookbooks_for_upload << cookbook
-            cookbook.freeze_version if config[:freeze]
-            version_constraints_to_update[cookbook_name] = cookbook.version
-          end
           if cookbooks_for_upload.any?
             begin
               upload(cookbooks_for_upload, justify_width)
-            rescue Exceptions::CookbookFrozen
+            rescue Chef::Exceptions::CookbookFrozen
               ui.warn("Not updating version constraints for some cookbooks in the environment as the cookbook is frozen.")
+              ui.error("Uploading of some of the cookbooks must be failed. Remove cookbook whose version is frozen from your cookbooks repo OR use --force option.")
+              upload_failures += 1
             end
-            ui.info("Uploaded all cookbooks.")
+            ui.info("Uploaded all cookbooks.") if upload_failures == 0
           else
             cookbook_path = config[:cookbook_path].respond_to?(:join) ? config[:cookbook_path].join(", ") : config[:cookbook_path]
             ui.warn("Could not find any cookbooks in your cookbook path: #{cookbook_path}. Use --cookbook-path to specify the desired path.")
           end
         else
-          cookbooks_to_upload.each do |cookbook_name, cookbook|
-            cookbook.freeze_version if config[:freeze]
+          tmp_cl.each do |cookbook_name, cookbook|
             begin
               upload([cookbook], justify_width)
               upload_ok += 1
-              version_constraints_to_update[cookbook_name] = cookbook.version
             rescue Exceptions::CookbookNotFoundInRepo => e
               upload_failures += 1
               ui.error("Could not find cookbook #{cookbook_name} in your cookbook path, skipping it")
@@ -147,6 +157,8 @@ class Chef
         unless version_constraints_to_update.empty?
           update_version_constraints(version_constraints_to_update) if config[:environment]
         end
+
+        tmp_cl.unlink!
       end
 
       def cookbooks_to_upload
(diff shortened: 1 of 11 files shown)


lamont-granquist:
berkshelf should already do this and has specs around it.

there's also a complicated case where you want to create the metadata.json file but need to be very careful to delete it so as not to pollute the users repository -- or else you need create a tempdir, copy all the files there, generate the metadata.json in that directory, then clean up the directory.

That is this bit:

https://github.com/berkshelf/berkshelf/blob/2f9b16496f9edafc3db72dad4304615238584f69/lib/berkshelf/uploader.rb#L63

And this horrible bit:

https://github.com/berkshelf/berkshelf/blob/2f9b16496f9edafc3db72dad4304615238584f69/lib/berkshelf/uploader.rb#L88-L95

Vasu1105:
> berkshelf should already do this and has specs around it.
> 
> there's also a complicated case where you want to create the metadata.json file but need to be very careful to delete it so as not to pollute the users repository -- or else you need create a tempdir, copy all the files there, generate the metadata.json in that directory, then clean up the directory.
> 
> That is this bit:
> 
> https://github.com/berkshelf/berkshelf/blob/2f9b16496f9edafc3db72dad4304615238584f69/lib/berkshelf/uploader.rb#L63
> 
> And this horrible bit:
> 
> https://github.com/berkshelf/berkshelf/blob/2f9b16496f9edafc3db72dad4304615238584f69/lib/berkshelf/uploader.rb#L88-L95

Yes , I was talking about the manual testing of berkshelf upload commands with this change and not the spec coverage.

@lamont-granquist I am not clear with your comments so may be I need more clarification. Do I need any updates in this PR ?

lib/chef/knife/cookbook_upload.rb
```diff
@@ -202,15 +205,46 @@ def assert_environment_valid!
...
+            Chef::CookbookUploader.new(cookbooks_to_upload, force: config[:force], concurrency: config[:concurrency]).upload_cookbooks
+          rescue Chef::Exceptions::CookbookFrozen => e
+            ui.error e
+            raise
+          ensure
+            # deletes the generated metadata from local repo.
+            if compiled_metadata
+              GC.start
+              File.unlink(compiled_metadata)
+            end
```
lamont-granquist: this ensure would need to go inside of the cookbooks loop so that for every cookbook the metadata gets deleted.

lib/chef/knife/cookbook_upload.rb
```diff
@@ -202,15 +205,46 @@ def assert_environment_valid!
...
+              if compiled_metadata
+                proxy_cookbook_path = "#{tmp_dir_path}/#{cb.name}"
+                # Make a symlink
+                file_class.symlink cb.root_dir, proxy_cookbook_path
+                proxy_loader = Chef::Cookbook::CookbookVersionLoader.new(proxy_cookbook_path)
+
+                proxy_loader.load_cookbooks
+                cookbook_to_upload = proxy_loader.cookbook_version
+                cookbook_to_upload.freeze_version if options[:freeze]
+                cookbooks_to_upload << cookbook_to_upload
```
lamont-granquist: If you're trying to implement my suggestion to do a tmpdir what that should probably look more like is:

```
  create the temp dir
  loop over all the cookbooks
    copy whole cookbook to "#{tmp_dir_path}/{cb.name}"  # do for every one
    use CookbookVersionLoader to build the CookbookVersion object
    call compile_metadata on the CookbookVersion object
    add to the cookbooks_to_upload list
  end
  upload all the cookbooks in the cookbooks_to_upload list
  <tempdir is deleted when it goes out of scope along with all the metadata.json>
```

if you do that, then you don't need the ensure/GC.start/File.unlink magic.

there's some excessive copying that isn't strictly necessary here, but i wouldn't worry about trying to speed it up, this isn't very performance critical code, and the upload across the network will take a lot longer than pushing files around on disk.

lamont-granquist: I suspect that a much cleaner way to write this would be:

```
# Chef::CookbookLoader.working_directory should be a factory method that produces a 
# Chef::CookbookLoader object after creating a tmpdir and copying all the cookbooks in the
# Hash passed to it here.  The tmpdir object needs to be saved on the CookbookLoader to
# prevent it from being deleted when the method exits (`saved_tmpdir` accessor)
cl = Chef::CookbookLoader.working_directory_from_hash(cookbooks)
# build the internal CookbookVersions
cl.load_cookbooks
# call compile_metadata on all CookbookVersions
cl.compile_metadata
# freeze all the versions in the collection
cl.freeze_versions if options[:freeze]
# upload all the versions
cl.upload(force: config[:force], concurrency: config[:concurrency])
# FileUtils.rm_rf the saved_tmpdir.path or throw and internal error if that is nil (doesn't have to be a pretty exception since it would mean a codebug
cl.unlink!
```

Vasu1105: Okay @lamont-granquist will have a look at this and update you accordingly. Thank you :)

lib/chef/knife/cookbook_upload.rb
Vasu1105: Ohh yes looks like I missed something in testing. Thank you.

lib/chef/knife/cookbook_upload.rb
Vasu1105: @lamont-granquist Are you proposing here to refactor the current code at line https://github.com/chef/chef/blob/master/lib/chef/knife/cookbook_upload.rb#L96-L142 ?

As per the current code if user select to upload all cookbooks, there is a single call to upload method. https://github.com/chef/chef/blob/master/lib/chef/knife/cookbook_upload.rb#L106

But in case of user selects specific cookbooks to upload then each cookbook gets uploaded separately https://github.com/chef/chef/blob/master/lib/chef/knife/cookbook_upload.rb#L116 and the progress is shown using upload_ok and upload_failure flags.

lamont-granquist: For the upload-everything use case you can probably do this instead of Chef::CookbookLoader.working_directory_from_hash:

```
temp_cl = Chef::CookbookLoader.working_directory_from_cl(cl)
```

So that routine should create a tmpdir and then copy all of the cookbooks in the specified cookbook_loader object to a new cookbook_loader object that references the tmpdir.

lamont-granquist: For the uploads I see where we have to upload them one-by-one and knife has to do its own error handling per-cookbook so the best that could be done would be this code:

```
          cookbook_repo.each do |cookbook_name, cookbook|
            cookbooks_for_upload << cookbook
            cookbook.freeze_version if config[:freeze]
            version_constraints_to_update[cookbook_name] = cookbook.version
          end
```

could look like:

```
          temp_cl = Chef::CookbookLoader.working_directory_from_cl(cookbook_repo)
          temp_cl.freeze_versions if config[:freeze]
          temp_cl.each do |cookbook_name, cookbook|
            cookbooks_for_upload << cookbook
            version_constraints_to_update[cookbook_name] = cookbook.version
          end
          [ .. and then call upload on every cookbook ... ]
```

lamont-granquist:
Looks good, I'll need to hold this until after I get back from vacation in 2 weeks though.

lib/chef/knife/cookbook_upload.rb
```diff
@@ -93,56 +96,85 @@ def run
...
-            exit 1
-          end
-        end
+            if config[:all]
+              if cookbooks_for_upload.any?
+                begin
+                  upload(cookbooks_for_upload, justify_width)
+                rescue Chef::Exceptions::CookbookFrozen
+                  ui.warn("Not updating version constraints for some cookbooks in the environment as the cookbook is frozen.")
+                  ui.error("Uploading of some of the cookbooks must be failed. Remove cookbook whose version is frozen from your cookbooks repo OR use --force option.")
```
Vasu1105: @lamont-granquist  This take cares of UX part for #9157. Let me know if there any changes needs to be done here in error message.

PR #9081, opened 2019-11-08
Add new chef_sleep resource

tas50:
We highly recommend you never use this, but there's times when you 100%
need to have the client sleep for a period of time after some work.
Example: windows_service will start a service and the OS will say that
it's "started" when it's not. Before doing additional work that relies
on the service being started you may need to sleep. This would be a good
time to notify on the restart or start action.


lib/chef/resource/chef_sleep.rb
```diff
@@ -0,0 +1,69 @@
...
+
+class Chef
+  class Resource
+    class ChefSleep < Chef::Resource
+      resource_name :chef_sleep
+      provides :chef_sleep
+
+      unified_mode true
+
+      description "Use the chef_sleep resource to sleep for a number of seconds during a #{Chef::Dist::PRODUCT} run. This resource should only be used when a command or service exits successfuly , but is not fully ready for the next step of the recipe."
```
kagarmoe:
```suggestion
      description "Use the chef_sleep resource to set the number of seconds to sleep during a #{Chef::Dist::PRODUCT} run. Only use this resource when a command or service exits successfully but is not ready for the next step of the recipe."
```

lib/chef/resource/chef_sleep.rb
```diff
@@ -0,0 +1,69 @@
...
+      introduced "15.5"
+      examples <<~DOC
+        Sleep for 10 seconds
+        ```ruby
+        chef_sleep '10'
+        ```
+
+        Sleep for 10 seconds with a friendly resource name for logging
```
kagarmoe: I don't get the "friendly resource name for logging" part.

lib/chef/resource/chef_sleep.rb
```diff
@@ -0,0 +1,69 @@
...
+        ```ruby
+        chef_sleep 'wait for the service to start' do
+          seconds 10
+        end
+        ````
+
+        Use a notify from another resource to sleep only when necessary
```
kagarmoe:
```suggestion
        Use a notification from another resource to sleep only when necessary
```

lib/chef/resource/chef_sleep.rb
tas50: Instead of the resource name being 10 it's an actual descriptive name

PR #9083, opened 2019-11-12
Fix action collection nilclass error early in runs

### LINKED ISSUE #9038: chef_version metadata violation raises incorrect error

lamont-granquist:
closes #9043
closes #9038

### DIFF
--- a/lib/chef/resource_reporter.rb
+++ b/lib/chef/resource_reporter.rb
@@ -183,7 +183,7 @@ class Chef
 
     # get only the top level resources and strip out the subcollections
     def updated_resources
-      @updated_resources ||= action_collection.filtered_collection(max_nesting: 0, up_to_date: false, skipped: false, unprocessed: false)
+      @updated_resources ||= action_collection&.filtered_collection(max_nesting: 0, up_to_date: false, skipped: false, unprocessed: false) || {}
     end
 
     def total_res_count


lamont-granquist:
yep, unfortunately i love the safe nav operator...

PR #9086, opened 2019-11-12
reserve a number for chef-sugar deprecation warnings

lamont-granquist:
the work of implementing the warnings is happening in the chef-sugar gem but we need to not have an id collision.


tas50:
Ideally we never need this, but sure

PR #9095, opened 2019-11-14
Additional chef-utils helpers for rhel6/7/8

lamont-granquist:
Kind of mixes up the distinction between platform_family vs platform_version helpers, but these seem to overwhelmingly make sense and are the kinds of things people will look for, and are common.

Stolen somewhat from facebook's helpers where they have `centos6/7/8?` helpers, but generalized instead to the rhel platform_family.  To get more specific users should probably `rhel6? && centos?` if they deeply care about centos versus actual redhat.

### DIFF
--- a/chef-utils/README.md
+++ b/chef-utils/README.md
@@ -82,7 +82,10 @@ There are helpers here which are also meta-families which group together multipl
 * `macos?`
 * `netbsd?`
 * `openbsd?`
-* `rhel?` - includes redhat, centos, scientific, oracle
+* `rhel?` - includes redhat, centos, scientific, oracle, clearos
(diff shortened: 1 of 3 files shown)


robbkidd:
I still love `rhelish?` as the umbrella term, but I get your meaning.

lamont-granquist:
Yeah, `rhelish?` would probably meet the same fate as `epic_fail true` through....  (and I used the latter all the time)

lamont-granquist:
Yes, to be clear this is pretty specifically for rhel as a special case.  Unless we could somehow easily autogenerate PV helpers for other distros I'm very opposed to the infinite yak shave that would occur.  Ubuntu helpers would be mostly terrible.

lamont-granquist:
Point releases like `rhel6_10?` are also a HARD NOPE.

Issue #6682, opened 2017-12-15
Checksum problem (similar to #4746?) - case-sensitivity


andybrucenet:
## Description
Installing SQL Server Express 2016 via chef-client fails due to case-sensitive SHA256 checksum match.

## Chef Version
12.7.2

## Platform Version
Windows Server 2016

## Replication Case
Follow the instructions to add a SQL Server database recipe as documented on https://learn.chef.io/modules/create-a-web-app-cookbook/windows/virtualbox/configure-sql-server#/.

My recipe looks like:
```
# recipes/database.rb
include_recipe 'sql_server::server'

```

Attributes are:
```
# attributes/default.rb
default['sql_server']['accept_eula'] = true
default['sql_server']['version'] = '2016'
```

## Client Output

The chef-client output is:
```
Recipe: sql_server::server
         * template[C:\Users\CHEF-U~1\AppData\Local\Temp\kitchen\cache\ConfigurationFile.ini] action create
         * windows_package[Microsoft SQL Server 2016 (64-bit)] action install
         Recipe: <Dynamically Defined Resource>
           * remote_file[C:\Users\CHEF-U~1\AppData\Local\Temp\kitchen\cache\package\SQLEXPR_x64_ENU.exe] action create

             ================================================================================
             Error executing action `create` on resource 'remote_file[C:\Users\CHEF-U~1\AppData\Local\Temp\kitchen\cache\package\SQLEXPR_x64_ENU.exe]'
             ================================================================================

             Chef::Exceptions::ChecksumMismatch
             ----------------------------------
             Checksum on resource (2a5b64) does not match checksum on content (2a5b64)

             Compiled Resource:
             ------------------
             # Declared in

             remote_file("C:\Users\CHEF-U~1\AppData\Local\Temp\kitchen\cache\package\SQLEXPR_x64_ENU.exe") do
        provider Chef::Provider::RemoteFile
        action [:create]
        retries 0
        retry_delay 2
        default_guard_interpreter :default
        source ["https://download.microsoft.com/download/9/0/7/907AD35F-9F9C-43A5-9789-52470555DB90/ENU/SQLEXPR_x64_ENU.exe"]
        use_etag true
        use_last_modified true
        checksum "2a5b64"
        path "C:\\Users\\CHEF-U~1\\AppData\\Local\\Temp\\kitchen\\cache\\package\\SQLEXPR_x64_ENU.exe"
        atomic_update true
             end


           ================================================================================
[...]

PR #9102, opened 2019-11-15
Improve input validation on windows_package

tas50:
- Make sure the correct installer_type symbol is passed and provide a useful error message when a bogus value is provided instead of just saying it needs to be a symbol and hoping a user eventually stumbles onto the docs
- Add descriptions fields to all the properties and update them since the ones on the docs site right now aren't the best
- Coerce the provided checksum value to lowercase. Right now if it's uppercase, but valid the resource will fail and the message provided is not useful. See https://github.com/chef/chef/issues/6682


tas50:
Ping @chef/docs-team for the descriptions. These end up on the docs site

Issue #9101, opened 2019-11-15
Figure out how to avoid shipping fauxhai with chef


tas50:
Right now we're shipping fauxhai with the chef package because we use it for testing chef-utils. It's 8 megs on disk and it would be great to avoid shipping it since it's not used on the customer side of things.

Possible solution:

Create a fauxhai-slim gem that doesn't include the actual dump files or the binary since fauxhai the library will reach out the github to download any dump files it needs. This way we could still test chef/chef, but avoid shipping 8 megs of ohai dumps.

tas50:
We did this

PR #9103, opened 2019-11-16
Use fauxhai-slim-ng

tas50:
This is 8 megs smaller. Let's see if it actually runs the tests in CI correctly. Works fine locally.

This is a short term bandaid until we can skip shipping dev deps into our packages.


tas50:
Fixes #9101

lamont-granquist:
well except you broke stuff

PR #9104, opened 2019-11-16
Don't ship the extra rake tasks in the gem

tas50:
We only need the rspec task which gets run from within the gem artifact
in the test phase. Skips all the other stuff.


tas50:
Fixes #9104

Issue #9080, opened 2019-11-08
linux_user: when modifying a users' homedir, tries to create it even if it exists


jaymzh:
## Description

When `linux_user` modifies a user to set it's homedir, if that homedir exists, it tries to create it, fails, and fails the run.

It looks like this is from passing -m always

## Chef Version
14.13.11

## Platform Version
Ubuntu 18.04

## Replication Case
setup a user. change it's homedir to something that exists.

## Client Output
```
---- Begin output of ["usermod", "-d", "/var/lib/openvpn/chroot", "-m", "nm-openvpn"]
 ----
STDOUT: 
STDERR: usermod: directory /var/lib/openvpn/chroot exists
---- End output of ["usermod", "-d", "/var/lib/openvpn/chroot", "-m", "nm-openvpn"] -
---
Ran ["usermod", "-d", "/var/lib/openvpn/chroot", "-m", "nm-openvpn"] returned 12
```

skippyj:
Based on what I know about the "usermod" command, that message is purely informational and should not be evaluating the "<x> directory already exists" message as an error. The manpage of usermod specifically addresses what happens if the passed "-d" directory already exists: https://linux.die.net/man/8/usermod 

tl;dr chef is not actually attempting to re-create the directory, it is simply reading the resulting message as an error when the directory already exists. Without too much looking (on my part) it's possible that the usermod command, despite being informational, is actually returning an err exit code.

jaymzh:
Yeah, it's def `usermod`. But the question is can we handle this better. The man page doesn't list the exit codes, but if 12 is unique to this, then we could set the allowed exit codes to `[0,12]`.

skippyj:
Yeah I agree. I haven't actively contributed to this project before, but I'll take a crack at it.

skippyj:
For reference, the entry reference in usermod that corresponds to code 12: https://github.com/shadow-maint/shadow/blob/b49712ed328ded0cd8161542ca13cf5d4cf55e5f/src/usermod.c#L92

"/* unable to complete home dir move */"

skippyj:
Created branch on my fork
https://github.com/skippyj/chef/commit/768d0f18a091622902e7eefcb6b2f68bc17d3145

skippyj:
Now that the fix is merged, can this be closed or are there further verification steps?

jaymzh:
Fixed by attached PRs

PR #9105, opened 2019-11-17
Resolve non-zero "success" error code issues with linux_user re…

skippyj:
This fixes an issue related to certain usermod commands returning non-0 exit codes but still functioning as expected.
Signed-off-by: Jonathan Jones <gitmaster@thespooky.house>

## Description
`usermod`, specifically, has the potential to issue non-zero return codes, despite the command functioning as expected. This can be seen here: https://github.com/shadow-maint/shadow/blob/b49712ed328ded0cd8161542ca13cf5d4cf55e5f/src/usermod.c#L92

Error code 12 fires when a user is modified to have a different home directory and that directory already exists. `shell_out` only accepts `0` return codes as success, when in this case it will receive `12`. Any error code 12 from usermod is now evaluated for the "already exists" stderr. It will still fail on error code 12 if this is not the error message. 

I am not sure if I need to do anything for additional test coverage - I don't _think_ so, but this is my first contribution. 

## Related Issue
Addresses https://github.com/chef/chef/issues/9080

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/user/linux.rb
+++ b/lib/chef/provider/user/linux.rb
@@ -24,23 +24,23 @@ class Chef
         provides :user, os: "linux"
 
         def create_user
-          shell_out!("useradd", universal_options, useradd_options, new_resource.username)
+          shell_out!("useradd", universal_options, useradd_options, new_resource.username, returns: return_codes)
         end
 
         def manage_user
-          shell_out!("usermod", universal_options, usermod_options, new_resource.username)
+          shell_out!("usermod", universal_options, usermod_options, new_resource.username, returns:  return_codes)
         end
 
         def remove_user
-          shell_out!("userdel", userdel_options, new_resource.username)
+          shell_out!("userdel", userdel_options, new_resource.username, returns: return_codes)
         end
 
         def lock_user
-          shell_out!("usermod", "-L", new_resource.username)
+          shell_out!("usermod", "-L", new_resource.username, returns: return_codes)
         end
 
         def unlock_user
-          shell_out!("usermod", "-U", new_resource.username)
+          shell_out!("usermod", "-U", new_resource.username, returns: return_codes)
         end
 
         # common to usermod and useradd
@@ -55,6 +55,19 @@ class Chef
           opts << "-o" if new_resource.non_unique
           opts
         end
+
+        def return_codes
+          ret_codes = [0]
+          if updating_home?
+            if new_resource.manage_home
+              home_dir_exist = shell_out!("test", "-d" new_resource.home, returns [1])
+              if home_dir_exist.error?
+                opts << 12
+              end
+            end
+          end
+          ret_codes.to_s
+        end    
 
         def usermod_options
           opts = []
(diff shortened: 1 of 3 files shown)


lib/chef/provider/user/linux.rb
```diff
@@ -24,23 +24,23 @@ class Linux < Chef::Provider::User
         provides :user, os: "linux"
 
         def create_user
-          shell_out!("useradd", universal_options, useradd_options, new_resource.username)
+          shell_out!("useradd", universal_options, useradd_options, new_resource.username, returns: return_codes)
         end
 
         def manage_user
-          shell_out!("usermod", universal_options, usermod_options, new_resource.username)
+          shell_out!("usermod", universal_options, usermod_options, new_resource.username, returns: return_codes)
```
lamont-granquist: can we simplify this and just pass `returns: [ 0, 12]` here?

skippyj: So, I was torn on this. Technically, return code 12 can fail for other reasons than just the one I'm addressing. Additionally, I thought it may be pertinent to add it as a variable going forward, if only for the sake of extensibility, should other cases arise. 

When I was initially researching the problem, I found several other instances of `usermod` failing "incorrectly" in the past, such as https://bugs.launchpad.net/ubuntu/+source/shadow/+bug/1023509

spec/support/shared/unit/provider/useradd_based_user_provider.rb
```diff
@@ -190,6 +192,10 @@
   end
 
   describe "when managing a user" do
+    let(:passwd_s_status) do
```
tas50: Do we need this? Doesn't seem to be used

lib/chef/provider/user/linux.rb
lamont-granquist: So I don't like building large abstractions for a one-off use case (see https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

And what it sounds like you need to do here is accept 12 as a return code and then parse the output and make sure it matches the expected warning text if the exit code is 12, and `error!` out if the warning text isn't there.  That would actually catch the cases where the homedir can't be created due to permissions problems or I/O errors due to a corrupted disk or anything else it sounds like.

lamont-granquist: (and by collecting all the error handling for all the actions in one mega-conditional-method you'd also make it difficult for the reader to be able to look at just how we handle managing users vs. irrelevant error conditions for the other actions -- burying code in methods and/or tiny helpers makes it often harder to read than easier)

skippyj: Can do. I'll make those changes and resubmit later today. Thank you for your feedback!

spec/support/shared/unit/provider/useradd_based_user_provider.rb
skippyj: Replaced with double for usermod shell_out spoofing

skippyj:
Any new changes I need to make before this can be merged?

lamont-granquist:
shipping!

Issue #8827, opened 2019-08-19
snap_package resource never exits


threeturn:
## Description
snap_package resource never exits

## Chef Version
Chef Infra Client: 15.2.20

## Platform Version
Ubuntu 19.04

## Replication Case

```
# -*- mode: ruby -*-
# vi: set ft=ruby :


Vagrant.configure("2") do |config|

  config.vm.box = "ubuntu/disco64"
  config.vm.provision "shell", inline: <<-SHELL
  mkdir -p /etc/chef/accepted_licenses/
  touch /etc/chef/accepted_licenses/chef_infra_client
  touch /etc/chef/accepted_licenses/inspec
  SHELL
  config.vm.provision "chef_apply" do |chef|
    chef.recipe = 'snap_package "kubectl"'
  end

end
```

## Client Output

```
[2019-08-19T06:45:57+00:00] INFO: Run List is []
[2019-08-19T06:45:57+00:00] INFO: Run List expands to []
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * snap_package[kubectl] action install
[2019-08-19T06:45:57+00:00] INFO: Processing snap_package[kubectl] action install ((chef-apply cookbook)::(chef-apply recipe) line 1)
```

lamont-granquist:
Suspicion is this has something to do with interactive prompting.

ndunn990:
I also appear to be having this issue with just a simple chef-client run.

Relevant portion of recipe:
```
service 'snapd.service' do
  action [ :enable, :start ]
end

snap_package 'install-snaps' do
  package_name node['snap_packages']
  timeout 90
  retries 3
  retry_delay 10
  action :install
end
```

Confirmed attribute value:
`default['snap_packages'] = ['wekan']`

Server Info:
```
cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

```
Chef-Client output using a Policyfile:
```
Starting Chef Infra Client, version 15.4.45
[...]

Issue #9099, opened 2019-11-15
snap_package resource gives requires classic confinement error


kapilchouhan99:
## Description
snap_package resource gives requires classic confinement error

## Chef Version
Chef Infra Client: 15.2.20
## Platform Version
Ubuntu 19.04

## Replication Case
```
snap_package 'install package' do
  package_name      ["kubectl"]
end
```
## Client Output
Facing this error:
```
[2019-11-15T15:01:51+05:30] FATAL: RuntimeError: snap_package[install package] ((chef-apply cookbook)::(chef-apply recipe) line 39) had an error: RuntimeError: status: Bad Request, kind: snap-needs-classic, message: snap "kubectl" requires classic confinement
```
After I have passed `classic` as a option, but getting the same error:
```
snap_package 'install package' do
  package_name      ["kubectl"]
  options           ["classic"]
end
```
```
[2019-11-15T15:05:43+05:30] FATAL: RuntimeError: snap_package[install package] ((chef-apply cookbook)::(chef-apply recipe) line 39) had an error: RuntimeError: status: Bad Request, kind: snap-needs-classic, message: snap "kubectl" requires classic confinement
```

kapilchouhan99:
As per my analysis, I think `requires classic confinement` issue is not from `chef` but from `snapd` side. I have verified it with the curl command: 
```
sudo curl --unix-socket /run/snapd.socket -H "Content-Type: application/json" -X POST -d '{"action": "install", "snaps": ["kubectl"], "classic": true}' http://localhost/v2/snaps | jq .
```
After running this command, facing the same error: 
```
{
  "type": "error",
  "status-code": 400,
  "status": "Bad Request",
  "result": {
    "message": "snap \"kubectl\" requires classic confinement",
    "kind": "snap-needs-classic",
    "value": "kubectl"
  }
}
```
Have added topic https://forum.snapcraft.io/t/facing-requires-classic-confinement-error/14163 and report a bug https://bugs.launchpad.net/snappy/+bug/1853122 regarding the same issue.

kapilchouhan99:
Its fixed in this #10042 , So closing this issue.

PR #9106, opened 2019-11-18
Socket response reading fixes while installing snap packages

kapilchouhan99:

## Description
- Earlier we used `socket.read` for reading the response from the socket, at that time ruby freezes, because the response body was not ending with the newline, So `socket.read` trying to read more data
- Fix this bug by using `socket.read_nonblock(socket.nread)` instead of `socket.read`
- Added `version` property to install a specific version of the snap package
- I found some minor bugs and fixed them here only
- Refactor some of the code
- Added test cases

## Related Issue
Fixes: #8827 #9099 

## Types of changes

## Checklist:


lib/chef/provider/package/snap.rb
```diff
@@ -62,27 +61,24 @@ def get_current_versions
...
         end
 
-        def install_package(names, versions)
+        def install_package(names, _versions)
           if new_resource.source
             install_snap_from_source(names, new_resource.source)
           else
-            resolved_names = names.each_with_index.map { |name, i| available_version(i).to_s unless name.nil? }
-            install_snaps(resolved_names)
+            names.each { |pkg| install_snaps(pkg) }
```
lamont-granquist: that does not look like the right way to fix this.

if the version is passed we should definitely be using that, so ignoring the argument is guaranteed to be buggy.

the problem with this looks like it is copying code from the dnf provider, where the available_version function does not return a version, but a name+version+arch tuple object called a Version, which when you call `#to_s` on it produces something like "name-version.arch" which can be passed to `dnf install name-version.arch`.

Version and available_version was probably the wrong word to use there.

kapilchouhan99: After did require analysis, I found, If we want to install a specific version of the snap package, then we can install a specific version with `channel` property like this: `sudo snap install kubectl --channel=1.6/stable --classic` , where `1.6` is the version and `stable` is the channel name.

so here we have two approaches:
1.) we need to pass `channel` property value with the version and there is no need to add version property. 
suppose if we want to install `node` package with `8` version, then we need to set `channal` value like this

```
snap_package 'install package' do
  package_name      ["node"]
  options           ["classic"]
  channel           "8/stable"
  action :install
end
```
OR

2.) Add new `version` property in the snap resource
In this approach, if we want to install `node` package with `8` version, then we need to set `version` value like this:
```
snap_package 'install package' do
  package_name      ["node"]
  options           ["classic"]
  version           "8"
  action :install
end
```
and also we need to update some of the code.

lib/chef/provider/package/snap.rb
```diff
@@ -356,6 +377,17 @@ def set_installed_package_conf(name, value)
           wait_for_completion(id)
         end
 
+        # Helper to construct error message
+        #
+        # @return [String] a formatted error message
+        def available_versions(elements)
```
lamont-granquist: that seems to be very confusing to have `available_version` and `available_versions`, this should probably be named something else.

lib/chef/provider/package/snap.rb
```diff
@@ -268,51 +283,61 @@ def generate_multipart_form_data(snap_name, action, options, path, content_lengt
...
         #   @param channel [String] The release channel.  Ex. stable
-        #   @param options [Hash] Misc configuration Options
+        #   @param options [Array] An array of configuration Options
+        #   @param version [String] A snap package version
         #   @param revision [String] A revision/version
-        def generate_snap_json(snap_names, action, channel, options, revision = nil)
+        def generate_snap_json(action, channel, options, version, revision = nil)
           request = {
               "action" => action,
-              "snaps" => snap_names,
```
lamont-granquist: So the problem with this is that if snap grows the concept of transactions that this change will break transactions (and this is something to be very worried about because package managers will inevitably grow this concern if snap doesn't have it already).  So when we're given multiple packages we should install them all together as a batch.

Since it looks like the API doesn't support installing individual versions when installing multiple packages that means we can't either, but we shouldn't shape the whole API around that edge case.  We should just throw if the user attempts to do a multipackage install with pinned versions and state that snap does not support that (yet?).  Then when given a single package, with a version, we can do the individual install of the pinned package.  That also makes it obvious to the end user that we cannot do a single-transaction install of multiple pinned packages, forces them to do a series of single snap_package calls, which is entirely the correct way for them to think about it.

tas50:
@kapilchouhan99 can you rebase this on master so we can see if that resolves the expeditor config warnings.

lib/chef/provider/package/snap.rb
```diff
@@ -46,6 +45,7 @@ def define_resource_requirements
             a.failure_message Chef::Exceptions::Package, "Package #{new_resource.package_name} not found: #{new_resource.source}"
             a.whyrun "assuming #{new_resource.source} would have previously been created"
           end
+          raise Chef::Exceptions::Package, "You can not install multipackages with pinned versions" if is_versions_passed?
```
tas50, line 48:
```suggestion
          raise Chef::Exceptions::Package, "You can not specify multiple packages via an array when also specifying pinned versions" if is_versions_passed?
```

tas50:
Closing this out now since @jaymzh pulled in some chunks and rolled with it in #10042

PR #9111, opened 2019-11-18
Fix failures in build_essential on rhel platforms

### LINKED ISSUE #9107: build_essential failing to run for CentOS
## Description
Using the build_essential on a CentOS 7 node now gives the following:

```
* build_essential[default] action install[2019-11-18T10:18:48-05:00] WARN:         The build_essential resource does not currently support the 'rhel' platform family. Skipping...
```

It would appear that the fedora_derived definition is not correctly encompassing RHEL platforms.

## Chef Version
Chef Infra Client version 15.5.9

## Platform Version
  "platform": "centos",
  "platform_version": "7.6.1810",
  "platform_family": "rhel",

## Replication Case
- Create a CentOS client.
- Run a cookbook that requires the build_essential resource.

## Client Output

```
* build_essential[default] action install[2019-11-18T10:18:48-05:00] WARN:         The build_essential resource does not currently support the 'rhel' platform family. Skipping...
```

lamont-granquist:
- fixes bug inside the action which was breaking build_essential on rhel
- wires up some more testing in the provider resolver spec
- adds a property that we can turn on so that this kind of breakage will raise an error in our end-to-end testing

i'm not entirely sure raise_if_unsupported true should be the default behavior, but that would be a breaking change and we definitely need to be able to bypass that for testing.

closes #9107

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/default.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/default.rb
@@ -2,7 +2,7 @@
 # Cookbook:: end_to_end
 # Recipe:: default
 #
-# Copyright:: 2014-2019, Chef Software, Inc.
+# Copyright:: 2014-2019, Chef Software Inc.
 #
 
 hostname "chef-bk-ci.chef.io"
@@ -36,7 +36,9 @@ yum_repository "epel" do
   only_if { platform_family?("rhel") }
 end
 
-build_essential
+build_essential do
+  raise_if_unsupported true
+end
 
 include_recipe "::packages"
 
--- a/lib/chef/resource/build_essential.rb
+++ b/lib/chef/resource/build_essential.rb
@@ -56,12 +56,11 @@ class Chef
         case
         when debian?
           package %w{ autoconf binutils-doc bison build-essential flex gettext ncurses-dev }
-        when false # rubocop:disable Lint/LiteralAsCondition
-        # when fedora_derived?
-          package %w{ autoconf bison flex gcc gcc-c++ gettext kernel-devel make m4 ncurses-devel patch }
+         when fedora_derived?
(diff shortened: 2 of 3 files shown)


lib/chef/resource/build_essential.rb
```diff
@@ -45,24 +45,29 @@ class BuildEssential < Chef::Resource
         description: "Install the build essential packages at compile time.",
         default: false, desired_state: false
 
+      property :raise_if_unsupported, [TrueClass, FalseClass],
+        description: "Raise a hard error on platforms where this resource is unsupported.",
+        default: false, desired_state: false # FIXME: make this default to true
```
tas50: making the default to continue on when we don't do what the user says seems like a really big anti-pattern. if that's what the user wants then they should specify it. Shouldn't be the default

kitchen-tests/cookbooks/end_to_end/recipes/default.rb
```diff
@@ -36,7 +36,9 @@
   only_if { platform_family?("rhel") }
 end
 
-build_essential
+build_essential do
+  raise_if_unsupported true
```
tas50: We shouldn't be using this in the kitchen tests. It would be better to have it blow up in our faces when we add new platforms so we can make sure we add support for those.

lib/chef/resource/build_essential.rb
lamont-granquist: Its the current behavior, it would likely be breaking to change it

kitchen-tests/cookbooks/end_to_end/recipes/default.rb
lamont-granquist: Right but that's what it does now by raising?

tas50: Slowly walking away from keyboard. Yep I'm an idiot

PR #9112, opened 2019-11-19
add new option bootstrap_product for install cinc via bootstrap

sawanoboly:
Signed-off-by: sawanoboly <sawanoboriyu@higanworks.com>

This changes allow install cinc or other projects such as compatible.

## Description

- add new option `--bootstrap-product PRODUCT` to  bootstrap

## Related Issue

- https://github.com/higanworks/knife-zero/issues/133

## Types of changes

## Checklist:


bobchaos:
Have you considered addressing this with a distribution constant (defined in lib/chef/dist.rb), probably `Chef::Dist::SHORT`? The end result would be the same, that is knife will install the correct product on remote targets, but it doesn't require any additional configuration. If knife was installed using Chef, it would distributes Chef, and same with CinC.

This incidentally prevents people using CinC from accidentally installing the Chef client.

sawanoboly:
I've asked https://gitlab.com/cinc-project/client/issues/9

bobchaos:
👍 It can be done when the PR implementing `SHORT` is merged and the CinC pipelines are adjusted for the new constant.

sawanoboly:
Does it mean that `short:"-? PROJECT",` should be added?

bobchaos:
What I'm proposing would not require a CLI switch at all. I'm still waiting on the PR implementing `Chef::Dist::SHORT`, but once it's in we'd reference it directly in the bootstrap template.

Your method has the advantage of allowing cross-installs (installing Chef with Cinc and vice versa), but I'm not sure that's desirable.

sawanoboly:
@bobchaos 

Thanks, I understood.

However,sy plugin depends on chef gem. So,I would be grateful if it could be used in the following configuration.

- local workstation => chef + [knife-zero](https://knife-zero.github.io)
- remote node => cinc


I have received requests from plugin users. => https://github.com/higanworks/knife-zero/issues/133

lib/chef/knife/bootstrap/templates/chef-full.erb
```diff
@@ -177,7 +177,7 @@ do_download() {
   else
     echo "-----> Installing Chef Omnibus (<%= @config[:channel] %>/<%= version_to_install %>)"
     do_download ${install_sh} $tmp_dir/install.sh
-    sh $tmp_dir/install.sh -P chef -c <%= @config[:channel] %> -v <%= version_to_install %>
```
tas50: The P is actually for product so using project here doesn't really make sense

sawanoboly: I'm trying to use a combination of the following two options:

-`--bootstrap-project cinc`
-`--bootstrap-url" https://omnitruck.cinc.sh/install.sh "`

The knife plug-in used in WorkStation relies on Chef gem, which is a bit troubling if you want to turn the remote into a cinc.

lamont-granquist: yeah that already has a meaning and it is the product.  don't want to rename that.  the ship already sailed.

lib/chef/knife/bootstrap.rb
```diff
@@ -296,6 +296,12 @@ class Bootstrap < Knife
         description: "URL to a custom installation script.",
         proc: Proc.new { |u| Chef::Config[:knife][:bootstrap_url] = u }
 
+      option :bootstrap_project,
+        long: "--bootstrap-project PROJECT",
+        description: "Project to install.",
+        proc: Proc.new { |u| Chef::Config[:knife][:bootstrap_project] = u },
```
lamont-granquist: you shouldn't need to do this Proc nonsense, even though we haven't ripped it out everywhere yet.

lib/chef/knife/bootstrap/templates/chef-full.erb
sawanoboly: Does that mean that the option name should be product instead of project?

lib/chef/knife/bootstrap.rb
sawanoboly: I see. removed.

lib/chef/knife/bootstrap/templates/chef-full.erb
lamont-granquist: yep

sawanoboly: thanks, i've changed to `bootstrap_product`

lamont-granquist:
LGTM

PR #9113, opened 2019-11-19
Add 15.5.15 release notes

tas50:

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@ This file holds "in progress" release notes for the current release under develo
 
 # Chef Infra Client 15.5.15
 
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,12 @@
 This file holds "in progress" release notes for the current release under development and is intended for consumption by the Chef Documentation team. Please see <https://docs.chef.io/release_notes.html> for the official Chef release notes.
 
-# Chef Infra Client 15.5
+# Chef Infra Client 15.5.15
+
+The Chef Infra Client 15.5.15 release includes fixes for a regression in the `build_essential` resource that caused failures on `rhel` platforms as well as a regression in starting Chef Infra Client with `enforce_path_sanity` enabled. As part of this fix we've added a new property `raise_if_unsupported` to the `build-essential` resource, which will cause the Chef Infra Client run to fail if an unknown platform is encountered instead of silently continuing.
```
IanMadd:
```suggestion
The Chef Infra Client 15.5.15 release includes fixes for two regressions. A regression in the `build_essential` resource caused failures on `rhel` platforms and a second regression caused Chef Infra Client to fail when starting with `enforce_path_sanity` enabled. As part of this fix we've added a new property, `raise_if_unsupported`, to the `build-essential` resource. Instead of silently continuing, this property will fail a Chef Infra Client run if an unknown platform is encountered.
```

PR #9114, opened 2019-11-19
fix enforce_path_sanity being set to true

lamont-granquist:
Obvious bug.  Prioritizing fixing this for now over figuring out where to add testing.


lamont-granquist:
manual testing against this branch worked.

PR #9120, opened 2019-11-21
kernel_module: Add new options property (backport from chef-15 to chef-14)

ramereth:
This backports #8887 for Chef 14. I'm not sure what the policy is for backporting something like this, so please let me know if I did it correctly or if it's even possible.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/kernel_module.rb
+++ b/lib/chef/resource/kernel_module.rb
@@ -23,6 +23,7 @@ class Chef
 
       property :options, Array,
         description: "An optional property to set options for the kernel module."
+        introduced: "15.4"
 
       property :load_dir, String,
                description: "The directory to load modules from.",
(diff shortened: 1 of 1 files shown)


lib/chef/resource/kernel_module.rb
```diff
@@ -16,11 +16,61 @@ class KernelModule < Chef::Resource
...
+        ```
+      DOC
 
       property :modname, String,
                description: "An optional property to set the kernel module name if it differs from the resource block's name.",
                name_property: true, identity: true
 
+      property :options, Array,
+        description: "An optional property to set options for the kernel module.",
+        introduced: "15.4"
```
tas50: You'll need to update this for the next 14 release number

ramereth: Would that be ``14.15`` or something else?

tas50: 14.15 seems right

PR #9147, opened 2019-12-09
Specify item path in Node.read! error message

zeusttu:
## Description
The `Node.read!` method specifies an empty error message when it fails, which complicates debugging. I added the dot-separated path to the key it was attempting to read to the error message.

It is not perfect because ideally the user would like to know which key in the hierarchy does not exist; instead it always specifies the entire requested path. Since the code itself is not aware of which key exactly does not exist, retrieving this information and putting it into the error message would increase complexity, and require refactoring to prevent it from incurring a performance penalty for successful calls. This is why I decided for now to make a simple change that always returns the entire requested path instead. If we keep it this way, we may want to document it, but I could not find API docs or any comments that look like auto-generation sources for API docs, so I do not know where to document it.

## Types of changes

( I am not sure, but I think it is a chore because it fixes a minor inconvenience that was never actually reported as a bug as far as I know.)

## Checklist:

- [ ] I have run the pre-merge tests locally and they pass. -> I don't know which tests constitute "pre-merge" tests. I ran the unit tests locally at least.
- [ ] I have updated the documentation accordingly. -> I think there is one detail that deserves documentation, but i don't know where to document it (see Description).
- [x] All new and existing tests passed. -> Some tests fail in the environment I tested in, but those same tests fail on `master` in the same environment, which has to do with `gem` somehow not reporting that `rspec` is installed. I have not re-run the tests after the most recent rebase on `master`. In the pipeline all tests pass.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco). -> I invoked the obvious fix clause, I'm guessing this is ok.

### DIFF
--- a/lib/chef/node/common_api.rb
+++ b/lib/chef/node/common_api.rb
@@ -100,7 +100,7 @@ class Chef
 
       # non-autovivifying reader that throws an exception if the attribute does not exist
       def read!(*path)
-        raise Chef::Exceptions::NoSuchAttribute unless exist?(*path)
+        raise Chef::Exceptions::NoSuchAttribute.new(path.join ".") unless exist?(*path)
 
         path.inject(self) do |memo, key|
           memo[key]
--- a/spec/unit/node/vivid_mash_spec.rb
+++ b/spec/unit/node/vivid_mash_spec.rb
@@ -163,15 +163,15 @@ describe Chef::Node::VividMash do
     end
 
     it "throws an exception when attributes do not exist" do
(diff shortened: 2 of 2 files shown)


zeusttu:
I found that your checks reported linting errors in the code I committed, specifically I used single quotes instead of double quotes. I fixed this by amending my existing commit, because it is really a fixup to that commit and not a change to any pre-existing code, and it does not look like anyone was reviewing yet anyway.

tas50:
Thanks @zeusttu and thanks for the solid descriptionn

Issue #9155, opened 2019-12-11
PolicyBuilder complains about run_list being set in JSON data even though run_list is not set. 


NAshwini:
## Description
When bootstrapping a node using the -j or --json-attribute-file options to pass JSON containing `policy_name` and `policy_group` definitions but no run_list definition,  Chef::PolicyBuilder::Policyfile complains that setting the run_list in JSON data is unsupported.
This internally set a empty runlist in first-boot attribute which leads to above issue.

## Chef Version
14.14.29

## Platform Version
Ubuntu18

## Replication Case
Create a json file with value as ` { "policy_group": "azuregrp", "policy_name": "azurepolicy" } `
Try to bootstrap a fresh node by command `knife bootstrap <IP-Address>  -x "azure" -P "<PASSWORD>" --json-attribute-file /etc/chef/bootattr.json -c "/etc/chef/knife.rb" --sudo`

## Client Output

```
Unexpected Error:
 Chef::PolicyBuilder::Policyfile::UnsupportedFeature: Policyfile does not support setting the run_list in json data.

```

## Stacktrace

```
Generated at 2019-12-11 09:43:06 +0000
Chef::PolicyBuilder::Policyfile::UnsupportedFeature: Policyfile does not support setting the run_list in json data.
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/policy_builder/policyfile.rb:107:in `initialize'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/policy_builder/dynamic.rb:155:in `new'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/policy_builder/dynamic.rb:155:in `select_implementation'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/policy_builder/dynamic.rb:74:in `load_node'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/client.rb:472:in `load_node'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/client.rb:267:in `run'
/opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.14.29/lib/chef/application.rb:303:in `run_with_graceful_exit_option'
[...]

PR #9156, opened 2019-12-11
When bootstrapping don't send an empty run_list if we are using policyfiles instead

NAshwini:

## Description
Don't send empty runlist if do not pass it.

## Related Issue
https://github.com/chef/chef/issues/9155

## Types of changes

- [ -] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -205,15 +205,14 @@ class Chef
         end
 
         def first_boot
-          (@config[:first_boot_attributes] || {}).tap do |attributes|
+          (@config[:first_boot_attributes] = Mash.new(@config[:first_boot_attributes]) || Mash.new).tap do |attributes|
             if @config[:policy_name] && @config[:policy_group]
               attributes[:policy_name] = @config[:policy_name]
               attributes[:policy_group] = @config[:policy_group]
             else
               attributes[:run_list] = @run_list
             end
-
-            attributes.delete(:run_list) if ( attributes[:policy_name] && !attributes[:policy_name].empty? ) || ( attributes["policy_name"] && !attributes["policy_name"].empty? )
+            attributes.delete(:run_list) if attributes[:policy_name] && !attributes[:policy_name].empty?
             attributes.merge!(tags: @config[:tags]) if @config[:tags] && !@config[:tags].empty?
           end
         end
--- a/lib/chef/knife/core/bootstrap_context.rb
+++ b/lib/chef/knife/core/bootstrap_context.rb
@@ -213,7 +213,7 @@ class Chef
               attributes[:run_list] = @run_list
             end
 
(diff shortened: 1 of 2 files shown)


lib/chef/knife/core/bootstrap_context.rb
```diff
@@ -213,7 +213,7 @@ def first_boot
               attributes[:run_list] = @run_list
             end
 
-            attributes.delete(:run_list) if attributes[:policy_name] && !attributes[:policy_name].empty?
+            attributes.delete(:run_list) if ( attributes[:policy_name] && !attributes[:policy_name].empty? ) || ( attributes["policy_name"] && !attributes["policy_name"].empty? )
```
tas50: Can you add a comment here so that we know why we're checking both the symbol and the string varients

NAshwini: When we pass policy_name via bootstrap options i.e `--policy-name` then we need to check it by symbol.
And when we pass policy_name by `--json-attribute-file` it comes in string. Hence both check is added.

lamont-granquist: They should just be strings, since we're coming from parsing JSON the symbols shouldn't ever be set.

Alternatively, the config setting should be a Mash so it doesn't matter.

NAshwini: Thanks @tas50 and @lamont-granquist,  When we pass `policy_name` via bootstrap options i.e `--policy-name` then we need to check it by symbol and when we pass `policy_name` by `--json-attribute-file` it comes in string. Hence both check is added.
I was trying to change config into Mash but then it will affect other bootstrap options which may impact a lot. looking into more if there is any another possible way to do so.

NAshwini:
Thanks @lamont-granquist, Optimized the code by using Mash.

spec/unit/knife/core/bootstrap_context_spec.rb
```diff
@@ -141,7 +141,7 @@
     let(:config) { { policy_name: "my_app_server", policy_group: "staging" } }
 
     it "includes them in the first_boot data and excludes run_list" do
-      expect(Chef::JSONCompat.to_json(bootstrap_context.first_boot)).to eq(Chef::JSONCompat.to_json(config))
+      expect(Chef::JSONCompat.to_json(bootstrap_context.first_boot)).to eq(Chef::JSONCompat.to_json(config.delete(:first_boot_attributes)))
```
lamont-granquist: is this delete necessary because the config hash above doesn't seem to have that key?

NAshwini: Thanks @lamont-granquist, Added appropriate expectation.

tas50:
@NAshwini can you rebase this against master so we can check this against the latest specs?

PR #9173, opened 2019-12-17
resource archive_file: apply ownership to extracted files only

bobchaos:

## Description
Currently, the `archive_file` file resource applies a `chown_R` to the entire destination directory, which could be dangerous if the user is extracting directly to places like `/etc` or `/bin`. This PR proposes fixing this by modifying ownership only of file listed in the archive, leaving the destination directory and anything else it contained prior to the run untouched.

## Related Issue
I cheated; I'm skipping right to the PR. The docs said `owner` and `group` would be applied to the extracted files, so  presumably editing the destination directory itself is considered a bug.

## Types of changes

## Checklist:

- [N/A] I have updated the documentation accordingly.
- [N/A] I have added tests to cover my changes.

### DIFF
--- a/lib/chef/resource/archive_file.rb
+++ b/lib/chef/resource/archive_file.rb
@@ -102,8 +102,11 @@ class Chef
         end
 
         if new_resource.owner || new_resource.group
-          converge_by("set owner of #{new_resource.destination} to #{new_resource.owner}:#{new_resource.group}") do
-            FileUtils.chown_R(new_resource.owner, new_resource.group, new_resource.destination)
+          converge_by("set owner of files extracted in #{new_resource.destination} to #{new_resource.owner}:#{new_resource.group}") do
+            archive = Archive::Reader.open_filename(new_resource.path)
+            archive.each_entry do |e|
+              FileUtils.chown(new_resource.owner, new_resource.group, "#{new_resource.destination}/#{e.pathname}")
+            end
           end
         end
       end


lib/chef/resource/archive_file.rb
```diff
@@ -102,8 +102,11 @@ class ArchiveFile < Chef::Resource
         end
 
         if new_resource.owner || new_resource.group
-          converge_by("set owner of #{new_resource.destination} to #{new_resource.owner}:#{new_resource.group}") do
-            FileUtils.chown_R(new_resource.owner, new_resource.group, new_resource.destination)
+          converge_by("set owner of files extracted in #{new_resource.destination} to #{new_resource.owner}:#{new_resource.group}") do
+            archive = Archive::Reader.open_filename(new_resource.path)
+            archive.each_entry do |e|
+              FileUtils.chown_R(new_resource.owner, new_resource.group, "#{new_resource.destination}/#{e.pathname}")
```
lamont-granquist: i think if each_entry includes directories then the `chown_R` is going to recursively chown files that might not be in the archive and will re-chown files that are in the archive which is probably not what you want.  OTOH if each_entry doesn't include directories then the chown_R isn't doing anything useful...

bobchaos: Good point, I think I can arrange that. I'll give it another go sometime this week by constructing a full list of the archive contents and applying using plain `chown` to it.

bobchaos: I played a bit with irb and confirmed `each_entry` returns the folders too

PR #9187, opened 2019-12-25
x509_certificate : Add the capability to automatically renew a certificate

julienhuon:
## Description
This change add the capability to automatically renew a certificate in the x509_certificate resource. It adds a new property (renew_before_expiry) wich contains the number of days before expiration. If this limit is reached, the certificate will be renewed.

Merry christmas ! 

## Types of changes

## Checklist:
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

### DIFF
--- a/lib/chef/mixin/openssl_helper.rb
+++ b/lib/chef/mixin/openssl_helper.rb
@@ -401,6 +401,30 @@ class Chef
         crl.sign(ca_private_key, ::OpenSSL::Digest::SHA256.new)
         crl
       end
+
+      # Return true if a certificate need to be renewed (or doesn't exist) according to the number
+      # of days before expiration given
+      # @param [string] cert_file path of the cert file or cert content
+      # @param [integer] renew_before_expiry number of days before expiration
+      # @return [true, false]
+      def cert_need_renewall?(cert_file, renew_before_expiry)
+        raise TypeError, 'cert_file must be a String object' unless cert_file.is_a?(String)
+        raise TypeError, 'renew_before_expiry must be a Integer object' unless renew_before_expiry.is_a?(Integer)
+
+        resp = true
+        cert_content = ::File.exist?(cert_file) ? File.read(cert_file) : cert_file
+        begin
+          cert = OpenSSL::X509::Certificate.new cert_content
+        rescue ::OpenSSL::X509::CertificateError
+          return resp
+        end
+
+        unless cert.not_after <= Time.now + 3600 * 24 * renew_before_expiry
+          resp = false
+        end
+
+        resp
+      end
     end
   end
 end
--- a/lib/chef/mixin/openssl_helper.rb
+++ b/lib/chef/mixin/openssl_helper.rb
@@ -408,8 +408,8 @@ class Chef
       # @param [integer] renew_before_expiry number of days before expiration
       # @return [true, false]
       def cert_need_renewall?(cert_file, renew_before_expiry)
-        raise TypeError, 'cert_file must be a String object' unless cert_file.is_a?(String)
-        raise TypeError, 'renew_before_expiry must be a Integer object' unless renew_before_expiry.is_a?(Integer)
+        raise TypeError, "cert_file must be a String object" unless cert_file.is_a?(String)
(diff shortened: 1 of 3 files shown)


lib/chef/mixin/openssl_helper.rb
```diff
@@ -401,6 +401,30 @@ def renew_x509_crl(crl, ca_private_key, info)
...
       end
+
+      # Return true if a certificate need to be renewed (or doesn't exist) according to the number
+      # of days before expiration given
+      # @param [string] cert_file path of the cert file or cert content
+      # @param [integer] renew_before_expiry number of days before expiration
+      # @return [true, false]
+      def cert_need_renewall?(cert_file, renew_before_expiry)
+        raise TypeError, "cert_file must be a String object" unless cert_file.is_a?(String)
+        raise TypeError, "renew_before_expiry must be a Integer object" unless renew_before_expiry.is_a?(Integer)
```
tas50: No need for this check if the property is an Int. Chef will blow up before it gets here.

lib/chef/resource/openssl_x509_certificate.rb
```diff
@@ -109,13 +109,17 @@ class OpensslX509Certificate < Chef::Resource
...
+        default: 5
+
       action :create do
         description "Generate a certificate"
 
-        unless ::File.exist? new_resource.path
+        if cert_need_renewall?(new_resource.path, new_resource.renew_before_expiry)
           converge_by("Create #{@new_resource}") do
             file new_resource.path do
-              action :create_if_missing
```
tas50: There was a design to not overwrite the cert on disk if present that was part of this design originally. You'll want to change that action based on renew_before_expiry being set.

lib/chef/mixin/openssl_helper.rb
julienhuon: True. I removed it.

lib/chef/resource/openssl_x509_certificate.rb
julienhuon: I assumed that overwrite the cert on disk only if it is expired should be enought. Ok I will overwrite it only if renew_before_expiry is set and the limit reached.

julienhuon: Corrected. Hope it was what needed.

lib/chef/resource/openssl_x509_certificate.rb
```diff
@@ -109,30 +109,41 @@ class OpensslX509Certificate < Chef::Resource
...
+        file new_resource.path do
+          action :create_if_missing
+          owner new_resource.owner unless new_resource.owner.nil?
+          group new_resource.group unless new_resource.group.nil?
+          mode new_resource.mode unless new_resource.mode.nil?
+          sensitive true
+          content cert.to_pem
+        end
+
+        unless new_resource.renew_before_expiry.nil?
```
tas50: We should do this all inline so we don't have a resource appear in the log that we skip:

```ruby
if !new_resource.renew_before_expiry.nil? && cert_need_renewall?(new_resource.path, new_resource.renew_before_expiry)
```

lib/chef/resource/openssl_x509_certificate.rb
```diff
@@ -109,30 +109,41 @@ class OpensslX509Certificate < Chef::Resource
       property :ca_key_pass, String,
         description: "The passphrase for CA private key's passphrase."
 
+      property :renew_before_expiry, Integer,
+        description: "The number of days before the expiry. The certificate will be automaticaly renewed when the value is reached."
```
tas50: we'll want to get an introduced field added here so this shows up in our docs correctly:
```ruby
introduced: 15.7
```

julienhuon: Done.

lib/chef/resource/openssl_x509_certificate.rb
julienhuon: Done.

tas50:
Looks good. Thanks for the work.

PR #9205, opened 2020-01-03
Remove support for Windows 7 / 2008 from windows_feature

tas50:
These are EOL platforms that we no longer produce artifacts for. Let's
simplify / speedup these resources by removing legacy platform checks.


lamont-granquist:
"If it ~~bleeds~~ is EOL we can kill it"

tas50:
We should just nuke this from chef 16 when we branch

PR #9216, opened 2020-01-14
Add pfx_properties property to windows_certificate resource

amunoz951:
Add pfx_properties property to windows_certificate resource to allow specifying of PFX key properties used when installing PFX certificates in Windows.

win32-certstore gem >=v0.4.0 allows these properties to be passed.
Gem updates to add_pfx function can be seen here: https://github.com/chef/win32-certstore/commit/fd9342e3e1d87011c29f8be1461642f85316043b

Signed-off-by: Alex Munoz <amunoz951@gmail.com>

## Description
This allows chef-client to install the certificate PFX key in the localsystem store even when being run as a user as well as any number of PFX key flags available to the PFXImportCertStore function.

See https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-pfximportcertstore 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/windows_certificate.rb
+++ b/lib/chef/resource/windows_certificate.rb
@@ -38,6 +38,10 @@ class Chef
       property :pfx_password, String,
         description: "The password to access the source if it is a pfx file."
 
+      property :pfx_properties, Integer,
+        description: "The dwFlags value desired for PFX storage. Automatically assigned based on the user_store property if omitted. See https://docs.microsoft.com/en-us/windows/desktop/api/wincrypt/nf-wincrypt-pfximportcertstore PFXImportCertStore function for available values.",
+        default: lazy { user_store ? 0x00001000 : 0x00000020 }
+
       property :private_key_acl, Array,
         description: "An array of 'domain\account' entries to be granted read-only access to the certificate's private key. Not idempotent."
 
@@ -134,7 +138,7 @@ class Chef
 
         def add_pfx_cert
           store = ::Win32::Certstore.open(new_resource.store_name)
-          store.add_pfx(new_resource.source, new_resource.pfx_password)
+          store.add_pfx(new_resource.source, new_resource.pfx_password, new_resource.pfx_properties)
         end
 
         def delete_cert


amunoz951:
This change would not affect most. The only case that will change is if the user_store property is set to false and chef-client runs as a specific user. In those cases, the certificate PFX key would now be installed in the localsystem store instead of currentuser. This would be a misconfiguration on the user's part for not setting user_store to false if they actually intended to have the PFX key installed under the current user only. In other words, in those cases, it was installing in the correct place, despite the misconfiguration up until this fix. In these cases where the user_store property is misconfigured, except in the rare case where the user does not have permissions to the localsystem store, this would not break anything, but could broaden the scope of access to the PFX key. Regardless, this is describing the potential effect of default behavior and can be overridden by specifying a value for the new property.

lib/chef/resource/windows_certificate.rb
```diff
@@ -38,6 +38,10 @@ class WindowsCertificate < Chef::Resource
       property :pfx_password, String,
         description: "The password to access the source if it is a pfx file."
 
+      property :pfx_properties, Integer,
+        description: "The dwFlags value desired for PFX storage. Automatically assigned based on the user_store property if omitted. See https://docs.microsoft.com/en-us/windows/desktop/api/wincrypt/nf-wincrypt-pfximportcertstore PFXImportCertStore function for available values.",
+        default: lazy { user_store ? 0x00001000 : 0x00000020 }
```
tas50: This is a really useful property to add, but I think this doesn't address it in a way that we'd want to merge it. Many of those pfx properties are things that our users would want to be able to set, but they don't want to have to add up hex values to determine the values they want to set. This will need to be something where the user set the properties in a friendly value and we passed handled passing in the proper hex value.

lib/chef/resource/windows_certificate.rb
```diff
@@ -134,7 +138,7 @@ def add_cert(cert_obj)
 
         def add_pfx_cert
           store = ::Win32::Certstore.open(new_resource.store_name)
-          store.add_pfx(new_resource.source, new_resource.pfx_password)
+          store.add_pfx(new_resource.source, new_resource.pfx_password, new_resource.pfx_properties)
```
tas50: Since this requires the newer gem we'll also want to bump the dependency to require the version that allows passing in pfx properties

tas50:
@amunoz951 I'm going to close this one out at this point since we haven't heard back. If you'd like to move forward with this change feel free to pop into the #chef-infra-dev channel on the Community Chef Slack and we can chat about where to take this.

PR #9235, opened 2020-01-17
Add windows helpers for install type

tas50:
I'm placing these in their own helper file since there are going to be
more of them and the only commonality is that they're for windows.


lamont-granquist:
yeah this makes sense.  `introspection` is the grab bag, but i don't think it should go there, and these aren't the platform_version ones.

lamont-granquist:
you probably should write some units for this one, you should be able to copypasta off of the existing ones reasonably easy.

Issue #9171, opened 2019-12-17
The new mac_user resource has several critical flaws


chilcote:
## Description

15.6.16 fixed the crashing bug in the `mac_user` resource, but now that it's not crashing, it has exposed several other critical issues.

- Setting a home directory with `sysadminctl` or `dscl` now require `SystemPolicySysAdminFiles` whitelisting, which cannot be set indefinitely without an MDM server (i.e. it will prompt every time chef calls the resource, asking you to "allow" Terminal to manage your system.)
- When creating a user, it does not respect `system: true` (i.e. all users are >500)
- When creating a user, it does not set the GID to its numerical GID; it uses the group "name" (i.e. "admin" instead 80)
    - this causes problems when doing things like "chown," or, more critically, using a `dir` resource to manage a home dir that's not under /Users.

I appreciate the work that went into fixing the crashing bug, and hopefully we can get these addressed as well. 

## Chef Version
15.6.16

## Platform Version
10.15.2

## Replication Case
It's as easy as trying to manage a user on macOS. For example, this resource fails all three of the points above.
```
user 'fooadmin' do
  gid 'admin'
  home '/Users/fooadmin'
  shell '/bin/bash'
  password 'this_is_a_terrible_password'
  iterations 40000
  manage_home true
  system true
  action :create
end
```

## Client Output
The output looks correct. because the user is being created (and if the process is being called via launchd, then thankfully we don't see the user permission prompts. However, if you set system to true, you can verify that the UID is still >500, and you can inspect the user plist to verify that the UniqueGID is not being set correctly.

## Stacktrace
n/a

chilcote:
Anyone?

jessepeterson:
Regarding the group ID issue just to expound a bit. It looks like macOS effectively disables a user on the system that's created with an invalid GID. I.e. the system acts like the user doesn't even exist (despite the record actually existing the Local DS plist 'database.'

I think @chilcote and I probably came up with the same solution at the same time to the `PrimaryGroupID` issue. Here's my diff:

```diff
$ diff -wu mac.rb.prev mac.rb
--- mac.rb.prev	2019-12-15 21:19:05.000000000 -0800
+++ mac.rb	2020-01-24 09:03:37.000000000 -0800
@@ -196,7 +196,7 @@
             end.run_action(group_action)
 
             converge_by("create primary group ID") do
-              run_dscl("create", "/Users/#{new_resource.username}", "PrimaryGroupID", new_resource.gid)
[...]

PR #9275, opened 2020-01-24
mac_user: fixing gid and system properties, and adding hidden property

chilcote:
Goals:
- `mac_user` provider will now use the numeric GID when creating a user, instead of passing what was literally in the resource (e.g. it will use 80 instead of "admin")
- brings back support for `system true` from the old `dscl` provider.
- adds support for new property `hidden` which will set the `IsHidden` value in the user plist.

## Description
The GID will fix issues where the system does not recognize the user (for `chown` operations, or `id`, etc)

The `system: true` works, but there are still macOS quirks. For instance, if you run chef manually in terminal, you'll be prompted to give Terminal.app privacy protection rights to be able to set the UID.  When run via launchd, which is how most orgs would use chef, macOS just ignores the UID you pass and pulls its own (usually starting at 502 and going up to the next available one). This seems to be a limitation in the `sysadminctl` tool itself, in conjunction with the added privacy protections around updating a UID for a user. The fix for this behavior would involve signing chef (or chef's ruby) and whitelisting it for `SystemPolicySysAdminFiles` via MDM.

Since what most of us are interested in is not specifically assigning a UID under 500 (which Apple doesn't even support), I'm also adding in support to this resource for the `IsHidden` dscl attribute. Setting this to `1` (which, apple terms, means "true"), the account will not show up in System Prefs or at the Login Window. 

## Related Issue
#9171

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/chef/provider/user/mac.rb
+++ b/lib/chef/provider/user/mac.rb
@@ -51,7 +51,10 @@ class Chef
             current_resource.home(user_plist[:home][0])
             current_resource.shell(user_plist[:shell][0])
             current_resource.comment(user_plist[:comment][0])
-            current_resource.hidden(user_plist[:is_hidden][0] == "1" ? true : false)
+
+            if user_plist[:is_hidden]
+              current_resource.hidden(user_plist[:is_hidden][0] == "1" ? true : false)
+            end
 
             shadow_hash = user_plist[:shadow_hash]
             if shadow_hash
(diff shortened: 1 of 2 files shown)


tas50:
@chilcote As of Chef Infra Client 15.8 we will be signing all the binaries in Chef to comply with all the new notarization requirements. So next week we'll be signed

chilcote:
@tas50 That's very interesting news! Is there info posted on what needs to be added to the kernel extension whitelist payloads?

Issue #9285, opened 2020-01-27
Knife cli fuzzy search does not work as documented


mimato:
###  Description
As documented here: https://docs.chef.io/knife_search.html#syntax, the behaviour for a node search that does not have a `:` in the search string should be to do a fuzzy search like `tags:*#{@query}* OR roles:*#{@query}* OR fqdn:*#{@query}* OR addresses:*#{@query}*`. However, after upgrading Chef from 12 to 14, we noticed this was no longer working.  

## Chef Version
14.12.3

## Platform Version
macOS 10.14.6

## Replication Case
Given a chef server with a node `mynode` and a properly configured knife set-up
* run `knife node show mynode` (the node should display)
* run `knife search node 'name:mynode'` (the node should display)
* run `knife search mynode` (the node should display but does not)

## Client Output
```
knife search node 'name:a-fake-node' -i
1 items found

a-fake-node
```

```
knife search 'name:a-fake-node' -i
1 items found

a-fake-node
```

```
knife search a-fake-node -i
0 items found
```

lamont-granquist:
closed by #9287

PR #9287, opened 2020-01-27
Fix fuzzy node search to work when the search type is a string rather than a symbol

mimato:
## Description
The knife CLI calls search with a `type` of `'node'` (a string) for node searches (or searches where the index is unspecified). The `if` statement here only checks for the symbol `:node` which meant that fuzzy searches from the knife CLI did not work.

This fixes the issue by checking if the type is  `:node` or `'node'`. I'm not sure if there's a preference to check this way or to just coerce it to a symbol beforehand. 

## Related Issue
[#9285 - Knife cli fuzzy search does not work as documented](https://github.com/chef/chef/issues/9285)

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/search/query.rb
+++ b/lib/chef/search/query.rb
@@ -63,7 +63,7 @@ class Chef
 
         args_h = hashify_args(*args)
         if args_h[:fuzz]
-          if type == :node
+          if [:node, 'node'].include?(type)
             query = fuzzify_node_query(query)
           end
           # FIXME: can i haz proper ruby-2.x named parameters someday plz?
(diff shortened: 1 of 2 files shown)


lib/chef/search/query.rb
```diff
@@ -63,7 +63,7 @@ def search(type, query = "*:*", *args, &block)
 
         args_h = hashify_args(*args)
         if args_h[:fuzz]
-          if type == :node
+          if [:node, "node"].include?(type)
```
lamont-granquist: `if type.to_sym == :node` would probably read a little better

tas50:
Nice fix. Thanks @mimato

lamont-granquist:
Yep, thanks for fixing the fix as well.

PR #9289, opened 2020-01-27
apt_repository: add a description for components when using a PPA

tas50:
When you're using a PPA there's a default here that doesn't apply to non-PPA repos.

### DIFF
--- a/lib/chef/resource/apt_repository.rb
+++ b/lib/chef/resource/apt_repository.rb
@@ -120,7 +120,7 @@ class Chef
 
       property :components, Array,
         description: "Package groupings, such as 'main' and 'stable'.",
-        default: lazy { [] }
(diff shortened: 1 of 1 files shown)


lib/chef/resource/apt_repository.rb
```diff
@@ -120,7 +120,7 @@ class AptRepository < Chef::Resource
 
       property :components, Array,
         description: "Package groupings, such as 'main' and 'stable'.",
-        default: lazy { [] }
+        default: lazy { [] }, default_description: "If using a PPA this will be 'main'."
```
mjingle: The "this" in the default description needs to be clarified in the default description unless there is other content providing context elsewhere. Grouping? Component? Property?

tas50: Updated

PR #9292, opened 2020-01-28
Move knife-acl gem commands into chef in their own namespaces

tas50:
Ship these knife commands so we can kill off knife-acl, but move them under their own namespaces.


lamont-granquist:
These all work against normal chef-server though (after the single-org open source chef-11 server was killed off)

tas50:
Reworked to just stick them into the appropriate namespaces

PR #9294, opened 2020-01-28
debian 10 ifconfig fix

lamont-granquist:
debian 10 version of net-tools switches its --version output to
stdout from stderr.


tas50:
This needs to get backported to 14

PR #9315, opened 2020-01-31
Add chef-sugar virtualization helpers

lamont-granquist:
this also adds `node.dig` as an alias for `node.read` so that Hashes can be used to mock Chef::Node because i don't want to break my own rule and take a dev time circular dep back on the chef gem (although VividMash probably needs to be moved into chef-utils eventually).

### DIFF
--- a/chef-utils/README.md
+++ b/chef-utils/README.md
@@ -148,8 +148,8 @@ Architecture Helpers allow you to determine the processor architecture of your n
 * `openvz?` - if the node is an openvz guest
 * `openvz_host?` - if the node is an openvz host
 * `guest?` - if the node is detected as any kind of guest
-* `virtual_host?` - if the node is detected as being any kind of virtual host
-* `physical?` - strictly the logical opposite of `guest?`
+* `hypervisor?` - if the node is detected as being any kind of hypervisor
+* `physical?` - the node is not running as a guest (may be a hypervisor or may be plain metal)
 * `vagrant?` - attempts to identify the node as a vagrant guest (this check may be error prone)
 
 ### Train Helpers
--- a/chef-utils/README.md
+++ b/chef-utils/README.md
@@ -135,14 +135,15 @@ Architecture Helpers allow you to determine the processor architecture of your n
 
 ### Virtualization Helpers
 
-* `kvm?` - if the node is a kvm guest
-* `lxc?` - if the node is an lxc guest
-* `parallels?`- if the node is a parallels guest
-* `vbox?` - if the node is a virtualbox guest
-* `vmware?` - if the node is a vmware guest
-* `openvz?` - if the node is an openvz guest
-* `virtual?` - if any of the above are true (guest of any detected virtualization system)
-* `physical?` - strictly the logical opposite of `virtual?`
+* `kvm?` - if the node is a kvm (guest or host)
+* `lxc?` - if the node is an lxc (guest or host)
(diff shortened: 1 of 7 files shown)


tas50:
We probably want to make sure we check that we’re on guests not just a system with VMware or KVM as a host

lamont-granquist:
heh, great.

this is just copypasta from chef-sugar so its just been broken then.

lamont-granquist:
so i left helpers like `kvm?` the way they were, so users will need to `if kvm? && guest?` or `if kvm? && virtual_host?` explicitly if they care.

tas50:
The KVM and VMware stuff should really get fixed as well. That just sorta worked back in the chef-suger days because ohai wasn't that great at detecting the virtualization hosts. Ohai detects the hosts and guests pretty well now so that's definitely not going to work in a consistent way or even the way people would have expected chef-sugar worked circa chef 13. It should all just be for guests so that someone with virtualbox or vmware fusion on their computer doesn't get flagged as a vm. This will pop up more and more as well do more desktop automation work.

lamont-granquist:
But then you've got no way to tell if someone has virtualbox or vmware fusion via a helper if those are only for guests.  Works great for guest detection, works bad for managing the hosts.

chef-utils/lib/chef-utils/dsl/virtualization.rb
```diff
@@ -0,0 +1,231 @@
...
+
+      alias_method :virtual?, :guest?
+
+      # Determine if the current node is running under any virutalization environment
+      #
+      # @param [Chef::Node] node
+      #
+      # @return [Boolean]
+      #
+      def virtual_host?(node = __getnode)
```
tas50: hypervisor?

chef-utils/README.md
```diff
@@ -133,6 +133,19 @@ Architecture Helpers allow you to determine the processor architecture of your n
 * `digital_ocean?` - if the node is running in digital ocean
 * `softlayer?` - if the node is running in softlayer
 
+### Virtualization Helpers
+
+* `kvm?` - if the node is a kvm (guest or host)
+* `lxc?` - if the node is an lxc (guest or host)
```
tas50: wanna add all the host checks here too?

chef-utils/lib/chef-utils/dsl/virtualization.rb
lamont-granquist: lxc is container based though and doesn't have a hypervisor so that word would be wrong, i think.

lamont-granquist: oh yeah, i see that could be read either way though so that is also terrible

lamont-granquist: i hate words

lamont-granquist: vmhost, hypervisor...

lamont-granquist: vmm or vmms which probably nobody would guess

tas50: I think hypervisor is less terrible than vm_host, but given a time machine container stuff would go in node['container'] not node['virtualization']

lamont-granquist: yeah i polled #general on slack and wound up at `hypervisor?` as being least-bad.

PR #9322, opened 2020-02-03
Document the new hidden field in mac_user ships in 15.8

tas50:
Make sure we expose this come docs


lamont-granquist:
:+1:

PR #9328, opened 2020-02-04
Add information about the resource_name in property exceptions

lamont-granquist:
🚧 WIP 🚧


lamont-granquist:
This is harder than I first thought.  I want it to work on the actual name property (not the name_property) so that if you do something like `user` where you've typo'd some local variable and picked up the resource instead, that it fails and the error message states that the name property is required for the user resource.  But the name property is defined on Chef::Resource so the decared_in variable only contains Chef::Resource as a class, which is totally useless.  This would need to be more extensive and inject the calling object into the property.

PR #9331, opened 2020-02-05
Add some version string backcompat APIs

lamont-granquist:
These are strictly necessary for backcompat with chef-sugar and
necessary for converting omnibus off of chef-sugar and onto
chef-utils.


chef-utils/lib/chef-utils/dsl/platform_version.rb
```diff
@@ -0,0 +1,39 @@
...
+module ChefUtils
+  module DSL
+    module PlatformVersion
+      include Internal
+
+      # Return the platform_version for the node. Acts like a String
+      # but also provides a mechanism for checking version constraints.
+      #
+      # @param [Chef::Node] node
+      #
```
tas50: Add a @since 15.8 here so we can get this documented out of the box in vscode

PR #9332, opened 2020-02-05
Pull the windows Ruby installer from S3 for tests

tas50:
This should prevent failures when we get throttled pulling it from
Github


btm:
Windows was good, we still need to fix the Centos failure on the TK tests:
```
dnf.exceptions.RepoError: Failed to synchronize cache for repo 'AppStream'
 ```

tas50:
Yeah I need to figure out what's up with that TK stuff now, but we're closer to green.

PR #9334, opened 2020-02-05
Add a default_description in windows_task

tas50:
This came up in https://github.com/chef/chef-web-docs/pull/2256 and we
should really be documenting that behavior here so we can generate the
docs with it.


tas50:
Backport of #9329

PR #9349, opened 2020-02-07
Add notify_group resource

### LINKED ISSUE #3426: Add a 'null' or 'notifies' NOP resource to hang chained notifies off of
Original request here:
https://tickets.opscode.com/browse/CHEF-5324

The item was obviously lost when the issues moved to Github.  Chef still has not provided a way to specify a resource call is to execute at the end of the Chef run.  The work-around was to have a dummy resource do the notify like as follows:

```
bash 'dummy-delay-httpd-start' do
  notifies :restart, 'service[httpd]', :delayed
end

service 'httpd' do
  action :nothing
end
```

However, now a warning is beind thrown that this won't work anymore in Chef 13:

```
 WARN: bash[dummy-delay-httpd-start]: No code attribute was given, resource does nothing, this behavior is deprecated and will be removed in Chef-13
```

You need to either abort the planned removal of this or add the original feature I requested.  It's absurd that there won't be a way to have something run at the end of a Chef run in all circumstances.

lamont-granquist:
This can be used to replace the abuse of the log resource for
grouping notifications.

Closes #3426

### DIFF
--- a/lib/chef/resource/notify_group.rb
+++ b/lib/chef/resource/notify_group.rb
@@ -1,0 +1,70 @@
+#
+# Copyright:: 2019-2020, Chef Software Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require_relative "../resource"
+require_relative "../dist"
+
+class Chef
+  class Resource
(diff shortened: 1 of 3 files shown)


lib/chef/resource/notify_group.rb
```diff
@@ -0,0 +1,70 @@
...
+
+class Chef
+  class Resource
+    class NotifyGroup < Chef::Resource
+      resource_name :notify_group
+      provides :notify_group
+
+      unified_mode true
+
+      description "The notify_group resource does nothing, and always fires notifications which are set on it.  Use it to DRY blocks of notifications that are common to multiple resources, and provide a single target for other resources to notify.  Unlike most resources, its default action is :nothing."
```
tas50:
```suggestion
      description "The notify_group resource does nothing, and always fires notifications which are set on it.  Use it to DRY blocks of notifications that are common to multiple resources, and provide a single target for other resources to notify. Unlike most resources, its default action is :nothing."
```

lamont-granquist: if i merge this it busts the DCObot though.

tas50: You can just paste a DCO sign off into the UI and it'll work

PR #9364, opened 2020-02-13
Add chef_vault_secret resource from chef-vault cookbook

tas50:
Move this common resource into Chef itself to avoid the need to include this cookbook.


lib/chef/resource/chef_vault_secret.rb
```diff
@@ -0,0 +1,134 @@
...
+      action :create_if_missing do
+        description "Calls the create action unless it exists."
+
+        action_create if current_resource.nil?
+      end
+
+      action :delete do
+        description "Deletes the item and the item's keys ('id'_keys)."
+
+        converge_by("remove #{new_resource.id} and #{new_resource.id}_keys from #{new_resource.data_bag}") do
```
lamont-granquist: delete doesn't look properly idempotent here

tas50: Totally isn't

lamont-granquist:
i assume that bug there is years old at this point, so wouldn't consider it a blocker.

PR #9370, opened 2020-02-14
windows_ad_join: Fix joining specific domains when domain trusts are involved

srb3:
Signed-off-by: srb3 <sbrown@chef.io>

## Description
A customer has an issue where they want the user who initiates the domain to join to be in a different domain to the one that is being joined to.

```
you cannot use the windows_ad_join resource to join a computer in for example domain-b, by a user in domain-a that has a trust between them
```

customer would like to do the following

```
windows_ad_join domain-b do
  domain_user     username + "@domain-a"
  domain_password domain_creds['password']
  action          :join
  notifies        :reboot_now, 'reboot[post-domain-join]', :immediately
end
```
however it ends up with something like this:

 ```username@domain-a@domain-b```

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.
- [ ]x All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/lib/chef/resource/windows_ad_join.rb
+++ b/lib/chef/resource/windows_ad_join.rb
@@ -70,7 +70,7 @@ class Chef
 
         unless on_desired_domain?
           cmd = "$pswd = ConvertTo-SecureString \'#{new_resource.domain_password}\' -AsPlainText -Force;"
-          cmd << "$credential = New-Object System.Management.Automation.PSCredential (\"#{new_resource.domain_user}@#{new_resource.domain_name}\",$pswd);"
+          cmd << "$credential = New-Object System.Management.Automation.PSCredential (\"#{sanitize_usename}\",$pswd);"
           cmd << "Add-Computer -DomainName #{new_resource.domain_name} -Credential $credential"
           cmd << " -OUPath \"#{new_resource.ou_path}\"" if new_resource.ou_path
           cmd << " -NewName \"#{new_resource.new_hostname}\"" if new_resource.new_hostname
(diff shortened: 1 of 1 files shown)


lib/chef/resource/windows_ad_join.rb
```diff
@@ -169,6 +169,20 @@ def on_desired_domain?
...
         end
 
+        #
+        # @return [String] the correct user and domain to use.
+        #   if the domain_user property contains an @ symbol followed by any number of non white space characheters
+        #   then we assume it is a user from another domain than the one specifed in the resource domain_name property.
+        #   if this is the case we do not append the domain_name property to the domain_user property
+        #   regex: https://rubular.com/r/VzupyBWemC8MMM
+        def sanitize_usename
+          if new_resource.domain_user =~ /^\S*@\S*$/
```
stevendanna: I don't know much about AD names, but the use of `*` here seems a bit fishy to me.  This maches:

`foo@`
`@`
`@foo`

But not:

`foo bar@baz`

I think it would also be good to link to the rules we think the domain_user needs to comply to. Do they allow spaces? If so, this regex will be incorrect I think.  Given the current regex, I think we would be better off with checking if the string contains '@' rather than using a regex because then the behavior would be easier to document and understand -- assuming that is compatible with whatever naming rules apply here.

srb3: Did some digging:
The UPN seems to be based off RFC 822 (which is now obsolete)
 https://docs.microsoft.com/en-us/windows/win32/ad/naming-properties#userprincipalname

https://tools.ietf.org/html/rfc822#section-6
is the address part of the rfc that UPN is based off.
However, It is not clear (to me) if white spaces are allowed by this spec
So I agree that matching on the @ is preferable to the current regex, given its simplicity in an unclear area.

PR #9380, opened 2020-02-20
Require compatible cheffish and chef-zero releases

tas50:
We should require releases that work with Ruby 2.6/2.7.


lamont-granquist:
yeah i think i'd rather leave the lower bounds where they are, there's nothing actually incompatible there.

PR #9387, opened 2020-02-21
fix ruby 2.7 URI.unescape deprecation

lamont-granquist:
Necessary for ruby 2.7 support

### DIFF
--- a/lib/chef/http/http_request.rb
+++ b/lib/chef/http/http_request.rb
@@ -176,8 +176,8 @@ class Chef
         @http_request.body = request_body if request_body && @http_request.request_body_permitted?
         # Optionally handle HTTP Basic Authentication
         if url.user
-          user = URI.unescape(url.user)
-          password = URI.unescape(url.password) if url.password
+          user = CGI.unescape(url.user)
+          password = CGI.unescape(url.password) if url.password
           @http_request.basic_auth(user, password)
         end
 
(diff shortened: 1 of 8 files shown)


lib/chef/http/http_request.rb
```diff
@@ -176,8 +176,8 @@ def configure_http_request(request_body = nil)
         @http_request.body = request_body if request_body && @http_request.request_body_permitted?
         # Optionally handle HTTP Basic Authentication
         if url.user
-          user = URI.unescape(url.user)
-          password = URI.unescape(url.password) if url.password
+          user = CGI.unescape(url.user)
```
tas50: do we need to require this?

lamont-granquist:
🚧  need to see if specs still pass or not

lib/chef/http/http_request.rb
lamont-granquist: thought it was another built in...

lamont-granquist: oh no i guess it isn't

lib/chef/resource/apt_repository.rb
```diff
@@ -378,7 +379,7 @@ def make_ppa_url(ppa)
         def build_repo(uri, distribution, components, trusted, arch, add_src = false)
           uri = make_ppa_url(uri) if is_ppa_url?(uri)
 
-          uri = URI.escape(uri)
+          uri = CGI.escape(uri)
```
lamont-granquist: this one is problematic because its ripping the escaping across the entire uri, the other ones are specifically targeting the user/password/path components.

lamont-granquist: fixed with Addressable

PR #9400, opened 2020-02-26
Convert the node[:platform_version] to a Chef::VersionString

### LINKED ISSUE #9154: Wrap node['platform_version'] in a Chef::VersionString
Inflate it into a Chef::VersionString after its read in from ohai and that will allow:

`node['platform_version'] > 6.10` to just work correctly without the inherent bugginess of `to_f`.

backport it to 14/15, then consider raising on `to_i/to_f` in Chef::VersionString and force people off of those poor APIs.

for bonus points get Chef::VersionString into chef-utils and use it in ohai so that you can do the same there.

lamont-granquist:
`node[:platform_version] =~ "~> 1.2"` will now just work.

This might conceivably be breaking if people are doing really weird
things by the class changing (Chef::VersionString still inherits from
String though, so it'll have to be weirdness on par with the bad
`thing.class.to_s == "Mash"` kind of comparisons that were in ohai).

closes #9154

### DIFF
--- a/chef-utils/lib/chef-utils/version_string.rb
+++ b/chef-utils/lib/chef-utils/version_string.rb
@@ -27,8 +27,13 @@ module ChefUtils
     #
     # @param val [String] Version string to parse.
     def initialize(val)
-      super
-      @parsed_version = ::Gem::Version.create(self)
+      val = "" unless val
+      super(val)
+      begin
+        @parsed_version = ::Gem::Version.create(self)
+      rescue ArgumentError
+        @parsed_version = nil
+      end
     end
 
     # @!group Compat wrappers for String
@@ -135,7 +140,12 @@ module ChefUtils
       when Regexp
         super
       else
-        Gem::Requirement.create(other) =~ parsed_version
+        begin
+          Gem::Requirement.create(other) =~ parsed_version
+        rescue ArgumentError
+          # one side of the comparison wasn't parseable
+          super
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


tas50:
How will this work on some of the odder versioned platforms:

arch: "4.10.13-1-ARCH"
freebsd: "12.1-RELEASE-p1"
gentoo: "4.9.95-gentoo"

lamont-granquist:
Works as well as it can.  Those will all work for the leading decimals, then it tends to go off he rails:

```
[8] pry(main)> Gem::Version.new("4.10.13-1-ARCH").segments
=> [4, 10, 13, "pre", 1, "pre", "ARCH"]
[9] pry(main)> Gem::Version.new("4.9.95-gentoo").segments
=> [4, 9, 95, "pre", "gentoo"]
[10] pry(main)> Gem::Version.new("12.1-RELEASE-p1").segments
=> [12, 1, "pre", "RELEASE", "pre", "p", 1]
```

The well over 99% use case is that nobody cares about the trailing nonsense.

Since these are still strings then totally insane things will fall back to behaving exactly like a string and nothing better can be done there.  The code added to Chef::VersionString handles a lot of strange cases in the node attribute tests where we're pushing in totally unparseable strings into the platform_version and it needed to be protected against throwing exceptions.

There's no coherent way to fix the problem accurately beyond this, without writing a library of routines that handle looking up the proper specific comparison operation based on platform which will never get written.  This is as good as it gets.

PR #9409, opened 2020-02-27
Update the rhsm_erata* and rhsm_register resources for RHEL 8

tas50:
These need to use dnf instead of yum_package where they can. I also
enabled these for unified_mode and moved some only_if logic outside the
resources for cleaner logging. These should really all just get
converted to use shell_out and a nice converge_by

### DIFF
--- a/lib/chef/resource/rhsm_register.rb
+++ b/lib/chef/resource/rhsm_register.rb
@@ -67,6 +67,12 @@ class Chef
         package "subscription-manager"
 
         unless new_resource.satellite_host.nil? || registered_with_rhsm?
+          declare_resource(package_resource, "katello-ca-consumer-latest") do
+            options "--nogpgcheck"
+            source "#{Chef::Config[:file_cache_path]}/katello-package.rpm"
+            action :nothing
+          end
+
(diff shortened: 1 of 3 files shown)


lib/chef/resource/rhsm_register.rb
```diff
@@ -69,11 +70,13 @@ class RhsmRegister < Chef::Resource
           remote_file "#{Chef::Config[:file_cache_path]}/katello-package.rpm" do
             source "http://#{new_resource.satellite_host}/pub/katello-ca-consumer-latest.noarch.rpm"
             action :create
-            notifies :install, "yum_package[katello-ca-consumer-latest]", :immediately
+            notifies :install, "{rhel8? ? 'dnf' : 'yum'}_package[katello-ca-consumer-latest]", :immediately
```
jaymzh: should this be a helper? We do it a lot...

Also, I worry about rhel9 breaking. I feel like this should be something like `(rhel? && platform_version.to_i < 8) ? 'yum' : 'dnf'`

tas50: That's fair. Let me rework that.

PR #9417, opened 2020-02-28
Remove all the code that checks for Windows Nano

tas50:
Windows Nano isn't a thing anymore so we shouldn't be spending compute time checking to see if we're on Windows Nano


lib/chef/platform/query_helpers.rb
```diff
@@ -26,27 +26,9 @@ def windows?
...
-          ::Win32::Registry::HKEY_LOCAL_MACHINE.open(key, access) do |reg|
-            nano = reg["NanoServer"]
-          end
-        rescue ::Win32::Registry::Error
-          # If accessing the registry key failed, then we're probably not on
-          # nano. Fail through.
-        end
-        nano == 1
+        false
       end
```
lamont-granquist: should probably just delete this and let it fail to force cleanup of anyone who found it and used it.

tas50: In the future yeah we should blow up here. We can cleanup any uses first.

PR #9427, opened 2020-02-29
Remove the last bits of Appveyor from the specs

tas50:
We don't have Appveyor anymore.


jaymzh:
Windows tests, specifically, are failing. I didn't go digging into them, but potentially related?

tas50:
Yep this failed which means we need to update some tests

PR #9433, opened 2020-03-03
Use the action DSL consistently

lamont-granquist:
May be a potentially breaking change if I noodle hard enough on it, it
does mean that all the actions are now (correctly) encapsulated in a
sub-resource collection, whereas before they never had one.


lib/chef/provider/route.rb
```diff
@@ -162,8 +162,10 @@ def action_delete
       def generate_config
         if platform_family?("rhel", "amazon", "fedora")
```
lamont-granquist: tried fixing this to use the new helpers, it blew up unit tests, already burned too much time on this

tas50:
Not a huge fan of super(), but I can live.

lamont-granquist:
yeah, dynamically defined methods can't call super without declaring the args.

Issue #9359, opened 2020-02-10
Remove highline gem


reedlaw:
### Describe the Enhancement:
Highline dependency is too restrictive.

### Describe the Need:
It prevents other gems from being updated.

### Current Alternative
#8381 already supplied the alternative

### Can We Help You Implement This?:
From searching the code it doesn't seem like highline is being used at all. It could be as simple as removing the gem and unused methods.

tas50:
@reedlaw There's a few other places we use highline that need to get updated to use the tty gems. #8381 was the first part of this work, but it was also the easy part.

lamont-granquist:
./lib/chef/knife/core/ui.rb and ./lib/chef/formatters/indentable_output_stream.rb still use it.

btm:
`highline.ask` should be pretty easily replaceable by `tty-prompt`.

I can't figure out a way to get `tty-table` to look like `highline.list`
```
2.7.0-preview1 :008 > HighLine.new.list(['one', 'two', '12' ], :columns_down)
 => "one  two  12 \n" 
2.7.0-preview1 :009 > HighLine.new.list(['one', 'two', '123' ], :columns_down)
 => "one  two  123\n" 
2.7.0-preview1 :010 > HighLine.new.list(['one', 'two', '1234' ], :columns_down)
 => "one   two   1234\n" 
2.7.0-preview1 :011 > HighLine.new.list(['one', 'two', '12345' ], :columns_down)
 => "one    two    12345\n" 
```

rmoriz:
There was a braking change introduced by switching to pastel:

```ruby
irb(main):015:0> highline.color("test", [:bold, :white])
=> "\e[1m\e[37mtest\e[0m"

irb(main):016:0> pastel.decorate("test", [:bold, :white])
Traceback (most recent call last):
       16: from /Users/rmoriz/.asdf/installs/ruby/2.7.1/bin/irb:23:in `<main>'
       15: from /Users/rmoriz/.asdf/installs/ruby/2.7.1/bin/irb:23:in `load'
       14: from /Users/rmoriz/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/irb-1.2.3/exe/irb:11:in `<top (required)>'
       13: from (irb):16
       12: from /Users/rmoriz/.asdf/installs/ruby/2.7.1/lib/ruby/2.7.0/forwardable.rb:235:in `decorate'
[...]

PR #9434, opened 2020-03-03
Replace highline.color with pastel.decorate

btm:
Contributes to the removal of the 'heavy' highline dependency (#9359) by
switching from highline.color to pastel.decorate, which is a drop-in
replacement. Pastel is part of the tty[1] family of libraries.

[1] https://github.com/piotrmurach/tty#3-components

### DIFF
--- a/lib/chef/formatters/indentable_output_stream.rb
+++ b/lib/chef/formatters/indentable_output_stream.rb
@@ -17,21 +17,12 @@ class Chef
         @semaphore = Mutex.new
       end
 
-      def highline
-        @highline ||= begin
-          require "highline"
-          HighLine.new
+      # pastel.decorate is a lightweight replacement for highline.color
+      def pastel
+        @pastel ||= begin
+          require "pastel"
+          Pastel.new
         end
-      end
-
-      # Print text. This will start a new line and indent if necessary
-      # but will not terminate the line (future print and puts statements
-      # will start off where this print left off).
-      #
-      # @param string [String]
-      # @param args [Array<Hash,Symbol>]
-      def color(string, *args)
-        print(string, from_args(args))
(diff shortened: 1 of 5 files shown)


lib/chef/formatters/indentable_output_stream.rb
```diff
@@ -17,23 +17,14 @@ def initialize(out, err)
...
         end
       end
 
-      # Print text. This will start a new line and indent if necessary
-      # but will not terminate the line (future print and puts statements
-      # will start off where this print left off).
-      #
-      # @param string [String]
-      # @param args [Array<Hash,Symbol>]
-      def color(string, *args)
```
btm: This appears to be cruft and not called from anywhere.

btm:
Just force-pushed the same fix for `lib/chef/formatters/indentable_output_stream.rb`. Didn't add the same unit test there. seems unnecessary and I mostly added the other unit test to be sure we actually called `pastel.decorate` somewhere in the tests.

I also manually tested a few commands like `knife node show NODE` and `knife config list profiles` to verify that color was working correctly.

btm:
bk failure is unrelated windows functional spec certificate failures.

lib/chef/formatters/indentable_output_stream.rb
tas50: do any plugins in the community use this?

btm: I can't imagine a reasonable reason for a plugin poking around in the internals of IndentableOutputStream. If this was in Core::UI or somewhere I would have left it.

tas50: 👍

PR #9441, opened 2020-03-04
Remove the canonical DSL

### LINKED ISSUE #9206: Remove provides, resource_name and canonical dsl complexity and confusion
Get rid of provides, probably get rid of the "canonical" DSL (have to research what it is), and do something like have the first provides name set the resource_name and/or allow a `resource_name: true` flag to set the name of the resource.

lamont-granquist:
Removes all the internal wiring of the "canonical DSL" in the ResourceResolver and the node_map.

The `resource_name` still has magical wiring to an automatic implicit `provides` of the given resource_name.  In auditing supermarket removing this magic would cause 22% of all otherwise correctly written custom resources to break because they only declare a resource_name with no provides.

The automatic wiring of the class name to the DSL and to the resource_name is removed.  This does not affect LWRPs or custom resources at all.  The default cookbook_name+file_name wiring of LWRPs+Custom Resources has been preserved.

The first provides line now wires up the resource_name if it has not been set yet.  I would highly encourage this to become the default way to write resources, because provides-with-automatic-resource_name still gets the any `os` or `platform_family` filters correct.   The alternative of resource_name-with-automatic-provides should IMO be discouraged since it leads to this kind of thing:

```
resource_name :windows_thingy
provides :windows_thingy, os: :windows
```

The provides line there is useless because due to the implicit provides this becomes:

```
resource_name :windows_thingy
provides :windows_thingy
provides :windows_thingy, os: :windows
```

Because of this I would argue that the correct thing to do would be to remove the implicit provides from the resource_name (but again this breaks 22% of resources on the supermarket today).

The alternative would be to embrace the `resource_name` as the thing to set everywhere and call that correct and establish that "yes the thing you call the resource should always be applied to the DSL` but that will mean that the correct way to write the above resource would simply be:

```
resource_name :windows_thingy  # implicitly calls provides :windows_thingy
```

Which then establishes that will commit to losing information about the os and platform_family applicability to most resources.  This is beneficial for cases like chefspec, which still does not handle dynamic resource resolution correctly, but will definitively break any hope of automatically generating per-OS self-documentation of which resources apply to which O/S.  I'd argue that is wrong and we should fix chefspec instead (I've waffled horribly about this idea in the past, I think I'm settling down to fixing chefspec as the correct direction).

At any rate -- we still support both bits of magic -- all the previous discussion is about the next step we should take.

closes #9206

### DIFF
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Lamont Granquist (<lamont@chef.io>)
-# Copyright:: Copyright 2014-2019, Chef Software Inc.
+# Copyright:: Copyright 2014-2020, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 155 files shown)


lamont-granquist:
well that's rather red and angry

tas50:
I'm a firm supporter of the magic in resource_name mostly because 1) it's out there already and walking that back would be unnecessarily painful for users 2) it's a logical name for what it does. We do have a docs gap here where we're not properly explaining how to use provides in a scenario where you wouldn't want to also use resource_name. We should get that into the docs.

PR #9445, opened 2020-03-04
Update all our links to use the new docs site format / cleanup descriptions

tas50:
When we migrated to hugo the URLs changed a bit. Nothing ends in .html and we moved all the resources into their own dir.

### DIFF
--- a/lib/chef/dsl/platform_introspection.rb
+++ b/lib/chef/dsl/platform_introspection.rb
@@ -252,7 +252,7 @@ class Chef
       # @deprecated Windows releases before Windows 2012 and 8 are no longer supported
       # @return [Boolean] Is the system older than Windows 8 / 2012
       def older_than_win_2012_or_8?(node = run_context.nil? ? nil : run_context.node)
-        node["platform_version"].to_f < 6.2
+        false # we don't support platforms that would be true
       end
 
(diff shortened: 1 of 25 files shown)


mjingle:
@tas50 If you're seeing italics in the release notes in wide bands of release notes, that's due to phrases like "*nix", which need to be formatted with appropriate `formatting`. See https://github.com/chef/chef/pull/9445/files#diff-f54e6ba45d43eee7cdc25838b79b012cR270

tas50:
@mjingle at some point I'll go through all the auto generated stuff and make sure to remove those.

lib/chef/dsl/platform_introspection.rb
```diff
@@ -252,7 +252,7 @@ def platform_family?(*args)
       # @deprecated Windows releases before Windows 2012 and 8 are no longer supported
       # @return [Boolean] Is the system older than Windows 8 / 2012
       def older_than_win_2012_or_8?(node = run_context.nil? ? nil : run_context.node)
-        node["platform_version"].to_f < 6.2
+        false # we don't support platforms that would be true
       end
```
phiggins: This seems like an unrelated change?

tas50: Yep that was a bit of cleanup from a previous PR

PR #9462, opened 2020-03-07
Turn off a lot of noise in test runs.

zenspider:
+ I don't know how ---color was ever parsed to begin with.
+ -fd results in >23k lines of output that nobody reads.
  + This output obscures ~800 lines of output that everyone should
    read and silence.
+ Turned off verbose in rspec rake task. It's huge and doesn't help anything.

Signed-off-by: Ryan Davis <zenspider@chef.io>


phiggins:
I'd be ok with merging this piecemeal while working on the details. Things like turning off `-fd` are 👍 for me.

tas50:
We merged in a good chunk of this which resulted in some nice noise reductions. Thanks @zenspider

PR #9466, opened 2020-03-07
Set env vars in the container setup script

tas50:
We're setting the same env vars in each job. Seems entirely pointless.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -26,10 +26,6 @@ steps:
       docker:
         image: rubydistros/ubuntu-18.04
         privileged: true
-        environment:
-          - FORCE_FFI_YAJL=ext
-          - CHEF_LICENSE=accept-no-persist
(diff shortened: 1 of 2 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -291,10 +246,6 @@ steps:
     executor:
       docker:
         image: ruby:2.7-buster
-        environment:
-          - FORCE_FFI_YAJL=ext
-          - CHEF_LICENSE=accept-no-persist
-          - UNIT_SPECS_27=1
```
phiggins: Makes sense overall but does this mean `UNIT_SPECS_27=1` and friends aren't doing anything?

tas50: I'm fairly certain that's just a leftover from travis where we used the env vars to name the jobs.

phiggins: I grepped the project and could only find references to them in this file, so this seems 👍

PR #9478, opened 2020-03-11
Bring in the extended Ruby cleanup used in chef-workstation

tas50:
This resulted in a pretty nice install time win / disk space win in Workstation. Let's see if we can achieve something similar in chef/chef.

This reduces the Ubuntu package size by 14%. I'll have additional install stats once our mac builder issues are resolved

### DIFF
--- a/omnibus/config/software/more-ruby-cleanup.rb
+++ b/omnibus/config/software/more-ruby-cleanup.rb
@@ -1,0 +1,103 @@
+#
+# Copyright:: 2019-2020 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


omnibus/config/software/more-ruby-cleanup.rb
```diff
@@ -0,0 +1,103 @@
...
+      # chef stores the powershell dlls in the ext dir
+      next if File.basename(File.expand_path("..", f)).start_with?("chef-")
+
+      puts "Deleting #{f}"
+      if File.directory?(f)
+        # recursively removes files and the dir
+        FileUtils.remove_dir(f)
+      else
+        File.delete(f)
+      end
```
phiggins: I'm not sure if the more verbose mode is necessary, but FileUtils has a `rm_r` function for deleting directories recursively: https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_r

PR #9479, opened 2020-03-11
Fix load path in test runs.

phiggins:
Fixes the load path in rspec runs to always load the local version of sub-gems.

## Description

* Add the local versions of `chef-config` and `chef-utils` to the load path so that rspec will always load the local versions when called without bundler.
* Use `$LOAD_PATH` instead of the cryptic `$:`.
* Remove some unnecessary additions to the load path that are automatically done by rspec.

## Types of changes

## Checklist:

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,17 +24,15 @@ module Shell
   IRB = nil unless defined? IRB
 end
 
-# Ruby 1.9 Compat
-$:.unshift File.expand_path("../..", __FILE__)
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -25,17 +25,16 @@ module Shell
 end
 
 # Ruby 1.9 Compat
```
tas50: we can probably nuke this comment at this point

phiggins: Nuked 💣

PR #9480, opened 2020-03-12
Fix unintuitive behavior of sources on gem resources

phiggins:
## Description
Fix for #6729 

The main motivation for this PR is to make the [airgapped Chef installation described in the official Chef docs](https://docs.chef.io/install_chef_air_gap/) work without having to do anything extra.

* Change the default value of the `clear_sources` property of `gem_package` and `chef_gem` resources to `nil`. The possible behaviors for `clear_sources` are now:
** `true`: Always clear sources.
** `false`: Never clear sources.
** `nil`: Clear sources if `source` is set, don't clear sources otherwise.
* Change the default value of the `include_default_source` property of `gem_package` and `chef_gem` resources to `nil`. The possible behaviors for `include_default_source` are now:
** `true`: Always include the default source.
** `false`: Never include the default source.
** `nil`: Include the default source if `Chef::Config[:rubygems_url]` is set or if `source` and `clear_sources` are not set on the resource.
* Change the default value of `Chef::Config[:rubygems_url]` to `nil`. Setting to `nil` previously  had similar behavior to setting `clear_sources` to `true`, but with some differences. The new behavior is to always use "https://rubygems.org" as the default rubygems repo unless explicitly changed, and whether to use this value is determined by `clear_sources` and `include_default_source`.
* Change the rubygems URL from "https://www.rubygems.org" to the canonical "https://rubygems.org".

## Related Issue
Fix for #6729

## Types of changes

## Checklist:


lib/chef/provider/package/rubygems.rb
```diff
@@ -555,7 +555,7 @@ def install_via_gem_command(name, version)
           if new_resource.source.is_a?(String) && new_resource.source =~ /\.gem$/i
             name = new_resource.source
           else
-            src << "--clear-sources" if new_resource.clear_sources
+            src << "--clear-sources" if new_resource.clear_sources?
```
tas50: was this meant to be changed?

phiggins: Yes, that's changing from checking the property value to calling this method https://github.com/chef/chef/pull/9480/files#diff-151fa83ee17ddfc007da9fe30edc0f91R59-R65

tas50: it should be just clear_sources? then.

tas50: Also if this is a method used in the provider you'll want to toss it in there

lib/chef/resource/gem_package.rb
```diff
@@ -55,6 +55,14 @@ class GemPackage < Chef::Resource::Package
       property :options, [ String, Hash, Array, nil ],
         description: "Options for the gem install, either a Hash or a String. When a hash is given, the options are passed to Gem::DependencyInstaller.new, and the gem will be installed via the gems API. When a String is given, the gem will be installed by shelling out to the gem command. Using a Hash of options with an explicit gem_binary will result in undefined behavior.",
         desired_state: false
+
+      def clear_sources?
+        if clear_sources.nil?
+          !!source
```
lamont-granquist: this should really get moved to the provider since its not interface and is logic.

should document the magic here as well with a comment.

chef-config/lib/chef-config/config.rb
```diff
@@ -1221,9 +1221,6 @@ def self.guess_internal_locale
     # can be set to a string or array of strings for URIs to set as rubygems sources
     default :rubygems_url, "https://www.rubygems.org"
 
-    # globally sets the default of the clear_sources property on the gem_package and chef_gem resources
-    default :clear_gem_sources, false
```
lamont-granquist: this should probably be changed to an explicit nil default rather than relying on chef-config autovivification (for documentation purposes if nothing else, but i think we actually consume chef-config and turn on strict checking for something like berkshelf as well -- we always intended to turn on strict checking in chef-client but never did)

phiggins: I wondered about that, I agree it makes sense to be more explicit.

lib/chef/resource/gem_package.rb
phiggins: If by document the magic you mean to leave a comment explaining what this is doing, I had intended to do that after I got confirmation that this is what we wanted 👍

tas50:
@phiggins Thanks for taking the time to sit down and really think through the scenarios we're trying to support and what it takes to make this work. I think this is going to help a lot of folks out.

PR #9481, opened 2020-03-12
Add alternatives resource for Linux

tas50:
This comes from https://github.com/vkhatri/chef-alternatives

### DIFF
--- a/lib/chef/resource/alternatives.rb
+++ b/lib/chef/resource/alternatives.rb
@@ -126,7 +126,7 @@ class Chef
         def current_path
           # https://rubular.com/r/ylsuvzUtquRPqc
           match = shell_out(alternatives_cmd, "--display", new_resource.link_name).stdout.match(/link currently points to (.*)/)
-          match.nil? ? nil : match[1]
+          match[1]
         end
 
         #
--- a/lib/chef/resource/alternatives.rb
+++ b/lib/chef/resource/alternatives.rb
@@ -102,10 +102,10 @@ class Chef
         # @return [String] The appropriate alternatives command based on the platform
         #
         def alternatives_cmd
-          if fedora_derived?
-            "alternatives"
-          else
+          if debian?
             "update-alternatives"
+          else
+            "alternatives"
           end
         end
 
--- a/lib/chef/resource/alternatives.rb
+++ b/lib/chef/resource/alternatives.rb
@@ -35,12 +35,20 @@ class Chef
       property :priority, [String, Integer], coerce: proc { |n| n.to_i }
 
       def define_resource_requirements
+        requirements.assert(:install) do |a|
+          a.assertion do
+            !new_resource.priority.nil?
+          end
+
+          a.failure_message("Could not set alternatives for #{new_resource.link_name}, you must provide the :priority property")
+        end
+
         requirements.assert(:install, :set, :remove) do |a|
           a.assertion do
             !new_resource.path.nil?
           end
 
-          a.failure_message("Could not set alternatives for #{new_resource.link_name}, must provide :path property")
+          a.failure_message("Could not set alternatives for #{new_resource.link_name}, you must provide the :path property")
         end
 
         requirements.assert(:install, :set, :remove) do |a|
@@ -54,8 +62,6 @@ class Chef
       end
 
       action :install do
-        raise "missing :priority" unless new_resource.priority
-
         if path_priority != new_resource.priority
           converge_by("adding alternative #{new_resource.link} #{new_resource.link_name} #{new_resource.path} #{new_resource.priority}") do
             output = shell_out(alternatives_cmd, "--install", new_resource.link, new_resource.link_name, new_resource.path, new_resource.priority)
--- a/lib/chef/resource/alternatives.rb
+++ b/lib/chef/resource/alternatives.rb
@@ -1,0 +1,131 @@
+#
+# Copyright:: 2020, Chef Software Inc.
+# Copyright:: 2016-2020, Virender Khatri
+#
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      action :refresh do
+        converge_by("refreshing alternative #{new_resource.link_name}") do
+          shell_out("#{alternatives_cmd} --refresh #{new_resource.link_name}")
+        end
+      end
+
+      action_class do
+        def alternatives_cmd
+          if platform_family?("rhel", "amazon", "fedora")
```
lamont-granquist: `if fedora_based?`

lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,131 @@
...
+          a.failure_message("Could not set alternatives for #{new_resource.link_name}, missing #{new_resource.path}")
+        end
+      end
+
+      action :install do
+        raise "missing :priority" unless new_resource.priority
+
+        if path_priority != new_resource.priority
+          converge_by("adding alternative #{new_resource.link} #{new_resource.link_name} #{new_resource.path} #{new_resource.priority}") do
+            output = shell_out("#{alternatives_cmd} --install #{new_resource.link} #{new_resource.link_name} #{new_resource.path} #{new_resource.priority}")
```
lamont-granquist: all the `shell_out` commands should use the splat-args version.  should probably have a `run_alternatives(*args)` helper that does `shell_out(alternatives_cmd, *args)` as well.

lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      provides(:alternatives) { true }
+
+      description "The alternatives resource allows for configuration of command alternatives in Linux using the alternatives or update-alternatives packages."
+      introduced "16.0"
+
+      property :link_name, String, name_property: true
+      property :link, String, default: lazy { |n| "/usr/bin/#{n.link_name}" }
+      property :path, String
+      property :priority, [String, Integer], coerce: proc { |n| n.to_i }
```
phiggins: If this is getting coerced with `to_i`, can it ever be a String?

lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,131 @@
...
+            ::File.exist?(new_resource.path)
+          end
+
+          a.whyrun("Assuming file #{new_resource.path} already exists or was created already")
+          a.failure_message("Could not set alternatives for #{new_resource.link_name}, missing #{new_resource.path}")
+        end
+      end
+
+      action :install do
+        raise "missing :priority" unless new_resource.priority
```
phiggins: Should this be done above with the requirements?

lib/chef/resource/alternatives.rb
tas50: The user can give it to us as a string and then we just make it an Integer immediately.

lib/chef/resource/alternatives.rb
tas50: Updated to ```fedora_derived?```

lib/chef/resource/alternatives.rb
phiggins: I must have been reading the docs wrong, I thought it happened in the opposite order.

lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,144 @@
...
+        end
+
+        #
+        # @return [String] The current path for the link_name alternative
+        #
+        def current_path
+          # https://rubular.com/r/ylsuvzUtquRPqc
+          match = shell_out("#{alternatives_cmd} --display #{new_resource.link_name}").stdout.match(/link currently points to (.*)/)
+          match.nil? ? nil : match[1]
+        end
```
phiggins: `String#[]` can take a regex and a position to return just that match:

```
irb(main):006:0> "foo"[/(f)oo/, 1]
=> "f"
irb(main):007:0> "flood"[/(f)oo/, 1]
=> nil
```

lib/chef/resource/alternatives.rb
tas50: I used the splat form, but kept away from a helper since this makes it easy to read exactly what we're shelling out to.

lib/chef/resource/alternatives.rb
tas50: Yeah I was wondering the same thing about this one.

lib/chef/resource/alternatives.rb
```diff
@@ -0,0 +1,143 @@
...
+
+        #
+        # @return [Boolean] does the path exist for the link_name alternative
+        #
+        def path_exists?
+          # https://rubular.com/r/ogvDdq8h2IKRff
+          escaped_path = Regexp.new(Regexp.escape("#{new_resource.path} - priority"))
+          shell_out(alternatives_cmd, "--display", new_resource.link_name).stdout.match?(escaped_path)
+        end
+      end
```
phiggins: If it's not too much extra work, it seems worth it to me to add unit tests for these three methods using the input in the rubular link. Including the rubular link is a good idea though!

lib/chef/resource/alternatives.rb
lamont-granquist: maybe?  it isn't that important because there's no useful why-run case.  i have all kinds of internal debate over that one.  i did think about making the comment that it should be moved though, but in this case it felt a bit nitpicky since i can't think of anything bad that would ever happen...

tas50: Seemed a lot simpler to keep this here.

lib/chef/resource/alternatives.rb
tas50: Yeah let me see about adding those.

lib/chef/resource/alternatives.rb
phiggins: It seemed to make sense to want to fail as early as possible, at definition time vs run time, but I should definitely defer to yall on this one.

PR #9484, opened 2020-03-13
Turn off notifications from log resource by default

### LINKED ISSUE #1727: CHEF-4028:  log resource should not report as updated but should fire notifications
The proposed patch for CHEF-4028 had to be reverted because people expect this to work:

``` ruby
execute "bar" do
  command 'echo "Bar"'
  action :nothing
end

log "foo" do
  notifies :run, "execute[bar]", :immediately
end
```

The proposed CHEF-4028 patch also still updated the log resource when it logged, which was explicitly not what the user wanted.

One problem is that people are abusing a side effect of the log resource in order to group notifications.    We could construct a 'null' resource instead that did nothing but was always updated (unless action :nothing'd) that people could hang their notifications off of.

It has been argued that the CHEF-4028 patch is correct and should be the way the log resource works, but this means that the log resource will fire and be updated if and only if it actually logs in the log level.  That means that the resource updated count and any notifications you hang off of it would differ based on the log level that chef-client was invoked as.   That would make hanging notifications off of the log resource largely crazy which is the driver for the null resource.  This also does not solve the stated problem in CHEF-4028 which is that the log resource must log, yet not be updated.

Ways I can think of to implement this:
1.  The log resource could never be updated and never fire notifications, and only log (this would be a breaking change and would require implementing the null resource)
2.  The log resource could set a flag that resource reporting and resource counting at the end of the run would use to ignore the updated status of the resource (marking itself as blacklisted for reporting).
3.  The Chef::Runner and Chef::Resource could be modified so that the log resource could tell them to fire notifications even though it didn't mark itself as updated (the more i think about it the worse of an idea this seems).

All of these break the resource model in one way or another.

We should arguably implement the 'null' resource in any case, and push folks towards using that instead of abusing the log resource.  That may be somewhat orthogonal.

lamont-granquist:
This should be switched to the notify_group resource now.

A word on flipping the default, and why its the right thing to do:

- This is what people naively expect from the start and they have to dig to find it.

- IF this breaks any community cookbooks THEN anyone who flips the default will be unable to consume those community cookbooks, and will now be deeply lost trying to debug those cookbooks and why they appear to be broken.

- The switch is therefore a poor solution.

- As a result, we should flip the default, and fix the cookbooks, get to the point where everything just works the way people expect it to.

- The setting can be retained so that anyone upgrading from Chef-15 can flip that to get the behavior back before auditing their cookbooks -- the solution of using notify_group instead will of course be backcompat with that setting.

closes #1727


spec/integration/recipes/notifies_spec.rb
```diff
@@ -135,8 +140,9 @@
           resource_name :notifying_test
 
           action :run do
-            log "bar" do
+            notify_group "bar" do
               notifies :write, 'log[foo]', :delayed
+              action :run
```
jaymzh: If the default is to only use these as a notifier, why are you setting an action on them?

lamont-granquist: `action :nothing` doesn't fire notifies, and that's hard coded in Chef::Resource and difficult to override.

tas50:
Delete the leftover provider that's floating around

spec/integration/recipes/notifies_spec.rb
lamont-granquist: eh, no, that was before coffee.

in those specs i'm abusing it deliberately just to test notifications, that isn't an example of how to actually use it.

jaymzh: got it.

Issue #7319, opened 2018-05-30
Port ulimit resource to core


tas50:
We should be able to manage ulimits in Chef out of the box

tas50:
This was done

PR #9487, opened 2020-03-13
Add user_ulimit resource from the ulimit cookbook

tas50:
This is a really handy resource for modifying ulimits on a system.


tas50:
Fixes https://github.com/chef/chef/issues/7319

PR #9505, opened 2020-03-16
backport chef-utils comment fixes

tas50:
Fix some tests and backport some yard comments we use for generating vscode plugin data.

### DIFF
--- a/chef-utils/lib/chef-utils/dsl/windows.rb
+++ b/chef-utils/lib/chef-utils/dsl/windows.rb
@@ -26,7 +26,8 @@ module ChefUtils
 
       # Determine if the current node is Windows Server Core.
       #
-      # @param [Chef::Node] node
+      # @param [Chef::Node] node the node to check
+      # @since 15.7
       #
       # @return [Boolean]
       #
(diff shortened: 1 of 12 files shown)


chef-utils/lib/chef-utils/dsl/windows.rb
```diff
@@ -56,17 +59,19 @@ def windows_server?(node = __getnode)
...
       #
       def windows_nt_version(node = __getnode)
         ChefUtils::VersionString.new(node["os_version"])
       end
 
-      # Determine the installed version of PowerShell
+      # Determine the installed version of PowerShell.
       #
-      # @param [Chef::Node] node
+      # @param [Chef::Node] node the node to check
```
phiggins: The text `node the node to check` appears in this PR 73 times. It should be search-and-replaced to `the node to check`, probably?

tas50: The first part is the param name. It's the format the vscode yard plugin and a few yard example pages I found use. Seems really strange, but 🤷‍♂

PR #9507, opened 2020-03-18
Add support for parsing YAML recipes with chef-apply

btm:
Enables `chef-apply foo.yml` and `chef-apply foo.whatever --yaml`.

Improves error messages around incorrectly formatted YAML.

 **Also changes/enforces the YAML file format** 
This removes an unnecessary array that we had been using.

New format:
```
---
resources:
  - name: "do something"
    type: "execute"
```

```
{ "resources" => [ { "name" => "do something", "type" => "execute" } ] }
```
Old format:
```
---
- resources:
  - name: "do something"
     type: "execute"
```

```
[ { "resources" => [ { "name" => "do something", "type" => "execute" } ] } ]
```

### DIFF
--- a/lib/chef/application/apply.rb
+++ b/lib/chef/application/apply.rb
@@ -105,6 +105,12 @@ class Chef::Application::Apply < Chef::Application
     description: "Enable whyrun mode.",
     boolean: true
 
+  option :yaml,
+    long: "--yaml",
+    description: "Parse recipe as YAML",
+    boolean: true,
+    default: false
+
   option :profile_ruby,
     long: "--[no-]profile-ruby",
     description: "Dump complete Ruby call graph stack of entire #{Chef::Dist::PRODUCT} run (expert only).",
@@ -148,6 +154,10 @@ class Chef::Application::Apply < Chef::Application
     if file_name.nil?
       Chef::Application.fatal!("No recipe file was provided", Chef::Exceptions::RecipeNotFound.new)
     else
+      if file_name =~ /\.yml$/
+        logger.info "Recipe file name ends with .yml, parsing as YAML"
+        config[:yaml] = true
+      end
       recipe_path = File.expand_path(file_name)
       unless File.exist?(recipe_path)
         Chef::Application.fatal!("No file exists at #{recipe_path}", Chef::Exceptions::RecipeNotFound.new)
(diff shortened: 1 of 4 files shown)


lib/chef/application/apply.rb
```diff
@@ -148,6 +154,10 @@ def read_recipe_file(file_name)
     if file_name.nil?
       Chef::Application.fatal!("No recipe file was provided", Chef::Exceptions::RecipeNotFound.new)
     else
+      if file_name =~ /\.yml$/
```
phiggins:
```suggestion
      if File.extname(file_name) == ".yml"
```

Should this use `File.extname` to avoid any potential regex weirdness?

lib/chef/application/apply.rb
```diff
@@ -198,7 +208,11 @@ def run_chef_recipe
       @recipe_text, @recipe_fh = read_recipe_file @recipe_filename
     end
     recipe, run_context = get_recipe_and_run_context
-    recipe.instance_eval(@recipe_text, @recipe_filename, 1)
+    if config[:yaml]
```
lamont-granquist: this should not really be in the config hash and should go into an instance variable at the very least.

i sort of feel like this should mostly be handled by Chef::Recipe rather than here, and if you pass a filename with .yml then Chef::Recipe should have an API which handles it.  but its complicated by the way that earlier in this method it optionally loads in a string off of STDIN or the cli.

you've also still got the filename here, and i'd almost prefer checking the extension of the filename twice rather than passing around state.

lib/chef/recipe.rb
```diff
@@ -88,26 +89,24 @@ def untag(*tags)
     def from_yaml_file(filename)
       self.source_file = filename
       if File.file?(filename) && File.readable?(filename)
-        from_yaml(IO.read(filename))
+        yaml_contents = IO.read(filename)
+        if ::YAML.load_stream(yaml_contents).length > 1
+          raise ArgumentError, "YAML recipe '#{filename}' contains multiple documents, only one is supported"
+        end
```
phiggins: TIL, I had no idea this was a thing.

btm: YIL YAML is based off rfc822 (ARPA INTERNET TEXT MESSAGES 1982).

lib/chef/application/apply.rb
btm: seems good, I was cargo culting.

lib/chef/application/apply.rb
btm: right, this isn't just being used to avoid checking the filename a second time, this is for the case where you're passing `chef-apply --yaml`. I guess the above conditional with the logging could move down here too, but if we just did checked the file extension here then `something | chef-apply -s --yaml` wouldn't work.

I'd probably accept the argument that it shouldn't, because if you're doing that the training wheels should probably be off. But that's what I was doing.

btm: I removed this check on this line, but did keep logic here rather than Chef::Recipe to avoid changing the arguments to Chef::Recipe to add one forcing yaml mode there.

lamont-granquist: Got it, i somehow missed the --yaml option.  but still not mutating that argument is probably better.

PR #9508, opened 2020-03-18
Add a `knife yaml convert` tool

btm:
Very basic interface, reads in a YAML recipe, ruby recipe goes to stdout.

```
---
resources:
# This is an example of how you create a text file in chef
#
#  - type: "file"
#    name: "/tmp/jmc/airplanes"
#    content: "they fly in the air"

#  This is an example of how you run a bash command in chef
#  - type: "bash"
#    name: "touch a file!"
#    code: "touch /tmp/jmc/bash"

# This is a multiline code block in YAML
  - type: "bash"
    name: "new chef user"
    code: "
    touch /tmp/jmc/something ;
    echo something >> /tmp/jmc/something ;
    rm -f /tmp/jmc/bash ;
    "

# This is ignored because it isn't in the resources hash    
something:
  - other: "dont use this"
```

gets converted to:

```

bash "new chef user" do
  code " touch /tmp/jmc/something ; echo something >> /tmp/jmc/something ; rm -f /tmp/jmc/bash ; "
end
```

### DIFF
--- a/lib/chef/knife/yaml_convert.rb
+++ b/lib/chef/knife/yaml_convert.rb
@@ -1,0 +1,67 @@
+#
+# Author:: Bryan McLellan <btm@loftninjas.org>
+# Copyright:: Copyright 2020, Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "yaml"
+require_relative "../knife"
+class Chef::Knife::YamlConvert < Chef::Knife
+
+  banner "knife yaml convert YAML_FILENAME"
+
+  def run
+    if name_args.empty?
+      ui.error("Please specify the file name of a YAML recipe to convert to Ruby")
+      exit 1
+    elsif name_args.size >= 2
+      ui.error("Only one recipe may converted at a time")
+      exit 1
+    end
+
+    filename = @name_args[0]
+    yaml_contents = IO.read(filename)
+
+    if ::YAML.load_stream(yaml_contents).length > 1
+      raise ArgumentError, "YAML recipe '#{filename}' contains multiple documents, only one is supported"
+    end
+
+    # Unfortunately, per the YAML spec, comments are stripped when we load, so we lose them on conversion
+    yaml_hash = ::YAML.safe_load(yaml_contents)
+    unless yaml_hash.is_a?(Hash) && yaml_hash.key?("resources")
+      raise ArgumentError, "YAML recipe '#{source_file}' must contain a top-level 'resources' hash (YAML sequence), i.e. 'resources:'"
+    end
+
+    ruby_contents = []
+    ruby_contents << "# Autoconverted recipe from #{filename}"
+    ruby_contents << ""
+
+    yaml_hash["resources"].each do |r|
+      type = r.delete("type")
+      name = r.delete("name")
+
+      ruby_contents << "#{type} \"#{name}\" do"
+      r.each do |p|
+        ruby_contents << "  #{p.shift} \"#{p.shift}\""
+      end
+      ruby_contents << "end"
+      ruby_contents << ""
+    end
+
+    ruby_contents.each do |l|
+      puts l
+    end
+  end
+end
--- a/lib/chef/knife/yaml_convert.rb
+++ b/lib/chef/knife/yaml_convert.rb
@@ -20,48 +20,72 @@ require "yaml"
 require_relative "../knife"
 class Chef::Knife::YamlConvert < Chef::Knife
 
-  banner "knife yaml convert YAML_FILENAME"
+  banner "knife yaml convert YAML_FILENAME [RUBY_FILENAME]"
 
   def run
     if name_args.empty?
-      ui.error("Please specify the file name of a YAML recipe to convert to Ruby")
-      exit 1
-    elsif name_args.size >= 2
-      ui.error("Only one recipe may converted at a time")
-      exit 1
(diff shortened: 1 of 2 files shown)


lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,66 @@
...
+      exit 1
+    elsif name_args.size >= 2
+      ui.error("Only one recipe may converted at a time")
+      exit 1
+    end
+
+    filename = @name_args[0]
+    yaml_contents = IO.read(filename)
+
+    if ::YAML.load_stream(yaml_contents).length > 1
```
tas50: Shouldn't this require yaml since it uses it?

btm: I did at first, and then I noticed that @lamont-granquist didn't add it in #8653, so thought maybe it was one of those newer things about how we're doing requires that I didn't understand, but this morning I realize it was probably just an oversight.

tas50:
I assume at some point we'd add a --o option here to save it without pipes?

btm:
@tas50 yeah maybe? I tossed around for a minute about if it should be a required in-file and out-file, and then have stdin / stdout flags. i think I ended up being a little lazy realizing that if I use stdout then I don't have to deal with checking for the outfile existence and raising or whatnot.

I also thought for a moment about would you ever want to do this in bulk? but I don't think we've thought up with a complex enough reality for the not-complex thing where that you would use that.

lamont-granquist:
```
# This is ignored because it isn't in the resources hash    
something:
  - other: "dont use this"
```

this has me a bit worried that we're doing it wrong.

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,67 @@
...
+      r.each do |p|
+        ruby_contents << "  #{p.shift} \"#{p.shift}\""
+      end
+      ruby_contents << "end"
+      ruby_contents << ""
+    end
+
+    ruby_contents.each do |l|
+      puts l
+    end
```
lamont-granquist: `puts ruby_contents.join("\n")` for a bit more succinctness.

lamont-granquist: should probably output to the `.rb` file of the same basename

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,67 @@
...
+      type = r.delete("type")
+      name = r.delete("name")
+
+      ruby_contents << "#{type} \"#{name}\" do"
+      r.each do |p|
+        ruby_contents << "  #{p.shift} \"#{p.shift}\""
+      end
+      ruby_contents << "end"
+      ruby_contents << ""
+    end
```
lamont-granquist: this bit is probably long enough to be extracted to a method to convert the resources array-of-hashes into a string (`string = resources_to_string(yaml_hash["resources"])`)

lamont-granquist:
The other thing to think about is maybe to convert all the recipes in a given cookbook.  So more of a `knife cookbook yaml covert` kind of API to bulk convert all yaml recipes to rb recipes in the cookbook.

btm:
> ```
> # This is ignored because it isn't in the resources hash    
> something:
>   - other: "dont use this"
> ```
> 
> this has me a bit worried that we're doing it wrong.

In what way? My presumption has been there could be other policy we want to add later that doesn't make obvious sense yet because we're stuck thinking about the traditional cookbook patterns. The resources namespace saves that option and is kind of slightly self-documenting as to what follows.

> The other thing to think about is maybe to convert all the recipes in a given cookbook. So more of a `knife cookbook yaml covert` kind of API to bulk convert all yaml recipes to rb recipes in the cookbook.

Yeah, I mentioned this in my comment to @tas50. Are people going to have more than a few yaml recipes in one cookbook? I don't know. It doesn't sound like a good idea but I imagine it'll end up happening. At the moment I'm not convinced that we need it yet because of that.

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      ruby_contents << "#{type} \"#{name}\" do"
+      r.each do |p|
+        ruby_contents << "  #{p.shift} \"#{p.shift}\""
+      end
+      ruby_contents << "end\n"
+    end
+
+    ruby_contents
+  end
```
phiggins: nitpick: This function is called `resources_to_string` but it doesn't return a string. It might make more sense to move the `join("\n")` here.

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,85 @@
...
+    ruby_contents << "# Autoconverted recipe from #{filename}\n"
+
+    resource_hash.each do |r|
+      type = r.delete("type")
+      name = r.delete("name")
+
+      ruby_contents << "#{type} \"#{name}\" do"
+      r.each do |p|
+        ruby_contents << "  #{p.shift} \"#{p.shift}\""
+      end
```
phiggins:
```suggestion
      r.each do |k,v|
        ruby_contents << "  #{k} \"#{v}\""
      end
```

Nitpicky, but it wasn't clear to me that this wasn't doing something tricky with the collection or values.

btm: I was trying to pass along inline comments before I realized YAML filtered them all out.

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,87 @@
...
+  banner "knife yaml convert YAML_FILENAME [RUBY_FILENAME]"
+
+  def run
+    if name_args.empty?
+      ui.fatal!("Please specify the file name of a YAML recipe to convert to Ruby")
+    elsif name_args.size >= 3
+      ui.fatal!("knife yaml convert YAML_FILENAME [RUBY_FILENAME]")
+    end
+
+    yaml_file = @name_args[0]
```
vsingh-msys: It would be nice if we can check the existing of the provided file here.

```suggestion
    yaml_file = @name_args[0]
    ui.fatal!("File #{yaml_file} does not exist/unadle to read") unless ::File.exist?(yaml_file)
```

lib/chef/knife/yaml_convert.rb
```diff
@@ -0,0 +1,87 @@
...
+    ruby_file = if @name_args[1]
+                  @name_args[1] # use the specified output filename if provided
+                else
+                  if File.extname(yaml_file) == ".yml" || File.extname(yaml_file) == ".yaml"
```
vsingh-msys: [Minor] might be we can replace `File` with `::File`?

PR #9518, opened 2020-03-20
Add chef_client_cron resource from chef-client cookbook

tas50:
Adds a simple resource for setting the chef-client to run on a schedule.

### TODO:

- [x] Add unit tests
- [x] Add log directory creation similar to chef_client_scheduled_task
- [x] Add usage to the Test Kitchen integration tests
- [x] Add descriptions to the properties


lamont-granquist:
yeah you probably want to have the shape of the API across all the chef_client_whatever resources to be as consistent as possible.

and i think you want to be using the cinc/dist values.

lib/chef/resource/chef_client_cron.rb
```diff
@@ -0,0 +1,181 @@
...
+          random = Random.new(seed.to_i)
+          random.rand(splay)
+        end
+
+        def cron_command
+          cmd = ""
+          cmd << "/bin/sleep #{splay_sleep_time(new_resource.splay)}; "
+          cmd << "#{new_resource.chef_binary_path} "
+          cmd << "#{new_resource.daemon_options.join(" ")} " unless new_resource.daemon_options.empty?
+          cmd << "#{new_resource.append_log_file ? ">>" : ">"} #{::File.join(new_resource.log_directory, new_resource.log_file_name)} 2>&1"
```
lamont-granquist: directly catting to the file seems like something we should discourage here in favor of the -L option to the client?

lib/chef/resource/chef_client_cron.rb
```diff
@@ -0,0 +1,181 @@
...
+      property :log_directory, String,
+        default: lazy { platform?("mac_os_x") ? "/Library/Logs/#{Chef::Dist::DIR_SUFFIX.capitalize}" : "/var/log/#{Chef::Dist::DIR_SUFFIX}" },
+        description: "The path of the directory to create the log file in."
+
+      property :log_file_name, String,
+        default: "client.log",
+        description: "The name of the log file to use."
+
+      property :append_log_file, [true, false],
+        default: false,
```
lamont-granquist: this should default to true

tas50: If it's true then folks have to setup log rotation which is why the cookbook historically didn't do that.

lib/chef/resource/chef_client_cron.rb
tas50: Pretty sure that was just to support appending vs. overwriting

lib/chef/resource/chef_client_cron.rb
```diff
@@ -0,0 +1,181 @@
...
+
+        def cron_command
+          cmd = ""
+          cmd << "/bin/sleep #{splay_sleep_time(new_resource.splay)}; "
+          cmd << "#{new_resource.chef_binary_path} "
+          cmd << "#{new_resource.daemon_options.join(" ")} " unless new_resource.daemon_options.empty?
+          cmd << "#{new_resource.append_log_file ? ">>" : ">"} #{::File.join(new_resource.log_directory, new_resource.log_file_name)} 2>&1"
+          cmd << " || echo \"#{Chef::Dist::PRODUCT} execution failed\"" if new_resource.mailto
+          cmd
+        end
```
phiggins: Does this need to include `environment`?

tas50: I'm passing that to the cron resource directly which handles parsing the hash in chef itself.

lib/chef/resource/chef_client_cron.rb
tas50: I updated appending to use -L

lib/chef/resource/chef_client_cron.rb
tas50: I've converted this to true

PR #9521, opened 2020-03-21
Revert windows server manager spec helper

tas50:
This was causing our specs to fail


tas50:
This turned out to not fix things. Digging further.

PR #9522, opened 2020-03-23
Move cron_d validation methods to a helper and use that in the cron resource

tas50:
1) Prevent duplicate code
2) Provide better validation in the cron resource
3) convert a few set_or_return methods in cron resource to true properties for better input 
validation
4) Update some specs to reflect the actual properties in cron now


lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,90 @@
...
+      # validate a provided value is between two other provided values
+      # we also allow * as a valid input
+      # @param spec the value to validate
+      # @param min the lowest value allowed
+      # @param max the highest value allowed
+      # @return [Boolean] valid or not?
+      def self.validate_numeric(spec, min, max)
+        return true if spec == "*"
+
+        #  binding.pry
```
phiggins:
```suggestion

```

lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,90 @@
...
+        if spec.respond_to? :to_int
+          return false unless spec >= min && spec <= max
+
+          return true
```
phiggins:
```suggestion
          return spec >= min && spec <= max
```

phiggins: You could even simplify the whole if block to:

```
return spec >= min && spec <= max if spec.respond_to? :to_int
```

but that's kinda dense.

lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,90 @@
...
+      # validate the provided month value to be jan - dec, 1 - 12, or *
+      # @param spec the value to validate
+      # @return [Boolean] valid or not?
+      def self.validate_month(spec)
+        return true if spec == "*"
+
+        if spec.respond_to? :to_int
+          validate_numeric(spec, 1, 12)
+        elsif spec.respond_to? :to_str
+          return true if spec == "*"
```
phiggins:
```suggestion

```

This is handled on the first line of the function, right?

tas50: I just noticed these all a few minutes ago. They seem entirely redundant.

lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,90 @@
...
+      # validate the provided day of the week is sun-sat, 0-7, or *
+      # @param spec the value to validate
+      # @return [Boolean] valid or not?
+      def self.validate_dow(spec)
+        return true if spec == "*"
+
+        if spec.respond_to? :to_int
+          validate_numeric(spec, 0, 7)
+        elsif spec.respond_to? :to_str
+          return true if spec == "*"
```
phiggins:
```suggestion

```

This gets tested on the first line of the function.

lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,85 @@
...
+          validate_numeric(spec, 0, 7)
+        elsif spec.respond_to? :to_str
+          # Named abbreviations are permitted but not as part of a range or with stepping
+          return true if %w{sun mon tue wed thu fri sat}.include? spec.downcase
+
+          # 0-7 are legal for days of week
+          validate_numeric(spec, 0, 7)
+        else
+          false
+        end
```
phiggins:
```suggestion
        return spec == "*" || 
          validate_numeric(spec, 0, 7) || 
          %w{sun mon tue wed thu fri sat}.include?(String(spec).downcase)
```

This could be collapsed into something like this, but this might be too dense.

tas50: as long as we're maintaining the comments as to why these things are allowed I'm fine.

lib/chef/resource/cron.rb
```diff
@@ -35,80 +36,33 @@ class Cron < Chef::Resource
...
-          kind_of: String
-        )
-      end
+      property :month, [Integer, String],
+        description: "The month in the year on which a cron entry is to run (1 - 12, jan-dec, or *).",
+        default: "*", callbacks: {
+          "should be a valid month spec" => ->(spec) { Chef::Mixin::CronValidations.validate_month(spec) },
+        }
 
       def weekday(arg = nil)
```
phiggins: Some of the code in `weekday` could be replaced by `validate_dow`.

phiggins: (maybe)

spec/unit/mixin/cron_validations_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    it "it accepts the string of *" do
+      expect(Chef::Mixin::CronValidations.validate_month("*")).to be true
+    end
+
+    it "returns false for an out of range integer" do
+      expect(Chef::Mixin::CronValidations.validate_month(13)).to be false
+    end
+
+    it "returns false for an invalid string" do
+      expect(Chef::Mixin::CronValidations.validate_month("janurary")).to be false
```
phiggins: Is this false because its the whole name of the month or because the month is misspelled? 🤔

tas50: I think the point was the typo. These are all just moved from the existing specs

spec/unit/mixin/cron_validations_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  context "#validate_numeric" do
+    it "returns true if the value is in the allowed range" do
+      expect(Chef::Mixin::CronValidations.validate_numeric(5, 1, 100)).to be true
+    end
+
+    it "returns false if the value is out of the allowed range" do
+      expect(Chef::Mixin::CronValidations.validate_numeric(-1, 1, 100)).to be false
+    end
+  end
```
phiggins: Is the range inclusive or exclusive? It might be good to add an example on the far end.

lib/chef/mixin/cron_validations.rb
```diff
@@ -0,0 +1,90 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  module Mixin
+    # a collection of methods for validating cron times. Used in the various cron-like resources
+    module CronValidations
```
phiggins: This is called `Chef::Mixin::CronValidations` but it's not usable as a mixin...

lib/chef/resource/cron.rb
tas50: Yeah that's going to take a lot more work because the old resource allowed you to pass things like Monday which is then mapped to a numeric value based on the position in an array. These resources are similar but different and my hope in unifying this validation logic was to prevent any further property uses. We still have to account for the old ones at some point.

phiggins: I was curious about that. They seemed similar on the surface but this makes sense. 👍

spec/unit/mixin/cron_validations_spec.rb
tas50: I expanded this to be 2 specs

spec/unit/mixin/cron_validations_spec.rb
tas50: I noted it's because of a typo in the description

lib/chef/mixin/cron_validations.rb
tas50: The word mixin wildly abused in chef/chef

tas50: Renamed

Issue #9166, opened 2019-12-16
windows_share: can't modify frozen String


Kundan22:
## Description
Compilation fails with error "can't modify frozen String". It shows error at **path** in the windows_share method.

```
19:  windows_share 'AppFabricConfig' do
20:    action :nothing
21>>   path config_path
22:  end
```

## Chef Version
**15.5.17**

## Platform Version
Windows Server 2016

System Info:
------------

> chef_version=15.5.17
> platform=windows
> platform_version=10.0.14393
> ruby=ruby 2.6.5p114 (2019-10-01 revision 67812) [x64-mingw32]
> program_name=C:/opscode/chef/bin/chef-client
> executable=C:/opscode/chef/bin/chef-client

## Replication Case
Occurs every time when I am trying to run a chef-client. It doesn't compile.

## Client Output

> Running handlers:
> [2019-12-16T15:28:29+01:00] ERROR: Running exception handlers
> Running handlers complete
> [2019-12-16T15:28:29+01:00] ERROR: Exception handlers complete
> Chef Infra Client failed. 0 resources updated in 19 seconds
> [2019-12-16T15:28:29+01:00] FATAL: Stacktrace dumped to c:/chef/cache/chef-stacktrace.out
> [2019-12-16T15:28:29+01:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
> [2019-12-16T15:28:29+01:00] FATAL: FrozenError: can't modify frozen String

```
```

lamont-granquist:
We would probably need the contents of the stacktrace.out file to track this down.

Kundan22:
Attached
[chef-stacktrace.out.log](https://github.com/chef/chef/files/3972357/chef-stacktrace.out.log)

Kundan22:
**can you please increase the priority of this issue as we are not able to compile our cookbooks because of this issue.**

lamont-granquist:
For escalating support issues you should inquire to our sales people about pricing for a support agreement and should be going through our helpdesk.

You can fix the root cause of this yourself though since the problem is going to be that the config_path string is frozen and you can either remove that .freeze or if that comes from somewhere else you should be able to dup the string one way or another to bypass it.

tas50:
@Kundan22 what type of value is the config_path variable and where is that being set?

dheerajd-msys:
@Kundan22 Any updates on ^

Kundan22:
@tas50 : C:\AppFabricConfig

[...]

PR #9524, opened 2020-03-24
Fixed windows share cannot modify frozen string bug

sanga1794:
## Description
Compilation fails with error "can't modify frozen String". It shows error at path in the windows_share method.

## Related Issue
fixed: https://github.com/chef/chef/issues/9166

## Types of changes

## Checklist:

### DIFF
--- a/spec/unit/resource/windows_share_spec.rb
+++ b/spec/unit/resource/windows_share_spec.rb
@@ -38,11 +38,9 @@ describe Chef::Resource::WindowsShare do
   end
 
   it "coerces path to a single path separator format" do
-    resource.path("C:/chef")
+    resource.path("C:/chef".freeze)
     expect(resource.path).to eql("C:\\chef")
     resource.path("C:\\chef")
-    expect(resource.path).to eql("C:\\chef")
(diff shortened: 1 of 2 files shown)


spec/unit/resource/windows_share_spec.rb
```diff
@@ -42,5 +42,9 @@
     expect(resource.path).to eql("C:\\chef")
     resource.path("C:\\chef")
     expect(resource.path).to eql("C:\\chef")
+    resource.path("C:/chef".freeze)
```
tas50: I'm not sure if we need these?

sanga1794: Okk. will update it. Thanks @tas50

tas50: If you're trying to make sure the path was frozen then you can just freeze the first one in this spec right?

sanga1794: Ok, I'll update it, thanks @tas50

sanga1794: Hello @tas50, All tests other than kitchen are passed.

PR #9525, opened 2020-03-24
Add : https_for_ca_consumer property to rhsm_register resource

jasonwbarnett:
## Description

This will enable people to increase their security posture by verifying the remote satellite host SSL certificate.

## Related Issue

N/A

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/rhsm_register.rb
+++ b/lib/chef/resource/rhsm_register.rb
@@ -16,6 +16,7 @@
 #
 
 require_relative "../resource"
+require_relative "../dist"
 require "shellwords" unless defined?(Shellwords)
 
 class Chef
@@ -60,6 +61,10 @@ class Chef
         description: "If true, the system will be registered even if it is already registered. Normally, any register operations will fail if the machine has already been registered.",
         default: false, desired_state: false
 
+      property :https_for_ca_consumer, [TrueClass, FalseClass],
+        description: "If true, #{Chef::Dist::PRODUCT} will fetch the katello-ca-consumer-latest.noarch.rpm from the satellite_host using HTTPS.",
+        default: false, desired_state: false
+
       action :register do
         description "Register the node with RHSM."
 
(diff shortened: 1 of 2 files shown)


lib/chef/resource/rhsm_register.rb
```diff
@@ -60,6 +60,10 @@ class RhsmRegister < Chef::Resource
         description: "If true, the system will be registered even if it is already registered. Normally, any register operations will fail if the machine has already been registered.",
         default: false, desired_state: false
 
+      property :use_https, [TrueClass, FalseClass],
+        description: "If true, chef will fetch the katello-ca-consumer-latest.noarch.rpm from the satellite_host using HTTPS.",
+        default: false, desired_state: false
```
lamont-granquist: should probably flip the default to true.  this is also a very generic sounding name of a property for a very specific function.  it should probably be `use_katello_https` or something like that.

lamont-granquist: also maybe consider just fixing this to be hardcoded?  would anyone really not want this?

jasonwbarnett: Out of the box internal Satellite servers use a self-signed certificate. Red Hat themselves even reference http (as opposed to https) urls for installation of katello-ca-consumer RPM.

Reference: https://access.redhat.com/documentation/en-us/red_hat_satellite/6.6/html-single/quick_start_guide/index#registering_existing_system

jasonwbarnett: > should probably flip the default to true. this is also a very generic sounding name of a property for a very specific function. it should probably be `use_katello_https` or something like that.

Agreed. I can't think of a short name that gives a clear gist of what it is for that isn't huge. Also, it's not katello, it's the katello-ca-consumer package which are different things.

lamont-granquist: `https_for_ca_consumer true` ?

jasonwbarnett: > `https_for_ca_consumer true` ?

works for me. I'll update and re-push in a bit.

lib/chef/resource/rhsm_register.rb
```diff
@@ -60,6 +60,10 @@ class RhsmRegister < Chef::Resource
         description: "If true, the system will be registered even if it is already registered. Normally, any register operations will fail if the machine has already been registered.",
         default: false, desired_state: false
 
+      property :https_for_ca_consumer, [TrueClass, FalseClass],
+        description: "If true, chef will fetch the katello-ca-consumer-latest.noarch.rpm from the satellite_host using HTTPS.",
```
tas50: The use of Chef needs to be a dist value here for PRODUCT

PR #9527, opened 2020-03-24
clean up chef_client_cron

lamont-granquist:
- add instance methods to the module so it can be mixed in
- extend the module into the resource class
- use explicit methods for "currying" arguments to avoid the cognitive
  overhead of the lambda syntax and allow using method()
- use method() to keep the callbacks readable

### DIFF
--- a/lib/chef/resource/chef_client_cron.rb
+++ b/lib/chef/resource/chef_client_cron.rb
@@ -1,5 +1,5 @@
 #
-# Copyright:: 2020, Chef Software Inc.
+# Copyright:: 2020-2020, Chef Software Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lib/chef/resource/chef_client_cron.rb
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright:: 2020, Chef Software Inc.
+# Copyright:: 2020-2020, Chef Software Inc.
```
tas50: ?

lamont-granquist: my vim script, never figured out what the regexp bug is there.

we could just delete all the dates.  they don't matter.

tas50: I'm fine with just nuking them

PR #9529, opened 2020-03-24
Add missing introduced value to the new property in rhsm_register

tas50:
I realized we needed this right after the merge


tas50:
Closing this out since @jasonwbarnett is going to backport this feature to 15.9 so we'll want this value to be 15.9

PR #9532, opened 2020-03-25
restore magic resource name ordering

lamont-granquist:
this restores the order that the magic wiring of LWRP / custom resources
to the cookbook_name + resource filename comes after the parsing of the
file that it is defined in.

i had moved this because i was running into issues with recursive calls
between provides and the resource_name method, which i later solved.

this restores the prior behavior, but calls provides instead of
resource_name, but relies on the implicit call to resource_name from
the first provides to be functionally identical to the prior behavior
anyway.

no idea if this actually fixes any bugs, but it was not really
intentional to change it.  stuff like chef_version_for_provides
needs to be called before calling provides, so it seems like the
behavior this fixes should have been buggy with respect to that
but it is obviously...  complicated.

this probably fixes issues where if the provides is called before the parsing of the resource file
itself the stuff like `chef_version_for_provides` will not work correctly since that needs to come
first (along with just restoring the prior order of parsing and probably fixing other edge conditions
that depend on that ordering).

### DIFF
--- a/lib/chef/resource/lwrp_base.rb
+++ b/lib/chef/resource/lwrp_base.rb
@@ -51,7 +51,6 @@ class Chef
 
           resource_class = Class.new(self)
           resource_class.run_context = run_context
-          resource_class.provides resource_name.to_sym
           resource_class.class_from_file(filename)
 
           # Make a useful string for the class (rather than <Class:312894723894>)
@@ -65,6 +64,10 @@ class Chef
           Chef::Log.trace("Loaded contents of #{filename} into resource #{resource_name} (#{resource_class})")
 
           LWRPBase.loaded_lwrps[filename] = true
+
+          # wire up the default resource name after the class is parsed only if we haven't declared one.
+          # (this ordering is important for MapCollision deprecation warnings)
+          resource_class.provides resource_name.to_sym unless Chef::ResourceResolver.includes_handler?(resource_name.to_sym, self)
 
           resource_class
         end


lamont-granquist:
weird, this should have worked but clearly does not.

lamont-granquist:
This is probably the right thing to do.  Across the whole supermarket this will break the following resources:

```
cerner_splunk/resources/sh_cluster.rb:provides :sh_cluster
k8s/resources/pod.rb:provides :kube_pod
k8s/resources/replication_controller.rb:provides :kube_replication_controller
k8s/resources/service.rb:provides :kube_service
kubernetes/resources/master.rb:provides :kube_master
kubernetes/resources/node.rb:provides :kube_node
kubernetes/resources/pod.rb:provides :kube_pod
kubernetes/resources/replication_controller.rb:provides :kube_replication_controller
kubernetes/resources/service.rb:provides :kube_service
qas/resources/vastool.rb:provides :vastool
sc-mongodb/resources/user.rb:provides :mongodb_user
```

What is broken about those is that they are old style LWRP resources which define a provides, which is NOT the magical cookbook_name+filename provides AND then the provider does NOT declare a provides line.

The fix is that the provider should declare a provides line...  or convert them to a custom resource.

lamont-granquist:
required fixing chefspec which was testing the very old chef-11 style LWRP provides behavior:  chefspec/chefspec#983

PR #9533, opened 2020-03-25
Switch our spec helper check to ChefUtils.windows_ruby

tas50:
This should work instead of ChefUtils.windows and is simpler than the RUBY_PLATFORM check.


tas50:
the adhoc for this failed

PR #9535, opened 2020-03-25
Add :swap action to yum_package resource

jasonwbarnett:
## Description

For folks who are running RHEL distros and need to swap in different packages from the [IUS repository](https://ius.io/usage), this would give them the ability to do so natively in Chef Infra Client.

### Example

```ruby
yum_package 'git222' do
  swap_from 'git'
  action :swap
end
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -41,6 +41,8 @@ class Chef
 
         provides :yum_package
 
+        attr_accessor :swap_from_current_resource
+
         #
         # Most of the magic in this class happens in the python helper script.  The ruby side of this
         # provider knows only enough to translate Chef-style new_resource name+package+version into
@@ -60,6 +62,11 @@ class Chef
           @current_resource = Chef::Resource::YumPackage.new(new_resource.name)
           current_resource.package_name(new_resource.package_name)
           current_resource.version(get_current_versions)
+
+          if new_resource.actions.include?(:swap)
+            @swap_from_current_resource = Chef::Resource::YumPackage.new(new_resource.swap_from)
+            swap_from_current_resource.version(get_swap_from_current_version)
+          end
 
           current_resource
         end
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/yum.rb
```diff
@@ -71,6 +78,18 @@ def define_resource_requirements
...
+            a.assertion { !new_resource.swap_from.nil? }
+            a.failure_message Chef::Exceptions::Package, ":swap action requires the swap_from property"
+            a.whyrun "assuming #{new_resource.source} would have previously been created"
+          end
+
+          requirements.assert(:swap) do |a|
+            a.assertion { new_resource.package_name.length == 1 }
+            a.failure_message Chef::Exceptions::Package, ":swap action only supports a single package_name"
+            a.whyrun "assuming #{new_resource.source} would have previously been created"
+          end
```
lamont-granquist: neither of those should have a whyrun clause, those are API-shape issues and aren't fixable via having other resources run first.

lib/chef/provider/package/yum.rb
```diff
@@ -61,6 +63,11 @@ def load_current_resource
           current_resource.package_name(new_resource.package_name)
           current_resource.version(get_current_versions)
 
+          if new_resource.actions.include?(:swap)
+            @swap_from_current_resource = Chef::Resource::YumPackage.new(new_resource.swap_from)
+            swap_from_current_resource.version(get_swap_from_current_version)
```
lamont-granquist: you probably don't need to use a current_resource-looking-thing here.  you can't ever wire it up to reporting, or to load_current_value or to converge_if_changed, so the whole machinery around that abstraction can't really be used.  you probably just want a helper that returns only the `swap_from_version(package_name)`

lamont-granquist:
closing this since it seems to have stalled

PR #9544, opened 2020-03-26
windows_firewall_rule: Fix idempotency and add icmp_type property

### LINKED ISSUE #9542: windows_firewall_rule does not support icmp_type when setting protocol with ICMP
## Description
When I set a new windows firewall rule with a protocol that starts with ICMP, I have to add icmp_type. This is not included right now.

## Chef Version
16.0.*

## Platform Version
Windows

## Replication Case
Add a default windows_firewall_rule with property protocol set to "ICMPv6" (example) and run the resource. It will fail in a powershell failure, which states to set -IcmpType.

## Client Output
not necessary, as missing feature

## Stacktrace
not necessary, as missing feature

### LINKED ISSUE #9543: windows_firewall_rule always overrides description even when not necessary
## Description
When I want to update a windows firewall rule with my specified properties, which do not contain any description in it, the description should not be updated.
Right now, this is the behavior, which should be changed by removing the property default of descriptiion.

## Chef Version
16.0.*

## Platform Version
Windows

## Replication Case
Update any existing firewall rule. It will always set 'Firewall rule' in the description, when no other value is present.

## Client Output
not necessary

## Stacktrace
not necessary

### LINKED ISSUE #9546: windows_firewall_rule missing tests for adding groups
## Description
The group property has been added, but do not have any tests for it. 

## Chef Version
16.0

## Platform Version
Windows

## Replication Case
Not needed

## Client Output
Not needed

## Stacktrace
Not needed

pschaumburg:
This PR contains several things, which are described in the linked issues.

## Description
- icmp_type (new property): this type is always needed, when a protocol starts with ICMP (default value "Any"). 
- description (change property) : this property is not needed for creating a firewall rule. When updating a standard default windows firewall rule, the description is always updated, even when not necessary. So the default value will be removed.
- add validation for protocol + icmp_type during action :creation
- add tests for group and icmp_type
- add examples to windows_firewall_rule out of the current docs

## Related Issue
closes #9542
closes #9543 
closes #9546 

## Types of changes

## Checklist:


lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -186,6 +226,7 @@ def firewall_command(cmdlet_type)
           cmd << " -RemotePort '#{new_resource.remote_port.join("', '")}'" if new_resource.remote_port
           cmd << " -Direction '#{new_resource.direction}'" if new_resource.direction
           cmd << " -Protocol '#{new_resource.protocol}'" if new_resource.protocol
+          cmd << " -IcmpType '#{new_resource.icmp_type}'"
```
Vasu1105: @pschaumburg may be we should add the condition as used for other options ?

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
...
+        # return the current firewall rule settings for icmp_type and protocol are correct
+        # @return [Boolean]
+        def is_set_properly?(icmp_type, protocol)
+          if icmp_type.to_s.empty?
+            return false # rubocop:disable Style/RedundantReturn
+
+          elsif icmp_type.is_a?(Integer)
+            return false unless protocol.start_with?("ICMP")
+            return false unless (0..255).include?(icmp_type)
+
```
Vasu1105: May be we can use or condition here and optimise it.

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
...
+            true
+
+          elsif icmp_type.is_a?(String)
+            return false if !protocol.start_with?("ICMP") && icmp_type !~ /^\D+$/
+            return false if icmp_type.count("a-zA-Z") > 0 && icmp_type.count(":") > 1
+            return false unless (0..255).include?(icmp_type.to_i)
+
```
Vasu1105: same  applies here. Instead of this multiple return statements use or for conditions and optimise.

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -139,10 +172,17 @@ class WindowsFirewallRule < Chef::Resource
       action :create do
         description "Create a Windows firewall entry."
 
+        unless is_set_properly?(new_resource.icmp_type, new_resource.protocol)
```
tas50: We have a method `define_resource_requirements` that gets called at the very beginning, which I think handles the scenario you're hitting here. Here's how I used it to validate inputs per action in the `alternatives` resource.

https://github.com/chef/chef/blob/master/lib/chef/resource/alternatives.rb#L37

lib/chef/resource/windows_firewall_rule.rb
pschaumburg: The condition is not needed for anything. It will always add the IcmpType to the Powershell and do not break anything.

lib/chef/resource/windows_firewall_rule.rb
pschaumburg: Will definitely look into this. Thanks a lot!

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
 
           cmd
         end
+
+        # return the current firewall rule settings for icmp_type and protocol are correct
+        # @return [Boolean]
+        def is_set_properly?(icmp_type, protocol)
```
phiggins: nitpick: it's considered bad style in ruby to preface method names with `is_`, `get_`, etc.

spec/unit/resource/windows_firewall_rule_spec.rb
```diff
@@ -457,13 +491,92 @@
...
+        expect(provider.is_set_properly?(345, "ICMPv6")).to eql(false)
+      end
+
+      it "icmp_type is '1:3' as code pair" do
+        expect(provider.is_set_properly?("1:3", "ICMPv6")).to eql(true)
+      end
+
+      it "icmp_type is '123:456' as code pair out of range" do
+        expect(provider.is_set_properly?("123:456", "ICMPv6")).to eql(false)
       end
```
phiggins: Is 0 a valid value? Is 254 or 255 the max allowed upper bound?

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
...
+            return false if !protocol.start_with?("ICMP") && icmp_type !~ /^\D+$/
+            return false if icmp_type.count("a-zA-Z") > 0 && icmp_type.count(":") > 1
+            return false unless (0..255).include?(icmp_type.to_i)
+
+            if protocol.start_with?("ICMP") && icmp_type.include?(":")
+              return icmp_type.split(":").all? { |type| (0..255).include?(type.to_i) }
+            end
+
+            # the following has to be disabled, as this works like a safeguard
+            true # rubocop:disable Style/RedundantReturn
```
phiggins:
```suggestion
          case icmp_type
          when Integer
            protocol.start_with?("ICMP") && (0..255).include?(icmp_type)
          when String
            # string logic
          else
            false
          end

```

Like others have mentioned, this code could likely be simplified for more clarity.

I would have filled in the string logic but it was more complicated than I realized and felt like more work than it was worth trying to fix in a github comment box. :)

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
...
+          end
```
phiggins:
```
          case icmp_type
          when Integer
            protocol.start_with?("ICMP") && (0..255).include?(icmp_type)
          when String
            # string logic
          else
            false
          end

```

Like others have mentioned, this code could likely be simplified for more clarity.

I would have filled in the string logic but it was more complicated than I realized and felt like more work than it was worth trying to fix in a github comment box. :)

spec/unit/resource/windows_firewall_rule_spec.rb
```diff
@@ -457,13 +491,92 @@
...
+        expect(provider.is_set_properly?(345, "TCP")).to eql(false)
+      end
+
+      it "icmp_type is '1:3' as code pair" do
+        expect(provider.is_set_properly?("1:3", "TCP")).to eql(false)
+      end
+
+      it "icmp_type is '123:456' as code pair out of range" do
+        expect(provider.is_set_properly?("123:456", "TCP")).to eql(false)
+      end
```
phiggins: This probably needs a test for the opposite case, where the first number in the pair is out of range (ie, `456:123`).

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +236,38 @@ def firewall_command(cmdlet_type)
...
         end
+
+        # return the current firewall rule settings for icmp_type and protocol are correct
+        # @return [Boolean]
+        def is_set_properly?(icmp_type, protocol)
+          if icmp_type.to_s.empty?
+            return false # rubocop:disable Style/RedundantReturn
+
+          elsif icmp_type.is_a?(Integer)
+            return false unless protocol.start_with?("ICMP")
```
phiggins: Do these checks need to be `protocol == "ICMP"`? Is a value of `ICMP2` valid?

pschaumburg: Yes, ICMPv4 and ICMPv6 are fine right now. This check is only needed for every protocol, which starts with ICMP, as every ICMP protocol requires an icmp_type. All other protocol types (which might become more) are only allowed to use icmp_type == 'Any'

spec/unit/resource/windows_firewall_rule_spec.rb
pschaumburg: Right now, the check of the definition simply splits those pairs into single numbers and checks them, if any of these are not in a range. But if you want to have this test in here, I'll add it.

spec/unit/resource/windows_firewall_rule_spec.rb
pschaumburg: 0 is the minimum and 255 the maximum value. 
This is related to the docs (https://docs.microsoft.com/en-us/powershell/module/netsecurity/new-netfirewallrule?view=win10-ps). Search there for "-IcmpType" and you'll see the possible options.

pschaumburg: @phiggins I've changed the logic based on `define_resource_requirements`.
For now, I've tested the general logic for that manually, but really want to learn and get some more background information about how I could add tests for it like I did before.

Maybe we can have a discussion on Slack regarding that? Just search for pschaumburg.

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -28,13 +28,40 @@ class WindowsFirewallRule < Chef::Resource
...
+        local_port '5985'
+        protocol 'TCP'
+        firewall_action :block
+        local_address '192.168.1.1'
+      end
+      ```
+
+      Deleting an existing rule
+      ```ruby
+      windows_firewall_rule 'Remove the SSH rule' do
```
tas50, line 53: can you add an example of the new functionality to this list

lib/chef/resource/windows_firewall_rule.rb
```diff
@@ -195,12 +228,53 @@ def firewall_command(cmdlet_type)
 
           cmd
         end
+
+        def define_resource_requirements
```
tas50, line 232: Traditionally we stick this above the actions.

pschaumburg: I would agree with that, but as I have tested this, it's not possible to move it out of the `action_class` block.
So it is in the same block like we have right now in the `locale` resource (https://github.com/chef/chef/blob/master/lib/chef/resource/locale.rb#L81)

lib/chef/resource/windows_firewall_rule.rb
pschaumburg: Example is added now.

tas50:
Nice fixes here. Thanks @pschaumburg

PR #9549, opened 2020-03-27
Skip the ifconfig functional tests if we don't have ifconfig

tas50:
I'm looking at you Ubuntu 20.04


tas50:
Merging based on slack convo

PR #9178, opened 2019-12-18
support Catalina softwareupdate output ("Label: ") and support updating CLI tools

w0de:
## Description
My proposed fix for #9177 

An example of Catalina output that this resource should support:

```
Software Update Tool

Finding available software
Software Update found the following new or updated software:
* Label: Command Line Tools for Xcode-11.2
	Title: Command Line Tools for Xcode, Version: 11.2, Size: 224942K, Recommended: YES,
* Label: Command Line Tools for Xcode-11.3
	Title: Command Line Tools for Xcode, Version: 11.3, Size: 224878K, Recommended: YES,
```

This resource will continue to work on Sierra and Mojave.

This PR also extracts the call to `softwareupdate` checking for packages, and uses that as a second criteria for deciding whether to install CL Tools. This has the effect of keeping the package updated by triggering a new install whenever one is available. I'm not sure if that's the behavior chef wants, however.

## Related Issue
#9177 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I have run the pre-merge tests locally and they pass.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

I'll need some guidance on where to put this test.
- [x] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #9550, opened 2020-03-30
build_essential resource: fix macOS 10.15 and improve installation support with a new :upgrade action for macOS

tas50:
This picks off the last DCO signed off portion of #9178 so we can recreate the conversion to native Ruby without using the code we can't accept due to license concerns.

### DIFF
--- a/lib/chef/resource/build_essential.rb
+++ b/lib/chef/resource/build_essential.rb
@@ -118,8 +118,8 @@ class Chef
         if macos?
           pkg_label = xcode_cli_package_label
 
-          # with upgrade action we should install if it's not install or if there's an available update to install
-          # xcode_cli_package_label will be nil if there's not update
+          # With upgrade action we should install if it's not installed or if there's an available update.
+          # `xcode_cli_package_label` will be nil if there's no update.
           install_xcode_cli_tools(pkg_label) if !xcode_cli_installed? || xcode_cli_package_label
         else
           Chef::Log.info "The build_essential resource :upgrade action is only supported on macOS systems. Skipping..."
--- a/lib/chef/resource/build_essential.rb
+++ b/lib/chef/resource/build_essential.rb
@@ -173,7 +173,7 @@ class Chef
           label_match = available_updates.stdout.match(/^\s*\* (?:Label: )?(Command Line Tools.*)/)
 
           # this will return the match or nil
-          label_match&.first
+          label_match[1] unless label_match.nil?
         end
       end
     end
--- a/lib/chef/resource/build_essential.rb
+++ b/lib/chef/resource/build_essential.rb
@@ -61,8 +61,8 @@ class Chef
           package "devel/m4"
           package "devel/gettext"
         when macos?
-          update_label = xcode_cli_package
-          unless xcode_cli_installed? && update_label.empty?
+          update_pkg_label = xcode_cli_package_label
+          unless xcode_cli_installed? && update_pkg_label.nil?
             # This script was graciously borrowed and modified from Tim Sutton's
             # osx-vm-templates at https://github.com/timsutton/osx-vm-templates/blob/b001475df54a9808d3d56d06e71b8fa3001fff42/scripts/xcode-cli-tools.sh
             execute "install XCode Command Line tools" do
@@ -71,7 +71,7 @@ class Chef
                 # in Apple's SUS catalog
                 touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
                 # find the CLI Tools update. We tail here because sometimes there's 2 and newest is last
-                PROD=#{update_label}
+                PROD=#{update_pkg_label}
                 # install it
                 softwareupdate -i "$PROD" --verbose
                 # Remove the placeholder to prevent perpetual appearance in the update utility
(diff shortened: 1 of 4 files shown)


phiggins:
As mentioned in standup, I have osx 10.14.6 and the output of `softwareupdate -l` looks like this:

```
Petes-MBP:chef pete$ softwareupdate -l
Software Update Tool

Finding available software
Software Update found the following new or updated software:
   * Security Update 2020-002-10.14.6
        Security Update 2020-002 (10.14.6), 1578350K [recommended] [restart]
```

tas50:
Thanks @phiggins that gives me an idea of the different for the parsing.

lib/chef/resource/build_essential.rb
```diff
@@ -61,7 +61,8 @@ class BuildEssential < Chef::Resource
           package "devel/m4"
           package "devel/gettext"
         when macos?
-          unless xcode_cli_installed?
+          update_label = xcode_cli_package
```
phiggins:
```suggestion
          xcode_version = xcode_cli_package
```

This variable name is kind of strange to me. `xcode_version`, maybe?

lib/chef/resource/build_essential.rb
```diff
@@ -70,7 +71,7 @@ class BuildEssential < Chef::Resource
                 # in Apple's SUS catalog
                 touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
                 # find the CLI Tools update. We tail here because sometimes there's 2 and newest is last
```
phiggins: This comment should get moved with the code below if it is still relevant.

lib/chef/resource/build_essential.rb
```diff
@@ -70,7 +71,7 @@ class BuildEssential < Chef::Resource
...
-                PROD=$(softwareupdate -l | grep "\*.*Command Line" | tail -n 1 | awk -F"*" '{print $2}' | sed -e 's/^ *//' | tr -d '\n')
+                PROD=#{update_label}
                 # install it
                 softwareupdate -i "$PROD" --verbose
```
phiggins:
```suggestion
                softwareupdate -i "#{xcode_version}" --verbose
```

This is a bit unnecessarily verbose.

lib/chef/resource/build_essential.rb
```diff
@@ -134,11 +135,23 @@ class BuildEssential < Chef::Resource
...
           # pkgutil returns an error if the package isn't found aka not installed
           cmd.error? ? false : true
         end
+
+        #
+        # Return to package label of the latest XCode Command Line Tools update, if available
+        #
+        # @return [String]
+        def xcode_cli_package
+          cmd = <<-EOH.gsub(/^ {14}/, "")
```
phiggins:
```suggestion
          cmd = <<-EOH.strip
```



Removing exactly 14 spaces from the front of this seems too specific (and also doesn't seem like it'll do anything here?). Using `.strip` will continue working if this code gets reindented, etc. There is also `.lstrip` if we want ot keep whitespace on the right side for whatever reason.

phiggins: There's one more occurrence of this in this file but I can't comment on it because it's not close enough to any lines modified in this PR.

lib/chef/resource/build_essential.rb
```diff
@@ -61,18 +61,17 @@ class BuildEssential < Chef::Resource
           package "devel/m4"
           package "devel/gettext"
         when macos?
-          unless xcode_cli_installed?
+          update_pkg_label = xcode_cli_package_label
```
phiggins: Since running `softwareupdate -l` is so slow, you could potentially save time in some cases by moving this method call inside the `unless` block on the next line.

tas50: I think the current behavior is also acting as an updater, which is wrong. We should have an :upgrade action that handles that scenario.

lib/chef/resource/build_essential.rb
```diff
@@ -128,16 +113,68 @@ class BuildEssential < Chef::Resource
...
       action_class do
         #
-        # Determine if the XCode Command Line Tools are installed
+        # Install Xcode Command Line tools via softwareupdate CLI
+        #
+        # @param [String] label The label (package name) to install
+        #
+        def install_xcode_cli_tools(label)
+          # This script was graciously borrowed and modified from Tim Sutton's
+          # osx-vm-templates at https://github.com/timsutton/osx-vm-templates/blob/b001475df54a9808d3d56d06e71b8fa3001fff42/scripts/xcode-cli-tools.sh
```
lamont-granquist: that comment is becoming less and less relevant.

lib/chef/resource/build_essential.rb
```diff
@@ -128,16 +113,68 @@ class BuildEssential < Chef::Resource
...
       end
 
+      action :upgrade do
+        description "Upgrade build essential (Xcode Command Line) tools on macOS"
+
+        if macos?
+          pkg_label = xcode_cli_package_label
+
+          # with upgrade action we should install if it's not install or if there's an available update to install
+          # xcode_cli_package_label will be nil if there's not update
```
phiggins:
```suggestion
          # With upgrade action we should install if it's not installed or if there's an available update.
          # `xcode_cli_package_label` will be nil if there's no update.
```

lib/chef/resource/build_essential.rb
```diff
@@ -128,16 +113,68 @@ class BuildEssential < Chef::Resource
...
+          available_updates = shell_out("softwareupdate", "--list")
+
+          # raise if we fail to check
+          available_updates.error!
+
+          # https://rubular.com/r/UPEE5P7mZLvXNs
+          label_match = available_updates.stdout.match(/^\s*\* (?:Label: )?(Command Line Tools.*)/)
+
+          # this will return the match or nil
+          label_match[1] unless label_match.nil?
```
phiggins:
```suggestion
          available_updates.stdout[/^\s*\* (?:Label: )?(Command Line Tools.*)/, 1]
```

This should be equivalent, but I've heard that this is not super readable for folks not familiar with ruby.

tas50: Used with a rubular link included

PR #9553, opened 2020-03-31
Fix cloud? helper to only report true on cloud instances

tas50:
Ohai always returns the cloud key so checking for the key isn't valid for cloud. We need to make sure it's not nil instead.

### DIFF
--- a/chef-utils/lib/chef-utils/dsl/cloud.rb
+++ b/chef-utils/lib/chef-utils/dsl/cloud.rb
@@ -30,7 +30,8 @@ module ChefUtils
       # @return [Boolean]
       #
       def cloud?(node = __getnode)
-        node.key?("cloud")
+        # cloud is always present, but nil if not on a cloud
+        !node["cloud"].nil?
       end
 
       # Return true if the current current node is in EC2.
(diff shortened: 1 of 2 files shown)


chef-utils/lib/chef-utils/dsl/cloud.rb
```diff
@@ -30,7 +30,8 @@ module Cloud
       # @return [Boolean]
       #
       def cloud?(node = __getnode)
-        node.key?("cloud")
+        # cloud is always present, but nil if not on a cloud
+        !node["cloud"].nil?
```
phiggins:
```suggestion
        !node["cloud"].nil?
```

I feel like `!!node["cloud"]` would be better in case whatever is setting this value ever sets it to an explicit `false`.

phiggins: Oh oops, I read further and saw in the tests that is a hash, not a boolean. In that case, maybe a check for `.empty?` is called for?

tas50: On a non-cloud system the value is nil though. The tests are giving the empty hashes to say "there would be some data here, but I'm not sticking 100 lines of cloud data into the spec"

phiggins: Gotcha 👍

lamont-granquist:
yeah this was probably always buggy on chef-sugar

PR #9562, opened 2020-04-01
Add after_resource to pair with current_resource

### LINKED ISSUE #6036: Implement `after_resource` that leverages `load_current_value`
The `current_value_does_not_exist!` API sets the `current_resource` to nil which seems bad.  The purpose of this is to have better checking around the object-does-not-exist-yet case.  So for example in the file resource now we build a current_resource but expect the mode, owner, group, etc to be 'nil' when the file simply doesn't exist yet.  It would seem to be better to build a default resource and then set a property like `exist` to false, and then have a `current_resource.exist?` helper to handle this case.  Right now I suspect that there are going to be lingering bugs in the resource reporter and data collector code (and possibly server-side in automate, etc) if anyone starts really using this feature and setting current_resource to nil.  I suspect that most people are not doing this.   The fact that it is hidden under the load_current_value and converge_if_changed api suggests that we can change this without changing the shape of the API.

I'd also like to work on creating an after_resource to solve all the horrible new_resource mutation issues we have where we're abusing the desired state with the final state and then trying to pass that into the resource_reporter/data_collector.  We really need to simply call a function to load an `after_resource` and let reporting diff `current_resource` and `after_resource` and then we can leave the `new_resource` alone.  In order to do that, though, we can't use `load_current_resource` because that function updates the current_resource as a side effect, we need to have a function which can load either the current_resource or the after_resource depending on how it is called.  The `load_current_value` function does that (and apparently even will optionally call itself with the resource as an argument).

#5974, #5980 are related, #5992 could probably use being looked at as well.

lamont-granquist:
For custom resources which correctly implement load_current_value the
new after_resource comes for "free" and load_current_value will be
called twice to load the current_resource (should be renamed the
"before_resource" but that ship sailed) and then the after_resource.

Appropriate wiring is added for a new event and capturing that into
the action_collection and then the data_collector.  The resource_reporter
has not and will not be updated.

For old style resources which are difficult to convert to custom
resources (thinking here particularly of the 1:N model of the service
resource in core chef, or stuff that just may be a lot of work like the
file resource) then they can override the load_after_resource to
manually wire up their own after_resource.

closes #6036

### DIFF
--- a/lib/chef/action_collection.rb
+++ b/lib/chef/action_collection.rb
@@ -1,5 +1,5 @@
 #
-# Copyright:: Copyright 2018-2019, Chef Software Inc.
+# Copyright:: Copyright 2018-2020, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -24,10 +24,6 @@ class Chef
 
     class ActionRecord
 
-      # XXX: this is buggy since we (ab)use this resource for "after" state and it may be
-      # inaccurate and it may be mutated by the user.  A third after_resource should be added
-      # to new_resource + current_resource to properly implement this.
-      #
       # @return [Chef::Resource] The declared resource state.
       #
       attr_accessor :new_resource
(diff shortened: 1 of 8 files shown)


tas50:
Is there a particular resource within chef that we'd convert to use this?

lamont-granquist:
Any custom resource that properly uses load_current_value gets it for free, so all the custom resources should really use it.

The service resource and the file resource can be fixed for all the janky garbage they do with the service resource's `running` and `enabled` properties, along with the file resource's `checksum` property.

PR #9568, opened 2020-04-01
Stop building S390x packages

tas50:
We will not be supporting these platforms for Chef 16


tas50:
@christopher-snapp Removed here as well.

PR #9570, opened 2020-04-02
Convert more resources to unified_mode

lamont-granquist:


lib/chef/resource/chef_vault_secret.rb
```diff
@@ -116,16 +118,14 @@ class ChefVaultSecret < Chef::Resource
       action :delete do
         description "Deletes the item and the item's keys ('id'_keys)."
 
-        converge_by("remove #{new_resource.id} and #{new_resource.id}_keys from #{new_resource.data_bag}") do
```
tas50, line 119: what's wrong with having a converge_by message here?

lamont-granquist: it isn't just a message and sets the resource to always being updated

tas50: 🤦‍♂ yep

lib/chef/provider/launchd.rb
```diff
@@ -89,24 +89,41 @@ def gen_path_from_type
       end
 
       def manage_plist(action)
-        return unless manage_agent?(action)
```
tas50, line 92: this resource terrifies me

lamont-granquist: yeah it should be somewhat better after this

lib/chef/resource/chef_vault_secret.rb
lamont-granquist: so once everything is unified_mode'd we can probably set up a log message to output nicely to the doc formatter to get the effect that everyone wants from converge_by without the side effects of updating the action.

that is another bit of "why we need unified_mode" ammo

tas50: I think I may have an issue from a few years ago asking for that. It would make log output a lot nicer when you have a ton of custom resource that contains dozens of out of the box resources under the hood and you want to all to just say "did foo" as if it was all magic.

PR #9581, opened 2020-04-02
Change the name_property to be the identity property by default, and to have desired_state: false by default

### LINKED ISSUE #4282: Make name_property imply identity
It is weird that this doesn't imply identity:

```
class X < Chef::Resource
  property :path, name_property: true
end
```

`name_attribute` doesn't do this, because an old resource written this way:

```
class X < Chef::Resource::LWRPBase
  attribute :path, name_attribute: true
end
```

Will _not_ be set as the identity attribute.

We can fix this by making `name_property` differ from `name_attribute` in this respect.

However, it should be a default, and this should work:

`property :path, name_property: true, identity: false`

lamont-granquist:
Setting `name_property: true` is now roughly equivalent to `name_property: true, identity: true, desired_state: false`.

If there is any other property which is marked as an identity though the name_property will no longer be considered an identity property.  A subtle detail of this is that the name_property itself is not marked as being `identity: true`, it is implemented via the `chef/mixin/properties.rb` mixin because the identity-ness of the name_property will be affected by other properties being marked as identity or not.  That means you have to inspect the whole property collection -- hence implemented in the properties mixin.  Consumers should check the identity_properties array on the resource class, and not inspect individual properties for being an identity themselves.

It makes no sense to have the name_property be set to being desired_state either since the name should not change before or after the resource converges.

closes #4282

### DIFF
--- a/lib/chef/resource/mount.rb
+++ b/lib/chef/resource/mount.rb
@@ -40,7 +40,7 @@ class Chef
         description: "Windows only:. Use to specify the password for username.",
         sensitive: true
 
-      property :mount_point, String, name_property: true, identity: false,
+      property :mount_point, String, name_property: true,
                description: "The directory (or path) in which the device is to be mounted. Defaults to the name of the resource block if not provided."
 
       property :device, String, identity: true,
--- a/lib/chef/resource/mount.rb
+++ b/lib/chef/resource/mount.rb
@@ -40,7 +40,7 @@ class Chef
         description: "Windows only:. Use to specify the password for username.",
         sensitive: true
 
-      property :mount_point, String, name_property: true,
+      property :mount_point, String, name_property: true, identity: false,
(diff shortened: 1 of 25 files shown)


tas50:
The resource specs seem to believe otherwise. That's the issue I ran into as I converted things to use the property DSL and why I had to add all these identity: true statements.

lamont-granquist:
oic, the "name" is the default identity property, not the "name_property"

lib/chef/resource/mount.rb
```diff
@@ -40,7 +40,7 @@ class Mount < Chef::Resource
         description: "Windows only:. Use to specify the password for username.",
         sensitive: true
 
-      property :mount_point, String, name_property: true,
+      property :mount_point, String, name_property: true, identity: false,
```
tas50, line 43: why does this one become false?

lamont-granquist: left over edit from an earlier version of the patch.  removed.

PR #9593, opened 2020-04-03
Implement eager_load_libraries metadata option

### LINKED ISSUE #3494: Implement RFC-040 and solve the oft requested recursive library loading problem
https://github.com/chef/chef-rfc/blob/master/rfc040-on-demand-cookbook-libraries.md

lamont-granquist:
This implements RFC-40 from the old chef rfc repo:

https://github.com/chef-boneyard/chef-rfc/blob/master/rfc040-on-demand-cookbook-libraries.md

Closes #3494

As far as I can see we have recursively loaded libraries for some time, this offers some better options to not load any library files but to push them into the LOAD_PATH or to push them into the LOAD_PATH and load a single "bootstrapping" hook.

This should also help with vendoring gems, although vendoring multiple gems with this mechanism isn't super straight forwards and we might want to have a top level first class "gems" segment for a cookbook that ships a pile of gems, along with their gemspecs in order to handle getting the gems activated correctly.

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -1,7 +1,7 @@
 # Author:: Adam Jacob (<adam@chef.io>)
 # Author:: AJ Christensen (<aj@chef.io>)
 # Author:: Seth Falcon (<seth@chef.io>)
-# Copyright:: Copyright 2008-2018, Chef Software Inc.
+# Copyright:: Copyright 2008-2020, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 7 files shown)


tas50:
This seems like a solid option as long as we don't plan on forcing this on folks later on. That seems like a pretty painful break for all the libraries out there.

lamont-granquist:
Uh, no.  Zero idea of ever flipping the default around, this should always be opt-in.  Most users shouldn't ever care, this is mostly for the one-percenters (like when btm asked me how to vendor a gem in a cookbook a few weeks back).

PR #9597, opened 2020-04-04
Skip the ifconfig functional tests if we don't have ifconfig

btm:
I'm looking at you Ubuntu 20.04

btm: this seems to fail in a surprising number of places. that smells funny.


btm:
more backports

PR #9605, opened 2020-04-05
14 backport: Improved Ruby download/install for functional tests

btm:
Adds more error checking and avoids trying to download Ruby if for some
reason it already exists on disk and matches the SHA256 hash.

The primary issue discovered when troubleshooting this script was that
aws.exe was running asynchronous. That was 'fixed' with '| Out-Null'.

Backport of #9603


btm:
Letting this one build to see if we can get a notarized package today.

PR #9606, opened 2020-04-05
14: fix some of the verify tests

btm:


scripts/bk_tests/bk_linux_exec.sh
```diff
@@ -39,10 +39,11 @@ echo "--- Installing Ruby 2.6.5"
 /opt/asdf/bin/asdf install ruby 2.6.5
```
tas50, line 39: while you're in here you might want to make this 2.6.6

btm:
Giving up on this. 14 goes EOL this month anyway.

PR #9608, opened 2020-04-06
Download and verify the Ruby installer, then move into place

btm:
This should reduce or remove race conditions while downloading the same
file on multiple container instances.

This cherry-picks 36d254fe1e2c2d0a274854ce627add1f622459b7 from https://github.com/chef/chef/pull/9604 to master.


btm:
I'm gonna let this one build to see if we get a notarized macos package today.

PR #9612, opened 2020-04-06
14: Really skip the reboot pending func test if a reboot is pending

btm:
This looks like it never worked but we happened to never run into it until we added a desktop platform to the pipeline.


btm:
backporting https://github.com/chef/chef/pull/9611

PR #9622, opened 2020-04-07
Add arm? helper to chef-utils

tas50:
It's tagged as first shipping in chef 15.10 since I plan to backport this and that'll make sense in the docs.

### DIFF
--- a/chef-utils/README.md
+++ b/chef-utils/README.md
@@ -108,17 +108,21 @@ The OS helpers provide an alternative to comparing data from `node['os']`.
 
 Architecture Helpers allow you to determine the processor architecture of your node.
 
-* `_64_bit?`
 * `_32_bit?`
+* `_64_bit?`
+* `aarch64?`
+* `arch64?`
+* `arm64?`
+* `arm?`
+* `armhf?`
 * `i386?`
 * `intel?`
-* `sparc?`
+* `powerpc?`
 * `ppc64?`
 * `ppc64le?`
-* `powerpc?`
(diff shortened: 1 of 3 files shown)


chef-utils/README.md
```diff
@@ -108,17 +108,21 @@ The OS helpers provide an alternative to comparing data from `node['os']`.
 
 Architecture Helpers allow you to determine the processor architecture of your node.
 
-* `_64_bit?`
 * `_32_bit?`
+* `_64_bit?`
+* `aarch64?`
+* `arch64?`
+* `arm64?`
```
lamont-granquist, line 115: these helpers don't actually exist (as the specs show)

tas50: Updated

chef-utils/lib/chef-utils/dsl/architecture.rb
```diff
@@ -103,6 +103,17 @@ def powerpc?(node = __getnode)
...
 
+      # Determine if the current architecture is arm
+      #
+      # @since 15.10
+      #
+      # @return [Boolean]
+      #
+      def arm?(node = __getnode)
+        %w{armhf aarch64 arm64 arch64}
+        .include?(node["kernel"]["machine"])
```
phiggins, line 114: Super nitpick but using two lines like this is kind of a weird way to write this statement.

tas50: True. Fixed

PR #9624, opened 2020-04-08
Add chef_client_systemd_timer resource

tas50:
This provides a resource for handling the base use case of running the chef-client binary under systemd using a timer. Why a timer and not a service? Running as a timer makes it significantly easier to update configs and handle updates. We never have to worry about Chef Infra Client restarting itself since the timer is a fresh instance running each time.

With nothing set this resource will run every 30 minutes with a 5-minute splay. This is a sensible default and that means with 1 line a user is ready to go. It provides the ability to modify the interval, modify the splay, use a custom config dir, use a custom chef-client binary path, disable runs when the system is on battery.

Things I plan to tackle in an upcoming PR:
- Killing long-running client runs which systemd supports natively
- Validating the interval inputs so we can make sure they're correct
- Failing gracefully when we're on a non-systemd system


lib/chef/resource/chef_client_systemd_timer.rb
```diff
@@ -48,43 +48,52 @@ class ChefClientSystemdTimer < Chef::Resource
...
       property :job_name, String,
-        default: Chef::Dist::CLIENT,
-        description: "The name of the system timer to create."
-      property :description, String, default: "Chef Infra Client periodic execution"
+        description: "The name of the system timer to create.",
+        default: Chef::Dist::CLIENT
+
+      property :description, String,
+        description: "The description to add to the systemd timer. This will be displayed when running `systemctl status` for the timer.",
+        default: "Chef Infra Client periodic execution"
```
phiggins, line 56: Does this need to be 
```suggestion
        default: "#{Chef::Dist::PRODUCT} periodic execution"
```

lib/chef/resource/chef_client_systemd_timer.rb
```diff
@@ -48,43 +48,52 @@ class ChefClientSystemdTimer < Chef::Resource
...
-        default: 300,
-        coerce: proc { |x| Integer(x) },
-        callbacks: { "should be a positive number" => proc { |v| v > 0 } },
-        description: "A random number of seconds between 0 and X to add to interval so that all #{Chef::Dist::CLIENT} commands don't execute at the same time."
+      property :delay_after_boot, String,
+        description: "The time to wait after booting before the interval starts. This is expressed as a systemd time span such as `300seconds`, `1hr`, or `1m`. See the `systemd.time` man page for a complete list of allowed time span values.",
+        default: "1min"
+
+      property :interval, String,
+        description: "The interval to wait between execution. This is expressed as a systemd time span such as `300seconds`, `1hr`, or `1m`. See the `systemd.time` man page for a complete list of allowed time span values.",
```
phiggins, line 67:
```suggestion
        description: "The interval to wait between executions. This is expressed as a systemd time span such as `300seconds`, `1hr`, or `1m`. See the `systemd.time` man page for a complete list of allowed time span values.",
```

tas50: Fixed outside review

PR #9633, opened 2020-04-08
Bump highline to 2.x

lamont-granquist:
We need 2.x for ruby 2.7 compat and warning suppresion

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -41,7 +41,7 @@ PATH
       ffi (~> 1.9, >= 1.9.25)
       ffi-libarchive
       ffi-yajl (~> 2.2)
-      highline (>= 1.6.9, < 2)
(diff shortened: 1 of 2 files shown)


tas50:
Do we know that we're not going to get burned by any of the deprecations they introduced in 2.0?

lamont-granquist:
uh, nope, but we're definitely going to get burned by all kinds of ruby 2.7 warning spam if we don't do this.

lamont-granquist:
```
/Users/lamont/.asdf/installs/ruby/2.7.0/lib/ruby/gems/2.7.0/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
```

tas50:
I did some local testing on this and it looks like we're not impacted by the breaking changes in 2.0

Issue #9639, opened 2020-04-09
ssh: ed25519 keys fail to auth from Windows host. 


marcparadise:
## Description
When using a command such as `knife bootstrap` or `knife ssh` from a Windows host and providing an ed25519 key,  it fails as follows:

```
 C:\opscode\chef> .\bin\knife ssh -m 192.168.1.133 -x marc -i C:\Users\marc\Documents\id_ed25519
 - snip - 
OpenSSH keys only supported if ED25519 is available (NotImplementedError)
net-ssh requires the following gems for ed25519 support:
 * ed25519 (>= 1.2, < 2.0)
 * bcrypt_pbkdf (>= 1.0, < 2.0)
See https://github.com/net-ssh/net-ssh/issues/565 for more information
```

These gems are [both installed](https://github.com/chef/chef/blob/master/chef.gemspec#L36-L37) in the chef package; however on Windows the installation only includes platform native extension for rubies 2.0-2.3, and fails to load a non-existent 2.6 extension

This is because the gem download only contained native extensions for ruby 2.0-2-3.  The PR linked in comments extends this to 2.0->2.7.

 
## Chef Version
15.9.17 (earlier too) 

## Platform Version

Windows.* 

## Replication Case
1. Install chef-client
2. Generate an ed25519 key
3. Use this key to authenticate to a known-good host using `knife ssh`: 
`knife ssh -m HOST -x USER -i ED25519-KEY-PATH

## Client Output

See description.

marcparadise:
This will be fixed by this PR, all that should be necessary is to bump the version we pull in: https://github.com/mfazekas/bcrypt_pbkdf-ruby/pull/2

lamont-granquist:
1.1.0.rc1 is released

tas50:
@marcparadise Can you test that using 1.1.0rc1 would fix this?

PR #9640, opened 2020-04-09
Add software def for bcrypt_pbkdf-ruby

marcparadise:
Fixes an issue where the named gem does not build a compatible
native extension  for Windows when it's installed indirectly via `bundle
install` of the `chef` gem

fixes: https://github.com/chef/chef/issues/9639

### DIFF
--- a/omnibus/config/projects/chef.rb
+++ b/omnibus/config/projects/chef.rb
@@ -47,6 +47,10 @@ instance_eval(IO.read(overrides_path), overrides_path)
 
 dependency "preparation"
 
+dependency "bcrypt_pbkdf-ruby" # In order for the correct windows extension to build, this gem must be
+                               # built directly, before we build the chef gem.
+                               # see https://github.com/chef/chef/issues/9639 for details.
+
 dependency "chef"
 
 #
--- a/omnibus/config/software/bcrypt_pbkdf-ruby.rb
+++ b/omnibus/config/software/bcrypt_pbkdf-ruby.rb
@@ -1,0 +1,39 @@
+#
+# Copyright:: 2020 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


marcparadise:
Note that `omnibus-software` does contain a definition for (bcrypt_pbkdf-ruby)[https://github.com/chef/omnibus-software/blob/master/config/software/bcrypt_pbkdf-ruby.rb] - that's where this was taken from - but the `--without-development` flag causes it to fail (native build needs the rake-compiler gem, which is declared as a dev dep).

I did not want to change the shared software def, as removing that flag could cause existing users of it to start pulling in things they did not expect.  If that's not a concern, we can update the original definition and skip adding one here.

marcparadise:
Ad-hoc pipeline build is unhappy about bundler version incompatibility: 
```
/opt/chef/embedded/lib/ruby/2.7.0/rubygems.rb:275:in `find_spec_for_exe': Could not find 'bundler' (1.17.3) required by your /var/cache/omnibus/chef/src/bcrypt_pbkdf-ruby/bcrypt_pbkdf/Gemfile.lock. (Gem::GemNotFoundException)
```

tas50:
2.1.4 is built into Ruby 2.7 so if this is failing here it's going to fail in Workstation this month as well

lamont-granquist:
Doing the bundle install here is fairly terrible for numerous reasons.  If all `rake gem` does is run `gem build` then you should probably just do that here.

lamont-granquist:
The gem seems to have zero runtime deps, so the bundle install is only going to be polluting your shipping artifact with duplicated gems.

lamont-granquist:
Oh, so you should probably just declare the `rake-compiler` gem in your Gemfile manually and then most likely just ship it.  If `rake-compiler` itself has dependencies which are problematic, then you could manually `gem install` that first outside of the bundle I suppose but then you wind up shipping dup unnecessary gems anyway.  The best solution though is to just get it into the Gemfile.  The author should really add `rake-compiler` to the runtime deps of the gem so that it becomes available at install time, which effectively would be the same as you adding it to the Gemfile.

lamont-granquist:
I also don't see any value to adding a software definition here.

Dependent project either need to add a software definition or need to manually add the dep to the Gemfile.  They should do the simpler thing and manually add the dep to the Gemfile.  We shouldn't violate the rule that we don't carry any rubygem software definitions, and we shouldn't violate the rule that we only do one bundle install.  Every time we violate that rule (always with the best of intentions) I seem to wind up having to clean it up later.

marcparadise:
Thanks @lamont-granquist and @tas50 

There's now an upstream PR that the maintainer is taking a look at; it's a better fix than this Chef PR: 

https://github.com/mfazekas/bcrypt_pbkdf-ruby/pull/2

PR #9646, opened 2020-04-10
Knife bootstrap options cleanup

### LINKED ISSUE #5425: remove config[:ssh_password_ng] from knife ssh
config[:ssh_password] should be changed to work like config[:ssh_password_ng] and the latter should be dropped

Chef 13 breaking change.

### LINKED ISSUE #8394: Improved Knife Config Merge
## Motivation

    As a knife maintainer,
    I want to only worry sourcing my config values from one place,
    So that I spend less time trying to figure out where config values are coming from.

## Specification

In `Knife#merge_configs` - invoked before any `knife` command runs - we currently merge Chef::Config and CLI options as follows:

* start with default options from CLI flags
* merge in populated user-knife config, constrained by available CLI flags
* merge in CLI options

However, in some cases commands will use values of out `Chef::Config[:knife]` that don't have a counterpart in the command's CLI options. For example, `Chef::Config[:knife][:architecture]`, referenced in the Windows bootstrap template. In the current merge behavior, this scenario means `Chef::Config[:knife]` values that don't match up to CLI flags will not be available in `Knife#config`, requiring the command to use `Chef::Config` as well.

Under this change, the merge behavior will be as follows:

* start with default options from CLI flags
* merge in all user-knife config, without constraints
* merge in CLI options

This will result in all Chef::Config[:knife] contents being available through `Knife#config`, without 
In the soon-to-be-merged chef#1843, we also introduced `config_source(key)`; this will be modified to check config hashes in merge-precedence to return the source of the config value. This will replace its current behavior of annotating knife- and cli-sourced config in separate hashes.

## Downstream Impact

Because this is only expanding the values present in `Knife#config` and we're preserving precedence, there are no effects to consumers of either `Knife#config` or `Chef::Config[:knife]`.

It is an existing pattern to set knife configuration directly when handling CLI options [(example)](https://github.com/chef/chef/blob/SUSTAINING-955/bootstrap-on-chef-core/lib/chef/knife/supermarket_search.rb#L27-L32). 

This is all done before the merge.  When `Config[:knife]` is modified before the merge, `Knife#config` will resolve correctly using the ordering above.

In some cases, multiple `Config[:knife]` keys can be set in response to one flag. These will also be present in the merged configuration.  If a command is setting keys that also correspond to CLI flags that were provided, the result will be unpredictable, depending on the order of evaluation of the flags. However, commands that do this tend to look explicitly in `Config[:knife]` to resolve their config to begin with - if they don't, their results are already unpredictable. 

## Additional References 

https://github.com/chef/chef/issues/3946

lamont-granquist:
We have issue that are caused by old code before merging of hash values
were done correctly.

The `config` hash correctly merges all options and should always be
used.

Knife plugins should never touch Chef::Config[:knife] values (either
reading or writing from them).

The `knife_config` should be converted to the `config` hash since it
directly accesses Chef::Config[:knife] values.

The `config_value()` helper should no longer be used.  Very clearly most
people started to use that when they should just use the config hash
directly.  That was intended to be used only when a knife cli option
was being renamed and the former configuration value needed to be
used as well.  It has been cargo culted around as the way to access
config values, and that should really stop.

The DataBagSecretOption mixin has been cleaned up so that the cli
options read+write only to the config[:cl_secret] and
config[:cl_secret_file] values.  The config file values go into
config[:secret] and config[:secret_file].  The fact that those are
the merged values in the `config` hash doesn't matter since only
the cli should be writing to the first two and only the config
file should be writing to the latter two.  I don't know why it was
made so complicated to begin with, but if there's some hidden
chef-11.early backcompat there, then chef-16 deliberately breaks that.

The use of `locate_config_value` helpers in all knife plugins is also
discouraged (but they all implement those themselves), just use the
config hash, which has the correct hash merge ordering.  All of those
need to be deleted.

Closes #8394

Bonus fix, removes the ssh_password_ng nightmare.

This means that knife plugins that use that ssh_password option need to be updated to have a `default: false` or else they will always prompt for a password, even if the flag was not given on the CLI.  For our plugins this looks like it only affects knife-windows.

Closes #5425

### DIFF
--- a/lib/chef/knife.rb
+++ b/lib/chef/knife.rb
@@ -343,31 +343,35 @@ class Chef
       exit(1)
     end
 
-    # keys from mixlib-cli options
-    def cli_keys
-      self.class.options.keys
+    # This is all set and default mixlib-config values.  We only need the default
+    # values here (the set values are explicitly mixed in again later), but there is
+    # no mixlib-config API to get a Hash back with only the default values.
+    #
+    # Assumption:  since config_file_defaults is the lowest precedence it doesn't matter
+    # that we include the set values here, but this is a hack and makes the name of the
+    # method a lie.  FIXME: make the name not a lie by adding an API to mixlib-config.
+    #
+    # @api private
+    #
+    def config_file_defaults
+      Chef::Config[:knife].save(true) # this is like "dup" to a (real) Hash, and includes default values (and user set values)
     end
 
-    # extracts the settings from the Chef::Config[:knife] sub-hash that correspond
-    # to knife cli options -- in preparation for merging config values with cli values
+    # This is only the user-set mixlib-config values.  We do not include the defaults
+    # here so that the config defaults do not override the cli defaults.
     #
-    # NOTE: due to weirdness in mixlib-config #has_key? is only true if the value has
-    # been set by the user -- the Chef::Config defaults return #has_key?() of false and
-    # this code DEPENDS on that functionality since applying the default values in
-    # Chef::Config[:knife] would break the defaults in the cli that we would otherwise
-    # overwrite.
+    # @api private
+    #
     def config_file_settings
-      cli_keys.each_with_object({}) do |key, memo|
-        if Chef::Config[:knife].key?(key)
-          memo[key] = Chef::Config[:knife][key]
-        end
-      end
+      Chef::Config[:knife].save(false) # this is like "dup" to a (real) Hash, and does not include default values (just user set values)
     end
 
     # config is merged in this order (inverse of precedence)
-    #  default_config       - mixlib-cli defaults (accessor from the mixin)
-    #  config_file_settings - Chef::Config[:knife] sub-hash
-    #  config               - mixlib-cli settings (accessor from the mixin)
+    #  config_file_defaults - Chef::Config[:knife] defaults from chef-config (XXX: this also includes the settings, but they get overwritten)
+    #  default_config       - mixlib-cli defaults (accessor from mixlib-cli)
+    #  config_file_settings - Chef::Config[:knife] user settings from the client.rb file
+    #  config               - mixlib-cli settings (accessor from mixlib-cli)
+    #
     def merge_configs
       # Update our original_config - if someone has created a knife command
       # instance directly, they are likely ot have set cmd.config values directly
@@ -375,7 +379,7 @@ class Chef
       @original_config = config.dup
       # other code may have a handle to the config object, so use Hash#replace to deliberately
       # update-in-place.
-      config.replace(default_config.merge(config_file_settings).merge(config))
+      config.replace(config_file_defaults.merge(default_config).merge(config_file_settings).merge(config))
     end
 
     #
@@ -385,8 +389,9 @@ class Chef
     # @return [Symbol,NilClass] return the source of the config key,
     # one of:
     #   - :cli - this was explicitly provided on the CLI
-    #   - :config - this came from Chef::Config[:knife]
+    #   - :config - this came from Chef::Config[:knife] explicitly being set
     #   - :cli_default - came from a declared CLI `option`'s `default` value.
+    #   - :config_default - this came from Chef::Config[:knife]'s defaults
     #   - nil - if the key could not be found in any source.
     #           This can happen when it is invalid, or has been
     #           set directly into #config without then calling #merge_config
@@ -394,6 +399,7 @@ class Chef
       return :cli if @original_config.include? key
       return :config if config_file_settings.key? key
       return :cli_default if default_config.include? key
+      return :config_default if config_file_defaults.key? key # must come after :config check
 
       nil
     end
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -86,7 +86,6 @@ class Chef
         short: "-w AUTH-METHOD",
         long: "--winrm-auth-method AUTH-METHOD",
         description: "The WinRM authentication method to use.",
-        proc: Proc.new { |protocol| Chef::Config[:knife][:winrm_auth_method] = protocol },
         in: WINRM_AUTH_PROTOCOL_LIST
 
       option :winrm_basic_auth_only,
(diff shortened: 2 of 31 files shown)


lamont-granquist:
hmm, don't merge this yet, buggy.

lamont-granquist:
still need to look this over for another couple days

lib/chef/knife/core/generic_presenter.rb
```diff
@@ -1,6 +1,6 @@
 #--
 # Author:: Daniel DeLeo (<dan@chef.io>)
-# Copyright:: Copyright 2011-2016, Chef Software, Inc.
```
tas50, line 3: your script went off the rails a bit here

lib/chef/knife.rb
```diff
@@ -343,39 +343,38 @@ def parse_options(args)
...
-      cli_keys.each_with_object({}) do |key, memo|
-        if Chef::Config[:knife].key?(key)
-          memo[key] = Chef::Config[:knife][key]
-        end
-      end
+      Chef::Config[:knife].save(false)
     end
 
     # config is merged in this order (inverse of precedence)
+    #  base_mixlib_config   - Chef::Config[:knife] defaults
```
tas50, line 365: This method could really use a more clear name. I get the order after reading it here, but seeing these methods used it's really not apparent where the config data is coming from

lib/chef/knife.rb
```diff
@@ -385,15 +384,17 @@ def merge_configs
     # @return [Symbol,NilClass] return the source of the config key,
     # one of:
     #   - :cli - this was explicitly provided on the CLI
-    #   - :config - this came from Chef::Config[:knife]
+    #   - :config - this came from Chef::Config[:knife] explicitly being set
```
tas50, line 387: that comment makes this 10x more clear. Thanks

lib/chef/knife/bootstrap.rb
```diff
@@ -309,26 +300,22 @@ class Bootstrap < Knife
       # bootstrap override: Do this instead of our own setup.sh from omnitruck. Causes bootstrap_url to be ignored.
       option :bootstrap_install_command,
         long: "--bootstrap-install-command COMMANDS",
-        description: "Custom command to install #{Chef::Dist::PRODUCT}.",
-        proc: Proc.new { |ic| Chef::Config[:knife][:bootstrap_install_command] = ic }
```
tas50, line 313: Best salmon ever here

lib/chef/knife/core/generic_presenter.rb
lamont-granquist: no, i got sick of the years, time for them to go away.

tas50: We should just nuke them all then

lib/chef/knife.rb
lamont-granquist: Yeah I'm open to a better name but I'm also at a loss.  I would have used `default_config` because that's what its supposed to be, but mixlib-cli already took that.  So I went with `default_mixlib_config`, but then its not really just the defaults, although that's the _intent_ but it actually includes all of mixlib-config.  So I went with "base_mixlib_config" because its really the base config that everything is layered on.  I hate it all.

lib/chef/knife/bootstrap.rb
```diff
@@ -262,21 +262,16 @@ class Bootstrap < Knife
...
-      # but only from the passed-in knife config; it does not use the
-      # config from the CLI for those values.  We cannot always used the merged
-      # config, because in some cases the knife keys thIn those cases, the option
-      # will have a proc that assigns the value into Chef::Config[:knife]
-
       # bootstrap template
       # Create ohai hints in /etc/chef/ohai/hints, fname=hintname, content=value
-      option :hint,
+      option :hints,
         long: "--hint HINT_NAME[=HINT_FILE]",
```
tas50, line 268: The long here doesn't match

lib/chef/knife/bootstrap/templates/chef-full.erb
```diff
@@ -164,20 +164,20 @@ do_download() {
...
+<% if @config[:bootstrap_install_command] %>
+  <%= @config[:bootstrap_install_command] %>
 <% else %>
-  install_sh="<%= config[:bootstrap_url] ? config[:bootstrap_url] : "https://omnitruck.chef.io/chef/install.sh" %>"
+  install_sh="<%= @config[:bootstrap_url] ? @config[:bootstrap_url] : "https://omnitruck.chef.io/chef/install.sh" %>"
   if test -f /usr/bin/<%= Chef::Dist::CLIENT %>; then
     echo "-----> Existing <%= Chef::Dist::PRODUCT %> installation detected"
   else
-    echo "-----> Installing Chef Omnibus (<%= @config[:channel] %>/<%= version_to_install %>)"
+    echo "-----> Installing Chef Omnibus (<%= @@config[:channel] %>/<%= version_to_install %>)"
```
tas50, line 178: Is the @@config intentional here?

lamont-granquist: lol no

lib/chef/knife/bootstrap.rb
```diff
@@ -270,30 +263,24 @@ class Bootstrap < Knife
...
         description: "Specify an Ohai hint to be set on the bootstrap target. Use multiple --hint options to specify multiple hints.",
-        proc: Proc.new { |h|
-          Chef::Config[:knife][:hints] ||= {}
-          name, path = h.split("=")
-          Chef::Config[:knife][:hints][name] = path ? Chef::JSONCompat.parse(::File.read(path)) : {}
+        proc: Proc.new { |hint, accumulator|
+          accumulator ||= {}
+          name, path = hint.split("=")
```
phiggins, line 273:
```suggestion
          name, path = hint.split("=", 2)
```

nitpick: The value on the right hand side could have an `=` in it.

lamont-granquist:
Notes to myself.  We need to "backport" to Chef-15:

- The code which merges the configs to pick up all of the keys and the Chef::Config defaults correctly.

- The change to accept the config keyword option (but not deprecate the knife_config keyword option) to the bootstrap client_builder / chef_vault_handler.

- Ensure that a falsey value as a default to config[:ssh_password] will not blow up on Chef-15 and should resolve
to nil (should be as simple as backporting the fix to get_stripped_unfrozen_value().

lib/chef/knife/data_bag_secret_options.rb
```diff
@@ -98,41 +94,26 @@ def validate_secrets
...
-      def has_cl_secret_file?
-        Chef::Config[:knife].key?(:cl_secret_file)
-      end
-
-      def self.set_cl_secret_file(sf)
-        Chef::Config[:knife][:cl_secret_file] = sf
-      end
-
       def knife_config
+        Chef.deprecated(:knife_bootstrap_apis, "The knife_config bootstrap helper has been deprecated, use the properly merged config helper instead")
```
phiggins, line 116:
```suggestion
        Chef.deprecated(:knife_bootstrap_apis, "The `knife_config` bootstrap helper has been deprecated, use the properly merged `config` helper instead")
```

I could see this message being unclear if you weren't aware `config` referred to a method

lamont-granquist: It pretty clearly suggests `knife_config` being rewritten to `config` as a solution though?

lamont-granquist: After this is merged there needs to be a deprecation page added which describes it in a lot more words and gives examples and remediations, which should make it clearer.

phiggins: What I mean is if I'm unfamiliar with chef I could read "use the properly merged config helper instead" not realizing that `config` is a method that I could call. It reads like "instead of doing the bad thing, call the correct method instead".

lamont-granquist: Still not quite getting it, think I'm too tired today.  What's your suggestion for a replacement?

Issue #9450, opened 2020-03-06
Update the locale resource to support Windows


tas50:
### Describe the Enhancement:
Update the locale resource to support Windows by using Get-WinSystemLocale and Set-WinSystemLocale.

### Describe the Need:
This is something people would want to configure on Windows and we can add it with minimal effort.

### Current Alternative
powershell_script can be used right now.

PR #9648, opened 2020-04-10
locale: Support setting locale on Windows

kapilchouhan99:

## Description
- Now users can set the system locale on Windows with the help of locale resource.
- Added test cases

## Related Issue
Fixes: #9450 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/locale.rb
+++ b/lib/chef/resource/locale.rb
@@ -30,7 +30,7 @@ class Chef
       LC_VARIABLES = %w{LC_ADDRESS LC_COLLATE LC_CTYPE LC_IDENTIFICATION LC_MEASUREMENT LC_MESSAGES LC_MONETARY LC_NAME LC_NUMERIC LC_PAPER LC_TELEPHONE LC_TIME}.freeze
       LOCALE_CONF = "/etc/locale.conf".freeze
       LOCALE_REGEX = /\A\S+/.freeze
-      LOCALE_PLATFORM_FAMILIES = %w{debian}.freeze
+      LOCALE_PLATFORM_FAMILIES = %w{debian windows}.freeze
 
       property :lang, String,
         description: "Sets the default system language.",
@@ -65,13 +65,32 @@ class Chef
         end
       end
 
+      load_current_value do
+        if windows?
+          lang get_system_locale_windows
+        else
+          begin
+            old_content = ::File.read(LOCALE_CONF)
+            locale_values = Hash[old_content.split("\n").map { |v| v.split("=") }]
+            lang locale_values["LANG"]
+          rescue Errno::ENOENT => e
+            false
+          end
+        end
+      end
+
+      # Gets the System-locale setting for the current computer.
+      # @see https://docs.microsoft.com/en-us/powershell/module/international/get-winsystemlocale
+      # @return [String] the current value of the System-locale setting.
+      #
+      def get_system_locale_windows
+        powershell_exec("Get-WinSystemLocale").result["Name"]
+      end
+
       action :update do
         description "Update the system's locale."
-        unless up_to_date?
-          converge_by "Updating System Locale" do
-            generate_locales unless unavailable_locales.empty?
-            update_locale
-          end
+        converge_if_changed do
+          set_system_locale
         end
       end
 
@@ -97,6 +116,21 @@ class Chef
         #
         def generate_locales
           shell_out!("locale-gen #{unavailable_locales.join(" ")}")
+        end
+
+        # Sets the system locale for the current computer.
+        #
+        def set_system_locale
+          if windows?
+            # Sets the system locale for the current computer.
+            # @see https://docs.microsoft.com/en-us/powershell/module/internationalcmdlets/set-winsystemlocale
+            #
+            response = powershell_exec("Set-WinSystemLocale -SystemLocale #{new_resource.lang}")
+            raise response.errors.join(" ") if response.error?
+          else
+            generate_locales unless unavailable_locales.empty?
+            update_locale
+          end
         end
 
         # Updates system locale by appropriately writing them in /etc/locale.conf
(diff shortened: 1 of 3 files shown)


Vasu1105:
@kapilchouhan99 looks it might not be idempotent.

lib/chef/resource/locale.rb
```diff
@@ -99,6 +104,15 @@ def generate_locales
...
         end
 
+        # Windows only: Sets the system locale for the current computer.
+        # @see https://docs.microsoft.com/en-us/powershell/module/internationalcmdlets/set-winsystemlocale
+        #
+        def set_system_locale
+          response = powershell_exec("Set-WinSystemLocale -SystemLocale #{new_resource.lang}")
+
+          raise response.errors.join(" ") if response.error?
+        end
```
Vasu1105, line 114: @kapilchouhan99 looks like it might not be idempotent. Considering it only running Set query and not checking existing values,

Vasu1105:
Also make a point this only support language part for Windows and not lc_env so this will require document update.

lib/chef/resource/locale.rb
Vasu1105: We should have idempotency check

lib/chef/resource/locale.rb
```diff
@@ -1,3 +1,4 @@
+
```
tas50, line 1:
```suggestion
```

lib/chef/resource/locale.rb
```diff
@@ -69,8 +70,12 @@ def lc_all(arg = nil)
         description "Update the system's locale."
         unless up_to_date?
           converge_by "Updating System Locale" do
-            generate_locales unless unavailable_locales.empty?
-            update_locale
+            if ChefUtils.windows?
```
tas50, line 73:
```suggestion
            if windows?
```

lib/chef/resource/locale.rb
```diff
@@ -69,8 +70,12 @@ def lc_all(arg = nil)
         description "Update the system's locale."
         unless up_to_date?
           converge_by "Updating System Locale" do
```
tas50, line 72: This resource needs a bit of modernizing. If we load the current value of :lang then we can use a converge_if_changed to update the value. That will provide us with proper idempotency and allows us to remove the up_to_date? method. That method just becomes the Linux side of the logic that loads the :lang value for current_resource.

kapilchouhan99: @tas50 In `up_to_date?` method we are comparing `/etc/locale.conf` content with the new contain Whether any modification is required in `/etc/locale.conf`

lib/chef/resource/locale.rb
```diff
@@ -67,10 +67,18 @@ def lc_all(arg = nil)
...
-            generate_locales unless unavailable_locales.empty?
-            update_locale
+        if windows?
+          unless new_resource.lang == get_system_locale
+            converge_if_changed :lang do
+              set_system_locale
+            end
+          end
+        else
+          unless up_to_date?
```
lamont-granquist, line 77: there shouldn't be a `if windows?` check at this higher level.  the `up_to_date?` method should really work on both windows and unix, and the `if windows`? check should be pushed into that method.

similarly there should probably be a single `update_locale` or `set_system_locale` or whatever that worked on both windows and unix (and the generate_locales/update_locale complexity should be pushed down into the details of that method).

lib/chef/resource/locale.rb
```diff
@@ -67,10 +67,18 @@ def lc_all(arg = nil)
 
       action :update do
         description "Update the system's locale."
-        unless up_to_date?
-          converge_by "Updating System Locale" do
-            generate_locales unless unavailable_locales.empty?
-            update_locale
+        if windows?
+          unless new_resource.lang == get_system_locale
+            converge_if_changed :lang do
```
lamont-granquist, line 72: `converge_if_changed` here isn't going to work if there's no load_current_value, this should be `converge_by` (and the conditional above `converge_if_changed` doesn't make any sense, either the pattern should be to use a conditional with `converge_by` or to use `converge_if_changed` without a conditional and with load_current_value).

kapilchouhan99: @lamont-granquist `new_resource.lang == get_system_locale` this condition checks if the user supplied the same `lang` value which is already set in the current system, so on that time, it should not be run `set_system_locale` method.

lib/chef/resource/locale.rb
```diff
@@ -99,6 +98,29 @@ def generate_locales
           shell_out!("locale-gen #{unavailable_locales.join(" ")}")
         end
 
+        # Windows only: Gets the System-locale setting for the current computer.
+        # @see https://docs.microsoft.com/en-us/powershell/module/international/get-winsystemlocale
+        #
+        def get_system_locale
+          response = powershell_exec("Get-WinSystemLocale")
```
tas50, line 105: This could all be done on one line without setting a variable

lib/chef/resource/locale.rb
```diff
@@ -99,6 +98,29 @@ def generate_locales
           shell_out!("locale-gen #{unavailable_locales.join(" ")}")
         end
 
+        # Windows only: Gets the System-locale setting for the current computer.
+        # @see https://docs.microsoft.com/en-us/powershell/module/international/get-winsystemlocale
+        #
+        def get_system_locale
```
tas50, line 104: if this is for windows only we should just name it `get_system_locale_windows`

lib/chef/resource/locale.rb
```diff
@@ -99,6 +98,29 @@ def generate_locales
           shell_out!("locale-gen #{unavailable_locales.join(" ")}")
         end
 
+        # Windows only: Gets the System-locale setting for the current computer.
+        # @see https://docs.microsoft.com/en-us/powershell/module/international/get-winsystemlocale
+        #
```
tas50, line 103: Can you add a yard comment that gives the type returned here

tas50:
So what you really want is something more like the windows_firewall resource:

https://github.com/chef/chef/blob/master/lib/chef/resource/windows_firewall_rule.rb

You'll want to get the lang resource for current_resource by powershelling out on windows or parsing the file on Linux. Then you can just use converge_if_changed(:lang) like is done in windows_firewall. That avoids a lot of the per platform logic here and is the correct way to load state and then change a system if the state doesn't match the desired state.

Issue #9383, opened 2020-02-20
msu_package fails to install cumulative updates correctly


gsreynolds:
## Description
Similar to https://github.com/chef/chef/issues/6091 (and a possible regression from the fix in https://github.com/chef/chef/pull/6100), on a Windows 2016 server with a single previous Cumulative Update package installed attempting to install a later Cumulative Update package does nothing as the resource is marked up to date.

Additionally, on a long lived Windows 2016 server with multiple Cumulative Update packages installed (where the previous packages have been marked `Superseded` in DISM) attempting to install a further Cumulative Update package results in the error message: `Found multiple packages installed matching name Package_for_RollupFix, found: 4 matches`. We are hitting a branch of the code that it was assumed would not be encountered, but it is in fact possible to reach https://github.com/chef/chef/blob/28653b9a54427dc295d8f3be31b032bc556e7b59/lib/chef/provider/package/cab.rb#L102-L103

Example output from DISM:
```
Package Identity : Package_for_KB4521858~31bf3856ad364e35~amd64~~14393.3262.1.0
State : Installed
Release Type : Security Update
Install Time : 10/9/2019 6:52 AM

Package Identity : Package_for_RollupFix~31bf3856ad364e35~amd64~~14393.3326.1.5
State : Superseded
Release Type : Security Update
Install Time : 11/13/2019 7:08 AM

Package Identity : Package_for_RollupFix~31bf3856ad364e35~amd64~~14393.3504.1.3
State : Installed
Release Type : Security Update
Install Time : 2/17/2020 8:03 PM
```

## Chef Version
Chef Infra Client 15.8.23

## Platform Version
Windows 2016

## Replication Case

kitchen.yml extract (AMI with the November 12, 2019—KB4525236 Cumulative Update installed
```yaml
platforms:
  - name: windows-2016
    driver:
      image_search: 
        owner-id: 801119661308
        name: Windows_Server-2016-English-Full-Base-2019.11.13
```

recipe
```ruby
windows_updates = [
  # Latest Servicing stack update
  { kb: 'KB4520724', source: 'http://download.windowsupdate.com/c/msdownload/update/software/secu/2019/11/windows10.0-kb4520724-x64_97604f0b532d6da814b4120fc43b2d9f6fd0b356.msu' },
  # February 11, 2020—KB4537764
[...]

Issue #9245, opened 2020-01-21
timeout property not honored by msu_package


danielfloyd:
# Describe the problem

While provisioning a Windows 8.1 x64 vagrant box using packer and the chef-solo provisioner, I am attempting to install the Windows 8.1 Update 1 update using `msu_package` from a recipe. The update is large and times out after the default 600 seconds. When I add a `timeout` property to my `msu_package` definition, it still times out after 600 seconds.

Looking in the code, it appears the issue is because `msu_package` uses `cab_package` to actually do the work and it is not passing its `timeout` property down. See here: https://github.com/chef/chef/blob/f154e2a113499b75ec48ea4118aa65292834b8f5/lib/chef/provider/package/msu.rb#L110

The expectation is that setting a `timeout` on `msu_package` would be honored.

## Software Version
14.13.11

But the link I referenced above is on `master` so I'm assuming upgrading would not improve the situation.

## Replication Case

Here is the code from the recipe I'm using to install the updates. KB2919442 installs fine and is a prereq for KB2919355, which is the one that times out. That said, the issue doesn't appear to be specific to KB2919355. Any update that takes longer than 600 seconds to install should trigger it. Note that the recipe checks for `x86_64`. KB2919355 on x86 installs within the 600 second timeout.

```
if node['platform_version'].to_f == 6.3 # 8.1 or 2012r2
  if node['kernel']['machine'] == 'x86_64'
    # KB2919442 is a prereq for KB2919355
    msu_package 'March 2014 servicing stack update for Windows 8.1 and Windows Server 2012 R2 x64' do
      source 'http://download.windowsupdate.com/c/msdownload/update/software/crup/2014/02/windows8.1-kb2919442-x64_f97d8290d9d75d96f163095c4cb05e1b9f6986e0.msu'
      action :install
    end
    # KB2919355 is a prereq for KB2999226 which is installed by VC++ 2015+ Redistributable on systems prior to Windows 10 which is a prereq by VMware guest tools
    msu_package 'Windows RT 8.1, Windows 8.1, and Windows Server 2012 R2 update: April 2014 x64' do
      source 'http://download.windowsupdate.com/d/msdownload/update/software/crup/2014/02/windows8.1-kb2919355-x64_e6f4da4d33564419065a7370865faacf9b40ff72.msu'
      action :install
      timeout 1800
    end
end
```

## Stacktrace

```
[...]

PR #9649, opened 2020-04-10
msu_package: Fix cumulative updates installation and provide a 3600s default timeout

dheerajd-msys:

## Description

-- Updated to match with version as well along with package name.
-- Identifies correct match for cumulative updates.
-- Added `timeout` property for setting new default 3600 because some packages which are larger in sizes gets time nearly 45mins to 60mins to complete.
## Related Issue

Fixes: https://github.com/chef/chef/issues/9383
Fixes: #9245 
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/cab.rb
+++ b/lib/chef/provider/package/cab.rb
@@ -74,7 +74,7 @@ class Chef
 
         def dism_command(command)
           with_os_architecture(nil) do
-            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout })
+            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout || 3600 })
             if result.exitstatus == -2146498530
               raise Chef::Exceptions::Package, "The specified package is not applicable to this image." if result.stdout.include?("0x800f081e")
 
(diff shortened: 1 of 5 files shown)


Vasu1105:
According to me we should cover the test for the version for which we made the changes. I haven't saw that part has been covered in specs.

lib/chef/provider/package/cab.rb
```diff
@@ -74,7 +74,7 @@ def remove_package(name, version)
 
         def dism_command(command)
           with_os_architecture(nil) do
-            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout })
+            result = shell_out("dism.exe /Online /English #{command} /NoRestart", { timeout: new_resource.timeout || 3600 })
```
dheerajd-msys, line 77: I am only concerned about timeout to be ORed 3600. During testing on 2016 I found the package to get installed in nearly 45mins. So I've added here.

tas50: If we want to change the default we should define a timeout property in the msu_package resource and set a new default. Doing it this way hides it from our documentation generation so no user will know there's a new default.

dheerajd-msys: Thanks @tas50 for suggestion. Will be adding `timeout` property to override default.

dheerajd-msys:
The msu package with which we got the issue reproduced is larger in size and we wont be able to cover in the test case including that package as source.

PR #9653, opened 2020-04-11
fix continuous install of same package version

jalberti:
subscriptions to dnf_package resources are triggered each time the resource converges with an install action

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -445,6 +445,9 @@ class Chef
                   elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
                     logger.warn("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{new_version}. Skipping...)")
                     target_version_array.push(nil)
+                  elsif version_equals?(current_version, candidate_version)
+                    logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+                    target_version_array.push(nil)
                   else
                     logger.trace("#{new_resource} #{package_name} #{current_version} needs updating to #{new_version}")
                     target_version_array.push(new_version)


tas50:
@jalberti We did a large amount of refactoring in dnf_package on master recently. This will be shipping in Chef Infra Client 16. Can you possibly test out your issue against a current build (https://downloads.chef.io/chef/current) and let us know if this is still an issue.

lamont-granquist:
Yeah this is also the wrong fix, it blows up CI.  Messing around with the package superclass is very hard to do correctly because of all the use cases, and there should be no reason to fix this problem there.  I'm pretty sure that I fixed this in Chef 16 and it boiled down to issues with EVRA-vs-EVR comparisons never succeeding and the workaround (if I recall correctly, is that including the full EVRA in the package resource works around it), but I can't tell if you don't include any linked issue or any reproduction steps.

jalberti:
Before making these changes I tried to back port all the changes from Chef 16/master, but those changes did not fix my issue, they break backward compatibility with existing attribute values.

All out pipelines are written against centos 6, i.e. they are using the yum_package, now we are porting to centos 8 ... the example where we saw this looks as follows ...

yum info ....
Available Packages
Name         : broken-package
Epoch        : 1
Version      : 0.1
Release      : 123
Architecture : x86_64

Now I define a resource ... 
e.g. 
new_version: 0.1-123
current_version: 1:0.1-123.x86_64
candidate_version: 1:0.1-123.x86_64

package will be always "re-installed" ... it does not really re-install, yum catches the issue and it results in a no-op from the host point of view, but the subscriptions all fire

lamont-granquist:
> those changes did not fix my issue, they break backward compatibility with existing attribute values.

That is a more interesting question to solve.  We are going to backport the Chef-16 behavior to Chef-15 so if we have breaking issues we need to know about it.

lamont-granquist:
If this is the problem where you have to include a full EVRA in the version that's deliberate.

`version 1.2.3-1` will only match 0:1.2.3-1 now and you must write `version 1:1.2.3-1` if that is your package's epoch.

That may block the backporting of any fixes to Chef-15, it'll become a tradeoff between forcing that breaking change in Chef-15 and fixing this bug.

jalberti:
> If this is the problem where you have to include a full EVRA in the version that's deliberate.
> 
> `version 1.2.3-1` will only match 0:1.2.3-1 now and you must write `version 1:1.2.3-1` if that is your package's epoch.
> 
> That may block the backporting of any fixes to Chef-15, it'll become a tradeoff between forcing that breaking change in Chef-15 and fixing this bug.

Yes, that is exactly the issue. I think you are taking the breaking change too lightly, and I'm also not sure why you are thinking my change is not safe for this case. Basically my change says ... the version that would have been installed is already installed ... so this from my point of view is exactly what you want here, but I'm certainly not the expert, as you rightfully said, one would expect that such a fundamental issue should sit deeper in the stack, i.e. dnf specific provider, but I was not able to figure out how to fix this lower in the stack, in the time I was willing to spend on this ...

lamont-granquist:
That breaking change is correct behavior.  The prior behavior of chef-client was to treat 1.2.3 like it matches 0:1.2.3, 1:1.2,3, 2:1.2.3, etc.  Which is not correct per the spec of RPMs or how rpm/yum/dnf treat version strings.  Per the spec and the behavior of the underlying tools 1.2.3 is only equivalent to 0:1.2.3 and is never equivalent to 1:1.2.3.  That was deliberately removed.  Chef now uses the underlying python libraries to do version comparisons and those return that 1.2.3 != 1:1.2.3 correctly.  That behavior will not be changed or reverted.  You must update your existing code to work around that.  That decision was very much not taken lightly and was due to years of issue involved with maintaining our own custom ruby implementation of RPM version parsing and comparison which was buggy and deliberately violated the RPM spec.  We did that in a major breaking change release with the yum_package provider that shipped in Chef 14 and the behavior of the dnf_package provider in Chef 16 now will match that behavior.  We aren't going to re-litigate that decision at this time.

The patch that you submitted can't be accepted because it breaks behavior:

```
Failures:

  1) Chef::Provider::Package - Multi when installing multiple packages installs the specified version when some are out of date
     Failure/Error: expect(new_resource).to be_updated
       expected `<package[emacs, vi] @package_name: ["emacs", "vi"] @name: "emacs, vi" @before: nil @params: {} @provider: nil @allowed_actions: [:nothing, :install, :upgrade, :remove, :purge, :reconfig, :lock, :unlock] @action: [:install] @updated: false @updated_by_last_action: false @source_line: nil @guard_interpreter: nil @default_guard_interpreter: :default @elapsed_time: 0 @version: ["1.0", "6.3"]>.updated?` to return true, got false
     # ./spec/unit/provider/package_spec.rb:523:in `block (3 levels) in <top (required)>'
```

It has also broken functional specs which run on opensuse:

```
Failures:

  1) Chef::Resource::ZypperPackage with versions downgrades when the installed version is higher than the package_name version
     Failure/Error: expect(zypper_package.updated_by_last_action?).to be true

       expected true
            got false
     # ./spec/functional/resource/zypper_package_spec.rb:129:in `block (3 levels) in <top (required)>'
```

That is why there's a red x next to the PR on buildkite (the kitchen test failures are some other issues that would go away if the branch was rebased against master and force-pushed).

There's lot of complicated edge conditions to the package providers and it is nearly impossible at this point to find a quick fix to something in the superclass without causing more problems than you solve.

jalberti:
I'm totally fine with that decision, it is your product, therefore your decision. Thanks for looking at and discussing this PR here. Feel free to close this PR. Best regards, Johannes

PR #9656, opened 2020-04-13
Remove copyright dates

lamont-granquist:
Legally incredibly dubious, particularly since we don't follow it
strictly as policy, and we have git history instead, which does it right.
This is just a waste of time and a cargo cult.


lamont-granquist:
regexp fail, one second...

lamont-granquist:
Auditing the commits with 2020 dates, the highest date on the file is right less than half the time:

```
   1 2015
  39 2016
   4 2017
  16 2018
  22 2019
  76 2020
```

Pretty sure that if there was any conceivable value from this policy that the fact that we do not adhere to the policy nullifies any potential benefit.  Policies that aren't followed don't have any utility.

lamont-granquist:
Also corporate copyrights last for 95 years which means that the game of updating it every time the file is touched to eek one more year out of the copyright is a bit academic.

lamont-granquist:
Yeah that makes a bit more sense since then if someone sees the file out of context somehow, the copyright owner and license is clear.

tas50:
We should backport this just to make the diffs easier for now

lamont-granquist:
we should stop generating diffs against the copyright lines now though?

tas50:
if you diff chef-15 and master branches it's 100% red now

lamont-granquist:
heh, i guess i never do that...

PR #9663, opened 2020-04-15
Remove deprecated Chef-10 constraint handling

damacus:
Remove deprecate Chef-10 constraint handling

I'll sign the DCO if this is the right thing to do 😄

### DIFF
--- a/lib/chef/cookbook/metadata.rb
+++ b/lib/chef/cookbook/metadata.rb
@@ -481,7 +481,7 @@ class Chef
         @maintainer_email             = o[MAINTAINER_EMAIL] if o.key?(MAINTAINER_EMAIL)
         @license                      = o[LICENSE] if o.key?(LICENSE)
         @platforms                    = o[PLATFORMS] if o.key?(PLATFORMS)
-        @dependencies                 = handle_deprecated_constraints(o[DEPENDENCIES]) if o.key?(DEPENDENCIES)
+        @dependencies                 = handle_incorrect_constraints(o[DEPENDENCIES]) if o.key?(DEPENDENCIES)
         @providing                    = o[PROVIDING] if o.key?(PROVIDING)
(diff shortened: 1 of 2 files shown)


tas50:
@damacus This needs a DCO sign off

tas50:
Also this method is used at https://github.com/chef/chef/blob/9681323e610b02586fe2803c4b63a33462e971d4/lib/chef/cookbook/metadata.rb#L484

tas50:
@damacus There's a pile of old specs that need to go away as well.

lamont-granquist:
oh please yes kill it with fire.  i was looking at this the other day but haven't had the time.

tas50:
This one needs a cookstyle rule for the 1 person left doing this.

damacus:
I'm leaving in the code to handle multiple versions. I may possibly remove it in subsequent PR. But that means unpicking way more code.

lib/chef/cookbook/metadata.rb
```diff
@@ -535,6 +535,25 @@ def source_url(arg = nil)
...
+      # multiple constraints to be placed on cookbooks, as well as the
+      # << and >> operators, which are now just < and >. For
+      # specifications with more than one constraint, we return an
+      # empty array (otherwise, we're silently abiding only part of
+      # the contract they have specified to us). If there is only one
+      # constraint, we are replacing the old << and >> with the new <
+      # and >.
+      def handle_incorrect_constraints(specification)
+        specification.inject(Mash.new) do |acc, (cb, constraints)|
+          constraints = Array(constraints)
```
tas50, line 551: Is there a reason to leave the commented out part here?

damacus: Not at all.
Thanks for the spot, I'll tidy that up tomorrow

tas50:
Merging this based off the approval in our client review session.

PR #9666, opened 2020-04-15
Add a warning to the end of the chef run for EOL releses

tas50:
We really want to encourage people to upgrade. One way we can do this is
to just warn them that they're on an EOL release. This simple warning
might nudge them to upgrade their client. It works off the base EOL
data of Chef 15 and takes the current release to add to that date. That
way we don't have to maintain anything and as long as we keep a yearly
release we're good. If we change the schedule then we either remove this
or update the logic. That's a risk in shipping code that doesn't fire
for 2 years.

### DIFF
--- a/lib/chef/client.rb
+++ b/lib/chef/client.rb
@@ -329,15 +329,12 @@ class Chef
     def warn_if_eol
       require_relative "version"
 
-      # we have a yearly release and we know 15 goes EOL in 2021 so calculate off that
-      # this way we don't have to update anything or maintain a hash of EOL dates
-      base_eol_year = 2022
-      base_release = Gem::Version.new(15)
+      # We make a release every year so take the version you're on + 2006 and you get
+      # the year it goes EOL
+      eol_year = 2006 + Gem::Version.new(Chef::VERSION).segments.first
 
-      diff_from_base_release = Gem::Version.new(Chef::VERSION).segments.first - base_release.segments.first
-
-      if Time.now > Time.new(base_eol_year + diff_from_base_release, 5, 1)
-        logger.warn("This release of #{Chef::Dist::PRODUCT} became end of life (EOL) on May 1st #{base_eol_year + diff_from_base_release}. Please update to a supported release to receive new features, bug fixes, and security updates.")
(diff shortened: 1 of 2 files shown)


lib/chef/client.rb
```diff
@@ -323,6 +325,22 @@ def run
     # @todo make this stuff protected or private
     #
 
+    # @api private
+    def warn_if_eol
+      require_relative "version"
+
+      # we have a yearly release and we know 15 goes EOL in 2021 so calculate off that
```
phiggins, line 332:
```suggestion
      # we have a yearly release and we know 15 goes EOL in 2022 so calculate off that
```

Is this comment wrong?

lib/chef/client.rb
```diff
@@ -323,6 +325,22 @@ def run
...
+      # this way we don't have to update anything or maintain a hash of EOL dates
+      base_eol_year = 2022
+      base_release = Gem::Version.new(15)
+
+      diff_from_base_release = Gem::Version.new(Chef::VERSION).segments.first - base_release.segments.first
```
lamont-granquist, line 337: `eol_year = 2007 + Gem::Version.new(Chef::VERSION).segments.first` ?

the double magic constants of 15 expiring in 2022 and doing arithmetic around that seems more confusing to me than just having a magic base year of 2007 that makes the math work out.

lib/chef/client.rb
lamont-granquist: isn't that comment right?  15 expires next year.

phiggins: One of the tests is testing that 15 gets an EOL message mentioning 2022. Is the comment right or is the test? Or am I just very confused?

lib/chef/client.rb
lamont-granquist: (or 2006)

lib/chef/client.rb
tas50: I set the dates in the specs to be right on the border now so it should line up better with the dates.

lib/chef/client.rb
tas50: Yeah that makes way more sense to me as well. I updated the PR

PR #9679, opened 2020-04-16
Ruby 2.7 fix for powershell_out

lamont-granquist:
Just makes a warning go away


lamont-granquist:
(don't merge until green)

PR #9682, opened 2020-04-17
Fix platform_version Chef::VersionString API

lamont-granquist:
The consumption of ohai attributes by the ohai segment plugin code
was breaking this feature.


lamont-granquist:
I don't think I have time to screw around with this one or noodle about the right way to test this, time is getting a bit too short.

PR #9688, opened 2020-04-17
Force requiring properties

### LINKED ISSUE #4596: Chef doesn't fail if a resource's required properties is missing

lamont-granquist:
All required properties are now required for all actions by default
even if the action does not reference the property.

In order to only make the property required for a subset of the
actions, specify them as an array of symbols to the required
options on the property.

```
property :whatever, String, required: %i{start stop}

action :start do
end

action :stop do
end

action :enable do
end

action :disable do
end
```

That property will be required for start+stop but not for
enable+disable.

There's an unaddressed edge case here where if you reference the
property in an action which was not specified that it will also
fail validation.  That is correct behavior.  We should probably
dig into how to warn the user that they must either remove the
reference to the property from that action or else to add the
action to the list of required actions on the property.

Closes #4596

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -303,8 +303,12 @@ class Chef
     #
     # @return [Boolean]
     #
-    def required?
-      options[:required]
+    def required?(action = nil)
+      if !action.nil? && options[:required].is_a?(Array)
+        options[:required].include?(action)
+      else
+        !!options[:required]
+      end
     end
 
     #
--- a/lib/chef/provider.rb
+++ b/lib/chef/provider.rb
@@ -152,6 +152,7 @@ class Chef
       new_resource.cookbook_name
     end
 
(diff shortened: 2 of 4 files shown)


tas50:
The API of required on certain actions is something people have asked for repeatedly. We need that internally to cleanup a lot of terrible code as well.

Issue #7783, opened 2018-10-26
Remove "Code Can" from the logo in macOS DMGs


tas50:
This image need to get updated to a higher resolution and the old slogan needs to get removed.

<img width="762" alt="screen shot 2018-10-25 at 8 39 25 pm" src="https://user-images.githubusercontent.com/1015200/47543266-2dce7800-d896-11e8-84df-3e6b123a071f.png">

lamont-granquist:
Did this ever get done?

tas50:
Nope https://github.com/chef/chef/blob/master/omnibus/resources/chef/dmg/background.png

tas50:
This was done in https://github.com/chef/chef/commit/2a1823174c72c655d942af79876a4abc8d231f8a

PR #9690, opened 2020-04-18
Remove "Code Can" from dmg background

tas50:
We don't use this slogan anymore


tas50:
Fixes https://github.com/chef/chef/issues/7783

PR #9695, opened 2020-04-20
[9431] Add remote_file support to archive_file

### LINKED ISSUE #9431: archive_file needs to accept accept remote files similar to poise_archive
### Describe the Enhancement:
poise_archive allowed the user to set the source as a remote file. We should handle this same scenario since that's one of the most common archive scenarios. This allows us to migrate users off poise_archive more easily.

### Describe the Need:

Add missing behavior from https://github.com/poise/poise-archive

damacus:
Fixes #9431

Signed-off-by: Daniel Webb <dan.webb@damacus.io>

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/default.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/default.rb
@@ -120,6 +120,13 @@ end
     path File.join(Chef::Config[:file_cache_path], archive)
     extract_to File.join(Chef::Config[:file_cache_path], archive.tr(".", "_"))
   end
+
+  archive_file "distant_tourism.tar.gz" do
+    url "https://github.com/chef/cookstyle/archive/v6.3.5.tar.gz"
+    checksum "4692955a785990423c856ce809063d230cc96672d2092871bf1e66735bd6edd5"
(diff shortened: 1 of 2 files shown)


damacus:
The failure here doesn't look like it's related to the resource update

tas50:
@damacus I really want to get the URL functionality added to Infra Client 16 (and backported to 15 eventually), but I'm not sure this is quite the way to go about handling it. Take a look at the UX of https://github.com/poise/poise-archive where you can just pass a URL and it does the right thing. I think that's what we want to adopt partially because I'd like to eventually autocorrect those resources to use archive_file and it would be much easier to do that if the API was somewhat similar. Take a look at https://github.com/poise/poise-archive/blob/master/lib/poise_archive/resources/poise_archive.rb#L89 for some idea of how Noah went about doing that. Not all of that is relevant, but the simple regex to determine if it's a path of a URL is something you could potentially run with. Let me know if you want to chat about it. You know where to find me.

damacus:
Sounds like a plan, this really was just to start the conversation off with the most minimal effort possible. 
I'll dig that code out some time soon. 

Thanks!

tas50:
Closing this out since we need to take a different approach

PR #9698, opened 2020-04-21
Avoid double pry by pinning to the exact version

tas50:
It's in the gemfile.lock once, but we're ending up with 0.12.2 and 0.13
which is super odd.


tas50:
This didn't fix it. #9700 did

PR #9726, opened 2020-04-23
Refactor scm, git and subversion resources & fix longstanding git issues

lamont-granquist:
The git provider now no longer checks out to a "deploy" branch by
default and now checks out to the branch (with a remote upstream) or
else checks out to a detatched head.

The prior behavior can be restored by using "checkout branch 'deploy'".

This also removes the SCM resource base class and replaces it with a
resource partial and does some internal reorganization.

It also introduces the RecipeDSLHelper for better functional tests and
cleans up the functional tests of the git provider.

Properties that were only ever implemented on the git provider were
removed from the subversion provider where they had been inherited from
the base class incorrectly.

Some additional env var handling was added to the subversion handler in
the process of sorting out the common properties, including HOME
handling for alternative users.


lib/chef/resource/scm/git.rb
```diff
@@ -31,6 +33,26 @@ class Git < Chef::Resource::Scm
...
+      property :enable_submodules, [TrueClass, FalseClass],
+        description: "Perform a sub-module initialization and update.",
+        default: false
+
+      property :enable_checkout, [TrueClass, FalseClass],
+        description: "Check out a repo from master. Set to false when using the checkout_branch attribute to prevent the git resource from attempting to check out master from master.",
+        default: true
+
+      property :remote, String,
+        default: "origin"
```
tas50, line 48: These should have descriptions so we generate the docs correctly

lib/chef/dsl/declare_resource.rb
```diff
@@ -297,13 +297,15 @@ def declare_resource(type, name, created_at: nil, run_context: self.run_context,
...
       #
-      def build_resource(type, name, created_at: nil, run_context: self.run_context, &resource_attrs_block)
+      def build_resource(type, name, created_at: nil, run_context: self.run_context, enclosing_provider: nil, &resource_attrs_block)
         created_at ||= caller[0]
 
         # this needs to be lazy in order to avoid circular dependencies since ResourceBuilder
         # will requires the entire provider+resolver universe
         require_relative "../resource_builder" unless defined?(Chef::ResourceBuilder)
 
+        enclosing_provider ||= self if is_a?(Chef::Provider)
```
tas50, line 307: maybe a comment on how this is used with the spec helper only

lib/chef/resource/scm/_scm.rb
```diff
@@ -0,0 +1,48 @@
...
+unified_mode true
+
+default_action :sync
+allowed_actions :checkout, :export, :sync, :diff, :log
+
+property :destination, String,
+  description: "The location path to which the source is to be cloned, checked out, or exported. Default value: the name of the resource block.",
+  name_property: true
+
+property :repository, String
```
tas50, line 28: This one needs a description as well

lib/chef/dsl/declare_resource.rb
lamont-granquist: as its generic though we can use the same trick elsewhere

lamont-granquist: (and its not just for the spec helper, the enclosing_provider goes back to chef-11-era)

tas50: If we're going to use it elsewhere then go for it. It just seems like passing it right now is a trick specifically for the specs. The docs on the spec side are solid for that which is probably fine.

PR #9740, opened 2020-04-25
Fix failing RHEL 6 kitchen tests

tas50:

### DIFF
--- a/kitchen-tests/kitchen.yml
+++ b/kitchen-tests/kitchen.yml
@@ -75,6 +75,9 @@ platforms:
     pid_one_command: /sbin/init
     intermediate_instructions:
       - RUN sed -i -e "s/Defaults.*requiretty.*/Defaults    !requiretty/g" /etc/sudoers
+      - RUN yum install -y centos-release-scl
+      - RUN yum install -y devtoolset-7
+      - RUN scl enable devtoolset-7 bash
 
 - name: centos-7
   driver:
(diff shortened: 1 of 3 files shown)


kitchen-tests/kitchen.yml
```diff
@@ -75,6 +75,11 @@ platforms:
     pid_one_command: /sbin/init
     intermediate_instructions:
       - RUN sed -i -e "s/Defaults.*requiretty.*/Defaults    !requiretty/g" /etc/sudoers
+      - RUN wget http://people.centos.org/tru/devtools-2/devtools-2.repo -O /etc/yum.repos.d/devtools-2.repo
+      - RUN yum install -y devtoolset-2-gcc devtoolset-2-binutils devtoolset-2-gcc-c++
+      - RUN ln -s /opt/rh/devtoolset-2/root/usr/bin/gcc /usr/local/bin/gcc
+      - RUN ln -s /opt/rh/devtoolset-2/root/usr/bin/c++ /usr/local/bin/c++
+      - RUN ln -s /opt/rh/devtoolset-2/root/usr/bin/g++ /usr/local/bin/g++
```
lamont-granquist, line 82: what about following https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/ ?

tas50: That's GCC 7 vs. 4.8. All we needed was 4.8 vs. 4.7, but your method is probably easier to install so I'll pull that in.

PR #9741, opened 2020-04-27
Remove the Windows version check in windows_share

tas50:
This was keeping this resource from working on Windows 8 / Windows
2012 which are 6.2. This works on all the versions of Windows we support
so there's no need to check this anymore.


btm:
It would be helpful for us (support wise) if we added a "feature" that checked that platform_version early in the chef run and emitted a warning if they were running on an unsupported version. 

I was thinking about what would happen if someone _tried_ to use IC16 on a Windows 7 system.

PR #9742, opened 2020-04-27
Add powershell_exec! helper to make conversion from powershell_out! easier

tas50:
This is something we have in powershell_out! and shell_out! already and we use it a lot. This will make updating resources much easier.

### DIFF
--- a/lib/chef/mixin/powershell_exec.rb
+++ b/lib/chef/mixin/powershell_exec.rb
@@ -20,7 +20,7 @@ require_relative "../powershell"
 # The powershell_exec mixin provides in-process access to PowerShell engine via
 # a COM interop (installed by the Chef Client installer).
 #
-# powershell_exec returns a Chef::PowerShell object that provides 3 methods:
+# powershell_exec returns a Chef::PowerShell object that provides 4 methods:
 #
 # .result - returns a hash representing the results returned by executing the
 #           PowerShell script block
@@ -28,6 +28,7 @@ require_relative "../powershell"
 #           PowerShell error stream during execution
 # .error? - returns true if there were error messages written to the PowerShell
 #           error stream during execution
+# .error! - raise Chef::PowerShell::CommandFailed if there was an error
 #
 # Some examples of usage:
 #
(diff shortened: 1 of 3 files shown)


spec/unit/mixin/powershell_exec_spec.rb
```diff
@@ -40,4 +40,14 @@
       expect(execution.errors[0]).to include("Runtime exception: this-should-error")
     end
   end
+
+  describe "#powershell_exec!" do
+    it "runs a basic command and returns a Chef::PowerShell object" do
+      expect(object.powershell_exec("$PSVersionTable")).to be_kind_of(Chef::PowerShell::CommandFailed
```
dheerajd-msys, line 46: Typo of `)` here.

lib/chef/mixin/powershell_exec.rb
```diff
@@ -20,14 +20,15 @@
 # The powershell_exec mixin provides in-process access to PowerShell engine via
 # a COM interop (installed by the Chef Client installer).
 #
-# powershell_exec returns a Chef::PowerShell object that provides 3 methods:
+# powershell_exec returns a Chef::PowerShell object that provides 4 methods:
```
lamont-granquist, line 23: `provides the following methods:`?

or just "amongst our methods are `result`, `errors`, `error?`, `error!`, an almost fanatical devotion to the Pope, and nice red uniforms..."

tas50: I was just updating the existing comment

spec/unit/mixin/powershell_exec_spec.rb
```diff
@@ -40,4 +40,14 @@
...
+
+  describe "#powershell_exec!" do
+    it "runs a basic command and returns a Chef::PowerShell object" do
+      expect(object.powershell_exec("$PSVersionTable")).to be_kind_of(Chef::PowerShell::CommandFailed)
+    end
+
+    it "raises an error if the command fails" do
+      expect { object.powershell_exec("$PSVersionTable") }.to raise_error(Chef::Exceptions::Script)
+    end
+  end
```
phiggins, line 52: Unless I'm missing something these tests aren't calling the correct method... ?

PR #9744, opened 2020-04-27
Stop pinning rake

tas50:
The removal of the gem install of appbundler probably fixes where this
was leaking in.


tas50:
Confirmed this has no impact via an adhoc build

PR #9764, opened 2020-04-29
Improve resource docs

tas50:
Add some missing docs we'll use to generate docs on the docs site. Fix
an incorrect resource name. Fix a bad link in the release notes.

### DIFF
--- a/lib/chef/resource/alternatives.rb
+++ b/lib/chef/resource/alternatives.rb
@@ -29,7 +29,7 @@ class Chef
       description "The alternatives resource allows for configuration of command alternatives in Linux using the alternatives or update-alternatives packages."
       introduced "16.0"
       examples <<~DOC
-      Install an alternatives
+      Install an alternative
 
       ```ruby
       alternatives 'python install 2' do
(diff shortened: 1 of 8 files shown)


lib/chef/resource/alternatives.rb
```diff
@@ -28,11 +28,72 @@ class Alternatives < Chef::Resource
 
       description "The alternatives resource allows for configuration of command alternatives in Linux using the alternatives or update-alternatives packages."
       introduced "16.0"
+      examples <<~DOC
+      Install an alternatives
```
phiggins, line 32:
```suggestion
      Install an alternative
```

tasks/docs.rb
```diff
@@ -266,7 +266,6 @@ def build_resource_data(name, data)
       r = {}
 
       # These properties are always set to these values.
-      r["draft"] = false
       r["resource_reference"] = true
       r["robots"] = nil
       r["syntax_code_block"] = nil
```
phiggins, line 271: Does it still make sense to set these two values to nil?

tas50: I think there's a bunch of these we can probably remove. I'm going to work with the docs team to figure out what we need to set and what we don't. Expect more PRs

PR #9765, opened 2020-04-29
More resource documentation improvements

tas50:
Make sure we don't document the deprecated properties and generally fix the formatting so we can auto generate the docs for the docs site properly.

### DIFF
--- a/lib/chef/resource/yum_repository.rb
+++ b/lib/chef/resource/yum_repository.rb
@@ -34,12 +34,12 @@ class Chef
       # http://linux.die.net/man/5/yum.conf as well as
       # http://dnf.readthedocs.io/en/latest/conf_ref.html
       property :baseurl, [String, Array],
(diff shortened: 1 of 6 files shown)


lib/chef/resource/yum_repository.rb
```diff
@@ -34,12 +34,12 @@ class YumRepository < Chef::Resource
       # http://linux.die.net/man/5/yum.conf as well as
       # http://dnf.readthedocs.io/en/latest/conf_ref.html
       property :baseurl, [String, Array],
-        description: "URL to the directory where the Yum repository's 'repodata' directory lives. Can be an http://, https:// or a ftp:// URL. You can specify multiple URLs in one baseurl statement."
+        description: "URL to the directory where the Yum repository's 'repodata' directory lives. Can be an `http://`, `https://` or a `ftp://` URLs. You can specify multiple URLs in one baseurl statement."
```
phiggins, line 37:
```suggestion
        description: "URL to the directory where the Yum repository's `repodata` directory lives. Can be an `http://`, `https://` or a `ftp://` URLs. You can specify multiple URLs in one `baseurl` statement."
```

Should these get backticks too?

PR #9767, opened 2020-04-29
More improvements to the docs generation

tas50:
Keeping this open to add resource updates to over time

### DIFF
--- a/lib/chef/resource/apt_preference.rb
+++ b/lib/chef/resource/apt_preference.rb
@@ -36,7 +36,7 @@ class Chef
         validation_message: "The provided package name is not valid. Package names can only contain alphanumeric characters as well as _, -, +, or *!"
 
       property :glob, String,
-        description: "Pin by a `glob()` expression or with a regular expressions surrounded by `/`."
+        description: "Pin by a `glob()` expression or with a regular expression surrounded by `/`."
 
       property :pin, String,
         description: "The package version or repository to pin.",
--- a/lib/chef/resource/apt_preference.rb
+++ b/lib/chef/resource/apt_preference.rb
@@ -36,7 +36,7 @@ class Chef
         validation_message: "The provided package name is not valid. Package names can only contain alphanumeric characters as well as _, -, +, or *!"
 
       property :glob, String,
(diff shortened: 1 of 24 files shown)


tasks/docs.rb
```diff
@@ -305,7 +305,7 @@ def build_resource_data(name, data)
       else
         puts "Writing out #{resource}."
         FileUtils.mkdir_p "docs_site/#{resource}"
-        File.open("docs_site/#{resource}/_index.md", "w") { |f| f.write(resource_data.to_yaml) }
+        File.open("docs_site/#{resource}/_index.md", "w") { |f| f.write(resource_data.to_yaml + "---") }
```
phiggins, line 308: Nitpick, but a comment about why the `"---"` got added would be good so it doesn't get left off in the future.

lib/chef/resource/apt_preference.rb
```diff
@@ -36,7 +36,7 @@ class AptPreference < Chef::Resource
         validation_message: "The provided package name is not valid. Package names can only contain alphanumeric characters as well as _, -, +, or *!"
 
       property :glob, String,
-        description: "Pin by glob() expression or with regular expressions surrounded by /."
+        description: "Pin by a `glob()` expression or with a regular expressions surrounded by `/`."
```
phiggins, line 39:
```suggestion
        description: "Pin by a `glob()` expression or with a regular expression surrounded by `/`."
```

tasks/docs.rb
```diff
@@ -153,19 +153,41 @@ def friendly_default_value(property)
...
+      end
+
+      # add the special case for nothing
+      list[:nothing] = { "shortcode" => "resources_common_actions_nothing.md" }
+
+      list
+    end
+
     # TODO:
     # - what to do about "lazy default" for default?
```
phiggins, line 177:
```suggestion
```

Extremely nitpick, but I added this comment and then I think you addressed that a day or two later?

tas50: the lazy bit is still there and I keep seeing this and wondering what the right solution is? Maybe some warnings in the script or something.

phiggins: Ahh ok.

PR #9771, opened 2020-04-30
Update registry helper behavior

jmassardo:
Signed-off-by: James Massardo <jmassardo@chef.io>

## Description
This change updates the behavior of the `registry_value_exists?` and `registry_data_exists?` helpers so they return `false` instead of an exception if the value or data does not exist.

There are several use cases where a one would use a registry key guard on resources that install and/or update certain applications where the data may or may not exist or may or may not contain the desired value. This change allows one to use a single guard to on their resource instead of needing to perform several checks beforehand.

## Related Issue
Here are a couple examples of users that struggle with this issue.
* https://chef-success.slack.com/archives/C0E8C3K97/p1587758901048000
* https://discourse.chef.io/t/windows-registry-key-only-if-not-if-registry-value-exists-help/8075/2

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have run the pre-merge tests locally and they pass.
- [ x ] I have updated the documentation accordingly.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.
- [ x ] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/lib/chef/win32/registry.rb
+++ b/lib/chef/win32/registry.rb
@@ -201,7 +201,7 @@ class Chef
       end
 
       def value_exists?(key_path, value)
-        key_exists!(key_path)
+        return false unless key_exists?(key_path)
         hive, key = get_hive_and_key(key_path)
         hive.open(key, ::Win32::Registry::KEY_READ | registry_system_architecture) do |reg|
           return true if reg.any? { |val| safely_downcase(val) == safely_downcase(value[:name]) }
@@ -210,7 +210,7 @@ class Chef
       end
 
       def data_exists?(key_path, value)
-        key_exists!(key_path)
+        return false unless key_exists?(key_path)
         hive, key = get_hive_and_key(key_path)
         hive.open(key, ::Win32::Registry::KEY_READ | registry_system_architecture) do |reg|
           reg.each do |val_name, val_type, val_data|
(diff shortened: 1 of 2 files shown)


tas50:
This seems like the right thing to do, but it's a breaking change for anyone who has programmed to the existing behavior so we can't ship it outside a major release. We can tag this for Chef 17 in April 2021 if you'd like?

jmassardo:
@tas50 can you clarify how this would be a breaking change? Anyone currently using these helpers would get an exception that will stop the chef run.

lamont-granquist:
> Anyone currently using these helpers would get an exception that will stop the chef run.

They can be rescuing those exceptions and then doing something where `false` is unexpected.  If they're manually suppressing the exceptions to nil this can case a boolean false value to be injected into their code at that point where its not expected.  It may most commonly be a not_if/only_if, but it could just be a conditional check in arbitrary code.

Honestly this is probably not worth the 8 years of people figuring this out and working around it.

lamont-granquist:
Oh actually this is entirely deliberate API

`value_exists?` needs the hive and the key to exist.  You must protect `value_exists?` with checks around the hive or the key not existing.  If the value along doesn't exist this API returns false.  Bit fussy API from the Chef 11.0 era.

PR #9778, opened 2020-05-01
Another round of description updates && Docs generator tweaks

tas50:

### DIFF
--- a/lib/chef/resource/apt_package.rb
+++ b/lib/chef/resource/apt_package.rb
@@ -25,6 +25,35 @@ class Chef
 
       provides :apt_package, target_mode: true
       provides :package, platform_family: "debian", target_mode: true
+      examples <<~DOC
+      **Install a package using package manager**:
+
+      ```ruby
+      apt_package 'name of package' do
+        action :install
+      end
+      ```
+
+      **Install a package without specifying the default action**:
+
+      ```ruby
+      apt_package 'name of package'
+      ```
+
+      **Install multiple packages at once**:
+
+      ```ruby
+      apt_package %(package1 package2 package3)
+      ```
+
+      **Install without using recommend packages as a dependency**
+
+      ```ruby
+      package 'apache2' do
+        options '--no-install-recommends'
+      end
+      ```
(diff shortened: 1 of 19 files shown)


lib/chef/resource/apt_package.rb
```diff
@@ -25,6 +25,35 @@ class AptPackage < Chef::Resource::Package
...
+      **Install multiple packages at once**:
+
+      ```ruby
+      apt_package %(package1 package2 package3)
+      ```
+
+      **Install without using recommend packages as a dependency**
+
+      ```ruby
+      package 'apache2' do
```
jaymzh, line 52: did you want this to be `apt_package` like the others?

tasks/docs.rb
```diff
@@ -189,85 +183,69 @@ def properties_list(properties)
...
-          false
-        end
+      # these package properties support passing arrays for the package name
+      properties["common_resource_functionality_multiple_packages"] = true if %w{yum_package apt_package zypper_package homebrew_package dnf_package pacman_package}.include?(name)
 
-      properties["common_resource_functionality_resources_common_windows_security"] = name == "remote_directory"
+      properties["common_resource_functionality_resources_common_windows_security"] = true if name == "remote_directory"
 
-      properties["cookbook_file_specificity"] = name == "cookbook_file"
+      properties["cookbook_file_specificity"] = true if name == "cookbook_file"
```
jaymzh, line 191: why the tautology here? Are you trying to keep some existing value in these if they are not true?

tas50: The intention is to avoid creating these yaml entries unless they're set the true. Everything in the hugo config defaults to false and including 30 false values make it really hard to quickly determine what's going on. It gets us cleanup like this https://github.com/chef/chef-web-docs/pull/2476

lib/chef/resource/apt_package.rb
tas50: That was in the existing examples on the docs site and I think the intention was to show the `provides :package` for the debian platform_family

PR #9791, opened 2020-05-04
Fix required errors in apt_preference and systctl resources

tas50:
The remove actions in both of these resources shouldn't require the properties that are only necessary for the creation of the configs.

I also added Test Kitchen tests from the resource docs examples so we catch this in the future.


lib/chef/resource/sysctl.rb
```diff
@@ -44,7 +104,7 @@ class Sysctl < Chef::Resource
       property :value, [Array, String, Integer, Float],
         description: "The value to set.",
         coerce: proc { |v| coerce_value(v) },
-        required: true
+        required: %i(:apply)
```
phiggins, line 107:
```suggestion
        required: [:apply]
```

Was this intentional?

```
irb(main):001:0> %i{foo}
=> [:foo]
irb(main):002:0> %i{:foo}
=> [:":foo"]
```

tas50: Just a typo

PR #9797, opened 2020-05-05
Chef 15 backport of Chef 16 knife cleanup

lamont-granquist:
Backport of some of #9646 to make it so Chef-16 style knife plugins
can run on Chef-15.

What this does:

1.  Supports the config: named option in the initializer of the ChefVaultHandler and the ClientBuilder objects, this is so that knife commands may switch to that and support Chef-15 + Chef-16.  The knife_config named parameter will be deprecated in Chef-16 and removed in Chef-17.

2.  Backports merging all the default Chef::Config options Chef-16 style at the lowest precedence level, so that Chef::Config options that have no corresponding CLI option will be included in the config hash.  There might be edge conditions where knife plugins break due to expecting that Chef::Config options *aren't* in this hash, but going forwards that is simply buggy behavior, and there's no sane way to leave this the way it is in Chef-15 and write plugins that are backwards compatible.  So I'm declaring that was never part of any published API, so was not covered under SemVer, so we're not going backwards to the old behavior here.  Knife plugins that break need to be fixed, this will not be reverted, otherwise we get stuck in the perverse situation where we can never fix this.  Going forwards the situation is simple and knife plugins should just use the config hash for all their CLI options and Config options, period, full stop.  The should stop worrying about where the options come from, and rip out all the "if I set it / don't set it in the CLI and/or set it / don't set it in the config has does it work right?" (and most of those don't call knife.merge_configs correctly and were never testing the correct behavior any way)

3. There's a subtle change to config_value which affects the "fallback" value, and is considered a bugfix.  Knife plugins should simply not use config_value at all, period, full stop to begin with.  The one argument form of config_value is equivalent to simply getting the value out of the config hash, and the latter should always be preferred.  For code that does use the two and three argument versions this code was buggy:

```
          if Chef::Config[:knife].key?(lookup_key) || config.key?(lookup_key)
            Chef::Config[:knife][lookup_key] || config[lookup_key]
```

If the `Chef::Config[:knife][:lookup_key]` was set to false, then the second line would trigger the `||` and would evaluate the `config[lookup_key]` (which should have always been `nil` based on the way this API was intended to be used since the CLI option should have been deprecated using mixlib-cli's deprecation feature).  That would convert a `false` value in the `Chef::Config` value to a `nil` value, that has been fixed and the code so it now returns `false` and has been simplified to make it clearer.  This is another "sorry, not sorry" if it breaks any code out there.  Stop using this API, it is not your friend.


lamont-granquist:
🚧 notice: i had issues running the tests locally, so need to see what BK says.

spec/unit/knife/bootstrap_spec.rb
```diff
@@ -1072,7 +1076,7 @@ def certificates
                 keys_only: false, # implied false from config password present
                 key_files: ["/identity.pem", "/gateway.pem"], # Config
                 sudo: true, # ccli
-                verify_host_key: nil, # Config
+                verify_host_key: false, # Config
```
lamont-granquist, line 1079: This is where the API has been fixed to be correct, the spec here was testing buggy behavior

spec/unit/knife/bootstrap_spec.rb
```diff
@@ -1056,6 +1059,7 @@ def certificates
               knife.config[:use_sudo] = true
               knife.config[:use_sudo_pasword] = true
               knife.config[:ssh_forward_agent] = true
+              knife.merge_configs
```
lamont-granquist, line 1062: Everyone forgets to call merge_configs when testing CLI/Config interactions

Issue #9806, opened 2020-05-06
windows_package fails due to version property as String when package is installed


Tensibai:
## Description

Follow up on #9770 

Due to the enforcement of single version package and the way load_current_resource is done, populating the version property fails when a package is already installed.

The current resource version property is done here https://github.com/chef/chef/blob/7235856d859bbb32482b3f3b12f758b63a0eb8eb/lib/chef/provider/package/windows.rb#L65-L66

With the version property being defined as String since [this MR](https://github.com/chef/chef/commit/6ed90220893285f80513c6155620ebd0d2fbda1e#diff-3c9b1a519cef717fb2fa5d2f52a30477) 

Both msi and exe provider used in `package_provider.installed_version` can ends up returning an array with map(..).uniq:

 - https://github.com/chef/chef/blob/7235856d859bbb32482b3f3b12f758b63a0eb8eb/lib/chef/provider/package/windows/exe.rb#L93-L98:

- https://github.com/chef/chef/blob/7235856d859bbb32482b3f3b12f758b63a0eb8eb/lib/chef/provider/package/windows/msi.rb#L48-L59

## Chef Version
16.0.275

## Platform Version
Non relevant

## Replication Case
Converge twice with a windows_package resource, the second run will fail.

## Client Output
```
================================================================================
    Error executing action `install` on resource 'windows_package[Microsoft Visual C++ 2013 Redistributable (x64) - 12.0.40660]'
    ================================================================================
    Chef::Exceptions::ValidationFailed
    ----------------------------------
    Property version must be one of: String!  You passed ["12.0.40660.0"].
    Resource Declaration:
    ---------------------
    # In c:/chef/local-mode-cache/cache/cookbooks/vcruntime/recipes/vc12.rb
     24:     windows_package node['vcruntime']['vc12']['x64'][node['vcruntime']['vc12']['version']]['name'] do
     25:       checksum node['vcruntime']['vc12']['x64'][node['vcruntime']['vc12']['version']]['sha256sum']
     26:       source node['vcruntime']['vc12']['x64'][node['vcruntime']['vc12']['version']]['url']
     27:       installer_type :custom
     28:       returns [0, 3010]
     29:       options '/q'
     30:     end
     31:     windows_package node['vcruntime']['vc12']['x86'][node['vcruntime']['vc12']['version']]['name'] do
    Compiled Resource:
    ------------------
[...]

PR #9810, opened 2020-05-06
Fix windows_package not allowing version to be an array

tas50:
Due to the way the resource is built we need to be able to store an
array of current versions. I added some tests and comments so we don't
refactor this away later.


tas50:
Fixes https://github.com/chef/chef/issues/9806

Issue #9788, opened 2020-05-04
Multiple issues with launchd resource


erikng:
## Description
Moving away from the official plist gem to the built in provider ported from Microsoft has resulted in significant regressions moving from the latest stable Chef 15 to Chef 16.

I believe the issue is firmly due to https://github.com/chef/chef/pull/9642

Regressions found:
#### launchd idempotence
```
  * launchd[com.googlecode.munki.ManagedSoftwareCenter] action enable
    * file[/Library/LaunchAgents/com.googlecode.munki.ManagedSoftwareCenter.plist] action create (up to date)
    * macosx_service[com.googlecode.munki.ManagedSoftwareCenter] action restart
      - restart service macosx_service[com.googlecode.munki.ManagedSoftwareCenter]
```

#### Creating a launchd item fails due to 0kb files being created on disk with the plist module unable to read the data.
```
    * launchd[com.uber.chef.uberinternal.umad.agent] action enable
      * file[/Library/LaunchAgents/com.uber.chef.uberinternal.umad.agent.plist] action create
        - create new file /Library/LaunchAgents/com.uber.chef.uberinternal.umad.agent.plist
      * macosx_service[com.uber.chef.uberinternal.umad.agent] action restart
        
        ================================================================================
        Error executing action `restart` on resource 'macosx_service[com.umad.agent]'
        ================================================================================
        
        Mixlib::ShellOut::ShellCommandFailed
        ------------------------------------
        Expected process to exit with [0], but received '1'
        ---- Begin output of plutil -convert xml1 -o - /Library/LaunchAgents/com.umad.agent.plist ----
        STDOUT: /Library/LaunchAgents/com.umad.agent.plist: Property List error: Cannot parse a NULL or zero-length data / JSON error: No value.
        STDERR: 
        ---- End output of plutil -convert xml1 -o - /Library/LaunchAgents/com.umad.agent.plist ----
        Ran plutil -convert xml1 -o - /Library/LaunchAgents/com.umad.agent.plist returned 1

```

#### launchd cannot find the right service (it does not look to be honoring the `type` any longer, resulting in it finding "duplicate" service names.

```
    * launchd[com.somethingumad] action enable
      * file[/Library/LaunchAgents/com.something.umad.plist] action create (up to date)
      * macosx_service[com.umad] action restart
        * Several plist files match service name. Please use full service name.
        ================================================================================
        Error executing action `restart` on resource 'macosx_service[com.something.umad]'
        ================================================================================
        
[...]

PR #9812, opened 2020-05-07
fix launchd provider for chef-16

lamont-granquist:
the refactoring to use the DSL (orthogonal to setting this up as
a unified_mode resource) meant we have to care about lexical scoping
now.


tas50:
Fixes #9788

phiggins:
Fixes a regression introduced in #9570.

PR #9817, opened 2020-05-07
Switch from ASDF to a Amazon Linux extras collection for Ruby

tas50:
Fingers crossed this speeds up our tests. We won't get the latest and greatest ruby, but we don't need that. We need a version of ruby we can install our kit in, which gives them a year to update.


tas50:
https://github.com/chef/chef/pull/9828

PR #9825, opened 2020-05-07
Fix a misspelling in knife subcommand.

phiggins:
"recind" is not a word.

### DIFF
--- a/lib/chef/knife/user_invite_rescind.rb
+++ b/lib/chef/knife/user_invite_rescind.rb
@@ -20,14 +20,14 @@ require_relative "../knife"
 
 class Chef
   class Knife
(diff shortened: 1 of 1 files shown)


tas50:
This is technically breaking, but I think that it makes sense here to just fix it. Horrible typos can't live on forever.

phiggins:
> This is technically breaking, but I think that it makes sense here to just fix it. Horrible typos can't live on forever.

I thought about that, and we could probably alias it by creating a class with the same name that's equal to the new one, but I'm not sure how that would work with all the class name to command line arg translation happening.

I also didn't see this documented anywhere so I'm not sure how widely used it is.

PR #9826, opened 2020-05-08
Add spellcheck config and fix a bunch of violations

phiggins:
Adds two rake tasks:

* `rake spellcheck` lists the spellcheck violations with the file and line number they occur on.
* `rake spellcheck:unknown_words` lists just the words of spellcheck violations, useful for generating exceptions lists.

The exception list is pretty long because I wanted to avoid having to change code and potentially break things.


cspell.json
```diff
@@ -0,0 +1,2195 @@
...
+    "zypper",
+    "Ásgeirsson"
+  ],
+  // flagWords - list of words to be always considered incorrect
+  // This is useful for offensive words and common spelling errors.
+  // For example "hte" should be "the"
+  "flagWords": [
+    "hte"
+  ],
+  "ignorePaths": [
```
phiggins, line 2172: I tried to ignore most things that did not have any prose in them, things that would not ever be user facing, and directories with binaries in them.

I avoided all of `spec/**` because main tests are full of nonsense words and gibberish that would have taken a long time to go over.

lib/chef/win32/security/sid.rb
```diff
@@ -261,23 +261,23 @@ def self.current_user
...
+        # Check if the user is in builtin system group
         #
         # @return [Boolean] True or False
         #
         def self.group_user?(user)
           BUILT_IN_GROUPS.include?(user.to_s.upcase)
         end
 
-        # Сheck if the user belongs to system users category
+        # Check if the user belongs to system users category
```
phiggins, line 280: These are not the latin letter `C`, they are a Cyrillic character. https://en.wikipedia.org/wiki/Es_(Cyrillic)

lib/chef/cookbook_loader.rb
```diff
@@ -27,7 +27,7 @@
 
 class Chef
   # This class is used by knife, cheffs and legacy chef-solo modes.  It is not used by the server mode
-  # of chef-client or zolo/zero modes.
```
tas50, line 30: This might have been intentional. Zero + Solo was referred to as zolo.

lib/chef/provider/remote_directory.rb
```diff
@@ -245,7 +245,7 @@ def directory_resource(dir)
         res = Chef::Resource::Directory.new(dir, run_context)
         res.cookbook_name = resource_cookbook
         if ChefUtils.windows? && rights
-          # rights are only meant to be applied to the toppest-level directory;
```
tas50, line 248: seems like this should be most top level

lib/chef/cookbook_loader.rb
phiggins: TIL, reverted.

Rakefile
```diff
@@ -108,3 +108,19 @@ begin
 rescue LoadError
   puts "yard is not available. bundle install first to make sure all dependencies are installed."
 end
+
+if `command -v cspell`.empty?
```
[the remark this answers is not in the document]
phiggins: `command` is supposed to be the POSIX compliant way of doing this so I'm not sure why it wouldn't work. I removed it, it wasn't super necessary to gate this.

tas50:
Rebased to resolve the conflict with the fix I already made to the release notes. Glad to get all the user facing typos fixed. Thanks @phiggins

PR #9827, opened 2020-05-08
Remove the config hash defaults

lamont-granquist:
The backport to Chef-15 made me realize that this all needs to be
deleted.

The :hints looks necessary to initialize that to a Hash and to
create the config context to begin with.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -847,33 +847,6 @@ module ChefConfig
 
     # knife configuration data
     config_context :knife do
-      # XXX: none of these default values are applied to knife (and would create a backcompat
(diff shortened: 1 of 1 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -847,33 +847,6 @@ def self.init_openssl
 
     # knife configuration data
     config_context :knife do
-      # XXX: none of these default values are applied to knife (and would create a backcompat
-      # break in knife if this bug was fixed since many of the defaults below are wrong).  this appears
```
lamont-granquist, line 851: ^ we fixed that bug, so now these defaults are wrong and need to go

PR #9828, opened 2020-05-08
Try removing asdf entirely and just use omnibus-toolchain

tas50:
It seems odd that we install ruby over ruby


lamont-granquist:
ENOTENOUGHRUBIES

Issue #3351, opened 2015-05-07
Cron resource symbols for weekday are broken (regression)


mynamewastaken:
I tried to create a cron job:

```
cron 'clean_db_archive_logs' do
  minute '0'
  hour '6'
  weekday :tuesday
  user 'postgres'
  command '/var/lib/postgresql/backup.py'
end
```

and knife'd it up to our chef server. I ran chef-client and the run failed with the following error:

```
================================================================================
Error executing action `create` on resource 'cron[clean_db_archive_logs]'
================================================================================

Chef::Exceptions::Cron
----------------------
Error updating state of clean_db_archive_logs, exit: 1

Resource Declaration:
---------------------
# In /var/chef/cache/cookbooks/<cb>/recipes/<recipe>

 42: cron 'clean_db_archive_logs' do
 43:   minute '0'
 44:   hour '6'
 45:   weekday :tuesday
 46:   user 'postgres'
 [..]
 49:   command '/var/lib/postgresql/backup.py'
 50: end

Compiled Resource:
------------------
# Declared in /var/chef/cache/cookbooks/<cb>/recipes/<recipe>:42:in `from_file'

cron("clean_db_archive_logs") do
  action :create
  retries 0
  retry_delay 2
  default_guard_interpreter :default
  minute "0"
  hour "6"
  day "*"
  month "*"
  weekday :tuesday
  command "/var/lib/postgresql/backup.py"
  user "postgres"
  [...]
  declared_type :cron
  cookbook_name "<cb>"
  recipe_name "<recipe>"
end
```

I then changed `weekday :tuesday` to `weekday '2'`, re-knife'd and re-ran chef-client (ver 12.0.3), which succeeded. It appears that the cron resource doesn't like mixed input for the times that a cron job will run.

tas50:
@Vasu1105 Let's make sure that this issue is resolved in both the cron and cron_d resources.

Vasu1105:
Sure @tas50 we will take care of it. Currently Antima from MSys is working on this. So I was trying to assign her this issue but I can't see her in our Assignees list. And it mistakenly got assigned to me. Informed this to @btm. And we will definitely will take care of cron_d too. Thank you.

tas50:
@Vasu1105 You should probably take a look at the cron_d weekday property. It does validation using a native property instead of a crazy set_or_return mess like cron. You'd have to coerce the symbols to strings and then the validation should do the rest. It was something I meant to do for chef 16, but never got around to do doing it.

Issue #7060, opened 2018-03-26
Convert cron resource to setup properties via property DSL


tas50:
The cron resource is still using set_or_return. We should convert this to setup properties via the property resource DSL.

tas50:
To expand upon this we have 2 cron resources right now that both have similar, but different properties for setting things like day, week, etc. We need to unify this code so that both behave the same and don't duplicate code.

tas50:
Properly solved by #9522

PR #9829, opened 2020-05-08
Cron and Cron_d resource weekday property fixes

antima-gupta:

## Description
- Cron resource symbols for weekday are broken.
- Fixed weekday property for cron_d and cron resources
- It will support weekday values: `0-7`, `mon-sun`, `monday-sunday`, or `*` .
- Also weekday property will be `full name` and or `short name` in a Symbol or String(for ex:-  `:mon`, `'mon'`, `'monday'`, `:monday` etc..)

## Related Issue
Fixes: #3351 
           #7060 

## Types of changes

## Checklist:


lib/chef/resource/helpers/cron.rb
```diff
@@ -0,0 +1,19 @@
...
+      WEEKDAYS = { sunday: '0', monday: '1', tuesday: '2', wednesday: '3', thursday: '4', friday: '5', saturday: '6',
+                    sun: '0', mon: '1', tue: '2', wed: '3', thu: '4', fri: '5', sat: '6'
+                  }
+
+      # Convert weekday input value into crontab format that
+      # could be written in the crontab
+      # @return [Integer, String] A weekday formed as per the user inputs.
+      def weekday_in_crontab(wday)
+        weekday = wday.to_s.downcase
+        weekday_in_crontab = WEEKDAYS[weekday.to_sym] || wday
```
jaymzh, line 13: nit: why do half the conversions in one line and the other on the other?
```suggestion
        weekday = wday.to_s.downcase.to_sym
        weekday_in_crontab = WEEKDAYS[weekday] || wday
```

lib/chef/resource/helpers/cron.rb
```diff
@@ -0,0 +1,19 @@
+class Chef
+  module ResourceHelpers
+    module Cron
+      WEEKDAYS = { sunday: '0', monday: '1', tuesday: '2', wednesday: '3', thursday: '4', friday: '5', saturday: '6',
+                    sun: '0', mon: '1', tue: '2', wed: '3', thu: '4', fri: '5', sat: '6'
+                  }
```
jaymzh, line 6: nit: a good rule of thumb is - either both braces line by themself, or none should be. Personally I'd go:

```
      WEEKDAYS = {
        sunday: '0', monday: '1', tuesday: '2', wednesday: '3', thursday: '4', friday: '5', saturday: '6',
        sun: '0', mon: '1', tue: '2', wed: '3', thu: '4', fri: '5', sat: '6'
      }
```

spec/functional/resource/cron_spec.rb
```diff
@@ -80,6 +80,16 @@ def cron_should_not_exists(cron_name)
       5.times { new_resource.run_action(:create) }
       cron_should_exists(new_resource.name, new_resource.command)
     end
+    
+    # Test cron for day of week
+    weekdays = { Mon: 1, tuesday: 2, '3': 3, 'thursday': 4, 'Fri': 5, 6 => 6 }
+    weekdays.each do|key, value|
+      it "should create crontab entry and set #{value} for mon as weekday" do
```
jaymzh, line 87: for `mon`? left over? You're using more than mondays...

tas50:
Let's take this to the next level. We introduced resource partials in Chef Infra 16 that let us clean this up further without the need for helpers.

We did this with the git/subversion resources here:

https://github.com/chef/chef/tree/master/lib/chef/resource/scm

We can create a cron directory that contains the cron and cron_d resources and a new _cron_shared.rb resource partial that includes the common properties and common validators. Then you just include it with use like we do here: https://github.com/chef/chef/blob/master/lib/chef/resource/scm/git.rb#L24

lib/chef/resource/helpers/cron.rb
antima-gupta: @jaymzh Agreed, I have updated it.

spec/functional/resource/cron_spec.rb
antima-gupta: @jaymzh yeah I missed to update it. I have updated.

lib/chef/resource/helpers/cron.rb
antima-gupta: @jaymzh Agreed, I updated the same.

antima-gupta:
> Let's take this to the next level. We introduced resource partials in Chef Infra 16 that let us clean this up further without the need for helpers.
> 
> We did this with the git/subversion resources here:
> 
> https://github.com/chef/chef/tree/master/lib/chef/resource/scm
> 
> We can create a cron directory that contains the cron and cron_d resources and a new _cron_shared.rb resource partial that includes the common properties and common validators. Then you just include it with use like we do here: https://github.com/chef/chef/blob/master/lib/chef/resource/scm/git.rb#L24

@tas50 Command property has different parameters like In cron resource we have **identity** parameter and in cron_d resource we have **required** parameter.We can add both parameters in cron as well as cron_d resource but If we add **required** parameter then it will become mandatory earlier for cron resource **command** property was not mandatory it can be nil so it will through exception i.e. `Chef::Exceptions::ValidationFailed: command is a required property`
Do we need to make command property common or keep it as currently?

tas50:
I was wondering the other day why command wasn't mandatory. It really should be. It's probably just a historical thing. When we added that resource originally we probably didn't have the ability to make properties required.

lib/chef/resource/cron/_cron_shared.rb
```diff
@@ -0,0 +1,102 @@
+unified_mode true
```
tas50, line 1: We should leave this in the invidual resources

lib/chef/resource/cron/_cron_shared.rb
```diff
@@ -0,0 +1,102 @@
...
+property :user, String,
+  description: "The name of the user that runs the command.",
+  default: "root"
+
+property :environment, Hash,
+  description: "A Hash containing additional arbitrary environment variables under which the cron job will be run in the form of `({'ENV_VARIABLE' => 'VALUE'})`.",
+  default: lazy { {} }
+
+
+TIMEOUT_OPTS = %w{duration preserve-status foreground kill-after signal}.freeze
```
tas50, line 65: let's put all the constants together

lib/chef/resource/cron/cron.rb
```diff
@@ -0,0 +1,50 @@
...
+      provides :cron
+
+      description "Use the **cron** resource to manage cron entries for time-based job scheduling. Properties for a schedule will default to * if not provided. The cron resource requires access to a crontab program, typically cron."
+
+      state_attrs :minute, :hour, :day, :month, :weekday, :user
+
+      default_action :create
+      allowed_actions :create, :delete
+
+      def initialize(name, run_context = nil)
```
tas50, line 38: We should be able to kill this whole thing now

spec/unit/provider/cron_spec.rb
```diff
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
+require 'pry'
```
tas50, line 18: nuke this

lib/chef/resource/cron/_cron_shared.rb
```diff
@@ -0,0 +1,102 @@
...
+property :hour, [Integer, String],
+  description: "The hour at which the cron entry is to run (`0 - 23`).",
+  default: "*", callbacks: {
+    "should be a valid hour spec" => ->(spec) { Chef::ResourceHelpers::CronValidations.validate_numeric(spec, 0, 23) },
+  }
+
+property :day, [Integer, String],
+  description: "The day of month at which the cron entry should run (`1 - 31`).",
+  default: "*", callbacks: {
+    "should be a valid day spec" => ->(spec) { Chef::ResourceHelpers::CronValidations.validate_numeric(spec, 1, 31) },
```
tas50, line 23: We should be able to move all the validation files out of the helper file and into this shared file. The format changes a bit, but if you look back in the history you can find where I pulled that into the shared file. It would be nice to kill off that helper and just include all those methods here. That will also require moving around some rspecs most likely.

lib/chef/resource/cron/_cron_shared.rb
```diff
@@ -0,0 +1,102 @@
...
+  introduced: "15.7",
+  coerce: proc { |h|
+    if h.is_a?(Hash)
+      invalid_keys = h.keys - TIMEOUT_OPTS
+      unless invalid_keys.empty?
+        error_msg = "Key of option time_out must be equal to one of: \"#{TIMEOUT_OPTS.join('", "')}\"!  You passed \"#{invalid_keys.join(", ")}\"."
+        raise Chef::Exceptions::ValidationFailed, error_msg
+      end
+      unless h.values.all? { |x| x =~ TIMEOUT_REGEX }
+        error_msg = "Values of option time_out should be non-empty string without any leading whitespaces."
```
phiggins, line 85:
```suggestion
        error_msg = "Values of option time_out should be non-empty strings without any leading whitespace."
```

lib/chef/resource/cron/_cron_shared.rb
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+
+private
+# Convert weekday input value into crontab format that
+# could be written in the crontab
+# @return [Integer, String] A weekday formed as per the user inputs.
+def weekday_in_crontab(wday)
+  weekday = wday.to_s.downcase.to_sym
+  weekday_in_crontab = WEEKDAYS[weekday] || wday
```
phiggins, line 101:
```suggestion
  WEEKDAYS[weekday] || wday
```

lib/chef/resource/helpers/cron_validations.rb
```diff
@@ -62,13 +62,16 @@ def validate_month(spec)
...
+      # Added crontab param to check cron resource
       # @param spec the value to validate
       # @return [Boolean] valid or not?
       def validate_dow(spec)
+        spec = spec.to_s
         spec == "*" ||
           validate_numeric(spec, 0, 7) ||
-          %w{sun mon tue wed thu fri sat}.include?(String(spec).downcase)
+          %w{sun mon tue wed thu fri sat}.include?(String(spec).downcase) ||
+          %w{sunday monday tuesday wednesday thursday friday saturday}.include?(String(spec).downcase)
```
phiggins, line 74:
```suggestion
          %w{sun mon tue wed thu fri sat}.include?(spec.downcase) ||
          %w{sunday monday tuesday wednesday thursday friday saturday}.include?(spec.downcase)
```

Since this is being converted to a string on the first line of the method, it doesn't need to be converted again.

lib/chef/resource/cron/_cron_shared.rb
antima-gupta: @tas50 I tried to find  in the history but I was not able to see anything related to helper.Do I need to move all the methods in `cron_validation` helper to the shared partial?

lib/chef/resource/helpers/cron_validations.rb
antima-gupta: Agreed, I have updated this.

lib/chef/resource/cron/_cron_shared.rb
antima-gupta: Agreed, I have fixed this.

lib/chef/resource/cron/_cron_shared.rb
tas50: You can just try to move the helpers into this resource partial and use them directly. There's no need to have a partial and the helpers.

tas50:
@antima-gupta can you rebase this so we can see if the kitchen test will pass now that we fixed the issue you're hitting in testing

antima-gupta:
@tas50 I have rebase this branch.I still need to work on move the helpers into this resource partial.Currently I am working on priority task i.e. Chef Workstation 2020 Testing after finish this, I'll continue on this.

antima-gupta:
@tas50 I have moved validations into the partial file.Please have a look and let me know if any changes required.

tas50:
I pulled this locally, rebased it and merged it to master. Thanks @antima-gupta

antima-gupta:
@tas50 I have fixed chef_spec failure for cron resource in the PR: [990](https://github.com/chefspec/chefspec/pull/990). Please have a look.

tas50:
I realized that moving the validators into the partial was actually the wrong call since we use those in chef_client_cron and we don't want to include the extra properties there. I opened a new PR with these changes but I reverted that last part. I'm going to merge that once the tests pass. Sorry for the extra work there. I forgot all the consumers of those validators.

antima-gupta:
Thanks @tas50 for update.

PR #9833, opened 2020-05-09
Add ssl_verify option for remote_file

### LINKED ISSUE #8897: Enable turning off ssl validation for single `remote_file` resource
### Describe the Enhancement:
Currently people can either turn off SSL validation for the entire client, or not. That means if you need to `remote_file` a single https endpoint with a self-signed cert, most people end up turning off all validation and thus being even more insecure than they need to be.

A property to control `:ssl_verify_mode` for that specific resource would be better.

### Describe the Need:
```
remote_file 'https://someserver.com/thing' do
  ssl_verify_mode :verify_none
end
```

### Current Alternative
I've tried various alternatives like:

```
ruby_block 'turn off ssl verify' do
  block { Chef::Config[:ssl_verify_mode] = :verify_none }
end
remote_file 'https://someserver.com/thing' do
  ...
end
ruby_block 'turn on ssl viery' do
  block { Chef::Config[:ssl_verify_mode] = :verify_peer }
end
```
but didn't have success.

If I get time I will poke at it...

jaymzh:
## Description

Different servers have different https requirements and enforcing
the API policy on all `remote_file` resources isn't reasonable.

The logic around the HTTP clients and policies in Chef is... complex.
This approach seemed like the best one, but I'm open to others.

By default here if the user specifies nothing, `remote_file`'s http
clients will fall back to the API policy, otherwise, it'll use whatever
the specify.

And yes, Tim, I can add tests, but first I want to make sure this
approach is palatable. :)

## Related Issue

This fixes #8897

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/http/ssl_policies.rb
+++ b/lib/chef/http/ssl_policies.rb
@@ -129,5 +129,23 @@ class Chef
       end
     end
 
+    # This policy is used when we want to explicitly turn on verification
+    # for a specific request regardless of the API Policy. For example, when
+    # doing a `remote_file` where the user specified `verify_mode :verify_peer`
+    class VerifyPeerSSLPolicy < DefaultSSLPolicy
+      def set_verify_mode
+        http_client.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      end
+    end
+
+    # This policy is used when we want to explicitly turn off verification
+    # for a specific request regardless of the API Policy. For example, when
+    # doing a `remote_file` where the user specified `verify_mode :verify_none`
+    class VerifyNoneSSLPolicy < DefaultSSLPolicy
+      def set_verify_mode
+        http_client.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      end
+    end
+
   end
 end
--- a/lib/chef/resource/remote_file.rb
+++ b/lib/chef/resource/remote_file.rb
@@ -94,6 +94,9 @@ class Chef
 
       property :show_progress, [ TrueClass, FalseClass ], default: false
 
+      property :ssl_verify_mode, equal_to: %i{verify_none verify_peer}, default: nil,
+        description: "Optional property to override SSL policy. If not specified, uses the SSL polify from config.rb."
+
       property :remote_user, String
 
       property :remote_domain, String
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -304,7 +304,8 @@ class Chef
 
         SocketlessChefZeroClient.new(base_url)
       else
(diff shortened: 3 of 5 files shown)


lib/chef/resource/remote_file.rb
```diff
@@ -94,6 +94,8 @@ def use_conditional_get(true_or_false)
 
       property :show_progress, [ TrueClass, FalseClass ], default: false
 
+      property :ssl_verify_mode, [ :verify_none, :verify_peer, nil ], default: nil
```
tas50, line 97: This should be type Symbol with equal_to set to the values. That way it generates the documentation correctly and still gives nice messages. Also even though the other properties don't have one there should be a description field and an introduced field here (16.1 probably) so we can generate all the docs.

jaymzh: Can do. But the overall approach is sane? If so, I'll do that, plus add tests.

lib/chef/http/ssl_policies.rb
```diff
@@ -129,5 +129,17 @@ def set_verify_mode
       end
     end
 
+    class VerifyPeerSSLPolicy < DefaultSSLPolicy
```
tas50, line 132: Can we get at least a basic comment here on the intent of these classes and where they'll get used. Help future folks out. I realize this entire file is uncommented, but we need to start somewhere

lib/chef/resource/remote_file.rb
tas50: Also this doesn't need nil. That comes for free

lamont-granquist: the overall approach of this is no less sane than the existing code...

jaymzh:
OK @tas50 and @lamont-granquist - I added tests to the policy classes themselves, as well as updated for all of the things requested in the review.

However, I could not figure out how to add a test to remote_file specifically to look at this. `http_class_opts` doesn't get called until fetch time, and even when I made it public I couldn't call it directly, and i didn't see a reasonable way to test it without significant surgery, but i could have missed something.

jaymzh:
The test failures seem to be all dnf stuff and unrealted.

jaymzh:
@tas50 and @lamont-granquist - thoughts?

tas50:
Sorry for not keeping you up to date on this. I'm holding back net-new features right now so we can get a 16.1 hotfix out with the rubygems performance fix and some bootstrap fixes. After that I plan to merge this.

jaymzh:
Got it, thanks! I just wanted to know if you had any change requests while I had a day off to do them. :) Sounds like it's good to go. Thanks for the update!

PR #9834, opened 2020-05-10
Move dist implementation into chef-utils

bobchaos:

## Description
All dist contants are moving to chef-utils under new namespaces for each product, ie: `ChefUtils::Dist::Infra::PRODUCT`. 

~~All that remains to be done is to go through all the relevant files' `require` statement to elemininate duplicate `require "chef-utils"` statement generated by automation and require only dist where other chef-utils features aren't in use.~~ Being enormous, I figured I'd open the PR sooner than later. Sorry :/

## Related Issue
As discussed on slack with @lamont-granquist 

## Types of changes

## Checklist:

- [N/A] I have updated the documentation accordingly. - This new implementation may be worth documenting, lemme know if you think so and I'll find some place for it in the docs
- [N/A] I have added tests to cover my changes.


bobchaos:
This should be ready to process but I'd like to hold off as long as possible on merging it so I can write a blog post or something informing redistributors that we're about to break everything o.O

bobchaos:
I'm seeing test failures here, but no recent PR is green... Should I worry? FWIW I looked through them and found nothing that screams "This is because of the dist implementation", and I have no error that are not also present in a PR dated a few hours before this one.

lib/chef/application.rb
```diff
@@ -27,7 +27,7 @@
 require "tmpdir" unless defined?(Dir.mktmpdir)
 require "rbconfig"
 require_relative "application/exit_code"
-require_relative "dist"
+require "chef-utils"
```
tas50, line 30: Let's require the specific part of chef-utils we need vs. the whole gem

chef-utils/lib/chef-utils/dist/org.rb
```diff
@@ -0,0 +1,15 @@
+module ChefUtils
```
tas50, line 1: Can we keep this class structure, but stick it all into a singel file so it's easier to read and require?

chef-utils/lib/chef-utils.rb
```diff
@@ -29,6 +29,7 @@
 require_relative "chef-utils/dsl/which"
 require_relative "chef-utils/dsl/windows"
 require_relative "chef-utils/mash"
+require_relative "chef-utils/dist"
```
tas50, line 32: We can skip this part and continue to require dist when we need it within the specific ruby files

lib/chef/application.rb
bobchaos: where I put a plain `require "chef-utils"` as opposed to `chef-utils/dist` it's because there was another chef-utils function being used, in this case `.windows?`. I could make 2, more specific require statements tho, would that be better?

chef-utils/lib/chef-utils/dist/org.rb
bobchaos: arg, I had it like that at one point too XD I'll fix it, should be much easier now that I've done an initial pass.

tas50:
@bobchaos can you rebase this. Merging #9878 with the dist fix broke this.

ramereth:
@tas50 @lamont-granquist I'm working on rebasing and updating this PR and was wondering if we should provide backwards compatibility support for ``lib/chef/dist.rb`` so give us a migration path for downstream gems. We could replace the references in that file to use their ``ChefUtils::Dist`` counter parts. The current version of this PR completely removes ``lib/chef/dist.rb``. Not sure if we should add some kind of deprecation warning or not. I figure once this gets merged and released, we can work and fix any downstream gems that depend on this.

Also, should all of the references in ``chef-config/lib/chef-config/dist.rb`` be migrated to ``chef-utils/lib/chef-utils/dist.rb`` as well? And the same question above regarding a migration path.

Please let me know what you think.

jgitlin-p21:
My personal vote would be to deprecate `lib/chef/dist.rb` and `chef-config/lib/chef-config/dist.rb` by keeping the files but adding a deprecation notice, and making their values equal to the `ChefUtils::Dist` counter parts, with the `lib/chef/dist.rb` and `chef-config/lib/chef-config/dist.rb` files to be removed completely in a future version.

Just my $0.02

ramereth:
FWIW I couldn't find any direct references to ``Chef::Dist`` or ``ChefConfig::Dist`` outside of this repository, so it might be safe to completely remove/deprecate both of those.

I did some additional investigation of all the places where ``dist.rb`` exists and found the following references:
- https://github.com/chef/omnitruck/blob/ab4ca84865decd69d2c41a609bc9e3bf43bc5019/lib/dist.rb
- https://github.com/chef/mixlib-install/blob/13bbdce54ef9e48a0cf25bccc4152bf76abbfaaa/lib/mixlib/install/dist.rb
- https://github.com/chef/chef-cli/blob/f521a80e3be037eede70cf7734326d82f4b5d881/lib/chef-cli/dist.rb
- https://github.com/chef/chef-zero/blob/ed38f23fef793f73fb690e5ea59bf32b57a994c5/lib/chef_zero/dist.rb
- https://github.com/chef/chef-dk/blob/67c02e3a500efaaef5337caed503296fd443ede7/lib/chef-dk/dist.rb
- https://github.com/inspec/inspec/blob/a35d37043e65ce5fd528d58cd48ecb369c74a432/lib/inspec/dist.rb

I know that we're replacing chef_zero and inspec already in this PR, but I'm wondering if all of those should eventually be migrated here as well(well, except for chef-dk since that's being deprecated). I'd be curious what others think. That would create including chef-utils as a gem dependency in each of those projects.

jgitlin-p21:
> That would create including chef-utils as a gem dependency in each of those projects

From these comments in the `#chef-infra-dev` slack channel:

> **tas50:** chef-config is parsing out the config files.
> chef-utils is various helpers that can be used throughout our products.
> the OS detection stuff is all in there
> which / shell_out helpers are in there
> it's a dep free helper collection
> **lamont:** chef-utils is designed to not have any dependencies on anything else to avoid circular deps and to be broadly usable across literally everything

It seems to me like the answer is "Yes, a dependency is fine and `chef-utils` is the proper place so all of those should eventually be migrated here as well" but again just adding my $0.02 😆

lamont-granquist:
Yes, we should kill off the exiting locations.  I think/hope that nobody has deps on those, definitely I wouldn't consider that a public API yet, we're still figuring it all out.  And yes, we need to have all the other dist.rb's removed and those projects need to take a dep on chef-utils and use the dist constants out of chef-utils.

That is mildly terrifying, and would suggest we do this in a few phases:

1. omnitruck/mixlib-install/chef-zero -- these are obviously relatively "safe"
2. chef-cli -- this has tighter integration with chef-client already, but is "bigger"
3. inspec -- this will be a whole new world

You can probably omit chef-dk since it is basically dead now.

lib/chef/application.rb
ramereth: I just realized I may have reverted his change when I rebased this. Do you want me to split this up to requiring ``chef-utils/dist`` and ``chef-utils/dsl/windows`` like it was before?

lib/chef/application/client.rb
```diff
@@ -21,6 +21,7 @@
 require_relative "../handler/error_report"
 require_relative "../workstation_config_loader"
 require "uri" unless defined?(URI)
+require "chef-utils"
```
ramereth, line 24: Should I change this to the following?
```suggestion
require "chef-utils/dsl/windows"
require "chef-utils/dist"
```

lib/chef/application/solo.rb
```diff
@@ -21,6 +21,7 @@
 require_relative "client"
 require "fileutils" unless defined?(FileUtils)
 require "pathname" unless defined?(Pathname)
+require "chef-utils"
```
ramereth, line 24: Should I change this to the following?
```suggestion
require "chef-utils/dsl/windows"
require "chef-utils/dist"
```

lib/chef/client.rb
```diff
@@ -55,7 +55,6 @@
 require "chef-utils" unless defined?(ChefUtils::CANARY)
 require "ohai" unless defined?(Ohai::System)
 require "rbconfig" unless defined?(RbConfig)
-require_relative "dist"
```
ramereth, line 58: Looks like maybe I should have this changed to the following?
```ruby
require "chef-utils/dist"
```

lib/chef/event_loggers/windows_eventlog.rb
```diff
@@ -19,7 +19,7 @@
 require_relative "base"
 require_relative "../platform/query_helpers"
 require_relative "../win32/eventlog"
-require_relative "../dist"
+require "chef-utils"
```
ramereth, line 22: Should this be?
```suggestion
require "chef-utils/dsl/windows"
require "chef-utils/dist"
```

lib/chef/knife.rb
```diff
@@ -32,12 +32,12 @@
 require_relative "http/http_request"
 require_relative "http"
 require "pp" unless defined?(PP)
-require_relative "dist"
+require "chef-utils"
```
ramereth, line 35: This should probably be the following, correct?
```suggestion
require "chef-utils/dist"
```

lib/chef/local_mode.rb
```diff
@@ -15,9 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "chef-utils"
```
ramereth, line 18: Should this be?
```suggestion
require "chef-utils/dist"
```

lib/chef/provider/file.rb
```diff
@@ -31,7 +31,7 @@
 require_relative "../util/diff"
 require_relative "../util/selinux"
 require_relative "../file_content_management/deploy"
-require_relative "../dist"
+require "chef-utils"
```
ramereth, line 34: Should this be?
```suggestion
require "chef-utils/dist"
```

spec/functional/event_loggers/windows_eventlog_spec.rb
```diff
@@ -19,6 +19,7 @@
 require "spec_helper"
 require "securerandom"
 require "chef/event_loggers/windows_eventlog"
+require "chef-utils"
```
ramereth, line 22: ?
```suggestion
require "chef-utils/dist"
```

spec/integration/solo/solo_spec.rb
```diff
@@ -5,10 +5,10 @@
 require "chef/config"
 require "timeout"
 require "fileutils"
+require "chef-utils"
```
ramereth, line 8: ?
```suggestion
require "chef-utils/dist"
```

lib/chef/application/client.rb
lamont-granquist: nope

lib/chef/client.rb
lamont-granquist: yeah you should need to change that, that file shouldn't exist so something should be blowing up in tests?

ramereth: Actually this is covered with via the require on ``chef-utils`` above so this can be ignored.

lib/chef/application.rb
ramereth: I think based on the conversation I had with @lamont-granquist this morning, this is probably better. to leave as it.

ramereth:
@tas50 @lamont-granquist I believe this is ready for another review. I'm not sure if the failing tests are related or not.

phiggins:
> I'm not sure if the failing tests are related or not.

The failing unit tests definitely look related:

```
  1) ChefConfig::Config#var_chef_path when forced to unix etc_chef_dir is /etc/chef
     Failure/Error: let (:dirname) { ChefConfig::Dist::DIR_SUFFIX }
 
     NameError:
       uninitialized constant ChefConfig::Dist
       Did you mean?  Digest
     # ./spec/unit/config_spec.rb:228:in `block (3 levels) in <top (required)>'
     # ./spec/unit/config_spec.rb:268:in `block (4 levels) in <top (required)>'
```

https://buildkite.com/chef-oss/chef-chef-master-verify/builds/6050#65521c8e-b41b-4803-906b-73ddaf9b14f3/88-1794

Rakefile
```diff
@@ -26,7 +26,7 @@ begin
   require_relative "tasks/announce"
   require_relative "tasks/docs"
   require_relative "tasks/spellcheck"
-  require_relative "lib/chef/dist"
+  require_relative "chef-utils/lib/chef-utils/dist"
```
tas50, line 29: you might want to slack the defined? bit in here to avoid that chefstyle issue when we bump that in chef/chef

lamont-granquist (approved):

PR #9847, opened 2020-05-12
Fix dnf_package version and arch property support and idempotency

### LINKED ISSUE #9763: dnf_package (and most likely yum_package) issues with version property
These all work correctly:

```
dnf_package "zabbix-agent-4.0.15"
dnf_package "zabbix-agent-4.0*"
dnf_package "zabbix-agent-4.0.15-1.fc31"
dnf_package "zabbix-agent-4.0*-1.fc31"
dnf_package "zabbix-agent-4.0.15-*"
dnf_package "zabbix-agent-4.0*-*"
dnf_package "zabbix-agent-1:4.0.15"
dnf_package "zabbix-agent-1:4.0.*"
dnf_package "zabbix-agent-1:4.0.15-1.fc31"
dnf_package "zabbix-agent-1:4.0*-1.fc31"
dnf_package "zabbix-agent-*:4.0.15-*"
dnf_package "zabbix-agent-*:4.0*-*"
```

These do all do not work (but correctly so):

```
#dnf_package "zabbix-agent-4.0"
#dnf_package "zabbix-agent-4.0.15-1"
#dnf_package "zabbix-agent-1:4.0.15-1"
#dnf_package "zabbix-agent-*:4.0.15-1"
#dnf_package "zabbix-agent-*:4.0.15-1.fc31"
```

The CLI does not support those five either:

```
[root@default-fedora-latest vagrant]# dnf install zabbix-agent-4.0
Last metadata expiration check: 0:20:31 ago on Wed 29 Apr 2020 01:09:14 AM UTC.
No match for argument: zabbix-agent-4.0
Error: Unable to find a match: zabbix-agent-4.0
[root@default-fedora-latest vagrant]# dnf install zabbix-agent-4.0.15-1
Last metadata expiration check: 0:20:37 ago on Wed 29 Apr 2020 01:09:14 AM UTC.
No match for argument: zabbix-agent-4.0.15-1
Error: Unable to find a match: zabbix-agent-4.0.15-1
[root@default-fedora-latest vagrant]# dnf install zabbix-agent-1:4.0.15-1
Last metadata expiration check: 0:20:44 ago on Wed 29 Apr 2020 01:09:14 AM UTC.
No match for argument: zabbix-agent-1:4.0.15-1
Error: Unable to find a match: zabbix-agent-1:4.0.15-1
[root@default-fedora-latest vagrant]# dnf install 'zabbix-agent-*:4.0.15-1'
Last metadata expiration check: 0:21:06 ago on Wed 29 Apr 2020 01:09:14 AM UTC.
No match for argument: zabbix-agent-*:4.0.15-1
Error: Unable to find a match: zabbix-agent-*:4.0.15-1
[root@default-fedora-latest vagrant]# dnf install 'zabbix-agent-*:4.0.15-1.fc31'
Last metadata expiration check: 0:21:16 ago on Wed 29 Apr 2020 01:09:14 AM UTC.
No match for argument: zabbix-agent-*:4.0.15-1.fc31
Error: Unable to find a match: zabbix-agent-*:4.0.15-1.fc31
```

When used as a version property the only use case which works entirely correctly is the fully specified case:

```
dnf_package "zabbix-agent" do
  version "1:4.0.15-1.fc31"
end
```

When used as a version property the following use case crash hard:

```
dnf_package "zabbix-agent" do
  version "*:4.0.15-*"
end

dnf_package "zabbix-agent" do
  version "-*:4.0*-*"
end
```

The exception is probably fixable and is:

```
[2020-04-29T01:35:31+00:00] FATAL: RuntimeError: dnf_package[zabbix-agent] ((chef-apply cookbook)::(chef-apply recipe) line 7) had an error: RuntimeError: dnf-helper.py had stderr/stdout output:

  File "/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.0.257/lib/chef/provider/package/dnf/dnf_helper.py", line 166, in <module>
    query(command)
  File "/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.0.257/lib/chef/provider/package/dnf/dnf_helper.py", line 101, in query
    q = q.filterm(epoch=int(command['epoch']))
ValueError: invalid literal for int() with base 10: '-*'

During handling of the above exception, another exception occurred:

  File "/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.0.257/lib/chef/provider/package/dnf/dnf_helper.py", line 173, in <module>
    base.closeRpmDB()
AttributeError: 'Base' object has no attribute 'closeRpmDB'
```

When used as a version property the following uses cases appear to work but are not idempotent:

```

dnf_package "zabbix-agent" do
  version "4.0.15"
end

dnf_package "zabbix-agent" do
  version "4.0*"
end

dnf_package "zabbix-agent" do
  version "4.0.15-1.fc31"
end

dnf_package "zabbix-agent" do
  version "4.0*-1.fc31"
end

dnf_package "zabbix-agent" do
  version "4.0.15-*"
end

dnf_package "zabbix-agent" do
  version "4.0*-*"
end

dnf_package "zabbix-agent" do
  version "1:4.0.15"
end

dnf_package "zabbix-agent" do
  version "1:4.0.*"
end

dnf_package "zabbix-agent" do
  version "1:4.0*-1.fc31"
end
```

Tested On:  Fedora release 31 (Thirty One)

Client version: Chef Infra Client: 16.0.257

Aha! Link: https://chef.aha.io/features/SH-111

lamont-granquist:
This does two things:

1.  Fixes the filtering by name, version + arch.  This introduces one incompatibility which is that putting the arch in the version property was supported at one point in time with a test that I committed which doesn't seem to be used anywhere.  But we also now should correctly support mixing and matching the name with versions and arches in them combined with a version property and/or arch property.  Which is a bit crazy but it supports doing `package "openssl.i686"` with a version property.  It should now behave the same using either the property or putting the version and arch into the name.

2.  It changes the meaning of the "current_version" in the superclass (or confusingly the "installed_version" in the subclass) so that if a version of the package is installed, which does not match the requested version+arch filtering, the value is set to nil.  This is done so that the superclass can check current_version.nil? in order to check to see if a "good" package is installed or not.  This changes the semantics of that variable where previously it reported whatever version was installed.  The subclass_directive that is introduced switches between those behaviors.  The result is 2 fewer comparison operations that the superclass asks of the subclass which also speeds up execution.

This fixes many edge conditions around idempotency, along with differences in the supported APIs and idemopotency behavior between specifying the version constraints in the name and specifying the version constraints in the version property of the resource.

There's an unrelated fix to exception handling and crashes.

closes  #9763

Aha! Link: https://chef.aha.io/features/SH-120

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -434,6 +434,8 @@ class Chef
               case action
               when :upgrade
                 if current_version.nil?
+                  # with installed_version_satisfies_version_constraints there may be a package installed, but it fails the user's
+                  # requested new_resource.version constraints
                   logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
                   target_version_array.push(candidate_version)
                 elsif version_equals?(current_version, new_version)
@@ -467,6 +469,8 @@ class Chef
                     target_version_array.push(new_version)
                   end
                 elsif current_version.nil?
+                  # with installed_version_satisfies_version_constraints there may be a package installed, but it fails the user's
+                  # requested new_resource.version constraints
                   logger.trace("#{new_resource} #{package_name} not installed, installing #{candidate_version}")
                   target_version_array.push(candidate_version)
                 else
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -53,6 +53,11 @@ class Chef
       # [ this may arguably be useful for all package providers and it greatly simplifies the logic
       #   in the superclass that gets executed, so maybe this should always be used now? ]
       #
+      # note that when using this feature that the current_resource.version must be loaded with the
+      # correct currently installed version, without doing the filtering -- for reporting and for
+      # correctly displaying version upgrades.  that means there are 3 different arrays which must be
+      # loaded by the subclass:  candidate_version, magic_version and current_resource.version.
+      #
       subclass_directive :use_magic_version
 
       #
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -58,6 +58,9 @@ class Chef
       # correctly displaying version upgrades.  that means there are 3 different arrays which must be
       # loaded by the subclass:  candidate_version, magic_version and current_resource.version.
       #
+      # NOTE: magic_version is a terrible name, but I couldn't think of anything better, at least this
+      #       way it stands out clearly.
+      #
       subclass_directive :use_magic_version
 
       #
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -28,13 +28,32 @@ class Chef
     class Package < Chef::Provider
       extend Chef::Mixin::SubclassDirective
 
-      # subclasses declare this if they want all their arguments as arrays of packages and names
+      # subclasses declare this if they want all their arguments as arrays of packages and names.
+      # any new packages using this should also use allow_nils below.
+      #
       subclass_directive :use_multipackage_api
-      # subclasses declare this if they want sources (filenames) pulled from their package names
+
+      # subclasses declare this if they want sources (filenames) pulled from their package names.
+      # this is for package providers that take a path into the filesystem (rpm, dpkg).
+      #
       subclass_directive :use_package_name_for_source
+
       # keeps package_names_for_targets and versions_for_targets indexed the same as package_name at
-      # the cost of having the subclass needing to deal with nils
+      # the cost of having the subclass needing to deal with nils.  all providers are encouraged to
+      # migrate to using this as it simplifies dealing with package aliases in subclasses.
+      #
       subclass_directive :allow_nils
+
+      # subclasses that implement complex pattern matching using constraints, particularly the yum and
+      # dnf classes, should filter the installed version against the desired version constraint and
+      # return nil if it does not match.  this means that 'nil' does not mean that no version of the
+      # package is installed, but that the installed version does not satisfy the desired constraints.
+      # (the package plus the constraints are not installed)
+      #
+      # [ this may arguably be useful for all package providers and it greatly simplifies the logic
+      #   in the superclass that gets executed, so maybe this should always be used now? ]
+      #
+      subclass_directive :installed_version_satisfies_version_constraints
 
       #
       # Hook that subclasses use to populate the candidate_version(s)
@@ -414,19 +433,19 @@ class Chef
             each_package do |package_name, new_version, current_version, candidate_version|
               case action
               when :upgrade
-                if version_equals?(current_version, new_version)
-                  # this is an odd use case
-                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
-                  target_version_array.push(nil)
-                elsif version_equals?(current_version, candidate_version)
-                  logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+                if current_version.nil?
+                  logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
+                  target_version_array.push(candidate_version)
+                elsif version_equals?(current_version, new_version)
+                  # this is a short-circuit to avoid needing to (expensively) query the candidate_version which must come later
+                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed")
                   target_version_array.push(nil)
                 elsif candidate_version.nil?
                   logger.trace("#{new_resource} #{package_name} has no candidate_version to upgrade to")
                   target_version_array.push(nil)
-                elsif current_version.nil?
-                  logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
-                  target_version_array.push(candidate_version)
+                elsif version_equals?(current_version, candidate_version)
+                  logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+                  target_version_array.push(nil)
                 elsif !allow_downgrade && version_compare(current_version, candidate_version) == 1
                   logger.trace("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{candidate_version}. Skipping...)")
                   target_version_array.push(nil)
@@ -436,15 +455,12 @@ class Chef
                 end
 
               when :install
-                if new_version
+                if new_version && !installed_version_satisfies_version_constraints?
                   if version_requirement_satisfied?(current_version, new_version)
                     logger.trace("#{new_resource} #{package_name} #{current_version} satisfies #{new_version} requirement")
                     target_version_array.push(nil)
                   elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
                     logger.warn("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{new_version}. Skipping...)")
-                    target_version_array.push(nil)
-                  elsif version_equals?(current_version, candidate_version)
-                    logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
                     target_version_array.push(nil)
                   else
                     logger.trace("#{new_resource} #{package_name} #{current_version} needs updating to #{new_version}")
@@ -511,8 +527,14 @@ class Chef
             each_package do |package_name, new_version, current_version, candidate_version|
               next if new_version.nil? || current_version.nil?
 
-              if !version_requirement_satisfied?(current_version, new_version) && candidate_version.nil?
-                missing.push(package_name)
+              if installed_version_satisfies_version_constraints?
+                if !current_version && candidate_version.nil?
+                  missing.push(package_name)
+                end
+              else
+                if !version_requirement_satisfied?(current_version, new_version) && candidate_version.nil?
+                  missing.push(package_name)
+                end
               end
             end
             missing
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -53,7 +53,7 @@ class Chef
       # [ this may arguably be useful for all package providers and it greatly simplifies the logic
       #   in the superclass that gets executed, so maybe this should always be used now? ]
       #
-      subclass_directive :installed_version_satisfies_version_constraints
+      subclass_directive :use_magic_version
 
       #
       # Hook that subclasses use to populate the candidate_version(s)
@@ -434,7 +434,7 @@ class Chef
               case action
               when :upgrade
                 if current_version.nil?
-                  # with installed_version_satisfies_version_constraints there may be a package installed, but it fails the user's
+                  # with use_magic_version there may be a package installed, but it fails the user's
                   # requested new_resource.version constraints
                   logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
                   target_version_array.push(candidate_version)
@@ -457,7 +457,7 @@ class Chef
                 end
 
               when :install
-                if new_version && !installed_version_satisfies_version_constraints?
+                if new_version && !use_magic_version?
                   if version_requirement_satisfied?(current_version, new_version)
                     logger.trace("#{new_resource} #{package_name} #{current_version} satisfies #{new_version} requirement")
                     target_version_array.push(nil)
@@ -469,7 +469,7 @@ class Chef
                     target_version_array.push(new_version)
                   end
                 elsif current_version.nil?
-                  # with installed_version_satisfies_version_constraints there may be a package installed, but it fails the user's
+                  # with use_magic_version there may be a package installed, but it fails the user's
                   # requested new_resource.version constraints
                   logger.trace("#{new_resource} #{package_name} not installed, installing #{candidate_version}")
                   target_version_array.push(candidate_version)
@@ -531,8 +531,8 @@ class Chef
             each_package do |package_name, new_version, current_version, candidate_version|
               next if new_version.nil? || current_version.nil?
 
-              if installed_version_satisfies_version_constraints?
-                if !current_version && candidate_version.nil?
+              if use_magic_version?
+                if !magic_version && candidate_version.nil?
                   missing.push(package_name)
                 end
               else
@@ -551,7 +551,7 @@ class Chef
       def each_package
         package_name_array.each_with_index do |package_name, i|
           candidate_version = candidate_version_array[i]
-          current_version = current_version_array[i]
+          current_version = use_magic_version? ? magic_version[i] : current_version_array[i]
           new_version = new_version_array[i]
           yield package_name, new_version, current_version, candidate_version
         end
--- a/lib/chef/provider/package/dnf/dnf_helper.py
+++ b/lib/chef/provider/package/dnf/dnf_helper.py
@@ -98,14 +98,24 @@ def query(command):
         q = q.available()
 
     if 'epoch' in command:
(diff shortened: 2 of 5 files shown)


breisig:
Can we get this merged.

lamont-granquist:
the "upgrade" tests include downgrades

lib/chef/provider/package.rb
```diff
@@ -414,19 +433,19 @@ def target_version_array
...
-                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
-                  target_version_array.push(nil)
-                elsif version_equals?(current_version, candidate_version)
-                  logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+                if current_version.nil?
+                  logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
+                  target_version_array.push(candidate_version)
+                elsif version_equals?(current_version, new_version)
+                  # this is a short-circuit to avoid needing to (expensively) query the candidate_version which must come later
+                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed")
```
jaymzh, line 441: I would put the warning back. pinning in :upgrade is weird and we still may kill it

lib/chef/provider/package.rb
```diff
@@ -414,19 +433,19 @@ def target_version_array
             each_package do |package_name, new_version, current_version, candidate_version|
               case action
               when :upgrade
-                if version_equals?(current_version, new_version)
-                  # this is an odd use case
-                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
-                  target_version_array.push(nil)
-                elsif version_equals?(current_version, candidate_version)
-                  logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+                if current_version.nil?
```
jaymzh, line 436: it's entirely unclear here how it's possible for current_version != new_version but current_version != candidate_version. I mean *I* get it because I've been talking to you about those stuff for weeks, but I would add a TON of comments here for future-Lamont and future-Phil.

lib/chef/provider/package/dnf/dnf_helper.py
```diff
@@ -98,14 +98,24 @@ def query(command):
         q = q.available()
 
     if 'epoch' in command:
-        q = q.filterm(epoch=int(command['epoch']))
+        if not dnf.util.is_glob_pattern(command['epoch']):
```
jaymzh, line 101: why drop this if it's a glob? Is there not a epoch__glob filter?

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -128,13 +127,114 @@ def pkg_arch
...
+          expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.2-1.#{pkg_arch}$")
+        end
+      end
+
+      %w{1.2 1.2-1 1.2-* 0:1.2 0:1.2-1 *:1.2-*}.each do |vstring|
+        it "is idempotent when #{vstring} is in the version property on upgrade and it doesn't match the candidate version" do
+          preinstall("chef_rpm-1.2-1.#{pkg_arch}.rpm")
+          dnf_package.package_name("chef_rpm")
+          dnf_package.version(vstring)
+          dnf_package.run_action(:upgrade)
```
jaymzh, line 209: this is a pretty big change... you're now saying that:
```
package 'foo' do
  version '1.2'
  action :install # or :upgrade
end
```

will work and install the latest-release of 1.2. That needs to be called out BIG TIME in the release notes. Before that would break.

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -715,32 +914,104 @@ def pkg_arch
...
+        dnf_package.version("1.10-1")
         dnf_package.run_action(:upgrade)
         expect(dnf_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be false
       end
 
-      it "with a prco equality pin in the name it upgrades a prior package" do
+      it "with a full version pin in the name it downgrdes the package" do
```
jaymzh, line 934:
```suggestion
      it "with a full version pin in the name it downgrades the package" do
```

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -715,32 +914,104 @@ def pkg_arch
...
+        dnf_package.package_name("chef_rpm")
+        dnf_package.version("1.10")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be true
+        expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be false
+      end
+
+      it "with a partial (no release) version pin in the name it downgrdes the package" do
```
jaymzh, line 956:
```suggestion
      it "with a partial (no release) version pin in the name it downgrades the package" do
```

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -715,32 +914,104 @@ def pkg_arch
...
+        preinstall("chef_rpm-1.2-1.#{pkg_arch}.rpm")
+        dnf_package.package_name("chef_rpm-1.10-1")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be true
+        expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be false
+      end
+
+      it "with a full version pin in the name it downgrdes the package" do
```
jaymzh, line 977:
```suggestion
      it "with a full version pin in the name it downgrades the package" do
```

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -715,32 +914,104 @@ def pkg_arch
...
+        dnf_package.package_name("chef_rpm-1.10")
         dnf_package.run_action(:upgrade)
         expect(dnf_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.10-1.#{pkg_arch}$")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be false
       end
 
-      it "with an equality pin in the name it downgrades a later package" do
+      it "with a partial (no release) version pin in the name it downgrdes the package" do
```
jaymzh, line 997:
```suggestion
      it "with a partial (no release) version pin in the name it downgrades the package" do
```

spec/functional/resource/dnf_package_spec.rb
```diff
@@ -715,32 +914,104 @@ def pkg_arch
...
         dnf_package.run_action(:upgrade)
         expect(dnf_package.updated_by_last_action?).to be true
         expect(shell_out("rpm -q --queryformat '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n' chef_rpm").stdout.chomp).to match("^chef_rpm-1.2-1.#{pkg_arch}$")
+        dnf_package.run_action(:upgrade)
+        expect(dnf_package.updated_by_last_action?).to be false
+      end
+
+      it "with a prco equality pin in the name it upgrades a prior package" do
+        preinstall("chef_rpm-1.2-1.#{pkg_arch}.rpm")
+        dnf_package.package_name("chef_rpm = 1.10")
```
jaymzh, line 1009: wat? That's a syntax?

jaymzh (commented):
seems generally OK.
This does not fix the open bug with multipackage, but I don't think you intended to.

lib/chef/provider/package/dnf/dnf_helper.py
lamont-granquist: no if its a glob, it is '*' so it matches anything so you don't filter.

i beat my head against that one for a bit.

i suppose i could try a comment.

spec/functional/resource/dnf_package_spec.rb
lamont-granquist: that's the bug that is being fixed.

spec/functional/resource/dnf_package_spec.rb
lamont-granquist: yep that's not new, this is just rearrangement from existing tests.  works the same way on the commandline

lamont-granquist: it also supports any `provides` lines in the RPMs, with that kind of version string syntax.

lib/chef/provider/package.rb
lamont-granquist: even if you delete this line because of the way that the candidate and installed versions are both filtered by the new_resource.version and/or the version constraints in the package_name it still works.  that line (at least for the DNF package provider now and soon the yum provider) is just a performance optimization now if you equality match (to avoid querying the candidate_version, which can be expensive in some of the subclasses).

lamont-granquist: (and we haven't deleted in that in the years that comment has been sitting there, i'm pretty confident at this point that we never will)

lib/chef/provider/package.rb
lamont-granquist: candidate_version and current_version are always fully materialized EVRs (in the yum/dnf subclasses) because they come back from queries.

the new_version comes from the user and can be "1.*" or whatever.  and "1.2-1" != "1.*".  i was going to go down the road of having a `installed_version.satisfied_by?(new_version)` API that could be called in the subclass, but that looked like it was going to require a lot of python code that would duplicate internal code in the dnf library.  making it so that the installed_version was just automatically satisfied_by the new_version (or else it being nil) was vastly simpler and will be more robust and gets what we actually need.  that way if either candidate_version or installed_version is nil it either doesn't exist, or doesn't satisfy the desired constraint and either way we have to do work to fix it.

lamont-granquist:
Note to self that I need to fix the condition when the installed_version will now be nil when the new_version doesn't match and the user will get green text about nil being upgrading/downgraded to the desired version, when the package was actually installed.

Similarly need to make sure that the "before" resource (current_resource) has the right version number for reporting.  So we need to populate it there and print that in the superclass for output to the user, while we use the installed_version in the subclass.  Existing subclasses generally set those to the same value so it shouldn't affect anything with other subclasses, they won't care.

lib/chef/provider/package/dnf/dnf_helper.py
jaymzh: what if it's a glob that's *not* that. what if my glob is `1?` ?

spec/functional/resource/dnf_package_spec.rb
jaymzh: I mean... sorta? You had to be explicit before. The bug you're fixing is that specifying arch in weird places would lead to lack of idempotency.
I'm not saying this is wrong, but this is definitely gonna lead to some unexpected surprises for people.

lib/chef/provider/package.rb
jaymzh: Oh I understand it. I'm just saying .... comment your damn code. :)

lib/chef/provider/package.rb
jaymzh: yeah I was totally wrong. This is a valid thing to do now. You might want to upgrade to whatever the latest release of 1.2 is

lib/chef/provider/package.rb
```diff
@@ -414,19 +433,19 @@ def target_version_array
...
+                  logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
+                  target_version_array.push(candidate_version)
+                elsif version_equals?(current_version, new_version)
```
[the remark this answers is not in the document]
epilatow: Well, I downloaded this code, monkeypatched it into our code base, ran my tests, and I'm happy to report that the tests succeeded. It seems we never tried to do the aforementioned broken comparisons because `current_version` was nil (even though there was a current version of the package installed). I think you tried to explain this earlier to me when you said:

> went a completely different direction so that the subclass can direct the superclass that the installed_version ("current_version" in the superclass) is nil if it doesn't match the version constraint.  so that value can be nil if there's a package installed but it doesn't match what the resource wants.

That said, while running chef in debug mode I am still seeing a lot of `versioncompare` calls, some with version numbers that we will parse incorrectly:
```
[2020-12-10T16:46:28-08:00] DEBUG: sending '{"action":"versioncompare","versions":["0:2020.2.41-80.0.el8_2.noarch","0:2020.2.41-80.0.el8_2.noarch"]}' to python helper
[2020-12-10T16:46:34-08:00] DEBUG: sending '{"action":"versioncompare","versions":["0:60.3-2.el8_1.x86_64","0:60.3-2.el8_1.x86_64"]}' to python helper
[2020-12-10T16:46:34-08:00] DEBUG: sending '{"action":"versioncompare","versions":["0:60.3-2.el8_1.x86_64","0:60.3-2.el8_1.x86_64"]}' to python helper
[2020-12-10T16:46:34-08:00] DEBUG: sending '{"action":"versioncompare","versions":["1:0.17-73.el8_1.1.x86_64","1:0.17-73.el8_1.1.x86_64"]}' to python helper
```

Most of these seem to be occurring with multi-package resources in the :upgrade path on line 446:
```
                 elsif version_equals?(current_version, candidate_version)
```

But in all these cases we're doing an equality comparison on identical strings, so it seems like things will work even though we're not parsing EVRAs correctly. It seems like your changes essentially eliminate all version comparisons (other than for equality comparisons for equal strings). If that's true then the dnf versioncompare code is dead, so why not remove it and update the dnf 'python_helper.rb`compare_versions()' function to assert v1 == v2 and always return 0? That'd make the code less complex and the constraints more obvious. (If that assumption is not true then it seems like there should be a way to make the version comparisons fail with the current parsing code.)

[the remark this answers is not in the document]
lamont-granquist: >  because current_version was nil (even though there was a current version of the package installed)

yep that's the point of `installed_version_satisfies_version_constraints` -- if it doesn't satisfy the version constraints then its nil, even though it is installed.

have to parse the rest of that tomorrow, but you're right that versioncompare is basically unimportant now.

lib/chef/provider/package/dnf/dnf_helper.py
lamont-granquist: looks like that is actually legal, but i don't know how to fix that since i don't see how the underlying dnf libraries do that globbing.

so that's just going to be known broken.  i can't imagine that is an actually useful use case though.  and the dnf libaries themselves have several bugs around handling wildcards in epoch versions (see the #9763 use cases which work and which don't).

PR #9863, opened 2020-05-13
Remove most of the emojis from the buildkite tests

btm:
Most of these emojis I can't remember so I spend longer waiting for the
mouse-overs that I did before when some were too long. :older_man:

Removed a duplicate Amazon Linux TK test

Shortened some text where I thought it made sense too

This is obviously a bikeshed but I really do find the emoji's difficult


tas50:
🤷

PR #9874, opened 2020-05-15
WIP: appbundle-updater

btm:
:construction:

### DIFF
--- a/kitchen-tests/kitchen.azure.yml
+++ b/kitchen-tests/kitchen.azure.yml
@@ -9,12 +9,30 @@ driver_config:
 
 provisioner:
   name: chef_zero
+  produce_name: chef
+  install_strategy: skip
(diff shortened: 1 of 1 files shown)


kitchen-tests/kitchen.azure.yml
```diff
@@ -9,13 +9,31 @@ driver_config:
 
 provisioner:
   name: chef_zero
+  produce_name: chef
```
lamont-granquist, line 12: produce?

btm:
This was timing out after 30 or 60 minutes or something. @christopher-snapp The Azure Pipelines mac build was doing this in a blazing unbelievable amount of speed for the mac stuff, so we should probably prioritize trying to do this on 2019 there and take advantage of the warm microsoft hosted queues.

PR #9886, opened 2020-05-20
Small cleanups to checksum logic in file/windows_package/chef::mixin::checksum

phiggins:
A few small things I noticed while reading code.

### DIFF
--- a/lib/chef/provider/file.rb
+++ b/lib/chef/provider/file.rb
@@ -334,7 +334,7 @@ class Chef
       end
 
       def do_validate_content
-        if new_resource.checksum && tempfile && ( new_resource.checksum.downcase != tempfile_checksum )
+        if new_resource.checksum && tempfile && ( new_resource.checksum != tempfile_checksum )
           raise Chef::Exceptions::ChecksumMismatch.new(short_cksum(new_resource.checksum), short_cksum(tempfile_checksum))
         end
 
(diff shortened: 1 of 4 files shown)


lib/chef/provider/package/windows.rb
```diff
@@ -38,7 +38,7 @@ class Windows < Chef::Provider::Package
         def define_resource_requirements
           if new_resource.checksum
             requirements.assert(:install) do |a|
-              a.assertion { new_resource.checksum.downcase == checksum(source_location) }
+              a.assertion { new_resource.checksum == checksum(source_location) }
```
phiggins, line 41: windows_package already does this coercion: https://github.com/chef/chef/blob/master/lib/chef/resource/windows_package.rb#L140-L142

lib/chef/provider/file.rb
```diff
@@ -334,7 +334,7 @@ def tempfile_checksum
       end
 
       def do_validate_content
-        if new_resource.checksum && tempfile && ( new_resource.checksum.downcase != tempfile_checksum )
```
tas50, line 337: are we coercing this in file as well?

phiggins: I did that in this PR https://github.com/chef/chef/pull/9886/commits/60f2c4a8dfed08656def3abef6f6a11fa2ef8f2f

Issue #9885, opened 2020-05-19
Custom resources with duplicate names silently fail


steelcowboy:
## Description
If two very similar custom resources are created that provide for different platforms, and then they are called in a unified way in a recipe, the resource itself will silently fail: it will say up-to-date on the action but not actually run anything else. 

See [the CPE Chrome default recipe](https://github.com/facebook/IT-CPE/blob/5e0bce5876e88faeb8a4d272566d2627af4423db/itchef/cookbooks/cpe_chrome/recipes/default.rb), and the two resources [cpe_chrome](https://github.com/facebook/IT-CPE/blob/5e0bce5876e88faeb8a4d272566d2627af4423db/itchef/cookbooks/cpe_chrome/resources/cpe_chrome.rb) and [cpe_chrome_win](https://github.com/facebook/IT-CPE/blob/5e0bce5876e88faeb8a4d272566d2627af4423db/itchef/cookbooks/cpe_chrome/resources/cpe_chrome_win.rb) for an example.

## Chef Version
Chef Infra Client: 16.1.0

## Platform Version
Fedora 32

## Replication Case
1. Create a cookbook with two custom resources
2. Give them the same `resource_name`, provide for different platform
3. See that instead of an error, it just says "up to date" and doesn't run the resource

## Client Output
https://gist.github.com/steelcowboy/d411bc080af9b46c11661c96a58e8753

## Stacktrace
N/A

btm:
@steelcowboy as I've commented over on https://github.com/facebook/IT-CPE/issues/243#issuecomment-631064185, I can't reproduce this.

Given this simple reproduction case:
```
btm@btm-thinkpad:/tmp/cookbooks$ find two_resources/
two_resources/
two_resources/resources
two_resources/resources/mirror_windows.rb
two_resources/resources/mirror_linux.rb
two_resources/recipes
two_resources/recipes/default.rb

btm@btm-thinkpad:/tmp/cookbooks$ cat two_resources/recipes/default.rb 
mirror_resource 'testing'

btm@btm-thinkpad:/tmp/cookbooks$ cat two_resources/resources/mirror_linux.rb 
resource_name 'mirror_resource'
provides :mirror_resource, :os => "linux"
default_action :test

action :test do
  file "/tmp/mirror_resource" do
    content "this is linux"
  end
end

btm@btm-thinkpad:/tmp/cookbooks$ cat two_resources/resources/mirror_windows.rb 
resource_name 'mirror_resource'
[...]

PR #9890, opened 2020-05-20
Chef-16.1 breaking change

lamont-granquist:
This is a breaking change to fix an intentionally breaking change in Chef 16.0 to remove the canonical DSL #9206 with unintended side effects #9885.  Breaking behavior differently, but correctly, without the side effects.

The new behavior is that a resource which relies upon the resource_name
statement to wire the resource up to the DSL will fail (although the
very old wiring of the "<cookbook_name>_<resource_filename>" continues
to work).

So for a resource which does not get its DSL wiring from the filename,
only declaring a resource_name will fail:

```
resource_name :foo
```

This can be fixed by adding an explicit `provides` line (which is
backwards compatible):

```
resource_name :foo
provides :foo
```

If backwards compatibility is not a concern, then post-16.0 the
resource_name call can be completely dropped:

```
provides :foo
```

This is a vastly simpler backwards compatibility break than the
unintentional break in #9885, which is difficult to describe and
to detect.

The rules going forward are fairly simple to explain:

1.  The resource_name now only sets the resource_name.
2.  The old behavior of the fallback resource_name and DSL wiring
    based on the filename is preserved (unless the values are set).
3.  In Chef 16, The first provides line will set the resource_name if
    it has not already been set, allowing the resource_name to be
    omitted.

It is recommended that all resources only set provides lines, the
use of the fallback filename-based wiring is discouraged and
explicitly setting the resource_name is discouraged.

### DIFF
--- a/spec/unit/resource_spec.rb
+++ b/spec/unit/resource_spec.rb
@@ -492,6 +492,20 @@ describe Chef::Resource do
       expect(r.resource_name).to eq :blah
       expect(r.declared_type).to eq :d
     end
+
+    # This tests some somewhat confusing behavior that used to occur due to the resource_name call
+    # automatically wiring up the old canonical provides line.
+    it "setting resoure_name does not override provides in prior resource" do
+      c1 = Class.new(Chef::Resource) do
+        resource_name :self_resource_name_test_4
+        provides :self_resource_name_test_4
+      end
+      c2 = Class.new(Chef::Resource) do
+        resource_name :self_resource_name_test_4
+        provides(:self_resource_name_test_4) { false } # simulates any filter that does not match
+      end
+      expect(Chef::Resource.resource_for_node(:self_resource_name_test_4, node)).to eql(c1)
+    end
   end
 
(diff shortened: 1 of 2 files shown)


erikng:
Would the rescue still mask the issue?

lamont-granquist:
no, that's just to handle the case of where you get something like `nil.to_sym` being called and to replace that with nil.

spec/unit/resource_spec.rb
```diff
@@ -492,6 +492,20 @@ def self.name
       expect(r.resource_name).to eq :blah
       expect(r.declared_type).to eq :d
     end
+
+    # This tests some somewhat confusing behavior that used to occur due to the resource_name call
```
jaymzh, line 496: I had to read this test... a lot... to understand what you were trying to test. More comments here on what the test is testing would be useful.

tas50:
I've renamed this to be 16.2 since that's when we ended up shipping it. It missed 16.1 by an hour or so.

PR #9894, opened 2020-05-20
Set up CI with Azure Pipelines

btm:
This PR adds testing of the end_to_end cookbook against linux, windows, and macos for all PRs.


azure-pipelines.yml
```diff
@@ -0,0 +1,88 @@
...
+
+variables:
+  FORCE_FFI_YAJL: 'ext'
+  CHEF_LICENSE: 'accept-no-persist'
+
+trigger:
+- master
+# TODO: 20190528 - should we be testing end_to_end on chef-15 and chef-14?
+#- chef-15
+#- chef-14
```
tas50, line 11: def not 14

kitchen-tests/cookbooks/end_to_end/recipes/default.rb
```diff
@@ -5,114 +5,6 @@
...
-include_recipe "::_apt" if platform_family?("debian")
-include_recipe "::_chef-vault" unless includes_recipe?("end_to_end::chef-vault")
-include_recipe "::_sudo"
-include_recipe "::_sysctl"
-include_recipe "::_alternatives"
-include_recipe "::_cron"
-include_recipe "::_ohai_hint"
-include_recipe "::_openssl"
-include_recipe "::_tests"
+include_recipe "::linux" if platform_family?("rhel", "debian")
```
tas50, line 8: We should just linux? this

PR #9897, opened 2020-05-21
macos_userdefaults: Fix ruby 2.7 keyword args warning and refactor a bit

### LINKED ISSUE #9891: macos_userdefaults shell_out syntax deprecation warning
## Description
The macos_userdefaults produces a deprecation warning on its invocation of `shell_out`

## Chef Version
Reported from 16.1.x

## Client Output
```
* macos_userdefaults[...] action write
/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.1.0/lib/chef/resource/macos_userdefaults.rb:81:
warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

phiggins:
Fixes #9891

### DIFF
--- a/lib/chef/resource/macos_userdefaults.rb
+++ b/lib/chef/resource/macos_userdefaults.rb
@@ -79,10 +79,10 @@ class Chef
         cmd << " | grep -qx '#{value}'"
 
         vc = if desired.user.nil?
-          shell_out(cmd)
-        else
-          shell_out(cmd, user: desired.user)
-        end
+               shell_out(cmd)
+             else
+               shell_out(cmd, user: desired.user)
+             end
 
         is_set !vc.error?
       end
--- a/lib/chef/resource/macos_userdefaults.rb
+++ b/lib/chef/resource/macos_userdefaults.rb
@@ -74,12 +74,17 @@ class Chef
 
       load_current_value do |desired|
         value = coerce_booleans(desired.value)
-        drcmd = "defaults read '#{desired.domain}' "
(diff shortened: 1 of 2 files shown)


lib/chef/resource/macos_userdefaults.rb
```diff
@@ -74,12 +74,17 @@ def coerce_booleans(val)
...
+        cmd << "'#{desired.key}' " if desired.key
+        cmd << " | grep -qx '#{value}'"
+
+        vc = if desired.user.nil?
+          shell_out(cmd)
+        else
+          shell_out(cmd, user: desired.user)
+        end
+
+        is_set vc.exitstatus == 0
```
tas50, line 87: You should be able to just use .error? here

phiggins: Good call!

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -74,12 +74,17 @@ def coerce_booleans(val)
...
+        cmd << "'#{desired.key}' " if desired.key
+        cmd << " | grep -qx '#{value}'"
+
+        vc = if desired.user.nil?
+          shell_out(cmd)
+        else
+          shell_out(cmd, user: desired.user)
+        end
+
+        is_set !vc.error?
```
tas50, line 87: This is definitely an improvement. Long term we need to kill this is_set entirely.

tas50:
@phiggins You have a minor chefstyle warning on this one

Issue #9918, opened 2020-05-27
chef_client_scheduled_task frequency_modifier default is problematic for daily, weekly, or even hourly schedules


chef-davin:
### Describe the Enhancement:

The new `chef_client_scheduled_task` resource is awesome.  The only issue I've found is with the default frequency modifier.  Having a modifier of 30 is great when you're using the default frequency of minutes, but it's awful when you try to use daily, weekly, or monthly, or hourly.  Ideally you should only need to use this parameter when you need an odd schedule (like bi-weekly, every other day, etc...).  Can this be adjusted to only be 30 when the default "minutes" frequency is used, and left at 1 otherwise?

### Describe the Need:

After setting up a scheduled task with the following code and running chef-client, you'll see that the next chef-client run will be 3am in a month.  The expected result of this resource block would be 3am the next day.

```ruby
chef_client_scheduled_task 'Run chef-client in audit mode' do
  start_time '03:00'
  frequency 'daily'
  splay 600
  daemon_options ['-n audit_only']
end
```

### Current Alternative

Obviously, you can adjust the frequency_modifier parameter, but I think this resource would be much more intuitive if we just put a little logic in the resource.  Something like
`if new_resource.frequency != minutes` that would then differentiate the default for minutes (30) and everything else (1)

```ruby
chef_client_scheduled_task 'Run chef-client in audit mode' do
  start_time '03:00'
  frequency 'daily'
  frequency_modifier 1
  splay 600
  daemon_options ['-n audit_only']
end
```

### Can We Help You Implement This?:

I'll try to create a pull request for this. I don't believe I'll need help with it, but will reach out if I do.

phiggins:
Fixed in #9920.

PR #9920, opened 2020-05-27
Update to the chef_client_scheduled_task resource frequency_modify default functionality

chef-davin:

Updates resource so that the frequency_modifier only defaults to 30 if the frequency is "minutes" and will otherwise default to 1

## Description

This update adds some logic into the pull request so that the frequency_modifier will be set to 1 if the frequency is set to hours, days, weeks, or months.  So that you don't end up scheduling a task to run once every 30 months by accident.

This could probably be more elegantly done, so if someone has improvements to make, I'd be very happy to get the input.

## Related Issue

#9918 

## Types of changes

## Checklist:


lib/chef/resource/chef_client_scheduled_task.rb
```diff
@@ -129,13 +139,22 @@ class ChefClientScheduledTask < Chef::Resource
...
         # According to https://docs.microsoft.com/en-us/windows/desktop/taskschd/schtasks,
         # the :once, :onstart, :onlogon, and :onidle schedules don't accept schedule modifiers
+
+        frequency_mod = if new_resource.frequency == "minute" && new_resource.frequency_modifier == 30
+                          30
+                        elsif new_resource.frequency_modifier != 30
+                          new_resource.frequency_modifier
+                        else
+                          1
+                        end
```
phiggins, line 149: This should probably be added as the default on the property, then tests could be written for all the different combos of inputs.

chef-davin: just so I understand what you mean, you're saying the default value should be 1, and then have tests for the frequency being "minute" that would modify this to 30? rather than the other way around?  I think "minute" is the only frequency option that would be bad to end up with a default frequency modifier of 1.

tas50: This logic should be used to determine the default and then we should have 

```ruby

Comment (phiggins) [reply lib/chef/resource/chef_client_scheduled_task.rb:149]: Sorry to be so vague. I meant this logic could be moved into the default of the `frequency_modifier` property. Here's an example of a property with a dynamic default that depends on the value of another parameter: https://github.com/chef/chef/blob/master/lib/chef/resource/execute.rb#L630-L632

With that changed, it'd be easier to test that the right values were being set depending on the input values.

Also, with a dynamic default a `default_description` should be added that describes the logic in English so that it'll show up in the autogenerated docs.

Comment (chef-davin) [reply lib/chef/resource/chef_client_scheduled_task.rb:149]: oh slick! I was hoping something like that existed for the default.  I'll try to get an update to this today to make the logic better.

Comment (chef-davin) [reply lib/chef/resource/chef_client_scheduled_task.rb]: New, better code:

```ruby
      property :frequency_modifier, [Integer, String],
        coerce: proc { |x| Integer(x) },
        callbacks: { "should be a positive number" => proc { |v| v > 0 } },
        description: "Numeric value to go with the scheduled task frequency",
        default: lazy { frequency == "minute" ? 30 : 1 }, 
        default_description: "30 if frequency is 'minute', 1 otherwise"
```

I believe this resolves this.  Thanks for the help @phiggins and @tas50

chef-davin:
Okay, so I fixed all the failing Windows unit tests for this resource, and I also put in unit tests that would validate that the frequency_modifier defaults to 1 unless the frequency is "minute", in which case it defaults to 30. So I feel like this is pretty solid.  Going to take this off of [WIP] and hopefully it can get pulled into the next update.

spec/unit/resource/chef_client_scheduled_task_spec.rb
```diff
@@ -70,12 +80,12 @@
...
   describe "#client_cmd" do
     it "creates a valid command if using all default properties" do
-      expect(provider.client_cmd).to eql("C:/opscode/chef/bin/chef-client -L /etc/chef/log/client.log -c /etc/chef/client.rb")
+      expect(provider.client_cmd).to eql("C:/opscode/chef/bin/chef-client -L /etc/chef/log/client.log -c /etc/chef/client.rb") | eql("C:/opscode/chef/bin/chef-client -L C:\\chef/log/client.log -c C:\\chef/client.rb")
     end
 
     it "uses daemon_options if set" do
       resource.daemon_options ["--foo 1", "--bar 2"]
-      expect(provider.client_cmd).to eql("C:/opscode/chef/bin/chef-client -L /etc/chef/log/client.log -c /etc/chef/client.rb --foo 1 --bar 2")
+      expect(provider.client_cmd).to eql("C:/opscode/chef/bin/chef-client -L /etc/chef/log/client.log -c /etc/chef/client.rb --foo 1 --bar 2") | eql("C:/opscode/chef/bin/chef-client -L C:\\chef/log/client.log -c C:\\chef/client.rb --foo 1 --bar 2")
```
tas50, line 88: Long term we need a better solution for cross platform specs, but this will do for now.

chef-davin: @tas50 could we do some regex with it instead?  
(I don't know if this is correct ruby regex syntax, I'm probably conflating with *nix syntax)

```ruby
expect(provider.client_cmd).to match(/opscode\/chef\/bin\/chef-client -L .*chef\/log\/client.log -c .*chef\/client.rb --foo 1 --bar 2/)
```
It's not exactly pretty, but it's still feels better than the or statement (at least to me).

tas50: Historically we've had a lot of really subtle issues building these commands which makes me nervious to use a regex. I think we may just have to per platform set expects here so it's more obvious what we're trying to achieve.

PR #9925, opened 2020-05-28
Update to ssl_verify_mode on remote_file

jaymzh:
So it turns out the `berkshelf` has been passing it's entire `options`
hash into `Chef::HTTP` this whole time, which, we should probably fix.

However, one of those options was `:ssl_verify_mode`. We ignore entries
in the `options` hash we don't recognize so this was fine, but suddenly
we recognize it. But we expected it to be class, and they're passing
in symbols. Whoops!

This makes that all work and puts the symbol->class conversion inside
of `Chef::HTTP`.

### DIFF
--- a/lib/chef/provider/remote_file/http.rb
+++ b/lib/chef/provider/remote_file/http.rb
@@ -135,12 +135,7 @@ class Chef
             opts[:disable_gzip] = true
           end
           if new_resource.ssl_verify_mode
-            opts[:ssl_verify_mode] = case new_resource.ssl_verify_mode
-                                     when :verify_none
-                                       Chef::HTTP::VerifyNoneSSLPolicy
-                                     when :verify_peer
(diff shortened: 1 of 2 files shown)


lib/chef/provider/remote_file/http.rb
```diff
@@ -135,12 +135,7 @@ def http_client_opts
             opts[:disable_gzip] = true
           end
           if new_resource.ssl_verify_mode
-            opts[:ssl_verify_mode] = case new_resource.ssl_verify_mode
-                                     when :verify_none
-                                       Chef::HTTP::VerifyNoneSSLPolicy
-                                     when :verify_peer
-                                       Chef::HTTP::VerifyPeerSSLPolicy
-                                     end
+            opts[:ssl_verify_mode] = new_resource.ssl_verify_mode
```
tas50, line 138: this should just get inlined now

PR #9928, opened 2020-05-28
hostname: Remove support for Solaris 5.10

tas50:
We don't support Solaris 5.10 in Chef anymore so there's no reason to make this check every time the resource runs on Solaris 5.11


tas50:
I fixed the failures in #9931

PR #9931, opened 2020-05-29
Make sure file is properly scoped in cron_access

tas50:
This was hitting the file resource not the file ruby class.


tas50:
Yeah we need to wire up chefstyle with some real rules

Issue #9940, opened 2020-06-01
dnf package resource version comparison failures.


epilatow:
While using a recent snapshot of the chef dnf package provider ([3494a29604b7719ab640a325becabbcdc57a7bce](https://github.com/chef/chef/pull/9500)) I saw chef failing to upgrade rpm packages when it should have. The failures occurred on Centos 8 with Chef 13.10 (after I monkeypatched in the aforementioned new dnf package provider into it).

Looking at one specific failure scenario, the following package was installed on a machine: grub2-efi-x64-2.02-78.el8.x86_64.rpm. In chef there was a "grub2-efi-x64" package resource that specified the following version: 2.02-78.el8_1.1

But when chef ran it did not upgrade the installed package. Doing a debug chef run I tracked the problem down to versioncompare in dnf_helper.py. The problem can be reproduced with the following dnf_helper.py invocation (which mirrors what chef was doing):

    echo '{"action": "versioncompare", "versions": ["1:2.02-78.el8.x86_64","1:2.02-78.el8_1.1"]}' | /usr/libexec/platform-python ~/dnf_helper.py 2>&1 | head -1

If you run this command (as root) you'll get an output of 0, saying that these two version strings are equal, when they are clearly not. This failure occurs because dnf_helper.py has been asked to do the impossible. It's been asked to parse and compare EVRs and EVRAs without knowing which it has been given. (As a simple example, imagine being given the following string: 1:2.02-78.el8. Is this an EVR or an EVRA? Now assume the existence of an el8 architecture and attempt to answer the question again.)

For dnf_helper.py versioncompare to work successfully it must know the type of version input strings (EVR or EVRA) that it is dealing with.

It's worth noting that this problem doesn't occur if, instead of using the package resource version property, we instead include a NEVR in the package resource package_name property. This is because, in this scenario chef doesn't detect the existing installed package, so it never invokes versioncompare to do version comparisons.

lamont-granquist:
closing as dup of #9763 and #9847

PR #9941, opened 2020-06-01
WIP: Provider::Package to use EVRAs for dnf rpm version comparisons

epilatow:
This is a potential bug fix for [Issue 9940](https://github.com/chef/chef/issues/9940).

Currently on dnf systems, Provider::Package does rpm version comparisons
between installed_version/candidate_version (EVRAs) and
new_resource.version / new_version_array (a VR or EVR). These
comparisons can fail in dnf_helper.py because versioncompare doesn't
know the type of version data it's comparing (and it's impossible to
deterministically parse an EVR[A] string without first knowing if it's
an EVR or EVRA).

To fix this problem, we first override new_version_array in
Chef::Provider::Package to just return nil values. This forces the
Package class to use candidate_version in place of new_version_array for
all version comparisons.

This doesn't fix rpm version comparisons in and of itself, but instead
it ensures that all version comparisons are done using EVRAs. This
allows us to update dnf_python.py versioncompare to properly parse and
compare EVRAs. This fixes rpm version comparisons for package resources
which contain a version specification and an :upgrade action.

To fix rpm version comparisons for package resources that have a version
specification and an :install action, we also need to update
target_version_array in Chef::Provider::Package to fallback to comparing
against candidate_version when new_version_array returns nil.

To test this fix I created four packages, each with two versions:
2.02-78.el8 and 2.02-78.el8_1.1.  I installed the 2.02-78.el8 version of
all the packages, and then executed a chef run with the following
resources configured:

---8<---
package 'empty-test-rpm-version-upgrade' do
  version '1:2.02-78.el8_1.1'
  action :upgrade
end

package 'empty-test-rpm-version-install' do
  version '1:2.02-78.el8_1.1'
  action :install
end

package 'empty-test-rpm-name-upgrade-1:2.02-78.el8_1.1' do
  action :upgrade
end

package 'empty-test-rpm-name-install-1:2.02-78.el8_1.1' do
  action :install
end
---8<---

and then I confirmed that all the packages were upgraded as expected.

I also regression tested that if an invalid new_version is requested
(ie, a version for which we can't find a candidate_version) that chef
fails (as expected).

Signed-off-by: Edward Pilatowicz <epilatow@fb.com>

### DIFF
--- a/lib/chef/provider/package/dnf.rb
+++ b/lib/chef/provider/package/dnf.rb
@@ -81,6 +81,31 @@ class Chef
           super
         end
 
+        #
+        # Override new_version_array in Chef::Provider::Package to to
+        # force package version comparisons to use the
+        # candidate_version values instead of new_resource.version.
+        #
+        # We do this because new_resource.version can specify a partial
+        # version spec, where as candidate_version has been expanded to
+        # a full EVRA, which can be accurately compared against
+        # installed_version later (also an EVRA, if it exists).
+        #
+        # If the there is no candidate_version we have to leave
+        # new_resource.version as is to preserve the fact that the user
+        # requested an unavailable version.
+        #
+
+        # @return [Array] new_version(s) as an array
+        def new_version_array
+          @new_version_array ||= \
+            [ new_resource.version ].flatten.each_with_index.map do |nv, i|
+              nv = nv.to_s.empty? ? nil : nv
+              cv = candidate_version[i]
+              nv and cv ? cv : nv
+            end
+        end
+
         def candidate_version
           package_name_array.each_with_index.map do |pkg, i|
             available_version(i).version_with_arch
--- a/lib/chef/provider/package/dnf/dnf_helper.py
+++ b/lib/chef/provider/package/dnf/dnf_helper.py
@@ -52,37 +52,27 @@ def flushcache():
         pass
     get_sack().load_system_repo(build_cache=True)
 
-def version_tuple(versionstr):
-    e = '0'
-    v = None
-    r = None
-    colon_index = versionstr.find(':')
-    if colon_index > 0:
-        e = str(versionstr[:colon_index])
-    dash_index = versionstr.find('-')
-    if dash_index > 0:
-        tmp = versionstr[colon_index + 1:dash_index]
-        if tmp != '':
-            v = tmp
-        arch_index = versionstr.find('.', dash_index)
-        if arch_index > 0:
-            r = versionstr[dash_index + 1:arch_index]
-        else:
-            r = versionstr[dash_index + 1:]
-    else:
-        tmp = versionstr[colon_index + 1:]
-        if tmp != '':
-            v = tmp
-    return (e, v, r)
+def evras_split(evras):
+    evrs = []
+    archs = []
+    for evra in evras:
+        tmp = evra.split('.')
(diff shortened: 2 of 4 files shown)


lib/chef/provider/package.rb
```diff
@@ -452,9 +452,15 @@ def target_version_array
...
                   logger.trace("#{new_resource} #{package_name} not installed, installing #{candidate_version}")
                   target_version_array.push(candidate_version)
-                else
+                elsif version_compare(current_version, candidate_version) == 0
                   logger.trace("#{new_resource} #{package_name} #{current_version} already installed")
                   target_version_array.push(nil)
+                elsif !allow_downgrade && version_compare(current_version, candidate_version) == 1
+                  logger.trace("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{candidate_version}. Skipping...)")
+                  target_version_array.push(nil)
+                else
```
jaymzh, line 461: `:install` doesn't upgrade like this. Unless the version is specified, `:install` simply installs if nothing is installed. You need to use `:upgrade` to upgrade-to-latest or `:install` with a version to move to a new version.

epilatow: Thanks  for looking at this.  Also, thanks for the description of the required behavior (I'm new to chef and didn't know this).

I added this code to fix the following test case:
---8<---
package 'empty-test-rpm-version-install' do
version '1:2.02-78.el8_1.1'
action :install
end
---8<---

With this change, new_version (which contained the user specified version) has been nil'd out (by the dnf subclass).  So I no longer have the context necessary to make the determination if we should do the upgrade in the scenario you've outlined.  :(

I'm hesitant to update package.rb to add special handling for this case since it's a generic resource handler and this information is only needed for this one dnf edge case.

My first idea for a fix would probably be to update the dnf version of new_version_array() such that: instead of always nil'ing out the entire array, if we're dealing with an install package resource, then we could set new_version to the available_version if both are non-nil (and set it to nil otherwise).

I'd be curious to get @lamont-granquist perspective on this idea (and the rest of the diff).  I noticed that he just closed the issue this diff is addressing as a dup of another issue+PR, so I suspect this PR may be obsoleted shortly.

epilatow: I've updated the PR with the changes I described above, which allowed me to revert all the updates to provider/package.rb (so I've restored the old install behavior).  I've verified that all my original tests (from the comments) pass.

epilatow:
We've been running with these dnf changes for since June, but last week we discovered a bug in them. Specifically, by nil'ing out new_versions when there is no candidate_version, chef would succeed if a package resource specified an invalid version. (Ie, when a requested version couldn't be mapped to a candidate version.) I've just updated this PR to address the issue. This required leaving new_versions alone if there was no candidate version, which complicated things because we can't use unresolved new_versions strings for comparisons against candidate versions (ie full EVRAs). So I added tagging for EVRAs so we can ensure that we don't accidentally compare them to non-EVRAs.

lamont-granquist:
> Specifically, by nil'ing out new_versions when there is no candidate_version, chef would succeed if a package resource specified an invalid version. 

yeah, i think that's why this PR stalled so hard because I knew that had to be wrong...

epilatow:
Obsoleted by PR #9847

PR #9944, opened 2020-06-02
Fix snap_package bugs

### LINKED ISSUE #9041: service provider doesn't handle `indirect` systemd units

jaymzh:
First and foremost this fixes the bug in which #8827 where
`snap_package` hangs forever.

See the comment in the code for the nitty-gritty details, but snap never
returns an EOF so you have to jump through GREAT hoops not to hang
forever.

UNLESS you want to use nonblock+read, but it takes MUCH longer:

```
[phil@ldt-hardwired ~]$ time /tmp/test.rb char

real    0m0.252s
user    0m0.207s
sys     0m0.044s
[phil@ldt-hardwired ~]$ time /tmp/test.rb select

real    0m5.244s
user    0m0.183s
sys     0m0.056s
```

That's just for doing `GET /v2/snaps/black`

Second, if you try to install a snap that doesn't exist on the `stable`
channel then Chef crashes with a `undefined method [] on nil`, so let's
check for that explicitly.

And a bunch of other fixes that were required to even install anything:

* prevent get_current_versions from returning `[nil]`
* don't try to 'update' when something is not yet installed
* don't try to install versions, install packages
* don't try to push ruby hashes into HTTP without JSONing them first
* when installing, install one at a time, it's the only API that accepts a channel
* options is an array, not a hash

NOTE... there's a lot of janky here. A lot. It's the minimalist stuff
from #9106 I could pull in, while also making the socket stuff not
painfully slow. That PR aims to support versions and lots of other and
should be fixed up and [eventually] merged... but in the meantime this
provider doesn't even come close to working and this at least gets
packages installing.

After this, I never want to touch snap again.

Aha! Link: https://chef.aha.io/features/SH-114

### DIFF
--- a/lib/chef/provider/package/snap.rb
+++ b/lib/chef/provider/package/snap.rb
@@ -59,15 +59,14 @@ class Chef
         def get_current_versions
           package_name_array.each_with_index.map do |pkg, i|
             installed_version(i)
-          end
+          end.compact
         end
 
         def install_package(names, versions)
           if new_resource.source
             install_snap_from_source(names, new_resource.source)
           else
-            resolved_names = names.each_with_index.map { |name, i| available_version(i).to_s unless name.nil? }
-            install_snaps(resolved_names)
+            install_snaps(names)
           end
         end
 
@@ -75,14 +74,16 @@ class Chef
           if new_resource.source
             install_snap_from_source(names, new_resource.source)
           else
-            resolved_names = names.each_with_index.map { |name, i| available_version(i).to_s unless name.nil? }
-            update_snaps(resolved_names)
+            if get_current_versions.empty?
+              install_snaps(names, versions)
+            else
+              update_snaps(names)
+            end
           end
         end
 
         def remove_package(names, versions)
-          resolved_names = names.each_with_index.map { |name, i| installed_version(i).to_s unless name.nil? }
-          uninstall_snaps(resolved_names)
+          uninstall_snaps(names)
         end
 
         alias purge_package remove_package
@@ -129,19 +130,72 @@ class Chef
             "Accept: application/json\r\n" +
             "Content-Type: application/json\r\n"
           if method == "POST"
-            request.concat("Content-Length: #{post_data.bytesize}\r\n\r\n#{post_data}")
+            pdata = post_data.to_json.to_s
+            request.concat("Content-Length: #{pdata.bytesize}\r\n\r\n#{pdata}")
           end
           request.concat("\r\n")
-          # While it is expected to allow clients to connect using HTTPS over a TCP socket,
-          # at this point only a UNIX socket is supported. The socket is /run/snapd.socket
-          # Note - UNIXSocket is not defined on windows systems
+
+          # while it is expected to allow clients to connect using https over
+          # a tcp socket, at this point only a unix socket is supported. the
+          # socket is /run/snapd.socket note - unixsocket is not defined on
+          # windows systems
           if defined?(::UNIXSocket)
-            UNIXSocket.open("/run/snapd.socket") do |socket|
-              # Send request, read the response, split the response and parse the body
-              socket.print(request)
-              response = socket.read
-              headers, body = response.split("\r\n\r\n", 2)
-              JSON.parse(body)
+            UNIXSocket.open('/run/snapd.socket') do |socket|
+              # send request, read the response, split the response and parse
+              # the body
+              socket.write(request)
+
+              # WARNING!!! HERE BE DRAGONs
+              #
+              # So snapd doesn't return an EOF at the end of its body, so
+              # doing a normal read will just hang forever.
+              #
+              # Well, sort of. if, after it writes everything, you then send
+              # yet-another newline, it'll then send its EOF and promptly
+              # disconnect closing the pipe and preventing reading. so, you
+              # have to read first, and therein lies the EOF problem.
+              #
+              # So you can do non-blocking reads with selects, but it
+              # makes every read take about 5 seconds. If, instead, we
+              # read the last line char-by-char, it's about half a second.
+              #
+              # Reading a character at a time isn't efficient, and since we
+              # know that http headers always have a blank line after them,
+              # we can read lines until we find a blank line and *then* read
+              # a character at a time. snap returns all the json on a single
+              # line, so once you pass headers you must read a character a
+              # time.
+              #
+              # - jaymzh
+
+              Chef::Log.trace(
+                "snap_package[#{new_resource.package_name}]: reading headers",
+              )
+              loop do
+                response = socket.readline
+                break if response.strip.empty? # finished headers
+              end
+              Chef::Log.trace(
+                "snap_package[#{new_resource.package_name}]: past headers, " +
+                'onto the body...',
+              )
+              result = nil
+              body = ''
+              socket.each_char do |c|
+                body << c
+                # we know we're not done if we don't have a char that
+                # can end JSON
+                next unless ['}', ']'].include?(c)
+                begin
+                  result = JSON.parse(body)
+                  # if we get here, we were able to parse the json so we
+                  # are done reading
+                  break
+                rescue JSON::ParserError
+                  next
+                end
+              end
+              result
             end
           end
         end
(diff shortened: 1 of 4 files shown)


lib/chef/provider/package/snap.rb
```diff
@@ -311,6 +349,9 @@ def get_latest_package_version(name, channel)
             raise Chef::Exceptions::Package, json["result"], caller
           end
 
+          unless json["result"][0]["channels"]["latest/#{channel}"]
+            fail "No version of #{name} in channel #{channel}"
```
phiggins, line 353: Since `fail` is a synonym of `raise`, should this use a more specific exception class like the `raise` a few lines up?

lib/chef/provider/package/snap.rb
```diff
@@ -138,10 +138,48 @@ def call_snap_api(method, uri, post_data = nil?)
...
+                begin
+                  result = JSON.parse(body)
+                  # if we get here, we were able to parse the json so we
+                  # are done reading
+                  break
+                rescue JSON::ParserError
+                  next
+                end
+              end
+              result
```
phiggins, line 182: If `read_nonblock` and friends don't work or are undesirable here, since a valid json document will end with a `]` or a `}` (right?), this could read until one of those characters is found before attempting to parse json.

lib/chef/provider/package/snap.rb
jaymzh: yes it should. Fixed.

lib/chef/provider/package/snap.rb
jaymzh: Nice optimization. Added.
Also note that in my new description I found you *can* make `read_nonblock` work, but it's 10x slower than this. ANd that was before your optimization.

lib/chef/provider/package/snap.rb
```diff
@@ -59,30 +59,31 @@ def candidate_version
...
             installed_version(i)
-          end
+          end.compact
         end
 
         def install_package(names, versions)
           if new_resource.source
             install_snap_from_source(names, new_resource.source)
           else
-            resolved_names = names.each_with_index.map { |name, i| available_version(i).to_s unless name.nil? }
```
jaymzh, line 69: this returns versions, not names.

lib/chef/provider/package/snap.rb
```diff
@@ -59,30 +59,31 @@ def candidate_version
...
-            install_snaps(resolved_names)
+            install_snaps(names)
           end
         end
 
         def upgrade_package(names, versions)
           if new_resource.source
             install_snap_from_source(names, new_resource.source)
           else
-            resolved_names = names.each_with_index.map { |name, i| available_version(i).to_s unless name.nil? }
```
jaymzh, line 78: same

lib/chef/provider/package/snap.rb
```diff
@@ -59,30 +59,31 @@ def candidate_version
...
+            if get_current_versions.empty?
+              install_snaps(names, versions)
+            else
+              update_snaps(names)
+            end
           end
         end
 
         def remove_package(names, versions)
-          resolved_names = names.each_with_index.map { |name, i| installed_version(i).to_s unless name.nil? }
```
jaymzh, line 84: same

lib/chef/provider/package/snap.rb
```diff
@@ -129,19 +130,72 @@ def call_snap_api(method, uri, post_data = nil?)
             "Accept: application/json\r\n" +
             "Content-Type: application/json\r\n"
           if method == "POST"
-            request.concat("Content-Length: #{post_data.bytesize}\r\n\r\n#{post_data}")
```
jaymzh, line 132: clearly no one ever tried this...

lib/chef/provider/package/snap.rb
```diff
@@ -129,19 +130,72 @@ def call_snap_api(method, uri, post_data = nil?)
...
+              #
+              # So snapd doesn't return an EOF at the end of its body, so
+              # doing a normal read will just hang forever.
+              #
+              # Well, sort of. if, after it writes everything, you then send
+              # yet-another newline, it'll then send its EOF and promptly
+              # disconnect closing the pipe and preventing reading. so, you
+              # have to read first, and therein lies the EOF problem.
+              #
+              # So you can do non-blocking reads with selects, but it
```
jaymzh, line 158: updated this with speed comment.

lib/chef/provider/package/snap.rb
```diff
@@ -211,20 +265,22 @@ def install_snap_from_source(name, path)
...
+          snap_names.each do |snap|
+            response = post_snap(snap, "install", new_resource.channel, new_resource.options)
+            id = get_id_from_async_response(response)
+            wait_for_completion(id)
+          end
         end
 
         def update_snaps(snap_names)
-          response = post_snaps(snap_names, "refresh", new_resource.channel, new_resource.options)
+          response = post_snaps(snap_names, "refresh", nil, new_resource.options)
```
jaymzh, line 277: afaict, updates generally happen within a channel, so I left this as batch, but that doesn't work with installs.

but when you do multiples you cannot specify a channel

lib/chef/provider/package/snap.rb
```diff
@@ -278,18 +334,20 @@ def generate_snap_json(snap_names, action, channel, options, revision = nil)
...
           end
 
           # No defensive handling of params
           # Snap will throw the proper exception if called improperly
           # And we can provide that exception to the end user
-          request["classic"] = true if options["classic"]
-          request["devmode"] = true if options["devmode"]
-          request["jailmode"] = true if options["jailmode"]
+          if options
+            request["classic"] = true if options.include?("classic")
```
jaymzh, line 345: lolz, not a hash.

tas50:
Before we can merge this or #9106 we need to add some kitchen tests or snap_package exercising as many scenarios as we can reasonably do in a quick kitchen run. It's something we can't really test in unit tests and we need to have a reasonable level of confidence when merging future changes.

jaymzh:
@tas50 - yup the few unittests that are there are basically useless. Yes, we should have a kitchen tests and I can add one. **however** this PR isn't meant to last. This whole thing needs a ground up re-write, and frankly, unless either the author of #9106 or lamont are going to actually do that, we should probably just *remove* this provider altogether. It doesn't work. I'm not even sure this will work reliably. Among other things it drops versions on the floor.... nothing here works as advertised, this just makes it not crash and at least attempt to install something.

At this point I'd almost rather just do:
```
execute 'snap install black' do
  not_if 'snap list | grep black'
end
```

for the 3 snaps I have to install. I guess what I"m saying is I can add a kitchen test that calls install (that's the ONLY thing that works at this point), or more, but I'm in no way going to take responsibility for making those tests work (only the simplest install case will work at this moment).

I actually pulled this monkey patch out of our cookbook at work and put the install into a shell script because I don't trust this provider at all, but I figured the least I could do is get *something* out so people who are following the attached bug might have some way forward.

jaymzh:
I attempted to add kitchen tests, @tas50 , but it turns out that snap has some issues running in containers:
```
    status: Internal Server Error, kind: , message: system does not fully support snapd: cannot mount squashfs image using "squashfs": mount: /tmp/sanity-mountpoint-649535134: failed to setup loop device for /tmp/sanity-squashfs-413258619.
```

I know that loopback mounts _can_ work in containers, but no idea what needs to be done to make it so.

jaymzh:
huh I take it back, looks like it worked on Azure. Won't look a gift horse in the mouth. Woot.

jaymzh:
@tas50 to aisle 5

tas50:
@jaymzh Can you rebase it and fix the failing spec?

jaymzh:
yeah if you're happy with the kitchentest

jaymzh:
rebased, fixed specs.

tas50:
I'll nuke the kitchen tests out later since it turns out that doesn't work in docker atm.

Issue #9610, opened 2020-04-06
Powershell_Package does not throw error when it cannot connect to powershell gallery


Xorima:
## Description
When using `powershell_package` chef marks the resource as (up to date) when it is not, if the client does not support tls 1.2 

## Platform Version
(2012, 2012R2 and 2016 do not by default)

## Replication Case
Try to install a powershell package with tls 1.2 disabled on the machine (IE a default windows vm) it will mark as up to date when an error is actually thrown

## Possible Fixes

To resolve this internally there are two options, one is invasive, one is not:

Putting this above the install-module line will allow it to run tls 1.2:
`[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12`

Alternatively settings the following registry keys will make powershell run tls 1.2 by default (It does turn off other ciphers such as sslv3 though):

```
registry_key 'HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\.NETFramework\\v4.0.30319' do
  values [{
    name: 'SchUseStrongCrypto',
    type: :dword,
    data: 1,
  }]
end
registry_key 'HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\.NETFramework\\v4.0.30319' do
  values [{
    name: 'SchUseStrongCrypto',
    type: :dword,
    data: 1,
  }]
end
```

tas50:
I know that releng worked around issues with chocolatey and TLS 1.2 by forcing powershell to use 1.2. We'll want to use that path forward since we can't modify TLS settings host wide with the registry key.

lamont-granquist:
yeah, we can't fiddle with the TLS settings in the registry key.  the other solution would be acceptable.

Xorima:
I am fully happy with the `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12` method, I guess where should I be looking at when implementing this as it is getting really really needed.

PR #9946, opened 2020-06-02
Fixed Powershell_Package does not throw error when it cannot connect …

sanga1794:
Signed-off-by: sanga17 <sausekar@msystechnologies.com>

## Description
Powershell_Package does not throw error when it cannot connect to powershell gallery

## Related Issue
Fixed: https://github.com/chef/chef/issues/9610

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/powershell.rb
+++ b/lib/chef/provider/package/powershell.rb
@@ -53,6 +53,9 @@ class Chef
 
         # Installs the package specified with the version passed else latest version will be installed
         def install_package(names, versions)
+          # To enable tls 1.2, which is disabled by default in some OS
+          powershell_out("[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12")
+
           names.each_with_index do |name, index|
             cmd = powershell_out(build_powershell_package_command("Install-Package '#{name}'", versions[index]), timeout: new_resource.timeout)
             next if cmd.nil?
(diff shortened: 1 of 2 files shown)


tas50:
Paging @Xorima

lib/chef/provider/package/powershell.rb
```diff
@@ -115,7 +118,8 @@ def build_powershell_package_command(command, version = nil)
           command = [command] unless command.is_a?(Array)
           cmdlet_name = command.first
           command.unshift("(")
-          %w{-Force -ForceBootstrap}.each do |arg|
+          # -WarningAction SilentlyContinue is used to suppress the warnings from stdout
```
Xorima, line 121: I guess I have a question as to why you would want to suppress warnings from stdout? 

The other thing is you might want `-erroraction stop` which will blow up powershell if the command errors (or at least throw the error so it can actually be caught ...)

sanga1794: Hi @Xorima , I  have used WarningAction because while storing the version from stdout, it also stores with the warnings e.g.
`WARNING: MSG:UnableToDownload �https://go.microsoft.com/fwlink/?LinkID=627338&clcid=0x409� ��
WARNING: Unable to download the list of available providers. Check your internet connection.
1.8.6`
This whole value is stored in the candidate_version as string.

lib/chef/provider/package/powershell.rb
```diff
@@ -115,7 +118,8 @@ def build_powershell_package_command(command, version = nil)
...
+          %w{-Force -ForceBootstrap -WarningAction SilentlyContinue}.each do |arg|
```
Vasu1105, line 122: if we are adding -WarningAction SilentlyContinue here then why in test cases https://github.com/chef/chef/pull/9946/files#diff-4efcbd80710bb0976c44af3650e95528R127 we added that for each command. If I am not wrong and getting it correctly isn't this get appended to those command automatically ?

sanga1794: @Vasu1105, I have added the 'WarningAction SilentlyContinue' in specs, because we are compairing the commands that generating from the parrticular method with the commands used in specs.

PR #9949, opened 2020-06-03
RFC: fixing systemd_services

jaymzh:
In #9063 I thought I had fixed a bug in the systemd service provider,
to handle indirect services. Here I'm adding the same "fix" for static
services. But it turns out, I had not. You get fun stuff like:

```
[2020-06-02T16:56:33-07:00] TRACE: service[gdm] cannot be disabled: it is static
[2020-06-02T16:56:33-07:00] INFO: service[gdm] disabled
```

womp womp.

This is because the actual `action :disable` in `providers/service.rb`
(the superclass) is what decides:

```
      action :disable do
        if current_resource.enabled
          converge_by("disable service #{new_resource}") do
            disable_service
            logger.info("#{new_resource} disabled")
          end
        else
          logger.trace("#{new_resource} already disabled - nothing to do")
        end
        load_new_resource_state
        new_resource.enabled(false)
      end
```

There are a few ways around this:

1. We could define our own action :disable and not use the superclass.
2. We could instead push the idea of static/indirect up to the parent
class, it's already in the resource, and have them default to `false`
which would make it a no-op for other subclasses
3. ?? Something i haven't thougt of?

Also I changed this not to run `is-enabled` 4 times for every service
(sigh).

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -51,15 +51,18 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
         current_resource.running(false)
         current_resource.enabled(false)
         current_resource.masked(false)
+	current_resource.static(false)
         current_resource.indirect(false)
       end
     else
       current_resource.running(is_active?)
     end
 
-    current_resource.enabled(is_enabled?)
-    current_resource.masked(is_masked?)
-    current_resource.indirect(is_indirect?)
+    e, m, s, i = get_enabled_masked_static_indirect
+    current_resource.enabled(e)
+    current_resource.masked(m)
+    current_resource.static(s)
+    current_resource.indirect(i)
     current_resource
   end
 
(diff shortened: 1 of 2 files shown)


jaymzh:
@lamont-granquist / @tas50  thoughts? Option 2 is more work, but I think it's the better option and I'm willing to do the work.

lamont-granquist:
what is 'static' or 'indirect'?  i don't use systemd enough to know what those nouns actually mean.

lamont-granquist:
It seems like aliases should be resolved by the subclass and shouldn't take any changes to the superclass, other than perhaps a hook or some cleanup not to directly trust the name given in the new_resource.

It doesn't need to be much more complicated than this:

```
# default behavior in the superclass
def name
  new_resource.name
end

# in the subclass
def name
  lookup_alias(new_resource.name)
end
```

jaymzh:
well indirect can be an `Also=` or a bunch of other things. See the `systemd_unit` for the logic here, it handles all this correctly...

jaymzh:
OK so... # 2? we good with 2?

lamont-granquist:
Well a very, very lightweight version of #2.

Issue #9577, opened 2020-04-02
archive_file resource casts mode incorrectly


sfiggins:
## Description

When archive file creates the destination folder it incorrectly casts the mode for the directory. What should be by default `0755` becomes `01363`.

## Chef Version

Noted with  15.8.23. Looks like it is still a problem with current master.

## Platform Version

Linux - a centos 7 target. Though this would probably be a problem on any unix like filesystem.

## Replication Case

Use archive_file and give a destination that does not already exist. Created directory will by default be assigned mode 01363.

Problem line is in `lib/chef/resource/archive_file.rb`

```ruby
            FileUtils.mkdir_p(new_resource.destination, mode: new_resource.mode.to_i)
```

Casting '0755' to integer yields 755. When cast back to octal this becomes 01363.

lamont-granquist:
There's code to handle this already in the file provider's mode handling (buried somewhere in a mixin I think).

lamont-granquist:
lib/chef/file_access_control/unix.rb has a `mode_from_resource` method.

that should probably go into a helper in chef-utils so its accessable as "how Chef, Inc coerces string things that might be octals into numbers".

tas50:
@sfiggins I merged in a fix for this. Unfortunately Ruby doesn't expose a way to tell the base of the Integer the user passed in and mkdir_p needs it to be in Octal form. That means there's no reliable way for us to ensure the passed in value is appropriate. I've added a deprecation for passing in Integer values since we can convert String values to octal without issues.

PR #9950, opened 2020-06-03
archive_file: better handle mode property and deprecate Integer values

tas50:
FileUtils.mkdir_p expects the mode to be in octal form. We accept Integers, but Ruby doesn't have a way to tell if what the user passed was octal or not. The safest thing to do here is to only accept mode as a String and then convert it to octal form with .to_i(8). This adds a new deprecation and correctly sets the file mode value as octal. 

**Old behavior**

String mode: applies wrong permission
Base 10 mode: applies wrong permissions
Octal mode: Works

**New behavior**

String mode: Works
Base 10 mode: Works but throws a deprecation warning
Octal mode: applies the wrong permissions and throws a deprecation warning

### DIFF
--- a/lib/chef/resource/archive_file.rb
+++ b/lib/chef/resource/archive_file.rb
@@ -53,7 +53,7 @@ class Chef
         description: "The group of the extracted files."
 
       property :mode, [String, Integer],
-        description: "The mode of the extracted files. Integer values are deprecated as octal strings (ex. 0755) would not be interpreted correctly.",
+        description: "The mode of the extracted files. Integer values are deprecated as octal values (ex. 0755) would not be interpreted correctly.",
         default: "755"
 
       property :destination, String,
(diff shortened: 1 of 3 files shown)


lib/chef/resource/archive_file.rb
```diff
@@ -53,7 +53,7 @@ class ArchiveFile < Chef::Resource
         description: "The group of the extracted files."
 
       property :mode, [String, Integer],
-        description: "The mode of the extracted files.",
+        description: "The mode of the extracted files. Integer values are deprecated as octal strings (ex. 0755) would not be interpreted correctly.",
```
phiggins, line 56:
```suggestion
        description: "The mode of the extracted files. Integer values are deprecated as octal values (ex. 0755) would not be interpreted correctly.",
```

phiggins:
Fix for #9577

PR #9951, opened 2020-06-03
archive_file: move ffi-libarchive into a simple helper method

tas50:
This way it always gets called in the action, but we can still avoid
spec failures on systems that don't have ffi-libarchive

### DIFF
--- a/lib/chef/resource/archive_file.rb
+++ b/lib/chef/resource/archive_file.rb
@@ -19,6 +19,7 @@
 #
 
 require_relative "../resource"
+require "fileutils" unless defined?(FileUtils)
 
 class Chef
   class Resource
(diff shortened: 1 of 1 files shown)


lib/chef/resource/archive_file.rb
```diff
@@ -72,11 +85,11 @@ class ArchiveFile < Chef::Resource
       alias_method :extract_options, :options
       alias_method :extract_to, :destination
 
-      require "fileutils" unless defined?(FileUtils)
-
       action :extract do
         description "Extract and archive file."
 
+        require_libarchive
```
phiggins, line 91: Any particular reason this needs to be a method instead of a plain `require` here?

tas50: The specs were exploding when I just did a require here.

phiggins: Weird

PR #9952, opened 2020-06-04
knife vault on windows 10 fails due to ERROR: Chef::Exceptions::InvalidDataBagPath

snehaldwivedi:
Signed-off-by: Snehal Dwivedi <sdwivedi@msystechnologies.com>

## Description
- updated exception `Chef::Exceptions::InvalidDataBagPath` message

## Related Issue
Fixes:  https://github.com/chef/chef-vault/issues/333
https://github.com/chef/chef-vault/issues/321

### DIFF
--- a/lib/chef/data_bag.rb
+++ b/lib/chef/data_bag.rb
@@ -94,7 +94,7 @@ class Chef
         names = []
         paths.each do |path|
           unless File.directory?(path)
-            raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' is invalid"
+            raise Chef::Exceptions::InvalidDataBagPath, "Please create databag folder at '#{path}'"
           end
 
           names += Dir.glob(File.join(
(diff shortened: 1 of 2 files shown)


phiggins:
This is definitely an improvement, but I feel like it could be better. If creating the directory is the solution to this problem, why isn't the directory automatically created instead of asking the user to do it manually? I don't know the history of this issue so there might be reasons why that can't happen. In that case, I think an error message more similar to the original might be more appropriate, something like:

> Data bag path '#{path}' does not exist.

Additionally, since multiple directories could be specified, it would be a frustrating experience to be told "Please create /directory1" then to run the command again and be told "Please create /directory2". This could be worked around by checking for the existence of all specified directories before doing any of the work, such as:

```
invalid_paths = paths.reject { |path| File.directory?(path) }
if invalid_paths.any?
  invalid_paths.map! { |path| %Q{"#{path}"} }
  raise Chef::Exceptions::InvalidDataBagPath, "Data bag path(s) do not exist: #{invalid_paths.join(", ")}"
end

paths.each do |path|
  # Do the rest of the work here
end
```

lib/chef/data_bag.rb
```diff
@@ -94,7 +94,7 @@ def self.list(inflate = false)
         names = []
         paths.each do |path|
           unless File.directory?(path)
-            raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' is invalid"
+            raise Chef::Exceptions::InvalidDataBagPath, "Please create databag folder at '#{path}'"
```
tas50, line 97:
```suggestion
            raise Chef::Exceptions::InvalidDataBagPath, "Data bag path '#{path}' not found. Please create this directory."
```

PR #9961, opened 2020-06-05
Fix wrong unit test exposed by cleaning up rspec deprecations.

phiggins:
These tests were broken in #9937 because they are doing exactly what that warning was about: hiding a real exception. It also turns out that the test was wrong because that code is supposed to raise an exception in this case.


lamont-granquist:
Yeah I've been forcing those warnings to be errors in all the knife plugins since I found a real issue in one of them.

PR #9963, opened 2020-06-07
Fix how enforce_license is set in run method for chef-apply

ramereth:
I recently discovered that when using ``Chef::Application::Apply.new.run(enforce_license: false)``, chef-apply will still run the license verification when I expected it to be disabled. It seems this feature works if you install this as a gem, however if you build chef using omnibus with this set, it refuses to work. I suspect this might be something with appbundler.

I noticed that other binaries (i.e. solo) were using ``run(enforce_license: false)`` instead of ``run(enforce_license = false)`` which chef-apply uses. If I changed this to ``run(enforce_license: false)``, this worked as expected.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/application/apply.rb
+++ b/lib/chef/application/apply.rb
@@ -233,7 +233,7 @@ class Chef::Application::Apply < Chef::Application
   end
 
   # Get this party started
-  def run(enforce_license = false)
+  def run(enforce_license: false)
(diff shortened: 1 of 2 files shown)


tas50:
@ramereth There's some failing specs here. Can you try rebasing this against master

PR #9969, opened 2020-06-09
Nicer error for misconfigured spellcheck in CI

phiggins:

### DIFF
--- a/.expeditor/verify_public.pipeline.yml
+++ b/.expeditor/verify_public.pipeline.yml
@@ -576,7 +576,7 @@ steps:
   commands:
     - ruby -rjson -e "JSON.parse(File.read('cspell.json'))" 2>/dev/null || (echo "Failed to parse config file 'cspell.json', skipping spellcheck" && exit 1)
     - npm install -g cspell
-    - cspell "**/*" "*.md"
+    - cspell "**/*"
   expeditor:
     executor:
       docker:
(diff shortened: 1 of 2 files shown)


.expeditor/verify_public.pipeline.yml
```diff
@@ -574,8 +574,9 @@ steps:
 
 - label: "Spellcheck"
   commands:
+    - ruby -rjson -e "JSON.parse(File.read('cspell.json'))" 2>/dev/null || (echo "Failed to parse config file 'cspell.json', skipping spellcheck" && exit 1)
```
phiggins, line 577: I tried to use node here, but because cspell uses a non-standard form of json that allows comments, node's json parser blew up but ruby's worked.

.expeditor/verify_public.pipeline.yml
```diff
@@ -574,8 +574,9 @@ steps:
...
     - npm install -g cspell
-    - cspell "**/*" "*.md"
+    - cspell "**/*"
```
phiggins, line 579: Everything in the `"*.md"` glob is also in the `"**/*` glob, so it caused those files to be spellchecked twice.

tas50:
Any chance we could setup a single cspell config and just pull that down somehow so that we could maintain a list of allowed spellings as a company?

PR #9972, opened 2020-06-09
Make sure a password is passed with knife bootstrap -P

tas50:
It's common with a lot of applications that the password flag w/o an option means prompt for password. That's not how it works with Knife and it's very hard to troubleshoot. Instead we eat the next flag and make that the password. Then we throw a very cryptic auth error since the password is not "--sudo". This validates the password field doesn't start with - and a single letter or -- and then letters or an underscore. The chances of a false positive seem incredibly low here and it should help folks with a day 1 problem that is quite awful.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -42,7 +42,7 @@ class Chef
         long: "--connection-password PASSWORD",
         description: "Authenticate to the target host with this password.",
         proc: Proc.new { |v|
-          # If a user passes -P and leaves out the password the knife interprets the next flag as the password
+          # If a user passes -P and leaves out the password knife interprets the next flag as the password
           # which is incredibly hard to troubleshoot. Let's help the user out here by seeing if a short or long
           # flag is in the password field.
           # regular expression explaination: https://rubular.com/r/gcknBouOOHokhT
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -40,7 +40,16 @@ class Chef
       option :connection_password,
         short: "-P PASSWORD",
         long: "--connection-password PASSWORD",
-        description: "Authenticate to the target host with this password."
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap.rb
```diff
@@ -40,7 +40,16 @@ class Bootstrap < Knife
...
         short: "-P PASSWORD",
         long: "--connection-password PASSWORD",
-        description: "Authenticate to the target host with this password."
+        description: "Authenticate to the target host with this password.",
+        proc: Proc.new { |v|
+          # If a user passes -P and leaves out the password the knife interprets the next flag as the password
+          # which is incredibly hard to troubleshoot. Let's help the user out here by seeing if a short or long
+          # flag is in the password field.
+          # regular expression explaination: https://rubular.com/r/gcknBouOOHokhT
+          if v.match?(/^(-([a-zA-Z])$)|(--([a-z]|_)*)$/)
```
phiggins, line 49: This should probably include `-` in the second part, as it is, `--connection-password --connection-password` wouldn't get caught by this.

lib/chef/knife/bootstrap.rb
```diff
@@ -40,7 +40,16 @@ class Bootstrap < Knife
       option :connection_password,
         short: "-P PASSWORD",
         long: "--connection-password PASSWORD",
-        description: "Authenticate to the target host with this password."
+        description: "Authenticate to the target host with this password.",
+        proc: Proc.new { |v|
+          # If a user passes -P and leaves out the password the knife interprets the next flag as the password
```
phiggins, line 45:
```suggestion
          # If a user passes -P and leaves out the password knife interprets the next flag as the password
```

tas50:
Yeah ideally we'd handle this in mixlib-cli somehow, but it's a bit of a mess and I'm not sure if there's other areas where you'd do this an not be able to see the error clearly.

tas50:
Closing this out since it's not the right approach

PR #9974, opened 2020-06-10
Warn during bootstrapping when using validation keys

tas50:
We should highly encourage users to move to validatorless bootstrapping
 1) It's more secure
 2) It requires less admin work since there no validation key to pass
    around or rotate

This also provides an improved message helping the user to move off the
old keys and fixes some bad formatting where the 2nd line was indented
about 40 columns.

### DIFF
--- a/lib/chef/knife/bootstrap.rb
+++ b/lib/chef/knife/bootstrap.rb
@@ -580,11 +580,8 @@ class Chef
 
           bootstrap_context.client_pem = client_builder.client_path
         else
-          ui.info <<~EOM
-            Performing legacy client registration with the validation key at #{Chef::Config[:validation_key]}...
-            Delete your validation key in order to use your user credentials for client registration instead.
-          EOM
-
+          ui.warn "Performing legacy client registration with the validation key at #{Chef::Config[:validation_key]}..."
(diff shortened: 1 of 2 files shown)


stevenoneill-jmfamily:
@tas50 my understanding has been that validator based bootstrap was the best option for things like nodes in AWS AutoScaleGroups.  Is there an easy ASG friendly validatorless pattern or should we basically ignore this warning if we're using validator-based bootstrapping in an ASG?

tas50:
My hope here (and I'm fine being told I'm totally wrong) is that no one should be passing around keys to admin's workstations like it's 2013. Since this is a message that humans see during that sort of process they should be presented with a warning. This just mostly flips the conversation. Before we assumed the default was validator based and we had to sell you on validatorless. Now we assume the default is validatorless and warn you against using shared validator keys. 

I made a similar docs change here: https://github.com/chef/chef-web-docs/commit/1456c0d84ea1d6d4aea46ab5ef9a86cd8909f446

I assume in ASG you use the validation key, but you don't actually run knife on the nodes right? When I used Chef (many moons ago now) with scaling groups we baked the validation key into our AMIs and then kicked off chef to register the node during provisioning.

stevenoneill-jmfamily:
Correct, no running knife bootstrap on ASG nodes at this time, so I suppose we won't be seeing this warning.  Sorry about that, got excited too quickly for the wrong reasons, my apologies 🤦‍♂️

Issue #9832, opened 2020-05-09
apt purges are not idempotent


jaymzh:
## Description

When apt does a `purge`, if any package in the list is installed, it tries to uninstall them all, cfengine2 style.

## Chef Version
15.9.17

## Platform Version
Ubuntu 18.04

## Replication Case
```ruby
unwanted = [
  'whoopsie',
  'rtkit',
  'speech-dispatcher',
  'speech-dispatcher-audio-plugins',
  'kerneloops-daemon',
  'puppetlabs-release-pc1',
  'puppet',
  'landscape-common',
  'landscape-client',
  'avahi-autoipd',
  'avahi-daemon',
]

package unwanted do
  action :purge
end
```

## Client Output

```
---- Begin output of ["apt-get", "-q", "-y", "purge", "whoopsie", "rtkit", "speech-dispatcher", "speech-dispatcher-audio-plugins", "kerneloops", "puppetlabs-release-pc1", "puppet", "landscape-common", "landscape-client", "avahi-autoipd", "avahi-daemon"] ----
STDOUT: Reading package lists...
Building dependency tree...
Reading state information...
STDERR: E: Unable to locate package puppetlabs-release-pc1
---- End output of ["apt-get", "-q", "-y", "purge", "whoopsie", "rtkit", "speech-dispatcher", "speech-dispatcher-audio-plugins", "kerneloops", "puppetlabs-release-pc1", "puppet", "landscape-common", "landscape-client", "avahi-autoipd", "avahi-daemon"] ----
Ran ["apt-get", "-q", "-y", "purge", "whoopsie", "rtkit", "speech-dispatcher", "speech-dispatcher-audio-plugins", "kerneloops", "puppetlabs-release-pc1", "puppet", "landscape-common", "landscape-client", "avahi-autoipd", "avahi-daemon"] returned 100
```

In this case one of the packages (`landscape-common`) was there so it needed to be removed, however, it tried uninstalling all of them as you can see above. Turns out that one of those packages isn't on the repos and thus apt barfed.

```
# dpkg -l 'whoopsie' 'rtkit' 'speech-dispatcher' 'speech-dispatcher-audio-plugins' 'kerneloops-daemon' 'puppetlabs-release-pc1' 'puppet' 'landscape-common' 'landscape-client'
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
[...]

PR #9976, opened 2020-06-10
 Fix apt purge idempotent

dheerajd-msys:

## Description

When apt does a purge, if any package in the list is installed, it tries to uninstall them all
## Related Issue

Fixes https://github.com/chef/chef/issues/9832
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -235,10 +235,11 @@ class Chef
         end
 
         def check_availability(name)
-          name.map do |pkg|
+          available_packages = name.map do |pkg|
             showpkg = run_noninteractive("apt-cache", "search", pkg).stdout
             showpkg.empty? ? logger.warn("Unable to locate package  #{pkg} ") : pkg
           end
+          available_packages.compact
         end
       end
     end
--- a/lib/chef/provider/package/apt.rb
+++ b/lib/chef/provider/package/apt.rb
@@ -89,6 +89,7 @@ class Chef
         end
 
         def install_package(name, version)
+          name = check_availablity(name)
           package_name = name.zip(version).map do |n, v|
             package_data[n][:virtual] ? n : "#{n}=#{v}"
           end
@@ -101,6 +102,7 @@ class Chef
         end
 
         def remove_package(name, version)
+          name = check_availablity(name)
           package_name = name.map do |n|
             package_data[n][:virtual] ? resolve_virtual_package_name(n) : n
           end
@@ -108,6 +110,7 @@ class Chef
         end
 
         def purge_package(name, version)
+          name = check_availablity(name)
           package_name = name.map do |n|
             package_data[n][:virtual] ? resolve_virtual_package_name(n) : n
           end
@@ -115,10 +118,12 @@ class Chef
         end
 
         def lock_package(name, version)
+          name = check_availablity(name)
           run_noninteractive("apt-mark", options, "hold", name)
         end
 
         def unlock_package(name, version)
+          name = check_availablity(name)
           run_noninteractive("apt-mark", options, "unhold", name)
         end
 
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/apt.rb
```diff
@@ -229,6 +234,12 @@ def package_data_for(pkg)
           }
         end
 
+        def check_availability(name)
+          name.map do |pkg|
+            showpkg = run_noninteractive("apt-cache", "search", pkg).stdout
+            showpkg.empty? ? logger.warn("Unable to locate package  #{pkg} ") : pkg
+          end
```
jaymzh, line 241: do you nee to call `.compact` here so you don't return a bunch of `nil`s?

lib/chef/provider/package/apt.rb
```diff
@@ -89,6 +89,7 @@ def locked_packages
         end
 
         def install_package(name, version)
+          name = check_availability(name)
```
jaymzh, line 92: This only fixes... part of the bug.

It will ensure that if you try to purge `['a', 'b', 'c']` and `c` isn't a known package, that it won't try to purge `c`... which is good...

But the actual bug is that if `a` is installed, it'll automatically try to uninstall `b`, and `c` even if they are not installed.

So imagine:
* a - installed
* b - uninstalled, but available
* c - uninstalled and unavailable

Current behavior is Chef will try to purge `a`, `b` and `c`

With your PR it'll just try to purge `a` and `b`

But the *right* behavior is that it should only try to purge `a`.

lib/chef/provider/package/apt.rb
dheerajd-msys: Thanks @jaymzh, Updated.

lamont-granquist:
This doesn't seem to be the right approach, I wouldn't expect that install_package would be getting updated. It should also be pushed into fixing the args for remove_package/purge_package so that the superclass passes the right set of packages into remove_package/purge_package.  That should be done by modifying stuff like candidate_version/installed_version/etc so that the superclass (unmodified) selects the right set of packages.

(I mean its possible that there's some across-the-board superclass bug, but its unlikely)

lamont-granquist:
so probably `resolve_package_versions` should be modified to correctly populate the current_versions and to filter out these packages so that they're not purged.  there may need to be a conditional put in there to switch based on the `action` being run.

dheerajd-msys:
@lamont-granquist Thanks for the review. We looked into the working of the remove/purge function of yum resource. To make sure it doesn't run/purge a non-existing package it checks the current_version and if it exists (valid version) it adds that to the remove/purge list, based on this we came up with a solution(for the apt resource) of the same sort where we are checking the current_version of the package to be removed and if that exists then only to remove the package.

So currently working on code update.

lamont-granquist:
You're still not going about fixing this correctly.

The provider most likely needs to be updated to have `allow_nils` added:

https://github.com/chef/chef/blob/master/lib/chef/provider/package/yum.rb#L36-L37

This may be 90% of the problem.  That will mean that the methods called from the superclass will have nil values in the name and version arguments to e.g.  `install_package(name, version)` so all those methods need to skip over nils there in the array of names.

Then the `current_resource.version` array needs to be populated correctly with nil values where the package is not installed.  The purpose of the nil check in the yum provider is to skip over those (but based on the information provided by the superclass).

Then once that is done the name array and version array passed to e.g. purge_package is then indexed exactly the same was as the package_name_array.  It would be better to introduce a similar installed_version and available_version helper that were indexed based on the index position rather than the name.

tas50:
@dheerajd-msys Is there an update on this issue?

tas50:
@dheerajd-msys Is this one you'll be picking back up or should we close it out?

dheerajd-msys:
@tas50 I won't be able to pick that up for now.

johnmccrae:
per last comment, closing this for now.

PR #9989, opened 2020-06-11
Use .match? not =~ when match values aren't necessary

tas50:
Autocorrected from RuboCop Performance which is now smart enough to detect when you use the match and when you don't. Using match? does not create any objects so it's slightly faster and uses less memory.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,7 +84,7 @@ eval_gemfile("./Gemfile.local") if File.exist?("./Gemfile.local")
 #
 # We copy (and overwrite) these files every time "bundle <exec|install>" is
 # executed, just in case they have changed.
-if RUBY_PLATFORM =~ /mswin|mingw|windows/
+if RUBY_PLATFORM.match?(/mswin|mingw|windows/)
   instance_eval do
     ruby_exe_dir = RbConfig::CONFIG["bindir"]
     assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", Dir.pwd) + "/*.dll")
--- a/lib/chef/application/base.rb
+++ b/lib/chef/application/base.rb
@@ -363,7 +363,7 @@ class Chef::Application::Base < Chef::Application
     Chef::Log.trace("Download recipes tarball from #{url} to #{path}")
     if File.exist?(url)
       FileUtils.cp(url, path)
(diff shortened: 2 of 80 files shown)


Gemfile
```diff
@@ -84,7 +84,7 @@ eval_gemfile("./Gemfile.local") if File.exist?("./Gemfile.local")
 #
 # We copy (and overwrite) these files every time "bundle <exec|install>" is
 # executed, just in case they have changed.
-if RUBY_PLATFORM =~ /mswin|mingw|windows/
+if RUBY_PLATFORM.match?(/mswin|mingw|windows/)
```
phiggins, line 87: String#match? was added in ruby 2.4, does chef support anything older than that?

tas50: We're 2.6+ now. I've been waiting a while to use this, but the rubocop autocorrect before destroyed code.

lib/chef/application/base.rb
```diff
@@ -363,7 +363,7 @@ def fetch_recipe_tarball(url, path)
     Chef::Log.trace("Download recipes tarball from #{url} to #{path}")
     if File.exist?(url)
       FileUtils.cp(url, path)
-    elsif url =~ URI.regexp
+    elsif url&.match?(URI.regexp)
```
phiggins, line 366: This nil check could be avoided by flipping these around.

```suggestion
    elsif URI.regexp.match?(url)
```

spec/data/lwrp/providers/embedded_resource_accesses_providers_scope.rb
```diff
@@ -24,5 +24,5 @@ def action_twiddle_thumbs
 end
 
 def generate_new_name(str, &block)
-  "#{str}, #{block.call}"
+  "#{str}, #{yield}"
```
phiggins, line 27: If you're not using the `block` parameter they can be removed from the method's signature.

tas50: Let me revertthis. I was trying to stick to just the match? fixes in this PR

lib/chef/application/base.rb
tas50: Fixed

PR #9994, opened 2020-06-15
Create the windows_firewall_profile resource for use with enabling/disabling and configuring the Windows firewall Domain, Private, and Public profiles

chef-davin:

## Description

Added the `windows_firewall_profile` resource with the `:enable` and `:disable` actions to allow you to enable Windows firewall profiles as well as modify specific settings for those profiles.  It can be used as so:

**Using the `:enable` action with some of the configuration options with the Private firewall profile**
  ```ruby
  windows_firewall_profile 'Private' do
    default_inbound_action 'Block'
    default_outbound_action 'Allow'
    allow_inbound_rules true
    display_notification false
    action :enable
  end
  ```
**Using the `:enable` action with some of the configuration options with the Public firewall profile**
  ```ruby
  windows_firewall_profile 'Public' do
    default_inbound_action 'Allow'
    default_outbound_action 'Allow'
    allow_inbound_rules false
    display_notification false
    action :enable
  end
  ```
**Using the `:disable` action against the Domain firewall profile**
  ```ruby
  windows_firewall_profile 'Disable Windows Firewall Domain Profile' do
    profile 'Domain'
    action :disable
  end
  ```
## Types of changes

## Checklist:


lib/chef/provider/package/snap.rb
```diff
@@ -137,7 +137,7 @@ def call_snap_api(method, uri, post_data = nil?)
 
           # while it is expected to allow clients to connect using https over
           # a tcp socket, at this point only a unix socket is supported. the
-          # socket is /run/snapd.socket note - unixsocket is not defined on
+          # socket is /run/snapd.socket note - unix socket is not defined on
```
phiggins, line 140: Very nitpicky, but I think referring to the class name is appropriate here.
```suggestion
          # socket is /run/snapd.socket note - UNIXSocket is not defined on
```

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,167 @@
...
+      ```ruby
+      windows_firewall_profile 'Configure and Enable Windows Firewall Private Profile' do
+        profiles 'Private'
+        profile_enabled true
+        default_inbound_block true
+        default_outbound_allow true
+        allow_inbound_rules true
+        display_notification false
+      end
+      ```
```
phiggins, line 38: Is it significant that this example does not have `action :configure` like the two below?

chef-davin: Not really. I'll update it. Probably deleted it by accident.

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,168 @@
...
+          $#{desired.profile}.DefaultOutboundAction
+          $#{desired.profile}.AllowInboundRules
+          $#{desired.profile}.AllowLocalFirewallRules
+          $#{desired.profile}.AllowLocalIPsecRules
+          $#{desired.profile}.AllowUserApps
+          $#{desired.profile}.AllowUserPorts
+          $#{desired.profile}.AllowUnicastResponseToMulticast
+          $#{desired.profile}.NotifyOnListen
+        CODE
+        profile_enabled case ps_results[0]; when 0 then false; when 1 then true; end
```
tas50, line 93: I think we can do this a bit differently.  Properties default to nil if not set so we can set them to true / false and not set them at all otherwise. In that case current_resource.WHATEVER should be nil. This would allow you to remove nil as an allowed value from all of these properties and allows removing these inline case statements. You might want to just create a simple method to return true/false from 1/2. It would really help the readability here.

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,168 @@
...
+        allow_inbound_rules case ps_results[3]; when 0 then false; when 1 then true; when 2 then nil; end
+        allow_local_firewall_rules case ps_results[4]; when 0 then false; when 1 then true; when 2 then nil; end
+        allow_local_ipsec_rules case ps_results[5]; when 0 then false; when 1 then true; when 2 then nil; end
+        allow_user_apps case ps_results[6]; when 0 then false; when 1 then true; when 2 then nil; end
+        allow_user_ports case ps_results[7]; when 0 then false; when 1 then true; when 2 then nil; end
+        allow_unicast_response case ps_results[8]; when 0 then false; when 1 then true; when 2 then nil; end
+        display_notification case ps_results[9]; when 0 then false; when 1 then true; when 2 then nil; end
+      end
+
+      action :configure do
```
tas50, line 105: I'd change this action to :enable. That way you remove the need for the profile_enabled property. That just becomes the action and then you add a disable action that only does the converge_by disable. The

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,168 @@
...
+        profile 'Domain'
+        profile_enabled false
+        action :configure
+      end
+      ```
+      DOC
+
+      unified_mode true
+
+      property :profile, String, required: true, equal_to: %w{ Domain Public Private }, description: "Set the Windows Profile being configured"
```
tas50, line 67: This should probably be the name_property so someone can just do 
```ruby
windows_firewall_profile %w(Domain Public) do
  whatever whatevs
end
```

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+
+      action :configure do
+        converge_if_changed :profile_enabled do
+          cmd = "Set-NetFirewallProfile -Profile #{new_resource.profile} "
+          cmd += "-Enabled #{new_resource.profile_enabled ? "True" : "False"} "
+          powershell_exec(cmd)
+        end
+        converge_if_changed :default_inbound_block do
+          cmd = "Set-NetFirewallProfile -Profile #{new_resource.profile} "
```
tas50, line 112: Can we create a method here that builds a single Set-NetFirewallProfile  that includes all the necessary values instead so we can just shell out once. If the user changes a bunch of things here this is a lot of shelling out.

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,173 @@
...
+      property :allow_local_firewall_rules, String, equal_to: %w{ True False NotConfigured }, description: "Merges inbound firewall rules into the policy"
+      property :allow_local_ipsec_rules, String, equal_to: %w{ True False NotConfigured }, description: "Allow users to manage local connection security rules"
+      property :allow_user_apps, String, equal_to: %w{ True False NotConfigured }, description: "Allow user applications to manage firewall"
+      property :allow_user_ports, String, equal_to: %w{ True False NotConfigured }, description: "Allow users to manage firewall port rules"
+      property :allow_unicast_response, String, equal_to: %w{ True False NotConfigured }, description: "Allow unicast responses to multicast and broadcast messages"
+      property :display_notification, String, equal_to: %w{ True False NotConfigured }, description: "Display a notification when firewall blocks certain activity"
+
+      load_current_value do |desired|
+        ps_get_net_fw_profile = load_firewall_state(desired.profile)
+        output = powershell_out(ps_get_net_fw_profile)
```
tas50, line 88: This should use powershell_exec since it's much faster

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,173 @@
...
+        if output.stdout.empty?
+          current_value_does_not_exist!
+        else
+          state = Chef::JSONCompat.from_json(output.stdout)
+        end
+
+        default_inbound_action state["default_inbound_action"]
```
tas50, line 95: Once this is using powershell_exec you shouldn't need to do any of the json conversion here

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,173 @@
...
+    class WindowsFirewallProfile < Chef::Resource
+      provides :windows_firewall_profile
+      description "Use the **windows_firewall_profile** resource to enable, disable, and configure the Windows firewall."
+      introduced "16.3"
+
+      examples <<~DOC
+      **Enable and Configure the Private Profile of the Windows Profile**:
+
+      ```ruby
+      windows_firewall_profile 'Configure and Enable Windows Firewall Private Profile' do
```
tas50, line 30: Let's make this first example not use the profile property and instead have it use the resource name so:

 ```ruby
      windows_firewall_profile 'Private' do
        default_inbound_action 'Block'
        default_outbound_action 'Allow'
        allow_inbound_rules 'True'
        display_notification 'False'
        action :enable
      end
      ```

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,173 @@
...
+        equal_to: %w{ Allow Block NotConfigured },
+        description: "Set the default policy for inbound network traffic"
+
+      property :default_outbound_action, [String, nil],
+        equal_to: %w{ Allow Block NotConfigured },
+        description: "Set the default policy for outbound network traffic"
+
+      property :allow_inbound_rules, String, equal_to: %w{ True False NotConfigured }, description: "Allow users to set inbound firewall rules"
+      property :allow_local_firewall_rules, String, equal_to: %w{ True False NotConfigured }, description: "Merges inbound firewall rules into the policy"
+      property :allow_local_ipsec_rules, String, equal_to: %w{ True False NotConfigured }, description: "Allow users to manage local connection security rules"
```
tas50, line 80: The true/false values should just take true/false and not a string.

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -0,0 +1,173 @@
...
+      DOC
+
+      unified_mode true
+
+      property :profile, String,
+        name_property: true,
+        equal_to: %w{ Domain Public Private },
+        description: "Set the Windows Profile being configured"
+
+      property :default_inbound_action, [String, nil],
```
tas50, line 70: You shouldn't need to allow Nil here. String defaults to nil if not set

phiggins:
Is there a reason to not use `nil` for the not-configured state instead of `"NotConfigured"`? That seems to be the convention for the "user didn't specify a value so they get the default behavior" case.

chef-davin:
> Is there a reason to not use `nil` for the not-configured state instead of `"NotConfigured"`? That seems to be the convention for the "user didn't specify a value so they get the default behavior" case.

This is a tricky one for me @phiggins. It's because "NotConfigured" is an actual state in the Set-NetFirewallProfile cmdlet.  I've tried to use this with `nil`, but if I translate `"NotConfigured"` to `nil` and I don't specify `nil` as an acceptable property type, the `load_current_value` spazzes out on me.  If I do make `nil` a property type, It makes my logic in building my `Set-NetFirewallProfile` powershell command annoyingly impossible.

I can keep trying to play around with it more, and put some complexity into the `load_current_value` bit, so that if it's not used it doesn't try to set a value.  I just don't know a good way to set up the logic to build the powershell cmd where it can distinguish between a `nil` value where the user doesn't want to change whatever the current value is, and someone wanting to force the settings back to defaults for that profile.

lamont-granquist:
nil should mean "leave whatever you find along because the user didn't care"

if the user omitting `default_inbound_block` from the resource means that a true or false value MUST be changed to "NotConfigured" then the resource is written correctly right now.

if the user omitting `default_inbound_block` should mean that any setting of that which is found in the current_value should be preserved, then you need to do some nil handling -- at least you'd need to make the property nillable and nil by default, and then preserve "NotConfigured" as an explicit option.

knowing exactly zero about the domain I can't tell you which to pick, but those are sort of the choices.

phiggins:
> This is a tricky one for me @phiggins. It's because "NotConfigured" is an actual state in the Set-NetFirewallProfile cmdlet.

Ok, that makes a lot more sense. I thought we were dealing with 3 states but its actually 4, so keeping things as-is seems fine.

tas50:
I think we're at the point where this is in a good state to get it in the hands of folks that can tell us if anything is mising. Thanks for putting this together @chef-davin

Issue #9872, opened 2020-05-14
Requesting revision master for a git resource fails in chef 16


tomhughes:
## Description

A git resource that requests revision `master` fails in chef 16 when it tries to force update master to the revision of it's own tip.

This works fine in chef 14 and 15 and is presumably related to the changes in the way the git resource works.

## Chef Version

16.0.287

## Platform Version

Ubuntu 18.04 with git 2.26.2 from the git-core PPA

## Replication Case

```
git "/usr/local/lib/supybot/plugins/Git" do
  action :sync
  repository "https://github.com/openstreetmap/supybot-git"
  revision "master"
  depth 1
  user "root"
  group "root"
end
```

## Client Output

```
Mixlib::ShellOut::ShellCommandFailed
------------------------------------
Expected process to exit with [0], but received '128'
---- Begin output of git branch -f master 1d6df75f919451a8c1ac7f2cb29766e98bf283d0 ----
STDOUT: 
STDERR: fatal: Cannot force update the current branch.
---- End output of git branch -f master 1d6df75f919451a8c1ac7f2cb29766e98bf283d0 ----
Ran git branch -f master 1d6df75f919451a8c1ac7f2cb29766e98bf283d0 returned 128
```

johngun3:
This problem is caused by the changes in this commit https://github.com/chef/chef/commit/165685fdcbd4fbe8138d37d1285b633e3958cf09 in the `fetch_updates` method . It used to just `git reset --hard` but now it uses `git branch -f` which causes a problem when the same branch is updated.

KindAchmed:
I'm having the same issue with 16.1.16. 
Is there any kind of work around in order to quickly fix this issue?

lotooo:
Same issue here.

The work around I want to use for now is doing a `execute "git checkout deploy"` to switch to a different branch before the call to the `git` resource. But it's annoying.

lotooo:
I pushed a PR (still in validation process) : https://github.com/chef/chef/pull/9998

This PR tests if it's already using the proper branch and does a `git reset` instead of the `git branch -f` 

I overrided the `git.rb` provider with this version and it seems to work for us.

gyamada619:
This issue is causing failures on a number of our Chef runs as well.

lotooo:
This [PR](https://github.com/chef/chef/pull/9998) has been merged to `master`

PR #9998, opened 2020-06-15
Fix git sync if the local branch already exist

lotooo:
Signed-off-by: Matthieu Serrepuy <lotooo@gmail.com>

This PR tried to fix the git resource when a `:sync` action is triggerred.

## Description

The `git` resouce is currently failing in chef-client 16 (tested with 16.1.16) when a new commit needs to be synced because you can't do a `git branch -f branch_name branch_name` if you are already on the  branch  `branch`.

## Related Issue
https://github.com/chef/chef/issues/9872

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/git.rb
+++ b/lib/chef/provider/git.rb
@@ -154,6 +154,11 @@ class Chef
         sha_hash?(result) ? result : nil
       end
 
+      def already_on_target_branch?
+        current_branch = git("rev-parse", "--abbrev-ref", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip
+        current_branch == (new_resource.checkout_branch || new_resource.revision)
+      end
+
       def add_remotes
         if new_resource.additional_remotes.length > 0
           new_resource.additional_remotes.each_pair do |remote_name, remote_url|
@@ -193,6 +198,9 @@ class Chef
             # detached head
             git("checkout", target_revision, cwd: cwd)
             logger.info "#{new_resource} checked out reference: #{target_revision}"
+          elsif already_on_target_branch?
+            # we are already on the proper branch
+            git("reset", "--hard", target_revision, cwd: cwd)
           else
             # need a branch with a tracking branch
             git("branch", "-f", new_resource.revision, target_revision, cwd: cwd)
@@ -222,13 +230,13 @@ class Chef
           logger.trace "Fetching updates from #{new_resource.remote} and resetting to revision #{target_revision}"
           git("fetch", "--prune", new_resource.remote, cwd: cwd)
           git("fetch", new_resource.remote, "--tags", cwd: cwd)
-          if new_resource.checkout_branch
+          if sha_hash?(new_resource.revision) || is_tag? || already_on_target_branch?
+            # detached head or if we are already on the proper branch
+            git("reset", "--hard", target_revision, cwd: cwd)
+          elsif new_resource.checkout_branch
             # check out to a local branch
             git("branch", "-f", new_resource.checkout_branch, target_revision, cwd: cwd)
             git("checkout", new_resource.checkout_branch, cwd: cwd)
-          elsif sha_hash?(new_resource.revision) || is_tag?
-            # detached head
-            git("reset", "--hard", target_revision, cwd: cwd)
           else
             # need a branch with a tracking branch
             git("branch", "-f", new_resource.revision, target_revision, cwd: cwd)
--- a/spec/functional/resource/git_spec.rb
+++ b/spec/functional/resource/git_spec.rb
@@ -239,6 +239,28 @@ describe Chef::Resource::Git, requires_git: true do
     end
   end
 
+  context "when updating a branch that's already checked out out" do
+    it "checks out master, commits to the repo, and checks out the latest changes" do
+      git deploy_directory do
+        repository origin_repo
+        revision "master"
+        action :sync
+      end.should_be_updated
+
+      # We don't have a way to test a commit in the git bundle
+      # Revert to a previous commit in the same branch and make sure we can still sync.
+      shell_out!("git", "reset", "--hard", rev_foo, cwd: deploy_directory)
+
+      git deploy_directory do
+        repository origin_repo
+        revision "master"
+        action :sync
+      end.should_be_updated
+      expect_revision_to_be("HEAD", rev_head)
+      expect_branch_to_be("master")
+    end
+  end
+
   context "when dealing with a repo with a degenerate tag named 'HEAD'" do
     before do
       shell_out!("git", "tag", "-m\"degenerate tag\"", "HEAD", "ed181b3419b6f489bedab282348162a110d6d3a1", cwd: origin_repo)


lib/chef/provider/git.rb
```diff
@@ -154,6 +154,14 @@ def find_current_revision
         sha_hash?(result) ? result : nil
       end
 
+      def already_on_target_branch?
```
johngun3, line 157: This function could take in the target branch and just look like this:
```
def already_on_target_branch?(branch)
  git("rev-parse", "--abbrev-ref", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip == branch
end
```

Or if you don't want to have any inputs you only need to write the git command a single time which makes for less repetitive code:
```
current_branch = git("rev-parse", "--abbrev-ref", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip
if new_resource.checkout_branch
  current_branch == new_resource.checkout_branch
else
  current_branch == new_resource.revision
end
```

lotooo: I think it is easier to read if no argument to that branch. I made the requested change !

lib/chef/provider/git.rb
```diff
@@ -154,6 +154,15 @@ def find_current_revision
...
+        current_branch = git("rev-parse", "--abbrev-ref", "HEAD", cwd: cwd, returns: [0, 128]).stdout.strip
+        if new_resource.checkout_branch
+          current_branch == new_resource.checkout_branch
+        else
+          current_branch == new_resource.revision
+        end
```
phiggins, line 163:
```suggestion
        current_branch == (new_resource.checkout_branch || new_resource.revision)
```

This could be made more succinct.

lotooo: ohh nice. Pushed it

lib/chef/provider/git.rb
```diff
@@ -224,16 +229,24 @@ def fetch_updates
...
-            git("checkout", new_resource.revision, cwd: cwd)
-            git("branch", "-u", "#{new_resource.remote}/#{new_resource.revision}", cwd: cwd)
+            if already_on_target_branch?
+              git("reset", "--hard", target_revision, cwd: cwd)
+            else
+              git("branch", "-f", new_resource.revision, target_revision, cwd: cwd)
+              git("checkout", new_resource.revision, cwd: cwd)
+              git("branch", "-u", "#{new_resource.remote}/#{new_resource.revision}", cwd: cwd)
+            end
           end
```
phiggins, line 250: I'm not very familiar with how this code is supposed to work, but I'm curious if some of the branches of this conditional could be combined to reduce the nested conditions. If the check for `if new_resource.checkout_branch` does not strictly need to come first, you could do something like this, for example:

```
if sha_hash?(new_resource.revision) || is_tag? || already_on_target_branch?
   git("reset", "--hard", target_revision, cwd: cwd)
elsif new_resource.checkout_branch
  # ...
else
  # ...
end
```

lotooo: I think the logic would work, but not sure if the code wouldn't be harder to read

lotooo: I pushed your change.

lotooo:
> This'll need a functional test, see https://github.com/chef/chef/blob/master/spec/functional/resource/git_spec.rb

How can I add a commit to the remote during this test ?

Basically I need this to run : 

https://github.com/chef/chef/blob/master/spec/functional/resource/git_spec.rb#L221-L230

```
    it "checks out master as the default revision, a checkout again to grab the latest commit of the same branch" do
      git deploy_directory do
        repository origin_repo
      end.should_be_updated
      expect_revision_to_be("HEAD", rev_head)
      expect_branch_to_be("master")

      COMMIT SOMETHING IN THE REMOTE

      git deploy_directory do
        repository origin_repo
      end.should_be_updated
      expect_revision_to_be("HEAD", rev_head)
      expect_branch_to_be("master")
    end
```

 I am not sure how to perform the step 2 of that test.

KindAchmed:
> > This'll need a functional test, see https://github.com/chef/chef/blob/master/spec/functional/resource/git_spec.rb
> 
> How can I add a commit to the remote during this test ?
> 
> Basically I need this to run :
> 
> https://github.com/chef/chef/blob/master/spec/functional/resource/git_spec.rb#L221-L230
> 
> ```
>     it "checks out master as the default revision, a checkout again to grab the latest commit of the same branch" do
>       git deploy_directory do
>         repository origin_repo
>       end.should_be_updated
>       expect_revision_to_be("HEAD", rev_head)
>       expect_branch_to_be("master")
> 
>       COMMIT SOMETHING IN THE REMOTE
> 
>       git deploy_directory do
>         repository origin_repo
>       end.should_be_updated
>       expect_revision_to_be("HEAD", rev_head)
>       expect_branch_to_be("master")
>     end
> ```
> 
> I am not sure how to perform the step 2 of that test.

Is someone looking into this?

phiggins:
> How can I add a commit to the remote during this test ?

I can't comment on whether or not this is the correct thing to do, but other places in this test file interact with git via commands in `shell_out` calls, for example: https://github.com/chef/chef/blob/master/spec/functional/resource/git_spec.rb#L56

Would that work?

lotooo:
> Would that work?

Executing a command and test the output (or the return code) is not where I am blocked.

If we want to properly test that the `:sync` action properly work, we need "something external" to commit into the git repo between the 2 sync. This is the part that is currently failing.

The something external" is where I am blocked (and I guess the test is not currently done as it is failing with chef-client 16.1.16).

tas50:
@lotooo Can you rebase this against master as well please?

lotooo:
I just pushed a working test. Instead of a commit/push, I am doing a `git reset --hard` which perform the same goal/test. It should be ready now !

spec/functional/resource/git_spec.rb
```diff
@@ -239,6 +239,29 @@ def expect_branch_to_be(branch)
     end
   end
 
+  context "when updating a branch already checkout out" do
```
phiggins, line 242:
```suggestion
  context "when updating a branch that's already checked out" do
```

spec/functional/resource/git_spec.rb
```diff
@@ -239,6 +239,29 @@ def expect_branch_to_be(branch)
...
+    it "checks out master, commit ot the repo and checkout the last changes" do
```
phiggins, line 243:
```suggestion
    it "checks out master, commits to the repo, and checks out the latest changes" do
```

phiggins: Sorry if I'm misunderstanding, but it seems like this is describing what the test is doing rather than the behavior the test is supposed to demonstrate. Does this accurately describe the changes you made in this PR?
```suggestion
    it "syncs changes made remotely" do
```

spec/functional/resource/git_spec.rb
```diff
@@ -239,6 +239,29 @@ def expect_branch_to_be(branch)
...
+      git deploy_directory do
+        repository origin_repo
+        revision "master"
+        action :sync
+      end.should_be_updated
+
+      # We don't have a way to test a commit in the git bundle
+      # Let's revert to a previous commit in the same branch and make sure we
+      # can still sync
```
phiggins, line 252:
```suggestion
      # We don't have a way to test a commit in the git bundle.
      # Revert to a previous commit in the same branch and make sure we can still sync.
```

lotooo:
Thanks for the review and the help !

PR #10020, opened 2020-06-18
Add more examples to the resource code

tas50:
Also update the sudo resource command description based on an escalation. The commands need to be the full path to the command not just the command.

### DIFF
--- a/lib/chef/resource/sudo.rb
+++ b/lib/chef/resource/sudo.rb
@@ -139,7 +139,7 @@ class Chef
 
       # handle legacy cookbook property
       def after_created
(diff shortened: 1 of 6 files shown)


lib/chef/resource/sudo.rb
```diff
@@ -53,7 +80,7 @@ class Sudo < Chef::Resource
         coerce: proc { |x| coerce_groups(x) }
 
       property :commands, Array,
-        description: "An array of commands this sudoer can execute.",
+        description: "An array of full paths to commands this sudoer can execute.",
```
phiggins, line 83: "absolute paths", maybe?

lib/chef/resource/timezone.rb
```diff
@@ -28,6 +28,21 @@ class Timezone < Chef::Resource
...
+      **Set the timezone to UTC**
+
+      ```ruby
+      timezone 'UTC'
+      ```
+
+      **Set the timezone to UTC with a friendly resource name**
+
+      ```ruby
+      timezone 'Set the hosts timezone to UTC' do
```
phiggins, line 41:
```suggestion
      timezone 'Set the host's timezone to UTC' do
```
?

PR #10021, opened 2020-06-19
Setup cspell to pull from our common dictionary

tas50:
This makes it easier for us to use this on other projects.


Rakefile
```diff
@@ -111,6 +111,7 @@ end
 
 namespace :spellcheck do
   task :run do
+    sh 'wget https://raw.githubusercontent.com/chef/chef_dictionary/master/chef.txt -O chef_dictionary.txt'
```
phiggins, line 114: This should get moved to its own task then added as a dependency to this and the `:unknown_words` task. It's unfortunate that the file will be downloaded each time spellcheck is run, but I can't think of how to avoid that without ending up in a caching quagmire.

kagarmoe: How many repos does the common dictionary cover so far?

lamont-granquist:
chefstyle is annoyed about something though

Rakefile
tas50: Just chef/chef right now. It just makes it easier for us to roll this out further in the future.

PR #10024, opened 2020-06-22
Resource doc updates

phiggins:
This started from wanting to fix the busted examples section for chef_gem and grew from there:

Screen Shot 2020-06-22 at 12 38 34 PM

### DIFF
--- a/lib/chef/resource/chef_gem.rb
+++ b/lib/chef/resource/chef_gem.rb
@@ -47,6 +47,37 @@ class Chef
         Use the **gem_package** resource to install all other gems (i.e. install gems system-wide).
       DESC
 
+      examples <<~EXAMPLES
+        **Compile time vs. converge time installation of gems**
+
+        To install a gem while Chef Infra Client is configuring the node (the converge phase), set the `compile_time` property to `false`:
+        ```ruby
+        chef_gem 'right_aws' do
+          compile_time false
+          action :install
+        end
+        ```
+
+        To install a gem while the resource collection is being built (the compile phase), set the `compile_time` property to `true`:
+        ```ruby
+        chef_gem 'right_aws' do
+          compile_time true
+          action :install
+        end
+        ```
+
+        Install MySQL for Chef
+        ```ruby
+        apt_update
+
+        build_essential 'install compilation tools' do
+          compile_time true
+        end
+
+        chef_gem 'mysql'
+        ```
+      EXAMPLES
+
       property :package_name, String,
         description: "An optional property to set the package name if it differs from the resource block's name.",
         identity: true
--- a/lib/chef/resource/chef_gem.rb
+++ b/lib/chef/resource/chef_gem.rb
@@ -49,11 +49,14 @@ class Chef
       property :version, String,
         description: "The version of a package to be installed or upgraded."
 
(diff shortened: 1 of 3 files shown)


lib/chef/resource/chef_gem.rb
```diff
@@ -22,38 +22,74 @@
...
-    # accept the gem_binary property because it always uses the CurrentGemEnvironment under which the chef-client is
-    # running. In addition to performing actions similar to the gem_package resource, the chef_gem resource does the
-    # following:
-    #  - Runs its actions immediately, before convergence, allowing a gem to be used in a recipe immediately after it is
-    #    installed
-    #  - Runs Gem.clear_paths after the action, ensuring that gem is aware of changes so that it can be required
-    #    immediately after it is installed
-
-    require_relative "gem_package"
-    require_relative "../dist"
```
phiggins, line 39: These requires are also at the top of the file.

lib/chef/resource/chef_gem.rb
```diff
@@ -22,38 +22,74 @@
...
+
       property :package_name, String,
         description: "An optional property to set the package name if it differs from the resource block's name.",
         identity: true
 
       property :version, String,
         description: "The version of a package to be installed or upgraded."
 
-      property :gem_binary, default: "#{RbConfig::CONFIG["bindir"]}/gem", default_description: "The `gem` binary included with #{Chef::Dist::PRODUCT}.",
-                            description: "The path of a gem binary to use for the installation. By default, the same version of Ruby that is used by #{Chef::Dist::PRODUCT} will be installed.",
```
phiggins, line 53: > By default, the same version of Ruby that is used by #{Chef::Dist::PRODUCT} will be installed.

I updated this sentence because it made it sound like ruby was getting installed.

lib/chef/resource/gem_package.rb
```diff
@@ -53,7 +86,7 @@ class GemPackage < Chef::Resource::Package
         default: lazy { Chef::Config[:clear_gem_sources] }, desired_state: false
 
       property :gem_binary, String, desired_state: false,
-        description: "The path of a gem binary to use for the installation. By default, the same version of Ruby that is used by #{Chef::Dist::PRODUCT} will be installed."
+        description: "The path of a gem binary to use for the installation. By default, the same version of Ruby that is used by #{Chef::Dist::PRODUCT} will be used."
```
phiggins, line 89: > By default, the same version of Ruby that is used by #{Chef::Dist::PRODUCT} will be installed.

I updated this sentence because it made it sound like ruby was getting installed.

tasks/docs.rb
```diff
@@ -250,6 +236,38 @@ def special_properties(name, data)
       properties
     end
 
+    # Breaks a block of text into the different sections expected for the description,
+    # using the markers "Note:" for "note" sections and "Warning:" for "warning" sections.
+    # TODO: has the limitation that the plain description section is assumed to come first,
+    # and is followed by one or more "note"s or "warning"s sections.
```
phiggins, line 242: It probably wouldn't be too difficult to fix this TODO if we wanted to adopt a convention of a blank line in between sections being the end of a section or something like that.

PR #10027, opened 2020-06-23
specs: Remove test file that added a method to top-level scope.

phiggins:

While debugging a strange test failure, I figured out that some of my code in `lib/` was referencing `run_context` where it shouldn't, and it was getting this definition of it instead of the one inside Resource that I was expecting. From looking at the test failures that happened after deleting it, nothing was depending on its behavior explicitly.


phiggins:
This has a lot of legit test failures I didn't see locally, so marking as WIP for now.

phiggins:
I think I fixed all of the relevant test failures. Any I didn't catch should be fairly obvious because they'll complain about the method `run_context` not being found.

tas50:
The windows functional tests are still broken here.

PR #10037, opened 2020-06-23
Fix rspec warning about `not_to raise_error` with a specific exception.

tas50:
WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives, since literally any other error would cause the expectation to pass, including those raised by Ruby (e.g. NoMethodError, NameError and ArgumentError), meaning the code you are intending to test may not even get reached. Instead consider using `expect { }.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`. This message can be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`. Called from /Users/pete/work/chef/spec/unit/resource_spec.rb:381:in `block (4 levels) in <top (required)>'.


lamont-granquist:
oo yeah, and set it in the spec_helper to make sure it never comes back

tas50:
This is just the backport. Doesn't pass either.

PR #10052, opened 2020-06-23
Fix broken tests after updating diff-lcs dep

phiggins:
https://rubular.com/r/5JdtZDqWnZhB9J


tas50:
@phiggins can you toss that rubular link after the spec so someone can figure out what we're trying to match in the future

phiggins:
@tas50 Done.

PR #10055, opened 2020-06-23
Fix some windows unit tests

lamont-granquist:
There's like 18 different bugs that need fixing in these one by one, no single cause.


lamont-granquist:
other failures look like diff-lcs

PR #10057, opened 2020-06-23
Bump diff-lcs to get bugfix.

phiggins:
This fixes some errors I was seeing, but I'm getting different errors locally.


halostatue:
An update will be released shortly that fixes the issue outlined here.

halostatue:
Released as 1.4.2.

tas50:
Thanks for the quick turnaround @halostatue

phiggins:
Thanks @halostatue!

PR #10058, opened 2020-06-24
Revert openssl digest update that broke FIPS

tas50:
We're going to have to figure out how to work around this before Ruby 3.0 this the constants go away.

### DIFF
--- a/lib/chef/chef_fs/file_system/chef_server/cookbook_file.rb
+++ b/lib/chef/chef_fs/file_system/chef_server/cookbook_file.rb
@@ -18,7 +18,7 @@
 
 require_relative "../base_fs_object"
 require_relative "../../../http/simple"
-require "openssl" unless defined?(OpenSSL)
+require "digest" unless defined?(Digest)
 
 class Chef
   module ChefFS
@@ -69,7 +69,7 @@ class Chef
           private
 
           def calc_checksum(value)
-            OpenSSL::Digest.hexdigest("MD5", value)
+            ::Digest::MD5.hexdigest(value)
           end
         end
       end
--- a/lib/chef/digester.rb
+++ b/lib/chef/digester.rb
@@ -19,6 +19,7 @@
 #
 
 require "openssl" unless defined?(OpenSSL)
+require "digest" unless defined?(Digest)
 require "singleton" unless defined?(Singleton)
 
 class Chef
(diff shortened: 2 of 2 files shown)


lamont-granquist:
SHA512/SHA256 should work, the problem should be in MD5 and maybe SHA1

lamont-granquist:
i think only MD5 is the FIPS non-compliant algorithm that causes issues?

btm:
I think we want something like this over in https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L1230-L1231

```
module FIPSHexdigest
  def hexdigest(hash, value)
    if hash == "MD5"
      ::Digest::MD5.hexdigest(value)
    else
      super(hash, value)
    end
  end
end

module OpenSSL
  class Digest
    extend FIPSHexdigest
  end
end
```

lib/chef/chef_fs/file_system/chef_server/cookbook_file.rb
```diff
@@ -69,7 +69,7 @@ def compare_to(other)
           private
 
           def calc_checksum(value)
-            OpenSSL::Digest.hexdigest("MD5", value)
+            OpenSSL::Digest::MD5.hexdigest(value)
```
lamont-granquist, line 72: can we just call `::Digest::MD5.hexdigest` here instead?

lamont-granquist: (similarly obviously for all the other MD5 calls?)

btm: @jaym any recollection of that choice?

jaym: https://github.com/chef/chef/pull/4481/files#diff-d19a42ae266a2919d7cfeb43c77dbaffR917
I'm guessing its relying on a similar pattern to that. So doing what lamont said would preserve the old behavior

lamont-granquist: Oh I bet that breaks in Ruby 3 as well, but at least being consistent would funnel everything through that one const_set

lamont-granquist:
so to be clear just change the MD5 calls to use `::Digest::MD5.hexdigest` instead of reverting this.

PR #10063, opened 2020-06-24
Update the windows_user_privilege resource to have a `:clear` action

chef-davin:

## Description

It would be useful to have an action in the windows_user_privilege resource that would allow you to easily remove all users from using a Windows user right.  This pull request adds a `:clear` action that does this, by looking for any users assigned a specific user right and then removing them.

It would work like this:
```ruby
windows_user_privilege 'Allow any user the Network Logon right' do
  privilege      'SeDenyNetworkLogonRight'
  action         :clear
end
```
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/windows_user_privilege.rb
+++ b/lib/chef/resource/windows_user_privilege.rb
@@ -112,6 +112,15 @@ class Chef
         action         :remove
       end
       ```
+
+      **Clear all users from the SeDenyNetworkLogonRight Privilege**:
+
+      ```ruby
+      windows_user_privilege 'Allow any user the Network Logon right' do
+        privilege      'SeDenyNetworkLogonRight'
+        action         :clear
+      end
+      ```
       DOC
 
       property :principal, String,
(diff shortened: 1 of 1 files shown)


sam1el:
it would be great to have this functionality. I have been overriding the resource to get it to work. Thanks @chef-davin

lib/chef/resource/windows_user_privilege.rb
```diff
@@ -132,8 +141,8 @@ class WindowsUserPrivilege < Chef::Resource
          }
 
       load_current_value do |new_resource|
-        unless new_resource.principal.nil?
-          privilege Chef::ReservedNames::Win32::Security.get_account_right(new_resource.principal) unless new_resource.action.include?(:set)
+        unless new_resource.principal.nil? || new_resource.action.include?(:set) || new_resource.action.include?(:clear)
```
phiggins, line 144: I know this is not all new code you're adding but I think this would be better with the logic reversed:

```suggestion
        if new_resource.principal && (new_resource.action.include?(:add) || new_resource.action.include?(:remove)
```

chef-davin: @phiggins thanks for the input. suggestion committed.

PR #10070, opened 2020-06-26
fix habitat based windows service tests

mwrock:
The windows service that gets started in the windows service functional tests will not inherit the current shell's environment. So it fails because it's `GEM_PATH` is not set and therefore is unable to consume the `win32-service` gem. This adds the `chef-infra-client` packages `GEM_PATH` to the machine scoped `GEM_PATH`.


btm:
@mwrock these went green, thanks! https://buildkite.com/chef/chef-chef-master-habitat-test

btm:
also confirmed chef/chef-infra-client/16.2.62/20200626032748 and later on windows made it to 'current'.

PR #10085, opened 2020-06-30
Pin dff-lcs, bump ohai, & misc omnibus updates

tas50:
- Pin diff-lcs since 1.4 has changed the output.  
- Bump Ohai and other deps while here. 
- Update some omnibus docs and comments to make diffing with workstation better and the docs a bit more accurate.
- Bring in the .gitkeep cleanup from Workstation to nuke those files in the install

### DIFF
--- a/omnibus/Gemfile.lock
+++ b/omnibus/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/chef/omnibus
-  revision: d75718522deb9faeb3c21b50c60e94daf70ce9b6
+  revision: 320d9350cdf39abdcc4c37d5bc2971b3c222a0a0
   branch: master
   specs:
-    omnibus (7.0.13)
+    omnibus (7.0.14)
       aws-sdk-s3 (~> 1)
       chef-cleanroom (~> 1.0)
       chef-sugar (>= 3.3)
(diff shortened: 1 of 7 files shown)


omnibus/Gemfile.lock
```diff
@@ -164,7 +164,7 @@ GEM
     citrus (3.0.2)
     cleanroom (1.0.0)
     concurrent-ruby (1.1.6)
-    diff-lcs (1.4.2)
+    diff-lcs (1.4.3)
```
phiggins, line 167: 🤔

omnibus/omnibus.rb
```diff
@@ -1,5 +1,5 @@
 #
-# This file is used to configure the Omnibus projects in this repo. It contains
+# This file is used to configure the chef infra client project. It contains
```
phiggins, line 2:
```suggestion
# This file is used to configure the Chef Infra Client project. It contains
```

omnibus/Gemfile.lock
tas50: It's fine in omnibus

lamont-granquist:
(clarification for posterity that diff-lcs 1.4.x actually has broken output through 1.4.3 at least)

PR #10089, opened 2020-06-30
Fix release build tests, part 2

phiggins:
A small thing I missed in #10088.


spec/functional/resource/cron_spec.rb
```diff
@@ -63,8 +63,6 @@ def cron_should_not_exists(cron_name)
   end
 
   let(:new_resource) do
-    run_context = Chef::RunContext.new(Chef::Node.new, {}, Chef::EventDispatch::Dispatcher.new)
```
phiggins, line 66: The one with real platform information is defined directly above this and I forgot to remove the stub.

PR #10097, opened 2020-07-02
Fix syslog logging on Chef-16

### LINKED ISSUE #9982: Fatal TypeError if `log_location` is set to `:syslog` on macOS
## Description
If the `:syslog` symbol is used for `log_location` value in `client.rb`, chef-client fails with the error:
```
FATAL: TypeError: no implicit conversion of Chef::Log::Syslog into String
```

## Chef Version
16.1.16 (not observed on 15.x)

## Platform Version
macOS 10.15.5

## Replication Case
Run `chef-client` with the line `log_location :syslog` in `/etc/chef/client.rb` on macOS. Manually specifying other log paths works as expected. Manually specifying the path _previously_ set by `:syslog` (`log_location '/Library/Logs/Chef/client.log'`) does work, but all chef-client output is logged twice for some reason:

```
Starting Chef Infra Client, version 16.1.16
Starting Chef Infra Client, version 16.1.16
Using policy 'apple_infra_builder' at revision 'f6841d4764c8664f8b4f37a9a548736cb6800c89972b36967687bcb0260ade23'
Using policy 'apple_infra_builder' at revision 'f6841d4764c8664f8b4f37a9a548736cb6800c89972b36967687bcb0260ade23'
```

## Client Output
```
FATAL: TypeError: no implicit conversion of Chef::Log::Syslog into String
```

lamont-granquist:
This forces the formatter output off for users that use the
winevt or syslog options to the log_location.

This changes nothing about the logging behavior from Chef-15, those
never supported formatter output.

It creates some more clear consistency with how STDOUT behaves when
those options are used.

- We do not have the ability (fairly fundamentally in the Chef::Log
  class itself) to have different log levels running at the same time
  so either you get :warn or you get :info and we'll likely never
  be able to support mix-and-match.

- Without mix-and-match if you're trying to use the formatter on
  STDOUT the syslog logger you have to either pick :warn -- which
  makes the syslog output totally useless, or else pick :info --
  which intersperses all of the info logger output with the
  formatter on STDOUT and leads to complaints and bug reports.

- So we force the logger on, although users can still get the old
  STDOUT logging behavior back with --force-formatter -l info

closes #9982

### DIFF
--- a/lib/chef/application.rb
+++ b/lib/chef/application.rb
@@ -194,14 +194,26 @@ class Chef
       chef_config[:log_location].map! do |log_location|
         case log_location
         when :syslog, "syslog"
+          force_force_logger
           logger::Syslog.new
         when :win_evt, "win_evt"
+          force_force_logger
           logger::WinEvt.new
         else
           # should be a path or STDOUT
           log_location
         end
       end
+    end
+
+    # Force the logger by default for the :winevt and :syslog loggers.  Since we do not and cannot
+    # support multiple log levels in a mix-and-match situation with formatters and loggers, and the
+    # formatters do not support syslog, we force the formatter off by default and the log level is
+    # thus info by default.  Users can add `--force-formatter -l info` to get back formatter output
+    # on STDOUT along with syslog logging.
+    #
+    def force_force_logger
+      chef_config[:force_logger] = true unless chef_config[:force_formatter]
     end
 
     # Use of output formatters is assumed if `force_formatter` is set or if `force_logger` is not set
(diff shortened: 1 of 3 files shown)


lib/chef/application.rb
```diff
@@ -204,6 +206,16 @@ def configure_log_location
       end
     end
 
+    # Force the logger by default for the :winevt and :syslog loggers.  Since we do not and cannot
+    # support multiple log levels in a mix-and-match situation with formatters and loggers, and the
+    # formatters do not support syslog, we force the formatter off by default and the log level is
+    # thus info by default.  Users can add `--force-formatter -l info` to get back formatter output
+    # on STDOUT along with syslog logging.
+    #
+    def force_force_logger
```
tas50, line 215: force force?

lamont-granquist: yep

it forces force_logger

PR #10098, opened 2020-07-02
Creating the `inspec_waiver_file_entry` resource for managing and formatting a waiver file

chef-davin:

## Description

Chef recently introduced the use of waiver files by InSpec, as well as use via the audit cookbook.  There isn't a corresponding chef-client resource to manage such a file through chef automation if a user wanted to.  This `inspec_waiver_file_entry` resource will provide that functionality.

**Add an InSpec waiver entry to a given waiver file**:

  ```ruby
    inspec_waiver_file_entry 'Add waiver entry for control' do
      file_path 'C:\\chef\\inspec_waiver.yml'
      control 'my_inspec_control_01'
      run_test false
      justification "The subject of this control is not managed by Chef on the systems in policy group \#{node['policy_group']}"
      expiration '2021-01-01'
      action :add
    end
  ```

**Add an InSpec waiver entry to a given waiver file using the 'name' property to identify the control**:

  ```ruby
    inspec_waiver_file_entry 'my_inspec_control_01' do
      justification "The subject of this control is not managed by Chef on the systems in policy group \#{node['policy_group']}"
      action :add
    end
  ```

**Remove an InSpec waiver entry to a given waiver file**:

  ```ruby
    inspec_waiver_file_entry "my_inspec_control_01" do
      action :remove
    end
  ```

## Related Issue

N/A

## Types of changes

## Checklist:


lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,159 @@
...
+            action :create_if_missing
+          end
+        end
+        converge_if_changed :control, :expiration, :run_test, :justification do
+          yaml_contents = IO.read(new_resource.file)
+          waiver_hash = if yaml_contents.empty?
+                          {}
+                        else
+                          waiver_hash = ::YAML.safe_load(yaml_contents)
+                        end
```
lamont-granquist, line 132: could probably be:

```
waiver_hash = {}
waiver_hash = ::YAML.safe_load(yaml_contents) unless yaml_contents.empty?
```

chef-davin: yeah... that would probably look a little cleaner. I'll update that.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,159 @@
...
+          waiver_hash["#{new_resource.control}"] = {}
+          control_hash = {}
+          control_hash["expiration_date"] = new_resource.expiration.to_s unless new_resource.expiration.nil?
+          control_hash["run"] = new_resource.run_test unless new_resource.run_test.nil?
+          control_hash["justification"] = new_resource.justification.to_s
+          waiver_hash["#{new_resource.control}"] = control_hash
+          waiver_hash = waiver_hash.sort.to_h
+          ::File.open(new_resource.file, "w") { |f| f.puts waiver_hash.to_yaml }
```
lamont-granquist, line 140: this should be part of the `file` subresource used above but pushed into the `content` property of it.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+
+      action :add do
+        converge_if_changed :file do
+          file "Create Waiver File #{new_resource.file}" do
+            path new_resource.file
+            action :create_if_missing
+          end
+        end
+        converge_if_changed :control, :expiration, :run_test, :justification do
```
lamont-granquist, line 126: it is probably better in this case to not try to use `converge_if_changed` but just use the idempotency of the file resource.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,159 @@
...
+          else
+            control ""
+          end
+        else
+          file ""
+          control ""
+          justification ""
+          expiration "1111-11-11"
+        end
+      end
```
lamont-granquist, line 117: if you just let the contents of the file subresource dictate idempotency then this whole method goes away.

danielcbright:
The example provided needed the `expiration` to be encapsulated with single quotes in order to not throw an error: `expiration '2021-01-01'`. Also, I've been using my own recipe for a few days now that takes a different approach, it uses a template and allows you to control the entire waiver file with attributes dynamically. Just wanted to bring it up as another possible route to go down, or maybe combine the dynamic nature of a template + `.each` loop for attributes with a nice resource like this that takes an attribute hash and iterates over it (kinda like how the profiles attribute is set in the audit cookbook).

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+require_relative "../dist"
+require "yaml" unless defined?(YAML)
+
+class Chef
+  class Resource
+    class InspecWaiverFile < Chef::Resource
+      provides :inspec_waiver_file
+      unified_mode true
+
+      description "Use the **inspec_waiver_file** resource to add or remove entries from an inspec waiver file. This can be used in conjunction with the audit cookbook"
```
phiggins, line 27:
```suggestion
      description "Use the **inspec_waiver_file** resource to add or remove entries from an inspec waiver file. This can be used in conjunction with the audit cookbook."
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+      property :control, String,
+        name_property: true,
+        description: "The name of the control being added or removed to the waiver file"
+
+      property :file, String,
+        required: true,
+        description: "The path to the waiver file being modified"
+
+      property :expiration, String,
+        description: "The expiration date of the given waiver - provided in YYYY-MM-DD format",
```
phiggins, line 72: Using a regex to validate the form of the string makes sense, but instead of validating the content (eg, that the third stanza can be 30 or 31 but not 32), this could rely on Date parsing:

```
irb(main):001:0> require 'date'
=> true
irb(main):002:0> s = "2020-01-31"
=> "2020-01-31"
irb(main):003:0> Date.valid_date?(*s.split("-").map(&:to_i))
=> true
irb(main):004:0> s = "2020-02-31"
=> "2020-02-31"
irb(main):005:0> Date.valid_date?(*s.split("-").map(&:to_i))
=> false
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+          },
+          }
+
+      property :run_test, [true, false],
+        description: "If present and true, the control will run and be reported, but failures in it won’t make the overall run fail. If absent or false, the control will not be run."
+
+      property :justification, String,
+        description: "Can be any text you want and might include a reason for the waiver as well as who signed off on the waiver."
+
+      property :backup, [false, Integers],
```
phiggins, line 90:
```suggestion
      property :backup, [false, Integer],
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+                          ""
+                        end
+        waiver_hash = {}
+        waiver_hash = ::YAML.safe_load(yaml_contents) unless yaml_contents.empty?
+        control_hash = {}
+        control_hash["expiration_date"] = new_resource.expiration.to_s unless new_resource.expiration.nil?
+        control_hash["run"] = new_resource.run_test unless new_resource.run_test.nil?
+        control_hash["justification"] = new_resource.justification.to_s
+
+        if waiver_hash.key?("#{new_resource.control}")
```
phiggins, line 108:
```suggestion
        if waiver_hash.key?(new_resource.control)
```

Since this is a string, the interpolation is unnecessary.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+          unless waiver_hash["#{new_resource.control}"] == control_hash
+            waiver_hash["#{new_resource.control}"] = {}
```
phiggins, line 110:
```suggestion
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+            waiver_hash = waiver_hash.sort.to_h
+            file "Update Waiver File #{new_resource.file} to update waiver for control #{new_resource.control}" do
+              path new_resource.file
+              content waiver_hash.to_yaml
+              backup new_resource.backup
+              action :create
+            end
+          end
+        else
+          waiver_hash["#{new_resource.control}"] = {}
```
phiggins, line 121:
```suggestion
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,152 @@
...
+          waiver_hash["#{new_resource.control}"] = control_hash
+          waiver_hash = waiver_hash.sort.to_h
+          file "Update Waiver File #{new_resource.file} to add waiver for control #{new_resource.control}" do
+            path new_resource.file
+            content waiver_hash.to_yaml
+            backup new_resource.backup
+            action :create
+          end
+        end
```
phiggins, line 130: It looks like the two branches of this conditional are the same except one uses `add` and one uses `update` in the description. Was that intentional? If so, there's probably a more succinct way to write this without repetition.

spec/unit/resource/inspec_waiver_file_spec.rb
```diff
@@ -0,0 +1,77 @@
+#
+# Author:: Cary Penniman (<cary@rightscale.com>)
+# Author:: Tyler Cloke (<tyler@chef.io>)
```
phiggins, line 3: This is probably not correct. :)

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,153 @@
...
+        callbacks: {
+          "Expiration date should be a valid calendar date and match the following format: YYYY-MM-DD" => proc { |e|
+            re = Regexp.new('\d{4}-\d{2}-\d{2}$').freeze
+            if re.match?(e)
+              Date.valid_date?(*e.split("-").map(&:to_i))
+            elsif e.nil?
+              true
+            else
+              false
+            end
```
phiggins, line 83:
```suggestion
            else
              e.nil?
            end
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,153 @@
...
+        control_hash["justification"] = new_resource.justification.to_s
+
+        if waiver_hash.key?(new_resource.control)
+          unless waiver_hash[new_resource.control] == control_hash
+            waiver_hash.delete(new_resource.control)
+            waiver_hash[new_resource.control] = control_hash
+          end
+        else
+          waiver_hash[new_resource.control] = control_hash
+        end
```
phiggins, line 112:
```suggestion
        waiver_hash[new_resource.control] = control_hash
```

If this code is making sure that `waiver_hash` has the new `control_hash` set for the key `new_resource.control`, can it just always set that value and avoid all of this logic? I walked through this a few times and it seems like this should be equivalent.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,153 @@
...
+        def load_waiver_file_to_hash(file_name)
+          if ::File.file?(file_name) && ::File.readable?(file_name) && !::File.zero?(file_name)
+            file_contents = IO.read(file_name)
+            contents_hash = {}
+            contents_hash = ::YAML.safe_load(file_contents) unless file_contents.empty?
+          else
+            contents_hash = {}
+          end
+          contents_hash
+        end
```
phiggins, line 147:
```suggestion
        def load_waiver_file_to_hash(file_name)
          if ::File.exist?
            ::YAML.load_file(file_name)
          else
            {}
          end
        end
```
If the file is not readable, etc, it seems like we'd want this to fail to let the user know something went with the path they supplied. This could be confusing if I'm expecting this resource to read my file and it comes back blank with no indication of what happened.

lib/chef/resource/inspec_waiver_file.rb
chef-davin: so, it was validating that there were no changes to the expiration, run_tests, or justification of a control and then not bothering to run the file resource.  but I guess we could always just wipe out the control from the hash, recreate it and then let the file resource determine if there's a diff or not...

phiggins: I was curious if that's what was going on, and in that case you could wrap the file resource in a `if waiver_hash[new_resource_control] == control_hash`.

lib/chef/resource/inspec_waiver_file.rb
chef-davin: that's a good point. I'll put this into my updates.

lamont-granquist:
`expeditor/cleanup_orphaned_test_resources.sh` seems unrelated and uncalled by anything...

other than that, this looks fine though.

chef-davin:
> `expeditor/cleanup_orphaned_test_resources.sh` seems unrelated and uncalled by anything...
> 
> other than that, this looks fine though.

Not sure how that got in there, but it's been removed.

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,151 @@
...
+require "yaml"
+require "date"
+
+class Chef
+  class Resource
+    class InspecWaiverFile < Chef::Resource
+      provides :inspec_waiver_file
+      unified_mode true
+
+      description "Use the **inspec_waiver_file** resource to add or remove entries from an inspec waiver file. This can be used in conjunction with the audit cookbook."
```
tas50, line 28:
```suggestion
      description "Use the **inspec_waiver_file** resource to add or remove entries from an inspec waiver file. This can be used in conjunction with Compliance Phase."
```

lib/chef/resource/inspec_waiver_file.rb
```diff
@@ -0,0 +1,151 @@
...
+      introduced "16.6"
```
tas50, line 29:
```suggestion
      introduced "17.1"
```

tas50:
From our Slack conversation on this:

- rename to inspec_waiver_file_entry
- rename file to file_path and make it default to /etc/chef/inspec_waivers.yml and C:\chef\inspec_waivers.yml

Issue #10073, opened 2020-06-26
Default configuration of `ssl_ca_file` is only set on Windows, but not on any other platforms


kapilchouhan99:
## Description
Default configuration of `ssl_ca_file` is only set on Windows, but not on any other platforms
## Related Issue
https://github.com/chef/customer-bugs/issues/102
## Replication Case
Execute the recipe locally on any non-Windows platform. Contents of /tmp/env.log will be empty after the run completes.
```
file '/tmp/env.log' do
  content "#{Chef::Config['ssl_ca_file']}" 
end
```

## Expected Results
Both Windows and non-Windows will return a valid path, when executing the above recipe.

## Actual Results:
The above recipe when executed on a Windows node, It returns `C:/opscode/chef/embedded/ssl/certs/cacert.pem`
When executed on Linux it returns empty (nil)

dheerajd-msys:
Closing based on https://github.com/chef/chef/pull/10100#discussion_r482012014

PR #10100, opened 2020-07-02
Fix for ssl_ca_file is only set on Windows, but not on any other platforms.

kapilchouhan99:

## Description
- Now Default configuration of `ssl_ca_file` will set for both types of platforms Windows and Unix.
- Added test cases.
- Ensured chef-style on the code changes made.

## Related Issue
Fixes: #10073 

## Types of changes

## Checklist:

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -634,7 +634,7 @@ module ChefConfig
     # Path to the default CA bundle files.
     default :ssl_ca_path, nil
     default(:ssl_ca_file) do
-      if ChefUtils.windows? && embedded_dir
+      if embedded_dir
         cacert_path = File.join(embedded_dir, "ssl/certs/cacert.pem")
         cacert_path if File.exist?(cacert_path)
       else
--- a/chef-config/spec/unit/config_spec.rb
+++ b/chef-config/spec/unit/config_spec.rb
@@ -644,12 +644,21 @@ RSpec.describe ChefConfig::Config do
 
         end
 
-        # On Windows, we'll detect an omnibus build and set this to the
-        # cacert.pem included in the package, but it's nil if you're on Windows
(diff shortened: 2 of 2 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -634,7 +634,7 @@ def self.target_mode?
     # Path to the default CA bundle files.
     default :ssl_ca_path, nil
     default(:ssl_ca_file) do
-      if ChefUtils.windows? && embedded_dir
+      if embedded_dir
```
lamont-granquist, line 637: so why wasn't this done on windows to begin with?  where did this code come from and is there any rationale in the original PR?  and what effect is this going to have on unix users?

kapilchouhan99: This code was added https://github.com/chef/chef/commit/eecb3f3cdbe1eb4e77794f33340a6690e49b9d34

In Unix/Linux machine the certificate is present under `/opt/chef/embedded/ssl/certs/cacert.pem`

So earlier it was set default configuration of `ssl_ca_file` only for Windows, and for Unix/Linux machine it was set as nil.

kapilchouhan99: @mwrock @lamont-granquist Can we close it or anything we can do if required, please put your suggestion so that i can proceed.

mwrock: My question is what is this fixing. I know it aims to "fix" #10073 but I'm not sure that is a bug but just an implementation detail. Do we have customer scenarios where 10073 actually break chef runs on non-windows platforms? If not, I think we should close.

kapilchouhan99: @mwrock @lamont-granquist @tas50 since we are not getting any chef failure so I think we can close PR.

Issue #10105, opened 2020-07-02
Resource override warning indicates incorrect behavior


kimbernator:
## Description
Upon startup, chef-client indicates the following:

```
WARN: Resource chef_client_scheduled_task from the client is overriding the resource from a cookbook. Please upgrade your cookbook or remove the cookbook from your run_list.
```

This is due to the `chef-client` cookbook still being present in the run-list (this cookbook contains a custom resource called `chef_client_scheduled_task`) while running chef 16, which now includes `chef_client_scheduled_task` natively. Since there is further functionality in that cookbook that didn't make it into chef 16 (namely client.rb management), the cookbook remains in the runlist.

The warning above implies to me that the native version of the resource takes precedence and the custom resource in the chef-client cookbook will be disregarded. However, after noticing the splay not being applied to the scheduled task, I found that the chef-client cookbook resource is indeed still the one being used. I put a few simple "puts" statements in each and ran chef, it indicated that I was on the cookbook version.

## Chef Version
16.1.16

## Platform Version
Windows Server 2016

## Replication Case

Have a cookbook with a dependency on the chef-client cookbook, edit the local copy of the scheduled_task resource's `:add` action to include a "puts" indicating that it is being executed, then attempt to execute a chef_client_scheduled_task resource and look for the puts output. I'm sure there's a more "correct" way to do this but it works.

## Client Output

None aside from the warning above.

lamont-granquist:
fixed the warning message in #10117

PR #10106, opened 2020-07-02
Fix resource mapping to prevent undesired override

kimbernator:
## Description

resolves an issue where custom resources with the same name as native ones successfully overwrite native ones, despite a warning showing that indicates the opposite.

## Related Issue

https://github.com/chef/chef/issues/10105

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/node_map.rb
+++ b/lib/chef/node_map.rb
@@ -84,6 +84,7 @@ class Chef
                           klass.superclass.to_s
                         end
         Chef::Log.warn( COLLISION_WARNING % { type: type_of_thing, key: key, type_caps: type_of_thing.capitalize } )
+        return map
       end
 
       # The map is sorted in order of preference already; we just need to find


kimbernator:
okay, so based on some of the errors in the unit testing I feel like the tests are looking for what I feel is the "incorrect" behavior, but I'm not really sure if that is an issue with the tests or a misunderstanding on my part.

lamont-granquist:
Pretty sure we changed the behavior, and the behavior is now considered correct, the unit tests are correct, the message is wrong though and didn't get updated when the behavior changed to how it currently is.

By default we don't want core resources overriding cookbook resources because then upgrades to new point releases of chef-client break people cookbook resources they've been happily using.  We use other mechanisms to force core to start overriding cookbooks on major version releases.

So this PR can't be merged, behavior cannot change.

kimbernator:
@lamont-granquist That makes sense. After looking through the commit history, I did start to find some indication that this is desired. 

However, this does appear to mean that my scenario in the linked issue has no workaround. The chef-client cookbook hasn't actually released some updates I made to the custom `chef_client_scheduled_task` resource, instead opting to include them in the core. Is there any way to specify that I want to use the core resource over the custom one?

tas50:
@kimbernator We generally adds provides lines like this to resources in cookbooks to prevent them loading on newer Chef Infra Client releases that ship with resources built in:

https://github.com/chef-boneyard/ohai/blob/master/resources/hint.rb#L1

tas50:
Closing this out since we updated the error message.

Issue #10095, opened 2020-07-01
macos_userdefaults cannot pass a string with a space in it


erikng:
## Description
When attempting to pass a string with a space, the value is not passed with quotes, causing the command to fail.

## Chef Version
16.2.50

## Platform Version
macOS 10.15.5 +

## Replication Case
```
macos_userdefaults 'Bad string test' do
  domain '/Library/Preferences/ManagedInstalls'
  key 'LogFile'
  value '/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'
end
```

## Client Output
```
[2020-07-01T06:44:43-07:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2020-07-01T06:44:43-07:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2020-07-01T06:44:43-07:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: cpe_munki_config[Manage Munki Settings] (cpe_munki::default line 18) had an error: Mixlib::ShellOut::ShellCommandFailed: macos_userdefaults[Configure ManagedInstalls - LogFile] (/var/chef/cache/cookbooks/cpe_munki/resources/cpe_munki_config.rb line 60) had an error: Mixlib::ShellOut::ShellCommandFailed: execute[defaults write '/Library/Preferences/ManagedInstalls' 'LogFile' /Library/Managed Installs/Logs/ManagedSoftwareUpdate.log] (/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.2.50/lib/chef/resource/macos_userdefaults.rb line 117) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
---- Begin output of defaults write '/Library/Preferences/ManagedInstalls' 'LogFile' /Library/Managed Installs/Logs/ManagedSoftwareUpdate.log ----
STDOUT: 
STDERR: 2020-07-01 06:44:41.358 defaults[6967:119919] Unexpected argument Installs/Logs/ManagedSoftwareUpdate.log; leaving defaults unchanged.
---- End output of defaults write '/Library/Preferences/ManagedInstalls' 'LogFile' /Library/Managed Installs/Logs/ManagedSoftwareUpdate.log ----
Ran defaults write '/Library/Preferences/ManagedInstalls' 'LogFile' /Library/Managed Installs/Logs/ManagedSoftwareUpdate.log returned 1
```

[...]

Issue #10084, opened 2020-06-30
macos_userdefaults cannot manage an array value idempotently.


natewalck:
## Description
When setting an array as a value for a key using `macos_userdefaults`, chef will set the array every chef run. 

## Chef Version
16.2.50

## Platform Version
10.16 (but probably <= 10.15 as well)

## Replication Case
Attempt to set an array as a value for a key using `macos_userdefaults`

## Client Output
Nothing relevant. Run chef 2-3 times, it will set the key every run as the idempotence check isn't working correctly.

erikng:
putting some debug prints in the macos_userdefaults

```
defaults read '/Library/Preferences/ManagedInstalls' 'AdditionalHttpHeaders' | grep -qx '["Authorization: Basic basdadfsdfwsdfswdfwerf=="]'
      * execute[defaults write '/Library/Preferences/ManagedInstalls' 'AdditionalHttpHeaders' -array 'Authorization: Basic basdadfsdfwsdfswdfwerf=='] action run
```

the grep is wrong, so it returns nothing.

chef is running
```
defaults read '/Library/Preferences/ManagedInstalls' 'AdditionalHttpHeaders' | grep -qx '["Authorization: Basic basdadfsdfwsdfswdfwerf=="]' 

$?
zsh: command not found: 1
```

```
defaults read '/Library/Preferences/ManagedInstalls' 'AdditionalHttpHeaders'                                                                                           
(
    "Authorization: Basic basdadfsdfwsdfswdfwerf=="
)
```

This is going to be very ugly to unpack (and why using `CFPreferences` API over `defaults read` is better) but one potential solution (that likely will not work with more advanced arrays):

```
defaults read '/Library/Preferences/ManagedInstalls' 'AdditionalHttpHeaders' | grep -qx '    "Authorization: Basic basdadfsdfwsdfswdfwerf=="'
$?
zsh: command not found: 0
```

The command above takes off the `[]` since that's not a format that defaults read uses and adds four `spaces`. This then allows the line to return correctly.

Issue #7622, opened 2018-09-06
macos_userdefaults shouldn't track state in is_set property


tas50:
We're using a property to track if a resource is set or not. We should do this with current_resource instead so a user doesn't set this by mistake and break the resource.

Issue #10086, opened 2020-06-30
potential fix for macos_userdefaults and complex values


erikng:
### Describe the Enhancement:
The following descriptions may be able to make this resource work for complex values

#### Reading a defaults file
```
/usr/bin/defaults export payload_domain -
```

The key is the `-` at the very end.

```
defaults export /Library/Preferences/com.apple.timezone.auto.plist -

<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>Active</key>
	<true/>
</dict>
</plist>
```

This would then allow you to consume this with the `plist` gem and properly compare values.

#### Writing a defaults file
```
defaults import payload_domain - < /path/to/file.plist
```

Using this, you would have to do the following:
1. Read the current configuration with `defaults import` and save it in memory
2. Append the desired changes to the plist, overwriting what is necessary
3. Saving the plist to disk
4. importing it with the above example command

```
defaults import /Library/Preferences/com.apple.timezone.auto.plist - < /tmp/tz.plist
(plist file has Active set to false)

defaults export /Library/Preferences/com.apple.timezone.auto.plist -

<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>Active</key>
	<false/>
</dict>
</plist>
```

One secondary concern is spawning multiple `import` commands simultaneously (say setting 10 values to one preference domain). There is no guarantee that `cfprefsd` will write to disk quickly enough, which would not be run-time safe.

One alternative would be for chef to look at all preferences being set for a specific domain and hashing the written plist with the expected plist. This is exactly what @natewalck did when implementing the `osx_profile` resource.

### Describe the Need:
There are several bugs
- https://github.com/chef/chef/issues/7131
 - this should be fixed with the `defaults export` command
- https://github.com/chef/chef/issues/10083
  - this should be fixed with the `defaults import` command
- https://github.com/chef/chef/issues/10084
  - this should be fixed with the `defaults export` command

I think these may all be solved with this

### Current Alternative
N/a

### Can We Help You Implement This?:
Not at this moment

erikng:
[...]

PR #10118, opened 2020-07-06
Rework macos_userdefaults resource

tas50:
This is a reboot of sorts on the macos_userdefaults resource to fix some fundamental flaws in the UX of the resource and how it gets state / sets the values.

Fixes https://github.com/chef/chef/issues/7622
Fixes https://github.com/chef/chef/issues/10095
Fixes https://github.com/chef/chef/issues/10084

**The good:**

- Uses the array format of values for `shell_out`, which avoids the previous issues with spaces in keys/values
- Rewrites how we load state to use `defaults export`, which allows us to parse complex data values accurately. The previous method of determining the current value just used grep for a string, which would fail on anything but the most basic value
- Overall this fixes a bunch of small failures to set values correctly. Things just didn't work throughout and unit testing exposed much of that.
- By properly loading state we now get much better convergence logging that shows before and after state. 
- Add a new `:remove` action for removing keys entirely. This seemed like a useful thing that was missing
- Replaced the 'global' property with a default value in the `domain` property instead. Previously it was meant that you could set the global property and skip domain to get `NSGlobalDomain` as your domain. Seems a bit overly complex to have two properties that control the domain, but you know whatever. Unfortunately, this didn't work because in `load_current_value` we still tried to load from the domain property and this would result in a domain of '' which would blow up the defaults command. So it's really questionable if this is a breaking change since it didn't work before. Now if you skip the domain property you just get the default of `NSGlobalDomain`. I'm fairly confident that no one ever used this, but I can create a Cookstyle rule if we want to be extra sure.
- Added specs for the `defaults export` and `defaults write` commands to validate how we build the commands for getting and setting state
- Added a better description to sudo to explain the requirements of setting this. 
- Added equal_to to the type property since it was entirely undocumented what values were allowed here. Since we don't support every defaults type you had to just read to code to know how to use it. 
- Added a new `host` property that lets you set per-host values. If you set this to :current it sets the `-currentHost` flag.
- Set desired_state to false on everything that isn't a state property
- Removes the `is_set` property which was being used to store the state

**The bad:**

This updated resource does a much better job of handling complex data types, but it's still deeply flawed since it relies on the `defaults` command, which can't set complex nested values. We could push it further, but we're close to the limits.

**Things I'm going to tackle in subsequent PRs:**

- booleans that are specified as strings with the type set to bool are not idempotent
- we need more examples


lib/chef/resource/macos_userdefaults.rb
```diff
@@ -28,40 +28,74 @@ class MacosUserDefaults < Chef::Resource
...
 
       property :key, String,
-        description: "The preference key."
+        description: "The preference key.",
+        required: true,
+        desired_state: false
 
       property :value, [Integer, Float, String, TrueClass, FalseClass, Hash, Array],
-        description: "The value of the key.",
+        description: "The value of the key. Note: When setting boolean values you can either specify 0/1 or you can pass true/false, 'true'/false', or 'yes'/'no' and we'll automattically convert these to the proper boolean values Apple expects.",
```
phiggins, line 81:
```suggestion
        description: "The value of the key. Note: When setting boolean values you can either specify 0/1 or you can pass true/false, 'true'/false', or 'yes'/'no' and we'll automatically convert these to the proper boolean values Apple expects.",
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,34 +107,38 @@ def coerce_booleans(val)
...
-
-        is_set !vc.error?
+        coerced_value = coerce_booleans(desired.value)
+
+        state_cmd = ['/usr/bin/defaults', 'read', desired.domain, desired.key]
+
+        state = if desired.user.nil?
+          shell_out(state_cmd)
+        else
+          shell_out(cmd, user: desired.user)
```
phiggins, line 117:
```suggestion
          shell_out(state_cmd, user: desired.user)
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,34 +107,38 @@ def coerce_booleans(val)
...
+        end
+
+        current_value_does_not_exist! if state.error?
```
phiggins, line 120: This should probably dump the error to the debug log or something in case of an unexpected error.

lib/chef/resource/macos_userdefaults.rb
tas50: Fixed

lib/chef/resource/macos_userdefaults.rb
tas50: Added debug logging

tas50:
Fixes #10095

tas50:
Fixes #10084

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +107,90 @@ def coerce_booleans(val)
...
-        vc = if desired.user.nil?
-               shell_out(cmd)
-             else
-               shell_out(cmd, user: desired.user)
-             end
-
-        is_set !vc.error?
+        coerced_value = coerce_booleans(desired.value)
+
+        state_cmd = ["/usr/bin/defaults", "read", desired.domain, desired.key]
```
jaymzh, line 112: the fact that key is required is new and backwards incompat... do we need to keep back-compat for that until 17?

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +107,90 @@ def coerce_booleans(val)
...
-          if type == "array"
-            value = value.join("' '")
-            value = "'#{value}'"
-          end
+          type = new_resource.type || value_type(value)
+
+          value = case type
+                  when "dict"
+                    # creates an array of Key1 Value1 Key2 Value2...
+                    value.flatten.map { |x| "'#{x}'" }
```
jaymzh, line 180: This comment isn't quite true, it takes each one and makes a string with single quote around it. That comment would be more accurate saying `creates and array of ["'key1'", "'key2'"]` 

```
irb(main):003:0> d = {'ast' => 'dink'}
irb(main):004:0> d.flatten.map { |x| "'#{x}'" }
=> ["'ast'", "'dink'"]
```

Did you instead want `.to_a.flatten`?

```
irb(main):006:0> d.to_a.flatten
=> ["ast", "dink"]
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +107,90 @@ def coerce_booleans(val)
...
+                  when "array"
+                    value.map { |x| "'#{x}'" }
+                  else
+                    "'#{value}'"
+                  end
+
+          cmd = ["defaults", "write", "'#{new_resource.domain}'", "'#{new_resource.key}'"]
```
jaymzh, line 187: the single-quotes shouldn't be needed there if you use the command as an array:

```
[phil@rider ~]$ cat /tmp/test.sh
#!/bin/bash

echo "\$1 is $1"
[phil@rider ~]$ /tmp/test.sh foo bar
$1 is foo
```
and:
```
irb(main):010:0> cmd = ['/tmp/test.sh', 'foo bar', 'baz']
irb(main):011:0> puts Mixlib::ShellOut.new(cmd).run_command.stdout
$1 is foo bar
```

erikng:
I think continuing to coerce the values is not ideal, because it's not truly desired state if you can cheat the system.

Doing something like https://github.com/chef/chef/issues/10086 seems more appropriate here so you can truly compare the real values.

lib/chef/resource/macos_userdefaults.rb
tas50: This resource had some pretty fundamental design flaws that require breaking it to fix.  The previous behavior where it wasn't required seems to 100% broken. You would only not provide the key if you were setting the global property to true and in that case we used key values in the domain property (wtf) and didn't specify a domain. It doesn't look like that ever worked. I tried to repro the examples and just got errors so I think anyone who ever tried to skip that property just had chef blow up in their face.

tas50:
@erikng Can you elaborate on "cheat the system". I'll freely admit that this is less than ideal, but I think it potentially gives us the performance we need while determining state (if done right). I'm not super opposed to exporting the data to get the current state, but I'd need to see various perf benchmarks there to show that we're not going to take a significant hit by adding that file io. Even if this method gets tossed in a few months I think the other bits of refactoring here are valuable to continue and certainly worth shipping as soon as possible.

lib/chef/resource/macos_userdefaults.rb
tas50: Fair point. Updated

tas50: The values need to be quoted as they go out to the command so that they don't fail if they have a space in them. That's one of the reported bugs that was fixed here and the test case added for our azure pipelines job.

lib/chef/resource/macos_userdefaults.rb
tas50: We need the quotes to supports keys with a space. You can pass whatever you want into defaults as long as you single quote it on the CLI:

If I try to chef-apply something like this with a space in the key:

```ruby
macos_userdefaults "value with space" do
  domain "/Library/Preferences/ManagedInstalls"
  key "Log File"
  value "/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log"
end
```

With the quoting code removed:

```
❰tsmith❙~/dev/work/chef(git≠macos_userdefaults)❱✔≻ bundle exec chef-apply test.rb
  * macos_userdefaults[value with space] action write

    Error executing action `write` on resource 'macos_userdefaults[value with space]'

    Expected process to exit with [0], but received '1'
    ---- Begin output of ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] ----
    STDOUT:
    STDERR: 2020-07-14 21:31:09.951 defaults[55022:8620703] Could not write domain /Library/Preferences/ManagedInstalls; exiting
    ---- End output of ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] ----
    Ran ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] returned 1

    Resource Declaration:
    ---------------------
    # In test.rb

      1: macos_userdefaults "value with space" do
      2:   domain "/Library/Preferences/ManagedInstalls"
      3:   key "Log File"
      4:   value "/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log"
      5: end

    Compiled Resource:
    ------------------
    # Declared in test.rb:1:in `run_chef_recipe'

    macos_userdefaults("value with space") do
      action [:write]
      declared_type :macos_userdefaults
      domain "/Library/Preferences/ManagedInstalls"
      key "Log File"
      value "/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log"
    end

    System Info:
    ------------
    chef_version=16.3.8
    platform=mac_os_x
    platform_version=10.15.6
    ruby=ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]
    program_name=/usr/local/lib/ruby/gems/2.7.0/bin/chef-apply
    executable=/usr/local/Cellar/ruby/2.7.1_2/bin/bundle

[2020-07-14T21:31:09-07:00] FATAL: Stacktrace dumped to /Users/tsmith/.chef/cache/chef-stacktrace.out
[2020-07-14T21:31:09-07:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2020-07-14T21:31:09-07:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: macos_userdefaults[value with space] ((chef-apply cookbook)::(chef-apply recipe) line 1) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
---- Begin output of ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] ----
STDOUT:
STDERR: 2020-07-14 21:31:09.951 defaults[55022:8620703] Could not write domain /Library/Preferences/ManagedInstalls; exiting
---- End output of ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] ----
Ran ["defaults", "write", "/Library/Preferences/ManagedInstalls", "Log File", "'/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log'"] returned 1
```

With the code as it is in this branch:

```
❰tsmith❙~/dev/work/chef(git✱≠macos_userdefaults)❱✘≻ bundle exec chef-apply test.rb
  * macos_userdefaults[value with space] action write
    - update value with space
    -   set value to "/Library/Managed Installs/Logs/ManagedSoftwareUpdate.log" (was {"Identifier"=>"HUBaseColorLightBlue"})
```

jaymzh: dafuq. that's... not how `exec()` is supposed to work. Goddamnit Apple, what have you done.

tas50: We're trying to run this kind of thing on the CLI:

```
defaults write '/Library/Preferences/ManagedInstalls' 'Log File' -dict 'Identifier' 'HUBaseColorLightBlue'
```

jaymzh: yes, but it shouldn't, unless a shell is interpreting it. When you use Mixlib::ShellOut with an array, it's calling `exec()` and the second argument is an array and the array has each argument as a single element in it - regardless of spaces. To put that in ruby it'd be `exec('defaults', ['write', '/Library/Preferences/ManagedInstalls' 'Log File', ...])` - in which case there's nothing to misunderstand that space.

The only tine that should be a problem is if it's being passed to a shell first, which it shouldn't be if Mixlib::ShellOut is called with an array instead of a string, which you are doing.

I can't tell if this is some weird bug in Mixlib on Mac, or some other crazy thing Mac is doing. Unless I'm misunderstanding something, which is always possible. @lamont-granquist might have thoughts here... am I crazy?

erikng:
@tas50 the current method will suffer from the same issue I've complained about since Chef 14.

Let's say I want to make an administrator _think_ I have automatic updates turned on. Instead of passing a boolean value of `True` to  the `AutomaticallyInstallMacOSUpdates` key, I will instead pass an integer of `1`. As coded (and previously coded) chef will see this value, and happily move on.

To me, this fundamentally means the DSC model for this resource is broken. Given that administrators may depend on this resource for critical security functionality, I think it would be prudent to ensure it properly compares the values, even at the expense of resource speed.

tas50:
@erikng I ran into the truthy issues today which led me to look at the perf of the export. It's actually not bad so I'm going to see what it takes to at least load current value with that. More changes incoming here.

lib/chef/resource/macos_userdefaults.rb
tas50: Everything is converted to array format and the single quotes nuked. Value was the thing holding it up.

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
+      #
+      # @return [Array] defaults command
+      #
+      def defaults_export_cmd(resource)
+        state_cmd = ["/usr/bin/defaults"]
+
+        if resource.host == "current"
+          state_cmd << "-currentHost"
+        elsif resource.host # they specified a non-nil value, which is a hostname
+          state_cmd << ["-host", resource.host]
```
jaymzh, line 131:
```suggestion
          state_cmd += ["-host", resource.host]
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
+        end
+
+        state_cmd << ["export", resource.domain, "-"]
```
jaymzh, line 134:
```suggestion
        state_cmd += ["export", resource.domain, "-"]
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
+        state_cmd.flatten
```
jaymzh, line 135:
```suggestion
        state_cmd
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
       end
 
       action_class do
+        def defaults_modify_cmd(defaults_action)
+          cmd = ["defaults"]
+
+          if new_resource.host == "current"
+            state_cmd << "-currentHost"
+          elsif new_resource.host # they specified a non-nil value, which is a hostname
+            state_cmd << ["-host", new_resource.host]
```
jaymzh, line 178:
```suggestion
            state_cmd += ["-host", new_resource.host]
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
+          end
+
+          cmd << [defaults_action, new_resource.domain, new_resource.key]
```
jaymzh, line 181:
```suggestion
          cmd += [defaults_action, new_resource.domain, new_resource.key]
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -73,54 +111,96 @@ def coerce_booleans(val)
...
+          cmd << processed_value if defaults_action == "write"
+          cmd.prepend("sudo") if new_resource.sudo
+          cmd.flatten
```
jaymzh, line 184:
```suggestion
          cmd
```

lib/chef/resource/macos_userdefaults.rb
tas50: Isn't += is going to create a new object each time I add to the array vs. flatten doing it once?

tas50: I switched to concat which avoids the object and avoids the flatten

jaymzh: `+=` afaik, is the same thing as `concat` - it's the operator that calls `concat` underneath. Just like `<<` called `append`.

tas50:
```
irb(main):039:0> foo = []
irb(main):040:0> foo.object_id
=> 220
irb(main):041:0> foo =+ 'something'
irb(main):042:0> foo.object_id
=> 240
```

```
irb(main):044:0> foo = []
irb(main):045:0> foo.object_id
=> 260
irb(main):046:0> foo.concat(['bar'])
=> ["bar"]
irb(main):047:0> foo.object_id
=> 260
```

jaymzh: That... code doesn't work:
```
irb#1(main):008:0> foo = []
irb#1(main):009:0> foo += 'stuff'
        2: from (irb#1):9
        1: from (irb#1):9:in `+'
TypeError (no implicit conversion of String into Array)
irb#1(main):010:0> 
```
But you're right, if you `+=` an array it makes a new object, that's strange:
```
irb#1(main):017:0> foo.object_id
=> 13540
irb#1(main):018:0> foo += ['stuff']
irb#1(main):019:0> foo.object_id
=> 13560
```

Is that more expensive than flatten?

Either way, I like `concat` as a solution. Best of both worlds.

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -28,99 +30,180 @@ class MacosUserDefaults < Chef::Resource
...
+        # convert the provided value into the format defaults expects
+        #
+        # @return [array] array of values starting with the type if applicable
+        #
+        def processed_value
+          type = new_resource.type || value_type(value)
+
+          # when dict this creates an array of values ["Key1", "Value1", "Key2", "Value2" ...]
+          cmd_vals = [ type == "dict" ? new_resource.value.flatten : new_resource.value ]
+          cmd_vals.prepend("-#{type}") if type
```
phiggins, line 197: Spellcheck does not like this variable name:

```
/workdir/lib/chef/resource/macos_userdefaults.rb:196:15 - Unknown word (vals)
/workdir/lib/chef/resource/macos_userdefaults.rb:197:15 - Unknown word (vals)
```

phiggins: Also if `type` is nil here this method will return `nil`.

tas50: I'm noodling on how to destroy this whole set of methods. It's all pretty fragile.

kitchen-tests/cookbooks/end_to_end/recipes/_macos_userdefaults.rb
```diff
@@ -11,16 +11,22 @@
...
 
-macos_userdefaults "set dock size" do
+macos_userdefaults "Set the dock size" do
   domain "com.apple.dock"
   type "integer"
   key "tilesize"
   value "20"
 end
+
+macos_userdefaults "value with space" do
```
tas50, line 28: lamont says this is blocking unless we capitalize this

lamont-granquist: literally unplayable

lib/chef/resource/macos_userdefaults.rb
lamont-granquist: bit weird that you slurp up all the values and then prepend the type rather than just appending.

prepend is a nifty trick but it should probably be very lightly used

kitchen-tests/cookbooks/end_to_end/recipes/_macos_userdefaults.rb
tas50: Fixed

lib/chef/resource/macos_userdefaults.rb
tas50: It avoided having to create an empty array, then optionally concat the type in, then toss the values in.

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -28,99 +30,210 @@ class MacosUserDefaults < Chef::Resource
...
+
+      property :host, [String, Symbol],
+        description: "Set either :current or a hostname to set the user default at the host level.",
+        desired_state: false,
+        introduced: "16.3"
 
       property :value, [Integer, Float, String, TrueClass, FalseClass, Hash, Array],
-        description: "The value of the key.",
-        required: true
+        description: "The value of the key. Note: If you set the `type` property to `bool` we'll automatically try to convert various `String` forms of the Boolean true/false that Apple accepts in the defaults command: 0/1, 'TRUE'/'FALSE,' 'true'/false', 'YES'/'NO', or 'yes'/'no'.",
```
phiggins, line 86: I thought this sounded better at first, now I'm not so sure...

```suggestion
        description: "The value of the key. Note: With the `type` property set to `bool`, `String` forms of Boolean true/false values that Apple accepts in the defaults command will be coerced: 0/1, 'TRUE'/'FALSE,' 'true'/false', 'YES'/'NO', or 'yes'/'no'.",
```

lib/chef/resource/macos_userdefaults.rb
```diff
@@ -28,99 +30,210 @@ class MacosUserDefaults < Chef::Resource
...
+        # @param [String, Integer, Boolean] input <description>
+        #
+        # @return [String] TRUE or FALSE
+        #
+        def bool_to_defaults_bool(input)
+          return ["TRUE"] if [true, "TRUE", "1", "true", "YES", "yes"].include?(input)
+          return ["FALSE"] if [false, "FALSE", "0", "false", "NO", "no"].include?(input)
+
+          # make sure it's very clear bad input was given
+          raise ArgumentError, "#{input} cannot be converted to a boolean value for use with Apple's defaults command. Acceptable values are: 'TRUE', 'YES', 'true, 'yes', '0', true, 'FALSE', 'false', 'NO', 'no', 1, or false."
```
phiggins, line 227:
```suggestion
          raise ArgumentError, "#{input} cannot be converted to a boolean value for use with Apple's defaults command. Acceptable values are: 'TRUE', 'YES', 'true, 'yes', '0', true, 'FALSE', 'false', 'NO', 'no', '1', or false."
```

spec/unit/resource/macos_user_defaults_spec.rb
```diff
@@ -20,16 +20,117 @@
...
+    it "autodetects boolean type from FalseClass value and passes a 'FALSE' string" do
+      resource.value = false
+      expect(provider.defaults_modify_cmd).to eql(["/usr/bin/defaults", "write", "NSGlobalDomain", "foo", "-bool", "FALSE"])
+    end
+
+    it "autodetects dict type from Hash value and flattens keys & values" do
+      resource.value = { "foo" => "bar" }
+      expect(provider.defaults_modify_cmd).to eql(["/usr/bin/defaults", "write", "NSGlobalDomain", "foo", "-dict", "foo", "bar"])
+    end
+  end
```
phiggins, line 135: Any particular reason these tests use `.eql` instead of `.eq`?

spec/unit/resource/macos_user_defaults_spec.rb
```diff
@@ -20,16 +20,117 @@
...
 
   it "has a resource name of :macos_userdefaults" do
     expect(resource.resource_name).to eql(:macos_userdefaults)
   end
 
+  it "the domain property defaults to NSGlobalDomain" do
+    expect(resource.domain).to eql("NSGlobalDomain")
+  end
+
+  it "the value property coerces keys in hashes to symbols so we can compare them with plist data" do
```
phiggins, line 33:
```suggestion
  it "the value property coerces keys in hashes to strings so we can compare them with plist data" do
```

spec/unit/resource/macos_user_defaults_spec.rb
tas50: habit?

tas50: Fixed

Issue #9242, opened 2020-01-20
windows_share is not idempotent when using local groups


Xorima:
## Description
The resource `windows_share` is not idempotent when using local groups

## Chef Version
`14.12.9`

## Platform Version
Windows 2016

## Replication Case
```
directory 'C:\foo' do
  recursive true
  rights :modify, "#{node['hostname']}\\mygroup"
end

windows_share 'foo' do
  path            'C:\foo'
  read_users      "#{node['hostname']}\\mygroup"
  change_users    "#{node['hostname']}\\mygroup"
end

```

## Client Output
```
* directory[C:\foo] action create (up to date)
         * windows_share[foo] action create
           - update foo
           -   set change_users to ["WIN-4EVKS4M49GI\\mygroup"] (was ["mygroup"])
           -   set read_users   to ["WIN-4EVKS4M49GI\\mygroup"] (was ["mygroup"])
```

It appears that the return is bringing back just the group name where the set has to have the server name for local groups to work.

PR #9796, opened 2020-05-05
Fixed windows share is not idempotent when using local groups

sanga1794:
## Description
windows_share is not idempotent when using local groups

## Related Issue
Fixed: https://github.com/chef/chef/issues/9242

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #10122, opened 2020-07-07
Fixed windows_share resource is not idempotent

sanga1794:
Signed-off-by: sanga17 <sausekar@msystechnologies.com>

## Description
windows_share is not idempotent when using local groups

## Related Issue
Fixed: 1) https://github.com/chef/chef/issues/9242
2) https://github.com/chef/chef/issues/10226

## Types of changes

## Checklist:


sanga1794:
review comment are addressed for related PR: https://github.com/chef/chef/pull/9796

lib/chef/resource/windows_share.rb
```diff
@@ -336,8 +331,28 @@ def bool_string(bool)
           # bool ? 1 : 0
           bool ? "$true" : "$false"
         end
+
+        # Grant-SmbShareAccess sets only one permission to one user so we need to remove from the lower permission access
+        # For change_users remove common full_users from it
+        # For read_users remove common full_users as well as change_users
+        def set_users
+          new_resource.change_users = new_resource.change_users - new_resource.full_users
```
tas50, line 339: Modifying new_resource properties mid run is something we generally don't recommend. This needs to be handled in a new variable.

sanga1794: Hi @tas50 , I have tried by using new variable, still it will fail to acheive idempotency, I have also tried by another approach(re arranging the current_resource) but it will require lot of changes in the code, so any suggestion for the solution?

lamont-granquist: i think the problem is that "change_users" is a verb-style property rather than a noun, which breaks the simple idempotency model, which means converge_if_changed cannot be used, since that strictly compares a desired state with a current state.

most likely abandoning converge_if_changed is the right thing to do here, this resource is outside of the scope of what it can handle.

lib/chef/resource/windows_share.rb
```diff
@@ -338,8 +335,38 @@ def bool_string(bool)
...
+          users["full_users"] = new_resource.full_users
+
+          users["change_users"] = new_resource.change_users - new_resource.full_users
+          users["read_users"] = new_resource.read_users - (new_resource.full_users + new_resource.change_users)
+          users
+        end
+
+        # Compare the full_users, change_users and read_users from current_resource and new_resource
+        # @returns boolean True/False
+        def compare_users
```
lamont-granquist, line 353: This should be named something like `def users_changed?` and the sense of it should be inverted (this makes it read nicer without the NOT when it gets used above)

lib/chef/resource/windows_share.rb
```diff
@@ -338,8 +335,38 @@ def bool_string(bool)
...
+
+        # Grant-SmbShareAccess sets only one permission to one user so we need to remove from the lower permission access
+        # For change_users remove common full_users from it
+        # For read_users remove common full_users as well as change_users
+        def new_resource_users
+          users = {}
+          users["full_users"] = new_resource.full_users
+
+          users["change_users"] = new_resource.change_users - new_resource.full_users
+          users["read_users"] = new_resource.read_users - (new_resource.full_users + new_resource.change_users)
```
lamont-granquist, line 347: this hash indirection should be removed and these should just be plain ruby variables in the method below.

lamont-granquist (changes requested):

lamont-granquist:
This really needs at least unit tests (if not figuring out how to do real functional tests, but i don't know how you could setup a 'test' SMB share and destroy it reliably on every test run).  For unit tests the powershell command can be mocked with the output that it would normally produce.  It should test the cases where the chef user requests the same user on multiple rights, and should test the domain mangling.

lib/chef/resource/windows_share.rb
sanga1794: @lamont-granquist, i have also used this method in other methods too e.g. https://github.com/MsysTechnologiesllc/chef/blob/3c268b8c94db20d64ae729ce0525a8951272c8b9/lib/chef/resource/windows_share.rb#L313

lamont-granquist: yeah but those accesses should change as well, these should just be `@full_users` ivars with accessors on them rather than doing access to the state through an unnecessary Hash.

sanga1794: ok will change accordingly, Thanks!! @lamont-granquist

sanga1794:
@tas50 and @lamont-granquist, I have rebased the PR so please have a look, Thanks!!

sanga1794:
@tas50 and @lamont-granquist  Please have a look into this PR. Thanks!!

sanga1794:
@tas50 @lamont-granquist @phiggins @marcparadise Please have look into the PR. Thanks!!

johnmccrae:
@sanga1794 can you please rebase this and push it again?

johnmccrae (approved):

johnmccrae:
We just added some important changes and updates to the code base in the last 24 hours. Can you please rebase and push one final time? I just approved your PR

tpowell-progress:
Topic for review for 2022-12-13 PR meeting

lib/chef/resource/windows_share.rb
```diff
@@ -306,13 +305,13 @@ def permissions_need_update?(type)
...
-          return false if current_resource.nil? && new_resource.send(property_name).empty?
+          return false if current_resource.nil? && instance_variable_get("@#{property_name}").empty?
 
           # brand new share with new permissions to set
-          return true if current_resource.nil? && !new_resource.send(property_name).empty?
+          return true if current_resource.nil? && !instance_variable_get("@#{property_name}").empty?
 
           # there's a difference between the current and desired state
-          return true unless (new_resource.send(property_name) - current_resource.send(property_name)).empty?
+          return true unless (instance_variable_get("@#{property_name}") - current_resource.send(property_name)).empty?
```
jaymzh, line 314: We've moved away from using the instance vars, and instead use the methods on the right object. This feels like a regression.

lib/chef/resource/windows_share.rb
```diff
@@ -284,7 +283,7 @@ def update_permissions
           revoke_user_permissions(users_to_revoke) unless users_to_revoke.empty?
 
           # set permissions for each of the permission types
-          %w{full read change}.each do |perm_type|
+          %w{read change full}.each do |perm_type|
```
jaymzh, line 286: what's the reason for this?

jaymzh (changes requested):
Overall seems like a good change, but have some concerns.

lib/chef/resource/windows_share.rb
neha-p6: I was backporting this fix to chef16 and def `new_resource_users` is using those instance variables instead of setting the properties on new_resource due to which `new_resource.send(property_name)` never actually has the correct values. So it then results in test failure https://buildkite.com/chef-oss/chef-chef-chef-16-verify/builds/1448#0185bfd7-35c0-4015-bda2-0363cfff6bf5/302-1202. It's working here because of `instance_variable_get`

Ref https://github.com/chef/chef/pull/13496/files#diff-4a31658e49ff21d95272bb16f9375a8a929d96388ba81eb837f7989c78b420daR317-R323

lib/chef/resource/windows_share.rb
```diff
@@ -331,8 +330,35 @@ def bool_string(bool)
...
         end
+
+        # Grant-SmbShareAccess sets only one permission to one user so we need to remove from the lower permission access
+        # For change_users remove common full_users from it
+        # For read_users remove common full_users as well as change_users
+        def new_resource_users
+          @full_users = new_resource.full_users
+          @change_users = new_resource.change_users - new_resource.full_users
+          @read_users = new_resource.read_users - (new_resource.full_users + new_resource.change_users)
+        end
```
jaymzh, line 341: Change the instance variables underneath the property is not only a bad idea, I'm fairly certain it's the reasons for the issues you're seeing.

Instead, do something like:

```ruby
def effective_full_users
  new_resource.full_users
end

def effective_change_users
  new_resource.change_users - new_resource.full_users
end

def effective_read_users
  new_resource.read_users - (new_resource.full_users + new_resource.change_users)
end
```

Then instead of using `instance_variable_get`, you can call `new_resource.send`.

This is the moral equivalent of what @tas50 suggested back in 2020 (though with a method instead of a var, he said "use a different variable"), you said it didn't work, but you didn't say what specifically didn't work, and it should definitely work. Lets do that and then debug it, because otherwise we're setting up a footgun here.

jaymzh (changes requested):

PR #10126, opened 2020-07-07
Fix `powershell_exec!` test.

phiggins:
This test was running the same code as the one right above it and expecting a different result.


tas50:
That's def my copy pasta there

PR #10127, opened 2020-07-07
Fix systemd unit test on Windows.

phiggins:
The systemd resource has a `user` property but no `password`, and `Mixlib::Shellout` fails when setting a user with no password on Windows. This should only reasonably happen while running the unit tests on Windows, so guard against that case.


phiggins:
After the discussion in Slack last night, I changed this to just skip the failing systemd tests on Windows.

PR #10129, opened 2020-07-08
Fix some windows package tests

phiggins:
Fixes some of the tests for the Windows package provider. There are other failing tests in this file that are beyond me at the moment.

### DIFF
--- a/spec/unit/provider/package/windows_spec.rb
+++ b/spec/unit/provider/package/windows_spec.rb
@@ -42,7 +42,7 @@ describe Chef::Provider::Package::Windows, :windows_only do
   let(:cache_path) { 'c:\\cache\\' }
 
   before(:each) do
-    allow(::File).to receive(:exist?).with(provider.new_resource.source).and_return(true)
+    allow(::File).to receive(:exist?).with(new_resource.source).and_return(true)
   end
 
   describe "load_current_resource" do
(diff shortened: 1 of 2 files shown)


spec/unit/provider/package/windows_spec.rb
```diff
@@ -233,9 +233,9 @@
       end
     end
 
-    it "returns @installer_type if it is set" do
-      provider.new_resource.installer_type(:downeaster)
-      expect(provider.installer_type).to eql(:downeaster)
+    it "returns the resource's installer_type if it is set" do
+      provider.new_resource.installer_type(:nsis)
+      expect(provider.installer_type).to eql(:nsis)
     end
```
phiggins, line 239: The resource has validation on the `installer_type` which probably means that behavior changed after this test was written and it was never updated.

spec/unit/provider/package/windows_spec.rb
```diff
@@ -414,51 +414,23 @@
...
-    context "checksum is in uppercase" do
-      before do
-        new_resource.checksum = new_resource.checksum.upcase
-      end
+      new_resource.checksum("ffffffffffffffff")
 
-      include_context "valid checksum"
-      include_context "invalid checksum"
+      expect { provider.run_action(:install) }.to raise_error(Chef::Exceptions::Package)
     end
```
phiggins, line 434: The `validate_content!` method was removed in #9790 and these tests weren't updated to match the new way of validating checksums. The resource does coercion of the checksum to lowercase so checking both uppercase and lowercase isn't necessary.

PR #10135, opened 2020-07-08
Fix windows knife bootstrap context tests

phiggins:
Fix Windows knife bootstrap context tests. I couldn't resist doing some cleanup while I was there. 😅

### DIFF
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -67,16 +67,16 @@ class Chef
             validation_client_name "#{chef_config[:validation_client_name]}"
           CONFIG
 
-          unless chef_config[:file_cache_path].nil?
-            client_rb << "file_cache_path   \"#{chef_config[:file_cache_path]}\"\n"
+          if chef_config[:file_cache_path]
+            client_rb << %Q{file_cache_path   "#{chef_config[:file_cache_path]}"\n}
           end
 
-          unless chef_config[:file_backup_path].nil?
-            client_rb << "file_backup_path  \"#{chef_config[:file_backup_path]}\"\n"
+          if chef_config[:file_backup_path]
+            client_rb << %Q{file_backup_path  "#{chef_config[:file_backup_path]}"\n}
           end
 
-          unless chef_config[:chef_license].nil?
-            client_rb << "chef_license \"#{chef_config[:chef_license]}\"\n"
+          if chef_config[:chef_license]
+            client_rb << %Q{chef_license "#{chef_config[:chef_license]}"\n}
           end
 
           if config[:chef_node_name]
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -30,16 +30,6 @@ class Chef
       # * @run_list - the run list for the node to bootstrap
       #
       class WindowsBootstrapContext < BootstrapContext
-        attr_accessor :config
-        attr_accessor :chef_config
-
-        def initialize(config, run_list, chef_config, secret = nil)
-          @config       = config
-          @run_list     = run_list
-          @chef_config  = chef_config
-          @secret       = secret
-          super(config, run_list, chef_config, secret)
-        end
 
         # This is a duplicate of ChefConfig::PathHelper.cleanpath, however
         # this presumes Windows so we can avoid changing the method definitions
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -65,10 +65,15 @@ class Chef
           client_rb = <<~CONFIG
             chef_server_url  "#{chef_config[:chef_server_url]}"
             validation_client_name "#{chef_config[:validation_client_name]}"
-            file_cache_path   "#{ChefConfig::Config.var_chef_dir(true)}/cache"
-            file_backup_path  "#{ChefConfig::Config.var_chef_dir(true)}/backup"
-            cache_options     ({:path => "#{ChefConfig::Config.etc_chef_dir(true)}/cache/checksums", :skip_expires => true})
           CONFIG
+
+          unless chef_config[:file_cache_path].nil?
+            client_rb << "file_cache_path   \"#{chef_config[:file_cache_path]}\"\n"
+          end
+
+          unless chef_config[:file_backup_path].nil?
+            client_rb << "file_backup_path  \"#{chef_config[:file_backup_path]}\"\n"
+          end
 
           unless chef_config[:chef_license].nil?
             client_rb << "chef_license \"#{chef_config[:chef_license]}\"\n"
--- a/spec/unit/knife/core/windows_bootstrap_context_spec.rb
+++ b/spec/unit/knife/core/windows_bootstrap_context_spec.rb
@@ -20,14 +20,12 @@ require "spec_helper"
 require "chef/knife/core/windows_bootstrap_context"
 describe Chef::Knife::Core::WindowsBootstrapContext do
   let(:config) { {} }
-  let(:chef_config) { Chef::Config.save } # "dup" to a hash
+  let(:chef_config) { {} }
   let(:bootstrap_context) { Chef::Knife::Core::WindowsBootstrapContext.new(config, nil, chef_config, nil) }
 
   describe "fips" do
     context "when fips is set" do
-      before do
-        chef_config[:fips] = true
-      end
+      let(:chef_config) { { fips: true } }
 
       it "sets fips mode in the client.rb" do
         expect(bootstrap_context.config_content).to match(/fips true/)
@@ -35,9 +33,7 @@ describe Chef::Knife::Core::WindowsBootstrapContext do
     end
 
     context "when fips is not set" do
-      before do
-        chef_config[:fips] = false
-      end
+      let(:chef_config) { { fips: false } }
 
       it "sets fips mode in the client.rb" do
         expect(bootstrap_context.config_content).not_to match(/fips true/)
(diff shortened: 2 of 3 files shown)


lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -30,16 +30,6 @@ module Core
...
-        attr_accessor :config
-        attr_accessor :chef_config
-
-        def initialize(config, run_list, chef_config, secret = nil)
-          @config       = config
-          @run_list     = run_list
-          @chef_config  = chef_config
-          @secret       = secret
-          super(config, run_list, chef_config, secret)
-        end
```
phiggins, line 42: This code is exactly the same as what's in the superclass.

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -75,13 +65,18 @@ def config_content
           client_rb = <<~CONFIG
             chef_server_url  "#{chef_config[:chef_server_url]}"
             validation_client_name "#{chef_config[:validation_client_name]}"
-            file_cache_path   "#{ChefConfig::Config.var_chef_dir(true)}/cache"
-            file_backup_path  "#{ChefConfig::Config.var_chef_dir(true)}/backup"
-            cache_options     ({:path => "#{ChefConfig::Config.etc_chef_dir(true)}/cache/checksums", :skip_expires => true})
```
phiggins, line 80: `cache_options` was ignoring the user's input and had a subtle path separator change that made the test fail. It is also marked deprecated so it seems bad to include it in new code: https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L804-L813

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -75,13 +65,18 @@ def config_content
           client_rb = <<~CONFIG
             chef_server_url  "#{chef_config[:chef_server_url]}"
             validation_client_name "#{chef_config[:validation_client_name]}"
-            file_cache_path   "#{ChefConfig::Config.var_chef_dir(true)}/cache"
-            file_backup_path  "#{ChefConfig::Config.var_chef_dir(true)}/backup"
```
phiggins, line 79: These were changed to accept user's input but Windows was not updated to match: https://github.com/chef/chef/pull/9361

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -20,24 +20,20 @@
 require "chef/knife/core/windows_bootstrap_context"
 describe Chef::Knife::Core::WindowsBootstrapContext do
   let(:config) { {} }
-  let(:chef_config) { Chef::Config.save } # "dup" to a hash
+  let(:chef_config) { {} }
```
phiggins, line 23: Nothing broke when I changed this, so ✂️

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -183,7 +171,7 @@
 
   describe "#start_chef" do
     it "returns the expected string" do
-      expect(bootstrap_context.start_chef).to match(%r{SET \"PATH=%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem;%SYSTEMROOT%\\System32\\WindowsPowerShell\\v1.0\\;C:\\ruby\\bin;C:\/opscode\/chef\\bin;C:\/opscode\/chef\\embedded\\bin\;%PATH%\"\n})
+      expect(bootstrap_context.start_chef).to match(%r{SET \"PATH=%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem;%SYSTEMROOT%\\System32\\WindowsPowerShell\\v1.0\\;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\;%PATH%\"\n})
     end
```
phiggins, line 175: This is one of the test fixes. Changed some path separators.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -96,74 +86,74 @@
...
 
     it "generates the config file data" do
       expected = <<~EXPECTED
         echo.chef_server_url  "http://chef.example.com:4444"
         echo.validation_client_name "chef-validator-testing"
-        echo.file_cache_path   "C:/chef/cache"
-        echo.file_backup_path  "C:/chef/backup"
-        echo.cache_options     ^({:path =^> "C:/chef/cache/checksums", :skip_expires =^> true}^)
+        echo.file_cache_path   "X:/my_cache_path/cache"
+        echo.file_backup_path  "X:/my_cache_path/backup"
```
phiggins, line 156: This is one of the text fixes. These paths were all busted.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -51,10 +47,7 @@
     let(:crt_files) { ::Dir.glob(::File.join(mock_cert_dir, "*.crt")) }
     let(:pem_files) { ::Dir.glob(::File.join(mock_cert_dir, "*.pem")) }
     let(:other_files) { ::Dir.glob(::File.join(mock_cert_dir, "*")) - crt_files - pem_files }
-
-    before do
-      bootstrap_context.instance_variable_set(:@chef_config, Mash.new(trusted_certs_dir: mock_cert_dir))
-    end
+    let(:chef_config) { { trusted_certs_dir: mock_cert_dir } }
```
phiggins, line 50: A lot of these tests were setting this instance variable directly instead of passing a hash while instantiating, which it turns out was completely unnecessary.

lib/chef/knife/core/windows_bootstrap_context.rb
phiggins: Discussed this in slack and the consensus was this is probably a bug.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -82,12 +75,9 @@
   end
 
   describe "validation_key" do
-    before do
-      bootstrap_context.instance_variable_set(:@chef_config, Mash.new(validation_key: "C:\\chef\\key.pem"))
-    end
+    let(:chef_config) { { validation_key: "C:\\chef\\key.pem" } }
 
     it "should return false if validation_key does not exist" do
-      allow(::File).to receive(:expand_path).with("C:\\chef\\key.pem").and_call_original
```
phiggins, line 90: `allow(...).to receive(...).and_call_original` is essentially a no-op.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
lamont-granquist: This makes me nervous since you're snipping out all the default values in Chef::Config (although perhaps I'm missing context here where chef_config != Chef::Config, but then the point still applies though where that design is probably fundamentally wrong).  If we're not testing against the actual Chef::Config here we have no confidence that it works with all the Chef::Config defaults set.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
lamont-granquist: lol, yeah.

lamont-granquist: particularly with "allow" and not "expect", that's just silly

spec/unit/knife/core/windows_bootstrap_context_spec.rb
```diff
@@ -96,74 +86,74 @@
...
       it "sets STDERR in client.rb" do
-        bootstrap_context.instance_variable_set(:@chef_config, Mash.new(config_log_location: STDERR))
         expect(bootstrap_context.get_log_location).to eq("STDERR\n")
       end
     end
 
     context "when config_log_location value is path to a file" do
+      let(:chef_config) { { config_log_location: "C:\\chef\\chef.log" } }
       it "sets file path in client.rb" do
-        bootstrap_context.instance_variable_set(:@chef_config, Mash.new(config_log_location: "C:\\chef\\chef.log"))
```
lamont-granquist, line 128: Yeah so this is awful, but its a symptom of excessive stubbing, the chef_config should just be Chef::Config.

That is a global variable so has all the global variable issues associated with testing but we're already calling Chef::Config.reset! in the spec_helper on every example to clear the global state, so we don't have to overly mock the Chef::Config.

lamont-granquist: I'm also less and less concerned about let()'ing and before()'ing everything:

```ruby
it "sets file path in client.rb" do
  bootstrap_context.chef_config.config_log_location "C:\\chef\\chef.log"
  [....]
end
```

spec/unit/knife/core/windows_bootstrap_context_spec.rb
phiggins: It is Chef::Config where it gets used in lib/ https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L518-L527

The other bootstrap context tests provide a few values by default, but then they get superseded in the majority of the tests: https://github.com/chef/chef/blob/master/spec/unit/knife/core/bootstrap_context_spec.rb#L25-L33

I was thinking along the same lines originally, and I wondered if that had something to do with the weird way the `chef_config` was being overwritten in most of the tests. I thought using an empty hash made more sense to prevent hidden dependencies that would cause problems if the user changed the defaults. Since it doesn't seem to make a difference in the tests as they exist now I'll change it back. 👍

spec/unit/knife/core/windows_bootstrap_context_spec.rb
phiggins: I had guessed it was because at one point `@chef_config` was not as easily injectable. With the initializer taking this as a param, and this not being used to avoid a weird side effect or something, it seems pretty pointless to me.

I do agree with you that setting the specific value a test is concerned with on the default config object makes more sense as a realistic testing scenario though, I can change these.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
phiggins: I changed the tests to just use `Chef::Config` and then add on to that if necessary for a more realistic testing scenario.

spec/unit/knife/core/windows_bootstrap_context_spec.rb
lamont-granquist: i would guess it was written this way just because someone thought that in unit tests you had to mock everything.

phiggins:
We decided this is the wrong way to go about fixing the underlying bootstrap issues.

PR #10137, opened 2020-07-08
Extract shell_out helper method to chef-utils for reuse in ohai, etc

lamont-granquist:
Okay I think I made some improvements which make this whole framework bound a lot less tightly bound to "global state" and calling class methods.  The `ChefUtils.systemd?` calling convention may be somewhat of a bad idea, but after ripping some of that out I put it back in due to not wanting to break public APIs, and I believe that this does not break public APIs.

What I did do though is in some of the modules that referenced class names, I instead switched to mixin in those modules and then calling the methods, which I believe (going forwards) is the more correct way of writing these -- it makes it easier to include those methods into external classes and override things like the __transport_connection, whereas if they call class methods then external software consumers would need to directly monkeypatch into ChefUtils classes to override those methods (which is clearly a poor API).

### DIFF
--- a/chef-utils/lib/chef-utils/dsl/path_sanity.rb
+++ b/chef-utils/lib/chef-utils/dsl/path_sanity.rb
@@ -30,27 +30,27 @@ module ChefUtils
         path_separator = ChefUtils.windows? ? ";" : ":"
         # ensure the Ruby and Gem bindirs are included for omnibus chef installs
         new_paths = env_path.split(path_separator)
-        [ ChefUtils::DSL::PathSanity.ruby_bindir, ChefUtils::DSL::PathSanity.gem_bindir ].compact.each do |path|
+        [ __ruby_bindir, __gem_bindir ].compact.each do |path|
           new_paths = [ path ] + new_paths unless new_paths.include?(path)
         end
-        ChefUtils::DSL::PathSanity.sane_paths.each do |path|
+        __sane_paths.each do |path|
           new_paths << path unless new_paths.include?(path)
         end
         new_paths.join(path_separator).encode("utf-8", invalid: :replace, undef: :replace)
       end
 
-      class << self
-        def sane_paths
-          ChefUtils.windows? ? %w{} : %w{/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin}
-        end
+      private
 
-        def ruby_bindir
-          RbConfig::CONFIG["bindir"]
-        end
+      def __sane_paths
+        ChefUtils.windows? ? %w{} : %w{/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin}
+      end
 
-        def gem_bindir
-          Gem.bindir
-        end
+      def __ruby_bindir
+        RbConfig::CONFIG["bindir"]
+      end
+
+      def __gem_bindir
+        Gem.bindir
       end
 
       extend self
--- a/chef-utils/lib/chef-utils/dsl/service.rb
+++ b/chef-utils/lib/chef-utils/dsl/service.rb
@@ -25,6 +25,7 @@ module ChefUtils
     module Service
       include Internal
       include TrainHelpers
+      include Introspection
 
       # Returns if debian's old rc.d manager is installed (not necessarily the primary init system).
       #
@@ -97,8 +98,8 @@ module ChefUtils
           file_exist?("/etc/rc.d/#{script}")
         when :systemd
           file_exist?("/etc/init.d/#{script}") ||
-            ChefUtils::DSL::Introspection.has_systemd_service_unit?(script) ||
-            ChefUtils::DSL::Introspection.has_systemd_unit?(script)
+            has_systemd_service_unit?(script) ||
+            has_systemd_unit?(script)
         else
           raise ArgumentError, "type of service must be one of :initd, :upstart, :xinetd, :etc_rcd, or :systemd"
         end
--- a/chef-utils/lib/chef-utils/internal.rb
+++ b/chef-utils/lib/chef-utils/internal.rb
@@ -75,9 +75,9 @@ module ChefUtils
       end
     end
 
-    # This should be set to a Train::FIXME instance.  You should wire this up to nil for not using a train transport connection.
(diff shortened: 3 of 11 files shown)


tas50:
:dope:

tas50:
Make sure chef-utils deps on mixlib-shellout now

lamont-granquist:
Yeah that's the problem with this.  I think that needs to be inverted and this shell_out helper has to go into mixlib-shellout and then mixlib-shellout needs to take a dep on chef-utils.

I stumbled over this a bit because that felt like mixlib-shellout would then take a dep on train, but that is the point of doing it in a dep-injection kind of style, which is that chef+ohai can DI train into mixlib-shellout and get the right behavior, without mixlib-shellout directly taking a dep on train.

tas50:
It always felt like the mixlib should have the helper to start with so I'm fine with that.

lamont-granquist:
Yeah I just definitely don't want mixlib-shellout taking a dep on train (although it means that mixlib-shellout may need a dep on train, and train may need to run those mixlib-shellout tests in its CI)

chef-utils/lib/chef-utils/dsl/path_sanity.rb
```diff
@@ -30,27 +30,27 @@ def sanitized_path(env = nil)
...
-      class << self
-        def sane_paths
-          ChefUtils.windows? ? %w{} : %w{/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin}
-        end
+      private
 
-        def ruby_bindir
-          RbConfig::CONFIG["bindir"]
-        end
+      def __sane_paths
```
phiggins, line 44: I know this is not all new code, but it'd be nice if we could avoid new uses of sane/sanity.

phiggins: Also is this method called outside of this module? If not, it could be a constant.

lamont-granquist: its a "dependency injection hook" so it can't be a constant and is sort of a public API.  instead of constructor injection we override the inherited methods in the "subclass" where its wired up to.

i don't know how much people have discovered "path sanity" and started to use it externally as a public API thing.  we can change it, but i think i'd prefer to do that deliberately as a separate issue.  we should probably batch that up with the allow/deny_list stuff for 17.0 or something.

lamont-granquist: (the "path_sanity" term dates back to like chef 0.8.x days or something along those lines)

chef-utils/lib/chef-utils/dsl/path_sanity.rb
```diff
@@ -30,27 +30,27 @@ def sanitized_path(env = nil)
         path_separator = ChefUtils.windows? ? ";" : ":"
         # ensure the Ruby and Gem bindirs are included for omnibus chef installs
         new_paths = env_path.split(path_separator)
-        [ ChefUtils::DSL::PathSanity.ruby_bindir, ChefUtils::DSL::PathSanity.gem_bindir ].compact.each do |path|
+        [ __ruby_bindir, __gem_bindir ].compact.each do |path|
           new_paths = [ path ] + new_paths unless new_paths.include?(path)
         end
-        ChefUtils::DSL::PathSanity.sane_paths.each do |path|
```
lamont-granquist, line 36: These references to the class methods are removed to make it easier to override

chef-utils/lib/chef-utils/dsl/service.rb
```diff
@@ -97,8 +98,8 @@ def service_script_exist?(type, script)
           file_exist?("/etc/rc.d/#{script}")
         when :systemd
           file_exist?("/etc/init.d/#{script}") ||
-            ChefUtils::DSL::Introspection.has_systemd_service_unit?(script) ||
-            ChefUtils::DSL::Introspection.has_systemd_unit?(script)
```
lamont-granquist, line 101: good example of the patter of removing the references to the class methods and using include intead

chef-utils/lib/chef-utils/internal.rb
```diff
@@ -65,13 +75,39 @@ def __env_path
...
 
+    # This should be set to a Train::FIXME instance.  You should wire this up to nil for not using a train transport connection.
+    #
+    # @return [Train::FIXME]
+    #
     # @api private
+    #
     def __transport_connection
+      # Software consumers MUST override this method with their own implementation.  The default behavior here is subject to change.
       return Chef.run_context.transport_connection if defined?(Chef) && Chef.respond_to?(:run_context) && Chef&.run_context&.transport_connection
```
lamont-granquist, line 86: I initially removed this line to force consumers to mixin the DSL methods and then force them to override this method (similar to __log and __config below).  I backed that out because that broke `ChefUtils.systemd?`.  The comment is meant to encourage any external consumers to treat this like `raise NotImplementedError` so that they are not tightly coupled to the magic here.  The `@api private` means that this code can change and break at any time.

spec/unit/mixin/shell_out_spec.rb
```diff
@@ -57,38 +56,38 @@ def env_path
           describe "and environment is an option" do
             it "should not change environment language settings when they are set to nil" do
               options = { environment: { "LC_ALL" => nil, "LANGUAGE" => nil, "LANG" => nil, env_path => nil } }
-              expect(Chef::Mixin::ShellOut).to receive(:shell_out_command).with(cmd, **options).and_return(retobj)
+              expect(shell_out_obj).to receive(:__shell_out_command).with(cmd, **options).and_return(retobj)
```
lamont-granquist, line 59: This is some fallout from pushing more stuff into instances and having less references to "global" class methods.  Don't think this is a public API.

chef-utils/lib/chef-utils/dsl/path_sanity.rb
tas50: We're going to need to maintain those existing names but mark them deprecated and just ship new methods. Sooner we drop those names the better

lamont-granquist:
shipping this without fixing this for "default_paths" in this PR, there is a surprisingly large amount of patching to do to fix that issue, it needs to be a separate concern.

PR #10140, opened 2020-07-10
Bump the server api version to 2

lamont-granquist:
see https://github.com/chef/chef-cli/issues/111 for related bug.

🚧 need to ponder the blast radius of this one 🚧 

although i think we clearly need to do this.

### DIFF
--- a/lib/chef/http/authenticator.rb
+++ b/lib/chef/http/authenticator.rb
@@ -24,7 +24,7 @@ class Chef
   class HTTP
     class Authenticator
 
-      DEFAULT_SERVER_API_VERSION = "1".freeze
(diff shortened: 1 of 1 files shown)


lamont-granquist:
I think there's another bug here that if we're on server_api < 2 then we render manifests for cookbook uploading using v2 no matter what.  I think only the downloading of cookbooks was made server_api aware, but i need to dig into that... there's clearly some other bug as well.

lamont-granquist:
Well this should be right, i still think there's a bug somewhere that we're getting ManifestV2 somehow when we're talking to a V1 server, but this shoudn't change that afaik.

PR #10147, opened 2020-07-13
Use powershell_out vs. powershell_script in hostname

### LINKED ISSUE #9823: hostname output on windows is less than ideal
We can do better by using powershell_out instead of powershell_script here:

```
           * powershell_script[set hostname] action run
             - execute "C:\Windows\system32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Bypass -InputFormat None -File "C:/Users/vagrant/AppData/Local/Temp/chef-script20200507-2624-t3izci.ps1"
           * reboot[setting hostname] action nothing (skipped due to action :nothing)
           * reboot[setting hostname] action request_reboot[2020-05-07T18:19:30+00:00] WARN: Reboot requested:'setting hostname'

             - request a system reboot to occur if the run succeeds

```

tas50:
This improves the log output and probably speeds up the run since we're
not writing out a temporary script file anymore and then executing it. I
also removed some delcare_resource uses since we enabled unified_mode
for this resource. There's a ton more to remove here.

Fixes #9823

### DIFF
--- a/lib/chef/resource/hostname.rb
+++ b/lib/chef/resource/hostname.rb
@@ -231,27 +231,26 @@ class Chef
             if xml_contents.empty?
               Chef::Log.warn('Unable to properly parse and update C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml contents. Skipping file update.')
             else
-              declare_resource(:file, 'C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml') do
+              file 'C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml' do
                 content xml_contents
               end
             end
           end
 
-          # update via netdom
-          declare_resource(:powershell_script, "set hostname") do
-            code <<-EOH
-              $sysInfo = Get-WmiObject -Class Win32_ComputerSystem
-              $sysInfo.Rename("#{new_resource.hostname}")
-            EOH
-            notifies :request_reboot, "reboot[setting hostname]"
-            not_if { Socket.gethostbyname(Socket.gethostname).first == new_resource.hostname }
-          end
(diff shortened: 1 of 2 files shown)


lamont-granquist:
> I also removed some delcare_resource uses since we enabled unified_mode for this resource

- you can use declare_resource because of nuking the "Core" DSL, nothing to do with unified_mode.

nearly all the declare_resource stuff can go away, unless people are actually doing dynamic resource generation.

lib/chef/resource/hostname.rb
```diff
@@ -231,27 +231,25 @@ def updated_ec2_config_xml
             if xml_contents.empty?
               Chef::Log.warn('Unable to properly parse and update C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml contents. Skipping file update.')
             else
-              declare_resource(:file, 'C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml') do
+              file 'C:\Program Files\Amazon\Ec2ConfigService\Settings\config.xml' do
```
phiggins, line 234: Not really your responsibility but it feels like this string should be a constant...

tas50: let me hit it with chefstyle

tas50: actually that's to avoid having to escape the \ so that's intentional

phiggins: I mostly meant to avoid repeating the file name in this file.

tas50: Updated

tas50:
Confirmed on a real system that the logging works here

PR #10153, opened 2020-07-13
Don't allow tests to set an expectation without specific error type

phiggins:


spec/functional/resource/remote_file_spec.rb
```diff
@@ -409,13 +409,7 @@ def create_resource
...
         # This can legitimately raise either Errno::EADDRNOTAVAIL or Errno::ECONNREFUSED
         # in different Ruby versions.
-        old_value = RSpec::Expectations.configuration.on_potential_false_positives
-        RSpec::Expectations.configuration.on_potential_false_positives = :nothing
-        begin
-          expect { resource.run_action(:create) }.to raise_error
-        ensure
-          RSpec::Expectations.configuration.on_potential_false_positives = old_value
-        end
+        expect { resource.run_action(:create) }.to raise_error(SystemCallError)
```
phiggins, line 412: `SystemCallError` is the parent class of all the `Errno` exceptions.

spec/support/shared/unit/application_dot_d.rb
```diff
@@ -41,7 +41,6 @@
       expect(IO).to receive(:read).with(Pathname.new("#{client_d_dir}/00-foo.rb").cleanpath.to_s).and_return("foo 0")
       expect(IO).to receive(:read).with(Pathname.new("#{client_d_dir}/01-bar.rb").cleanpath.to_s).and_return("bar 0")
       expect(IO).to receive(:read).with(Pathname.new("#{client_d_dir}/02-strings.rb").cleanpath.to_s).and_return("strings 0")
-      allow(app).to receive(:apply_config).with(anything, Chef::Config.platform_specific_path("/etc/chef/client.rb")).and_call_original.ordered
```
phiggins, line 44: I changed this to a plain `expect` with no `.ordered` on the end and the test failed, so it doesn't seem like this gets called with these arguments at all.

spec/unit/chef_fs/parallelizer_spec.rb
```diff
@@ -405,7 +405,7 @@ def parallelize(inputs, options = {}, &block)
         expect do
           b = enum.enum_for(:each)
           b.next
-        end.to raise_error
+        end.to raise_error(RuntimeError, "each() called on parallel enumerable twice simultaneously! Bad mojo")
```
phiggins, line 408: Expecting `RuntimeError` here seemed a bit too generic so I added the test message.

PR #10154, opened 2020-07-14
Don't allow setting expectations on nil.

phiggins:


spec/unit/cookbook/gem_installer_spec.rb
```diff
@@ -95,7 +95,8 @@
 
   it "skip metadata installation when Chef::Config[:skip_gem_metadata_installation] is set to true" do
     Chef::Config[:skip_gem_metadata_installation] = true
-    expect(gem_installer.install).to_not receive(:shell_out!)
+    expect(gem_installer).to_not receive(:shell_out!)
+    expect(gem_installer.install).to be_nil
```
phiggins, line 99: This test was not really testing anything previously.

PR #10133, opened 2020-07-08
Support legacy DSS host keys with knife-ssh

tas50:
net-ssh 6.0 disabled these legacy keys by default. We need to enable the
support by passing this option.

Signed-off-by: Tim Smith <tsmith@chef.io>

PR #10164, opened 2020-07-15
Support legacy DSS host keys with knife-ssh

tas50:
net-ssh 6.0 disabled these legacy keys by default. We need to enable the
support by passing this option.

Backport of https://github.com/chef/chef/pull/10133

### DIFF
--- a/lib/chef/knife/ssh.rb
+++ b/lib/chef/knife/ssh.rb
@@ -298,6 +298,10 @@ class Chef
             opts[:keepalive] = true
             opts[:keepalive_interval] = ssh_config[:keepalive_interval]
           end
+          # maintain support for legacy key types / ciphers / key exchange algorithms.
+          # most importantly this adds back support for DSS host keys
+          # See https://github.com/net-ssh/net-ssh/pull/709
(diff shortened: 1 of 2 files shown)


lib/chef/knife/ssh.rb
```diff
@@ -298,6 +298,10 @@ def session_options(host, port, user = nil, gateway: false)
             opts[:keepalive] = true
             opts[:keepalive_interval] = ssh_config[:keepalive_interval]
           end
+          # maintain support for legacy key types / ciphers / key exchange algorithms.
+          # most importantly this adds back support for DSS host keys
+          # See https://github.com/net-ssh/net-ssh/pull/709
+          opts[:append_all_supported_algorithms] = true
```
phiggins, line 304: The comment on this method says it returns a hash, but this is going to return `true`.

```suggestion
          opts[:append_all_supported_algorithms] = true
          
          opts
```

phiggins: Nevermind, didn't notice this was in a `tap` block.

PR #10171, opened 2020-07-17
 Fixed mount Resource for bind mounts is not idempotent.

### LINKED ISSUE #4660: Mount Resource for bind mounts is not idempotent
## Description

Mount Resource for bind mounts is not idempotent
Here https://tickets.opscode.com/browse/CHEF-5254 is a similar bug opened with version 11.12.2 and fixed in version 11.14.0
## Chef Version

12.7.2 + Opscode server version
## Platform Version

Debian Jessie 8.3
## Replication Case

This is the provider code.

```
mount '/mnt' do
  device '/tmp'
  fstype 'ext4'
  options %w(bind rw)
  action [:mount, :enable]
end
```

Bind mountpoints in fstab file are mounted again in every chef-run, if you run chef-client four times you are going to have the resource mounted four times.

Fstab file is ok just one line per mount. 

Now mounted binds in jessie are not displaying the word bind as in squeeze with the mount command and using cat /proc/mounts  
## Client Output

``` shell
sudo chef-apply mount.rb

  * mount[/mnt] action mount
    - mount /tmp to /mnt
  * mount[/mnt] action enable
    - enable /tmp
vagrant:~/code$
vagrant:~/code$ sudo chef-apply mount.rb
  * mount[/mnt] action mount
    - mount /tmp to /mnt
  * mount[/mnt] action enable (up to date)
vagrant:~/code$ mount
/dev/sda1 on /mnt type ext4 (rw,relatime,errors=remount-ro,data=ordered)
/dev/sda1 on /mnt type ext4 (rw,relatime,errors=remount-ro,data=ordered)
```
### NOTE: CHEF CLIENT BUGS ONLY

This issue tracker is for the code contained within this repo -- `chef-client`, base `knife` functionality (not
plugins), `chef-apply`, `chef-solo`, `chef-client -z`, etc.
- [Server issues](https://github.com/chef/chef-server/issues/new)
- [ChefDK issues](https://github.com/chef/chef-dk/issues/new)
- Cookbook Issues (see the https://github.com/chef-cookbooks repos or search [Supermarket](https://supermarket.chef.io) or GitHub/Google)

antima-gupta:

## Description
- Added a separate linux mount provider.
- Define `mounted?` method and used `findmnt -rn` command to list mount points along with device.
- Added linux_mount recipe to kitchen-tests.

## Related Issue
Fixes #4660 

## Types of changes

## Checklist:


tas50:
findmnt is not a command on macOS so this is going to break those systems.

tas50:
@antima-gupta I'm also a bit concerned that no unit tests went red with this change since this won't work on FreeBSD or macOS. It might be worth looking at the unit tests for the provider and seeing if there's places in these methods where we can expand testing.

phiggins:
@tas50 Would this be appropriate for a functional/kitchen test?

tas50:
@antima-gupta The tests on Linux are great to have, but this change is still going to break non-Linux platforms like BSD/Solaris/macOS

antima-gupta:
@tas50 We have limited hardware so not able to find command which is compatible to `findmnt`  for platforms (BSD/Solaris/macOS). Could please give any suggestion to fix this issue for non-linux platforms (BSD/Solaris/macOS).

tas50:
@antima-gupta Without digging into this issue in great detail I can't offer you a particular command that will be fully cross platform, but as-is this will break a great number of chef users, so it's not a direction we can head down. It may be worth creating a overall unix mount provider and a linux mount provider which differ only in the commands here.

antima-gupta:
@tas50 I have pushed the latest changes, please have a look and let me know if any changes required.

spec/unit/provider/mount/linux_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    @new_resource.supports remount: false
+
+    @provider = Chef::Provider::Mount::Linux.new(@new_resource, @run_context)
+
+    allow(::File).to receive(:exists?).with("/dev/sdz1").and_return true
+    allow(::File).to receive(:exists?).with("/tmp/foo").and_return true
+    allow(::File).to receive(:realpath).with("/dev/sdz1").and_return "/dev/sdz1"
+    allow(::File).to receive(:realpath).with("/tmp/foo").and_return "/tmp/foo"
+  end
```
phiggins, line 23:
```suggestion

  let(:run_context) do
    node = Chef::Node.new
    events = Chef::EventDispatch::Dispatcher.new
    run_context = Chef::RunContext.new(node, {}, events)
  end

  let(:new_resource) do
    new_resource = Chef::Resource::Mount.new("/tmp/foo")
    new_resource.device      "/dev/sdz1"
    new_resource.device_type :device
    new_resource.fstype      "ext3"
    new_resource.supports remount: false
    new_resource
  end

  let(:provider) do
    Chef::Provider::Mount::Linux.new(@new_resource, @run_context)
  end

  before(:each) do
    allow(::File).to receive(:exists?).with("/dev/sdz1").and_return true
    allow(::File).to receive(:exists?).with("/tmp/foo").and_return true
    allow(::File).to receive(:realpath).with("/dev/sdz1").and_return "/dev/sdz1"
    allow(::File).to receive(:realpath).with("/tmp/foo").and_return "/tmp/foo"
  end
```

Using instance variables in tests is not a good pattern, we should use rspec stuff for this.

spec/unit/provider/mount/linux_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  context "to see if the volume is mounted" do
+
+    it "should set mounted true if the mount point is found in the mounts list" do
+      allow(@provider).to receive(:shell_out!).and_return(OpenStruct.new(stdout: "/tmp/foo /dev/sdz1 type ext3 (rw)\n"))
```
phiggins, line 28:
```suggestion
      allow(@provider).to receive(:shell_out!).and_return(double(stdout: "/tmp/foo /dev/sdz1 type ext3 (rw)\n"))
```

Since an OpenStruct instance will respond to any method, this can cause problems in the future when the code changes in unexpected ways. Instead the tests should be explicit about what behavior is expected of the test objects.

antima-gupta: Agreed, I have updated the changes.

spec/unit/provider/mount/linux_spec.rb
antima-gupta: Agreed, I have updated the changes.

antima-gupta:
@tas50 I have done with suggested changes, please let me know if any changes required.

lib/chef/provider/mount/linux.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          real_mount_point = if ::File.exists? @new_resource.mount_point
+                               ::File.realpath(@new_resource.mount_point)
+                             else
+                               @new_resource.mount_point
+                             end
+
+          shell_out!("findmnt -rn").stdout.each_line do |line|
+            case line
+            when /\A#{Regexp.escape(real_mount_point)}\s+#{device_mount_regex}\s/
```
jaymzh, line 44: This looks really nice. One thing that would be useful though would be to add a comment for each `when` that has an example line you're matching.

spec/unit/provider/mount/linux_spec.rb
```diff
@@ -0,0 +1,98 @@
+require "spec_helper"
+require "ostruct"
```
phiggins, line 2:
```suggestion
```

Not needed anymore.

lib/chef/provider/mount/linux.rb
tas50: Permalinks to rubular are super handy in this sort of case https://rubular.com/

lib/chef/provider/mount/linux.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        provides :mount, os: "linux"
+
+        # Check to see if the volume is mounted.
+        # "findmnt" outputs the mount points with volume.
+        # Convert the mount_point of the resource to a real path in case it
+        # contains symlinks in its parents dirs.
+
+        def mounted?
+          mounted = false
```
tas50, line 34: Instead of setting a mounted value you can just return true/false wherever you set mounted equal to true/false

kitchen-tests/cookbooks/end_to_end/recipes/linux.rb
```diff
@@ -122,6 +122,7 @@
 include_recipe "::_ohai_hint"
 include_recipe "::_openssl"
 include_recipe "::_tests"
+include_recipe "::_linux_mount"
```
tas50, line 125: Please rename this to _mount to keep the pattern of matching the resource name going. It's fine if we only use it on Linux for now. If we want to expand mount testing to other platforms we can add some platform logic there later.

tas50:
Overall this looks nice. It'll probably be much more reliable to use this method of determining mounts on linux.

lib/chef/provider/mount/linux.rb
phiggins: I was curious about that too, this is following the pattern of the method in the superclass that sets the value of the property at the end: https://github.com/chef/chef/blob/d0764e7849a69f5ce9b0a7bf24a7c0263a9f2c8e/lib/chef/provider/mount/mount.rb#L82

tas50: We should nuke that pattern in the main provider as well.

lib/chef/provider/mount/linux.rb
antima-gupta: Agreed, I have added Permalink.

spec/unit/provider/mount/linux_spec.rb
antima-gupta: I have removed it.

lib/chef/provider/mount/linux.rb
antima-gupta: @tas50 I am setting a mounted value because we need to iterate all lines in the mount point list then only we can identify the device is mounted or not.

kitchen-tests/cookbooks/end_to_end/recipes/linux.rb
antima-gupta: Agreed, I have updated this.

PR #10177, opened 2020-07-20
fix windows package for Chef-16 regression

lamont-granquist:
I didn't realize that the source_package was being used to
collect the merged state off of the property.


tas50:
Issue to add a test is here: https://github.com/chef/chef/issues/10179

Issue #10186, opened 2020-07-22
Lots of Chef::Log logspam output in the functional tests.


lamont-granquist:
https://buildkite.com/chef-oss/chef-chef-master-verify/builds/5295#a687e517-3c86-426f-8c40-5b625ce42e34

Note that didn't start in that PR, I rewound to the prior commit and I'm still seeing it.

I know I've seen a line of code somewhere in our spec helper stuff which was responsible for suppressing that output logging.

lamont-granquist:
I think this was what did it before?

https://github.com/chef/chef/blob/master/spec/support/chef_helpers.rb#L21-L27

I have no idea though why that wasn't being reset by the Chef::Config.reset call in the spec_helper.rb

lamont-granquist:
Ah so the Chef::Config lines probably became "no-ops" other than their local effects there when we started to reset the Chef::Config every time.  They can probably be deleted now.  The Chef::Log state was probably global and that's how the tests used to run before the started getting reset.  Although it doesn't seem to have started with pete's `Chef::Log.setup!` PR.

PR #10188, opened 2020-07-22
Set log level to :fatal in test to prevent excess logging.

phiggins:
Set the log level to :fatal for all tests after resetting the logger. Attempted fix for #10186.

### DIFF
--- a/spec/functional/knife/ssh_spec.rb
+++ b/spec/functional/knife/ssh_spec.rb
@@ -36,17 +36,6 @@ describe Chef::Knife::Ssh do
     allow(Net::SSH).to receive(:configuration_for).and_return(ssh_config)
   end
 
-  # Force log level to info.
-  around do |ex|
-    old_level = Chef::Log.level
-    begin
-      Chef::Log.level = :info
-      ex.run
-    ensure
-      Chef::Log.level = old_level
-    end
-  end
-
   describe "identity file" do
     context "when knife[:ssh_identity_file] is set" do
       before do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -243,6 +243,9 @@ RSpec.configure do |config|
 
     Chef::Log.setup!
 
+    Chef::Config[:log_level] = :fatal
+    Chef::Log.level(Chef::Config[:log_level])
+
     # By default, treat deprecation warnings as errors in tests.
     Chef::Config.treat_deprecation_warnings_as_errors(true)
 
--- a/spec/support/chef_helpers.rb
+++ b/spec/support/chef_helpers.rb
@@ -18,14 +18,6 @@ CHEF_SPEC_DATA = File.expand_path(File.dirname(__FILE__) + "/../data/")
 CHEF_SPEC_ASSETS = File.expand_path(File.dirname(__FILE__) + "/../functional/assets/")
 CHEF_SPEC_BACKUP_PATH = File.join(Dir.tmpdir, "test-backup-path")
 
-Chef::Config[:log_level] = :fatal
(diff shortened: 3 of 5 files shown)


spec/support/shared/functional/file_resource.rb
```diff
@@ -366,7 +366,6 @@ def real_file?(file_path)
   include_context "diff disabled"
 
   before do
-    Chef::Log.level = :info
```
phiggins, line 369: This seemed to be the culprit for all the log spam running the functional tests locally.

spec/functional/knife/ssh_spec.rb
```diff
@@ -36,17 +36,6 @@
...
-  # Force log level to info.
-  around do |ex|
-    old_level = Chef::Log.level
-    begin
-      Chef::Log.level = :info
-      ex.run
-    ensure
-      Chef::Log.level = old_level
-    end
-  end
```
phiggins, line 48: These tests didn't output anything when run alone so it didn't seem worth keeping this.

spec/support/chef_helpers.rb
```diff
@@ -18,14 +18,6 @@
 CHEF_SPEC_ASSETS = File.expand_path(File.dirname(__FILE__) + "/../functional/assets/")
 CHEF_SPEC_BACKUP_PATH = File.join(Dir.tmpdir, "test-backup-path")
 
-Chef::Config[:log_level] = :fatal
-Chef::Config[:persistent_queue] = false
-Chef::Config[:file_backup_path] = CHEF_SPEC_BACKUP_PATH
-
-Chef::Log.init(StringIO.new)
-Chef::Log.level(Chef::Config.log_level)
-Chef::Config.solo(false)
```
phiggins, line 27: I moved what was still relevant from this into the global `before` block in `spec_helper`, the rest is either set where it needs to be or didn't seem to have any effect.

phiggins:
The functional test output looks a lot better now: https://buildkite.com/chef-oss/chef-chef-master-verify/builds/5365#3bd27600-6f1b-42d3-8d19-12007a17fbb5

spec/functional/knife/ssh_spec.rb
lamont-granquist: yeah thats weird looking, if it didn't break the tests that's probably fine.

spec/spec_helper.rb
```diff
@@ -243,6 +243,9 @@ def self.from(system_exit)
 
     Chef::Log.setup!
 
+    Chef::Config[:log_level] = :fatal
+    Chef::Log.level(Chef::Config[:log_level])
```
lamont-granquist, line 247: `Chef::Config[:log_level]` gets reset on every `:each` down below this doesn't it?

spec/support/shared/functional/file_resource.rb
lamont-granquist: yep, global state is horrible

spec/support/chef_helpers.rb
lamont-granquist: 👍

spec/spec_helper.rb
phiggins: I added these two lines just below where `Chef::Config` and `Chef::Log` are reset. I'm fairly sure `Chef::Config[:log_level]` is being set because of the changes required to fix the test failures in `spec/unit/application_spec.rb` that started in this PR. https://github.com/chef/chef/pull/10188/files#diff-874b8ae1db0389e90526e97d2d53695b

lamont-granquist: ah right, i thought Chef::Log.setup! was outside the tight loop for some reason...

PR #10190, opened 2020-07-22
Add integration and unit tests back to omnibus tests in buildkite

### LINKED ISSUE #6708: Chef >= 13 gem_package on windows cannot find external gem command via `which` helper
## Description
The documentation for the `gem_package` resource, https://docs.chef.io/resource_gem_package.html , makes it sound like the `gem_binary` property is optional:

```
A property for the gem_package provider that is used to specify a gems binary.
By default, the same version of Ruby that is used by the chef-client will be installed.
```

I have a recipe with a block like this:

```
gem_package 'my_gem' do
  action :install
  source 'my-artifact-repository'
end
```

I can run kitchen converge fine with Chef client 12. But with Chef client 13, I get an error like this:

```
Chef::Exceptions::ValidationFailed: Property gem_binary must be one of String!  
You passed false.
```

I have tested with Chef client 13.2.20-1 and 13.4.19-1.

See also https://github.com/chef/chef-web-docs/issues/1169

## Chef Version

I have tested with Chef client 13.2.20-1 and 13.4.19-1.

## Platform Version

Windows Server 2012 R2, both workstation and Kitchen instances

## Replication Case

^^

## Client Output

```

```

lamont-granquist:
Adds integration and unit testing back to omnibus testers.

This discovered a bug in the gem_package provider that only occurs in an omnibus build (we do bundle-installed tests on windows without omnibus) and more generally with the which helper in chef-utils not being PATHEXT aware on windows at all.

That is the kind of reason why we need to be doing this properly from now on.

We are still not doing the chocolatey tests, although I installed chocolatey, but turning them on gave some super confusing windows error that I couldn't usefully google.  Bit worried there's a bug there.

Fixed some other cleanliness issues, went back to the progress formatting because the doc formatter is just too damn long to scroll down.

Also discovered some test issues around the ServerAPIVersions stuff which seem to crop up on boxes that are being chef-managed and have an /etc/chef/client.rb

The application_dot_d tests were similarly failing due to not being properly mocked on nodes that are being managed by chef.

The powershell tests now need to be turned off on windows os version < 10, we now have testing of versions less than 10 where for PRs we only test against 10.  This is correct since powershell_exec and the powershell resource only support >= 10.

The run_lock test and the parallelizer tests are turned off on mac os for now because our testers are horribly slow.  That is intended to be temporary and once we get faster testers that gets reverted.

This will still be red because we have red windows unit tests now so all the windows omnibus testers will be red with the same two test failures.

And it looks like windows 2012r2 32-bit, and ONLY 32-bit is still completely busted though.

fixes #6708


spec/unit/provider/package/powershell_spec.rb
```diff
@@ -19,7 +19,7 @@
 require "spec_helper"
 require "chef/mixin/powershell_out"
 
-describe Chef::Provider::Package::Powershell, :windows_only do
+describe Chef::Provider::Package::Powershell, :windows_only, :windows_gte_10 do
```
tas50, line 22: if these aren't working on older PS I think we have a bug. This should work pretty far back

lamont-granquist: those were busted just like the powershell_exec ones

lamont-granquist: https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/310 sending a new build through to verify

lamont-granquist: that blows up on all the 2012 testers with " Minimum installed PowerShell Version required is 5"

lamont-granquist: note this is the powershell package provider not the powershell script provider

PR #10185, opened 2020-07-21
Remove cookbook uploader global state

phiggins:
An attempt at fixing #9860.

Some of the cheffish tests fail intermittently on CI: https://buildkite.com/chef-oss/chef-chef-master-verify/builds/5295#cfb74720-f5be-455e-8ba4-3bde3ed4cd0a

Running these tests locally, the same 3 tests fail consistently (with an occasional success), either getting the 404 HTTP error or various errors about being unable to read files:

```
 1) Chef::Resource::ChefMirror When the Chef Zero server is in multi-org mode cookbook upload, chef_repo_path and versioned_cookbooks when the chef repo has cookbooks in non-versioned format chef_mirror :upload uploads everything
     Failure/Error:
       paths.collect!{|d| raise Errno::ENOENT, d unless File.exist?(d); d.dup}.each do |path|
         path = path.to_path if path.respond_to? :to_path
         enc = path.encoding == Encoding::US_ASCII ? fs_encoding : path.encoding
         ps = [path]
         while file = ps.shift
           catch(:prune) do
             yield file.dup.taint
             begin
               s = File.lstat(file)
             rescue Errno::ENOENT, Errno::EACCES, Errno::ENOTDIR, Errno::ELOOP, Errno::ENAMETOOLONG

     Errno::ENOENT:
       chef_mirror[] (basic_chef_client::block line 242) had an error: Errno::ENOENT: No such file or directory - /var/folders/3d/kw7h7x0s39jdbhdyycl4jp_40000gn/T/tmp_working_dir_path20200721-75275-85siyh/y-3.0.0/metadata.rb
     # /Users/pete/work/chef/lib/chef/cookbook/cookbook_version_loader.rb:224:in `block in load_all_files'
     # /Users/pete/work/chef/lib/chef/cookbook/cookbook_version_loader.rb:215:in `each'
     # /Users/pete/work/chef/lib/chef/cookbook/cookbook_version_loader.rb:215:in `load_all_files'
     # /Users/pete/work/chef/lib/chef/cookbook/cookbook_version_loader.rb:73:in `load!'
     # /Users/pete/work/chef/lib/chef/cookbook_loader.rb:98:in `load_cookbook'
     # /Users/pete/work/chef/lib/chef/cookbook_loader.rb:80:in `block in load_cookbooks'
     # /Users/pete/work/chef/lib/chef/cookbook_loader.rb:79:in `each_key'
[...]

PR #10192, opened 2020-07-23
Initial 16.3 release notes

tas50:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,83 @@
...
+- `apt-cdrom` repositories are now skipped when parsing out the list of apt repositories
+- Faulty profiles are now reported instead of causing a crash
+- Errors are no longer logged to stdout with the `html2` reporter
+- macOS Big Sur is now correctly identified as macOS
+
+## New Resources
+
+### windows_firewall_profile
+
+The `windows_firewall_profile` allows you to `enable`, `disable`, or `configure` Windows Firewall profiles. For example you can now setup default actions and configure fules for a the `Public` profile using this single resource instead of managing your own PowerShell code in a `powershell_script` resource:
```
phiggins, line 19:
```suggestion
The `windows_firewall_profile` allows you to `enable`, `disable`, or `configure` Windows Firewall profiles. For example you can now setup default actions and configure rules for the `Public` profile using this single resource instead of managing your own PowerShell code in a `powershell_script` resource:
```

`fules` is not getting recognized as a typo because it is in `chef_dictionary.txt`. Is that intentional? What are `fules`?

phiggins:
#10185 was done to fix flaky CI failures, but I believe it was a thread safety bug and if users have used that code like how we did in the tests someone has likely run into it.

tas50:
@phiggins I added a vague reference to that fix.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+### windows_firewall_profile
+
+The `windows_firewall_profile` allows you to `enable`, `disable`, or `configure` Windows Firewall profiles. For example you can now setup default actions and configure rules for the `Public` profile using this single resource instead of managing your own PowerShell code in a `powershell_script` resource:
+
+```ruby
+windows_firewall_profile 'Public' do
+  default_inbound_action 'Block'
+  default_outbound_action 'Allow'
+  allow_inbound_rules 'False'
+  display_notification 'False'
```
chef-davin, line 26: The final version of this had this using `true` and `false` rather than the Strings.  Though I realize I didn't update the examples as I thought I had, so I'm putting in a PR to fix that now, but it should be fixed here as well.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+- `apt-cdrom` repositories are now skipped when parsing out the list of apt repositories
+- Faulty profiles are now reported instead of causing a crash
+- Errors are no longer logged to stdout with the `html2` reporter
+- macOS Big Sur is now correctly identified as macOS
+
+## New Resources
+
+### windows_firewall_profile
+
+The `windows_firewall_profile` allows you to `enable`, `disable`, or `configure` Windows Firewall profiles. For example you can now setup default actions and configure rules for the `Public` profile using this single resource instead of managing your own PowerShell code in a `powershell_script` resource:
```
chef-davin, line 19: I think this might be better as "and" `configure` rather than "or".  Since the configuration is part of the `:enable` action, and no longer a separate action on its own.

RELEASE_NOTES.md
chef-davin: PR #10198

tas50: If you left the properties empty wouldn't you get just an enable?

RELEASE_NOTES.md
tas50: If you left the properties empty wouldn't you get just an enable?

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+The `git` resource will no longer fail if syncing a branch that already exists locally. Thanks for fixing this [@lotooo](https://github.com/lotooo/).
+
+### macos_user_defaults
+
+The `macos_user_defaults` has received a ground-up refactoring with new actions, additional properties, and better overall reliability:
+
+  - Improved idempotency by properly loading the current state of domains.
+  - Improved how we set `dict` and `array` type data.
+  - Improved logging to show the existing key/value pair that are being changed and improved the property state data we send to handlers and/or Chef Automate.
```
IanMadd:
```suggestion
  - Improved logging to show the existing key/value pair that is changed, and improved the property state data that the resource sends to handlers and/or Chef Automate.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+  - Fixed a failure when setting keys or values that included a space.
+  - Replaced the existing non-functional `global` property with a new default for the `domain` property. To set a key/value on the `NSGlobalDomain` domain you can either set that value explicitly or just skip `domain` property entirely and Chef Infra Client will default to `NSGlobalDomain`. The existing property has been marked as deprecated and we will ship a Cookstyle rule to detect cookbooks using this property in the future.
```
IanMadd:
```suggestion
  - Replaced the existing non-functional `global` property with a new default for the `domain` property. To set a key/value pair on the `NSGlobalDomain` domain, you can either set that value explicitly or just skip the `domain` property entirely and Chef Infra Client will default to `NSGlobalDomain`. The existing property has been marked as deprecated and we will ship a Cookstyle rule to detect cookbooks using this property in the future.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+  - Fixed the type property to only accepts valid inputs. Previously typos or otherwise incorrect values would just be ignored, resulting in unexpected behavior. This may cause failures in your codebase if you previously used incorrect values, and we will be shipping a Cookstyle rule to detect and correct these values in the future.
```
IanMadd:
```suggestion
  - Fixed the `type` property to only accept valid inputs. Previously, typos or otherwise incorrect values would just be ignored resulting in unexpected behavior. This may cause failures in your codebase if you previously used incorrect values. We will be shipping a Cookstyle rule to detect and correct these values in the future.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+ - Added a new `delete` action to allow users to remove a key from a domain.
+ - Added a new host property that lets you set per-host values. If you set this to :current it sets the -currentHost flag.
```
IanMadd:
```suggestion
 - Added a new `host` property that lets you set per-host values. If you set this to `:current` it sets the -currentHost flag.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+The `windows_dns_record` resource includes a new optional property, `dns_server`, allowing you to make changes against remote servers. Thanks for this addition [@jeremyciak](https://github.com/jeremyciak/).
+
+### windows_package
+
+A Chef Infra Client 16 regression within `windows_package` that prevented specifying `path` in the `remote_file_attributes` property has been resolved. Thanks for reporting this issue [@asvinours](https://github.com/asvinours/)
+
+### windows_security_policy
+
+The `windows_security_policy` resource has been refactored to improve idempotency and improve log output when changes are made. You'll now see more complete change information in logs and any handler consuming this data will also receive more detailed change information.
```
IanMadd, line 74:
```suggestion
The `windows_security_policy` resource has been refactored to improve idempotency and improve log output when changes are made. You will now see more complete change information in logs and any handler consuming this data will also receive more detailed change information.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+## Knife Improvements
+
+- Ctrl-C can now be used to exit knife even when being prompted for input.
+- `knife bootstrap` will now properly error if attempting to bootstrap an AIX system using an account with an expired password.
+- `knife profile` commands will no longer error if an invalid profile was previously set.
+- The `-o` flag for `knife cookbook upload` can now be used on Windows systems.
+- `knife ssh` now once again accepts legacy DSS host keys although we highly recommend upgrading to a more secure key algorithm if possible.
```
IanMadd, line 82:
```suggestion
- `knife ssh` now once again accepts legacy DSS host keys, although we highly recommend upgrading to a more secure key algorithm if possible.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+- Several changes were made to knife to that may prevent intermittent failures running cookbook commands
```
IanMadd, line 83:
```suggestion
- Several changes were made to Knife that may prevent intermittent failures while running cookbook commands.
```

RELEASE_NOTES.md
chef-davin: yes. that's true.

RELEASE_NOTES.md
tas50: Fixed

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+## Habitat Package Improvements
+
+Habitat packages for Windows, Linux and Linux2 are now built and tested against each pull request to Chef Infra Client. Additionally we've improved how these packages are built to reduce the size of the package, which reduces network utilization when using the Effortless deployment pattern.
```
IanMadd, line 87:
```suggestion
Habitat packages for Windows, Linux, and Linux2 are now built and tested against each pull request to Chef Infra Client. Additionally, we've improved how these packages are built to reduce the size of each package. This reduces network utilization when using the Effortless deployment pattern.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+The `git` resource will no longer fail if syncing a branch that already exists locally. Thanks for fixing this [@lotooo](https://github.com/lotooo/).
+
+### macos_user_defaults
+
+The `macos_user_defaults` has received a ground-up refactoring with new actions, additional properties, and better overall reliability:
+
+- Improved idempotency by properly loading the current state of domains.
+- Improved how we set `dict` and `array` type data.
+- Improved logging to show the existing key/value pair that are being changed and improved the property state data we send to handlers and/or Chef Automate.
```
IanMadd, line 57:
```suggestion
- Improved logging to show the existing key/value pair that is changed, and improved the property state data that the resource sends to handlers and/or Chef Automate.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+- Fixed a failure when setting keys or values that included a space.
+- Replaced the existing non-functional `global` property with a new default for the `domain` property. To set a key/value on the `NSGlobalDomain` domain you can either set that value explicitly or just skip `domain` property entirely and Chef Infra Client will default to `NSGlobalDomain`. The existing property has been marked as deprecated and we will ship a Cookstyle rule to detect cookbooks using this property in the future.
```
IanMadd, line 59:
```suggestion
- Replaced the existing non-functional `global` property with a new default for the `domain` property. To set a key/value pair on the `NSGlobalDomain` domain, you can either set that value explicitly or just skip `domain` property entirely and Chef Infra Client will default to `NSGlobalDomain`. The existing property has been marked as deprecated and we will ship a Cookstyle rule to detect cookbooks using this property in the future.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+- Fixed the type property to only accepts valid inputs. Previously typos or otherwise incorrect values would just be ignored, resulting in unexpected behavior. This may cause failures in your codebase if you previously used incorrect values, and we will be shipping a Cookstyle rule to detect and correct these values in the future.
```
IanMadd, line 60:
```suggestion
- Fixed the `type` property to only accept valid inputs. Previously typos or otherwise incorrect values would just be ignored resulting in unexpected behavior. This may cause failures in your codebase if you previously used incorrect values. We will be shipping a Cookstyle rule to detect and correct these values in the future.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+- Added a new `delete` action to allow users to remove a key from a domain.
+- Added a new host property that lets you set per-host values. If you set this to :current it sets the -currentHost flag.
```
IanMadd, line 62:
```suggestion
- Added a new `host` property that lets you set per-host values. If you set this to `:current` it sets the -currentHost flag.
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,91 @@
...
+
+### windows_dns_record
+
+The `windows_dns_record` resource includes a new optional property, `dns_server`, allowing you to make changes against remote servers. Thanks for this addition [@jeremyciak](https://github.com/jeremyciak/).
+
+### windows_package
+
+A Chef Infra Client 16 regression within `windows_package` that prevented specifying `path` in the `remote_file_attributes` property has been resolved. Thanks for reporting this issue [@asvinours](https://github.com/asvinours/)
```
IanMadd, line 70:
```suggestion
A Chef Infra Client 16 regression within `windows_package` that prevented specifying `path` in the `remote_file_attributes` property has been resolved. Thanks for reporting this issue [@asvinours](https://github.com/asvinours/).
```

RELEASE_NOTES.md
tas50: Fixed

RELEASE_NOTES.md
tas50: Fixed

PR #10193, opened 2020-07-23
Implement default_paths API

lamont-granquist:


chef-config/lib/chef-config/config.rb
```diff
@@ -313,9 +313,12 @@ def self.derive_path_from_chef_repo_path(child_path)
     # Defaults to <chef_repo_path>/users.
     default(:user_path) { derive_path_from_chef_repo_path("users") }.writes_value { |path| expand_relative_paths(path) }
 
-    # Turn on "path sanity" by default.
+    # DEPRECATED
     default :enforce_path_sanity, false
 
+    # Enforce default paths by default for all APIs, not just the default internal shell_out
+    default :enforce_default_paths, false
```
phiggins, line 320: Should we make this one default to the value of the old one, in case someone is already using the old setting?

lamont-granquist: at some point they gotta go fix their settings by the time we remove this, so i think not.

the code itself does `Chef::Config[:enforce_default_paths] || Chef::Config[:enforce_path_sanity]` where it matters.

we'll need to throw a deprecation if the old one is set and the new one isn't -- copying the value makes that difficult.

tas50: Is the point of using the || everywhere just to make it easier to toss a deprecation in?

lamont-granquist: Yeah throw a `Chef::Log.deprecated "whatever" if Chef::Config[:enforce_path_sanity]` above it.

Note that I've seen VERY few people that have found this API over the years.  It tends to break more stuff than it fixes because it also injects the embedded/bin dir so everyone's ruby apps or whatever would start picking up chef's ruby, etc.

A bigger question is if we should just deprecate this config switch entirely

Issue #9913, opened 2020-05-26
Add new 'json_file' resource


mattray:
A resource for writing out JSON files would replace many of the use of templates or the `file` resource to write these out. It would be an extension of the `file` resource, replacing the `content` with a hash and keeping all the existing properties.

## Motivation

As a user, I want to pass a hash of attributes to a `json_file` resource and have it render them appropriately.

## Specification

```
a = {
    "golf": "hotel",
    "kilo": ["lima", "mike", "november"],
    "india": {
        "juliett": true
        },
    "alpha": {
        "charlie": "delta",
        "bravo": 10,
        },
    "echo": "foxtrot"
    }

json_file '/tmp/file.json' do
    content a
end
```
produces the file `/tmp/file.json`
```
{
  "alpha": {
    "bravo": 10,
    "charlie": "delta"
  },
  "echo": "foxtrot",
  "golf": "hotel",
  "india": {
    "juliett": true
  },
  "kilo": [
    "lima",
    "mike",
    "november"
  ]
}
```

lamont-granquist:
closing to track in #9911

tas50:
a new helper for handling this has been added for 17.3

Issue #9912, opened 2020-05-26
Add new 'yaml_file' resource


mattray:
A resource for writing out YAML files would replace many of the use of templates or the `file` resource to write these out from strings. It would be an extension of the `file` resource, replacing the `content` with a hash and keeping all the existing properties.

## Motivation

As a user, I want to pass a hash of attributes to a `yaml_file` resource and have it render them appropriately.

## Specification

```
a = {
    'golf': 'hotel',
    'kilo': ['lima', 'mike', 'november'],
    'india': {
        'juliett': true
        },
    'alpha': {
        'charlie': 'delta',
        'bravo': 10,
        },
    'echo': 'foxtrot'
    }

yaml_file '/tmp/file.yaml do
    content a
end
```

produces the file `/tmp/file.yaml`
```
---
golf: hotel
kilo:
- lima
- mike
- november
india:
  juliett: true
alpha:
  charlie: delta
  bravo: 10
echo: foxtrot
```

```

mattray:
Example implementation with tests here: https://github.com/mattray/yaml-cookbook

lamont-granquist:
closing to track in #9911

tas50:
a new helper for handling this has been added for 17.3

Issue #9911, opened 2020-05-26
Add new 'toml_file' resource


mattray:
A resource for writing out [TOML](https://github.com/toml-lang/toml) files would replace many of the use of templates or the `file` resource to write these out. It would be an extension of the `file` resource, replacing the `content` with a hash and keeping all the existing properties.

## Motivation

As a user, I want to pass a hash of attributes to a `toml_file` resource and have it render them appropriately.

## Specification

```
a = {
    'golf': 'hotel',
    'india': {
        'juliett': true
        },
    'alpha': {
        'charlie': 'delta',
        'bravo': 10,
        },
    'echo': 'foxtrot'
    }

toml_file '/tmp/file.toml' do
    content a
end
```

produces the file `/tmp/file.toml`
```
echo = "foxtrot"
golf = "hotel"
[alpha]
bravo = 10
charlie = "delta"
[india]
juliett = true
```

mattray:
Example implementation with tests here: https://github.com/mattray/toml-cookbook

lamont-granquist:
If you're going to do this it'd be better to actually subclass the File class for this one, which eliminates all of the boilerplate code.

It also feels like this is a bit too trivial, you've got literally one line of code in here.  While the cost is that this is yet another resource in the docs.  I'm not sure that's worth it just so people don't have to write `content TomlRB.dump(content)`.

I think I'd rather push that into a property of the file resource itself (toml_content or whatever).  Although I'm not sure how that'd behave when inherited by the remote_file resource and I just woke up...

lamont-granquist:
I think I'd rather seem some kind of DSL sugar added:

```
hash = { ... }
file "/etc/whatever.xyz" do
  content render_toml(hash)
end
```

mattray:
I agree that the implementation for this (and the YAML and JSON variants) will all subclass File, they're intended to be "batteries included" approaches. I wanted to avoid "DSL sugar" because that implies that the user has to know about Ruby. My goal was for a user to populate some attributes in a Policyfile and have them rendered as TOML.

lamont-granquist:
> because that implies that the user has to know about Ruby

Except that the `render_toml` call in my example isn't any more ruby than your file resources are.

lamont-granquist:
I'm really very, very hard "NO" on doing this any way other than the DSL helper to be abundantly clear.

lytao:
[...]

PR #10194, opened 2020-07-23
New render_json, render_toml, render_yaml helper methods

mattray:
Provide new syntactic sugar methods for rendering JSON, TOML, and YAML content from hashes.

Related to 
https://github.com/chef/chef/issues/9911 
https://github.com/chef/chef/issues/9912 
https://github.com/chef/chef/issues/9913

The toml code is migrated from the Habitat cookbook: https://github.com/chef-cookbooks/habitat/blob/master/libraries/habitat_toml.rb

Signed-off-by: Matt Ray <github@mattray.dev>

### DIFF
--- a/chef-utils/lib/chef-utils/dsl/render_helpers.rb
+++ b/chef-utils/lib/chef-utils/dsl/render_helpers.rb
@@ -1,0 +1,49 @@
+#
+# Copyright:: Copyright (c) Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require_relative "../internal"
+require_relative "../toml"
+
+require 'json'
+require 'yaml'
+
+module ChefUtils
+  module DSL
+    module RenderHelpers
+      include Internal
+
+      # pretty-print a hash as a JSON string
+      def render_json(hash)
+        JSON.pretty_generate(hash) + "\n"
+      end
+
+      # pretty-print a hash as a TOML string
+      def render_toml(hash)
+        ChefUtils::Toml::Dumper.new(hash).toml_str
+      end
+
+      # pretty-print a hash as a YAML string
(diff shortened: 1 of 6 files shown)


chef-utils/lib/chef-utils/dsl/render_helpers.rb
```diff
@@ -0,0 +1,49 @@
...
+      # pretty-print a hash as a TOML string
+      def render_toml(hash)
+        ChefUtils::Toml::Dumper.new(hash).toml_str
+      end
+
+      # pretty-print a hash as a YAML string
+      def render_yaml(hash)
+        yaml_content = hash.transform_keys(&:to_s).to_yaml
+        #above replaces first-level keys with strings, below the rest
+        yaml_content.gsub!(' :',' ')
```
phiggins, line 43: This does not seem like a safe thing to do on arbitrary data. In ruby it is straighforward to recursively convert a hash's keys to strings though:

```
def render_yaml(hash)
  converter = lambda do |object|
    case object
    when Hash
      object.transform_keys(&:to_s).transform_values(&converter)
    else
      object
    end
  end

  converter.call(hash).to_yaml
end
```

chef-utils/spec/spec_helper.rb
```diff
@@ -21,6 +22,7 @@
 OS_HELPERS = (ChefUtils::DSL::OS.methods - Module.methods).freeze
 PLATFORM_FAMILY_HELPERS = (ChefUtils::DSL::PlatformFamily.methods - Module.methods).freeze
 PLATFORM_HELPERS = (ChefUtils::DSL::Platform.methods - Module.methods).freeze
+RENDER_HELPERS = (ChefUtils::DSL::RenderHelpers.methods - Module.methods).freeze
```
phiggins, line 25: I know this is copying an existing pattern, but I feel like it's kind of overkill to define a constant in spec_helper that is used in one test file.

lamont-granquist: yeah i did it centralized though because otherwise people will probably copypasta around the constants like I did initially and defining them on the top level will get constant redefinition warnings and will break stuff.

lamont-granquist: (it ain't good, but the alternative i had first was worse and this at least makes it obvious you need unique names, but that's about all)

chef-utils/lib/chef-utils/dsl/render_helpers.rb
lamont-granquist: probably needs a `when Array` that recursively calls the converter on all the elements.

i think the api probably needs to take an optional flag to turn this convert off as well in case users want to actually render symbols.

tas50:
@mattray ping?

tas50:
@mattray I'm going to close this out at this point. If you still want to make this happen that'd be great. New PR it and we'll move forward.

PR #10205, opened 2020-07-24
Remove some test cruft around diffing

phiggins:


lamont-granquist:
i think i recognize some of this code...

PR #10207, opened 2020-07-27
Combine attr_readers onto a single line

tas50:
The method takes more than 1 argument


tas50:
Closing as this prevents us from documenting these later

Issue #10211, opened 2020-07-27
Chef 16.3.38 on macOS has a NoMethodError


erikng:
## Description
I can't get chef to run

## Chef Version
Chef 16.3.38

## Platform Version
macOS 10.16.0 Beta 3

## Replication Case
Install chef, run chef

## Client Output
```
Starting Chef Infra Client, version 16.3.38
[2020-07-27T13:54:47-05:00] INFO: *** Chef Infra Client 16.3.38 ***
[2020-07-27T13:54:47-05:00] INFO: Platform: x86_64-darwin17
[2020-07-27T13:54:47-05:00] INFO: Chef-client pid: 34423

================================================================================
Chef encountered an error attempting to load the node data for "snipped"
================================================================================

Unexpected Error:
-----------------
NoMethodError: undefined method `possible_requests' for nil:NilClass

System Info:
------------
chef_version=16.3.38
ruby=ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin17]
program_name=/usr/local/bin/chef-client
executable=/opt/chef/bin/chef-client


Running handlers:
[2020-07-27T13:54:50-05:00] ERROR: Running exception handlers
[2020-07-27T13:54:50-05:00] INFO: Could not get failed cookbook from backtrace
[2020-07-27T13:54:50-05:00] INFO: Could not get failed rescipe from backtrace
[2020-07-27T13:54:50-05:00] INFO: Could not get failed line from backtrace
[2020-07-27T13:54:50-05:00] ERROR: Report handler Chef::Handler::Runstats raised #<NoMethodError: undefined method `[]' for nil:NilClass>
[2020-07-27T13:54:50-05:00] ERROR: /etc/chef/handlers/chef_runstats.rb:99:in `report'
[2020-07-27T13:54:50-05:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.3.38/lib/chef/handler.rb:256:in `run_report_unsafe'
[2020-07-27T13:54:50-05:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.3.38/lib/chef/handler.rb:244:in `run_report_safely'
[2020-07-27T13:54:50-05:00] ERROR: /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.3.38/lib/chef/handler.rb:149:in `block in run_exception_handlers'
[...]

PR #10216, opened 2020-07-27
Fix NoMethodError when server returns a 406.

phiggins:
Fix for #10211.

### DIFF
--- a/lib/chef/http.rb
+++ b/lib/chef/http.rb
@@ -58,7 +58,6 @@ class Chef
           handler.handle_chunk(chunk)
         end
       end
-
     end
 
     def self.middlewares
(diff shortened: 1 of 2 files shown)


lib/chef/http.rb
```diff
@@ -471,7 +470,11 @@ def retrying_http_errors(url)
     end
 
     def version_retries
-      @version_retries ||= options[:version_class].possible_requests
+      @version_retries ||= if options[:version_class]
+                             options[:version_class].possible_requests
+                           else
+                             0
```
lamont-granquist, line 476: not sure this is correct because we could be on a server and client (in the future) that have both dropped support for v0

phiggins: I'm obviously not the expert on this, but this code is not returning the version of the protocol, it's returning the number of times to retry the request:

https://github.com/chef/chef/blob/09cdc96cb25956b3ef053fbe093d06453a7af92b/lib/chef/http.rb#L157

I set it to 0 to indicate "do not retry". Like I said in slack this is not great because you're still getting an error (HTTP 406), just not the one in our code (NoMethodError).

PR #10219, opened 2020-07-28
Update mixlib-shellout to 3.1.2

tas50:


phiggins:
`Kitchen: CentOS 6` is blowing up.

PR #10232, opened 2020-07-29
Fix warning in test for redefined shared example group

phiggins:
Because I get annoyed seeing this warning whenever I run tests:

```
WARNING: Shared example group 'with a chef repo' has been previously defined at:
  /Users/pete/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0/gems/cheffish-16.0.2/lib/cheffish/rspec/repository_support.rb:12
...and you are now defining it at:
  /Users/pete/work/chef/spec/support/shared/integration/integration_helper.rb:109
The new definition will overwrite the original one.
```

90% of the lines changed in this PR are indentation changes so viewing this with whitespace changes turned off will likely make this much more readable: https://github.com/chef/chef/pull/10232/files?w=1

### DIFF
--- a/spec/integration/knife/config_get_profile_spec.rb
+++ b/spec/integration/knife/config_get_profile_spec.rb
@@ -22,92 +22,93 @@ describe "knife config get-profile", :workstation do
   include KnifeSupport
 
   include_context "default config options"
-  include_context "with a chef repo"
 
   let(:cmd_args) { [] }
 
-  subject do
-    cmd = knife("config", "get-profile", *cmd_args, instance_filter: lambda { |instance|
-      # Fake the failsafe check because this command doesn't actually process knife.rb.
-      $__KNIFE_INTEGRATION_FAILSAFE_CHECK << " ole"
-    })
-    cmd.stdout
-  end
-
-  around do |ex|
-    # Store and reset the value of some env vars.
-    old_chef_home = ENV["CHEF_HOME"]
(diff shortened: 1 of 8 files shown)


spec/integration/knife/config_get_profile_spec.rb
```diff
@@ -22,92 +22,93 @@
   include KnifeSupport
 
   include_context "default config options"
-  include_context "with a chef repo"
```
tas50, line 25: 😄

spec/integration/knife/config_get_profile_spec.rb
```diff
@@ -22,92 +22,93 @@
...
-      old_chef_profile = ENV["CHEF_PROFILE"]
+      # Store and reset the value of some env vars.
+      old_chef_home = ENV["CHEF_HOME"]
+      old_knife_home = ENV["KNIFE_HOME"]
+      old_home = ENV["HOME"]
+      old_wd = Dir.pwd
+      ChefConfig::PathHelper.per_tool_home_environment = "KNIFE_HOME"
+      # Clear these out because they are cached permanently.
+      ChefConfig::PathHelper.class_exec { remove_class_variable(:@@home_dir) }
+      Chef::Knife::ConfigGetProfile.reset_config_loader!
```
phiggins, line 46: All of this state, `ENV`, `pwd`, `PathHelper`, should probably get reset in between tests but that's a dragon to slay some other time.

Issue #10099, opened 2020-07-02
Resources with `ignore_failure true` don't report failure details to A2


trickyearlobe:
# Description

An enterprise customer reported that if a resource in a Chef run fails, but it has `ignore_failure true` the error doesn't show up in the A2 UI as a stack trace.

![image](https://user-images.githubusercontent.com/3001322/86259384-dc6ea400-bbb3-11ea-8d3b-f55b2125c142.png)

In the screenshot you can see one resource has a `Error Log` button against it, but when clicked, the stack trace is empty.

## Chef Version
15.6.18

## Platform Version
Windows 10.0.14393

## Replication Case

- Run a cookbook with a failing resource that has `ignore_failure true` set.
- Look at the list of updated resources and see one has an `Error Log` button.
- Click the `Error Log` button and observe the empty stack trace.
- Observe the stack trace is visible in the Chef client logs
- Observer the client run completes successfully.

## The actual stacktrace from the example

```
    ================================================================================
    Error executing action `run` on resource 'powershell_script[Disbale administrator]'
    ================================================================================

    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    Expected process to exit with [0], but received '2'
    ---- Begin output of "C:\windows\system32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NonInteractive -NoProfile
-ExecutionPolicy Bypass -InputFormat None -File "C:/Users/t635414/AppData/Local/Temp/chef-script20200213-7040-iblr4x.ps1
" ----
    STDOUT:
    STDERR: The user name could not be found.
    More help is available by typing NET HELPMSG 2221.
    ---- End output of "C:\windows\system32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NonInteractive -NoProfile -E
xecutionPolicy Bypass -InputFormat None -File "C:/Users/t635414/AppData/Local/Temp/chef-script20200213-7040-iblr4x.ps1"
----
    Ran "C:\windows\system32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NonInteractive -NoProfile -ExecutionPolicy
Bypass -InputFormat None -File "C:/Users/t635414/AppData/Local/Temp/chef-script20200213-7040-iblr4x.ps1" returned 2

    Resource Declaration:
    ---------------------
    # In c:/chef/cache/cookbooks/cis-windows-ms-2016/recipes/account_and_local_policies.rb

[...]

PR #10237, opened 2020-07-31
client-run per resource error detail

vsingh-msys:
## Description

- Resources with `ignore_failure true` error backtrace and detail are not populating to A2.
- In order to solve this add error_description to action_collection of the resource-based object.
- Add error hash message class, message, and detail to end-run message resource object.

## Related Issue

https://github.com/chef/chef/issues/10099

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/action_collection.rb
+++ b/lib/chef/action_collection.rb
@@ -44,6 +44,9 @@ class Chef
       # @return [Exception] The exception that was thrown
       attr_accessor :exception
 
+      # @return [Hash] JSON-formatted error description from the Chef::Formatters::ErrorMapper
+      attr_accessor :error_description
+
       # @return [Numeric] The elapsed time in seconds with machine precision
       attr_accessor :elapsed_time
 
@@ -223,6 +226,7 @@ class Chef
 
       current_record.status = :failed
       current_record.exception = exception
+      current_record.error_description = Formatters::ErrorMapper.resource_failed(new_resource, action, exception).for_json
     end
 
     # Hook called after an action is completed.  This is always called, even if the action fails.
--- a/lib/chef/data_collector/run_end_message.rb
+++ b/lib/chef/data_collector/run_end_message.rb
@@ -135,6 +135,8 @@ class Chef
           hash["conditional"] = action_record.conditional.to_text if action_record.status == :skipped
 
           unless action_record.exception.nil?
+            hash["error_message"] = action_record.exception.message
+
             hash["error"] = {
               "class" => action_record.exception.class,
               "message" => action_record.exception.message,
--- a/lib/chef/data_collector/run_end_message.rb
+++ b/lib/chef/data_collector/run_end_message.rb
@@ -133,7 +133,15 @@ class Chef
           end
 
           hash["conditional"] = action_record.conditional.to_text if action_record.status == :skipped
-          hash["error_message"] = action_record.exception.message unless action_record.exception.nil?
+
+          unless action_record.exception.nil?
+            hash["error"] = {
+              "class" => action_record.exception.class,
+              "message" => action_record.exception.message,
+              "backtrace" => action_record.exception.backtrace,
+              "description" => action_record.error_description,
+            }
+          end
 
           hash
         end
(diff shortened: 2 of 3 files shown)


lib/chef/data_collector/run_end_message.rb
```diff
@@ -133,7 +133,15 @@ def action_record_for_json(action_record)
           end
 
           hash["conditional"] = action_record.conditional.to_text if action_record.status == :skipped
-          hash["error_message"] = action_record.exception.message unless action_record.exception.nil?
```
vsingh-msys, line 136: Deliberately remove `error_message` attribute as going to add more error info.

lamont-granquist: That probably breaks A2 though.  Even if it doesn't this is a public API that anyone can use, and that breaks backwards compatibility of the API.  This needs to be backed out and included.

vsingh-msys:
@lamont-granquist sure will leave it as it is. thanks!

lamont-granquist:
unit tests around the data collector are broken now

lib/chef/action_collection.rb
```diff
@@ -223,6 +226,7 @@ def resource_failed(new_resource, action, exception)
 
       current_record.status = :failed
       current_record.exception = exception
+      current_record.error_description = Formatters::ErrorMapper.resource_failed(new_resource, action, exception).for_json
```
tas50, line 229: Any reason we don't just use this:

https://github.com/chef/chef/blob/6e73617243baa21815258f9b51352e9f3753af1d/lib/chef/data_collector/error_handlers.rb#L110

vsingh-msys: We don't have data collector object reference, also `error_description` gets assigned with the call of `resource_failed`.

please let me know if I am not getting it correctly? Thanks

spec/unit/data_collector_spec.rb
```diff
@@ -39,6 +39,7 @@
   let(:new_resource) do
     new_resource = Chef::Resource::File.new("/tmp/a-file.txt")
     new_resource.checksum nil
+    new_resource.path
```
vsingh-msys, line 42: Not sure why I need this magically this fixes all failing specs with `error_descrition` contains path attributes as well

vsingh-msys: Without `new_resource.path` define let object `error_description` return

```
{"title"=>"Error executing action `create` on resource 'file[/tmp/a-file.txt]'", "sections"=>[{"Exception"=>"imperial to metric conversion error"}, {"Compiled Resource:"=>"# Declared in \n\nfile(\"/tmp/a-file.txt\") do\n  action [:create]\n  default_guard_interpreter :default\n  checksum nil\n  cookbook_name \"monkey\"\n  recipe_name \"atlas\"\n  owner nil\n  group nil\n  mode nil\n  verifications []\n  path \"/tmp/a-file.txt\"\nend\n"}]}
```

After `new_resource.path` seems to be expected json values

```
{"title"=>"Error executing action `create` on resource 'file[/tmp/a-file.txt]'", "sections"=>[{"Exception"=>"imperial to metric conversion error"}, {"Compiled Resource:"=>"# Declared in \n\nfile(\"/tmp/a-file.txt\") do\n  action [:create]\n  default_guard_interpreter :default\n  checksum nil\n  path \"/tmp/a-file.txt\"\n  cookbook_name \"monkey\"\n  recipe_name \"atlas\"\n  owner nil\n  group nil\n  mode nil\n  verifications []\nend\n"}]}

```

lib/chef/action_collection.rb
lamont-granquist: Yeah actually that's correct, the data_collector isn't responsible for per-resource analysis and while it gets the error_description after resource_failed() is called, it doesn't do anything with it, and as it collects more ignored errors during the run it just discards them.  To collect the errors in the data_collector for the actions we'd have to reimplement chunks of the action_collection there (violating SRP and defeating the purpose of extracting the action_collection).  So it does make more sense to collection the error descriptions in the action_record like this.

spec/unit/data_collector_spec.rb
lamont-granquist: That's due to the fact that the @path instance variable doesn't exist until the first time that `path` is called on the object where it sees that it is the name attribute so the getter pulls the value from the name attribute and sets itself.

lamont-granquist: Ultimately there seems to be a bit of a bug in Chef::Resource#to_text around handling of the name property when it hasn't been accessed yet.

PR #10239, opened 2020-07-31
Convert osx_profile to custom resource

lamont-granquist:
This also does some code cleanup.

The path property has been removed as a property since it was never
an input to the resource, and just some state that was necessary
to abuse the tempfile mechanism.

Documented the abuse of the tempfile mechanism for now (I think I
actually recall signing off on this awhile back because there was
no better way to do it, so its all half my fault, it's still horrible)

The actual shell_out calls now raise.  This might cause problems?  I
don't know, but it seems real bad to just blindly swallow errors and
continue.  Feels like more of a bugfix than a breaking change, but it
isn't documented clearly as to why those errors were being ignored
anywhere and didn't break any tests so it appears to be a coding
error.

The mutation of the new_resource was removed (other than the tempfile
hack) and migrated to state on the provider.

Removed some weird return values on things that are called from a
void context.

Added some lazy initializers for stuff that made it a lot more readable

Converted the cookbook_file resource to a real cookbook_file now that
we're in unified_mode.


lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,241 @@ class OsxProfile < Chef::Resource
...
+              item["ProfileIdentifier"] == new_profile_identifier
+            end
+          end
+          nil
+        end
+
+        def check_resource_semantics!
+          if action == :remove
+            if new_profile_identifier
+              unless !new_profile_identifier.nil? && !new_profile_identifier.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_identifier)
```
phiggins, line 81: This check is redundant.
```suggestion
              unless !new_profile_identifier.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_identifier)
```

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,241 @@ class OsxProfile < Chef::Resource
...
+                raise "when removing using the identifier property, it must match the profile identifier"
+              end
+            else
+              new_profile_name = new_resource.profile_name
+              unless !new_profile_name.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_name)
```
phiggins, line 86: This regex is used twice here, might be worth moving to a constant with some descriptive name.

phiggins: Actually these two conditionals are testing for the same thing on different values, so extracting that to a function with unit tests, etc might be a good idea.

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,241 @@ class OsxProfile < Chef::Resource
...
+          if update
+            node.run_state[:config_profiles] = query_installed_profiles
+          else
+            node.run_state[:config_profiles] ||= query_installed_profiles
+          end
+        end
+
+        def query_installed_profiles
+          # Dump all profile metadata to a tempfile
+          tempfile = generate_tempfile
```
phiggins, line 252: Could this use the block form of Tempfile to avoid doing manual cleanup?

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,241 @@ class OsxProfile < Chef::Resource
...
+              new_profile_name = new_resource.profile_name
+              unless !new_profile_name.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_name)
+                raise "When removing by resource name, it must match the profile identifier"
+              end
+            end
+          end
+
+          if action == :install
+            if new_profile_hash.is_a?(Hash)
+              unless new_profile_hash.include?("PayloadIdentifier")
```
phiggins, line 94:
```suggestion
            if new_profile_hash.is_a?(Hash) && !new_profile_hash.include?("PayloadIdentifier")
```

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,241 @@ class OsxProfile < Chef::Resource
...
+                raise "The specified profile does not seem to be valid"
+              end
+            end
+            if new_profile_hash.is_a?(String)
+              unless new_profile_hash.end_with?(".mobileconfig")
```
phiggins, line 99:
```suggestion
            if new_profile_hash.is_a?(String) && !new_profile_hash.end_with?(".mobileconfig")
```

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,245 @@ class OsxProfile < Chef::Resource
...
+          if all_profiles && all_profiles.key?("_computerlevel")
+            return all_profiles["_computerlevel"].find do |item|
+              item["ProfileIdentifier"] == new_profile_identifier
+            end
+          end
+          nil
+        end
+
+        def check_resource_semantics!
+          if mac? && node["platform_version"] > 10.15
```
phiggins, line 79: Does this need to be a string?
```suggestion
          if mac? && node["platform_version"] > '10.15'
```

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,245 @@ class OsxProfile < Chef::Resource
...
+          write_installed_profiles(tempfile)
+          installed_profiles = read_plist(tempfile)
+          logger.trace("Saved profiles to run_state")
+          # Clean up the temp file as we do not need it anymore
+          ::File.unlink(tempfile)
+          installed_profiles
+        end
+
+        def generate_tempfile
+          tempfile = ::Dir::Tmpname.create("allprofiles.plist") {}
```
phiggins, line 266:
```suggestion
          ::Dir::Tmpname.create("allprofiles.plist") {}
```

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,245 @@ class OsxProfile < Chef::Resource
...
+        def query_installed_profiles
+          # Dump all profile metadata to a tempfile
+          tempfile = generate_tempfile
+          write_installed_profiles(tempfile)
+          installed_profiles = read_plist(tempfile)
+          logger.trace("Saved profiles to run_state")
+          # Clean up the temp file as we do not need it anymore
+          ::File.unlink(tempfile)
+          installed_profiles
+        end
```
phiggins, line 263: Unless they're needed for stubbing or something, I'd inline the three small functions used here.

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,245 @@ class OsxProfile < Chef::Resource
...
+      action_class do
+        private
+
+        def profile
+          @profile ||= new_resource.profile || new_resource.profile_name
+        end
+
+        def new_profile_hash
+          @new_profile_hash ||= get_profile_hash(profile)
+        end
```
phiggins, line 140: Could this be set to `{}` if `get_profile_hash` returns nil, or make it so that function always returns a hash? This gets memoized which means it could be run multiple times if its nil, and there's a lot of checking to make sure this function does not return nil.

lib/chef/resource/osx_profile.rb
lamont-granquist: "> 10.15" i think?

lamont-granquist: fixed

lib/chef/resource/osx_profile.rb
lamont-granquist: Changing it to `{}` would mean that the nil checks need to be `empty?` checks and without any domain knowledge of what is going on here, i don't know if `{}` is a separate valid case or not.  I thought about it, but being backwards compatible with the edge case seemed to be better for this refactoring than worrying about speed.

lib/chef/resource/osx_profile.rb
lamont-granquist: what would you call that conditional, because i can't actually tell what is going on there, the error text reads like we're comparing two values when we're just checking a single value?

lib/chef/resource/osx_profile.rb
```diff
@@ -42,8 +43,226 @@ class OsxProfile < Chef::Resource
...
+            return all_profiles["_computerlevel"].find do |item|
+              item["ProfileIdentifier"] == new_profile_identifier
+            end
+          end
+          nil
+        end
+
+        def check_resource_semantics!
+          if mac? && node["platform_version"] =~ "> 10.15"
+            raise "The osx_profile resource is not available on macOS Bug Sur or above due to the removal of apple support for CLI installation of profiles"
```
phiggins, line 80:
```suggestion
            raise "The osx_profile resource is not available on macOS Big Sur or above due to the removal of apple support for CLI installation of profiles"
```

tas50: due to the removal of support for installing profiles from the local machine.

lib/chef/resource/osx_profile.rb
phiggins: This:

`unless !new_profile_identifier.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_identifier)`

can be converted to this:

`if !( !new_profile_identifier.end_with?(".mobileconfig") && /^\w+(?:(\.| )\w+)+$/.match(new_profile_identifier) )`

Applying DeMorgan's law gives you:

`if new_profile_identifier.end_with?(".mobileconfig") || !/^\w+(?:(\.| )\w+)+$/.match(new_profile_identifier)`

So it raises an exception when:
* The string ends with `.mobileconfig`
* The string is not two or more groups of ascii characters separated by a space or `.`

It sounds like `.mobileconfig` files are specific to ios devices so excluding those makes sense. I don't understand why `"MyProfile"` or `"My      Profile"` are excluded.

In any case, I think the error message is trying to say something like `"When removing by resource name, the argument must be a valid profile identifier"` but I'm not sure why you'd want to proactively check for that rather than just allowing the underlying thing to fail.

PR #10242, opened 2020-07-31
Cleanup more files from our package's gem installs

tas50:
Nuke more useless files in gems
Remove all the empty gem dirs for the ruby built in gems so we can skip shipping 41 empty folders


tas50:
Waiting on this adhoc: https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/278

PR #10244, opened 2020-08-01
Further slim down libxml2

tas50:


tas50:
In its current state this breaks the nokogiri install.

tas50:
this works now after some fixes

PR #10246, opened 2020-08-03
Use .key? instead of keys.include

tas50:
We don't need to create an array that we then search when we can just search the hash.


lamont-granquist:
tests seem VERY pissed off

tas50:
Yeah I assumed this one would blow up the world

tas50:
I disabled the false positives in rubocop and reverted those

PR #10250, opened 2020-08-03
Use Hash#transform_values

007lva:
Use Hash#transform_values where makes sense

## Description
Now that the minimum Ruby ruby version is > 2.4, it's safe to use Hash#transform_values.

## Related Issue
https://github.com/chef/chef/commit/dbb339175b445bcd4dfd6c54bababf9dd7908993
https://github.com/chef/chef/pull/9247

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/template/content.rb
+++ b/lib/chef/provider/template/content.rb
@@ -40,14 +40,7 @@ class Chef
           visitor = lambda do |obj|
             case obj
             when Hash
-              # If this is an Attribute object, we need to change class otherwise
-              # we get the immutable behavior. This could probably be fixed by
-              # using Hash#transform_values once we only support Ruby 2.4.
-              obj_class = obj.is_a?(Chef::Node::ImmutableMash) ? Mash : obj.class
-              # Avoid mutating hashes in the resource in case we're changing anything.
-              obj.each_with_object(obj_class.new) do |(key, value), memo|
-                memo[key] = visitor.call(value)
-              end
+              obj.transform_values { |value| visitor.call(value) }
             when Array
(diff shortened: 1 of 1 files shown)


lib/chef/provider/template/content.rb
```diff
@@ -40,14 +40,7 @@ def file_for_provider
...
             when Hash
-              # If this is an Attribute object, we need to change class otherwise
-              # we get the immutable behavior. This could probably be fixed by
-              # using Hash#transform_values once we only support Ruby 2.4.
-              obj_class = obj.is_a?(Chef::Node::ImmutableMash) ? Mash : obj.class
-              # Avoid mutating hashes in the resource in case we're changing anything.
-              obj.each_with_object(obj_class.new) do |(key, value), memo|
-                memo[key] = visitor.call(value)
-              end
+              obj.transform_values { |value| visitor.call(value) }
```
lamont-granquist, line 43: The problem that I can see with this is that when passed an ImmutableMash this will change the return type to a Hash rather than a Mash (probably due to something like `dup` on the ImmutableMash object calling `to_hash` and returning a Hash rather than a Mash, which is a somewhat unfortunate API that we have).

lamont-granquist: Yeah and that's a pretty hard break, i think, this likely breaks any node objects passed into templates where the consumers use symbols as keys rather than strings.  Likely not just a theoretical break.

007lva: Oh, I see, thanks! I'll take a look and try again later.

PR #10253, opened 2020-08-03
apt_repository: Small code refactor in key_is_valid? method

phiggins:

Followup to  #10248.


tas50:
Much better

Issue #10240, opened 2020-07-31
Resources that abuse node state for temporary data need to be addressed (osx_profile, windows_feature_powershell, windows_feature_dism)


lamont-granquist:
This makes me very hangry.

If this is just temporary cache state it should be moved to a Singleton cache object.  If it is intended to be a public-facing API where people expect to run the resources, then inspect the node state for the data and do things based on it in recipe mode that is awful and we need a better solution.

Even if it just means moving it to a singleton object and exposing an API off the provider and allowing users to access it that is better than just jamming it into node data as a side effect.

Possibly a breaking change.

lamont-granquist:
Due to the fact that people inevitably copy the WORST design patterns from the chef/chef codebase everywhere, we need to clean this up, because this should not be the way we support this kind of use case, and we need POSITIVE use cases to show people how to do things the correct way.

lamont-granquist:
One of the things I hate about these approaches is that you have to call the resource first to get the side effect of populating the global state.

A better pattern would be to have an ohai plugin which populated the data, and the resource could fire the ohai resource to redo the ohai plugin to update it when the resource updated.

At a very minimum the right design pattern should be some kind of globalish state that had a lazy initializer so that the first read would populate the data.  I wouldn't be opposed at all to some kind of proper design pattern so that resources could define their own global state (separate from the node attributes and run_state) and then expose that to consumers in a way that it all worked somewhat elegantly -- something that looked like it was a singleton pattern but was actually hanging off of the run_context would be nice.

```
provides :my_resource_name

expose_global_state :my_hashy_thing do
  hash = build_up_some_stuff
  return hash
end

action :whatever do
  hash = get_global_state(:my_resource_name, :my_hashy_thing) # acts like ||=
  hash = update_global_state(:my_resource_name, :my_hashy_thing) # updates even if already set
end
```

kind of like data_bag (one per resource) and data_bag_item for resources to store stuff, with a block that was a lazy initializer.

tas50:
Closing this out while we figure out what actual state needs to be. We have warnings in code in a lot of these cases so the chances that someone copies these are minimal.

PR #10260, opened 2020-08-04
Convert windows custom resources to unified_mode

lamont-granquist:
Tiny bit of code rage included to wipe out a completely unnecessary
nodoc property I found in two of these.

I also renamed the "do_it" method in two of these.

Plus added the sensitive flag to a password property.

And a comment yelling about #10240 

There are 7 other non-custom-resource-style windows resources that
need converting that are all that is left.


tas50:
Solid improvement

PR #10262, opened 2020-08-05
Simplify how we detect fips in chef-config

tas50:
Use the same code we use in ohai.

### DIFF
--- a/chef-config/lib/chef-config/fips.rb
+++ b/chef-config/lib/chef-config/fips.rb
@@ -46,8 +46,8 @@ module ChefConfig
         false
       end
     else
-      fips_path = "/proc/sys/crypto/fips_enabled"
-      File.exist?(fips_path) && File.read(fips_path).chomp != "0"
+      require "openssl" unless defined?(OpenSSL)
+      OpenSSL::OPENSSL_FIPS
     end
   end
 end


chef-config/lib/chef-config/fips.rb
```diff
@@ -21,33 +21,11 @@
...
-        false
-      end
+    require "openssl" unless defined?(OpenSSL)
+    if defined?(OpenSSL.fips_mode) && OpenSSL.fips_mode
+      true
     else
-      fips_path = "/proc/sys/crypto/fips_enabled"
-      File.exist?(fips_path) && File.read(fips_path).chomp != "0"
+      false
     end
```
lamont-granquist, line 29: kill the if-else and just return the boolean expression

lamont-granquist: (there should be a rubocop rule for this if there isn't one already

lamont-granquist: someone needs to write a rule to finding unmatched parens in comments as well)

lamont-granquist: (although someone will ban multiple nesting levels of parents (and i'd hate that))

lamont-granquist: *parens

tas50: Fixed. I wonder how that ended up that way in Ohai. There must have been more there to start.

tas50:
In order to go down this route we're going to need to do some fips testing on real boxes. Ideally this is 1:1 compatible, but faster than the old method, but who knows in FIPS-land.

chef-config/lib/chef-config/fips.rb
phiggins: > there should be a rubocop rule for this if there isn't one already

My continual rubocop experience is being surprised it doesn't handle all of my annoyances.

PR #10265, opened 2020-08-05
Cleanup extra binaries from libxml2 and libxslt

tas50:
Neither of these libraries let you skip building these. Directory cleanup has been removed since that happens in the omnibus-software ruby-cleanup def.

image

### DIFF
--- a/omnibus/config/software/more-ruby-cleanup.rb
+++ b/omnibus/config/software/more-ruby-cleanup.rb
@@ -106,13 +106,22 @@ build do
     end
   end
 
-  block "Remove empty gem dirs from Ruby's built-in gems" do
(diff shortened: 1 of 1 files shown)


omnibus/config/software/more-ruby-cleanup.rb
```diff
@@ -106,13 +106,22 @@
...
+      xmlcatalog
+      xmllint
+      xslt-config
+      xsltproc
+    }.each do |f|
+      file_path = "#{install_dir}/embedded/bin/#{f}"
+
+      if ::File.exist?(file_path)
+        puts "Deleting binary at #{file_path}"
+        FileUtils.rm_rf(file_path)
```
phiggins, line 121:
```suggestion
        FileUtils.rm_f(file_path)
```

nitpick, but if this is not expecting a directory here it shouldn't recursively delete.

PR #10269, opened 2020-08-06
Resolve Style/RedundantAssignment warnings

tas50:
Avoid assigning when we don't need to


lamont-granquist:
i'm really kind of surprised there's not a lot more

tas50:
I was blown away by the small number that came back.

PR #10274, opened 2020-08-07
powershell_script: Use native properties and don't coerce user provided flags with powershell defaults

phiggins:
Some code I had sitting in a local branch. If I remember right this was part of a bigger change that I abandoned and I forgot to PR the remainder.

### DIFF
--- a/lib/chef/resource/execute.rb
+++ b/lib/chef/resource/execute.rb
@@ -509,7 +509,6 @@ class Chef
       def initialize(name, run_context = nil)
         super
         @command = name
-        @backup = 5
         @default_guard_interpreter = :execute
         @is_guard_interpreter = false
       end
--- a/spec/unit/resource/powershell_script_spec.rb
+++ b/spec/unit/resource/powershell_script_spec.rb
@@ -130,24 +130,7 @@ describe Chef::Resource::PowershellScript do
     let(:resource_instance_name ) { @resource.command }
     let(:resource_name) { :powershell_script }
     let(:interpreter_file_name) { "powershell.exe" }
-    before do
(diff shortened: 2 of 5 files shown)


lib/chef/resource/execute.rb
```diff
@@ -509,7 +509,6 @@ class Execute < Chef::Resource
       def initialize(name, run_context = nil)
         super
         @command = name
-        @backup = 5
```
tas50, line 512: where did this default go?

tas50: I bet this is leftover from when backup wasn't a property

spec/unit/resource/powershell_script_spec.rb
```diff
@@ -130,24 +130,7 @@
...
-    before do
-      allow(@resource).to receive(:default_flags).and_return(nil)
-    end
-    it_behaves_like "a Windows script resource"
-  end
-
-  describe "#flags" do
-    let(:resource) { @resource }
 
-    it "appends user's flags to the defaults" do
```
tas50, line 142: This was a particular bug that we wanted to make sure we didn't introduce a regression for. If someone has an option that was basically foo false and default set it to true we want to make sure we're always applying the user supplied values last. We need some sort of test to make sure refactoring later doesn't break the order.

phiggins: Good call. I can probably add a test for the command output.

lib/chef/resource/execute.rb
phiggins: It's probably this: https://github.com/chef/chef/blob/master/lib/chef/resource/file.rb#L65-L66

PR #10281, opened 2020-08-10
Bump Ohai / Cheffish to the latest

tas50:


tas50:
Looks like new Ohai broke ip address detection. I'll look into this before merging.

PR #10141, opened 2020-07-10
WIP : Fix windows winrm bootstrap : 

dheerajd-msys:
Signed-off-by: dheerajd-msys <dheeraj.dubey@msystechnologies.com>

## Description

Code update to fix windows-->windows bootstrap
Checked that windows-->windows bootstrap was failing on latest `chef-15` i.e. `15.12.22` too
## Related Issue

Fixes https://github.com/chef/chef/issues/9922
https://github.com/chef/chef-workstation/issues/1226#issuecomment-635182104
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #10298, opened 2020-08-14
Fix bootstrapping windows-from-linux

lamont-granquist:
replaces #10141

### DIFF
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -61,9 +61,9 @@ class Chef
           client_rb = <<~CONFIG
             chef_server_url  "#{chef_config[:chef_server_url]}"
             validation_client_name "#{chef_config[:validation_client_name]}"
-            file_cache_path   "#{ChefConfig::Config.var_chef_dir(windows: true)}/cache"
-            file_backup_path  "#{ChefConfig::Config.var_chef_dir(windows: true)}/backup"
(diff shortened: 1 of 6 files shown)


lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -72,12 +58,21 @@ def trusted_certs_script
...
         def config_content
+          # The windows: true / windows: false in the block that follows is more than a bit weird.  The way to read this is that we need
+          # the e.g. var_chef_dir to be rendered for the windows value ("C:\chef"), but then we are rendering into a file to be read by
+          # ruby, so we don't actually care about forward-vs-backslashes and by rendering into unix we avoid having to deal with the
+          # double-backwhacking of everything.  So we expect to see:
+          #
+          # file_cache_path "C:/chef"
+          #
+          # Which is mildly odd, but should be entirely correct as far as ruby cares.
+          #
```
lamont-granquist, line 69: So I wrote this comment out with the explanation and I'm now I'm second guessing myself.

The issue is that ultimately the `Chef::Config[:file_cache_path]` gets used everywhere.  We could expect users to wind up trying to put this into shell_out/powershell_out commands and expect it to be rendered correctly into cmd/powershell.  If we have unix-style forward-slashes then we force users to do their own manual cleanpath'ing of this value on windows.

I'm starting to think we should eat the pain of double-backslash escaping the backslashes here...

lamont-granquist:
ffi-yajl release is trolling centos-6, nothing to do with this so merging.

PR #10300, opened 2020-08-14
Change output to clearly show checksums are displayed incompletely

tecracer-theinen:
## Description

When working with templates, people can get confused by a checksum mismatch displaying an apparently shorter hash and subsequently trying to add this one to the recipe:

```
Chef::Exceptions::ChecksumMismatch
Checksum on resource (1de48f) does not match checksum on content (8222e7)
``` 

I propose to add ellipsis ("...") and the end to show this is a truncated hash instead.

## Related Issue

No filed issue, small usability improvement only

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/exceptions.rb
+++ b/lib/chef/exceptions.rb
@@ -423,7 +423,7 @@ class Chef
 
     class ChecksumMismatch < RuntimeError
       def initialize(res_cksum, cont_cksum)
-        super "Checksum on resource (#{res_cksum}) does not match checksum on content (#{cont_cksum})"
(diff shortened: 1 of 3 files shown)


lamont-granquist:
You might want to jam the ellipses into the checksum mismatch exception.  Probably that is only used there and its the rendering method for those checksum mismatch errors.  If not you could probably conditionally add ellipses if you got a checksum that was less than a few characters (we won't ever support like CRC-32 here so there should be no confusion).

tecracer-theinen:
Yeah, maybe that felt too easy yesterday ;-)

PR #10302, opened 2020-08-14
New exit code to signal chef-client exits due to configuration errors

### LINKED ISSUE #10285: Cannot detect chef-client failures due to invalid configuration
### Describe the Enhancement:
Provide a unique exit code that can signal to outside tools if a chef-client run has failed due to an invalid client config vs any other exception.

### Describe the Need:
Failures in `client.rb` usually happen before any handlers or events are loaded and available and cannot be fixed with chef (for obvious reasons). If we emit a unique exit code for configuration errors, users could orchestrate their own tooling to monitor or remediate a problematic config.

### Current Alternative
I could parse the logs and do string matching to isolate these types of errors.

I think I can write this, it looks like a basic implementation would be to add a rescue in [configure_chef](https://github.com/chef/chef/blob/master/lib/chef/application.rb#L97-L103), and raise a `Chef::Application.fatal!(e)` with an error and then handle it in the [exit codes](https://github.com/chef/chef/blob/master/lib/chef/application/exit_code.rb).

NaomiReeves:
## Description
This adds a new exit code, `43`, to more easily distinguish between `chef-client` exits due to a configuration error vs a generic error. I chose `43` since it was the next available exit-code and didn't interfere with existing system exit codes.

This will allow users to easily notify monitoring or other tooling of the configuration error.

## Related Issue
Fixes #10285 

## Types of changes

## Checklist:


spec/unit/application_spec.rb
```diff
@@ -491,16 +491,14 @@
...
 
     def raises_informative_fatals_on_configure_chef
       config_file_regexp = Regexp.new @app.config[:config_file]
-      expect(Chef::Log).to receive(:fatal)
-        .with(/Configuration error/)
-      expect(Chef::Log).to receive(:fatal)
-        .with(config_file_regexp)
-        .at_least(1).times
+      expect(Chef::Log).to receive(:fatal).with(/Configuration error/)
+      expect(Chef::Log).to receive(:fatal).with(config_file_regexp).at_least(:once)
```
phiggins, line 500:
```suggestion
      expect(Chef::Log).to receive(:fatal).with(config_file_regexp)
```

I know you modified the existing code, but this seems unnecessary.

PR #10308, opened 2020-08-17
Remove the unused win32-dir dependency

tas50:
I'm not seeing this being required anywhere in our kit.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 
 GIT
   remote: https://github.com/chef/ohai.git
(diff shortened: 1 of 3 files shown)


tas50:
Adhoc: https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/312#20d9ae03-361b-4f15-aa70-d6db3c729305

tas50:
I searched all the repos I have checked out for a "win32/dir" and came back with nothing so there's not deps in our codebase that just relied on this, but yeah it's possible someone else was using it without a proper dep.

PR #10321, opened 2020-08-18
moving some requires inline to improve cli startup

mwrock:
On my system, this improves `chef-client --version` time by about 2-3 seconds (~30%).

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -35,7 +35,6 @@ require "rubygems/spec_fetcher"
 require "rubygems/platform"
 require "rubygems/package"
 require "rubygems/dependency_installer"
-require "rubygems/uninstaller"
 require "rubygems/specification"
 
 class Chef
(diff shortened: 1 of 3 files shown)


lamont-granquist:
I'd kind of like to see this redone using autoload and if the timing improvement is more or less identical we aggressively ignore matz being worried about it and use autoload everywhere.

lamont-granquist:
(and if matz ever rips out autoload i'll write and publish a gem to add it back if someone else doesn't beat me to it)

tas50:
I'm ok with moving our requires into the method where we actually use them. This seems pretty straight forward.

lamont-granquist:
That "doesn't scale" though.

If you broadly apply the pattern if you have 10 methods that hit the symbol and may have random access you need to add the require to 10 different methods.  Using autoload is a better generic pattern to this problem.

lamont-granquist:
going from require to autoload is also something that could be hacked up into relativity.rb and be applied automatically across the whole codebase.

moving requires into the methods that use them is something that is a lot more complicated and difficult to automate (even with chefstyle that'd be much more difficult).

mwrock:
Yeah I agree about the "scalability." I think one naturally falls into the pattern we have here (and just about any other OSS ruby project) and then it is a constant uphill battle to hunt down and shift around expensive requires. If we can establish a pattern of autoload that makes it easy to apply a lazyloading of gems across the board that could potentially drastically improve startup cost beyond what this PR achieves.

tas50:
Closing since we agreed to go a different route

PR #10323, opened 2020-08-18
Refactor the timezone resource to properly load the current timezone

tas50:
This way we get a proper state change instead of just showing a converge. This also removes awk/grep in not_if statements and overall improves logging. It also wraps the package on linux systems to occur only if the timezone needs to be updated. That was we can skip that whole package state check and logging on every subsequent run.

New systemd linux output:
```
  * timezone[America/Los_Angeles] action set
    * dnf_package[tzdata] action install (up to date)
    - update Etc/UTC
    -   set timezone to "America/Los_Angeles" (was "Etc/UTC")
```

New Windows output:
```
  * timezone[Pacific Standard time] action set
    - setting timezone to 'Pacific Standard time'
```

New Mac output:
```
  * timezone[America/Los_Angeles] action set
    - update GMT
    -   set timezone to "America/Los_Angeles" (was "GMT")
```

Nice clean no-op output:
```
  * timezone[America/Los_Angeles] action set (up to date)
```


spec/unit/resource/timezone_spec.rb
```diff
@@ -36,4 +61,42 @@
     expect { resource.action :set }.not_to raise_error
     expect { resource.action :unset }.to raise_error(Chef::Exceptions::ValidationFailed)
   end
+
+  describe "#current_darwin_tz?" do
```
phiggins, line 65:
```suggestion
  describe "#current_darwin_tz" do
```

spec/unit/resource/timezone_spec.rb
```diff
@@ -36,4 +61,42 @@
...
+
+    context "without admin privs" do
+      it "returns the TZ" do
+        expect(resource).to receive(:shell_out!).and_return(shellout_systemsetup_fail)
+        expect { resource.current_darwin_tz }.to raise_error(RuntimeError)
+      end
+    end
+  end
+
+  describe "#current_systemd_tz?" do
```
phiggins, line 81:
```suggestion
  describe "#current_systemd_tz" do
```

spec/unit/resource/timezone_spec.rb
```diff
@@ -36,4 +61,42 @@
...
+    it "returns the TZ" do
+      expect(resource).to receive(:shell_out!).and_return(shellout_timedatectl)
+      expect(resource.current_systemd_tz).to eql("Etc/UTC")
+    end
+  end
+
+  describe "#current_windows_tz?" do
```
phiggins, line 88:
```suggestion
  describe "#current_windows_tz" do
```

spec/unit/resource/timezone_spec.rb
```diff
@@ -36,4 +61,42 @@
...
+    it "returns the TZ" do
+      expect(resource).to receive(:shell_out!).and_return(shellout_tzutil)
+      expect(resource.current_windows_tz).to eql("UTC")
+    end
+  end
+
+  describe "#current_rhel_tz?" do
```
phiggins, line 95:
```suggestion
  describe "#current_rhel_tz" do
```

lib/chef/resource/timezone.rb
```diff
@@ -48,90 +56,118 @@ class Timezone < Chef::Resource
...
-        # some linux systems may be missing the timezone data
-        if linux?
-          package "tzdata" do
-            package_name suse? ? "timezone" : "tzdata"
-          end
+      # detect the current TZ on darwin hosts
+      #
+      # @since 14.7
+      # @return [String] TZ database value
+      def current_darwin_tz
```
phiggins, line 63: Nitpick: should this be `curent_macos_tz`?

lib/chef/resource/timezone.rb
```diff
@@ -48,90 +56,118 @@ class Timezone < Chef::Resource
...
-        # Modern SUSE, Amazon, Fedora, RHEL, Ubuntu & Debian
-        if systemd?
-          cmd_set_tz = "/usr/bin/timedatectl --no-ask-password set-timezone #{new_resource.timezone}"
+      # detect the current timezone on windows hosts
+      #
+      # @since 14.7
+      # @return [String] timezone id
+      def current_windows_tz
+        tz_shellout = shell_out!("tzutil /g")
+        raise "There was an error running the tzutil command" if tz_shellout.exitstatus == 1
```
phiggins, line 78: There's a few `.exitstatus == 1` calls in here. Any reason to be so specific about the exit code instead of `.error?`?

lib/chef/resource/timezone.rb
```diff
@@ -48,90 +56,118 @@ class Timezone < Chef::Resource
...
-          execute cmd_set_tz do
-            action :run
-            not_if cmd_check_if_set
-          end
+      # detect the current timezone on systemd hosts
+      #
+      # @since 16.5
+      # @return [String] timezone id
+      def current_systemd_tz
+        tz_shellout = shell_out!(["/usr/bin/timedatectl", "status"])
```
phiggins, line 88: `timedatectl status` is for human readable output. There's also a `timedatectl show` that might be less obnoxious to parse: https://www.freedesktop.org/software/systemd/man/timedatectl.html

lib/chef/resource/timezone.rb
```diff
@@ -48,90 +56,118 @@ class Timezone < Chef::Resource
...
+              when "rhel", "amazon"
+                # make sure we have the tzdata files
+                package "tzdata"
+
+                file "/etc/sysconfig/clock" do
+                  owner "root"
+                  group "root"
+                  mode "0644"
+                  action :create
+                  content %{ZONE="#{new_resource.timezone}"\nUTC="true"\n}
```
phiggins, line 152:
```suggestion
                  content <<~CONTENT
                    ZONE="#{new_resource.timezone}"
                    UTC="true"
                  CONTENT
```

spec/unit/resource/timezone_spec.rb
```diff
@@ -36,4 +61,42 @@
...
+      it "returns the TZ" do
+        expect(resource).to receive(:shell_out!).and_return(shellout_systemsetup)
+        expect(resource.current_darwin_tz).to eql("UTC")
+      end
+    end
+
+    context "without admin privs" do
+      it "returns the TZ" do
+        expect(resource).to receive(:shell_out!).and_return(shellout_systemsetup_fail)
+        expect { resource.current_darwin_tz }.to raise_error(RuntimeError)
```
phiggins, line 76: Given RuntimeError is pretty generic this could also check for some portion of the error message.

lib/chef/resource/timezone.rb
tas50: We'd have to see how far back that goes. A lot of the systemd distros we support have early versions of systemd that lack the machine parsable output. I can cut an issue to investigate switching over to that though.

lib/chef/resource/timezone.rb
tas50: We have a mix historically. I switched this over to error? for reability

spec/unit/resource/timezone_spec.rb
tas50: Fixed

lib/chef/resource/timezone.rb
tas50: Fair enough. I fixed the name

spec/unit/resource/timezone_spec.rb
tas50: Fixed

spec/unit/resource/timezone_spec.rb
tas50: Fixed

spec/unit/resource/timezone_spec.rb
tas50: Fixed

spec/unit/resource/timezone_spec.rb
tas50: Fixed

PR #10325, opened 2020-08-18
Fix dll copying in Gemfile to remove Dir.pwd

lamont-granquist:
We can invoke bundler from other directories

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,7 @@ eval_gemfile("./Gemfile.local") if File.exist?("./Gemfile.local")
 if RUBY_PLATFORM.match?(/mswin|mingw|windows/)
   instance_eval do
     ruby_exe_dir = RbConfig::CONFIG["bindir"]
-    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", Dir.pwd) + "/*.dll")
+    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", File.dirname(__FILE__)) + "/*.dll")
     FileUtils.cp_r assemblies, ruby_exe_dir, verbose: false unless ENV["_BUNDLER_WINDOWS_DLLS_COPIED"]
     ENV["_BUNDLER_WINDOWS_DLLS_COPIED"] = "1"
   end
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@mwrock i'm kinda wondering if the dll copying here has something to do with the "invoking a 64-bit dll on 32-bit windows" problem...

this whole bit of code here seems very dubious.

mwrock:
Yes. I had to look into the powershell wrapper about a month ago and it is a 64bit only binary. This would definitely fail trying to use `powershell_exec` on a 32 bit node. I'm not sure if that failing test is using powershell_exec but if it is, this would totally explain the failure.

lamont-granquist:
I don't think its trying to use powershell_exec, but I wonder if its tickling it somehow anyway?

Gemfile
```diff
@@ -87,7 +87,7 @@ eval_gemfile("./Gemfile.local") if File.exist?("./Gemfile.local")
 if RUBY_PLATFORM.match?(/mswin|mingw|windows/)
   instance_eval do
     ruby_exe_dir = RbConfig::CONFIG["bindir"]
-    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", Dir.pwd) + "/*.dll")
+    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", File.dirname(__FILE__)) + "/*.dll")
```
phiggins, line 90:
```suggestion
    assemblies = Dir.glob(File.expand_path("distro/ruby_bin_folder", __dir__) + "/*.dll")
```

lamont-granquist: I was gonna start using this a lot but then ryan davis suggested that __dir__ was a terrible thing, now i'm like ???

phiggins: Hmm, I'm curious what his reasoning is. It's no biggie to stick with what's there, just less parents and things.

lamont-granquist: rubocop has a rule to enforce it (although it doesn't seem to do anything, i'm very confused, but i can't actually find anything against it now).  think i'm leaning towards using `__dir__` now and we can always undo that later if its some kind of giant mistake....

lamont-granquist:
i'm hoping the hab plan failures have nothing to do with this, they certainly don't look like it to me.

PR #10331, opened 2020-08-19
Add new chef_client_trusted_certificate resource

tas50:
This is a pretty simple resource for something that users can certainly do on their own right now, but most people don't even realize is an option. Giving it a dedicated resource exposes the functionality better and just makes it that much easier to write a simple cross-platform cookbook that involves configuring the client.

I'm not a huge fan of having to provide the certificate as a string, but I didn't want to add yet another place where we allow string values or URI. They all have slightly different APIs and we need to just do it via a mixin or something. This can certainly move that direction in the future, but this solves a real problem today.


lib/chef/resource/chef_client_trusted_certificate.rb
```diff
@@ -0,0 +1,101 @@
...
+        xPxTuW1CrbV8/q71FdIzSOciccfCFHpsKOo3St/qbLVytH5aohbcabFXRNsKEqve
+        ww9HdFxBIuGa+RuT5q0iBikusbpJHAwnnqP7i/dAcgCskgjZjFeEU4EFy+b+a1SY
+        QCeFxxC7c3DvaRhBB0VVfPlkPz0sw6l865MaTIbRyoUCAwEAAaMyMDAwCQYDVR0T
+        BAIwADAjBgNVHREEHDAaggwqLmJhZHNzbC5jb22CCmJhZHNzbC5jb20wDQYJKoZI
+        hvcNAQELBQADggEBAGlwCdbPxflZfYOaukZGCaxYK6gpincX4Lla4Ui2WdeQxE95
+        w7fChXvP3YkE3UYUE7mupZ0eg4ZILr/A0e7JQDsgIu/SRTUE0domCKgPZ8v99k3A
+        vka4LpLK51jHJJK7EFgo3ca2nldd97GM0MU41xHFk8qaK1tWJkfrrfcGwDJ4GQPI
+        iLlm6i0yHq1Qg1RypAXJy5dTlRXlCLd8ufWhhiwW0W75Va5AEnJuqpQrKwl3KQVe
+        wGj67WWRgLfSr+4QG1mNvCZb2CkjZWmxkGPuoP40/y7Yu5OFqxP5tAjj4YixCYTW
+        EVA0pmzIzgBg+JIe3PdRy27T0asgQW/F4TY61Yk=
```
phiggins, line 54: I'm impressed this made it past the spellchecker unscathed.

spec/unit/resource/chef_client_trusted_certificate_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  describe "#cert_path" do
+    it "appends .pem to new_resource.cert_name value" do
+      resource.cert_name "something"
+      expect(provider.cert_path).to match(/trusted_certs\/something.pem$/)
+    end
+
+    it "does not append .pem if cert_name already ends in .pem" do
+      resource.cert_name "something.pem"
+      expect(provider.cert_path).to match(/trusted_certs\/something.pem$/)
```
phiggins, line 51: chefstyle does not like these regex.

tas50: Fixed

cspell.json
```diff
@@ -1879,6 +1880,7 @@
     "Zanetti",
     "Zapp",
     "zeproc",
+    "PATHEXT",
```
phiggins, line 1883: Where'd this one come from?

tas50: #10190 added that

PR #10335, opened 2020-08-20
Simplify macos detection in specs to include big sur

tas50:

### DIFF
--- a/spec/support/platform_helpers.rb
+++ b/spec/support/platform_helpers.rb
@@ -95,23 +95,10 @@ def windows_user_right?(right)
   Chef::ReservedNames::Win32::Security.get_account_right(ENV["USERNAME"]).include?(right)
 end
 
-def mac_osx_1014?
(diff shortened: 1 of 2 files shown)


spec/support/platform_helpers.rb
```diff
@@ -95,26 +95,17 @@ def windows_user_right?(right)
...
-    result.stdout.each_line do |line|
-      if /^ProductName:\sMac OS X.*$/.match?(line)
-        return true
-      end
-    end
-  end
-
-  false
+def macos?
+  ::File.exist? "/usr/bin/sw_vers" ? true : false
```
phiggins, line 108:
```suggestion
  ::File.exist? "/usr/bin/sw_vers"
```

tas50:
I realized there was actually an existing macos? method so I just deleted this one entirely. The other one wasn't being used, but did detection via ruby and worked fine in an adhoc

Issue #10336, opened 2020-08-21
windows_package isn't idempotent


jaymzh:
## Description

Windows package reinstalls the package every run.

## Chef Version
15.10.12 and 16.4.41

## Platform Version
Windows 10

## Replication Case
Given this resource:
```
windows_package 'vim' do
  source 'https://github.com/vim/vim-win32-installer/releases/download/' +
    'v8.2.1364/gvim_8.2.1364_x64.exe'
end
```

It seems to reinstall gvim each run.

## Client Output

The icons all go away and come back as you can see it re-install each time. Chef also tells you it installed it:

```
[2020-08-20T19:45:31-07:00] INFO: Processing windows_package[vim] action install (vcrs_windows::default line 9)
[2020-08-20T19:45:31-07:00] INFO: Processing remote_file[C:\chef\cache\package\gvim_8.2.1364_x64.exe] action create (dynamically defined)
[2020-08-20T19:45:49-07:00] INFO: windows_package[vim] installed vim at latest
```

jaymzh:
Note I also tried adding `version '8.2.1364'` but that did not fix it.

phiggins:
@jaymzh This was already mentioned in the docs, but it was a bit buried. I opened a PR to improve things here if you'd like to chime in: #10342

tas50:
I think with the new ohai package behavior on windows and the updated docs we're probably good to close this one out. Feel free to correct me if I missed something @jaymzh

PR #10342, opened 2020-08-24
More prominently mention quirks of package_name in windows_package resource docs.

phiggins:

In response to #10336.

The quirks of the `package_name` property were already mentioned, but it was buried in the description of the `source` property and only mentioned it in context of the `:remove` action. 

The text was copied from the docs for the `windows_package` resource on the docs site and modified: https://docs.chef.io/resources/windows_package/

### DIFF
--- a/lib/chef/resource/windows_package.rb
+++ b/lib/chef/resource/windows_package.rb
@@ -30,7 +30,30 @@ class Chef
       provides(:windows_package) { true }
       provides :package, os: "windows"
 
-      description "Use the **windows_package** resource to manage packages on the Microsoft Windows platform. The windows_package resource supports these installer formats:\n\n Microsoft Installer Package (MSI)\n Nullsoft Scriptable Install System (NSIS)\n Inno Setup (inno)\n Wise\n InstallShield\n Custom installers such as installing a non-.msi file that embeds an .msi-based installer\n"
+      description <<~DESC
+        Use the **windows_package** resource to manage packages on the Microsoft Windows platform.
+        The **windows_package** resource supports these installer formats:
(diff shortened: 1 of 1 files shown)


tas50:
Much better

lib/chef/resource/windows_package.rb
```diff
@@ -30,7 +30,30 @@ class WindowsPackage < Chef::Resource::Package
...
+          * Custom installers such as installing a non-.msi file that embeds an .msi-based installer
+
+        To enable idempotence of the `:install` action or to enable the `:remove` action with no source property specified,
+        `package_name` MUST be an exact match of the name used by the package installer. The names of installed packages
+        Windows knows about can be found in **Add/Remove programs**, in the output of `ohai packages`, or in the
+        `DisplayName` property in one of the following in the Windows registry:
+
+        * `HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall`
+        * `HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall`
+        * `HKEY_LOCAL_MACHINE\\Software\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall`
```
phiggins, line 50: TODO: I don't think this part is accurate, it seems to be a mix of the approaches of chef-client and ohai.

https://github.com/chef/chef/blob/master/lib/chef/provider/package/windows/registry_uninstall_entry.rb#L32-L34
https://github.com/chef/ohai/blob/master/lib/ohai/plugins/packages.rb#L132-L138

PR #10348, opened 2020-08-25
Add initial take at chef_client_launchd

tas50:
The good: We have a resource to configure chef-client to run as a service on macos
The bad: Launchd will stop the service if a reconfigure is necessary and it will not start back up. This is a known issue and it exists in the chef-client cookbook as well, but we need to solve it by forked off the chef-client process before we restart the launchd agent.


lib/chef/resource/chef_client_launchd.rb
```diff
@@ -0,0 +1,139 @@
...
+        default: "/opt/#{Chef::Dist::DIR_SUFFIX}/bin/#{Chef::Dist::CLIENT}",
+        description: "The path to the #{Chef::Dist::CLIENT} binary."
+
+      property :daemon_options, Array,
+      default: lazy { [] },
+      description: "An array of options to pass to the #{Chef::Dist::CLIENT} command."
+
+      property :environment, Hash,
+        default: lazy { {} },
+        description: "A Hash containing additional arbitrary environment variables under which the cron job will be run in the form of `({'ENV_VARIABLE' => 'VALUE'})`."
```
phiggins, line 85: Is this using cron?

tas50: Fixed

PR #10350, opened 2020-08-25
Fix habitat test script

phiggins:

With the extra debug information added, I could see that the failure was printing this message:

```
ERROR: chef-client is not the expected version. Expected '16.5.15', got 'hab-pkg-exec 1.6.139/20200824142405'
```

See: https://buildkite.com/chef-oss/chef-chef-master-verify/builds/6001#54bd468a-7edc-42fc-999d-c7ba90c380fa

That second string is the version of hab. Since these testing scripts pull the latest release of habitat, something must have changed in the way hab parses CLI flags that caused this to break.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -590,9 +590,6 @@ steps:
       linux:
         privileged: true
         single-use: true
-        environment:
-          - HAB_NONINTERACTIVE=true
-
 
 - label: ":habicat: Linux plan (kernel2)"
   commands:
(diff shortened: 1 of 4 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -616,5 +611,3 @@ steps:
         privileged: true
         single-use: true
         shell: ["powershell", "-Command"]
-        environment:
-          - HAB_NONINTERACTIVE=true
```
phiggins, line 620: This didn't work, not sure why. Moved inside the bash scripts run by hab instead.

PR #10358, opened 2020-08-26
Add back nice functionality to chef_client_cron

tas50:
This was in the cookbook recipe before and was useful for low resource
systems where the client could impact host operations. Since we support
this in launchd now, it makes sense to do it here as well.

### DIFF
--- a/lib/chef/resource/chef_client_cron.rb
+++ b/lib/chef/resource/chef_client_cron.rb
@@ -138,6 +138,12 @@ class Chef
         default: lazy { {} },
         description: "A Hash containing additional arbitrary environment variables under which the cron job will be run in the form of `({'ENV_VARIABLE' => 'VALUE'})`."
 
+      property :nice, [Integer, String],
+        description: "The process priority to run the #{Chef::Dist::CLIENT} process at. A value of -20 is the highest priority and 19 is the lowest priority.",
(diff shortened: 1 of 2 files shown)


lib/chef/resource/chef_client_cron.rb
```diff
@@ -138,6 +138,12 @@ class ChefClientCron < Chef::Resource
         default: lazy { {} },
         description: "A Hash containing additional arbitrary environment variables under which the cron job will be run in the form of `({'ENV_VARIABLE' => 'VALUE'})`."
 
+      property :nice, [Integer, String],
+        description: "The process priority to run the #{Chef::Dist::CLIENT} process at. A value of -20 is the highest priority and 19 is the lowest priority.",
+        introduced: "16.5",
+        coerce: proc { |x| Integer(x) },
+        callbacks: { "should be a Integer between -20 and 19" => proc { |v| v >= -20 && v <= 19 } }
```
phiggins, line 145:
```suggestion
        callbacks: { "must be an Integer between -20 and 19" => proc { |v| v >= -20 && v <= 19 } }
```

?

tas50: We usually use should in the callbacks since that's the message presented on error.

PR #10361, opened 2020-08-26
chef_client_launchd: reorder properties and fix log permissions

tas50:
Use 750 as we do in the cron resource
Fix property order so we can diff things more easily


phiggins:
Indirectly related to this PR, but some of the cron and launchd tests are failing on Windows with path issues.

tas50:
Let me cut a PR to fix up that spec

PR #10365, opened 2020-08-28
chef_client_config resource

tas50:
Initial stab at a chef_client_config resource that will allow people to easily write out a client.rb. This isn't meant to cover every single use case, but instead to provide a simple resource that works for 99% of users.


tas50:
We can probably start conversations around this

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,244 @@
...
+      # @todo Allow passing this as a string and convert it to the symbol
+      property :ssl_verify_mode, [Symbol, String],
+        equal_to: %i{verify_none verify_peer},
+        coerce: proc { |x| string_to_symbol(x) },
+        description: <<~DESC
+        Set the verify mode for HTTPS requests.
+
+        * Use :verify_none for no validation of SSL certificates.
+        * Use :verify_peer for validation of all SSL certificates, including the #{ChefUtils::Dist::Server::PRODUCT} connections, S3 connections, and any HTTPS remote_file resource URLs used in #{ChefUtils::Dist::Infra::PRODUCT} runs. This is the recommended setting.
+        DESC
```
phiggins, line 93: Does this need a default value?

lib/chef/resource/support/client.erb
```diff
@@ -0,0 +1,64 @@
...
+      minimal_ohai
+      named_run_list
+      no_proxy
+      ohai_disabled_plugins
+      ohai_optional_plugins
+      pid_file
+      policy_group
+      policy_name
+      run_path
+      ssl_verify_mode).each do |prop| -%>
```
phiggins, line 23: You could probably `.map { |prop| instance_variable_get('@' + prop) }` to avoid calling `eval` multiple times.

lib/chef/resource/chef_client_config.rb
tas50: We're just taking the value from chef-config as the default. 99.9% of people shouldn't touch this.

lib/chef/resource/support/client.erb
```diff
@@ -0,0 +1,64 @@
...
+<%=prop %> = <%= eval('@' + prop).inspect %>
+<% end -%>
+<%# log_location is special due to STDOUT/STDERR from String -> IO Object %>
+<% unless @log_location.nil? %>
+  <% if @log_location.is_a? String && %w(STDOUT STDERR).include?(@log_location) -%>
+log_location = <%= @log_location %>
+  <% else -%>
+log_location = <%= @log_location.inspect %>
+  <% end -%>
+<% end -%>
```
phiggins, line 34: Could you move this logic to a function in the resource and avoid the nested conditional here?

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,244 @@
...
+      #
+      # @return [Symbol] The symbol form of the symbol-like string, string, or symbol value
+      #
+      def string_to_symbol(prop_val)
+        if prop_val.is_a?(String)
+          prop_val.start_with?(":") ? prop_val[1..-1].to_sym : prop_val.to_sym
+        else
+          prop_val
+        end
+      end
```
phiggins, line 68:
```suggestion
      def string_to_symbol(prop_val)
        if prop_val.is_a?(String) && prop_val.start_with?(":")
          prop_val[1..-1].to_sym
        else
           prop_val.to_sym
        end
      end
```

If you don't have to care about things that aren't strings or symbols this can be simplified a bit.

tas50: I think at one point I did but now it shouldn't matter.

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,244 @@
...
+
+      property :config_directory, String,
+        description: "The directory to store the client.rb in.",
+        default: ChefConfig::Config.etc_chef_dir
+
+      property :user, String,
+        description: "The user that should own the client.rb file and the configuration directory if it needs to be created."
+
+      property :node_name, String,
+        description: "The name of the node. This determines which configuration should be applied and sets the `client_name`, which is the name used when authenticating to a #{ChefUtils::Dist::Server::PRODUCT}. If this value is not provided Chef Infra Client will use the node's FQDN as the node name. In general, Chef recommends that you leave this setting blank and let the client assign the FQDN of the node as the node_name during each Chef Infra Client run."
```
phiggins, line 78:
```suggestion
        description: "The name of the node. This determines which configuration should be applied and sets the `client_name`, which is the name used when authenticating to a #{ChefUtils::Dist::Server::PRODUCT}. If this value is not provided #{ChefUtils::Dist::Infra::PRODUCT} will use the node's FQDN as the node name. Leaving this setting blank and letting the client assign the FQDN of the node as the node_name during each #{ChefUtils::Dist::Infra::PRODUCT} run is recommended."
```

tas50: That was a copy/paste from the docs site, but this sounds better

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,265 @@
...
+        coerce: proc { |x| string_to_symbol(x) },
+        description: <<~DESC
+        Set the verify mode for HTTPS requests.
+
+        * Use :verify_none for no validation of SSL certificates.
+        * Use :verify_peer for validation of all SSL certificates, including the #{ChefUtils::Dist::Server::PRODUCT} connections, S3 connections, and any HTTPS remote_file resource URLs used in #{ChefUtils::Dist::Infra::PRODUCT} runs. This is the recommended setting.
+        DESC
+
+      property :formatters, Array,
+        description: "",
```
phiggins, line 113: Does this need to get filled in?

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,265 @@
...
+        default: []
+
+      property :event_loggers, Array,
+        description: "",
```
phiggins, line 117: Does this need to get filled in?

tas50: Yep there's some missing descriptions at the moment due to descriptions missing on the docs site overall*

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,265 @@
...
+        description: "The location in which a process identification number (pid) is saved. An executable, when started as a daemon, writes the pid to the specified file. "
+
+      property :file_cache_path, String,
+        description: "The location in which cookbooks (and other transient data) files are stored when they are synchronized. This value can also be used in recipes to download files with the `remote_file` resource."
+
+      property :file_backup_path, String,
+        description: "The location in which backup files are stored. If this value is empty, backup files are stored in the directory of the target file"
+
+      property :run_path, String,
+        description: "The location where #{ChefUtils::Dist::Infra::PRODUCT} should write the PID file."
```
phiggins, line 196: Is this description accurate? How does this differ from `pid_file`? I don't see `run_path` in the client.rb docs: https://docs.chef.io/config_rb_client/

tas50: I was wondering about this one last night actually. I'll see why we historically had both

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,265 @@
...
+
+      property :https_proxy, String,
+        description: "The proxy server to use for HTTPS connections."
+
+      property :ftp_proxy, String,
+      description: "The proxy server to use for FTP connections."
+
+      property :no_proxy, [String, Array],
+        description: "A comma-separated list or an array of URLs that do not need a proxy.",
+        coerce: proc { |x| x.is_a?(Array) ? x.join(",") : x },
```
phiggins, line 143:
```suggestion
        coerce: proc { |x| x.is_a?(Array) ? x.join(",") : String(x) },
```

nitpick: this could end up nil

phiggins:
Whoops 😅

lib/chef/resource/support/client.erb
tas50: I moved the actual creation of the handler string into the resource itself.

lib/chef/resource/chef_client_config.rb
```diff
@@ -50,6 +52,8 @@ class ChefClientConfig < Chef::Resource
 
       **Adding additional config content to the client.rb**:
 
+      This resource aims to provide common configuration options. Some configuration options are missing and some users may want to use arbitrary Ruby code within their configuration. For this we offer an `additional_config` property that can be used to add any configuration or code to the bottom of the `client.rb` file. Also keep in mind that within the configuration directory is a `client.d` directory where you can stick additional `.rb` files containing configuration options. These can be dropped off using `file` or `template` resources witin your cookbooks as necessary.
```
phiggins, line 55:
```suggestion
      This resource aims to provide common configuration options. Some configuration options are missing and some users may want to use arbitrary Ruby code within their configuration. For this we offer an `additional_config` property that can be used to add any configuration or code to the bottom of the `client.rb` file. Also keep in mind that within the configuration directory is a `client.d` directory where you can put additional `.rb` files containing configuration options. These can be created using `file` or `template` resources within your cookbooks as necessary.
```

lib/chef/resource/chef_client_config.rb
tas50: I believe we only coerce if set so the value couldn't be nil.

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,307 @@
...
+          prop_val.to_sym
+        end
+      end
+
+      property :config_directory, String,
+        description: "The directory to store the client.rb in.",
+        default: ChefConfig::Config.etc_chef_dir
+
+      property :user, String,
+        description: "The user that should own the client.rb file and the configuration directory if it needs to be created."
```
lamont-granquist, line 115: should probably have group+mode if you're gonna have user.

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,307 @@
...
+
+      property :node_name, String,
+        description: "The name of the node. This determines which configuration should be applied and sets the `client_name`, which is the name used when authenticating to a #{ChefUtils::Dist::Server::PRODUCT}. If this value is not provided #{ChefUtils::Dist::Infra::PRODUCT} will use the node's FQDN as the node name. Leaving this setting blank and letting the client assign the FQDN of the node as the node_name during each #{ChefUtils::Dist::Infra::PRODUCT} run is recommended."
```
lamont-granquist, line 118: It is actually probably better to set this to the node.name value, since otherwise the construction of the node name from the FQDN can fail due to lack of DNS.  The value in the config file isn't brittle like that, and shouldn't change.

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,307 @@
...
+        description: "The level of logging performed by the #{ChefUtils::Dist::Infra::PRODUCT}.",
+        equal_to: %i{auto trace debug info warn fatal}
+
+      property :log_location, [String, Symbol],
+        description: "The location to save logs to. This can either by a path to a log file on disk `:syslog` to log to Syslog, `:win_evt` to log to the Windows Event Log, or `'STDERR'`/`'STDOUT'` to log to the *nix text streams.",
+        callbacks: {
+          "accepts Symbol values of ':win_evt' for Windows Event Log or ':syslog' for Syslog" => lambda { |p|
+            p.is_a?(Symbol) ? %i{win_evt syslog}.include?(p) : true
+          },
+        }
```
lamont-granquist, line 153: stuff like `log_location Syslog::Logger.new(“chef-client”)` is also valid config.

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,307 @@
...
+        default: []
+
+      # @todo we need to fixup bad plugin naming inputs here
+      property :ohai_optional_plugins, Array,
+        description: "Optional Ohai plugins that should be enabled to provide additional Ohai data for use in cookbooks.",
+        coerce: proc { |x| x.map { |v| string_to_symbol(v).capitalize } },
+        default: []
+
+      property :minimal_ohai, [true, false],
+        description: "Run a minimal set of Ohai plugins providing data necessary for the execution of #{ChefUtils::Dist::Infra::PRODUCT}'s built-in resources. Setting this to true will skip many large and time consuming data sets such as `cloud` or `packages`. Setting this this to true may break cookbooks that assume all Ohai data will be present."
```
lamont-granquist, line 182: this was never intended for "production" use probably shouldn't be in this API

lib/chef/resource/chef_client_config.rb
```diff
@@ -0,0 +1,307 @@
...
+        description: "The location where #{ChefUtils::Dist::Infra::PRODUCT} should write the PID file."
+
+      property :file_staging_uses_destdir, String,
+        description: "How file staging (via temporary files) is done. When `true`, temporary files are created in the directory in which files will reside. When `false`, temporary files are created under `ENV['TMP']`"
+
+      property :additional_config, String,
+        description: "Additional text to add at the bottom of the client.rb config. This can be used to run custom Ruby or to add less common config options"
+
+      action :create do
+        unless ::Dir.exist?(new_resource.config_directory)
```
lamont-granquist, line 228: why the extra idempotency check here and below?

lib/chef/resource/support/client.erb
lamont-granquist: yeah definitely convert the eval stuff to instance_variable_get

lib/chef/resource/chef_client_config.rb
tas50: They're the equivalent of "create_if_missing" which we don't have for directory. People have historically sharpened pitch forks over the details of creating these directories so this way if they want to do it themselves they do it and this resource won't touch them. It creates the dirs if they aren't there, but it's not actively managing them since it's concerned with the client.rb only. The same thing is done in the other chef_client_* resources.

lib/chef/resource/support/client.erb
tas50: TIL ﻿instance_variable_get. I updated it for that.

lib/chef/resource/chef_client_config.rb
tas50: I had a default of node.name originally, but decided to go with the default out of config. I guess it does make sense if DNS is ever flakey to hard code that value.

PR #10370, opened 2020-09-01
Remove a redundant spec loop

tas50:
We were looping over the same object twice here.


tas50:
It looks like we're relying on one of these actually

tas50:
Sticking to just the first spec loop here for now

PR #10371, opened 2020-09-01
Add an ohai timing test to find busted DNS on CI testers

lamont-granquist:
I disabled the first test here entirely because we don't seem to have stable enough timing to shell_out to ohai so that we wouldn't suffer, so it all becomes a bit more aspirational than functional.

The second test should be good enough to find busted DNS issues like the one last week though.

### DIFF
--- a/spec/integration/ohai/ohai_spec.rb
+++ b/spec/integration/ohai/ohai_spec.rb
@@ -1,0 +1,61 @@
+require "spec_helper"
+require "chef/mixin/shell_out"
+require "benchmark" unless defined?(Benchmark)
+
+describe "ohai" do
+  include Chef::Mixin::ShellOut
+
+  let(:ohai) { "bundle exec ohai" }
+
+  describe "testing ohai performance" do
+    # The purpose of this test is to generally find misconfigured DNS on
+    # CI testers.  If this fails, it is probably because the forward+reverse
+    # DNS lookup that node[:hostname] needs is timing out and failing.
+    #
+    # If it is failing spuriously, it may mean DNS is failing spuriously, the
+    # best solution will be to make sure that `hostname -f`-like behavior hits
+    # /etc/hosts and not DNS.
+    #
+    # If it still fails supriously, it is possible that the server has high
+    # CPU load (e.g. due to background processes) which are contending with the
+    # running tests (disable the screensaver on servers, stop playing Fortnite
+    # while you're running tests, etc).
+    #
+    # If this just fails due to I/O being very slow and ruby being very slow to
+    # startup then that still indicates that the tester configuration needs
+    # fixing.  The fact that this will fail on a windows box on a virt that doesn't
+    # use an SSD is because we have a higher bar for the tests to run successfully
+    # and that configuration is broken, so this test is red for a reason.
+    #
+    # This will probably fail on raspberry pi's or something like that as well.  That
+    # is not a bug.  We will never accept a raspberry pi as a CI tester for our
+    # software.  Feel free to manually delete and thereby skip this file in your
+    # own testing harness, but that is not our concern, we are testing behavior
+    # that is critical to our infrastructure and must run in our tests.
+    #
+    # XXX: unfortunately this is so slow on our windows testers (~9 seconds on one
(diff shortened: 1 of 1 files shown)


spec/integration/ohai/ohai_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    # use an SSD is because we have a higher bar for the tests to run successfully
+    # and that configuration is broken, so this test is red for a reason.
+    #
+    # This will probably fail on raspberry pi's or something like that as well.  That
+    # is not a bug.  We will never accept a raspberry pi as a CI tester for our
+    # software.  Feel free to manually delete and thereby skip this file in your
+    # own testing harness, but that is not our concern, we are testing behavior
+    # that is critical to our infrastructure and must run in our tests.
+    #
+    it "the hostname plugin must return in under 4 seconds" do
```
tas50, line 35: Would it make sense to say "see code comment in case of failure". You have a pretty nice comment up there but if this fails in 2 years we're going to miss it

spec/integration/ohai/ohai_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      # we time the command and then compare to the timing result in order to display how long it actually took
+      start = Time.now
+      shell_out!("#{ohai} hostname")
+      delta = Time.now - start
+      expect(delta).to be < 0
+    end
```
phiggins, line 41: To avoid random failures, using something like `Benchmark.realtime` to use the monotonic timer stuff will be better than `Time.now`: https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html#method-c-realtime

spec/integration/ohai/ohai_spec.rb
phiggins: tbh I'd be fine with printing the whole comment if this is expected to fail very rarely if ever.

spec/integration/ohai/ohai_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    # software.  Feel free to manually delete and thereby skip this file in your
+    # own testing harness, but that is not our concern, we are testing behavior
+    # that is critical to our infrastructure and must run in our tests.
+    #
+    it "the hostname plugin must return in under 4 seconds" do
+      # we time the command and then compare to the timing result in order to display how long it actually took
+      start = Time.now
+      shell_out!("#{ohai} hostname")
+      delta = Time.now - start
+      expect(delta).to be < 0
```
jaymzh, line 40: The test says under 4 seconds, but you're checking if it's under 0 seconds...

spec/integration/ohai/ohai_spec.rb
lamont-granquist: windows though?

spec/integration/ohai/ohai_spec.rb
lamont-granquist: this test probably doesn't work, if i do anything it'll probably be the other test.

lamont-granquist: (windows is just way too slow normally)

lamont-granquist: (although maybe autoloading and related approaches would help a lot)

spec/integration/ohai/ohai_spec.rb
phiggins: It works in windows. If I remember right, if the system doesn't support it it falls back on the system clock.

```
irb(main):006:0> t = Process.clock_gettime(Process::CLOCK_MONOTONIC) ; sleep 1.5; Process.clock_gettime(Process::CLOCK_MONOTONIC) - t
=> 1.5119957999995677
```

lamont-granquist: fixed, looks like Benchmark.realtime works fine on windows

tas50:
Similar to what we actually do in ohai itself on the CLI: https://github.com/chef/ohai/blob/345243c94df0071c65821782263c3f9e453d858a/lib/ohai/application.rb#L80

PR #10372, opened 2020-09-01
Shell test cleanup

phiggins:
I recently noticed that one test was printing a lot of output in the unit tests, as can be seen in this recent buildkite output: https://buildkite.com/chef-oss/chef-chef-master-verify/builds/6053#4785c4cb-14f3-4937-b6ca-f42a83a6f3ec/86-886

The test was running ohai when it was supposed to be stubbed, meaning it was printing a lot of output and taking a lot longer than it should. In CI it was a few hundred lines and 1.5 seconds, but in my local environment it was several thousand lines of output and the test took ~15 seconds.

* The first commit is rearranging some test code to remove unspecified dependencies in the tests so that each file could be run separately.
* The second commit is cleaning up the test code to be more consistent to allow me to see the differences in test setup/assertions.
* The last commit is making the meaningful change to fix the test.

### DIFF
--- a/spec/unit/shell/shell_session_spec.rb
+++ b/spec/unit/shell/shell_session_spec.rb
@@ -161,7 +161,7 @@ describe Shell::SoloLegacySession do
 
   before do
     Chef::Config[:shell_config] = { override_runlist: [Chef::RunList::RunListItem.new("shell::override")] }
-    Chef::Config[:shell_solo] = nil
+    Chef::Config[:solo_legacy_shell] = true
     session.node = node
     session.json_configuration = json_attribs
     session.run_context = run_context
--- a/spec/unit/shell/shell_session_spec.rb
+++ b/spec/unit/shell/shell_session_spec.rb
@@ -98,20 +98,10 @@ end
 
 describe Shell::StandAloneSession do
   let(:json_attribs) { { "a" => "b" } }
-  let(:chef_rest) { double("Chef::ServerAPI") }
   let(:node) { Chef::Node.new }
-  let(:events) { Chef::EventDispatch::Dispatcher.new }
   let(:session) { Shell::StandAloneSession.instance }
-  let(:client) do
-    double("Chef::Client.new",
-      run_ohai: true,
-      load_node: true,
-      build_node: true,
-      register: true,
-      sync_cookbooks: {})
-  end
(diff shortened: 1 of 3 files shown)


spec/unit/shell/shell_session_spec.rb
```diff
@@ -180,68 +154,65 @@ def loading_complete
...
+    expect(session.resource_collection.include?(kitteh_cat)).to be true
+    expect(session.resource_collection.include?(keyboard_cat)).to be true
   end
 
   it "runs chef with a resource collection from the compiled cookbooks" do
-    allow(@session).to receive(:node_built?).and_return(true)
+    allow(session).to receive(:node_built?).and_return(true)
     chef_runner = double("Chef::Runner.new", converge: :converged)
-    expect(Chef::Runner).to receive(:new).with(an_instance_of(Chef::RunContext)).and_return(chef_runner)
+    expect(Chef::Runner).to receive(:new).with(session.recipe.run_context).and_return(chef_runner)
```
phiggins, line 203: This was the first clue as to what was wrong. This was asserting that the argument would be any instance and changing it to be the specific instance in this test failed.

spec/unit/shell/shell_session_spec.rb
```diff
@@ -180,68 +154,65 @@ def loading_complete
...
-    @events = Chef::EventDispatch::Dispatcher.new
-    @run_context = @session.run_context = Chef::RunContext.new(@node, {}, @events)
-    @session.node = @node
-    @recipe = @session.recipe = Chef::Recipe.new(nil, nil, @run_context)
-    Shell::Extensions.extend_context_recipe(@recipe)
-  end
-
-  after do
-    Chef::Config[:shell_solo] = nil
+    Chef::Config[:solo_legacy_shell] = true
```
phiggins, line 164: This is the meaningful change. This method determines which subclass of `Shell::Session` is used and it is looking for this key to be set in `Chef::Config`: https://github.com/chef/chef/blob/master/lib/chef/shell.rb#L189-L196

The string `shell_solo` is not present anywhere else in the project so this is likely some leftover cruft.

lamont-granquist: yeah, this is some veeeery old code...

Issue #10367, opened 2020-08-31
ZD 25979 - /Users created on nodes after knife bootstrap


vsingh-msys:
#### Describe the bug

----------------------------------------------

```
r-goto@Mac:1.r-goto_admin +[master]%chef -v
Chef Workstation version: 20.8.125
Chef Infra Client version: 16.4.38
Chef InSpec version: 4.22.8
Chef CLI version: 3.0.24
Chef Habitat version: 1.6.56
Test Kitchen version: 2.6.0
Cookstyle version: 6.15.5
```

```
[vagrant@localhost /]$ ls /Users
ls: cannot access Users: no such directory
```

After `knife bootstrap`:

```
[vagrant@localhost /]$ ls /Users/r-goto/.chef/cache/chef-client-running.pid 
Users/r-goto/.chef/cache/chef-client-running.pid
```

`knife bootstrap` is creating chef directory right under `root` directory, which is not acceptable.
How can we change the directory location and where is it recommended to place? 

**To Reproduce**

Steps to reproduce the behavior.

**Expected behavior**

**Actual behavior**

A clear and concise description of what actually happened.

**Additional context**

Any other additional context here.

**Associated GitHub Issue**

Any associated public GitHub Issue URL link will be included here

vsingh-msys:
it seems that by default `cache_path` picking up workstation home dir instead of node home dir

```
> cat /etc/chef/client.rb 

chef_server_url  "https://chef-server.test/organizations/4thcoffee"
validation_client_name "4thcoffee-validator"
chef_license "accept"
log_location   STDOUT
node_name "node2.test"
trusted_certs_dir "/etc/chef/trusted_certs"
file_cache_path "/home/vivek.singh/.chef/cache"  # workstion home dir /home/vivek.singh
file_backup_path "/home/vivek.singh/.chef/backup"

```

r-goto:
Adding these lines in the `knife.rb` will place the `Users` directory where it's used to be.  

```
file_cache_path "/var/chef/cache"
file_backup_path "/var/chef/backup"
```

Could this be a workaround for this bug? 
Could this change cause any impact after fix release?

--

Also for those already bootstrapped and `Users`/ already placed under /, is it okay to relocate the directory to any intended location manually or should those nodes be bootstrapped again with above lines in `knife.rb`?

vsingh-msys:
Hi @r-goto 

> Could this be a workaround for this bug?

[...]

PR #10374, opened 2020-09-02
[WIP] Introduce Chef.bootstrap_context to wire train transport_connection

vsingh-msys:
#### [POC] This PR is a footprint of mapping bootstrapping train transport connections with chef-utiles.

A couple of bugs have been observed related to chef config picking up workstation info to decide the node configurations.

## Related Issue

Fixes: https://github.com/chef/chef/issues/10367

## Types of changes

## Checklist:

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -379,7 +379,7 @@ module ChefConfig
     #
     # @param path [String]
     def self.path_accessible?(path)
-      File.exist?(path) && File.readable?(path) && File.writable?(path)
+      ChefUtils.file_exist?(path) && File.readable?(path) && File.writable?(path)
     end
 
     # Where cookbook files are stored on the server (by content checksum)
@@ -920,7 +920,7 @@ module ChefConfig
     end
 
     # returns a platform specific path to the user home dir if set, otherwise default to current directory.
-    default( :user_home ) { PathHelper.home || Dir.pwd }
+    default( :user_home ) { ChefUtils.home || PathHelper.home || Dir.pwd }
 
     # Enable file permission fixup for selinux. Fixup will be done
     # only if selinux is enabled in the system.
--- a/chef-utils/lib/chef-utils/dsl/train_helpers.rb
+++ b/chef-utils/lib/chef-utils/dsl/train_helpers.rb
@@ -57,6 +57,14 @@ module ChefUtils
         end
       end
 
+      def home
+        if __transport_connection
+          __transport_connection.run_command("echo $HOME").stdout&.strip
+        else
+          ENV["$HOME"]
+        end
+      end
+
       extend self
     end
   end
--- a/chef-utils/lib/chef-utils/internal.rb
+++ b/chef-utils/lib/chef-utils/internal.rb
@@ -85,6 +85,8 @@ module ChefUtils
       # Software consumers MUST override this method with their own implementation.  The default behavior here is subject to change.
       return Chef.run_context.transport_connection if defined?(Chef) && Chef.respond_to?(:run_context) && Chef&.run_context&.transport_connection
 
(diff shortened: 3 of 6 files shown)


chef-config/lib/chef-config/config.rb
```diff
@@ -379,7 +379,7 @@ def self.derive_path_from_chef_repo_path(child_path)
     #
     # @param path [String]
     def self.path_accessible?(path)
-      File.exist?(path) && File.readable?(path) && File.writable?(path)
+      ChefUtils.file_exist?(path) && File.readable?(path) && File.writable?(path)
```
vsingh-msys, line 382: This is only an example of how we are using train helpers `ChefUtiles.file_exist?` which hitting the node to check whether the dir exists or not.

Ref: https://github.com/chef/chef/blob/fb4197f19f60740780c35f4ff37adbd493fdb2dc/chef-utils/lib/chef-utils/dsl/train_helpers.rb#L37-L43

lib/chef/knife/bootstrap.rb
```diff
@@ -522,7 +522,7 @@ def bootstrap_context
             Knife::Core::WindowsBootstrapContext.new(config, config[:run_list], Chef::Config, secret)
```
vsingh-msys, line 522: Pending to wire this for windows bootstrap context

chef-utils/lib/chef-utils/internal.rb
```diff
@@ -85,6 +85,8 @@ def __transport_connection
       # Software consumers MUST override this method with their own implementation.  The default behavior here is subject to change.
       return Chef.run_context.transport_connection if defined?(Chef) && Chef.respond_to?(:run_context) && Chef&.run_context&.transport_connection
 
+      return Chef.bootstrap_context.transport_connection if defined?(Chef) && Chef.respond_to?(:bootstrap_context) && Chef&.bootstrap_context&.transport_connection
+
```
vsingh-msys, line 89: - As run context denoting the chef-client run and have ohai node object in place to respond query node info like platform etc.

- Here we have added bootstrap_context which denoting we have actively made connection to the node able to answer query about the node using train transport connection.

chef-utils/lib/chef-utils/dsl/train_helpers.rb
```diff
@@ -57,6 +57,14 @@ def file_open(*args, &block)
         end
       end
 
+      def home
+        if __transport_connection
+          __transport_connection.run_command("echo $HOME").stdout&.strip
+        else
+          ENV["$HOME"]
+        end
+      end
```
lamont-granquist, line 66: This shouldn't be added here, we pull in a train-aware `shell_out` into Chef::Mixin::ShellOut so that should be used directly with `shell_out("echo $HOME").stdout&.strip` in the caller.

vsingh-msys: sounds reasonable to me will look into it thanks!

chef-utils/lib/chef-utils/internal.rb
lamont-granquist: This likely isn't the right way to do this for knife.  Chef has to do this because we support calling the DSL methods from library files where we don't have any kind of useful context or DSL so need to use the Chef.run_context to get at the state of the chef-client run.  That itself is probably a bad idea because it isn't threadsafe.  What should be done here is that under lib/chef/knife there should be a mixin created which pulls in the train_helpers and wires up the transport_connection similarly to lib/ohai/mixin/train_helpers.rb in https://github.com/chef/ohai/pull/1503/files

vsingh-msys:
Closing this PR, based on the comment https://github.com/chef/chef/pull/10374#discussion_r489095424, will raise new PR based on ohai target node updates.

PR #10379, opened 2020-09-02
Chef 17: Assume Rubygems 1.8 in the rubygems provider / specs

tas50:
Rubygems 1.8 came out May 4, 2011 and we're still checking for rubygems older than that.

For the record the other checks for Rubygems 2.7 are Nov 2017+

### DIFF
--- a/lib/chef/provider/package/rubygems.rb
+++ b/lib/chef/provider/package/rubygems.rb
@@ -69,7 +69,6 @@ class Chef
           # A rubygems specification object containing the list of gemspecs for all
           # available gems in the gem installation.
           # Implemented by subclasses
-          # For rubygems >= 1.8.0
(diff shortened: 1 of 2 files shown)


lamont-granquist:
We're ruby >= 2.6 now, so we can probably drop rubygems 1.8 support without a major version bump.  Someone would really be going out of their way to try to get rubygems 1.7 onto ruby 2.6 and remove the built in gem.

phiggins (approved):

tas50:
This would impact system ruby though not just chef_gem

lamont-granquist:
hmmm yeah forgot what is on rhel-6, its got some kind of horrible ruby-1.9.1/1.9.2 i think by default

PR #10381, opened 2020-09-03
chef_client_systemd_timer: Add the ability to set CPUQuota on the chef-client unit

tas50:
This allows us to set a quota like 50% on everything handled by the chef-client including all the sub-processes like ohai or inspec. I looked at using systemd's nice functionality since that would give us a unified property for all the *nix processes, but that only applies to the chef-client process itself. This is much more powerful and gives users what user's probably really want which is "chef and all its stuff should not consume all the CPU on my host".

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/linux.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/linux.rb
@@ -100,7 +100,7 @@ end
 
 chef_client_systemd_timer "Run chef-client as a systemd timer" do
   interval "1hr"
-  cpu_quote 50
+  cpu_quota 50
   only_if { systemd? }
 end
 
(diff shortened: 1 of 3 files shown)


kitchen-tests/cookbooks/end_to_end/recipes/linux.rb
```diff
@@ -100,6 +100,7 @@
 
 chef_client_systemd_timer "Run chef-client as a systemd timer" do
   interval "1hr"
+  cpu_quote 50
```
phiggins, line 103:
```suggestion
  cpu_quota 50
```

phiggins: It looks like the CI failures are from this.

phiggins (approved):

ecottom:
Will this be ported over to https://github.com/chef-cookbooks/chef-client/blob/master/recipes/systemd_service.rb ?

tas50:
I was planning to backport this all today

PR #10382, opened 2020-09-03
Fix nil deep_merging

lamont-granquist:
While this behavior was fixed for attributes a long, long time ago it
turns out we don't use this deep_merge mixin for constructing the
deep merge of attribute precedence levels any more.

This bug however surfaces because we still use deep_merge to merge
json attributes into normal attributes.  That means that it is not
possible to pass a nil / null value in through -j and have it override
a normal attribute which is pulled back in from the node object on
the chef-server.

So this fixes that behavior to be consistent with the behavior of
attributes that we've had for several years now.

### DIFF
--- a/lib/chef/mixin/deep_merge.rb
+++ b/lib/chef/mixin/deep_merge.rb
@@ -75,14 +75,7 @@ class Chef
       # @api private
       #
       def deep_merge!(source, dest)
-        # if dest doesn't exist, then simply copy source to it
-        if dest.nil?
-          dest = source; return dest
-        end
-
         case source
-        when nil
-          dest
         when Hash
           if dest.is_a?(Hash)
             source.each do |src_key, src_value|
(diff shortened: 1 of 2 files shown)


phiggins:
Is the CI failure related? I haven't seen it anywhere but this PR:

```
   Context chef-client
    [+] is an executable 529ms
    [-] is the expected version 125ms
      RuntimeException: You cannot call a method on a null-valued expression.
      at <ScriptBlock>, C:\bkb0513b9d\habitat\tests\test.pester.ps1: line 16
```

https://buildkite.com/chef-oss/chef-chef-master-verify/builds/6077#b0513b9d-46d6-4355-970b-1ab400cb3a42/1126-1133

lamont-granquist:
I don't know where that is coming from or how to replicate it and the error message is completely cryptic...

lamont-granquist:
something to do with `chef-client.bat --version` on windows not working, but i can't see how that's related

phiggins:
https://github.com/chef/chef/blob/master/habitat/tests/test.pester.ps1#L16

It looks like the same bug I fixed in #10350 but with windows this time. Probably not related, will see if I can repro on master.

phiggins (approved):

PR #10383, opened 2020-09-03
Improve cli boot performance by prefering autoload over requires

mwrock:
On my windows system, this improves `chef-client --version` performance by ~2.5x. It will also improve other scenarios to a lesser degree. I think we could yield even more but this is pretty much all the low hanging fruit and then some.

(https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).


lib/chef/mixin/openssl_helper.rb
```diff
@@ -20,7 +20,7 @@ module Mixin
     # various helpers for use with openssl. Currently used by the openssl_* resources
     module OpenSSLHelper
       def self.included(_base)
-        require "openssl" unless defined?(::OpenSSL)
+        autoload :OpenSSL, "openssl"
```
lamont-granquist, line 23: you can probably now delete this handler and move it to the top level?

lib/chef/util/diff.rb
```diff
@@ -86,6 +83,9 @@ def diff(old_file, new_file)
       # produces a unified-output-format diff with 3 lines of context
       # ChefFS uses udiff() directly
       def udiff(old_file, new_file)
+        require "diff/lcs"
+        require "diff/lcs/hunk"
```
lamont-granquist, line 87: that could probably autoload both on Diff::LCS and don't worry about calling out "hunk" specifically

chef-config/lib/chef-config/config.rb
```diff
@@ -29,7 +29,7 @@
 require_relative "path_helper"
 require_relative "mixin/fuzzy_hostname_matcher"
 
-Mixlib.autoload :Shellout, "mixlib/shellout"
+Mixlib.autoload :ShellOut, "mixlib/shellout"
```
lamont-granquist, line 32: Oh right, nested class constants...  There's gotta be a better syntax to use here...

lamont-granquist: (it appears there isn't)

mwrock: I've been poking around and it looks like a better convention is:
```
module Mixlib
    autoload :ShellOut, "mixlib/shellout"
end
```
It srill feels clumsy but it is more precise for the cases where the module` may not be loaded.

lamont-granquist: yeah that would be my concern, and yeah that would be my suggestion.

(also means you have to know if its a module or a class at the top level, which is mildly annoying)

lamont-granquist: (although that may only matter for `class Chef` and everything else should be properly a module I hope)

lib/chef/util/diff.rb
mwrock: As far as I can tell you can only autoload one require per symbol.

phiggins:
The buildkite CI job should pass if you rebase this on master.

lib/chef/util/diff.rb
lamont-granquist: well that's annoying.

i wonder if we need to write a better autoload implementation or look and see what activesupport has that we might be able to copy....

lamont-granquist (approved):

tas50:
Unrelated failures in unit. Nice stuff @mwrock

PR #10388, opened 2020-09-04
Add a :reboot_delay property to the windows_ad_join resource

chef-davin:

## Description

In many situations, when using the windows_ad_join resource, you can't utilize the build-in reboot functionality because the windows system reboots too quickly and the chef run can't close out with the proper error code. We often see this failure when in test-kitchen, where the appropriate chef exit codes don't get passed to test-kitchen in order for it to trigger a retry.

I've added a `:reboot_delay` property to the resource to leverage the `:delay_mins` property in the `reboot` resource.  This will allow users to leverage that `:delay_mins` property without having to use a `notify` call from their use of the `windows_ad_join` resource.

Example:

** Join a domain and reboot the system after a 2 minute delay **

```ruby
windows_ad_join 'ad.example.org' do
  domain_user 'nick'
  domain_password 'p@ssw0rd1'
  new_hostname 'win-workstation'
  reboot: :reboot_now
  reboot_delay: 2
end
```

## Related Issue

This is allowing the fix for this [Test-Kitchen Issue](https://github.com/test-kitchen/test-kitchen/issues/1227) to be available to the windows_ad_join resource.

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/windows_ad_join.rb
+++ b/lib/chef/resource/windows_ad_join.rb
@@ -80,6 +80,11 @@ class Chef
         description: "Controls the system reboot behavior post domain joining. Reboot immediately, after the #{Chef::Dist::PRODUCT} run completes, or never. Note that a reboot is necessary for changes to take effect.",
         default: :immediate
 
+      property :reboot_delay, Integer,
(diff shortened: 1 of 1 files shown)


tas50:
Thanks @chef-davin

PR #10390, opened 2020-09-05
chef_client_launchd: create a launchd daemon to handle the client restart

tas50:
Launchd doesn't have the concept of a reload aka restart. Instead to update a daemon config you have to unload it and then reload the new plist. That's usually fine (actually just like upstart), but not if Chef is trying to restart itself. Currently if you change the nice level or interval the client will unload its own config which kills the run and it never gets started back up.  So to work around this we're going to install a launchd daemon that has a watcher on the chef-client daemon plist file. It's only job to to unload and reload the client. That way when the daemonized chef-client process updates the plist the restarter job will handle the restart. To ensure we're not in the middle of the next recipe we use a chef_sleep resource to sleep a bit post launchd update.

I also updated the resource to work better with cinc while I was in here.

### DIFF
--- a/lib/chef/resource/chef_client_launchd.rb
+++ b/lib/chef/resource/chef_client_launchd.rb
@@ -119,7 +119,7 @@ class Chef
           nice new_resource.nice
           low_priority_io true
           notifies :sleep, "chef_sleep[Sleep before client restart]", :immediately
-          action :enable
+          action :create # create only creates the file. No service restart triggering
         end
 
         # Launchd doesn't have the concept of a reload aka restart. Instead to update a daemon config you have
@@ -135,7 +135,7 @@ class Chef
           program_arguments ["/bin/bash",
                              "-c",
                              "echo; echo #{Chef::Dist::PRODUCT} launchd daemon config has been updated. Manually unloading and reloading the daemon; echo Now unloading the daemon; launchctl unload /Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist; sleep 2; echo Now loading the daemon; launchctl load /Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist"]
-          action :enable
+          action :enable # enable creates the plist & triggers service restarts on change
         end
 
         # We want to make sure that after we update the chef-client launchd config that we don't move on to another recipe
@@ -149,8 +149,12 @@ class Chef
       end
 
       action :disable do
-        service "#{Chef::Dist::CLIENT}" do
+        service "#{Chef::Dist::PRODUCT}" do
           service_name "com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}"
+          action :disable
+        end
+
+        service "com.#{Chef::Dist::SHORT}.restarter" do
           action :disable
         end
       end
--- a/lib/chef/resource/chef_client_launchd.rb
+++ b/lib/chef/resource/chef_client_launchd.rb
@@ -122,6 +122,11 @@ class Chef
           action :enable
         end
 
+        # Launchd doesn't have the concept of a reload aka restart. Instead to update a daemon config you have
+        # to unload it and then reload the new plist. That's usually fine, but not if chef-client is trying
+        # to restart itself. If the chef-client process uses launchd or macosx_service resources to restart itself
+        # we'll end up with a stopped service that will never get started back up. Instead we use this daemon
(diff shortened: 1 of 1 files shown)


lib/chef/resource/chef_client_launchd.rb
```diff
@@ -110,21 +110,38 @@ class ChefClientLaunchd < Chef::Resource
           end
         end
 
-        launchd "com.chef.chef-client" do
+        launchd "com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}" do
```
lamont-granquist, line 113: yeah we need to get that chef-utils dist constants PR merged because you just broke it here

tas50: Truth

lib/chef/resource/chef_client_launchd.rb
```diff
@@ -110,21 +110,38 @@ class ChefClientLaunchd < Chef::Resource
...
+          watch_paths ["/Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist"]
+          standard_out_path "#{::File.join(new_resource.log_directory, new_resource.log_file_name)}"
+          standard_error_path "#{::File.join(new_resource.log_directory, new_resource.log_file_name)}"
+          program_arguments ["/bin/bash",
+                             "-c",
+                             "echo; echo #{Chef::Dist::PRODUCT} launchd daemon config has been updated. Manually unloading and reloading the daemon; echo Now unloading the daemon; launchctl unload /Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist; sleep 2; echo Now loading the daemon; launchctl load /Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist"]
+          action :enable
+        end
+
+        chef_sleep "Sleep before client restart" do
```
lamont-granquist, line 136: this could probably use a comment because it isn't clear at all why this sleep is in here.

tas50: I half typed it and then deleted it. Clearly I should have kept going

tas50:
I think we can actually skip the sleep so I'm going to test that further. I'm pretty sure we never get there.

lib/chef/resource/chef_client_launchd.rb
```diff
@@ -110,21 +110,51 @@ class ChefClientLaunchd < Chef::Resource
...
+        # Launchd doesn't have the concept of a reload aka restart. Instead to update a daemon config you have
+        # to unload it and then reload the new plist. That's usually fine, but not if chef-client is trying
+        # to restart itself. If the chef-client process uses launchd or macosx_service resources to restart itself
+        # we'll end up with a stopped service that will never get started back up. Instead we use this daemon
+        # that triggers when the chef-client plist file is updated, and handles the restart outside the run.
+        launchd "com.#{Chef::Dist::SHORT}.restarter" do
+          username "root"
+          watch_paths ["/Library/LaunchDaemons/com.#{Chef::Dist::SHORT}.#{Chef::Dist::CLIENT}.plist"]
+          standard_out_path "#{::File.join(new_resource.log_directory, new_resource.log_file_name)}"
+          standard_error_path "#{::File.join(new_resource.log_directory, new_resource.log_file_name)}"
```
lamont-granquist, line 134: should probably stop using File.join and just use interpolation unless double-slashes are super important to remove, but that doesn't matter.

lamont-granquist (approved):

lib/chef/resource/chef_client_launchd.rb
tas50: I'm not sure how it would handle a double slash here

phiggins, line 134:
```suggestion
          standard_out_path ::File.join(new_resource.log_directory, new_resource.log_file_name)
          standard_error_path ::File.join(new_resource.log_directory, new_resource.log_file_name)
```

These shouldn't need interpolation.

tas50: Yep. We should turn on that RuboCop cop. I found a few in Ohai the other day using it.

lamont-granquist: oh yeah, or that.  did chefstyle not complain about useless interpolation?

tas50: https://github.com/chef/chef/pull/10394

tas50: We'd need to turn on Style/RedundantInterpolation to get this everywhere

Issue #9295, opened 2020-01-29
rhsm_register resource creates high load on Red Hat Satellite


donwlewis:
## Description
the `rhsm_register` resource seems to create quite a bit of load on a Red Hat Satellite server when a large number of machines are registered to it.  It seems that when running a `subscription-manager status` this generates several tasks on the satellite that can cause load to spike when a large number of hosts are doing it all at once.  We reached out to Red Hat to help investigate the problem, and they seemed to point to this as being the culprit.  

## Chef Version
15.4.45

## Platform Version
RHEL7

Please let me know if you need and data or testing from me.  I realize that satellite is a pretty niche thing, so I will be more than happy to help or contribute to get this resolved.

donwlewis:
Here is some more information:

* When running `subscription-manager status` on a node, this will reach out to the satellite server and run a `Katello::Host::Update` on the satellite.  Process does several things on the backend.

* when running `subscription-manager repos --list-enabled`, this also reaches out to the satellite and runs `Katello::Host::Update` on the satellite.

As you can see, if you have several nodes running these process, especially multiple times if you are enabling multiple repos, this could cause the number of tasks to build up on the satellite, increasing the load.  I am trying to see if Red Hat can provide any type of cache data that can be reviewed on the node vs having to reach out to the satellite all of the time.  Seems silly to me that has to do this.

donwlewis:
Here is more information.  I saw the following displayed when running the `rhsm_subscribe` resource when it actually executed:

```
      [execute] Internal Server Error): Required lock is already taken by other running tasks.
                Please inspect their state, fix their errors and resume them.

                Required lock: update
                Conflicts with tasks:
                  https://REDACTED/foreman_tasks/tasks/304c096b f4e6 407c 84a5 a0af02ee7b5b

                Successfully attached a subscription for: Enterprise Linux Extra Packages
```

The way I am doing the subscription attachment is I am looping through an array of subscription IDs.  Maybe not ideal, but not sure how else it should be done since you can't pass an array to the resource itself.

As far as checking subscription, we could possible use `subscription-manager list` to get the status of the subscription, as this shows the consumed base Red Hat subscription and seems to pull the data from scratch.

The best option for reviewing the repos that are enabled at this point seems to be to read through the `/etc/yum.repos.d/redhat.repo` file that gets generated when a system is subscribed.

tas50:
[...]

PR #10395, opened 2020-09-08
rhsm_register: Avoid potentially checking if we need to register twice

tas50:
Switch to shell_out
Use include? instead of a regex
Add some simple YARD
Memoize the rhsm registration check to avoid potentially running that check twice
Stop using grep to find text. Let ruby do it all

May help with:
https://github.com/chef/chef/issues/9295

### DIFF
--- a/lib/chef/resource/rhsm_register.rb
+++ b/lib/chef/resource/rhsm_register.rb
@@ -139,7 +139,7 @@ class Chef
         # @return [Boolean] is katello-ca-consumer installed
         #
         def katello_cert_rpm_installed?
-          shell_out("rpm -qa | grep katello-ca-consumer").stdout.include?("katello-ca-consumer")
+          shell_out("rpm -qa").stdout.include?("katello-ca-consumer")
         end
 
         #
--- a/lib/chef/resource/rhsm_register.rb
+++ b/lib/chef/resource/rhsm_register.rb
@@ -129,7 +129,7 @@ class Chef
         end
 
         #
-        # @return [Boolean] is katello-ca-consumer installed
+        # @return [Boolean] is the node registered with RHSM
         #
         def registered_with_rhsm?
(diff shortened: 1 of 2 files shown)


lib/chef/resource/rhsm_register.rb
```diff
@@ -121,24 +121,30 @@ class RhsmRegister < Chef::Resource
...
 
+        #
+        # @return [Boolean] is katello-ca-consumer installed
+        #
         def registered_with_rhsm?
-          # FIXME: use shell_out
-          cmd = Mixlib::ShellOut.new("subscription-manager status", env: { LANG: "en_US" })
-          cmd.run_command
-          !cmd.stdout.match(/Overall Status: Unknown/)
+          @registered ||= !shell_out("subscription-manager status").stdout.include?("Overall Status: Unknown")
```
phiggins, line 135: Was leaving off the LANG intentional?

lib/chef/resource/rhsm_register.rb
```diff
@@ -121,24 +121,30 @@ class RhsmRegister < Chef::Resource
...
       action_class do
+        #
+        # @return [Symbol] dnf_package or yum_package depending on OS release
+        #
         def package_resource
           node["platform_version"].to_i >= 8 ? :dnf_package : :yum_package
         end
 
+        #
+        # @return [Boolean] is katello-ca-consumer installed
```
phiggins, line 132: Was this supposed to have the same description as the following method?

lib/chef/resource/rhsm_register.rb
tas50: shell_out should be defaulting to en-us with the new changes

phiggins (approved):

lib/chef/resource/rhsm_register.rb
tas50: fixed

PR #10396, opened 2020-09-08
Use include? to example strings when we don't need a regex

tas50:
There's no need for match? with regex here.

### DIFF
--- a/spec/support/platform_helpers.rb
+++ b/spec/support/platform_helpers.rb
@@ -26,11 +26,11 @@ class DependencyProc < Proc
 end
 
 def ruby_64bit?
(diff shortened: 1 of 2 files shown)


spec/support/platform_helpers.rb
```diff
@@ -199,8 +199,7 @@ def selinux_enabled?
 end
 
 def suse?
-  ::File.exist?("/etc/SuSE-release") ||
```
tas50, line 202: This was only necessary on the older SLES boxes. Our new builders should all have the os-release file since we're not building for SLES 11 anymore

phiggins (approved):

PR #10401, opened 2020-09-09
Use __dir__ instead of getting the dir of __FILE__

tas50:
This is a bit easier to read

### DIFF
--- a/tasks/rspec.rb
+++ b/tasks/rspec.rb
@@ -20,7 +20,7 @@
 require "rubygems"
 require "rake"
 
-CHEF_ROOT = File.join(File.dirname(__FILE__), "..")
+CHEF_ROOT = File.join(__dir__, "..")
 
 begin
   require "rspec/core/rake_task"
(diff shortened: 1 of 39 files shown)


phiggins:
Looks like this is also running into this issue: https://bugs.ruby-lang.org/issues/13126

tas50:
Time to update that chefstyle rule!

mwrock (approved):

lamont-granquist:
for some reason "the dir of __FILE__" sounds like a royal title in my brain for some reason.

lamont-granquist (approved):

tasks/rspec.rb
```diff
@@ -20,7 +20,7 @@
 require "rubygems"
 require "rake"
 
-CHEF_ROOT = File.join(File.dirname(__FILE__), "..")
+CHEF_ROOT = File.join(__dir__, "..")
```
phiggins, line 24:
```suggestion
```

This isn't referenced anywhere.

phiggins (approved):

PR #10405, opened 2020-09-09
autoload license_acceptance/acceptor in knife loading

mwrock:
I had previously done this for the `chef-client` scenario but missed this which is pulled in when loading knife commands. The deps of license_acceptance/acceptor are pretty heavy.


tas50:
Kitchen test failures are unrelated

PR #10421, opened 2020-09-11
WIP: Update the validation of the privilege property on the windows_user_privilege resource

chef-davin:
## Description

This modifies the validation of the `privilege` property in order to ensure each value in the array or string provided exists in the array of allowed/existing windows privileges.

## Related Issue

#10420 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Issue #10420, opened 2020-09-11
Resource "windows_user_privilege" Does Not Work as Documented


jeremyciak:
## Description

Using the `windows_user_privilege` resource as documented results in a validation error.

## Chef Version

16.3.45

## Platform Version

Windows Server 2019

## Replication Case

```ruby
windows_user_privilege 'Administrators' do
  privilege %w(SeAssignPrimaryTokenPrivilege SeIncreaseQuotaPrivilege)
end
```

## Client Output

```
================================================================================
Error executing action `add` on resource 'windows_user_privilege[Administrators]'
================================================================================

Chef::Exceptions::ValidationFailed
----------------------------------
[...]

PR #10422, opened 2020-09-11
Update the validation of the privilege property on the windows_user_privilege resource

tas50:
Built off #10421

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
@@ -38,11 +38,6 @@ windows_firewall_profile "Domain" do
   action :enable
 end
 
-windows_user_privilege "BUILTIN\\Administrators" do
-  privilege %w{SeAssignPrimaryTokenPrivilege SeIncreaseQuotaPrivilege}
-  action :add
-end
-
 windows_firewall_profile "Public" do
   action :disable
 end
--- a/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
@@ -38,6 +38,11 @@ windows_firewall_profile "Domain" do
   action :enable
 end
 
+windows_user_privilege "BUILTIN\\Administrators" do
+  privilege %w(SeAssignPrimaryTokenPrivilege SeIncreaseQuotaPrivilege)
+  action :add
+end
+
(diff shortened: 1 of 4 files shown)


lib/chef/resource/windows_user_privilege.rb
```diff
@@ -137,10 +137,20 @@ class WindowsUserPrivilege < Chef::Resource
...
+              these_options = Array[n]
+            else
+              these_options = n
+            end
+
+            if (these_options - PRIVILEGE_OPTS).empty?
+              true
+            else
+              false
+            end
```
phiggins, line 151:
```suggestion
            (Array(n) - PRIVILEGE_OPTS).empty?
```

`Array()` is a function that will convert its parameter to an array. `Array[]` is the same as bare `[]`, that is, it creates an array containing its parameters.

phiggins: If the `coerce` proc is run first the conversion might not be necessary.

tas50: You're too fast here @phiggins. I already have those same changes locally. This isn't done

lib/chef/resource/windows_user_privilege.rb
```diff
@@ -137,10 +137,20 @@ class WindowsUserPrivilege < Chef::Resource
         required: true,
         coerce: proc { |v| v.is_a?(String) ? Array[v] : v },
         callbacks: {
-           "Option privilege must include any of the: #{privilege_opts}" => lambda { |v|
-             (privilege_opts & v).size == v.size
-           },
-         }
+          "Option privilege must include any of the: #{PRIVILEGE_OPTS}" => lambda { |n|
```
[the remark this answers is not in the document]
tas50: I pulled this in but changed option to property

kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
```diff
@@ -38,6 +38,11 @@
   action :enable
 end
 
+windows_user_privilege "BUILTIN\\Administrators" do
```
tas50, line 41: @chef-davin Am I missing something here. Isn't the name property the principal not the users?

phiggins:
I read #10420 and I'm confused what the actual problem was and how this PR addresses it.

phiggins (approved):

tas50:
There's more to that issue than this fix, but this resolves some strange code in the validation, gives us users as strings and adds specs

Issue #10425, opened 2020-09-11
Azure Pipelines CI job broken


phiggins:
The `Post-job: Checkout chef/chef@master to s` step of the `ubuntu_kitchen_tests` job of the Azure pipelines CI has been failing since #10171 was merged: https://dev.azure.com/chef-software/chef/_build?definitionId=10

 It fails with a message like this every time:

```
Starting: Checkout chef/chef@master to s
==============================================================================
Task         : Get sources
Description  : Get sources from a repository. Supports Git, TfsVC, and SVN repositories.
Version      : 1.0.0
Author       : Microsoft
Help         : [More Information](https://go.microsoft.com/fwlink/?LinkId=798199)
==============================================================================
Failed to create CoreCLR, HRESULT: 0x80004005
##[error]Exit code 137 returned from process: file name '/home/vsts/agents/2.174.1/bin/Agent.PluginHost', arguments 'task "Agent.Plugins.Repository.CleanupTask, Agent.Plugins"'.
Finishing: Checkout chef/chef@master to s
```

In #10424 I was able to get this passing again by commenting out the mount recipe in `kitchen-tests/cookbooks/end_to_end/recipes/linux.rb`: https://github.com/chef/chef/pull/10424/files#diff-062613025481c246313ddc7ea9194c86R127

I'm not sure how to debug this or fix those tests.

phiggins:
Not relevant since this test job was removed in #10434.

PR #10426, opened 2020-09-11
Disable and unmount devices in mount end-to-end test.

phiggins:

#10425


phiggins:
Didn't work.

PR #10432, opened 2020-09-15
Add examples to the ohai resource

tas50:
Make it a bit easier to generate the docs for this resource

### DIFF
--- a/lib/chef/resource/ohai.rb
+++ b/lib/chef/resource/ohai.rb
@@ -29,7 +29,7 @@ class Chef
 
       provides :ohai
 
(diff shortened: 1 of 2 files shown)


lib/chef/resource/ohai.rb
```diff
@@ -31,6 +31,49 @@ class Ohai < Chef::Resource
...
+      examples <<~DOC
+      Reload All Ohai Plugins
+
+      ```ruby
+      ohai 'reload' do
+        action :reload
+      end
+      ```
+
+      Reload A Single Ohai Plugins
```
phiggins, line 43:
```suggestion
      Reload A Single Ohai Plugin
```

phiggins (approved):

lib/chef/resource/ohai.rb
tas50: Fixed before I saw this

PR #10436, opened 2020-09-16
Move TrainTransport to ChefConfig

lamont-granquist:
Also pull in some helpers from the Ohai work so that Ohai can
use them later.

This is code rearrangement to support the remote ohai work but
does not change any existing behavior yet.

### DIFF
--- a/chef-config/lib/chef-config/mixin/train_transport.rb
+++ b/chef-config/lib/chef-config/mixin/train_transport.rb
@@ -1,0 +1,141 @@
+# Author:: Bryan McLellan <btm@loftninjas.org>
+# Copyright:: Copyright (c) Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


chef-config/lib/chef-config/mixin/train_transport.rb
```diff
@@ -0,0 +1,141 @@
...
+        train_config[:logger] = logger
+
+        # Train handles connection retries for us
+        Train.create(protocol, train_config)
+      rescue SocketError => e # likely a dns failure, not caught by train
+        e.message.replace "Error connecting to #{train_config[:target]} - #{e.message}"
+        raise e
+      rescue Train::PluginLoadError
+        logger.error("Invalid target mode protocol: #{protocol}")
+        exit(false)
```
phiggins, line 133:
```suggestion
        exit(1)
```

TIL: 

```
Petes-MBP:chef pete$ ruby -e "exit(false)" ; echo $?
1
```

spec/unit/train_transport_spec.rb
```diff
@@ -19,32 +19,33 @@
 require "spec_helper"
 
 describe Chef::TrainTransport do
+  let(:transport) { Chef::TrainTransport.new(Chef::Log) }a
```
phiggins, line 22:
```suggestion
  let(:transport) { Chef::TrainTransport.new(Chef::Log) }
```

phiggins (approved):
Seems like a good intermediate step 👍

PR #10437, opened 2020-09-16
Resolve Lint/RedundantRequireStatement & Style/RedundantCondition warnings

tas50:
**Lint/RedundantRequireStatement:** These are required by Ruby out of the box so lets avoid trying to require them again.

**Style/RedundantCondition:** This lets us nuke a pile of ternary operators and other checks for values that weren't necessary


phiggins (approved):
Whoa

PR #10439, opened 2020-09-16
Resolve Style/RedundantSort warnings

tas50:
Avoid a sort then last and just grab the max.

### DIFF
--- a/lib/chef/mixin/versioned_api.rb
+++ b/lib/chef/mixin/versioned_api.rb
@@ -50,8 +50,7 @@ class Chef
           # in which case we'll just start with the highest version and see what happens
           ServerAPIVersions.instance.min_server_version.nil? || (version >= ServerAPIVersions.instance.min_server_version && version <= ServerAPIVersions.instance.send(type))
         end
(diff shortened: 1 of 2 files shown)


lib/chef/mixin/versioned_api.rb
```diff
@@ -50,8 +50,7 @@ def get_class_for(type)
           # in which case we'll just start with the highest version and see what happens
           ServerAPIVersions.instance.min_server_version.nil? || (version >= ServerAPIVersions.instance.min_server_version && version <= ServerAPIVersions.instance.send(type))
         end
-          .sort { |a, b| a.send(:minimum_api_version) <=> b.send(:minimum_api_version) }
-          .last
+          .max { |a, b| a.send(:minimum_api_version) <=> b.send(:minimum_api_version) }
```
phiggins, line 53:
```suggestion
          .max_by { |a| a.send(:minimum_api_version) }
```

Assuming the `send` is necessary, otherwise `.max_by(&:minimum_api_version)` would work.

phiggins (approved):

Issue #10352, opened 2020-08-26
osx_profile no longer idempotent


chilcote:
## Description
osx_profile resource is no longer idempotent.  On every run I get (using the loginwindow profile as an example):

## Chef Version
16.4.41

## Platform Version
macOS 10.15.6

## Replication Case
Using the open source cookbook here https://github.com/pinterest/it-cpe-cookbooks/tree/master/cpe_loginwindow

And feeding it a node attribute like so:

```
{
  'LoginwindowText' =>
    'I think chef done broke it',
}.each { |k, v| node.default['cpe_loginwindow'][k] = v }
```

## Client Output
```
  * osx_profile[com.facebook.chef.loginwindow] action install
    - install profile com.facebook.chef.loginwindow
```

(16.2.50 for comparison)
```
  * osx_profile[com.facebook.chef.loginwindow] action install (up to date)
```

PR #10444, opened 2020-09-17
Fix idempotency in the osx_profile resource and avoid writing data to disk

tas50:
This had 2 different issues that were causing it to not fetch the current resource state properly. The resource didn't fail, but it would converge each time because it believed the profile was not installed.

1) The profiles command was failing to write the profile state to disk because the temp file already existed. We don't actually need to write to disk and we can avoid that extra disk space / IO entirely by writing to stdout and parsing that.

2) Our fetching of the current profile state from node.state was always returning nil because of an extra logger at the end of the method. That has been moved to the top so that we return the objects correctly.

### DIFF
--- a/lib/chef/resource/osx_profile.rb
+++ b/lib/chef/resource/osx_profile.rb
@@ -308,12 +308,12 @@ class Chef
         #
 
         def get_installed_profiles(update = nil)
+          logger.trace("Saving profile data to node.run_state")
           if update
             node.run_state[:config_profiles] = query_installed_profiles
           else
             node.run_state[:config_profiles] ||= query_installed_profiles
           end
-          logger.trace("Saved profiles to run_state")
         end
 
         def query_installed_profiles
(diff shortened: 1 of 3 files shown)


phiggins (approved):

lamont-granquist (approved):

lib/chef/resource/osx_profile.rb
```diff
@@ -308,19 +308,18 @@ def remove_profile
...
 
         def get_installed_profiles(update = nil)
+          logger.trace("Saving profile data to node.run_state")
           if update
             node.run_state[:config_profiles] = query_installed_profiles
           else
             node.run_state[:config_profiles] ||= query_installed_profiles
           end
-          logger.trace("Saved profiles to run_state")
         end
```
phiggins, line 317: I feel like this function should be changed to return a value or cache the state, but that's probably outside the scope of fixing this for the next release.

tas50: Also keep in mind this whole resource is basically EOL since it won't work on macOS 11. We need it to keep working for a few more years.

tas50:
Fixes #10352

PR #10449, opened 2020-09-20
Remove unused method

007lva:
I've just found a private method that is not being used.


lamont-granquist (approved):

lamont-granquist:
Looks like the code that used it was deleted ~10 years ago, so hopefully we don't need this anymore.  thanks!

PR #10451, opened 2020-09-21
Update require gating to include chef-utils/chef-config & gate more

tas50:
Update the cop config to handle those as well

### DIFF
--- a/chef-utils/spec/spec_helper.rb
+++ b/chef-utils/spec/spec_helper.rb
@@ -1,4 +1,4 @@
-require "chef-utils" unless defined?(ChefUtils)
+require "chef-utils"
 
 # FIXME: dynamically generate this for accuracy
 HELPER_MODULES = [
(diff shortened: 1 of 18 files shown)


mwrock (approved):

lamont-granquist (approved):

lamont-granquist:
busted unit tests

chef-utils/spec/spec_helper.rb
```diff
@@ -1,4 +1,4 @@
-require "chef-utils"
+require "chef-utils" unless defined?(ChefUtils)
```
lamont-granquist, line 1: this might not work at all.

but i think you may have discovered why i was using `unless defined?(ChefUtils::CANARY)`

mwrock:
Hmm. looks like tests are failing due to `uninitialized constant ChefUtils::DSL` which may be related.

chef-utils/spec/spec_helper.rb
tas50: I updated chefstyle to enforce that

tas50:
Fixed things, but excluding specs and using ChefUtils::CANARY

PR #10463, opened 2020-09-24
Use Ruby 2.6 endless Range syntax

007lva:
I find `array[n..]` more readable than `array[n..-1]` and thought that would not be difficult to applied here.

https://blog.bigbinary.com/2018/07/04/ruby-2-6-adds-endless-range.html


tas50 (approved):

tas50:
Thanks @007lva

tduffield:
Just as a note, this breaks anyone who might use this library with 2.5 (i.e. license_scout)

Issue #2908, opened 2015-02-13
mount provider changes order of existing entries


scarpe01:
chef-solo --version
Chef: 11.12.4

This resulted in some seriuous headaches the past few days. The problem is pretty clear, if mount sees that a resource should be enabled, but does not match current state, it disables then enables the resource. This changes the order of entries in fstab which can have side effects of changing when in the boot process filesystems are loaded.

Here is how we produced the problem:
Our initial config has /var and /var/log setup with default options out of the installer.
Our requirement is to add a few options.

After we add our options with the mount resource, /var/log is defined before /var, causing it to not mount until netfs runs. Since rsyslogd starts up at S12 (netfs is at S26), so it begins logging on /var and then its logs become invisiable when the correct /var/log is mounted over them.,

In short, if entries exist before the provider is called, it should either attempt to preserve that order, especially for volumes important to the system.

lamont-granquist:
- you need to replicate this on chef 12, there's been changes to the mount provider since 11.12.4
- you need to include your distro and version since there's different mount providers with different behaviors

scarpe01:
Platform is OL6. sorry I buried that in the text.

I will give chef12 a try but, looking at the current code on github, I can see the issue is still there on the master branch anyway:
https://github.com/chef/chef/blob/master/lib/chef/provider/mount/mount.rb lines 153 to 160, there is clearly no logic to do anything but an append

Thopugh as you say, there may be different providers etc, so I will give it a try with the latest anyway, its easy enough to try.

lamont-granquist:
OL6 == Oracle 6.x?

scarpe01:
from chef-12.0.3-1.x86_64:

Oh yah, sorry we use that shorthand alot. Latest OL6 fully updated 6.6 latest. 

Before:
..
/dev/mapper/rootvg-varlv /var                    ext4    defaults        1 2
/dev/mapper/rootvg-loglv /var/log                ext4    defaults        1 2
..
proc                    /proc                   proc    defaults        0 0

After:
...
/dev/mapper/rootvg-loglv /var/log                ext4    defaults        1 2
...
/dev/mapper/rootvg-varlv /var auto defaults 0 2
...

Confirmed.

bby-bishopclark:
Can confirm this is still occurring.

OS:  EL7.7  (RHEL/CentOS and friends) 
Chef Client : 14.13.11

Results:
- all changed lines are created at the tail of fstab
- old lines ARE NOT REMOVED

[...]

PR #10472, opened 2020-09-30
mount: Changes to fix creating multiple entries in fstab

antima-gupta:

## Description
- Updated `enabled?` method in mount provider.
- Added `update_or_delete_fs` method to fix the order change
- Added spec for the changes
- Fixed test case failure.

## Related Issue
#2908 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/mount/mount.rb
+++ b/lib/chef/provider/mount/mount.rb
@@ -166,7 +166,7 @@ class Chef
             # The current options don't match what we have, so
             # update the last matching entry with current option
             # and order will remain the same.
-            update_or_delete_fs(true)
+            edit_fstab
           else
             ::File.open("/etc/fstab", "a") do |fstab|
               fstab.puts("#{device_fstab} #{@new_resource.mount_point} #{@new_resource.fstype} #{@new_resource.options.nil? ? default_mount_options : @new_resource.options.join(",")} #{@new_resource.dump} #{@new_resource.pass}")
@@ -176,7 +176,7 @@ class Chef
         end
 
         def disable_fs
-          update_or_delete_fs
+          edit_fstab(:remove)
         end
 
         def network_device?
@@ -248,7 +248,7 @@ class Chef
         end
 
         # It will update or delete the entry from fstab.
-        def update_or_delete_fs(update_fs = false)
+        def edit_fstab(action = :update)
           if @current_resource.enabled
             contents = []
 
@@ -256,7 +256,7 @@ class Chef
             ::File.readlines("/etc/fstab").reverse_each do |line|
               if !found && line =~ /^#{device_fstab_regex}\s+#{Regexp.escape(@new_resource.mount_point)}\s/
                 found = true
-                if update_fs
+                if action == :update
                   logger.trace("#{@new_resource} is updated with new content in fstab")
                   contents << ("#{device_fstab} #{@new_resource.mount_point} #{@new_resource.fstype} #{@new_resource.options.nil? ? default_mount_options : @new_resource.options.join(",")} #{@new_resource.dump} #{@new_resource.pass}")
                 else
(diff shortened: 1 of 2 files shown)


phiggins (approved):
Looks good to me 👍

spec/unit/provider/mount/mount_spec.rb
```diff
@@ -247,15 +247,6 @@
       expect(@provider.current_resource.enabled).to be_falsey
     end
 
-    it "should set enabled to false if the mount point is not last in fstab" do
```
tas50, line 250: It seems like we still probably want this test. I'm confused why we're removing it.

tas50:
Can we get a test added for the actual scenario in the linked issue so we make sure this solves that problem and we don't introduce a regression in the future.

antima-gupta:
@tas50 Do I need to add test scenario for the multiple entry in fstab fix?

tas50:
@antima-gupta We definitely want to test this scenario and understand why it was Ok to remove a previous test here.

spec/unit/provider/mount/mount_spec.rb
antima-gupta: I removed the [code](https://github.com/chef/chef/pull/10472/files#diff-ee080140ef5e117fe81d326426a69ea8b1ef913dee4ed1267fe07350f7b59610L73-L76) because this was causing multiple entry in `fstab`, that's why I need to remove this test.

antima-gupta:
@tas50 I have added test please have a look into this.

lib/chef/provider/mount/mount.rb
```diff
@@ -267,6 +246,35 @@ def mount_options_unchanged?
             @current_resource.pass == @new_resource.pass
         end
 
+        # It will update or delete the entry from fstab.
+        def update_or_delete_fs(update_fs = false)
```
tas50, line 250: If we're going to go down the route of having a helper that actually does the work it should just be `update_fstab` and takes an action value. We've done that same sort of thing in other places. When you see how this is used right now it's not clear what the parameter is actually doing unless you look at the definition

tas50: It becomes something like this

```ruby
edit_fstab(:remove)
```

or 

```ruby
edit_fstab(:update)
```

antima-gupta: @tas50 Agreed, I have updated it. Please have a look.

lib/chef/provider/mount/mount.rb
```diff
@@ -267,6 +246,35 @@ def mount_options_unchanged?
             @current_resource.pass == @new_resource.pass
         end
 
+        # It will update or delete the entry from fstab.
+        def edit_fstab(action = :update)
```
tas50, line 250: Let's remove the default action value so it's more obvious above what a plain call to edit_fstab does. Ideally you can see where a method is called and understand what it's doing.

antima-gupta: @tas50 I have updated it, please have a look.

lib/chef/provider/mount/mount.rb
```diff
@@ -268,6 +247,35 @@ def mount_options_unchanged?
             @current_resource.pass == @new_resource.pass
         end
 
+        # It will update or delete the entry from fstab.
+        def edit_fstab(action)
```
lamont-granquist, line 251: this should not be called `action` since that collides with the `action` on the resource, and `:update` and `:remove` are not a valid actions for the mount provider.  different concepts must have different names.

you can just call this `def edit_fstab(remove: false)` and then do `if remove` later, then call it with `edit_fstab(remove: true)` with a named parameter to make it clear what is happening.

lamont-granquist (changes requested):

lib/chef/provider/mount/mount.rb
antima-gupta: @lamont-granquist I update this, please have a look.

lamont-granquist (approved):

PR #10473, opened 2020-09-30
Update to the windows_audit_policy resource to fix a bug on failure-only auditing

chef-davin:
## Description

A bug was discovered by @sam1el where the resource would fail to set the policy when wanting to only audit on failure events and not on success events.  This fixes the powershell regex used in the Select-String cmdlet so that the resource will correctly identify if an advanced audit policy is failure only or not.

It also moves the methods into the action_class as is appropriate.

## Types of changes

## Checklist:


kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
```diff
@@ -38,10 +38,39 @@
...
+  privilege %w{SeAssignPrimaryTokenPrivilege SeIncreaseQuotaPrivilege}
+  action :add
+end
+
 windows_firewall_profile "Public" do
   action :disable
 end
 
+windows_audit_policy "Update Some Advanced Audit Policies to Success and Failure" do
+  subcategory subcategory ["Application Generated", "Application Group Management", "Audit Policy Change"]
```
phiggins, line 51:
```suggestion
  subcategory ["Application Generated", "Application Group Management", "Audit Policy Change"]
```

I suspect that all of these need to change and might be the cause of some of the CI failures.

lib/chef/resource/windows_audit_policy.rb
```diff
@@ -225,6 +201,32 @@ def option_configured?(option_name, option_setting)
...
       end
+
+      action_class do
+        def subcategory_configured?(sub_cat, success_value, failure_value)
+          setting = if success_value && failure_value
+                      "Success and Failure$"
+                    elsif success_value && !failure_value
+                      "Success$"
+                    elsif !success_value && failure_value
+                      "#{sub_cat} \\s+ Failure$"
```
phiggins, line 212:
```suggestion
                      "#{sub_cat}\\s+Failure$"
```

This was saying there should always be 3+ spaces in the middle. Was that intentional?

phiggins (approved):
The code looks good to me, I don't really have an opinion on the correctness of the fix though.

PR #10476, opened 2020-10-01
add interpreter arg to powershell_exec allowing it to run powershell core

### LINKED ISSUE #10430: Support cross platform pwsh in powershell_exec
Related to epic #10330

Tentatively users of powershell_exec may specify an "interpreter" as an optional second string argument to powershell_exec. The default will be powershell. If pwsh is the interpreter, powershell exec will invoke a .net core based wrapper for creating a cross platform powershell based runspace.

"interpreter" may not be the correct path forward here so that implementation detail is open. We could also support a boolean flag.

mwrock:
fixes #10430 

This adds an optional `interpreter` arg to `powershell_exec` which defaults to `:powershell` but can be set to `:pwsh`. `:powershell` will perform the same behavior that this mixin performs today. Passing `:pwsh` will use the powershell core based wrapper.

Note that this does add a bunch of dlls which is basically the .net core runtime. These were generated by https://github.com/chef/chef-powershell-shim/pull/12.


lib/chef/mixin/powershell_exec.rb
```diff
@@ -90,22 +96,35 @@
...
-        cmd = Chef::PowerShell.new(script)
+      def powershell_exec!(script, interpreter = :powershell)
+        case interpreter
+        when :powershell
+          cmd = Chef::PowerShell.new(script)
+        when :pwsh
+          cmd = Chef::Pwsh.new(script)
+        else
+          raise ArgumentError, "Expected interpreter of :powershell or :pwsh"
+        end
```
phiggins, line 127:
```suggestion
        cmd = powershell_exec(script, interpreter)
```

Any reason to not lean on the other method?

mwrock: oh of course!

lib/chef/pwsh.rb
```diff
@@ -0,0 +1,59 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+class Chef
+  class Pwsh < Chef::PowerShell
+
+    @@wrapper_dll ||= Dir.glob("#{RbConfig::CONFIG["bindir"]}/**/Chef.PowerShell.Wrapper.Core.dll").last
```
phiggins, line 21: Rather than use a class variable for this and overwrite the initializer, I think I'd rather see it as a class method that's called in the superclass's `exec` or make it an argument that gets passed to that function. Class variables in general have kind of strange behavior and I try to avoid them, and since this is doing "work" it could be made lazily memoized to avoid having to pay that cost just for loading the file.

jaymzh:
perhaps a dumb question, but should some of this get rolled into `Mixlib::ShellOut` instead?

mwrock:
@jaymzh the main reason for calling `powershell_exec` is to avoid shelling out. The startup cost of a powershell process is rather expensive. This allows us to interact with a dll wrapper around the powershell plumbing directly inside ruby.exe resulting in much better performance than shelling out to powershell.exe or pwsh.exe.

lib/chef/pwsh.rb
mwrock: yeah this felt totally weird. sadly all of my ruby knowledge around scoping and the proper way to do class initialization has atrophied. will make this better.

phiggins (approved):
I have some code style reckons but overall this looks good. I will try it out manually once my windows vm is finished rebuilding. 😅

jaymzh:
@mwrock I understand that. But it seems generically useful beyond chef. I would live to be able to use this from Ohai, for example. If Mixlib::ShellOut exposed cmdlets natively like this, it'd be generically useful. Assuming i understand it properly.

mwrock:
oh I see @jaymzh . I'm not sure mixlib shellout is the place because this feels "different". But I could see the value in making it into a separate gem altogether.

jaymzh:
It feels like you'd re-invent a bunch of what's in M::SO - it's just a way to access shell even if it doesn't fork a shell. But a separate gem (or even chefutils) would be useful

lib/chef/pwsh.rb
mwrock: ok I moved it from a class variable to a class method. I still had to override the instance variable because I don't think you can override a class method in a subclass??

tas50:
As much as I'd like to move this into Mixlib::Shellout right now it's incompatible with many older Windows releases and 32bit windows releases, which makes it problematic. We need to work out our platform compatibility before we can push this into shellout.

PR #10478, opened 2020-10-01
add interpreter arg to powershell_out allowing it to call pwsh.exe

### LINKED ISSUE #10429: Add optional second interpreter arg to powershell_out to support pwsh
Related to epic #10330 

Users of `powershell_out` may specify an "interpreter" as an optional second string argument to `powershell_out`. The default will be `powershell.exe`. This will not break current instances of `powershell_out` because it currently expects the first argument to be the script to execute and the optional last argument to be a hash of options. This will continue to be the case.

mwrock:
fixes #10429

### DIFF
--- a/spec/support/platform_helpers.rb
+++ b/spec/support/platform_helpers.rb
@@ -243,5 +243,10 @@ end
 
 def choco_installed?
   result = ShellHelpers.powershell_out("choco --version")
-  result.stderr.empty? ? true : false
+  result.stderr.empty?
+end
+
+def pwsh_installed?
+  result = ShellHelpers.powershell_out("pwsh.exe --version")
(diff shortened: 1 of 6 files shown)


spec/support/platform_helpers.rb
```diff
@@ -245,3 +245,8 @@ def choco_installed?
   result = ShellHelpers.powershell_out("choco --version")
   result.stderr.empty? ? true : false
 end
+
+def pwsh_installed?
+  result = ShellHelpers.powershell_out("pwsh.exe --version")
+  result.stderr.empty? ? true : false
```
phiggins, line 251:
```suggestion
  result.stderr.empty?
```

phiggins (approved):

PR #10482, opened 2020-10-04
Remove extra safe navigation

tas50:
This is not needed here.


lamont-granquist (approved):

tas50:
We can turn on that new rubocop rule now

PR #10484, opened 2020-10-05
fix specs on powershell v4 and below

mwrock:
Some of our tests are interrogating the `PSEdition` property of `$PSVersionTable`. This prop does not exist in versions of powershell under 5.1. So we either check the version of the existence of `CLRVersion`.


tas50:
@mwrock Looks good other than the minor chefstyle issue

phiggins (approved):

PR #10487, opened 2020-10-05
Improve handling of multilib packages in yum_package

### LINKED ISSUE #10485: yum_package cannot :install libstdc++.i686 in common situations
It is currently not possible to use the `:install` action of `yum_package` to ensure that multilib packages (those available as multiple packages for different architectures, such as `libstdc++.i686` and `libstdc++.x86_64`) are always present on a system; it fails in common scenarios.  (`dnf_package` probably has the same problem.)

## Description

I need to ensure that a number of prerequisite packages, including `libstdc++.i686`, are present during a converge before I begin to configure another application.  Ideally, I could just declare `package 'libstdc++.i686'` in the cookbook, but this often produces a YUM error.

`libstdc++.x86_64` gets installed with the core packages on every CentOS system.  When `yum_package` installs `libstdc++.i686`, it tries to install the latest version available.  If the version of `libstdc++.i686` that `yum_package` tries to install does not match the version of `libstdc++.x86_64` already present on the system, `yum` produces an error regarding a multilib version mismatch.  (Note: Where I talk about package version, I'm referring to the whole EVR (Epoch:Version-Release) of an RPM, not just an RPM's "version" attribute.)

I know of two methods for addressing this with YUM.  The first is to install the same version of the `libstdc++.i686` package as is already installed for `libstdc++.x86_64`.  The second is to pass both packages to yum in the install command, such as `yum install libstdc++.i686 libstdc++.x86_64`, in which case `yum` will install `libstdc++.i686` and update `libstdc++.x86_64` to a matching version at the same time.  The replication cases show the challenges with trying to make chef do each of these things, below.

Regardless whether I can find a good work around for this problem, I believe that a basic resource definition like `package 'libstdc++.i686'` should not fail on supported platforms when the package is readily available to install.  I am considering this behavior to be a bug.

## Chef Version
16.5.77

## Platform Version
All platforms that use YUM.  I used CentOS 7 to replicate, below.

## Replication Case

1. Create  the following files:
   * `metadata.rb` containing:

      ```ruby
      name 'example'
      version '1.0'
      ```

   * `recipes/scenario1.rb` containing:

      ```ruby
      package 'libstdc++.i686'
      ```

   * `recipes/scenario2.rb` containing:

      ```ruby
      package 'libstdc++' do
        arch 'i686'
        pkg = node['packages']['libstdc++']
        if pkg.key?('versions')
          pkg = pkg['versions'].find { |p| p['arch'] == 'x86_64' }
        end
        version "#{pkg['version']}-#{pkg['release']}"
      end
      ```

   * `recipes/scenario3.rb` containing:

      ```ruby
      package [ 'libstdc++.i686', 'libstdc++.x86_64' ]
      ```

   * `kitchen.yml` containing:

      ```yaml
      driver: { name: vagrant }
      provisioner: { name: chef_zero }
      platforms: [ { name: centos-7.6 } ]
      suites:
        - name: scenario1
          run_list: [ 'recipe[example::scenario1]' ]
        - name: scenario2
          run_list: [ 'recipe[example::scenario2]' ]
        - name: scenario3
          run_list: [ 'recipe[example::scenario3]' ]
      ```

1. Run `kitchen converge scenario1-centos-76` (see gist for output).

   `yum` will produce a large error message and `chef-client` will fail.  Specifically note the last few lines of output:

   ```
   Protected multilib versions: libstdc++-4.8.5-39.el7.i686 != libstdc++-4.8.5-36.el7_6.2.x86_64
   ---- End output of ["yum", "-y", "install", "libstdc++-0:4.8.5-39.el7.i686"] ----
   Ran ["yum", "-y", "install", "libstdc++-0:4.8.5-39.el7.i686"] returned 1
   ```

   Even though the package resource definition did not specify a package version, `chef-client` found the latest candidate version and specifically required that version when it ran `yum -y install libstdc++-0:4.8.5-39.el7.i686`.  `yum` aborted because installing the requested version would have produced a multilib version mismatch between `libstdc++.i686` and `libstdc++.x86_64`.  This is a rudimentary operation that I believe Chef should be able to handle successfully on the platforms it supports.

1. Run `kitchen converge scenario2-centos-76` (see gist for output).

   It will also fail.  This is my attempt to tell Chef to install the same version of `libstdc++.i686` as is being used by the already installed `libstdc++.x86_64`.  It works in many cases, but this time it fails because `libstdc++-4.8.5-36.el7_6.2.i686` is no longer in CentOS's primary yum repository; it has been rotated to the `C7.6.1810-updates` vault.  To install the package on this platform, you would first need to update your CentOS repo definitions so you could see the vault with `yum -y update centos-release`.  You could then install the package with `yum -y --enablerepo=C7.6.1810-updates install libstdc++-4.8.5-36.el7_6.2.i686`.

   This mostly seems like a CentOS issue, but in this case I can't easily predetermine whether a package has already been rotated out of a repository from the cookbook level, nor is this an error I can easily handle because `chef-client` aborts.  Also note that the code is getting hard to read; making chef choose a package version that `yum` can successfully install should not be this complicated.

1. Run `kitchen converge scenario3-centos-76` (see gist for output).

   It will fail with another multilib version conflict.  In this case, I tried to tell Chef to have yum install both packages at the same time.  Chef determined that `libstdc++.x86_64` was already installed and left it out of the request, as you can see from the last line of output `Ran ["yum", "-y", "install", "libstdc++-0:4.8.5-39.el7.i686"] returned 1`.  If I also set action `:update` for the resource, `chef-client` will succeed in installing a missing package, but it will also update the packages even when both are already present, which is not what I am trying to accomplish.

## Client Output
https://gist.github.com/karmix/4071eb9494adba4613b71714c8565758

## Stacktrace
Not Applicable.

karmix:
## Description

When `yum_package` is used to install a package, and there is already a package with the same name but different architecture installed on the system, and the package version that `yum_package` chooses for the new package is different than the version of the already installed package, the `yum` command generates an multilib version error and `chef-client` aborts.  This PR makes two behavioral changes that help prevent `yum_package` from triggering errors regarding multilib version conflicts when calling out to `yum` in this situation.

1. When `yum_package` searches for candidates to satisfy an `:install` action that has no version constraints, it tries to install the latest version of the package available.  The first part of this PR makes `yum_package` prefer those candidates with versions that match any installed packages that have the same name (but different architectures).

1.  When `yum_package` is handling packages in the same `:install` action (i.e. `yum_package` [ 'pkg.arch1', 'pkg.arch2', 'other-packages...' ]), `yum_package` only passes those packages that are not installed on the system to `yum` for installation.  The second part of this PR identifies when the version chosen for the new package is incompatible with the version of the installed package and instructs `yum` to install both packages in this case.  When told to install both packages, `yum` responds by updating the installed package to the same version as the new package being installed, avoiding a conflict.

There is an edge case where if a user sets `protected_multilib=0` in their YUM configuration, `yum` will install different versions of multilib packages without generating an error.  The yum documentation discourages this configuration wherever `protected_multilib` is discussed.  If a user has this configuration, the first part of this PR will detect it and resort to always choosing the most recent version of installation candidates.  The second part of this PR, however, does not have visibility into the YUM configuration; it will sill update an installed package to maintain version consistency with new packages being installed when both packages are defined in the resource's package list.

## Related Issue

Resolves #10485

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/package.rb
+++ b/lib/chef/provider/package.rb
@@ -412,57 +412,58 @@ class Chef
           begin
             target_version_array = []
             each_package do |package_name, new_version, current_version, candidate_version|
-              case action
-              when :upgrade
-                if version_equals?(current_version, new_version)
-                  # this is an odd use case
-                  logger.trace("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
-                  target_version_array.push(nil)
-                elsif version_equals?(current_version, candidate_version)
-                  logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
-                  target_version_array.push(nil)
-                elsif candidate_version.nil?
-                  logger.trace("#{new_resource} #{package_name} has no candidate_version to upgrade to")
-                  target_version_array.push(nil)
-                elsif current_version.nil?
-                  logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
-                  target_version_array.push(candidate_version)
-                elsif !allow_downgrade && version_compare(current_version, candidate_version) == 1
-                  logger.trace("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{candidate_version}. Skipping...)")
-                  target_version_array.push(nil)
-                else
-                  logger.trace("#{new_resource} #{package_name} is out of date, will upgrade to #{candidate_version}")
-                  target_version_array.push(candidate_version)
-                end
-
-              when :install
-                if new_version
-                  if version_requirement_satisfied?(current_version, new_version)
-                    logger.trace("#{new_resource} #{package_name} #{current_version} satisfies #{new_version} requirement")
-                    target_version_array.push(nil)
-                  elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
-                    logger.warn("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{new_version}. Skipping...)")
-                    target_version_array.push(nil)
-                  else
-                    logger.trace("#{new_resource} #{package_name} #{current_version} needs updating to #{new_version}")
-                    target_version_array.push(new_version)
-                  end
-                elsif current_version.nil?
-                  logger.trace("#{new_resource} #{package_name} not installed, installing #{candidate_version}")
-                  target_version_array.push(candidate_version)
-                else
-                  logger.trace("#{new_resource} #{package_name} #{current_version} already installed")
-                  target_version_array.push(nil)
-                end
-
-              else
-                # in specs please test the public interface provider.run_action(:install) instead of provider.action_install
-                raise "internal error - target_version_array in package provider does not understand this action"
-              end
+              target = target_version(package_name, new_version, current_version, candidate_version)
+              target_version_array.push(target == current_version ? nil : target)
             end
-
             target_version_array
           end
+      end
+
+      # Returns the version a package should have after the install/upgrade operation is complete.
+      #
+      # @return [String] target package version
+      def target_version(package_name, new_version, current_version, candidate_version)
+        case action
+        when :upgrade
+          if version_equals?(current_version, new_version)
+            # this is an odd use case
+            logger.trace("#{new_resource} #{package_name} #{new_version} is already installed -- you are equality pinning with an :upgrade action, this may be deprecated in the future")
+          elsif version_equals?(current_version, candidate_version)
+            logger.trace("#{new_resource} #{package_name} #{candidate_version} is already installed")
+          elsif candidate_version.nil?
+            logger.trace("#{new_resource} #{package_name} has no candidate_version to upgrade to")
+          elsif current_version.nil?
+            logger.trace("#{new_resource} has no existing installed version. Installing install #{candidate_version}")
+            return candidate_version
+          elsif !allow_downgrade && version_compare(current_version, candidate_version) == 1
+            logger.trace("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{candidate_version}. Skipping...)")
+          else
+            logger.trace("#{new_resource} #{package_name} is out of date, will upgrade to #{candidate_version}")
+            return candidate_version
+          end
+
+        when :install
+          if new_version
+            if version_requirement_satisfied?(current_version, new_version)
+              logger.trace("#{new_resource} #{package_name} #{current_version} satisfies #{new_version} requirement")
+            elsif current_version && !allow_downgrade && version_compare(current_version, new_version) == 1
+              logger.warn("#{new_resource} #{package_name} has installed version #{current_version}, which is newer than available version #{new_version}. Skipping...)")
+            else
+              logger.trace("#{new_resource} #{package_name} #{current_version} needs updating to #{new_version}")
+              return new_version
+            end
+          elsif current_version.nil?
+            logger.trace("#{new_resource} #{package_name} not installed, installing #{candidate_version}")
+            return candidate_version
+          else
+            logger.trace("#{new_resource} #{package_name} #{current_version} already installed")
+          end
+
+        else
+          # in specs please test the public interface provider.run_action(:install) instead of provider.action_install
+          raise "internal error - target_version in package provider does not understand this action"
+        end
+        current_version
       end
 
       # Check the list of current_version_array and candidate_version_array. For any of the
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -220,7 +220,8 @@ class Chef
           @available_version[index] ||= if new_resource.source
                                           resolve_source_to_version_obj
                                         else
-                                          python_helper.package_query(:whatavailable, package_name_array[index], version: safe_version_array[index], arch: safe_arch_array[index], options: options)
+                                          mc = (action == :install) || nil
+                                          python_helper.package_query(:whatavailable, package_name_array[index], version: safe_version_array[index], arch: safe_arch_array[index], options: options, multilibcompat: mc)
                                         end
 
           @available_version[index]
--- a/lib/chef/provider/package/yum.rb
+++ b/lib/chef/provider/package/yum.rb
@@ -153,6 +153,41 @@ class Chef
 
         private
 
+        def target_version(package_name, new_version, current_version, candidate_version)
+          return super if action != :install || current_version.nil? || new_version
+
+          # Install-only packages cannot produce version conflicts.
+          package_idx = package_name_array.find_index(package_name)
+          return super if python_helper.install_only_packages(installed_version(package_idx).name)
+
+          # Before telling the user that no action is being taken because the package is already installed, make sure
+          # that it doesn't require upgrading to maintain multilib version consistency with another package being
+          # installed in the same package set.
+          package_name_array.each_with_index do |_other_name, other_idx|
+            if multilib_update_required(package_idx, other_idx)
+              logger.trace("#{new_resource} #{package_name} needs updating to #{candidate_version} to maintain multilib version consistency")
+              return candidate_version
+            end
+          end
+
+          super
+        end
+
+        # Returns true if other_idx is a variant of package_idx for a different architecture and package_idx needs to
+        # update so that other_idx can install without generating a multilib version conflict.
+        def multilib_update_required(package_idx, other_idx)
+          return false if current_version_array[other_idx] && !new_version_array[other_idx]
+
+          other_available = available_version(other_idx)
+          package_installed = installed_version(package_idx)
+          package_available = available_version(package_idx)
+          return false if other_available.name != package_installed.name
+          return false if other_available.version != package_available.version
+          return false if other_available.version == package_installed.version
+
+          true
+        end
+
         # this will resolve things like `/usr/bin/perl` or virtual packages like `mysql` -- it will not work (well? at all?) with globs that match multiple packages
         def resolved_package_lock_names(names)
           names.each_with_index.map do |name, i|
--- a/lib/chef/provider/package/yum/yum_helper.py
+++ b/lib/chef/provider/package/yum/yum_helper.py
@@ -82,6 +82,28 @@ except NameError:
                 return True
         return False
 
+# Compares a set of candidate packages with the packages installed on a system
+# and returns the set of packages that would not generate a multilib version
+# conflict when installed.
+def remove_multilib_incompat(pkgs):
+    compat_pkgs = []
+    installed = {}
+
+    for pkg in pkgs:
+        if base.allowedMultipleInstalls(pkg):
+            compat_pkgs.append(pkg)
+            continue
+        installed.setdefault(pkg.name, base.rpmdb.searchNames([pkg.name]))
+        for inst in installed[pkg.name]:
+            if pkg.arch == inst.arch:
+                continue
+            if not pkg.verEQ(inst):
+                break
+        else:
+            compat_pkgs.append(pkg)
+
+    return compat_pkgs
+
 def query(command):
     base = get_base()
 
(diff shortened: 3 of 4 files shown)


karmix:
**This PR is not yet ready to merge.**  I am creating it show work done on #10485, and to facilitate discussion.

lib/chef/provider/package.rb
```diff
@@ -412,59 +412,60 @@ def target_version_array
...
+              target_version_array.push(target == current_version ? nil : target)
             end
-
             target_version_array
           end
       end
 
+      # Returns the version a package should have after the install/upgrade operation is complete.
+      #
+      # @return [String] target package version
```
jaymzh, line 424: now that you've separated this out the `nil` case(s) look like a bug. It's worth adding documentation here that this returns nil if there is nothing to do.

lib/chef/provider/package/yum.rb
```diff
@@ -153,6 +153,41 @@ def unlock_package(names, versions)
 
         private
 
+        def target_version(package_name, new_version, current_version, candidate_version)
+          return super if action != :install || current_version.nil? || new_version
```
jaymzh, line 157: couldn't you run into the case where you have a x86_64 pkg installed at X, but you're calling `:upgrade` on on the i386 pkg whose candidate version is Y (where Y > X), and thus need the logic below?

lib/chef/provider/package/yum/yum_helper.py
```diff
@@ -82,6 +82,28 @@ def any(s):
                 return True
         return False
 
+# Compares a set of candidate packages with the packages installed on a system
+# and returns the set of packages that would not generate a multilib version
+# conflict when installed.
+def remove_multilib_incompat(pkgs):
```
jaymzh, line 88: remove sounds like packages are gonna be uninstalled. Maybe reject_multilib_incompat ?

lib/chef/provider/package/yum.rb
karmix: I do not believe there is a legitimate way for that to occur.  In the scenario where Y is the latest candidate for an i386 package, there must also be an x86_64 candidate at version Y, otherwise the repository is incomplete and could never install version Y of the i386 package.  I believe performing a bare `yum upgrade` in a on a system with a repository in such a state would fail unless `--skip-broken` was used.

Of course, when the action is `:upgrade` with the x86_64 package at version X and candidate Y is in the repository, `super` will display a message about updating the x86_64 pacakge to version Y, and another message about installing or updating to version Y of the i386 package, as appropriate.

lib/chef/provider/package.rb
karmix: I am not sure I am understanding what you are referring to.  I (also?) thought it would be confusing to have `nil` as a target version, so the `target_version` method never returns `nil`.  The cases that do not immediately return a new version of the package to install/update fall through to the end of the function and return `current_version`.  Since `current_version.nil?` is handled for each value of `action`, `current_version` should never be `nil` when it is returned.

The `target_version_array` method translates a target version of `current_version` to `nil` when constructing its array, and the meaning of `nil` in that case is documented in the description of `target_version_array`.

lib/chef/provider/package/yum/yum_helper.py
karmix: Agreed.  I like reject better, too.  I'll change this with the next update to this PR.

lib/chef/provider/package.rb
jaymzh: oh I missed the final return here.

But that ... won't work. I think you break :upgrade here.

`nil` in the case of multipackage means 'latest' for :upgrade (ie do work if there's a newer version in the repo), or "anything" for :install (ie do nothing if anything is installed). For the :upgrade path, you'll likely need to return `nil` specifically.

lib/chef/provider/package/yum.rb
jaymzh: OK so, current state of system:
* foo.x86_64 installed at 1.2-1
* foo.i386 not installed
Recipe:
* `package 'foo.i386' { :upgrade }`
Repo:
* 1.2-1 and 1.3-1 are both available

I believe this is the case where you need the logic below and the return here will break things. I think.

jaymzh:
@karmix - you still planning on updating this?

lamont-granquist:
Things to think about (notes really for me in this issue post hashing this out for an hour):

If "foo.x86_64" and "foo.i686" have versions 1.0, 2.0 and 3.0 and "bar" depends on "foo.i686 < 3.0" and foo-1.0.i686" is installed then 3.0 is unsolvable, while advancing both to 2.0 is possible, and installing only "foo-1.0.x86_64" should hopefully always be possible.

I also want to validate what happens if we simply don't add the candidate version to the yum command if the new_resource.version is nil since that would presumably leave it up to yum install as to how it wants to do its depsolving.  Phil mentioned this may fail before something like rhel 7.3.

I am fairly worried that we may wind up going down the route of needing to write a depsolver in the package provider and want to avoid something like that.

The problem that we have right now at its most abstract level is that the candidate_version we are returning is just the latest one that we can install, we get no information on if that solves against the current set of installed RPMs.  Stated that way the right thing to do feels more like we need to ask the python API not for the latest version but for the latest version which can be installed against the set of RPMs.  That shouldn't require any surgery to the superclass (and would also make it work the way that Phil wants which is to not upgrade the existing installed version of the alternative RPM).

The problem with that solution however is if we're doing a multipackage upgrade of both arches then we need to let the to-be-installed RPMs float when we solve for each individual RPMs candidate_version.  So the python API would have to get a list of all the packages being installed/upgraded along with their requested version pins when asking for each candidate_version.   It may be better at that point to do one request to solve the whole solution.

The :upgrade case should also carefully be considered, we can't just solve the :install case.

I don't like extracting the whole target_version API into a method and then subclassing and calling super on the entire mess.  That should get inverted and the main loop should have a hook added which is normally empty (or false or nil or something suitable like that) which is then overridden by the subclass.  The subclass should never be overriding the main behavior there, that is creating an inheritence mess making it even more difficult to work on the superclass and opens up violating Lisikov Substitution.

And the definition of done needs to include tests, and they need to cover pretty much all those use cases.  So that means we need multiarch library mock packages added to the integration tests, we need to have install, upgrade and downgrade tests written which exercise the use cases where the version is either pinned or floating and the alternative package is installed at different versions.

And given how complicated all this is already, I'm not sure how much I honestly care about the "protected_multilib=0" case.  At a certain point I'm okay with people having to write an execute resource with an `rpm -q` guard.  Yeah, actually that's the correct answer there, its okay for the yum_package provider to not handle literally every single use case generated by arbitrarily complicated and officially discouraged configuration.

lamont-granquist:
Another bit of horribleness to consider is the case where the installed secondary arch version is on a version which is not available for the arch which is being requested.  So 'foo-1.0.i686' was installed on an image a long time ago, that is brought up and we want to install 'foo.x86_64' but the only available version is 2.0 and 1.0 of both library versions has been removed from mirrors.  This is the use case which leans me more towards the side where if 'foo.x86_64' is not installed, then it should be installed to the latest version, with upgrades of any other dependencies being allowed.

lamont-granquist:
Testing on RHEL8 I'm not fairly confused because it seems to work:

With libtevent-0.9.39-2.el8.x86_64 installed running:

```
yum install libtevent.i686
```

Results ultimately in the 32-bit version being installed to the latest version without the 64-bit version being upgraded, although yum upgrades the glibc in the process:

```
Upgraded:
  glibc-2.28-101.el8.x86_64               glibc-common-2.28-101.el8.x86_64               glibc-langpack-en-2.28-101.el8.x86_64               libtalloc-2.2.0-7.el8.x86_64

Installed:
  glibc-2.28-101.el8.i686                    libtalloc-2.2.0-7.el8.i686                    libtevent-0.10.0-2.el8.i686                    libxcrypt-4.1.1-4.el8.i686

Complete!
```

```
# rpm -qa | egrep libtevent
libtevent-0.9.39-2.el8.x86_64
libtevent-0.10.0-2.el8.i686
```

lamont-granquist:
The RHEL8 DNF docs on the yum-to-dnf-compatibility chart list protected_multilib as "removed"

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/considerations_in_adopting_rhel_8/software-management_considerations-in-adopting-rhel-8

Screen Shot 2021-02-03 at 9 05 06 PM

That suggests to me that while YUM docs may discourage that setting that it seems DNF has abandoned that viewpoint.

This now seems a lot to me like the `allow_downgrade` problem as well and there's an easy solution which is to set `protected_multilib=0`

Amusingly enough I find this problem referenced on this "The tar pit of Red Hat overcomplexity" page:

http://www.softpanorama.org/Commercial_linuxes/RHEL/index.shtml#Pretty_convoluted_RPM_packaging_system_which_creates_problems

karmix:
Oh, that makes more sense.  RHEL 8 uses `dnf`, and it looks like `dnf` dropped this multilib mess long ago (https://github.com/rpm-software-management/dnf/commit/f7dd24e10fdc6d1edc7b0924af0201dfb114223b).  This issue only exists on systems that still use `yum`.  `dnf` packages are handled by the `dnf` provider, which has its own helper.  I don't believe these changes should affect the behavior of the `dnf` provider in any way.

lib/chef/provider/package/yum.rb
karmix: If by "break things" you mean fall back to existing behavior, that is correct.  These modifications only change the behavior of the `:install` action, not the `:upgrade` action.  That is because when you use `:install` and don't specify a version, you are telling `chef` that you only care that the package gets installed, in which case it would be acceptable for `chef` to choose whichever version it determines is most appropriate.  When you use the `:upgrade` action, you are telling `chef` to ensure the package is at the latest version.  It seems more appropriate to have `chef` continue to generate an error when `:upgrade` is used for `foo.i386` and version 1.3-1 is available than to install an out of date version of `foo.i386`.

The alternative would be to try to have `chef` implicitly upgrade `foo.x86_64` every time there is an update for `foo.i386` so that it could complete the operation successfully.  However, in this case, `chef` would be modifying `foo.x86_64`, which it was never told to manage.  I would not expect this behavior and would view it as a side effect.  I believe it would be better for the recipe to be more explicit.  Specifically,
```ruby
package %i[foo.i686 foo.x86_64] do
    action :upgrade
end
```
behaves as expected without this patch, whereas `action :install` does not.

It is also worth noting that the provider only knows about the packages it was told to mange.  In your scenario, the provider does not know whether `foo.x86_64` is present.  Even though the helper script has access to this information, the provider doesn't have a way to ask for it.  We would have to make significant changes to the provider  and helper to allow the provider to implicitly identify and manage packages it was not instructed to manage.  Given that I could find no tests for this provider and I rarely see the `:upgrade` action used in the wild, I think refactoring the code to add such a feature could introduce more risk than the benefits merit.

lib/chef/provider/package.rb
karmix: I do not understand to what you are referring.  `nil` in the `target_version_array` means "the package is not being modified" for both the `:upgrade` and `:install` cases (per line 407).  In all of the cases where `nil` was being pushed onto the stack in the `target_version_array` method, the `target_version` method returns `current_version`, meaning the package is expected to be at the same version as is currently installed.  The `target_version_array` method now simply retrieves the `target_version` for each package, and maps `current_version` responses back to the `nil`s it uses to represent a no-op on the package in its array.

Are we perhaps talking about the use of `nil` in different parts of the code?

lamont-granquist:
Yeah, chef handling every single edge case to provide a yummy perfect abstraction over all the package providers sounds good in principle, but in practice that is terrible software design.  The `allow_downgrade` mess is a testament to that.  The code here is going to have to be maintained and adds complexity to something which is already nuclear-reactor-grade complicated.  The solution of `protected_multilib=0` is vastly simpler and makes that exception just go away.  I don't normally like to tell people how to run their infrastructure, but this is one time where I think its more than worth it.  It looks like RedHat already jettisoned that concept into the Sun, so that's the best solution for everyone.

Issue #10457, opened 2020-09-22
ifconfig resource can create lots of blank lines for an interface config file


haidangwa:
### Describe the Enhancement:

Improve the template defined by https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig/redhat.rb#L29-L47 and used to generate `/etc/sysconfig/network-scripts/ifcfg-*` files so it isn't loaded with blank lines when the if statements return a false value.

### Repro
Define this resource in a chef recipe:
```
ifconfig 'my-interface' do
  device 'my-interface'
  mask '255.255.255.255'
  bootproto 'none'
  onboot 'no'
end
```

### Describe the Need:

We use the ifconfig resource and many of the options referenced by https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig/redhat.rb#L29-L47 return false. This results in a file that has a lot of blank lines. It would be preferrable to have a compact file with little or no empty lines.

I setup an ifconfig resource configures a dummy interface successfully, then I look in the file to verify the contents I noticed a lot of empty lines.

```
# cat /etc/sysconfig/network-scripts/ifcfg-my-interface

DEVICE=my-interface

BOOTPROTO=none
IPADDR=192.168.20.18
NETMASK=255.255.255.255












          
```
Everything you see above after the `NETMASK` line was added by the ifconfig resource.

### Current Alternative

none that I know of

jmherbst:
Would this just entail using the `'-'` trim mode on the ERB object https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig.rb#L239
```
template = ::ERB.new(@config_template, nil, '-')
```

Then refactoring the `config_template`s to leverage that trim mode? 
Example: 
```
<% if new_resource.device -%>
DEVICE=<%= new_resource.device %>
<% end -%>
<% if new_resource.onboot == "yes" -%>
ONBOOT=<%= new_resource.onboot %>
<% end -%>
```

I'd be happy to make this change, if desired.

Thanks

phiggins:
Hi @jmherbst,

That is my impression as well, that using ERB's trim mode would fix the problem. If you'd like to submit a patch for that it would be appreciated.

Alternatively, using ERB for this seems a bit overkill so it might make sense to refactor the code to generate a string in a simpler fashion.

jmherbst:
I believe this can be closed out now. Thanks for the suggestion @haidangwa and for the help @phiggins 👍

PR #10489, opened 2020-10-06
Changing ifconfig provider to reduce blank lines in redhat and debian ifconfigs

jmherbst:
Changing ifconfig provider to use '-' trim_mode for its ERB templates and updating the redhat and debian ifconfigs to leverage the trim_mode so they are not constructed with any blank lines when nil values are encountered.

This addresses https://github.com/chef/chef/issues/10457

## Description
This improves the `ifconfig` templates defined for debian and redhat so that `nil` values do not generate blank lines in the resulting ifconfig file.

Ruby ERB supports a `trim_mode` parameter of `'-'` to `"-  omit blank lines ending in -%>"`

I've changed the ERB call in https://github.com/chef/chef/blob/master/lib/chef/provider/ifconfig.rb#L239 to leverage this trim mode.

I've also changed each of the distro's template generation scripts to generate a template that omits blank lines when a `nil` value is encountered 

## Related Issue
https://github.com/chef/chef/issues/10457

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/ifconfig/debian.rb
+++ b/lib/chef/provider/ifconfig/debian.rb
@@ -32,25 +32,43 @@ class Chef
         def initialize(new_resource, run_context)
           super(new_resource, run_context)
           @config_template = %{
-<% if new_resource.device %>
-<% if new_resource.onboot == "yes" %>auto <%= new_resource.device %><% end %>
+<% if new_resource.device -%>
+<% if new_resource.onboot == "yes" -%>
+auto <%= new_resource.device %>
+<% end -%>
 <% case new_resource.bootproto
-   when "dhcp" %>
+   when "dhcp" -%>
 iface <%= new_resource.device %> <%= new_resource.family %> dhcp
-<% when "bootp" %>
+<% when "bootp" -%>
 iface <%= new_resource.device %> <%= new_resource.family %> bootp
-<% else %>
+<% else -%>
 iface <%= new_resource.device %> <%= new_resource.family %> static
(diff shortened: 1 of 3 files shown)


lib/chef/provider/ifconfig/debian.rb
```diff
@@ -32,25 +32,43 @@ class Debian < Chef::Provider::Ifconfig
         def initialize(new_resource, run_context)
           super(new_resource, run_context)
           @config_template = %{
-<% if new_resource.device %>
-<% if new_resource.onboot == "yes" %>auto <%= new_resource.device %><% end %>
+<% if new_resource.device -%>
```
phiggins, line 35: If the whole thing depends on `new_resource.device`, that conditional could be moved out of the ERB string and just set `@config_template = ""` or something.

phiggins:
The code looks fine, thanks for the fix! Would it be possible to add some tests for this?

jmherbst:
What are you thinking of In addition to the existing unit tests doing string matching ([redhat](https://github.com/chef/chef/blob/master/spec/unit/provider/ifconfig/redhat_spec.rb#L57), [debian](https://github.com/chef/chef/blob/master/spec/unit/provider/ifconfig/debian_spec.rb#L70-L75)) @phiggins ?

phiggins:
Oops, I was thinking that any existent tests would have failed with these changes, but I guess I was wrong. This seems good to me 👍

phiggins (approved):

PR #10494, opened 2020-10-09
Refactor ResourceGuardInterpreter

phiggins:

I started refactoring ResourceGuardInterpreter as part of a different PR a few months ago but moved the changes onto a different branch when I realized I didn't have to change it for that PR's behavior. This doesn't fix anything other than making the code a lot easier to follow (IMO).

### DIFF
--- a/lib/chef/guard_interpreter/resource_guard_interpreter.rb
+++ b/lib/chef/guard_interpreter/resource_guard_interpreter.rb
@@ -20,18 +20,29 @@ require_relative "../guard_interpreter"
 
 class Chef
   class GuardInterpreter
-    class ResourceGuardInterpreter < DefaultGuardInterpreter
-
+    class ResourceGuardInterpreter
       def initialize(parent_resource, command, opts)
-        super(command, opts)
+        @command = command
+        @opts = opts
+
         @parent_resource = parent_resource
         @resource = get_interpreter_resource(parent_resource)
+      end
+
+      # This class used to inherit from DefaultGuardInterpreter and it responds
+      # to #output, so leave this in for potential backwards compatibility.
+      def output
+        nil
       end
 
       def evaluate
         # Add attributes inherited from the parent class
         # to the resource
         merge_inherited_attributes
+
+        @opts.each do |attribute, value|
+          @resource.send(attribute, value)
+        end
 
         # Only execute and script resources and use guard attributes.
         # The command to be executed on them are passed via different attributes.
@@ -42,9 +53,9 @@ class Chef
         # We need to make sure we check for Script first because any resource
         # that can get to here is an Execute resource.
         if @resource.is_a? Chef::Resource::Script
-          block_attributes = @command_opts.merge({ code: @command })
+          @resource.code @command
         else
-          block_attributes = @command_opts.merge({ command: @command })
+          @resource.command @command
         end
 
         # Handles cases like powershell_script where default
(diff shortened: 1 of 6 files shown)


lib/chef/guard_interpreter/resource_guard_interpreter.rb
```diff
@@ -20,19 +20,28 @@
 
 class Chef
   class GuardInterpreter
-    class ResourceGuardInterpreter < DefaultGuardInterpreter
```
phiggins, line 23: No behavior was inherited other than `initialize` which made this more convoluted.

lib/chef/guard_interpreter/resource_guard_interpreter.rb
```diff
@@ -53,33 +62,24 @@ def evaluate
...
         if @resource.class.respond_to?(:get_default_attributes)
-          block_attributes = @resource.class.send(:get_default_attributes, @command_opts).merge(block_attributes)
+          @resource.class.send(:get_default_attributes).each do |attribute, value|
+            @resource.send(attribute, value)
+          end
         end
 
-        resource_block = block_from_attributes(block_attributes)
-        evaluate_action(nil, &resource_block)
-      end
```
phiggins, line 61: `evaluate` is the entry point for all the behavior in this class, and it:
* builds a hash of parameters to be set on the guard resource
* makes a proc that sets the parameters on the guard resource
* passes the proc to another function that `instance_eval`s the proc on the guard resource

Instead it can set the parameters on the guard resource directly and avoid all the indirection.

lib/chef/guard_interpreter/resource_guard_interpreter.rb
```diff
@@ -106,30 +106,17 @@ def get_interpreter_resource(parent_resource)
...
         end
 
-        if inherited_attributes && !inherited_attributes.empty?
-          inherited_attributes.each do |attribute|
-            if @parent_resource.respond_to?(attribute) && @resource.respond_to?(attribute)
-              parent_value = @parent_resource.send(attribute)
-              child_value = @resource.send(attribute)
-              if parent_value || child_value
-                @resource.send(attribute, parent_value)
-              end
```
phiggins, line 131: This code was potentially hiding a bug, because `nil` and `false` are treated different in resource properties but are considered the same here.

spec/unit/resource/powershell_script_spec.rb
```diff
@@ -47,81 +47,11 @@
...
-    it "passes convert_boolean_return as false if it was specified as true in a guard parameter" do
-      other_guard_parameters = { cwd: "/etc/chef", architecture: :x86_64 }
-      parameters_with_boolean_disabled = other_guard_parameters.merge({ convert_boolean_return: false, code: "$true" })
-      allow_any_instance_of(Chef::GuardInterpreter::ResourceGuardInterpreter).to receive(:block_from_attributes).with(
-        parameters_with_boolean_disabled
-      ).and_return(Proc.new {})
-      resource.only_if("$true", parameters_with_boolean_disabled)
-    end
+    expect(inherited_difference).to eq([])
   end
```
phiggins, line 55: I tried to figure out a way to make these tests meaningful for a while and it didn't seem possible without being very tied to the implementation of `Conditional`.

These tests currently do not set any expectations so there is no way they could fail. If they did set expectations on how `ResourceGuardInterpreter` worked, they would be failing now because calling `only_if` does not evaluate any `Conditional`s, it just gets them ready for later evaluation.

tas50:
@phiggins can you rebase this and fix that chefstyle issue?

phiggins:
@tas50 Good call, I didn't even notice. Done 👍

PR #10504, opened 2020-10-15
Make sure to thank tecRacer

tas50:
I blanked on this one when writing the release notes because Lamont
wrote the final implementation, but tecracer put a huge amount of
effort into this functionality and we should call them out for that.


phiggins (approved):
Big 👍

PR #10519, opened 2020-10-16
Use action not the alias of allowed_actions in our resources

tas50:
Use the actual method not the aliases method name.

Also fixup weird formatting in an error message


tas50:
Yeah nm there's more layers to this mess

PR #10520, opened 2020-10-16
Move the alias for attribute to property right into the property mixin

tas50:
Stick it right next to the property def so it's pretty obvious how this works.


tas50:
Merged via Zoom approval

PR #10524, opened 2020-10-16
WIP: Avoid making WMI calls to find the windows version on each symlink creation

tas50:
We can gather this information via ChefUtils which abstracts the Ohai data. That way we only make these WMI calls once.


mwrock (approved):

mwrock:
Looks like there are some unhappy unit tests.

tas50:
Sadly this won't work because we use this all over cheffs and that won't have node data.

PR #10525, opened 2020-10-16
Add required_ruby_version to chef-utils and chef-config

tas50:
These are getting pulled in all over and that's causing problems on
older Ruby releases.


mwrock (approved):

lamont-granquist (approved):

lamont-granquist:
i thought we equality locked this in the chef gem, but yeah this is obviously correct

PR #10530, opened 2020-10-19
Test ifconfig in Test Kitchen and add examples to the resource

tas50:
We lacked an actual test of this resource. Let's see if it works.

### DIFF
--- a/lib/chef/resource/ifconfig.rb
+++ b/lib/chef/resource/ifconfig.rb
@@ -26,7 +26,7 @@ class Chef
 
       provides :ifconfig
 
-      description "Use the **ifconfig** resource to manage interfaces on Unix and Linux systems."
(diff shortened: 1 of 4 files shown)


lib/chef/resource/ifconfig.rb
```diff
@@ -21,16 +21,67 @@
...
+      ```
+
+      will create the following interface configuration:
+
+      ```
+      iface eth1 inet static
+        address 33.33.33.80
+      ```
+
+      **Specify a boot protocol**
```
phiggins, line 46: Is this example missing `bootproto`?

tas50: That whole example seems bogus. I just nuked it.

lamont-granquist (approved):

PR #10531, opened 2020-10-19
Remove custom constant mocking in favor of rspec's.

phiggins:

Followup to #10522, I searched for uses of `constantize` and was surprised to find this in the test suite.

* Use rspec's constant mocking instead of dynamically setting constants and globally modifying ENV.
* Avoid defining methods in the global namespace when they are only used inside rspec tests.


tas50:
Nice

PR #10545, opened 2020-10-21
Improve Windows resource performance by converting powershell_out usage to powershell_exec

mwrock:
`powershell_package` and `windows_feature_powershell` are still using powershell_out due to timeout.

I've tested every affected resource and added functional tests where feasible. Interestingly a couple resources were busted and I don't think the `windows_dfs_server` could have ever possibly worked at all.

### DIFF
--- a/lib/chef/powershell.rb
+++ b/lib/chef/powershell.rb
@@ -34,8 +34,6 @@ class Chef
     # @param script [String] script to run
     # @return [Object] output
     def initialize(script)
-      raise "Chef::PowerShell can only be used on the Windows platform." unless RUBY_PLATFORM.match?(/mswin|mingw32|windows/)
-
       @dll ||= "Chef.PowerShell.Wrapper.dll"
       exec(script)
     end
--- a/lib/chef/powershell.rb
+++ b/lib/chef/powershell.rb
@@ -63,7 +63,15 @@ class Chef
     def exec(script)
       FFI.ffi_lib @dll
       FFI.attach_function :execute_powershell, :ExecuteScript, [:string], :pointer
-      execution = FFI.execute_powershell(script).read_utf16string
+      # This is a temporary fix for running in a Habitat environment
+      # In habitat we set CHEF_POWERSHELL_BIN so that .Net resolves our
+      # managed shim assembly from the correct location.
+      # It seems that that is preventing .Net from successfully loading GAC assemblies
+      # and can break all sorts of edge (and not so edge) scenarios. Once we are actually
+      # inside the powershell run space, we know our shim was loaded and can unset
+      # CHEF_POWERSHELL_BIN which will bypass our custom resolver logic. The real fix is
+      # to fix our resolver. Oh and OH MY GOD this was a pain to track down.
+      execution = FFI.execute_powershell("$ENV:CHEF_POWERSHELL_BIN=$NULL;#{script}").read_utf16string
       hashed_outcome = Chef::JSONCompat.parse(execution)
       @result = Chef::JSONCompat.parse(hashed_outcome["result"])
       @errors = hashed_outcome["errors"]
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -151,7 +151,7 @@ class Chef
           @choco_exe ||= begin
               # if this check is in #define_resource_requirements, it won't get
               # run before choco.exe gets called from #load_current_resource.
-              exe_path = ::File.join(choco_install_path.to_s, "bin", "choco.exe")
+              exe_path = ::File.join(choco_install_path, "bin", "choco.exe")
               raise Chef::Exceptions::MissingLibrary, CHOCO_MISSING_MSG unless ::File.exist?(exe_path)
 
               exe_path
@@ -160,9 +160,9 @@ class Chef
 
         # lets us mock out an incorrect value for testing.
         def choco_install_path
-          @choco_install_path ||= powershell_out!(
-            PATHFINDING_POWERSHELL_COMMAND
-          ).stdout.chomp
+          result = powershell_exec!(PATHFINDING_POWERSHELL_COMMAND).result
+          result = "" if result.empty?
+          result
         end
 
         # Helper to dispatch a choco command through shell_out using the timeout
--- a/lib/chef/resource/powershell_package_source.rb
+++ b/lib/chef/resource/powershell_package_source.rb
@@ -16,7 +16,6 @@
 #
 
 require_relative "../resource"
-require_relative "../json_compat"
 
 class Chef
   class Resource
@@ -57,11 +56,11 @@ class Chef
 
       load_current_value do
         cmd = load_resource_state_script(source_name)
-        repo = powershell_out!(cmd)
-        if repo.stdout.empty?
+        repo = powershell_exec!(cmd)
+        if repo.result.empty?
           current_value_does_not_exist!
         else
-          status = Chef::JSONCompat.from_json(repo.stdout)
+          status = repo.result
         end
         url status["url"]
         trusted status["trusted"]
@@ -78,28 +77,28 @@ class Chef
           if package_source_exists?
             converge_if_changed :url, :trusted, :publish_location, :script_source_location, :script_publish_location do
               update_cmd = build_ps_repository_command("Set", new_resource)
-              res = powershell_out(update_cmd)
-              raise "Failed to update #{new_resource.source_name}: #{res.stderr}" unless res.stderr.empty?
+              res = powershell_exec(update_cmd)
+              raise "Failed to update #{new_resource.source_name}: #{res.errors}" if res.error?
             end
           else
             converge_by("register source: #{new_resource.source_name}") do
               register_cmd = build_ps_repository_command("Register", new_resource)
-              res = powershell_out(register_cmd)
-              raise "Failed to register #{new_resource.source_name}: #{res.stderr}" unless res.stderr.empty?
+              res = powershell_exec(register_cmd)
+              raise "Failed to register #{new_resource.source_name}: #{res.errors}" if res.error?
             end
           end
         else
           if package_source_exists?
             converge_if_changed :url, :trusted, :provider_name do
               update_cmd = build_package_source_command("Set", new_resource)
-              res = powershell_out(update_cmd)
-              raise "Failed to update #{new_resource.source_name}: #{res.stderr}" unless res.stderr.empty?
+              res = powershell_exec(update_cmd)
+              raise "Failed to update #{new_resource.source_name}: #{res.errors}" if res.error?
             end
           else
             converge_by("register source: #{new_resource.source_name}") do
               register_cmd = build_package_source_command("Register", new_resource)
-              res = powershell_out(register_cmd)
-              raise "Failed to register #{new_resource.source_name}: #{res.stderr}" unless res.stderr.empty?
+              res = powershell_exec(register_cmd)
+              raise "Failed to register #{new_resource.source_name}: #{res.errors}" if res.error?
             end
           end
         end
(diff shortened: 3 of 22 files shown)


lib/chef/resource/windows_certificate.rb
```diff
@@ -207,16 +207,16 @@ def export_cert(cert_obj, cert_path)
           when ".der"
             out_file.puts(cert_obj.to_der)
           when ".cer"
-            cert_out = powershell_out("openssl x509 -text -inform DER -in #{cert_obj.to_pem} -outform CER").stdout
+            cert_out = powershell_exec("openssl x509 -text -inform DER -in #{cert_obj.to_pem} -outform CER").result
```
tas50, line 210: What's the perf for powershell_exec vs. standard shellout because this seems like PS isn't needed

lib/chef/resource/windows_dfs_server.rb
```diff
@@ -49,14 +49,14 @@ class WindowsDfsServer < Chef::Resource
         default: 3600
 
       load_current_value do
-        ps_results = powershell_out("Get-DfsnServerConfiguration -ComputerName '#{ENV["COMPUTERNAME"]}' | Select LdapTimeoutSec, PreferLogonDC, EnableSiteCostedReferrals, SyncIntervalSec, UseFqdn | ConvertTo-Json")
+        ps_results = powershell_exec("Get-DfsnServerConfiguration -ComputerName '#{ENV["COMPUTERNAME"]}' | Select LdapTimeoutSec, PreferLogonDC, EnableSiteCostedReferrals, SyncIntervalSec, UseFqdn | ConvertTo-Json")
```
tas50, line 52: ConvertTo-Json can go away with powershell_exec

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -40,8 +40,13 @@ def cleanup(store)
 
   def no_of_certificates
     path = "Cert:\\LocalMachine\\" + store
+    # Seems weird that we have to call sir twice right?
```
phiggins, line 43:
```suggestion
    # Seems weird that we have to call dir twice right?
```

spec/support/platform_helpers.rb
```diff
@@ -243,10 +243,10 @@ def ifconfig?
...
 def choco_installed?
   result = ShellHelpers.powershell_exec("choco --version")
-  result.error?
+  !result.error?
 end
 
 def pwsh_installed?
   result = ShellHelpers.powershell_exec("pwsh.exe --version")
-  result.error?
+  !result.error?
```
phiggins: This seems like kind of a big deal, are these just broken right now?

mwrock: yeah. this is all a bit rough right now. and ^^^ broke alot of stuff :)

lib/chef/resource/windows_certificate.rb
mwrock: we should use shell_out when calling narive commands because shell_out will make sure the path is properly set it is also faster since there is some overhead with powershell_exec. Locally I compared and there was ~300ms difference.

lib/chef/powershell.rb
```diff
@@ -34,8 +34,6 @@ class PowerShell
     # @param script [String] script to run
     # @return [Object] output
     def initialize(script)
-      raise "Chef::PowerShell can only be used on the Windows platform." unless RUBY_PLATFORM.match?(/mswin|mingw32|windows/)
```
tas50, line 37: Any particular reason for removing this one?

lib/chef/provider/package/chocolatey.rb
```diff
@@ -160,9 +160,9 @@ def choco_exe
 
         # lets us mock out an incorrect value for testing.
         def choco_install_path
-          @choco_install_path ||= powershell_out!(
+          @choco_install_path ||= powershell_exec!(
```
tas50, line 163: Isn't this going to return a JSON object now instead of a string?

tas50, line 163: Also it looks like there's no need to memoize this since it just goes into the `choco_exe` variable which is memoized

lib/chef/resource/powershell_package_source.rb
```diff
@@ -78,28 +77,28 @@ class PowershellPackageSource < Chef::Resource
           if package_source_exists?
             converge_if_changed :url, :trusted, :publish_location, :script_source_location, :script_publish_location do
               update_cmd = build_ps_repository_command("Set", new_resource)
-              res = powershell_out(update_cmd)
-              raise "Failed to update #{new_resource.source_name}: #{res.stderr}" unless res.stderr.empty?
+              res = powershell_exec(update_cmd)
+              raise "Failed to update #{new_resource.source_name}: #{res.errors}" if res.error?
```
tas50, line 81: Do we need to join the errors in these?

lib/chef/resource/windows_firewall_profile.rb
```diff
@@ -166,12 +166,7 @@ def firewall_enabled?(profile_name)
                 return $true
             } else {return $false}
           CODE
-          firewall_status = powershell_out(cmd).stdout
-          if /True/.match?(firewall_status)
-            true
-          elsif /False/.match?(firewall_status)
-            false
-          end
+          powershell_exec!(cmd).result
```
tas50, line 169: Does the boolean parsing just go away with our better JSON parsing from powershell_exec?

mwrock: yes. They are parsed as true booleans.

cspell.json
```diff
@@ -236,6 +236,7 @@
     "cname",
     "codepage",
     "coderanger",
+    "coderay",
```
tas50, line 239: This can go away now that we have the same thing in the global dictionary

lib/chef/resource/windows_firewall_profile.rb
tas50: 😄

lib/chef/resource/windows_share.rb
```diff
@@ -246,14 +244,16 @@ def delete_share
           delete_command = "Remove-SmbShare -Name '#{new_resource.share_name}' -Force"
 
           Chef::Log.debug("Running '#{delete_command}' to remove the share")
-          powershell_out!(delete_command)
+          powershell_exec!(delete_command)
         end
 
         def update_share
-          update_command = "Set-SmbShare -Name '#{new_resource.share_name}' -Description '#{new_resource.description}' -Force"
+          update_command = "Set-SmbShare -Name '#{new_resource.share_name}' -Description '#{new_resource.description}'  -ConcurrentUserLimit #{new_resource.concurrent_user_limit} -CATimeout #{new_resource.ca_timeout} -EncryptData:#{bool_string(new_resource.encrypt_data)} -ContinuouslyAvailable:#{bool_string(new_resource.continuously_available)} -Force"
```
tas50, line 251: Now I'm very confused why we didn't do a full update before

lib/chef/provider/package/chocolatey.rb
mwrock: In cases where the powershell is emitting just a string, it gets parsed as a plain string.

tas50: Very fancy. Thanks for the clarification.

mwrock: However, looking this over more closely. It could blow up if choco is not installed and the command returns an empty string. In that case the `result` is `{}` which has no `chomp`. So I'm gonna have it check `empty?`.

lib/chef/resource/powershell_package_source.rb
```diff
@@ -110,16 +109,16 @@ class PowershellPackageSource < Chef::Resource
...
           end
         end
       end
 
       action_class do
         def package_source_exists?
-          cmd = powershell_out!("(Get-PackageSource -Name '#{new_resource.source_name}' -WarningAction SilentlyContinue).Name")
-          cmd.stdout.downcase.strip == new_resource.source_name.downcase
+          cmd = powershell_exec!("(Get-PackageSource -Name '#{new_resource.source_name}' -ErrorAction SilentlyContinue).Name")
+          !cmd.result.empty? && cmd.result.to_s.downcase.strip == new_resource.source_name.downcase
```
phiggins, line 121:
```suggestion
          String(cmd.result).downcase.strip == new_resource.source_name.downcase
```

lib/chef/resource/windows_ad_join.rb
```diff
@@ -182,10 +182,10 @@ def node_domain
         #   workgroup.
         #
         def node_workgroup
-          node_workgroup = powershell_out!("(Get-WmiObject Win32_ComputerSystem).Workgroup")
+          node_workgroup = powershell_exec!("(Get-WmiObject Win32_ComputerSystem).Workgroup")
           raise "Failed to check if the system is currently a member of a workgroup" if node_workgroup.error?
 
-          node_workgroup.stdout.downcase.strip
+          node_workgroup.result
```
phiggins, line 188: Any reason to remove the `.downcase.strip` here and not in `node_domain`?

cspell.json
tas50, line 239:
```suggestion
```

lib/chef/resource/windows_ad_join.rb
mwrock: yeah. workgroup is only ever checked for `empty?` and if one is domain joined, it will return `{}` and break on `downcase.strip`.

lib/chef/resource/powershell_package_source.rb
mwrock: the problem here is that `String(cmd.result)` will return `"{}"` if it is empty.

lib/chef/resource/windows_workgroup.rb
```diff
@@ -117,10 +117,10 @@ def join_command
 
         # @return [Boolean] is the node a member of the workgroup specified in the resource
         def workgroup_member?
-          node_workgroup = powershell_out!("(Get-WmiObject -Class Win32_ComputerSystem).Workgroup")
+          node_workgroup = powershell_exec!("(Get-WmiObject -Class Win32_ComputerSystem).Workgroup")
           raise "Failed to determine if system already a member of workgroup #{new_resource.workgroup_name}" if node_workgroup.error?
 
-          node_workgroup.stdout.downcase.strip == new_resource.workgroup_name.downcase
+          !node_workgroup.result.empty? && node_workgroup.result.downcase.strip == new_resource.workgroup_name.downcase
```
phiggins, line 123:
```suggestion
          String(node_workgroup.result).downcase.strip == new_resource.workgroup_name.downcase
```

lib/chef/resource/powershell_package_source.rb
mwrock: I don't think so. The vast majority will have only one error. In the cases where there are multiple errors, I think having them comma separated in brackets makes it clear to view.

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -16,34 +16,39 @@
...
       Remove-Item -Path #{path} -Recurse
     EOC
-    powershell_out(command)
+    powershell_exec(command)
   end
 
   def no_of_certificates
     path = "Cert:\\LocalMachine\\" + store
+    # Seems weird that we have to call dir twice right?
+    # The powershell cert provider seems to cache the last dir
```
phiggins, line 44: I was a bit confused by this comment, I thought this was referring to an infra client provider not realizing "Certifcate Provider" is a Windows thing. https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.security/about/about_certificate_provider?view=powershell-7

mwrock: hoho. yeah I can see that would be confusing. I'll clarify thanks!

tas50 (approved):
Very exciting stuff here for windows perf

Issue #10499, opened 2020-10-13
Warnings from our net/http monkeypatch


tas50:
We need to clean this monkeypatch up:

```
/usr/local/Cellar/ruby/2.7.2/lib/ruby/2.7.0/net/http/exceptions.rb:25: warning: already initialized constant Net::HTTPClientException
/Users/tsmith/dev/work/chef/lib/chef/monkey_patches/net_http.rb:16: warning: previous definition of HTTPClientException was here
```

tas50:
This may require Ruby 2.7.2 to repro

tas50:
Fixed!

PR #10548, opened 2020-10-26
Remove unused monkeypatch on net/http.

phiggins:

#10499


lamont-granquist:
those respond_to? checks in chef/http should probably go as well since those are going to be real confusing now.

phiggins:
> those respond_to? checks in chef/http should probably go as well since those are going to be real confusing now.

Good catch, I had removed those also but I think I undid it accidentally.

PR #10555, opened 2020-10-26
ifconfig is not compatible with Fedora 33 or later

tas50:
This resource only works with the legacy redhat /etc/networks configuration format. That format was slapped onto NetworkManager when RHEL moved off legacy init scripts. Fedora 33 switches to the keyfile format so this resource no longer works on those systems. Stop testing it there and add a note to the resource description so we get that up on the docs site.

### DIFF
--- a/lib/chef/resource/ifconfig.rb
+++ b/lib/chef/resource/ifconfig.rb
@@ -26,7 +26,7 @@ class Chef
 
       provides :ifconfig
 
-      description "Use the **ifconfig** resource to manage interfaces on Unix and Linux systems. Note: This resource requires the ifconfig binary to be present on the system and may require additional packages to be installed first. On Ubuntu 18.04 or later you will need to install the `ifupdown` package, which disables the built in Netplan functionality. This resource will not work with Fedora release 33 or later."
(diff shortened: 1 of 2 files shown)


lib/chef/resource/ifconfig.rb
```diff
@@ -26,7 +26,7 @@ class Ifconfig < Chef::Resource
 
       provides :ifconfig
 
-      description "Use the **ifconfig** resource to manage interfaces on Unix and Linux systems. Note: This resource requires the ifconfig binary to be present on the system and may require additional packages to be installed first. On Ubuntu 18.04 or later you will need to install the `ifupdown` package, which disables the built in Netplan functionality."
+      description "Use the **ifconfig** resource to manage interfaces on Unix and Linux systems. Note: This resource requires the ifconfig binary to be present on the system and may require additional packages to be installed first. On Ubuntu 18.04 or later you will need to install the `ifupdown` package, which disables the built in Netplan functionality. This resource will not work with Fedora release 33 or later."
```
phiggins, line 29:
```suggestion
      description "Use the **ifconfig** resource to manage interfaces on Unix and Linux systems. Note: This resource requires the ifconfig binary to be present on the system and may require additional packages to be installed first. On Ubuntu 18.04 or later you will need to install the `ifupdown` package, which disables the built in Netplan functionality. Warning: This resource will not work with Fedora release 33 or later."
```

Give this its own callout box instead of being added onto the previous one.

phiggins (approved):

PR #10557, opened 2020-10-27
Final batch of unified_mode providers

lamont-granquist:
Also converts some of them to custom resources.

There isn't a ton of cleanup here because everything is a bit brittle.  A couple of resources were changed from declare_resource.  The `key_exists?` stuff in windows_env was a bit of cleanup that I couldn't resist, but I was punished for trying that with some crazy brittle spec tests.  The rest of the code that has been moved to the resources is pretty faithful to what was removed from the providers warts and all.

There was some deeply crazy stuff in here where we have a `module ::Win32` and a `class ::Chef::Win32` which becomes ambiguous depending on where you are in the class and inheritance search tree and so just needed to be explicit around `include ::Win32` but then that seemed to break some of the references to `TaskScehduler` constants which I never resolved as to why but just changing those to all be `::Win32::TaskScheduler` and being explicit fixed them.


lib/chef/resource/windows_task.rb
```diff
@@ -16,12 +16,18 @@
 # limitations under the License.
 #
 
+require "chef-utils" unless defined?(ChefUtils::CANARY)
 require_relative "../resource"
-require_relative "../win32/security" if Chef::Platform.windows?
+require_relative "../win32/security" if ChefUtils.windows_ruby?
+autoload :ISO8601, "iso8601" if ChefUtils.windows_ruby?
```
tas50, line 22: do we actually need to gate the autoload?

lib/chef/resource/windows_task.rb
```diff
@@ -464,6 +461,609 @@ def sec_to_dur(seconds)
       def sec_to_min(seconds)
         seconds.to_i / 60
       end
+
+      action_class do
+        if ChefUtils.windows_ruby?
+          include ::Win32
```
tas50, line 467: ?

lamont-granquist: that's from the original code

tas50: yeah I see the comment above. That's some serious wtf there

PR #10558, opened 2020-10-27
Merge repetitive conditionals

tas50:
So much easier to read and probably faster this way too.


mwrock (approved):

phiggins:
> So much easier to read and probably faster this way too.

Definitely easier to read, but I'd be surprised if creating an array and calling a method on it is faster than the bare conditional.

phiggins (approved):

tas50:
It is actually quite a bit slower per the benchmarks.

PR #10560, opened 2020-10-27
Simplify a weird conditional in chef-config

tas50:
if with ! is a brain twister.

### DIFF
--- a/chef-config/lib/chef-config/mixin/fuzzy_hostname_matcher.rb
+++ b/chef-config/lib/chef-config/mixin/fuzzy_hostname_matcher.rb
@@ -21,7 +21,7 @@ module ChefConfig
     module FuzzyHostnameMatcher
 
       def fuzzy_hostname_match_any?(hostname, matches)
-        if (!hostname.nil?) && (!matches.nil?)
+        unless hostname.nil? || matches.nil?
           return matches.to_s.split(/\s*,\s*/).compact.any? do |m|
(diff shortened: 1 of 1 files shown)


chef-config/lib/chef-config/mixin/fuzzy_hostname_matcher.rb
```diff
@@ -21,7 +21,7 @@ module Mixin
     module FuzzyHostnameMatcher
 
       def fuzzy_hostname_match_any?(hostname, matches)
-        if (!hostname.nil?) && (!matches.nil?)
+        unless hostname.nil? || matches.nil?
```
phiggins, line 24: How about `if hostname && matches`?

tas50: I know historically we've battled over using .nil? or not. I was trying to maintained the .nil? behavior while untangling the knot in my brain.

lamont-granquist: if there's no ambiguity over false being valid and nil meaning "invalid value that deliberately needs to avoid triggering on false" then dropping the `nil?` check reads better.

the only reason to keep that is worry over false.

tas50: Updates to plain hostname/matches. For our use nil means an env var wasn't there and that's unset so there's no match.

PR #10562, opened 2020-10-27
Avoid calculating self.etc_chef_dir multiple times on startup

tas50:
Running chef-apply we calculated this value 5 times. We also have several other resources that use this method which would again have to be calculated. Each of those calculations run the PathHelper.join which involves a whole series of terrible string manipulations. Just calculate this once and be done with it.

### DIFF
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -84,8 +84,10 @@ module ChefConfig
     # @return [String] the platform-specific path
     #
     def self.etc_chef_dir(windows: ChefUtils.windows?)
-      path = windows ? c_chef_dir : PathHelper.join("/etc", ChefUtils::Dist::Infra::DIR_SUFFIX, windows: windows)
-      PathHelper.cleanpath(path, windows: windows)
+      @etc_chef_dir ||= begin
+        path = windows ? c_chef_dir : PathHelper.join("/etc", ChefUtils::Dist::Infra::DIR_SUFFIX, windows: windows)
+        PathHelper.cleanpath(path, windows: windows)
+      end
     end
 
     # On *nix, /var/chef, on Windows C:\chef
--- a/chef-config/lib/chef-config/config.rb
+++ b/chef-config/lib/chef-config/config.rb
@@ -84,7 +84,8 @@ module ChefConfig
     # @return [String] the platform-specific path
     #
     def self.etc_chef_dir(windows: ChefUtils.windows?)
-      @etc_chef_dir ||= begin
+      @etc_chef_dir ||= {}
+      @etc_chef_dir[windows] ||= begin
         path = windows ? c_chef_dir : PathHelper.join("/etc", ChefUtils::Dist::Infra::DIR_SUFFIX, windows: windows)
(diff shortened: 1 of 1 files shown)


tas50:
Spec failures are an easy cleanup. I'm mostly curious how folks feel about this small optimization. I can fix those if this is something we'd want to do.

phiggins:
I think the idea is good, but for the implementation I think it'd be better to use an instance variable and tie it into the `Chef::Config.reset` mechanism that's already used in the tests.

lamont-granquist:
yeah don't use the class variable, it shouldn't be necessary.

tying into `.reset` would probably fix the test failures.

tas50:
This is one of those fun ones where the local config on your workstation leaks into tests and causes tons of spec failures. Works in CI though.

phiggins (approved):

jaymzh (approved):

chef-config/lib/chef-config/config.rb
```diff
@@ -84,8 +84,10 @@ def self.platform_specific_path(path)
     # @return [String] the platform-specific path
     #
     def self.etc_chef_dir(windows: ChefUtils.windows?)
-      path = windows ? c_chef_dir : PathHelper.join("/etc", ChefUtils::Dist::Infra::DIR_SUFFIX, windows: windows)
-      PathHelper.cleanpath(path, windows: windows)
+      @etc_chef_dir ||= begin
```
phiggins, line 87:
```suggestion
      @etc_chef_dir ||= {}
      @etc_chef_dir[windows] ||= begin
```

Kinda gross, but what about this?

jaymzh: I'm not clear why that's better?

phiggins: This function takes a boolean that controls which of two values are returned, but this new caching mechanism will only cache the value for whichever param it is called with first.

lamont-granquist:
@tas50 you want to rebase this and see if it passes?

tas50:
It needs some spec love. The land of mocks is a brutal place

lamont-granquist:
I can't replicate the spec failures.

I think the best thing to do is just write some new specs for this behavior.

The existing specs are in chef-config/spec/unit/config_spec.rb

I think the only test I can usefully think of is to have a single test which tests both the `windows: true` and `windows: false` cases and makes sure that the results are correct.

```
  expect(ChefConfig::Config.etc_chef_dir(windows: false)).to eql("/etc/#{dirname}")
  expect(ChefConfig::Config.etc_chef_dir(windows: true)).to eql("C:\\#{dirname}")
```

Which is just to ensure we don't memoize the same single result for both, which was a problem in the first iteration of this change.

Oh you could write some tests like:

```
  expect(ChefConfig::Config.etc_chef_dir(windows: false)).to eql("/etc/#{dirname}")
  expect(PathHelper).not to receive(:cleanpath)
  expect(ChefConfig::Config.etc_chef_dir(windows: false)).to eql("/etc/#{dirname}")
```

That'd assert that on the second call we don't redo the work.

tas50:
Closing in favor of https://github.com/chef/chef/pull/12581

PR #10567, opened 2020-10-28
Improve the package docs generation + resolve rubocop warnings

tas50:
Resolve a few valid warnings there and then add a few more resources where we should be showing package warnings.

### DIFF
--- a/tasks/docs.rb
+++ b/tasks/docs.rb
@@ -172,7 +172,10 @@ namespace :docs_site do
       properties = {}
 
       # these package properties support passing arrays for the package name
-      properties["common_resource_functionality_multiple_packages"] = true if %w{snap_package dpkg_package yum_package apt_package zypper_package homebrew_package dnf_package pacman_package homebrew_package}.include?(name)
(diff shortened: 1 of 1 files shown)


tasks/docs.rb
```diff
@@ -168,11 +168,11 @@ def properties_list(properties)
       end
     end
 
-    def special_properties(name, data)
+    def special_properties(name)
       properties = {}
 
       # these package properties support passing arrays for the package name
-      properties["common_resource_functionality_multiple_packages"] = true if %w{yum_package apt_package zypper_package homebrew_package dnf_package pacman_package homebrew_package}.include?(name)
+      properties["common_resource_functionality_multiple_packages"] = true if %w{snap_package dpkg_package yum_package apt_package zypper_package homebrew_package dnf_package pacman_package homebrew_package}.include?(name)
```
phiggins, line 175: I'm curious the difference between this and `"properties_multiple_packages"` 🤔

phiggins (approved):

tasks/docs.rb
tas50: It looks like we have some duplicate yaml items there. We should be setting both the same and I updated that. Longterm we should nuke one of those.

PR #10570, opened 2020-10-28
Improve resource documentation

tas50:
Get closer to just being able to auto-generate the world

### DIFF
--- a/lib/chef/resource/scm/_scm.rb
+++ b/lib/chef/resource/scm/_scm.rb
@@ -33,7 +33,8 @@ property :revision, String,
   default: "HEAD"
 
 property :user, [String, Integer],
-  description: "The system user that will own the checked-out code."
+  description: "The system user that will own the checked-out code.",
+  default_description: "`HOME` environment variable of the current user"
 
 property :group, [String, Integer],
   description: "The system group that will own the checked-out code."
--- a/lib/chef/resource/scm/_scm.rb
+++ b/lib/chef/resource/scm/_scm.rb
@@ -34,7 +34,7 @@ property :revision, String,
 
 property :user, [String, Integer],
   description: "The system user that will own the checked-out code.",
-  default_description: "`HOME` environment variable of the current user"
+  default_description: "`HOME` environment variable of the user running #{ChefUtils::Dist::Infra::CLIENT}"
 
 property :group, [String, Integer],
   description: "The system group that will own the checked-out code."
--- a/lib/chef/resource/scm/git.rb
+++ b/lib/chef/resource/scm/git.rb
@@ -28,6 +28,87 @@ class Chef
       provides :git
 
       description "Use the **git** resource to manage source control resources that exist in a git repository. git version 1.6.5 (or higher) is required to use all of the functionality in the git resource."
+      examples <<~DOC
+      **Use the git mirror**
+
+      ```ruby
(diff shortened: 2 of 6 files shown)


lib/chef/resource/scm/_scm.rb
```diff
@@ -33,7 +33,8 @@
   default: "HEAD"
 
 property :user, [String, Integer],
-  description: "The system user that will own the checked-out code."
+  description: "The system user that will own the checked-out code.",
+  default_description: "`HOME` environment variable of the current user"
```
phiggins, line 37: Is this description accurate? If so, I'm not sure I understand it.

tas50: It's ENV['HOME'] of the user running chef-client. I tried to write something brief so it fits in the format of the docs site.

lib/chef/resource/scm/git.rb
```diff
@@ -28,6 +28,87 @@ class Git < Chef::Resource
...
+      git '/home/user/deployment' do
+         repository 'git@github.com:gitsite/deployment.git'
+         revision branch_name
+         action :sync
+         user 'user'
+         group 'test'
+      end
+      ```
+
+      Where the branch_name variable is set to staging or master, depending on the environment of the node. Once this is determined, the branch_name variable is used to set the revision for the repository. If the git status command is used after running the example above, it will return the branch name as deploy, as this is the default value. Run Chef Infra Client in debug mode to verify that the correct branches are being checked out:
```
phiggins, line 62:
```suggestion
      Where the `branch_name` variable is set to staging or master, depending on the environment of the node. Once this is determined, the `branch_name` variable is used to set the revision for the repository. If the git status command is used after running the example above, it will return the branch name as `deploy`, as this is the default value. Run #{Chef::Dist::PRODUCT} in debug mode to verify that the correct branches are being checked out:
```

tas50: I know previously having variables in the descriptions caused a lot of issues so I was avoiding that all together. We have tons of branding in the examples sections. cinc isn't consuming that really anywhere so I've punted on pulling the variables in. If that's something they want to tackle in the future, it's best done there.

lib/chef/resource/systemd_unit.rb
```diff
@@ -29,6 +29,47 @@ class SystemdUnit < Chef::Resource
...
+                }})
+        action [:create, :enable]
+      end
+      ```
+
+      **Create systemd service unit file from a String**
+
+      ```ruby
+      systemd_unit 'sysstat-collect.timer' do
+        content <<-EOU.gsub(/^\s+/, '')
```
phiggins, line 58:
```suggestion
        content <<~EOU
```

lib/chef/resource/windows_task.rb
```diff
@@ -75,7 +75,7 @@ class WindowsTask < Chef::Resource
         command 'chef-client'
         run_level :highest
         frequency :once
-        start_time "16:10"
+        start_time 16:10'
```
phiggins, line 78:
```suggestion
        start_time '16:10'
```

phiggins (approved):

lib/chef/resource/scm/git.rb
tas50: I pulled in the just the ``

PR #10572, opened 2020-10-29
Remove the provides :package for solaris_package

tas50:
This only worked on Solaris 5.10, which we don't support anymore. Just
nuke the line entirely


phiggins (approved):

phiggins:
Looks like there are some tests to delete as well.

tas50:
Yeah I want to do the cleanup of the old solaris derivatives first. This led to that and there's more cleanup / fixes here.

tas50:
This goes green once we merge https://github.com/chef/chef/pull/10576

PR #10573, opened 2020-10-29
Remove support for nexentacore and opensolaris which are both a decade EOL

tas50:
### Nexenta

https://en.wikipedia.org/wiki/Nexenta_OS

"In late 2011, the Nexenta OS brand was terminated and replaced with Illumian, which is derived from community development for illumos and OpenIndiana, but was distinguished by its use of Debian packaging. Illumian version 1.0 was released in February 2012. Following the initial release of Illumian in 2012, the Illumian project was discontinued.[3]"

and

"Final release	3.1.3.5 (October 31, 2012; 7 years ago)"

### OpenSolaris

https://en.wikipedia.org/wiki/OpenSolaris

"OpenSolaris is a discontinued open source computer operating system based on Solaris and created by Sun Microsystems."

and

Latest release	2009.06 / June 1, 2009; 11 years ago

These are beyond dead and we never supported either of them officially. Just nuke it along with the helpers. There's no way anyone used these.

### DIFF
--- a/lib/chef/resource/solaris_package.rb
+++ b/lib/chef/resource/solaris_package.rb
@@ -25,7 +25,6 @@ class Chef
       unified_mode true
 
       provides :solaris_package
-      provides :package, os: "solaris2", platform_family: "nexentacore"
       provides :package, os: "solaris2", platform_family: "solaris2", platform_version: "<= 5.10"
 
       description "Use the **solaris_package** resource to manage packages on the Solaris platform."
--- a/spec/functional/resource/cron_spec.rb
+++ b/spec/functional/resource/cron_spec.rb
@@ -26,7 +26,7 @@ describe Chef::Resource::Cron, :requires_root, :unix_only do
   # Platform specific validation routines.
   def cron_should_exists(cron_name, command)
     case ohai[:platform]
-    when "aix", "opensolaris", "solaris2", "omnios"
+    when "aix", "solaris2", "omnios"
       expect(shell_out("crontab -l #{new_resource.user} | grep \"#{cron_name}\"").exitstatus).to eq(0)
       expect(shell_out("crontab -l #{new_resource.user} | grep \"#{cron_name}\"").stdout.lines.to_a.size).to eq(1)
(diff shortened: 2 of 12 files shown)


phiggins (approved):

lib/chef/resource/solaris_package.rb
```diff
@@ -25,7 +25,6 @@ class SolarisPackage < Chef::Resource::Package
       unified_mode true
 
       provides :solaris_package
-      provides :package, os: "solaris2", platform_family: "nexentacore"
       provides :package, os: "solaris2", platform_family: "solaris2", platform_version: "<= 5.10"
```
jaymzh, line 28: should openindiana be here?

spec/functional/resource/cron_spec.rb
```diff
@@ -26,7 +26,7 @@
   # Platform specific validation routines.
   def cron_should_exists(cron_name, command)
     case ohai[:platform]
-    when "aix", "opensolaris", "solaris2", "omnios"
+    when "aix", "solaris2", "omnios"
```
jaymzh, line 29: openindiana?

spec/functional/resource/cron_spec.rb
```diff
@@ -113,7 +113,7 @@ def cron_should_not_exists(cron_name)
     end
   end
 
-  exclude_solaris = %w{solaris opensolaris solaris2 omnios}.include?(ohai[:platform])
+  exclude_solaris = %w{solaris solaris2 omnios}.include?(ohai[:platform])
```
jaymzh, line 116: same

lib/chef/resource/solaris_package.rb
tas50: OpenIndiana is 5.11 based so it's IPS

spec/functional/resource/cron_spec.rb
tas50: I think we can skip these just because we're in functional and that's never gonna happen

jaymzh (approved):

PR #10576, opened 2020-10-29
Remove the provider_resolver specs that are not helpful

tas50:
This is a pretty useless test at this point:

- Platform/platform_family/version combinations are invalid in a way
  that makes tests pass when they should not. Ex: all the solaris stuff
- Platforms we don't support anymore
- Fake version numbers that mean we're not testing what would actually
  make sense


phiggins (approved):
Here's a 👍, but I don't know the history here and I'm just taking your word for it that these aren't useful.

PR #10579, opened 2020-10-29
Remove rspec_junit_formatter and rspec version pins

phiggins:
This was added a long time ago (https://github.com/chef/chef/commit/87612bf41873efd9c783be7732415c23a171f58d) and exists to write test output in junit format to a file named "test.xml" that does not appear to be consumed by anything.

I figure if this breaks something we'll find out pretty quickly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,9 +54,7 @@ end
 
 group(:development, :test) do
   gem "rake"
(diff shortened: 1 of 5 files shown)


tas50:
You had me at removing a dep

tas50 (approved):

PR #10584, opened 2020-10-30
Simplify regexes by removing extra character classes

tas50:
We're working too hard here. No need for these.


phiggins (approved):
Looks ok to me 👍

PR #10601, opened 2020-11-04
Use .compact instead of .select/.reject to remove nils

tas50:
This is 3-5x faster from the benchmarks I've run locally. There's a cop
coming in the next release of RuboCop to enforce this going forward.

### DIFF
--- a/lib/chef/provider/package/chocolatey.rb
+++ b/lib/chef/provider/package/chocolatey.rb
@@ -79,7 +79,7 @@ class Chef
           name_versions_to_install = desired_name_versions.select { |n, v| lowercase_names(names).include?(n) }
 
           name_nil_versions = name_versions_to_install.select { |n, v| v.nil? }
-          name_has_versions = name_versions_to_install.reject { |n, v| v.nil? }
+          name_has_versions = name_versions_to_install.compact
 
           # choco does not support installing multiple packages with version pins
           name_has_versions.each do |name, version|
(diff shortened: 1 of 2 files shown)


lib/chef/provider/package/chocolatey.rb
```diff
@@ -79,7 +79,7 @@ def install_package(names, versions)
           name_versions_to_install = desired_name_versions.select { |n, v| lowercase_names(names).include?(n) }
 
           name_nil_versions = name_versions_to_install.select { |n, v| v.nil? }
-          name_has_versions = name_versions_to_install.reject { |n, v| v.nil? }
+          name_has_versions = name_versions_to_install.compact
```
tas50, line 82: I need to double check this one. This seems like it's not a compatible change

phiggins: I don't think this is going to do the same thing. You could replace the dual calls to select and reject with partition though:

```suggestion
          name_nil_versions, name_has_versions = name_versions_to_install.partition { |_, v| v.nil? }
```

tas50: TIL compact actually does the right thing in a hash.

```
irb(main):001:0> {foo: "bar", baz: nil}
=> {:foo=>"bar", :baz=>nil}
irb(main):002:0> {foo: "bar", baz: nil}.compact
=> {:foo=>"bar"}
```

phiggins: wat

phiggins: TIL Hash has its own implementation of compact and doesn't just inherit from Enumerable or whatever I was expecting https://ruby-doc.org/core-2.7.2/Hash.html#method-i-compact

tas50: I was all ready to fix this in RuboCop when I realized they're actually testing for it

PR #10602, opened 2020-11-05
Remove duplicate Gemfile gems + update ohai to 16.7.13

tas50:
Remove gemfile gems that are also in the gemspec.


btm (approved):
yeah, although I have no idea if one is better than the other right now.

tas50:
It should really be in the gemspec so knife works in a gem install

PR #10604, opened 2020-11-06
Collapse several duplicate branches down

tas50:
Write more compact case statements to prevent duplicate branches.

### DIFF
--- a/lib/chef/file_access_control/windows.rb
+++ b/lib/chef/file_access_control/windows.rb
@@ -255,10 +255,7 @@ class Chef
             flags |= CONTAINER_INHERIT_ACE
           when :objects_only
             flags |= OBJECT_INHERIT_ACE
-          when true
-            flags |= CONTAINER_INHERIT_ACE
-            flags |= OBJECT_INHERIT_ACE
-          when nil
+          when true, nil
             flags |= CONTAINER_INHERIT_ACE
             flags |= OBJECT_INHERIT_ACE
           end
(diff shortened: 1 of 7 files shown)


lib/chef/file_access_control/windows.rb
```diff
@@ -255,10 +255,7 @@ def calculate_flags(rights)
             flags |= CONTAINER_INHERIT_ACE
           when :objects_only
             flags |= OBJECT_INHERIT_ACE
-          when true
-            flags |= CONTAINER_INHERIT_ACE
-            flags |= OBJECT_INHERIT_ACE
-          when nil
+          when true, nil
```
phiggins, line 258: Maybe just `else` here?

lib/chef/node/attribute_collections.rb
```diff
@@ -73,9 +73,7 @@ def to_yaml(*opts)
 
       def convert_value(value)
         case value
-        when VividMash
-          value
-        when AttrArray
+        when VividMash, AttrArray
```
phiggins, line 76: This code is repeated about 50 lines down.

phiggins (approved):

lib/chef/node/attribute_collections.rb
tas50: Fixed

lib/chef/file_access_control/windows.rb
tas50: I'm not sure if there are additional values we wouldn't want to perform the action for though. Else might be fine, but it could also change the behavior.

tas50: False for example would get caught by the else

Issue #10607, opened 2020-11-08
Error if log_location is set in chef_client_config resource


srb3:
## Description
Setting the log_location in the chef_client_config resource causes the client run to fail

## Chef Version
16.6.14

## Platform Version
CentOS 7.9

## Replication Case
Set the following recipe code
```
chef_client_config 'create client.rb' do
    chef_server_url  Chef::Config[:chef_server_url]
    log_location :syslog
    ssl_verify_mode :verify_none
  end
```

## Client Output

```
   
      ================================================================================
      Error executing action `create` on resource 'template[/etc/chef/client.rb]'
 [10.0.1.34]       ================================================================================
      
      Chef::Mixin::Template::TemplateError
      ------------------------------------
      class or module required
      
      Resource Declaration:
      ---------------------
      # In /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-16.6.14/lib/chef/resource/chef_client_config.rb
      
      249:         template ::File.join(new_resource.config_directory, "client.rb") do
      250:           source ::File.expand_path("support/client.erb", __dir__)
      251:           user new_resource.user unless new_resource.user.nil?
      252:           group new_resource.group unless new_resource.group.nil?
      253:           local true
      254:           variables(
      255:             chef_license: new_resource.chef_license,
      256:             chef_server_url: new_resource.chef_server_url,
      257:             event_loggers: new_resource.event_loggers,
      258:             exception_handlers: format_handler(new_resource.exception_handlers),
      259:             file_backup_path: new_resource.file_backup_path,
      260:             file_cache_path: new_resource.file_cache_path,
      261:             file_staging_uses_destdir: new_resource.file_staging_uses_destdir,
      262:             formatters: new_resource.formatters,
      263:             http_proxy: new_resource.http_proxy,
      264:             https_proxy: new_resource.https_proxy,
      265:             ftp_proxy: new_resource.ftp_proxy,
      266:             log_level: new_resource.log_level,
      267:             log_location: new_resource.log_location,
      268:             minimal_ohai: new_resource.minimal_ohai,
      269:             named_run_list: new_resource.named_run_list,
      270:             no_proxy: new_resource.no_proxy,
      271:             node_name: new_resource.node_name,
      272:             ohai_disabled_plugins: new_resource.ohai_disabled_plugins,
      273:             ohai_optional_plugins: new_resource.ohai_optional_plugins,
      274:             pid_file: new_resource.pid_file,
[...]

Issue #10606, opened 2020-11-08
chef_client_config is not rendering a correct client.rb


srb3:
## Description
It seems that the client_config template is inserting an `=` sign in-between properties and values.
This does not seem to be valid config causing the chef-client binary to ignore the config values.

## Chef Version
16.6.14

## Platform Version
CentOS 7.9
Windows 2019

## Replication Case
Create a recipe with:
```
chef_client_config 'create client.rb' do
    chef_server_url  Chef::Config[:chef_server_url]
    ssl_verify_mode :verify_none
end
```

run chef-client on a node with that cookbook and recipe 
inspect the content of `/etc/chef/client.rb`
```
[root@ip-10-0-1-34 ~]# cat /etc/chef/client.rb                                                            
node_name = "linux-client-0"                                                                              
chef_server_url = "https://10.0.1.5/organizations/acmecorp"
ssl_verify_mode = :verify_none
```
log output from running chef client
```
[root@ip-10-0-1-34 ~]# chef-client -c /etc/chef/client.rb                                                                                                                                                            
Starting Chef Infra Client, version 16.6.14
Patents: https://www.chef.io/patents                                                                      
[2020-11-08T11:06:23+00:00] ERROR: Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, retry 1/5
[2020-11-08T11:06:28+00:00] ERROR: Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, retry 2/5
[2020-11-08T11:06:33+00:00] ERROR: Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, retry 3/5
[2020-11-08T11:06:38+00:00] ERROR: Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, retry 4/5
[2020-11-08T11:06:43+00:00] ERROR: Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, retry 5/5
                                                     
================================================================================
Chef encountered an error attempting to load the node data for "ip-10-0-1-34.eu-west-1.compute.internal"
================================================================================
                                                     
Networking Error:                   
-----------------
Connection refused - Connection refused connecting to https://localhost/nodes/ip-10-0-1-34.eu-west-1.compute.internal, giving up
                                                     
Your chef_server_url may be misconfigured, or the network could be down.
[...]

PR #10608, opened 2020-11-08
chef_client_config: Resolve invalid configuration in client.rb

srb3:
## Description

Address two issues in the chef_client_config resource. It removes unneeded `=` signs and fixes logic for log_location

## Related Issue
#10606 
#10607 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/support/client.erb
+++ b/lib/chef/resource/support/client.erb
@@ -20,12 +20,12 @@
       @policy_group
       @policy_name
       @ssl_verify_mode).each do |prop| -%>
-<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? -%>
+<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? || prop == '@log_location' -%>
 <%=prop.delete_prefix("@") %> <%= instance_variable_get(prop).inspect %>
 <% end -%>
 <%# log_location is special due to STDOUT/STDERR from String -> IO Object -%>
 <% unless @log_location.nil? %>
-  <% if @log_location.is_a? String && %w(STDOUT STDERR).include?(@log_location) -%>
+  <% if (@log_location.is_a? String) && (%w(STDOUT STDERR).include?(@log_location)) -%>
 log_location <%= @log_location %>
   <% else -%>
 log_location <%= @log_location.inspect %>
--- a/lib/chef/resource/support/client.erb
+++ b/lib/chef/resource/support/client.erb
@@ -10,7 +10,6 @@
       @https_proxy
       @ftp_proxy
       @log_level
-      @log_location
       @minimal_ohai
       @named_run_list
       @no_proxy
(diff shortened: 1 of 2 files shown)


lib/chef/resource/support/client.erb
```diff
@@ -20,15 +20,15 @@
...
-<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? -%>
-<%=prop.delete_prefix("@") %> = <%= instance_variable_get(prop).inspect %>
+<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? || prop == '@log_location' -%>
+<%=prop.delete_prefix("@") %> <%= instance_variable_get(prop).inspect %>
 <% end -%>
 <%# log_location is special due to STDOUT/STDERR from String -> IO Object -%>
 <% unless @log_location.nil? %>
-  <% if @log_location.is_a? String && %w(STDOUT STDERR).include?(@log_location) -%>
-log_location = <%= @log_location %>
+  <% if (@log_location.is_a? String) && (%w(STDOUT STDERR).include?(@log_location)) -%>
```
tas50, line 28: Is this line change necessary? Ruby should do the right thing with the order of operations without it.

srb3: I agree it should, but without the parentheses, it throws the following:
```
  
      Chef::Mixin::Template::TemplateError
      class or module required
```
I don't understand why this error is produced, but adding the parentheses stops erb getting upset.

tas50: Maybe try @log_location.is_a?(String)

lib/chef/resource/support/client.erb
```diff
@@ -20,15 +20,15 @@
       @policy_group
       @policy_name
       @ssl_verify_mode).each do |prop| -%>
-<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? -%>
-<%=prop.delete_prefix("@") %> = <%= instance_variable_get(prop).inspect %>
+<% next if instance_variable_get(prop).nil? || instance_variable_get(prop).empty? || prop == '@log_location' -%>
```
phiggins, line 23: Since `log_location` gets special treatment below, it should be removed from the list above instead of being skipped like this.

lib/chef/resource/support/client.erb
phiggins: Is it necessary to check that `@log_location` is a string, then immediately also check that it is equal to one of two other strings? We already know from the check above that it is not `nil`, so what other scenarios are we trying to guard against?

```suggestion
  <% if %w(STDOUT STDERR).include?(@log_location) -%>
```

phiggins: If this does need some more complicated logic it might be worth moving it into the resource where it could be potentially unit tested, etc, instead of embedded in the erb file.

lib/chef/resource/support/client.erb
srb3: agreed

lib/chef/resource/support/client.erb
srb3: @phiggins to my knowledge the other scenarios are:
 `@log_location` can be a symbol as in `:syslog` or `:win_evt` or a path to a file, or the ruby constants STDOUT or STDERR

@tas50 `@log_location.is_a?(String)` works and I am happy the logic is simple enough to remain in the template

tas50:
Thanks @srb3

PR #10609, opened 2020-11-08
Use ||= where we can

tas50:

### DIFF
--- a/lib/chef/cookbook_version.rb
+++ b/lib/chef/cookbook_version.rb
@@ -588,7 +588,6 @@ class Chef
 
     def file_vendor
       @file_vendor ||= Chef::Cookbook::FileVendor.create_from_manifest(cookbook_manifest)
-      @file_vendor
     end
 
   end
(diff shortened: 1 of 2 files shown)


lib/chef/cookbook_version.rb
```diff
@@ -587,9 +587,7 @@ def yml_filenames_by_name(records)
     end
 
     def file_vendor
-      unless @file_vendor
-        @file_vendor = Chef::Cookbook::FileVendor.create_from_manifest(cookbook_manifest)
-      end
+      @file_vendor ||= Chef::Cookbook::FileVendor.create_from_manifest(cookbook_manifest)
       @file_vendor
```
jaymzh, line 591:
```suggestion
      @file_vendor ||= Chef::Cookbook::FileVendor.create_from_manifest(cookbook_manifest)
```

jaymzh: the ||= should return the value either way...

tas50: Yep. Added

jaymzh (approved):

Issue #10453, opened 2020-09-21
Differences in mount vs findmount output causing idempotency issues


axelrtgs:
## Description
after some digging looks like the switch to findmnt may be causing the issue as we didn’t have a problem before and chef mounts the drive so its not something manually mounted. need to investigate more but not sure if anyone might have an idea on the extra section in the path shown in the findmnt output?
findmnt -rn output
`/backups //MTLNAS/Servers/backup[/backup] cifs rw,relatime`
mount output
`//MTLNAS/Servers/backup on /backups type cifs (rw,relatime)`

double checking the regex matching in the PR it looks like it doesnt account for the random [/backup] that gets outputted in the findmnt output vs mount output. fstab also doesnt have that extra notation and matches the mount output

one of my colleagues looked into it and found this recent PR and figured something might be the cause: https://github.com/chef/chef/pull/10171/files

## Chef Version
16.5

## Platform Version
Linux (CentOS/Ubuntu)

## Replication Case
Use a CIFs mount that has a different output in mount vs findmnt. mounts will keep trying to get remounted but fail because the device is already mounted.

## Client Output

```

```

lamont-granquist:
problem seems to be the argument in square brackets here:

```
 /backups //MTLNAS/Servers/backup[/backup] cifs rw,relatime
```

offhand it looks like that is the path on the cifs server which isn't very useful, and should just be stripped out?

lamont-granquist:
It looks like there's a pattern match in the provider which matches stuff that has square brackets but fails on this, and its not clear that it would be appropriate (since it seems like we need to match against `//MTLNAS/Servers/backup` and not `/backup` in that case, and of course there's no tests -- although there is a rubular link for the kind of moutpoint that the existing code is attempting to mount against.

i have no idea.

tas50:
@antima-gupta Any chance you can take a look at this one along with your other mount work?

antima-gupta:
@tas50 Sure I'll look into this, currently working on another priority stuff.

antima-gupta:
@axelrtgs To make `mount` resource idempotent, I used `findmnt -rn` command. `findmnt -rn` list the currently mounted file systems along with target mount point, and the source device.
Since it's idempotent, so If mounts will keep trying to get remounted, it will fail because the device is already mounted.

[...]

PR #10614, opened 2020-11-10
mount: Fixes for findmount output causing idempotency issues

antima-gupta:

## Description
- Fixes for CIFS mount not idempotent issue.
- Added spec for network device found as mounted.
- Removed '/' from end of the device value.

## Related Issue
#10453 

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/mount/linux.rb
+++ b/lib/chef/provider/mount/linux.rb
@@ -53,8 +53,8 @@ class Chef
             when %r{\A#{Regexp.escape(real_mount_point)}\s+([/\w])+\[#{device_mount_regex}\]\s}
               mounted = true
               logger.trace("Bind device #{device_logstring} mounted as #{real_mount_point}")
-            # Permalink for network device mounted to an existing mount point: https://rubular.com/r/HicK63SqchPLvk
-            when %r{\A#{Regexp.escape(real_mount_point)}\s+#{device_mount_regex}\[([\/\w])+\]\s}
+            # Permalink for network device mounted to an existing mount point: https://rubular.com/r/JRTXXGFdQtwCD6
+            when %r{\A#{Regexp.escape(real_mount_point)}\s+#{device_mount_regex}\[}
               mounted = true
               logger.trace("Network device #{device_logstring} mounted as #{real_mount_point}")
             end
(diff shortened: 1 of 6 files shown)


lib/chef/provider/mount/linux.rb
```diff
@@ -53,6 +53,10 @@ def mounted?
             when %r{\A#{Regexp.escape(real_mount_point)}\s+([/\w])+\[#{device_mount_regex}\]\s}
               mounted = true
               logger.trace("Bind device #{device_logstring} mounted as #{real_mount_point}")
+            # Permalink for network device mounted to an existing mount point: https://rubular.com/r/HicK63SqchPLvk
+            when %r{\A#{Regexp.escape(real_mount_point)}\s+#{device_mount_regex}\[([\/\w])+\]\s}
```
lamont-granquist, line 57: maybe just this?

`%r{\A#{Regexp.escape(real_mount_point)}\s+#{device_mount_regex}\[}`

does any of the stuff after the bracket really matter in terms of determining a match or not?

lamont-granquist:
seems fine either way though, but the regexp may be overly complicated

lib/chef/provider/mount/linux.rb
lamont-granquist: the extra matching on the allowed characters within the brackets may actually be brittle as well, if there's ever a mountpoint in the world which doesn't match those characters there.

antima-gupta: Agreed, I have updated the Regex.

antima-gupta:
@lamont-granquist I have updated the PR, please have a look.

lib/chef/resource/mount.rb
```diff
@@ -65,7 +66,7 @@ class Mount < Chef::Resource
 
       property :options, [Array, String, nil],
         description: "An array or comma separated list of options for the mount.",
-        coerce: proc { |arg| arg.is_a?(String) ? arg.split(",") : arg },
+        coerce: proc { |arg| mount_options(arg) }, # Please see #mount_options method.
```
tas50, line 69: Any time we do a coerce we should make sure to add specs to test the various scenarios

tas50: I added those here so we don't block on merging this: https://github.com/chef/chef/pull/10625

antima-gupta: > Any time we do a coerce we should make sure to add specs to test the various scenarios

@tas50 Okay, I'll do that.

PR #10621, opened 2020-11-11
Add new homebrew_install resource

tas50:
This was originally done in the desktop cookbook, but makes more sense here


lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,107 @@
...
+          end
+        else
+          build_essential "install Command Line Tools for Xcode" do
+            action :install
+          end
+        end
+
+        directory "/usr/local/Homebrew"
+
+        remote_file "download the Homebrew archive" do
```
tas50, line 81: I still need to investigate if we can just execute the bash install script

tas50: We can't do that since that script has to be run using sudo and we can't do that in a chef resource.

tas50:
This is going to require further investigation to get working. The simple create the homebrew dir and wire up the binlink is not enough for homebrew.

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,152 @@
...
+        description: "The user to install Homebrew as. Note: Homebrew cannot be installed as root.",
+        required: true
+
+      action :install do
+        # Avoid all the work in the below resources if homebrew is already installed
+        return if ::File.exist?('/usr/local/bin/brew')
+
+        BREW_REPO = 'https://codeload.github.com/Homebrew/brew/zip/master'.freeze
+        USER_HOME = Dir.home(new_resource.user).freeze
+        HOMEBREW_PREFIX = '/usr/local'.freeze
```
tas50, line 70: We should probably just kill this and add the prefix the the dirs so it's more obvious what we're trying to create

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,152 @@
...
+        HOMEBREW_REPOSITORY = '/usr/local/Homebrew'.freeze
+        HOMEBREW_CACHE = "#{USER_HOME}/Library/Caches/Homebrew".freeze
+
+        # Creating the basic directory structure needed for Homebrew
+        directories = ['bin', 'etc', 'include', 'lib', 'sbin', 'share', 'var', 'opt',
+                        'share/zsh', 'share/zsh/site-functions',
+                        'var/homebrew', 'var/homebrew/linked',
+                        'Cellar', 'Caskroom', 'Homebrew', 'Frameworks'
+                      ]
```
tas50, line 79:
```suggestion
                      ].freeze
```

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,152 @@
...
+
+class Chef
+  class Resource
+    class HomebrewInstall < Chef::Resource
+      unified_mode true
+
+      provides :homebrew_install
+
+      description "Use the **homebrew_install** resource to install the Homebrew package manager on macOS systems."
+      introduced "16.7"
```
tas50, line 27:
```suggestion
      introduced "16.8"
```

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+        tools_url 'https://somewhere.something.com/downloads/command_line_tools.dmg'
+        tools_pkg_name 'Command Line Tools.pkg'
+        user 'someuser'
+        action :install
+      end
+      ```
+      DOC
+
+      property :tools_url, String,
+        description: "A url pointing to a local source for the Command Line Tools for Xcode dmg"
```
phiggins, line 48: I know we had a discussion about this somewhere, but "A url pointing to a local source" sounds like it would have to be a path to a file on disk or something.

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      property :tools_pkg_name, String,
+        description: "The name of the pkg inside the dmg located at the tools url"
+
+      property :brew_source, String,
+        description: "A url pointing to a Homebrew installer",
+        default: "https://github.com/Homebrew/brew/tarball/master"
+
+      property :user, String,
+        description: "The user to install Homebrew as. Note: Homebrew cannot be installed as root.",
```
phiggins, line 58: > Note: Homebrew cannot be installed as root.

I know the Homebrew install script will bail out if you run it as root, but is there a technical reason someone couldn't do this with this code? If so, should there be a guard against it with an informative error message?

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+      introduced "16.8"
+      examples <<~DOC
+      **Install Homebrew using the Internet to download Command Line Tools for Xcode**:
+      ```ruby
+      homebrew_install 'Install Homebrew and xcode command line tools if necessary' do
+        user 'someuser'
+        action :install
+      end
+      ```
+      **Install Homebrew using a local source to download Command Line Tools for Xcode from**:
```
phiggins, line 36:
```suggestion

      **Install Homebrew using a local source to download Command Line Tools for Xcode from**:
```

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+        action :install
+      end
+      ```
+      DOC
+
+      property :tools_url, String,
+        description: "A url pointing to a local source for the Command Line Tools for Xcode dmg"
+
+      property :tools_pkg_name, String,
+        description: "The name of the pkg inside the dmg located at the tools url"
```
phiggins, line 51: Should the `tools_` properties be renamed `xcode_`? Is tools vs xcode an important distinction?

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      property :brew_source, String,
```
phiggins, line 53: This one should probably be `brew_url` if it has to be a URL.

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      property :user, String,
+        description: "The user to install Homebrew as. Note: Homebrew cannot be installed as root.",
+        required: true
+
+      action :install do
+        # Avoid all the work in the below resources if homebrew is already installed
+        return if ::File.exist?("/usr/local/bin/brew")
+
+        BREW_REPO = "https://codeload.github.com/Homebrew/brew/zip/master".freeze
```
phiggins, line 65: Instead of being hardcoded here, this should probably be the default of the `brew_source` property since that appears to not be used anywhere currently.

lib/chef/resource/homebrew_install.rb
tas50: brew_source_url is probably what we want here

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+          end
+        end
+
+        script "Download and unpack Homebrew" do
+          interpreter "bash"
+          cwd "/usr/local/Homebrew"
+          code <<-CODEBLOCK
+            git init -q
+            curl #{BREW_REPO} -o brew-master.zip
+            unzip brew-master.zip -d /usr/local/Homebrew
```
phiggins, line 113: Should the archive get deleted at some point?

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+            type "pkg"
+          end
+        else
+          build_essential "install Command Line Tools for Xcode" do
+            action :upgrade
+          end
+        end
+
+        script "Download and unpack Homebrew" do
+          interpreter "bash"
```
phiggins, line 108: Any reason not to make this a `bash` resource?

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+          code <<-CODEBLOCK
+            git init -q
+            curl #{BREW_REPO} -o brew-master.zip
+            unzip brew-master.zip -d /usr/local/Homebrew
+          CODEBLOCK
+          user new_resource.user
+        end
+
+        script "move files to their correct locations" do
+          interpreter "bash"
```
phiggins, line 119: Same, should this be a bash resource?

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,147 @@
...
+        cmd.run_command
+
+        cmd = Mixlib::ShellOut.new("/usr/local/bin/brew", "update", "--force", :user => new_resource.user, :environment => nil, :cwd => "/usr/local/Homebrew")
+        cmd.run_command
+
+        local_shell = shell_out("echo $SHELL")
+        if local_shell.stdout.match(/zsh/)
+          shell_out("export PATH="/usr/local/bin:$PATH" >> ~/.zshrc")
+        else
+          shell_out("export PATH="/usr/local/bin:$PATH" >> ~/.bash_profile")
```
phiggins, line 142: These two strings aren't valid syntax, you're putting double quotes inside double quotes with no escaping.

lib/chef/resource/homebrew_install.rb
johnmccrae: I changed the description to make it clearer.

lib/chef/resource/homebrew_install.rb
johnmccrae: done

lib/chef/resource/homebrew_install.rb
johnmccrae: done

lib/chef/resource/homebrew_install.rb
johnmccrae: sorted

lib/chef/resource/homebrew_install.rb
johnmccrae: as discussed, will file a bug

lib/chef/resource/homebrew_install.rb
johnmccrae: as discussed, will file a bug

lib/chef/resource/homebrew_install.rb
johnmccrae: fixed

lib/chef/resource/homebrew_install.rb
johnmccrae: done

lib/chef/resource/homebrew_install.rb
johnmccrae: added an exception to manage this.

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,148 @@
...
+        # Avoid all the work in the below resources if homebrew is already installed
+        return if ::File.exist?("/usr/local/bin/brew")
+
+        # check if 'user' is root and raise an exception if so
+        if Etc.getpwnam(new_resource.user).uid == 0
+          msg = "You are attempting to install Homebrew as Root. This is not permitted by Homebrew. Please run this as a standard user with admin rights"
+          raise Chef::Exceptions::InsufficientPermissions, msg
+        end
+
+        USER_HOME = Dir.home(new_resource.user).freeze
```
tas50, line 71: we should make these lowercase since they're not really constants and that gets confusing

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+
+        if new_resource.xcode_tools_url
+          dmg_package new_resource.xcode_tools_pkg_name do
+            source new_resource.xcode_tools_url
+            type "pkg"
+          end
+        else
+          build_essential "install Command Line Tools for Xcode" do
+            action :upgrade
```
tas50, line 108: Does this need to be an upgrade. That's a very slow code path and maybe not what users are expecting vs. the install

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+
+        script "Download and unpack Homebrew" do
+          interpreter "bash"
+          cwd "/usr/local/Homebrew"
+          code <<-CODEBLOCK
+            git init -q
+            curl #{new_resource.brew_source_url} -o brew-master.zip
+            unzip brew-master.zip -d /usr/local/Homebrew
+            rm /usr/local/Homebrew/brew-master.zip
```
phiggins, line 119: `tar` on macOS can handle zip files and can read from stdin, so this could be done as one command rather than 3. Here's an example of me trying this locally:

```
Petes-MBP:tmp pete$ mkdir -p brew
Petes-MBP:tmp pete$ cd brew/
Petes-MBP:brew pete$ curl "https://codeload.github.com/Homebrew/brew/zip/master" | tar -x -C .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 3274k    0 3274k    0     0  2524k      0 --:--:--  0:00:01 --:--:-- 2524k
Petes-MBP:brew pete$ ls -l
total 0
drwxr-xr-x  19 pete  wheel  608 Dec 10 11:15 brew-master
```

tas50: We should probably convert the download over to remote_file to handle it in a native chef way and download that file to our file cache dir

lib/chef/resource/homebrew_install.rb
johnmccrae: done

lib/chef/resource/homebrew_install.rb
johnmccrae: It does not. I will make it an Install

lib/chef/resource/homebrew_install.rb
johnmccrae: got it

johnmccrae: OK, updated it to use remote_file to get the zip file. Testing now.

phiggins:
It would be cool if we could add this to the kitchen tests, but homebrew must be present for the other tests to work and I'm not sure how it's getting there in the first place: https://github.com/chef/chef/blob/9604db9ced992079daaaf31ec64b82758465cd65/kitchen-tests/cookbooks/end_to_end/recipes/macos.rb#L76

tas50:
So I dug into this resource and right now it's not functional. There's a large amount of setup that happens in the install.sh file for homebrew which we haven't replicated. I started down the path of trying to fix some of that over the weekend, but there is quite a bit more to work on before this is ready.

kagarmoe:
@tas50 What is the status of this? Do we merge or close?

tas50:
We ran into a number of technical challenges in making this work and it was put on hold

kagarmoe (approved):
tenor-13267889

lib/chef/resource/homebrew_install.rb
```diff
@@ -0,0 +1,165 @@
...
+        description: "The user to install Homebrew as. Note: Homebrew cannot be installed as root.",
+        required: true
+
+      property :force, [TrueClass, FalseClass],
+        description: "Force the installation of Homebrew even if the `/usr/local/bin/brew` command already exists on the system.",
+        default: false
+
+      action :install do
+        # Avoid all the work in the below resources if homebrew is already installed
+        return if !new_resource.force && ::File.exist?("/usr/local/bin/brew")
```
khramtsoff, line 70: Location of brew is different for apple silicon (m1)
/opt/homebrew/bin

PR #10628, opened 2020-11-11
Use the full path to the which helper to avoid conflicts with chef-sugar

tas50:
When uses are migrating from legacy chef-client releases with chef-sugar to Chef Infra Client 16+ they need to continue using chef-sugar in their runlists for some time period. We want to make sure that we don't break in those upgrade scenarios. Using the complete path to our which helper makes sure that we get that helper instead of the one from chef-sugar which injects itself into the recipe dsl.


tas50:
No go here

PR #10630, opened 2020-11-11
provide a registry_key example that creates a multibyte binary value

### LINKED ISSUE #10600: registry_key for :binary type is bugged and/or undocumented
## Description

When trying to write a `:binary` value with the `registry_key` resource, the "common" ways of doing it won't write binary data but their ASCII values.

## Chef Version

16.2.73

## Platform Version

Windows 2019

## Replication Case

When writing

```ruby
registry_key 'HKEY_LOCAL_MACHINE\SOFTWARE\TEST' do
  values [{
      name: 'test',
      type: :binary,
      data: 0
    }]
  recursive true
  action :create
end
```

The value will be written but will be 48 (0x30) which is the ASCII character of "0"

The correct way of doing this would be:

```ruby
registry_key 'HKEY_LOCAL_MACHINE\SOFTWARE\TEST' do
  values [{
      name: 'test',
      type: :binary,
      data: 0.chr
    }]
  recursive true
  action :create
end
```

or even be more complicated with multi-byte (`[0, 1, 2].tr(&:chr)`).

This requirement could either be documented in an example or a syntax of providing `[0, 1, 2]` arrays with handling inside `win32/registry.rb` around [line 65](https://github.com/chef/chef/blob/7558c414030d082194d0d2b5f74279ce151517e9/lib/chef/win32/registry.rb#L65):

```ruby
if value[:type] == :binary
  raise Chef::Exceptions::Win32RegTypesMismatch, "Data provided for registry key #{key_path} needs to be an Array of Integers for :binary types" unless data.is_a?(Array) && data.all? { |e| e.is_a? Integer)

  data.map!(&:chr)
end
```

## Client Output

```
       Recipe: test::default
         * registry_key[HKEY_LOCAL_MACHINE\SOFTWARE\TEST] action create
           - set value {:name=>"test", :type=>:binary, :data=>0}
```

## Stacktrace

none

mwrock:
fixes #10600

### DIFF
--- a/lib/chef/resource/registry_key.rb
+++ b/lib/chef/resource/registry_key.rb
@@ -27,6 +27,91 @@ class Chef
       provides(:registry_key) { true }
 
       description "Use the **registry_key** resource to create and delete registry keys in Microsoft Windows."
+      examples <<~DOC
+      **Create a registry key**
+      ```ruby
+      registry_key 'HKEY_LOCAL_MACHINE\path-to-key\Policies\System' do
+        values [{
+          name: 'EnableLUA',
+          type: :dword,
+          data: 0
+        }]
+        action :create
+      end
+      ```
+      
+      **Create a registry key with binary data: "\x01\x02\x03"**:
+      ```ruby
+      registry_key 'HKEY_CURRENT_USER\\ChefTest' do
+        values [{
+          :name => "test",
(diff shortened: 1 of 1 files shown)


tas50:
Nice. Thanks for getting us closer to automated docs

Issue #10641, opened 2020-11-13
Group member administration is not idempotent on Windows


jaymzh:
## Description

The `group` resource on Windows doesn't seem to translate group SIDs to names and thus isn't idempotent as it re-sets the members every run.

## Chef Version
15.9.17

## Platform Version
Windows 10

## Replication Case
```
group 'Administrators' do
  members ['john', 'sarah']
end
```

## Client Output

There is no useful output due to https://github.com/chef/chef/issues/10640
but I added some and here's what you get:

```
[2020-11-12T17:11:43-08:00] INFO: Processing group[Administrators] action create (C:/chef/cache/cookbooks/fb_users/resources/default.rb line 153)
[2020-11-12T17:11:43-08:00] INFO: Set members to ["Administrator", "user1", "user2", "user3"] from ["S-1-5-21-2734142845-4142502760-859509394-500", "S-1-5-21-2734142845-4142502760-859509394-1000", "S-1-5-21-2734142845-4142502760-859509394-1020", "S-1-5-21-2734142845-4142502760-859509394-1023"] for Administrators
[2020-11-12T17:11:43-08:00] INFO: group[Administrators] altered
```

Issue #10640, opened 2020-11-13
group resource doesn't say what changed on Windows


jaymzh:
## Description

Just says "group altered" which is not useful, it should say what changed as other resources do.

## Chef Version
16.6.14

## Platform Version
Windows

I'll have a PR shortly...

PR #10642, opened 2020-11-13
Fix group output and windows support

jaymzh:
Fixes both #10640 and #10641.

* Take the description we build for doc-formatters and put them in the
logs
* Extract out parts of `compare_group` that need to be different on
Windows and implement them in the subclass


jaymzh:
We could do this the way we do it for file/template/etc. where there's a line per type of change ("changed owner from x to y" "changed group from X to Y" etc.) but since a reasoanble changelog was already been created in the parent class I just used that (the other way would require changes in each subclass).

I also want to fix `user`, but ... it's harder, so I'll do that separately.

lib/chef/provider/group.rb
```diff
@@ -103,13 +103,21 @@ def compare_group
           unless members_to_be_removed.empty?
             @change_desc << "remove existing member(s): #{members_to_be_removed.join(", ")}"
           end
-        elsif new_resource.members != current_resource.members
-          @change_desc << "replace group members with new list of members"
+        elsif !group_members_match?
+          @change_desc << "replace group members with new list of members: #{new_resource.members}"
```
tas50, line 107: do we want to join on members here: new_resource.members.join(", ")

jaymzh: yeah slightly nicer output, can do.

jaymzh:
that should sort the tests out as well as add the join.

mwrock (approved):
Very Nice!

PR #10647, opened 2020-11-13
Fix idempotency issues in build_essential on the mac

tas50:
2 issues this solves:

- On some systems the Command Line Tools aren't showing up in the install history plist we're parsing. This means we attempt to install the tools each time chef runs, which takes a LONG time even on a top of the line mac. It was suggested in mac admins slack to just use xcode-select -p which lists the command line tools path or exits 2 if not installed.

- When the :upgrade action is used we were parsing out the software update list twice for no reason, which made it take even longer.


phiggins (approved):
Nice! 👍

PR #10652, opened 2020-11-17
update pwsh in powershell_exec to 7.1.0 and add comments explaining how to pull in updates

mwrock:
This updates the powershell_exec binaries to those build with powershell 7.1.0 deps. This also adds a rake task to automate pulling in updated binaries in the future and adds comments in a few places that explain where the binaries come from.


mkmf.log
```diff
@@ -0,0 +1,4 @@
+find_executable: checking for hab... -------------------- yes
+
+--------------------
+
```
phiggins, line 4: Looks like this got accidentally committed.

PR #10674, opened 2020-11-22
Use URI::DEFAULT_PARSER.make_regexp instead of URI.regexp

tas50:
RuboCop claims that URI.regexp is the obsolete way of doing this. Both produce the same regex.


phiggins (approved):
I could have sworn there was a constant you could reference to do this but I can't find any sources to back that up. The rubocop folks probably know better than I do!

PR #10680, opened 2020-11-23
Remove macOS 10.13 from the build matrix

tas50:
We support N-2 macOS releases so 10.13 can be dropped now that Big Sur
is out.


lamont-granquist (approved):

tas50:
Putting this on hold pending releng work to get macOS 10.14 builders working.

PR #10683, opened 2020-11-24
replace usages of Cmdlet class with powershell_exec

### LINKED ISSUE #10665: utilize powershell_exec in dsc_resource
`dsc_resource` uses the `cmdlet` class to delegate all powershell commands which in turn shells out to `powershell.exe.` The `cmdlet` class and its supporting classes should be refactored to use powershell_exec. This will involve some changes to the underlying powershell shim dll to perform the same verbose stream and json depth special handling.

mwrock:
fixes #10665 

This primarily involved refactoring the dsc_resourse and dsc_script resources.


lib/chef/provider/dsc_script.rb
```diff
@@ -85,20 +84,24 @@ def run_configuration(operation)
...
-          environment: @dsc_resource.environment,
-          timeout: @dsc_resource.timeout,
-        }
+        cwd = @dsc_resource.cwd || Dir.pwd
+        original_env = ENV.to_hash
 
         begin
-          configuration_document = generate_configuration_document(config_directory, configuration_flags)
-          @operations[operation].call(config_manager, configuration_document, shellout_flags)
+          ENV.update(@dsc_resource.environment || original_env)
```
phiggins, line 91:
```suggestion
          ENV.replace(@dsc_resource.environment) if @dsc_resource.environment
```

* I'm not sure what side effects this could have, but given that avoiding the method call is just as easy, this feels midly safer.
* `.update` will merge the provided env variables into the existing ones. Is that what is intended? This appears to be undocumented so I'm not sure.

lib/chef/provider/dsc_script.rb
```diff
@@ -85,20 +84,24 @@ def run_configuration(operation)
...
+              @operations[operation].call(config_manager, configuration_document)
+            end
+          end
         rescue Exception => e
           logger.error("DSC operation failed: #{e.message}")
           raise e
         ensure
           ::FileUtils.rm_rf(config_directory)
+          ENV.clear
+          ENV.update(original_env)
```
phiggins, line 104:
```suggestion
          ENV.replace(original_env)
```

lib/chef/util/dsc/configuration_generator.rb
```diff
@@ -59,6 +57,49 @@ def configuration_document_from_script_path(script_path, configuration_name, con
 
     protected
 
+    def validate_switch_name!(switch_parameter_name)
+      if !!(switch_parameter_name =~ /\A[A-Za-z]+[_a-zA-Z0-9]*\Z/) == false
```
phiggins, line 61:
```suggestion
      unless switch_parameter_name =~ /\A[A-Za-z]+[_a-zA-Z0-9]*\Z/
```

lib/chef/util/dsc/lcm_output_parser.rb
```diff
@@ -75,15 +75,16 @@ module Parser
           #
 
           def self.parse(lcm_output, test_dsc_configuration)
+            lcm_output ||= ""
+            lcm_output = lcm_output.split("\n")
```
phiggins, line 79:
```suggestion
            lcm_output = String(lcm_output).split("\n")
```

lib/chef/util/dsc/configuration_generator.rb
tas50: Since we don't care about the match group we should just use .match? here and avoid a few k of memory usage

spec/functional/resource/dsc_script_spec.rb
```diff
@@ -261,10 +261,10 @@ def delete_user(target_user)
...
       it "should raise an exception if the cwd is etc" do
         dsc_test_resource.cwd(dsc_environment_fail_etc_directory)
-        expect { dsc_test_resource.run_action(:run) }.to raise_error(Chef::Exceptions::PowershellCmdletException)
+        expect { dsc_test_resource.run_action(:run) }.to raise_error(Chef::PowerShell::CommandFailed)
         begin
           dsc_test_resource.run_action(:run)
-        rescue Chef::Exceptions::PowershellCmdletException => e
+        rescue Chef::PowerShell::CommandFailed => e
           expect(e.message).to match(exception_message_signature)
         end
```
phiggins, line 269: This doesn't need to run the code twice to test both of these:

```
expect {
  dsc_test_resource.run_action(:run)
}.to raise_error(Chef::PowerShell::CommandFailed, /#{exception_message_signature}/)
```

lib/chef/provider/dsc_script.rb
mwrock: yeah we want the merge (`update`). Previously we shelled out to powershell and passed the `environment` as a shell arg. This does the equivalent in proc.

spec/unit/util/dsc/configuration_generator_spec.rb
```diff
@@ -25,6 +25,85 @@
...
+    it "should not raise an error if the name contains all lower case letters" do
+      conf_man.send(:validate_switch_name!, "hello")
+    end
+
+    it "should not raise an error if no special characters are used except _" do
+      conf_man.send(:validate_switch_name!, "hello_world")
+    end
+
+    %w{! @ # $ % ^ & * & * ( ) - = + \{ \} . ? < > \\ /}.each do |sym|
+      it "raises an Argument error if it configuration name contains #{sym}" do
```
phiggins, line 42:
```suggestion
      it "raises an ArgumentError if configuration name contains #{sym}" do
```

spec/unit/util/dsc/configuration_generator_spec.rb
```diff
@@ -25,6 +25,85 @@
...
+          conf_man.send(:validate_switch_name!, "Hello#{sym}")
+        end.to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  describe "#escape_parameter_value" do
+    # Is this list really complete?
+    %w{` " # '}.each do |c|
+      it "escapse #{c}" do
```
phiggins, line 53:
```suggestion
      it "escapes #{c}" do
```

phiggins (approved):
Definitely seems like an improvement. 👍

PR #10691, opened 2020-11-25
Resolve new spacing offenses in RuboCop 1.4

tas50:
If we really care we can turn this off, but it seems fine


phiggins (approved):
Not sure I'm a fan of this one...

tas50:
@phiggins We can disable this particular part of the cop in chefstyle if you want

phiggins:
If you like it you should keep it, it's not going to be my concern for too much longer. :)

tas50:
I disabled the 1 offenses here, which made this much more reasonable: https://github.com/chef/chefstyle/pull/115

PR #10699, opened 2020-11-30
Refactor the code for windows_security_policy resource

chef-davin:
## Description

This pull request modifies the backend function of the `windows_security_policy` resource to generate an inf file for secedit to import, rather than doing an export, modifying it with regex in powershell, and then importing it. 

There is no functional change to the resource from a DSL perspective.  Users will use the resource in the same way that they have been.

## Related Issue

Changing the resource to work in this new way resolves a couple issues seen where some users were reporting that changing the Administrator or Guest username was failing, and if there was a multi-line complex legal notification text set in the system, the file couldn't be imported by secedit.

## Types of changes

## Checklist:


chef-davin:
My only real concern is that the resource is now a lot chattier in the logs as its calling three sub-resources in the logs, but I don't know of a way to call the file resource in a quiet manner unless I just want to use ruby File methods instead.  If there are any suggestions on making that a bit more streamlined, I'm all ears.

mwrock:
I'd use a ruby [`Tempfile`](https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html) which should just delete itself and the ShellOut mixin for the secedit call.

chef-davin:
Going to take this out of WIP as all tests are passing and its been rewritten again to have much nicer output in the stdout of a chef-client run.  It also pulls troublesome parts out of secedit import/export and uses `net accounts` commands to do the work when needed.

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,58 @@ class WindowsSecurityPolicy < Chef::Resource
...
+          security_option = new_resource.secoption
+          security_value = new_resource.secvalue
+
+          file = Tempfile.new(["#{security_option}", ".inf"])
+          if security_option == "LockoutBadCount"
+            cmd = "net accounts /lockoutthreshold:#{security_value}"
+          elsif security_option == "ResetLockoutCount"
+            cmd = "net accounts /lockoutwindow:#{security_value}"
+          elsif security_option == "LockoutDuration"
+            cmd = "net accounts /lockoutduration:#{security_value}"
```
phiggins, line 116: Since these three values are flags passed to a windows command, they are likely not case sensitive, so you could use the mixed case version (ie, `LockoutDuration`) and avoid a spelling violation.

Additionally, you could cut down on noise a bit by turning this from if/elsif into a case statement, like:

```
          case security_option
          when "LockoutBadCount"
            cmd = "net accounts /lockoutthreshold:#{security_value}"
          when "ResetLockoutCount"
            cmd = "net accounts /lockoutwindow:#{security_value}"
          when "LockoutDuration"
            cmd = "net accounts /lockoutduration:#{security_value}"
          when "NewAdministratorName", "NewGuestName"
```

etc

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,59 @@ class WindowsSecurityPolicy < Chef::Resource
       description: "Policy value to be set for policy name."
 
       load_current_value do |desired|
+        current_state = load_security_options
+
+        if desired.secoption == "ResetLockoutCount"
+          if desired.secvalue.to_i > 30
```
tas50, line 90: You might want to just do this with a callback right in the property:

https://docs.chef.io/custom_resources/#validators

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,59 @@ class WindowsSecurityPolicy < Chef::Resource
...
+            cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
+          else
+            policy_line = "#{security_option} = #{security_value}"
+            file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
+            file.close
+            file_path = file.path.gsub("/", '\\')
+            cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
+          end
+          powershell_out!(cmd)
+          file.unlink
```
phiggins, line 132:
```suggestion
          Tempfile.new(["#{security_option}", ".inf"]) do |file|
            case security_option
            when "LockoutBadCount"
              cmd = "net accounts /LockoutThreshold:#{security_value}"
            when "ResetLockoutCount"
              cmd = "net accounts /LockoutWindow:#{security_value}"
            when "LockoutDuration"
              cmd = "net accounts /LockoutDuration:#{security_value}"
            when "NewAdministratorName", "NewGuestName"
              policy_line = "#{security_option} = \"#{security_value}\""
              file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
              file_path = file.path.gsub("/", '\\')
              cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
            else
              policy_line = "#{security_option} = #{security_value}"
              file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
              file_path = file.path.gsub("/", '\\')
              cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
            end
            powershell_out!(cmd)
          end
```

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,59 @@ class WindowsSecurityPolicy < Chef::Resource
...
+            file_path = file.path.gsub("/", '\\')
+            cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
+          else
+            policy_line = "#{security_option} = #{security_value}"
+            file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
+            file.close
+            file_path = file.path.gsub("/", '\\')
+            cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
+          end
+          powershell_out!(cmd)
```
tas50, line 131: You might want to look at powershell_exec here

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,59 @@ class WindowsSecurityPolicy < Chef::Resource
...
+            policy_line = "#{security_option} = \"#{security_value}\""
+            file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
+            file.close
+            file_path = file.path.gsub("/", '\\')
+            cmd = "C:\\Windows\\System32\\secedit /configure /db C:\\windows\\security\\new.sdb /cfg #{file_path} /areas SECURITYPOLICY"
+          else
+            policy_line = "#{security_option} = #{security_value}"
+            file.write("[Unicode]\r\nUnicode=yes\r\n[System Access]\r\n#{policy_line}\r\n[Version]\r\nsignature=\"$CHICAGO$\"\r\nRevision=1\r\n")
+            file.close
+            file_path = file.path.gsub("/", '\\')
```
tas50, line 128: Do we actually need to convert the / to \\. Windows generally just does the right thing.

lib/chef/resource/windows_security_policy.rb
```diff
@@ -106,46 +160,10 @@ class WindowsSecurityPolicy < Chef::Resource
             MinimumPasswordAge = $security_options_hash.MinimumPasswordAge
             NewGuestName = $security_options_hash.NewGuestName
             LockoutBadCount = $security_options_hash.LockoutBadCount
-          })
+          }) | ConvertTo-Json
```
tas50, line 163: Powershell exec will handle this pipe for you automatically

lib/chef/resource/windows_security_policy.rb
chef-davin: in my testing on 2016, it very much didn't do the right thing.  It was very unhappy when I used `file.path` directly.

lib/chef/resource/windows_security_policy.rb
mwrock: A plain Shellout would be the best option. Powershell is just going to shell out to `net.exe` or `secedit.exe`. Using mixlib/shellout gets rid of an additional powershell process (or the .Net overhead of powershell_exec).

lib/chef/resource/windows_security_policy.rb
chef-davin: so, I tested this, and `powershell_out` and `powershell_exec` both stopped applying the changes in the cmd.  Not sure why that would be, if there's a bug in putting shellouts in a wrapper like that or something, but it definitely stopped actually running the shellouts on the system.

phiggins: Huh. It could be a buffering issue or something like that, but I understand if you'd rather just move on. 👍

phiggins: If you're curious, I just ran into this myself. `Tempfile.new { }` will not call the block, `Tempfile.create { }` will. 🤦

lib/chef/resource/windows_security_policy.rb
```diff
@@ -83,6 +84,59 @@ class WindowsSecurityPolicy < Chef::Resource
...
+        end
+        if desired.secoption == "ResetLockoutCount" || desired.secoption == "LockoutDuration"
+          if current_state["LockoutBadCount"] == "0"
+            raise Chef::Exceptions::ValidationFailed, "#{desired.secoption} cannot be set unless the \"LockoutBadCount\" security policy has been set to a non-zero value"
+          else
+            secvalue current_state[desired.secoption.to_s]
+          end
+        else
+          secvalue current_state[desired.secoption.to_s]
+        end
```
phiggins, line 102:
```suggestion
        if (desired.secoption == "ResetLockoutCount" || desired.secoption == "LockoutDuration") && current_state["LockoutBadCount"] == "0"
          raise Chef::Exceptions::ValidationFailed, "#{desired.secoption} cannot be set unless the \"LockoutBadCount\" security policy has been set to a non-zero value"
        end
        secvalue current_state[desired.secoption.to_s]
```

phiggins (approved):

PR #10702, opened 2020-12-01
Pin our InSpec version and use Chefstyle from rubygems for now

tas50:
We have an inspec / rubocop conflict that's getting worked out right now, but for now we need to make sure a bundle update doesn't downgrade inspec.


phiggins (approved):

tas50:
@phiggins You'll need to rebase off this to move that inspec version pin into the gemspec

PR #10711, opened 2020-12-07
windows_certificate: Add exportable option to pfx certificate

### LINKED ISSUE #9355: Add the ability for the windows_certificate resource to be exportable.
### Describe the Enhancement:
Add the ability for the `windows_certificate` resource to be exportable.
### Describe the Need:
There is a need from many customers to be able to import a certificate into the Windows Certificate Store and it be marked as `exportable` for certain use cases.
### Current Alternative
Currently, customers are using powershell scripts to accomplish the task of importing the cert, the default when using the `Import-PfxCertificate` is to set the `exportable` flag to `true`, which solves the problem

I can provide more specific information as needed. An example import that enables the `exportable` flag looks like this:

`pfx.Import($certPath,$certPass,"Exportable,PersistKeySet")`

dheerajd-msys:

Improvement in the PFX certificate so as to imported certificate can be exported as user wants.
## Description

When we import the PFX certificate, it's unable to export along with its keys so this improvement by adding `exportable` parameter to the underlying function will ensure the certificate to be exportable as well.
- Exported value as integer can be passed to https://github.com/chef/win32-certstore/blob/master/lib/win32/certstore/store_base.rb#L69 so that certificate become exportable.
- Reference doc : https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-pfximportcertstore
## Related Issue

Fixes #9355
## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/resource/windows_certificate.rb
+++ b/lib/chef/resource/windows_certificate.rb
@@ -87,9 +87,9 @@ class Chef
         description: "Ensure that sensitive resource data is not logged by the #{ChefUtils::Dist::Infra::CLIENT}.",
         default: lazy { pfx_password ? true : false }, skip_docs: true
 
-      property :exportable, Integer,
-        description: "Ensure that imported pfx certificate is exportable. Please provide '1' if you want the certificate to be exportable."
-        default: 0
+      property :exportable, [TrueClass, FalseClass],
+        description: "Ensure that imported pfx certificate is exportable. Please provide 'true' if you want the certificate to be exportable.",
+        default: false
 
       action :create do
         description "Creates or updates a certificate."
@@ -166,8 +166,9 @@ class Chef
         end
 
         def add_pfx_cert
+          exportable = new_resource.exportable ? 1 : 0
           store = ::Win32::Certstore.open(new_resource.store_name)
-          store.add_pfx(new_resource.source, new_resource.pfx_password, new_resource.exportable)
+          store.add_pfx(new_resource.source, new_resource.pfx_password, exportable)
         end
 
         def delete_cert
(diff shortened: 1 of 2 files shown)


lib/chef/resource/windows_certificate.rb
```diff
@@ -87,6 +87,10 @@ class WindowsCertificate < Chef::Resource
         description: "Ensure that sensitive resource data is not logged by the #{ChefUtils::Dist::Infra::CLIENT}.",
         default: lazy { pfx_password ? true : false }, skip_docs: true
 
+      property :exportable, Integer,
+        description: "Ensure that imported pfx certificate is exportable. Please provide '1' if you want the certificate to be exportable."
+        default: 0
+
       action :create do
```
mwrock, line 94: I think this would make more sense to a user if it was true/false instead of an integer.

lib/chef/resource/windows_certificate.rb
```diff
@@ -87,6 +87,10 @@ class WindowsCertificate < Chef::Resource
         description: "Ensure that sensitive resource data is not logged by the #{ChefUtils::Dist::Infra::CLIENT}.",
         default: lazy { pfx_password ? true : false }, skip_docs: true
 
+      property :exportable, Integer,
```
tas50, line 90: This should be true/false

lib/chef/resource/windows_certificate.rb
dheerajd-msys: Updated! Thank you @mwrock

lib/chef/resource/windows_certificate.rb
```diff
@@ -87,6 +87,10 @@ class WindowsCertificate < Chef::Resource
         description: "Ensure that sensitive resource data is not logged by the #{ChefUtils::Dist::Infra::CLIENT}.",
         default: lazy { pfx_password ? true : false }, skip_docs: true
 
+      property :exportable, [TrueClass, FalseClass],
```
tas50, line 90: This needs the introduced field so the documentation will be correctly built

dheerajd-msys: Added

dheerajd-msys:
Buildkite failure seems to be unrelated

PR #10723, opened 2020-12-09
Ensure we can still install RHEL 7 GCC on RHEL 6 in testing

tas50:
These are the terrible things you get to do to support an EOL platform.


lamont-granquist (approved):

tas50:
Merging since this gets us further

PR #10727, opened 2020-12-10
Update InSpec 4.24 in gemspec.

phiggins:

I wasn't sure if this was an intentional omission from #10726 or not.


phiggins:
Discussed in Slack, @tas50 confirmed it was intentional.

PR #10761, opened 2020-12-18
Stub http requests in rubygems tests.

phiggins:

Stub HTTP requests in tests that were hitting the live rubygems servers and timing out in some environments.

I changed the tests to use `sexp_processor` instead of `rspec` as the gem to check for because each dependency caused 2+ API requests, and `rspec` has a dozen or so transitive dependencies and `sexp_processor` was the first gem I found that has none.


spec/unit/provider/package/rubygems_spec.rb
```diff
@@ -113,28 +115,55 @@ def self.rspec_version_string
...
       Chef::Config[:rubygems_cache_enabled] = true
+
+      stub_request(:get, "https://rubygems.org/latest_specs.4.8.gz")
+        .to_return(status: 200, body: File.binread(File.join(CHEF_SPEC_DATA, "rubygems.org", "latest_specs.4.8.gz")))
     end
 
     it "finds a matching gem candidate version on rubygems 2.0.0+" do
-      dep = Gem::Dependency.new("rspec", ">= 0")
+      stub_request(:get, "https://rubygems.org/quick/Marshal.4.8/sexp_processor-4.15.1.gemspec.rz")
+        .to_return(status: 200, body: File.binread(File.join(CHEF_SPEC_DATA, "rubygems.org", "sexp_processor-4.15.1.gemspec.rz")))
```
phiggins, line 164: Stubbing this call was not necessary in my local env on macOS or in a windows VM, but the tests were blowing up on buildkite across all the jobs because this request was not stubbed. My hunch here is that the rubygems version being different between local and CI was the problem, but it did not seem worth the effort to track this down.

spec/unit/provider/package/rubygems_spec.rb
```diff
@@ -144,8 +173,11 @@ def self.rspec_version_string
     end
 
     it "finds a matching gem from a specific gemserver when explicit sources are given" do
-      dep = Gem::Dependency.new("rspec", ">= 0")
-      expect(@gem_env.candidate_version_from_remote(dep, "http://production.cf.rubygems.org")).to be_kind_of(Gem::Version)
+      stub_request(:get, "https://rubygems.org/quick/Marshal.4.8/sexp_processor-4.15.1.gemspec.rz")
+        .to_return(status: 200, body: File.binread(File.join(CHEF_SPEC_DATA, "rubygems.org", "sexp_processor-4.15.1.gemspec.rz")))
+
+      dep = Gem::Dependency.new("sexp_processor", ">= 0")
+      expect(@gem_env.candidate_version_from_remote(dep, "http://rubygems2.org")).to be_kind_of(Gem::Version)
```
phiggins, line 180: It appears that I can change this string argument to whatever I want and the requests still go to `https://rubygems.org`, so I suspect that means this configuration just does not work.

tas50 (approved):
Just fix the chefstyle bit and this is good to go.

Issue #10620, opened 2020-11-11
Chef 17: Remove Ubuntu 16.04 packages


tas50:
We can remove Ubuntu 16.04 packages after April 2021 when 16.04 goes EOL.

Edit https://github.com/chef/chef/blob/master/.expeditor/release.omnibus.yml to build on 18.04 and test on 18.04/20.04.

Run an ad-hoc job https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/ of the branch

PR #10765, opened 2020-12-19
Remove support for Ubuntu 16.04

tas50:
By the time we ship Chef 17 Ubuntu 16.04 will be EOL.


tas50:
Fixes https://github.com/chef/chef/issues/10620

Issue #10775, opened 2020-12-23
load_current_resource for systemd service could be more efficient


joshuamiller01:
### Describe the Enhancement:
Establishing the status of a systemd service (per `load_current_resource`) currently shells out to `systemctl` four times; this could be refactored to perform a single shellout for speed and CPU wins.  In my crude experimentation, a single `systemctl` call costs ~20ms; a single typical service resource has two actions ([:start, :enable]), each of which triggers the four shellouts.  This reduces the steady-state cost from ~160ms (2*4*20ms) to 40ms (~75% reduction).

### Describe the Need:
Anybody managing service resources on a systemd host

### Current Alternative
None

### Can We Help You Implement This?:
I'll submit a PR shortly.

PR #10776, opened 2020-12-23
load_current_resource for systemd service more efficiently

joshuamiller01:
Signed-off-by: Joshua Miller <joshmiller@fb.com>

## Description
Refactors how `load_current_resource` gathers state for systemd services by requesting all pieces of state in a single call and memoizing the results.

The only material change here is in how `masked-runtime` is handled; previously it was reported as 'masked', but I think that's incorrect.  Asserting whether a unit is masked should be against the persistent system state (not volatile).  The remaining service states are all the same.

Performance outcomes are demonstrated with a naive loop, against a system where no such services are running:
```
(1..100).each do |x|
  service "josh#{x}" do
    action [:stop, :disable]
  end
end
```

Previously this takes ~15s; with this patch, ~6s.  This testing was done with an older version of chef which has fewer shellouts, so the perf improvement should be better than this in newer chef releases.

## Related Issue
https://github.com/chef/chef/issues/10775

## Types of changes

## Checklist:

### DIFF
--- a/lib/chef/provider/service/systemd.rb
+++ b/lib/chef/provider/service/systemd.rb
@@ -76,6 +76,30 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
     end
   end
 
+  def systemd_service_status
+    @systemd_service_status ||= begin
+      # Collect all the status information for a service and returns it at once
+      options, args = get_systemctl_options_args
+      s = shell_out!(systemctl_path, args, "show", "-p", "UnitFileState", "-p", "ActiveState", new_resource.service_name, options)
+      # e.g. /bin/systemctl --system show  -p UnitFileState -p ActiveState sshd.service
+      # Returns something like:
+      # ActiveState=active
+      # UnitFileState=enabled
+      status = {}
+      s.stdout.each_line do |line|
+        k, v = line.strip.split("=")
+        status[k] = v
+      end
+
+      # Assert requisite keys exist
+      unless status.key?("UnitFileState") && status.key?("ActiveState")
+        raise Chef::Exceptions::Service, "'#{systemctl_path} show' not reporting status for #{new_resource.service_name}!"
+      end
+
+      status
+    end
+  end
+
   def get_systemctl_options_args
     if new_resource.user
       raise NotImplementedError, "#{new_resource} does not support the user property on a target_mode host (yet)" if Chef::Config.target_mode?
@@ -173,25 +197,25 @@ class Chef::Provider::Service::Systemd < Chef::Provider::Service::Simple
   end
 
   def is_active?
-    options, args = get_systemctl_options_args
-    shell_out(systemctl_path, args, "is-active", new_resource.service_name, "--quiet", **options).exitstatus == 0
+    # Note: "activating" is not active (as with type=notify or a oneshot)
+    systemd_service_status["ActiveState"] == "active"
   end
 
   def is_enabled?
-    options, args = get_systemctl_options_args
-    shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, "--quiet", **options).exitstatus == 0
+    # See https://github.com/systemd/systemd/blob/master/src/systemctl/systemctl-is-enabled.c
+    # Note: enabled-runtime is excluded because this is volatile, and the state of enabled-runtime
+    # specifically means that the service is not enabled
+    %w{enabled static generated alias indirect}.include?(systemd_service_status["UnitFileState"])
   end
 
   def is_indirect?
-    options, args = get_systemctl_options_args
-    s = shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, **options)
-    s.stdout.include?("indirect")
+    systemd_service_status["UnitFileState"] == "indirect"
   end
 
   def is_masked?
-    options, args = get_systemctl_options_args
-    s = shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, **options)
-    s.exitstatus != 0 && s.stdout.include?("masked")
+    # Note: masked-runtime is excluded, because runtime is volatile, and
+    # because masked-runtime is not masked.
+    systemd_service_status["UnitFileState"] == "masked"
   end
 
   private
--- a/spec/unit/provider/service/systemd_service_spec.rb
+++ b/spec/unit/provider/service/systemd_service_spec.rb
@@ -297,6 +297,55 @@ describe Chef::Provider::Service::Systemd do
         end
       end
 
+      enabled_and_active = <<-STDOUT
+        ActiveState=active
+        UnitFileState=enabled
+      STDOUT
+      disabled_and_inactive = <<-STDOUT
+        ActiveState=disabled
+        UnitFileState=inactive
+      STDOUT
+      # No unit known for this service, and inactive
+      nil_and_inactive = <<-STDOUT
+        ActiveState=inactive
+        UnitFileState=
+      STDOUT
+
+      def with_systemctl_show(systemctl_path, stdout)
+        systemctl_show = [systemctl_path, "--system", "show", "-p", "UnitFileState", "-p", "ActiveState", service_name]
+        expect(provider).to receive(:shell_out!).with(*systemctl_show, {}).and_return(double(stdout: stdout, exitstatus: 0, error?: false))
+      end
+
+      describe "systemd_service_status" do
+        before(:each) do
+          provider.current_resource = current_resource
+          current_resource.service_name(service_name)
+        end
+
+        it "should return status if '#{systemctl_path} --system show -p UnitFileState -p ActiveState service_name' returns 0 and has nil" do
+          nil_and_inactive_h = {
+            "ActiveState" => "inactive",
+            "UnitFileState" => nil,
+          }
+          with_systemctl_show(systemctl_path, nil_and_inactive)
+          expect(provider.systemd_service_status).to eql(nil_and_inactive_h)
+        end
+
+        it "should error if '#{systemctl_path} --system show -p UnitFileState -p ActiveState service_name' misses fields" do
+          partial_systemctl_stdout = <<-STDOUT
+            ActiveState=inactive
+          STDOUT
+          with_systemctl_show(systemctl_path, partial_systemctl_stdout)
+          expect { provider.systemd_service_status }.to raise_error(Chef::Exceptions::Service)
+        end
+
+        it "should error if '#{systemctl_path} --system show -p UnitFileState -p ActiveState service_name' returns non 0" do
+          systemctl_show = [systemctl_path, "--system", "show", "-p", "UnitFileState", "-p", "ActiveState", service_name]
+          allow(provider).to receive(:shell_out!).with(*systemctl_show, {}).and_return(shell_out_failure)
+          expect { provider.systemd_service_status }.to raise_error(Chef::Exceptions::Service)
+        end
+      end
+
       describe "is_active?" do
         before(:each) do
           provider.current_resource = current_resource
@@ -304,13 +353,22 @@ describe Chef::Provider::Service::Systemd do
           allow(provider).to receive(:which).with("systemctl").and_return(systemctl_path.to_s)
         end
 
-        it "should return true if '#{systemctl_path} --system is-active service_name' returns 0" do
-          expect(provider).to receive(:shell_out_compacted).with(systemctl_path, "--system", "is-active", service_name, "--quiet", timeout: 900).and_return(shell_out_success)
+        it "should return true if service is active" do
+          with_systemctl_show(systemctl_path, enabled_and_active)
           expect(provider.is_active?).to be true
         end
 
-        it "should return false if '#{systemctl_path} --system is-active service_name' returns anything except 0" do
-          expect(provider).to receive(:shell_out_compacted).with(systemctl_path, "--system", "is-active", service_name, "--quiet", timeout: 900).and_return(shell_out_failure)
+        it "should return false if service is not active" do
+          with_systemctl_show(systemctl_path, disabled_and_inactive)
+          expect(provider.is_active?).to be false
+        end
+
+        it "should return false if service is activating" do
+          enabled_and_activating = <<-STDOUT
+            ActiveState=activating
+            UnitFileState=enabled
+          STDOUT
+          with_systemctl_show(systemctl_path, enabled_and_activating)
           expect(provider.is_active?).to be false
         end
       end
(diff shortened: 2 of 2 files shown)


lib/chef/provider/service/systemd.rb
```diff
@@ -76,6 +76,31 @@ def define_resource_requirements
...
+      status[k] = v
+    end
+    # Assert requisite keys exist
+    unless status.key?("UnitFileState") && status.key?("ActiveState")
+      raise Mixlib::ShellOut::ShellCommandFailed, "'#{systemctl_path} show' not reporting status for #{new_resource.service_name}!"
+    end
+
+    @service_status = status
+    status
+  end
```
phiggins, line 102: The idiomatic ruby way to write this is memoized with a block:

```suggestion
  def systemd_service_status
    @systemd_service_status ||= begin
      # Collect all the status information for a service and returns it at once
      options, args = get_systemctl_options_args
      s = shell_out!("#{systemctl_path} #{args} show -p UnitFileState -p ActiveState #{new_resource.service_name}", options)
      # e.g. /bin/systemctl --system show  -p UnitFileState -p ActiveState sshd.service
      # Returns something like:
      # ActiveState=active
      # UnitFileState=enabled
      status = {}
      s.stdout.each_line do |line|
        k, v = line.strip.split("=")
        status[k] = v
      end
      # Assert requisite keys exist
      unless status.key?("UnitFileState") && status.key?("ActiveState")
        raise Mixlib::ShellOut::ShellCommandFailed, "'#{systemctl_path} show' not reporting status for #{new_resource.service_name}!"
      end
    end
  end
```

lib/chef/provider/service/systemd.rb
```diff
@@ -173,25 +198,29 @@ def unmask_service
...
   end
 
   def is_enabled?
-    options, args = get_systemctl_options_args
-    shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, "--quiet", **options).exitstatus == 0
+    status = systemd_service_status
+    # See https://github.com/systemd/systemd/blob/master/src/systemctl/systemctl-is-enabled.c
+    # Note: enabled-runtime is excluded because this is volatile, and the state of enabled-runtime
+    # specifically means that the service is not enabled
+    %w{enabled static generated alias indirect}.include?(status["UnitFileState"])
```
phiggins, line 211:
```suggestion
    # See https://github.com/systemd/systemd/blob/master/src/systemctl/systemctl-is-enabled.c
    # Note: enabled-runtime is excluded because this is volatile, and the state of enabled-runtime
    # specifically means that the service is not enabled
    %w{enabled static generated alias indirect}.include?(systemd_service_status["UnitFileState"])
```

lib/chef/provider/service/systemd.rb
```diff
@@ -173,25 +198,29 @@ def unmask_service
...
   end
 
   def is_indirect?
-    options, args = get_systemctl_options_args
-    s = shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, **options)
-    s.stdout.include?("indirect")
+    status = systemd_service_status
+    status["UnitFileState"] == "indirect"
```
phiggins, line 216:
```suggestion
    systemd_service_status["UnitFileState"] == "indirect"
```

lib/chef/provider/service/systemd.rb
```diff
@@ -173,25 +198,29 @@ def unmask_service
...
   end
 
   def is_masked?
-    options, args = get_systemctl_options_args
-    s = shell_out(systemctl_path, args, "is-enabled", new_resource.service_name, **options)
-    s.exitstatus != 0 && s.stdout.include?("masked")
+    # Note: masked-runtime is excluded, because runtime is volatile, and
+    # because masked-runtime is not masked.
+    status = systemd_service_status
+    status["UnitFileState"] == "masked"
```
phiggins, line 223:
```suggestion
    systemd_service_status["UnitFileState"] == "masked"
```

lib/chef/provider/service/systemd.rb
```diff
@@ -76,6 +76,31 @@ def define_resource_requirements
...
+    # ActiveState=active
+    # UnitFileState=enabled
+    status = {}
+    s.stdout.each_line do |line|
+      k, v = line.strip.split("=")
+      status[k] = v
+    end
+    # Assert requisite keys exist
+    unless status.key?("UnitFileState") && status.key?("ActiveState")
+      raise Mixlib::ShellOut::ShellCommandFailed, "'#{systemctl_path} show' not reporting status for #{new_resource.service_name}!"
```
phiggins, line 97: I don't think raising a Mixlib::Shellout exception here is the right thing to do. Using that exception class implies that the error came from Mixlib::Shellout which isn't the case. With no better option, a RuntimeException (ie, a bare call to `raise` is probably what you want here.

spec/unit/provider/service/systemd_service_spec.rb
```diff
@@ -297,20 +297,80 @@ def setup_current_resource
...
+      disabled_and_inactive = {
+        "UnitFileState" => "disabled",
+        "ActiveState" => "inactive",
+      }
+      # No unit known for this service, and inactive
+      nil_and_inactive = {
+        "ActiveState" => "inactive",
+        "UnitFileState" => nil,
+      }
+      describe "systemd_service_status" do
```
phiggins, line 313: The fact that this method exists is an implementation detail, no one outside of this class should care that it exists or be calling it, so explicitly testing it seems like the wrong thing to do here. 

A case could be made for keeping the tests for the error cases though.

lib/chef/provider/service/systemd.rb
```diff
@@ -173,25 +198,29 @@ def unmask_service
   end
 
   def is_active?
-    options, args = get_systemctl_options_args
-    shell_out(systemctl_path, args, "is-active", new_resource.service_name, "--quiet", **options).exitstatus == 0
+    status = systemd_service_status
+    # Note: "activating" is not active (as with type=notify or oneshots)
```
phiggins, line 202:
```suggestion
    # Note: "activating" is not active (as with type=notify or one-offs)
```

`oneshots` is not a dictionary word so this is failing our spellchecker.

lib/chef/provider/service/systemd.rb
joshuamiller01: My intent here is to surface that the shellout to systemctl failed; if `RuntimeException` is the preference, that works for me.

spec/unit/provider/service/systemd_service_spec.rb
joshuamiller01: The intent here is to be testing `systemd_service_status` and how it handles the various outputs and exit codes of `systemctl`; the other tests (e.g. `is_active?`) don't know anything about `systemctl` or its quirks - but that's by design.

lib/chef/provider/service/systemd.rb
joshuamiller01: 'oneshot' is specific systemd terminology.  Is there an incantation to tell the spellchecker to back off?

lib/chef/provider/service/systemd.rb
phiggins: @tas50 @lamont-granquist thoughts on this?

lib/chef/provider/service/systemd.rb
phiggins: Ahh, then in that case you'll need to add it to the list of exceptions in the `cspell.json` file: https://github.com/chef/chef/blob/master/cspell.json#L16

spec/unit/provider/service/systemd_service_spec.rb
phiggins: `systemd_service_status` is not the interface we care about, so it is not important if it changes as long as the methods that depend on it still work, since they are the interfaces we care about. The danger of the tests as they are written in this PR is that changing the implementation of `systemd_service_status` will require updating the tests of `is_active?` and friends, but that won't necessarily be apparent because they will continue to work thanks to the stubs hiding the actual changes.

I agree that all of `is_active?` and similar should not test every detail like the error conditions, those can remain as tests on `systemd_service_status`.

joshuamiller01: I think you're suggesting to do the following, correct?
```
  describe "is_active?" do
  ...
     it "should return true if service is active" do
       systemd_stdout = <<-STDOUT
          ActiveState=active
          UnitFileState=enabled
       STDOUT
       allow(provider).to receive(:shell_out!).with("#{systemctl_path} --system show -p UnitFileState -p ActiveState #{service_name}", {}).and_return(double(stdout: systemd_stdout, exitstatus: 0, error?: false))
       expect(provider.is_active?).to be true
    end
  end
  ```
  
  I can do this, but it'll mean each of the subsequent tests will be explicitly bound to the implementation of `systemd_service_status`; each test will be running it, and a change to the shellout will require a corresponding change to each test (which seems to be at odds with the concern you raised).

File: lib/chef/provider/service/systemd.rb [hunk 8]
```diff
@@ -76,6 +76,30 @@ def define_resource_requirements
     end
   end
 
+  def systemd_service_status
+    @service_status ||= begin
```
jaymzh, line 80: I worry a bit about the memoization here.

There should be one provider class instantiation per resource, I think, so in theory I don't think this will break:
```
service 'foo' do  # we check, see it's running
  action :stop
end
# something
service 'foo' do   # we need to check again, since it's now not running
  action :start
end
```
So I think that case is OK, but we probably want a test for it.

What about if we call this during `define_resource_requirements`? The wiring up of dynamic providers is a bit fuzzy to me, these days, but @lamont-granquist does that run during compile time? If so, we'll definitely break the entire world, that'd be like depending on Ohai to figure out if your service is running.

We also want a test around that case.

jaymzh (changes requested):

lib/chef/provider/service/systemd.rb
joshuamiller01: I didn't test cloned resources, but I did test multiple actions on a single resource, with the very funny:
```
service 'chronyd' do
  action [:start, :stop, :start, :stop, :start, :start, :start]
end
```

This still works; i.e. each `action` re-evaluates the state of the service, and idempotently fixes it when necessary.

spec/unit/provider/service/systemd_service_spec.rb
phiggins: You're right, these tests are going to be coupled to something because these are unit tests that test running a faked shell command, it's an issue of where the seams are. I don't like coupling the predicate methods so closely to the internal implementation, but I also don't like the scenario you pointed out of having to copy-paste the shell_out stub into each test. In that case, you could make the tests change just the output:

```
  def with_systemd_status(text)
    expect(provider).to receive(:shell_out!).with("#{systemctl_path} --system show -p UnitFileState -p ActiveState #{service_name}", {}).and_return(double(stdout: text, exitstatus: 0, error?: false))
  end

  describe "is_active?" do
  ...
     it "should return true if service is active" do
       with_systemd_status <<~STDOUT
          ActiveState=active
          UnitFileState=enabled
       STDOUT

       expect(provider.is_active?).to be true
    end
  end
```

That could be further encapsulated by reusing the output strings, etc.

Another thought I had is to reorient the tests so that they aren't grouped by method call but by status, so you could do something like:

```
context "when active and enabled" do
  before do
    with_systemd_status <<~STDOUT
      ActiveState=active
      UnitFileState=enabled
    STDOUT
  end
  it "is active" do
   expect(provider.is_active?).to eq(true)
  end

  it "is not masked" do
    expect(provider.is_masked?).to eq(false)
  end
end
```

This could make the tests harder to understand, it was just a stray thought I had.

joshuamiller01: I've fixed the tests to all mock the shellout with a function per your suggestion.

phiggins (approved):
Nice work, thanks for taking this on! 👍

lib/chef/provider/service/systemd.rb
jaymzh: I assume you mean tested manually yourself. But that's not why I want a test. I want a test so that in the future when someone innocently adds a call to this function from the wrong place and or we change up some wiring, we don't then break this code as a side effect.

You can probably add a similar test to the above in the `integration` test directory where you can just write a simple recipe. But one way or another, if we're gonna memoize stuff, we shouldn't merge this without a test to ensure we never have a stale memo.

jaymzh (changes requested):

lib/chef/provider/service/systemd.rb
lamont-granquist: the ivar and the memoization helper should be really named the same thing, so either service_status or systemd_service_status.

lib/chef/provider/service/systemd.rb
```diff
@@ -76,6 +76,30 @@ def define_resource_requirements
...
+      # Collect all the status information for a service and returns it at once
+      options, args = get_systemctl_options_args
+      s = shell_out!("#{systemctl_path} #{args} show -p UnitFileState -p ActiveState #{new_resource.service_name}", options)
```
lamont-granquist, line 83: you should use the *splat args style of calling here so:

```
shell_out!(systemctl_path,  args,  "show", "-p", "UnitFileState", "-p", "ActiveState", new_resource.service_name, options)
```

if that gets done correctly then it should avoid invoking the shell to parse the arguments, which should increase the perf even more (and has other benefits in not needing to worry about filenames with spaces and quotes around them, etc, etc).

lib/chef/provider/service/systemd.rb
lamont-granquist: so the memoization approach should be fine, every time an action is called on a resource it gets a new provider ("action class") with fresh state, so repeated resources or array actions will be fine.  even `run_action(:whatever)` on a saved resource will flush the provider.  we do a lot of memoization in other resources.

lib/chef/provider/service/systemd.rb
joshuamiller01: I'm going to change this to `raise Chef::Exceptions::Service` per Lamont's suggestion

lib/chef/provider/service/systemd.rb
joshuamiller01: @jaymzh - since every `action` gets a new provider (per Lamont and corroborated by my testing), I don't think we need a test to cover this case.  Browsing through the `integration` dir, we could mock out a resource running multiple actions, but since each provider is a separate instance, that wouldn't be testing the memoization in any way.

jaymzh (approved):
thanks for the double-check @lamont-granquist 
Nice work @joshuamiller01

lamont-granquist (approved):

tas50:
This appears to have broken things on Ubuntu 16.04, but since that's EOLing before Chef 17 we're probably fine there. Let's just make sure not to backport this to Chef 16.

ramereth:
To be clear, it seems to have only broke idempotency with the enable action and only on Ubuntu 16.04.

rpunt:
@ramereth sorry, not true. I can demonstrate on RHEL7. Please see #11980

PR #10786, opened 2020-12-29
Stop updating bundler in CI

tas50:
We don't update it in our builds anymore. We just use the version that's built into Ruby. This omnibus override was only used for CI and we had to match the versions, which was a real pain and pretty pointless.

### DIFF
--- a/.expeditor/scripts/bk_linux_exec.sh
+++ b/.expeditor/scripts/bk_linux_exec.sh
@@ -32,9 +32,9 @@ echo "--- Config information"
 echo "!!!! RUBY VERSION !!!!"
 ruby --version
 echo "!!!! BUNDLER LOCATION !!!!"
-which bundler
+which bundle
 echo "!!!! BUNDLER VERSION !!!!"
-bundler -v
+bundle -v
 echo "!!!! DOCKER VERSION !!!!"
 docker version
 echo "!!!! DOCKER STATUS !!!!"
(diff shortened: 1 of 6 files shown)


.expeditor/scripts/bk_linux_exec.sh
```diff
@@ -31,8 +31,10 @@ echo "--- Config information"
 
 echo "!!!! RUBY VERSION !!!!"
 ruby --version
-echo "!!!! BUNDLE LOCATION !!!!"
-which bundle
+echo "!!!! BUNDLER LOCATION !!!!"
+which bundler
+echo "!!!! BUNDLER VERSION !!!!"
+bundler -v
```
lamont-granquist, line 37: `bundle -v`?

omnibus_overrides.rb
```diff
@@ -4,7 +4,6 @@
 # NOTE: You MUST update omnibus-software when adding new versions of
 # software here: bundle exec rake dependencies:update_omnibus_gemfile_lock
 override :rubygems, version: "3.1.4" # pin to what ships in the ruby version
-override :bundler, version: "2.1.4" # pin to what ships in the ruby version
```
lamont-granquist, line 7: don't we still pull in the bundler omnibus-software definition from the ruby definition which will then install whatever its default version is?

tas50: We had to nuke all that junk when ruby integrated it. It causes issues when we need to build stuff for old ruby releases.

.expeditor/scripts/bk_linux_exec.sh
tas50: it's an alias, but I updated the 3 places I used bundler not bundle

lamont-granquist (approved):

PR #10794, opened 2020-12-29
Consolidate Windows cert tests to improve CI runtime

### LINKED ISSUE #10696: Determine if we can speed up the 25 minute windows certificate tests
By far the worst offender in our Windows tests is the windows_certificate testing we perform. The tests in spec/functional/resource/windows_certificate_spec.rb take ~25 minutes to run in CI. Anything we can do there would be a huge win for the overall test times.

phiggins:

Attempt to fix #10696.

These tests were identified as running slowly on CI. This PR takes their runtime from ~50s to ~18s in my local Windows 2019 VM.

### DIFF
--- a/lib/chef/resource/windows_certificate.rb
+++ b/lib/chef/resource/windows_certificate.rb
@@ -309,11 +309,7 @@ class Chef
         # @raise [OpenSSL::PKCS12::PKCS12Error] When incorrect password is provided for PFX certificate
         #
         def fetch_cert_object(ext)
-          contents = if binary_cert?
-                       ::File.binread(new_resource.source)
-                     else
-                       ::File.read(new_resource.source)
-                     end
+          contents = ::File.binread(new_resource.source)
 
           case ext
           when ".pfx"
@@ -328,12 +324,6 @@ class Chef
           else
             OpenSSL::X509::Certificate.new(contents)
           end
-        end
-
-        # @return [Boolean] Whether the certificate file is binary encoded or not
-        #
-        def binary_cert?
-          shell_out!("file -b --mime-encoding #{new_resource.source}").stdout.strip == "binary"
         end
 
         # Imports the certificate object into cert store
--- a/spec/functional/resource/windows_certificate_spec.rb
+++ b/spec/functional/resource/windows_certificate_spec.rb
@@ -19,478 +19,298 @@ require "spec_helper"
 require "chef/mixin/powershell_exec"
 require "chef/resource/windows_certificate"
 
-module WindowsCertificateHelper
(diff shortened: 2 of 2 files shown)


lib/chef/resource/windows_certificate.rb
```diff
@@ -309,11 +309,7 @@ def acl_script(hash)
         # @raise [OpenSSL::PKCS12::PKCS12Error] When incorrect password is provided for PFX certificate
         #
         def fetch_cert_object(ext)
-          contents = if binary_cert?
-                       ::File.binread(new_resource.source)
-                     else
-                       ::File.read(new_resource.source)
-                     end
+          contents = ::File.binread(new_resource.source)
```
phiggins, line 312: Reading non-binary certs with `File.binread` appears to cause no effect, so avoid shelling out to find out if the file is binary. This is already done in other places (#10751), so it seems safe to do here as well.

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -19,478 +19,298 @@
...
-    # Seems weird that we have to call dir twice right?
-    # The powershell pki module cache the last dir in module session state
-    # Issuing dir with a different arg (-Force) seems to refresh that state.
-    command = <<~EOC
-      dir #{path} -Force | Out-Null
-      (dir #{path} | measure).Count
+  def certificate_count
+    powershell_exec(<<~EOC).result.to_i
+      (Get-ChildItem -Force -Path Cert:\\LocalMachine\\#{store} | measure).Count
     EOC
```
phiggins, line 40: Using the `-Force` flag resulted in getting the correct results in my local testing, avoiding the need for the double command and confusing comment (`dir` and `Get-ChildItem` are aliases in powershell).

spec/functional/resource/windows_certificate_spec.rb
```diff
@@ -19,478 +19,298 @@
...
-             "CA", "AUTHROOT", "TRUSTEDPEOPLE", "MY", "SMARTCARDROOT", "TRUST",
-             "DISALLOWED"]
-
-    # Byepassing the validation so that we may create a custom store
+    # Bypass validation of the store name so we can use a fake test store.
     allow_any_instance_of(Chef::Mixin::ParamsValidate)
       .to receive(:_pv_equal_to)
-      .with(opts, key, to_be)
+      .with({ store_name: store }, :store_name, anything)
       .and_return(true)
```
phiggins, line 68: There did not seem to be a better way to allow these tests to use a custom cert folder other than stubbing this.

PR #10797, opened 2020-12-30
Coerce uid and gid to integer to prevent failures in Windows.

### LINKED ISSUE #10454: `no implicit conversion of FFI::MemoryPointer into Integer` when you specify `uid`
This seems a lot like this bug from forever ago: https://github.com/chef/chef/issues/4024

## Description
the `user` resource on windows blows up with `no implicit conversion of FFI::MemoryPointer into Integer` if you specify `uid`.

If you remove `uid`, then it works fine.

## Chef Version
15.9.7

## Platform Version
Windows 10

## Replication Case

```
user 'dude' do
  uid "8019"
end
```

And it will blow up.

## Client Output

```
TypeError
---------
no implicit conversion of FFI::MemoryPointer into Integer

Resource Declaration:
---------------------
# In C:/chef/cache/cookbooks/fb_users/resources/default.rb

109:     user username do
110:       uid mapinfo['uid']
111:       # the .to_i here is important - if the usermap accidentally
112:       # quotes the gid, then it will try to look up a group named "142"
113:       # or whatever.
114:       #
115:       # We explicityly pass in a GID here instead of a name to ensure that      
116:       # as GIDs are moving, we get the intended outcome.
117:       unless node.windows?
118:         gid ::FB::Users::GID_MAP[pgroup]['gid'].to_i
119:       shell info['shell'] || node['fb_users']['user_defaults']['shell']
120:       manage_home manage_homedir
121:       home homedir
122:       comment mapinfo['comment'] if mapinfo['comment']
123:       password pass if pass
124:       end
125:       action :create

Compiled Resource:
------------------
# Declared in C:/chef/cache/cookbooks/fb_users/resources/default.rb:109:in `block (2 
levels) in class_from_file'

windows_user("someuser") do
  action [:create]
  declared_type :user
  cookbook_name "fb_users"
  uid "8019"
end
```

phiggins:

Fixes #10454.

There doesn't appear to be a valid usecase for a uid or gid that is not an integer string, ie `uid "pete"`. 

Running the test script in the linked issue after this PR:

```
PS C:\vagrant\chef> ruby -Ilib chef-bin\bin\chef-apply .\test_user_uid.rb
  * windows_user[dude] action create
    - alter user dude
    - change uid from 1001 to 8019
```

### DIFF
--- a/lib/chef/resource/user.rb
+++ b/lib/chef/resource/user.rb
@@ -67,10 +67,12 @@ class Chef
         default: false
 
       property :uid, [ String, Integer, NilClass ], # nil for backwards compat
-        description: "The numeric user identifier."
+        description: "The numeric user identifier.",
+        coerce: proc { |n| n && Integer(n) }
 
       property :gid, [ String, Integer, NilClass ], # nil for backwards compat
-        description: "The numeric group identifier."
+        description: "The numeric group identifier.",
+        coerce: proc { |n| n && Integer(n) }
 
       alias_method :group, :gid
     end
--- a/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/windows.rb
@@ -126,3 +126,11 @@ include_recipe "::_ohai_hint"
 hostname "new-hostname" do
   windows_reboot false
 end
+
+user "phil" do
+  uid "8019"
+end
+
+user "phil" do
+  action :remove
+end
--- a/spec/unit/resource/user_spec.rb
+++ b/spec/unit/resource/user_spec.rb
@@ -85,7 +85,7 @@ end
 
     it "allows a string" do
       resource.send(attrib, "100")
-      expect(resource.send(attrib)).to eql("100")
+      expect(resource.send(attrib)).to eql(100)
     end
 
     it "allows an integer" do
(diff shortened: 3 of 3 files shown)


lib/chef/resource/user.rb
```diff
@@ -67,10 +67,12 @@ class User < Chef::Resource
...
 
       property :uid, [ String, Integer, NilClass ], # nil for backwards compat
-        description: "The numeric user identifier."
+        description: "The numeric user identifier.",
+        coerce: proc { |n| n && Integer(n) }
 
       property :gid, [ String, Integer, NilClass ], # nil for backwards compat
-        description: "The numeric group identifier."
+        description: "The numeric group identifier.",
+        coerce: proc { |n| n && Integer(n) }
```
phiggins, line 75: This allows setting uid and gid to nil or something coercible to Integer. I could preserve the existing error behavior by changing this to `coerce: proc { |n| n && Integer(n) rescue n }`, but there are many other places in resources where a `coerce: proc { |n| Integer(n) }` is used.

phiggins:
Based on the test failures I made some wrong assumptions in here, will revisit.

phiggins:
I do have a fix for this, but I can't push to this branch anymore. I'll open a new one. 😅

PR #10813, opened 2021-01-01
adapt to FreeBSD pkgng sysexit changes

mrtazz:
## Description

The exit code for pkgng changed with https://github.com/freebsd/pkg/commit/f446c790707993656934e0c378b64ae28cb96575 from `70` to `1` making the provider fail on any non-converged node because `pkg info <pkg>` raises an error now. E.g.

```
    Error executing action `install` on resource 'freebsd_package[tmux]'

    Expected process to exit with [0, 70], but received '1'
    ---- Begin output of ["pkg", "info", "tmux"] ----
    STDOUT:
    STDERR: pkg: No package(s) matching tmux
    ---- End output of ["pkg", "info", "tmux"] ----
    Ran ["pkg", "info", "tmux"] returned 1
```

I'm a bit wary of how this could be masking legitimate errors of the `pkg info` command.  So I'd definitely love some input on whether there should be additional checking going on to make sure it's not an exit code that should `raise`. That's why I also marked "Breaking Change" below.

This breaks on any FreeBSD node with pkgng version `1.15.99.7` or newer. The current version being `1.16.1`.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I have added tests to cover my changes.


tas50:
@mrtazz Can you rebase this against the current master. Also should we continue to allow 70 in the event that the user is on an older release?

mrtazz:
> Can you rebase this against the current master.

@tas50 yup, will do.

> Also should we continue to allow 70 in the event that the user is on an older release?

Good call. Looking at https://www.freshports.org/ports-mgmt/pkg/ it seems that a couple of architectures still only have the slightly older pkg version with the old behavior. So if that makes sense from a maintainer perspective I'll add `70` as an expected exit code back in.

tas50:
Yeah let's leave 70 and throw a comment above each of these executes that summarizes why we need to support those exit codes. Something that will allow us to circle back years from now and know that it's safe to remove 70.

mrtazz:
updated according to feedback. The force push was because I forgot to sign off on the commit

tas50:
Thanks @mrtazz

PR #10816, opened 2021-01-04
Update to Ruby 3.0.0

jasonwbarnett:
This should improve performance across the board 🤞


tas50:
We have a good number of blockers to work out before we can do this one. This PR attempts to add testing to CI which tests out how the app performs on Ruby 3. We have issues right now with ruby-shadow, inspec (fixed, but not yet merged to chef/chef) and ffi. It's happening, but there's a TON of little things to handle first.

PR #10823, opened 2021-01-05
YUM package idempotency fixes

lamont-granquist:
This applies the same kinds of fixes in #9763 that were applied to the `dnf_package` provider to the `yum_package` provider.

The following ways of calling `yum_package` with a single name argument have been fixed and are now supported:

```ruby
yum_package "ypbind-3:1.37.1"
yum_package "ypbind-3:1.37.1-9.el7"
```

The following ways of calling `yum_package` with a single name argument are still supported:

```ruby
yum_package "ypbind-1.37.1"
yum_package "ypbind-1.37*"
yum_package "ypbind-1.37.1-*"
yum_package "ypbind-1.37*-*"
yum_package "ypbind-*:1.37.1-*"
yum_package "ypbind-*:1.37*-*"
yum_package "ypbind-1.37.1-9.el7"
yum_package "ypbind-1.37*-9.el7"
yum_package "ypbind-1.37.1-9.*"
```

The following ways of calling `yum_package` with a single named argument do not properly match candidate versions (due to support in the underlying yum libraries) and will fail:

```ruby
yum_package "ypbind-3:1.37*-9.el7"
yum_package "ypbind-3:1.37.*"
yum_package "ypbind-3:1.37.1-9.*"
```

The following ways of calling `yum_package` with a version property have been fixed and are now supported:

```ruby
yum_package "ypbind" do
  version "1.37*"
end
yum_package "ypbind" do
  version "1.37.1-*"
end
yum_package "ypbind" do
  version "1.37*-*"
end
yum_package "ypbind" do
  version "3:1.37.1-9.el7"
end
yum_package "ypbind" do
  version "*:1.37.1-*"
end
yum_package "ypbind" do
  version "*:1.37*-*"
end
yum_package "ypbind" do
  version "1.37.1-9.el7"
end
yum_package "ypbind" do
  version "1.37*-9.el7"
end
yum_package "ypbind" do
  version "1.37.1-9.*"
end
```

The following ways of calling` yum_package` with a version property were not idempotent and that has been fixed (there was previously no way to correctly use the version property of the `yum_package` resource and have it behave idempotently):

```ruby
yum_package "ypbind" do
  version "1.37.1"
end
yum_package "ypbind" do
  version "3:1.37.1"
end
```

The following ways of calling `yum_package` with a version property do not properly match candidate versions (due to support in the underlying yum libraries) and will fail:

```ruby
yum_package "ypbind" do
  version "3:1.37*-9.el7"
end
yum_package "ypbind" do
  version "3:1.37.*"
end
yum_package "ypbind" do
  version "3:1.37.1-9.*"
end
```

Users are highly encouraged to always provide the full EVR (epoch-version-release) of an RPM as the chef version since omitting the epoch is ambiguous and rpm/yum/dnf libraries tend to assume zero when omitted.  Event though the version with the epoch and release omitted is accepted by the underlying yum libraries (`1.37.1`)  it is highly recommended to never rely on being able to omit the epoch.  Omitting the release is also not a best practice, but may be required due to the convention of embedding the distro string in the release.

The same fixes apply to the `:upgrade` action.  The `:remove` action may have been similarly fixed and many additional tests have been added to the test suite around idempotency and correctness.


lib/chef/provider/package/yum.rb
```diff
@@ -64,6 +65,16 @@ def load_current_resource
           current_resource
         end
 
+        def load_after_resource
+          # force the installed version array to repopulate
+          @current_version = []
+          @after_resource = Chef::Resource::DnfPackage.new(new_resource.name)
```
jaymzh, line 71: I don't think you meant Dnf. Yum.

lib/chef/provider/package/yum/python_helper.rb
```diff
@@ -115,17 +115,70 @@ def options_params(options)
             end
           end
 
+          def is_arch?(arch)
+            arches = %w{alpha alphaev4 alphaev45 alphaev5 alphaev56 alphaev6 alphaev67 alphaev68 alphaev7 alphapca56 amd64 armv5tejl armv5tel armv6l armv7l athlon geode i386 i486 i586 i686 ia32e ia64 noarch ppc ppc64 ppc64iseries ppc64pseries s390 s390x sh3 sh4 sh4a sparc sparc64 sparc64v sparcv8 sparcv9 sparcv9v x86_64}
```
tas50, line 119: 1/2 these need to get shoved into https://github.com/chef/chef/blob/main/cspell.json#L16

lamont-granquist: i'm trying to go the other way and disable spellchecking on that line, its useless to have it

lib/chef/provider/package/yum/python_helper.rb
```diff
@@ -175,9 +229,9 @@ def build_query(action, parameters)
             end
 
             # Special handling for certain action / param combos
-            if %i{whatinstalled whatavailable}.include?(action)
-              add_version(hash, parameters["version"]) unless parameters["version"].nil?
-            end
+            # if %i{whatinstalled whatavailable}.include?(action)
+            #  add_version(hash, parameters["version"]) unless parameters["version"].nil?
+            # end
```
lamont-granquist, line 234: why did i do this?  why does this not break anything?

tas50:
@jaymzh I assume you have thoughts on this

jaymzh (approved):
Nice.

PR #10824, opened 2021-01-05
Update Ohai to 16.8.2 and pin chefstyle

tas50:
This ohai should fix build failures and we don't need to pull chefstyle from git anymore since we only want the latest coming in on chef/chef master.


jaymzh (approved):

tas50:
This looks to have fixed the failing TK tests, but some jobs still failed due to the docker rate limiting.

PR #10836, opened 2021-01-05
Extend the pure ruby Ohai timeout to 4 seconds

tas50:
This times out from time to time when the system is configured
correctly. It just makes the test cycles take an hour longer.


jaymzh (approved):

tas50:
I think we can hold off on this one for now, but it sure is annoying when s390 or aix times out

Issue #10345, opened 2020-08-25
Add lazy node values to support delayed evaluation in attributes files and derived values


lamont-granquist:
https://github.com/poise/poise-derived

Dunno why I never noticed this cookbook before but this is the least bad solution I've seen to this problem.

tas50:
Done in 17

PR #10861, opened 2021-01-12
Add support for lazy attributes

lamont-granquist:
Implements #10345

This is very much not the poise-derived solution.

This allows the user to store procs/lazy-evaluators into the attribute tree.  Those are saved into the merged view of the node object.  On access, the proc is evaluated before the value is given back to the user.

This allows the user to lazy derived attributes:

```ruby
default['user'] = "Avasarala"
default['homedir'] = lazy { "/home/#{node['user']}" }
```

Since the proc is saved in the deep merge cache rather than the evaluated value, the result is truly lazy:

```ruby
default['user'] = "Avasarala"
default['homedir'] = lazy { "/home/#{node['user']}" }

# prints "/home/Avasarala"
puts node['homedir']

override['user'] = "Amos"

# prints "/home/Amos"
puts node['homedir']
```

The user can arbitrarily mess with the dependent variable in recipe code and the lazied attribute should change accordingly.

The access through the precedence level is still the proc (this is pretty much a feature not a bug and trying to change this behavior would cache the lazy evaluation on first access):

```ruby
default['user'] = "Holden"
default['homedir'] = lazy { "/home/#{node['user']}" }

# prints something about being a proc defined on some source line
puts default['homedir'] 
# prints "/home/Holden"
puts default['homedir'].call 
```

It also supports recursive laziness

```ruby
default["program"] = "ganymede"
default["program_var"] = lazy { "/var/#{node["program"}" }
default["program_cache"] = lazy { "#{node["program_var"]}/cache" }

default["program"] = "phoebe"

# prints "/var/phoebe/cache"
puts node["program_cache"] 
```

This will not work with poise-derived, users will get poise-derived behavior (which I suspect has been broken for several major releases of chef-client now)

Note that since it returns a real value on access, that means it does not lazy the value all the way to converge time in the recipe context, it will still be necessary to lazy the access in recipe code as well.  There is technically a breaking change issue here, because users may have figured out that they can already pass a lazy proc into an attribute and pass it to a property which will lazy the evaluation (and lazy it all the way to converge time).  This change enhances the former behavior -- any user that was trying to work around the derived attribute issue will have found that it fails when passed into e.g. template variables without littering their templates with ".call" modifiers and such, which this PR fixes.  However it necessarily breaks lazy procs in attributes from being lazied all the way to converge time, which uses may have abused as a feature, or may have just accidentally become dependent upon.  This is actually better behavior since having the contents of the attribute control the evaluation at converge time is going to surprise people who overwrite the attribute and lose the behavior as well.  There is no good way to deprecate this behavior since it requires very, very spooky action at a distance (the evaluation of the proc by the property does not have any information about the proc being declared immediately in the recipe code or distantly from the attributes file, and a lazy proc in the attributes file might not require the behavior of lazying to compile time so some very distant coordination would be required to accurately emit a deprecation -- and doing this with cookstyle would be impossible).  At the same time the amount of cookbooks using this trick that will break should be very, very low.

It looks like there's only two groups of cookbooks on the supermarket that have discovered this trick:

```
alfresco-appserver/attributes/default.rb
alfresco-appserver/attributes/tomcat.rb
alfresco-db/attributes/mysql_local.rb
alfresco-transformations/attributes/default.rb
kafka/attributes/default.rb
kkafka/attributes/default.rb
```

Those seem to be using `lazy {}` in order to do derived attributes so presumably would still work okay, but no other cookbooks should be affected.

BONUS FIX:

The deep merge code was calling ImmutableMash#[] while merging arrays which was incurring additional expense and complexity in invoking all the decorated state_tracking stuff and other calls in the inheritance heirarchy on those.  Similarly to how we use the Mash#internal_set API in that routine to set values without going through convert_value this introduces a change to use the new Mash#internal_get API which directly gets the value from the underlying Hash without any magic associated with it.  This should provide at least a little bit of perf boost to the deep merge caching routines, and it eliminates an infinite loop that I created here.

DOUBLE-SUPER-BONUS FIX:

The Chef::DelayedEvaluator class seems to exercise a ruby bug where dup returns a Proc rather than the subclass, so fix that.

### DIFF
--- a/chef-utils/lib/chef-utils/mash.rb
+++ b/chef-utils/lib/chef-utils/mash.rb
@@ -94,12 +94,21 @@ module ChefUtils
       end
     end
 
+    unless method_defined?(:regular_reader)
+      alias_method :regular_reader, :[]
+    end
+
     unless method_defined?(:regular_writer)
       alias_method :regular_writer, :[]=
     end
 
     unless method_defined?(:regular_update)
       alias_method :regular_update, :update
+    end
+
+    # @param key<Object> The key to get.
+    def [](key)
+      regular_reader(key)
     end
 
     # @param key<Object> The key to set.
@@ -110,6 +119,12 @@ module ChefUtils
     # @see Mash#convert_value
     def []=(key, value)
       regular_writer(convert_key(key), convert_value(value))
+    end
+
+    # internal API for use by Chef's deep merge cache
+    # @api private
+    def internal_get(key)
+      regular_reader(key)
     end
 
     # internal API for use by Chef's deep merge cache
--- a/lib/chef/delayed_evaluator.rb
+++ b/lib/chef/delayed_evaluator.rb
@@ -17,5 +17,9 @@
 
 class Chef
   class DelayedEvaluator < Proc
+    def dup
+      # super returns a "Proc" (which seems buggy) so re-wrap it
+      self.class.new(&super) # rubocop:disable Layout/SpaceAroundKeyword
+    end
   end
 end
--- a/lib/chef/node/attribute.rb
+++ b/lib/chef/node/attribute.rb
@@ -596,7 +596,7 @@ class Chef
           merge_with.each do |key, merge_with_value|
             value =
               if merge_onto.key?(key)
-                deep_merge!(safe_dup(merge_onto[key]), merge_with_value)
+                deep_merge!(safe_dup(merge_onto.internal_get(key)), merge_with_value)
               else
                 merge_with_value
               end
@@ -632,7 +632,7 @@ class Chef
           merge_with.each do |key, merge_with_value|
             value =
               if merge_onto.key?(key)
-                hash_only_merge!(safe_dup(merge_onto[key]), merge_with_value)
+                hash_only_merge!(safe_dup(merge_onto.internal_get(key)), merge_with_value)
               else
                 merge_with_value
               end
--- a/lib/chef/node/immutable_collections.rb
+++ b/lib/chef/node/immutable_collections.rb
@@ -19,6 +19,7 @@ require_relative "common_api"
 require_relative "mixin/state_tracking"
 require_relative "mixin/immutablize_array"
 require_relative "mixin/immutablize_hash"
+require_relative "../delayed_evaluator"
 
 class Chef
   class Node
@@ -111,7 +112,7 @@ class Chef
 
       def [](*args)
         value = super
-        value = value.call while value.is_a?(Proc)
+        value = value.call while value.is_a?(::Chef::DelayedEvaluator)
         value
       end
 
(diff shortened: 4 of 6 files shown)


jaymzh:
Just to be clear, does the proc converge on first-read, or will it be re-evaluated on every read? In other words if I do:

```
file 'whatever' do
  owner node['user']
end

file 'whatever2' do
  owner lazy { node['user'] }
end
```

and assuming `user` is lazy, per your example, does the second `file` get screwed by the value already being resolved in the first `file`?

Your description implies so, but that doesn't seem super bad. I'm not familiar enough with the code to know for sure at a quick glance without paging in the attribute code.

lamont-granquist:
Yeah that's the second example, it is truly lazy on every read rather than caching the first evaluated return value.

```
node['user'] = lazy { node['real_user'] }

node['real_user'] = "phil"

file 'whatever' do
  owner node['user'] # this would have magically been "lazied" to converge time before this change and would be "dibowitz", but now is "phil"
end

file 'whatever2' do
  owner lazy { node['user'] } # this will be "dibowitz" 
end

ruby_block "print it" do
 block do
   puts node['user'] # this will now print "dibowitz" (before it would print out something about it being a Proc)
end

node['real_user'] = "dibowitz"
```

Note that in the new behavior only the second property is evaluated lazily at converge time.

In the old behavior `node['user']` was implicitly lazied all the way to converge time because the attribute carried a lazy block which was presented directly to the property.

In the new behavior the laziness is more directly exactly what is written and the first property will not have a lazy value.  When ruby parses the attribute at compile time the lazy proc will be evaluated and a string will be presented into the property.

This seems more correct to me, since if you want lazy properties at converge time you should have `lazy {}` around the value in the property and relying on a lazy proc being shoved into the node attribute is SUPER confusing spooky-action-at-a-distance and overwriting it with `node['user'] = "phil"` will unlazy the use of the attribute and pass the string to the first file property at compile time.  So breaking that seemed okay.

The rule becomes simple "if you want compile time laziness for a property, you need to write a lazy {} in the property" then you don't have data-driven modification of compile-converge laziness.

jaymzh:
OK that makes sense to me.

If I use `proc` instead of `lazy`, does that change things? We have a bunch of latterns where we do `node.default['foo'] = proc { bla }` and then we want to eval that proc at a specific time, are we going to try to read node['foo'] and get it eval'd _for_ us? Or is this *only* for "lazy"?

lamont-granquist:
oh interesting.  no its not just for lazy, and it does procs as well.

that would be easily fixable though.  in fact it sounds like i should fix that.

lamont-granquist:
```
[4] pry(main)> a = Chef::DelayedEvaluator.new { puts "foo" }
=> #<Chef::DelayedEvaluator:0x00007ffe0c03a4c8 (pry):4>
[5] pry(main)> a.dup
=> #<Proc:0x00007ffe0c058e28 (pry):4>
```

ruby just trolling me now.

lamont-granquist:
Okay, fixed.

lamont-granquist:
Ruby bug report also submitted: https://bugs.ruby-lang.org/issues/17545

PR #10880, opened 2021-01-14
Add support for client.d files in chef-shell

### LINKED ISSUE #10748: chef-shell doesn't read client.d/*
## Description

chef-shell doesn't read client.d/*

## Chef Version
16.7.61

## Platform Version
Ubuntu 18.04

## Replication Case

put `raise 'this is a test'` in `/etc/chef/client.d/test.rb`
then run `chef-shell -z`

which it not die on that.

jaymzh:
This brings chef-shell a bit closer to behaving properly. This code is
tons of copy-pasta and needs some love. Because it's not using *any* of
`Chef::Application` (not `Application` or `Application::Base` or even
it's own fork of `Application` like half the other stuff in the repo),
it doesn't have access to _any_ of the many versions of
`load_config_file` (note to self: collapse those into one).

Pulling in Application will be a significant chunk of work, so in the
mean time I did the minimal amount of work to make the `parse_opts` in
shell.rb create the right bits so that Chef::Config and Mixin::DotD can
work... so while this doesn't fix the problem that we're invoking our
own config parsing, it at least uses as much of the common code as I can
without reworking the entirety of chef-shell.

Here's an example of it working:

```
root@ldt-hardwired:~# cinc-shell -s -c /etc/chef/client.rb
loading configuration: /etc/chef/client.rb
true
Session type: solo
```

From there it loads not only `/et/chef/client.rb` but also
`/etc/chef/client.d/ohai.rb`. Neat.

None of this portion of the code is tested in the specs and frankly, I'm
not entirely certainly how to go about writing a test, but I'll poke
around.

fixes #10748

### DIFF
--- a/lib/chef/shell.rb
+++ b/lib/chef/shell.rb
@@ -25,6 +25,7 @@ require "pp" unless defined?(PP)
 require "etc" unless defined?(Etc)
 require "mixlib/cli" unless defined?(Mixlib::CLI)
 require "chef-utils/dist" unless defined?(ChefUtils::Dist)
+require "chef-config/mixin/dot_d"
 
 require_relative "../chef"
 require_relative "version"
@@ -211,6 +212,7 @@ module Shell
 
   class Options
     include Mixlib::CLI
+    include ChefConfig::Mixin::DotD
 
     def self.footer(text = nil)
       @footer = text if text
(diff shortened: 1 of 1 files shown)


lamont-granquist (approved):

lamont-granquist:
i would honestly not bother with specs for this unless you're feeling particularly masochistic

jaymzh:
The tests all seem to be failing because of gem version constraint conflicts...

jaymzh:
So all the unit tests don't even get past:
```
ohai-17.0.7 requires ruby version >= 2.7, which is incompatible with the current version, ruby 2.6.6p146
```

And all the kitchen tests are failing on:
```
The git source https://github.com/chef/ohai.git is not yet checked out. Please run `bundle install` before trying to start your application
```

PR #10887, opened 2021-01-15
Split out the rubocop runs so we can bump to 2.7

tas50:


tas50:
Due to the way rubocop inherits config this is not something we should do. The exclude of chef-config in the main config is inherited to the chef-config .rubocop.yml and nothing is scanned. You can disable that on the CLI, but not in the config.

PR #10914, opened 2021-01-21
Backport of changes necessary for removal of dev gems

lamont-granquist:

PR #10919, opened 2021-01-22
backport some chef-16 code to chef-15 to fix builds

lamont-granquist:
This is necessary for new omnibus-software.

This is not really similar to #10914 which is related to
chef-utils stuff which doesn't exist in chef-15.

There are other fixes I pushed into chef-16 at some point
to fix running rake install outside of bundler which we need
here and this backports a bit of other cleanup that exists
in chef-16/17 in the Rakefiles, making them more similar


Rakefile
```diff
@@ -82,15 +88,3 @@ begin
 rescue LoadError
   puts "chefstyle/rubocop is not available. bundle install first to make sure all dependencies are installed."
 end
-
-begin
```
tas50, line 86: 🍻

PR #10922, opened 2021-01-22
Ruby 3.0 fixes and post-bundle-install hook 

lamont-granquist:
Mostly this is all fixes necessary for ruby 3.0

There's the addition of the appbundle hook which lets us better pull
git gems into appbundler

Note carefully how after adding the post-bundle-install.rb that
trying to pre appbundle-update ohai pulls in chef/chef as bundle
installed git gem which fails to install so we go back to only
using one appbundle-update on chef/chef and removing the chef/ohai
one (which may fix other bugs).


.expeditor/verify.pipeline.yml
```diff
@@ -12,10 +12,10 @@ expeditor:
 steps:
 
 #########################################################################
-  # Tests Ruby 2.6
+  # Tests Ruby 2.7
```
tas50, line 15: chef-utils and chef-config were still on ruby 2.6 by design. These tests need to stay

jaymzh (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FxT9DPrC1YWXBO1coyQ%2Fgiphy.gif&token=lgtm" data-gitmeme-token="lgtm">lgtm</a>

tas50 (changes requested):
We need 2.7 and 3.0 testing to match our previous 2.6 and 2.7 testing. 2.7 doesn't need to be testing every single bit, but it should still touch it.

tas50:
Add this to the .rubocop.yml to reduce the Chefstyle warnings:


```yaml
# This cop shouldn't alert on the helper / specs itself
Chef/Ruby/LegacyPowershellOutMethods:
  Exclude:
    - 'lib/chef/mixin/powershell_out.rb'
    - 'spec/functional/mixin/powershell_out_spec.rb'
```

tas50:
@lamont-granquist if you rebase this the chefstyle issue should go away

Gemfile
```diff
@@ -24,6 +24,9 @@ gem "cheffish", ">= 14"
 
 gem "chef-telemetry", ">=1.0.8" # 1.0.8 removes the http dep
 
+# yolo master for a bit until they release a ruby-3.0 compatible ffi
+gem "ffi", git: "https://github.com/ffi/ffi", branch: "master", submodules: true
```
tas50, line 28: Do we still need this now that 1.15 is out?

chef.gemspec
```diff
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_dependency "ohai", "~> 17.0"
   s.add_dependency "inspec-core", "~> 4.23"
 
-  s.add_dependency "ffi", ">= 1.9.25"
+  s.add_dependency "ffi" # FIXME: needs to be >= 1.14.3 at some point
```
tas50, line 41: we should just make this 1.15 for the m1 stuff

Issue #10921, opened 2021-01-22
load_current_resource for systemd_unit could be more efficient


joshuamiller01:
### Describe the Enhancement:
Establishing the status of a systemd_unit resource (per load_current_resource) currently shells out to systemctl five times; this could be refactored to perform a single shellout for speed and CPU wins. In my crude experimentation, a single systemctl call costs ~20ms; a single typical systemd_unit resource has three actions ([:create, :start, :enable]), each of which triggers the five shellouts. This reduces the steady-state cost from ~300ms (3 actions * 5 shellouts *20ms) to 60ms (~80% reduction).

### Describe the Need:
Anybody managing a systemd_unit resources on a systemd host

### Current Alternative
None

### Can We Help You Implement This?:
I'll submit a PR shortly.

lamont-granquist:
merged

PR #10925, opened 2021-01-23
load_current_resource for systemd_unit more efficiently

joshuamiller01:
Signed-off-by: Joshua Miller <joshmiller@fb.com>

## Description
Refactors how load_current_resource gathers state for systemd_unit resources by requesting all pieces of state in a single call and memoizing the results.

The only material change here is in how masked-runtime is handled; previously it was reported as 'masked', but I think that's incorrect. Asserting whether a unit is masked should be against the persistent system state (not volatile). The remaining systemd_unit resource states are all the same.

Performance outcomes are demonstrated with a naive loop, against a system where no such services are running:

```
(1..100).each do |x|
  systemd_unit "josh#{x}.service" do
    action [:stop, :disable]
  end
end
```

Previously this takes ~26s; with this patch, ~6s.  More actions (like :delete) compound this even more.

## Related Issue
https://github.com/chef/chef/issues/10921

## Types of changes

## Checklist:
- [x] All commits have been signed-off for [the Developer Certificate of Origin]

### DIFF
--- a/lib/chef/provider/systemd_unit.rb
+++ b/lib/chef/provider/systemd_unit.rb
@@ -55,6 +55,26 @@ class Chef
         end
       end
 
+      def systemd_unit_status
+        @systemd_unit_status ||= begin
+          # Collect all the status information for a unit and return it at once
+          # This may fail if we are managing a template unit (e.g. with '@'), in which case
+          # we just ignore the error because unit status is irrelevant in that case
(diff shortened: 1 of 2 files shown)


jaymzh (approved):
This actually fixes a few bugs too. Nice.

lamont-granquist (approved):

gene1wood:
Related PR : #10776

PR #10928, opened 2021-01-23
Chef 17: Remove windows service manager capabilities

tas50:
This is something we've recommended against using for many years now. Windows users should instead run Chef Infra Client using a Scheduled Task which is far more reliable and has the added benefit of requiring zero CPU or memory while not running.


jaymzh (approved):

lamont-granquist (approved):

tas50:
There's a few more things to do before this can get merged and I think we need to look at the MSI config still.

PR #10930, opened 2021-01-25
WIP - add support for action descriptions

### LINKED ISSUE #10660: Create a way of documentation actions in resources
We previously added a description method inside of actions, which we had hoped could be used to automation the documentation of resources. Due to how resources and providers work these methods are not available to use in the resource inspector. We need a way to document the action descriptions within the resource.

Proposal:

Extend the previous `actions`  method to allow accepting a hash instead of just an array, with the hash including descriptions.

Current usage:

```ruby
actions :create, :delete
```

Proposal usage:

```ruby
actions :create => 'Create the thing', :delete => 'delete the thing'
```

marcparadise:
## Description

This is a WIP proposed solution to resolve #10660.  This PR is not complete, but I did want to gather feedback as I'm not sure I like it as a direction; it touches internal state I'd rather not touch, but it may be a worthwhile tradeoff for the feature.  The net effects of this change are that actions can now be declared with descriptions; and these descriptions will be captured by the resource-inspector (and doc task).

With this change, the following is supported: 

```
action :action, description: "my own action" do 
...
end
```
and
```
allowed_actions action_a: "This is action a", 
  action_b: "This is action b" 
```
The original usage (just a symbol) is still supported as well. 

### Temporary breaking change

`allowed_actions` now returns a hash istead of an array. This resulted in a leakage of implementation so that we now have `.keys` scattered all over the place. This is temporary and will get reverted shortly - it was an easier to start with it like this while I pinned down all the places this (and related) fields were coming in/going out, but I don't intend to keep it this way. 

### Changes in behavior

LWRPs will now inherit their base class's  allowed_actions; this appears to be a change in which LWRPs previously would have only their own  declared actions as 'allowed_actions'.  Anyone know historically if we explicitly avoided that?
### Other
Existing unit tests are passing (except one it seems), but new tests are still needed; and I've seen a couple places where we can expand coverage around allowed_actions/action/default_action.

## Related Issue

#10660

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -120,7 +120,7 @@ class Chef
       @before = nil
       @params = {}
       @provider = nil
-      @allowed_actions = self.class.allowed_actions.to_a
+      @allowed_actions = self.class.allowed_actions
       @action = self.class.default_action
       @updated = false
       @updated_by_last_action = false
@@ -887,8 +887,10 @@ class Chef
     attr_writer :allowed_actions
 
     def allowed_actions(value = NOT_PASSED)
+      # TODO - why do we support assignment of allwoed actions
+      # from the instance?
       if value != NOT_PASSED
-        self.allowed_actions = value
+        @allowed_actions = Chef::Resource.convert_actions(value)
       end
       @allowed_actions
     end
@@ -994,7 +996,7 @@ class Chef
     #
     # @param actions [Array<Symbol,Hash>] The list of actions to add to allowed_actions.
     #
-    # @return [Array<Symbol>] The hash of allowed actions { ACTION => DESCRIPTION }
+    # @return Hash<Symbol,String> The hash of allowed actions { :ACTION => DESCRIPTION }
     def self.allowed_actions(*actions)
       if superclass.respond_to?(:allowed_actions)
         superclass.allowed_actions.each_pair { |k, v| add_allowed_action(k, v) }
@@ -1002,37 +1004,40 @@ class Chef
         add_allowed_action(:nothing, nil)
       end
 
-      actions.each do |action|
-        case action
-        when String, Symbol
-          add_allowed_action(action, nil)
-        when Hash
-          action.each_pair do |k, v|
-            add_allowed_action(k, v)
-          end
-        when Array
-          action.each do |k|
-            add_allowed_action(k, nil)
-          end
-
-        end
+      convert_actions(*actions).each_pair do |k, v|
+        add_allowed_action(k, v)
       end
       @allowed_actions
     end
 
-    # TODO - should this be addaction too?  Does it need to multi-input -
-    def self.allowed_actions=(value)
-      @allowed_actions = {}
-
-      # Hey this looks just like that function above, let's consolidate it?
-      case value
-      when String,Symbol
-
-      @allowed_actions = value.uniq
+    def self.allowed_actions=(*actions)
+      @allowed_actions = nil
+      convert_actions(*actions).each_pair do |k, v|
+        add_allowed_action(k, v)
+      end
     end
 
-    # @private
-    #
+    # private
+    def self.convert_actions(*actions)
+      {}.tap do |o|
+        actions.each do |action|
+          case action
+          when String, Symbol
+            o[action.to_sym] = nil
+          when Hash
+            action.each_pair do |k, v|
+              o[k] = v
+            end
+          when Array
+            action.each do |k|
+              o[k] = nil
+            end
+          end
+        end
+      end
+    end
+
+    # private
     def self.add_allowed_action(action, description)
       action = action.to_sym
       @allowed_actions ||= {}
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -152,7 +152,7 @@ class Chef
         arg.each do |action|
           validate(
             { action: action },
-            { action: { kind_of: Symbol, equal_to: allowed_actions } }
+            { action: { kind_of: Symbol, equal_to: allowed_actions.to_h.keys } }
           )
         end
         @action = arg
@@ -176,7 +176,7 @@ class Chef
       arg.map do |action|
         validate(
           { action: action },
-          { action: { kind_of: Symbol, equal_to: allowed_actions } }
+          { action: { kind_of: Symbol, equal_to: allowed_actions.to_h.keys } }
         )
         # the resource effectively sends a delayed notification to itself
         run_context.add_delayed_action(Notification.new(self, action, self, run_context.unified_mode))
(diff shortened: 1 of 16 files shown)


lib/chef/resource.rb
```diff
@@ -152,7 +152,7 @@ def action(arg = nil)
         arg.each do |action|
           validate(
             { action: action },
-            { action: { kind_of: Symbol, equal_to: allowed_actions } }
+            { action: { kind_of: Symbol, equal_to: allowed_actions.to_h.keys } }
```
marcparadise, line 155: All of these `to_h` will be made better once I update `allowed_actions` in my next push.

lib/chef/resource/lwrp_base.rb
```diff
@@ -74,13 +75,9 @@ def build_from_file(cookbook_name, filename, run_context)
 
         # Adds +action_names+ to the list of valid actions for this resource.
         # Does not include superclass's action list when appending.
+        # # TODO - do we want to support
```
marcparadise, line 78: That comment is also wrong, witht this change it _does_ include superclass's actions because that's the default behavior of `allowed_actions`.

spec/unit/resource_spec.rb
```diff
@@ -189,12 +188,16 @@
     let(:prior_resource) do
       prior_resource = Chef::Resource.new("funk")
       prior_resource.source_line
-      prior_resource.allowed_actions << :funkytown
```
marcparadise, line 192: But this original behavior will continue to not work right, even after changing allowed_actions. My first thought is that no subclass  should be directly modifying that list anyway.

lamont-granquist:
I'm going to have ponder the changes on subclassing.  Offhand it feels like subclasses should inherit because otherwise you violate Liskov.

Of course I don't think anyone should write resources that subclass because "prefer composition over inheritance" and all the abuse of inheritance in resources that never ends well and the Custom Resources DSL explicitly makes that impossible.

Mostly would need to check against the mysql and httpd cookbooks I think which have a lot of abuse of inheritance in them...  As long as they don't explode I'm leaning towards this being better.

tasks/docs.rb
```diff
@@ -266,6 +262,8 @@ def build_resource_data(name, data)
       r["resource"] = name
       r["resource_description_list"] = build_description(name, data["description"])
       r["resource_new_in"] = data["introduced"] unless data["introduced"].nil?
+      #require 'pry'; binding.pry
```
tas50, line 265:
```suggestion
```

lamont-granquist:
So I've got bad news.  `@allowed_actions` being an actual Array object seems to be a public API that has been used and abused quite a bit over the years:

```
activelamp_symfony/libraries/resource_symfony_project.rb:        @allowed_actions.push(:set_permissions)
boundary-meter/libraries/resource_boundary_meter.rb:        @allowed_actions.push(:create)
boundary-meter/libraries/resource_boundary_meter.rb:        @allowed_actions.push(:delete)
chef_server_omnibus/libraries/chef_server_omnibus_service.rb:        @allowed_actions.push(:reconfigure)
couchbase/libraries/node_resource.rb:        @allowed_actions.push(:modify)
couchbase/libraries/settings_resource.rb:        @allowed_actions.push :modify
couchbase/libraries/bucket_resource.rb:        @allowed_actions.push :create
couchbase/libraries/cluster_resource.rb:        @allowed_actions.push :create_if_missing
couchbase-ng/libraries/node_resource.rb:        @allowed_actions.push(:modify)
couchbase-ng/libraries/settings_resource.rb:        @allowed_actions.push :modify
couchbase-ng/libraries/bucket_resource.rb:        @allowed_actions.push :create
couchbase-ng/libraries/cluster_resource.rb:        @allowed_actions.push :create_if_missing
database/libraries/resource_database.rb:        @allowed_actions.push(:create, :drop, :query)
database/libraries/resource_database_user.rb:        @allowed_actions.push(:create, :drop, :grant, :revoke)
database/libraries/resource_postgresql_database_schema.rb:        @allowed_actions.push(:create, :drop)
database/libraries/resource_postgresql_database_user.rb:        @allowed_actions.push(:create, :drop, :grant, :grant_schema, :grant_table, :grant_sequence, :grant_function)
database/libraries/resource_sql_server_database_user.rb:        @allowed_actions.push(:alter_roles, :alter_sys_roles)
delivery-sugar/libraries/delivery_test_kitchen_resource.rb:          @allowed_actions.push(a.to_sym)
delivery-sugar/libraries/delivery_github_resource.rb:        @allowed_actions.push(:push)
delivery-sugar/libraries/delivery_chef_cookbook_resource.rb:        @allowed_actions.push(:upload)
delivery-sugar/libraries/delivery_push_job_resource.rb:        @allowed_actions.push(:dispatch)
delivery-sugar/libraries/delivery_inspec_resource.rb:        @allowed_actions.push(:test)
docker_rancher/libraries/resource_agent.rb:          @allowed_actions.push(:create, :delete, :nothing)
docker_rancher/libraries/resource_manager.rb:          @allowed_actions.push(:create, :delete, :nothing)
docker_rancher/libraries/resource_auth_local.rb:          @allowed_actions.push(:enable)
docker_rancher/libraries/resource_api_key.rb:          @allowed_actions.push(:create, :deactivate, :activate, :delete)
dsc/libraries/dsc_resource.rb:    @allowed_actions.push(:set)
dsc/libraries/dsc_resource.rb:    @allowed_actions.push(:test)
gitlab_omnibus/libraries/gitlab_omnibus_service.rb:        @allowed_actions.push(:reconfigure)
machine_tag/libraries/resource_machine_tag.rb:        @allowed_actions.push(:create, :delete)
mysqler/libraries/provider_chef_gem.rb:        @allowed_actions.push(:pristine)
ndb/libraries/resource_ndb.rb:#         @allowed_actions.push(:install, :dump, :init)
netdev/libraries/resource_vlan.rb:      @allowed_actions.push(:create, :delete)
netdev/libraries/resource_l2_interface.rb:      @allowed_actions.push(:create, :delete)
netdev/libraries/resource_interface.rb:      @allowed_actions.push(:create, :delete)
netdev/libraries/resource_lag.rb:      @allowed_actions.push(:create, :delete)
openstack-common/libraries/resource_database.rb:        @allowed_actions.push(:create, :drop, :query)
openstack-common/libraries/resource_database_user.rb:        @allowed_actions.push(:create, :drop, :grant, :revoke)
rightscale_backup/libraries/resource_rightscale_backup.rb:      #  @allowed_actions.push(:create, :restore, :cleanup)
rightscale_volume/libraries/resource_rightscale_volume.rb:      #  @allowed_actions.push(:create, :delete, :attach, :detach, :snapshot, :cleanup)
service_factory/libraries/provider.rb:        @allowed_actions.push(:install, :uninstall)
ssl_certificate/libraries/resource_ssl_certificate.rb:        @allowed_actions.push(@action)
symfony_project/libraries/resource_symfony_project.rb:        @allowed_actions.push(:set_permissions)
tgw_uwsgi/libraries/resource_uwsgi.rb:        @allowed_actions.push(:create, :delete, :enable, :disable)
tomcat6/libraries/tomcat_manager.rb:        @allowed_actions.push(:install)
tomcat6/libraries/tomcat_manager.rb:        @allowed_actions.push(:start)
tomcat6/libraries/tomcat_manager.rb:        @allowed_actions.push(:stop)
tomcat6/libraries/tomcat_manager.rb:        @allowed_actions.push(:update)
tomcat6/libraries/tomcat_manager.rb:        @allowed_actions.push(:undeploy)
windows-sdk/libraries/windows_sdk_feature.rb:        @allowed_actions.push(:install, :uninstall)
wmq/libraries/queue.rb:        @allowed_actions.push(:create)
wsus-server/libraries/base_resource.rb:      @allowed_actions.push(:configure)
xml_file/libraries/xml_file.rb:        allowed_actions.push(:edit)
zap/libraries/default.rb:      @allowed_actions.push(:delete, :remove, :disable)
```

There's a bunch more which are just doing `@allowed_actions = [ <...stuff...>]`

marcparadise:
Thanks @lamont-granquist . I was hoping at first that continuing to keep it exposed as an array would smooth over usage differences, but not when direct access is common.  

Okay, that  means we'll have to split managing actions from adding descriptions in `Resource`.  Shouldn't be too bad, and should still be able to support declaring action docs with the same DSL already in use here.

Issue #10938, opened 2021-01-26
Add CLI output to the Compliance Phase


aknarts:
### Describe the Enhancement:
In deprecated Audit phase the results of the audit were put at the end of the chef converge log
![Screenshot from 2021-01-26 21-37-35](https://user-images.githubusercontent.com/252905/105902313-c1b3bf00-601e-11eb-857c-079b4a154d1c.png)

This was removed and is now replaced by inspec and the Compliance Phase

### Describe the Need:
We want to be able to easily see the results of the compliance during the chef converge.

### Current Alternative
The only alternative is to create a wrapper around chef client that parses the resulting json file and/or rely on Chef Automate for the reporting

### Can We Help You Implement This?:
I do have a PR ready that implements this in a reasonable way for us: https://github.com/chef/chef/pull/10939

tas50:
Merged

PR #10939, opened 2021-01-26
Compliance cli report

aknarts:
## Description
Add a cli output option to the node["audit"]["reporter"]

This results in the following output at the end of the chef run:
Screenshot from 2021-01-26 21-33-30

## Related Issue
https://github.com/chef/chef/issues/10938

## Types of changes

## Checklist:


tas50:
@kagarmoe Would this need to be fit into the open PR we have for docs in the Compliance Phase?

lib/chef/compliance/reporter/cli.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        private
+
+        def format_result(result)
+          output = []
+          found = false
+          if result[:status] == "failed"
+            if result[:code_desc]
+              found = true
+              output << "         \033[31m\xE2\x9D\x8C #{result[:code_desc]}\033[0m"
```
lamont-granquist, line 33: those codes should be using the tty gem, not just doing in-line escape sequences in strings

lamont-granquist (approved):

lib/chef/compliance/reporter/cli.rb
aknarts: I did look into the tty-gem, while I agree that there should be a better way of doing this(like checking if TTY::Color.support?) I did not want to introduce a gem dependency. I can remove the special characters, and colours, but they do provide some clarity into the success of the result. There are not that many uses of the tty gems.

So I did remove the hardcoded colours and characters and replaced it with pastel that is already being used elsewhere.

lamont-granquist: i'm fairly certain that knife already depends on the tty gem and that is definitely the direction that we'd like to go with output formatting

lamont-granquist: oh nevermind me, i see you've updated to use Pastel and that's what we're actually using in knife for colors...

aknarts: Yes I checked your codebase for what is being used because I know there is coloured output and adjusted accordingly.

aknarts: I am testing this with the latest Chef 16 and in the audit cookbook(the code is pretty much the same, just a different class in the cookbook)

tas50:
Thanks @aknarts

tas50:
@aknarts I used this in our Compliance Phase webinar today. It was a great way to visualize the Compliance Phase running. Thanks!

aknarts:
@tas50 Glad it is of some use to somebody else than us.

PR #10952, opened 2021-01-29
Add support for resource action descriptions

marcparadise:
This will allow us to generate better documentation by including
action descriptions in the resource help, and includes updates to the
resource inspector and doc generation task to include action descriptions
in resource action data.

This change permits actions to be defined with the optional new
'description' parameter as follows:

```
action :my_action, description: "my_action changes the state of this world"
...
end
```

Action description can be looked up using the
`.action_description(action)` method of the resource class.  Action
descriptions are inheritable and overridable.

---

Replaces #10930 

Implements #10660

### DIFF
--- a/lib/chef/resource.rb
+++ b/lib/chef/resource.rb
@@ -1062,6 +1062,7 @@ class Chef
     # action for the resource.
     #
     # @param name [Symbol] The action name to define.
+    # @param description [String] optional description for the action
     # @param recipe_block The recipe to run when the action is taken. This block
     #   takes no parameters, and will be evaluated in a new context containing:
     #
@@ -1071,12 +1072,35 @@ class Chef
     #
     # @return The Action class implementing the action
     #
-    def self.action(action, &recipe_block)
+    def self.action(action, description: nil, &recipe_block)
       action = action.to_sym
       declare_action_class
       action_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
+      # Accept any non-nil description, which will correctly override
+      # any specific inherited description.
+      action_descriptions[action] = description unless description.nil?
       default_action action if Array(default_action) == [:nothing]
+    end
+
+    # Retrieve the description for a resource's action, if
+    # any description has been included in the definition.
+    #
+    # @param action [Symbol,String] the action name
+    # @return the description of the action provided, or nil if no description
+    # was defined
+    def self.action_description(action)
+      action_descriptions[action.to_sym]
+    end
+
+    # @api private
+    #
+    # @return existing action description hash, or newly-initialized
+    # hash containing action descriptions inherited from parent Resource,
+    # if any.
+    def self.action_descriptions
+      @action_descriptions ||=
+        superclass.respond_to?(:action_descriptions) ? superclass.action_descriptions.dup : { nothing: nil }
     end
 
     # Define a method to load up this resource's properties with the current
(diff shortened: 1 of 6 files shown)


marcparadise:
https://github.com/chefspec/chefspec/pull/1001 - this PR will fix the failing chefspec gem test.

lib/chef/resource.rb
```diff
@@ -1071,14 +1091,34 @@ def self.default_action=(action_name)
...
       declare_action_class
       action_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
+      # TODO So this is weird and makes me think we shouldn't accept description in
+      # allowed_actions:  if someone for some reason has an action with a description,
+      # and allowed_actions also gives a description for the same action - which
+      # description prevails is load order dependent.
+      #
+      # In any case, accept any non-nil description, which will correctly override
+      # any specific inherited description.
```
jaymzh, line 1105: I definitely thing that the description should be with the action and we should not accept descriptions in `allowed_actions`. Allowed_actions - by it's very name, is a list of existing actions, and nothing more. Actions themselves should define their description just like resources themselves can.

marcparadise: That makes sense. The reason I included it there in the first place is that  `allowed_actions` itself will otherwise add the named actions as valid actions - allowing an action to be declared without ever defining it.

tas50: As long as we maintain our ability to document both custom resources and LWRP/HWRP style resources it's fine with me to drop this.

lib/chef/resource.rb
```diff
@@ -1071,14 +1072,37 @@ def self.default_action=(action_name)
...
     #
-    def self.action(action, &recipe_block)
+    def self.action(action, description: nil, &recipe_block)
       action = action.to_sym
       declare_action_class
       action_class.action(action, &recipe_block)
       self.allowed_actions += [ action ]
+      # Accept any non-nil description, which will correctly override
+      # any specific inherited description.
+      action_descriptions[action] = description unless description.nil?
```
jaymzh, line 1082: I don't think the `unless` here is necessary. `action_descrptions[action]` will be `nil` if description is `nil`, and that's fine, ruby finds that falsey.

lib/chef/resource.rb
```diff
@@ -1196,9 +1220,9 @@ def self.preview_resource(value = nil)
...
-    FORBIDDEN_IVARS = %i{@run_context @logger @not_if @only_if @enclosing_provider @description @introduced @examples @validation_message @deprecated @default_description @skip_docs @executed_by_runner}.freeze
+    FORBIDDEN_IVARS = %i{@run_context @logger @not_if @only_if @enclosing_provider @description @introduced @examples @validation_message @deprecated @default_description @skip_docs @executed_by_runner @action_descriptions}.freeze
     # HIDDEN_IVARS do not show up when the resource is displayed to the user as text (ie. in the error inspector output via #to_text)
-    HIDDEN_IVARS = %i{@allowed_actions @resource_name @source_line @run_context @logger @name @not_if @only_if @elapsed_time @enclosing_provider @description @introduced @examples @validation_message @deprecated @default_description @skip_docs @executed_by_runner}.freeze
+    HIDDEN_IVARS = %i{@allowed_actions @resource_name @source_line @run_context @logger @name @not_if @only_if @elapsed_time @enclosing_provider @description @introduced @examples @validation_message @deprecated @default_description @skip_docs @executed_by_runner @action_descriptions}.freeze
```
jaymzh, line 1225: why make this hidden/forbidden? Seems like a good thing to dump when people look at the objects.

jaymzh (commented):
I like this. few minor comments.

lib/chef/resource.rb
lamont-granquist: no that's going to be messy, definitely don't want it when its converted to JSON at all, and i don't want it in the debugging output.

lib/chef/resource_inspector.rb
```diff
@@ -41,7 +41,11 @@ def self.extract_resource(resource, complete = false)
       data[:description] = resource.description
       # data[:deprecated] = resource.deprecated || false
       data[:default_action] = resource.default_action
-      data[:actions] = resource.allowed_actions
+      data[:actions] = {}
+      resource.allowed_actions.each do |action|
+        data[:actions][action] = resource.action_description(action)
+      end
```
lamont-granquist, line 47: FWIW this could have been each_with_object, but it only saves a line at the cost of added overhead of knowing that method.

lib/chef/resource.rb
marcparadise: This handles the case where if a base class has a description for `action`, it won't get replaced with `nil` if the inheriting resource overrides that action without a description. It may already be nil if the base class didn't describe it, but in case it's not we won't overwrite it with a nil.

PR #10960, opened 2021-02-02
windows_certificate: Fix the `user_store` property to actually install certificates to the user store 

johnmccrae:
Updated chef to use the CurrentUser certificate store. Previously it only used LocalMachine. Added tests to the windows.rb file, added a reserved name to the cspell file. Updated Gem files to require 0.5.3 of win32-certstore


lib/chef/resource/windows_certificate.rb
```diff
@@ -75,8 +75,9 @@ class WindowsCertificate < Chef::Resource
         description: "The certificate store to manipulate.",
         default: "MY", equal_to: ["TRUSTEDPUBLISHER", "TrustedPublisher", "CLIENTAUTHISSUER", "REMOTE DESKTOP", "ROOT", "TRUSTEDDEVICES", "WEBHOSTING", "CA", "AUTHROOT", "TRUSTEDPEOPLE", "MY", "SMARTCARDROOT", "TRUST", "DISALLOWED"]
 
+      # prior to Chef 16.10 this property was ignored.
       property :user_store, [TrueClass, FalseClass],
-        description: "Use the user store of the local machine store if set to false.",
+        description: "Use the `CurrentUser` store instead of the default `LocalMachine` store.",
```
tas50, line 80:
```suggestion
        description: "Use the `CurrentUser` store instead of the default `LocalMachine` store. Note: Prior to #{ChefUtils::Dist::Infra::CLIENT}. 16.10 this property was ignored.",
```

lib/chef/resource/windows_certificate.rb
```diff
@@ -75,8 +75,9 @@ class WindowsCertificate < Chef::Resource
         description: "The certificate store to manipulate.",
         default: "MY", equal_to: ["TRUSTEDPUBLISHER", "TrustedPublisher", "CLIENTAUTHISSUER", "REMOTE DESKTOP", "ROOT", "TRUSTEDDEVICES", "WEBHOSTING", "CA", "AUTHROOT", "TRUSTEDPEOPLE", "MY", "SMARTCARDROOT", "TRUST", "DISALLOWED"]
 
+      # prior to Chef 16.10 this property was ignored.
```
tas50, line 78:
```suggestion
```

lamont-granquist:
so to fix this you'll probably need to do it mostly manually.

you need to find the commit which is directly BEFORE yours and get the gitsha of it and do a reset (bit WITHOUT --hard to leave the repo contents as-is):

```
git reset abcdef1234..
```

you can always get back to normal if you pick the wrong one:

```
git reset origin/jfm/windows_certificate_resource
```

the end result should be a `git diff` which is identical to this PR and which contains no other changes.  if there's other changes in here or missing changes then you picked the wrong commit to reset to.

then you need to remove the file diffs of the stuff that you are not changing.  this is deleting changes without any record of them so you kind of need to be careful:

```
git checkout -- CHANGELOG.md Gemfile.lock VERSION
```

keep removing changes and checking with `git diff` until the repo starts to look like only the changes you want.

if you have files which are mix-n-match of your changes and changes that have been pulled in, then you need to manually edit those and get them to the point where they only show the changes you want.

then double check with `git diff` that it looks like you have ALL your changes and NONE of the stray changes.  then triple check.

then just commit the changes normally:

```
git commit -as
```

then force push since you've rewritten history:

```
git push -f origin jfm/windows_certificate_resource
```

that's the real point of absolutely no return.  then you lose this PR to refer to and overwrite the origin and it gets very difficult to get any information back.

before this point you can start all over completely with:

```
git reset --hard origin/jfm/windows_certification_resource
```

after you force push though you'll rewrite the origin and then you're probably manually recreating changes (although if you keep the gitsha of the old commit you can reset --hard to that gitsha provided you do so before git runs a garbage collection since it'll still technically be hanging around, but you really need to never rely on that -- if you ever get into that position you need to immediately stop what you're doing and get some help recovering the orphaned commit before you trigger a garbage collection accidentally and its gone).

tas50:
Using #10977  instead

PR #10977, opened 2021-02-03
windows_certificate: Fix the `user_store` property to actually install certificates to the user store

tas50:
This is #10960 without the merge conflict


johnmccrae (approved):

lib/chef/resource/windows_certificate.rb
```diff
@@ -161,35 +161,69 @@ class WindowsCertificate < Chef::Resource
...
+            store.get(thumbprint)
+          end
+        end
+
+        # this Regex verifies whether a string is a hex value (ie. thumbprint) or not
+        def check_string(string)
+          string.scan(/\H/).empty?
+        end
+
+        def get_thumbprint(store_name, location, source)
```
marcparadise, line 204: If you named this `is_thumbprint` you could do both the length and hex check in the same function. It would  also simplify usage and  be more descriptive in the places it's being called from.

lib/chef/resource/windows_certificate.rb
```diff
@@ -230,13 +264,15 @@ def export_cert(cert_obj, cert_path)
           out_file.close
         end
 
-        def cert_location
```
marcparadise, line 233: Minor, but `cert_location[0]` scans better than `get_cert_location[0]`.

lib/chef/resource/windows_certificate.rb
```diff
@@ -230,13 +264,15 @@ def export_cert(cert_obj, cert_path)
...
+        # this array structure is solving 2 problems. The first is that we need to have support for both the CurrentUser AND LocalMachine stores
+        # Secondly, we need to pass the proper constant name for each store to win32-certstore but also pass the short name to powershell scripts used here
+        def get_cert_location
+          new_resource.user_store ? ["CurrentUser", CERT_SYSTEM_STORE_CURRENT_USER] : ["LocalMachine", CERT_SYSTEM_STORE_LOCAL_MACHINE]
         end
 
         def cert_script(persist)
           cert_script = "$cert = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2"
-          file = Chef::Util::PathHelper.cleanpath(new_resource.source)
+          file = Chef::Util::PathHelper.cleanpath(new_resource.source, get_cert_location[1])
```
lamont-granquist, line 275: You've got magic `0` and `1` numbers here which makes for fairly poor reading.  I also don't like the `get_` prefix (there's actually a rubocop rule to not do this, but we haven't enabled it because we violate it so much, but I do sort of agree that we should avoid that where possible).

So, I'd suggest something like:

```
def cert_location_name
  new_resource.user_store ? "CurrentUser" : "LocalMachine"
end

def cert_location_symbol
  new_resource.user_store ? CERT_SYSTEM_STORE_CURRENT_USER : CERT_SYSTEM_STORE_LOCAL_MACHINE
end
```

That is more verbose in the declaration, but it leads to a lot easier to read with the substitutions `cert_location_name` for `get_cert_location[0]` and `cert_location_symbol` for `get_cert_location[1]`.

lamont-granquist (changes requested):

lib/chef/resource/windows_certificate.rb
johnmccrae: The powershell code in here needs to see CurrentUser instead of CERT_SYSTEM_STORE_CURRENT_USER

lib/chef/resource/windows_certificate.rb
```diff
@@ -161,35 +161,69 @@ class WindowsCertificate < Chef::Resource
...
         # But if it is present but expired, it returns a Boolean: false
         # Otherwise, it returns a Boolean: true
+        # updated this method to accept either a subject name or a thumbprint - 1/29/2021
+
         def verify_cert(thumbprint = new_resource.source)
-          store = ::Win32::Certstore.open(new_resource.store_name)
-          store.valid?(thumbprint)
+          store = ::Win32::Certstore.open(new_resource.store_name, store_location: get_cert_location[1])
+          status = check_string(thumbprint).to_s
+          if thumbprint.length == 40 && status == "true"
```
lamont-granquist, line 221: that looks like you can `if check_string(thumbprint) && thumbprint.length == 40` here

lib/chef/resource/windows_certificate.rb
johnmccrae: done, thanks

lib/chef/resource/windows_certificate.rb
```diff
@@ -161,35 +161,69 @@ class WindowsCertificate < Chef::Resource
...
 
         def delete_cert
           store = ::Win32::Certstore.open(new_resource.store_name)
-          store.delete(new_resource.source)
+          if check_string(new_resource.source) && new_resource.source.length == 40
+            store.delete(new_resource.source)
+          else
+            thumbprint = powershell_exec!(get_thumbprint(new_resource.store_name, get_cert_location[0], new_resource.source)).result
+            store.delete(thumbprint)
+          end
```
lamont-granquist, line 186: You should extract a helper method out of this so that all three of these similar blocks of code get extracted...

```
store.delete(thumbprintcode(new_resource.source))
store.get(thumbprintcode(new_resource.source))
store.valid?(thumbprintcode(thumbprint))
```
with:

```
def thumbprintcode(thumbprint)
  if check_string(thumbprint) && thumbprint.length == 40
    store.get(thumbprint)
  else
    thumbprint = powershell_exec!(get_thumbprint(new_resource.store_name, get_cert_location[0], new_resource.source)).result
    store.get(thumbprint)
  end
end
```

or:

```
def thumbprintcode(thumbprint)
  if !check_string(thumbprint) || thumbprint.length != 40 
    thumbprint = powershell_exec!(get_thumbprint(new_resource.store_name, get_cert_location[0], new_resource.source)).result
  end
  store.get(thumbprint)
end
```

you could also invert check_string, add the length != 40 check to it and turn it into `def invalid_thumbprintcode?` and then you'd have:

```
def thumbprintcode(thumbprint)
  if invalid_thumbprintcode?(thumbprint)
    thumbprint = powershell_exec!(get_thumbprint(new_resource.store_name, get_cert_location[0], new_resource.source)).result
  end
  store.get(thumbprint)
end
```

With that method now looking like:

```
def invalid_thumbprintcode?(code)
  !(code.scan(/\H/).empty? && code.length == 40)
end
```

marcparadise (dismissed):

lib/chef/resource/windows_certificate.rb
johnmccrae: there we go, thanks

lib/chef/resource/windows_certificate.rb
johnmccrae: Lamont suggested the same thing. I added a second method and changed them to ps_cert_location and native_cert_location to reflect PowerShell versus the actual name of the store.

lamont-granquist (approved):

PR #10993, opened 2021-02-07
Fix downgrades in apt_package

jaymzh:
`apt_package` never even looks at `new_resource.version` when deciding
what version to install. That's silly.

This fixes that. It requires the version be exact (as the `yum_package`
did for most of Chef's history), and if you specify a non-existent
version, it'll raise an exception.

The unit tests for this provider are a bit of a mess, so I kinda stuck
in the tests where I could.

```
[2021-02-07T15:59:35-08:00] TRACE: apt_package[hub] installed version for hub is 2.14.2-1vcrs.1
[2021-02-07T15:59:35-08:00] TRACE: apt_package[hub] candidate version for hub is 2.14.2-1vcrs.1
[2021-02-07T15:59:35-08:00] TRACE: apt_package[hub] hub 2.14.2-1vcrs.1 needs updating to 2.7.0~ds1-1vcrs.1
[2021-02-07T15:59:38-08:00] INFO: apt_package[hub] installed hub at 2.7.0~ds1-1vcrs.1
```

### DIFF
--- a/spec/unit/provider/package/apt_spec.rb
+++ b/spec/unit/provider/package/apt_spec.rb
@@ -48,6 +48,21 @@ describe Chef::Provider::Package::Apt do
     @timeout = 900
   end
 
+  def ubuntu1804downgrade_stubs
+    so = instance_double(Mixlib::ShellOut, stdout: "apt 1.6~beta1 (amd64)\notherstuff\n")
+    so2 = instance_double(Mixlib::ShellOut, error?: false)
+    allow(@provider).to receive(:shell_out).with("apt-get --version").and_return(so)
+    allow(@provider).to receive(:shell_out).with("dpkg", "--compare-versions", "1.6~beta1", "gt", "1.1.0").and_return(so2)
+  end
+
+  def ubuntu1404downgrade_stubs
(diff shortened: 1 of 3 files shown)


spec/unit/provider/package/apt_spec.rb
```diff
@@ -590,7 +650,7 @@ def ubuntu1404downgrade_stubs
       end
     end
 
-    describe "#action_install" do
+    describe "#action_upgrade" do
```
jaymzh, line 653: all of these tests call action upgrade, not action install, so I fixed the name.

jaymzh:
OK. tested and linted... should be ready for review

jaymzh:
hmm looks like I broke multipackage cookbook thing :/

jaymzh:
OK @lamont-granquist / @tas50 - all relevant tests finally pass... there's some kitchen failures, but that's around Ohai versions, seems pretty unrelated.

lamont-granquist (approved):

tas50:
I assume we don't want to backport this since the raises behavior is technically breaking?

jaymzh:
I could argue either way. 16 is pretty broken today - it blindly ignores `version`, so it seems OK to backport, but also, raising when it didn't before also sucks... so.... I'm open.

PR #11020, opened 2021-02-11
Split a knife gem out of chef

### LINKED ISSUE #11019: Split the knife gem out of the chef gem
For Chef Infra Client 17 we have a great opportunity to slim down the size of the chef install while resolving some longstanding customer security concerns by splitting knife out of the chef gem.

## Why

chef the repo historically was everything chef including the server. It was pretty natural at to just gem install chef and manage nodes that way. A lot has changed since that time, especially with the introduction of ChefDK and Chef Workstation. Users aren't (or at least shouldn't) be gem installing chef in order to manage nodes. That task should be performed in Workstation either on a system with the Workstation package installed or a workstation docker container.

Knife being in each Infra Client is pretty problematic. Knife itself has bloated over the years as have the dependencies it brings along to each client install. We'd like to pull more into the standard knife offering, but that's difficult as each new plugin adds more dependencies to each Infra Client install. For instance, adding the knife-windows functionality brings in 3 more winrm-related gems that would have to ship with each Infra Client.

Additionally, security audits have raised concerns having management code on each client node in their fleet. Removing knife from the standard Infra Client package resolves that audit concerns.

## What to do

Split current knife code into a knife gem. That gem code would continue to live in the chef/chef repo. This would be really similar to our chef-utils and chef-config pattern. Version bumping would be handled via our existing pipelines as would promotion. This keeps the maintenance low. Chef Workstation would include the new knife dependency and the omnibus chef package would no longer directly depend on knife, allowing us to shed a good number of dependencies.

marcparadise:
This isn't anywhere near ready for merging. 
This change will split knife into its own gem, resolves #11019


knife/lib/chef/knife/client_bulk_delete.rb
```diff
@@ -0,0 +1,104 @@
+#
```
marcparadise, line 1: This should have been a rename-then-change, not sure why this and others are showing new.  THat'l make this more challenging to review...

PR #11036, opened 2021-02-13
Remove the legacy chef-apply command

tas50:
chef-apply is a low value command with a high value name which conflicts with the actual code of the chef run command. It's been targeted for Infra Client 17 removal.


kagarmoe:
Related docs PRs: 

https://github.com/chef/chef-web-docs/pull/2961
https://github.com/chef/chef-workstation/pull/1776

jaymzh:
I use this tool quite a lot. you said the same code exists in the `chef` tool, but that's part of WS, not client. What's the replacement here?

lamont-granquist:
Yeah I'm a fan of killing the docs but not a fan of removal.  I could be persuaded to rename it, but `chef apply` got renamed to `chef run` and no longer collides with that name, and people crawling around the internals are on their own.  We haven't seen any bugs over confusion with chef apply in literally years and I don't think its a thing that needs fixing at this point.  Mostly that was our own goal from putting `chef-apply` in front of users in learnchef back in the day, which we stopped doing.

tas50:
Docs are gone, feature stays for now

PR #11055, opened 2021-02-16
Experimental: attempt to directly load known knife commands

marcparadise:
Instead of loading all of them then searching, try to
load just the requested commanmd if we can find it by path.

### DIFF
--- a/lib/chef/knife/acl_base.rb
+++ b/lib/chef/knife/acl_base.rb
@@ -22,6 +22,7 @@ require_relative "../knife"
 class Chef
   class Knife
     module AclBase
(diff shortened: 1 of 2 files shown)


lib/chef/knife/acl_base.rb
```diff
@@ -22,6 +22,7 @@
 class Chef
   class Knife
     module AclBase
+      Kernel.puts "LOADING ANYWAY HAHAHAH"
```
marcparadise, line 25: canary to be deleted, so I can tell easily if I loaded all commands anyway.

PR #11079, opened 2021-02-19
Remove the Ohai specs for our CI config

tas50:
These are preventing a release and the fix is in the build script for
the CI boxes now


lamont-granquist:
These still have a point since they surface issues on anyone's laptops or any other environments running the tests.

And the whole purpose behind these tests is that just that I couldn't explain to @christopher-snapp how I debugged the problem (and it boiled down to "have 10 years of experience and get lucky in picking a simple test to debug") so I wrote a test to surface the issue without anyone needing to directly engage me (and it worked).  It was generated as the result of pair debugging with releng.

tas50:
It just caused me a ton of work so next time it's someone else

PR #11086, opened 2021-02-22
Update resolver cookbook usage in test-kitchen tests

ramereth:
Recently, the Sous Chefs released a new resource driven resolver cookbook and removed the default recipe, thus breaking tests. This should fix the tests related to this.

### DIFF
--- a/kitchen-tests/cookbooks/end_to_end/recipes/macos.rb
+++ b/kitchen-tests/cookbooks/end_to_end/recipes/macos.rb
@@ -23,7 +23,10 @@ timezone "America/Los_Angeles"
 
 include_recipe "ntp"
 
-include_recipe "resolver"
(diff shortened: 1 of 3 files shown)


ramereth:
For some reason, test-kitchen is not executing any of the recipes on the linux docker hosts. I cannot for the life of me figure out why that is happening.

kitchen-tests/cookbooks/end_to_end/recipes/macos.rb
```diff
@@ -23,7 +23,10 @@
 
 include_recipe "ntp"
 
-include_recipe "resolver"
+resolver_config '/etc/resolv.conf' do
```
tas50, line 26: This is going to fail chefstyle here

PR #11095, opened 2021-02-24
Dup default property values

lamont-granquist:
This largely makes frozen default value errors go away.  Each
copy of a resource will get a dup of the default.  Defaults are
still frozen so we have defense-in-depth against them being
mutated.  This eliminates the need to `default: lazy{ {} }`
things like Hashes or Arrays.

### DIFF
--- a/lib/chef/property.rb
+++ b/lib/chef/property.rb
@@ -412,6 +412,7 @@ class Chef
             # Otherwise, we have to validate it now.
             value = input_to_stored_value(resource, default, is_default: true)
           end
+          value = deep_dup(value)
           value = stored_value_to_output(resource, value)
 
           # If the value is mutable (non-frozen), we set it on the instance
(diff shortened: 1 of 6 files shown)


spec/unit/property_spec.rb
```diff
@@ -708,10 +709,10 @@ def blah
           expect(resource.x).to eq "hi1"
           expect(Namer.current_index).to eq 1
         end
-        it "when x is retrieved, coercion is run each time" do
+        it "when x is retrieved, coercion is run exactly once" do
           expect(resource.x).to eq "101"
-          expect(resource.x).to eq "102"
-          expect(Namer.current_index).to eq 2
+          expect(resource.x).to eq "101"
+          expect(Namer.current_index).to eq 1
```
lamont-granquist, line 715: This is one of the weirdest tests I've ever seen in the codebase and I'm unsure of what its purpose is, or if it was just sort of documentation-through-code of an edge condition.

tas50:
Add a link to the 17 section of the release notes to this PR so I can put them together come april

lib/chef/property.rb
```diff
@@ -748,5 +749,22 @@ def coerce_and_validate(resource, value)
 
       result
     end
+
+    # recursively dup the value
+    def deep_dup(value)
```
jaymzh, line 754: Do we really not have a deep_dup somewhere? We do this in the Node constnatly, right?

spec/unit/property_spec.rb
jaymzh: we are changing the behavior here.... which this behavior seems better.... but like this will break SOMETHING. We'll need to document that.

lamont-granquist: I literally have no idea what this would ever break though or why anyone would ever have done this.

lib/chef/property.rb
lamont-granquist: That is all overly complicated stuff hanging off of the various attribute classes, which would need to be injected into Hash and Array to function similarly.

jaymzh (approved):

PR #11123, opened 2021-03-01
Update update_dep to allow passing in the dep to update

tas50:
We need to know the dep we're updating throughout this script and if
it's not a gem we don't know the name. Allow passing in the name via arg

### DIFF
--- a/.expeditor/update_dep.sh
+++ b/.expeditor/update_dep.sh
@@ -11,11 +11,14 @@
 
 set -evx
 
+REPONAME=$(echo $EXPEDITOR_REPO | cut -d '/' -f 2)
+DEPNAME="${EXPEDITOR_GEM_NAME:-${REPONAME:?Could not find gem name}}"
+
 function new_gem_included() {
-  git diff | grep -E '^\+' | grep "${EXPEDITOR_GEM_NAME} (${EXPEDITOR_VERSION})"
+  git diff | grep -E '^\+' | grep "${DEPNAME} (${EXPEDITOR_VERSION})"
 }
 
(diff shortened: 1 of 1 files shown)


.expeditor/config.yml
```diff
@@ -211,8 +211,8 @@ subscriptions:
   # NOTE: The branch of Ohai here needs to be updated when setting up a stable branch of chef/chef
   - workload: chef/ohai:master_completed:pull_request_merged:chef/ohai:master:*
     actions:
-      - bash:.expeditor/update_dep.sh
+      - bash:.expeditor/update_dep.sh ohai
```
nkierpiec, line 214: Need to remove the `ohai` here

.expeditor/config.yml
```diff
@@ -211,8 +211,8 @@ subscriptions:
...
   # NOTE: When the stable branch of chef/chef is being cut you probably want to remove this subscription
   - workload: chef/chefstyle:master_completed:pull_request_merged:chef/chefstyle:master:*
     actions:
-      - bash:.expeditor/update_dep.sh
+      - bash:.expeditor/update_dep.sh chefstyle
```
nkierpiec, line 218: Need to remove the `chefstyle` here

nkierpiec (changes requested):
Small fixes to the config then you should be golden

nkierpiec (approved):

tduffield:
So the only issue here is that `pull_request_merged` does not have an `EXPEDITOR_VERSION` variable. You'll likely need to fetch that from the VERSION file via a curl command or have a mode of operating where it does not exist.

